Issue #1165, opened 2014-12-17
transition to SVG icons


nebulon42:
I think most will agree that moving all icons to SVG is beneficial (line and pattern symbolizers will have to wait for the transition to Mapnik 3.0). This issue is intended to gather ideas how the transition could work. Issues #632, #658 and #659 are related.

As pointed out by @math1985 in #659 there are currently several efforts for icon repositories. While a central repository is a good idea, we might have to alter icons before using them here and hence not every icon might be useful for each project. So should we create our own repository or just keep track of the icons we include?

https://github.com/twain47/Open-SVG-Map-Icons is a good starting point for a repository, but for me it is not clear what its current status is. Is it still active? While the icons are very good, I also have the impression that some are too complex for small sizes and some have to be altered to be used in this project. I don't follow the puristic view on SVG code of @pnorman regarding Inkscape (I might be wrong on that, please correct me), but still think that code quality matters a bit here (file size).

I'm not sure what the best approach might be. SJJB icons are a good starting point, but some will require alteration or even re-drawing for using them with CartoCSS (e.g. colouring). So I think we cannot use them directly. Should we fork the repository and evolve it further for using the icons with ths project? I would be up for maintaining an icon repository, but this might be overkill and maybe we should concentrate on the icons we will really use.

repository for my icons: https://github.com/nebulon42/osmic

declined (just for reference)
`amenity=car_sharing` (~~#1283~~) - removed in #1901 

completed
- [x] `amenity=hospital` (#1169)
- [x] `amenity=pharmacy` (#1169)
- [x] `amenity=atm` (#1171)
- [x] `amenity=bank` (#1171)
- [x] `tourism=museum` (#1174) 
- [x] `amenity=recycling` (#1175)
- [x] `leisure=playground` (#1177)
- [x] `amenity=place_of_worship` (#1185)
- [x] `amenity=bar` (#1188)
- [x] `amenity=pub` (#1188)
- [x] `amenity=cafe` (#1188)
- [x] `amenity=biergarten` (#1188)
- [x] `amenity=fast_food` (#1188)
- [x] `amenity=restaurant` (#1188)
- [x] `amenity=cinema` (#1211)
- [x] `shop=bakery` (#1212)
- [x] `tourism=camp_site` (#1218)
- [x] `amenity=police` (#1220)
- [x] `amenity=fire_station` (#1220)
- [x] `amenity=embassy` (#1235)
[...]

PR #1218, opened 2015-01-11
move camping icon to SVG

nebulon42:
Implements part of #1165 by moving the camping icon to SVG.

before
camping_before

after
camping_after


math1985:
Thanks!

PR #1220, opened 2015-01-13
move emergency icons to SVG

nebulon42:
Implements part of #1165 by changing `amenity=police` and `amenity=fire_station` to SVG.

before
fire_before police_before

after
fire_after  police_after

Rationale for police station:
I consider handcuffs to fragile and complex for small sizes. The proposed icon is also used by the HOT style and (a variant) is commonly used to depict border control, so it should be quite familiar.


math1985:
Looks fine to me.

dieterdreist:
I like your fire icon, it is much more recognizable at this size and looks
great. The police icon looks a bit like customs to me.

mkoniecz:
The police is not obvious - enough to guess broad range, but it is hard to guess that it is police. But old icon for me is great one to depict prisons, so it is a significant improvement.

Fire station is great and much better. I thought about depicting truck with ladder, but I am not sure is it universal and possible to fit into a small icon.

SomeoneElseOSM:
(late unfortunately, but) I just spotted the new "police" icon yesterday and had to look at the data to see what it represented.  It just looked like "a person" to me.

nebulon42:
The "policeman" representation is quite common IMO (try a search for "police icon"). Still open for improvements except "star and shield" (too specific for Americas IMO) and "handcuffs" (not really feasible at that size).

kocio-pl:
For me it works.

PR #1222, opened 2015-01-13
Make living street casing darker to increase contrast

### LINKED ISSUE #1194: Living streets are hard to see
After #1153, I think living streets are harder to see when there are lots of houses around. Example: http://www.openstreetmap.org/#map=17/48.36770/10.90188

Possibly a lighter grey would now look better? It would need testing on other backgrounds.

math1985:
This resolves #1194

### DIFF
--- a/roads.mss
+++ b/roads.mss
@@ -31,7 +31,7 @@
 @tertiary-casing: #c6c68a;
 @residential-casing: #bbb;
 @service-casing: @residential-casing;
-@living-street-casing: @residential-casing;
+@living-street-casing: #999;
 @pedestrian-casing: @residential-casing;
 @path-casing: @default-casing;
 @footway-casing: @default-casing;


math1985:
We can look at a further change of colours of pedestrian/living_street later, but this is a first step.

nebulon42:
Could you please provide a preview?

math1985:
Before:
screenshot from 2015-01-25 18 31 29

After:
screenshot from 2015-01-25 18 32 09

kocio-pl:
It is a clear improvement for me. How would it look like on residential landuse?

math1985:
Before:
screenshot from 2015-01-25 19 32 51

After:
screenshot from 2015-01-25 19 28 50

kocio-pl:
Not good enough for me, but a bit better.

polarbearing:
The grey gives them a lot of prominence when there are many residential streets in an area, see here:
http://www.openstreetmap.org/#map=16/52.4606/13.7238

Maybe we could render them narrower, such as highway=service, while keeping the grey?

daganzdaanda:
> Maybe we could render them narrower, such as highway=service, while keeping the grey?

Yes, that would make a lot of sense.  Most of the living_streets that I know are really narrow, or at least made to look narrow so that drivers feel they can't go fast there.

mkoniecz:
> The grey gives them a lot of prominence when there are many residential streets in an area, see here:
http://www.openstreetmap.org/#map=16/52.4606/13.7238

Yes, this is a problem. And there are more affected places like http://www.openstreetmap.org/#map=15/50.0769/20.0377

> Maybe we could render them narrower, such as highway=service, while keeping the grey?

It seems like a good idea.

@math1985 Is it intentional to make living street stronger and more visible at z13?

HolgerJeromin:
are living streets really so different to deserve such a complete other color?

polarbearing:
highway={footway|pedestrian|living_street} are all ways where pedestrians have some priority, though footway is visualised completely different, the red colour even carrying some restrictive semantic. Some convergence between them would be useful.

mkoniecz:
@HolgerJeromin 

> are living streets really so different to deserve such a complete other color?

It inspired #1326 ("render highway=living_street exactly like highway=residential and highway=unclassified")

math1985:
We could consider merging this now as a temporary solution and in the meanwhile continue exploring a way to change the rendering of living streets more radically.

pnorman:
I think it's an improvement, but we may need to look at it further.

PR #1226, opened 2015-01-14
add rendering for amenity=dentist

nebulon42:
Was not rendered before. I also felt that `text-dy`might a bit too small for pharmacy, doctors and dentist, so I increased it to 12.
Icon: [nebulon42/osmic/dentist-16.svg](https://github.com/nebulon42/osmic/blob/master/health/dentist-16.svg)

Preview
dentist


althio:
I would like to see the same cross used in doctor / pharmacy.
I am not sure how it fits inside or next to the tooth though.

HolgerJeromin:
perhaps the cross could be behind like the calendar in [this icon](http://www.flaticon.com/free-icon/dentist-date-for-personal-care_30904).

nebulon42:
I agree that it might be good to have a cross in all health-related icons, but I don't think that it is needed here for the following reasons: The icon colour gives a hint and this icon is pretty distinctive. I can't think of much other uses for a tooth icon, but there are many for bags (doctors) and also for bottles (pharmacy).

Issue #1189, opened 2015-01-03
Icon outlines


nebulon42:
We might need outlines or halos around icons to make them distinguishable from the background. Especially since landuse colours are so saturated. See https://github.com/gravitystorm/openstreetmap-carto/issues/325#issuecomment-68612835 (second image) for why this might be needed.

Outlines cannot be part of the SVG (as Maki does for some icons) otherwise re-colouring with `marker-fill` won't work as intended. Unfortunately, `marker-line-` rules only work for markers that are specified with `marker-type`.

I have thought of stacking outline and fill with two separate SVGs. This has several drawbacks: First, we need two SVGs. The one comprising the outline has to be slightly bigger and should not have any inner holes. Second, we need to use `marker-allow-overlap: true`, which also allows overlap with any marker nearby. So this apparently does not work.

Did I miss something? IMO it would be easiest to be able to use `marker-line-` rules. I'm not sure if this is even technically possible. Should this be reported upstream?

Follow-up: If https://github.com/mapnik/mapnik-support/issues/27 and https://github.com/mapnik/mapnik/issues/2609 are related then at least Mapnik 3 could support it.

nebulon42:
I still think this is rather important, but there was no support for it. A pity. Closing this.

StyXman:
I have an update, I think, about this.

`mapnik` now supports marker stroke, stroke-width and stroke-opacity, which impacts the SVG stroke. The issue is that a stroke of N pixels means that half of those N pixels 'invade' the fill, while the rest become a 'halo'.

One thing that could be implemented is to add another layer with the same data and, for a halo of N pixels, set the stroke width to 2N. This of course means duplicating the layers, maybe it can be optimized with caching?

I also opened mapnik/mapnik#4246 , but I don't expect this to bubble up the stack (mapnik -> carto -> osm-carto) any time soon :(

StyXman:
I hit mapnik/mapnik#4247 :(

PR #1228, opened 2015-01-14
move library icon to SVG

nebulon42:
Implements part of #1165 by moving the icon for `amenity=library` to SVG.
Icon: [nebulon42/osmic/library-16.svg](https://github.com/nebulon42/osmic/blob/master/amenity/library-16.svg)

before
library_before

after
library_after


kocio-pl:
I don't like horizontal space - now it looks more like an arrow down or a gift than a book.

nebulon42:
I can't follow you regarding the gift I have to say.

kocio-pl:
Well, the lower part of the package with the gift. =} 

Anyway, I can't follow the horizontal space in the book icon - what does it represent?

nebulon42:
It is intended to represent the pages, the thickness of the book.

kocio-pl:
Hm, I was not sure, since I expect it to be on top. And I guess the cover is too short (vertical cover line should be longer than horizontal - this one is too square).

nebulon42:
> I was not sure, since I expect it to be on top

That seems to be a problem with this icon. The book is meant to show the inside but some are seeing the back cover (me too sometimes). I will try to adjust the fake perspective.

althio:
If you want to give a try to another idea and get rid of perspective how about several books on a shelf.
Found on the web as example:

nebulon42:
Yes, tried it already but didn't work so well.

nebulon42:
Slightly adjusted: library-162

kocio-pl:
I still consider it strange because of a square aspect ratio (if I understand correctly those small parts on the bottom are the thickness/z axis, so they don' count for height) - rectangles with one dimension longer are more common: [https://commons.wikimedia.org/wiki/File:Comparison_paper_sizes.svg]

nebulon42:
I have updated the icon and removed the z-axis part.

HolgerJeromin:
I think most names of the poi will clarify the icon for those people who don't recognize it.

mkoniecz:
Unfortunately at least for me the new icon is not recognizable and current one is.

nebulon42:
> Unfortunately at least for me the new icon is not recognizable and current one is.

I have experimented quite a bit with different options before the last update and I'm afraid I will not be able to come up with a different/better version.

vholten:
These are some variations:
book icons
book icons preview
(Inspired by [this icon](http://commons.wikimedia.org/wiki/File:Book-of-myst.png))

kocio-pl:
Both are much better for me!

nebulon42:
@vholten Are your previews output by TileMill or did you do some image editing?

vholten:
@nebulon42 The previous previews were done by image editing (downsampling the icons to 16 pixel width). 
This is how it looks in TileMill using SVG:
book icons tilemill

nebulon42:
Looks a bit blurry at the edges, but not bad. Maybe it helps to have the parts at the left/right side in solid colour as it is done for the lower part? I suggest that you create another Pull Request for it.

mkoniecz:
@vholten It looks really well, better than the current one.

althio:
@vholten I like them too.
I would prefer the left and right edges really vertical with no apparent perspective.

vholten:
OK, I will try some variations and create a new pull request.

math1985:
For what it's worth, tester recognizes the image proposed in this PR but prefers the current image.

nebulon42:
Thanks, but let's wait for @vholten's PR.

nebulon42:
I tried to recreate @vholten's proposal and it looks like this:
library_new

@math1985: Here I tried to use transparency (0.1) for the fill. What do you think?

As the outlines are a bit thin and the fill (pages) are quite big I think the icon is not prominent enough compared to the others, especially at lower zoom levels:
library_low_zoom

nebulon42:
Please give feedback on the new icon so we can move this forward, thanks.

math1985:
> Please give feedback on the new icon so we can move this forward, thanks.

I'm very busy at the moment, but will get back to it as soon as I have time.

kocio-pl:
I like this icon - it's a bit less visible then current one, but then I think libraries are now too visible. New icon is satisfying recognizable for me.

math1985:
I also like the new icon. It's more recognizable then the 'arrow', and as far as I'm concerned still prominent enough.

We could also consider swapping your proposed library and book shop icon. A (messy) stack of books reminds me more of a library than a book shop.

lest69:
I also like the new icon. I think it's visible enough, especially when combined with the label.

I'm not so sure about swapping the icon with book shop. In my experience, the open book is a fairly universal symbol for representing libraries. Regardless, the colour of the icon and label is different for the two, so it should be relatively easy to determine which is the amenity and which is the shop.

mkoniecz:
>  In my experience, the open book is a fairly universal symbol for representing libraries.

Situation in Poland is the same - IMHO this symbol or similar is more likely to represent library.

nebulon42:
I have to delay this a bit because this involves transparency and I have to address https://github.com/nebulon42/osmic/issues/3 first to see if the transparency is still needed. We then will have to decide if we want to have icon halos in here too (#1189).

mkoniecz:
@nebulon42 Is there anything blocking this PR? This icon is a nice improvement, is there anything that needs to be done before reviewing and hopefully merging it?

nebulon42:
@matkoniecz After careful consideration I cannot use the last proposed variant, because of the large inner holes that lead to difficulties with legibility on noisy background.
Example: noise_library

I therefore have reverted the library icon in my repository and will go on proposing my version as depicted in the preview in the first post. You can still find the other icon [here](https://github.com/nebulon42/osmic/blob/b20bf75298c6a268ceacc247bccea1d5b902030d/amenity/library-16.svg) if you choose to go ahead with that version.

matkoniecz:
@nebulon42 Yes, it is not working in cases like this example.

kocio-pl:
Do the pages have to be completely translucent or fully brown, or there could be a shade?

nebulon42:
Technically a transparent fill (say 10/15/20%) could work. But translucent fills clash with halos, so I cannot add that. (Although halos are not considered in this repository.)

lest69:
Rather than transparency, is it possible to use a lighter, but still solid, shade? Or can it only have on/off for each pixel? I'm not completely familiar with how this all works, but I envision something like having the pages coloured by the primary colour -50% saturation, for example.

nebulon42:
@lest69 When using `marker-fill` as we do it is not possible to use more than one colour.

Please tell me if you consider to merge this, otherwise I will close it.

math1985:
I'm fine with merging the current proposal, what do the others think?

kocio-pl:
For me it's just passable - visually not better than current icon (too strong), but not worse.

pnorman:
I'd rather keep the current one or use something similar to what was proposed in the PR comments. I don't like the one in the PR itself.

nebulon42:
I'm closing this, because we are re-iterating the same discussion over and over again.

PR #1229, opened 2015-01-15
fix dy for leisure=water_park

mkoniecz:
leisure water_park name eeeeee eeeeee eeeeee ref 1 ele 8000 closed_way master - dy

### DIFF
--- a/amenity-points.mss
+++ b/amenity-points.mss
@@ -860,7 +860,7 @@
       text-name: "[name]";
       text-size: 11;
       text-fill: darken(@park, 60%);
-      text-dy: 11;
+      text-dy: 13;
       text-face-name: @book-fonts;
       text-halo-radius: 1;
       text-halo-fill: rgba(255,255,255,0.6);


kocio-pl:
Hm, I can't recognize which one is "before" and which is "after" - seems like it was meant just for simple 1:1 comparison.

mkoniecz:
@kocio-pl - yes, it is comparison not on real data, but with a single generated element. It is not mentioned anywhere, mainly because I am not sure how to name it. Internally I am using term "synthetic data" but I think that it is not making situation clear.

> can't recognize which one is "before" and which is "after"

I expected this problem. Have you noticed "before", "after" labels? Or is there something else causing this problem?

kocio-pl:
It was just the labels.

mkoniecz:
Diff in the new style. Is it better?
leisure water_park name eeeeee eeeeee eeeeee ref 1 ele 8000 closed_way master - dy

kocio-pl:
I see no difference, just the bigger fonts. However the problem isn't serious, it's just a minor "hm?..." and if it couldn't be easily fixed, so be it.

HolgerJeromin:
> Is it clear why first image is displayed only once?

Because it is not rendered in this zoom levels? I think printing this two times would be a bit clearer. Especially the label "after" would make more sense then. The labels are exact over an image where it is wrong. Perhaps centering the label "before" and "after" on the columns would be helpful, too.
I would perhaps name them "unit tests". Users with a programmers background will have the right idea.

PR #1234, opened 2015-01-16
Fix contradictory unzip options

### LINKED ISSUE #1233: Incompatible unzip options
with `./get-shapefiles.sh`

```
simplified-land-polygons-complete-3857...
unzip: -n, -o and -u are contradictory
```

Maybe `-o` is unnecessary with a list of file specified?

BSD zip appears to have stricter option parsing than Info-ZIP, but it's a bug in the script under either.

pnorman:
-u and -n are mutually exclusive options. Info-ZIP wasn't reporting
this as an error, but BSD zip caught it.

Fixes #1233

### DIFF
--- a/get-shapefiles.sh
+++ b/get-shapefiles.sh
@@ -1,7 +1,7 @@
 #!/bin/sh
 set -e -u
 
-UNZIP_OPTS=-qqun
+UNZIP_OPTS=-qqu
 
 # create and populate data dir
 


math1985:
I get a prompt now:

> replace data/simplified-land-polygons-complete-3857/simplified_land_polygons.cpg? [y]es, [n]o, [A]ll, [N]one, [r]ename: 

Is that desirable? I can imagine it might break users who rely on this script being non-interactive.

This is `UnZip 6.00 of 20 April 2009, by Debian.`

pnorman:
> I get a prompt now:
> 
> replace data/simplified-land-polygons-complete-3857/simplified_land_polygons.cpg? [y]es, [n]o, [A]ll, [N]one, [r]ename:
> Is that desirable? I can imagine it might break users who rely on this script being non-interactive.
> 
> This is UnZip 6.00 of 20 April 2009, by Debian.

ugh. I got this intermittently in testing before, I'll have to rework this to not rely on those options but instead remove the files and replace them

PR #1235, opened 2015-01-16
move embassy icon to SVG

nebulon42:
Implements #1165 by changing the icon for `amenity=embassy` to SVG.
Icon: [nebulon42/osmic/embassy-16.svg](https://github.com/nebulon42/osmic/blob/master/amenity/embassy-16.svg)

before
embassy_before

after
embassy_after


mkoniecz:
Please, test it with tall name like "ÉÉÉÉÉÉ ÉÉÉÉÉÉ ÉÉÉÉÉÉ" (text-dy is too low).

nebulon42:
The text would have worked, but I nevertheless increased `text-dy` by 1.

mkoniecz:
Weird, according to my test it was failing.

nebulon42:
With your tool? I changed the text to ÈÈÈÈÈ, but I don't know, which value do you suggest?

z19
embassy_text

pnorman:
I'd give it one more pixel over what you have in the image - because the text isn't uniformly tall, it looks to not be hitting the tallest point.

nebulon42:
Right, I did that already, but didn't update the preview.

nebulon42:
I have updated the icon to a lighter version that looks better and is more similar to the old version.

math1985:
The latest preview does not match with the icon in the repository. Which one is the correct version?

nebulon42:
> Which one is the correct version?

The one in the inital post should match with the repository and the file in this PR.

math1985:
The preview in https://github.com/gravitystorm/openstreetmap-carto/pull/1235#issuecomment-70380791 is different from the file in the repository though. Can I ignore this preview?

nebulon42:
Yes, this was the first version, which has been superseded. The comment was intended to show to @mkoniecz that I tested with this particular name.

I probably should not update the first comment anymore to avoid confusion about updates.

math1985:
> The comment was intended to show to @mkoniecz that I tested with this particular name.

I actually like the updated first post. In any case I merged this now.

mkoniecz:
> I actually like the updated first post.

I also prefer first post to contain current state (or at least - link to post with a current state), not what was initially submitted.

PR #1238, opened 2015-01-17
Improve rendering of construction=residential on zoom 13

sb12:
Before:

constructionbefore

After:

constructionafter

### DIFF
--- a/roads.mss
+++ b/roads.mss
@@ -743,6 +743,11 @@
           [zoom < 13] {
             line-width: 0;
             b/line-width: 0;
+          }            
+          [zoom >= 13][zoom < 14] {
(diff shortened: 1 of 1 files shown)


math1985:
I agree highway=construction, construction=residential should be more narrow on z13.

However, this proposal also makes highway=construction, construction=residential more narrow on z14 and higher, and on these zoom levels, in particular from z15 on, it is rendered too narrow now.

sb12:
You are right. Thank you for spotting this. I probably should have tested it better.

I fixed it now and also improved the code a little bit.

math1985:
Thanks, that's better.

PR #1240, opened 2015-01-18
move picnic table icon to SVG

nebulon42:
Implements part of #1165.
Icon: [nebulon42/osmic/table-16.svg](https://github.com/nebulon42/osmic/blob/master/outdoor/table-16.svg)

before
table_before

after
picnic


mkoniecz:
My problem with this icon that it looks like information board (at least for me). New ones also have this problem.

nebulon42:
You mean like a scaffold for a board (as in painting)? This is a very common symbol for a picnic table, I'm not able to think about something else. But if you have some hints, maybe there is something better.

mkoniecz:
I know. I tried some variations - but it made problem even worse.

althio:
Based on this picture:
https://commons.m.wikimedia.org/wiki/File:Picnic_table.jpg

I think the bench could be represented with extra height and the tabletop with reduced height.
Rough idea:image2

With extra features like glass and bottle?
Rough idea 2:image4

pnorman:
I prefer the simpler bench to those more complex ones - either the existing one or the new one.

math1985:
Tester thinks icon symbolizes a bench in the park. Good enough I think.

nebulon42:
By saying "simple" do you refer to the thicker or the thinner version?

math1985:
The thinner one without bottles.

althio:
Thanks for your feedback. That was mostly illustrative hints to @nebulon42 because sometimes explaining by words is more difficult. Eventually if he thinks he can integrate a few concepts in his own design he will certainly do better than my attempts.

As for preference, I am OK with either. However I think the after-thick from @nebulon42 has a better shape but is a bit too thick, while after-lighter has the correct weight but shape of legs and benches feel less natural.

Another attempt based on the lighter version: picnic-lighter-width-bench3-narrower

nebulon42:
@althio Thanks for the concrete suggestions, much appreciated.

I'm still undecided on thick vs. thin. I think that thin is too thin, but thick is a bit too thick. A solution would be to use 1,5px wide elements, but this would increase blurriness.

pnorman:
> A solution would be to use 1,5px wide elements, but this would increase blurriness.

It's not too serious an issue if the unaligned side is on the interior.

nebulon42:
Thanks @pnorman for the suggestion, you were right. I'm now quite happy with the result, see preview above.

althio:
@nebulon42 I am happy also with your new suggestion, 1.5 is a nice middle ground.
Last nitpick -- feel free to ignore -- horizontally maybe the benches should extend more or the table extend less.

Issue #1066, opened 2014-10-15
Render memorial:type=stolperstein only in z19


tilmanb:
There are lots[0] of "Stolpersteine"[1] in the OSM data. Each Stolperstein is a memorial for one or more persons deported and killed by the nazi regime.

The number of Stolpersteine is huge, and they someplace[2] clutter the map output at z18.

This has lead to users deleting them "for simplicity's sake" and to "clean up the map". This is unacceptable, of course, but the problem of map clutter remains.

So i propose to render features tagged with

```
historic=memorial
memorial:type=stolperstein
```

only at z19.

[0] 12567 nodes as of this writing, see http://taginfo.openstreetmap.org/tags/memorial%3Atype=stolperstein
[1] https://en.wikipedia.org/wiki/Stolperstein
[2] http://www.openstreetmap.org/#map=18/50.12166/8.68692

pnorman:
👎 to memorial:type=stolperstein. The usage of it is too country-specific. I'd be more open to memorial=plaque along with the other `memorial` subkeys, but usage isn't great.

dieterdreist:
sent from a phone

> On 15 Dec 2016, at 06:35, Paul Norman <notifications@github.com> wrote:
> 
> 👎 to memorial:type=stolperstein. The usage of it is too country-specific. I'd be more open to memorial=plaque along with the other memorial subkeys, but usage isn't great.
> 


although this had started in Germany as a project by a German artist, there are now occurrences in other countries as well. As long as there are many, why does it matter those are not equally distributed?

imagico:
We aim to support generic tagging.  `memorial:type=stolperstein` on the other hand is highly specialized.  If `memorial=plaque` (3617 uses) and `memorial:type=plaque`(476 uses) are rendered this could be considered but otherwise it would support incorrect tagging.

Note the fact that we have competing `memorial=*` and `memorial:type=*` is in itself an issue.

math1985:
>  The usage of it is too country-specific.

There are stolpersteinen in 18 different countries now.

However, I agree usage is too specialized. Not too mention that there are Stolpersteinen every couple of houses in some German cities.

dieterdreist:
2016-12-15 9:27 GMT+01:00 Christoph Hormann <notifications@github.com>:

> We aim to support generic tagging. memorial:type=stolperstein on the
> other hand is highly specialized. If memorial=plaque (3617 uses) and
[...]

Issue #758, opened 2014-07-25
Add rendering for historic=monument


mrwojo:
An icon and name should do (as with its cousin historic=memorial). There's already a CC0 icon for it at http://www.sjjb.co.uk/mapicons/ which is the icon osm.org's browse page uses for this tag.

historic=monument has about 30k uses.

This may reveal some latent weirdness in tagging... like how every stone in Stonehenge is marked historic=monument.

nebulon42:
I can offer this ([nebulon42/osmic/monument-16.svg](https://github.com/nebulon42/osmic/blob/master/tourism/monument-16.svg)) as icon if it is decided to be rendered.

kocio-pl:
I like this icon - of course I wait for a subtags support (wall-mounted memorial is very different from a big memorial monument - see the examples of its variety: https://en.wikipedia.org/wiki/Memorial#Examples_of_notable_memorials ), but we need it for many landmarks.

kocio-pl:
@nebulon42: Whatever happened to this icon? Long time no hear... =}

nebulon42:
I never intended to do a PR here, because first it has to be clear if it should be rendered or not and I'd like to avoid another #1304 where in the PR discussion it became clear that there is no support for additional rendering. The icon is available, feel free to use it.

kocio-pl:
Hm, because historic=monument looks like designed especially for buildings, which are visible on the map anyway (even with the name), this icon is optional, but I think it would do no harm. Probably we need a visual before/after test to know.

PR #1241, opened 2015-01-18
move memorial icon to SVG

nebulon42:
Implements part of #1165.
Icon: [nebulon42/osmic/memorial-16.svg](https://github.com/nebulon42/osmic/blob/master/tourism/memorial-16.svg)

before
memorial_before

after
memorial_after


math1985:
Tester recognizes neither before nor after.

I'm not sure if this should rendered significantly/on low zoomlevels at all. Some mrmorials are significant landmarks, but others are very minor compared to yhe surroundings.

pnorman:
I think the new icon is an improvement - but the basic idea of the icon needs revisiting, as in my experience it's not clear what it means.

HolgerJeromin:
yes,  the icon is good, but especially the german "Stolpersteine" are much too prominent.

nebulon42:
Prominence: Agreed, but the icon is now smaller although sharper. This needs adressing by categorising using other tags.

Other icon: I thought about this a bit and it is hard to come up with something distinctive for memorial. The idea of depiciting a plate or stone with text on it is not so bad. Sometimes this are statues, but depiciting e.g. a head would not include all the other possibilities of memorial. The other icon I meant to use for monument ([monument-16.svg](https://github.com/nebulon42/osmic/blob/master/tourism/monument-16.svg)) might imply a larger structure.

althio:
Regarding the icon itself: :+1: 

Regarding previous comments: I very much agree with @math1985 and @nebulon42 
> I'm not sure if this should rendered significantly/on low zoomlevels at all. Some mrmorials are significant landmarks, but others are very minor compared to yhe surroundings.
> Prominence: [...] This needs adressing by categorising using other tags.

As many other features, it would be great to have subtags and established rules (tiers for zoom threshold and rendering) so that:
* first tier: important/landmark/monumental features are rendered early (z16?) and maybe special rendering when second tier appears (slightly larger icon size? icon in square/circle?)
* second tier: common features are rendered at closer zoom levels (z18?)
* negligible features are not rendered or only a maximal zoom level and maybe special rendering (lighter as in private POI?).

polarbearing:
@althio - would be good, but apparently (see #1066) subtags are not yet supported. 
If the monumental features are tagged as monument they could get a different icon as proposed in #758.

kocio-pl:
Sooner or later we will need the subtags rendering - memorial table and memorial cemetery are very different beasts.

math1985:
Memorial rendering clearly has issues, but at least everyone agrees that the new icon is an improvement, so I don't mind merging this.

Issue #937, opened 2014-09-10
PolygonPatternSymbolizer metatile alignment


pnorman:
To use polygon pattern, the pattern either needs to fit into fit into a metatile an integer number of times with `polygon-pattern-alignment: global` or have `polygon-clip: false`, or artifacts at meta-tile boundaries will occur.

Without a fixed metatile size, this implies both pattern dimensions should be a multiple of two, and at most 256.

Because the defaults used to be `polygon-clip: false`, many of the icons do not meet this criteria. We need to check if there is more performance from `polygon-clip: false`, and if so, set that, otherwise, change the PNGs.

Some shell scripting (`./node_modules/.bin/carto -a 3.0.12 project.mml | grep PolygonPatternSymbolizer | sed 's| <PolygonPatternSymbolizer [^>]*file="\([^"]*\)".*/>|\1|' | sort -u | xargs pnginfo | grep -B1 'Image Width'`) lets us get the image sizes.

The following files are not powers of two

```
symbols/danger.png...
  Image Width: 30 Image Length: 30
symbols/grave_yard_jewish.png...
  Image Width: 18 Image Length: 18
```

pnorman:
Updated the list.

pnorman:
Updated, including a minor shell script change.

pnorman:
Updated

pnorman:
Updated, we're close

kocio-pl:
We have the last one which is still not power of two:

```
symbols/danger.png...
  Image Width: 30 Image Length: 30
--
```

but we have more above 256 px:

```
symbols/wetland_bog.png...
  Image Width: 512 Image Length: 512
--
symbols/wetland_mangrove.png...
  Image Width: 512 Image Length: 512
--
symbols/wetland_marsh.png...
  Image Width: 512 Image Length: 512
--
symbols/wetland_reed.png...
  Image Width: 512 Image Length: 512
--
symbols/wetland_swamp.png...
  Image Width: 512 Image Length: 512
```

Why would they be a problem?

pnorman:
> Why would they be a problem?

Because they won't align when rendering tile by tile

kocio-pl:
Should we recreate them all in 256 px then?

pnorman:
> Because they won't align when rendering tile by tile

They also won't align with some metatiling systems when taking into account buffer.

> Should we recreate them all in 256 px then?

Yes.

sommerluk:
[...]

Issue #822, opened 2014-08-03
Distinguish deciduous and coniferous forests


math1985:
At the moment, we don't distinguish deciduous and coniferous forests. Distinguishing them in rendering might improve the map.

See https://trac.openstreetmap.org/ticket/1526.

matkoniecz:
see https://github.com/gravitystorm/openstreetmap-carto/pull/1242#issuecomment-126958042 for ideas and partial implementation how this data may be displayed

BushmanK:
I made several sample patterns using current pattern used for natural=wood.

leaf_type=needleleaved no leaf_cycle
![needleleaved](https://habrastorage.org/files/bc6/fba/7c5/bc6fba7c5bcf4f1eb684c4cafc103ddf.png)

leaf_type=broadleaved no leaf_cycle
![broadleaved](https://habrastorage.org/files/4a9/241/392/4a9241392ed14f18bb947cb831f8b57c.png)

leaf_type=broadleaved leaf_cycle=deciduous
![broadleaved deciduous](https://habrastorage.org/files/6e4/2aa/918/6e42aa918bed495fb2eac710cd720a0b.png)

For leaf_cycle=evergreen, leaf_cycle=semi_evergreen,  leaf_cycle=semi_deciduous I suggest using pattern with fully filled tree outlines, but any other ideas are welcome, for sure.

So, basically, for unknown life cycle outline should be empty; for trees, looking differently through the year - half-filled, for trees, not changing or barely changing its appearance - fully filled.

While not all needle-leaved plants have similar conic shape as fir or cypress (mature pine, for example), in majority of cases, this shape is associated with them pretty well. So, I believe, using conic and rounded canopy outline is legitimate way to show broadleaved and needleleaved trees.

Actually, I have no  idea how to show leafless plants - hard to imagine unified symbol for them, they are too different.

d1g:
> It doesn't sound like the tagging is fully settled out. Worth noting: we have wood in the database, but not leaftype.

Things changed since 2014, probably it's time to review (now 420K leaf_cycle objects vs 366K instances of the deprecated wood)

http://taginfo.openstreetmap.org/compare/wood/leaf_cycle

HolgerJeromin:
We are waiting here for #1504

kocio-pl:
[...]

Issue #647, opened 2014-06-18
natural=wood and landuse=forest should render using the same style


matkoniecz:
Is it intentional that there is a significant difference between natural=wood and landuse=forest?

See http://www.openstreetmap.org/?mlat=50.07725&mlon=19.88101#map=19/50.07725/19.88101

I would suggest to render natural=wood using exactly the same rules as landuse=forest

![przechwytywanie](https://cloud.githubusercontent.com/assets/899988/3320585/3956d46a-f72c-11e3-886c-e49f1ec99ccd.PNG)

Top left - natural, bottom right - landuse.

PR #1210, opened 2015-01-08
Add transparency to halo placename labels

math1985:
This looks prettier, and without compromising legibility as far as I see.

PR #1242, opened 2015-01-19
Use a random symbology for forests

### LINKED ISSUE #938: Randomized symbology for forests
See towards the end of http://sk53-osm.blogspot.com/2014/09/woodland-cartography.html by @sk53 as a reference. We may want to consider this for other features, but forests are a good start.

There are a few ways to do this

## Use a large texture
See https://github.com/gravitystorm/openstreetmap-carto/issues/937#issuecomment-55294102 and @imagico's example of 
Randomly placed trees

There is no particular reason why the texture must be 256x256. The constraint is that any sub-square in the top-left of a potential metatile size forms a valid repeating pattern (i.e. the edges align). In production we use a 2048x2048 metatile, which is certainly big enough to appear random, as not all of it will appear on most screens at once.

Advantages
* Minimal style changes
* No SQL changes

Disadvantages
* Detectable repetitions for small meta-tile sizes.
* Image needs to be designed carefully so any sub-square in a MT size meets tiling criteria

## Generate point locations in SQL
We could plot tree icons on top of a solid green fill by generating points in SQL.

It is possible to generate a grid in SQL, and then offset each point based on some criteria that isn't visually obvious (e.g. md5 of location)

The SQL to generate a 16 pixel grid is approximately

```sql
SELECT 
    ST_SetSRID(
      ST_Point(x+floor(st_xmin(!bbox!)/(!pixel_width!*16))*!pixel_width!*16,
      ST_Point(y+floor(st_ymin(!bbox!)/(!pixel_height!*16))*!pixel_height!*16),900913) 
  FROM 
    generate_series(0,ceil((st_xmax(!bbox!)-st_xmin(!bbox!))/!pixel_width!,16) AS hor(x), 
    generate_series(0,ceil((st_ymax(!bbox!)-st_ymin(!bbox!))/!pixel_height!,16)  AS ver(y);
```

Advantages
* Never repeats
* With ``ST_Buffer(way, -sqrt(!pixel_width!*!pixel_height!)*N) AS way`` you can avoid placing an icon where it gets cut off at an edge

Disadvantages
* Adds another layer
* More complicated SQL

Mixed
* Different styling. The trees will essentially be overlays

pnorman:
This adds a random symbology for forests, lightens forests, and unifies natural=wood and landuse=forest rendering.

The colours for `landuse=forest` and `natural=wood` have been unified and slightly lightened.

[z10](http://www.openstreetmap.org/#map=10/49.2902/-122.3245)
image

At higher zooms, a randomized symbology has been added, designed to not imply a particular type of forest.

[z15](http://www.openstreetmap.org/#map=15/49.2138/-122.9021)
image

[z16](http://www.openstreetmap.org/#map=16/49.2363/-122.8884)
image

As the new symbols are an overlay, they show even if there is another landcover within the forest.
[z15](http://www.openstreetmap.org/#map=15/49.2807/-122.9366)
image

They are a repeating 512x512 pattern, which is hopefully large enough that the repeating nature of the pattern can't be seen.

[z16, original not shown](http://www.openstreetmap.org/#map=16/49.2570/-122.8242)
image

With woods and forests no longer being primarily one colour, the name text and halo was reworked

[z17](http://www.openstreetmap.org/#map=17/49.25220/-122.81239)
image

Unfortunately I was unable to find a tree symbol that does not imply something about the type of tree, but this does open up the door to using different symbols for different types of forests (#822)

`landuse=forest` and `natural=wood` now render the same. For some mappers there was a difference between what they meant - unfortunately the difference varied among mappers, making the two different tags effectively the same globally (https://github.com/gravitystorm/openstreetmap-carto/issues/647#issuecomment-52816195) and the distinction between them useless. This was also not captured in the existing cartography.

Fixes #938
Fixes forest part of #937
Closes pnorman#6

Thanks to @SK53 for symbol selection and @imagico for the point generation tool.


pnorman:
Some more examples

[streams](http://www.openstreetmap.org/#map=15/49.3430/-122.7430)
image

[lower-zoom footways](http://www.openstreetmap.org/#map=14/49.1745/-122.8229)
image

[narrow woods and military/wood combo](http://www.openstreetmap.org/#map=14/49.1595/-123.1026)
image

HolgerJeromin:
IMO the "icons" are to turbulent if there are large areas shown.

pnorman:
> IMO the "icons" are to turbulent if there are large areas shown. Probably because it is sharper than the old version.

How would you suggest improving it?

kocio-pl:
New icons are too visible for me - comparing to the old ones it is bigger and consists of pair of symbols. I would make it the same size as before and use only one symbol instead of a pair (512x512 box gives us a comfort of using both symbols semi-randomly, we don't have to squeeze them all in every point).

> As the new symbols are an overlay, they show even if there is another landcover within the forest.

Is there a way to avoid it or we're bound to it by the nature of the overlays? 

I think that for example glades/clearings should be clean. It is easy to recognize them as part of the forest just because they are completely surrounded by it.

Also some residential areas can be placed inside the forest/woods, but if we need to show the trees there, we should just extend the forest area to cover it.

EdLoach:
I might have misunderstood the comment above, but glades/clearings where there aren't trees will be holes in the existing multipolygon (or need fixing). Similarly for residential areas inside such areas.

kocio-pl:
Maybe I got it wrong - the icons will be visible only if the other landuse is NOT an inner relation element (forest being the outer one)? I would support it then, since sometimes included landuse is so full of trees, that you can't really say the forest stops on the edge of it, yet it is useful to see this landuse being there.

imagico:
First of all the color seems an improvement from the examples shown and unifying it with forest seems a good idea as well.

Now the critical remarks:

* Color needs to be tested against other vegetation landcovers, in particular scrub.  Also how does it look together with trunk roads?
* Contrast between wood and water still seems suboptimal - maybe considering the radical idea of darkening the water color would be an option.
* The pattern appers somewhat strong in many cases which could be both due to the symbols, their relatively large size and the choice of color.
* I do not really like the paired rendering of tree types since it introduced a regular arrangement into an otherwise irregular pattern. It would be better to randomly vary the symbols i think.  jsdotpattern already supports this.  You can specify an array of strings instead of a single SVG string in [https://github.com/imagico/jsdotpattern/blob/master/symbols.js](symbols.js) to use this.  I included your symbols and a variant with a triangle tree symbol (see next point) in the deployed version.  Note the automatic offset calculation is still broken as you noticed.
* A triangle shaped rendering of the conifer tree might be a better fit than the branched version used right now.
* How about using a regular pattern as before instead of the irregular one to indicate landuse=forest in contrast to natural=wood.

Generally with regards to landcover overlays (in particular also for wetland) i thought about introducing a lower way_area limit in the range of 10-100 pixel to avoid irritating isolated partial symbols.  This of course will lead to problems with fine grained splitting of mapping but i am not sure how much problem this would be in reality.

Here my quick try with random symbol selection and triangle shaped conifer symbols:

tree pattern

I would also consider making the symbols smaller even if this makes them blurry - at 80 percent size they are still recognizable and the more blurry lines could make them less disturbing.

HolgerJeromin:
The name of the forest is more important than the icons. So i would make the icons much lighter.

SK53:
Broadly I think this is a good discussion with plenty of salient points. Many thanks to Paul for the work he has put in so far.

I really have only one point to add which is about the choice of a dual tree symbol rather than a single tree. The idea was taken from the Thunderforest Landscape layer, but the principle behind this choice is important. 

A choice of any single tree symbol as a generic symbol will always be inappropriate somewhere in the world, the current (conifer) symbol is really only appropriate in northern areas (boreo-temparate zones). A broad trawl through a variety of sources failed to find single tree symbols which are not instantly recognisable as either conifer or broad-leaved, so either one has a dual symbol or a mixture of two symbols (as suggested by @imagico).

At least for me, one of the ideas of this woodland cartography is to encourage better tagging of forests and woods, either with the currently supported wood tag or the leaf_type tag. I would therefore like the symbolisation to discriminate between undifferentiated woodland and the main types: coniferous, broad-leaved, and mixed. If the symbology of the undifferentiated woodland is less attractive than those tagged with a type, this may result in more attention to this matter by mappers. In this case the paired symbology would be distinct from the random symbology which I would reserve for wood/leaf_type=mixed.

I know advocating use of a perhaps sub-optimal symbology as a nudge for better mapping might go against the grain, but there it is. 

A minor point, I do thing @imagico's triangular conifers work better, not necessarily in their own right, but the in the current version the broad-leaf symbol looks much more stylised than the conifer symbol. Changing to the triangular form reduces that discrepancy.

nebulon42:
I like this very much and think that this is definitely an improvement. I'm also in favour of lightening forest. Some comments/suggestions:

* I'm not sure if we should adjust the colour more, I will have get used to it first. (similarity with trunk road remains)
* I like that the symbols are showing up on landuses that are not in the multipolygon relation, this will encourage cleaning up these things.
* The symbols are rather prominent, lowering the opacity to 0.5 or even to 0.4 would help. Also using alternating symbols as proposed could be an improvement. They should be more unobtrusive.
* The symbols are showing up too early and thus adding a lot of noise to the map. I would propose to render them at z13 at the earliest or maybe at z14 depending on how noticable they are. Maybe fading them in at increasing zoom level by using opacity would be also an option. (This may be no longer an issue when symbols are less prominent.)
* Forest names seem too yellowish, because of the halo. Maybe using white with transparency like `rgba(255, 255, 255, 0.7)` for it (as in #1210)?

forest_name

althio:
I also find it an improvement
Other people already voiced concerns with good suggestions, so I will just 'add +1' to:
* name text and halo: not visible enough, lacks contrast
* contrast forest/water is worse now
* paired symbols are heavy, alternate symbols are better
* regular pattern for forest vs random for wood

imagico:
> If the symbology of the undifferentiated woodland is less attractive than those tagged with a type, this may result in more attention to this matter by mappers. In this case the paired symbology would be distinct from the random symbology which I would reserve for wood/leaf_type=mixed.

That is a good point.  Still as long as the type is not rendered it is really not necessary to reserve the good looking style for mixed type woodland.  When support for differentiating types in rendering is added the unspecified type could be easily downgraded to a less attractive pattern.

One option to improve the paired rendering would be to randomly swap the pairs.

Triangle vs. branches - i have to admit i am probably somewhat biased in this regard since i am very used to the German tradition of even more abtract open triangle/circle shape rendering.  I do not consider this a very important aspect, keeping the connection to the traditional rendering in the OSM-style has advantages too.

pnorman:
> > As the new symbols are an overlay, they show even if there is another landcover within the forest.
> > Is there a way to avoid it or we're bound to it by the nature of the overlays?
> 
> I think that for example glades/clearings should be clean. It is easy to recognize them as part of the forest just because they are completely surrounded by it.

Glades and clearings would not normally be within a forest polygon.

kocio-pl:
> Glades and clearings would not normally be within a forest polygon.

What about multipolygons?

pnorman:
Colours
====

The colour has not significantly changed from the old wood colour, so basically any issues with 

> Color needs to be tested against other vegetation landcovers, in particular scrub. 

Yes - new forest to scrub distance is insufficient at 5.9. It wasn't great before with old wood/scrub at 7.1. for reference, residential/land is 6.4 and commercial/retail is 7.7.

image

Of course, they both have symbols. I'll take another look at all three colours, as I'd like to have a consistent hue for all three.

> Also how does it look together with trunk roads?

About as bad as before. I want to adjust road colours, but not in this PR.

> Contrast between wood and water still seems suboptimal - maybe considering the radical idea of darkening the water color would be an option

new forest/water contrast is 27.2, which is actually an increase from old wood/water and scrub/water. Darkening the water (or forests) wouldn't increase contrast much - the difference between them comes from the chroma and hue angle.

Polygons
====
> What about multipolygons?

Multipolygons do not normally exist in the rendering database but in any case, it remains the same - a clearing within a forest will have had a hole cut out, and that hole is not within the polygon.

Regular patterns
====

> How about using a regular pattern as before instead of the irregular one to indicate landuse=forest in contrast to natural=wood.

Removing the difference between `landuse=forest` and `natural=wood` is intentional, as there is no consistent difference between the tags.

Symbols
====
The choice of a dual symbol is quite intentional - I need a symbol that does not imply one type of tree or mixed trees, but an unknown type. Unfortunately, I can find no single icon that does that. Suggestions are of course welcome, but both myself and @SK53 looked for a generic tree symbol and couldn't find one.

The triangle symbol is an improvemnt, as it has the same amount of stylization, and also will probably work better at smaller sizes.

I will be re-generating with a new symbol, making use of the undocumented symbol [list](https://github.com/gravitystorm/openstreetmap-carto/pull/1242#issuecomment-70475572) feature to give different left/right pairs.

Text
====

> Forest names seem too yellowish, because of the halo. Maybe using white with transparency like `rgba(255, 255, 255, 0.7)` for it (as in #1210)?

That is basically what it is doing - the colour chosen is a mix of the forest background and white. The name colour could use some adjustment. I'm not particularly sure it's any worse than the old labels or the old park label.

More in the next couple of days when I can work on it again

polarbearing:
> The choice of a dual symbol is quite intentional

While I would prefer the randomised mix of symbols, if you want to go with the dual symbol the trees could be slightly shifted vertically against each other, so they do not stand too neatly on the same line.

kocio-pl:
I thought that if we're not sure what type of trees are in the forest/woodland, we may make a pair of different trees, but in a V-shaped icon (common roots would suggest ambiguity more than two paralell symbols). 

But then I thought this may be too complicated (especially because we have limited pixel matrix) and we could make one mixed symbol like that: a triangle based symbol with peak on top, but also with a rounded bottom.

imagico:
Thanks for addressing the issues brought up.

## Color

Your deltaE values are in line with what i would have assumed - other close colors are probably orchard and vineyard.  But in practial application differences in brightness are much more important than in tone since output devices, in particular the more crappy ones (laptops, office monitors) are much worse in differentiating color than in differentiating brightness.

My suggestion would probably be to lighten the pattern color a bit but to keep the base color as it is for the moment - the number of very similar green tones currently used for area coloring is not really discernable though - this will need to be either extended to brighter colors for the various grass tones or by unifying different types and differentiating only by pattern.

## Symbols

I am not really convinced that a mixed pattern is not an option as long as the woodland type is not taken into account at all - see my reply to @SK53.  The paired pattern works very well as an indicator for missing tags encouraging the mapper to add them but if more precise tagging does not lead to a change in pattern this effect is wasted and counterproductive.

The paired trees do look much better with random swapping IMO - if they are put fairly close together so they really function as a single symbol:

randomized tree pair pattern

SK53:
@polarbearing I had already mention to pnorman that offsetting the two icons might enable getting them even closer, but havent yet tried it.

@kocio-pl the common root is a great idea, but probably doesnt work at this scale.

One other thing I noticed looking at British Ordnance Survey maps is that tree symbols have a 'foot', presumably representing a shadow (from the NW):

os7series_wood_symbology

Again this might be too busy for the scale, but again I think its probably something to note for future use.

@imagico I experience a slight optical illusion with the icons appearing to slant away from the triangle (conifer) side so it might need a bit of tweaking.

imagico:
> I experience a slight optical illusion with the icons appearing to slant away from the triangle (conifer) side so it might need a bit of tweaking.

This is to be expected with the triangle shape close to something else on one side.  But i suppose such effects are much weaker in actual use of the pattern (lower contrast and  combined with other elements).  But there certainly is room for variation.

kocio-pl:
@imagico: Such smaller and more abstract symbols glued together are acceptable for me for a general forest. What I'm also interested in is how will we render each type of forest (especially mixed) in this case - it should be visually more "attractive", to encourage mappers to tag more details.

mboeringa:
@kocio-pl 
> New icons are too visible for me - comparing to the old ones it is bigger and consists of pair of symbols.

@HolgerJeromin 
> The name of the forest is more important than the icons. So i would make the icons much lighter.

@nebulon42 
> The symbols are rather prominent, lowering the opacity to 0.5 or even to 0.4 would help. Also using alternating symbols as proposed could be an improvement. They should be more unobtrusive.

Agree with all the above remarks. It needs more subtlety, most likely starting with a lighter color of the symbol, and possibly smaller icons, and maybe less dense. The *first* - not paired - example of randomized symbology of @imagico also seems to hold some promise.

pnorman:
The name colour has been changed back to darkening by 30. The calculation is carried out in a perceptual colourspace, avoiding the hue and chroma shift inherent with `darken()` which changes more than lightness.

image

CIE76 ΔE values
image

I am fairly happy with the new symbols

image

image

Grass has been shifted away from yellow
image

scrub is distinguishable
image
image

althio:
forest-label
I think that the labels should be compared to the fill color but also to the pattern color.
Patterns are quite disturbing when you read something on them.
I don't know if you have any guidelines on color for name labels(?).
For forest name on forest landuse: I would suggest to radically improve readability with good contrast like the island name label on forest landuse.

kocio-pl:
As for me, the name name of the forest is hardly visible and the icons are a bit too dark (but at least now they have proper size and shape).

imagico:
I think this is an improvement, the changed base color is now a bit closer to the water color again though.

I would still lighten the pattern symbols to be somewhat less prominent, maybe in turn use a somewhat denser pattern.  This would improve readability of the labels as well.

Not too sure about the grass color change - unifying grass and grassland is a good idea - on high zooms grassland could use a pattern for distinction though (but that's for a separate change).  Making it less yellow makes it less distinguishable from other greens however.  The yellow tone might seem less intuitive in humid regions but it is actually quite fitting in other parts of the world where grassland usally dries up during the dry season.

pnorman:
> I think this is an improvement, the changed base color is now a bit closer to the water color again though.

Actually it's farther from the water colour than the old wood colour, which is basically unchanged. It has a ΔE of 27.

> Not too sure about the grass color change - unifying grass and grassland is a good idea - on high zooms grassland could use a pattern for distinction though (but that's for a separate change). Making it less yellow makes it less distinguishable from other greens however.

`@grassland` was Lch(87.4,27.35,132.4), and all the new greens are 135. `@grass` was the yellow one Lch(89.9,36.5,125). `@grass` is also used for meadow, common, garden, village_green and conservation - it's really a bit of a mess, but I don't consider it in scope for this.

landcover.mss
```diff
@@ -546,13 +527,18 @@
 }
 
 #landuse-overlay {
-  [landuse = 'military'][zoom >= 10]::landuse {
+  [feature = 'military'][zoom >= 10] {
```
math1985: I would prefer to have this as `feature = 'landuse_military'` to keep consistency with the other feature constructs.

landcover.mss
```diff
@@ -546,13 +527,18 @@
...
     polygon-pattern-file: url('symbols/military_red_hatch.png');
     polygon-pattern-alignment: global;
     line-color: @military;
     line-width: 3;
     line-opacity: 0.329;
   }
+  [feature = 'forest'][zoom >= 12]  {
+    polygon-pattern-file: url('symbols/forest.png'); // Lch(50,30,135)
+    polygon-pattern-alignment: global;
+    opacity: 0.6; // The entire layer has opacity in case of overlapping forests
```
math1985: Would that result in overlapping symbols? I'm not sure if that would look good.

pnorman: They overlap at 100% opacity and then are reduced to 60%. This might result in some anti-aliasing artifacts if you look *very* closely - the edges of the symbols aren't 100% opacity in the PNG - but I couldn't see any issues.

If polygon-pattern-opacity were used then overlapping forests would have stronger symbols.

project.yaml
```diff
@@ -503,9 +503,10 @@ Layer:
       <<: *osm2pgsql
       table: |-
         (SELECT
-            way, landuse, leisure
+            way, COALESCE(landuse, CASE WHEN "natural" = 'wood' THEN 'forest' END) AS feature
```
math1985: I would add an explicit `ELSE NULL` case. If not, I believe this will give an error whenever the WHERE part is extended. In fact I think the current code is equivalent to `COALESCE(landuse, 'forest')`.

landcover.mss
pnorman: I'm wondering if I should name the column `landuse`, I'm just normalizing the two different wood taggings into one rather than a full feature column setup.

landcover.mss
```diff
@@ -546,13 +527,18 @@
 }
 
 #landuse-overlay {
```
math1985: Could you give a rendering example for forests within military areas?

project.yaml
pnorman: Yes, with the current WHERE clause it is equivalent to `COALESCE(landuse, 'forest')`

landcover.mss
pnorman: image

pnorman:
The forest label text is now as it was before, there was an error in a conversion

image

I'd like to make it stronger, but that would be inconsistent with similar labels and changing all of them is out of scope of this PR in my view.

landcover.mss
math1985: Thanks, looks fine.

althio:
As you wish, it can be treated as a new issue/PR. IMO it needs to be addressed somehow because the pattern is a improvement for the general look but a regression for the readability of labels. Note: I don't know what you mean with 'inconsistent with _similar labels_', that is a bit unclear at least for me. Is it labels for green things, labels for landuse, labels for filled areas, labels for areas with pattern?

My 2 cents for the consistent rule:
If the consistent rule is to apply 30% darker (compared to a reference color),
if the color of icons in pattern is darker than the fill color,
the label color should be 30% darker than the icon color, not the fill color.

nebulon42:
New symbols are good, but I still think that they render far too early and divert attention from more important parts of the map.

Like here at z12 (http://www.openstreetmap.org/#map=12/46.8793/13.5441): 
forest

I would suggest to let symbols render beginning from z14 (as before) and make them less prominent.
Less yellow for grass is also an improvement, but I'm wondering if slightly more yellow as it is now would be better.

imagico:
> `@grassland` was Lch(87.4,27.35,132.4), and all the new greens are 135. `@grass` was the yellow one Lch(89.9,36.5,125). `@grass` is also used for meadow, common, garden, village_green and conservation - it's really a bit of a mess, but I don't consider it in scope for this.

Yes, but landuse=grass and landuse=meadow each have over one million uses while natural=grassland only about 120k - hence this change primarily makes the grassy landuses less yellow.

In general i would make natural=grassland, landuse=grass/meadow/village_green a common yellowish green, maybe somewhat less yellow than grass right now but definitely more than your proposal and landuse=garden different from that.  landuse=conservation should not get a green fill at all.  If this is out of scope here i would leave the colors as they are for now.

math1985:
> I would suggest to let symbols render beginning from z14 (as before) and make them less prominent.

I agree the symbols are too prominent,  especially on z12-z13 and even z14. And even on higher zoom levels, the symbols make it much harder to read paths in forests. I'm not sure what to do about that - maybe strongly reduce the contrast between the background and symbols?

Otherwise great work!

math1985:
By the way, quarry has symbols that are lighter than the background, rather than darker. Even that might be an option.

mboeringa:
> I agree the symbols are too prominent, especially on z12-z13 and even z14. And even on higher zoom levels, the symbols make it much harder to read paths in forests. I'm not sure what to do about that - maybe strongly reduce the contrast between the background and symbols?

Here is a quick and dirty mockup of a much lower contrast symbol made in Photoshop:

forests_low_contrast_symbol

math1985:
@mboeringa I think your preview has too many JPEG artifacts to accurately judge its merits.

mboeringa:
> @mboeringa I think your preview has too many JPEG artifacts to accurately judge its merits.

It was saved at quality setting "9-high". You may actually be perceiving the consequence of the color selection-with-tolerance to selectively pick out the symbols, as JPEG artefacts around the symbols.

As I wrote, it is just a quick and dirty mock-up, feel free to create a better image if you have the tools to generate it (I don't).

math1985:
> You may actually be perceiving the consequence of the color selection-with-tolerance to selectively pick out the symbols, as JPEG artefacts around the symbols.

Yes, that might be what I noticed.

nebulon42:
> maybe strongly reduce the contrast between the background and symbols?

This would be possible by either adjusting transparency or lightness or both.
Below lightness is adjusted.

forest_contrast

> quarry has symbols that are lighter than the background, rather than darker. Even that might be an option.

The last step in the graphics above shall visualise that option.

imagico:
Here an example (from [here](http://www.openstreetmap.org/#map=13/48.0998/7.6841) - a good area for evaluating all the green stuff together by the way) using Lch(66,30,135) and the somewhat finer pattern i showed above as well as `#cdebb0` for grass:

z=13
z=15

In addition i noticed that wetland seems to be rendered above the wood pattern overlay - should be the other way round i think.

mboeringa:
@imagico:
> Here an example (from here - a good area for evaluating all the green stuff together by the way) using Lch(66,30,135) and the somewhat finer pattern i showed above as well as #cdebb0 for grass:

I do think the pattern could be more sparse based on these examples, or only rendered on higher zoom, as @math1985 suggested. Or maybe your non-paired example rendered in these same areas?

I also still wonder about the function of the double symbol: mixed broadleaved and needleleaved, or general "forest" symbol?

nebulon42:
> I also still wonder about the function of the double symbol: mixed broadleaved and needleleaved, or general "forest" symbol?

It's a general symbol. Leaf-types cannot be rendered currently as they are not available in the rendering database.

kocio-pl:
> Leaf-types cannot be rendered currently as they are not available in the rendering database.

But I guess we will have it one day, because it's one of the most rather important feature of the forest, so I'd like to test all the cases beforehand (only "mixed" and "unknown/general" are interesting, other two are dead simple). 

I'd also like to see some rendering examples with the name of the forest, to check if this mix (green+symbols+name) works good together.

math1985:
I agree it would be good to be able to distinguish between 'unknown' and 'mixed' forests,

kocio-pl:
My proposition here would be:
1. Mixed forest - a pattern consisting of _single_ icons of both types entwined on the area
2. General/unknown forest - a pattern of both icons together (_a pair_, like we see it in new propositions) multiplied all over the area

What I like in this idea is that all known/specific types would be consistent and similar then (be it one icon, the other one, or two types scattered) and the unknown/general would look a little different (pairs of trees everywhere).

I'm not sure if this is clear enough, but if not, I'll try to find a better wording. =}

pnorman:
mixed/unknown forests are out of scope for this PR - we don't have leaf_type in the DB, and I don't want to hold this up while figuring out if wood=* is adequate.

> I also still wonder about the function of the double symbol: mixed broadleaved and needleleaved, or general "forest" symbol?

I'd love a single general forest symbol, but as indicated above, one couldn't be found.

> If this is out of scope here i would leave the (grass) colors as they are for now

Yes, the overloading of `@grass` for all kinds of landcover needs to be looked at.

> In addition i noticed that wetland seems to be rendered above the wood pattern overlay - should be the other way round i think

wetland is not an overlay. It probably should be.

pnorman:
Rather than change the lightness I opted to change the opacity. This is equivalent when on top of the forest fill colour, but different when on top of another colour.

image
Note Watershed Park label is a park label, not a forest label

image

image

pnorman:
Visibility on z14 of footway/cycleway isn't outstanding, but I don't think it's really much worse than before. We probably need to make those a bit stronger.

SomeoneElseOSM:
I'm not sure that I'd worry too much about changes to footpath legibility in the standard layer - it isn't possible to use the standard layer at z13 or z14 for route planning since last year's changes that made them mostly illegible at z13 and z14.  See for example Cannock Chase:

http://www.openstreetmap.org/#map=14/52.7413/-1.9926

math1985:
I can read the map in that area just fine, even on mobile. What device do you use?

nebulon42:
The new opacity would roughly correspond to Lch(68, 30, 135), which is good. I'd still start rendering the symbols from z14 onwards as there is still a good amount of noise on z12, z13 when there is a lot of wood/forest present.

In https://github.com/gravitystorm/openstreetmap-carto/pull/1242#issuecomment-70441423 you can see that the overlay covers streams. Is this desirable? This reduces legibility of those features a bit.

math1985:
>  I'd still start rendering the symbols from z14 onwards as there is still a good amount of noise on z12, z13 when there is a lot of wood/forest present.

+1

mboeringa:
+1

pnorman:
I've moved the symbols to z13. I'm happy with how that zoom looks.

Assigning for @gravitystorm to make the call on merging, as it's a significant cartographic change.

> In #1242 (comment) you can see that the overlay covers streams. Is this desirable? This reduces legibility of those features a bit.

I'm not sure if it's desirable - but layer ordering is unchanged.

imagico:
I once more would suggest to remove the grass color change until a proper solution can be established for it that it clearly an overall improvement.  This is a high impact change affecting ~3 million features, it is really quite unrelated to the rest of this PR and has not really received significant attention in discussion here.

math1985:
> Multipolygons do not normally exist in the rendering database but in any case, it remains the same - a clearing within a forest will have had a hole cut out, and that hole is not within the polygon.

I had some more thought about this, and I think this (rendering the pattern on top of other elements within a forest) will cause a lot of problems. Rendering the pattern on top of forest creeks was already mentioned. Another problem are houses in forests. Do we really expect users to modify the multipolygon for every house drawn in the forest? See [here](http://www.openstreetmap.org/#map=18/49.37050/-123.18697) for an example area near Vancouver. The same problem holds for roads through forests - I don't think they should be necessarily cut out of the multipolygon, and rendering icons on top of them wouldn't look good. So I would now be in favour of tying the forest icon rendering to the rendering of the forest background.

imagico:
I agree on rivers and houses but roads seem to be already drawn above the forest pattern and streams are fine below it i think.

Best would IMO be to draw the pattern above all other landcovers (including wetland) but below anything else.

math1985:
> I once more would suggest to remove the grass color change until a proper solution can be established for it that it clearly an overall improvement.

I agree it would be better to split independent changes (which this is IMO) into separate pull requests.

mkoniecz:
> Best would IMO be to draw the pattern above all other landcovers (including wetland) but below anything else.

I agree with this. Especially as typically buildings are within forest (under trees), so creating multipolygons would be incorrect.

gravitystorm:
Unassigning, as per previous comments regarding independent changes and grassland.

pnorman:
Rolled back the `@grass` change.

For reference, there's the building difference.

image

I'm okay with buildings under trees being rendered under trees.

pnorman:
> Unassigning, as per previous comments regarding independent changes and grassland.

Resolved, so reassigning.

mboeringa:
> I'm okay with buildings under trees being rendered under trees.

I am not so... There is no way to tell if these buildings are really "under trees" in reality. It also goes against many cartographic conventions I think, although I am no cartographer from profession, but I haven't seen this ever on a professionally designed map. It also starts suggesting these features aren't buildings at all, but maybe some kind of plantation...

math1985:
I agree, I also don't like the proposed rendering in the image.

Rovastar:
I not really like it either. The tree icons over the buildings just don't look right.

pnorman:
per discussion moving symbols above other landcover but below buildings

dieterdreist:
2015-01-26 20:04 GMT+01:00 mboeringa <notifications@github.com>:

> I'm okay with buildings under trees being rendered under trees.
>
> I am not so... There is no way to tell if these buildings are really
> "under trees" in reality. It also goes against many cartographic
> conventions I think, although I am no cartographer from profession, but I
> haven't seen this ever on a professionally designed map. It also starts
> suggesting these features aren't buildings at all, but maybe some kind of
> plantation...
>


+1, to me these look like underground buildings now. The wider buildings in
this spot definitely aren't "under trees" in the real world in a way the
rendering might suggest, and if I imagine even bigger buildings this will
become uglier.

To me this is a mapping error (as the landuse isn't "forest" on a spot that
is covered with a building or a pitch, or a road, or ...).

I am not yet tired of repeating the idea to use _landuse_ as an attribute
for the usage of land, and _natural_ to define "geo-features" (e.g. with
name). Having a natural=forest name=xy overlap with areas of landuse=forest
(including recently cut areas with no trees on them currently) and yet
other areas of landcover=trees (where there actually are trees) would make
perfect sense IMHO.

daganzdaanda:
I'm impressed, this looks great. I also agree that buildings should not be covered by the symbols.
We will probably still see some weird situations that result from bad tagging. But that's a good thing, they will hopefully be cleaned up sooner that way. 

@dieterdreist I really like the "landcover" tag for what is actually physically on the ground. It should definitely be used more (I resolve to try and remember that...) Also, how about getting it into JOSM and other editors?

project.yaml
```diff
@@ -503,9 +503,10 @@ Layer:
       <<: *osm2pgsql
       table: |-
         (SELECT
-            way, landuse, leisure
```
pnorman: I'm not sure why we were returning leisure before

daganzdaanda:
I've looked at some German topographical maps and they use tree symbols with a lot more space, maybe a third of the density of what is suggested here. It still works at explaining the forest type: 
http://geoportal.bayern.de/bayernatlas/?X=5311549.22&Y=4470867.01&zoom=10&lang=de&bgLayer=tk&topic=ba
They start with the symbols at zoom=10, which looks closer to z15 than z14 in OSM.

mboeringa:
> I've looked at some German topographical maps and **they use tree symbols with a lot more space**, maybe a third of the density of what is suggested here. It still works at explaining the forest type:

The same for OpenTopoMap:

http://opentopomap.org/#map=15/46.88683/8.66075

RAytoun:
I am in favour of the mixed trees but am uncomfortable with the "two together" random pattern as it makes it look "busy". The random single trees still depict a mixed forest/wood but without the clutter. The green background already clearly shows the extent of the feature so the symbol is just to indicate type of trees so does not need to be so dense, as shown in the examples indicated on the topo maps. 
I am not in favour of showing the trees growing on top of buildings.

imagico:
Rendering the wood overlay together with the wetland overlay seems a reasonable solution, the only situation where this might cause troubles is areas tagged both natural=wetland and landuse=forest - these would not get a trees overlay.  There are only ~1000 features with that tag combination though and generally the more appropriate and more popular tagging would be natural=wetland + wetland=swamp (41k uses).

IMO this is ready to be merged - some fine tuning of the pattern density could be in order but it makes much more sense to discuss this after broad testing globally.

pnorman:
Assigned for review.

mkoniecz:
It solves problem of something is forest and university/park/playground/foobar. During testing rendering I found multiple multipolygon problems in a single forest.

 But new forest is quite noisy what is causing problems - for example zoo in forest is rendered like this:

selection_003

Maybe lower density of symbols would be a good idea? Both for forests and other noisy features like wetlands?

mkoniecz:
A small bug - closed way tagged with both landuse=forest and natural=wetland is not rendering tree pattern - solid green background and wetland patter is rendered.

Area with two separate closed ways - landuse=forest and natural=wetland has both patterns.

pnorman:
As both wetlands and forests *may* use the natural key, I decided against trying to handle an object tagged as both. There are approximately 1000 objects ttagged as both.

imagico:
It would be great if this could get out of the door some time soon - i have some changes w.r.t. wetland in preparation that would significantly interfere with this and that would be much less work to do if this is already merged.

imagico:
I am not sure if this is to be considered rejected or on hold indefinitely.  Considering the strong visual impact it would be important for other changes to have a decision here, i.e. if wood/forest color is going to stay as it is or going to be like in this PR.

mboeringa:
> **Considering the strong visual impact**

I think this is exactly the main problem with this one being accepted / implemented.

Even in the latest proposals, the visual impact of the change is to big IMO (and probably others...). 

Why can't we have something more subtle like OpenTopoMap (e.g.http://opentopomap.org/#map=15/48.08533/9.20581), with a much more dispersed, and significantly smaller non-obtrusive symbols, like in topographic maps?

In fact, I suggest maybe copying these OpenTopoMap symbols to standard... while using the randomness algorithm by @pnorman, but spaced like in OpenTopoMap.

kocio-pl:
@mboeringa +1

brycenesbitt:
I think it worth trying a stippled randomized green/dark green pattern.  It would be less flat than the current solid green, without the overwhelming visual strength of the two trees.  The two trees demand visual attention, for a feature that's often a background feature.

kocio-pl:
This is a big and important change, so it should be well discussed, but after 6 months and almost hundred of comments given I think we could summarize current state of affairs and at least decide how we want to carry on with this issue.

matkoniecz:
I think that it would be a good idea to resurrect "use a random symbology for forests". I liked idea, but I had doubts about how new forests looked like. After reviewing it again - I am no longer against new colours, especially new natural=grassland is much better (current natural=grassland is closer to shrub or bushes than grass). New natural=scrub is also better, current colour seems wrong too me. Current landuse=forest still seems more "alive" than desatured version from PR but it is not good argument against this change, especially as current forest colour is rather too strong.

But I still think that forest symbol should be more subtle (closer to colour of forest itself). Especially in areas with paths or footways it is currently too distracting. 

for example from test on http://www.openstreetmap.org/#map=18/50.05195/19.85797




I am also confused by

> opacity: 0.4; // The entire layer has opacity to handle overlapping forests

How opacity is helping in that?

Also, how symbols/forest.png file was generated? I want to experiment with lower symbol density and different color and it would be a pity to do manually something that may be easily automated.

kocio-pl:
Just a note: since we've changed the way we display zoo (colored border instead of "z" pattern), [this particular problem](https://github.com/gravitystorm/openstreetmap-carto/pull/1242#issuecomment-76175505) no longer exists.

matkoniecz:
Yes, zoo in forest is no longer problematic. But paths are still less readable than I would prefer.

imagico:
Looking into this and trying out some more options for the pattern is on my list of things to do when time permits.  In general i think taking the paired symbol as a design goal is probably not the best approach since ultimately this will only serve as a fallback pattern when no leaf_type is specified.  It would be an option to use smaller (and less readable) symbols or not use a pattern at all for that - as long as wood/forest with detailed tagging is looking good.

Trying some options for the symbol color is another possibility.  I also thought about the possibility of encoding leaf_cycle in the symbol color (something greener for evergreen and something more brown for deciduous)

kocio-pl:
> Also, how symbols/forest.png file was generated? I want to experiment with lower symbol density and different color and it would be a pity to do manually something that may be easily automated.

I guess @imagico [jsdotpattern](http://www.imagico.de/map/jsdotpattern.php) is the tool you are looking for. :smiley:

matkoniecz:
@kocio-pl Thanks!

I experimented with not so strong patterns and I found one that I like. My goal was to makekeep path and tracks in forests readable, also on lower zoom levels. Not so strong pattern seems to be working also on z12.

paler

heavily forested area (Kosice): https://cloud.githubusercontent.com/assets/899988/8937158/0d5bbba4-3557-11e5-8a5c-c3f441ffe012.png

forest near city (Vienna, Kraków): 
https://cloud.githubusercontent.com/assets/899988/8937306/cf320e40-3557-11e5-9747-6b882a0cfa26.png
https://cloud.githubusercontent.com/assets/899988/8937307/cf3bc2d2-3557-11e5-93d6-41067a1079b5.png

code: https://github.com/matkoniecz/openstreetmap-carto/commits/test_forest (I attempted rebasing - but it is still far, far beyond current state for now)

Maybe use this more delicate pattern on earlier zoom levels and later a stronger one?

kocio-pl:
Much better pattern density and color for me, but I'm not sure if it wouldn't be even better with even more lightening.

I also don't think we need this pattern earlier than currently (z>=14).

matkoniecz:
and for location provided by @nebulon42 (https://github.com/gravitystorm/openstreetmap-carto/pull/1242#issuecomment-71189197)

https://cloud.githubusercontent.com/assets/899988/8937920/ae130a8a-355b-11e5-9eb5-603cd4c6de60.png

kocio-pl:
For me this example proves it further that z=13 is too early: z12 is worse and z=13 is better, but worse than with no pattern, because tracks are too dense. Probably z=14 and z=15 could gain some readability if the pattern was toned down.

imagico:
Ok - i have some more options here.  First of all i moved away from the tree pair concept.  We have seen a lot of variations of this already but it is very strong and possibilities to weaken its appearance while keeping it readable are limited.  And ultimately it might make more sense to simply use no pattern at all for woodland of unknown composition - it well communicates something is missing (i.e. supplementary tags).

So the following suggestions include different variants for `leaf_type=broadleaved`, `leaf_type=needleleaved` and `leaf_type=mixed`.  This is currently not available of course.  As a temporary solution until this changes it would be possible to use the mixed variant of course.

First the simple tree symbols that were already used above in a denser and a coarser arrangement.  Color is fairly low contrast similar to the last one @matkoniecz showed.  The coarser pattern looks less noisy in the abstract sample of course but keep in mind that with small polygons large symbol distances often do not work well.

simple symbols

Second are somewhat more complex tree symbols - the broadleaved one is already used in the swamp pattern.  In my opinion these might work better since they are more intuitive especially in cases where the shape is not perfectly clear (due to low contrast or other interfering elements).

complex symbols

As a third possibility i tried a pure structure pattern without distinct symbols.  This is kind of meant to resemble a view from above.  The mixed variant is somewhat difficult to get right of course.  Two variants with slightly different settings:

structure patterns

The huge advantage of this would be it does not require the symbols to be readable, it works with less contrast between pattern and background color and has less problems with small polygons.  But of course it is also less explicit in meaning.

And finally a quick glance on the possibility to illustrate leaf_cycle by varying the overlay color.  Not sure if this is a good idea but might be worth testing.

leaf_cycle

I won't test all the different variants in actual rendering but you can try out the ones you find promising yourself - here are the pattern images:

http://www.imagico.de/files/smixed2_overlay.png
http://www.imagico.de/files/smixed_overlay.png
http://www.imagico.de/files/sbroad2_overlay.png
http://www.imagico.de/files/sbroad_overlay.png
http://www.imagico.de/files/sneedle2_overlay.png
http://www.imagico.de/files/sneedle_overlay.png

http://www.imagico.de/files/cmixed2_overlay.png
http://www.imagico.de/files/cmixed_overlay.png
http://www.imagico.de/files/cbroad2_overlay.png
http://www.imagico.de/files/cbroad_overlay.png
http://www.imagico.de/files/cneedle2_overlay.png
http://www.imagico.de/files/cneedle_overlay.png

http://www.imagico.de/files/mixed4_overlay.png
http://www.imagico.de/files/mixed3_overlay.png
http://www.imagico.de/files/broad_overlay.png
http://www.imagico.de/files/broad3_overlay.png
http://www.imagico.de/files/needle_overlay.png
http://www.imagico.de/files/needle2_overlay.png

dieterdreist:
very nice work Christoph. From the icons I prefer the more complex ones, especially that they are "grounded" by a small piece of soil/cast shadow, this is an important graphical detail the simple symbols are missing.

I also like the abstract "top view" a lot from a graphical point of view but I feel that it might suggest more detail than there is actually mapped (especially in the mixed version but also in the others, people might think that these are actual trees rather than just an abstract representation of a kind of tree).

mboeringa:
@imagico: good work, I like the subtle colors and symbols of the latest suggestions, we're finally getting in the direction of something that would not dominate the entire map, but meld-in with the rest.

Personally, I am in favour of the simple small symbols and a coarser pattern density, but with the subtle colors you used, the complex symbols aren't that distracting anymore, although I would still recommend the sparse distribution for those symbols.

The idea of the "patterns" is a nice one as well, although you may be right, that this is less meaningful. **I do think it would be nice to see some real world examples of the use of this pattern though** against the true data, instead of the sample squares presented here, to get a more informed opinion about this option.

SK53:
Minor comments:

* I never measured symbol density on the maps in the original blog posts: perhaps I should have done so. I prefer the coarser symbol distribution, but suspect that I would like something in-between the two examples more. (I am also interested in a later possibility of interspersing tree symbology with other symbols to show other wood/forest properties but **not** on the standard render)
* The tree 'footer' was something I had noted in many prior cartographic examples and does really add something.
* The colour distinction for leaf_cycle is very subtle, but I thing familiarity would increase recognition. Of course I'm very much in favour of something which will encourage better mapping.
* For completeness a very abstract (circles & triangles) style perhaps ought to be considered.

Again I think the Krakow example should be tested with these: I was quite surprised how intrusive the symbology was at that scale, and felt that we had missed capturing something subtle from similar printed maps.

matkoniecz:
> felt that we had missed capturing something subtle from similar printed maps

From your examples - note how on  and https://upload.wikimedia.org/wikipedia/commons/6/64/Country-side_around_Troitsk%2C_Moscow.png tree symbols are kept from intersecting paths/roads.

SK53:
@matkoniecz Indeed, but this is not something we can do with the current approach as an overlay. I suspect the OSGB StreetView actually cuts out the green pattern for forests/woods entirely when a path or track is shown, but 1:50k OSGB does not do that http://binged.it/1VUHvdf (and multiple overlays make it fairly difficult to read).

imagico:
Thanks for the comments.

* regarding symbol density - i chose the extremes of the range i consider useful here for single tree symbols (for pairs you might make it even coarser but that's part of the problem why this works poorly).  In between densities are possible of course - densities used in jsdotpattern were 17 and 28.
* regarding color - in the abstract samples the low contrast is not much of a problem of course - if this also works in real use remains to be seen.  There is of course also the possibility to fade in the symbols when zooming in, i.e. make them more subtle at the low zooms but stronger at the higher levels.  For the leaf_cycle coloring - the required color difference depends a lot on how heavy the pattern is - for the coarse one the current choice is fairly weak and might be difficult to recognize but for the heavier patterns it might already be too strong.  You have to be careful not to influence the overall color impression so much that different areas are no more commonly recognizable as the same base color.

Abstract 'german style' triangle/circle symbols are here in two different sizes:

abstract tree symbols
abstract tree symbols large

The small ones probably need a better contrast to be readable.

Vort:
I think this symbols better than others:
https://camo.githubusercontent.com/e82ede7ebc3f4f154e8e617368c67cac98862918/687474703a2f2f7777772e696d616769636f2e64652f66696c65732f776f6f645f7061747465726e7332612e706e67
Because they are better recognizable as trees.

kocio-pl:
I strongly support dropping pairs. This was more or less my preferred idea (I just wanted the pairs to be used only for unspecified type of forest), thanks for picking it up!

I prefer complex tree symbols sparse, because they are easiest to recognize and won't clutter the area. I guess even natural=wood, which is commonly used for smaller tree areas, it would be still usable, however field test would be needed.

BTW: I think unification could also include [landcover=trees tag](http://taginfo.openstreetmap.org/tags/landcover=trees), since it's quite popular and we have probably no other neutral way of describing tree areas. Lack of wiki page is easy to fill, I will ask about it on Tagging list.

dieterdreist:
sent from a phone

> Am 02.08.2015 um 13:16 schrieb Mateusz Konieczny <notifications@github.com>:
> 
> From your examples - note how on  and https://upload.wikimedia.org/wikipedia/commons/6/64/Country-side_around_Troitsk%2C_Moscow.png tree symbols are kept from in tersecti ng paths/roads.


these look like actual trees (surveyed), not merely abstract symbolic fills

pnorman:
I would not worry about leaf type right now, as that's out of scope of this issue.

imagico:
> I would not worry about leaf type right now, as that's out of scope of this issue.

IMO it makes sense to plan ahead and consider the possibilities to show leaf_type and possibly leaf_cycle.  As said the idea would be later when leaf_type can be rendered to not render the pattern any more in areas not tagged with leaf_type and until then use the mixed version as temporary pattern.

I did not mean to hijack this PR - if you or anyone else want to further evaluate the tree pair concept of a different generic approach that is fine - i just have my doubts this will lead to good results now and the question remains anyway how this can be extended to differentiate leaf_type.

matkoniecz:
@Imagico 

I think that pairs are slightly uglier than individual tree - and it is not a problem. Once hstore will be available this style may represent forests with missing leaf_type, and forests with supplied leaf_type may use a proper and prettier pattern like proposed in https://github.com/gravitystorm/openstreetmap-carto/pull/1242#issuecomment-126958042

pnorman:
closing in favour of #1728

Issue #954, opened 2014-09-16
Render place=borough and place=municipality


sanderd17:
place=borough is the more important version of place=suburb (i.e. boroughs mostly have a separate council and separate governance).

Boroughs are used in Berlin, but not rendered: http://overpass-turbo.eu/map.html?Q=area[%22name%22%3D%22Berlin%22]%3B%0A%28%0A%20%20node[place%3Dborough]%28area%29%3B%0A%29%3B%0Aout%3B%0A

Boroughs are certainly more important than suburbs. But probably less important than towns.

Municipalities are also used in Germany, but not rendered. F.e. http://www.openstreetmap.org/node/240123243#map=13/47.7520/8.0760

Municipalities should probably be between suburb and village in importance.

In Belgium, we're planning to replace the Antwerp suburbs by boroughs (since they have a district council and governance), and we are discussing to put the place=municipality tag on municipalities (sometimes replacing place=village nodes when the village is also the centre of a municipality).

The proposed tagging schema for Belgium will clear up some inconsistencies. But I'd like to get these tags rendered to avoid edit wars, or edits from newbies who like to see their names rendered.

The tags themselves are not new, and documented in the wiki: http://wiki.openstreetmap.org/wiki/Key:place

This is related to #948 , but I didn't want to interrupt the discussion there.

PR #1247, opened 2015-01-21
Clean up place name rendering

### LINKED ISSUE #107: Render place=county/neighbourhood/block
It would be nice if more place types were rendered at appropriate zooms. For example county, neighbourhood, block etc. Both on nodes and closed ways/relations.

### LINKED ISSUE #522: Non-capital=yes values
We currently check for ``capital='yes'`` on #placenames-capital, but this doesn't match capital=4. We should probably check either for ``capital = ANY (ARRAY['1','2','3','4','yes'])`` or ``capital IS NOT NULL AND capital != 'no'``

(note: is the NOT NULL check necessary, or implied by the inequality check?)

### LINKED ISSUE #734: Suburb font size too big in place=town or place=village
For example look at this place: http://www.openstreetmap.org/#map=14/45.4914/18.0892

Font for place seams too small or font for suburb is too big. Can this be fixed to place is more clearly visable and not to be mistaken for suburb?

See also https://trac.openstreetmap.org/ticket/2291.

### LINKED ISSUE #798: Add rendering for place=quarter
A quarter is an area of a city smaller than place=suburb, but larger than place=neighbourhood. A quarter may have an administrative boundary, or only a node marking its nominal centre.

The proposal for this tag was approved in 2012-01.
- http://wiki.openstreetmap.org/wiki/Proposed_features/place%3Dquarter
- http://wiki.openstreetmap.org/wiki/Tag:place%3Dquarter

It would be nice if this tag could be rendered.

See also https://trac.openstreetmap.org/ticket/5199.

### LINKED ISSUE #948: Make place=locality label to be less strong
place=locality is rendered too strong considering its meaning. At http://www.openstreetmap.org/#map=14/49.1133/19.7303 you can see how various localities makes villages hard to recognise. Please make it to render less strong, eg. by making font smaller and more fade.

### LINKED ISSUE #1204: Adjust size of placenames at z14
The size of the placenames at z14 looks a little bit strange, because town (e.g. Horw) and suburb (e.g. Ennethorw) as well as village (e.g. Hergiswil) and isolated_dwelling (e.g. Halti) have similar sizes.

placename size
http://www.openstreetmap.org/#map=14/47.0005/8.3137

I suggest to reduce the size of suburb names and to increase the size of village names. In fact, some suburbs have a higher population than towns, but suburbs can be part of towns (see the example) and I guess town names are more important than suburb names in most cases.

math1985:
* Don't render place=town on z16 and higher, and place=suburb and place=village
  on z17 and higher.
* Render place=hamlet, place=locality, place=neighbourhood,
  place=isolated_dwelling, and place=farm from z15 rather than from z14 (resolves
  #948).
* Render place=town black rather than gray.
* Make font sizes more consistent by increasing font size for place=town on z12
  and z13, for place=suburb on z16, and for place=village on z14 and z16
  (resolves #734, resolves #1204).
* Treat capital=4 like capital=yes (resolves #522).
* Make text-wrap consistent.

### DIFF
--- a/placenames.mss
+++ b/placenames.mss
@@ -46,15 +46,15 @@
     text-face-name: @book-fonts;
     text-halo-radius: 1.5;
     text-halo-fill: rgba(255,255,255,0.6);
-    text-wrap-width: 25;
+    text-wrap-width: 45;
     text-min-distance: 10;
     [zoom >= 6] {
       text-size: 12;
-      text-wrap-width: 70;
+      text-wrap-width: 60;
     }
     [zoom >= 11] {
       text-size: 15;
-      text-wrap-width: 90;
+      text-wrap-width: 75;
     }
   }
 }
@@ -68,15 +68,15 @@
       text-face-name: @book-fonts;
       text-halo-radius: 1.5;
       text-halo-fill: rgba(255,255,255,0.6);
-      text-wrap-width: 25;
+      text-wrap-width: 45;
       text-min-distance: 10;
       [zoom >= 9] {
         text-size: 12;
-      text-wrap-width: 70;
+        text-wrap-width: 60;
       }
       [zoom >= 11] {
         text-size: 15;
-      text-wrap-width: 90;
+        text-wrap-width: 75;
       }
     }
   }
@@ -84,51 +84,57 @@
 
 #placenames-medium::town {
   [place = 'town'] {
-    [zoom >= 9] {
+    [zoom >= 9][zoom < 16] {
       text-name: "[name]";
       text-size: 9;
       text-fill: @placenames;
       text-face-name: @book-fonts;
       text-halo-radius: 1.5;
       text-halo-fill: rgba(255,255,255,0.6);
-      text-wrap-width: 50;
+      text-wrap-width: 45;
       text-min-distance: 10;
-    }
-    [zoom >= 11] {
-      text-size: 11;
-      text-wrap-width: 70;
-    }
-    [zoom >= 14] {
-      text-size: 15;
-      text-wrap-width: 90;
-      text-fill: @placenames-light;
-      text-halo-fill: white;
+      [zoom >= 11] {
+        text-size: 11;
+        text-wrap-width: 55;
+      }
+      [zoom >= 12] {
+        text-size: 13;
+        text-wrap-width: 65;
+      }
+      [zoom >= 14] {
+        text-size: 15;
+        text-wrap-width: 75;
+      }
     }
   }
 }
 
 #placenames-small::suburb {
-  [place = 'suburb'][zoom >= 12] {
+  [place = 'suburb'][zoom >= 12][zoom < 17] {
     text-name: "[name]";
     text-size: 11;
     text-fill: @placenames;
     text-face-name: @book-fonts;
     text-halo-radius: 1.5;
     text-halo-fill: rgba(255,255,255,0.6);
-    text-wrap-width: 70;
+    text-wrap-width: 55;
     text-min-distance: 10;
     [zoom >= 14] {
       text-size: 14;
-      text-wrap-width: 90;
+      text-wrap-width: 70;
       text-fill: @placenames-light;
       text-halo-fill: white;
+    }
+    [zoom >= 16] {
+      text-size: 15;
+      text-wrap-width: 75;
     }
   }
 }
 
 #placenames-small::village {
   [place = 'village'] {
-    [zoom >=12] {
+    [zoom >= 12][zoom < 17] {
       text-name: "[name]";
       text-size: 10;
       text-fill: @placenames;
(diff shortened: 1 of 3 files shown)


pnorman:
I like the other changes (in principle, haven't previewed yet) but I'm not sure about adding the new tags. In particular, borough was rejected (#954)

https://github.com/gravitystorm/openstreetmap-carto/issues/954#issuecomment-55747887
> "borough" and "municipality" are both, in British English, terms for administrative levels, not places. They are not in the hierarchy of hamlet/village/town/city, nor are they in the similar set of place types suburb/metropolis/conurbation etc.

> I've no idea how they came into being in the place hierarchy on the wiki instead of either a) being part of the admin_level hierarchy or b) better values, related to places, being chosen.

> I suspect this is just adding more tags for dubious overall value. Having a limited set of place tags provides *more* value than having an large number of place tags that are hard if not impossible to choose between.

> If a village is also a municipality, then that doesn't mean it stops being a village. If a suburb has a local council, then it's still a suburb. The administrative classification should be recorded separately.

math1985:
Thanks, I will remove borough. What about the other new tags?

pnorman:
I'm skeptical - but haven't looked into them in any detail. I'm also not a native en_GB speaker, only en_CA.

We do not want to encourage a proliferation of place values as that reduces their value for data consumers

math1985:
* About 60% of the place=quarter occurrences are in Japan.
* About half the place=city_block occurrences are in Stockholm.
* About 99% of the place=block occurrences are in Malmo, Lund and Landskrona.

Block names seem rather particular to Sweden, but they seem to be important toponyms there.

math1985:
So, should I drop quarter/block/city_block from this PR?

math1985:
> 

So, should I drop quarter/block/city_block from this PR?

math1985:
> So, should I drop quarter/block/city_block from this PR?

Any opinions on this?

pnorman:
I'd drop the new tags from the PR.

math1985:
> I'd drop the new tags from the PR.

Dropped the tags and rebased.

Do we consider the proposal to add rendering for quarter/block/city_block as 'rejected'?

gravitystorm:
> Do we consider the proposal to add rendering for quarter/block/city_block as 'rejected'?

Yes, no more place name values please.

dieterdreist:
2015-02-04 14:12 GMT+01:00 Andy Allan <notifications@github.com>:

> Do we consider the proposal to add rendering for quarter/block/city_block
> as 'rejected'?
>
> Yes, no more place name values please.
>


IMHO there is a point in having at least some of them for places where they
are used (probably for good reason). E.g. in the centre of Mannheim there
are blocks but only street names for some principal road:
http://de.wikipedia.org/wiki/Quadratestadt

In other cases, "suburb" and "neighbourhood" are not sufficient to
represent the hierarchy of toponyms (not talking about administrative units
here), e.g. Rome. (We have not yet mapped all those toponyms that are in
use, but we will sooner or later do it). That's where the proposal to have
also a place=quarter tag comes from.

Also, using "suburb" for stuff that is not a "suburb" is neither intuitive
nor semantically correct, so despite all the historical reasons it wouldn't
be too bad IMHO to change this sooner or later to something more pertinent.

If the main style decides to ignore these tags that are at least in some
cases there for good reason (the established tags don't hit it), we will
encourage people to map with tags that don't apply (tagging for the
renderer) in order to retain the usability of the map.

RAytoun:
Many old towns have historic areas such as "old quarter", which are still  in common use and named on town and tourist maps as a place to visit. These areas do not necessarily coincide with administrative boundaries and do not have administrative status.

placenames.mss
```diff
@@ -84,51 +84,57 @@
...
+      }
+      [zoom >= 14] {
+        text-size: 15;
+        text-wrap-width: 75;
+      }
     }
   }
 }
 
 #placenames-small::suburb {
```
pnorman: We probably want to move these out of their own attachments and do an order by on the query - but out of scope of this PR

math1985: @pnorman Is there an advantage to that? Performance?

kocio-pl:
> In other cases, "suburb" and "neighbourhood" are not sufficient to represent the hierarchy of toponyms (not talking about administrative units here), e.g. Rome. (We have not yet mapped all those toponyms that are in use, but we will sooner or later do it). That's where the proposal to have also a place=quarter tag comes from. 

That's also the case in Warsaw. We have three levels deep place naming hierarchy (suburb/quarter/neighbourhood - and it's not administrative too) and quarters are not easily negligible because its name is visible on every house number and city name plate, so it's used as a main navigation unit:

https://pl.wikipedia.org/wiki/Miejski_System_Informacji_%28Warszawa%29#mediaviewer/File:MSI_Trakt_Kr%C3%B3lewski.jpg
https://pl.wikipedia.org/wiki/Miejski_System_Informacji_%28Warszawa%29#mediaviewer/File:Tablica_MSI_ulica_Alternatywy_w_Warszawie.JPG

PR #978, opened 2014-09-25
Use French-style tree rendering

math1985:
Merge tree rendering style developed by @cquest for French rendering, see
https://github.com/cquest/osmfr-cartocss/.

* Resolves #138
* Resolves #838

PR #1249, opened 2015-01-22
Add icon for amenity=bench and amenity=waste_basket

### LINKED ISSUE #194: Add rendering for amenity=bench
I think "amenity=bench" should be rendered, specially those that are mapped as ways, since that often means long stone benches that also act as a barrier.

Bench icons could start showing at level 20, while bench ways should start at 18.
Maybe a brownish color for both symbols and lines, or use the "color" tag if available (is that possible?) or even a color depending on the material (stone=grey, wood=brown, etc).

### LINKED ISSUE #418: Add rendering for amenity=waste_basket
Popular tag is not rendered on the map.
Please add this tag.

http://wiki.openstreetmap.org/wiki/Tag:amenity%3Dwaste_basket

math1985:
* Add icon for amenity=bench from z18 (resolves #194)
* Add icon for amenity=waste_basket from z19 (resolves #418)

These icons are taken from the SJJB repository, so they are still not aligned to the pixels.

Preview:
screenshot from 2015-01-30 21 30 06

### DIFF
--- a/amenity-points.mss
+++ b/amenity-points.mss
@@ -701,6 +701,18 @@
     marker-placement: interior;
     marker-clip: false;
   }
+
+  [feature = 'amenity_bench'][zoom >= 18] {
+    marker-file: url('symbols/bench.16.svg');
+    marker-fill: #744a08;
+    marker-placement: interior;
+  }
+
+  [feature = 'amenity_waste_basket'][zoom >= 19] {
+    marker-file: url('symbols/waste_basket.16.svg');
+    marker-fill: #744a08;
+    marker-placement: interior;
+  }
 }
 
 .amenity-low-priority {
--- a/amenity-points.mss
+++ b/amenity-points.mss
@@ -763,14 +763,14 @@
     }
   }
 
-  [amenity = 'bench'][zoom >= 18]::amenity {
+  [amenity = 'bench'][zoom >= 19]::amenity {
     marker-file: url('symbols/bench.16.svg');
     marker-fill: #744a08;
     marker-placement: interior;
   }
 
   [amenity = 'waste_basket'][zoom >= 19]::amenity {
-    marker-file: url('symbols/waste_basket.16.svg');
+    marker-file: url('symbols/waste_basket.10.svg');
     marker-fill: #744a08;
     marker-placement: interior;
   }
--- a/amenity-points.mss
+++ b/amenity-points.mss
@@ -701,18 +701,6 @@
     marker-placement: interior;
     marker-clip: false;
   }
-
-  [feature = 'amenity_bench'][zoom >= 18] {
-    marker-file: url('symbols/bench.16.svg');
-    marker-fill: #744a08;
-    marker-placement: interior;
-  }
-
-  [feature = 'amenity_waste_basket'][zoom >= 19] {
-    marker-file: url('symbols/waste_basket.16.svg');
-    marker-fill: #744a08;
-    marker-placement: interior;
-  }
 }
 
 .amenity-low-priority {
@@ -773,6 +761,18 @@
         marker-width: 4;
       }
     }
+  }
+
+  [amenity = 'bench'][zoom >= 18]::amenity {
+    marker-file: url('symbols/bench.16.svg');
+    marker-fill: #744a08;
+    marker-placement: interior;
+  }
+
+  [amenity = 'waste_basket'][zoom >= 19]::amenity {
+    marker-file: url('symbols/waste_basket.16.svg');
+    marker-fill: #744a08;
+    marker-placement: interior;
   }
 }
 
--- a/project.mml
+++ b/project.mml
@@ -1999,7 +1999,7 @@
       "srs": "+proj=merc +a=6378137 +b=6378137 +lat_ts=0.0 +lon_0=0.0 +x_0=0.0 +y_0=0.0 +k=1.0 +units=m +nadgrids=@null +wktext +no_defs +over",
       "Datasource": {
         "extent": "-20037508,-20037508,20037508,20037508",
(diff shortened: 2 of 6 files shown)


HolgerJeromin:
At least the bench should be included direct as SVG.

floscher:
> At least the bench should be included direct as SVG.

:+1:, I've opened pull request math1985/openstreetmap-carto#5 for this.

nebulon42:
As with #1240 for picnic table I'm not sure if the bench should be slightly thicker? With 1.5px thickness it would look like that (more blurry of course):
bench_thicker
([bench-16.svg](https://github.com/nebulon42/osmic/blob/master/outdoor/bench-16.svg))

pnorman:
I prefer 1.5 pixels

nebulon42:
This would be my proposal for waste basket ([waste-basket-16.svg](https://github.com/nebulon42/osmic/blob/master/outdoor/waste-basket-16.svg)):
waste_basket

math1985:
I merged in the svg bench icon.

I don't particularly like the bin icon - not sure what to do about it, maybe it would look better with a straight bottom?

nebulon42:
Maybe this: waste-basket-16

mkoniecz:
@nebulon42 - I like this icon.

math1985:
Yes, looks better to me. Thanks!

nebulon42:
I have updated the corresponding icon: [waste-basket-16.svg](https://github.com/nebulon42/osmic/blob/master/outdoor/waste-basket-16.svg)

math1985:
I updated the PR with both svg icons.

nebulon42:
To me bench.16.svg is still displayed as 1px version. Did you mean to keep that?

math1985:
>  Did you mean to keep that?

No I forgot that one, it should be updated now.

HolgerJeromin:
A bench is larger than a basket. They are often near. It looks strange if the basket has more visual weight.

nebulon42:
Normally, I'd like to use every pixel of the restricted 16px frame for the icon, but I agree in a way. I have reduced the profile of the waste basket a bit: waste-basket-16

I like the low profile of the bench. An option would be to add a back-rest to it as is done in the Humanitarian style, but I'm not too fond of that.

daganzdaanda:
The new waste basket looks good to me.
I'm happy that benches will appear on the map soon!
But let's not forget that some benches are tagged on lines. (See also https://github.com/gravitystorm/openstreetmap-carto/issues/194#issuecomment-25430056 ) I would like to see those, too, but I have no good idea. Just a fat grey line, maybe with the symbol in the middle?

HolgerJeromin:
Yes. I would show the Icon for benches as ways, too. But showing the dimension seems not easy possible, so drop this information.

math1985:
How should we do that? Does it make sense to add an extra query specifically for linear benches?

math1985:
Added new bin icon and updated preview.

mkoniecz:
Maybe bin icon can be even smaller?

math1985:
I'd prefer to leave the icon as is.

mkoniecz:
I propose moving these two icons to .amenity-low-priority to avoid blocking more important icons and labels.

althio:
The guidance https://github.com/gravitystorm/openstreetmap-carto/pull/1282#issuecomment-74496085 about icon size and weight is IMO very applicable here. As mentioned earlier https://github.com/gravitystorm/openstreetmap-carto/pull/1249#issuecomment-71399247 weight of bin vs bench is strange.

We do not have practical guidelines yet, just general guidance.
This present issue on bench / bin can be a useful practical case to determine size / weight / priority / tiers of icons. First in a 'fight' bench vs bin, and also as in bench&bin vs other icons.

This is very related to @mkoniecz last comment
> moving these two icons to .amenity-low-priority to avoid blocking more important icons and labels

daganzdaanda:
> moving these two icons to .amenity-low-priority 
definitely!

math1985:
Rebased.

mkoniecz:
@math1985 Can you move these two icons to .amenity-low-priority to avoid blocking more important icons and labels?

math1985:
Sorry I forgot, done now.

mkoniecz:
It is hard to find place with fully mapped benches and waste baskets, but in such places rendering of these objects is too prominent. I would propose pushing benches to z19 and waste baskets to z20.

http://www.openstreetmap.org/#map=19/50.07081/19.91222

selection_002
selection_003


Also, is it possible to ensure than benches have priority over waste baskets?

http://www.openstreetmap.org/#map=19/50.06866/19.98983

selection_004
selection_005

pnorman:
tile.osm.org doesn't render z20. I'd be against any changes for zoom level 20, except if [continuing a size progression](https://github.com/gravitystorm/openstreetmap-carto/pull/978/files#diff-5d9237083037cf1d6d91c329a0ff61adR1971). These changes would only be visible to those running their own deployment of osm-carto.

We've used z17 for the cutoff at which everything renders for other features, and I'd prefer to do so for benches and bins. If we feel benches/bins are too strong, perhaps reducing the size? Saying z19 only feels like a cartographic cop-out.

mkoniecz:
> tile.osm.org doesn't render z20

Right, I forgot. Than I repeat my

> Maybe bin icon can be even smaller?

nebulon42:
I made a smaller version of waste basket: https://github.com/nebulon42/osmic/blob/master/outdoor/waste-basket-10.svg

math1985:
Rebased, and made the following changes:

* Render waste_basket smaller
* Render bench only from z19
* Give bench priority over waste_basket

This is again ready for review.

matkoniecz:
I would support merging. @math1985 - would it be OK for me to squash 3 commits from this PR into one before merging? It is case of "improve previous commit" rather than separation of independent changes.

math1985:
Yes, please squash. In general I think it's clearer if the reviewer squashes rather than the author of the PR, as it makes it easier for the reviewer to see what has changed.

matkoniecz:
Merge done via cli in https://github.com/gravitystorm/openstreetmap-carto/commit/376a1b4a230a12faff5eb6979f7822da3029a13f

PR #1250, opened 2015-01-23
Let yaml2mml.py remove trailing whitespace from project.mml

floscher:
Currently the yaml2mml.py conversion script adds [trailing](https://github.com/gravitystorm/openstreetmap-carto/blob/bc2eb949f28992a354b5ff04226780226b0f79df/project.mml#L2) [whitespaces](https://github.com/gravitystorm/openstreetmap-carto/blob/bc2eb949f28992a354b5ff04226780226b0f79df/project.mml#L5) [all](https://github.com/gravitystorm/openstreetmap-carto/blob/bc2eb949f28992a354b5ff04226780226b0f79df/project.mml#L6) [over](https://github.com/gravitystorm/openstreetmap-carto/blob/bc2eb949f28992a354b5ff04226780226b0f79df/project.mml#L7) [the](https://github.com/gravitystorm/openstreetmap-carto/blob/bc2eb949f28992a354b5ff04226780226b0f79df/project.mml#L8) [place](https://github.com/gravitystorm/openstreetmap-carto/blob/bc2eb949f28992a354b5ff04226780226b0f79df/project.mml#L9).
This makes diff-tools (e.g. [git-diff](http://git-scm.com/book/en/v2/Distributed-Git-Contributing-to-a-Project#Commit-Guidelines)) complain and it is [a known python bug](http://bugs.python.org/issue16333), that was fixed back in 2012.
This will probably cause problems when either a developer or Travis start using a version of python, that includes this bugfix. Then [Travis will complain, that project.mml is different from project.mml2](https://github.com/gravitystorm/openstreetmap-carto/blob/master/.travis.yml#L23).

This pull request solves this problem by stripping all trailing whitespace from project.mml everytime yaml2mml.py is run. So the output should be identical for both versions of python that include the mentioned bugfix and those that don't.

Personally I'd like to also add a Travis check for trailing whitespace or a script that removes trailing whitespace (or both) to ensure, that [other files also don't have trailing whitespace](https://github.com/gravitystorm/openstreetmap-carto/commit/bc2eb949f28992a354b5ff04226780226b0f79df#diff-f61408bc1ffb6636dbaf2bf9e38a4177R783). What is your opinion on that?


math1985:
I agree with the symptoms, but:

> Currently the yaml2mml.py conversion script adds trailing whitespaces all over the place.

The whitespace is not added by the yaml2mml.py script, but rather caused the fact that project.mml itself contains a lot of trailing white space itself.

math1985:
> Personally I'd like to also add a Travis check for trailing whitespace or a script that removes trailing whitespace (or both) to ensure, that other files also don't have trailing whitespace. What is your opinion on that?

Yes, I'd like a Travis check to ensure the files have no trailing whitespace.

mkoniecz:
> Travis check for trailing whitespace

A great idea.

pnorman:
note - when merging this, it should probably be done with ``--no-commit`` and run yaml2mml again.

The python bug suggests

```python
print(json.dumps({'4': 5, '6': 7}, sort_keys=True, indent=4, separators=(',', ': ')))
```
Would this work?

floscher:
> note - when merging this, it should probably be done with --no-commit and run `yaml2mml` again.

I already ran `yaml2mml.py`, otherwise [the Travis CI build](https://travis-ci.org/gravitystorm/openstreetmap-carto/builds/48051351) would not be green.

> Would this work?
Yes, that is more elegant. I updated my commit.

pnorman:
> I already ran yaml2mml.py, otherwise the Travis CI build would not be green.

Yes, but if there's any commits to project.yaml since you branched there will be a merge conflict in project.mml

PR #1251, opened 2015-01-23
Added script to detect trailing whitespaces

floscher:
The script prints the number of lines that contain trailing whitespace in files inside of a given directory.
Usage: e.g. `./scripts/check_trailing_whitespace` or `./check_trailing_whitespace ../`.
Currently it checks files that have the following file extensions: .mss, .mml, .style, .svg, .yaml, .yml, .md, .rb, .py, .sh, .gitignore and files without file extension.
Travis CI will execute this script for the root directory of the repository.

See #1250 for discussion regarding this topic.

When pulling you should use the `--no-commit`-option and check if the `check_trailing_whitespace`-script returns 0 before committing.


pnorman:
Can't git do this check? Also, are git commit hooks are more appropriate way to do this?

floscher:
> Can't git do this check? Also, are git commit hooks are more appropriate way to do this?

Yes, it can. I've created an [update-hook](https://gist.github.com/floscher/ad41bac476bdcc2dee2f), which can check, if someone introduces whitespace problems by pushing to the repository.

There is even a setting in the git-configuration (core.whitespace), that controls, which whitespace problems should be reported (see [the git-docs for possible values](http://www.git-scm.com/docs/git-config)).

The downside is, that git only complains once somebody tries to push the change to the server-repository (or does GitHub check if a hook would fail, when opening a pull request). You could turn on the default pre-commit-hook on the client side to produce errors when trying committing such changes, but that in turn has to be activated in every single client-side repository.

pnorman:
> The downside is, that git only complains once somebody tries to push the change to the server-repository

Yes, but that's sooner than travis building after a PR. Erroring on client-side commit would be ideal, but erroring on push or opening PR is better than erroring on travis build.

If we do have to do a manually triggered check of whitespace, then we should still look at having git do the actual check. `git diff --check` can probably do this.

I don't have an issue with whitespace checks for myself, but I'm worried that we might be increasing barriers to contribution. @gravitystorm, any thoughts?

floscher:
Closing, as it is not an important change and could discourage new contributors from contributing.

PR #1252, opened 2015-01-24
Clean up low-zoom road definitions

### LINKED ISSUE #400: z=12 railway tunnels loss of layering: rendered on top of everything
Possibly as a result of the commit in issue #395 ; railway tunnels at z=12 are rendered on top of everything (constrast with z=11 and z=13+).  Most obvious at Malmo City Tunnel, Folkestone Channel Tunnel/Shakespeare Tunnel, Berlin Tiergarten Tunnel:

 * http://osm.org/#map=12/55.6/13
 * http://osm.org/#map=12/51.1/1.3
 * http://osm.org/#map=12/52.5/13.4

math1985:
This merges code for low-zoom road definitions into the corresponding
high-zoom road definitions.

Data on z&lt;10 now comes from osm_planet_roads, data on z>=10 comes from
osm_planet_line. Chosen is for zoom level 10 as cut-off, because residential
is rendered from z10 and is not included in osm_planet_roads, so on z>=10
osm_planet_line needs to be queried anyway.

Advantages:
* This makes the zoomlevel cut-off for querying from the roads vs line
  table independent from the zoomlevel cut-off for the lowzoom rendering style,
instead of using one zoomlevel cut-off for both.
* This brings related code (i.e. code per feature) closer together in the file,
  which should make the code more easy to modify.

This is mainly a code cleanup, but it causes some small changes in rendering:
* Take into account layering of roads on z10 and z11.
* Fix the loss of layering problem of railway tunnels on z12 (resolves #400).
* Fix bug that prevented tunnel rendering style for railway on z11/12, and add
  tunnel style rendering for spur/siding/yard for z<13.
* Give railways on z<12 a round line-join.

This is an improvement of #555.


project.yaml
```diff
@@ -891,7 +891,9 @@ Layer:
     Datasource:
       <<: *osm2pgsql
       table: |2-
-              (SELECT way, COALESCE(('highway_' || (CASE WHEN substr(highway, length(highway)-3, 4) = 'link' THEN substr(highway, 0, length(highway)-4) ELSE highway end)), ('railway_' || (CASE WHEN (railway = 'rail' AND service IN ('spur', 'siding', 'yard')) THEN 'INT-spur-siding-yard' WHEN railway IN ('rail', 'tram', 'light_rail', 'funicular', 'narrow_gauge') THEN railway ELSE NULL END))) AS feature, tunnel
+              (SELECT way, COALESCE(('highway_' || (CASE WHEN substr(highway, length(highway)-3, 4) = 'link' THEN substr(highway, 0, length(highway)-4) ELSE highway end)), ('railway_' || (CASE WHEN (railway = 'rail' AND service IN ('spur', 'siding', 'yard')) THEN 'INT-spur-siding-yard' WHEN railway IN ('rail', 'tram', 'light_rail', 'funicular', 'narrow_gauge') THEN railway ELSE NULL END))) AS feature,
+                 CASE WHEN tunnel = 'yes' OR tunnel = 'building_passage' OR covered = 'yes' THEN 'yes' ELSE 'no' END AS int_tunnel,
```
pnorman: We've been using INT- elsewhere as a prefix

math1985: `int_tunnel` was already in use as well for the water tunnels. This should be made consistent some time, but that would require also editing the water features, which I don't want to do in this PR.

math1985:
This is ready for review/merge.

math1985:
Rebased.

PR #1254, opened 2015-01-24
move theatre icon to SVG

nebulon42:
Implements part of #1165.
Icon: [theatre-16.svg](https://github.com/nebulon42/osmic/blob/master/tourism/theatre-16.svg)

before
theatre_before

after
theatre_after


kocio-pl:
I like the current icon we have - could this new one have one mask in "negative" color, to make them appear more distinct? The diagonal positioning would be also nice, I think.

math1985:
> I like the current icon we have - could this new one have one mask in "negative" color, to make them appear more distinct? The diagonal positioning would be also nice, I think.

+1

mboeringa:
I'm OK with both versions, the new one works just as well to me, and it is similar enough to the old one in terms of the way the "theatre" aspect is conveyed, to be easily recognizable / familiar.

nebulon42:
Agreed, the inverse mask looks better. But I'm not in favour of tilting the masks. I have updated the preview.

mboeringa:
Looks good to me...

dieterdreist:
2015-01-24 17:54 GMT+01:00 kocio-pl <notifications@github.com>:

> I like the current icon we have - could this new one have one mask in
> "negative" color, to make them appear more distinct? The diagonal
> positioning would be also nice, I think.



+1
please note that "black and white" (similar to yin and yang) has additional
meaning (good vs. evil, the main topic in the ancient theatre) that now
gets almost completely lost in the current svg icon. (almost because
there's the tiny detail of smiling vs. bad mouth). Also the tilted masks
look a lot better (lighter, more vivid) than the rigid ones in the new
proposal.

PR #1256, opened 2015-01-25
move drinking water icon to SVG

nebulon42:
Implements part of #1165.
Icon: [drinking-water-16.svg](https://github.com/nebulon42/osmic/blob/master/outdoor/drinking-water-16.svg)

before
drinking_water_before

after
drinking_water_after


mboeringa:
Looks nice, the half filled cup is OK too.

PR #1259, opened 2015-01-26
move prison icon to SVG

nebulon42:
Implements part of #1165.
Icon: [prison-16.svg](https://github.com/nebulon42/osmic/blob/master/administration/prison-16.svg)

before
prison_before

after
prison_after


polarbearing:
Nice improvement, the old one looked a bit like an owl in a cage.

math1985:
The icon is improved a lot, but perhaps still not recognizable enough. Any chance of basing an icon on something like [1](http://zizaza.com/cache/small_wm/iconset/582181/582214/PNG/256/nasty/simple_prison_nasty_line_icon_png_prison_png_prison_icon.png) or [2](http://t1.gstatic.com/images?q=tbn:ANd9GcQym0gbmT_5FsrMjM09K3LvqONGNeJJpEaKq0pxr2AVEpNei2HhopLsQA)?

nebulon42:
@math1985 I think it will be hard to have the hands recognisable. Your second suggestion is quite similar to my proposal, though hands are absent. Could you explicate where you see lack of recognisability?

I see increasing the space between bars and the person as one possible improvement, but I'm not sure if that is necessary.

kocio-pl:
What I like in current version is that the person behind the bars is lighter, so it's easier for me to recognize both objects are different - example 2 from @math1985 also goes this way (but it's too much abstract for me).

daganzdaanda:
I like the new one already, it's much better than the owl! 

I kind of feel that the bars could be a little further from the head, and maybe then the body could be wider. Or maybe three bars, one of them in the middle in front of the person, could actually work? 
Maybe the window outline could also be stronger?

Rovastar:
I like the new one already, it's much better than the owl! 

Owl!! Surely it is a "jailbird". ;)

nebulon42:
I have tried some variations like increased space and a thicker frame, but nothing looked good. So I have to leave the proposal as it is.

pnorman:
:shipit:

math1985:
Is the change in tourism_viewpoint intentional? Might be better in a separate PR (or add least add it to the PR description).

nebulon42:
@math1985 This is caused by the collapsed view of the code, if you expand it to full then you see that the change is made near the prison rule.

math1985:
Oops, thanks.

Issue #1012, opened 2014-10-06
render more private POI icons lighter


HolgerJeromin:
Parking has amenity-points.mss#L172

```
[access != ''][access != 'yes'] {
   marker-opacity: 0.33; 
```

This works, as the marker is in SVG format. I think this could be extended. It is better to change the icon SVG format, but even PNGs can have point-opacity:
- [x] leisure_playground (opacity #1191, SVG icon: #1177 merged)
- [x] amenity_recycling (opacity #1191, SVG icon: #1175 merged)
- [ ] amenity_toilets (SVG icon: #1328 merged)
- [x] <s>natural_cave_entrance (for security reasons) (opacity #1262 btw the cave entrace icon is in another mss-file than its name)</s> (rejected)

less important
- [ ] tourism_picnic_site
- [ ] leisure_picnic_table
- [ ] amenity_drinking_water

and perhaps:
- [ ] tourism_museum (SVG icon: #1174 merged)
- [ ] historic_archaeological_site
- [ ] amenity_library (SVG icon: #1228 )

For parking the name has another text color (hard coded, and not defined btw) which has to be decided as well.
- [ ] A logic the other way round could be implemented for barrier nodes.

mkoniecz:
So toilets and archaeological site are now waiting for SVG icons.

natural=cave_entrance is problematic, as it is not clear what access on it means. It may declare whatever cave entrance is accessible or is it about cave? How should one tag cave entrance that is reachable, but further travel is impossible (door, bars etc)?

Maybe definition is hidden somewhere on OSM wiki, but as it is not intuitive big part of tags would ignore it anyway.

nebulon42:
IMO access on the cave entrance means the cave. Reachability has to be represented on ways towards the entrance. If there are none then everybody would have to judge for his/her own if it would be dangerous/impossible to reach the entrance.

HolgerJeromin:
Most "private" museums i found requires pre-registration. IMO a very valid reason to render them with opacity.

nebulon42:
There is now a icon proposal for library (see #1228).

pnorman:
> This works, as the marker is in SVG format. I think this could be extended to at least, but the icon has to be changed to SVG format:

There is no need to change the icons to SVG for opacity. See [point-opacity](https://github.com/mapbox/carto/blob/master/docs/latest.md#point-opacity-float).

SVG gains us two things over PNG

[...]

PR #1262, opened 2015-01-27
render cave symbols less opaque when access is restricted

tilmanb:
natural=cave_entrance is rendered less opaque when access is restricted. This is a partial fix for #1012 .

Visual comparison:
left: old, right: opaque 0.55
natural_cave_entrance_with_name | natural_cave_entrance_with_name_0 55
natural_cave_entrance | natural_cave_entrance_0 55

More opacity values (0.1 to 0.8):

natural_cave_entrance_with_name_opacity1-8

### DIFF
--- a/amenity-points.mss
+++ b/amenity-points.mss
@@ -19,7 +19,7 @@
 @standard-wrap-width: 30;
 
 .points {
-  [feature = 'tourism_alpine_hut'][zoom >= 1c3] {
+  [feature = 'tourism_alpine_hut'][zoom >= 13] {
     point-file: url('symbols/alpinehut.p.16.png');
     point-placement: interior;
   }
--- a/amenity-points.mss
+++ b/amenity-points.mss
@@ -552,7 +552,7 @@
     point-file: url('symbols/poi_cave.p.16.png');
     point-placement: interior;
     [access != ''][access != 'yes'] {
-      point-opacity: 0.55;
+      point-opacity: 0.4;
     }
   }
 
@@ -796,7 +796,7 @@
     text-wrap-width: @standard-wrap-width;
     text-placement: interior;
     [access != ''][access != 'yes'] {
-      text-opacity: 0.55;
+      text-opacity: 0.4;
     }
   }
 
--- a/amenity-points.mss
+++ b/amenity-points.mss
@@ -19,7 +19,7 @@
 @standard-wrap-width: 30;
 
 .points {
-  [feature = 'tourism_alpine_hut'][zoom >= 13] {
+  [feature = 'tourism_alpine_hut'][zoom >= 1c3] {
     point-file: url('symbols/alpinehut.p.16.png');
     point-placement: interior;
   }
@@ -551,6 +551,9 @@
   [feature = 'natural_cave_entrance'][zoom >= 15]::natural {
     point-file: url('symbols/poi_cave.p.16.png');
     point-placement: interior;
+    [access != ''][access != 'yes'] {
+      point-opacity: 0.55;
+    }
   }
 
   [feature = 'natural_spring'][zoom >= 14]::natural {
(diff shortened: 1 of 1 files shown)


mkoniecz:
I admit that I am not sure is it a good idea - as tagging in that case is not clear. [natural=cave_entrance, access=no] may be interpreted as lack of access to the cave tunnels or as cave entrance that may not be reached.

Can you provide before/after comparison?

HolgerJeromin:
thanks for your pull request!
i too would interprete access no as collapsed or something like that and private as reachable but not allowed to enter.
I am not sure if a non reachable cave should be opaque or not :-)

tilmanb:
As far as I understand, the tag natural=cave_entrance says ( on http://wiki.openstreetmap.org/wiki/Tag:natural%3Dcave_entrance ) on the access tag:
*"no" if bricked, locked or forbidden
"private" if locked and you can get the key*

and on using barrier

*gate/wall/etc.*

So it is unclear, whether acces means physical access or legal access. I think the topic of reachability is another one, and does not need to be encoded in the rendering of the cave entrance, but on the tracks/paths (if any) leading towards the cave entrance.

For the technical part: I will provide visual comparisons this evening. To squash those two commits into one I should open a new branch and re-perform the changes there, and then open another pull request?

pnorman:
You can use this pull request - just push more commits to the branch

tilmanb:
Can't get the table to show correctly when editing the first post, so here in table style:

Old | New (Opaque 0.55)
------------ | -------------
natural_cave_entrance_with_name | natural_cave_entrance_with_name_0 55
natural_cave_entrance | natural_cave_entrance_0 55

HolgerJeromin:
oh, i did not checked the wiki page for special access definition. So rendering no and private with opacity seems ok for me. In both cases you need preparation before entering.

could you try rendering with 0.3? on my phone 0.5 is not very different if you have nothing to compare.

tilmanb:
see updated first post for comparison of opacity values (text and marker) from 0.1 to 0.8 on forest.

HolgerJeromin:
Thank you.
If the feature has no name 0.55 is not very clear (on my mobile and desktop). I would propose 0.4 for opacity. Other opinions?

tilmanb:
Yes, on closer inspection I am in favor of using 0.4. amenity=parking uses 0.33, but as caves are often drawn on darker backgrounds compared to parkings, I propose using something more visible. So 0.4 seems good to me.

tilmanb:
@mkoniecz have you had a chance to look at the before and after comparisons? Note that 0.4 seems to be the best opacity value (as seen in the opacity row in the updated first post).
Also, the wiki page is now updated and further clarified regarding access and barrier.
I can also provide renderings with 0.4 opacity (or the complete opacity range) on different backgrounds, if needed.

tilmanb:
Here are more opacity values on different backgrounds. Again, I propose 0.4.
caveentrance_opacityrow_4bg

pnorman:
.4 looks good, as would .3 on the icon and and .4 or higher on the text.

floscher:
Just a sidenote:
> render cave symbols opaque when access is restricted

You want to render it **less** opaque. It took me quite some time to realize that you actually mean it the other way around. :wink:

tilmanb:
Is there anything left to do on my part? I am unsure on how to "squash two commits into one" as requested by mkoniecz without using another branch.

math1985:
As far as I'm concerned, this is too specialistic for the main map. Before going into a cave, you'd need to check if you need specialistic equipment etc. anyway. So I don't think showing the access restrictions of caves is very useful.

HolgerJeromin:
we had a goverment in germany deleting cave entrance osm points. Thats why i proposed this change on the main map.

math1985:
Thanks for the additional information, I wasn't aware of that.

pnorman:
I'm not sure why, but I'm not finding that the opacity indicates access as clearly as it does for parking icons.

math1985:
I don't think a cave icon on a map intuitively implies that you can (legally and/or safely) enter the cave.

I also agree with @pnorman that the opacity does not make its meaning very clear in this case.

So I'd propose to not merge this.

mkoniecz:
@tilmanb Thanks for your work, but unfortunately this pull request will not be merged.

PR #1266, opened 2015-01-28
add missing landcovers to text-poly-low-zoom

imagico:
Adds missing labels for glaciers and bare ground land covers for low zooms <10.  This is purely a fix for a coding error - styling already has the correct settings, just missing in project.yaml/mml.

In general the separate layers are prone to this kind of mistake, would probably be good to prominently mention this in CONTRIBUTING.md.


math1985:
Merged - although I think I would prefer to start rendering scree etc. from z10 rather than from z9, like most other landcover.

imagico:
I don't think there is currently any scree or bare_rock areas with name tag making the size limit at z=9.  natural=sand there are probably a few like http://www.openstreetmap.org/way/249396361 - and this could IMO well be shown already at z=8.  In general the subtle colors of scree, bare_rock and sand make them less disturbing than wood/forest at these low levels.  But if there are actually cases where this leads to bad results this can be reconsidered.

PR #1270, opened 2015-01-30
Improve widths aeroway

math1985:
Widths on z11/12/13/14/15/16:

* Runway old: 2/4/7/18/18/18
* Runway new: 2/4/6/12/18/24
* Taxiway old: 1/1/1/1/6/6
* Taxiway new: 1/1/2/4/6/8

### DIFF
--- a/roads.mss
+++ b/roads.mss
@@ -519,7 +519,9 @@
     [feature = 'aeroway_runway'] {
       .bridges-casing {
         [zoom >= 14] {
-          line-width: 19;
+          line-width: 13;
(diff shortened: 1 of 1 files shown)


imagico:
This could also well use scale factor adjustment as explained in #978

compare:

http://www.openstreetmap.org/#map=14/69.6837/18.9272
http://www.openstreetmap.org/#map=14/1.3559/103.9945

math1985:
@gravitystorm @pnorman What do you think?

mkoniecz:
Thanks!

PR #1271, opened 2015-01-30
Also render junction names on areas

### LINKED ISSUE #1253: Render junction=yes on areas

math1985:
This resolves #1253.

### DIFF
--- a/project.yaml
+++ b/project.yaml
@@ -1294,9 +1294,13 @@ Layer:
     Datasource:
       <<: *osm2pgsql
       table: |2-
-             (SELECT way, highway, junction, ref, name
-              FROM planet_osm_point
-              WHERE highway = 'motorway_junction' OR highway = 'traffic_signals' OR junction = 'yes'
+             (  SELECT way, highway, junction, ref, name
+                FROM planet_osm_point
+                WHERE highway = 'motorway_junction' OR highway = 'traffic_signals' OR junction = 'yes'
+              UNION
+                SELECT ST_BuildArea(way), highway, junction, ref, name
+                FROM planet_osm_line
+                WHERE highway = 'motorway_junction' OR highway = 'traffic_signals' OR junction = 'yes'
              ) AS junctions
     advanced: {}
   - id: "roads-text-ref"
--- a/project.mml
+++ b/project.mml
@@ -1354,7 +1354,7 @@
       "srs": "+proj=merc +a=6378137 +b=6378137 +lat_ts=0.0 +lon_0=0.0 +x_0=0.0 +y_0=0.0 +k=1.0 +units=m +nadgrids=@null +wktext +no_defs +over",
       "Datasource": {
         "extent": "-20037508,-20037508,20037508,20037508",
-        "table": "     (SELECT way, highway, junction, ref, name\n      FROM planet_osm_point\n      WHERE highway = 'motorway_junction' OR highway = 'traffic_signals' OR junction = 'yes'\n     ) AS junctions",
+        "table": "     (  SELECT way, highway, junction, ref, name\n        FROM planet_osm_point\n        WHERE highway = 'motorway_junction' OR highway = 'traffic_signals' OR junction = 'yes'\n      UNION\n        SELECT ST_BuildArea(way), highway, junction, ref, name\n        FROM planet_osm_line\n        WHERE highway = 'motorway_junction' OR highway = 'traffic_signals' OR junction = 'yes'\n     ) AS junctions",
         "geometry_field": "way",
         "type": "postgis",
         "key_field": "",
--- a/roads.mss
+++ b/roads.mss
@@ -1842,6 +1842,7 @@
       text-halo-radius: 1;
       text-wrap-width: 30;
       text-min-distance: 2;
+      text-placement: interior;
       [zoom >= 14] {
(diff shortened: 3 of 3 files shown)


math1985:
@pnorman This explicitly converts LINESTRINGs into POLYGONs. Do you think this code makes sense? It seems to work, even if there are LINESTRINGs in the database that are not POLYGONs (they get ignored).

project.yaml
```diff
@@ -1294,9 +1294,13 @@ Layer:
...
-             (SELECT way, highway, junction, ref, name
-              FROM planet_osm_point
-              WHERE highway = 'motorway_junction' OR highway = 'traffic_signals' OR junction = 'yes'
+             (  SELECT way, highway, junction, ref, name
+                FROM planet_osm_point
+                WHERE highway = 'motorway_junction' OR highway = 'traffic_signals' OR junction = 'yes'
+              UNION
+                SELECT ST_BuildArea(way), highway, junction, ref, name
+                FROM planet_osm_line
+                WHERE highway = 'motorway_junction' OR highway = 'traffic_signals' OR junction = 'yes'
```
pnorman: if using this you need a you need a `!bbox! && way` condition in here

project.yaml
```diff
@@ -1294,9 +1294,13 @@ Layer:
     Datasource:
       <<: *osm2pgsql
       table: |2-
-             (SELECT way, highway, junction, ref, name
-              FROM planet_osm_point
-              WHERE highway = 'motorway_junction' OR highway = 'traffic_signals' OR junction = 'yes'
+             (  SELECT way, highway, junction, ref, name
```
pnorman: two stray spaces before the `SELECT`

pnorman:
You'll want to check `ST_IsClosed` if using `ST_BuildArea`.

@gravitystorm the `ST_IsClosed` on lines table is technically sound, but is doing this an approach we **want** to take?

project.yaml
sommerluk: Maybe it’s better not to add highway=motorway_junction and highway=traffic_signals here. Both are only defined for nodes, not for areas. (Or is there a technical reason to keep this?)

math1985: No technical reason, we could drop it. Are you sure highway=traffic_signals is not defined on junctions, with the recent changes to the scheme?

gravitystorm:
@pnorman as I understand it, we're hacking around our .style being incorrect, and all the SQL cleverness is to avoid changing it and therefore avoid people having to reload their databases.

So we need to weigh up whether this is important enough to carry the baggage of complexity, or whether we can wait for the next change in the .style file. We don't have a cadence for that yet (i.e. we've never done it).

I'd prefer to wait and do it properly, but without a cadence that's asking to wait forever. So I say we take the complexity, and remove the complexity as soon as we can.

All this presupposes that we're only talking about complexity of the code, and that there's no particular performance issue.

math1985:
@gravitystorm @pnorman As far as I know, this would not only require a change to the style file, but also a change to the osm2pgsql code that allows it to make the polygon/line decision based on the value instead of the key only. Can such a change be expected in the near future?

gravitystorm:
If that's something we need, we should make sure there's a ticket for it upstream.

project.yaml
sommerluk: Yes. Finally it was only about junction=yes, leaving complex traffic signal systems for later.

(Maybe I’ll try to re-launch discussion about the traffic signal systems, but I would not wait for this. Currently, I would drop on areas both, highway=motorway_junction and highway=traffic_signals)

pnorman:
> Can such a change be expected in the near future?

As an osm2pgsql maintainer, I'd ask if Lua tag tranforms are an option for your needs.

math1985:
> As an osm2pgsql maintainer, I'd ask if Lua tag tranforms are an option for your needs.

I'm not sure what Lua tag transform can do exactly - could we create a lua file that makes sure that closed ways with highway=junction are treated like areas and closed ways with other highway values are treated as lines?

Also, I remember last time Lua tag transforms came up, somebody warned that this might still be buggy - are they stable now?

pnorman:
See https://github.com/openstreetmap/osm2pgsql/blob/master/docs/lua.md

https://github.com/openstreetmap/osm2pgsql/blob/master/style.lua is an almost replacement for the C tag processing.

I consider the pgsql backend and Lua transforms production ready - and know those like @systemed who use it in production. Obviously you want to test any transform you write.

I've been working with the multi backend and Lua transforms, which are not production ready, but the example multi backend scripts do show some other cases of pre-processing that can be useful. For example, yes/no/unset values can be [normalized so you can use boolean columns](https://github.com/pnorman/osm2pgsql/blob/1278df42534baeff530f69f0a4aa7ffe60628ac7/multi.lua#L54-L76).

My main objections to adding logic to allow some values for keys to be areas and other values for the same keys to be linear is the complexity it adds to the .style files, which already almost no one fully understands.

math1985:
After having looked at the lua transforms, I think that's the way to go, especially since treating junction=yes on polygons as areas is something that should be style independent.

We could add something like this to the end of `filter_tags_way`:

``
if (keyvalues["junction"] == "yes") then
  poly = 1;
end
``
Or alternatively something more complex with an array of values that are always be treated as area.

I will close this PR.

PR #1273, opened 2015-01-30
Make halos transparent (alpha 0.6)

math1985:


math1985:
Could anybody have a look at this? It's already open for four weeks, would be nice to have it merged at some point.

PR #1274, opened 2015-01-31
Render tourism=zoo like tourism=theme_park

### LINKED ISSUE #778: Render tourism=zoo as half transparent overlay
The following issue has been moved over from [trac](https://trac.openstreetmap.org/ticket/3962):

Rendering tourism=zoo as half transparent "overlay" (like nature reserves are rendered currently) would be better. Currently landuse and nature areas are drawn over zoo area and in places where the entire zoo area is covered with landuse/natural areas there is no way to say that this area is a zoo area.

Example (not jet covered fully with landuse tags but the area covered with forest shows the problem): http://www.openstreetmap.org/?lat=59.42285&lon=24.66418&zoom=15&layers=M

math1985:
This resolves #778.


mkoniecz:
I think that zoos should not be restricted to z13+. Typical ones are big enough to display at z12, even z11 and z10 may be displayable - see for an example http://www.openstreetmap.org/?mlat=43.8246&mlon=-79.1867#map=11/43.8246/-79.1867.

math1985:
Dropped minimum zoom level to z10.

Rebased.

math1985:
Rebased.

This is still waiting for review.

HolgerJeromin:
I think symbols/zoo.png can be deleted now.

pnorman:
fixed in 7461ed5

PR #1275, opened 2015-01-31
Decrease polygon-gamma for large water areas

### LINKED ISSUE #671: Gap between water and sea
There's a small white line between a polygon tagged natural=water and the 'sea', see for example the Venetian Lagoon
https://www.openstreetmap.org/#map=17/45.33119/12.34303

math1985:
This decreases polygon-gamma to 0.3 for large water areas.
This likely resolves #671.

### DIFF
--- a/water.mss
+++ b/water.mss
@@ -24,7 +24,7 @@
         polygon-gamma: 0.75;
       }
       [way_pixels >= 64] {
-        polygon-gamma: 0.6;
+        polygon-gamma: 0.3;
       }
     }
   }
@@ -35,7 +35,7 @@
       polygon-gamma: 0.75;
     }
     [way_pixels >= 64] {
-      polygon-gamma: 0.6;
+      polygon-gamma: 0.3;
     }
   }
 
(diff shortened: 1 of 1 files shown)


pnorman:
How bad is the impact on large complex water edges?

math1985:
I don't notice any problems with limited tests.

math1985:
This is ready for merge+review.

pnorman:
I'd like to test on a large waterbody with a complex edge. Probably something in Canada - we have a lot of those.

pnorman:
0.3
image

0.6
image

1.0
image

I can see the difference unaided :(

I'd rather have #671 than mis-state the water area like this.

math1985:
Self-assigned.

pnorman:
moving from land polygons to ocean polygons might help a bit with this. ultimately, it's an upstream issue that we're trying to work around.

imagico:
I would not put too much hope in ocean polygons here - as i see it the only way this would help is by manipulating gamma on the shapefile layer and this is a bad idea with the many small islands.

math1985:
Would there be any point in only setting low polygon-gamma for areas with huge way_pixels?

pnorman:
No, not really.

math1985:
In that case I think there's nothing else we can do then marking this as wontfix.

PR #1281, opened 2015-02-01
move bicycle shop icon to SVG

nebulon42:
Implements part of #1165.
Icon: [nebulon42/osmic/bicycle-16.svg](https://github.com/nebulon42/osmic/blob/master/shop/bicycle-16.svg)

before
bicycle_before

after
bicycle-after

PR #1282, opened 2015-02-01
move bicycle rental icon to SVG

nebulon42:
Implements part of #1165.
Icon: [nebulon42/osmic/rental-bicycle-16.svg](https://github.com/nebulon42/osmic/blob/master/transport/rental-bicycle-16.svg)

before
bike_rental_before

after
bike_rental_after


mkoniecz:
Unfortunately, with new icon It is clear that it is bicycle. Probably because it is smaller.

nebulon42:
I guess you mean that it's not clear. With what do you think it is likely to be confused with?

mkoniecz:
Sorry, I missed negation.

With old icon mi first impression is that it is key and bicycle. With new - key and some blob and I need to figure out what it is supposed to be. Note that it is heavily influenced by fact that I am looking at new icon for the first time and old is for me well known.

mkoniecz:
Test on pseudorandom person without exposure to the style: both are recognized as bicycle, old one is preferred as bicycle is larger.

nebulon42:
Thanks for testing. As there is no apparent reason (except legibility maybe) why some icons are 20px and some are 16px I suggest to stick to the 16px canvas size (as I so far already did for prison and theatre if I remember correctly).

math1985:
> Test on pseudorandom person without exposure to the style: both are recognized as bicycle, old one is preferred as bicycle is larger.

Same results for second pseudorandom person.

I don't have a strong opinion on whether to favour legibility or consistency in size.

nebulon42:
I see no problem with legibility here, both tests confirmed it to be seen as bicycle. As far as my icons are concerned, they have all 16px canvas size. But if you want to keep the larger icon this is fine too.

kocio-pl:
For me the new one is a bit too small, but if that way we can achieve better visual consistency, I can live with it.

nebulon42:
Updated after update of #1281.

math1985:
@gravitystorm This needs a strategic decision: do we go for icons that are all the same size at the cost of legibility? If not, how do we decide which icons we give a large size?

gravitystorm:
Icons certainly don't need to be the same size, especially when they have widely varying "weights" i.e. number of coloured pixels. There is, of course, benefit to having them of both similar size and weight.

Larger icons would be for things which are more important, or are more useful for self-orientation when looking at the map. Smaller icons would be good for things that are repeated many times in the one map view.

I don't have any particular comments on this proposal, but I hope the above guidance is useful in general.

math1985:
@gravitystorm Yes, such high-level guidance is very useful.

Because @gravitystorm didn't mention allowing finer detail as a reason to accept to use larger icons, and because bicycle rental is not more important than other POI, and also to prevent inflation in icon sizes, I'd be inclined to accept this proposal.

daganzdaanda:
I'd like to have all icons more or less the same "optical" size. This might mean that a very dense (=lots of filled pixels) icon could be smaller in real pixels, but carry the same optical impact.

Vector icons are great in that they can get scaled better (in theory at least), so that we could one day decide to show some more important icons larger.
Scaling a 16px icon up will probably give better results than scaling something larger down. Of course, ideally we'd have pixel-aligned icons for many sizes... 

About the bike rental icon, let's take it like that. I don't believe any of my versions in #1281 would look better at that size.

mkoniecz:
And thanks for your work!

math1985:
Nice work!

PR #1283, opened 2015-02-01
move car sharing icon to SVG

nebulon42:
Implements part of #1165.
Icon: [nebulon42/osmic/shared-car-16.svg](https://github.com/nebulon42/osmic/blob/master/transport/shared-car-16.svg)

before
car_sharing_before

after
car_sharing_after


math1985:
Tester prefers first icon, because it's not clear that there are people in the new icon.

nebulon42:
I can offer this as quick variant: shared-car-16

mboeringa:
I think I also prefer the "old" version. As  @math1985 pointed out, the symbols for people don't work as such. I get the feeling they are actually cars, in a kind of perspective drawing similar to road-signs warning for traffic jams.

math1985:
Is there a way to make the bodies of the people taller? Maybe add one or a half pixel to the top and bottom of the bodies?

math1985:
> Is there a way to make the bodies of the people taller? Maybe add one or a half pixel to the top and bottom of the bodies?

@nebulon42 Did you have the time to look at this?

nebulon42:
Not yet, sorry.

nebulon42:
I don't see the possibility of making the bodies higher, only of making the car lower. But this would mean breaking consistency with other car related icons.

Another thing which occured to me: The distinction between car rental and car sharing is fuzzy IMO. What was the rationale for having a distinct icon for that?

For bike rental and bike sharing (e.g. public city bikes that can be rented and brought back at/to several locations) we use the same icon. I think the situation with cars re distinction of renting/sharing is not so much different.

brycenesbitt:
@nebulon42  car sharing and car rental are highly distinct.  Car sharing are unattended locations: if you show up unannounced you can't do anything other than look at the vehicles.

Car rental companies on the other hand, will gladly take your credit card and driver's licence, and let you hire a car.

brycenesbitt:
I prefer the old icon.  The people are the crucial graphic element, not the car, for the car sharing case.

nebulon42:
I can offer the two variants (original post and https://github.com/gravitystorm/openstreetmap-carto/pull/1283#issuecomment-72525584). Please tell me if you consider to merge either version, otherwise I will close this.

mboeringa:
Here is another variation of @nebulon42's last version for comparison, with lengthened bodies. Actually quite some changes here done, so don't know if it can work in SVG, this is just bitmap editing...

car_sharing

kocio-pl:
For me it would work.

nebulon42:
@mboeringa Thanks for trying this out. The compressed car is not really an option for me.

brycenesbitt:
@mboeringa the compressed car looks much better, and helps get the "sharing" part across.  Car sharing and car rental are very different.

math1985:
What about @mboeringa 's version but with the car 1 or 2px more narrow? How would that look?

nebulon42:
That might help, but I have to aim for consistency. If I make the car less tall I will have to change five other icons too. Maybe somebody will come up with another more suitable SVG proposal. Mine is available in the Osmic repository if you want to use it.

mboeringa:
> If I make the car less tall I will have to change five other icons too.

Yes, this is something I realized recently too. My adjusted proposal conflicts with for example the recently accepted  and merged **move car repair icon to SVG** #1377. As you say, for consistency, it is probably best to either stick with the original proposal and car icon, or change all other car related icons too that may already have been moved to SVG (I am neutral here on this, and leave it to others to judge what they want).

brycenesbitt:
Consistency is getting in the way of function here.  This is not a "facility for cars" needing the same look as other car facilities.  It's a type of transportation that happens to use cars.

mboeringa:
> **Consistency is getting in the way of function here**. This is not a "facility for cars" needing the same look as other car facilities. It's a type of transportation that happens to use cars.

Not necesarily. Personally, I can live with two icons, but I think it is up to others to decide or give their opinion. In addition, there is the option of adjusting all other icons for consistency, but that requires extra labour. I think we need to give credit to @nebulon42 for all the work in creating these icons and posting them here. I can understand he may be a bit reluctant to adjust all of these, if not for anything else, just because of the time involved. Maybe someone else can do it who has experience with the creation of these SVG icons.

PR #1297, opened 2015-02-08
Restore railways on z12 and roads and aeroways on z10/z11

### LINKED ISSUE #1296: Missing railways at z12
There is no railways at zoom 12 with the latest update 
http://a.tile.openstreetmap.org/11/1229/694.png
http://c.tile.openstreetmap.org/12/2459/1389.png
http://a.tile.openstreetmap.org/13/4919/2779.png

math1985:
This resolves #1296.

### DIFF
--- a/project.mml
+++ b/project.mml
@@ -943,7 +943,7 @@
         85.05112877980659
       ],
       "properties": {
-        "maxzoom": 11,
+        "maxzoom": 12,
         "minzoom": 5
       },
       "advanced": {}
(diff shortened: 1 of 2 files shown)


math1985:
@pnorman Could you review this?

math1985:
For information, #1252 aims to fix the code mess that triggered this bug in the first place.

math1985:
@gravitystorm  GIven the impact of this, I think we should consider releasing a v2.28.1 with just this commit.

PR #1304, opened 2015-02-09
SVG icons for mast and add conditional rendering based on tower:type

### LINKED ISSUE #1225: Masts are not always for communication
The tag man_made=mast only says something about the physical look of the structure, not it's function. See http://wiki.openstreetmap.org/wiki/Tag:man_made%3Dmast

Next to communications, it can also be used for lightening (of football stadions f.e.), or for sports like popinjay archery. 

Currently, the icon communications.p.20.png is used for all man_made=mast objects, that icon has a visualisation of "broadcasting" next to the mast view.

According to the wiki, it's advised to add tower:type=communication if the mast is used for communication purposes. 

So I propose to use a version without broadcasting for a simple man_made=mast, and a version with broadcasting when tower:type=communication is added.

I made a simple icon of a mast, based on the communications icon, but where I just coloured the broadcasting stuff away. I hope it's good enough.
mast

nebulon42:
Adds SVG icons for `man_made=mast` and thus implements part of #1165.
Removes a duplicate rendering rule for mast that was present in amenity-points.

Adds conditional rendering for masts based on `tower:type`. Only if `tower:type=communication` is present the mast is visualised with communication waves. This resolves #1225.

`tower:type` should be present in the database. As my setup uses hstore this should be double-checked.

Previews:

with `tower:type=communications`
before
comm_before

after
comm_after

without `tower:type=communications`
before
no_comm_before

after
no_comm_after


daganzdaanda:
I'm not too keen on the "A" shape. I think it's a rare exception for masts to look like that. Of course, that kind of icon has been used historically on some maps, maybe that would be a reason to use that style.

A mast should be a lot higher than wide. I'd even argue that a single 1.5px line with a dot on top could be clearer. Similar to 
http://wiki.openstreetmap.org/wiki/File:Communications_tower.svg
Or a line with a tiny base at right angles, and no dot at top.

Another point: When the "waves" are not there, it looks like an oil drilling rig. Which would maybe better be tagged as a man_made=tower?

I can't think of many masts that are used for something else than communication.  @sanderd17 wrote they "can also be used for lightening (of football stadions f.e.), or for sports like popinjay archery." Also man_made=maypole and man_made=flagpole are used already, they could be a subtype of mast. These all look more like a single line, not like an A or the Eiffel tower.

sanderd17:
In my head, a pole is a round feature, while a mast has diagonal connections to have a bigger strength and be able to go higher.

And about how often you get what type, well, that depends entirely on your region. Popinjay archery is very popular in Flanders. See a map of all recognised archery clubs in Flanders: http://www.bloso.be/SPORTDATABANK/SPORTCLUBSPERSPORTTAK/Pages/Boogschietclubs.aspx most of those clubs have one or two masts for popinjay archery (sadly, not a lot are mapped in osm due to not being visible from aerial imagery, and it's never next to a road - nobody wants arrows on his head - but I hope that a better rendering would lead to more mappers).

However, I agree that it won't hurt if the icon looks a bit taller than currently. But thanks a lot for working on it.

mkoniecz:
I prefer old mast icon.

nebulon42:
Please let's not get into philosophical discussions about masts. For your reference I made you a not A-shaped mast icon: mast The information content of this icon is close to zero.

As the resistance level against new icons seems to be increasing I have to think about if I will continue like that.

math1985:
Tester prefers the old icon.

> As the resistance level against new icons seems to be increasing I have to think about if I will continue like that.

I'd be happy to move to the crisper icons, and I'm very glad with all time and effort you're putting in the design of the icon. However, I'm hesitant to replace icons if testers (who have not seen either of the icons before) indicate that they prefer the old icons. It also seems that many of the old icons are in fact quite good, so it might be hard to improve on them. 

Maybe it would be an idea to stay a bit closer to the old icons, accepting some more blurred pixels?

Of course, we should take into account that in many cases we are moving from a 20px icon to a 16px icon. That inevitable forces us to simplify the icons.

@all Would it be an idea to keep some of the icons at 20px?

daganzdaanda:
@nebulon42 You are doing very good work. Please don't take it personal when I criticise an icon idea. I do agree that the switch to svg is necessary, and that it's a good opportunity to think about the icons. 

As for the "I" instead of an "A", yes that wouldn't work very well without radio waves ;-)
That's why I thought of a short "foot" or shadow line like an L, similar to the way some maps abstract trees as open triangles and circles with a shadow line. I guess it would still be very small and quiet on a busy map, but that's OK with me. 

I guess the difficulty is that anything with radio waves is easy enough to understand -- also, we have seen those icons before. But I haven't seen an icon for non-communication masts before, so I need to learn the new symbol, whatever it is.

sanderd17:
Although I like masts (wouldn't have started one of the issues otherwise), I don't think they belong to the group of "important" icons for a general map style, so I guess they should be in the smallest size category (even if it's difficult to create an icon for that size).

And again, thanks a lot for working on it, the things you have now are already way better than what I would be able to make.

daganzdaanda:
A quick test:
mmm-radio
mmm-nonradio
These are 16x16 px badly done with MS-paint. It's not the exact same mast in both.

I guess it shows that without radio waves, nobody will know what it's supposed to be... So maybe the Eiffel tower shape is indeed the best, it's unique enough that it may be remembered even without the waves.

math1985:
Another issue is that `tower:type=communication` is used quite rarely, and is likely often missing from communication masts.

HolgerJeromin:
We could add the waves if it is communication or empty.
But it is probably better to promote the sub Tag with the proposed rendering.

nebulon42:
I'm not taking critique personally, but everytime the discussion moves near a "this proposal is bad, next one please" I will voice my discomfort with it. I have always found concrete ideas/proposals for improvement very helpful. For example at #1281 or also here at https://github.com/gravitystorm/openstreetmap-carto/pull/1304#issuecomment-73604300. Lack of such concrete ideas (maybe because it's difficult to come up with some) as in #1283 can be a bit frustrating sometimes.

> Tester prefers the old icon.

If the tester has a neutral opinion (as you wrote: seen both icons for the first time), this is perfectly valid. Thanks for testing. I might have to move towards such an approach myself, but I was hesitant up until now to do that because coming up with new icons was enough work already.

> Would it be an idea to keep some of the icons at 20px?

I'm more in favour of consistency.

> Maybe it would be an idea to stay a bit closer to the old icons, accepting some more blurred pixels?

The new icons are not deliberately different from the old ones. The issue with the existing mast icon for me is that the radio waves are not visible/distinct enough and I wanted to emphasise the representation of the mast. Maybe I should start giving rationales for my icon design decisions? Possibly, that could provide a starting point for discussion.

I'm also fine with some old icons being kept, but I have quite a strong opinion against me "converting" the old ones to SVG. I will not do that.

> Another issue is that `tower:type=communication` is used quite rarely, and is likely often missing from communication masts.

This was a deliberate decision for me to encourage correct tagging. Decisions like that have often occured here already. But if you think distinction between masts are not that important that should be discussed too, maybe in #1225.

@daganzdaanda These are good looking proposals and helpful for me, thank you. Maybe this will work, I will give it a try. Ideas/comments that are visualised like that are most helpful for me. They don't need to look good, quick sketches are enough. They should work at small scales though.

kocio-pl:
@nebulon42: I appreciate your systematic efforts to standardize and refresh the icon set on the main OSM map. I think it's great somebody do this on such a big scale and so fast! It's obvious that people are used to what they see for a long time, but I see this transformation as necessary in the long run, even if I like some older icons more. Theres also a lot new icons missing and I hope there will be more of them.

imagico:
The problem here is that design has a lot to do with personal taste.  If i for example find an icon you created not so good it is not usually possible to shape this dislike into a suggestion how to improve it - apart from maybe 'make it more like the old icon' which - as you said - is not really that helpful.

We probably all agree that the aim of icon design is to create symbols that are appealing and well understandable for the users of the map.  With this in mind it might be more helpful for you if you have feedback of the most basic form (i.e. like/dislike) from a larger number of people and reserve discussion here for more substantial critique.

Specific suggestions from me here:  I think for communication masts the 'spherical shells illustrating radio waves from the top of a simple mast depiction' has a long tradition in map design and should be used.  In your draft the spherical shells are not that rounded which might make it difficult to identify them as radio waves.  But more importantly the source of the radio waves, i.e. the top of the tower, seems missing.

I have no problem with the A-shape of the tower but this would acually be implying man_made=tower and not man_made=mast.

And i have troubles imagining a man_made=mast for other purposes than communication which makes it difficult to form an opinion on this variant.

In general i think a lot of your new icons work very nicely, like fuel, fire, embassy, camping.  Others i quite strongly dislike (like the banking stuff).  But that's life...

al--:
Currently man_made=mast isn't rendered at all, with or without tower:type=communication. Bug?

And I'd prefer the old mast icon. BTS masts often are narrow masts made of steel, concrete or else.

/al

lest69:
> Currently man_made=mast isn't rendered at all, with or without tower:type=communication.

It is rendered. Here's an example without tower:type: http://www.openstreetmap.org/node/1866943739
...and one with tower:type=communication: http://www.openstreetmap.org/node/2975227644

nebulon42:
Thanks for all comments. As there seems to be currently no consensus to whether `tower:type=communication` should be used for conditional rendering I'm closing this. I will try to improve the mast icon in a separate PR.

PR #1311, opened 2015-02-14
add icon for shop=books

nebulon42:
Preview:
books


math1985:
Tester recognizes icon as books.

mkoniecz:
Tester recognizes icon as books. Maybe it would be a good idea to have the same icon in two colours for shop=books and library?

nebulon42:
This could work as shops have a quite distinct colour. But if we succeed in creating two different recognisable icons this would be better. See also https://github.com/gravitystorm/openstreetmap-carto/pull/1228#issuecomment-74369560 for remaining issues I have with an updated version of the library icon.

mkoniecz:
@nebulon42 It would be better to avoid reordering names in defition of shop name.

On topic of adding new icon - I suppport it, bookstores are common shop type. Currently most mappers in Poland specify that shop is bookstore, as otherwise in default rendering it is impossible to guess shop type. Separate icon should fix the biggest source of this poor tagging.

nebulon42:
> It would be better to avoid reordering names in defition of shop name.

What issues do you see with that? I thought it would be better to have it ordered alphabetically for quick overview.

mkoniecz:
> What issues do you see with that? I thought it would be better to have it ordered alphabetically for quick overview.

Maybe, but it would be better as a separate commit - or even better a separate PR. This way it is easier to check and ensure that nothing goes wrong.

daganzdaanda:
Nice icon! 
I think the transparency is really needed here, you couldn't make the "pages" solid, and you couldn't leave them empty.

nebulon42:
removed reordering of shop name definitions.
@daganzdaanda If you look at the icon it is not really transparent, but looks like it at this size. It works quite well, though.

math1985:
@gravitystorm This also needs a strategic decision. Do we want to add more icons for shoptypes we currently only render as dot? Personally I support more icons as long as they are rezognizable, but I know you are more hesitant on this aspect.

matkoniecz:
@gravitystorm @pnorman What is your opinion? Is anybody opposed to adding more icons for shops?

Currently I am planning to merge it in a near future.

kocio-pl:
I would love to add special icon for every possible type - visual hints work very good. Regarding this icon:  @daganzdaanda is right about making it more solid, but given the choice solidness or just a dot, I would take the first one.

math1985:
Nobody opposed, so as far as I'm concerned we merge this.

nebulon42:
Updated.

pnorman:
I'm iffy on adding more icons. Slightly less so on shop icons.

kocio-pl:
Why? If they aren't visible too early, I see no problem with more icons. As a matter of fact in micromapping we lack many of them - including shops, but if we care only for them, we would be not much better than Google Maps with their clear commercial attitude.

matkoniecz:
@nebulon42 Can you rebase it? I want to merge this PR.

matkoniecz:
@nebulon42 Also, why yaml and mml files are modified?

nebulon42:
Rebased.

matkoniecz:
Thanks!

kocio-pl:
Something probably went wrong a bit, since the icon seems to be rendered for areas:

https://www.openstreetmap.org/way/289091540

but not for the nodes (just created, so that can't be old dot):

http://www.openstreetmap.org/node/3410976366

nebulon42:
@kocio-pl There was a problem with the merge, which got fixed by now, but, unfortunately, didn't make it into the current release (yet).

PR #1317, opened 2015-02-16
Render paths under construction more narrow

### LINKED ISSUE #876: Paths under construction are too prominent
Example http://www.openstreetmap.org/#map=19/47.43506/8.35175 . Could be related to #592 For now pro memoria aka very low priority.

math1985:
* Render footway, bridleway, path and track under construction as narrow
as cycleway under construction. This resolves #876.

* Do not render names of roads under construction on zoomlevels in which they are not rendered.


mkoniecz:
At z13 and z14 only name is rendered (from https://github.com/math1985/openstreetmap-carto/blob/construction-path-width/roads.mss#L2247 ), way itself is missing. I think it would be better to not render names.

highway construction construction footway name eeeeee eeeeee eeeeee ref 1 ele 8000 way master - math_construction-path-width

math1985:
@mkoniecz  I think your preview is wrong for z14? In my rendering, z14 is like z15, not like z13.

math1985:
Dropped text rendering from minor roads on z14. The bug reported by @mkoniecz should now be fixed.

Ready for review/merge.

mkoniecz:
@math1985 
> I think your preview is wrong for z14?

I am not sure where the bug is (my preview generator, this style, Mapnik, maybe something else) - but in the worst case this pull reduces this bug (z13 is no longer affected).

In addition named footways/cycleways/etc under construction are really rare and places where this label have chance to appear at z14 is ridiculously rare.

I will investigate this, but I see no reason to block this merge - as it is unsure whatever this bug happens, bug is minor, bug is at worst reduced and much bigger problem is solved.

math1985:
Thanks! It would be nice to find out what causes the bug reported by the tool though.

PR #1319, opened 2015-02-16
cliff rendering improvements

imagico:
This changes the following things:

* fixes the general offset in cliff rendering resulting from a non-centered line pattern image
* use a different line pattern for low zoom (13-14) cliff rendering and man_made=embankment
* render cliff at low zooms in the same color as high zooms

The first point is obvious, here comparison before-after:

cliff line offset

Here the new pattern for man_made=embankment:

embankment rendering

And here zoom 14/15 showing the unified color ([this area](http://www.openstreetmap.org/#map=14/44.0867/39.9164)):

cliff at z=14
cliff at z=15

This improves consistency between zoom levels and avoids too heavy appearance in areas with detailed cliff mapping like [here](http://www.openstreetmap.org/#map=14/-30.1325/28.4522) - labels and other elements are better visible at z=13/14.

Rendering quality of cliffs is of course still fairly poor based on a pattern image - a vector based pattern and adjustable joins would be very good.


ximex:
:+1:

mkoniecz:
First change is great, I am less enthusiasiastic about a separate style for man_made=embankment.

imagico:
Currently embankment reuses the z=13/14 pattern of cliffs for z>=15.  This means cliff and embankment are rendered differently but there is a strange continuity between cliff at z<15 and embankment at z>=15.

There would be two alternative solutions:

* keep reusing the low zoom cliff pattern for embankment with the above changes - this would mean cliff and embankment render identically except for the size of the ticks at high zooms.
* keep using the old, darker cliff pattern for embankment while changing the one for cliffs as above.

What i proposed is very close to the second solution, just that i also modified the geomentry a bit so the pattern differs both in geometry and color.  For comparison: the area shown above is currently rendered like [this](http://www.openstreetmap.org/#map=19/48.07776/7.65022).

mkoniecz:
OK, it makes sense. I noticed also something else: new natural=cliff is paler, with wider line and blurrier ticks.

Paler at low zoom levels makes sense, but I not convinced that making it paler at high zoom levels is also a good idea. And I see no reason for wider line and blurrier ticks.

natural=cliff: Old on left, new on right. z13, z14 higher, z15+ lower. Magnification 600%.

imagico:
First the line at high zooms should be identical to before, the line at low zooms is just a different color.

The difference in your test is probably due to the changed pattern image dimensions (11 pixel vs. 6 pixel high) and as a result a different alignment of the pattern image to the grid of the rendered map.  I don't know how exactly mapnik does the resampling here but i would assume by slightly moving the cliff you could also get the old line blurry and the new one sharp.

The ticks were not antialiased before leading to a sharper but jaggy appearance.  The new ones are somewhat more blurry when rotated.  I think this is preferrable but this is of course a matter of taste.  None of this is ideal, as said a vector based pattern would be much better.

Compare:

oldnew

math1985:
Thanks!

PR #1324, opened 2015-02-20
move mast icon to SVG

nebulon42:
Implements part of #1165.
Mast icon has been inspired by https://github.com/gravitystorm/openstreetmap-carto/pull/1304#issuecomment-73604300. (Thanks @daganzdaanda)

Please note that the icon is now smaller (16px instead of 20px).
This PR also deletes a duplicate mast definition.

before
mast_before

after
mast_after


math1985:
If we don't implement non-communication masts, I think this also works just as well, or better, without foot.

What do others think?

dieterdreist:
2015-02-24 22:41 GMT+01:00 math1985 <notifications@github.com>:

> If we don't implement non-communication masts, I think this also works
> just as well, or better, without foot.



I clearly prefer the foot.

daganzdaanda:
Of course I like it :)

HolgerJeromin:
i like it. 
would like different rendering for communication type, too. But this alone is an improvement.

nebulon42:
I would like to keep the foot if possible.

Issue #1321, opened 2015-02-19
Rendering cycleway with pedestrians


tvainika:
Currently highway=cycleway is always rendered in blue dashed line. However, especially in continental Europe, it is very common that cycleway is shared with pedestrians (foot=yes tag) in some areas, but not in other areas. In addition, there are two types of sharing, which is tagged with segregated=yes/no.

Different types of cycle infrastructure have different quality properties. Also, in some countries different traffic rules apply (if it is obligatory for bicyclist to use that part of the road or not). Third, having different rendering highly visible makes it easier to fix tags.

Therefore, following four different cases should be rendered differently:
1. Dedicated cycleway with tags highway=cycleway + foot=no (implicit): Render non-dashed blue line.
2. Segregated cycleway and footway with tags highway=cycleway + foot=designated + segregated=yes: Render something different, for example different dashing style, which has longer lines every other time
3. Combined cycleway and footway  with tags highway=cycleway + foot=designated + segregated=no: Render something different, for example mixed blue and red in dashed line.
4. Unknown type of cycleway (highway=cycleway, none of the above): Dashed blue line as it is now.

mkoniecz:
I have serious doubts whatever such detail is reasonable on general purpose map.

I think that it belongs rather to a special rendering displaying bicycle-related features in detail.

kocio-pl:
I always thought that mixed foot+bocycle highways should be rendered as mixed red+blue dashed line - dead simple, of course if we don't try to rethink all the cases: than we need something like @tvainika complex proposition before we change rendering.

Speaking of that: I think segregated cycleway and footway should always be mapped as two independent highways, because they can be not really parallel in some parts (think of the tree or grass between them) and we have such a good imagery for some time (at least here in Poland), that we should start mapping and rendering them as areas. Of course, if the tagging is like you suggested, it's better to show it on the map.

HolgerJeromin:
I don't want two independent highways for a way like this:
https://commons.wikimedia.org/wiki/File:Cycle_path_along_Kingsclere_Road_-_geograph.org.uk_-_701953.jpg

polarbearing:
Seems @kocio-pl meant "segregated+physically separated", as she mentioned grass.

kocio-pl:
[...]

Issue #1243, opened 2015-01-19
Timeframe for move to v3.x and availability of new keys


imagico:
The roadmap indicates current development work in v2.x is focussed on fixing problems and internal redesign with only smaller changes to the output.  Many recent changes have however made quite large modifications to the appearance of the style, in particular for example #747, #941 and #1153.  So with regards to the development priorities outlined in the roadmap we are definitely already in the v3.x domain to some extent.

With this background there are many situations where tags currently not available in the rendering database are seriously needed for rendering, in other words there are many widely used tags that could be well used in rendering that are not currently available and this severely limits the style's ability to properly show the work of mappers.

The questions here are

* what are the current plans as to when and how the formal move to v3.x is going to happen?
* what are the factors holding this back, like are there technical requirements that are not met like in terms of the hardware?
* in what form is this going to happen, will there be a simple update to the fixed list of tags or will there be a dynamic approach without a limitation to a fixed list?

I realize there might not yet be answers to these questions but if not it might be time working on them.

nebulon42:
+1 for bringing this up. I assumed this was also somehow tied to the availability of Mapnik 3.x, which is now close ahead (at least there was a first RC).

math1985:
I think this is a question for @gravitystorm and @pnorman.

HolgerJeromin:
This is probably also a question for the helping hands behind the rendering database. 
If i understood correct we need a full reimport of the OSM database with the hstore to have the full (?) osm madness for rendering available.
Perhaps we have an reimport in the mid term future (for hardware reasons?) where we could activate the hstore and after that we can start using this.

pnorman:
> I assumed this was also somehow tied to the availability of Mapnik 3.x, which is now close ahead (at least there was a first RC).

@springmeyer says it will be at least March for Mapnik 3 to be released. Mapnik 3 is a hard requirement for some fixes, and I had hoped it would be out for 3.0 of the style so we could start using its features for other enhancements. If the release drags on, then maybe 4.0 of the style will start to use Mapnik 3 features.

I would say the following are blocked by not having a Mapnik 3 release

* Road shield improvements
* SVGs on anything other than POIs
* Support for non-European languages
* Rotating symbols

-----

Right now I would place fixing some of the remaining big style issues (e.g. roads colours) as more important than hstore.

[...]

PR #1325, opened 2015-02-21
Render shared cycleway+footways different from cycleways

tvainika:
Add 'segregated' tag to .style file to allow using it in styles.

Add different rendering to non-segregated shared cycleway and footway
by alternating red and blue just like separate cycleway and footway
are rendered.

Add longer cycleway dashes to segregated cycleway and footway.

This resolves issue #1321.

Compared to my original proposal in #1321, this has only three different renderings: Current for cycleways without foot=yes/designated tag, and then two different dash patterns for segregated cases yes or no (default). In case of not segregated, rendering is like alternating footway and cycleway. In case of segregated cycleway and footway, it has a bit longer dashes for cycleway, because such segregation is usually considered more suitable for bicycling.


mkoniecz:
Please add before/after images.

tvainika:
Here is before/after images.

Let me know what you think. I actually started thinking myself if this segregated=yes case needs still polishing if you compare it to non-combined cycleway.

cycleway_before
cycleway_after

math1985:
This requires a database reload. See also #1243.

tvainika:
Let me know if you think it would be worth doing without separate case for segregated=yes tag: Combined cycleway and footway rendered by alternating blue and red dots can be done without a database reload. (Just first part of the changes made to roads.mss.) That would resolve my original issue partially.

kocio-pl:
I think red-blue-red-blue pattern is not too visible, maybe it should be less interleaved (like 3xred-3xblue-3xred-3xblue or something)?

tvainika:
Current non-combined cycleway is 1×blue+3×space repeated, non-combined footway is same pattern in red.

If combined cycleway and footway uses something like 3×blue(-3×space)-3×red(-3×space) pattern, should also non-combined cycleways and footways be modified?

math1985:
I think the blue/red mixed combined looks too busy. In addition, I don't think rendering shared cycleways different from cycleway-only is really necessary.

All cycleway-only roads I know are either of two categories: either there is a footway immediately next to it, or pedestrians are officially not allowed to use the cycleway but everyone does so anyway. In neither case, it's not really useful for pedestrians to know that they can't use the cycleway. Is this different in Finland (or anywhere else)?

So I will reject this proposal. But thank you anyway for the effort @tvainika!

mkoniecz:
> either there is a footway immediately next to it, or pedestrians are officially not allowed to use the cycleway but everyone does so anyway

In Poland in the second case law allows pedestrians to use cycleways.

Issue #1307, opened 2015-02-12
highway=steps are too prominent


bubalici:
I suggest a change for the bright red color that the highway=steps has at the moment. Ever since the building color was changed, the steps look to strident, especially in areas where there are lots of such roads (example: https://www.openstreetmap.org/#map=18/37.44711/24.94306 ).
This kind of red is more fitted for restrictions (e.g. access=no tags) or important roads.
At z15 they look even more strange.

As to what color should they have... don't know, perhaps something near the color of pedestrian roads.

HolgerJeromin:
i think step is logical linked to footway. changing both would be a major change. Making footway and pedestrian similar seems good, but footway has to be visible on forest and other landuses.

matkoniecz:
Change of colour is a poor idea, but it is possible to reduce visibility of highway=steps (just a bit, this feature should be still noticeable).

Issue #1248, opened 2015-01-21
Admin borders look like paths


math1985:
For someone not familiar with the style, it is really hard to distinguish paths and admin borders.

Example area: http://www.openstreetmap.org/#map=18/49.60996/6.14509

It would be hard to understand for somebody that he can follow the blue, but not the purple dotted lines.

mboeringa:
I am not sure if I agree here. Yes, this lower level admin border, with its simple dotted line rendering, looks a bit like a path. Usually, in real printed maps, border lines are some kind of mix of long and short dashes, and possibly dots. But I guess with so many admin levels, possibilities are exhausted.

However, since this particular color is consistently used for all admin borders, including the easily recognizable ones like country or municipal borders that people are familiar with, and the fact that all admin borders have a distinct type of labelling with the names of the "regions" drawn on both side of the lines, I think it is recognizable.

In addition, these borders tend to take, or may take, "unusual" path, like going straight through a building or crossing major roads like motorways or primaries without connecting, clearly signifying them as something peculiar like borders.

And at some point, users will simply need to become familiar with a style. Maybe we need a "map" equivalent of RTFM: RTFL(egend) ;-)

mkoniecz:
Potential solution is to make cycleways and footways continous/dashed, not dotted.

math1985:
Would something like this help?
![screenshot from 2015-01-30 13 27 41](https://cloud.githubusercontent.com/assets/5251909/5976269/c8cb93de-a883-11e4-94ff-5a5d7095049e.png)

@RAytoun Do you have any opinion on this issue?

RAytoun:
Hi,
Admin boundaries are a part of the Cadastral range of symbols which are
always going to be a problem overlaying the map base.
Set aside for admin is a range of line and dot symbols with a transparent
screen background highlight to make underlying features still visible. The
line symbol can be black or purple with red purple or light green highlight.
The line starts with International which would be a bold pecked line (Long
dash followed by three shorter dashes) and next level down would be a less
bold line with same pattern of pecks but with dots instead of short dashes)
Next level would contain One long and two short. The linework would be
positioned along the actual border.
The transparent background highlight can be straddling the line as a
highlight or to one side of the line.
I have attached some quick scans (that I was able to lay my hands on
straight away) to illustrate my meaning but not that I am intending them to
be used as such.
I can put together some better examples if need be. Let me know.

gravitystorm:
[...]

PR #747, opened 2014-07-20
Clean-up paths and tracks, in particular on z13/z14

matthijsmelissen:
- Make paths and tracks less visible on z13/z14, by making them narrower and
  remove the casing at these zoomlevels. This declutters in high-density
  environments, while at the same keeping the paths and tracks visible in
  low-density environments.
- Hide private paths and tracks (access=no and access=private) on z13/z14.
- Define path/track widths with variables.
- Make widths of path/track bridges and tunnels more consistent.
- Add rendering for steps in tunnels.
- Add background (glow) to steps, just like footways.

This solves the following issues:
- #211 (Poor result of rendering areas with many footways in zoom level 13)
- #620 (Tracks too dominant on z13)
- Trac 1508: don't render tracks of high tracktype at low zoom (-> wontfix)
- Trac 3788: Don't render highway=track, access=private at z13&14
- Trac 4015: Rendering of highway=path with access=no is inconsistent
- #634: private footways should not be rendered up to z13

PR #1327, opened 2015-02-21
New path rendering

sb12:
I've recently seen a lot of issues regarding cycleways, footways and stairs, so I thought what about a complete new rendering for paths similar to the German OSM Style.

This pull requests changes the following:

- Paths are not rendered as thin lines/dots any more, but similar to roads with a width.
- Footways and stairs are not rendered in "Salmon" anymore, but in the same color as pedestrian areas.
- Paved paths are rendered same as footways
- Paths that are shared by bicycles and footways (path with bicyle=designated and foot=designated, footway with bicycle=yes, cycleway with foot=yes) are rendered with a blue grey dashed outline

This solves the following issues:

- solves #1321 Rendering cycleway with pedestrians 
- solves #1307 highway=steps color change
- solves #1248 Admin borders look like paths
... and probably some more

Some possible issues with this style:

- Confusion between cycleways and water features
- Tracks should be adapted as well for consistency
- Service roads have the same width as paths

Example images:

Low zoom:
path_z14_karlsruhe

High zoom:
path_z17_kit

Bridges:
path_z18

Tunnels:
path_z18_hbf

Bridleway / path high zoom:
path_z18_bridleway

This is just a suggestion and there is still some room for improvement (especially for bridlway and unpaved paths).
I just first wanted to see what people think about such a change before putting more work into it. 

As this is my first major pull request for OSM carto I'm also very happy about any suggestions for code improvement.

### DIFF
--- a/amenity-points.mss
+++ b/amenity-points.mss
@@ -54,6 +54,7 @@
     marker-file: url('symbols/rental_bicycle.16.svg');
     marker-fill: #0092da;
     marker-placement: interior;
+    marker-clip: false;
   }
 
   [feature = 'highway_bus_stop'] {
@@ -62,6 +63,7 @@
       marker-fill: @transportation-icon;
       marker-placement: interior;
       marker-width: 6;
+      marker-clip: false;
     }
     [zoom >= 17] {
       marker-file: url('symbols/bus_stop.p.12.png');
@@ -78,6 +80,7 @@
     marker-file: url('symbols/traffic_light.svg');
     marker-fill: #0a0a0a;
     marker-placement: interior;
+    marker-clip: false;
   }
 
   [feature = 'amenity_cafe'][zoom >= 17] {
@@ -198,6 +201,7 @@
       marker-width: 9;
       marker-line-width: 0;
       marker-placement: interior;
+      marker-clip: false;
     }
   }
 
@@ -205,6 +209,7 @@
     marker-file: url('symbols/communications.16.svg');
     marker-fill: #555;
     marker-placement: interior;
+    marker-clip: false;
   }
 
   [feature = 'tourism_museum'][zoom >= 16] {
@@ -363,6 +368,7 @@
     marker-file: url('symbols/drinking_water.16.svg');
     marker-fill: #734a08;
     marker-placement: interior;
+    marker-clip: false;
   }
 
   [feature = 'amenity_prison'][zoom >= 17] {
@@ -517,6 +523,7 @@
     marker-file: url('symbols/picnic.16.svg');
     marker-fill: #734a08;
     marker-placement: interior;
+    marker-clip: false;
   }
 
   [feature = 'leisure_slipway'][zoom >= 17] {
@@ -526,12 +533,14 @@
 
   [feature = 'aeroway_helipad'][zoom >= 16] {
     marker-file: url('symbols/helipad.svg');
+    marker-placement: interior;
     marker-clip: false;
     marker-fill: @airtransport;
   }
 
   [feature = 'aeroway_aerodrome'][zoom >= 10][zoom < 14] {
     marker-file: url('symbols/aerodrome.svg');
+    marker-placement: interior;
     marker-clip: false;
     marker-fill: @airtransport;
   }
@@ -545,18 +554,21 @@
     marker-file: url('symbols/peak.svg');
     marker-fill: #d08f55;
     marker-placement: interior;
+    marker-clip: false;
   }
 
   [feature = 'natural_volcano'][zoom >= 11] {
     marker-file: url('symbols/peak.svg');
     marker-fill: #d40000;
     marker-placement: interior;
+    marker-clip: false;
   }
 
   [feature = 'natural_saddle'][zoom >= 15] {
     marker-file: url('symbols/saddle.svg');
     marker-fill: #d08f55;
     marker-placement: interior;
+    marker-clip: false;
   }
 
   [feature = 'natural_cave_entrance'][zoom >= 15] {
@@ -567,6 +579,7 @@
   [feature = 'natural_spring'][zoom >= 14] {
     marker-file: url('symbols/spring.svg');
     marker-placement: interior;
+    marker-clip: false;
   }
 
   [feature = 'power_generator']['generator:source' = 'wind'],
@@ -595,19 +608,22 @@
   [highway = 'mini_roundabout'][zoom >= 16]::highway {
     marker-file: url('symbols/mini_roundabout.svg');
     marker-placement: interior;
+    marker-clip: false;
   }
 
   [barrier = 'gate']::barrier {
     [zoom >= 16] {
       marker-file: url('symbols/gate.svg');
       marker-placement: interior;
+      marker-clip: false;
     }
   }
 
   [barrier = 'lift_gate'][zoom >= 16]::barrier {
     marker-file: url('symbols/liftgate.svg');
     marker-fill: #3f3f3f;
-    marker-placement: interior
+    marker-placement: interior;
+    marker-clip: false;
   }
 
   [barrier = 'bollard'],
@@ -1887,6 +1903,7 @@
       text-wrap-width: @landcover-wrap-width-size;
       [way_pixels > 12000] {text-wrap-width: @landcover-wrap-width-size-big; }
       [way_pixels > 48000] {text-wrap-width: @landcover-wrap-width-size-bigger; }
+      text-placement: interior;
     }
   }
 
@@ -1905,6 +1922,7 @@
       text-wrap-width: @landcover-wrap-width-size;
       [way_pixels > 12000] {text-wrap-width: @landcover-wrap-width-size-big; }
       [way_pixels > 48000] {text-wrap-width: @landcover-wrap-width-size-bigger; }
+      text-placement: interior;
     }
   }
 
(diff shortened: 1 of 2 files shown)


RAytoun:
@sb12  I am in favour of paths and tracks rendered similar to roads at this zoom level and it allows for a number of different lines and fills to extend the classifications of use and leaves a lot more single line options open for the other features that require single line rendering.

mkoniecz:
> Failed — The Travis CI build failed

@sb12 See https://github.com/gravitystorm/openstreetmap-carto/blob/master/CONTRIBUTING.md#editing-layers - it seems that you need to run `./scripts/yaml2mml.py < project.yaml > project.mml && touch project.mml`

sb12:
> Failed — The Travis CI build failed

Fixed. Thanks @mkoniecz

mkoniecz:
@sb12 

> As this is my first major pull request for OSM carto I'm also very happy about any suggestions for code improvement. 

> substr(surface, 0, 11) = 'cobblestone'

Wildcards are considered inferior to explicit values. Also, some known paved surface values are missing (I noticed lack of surface=wood and surface=metal).

> Paved paths are rendered same as footways

It is important decision whatever highway values should directly correspond to render style. What about unpaved highway=footway? What about paved tracks? Is it necessary to make clear whatever way is tagged as highway=track/path like it is done now?

Rovastar:
This will need further discussion as it is a large effecting change.

I would like to see before and after pics.
pics showing multiple ways like pedestrian,  path, etc

Not sure about fixing a style for paved for this one thing.

I would prefer a integrated solution for all paved./unpaved ways.
just doing one tiny aspect of this can led to confusion now or later.

Probably best doing separate pull requests for each issue.

imagico:
I think this would need some more work - rendering foot/cycleways wider than tracks does not seem a good idea.  Also make sure to test it in areas with very dense path mapping - like [here](http://www.openstreetmap.org/#map=16/47.9953/7.8598).

sb12:
Thanks for your feedback.

> Wildcards (like in substr(surface, 0, 11) = 'cobblestone') are considered inferior to checking for explicit values (see implementation of rare shop types). Also, some known paved surface values are missing (I noticed lack of surface=wood and surface=metal).

For now I added all values that are mentioned in the wiki as paved (for some reason wood and metal are considered to be unpaved in the wiki?).
The aim of the wildcards was to also cover values like `surface=paving_stones:20` and `surface=cobblestone:10` etc. which are valid values according to the wiki.

But anyway:
>I would prefer a integrated solution for all paved./unpaved ways.
just doing one tiny aspect of this can led to confusion now or later.

I added the surface value because I wanted to get rid of the path controversies. (What's the difference between a path and a footway etc.)
It's probably better to ignore the surface for now and just have three different styles of paths:

- path for pedestrians
- path for pedestrians and cyclists
- path only for cyclists

For low zoom levels (e.g. <= 16) it might be an option to have the same style for all paths. This also solves the issue of cycleways being confused with rivers. 

The surface issue can be solved later then.

> I think this would need some more work - rendering foot/cycleways wider than tracks does not seem a good idea.

I agree that it does need more work, especially regarding tracks. But this pull request wasn't meant to be merged like it is, it was mainly supposed to be a base for discussion whether such a major change in style could solve the footway and path problems.

If we agree that this might be a solution, the process would be probably like this:

- change track style, so that tracks are wider than footways.
- change the footway style similar to this pull request.
- optional: solve the paved/unpaved issue

>I would like to see before and after pics.
pics showing multiple ways like pedestrian, path, etc

Some before and after pictures:

npath_z17_schlosskarlsruhe npath_z17_schlosskarlsruhe_after

npath_z18_fasanengarten npath_z18_fasanengarten_after


>Also make sure to test it in areas with very dense path mapping - like here.

npath_z13_schwetzingen npath_z13_schwetzingen_after

npath_z14_schwetzingen npath_z14_schwetzingen_after

npath_z16_heidelberg npath_z16_heidelberg_after

kocio-pl:
These pictures are quite convincing for me, but I would also add a z=13 example - just for completeness.

imagico:
Thanks for the additional examples - i think the contrast on urban landuses is not sufficient and i also don't really like the dashed casing which seems quite disturbing and not really that readable.

In general it seems to look quite reasonable in urban environments but much less so in for rural paths.  Here the black dashed rendering currently used for highway=path seems to work quite well and it will be difficult to improve on that in a way that looks well readable on all kinds of background.

mkoniecz:
Footways are significantly prettier and (for me) no readability cost. But cycleways are not recognizable (cycleway looks like footway with ditches). Rendering fill blue would make it river-like, so I see no obvious solution.

Is it possible to make some before/after in rural area with low road and footway density? For now only z16 is presented - and it is in forest what makes it easier to notice white lines.

Maybe it would be better to ignore highway=path and surface for now (it has many open questions, starting from "what should be considered as paved").

mkoniecz:
And thanks for this proposal, it is really promising idea.

daganzdaanda:
Nice, I like the close-ups a lot already! 

For the record, here is the last time footways and cycleways have been changed and discussed at length: #747
How would your style look in an area like https://www.openstreetmap.org/#map=14/52.9232/-0.8588 ?

sb12:
>These pictures are quite convincing for me, but I would also add a z=13 example - just for completeness. 

See first picture of "Schwetzingen" in my last post.

>Is it possible to make some before/after in rural area with low road and footway density? For now only z16 is presented - and it is in forest what makes it easier to notice white lines.

Sorry I didn't really find a good rural area with paths and without forest in the area my database covers.
But here are some footways (actually mapped as cycleways) on agricultural area (see http://www.openstreetmap.org/#map=14/49.3003/8.5597 for before pictures):
path_rural_z13
path_rural_z14
path_rural_z15
path_rural_z16

The contrast on farmland is not very good, yet, especially on level 13 and 15, so this needs a bit more tweaking. Maybe for levels 13 to 15 it is enough to have a grey or light salmon line without casing.

> i think the contrast on urban landuses is not sufficient 

Would it help to give footways and pedestrian roads a red (salmon) casing:
path_red_casing_z16
path_red_casing_z17
path_redfootway_z14

>i also don't really like the dashed casing which seems quite disturbing and not really that readable.

What about smaller dashes:
path_small_dashes_z19

> But cycleways are not recognizable (cycleway looks like footway with ditches).

I'm also not that happy with the cycleways. See for example here they look like water features:
path_cycleways_z15
path_cycleways_z16

A possible solution might be to make the blue less strong with the drawback that it is harder to distinguish between footways and cycleways. This way also the dashes for mixed cycle/footways might be less disturbing. See e.g. the German OSM Style for a reference: http://openstreetmap.de/karte.html?zoom=15&lat=48.99818&lon=8.3766&layers=B000TT

mkoniecz:
> I'm also not that happy with the cycleways. See for example here they look like water features:

Random untested (for now) idea: what about solid fill like light orange/pink/green or strong green?

> Would it help to give footways and pedestrian roads a red (salmon) casing:

Main result is making map uglier.

math1985:
Thanks for the work so far. Some more comments:
- I'm not sure if with the wider rendering it's clear footways can't be used by cars.
- Footways look bad on forest backgrounds on z<14.
- It might be confusing for mappers that pedestrian and footway look so similar.

mkoniecz:
> It might be confusing for mappers that pedestrian and footway look so similar.

Maybe bigger difference would be good but it is way better than current situation with massive difference despite minor difference (compared to other features).

kocio-pl:
@mkoniecz: +1

RobJN:
I feel that we should focus most on the paved/unpaved tag, and bicyle routes. The bicycle=yes, foot=not combination seems a bit specialist for a general map. As does the segregated, non-segretad tag.

My suggestion:

* Up to zoom 16 render as lines (like highway=path for non-bicycle access, and like highway=cycleway for bicycle access)
* From zoom 17+ start rendering paved with a width. The red casing is not needed but perhaps keep blue casing for bicycle access. Unpaved continues as line.
* Change colour of highway=steps.

This drops the use of red (like highway=footway) entirely.

scaidermern:
Actually I like the current red dashed rendering of footways. It makes it really obvious that these ways can't be used by motor vehicles. Changing the style like suggested above makes the map look nicer, but it becomes less useful as footways look too similar to residential ways.

kocio-pl:
Highway=pedestrian already looks more like residential ways, but I heard no complaints about it.

sb12:
I now updated the pull request:

- I dropped the changes regarding surfaces from the PR. This should be discussed separately
- This PR now only covers footways and cycleways
- I changed the style for low zoom levels
- Footways and Cycleways have the same style on level 16 and lower
- Shared Foot- and Cycleways are shown dashed only from level 17.

Before and after renderings:

Low zoom levels:

npath_z13_schwetzingen
npath_z13_schwetzingen_after_new

npath_z14_schwetzingen
npath_z14_schwetzingen_after_new

forest:
npath_z14_mannheim
npath_z14_mannheim_after_new

Cemetery and allotments:
npath_z15_cemetery
npath_z15_cemetery_after_new

Higher zoom levels:
npath_z16_heidelberg
npath_z16_heidelberg_after_new
npath_z17_schlosskarlsruhe
npath_z17_schloss_karlsruhe_after_new
npath_z18_fasanengarten
npath_z18_fasanengarten_after_new

PR #1331, opened 2015-02-22
Made Travis output on yaml2mml.py failure more comprehensive

### LINKED ISSUE #1329: Make Travis CI build log less cryptic
Currently it is rather hard to guess how 

> The command "scripts/yaml2mml.py < project.yaml > project.mml2 && diff project.mml project.mml2" exited with 1.

should be fixed. I think that mentioning https://github.com/gravitystorm/openstreetmap-carto/blob/master/CONTRIBUTING.md#editing-layers would be a good idea.

Also, is it possible to eliminate wall of text (lines 142-156 in https://travis-ci.org/gravitystorm/openstreetmap-carto/builds/51634644)? Or is it useful for some purpose?

floscher:
Fixes #1329.

See https://travis-ci.org/floscher/openstreetmap-carto/builds/51741837 for a failing build and https://travis-ci.org/floscher/openstreetmap-carto/builds/51741018 for a sucessful build.

This suppresses the output of the complete diff, because I'd say that in only a very few cases it would be of interest **what** differs between `project.mml` and `project.mml2`, the information **that** there are differences should be enough. In most cases the author of the commit simply forgot to run `yaml2mml.py` (like in https://travis-ci.org/gravitystorm/openstreetmap-carto/builds/51634644).

This could surely be done in a oneliner, but I wanted to keep it readable. Additionally Travis will always print out the full command. That means it would also print the string saying that the files are out of sync, even when that is not the case. This is avoided by putting the code in a separate script-file.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,4 +20,4 @@ install:
 script:
   - jsonlint project.mml
   - ./node_modules/carto/bin/carto project.mml | xmllint - | wc -l
-  - scripts/yaml2mml.py < project.yaml > project.mml2 && diff project.mml project.mml2
+  - ./scripts/travis_check_project_files
--- a/scripts/travis_check_project_files
+++ b/scripts/travis_check_project_files
@@ -1,0 +1,12 @@
+#!/bin/bash
+scripts/yaml2mml.py < project.yaml > project.mml2 && diff -q project.mml project.mml2
+exitcode=$?
+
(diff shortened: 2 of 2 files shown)


mkoniecz:
Thanks! One small thing - I think that there should be a full stop after "for further information".

floscher:
OK, the commit is updated. There was also a "t" missing in "committing".

Issue #325, opened 2014-02-02
Change rendering for natural=spring


matkoniecz:
http://www.openstreetmap.org/?mlat=49.75063&mlon=19.79288#map=19/49.75063/19.79288

1) It looks like 5, not like S
2) I am unsure whatever S = Spring is obvious even for native speakers

I suggest some more obvious icon - maybe something like https://commons.wikimedia.org/wiki/File:RiverIcon-Spring.svg ?

Also orienteering maps have a good symbol for springs:

![beztytuu](https://f.cloud.github.com/assets/899988/2060360/5c2b8bec-8c11-11e3-94df-c1926a60b1ab.png)

from http://www.maprunner.co.uk/simon/mapsymbols.jpg

nebulon42:
For nodes that are part of a waterway a simple circle would be sufficient.
Like here: http://www.openstreetmap.org/#map=17/48.18177/16.24524
![bildschirmfoto vom 2015-01-03 23 22 42](https://cloud.githubusercontent.com/assets/3531092/5603973/999bd266-939f-11e4-85d4-1f7f8d6ac147.png)

Detached springs would look weird, though:
![spring](https://cloud.githubusercontent.com/assets/3531092/5603983/21842e58-93a0-11e4-8d91-6df4c842817b.png)

While I like the icon of @zdila it does not look so good connected to a waterway if not rotated properly:
![bildschirmfoto vom 2015-01-03 23 37 32](https://cloud.githubusercontent.com/assets/3531092/5604002/90b1dfa4-93a1-11e4-8ff3-2bdd599f9447.png)

Or something like @javbw proposed?
![bildschirmfoto vom 2015-01-03 23 50 25](https://cloud.githubusercontent.com/assets/3531092/5604025/6155fd88-93a3-11e4-961c-369cd8f33100.png)

Admittedly, this needs more work so that it doesn't look like a flower, but I like the last variant best so far.

kocio-pl:
It's quite clear that it's not a flower because it has a water color and is not symmetric. It's much better than current symbol, so I would be happy to replace it as soon as possible. I think I will also use [your icon for the fountain](https://github.com/nebulon42/osmic/blob/master/outdoor/fountain-14.svg), so there's no risk of confusing them.

kocio-pl:
What about something like this on z14:
[...]

PR #1341, opened 2015-02-28
add icon halos (to refreshed SVG icons)

### LINKED ISSUE #1189: Icon outlines

nebulon42:
Adds white 1px wide halos with 30% opacity to refreshed SVG icons.
This technique is similar to how Maki icons handle this.
Partly fixes #1189.

This improves the situation on noisy background especially for icons with larger inner holes such as ATM:

before
halo_atm_before

after
halo_atm_after

Some general overview (click for full size):
before
halo_before

after
halo_after

Since icons are now 2px larger than before this would need systematic `text-dy` checking.
@mkoniecz Could this be done with your tool and is it a lot of work?


mkoniecz:
@nebulon42

> Could this be done with your tool

Yes. 

> and is it a lot of work?

Cost of human work (setup) is negligible but it is significant and long work for computer. I will run it, but it may wait day or two for a feasible time (computer not used for anything for a significant part of day - as at least on my hardware it significantly degrades performance of other software).

math1985:
> Since icons are now 2px larger than before this would need systematic text-dy checking.

Does it mean that all/most text need to be shifted down 1px? I think the distance between text and icons is already quite large, would be a shame if we'd have to enlarge it even further.

nebulon42:
>  I will run it, but it may wait day or two for a feasible time

Great, thanks very much for checking this.

> Does it mean that all/most text need to be shifted down 1px?

I don't know but I think not necessarily because the distance is already quite large as you said. But it has to be checked to be sure.

pnorman:
Yes - adding a 2px buffer will require adding 2px to dy.

Are there any ways to do the halo within mapnik with comp-op or other techniques? I'd much rather not have to do them in all the SVGs.

gravitystorm:
How does adding halos affect the svg recolouring? One advantage of the SVG icons for downstream users is being able to set `marker-fill` but would setting this change the colour of the halo too?

nebulon42:
The halos consist of line strokes that are not affected by `marker-fill`. With `marker-line-color` and `marker-line-width` it is possible to change the colour and width of the halo as can be seen in this example:
halo_green

Regarding comp-op I'm not sure. I have experimented with stacking icons some time ago and don't remember all the details, but as far as I remember the problems where:
* Mapnik currently seems to be unable to add line strokes with `marker-line-width` when they were not present before. That would mean that we would require two different files per icon at the moment.
* Stacking of icons seemed to involve the use of `marker-allow-overlap: true` which has the undesired side effect of also allowing overlap with other markers nearby.

But I might be wrong on both of them, so it might be good to experiment more with it.

mkoniecz:
> Yes - adding a 1px buffer will require adding 1px to dy.

Yes, this propose change would force increasing text-dy - that is anyway really big.

Also, I think that change in example is not improvement. Loss of labels is not really improving readability, and there is loss of information.

> icons with larger inner holes such as ATM

The problem is with inner hole, not with nearby labels/icons.

nebulon42:
First thanks for systematically testing this. Could you please quantify your statement a bit? Like "all text-dy have to be increased" or "30% of text-dy have to be increased". This does not have to be exact, but I'd love to see how many POIs would be affected.

I don't know what you mean with loss of information. I have always said that text-dy would have to be adjusted before this can be merged.

On the other hand, I really think that icons need halos. Text has halos, some even quite unnecessarily big. So why should icons don't have it? The example should not be seen as ideal or as really illustrative for this change. There are certain noisy backgrounds and combinations where halos will be helpful. See again my example in https://github.com/gravitystorm/openstreetmap-carto/issues/325#issuecomment-68612835 with spring within forest where a halo would certainly improve visibility.

HolgerJeromin:
is there an option in mapnik to prevent the collision of text and icon at all? I have no Problem with a high text overlapping the icon halo.

mkoniecz:
> I don't know what you mean with loss of information.

Reducing amount of displayed relevant labels means that some information is removed from displayed map.

> Could you please quantify your statement a bit?

Now entire test finished, with following results:

```
amenity=biergarten - name is missing for name 'ÉÉÉÉÉÉ ÉÉÉÉÉÉ ÉÉÉÉÉÉ' on z22
amenity=cafe - name is missing for name 'ÉÉÉÉÉÉ ÉÉÉÉÉÉ ÉÉÉÉÉÉ' on z22
amenity=cinema - name is missing for name 'ÉÉÉÉÉÉ ÉÉÉÉÉÉ ÉÉÉÉÉÉ' on z22
amenity=embassy - name is missing for name 'ÉÉÉÉÉÉ ÉÉÉÉÉÉ ÉÉÉÉÉÉ' on z22
amenity=fast_food - name is missing for name 'ÉÉÉÉÉÉ ÉÉÉÉÉÉ ÉÉÉÉÉÉ' on z22
amenity=fire_station - name is missing for name 'ÉÉÉÉÉÉ ÉÉÉÉÉÉ ÉÉÉÉÉÉ' on z22
amenity=food_court - name is missing for name 'ÉÉÉÉÉÉ ÉÉÉÉÉÉ ÉÉÉÉÉÉ' on z22
amenity=fuel - name is missing for name 'ÉÉÉÉÉÉ ÉÉÉÉÉÉ ÉÉÉÉÉÉ' on z22
amenity=hospital - name is missing for name 'ÉÉÉÉÉÉ ÉÉÉÉÉÉ ÉÉÉÉÉÉ' on z22
amenity=police - name is missing for name 'ÉÉÉÉÉÉ ÉÉÉÉÉÉ ÉÉÉÉÉÉ' on z22
amenity=pub - name is missing for name 'ÉÉÉÉÉÉ ÉÉÉÉÉÉ ÉÉÉÉÉÉ' on z22
amenity=restaurant - name is missing for name 'ÉÉÉÉÉÉ ÉÉÉÉÉÉ ÉÉÉÉÉÉ' on z22
amenity=theatre - name is missing for name 'ÉÉÉÉÉÉ ÉÉÉÉÉÉ ÉÉÉÉÉÉ' on z22
historic=memorial - name is missing for name 'ÉÉÉÉÉÉ ÉÉÉÉÉÉ ÉÉÉÉÉÉ' on z22
leisure=golf_course - name is missing for name 'ÉÉÉÉÉÉ ÉÉÉÉÉÉ ÉÉÉÉÉÉ' on z22
leisure=miniature_golf - name is missing for name 'ÉÉÉÉÉÉ ÉÉÉÉÉÉ ÉÉÉÉÉÉ' on z22
tourism=museum - name is missing for name 'ÉÉÉÉÉÉ ÉÉÉÉÉÉ ÉÉÉÉÉÉ' on z22
```

nebulon42:
Thanks very much, that means that this affects roughly 37% of the icons in this PR.

math1985:
To me, adding the halo is not a visual improvement. I don't like the halo for the labels either, but there we really need them for readability (although I succeeded in reducing the visual impact of the label halos). I think the icons are readable enough without halo.

I also think the text-dy is already too big, and really wouldn't like to increase it even more.

So in short, I would suggest not to add halos to the icons.

pnorman:
I also don't think this is worth it.

daganzdaanda:
Looking back at this - why would we even need more text-dy? 
Is it possible that an icon's halo might render on top of the text? Or are the halos for text and icon on the same "layer" in the background? 

In the latter case, we could make the halos for text and icon as large as we like, while keeping the current text-dy. But even in the first case, a 1px outline around an icon could only touch the outline of the text. That is not a problem at all, IMHO.

daganzdaanda:
may be related to #1853

gravitystorm:
> Looking back at this - why would we even need more text-dy?

Adding a halo to an icon makes for a bigger icon. Bigger icon, bigger text-dy.

> Is it possible that an icon's halo might render on top of the text?

No. The halo is part of the icon. It just makes the image larger.

> Or are the halos for text and icon on the same "layer" in the background?

I'd encourage you to try experimenting with tilemill/kosmtik to get a better feeling for how it works. But in short, the icon is "one thing" and the text(+text-halo) is "one thing" and if they overlap, the text isn't shown.

Issue #657, opened 2014-06-22
Provide installation script


matkoniecz:
Version for Ubuntu 14.04: https://github.com/gravitystorm/openstreetmap-carto/issues/657#issuecomment-121279334

Current setup instructions require unnecessary figuring out what and how it should be done.

But it is possible to assume that user have certain Linux (for example in VM) and provide installation script. It would be possible to provide entire installation instruction also in form:

install Lubuntu 14.04, it may be a virtual machine
$ sudo apt-get install git
$ git clone openstreetmap-carto
$ sudo sh ./openstreetmap-carto/lubuntu14.04_installation_script.sh

script would include
- set up of PostGIS database
- installing listed dependencies and fonts
- running get-shapefiles.sh
- fetching some small area suitable for testing and loading this data into database

Note that currently some people even make pull requests without testing them (see for example #654 ).

I may make and test the installation script (it would be mostly copy of http://switch2osm.org/loading-osm-data/ ).

mkoniecz:
I tested and improved previous comment (mostly wording - and two instructions "Open Tilemill and close it" + final one).

math1985:
As far as I'm concerned, we add this to https://github.com/math1985/openstreetmap-carto/blob/master/INSTALL.md .

pnorman:
I have had nothing but problems with ppa:developmentseed/mapbox causing conflicts with other versions.

mkoniecz:
> I have had nothing but problems with ppa:developmentseed/mapbox causing conflicts with other versions.

So what is the proper way of installing Tilemill? There are recommending this method on their website.

Or is there a suitable replacement for Tilemill?

pnorman:
I'd prefer recommending Kosmtik over Tilemill.

* Kosmtik is better suited to large complex stylesheets. Tilemill tries to be a text editor, but can't deal with more than 3-4 .mss files
* Tilemill is essentially abandoned. Kosmtik is more of a community driven effort
* Kosmtik is smaller

The main differences with Kosmtik install and setup are

* You need to install it from source (@yohanboniface, any progress there?)
* You point it directly at the project.yaml file
* While running, it only deals with one project

For a new user creating new projects which will only involve 1 MSS file, I could see recommending Tilemill, as the built-in text editor makes it easier to start, but it just doesn't work on large projects.

yohanboniface:
> You need to install it from source (@yohanboniface, any progress there?)

[...]

PR #1342, opened 2015-03-01
Quick dev. environment using Vagrant & Salt

thor:
Following straight from the commit:
> Ubuntu 14.04.1 with osm2pgsql, postgis and kosmtik ready to use
 - Uses Vagrant to quickly create a VM based on Ubuntu 14.04.1 LTS
    - Adjustable amounts of CPU cores and memory
 - Easily have your own kosmtik plugins ready if you want to start from scratch
    - Sets up kosmtik as a service, so you can easily start/stop using $ service kosmtik <start/stop/status/restart>
 - Added .vagrant/ and Vagrantfile to the ignore list to avoid unintentional commits with edits
 - Does NOT import any .pbfs automatically, as it can quickly get messy
   doing such. However, I've played around with it and it works -- as
   long as you've got the memory and you're not importing overlapping
   areas.
 - However, it does execute get-shapefiles for you -- just `vagrant up` or `vagrant provision`
 - Edit as you please and reload it for kosmtik to pick it up
 - All you need to do to get started is to import with osm2pgsql

I queried @pnorman about this very briefly in #osm-dev at some point, and so here is a PR to see if there is any interest in adding this as part of the repo.
The benefits are mainly listed above, but the recap would be the following:
  - No messing around with relatively long lists of instructions
    - *The configuration is combined from switch2osm and the notepad with instructions that I've seen linked around here*
  - Allows for a "shared" development environment
    - *This is only to some degree; you all have vastly different HW requirements and desires for what and how to render, but it's an easy way to get an identical setup*
  - Makes it a whole lot easier for people on Windows to join the party
    - *I started it while I was using Windows myself, but even on any other distribution it's nice to not have to worry about the development environment too much. Note the .gitattributes addition is to avoid gits conversion of line endings for the shell scripts, so that errors don't occur when running i.e. `get-shapefiles.sh`*

For those of you not familiar with [Vagrant](http://docs.vagrantup.com/v2/getting-started/index.html), you won't be adding a whole VM to the repository -- just a configuration file for Vagrant and Salt roots for configuring the VM for osm2pgsql, kosmtik and postgis. (Though, not in _that_ order.) You can also clone my branch and simply run `vagrant up` once you've installed Vagrant on your computer, and have a look. 

Usage is currently as follows:
1. If not already present, install Vagrant and VirtualBox.
2. Clone the repository.
3. If you'd like kosmtik plugins, edit the Vagrantile and add the ones you want.
4. Run `vagrant up` from within the repository. 
5. Run `vagrant ssh` to SSH in, so that you can run your own `osm2pgsql` import. (Place any files in data/ and they'll be available at `/srv/openstreetmap-carto/data/` inside the VM.)
6. Once data is imported, the map will magically be visible at http://127.0.0.1:6789.
7. Ideally, edit stuff and see how simple it is to `vagrant destroy && vagrant up` if you want to recreate it.

Either way, if it's not something for you folks, that's completely understandable: the PR addes one file to the root, as well as an entire folder containing 4 files. (These can be put elsewhere if desired.) Should it be something that is a bit too meta, then I'll just keep it separated in its own repository. However, at least for me it's really practical when I can clone a repo, run `vagrant up` and be at the finish line for what I need to contribute to a project. :sunny:

If there is an interest in it, but something is amiss, let me know and I'll try what I can to fix & improve it.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,3 @@
+# Keep get-shapefiles.sh with LF-endings as to avoid Vagrant issues
+# Should be the case for all shell-files, and as such is *.sh
+*.sh text eol=lf
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,3 +1,4 @@
 # Keep get-shapefiles.sh with LF-endings as to avoid Vagrant issues
 # Should be the case for all shell-files, and as such is *.sh
 *.sh text eol=lf
+
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,5 @@ layers/
 data/
 *.xml
 node_modules/
+.vagrant/
+Vagrantfile
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -1,0 +1,73 @@
+# -*- mode: ruby -*-
+# vi: set ft=ruby :
+
+# Vagrantfile API/syntax version. Don't touch unless you know what you're doing!
+VAGRANTFILE_API_VERSION = "2"
+
+Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
+  # All Vagrant configuration is done here. The most common configuration
+  # options are documented and commented below. For a complete reference,
+  # please see the online documentation at vagrantup.com.
+
+  # Every Vagrant virtual environment requires a box to build off of.
+  # This can be changed to anything else, but keep in mind the states
+  # are only written with 14.04.1 in mind.
+  config.vm.box = "ubuntu/trusty64"
+
+  # Setting up port-forwarding for kosmtik, so you can access it at
+  # http://127.0.0.1:6789
+  config.vm.network "forwarded_port", guest: 6789, host: 6789
+
+  # Configurations specific to customizing the virtual machine for the 
+  # provider we use. Here, we use virtualbox, but you can use another if
+  # you'd like -- just remember to make a separate section for it, as
+  # this only applies to VirtualBox.
+
+  config.vm.provider "virtualbox" do |vb|
+    # We're dealing with memory intensive business: minimum should be 2048MB.
+    # Default here is 4096MB, but you can decrease or increase it as you need.
+    vb.customize ["modifyvm", :id, "--memory", "4096"]
+    
+    # The following are recommended values for the VM for relatively high 
+    # degrees of responsiveness. Note that you need virtualization extensions
+    # enabled on your computer to use the last option.
+    # The extra CPU cores for the VM are highly recommended for importing
+    # larger areas, especially when dealing with complete countries.
+    # vb.customize ["modifyvm", :id, "--cpus", "4"]
+    # vb.customize ["modifyvm", :id, "--hwvirtex", "on"]
+  end
+
+  # Share an additional folder to the guest VM. The first argument is
+  # the path on the host to the actual folder. The second argument is
+  # the path on the guest to mount the folder. And the optional third
+  # argument is a set of non-required options.
+
+  # Here we share the same folder to /srv/openstreetmap-carto/ so that
+  # we can have the sources available for kosmtik to render.
+  config.vm.synced_folder ".", "/srv/openstreetmap-carto"
+  # Since we want to run Salt masterless, we've got to setup the Salt root
+  # as well, resulting in us being able to get the dev. environment up.
+  config.vm.synced_folder "salt/", "/srv/salt/"
+
+  # Set up the Salt provisioner, without much special configuration.
+  config.vm.provision :salt do |salt|
+    salt.minion_config = "salt/minion"
+    salt.run_highstate = true
+
+    # IF you'd like to use any plugins for kosmtik, you can provide them 
+    # underneath here by uncommenting the block. Check out the kosmtik
+    # project on github for a list over plugins, or execute
+    # $ node index.js plugins --available 
+    # while in /usr/local/kosmtik
+    
+    #salt.pillar({
+    # "kosmtik" => {
+    #   "plugins" => [
+    #     "kosmtik-map-compare",
+    #     "kosmtik-place-search"
+    #   ]
+    # }
+    #})
+  end
+  
+end
--- a/salt/gis.sls
+++ b/salt/gis.sls
@@ -1,4 +1,9 @@
-# get the software packages installed
+# State for configuring the packages and requirements for PostGIS locally.
+# Configures a very basic setup with increased memory for postgresql,
+# along with software needed for importing OpenStreetMap data.
+
+
+# Install basic tools that are useful to have for installing this and that.
 base:
   pkg.installed:
     - install_recommends: false
@@ -22,22 +27,20 @@ gis:
       - postgresql-contrib-9.3
       - proj-bin
       - libgeos-dev
-  # Let's make it easy for our GIS-in-aid software to access our PostGIS database
-  file.blockreplace:
-    - name: /etc/postgresql/9.3/main/pg_hba.conf
-    - marker_start: "# TYPE  DATABASE"
-    - marker_end: '# "local" is'
-    - content: "local   all             gisuser                                 trust"
+  # Making sure the postresql service is running, and that it's restarted
+  # when changes to the configuration are processed. Requires gis beforehand.
   service:
     - name: postgresql
     - running
     - watch:
-      - file: postgresql.conf.work_mem
-      - file: postgresql.conf.maintenance_work_mem
+      - file: postgis-postgresql.conf
       - file: gis
     - require:
       - pkg: gis
       - file: gis
+  user.present:
+    - name: gisuser
+    - empty_password: True
   postgres_user.present:
     - name: gisuser
     - superuser: true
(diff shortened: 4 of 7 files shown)


.gitignore
```diff
@@ -3,3 +3,5 @@ layers/
 data/
 *.xml
 node_modules/
+.vagrant/
+Vagrantfile
```
pnorman: You're checking in the Vagrantfile, should it really be ignored?

.gitattributes
```diff
@@ -0,0 +1,3 @@
+# Keep get-shapefiles.sh with LF-endings as to avoid Vagrant issues
+# Should be the case for all shell-files, and as such is *.sh
+*.sh text eol=lf
```
pnorman: There's a missing EOL in this file

salt/gis.sls
```diff
@@ -0,0 +1,126 @@
...
+  # add the ppa for osm2pgsql with osmctools
+  pkgrepo.managed:
+    - ppa: kakrueger/openstreetmap
+  # install the packages from above-mentioned repo!
+  pkg.latest:
+    - install_recommends: false
+    - refresh: true
+    - pkgs:
+      - osm2pgsql
+      - osmctools
```
pnorman: osmctools is actually part of 14.04, so we could install it earlier.

.gitignore
thor: @pnorman It's mentioned in the commit, but it's ignored to avoid accidentally pushing a modified Vagrantfile unless it really is your intention, as you'd use it to configure the virtual machines number of CPU cores, memory, on/off for the hardware extensions for virtualization -- as well as changing which kosmtik plugins get installed during/by the Salt highstate / `vagrant provision`.

.gitattributes
thor: @pnorman Added EOL in d6659b2.

salt/gis.sls
thor: @pnorman I hadn't noticed that! I updated the comments surrounding it, but it's the only OSM specific ID in the GIS state. So far I've let it stand alone in the event I'd want to `require` it for another step, such as importing during `vagrant provision`. 

(I have another state for that, but I'm not satisfied with it, as it ends up with the whole provisioning taking a loooong time and there's no user feedback while `osm2pgsql` does its job.)

Vagrantfile
```diff
@@ -0,0 +1,73 @@
...
+VAGRANTFILE_API_VERSION = "2"
+
+Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
+  # All Vagrant configuration is done here. The most common configuration
+  # options are documented and commented below. For a complete reference,
+  # please see the online documentation at vagrantup.com.
+
+  # Every Vagrant virtual environment requires a box to build off of.
+  # This can be changed to anything else, but keep in mind the states
+  # are only written with 14.04.1 in mind.
```
pnorman: 14.04.2 now?

salt/gis.sls
```diff
@@ -0,0 +1,127 @@
...
+      - postgresql-9.3-postgis-2.1
+      - postgresql-contrib-9.3
+      - proj-bin
+      - libgeos-dev
+  # Let's make it easy for our GIS-in-aid software to access our PostGIS database
+  file.blockreplace:
+    - name: /etc/postgresql/9.3/main/pg_hba.conf
+    - marker_start: "# TYPE  DATABASE"
+    - marker_end: '# "local" is'
+    - content: "local   all             gisuser                                 trust"
```
pnorman: do we need this with the `postgres_user.present` later in the file?

salt/gis.sls
```diff
@@ -0,0 +1,127 @@
...
+postgis:
+  postgres_extension.present:
+    - db_user: gisuser
+    - maintenance_db: gis
+    - require:
+      - service: gis
+
+# Conservative tweaks to PostgreSQL conf, based on 
+# "Loading OSM data" located at switch2osm.org.
+# TODO: figure out if it's possible to get rid of the other ID, so I can sleep better at night.
```
pnorman: Other ID?

fwiw, I've started changing postgres settings by adding `include_dir = 'conf.d'` to postgresql.conf and then dumping my custom settings in conf.d/base.conf. Probably not an issue here, but it means there's only one change in postgresql.conf if I later upgrade it.

salt/kosmtik.sls
```diff
@@ -0,0 +1,106 @@
...
+      - ttf-unifont
+      - fonts-droid
+      - fonts-sipa-arundina
+      - fonts-sil-padauk
+      - fonts-khmeros
+
+# nodejs is what kosmtik runs on
+nodejs:
+  pkgrepo.managed:
+    - humanname: NodeSource node.js
```
pnorman: the version of node in 14.04 is recent enough for kosmtik

.gitignore
pnorman: it sounds like switching between branches might cause problems then?

salt/kosmtik.sls
```diff
@@ -0,0 +1,106 @@
...
+    - depth: 1
+    - require:
+      - pkg: package-requirements
+  npm.bootstrap:
+    - name: /usr/local/kosmtik
+    - require:
+      - pkg: nodejs
+      - git: kosmtik
+  # localconfig.json -- can be edited, is stored in the home directory
+  file.managed:
```
pnorman: this shouldn't be needed with a database named gis

salt/kosmtik.sls
```diff
@@ -0,0 +1,106 @@
...
+    - name: /root/localconfig.json
+    - contents: |
+        [
+          { 
+            "where": "Layer",
+            "if": {
+              "Datasource.type": "postgis"
```
pnorman: this will override **all** postgis layers, not just ones pulling from the gis db. You want to restrict it to just changing databases named gis, if you need to change it.

salt/kosmtik.sls
```diff
@@ -0,0 +1,106 @@
...
+            },
+            "then": {
+              "Datasource.dbname": "gis",
+              "Datasource.password": "",
+              "Datasource.user": "gisuser",
```
pnorman: setting the PGUSER env var before launching kosmtik is my preferred method if there isn't a need for multiple users.

pnorman:
It looks useful and I've added some comments, but is this the right repo for this? I'm not the best judge, as I know **I'd** never use it.

Vagrantfile
thor: @pnorman Not as of yet; the trusty64 box has yet to be updated to 14.04.2. This isn't much of a problem, as .2 mainly consists of kernel and hardware support updates (beyond other typical package updates).

salt/gis.sls
thor: @pnorman I believe so, unless I give the `gisuser` a password and tell it to connect using TCP/IP, instead of a local Unix socket. A local Unix socket should give better transfer rates, so unless the user specifies otherwise in the Vagrantfile for what database and host to use, it should probably remain, unless there are other concerns.

.gitignore
thor: Yeah, you'll have to commit or stash your changes, otherwise git will get vocal. Stashing would be the most convenient, then you can just pop wherever you checkout, and potentially `git update-index --assume-unchanged Vagrantfile` if you'd like to tell git to be quiet about it.

An alternative approach is to never commit the Vagrantfile, and instead have a Vagrantfile.default or similarly that a user would have to rename to Vagrantfile. That sort of makes a simple thing somewhat more complex, but also can avoid minor things down the road.

thor:
> It looks useful and I've added some comments, but is this the right repo for this?

:+1: Glad to hear.
Also, that's a good question. I alone definitely can't answer that. What I can say is that having a simple Vagrantfile in a project allows for people to quickly contribute, as well as see what the project "does" with a higher degree of ease - generally speaking. In this case I think it's extra valuable, at least if you haven't setup the whole of PostGIS before, or your Unix knowledge is limited, while your CartoCSS and SQL experience is more useful.

If it'd make a difference I could link to other projects doing so, but it shouldn't be included on the merit of "somebody else does it!", rather if it makes sense. Potentially as mentioned in the PR, I could split it out to its own repository, but then it'd be harder to find, which again would decrease the usefulness of it.

> I'm not the best judge, as I know I'd never use it.

I'm not going to try to convince you to do so, but could you give me some pointers on what makes it less useful for you, or irrelevant, so that I could potentially improve it for other people alike?

salt/gis.sls
pnorman: if you're connecting with a local socket from the same user you shouldn't need to modify `pg_hba.conf`

thor: @pnorman True -- I'll see if it's no hassle to run kosmtik as the `gisuser`.

pnorman:
> I'm not going to try to convince you to do so, but could you give me some pointers on what makes it less useful for you, or irrelevant, so that I could potentially improve it for other people alike?

I have multiple database servers running on bare hardware. I will be running a minutely updated osm2pgsql database on my dev box. I edit multiple stylesheets.

I could virtualize Kosmtik and point it at an existing database host with PGHOST, but I wouldn't gain much out of that.

salt/kosmtik.sls
thor: Leaving these here -- I've altered it to use the configuration from the Vagrantfile, so it can be pointed towards any other host.

salt/kosmtik.sls
thor: Will correct to match against `"Datasource.type": "postgis"` and `"Datasource.dbname": "gis"`.

salt/kosmtik.sls
pnorman: Is there a reason for adjusting it here rather than PG env vars? Both will work for any other host

thor: It seems more reasonable to split it into a localconfig.json file for defining where kosmtik should connect, rather than editing it within the upstart service file for kosmtik. What sort of host are you thinking of though?

gravitystorm:
I think this is useful, but it should be managed in a separate repository. I'm happy to link to that repository in the documentation.

Like @pnorman I wouldn't use it myself so I'm reluctant to have it in the main repository - it would be misleading. I run multiple stylesheets and databases, and use multiple tools (kosmtik, tilemill) and multiple text editors - never mind compiling my own mapnik (for testing various things) and linking node-mapnik against that custom mapnik. All in all, I wouldn't want to run tilemill/kosmtik in an isolated environment.

I can see that it will be useful for other people, but it's best managed separately.

thor:
> I think this is useful, but it should be managed in a separate repository. I'm happy to link to that repository in the documentation.

Understandable. I'll close the pull request and see if I can put the files together in a way that makes it as convenient as possible to get started with it. 

When it's ready, how should I get in touch re: linking to it, if desireable?

mkoniecz:
@thor 
Maybe it would be a good idea to setup repository with alternative ways to setup environment? Maybe for some mine "notepad with instructions that I've seen linked around here" would be better (at least for me it is, as on my machine carto inside VM is really slow).

thor:
@mkoniecz Absolutely -- it's a complex combination of different components, and there's no reason why I shouldn't try to make it useful for power users in the same way you'd try to make it quick and easy for new users. :+1: 

A thought that did occur to me is that it is silly to have the PostGIS database on the same VM as mapnik/kosmtik/tilemill, simply because you'd never want to constantly be destroying and re-importing your PostGIS data -- it's time consuming and useless.

It seems the main target audience isn't you folks ( ;-) ), but even I grow tired of importing Great Britain among others! I'm setting it up so that either way, even if you're a new user, the PostGIS database and your choice of rendering/editing are on two different machines. (If you use a "cloud" provider or OpenStack provider, vSphere, KVM, etc, Vagrant can still take care of deploying the VM to your own stack. :bulb: )

math1985:
> I think this is useful, but it should be managed in a separate repository.

So I'm closing the PR. Feel free to continue discussing here about setting this up elsewhere.

pnorman:
[furry-sansa](https://github.com/Komzpa/furry-sansa) is also related, being a tool to manage importing osm2pgsql databases and set up config files. Slightly different use-case.

kocio-pl:
Just a quick note: I'm [interested in Docker](https://github.com/gravitystorm/openstreetmap-carto/issues/657#issuecomment-88671581) and I know that Vagrant 1.6+ [can use Docker images](https://www.vagrantup.com/blog/feature-preview-vagrant-1-6-docker-dev-environments.html), so probably that could be a perfect match for all the OS-es.

PR #1344, opened 2015-03-02
Sql cleanup

pnorman:
More hstore prep

### DIFF
--- a/project.yaml
+++ b/project.yaml
@@ -1719,10 +1719,12 @@ Layer:
     <<: *extents
     Datasource:
       <<: *osm2pgsql
-      table: |2-
-        (SELECT way
-        FROM planet_osm_line
-        WHERE "addr:interpolation" IS NOT NULL) AS interpolation
+      table: |-
+        (SELECT
+            way
+          FROM planet_osm_line
+          WHERE "addr:interpolation" IS NOT NULL
+        ) AS interpolation
     properties:
       minzoom: 17
     advanced: {}
--- a/project.yaml
+++ b/project.yaml
@@ -1749,12 +1749,22 @@ Layer:
     <<: *extents
     Datasource:
       <<: *osm2pgsql
-      table: |2-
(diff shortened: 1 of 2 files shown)


project.yaml
```diff
@@ -1733,12 +1735,23 @@ Layer:
...
-               ORDER BY way_area DESC
-              ) AS housenumbers
+      table: |-
+        (SELECT
+            way,
+            "addr:housenumber",
+            way_area/NULLIF(!pixel_width!::real*!pixel_height!::real,0) AS way_pixels
+          FROM planet_osm_polygon
+          WHERE "addr:housenumber" IS NOT NULL
+            AND building IS NOT NULL
```
pnorman: I'm not sure why there's the building not null check - as far as I can tell it came from the original XML

project.yaml
```diff
@@ -1733,12 +1735,23 @@ Layer:
...
-              (SELECT way,"addr:housenumber", way_area FROM planet_osm_polygon WHERE "addr:housenumber" IS NOT NULL AND building IS NOT NULL
-               UNION
-               SELECT way,"addr:housenumber", NULL AS way_area FROM planet_osm_point WHERE "addr:housenumber" IS NOT NULL
-               ORDER BY way_area DESC
-              ) AS housenumbers
+      table: |-
+        (SELECT
+            way,
+            "addr:housenumber",
+            way_area/NULLIF(!pixel_width!::real*!pixel_height!::real,0) AS way_pixels
```
pnorman: strictly speaking it's not required to return way_pixels, but it's useful if we want to start filtering by it in the MSS

project.yaml
gravitystorm: I think there's a building is not null index in production on the osmf hardware.

pnorman: Unless it's been added since we switched to carto, I don't think so. I've recommended adding a `building IS NULL` index, which is the inverse.

PR #1345, opened 2015-03-02
Always render road area names from z17 independent of size

### LINKED ISSUE #1339: Name of attraction, pedestrian area is changing with zoom
http://www.openstreetmap.org/way/95334058#map=18/50.76381/6.11586
this petting zoo is mapped as:

    highway=pedestrian
    name=Zwergziegen-Streichel-Gehege
    tourism=attraction
    attraction=animal

and rendered >zoom 18 with a black name, but on zoom 17 it transforms into the attraction font color.
I am not sure if pedestrian is a good tag, but i think footway would render the same.
Perhaps it is not worth any fine tuning.

math1985:
This resolves #1339.

### DIFF
--- a/roads.mss
+++ b/roads.mss
@@ -2366,7 +2366,8 @@ residential is rendered from z10 and is not included in osm_planet_roads. */
 }
 
 #roads-area-text-name {
-  [way_pixels > 3000] {
+  [way_pixels > 3000],
+  [zoom >= 17] {
     [zoom >= 15] {
       text-name: "[name]";
       text-size: 8;


mkoniecz:
According to my testing it is a positive chance, but note that really big [highway=pedestrian; tourism=attraction] may be rendered as tourism attraction up to z15.

Likely it is completely theoretical.

highway pedestrian area yes name eeeeee eeeeee eeeeee tourism attraction ref 1 ele 8000 closed_way master - math_minzoom-road-area

PR #1348, opened 2015-03-03
add rendering for amenity=nightclub

### LINKED ISSUE #456: Add rendering for amenity=nightclub
This was rendered before change to openstreetmap-carto. Now its gone.

This are often huge places with tons of people. This should be rendered again.

http://wiki.openstreetmap.org/wiki/Tag:amenity%3Dnightclub

nebulon42:
Adds rendering for nightclubs, resolves #456.
In order not to fully repeat the icon-related discussion please also see https://github.com/nebulon42/osmic/issues/8

before
not rendered

after
nightclub

### DIFF
--- a/amenity-points.mss
+++ b/amenity-points.mss
@@ -121,6 +121,13 @@
     marker-clip: false;
   }
 
+  [feature = 'amenity_nightclub'][zoom >= 17] {
+    marker-file: url('symbols/nightclub.18.svg');
+    marker-fill: #734a08;
+    marker-placement: interior;
+    marker-clip: false;
+  }
+
   [feature = 'amenity_fire_station'][zoom >= 16] {
     marker-file: url('symbols/firestation.16.svg');
     marker-fill: #734a08;
@@ -648,7 +655,8 @@
   [feature = 'amenity_cafe'],
   [feature = 'amenity_fast_food'],
   [feature = 'amenity_biergarten'],
-  [feature = 'amenity_bar'] {
+  [feature = 'amenity_bar'],
+  [feature = 'amenity_nightclub'] {
     [zoom >= 17] {
       text-name: "[name]";
       text-fill: #734a08;
--- a/symbols/nightclub.18.svg
+++ b/symbols/nightclub.18.svg
@@ -1,0 +1,47 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<!-- Created with Inkscape (http://www.inkscape.org/) -->
+
+<svg
+   xmlns:dc="http://purl.org/dc/elements/1.1/"
+   xmlns:cc="http://creativecommons.org/ns#"
+   xmlns:rdf="http://www.w3.org/1999/02/22-rdf-syntax-ns#"
+   xmlns:svg="http://www.w3.org/2000/svg"
+   xmlns="http://www.w3.org/2000/svg"
+   version="1.1"
+   width="100%"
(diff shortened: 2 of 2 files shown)


dieterdreist:
I think this icon it too much music centered and too generic (could be
anything from a music venue to a record store to a music school etc.).
Something suggesting dancing would be closer to the tag IMHO.

These are likely copyrighted (got them from a websearch), but I post them
to give some examples how it might look like:
-
http://www.bishopsstortford.org/sites/default/files/styles/square_thumbnail/public/Party_icon_svg.png?itok=eqzTYEke
-
http://www.termlancer.com/wp-content/themes/directory/tn.php?src=http%3A%2F%2Fwww.termlancer.com%2Fwp-content%2Fuploads%2F2014%2F09%2Ficon_25440.png&w=140&h=200
-
http://www.firstamericanmerchant.com/wp-content/uploads/2013/04/FAM_BEVEL_ICONS_CLUB.png

nebulon42:
I agree that there is a certain ambiguity involved with this symbol, but it seemed to me the most reasonable choice after discussion in https://github.com/nebulon42/osmic/issues/8.

However, regarding the dancer (because likely to come up more often): If somebody is able to draw a dancer at 14x14px (not downscaled!) that is recognisable, free of rights and is not confused with walking or running I'm more than happy to use it.

dieterdreist:
2015-03-03 11:24 GMT+01:00 nebulon42 <notifications@github.com>:

> However, regarding the dancer (because likely to come up more often): If
> somebody is able to draw a dancer at 14x14px (not downscaled!) that is
> recognisable, free of rights and is not confused with walking or running
> I'm more than happy to use it.



Skype had a nice one in their classical emoticon section (pre version 5.5),
but that is likely not free of rights...

imagico:
> I agree that there is a certain ambiguity involved with this symbol, but it seemed to me the most reasonable choice after discussion in nebulon42/osmic#8.

As a general rule and since this came up in icon design discussion here repeatedly: if creating a clear and non-ambiguous symbol within the system- and self-imposed contraints is not possible then it should not be rendered IMO - missing information is better than misleading information.

Specifically here i think the 'cocktail + music' concept is doable, a dancer might be as well although it does not really target the nightclub specifically.

nebulon42:
> Specifically here i think the 'cocktail + music' concept is doable, a dancer might be as well although it does not really target the nightclub specifically.

I'm not too fond of depicting two concepts side by side. Regarding the dancer I then suggest to do a proposal.

nebulon42:
We also discussed a disco ball, which is promising as most distinct symbol, but is hard to recognize.
Here is a preview: disco

What are your opinions about that?

dieterdreist:
not recognizable in this form, maybe with some "glitter" added?

nebulon42:
I think that anything the size of "glitter" is not really feasible at this size. Agreed, none of the options is optimal but so far I haven't seen anything that works at this size. Since an icon for nightclub was requested by @math1985 it would be good to have some guidance on how to proceed here.

math1985:
As we currently don't have a better icon, I would propose merging this. I don't think the current icon is so misleading that we shouldn't render it.

kocio-pl:
Maybe the line on the top (symbolizing supporting line) would help a bit?

math1985:
Merged as this is the best proposal we have.

If somebody wants to write a proposal for an alternative, I'd be happy to consider it.

Issue #1232, opened 2015-01-16
Hostel icon not rendered on recreation_ground


polarbearing:
tourism=hostel + landuse=recreation ground is a typical combination for bungalow villages.
http://overpass-turbo.eu/s/73y

Usually the label is written in the landuse style, thus the hostel icon is not rendered. 
In the case the map is centred to, the playground node in the centre seems to block any label or icon. 

Shouldn't *-placement = interiour also look for positions that are not in the centre?

Edit by @math1985:
* [x]    Recreation ground names have higher priority than hostel names
* [x]    The playground icon blocks the rendering of the hostel name

pnorman:
> Shouldn't *-placement = interiour also look for positions that are not in the centre?

No, I don't believe point symbolizers will try multiple locations without explicitly setting the other locations, which is not possible in CartoCSS without some hackish workarounds.

polarbearing:
Should a tourism or amenity label be written with higher priority than the landuse (in particular as they have the same size)?

math1985:
There are two issues here:
* Recreation ground names have higher priority than hostel names
* The playground icon blocks the rendering of the hostel name

math1985:
>  The playground icon blocks the rendering of the hostel name

This is duplicate with #964.

PR #1038, opened 2014-10-10
Rewrite SQL queries so the COALESCE part works correctly

math1985:
* Resolves #985
* Resolves #1029

PR #1349, opened 2015-03-04
Resolve bugs related to COALESCE problems

### LINKED ISSUE #963: amenity=fuel with shop=kiosk is rendered as unknown shop in purple instead of fuel icon
Fuel stations which are also tagged with shop=kiosk (prob. also shop=convenience and shop=yes) are rendered as purple dot as unknown shop type. For example: http://osm.org/go/0DaQVor3r

### LINKED ISSUE #985: shop=no shouldn't get rendered
shop=no shouldn't get rendered: http://www.openstreetmap.org/node/1031607377

### LINKED ISSUE #1029: tourism=attraction not rendered when object also has unrendered amenity tag
[This area](http://www.openstreetmap.org/way/243426059) is tagged `amenity=casino` and `tourism=attraction`. Normally `tourism=attraction` gets a red label ([example](http://www.openstreetmap.org/way/37492969)), but `amenity=casino` seems to override that rule, removing the label at all zoom levels. (The fuschia-ish fill color remains.) A label should appear based on the area’s area (like #941).

### LINKED ISSUE #1269: alpine_hut should have precedence over restaurant
Sometimes alpine huts are both tagged `tourism=alpine_hut` and `amenity=restaurant`, which is not wrong IMO. In those cases restaurant seems to have precedence over alpine_hut, which causes the icon to appear at z17 instead of z13. This is undesirable as huts are important POIs for hikers.

Examples: http://www.openstreetmap.org/way/176377262 or http://www.openstreetmap.org/node/3007429464

### LINKED ISSUE #1336: Unrendered man_made overrules natural
Opening a new issue after https://github.com/gravitystorm/openstreetmap-carto/issues/1060#issuecomment-75535146

Nodes with man_made=cross and natural=peak don't show up on the map.
Example: https://www.openstreetmap.org/node/26863070
More examples: http://overpass-turbo.eu/s/7OE

As @nebulon42 notes, *in the SQL  man_made  takes precedence over  natural , but since  man_made=cross  is not rendered nothing is rendered.*
Related to #1269 (amenity vs tourism)

Possible solutions I can think of: 

1 - reverse priority, render all natural over all man_made on the same node
   ...Likely to cause other issues?
1a - selectively render natural=peak over all man_made on the same node

2 - render something for man_made=cross
2a - render something special for man_made=cross and natural=peak combo

3 - more general: Have a list of things that get rendered for each key, and if the value is not on that list, look for another possible key to render. No idea if that is realistically possible, or if it would open other problems.

math1985:
* Make sure conditions are checked in both WHERE and COALESCE
  Checking the condition in the WHERE class is necessary for performance.
  Checking the condition in the COALESCE is necessary to prevent unrendered tags blocking rendering of tags further down in the COALESCE.
  See also the discussion [here](https://github.com/gravitystorm/openstreetmap-carto/pull/1038#issuecomment-58734334).

  This is an improvement of #1038.

  This resolves #985, resolves #1029, and resolves #1336.

* Give amenity priority over shop.

  This resolves #963.

* Give tourism priority over amenity, shop, leisure, landuse, man_made, natural,
  and place.

  This resolves #1269, and resolves item of #1232.


pnorman:
image

http://www.openstreetmap.org/node/960913460 with amenity=post_office and shop=variety_store

Mismatch between icon (post office) and text (shop)

math1985:
> Mismatch between icon (post office) and text (shop)

Fixed. We should probably additionally add labels for post offices.

mkoniecz:
There is a typo in comments - 'beween' -> 'between'

mkoniecz:
OK, I will make a separate PR.

PR #1351, opened 2015-03-04
Use >= rather than >

math1985:
This brings the code in line with the coding standard

### DIFF
--- a/amenity-points.mss
+++ b/amenity-points.mss
@@ -179,12 +179,12 @@
     point-placement: interior;
   }
 
-  [feature = 'amenity_courthouse'][zoom > 16] {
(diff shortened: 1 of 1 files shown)


pnorman:
I suppose the one question is was it intended to be `>= 16` or `>= 17`.

math1985:
Considering the other POI zoom levels, I think probably ''>= 16'' was intended.

Updated.

pnorman:
Should the [text zoom](https://github.com/gravitystorm/openstreetmap-carto/blob/ece00c1243b4894c8211a2ac6faf62c936b6b876/amenity-points.mss#L688) be adjusted too?

math1985:
> Should the text zoom be adjusted too?

No, we don't render labels for the other POI on z16 either.

Issue #591, opened 2014-06-02
Add rendering for amenity=bicycle_parking


Grillmannen:
Please add rendering for amenity=bicycle_parking, both nodes and areas. The icon could be the same as car parking but with a small bicycle underneath, like the JOSM "rendering".

If i remember correctly, these used to render. Areas like these look very empty with no bicycle parking rendering: http://www.openstreetmap.org/#map=19/55.86100/12.91229 .

Also, there are about 96 000 uses in the database.

nebulon42:
FYI there is now an icon available: ![parking-bicycle-18](https://cloud.githubusercontent.com/assets/3531092/6541460/718b678c-c4d2-11e4-945c-1bc7849a23e6.png)
[nebulon42/osmic/parking-bicycle-18.svg](https://github.com/nebulon42/osmic/blob/master/transport/parking-bicycle-18.svg)

mboeringa:
I just hope people aren't going to "tag for the renderer" with this one in the Netherlands... With +20 Million bikes or so, and about every tree, lamp- or sign pole in cities used as "bicycle_parking" around here, I could see this icon going "viral" ;-)

Grillmannen:
That kind of usage is not the kind that is normally called "tag for the renderer"... Actually, the opposite would be true. If people would avoid tagging bicycle_parkings because the map would be congested with icons, that would be mapping for the renderer, and that is a renderer problem.

mboeringa:
> Actually, the opposite would be true. If people would avoid tagging bicycle_parkings because the map would be congested with icons, ***that would be*** **mapping** ***for the renderer***

You are right... ;)

mkoniecz:
> I just hope people aren't going to "tag for the renderer" with this one in the Netherlands... With +20 Million bikes or so, and about every tree, lamp- or sign pole in cities used as "bicycle_parking" around here, I could see this icon going "viral" ;-)

see http://wiki.openstreetmap.org/wiki/Tag:amenity%3Dbicycle_parking 

> The term ‘bicycle parking’ refers to any parking space designed for bicycles, where one can leave a pedal cycle unattended in reasonable security.

Though on http://wiki.openstreetmap.org/wiki/Key:bicycle_parking there is bicycle_parking=informal but even this is restricted to "Notably large or well-used railings, fences, or other street furniture(...)".

matkoniecz:
[...]

PR #1353, opened 2015-03-04
amenity=bicycle_repair_station symbol (public 24/7 tools for fixing bikes)

brycenesbitt:
An amenity=bicycle_repair_station is usually available 24/7,and hosts tools for repairs of bicycles.  There are now over 500 mapped.  Many people are not aware these stations even exist, thus the mapping.

The svg file is licence compatible.


mkoniecz:
Please, add before/after images.

pnorman:
Does Mapnik even support all the stuff you're using in that SVG?

symbols/bicycle_repair_tools.svg
```diff
@@ -0,0 +1,150 @@
...
+     id="g3138"
+     transform="matrix(0.98757845,0,0,1,-171.61601,7.0110373)"
+     inkscape:export-xdpi="4.2362185"
+     inkscape:export-ydpi="4.2362185">
+    <circle
+       transform="matrix(0.612193,0,0,0.612193,156.40859,178.29391)"
+       r="161"
+       cy="409"
+       cx="215"
+       style="fill:none;stroke:#000000;stroke-width:40px;stroke-opacity:1"
```
pnorman: I've had problems with stroke before in Mapnik

pnorman:
Regardless of the technical issues with the SVG, I'm against adding `amenity=bicycle_repair_station`. I feel it's too specialized and is similar to some other bicycle-specific features we've decided not to render (#591).

I'd like to see the example images covering the proposed rendering in a variety of settings.

mkoniecz:
> is similar to some other bicycle-specific features we've decided not to render (#591).

In case of #591 I think that current situation is that there is no consensus for small ones, but gigantic ones should be somehow rendered.

In this case I am unsure what counting my positive bias toward anything cycling-related may be counted as being against rendering amenity=bicycle_repair_station.

math1985:
I see no harm in rendering this, but only from high zoom levels to avoid clutter (z18 or z19). I think z17 is too early for a general purpose map.

nebulon42:
@brycenesbitt re consistency: Here is my suggestion for bike repair that is consistent with bike rental (which is already in master) repair-bicycle-18 [nebulon42/osmic/repair-bicycle-18.svg](https://github.com/nebulon42/osmic/blob/master/transport/repair-bicycle-18.svg)

Here is a version without halo (preferred by maintainers): https://gist.githubusercontent.com/nebulon42/bcae5a3d535ff862a99a/raw/6dda54bc284e36fb430dd6e130ed6a2b668da92d/repair-bicycle-16.svg

math1985:
I think 500 items is still too little for it too deserve its own icon. Maybe in the future! Thanks anyway @brycenesbitt!

brycenesbitt:
This is a worldwide feature, growing, with the capability of drawing new mappers to the project.  A fraction of the existing stations are mapped, but the feature will still be far more rare than toilets or bike parking:
selection_189

kocio-pl:
Now the number of uses is much higher (2800), the new icon has been proposed in a duplicate ticket, but we already have it in Osmic: 

https://github.com/gmgeo/osmic/blob/master/shop/repair-bicycle-14.svg

kocio-pl:
Oh, sorry - it's the old PR, not an issue...

PR #1354, opened 2015-03-04
Match label colors to icons

math1985:
Match the color of the following labels to their respective icons:

* natural=peak
* natural=volcano
* natural=saddle
* natural=cave_entrance
* historic=memorial
* historic=archaeological_site
* man_made=lighthouse

### DIFF
--- a/amenity-points.mss
+++ b/amenity-points.mss
@@ -755,7 +755,7 @@
     [zoom >= 13] {
       text-name: "[name]";
       text-size: 10;
-      [feature = 'natural_peak'] { text-fill: #d08f55; }
+      [feature = 'natural_peak'] { text-fill: darken(#d08f55, 30%); }
       [feature = 'natural_volcano'] { text-fill: #d40000; }
       text-dy: 7;
       text-face-name: @book-fonts;
@@ -769,7 +769,7 @@
     [zoom >= 15] {
       text-name: "[name]";
       text-size: 10;
-      text-fill: #d08f55;
+      text-fill: darken(#d08f55, 30%);
       text-dy: 7;
       text-face-name: @book-fonts;
       text-halo-radius: 1;
--- a/amenity-points.mss
+++ b/amenity-points.mss
@@ -755,7 +755,8 @@
     [zoom >= 13] {
       text-name: "[name]";
       text-size: 10;
-      text-fill: brown;
+      [feature = 'natural_peak'] { text-fill: #d08f55; }
+      [feature = 'natural_volcano'] { text-fill: #d40000; }
       text-dy: 7;
       text-face-name: @book-fonts;
       text-halo-radius: 1;
@@ -768,7 +769,7 @@
     [zoom >= 15] {
       text-name: "[name]";
       text-size: 10;
-      text-fill: brown;
+      text-fill: #d08f55;
       text-dy: 7;
       text-face-name: @book-fonts;
       text-halo-radius: 1;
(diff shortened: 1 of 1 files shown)


math1985:
Before:
screenshot from 2015-03-04 13 45 47

After:
screenshot from 2015-03-04 13 43 49

mkoniecz:
Maybe also caravan site should be changed?

imagico:
I think the peak/saddle label is too weak to be well readable.

math1985:
> Maybe also caravan site should be changed?

For caravan site, see #1355.

math1985:
The peaks are readable for me, but I could make them darker. This is how they look now:

screenshot from 2015-03-06 18 17 51

mkoniecz:
I think that for peaks and saddles old version is better. This is barely readable - and what about peaks on natural=bare_rock areas?

imagico:
Having peak/volcano the same color except for the symbol itself (as it is now) seems a well working concept to me.  Reducing the color difference between the peak/saddle symbol and the label would not hurt but with your current setting the label appears even brighter (due to the thin lines and the white casing) than the symbol.  At the same time the symbol color could be somewhat stronger/darker IMO - to compensate the symbol could be slightly reduced in size for z=11/12.

Here your current color on different backgrounds

sample 1
sample 2
sample 3
sample 4

mkoniecz:
For me it is not really readable. It is possible to decipher it, but it requires effort that should not be needed.

Maybe new color for peak symbol/label may be something between current symbol and current label colour?

pnorman:
I also prefer the current colour to the proposed one.

mboeringa:
I would simply darken the color for the label, and stay with the same adjusted hue. I am pretty sure that will solve the readability issue. There is no reason why a label could not have a darker color than the icon, as long as the hue stays the same. I think the recognizability of the combination will not suffer.

Any specific reason why playground and water_park aren't adjusted as well? (I noticed the playgrounds don't have labels yet... but for consistency maybe?)

HolgerJeromin:
playground have a label: http://osm.org/?mlat=50.78457&mlon=6.05799#map=19/50.78457/6.05799

mboeringa:
> playground have a label: http://osm.org/?mlat=50.78457&mlon=6.05799#map=19/50.78457/6.05799

Then I was unlucky to only look at a couple of ones actually missing a name tag...

math1985:
> Any specific reason why playground and water_park aren't adjusted as well? (I noticed the playgrounds don't have labels yet... but for consistency maybe?)

This depends on #1355, for which there is still no clear preferred option.

math1985:
I darkened the labels.

screenshot from 2015-03-09 21 00 34

mboeringa:
That looks a lot better...

mkoniecz:
Yes, it is now much better.

pnorman:
Looks good - I'll try for a full review tomorrow.

pnorman:
Peaks start at z13 in production with text at z13, but on this branch they start at z12 with text at z14.

Is this intentional, or is there something else that has been merged in and not released yet?

math1985:
> Peaks start at z13 in production with text at z13, but on this branch they start at z12 with text at z14.

Are you sure? As far as I can see, in both production and branch, icons start at z11 and text at z13.

mkoniecz:
> Are you sure? As far as I can see, in both production and branch, icons start at z11 and text at z13.

The same happens according to my test.

natural peak name eeeeee eeeeee eeeeee ref 1 ele 8000 node master - math_text-colors

math1985:
> The same happens according to my test.

If I understand @mkoniecz's picture correctly, 'the same' refers to the same as I reported.

mkoniecz:
> If I understand @mkoniecz's picture correctly, 'the same' refers to the same as I reported.

Yes, at least for lone peaks only label color is changed.

PR #1359, opened 2015-03-06
New path rendering

sb12:
(See #1327 for old discussion. Unfortunately something went wrong while rebasing and I had to do a new pull request)
I've recently seen a lot of issues regarding cycleways, footways and stairs, so I thought what about a complete new rendering for paths similar to the German OSM Style.

This pull requests changes the following:

- Paths are not rendered as thin lines/dots any more, but similar to roads with a width.
- Footways and stairs are not rendered in "Salmon" anymore, but in the same color as pedestrian areas.
- ~~Paved paths are rendered same as footways~~
- Paths that are shared by bicycles and footways (path with bicyle=designated and foot=designated, footway with bicycle=yes, cycleway with foot=yes) are rendered with a blue grey dashed outline from zoom level 17
- Cycleways are not rendered special any more on zoom level 15 and lower

This solves the following issues:

- solves #1321 Rendering cycleway with pedestrians
- solves #1307 highway=steps color change
- solves #1248 Admin borders look like paths ... and probably some more

Some possible issues with this style:
- Tracks should be adapted as well for consistency
- Service roads have the same width as paths

Before and after renderings:

Low zoom levels:

npath_z13_schwetzingen
npath_z13_schwetzingen_after_new

npath_z14_schwetzingen
npath_z14_schwetzingen_after_new

forest:
npath_z14_mannheim
npath_z14_mannheim_after_new

Cemetery and allotments:
npath_z15_cemetery
npath_z15_cemetery_after_new

Higher zoom levels:
npath_z16_heidelberg
npath_z16_heidelberg_after_new
npath_z17_schlosskarlsruhe
npath_z17_schloss_karlsruhe_after_new
npath_z18_fasanengarten
npath_z18_fasanengarten_after_new

### DIFF
--- a/roads.mss
+++ b/roads.mss
@@ -13,8 +13,7 @@
 @raceway-fill: pink;
 @road-fill: #ddd;
 @path-fill: black;
-@footway-fill: salmon;
-@steps-fill: @footway-fill;
+@steps-fill: #aaa;
 @cycleway-fill: blue;
 @bridleway-fill: green;
 @track-fill: #996600;
@@ -34,9 +33,9 @@
 @living-street-casing: @residential-casing;
 @pedestrian-casing: @residential-casing;
 @path-casing: @default-casing;
-@footway-casing: @default-casing;
+@footway-casing: #aaa;
 @steps-casing: @default-casing;
-@cycleway-casing: @default-casing;
+@cycleway-casing: blue;
 @bridleway-casing: @default-casing;
 @track-casing: @default-casing;
 
@@ -56,6 +55,7 @@
 @tertiary-tunnel-fill: lighten(@tertiary-fill, 5%);
 @residential-tunnel-fill: darken(@residential-fill, 5%);
 @living-street-tunnel-fill: lighten(@living-street-fill, 10%);
+@pedestrian-tunnel-fill: lighten(@pedestrian-fill, 3%);
 
 @motorway-width-z12:              3.5;
 @motorway-link-width-z12:         1.5;
@@ -74,8 +74,7 @@
 @living-street-width-z13:         2;
 @pedestrian-width-z13:            2;
 @bridleway-width-z13:             0.3;
-@footway-width-z13:               0.7;
-@cycleway-width-z13:              0.7;
+@footway-width-z13:               0.1;
 @path-width-z13:                  0.2;
 @track-width-z13:                 0.5;
 @track-grade1-width-z13:          0.5;
@@ -86,6 +85,7 @@
 @residential-width-z14:           4.5;
 @living-street-width-z14:         4;
 @pedestrian-width-z14:            4;
+@footway-width-z14:               0.5;
 @service-width-z14:               2.5;
 
 @motorway-width-z15:             12.8;
@@ -97,18 +97,19 @@
 @residential-width-z15:           8.3;
 @living-street-width-z15:         6;
 @pedestrian-width-z15:            6;
-@bridleway-width-z15:             1.2;
-@footway-width-z15:               1.5;
+@bridleway-width-z15:             2;
+@footway-width-z15:               1.2;
 @cycleway-width-z15:              1.2;
 @path-width-z15:                  0.5;
 @track-width-z15:                 1.5;
 @track-grade1-width-z15:          0.75;
 @track-grade2-width-z15:          0.75;
-@steps-width-z15:                 4;
+@steps-width-z15:                 3;
 
 @residential-width-z16:          11.2;
 @living-street-width-z16:         9;
 @pedestrian-width-z16:            9;
+@footway-width-z16:               3;
 @service-width-z16:               6;
 @minor-service-width-z16:         4;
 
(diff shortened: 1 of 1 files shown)


math1985:
I'm on two minds about this change. I think visually the new style is much prettier. The main risk is that footpaths are now much harder to distinguish from other types of roads, such as service roads.

What do @gravitystorm and @pnorman think?

pnorman:
I haven't loaded up and looked around locally yet, but will do so.

kocio-pl:
> The main risk is that footpaths are now much harder to distinguish from other types of roads, such as service roads.

I think in a micro- scale it would probably make sense! The service roads and the footpaths (for both of which this scale is the most common one) are not totally exclusive - their space is shared by different uses:

1. People walk freely over service roads in different directions (hence no formal crossings there!).

2. Pedestrian ways, especially asphalt-covered in some housing estates/residential areas, are often accessible by cars (I have a lot of problems to pick the right tagging there, especially when armchair mapping - exactly because of this "multipurposeness").

3. Many sidewalks are used as a car parking space.

So this is closer to the "ground truth" than we expected... =}

scaidermern:
> The main risk is that footpaths are now much harder to distinguish from other types of roads, such as service roads.
That's also my only point of criticism. The new style looks pretty but makes it hard to distinguish footways from roads. Can you try to draw them slightly thinner?

HolgerJeromin:
perhaps we can remove the chasing to be able to have it thinner. This could be a visual difference to roads. Not sure if this will work

roads.mss
```diff
@@ -1400,7 +1443,7 @@ residential is rendered from z10 and is not included in osm_planet_roads. */
         [zoom >= 15] { line/line-width: @path-width-z15; }
       }
     }
-
+
```
pnorman: whitespace error

pnorman:
The casing for paths on z16 seems a bit strong.
image

z15 on forests seems weak
image

I find it jarring next to paths
image

In isolation I think the changes are mostly okay, but when combined with our path and track rendering, I'm not so sure that it fits in.

kocio-pl:
@pnorman I think that this is not as bad as it seemed:

1. They're just visible (just like the streets are visible) and that's good, because they're important here. If we make it lighter, they could disappear. And don't you think red lines we have now are much stronger and this change would help with that?

2. I agree:
- could adding a glow make a difference there?
- maybe on some lower levels they should be rendered as before?

3. No more than usual highway and track, so this may be another problem (how to change rendering of paths and tracks in the future). However I feel path is different enough so that it can look different.

ftrebien:
Oh, it's just been pulled... How about exchanging the colour of living_street (dark gray) and pedestrian ways (light gray), would that make pedestrian areas too dark? If not, footways could be dark gray like the new pedestrian ways, which would make them more visible against vegetation and residential areas. It would also make sense for drivers: anything dark gray (pedestrian ways and footways) would not be driveable. Living streets, which are driveable, would look more similar to residential ways, which are also driveable. Dark gray would make footways remarkably different from service ways.

matkoniecz:
I was experimenting with it and main problem was that it still seemed to be too close to roads. In my tests most people claimed that it looks better but dotted pattern is more footway-like and less road-like. Dark grey footways were hard to notice in forests and lighter ones keep disappearing on residential areas. Also, grey pedestrian areas are problematic. It is either too close to white (roads), grey (of residential areas), brown (of buildings) or really dark.

las wolski master-_googlify 800px master - googlify
living street many master-_googlify 800px master - googlify
many living streets master-_googlify 800px master - googlify

I have one version that is quite promising - what is quite interesting started as a joke, so many promising ideas failed, lets try something absurd. To my surprise in tests it had really good opinions among most people. But it requires change of landuse=railway/industrial colors, I am currently tweaking this part.

las wolski master-_purple 800px master - purple
living street many master-_purple 800px master - purple
many living streets master-_purple 800px master - purple
rural area where highway footway are important master-_purple 800px master - purple
rural area where highway footway are important master-_googlify 800px master - googlify

ftrebien:
Your changes look very intuitive to me, really!

Railway and industrial areas kind of remind me of highways, so how about changing the colour hue to a little bit closer to orange? Not too much in order to avoid confusion with farms (unless their colour is changed a little to a little bit closer to green). Another way to increase contrast with purple might be to make these red areas a little bit grayer.

matkoniecz:
Moving industrial/railway landuse to Lch(89,9,104) nicely solves conflict with new footways - but it is closer than before to residential, garages and school landcover.

000060_landuse industrial closed_way 50 1 19 9 16 16 master - purple

kocio-pl:
School may be less of a problem since it looks like [we may end up with dropping filling color for educational areas](https://github.com/gravitystorm/openstreetmap-carto/pull/1450#issuecomment-117227190) (boundary color will take over). Residential is sufficiently different on this example for me, so the garages are the only one left - but they are much less important, so we may change them too. However your second example was not uploaded properly, so I can't tell for sure (I guess there was landuse=garages probably?).

matkoniecz:
Unfortunately there is no consensus that either this version or versions based on that idea are superior to current footway rendering...

Therefore this PR will not be merged - but thanks for idea and code!

I really like this rendering - but unfortunately...

PR #1363, opened 2015-03-10
Render implied oneways (roundabout, motorway, motorway_link)

### LINKED ISSUE #467: Render junction=roundabout like oneway=yes
According to the wiki junction=roundabout implies oneway=yes.
Therefore, I propose to render all highways with junction=roundabout and oneway!=no in the same way as oneway ways.

sommerluk:
Some tags imply oneway=yes. This PR renders them as if oneway=yes would be present. (Condition: There is no oneway=no present.)

junction=roundabout
highway=motorway
highway=motorway_link

This resolves #467 

Are there any other tags that imply oneway=yes?


project.yaml
```diff
@@ -420,7 +420,10 @@ Layer:
                 CASE WHEN service IN ('parking_aisle', 'drive-through', 'driveway') THEN 'INT-minor'::text
                   ELSE 'INT-normal'::text
                 END AS service,
-                CASE WHEN oneway IN ('yes', '-1') THEN oneway ELSE NULL END AS oneway,
+                CASE WHEN oneway IN ('yes', '-1') THEN oneway
```
pnorman: if turning this into a multi-line CASE then have each WHEN on its own line

project.yaml
```diff
@@ -420,7 +420,10 @@ Layer:
...
+                  WHEN (highway IN ('motorway', 'motorway_link') OR junction IN ('roundabout')) AND (oneway <> 'no' OR oneway IS NULL) THEN 'yes'
+                  ELSE NULL
+                  END AS oneway,
```
pnorman: have we been tabbing the END for a case the same as the first line of it or the same as the contents of it?

sommerluk: In the current project.yaml there are both variants.

mkoniecz:
List of tags is consistent with https://github.com/bhousel/iD/blob/master/js/id/core/oneway_tags.js

mkoniecz:
But is contradicted by http://wiki.openstreetmap.org/wiki/Tag:highway%3Dmotorway_link#Tagging_oneway "Most motorway_link roads will be one way, and should be tagged oneway=yes. Any unusual motorway link road which is two-way should be explicitly tagged oneway=no.".

sommerluk:
> "Most motorway_link roads will be one way, and should be tagged oneway=yes. Any unusual motorway link road which is two-way should be explicitly tagged oneway=no."

This seems to requiere _always_ a oneway=* tag (either “yes” or “no” or “-1”). But for me, this sentence seems to make no assert on the default value.

On the other side, the ValueDescription box on the wiki page lists oneway=yes in the section “Implies”.

How should we proceed?

dieterdreist:
2015-03-12 14:44 GMT+01:00 Lukas Sommer <notifications@github.com>:

> "Most motorway_link roads will be one way, and should be tagged
> oneway=yes. Any unusual motorway link road which is two-way should be
> explicitly tagged oneway=no."
>
> This seems to requiere *always* a oneway=* tag (either “yes” or “no” or
> “-1”). But for me, this sentence seems to make no assert on the default
> value.
>
> On the other side, the ValueDescription box on the wiki page lists
> oneway=yes in the section “Implies”.
>
> How should we proceed?
>



IMHO there is no such implication for oneway=yes. Have a look at actual
tagging, 94% of all motorway_links have an additional tag oneway=yes, if
this was an automatic implication, mappers won't add this tag. In
preliminary mapping, motorways often get mapped as one way rather than
mapping 2 oneway carriageways (OK, this is preliminary, and maybe we're
over this phase for most parts of the world).

math1985:
98% of the motorways and 99% of the motorway links have a oneway tag. So I don't think adding implicit oneway=yes arrows is useful here.

27% of the roundabouts has a oneway tag. In this case, adding the arrows might be useful.

sommerluk:
Okay. I propose to close this PR and open a new PR only for roundabouts.

What should I do with the indent for the “END AS oneway,” line?

project.yaml
pnorman: I'd probably do it the same as the former.

sommerluk:
Closing in favor of #1367

PR #1361, opened 2015-03-09
Add rendering for bicycle_parking

math1985:
* Add bicycle_parking icon from z18 or (for areas) 750 pixels, whichever is
  earlier.
* Add yellow background for bicycle_parking (like regular parking).

Thanks to @nebulon42 for the icon.

This resolves #591 and part of #108.

PR #1364, opened 2015-03-11
Add rendering for amenity=bicycle_parking, change rendering for amenity=parking

matkoniecz:
Improved version of #1361 
- unified rules for amenity=parking and amenity=bicycle_parking
- small parkings are in low-priority layer what fixes problems with bicycle parking next to bus station blocking an important label
- reduces undue prominence of amenity=parking
- icons for big parkings are rendered earlier
- labels of small parkings are rendered alter or for really small ones not rendered

Still work in progress - I am reviewing how thresholds work for various areas and it will be necessary to squash commits.

Comparisons below, unless noted otherwise are made with rendering from #1361 

https://cloud.githubusercontent.com/assets/899988/6600231/b8244224-c80e-11e4-8845-2b78425c7de2.png
https://cloud.githubusercontent.com/assets/899988/6600204/957a1a14-c80e-11e4-932c-f75ac788177d.png
https://cloud.githubusercontent.com/assets/899988/6600201/93096b86-c80e-11e4-955b-100544e5e8e5.png


mkoniecz:
I am considering moving all parking symbols into low priority layer.

Also, here is example for label vs bicycle_parking:

https://cloud.githubusercontent.com/assets/899988/6600361/7789f122-c80f-11e4-8739-75ddeadad95d.png

kocio-pl:
Looks perfect for me!

I would also love to move all parking symbols from z16 to z18. If the parking is big enough, we always see the yellow area, so it's hard to miss, but P symbol should not be bigger than small areas (like 5 places on the street), so z17 is still not enough:

http://www.openstreetmap.org/#map=17/52.22925/20.99511

SomeoneElseOSM:
(re parking symbols at z18) isn't there a danger here that the rendering is being tailored to work only in micro-mapped central european cities?

For example, for my own use I changed the rendering of the "P" symbol to z13 so that if you're looking at a broad countryside area, you can see where to park:



Obviously z13 is far too low a zoom level for a more general purpose map - I wouldn't suggest it for that.  But z18 is surely too far in the other direction.

mkoniecz:
> For example, for my own use I changed the rendering of the "P" symbol to z13 so that if you're looking at a broad countryside area, you can see where to park:

Can you link this location? Seems to be a well-mapped area with low density of features, I was looking for such place for a long time.

SomeoneElseOSM:
(location linked in previous comment)

mkoniecz:
@SomeoneElseOSM 

> re parking symbols at z18

In low density areas parking symbols would appear from z17. Areas are anyway visible much earlier.

> isn't there a danger here that the rendering is being tailored to work only in micro-mapped central european cities?

I am currently rendering my other test locations, list of where I am testing is available at https://github.com/mkoniecz/CartoCSSHelper/blob/master/CartoCSSHelper.rb (function test_tag_on_real_data). I would welcome suggestions for additional well-mapped places with different situation.

mkoniecz:
> Edit: You mean just a bicycle parking?

In this pull requests I am proposing applying the same rules for parkings and bicycle parkings.

> http://www.openstreetmap.org/#map=17/52.22925/20.99511

In that case parkings would no longer block shop icons and parking icons would disappear from z17.

mkoniecz:
@SomeoneElseOSM 

> Edit: The location is here: http://www.openstreetmap.org/#map=13/53.2127/-1.8103

Thanks, added to my list. In that case parking was mapped as node so it would appear from z17 instead current z16. I mapped it as area, but it is not going to change anything as it is not covering significant area.

RobJN:
Surely parking in the city is what we should prioritise the rendering to. In the countryside you can just pull up on the side of the road (assuming it's a safe location) as there aren't all the parking restrictions that you have in a city.

math1985:
@mkoniecz Can you try to find a version of the svg without halo?

nebulon42:
https://github.com/nebulon42/osmic/blob/8f7aebf8dd6980ca57f11ccfe92ba1deeb444c06/transport/parking-bicycle-16.svg

matkoniecz:
OK, now it is prepared for review. Is anybody interested in seeing rendering of some particular location?

pnorman:
assigning to @gravitystorm - he came to the conclusion not to render in https://github.com/gravitystorm/openstreetmap-carto/issues/591#issuecomment-44988453

math1985:
Looks good to me.

Maybe add a comment to explain why the parking icon rule is duplicated? Or even better, avoid the duplication entirely by doing something like

.amenity-points[amenity = 'parking'][zoom >= 17],
.amenity-low-priority[amenity = 'parking'][way_pixels > 750] {

pnorman:
You've mentioned changing the amenity=parking rendering but I don't see changes - what are they?

matkoniecz:
@pnorman 

> You've mentioned changing the amenity=parking rendering but I don't see changes - what are they?

currently:

Render amenity=parking icons from zoom 16.

after this PR:

Render amenity=parking, amenity=bicycle_parking icons for areas larger than 900 pixels and display all amenity=parking, amenity=bicycle_parking icons from zoom 17 as low priority icons.

Also, display labels for parkings for areas larger than 900 pixels.

Intent of this change is to lower prominence of parkings, especially in areas where are other more important features, render earlier icons for gigantic parking. Also, to render amenity=bicycle_parkings, especially really big ones,

PR #1367, opened 2015-03-12
Render implied “oneway” on roundabouts

### LINKED ISSUE #467: Render junction=roundabout like oneway=yes

sommerluk:
Treat roads with “junction=roundabout” like oneways, as long as no “oneway=no” is present.

Resolves #467 

Substitutes #1363


math1985:
Looks fine to me.

mkoniecz:
What about oneway=reversible? It is mentioned on wiki[1] and is used (rarely, 900 instances but it is). This pull requests would result in rendering oneway arrows for such ways.

[1]http://wiki.openstreetmap.org/wiki/Tag%3Aoneway%3Dreversible#Time_conditional_restrictions

More broadly, I am unsure about benefit of implicit values in that case.

sommerluk:
> What about oneway=reversible?

Good catch!

Currently there is no element with both (junction=roundabout AND oneway=reversible) in the database, but the PR should be adapted.

I assume also the most combinations of junction=roundabout AND oneway=yes are tagging errors, but that’s another topic.

However, I think first there should be made a decision about this:
– If it is decided that implied oneway on roundabout is rendered, I can adapt this PR.
– If it is decided that implied oneway on roundabout is not rendered, than issue #467 should be closed.

math1985:
I see both the advantages and disadvantages of rendering this, so I'm rather neutral.

pnorman:
> I assume also the most combinations of junction=roundabout AND oneway=yes are tagging errors, but that’s another topic.

Why would you assume this? An explicit oneway is preferable.

sommerluk:
Oh, sorry, I did want to say “I assume also the most combinations of junction=roundabout AND oneway=no are tagging errors”.

math1985:
> Oh, sorry, I did want to say “I assume also the most combinations of junction=roundabout AND oneway=no are tagging errors”.

Many of them are tagging mistakes, but this tag combination is also quite common and correct on bicycle paths in the Netherlands.

sommerluk:
I’m also rather neutral (I don’t have a personal interest in this, I just saw #467 and thought that this is something that’s easy to implement.)

Rendering implied oneway on roundabouts but not on motorways would mean rendering implied oneways for some tags, but don’t do it for other tags – which could maybe also be confusing. So maybe it’s better to omit this PR (and close also #467 as WONTFIX) …

math1985:
Given that oneway=yes is often omitted, it's clear that that's the default, so I think treating oneway=no in our code too makes sense.

@sommerluk If you add the reversible case I will merge this.

sommerluk:
@math1985 Done.

matkoniecz:
@sommerluk 
It would be a good idea to rebase this PR.

sommerluk:
Sorry. I failed in doing a rebase.

Follow-up PR at #1393

math1985:
> I failed in doing a rebase.

For the future, something like this should work:
only one
```
git remote add upstream https://github.com/gravitystorm/openstreetmap-carto.git
```
and then
``
git checkout master
git pull upstream master
git checkout name_of_the_branch
git rebase master
git push -f origin name_of_the_branch
```

sommerluk:
@math1985 Thanks a lot! I’ve tried this with an old branch and it worked. So I hope in the future I’ll do correct rebasing. (Git is sometimes really, really complicate …)

PR #1376, opened 2015-03-14
Do not render intermittent rivers on low zoom levels

### LINKED ISSUE #1305: intermittent rivers at low zooms
The recent addition of special rendering of intermittent rivers (#1000) only uses the dashed rendering at z>=12, from 8 to 11 intermittent rivers are rendered like normal rivers as solid lines.  Not sure if this is intentional but it does not seem like a good idea.  An intermittent river is sort of a partial river, both in meaning and the way it is rendered as a dashed line.  First showing the river in full (at z>=8) and then showing it only partially (at z>=12) is irritating.

If the dashed rendering is considered unsuitable for the lower zooms i would suggest a lighter or semitransparent color.

sommerluk:
This resolves #1305

### DIFF
--- a/project.mml
+++ b/project.mml
@@ -204,7 +204,7 @@
       "srs": "+proj=merc +a=6378137 +b=6378137 +lat_ts=0.0 +lon_0=0.0 +x_0=0.0 +y_0=0.0 +k=1.0 +units=m +nadgrids=@null +wktext +no_defs +over",
       "Datasource": {
         "extent": "-20037508,-20037508,20037508,20037508",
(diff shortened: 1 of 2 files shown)


imagico:
I don't think this is a very good approach - it would discourage people from properly tagging large intermittent rivers since it would not show them (at low zooms) rather than showing them in a distinct form.

For example in the interior of Australia nearly all major rivers are intermittent - which however does not make them any smaller or less important as structuring and defining elements of the geography than a permanent river.

sommerluk:
Substituted by #1399

PR #1377, opened 2015-03-14
move car repair icon to SVG

nebulon42:
Implements part of #1165.

before
car_repair_before

after
car_repair_after


matkoniecz:
New icon is much better, in old it is looks like car and mop.

math1985:
Looks good.

kocio-pl:
@matkoniecz: +1

PR #1379, opened 2015-03-14
move butcher icon to SVG

nebulon42:
Implements part of #1165.

before
butcher_before

after
butcher_after


matkoniecz:
In that case I prefer the current icon, new symbol looks for me like a credit card.

kocio-pl:
I could recognize this icon only afterwards if the name was distinctive/familiar. Probably something like 2.1 from here:

http://cdn.vectorstock.com/i/composite/11,52/black-meat-icons-set-vector-1791152.jpg

would be even better (the emblematic meat itself, not how it's done...).

matkoniecz:
@kocio-pl Is it happening also for a current icon?

kocio-pl:
Do I recognize the current one? Yes, I do. Or what are you asking me?

nebulon42:
Butcher is not ready for merge yet.

pnorman:
> In that case I prefer the current icon, new symbol looks for me like a credit card.

I also saw it as a credit card before reading your comment.

> @kocio-pl Is it happening also for a current icon?

I didn't see the current icon as a credit card. The curve might be part of that. I like the hollow inside more too.

matkoniecz:
> Butcher is not ready for merge yet.

Please, make a new pull once it will be ready.

PR #1380, opened 2015-03-14
move DIY shop to SVG

nebulon42:
Implements part of #1165.

before
diy_before

after
diy_after

(this particular shop is tagged incorrectly, but nevertheless useful as preview)


matkoniecz:
In that case I am not recognizing new symbol on the right. Is it a soldering iron?

nebulon42:
Yes, that is a bit tricky :). It should be a screwdriver.

math1985:
A nut might also be a possible visualization.

kocio-pl:
For me it's 50/50 - I'm not sure if I would recognize it not knowing what this icon represents. For me it's also too "square" (like new hairdresser proposition).

HolgerJeromin:
probably the hammer could be rounder on one side.

neiljp:
While I really do appreciate the new designs and idea of transitioning to SVG, sometimes the older designs just do work better, for me at least.
IMO the previous 'claw' hammer design was definitely more distinctive and DIY-like.
I suppose DIY can imply electrical, so a screwdriver could be valid, but a spanner is a more recognisable and drawable shape.

pnorman:
I prefer the old one also.

matkoniecz:
> It should be a screwdriver.

It is not recognizable. Can you try something different for tool on the right?

nebulon42:
What about this alternative? doityourself-183

matkoniecz:
I like it.

kocio-pl:
It's better now, but I would still like the right one's head to be more distinctive (maybe body slimming would be enough).

nebulon42:
This is the most I could get out of it: doityourself-183

More is not possible.

math1985:
Much better. Maybe the hammer would be clearer without the top pixel?

What about the bolt idea? Something like [this](https://cdn3.iconfinder.com/data/icons/pyconic-icons-3-1/512/bolt-mech-128.png)?

nebulon42:
I somehow prefer the top-stuff added even if it is slightly blurrier. Without: doityourself-183

If you mean depicting the bolt like that: It is (almost) square which makes it difficult to depict alongside the hammer. Either it will be too small or too wide.

math1985:
No, I meant the bolt on its own. Not sure if it's iconic enough to be recognizable as DIY though.

nebulon42:
@math1985 While I find the use of one symbol rather than two desirable I'm not sure if this would be recognisable enough. For simplicity reasons I decided to stick with the icon which I already had.

Updated. From my side this is now ready.

PR #1381, opened 2015-03-14
move florist icon to SVG

nebulon42:
Implements part of #1165.

before
florist_before

after
florist_after


matkoniecz:
Really nice, new icon is much better.

kocio-pl:
The top is great, but I would make leaves thinner on the bottom (like two arches together, not that soft edge).

nebulon42:
I guess you mean something like that: florist-182
Which is actually more close to how tulip leaves are shaped, yes.

kocio-pl:
Yes, that one would be good for me.

math1985:
I also prefer the tulip.

nebulon42:
Updated icon and preview.

Issue #705, opened 2014-07-06
Add rendering for amenity=fountain


kocio-pl:
Hi,

I'd like to know what is the current state of affairs with the fountain rendering?

We have reasonable proposition to fix this issue (https://github.com/gravitystorm/openstreetmap-carto/pull/47) - there is the code and even the icon (https://github.com/Rovastar/openstreetmap-carto/blob/e5320dbd98981fdeb1944771d54d6984b4bf4b73/symbols/fountain.png). I know the original reporter mixed too many things in one basket, but this one should be easy to pick up and merge - almost no-brainer.

So what are technical - or maybe other - obstacles keeping this issue unresolved and what should be done to fix it? I want to help if it's needed and possible.

nebulon42:
Possible icon for fountain: ![fountain-18](https://cloud.githubusercontent.com/assets/3531092/6652543/942545c4-ca72-11e4-805e-e1e70419de35.png)
[nebulon42/osmic/outdoor/fountain-18.svg](https://github.com/nebulon42/osmic/blob/master/outdoor/fountain-18.svg)

kocio-pl:
I would expect more symmetry (possibly with 4 or 6 water streams instead of just 2), since the fountains are mostly big, decorative objects. This one I would recognize as a small water "fountains" for drinking or washing hands.

mboeringa:
One warning here: often, people tag each and every fountain in a fountain group. This may mean dozens of "fountains" in a very small area. Prime examples are the fountains in the basin near Palais de Chaillot in Paris, and the famous "Fountains of Bellagio" in Las Vegas.

You won't be able to render this with an icon on any but the highest zoom... and even on Z19, it is likely icons may overlap, unless using routines to avoid collisions.

nebulon42:
I just wanted to notify this thread that there is a possible icon available. I didn't imply anything about feasibility or such and I also didn't intend to propose a PR.

matkoniecz:
>  it is likely icons may overlap, unless using routines to avoid collisions

In this style all labels and icons have enabled collision detection and no overlap is allowed.

mboeringa:
> In this style all labels and icons have enabled collision detection and no overlap is allowed.

Good. Still, cartographically, it may not lead to the nicest results if multiple icons are displayed close to each other, even if they do not overlap. At least something to consider and possibly try out with the example locations I pointed out, where this potential problem is likely to surface.

dieterdreist:
[...]

PR #1384, opened 2015-03-14
change rendering for natural=spring

### LINKED ISSUE #325: Change rendering for natural=spring

nebulon42:
Resolves #325.

before
spring_waterway_before
spring_forest_before

after
spring_waterway_after
spring_forest_after

Here I would again recommend icon halos that have been rejected in #1341.

### DIFF
--- a/amenity-points.mss
+++ b/amenity-points.mss
@@ -584,9 +584,10 @@
   }
 
   [feature = 'natural_spring'][zoom >= 14] {
-    marker-file: url('symbols/spring.svg');
+    marker-file: url('symbols/spring.16.svg');
     marker-placement: interior;
     marker-clip: false;
+    marker-fill: @water-color;
   }
 
   [feature = 'power_generator']['generator:source' = 'wind'],
--- a/symbols/spring.16.svg
+++ b/symbols/spring.16.svg
@@ -1,0 +1,40 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<!-- Created with Inkscape (http://www.inkscape.org/) -->
+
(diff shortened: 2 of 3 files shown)


imagico:
To me this looks like a group of small ponds and not like an icon.

Would it be feasible performance wise to distinguish (via SQL) between detached springs and springs at the start of a waterway?

kocio-pl:
I would take something like 6 symmetrical streams (2 small arches on top, 2 medium in the middle and 2 wide arches on the bottom) with common trunk. I think if we add then a "cup" on the bottom we would have also nice fountain:

https://github.com/gravitystorm/openstreetmap-carto/issues/705

Of course it doesn't help with visibility at second "after" image - halo looks like a hard requirement.

math1985:
Maybe just a round circle in water colour, slightly bigger than a stream, would do?

nebulon42:
I have tried that in #325. For spring to be recognisable within forests it should either not coloured like water or have an outline. I was thinking of a slightly darker blue, but it's still on my list.

nebulon42:
The symbol requires redesign too (too big), closing.

PR #1385, opened 2015-03-14
move lighthouse icon to SVG

nebulon42:
Implements part of #1165.
Please note that the symbol is now smaller.

before
lighthouse_before

after
lighthouse_after_2


imagico:
To me the light cones appear to be part of the structure and there seems to be a gap between the top and bottom part.  This makes the whole thing less recognizable IMO.

kocio-pl:
Another "too square" icon for me - and I would recognize it probably only in the right context (that should be less of a problem, since lighthouses are always near water I guess).

pnorman:
I think it's better with a taller aspect ratio.

nebulon42:
Making it appear taller is no problem but for the light cones I see only the possibility of leaving them out completely lighthouse-183 or replacing them with another (more ugly?) variant  lighthouse-184 or yet another variant lighthouse-185.

kocio-pl:
I strongly prefer 2, but I'd like to see all of them on the real map to be sure.

math1985:
I like the second option.

nebulon42:
That's how it looks on the map
lighthouse_after_2

Unfortunately it looks a bit ugly for larger sizes
lighthouse-184

math1985:
Both large and small look fine to me.

kocio-pl:
+1

mboeringa:
I am OK with the version in the last post (I also kind of liked the original one, although I can understand the objections of it being to abstract), however, maybe putting back the "lamp" in the icon, like in the old version, could improve recognizability even more for the last version.

nebulon42:
I have updated the rays of light as I have no better idea. @mboeringa I do not want to add the lamp to avoid that the upper part looks like an eye. I have experimented a bit with it and it is better to leave it out.

mboeringa:
> I have updated the rays of light as I have no better idea. @mboeringa **I do not want to add the lamp to avoid that the upper part looks like an eye**. I have experimented a bit with it and it is better to leave it out.

Fair enough, the final version looks fine to me...

PR #972, opened 2014-09-23
Render brand=* additionally to name=*

floscher:
This is supposed to solve parts of #698.
I've picked some amenities, for which it makes the most sense to render brand. These are:
* amenity=fuel (Shell, Esso, Agip, Total, BP, …)
* amenity=bank (Sparkasse, Volksbank, Commerzbank, …)
* amenity=fast_food (Mc Donald's, Burger King, Subway, …)
* shop=supermarket (Aldi, Lidl, Rewe, Coop, …)
* shop=department_store (Ikea, Woolworth, Galeria Kaufhof, …)

Depending on what tags are present (only name, only brand or both), one of the following variants is rendered:
* [name]
* [brand]
* [name] + "\n(" + [brand] + ")"

Currently this will result in many amenities (especially petrol stations) showing up as "Brand xy\n(Brand xy)", because many of them have the same value for the brand- and name-key. I even had difficulties to find a petrol station, that has different values for name and brand. But this is in my eyes a tagging-mistake (tagging for the renderer), which will hopefully be corrected as soon as brand renders in this style (see also [the comment by Rovastar to issue #698](https://github.com/gravitystorm/openstreetmap-carto/issues/698#issuecomment-49944999)).

PR #1388, opened 2015-03-15
add rendering for amenity=car_rental

### LINKED ISSUE #857: Add rendering for amenity=car_rental
amenity = bicycle_rental is rendered ok
but -
amenity = car_rental is completely invisible - no icon - no name - nothing at any zoom level.

In my opinion, this is too bad to be left uncorrected for long...

nebulon42:
Resolves #857.
I left aside the name vs. brand issue for now. #972 could possibly take this into account as well.

before
not rendered

after
car_rental


kocio-pl:
Good and easy to recognize. Maybe lights could be bigger a bit, but that's not crucial and we'd like to have all the cars-related symbols visually synchronized.

HolgerJeromin:
very nice. Thanks

PR #1389, opened 2015-03-15
move hotel icon to SVG

nebulon42:
Implements part of #1165.
The symbol is now smaller.

before
hotel_before

after
hotel_after


kocio-pl:
I would like the barrier lowered, the right bed leg on bed end and head being more visible - for me current is better visually and easier to recognize.

HolgerJeromin:
Could you try to add a pillow?

neiljp:
Making it smaller/consistent is admirable, but this is really a reduction in quality from my view. Zoomed in, I can see it is a bed shape. I know there is probably not space to show the pillow and head like in the old design, but maybe showing bigger pillow(s) would improve the visibility? Separating pillow from headboard? Support at foot of bed I'm unsure whether is better at end or closer as in new design.

math1985:
I would also prefer a bigger/higher head/pillow.

nebulon42:
Updated accordingly after changes to hostel icon. Pillow was not possible, but I have elevated the head and moved the bed stand to the right end.

math1985:
For comparison, the SJJB icon at 16px:

http://www.sjjb.co.uk/mapicons/png/accommodation_hotel2.p.16.png

nebulon42:
From my side this is now ready.

math1985:
I think I prefer the original, even the version downsized to 16px.

math1985:
Any other opinions?

matkoniecz:
I also prefer original icon. But for me svg is better than downsized version.

I also have idea for small change that makes this icon better for me.

e9ab949e-ce91-11e4-925e-d24912857663

kocio-pl:
Both new SVG variations are too small for me, especially head, which makes the icon harder to recognize (I see blue rectangle with some white spaces inside). @matkoniecz: could you enlarge the head in your version?

matkoniecz:
> @matkoniecz: could you enlarge the head in your version?

Unfortunately no, but enlarging anything in icons is beyond my abilities.

math1985:
@nebulon42 Any response to the comments above?

nebulon42:
On my list. :)

matkoniecz:
From looking at icon itself it seemed to not be an improvement - but from looking at how map changes it is in fact improvement. I support merging both #1389 and #1390. Note, I still think that edit proposed in https://github.com/gravitystorm/openstreetmap-carto/pull/1389#issuecomment-85626745 would be a good idea.

nebulon42:
Thanks for merging. The change by @matkoniecz might be contribute to a more clear look, but another problem is that the icon is already too lightweight. I will have to see how I can change this.

pnorman:
> Thanks for merging. The change by @matkoniecz might be contribute to a more clear look, but another problem is that the icon is already too lightweight. I will have to see how I can change this.

For future reference, if PRing a new change, it's probably best to do hotel and hostel in the same branch.

PR #1390, opened 2015-03-15
move hostel icon to SVG

nebulon42:
Implements #1165.
Symbol is now smaller.

before
hostel_before

after
hostel_after


kocio-pl:
Almost the same remarks as in https://github.com/gravitystorm/openstreetmap-carto/pull/1389#issuecomment-80988986, but I think in this case the barrier is not too too high.

dieterdreist:
> Am 15.03.2015 um 13:31 schrieb nebulon42 <notifications@github.com>:
> 
> Symbol is now smaller.


I find this hard to read because it is very dense, maybe enlarge the distance of the sleepers or make the bed thinner

nebulon42:
Yes, it is dense because it is hard to fit a bunk bed into 14x14px.
Maybe this would be an improvement: hostel-183

kocio-pl:
Maybe, but it's hard to compare - could you show how would it look like on the map?

neiljp:
Is the stand/support on the bottom bunk really necessary? Would removing it help with improving the shape/visibility? Maybe more space between bunks? bigger pillows/heads?

Similarly 'should' there be a vertical line on the right-hand side? Conversely, is removing the left-hand one valid? Does it leave the design 'intact'?

mboeringa:
Here is a small attempt at modification, not a lot that can be done in 14x14 pixels...

hostel_icon
:

kocio-pl:
For me it's acceptable.

nebulon42:
The person is a bit short, but not bad. @mboeringa If that is a SVG, could you please provide it as Gist?

mboeringa:
> The person is a bit short, but not bad. @mboeringa If that is a SVG, could you please provide it as Gist?

Sorry, this was just quick and dirty in MS Paint, to see if anything sensibly could be changed. If you download the PNG and zoom in to 800% in MS Paint, you can see what pixels I changed compared to the original rasterized version you supplied in your first post.

nebulon42:
Updated after input of @mboeringa.

mboeringa:
@nebulon42 ,

In my version, I lowered the "bedding" by 1 pixel from 3 to 2, as this looked a bit more natural in my eyes by creating more "space" between the lower figure and the top bed. Is there a technical hurdle to implement this in your SVG version?

Also, the "heads" appears higher, looking like they float. Here is a side-by-side comparison, left my pixel manipulated version, right @nebulon42's SVG version:
hostel_icon
hostel_icon_enlarged

nebulon42:
Yes, looks better.

nebulon42:
From my side this is now ready.

mboeringa:
Updated version in the original post looks fine to me too.

math1985:
Looks fine to me, but waiting for #1389 for consistency reasons.

matkoniecz:
From looking at icon itself it seemed to not be an improvement - but from looking at how map changes it is in fact improvement. I support merging both #1389 and #1390. Note, I still think that edit proposed in https://github.com/gravitystorm/openstreetmap-carto/pull/1389#issuecomment-85626745 would be a good idea.

PR #1397, opened 2015-03-17
Sql cleanup

pnorman:
more of the same.


project.yaml
```diff
@@ -1315,11 +1343,16 @@ Layer:
...
-              ) AS placenames_capital
+      table: |-
+        (SELECT
+            way,
+            place,
+            name,
+            ref
+          FROM planet_osm_point
+          WHERE place IN ('city', 'town')
+            AND capital IN ('yes', '4')
```
pnorman: This seems odd. Should this include 2 and 3?

math1985:
The Travis CI build failed.

project.yaml
math1985: I think we shouldn't include 2. Normally capital=2 is tagged as capital=yes. There are only 2 objects with capital=2 in the database.

We could include capital=3, but it's not urgent as there are only 8 instances in the database.

pnorman:
fixed

PR #1401, opened 2015-03-17
Render intermittend rivers with dashed line also on low zoom levels

### LINKED ISSUE #1305: intermittent rivers at low zooms

sommerluk:
Render intermittend rivers with dashed line also on low zoom levels.

Screenshot:

preview

Resolves #1305 and substitutes #1376 and #1399


math1985:
I agree this is the best solution.

Issue #1372, opened 2015-03-14
problem with peaks that are viewpoints


mkoniecz:
Peaks are frequently also great viewpoints. On printed map it is typically solved either by special icon for such situation or by shifting where viewpoint is marked (typically in direction of the most impressive view).

We are not doing the first one and second is tagging for renderer.

Before #1349 [natural=peak; tourism=viewpoint] was displayed as peak, now it is displayed as viewpoint - https://cloud.githubusercontent.com/assets/899988/6650865/ade346d2-ca27-11e4-8bd6-b58cefe5462b.png.

I propose to display it as special icon that would be merge of peak and viewpoint icons. @nebulon42 - can you maybe produce such icon?

nebulon42:
Yes, I can produce such an icon as viewpoint and peak is on my list anyway. But first I would like to wait for some consensus on whether a special icon is required/wanted or not.

brycenesbitt:
I think a peak icon with a ring of viewpoint icon would look great.
Better yet if the direction of view were properly taggable.

Tomasz-W:
Peak and viewpoint icons merged:

![natural peak tourism viewpoint](https://user-images.githubusercontent.com/25656654/36922089-0e1d0704-1e67-11e8-953d-bc05d1530566.png)

Tomasz-W:
@matkoniecz @nebulon42 @kocio-pl 
Are you satysfied by icon above, or you want something different?

kocio-pl:
First of all I'm not sure if that's good idea, but I would have to see the rendering to tell how it works visually. For example: which color should it be?

matkoniecz:
> "By definition", a peak is a viewpoint unless it's a smoothed hilltop covered by forest.

Nearly all peaks covered by forest are not viewpoints, many small peaks are not viewpoints, sometimes viewpoint is not at this highest point (peak), sometimes peak is not considered as viewpoint as view is blocked...

jeisenbe:
@Tomasz-W - is there an SVG icon available to try a test rendering? Or could you make a test by just copy-and-pasting the icon over a screenshot?

ppete2:
I'm not quite sure wheter it's a good idea or not to render them with a combined icon. Situation right now is, that peak icon+name+elevation is rendered.

[...]

PR #1421, opened 2015-03-19
improve coalesce rules - fixes #1371, improves #1372

### LINKED ISSUE #1371: tourism=attraction is selected too early in COALESCE
After #1349 [landuse=\*; tourism=attraction; name=\*] label is rendered in tourism attraction style. Old rendering in that case was better. 

landuse village_green tourism attraction name eeeeee eeeeee eeeeee ref 1 ele 8000 closed_way v2 28 1 - master
https://cloud.githubusercontent.com/assets/899988/6650786/b575c936-ca24-11e4-93fe-d3c470635102.png) - see z14 to z17
https://cloud.githubusercontent.com/assets/899988/6650789/b85cbcf4-ca24-11e4-80d2-9312f0e489f9.png - see z15+

It may be solved by splitting COALESCE for tourism key in two parts. Remove value attraction from the current one and add new at the end containing only value attraction. It would reduce priority of tourism=attraction.

>  'tourism_' || CASE WHEN tourism IN ('attraction')

tourism=attraction overrides also values of other keys COALESCE, except key aeroway.

matkoniecz:
tourism=attraction/viewpoint may be applied to anything and should not override label styling of other objects

fixes regression from #1349 

I am currently testing this PR, initial testing indicates that everything is OK.

amenity restaurant tourism attraction name eeeeee eeeeee eeeeee ref 1 ele 8000 node master - co


math1985:
Yes, looks fine.

Issue #710, opened 2014-07-10
Inconsistencies in amenity-points


math1985:
There are several inconsistencies in the amenity-points (https://github.com/gravitystorm/openstreetmap-carto/blob/master/amenity-points.mss) specification:

- [ ] The name of some POI is printed in bold or oblique, but it is not clear what the rationale behind that is
- [x] The name of POIs are printed in different text-sizes, again without a clear rationale.
- [x] Some POI have a larger halo-radius than others, without a clear reason.
- [x] Some landuse labels are rendered in black and others in the background color
- [x] Some POI have a smaller text-dy than others. This is likely to cause that some labels containing capitals or accents are never rendered. Note that text-dy depends both on the icon size and font size. It would probably be best to test this with a tall string such as 'ÉÉÉÉÉÉ'.
- [x] Some POI with icons, such as leisure = 'playground' (but potentially others), have no text-dy at all, meaning their text will never be rendered.
- [x] The length of text-wrap-width is inconsistent.
- [x] <s>The order of the attributes per selector is inconsistent (best resolve in separate commit)</s> =- now as #1172 .
- [x] Replace hardcoded colour values by variables

nebulon42:
I think the last point on the list was already done by #1378.

matkoniecz:
> Some POI have a smaller text-dy than others. This is likely to cause that some labels containing capitals or accents are never rendered. Note that text-dy depends both on the icon size and font size. It would probably be best to test this with a tall string such as 'ÉÉÉÉÉÉ'.

Marked as solved. I made and sometimes use https://github.com/matkoniecz/CartoCSSHelper to catch this problem.

matkoniecz:
halos will be nearly finished by #1810 - but there are also shop labels

matkoniecz:
> I tested and recommend switching to bold for all POI labels, as it significantly improves readability at cost of minor label size increase.

And now almost two years later I tested it again, and to my surprise book seems to be preferable. It is much prettier at very minor readability cost.

But I have no strong opinion on this topic.

https://cloud.githubusercontent.com/assets/899988/16176415/215deece-360f-11e6-9f0a-d6274986a06d.png
https://cloud.githubusercontent.com/assets/899988/16176416/2160c0b8-360f-11e6-81c2-ea4034bc25d7.png

https://cloud.githubusercontent.com/assets/899988/16176417/2161001e-360f-11e6-856b-0bdf30279c57.png
[...]

Issue #459, opened 2014-04-07
Add label for tourism=viewpoint


1ec5:
[This observation deck](http://www.openstreetmap.org/way/91292276) is tagged `tourism=viewpoint`. The stylesheet applies an ![asterisk](https://raw.githubusercontent.com/gravitystorm/openstreetmap-carto/master/symbols/view_point.p.16.png) icon but ignores the `name` tag. When the viewpoint sits atop a landmark building, it’s important to render a label because the viewpoint will obscure any label for the building as a whole.

For now, the workaround is to tag the observation deck as a less specific `tourism=attraction`, as with the [Sears Tower](http://www.openstreetmap.org/node/2311635030).

matkoniecz:
It is still not rendered on closed ways.

PR #1429, opened 2015-03-20
unify some label definitions, closes #1231

### LINKED ISSUE #1231: Is it a good idea to reuse code during defining various features?
Frequently style definitions are repeated for various features, maybe with minor change like text-dy.

Is it a good idea to attempt reuse of this code? See my proposed refactoring at https://github.com/mkoniecz/openstreetmap-carto/commit/1d68047b7cf516fb138a5e35685b9b95eff28c7f

Benefits - lower amount of code, it is obvious that style is repeated.
Problems - may be harder to understand.

I am not 100% sure whatever increased clarity by removal of repetition is greater than what was lost by using slightly more complicated construction (for somebody new - I think that I may now underestimate difficulty for such people).

Maybe it should be limited to cases where style is exactly the same, except zoom level range?

matkoniecz:
it will allow easier 
- adding labels
- customization
- keeping style of similar labels synchronized

It is also preparation for #459 and part of #710

Rendering is not affected (except elements of exactly the same area).

### DIFF
--- a/amenity-points.mss
+++ b/amenity-points.mss
@@ -692,6 +692,7 @@
   }
 
   [feature = 'amenity_library'],
+  [feature = 'tourism_museum'],
   [feature = 'amenity_theatre'],
   [feature = 'amenity_courthouse'],
   [feature = 'amenity_townhall'],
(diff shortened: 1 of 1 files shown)


matkoniecz:
thanks!

PR #1433, opened 2015-03-21
Don't exclude supermarket and station from building name rendering

### LINKED ISSUE #1396: building-text exclusion oddities
building-text has the condition `building NOT IN ('no', 'station', 'supermarket')`. This does not seem matched by anything on another layer which renders building=supermarket text. Probably left-over from when we rendered supermarkets differently (#520) and we missed this layer in #1153

math1985:
This resolves #1396.


matkoniecz:
It is rather "Don't exclude" (also in commit description).

math1985:
You're right, fixed.

matkoniecz:
Thanks!

PR #1443, opened 2015-03-21
Fix SQL for rail service/yard

### LINKED ISSUE #1442: Railway yard tracks suddenly stopped rendering
See https://www.openstreetmap.org/way/219993787#map=19. There are supposed to be about 15 rail yard tracks there, but they stopped getting rendered for some reason. (Tags: <code>railway=rail</code>, <code>service=yard</code>.) However, tracks with <code>railway=subway</code>, <code>service=yard</code> are <a href=https://www.openstreetmap.org/way/207717060>still rendered</a>.

pnorman:
80a990a9 introduced a regression where railway service/siding/yard
wasn't being rendered due to an internal transformation which wasn't
also being done elsewhere in the SQL.

Fixes #1442


pnorman:
cc @gravitystorm probably worth tagging v2.29.1 to fix this

math1985:
Thanks for quick fix!

Issue #1028, opened 2014-10-07
Stop landuse labels based on area


pnorman:
#941 gave us labels which start rendering based on area.We should be doing the counterpart to this and stop rendering when an area gets too big.

As one of the more extreme examples, http://www.openstreetmap.org/#map=19/39.18521/-96.81765
![image](https://cloud.githubusercontent.com/assets/1190866/4550284/2ddaa842-4e65-11e4-9e40-585e21ce4877.png)
[Continued in all directions]

From that zoom, I have to zoom out 5 times (128x) to see a boundary, meaning placing text in the middle is rather meaningless. I have a 1080p screen, so most people will be 5 or 6 times.

A few options
- Stop rendering labels based on way_pixels
- Scale the label to _very_ large, keeping it a reasonable percentage of the overall size. Requires expr everywhere and Mapnik 3
- Repeat the label. Not technically trivial, and probably would cause issues in some other cases.

math1985:
A solution for the Carto bugs that prevented doing this (https://github.com/mapbox/carto/issues/369 and https://github.com/mapbox/carto/issues/370) have been merged today.

Writing a

math1985:
A solution for the Carto bugs that prevented doing this (https://github.com/mapbox/carto/issues/369 and https://github.com/mapbox/carto/issues/370) have been merged today.

Solving the issue here would imply that we require the tileservers to run the latest Carto version. @gravitystorm @pnorman Any opinions on this?

gravitystorm:
I want to ensure that this style can be run on a variety of systems, such as tilemill and kosmtik. So I certainly don't want to depend on anything that is in an unreleased version of carto, and I'd prefer to wait until kosmtik has a release containing that version of carto too (ideally Tilemill but that seems unlikely).

Of course, the OSMF tileservers are a separate task, but after a carto release that should be straightforward.

math1985:
>  I'd prefer to wait until kosmtik has a release containing that version of carto too (ideally Tilemill but that seems unlikely).

Would that involve more than bumping the carto version in package.json in both respective projects?

gravitystorm:
I don't know exactly what's involved, tbh.

math1985:
Confirmed that changing the node version in kosmtik can simply be accomplished by changing the version number in package.json and running npm install from the kosmtik directory.

In Tilemill, it should work likewise. Could somebody running Tilemill confirm that?

pnorman:
We need

[...]

PR #1444, opened 2015-03-21
Do not render island/islet name when way_pixels is big

sommerluk:
Do not render island/islet name when way_pixels is big

If islands are big and the zoom level is high, we still have a label rendered for these islands. This seems a little confusing to me.

This PR uses way_pixels to discard label rendering if the island surface is quite big (relative to your screen).

Maybe this approach may be also useful for other way_pixel based labels?

### DIFF
--- a/amenity-points.mss
+++ b/amenity-points.mss
@@ -687,10 +687,10 @@
 
 .text-low-zoom[zoom < 10],
 .text[zoom >= 10] {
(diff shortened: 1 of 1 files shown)


math1985:
Thanks!

matkoniecz:
> Maybe this approach may be also useful for other way_pixel based labels?

See #1028

PR #1445, opened 2015-03-21
Repair island labels on low zoom levels

sommerluk:
At amenity-points.mss we have code that tries to render island labels starting from z=7.

It seems that during a SQL refactoring there were some changes. So currently it renders only from z=10.

This PR fixes the SQL, so that island labels are rendered from z=7 again.


imagico:
This may be beyond the scope of this but as i mentioned in my comment on #990 if place=archipelago is rendered with priority over place=island it would make sense to start earlier than z=7 (the french style starts with z=4).  For performance reasons you would want to move it out of text-poly-low-zoom then though.

Specific sugggestion - move place in front of landuse/natural in COALESCE().

PR #1446, opened 2015-03-21
Add icon for hunting stand

### LINKED ISSUE #966: Add rendering for amenity=hunting_stand
Please use http://goo.gl/I6e5XW for amenity=hunting_stand.

I am the author of the icon and I am releasing it under public domain.

If you need to do any modification to it, please feel free or contact me.

math1985:
Thanks to @nebulon42 for the icon.
This resolves #966.

screenshot from 2015-03-21 10 50 59


matkoniecz:
Is it a good idea to display hunting stands mapped as areas?

dieterdreist:
2015-03-24 6:01 GMT+01:00 Mateusz Konieczny <notifications@github.com>:

> Is it a good idea to display hunting stands mapped as areas?



why not? IMHO everything that is more than a spot (e.g. a pole, an abstract
entity like a "center" or other small objects like a waste bin) should be
represented as an area. Even for benches it makes sense to use more than a
node (because of the orientation).

kocio-pl:
Since we have so good aerial images, we can go down to about decimeters - many things in micro-scale have at least 2 dimensions.

matkoniecz:
@kocio-pl @dieterdreist Maybe editing http://wiki.openstreetmap.org/wiki/Tag:amenity%3Dhunting_stand would be a good idea? Currently it recommends using nodes and only nodes.

dieterdreist:
2015-03-24 12:32 GMT+01:00 Mateusz Konieczny <notifications@github.com>:

> Maybe editing
> http://wiki.openstreetmap.org/wiki/Tag:amenity%3Dhunting_stand would be a
> good idea?
>

done

SomeoneElseOSM:
FWIW I had a look around the local area for things that either are already mapped as hunting stands or could conceivably be mapped that way.  Quite a lot are other things mistagged (that's neither the fault nor the concern of the stylesheet, obviously), but there are certainly things that you could perhaps map as an area - the  building at http://www.openstreetmap.org/node/706799650 for example (I went with bird_hide when it was previously accessible, which it isn't currently, but originally it was for shooting ducks from).

So yes, hunting_stand makes sense on areas, even though there is alas no "decimeter accuracy aerial imagery" at that location - Bing's offset by about 1-2m there.

kocio-pl:
Of course the accuracy is not always that good, but area rendering means only it _can_ be rendered when we know more than "there is a POI". We should "enable" (on Wiki) tagging as areas for many other objects, BTW.

PR #1447, opened 2015-03-21
Add nature reserve text to lowzoom layer

### LINKED ISSUE #1164: Add leisure=nature_reserve to text-low-zoom layer
It seems that the tags `leisure=nature_reserve` and `boundary=national_park` behave differently when it comes to showing a label in low zoom leves, and I believe they shouldn't.  For instace, in the followin tile:

http://www.openstreetmap.org/#map=8/0.698/-65.896

the Venezolan Parque Nacional Serranía de la Neblina (http://www.openstreetmap.org/relation/3404666, tagged with boundary=national_park) has a label, while its Brazilian counterpart (http://www.openstreetmap.org/relation/3424547, tagged with boundary=protected_area) doesn't have one.

math1985:
This resolves #1164


matkoniecz:
```OR "place" IN ('island')``` changed to ```OR "place" IN ('island')``` - in my opinion it is better to separately commit refactoring and new features, even in case of trivial ones.

math1985:
Updated.

PR #1449, opened 2015-03-21
Change rendering man_made=cutline

### LINKED ISSUE #781: Change rendering man_made=cutline
The following issue has been moved over from [trac](https://trac.openstreetmap.org/ticket/4253):

The rendering of man_made=cutline doesn't fit into the rest of the map style. The line is far to wide within serveral zoom levels and to "strong" (continious line) so it can be easily misread as highway=residential or highway=unclassified and the "way importance" can therefor be overrated compared to highway=track and highway=path.

Current rendering can bee seen here:

http://www.openstreetmap.org/?lat=50.01387&lon=7.73493&zoom=16&layers=M

wiki description:

 http://wiki.openstreetmap.org/wiki/Cutline

Suggestion of rendering style (Better then the wiki suggestion which can be missunderstood as highway=path !!!):

    Rendering width identical to highway=track or smaller like highway=path
    dashed line 50% white, 50% transparent, 50 white, and so on...

math1985:
This resolves #781

Before:
screenshot from 2015-03-21 19 25 12

After:
screenshot from 2015-03-21 19 22 33

### DIFF
--- a/landcover.mss
+++ b/landcover.mss
@@ -536,12 +536,13 @@
 /* man_made=cutline */
 #landcover-line {
   [zoom >= 14] {
-    line-width: 3;
+    line-width: 2;
     line-join: round;
     line-cap: square;
-    line-color: @land-color;
+    line-color: white;
+    line-opacity: 0.3;
     [zoom >= 16] {
-      line-width: 6;
+      line-width: 4;
     }
   }
 }


kocio-pl:
Makes a lot of sense.

pnorman:
Probably should follow on to randomized symbology

imagico:
I don't think using transparency is a good idea here - while the wiki suggests cutline to be only used on forest/wood firebreak cutlines make also sense on scrub/grassland and transparency will just lead to a huge number of additional green tones to be introduced.

Apart from that it looks fine.

math1985:
>  firebreak cutlines make also sense on scrub/grassland

Wouldn't that be rather an argument in favour of transparency, rather than defining a fixed color?

imagico:
> Wouldn't that be rather an argument in favour of transparency, rather than defining a fixed color?

If there were only very few green tones in the style it would but with the overwhelming number of greens it will just be a mess IMO.

With  #1242 you could probably use `mix(@land-color, @forest, 50%)` without transparency and still get something distinguishable on scrub and grassland.

math1985:
@matkoniecz What do you think?

I still think transparency works better than a fixed colour if it's supposed to work on different backgrounds.

matkoniecz:
@math1985 I am not sure, I would need to test it to give an opinion.

imagico:
My objection should probably not weight too heavy in the light of practical use of the tag - more than half of it is from imports, often of questionable reliability and a lot of uses are without any background landcover.  Here a greenish solid color would of course be counterproductive.  It might be worth considering not rendering the tag at all.

matkoniecz:
Rendering after this PR is better, also after considering glitches with white transparency rendered twice (in crossing points of two cutlines - in X or Y shaped junctions).

Have you tried version without transparency - with narrow line in land colour? Maybe that would be OK and it would avoid transparency.

> It might be worth considering not rendering the tag at all.

Maybe, I have no opinion (I area where I map nobody introduced it - neither by mapping nor by imports).

matkoniecz:
In addition - there are tow ways of mapping cutlines through some vegetation area:

a) replace area of vegetation by multipolygon/two areas, with cutline excluded ( like at http://www.openstreetmap.org/?mlat=50.44101&mlon=18.15309#map=18/50.44101/18.15309 )
b) draw man_made=cutline line

It would preferable to end with nearly the same rendering, so land-coloured line like current rendering is in my opinion. It is only far too wide.

dieterdreist:
> Am 30.04.2015 um 07:21 schrieb Mateusz Konieczny <notifications@github.com>:
> 
> a) replace area of vegetation by multipolygon/two areas, with cutline excluded ( like at http://www.openstreetmap.org/?mlat=50.44101&mlon=18.15309#map=18/50.44101/18.15309 )
> b) draw man_made=cutline line


you could do only a), only b) or both. All should look the same (ideally).

math1985:
> Rendering after this PR is better, also after considering glitches with white transparency rendered twice (in crossing points of two cutlines - in X or Y shaped junctions).

Yes, you're right, that's indeed a problem.

matkoniecz:
Superseded by #1526

Issue #1295, opened 2015-02-07
background color for amenity=social_facility


jengelh:
Current configuration on osm.org does not highlight the area: http://osm.org/relation/4568610
Proposing to handle it just like amenity=university.

dieterdreist:
2015-02-07 21:59 GMT+01:00 jengelh <notifications@github.com>:

> Current configuration on osm.org does not highlight the area:
> http://osm.org/relation/4568610
> Proposing to handle it just like amenity=university.
>


-1, these are clearly very different. While universities are almost always
very important (even if you don't want to go there, they are often known by
lots of people and typically big, and can serve for orientation), while
social facilities come in every size and importance level imaginable.

kocio-pl:
Universities can be also private university-level schools and we have it a lot in Poland - some are big (multi-building with the common area), some other are just using rented office space. I also know about social facilities that are located in such a rented space and also big like a school - especially orphanage.

And the ruke is simple - if it is small, no amenity background (just POI or building), if big - the background should be used. So no need to virtually differentiate both types by definition - the size will speak for itself.

dieterdreist:
2015-02-18 14:16 GMT+01:00 kocio-pl <notifications@github.com>:

> Universities can be also private university-level schools and we have it a
> lot in Poland - some are big (multi-building with the common area), some
> other are just using rented office space. I also know about social
> facilities that are located in such a rented space and also big like a
> school - especially orphanage.




in my understanding, these would not be "universities" but "colleges".

kocio-pl:
So I have to get deeper into anglo-saxon conventions of educational system. Once again there's something not clear and universal enough to me.

However it's irrelevant in this case. I sustain my claim, that if something like this is big enough, it should be shown and we don't have to judge it beforehand, because if we have the area with amenity=social_facility tagging, we can be sure it's about the right size. For example [orphanage may look just like a regular school with a pitch](http://www.openstreetmap.org/way/304204007) or [the social facilities can be grouped together in one fenced location](http://www.openstreetmap.org/way/39230878). If they are just inside the building or have no area to tag outside the building, we would treat automatically them the same as schools with no additional area.

Also amenity=childcare can also look like kindergarten with a playground, so i would add it too.

matkoniecz:
I think that it is impossible to continue increasing number of different landcovers.

Using the same colour as educational/hospital areas seems to be a poor idea - this way it will not mean anything.

I close this as it is not really possible to render more landcovers.

[...]

PR #1450, opened 2015-03-21
Render police grounds like hospital grounds

### LINKED ISSUE #848: Add rendering for amenity=police area
Areas mapped as amenity=police should be rendered.

[Example](http://www.openstreetmap.org/#map=18/51.91870/-0.65427)

See also https://trac.openstreetmap.org/ticket/3622.

math1985:
This resolves #848.


matkoniecz:
I am not entirely sure whatever rendering all amenities with this yellow colour is a good idea. Maybe restricting it to education areas would be better? But that would require either lack of area rendering for some features like police or hordes of new landcover colours/patterns.

I am not happy about any of this solutions.

kocio-pl:
I thought we can use the same color for social facilities, but the problem is much harder. This is something in the roots of OSM tagging scheme: while we have very useful building scheme, we don't have anything even remotely clean and generic for areas. 

1. We have building=yes well defined and rendered, but we don't use area=yes as a general area type (just an option for typical way to behave like the area).

2. Areas are scattered between different entities: landuse, some amenities (like in school), natural (but not tree probably), highway=*+area=yes and maybe others too. It means we simply don't have the idea of an area as an object - just buildings (very special areas), POIs ("points", which are probably always areas in micro-scale) and some special areas we recognize.

3. The building could be represented as a node if we don't know the geometry and the same we do for some areas - correctly (place=*) or in a slightly twisted way: some areas are symbolized by the central building - be it because we don't know the geometry (justified use) or because the area is not represented on the map (many of them have no color, icon, nor even the label), which is bad.

I think we need general cleaning here on the project primitives level. Of course we may use some rendering strategies in the meantime for using areas colors, icons and labels, because this process may last for years and have no real solution. But we have to remember it will be very limited and case-based until we know how we should properly deal with areas.

pnorman:
I'm still on the road so I haven't had a chance to review it, but I am somewhat skeptical about adding too many different features to the same color rendering.

I don't see that concerns about the OSM data model are relevant here. By the time we have the data it has been processed by osm2pgsql and there are no issues telling polygons and linestrings apart.

kocio-pl:
It's not a technical problem per se, but it means we have no system, so we have to think carefully case-by-case (and the code won't be such clean as for buildings or POIs). 

On the Tagging list there's a discussion about landuse=civic, so at least one more type we may consider worth showing. But what is more important is that is hard to know which criteria we should use when thinking about rendering areas. Buildings are easy (normal rendering like building=yes, special rendering like building=*+amenity=place_of_worship, no rendering for building=no), POIs and highways are currently quite easy too (different sets of icons+colors for categories+labels), but areas are a mess - no area categories (colors are tied to specific objects) and just a few tagging schemes are recognized as areas.

The problem is we need some categories for coloring/labeling areas to avoid "rainbow" map (if we keep just adding new types or colors), but it is hard to know what categories we should use, because we don't have a clear understanding what is an area at the most fundamental level. If we can do the area categorization, rendering them would be much easier and more consistent.

dieterdreist:
> Am 27.03.2015 um 13:26 schrieb kocio-pl <notifications@github.com>:
> 
> The problem is we need some categories for coloring/labeling areas to avoid "rainbow" map


I believe the crucial part is labeling, no need to color every area that has a tag which we care for, a label is the important thing when you look at the map

cheers 
Martin

kocio-pl:
I would agree with you - it's easier to decide, less risky for the final look, and probably more important than coloring (it's good to know the area type and borders, but the name brings more informations). That would be close analogy to buildings: for generic buildings name tag we use standard text styling, but if we know what it is, we can use different style and even the icon.

So which cases of areas we want - or definitely don't want - to show its labels, if there is a name of course? General namespaces, that can be areas, which we have now (quick and dirty list):
1. Landuse
2. Landcover
3. Amenity
4. Natural
5. Leisure
6. Admin_level
7. Other tagging schemes with area=yes

However labels (and icons) have their own problems the coloring does not have: what label/icon should be visible when it clashes with buildings/POIs inside? For me last resort strategy would be to render areas over the objects inside. But I think first try would be to shift that name/icon somewhere within area borders if there are some important objects (more advanced version of labels placing: SNEW) - is this technically possible now?

math1985:
I still think that it would be useful to render the outline of police grounds. I would prefer using the same colour for multiple objects over using a map that looks like a rainbow.

We currently use light yellow only for relatively small objects (hospitals and schools), so from the context (name and icon) is should be clear what object it is.

matkoniecz:
After thinking more - I think that I prefer removal light yellow fill from hospitals and restricting it to education areas rather than expanding it to even more diverse meanings. I think that the end result may be that it is used for everything and therefore losing any meaning.

polarbearing:
All these campus structures need some coherence. While a police campus often has a mappable fence that currently provides its boundaries, modern hospitals and universities are often fenceless. University hospitals have both educational and medical function - the same colour reflects that currently quite well. 

Thus I'd be happy to see/keep them filled, being it the same colour for all, or two different, with education+medical tending into one, and police+civic_* (once settled) tending into the other, fire-fighters to be decided.

dieterdreist:
2015-05-07 17:05 GMT+02:00 Mateusz Konieczny <notifications@github.com>:

> After thinking more - I think that I prefer removal light yellow fill from
> hospitals and restricting it to education areas rather than expanding it to
> even more diverse meanings.



+1

polarbearing:
So should we have one particular area colour for all non-educational services to the community, including hospitals, police, and some that lack area rendering currently such as amenity=social_facility #1295 (32000), community_centre (27000), nursing_home (11000) which often have a campus.

kocio-pl:
I'm curious why do you split social amenities/services like this (educational vs the rest)?

But if we agree this is the right kind of split, I see the only problem that this would be a change for hospitals (now also clinics are like this and I think doctors_office campuses should be the same too). However this makes sense for me and would be huge win for many other social facilities, so there are more pros than cons. And if we include police stations, I think also about fire stations here.

Do you think about any particular filling color (or maybe border color)?

polarbearing:
> I'm curious why do you split social amenities/services like this (educational vs the rest)?

To reach a consensus with folks that want to separate them.

> And if we include police stations, I think also about fire stations here.

Oh yes, the list above were just a few examples. Added the fire people. 

> Do you think about any particular filling color (or maybe border color)

For a fill, I was thinking about a yellow with a bit more grey in it than the current educational.

Border is a good alternative, it would fit in the consideration in #1624 to treat all humanly defined sites similar. Consequently, that might then apply to the educational sector as well, and we would need a colour scheme for all these boundaries. 

Such as tourism-brown, educational-yellow/light, social+civic services-yellow/dark, transportation-blue including the accommodation, nature-green.

kocio-pl:
I still don't get why @matkoniecz and @dieterdreist think educational areas should be so special (what about hospital campuses - are they less special or what)?

I think boundary color may be indeed better than fill color, because some well-micromapped places are "vanishing" because of being eclipsed (especially by grass and parkings, also by buildings - like for example [this hospital](http://www.openstreetmap.org/way/66266360)). Current zoo rendering shows it really works. I guess educational areas should also be rendered in this manner (without a filling color), so we the problem of being "special" or not would be non-existent then. Colored borders are also less obtrusive than colored filling, so the map will not be so "rainbow" even when we use a couple of colors as you suggested in the last paragraph.

Could you provide some code draft to test the idea (maybe also with some interesting places rendering)?

dieterdreist:
2015-06-30 18:09 GMT+02:00 kocio-pl <notifications@github.com>:

> I still don't get why @matkoniecz <https://github.com/matkoniecz> and
> @dieterdreist <https://github.com/dieterdreist> think educational areas
> should be so special (what about hospital campuses - are they less special
> or what)?
>
>
I am not particularily keen on highlighting schools and universities (it
just happens to have been decided by the old team and is like this for
years), but I think filling hospitals with the same colour is somehow
confusing.

matkoniecz:
I am not thinking that educational areas deserve to have fill more than hospitals and police grounds, just that using the same style for everything would make it meaningless.

polarbearing:
So we could have the boundary colour define the category (tourism-brown, services-yellow, etc), and the icon define the specific type (zoo, museum; school, hospital, police), many of the icons exist.

math1985:
Closing this as there seems to be no support for this change.

PR #1455, opened 2015-03-22
remove rendering on areas for natural=peak, volcano, saddle, cave_entrance

matkoniecz:
It makes code less confusing.


dieterdreist:
in the case of a vulcano it would seem to make more sense mapping it as an area

math1985:
> in the case of a vulcano it would seem to make more sense mapping it as an area

According to the wiki, it may only be used on notes.

math1985:
This removes spring icons but not labels.

matkoniecz:
Spring is no longer affected.

dieterdreist:
2015-03-22 19:14 GMT+01:00 math1985 <notifications@github.com>:

> in the case of a vulcano it would seem to make more sense mapping it as an
> area
>
> According to the wiki, it may only be used on notes.
>

we could change this, according to taginfo, a small number (around 5%) is
already tagged on ways depite the misleading wiki page.

dieterdreist:
btw.: just discovered: this was recommended to be mapped on areas or nodes,
and just one year ago changed (unilaterally I guess) to nodes only:
http://wiki.openstreetmap.org/w/index.php?title=Tag%3Anatural%3Dvolcano&diff=996213&oldid=995784

PR #337, opened 2014-02-11
order placement of city/towns names

cquest:
use population tag to place city/town names of major cities/towns first.

Fixes #1083

PR #1461, opened 2015-03-23
Order placenames by population and substitute NE usage on low zoom level...

### LINKED ISSUE #50: Remove ogr2ogr dependency
Given that we only need it for one minor task, we should remove the ogr2ogr dependency from the setup. It's not too bad on linux systems (yay package management) but it's an unnecessary hurdle elsewhere.

If we can't get the NE shapefile fixed upstream, then lets host a fixed version of the file somewhere and grab it from there instead.

### LINKED ISSUE #255: Replace Natural Earth labels with OSM-derived version
Long term project, but we *should* be able to use OSM data for labels. This will probably have to be something pre-processed into a shapefile.

### LINKED ISSUE #482: Add information to INSTALL.md on ogr2ogr
The fix the README tells us [to be required](https://github.com/gravitystorm/openstreetmap-carto#populated-places-shapefile) is not necessary anymore I think. The referenced bug https://github.com/mapnik/mapnik/issues/1605 has been closed a year ago. Could you please update the README and fix the naming of the files in the TileMill project?

Vale,
Quintus

### LINKED ISSUE #1083: Bigger town's names occluded by clearly smaller town's names
When I see Vilanova i la Geltrú (south of Barcelona) 
http://www.openstreetmap.org/#map=12/41.22392/1.72511
, a town with 65000 population, and all these tags 

addr:country_code	ES
addr:postcode	08800
admin_level	8
capital	8
ele	25
is_in:country	Spain
is_in:country_code	ES
is_in:municipality	Vilanova i la Geltrú
is_in:province_code	08
name	Vilanova i la Geltrú
name:ca	Vilanova i la Geltrú
name:es	Villanueva y Geltrú
place	town
population	66532
population:date	2009
ref:ine	08307000100
source	Instituto Geográfico Nacional
source:date	2011-06
source:ele	MDT5
source:file	http://centrodedescargas.cnig.es/CentroDescargas/equipamiento/BD_Municipios-Entidades.zip
source:name	Nomenclátor Geográfico de Municipios y Entidades de Población
wikipedia	ca:Vilanova i la Geltrú

It gets hidden by other smaller towns, with less priority like Sant Pere de Ribes

name	Sant Pere de Ribes
name:ca	Sant Pere de Ribes
name:es	San Pedro de Ribas
place	town
population	11904

and same priority but less population as Vilafranca del Penedès

admin_level	8
capital	8
designation	town
ele	229
is_in:country	Spain
is_in:country_code	ES
is_in:municipality	Vilafranca del Penedès
is_in:province_code	08
name	Vilafranca del Penedès
place	town
population	35864
ref:ine	08305000500
source	Institut Cartogràfic de Catalunya, IDESCAT
source:date	2011-06
source:ele	MDT5
source:file	http://www.idescat.cat/cat/idescat/publicacions/cataleg/pdfdocs/nomencat2010.pdf
source:name	Nomenclátor Geográfico de Municipios y Entidades de Población
website	http://www.vilafranca.cat/
wikipedia	ca:Vilafranca del Penedès

When I ask that to the catalan list they say me it happens in other towns like Mataró pop. 116698 vs. Argentona pop. 11544.
All these happens at lower levels than 11.These towns (Vilanova i la Geltrú and Mataró) should start to appear from level 9 http://www.openstreetmap.org/#map=9/41.7734/1.7413 and the smaller towns would start to appear from level 11 http://www.openstreetmap.org/#map=11/41.3552/1.9398 .

But also I notify other thing.As OSM standards tags a city starting from 100000 population http://wiki.openstreetmap.org/wiki/Tag:place%3Dcity (spanish OSM standards also http://wiki.openstreetmap.org/wiki/ES:Key:place )  , in Spain there are so many cities that also be with same tag (town) and same name's size but not the same population (from 10000 to 100000 there are more than 400,some near to others so it is important to make a difference).Is there anyway render would count population to make the name bigger than other towns with less population?

### LINKED ISSUE #1505: More fine grained capital ordering
Perhaps there should be one more "level" of capital rendering. Currently capital=yes and =4 are treated the same way (#1247 and f02d12c3a66bb1e6c011a4d0faaea01b1f29d86a) which leads to priority conflicts between national capitals and regional capitals. Maybe capital=yes and capital=2 should be placed in placenames-capital and a new category, placenames-medium++ ;), should be created where ≈4 and would reside.

Example of Malmö clobbering Copenhagen: http://www.openstreetmap.org/#map=6/55.260/12.722

sommerluk:
Order placenames by population and substitute NE usage on low zoom levels.

Resolves  #50 
Resolves #255 
Resolves #482
Resolves #1083 
Resolves #1505 

It tries to order cities and towns by population. This would improve the rendering regarding which label has to be supressed in case of collision.

It substitutes Natural Earth populated places labels on low zoom levels (3–4). It uses OSM data instead, making a choise based on population=*. It uses the same font color at 3–4 like on the higher zoom levels.

It tries to guarantie that once a label has been rendered at a specific zoom level, it is not supressed at higher zoom levels by other labels of start rendering only at higher zoom levels (as far as possible).

It tries to preserve more or less the current placenames density.


matkoniecz:
Can you provide before/after comparison?

imagico:
Regarding ordering by population - this might be a good idea but would need some examples.

Regarding replacing the NE data with a high population cutoff - bad idea.  I think this has been discussed before, importance rating of places does not work well based on population alone, you would need to take into account the surrounding, a huge city next to even bigger cities should not be rendered (think of the US east coast for example) while a medium size city with nothing comparable around for hundreds of kilometers should (think of Anchorage, Alaska or Murmansk, Russia).

sommerluk:
Here comes a comparision for z=3.

screen3

(But I don’t have a complete planet in the database. Only place=city here, with the country names missing.)

In general, for z3 and z4, where “Natural Earth” is substituted, this means less city labels in Europe and North America and more city labels in Asia and Africa. The threshold can be tweaked…

On the higher zoom levels it changes mostly which label will be supressed in case of collision.

project.yaml
```diff
@@ -1382,10 +1392,21 @@ Layer:
             way,
             place,
             name,
-            ref
+            ref,
+            CASE
+              WHEN (population ~ '^[0-9]+$') THEN CAST(population AS integer)
+              WHEN (place = 'city') THEN 100000
+              ELSE 10000
+            END AS population,
```
math1985: Maybe use INT_population here or something like that, otherwise it's very easy to get confused.

project.yaml
```diff
@@ -1382,10 +1392,21 @@ Layer:
             way,
             place,
             name,
-            ref
+            ref,
+            CASE
+              WHEN (population ~ '^[0-9]+$') THEN CAST(population AS integer)
```
math1985: Is there a difference between `CAST(population AS integer)` and `population::integer`?

pnorman: they are the same

kocio-pl:
On low zoom levels I'd like to see countries/states and capitals, no matter how much people live there. Population is one of the most important hints for rendering when we have conflicting name labels (see https://github.com/gravitystorm/openstreetmap-carto/pull/337).

sommerluk:
@imagico 

> a huge city next to even bigger cities should not be rendered (think of the US east coast for example) while a medium size city with nothing comparable around for hundreds of kilometers should

Maybe we can get closer to the desired behavior by raising text-min-distance and at the same time render places with lower populations earlier. So, we get more collisions, and these collisions can be solved e.g. based on the population=* tag.

However, I think that Natural Earth does a quite bad job regarding your criteria. It displays a lot of labels in very small continent which is called “Europe”, and it displays less labels in a much bigger continent which is called “Africa”. So, I would say we don’t loose anything.

@imagico 

In #337 you have proposed to use a score value:

>  For example you could use the population as score and multiply it with a factor of for example three in case it is a national capital. This would mean a larger city is only rendered with preference if its population is more than three times that of the capital.

Currently, we treat not only national capitals (“capital=yes”), but also regional capitals of a givel level (“capital=4”) in a special way. So, maybe we could do this:

national capital → population × 3
regional capital → population × 2
all others → population × 1

What do you think about?

Currently, we have also various layers — in this order:

↗ Natural Earth for z3-z4
↗ a layer for the capitals
↗ a layer for the rest of cities and town (with different font sizes for cities and towns at most zoom levels)

If we would adopt the solution with the score: Could we even use only one layer (with different font sizes based on the score)?

pnorman:
> However, I think that Natural Earth does a quite bad job regarding your criteria. It displays a lot of labels in very small continent which is called “Europe”, and it displays less labels in a much bigger continent which is called “Africa”. So, I would say we don’t loose anything.

That is a result of the cutoffs we have set for rendering. It would be fairly trivial to display more cities, and in a sensible rendering order.

imagico:
> However, I think that Natural Earth does a quite bad job regarding your criteria.

Indeed - this is one of the reasons it should be replaced.  My favorite example is:

Miami: population 415000, shown
Havana: population 2.1 million, capital of Cuba, not shown

But i would be against using this as an excuse to take the easy way and choose an alternative approach nearly as bad while better methods exist.

If you can use Mapnik's collision detection to do the necessary proximity processing that would be fine with me - i am not sure if this is workable though.  Because you probably want to show some labels even if they are close together (like New York and Washington DC) but do not want the same density of labels globally.  On the other hand at z=3/4 rendering performance is not an issue so you could reasonably use even an expensive SQL query to do the necessary processing.

> Currently, we treat not only national capitals (“capital=yes”), but also regional capitals of a givel level (“capital=4”) in a special way. So, maybe we could do this:
> 
> national capital → population × 3
> regional capital → population × 2
> all others → population × 1

Whatever formula you use - it will require some tuning.  This is best done with fast interactive feedback, i.e. not with the rendered map.

The more important questions you would need to consider is how you treat populated places with no population tag.  You could use the minimum number of a settlement of this class of course (i.e. city -> 100k)

And yes, if you render the low zoom levels based on a score calculated directly from attributes you could use the same layer for all zoom levels.  It might be somewhat inefficient w.r.t. rendering performance though.

sommerluk:
> You could use the minimum number of a settlement of this class of course (i.e. city -> 100k)

Yes. The current code of this PR uses 100 000 as default for place=city and 10 000 as default for place=town. The default is used when either there is no value or if the value is not valid (contains non-digit characters).

[I suppose I should catch special cases of wrong tagging like population=123456789123456789123456789 in the regular expression. The current regular expression considers them as valid, but they are too big for the integer datatype, and this leads to an SQL error.]

sommerluk:
> This is best done with fast interactive feedback, i.e. not with the rendered map.

Have you something special in mind?

pnorman:
> Indeed - this is one of the reasons it should be replaced. My favorite example is:
> 
> Miami: population 415000, shown
> Havana: population 2.1 million, capital of Cuba, not shown

What class are the two cities in NE?

imagico:
@sommerluk - you can do that in QGIS for example.

@pnorman - SCALERANK is 1 for Miami and 2 for Havana.

math1985:
INSTALL.md would also need to be updated.

pnorman:
Can you do a before/after out of the same database?

dieterdreist:
> Am 24.03.2015 um 20:44 schrieb imagico <notifications@github.com>:
> 
> You could use the minimum number of a settlement of this class of course (i.e. city -> 100k)


You would need regional min numbers though, in China a place with 100k is likely not considered a city

sommerluk:
@math1985 

> INSTALL.md would also need to be updated.

And also get-shapefiles.sh

sommerluk:
@imagico 

> Maybe we can get closer to the desired behavior by raising text-min-distance and at the same time render places with lower populations earlier. So, we get more collisions, and these collisions can be solved e.g. based on the population=* tag.

> If you can use Mapnik's collision detection to do the necessary proximity processing that would be fine with me - i am not sure if this is workable though. Because you probably want to show some labels even if they are close together (like New York and Washington DC) but do not want the same density of labels globally.

I’ve tried it. I had raised text-min-distance. But it doesn’t work well. Imagine the following situation:


Big city

Medium city
Small city


At low zoom levels, “Big city” will hide “Medium city” because it’s to close to “Big city”. However, “Small city” is displayed. At higher zoom levels, the distance between “Big city” and “Medium city” will be enough to display both, but the side effect is that “Small city” can’t be rendered anymore because it’s to close to the – now visible – “Medium city”. Result: “Small city” is e.g visible at z=4, not visible at z5–6, and visible again starting with z=7. I suppose that there will always be some of these inconsistent cases. But when I tested it, there were much more of these inconsistent cases when text-min-distance was raised.

sommerluk:
I’ve updated the code. Now, it’s using a score – as proposed by @imagico 

@pnorman 

Here are some screenshots of the updated code (left: current behaviour / right: behaviour of this PR). They are made out of a limited database with only the city nodes.

zoom3

zoom4africa

zoom4asia

zoom4europe

zoom5africa

zoom5europe

zoom6africa

zoom6europe

sommerluk:
Maybe the value should be tweaked, so that we have more labels at z6 (and maybe also z5).

Should there be a third font size for placenames-medium (current PR code: only 2 font sizes: high-score and low-score)?

math1985:
Current results look good to me, as far as I'm concerned this can be merged.

imagico:
Just wanted to mention that this change will likely increase the tendency to tag for the renderer with place classification and population numbers.  This is already an issue right now (see for example the number of obvious bogus points in http://taginfo.openstreetmap.org/tags/place=city#map)

A possible way to slightly counteract this would be to create an upper limit for population of place=town so pushing a small place significantly would require manipulating both population and place classification.

dieterdreist:
2015-04-27 12:33 GMT+02:00 imagico <notifications@github.com>:

> A possible way to slightly counteract this would be to create an upper
> limit for population of place=town so pushing a small place significantly
> would require manipulating both population and place classification.
>


this should not be done (or had to be implemented in a more complex way),
because population and place importance / classes are not the same
worldwide, e.g. China or India tend to have more population in less
important settlements compared to Europe.

sommerluk:
We could change the ORDER BY statement, so that first criterion is place=city/place=town, and the second criterion is the score that we have calculated. So place=city would have always a higher priority than place=town. Within each group, the elements would be orderd by score. Advandage: This approach would avoid to use a hard-coded threshold value and should be comprehensible for mappers (they get what they expect). (Minor) inconsistence: A city of 110 000 peoples without capital=* would have a higher priority than a town of 90 000 peoples with capital=*…

imagico:
This would avoid prioritizing a town with high population number over a city with lower one but the town would still have a high score and would be shown early if there are no competing cities.

And i disagree with @dieterdreist - place classification is not an importance rating, the wiki bases it primarily on the function of the place for the people around it and this is very similar globally for places with a certain number of inhabitants.  The population ranges of places with the various tags in different regions seem more a matter of local tagging habits than of actual differences in reality. So i think limiting the population of place=town to something like 200k in score calculation does seem appropriate.

pnorman:
We should be rendering city before town, even if the city has a lower population. Capital might change that, but that's a cartographic decision which I've seen done both ways

sommerluk:
This PR now renders cities always earlier than towns.

Exception: Towns that are capitals of a country may be treated like cities if their score is high enough.

Within each group, the elements are ordered by score.

pnorman:
I did a proper comparison for Europe, with full data loaded so there are the normal label collisions.

I was mainly interested in how it stacked up against Natural Earth. I'll do some more for higher zooms where it's already using OSM data

z3




z4

matkoniecz:
What is causing the difference in status of United Kingdom label on z3 and label for Poland on z4? It seems that this PR should affect only display of cities (and features colliding with city labels).

pnorman:
I think for Poland it is conflicting with the MA label, and [we set no order](https://github.com/gravitystorm/openstreetmap-carto/blob/master/project.yaml#L1358-L1367)

pnorman:
z5

old

imagico:
You could also test http://www.imagico.de/map/osm_populated_en.php#download which is using a preprocessing approach for comparison.

In general combining importance rated places with the crude hit-and-miss labeling approach of mapnik will always lead to bad results, especially if as here there are already a lot of labels in the map for admin units.  You end up skipping a lot of labels with high rating and possibly show a lot of labels with lower rating which counteracts the initial idea of having places rated.

One approach would be to interleave priorities of place and admin unit labels, i.e label national entities with priority but give important places priority over subnational entities.

sommerluk:
Rebased.

sommerluk:
> One approach would be to interleave priorities of place and admin unit labels, i.e label national entities with priority but give important places priority over subnational entities.

I do not know anything about how ordering works. Would it be (for example) enough to move the block “#placenames-medium::high-importance” to be before the block “.state” if we would want that “#placenames-medium::high-importance” renders always before “.state”?

daganzdaanda:
> One approach would be to interleave priorities of place and admin unit labels, i.e label national entities with priority but give important places priority over subnational entities.

Maybe we should have a discussion on the boards / lists or a survey about what people want to prioritize on the main map. I'd like to see the big cities earlier, and don't care so much about the admin-boundaries (or names of these), but the priorities or tastes might be different in other parts of the world. I like to compare styles, right now I feel the German style gets a lot right about city names.

Sorry if this is a stupid question: 
Would it be possible to have a list that decides which names get shown early on z=3/4?
Or maybe even later, allow some "editorial decisions" for difficult situations?

math1985:
Sorry for not coming back to this earlier. Is there any reason why this cannot be merged in its current form (apart from the fact that it needs rebasing)?

pnorman:
> Sorry for not coming back to this earlier. Is there any reason why this cannot be merged in its current form (apart from the fact that it needs rebasing)?

I'm not particularly happy with the priority at low zooms.

matkoniecz:
#1689 that was described as "Similar to #1461, but does not adjust low zoom or go as far." was merged.

It seems that this PR should at least be updated to fix merge conflicts, as part of this idea was implemented. 

It seems that "substitute NE usage on low zoom level" was not done in #1689

sommerluk:
Okay. I can take a look on a rebase in the next days.

However …
@pnorman

> I'm not particularly happy with the priority at low zooms.

I need something more detailed to propose an alternative

sommerluk:
Rebased.

pnorman:
> Rebased.

Far fewer cities being rendered, and I'm not sure the order is better.

sommerluk:
Okay. I’ve tweaked the values on low zoom levels. The number of visible cities should now be similiar to the current master.

kocio-pl:
@sommerluk Could you provide some test images for comparison?

sommerluk:
Left: before ## right: after

z3

z4

z5

HolgerJeromin:
fixes #482 too

mboeringa:
> Left: before ## right: after

Looks pretty good.

imagico:
Since the key problem is label collision with admin unit labels these samples do not really tell very much.  The selection seems fine but as long as higher level admin unit labels have priority over place labels this is not really going to work too well - no matter how smart the actual place labeling is.  For example the 'Brandenburg' label will always prevent the 'Berlin' label from showing.

kocio-pl:
Clear improvement - at least in Poland/Europe, but the biggest winner here is India!

pnorman:
I know it's a pain, but for a sensible preview, admin boundaries are needed too.

sommerluk:
> as long as higher level admin unit labels have priority over
> place labels this is not really going to work too well

Yes, absolutly.

But I would like to focus this PR to the internal placenames order. And maybe later in another issue we can discuss reordering the admin-placemenes relationship. I think this PR makes sense also without touching admin features.

sommerluk:
> as long as higher level admin unit labels have priority over
> place labels this is not really going to work too well

Yes, absolutly.

But I would like to focus this PR to the internal placenames order. And maybe later in another issue we can discuss reordering the admin-placemenes relationship. I think this PR makes sense also without touching admin features.

sommerluk:
> I know it's a pain, but for a sensible preview, admin boundaries are needed too.

Is there anyone who could do that?

(I’m not sure how the get the necessary data, however I fear it’s pretty much data and I’m living in West Africa and the internet connection is really much more limited here than in Europe … Getting the placenames took me half a day for the download)

sommerluk:
> I know it's a pain, but for a sensible preview, admin boundaries are needed too.

Is there anyone who could do that?

(I’m not sure how the get the necessary data, however I fear it’s pretty much data and I’m living in West Africa and the internet connection is really much more limited here than in Europe … Getting the placenames took me half a day for the download)

matkoniecz:
> I’m not sure how the get the necessary data,

I am using https://lists.openstreetmap.org/pipermail/dev/2015-July/028625.html for generating low zoom previews. But it is a 26GB download.

I may generate comparison, but after #1736 (currently I am constantly generating previews of various road renderings).

imagico:
You can probably get all required data for the lower zoom levels by filtering out all ways/relations with tag `admin_level<=4` - either via osmfilter from a planet file or via overpass api (have not tried that - might be too big) in addition to the places.  Then import this with osm2pgsql.

`osmfilter "$PLANET" --keep="admin_level<=4" --keep="place=city" --keep="place=town" --keep="place=village" -o="osm_admin_places.osm"`

generates a 6GB OSM XML file which i suppose should include all you need.

kocio-pl:
@sommerluk Thanks a lot for your work! Could you make some benchmarking - if there are still some labels which we should make visible - and try some fine tuning if needed? 

I think we could check especially:
- which country capitals are not visible and why?
- how about province/state capitals?
- any other big and important cities missing on the map?
- or maybe some labels are shown too early?

Is there a way to count it automatically (like probably with debugging features of Mapnik)? And maybe we could use computers to test changing parameters within specified range to show missing labels, to avoid tedious manual actions?

kocio-pl:
BTW: I think we could render country capitals labels different than the rest - in bold, with star or anything like that.

sommerluk:
Rebased again and melted to one single commit.

About benchmarking, rendering priority in general and capitals in special: Basically, this PR tries to base the priority on the OSM data for cities and towns. (Smaller places not considered). It gives priority by using a score which depends on the population, but capitals get additional bonus score. So a capital will always have a higher priority than a neighbour city with the same population. I’ve than played around with the threesholds until the overall placenames density on each zoom level was more or less similar to the current rendering.

If there is a consens to render capitals different, I could adopt the code. But this could also wait for a separate PR.

The next few days I’ll stay at a place where is no Internet connection. When I’m back I’ll look at this issue again …

kocio-pl:
OK. 

I hope for some automatic benchmarking because checking if all the capitals are visible is possible, but not convenient, while checking anything below them by hand is just crazy. Comparative statistics would be good estimation of success rate: how many of capitals/cities was shown before (I'm counting what _would be seen_ excluding other effects like province/state labels) and how many will be with this PR.

Rovastar:
I cannot tell from any of the pics in the thread but surely we want to see country names ahead of cities. I wouldn't want to display cities at these zooms without countries.

HolgerJeromin:
fixes #482 and fixes #50

sommerluk:
@Rovastar This PR does not make any changes to country names or boundaries. I just did not download them into the database for the examples because the boundry data is really big and my Internet connection is slow.

pnorman:
> This PR does not make any changes to country names or boundaries

Admin labels (`placenames-large`) will remain the same, but their presence will have an impact on the layers that are changed.

I'll see if I can do something to make a larger preview available, but the road changes are currently my priority.

matkoniecz:
One thing that is certainly much better is unifying style for city names and new text-wrap-width.

Due to low zoom levels generation of comparison images is slow (and way too slow to compare anything in TileMill) so according to limited testing some places would be better, some - especially ones with high population density like UK would be worse as lower amount of placenames would be rendered.

Overall, I think that reducing usage of NE data is worth even some decrease in rendering quality - though I want to continue testing.

sommerluk_lowzoomcities10 4 14 master - sommerluk_lowzoomcities10 500px
sommerluk_lowzoomcities10 4 10 master - sommerluk_lowzoomcities10 500px

(note z4 - is weirdly rendered both ion before and after image - I ma not entirely sure what is causing this problem).

sommerluk:
Thanks @matkoniecz !

matkoniecz:
At this moment my main complaint is significantly reduced number of displayed labels on z6. In Romania it works well - I am unsure why cities there are now so prominent. But in Germany, Poland, Hungary, Italy current situation is better.

sommerluk_lowzoomcities10 50 20 6 6 v2 34 0 - sommerluk_lowzoomcities10 1000px

Also, why Brookhaven is selected over New York on z5?

sommerluk_lowzoomcities10 a 4 8 master - sommerluk_lowzoomcities10 150px


Note that comparisons below are between PR proposal and modification of current master that modifies styling of admin area alables and labels for cities from NE - as in comparisons to current master the most visible change is modification of how placenames are styled on lower zoom levels.

BTW, it seems that decreasing text-wrap-width for admin areas improves rendering and reduces number of label collisions.

Australia: https://cloud.githubusercontent.com/assets/899988/9564098/231a994e-4e9b-11e5-9970-713703f80c70.png (centred on [26, -135])

North America: https://cloud.githubusercontent.com/assets/899988/9564100/2322ebf8-4e9b-11e5-98e5-2674cd7b8b2d.png (centred on [40, -75])

East Europe: https://cloud.githubusercontent.com/assets/899988/9564102/23233cac-4e9b-11e5-8bf2-22b626aa26f5.png (centred on [50, 40])

Central Europe: https://cloud.githubusercontent.com/assets/899988/9564101/232311be-4e9b-11e5-9ae6-927b15cd20a7.png (centred on [50, 20])

Major cities (like New York on z6, z7, z8) are appearing, NE data is no longer used. Except z6 and z5 New York I consider it either improvement or not worse (or I am unable to evaluate changes - for example in the China).

matkoniecz:
India: https://cloud.githubusercontent.com/assets/899988/9564213/077c75fa-4e9f-11e5-900e-3cb14c379ce5.png (main problem is Hyderabad - it seems that it should be far more prominent)

Note - to generate previews I am using https://lists.openstreetmap.org/pipermail/dev/2015-July/028625.html (therefore data used to generate maps is quite old)

kocio-pl:
> Also, why Brookhaven is selected over New York on z5?

Do we have some tool for debugging Mapnik exactly to tell us "within this set of points - which ones are not rendered and why"?

sommerluk:
Sorry for the late response, my hard disk was broken and I could not test further things.

About the issue with New York/Brookshaven: I have no idea why this happens. Maybe a data issue? I’ve tried to reproduce it with my cities-and-town-only database, and also with new-york-latest.osm.pbf (from last week) from Geofabrik. With both, I could not reproduce this issue.

I’ve tuned now z6 to be more similar to the current situation on z6.

sommerluk:
Solves #1083

matkoniecz:
> About the issue with New York/Brookshaven: I have no idea why this happens. Maybe a data issue? I’ve tried to reproduce it with my cities-and-town-only database, and also with new-york-latest.osm.pbf (from last week) from Geofabrik. With both, I could not reproduce this issue.

Maybe it is data issue that is now fixed.

HolgerJeromin:
@sommerluk The "fixes #xy" things must be in the commit message or in the first comment/description from you. Please add my [issues](https://github.com/gravitystorm/openstreetmap-carto/pull/1461#issuecomment-132109874), too :-)

Teester:
I've applied this patch to my local dataset (Ireland).  It is an area where few towns have populations above 10000, so with this patch as it stands, many towns with no population data are superseding towns with population data, by virtue them being assigned a population of 10000 by the patch.  So, major towns with population data are not appearing in favour of smaller towns which have no population data.

Changing the population assigned to 1000 seems to resolve the issue, for Ireland at least. I have not noticed any problems with cities displayed in my dataset.  Is there information available on average town population in OSM or proportion of towns with population included.

matkoniecz:
@Teester Thanks for testing!

Probably setting it low and encouraging to tag population for at least major towns is better than setting it high and expecting that every single small place=town will be tagged with population data.

kocio-pl:
+1

sommerluk:
@HolgerJeromin Thanks for the hint!

sommerluk:
The fallback score for towns without population key is now 1000. Thanks for testing!

matkoniecz:
100 000 for cities is also too high - there is plenty of place=city with lower population  - see http://overpass-turbo.eu/s/bqR

imagico:
Selecting a low fallback population number will probably not be much encouragement for tagging correct populations but likely significantly push tagging for the renderer with the place classes (if a town does not show up it is much more straight forward to tag it place=city instead of finding out the population number).

The most encouraging approach for the mapper for correct tagging overall is quite surely interpreting the tags exactly as they are supposed to be used.

The  most important problem of the approach here is that is does not display smaller places even if they are far away from other places.  Like:

http://www.openstreetmap.org/node/244081999
http://www.openstreetmap.org/node/31219396
http://www.openstreetmap.org/node/29341070

The first two are correctly tagged as city, the latter is tagged for the renderer.  They could all well show up at z5 although they do not.  In the first case tagging the correct population will probably help, in the other two probably not.

Apart from that i still think major cities should have priority over subnational admin entities.

sommerluk:
I’m not so sure if this is a good idea. For towns,
http://wiki.openstreetmap.org/wiki/Key:place says: “… often with a
population of 10,000 people and good range of local facilities… In
areas of low population, towns may have significantly lower
populations.” This is a good reason to use a fallback score below 10
000.

However, for cities, the very same wiki page says: “Should normally
have a population of at least 100,000 people and be larger than nearby
towns.” I know that the place value is not just population-dependent,
but following the wiki a place=city with a population lower than 100
000 should be an exception.

2015-09-14 9:36 GMT+02:00, Mateusz Konieczny <notifications@github.com>:
> 100 000 for cities is also too high - there is plenty of place=city with
> lower population  - see http://overpass-turbo.eu/s/bqR
>
> ---
> Reply to this email directly or view it on GitHub:
> https://github.com/gravitystorm/openstreetmap-carto/pull/1461#issuecomment-139985018


-- 
Lukas Sommer

pnorman:
> I’m not so sure about this. For towns, http://wiki.openstreetmap.org/wiki/Key:place says: “… often with a population of 10,000 people and good range of local facilities… In areas of low population, towns may have significantly lower populations.” This is a good reason to use a fallback score below 10 000.
> 
> However, for cities, the very same wiki page says: “Should normally have a population of at least 100,000 people and be larger than nearby towns.” I know that the place value is not just population-dependent, but following the wiki a place=city with a population lower than 100 000 should be an exception.

The 10%, median, and 90% population values for place=city are 30k, 131k, and 800k. For town they are 2.4k, 12k, and 45k.

math1985:
Thanks a lot for the PR @sommerluk! And sorry for the show review process.

I think this is a significant improvement in terms of code, so I'm happy to merge it. We can always finetune later.

By the way, it seems you are sorting by category both in the stylesheet (wit attachments) and in the SQL statement. Wouldn't sorting at only one place be sufficient?

sommerluk:
Thanks @math1985 for merging.

> you are sorting by category both in the stylesheet (wit attachments) and in the SQL

Good catch! PR at #1843 …

yopaseopor:
Sorry but #1083 still happens. Still happens at level 9 http://c.tile.openstreetmap.org/9/258/191.png (the first city I talk about does not appear). Level 10 fixed.

pnorman:
This commit is not in the version used on tile.openstreetmap.org as it's not in a released version yet.

PR #1466, opened 2015-03-29
Toilet dump stations

brycenesbitt:
It's not possible to completely map a typical campground without mapping the sewer dump point for RV's and boats.  OSM has not had very many of these mapped to date perhaps because of the strength of http://www.sanidumps.com/ and perhaps because of lack of rendering.
http://wiki.openstreetmap.org/wiki/Tag:amenity%3Dsanitary_dump_station

Note that some campgrounds are also tagged sanitary_dump_station=yes, which is not part of this pull request.

There are hundreds of thousands of such stations in the world, most unmapped in OSM.
This is also a feature that can draw a certain set of enthusiasts to OSM.


matkoniecz:
0) amenity=sanitary_dump_station is used 59 times worldwide, too early for rendering (usage less than 1000 times, not some special case) http://taginfo.openstreetmap.org/tags/amenity=sanitary_dump_station
1) Three svg files are added, only one is used
2) Have you tested whatever this code works? It is almost certain that some code is missing

brycenesbitt:
0) There are some 400 mapped right now.  There are tens of thousands of these stations all over the world.  It's special in that this is a common map symbol in campgrounds and natural areas: any complete campground map would have it, and there are thousands of camp grounds in OSM.  As Andy Allen says “render and they will map”.

1) The three icons follow the Maki convention of offering icons optimized for various pixel size renderings.

2) No.

brycenesbitt:
Various hacks have been used to get this tagged and in some cases rendered:

man_made=wastewater_treatment_plant
man_made=pumping_station
amenity=rv_dumpstation,
amenity=waste_disposal,waste=excrement
amenity=pumpout
amenity=recycling,note=RV Dump Station
name=Dumping Station
amenity=caravan_site,note=RV Dump Station
dumping_site
dumpsite
recycling:excrement
amenity=waste_basket
waterway=pumpout
amenity=pump_out

People appear to have been "tagging for the rendering", working around the lack of osm-carto rendering support.

aceman444:
http://wiki.openstreetmap.org/wiki/Tag:amenity%3Dsanitary_dump_station only shows 194 uses today. How many occurrences of the other tag hacks can you count as being this feature?

aceman444:
Yeah, if you also count sanitary_dump_station=yes then there can be above 400.

brycenesbitt:
@aceman444 I'll count the potential current size of this tag space on Monday.


But that said, I don't think the number of uses is at the important consideration for rendering.  What's important is how useful the feature if if rendered, how commonly mapped on other maps, or the potential for attracting new mappers or new map viewers.

There are features with tens of thousands of instances that are not good choices to render.
There are features with zero instances, where the project would benefit if rendering support were added.

--
In addition: rendering Sanitary Dump Stations would encourage users to converge on a single tag for the feature, and may help encourage more camping mapping.

aceman444:
I am sure nobody questions the usefulness of mapping this facility. But the "number of uses" rule probably wants to determine whether users actually decided on the precise tagging (the syntax). Otherwise how can anybody confirm that really amenity=sanitary_dump_station is the chosen tag and all programs should add a rule for this exact tag. It is only your word (and some ~20 voters) for now. But this is a generic opinion, nothing personally against you :) I had the same complaint when JOSM bended and added support for amenity=bicycle_repair_station that also had no prior use.

brycenesbitt:
About half the waste=excrement are this feature, as are most recycling:excrement, and waterway=pump_out, along with seamark:small_craft_facility:category%3Dpump-out.

Mostly gone tags include amenity=dumpstation, amenity=sanitary_station, amenity=pumpout, amenity=pump-out and amenity=pump out along with amenity=dump station, amenity=dump, and amenity=RV dump, amenity=recycling,waste=excrement, and finally tourism=caravan_site,note=RV Dump.

Fizzie41:
Just came across this & #1507 as I started searching for caravan waste dump points.

According to taginfo https://taginfo.openstreetmap.org/tags/?key=amenity&value=sanitary_dump_station, the count is now at almost 3000 (2899 to be precise), so could I please suggest that this matter be reopened?

kocio-pl:
I guess we need an icon (or some other form of rendering) to really think about it.

The usage looks high enough and it's growing:

taghistory 46

brycenesbitt:
The use case for the sanitary dump station is not (yet) that it's useful
for finding dump stations.
There are specialized websites with far more comprehensive listings,
prices, terms and user reviews.

But if you're creating a detailed map of a given camp site or area, the
dump station is a very relevant feature
that you'd want to show.  OSM is often the best source for maps of such
camping and natural areas.

---
bicycle_repair_station is a tag that probably should have gone through more
process.
The problem is people confuse the tag and map bike repair shops.  The tag
was intended only for unattended tool stands.

sanitary_dump_station however seems fine: the tag has momentum, and it
seems to be correct momentum.

Fizzie41:
> I guess we need an icon (or some other form of rendering) to really think about it.

A few examples were discussed in one of the earlier posts

Good ones would be either https://pixabay.com/en/dumping-waste-mobile-home-trailer-99231/
image
or
https://goo.gl/images/4sTTbz
image

kocio-pl:
We have only 14 px matrix to use, so it has to be designed with such small size in mind, usually scaling down icons doesn't work good.

Adamant36:
I was thinking something like the current caravan symbol with a down arrow in the back. I think it would be doable in 14x14, but maybe not with the square part added to it. If you don't include the caravan it risks being to abstract.

kocio-pl:
Or just a half of caravan.

Fizzie41:
> usually scaling down icons doesn't work good.

Understood

> the current caravan symbol with a down arrow in the back.

"In the back" do you mean at the back of the van, or behind it as you're looking at it? Would an arrow underneath work?

brycenesbitt:
The attached pull request from the original request has SVG versions of the symbol.

Adamant36:
> "In the back" do you mean at the back of the van, or behind it as you're looking at it? Would an arrow underneath work?

Underneath it you only have like 5 pixels down or sideways to work with, if even. If you cut the cravan in half though by getting rid of the front half, then move the back half to the right side of the canvas, you have a bunch of room left it to do the arrow and maybe even the square. With a lot of caravans the hookups aren't directly underneath them anyway. Their usually about at the back around bumper height or in the middle at the same level. You don't have to crawl under the caravan to access them though.

Fizzie41:
Just reading a camping article https://www.freerangecamping.com.au/directory/item/toowoomba-showground-cg/?utm_medium=email&utm_source=newsletter&utm_campaign=e-update43&utm_content=toowoomba-showground :-)

If you scroll down to the bottom of the page, they have all the icons that apply to that Park, including a dump point.

They've used their normal caravan icon, but drawn it "half size" so they can also include the arrow down into the tank underneath it.

Would that work, or would it make the caravan too small to recognise?

PR #1468, opened 2015-04-01
move caravan parking icon to SVG

nebulon42:
Implements part of #1165.
Symbol is now smaller.

before
caravan_before

after
caravan_after


kocio-pl:
The shape is now much harder to recognize - could it be longer and more rectangle?

mboeringa:
I am OK with it considering the restrictions on icon size, although it may indeed take a little bit more time to figure out what it is. But it is still recognizable, especially when combined with labelling and possible context (normal camping nearby) or a legend. People can get used to it I think.

lest69:
I agree with @mboeringa. The existing one is more recognizable, but the proposed one is probably the best we can do with the limited space and should be okay in context.

nebulon42:
What is blocking this PR?

math1985:
I'm not fully convinced the icon is recognizable enough.

For comparison, this is a non-pixel perfect 16x16px version of the old icon:

[!http://www.sjjb.co.uk/mapicons/png/accommodation_caravan_park.p.16.png](caravan)

math1985:
I'm not fully convinced the icon is recognizable enough.

For comparison, this is a non-pixel perfect 16x16px version of the old icon:

caravan

The two icons side by side:
caravan_park 16 caravan

nebulon42:
The oblong shape is probably more recognisable. Updated.

kocio-pl:
That was what I suggested and I like the result - +1.

mboeringa:
> The oblong shape is probably more recognisable. Updated.

Great.

gmbo:
I mean, this symbol is incorrect because caravan_site is a parking space for campers. There caravans are not or rarely allowed.
it should be used there a symbol for campers to avoid errors when tagging and the mapping use.

[caravan_site](http://wiki.openstreetmap.org/wiki/DE:Tag:tourism%3Dcaravan_site)

Likewise, the tent symbolizing camp_site is not correct. There likely this caravan icon should be used. These general campsites are mostly used by caravans, as well as tents and campers are welcome.

[camp_site](http://wiki.openstreetmap.org/wiki/DE:Tag:tourism%3Dcamp_site)

DE:ich meine, dieses Symbol ist falsch, da caravan_site einen Stellplatz für Wohnmobile darstellt. Dort sind Wohnwagen nicht oder nur selten erlaubt.
es sollte dort ein Symbol für Wohnmobile verwendet werden um Irrtümer beim Taggen und der Landkartenbenutzung zu vermeiden.

Ebenso ist das Zeltsymbol für camp_site nicht richtig. Dort sollte eher dieses Caravan-Icon genutzt werden. Diese allgemeinen Campingplätze werden hauptsächlich von Wohnwagen genutzt, aber auch Zelte sowie Wohnmobile sind hier willkommen.


[File:Caravan_site.svg](http://wiki.openstreetmap.org/wiki/File:Caravan_site.svg)

pnorman:
> I mean, this symbol is incorrect because **caravan_site** is a parking space for campers. There **caravans are not or rarely allowed**.

Quoting from the Wiki,

> A caravan site, caravan park or RV park is a place where people with caravans / motorhomes / recreational vehicles can stay overnight, or longer, in allotted spaces known as "pitches" or "sites".

Your view disagrees with the plain language meaning of caravan_site and the wiki documentation.

Issue #101, opened 2013-08-06
ele with alpine_hut when mapped as a building


al--:
Wish: If a building is tagged as tourism=alpine_hut, it would be fine if the ele was shown (as is done if it is mapped as a node).

![91464 1](https://f.cloud.github.com/assets/1458456/916010/7fc54ae0-fe7f-11e2-8678-e2866fa3fcd5.png)
http://www.openstreetmap.org/browse/way/232521425
![91450 1](https://f.cloud.github.com/assets/1458456/916026/dd974196-fe7f-11e2-8116-0b266eec7599.png)
http://www.openstreetmap.org/browse/node/389786530

pnorman:
Reopening, I think there's still some changes that need to be done to enable this.

math1985:
After #2823 is merged, this will be the oldest open issue!

kocio-pl:
@sommerluk I guess you could extend elevation rendering for polygons or help me with this? 

I've tried to copy this block:

https://github.com/gravitystorm/openstreetmap-carto/blob/8862ee40103ce9addd0b45c79dcfcd436c3dcd91/project.mml#L2020-L2024

and this one (without "natural" part):

https://github.com/gravitystorm/openstreetmap-carto/blob/8862ee40103ce9addd0b45c79dcfcd436c3dcd91/project.mml#L2085-L2092

to a `text-poly` layer, but then I get:
```
Postgis Plugin: ERROR:  column "elevation" does not exist
```

The `text-point` layer doesn't have this problem.

sommerluk:
I’m not sure. Could you make the code available on Github?

Just raw guess: `text-point` is composed by a `SELECT` and a sub-`SELECT`. The code for the name-elevation-combination is part of the outer `SELECT` and needs the column `elevation` which does _not_ come from the database (the database only provides `ele`), but it created in the inner `SELECT`. `text-poly` does not have nested `SELECT`. If you haven’t created one, you cannot query `elevation` at this level.

kocio-pl:
You were right probably, I didn't spot the SELECT is nested. Could you try to fix this issue?

PR #1472, opened 2015-04-05
Elevation rendering

### LINKED ISSUE #1084: Render peaks in order of eleveation
variant of #1083

Peak with untagged height currently may block one where height is tagged, lower may block label of higher one. In case of peaks showing on the same map ordering by elevation should work in nearly all situations.

Peaks below see level may be in theory troublesome - ordering by elevation may not work, missing ele tag should be assumed to be below lowest possible peak height to ensure that tag with missing ele tag will not be considered more important by one where peak is below see level. But at least currently these problems are theoretical. It seems that such peaks are rare http://www.openstreetmap.org/?mlat=31.1022&mlon=35.3796#map=13/31.1022/35.3796 

see http://www.openstreetmap.org/node/385115017#map=14/49.1838/20.0878 (Česká veža is not tagged with ele tag)

selection_005

compare http://www.openstreetmap.org/node/385115017#map=15/49.1868/20.0864

selection_004

sommerluk:
Refactor elevation rendering.

Interpretats the content of “ele”, instead of simply treating it like a text. Accepts standard floating point values. Ignores other values. Rounds to full integers. Renders the elevation with “ m” to clarify that it is measured in meters (using a non-breaking space) – IMO especially usefull as feedback for mappers in the USA. In case of negative values, uses the typografically correct minus sign. Uses a SELECT within a SELECT because this avoids duplicate querries.

Orders by elevation (from high to low) all features that are rendered with their elevation. Seems to be a good idea for peaks. Resolves #1084. Seems also to be a good idea for volcano and saddle. Is it also a good idea for alpine hut and shelter, or should it be here even the opposite (low to high)? I suppose the order would break the icon-text-missmatch that was described here: https://github.com/gravitystorm/openstreetmap-carto/pull/1349#issuecomment-77805678 ?


matkoniecz:
> good idea for alpine hut and shelter, or should it be here even the opposite (low to high)

I am not sure is it even changing anything here, is there any place where huts/shelters are so close that it is affecting anything? Or is it also applied for alpine hut/peak collisions?

Some before/after images would be a good idea.

matkoniecz:
In 1349 mismatch was caused by combination of 
- lack of rendering of label for post office
- higher priority for post_office icon than variety_shop icon

As result there was an empty space below icon where shop label from the same osm element was displayed. This kind of mismatch should not affect peaks etc neither before nor after this PR.

matkoniecz:
I am not sure is it the best idea to accept floating point and convert it to integers (converting it to integers seems a bit of magic that is not truly necessary and likely to confuse people).

matkoniecz:
Sorting of ele seems to affect only labels but it should be also applied to icons. Note that alpine huts and shelters may be also areas.

I would also consider splitting this into two separate PRs, especially as second part is currently discussed on mailing list.

matkoniecz:
Useful Overpass querries:

utterly wrong ele tags: http://overpass-turbo.eu/s/8zi
wrong ele tags: http://overpass-turbo.eu/s/8zk

project.yaml
```diff
@@ -1846,47 +1846,70 @@ Layer:
...
-              'waterway_' || CASE WHEN waterway IN ('dam', 'weir') THEN waterway ELSE NULL END,
-              'tourism_' || CASE WHEN tourism IN ('attraction', 'viewpoint') THEN tourism ELSE NULL END
-            ) AS feature,
+            way_pixels,
+            feature,
             access,
-            (CASE WHEN "natural" IN ('peak', 'volcano', 'saddle') OR tourism = 'alpine_hut' OR amenity = 'shelter' THEN CASE WHEN ele IS NOT NULL THEN CASE WHEN name IS NOT NULL THEN CONCAT(name, E'\n', ele) ELSE ele END ELSE name END ELSE name END) AS name,
+            CONCAT(
+                name,
+                CASE WHEN (name IS NOT NULL) AND (elevation IS NOT NULL) THEN E'\n' ELSE NULL END,
```
pnorman: the `(` and `)` shouldn't be needed here

project.yaml
```diff
@@ -1846,47 +1846,70 @@ Layer:
...
+                  'power_' || CASE WHEN power IN ('station', 'generator', 'sub_station', 'substation') THEN power ELSE NULL END,
+                  'boundary_' || CASE WHEN boundary IN ('national_park') THEN boundary ELSE NULL END,
+                  'waterway_' || CASE WHEN waterway IN ('dam', 'weir') THEN waterway ELSE NULL END,
+                  'tourism_' || CASE WHEN tourism IN ('attraction', 'viewpoint') THEN tourism ELSE NULL END
+                ) AS feature,
+                access,
+                CASE
+                  WHEN "natural" IN ('peak', 'volcano', 'saddle') OR tourism = 'alpine_hut' OR amenity = 'shelter' THEN
+                    CASE
+                      WHEN ele ~ '^-?\d{1,4}(\.\d+)?$' THEN CAST (ele AS numeric)
```
pnorman: Why `numeric`?

Also, we've been using the PostgreSQL :: syntax more than CAST. I know the latter is in the SQL standard and the former is not, but :: works better in the other cases where Mapnik is substituting in text.

sommerluk:
> I am not sure is it the best idea to accept floating point and convert it to integers (converting it to integers seems a bit of magic that is not truly necessary and likely to confuse people).

Yes, that’s a little bit …

The idea is to round values. That means, 12.4 becomes 12. But 12.6 becomes 13 (and not 12). That’s the reason why I want to accept the floating point. The intuitive way to round it is ROUND(). Following to the PostGre Documentation, the return type of ROUND is “numeric”. I preferred to have “integer” as type – just to make sure that there are no precision problems when converting to text (I don’t want to be supprised by something like 12.00000000000000001) – but maybe that’s a rather theoretical concern. The type cast from “numeric” to “integer” rounds the values correctly, but you are right that that’s not obvious.

pnorman:
> I preferred to have “integer” as type – just to make sure that there are no precision problems when converting to text (I don’t want to be supprised by something like 12.00000000000000001) – but maybe that’s a rather theoretical concern

Unless you're worried about precision problems in *Mapnik*, not PostgreSQL, that's not a concern.

nebulon42:
Regarding ordering alpine huts by elevation: I think this is a rather arbitrary importance measure. For peaks this works, but I can't imagine why lower or higher huts should be more important.

HolgerJeromin:
I would propose rendering all digits to find bogus data like ele=0.000000000

sommerluk:
I’ve started yet doing some changes. Just a few questions:

> Note that alpine huts and shelters may be also areas, not only points.
I supppose that https://github.com/gravitystorm/openstreetmap-carto/issues/101#issuecomment-25358923 is still valid. (No ele in database for ways) Is this correct?

Can somebody explain me how the rendering priority works in detail? How does the SQL querry, the classes in the .mss files and the layers in the .mss files work together? Does the the order of the querries/layers in .yaml matter? Does the order of layers and/or classes in the .mss files matter? Does the order of the list of .mss files in .yaml matter?

sommerluk:
Okay. Here is the new version with the following behaviour:

– Huts and shelters are not ordered by elevation. Only peaks, sattles and volcanos are ordered by elevation.

– The values are rounded to integer. This is no longer done by a cast, but by an explicit ROUND() call, which makes the code easier to understand.

– Now both, the text and the icon, are ordered.

– There is still no support for elevation on polygons, because the current openstreetmap-carto.style imports the elevation only for points.

– Here a before-after-image:
compare

matkoniecz:
> There is still no support for elevation on polygons

It should not be a problem. Peaks, saddles and volcanoes are anyway rendered only for points (it affected huts and shelters).

I still think that ele=1.9 should be rendered as 1.9, not 2.

Interesting cases for testing: peak with ele=10 near
- peak without ele value
- peak with invalid ele value
- peak with ele=1,8 (should it be considered as valid? Unfortunately ele proposal failed to define is it OK tagging)

sommerluk:
> Interesting cases for testing

Peaks without ele value and peaks with invalid ele value are both ordered with lower priority than valid ele values. I’ve tested this.

> peak with ele=12,8 (should it be considered as valid? Unfortunately ele proposal failed to define is it OK tagging)

For most other keys that take numbers, the documentation requires that “.” is the decimal separator. I think it would be reasonable to apply this established rule also for elevation. Otherwise, we would also have big problems: How to determine if 123,456 is one houndrid twenty three meters and four houndrid fifty six millimeters or if it is one houndrid twenty three thousand four houndrid fifty six meters? While a human who knows a bit about geografiy knows that there will be no peaks with more than 9000 m on earth, it requires a lot of magic to interpretate this. And for the mappers it would not be obvious which combination is valid and which not. Also we have wordwide 379 297 elements with a combination of ele and natural key (following taginfo), but at only 4 045 nodes with these two keys the pattern [ele~","] applies. I would clearly stay with treating it as invalid.

> I still think that ele=1.9 should be rendered as 1.9, not 2.

Do you think it should still be rounded, but guarding one digit after the decimal separatos (1.91 would be rendered as 1.9). Or do you think it should not be rounded at all – independent of the number of digits after the decimal separator? As there are elements like http://www.openstreetmap.org/node/3018784788#map=19/60.52291/15.07693 I would prefer to round at least a little bit.

matkoniecz:
I would consider ele=215.09677099999999 as (tagging error)[http://www.wolframalpha.com/input/?i=1%2F10^14+m] that should not be masked.

> Or do you think it should not be rounded at all

yes

project.yaml
```diff
@@ -1843,50 +1852,79 @@ Layer:
       <<: *osm2pgsql
       # Include values that are rendered as icon without label to prevent mismatch between icons and labels,
       # see https://github.com/gravitystorm/openstreetmap-carto/pull/1349#issuecomment-77805678
-      table: |-
+      table: |2-
```
pnorman: don't use an explicit number of spaces (`|2-`) but use an implicit instead (`|-`)

pnorman:
I'm okay with rounding. `ele=215.09677099999999` is a tagging error, but `ele=1234.5` isn't, and it's a reasonable cartographic choice to not display decimal places.

pnorman:
Our method of rendering icons then names on a second pass is annoying here.

image
image

The north peak is never going to be rendered at the lower zoom because the south peak triangle gets rendered before the north peak text. I don't think there's a solution to this, but the elevation sorting will have less effect than might be expected.

HolgerJeromin:
Perhaps we can display the full tagging error (as `ele=215.09677099999999`) if
`character_length(ele) > 6` and make a nice rounding if the content is shorter (more likely valid).

HolgerJeromin:
should i reopen a new issue for the character_length idea?

PR #1476, opened 2015-04-09
move wind power symbol to SVG

nebulon42:
Implements part of #1165.

before
wind_power_before

after
wind_power_after


kocio-pl:
Could it be more rotated (to not overlap one part with another)? Otherwise I think it's OK.

nebulon42:
Unfortunately not, otherwise the blades would have to be shorter.

kocio-pl:
For me it's hard to say if I could recognize it not knowing what's this shape, we need more feedback probably.

math1985:
Tester (unfamiliar with osm) recognizes the new icon as wind turbine.

mboeringa:
Given the restricted icon size, I think this is a pretty good shot at it. It may be slightly less recognizable, but I never really liked the huge current ones, as they are so big as to overlap to much other stuff or each other.

PR #1477, opened 2015-04-09
move courthouse icon to SVG

nebulon42:
Implements part of #1165.

before
courthouse_before

after
courthouse_after


kocio-pl:
Very nice, but I would make shoulders a bit stronger (more visible) if possible.

mboeringa:
Looks entirely fine to me.

HolgerJeromin:
nice and sharp. thanks

PR #1478, opened 2015-04-11
move bus icons to SVG

nebulon42:
Implements part of #1165.

Bus stop
Bigger symbol as before.

before 
bus_stop_before

after
bus_stop_after

Bus station
Remove shielded version as most other symbols are without shield.
Note that I'm not satisfied with legibility on the noisy background, but since #1189 was not considered there is not much I can do about it. For consistency reasons I prefer dropping the shield nevertheless.

before
bus_station_before

after
bus_station_after


mboeringa:
I like the top bus icon, less so the lower one for bus-station, which actually seems lower classes

@nebulon42, I hope you don't mind my 'bitmap variations' based on your proposals, but would something like this be realizable in SVG? And what is your opinion about this variation? To tackle the problem of the "station" aspect, I added a "roof" in the left icon. So the left icon represent the bus station, the right one a simple stop.
bus_icons_normal
bus_icons_enlarged

math1985:
Bus stops are quite common, so I think we should consider keeping a smaller bus stop icon.

kocio-pl:
@math1985: +1

I also like the "roof" variation, because we tend to make similar things looking alike and only differentiate them by some details (like adding key or people silhouettes). Both new SVG icons are good for me, but it would be much harder for me to tell if that's a bus stop or a bus station. But maybe we can use the same icon, keep the z16+ / z17+ difference and probably make bus station label bold?

nebulon42:
@mboeringa Thanks for the suggestion. I'm not able to follow your "lower classed" perception. My intention was to get away from the irregular differentiation (shield vs. no-shield) towards showing the difference by depicting a bus and a coach. The coach is intended to look more streamlined, but maybe I didn't succeed in that. In your variation you used the icon you liked less? Or did I understand that wrong?

I did this differentiation also for commuter train (railway halt) and express train (railway station), see [my repository](https://github.com/nebulon42/osmic).

I certainly don't mind this bitmap variations but you have to keep in mind the canvas size. Here you just added to the icon size, which is not possible in the SVG and that may lead to all sorts of other necessary adjustments because the coach needs to be smaller with the roof added. More helpful for me would be a variation that respects the canvas size, it hasn't to be a SVG though.

I'm not sure if the particular roof conveys the station character recognisable enough and it would certainly have to be more prominent.

@math1985 Yes, a smaller icon for bus stop might be necessary but I still have to enter the world of different icon sizes.

@kocio-pl "much harder to tell which is which" Is it really obvious now that the bus with the shield represents a bus station? Or is it just familiarity?

kocio-pl:
For me shield version is visually "stronger" than no-shield version.  If not the shield, it would be also hard for me to recognize.

mboeringa:
> . The coach is intended to look more streamlined, but maybe I didn't succeed in that.

I hadn't recognized it so much as that, and was mainly judging based on icon size (where the top "normal busline" icon is actually bigger and suggests being related to something bigger).

> Here you just added to the icon size,

Actually, I thought I didn't. The icon on the left with the "roof" is 14x14 pixels, wasn't that the maximum allowable? (I think I counted the "normal busline" icon on 14 pixels high too?).

Yes, I used the icon I less liked because of the size limitations (it was less high to begin with), but adjusted it a bit to lower it even more to allow for the roof. This also lead to a more pleasing headlights depiction, with flat instead of square ones (IMO...), and a relatively bigger window, which actually comes closer to your intention of depicting a "coach" (well, at least that is my subjective feeling).

nebulon42:
Ok, I thought the right version was unmodified. From that I judged that you just added the roof. The original icon does not have square headlights though (https://github.com/nebulon42/osmic/blob/master/transport/bus-station-14.svg). Maybe the windshield should be bigger, yes.

math1985:
@nebulon42 Did you have any time to look at this issue?

nebulon42:
I'm not really convinced by the roof and a smaller version of the bus is on my list. @math1985 What is your opinion on the coach vs. bus design approach?

math1985:
amenity=bus_station is also used for local bus stations (central bus stops of the local bus network) in the areas where I am mapping, so the coach/bus distinction is not intuitive for me.

nebulon42:
I'll create separate PRs for the icons and am therefore closing this.

math1985:
I would like to have see bus_station converted to SVG. I have some other changes that depend on it.

Would the bus inside a square or circle fit?

kocio-pl:
I was thinking exactly about it! I just need a visual proof that it really works.

nebulon42:
I'm not a big fan of circles and squares as they are a bit arbitrary and are not immediately intuitive. I'd like to have them only as last resort when nothing else works.

Since bus stop now uses the smaller icon one immediate solution would be to use the bigger version for bus stations. Maybe I will also succeed in depicting two busses for bus station.

kocio-pl:
For me they're not that exceptional, just making icon stronger like bolding the text (that's why I made a clinic icon simply by removing the circle from the hospital to show it's less important entity), but have no idea what would work and have no strong expectation other than bigger version probably will not be too clear. Maybe two buses eclipsing each other like the masks in the theater?

daganzdaanda:
> Maybe two buses eclipsing each other (like the masks in the theater icon)?

That's what I was thinking. Showing more than one bus may be an intuitive solution. 
The hard part is how to show more than one bus... 
- One bus in front, eclipsing another (smaller) one, so that about half or a third of it can be seen?
- Or 2 buses that stand next to each other, so that only half of each can be seen?

A roof could be OK too, but then I'd take the same roof as for indoor parking.
For all of these, I guess the bus(ses) need to be smaller versions of the bus-stop icon.

PR #1489, opened 2015-04-17
Remove trailing whitespace

gravitystorm:
I noticed this has crept in while I was working on other things.


pnorman:
project.mml is out of sync

gravitystorm:
Fixed (hopefully), and I've dealt with the mss files too.

Issue #1460, opened 2015-03-23
Icons needed for extremely popular shops [meta-issue]


kocio-pl:
We should start rendering extremely popular shop types (let's say above 10 000 uses [on taginfo](http://taginfo.openstreetmap.org/keys/shop#values)), but we probably have no icons for them:

* [x] 1. electronics (19 671) -  see: https://github.com/gravitystorm/openstreetmap-carto/pull/1493
* [x] 2. hardware (17 504) - see: https://github.com/gravitystorm/openstreetmap-carto/pull/1522
* [x] 3. gift (14 457) - see: https://github.com/gravitystorm/openstreetmap-carto/pull/1493
* [x] 4. beauty (14 258) - see: https://github.com/gravitystorm/openstreetmap-carto/issues/1525
* [x] 5. chemist (13 281) - see: https://github.com/gravitystorm/openstreetmap-carto/issues/1528
* [x] 6. garden_centre (11 509) - see: https://github.com/gravitystorm/openstreetmap-carto/issues/1523
* [x] 7. sports (10 554) - see: https://github.com/gravitystorm/openstreetmap-carto/issues/1530
* [x] 8. variety_store (10 249) - see: https://github.com/gravitystorm/openstreetmap-carto/issues/1536
* [x] 9. confectionery (10 091) - see: https://github.com/gravitystorm/openstreetmap-carto/issues/1534
* [x] 10. car_parts (10 003) - see https://github.com/gravitystorm/openstreetmap-carto/issues/1537
* [x] 11. travel_agency (10 120) - see: https://github.com/gravitystorm/openstreetmap-carto/issues/1409
* [x] 12. laundry + dry_cleaning (9 893 + 7 786) - see: https://github.com/gravitystorm/openstreetmap-carto/issues/1411
* [x] 13. newsagent + kiosk (8 913 + 46 299) - see: https://github.com/gravitystorm/openstreetmap-carto/issues/1459
* [x] 14. stationery - see: https://github.com/gravitystorm/openstreetmap-carto/pull/1588
* [x] 15. beverages  - see: https://github.com/gravitystorm/openstreetmap-carto/issues/1592
* [x] 16. toys - see: https://github.com/gravitystorm/openstreetmap-carto/issues/1414

Getting there (items above 5 000 uses):
* [x] amenity/shop=ice_cream - see: https://github.com/gravitystorm/openstreetmap-carto/issues/1479
[...]

Issue #1412, opened 2015-03-18
Add special rendering for shop=mobile_phone


kocio-pl:
We should start rendering shop=mobile_phone, probably with this icon from osmfr-cartocss fork:

https://github.com/cquest/osmfr-cartocss/blob/master/symbols/mobile_phone.svg

[It was proposed in meta-issue #1402]

nebulon42:
[mobile-phone-18.svg](https://github.com/nebulon42/osmic/blob/master/shop/mobile-phone-18.svg): ![mobile-phone-18](https://cloud.githubusercontent.com/assets/3531092/6778209/03c1ddb8-d151-11e4-87f2-6ca13513081c.png)

kocio-pl:
New model with the bigger screen - great! ;-P

HolgerJeromin:
phones does not have an antenna anymore, but it is useful. Its no iPod :-)

nebulon42:
Apparently closing with the commit in #1493 did not work. Should be closed.

Issue #1408, opened 2015-03-18
Add special rendering for shop=optician


kocio-pl:
We should start rendering shop=optician, probably with this icon from osmfr-cartocss fork:

https://github.com/cquest/osmfr-cartocss/blob/master/symbols/optician.svg

[It was proposed in meta-issue #1402]

matkoniecz:
shop=optician is already rendered, with generic shop icon (a purple dot). Is it intended as "render shop=optician with a special icon?" For list of rendered tag you may consult https://github.com/matkoniecz/CartoCSSHelper/blob/master/style_specific/default_osm_style.rb or search in code - https://github.com/gravitystorm/openstreetmap-carto/search?utf8=%E2%9C%93&q=optician

matkoniecz:
The same applies to other clones of this ticket.

kocio-pl:
I thought that is clearly written in the body of the issue - but of course I can also make title more clear too.

nebulon42:
[optician-18.svg](https://github.com/nebulon42/osmic/blob/master/shop/optician-18.svg): ![optician-18](https://cloud.githubusercontent.com/assets/3531092/6778233/2822f156-d151-11e4-90b4-2c8e3dc5bb88.png)

kocio-pl:
Good and recognizable.

nebulon42:
Apparently closing with the commit in #1493 did not work. Should be closed.

Issue #1407, opened 2015-03-18
Add special rendering for shop=alcohol and shop=beverages


kocio-pl:
We should start rendering shop=alcohol and shop=beverages, probably with this icon from osmfr-cartocss fork:

https://github.com/cquest/osmfr-cartocss/blob/master/symbols/alcohol.svg

[It was proposed in meta-issue #1402]

nebulon42:
[alcohol-18.svg](https://github.com/nebulon42/osmic/blob/master/shop/alcohol-18.svg): ![alcohol-18](https://cloud.githubusercontent.com/assets/3531092/6778243/4d516066-d151-11e4-938a-6019933d7c75.png)

kocio-pl:
Both the bottle and the glass could be taller, I think, but otherwise it's good icon.

nebulon42:
What is the distinction to beverages? Should both use the same icon?

kocio-pl:
It looks like beverages can be both alcoholic and non-alcoholic: 
- [OSM Wiki](http://wiki.openstreetmap.org/wiki/Tag:shop%3Dbeverages)
- [Wikipedia](https://en.wikipedia.org/wiki/Drink)

so probably we need also a generic icon for all the beverages.

math1985:
A percentage symbol is often used to indicate alcohol.

kocio-pl:
Maybe beverages shop should have a bottle+cup pair instead of bottle+glass?

dieterdreist:
2015-04-21 0:15 GMT+02:00 kocio-pl <notifications@github.com>:

> Maybe beverages shop should have a bottle+cup pair instead of bottle+glass?



I would think about a cafe and not about soft drinks when seeing a cup. All
beverages shops I have ever been to were selling soft drinks, juices,
water, beer, wine, maybe spirits but not tea or coffee and not even milk.

kocio-pl:
I meant glass [like this](https://commons.wikimedia.org/wiki/File:Glass_empty.jpg) (wider-on-top and thinner-on-bottom shape would be easier to recognize than rectangle), but was unable to tell it exactly in English (in Polish we have "szklanka" for beverages and "kieliszek" for an alcohol). =}

kocio-pl:
[...]

PR #1493, opened 2015-04-20
add icons for popular shop types

### LINKED ISSUE #1415: Add special rendering for shop=jewelry
We should start rendering shop=jewelry, probably with this icon from osmfr-cartocss fork:

https://github.com/cquest/osmfr-cartocss/blob/master/symbols/jewelry.svg

[It was proposed in meta-issue #1402]

nebulon42:
Resolves #1415, #1412, #1408 and part of #1460.

alcohol ~~(distinction to beverages unclear #1407)~~
alcohol

electronics
electronics

furniture ~~(distinction to interior_design unclear #1406)~~ (interior_decoration not considered for now)
furniture

gift
gift

jewelry, jewellery
jewelry

mobile_phone
mobile_phone

optician
optician


math1985:
For Jewelry, I also like an icon like this:

http://wiki.openstreetmap.org/wiki/File:Shop-jewelry.svg

Simpler than the ring, and the simpler the better on this scale, I guess.

nebulon42:
I had a try with the diamond, but the inverse version of what you linked here did not work out. I'm not sure regarding the linked version, it's rather noisy for me. I seriously considered the diamond, but then moved towards the ring in the end.

math1985:
Ok, good that you had a look at it at least!

matkoniecz:
> alcohol (distinction to beverages unclear #1407)

Wiki has "shop=alcohol Shop selling alcoholic drinks" and "shop=beverages Shop focused on selling alcoholic and non-alcoholic beverages.".

matkoniecz:
Maybe shop=beverages may have something else rather than wine glass? Second bottle? Can?

dieterdreist:
> Am 21.04.2015 um 07:10 schrieb Mateusz Konieczny <notifications@github.com>:
> 
> alcohol (distinction to beverages unclear #1407)
> 
> Wiki has "shop=alcohol Shop selling alcoholic drinks" and "shop=beverages Shop focused on selling alcoholic and non-alcoholic beverages.".
> 


it's mainly a cultural distinction, in countries with restricted sale of alcohol (e.g. USA, Iran) there are special alcohol shops, while in other countries you will get alcohol also in "normal" beverage shops, but also in these countries there will typically be shops that sell alcohol only, often just of some particular  kind (typically high quality), like wine or whiskey.

nebulon42:
Ok, so beverages should be distinct. I will work on that later. Any further comments on these icons?

imagico:
optician, mobile_phone and gift seem fine to me.

furniture i would make the 'line' less prominent - right now it appears to be a two part object

jewelry seems too abstract to be easily recognizable - could also resemble some kind of doll/puppet or some technical stuff.

electronics seems ambiguous - could also mean a kind of public viewing location.

dieterdreist:
2015-04-20 23:34 GMT+02:00 nebulon42 <notifications@github.com>:

> alcohol (distinction to beverages unclear)
> [image: alcohol]
> <https://cloud.githubusercontent.com/assets/3531092/7240957/8fa5b1f4-e7b5-11e4-8e7d-9788b4f90a45.png>
>

There is also the tag shop=wine (used 1933 times currently), that might be
more suitable for those places specialized in wine, and your icon seems to
fit perfectly for wine.
Used not only in Europe: http://taginfo.osm.org/tags/shop=wine#map

math1985:
The optician icon is 3d, while the other icons look more 'flat'. I think that was also a design principle. I don't oppose the proposed icon, but it does stand out a bit. I wonder if there would be a way to create a icon without 3d-effect? If not, I'm find with going fo the proposed icon.

matkoniecz:
> There is also the tag shop=wine (used 1933 times currently), that might be
more suitable for those places specialized in wine, and your icon seems to
fit perfectly for wine.

I think that different icons for shop=alcohol and shop=wine would be an overkill, the same may be used for both.

nebulon42:
I also would use the same icon for alcohol and wine shop. But that should be done afterwards if desirable.

Optician is 3D: true, but glasses without sidepieces look not that good e.g. optician-16 vs. optician
I'm also not satisfied with fake perspective here: optician-16

furniture: I'm not sure on the gap width but slightly more inclined towards the larger one. furniture 16 vs. furniture-16 (0.25px less wide)

For jewelry I have nothing else in store as the diamond did not work.

I think that TV might equal public viewing is a bit far fetched, but electronics shop is itself ambiguous so it's quite hard to settle onto a specific depiction.

math1985:
Thanks!

Issue #1158, opened 2014-12-11
Wetlands are too prominent


Zverik:
![wetlands](https://cloud.githubusercontent.com/assets/766031/5392349/1eef519e-8136-11e4-8144-a3505004e6ce.png)

I'd suggest making lines less than a pixel wide, 1.5 times longer and with opacity around 0.6-0.7.

math1985:
Fixed by #1497.

Issue #545, opened 2014-05-20
Add rendering for landuse=salt_pond


matthijsmelissen:
The following landcover types are currently not rendered:
- ~~landuse=harbour, see also https://trac.openstreetmap.org/ticket/3612~~ rejected in https://github.com/gravitystorm/openstreetmap-carto/issues/545#issuecomment-277207366 and https://github.com/gravitystorm/openstreetmap-carto/issues/545#issuecomment-341894371
- landuse=salt_pond, see also https://trac.openstreetmap.org/ticket/4470
- ~~natural=dune~~, see also https://trac.openstreetmap.org/ticket/3815
- ~~natural=fell~~ rejected in https://github.com/gravitystorm/openstreetmap-carto/issues/545#issuecomment-98479426 and https://github.com/gravitystorm/openstreetmap-carto/issues/545#issuecomment-341895192 , see also https://trac.openstreetmap.org/ticket/3033
- ~~and natural=scree~~ (done), see also https://trac.openstreetmap.org/ticket/3033
- ~~natural=reef, see also https://trac.openstreetmap.org/ticket/4416~~ done

It would be nice to have these tags rendered.

mkoniecz:
natural=dune conflicts with more widely used natural=sand, is used on areas less than 800 times, never went through any kind of approval process. I would suggest dropping it from this list.

math1985:
> natural=dune conflicts with more widely used natural=sand

And with natural=scrub, depending on the type of dunes. See [https://www.google.co.uk/search?q=dunes+netherlands&tbm=isch](here).

polarbearing:
The problem is that natural formations are mixed with the substances in the tags. Beach can be sandy or stony, dune can be sand or nicely grown over with dune grass.

dieterdreist:
2015-01-11 1:01 GMT+01:00 polarbearing <notifications@github.com>:

> The problem is that natural formations are mixed with the substances in
> the tags. Beach can be sandy or stony, dune can be sand or nicely grown
> over with dune grass.




+1, this is clearly a tagging problem. natural=sand/mud/bare_rock are "bad"
because they are describing properties of an object rather than a class of
object, thereby creating tagging conflicts/incompatibilities. They should
go under another key where they would fit, e.g. "surface" or "landcover".
Unfortunately these tags are rendered currently and there are presets for
them, so that mappers are encouraged to continue using them.

fkv1:
[...]

Issue #115, opened 2013-08-14
natural=mud doesn't render in rivers


chillly:
natural=mud now renders around the coastline, but not in rivers. The example below shows the join between coast and river and the cutoff is obvious.

Example: http://www.openstreetmap.org/#map=16/53.6995/-0.4381

malcolmh:
If the waterway=riverbank tag is changed to natural=water, then the rendering is correct. These two tags ought to produce the same result, but don't.

imagico:
I see no indication of areas tagged natural=water and waterway=riverbank being treated any differently - if they are please open a new issue and show an example.  Note any special issues with natural=mud will be solved by #1497.

This problem here (different handling of coastline and water areas in relation to landcovers) is still unsolved - it will require use of water polygons for the ocean rendering.

malcolmh:
Look here: http://www.openstreetmap.org/#map=14/53.7058/-0.5772. I changed one of the waterway=riverbank polygons to natural=water and the mud re-appeared.

imagico:
#1497 is not yet rolled out - if there was any difference before it should be gone after this change.

By the way - if you change waterway=riverbank to natural=water please always add water=river.

malcolmh:
OK, I will wait until roll-out to see if this issue is then cleared. Thanks!

math1985:
Closing this now, please let us know if there is still a problem after #1497 is rolled out.

PR #1497, opened 2015-04-22
wetland rendering differentiated by wetland type

### LINKED ISSUE #387: Rendering tidal flats
It appears that all areas tagged "natural=wetland" are rendered with the reed-bed pattern, regardless of the "wetland=xxx" or "surface=xxx" tags. Since the reed-bed pattern implies vegetation, this rendering is inappropriate for areas without  vegetation. In particular, areas that have "wetland=tidalflat" together with "surface=mud" or "surface=sand" should be rendered with the same pattern as the existing rendering of "natural=mud" and "natural=sand"

imagico:
resolves #387, #1158

This adds specialized patterns for various types of wetland as indicated by the wetland tag.  It also unifies natural=mud with wetland in terms of style - it is rendered like wetland=tidalflat.

As before the wetland pattern is rendered as an overlay.  The base color - if it exists - is rendered in the landcover layer.  There wetland has priority over landuse so if a wetland polygon has also a landuse tag (which is quite common) the wetland type supersedes the landuse for the base color.  Rendering with the basic wetland pattern starts at z=10 as before.  The specialized patterns are used at z=14 and above.  Also the wetland pattern is less strong than before (fixing #1158)

Here are generic styling examples for all different pattern types as used at high zooms:

bog and swamp:

bogswamp

bog uses the base color from heath, swamp from wood.

saltmarsh and reedbed:

saltmarshreedbed

marsh and wet_meadow are rendered the same as saltmarsh, just with the grass color as base color.

tidalflat/mud and mangrove:

tidalflatmangrove

As can be seen the last two use a semitransparent base color.  This is to make sure the coastline is always clearly visible even if a wetland polygon is drawn across it and transparency is the only way to do this.  The colors are designed to match the previous mud color and wood color when used over land.  If there is fear this might cause confusing colors when overlapping with other landcovers i can move these to a separate layer rendered before the landcover layer.  This would ensure only the coastline layer is below, it would however also complicate the code quite a bit.

Practical rendering examples:

z=13:
z=13

z=14:
z=14

Showing how the overlay renders above water areas:
bog

I did not touch the labels - they are still rendered in strong blue for all types of wetland.  This seems fine to me but can be changed of course.

Mangrove rendering showing which areas are inside and outside the coastline:
mangrove


pnorman:
impacts #1242

dieterdreist:
2015-04-23 0:34 GMT+02:00 imagico <notifications@github.com>:

> resolves #387
> <https://github.com/gravitystorm/openstreetmap-carto/issues/387>, #1158
> <https://github.com/gravitystorm/openstreetmap-carto/issues/1158>
>
> This adds specialized patterns for various types of wetland as indicated
> by the wetland tag.
>

very nice!

imagico:
Made some small changes:  For consistency and to avoid irritating color changes when zooming the base color of swamp/mangrove is rendered from z=8 like wood/forest.  Tidalflat is shown from z=9 which is consistent with the other bare ground tags and seems appropriate since tidal flats are frequently larger areas.

nebulon42:
It would be great to include also the source files of the patterns - if feasible e.g. because they are SVG - to allow for easier customisation.

imagico:
The patterns as a whole do not exist in SVG form, they are assembled using raster processing (i.e. masking the base dash pattern around the symbols).  I will however include the symbols used with jsdotpattern and can also make available the point arrangement - here with the mangrove symbol:

mangrove pattern svg

imagico:
Mud labels now also match the wetland labels:

wetland labels

nebulon42:
>  I will however include the symbols used with jsdotpattern and can also make available the point arrangement

Yes that's what I meant. Probably the form of including this in the style has to be discussed, but I always thought that it would be important to offer forks the possibility to change the patterns easily without having to recreate them from scratch.

gravitystorm:
We'll need the patterns in SVG too, eventually. Mapnik 2.x can only use raster patterns in patternsymbolizers, but this limitation is removed in mapnik 3.x so we'll prefer SVG patternsymbolizers at that point.

imagico:
I agree having SVGs for patterns could be useful, especially for rendering the style at different resolutions - however the design of the patterns here currently relies on raster processing and some parts of this (in particular use of hit-and-miss pattern matching - see [http://www.imagemagick.org/Usage/morphology/#hmt]) cannot easily be translated into vector operations.  This means creating matching SVGs will be complicated and this only makes sense to do when a pattern is finalized.

@pnorman also mentioned in imagico/jsdotpattern#3 that mapnik 3 has limitations in SVG support that would significantly blow up SVG file sizes for patterns with lots of or complex symbols - the current bare ground patterns for example would each be several megabytes in size as SVGs in mapnik compatible form.

Long story short - i think the concept of SVG patterns has still quite a lot of open questions.  Re. editability as @nebulon42 hinted the question is for example if and to what extent it makes sense to manually edit pattern SVGs in programs like Inkscape where maintaining the periodicity is not ensured.

Another thing - does anyone know if mapnik 3 pre-rasterizes SVG patterns when they are used?  If not i'd estimate using them will have a significant negative performance impact.

imagico:
rebased - from my side this is ready now.

landcover.mss
```diff
@@ -442,6 +444,45 @@
     }
   }
 
+  [feature = 'wetland_mud'],
```
math1985: Should this not be natural_mud?

water.mss
```diff
@@ -52,19 +51,37 @@
...
-      polygon-pattern-file: url('symbols/wetland.png');
+  [zoom >= 10] {
+    polygon-pattern-file: url('symbols/wetland.png');
+    polygon-pattern-alignment: global;
+  }
+  [zoom >= 14] {
+    [int_wetland = 'marsh'],
+    [int_wetland = 'saltmarsh'],
+    [int_wetland = 'wet_meadow'],
+    [int_wetland = 'fen'] {
```
math1985: I would drop wetland=fen. It is undocumented and has only 65 instances.

landcover.mss
imagico: No, in https://github.com/imagico/openstreetmap-carto/blob/wetland-improvements/project.yaml#L151 mud and marsh are translated to wetland types so natural=marsh/mud and natural=wetland+wetland=marsh/mud are treated exactly the same.  This could of course also be done in the style sheet but this seems more consistent in result, esp. with regard to priorities in COALESCE() - otherwise a polygon tagged natural=marsh + landuse=* would be rendered different to one tagged natural=wetland + wetland=marsh + landuse=*.

water.mss
imagico: Would be fine with me - i included it since it is on http://wiki.openstreetmap.org/wiki/Key:wetland.

landcover.mss
math1985: Thanks, I see now.

math1985: Ok, I see.

math1985:
Looks nice!

The only thing I'm not happy about it is the transparency. I think indeed that it would give undesired colours when overlapping with other landuse. 

I also don't think that the transparency over land/water is intuitive. In pictures like this:

https://camo.githubusercontent.com/c2591e894ceefa74086ba35d30069b7cb18a9191/687474703a2f2f7777772e696d616769636f2e64652f66696c65732f7765746c616e645f72656e646572342e706e67

it is not clear to me that the darker parts are actually outside the coastline.

math1985:
Looks nice!

The only thing I'm not happy about it is the transparency. I think indeed that it would give undesired colours when overlapping with other landuse. 

I also don't think that the transparency over land/water is intuitive. In pictures like this:

Mangrove

it is not intuitive to me that the darker parts are actually outside the coastline.

I'm also not sure if the coastline is a relevant physical feature on high zoom levels, so I wouldn't mind if it weren't visible.

matkoniecz:
> it is not intuitive to me that the darker parts are actually outside the coastline.

Yes, I expected that darker part is a land.

imagico:
Ok - first the sample image is not actually very good since it is relatively bad mapping there and the boundary also adds to the confusion.  But i see that the different color over water is not very clear in case of mangrove which probably has partly to do with the fact that transparency is fairly low and to get the same color as wood over land the color itself has to be very dark.

> I'm also not sure if the coastline is a relevant physical feature on high zoom levels, so I wouldn't mind if it weren't visible.

The point is that by hiding it mappers are discouraged from proper placement - it does not seem to matter since it has no effect on the rendering result.  We can already widely see this in case of natural=coastline vs. natural=water.  So i would rather render mangrove and tidalflat without a base color than with one that obscures the coastline.

Note this is not a high zoom level problem only - both these types of wetland cover quite large areas in some cases and an arbitrary placement of the coastline within such wetlands will have a very bad effect on the coastline quality.

I could imagine three measures:

* make the mangrove color more transparent so the distinction over water/over land is clearer.  In any case the part over water will be darker than over land - unless i use a different comp-op - which however would probably be even less intuitive.
* create an additional wetland base color layer to avoid interference with other landcovers.  Note the risk of problems because of this is fairly low since the wetlands in questions tend to be fairly large and the ordering is by size - you would need a different larger landcover polygon overlapping with a smaller mangrove/tidalflat polygon.  I am sure there are cases of this but it is definitely not a very common scenario.
* remove the base color from mangrove/tidalflat - tidalflat would then be rendered like generic wetland and mangrove like saltmarsh, just with a different symbol.

below is an example for 50/60 percent transparency for mangrove (current is 25):




and here using exclusion comp-op:

exclusion

imagico:
Found a somewhat better area for mangrove mapping crossing the coastline southeast of Mumbai:

mangrove example 1
mangrove example 2

Yes, the difference is fairly subtle - could be emphasized more as shown in the previous comment.  But i don't think there is normally confusion which part is inside and which is outside the coastline.

math1985:
Actually I still don't recognize it as coastline, rather as two different types of vegetation.

imagico:
Well - in a way it is - one is mangrove forest above the high water line, the other is mangrove forest below it.

Some level of obfuscation of the coastline seems appropriate considering mangrove does make it less recognizable in reality.  But having it still visible, at least in a subtle way tells the mapper: it still matters, take care about placing it sensibly.  Hiding it fully conveys the message: once you mapped the mangroves the coastline is not important any more, just close it somehow as it is convenient.

I tried the more transparent colors shown above but the part over water is too dark IMO compared to the rest of the style.

Using water polygons instead of land polygons would give some more options of course.

pnorman:
> Using water polygons instead of land polygons would give some more options of course.

This is on the to do list

imagico:
I removed the mangrove base color now since it is generally not found favorable.  I also moved the tidalflat/mud base color rendering to a separate symbolizer which i think makes sure it is rendered before the other landcovers without the need for a separate layer.

Sample:

mangrove without base color

sommerluk:
> I removed the mangrove base color now

I like the approach without base colour! Especially because your intention is to make clear that it matters where the coastline passes – and the version without base colour makes this clearer.

imagico:
It is not too bad IMO - but since mangrove would also usually be considered a type of woodland it would make sense to for it to use the base color of natural=wood - especially since at least in the inner tropics mangrove usually transits seamlessly into tropical rainforest under natural conditions.

Not using a base color might encourage mappers to map wood/forest in addition but not more than right now and practical mapping of mangrove areas right now is overwhelmingly using natural=wetland without another solid color area tag in addition.

When we move to water polygons it would be easily possible to use the wood base color over land and a mixture of wood and water color over water - this would be the best solution i think.

math1985:
> I also moved the tidalflat/mud base color rendering to a separate symbolizer which i think makes sure it is rendered before the other landcovers without the need for a separate layer.

Yes, that works.

However, what about other landcover inside tidalflat/mud? I can think of leisure=track inside landuse=mud, for example. Your approach would not render the track. Currently we order landuse by area size, and this would be inconsistent with that approach.

imagico:
Other landcovers within tidalflat/mud work fine, here natural=sand within natural=mud (somewhat suboptimal tagging of course but that's not the issue here):

sand within mud

Of course a small tidalflat/mud area within a larger other landcover would probably only render with the wetland overlay.

math1985:
You're right, I read

> I also moved the tidalflat/mud base color rendering to a separate symbolizer which i think makes sure it is rendered before the other landcovers without the need for a separate layer.

the wrong way around.

math1985:
Thanks, looks very good!

imagico:
Ok, thanks for reviewing and merging.

nebulon42:
I'd like to mention again that it would be nice having also source files/symbols included in the style somehow. Especially since most symbols are now going to be SVG it would be great if also patterns could be easily customised. I can open an issue for that though, if that is better suited.

math1985:
> I can open an issue for that though, if that is better suited.

Yes, please. I agree, but thought it was not worth postponing the merge for this reason.

nebulon42:
No, of course not.

malcolmh:
Could you please summarise the tagging used for the above illustrations?

imagico:
Places shown are:

http://www.openstreetmap.org/#map=14/53.7692/7.7091
http://www.openstreetmap.org/#map=16/51.5730/12.7657
http://www.openstreetmap.org/#map=15/22.6919/-84.0193
http://www.openstreetmap.org/#map=13/53.3186/-2.8796
http://www.openstreetmap.org/#map=13/18.8347/72.9652

general wetland tagging is explained on:

http://wiki.openstreetmap.org/wiki/Tag:natural%3Dwetland
http://wiki.openstreetmap.org/wiki/Key:wetland

malcolmh:
Interesting. Looking at those examples, I see that natural=water is used in preference to waterway=riverbank to produce the correct rendering. There is a long-unfixed issue (#115) regarding this.

malcolmh:
Tidal flats can expose various surfaces: mud, sand, shingle, etc. Will these cases be rendered with different background colours?

imagico:
Currently there is no established tagging for that - wetland=tidalflat is not used significantly in combination with surface=*.  Larger flat tidal areas (which is the meaning of tidalflat) also generally feature very small grained material at the surface - hence are usually muddy.

Sand/gravel features in tidal areas would often most accurately be tagged natural=shoal - #545 mentions the possibility of rendering this.  Another possible improvement is rendering the tidal part of natural=beach different from that above the high water line.  This once again would probably require use of water polygons.

malcolmh:
Shoal is not the same as tidalflat. A shoal is an area of shallow water, whether tidal or not, the seabed of which is not necessarily exposed at low water.

I could find no explanation of "water polygons" in the Wiki. What is meant by this term in this context?

imagico:
This is not the place for discussing tagging - if there are a significant number of areas you consider rightfully tagged natural=wetland + wetland=tidalflat + surface!=mud it might be worth looking into rendering them in a distinct form - i am not sure such areas exist though.  Best to discuss this first in local community or on tagging ML.

water polygons refers to use of

http://openstreetmapdata.com/data/water-polygons

rather than (as right now)

http://openstreetmapdata.com/data/land-polygons

for rendering the ocean/land distinction.  This has been in discussion occasionally for a long time.

PR #1498, opened 2015-04-23
change symbols of shop=supermarket and shop=convenience

nebulon42:
There is nothing wrong with the existing SVGs for `shop=supermarket` and `shop=convenience`, but the depiction of the grid makes them a bit noisy. I therefore propose to replace them with solid fill symbols. Especially supermarket becomes more visible.

supermarket - before
supermarket_before

after
supermarket_after

convenience - before
convenience_before

after
convenience_after


pnorman:
I prefer the old SVGs, particularly for convenience

kocio-pl:
I don't see the problem with current icons, plus I find the new ones too noticable (just like current solid icon for libraries).

math1985:
I also prefer the old icons.

dieterdreist:
2015-04-24 11:14 GMT+02:00 kocio-pl <notifications@github.com>:

> I don't see the problem with current icons, plus I find the new ones too
> noticable (just like current solid icon for libraries).



+1, visual impact of solid ones to too heavy and the lighter version seems
also more recognizable.

ximex:
i also prefer the old icons

dieterdreist:
Actually there is one part of the new supermarket icon which I prefer over
the old one: the wheels. Maybe these could be integrated into the old icon?

pnorman:
> Actually there is one part of the new supermarket icon which I prefer over
> the old one: the wheels. Maybe these could be integrated into the old icon?

There are non-detached wheels in the old icon which could be made stronger.

PR #1499, opened 2015-04-25
Make clothes shop icon more distinct

nebulon42:
...by adding a neckline.

before
clothes_before

after
clothes_after


kocio-pl:
Very good improvement!

PR #1501, opened 2015-04-27
Make WHERE statement in landcover more consistent

math1985:


pnorman:
I suspect this would not result in any performance improvements, and potentially a degradation.

PR #1512, opened 2015-05-01
Move link trimming to outer select

### LINKED ISSUE #1456: *_link highways now rendering above 'normal' highways
https://www.openstreetmap.org/#map=16/40.4500/-78.8399
https://www.openstreetmap.org/#map=16/40.4898/-78.7496
https://www.openstreetmap.org/#map=17/40.44787/-78.43308

This looks pretty ugly IMO.  Seems something got changed in the last update to the map style that shouldn't have been.  Know this is happening with 3 of the 5 *_link type highways.  Still looking for tertiary & secondary bugged ones.

Hope this can be an easy fix to get it back to where the *_link renders under the main highways again.

gravitystorm:
Fixes #1456

By delaying the trim to the outer select, it means that the raw
highway value can be used when joining to the priority table. Without
this, *_links are treated as the same priority as the base road type,
and that leads to rendering artefacts


math1985:
Thanks, code and output looks fine to me.

pnorman:
we might want to consider tagging a release now

PR #1517, opened 2015-05-02
move bus stop icon to SVG

nebulon42:
Implements part of #1165.
Size now remains unchanged after https://github.com/gravitystorm/openstreetmap-carto/pull/1478#issuecomment-91868336.

before
bus_stop_before

after
bus_stop_after


mboeringa:
Looks nice. I was fine with the bigger version too, but this is OK. The bus now also looks a bit more modern compared to the old "before" one, which more looks like a '50s bus.

matkoniecz:
It is amazing how quickly and how much I started hating current bus icon. A great improvement, thanks!

nebulon42:
Thanks too! :)

daganzdaanda:
Thanks, also from me. The old bus looks ancient, maybe the icon was made in the sixties ;-)

kocio-pl:
And the old one was sooo blurry. =}

PR #1518, opened 2015-05-03
move archaeological site to SVG

nebulon42:
Implements part of #1165.

before
archaeological_site_before

after
archaeological_site_after


mboeringa:
Way better! The old icon looked like some fire pit for a camp site...

kocio-pl:
The old one was quite good, but this one is more crisp. I like previous icon's idea to "stick out" of the ground, but this is not needed IMO.

mboeringa:
> I like previous icon's idea to "stick out" of the ground, but this is not needed IMO.

I always have had a hard time recognizing it as such, as the tilted "amphora" isn't recognizable enough in the original. My first impression was always something different when seeing the icon...

daganzdaanda:
> I like previous icon's idea to "stick out" of the ground

So THAT is what that was supposed to be! I never understood that icon! The new one is a million times better for me.

PR #1452, opened 2015-03-22
Add labels to various objects

math1985:
Add labels for:
* amenity=post_office (resolves #840)
* natural=tree (resolves #801)
* amenity=car_sharing (operator, resolves #800)
* amenity=atm (operator)
* amenity=bicycle_rental
* man_made=mast
* amenity=recycling
* amenity=drinking_water
* man_made=water_tower
* tourism=picnic_site
* leisure=picnic_table
* leisure=slipway

Resolves #1091.

PR #1519, opened 2015-05-03
move water tower icon to SVG

nebulon42:
Implements part of #1165.
Symbol is now smaller.

before
water_tower_before

after
water_tower_after

~~It's odd that the new symbol is not centred within the way. If someone knows the reason for that, I'm happy to hear it. The shape within the SVG is not misaligned.~~


mboeringa:
@nebulon42 , what about this variation?

water_tower

The original proposal slightly starts to look like a figure...

kocio-pl:
@mboeringa +1

math1985:
> It's odd that the new symbol is not centred within the way.

Could it be that the example is tagged both as way and as node?

nebulon42:
@math1985 No, I have checked. (http://www.openstreetmap.org/way/27501807)

matkoniecz:
@mboeringa The same happens for me. With new icon it is not immediately obvious what is the meaning. Current icon and one proposed by @mboeringa is immediately clear (at least for me).

mboeringa:
The **name** of the tower is never displayed at any scale as far as I can see, is that normal as well?...

nebulon42:
@mboeringa Is the little blank space beneath the fill intentional? Looks a bit strange. re name: #1452 deals with that.

nebulon42:
This would be an alternative for me: water-tower-142

mboeringa:
> Is the little blank space beneath the fill intentional?

I actually simply extended the fill to the left and right to close the gaps there. That is the only change compared to the original you proposed, which also has some blank space beneath the water body.

Here is an adjusted version next to the first one, I'll leave it to others to judge whether one is better than the other, I am kind of neutral here:

water_tower
water_tower_adjusted

mboeringa:
> This would be an alternative for me:

That slightly starts to look like these modern fancy "grills" you nowadays often see outside restaurants... The top part looks a bit like a lit with gap. Can you try it without the gap?

nebulon42:
That sounds quite far fetched for me.

nebulon42:
Smaller gap for the fill: water-tower-143

Different water fill level: water-tower-144 water-tower-145 water-tower-146

I'm still in favour of the fill gap.

mboeringa:
> That sounds quite far fetched for me.
;-)

grill_looking_icon
grill

nebulon42:
If you wanted to get on my nerves you just succeeded.

imagico:
My order of preference would be:

1. the solid high water fill
2. the one with lid
3. the small gap one
4. the original proposal

matkoniecz:
> That sounds quite far fetched for me. Without the gap it is too solid. Smaller gap may work though: water-tower-142a

This is my favorite.

mboeringa:
> **If you wanted to get on my nerves** you just succeeded.

No, that most certainly was not my intention... I do appreciate your work here on these icons. I'll leave it to the others to give their final judgements on the different proposals (as they are already doing).

kocio-pl:
I prefer solid fill 1, then solid fill 2, then solid fill 3 (because they resemble something filled with water - the more water, the better for me =} ), than the one with lid (but see the remark below), then smaller gap (because it doesn't look like human anymore, but still no resembling of the water).

I'd like to save the lid version to show a gas holder/gasometer - [man_made=gasometer has already 12 671 uses](http://taginfo.openstreetmap.org/tags/?key=man_made&value=gasometer), so it clearly deserves its own icon! The nice thing would be it would be similar to water tank, so the map users can make a better guess using just the shape.

Original proposition - big gap - is unclear for some of us. Current icon is clear, but making OSM icons more uniform is a noble goal and I second that.

nebulon42:
@mboeringa I would love to hear your order of preference too, if any. Thanks.

mboeringa:
I think I am with @imagico with this one:

1) the solid high water fill
2) the one with lid (your adjusted one with smaller gap though)
3) the small gap one
4) the original proposal

I also like @kocio-pl's proposal of possibly using 2) for gasometers, although gasometers generally are not real tower constructions (but they can be high and have a mild "tower" appearance with decorative brick work and all, like defensive towers of fortresses and castles, especially if dating from the 19th or beginning 20th century).

kocio-pl:
I try to think what other icons we may need in the not-so-distant future, and [looking at man_made key on Taginfo](http://taginfo.openstreetmap.org/keys/man_made#values), we see a dozen of popular objects (let's say >10k uses, as with [extremely popular shop types](https://github.com/gravitystorm/openstreetmap-carto/issues/1460)):

1) survey_point - 288 463
2) pier - 190 648
3) tower - 160 316
4) storage_tank - 114 230
5) cutline - 112 918
6) works - 75 510
7) pipeline - 61 557
8) water_tower - 44 226
9) water_well - 36 985
10) silo - 35 356
11) wastewater_plant - 33 770
12) surveillance - 31 576
13) petroleum_well - 30 015
14) chimney - 20 389
15) groyne - 17 633
16) reservoir_covered - 15 132
17) embankment - 14 719
18) mast - 14 036
19) gasometer - 12 671
20) breakwater - 12 256
21) flagpole - 11 556
22) water_works - 10 676

Some will probably not be visible on default map (like 1) survey_point) or just marked by a dot on a z=max, some others already have good visual representation (like 17) embankment) or have different shape (like 5) cutline), some other I'm not aware what do they look like or have to check what are they, but we're still left with a few somehow similar objects like storage_tank, silo, gasometer and water_tower. 

It would be great to take them all into consideration and design as a set of icons just to avoid problems later. I guess we already had this kind of conversation over redesigned symbol for tower/mast/flagpole (but I'm not able to find it now ={ ).

nebulon42:
I'm taking more of a bottom-up approach instead of the top-down approach you are suggesting here. I can only deal with that when we are there and cannot take all sorts of considerations into account.

kocio-pl:
Sure! These 3 more similar things I wrote about regarding to water_tower (storage_tank, silo and gasometer) are just a hint, because I'm in no way an expert in industrial objects.

mboeringa:
I don't think that all the objects that @kocio-pl listed need icons, in fact, things like embankments, groynes and breakwaters, should probably only have a polygon and line representation... There is really not much benefit, if any at all, for features like these ones having icons.

kocio-pl:
That's exactly what I said =} (please read my comment just below the list). What I put here is just a list of popular man_made objects, not all of them need an icon (or any representation at all!) on general map, we can just pick similarities to reuse icons whenever possible or avoid visual identity issues.

nebulon42:
Ok, I have listened and updated the symbol accordingly. In the meantime #1452 got merged, so I adjusted text-dy as well.

matkoniecz:
I dislike currently suggested icon ( https://github.com/gravitystorm/openstreetmap-carto/pull/1519#issue-72841985 ). One that is now displayed looks better at least in my opinion. New one is less clear.

daganzdaanda:
I disagree with @matkoniecz - I think the proposal is better than the current one. Though I'm not sure it needs the wave to symbolize water.

kocio-pl:
@daganzdaanda +1

nebulon42:
Rebased.

PR #1520, opened 2015-05-03
reuse symbol of florist for garden_centre

nebulon42:
Solves part of #1460.


kocio-pl:
I think watering can metaphor would be much better. Florist is a ready-made flowers shop, while garden_centre is like a DYI flowers/greens shop.

HolgerJeromin:
Yes, they are different. But I think a reuse would be an easy solution. Most of the time the name will help.

kocio-pl:
Drawing the water can is also quite easy, even with my very basic design skills =} : garden_centre-14 garden_centre-14-3

nebulon42:
While the water can is nice I'm not quite sure if it is really representative for a garden centre. If we want a distinct icon I was more thinking of a tree and a flower.

HolgerJeromin:
Nice icon!I misread your comment ("watering can" vs. the verb can) so i did not get your intention right. :-)

kocio-pl:
@HolgerJeromin English is not my native language, so I have to check at Wikipedia for some names and in cases like this I find Polish to be more precise ("konewka" has just one meaning). =}

@nebulon42 You're the design master here ;-) , so I welcome your visual projects, however in general I think you tend to underestimate the question of visual identity. For me the main objects you look for when going to florist are - well - flowers, but in the garden center you look for a garden tools. Sure, you can also buy flowers and trees, but in the essence they are also kind of "tools" for making/caring for/transforming your garden, so the tool/DYI is a key concept here and I'd like it to be easily shown.

In a nutshell: you can depict many things with a flower (florist shop, garden bed) or a tree (a tree itself, a forest...), but the water can (or maybe [a leaf rake](https://commons.wikimedia.org/wiki/File:A_leaf_rake.jpg)?) relates mainly to the gardening, so it's harder to misunderstand and that's what we want it to be for a mapping symbol.

nebulon42:
Maybe flower and tree could be ambiguous. The only flaw with your argumentation for the water can here is that you assume what works well for you will also work well for others. I do the same quite often. So you say water can works well for you. I say I'm not sure about it. We both are in a way right. :) The question is what works best for most. Leaf rake might be possible, but might have depiction issues.

But this can be discussed further. You convinced me about the need for a different icon, so I'm closing this. Discussion about which icon to choose could possibly happen in #1460.

Issue #720, opened 2014-07-10
Add rendering for elevators


math1985:
Elevators should be rendered as icon.

See also https://trac.openstreetmap.org/ticket/3545.

k1wiosm:
This would be nice. Any idea when will this be added?

math1985:
The problem is that some elevators are very important (the public elevators in Lisbon and Luxembourg come to mind), but others are totally irrelevant for a general purpose map (elevators inside small offices for example). We currently have no way to distinguish them.

@k1wiosm Did you have any particular elevator in mind?

k1wiosm:
I was thinking about a public elevator in this university (http://www.openstreetmap.org/#map=19/32.77629/-117.07330).

Definitely, only public elevators shoud be shown on the map. Are there really elevators inside small offices mapped on OSM?

mkoniecz:
@k1wiosm Certainly some people will start mapping anything that is rendered. Includinh indoor elevators in any high building.

pnorman:
I'm uncertain on this. Elevators would be more useful if we rendered elevation via hillshading and/or contour lines.

kocio-pl:
Elevators are important on viaducts, subway/train stations or underground footway passages and other transportation-related places, even without other ground elevation context. I would be very happy to see them there as I draw a lot of footways and it's not always clear how are they linked between the levels.

But I don't know how to easily recognize which one elevator is important public amenity and which one is private or useful only in the building context. For example train station elevator is of course a not-that-interesting building feature to get to the x floor, but it works mainly as the important link between footway and the platform in a journey planning (be it with heavy luggage, on wheelchair or with a bicycle), so it is useful in a broader context than the building itself. 

I also remember people with physical disabilities needs them also in places like townhalls, shopping centers, so I tend to think it would be good to know they are there.

RAytoun:
@kocio-pl You have raised an important strong case for showing elevators as a means for the physically disabled to plan their way through towns even if it includes shop and office elevators. By showing a visual difference to the rendered symbols it will highlight which elevators are a part of a route through the town and which are specific to a building.

kocio-pl:
Maybe we need to sort them out by the access=* tag (just like we do for public/private parkings). I think all the public places and some semi-public (like shopping centers, which are open to the public even if privately-owned) should have the elevators visible.

kocio-pl:
My proposition for elevator icon (10 px):
[...]

PR #1521, opened 2015-05-04
change symbol for railway=subway_entrance

### LINKED ISSUE #851: Change icon for railway=subway_entrance
Subway entrances are currently rendered with an icon that seems to symbolize 'walking'. This needs to be changed.

[Example](http://www.openstreetmap.org/#map=19/40.41912/-3.69332)

See also https://trac.openstreetmap.org/ticket/3920.

nebulon42:
Fixes #851, implements part of #1165.

before (click on images to see full preview)
subway_entrance_before

after
subway_entrance

issues:
~~subway entrances that are elevators are depicted with stairs (see http://www.openstreetmap.org/node/2703842678)~~


matkoniecz:
Neither old nor new symbol works for me. Old one is clearly somebody walking what is not helpful and not representing anything. New one is not looking like stairs (also at full sized image) - it is rectangle with line.

Maybe just stairs would work? Or arrow pointed toward bottom (I am not really convinced that this is a good idea)?

daganzdaanda:
Could you show the new symbol in train-blue? 
I think the idea works, if it is connected visually to other public transport features.
Or maybe it needs a filled rectangle to be more visible, or a tunnel-shape instead of a rectangle?

> Maybe just stairs would work? 

I've seen street signs where "stairs" stand for a pedestrian underpass. We probably don't need an icon for these, since we are rendering stairs and tunnels. So maybe it could work for subway entrance, too.

nebulon42:
@daganzdaanda See https://github.com/gravitystorm/openstreetmap-carto/issues/851#issuecomment-98685860.

nebulon42:
@matkoniecz Right, the stairs are barely visible, but visible. I still think the proposed icon is an improvement. Other opinions/ideas?

stations.mss
```diff
@@ -2,8 +2,10 @@
 
 .stations {
   [railway = 'subway_entrance'][zoom >= 18] {
-    point-file: url('symbols/walking.n.12.png');
-    point-placement: interior;
+    marker-file: url('symbols/subway_entrance.10.svg');
+    marker-placement: interior;
+    marker-fill: #555;
```
pnorman: this should probably be transit blue

kocio-pl:
Could we see how it looks now in blue on the map?

nebulon42:
Some color variations:

grey
subway_entrance_after

station color
subway_entrance_station_color

station text color
subway_entrance_station_text_color

transportation color
subway_entrance_transport_color

kocio-pl:
I see your examples in this order:
1. transportation (lighter than current, but still visible enough for this object - good for me)
2. grey
3. station
4. station text (map looks too much like a blue-red tonal rainbow =} )

nebulon42:
Any other opinions on different colours?

pnorman:
transportation colour.

nebulon42:
Alternatively we could use a generic entrance icon:
subway_entrance

kocio-pl:
Nice idea.

nebulon42:
I presented two options that would work more or less and are clearly improvements over the current icon. A decision on this would be great.

math1985:
I will get back to this as soon as possible.

kocio-pl:
I think general entrance version is even better, since it's more clean and readable (which is important given it's smaller than most icons) and if we decide to render plain building entrances, they'll be probably black, so the distinction will be clear.

BTW: I think we can make [elevator icon](https://github.com/gravitystorm/openstreetmap-carto/issues/720) to be in the same style, just with arrows up and down.

daganzdaanda:
Yeah, the general entrance with arrow is the better idea. The stairs are hard to decipher at that size and with the busy backgrounds. You probably already tried this, but maybe the arrow could point down 45° to symbolize entrance to the underground?

kocio-pl:
Nice idea.

matkoniecz:
Generic entrance works better for me.

nebulon42:
The consensus seems to be the generic entrance icon. Updated.
For other PRs that might want to use a generic entrance icon, it is now available.

PR #1522, opened 2015-05-04
reuse symbol of DIY shop for hardware shop

nebulon42:
Fixes part of #1460.


matkoniecz:
It is a bit different shop, but difference is probably not big enough to make a separate symbol.

nebulon42:
Rebased. Yes, it is a bit different, but the definition is a bit fuzzy. I think it might also be a good idea to have closely related shops reuse the same icon to prevent icon explosion, which might be confusing.

gravitystorm:
> I think it might also be a good idea to have closely related shops reuse the same icon to prevent icon explosion, which might be confusing.

On the other hand, reusing icons can mislead people into thinking that their is no difference between the tags. This is less likely to happen with an obviously generic icon (e.g. a dot) but more likely to happen with seemingly specific icons that are used for multiple different (even if similar) features.

kocio-pl:
Yes, we have to be careful, but it's good to know there's also this possibility.

nebulon42:
True, I was not suggesting to do this on a large scale. And of course there need not to be icons for everything.

PR #1526, opened 2015-05-05
Render cutline as thinner line

### LINKED ISSUE #781: Change rendering man_made=cutline

math1985:
This resolves #781.

Supersedes #1449.

Before:
Before

After:
After

### DIFF
--- a/landcover.mss
+++ b/landcover.mss
@@ -574,12 +574,12 @@
 /* man_made=cutline */
 #landcover-line {
   [zoom >= 14] {
-    line-width: 3;
+    line-width: 1.5;
     line-join: round;
     line-cap: square;
     line-color: @land-color;
     [zoom >= 16] {
-      line-width: 6;
+      line-width: 2;
     }
   }
 }


matkoniecz:
Also - I am not entirely sure whatever rendering it at all is a good idea, since I discovered this tag it seemed quite weird.

BushmanK:
Okay, now it's almost impossible to see cutline combined with footway. What was the point of making it thinner, to make map less readable?

math1985:
Do you have an example of this?

BushmanK:
Sure.
Here http://www.openstreetmap.org/#map=19/55.85272/37.76072 you can see an intersection of path, path+cutline and cutline
Here http://www.openstreetmap.org/#map=19/55.85231/37.75187 even worse case path and path+cutline
Here http://www.openstreetmap.org/#map=19/55.85199/37.74395 - path+surface=ground+cutline and path+surface=ground
in case of surface=ground, it's even harder to see the difference.

Cutline is usually something that is significantly wide. It makes a real gap between the trees, and was completely acceptable before. Maybe, on lower zooms it's okay to have it really thin, because it's just an overview map, but on higher zooms (16-18 or something) it has to be easily recognizable and it shouldn't make paths less visible (like in case of surface=ground).

Generally speaking, just try to look here http://www.openstreetmap.org/#map=15/55.8509/37.7437 and to say, where is what, especially if you have not ideal vision.

pnorman:
> Okay, now it's almost impossible to see cutline combined with footway. What was the point of making it thinner, to make map less readable?

Please don't insinuate that people are trying to degrade the map. There may be an issue with interactions between highway=footway and man_made=cutline, but it can be raised differently.

BushmanK:
I'm not trying to insult anyone, but making anything smaller/dimmer and so on (without any clear reason - I tried to find an explanation here, but unsuccessfully) very likely leads to less distinctive features and less readable map regardless of someone's intention. That's just a feature of human vision we should keep in mind when making changes.

kocio-pl:
@BushmanK See #781:
> The rendering of man_made=cutline doesn't fit into the rest of the map style. The line is far to wide within serveral zoom levels and to "strong" (continious line) so it can be easily misread as highway=residential or highway=unclassified and the "way importance" can therefor be overrated compared to highway=track and highway=path.

BushmanK:
@kocio-pl Okay, I see your point, but is changing cutline thickness the only way to make it look different from minor roads? What about making it lighter green and keeping continuous tree symbol pattern overlaying it (unlike in case of roads), for example? Here https://github.com/gravitystorm/openstreetmap-carto/issues/781#issuecomment-50401674 , relatively good suggestion was made (too dark for my taste, but still better than invisible thing line or buffer).

Because playing with width only obviously makes one thing better and another thing worse. Fixing and breaking something in the same time is usually considered nonsense.

kocio-pl:
It was not my point, I just answered you shortly what was the root cause to make any changes. 

I also like the comment you mentioned very much: no transparency is needed, just slightly lighter green plus pattern overlay (which is default now). I think it would be the best choice, since it completely avoid suggesting that there is any highway at all and even looks more like real cutline (typically grass I guess), while not being forest meadow (pattern overlay).

BushmanK:
@kocio-pl Good. Anyway, at least there is certain relatively common view on this.

HolgerJeromin:
A cutline is a line of missing trees in the forest. On its own you don't know if it is passable. 
With a footway the cutline is imo irrelevant as the thing "no dense trees here" is transported via the Footway itself. 
So imo we have no issue to solve.

BushmanK:
@HolgerJeromin 
Yes, cutline is not a passable way by itself.
But cutline is not just a line, where no dense trees are growing. Normally, it is significantly wider than just a gap between trees, and it always should (if it's not an old unmanaged cutline) give significant amount of clear sky between two rows of foliage.
When we have both cutline and passable way, going along it (located within a cutline, so we can map both passable way and cutline by tagging single osm way) these are two separate entities. Previously, it was easy to see these two entities sharing the same space, now it's problematic. Also, passable way in a forest is absolutely not equivalent to passable way on a cutline.

If you personally don't care about being able to see, if path goes through the forest or it goes through along the cutline in the forest - it doesn't mean nobody else should care. Especially keeping in mind that it was possible before.

I can give several examples, where it is important:
- if you need better GPS reception (open sky, especially when it's raining)
- if you do care about sunlight exposure (you want it or you need to avoid it)
- paths on cutlines are safer in terms of meeting large mammals (exactly in place of my examples above - Eurasian elk)
- if you need to know, where the cutline is by wildfire safety reasons.
It's easy to continue.

kocio-pl:
@BushmanK +1

Fast rendering test at z14 and z16 (thanks for the interesting test case!) - previous widths, just the color change from land to scrub:
cutline-14
cutline-16

We can change the line width, I just think it should be thin enough to not make artifacts (thinner than typical cutline) and to encourage people to draw real borders, like those scrubs. I like the color however and this concept works for me as expected.

The whole PR code is trivial, we just need to find what numbers we should use for width.

BushmanK:
@kocio-pl second variant looks better.
It gives space for paths (especially new thinner rendering) to be seen in contrast to background.
I also think, that width of cutline should be constant (like in your second variant), because currently on high zooms width is varying (thicker, where path is).
Color seems to be okay for me too.
Typical managed cutline is at least about 2-2.5m wide, which is equivalent to 7-8 pixels at zoom 18 in Central Europe. So, I think, that your second variant is more realistic.

kocio-pl:
These are not variants, it is just that at z=14-16 the line is depicted as thinner (3 px) than at z16+ (6 px). TileMill makes it hard to draw real box corresponding to given zoom level, I have to set it manually, because the zoom level means just "use the rules for given level at this box", it doesn't really "zoom" the area in the box.

BushmanK:
@kocio-pl I see.

pnorman:
> TileMill makes it hard to draw real box corresponding to given zoom level, I have to set it manually, because the zoom level means just "use the rules for given level at this box", it doesn't really "zoom" the area in the box

Use screenshots.

kocio-pl:
I somehow hate screen shooting. What's more important I'm not sure that what I see on the TileMill screen is the same as what I get by clicking Export, because I remember at least one time when on the screen I had some artifacts (namely multiple elevator symbols instead of one) which was not on the exported image. I don't understand it.

gravitystorm:
> what I see on the TileMill screen is the same as what I get by clicking Export

If I recall correctly, the Export button redraws the map image in one go, so it doesn't show e.g. boundary artefacts from metatiling. So a screenshot can be better, since if there's a metatile-boundary problem there, it's likely to be a problem on the tileservers too.

kocio-pl:
Thanks for explanations, it makes sense now! It looks like for now I have no better way for showing proper zooming than screenshots then.

OK, so what do we do with cutlines? The code is dead simple, but should I create new PR based on current state or this PR should be reverted before that - or maybe we need to further discuss the issue (new issue ticket? reopening of https://github.com/gravitystorm/openstreetmap-carto/issues/781)? If the previous version's (which is still current on the main website) width is good, then it would be just the color change.

matkoniecz:
> create new PR based on current state or this PR should be reverted before that

Create a new PR based on current code. Revert would be also a new PR based on the current code.

Generally it is better to open new issues than comment in merged PRs. Commenting in open PRs is a good idea as somebody is currently working on proposal and feedback is valuable. But comments in closed PR are likely to be lost.

PR #1471, opened 2015-04-04
render historic=monument like historic=memorial, fixes #758

matkoniecz:
historic monument name eeeeee eeeeee eeeeee ref 1 ele 8000 operator operator ee ee ee brand brand ee ee ee ee closed_way master - monument

two before/after on real data:

https://cloud.githubusercontent.com/assets/899988/6992697/a068e192-dada-11e4-87b7-89030d04179d.png

https://cloud.githubusercontent.com/assets/899988/6992698/a093a300-dada-11e4-95d3-f3d715944c5f.png

PR #1532, opened 2015-05-06
add rendering for historic=monument

### LINKED ISSUE #758: Add rendering for historic=monument

nebulon42:
Fixes #758, picking up after #1471.

I thought the landmark nature of monument deserves rendering from z16 on, but I'm fine with z17 too.

preview
monument

The preview is most likely wrong tagging, but you get the idea.


math1985:
I made the tagging list aware of this change:
https://lists.openstreetmap.org/pipermail/tagging/2015-May/024249.html

PR #1533, opened 2015-05-06
stop rendering railway=rail on z6 and z7 - fixes #1123

### LINKED ISSUE #1123: Drop railways from z6/z7
Zoom level 6 and 7 are currently very crowded. Both railways and some types of roads are rendered with a gray line. It would probably improve the legibility of the map if we drop railroads from these zoomlevels.

matkoniecz:


pnorman:
Could you provide some example screenshots? The area I have is not outstanding for this.

z6
image
image

z7
image
image

matkoniecz:
@pnorman In that area it is not a problem, but in eastern USA it is worse - and in Europe it is simply bad.

I tested on relatively small area, I plan on testing it with entire country in Europe (but before that I need to figure out why osm2pgsql on loading 0,5 GB pbf file file is managing to use 6GB of hard drive and later is not deleting temporary files).

BTW, is there any good way to download only cities, railroads and major roads, without downloading other data? Currently I am downloading extract for single country as extract for entire Europe is far too big - despite that only small amount of data is used at z6/z7.

I know that I may download big extract and filter it creating smaller one or download data in many parts from Overpass.

imagico:
> BTW, is there any good way to download only cities, railroads and major roads, without downloading other data?

You can do that with overpass - all railroads worldwide should be possible for example - could take some time of course.

The most problematic areas for not rendering railways are probably those regions where railways are an important means of transport and there are no larger roads parallel to the railway routes.  I am thinking of some parts of India, Australia, eastern Russia.

matkoniecz:
@imagico >The most problematic areas for not rendering railways are probably those regions where railways are an important means of transport and there are no larger roads parallel to the railway routes.

Yes, Trans-Siberian Railway on one side, regions with plenty of relatively unimportant railways on the another.

mboeringa:
> BTW, is there any good way to download only cities, railroads and major roads, without downloading other data? Currently I am downloading extract for single country as extract for entire Europe is far too big - despite that only small amount of data is used at z6/z7.

Have you ever tried QGIS? I found it very handy in downloading small to medium sized sections, like cities or provinces of countries. You can simply zoom in to the section you want, and it sets the extent to download automatically based on what is visible on the screen. The output is not *.pbf but *.osm though.

matkoniecz:
> Btw, if you are looking for a "worst-case" scenario, I recommend London. I know of no other city with such a dense railroad network... "All the railways lead to London"? ;)

Looking at the extremes is not really useful, I wanted to try something more balanced.

Example rendering - Poland, where I think that dropping railways is a clear improvement. To render larger area I will need to set up filter and keep only relevant data, 3 hours and 15GB to load a single country is too much for low zoom level testing.
selection_005
selection_003

selection_006
selection_004

pnorman:
I loaded up Europe overnight (10h) and had a look. Click on the images and make sure you're at 100% to get them at the right scale.
z6
production
rail

z7
production
rail

pnorman:
@gravitystorm @richardf could use a cartographic opinion on those samples

matkoniecz:
After hallway testing, sample size=2:

person A: dropping railways is clearly better on z6, was unsure about z7 (information vs readability) unsure, but dropping railways is slightly better.
person B: dropping railways is great idea on z6 as otherwise map is not readable, generally a good idea but z7 ("but maybe somebody would want to see railways during planning railway trip")

Also, in my personal opinion is that it is better without railways.

imagico:
Note this has the risk of being fairly culturally biased.  The relative importance of roads and railways for transport varies strongly around the world - there are countries with nearly no railway infrastructure at all while elsewhere it is the primary means of transport.  If you'd ask people in Germany for example if they prefer a coarse scale general purpose maps with or without railways most likely >90 percent will prefer it without - Germany is after all an automobile country.

A possible compromise approach would be to only show some railways - like those that belong to a railway route relation with at least a certain size.  Not sure if this is possible in the current setup though.

pnorman:
merged in d12a5bcc080ba825849c64ef74aaea26e182bbc9 for z6, not z7.

Given the other transportation shown on z7 (ferries, highway down to primary), I think it's best to keep them rendered.

PR #1540, opened 2015-05-11
adding antarctic icesheet rendering based on preprocessed shapefiles

### LINKED ISSUE #960: Making the Antarctic ice sheet show up in the map
Opening this issue as per suggestion by @gravitystorm in #646.  I am trying to summarize the current status here and add my thoughts on this at the bottom.  Please correct me if i misunderstood something about the proposed approach.

The problem:
-----------------

The [Antarctic ice sheet](http://en.wikipedia.org/wiki/Antarctic_ice_sheet) is currently not shown in the map since it is not explicitly mapped as a Multipolygon relation as per [wiki](http://wiki.openstreetmap.org/wiki/Antarctica/Tagging).  Solving this in a way as proposed in #646 is not considered a good solution.

The proposed solution:
------------------------------

Running a separate osmcoastline-like process to generate a shapefile for the ice sheet area, making this shapefile available and including it in the style to be rendered as the first layer after the coastlines in the same style as normal glaciers.

Comments:
--------------

First of all it is not clear to me if this suggestion is meant to produce a shapefile that represents the areas defaulting to ice as per the current mapping convention or if it is meant to represent the actual ice sheet area excluding all currently mapped non-ice areas and non-icesheet glaciers, in particular the bare_rock areas that are currently not rendered.  I suppose the latter since the former would be pointless (the shapefile would be identical to that generated by ogr2ogr in #646)

In the latter case the approach seems possible to me and would - when implemented - be an elegant solution for the problem.  It would also be more flexible in terms of styling than the approach in #646.  It would produce a number of difficulties however:

* It would require adding an identifying tag to all ways that need to be processed (in a similar way as natural=coastline identifies the coastline ways).  This cannot be done from the existing tagging alone.  This applies to about 60000-90000 ways probably.  Orientation of some of the ways would need to be switched as well.

* Processing would be subject to edits breaking it similar to the coastline.  Although editing frequency is low in the areas involved support for this in editors and QA tools would be essential if this is supposed to work with frequent updates.

* Since glaciers are normally drawn with an outline doing the same for the Antarctic ice sheet would require using an unsplit polygon (similar to the unsplit water polygons from osmcoastline) which is probably a problem performance wise.

I would not be able to invest the time required to address these difficulties but if there is someone willing to delve into this i can offer my advise based on my knowledge of the data involved.

If however an unsplit polygon and an osmcoastline like process are not required the problems above would not apply and it would be pretty much a piece of cake to produce such a shapefile - just collecting all non-icesheet areas in the region and subtracting them from the coastline polygons.

imagico:
Since the preprocessed Antarctic icesheet data is now regularly available on [openstreetmapdata.com](http://openstreetmapdata.com/data/icesheet) (where also the coastlines come from) this PR adds rendering of the Antarctic ice based on that data.  Style-wise this should be an exact equivalent to normal glacier rendering.

See also #960 for more background information.

Rendering examples:

z=6
McMurdo dry valleys
with explicitly mapped glacier

Resolves #960

### DIFF
--- a/get-shapefiles.sh
+++ b/get-shapefiles.sh
@@ -11,6 +11,8 @@ mkdir -p data/simplified-land-polygons-complete-3857
 mkdir -p data/ne_110m_admin_0_boundary_lines_land
 mkdir -p data/ne_10m_populated_places
 mkdir -p data/land-polygons-split-3857
+mkdir -p antarctica-icesheet-polygons-3857
+mkdir -p antarctica-icesheet-outlines-3857
 
 # world_boundaries
 echo "dowloading world_boundaries..."
@@ -65,6 +67,27 @@ unzip $UNZIP_OPTS data/land-polygons-split-3857.zip \
   land-polygons-split-3857/land_polygons.cpg \
   -d data/
 
+# antarctica-icesheet-polygons-3857
+echo "dowloading antarctica-icesheet-polygons-3857..."
+curl -z "data/antarctica-icesheet-polygons-3857.zip" -L -o "data/antarctica-icesheet-polygons-3857.zip" "http://data.openstreetmapdata.com/antarctica-icesheet-polygons-3857.zip"
+echo "expanding antarctica-icesheet-polygons-3857..."
+unzip $UNZIP_OPTS data/antarctica-icesheet-polygons-3857.zip \
+  antarctica-icesheet-polygons-3857/icesheet_polygons.shp \
+  antarctica-icesheet-polygons-3857/icesheet_polygons.shx \
+  antarctica-icesheet-polygons-3857/icesheet_polygons.prj \
+  antarctica-icesheet-polygons-3857/icesheet_polygons.dbf \
+  -d data/
+
+# antarctica-icesheet-outlines-3857
+echo "dowloading antarctica-icesheet-outlines-3857..."
+curl -z "data/antarctica-icesheet-outlines-3857.zip" -L -o "data/antarctica-icesheet-outlines-3857.zip" "http://data.openstreetmapdata.com/antarctica-icesheet-outlines-3857.zip"
+echo "expanding antarctica-icesheet-outlines-3857..."
+unzip $UNZIP_OPTS data/antarctica-icesheet-outlines-3857.zip \
+  antarctica-icesheet-outlines-3857/icesheet_outlines.shp \
+  antarctica-icesheet-outlines-3857/icesheet_outlines.shx \
+  antarctica-icesheet-outlines-3857/icesheet_outlines.prj \
+  antarctica-icesheet-outlines-3857/icesheet_outlines.dbf \
+  -d data/
 
 #process populated places
 echo "processing ne_10m_populated_places..."
@@ -77,6 +100,8 @@ echo "indexing shapefiles"
 shapeindex --shape_files \
 data/simplified-land-polygons-complete-3857/simplified_land_polygons.shp \
 data/land-polygons-split-3857/land_polygons.shp \
+data/antarctica-icesheet-polygons-3857/icesheet_polygons.shp \
+data/antarctica-icesheet-outlines-3857/icesheet_outlines.shp \
 data/ne_10m_populated_places/ne_10m_populated_places_fixed.shp \
 data/ne_110m_admin_0_boundary_lines_land/ne_110m_admin_0_boundary_lines_land.shp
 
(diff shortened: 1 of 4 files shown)


pnorman:
@gravitystorm thoughts, as this is a big change, involving new pre-processed data?

imagico:
Could we get this out some time soon?

If you have a look at previous discussions you can see this is my third attempt at getting this to render in the style and technically this is about as easy as it gets.

If getting support for this into operations takes some time i understand but it would be good to have a decision.

gravitystorm:
I'm happy to have these used, and taken from openstreetmapdata.com, but we need to have some source available to generate them in case that site is unavailable.

I realise there's some "magic sauce" with the coastlines wrt comparing them against each other every day, but I'm not concerned about that. We can generate alternative shapefiles from available code if required. Is there existing code available (not necessarily that run by openstreetmapdata.com, but that would be nice) to generate these ice sheets? If so, where does it live?

tomhughes:
See http://blog.imagico.de/easier-rendering-of-the-antarctic-land-ice/ for a description of how it is done and https://github.com/imagico/icesheet_proc for the code.

imagico:
Tom already provided the link - i just pushed the version running on openstreetmapdata.com - the only magic sauce running there in addition is zipping the results and maintaining feature count statistics from osmium_noice (if anyone wants to know - since processing began in March noice feature count decreased from 57278 to 57272 features, explicitly mapped glaciers increased from 634 to 793).  And of course we directly use the SQLite db produced by osmcoastline as input and not the shapefiles.

math1985:
@gravitystorm Does that answer your question, or is there anything more we need?

gravitystorm:
@math1985 I'm happy

HolgerJeromin:
For all involved people this is probably clear, but on the osmf tile servers there is more adaption needed on their side:
https://github.com/openstreetmap/chef/blob/master/roles/tile.rb#L53-74
A note should be added to a release note of a new osm-carto release.

math1985:
@imagico Thanks a lot!

@pnorman @gravitystorm @matkoniecz Please don't forget to run `get-shapefiles.sh` when pulling this revision. On deployment, `get-shapefiles.sh` should also be ran on the rendering server.

get-shapefiles.sh
```diff
@@ -11,6 +11,8 @@ mkdir -p data/simplified-land-polygons-complete-3857
 mkdir -p data/ne_110m_admin_0_boundary_lines_land
 mkdir -p data/ne_10m_populated_places
 mkdir -p data/land-polygons-split-3857
+mkdir -p antarctica-icesheet-polygons-3857
+mkdir -p antarctica-icesheet-outlines-3857
```
math1985: @imagico Shouldn't these two lines start with `data/`? In fact they don't seem to be necessary at all.

pnorman:
> On deployment, `get-shapefiles.sh` should also be ran on the rendering server

get-shapefiles.sh is not run on the rendering servers at all, nor is it particularly well-suited for production use, with a lack of consideration of failure handling.

math1985:
In any case there should be something on production that downloads the shapefiles...

tomhughes:
Already done in https://github.com/openstreetmap/chef/commit/cecd219c7ab80a25ceeb3e99879c9eb52f6ea6ff ;-)

math1985:
Thanks @tomhughes!

get-shapefiles.sh
matkoniecz: see #1699 for PR removing this

matkoniecz:
Is it intentional that borders are no longer rendered (unlike preview images)?

selection_002

imagico:
@math1985 - yes, paths were off but apparently this did not matter.  In general get-shapefiles.sh is very testing-unfriendly - you can only test it all together newly downloading all files.  This would be good to improve.

@matkoniecz outlines are drawn at ice-water edges and ice-land edges but not at ice-ice edges since those are already drawn by the explicitly mapped glaciers.  So if you don't have any glacier data in your database you get missing outlines at the edges, in particular towards the ice shelves.

Thanks everyone to finally get this out, after in total nearly 2.5 years - the land ice polygon in Antarctica was removed on March 16 2013.

matkoniecz:
@imagico Yes, I checked it without data loaded for this location.

gregrs-uk:
It would be great if the Antarctica shape files could be mentioned under 'Manual download' in INSTALL.md too. Thanks.

PR #1541, opened 2015-05-12
Fix explicit indentation modifiers in YAML

pnorman:
Not all of the SQL was reformatted with the change to YAML, and this fixes much of what was not done.

I believe there are no more big cleanups left.


pnorman:
rebased after zoo changes in #1274

PR #1558, opened 2015-05-19
move post icons to SVG

nebulon42:
Implements part of #1165.

before
post_before

after
post_after

While I like to keep the proposed design of post office as I don't like the old circled one I'm not so sure for post box. I have tried various things but none really worked.

What do you think of the larger/small approach? Or: Any suggestions for post box renderings (they should work at 10px width though)?


kocio-pl:
Maybe post box should use similar approach to the ATM (vertical with a "slot")?

RobJN:
I think they need a different design, not just different size. It's easy to tell them apart here as they appear next to each other, but when viewing just one on it's own it would be much harder.

kocio-pl:
Simple redesign as I suggested - works for me better than ATM:

1) post_box 12-1

nebulon42:
@kocio-pl I tried this already. If you successfully fit this into a 10px frame and make it look good I'm more than happy to include it.

kocio-pl:
I see no problem with it and I don't get what was yours?:
[2)](https://github.com/kocio-pl/openstreetmap-carto/blob/post_box-icon/symbols/post_box.10-1.svg) post_box 10-1

nebulon42:
IMO the "make it look good" part isn't really satisfied here, but I also didn't succeed in that.

kocio-pl:
I thought that may be the case! =} 

For me it works, but you can also use 1) without a slot (BTW: why it has to be 10x10? Maybe 12x12 with a slot is better). Just smaller icon doesn't work for me, because one can be easily mistaken.

RobJN:
How about a stamp instead? Perhaps just the shape and perforated (wiggly)
edge.

math1985:
> I think they need a different design, not just different size. It's easy to tell them apart here as they appear next to each other, but when viewing just one on it's own it would be much harder.

+1

nebulon42:
@kocio-pl The post box icon is currently 10x10 and there might be lots of them that should not clutter the map. But maybe rendering rules for post box have to be re-thought completely.
@RobJN Interesting idea, but could be hard to recognize.

Right, so a different shape, that was what I thought too, but I still need a good workable solution at this small size.

kocio-pl:
> The post box icon is currently 10x10 and there might be lots of them that should not clutter the map. 

So let's look at the current numbers of small, popular amenities:
- benches: 481 856
- post_box: 178 869
- recycling: 142 614
- waste_basket: 115 322
- atm: 83 588

(BTW - post_office: 116 016)

I see that the benches are the most dangerous thing if we talk about clutter (~500k), but the rest is much in the same league (~100k), so a danger is pretty the same for them. Making post box icon smaller than ATM is reasonable then (2x more), but it doesn't have to be so big difference in size. We can also use the rendering level to avoid clutter (see the recent PR "Move recycling icon to z17" https://github.com/gravitystorm/openstreetmap-carto/pull/1580), not only the size.

nebulon42:
I think I have found a pragmatic solution for the post box. Very similar what I have already tried before and was also suggested here.

daganzdaanda:
> I think I have found a pragmatic solution for the post box. 

I think it works! It's just a very small difference, but I think it can be enough, combined with the size difference.
But I still tried some versions with the letter rotated 45° (couldn't help it)...
https://gist.github.com/daganzdaanda/2f80c5a3c4a8ffd08d41
But these are not as clean. It's always about that tiny flap! Maybe if we showed the front of the letter, e.g. 3 lines for an address, it would work? I think it definitely would work with a 90° rotation as in @kocio-pl 's number 2.

nebulon42:
Thanks for the variations! I agree that the rotated versions all have a little glitch. The aligning of the gap was a nice idea though. I'm going to stick with the version I have now, but I accept that there might be other opinions. If necessary, I can also remove the post box from the PR and @kocio-pl or @daganzdaanda can try their suggestions.

kocio-pl:
I think that you current version may work only when both post items are close, because post box difference is too subtle. I made my test and I choose [12px vertical version](https://github.com/kocio-pl/openstreetmap-carto/blob/post-icon/symbols/post_box.12.svg) for it - it's more readable than 10px version and since the shapes are clearly different, we don't have to play hard with size:

post-z19-vertical

mboeringa:
I have a preference for @kocio-pl's last versions. I think the simple act of the rotated envelope is more easily recognizable as postbox and generally more universal "iconic-ally"

nebulon42:
Rebased. A maintainer decision is required here. I have no strict preference for my version, but I don't like the rotated versions so much. As I said I would just remove post box from this PR and this could then being added by somebody else.

RobJN:
I agree that @kocio-pl's last version works best although I wonder whether the slot should be at the top instead (the same as with the new ATM icon). Also, why are bank and ATM icons grey whilst these are brown? The two overlap quite a bit in the UK now (post offices provide basic banking and currency exchange services).

kocio-pl:
Both slot positions are right IMO - you typically get the money out of ATM, but put the letters inside the box.

You're also right with bank color - I think it should probably be amenity-brown, as all the other amenities.

RobJN:
Well the slot in post boxes here in the UK require you to push letters in
from below as a upper lip is there to keep rain out (hence my suggestion)
but other countries may be different.

It's a bit of extra detail ;-)

kocio-pl:
@RobJN Yes, physically it's the same in Poland, I just thought about abstraction of putting in/getting out, but your idea may work too. Actually I like it more than previous one now:

post_box-z19-up

@nebulon42 Maybe just do as you said and remove the post box part - I can handle it and make the PR soon, once we decide which shape we should use.

daganzdaanda:
For the icon, the line representing the slot needs to be on the bottom, IMHO. I think it has to do with the "reading direction" from top to bottom. So the letter "moves" from top to bottom, and if the slot is on the top, it moves away from it, it falls out of it.

pnorman:
I prefer the slot on top.

kocio-pl:
@daganzdaanda This is what I said about abstraction of putting in/getting out (that's exactly reading top down), but it looks like more people prefer physical interpretation.

math1985:
I'm happy to merge either of the vertical icons. I slightly prefer slot on bottom. I think vertical is clearly better than horizontal.

@nebulon42 WIll you still finish this PR, or would you prefer someone else to take over?

nebulon42:
@math1985 I have removed the post box icon from this PR so that others can deal with it.

math1985:
I don't think the new post office icon works very well together with the old post box icon, so I will close this for the time being until someone creates a PR that addresses both icons.

nebulon42:
Oh my. Why didn't you just leave it open until somebody did?

math1985:
At this point it's rather an issue than a PR right?

kocio-pl:
Don't worry, this way or another it will be done, I was just waiting for people to give more votes for both versions of post box icon.

PR #1559, opened 2015-05-20
move water park icon to SVG

nebulon42:
Implements part of #1165.
Smaller symbol.

before
water_park_before

after
water_park_after


math1985:
Looks good!

kocio-pl:
I would make the waves bigger, as I showed with the waterfall: https://github.com/gravitystorm/openstreetmap-carto/pull/1564#issuecomment-105233951

PR #1562, opened 2015-05-20
add rendering for amenity=taxi

### LINKED ISSUE #879: Add rendering for amenity=taxi
I think amenity=taxi should be visible as an icon - with name, if it has one - when it's a point. An area tagged like this could be (maybe) rendered like parking area but with taxi icon instead of "P".

nebulon42:
Resolves #879.
Rendered from z16 on like car sharing.

I'm not sure if areas should get a fill. It might be easier to spot big/important taxi places. On the other hand, which colour to use? I'm not in favour of the parking colour as it's not really parking and could be confusing.

preview
taxi


HolgerJeromin:
An area could get a parking-yellow stroke but no fill. Could work together with the icon.

math1985:
Some cities have loads of very small taxi stands, wouldn't it be better to move this icon to z17 or z18?

nebulon42:
I'm open for that. For Vienna z16 seemed to work.

Issue #1528, opened 2015-05-05
 Add special rendering for shop=chemist


kocio-pl:
[My proposition](https://github.com/kocio-pl/openstreetmap-carto/blob/shop-chemist-icon/symbols/chemist-14.svg) for this type of shop is this: ![chemist-14](https://cloud.githubusercontent.com/assets/5439713/7478217/64681ce8-f359-11e4-9d87-42e431d88d47.png)

[Mentioned on meta-issue #1460]

neiljp:
From the scale of the image that I can see, this is really quite unidentifiable. Zoomed in, I guess it's a toothpaste/brush pair?

kocio-pl:
Yes, I was looking for something identifiable by the shape and thought about this pair. Maybe you have other ideas?

daganzdaanda:
FWIW, I understood it as toothpaste and brush before I read the first comment. But that may be because I new that it was to signify a chemist.
On the map, the labels will help tell people what the icon means, until they have "learned" it. So it's probably fine. Anyhow, I can't think of a better way to signify a chemist, they sell all kinds of small things in tubes and boxes, not very iconifiable (is that a word?)

kocio-pl:
You may be right, but any criticism makes me think what could be done different, so I just come up with another idea, so we could at least compare which try is better:

2) ![chemist-14-4](https://cloud.githubusercontent.com/assets/5439713/7499206/5623fba2-f426-11e4-9a2c-73f984ce4d25.png)

daganzdaanda:
Is that a roll of toilet paper? cool ;)
...but actually, its not what I would think of first with a chemist. And, @nebulon42 tries to keep perspective in the new icons to a minimum, and I think we should try to follow that guideline.

kocio-pl:
Perspective - sure, but toilet paper clearly belongs to "A shop selling articles of personal hygiene, cosmetics, and household cleaning products." Meanwhile:
 
3) ![chemist-14-5](https://cloud.githubusercontent.com/assets/5439713/7499439/efcd7336-f427-11e4-9084-d909e4ebc1c6.png)

kocio-pl:
4) ![chemist-14-5](https://cloud.githubusercontent.com/assets/5439713/7505904/93d8ae98-f458-11e4-9788-d7828d873810.png)

[...]

PR #1563, opened 2015-05-20
Adding shop=chemist icon

kocio-pl:
Resolves https://github.com/gravitystorm/openstreetmap-carto/issues/1528.

chemist-after

Check carefully as I'm newbie in PRs.


nebulon42:
You should squash commits into one (maybe just before merge) and also mention the resolved issue in your commit message so it gets automatically closed (see https://help.github.com/articles/closing-issues-via-commit-messages/).

Please add a preview.

kocio-pl:
Preview added.

How to squash them? I thought git rebase will do it, but now I do it and no such effect appears.

nebulon42:
Try http://stackoverflow.com/a/5189600

kocio-pl:
$ git rebase -i master
warning: Cannot merge binary files: project.mml (HEAD vs. 897bdcf... Icon and YAML for shop=chemist)
error: could not apply 897bdcf... Icon and YAML for shop=chemist

nebulon42:
Have you edited project files more often than once? Maybe that's the reason. Have you force pushed?

Regarding the icon: This one is a bit blurry, especially at the bottom of the tube and at the brush head. I suggest making the brush handle thicker and replace the individual parts of the head by a solid part with zigzag outline on the left. Maybe a more straight tube could help against blurriness. On close inspection of the file I saw that you are using very subtle arcs that are unlikely to be visible at small sizes at all, either more curved or not curved at all would be better. Your nodes could be more aligned to the pixel grid.

kocio-pl:
New version:

chemist-after-nebulon

1. Brush head redone and handle is thicker. I liked the previous version because the head was visibly softer (like the material it is made of) and the handle had better proportion, but now it's more visible, so this is acceptable.

2. I see nothing wrong with the tube and it seems to me (previous session and today also) it can be only worse: straight tube (1 px less from both sides) is just a rectangle and it's harder to recognize, wider a bit and we loose the proper ratio.

3. Arcs was so subtle that it made sense only when looking at higher sizes - and that made it look better with SVG not damaging the small PNG. Now it's more aligned, but I don't see the real improvement on PNG, while SVG looks much worse to me now (head zigzag looks also worse here). I don't know if we don't care too much for SVGs, but I would like it to also get some love, because what's the use of trying to have them rather then PNGs then?

nebulon42:
Definitely an improvement! The tube bottom has a bit of blur but let's leave it at that for now.

daganzdaanda:
We're really deep in nit-picking territory here ;) but I'd combine the toothbrush with the fat handle and the first kind of brush-hairs. Like you said, the anti-alias gives a semi-transparent look and for this part it works well for me (same as in shop=books)

kocio-pl:
Probably the best:

chemist-after-daganzdaanda

> We're really deep in nit-picking territory here ;)

Well said. =} But that's good - it means to me that overall shape works and it's fun to test other people ideas, and sometimes they are really better or at least interesting and just different.

PR #1564, opened 2015-05-20
add rendering for waterway=waterfall

### LINKED ISSUE #336: Add rendering for waterway=waterfall
It would be nice if mapnik would display a waterfall icon for nodes marked with waterway=waterfall

I have created http://wiki.openstreetmap.org/wiki/Waterfall some time ago and this part  seems fairly stable.

nebulon42:
Resolves #336.
Renders waterfall with name from z16 onwards.
Does not render `natural=waterfall` because not documented on the Wiki, does not render lines and ways because this is discouraged on the Wiki.

preview
waterfall

It is hard to find a colour that works within forest. It should be blue, so I decided to reuse the colour for hotels, campsites, transportation, etc. Still to be discussed.


kocio-pl:
For me and another mapper the icon doesn't work.

imagico:
I would much prefer a bar drawn across the waterway to symbolize a waterfall but i know it is technically difficult to get the necessary orientation.

Color wise i thought this blue is reserved for transport/travel stuff.

HolgerJeromin:
I think a blue is needed.
You could generate it via something like `darken(@water-color)` to have them linked.
Rotating the icon would be nice, but i think it would make recognizing it also harder. We have probably no idea which is the flow direction, so a rotation is probably wrong by 180 deg.

nebulon42:
@kocio-pl Since you discussed so many shop icons lately I would expect a bit more detail than just "does not work". :)

@imagico I thought that it is not possible to compute the direction from a point location, same problem as with mountain pass? A bar might be probably confused with weirs, but that depends on the actual look. This blue is also used for accomodation and camping, so not really reserved. But the question is if it is better to invent a new colour or reuse an existing one. IMO it would be best to have the same blue for all water features (maybe also barriers). It should be darker than the waterway blue, but maybe the transport colour was not the best choice.

@HolgerJeromin I have tried that already, but if you darken and saturate the water colour it gets greenish very quickly.

I thought maybe the blue of ferry routes would be more connected to water features? What do you think?
waterfall_ferry_blue

imagico:
Direction should be possible to compute by querying the intersecting waterway - i think there are no precedents for this within the style at the moment though.

A bar across the waterway is a very common rendering of waterfalls in maps - like in German and Swiss topographic maps.  I don't think confusion with weir is an issue, a weir is essentially an artificial waterfall and you can vary the styling to make the distinction (like cliff/embankment)

kocio-pl:
> Since you discussed so many shop icons lately I would expect a bit more detail than just "does not work". :)

Sorry, trying to make 2 PRs ready made me that dumb... ;-P

I guess the color (dark blue) is right, but the picture looks more like underscored letter "M" for me, so I would look for another visual metaphor. I think @imagico may have some clue: what if waterfall was basically a blue cliff rendered with a name? It can be especially true for wide rivers and would have a proper orientation and shape as a bonus.

Maybe 2 more distinct streams (instead of 3) would look better, if we still try with the icon.

Weirs are very similar, maybe they can be depicted the same way (+ a wall when needed)?

RicoZ-OSM:
regarding direction, the wiki already suggests drawing a cliff as the waterfall edge which should share the node waterway=waterfall with the river.

So if it could be done to draw that cliff in a special way it would be nice. If it is technically difficult to query whether a cliff has a node marked with waterway=waterfall there is the option to tag the cliff-way with something like waterfall_edge=yes.

StyXman:
@RicoZ-OSM: that works if the waterfall is the size of the cliff, like in Iguazu, but not for waterfalls that are just a node like:

http://www.openstreetmap.org/node/1315639699

I understand that in fact the mapping is not correct; the cliff is missing. But if it were there, the cliff would be longer than the waterfall. The only way would be to make a segment of the cliff the width of the waterfall and tag as as such, not a node.

kocio-pl:
I don't have the data with waterfall area currently to show real life test, but that's what I think the icon should look like more in general:

2) waterfall 16

- bigger waves (to not look like an underline)
- 2 streams (to give it more space)
- disconnection between streams (to not like a letter - "n" in this case)
- more vertical bottom of streams (to make the water flow more visible)

nebulon42:
@kocio-pl You did not link your icon, could you provide a side-by-side comparison between mine and yours?

kocio-pl:
I just wanted it to be a proof of concept, so it's ugly inside, but [here you are](https://github.com/kocio-pl/openstreetmap-carto/blob/waterfall/symbols/waterfall.16-2.svg). =}

My initial thought was that it looks too much like a letter, but after I reworked your icon I think bigger waves are the single most important element, but I would still try to disconnect upper parts of streams.

nebulon42:
Concentrating on moving the rest of symbols to SVG, therefore closing this.

daganzdaanda:
I guess I've never seen an icon for waterfall. So I have no idea what it is "supposed" to look like, and I would need to learn any new icon.That said, both two and three sloshes(?) would be OK for me, but I do like the bigger waves. I tried to sketch some variants on paper with a cloud to show the mist, but those did not turn out better.

Issue #763, opened 2014-07-25
Add rendering for amenity=clinic


math1985:
Jóhannes Birgir Jensson reports on the talk mailinglist:

"Also worrying that amenity=clinic still isn't displayed with the cross like pharmacy and others, in many rural places this is the only healthcare around.
http://www.openstreetmap.org/node/2303511227"

kocio-pl:
Let it be like this - or maybe the same as the hospital or the same as the amenity=doctors? I'm not sure what are key difference between such objects and having the clinics visible is probably even more important than rendering "doctors".

Stalfur:
Amenity=doctors is now rendered. As most clinics have doctors (but not all) it seems like a easy and acceptable solution

kocio-pl:
So what icon shape do we choose:
1. Hospital cross without the circle.
2. Doctors bag.
3. hospital cross (maybe the same as in hospital)?

According to english Wikipedia clinic may be small local facility, but "[s]ome clinics grow to be institutions as large as major hospitals, or become associated with a hospital or medical school, while retaining the name 'clinic.'"
[ https://en.wikipedia.org/wiki/Clinic ]

So this can be different objects in reality (small/medium/large, independent/part of hospital), hence the difficulty with finding the right symbol.

Stalfur:
I for one would like to see 1. Hospital ross without circle

In my area the clinic has general doctors (who refer cases onwards to specialists if needed) and nurses, while also being responsible for infant checkups and inoculation. I would only go to a hospital for a major thing.

I believe the same situation is in other less developed areas where access to a hospital is even more ardous.

kocio-pl:
I still have no clear preference for symbol, but I think that it should be rendered form z>=16 (simply because hospital is z>=15 and doctors is z>=17).

kocio-pl:
I made the simple cross icon at last (baby steps with Inkscape), but I don't know how should I publish it now, because GitHub doesn't accept SVG ("Unfortunately, we don't support that file type. Try again with a PNG, GIF, or JPG."). What should I do now?

matkoniecz:
Are you trying to edit repository using online editor? 

I would recommend using git client (for starting with git I used https://try.github.io/levels/1/challenges/1 and http://git-scm.com/book/en/v1/Getting-Started )

But it is possible to add SVG file using online editor.

[...]

PR #1565, opened 2015-05-20
Adding amenity=clinic icon

kocio-pl:
Resolves https://github.com/gravitystorm/openstreetmap-carto/issues/763.

You can compare it with the hospital icon. As clinic may be similar, a part of the hospital (as here) or just something like doctors, it's similar to hospital, but rendered later and with less prominent name.

z16
clinic-after-z16

z17
clinic-after-z17

z18
clinic-after-z18

z19
clinic-after-z19


dieterdreist:
> Am 21.05.2015 um 00:27 schrieb kocio-pl <notifications@github.com>:
> 
> As clinic may be similar, a part of the hospital (as here) or just something like doctors, it's similar to hospital, but rendered later and with less prominent name.


sounds like a tagging problem, if it's big I'd rather use the hospital tag even if it's called "clinic" in your language, for reference see here
http://wiki.openstreetmap.org/wiki/Tag:amenity%3Dclinic#Meaning

and here(eg):
http://www.openstreetmap.org/relation/3334981

kocio-pl:
The name is not important, the tagging is, and this example shows this (I also know some fast foods with "restaurant" in the name).

nebulon42:
I'm not sure if clinic should get its own icon. Maybe either hospital or doctors icon? This one is a bit too close to hospital. I have no different proposal for another icon though. For me the concept of clinic is not yet clear, maybe because we don't have them here.

kocio-pl:
See the discussion in https://github.com/gravitystorm/openstreetmap-carto/issues/763 - I already asked about how should it look like and ended up with it (cross alone was the first idea, however I was not really satisfied with it). I would like more people replying, but the response was too weak to really know, so I decided it would be good to be like this.

polarbearing:
> I'm not sure if clinic should get its own icon.

Yes please. The concept fits for different aggregations of doctors in different cultures, or a rural facility that is visited by a doctor regularly. 

In eastern Europe you have Polyclinics, in Germany you have the Ärztehaus, an aggregation of legally independent (self-employed) doctors in one building. From the mapping perspective, that is easier to tag in the first step than researching a dozen surgeries within.

The lighter cross in the circle works well in proximity to a hospital, but is harder to recognise when on its own. 

Maybe the doctor's bag with a circle, or a roof?

kocio-pl:
I will try with the circle, but what's the metaphor/rationale behind adding roof to the doctor's bag?

polarbearing:
> what's the metaphor/rationale behind adding roof to the doctor's bag

doctor's bag currently used for amenity=doctors, and clinic would be "a house of doctors", roughly translating the German "Ärztehaus". 

A plurality of doctors in the icon would be even better, but I have no idea how to symbolise that in a small icon.

dieterdreist:
> Am 22.05.2015 um 11:29 schrieb polarbearing <notifications@github.com>:
> 
> doctor's bag currently used for amenity=doctors, and clinic would be "a house of doctors", roughly translating the German "Ärztehaus".
> 
> A plurality of doctors in the icon would be even better, but I have no idea how to symbolise that in a small icon.
> 


clinic is a bad tag because it is ambiguous (several meanings), so it remains unclear what kind of object it is

kocio-pl:
Still got no satisfaction:
2) clinic 16-2
3) clinic 16-3 (even with using 15 pixels instead of 14)
4) clinic 16-4

The key point for me is even if there's a subtle tagging problem, the most important is what we really try to show and the clinic is definitely important health facility. Last resort for me would be showing clinic with just a plain doctor's bag, because this way we would hide all the nuances and show that this is less-than-a-hospital, but still a medical POI. This may be even more true because the doctor's office definition is also not clear: "These range from a single doctor's office to larger establishments". 

What about that?

nebulon42:
Doctor's bag would be fine for me, but let's wait for other opinions.

kocio-pl:
Sure - the more opinions, the more probability of doing something in a proper way.

daganzdaanda:
After reading the wiki pages (also on doctors and health care proposals) I'm inclied to say that the doctor's bag alone is enough for "clinic". It's kind of the lowest common denominator: there will be medical help at the location, but it can be much less than a full hospital, so we should not have an icon that might be mistaken for a hospital.

math1985:
+1 for doctor's bag.

kocio-pl:
General question: when should I merge new master changes? I've been fixing merge few times already with my few PRs and it's good as a learning experience, but it's tedious and error prone (very long YAML lines...), so I'd rather do it only when I know someone wants to merge given PR.

math1985:
If it's simple changes I guess we can resolve the conflicts when we do the merging.

project.yaml
```diff
@@ -143,7 +143,7 @@ Layer:
           FROM (SELECT
               way, COALESCE(name, '') AS name,
               ('aeroway_' || (CASE WHEN aeroway IN ('apron', 'aerodrome') THEN aeroway ELSE NULL END)) AS aeroway,
-              ('amenity_' || (CASE WHEN amenity IN ('parking', 'bicycle_parking', 'university', 'college', 'school', 'hospital', 'kindergarten', 'grave_yard', 'prison', 'place_of_worship') THEN amenity ELSE NULL END)) AS amenity,
+              ('amenity_' || (CASE WHEN amenity IN ('parking', 'bicycle_parking', 'university', 'college', 'school', 'hospital', 'kindergarten', 'grave_yard', 'prison', 'place_of_worship', 'clinic') THEN amenity ELSE NULL END)) AS amenity,
```
math1985: Do we need to change the landcover layer?

kocio-pl:
I treat the clinic as a hospital, so if it's big enough, we should indicate it IMO (still it's less important, so icon is rendered bit later). I would also do it for doctor's office, because I know some of them that occupy the whole building and the area around it.

kocio-pl:
[Here](http://www.openstreetmap.org/way/327763405) is an example of doctor's office which is big, and [here](http://www.openstreetmap.org/node/3559448622) we have a hospital that probably does not occupy even one full building (I know both, so it's not a mistake). The differences between doctor's office, clinic and the hospital are rather blurry then, because one may think it's the size which helps recognize them, but it's not, and the name can be also misleading. That's why I support rendering areas for all such amenities as we do with hospital currently. I think different icons and zoom levels are good hints how important health facility may it be, but the area rendering should depend only on the size.

lest69:
At least in Canada, clinics are much more similar to a doctor's office than a hospital. A doctor's office and a clinic will tend to have similar levels of care and equipment and are outpatient facilities, while a hospital will tend to be much more advanced and include things like emergency care, inpatient beds, dedicated specialty departments (e.g. renal, oncology, ICU, maternity), etc.

My opinion is that the doctor's bag would be the best choice, or possibly another different icon if someone can think of one. For me, anything similar to the hospital icon is a no-go, since they're very different facilities.

Stalfur:
I see the doctors bag is on the map now for clinic, a choice I find to be odd since many clinics, particularly in rural 3rd world, don't have doctors but nurses - with occasional visits by doctors.

Issue #1523, opened 2015-05-04
Add special rendering for shop=garden_centre


kocio-pl:
My propositions: ![garden_centre-14](https://cloud.githubusercontent.com/assets/5439713/7452238/48246e32-f25b-11e4-9bbe-988c7810cdd8.png) ![garden_centre-14-3](https://cloud.githubusercontent.com/assets/5439713/7452232/24acbb12-f25b-11e4-854d-4a67741f4c1a.png) ![garden_centre-14-6](https://cloud.githubusercontent.com/assets/5439713/7452529/cdb49cfe-f25e-11e4-96bb-10f13caf2bfe.png) (and [this one](https://github.com/kocio-pl/openstreetmap-carto/blob/shop-garden_centre/symbols/garden_centre-14.svg) ![garden_centre-14](https://cloud.githubusercontent.com/assets/5439713/7455304/2eb13f6e-f27b-11e4-96f0-4e4aa8bfb1a0.png) is probably the best of them).

[Proposed and talked about in https://github.com/gravitystorm/openstreetmap-carto/pull/1520, mentioned on meta-issue https://github.com/gravitystorm/openstreetmap-carto/issues/1460]

matkoniecz:
This seems a good icon, also for me the last one is my favorite.

daganzdaanda:
Agree, a good idea, and the last icon is probably the best execution.

PR #1567, opened 2015-05-21
Adding shop=garden_centre icon

kocio-pl:
Resolves https://github.com/gravitystorm/openstreetmap-carto/issues/1523.

garden_centre-after-z19


nebulon42:
Here I would suggest making the "pipe" and handle thicker and the handle less wide as the body. The tilted right side looks a bit odd, maybe having a straight line with a rounded corner at the bottom would look better?

kocio-pl:
Next version then:

garden_centre-after-z19-nebulon

nebulon42:
Better regarding thickness, I'm still not satisfied with the right side and the handle spanning the whole width of the bucket. Your decision.

kocio-pl:
Why should I not try? But I hope I misunderstood you, because that would be unrecognizable and awful for me. =} - garden_centre-14-7 

If I really did, please try to describe it with other words maybe.

nebulon42:
I tried to make it more symmetrical: garden_centre-14
But I'm not really satisfied with the shape overall. Maybe it's enough if you move the lower right corner in your icon a bit to the right.

kocio-pl:
For me the shape is perfect now and even this slight correction you suggested made it less good, so I will stick to the current one. 

I also liked my original icon because its design was seen regardless such small size, but it's not a design contest and more typical shape is probably better for default map.

matkoniecz:
https://github.com/gravitystorm/openstreetmap-carto/pull/1567#issuecomment-104444340 is my favorite

I wonder is it because that watering cans of that shape are more popular in Poland.

kocio-pl:
Even if it was local influence, I like when the shape is not too square, because even 2x1 px difference between top and the bottom makes recognizing much easier ().

kocio-pl:
Sorry, that was just accidental key pres...

Issue #1534, opened 2015-05-06
Add special rendering for shop=confectionery


kocio-pl:
My propositions for this icon are: 
1) ![confectionery-14-1](https://cloud.githubusercontent.com/assets/5439713/7497942/bcc7e55c-f41d-11e4-9bef-cf8532b80f28.png)
2) ![confectionery-14-3](https://cloud.githubusercontent.com/assets/5439713/7497945/c039dbe6-f41d-11e4-8d84-a4f7d79e2587.png)

2) is probably better, because it's more readable.

[Mentioned on meta-issue #1460]

math1985:
2) looks like a pretzel to me, which is usually not sweet and thus not confectionery.

daganzdaanda:
Yep, we had the pretzel for bakery up to a few months ago! 
I like the sweet, but it might look a bit small compared to other icons. Also, if you put it in the middle of the 14px box, it will have some extra distance from the label. 
Maybe rotate the sweet by 45° to use more of the space?

kocio-pl:
IMO rotation doesn't help too much - it even does not give us more space: ![confectionery-14](https://cloud.githubusercontent.com/assets/5439713/7506707/aef1bd72-f463-11e4-99e8-139b7ace23fb.png)

HolgerJeromin:
I like the 45° sweet. 
Perhaps a 0° sweet would be better if it would have a round core, too.

Stalfur:
The sweet is nice and something like that is the best bet. The pretzel and the cake remind me more of a bakery.

matkoniecz:
> I like the 45° sweet.
> Perhaps a 0° sweet would be better if it would have a round core, too.

I have the same opinion.

kocio-pl:
OK, so 5) ![confectionery-14-5](https://cloud.githubusercontent.com/assets/5439713/7527581/23e40d06-f51d-11e4-8869-962fb6186553.png) or 3) ![confectionery-14-1](https://cloud.githubusercontent.com/assets/5439713/7506937/c80f7822-f467-11e4-807d-49ea06280a87.png) then?
6) Or maybe this: ![confectionery-14-6](https://cloud.githubusercontent.com/assets/5439713/7527657/175b73b6-f51e-11e4-80d9-22422ea22693.png)?

Stalfur:
5 or 6, 3 looks like a Star Wars tie fighter!

kocio-pl:
[...]

PR #1568, opened 2015-05-21
Adding shop=confectionery icon

kocio-pl:
Resolves https://github.com/gravitystorm/openstreetmap-carto/issues/1534.

confectionery-after-z19


nebulon42:
Looks good. I would suggest making the darker bottom area less tall and the top line beneath the cherry thicker instead.

kocio-pl:
OK, what about now?:

confectionery-after-z19-nebulon

nebulon42:
That's better.

PR #1574, opened 2015-05-23
move windmill to SVG

nebulon42:
Implements part of #1165.

before
windmill_before

after
windmill


kocio-pl:
I would cut a triangle's corners on the bottom, so it has two vertical borders (that would be pentagon than, to be strict).

math1985:
I think it might look better without the small element between the top wings.

nebulon42:
I'm not sure if it looks better, but it is definitely problematic at bigger sizes. Here is the preview:
windmill_2

I have cut the triangle at the bottom.

kocio-pl:
I would make the cut more visible, like this: 
windmill-16-1

pnorman:
I'd prefer no triangle in the top, cut off the bottom corners

kocio-pl:
Very small element on top is acceptable? It would make SVG better and IMO doesn't spoil the shape:
windmill-16-2

pnorman:
At the target size for the icons, I don't think it ever makes sense

kocio-pl:
I like it even with this size, because it suggests the building behind (so is less sketchy or abstract), but if it just doesn't hurt, it's better to keep SVG nice, because the icon might be rasterized one day for more pixels (for example on client's system, not on the server).

nebulon42:
I made the top part smaller and the lower part a bit less wide. I'm not able to remove the top part completely because that wouldn't look good at bigger sizes. I do not want to make it less wide at the bottom because the tower should become narrower at the top.

daganzdaanda:
I would prefer the topless one from https://github.com/gravitystorm/openstreetmap-carto/pull/1574#issuecomment-107404595  It's not bad that a symbol looks reduced, IMHO.
Just for reference: https://de.wikipedia.org/wiki/Datei:Map_symbol_windmill.svg

Issue #1405, opened 2015-03-18
Add special rendering for shop=photo and shop=photo_studio


kocio-pl:
We should start rendering shop=photo and shop=photo_studio, probably with this icon from osmfr-cartocss fork:

https://github.com/cquest/osmfr-cartocss/blob/master/symbols/shopping_photo.svg

[It was proposed in meta-issue #1402]

kocio-pl:
[My proposition](https://github.com/kocio-pl/openstreetmap-carto/blob/shop-photo-icon/symbols/photo-14.svg): ![photo-14](https://cloud.githubusercontent.com/assets/5439713/7485361/b1b74a2e-f394-11e4-8e2d-c4483fa0c35b.png)

daganzdaanda:
I like it!

PR #1576, opened 2015-05-24
Adding shop=photo icon

kocio-pl:
Resolves https://github.com/gravitystorm/openstreetmap-carto/issues/1405.

photo-after-z19


mboeringa:
Nice one, does the job well... although here in the Netherlands, dedicated photo shops are starting to get "endangered species" status... but that is another story!

Rovastar:
I think everywhere they are an "endangered species" it does look like another endangered shop those selling cameras to an average browser though.

Issue #1537, opened 2015-05-07
Add special rendering for shop=car_parts


kocio-pl:
1) [My proposition](https://github.com/kocio-pl/openstreetmap-carto/blob/shop-car_parts-icon/symbols/car_parts-14.svg) for this kind of shop: ![car_parts-14](https://cloud.githubusercontent.com/assets/5439713/7517925/2e9c8a2e-f4d8-11e4-92ea-62c30629fc5c.png)

[Mentioned on meta-issue #1460]

matkoniecz:
This icon is not really working for me.

kocio-pl:
One-man survey indicated it would be a scrap, however it's [far in the queue for adoption](http://wiki.openstreetmap.org/wiki/Tag:industrial%3Dauto_wrecker) (about 0,1k...). I was expecting that may be too distant association, so - back to reality:

2) ![car_parts-14-2](https://cloud.githubusercontent.com/assets/5439713/7527867/c3cf9c88-f520-11e4-8c12-5679b7b46291.png)
3) ![car_parts-14-3](https://cloud.githubusercontent.com/assets/5439713/7527962/c26fe04a-f521-11e4-9270-7d0e99958a13.png)
4) ![car_parts-14-4](https://cloud.githubusercontent.com/assets/5439713/7527964/dd702d1e-f521-11e4-9b5f-afdfde230638.png)

marcgemis:
what about a car exhaust or shock absorber ? would that be doable on such a small space ?
I wonder whether I would recognize 2 as a wheel on a map.
and 3 is pretty close to the icon of Mercedes , not ?

kocio-pl:
I doubt car exhaust or shock absorber would work - maybe some photos/icons showing their shape could help here? Another idea is to use spark plug or headlamp, but the first one will be probably too geometric, and lamps have many shapes, not always easy to recognize. 

Maybe old car horn ([like this](https://commons.wikimedia.org/wiki/File:Znak_B-29.svg) or [this](http://cdn.mysitemyway.com/etc-mysitemyway/icons/legacy-previews/icons/yellow-road-sign-icons-media/001791-yellow-road-sign-icon-media-horn-sc48.png)) would do, because the shape is great, but I have to try if it fits so small matrix well enough.

marcgemis:
http://www.gettyimages.com/detail/illustration/car-parts-performance-icons-royalty-free-illustration/166007126  second on the first row, first on the second row 

[...]

PR #1581, opened 2015-05-25
Adding shop=car_parts icon

kocio-pl:
Resolves https://github.com/gravitystorm/openstreetmap-carto/issues/1537.

car_parts-large-after-z19


nebulon42:
I think that the plus and minus sign could be a bit thinner and the connectors on top maybe a bit wider.

ximex:
i agree with @nebulon42

kocio-pl:
Sounds reasonable and looks good too:
car_parts-z19-nebulon

What do you think?

ximex:
much better

Issue #1459, opened 2015-03-23
Add special rendering for shop=kiosk


kocio-pl:
We should start rendering shop=kiosk (over 45k objects on taginfo), probably with this icon from osmfr-cartocss fork:

https://github.com/cquest/osmfr-cartocss/blob/master/symbols/kiosk.svg

[It was proposed in meta-issue #1402]

nebulon42:
[kiosk-18.svg](https://github.com/nebulon42/osmic/blob/master/shop/kiosk-18.svg): ![kiosk-18](https://cloud.githubusercontent.com/assets/3531092/6779943/14884b68-d161-11e4-94c6-22a66aca0fd9.png)

kocio-pl:
I could not recognize it probably (for me it looks now like an oven covered with with bananas... =} ). I'm not sure what to do, but maybe just removing the top bar could help? I can think also that the roof part can be made simpler (just those hanging elements?).

BTW: I'd like to start creating icons myself with Inkscape. What are the most important tips for making them pixel-aligned?

nebulon42:
For a start you could take the kiosk icon and try to improve it. :) Pixel-aligned means that the nodes of the path are snapped to the grid. Bring up the grid in Inkscape by pressing '#' and compare where the nodes are located. See also #632.

kocio-pl:
I may start with the clinic, since I'm not too familiar with vector graphics. =} 

I think basic guidelines regarding pixel-aligning could be added here:
https://github.com/nebulon42/osmic/blob/master/CONTRIBUTING.md#technical-aspects

kocio-pl:
My proposition for kiosk: https://github.com/kocio-pl/openstreetmap-carto/blob/shop-kiosk-icon/symbols/kiosk-14.svg

math1985:
I think @nebulon42's version is clearer.

matkoniecz:
I prefer @kocio-pl version so it may be cultural thing (typical kiosk style in a given area). Note that I tested neither on a map, it is based just on looking at svgs.

kocio-pl:
I was not fully satisfied with my take, it was just a first try to make something more simple than Osmic. @nebulon42 has a lot of creative ideas when drawing the icons, but sometimes it hurts recognizability too bad for my taste and kiosk was beyond my power to stay calm. ;-)

[...]

PR #1584, opened 2015-05-25
Adding shop=kiosk and shop=newsagent icon

kocio-pl:
Resolves https://github.com/gravitystorm/openstreetmap-carto/issues/1459.

kiosk-z19-name


nebulon42:
The icon is fine (for newsagent), but do kiosk and newsagent overlap enough that the could use the same icon?

kocio-pl:
I'm pretty sure they are. The main difference in a form - kiosk is a small space and the customers buy through the window, while newsagent is a more typical shop, where you go inside. And this is probably even clear from the context most of the times: small building alone means rather kiosk, a booth inside a building is rather newsagent, but I guess for a user that doesn't really matter, it's what they expect to buy there and (speaking from my experience) in both types there are newspapers, sweets, cigarettes and a lot of small household things, so I can spot no difference.

And [the definition](http://wiki.openstreetmap.org/wiki/Tag:shop%3Dkiosk) also supports it:

> A Kiosk is a little stand-alone pavillon selling products like cigarettes, newspapers, sweets, snacks and beverages. The shop is normally so small that customers cannot enter the shop. Kiosks are usually built on the pavement, or inside other buildings like train stations or shopping centres. 

>  shop=newsagent is a very similar type of shop to a kiosk in terms of products sold (primarily newspapers & magazines). However a newsagent is a proper shop, so a slightly bigger building or a normal sized shop in a row of shops. Note that this even holds for shops with the word 'Kiosk' in the name, such as K Kiosk in Switzerland and Luxembourg

dieterdreist:
> Am 26.05.2015 um 12:16 schrieb kocio-pl <notifications@github.com>:
> 
> but I guess for a user that doesn't really matter, it's what they expect to buy there and (speaking from my experience) in both types there are newspapers, sweets, cigarettes and a lot of small household things, so I can spot no difference.


as a kiosk is about a building type, it could also be a place that sells food or coffee or tickets for something or flowers etc, or am I wrong? The kind of goods that are sold will vary according to the region

cheers 
Martin

kocio-pl:
You have the florist kiosk on the same picture. =} We can't check the real meaning of tagging in each case, but the definition and usage are clear for me.

You can also buy coffee at the convenience shop or at the gas station shop and this is a big general trend I see for mixed-type shops (at least here in Poland).

As for regions: "big scale" problem will hit us in many ways and regional differences are very important. Trying to map the whole world at all levels means a lot of "uncommon" things outside the West (and even inside). However I don't think it's a rendering problem. We were discussing also [shop=confectionery / pastry / candy / sweets](https://lists.openstreetmap.org/pipermail/tagging/2015-May/024374.html) lately and it become apparent that our current tagging model is not flexible enough, because even such simple item as bakery may sell sweets too (like in my city) - but that would make any tagging impossible. 

I would like to start the gradual change to have more granular tagging (like "small + shop + flowers") and flexible categories (to avoid philosophical questions like "is the travel agent an office or shop?"), but currently the community seems to not see the need for it (even the general problem, only the problematic cases), so I try to stick to current rules.

dieterdreist:
> Am 26.05.2015 um 13:17 schrieb kocio-pl <notifications@github.com>:
> 
> You can also buy coffee at the convenience shop or at the gas station shop and this is a big general trend I see for mixed-type shops (at least here in Poland).
> 


yes, my suggestion for these cases is to use a property if you care about a particular item, eg 
sells:tobacco=yes
sells:ice_cream=yes
sells:coffee_to_go=yes
...

PR #1588, opened 2015-05-26
Adding shop=stationery icon

kocio-pl:
Mentioned on https://github.com/gravitystorm/openstreetmap-carto/issues/1460

stationery-z19


nebulon42:
Good idea, maybe a bit edgy. Would it be possible to have it thicker?

kocio-pl:
May be hard, because we would probably loose the proper ratio then (we have no spare vertical space).

kocio-pl:
As I thought - not gonna work:

2) stationery-14-4
3) stationery-14-5

so I stay with the original one.

Issue #1418, opened 2015-03-18
Add special rendering for shop=hifi


kocio-pl:
We should start rendering shop=hifi, probably with this icon from osmfr-cartocss fork:

https://github.com/cquest/osmfr-cartocss/blob/master/symbols/hifi.svg

[It was proposed in meta-issue #1402]

pmailkeey:
I don't like "hifi" as it's a bit precise - would prefer a good equivalent of "multimedia entertainment" - covering TVs, hifi, even computery games electronics / consoles.

kocio-pl:
Probably, but the tag is already used and we have [almost 6k uses](http://taginfo.openstreetmap.org/tags/?key=shop&value=hifi), so rendering of shop=hifi would be rational.

math1985:
> I don't like "hifi" as it's a bit precise - would prefer a good equivalent of "multimedia entertainment" - covering TVs, hifi, even computery games electronics / consoles.

I agree.

Typical hifi shop: http://www.bang-olufsen.com/en

Unfortunately, this is not the right place to do something about it - try the tagging mailing list.

dieterdreist:
> Am 30.04.2015 um 15:17 schrieb math1985 <notifications@github.com>:
> 
> I don't like "hifi" as it's a bit precise - would prefer a good equivalent of "multimedia entertainment" - covering TVs, hifi, even computery games electronics / consoles.
> 
> I agree.
> 


"hifi" is a different kind of shop, it doesn't include all kind of multimedia entertainment or computer games.

Cheers 
Martin

kocio-pl:
My icon proposition for shop=hifi: ![hifi-14](https://cloud.githubusercontent.com/assets/5439713/7422433/19f048cc-ef8d-11e4-9fde-ebd977133376.png)

pmailkeey:
I'll take your icon and raise you:

![image](https://cloud.githubusercontent.com/assets/11827386/7422532/8feb6500-ef85-11e4-8890-23d82222c695.png)

This icon can be used for a proper hifi shop - that has hifi TVs, DVD players and hifi computer games. Let's just call it an AV shop - that's simple and understandable.

This message, by the way, is in in HD with 5.1 surround sound.

kocio-pl:
Funny. =} 

Maybe you mean shop=electronics? ([icon for it in Osmic](https://github.com/nebulon42/osmic/blob/master/shop/electronics-14.svg))

[...]

PR #1593, opened 2015-05-27
Adding shop=hifi icon

kocio-pl:
Resolves https://github.com/gravitystorm/openstreetmap-carto/issues/1418.

hifi-z19


nebulon42:
The situation for `shop=hifi` on the wiki is a bit unclear at this moment, maybe we should defer this a bit? Iconwise it works. How would it look like with the inner circle being larger?

kocio-pl:
I don't remember any particular output from discussion on the mailing list and this icon seems to me as a no-brainer if we count recognizability. What are the issues you think should be resolved before we might start rendering it?

I will try with bigger circle.

nebulon42:
It says on the wiki "proposed to be merged with electronics".

kocio-pl:
I just asked on tagging list what to do with it.

nebulon42:
Thanks. I know I'm really picky, but now the inner circle feels too big. What about having it larger than before but not as large as in 2) leaving a bit of white space for the "ring"? I think that would be perfect.

kocio-pl:
No problem, we're just playing after all. =} 

3) hifi-14-3 

It may be really the best if this tagging will be held.

kocio-pl:
Wiki page is now clean, since nobody on the list objected, and the icon is updated to version 3), so it's ready for merging.

PR #1594, opened 2015-05-28
Add name IS NULL to placename layers

pnorman:
Only names are rendered from these layers, so there's no change in rendering.


math1985:
Should be name IS *NOT* NULL in commit message...

pnorman:
Fixed commit message, merging

Issue #1411, opened 2015-03-18
Add special rendering for shop=laundry and shop=dry_cleaning


kocio-pl:
We should start rendering shop=laundry and shop=dry_cleaning, probably with this icon from osmfr-cartocss fork:

https://github.com/cquest/osmfr-cartocss/blob/master/symbols/laundry.svg

[It was proposed in meta-issue #1402]

kocio-pl:
My sketches:
1) ![laundry-14-1](https://cloud.githubusercontent.com/assets/5439713/7784548/16390932-016a-11e5-9704-c81d16211d0b.png)
2) ![laundry-14-2](https://cloud.githubusercontent.com/assets/5439713/7784551/1edbbe40-016a-11e5-938b-7d2c1c3fb317.png)

kocio-pl:
Some more:
3) ![laundry-14-3](https://cloud.githubusercontent.com/assets/5439713/7798695/6970d140-02fd-11e5-9701-b92bb6de7b7d.png)
4) ![laundry-14-4](https://cloud.githubusercontent.com/assets/5439713/7798696/7025ed2c-02fd-11e5-8282-40adbac538fa.png)
5) ![laundry-14-5](https://cloud.githubusercontent.com/assets/5439713/7798698/7607a294-02fd-11e5-9a71-7baea01570f4.png)

I like 5) the most, then 3).

kocio-pl:
Test - for me it works good:

![laundry-z19](https://cloud.githubusercontent.com/assets/5439713/7819050/9cdda12e-03df-11e5-96e7-983027561094.png)

lest69:
It looks too much like a camera to me.

What about a shirt partially inside of a washing basin? It may be old-fashioned and doesn't accurately depict the dry-cleaning process, but it would convey the concept of the cleaning of clothing. I'm not sure if you can fit that into the available space, though.

kocio-pl:
The idea was not bad, but I think it doesn't work too good: ![laundry 16](https://cloud.githubusercontent.com/assets/5439713/8317367/28eb5e12-19fd-11e5-9e57-8729e8b7083c.png)

(BTW: the merge of PR was done just before the moment).

PR #1595, opened 2015-05-28
Adding shop=laundry and shop=dry_cleaning icon

kocio-pl:
Resolves https://github.com/gravitystorm/openstreetmap-carto/issues/1411.

laundry-z19


nebulon42:
The "eye" at the centre is a bit blurry in contrast to the rest of the icon, otherwise good. Maki has quite a similar icon, maybe inspiration could come from there too?

kocio-pl:
I would be delighted if the eye is even more blurry =} (=lighter), but I will check Maki icon of course.

nebulon42:
Alternatively, you could also add some bluriness to the top inner part by e.g. add rounded corners to the two "holes".

Issue #1592, opened 2015-05-27
Add special rendering for shop=beverages


kocio-pl:
Mentioned in https://github.com/gravitystorm/openstreetmap-carto/issues/1460

My propositions (derived from Osmic icon for alcohol shop) are:
1) ![beverages-14-1](https://cloud.githubusercontent.com/assets/5439713/7835199/358478a2-0477-11e5-9e98-2f1dbf6df8ca.png)
2) ![beverages-14-2](https://cloud.githubusercontent.com/assets/5439713/7835251/8621a226-0477-11e5-937e-88d5214fd3f2.png)
3) ![beverages-14-6](https://cloud.githubusercontent.com/assets/5439713/7835321/32bfedee-0478-11e5-8b70-06b1032b46be.png)

I guess 3) is the best.

nebulon42:
The straw or something similar might help to differentiate from alcohol shop though.

kocio-pl:
Then I would go for 2) - with higher glass.

PR #1596, opened 2015-05-28
Adding shop=beverages icon

kocio-pl:
Resolves https://github.com/gravitystorm/openstreetmap-carto/issues/1592.

beverages-z19


nebulon42:
Just more thoughts: Maybe make also the bottle different and maybe have a thicker straw?

kocio-pl:
I started with it =} : 
2) beverages-14-0 and 3) beverages-14-straw-1 and 3) was too light for me, so maybe 2) would be good.
For more clear comparison - now it is 1) beverages-14-2. What's your opinion?

nebulon42:
1) seems to be the thickest, would even thicker be possible?

kocio-pl:
Sorry, I misunderstood you of course - thicker, not thinner!... =}}} IMO that would be hard, because straw should be long and thin (and again we have no sufficient vertical space), but I will experiment with your idea. 

Or maybe lemon - however there are also alcohols you drink with a straw and/or with a lemon, so I guess glass cup alone may be as good as with these attributes. It's enough to be visually different and eye pleasing if possible.

Rovastar:
This is one of the tags that I never understood. 
I imagine most shops selling alcohol also sell some non-alcoholic drinks like Coke etc. BUt of course this is not the place to discuss this only obviously to encourage the usage....

dieterdreist:
2015-05-29 3:13 GMT+02:00 Rovastar <notifications@github.com>:

> This is one of the tags that I never understood.
> I imagine most shops selling alcohol also sell some non-alcoholic drinks
> like Coke etc. BUt of course this is not the place to discuss this only
> obviously to encourage the usage....
>

I believe the "alcohol" shop typology is something mainly found in
countries like Iran or the US, i.e. it is context dependent whether people
will use the beverages shop (access to alcohol is "easy", it can also be
found in supermarkets etc.) or the alcohol shop (alcohol is strictly
regulated and special shops exist).

kocio-pl:
4) beverages-14-7
5) beverages-14-8

For me neither the tallest bottle nor the thicker straw work at all.

PR #472, opened 2014-04-13
Shift minimal zoom levels for villages and hamlets.

jekhor:
At current setup, villages are shown at zoom levels >= 12 and hamlets at
zoom levels >= 14. In many countries the map seems 'uninhabited' with such
settings and this make it unusable for trip planning.

Shift minimal visible zoom level for village to 10 and to 12 for hamlets.

Closes bug #2057 in trac.

Tested for Belarus, Estonia, Switzerland.

Images (before and after patch) at 10 zoom in hight-density area (Switzerland):

switzerland-osm
switzerland-new

Issue #1560, opened 2015-05-20
Too many icons at z17?


nebulon42:
z17 might start getting crowded within cities. Example from @SomeoneElseOSM http://www.openstreetmap.org/#map=17/52.95354/-1.14638

I'm not sure if this can be considered a problem, but maybe a discussion is needed if some POIs should get rendered only at higher zoom levels e.g. some shops. For shops we could also switch from a two-tier to a three-tier rendering process.

See also https://github.com/gravitystorm/openstreetmap-carto/issues/1460#issuecomment-103860484, https://github.com/gravitystorm/openstreetmap-carto/issues/1460#issuecomment-103862526, https://github.com/gravitystorm/openstreetmap-carto/issues/1460#issuecomment-103862597.

joakimfors:
IMHO that doesn't look like a problem as it pretty accurately describes the street level appearance of areas crowded with shops and cafés. It also helps with navigation as you quickly realise where the busiest areas of a town/city is without obscuring the road network.

imagico:
This is a more general problem and not specific to POIs - in fact it is almost the same as #1143.  It comes from the fact that icons and labels of a certain type are generally simply drawn until the map is full - which is always a bad idea since the map needs to show other features as well.

And note the shown example is from 53 degrees latitude where the map already offers more than twice the real estate compared to around the equator.

nebulon42:
I thought #1143 results out of a tagging/classification deficiency where we cannot decide to leave out unimportant airports easily whereas we can always decide which POIs to render on a certain zoom level.

imagico:
It's the same problem, you want to display features on an early zoom level since in some areas they are rare and if they exist they are important while in areas where they are abundant you only want to display the more important ones.

The rating problem also exists for shops - when they are as dense as in the example which shops are shown and which left out is fairly arbitrary.  If you now decide shop type A is more important than shop type B and is rendered at an earlier zoom level and with higher priority at the later zoom levels this will essentially be an overall importance rating based on shop type which is not really a good importance measure.  Using an objective quantity - like length of a nearby runway in case of airports or area of the building the node is located in for shops - could be much better.

nebulon42:
True.

SK53:
[...]

PR #1597, opened 2015-05-30
Shift minimal zoom levels for villages and hamlets.

jekhor:
In continue of discussiion in the https://github.com/gravitystorm/openstreetmap-carto/pull/472

Shift minimal visible zoom level for village to 11 and to 12 for
hamlets. Set minimal text distance to 30 for villages names at zoom 11.

Switzerland, z11:
sw-z11
Switzerland, z12:
sw-z12
Switzerland, z13:
sw-z13

Belarus, z11:
bel-z11
Belarus, z12:
bel-z12
Belarus, z13:
bel-z13

### DIFF
--- a/placenames.mss
+++ b/placenames.mss
@@ -134,7 +134,7 @@
 
 #placenames-small::village {
   [place = 'village'] {
-    [zoom >= 12][zoom < 17] {
+    [zoom >= 11][zoom < 17] {
       text-name: "[name]";
       text-size: 10;
       text-fill: @placenames;
@@ -143,6 +143,7 @@
       text-halo-fill: rgba(255,255,255,0.6);
       text-wrap-width: 50;
       text-min-distance: 10;
+      [zoom <=11] {text-min-distance: 30;}
       [zoom >= 14] {
         text-fill: @placenames-light;
         text-halo-fill: white;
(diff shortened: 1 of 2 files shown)


imagico:
I don't think this is going to work well - especially shifting the threshold for hamlet (and locality/neighbourhood/isolated_dwelling/farm) from 15 to 12.  Keep in mind that these labels will prevent most other labels from showing up.

And whenever you try to show more labels than fit at the current zoom level the choice which labels are shown and which not gets hugely arbitrary unless you have a way to sort them by priority.  At z=12 you will in many areas probably not be able to show more than 10-20 percent of the hamlets.

The problem you are trying to address, i.e. that labeling is much too sparse in less densely populated areas, esp. at higher latitudes is a serious one but the solution is not to universally show more labels even in densely populated regions.  See also #1461.

Technically you will need to modify project.yaml and generate project.mml using the yaml2mml.py scripts.

kocio-pl:
I guess zoom level tuning (which zoom level should some kind of tagging start being visible) is too general tool, which is not able to embrace whole world's diversity. We should start using local density limits more, as described in https://github.com/gravitystorm/openstreetmap-carto/issues/1560#issuecomment-105205450, for all kind of points, including icons and names.

In practice that would mean we make the zoom levels suitable for low density places and fine tune the limits on the most dense places, so we have something working universally and limited (automatically!) only when there's too many items.

How about that?

BTW - the real command line for MML file generation should look like (issued from the directory with project.* files): 
./scripts/yaml2mml.py < project.yaml > project.mml

matkoniecz:
See https://travis-ci.org/gravitystorm/openstreetmap-carto/builds/64689058

matkoniecz:
Closing for now, since project.yaml and project.mml are not in sync (@imagico mentioned it on 30 May).

PR #1599, opened 2015-05-30
Remove castle_walls as per #1598.

### LINKED ISSUE #1598: Stop rendering historic=castle_walls
Simple reasons:
* There are [two](https://github.com/gravitystorm/openstreetmap-carto/blob/master/project.mml#L565) [layers](https://github.com/gravitystorm/openstreetmap-carto/blob/master/project.mml#L591) trying to render it.
* The value is not listed in [the OSM wiki](http://wiki.openstreetmap.org/wiki/Key:historic).
* There is not even a [tag page](http://wiki.openstreetmap.org/wiki/Tag:historic%3Dcastle_walls) for it.
* There are [17 ways in the whole planet](http://taginfo.openstreetmap.org/tags/?key=historic&value=castle_walls#overview) which probably can be retagged as citywalls.

StyXman:
This fixes #1598.

### DIFF
--- a/citywalls.mss
+++ b/citywalls.mss
@@ -11,14 +11,3 @@
     }
   }
 }
-
-.castlewalls {
-  [zoom >= 14] {
-    line-color: #999;
-    line-width: 1;
-  }
-  [zoom >= 16] {
-    line-color: #888;
-    line-width: 2;
-  }
-}
--- a/project.mml
+++ b/project.mml
@@ -562,58 +562,6 @@
       "advanced": {}
     },
     {
-      "name": "castlewalls",
-      "srs-name": "900913",
-      "geometry": "linestring",
-      "class": "castlewalls",
-      "id": "castlewalls",
-      "srs": "+proj=merc +a=6378137 +b=6378137 +lat_ts=0.0 +lon_0=0.0 +x_0=0.0 +y_0=0.0 +k=1.0 +units=m +nadgrids=@null +wktext +no_defs +over",
-      "Datasource": {
-        "extent": "-20037508,-20037508,20037508,20037508",
-        "table": "(SELECT\n    way\n  FROM planet_osm_line\n  WHERE historic = 'castle_walls'\n) AS castlewalls",
-        "geometry_field": "way",
-        "type": "postgis",
-        "key_field": "",
-        "dbname": "gis"
-      },
-      "extent": [
-        -180,
-        -85.05112877980659,
-        180,
-        85.05112877980659
-      ],
-      "properties": {
-        "minzoom": 14
-      },
-      "advanced": {}
-    },
-    {
-      "name": "castlewalls-poly",
-      "srs-name": "900913",
-      "geometry": "polygon",
-      "class": "castlewalls",
-      "id": "castlewalls-poly",
-      "srs": "+proj=merc +a=6378137 +b=6378137 +lat_ts=0.0 +lon_0=0.0 +x_0=0.0 +y_0=0.0 +k=1.0 +units=m +nadgrids=@null +wktext +no_defs +over",
-      "Datasource": {
-        "extent": "-20037508,-20037508,20037508,20037508",
-        "table": "(SELECT\n    way\n  FROM planet_osm_polygon\n  WHERE historic = 'castle_walls'\n) AS castlewalls_poly",
-        "geometry_field": "way",
-        "type": "postgis",
-        "key_field": "",
-        "dbname": "gis"
-      },
-      "extent": [
-        -180,
-        -85.05112877980659,
-        180,
-        85.05112877980659
-      ],
-      "properties": {
-        "minzoom": 14
-      },
-      "advanced": {}
-    },
-    {
       "name": "landuse-overlay",
       "srs-name": "900913",
       "geometry": "polygon",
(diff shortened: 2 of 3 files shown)


pnorman:
I was looking at removing the layer by combining with the citywalls layer, but this solves the layer proliferation better.

math1985:
Thanks @StyXman!

mboeringa:
The Germans (that are most active in the historic tagging arena I think with the hosting of the historical map - http://gk.historic.place/historische_objekte/translate/en/index-en.html), are pushing for **barrier=wall / wall=castle_wall** double tag, although that undoubtedly means a fair amount of castle walls in the future will be tagged with just **wall=castle_wall**...

See this (German) reference to it on the OpenStreetMap forums:
http://forum.openstreetmap.org/viewtopic.php?pid=502578#p502578

and more importantly the now listed tag on this Wiki "barrier/wall=x" page (although the tag itself doens't yet have a page):

http://wiki.openstreetmap.org/wiki/Tag:barrier%3Dwall

The German text in the forum actually says that they will document the wall=castle_wall tag in the near future more precisely, so someone is working on it.

So it is probably good to remove **historic=castle_walls**, but be aware that there is new tag that will undoubtedly be picked up by the Germans and others, and may need rendering in the future (it is currently at just 86 occurrences, but undoubtedly rising with a listing on a Wiki page: https://taginfo.openstreetmap.org/search?q=wall%3Dcastle_wall)

dieterdreist:
2015-06-01 2:47 GMT+02:00 mboeringa <notifications@github.com>:

> See this (German) reference to it on the OpenStreetMap forums:
> http://forum.openstreetmap.org/viewtopic.php?pid=502578#p502578
>
> and more importantly the now listed tag on this Wiki "barrier/wall=x" page
> (although the tag itself doens't yet have a page):
>
> http://wiki.openstreetmap.org/wiki/Tag:barrier%3Dwall
>
>
>
>

how can we decide on a rendering if we don't know what the tag is used for?
There is no definition in the wiki, and in German there is no translation
for the term "castle", or better, there are different translations
according to the kind of "castle" (e.g. Burg, Schloss, Kastell). For Italy
it is even harder to say what a "castle" is supposed to be (e.g. "borgo",
castello, villa, rocca, bicocca, fortificazione, maniero, ...)

mboeringa:
@dieterdreist, I think the problem is less big than you suggest here. 

There is no reason to differentiate between "possible" distinct "castle" types. As far as I understand, there is just a desire, with this tag, to differentiate a "castle" defensive wall from a city wall (and you well know how such defensive walls can look, irrespective if it is a burcht / slot / kasteel / fort (as in Dutch...)).

A city wall circumvents a town or city, it doesn't circumvent just one (small group of) buildings of a single historic owner, hence the tag just doesn't "fit" for castle walls...

Of course, people could just tag barrier=wall, but that doesn't feel entirely right either, as it puts no special significance to what are definitely noteworthy features.

Anyway, not before long, some of the German OSMers will probably have defined it on the Wiki...

dieterdreist:
> Am 01.06.2015 um 17:40 schrieb mboeringa <notifications@github.com>:
> 
> There is no reason to differentiate between "possible" distinct "castle" types. As far as I understand, there is just a desire, with this tag, to differentiate a "castle" defensive wall from a city wall (and you well know how such defensive walls can look, irrespective if it is a burcht / slot / kasteel / fort (as in Dutch...)).
> 
> A city wall circumvents a town or city, it doesn't circumvent just one (small group of) buildings of a single historic owner, hence the tag just doesn't "fit" for castle walls...
> 


I am not sure, city walls sometimes come nested, every big city once was small, and villages might have city walls as well? Or are these latter exactly those castle walls we are talking about here? In other words: do castle walls need a castle or is this intended as a generic tag for defensive walls that aren't city walls? Then I'd prefer a tag like defensive_walls

Also "castle walls" doesn't seem to be an expression in English, as opposed to city walls which have a defined meaning


cheers 
Martin

HolgerJeromin:
The 17 castle_walls in the database were walls of castle ruins.

mboeringa:
> Then I'd prefer a tag like defensive_walls

City walls are also defensive_walls, so this would again nullify the difference, and it is exactly the desire to differentiate, that seems the root of the new tag.

> The 17 castle_walls in the database were walls of castle ruins.

Some of the new **wall_castle** are also, but not all, some are for (relatively) intact structures". See these links:

http://www.openstreetmap.org/way/226854323
http://www.openstreetmap.org/way/36973311
http://www.openstreetmap.org/way/337682272
http://www.openstreetmap.org/way/118766759
http://www.openstreetmap.org/way/338807725#map=19/52.63042/4.75573

Of course, many historic city and castle walls have not survived the centuries unscaved, whether due to neglect and natural decay, destruction in war, or simply (partial) deliberate destruction or removal in more modern times to make place for other buildings.

Many city and castle walls are therefore heavily reconstructed or restored. They are often still notable features though, even if only sections survive, and can be major tourist attractions.

Anyway, this discussion may better be continued on the German section of the OpenStreetMap Forum:
http://forum.openstreetmap.org/viewtopic.php?pid=502578#p502578
as this seems to be where most of the decisions are being made by the Germans... The German subforum is one of the most active ones on the OpenStreetMap Forums, and maybe even more used than the OpenStreetMap (tagging) mailing lists? (can't tell, I don't visit it regularily, I usually browse the Forums).

PR #1601, opened 2015-06-02
Adding shop=cosmetics and shop=perfumery icon

kocio-pl:
While trying to design the icon for shop=chemist I created another icon, which may work for shop=cosmetics and shop=perfumery:
perfumery-z19


nebulon42:
Works, but I'm not really satisfied with the look.

kocio-pl:
I am more satisfied with it than with many more icons I have drawn, even if it was just a spin-off - I like both the visual style and recognizability of it.

Any hints what you'd like to change or which approach we may take instead?

math1985:
Thanks a lot for all changes! And sorry for slow review, I moved back to the Netherlands last month, and have been quite busy.

Issue #116, opened 2013-08-14
Render a generic icon for shop=*


dkniffin:
https://trac.openstreetmap.org/ticket/3043

I'm working on this right now. I will submitting a pull request shortly.

PR #1602, opened 2015-06-03
Adding shop=erotic icon

kocio-pl:
Variation of this icon was meant for shop=beauty, but now, when it has another icon merged, I think this one may be the best for shop=erotic:
erotic-z19


math1985:
What about a heart?

kocio-pl:
I thought this might be needed for emergency=defibrillator (heart with a flash) - 8 621 uses according to Taginfo is quite a good reason to show it. Plus I think that fat lips are more "erotic" than a heart. 

I can also think of another approach:
2) erotic-14-2
3) erotic-14-3

but these include fragile, small elements, which are hard to design properly in such a limited space, and I still like 1) more.

dieterdreist:
2015-06-03 13:21 GMT+02:00 kocio-pl <notifications@github.com>:

> I thought this might be needed for emergency=defibrillator (heart with a
> flash) - 8 621 uses according to Taginfo is quite a good reason to show it.
> Plus I think that fat lips are more "erotic" than a heart.
>
> I can also think of another approach:
> 2) [image: erotic-14-2]
> <https://cloud.githubusercontent.com/assets/5439713/7958675/cc0e1100-09f1-11e5-87ed-3bffd820a1f3.png>
> 3) [image: erotic-14-3]
> <https://cloud.githubusercontent.com/assets/5439713/7958816/427b0b12-09f3-11e5-9780-7af78ad7d30a.png>
>



I'm against the display of these shops for political reasons, because we
are not a project for adults but for kids as well. I also believe the tag
is a misnamer / euphorism, these should better be called "sex shop" or "sex
gadgets and pornography" shop. Any icon we can choose will be kind of a
statement, e.g. your proposals 2 and 3 are clearly aiming towards
heterosexual couples and will not be universally applicable to all kind of
these shops. Also the lips are referring to women which might not be
suitable for all kind of gay sex shops.

kocio-pl:
Just a note: we already display them (as a generic dot).

Regarding to political question - this may be an issue for someone (I wish more people replying, because any small discussions can be very biased!), but I don't see this as a political problem we should really deal with. There are other things people recognize as a problem - not showing the type of shop, even if we have this data and can easily do it, is really disturbing me much more than "someone somewhere might not like this kind of shops"! There's a reason people collect this data - why do they do it then and why should we, partially but deliberately (not because we have no resources as usual), hide their work?

Erotic is a broader term than sex and I think that's great the tag is named like this (not that I find "sex shop" name wrong in any way, it's just a blurry line and it's safer to use broader term - what if the shop is selling mainly velvet lingerie?). And I can find a lot of pornography in typical convenience or newsagent shop, so why not call them respectively too?...

I avoid being too direct, as you can see, so depiction shouldn't be a problem for children.

There's also non-removable practical problem with any icon: we choose something typical and - well - iconic, because we're not gonna show every object. Even the "obvious" convenience shop icon does not show anything about food or drink, which are the key objects there. But not even very narrow and easy to identify objects are safe - for example current shoes shop is misleading, because we have chosen typical men's shoes as an icon and those are frequently for women's. Icons are just are highly symbolic, convenient hint. So we have to live with this, as showing anything would be impossible otherwise (highways also are rarely colored like we render them... =} ).

And BTW - this is  sketch of a heart for AED:
defibrillator-14-2

dieterdreist:
> Am 03.06.2015 um 14:53 schrieb kocio-pl <notifications@github.com>:
> 
> what if the shop is selling mainly velvet lingerie?).


I believe they would not call themselves erotic shop (and would not be tagged like that), but rather lingerie or underwear or sth like that


> And I can find a lot of pornography in typical convenience or newsagent shop, so why not call them respectively too?...


yes, that's quite possible in some countries, but not an issue here because we don't call or tag them erotic shops.

HolgerJeromin:
I think the comments on #116 would speak to include a special icon.
I would like to have a special icon, but gender-neutral (for the gay audience). So i think a heart would be better.

math1985:
I think we should follow the wikipedia policy: we are not censored.

We map many shops that some people consider offensive or not suitable for children: not only erotic shops, but also for example gun shops, alcohol shops. If we start censoring erotic shops, why not also censor gun shops or alcohol shops?

If people take offense to them, they should lobby to get the shops removed from their neighbourhood, not lobby to remove them from the map.

kocio-pl:
@HolgerJeromin What do you think about possible clash with an icon for AED? 

I think defibrillator should be in "health" color, which is not that distant from "shop" purple, however I think it should be rendered from z>=18 (or even 19), because it's smaller than doctor's office. Do you think the different shade, minimal zoom level and a flash would be enough to separate them visually?

I also can think of small modification, which I think will do:
4) erotic-14-5

math1985:
> @HolgerJeromin What do you think about possible clash with an icon for AED? 

I think it's sufficiently different.

HolgerJeromin:
I second math opinion.

math1985:
Now I think of it, openchildmap.org would be a quite cool rendering :).

kocio-pl:
So, here we are - it works for me:
erotic-z19-heart

kocio-pl:
BTW: there are only 2 shop=sex items, so not worth adding.

And openchildmap.org could be useful if we render kid's clubs, cafes for parents with children, ice-cream shops and amenities, but also day care facilities for children (there was a discussion about it) an even orphanages - so this is not just a nice joke. =}

nebulon42:
I'm quite against the heart because it is clearly an euphemism and is "sugarcoating" the topic. But I'm not against rendering these shops nor against a special icon.

matkoniecz:
I am strongly against the heart icon as meaning is not obvious. I also share @nebulon42's concerns.

I would be also against ideas like penis-shaped icons.

kocio-pl:
OK - one developer is against exposing children for any erotic-related materials and the other one is against sugarcoating things. Hard tie. I wasn't expecting we have representation of both ends here, but that's great, because we can see more problems and try to solve them before anything is committed, not after.

I remember the same kind of discussion in polish Wikipedia years ago about "sex" related articles. The compromise was not showing photos, but rather graphics and other indirect depictions - and I see it's still holding (also on English Wikipedia). I suppose the compromise is needed also here, and I don't mean just ideological one, but also technical and cognitive.

There's an old joke about the man looking for a watchmaker, but when he finally finds a shop with a clock, it appears it's a brothel instead. The client is confused and angry: "So why did you put a clock in the window?!" - and the answer is: "Well, what I was supposed to show then?"...

And really: what are the universal symbols of sex as such? I guess fat lips are quite good (although not including all LGBT rainbow, indeed), a heart with a spice is also quite close to the subject. But should we just show sexual intercourse or anatomic details? I guess it would be not only pointlessly rude, but also naive and wrong, because human erotism is much more than anatomy and physiology (although they are a part of it). I can't also think of any erotic accessory which could be both unique and easy to show: DVDs, handcuffs, magazines, perfumes or lingerie are not unique, while typical vibrators are hard to draw/recognize (I was trying to make a proper lipstick and I still was not sure if it's one of them or the bullet maybe...). So let's look for something more general and less physical than hard sex.

I was just afraid if the heart symbol could be confused with something else and it would be hard to find much different idea for AED. But since you told that's not the issue, plain heart is acceptable for me, because love in general means also erotic fascination (sex included), and I'm not aware of any other POIs which it could be confused with. And if it's different enough from the rest (visually and metaphorically), we don't need to look any further.

No sugarcoating anything then, just hard searching for an adequate and universal symbol - business as usual. Also please remember that icons are _highly_ symbolic and if you try to compare the icons on different map styles, you would be surprised how the meaning is not hard coded in symbols (like for example a basket for any shop in humanitarian, not just a convenience).

nebulon42:
Maybe to make myself more clear, what I meant was that the heart is too "nice" for that. Even worse would be a heart for a brothel or something. Of course, there shouldn't be anything explicit too. Probably that's overreacting a bit, because an icon is a quite innocent thing. But knowing some of the debates around diversity regarding OSM I think the heart would make this style a target of that kind of debates. 

I was quite okay with the lips and I don't see any strong female or hetero focus here, but if there are concerns about that than that's fine. Currently, I have no other idea iconwise though.

math1985:
Some more ideas: http://image.shutterstock.com/display_pic_with_logo/619873/224359252/stock-vector-sex-icons-224359252.jpg

Would something with triple x work?

For me the heart is still most iconographic, but this might be culture-dependent. I don't mind that it's euphemism, we use a euphemism for toilets too. I'm fine with the kissing mouth too if there is too much resistance against the heart.

kocio-pl:
Yes, I was already searching and found something like that. They might work, of course, but we have not enough space for all of given symbols. 3x looks like this and I think it's less clear (and also culture dependent, but all the symbols are =} ): 

3) erotic-14-x2

I think condom may be interesting, as it's more sex related and has nice shape, so I might try it later.

math1985:
Agree 3) does not work.

pnorman:
At ~1.3k uses, I don't think it warrants a rendering different than the default shop rendering.

kocio-pl:
@pnorman : Sure, but what amount of uses really warrants anything? We don't even have a guidelines, not even what we will _not_ render. And I understand that, we just try the things case by case. I also understand if nobody had the time, skills to do it or we just could not reach the agreement. It happens and I take it as a fact of life - nothing to be upset or worry. But I already asked in this issue (and nobody answered):
 
> not showing the type of shop, even if we have this data and can easily do it, is really disturbing me much more than "someone somewhere might not like this kind of shops"! There's a reason people collect this data - why do they do it then and why should we, partially but deliberately (not because we have no resources as usual), hide their work?

How much better is showing just a dot of already used and approved type of shop than the shape you have a chance to recognize?

math1985:
Let's focus on objects with more than 5000 occurrences in the database (see taginfo) for now. Thanks for the contributions though!

aceman444:
http://openlovemap.de/ :)
But even JOSM has a preset and icon for shop=erotic (a red heart).

nebulon42:
Nothing against an icon for this kind of shop, but as I have said I think a heart is unsuitable for it (especially on the standard style). Regarding the map, it should have been called openf***map, that would have been more accurate. I have nothing against that kind of maps either, but would in the same vein vote against an hypothetical inclusion on the osm.org page.

Issue #1504, opened 2015-04-28
Plan database re-import


math1985:
This is a management issue for the upcoming re-import of the database. A re-import of the database gives us the opportunity to change the keys that are available to the stylesheet.

We will use the lua tag transformation option from osm2pgsql, and enable the hstore option.
# Questions to be answered
- Decide for which keys we want to change the polygon vs linestring behaviour for closed ways.
- Decide for which key/value combinations we want to change the polygon vs linestring behaviour for closed ways.
- Is there anything in particular that needs to be done to prepare for hstore?
- Do we want to move the z_order code for roads to lua?
- Are there any other changes we want to make?
# Action plan
- [ ] Decide on the questions above.
- [ ] Prepare SQL for hstore
- [ ] Implement changes in lua file.
- [ ] Contact server admins to discuss date of database reload.
- [ ] Database reload.
- [ ] Start resolving issues that need the new database scheme.
# Requested changes

The following changes have been requested:
## Keys to be added

Adding hstore will avoid the need to manually add these keys.
- emergency #318, #1160
- entrance #740, #517
- golf #661
- addr:flats #264, #605
- allotments #432
- area:highway #180
- mountain_pass #244
- ford #267
## Keys to be added

Adding hstore will avoid the need to manually add these keys.
- backcountry #1485
- memorial:type #1066
- lock_name #828
- bridge_name #828
- leaf_type #822
- int_name #803
- basin #709
- location #552, #640
- protect_class #603
- proposed #592
- information #304, #324
- network #508
- height #1017
- generator:output:electricity #1017
- lock_name #157
- bus #311, #435
- vehicle #214
- motor_vehicle #214
- motorcar #214 
## Keys to be treated as linestring
- sport https://github.com/gravitystorm/openstreetmap-carto/issues/1362#issuecomment-78860353
## Keys to be treated as polygon
- junction #1253
- boundary #468
## Key/value combinations to be treated as linestring

Pull request written, see https://github.com/openstreetmap/osm2pgsql/pull/346.
- man_made=embankment #892
- natural=cliff #268
- historic=citywalls #1362
- man_made=breakwater #1362
- man_made=groyne #1362
- natural=tree_row #1362
- waterway=derelict_canal #1362
- waterway=ditch #1362
- waterway=drain #1362
- waterway=river #1362
- waterway=stream #1362
- waterway=wadi #1362
- waterway=weir #1362
- power=line #1362
- power=minor_line #1362
[...]

PR #1605, opened 2015-06-08
update symbol of historic=memorial

nebulon42:
I know I have changed the memorial icon recently, but  this tries to fix two issues with the current symbol

* looks a lot like a tombstone
* large visual profile

by changing it into a plaque-like design and making it smaller. This shall also mitigate #1066 a bit.

preview (left=before, right=after)
memorial


kocio-pl:
I'm not sure what to do now. 

Ultimately we should use this one for a plaque probably, as it should be really much less visible than currently, but look at the squares with statues in the centre - they loose their dominant function with this new rendering. I guess we need a common symbol for memorial and plain artwork statues, because they are similar (despite being tagged differently) and maybe some other for other types. But this will be effective once we have memorial types available, so it's a question of database reload (https://github.com/gravitystorm/openstreetmap-carto/issues/1504) and we don't know when it may be and there is no sign of progress in last month and even the action plan has no checked items.

This PR can't resolve the problem of different forms of memorial forms, it just tries to change default icon and for some it will be better, while being worse for the other.

nebulon42:
Solving the memorial types problem was not intended anyway. This is meant as a pre-database reload PR. I cannot see depicting a statue as tombstone to be good either.

kocio-pl:
But there are plenty of memorial places which [looks exactly like this](https://commons.wikimedia.org/wiki/File:Flight_810_Memorial.jpg) and rightly so, because they do the same: try to commemorate someone (or some event); the only real difference is a lack of tomb. I find them more typical for such a broad category and I'm reluctant to make this change, however I will not be satisfied until we have the right solution to make [such places](http://www.openstreetmap.org/relation/4772652) more readable, for example.

dieterdreist:
2015-06-08 22:10 GMT+02:00 Michael Glanznig <notifications@github.com>:

> preview (left=before, right=after)
>
> <https://cloud.githubusercontent.com/assets/3531092/8044124/0009a6e4-0e2b-11e5-949c-d74709adcb98.png>
>


I am not sure what is the reason, but these 2 renderings are very
different, the new version has much fewer labels and different labels. You
should make exactly the same before/after renderings with the only
difference what you have changed. FWIW, I prefer the old symbol because the
new one cannot be understood.

nebulon42:
@dieterdreist The differences are because of Kosmtik switching to Mapnik 3 recently. I'm not debugging this, because this has to be done anyway later. I think you are quite capable of sorting out what is part of this change and what not, just have a look at the diffs (Files changed).

>  I prefer the old symbol because the new one cannot be understood.
I guess you will have to explain this.

@kocio-pl The image you have linked does make up just a small amount of memorials. There are lots of others. Therefore, no single symbol will be suitable for such a broad category. Nevertheless, I think the new symbol is much more suitable as for the reasons I have mentioned in the first post.

kocio-pl:
Memorials can be as small as the plaque (your new icon), similar in size and look to the tombs (your old icon) or big as a statue - or even bigger (monuments seem to be much bigger). So I prefer something in medium size (current one) to match, at least partially, both extremes (let's say with a "resemblance factor" for 3 sizes like 0,5+1+0,5=2) rather than mostly small ones (1+0,5+0=1,5). The context is clear enough for me to know it's not a tombstone.

daganzdaanda:
If you want to move away from a too literal depiction, maybe the symbol for "memorial" here https://commons.wikimedia.org/wiki/Simple_map_icons works. IIRC, that "abstract human" thing is used on some German maps, too.
I did not know that the "(tomb)stone shape" is supposedly also used in Japan: https://commons.wikimedia.org/wiki/File:Japanese_Map_symbol_(Monument).svg (Check the category, it's very interesting)

dieterdreist:
sent from a phone

> Am 14.06.2015 um 22:03 schrieb daganzdaanda <notifications@github.com>:
> 
> If you want to move away from a too literal depiction, maybe the symbol for "memorial" here https://commons.wikimedia.org/wiki/Simple_map_icons works. IIRC, that "abstract human" thing is used on some German maps, too.


yes, it is the German symbol for these, used for a very long time http://geoportal.geodaten.niedersachsen.de/legende/wms_pl25.gif

math1985:
I think the old icon is more recognizable (even though not always literally correct), so I'd prefer to keep the old one. @nebulon42 Thanks for your efforts though!

PR #1606, opened 2015-06-09
move emergency phone to SVG

nebulon42:
Implements part of #1165.

before
sos_before

after
sos_after


kocio-pl:
Less visible (that may be what we want - to see such items but not eclipse more typical elements), but for me as good as current.

daganzdaanda:
I like how clear the SOS reads. As with "phone", I think the gap isn't really necessary. 
Should we define a common colour for all "emergency" icons?

nebulon42:
> Should we define a common colour for all "emergency" icons?

That should be discussed seperately. But could be an idea to pursue.

nebulon42:
Rebased and ready for review/merge I think.

matkoniecz:
Thanks for the new icon!

PR #1607, opened 2015-06-09
move telephone icon to SVG

nebulon42:
Implements part of #1165.

before
phone_before

after
phone_after


kocio-pl:
I guess the arch should be more visible to make the shape easier to recognize, but overall impression is good.

mboeringa:
What if you drop the small gap between handle and speaker/microphone, have you tried that? I don't think it is actually necessary, and makes it slightly less recognizable (also in the "emergency" variant).

I know the gap is there in the original one, but it is so small, so as to be virtually invisible on coarse low dpi screens...

daganzdaanda:
Agree about the gap not being necessary. Otherwise, good icon.

math1985:
@nebulon42 Have you looked into dropping the gap?

nebulon42:
I like the gap better. I have also to make sure that my icons still look good at bigger sizes, even if that is not a concern for this style. It's a concern that I have to have.

Without gap: telephone-141 telephone-142

I can offer a less wide gap though and have updated the branch:
telephone

daganzdaanda:
I'm fine with the smaller gap, actually it's better than the full version.

kocio-pl:
+1

matkoniecz:
Yes, smaller gap works really well.

mboeringa:
Looks acceptable now.

nebulon42:
Rebased. Since the smaller gap is accepted this should be ready for review/merge now.

matkoniecz:
Thanks for the new icon!

PR #1611, opened 2015-06-17
Icon and YAML for amenity=motorcycle_parking

porjo:
Used `bicycle_parking` as a template:

motorcycle_parking 16

The motorcycle graphic was found here http://www.clker.com/clipart-147238.html and is released under public domain.


pnorman:
Travis is failing, see https://github.com/gravitystorm/openstreetmap-carto/blob/master/CONTRIBUTING.md#editing-layers

That icon is not suited for small sizes, containing far too much detail.

porjo:
Travis build fixed. I've modified the icon to make it look better at small sizes. New graphic is based on PNG take from [here](https://pixabay.com/en/bike-motor-cycle-transportation-39131/) (public domain)

motorcycle_parking2 16

Comparison of bicycle and motorcycle parking at 16x16:

bicycle_parking 16 motorcycle_parking2 16

HolgerJeromin:
The bike is nice, but it would be better if the P would be exactly the same with all parking features.

dieterdreist:
sent from a phone

> Am 17.06.2015 um 05:42 schrieb Ian Bishop <notifications@github.com>:
> 
> Travis build fixed. I've modified the icon to make it look better at small sizes. New graphic is based on PNG take from here (public domain)
> 
> 
> 


I prefer the first one, maybe you can add some abstraction of a driver (eg a filled circle) to the second one? This one looks like a buggy in the small version.

kocio-pl:
First one looks like a bicycle parking indeed. 

We also prefer to have all the new icons to be pixel-aligned SVGs - could you try that way (with Inkscape probably)? And we miss the motorcycle dealer shop icon (and shop=motorcycle_repair, which would be easy to crate when we have parking) - maybe you could try also make it? I could help you if you want.

HolgerJeromin:
> first one looks like a bicycle parking

This is probably because it IS bicycle parking :-) "Comparison of bicycle and motorcycle parking"

porjo:
> We also prefer to have all the new icons to be pixel-aligned SVGs - could you try that way (with Inkscape probably)?

@kocio-pl the samples provided are PNGs (only because Github doesn't support SVGs in comments), and they have the usual aliasing artifacts. If you meant something else, please provide more detail - thanks.

> I prefer the first one, maybe you can add some abstraction of a driver 

@dieterdreist I removed the driver in an effort to make it look better at very small sizes. I think also, the bicycle does not include the rider so, for consistency, the motorcycle shouldn't either.

dieterdreist:
sent from a phone

> Am 18.06.2015 um 02:59 schrieb Ian Bishop <notifications@github.com>:
> 
> I think also, the bicycle does not include the rider so, for consistency, the motorcycle shouldn't either


the consistency here should (IMHO) not be between different means of transport, but between common traffic signs and symbology. The bicycle symbol typically has no driver, while the motorcycle typically has it. You need it for easier recognizability, for a motorcycle it is expected.

kocio-pl:
I meant technical aspect like in @nebulon42 's [technical guide for drawing icons](https://github.com/nebulon42/osmic/blob/master/CONTRIBUTING.md#technical-aspects).

I would make the silhouette of the motorcycle higher (especially the front, but the seat is also too low). It could be even remake of bicycle parking to make the space between the wheels more solid but keep nice shape of the wheels (the first image had also nice wheels).

porjo:
Here are the original, original with higher front, and rider versions side-by-side:
motorcycle_parking2 16 motorcycle_parking4 16  motorcycle_parking3 16

While I agree that having a rider would be better, it doesn't look good at low resolution.

kocio-pl:
For me even 3 is acceptable and 2 is indeed better than 1. But is there possibility to make 2 even higher? 

And where can I find the SVG version to play with it myself? You can just make a fork of this style to make it easier with testing how it really looks on the map. When in testing phase, I make numbered versions of sketches in the new branch, so others can look at it in detail and try to redo as they think it may be better.

porjo:
@kocio-pl I've created a [new branch](https://github.com/porjo/openstreetmap-carto/tree/variations) containing the 2 variations in SVG format:

[motorcycle_parking_highfront.16.svg](https://github.com/porjo/openstreetmap-carto/blob/variations/symbols/motorcycle_parking_highfront.16.svg)
[motorcycle_parking_rider.16.svg](https://github.com/porjo/openstreetmap-carto/blob/variations/symbols/motorcycle_parking_rider.16.svg)

kocio-pl:
Thanks! =} 

My variations of your highfront version (links lead to SVG versions):
[2)](https://github.com/kocio-pl/openstreetmap-carto/blob/motorcycle/symbols/motorcycle_parking_highfront.16-2.svg) motorcycle_parking_highfront 16-2
[3)](https://github.com/kocio-pl/openstreetmap-carto/blob/motorcycle/symbols/motorcycle_parking_highfront.16-3.svg) motorcycle_parking_highfront 16-3
[4)](https://github.com/kocio-pl/openstreetmap-carto/blob/motorcycle/symbols/motorcycle_parking_highfront.16-4.svg) motorcycle_parking_highfront 16-4

Remarks - all models are nice when big, but as a small PNG I like 3 the most. 2 is still to low, and 4 differs from 3 just with a dot in the wheels, but since the wheels are not pixel aligned, it looks worse (whole wheels are not clear). Therefore I consider 3 as a base also for motorcycle repair shop (screwdriver instead of P) and a motorcycle dealer (just the big motorcycle shape).

mboeringa:
> Therefore I consider 3 as a base also for motorcycle repair shop (screwdriver instead of P) and a motorcycle dealer (just the big motorcycle shape).

I think this one looks nice and is understandable, I don't think it necessarily needs the figure (although the mini version @porjo showed wasn't bad considering the small size).

math1985:
@porjo Thanks for the pull request! What do you think of replacing your image by 3)?

porjo:
@math1985 yes, I'm happy to go with 3). I've updated the commit.

math1985:
@porjo There are still some things missing. It's really best too test it first in Tilemill or Kosmtik yourself, it's very difficult to write a correct pull request without testing.

Have a look at bicycle_parking rendering in the landcover.mss file and the .amenity-low-priority layer in the amenity-points.mss file.

porjo:
Sorry, but I don't have the time or motivation to go any further with this. I will close the PR but leave my branch intact. Anyone who would like to use is free to do so.

PR #1612, opened 2015-06-18
model aerialways with dashed lines instead of using line patterns

nebulon42:
By getting rid of the patterns it is easier to customise. Of course, the possibilities of `line-dasharray` are limited, but for those simple patterns it should be sufficient.

It's just an idea I had.

preview (right = after)
aerialways


kocio-pl:
I was staring at both images and couldn't spot the difference. ={ Could you explain the change in simple words?

mboeringa:
> I was staring at both images and couldn't spot the difference. ={ Could you explain the change in simple words?

I think the whole point of @nebulon42's proposal was to show you can create exactly the same symbols using an apparently easier coding / styling method...

pnorman:
you need `line-clip:false` any line with dashes both to avoid rendering problems at meta-tile boundaries

pnorman:
Just for an example, so that people understand what I mean
without line-clip
image

with line-clip
image

nebulon42:
Thanks for the hint. There is no visual difference with/without `line-clip: false` at metatile borders in my Kosmtik setup. Could this be due to changed behaviour in soon-to-be Mapnik 3?

math1985:
Cool idea, well done!

pnorman:
> Thanks for the hint. There is no visual difference with/without line-clip: false at metatile borders in my Kosmtik setup. Could this be due to changed behaviour in soon-to-be Mapnik 3?

Defaults may have changed, but also it requires some luck to hit a visually obvious case of the problem. The first few MT bounds I checked at didn't have anything I could clearly point at.

math1985:
Also, Kosmtik handles some things related to metatiles differently from the production server, as we have seen before.

gravitystorm:
> Could this be due to changed behaviour in soon-to-be Mapnik 3?

Yes, the defaults for clipping have changed from `true` to `false` in mapnik 3 -  see https://github.com/mapnik/mapnik/issues/2146

PR #1613, opened 2015-06-22
different rendering for aerialway=goods

### LINKED ISSUE #875: different rendering for aerialway=goods
Of all the types of aerialways specified on the wiki page there is only one, which normally doesn't allow passenger transport: aerialway=goods. It is currently rendered the same as aerialway=gondola and aerialway=cable_car. Should this be rendered differently to prevent confusing it as being part of a ski resort or being suitable for passenger transport?

Two examples in Austria:
http://www.openstreetmap.org/#map=13/46.8189/13.7329
In the west there is an aerialway used in mining, while in the east there is the ski resort "Bad Kleinkirchheim".

http://www.openstreetmap.org/#map=13/47.0225/12.6332
In the west there is the ski resort "Kals", while in the east there are three aerialways for transporting goods to alpine huts. These are even extending into the national park "Hohe Tauern", which would be prohibited for passenger transport.

nebulon42:
Fixes #875.

Symbols are grey instead of black and square instead of rounded. Square vs. round may not be really visible. But since #1612 it will be easier to adjust the pattern in discussion.

preview (right=new) (outdated)
aerialway=goods on the right side vs. other aerialways on the left side of each half.
aerialway_goods

### DIFF
--- a/aerialways.mss
+++ b/aerialways.mss
@@ -1,7 +1,6 @@
 #aerialways {
   [aerialway = 'cable_car'],
-  [aerialway = 'gondola'],
-  [aerialway = 'goods'] {
+  [aerialway = 'gondola'] {
     [zoom >= 12] {
       line/line-width: 1;
       line/line-join: round;
@@ -13,6 +12,21 @@
       dash/line-cap: round;
       dash/line-color: black;
       dash/line-dasharray: 0.4,13;
+      dash/line-clip: false;
+    }
+  }
+
+  [aerialway = 'goods'] {
+    [zoom >= 12] {
+      line/line-width: 1;
+      line/line-join: round;
+      line/line-cap: round;
+      line/line-color: #808080;
+
+      dash/line-width: 3.5;
+      dash/line-join: round;
+      dash/line-color: #707070;
+      dash/line-dasharray: 6,25;
       dash/line-clip: false;
     }
   }


kocio-pl:
Not looking at the code - visually it works for me.

mboeringa:
I'll be honest, I think this difference is to small and subtle to meet the goal that you actually tried to achieve here with the pull request. I really think the difference needs to be much more bold, if you seek to avoid confusion between the two types of aerialways. I don't think people will understand this as it is now.

This is also why I chose the red warning color for aerialway = goods in my ArcGIS renderer, as it stands out.

nebulon42:
I'm not convinced that a red colour would be better. I think it should be more subtle than other aerialways. Could you please provide a suitable preview (if possible of this area https://www.openstreetmap.org/#map=13/47.0312/12.6526)?

mboeringa:
> I think it should be more subtle than other aerialways.

In general, I agree that making stuff less prominent is usually the way to go, especially since there is already a lot of boldly styled data in carto. However, in this particular case regarding the already quite subtly styled aerialways, I just think it doesn't work...

> Could you please provide a suitable preview

I just recently downloaded and rendered entire Austria. I will try to output the same region and create a screenshot to show you.

mboeringa:
Ok, here is the same region rendered by my ArcGIS renderer:

arcgis_renderer_for_openstreetmap - aerialways

nebulon42:
Thanks for the example, now I understand better what you mean. I have no strong opinion in favour or against some particular way of styling, but I played a bit with red and was not really satisfied. This would be still an option, it would be good to hear other opinions on this.

I have now tried to make the style more distinct and less subtle:
aerialways

dieterdreist:
are those points actual poles/supports or is this just a symbolic linestyle? it would be nice to have them at the positions of the poles (if they are mapped)

nebulon42:
Currently none of the aerialways indicate the positions of poles and this should be kept separate from this change. Consider opening an issue for it.

mboeringa:
> I have now tried to make the style more distinct and less subtle:

If this particular symbol is not used for any other type of aerialway, I think it is OK. There is a clear distinction between the aerialways for goods and others with this symbolization. Just the "no passenger" aspect may still be left a bit in doubt, but a map legend could solve that.

nebulon42:
I found an even better example. Which one can you take? Only one out of three as it appears.
I have thought the dashes could be a bit closer together, updated.

aerialways

matkoniecz:
>Just the "no passenger" aspect may still be left a bit in doubt, but a map legend could solve that.

Maybe make it fainter? It alread appears a bit fainter, maybe making it even paler would suggest that it is less likely to be usable?

kocio-pl:
@matkoniecz +1, for me the message would be then "they are different and less important (for general public)".

mboeringa:
+1, latest rendering looks fine, a bit lighter grey might possibly help.

nebulon42:
This is the lightest possible without changing the base line colour:
aerialway

nebulon42:
@martenjj You might be interested to join the discussion.

martenjj:
Apologies for joining the discussion late;  I was intending to produce an example for #875 but haven't got around to setting up a local test server yet.  This screenshot is from OsmAnd which originally had a similar rendering with no difference between goods and passenger aerialways;  it keeps the same pattern but uses the brown track colour.  The goods aerialway is on the right.

+1 from me, though, for the lighter grey colour as the screenshot immediately above.

aerialway-goods

nebulon42:
I'm personally more in favour of a different pattern. Since nobody commented/objected for a while I guess the last preview version is now ready for review/merge.

mboeringa:
> Since nobody commented/objected for a while I guess the last preview version is now ready for review/merge.

I think this is a good conclusion, I was silently OK with the last version, which was in line with the last suggestions.

Good to see this merged.

Issue #1413, opened 2015-03-18
Add special rendering for shop=copyshop


kocio-pl:
We should start rendering shop=copyshop, probably with this icon from osmfr-cartocss fork:

https://github.com/cquest/osmfr-cartocss/blob/master/symbols/copyshop.svg

[It was proposed in meta-issue #1402]

kocio-pl:
My simple proposition (again the original is too detailed to edit it):
1) ![copyshop-14-1](https://cloud.githubusercontent.com/assets/5439713/8070874/1f2573c0-0f05-11e5-97fd-1f5dda7d4837.png)

daganzdaanda:
It does the job, I think.

matkoniecz:
Rejected in #1615 (at least for now).

PR #1615, opened 2015-06-23
Adding shop=copyshop and shop=printing icon

kocio-pl:
Resolves https://github.com/gravitystorm/openstreetmap-carto/issues/1413 .

copyshop-z19


math1985:
I like the icon, but it doesn't pass the 5000 instances threshold (yet), so for now I will close this PR.

dieterdreist:
sent from a phone

> Am 24.06.2015 um 09:37 schrieb math1985 <notifications@github.com>:
> 
> I like the icon, but it doesn't pass the 5000 instances threshold (yet), so for now I will close this PR.


5000 is a lot, there aren't even 500 country capitals in the world but we do render them (naturally). Is the 5000 a hard threshold? Is this just for shops or for all kind of features? Is there a way to apply for an exception in case a feature is deemed important but rare? Shall we look for already rendered features below the 5000 threshold so that they can be removed?

IMHO 4602 (currently used) instances of shop=copyshop are near the arbitrary 5000 line and copyshops in general are a feature of general interest and should be rendered

kocio-pl:
We have just a few such shops left (>5k uses) and I will take care of them in the short time or they are just waiting for the decision and PR, so I'd like to know what should I try in the longer run.

1. My general strategy would be to:
 * have all the icons for >10k uses (nearly all done)
 * have as many icons as it's easy doable for >1k uses (just a few done, like that one, more on the way)
 * have icons only for rare but outstanding items or just when we already have the right icon and we don't need new one <1k uses (no plan to look for them at all, purely opportunistic strategy)

2. Another key I'd like to use is to check all the shops we already show as a dot and try to make the icon for them.

3. One more thing I'd like to do is to sort the names of shops in YAML files, because it's too error prone now to edit them by hand. And maybe we could have only one name per line instead of long lines we have now?

What's your opinion about that? What do you think is worth doing first and where do you disagree with such plans? Any other hints or wishes?

math1985:
I'm quite busy now, but it's an important issue so I will come back to this.

kocio-pl:
Thanks for the info. I still have some other things to do in the meantime, so I'll just wait. =}

pnorman:
I've been planning to have a look at if there are any icons for shops in the <2k range and do a PR to remove the icon, moving them to the default icon. But this is not related to copyshop/printing, so does not belong here.

kocio-pl:
The issue ticket would be better to discuss it first. As you already know =} , I care for showing POI names and specific icons when available and since we have no policy on using Wiki or Taginfo status, I tend to check individual cases. I took the 10k/1k just out of the blue (because it was convenient to start with something we know), and 5k or 2k limits are not different.

kocio-pl:
I guess [this general discussion](https://github.com/gravitystorm/openstreetmap-carto/issues/1630) shows that we don't care so much for number of uses anyway (it can however warn us if given tagging scheme is not established, like with shop=winery) and the icons for POIs are rather welcome, so I think this PR may be safely reopened and merged.

kocio-pl:
I guess osmic version is clearly better and more subtle:
copyshop-osmic

kocio-pl:
@math1985 TagInfo says [we've just passed the 5k limit](http://taginfo.openstreetmap.org/tags/shop=copyshop). Should I rebase it or there are some other problems preventing it from being reopened?

matkoniecz:
Github is blocking reopening this PR as "this branch was force-pushed or recreated".

On topic of PR - I have mixed feeling about growing number of icons and I would probably neither merge nor propose rejecting.

Issue #1416, opened 2015-03-18
Add special rendering for shop=computer


kocio-pl:
We should start rendering shop=computer, probably with this icon from osmfr-cartocss fork:

https://github.com/cquest/osmfr-cartocss/blob/master/symbols/computer2.svg

[It was proposed in meta-issue #1402]

kocio-pl:
Just a basically corrected version:
1) ![computer-14-2](https://cloud.githubusercontent.com/assets/5439713/8069586/c4219f98-0efb-11e5-99ca-8ab23a2e3c2e.png)

math1985:
I'm very busy at the moment, sorry for my lack of comments. I will get back to the PRs as soon as possible.

kocio-pl:
Thanks for the message.

daganzdaanda:
I'd take 2) or 3), even if that leaves a big "hole" in the icon.

PR #1616, opened 2015-06-23
Adding shop=computer icon

kocio-pl:
Resolves https://github.com/gravitystorm/openstreetmap-carto/issues/1416 .

computer-z19


nebulon42:
I promised to notify you when I have created a similar icon: https://github.com/nebulon42/osmic/blob/master/shop/computer-14.svg

kocio-pl:
Thanks. =} Could you make a rendering to compare it easier? The concept is basically the same, so details matter, I guess.

nebulon42:
Here:
computer

kocio-pl:
It's less clear at this scale, but might work (probably I should copy your screen anyway). I guess more opinions would be needed to make the decision.

nebulon42:
Yes, that is to be expected because I added more detail.

kocio-pl:
I think that's rather because of the bottom part, where I have used more of perspective for the keyboard. I like your details here very much (clever!), we just have so limited space here that keeping the whole design nice at higher resolutions bites the readability at lower ones. This is a borderline case for me - I'm not sure if it's recognizable enough.

nebulon42:
For me it's still acceptable, but you are right about bigger vs smaller sizes. I have to make sure that my icons look okay at nearly all sizes, because uses might differ. They are intended for 14px, but in the font they could be used at virtually every size. Icons that are intended for osm-carto only, don't have this requirement.

Issue #1479, opened 2015-04-12
Icon for amenity=ice_cream


seebk:
amenity=ice_cream is curently not rendered in the map. Can this be changed?

kocio-pl:
We should render it, since it's very popular ([4 880 uses on taginfo](http://taginfo.openstreetmap.org/tags/amenity=ice_cream)), but we don't have the icon for it now for a start.

I wonder however if we should also render shop=ice_cream? It's far less used, but still [almost thousand](http://taginfo.openstreetmap.org/tags/shop=ice_cream) such tags exist in our database.

dieterdreist:
2015-04-13 11:19 GMT+02:00 kocio-pl <notifications@github.com>:

> We should render it, since it's very popular (4 880 uses on taginfo
> <http://taginfo.openstreetmap.org/tags/amenity=ice_cream>), but we don't
> have the icon for it now for a start.
>
> I wonder however if we should also render shop=ice_cream? It's far less
> used, but still almost thousand
> <http://taginfo.openstreetmap.org/tags/shop=ice_cream> such tags exist in
> our database.
>

similar to amenity=ice_cream there is also amenity=cafe (and maybe
amenity=fast_food) with cuisine=ice_cream (6836 uses currently:
http://taginfo.osm.org/tags/cuisine=ice_cream#overview ), but I am not sure
if we have the cuisine key currently. AFAIK this is not similar to
shop=ice_cream (which is AFAIK a take-away place and not a place with
seats).

kocio-pl:
We could reuse the same icon with different standard color: brown for cafe, violet for shop/amenity. That would be simple and coherent for me.

Most probably we don't have cuisine key yet, but if we agree on that, it would be just a matter of time (i.e. database refresh).

seebk:
I am no expert in icon design, but I wanted to give it a try:
https://cloud.sebastiankraft.net/index.php/s/9q1JQnpwzYSzCPS

If you like it feel free to use or modifiy it!

seebk:
And here already is another one
http://osm-icons.org/wiki/File:12077.svg

kocio-pl:
Could you try to rework it according to [this draft](https://github.com/nebulon42/osmic/blob/master/CONTRIBUTING.md#technical-aspects)? That would be more consistent with what we implement in OSM lately.

sommerluk:
cuisine=* is currently not in the database.

[...]

PR #1619, opened 2015-06-24
Adding amenity=ice_cream and shop=ice_cream icon

kocio-pl:
Resolves https://github.com/gravitystorm/openstreetmap-carto/issues/1479 .

Amenity alone:
ice_cream-z19-amenity

Shop alone:
ice_crem-z19-shop

Amenity+shop:
ice_cream-z19-amenity-and-shop


kocio-pl:
I want to note that I don't know which property should win over the second one if both are tagged, amenity>shop is just OK in case of ice creams (especially since people are not sure what is the difference), but that was not my choice - it's just automatic in this style that amenity takes precedence.

matkoniecz:
I like this icon, clear and obvious meaning. I have no preference whatever shop or amenity should have higher priority.

nebulon42:
Thanks for preparing the PR.

Issue #1536, opened 2015-05-07
Add special rendering for shop=variety_store


kocio-pl:
1) [My proposition](https://github.com/kocio-pl/openstreetmap-carto/blob/shop-variety_store-icon/symbols/variety_store-14-1.svg) for this kind of shop: ![variety_store-14-1](https://cloud.githubusercontent.com/assets/5439713/7507930/0f03ed02-f478-11e4-9929-92bc450fb223.png)

[Mentioned on meta-issue #1460]

HolgerJeromin:
I read it as a smart phone in landscape mode with a % on the display.

kocio-pl:
I'm quite happy with this icon, given the circumstances, but any other metaphor to test (like typical products found there) or maybe other means of showing "low price"?

math1985:
To me the proposed symbol is not recognizable (but I also don't have an alternative).

kocio-pl:
Other ideas:

2) ![variety_store-14-2](https://cloud.githubusercontent.com/assets/5439713/7518607/50ed2666-f4dc-11e4-9049-27bc74eca0ba.png)

3) ![variety_store-14-3](https://cloud.githubusercontent.com/assets/5439713/7518613/572686d0-f4dc-11e4-89d0-1b8ffe4bbed0.png)

pnorman:
I don't know that we're going to get a suitable icon for this store.

kocio-pl:
It's always the best if you can recognize the item just looking at its icon, but that's not the hard requirement. For example archaeological site icon [was misinterpreted by some of us](https://github.com/gravitystorm/openstreetmap-carto/pull/1518), but that doesn't mean it was useless to have that icon for such an important/popular item. "Suitable" for me means "good enough", not "perfect".

matkoniecz:
In that case none of proposed icon works and I doubt that that there will be good enough to replace generic shop icon.

kocio-pl:
You don't know till you try. =} 

I like the first one and, as I said in [different comment](https://github.com/gravitystorm/openstreetmap-carto/issues/1534#issuecomment-100280638) more or less, I do believe that generic dot is better just from terrible housenumbers for POIs, but much worse than visual hint, even if not recognizable just from the shape. So for me that is a fine (although not great) solution and clearly better than current state.

[...]

PR #1620, opened 2015-06-27
Adding shop=variety_store icon

kocio-pl:
Resolves https://github.com/gravitystorm/openstreetmap-carto/issues/1536.

variety_store-after-z19-digital


matkoniecz:
For me it still looks like battery and icon is not intuitive.

kocio-pl:
Luckily we have only [22 battery shops in our database](http://taginfo.openstreetmap.org/tags/shop=battery), so even if it's not clear just looking at the icon, the confusion will be very limited IMO, because probably nobody thinks about battery shops.

nebulon42:
While you did a great job in making this recognizable I'm also not convinced by it. Maybe we are trying to put too much detail into it? On the other hand, I also don't have any other design in mind.

matkoniecz:
@kocio-pl 

> Luckily we have only 22 battery shops in our database

This is not really convincing. I think that it is not enough "hard to confuse with other potential and existing icons" - goal me goal is "intuitive for a typical person".

In that case I would recommend rejecting this icon - I think that overall generic one is preferable.

kocio-pl:
Another try:
variety-bag99
variety-bagpercent

matkoniecz:
First one is clearly better than previously presented icons.

math1985:
> First one is clearly better than previously presented icons.

+1, could you update the pull request?

kocio-pl:
Sure.

Issue #1419, opened 2015-03-18
Add special rendering for shop=motorcycle


kocio-pl:
We should start rendering shop=motorcycle, probably with this icon from osmfr-cartocss fork:

https://github.com/cquest/osmfr-cartocss/blob/master/symbols/motorcycle.svg

[It was proposed in meta-issue #1402]

kocio-pl:
I guess now it makes more sense to use [3) from this PR](https://github.com/gravitystorm/openstreetmap-carto/pull/1611#issuecomment-114456754) as a base.

kocio-pl:
This is first sketch - just resized icon from motorcycle parking: 
[1)](https://github.com/kocio-pl/openstreetmap-carto/blob/icons/symbols/motorcycle.16.svg) ![motorcycle 16](https://cloud.githubusercontent.com/assets/5439713/8320112/e1254524-1a16-11e5-8392-c528a69977b0.png)

PR #1621, opened 2015-06-27
Adding shop=motorcycle icon

kocio-pl:
Resolves https://github.com/gravitystorm/openstreetmap-carto/issues/1419.

motorcycle-z19


nebulon42:
This looks a bit more like a motor scooter? I'm strongly recommending to use the same motorcycle representation for this shop and for #1611, but without putting any preference into that.

kocio-pl:
You're right and this is exactly the version 3) from #1611, which is already accepted and put into PR.

If this shape is not good enough to you, you should also try to change it in motorcycle parking symbol first, I don't feel like touching this shape because it looks OK for me.

nebulon42:
Ah, missed that.

Issue #267, opened 2013-11-22
Add rendering for ford=yes


23cpo:
The abandoned tag highway=ford is rendered while the new tag ford=yes isn't shown on the map. 

I would suggest to use the same icon for ford=yes which is already in use for highway=ford

http://wiki.openstreetmap.org/wiki/Key:ford

PR #1625, opened 2015-06-30
move ford icon to SVG

### LINKED ISSUE #118: ford icon not appropriate for non-car ways
The ford icon is a car with a wavy line beneath, as in http://www.openstreetmap.org/#map=19/51.09553/13.81686. However, in areas where no cars are allowed, eg the forest above, I don't expect to see a car icon. I propose only to use the wavy line without the car as ford icon, so that it is appropriate for both car and non-car ways.

nebulon42:
Implements part of #1165 and fixes #118.

before
ford_before

after
ford_after


kocio-pl:
Could you try also [the idea of @daganzdaanda](https://github.com/gravitystorm/openstreetmap-carto/issues/118#issuecomment-67555409)? I think it may be more clear visually. Otherwise your icon is a progress.

nebulon42:
The idea is good, but I'm avoiding perspective, so that is no option for me.

kocio-pl:
The problem with arrows is they are small and less readable when the background is noisy - like when the highway is crossing with waterway ([or with elevator over some highways...](https://github.com/gravitystorm/openstreetmap-carto/issues/720#issuecomment-116312441)).

I have also the same problem with waves, as it was in case of waterfall -  could they be bigger?

imagico:
I don't see the connection between the arrows and a ford.  Intuitively arrows indicate a measuring of some sort, like a narrow in a river, only passable by ships up to a certain size.  IMO they don't improve understanding here at all.

The waves seem too small to be well recognizable.

How about somewhat larger waves (or only a single one as before) and a diagonal cross - indicating a 'crossing water' idea and the diagonal cross being a fairly widespread symbols for crossings of all kind.

nebulon42:
I would not say that this is specific to arrows, if you compare the following preview with the one on the top you see that by adding a semi-transparent halo readability is improved:
ford_halo

It is even more improved if you use a plain background (of course), here a preview from JOSM:
ford_josm

What I try to say here is, that if we have to limit ourselves to what is really well recognizable given that we can neither use halos nor shields and have over-saturated backgrounds, then that is really limiting.

@imagico diagonal cross is not bad, I'm trying that.

nebulon42:
Might be better:
ford2

kocio-pl:
For me this version is worse, because the cross looks more like forbidden access to water than a ford. I also don't feel the waves are any bigger.

Hope I will draft my version based on @daganzdaanda proposition to test it.

imagico:
I think it's better, there is a misinterpretation risk but IMO it is not bigger than for the double arrow, especially since it is usually located where a waterway and highway cross which adds helpful context.

nebulon42:
There is the risk to interpret it like @kocio-pl, yes. Need more opinions on that. More magnitude for the waves doesn't look good and less frequency does not add to better readability. I think the slightly larger magnitude is an improvement.

matkoniecz:
For me the cross version looks like unpassable waterway. Initial version was OK. Maybe with a single bigger wave instead of two it would be better?

lest69:
Here's an alternate proposal:
What about having the waves (or a single wave), with the highway going down into them and then back up on the other side? Something like this: \_/, but with lower slope on the sides and wave(s) spanning the bottom. I'm not sure if there would be enough room to have the slopes as well as a gap in between, but it's an idea. If my description isn't clear enough, I can make a mock-up.

mboeringa:
> I think it's better, there is a misinterpretation risk but IMO it is not bigger than for the double arrow, especially since it is usually located where a waterway and highway cross which adds helpful context.

+ 1. I think the cross could even be removed, just leaving the curly water symbols. As you say, the waterway and highway provide the context.

Anyway, this feature is so uncommon in many (developed) countries, that for those people actually crossing them, they are likely to be "locals" knowing about them.

gravitystorm:
> Anyway, this feature is so uncommon in many (developed) countries, that for those people actually crossing them, they are likely to be "locals" knowing about them.

That's not a very convincing argument! The icon should be recognisable to more people than only those who have local knowledge as to what it represents.

kocio-pl:
I think it works for me as expected: 
[1)](https://github.com/kocio-pl/openstreetmap-carto/blob/icons/symbols/ford-14-3.svg) ford-14-3 ford-14-3-blue

I have technical problem here - this draft needs to be done with layers, but I don't know how to it properly, so I doubt I can make a proper SVG for automatic rendering. I made a transportation-blue version to let show how it may look like in reality, but that's all for now.

I hope I will also play with @lest69 idea.

mboeringa:
> That's not a very convincing argument! The icon should be recognisable to more people than only those who have local knowledge as to what it represents.

I think the "you can get wet feet here" aspect of the icon was quite well covered by @nebulon42... I was just making an argument for the recognizability. I don't think adding other stuff to the "waves" is increasing this aspect, looking at the proposals up to know...

kocio-pl:
I think it looks less interesting than my 1) version and I'm not sure it's recognizable by others:
[2)](https://github.com/kocio-pl/openstreetmap-carto/blob/icons/symbols/ford-14-5.svg) ford-14-5

However I started with @nebulon42 icon, which is just 12px indeed, so maybe 1) and 2) could be better if done as real 14px.

kocio-pl:
Ok, finally I managed to unlock another achievement in Inkscape and here is my 1) icon in the wild:

ford-z19-montenegro

I was not able to find the same point ([this one](http://www.openstreetmap.org/node/2490053410) is in Montenegro), however I was trying to get as much dark and busy background as possible to make comparison more fair. The code ready to be PR-ed is [here](https://github.com/gravitystorm/openstreetmap-carto/compare/master...kocio-pl:ford-icon). 

Now we just have to decide which version we want to be used as a new ford icon. I still think this one is the best - we don't have to follow no-perspective design closely (even optics icon deviate from this rule). The arrow is a close second one for me, the cross is misleading and slopes are both unrecognizable and ugly.

mboeringa:
> I still think this one is the bes

I personally had great difficulty understanding this one until you wrote about the "perspective" view of a road crossing the water. Problem is, the road gets the same blue color as the water, and hence doesn't work as a separate entity. I still think @nebulon42's original with the two way arrow and water is the best option until now, or simply showing the curly water symbol at the intersection of road and waterway without the arrows might even work. As I wrote before, the most important aspect is the "you can get wet feet here", and the curly water symbol by itself already symbolizes this.

matkoniecz:
Also for me the first icon with arrow is still the best of presented and much better than current one.

imagico:
Could someone explain to me the idea behind the double arrow?  I might be biased as an engineer here but i don't get it - to me there does not seem to be any gain in intuitive understanding compared to the waves only and plenty of room for misunderstanding.

IMO the diagonal cross, the second option by @kocio-pl (possibly with some fine tuning) and the 'waves only' concept would all be fine - not sure which is the best.  The other two less so.  The cross by the way could be drawn vertically compressed, i.e. at a different angle to avoid the 'forbidden' association.

dieterdreist:
2015-07-03 9:40 GMT+02:00 Christoph Hormann <notifications@github.com>:

> Could someone explain to me the idea behind the double arrow? I might be
> biased as an engineer here but i don't get it - to me there does not seem
> to be any gain in intuitive understanding compared to the waves only and
> plenty of room for misunderstanding.



+1, the waves alone might do it, otherwise an additional cross (slightly
improved version of the one suggested above) could also work. Arrows or
symbolic ways like those shown before are not understandable.

kocio-pl:
Even more choices... :anguished: 

I guess 'waves only' in the transportation blue color may work, but I need the ford with the river (and possibly busy background) to really test it, because we need to see how it differs from the water color. Cross with different angle (like [on this sign](https://commons.wikimedia.org/wiki/File:Znak_G-3.svg)) may work too.

@nebulon42, can you give the location of your example?

imagico:
To test on larger rivers:

http://www.openstreetmap.org/way/71838660
http://www.openstreetmap.org/way/306070956
http://www.openstreetmap.org/way/214125180

kocio-pl:
Up to this moment I was able to test most almost everything in my home city through [Metro extracts](https://mapzen.com/data/metro-extracts) ready PBF files, but the whole Earth (or even whole countries) may be too much for my computer. So I look for the instructions how should I find and import data for testing places like these.

imagico:
You can simply get a small area via overpass/xapi and add it to your database with `osm2pgsql -a`.  As long as this does not include features that are already in your database (which osm2pgsql can't handle) this works well.

dieterdreist:
2015-07-03 12:16 GMT+02:00 kocio-pl <notifications@github.com>:

> Cross with different angle (like on this sign
> <https://commons.wikimedia.org/wiki/File:Znak_G-3.svg>) may work too.
>


exactly

matkoniecz:
> how should I find and import data for testing places like these.

Downloading - the easiest is to use "Export" button on OSM website or download area with JOSM and use save as (in case of JOSM - do it before editing).

Locating - Overpass Turbo. Go to http://overpass-turbo.eu/ hit wizard button, type "highway=ford", select "build and run query". On failure to find any hits zoom out and repeat. On warning about finding too many zoom in. 

In that case it would be a good idea to remove

>   way["highway"="ford"]({{bbox}});
>   relation["highway"="ford"]({{bbox}});

from query as only nodes are currently rendered.

Doing it nth time is tedious, https://github.com/matkoniecz/CartoCSSHelper automates this process making it faster and eliminating repetition (locating place with given tags near given location, downloading place, loading into database, generating images) but it requires setup and is more complicated to use.

dieterdreist:
sent from a phone

> Am 03.07.2015 um 18:20 schrieb Mateusz Konieczny <notifications@github.com>:
> 
> Locating - Overpass Turbo. Go to http://overpass-turbo.eu/ hit wizard button, type "highway=ford", select "build and run query". On failure to find any hits zoom out and repeat. On warning about finding too many zoom in.
> 
> In that case it would be a good idea to remove
> 
> way"highway"="ford";
> relation"highway"="ford";
> 
> from query as only nodes are currently rendered.
> 


highway=ford is the old tagging style, the new one is ford=*
Transition is almost completed

http://taginfo.openstreetmap.org/keys/ford
http://taginfo.openstreetmap.org/tags/highway=ford

matkoniecz:
@dieterdreist That is covered by "Add rendering for ford=yes #267". Currently only highway=ford is rendered.

nebulon42:
Location is here: https://www.openstreetmap.org/#map=18/48.20338/16.21264
After looking at all the comments I'm keeping the arrows as I'm also not really convinced by the other versions. The one with the highway in perspective might be mistaken by some kind of board, but thanks for trying this.

I have updated the waves so that they are more visible:
ford

math1985:
> Up to this moment I was able to test almost everything in my home city through Metro extracts ready PBF files, but the whole Earth (or even whole countries) may be too much for my computer. So I look for the instructions how should I find and import data for testing places like these.

In addition in testing in real cities, I also use a file generated in JOSM for testing. I have a workflow with three tiny scripts that works quite quickly:

A script `osm_lux` for switching to the Luxembourg database:
    echo "alter database gis rename to gis_test;
    alter database gis_luxembourg rename to gis;
    \q" | psql postgres
       
A script `osm_test` for switching to the JOSM database:
     echo "alter database gis rename to gis_luxembourg;
    alter database  gis_test rename to gis;
    \q" | psql postgres

A script `osm_import_text` for re-importing the JOSM file `rendertest.osm`:
    osm_test
    sed -i s/action=\'modify\'\ //g rendertest.osm
    osm2pgsql -s -d gis -C 512 rendertest.osm


https://github.com/math1985/openstreetmap-test-data

pnorman:
If using Kosmtik you can use a [local config](https://github.com/kosmtik/kosmtik#local-config) to switch databases.

I have a file named europeconfig.json

```json
[
    {
        "where": "Layer",
        "if": {
            "Datasource.type": "postgis"
        },
        "then": {
            "Datasource.dbname": "osm2pgsql_europe"
        }
    }
]
```

When I want to use it, I just create a symbolic link ``ln -s ../europeconfig.json localconfig.json``

yohanboniface:
> When I want to use it, I just create a symbolic link ln -s ../europeconfig.json localconfig.json

You can also use the `--localconfig` option at runtime: `kosmtik serve project.yml --localconfig europeconfig.json`

And this is an example of file a bit more dynamic: https://gist.github.com/yohanboniface/dc95cf060e709c83fe67#file-localconfig-js (I just change the "country" value when I want to switch country of test; indeed, my db are named by the country name)

kocio-pl:
Thank you all for the tips!

HolgerJeromin:
Perhaps this tips should be written in a wiki page. On github or osm.

matkoniecz:
I think that the version with arrow is better than a current icon and solves #118.

There is no obvious icon for ford but from context it is quite clear what is represented (and certainly it is better than a current icon that was highly confusing for for non-car ways).

kocio-pl:
+1 - I also think the version with arrow is at least better than what we have now.

imagico:
If there is no consensus on which of the four versions suggested so far is best the neutral 'waves only' version might be the best approach.

matkoniecz:
> I don't see the connection between the arrows and a ford. Intuitively arrows indicate a measuring of some sort, like a narrow in a river, only passable by ships up to a certain size. IMO they don't improve understanding here at all.

> Could someone explain to me the idea behind the double arrow? I might be biased as an engineer here but i don't get it - to me there does not seem to be any gain in intuitive understanding compared to the waves only and plenty of room for misunderstanding.

Arrows also are symbol of movement, so hopefully it would be understood as "movement across water".

> neutral 'waves only' version might be the best approach.

Just like current version ( https://github.com/gravitystorm/openstreetmap-carto/pull/1625#issuecomment-118623567 ) but without arrow?

imagico:
> Arrows also are symbol of movement, so hopefully it would be understood as "movement across water".

But movement is more often illustrated by two separate arrows in opposite direction, see

https://de.wikipedia.org/wiki/Datei:Zusatzzeichen_1000-32.svg

The double arrow is frequently used as a symbol for measurements, see

https://de.wikipedia.org/wiki/Datei:Zeichen_266.svg

and this connection is also emphasized by the object displayed together with it matching the arrow length.  Maybe two smaller arrows and a single wave symbol would work.

> Just like current version ( #1625 (comment) ) but without arrow?

Yup.

nebulon42:
> But movement is more often illustrated by two separate arrows in opposite direction, see https://de.wikipedia.org/wiki/Datei:Zusatzzeichen_1000-32.svg

Ah, another hint for improvement. Might be worth trying.

nebulon42:
ford-16 ford-16a ford-16b

kocio-pl:
I have tried making new version of elevator icon based on the concept mentioned here, but only [thick arrow works well enough](https://github.com/gravitystorm/openstreetmap-carto/pull/1629#issuecomment-124623306), so I guess 1) is still the best option. You can render them on the map to be sure, however.

matkoniecz:
@nebulon42 

I like both first and third, without preference for either one. Both are for me much better than second one.

imagico:
Both two arrow versions look good i think.

math1985:
> I like both first and third, without preference for either one. Both are for me much better than the second one.

+1. If you update the PR I can merge this.

kocio-pl:
I prefer to see the real life test before this is merged, because in such cases the details really matter (that's how it was with [elevator arrows](https://github.com/gravitystorm/openstreetmap-carto/pull/1629) for example).

kocio-pl:
There's a [discussion about adding ferry terminal icon](https://github.com/gravitystorm/openstreetmap-carto/pull/1673) and there is general feeling the ship shape is not easy to recognize, so I think one of these icons may be of use. What do you think about it?

matkoniecz:
It would make situation quite confusing.

kocio-pl:
So the question is: if we want to distinguish those icons, in what way should they be different? I think this is perfect moment to think about them both at he same time to not confuse people and to know how we want to depict those two "way across the water" items.

nebulon42:
After due consideration I'm tending towards keeping the double-sided arrow, since it is more distinguishable as also noted by @kocio-pl. I do not share the concerns of @imagico re confusion arising out of the double-sided arrow symbolizing measuring, but I see the possible association.

@math1985 Since you seconded @matkoniecz statement re first or third icon, but said I should update the PR: did you mean you prefer Nr. 3? Because the PR already contains Nr. 1 and could be merged as it is.

matkoniecz:
I will test soon current version of this PR - given that there is no consensus what kind of arrow is preferable and it is clearly better than what we have now.

@nebulon42 Sorry for the long waiting time.

DaveF63:
I'm struggling to understand what relevance arrows have. Nothing needs sizing or measuring. To be clear on its meaning it should just indicate water (waves) crossing a road (lines). 

[This] (https://github.com/gravitystorm/openstreetmap-carto/issues/118#issuecomment-67555409) by @daganzdaanda (with a bit of tidying of the waves) is the most intuitive idea I've seen so far.

Also, as pointed out by @dieterdreist ford=yes is the current & by far the more popular tag. Is there a reason why it's not being rendered?

polarbearing:
> ford=yes is the current & by far the more popular tag. Is there a reason why it's not being rendered

I would assume it is not in the database and needs to wait for the 3.x reload.

kocio-pl:
I have tried the implementation of this idea [earlier in this thread](https://github.com/gravitystorm/openstreetmap-carto/pull/1625#issuecomment-118013806), but it was not choosen.

Tagging ford=yes probably means hstore, which is [not high on priority list](https://github.com/gravitystorm/openstreetmap-carto/issues/1975#issuecomment-157134906) of @pnorman, so we don't have the idea when this would be possible.

DaveF63:
>Tagging ford=yes probably means hstore
I'm still struggling to understand I'm afraid. What's the different between ford=yes & the numerous POIs that have had icons added in the recent carto updates & why is it only this combination that "means hstore"?

matkoniecz:
For example key amenity is loaded in the database used to render this style on the
OSM website, key ford is not (keys that are available for Default style are listed at https://github.com/gravitystorm/openstreetmap-carto/blob/master/openstreetmap-carto.style).

PR #1626, opened 2015-07-01
SQL tweaks to improve Mapnik 3 performance for admin areas

pnorman:
Admin areas are labeled on the edges, so those where the bbox lies completely inside can be discarded in SQL

This helps with https://github.com/mapnik/mapnik/issues/2937 and generally slow performance with offsetting of polygons in the latest RC

https://github.com/mapnik/mapnik/issues/2942 prevents turning on clip for now, which would offer gains for nature reserve fill/edge


math1985:
(@pnorman To avoid confusion please add full URLs rather than bug numbers when referring to issues from different projects.)

pnorman:
> @pnorman To avoid confusion please add full URLs rather than bug numbers when referring to issues from different projects

They're not bare numbers - github autolinks them to the Mapnik PPA

math1985:
> They're not bare numbers - github autolinks them to the Mapnik repo

Only after I edited your post :)

PR #1629, opened 2015-07-02
Adding highway=elevator icon (only for public elevators)

kocio-pl:
Resolves https://github.com/gravitystorm/openstreetmap-carto/issues/720.

elevator-12-z19
elevator12-z19-tarasy

It also fixes double bus_stop appearance in project.yaml.


nebulon42:
I think it is a bit too lightweight. Maybe try to make that elevator box solid and subtract the arrows from that shape (like the cross in pharmacy).

kocio-pl:
Sounds interesting - I will try this idea.

kocio-pl:
I doubt it's any better:
elevator-z19-solid-gdanski
elevator-z19-centralny

ximex:
the first version looks better but is very hard to see in pic1. Sooo i think version 2 is better...

kocio-pl:
I think 3) could be sane compromise, because 2) is harder to recognize for me and this one is visible even in such a cluttered place:
elevator-11-gdanski
elevator-11-centralny

math1985:
The proposed version looks fine to me. If anything can be improved, feel free to write a next PR.

PR #1632, opened 2015-07-05
Adding alcohol icon for shop=wine and shop=winery

kocio-pl:
Makes a lot of sense to show shop=wine and shop=winery not as a general dot when we already have the specific icon for shop=alcohol.


aceman444:
Shop=winery is an undocumented tag with unknown meaning (as what the difference to shop=wine should be). You probably want craft=winery. But then I think there is no rendering for craft=* yet.

kocio-pl:
We already display them, but with a general dot. Either they are flawed (and we remove one or both of them) or they're OK (and we just use this fix to display them properly).

aceman444:
I'm all for keeping shop=wine and showing the icon as shop=alcohol. I just think shop=winery should be removed from the patch as not to promote undocumented tags by rendering them.

pnorman:
I don't see documentation as a barrier to rendering shop=winery, but I'd rather not render it to avoid encouraging tag fragmentation.

kocio-pl:
OK, so should I left the icon for shop=wine and remove shop=winery completely from rendering in this style?

dieterdreist:
sent from a phone

> Am 05.07.2015 um 23:15 schrieb Paul Norman <notifications@github.com>:
> 
> I don't see documentation as a barrier to rendering shop=winery, but I'd rather not render it to avoid encouraging tag fragmentation.


+1

PR #1635, opened 2015-07-05
stop displaying really small zoos and theme parks

matkoniecz:
xxx

selection_003

Unfortunately I was unable to eliminate weird effect for multypolygon observed in London - only one section receives full border. See http://www.openstreetmap.org/?mlat=51.5391&mlon=-0.1566#map=12/51.5391/-0.1566

selection_002

https://cloud.githubusercontent.com/assets/899988/8512754/97f0c164-2351-11e5-9394-ab665e94e056.png
https://cloud.githubusercontent.com/assets/899988/8512753/97e99128-2351-11e5-82d9-37aec99d8837.png
https://cloud.githubusercontent.com/assets/899988/8512755/97f18c0c-2351-11e5-89b9-2e9d0099ab8c.png
https://cloud.githubusercontent.com/assets/899988/8512750/7bf017f8-2351-11e5-9e42-f8b374387fd8.png
https://cloud.githubusercontent.com/assets/899988/8512751/7bf055ec-2351-11e5-8882-eaed417faa08.png

EDIT: reported also on PL forum - see http://forum.openstreetmap.org/viewtopic.php?pid=514230#p514230

### DIFF
--- a/landcover.mss
+++ b/landcover.mss
@@ -624,20 +624,23 @@
 }
 
 #tourism-boundary {
-  [tourism = 'zoo'][zoom >= 10],
-  [tourism = 'theme_park'][zoom >= 10] {
+  [tourism = 'zoo'][zoom >= 10][way_pixels >= 20],
+  [tourism = 'theme_park'][zoom >= 10][way_pixels >= 20] {
     a/line-width: 1;
     a/line-offset: -0.5;
     a/line-color: @tourism;
     a/line-opacity: 0.5;
     a/line-join: round;
     a/line-cap: round;
-    b/line-width: 4;
-    b/line-offset: -2;
-    b/line-color: @tourism;
-    b/line-opacity: 0.3;
-    b/line-join: round;
-    b/line-cap: round;
+    [zoom >= 17],
+    [way_pixels >= 60] {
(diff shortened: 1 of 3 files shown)


kocio-pl:
+1

imagico:
Even the larger zoos seem very strong at the low zooms:

http://www.openstreetmap.org/#map=10/54.0586/10.2626

IMO these could well start at a later zoom level (like 12).  The nature reserve boundaries can also be counterproductive for clarity at the lower zooms (size threshold here is 100 pixels).  maybe unifying both with a size threshold of ~1000 pixel at z<12 and 100 pixel above would work.

jojo4u:
The example from imagico indeed looks like a scar on zoom 10 and 11.

math1985:
My bad, thanks for fixing.

@imagico @matkoniecz Feel free to open a new issue or PR if more changes are needed.

matkoniecz:
@math1985 It is partially my fault given that it was my idea to start rendering so early.

kocio-pl:
My feeling is that zoom range adjusting is for general rendering, while way_area is for fine tuning it - they are just different tools, so no need to look who was wrong. =}

PR #1638, opened 2015-07-07
rework display of railway=tram

### LINKED ISSUE #1637: railway=tram is visible from z8, tram is too visible at z13, z15, z16
During reworking road style I decided to try changing closely related tram style as on z13, z15, z16 it is really strong.

During more close checking of rendering I discovered that tram tracks are rendered also on z10, z11 and z12 what was quite surprising as I never noticed it.

On reading code I discovered that railway=tram is rendered from z8, what makes no sense. Especially with current style as even during looking for tram lines at known positions I failed to notice it.

Overall this contributes to really bad rendering of cities around z10. 

Even in Toronto, city with "the largest streetcar system in the Americas"[1] and quite far on the north (relevant due to distortion of scale) it makes sense to render trams up to z11 - and probably z12 would be more than enough. See http://overpass-turbo.eu/s/ai3 (I linked overpass turbo, as in current rendering tram lines are not possible to find - despite rendering them).

Other example, Vienna - "With 173.4 km of track, Vienna's network is one of the largest in the world."[2] z11 is the lowest zlwevel where rendering tram makes sense ( http://overpass-turbo.eu/s/ai4 - note, query returns large amount of data).

In general it seems that even with large networks rendering tram up to z12 makes sense, z11 may be justified but it is not the best idea but later it is pointless.

My original reason to start tweaking trams is not so clearly justified - but I am convinced that at z13, z15, z16 tram lines are too visible. See for example

selection_001

I propose to make tram lines visible at z12 (darker colour), slightly reduce widths at z13, z15, z16 and remove it completely from zlevels 8 to 11.

I think that it makes sense to make PR for this independent from the giant "rework entire road styling" as merge conflicts are highly unlikely and current tram line rendering is causing the same problems with current road style, proposed new road style and its variants.

The current code:

```
    [feature = 'railway_tram'] {
      [zoom >= 8] {
        line-color: #ccc;
        [zoom >= 10] { line-color: #aaa; }
        [zoom >= 13] { line-color: #444; }
        line-width: 1;
        [zoom >= 15] {
          line-width: 2;
        }
        .tunnels-fill {
          line-dasharray: 5,3;
        }
      }
    }
```

[1] https://en.wikipedia.org/wiki/Tram_and_light_rail_transit_systems
[2] https://en.wikipedia.org/wiki/Trams_in_Europe

matkoniecz:
dropped invisible clutter from z8 to z11
visible railway=tram on z12
subtler at z13, z15, z16

fixes #1637

https://cloud.githubusercontent.com/assets/899988/8548012/b85467c2-24c2-11e5-92d5-2890be60ade0.png

### DIFF
--- a/roads.mss
+++ b/roads.mss
@@ -1572,10 +1572,10 @@ residential is rendered from z10 and is not included in osm_planet_roads. */
 
     [feature = 'railway_tram'] {
       [zoom >= 12] {
-        line-color: #444;
-        line-width: 0.6;
+        line-color: #6E6E6E;
+        line-width: 0.75;
         [zoom >= 13] {
-          line-width: 0.75;
+          line-color: #444;
         }
         [zoom >= 14] {
           line-width: 1;
--- a/roads.mss
+++ b/roads.mss
@@ -1571,12 +1571,16 @@ residential is rendered from z10 and is not included in osm_planet_roads. */
     }
 
     [feature = 'railway_tram'] {
-      [zoom >= 8] {
-        line-color: #ccc;
-        [zoom >= 10] { line-color: #aaa; }
-        [zoom >= 13] { line-color: #444; }
-        line-width: 1;
+      [zoom >= 12] {
+        line-color: #444;
+        line-width: 0.75;
+        [zoom >= 14] {
+          line-width: 1;
+        }
         [zoom >= 15] {
+          line-width: 1.5;
+        }
+        [zoom >= 17] {
           line-width: 2;
         }
         .tunnels-fill {
--- a/roads.mss
+++ b/roads.mss
@@ -1573,7 +1573,7 @@ residential is rendered from z10 and is not included in osm_planet_roads. */
     [feature = 'railway_tram'] {
       [zoom >= 12] {
         line-color: #444;
-        line-width: 0.4;
+        line-width: 0.55;
         [zoom >= 13] {
           line-width: 0.75;
         }
--- a/roads.mss
+++ b/roads.mss
@@ -1573,7 +1573,7 @@ residential is rendered from z10 and is not included in osm_planet_roads. */
     [feature = 'railway_tram'] {
       [zoom >= 12] {
         line-color: #444;
-        line-width: 0.55;
+        line-width: 0.6;
         [zoom >= 13] {
           line-width: 0.75;
         }
--- a/roads.mss
+++ b/roads.mss
@@ -1573,7 +1573,10 @@ residential is rendered from z10 and is not included in osm_planet_roads. */
     [feature = 'railway_tram'] {
       [zoom >= 12] {
         line-color: #444;
-        line-width: 0.75;
+        line-width: 0.4;
+        [zoom >= 13] {
+          line-width: 0.75;
+        }
         [zoom >= 14] {
           line-width: 1;
         }


imagico:
In general looks good, the effect at z=10-12 is difficult to judge since the roads are such a mess - with a cleaner road display it might make sense to show trams at these zooms.  I am not too sure about the dark rendering at z=12 - this seems very disturbing and affects readability of labels.  I think keeping them in railway-gray at this scale looks better.

matkoniecz:
I am not aware about any tram/light rail system where display at z10 would make sense.

> I am not too sure about the dark rendering at z=12 - this seems very disturbing and affects readability of labels. I think keeping them in railway-gray at this scale looks better.

With current rendering tram lines are nearly invisible. Maybe it would be worth trying something between current railway gray and dark gray used for higher zooms.

imagico:
In 'z=12 - before' i can well see the tram lines.  It is not obvious they are not roads but this is more a fault of the road display than of the tram color.  A compromise might work but it would also mean yet another color.

Rovastar:
All for a colour change. It doesn't make sense to have this dark black for trams. I expect legacy styling.

matkoniecz:
> It doesn't make sense to have this dark black for trams.

I am not planning on proposing something that is not "dark lines". Or at least I have no good idea and I am happy with current idea for trams (but I see problems with execution of that idea).

@imagico I will make a next version, are you interested in giving a test location? (Preferably - city with available pbf extract, for example from https://mapzen.com/data/metro-extracts )

daganzdaanda:
Could you show Vienna or Toronto at z12, to check if the dark lines are a problem? 

> In 'z=12 - before' i can well see the tram lines. 

Me too, but only now since I know where to look. I hadn't noticed the trams on z12 in cities I know before! And I think at that scale, it makes good sense to really see the trams. But maybe a grey halfway between black and the railway-grey would do the job as well without so much contrast.

imagico:
If you want to test different scales due to different latitudes Helsinki and Medellin would be the most extreme probably - Medellin is only very small though.  Prague also has a very extensive and complex structure network that could make a good test case.

matkoniecz:
> Me too, but only now since I know where to look. I hadn't noticed the trams on z12 in cities I know before! 

Yes, I am not claiming that it is impossible - but given that somebody editing OSM and contributing to this style failed to notice these tram lines in its own city is hinting that rendering is not working well. I am able to find traces on z11 and z12 but only after careful looking. Trying to find anything on z8 through z10 is a joke.

matkoniecz:
I tried new subtler z12 - and it seems that I went to far in the other direction.

tram infestation - krakow master-_tram 50 06864 19 94569 450px master - tram
tram infestation - medellin master-_tram 6 24238 -75 55012 450px master - tram

matkoniecz:
The main problem is that in some places railway=tram is mapped twice, each track once. In other places either both tracks are mapped as one or there is really only one track. Anyway, now I have something that should be quite good.

tram infestation - medellin master-_tram 6 24238 -75 55012 450px master - tram
tram infestation - krakow master-_tram 50 06864 19 94569 450px master - tram

pnorman:
Can you push what you've got so we can see?

matkoniecz:
Done, needs rebasing - but for now I prefer to keep log of what was tried and failed.

pnorman:
I'm okay with it unsquished.

PR #1647, opened 2015-07-10
start displaying minor rail from z13, fixes #1645

### LINKED ISSUE #1645: stop rendering railway=rail with service=spur/siding/yard on z11 and z12
service tag gives information that these tracks are relatively unimportant. Currently, rail sidings/yards are displayed as gray blobs on z11/z12 as these tracks merge together in rendering.

After removal of rendering function of area will be displayed by showing landuse.

Removal of excess tracks would also make possible to make rendering of railway tracks distinguishable from roads on lower zoom levels by making railways darker (something similar to #1637).

See for example http://www.openstreetmap.org/?mlat=55.7899&mlon=37.6495#map=11/55.7899/37.6495

selection_001

matkoniecz:
small changes:

https://cloud.githubusercontent.com/assets/899988/8616741/f99326a0-26fd-11e5-9f0e-cd2c1a939a74.png
https://cloud.githubusercontent.com/assets/899988/8616743/f994ba6a-26fd-11e5-8698-77c3bbe7c85b.png
https://cloud.githubusercontent.com/assets/899988/8616742/f9944738-26fd-11e5-8b82-b2a14779a233.png
https://cloud.githubusercontent.com/assets/899988/8616745/f998f2f6-26fd-11e5-9c05-54c9c2d5cd73.png
https://cloud.githubusercontent.com/assets/899988/8616922/5d0a2494-26ff-11e5-9558-072ba1bae43a.png

big changes in rendering:

https://cloud.githubusercontent.com/assets/899988/8604275/ddaa6460-267d-11e5-8899-9e5f76b1b435.png
https://cloud.githubusercontent.com/assets/899988/8616915/5143aa5e-26ff-11e5-86f4-107fd33e7fb6.png
https://cloud.githubusercontent.com/assets/899988/8616746/f99ee314-26fd-11e5-9207-863603b6b60d.png

@kocio-pl 

> I was not convinced after just reading the description, but images show me it really works.

Yet another good reason to attach before/after images :)

@dieterdreist 

> I like the current rendering, it gives a good sense how much space is taken up by railway infrastructure, and where/which shape/direction. The color makes it intuitive to understand.

"how much space" is also visible thanks to landuse. Yes, industrial and railway is the same but z11 and z12 is extremely busy and something should be removed from these zoom levels. Shape/direction is for me not really clear as rails merge in one blob. I think that it is rather improved when only main rails are as it makes them distinguishable from other features.

### DIFF
--- a/roads.mss
+++ b/roads.mss
@@ -1472,7 +1472,7 @@ residential is rendered from z10 and is not included in osm_planet_roads. */
     }
 
     [feature = 'railway_rail'][zoom >= 7],
-    [feature = 'railway_INT-spur-siding-yard'][zoom >= 11] {
+    [feature = 'railway_INT-spur-siding-yard'][zoom >= 13] {
       [zoom < 13] {
         line-color: #aaa;
         [feature = 'railway_rail'] {


dieterdreist:
sent from a phone

> Am 10.07.2015 um 12:30 schrieb Mateusz Konieczny <notifications@github.com>:
> 
> "how much space" is also visible thanks to landuse. Yes, industrial and railway is the same but z11 and z12 is extremely busy


IMHO it gets busier now, because what was one big blob becomes a skeleton with a different background colour (introducing ambiguity with respect to industrial areas which also happen to have rail connectivity sometimes).



> and something should be removed from these zoom levels. Shape/direction is for me not really clear as rails merge in one blob.


Shape and direction seem pretty clear in these examples, but also mostly remain clear in the new proposal, I agree.

HolgerJeromin:
I am unsure of this change. The area is much cleaner, but you could think this area is unmapped and needs work until you zoom in.

HolgerJeromin:
Is rendering them lighter, as in private parking, an option?

matkoniecz:
> Is rendering them lighter, as in private parking, an option?

It still would be gray blob. At least currently it is a dark gray blob making it a bit distinct from roads.

math1985:
We lose some information with this on the mid zoomlevels, but I agree the mid zoomlevels are so crowded that the simplification outweighs the negative consequences.

HolgerJeromin:
But the main railway would be visible beneath the gray blob. 
But I think not rendering something works well for small roads. We will get used to zoom in to find empty spaces to map.

PR #1650, opened 2015-07-11
Adding amenity=motorcycle_parking icon

kocio-pl:
Taken over after https://github.com/gravitystorm/openstreetmap-carto/pull/1611.

The same shape of the motorcycle as in the shop=motorcycle:
motorcycle_parking-z19-warsaw

I would like to test 2 more things to be sure everything is OK:
- parking area with name ([named nodes](http://www.openstreetmap.org/node/841558023) don't show the name - maybe a subject for separate PR?)
- parking with access tag other than "yes"

but I was not able to find it in the wild (Overpass Turbo was trying too long and after all gave me some errors). 

So I have tried to make it up using [math1985/openstreetmap-test-data](https://github.com/math1985/openstreetmap-test-data) and failed initially - saving changes with JOSM gave me [node_changed_mark failed ERROR](http://forum.openstreetmap.org/viewtopic.php?pid=83475), but after careful manually removing "action='modify'" tags from the resulting *.osm file it worked:

motorcycle_parking-testdata


math1985:
> So I have tried to make it up using math1985/openstreetmap-test-data and failed initially - saving changes with JOSM gave me node_changed_mark failed ERROR, but after careful manually removing "action='modify'" tags from the resulting *.osm file it worked:

That's why there is a line `sed -i s/action=\'modify\'\ //g rendertest.osm` in https://github.com/gravitystorm/openstreetmap-carto/pull/1625#issuecomment-120532198 :)

kocio-pl:
I wanted to look closer to learn how to switch database content, but for now I just mentally bookmarked it and didn't see that you have included there also this trick. =}

ximex:
:+1:

kocio-pl:
I think this is non-controversial commit to pull in, but please let me know if it's to be merged any time soon. The problem is that I can update the code to remove conflicts, but it's as hard as making it from the beginning, so I am not able to keep updating my PRs just in case.

math1985:
The reason I didnt merge it yet is that I noticed that there are quite a few motorcycle parkings tagged, and I want to be sure this won't flood the map.

Don't worry about the merge conflicts, I can resolve them once merging them.

matkoniecz:
> this won't flood the map

It is quite unlikely - it is using the same rules as other parkings that were both tested and seem to work fine.

kocio-pl:
The numbers are like:
- (car) parking - 1 758 628
- bicycle_parking - 127 417
- motorcycle_parking - 2 823

so I wouldn't be afraid of flooding anything as we have first two types already. =}

I think it makes sense to show all related elements of infrastructure (shops + parkings + service points for given type of transport), however motorcycle_repair usage is so low that I'm not sure yet it's really established scheme for motorcycle servicing, but once I make sure, I will try to show it also.

math1985:
> It is quite unlikely - it is using the same rules as other parkings that were both tested and seem to work fine.

However, there are much more cyclists than motorcyclists, so it is much more likely that someone uses the map to look for a bike park rather than a motorcycling parking. Moreover, many bike parkings accomodate many bikes, while the motorbike parkings that I know of typically only have one spot.

I know I really should have discussed the desirability of implementing this first, before asking people to spend time on fixing the technical issues for rendering this - sorry about that!

kocio-pl:
No problem! :smiley: 

I still hold my position that once it's rather clear that we don't want to rely on numbers too much (they are just a proof that given scheme is established), it's better to think about "clusters" than just individual POIs. For example when we have a car shop, it makes sense to show also car repair, car washing, shop with tyres and with other spare parts, because they make an ecosystem, even if some of them are very popular and the others are rare. Motorcycles have less features probably, but motorcycle parking looks like sane addition for such a shop.

dieterdreist:
sent from a phone

> Am 29.07.2015 um 00:41 schrieb math1985 <notifications@github.com>:
> 
> However, there are much more cyclists than motorcyclists


around here there are much more motorcyclists than cyclists, I think it depends on the area...

PR #1655, opened 2015-07-14
meadow/grassland and orchard/vineyard color unification

imagico:
Attempting to reduce clutter and variety of green tones:

* unify landuse=meadow and natural=grassland (already hinted in #1242)
* unify base color of landuse=orchard and landuse=vineyard (slightly stronger than current vineyard to improve contrast to scrub)
* new fine grained pattern for landuse=vineyard, this is based on the standard pattern in German and Swiss maps - not sure how recognizable it is in other cultures.  But it seems a huge improvement compared to the very coarse previous pattern which works particularly badly on fine grained vineyard parcels.
* patterns for orchard/vineyard start at z=14 (vineyard was starting at z=13, orchard was always rendered with pattern).
* ~~move landuse=conservation to nature reserve boundaries (long overdue IMO but not really tested on real world objects).~~

Examples (from around [here](https://www.openstreetmap.org/#map=13/48.0979/7.6791)):

z=17
z=16
z=15
z=14
z=13
z=12


kocio-pl:
Very good idea.

matkoniecz:
I think it would be better as separate PRs, at least  "move landuse=conservation to nature reserve boundaries" should be separate. Maybe also "meadow/grassland" and "orchard/vineyard color unification" would be better as separate.

It simplifies testing, discussion and code review.

matkoniecz:
> new fine grained pattern for landuse=vineyard, this is based on the standard pattern in German and Swiss maps - not sure how recognizable it is in other cultures

Completely unrecognisable for me. There is no standard pattern for vineyards for maps made in Poland as this feature is extremely rare here.

matkoniecz:
@imagico - is this test place is maybe an area close to you where you may perform local survey? In that case - can you look at https://www.openstreetmap.org/note/396185 https://www.openstreetmap.org/note/396187 https://www.openstreetmap.org/note/396188 https://www.openstreetmap.org/note/396189

kocio-pl:
@matkoniecz BTW: there is a [lengthy discussion going on tagging list](https://lists.openstreetmap.org/pipermail/tagging/2015-July/025565.html) about rural highways - a proposition was given to tag them as highway=service+service=rural when they are not really highway=track. I don't know however if this is the case.

imagico:
Actually i did this together to be able to test how these colors fit together.  Changing colors one at a time is difficult since there are always interdependencies.  You'd start changing the orchard color and then notice you'd better change it again to better accommodate the change in grass color.

I can take the change of landuse=conservation out here if there are doubts about it.  Just wanted to make sure that the color change is not meant to apply to landuse=conservation.

Regarding the pattern - this type is also used on [french maps](http://geoportail.fr/url/7Fccn7), on Italian maps i have seen a somewhat similar 'L' shaped symbology - see page 6 [here](http://www.provincia.bz.it/natura-territorio/images/content/Segni_convenzionali_CTP_05k_PAB.pdf), the pattern also appears to be at least somewhat intuitive considering the appearance of current industrialized vineyards - at least not worse than the orchard pattern.  Spanish maps use a [different symbol](http://signa.ign.es/signa/default.aspx?crs=EPSG%3A25830&bbox=897110.2316735298,4590521.762132031,899512.0282766944,4591827.05172592&variant=SIGNAMapa) which however seems much less intuitive.  And that together covers more than two thirds of the world wine production...

matkoniecz:
> Actually i did this together to be able to test how these colors fit together.

OK, I hoped that at least landuse=conservation may be tested independently.

matkoniecz:
> And that together covers more than two thirds of the world wine production...

To make it clear - I am not thinking that lack of recognition for this symbol in Poland as a valid reason for rejecting this change (Poland has about 50 mapped vineyards).

vincentdephily:
Consider the opinion of a French man if you will : I don't recognize those patterns either. I think the original pattern (3 raisins on the vine) is much more recognizable for the vineyards. The orchard pattern isn't great, but I'm not sure what would be better. Perhaps an apple is the most universally-recognizable fruit ?

As for the color changes, they look good to me.

imagico:
I realize this is a question on how much abstraction is desirable and how abstraction is made.  The grape symbol uses fairly little abstraction in design but is quite abstract in the way it connects the landuse to the product produced (a bit like symbolizing forest with piles of logs).  Someone unfamiliar with the product, visiting the vineyard only during winter, will not be able to make this connection.  This makes the pattern unique in this style at the moment, all other patterns use symbols that are abstractions of the appearance of the area in question itself.

One problem of this is it cannot be made any smaller than it is without being unrecognizable.  This means this needs to be a very coarse pattern which essentially makes it unsuitable for the lower zoom levels.

I would be open to a less abstract visualization of the appearance of a vineyard but this seems difficult without being misleading.  My current suggestion might be unclear but i don't think it is misleading in the way it positively creates a wrong association.

Does anyone have examples for use of the grapes symbology in patterns in non-OSM maps?

pnorman:
> Actually i did this together to be able to test how these colors fit together. Changing colors one at a time is difficult since there are always interdependencies. You'd start changing the orchard color and then notice you'd better change it again to better accommodate the change in grass color.

One solution is to first do the colour changes in a spreadsheet, calculating colour differences. This can cut back the number of tests you have to do on the map, and let you be more precise in what to check on each iteration.

imagico:
It seems there is agreement about the colors but some objections concerning the vineyard pattern.  I can change this back to the old one but since all objections so far have been based on a subjective difficulty to recognize the meaning of the pattern i am not really sure this is sufficient basis to decide this is really overall the preferable solution.

So again my primary reasons for using a different pattern summarized:

* the old pattern due to its coarse structure is difficult to read with fine grained vineyard mapping - see [here](http://www.openstreetmap.org/#map=14/48.0942/7.6674) for example.  While there are many wine growing areas where mapping currently is fairly coarse and not dividing the individual plots it would be preferable if the style did actually encourage fine grained mapping.  Areas like [here](http://mc.bbbike.org/mc/?lon=11.147834&lat=46.19835&zoom=17&num=3&mt0=bing-satellite&mt1=mapnik&mt2=mapbox-satellite&marker=) and [here](http://mc.bbbike.org/mc/?lon=-0.218079&lat=44.578474&zoom=16&num=3&mt0=bing-satellite&mt1=mapnik&mt2=mapbox-satellite&marker=) look all right with the current styling at the moment but would look much worse when mapped in minute detail at z=13-15.
* the proposed pattern is a fairly common way to depict vineyards in maps, it is used in many non-OSM maps and also in several OSM maps (like [here](http://opentopomap.org/#map=15/48.09468/7.63734) and [here](http://www.opencyclemap.org/?zoom=15&lat=48.09535&lon=7.64343&layers=00B00)).
* the pattern works well together with the orchard pattern underlining the idea that vineyards are more or less a subtype of orchard.

Further input (arguing either for or against the pattern change or suggesting a third option) would be good.

pnorman:
The new pattern looks okay to me, but i haven't had a chance to load it up in Kosmtik.

yohanboniface:
> i haven't had a chance to load it up in Kosmtik.

Kosmtik bug?

pnorman:
> Kosmtik bug?

E_NOTIME

yohanboniface:
> E_NOTIME

Huhu ok ;)

matkoniecz:
>  I can change this back to the old one but since all objections so far have been based on a subjective difficulty to recognize the meaning of the pattern

And I am not considering my doubts as good reason to reject/change PR - but I will not merge it.

Maybe @math1985 or @gravitystorm may try looking at this PR?

math1985:
I haven't had time to look at it yet.

math1985:
I think both having a more pictographic and a more simplified representation have their merits. I find it hard to pick between the two.

@gravitystorm What do you think?

gravitystorm:
I'm happy with the abstract symbols for pattern fills. I'm slightly biased since (as @imagico links above) I use similar symbols in my own maps. But more importantly, pattern fills can repeat the symbol many hundreds of times, so they should be as simple as possible to reduce visual noise - it's more likely that an abstract symbol will be visually simpler than a (often complex) icon.

matkoniecz:
@imagico - are you planning to make & test new PR for

> move landuse=conservation to nature reserve boundaries

?

imagico:
Opened a new issue for that: #1826

Issue #1649, opened 2015-07-11
Add rendering for amenity=veterinary


kocio-pl:
We should start rendering this amenity: we already show related shop=pet and this tagging scheme is even more popular (15 306 uses). The only problem for me is shape - the paw is probably too small to add a cross inside, so I think about just showing the same icon with health color or maybe adding a circle around like in hospital.

kocio-pl:
This is the simplest form - [just the color change](https://github.com/kocio-pl/openstreetmap-carto/tree/veterinary-icon), you can see both pet shop and veterinary clinic here:

![veterinary-inwalidow](https://cloud.githubusercontent.com/assets/5439713/8635530/cd251342-2826-11e5-80a3-013f65244106.png)

kocio-pl:
Resolved by https://github.com/gravitystorm/openstreetmap-carto/pull/1656.

PR #1656, opened 2015-07-14
Adding amenity=veterinary icon

kocio-pl:
Resolves https://github.com/gravitystorm/openstreetmap-carto/issues/1649.

veterinary-inwalidow


HolgerJeromin:
The idea on reusing the icon is good, both colors are much too similar for me.

kocio-pl:
Any other ideas then?

HolgerJeromin:
moving healthcolor more to red? *duck* :)

daganzdaanda:
Agree, the colour difference alone does not work. I tried to put the cross in the paw:
vet16a
https://gist.github.com/daganzdaanda/ded21489b7056d2b7609
It's not great yet, but it could be possible.

kocio-pl:
The second icon (+ and paw) is what I was also considering, but I don't know when I will have the time to try. Could you make 14px versions of the icons to test if they will be readable at all?

daganzdaanda:
vet16a16
Not very great, but I did not try to improve the pixel-alignment.

dieterdreist:
sent from a phone

> Am 15.07.2015 um 16:02 schrieb daganzdaanda <notifications@github.com>:
> 
> It's not great yet, but it could be possible.
> 


last one is great

kocio-pl:
I don't think so - we should retain as much resemblance between full version icon and smaller versions (like bicycle parking or car service) as it's possible. This case is hard because original paw is too small for a cross and we don't see the upper border (which makes cross look less like a cross), but even then there it should be at least more symmetrical, like in the original.

nebulon42:
@daganzdaanda The circle one look promising (second row, first). Maybe try without white colour, but with subtracted shape as this will likely look quite different (less contrast). Maybe a bit more symmetry and a bit less blur. But great progress here, I have tried myself several times but did not find something suitable.

@kocio-pl While I would normally agree with your concerns re resemblance and symmetry I think for this hard case we should surrender these to the possibilities at hand.

kocio-pl:
@nebulon42 Symmetry is important I think, because the last one brought to my mind human foot clinic, while dog's paw can look [nearly exactly](https://commons.wikimedia.org/wiki/Category:Dog_feet#/media/File:Vasculitis_lesions.jpg) as in your design.

BTW: what do you think about the second icon?

nebulon42:
The second might work too, but is a bit noisy IMO.

rmikke:
Both icons in upper row are OK. The color scheme change is OK.
All the designs with cross inside a paw will be unreadable, IMHO, just because on map it will be not so easy to guess it's a paw...

daganzdaanda:
https://gist.github.com/daganzdaanda/74ca85f6f4dc5fe2ba9b
vet16b32
vet16b16
The leftmost icons are unchanged from the previous try, except for subtracting the negative cross.

kocio-pl:
I still like 1) the most - 2) is less readable at 14 px (the spaces are too small, so the details are less distinct). 3), 4) and 5) all look alike for me at 14 px and I like them less, however they're good too. I will try to test both soon to show them on real map.

daganzdaanda:
@nebulon42 has added an icon to osmic https://github.com/nebulon42/osmic/commit/54eb9611aaf9fc3af70bf8d834a92b69d18e622e
https://github.com/nebulon42/osmic/blob/54eb9611aaf9fc3af70bf8d834a92b69d18e622e/health/veterinary-14.svg
It looks better than my version, because the toes have more space.
The main question is, whether the circle-paw is still recognized as a paw and therefore the icon as a vet icon.

kocio-pl:
The same place (you can compare it with the pet shop and pharmacy here):

this is how @nebulon42 version looks like:
veterinary-circle

and this is @daganzdaanda version (with small fix to be better aligned in 14px square, which adds a bit more space between a paw and a cross):
veterinary-small

For me both are good - what do you think?

matkoniecz:
> This branch has conflicts that must be resolved

matkoniecz:
I prefer version by @nebulon42 Icon by @daganzdaanda seems to be a two separate objects.

Random person was unable to recognise both icons as veterinary (though he has no pet, so it was for him far less likely to consider this possibility) - but it was recognised as healthcare-related thanks to the cross. He also claims that both icons are not resembling paw.

kocio-pl:
I will fix it and rebase once we've decided which one to use.

nebulon42:
Actually are both versions by @daganzdaanda, I merely made some parts symmetrical.

kocio-pl:
Random mapper told me that both are very nice and recognizable (no pet owner too =} ), but the second one is more clear. However I like first more, because field test proves that second may be too small for some users and its elements are really detached.

I have also some other ideas, like dogs/horse head with cross inside or half the head (or paw) with syringe. Would anybody try to test it?

kocio-pl:
Another design idea:
3) veterinary-14-cat-1

ximex:
i like the first one. (the one with the + on the top left)

kocio-pl:
Icon 1) is similar but more readable [than in german style](http://www.openstreetmap.de/karte.html?zoom=18&lat=52.26393&lon=20.9905&layers=B000TT). Still I like icon 2) more, because it's more coherent and its parts are not so tiny.

math1985:
Current symbol looks fine to me.

kocio-pl:
AFAIK amenity=veterinary has no symbol yet... :smile: Which one are you referring to? Maybe you've just thought about shop=pet?

math1985:
I meant the currently proposed symbol. It was closed because the PR was merged :).

PR #605, opened 2014-06-06
Render addr:flats

saerdnaer:
Render addr:flats the same way as addr:housename (fixes #264)

Issue #264, opened 2013-11-13
Render addr:flats or addr:unit tags like housenumber or housename


saerdnaer:
Before I create a pull request, a little discussion:
We in munich have a problem with the tagging of one of the student home neighborhoods. There is a area where every building has the same offical housenumber, but each apartment has it's own building entrance and a own apartment number. Basically an extreme version of terraced housing.

Some user used this appartment number as content for addr:housenumber. In the discussion at our local meeting we came to the conclusion to use addr:flats insteat, but this is not rendered yet. 

Is this the right tag to use, or would you prefer an other one?

More information: 
http://www.openstreetmap.org/#map=19/48.17930/11.55186
http://www.studentenwerk-muenchen.de/typo3temp/pics/b9b5aaba19.jpg

ERnsTL:
There is [another case example](https://help.openstreetmap.org/questions/39962/how-to-tag-block-of-apartments-with-units) of an apartment complex, using addr:housenumber and addr:unit, with detailed info over at the OSM Help Forum.

I am open to any solution tagging-wise, but the rendered result should look better than right now, see the link, and if not too cluttered, include the flat/unit number at high zoom level.

Post solution here please, so that I can finish tagging that apartment complex. Thanks.

1ec5:
> alternatives might be addr:door or addr:unit, but as addr:flats is used 10 times more often I'd probably go with it.

Now `addr:unit` is used twice as much as `addr:flats`. (openstreetmap/iD#3980 would add `addr:unit` to iD, which would likely increase usage of this field where unit numbers are erroneously placed in `addr:housenumber` or `addr:street`.)

BalooUriza:
I would prefer addr:unit instead, as this handles everything along that
same general concept, including private mailboxes, apartments, office
suites, townhouses, trailer park lots, campsites, etc.

saerdnaer:
For reference – here is link to a TagInfo comparison view: https://taginfo.openstreetmap.org/compare/addr:unit/addr:flats/addr:door/addr:housename

tpikonen:
I don't have local knowledge on the original example, but I think the correct way to tag it would be to use, for example, addr:unit="Q" and addr:flats="59" and perhaps also building:part=yes for the ways defining separate flats. Currently the tagging is addr:housename="Q59" and building=yes.

Release 4.4.0 added rendering of addr:unit, so rendering of addr:flats would still be needed to fix this issue.

[...]

PR #1665, opened 2015-07-19
Added support for rendering flat numbers in addr:flats

derickr:
This patch adds support for rendering addr:flats instead of addr:housename - which makes a lot of sense if you see things like:

before

The nodes all have addr:flats, addr:housename and addr:street - but the housename shows up 20 times or so.

With this patch, it renders like:

after

I've made the number one point smaller than real house numbers, but I'm happy to have it render differently - just thought it would be nice that it shows up. It makes a lot of sense in this area as there are *so* many instances. Annoyingly, it requires a new column in the database.

### DIFF
--- a/addressing.mss
+++ b/addressing.mss
@@ -6,6 +6,18 @@
   }
 }
 
+#flatnumbers {
+  [zoom >= 17] {
+    text-name: "[addr:flats]";
+    text-placement: interior;
+    text-min-distance: 1;
+    text-wrap-width: 0;
+    text-face-name: @book-fonts;
+    text-fill: #666;
+    text-size: 8;
+  }
+}
+
 #housenumbers {
   [zoom >= 17] {
     text-name: "[addr:housenumber]";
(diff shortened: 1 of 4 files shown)


pnorman:
Closing. Still working out what tags will be in columns vs hstore, but `addr:flats` is extremely unlikely to be in a column, so this PR would need complete rewriting at that time. Same as https://github.com/gravitystorm/openstreetmap-carto/pull/605#issuecomment-101090230

Some things to consider when it comes time

- what if all the addresses have a flat? Then housenumber/housename wouldn't appear at all. Is this a common situation?
- flat numbers shouldn't be a higher label placement priority than housenumbers. the right approach is probably to filter out flat numbers from the housenumber/housenamelayer, this way if there is a conflict between a nearby housenumber/housename and a flat number, the latter gets displayed. 
- Should all three address layers be combined into one?

I suggest adding comments on #264

PR #1667, opened 2015-07-19
make variable name less misleading

matkoniecz:
During #1662 I completely missed that it affects also hospitals (now I am testing whatever it resulted in worse rendering).

This change should make code less confusing.

It it is likely that this name may be improved - it is better than current one but I am not truly happy about it.

EDIT: Tested hospitals, it seems that hospitals are similar in size too universities - therefore there is no problem with rendering.

### DIFF
--- a/amenity-points.mss
+++ b/amenity-points.mss
@@ -1829,7 +1829,7 @@
       text-size: @landcover-font-size;
       [way_pixels > 12000] { text-size: @landcover-font-size-big; }
       [way_pixels > 48000] { text-size: @landcover-font-size-bigger; }
-      text-fill: darken(@school, 70%);
+      text-fill: darken(@educational_areas_and_hospital, 70%);
       text-face-name: @landcover-face-name;
       text-halo-radius: 1;
       text-halo-fill: rgba(255,255,255,0.6);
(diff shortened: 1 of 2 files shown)


kocio-pl:
+1

ximex:
+1

HolgerJeromin:
Perhaps we should not hide the hospital between the education stuff, but reorder this a bit:

    [feature = 'amenity_university'],
    [feature = 'amenity_college'],
    [feature = 'amenity_school'],
    [feature = 'amenity_hospital'],
    [feature = 'amenity_kindergarten'] {

so start with the hospital and than add the education features:

    [feature = 'amenity_hospital'],
    [feature = 'amenity_university'],
    [feature = 'amenity_college'],
    [feature = 'amenity_school'],
    [feature = 'amenity_kindergarten'] {

gravitystorm:
The order of the entries in the cartocss files doesn't affect the order in which the features are drawn. In short, the features are selected from the database in the SQL query, and then each feature in the collection is rendered one at a time, by checking all the rules for that feature, drawing whatever is appropriate for that one feature, then moving onto the next feature, and checking all the rules for that one, etc.

So if you want to change priorities, you need to consider the order of the results in the SQL queries, not the cartocss selectors.

math1985:
I think @HolgerJeromin meant un-hiding the line in the code (and I think I agree with him).

Issue #869, opened 2014-08-09
Add label for amenity=ferry_terminal


kocio-pl:
Name of ~~the route=ferry~~ and amenity=ferry_terminal should be visible. Currently only MapQuest does this on the main OSM page. Example:

https://www.openstreetmap.org/way/37930736
https://www.openstreetmap.org/node/33238753

pnorman:
c.f. https://phabricator.wikimedia.org/T156205 and my problems identify terminal details

PR #1673, opened 2015-07-21
Adding amenity=ferry_terminal icon

kocio-pl:
Resolves https://github.com/gravitystorm/openstreetmap-carto/issues/869.

Rendered like bus stop - z=16:
ferry-z16

z>=17 (with osmic icon):
ferry-z17-osmic


kocio-pl:
Edited version, based on osmic with top inspired by MapQuet Open icon (both are too solid and look rather like bollard for me):
ferry-z17-my-turkawka
ferry-z17-my-pliszka

imagico:
I don't think there is a gain from adding a large specific for this purpose.  The meaning is perfectly clear with just a generic square symbol which uses less space and adds less clutter to the map.

matkoniecz:
Yes, especially that symbol is not really recognisable. Generic square seems much better.

kocio-pl:
Hm, you may be right. However I'd like to hear if there are some counterexamples, where just the dot with label would be suboptimal choice maybe?

math1985:
At larger terminals, such as [here](http://www.openstreetmap.org/#map=16/51.9740/4.1302), it might not always be immediately clear which building is the ferry terminal.

kocio-pl:
So maybe we should show this icon just for areas?

ximex:
the icon itself looks good.

math1985:
Is is not possible to come up with a simpler symbol for 'ship'? The 'terminal' concept will be clear enough from the ship, I think.

kocio-pl:
Any ideas are welcome. How should it look like? 

I can also split this PR into generic blue square part, which is not controversial, and we can add the icon later, once there is one drawn and approved.

matkoniecz:
I think that splitting is a good idea.

pnorman:
Not sure how well it works - it can't always be distinguished from a bus stop

image

Remember, the marked points are not on the water's edge.

See for example, [Tsawwassen](http://www.openstreetmap.org/node/660826491). Now, this should probably be an area, but the centroid of the area isn't going to be at the coastline either.

[Sidney International](http://www.openstreetmap.org/way/136695687) is an area, as are [Swartz Bay](http://www.openstreetmap.org/way/131450912) and [Langdale](http://www.openstreetmap.org/way/299739618).

A ferry icon is useful at lower zooms to show where the terminal is, but at high zooms its pretty eident where the road meets the water. If we wanted an icon there, we should do it based on where the lines touch, not ferry terminals, which are different.

kocio-pl:
I guess the best choice for terminal icon could be [something like what we will shortly use for the ford](https://github.com/gravitystorm/openstreetmap-carto/pull/1625): just the water and the arrows for showing there is a way over them. However the question is if we want to distinguish those icons, and if yes - in what way should they be different?

kocio-pl:
What about simpler shape like this:
ferry-arrows

matkoniecz:
How it works for examples from https://github.com/gravitystorm/openstreetmap-carto/pull/1673#issuecomment-125102218 ?

kocio-pl:
Sidney:
ferry-arrows-sidney

Swartz Bay:
ferry-swartz

Langdale:
ferry-langdale

Tsawwassen node is overcrowded with tags (ferry_terminal+locality+traffic_signals), so not suitable for testing this icon.

kocio-pl:
Maybe we should use the aeroway color for ferry terminal areas too. I guess this might be even used for all the transport terminals (like bus stations).

sabas:
Rendering the area could be useful.
Should the single berthing points inside the terminal be tagged differently?
Sample area http://overpass-turbo.eu/s/aW6

jojo4u:
@sabas From what I understand on a complex facility amenity=ferry_terminal should be the whole area (buildings, parking spaces), waterway=mooring, mooring=ferry should the the point where highway and ferry meet. This extends the meaning ferry_terminal a bit but in 2007 there was not so much hindsight (http://wiki.openstreetmap.org/wiki/Talk:Proposed_features/Ferry_Terminal)

jojo4u:
@kocio-pl I improved Tsawwassen.
The other examples suffer from "too much highway=service", something I already fixed on Tsawwassen and will be on Swartz Bay (https://www.openstreetmap.org/note/393069)

kocio-pl:
First I will have to learn how to tag the airport, then find out what parts ferry terminal consists of, and in the end try to find some analogies. It would be also good to study bus and railway station - or maybe also other transportation stations, to be as universal as possible. However I'm not into transportation too much and try to do few other things, so the help would be welcome. 

I just thought it would be dead easy task... =}

matkoniecz:
Unfortunately in current state it is not ready for merge and it seems unlikely to change in the near future - so I will close it for now. 

Updated/new PRs solving this problem are welcomed!

> I just thought it would be dead easy task... =}

Almost everything is harder than it seems before starting it.

kocio-pl:
Sure, currently I plan to start with defining airport/heliport/ferry(port) as a similar big areas.

Issue #1681, opened 2015-07-24
Change pattern icon for landuse=quarry


kocio-pl:
I think we should change "q" pattern icon we use currently for quarries for something more universal, like pickaxe icon for example (here's standard 14 px icon - instead of just 7 px - on 30px pattern tile just for test rendering):

![quarry-png-white-30](https://cloud.githubusercontent.com/assets/5439713/8878374/e9b3e520-322a-11e5-82e1-c7ee8c610f91.png)

We could also convert all the pattern icons to SVG, but I was not able to make it work with SVG and it looks like [polygon patterns can't have icon colors defined](https://github.com/mapbox/carto/blob/master/docs/latest.md#polygon-pattern): not polygon-pattern-fill nor polygon-pattern-color did work in my test. Also all the examples I saw include just PNG icons.

math1985:
That pattern was supposed to denote a Q?

I like the idea of changing this, how do other maps represent quarries?

kocio-pl:
Surprise, isn't it? :smile: Initially I thought it's a magnifying glass, but after I've checked it, it became obvious what it was supposed to be. However it's still not clear in general because it uses name in English to describe something universal.

Symbols can be made less contrast to not drag too much user's attention.

imagico:
IMO not bad but the symbol is much too heavy and even when weaker it would not work well at the lower zooms - you would need to go without at least for z=10-12.

Instead of a single pick two crossed hammers/picks are also a common symbol for mines and could be used.  See [here](https://www.google.de/search?q=mine+symbol&tbm=isch) for a few examples.

The standard pattern for quarries on German topographic maps is a brick pattern, like http://opentopomap.org/#map=14/50.91023/6.50876 - this however does not really work that well here due to the fairly broad meaning of landuse=quarry in OSM.

And you need a power of two for the pattern size which gives you additional constraints - but you do not necessarily have to use a square arrangement.

math1985:
> Instead of a single pick two crossed hammers/picks are also a common symbol for mines and could be used.

+1

kocio-pl:
Crossed tools are very common and even look surprisingly good at 7 px (the same size as "q"):
![quarry-x-7](https://cloud.githubusercontent.com/assets/5439713/8882505/9a9b8a84-3247-11e5-8d00-ec09a6e2c000.png)

[...]

PR #1696, opened 2015-07-28
New icon for landuse=quarry

kocio-pl:
Resolves https://github.com/gravitystorm/openstreetmap-carto/issues/1681.
quarry-grey10


imagico:
That looks much better already but the rotated square pattern might not be ideal for a diagonal cross shaped symbol.

And of course such a coarse pattern would not make sense from z=10 on - you should either make it denser or start at a later zoom level.

And you might want to test in in areas where stuff is mapped within the quarry, like:

http://www.openstreetmap.org/#map=15/37.4159/22.0923
http://www.openstreetmap.org/#map=15/-31.9847/141.4359

kocio-pl:
I'm ready to test it more, now the symbol shape, size and color are good enough.

Which pattern do you think would be good? This is not rotated square pattern, I think that's rotated triangular pattern - I have used jsdotpattern, so probably this or maybe regular triangular.

I don't know if we want it to be denser - @matkoniecz asked me to make distance bigger than default 30 yet smaller than 60 (double the standard), so I took 45.

kocio-pl:
I don't know how to test real zooming (not just the slider in TileMill just showing the objects for given zoom level of the same box), but here's a series of zoom levels z10-z15 in the busiest Australian quarry:
quarry-au-z10
quarry-au-z11
quarry-au-z12
quarry-au-z13
quarry-au-z14
quarry-au-z15

pnorman:
> not just the slider in TileMill showing the objects for given zoom level in the same box

Take screenshots and clip them with Paint or something similar

imagico:
When you select a triangular pattern in jsdotpattern it gets stretched to fit the square pattern size.  In extreme cases like the one you chose it becomes a rotated square pattern.  If you want a distance of 45 with a triangular pattern you better use a pattern size of 512 - you get an aspect ratio of 8:11 then which is much closer to the ideal 1:sqrt(2).

kocio-pl:
@pnorman I guess I miss the box keeping the same size in TileMill (instead of keeping the same boundaries), but probably it's better to do it manually than crop, because this will lead to big pixels, which is not suitable for real testing.

@imagico OK, that was what I suspected looking at previous pictures with d=30 and d=60. I will try this way.

kocio-pl:
New pattern with "manual zoom" testing (z10-15):
quarry-10
quarry-11
quarry-12
quarry-13
quarry-14
quarry-15

Given that the shape is non uniform in both dimensions and the area is especially noisy, I am happy with the result.

kocio-pl:
When the shape is more regular and there are not much details on top, even z10 works as expected: 
quarry-pl-10

matkoniecz:
Very nice improvement!

For z11, z12 something like first pattern in https://github.com/gravitystorm/openstreetmap-carto/issues/1681#issuecomment-124766142 but with smaller symbol would be probably even better.

https://cloud.githubusercontent.com/assets/899988/9037487/431fcafc-39ec-11e5-9797-0d0f7b3faaed.png
https://cloud.githubusercontent.com/assets/899988/9037488/43236dc4-39ec-11e5-8758-ea3b1d3d6323.png
ttps://cloud.githubusercontent.com/assets/899988/9037490/4323c3f0-39ec-11e5-92a8-05fadfaf0f1b.png
https://cloud.githubusercontent.com/assets/899988/9037489/4323979a-39ec-11e5-96da-b940a2a43c9c.png
https://cloud.githubusercontent.com/assets/899988/9037491/4323aa0a-39ec-11e5-83ad-acccd6c14325.png
https://cloud.githubusercontent.com/assets/899988/9037492/4325de7e-39ec-11e5-8aab-0a45e3543feb.png
https://cloud.githubusercontent.com/assets/899988/9037493/4338af36-39ec-11e5-9d06-87bdfd67ee2c.png
https://cloud.githubusercontent.com/assets/899988/9037495/433df630-39ec-11e5-9f27-e6131dffab12.png
https://cloud.githubusercontent.com/assets/899988/9037494/433daafe-39ec-11e5-857d-e4169eba1852.png
https://cloud.githubusercontent.com/assets/899988/9037496/43403b84-39ec-11e5-92e9-94cee887aa65.png

matkoniecz:
Are you interested in tweaking z11, z12?

kocio-pl:
I think I'm done here, because I don't have any more ideas - you can take over this issue if you like. BTW: thanks for "real" mockups! :smile:

matkoniecz:
I thought about 

> For z11, z12 something like first pattern in #1681 (comment) but with smaller symbol would be probably even better.

but it may be done as separate PR.

matkoniecz:
I think that https://commons.wikimedia.org/wiki/File:Mining_symbol.svg should also be added to repo to allow easier modification to quarry style - as svg file is not recoverable from png pattern. 

Maybe new subfolder in /symbols/ would be a good idea?

imagico:
I agree that the proposed pattern is unsuited for the lower zooms (z=10-12) for all but the largest quarries.  It would probably make sense to either use a denser pattern in general or at the lower zooms or not start showing the pattern before z=13.

kocio-pl:
Yes, subfolder "patterns" for such icons with rudimentary documentation how it was done (parameters for jsdotpattern or other software and further image manipulations) would make sense, so other users can have real source and making instructions.

BTW: I have used [another file](https://commons.wikimedia.org/wiki/File:Schlaegel_und_Eisen_nach_DIN_21800.svg) here, because it creates thinner lines.

imagico:
By the way i added the two quarry symbols to jsdotpattern - i suppose what @matkoniecz had in mind was something like [this](http://www.imagico.de/map/jsdotpattern.php#x,512,jdp76982;gt,32,32,32;rd,1,0,0,quarry1,0.62,7,6,0,jdp89011;).

gravitystorm:
> unsuited for the lower zooms (z=10-12) for all but the largest quarries

Should we instead be considering using the pattern only when the quarry has a large enough way_pixels?

kocio-pl:
This could be useful condition. The most important factor here is probably shape, but unfortunately we can't measure it.

imagico:
> Should we instead be considering using the pattern only when the quarry has a large enough way_pixels?

Yes, i think so.

It might also be considered taking this approach for patterns elsewhere but in many cases this can cause problems and confusion due to arbitrary splits in polygons etc.  For quarries however this is fairly uncritical.

kocio-pl:
Is somebody to take over this? I'm more happy with this PR than with current state, so if there are no substantial objections, I suggest to pull it and open another ticket for ironing out the details.

math1985:
Sorry, I had so far no time to look at this.

math1985:
Much better, thanks a lot! This only changes the icon, not the code, right?

kocio-pl:
Yes, as you can clearly see. :smile: 

BTW: how should we make documentation for recreating patterns if somebody wants to play with them? Should it be one file like patterns.md or maybe one file per image like quarry.md?

math1985:
> Yes, as you can clearly see. :smile: 

Yes, just doublechecking you didn't forget to add a file (happens to me all the time).

matkoniecz:
> BTW: how should we make documentation for recreating patterns if somebody wants to play with them? Should it be one file like patterns.md or maybe one file per image like quarry.md?

I would make patterns.md and put it in symbols folder.

Issue #1691, opened 2015-07-27
suggestions for tuning farmland color


imagico:
Current farmland color is fairly dark and strong, same brightness as the urban land covers, although mapping is frequently on very large areas.  This causes problems with road colors as well as similarity with several other landcovers.  This matter was already discussed in #1006.  In general it would make more sense to have farmyard in the same brightness as the urban landcovers and make farmland brighter.

Here are three suggestions for changing it to a lighter tone - with comparison to other landcover colors it might conflict with:

The current color:

![current](http://www.imagico.de/files/carto-farmland.png)

Suggestion 1:

![current](http://www.imagico.de/files/carto-farmland1.png)

Suggestion 2:

![current](http://www.imagico.de/files/carto-farmland2.png)

Suggestion 3:

![current](http://www.imagico.de/files/carto-farmland3.png)

All improve contrast with urban landuses and mud but reduce contrast with bare_ground and sand.  In reality farmland edging to sand or other bare ground is rare so this might not be that much of a problem but ultimately this of course needs some practical testing

kocio-pl:
I like 1) scheme - maybe this will be enough. Could you make some comparative renderings?

imagico:
I don't plan to do more elaborate testing at the moment - just wanted to put up these ideas since discussion w.r.t. road colors brought up the possibility.  Anyone who wants to implement a color change and make more thorough testing on real world samples is welcome.

Note in #338 it was rightfully mentioned that brown color is often not really that intuitive for farmland - a more greenish tone might therefore be better.  Since grass is now somewhat less yellow there is a bit of room in the bright yellow-green area that could be used.

kocio-pl:
BTW: I think farmyard can be the same color as landuse=residential, since they are basically telling the same (close environment where people live enclosing basic amenities), for me farmyard is just a special kind of residential area.

imagico:
No, farmyard is more like a subtype of industrial, even if there are often residential buildings present.  It makes sense to have a separate color for that.  Sadly a lot of remote mapping these days does not make this distinction and a residential landuse polygon is drawn around any group of buildings in rural areas.

dieterdreist:
sent from a phone

> Am 27.07.2015 um 19:20 schrieb kocio-pl <notifications@github.com>:
> 
> for me farmyard is just a special kind of residential area


TL;DR;
They are more producing than residential places 

***
[...]

PR #1701, opened 2015-07-29
new farmland color - makes farmland less prominent

matkoniecz:
new_farmland

Currently mapping landuse=farmland in most locations results in worse and less readable map. It is caused by too strong color and large size of landuse=farmland areas. In result it is frequently the most noticeable (or at least one of the most noticeable) features. Frequently it is more prominent than far more important features like villages or towns. There is also problem of residential, industrial, farmyard and other landuses not visible enough in sea of landuse=farmland. This problem is larger than implied by color difference computations due to differences in size of typical landuses and distribution of obscuring features. Density of buildings, names, roads etc are significantly higher on landuse=residential/industrial etc than on landuse=farmland.

All distances are in delta_e94 metric, as compute by color gem for ruby. Following data describes how close colors for current (first entry) or potential (next) are close to other landcovers. It is preferable to keep this values high to make landcovers differentiable. Colors rejected on grounds of clearly too small color differences or resulting in an ugly map were omitted from list below.

current style - lch(89, 12, 80)
landuse=garages: 4.55
foreground of @bare_ground: 4.84
landuse=farmyard: 5.45
amenity=prison hatches: 6.25
/////
natural=sand: 6.48
unmapped land: 8.05

farmland lch(91, 12, 80)
foreground of @bare_ground: 4.13
landuse=garages: 5.39
natural=sand: 5.64
aeroway=aerodrome: 6.18
/////
unmapped land: 6.72

farmland lch(93, 12, 80)
foreground of @bare_grounde: 4.65
natural=sand: 5.11
hospital/educational area: 5.27
/////
unmapped land: 6.07

farmland lch(94, 12, 80)
hospital/educational area: 4.85
foreground of @bare_ground: 5.02
natural=sand: 5.23
unmapped land: 5.8

farmland imagico #1
foreground of @bare_ground: 3.92
landuse=garages: 4.45
amenity=prison hatches: 5.72
aeroway=aerodrome: 5.92
/////
natural=sand: 6.13
unmapped land: 6.86

farmland imagico #2
natural=sand: 4.02
landuse=garages: 4.48
hospital/educational area: 4.49
foreground of @bare_ground: 5.18
/////
unmapped land: 7.16

farmland imagico #3
natural=sand: 3.47
hospital/educational area: 3.85
landuse=garages: 5.7
leisure=nature_reserve border: 6.09
/////
unmapped land: 8.31

raw data for distances: https://gist.github.com/matkoniecz/3d59bfde1aefb75a6596

Overall lch(94, 12, 80) [#fbecd7] was the most promising. In my tests it was an improvement - farmland is no longer so dominating. It is quite bright what may be not liked by some - but it also makes features on it more visible. Areas of other landuses neighbouring with farmland are now more visible.

https://cloud.githubusercontent.com/assets/899988/8958284/3962b8bc-3604-11e5-80b1-abe95e6ab0e8.png
https://cloud.githubusercontent.com/assets/899988/8958283/3962b934-3604-11e5-8ccf-1b5d18600e18.png
https://cloud.githubusercontent.com/assets/899988/8958285/3965307e-3604-11e5-9752-61f51d933d1d.png
https://cloud.githubusercontent.com/assets/899988/8958286/396ccb72-3604-11e5-81d6-a7839518307d.png
https://cloud.githubusercontent.com/assets/899988/8958287/396e5dd4-3604-11e5-8f65-e2860b756007.png

### DIFF
--- a/landcover.mss
+++ b/landcover.mss
@@ -26,8 +26,8 @@
 @industrial-line: #C6B3C3;  // Lch(75,11,330)
 @railway: @industrial;
 @railway-line: @industrial-line;
-@farmland: #EDDDC9;         // Lch(89,12,80) (Also used for farm)
-@farmland-line: #C8B69E;    // Lch(75,15,80)
+@farmland: #fbecd7;         // Lch(94,12,80) (Also used for farm)
+@farmland-line: #d6c4ab;    // Lch(80,15,80)
 
 @farmyard: #EFD6B5;         // Lch(87,20,80)
 @farmyard-line: #D1B48C;    // Lch(75,25,80)


matkoniecz:
I know that it break constant lightness for major landuses, but landuse=farmland is unlike highway=residential/industrial/retail - its importance is lower, covers much bigger surfaces and it is usually without other features like buildings covering its area.

imagico:
IMO the consistency should be between residential/industrial/retail/commercial and farmyard, not farmland - and that is easy to achieve.

Here the abstract sample for comparison:

new farmland

I think it looks good - i did not originally consider making it brighter than bare_ground/sand but it works fine and creates good contrast towards urban landcovers which is important for the overall result.  I'd suggest trying less red and more green but this would put you in conflict with the hospital/education color.

math1985:
> I know that it break constant lightness for major landuses

@pnorman What do you think of this?

daganzdaanda:
IMHO it's an improvement. Good things about it: 
- it gives more leeway to making the greens paler
- it makes the more important things stand out more by improving contrast

The only thing that might be irritating (to me) is that it's so pale that the contrast to "empty map gray" is very little. This might lead to some unsightlyness when people leave little empty areas next to farmland, as in z15 in the south of london example https://cloud.githubusercontent.com/assets/899988/8958287/396e5dd4-3604-11e5-8f65-e2860b756007.png

nebulon42:
I think it doesn't necessarily have to be this bright, but should be desaturated a bit more. Now it looks peach-like and makes the map even more colourful.

pnorman:
> > I know that it break constant lightness for major landuses
> @pnorman What do you think of this?

I do think it's an issue, and would much prefer results obtained with just adjusting the hue and chroma.

matkoniecz:
> adjusting the hue and chroma

This is quite problematic to achieve intended results without changing lightness. It is too prominent and too close to residential/industrial (situation is worse than implied by checking just colour difference due to differences in distribution of map features - image from the first post, attached here again is a good example).

Reducing chroma may reduce problem of undue prominence, but at the same time problem of being too close to other main landuses (especially distance to residential is smaller).

[](https://cloud.githubusercontent.com/assets/899988/8958274

daganzdaanda:
Re "constant lightness for major landuses" - having two groups might make sense: 
One group for mainly built-up or paved areas (residential, industrial, retail, commercial, parking, farmyard), another group for "natural" and farmland landuses (farmland and forest, wood, meadows / grasslands, scrub, heath, bare_rock, scree,.....?)

That way, a built-up area would be easier to spot in a big "natural" environment.

imagico:
To me the recent changes that significantly brighten the urban areas on z=10-12 in particular call for brightening the farmland color.  Both the French and the German style by the way use a brighter farmland even though in both the urban areas are darker at these zooms overall - see:

http://mc.bbbike.org/mc/?lon=9.944944&lat=52.161734&zoom=12&num=3&mt0=osmfr&mt1=mapnik&mt2=mapnik-german&marker=

@matkoniecz - if you brighten the farmyard color slightly to the same lightness as the urban landcovers this would maintain the lightness consistency in that regard.  Practically it would also slightly improve visibility of buildings on farmyards.

imagico:
After the road color change this still seems a serious improvement, especially at z10/11 in rural areas with many villages like:

http://www.openstreetmap.org/#map=11/48.3092/7.5370

It would also mean a significant improvement regarding the widespread feeling the map is becoming too pale (#1863).

nebulon42:
I think that lighter is better, but please not this saturated peach-like orange-brown.

imagico:
If you have a suggestion for a better color that would be welcome but since the suggested color is not more saturated than the current one and similarity with other colors is an issue i am not sure there is much room.

polarbearing:
Could we please see the proposed farmland colour in the context of the new road rollout?

mboeringa:
> I think that lighter is better, but please not this saturated peach-like orange-brown.

\+ 1, I agree a mildly de-saturated, and maybe slightly lighter, version might work better.

kocio-pl:
@nebulon42 @mboeringa What about one of these propositions then: https://github.com/gravitystorm/openstreetmap-carto/issues/1691#issue-97453449?

kocio-pl:
@polarbearing z12 with current tertiary (white):
c vn3vbb
hrkgvag5

However there is a proposition for changing the tertiary color ([see the test](https://github.com/gravitystorm/openstreetmap-carto/issues/1974#issuecomment-157263919)).

lsces:
In my own private rendering I disabled 'farmland' so that the 'farmyard' elements could actually be seen, but the main problem with farmland in the UK is the fact that large areas are tagged in some counties and not in others so the whole thing just looks a mess. This is a bit like "What should the default be" and to be honest I don't think those areas currently tagged as 'farmland' are really representative of the default for the areas that have been tagged currently, and the fact that an adjacent county does not have a default tagged is even more confusing.

dieterdreist:
sent from a phone

> Am 17.11.2015 um 08:44 schrieb Lester Caine <notifications@github.com>:
> 
> This is a bit like "What should the default be" and to be honest I don't think those areas currently tagged as 'farmland' are really representative of the default for the areas that have been tagged currently, and the fact that an adjacent county does not have a default tagged is even more confusing.


a default does not have to be tagged. If you don't want to add features, just don't add them, but when you add them, try to be precise (e.g. don't include areas into farmland polygons that aren't farmland, make them not too big in order to facilitate maintenance and keep complexity low)

cheers 
Martin

pnorman:
I think it's an incremental improvement.

lsces:
That does at least make it a little easier to spot the farmyards from the 'background' but comparing the first images with the last ones on this thread, I still find the originals have a much better contrast on everything in general.

math1985:
This also improves #1755.

PR #1702, opened 2015-07-29
New icons and YAML for amenity=post_office and amenity=post_box

kocio-pl:
Taken over after https://github.com/gravitystorm/openstreetmap-carto/pull/1558.
Implements part of #1165.

Please test the rendering. I see no problem in code, which is pretty simple, but I get rendering like this (black instead of brown):
osm-carto-test


amenity-points.mss
```diff
@@ -336,13 +336,17 @@
...
+    point-file: url('symbols/post_box-12.svg');
+    marker-fill: @amenity-brown;
+    marker-placement: interior;
+    marker-clip: false;
   }
 
   [feature = 'amenity_post_office'][zoom >= 17] {
-    point-file: url('symbols/post_office.p.16.png');
-    point-placement: interior;
+    point-file: url('symbols/post_office-14.svg');
```
math1985: Should be `marker-file`.

kocio-pl:
Thanks! Now it's OK (ATM is visible here for comparison):
post-icons

PR #1703, opened 2015-07-29
inherit fill for living street and pedestrian

matkoniecz:
rendering is not changed


HolgerJeromin:
Thumbs up for cleaning up the code. :-)

Issue #1530, opened 2015-05-05
Add special rendering for shop=sports


kocio-pl:
1) [My proposition](https://github.com/kocio-pl/openstreetmap-carto/blob/7c20bed8f090cbda254ddf46ea182c8bad63e695/symbols/sports-14.svg) for this type of shop is this: ![sports-14](https://cloud.githubusercontent.com/assets/5439713/7484224/7e7abf00-f386-11e4-817d-e5e544c8f9c5.png)

[Mentioned on meta-issue #1460]

math1985:
The icon is not recognizable to me. Maybe some kind of ball would make sense?

kocio-pl:
I didn't want to duplicate this shape, because I tend to think it may be needed for something else (like to indicate the soccer/rugby stadium/tennis court etc), so I'd like something not too common outside the shop/home use. The same with a tennis rocket (I started with it plus something else: ![sports-14-1](https://cloud.githubusercontent.com/assets/5439713/7484733/5061ef02-f38c-11e4-9ae9-06effabe984c.png)) or a boxing glove. But maybe I'm wrong and we should use them anyway?

I guess the most iconic soccer ball can be quite hard to draw, because color patterns could be barely visible given small number of pixels.

kocio-pl:
Hm, what about leaving the accessories altogether and use more general metaphor? ![sports-14](https://cloud.githubusercontent.com/assets/5439713/7484805/76e14de8-f38d-11e4-9e20-b83a9a4ddfe6.png)

HolgerJeromin:
[The Olympic Movement is very protective of its symbols](http://en.wikipedia.org/wiki/Olympic_symbols#Intellectual_property)

kocio-pl:
Wow, quite impressive. =}

I try also other sport metaphors like podium or medal, but the sketches are still not good enough.

math1985:
What about a diagonal tennis racket?

kocio-pl:
Diagonal shapes are harder to be kept symmetric/regular in vector form, but the raster proof of concept looks like this: ![sports-14-4](https://cloud.githubusercontent.com/assets/5439713/7485890/ae18e51e-f39c-11e4-9939-7519bf52e387.png)

daganzdaanda:
The tennis racket and ball looks good, as does the dumbbell. But both might be mistaken for a place where you use them (e.g. tennis club or fitness studio / gym). Would it be clear enough through the colour only that these are shops?

[...]

Issue #660, opened 2014-06-23
Guidelines on adding new icons


matthijsmelissen:
There are quite a few requests for adding new icons (#591 #540 #570 #487 #456 #452 #450 #419 #418 #336 etc), and even more features that we currently do not render, but might be nice to have on the map.

How do decide which POIs we render as icon?

Do we want to increase the icon set, or keep the set of icons small?

@gravitystorm you indicated in the past that you are in favour of an approach that doesn't involve hundreds of new icons. Could you give some more background to that statement?

A large icon set might give more information. But having too many icons might also make the map unreadable. We should also make sure that all icons are understandable.

brycenesbitt:
I think it's time to reframe this debate.

POI's, particularly obscure ones, don't clutter the map.
Line styles and fills get confusing quickly.

As such I think the "1000 uses" threshold is particularly lacking in vision.  It's an electronic map: it can support many constituencies, including small ones.

imagico:
As @math1985 said the main purpose of displaying POI icons is to allow map users to search for a particular type of feature.  This only works if there is a limited set of symbols, if there are thousands of different icons it does not (needle in a haystack problem).  With the brown icons this mostly still works, with the purple ones it no more does in most cases (IMO).

matkoniecz:
> main purpose of displaying POI icons is to allow map users to search for a particular type of feature

Ad second purpose is too recognize object based on icon and context (IMHO works for shop icons like ones added in #1493 and is not working so well in case of some other icons like #1776)

brycenesbitt:
My main purpose for POI icons is to see what's in an area.  I rarely search.
OSM is one dataset, but many many uses.

I have no needle in the haystack problem with a multitude of POI types.
Rather the opposite.
If anyone does have a needle in the haystack problem, that's a problem with
the client.  It should in search show the most commonly needed
options, with a pathway to getting more.

kocio-pl:
I generally think the same as @brycenesbitt - I don't look for exact objects, I rather want to know what's up there (as much as it's possible) and I think something is wrong when there are holes or just some unidentified structures which don't look like plain buildings. POIs are helping me understand the area along with roads, land colors, names and boundaries. The higher zoom, the more important POIs and icons are.

@imagico 
> This only works if there is a limited set of symbols, if there are thousands of different icons it does not (needle in a haystack problem).

[...]

PR #1707, opened 2015-07-31
Adding shop=sports icon

kocio-pl:
Resolves https://github.com/gravitystorm/openstreetmap-carto/issues/1530.

New icon alone and in comparison with shop=clothes:
sports-alone
sports-clothes


RobJN:
Doesn't say "sport" to me - still says clothes. Why not try a ball or similar.

matkoniecz:
For me it is good icon for shop=sport_clothes, not shop=sport.

kocio-pl:
Another idea:
2) sports-14-13

kocio-pl:
@RobJN Ball was hard to be drawn in such a small space, but I also don't like to use attributes of any kind of sport because I think we may start adding icons for different sport objects. You can read more about problems with creating right sport icon [in the corresponding issue](https://github.com/gravitystorm/openstreetmap-carto/issues/1530).

mboeringa:
I am fine with the original "cloth" type proposal with shorts. Shorts are quite strongly associated with "sport". The second icon has a conflict with the archaeological site icon, and is less signifying.

> For me it is good icon for shop=sport_clothes, not shop=sport.

What's the supposed difference? I doubt people could differentiate, all shops related to sport, except maybe those specialized ones selling trophies, sell clothes... And the official tag seems **shop=sports** ;) (http://wiki.openstreetmap.org/wiki/Key:shop), shop=sport_clothes is not even listed on the Wiki...

kocio-pl:
@mboeringa It was in fact derived from archaeological site icon and I was thinking about making it even more slim ([like this one](http://www.kajasport.pl/pol_pl_Puchar-T-7090-6213_2.jpg)), but first I wanted to know if this metaphor is acceptable at all. BTW: archaeological icon is amenity-brown, while sports will be shop-violet, so I don't think it could be misleading.

Maybe single item is too hard to use as an icon - we can resort to multiple objects, [as I have already proposed in #1530](https://github.com/gravitystorm/openstreetmap-carto/issues/1530#issuecomment-105257929), but I want to test every possible way to have something more clear.

kocio-pl:
Slimmed down version with more typical foot:
sports-cup-tall

math1985:
Would something like sport-icon work?

kocio-pl:
One of my draft was a football ball, but it was unrecognizable, given the available pixel resolution, so I guess this one would be even worse, because the were patches and there are thin lines.

However I don't like the idea of showing any particular sport discipline as a general symbol, because I foresee they might be needed in the future for different sport facilities (like pitches for example) and strongly prefer something neutral in this regard. It would be better to have one piece like this one, but if not, the collection of different symbols is acceptable for me (just less clean).

kocio-pl:
I have just checked that leisure=sports_centre has no icon, so I'd like to reserve collection of symbols for it, because that will encourage people to tag it more precise (with sport=*).

mboeringa:
> Slimmed down version with more typical foot:

If you want to stick with the "cup" metaphor, that is indeed an improvement. It looks more like a winning cup instead of an ancient amphora, and I think it now has enough distinction combined with the different colour.

kocio-pl:
Yes, I think it's decent enough now and I like it this way.

matkoniecz:
The problem is that it is meaning is not obvious - what is problematic enough for me to not merge it but not enough to reject it (as icon is fitting this shop type).

Unfortunately I have no idea for a better icon and cup icon is the best one from already proposed.

kocio-pl:
I have already wrote down all my arguments regarding "obviousness" and why it's overrated and inconsistent (see [here](https://github.com/gravitystorm/openstreetmap-carto/issues/660#issuecomment-135059905) for example), but nobody have even tried to discuss with them (which is a big surprise for me!). I guess it means it's probably just a matter of personal feelings - and I see no way to really discuss them... :no_mouth:

imagico:
> what is problematic enough for me to not merge it but not enough to reject it.

I think it helps to look at the balance in terms of solving problems adding problems. Here:

Solved:
* it would solve the problem that a certain type of shop is not indicated with a specific symbol and thereby adds useful semantic information to the map.

Added:
* it adds an additional type of icon making the map as a whole more difficult to read.
* it adds ambiguity and confusion by use of a symbol that is no clearly associated by most people with the type of feature it illustrates (people don't go to a sports shop to buy cups, at least not AFAIK).

Which weights more is a matter of opinion of course.

mrwojo:
The icon looks specifically like [shop=trophy](http://wiki.openstreetmap.org/wiki/Tag:shop%3Dtrophy). (That tag is underrepresented in OSM. There's far more than 48 of those in the real world.)

kocio-pl:
@imagico I would describe it in a different way, with more specific gains than just one big "more informations" on the positive side. For me it involves wider context  (like "makes high zoom levels more usable"), because I also care for consistency (what about other equally not clear icons/lines/patterns we already have - should we remove vineyard pattern or any road colors, because they are not clear just by looking at them?) and the relations with the rest of ecosystem (they're very popular, so it's probably important for the people to see them different than a standard shop). But you're right - in the end kind of +/- list might help when in doubt.

There's also multiple-objects version available. I just like it less and I think it could be used for something different, but it's still not being used and those items are what you can really see in such a shop.

@mrwojo There are basket shops, which would be more relevant for current "convenience" icon, yet I think the look is much less important factor than popularity - and lack of better alternatives probably. Which is exactly the same as in this case (trophy shops are less popular than general sports shops anyway).

kocio-pl:
Interesting example from MapQuest Open: [shop=sports+sport=skying](http://www.openstreetmap.org/node/2182480589#map=13/52.2392/20.9746&layers=Q) is visible from z=13 and before I have checked it, I thought it's a grit bin. However under the closer inspection it looks like a side effect of trying to show [snow sports pitch](http://www.openstreetmap.org/way/284300930) without checking if it's a pitch at all.

math1985:
Unfortunately I don't think the current symbol is recognizable enough - test person also thinks it sells trophies rather than sport articles.

If anybody has a good idea, please let us know!

PR #1713, opened 2015-08-01
unify=path/footway styling, show surface for highway=path/footway/cycleway

### LINKED ISSUE #547: paths and tracks renderings are indistinguishable
Initial text on old tracker ticket   https://trac.openstreetmap.org/ticket/5163::
The rendering of highway=path and highway=track is barely distinguishable.
For example, many huge width highway=residential continue to an extra thin =track and then invisibly change to a =path.
Just like on my National Geographic maps, I suppose that highway=track should be drawn with a distinguishable width something like half that of residential highways.

your reply: I don't think path and track are barely distinguishable.

my reply: Yes they are.
Look at this part of the OSM map

stockis-which

and ask anyone, including yourself, which is the path and which is the track where vehicles can pass.
In case of doubt, look at the key:
track-1ab8881461027c46960a65123f176022  Track
footway-10e2ffa7697210623f36a18829831ec6  Footway 
Just like any sensible rendering like the IGN map shown below, the key makes tracks very distinguishable from foot, and more important, although I still maintain that tracks should be wider on the OSM map, as in reality.
But look carefully, <b>the map does not correspond to the key at all !!!</b>
<b>It even looks like the opposite</b>: smaller red dashes south is the track and longer black north is the path !!!
So that my request could also be stated "make the map like the key (and reality)" ;-)

 This is @ http://www.openstreetmap.org/way/281165129#map=18/50.52934/5.81330&layers=D

stockis osm marked

Look at Le Stockis : residential -> track -> path transitions in the circles where they meet.
The tracks go to a meadow and uphill through the forest.
Compare with the following IGN map.

stockis ign-1

Isn't it obvious on this IGN map which are tracks and which are paths? Even without a key.
The main hint of the importance of a way is its width, also dashes.
The color means little obvious, but they're fine to follow similar roads when the beginning is grasped.
And visibility when the roads are thin is best achieved with black on any background.

I made a bit of painting to simulate the difference:

stockis osm new

Isn't that is a more pleasant map to look at and grasp? !!!
And consequently much more pleasurable for the OSMers to map !!!

Please note that I don't claim I reached the optimal aesthetics: I just focused on visibility and differentiating what can be used by vehicle and which by foot(/bike) only.  The paths also need some touch up for visibility.

### LINKED ISSUE #1698: stop displaying highway=footway and highway=path differently
In theory highway=path may be nearly anything, in practise it is used mainly as

(1) equivalent of highway=footway; footway=unpaved
(2) equivalent of highway=footway
(3) with both foot=designated and bicycle=designated, interpreted as combined footway and cycleway (may or may not be segregated)
(4) to indicate that it used by pedestrians but this feature was not constructed with intention to be used by pedestrians
(5) with bicycle=designated, interpreted as equivalent of highway=cycleway
(6) rare features like mountain bike trails 

Overall difference between highway=path and highway=footway is minor compared to other features and huge difference in rendering is not justified.

Maybe it would be a good idea to render for example mountain bike trails differently but highway=path/footway difference is not allowing to detect cases like that.

Differentiation of styling would be better based on surface tag what would encourage proper mapping rather than selecting tag that renders like something paved/unpaved as it is done currently.

Such change would basically mean that attempt to use highway=path for highly varied features is considered as failed and it encouraged treating it as duplicate of highway=footway.

I tested my region (with highway=path typically used in tagging for renderer to indicate surface=unpaved) with unified rendering (with highway=path/footway using current style for path and later using current footway style for both). Overall, I was surprised by limited negative effect and how noticeable was improvement due to stopping of representing two nearly the same things using the same style. 

I think that unified to highway=path map was prettier and unified to highway=footway style it was more readable.

Code is available at https://github.com/matkoniecz/openstreetmap-carto/tree/to-footway and https://github.com/matkoniecz/openstreetmap-carto/tree/to-path

examples of before/after

to footway style:

https://cloud.githubusercontent.com/assets/899988/8944927/b5ffd00c-3583-11e5-81a6-763221f43786.png
https://cloud.githubusercontent.com/assets/899988/8944929/b6031c6c-3583-11e5-8ec7-704b6e51e39f.png
https://cloud.githubusercontent.com/assets/899988/8944928/b6023acc-3583-11e5-8bb7-46880d5adc1a.png
https://cloud.githubusercontent.com/assets/899988/8944930/b605ec62-3583-11e5-8dc7-78df3515619f.png

to path style:

https://cloud.githubusercontent.com/assets/899988/8944942/c3b777a4-3583-11e5-8670-d6d1bf57639e.png
https://cloud.githubusercontent.com/assets/899988/8944945/c3f81b92-3583-11e5-8afd-b5f7ffd98eeb.png
https://cloud.githubusercontent.com/assets/899988/8944944/c3f80d1e-3583-11e5-8d8b-b1b8c3ea2609.png
https://cloud.githubusercontent.com/assets/899988/8944946/c3f919b6-3583-11e5-83ea-1825a928840b.png

It is next part of road redesign that I think may be considered independently.

This idea was proposed in https://www.openstreetmap.org/user/Mateusz%20Konieczny/diary/35389 (section "highway=path, highway=footway problems") and discussed a bit in comments.

matkoniecz:
fixes #1698, fixes #547, implements beginnings and tiny part of #110

unification

proposed in this PR/current rendering
selection_002

paved/unpaved rendering

selection_003

selection_004

selection_005

Reasons for unification of highway=footway and highway=path are discussed in #1698 (massive difference between highway=path and highway=footway is confusing and is not useful).

Goal was to make difference between paved and unpaved noticeable but not overwhelming. Especially as paved footway/path may not be too prominent (problem in cities) and unpaved should be visible without problems (problem in remote areas).

surface is assumed to be paved unless explicitly described as unpaved ('unpaved', 'compacted', 'dirt', 'earth', 'fine_gravel', 'grass', 'grass_paver', 'gravel', 'ground', 'mud', 'pebblestone', 'salt', 'sand', 'woodchips', 'clay').

Fixes minor bug without effects for the current rendering.

https://cloud.githubusercontent.com/assets/899988/9023918/8ba15642-38b1-11e5-8ebb-623796c4f690.png
https://cloud.githubusercontent.com/assets/899988/9023919/8bb7ef24-38b1-11e5-8d1b-9f55a7849dbd.png
https://cloud.githubusercontent.com/assets/899988/9023921/8be01b34-38b1-11e5-99ef-c72743ded263.png
https://cloud.githubusercontent.com/assets/899988/9023920/8bd05b0e-38b1-11e5-8ac7-f3a387217bef.png
https://cloud.githubusercontent.com/assets/899988/9023922/8bf41fc6-38b1-11e5-86c8-282cffa0667f.png
https://cloud.githubusercontent.com/assets/899988/9023923/8bf452fc-38b1-11e5-9fcc-8fa14c64cef9.png

https://cloud.githubusercontent.com/assets/899988/9023942/c3f67990-38b2-11e5-9505-1ba6d79f8e03.png
https://cloud.githubusercontent.com/assets/899988/9023943/c3facf04-38b2-11e5-8202-592798d5571d.png

https://cloud.githubusercontent.com/assets/899988/9023945/ca79553a-38b2-11e5-9489-479dc4b0ac27.png
https://cloud.githubusercontent.com/assets/899988/9023944/ca74faee-38b2-11e5-8fdb-c4931a4e6c4f.png

https://cloud.githubusercontent.com/assets/899988/9023946/cda8f9f4-38b2-11e5-80b8-ab0c33df8c42.png
https://cloud.githubusercontent.com/assets/899988/9023947/cdaba154-38b2-11e5-89a3-79ed82c17915.png


project.yaml
```diff
@@ -742,6 +745,7 @@ Layer:
                 foot,
                 bicycle,
                 tracktype,
+                CASE WHEN surface IN ('unpaved', 'compacted', 'dirt', 'earth', 'fine_gravel', 'grass', 'grass_paver', 'gravel', 'ground', 'mud', 'pebblestone', 'salt', 'sand', 'woodchips', 'clay') THEN 'unpaved' ELSE 'paved' END AS surface,
```
pnorman: It would be better to use a name other than surface for this, as it's a normalized surface, and we may want to consider tracktype in the future.

pnorman:
I support unifying them, but long term we may want to look at if the dashed red is the clear cartography for a footpath.

I need take a good look at some local paths to see how it looks

jojo4u:
So a path without surface is considered paved?

pnorman:
> So a path without surface tag is considered paved?

Yes, that's the logic implemented. A more sophisticated logic would be to consider rural vs urban, but we don't have that data, and it's probably an undesired transformation, even if it does produce a better result.

jojo4u:
I'm a bit surprised to see more red dots instead of less. I hoped we could keep current path style for unsurfaced - but this won't fix #547. Over there it was proposed to make track wider.

math1985:
> I'm a bit surprised to see more red dots instead of less. I hoped we could keep current path style for unsurfaced - but this won't fix #547. Over there it was proposed to make track wider.

In the UK, this would lead to less red dots.

matkoniecz:
> I support unifying them, but long term we may want to look at if the dashed red is the clear cartography for a footpath.

From my looking at alternatives (I really dislike this red dot style and I wanted to replace it by something else) it works quite well. highway=path is also well recognisable but it is closer to already used style for highway=tracks.

> I need take a good look at some local paths to see how it looks

Yes, feedback from others would be really useful.

matkoniecz:
More about alternatives - there are styles that look better and/or are better at recognisability as footways but problems start once it turns out that rendering is too close to some other features (landuse=industrial, highway=service, boundaries, highway=track, landuse=retail etc).

project.yaml
matkoniecz: I am not sure what would be better name. quality? normalized_surface?

matkoniecz:
> I need take a good look at some local paths to see how it looks

And in case that somebody want to review some area and is not interested in setting up environment  for rendering ( https://github.com/gravitystorm/openstreetmap-carto/issues/657#issuecomment-56711313 ) I may generated some requested area.

project.yaml
gravitystorm: `int_surface`, for consistency - it goes with other `int_` prefixes e.g. `int_tunnel`, `int_tc_type`

matkoniecz: Done in https://github.com/matkoniecz/openstreetmap-carto/commit/898339d4c7137efcdbed7f60c312f280fff1f66b

See also https://github.com/gravitystorm/openstreetmap-carto/commit/cc17b038055aeab387af3e374abc953e0bb48f99 that split overly long lines.

polarbearing:
As most of us dislike the red dots, I'd see it more productive to unify towards the other end, the current path style, with slight differentiation depending on paving.

In the future we might then migrate the built-up urban footway towards the narrow white cased style that has already been proposed. 

Finally I wonder why we need to emphasize different traffic permissions on the general map so significantly, i.e. the blue for cycleways. This could be left to the Cycle map, while the general map focuses on the physical features.

matkoniecz:
The problem with path is that this style is close to highway=track. It is not as bad as #547 claims, but it makes harder to find way to distinguish between paved/unpaved.

go

> emphasize different traffic permissions on the general map so significantly, i.e. the blue for cycleways

I thought about unifying highway=bridleway to either highway=track or to highway=path/footway.

polarbearing:
> I thought about unifying highway=bridleway to either highway=track or to highway=path/footway.

Probably towards path/footway. In my area, I have never seen a dedicated bridleway as in [wiki](https://wiki.openstreetmap.org/wiki/Tag:highway%3Dbridleway#Germany:_Public_ways_vs._forests.2Ffarmland) in my area, they are all forestry paths or tracks that have a little horse=yes permission. As highway=bridleway gives us no clue if it is suitable for two-track forestry vehicles, chosing the path/footway is safer.

kocio-pl:
+1

jojo4u:
Bridleway defaults to motorcar=no and should not be suited for motorcars in width and smoothness. Track seems to be the wrong choice. To me bridleway is just some path which is designated for horses.

matkoniecz:
@math1985 @gravitystorm @everybody

Can you try how it works in your region?

pnorman:
Looks good here. Long term I'd like to look at a different styling, but that should wait till after unification as we shouldn't hold up the former for the latter.

matkoniecz:
I changed rendering that improves display of unpaved, especiall on dark surfaces and is nor amking harder to distinguish paved from unpaved.

Also, rebased.

_54 73019_-1 37278 master-_footwayify-show-surface 15 15 350px master - footwayify-show-surface 350px
_54 95349_-1 83316 master-_footwayify-show-surface 15 15 350px master - footwayify-show-surface 350px
_54 42548_-0 92952 master-_footwayify-show-surface 15 15 350px master - footwayify-show-surface 350px
_54 95349_-1 83316 master-_footwayify-show-surface 19 19 350px master - footwayify-show-surface 350px

https://cloud.githubusercontent.com/assets/899988/9157193/6cc8d18e-3ef4-11e5-9984-8afa65c4fe6a.png
https://cloud.githubusercontent.com/assets/899988/9157195/6cf18052-3ef4-11e5-922b-b2a4e5c6aeb9.png
https://cloud.githubusercontent.com/assets/899988/9157196/6cf264d6-3ef4-11e5-8ed7-00868a36ac1f.png
https://cloud.githubusercontent.com/assets/899988/9157194/6cf0b71c-3ef4-11e5-9f59-ef90764bae2a.png

math1985:
I think the rendering difference between footway and path is still too subtle to be useful, unfortunately.

kocio-pl:
Maybe the distance between the dots could be higher to make paths more diferent?

neiljp:
Just to be clear, following replies from @math1985 and @kocio-pl, the change currently planned is for footway/path to render the same, and the 'dot' difference visible in the images shown by @matkoniecz above are for paved/unpaved, right?

matkoniecz:
> are for paved/unpaved, right

yes

kocio-pl:
OK, so I like it already, but I'm still curious if bigger spacing between the dots would make the difference more obvious.

matkoniecz:
And here is version with bigger difference, as cost of visibility of unpaved ways.

_54 42548_-0 92952 master-_footwayify-show-surface 15 15 350px master - footwayify-show-surface 350px
http___www openstreetmap org__mlat 54 73044 mlon -1 37209 map 18_54 73044_-1 37209 master-_footwayify-show-surface 19 19 350px master - footwayify-show-surface 350px
_54 95349_-1 83316 master-_footwayify-show-surface 19 19 350px master - footwayify-show-surface 350px
_54 73019_-1 37278 master-_footwayify-show-surface 15 15 350px master - footwayify-show-surface 350px
_54 95349_-1 83316 master-_footwayify-show-surface 15 15 350px master - footwayify-show-surface 350px
everything in forest near kosice 15 19 master - footwayify-show-surface 300px

https://cloud.githubusercontent.com/assets/899988/9159325/9941a24a-3f26-11e5-932b-63328412cc3c.png
https://cloud.githubusercontent.com/assets/899988/9159327/99669e06-3f26-11e5-93d1-5f6fdbc0cb46.png
https://cloud.githubusercontent.com/assets/899988/9159326/99651072-3f26-11e5-991a-ec004eae0fcb.png
https://cloud.githubusercontent.com/assets/899988/9159328/9968e76a-3f26-11e5-905a-a48ccd27bd3a.png

kocio-pl:
Thanks for the test! Unfortunately this looks worse for unpaved - they are visible only in forest, on plain land or on the grass they loose much of visibility, so I would stay with your current proposition.

matkoniecz:
It seems that it will be necessary to either make paved/unpaved also in different colours, use current path rendering for unpaved, do not render paved status or just keep as it is now.

math1985:
What about rendering unpaved as current path and paved as current footway?

kocio-pl:
We could also use the current path style to suggest more solid surface than dotted line.

pnorman:
> What about rendering unpaved as current path and paved as current footway?

This gives rather different renderings for rather similar objects, which is not desirable.

neiljp:
Is there any cartographic conflict with having a dashed red line as an alternative to the dots, to distinguish between the surface type?
If no, for backwards compatibility my inclination would be to move to dashes for the unpaved.
However, for general consistency I would be more inclined to work to unify with the theme from track, namely that a 'more solid' (more dashed) line indicated a higher quality surface, which would suggest switching paved to dashed, and unpaved to dotted. Other than backwards compatibility, this also presents the issue that paved paths/footways often change shape/direction, making the dots more applicable, I think?

JBosm:
I may arrive late here, know that tagging footway/path is still a mess in OSM.
But, would considering footway as paved and path as unpaved by default be conceivable?

matkoniecz:
>  But, would considering footway as paved and path as unpaved by default be conceivable?

It would only support widespread tagging for renderer (selection of highway=path/footway is not supposed to imply anything about surface or quality).

vvoovv:
I strongly oppose this commit.
path and footway are totally different entiites according to OSM-wiki:

Quotes from the OSM-wiki:
**highway=path** is a generic path, either multi-use or unspecified usage, open to all non-motorized vehicles. The path may have any type of surface.
This includes walking and hiking trails, bike trails and paths, horse and stock trails, mountain bike trails as well as combinations of the above.

The tag **highway=footway** is used for mapping minor pathways which are used mainly or exclusively by pedestrians.

I kindly asking to revert your commit and discuss the issue with the largest international communities. Otherwise remapping of all outdoors area will be required.

polarbearing:
Could you explain what would have to be remapped into what, and why?

vvoovv:
Path should be used for outdoors according to the OSM-wiki. With this commit paths and footways are rendered in the same way.
To distinguish between path and footway an additional tag (surface) will be required. So every occurance of **path** must be complemented with surface tag. That what I meant under remapping.

vvoovv:
Again, path and footway are totally different entities according to the OSM-wiki.
They should be rendered differently.

math1985:
Unfortunately, mappers use different definitions to distinguish between footway and path, so we cannot draw any general conclusions about the difference between footway and path

> Otherwise remapping of all outdoor areas will be required.

This is not clear to me, what needs to be remapped?

> To distinguish between path and footway an additional tag (surface) will be required.

I'm not sure what you mean with that. Paths can be paved, and footways can be unpaved, so surface does not distinguish between footway and path.

pnorman:
> So every occurance of path must be complemented with surface tag.

To be usable by data consumers, yes, you should never use highway=path without surface and access tags.

@systemed's [what does the path say](http://www.openstreetmap.org/user/Richard/diary/20333) is one introduction to the problem, but certainly not the only one.

imagico:
As i have pointed out in #1748 the primary problem here is that the current rendering assumes highway=path without surface tag to be paved.  Is there any indication most of the ways tagged this way are actually paved?  If not assuming the opposite (unpaved by default) would be more sensible probably.

Some basic numbers: there are 3.9 million ways with highway=path of which 800k have a surface tag, about 200k have surface=paved/asphalt or other paving, the rest is unpaved of some sort.  My guess is that among the other ~3 million ways the unpaved:paved ratio is at least the same, i.e. 3:1.

Alternatively it would be possible to use a specialized rendering for highway=path without a surface tag indicating the missing information somehow.

vvoovv:
I agree with @imagico that **surface=unpaved** should be assumed by default for **highway=path**.

I refer to the OSM-wiki again.

_This includes walking and hiking trails, bike trails and paths, horse and stock trails, mountain bike trails as well as combinations of the above.
...
For urban paths which a designated for pedestrians only, better use highway=footway._

Just think logically. If **highway=footway** is for urban paths, then **highway=path** is for outdoor paths which are most likely unpaved. That's how I understand the article **highway=path** and how I have been used **highway=path** since 2008 when I joined OSM. I believe the same is true for the other mappers and that's confirmed by @imagico's statistics

dieterdreist:
sent from a phone

> Am 15.08.2015 um 23:44 schrieb math1985 <notifications@github.com>:
> 
> To distinguish between path and footway an additional tag (surface) will be required.
> 
> I'm not sure what you mean with that. Paths can be paved, and footways can be unpaved, so surface does not distinguish between footway and path.
> 


it is generally advisable to add surface tags to all highways, in particular the minor ones, because you hardly can find any global default because people assume different values to be "standard "

HolgerJeromin:
The josm preset for combined foot and bicycle ways has a German urban sign as an icon. It generates a path with some access tags but no default surface tag.
So the "path is unpaved" is not used by the editor.

Klaus-Tockloth:
> I strongly oppose this commit.

As already said: The default for path (typically encountered outdoor) is basically unpaved and the default for footway (typically encountered in urban areas) is basically paved. Please regard this.

polarbearing:
> encountered outdoor ... encountered in urban areas

The terminology is confusing here. Urban is outdoor well. The opposite of outdoor is indoor.

math1985:
I would be against rendering highway=path and highway=footway different based on the assumption that most paths are unpaved and most footways are paved. Rendering them differently propagates the assumption that data users can assume something about the surface of roads based on footway versus path.

imagico:
Numbers for highway=footway:

5.3 million total
840k with surface tag
\>600k with a paved surface tag (paved/asphalt/paving_stones etc.)
~200k with an unpaved surface tag

That is a >3:1 ratio for paved:unpaved, fairly symmetrical to highway=path.  I think it is a fairly reasonable idea to suggest based on these numbers that assuming a default of paved for highway=footway and unpaved for highway=path would be much better than any other default assumption in a two class system.

As already said the alternative would be to render all without surface tag in a different neutral styling that encourages the mappers to add missing information (a bit like highway=road).

systemed:
> The default for path (typically encountered outdoor) is basically unpaved and the default for footway (typically encountered in urban areas) is basically paved

I'm afraid it's not that simple. Most rural footpaths in the UK are unpaved and tagged as highway=footway. To make the urban/rural distinction you need a set of urban area polygons, which is what I do at cycle.travel/map but couldn't sensibly be applied to the osm-carto toolchain.

> As already said the alternative would be to render all without surface tag in a different neutral styling that encourages the mappers to add missing information (a bit like highway=road).

Good idea.

imagico:
> I'm afraid it's not that simple. Most rural footpaths in the UK are unpaved and tagged as highway=footway.

It seems to me by the way it is more that the boundary between use of highway=footway and highway=path is moved more to the rural domain than in other countries.  Mountain hiking trails seem equally mostly tagged as highway=path in the UK as they are elsewhere:

http://mc.bbbike.org/mc/?lon=-3.206108&lat=54.486092&zoom=13&num=2&mt0=osmfr&mt1=mapnik&marker=
http://mc.bbbike.org/mc/?lon=-4.664833&lat=56.317301&zoom=14&num=2&mt0=osmfr&mt1=mapnik&marker=
http://mc.bbbike.org/mc/?lon=-6.242101&lat=57.205202&zoom=14&num=2&mt0=osmfr&mt1=mapnik&marker=
http://mc.bbbike.org/mc/?lon=-3.638513&lat=57.092691&zoom=14&num=2&mt0=osmfr&mt1=mapnik&marker=

It is also natural that relative to highway=footway the number of these is much lower in the UK than in a country like Russia.

How meaningful this distinction is even when only considered locally is a different matter of course.

vvoovv:
The statistics supplied by @imagico is against this commit. The description of **highway=path** in the OSM-wiki is also against this commit.

So I'd suggest to revert the commit and continue the discussion to find a nice solution.

imagico:
I don't think there is need to rush things this way - the basic idea of rendering the 200k highway=path paved and the 600k+ highway=footway paved in the same styling and equally for the 200k highway=footway unpaved and the 600k highway=path unpaved is very good.  The question is what to do with the rest, this should be changed IMO but there is no need to make a snap decision here.  I suggest to open a new issue for that (since #1748 is primarily about a different matter).  If no solution is found in time for the next release it could be considered to render highway=path/footway without surface tag in the old path styling as a provisional measure - this would avoid the problems introduced by this change and keep the undeniable improvements it brought.

SomeoneElseOSM:
@vvoovv You can't trust the OSM wiki on this.  It's been edited by people with a personal point of view about "what the difference between a footway and a path is", and in a number of cases the current state of it only reflects the POV of the last editor, not the (lack of) concensus.

systemed:
> Mountain hiking trails seem equally mostly tagged as highway=path in the UK as they are elsewhere

The comparatively small number of mountains in the UK are not really the issue, because these are often 'access land' (= public allowed to walk across any area). In these places, it's fairly common that the paths worn by usage - rather than defined by statute - are indeed tagged as highway=path.

95% of the England & Wales landmass is not mountainous, yet has defined [Rights of Way](https://en.wikipedia.org/wiki/Rights_of_way_in_England_and_Wales) - paths, usually unmade, where there is a statutory right for people to walk. Many more of these are tagged as highway=footway than highway=path.

matkoniecz:
> As already said the alternative would be to render all without surface tag in a different neutral styling that encourages the mappers to add missing information (a bit like highway=road).

This was my initial idea - unfortunately finding two different styles for paved/unpaved was already quite problematic and failed to find a third style that would be also recognisable as footway and convey that data necessary to display it properly is missing.

> I would be against rendering highway=path and highway=footway different based on the assumption that most paths are unpaved and most footways are paved. Rendering them differently propagates the assumption that data users can assume something about the surface of roads based on footway versus path.

This was one of my main reasons for proposing and implementing this change.

> You can't trust the OSM wiki on this.

In that case - please, edit wiki. It is main tool to check both intended meaning of tags and real usage. Other methods are burdened with even worse distortions.

> The description of highway=path in the OSM-wiki is also against this commit.

Neither current or initial definition recommends assuming that surface is unpaved.

kocio-pl:
What about making such paths even weaker than unpaved somehow (paler color or bigger distance between the dots - much like [what you were testing lately](https://github.com/gravitystorm/openstreetmap-carto/pull/1713#issuecomment-129293859))? This could be a huge encouragement for adding surface tag! :stuck_out_tongue_winking_eye:

matkoniecz:
> What about making such paths even weaker than unpaved somehow

This may be a solution, but it would make large amount of paths poorly visible (in many places it would affect all of them).

Or maybe for missing data use paved/unpaved style randomly (for example - based on way id).

geowOSM:
I fully oppose the unification of the styling: path and footway are different in many aspects and should be distinguishable in Mapnik. Rendering of surface is a major improvement. But there are minor issues due to low contrast on certain background textures (i.e. scree, bare_rock) and the very small perceivable difference of paved vs unpaved should be strengthened, see example from http://osm.org/go/0IXkNEAbg-

path_sree_bare_rock

Also assuming a default of surface=paved if a suface tag is missing might be reevaluated. IMHO paved for highway=footway and unpaved for highway=path would be more reasonable as imagico mentioned before.

We might consider a revert, until a more considerate solution is found, based on a broader participation of the OSM community.

Some more thoughts:

highway=footway is undoubtly mainly a single-use way for pedestrians, whereas highway=path is a universal key for multi-use or non specific use ways, from rural, wilderness and mountain trails to all kinds of transport infrastrucure in the so called developing countries and may be used by all kinds of non 4-wheel traffic, including stock, mule, yaks and - yes - in some countries even by motorcycles/mopeds etc. 

We have 9 million paths and footways in OSM, the ratio of path:footway varies significantly according to country. These differences should be reflected in rendering.

Globally 1:1.4, 
France it's 1:1, 
Poland 1:3 and 
UK 1:5

This is partly because of national access restrictions, partly because of  the "duck tagging" mapping tradition, which is understandable and historically determined. In the UK (especially England and Wales) a highway with designation=public_footpath is mainly attributed as footway (partly even if it's physically a track).

gravitystorm:
> highway=footway is undoubtly...

As has been said many times already, there are so many different interpretations of the difference between footway and path that none of them are useful.

It is good, however, to see this reinforced, in that everyone who comes here to complain about this change and explain that they are different comes here with yet another (different) explanation than everyone else who complained before.

If anyone has any suggestions for smaller path/footway changes (either to the styling, or to the default surface assumptions, etc) then I would suggest more focussed new issues would be better.

mboeringa:
> As has been said many times already, there are so many different interpretations of the difference between footway and path **that none of them are useful.**

\+ 1

This actually just leaves two options:
1) - Re-instate the difference, so that people - ***regardless of interpretation of the tags*** - can see the difference and interpret in any personal way they like.
2) - A bold decision by you, as the repositories final owner, to support the current style change and not allow changes / revert *at this time*.

**Option 2 doesn't exclude future changes**, there have been continuous changes over the years after all. If anyone can come up with something good, preferably not just in words but in true visual proposals as @matkoniecz did, I guess we will see re-evaluation as in any normal situation in the past.

Any other discussions about the exact meaning of the two tags seems hopeless and senseless at best, as there will not be agreement on that topic any time soon...

imagico:
> This was my initial idea - unfortunately finding two different styles for paved/unpaved was already quite problematic and failed to find a third style that would be also recognisable as footway and convey that data necessary to display it properly is missing.

Here a basic test for how this could be done in a similar way as for tracks with solid line for paved and a mixed dash pattern for missing surface tag:

test area 1
test area 2

In addition this also uses stronger colors and thinner lines - this might require testing and tuning of course - it is probably a bit too strong as is but it would need more testing to see how much it can be tuned down without being difficult to read over bare_rock and other structured backgrounds.  I also tried using uniformly non-dashed thin lines at the lower zooms which would reduce clutter and address #1748:

test area 2

I don't have the time to further develop this, code is here (incomplete, only for footway so far):

https://github.com/imagico/openstreetmap-carto/tree/path-nosurface

mboeringa:
> Here a basic test for how this could be done in a similar way as for tracks with solid line for paved and a mixed dash pattern for missing surface tag:

Would it be possible to provide a kind of small mock-up quick and dirty legend of all the line types in the images, so that people joining the discussion from outside can understand the differences?

imagico:
Thought this is self explaining:

solid line: paved
short dashes: unpaved
mixed long/short dashes: no surface tag

kocio-pl:
I like this idea: mixed are better visible than not paved, but the pattern "noise" is unpleasant and I think it may encourage people to tag the surface.

The only possible problem I see is that it looks like typical boundaries representation, however this is not a deal breaker for me.

mboeringa:
> Thought this is self explaining:
> 
> solid line: paved
> short dashes: unpaved
> mixed long/short dashes: no surface tag

Thanks, I made my own mock-up. Please correct if I mis-interpreted something...

imagico_proposal

pnorman:
> If anyone has any suggestions for smaller path/footway changes (either to the styling, or to the default surface assumptions, etc) then I would suggest more focussed new issues would be better.

Given this is being ignored, I'm going to lock this issue. Please create a new issue to discuss a rendering for unknown surface like there is for unknown tracktype.

Issue #1706, opened 2015-07-30
Add rendering for shop=boutique and shop=fashion


kocio-pl:
My proposition for such shops (shop=boutique tag is quite popular, but shop=fashion is also established):
1) ![boutique-14-21](https://cloud.githubusercontent.com/assets/5439713/8990530/40137dd2-36f3-11e5-9ff9-f3e80a7b7a6d.png)

HolgerJeromin:
This was my first association:
https://www.maxima.at/uploads/pics/0_header_zopf.jpg
Do we really need a different icon from shop=clothes?

kocio-pl:
Funny. :smile: 

I think it is kind of specialized kind of clothes shop and rather easy to visualize. There's a reason why people tag it this way, just like we have hairdresser/beauty shop, convenience/supermarket or alcohol/wine/drink shops, which all are closely related or just are more specialized kind of the other. I also think we could separate shop with shoes for women if specific tagging scheme will be established.

imagico:
Not a good idea - visualizing things with icons always requires generalization of some sort and sometimes a pars pro toto approach.  But the symbol used should always be more fitting for the class of objects they are used for than for other things visualized in other ways.  If you use just a different piece of clothing for shop=boutique than for shop=clothes this essentially just emphasizes that there is no meaningful difference between these two tags.

There is no point in inventing new icons for various tags just for the sake of having a larger variety of icons, there should be a meaningful and matching difference in the tags definition, its use and the proposed icon.

kocio-pl:
For me personally this *is* meaningful difference - I wouldn't be interested in boutiques, while I may be interested in general clothes shop. And for some mappers this is also meaningful, because such tags exist and are used well over 5k times, which is a lot - isn't it?

If some sub-category is popular enough, it deserves it's own representation. Otherwise we could have only "food" shops instead of convenience/supermarket/bakery/fishmonger/confectionery/deli/greengrocer/meat/ice_cream (and it's probably is not everything). Showing just different food item would mean they are all the same - taking your argument. Categories are always just subjective choice and amount of uses at least gives them some legs.

imagico:
It would help if you'd actually explain the difference in meaning between shop=clothes, shop=boutique and shop=fashion, show that actual use of these tags represents this difference and propose icons that illustrates this difference.

The wiki [explicitly says](http://wiki.openstreetmap.org/wiki/Tag:shop%3Dfashion) the difference between these tags is unclear.

[...]

PR #1719, opened 2015-08-03
Adding shop=boutique icon

kocio-pl:
Resolves https://github.com/gravitystorm/openstreetmap-carto/issues/1706.
boutique-clothes

It looks like there is no way to really recognize the difference between fashion, boutique and general clothes shops. As shop=fashion is already displayed the same as shop=clothes, only shop=boutique rendering change was needed.


math1985:
As far as I'm concerned, shop=clothes is preferred over shop=boutique for boutique clothes shops, so I'm not sure whether we should encourage the shop=boutique tagging?

kocio-pl:
Hm, I'm not sure. This "preference" is only a suggestion on Wiki and a "discussion" on this topic is short and it rather supports the existence of such shops. 

I will open next thread on Tagging to make sure what community thinks about it now and will report the outcome, because maybe there will be some clue how to deal with it and it's always good to at least try to make things straight once in a while.

kocio-pl:
Seems I was already asking about it and [this is the only response I have received](https://lists.openstreetmap.org/pipermail/tagging/2015-August/025842.html). What do you think about it?

kocio-pl:
My feeling now is it's in most cases the same as fashion, so should be rendered the same - the number of uses is so high, that we shouldn't ignore it and let tagging dept resolve the issue of clothes/fashion/boutique real meaning. I leave it to your discretion if you merge this code or just close, because for now I can't see anything more we could do with it.

math1985:
I still think these shops would be better tagged as shop=clothes, so I'll closed this PR. Thanks anyway!

Issue #1420, opened 2015-03-18
Add special rendering for shop=musical_instrument


kocio-pl:
We should start rendering shop=musical_instrument, probably with this icon from osmfr-cartocss fork:

https://github.com/cquest/osmfr-cartocss/blob/master/symbols/musical_instrument.svg

[It was proposed in meta-issue #1402]

nebulon42:
This icon is already in use for `amenity=nightclub`.

kocio-pl:
I think it could be more specific - like guitar or piano. But wasn't it disco ball that we were trying to apply here or was it some other tag?

kocio-pl:
My proposition:
[1)](https://github.com/kocio-pl/openstreetmap-carto/blob/icons/symbols/musical_instrument-14-4.svg) ![musical_instrument-14-4](https://cloud.githubusercontent.com/assets/5439713/8336080/1f62a080-1aa2-11e5-8314-2ebde9d24400.png)

lest69:
I know what it's supposed to be, but I'm concerned it could be confused with other things (shoe, ice skate, cleats, etc.). Can you maybe try a guitar?

kocio-pl:
That was what I started with, but the result was rather poor, so I'm still trying:
2) ![musical_instrument-14-1](https://cloud.githubusercontent.com/assets/5439713/8354360/dabfb426-1b47-11e5-8580-79739e026ff7.png)
3) ![musical_instrument-14-3](https://cloud.githubusercontent.com/assets/5439713/8354373/e47098b4-1b47-11e5-965e-ad462aad7c20.png)

kocio-pl:
Another take:
4) ![musical_instrument-14-5](https://cloud.githubusercontent.com/assets/5439713/8360629/36a44a38-1b6e-11e5-93b2-d769d8204803.png)

kocio-pl:
Hm, maybe saxophone would do...

@nebulon42 Do you have any ideas or actual drafts for this shop type?

matkoniecz:
3) is the best from proposed ones

nebulon42:
I have no special ideas. I think any musical instrument could work. 2 or 3 maybe, with some tweaking.

daganzdaanda:
A trumpet is a good symbol, I think. Number 3) has no piston valves in the middle, maybe they could be added to reduce the "hole".

kocio-pl:
So the popular opinion is clear - I will try to work with the trumpet.

[...]

PR #1720, opened 2015-08-04
Adding shop=musical_instrument icon

kocio-pl:
Resolves https://github.com/gravitystorm/openstreetmap-carto/issues/1420.

musical_instrument


matkoniecz:
As mentioned in #1420 I prefer version 3 for icon (but here it seems that difference is smaller than from my impression there - as result of colour change).

kocio-pl:
7, 8 and 9 comparative field test (3 was too rough in SVG):
musical_instrument
musical-trumpet8
musical-trumpet9

I would choose 8.

matkoniecz:
I think that 9 is slightly better.

kocio-pl:
I just forgot to change the code because I didn't expect @math1985 will be so quick in this case, but when I look at it now, I still think 7 was the best (with 9 being the worst) and nobody else complained, so I'm happy with the outcome.

PR #1728, opened 2015-08-06
Use a random symbology for forests v1.1

### LINKED ISSUE #1724: Stop displaying natural=wood and landuse=forest differently
Current woods (natural=wood) are rendered as a solid dark olive green, and forests (landuse=forest) as a brighter light green with trees. Example: https://www.openstreetmap.org/#map=19/26.54415/-81.75739

The differences between the two are discussed here: https://help.openstreetmap.org/questions/324/when-should-we-use-landuseforest-rather-than-naturalwood. The primary conclusion being reached is that the traditional ways of choosing between the two are misleading and the tags are essentially used interchangeably. Some people use all forests, some use all woods, some use forests for managed areas and woods for unmanaged areas, and some just use them arbitrarily.

That being the case, it makes no sense to render the two differently when they essentially mean the same thing -- an area with dense trees that's not being used for agriculture. I reccomend that they be merged into a single style.

matkoniecz:
This is #1242 - rebased and with subtler pattern.

Hopefully I caught all bug introduced by rebasing.

I am considering splitting landcover-area-symbols into two layers, as SQL is quite ugly.

Forest with high path density:
https://cloud.githubusercontent.com/assets/899988/9140119/05e767d0-3d2f-11e5-94af-97258b955869.png

Other examples from #1242:
https://cloud.githubusercontent.com/assets/899988/9140108/fba4c524-3d2e-11e5-87c8-4f471b250074.png
https://cloud.githubusercontent.com/assets/899988/9140109/fba558a4-3d2e-11e5-9232-561593c5ff8f.png
https://cloud.githubusercontent.com/assets/899988/9140110/fba7bd7e-3d2e-11e5-8448-a68f8e70f1a6.png
https://cloud.githubusercontent.com/assets/899988/9140107/fba2a6e0-3d2e-11e5-9c82-95cb15932ee9.png

Fixes #1724


ximex:
i like [this](https://github.com/gravitystorm/openstreetmap-carto/pull/1242#issuecomment-127045263) pattern more but this one is ok for me too

imagico:
The wood/forest background and pattern color are fine but apart from that i have the following critiques/remarks:

* For the pattern i would prefer a single mixed symbol version as shown on #1242 - that would help probably especially at z=13.  It would also directly offer  a path to rendering leaf_type later.
* Your pattern does not look quite ideal with the symbols lining up on somewhat irregular curves and individual symbols having strongly varying distances to their different neighbours.  This is not bad and hardly visible in many cases but it is noticeable with larger areas.  If you share your parameters i can see if i can improve it.
* If you use `convert  +gamma -` (preventing imagemagick to generate png gamma information) for generating your preview pngs you can probably make them show up with correct colors in firefox (why firefox and chrome show this differently is a mystery to me).
* Your change to park color seems fine but the change to scrub is way too close to the new vineyard/orchard color introduced in #1655 - this was specifically selected to be well distinguishable from scrub.  In fact i don't think there is anything wrong with the current scrub color.  Note this also introduces yet another green since golf courses still use the old scrub color.

matkoniecz:
> For the pattern i would prefer a single mixed symbol version as shown on #1242 - that would help probably especially at z=13. It would also directly offer a path to rendering leaf_type later.

I agree that this pattern are prettier - and I would like to use them for showing leaf_type (this from PR - without leaf_type, singular mixed symbol - leaf_type=mixed, singular symbol for leaf_type=broadleaved/needleleaved, something else for leaf_type=leafless).

> If you share your parameters i can see if i can improve it.

Ops, it seems that I lost parameters. Anyway, it was something like http://www.imagico.de/map/jsdotpattern.php#x,512,jdp54442;g,30,32,32;rx,25,2,32,32;rx,25,2,32,32;s,jdp65794;s,jdp70741;rx,25,2,32,32;

> png gamma information

Thanks, I opened https://github.com/matkoniecz/CartoCSSHelper/issues/17 to remember about it (I would prefer to figure way to do it via rmagick)

> Your change to park color seems fine but the change to scrub is way too close to the new vineyard/orchard

OK, I will test it. I forgot about this change and that greens need to be retested.

imagico:
> I agree that this pattern are prettier - and I would like to use them for showing leaf_type (this from PR - without leaf_type, singular mixed symbol - leaf_type=mixed, singular symbol for leaf_type=broadleaved/needleleaved, something else for leaf_type=leafless).

As said i see no reason not to use a mixed symbol pattern universally for now and change to using it only for leaf_type=mixed later.  OTOH changing the pattern now to something suboptimal and then inevitably changing it again when leaf_type is rendered seems pointless.

For the pattern try

http://www.imagico.de/map/jsdotpattern.php#x,512,jdp6894;g,30,32,32;s,jdp33742;s,jdp81637;rx,250,2,32,32;s,jdp28824;s,jdp59702;s,jdp91550;s,jdp27774;rx,250,2,64,64;rd,1,0,0,tree%20pair,1,5,5,0,jdp48960;

kocio-pl:
In general - another big +1! I also like the way we care for additional extension, like suggesting leaf types etc.

And special big thanks for @pnorman for starting to fix this very complex subject.

matkoniecz:
> OTOH changing the pattern now to something suboptimal and then inevitably changing it again when leaf_type is rendered seems pointless.

My idea it to not change it once leaf_type is available and keep using it for forests where leaf_type is missing.

matkoniecz:
> For the pattern try

Something went wrong. I am getting file with trees clustered on borders. Can you maybe produce and share generated svg (with trees in Lch(70,30,135) colour)?

index

> scrub

The new scrub is a problem. I think that I will keep it unchanged - and it may be synchronized later (reverting scrub change is not making color distances problematic - it increases also distance from forest and it is not getting to close to other greens).

I tried to rebalance it again, without good results.

matkoniecz:
forest, scrub, orchard
https://cloud.githubusercontent.com/assets/899988/9135901/521afe7c-3d12-11e5-8f28-f912af9e42c7.png

scrub, orchard
https://cloud.githubusercontent.com/assets/899988/9135900/521a2f10-3d12-11e5-93de-d57f81752a6c.png

imagico:
For the pattern generator make sure you use the version on imagico.de and you reload all source files - i had changed some code fixing a bug that prevented you from changing the radius after generating the pattern.  This is not yet on github.

> My idea it to not change it once leaf_type is available and keep using it for forests where leaf_type is missing.

Even if for illustrating leaf_type the same symbols are used this does not appear to be the best solution to me. Using no pattern for no leaf_type seems the most intuitive approach.

gravitystorm:
> Using no pattern for no leaf_type seems the most intuitive approach.

No, that's not intuitive. We have far too many "plain greenish fill" polygons to be able to tell which is a forest/wood unless there are symbols.

matkoniecz:
> For the pattern generator make sure you use the version on imagico.de and you reload all source files

Thanks, now it is working!

kocio-pl:
I think now forest, scrub, orchard are different enough.

imagico:
> No, that's not intuitive. We have far too many "plain greenish fill" polygons to be able to tell which is a forest/wood unless there are symbols.

I meant intuitive in the way it clearly indicates something is missing in contrast to areas with leaf_type tagged that are rendered with a pattern.

OTOH if a paired symbol pattern is used in some areas and single symbol patterns in others this is less obvious, especially if the same pattern was used indiscriminately for all forest/wood areas before.

math1985:
> I meant intuitive in the way it clearly indicates something is missing in contrast to areas with leaf_type tagged that are rendered with a pattern.

I don't think that leaf_type is so important that we should give users of the map the impression something 'is missing' when forests are tagged without leaf_type.

imagico:
It was my understanding that the display of paired symbols is meant to create an even stronger indication of something being wrong than the lack of a pattern.  If there is another reason for choosing symbol pairs over mixed individual symbols please explain.

kocio-pl:
I've just noticed that [leaf_type has 796 665 uses](http://taginfo.openstreetmap.org/keys/leaf_type), while [natural=wood has 3 044 417 uses](http://taginfo.openstreetmap.org/tags/natural=wood) and [landuse=forest has 2 704 306 uses](http://taginfo.openstreetmap.org/tags/landuse=forest), so - roughly estimating - only 14% of popular wood tags has leaf_type assigned. 

While I think idea of plain green for unknown tree areas is a clever way of encouraging people to be more specific, this would be rather demanding. Some less interesting pattern will be also valuable feedback, while not making rendering of about 86% of tree areas worse (less obvious).

matkoniecz:
> It was my understanding that the display of paired symbols is meant to create an even stronger indication of something being wrong than the lack of a pattern.

I think that pair at the same time indicates that it is forest and will subtly encourages to add leaf_type once leaf_type will result in appearance of other patterns.

Lack of pattern for missing leaf_type would also encourage adding missing data but would fail in marking area as forest.

> only 14% of popular wood tags has leaf_type assigned. 

It is even worse as leaf_type is applied also to natural=tree etc. See http://taginfo.openstreetmap.org/tags/landuse=forest#combinations http://taginfo.openstreetmap.org/tags/natural=wood#combinations - both lower than 7%.

imagico:
> I think that pair at the same time indicates that it is forest and will subtly encourages to add leaf_type once leaf_type will result in appearance of other patterns.

I have my doubts this is going to be the case, especially for map users who got used to this being the normal forest/wood pattern used everywhere.  But you could be right, i can't be sure.

By the way the limited prevalence of leaf_type is partly due to the fact that this tag has not been around for long - there are probably still more maps rendering wood=* than there are rendering leaf_type.  There are still >350k uses of wood=* without a leaf_type tag.  But of course whatever styling is used for wood/forest without leaf_type will for quite some time surely be the dominating styling - especially of course in areas where forest data has been imported.

matkoniecz:
PR updated, examples updated (thanks @imagico for the new pattern! And the pattern generator!).

Forest with high path density:
https://cloud.githubusercontent.com/assets/899988/9140119/05e767d0-3d2f-11e5-94af-97258b955869.png

Other examples from #1242:
https://cloud.githubusercontent.com/assets/899988/9140108/fba4c524-3d2e-11e5-87c8-4f471b250074.png
https://cloud.githubusercontent.com/assets/899988/9140109/fba558a4-3d2e-11e5-9232-561593c5ff8f.png
https://cloud.githubusercontent.com/assets/899988/9140110/fba7bd7e-3d2e-11e5-8448-a68f8e70f1a6.png
https://cloud.githubusercontent.com/assets/899988/9140107/fba2a6e0-3d2e-11e5-9c82-95cb15932ee9.png

dieterdreist:
> Am 07.08.2015 um 17:26 schrieb Christoph Hormann <notifications@github.com>:
> 
> There are still >350k uses of wood=* without a leaf_type tag. But of course whatever styling is used for wood/forest without leaf_type will for quite some time surely be the dominating styling - especially of course in areas where forest data has been imported.
> 


there are also 225.000+ occurences of the (for non-relations deprecated) "type" key with leaf type values.
http://taginfo.openstreetmap.org/keys/type#values

pnorman:
> I think that pair at the same time indicates that it is forest and will subtly encourages to add leaf_type once leaf_type will result in appearance of other patterns.

That's why I preferred it. I looked **hard** for a single symbol that indicated a generic tree without anything about what type of tree, but could not find one.

HostedDinner:
It seems to me, the ordering of layers is not optimal. Now the forest is above lakes, which are not excluded by a multipolygon.
zwischenablage01

matkoniecz:
According to current tagging of this area it is both lake and covered by forest (what may happen in some cases like mangroves - https://en.wikipedia.org/wiki/Mangrove#/media/File:Hunting_For_Worms.JPG ).

In cases where it is not forested area it should be inner part of forest multypolygon.

There are also situations where forested area and locality named "Forest Foobar" are not equivalent (lake that is not part of forested area but is part of locality) but there is no established tagging for this situations.

DaveF63:
Does this new development include landcover=trees? 

I need to see it on the map to verify, but ATM the symbols appear a bit intrusive & the problem noted by  @HostedDinner needs to be sorted.

fkv1:
> I don't think that leaf_type is so important that we should give users of the map the impression something 'is missing' when forests are tagged without leaf_type.

I don't think that leaf_type is important enough to render it at all. Those who introduced a random symbology may be rightfully proud of it, but it should not be used for forests/woods. Just render them plain green without any symbols.

kocio-pl:
Leaf type is one of the most important property of forest/woodland and it's easy to implement without disturbing anyone, so I'd like to have it in this style.

fkv1:
> it's easy to implement without disturbing anyone

This is not true. I feel disturbed, and I am sure that many others feel the same.

Most official maps (such as http://www.amap.at) and commercial maps (such as http://maps.kompass.at) render forests plain green, and that's one of the reasons why they look considerably better than the OSM carto map.

kocio-pl:
You misunderstood me: I just meant implementing leaf types rendering will not make it any more disturbing than it is now (which BTW are less intrusive than what we had before). For me plain green would hide too much, I prefer subtle hint and this is what we have now.

imagico:
Fortunately there is no need to decide if this is important from an absolute standpoint, >300k uses quite clearly tells mappers consider it important and since displaying leaf_type is well possible to do in a relatively subtle way it is likely this will happen as soon as it is technically feasible here.  

If there are specific issues with the new woodland rendering i would suggest to open a new issue but i would strongly suggest first reading the previous discussion on #1242 as to not repeat arguments from there unnecessarily.

matkoniecz:
> I don't think that leaf_type is important enough to render it at all.

See #822 for discussion of this topic.

SK53:
@matkoniecz Historically wood=deciduous and wood=coniferous were rendered in the standard style (which is why this tag persists in the style sheet). Leaf_type is largely a semantically cleaner form of this distinction. In many parts of the world, largely those in temperate and boreo-temperate climatic zones, this distinction is an important aspect of interpreting the landscape from a map. Equally leaf_cycle also plays a part, but we have no history of rendering this distinction because wood=deciduous was treated (erroneously) as a synonym for broad-leaved woodland. As others have stated above, the various tags used to describe this are very heavily used: wood, 400k instances, leaf_type, 300k instances, type, 200k instances (I exclude leaf_type on nodes, and of course these usages are not additive). I take this to mean that there is considerable interest in collecting and displaying this type of information.

Woodland covers about 23% of the world's land area: there are good cartographic reasons why a slightly finer division of types of woodland is worthwhile. More detailed division is clearly something for more specialised projects, but I don't think it is reasonable to dismiss out-of-hand leaf_type and leaf_cycle. This is particularly true now that the hard work of determining suitable ways to add symbols for woodland is nearing fruition.

fkv1:
> Leaf_type is largely a semantically cleaner form of this distinction.

Maybe, but it's botanically unfounded, at least when it comes to whole woods or forests. E.g. in central Europe, you encounter broad leaved plants from sea level up to the alpine zones in all natural plant communities. Even in spruce forests, you'll see one or the other naturally grown young beech. And reversely, you'll hardly find any forest without any Pinacaea, because so many of them are planted or spared. So there's essentially no needle_leaved or broad_leaved forest. You could rather tag the percentage at best, like needleleaved=70%.

I suspect that most wood=* come from imports, and also that most wood=* and leave_type=* on areas are incorrect.

> In many parts of the world, largely those in temperate and boreo-temperate climatic zones, this distinction is an important aspect of interpreting the landscape from a map.

This is wrong, because the foliage is only visible from a distance of a few metres.

> Woodland covers about 23% of the world's land area: there are good cartographic reasons why a slightly finer division of types of woodland is worthwhile.

How about different shades of green? We could recycle the one we used for natural=forest. The fact that woodland covers big parts of the world suggests NOT to use any symbols. Symbols should only be used for areas that are rare and therefore not obvious to the viewer.

Just compare woodland to water, residential areas etc.  Imagine what would happen if all water areas were ornamented with wave symbols, and all residential areas with house symbols, etc.  You would not recognize any streets, POIs, nothing. You would be overwhelmed by the ornaments. That's what we experience within woodlands now.

@imagico:
I disagree with you for the above reasons, but speaking of #1242, I think that the rendering you suggested on 2015-01-19 would at least be much better than the one we have right now, first of all because your rendering is less dense, and second because the tree pairs we have right now look like toilet symbols.

Issue #300, opened 2013-12-29
Support some marine tagging


sabas:
(generic issue for discussion)

Remembering the discussion about OpenSeaMap tagging, it's possibile to support some tags for harbours and ports?
This wouldn't influence greatly the current rendering because the features to be handled are found only in limited areas.

For example, how about starting with rendering port areas (other than marinas)? (http://wiki.openstreetmap.org/wiki/OpenSeaMap/Harbours#Harbour_Categories)

In this area the only visible landuse is that tagged with landuse=industrial
http://www.openstreetmap.org/#map=15/44.4047/8.8932

Issue #1716, opened 2015-08-02
Add rendering for landuse=depot


kocio-pl:
I think we should start rendering landuse=depot. It has only [about 500 uses](http://taginfo.openstreetmap.org/tags/landuse=depot), but it's one of those features which never will have much higher number of uses (because it's limited in real life, of course) and is in fact just a subcategory of landuse=industrial, so we can just make it look exactly the same.

polarbearing:
Sorry I disagree. Depending if the public transport stores their buses, or the chemical company their raw materials, it is landuse=commercial or industrial. There are certainly more depots in the world than 500, compare the 70000 mapped townhalls. In the wiki, landuse=depot is just a proposal with little support.

matkoniecz:
I consider landuse=depot as a terrible tagging and I created competing proposal at http://wiki.openstreetmap.org/wiki/Proposed_features/*%3Ddepot

kocio-pl:
OK, so we can safely close this issue.

matkoniecz:
Closing on request from reporter.

kocio-pl:
I like your proposition in general, because it better uses existing categories, however it looks it would be problem when tagging tram depot: how on should railway=tram + railway=depot? Probably you need some other description, like: railway:tram=depot.

matkoniecz:
@kocio-pl Can you ask on tagging mailing list or on wiki? I will reply there.

kocio-pl:
Sure, done on wiki.

PR #1731, opened 2015-08-08
Add landuse=port same as landuse=industrial

sabas:
Simple addition of the landuse=port tag (http://wiki.openstreetmap.org/wiki/Tag:landuse%3Dport)

Same as #437 but as a pull request, original idea proposed in #300


matkoniecz:
> Files project.mml and - differ

Also, I think that accepting this would encourage fragmentation of landuse. I think that it is similar situation as #1716.

Most likely landuse=industrial, industrial=port or landuse=commercial, commercial=port (depending on a port type) would be a better tagging.

matkoniecz:
I added industrial=port to http://wiki.openstreetmap.org/wiki/Proposed_features/industrial#Industrial

imagico:
Although the tag page on the wiki mentions the tag is for land areas only current use frequently covers water which cannot currently be rendered in a reasonable way.

Apart from that i partly share the concerns of @matkoniecz - a port is not necessarily an industrial area.  It could make sense to better render port features in a distinct way (in a similar way as airport features are rendered for example) but this is technically challenging, especially from the water side.

sabas:
@matkoniecz "commercial" isn't needed imho (as in available to general public): I left out ferries and marinas because are already covered by other established tags.

My test area is http://overpass-turbo.eu/s/aQx
As you can see I added cargo=* to a ferry_terminal although unnecessary.

I could retag if the combination landuse=industrial + industrial=port is a better fit (as the majority of this tag was added by myself)

matkoniecz:
From view of data consumer landuse=industrial + industrial=port is far better - one of the most frustrating things in using OSM data is discovering that there is yet another way to express the same thing.

With such cascading tagging scheme one may support just landuse=industrial and ports will also be handled. And somebody interested in special display for port areas may check key industrial.

But it is better to continue this discussion on tagging mailing list or on wiki - http://wiki.openstreetmap.org/wiki/Talk:Proposed_features/industrial or http://wiki.openstreetmap.org/wiki/Talk:Tag:landuse%3Dport

pnorman:
Low number of uses, tagging needs working out still.

I'm skeptical about adding *another* landuse tag, as this makes consuming data more difficult.

dieterdreist:
sent from a phone

> Am 08.08.2015 um 20:40 schrieb Mateusz Konieczny <notifications@github.com>:
> 
> Most likely landuse=industrial, industrial=port or landuse=commercial, commercial=port (depending on a port type) would be a better tagging.


a lot of ports will be industrial, (not sure for passenger oriented ports), commercial doesn't seem right for a port. Anyway, rather than looking at subtags for attributes like landuse we should focus on main tags like man_made=port or similar

mboeringa:
> **From view of data consumer landuse=industrial + industrial=port is far better - one of the most frustrating things in using OSM data is discovering that there is yet another way to express the same thing.**
> 
> With such cascading tagging scheme one may support just landuse=industrial and ports will also be handled. And somebody interested in special display for port areas may check key industrial.

+1, this is a major issue, especially with only marginally documented tagging schemes. However, even with better documented tagging schemes, the issue of multiple tagging schemes remains a headache from the point of data consumers, like you say. The SQL statements or filter expressions needed to extract what are essentially the same objects, can get really complicated to the point of being completely incomprehensible for novice users.

dieterdreist:
sent from a phone

> Am 08.08.2015 um 20:54 schrieb Christoph Hormann <notifications@github.com>:
> 
> Apart from that i partly share the concerns of @matkoniecz - a port is not necessarily an industrial area. It could make sense to better render port features in a distinct way (in a similar way as airport features are rendered for example) but this is technically challenging, especially from the water side.


I agree.

If mappers were starting to adopt this proposal it would mean moving to a different way of tagging. Until now we had quite rough landuse attributes (at least for the main built up uses like residential industrial etc) and used keys like man_made and amenity for features. If I read this proposal correctly it suggests to use a landuse subtag to say eg something is an oil mill.

My suggestion is to deprecate the proposal in its current form and use the industrial key  to subtag industrial landuse on a less specific level (eg heavy industry vs light industry, producing vs storage, etc.)

matkoniecz:
It is better to continue this discussion on tagging mailing list or on wiki - http://wiki.openstreetmap.org/wiki/Talk:Proposed_features/industrial or http://wiki.openstreetmap.org/wiki/Talk:Tag:landuse%3Dport

PR #1733, opened 2015-08-09
new rendering for natural=heath, fixes #780

### LINKED ISSUE #780: Improve natural=heath color
The following issue has been moved over from [trac](https://trac.openstreetmap.org/ticket/3578):

Nyah. I think that current heath color is inharmonious. I propose to replace the color on #ddebbb (see the bottom picture).

matkoniecz:
london 17 17 master - heath 300px

https://cloud.githubusercontent.com/assets/899988/9154970/cf434d00-3ea7-11e5-8f68-6e309755692e.png
http://www.openstreetmap.org/?mlat=51.44040&mlon=-0.23186#map=17/51.44040/-0.23186


matkoniecz:
It may be a good idea to also make scrub symbols bigger or in other way better imply that this is describing features like  (from https://github.com/gravitystorm/openstreetmap-carto/issues/780#issuecomment-129184704)

imagico:
Would be strongly against it as explained in https://github.com/gravitystorm/openstreetmap-carto/issues/780#issuecomment-129184704

All other pattern variations in this style illustrate smaller differences between areas with common major properties.  natural=scrub and natural=heath have no more in common than natural=scrub and natural=wood or natural=heath and natural=grassland.

Please don't turn this into a style showing major ecological differences only as minor footnotes while showing subtle and often non-verifiable differences in urban context in dozens of different tones.

matkoniecz:
>  natural=scrub and natural=heath have no more in common than natural=scrub and natural=wood or natural=heath and natural=grassland

Thanks! I thought that natural=scrub and natural=heath are closer than these other natural areas.

matkoniecz:
I had new version (that has better differentiation).

london 17 17 master - heath4 300px

imagico:
It could be worth considering, if this is useful for interaction with other colors, to make heath somewhat less brown as long as it is still sufficiently different to be well recognizable.  You would also need to consider that wetland=bog uses the same base color and here it is important for it to not appear as some intermediate between marsh and swamp.  The brownish color is fairly good here to indicate the low nutrient environment and relatively sparse vegetation typical for bogs.

#d2dea3 would seem suitable.

And a pattern could be worth considering for the higher zooms anyway to both clarify the meaning and possibly to differentiate different types of heathland (although there is currently no tagging for that).

kocio-pl:
Interesting idea, especially using small symbol to show how close heath is related to scrub in terms of size - ecological difference may be symbolized by the color (if I understand correctly, this is what @imagico suggests in the last post, but I'm not sure).

jeisenbe:
@matkoniecz, do you still have the icon svg that was used to make this heath pattern? 

I would like to try it with some other colors for #780

matkoniecz:
No, but it was reuse of one of icons. Probably one used in the style at this time.

Issue #1705, opened 2015-07-30
Showing some POIs earlier when outside the landuse=residential


kocio-pl:
I took this idea after thinking for some time that while we have problems with some [crowded places](https://github.com/gravitystorm/openstreetmap-carto/issues/1560), we have also much more places where the problem is quite reverse, because they have much less POIs; not because it's not fully mapped, it's just reality. Because of that POIs are generally more important there - in itself and as the landmarks.

Look for example at [this restaurant/inn](http://www.openstreetmap.org/way/362815861): it is rendered from z=17, but for me even z>=13 would be much better.

Of course maybe separate tagging scheme is needed for such objects located by the highways, but I have noticed that it should be easy to tell that it's rural place, because it's not surrounded by landuse=residential area!

Other good examples of useful POIs in some remote places could be:
- health facilities (like clinics)
- convenience shops
- [crosses and shrines](https://github.com/gravitystorm/openstreetmap-carto/issues/131#issuecomment-142845653)
- memorials/[monuments](http://www.openstreetmap.org/way/288351923)
- shelters
  and many others.

I realize that landuse=residential tag is not as universally present as it probably should be, but this is real chance to dynamically adjust rendering in different places all over the world. We can start with some carefully chosen POIs as a test drive to get some experience, make detailed analysis which ones and how many such POIs are rendered outside residential area and maybe tune minimal distance between the POIs, but I think this is promising idea for the future.

dieterdreist:
sent from a phone

> Am 30.07.2015 um 17:53 schrieb kocio-pl <notifications@github.com>:
> 
> I realize that landuse=residential tag is not as universally present as it probably should be, but this is real chance to dynamically adjust rendering in different places all over the world.


from everything you write I believe you mean the place area and not landuse=residential

StyXman:
Gas stations and pharmacies come to mind too.

kocio-pl:
@StyXman Of course - thanks!

@dieterdreist Sure, we may exclude both in case there are some residential landuses outside place areas. We can also exclude some other human-related areas, like industrial maybe, or most everything not being natural (but including farmlands maybe)...

The truth is I don't know exactly what areas to include/exclude and which POIs are more or less suitable to experiment with. We need to start with good research of current status anyway, because it's probably really outside-the-box idea (at least in OSM rendering - I guess GIS-aware users could know this trick already, but for me it was unexpected discovery).

matkoniecz:
[...]

PR #1734, opened 2015-08-09
consider airport size in rendering decisions

### LINKED ISSUE #1143: Too many tiny private airports on the map at small scales
screen shot 2014-11-30 at 11 12 31 pm
Any suggestions on how to handle lots of tiny private airports? It doesn't look like there is a way to differentiate between tags but perhaps only airports with an ICAO or IATA ref should appear at the z10ish.

Perhaps only areas or relations with the `aeroway=aerodrome` could be shown. Any way to handle this programmatically?

matkoniecz:
consider airport size in rendering decision, includes mall change to island label rendering

small airports are rendered later, small and medium airports are not losing labels too early
fixes #1143

world z 10 10 _ master -_ airport-unify 39 6316 -86 0889 10 10 master - airport-unify 250px

https://cloud.githubusercontent.com/assets/899988/9156855/a24e9d86-3ee8-11e5-9cd8-c2bb1290cf98.png
https://cloud.githubusercontent.com/assets/899988/9156857/a273b5da-3ee8-11e5-8303-813d9d4c95ef.png
https://cloud.githubusercontent.com/assets/899988/9156858/a276731a-3ee8-11e5-89d9-a6827dcf4d12.png
https://cloud.githubusercontent.com/assets/899988/9156860/a277224c-3ee8-11e5-8b3f-6bd158a9d9e0.png
https://cloud.githubusercontent.com/assets/899988/9156859/a276d4b8-3ee8-11e5-90f9-335d131d9d8d.png
https://cloud.githubusercontent.com/assets/899988/9156856/a266fa16-3ee8-11e5-9587-476a198c77e8.png

### DIFF
--- a/amenity-points.mss
+++ b/amenity-points.mss
@@ -16,6 +16,8 @@
 @landcover-wrap-width-size-bigger: 45;
 @landcover-face-name: @oblique-fonts;
 
+@max-labeling-area-in-pixels: 900000;
+
 @standard-wrap-width: 30;
 
 /* Note that .points is also used in water-features.mss */
@@ -873,7 +875,9 @@
     marker-fill: @airtransport;
   }
 
-  [feature = 'aeroway_aerodrome'][zoom >= 10][zoom < 14] {
+  [feature = 'aeroway_aerodrome'][zoom >= 14][way_pixels < @max-labeling-area-in-pixels],
+  [feature = 'aeroway_aerodrome'][zoom >= 14][way_pixels = null],
+  [feature = 'aeroway_aerodrome'][zoom >= 10][way_pixels > 5][way_pixels < @max-labeling-area-in-pixels] {
     marker-file: url('symbols/aerodrome.svg');
     marker-placement: interior;
     marker-clip: false;
@@ -1024,10 +1028,10 @@
 /* Note that .text is also used in water.mss */
 .text-low-zoom[zoom < 10],
 .text[zoom >= 10] {
-  [feature = 'place_island'][zoom >= 7][way_pixels > 3000][way_pixels < 800000],
-  [feature = 'place_island'][zoom >= 16][way_pixels < 800000],
-  [feature = 'place_islet'][zoom >= 14][way_pixels > 3000][way_pixels < 800000],
-  [feature = 'place_islet'][zoom >= 17][way_pixels < 800000] {
+  [feature = 'place_island'][zoom >= 7][way_pixels > 3000][way_pixels < @max-labeling-area-in-pixels],
+  [feature = 'place_island'][zoom >= 16][way_pixels < @max-labeling-area-in-pixels],
+  [feature = 'place_islet'][zoom >= 14][way_pixels > 3000][way_pixels < @max-labeling-area-in-pixels],
+  [feature = 'place_islet'][zoom >= 17][way_pixels < @max-labeling-area-in-pixels] {
     text-name: "[name]";
     text-fill: #000;
     text-size: 10;
@@ -2242,7 +2246,9 @@
     text-wrap-width: @standard-wrap-width;
   }
 
-  [feature = 'aeroway_aerodrome'][zoom >= 10][zoom < 14] {
+  [feature = 'aeroway_aerodrome'][zoom >= 14][way_pixels < @max-labeling-area-in-pixels],
+  [feature = 'aeroway_aerodrome'][zoom >= 14][way_pixels = null],
+  [feature = 'aeroway_aerodrome'][zoom >= 10][way_pixels > 5][way_pixels < @max-labeling-area-in-pixels] {
     text-name: "[name]";
     text-size: 8;
     text-fill: darken(@airtransport, 15%);


imagico:
Note this would defer all airports that are not mapped as areas (that is 3/4 of them - and not necessary the smallest 3/4) to z=14+.  This could lead to fairly pointless drawing of airport polygons just to make them show up in the map.

dieterdreist:
sent from a phone

> Am 09.08.2015 um 23:39 schrieb Christoph Hormann <notifications@github.com>:
> 
> This could lead to fairly pointless drawing of airport polygons just to make them show up in the map.
> 

why is it pointless to define the area?

matkoniecz:
> and not necessary the smallest 3/4

From what I checked all really important airports are marked as areas.

amenity-points.mss
```diff
@@ -16,6 +16,8 @@
 @landcover-wrap-width-size-bigger: 45;
 @landcover-face-name: @oblique-fonts;
 
+@area-too-large-to-label-size-in-pixels: 900000;
```
pnorman: A shorter variable name would be good. `max-labeling-area`?

amenity-points.mss
```diff
@@ -852,7 +854,8 @@
     marker-fill: @airtransport;
   }
 
-  [feature = 'aeroway_aerodrome'][zoom >= 10][zoom < 14] {
+  [feature = 'aeroway_aerodrome'][zoom >= 14][way_pixels < @area-too-large-to-label-size-in-pixels],
+  [feature = 'aeroway_aerodrome'][zoom >= 10][way_pixels > 5][way_pixels < @area-too-large-to-label-size-in-pixels] {
```
pnorman: 5 seems *very* small, does it actually work well?

amenity-points.mss
matkoniecz: max-labeling-area-in-pixels?

amenity-points.mss
matkoniecz: I am not too sure as I failed to find place with many minor airports tagged as an area - almost all are added solely as nodes.

matkoniecz: But it should fairly easy to modify - and I preferred to give too low value rather than too big.

matkoniecz:
I see that sending code late in night was a poor idea - airports rendered as nodes were never rendered. This is now fixed.

The same situation is with already existing code for place=island* (but I am not opening issue for that - I think that it is valid method to encourage proper mapping, island should not be mapped as nodes).

*http://overpass-turbo.eu/s/aRz

imagico:
> why is it pointless to define the area?

If an airfield consists of nothing more than the runway and maybe some shed or storage building there is no verifiable geometry to map except for those.  Drawing a polygon aeroway=aerodrome here would be a mere abstract design, in terms of semantic data it would be pointless work.

Examples:

http://mc.bbbike.org/mc/?lon=-150.395392&lat=70.429607&zoom=16&num=2&mt0=bing-satellite&mt1=mapnik&marker=
http://mc.bbbike.org/mc/?lon=-164.555411&lat=67.733584&zoom=15&num=2&mt0=google-satellite&mt1=mapnik&marker=
http://mc.bbbike.org/mc/?lon=-133.027696&lat=69.435128&zoom=15&num=2&mt0=google-satellite&mt1=mapnik&marker=
http://mc.bbbike.org/mc/?lon=-163.336014&lat=65.902401&zoom=15&num=2&mt0=bing-satellite&mt1=mapnik&marker=

A polygon for the airport as a whole only makes sense if that airport has a defined outline, like in form of a fence or at least a maintained surface (regularly mown grass etc.)

Note we are talking about relative importance here - like in case of populated places it makes little sense to only display airports of globally high importance at z=10.  Many of the ones listed above would for example all make sense to show at z=10 since they are hundreds, often thousands of pixels away from the nearest other airfield at that scale.

kocio-pl:
I think more usable metrics would be airports density for example. For some remote places even smallest airport could be the most important point and should be rendered early - it somehow similar issue to https://github.com/gravitystorm/openstreetmap-carto/issues/1705, but I'm not sure if using landuse+city borders would be suitable for airports.

dieterdreist:
sent from a phone

> Am 10.08.2015 um 09:28 schrieb Christoph Hormann <notifications@github.com>:
> 
> If an airfield consists of nothing more than the runway and maybe some shed or storage building there is no verifiable geometry to map except for those. Drawing a polygon aeroway=aerodrome here would be a mere abstract design, in terms of semantic data it would be pointless work.


well, make a multipoligon with those 2 areas? We were discussing here (as I understood it) major civil airports currently mapped just with a node (and parts ). These cases almost everytime are fenced and thus have a very precise border.

cheers 
Martin

matkoniecz:
> A polygon for the airport as a whole only makes sense if that airport has a defined outline, like in form of a fence or at least a maintained surface (regularly mown grass etc.)

In cases like that it may be really hard or impossible to map airport area from satellite/aerial images but I am pretty sure that there is a defined area findable by a local survey.

imagico:
You can always find some outline but the next mapper finds it completely different and that is the very definition of non-verifiable.

My main point here is this path is clearly chosen out of convenience for the rendering side - way_area is easily available.  As i pointed out in #1143 the runway length would be a much more suitable importance measure - nearly all airports have a runway mapped and it is much less arbitrary and less subject to unrelated influence factors than the airport polygon area.

Of course this is complicated to implement and could have a notable performance impact - still i don't think any of this is a good reason to tell the mappers to change a perfectly good mapping practice (i.e. map airports as nodes unless there is a verifiable delineation of the airport area).

pnorman:
The root cause is that a single grass strip and Heathrow are both tagged with aeroway=aerodrome. This didn't used to be the case, and there used to be a hierarchy of airports.

matkoniecz:
> The root cause is that a single grass strip and Heathrow are both tagged with aeroway=aerodrome.

Yes, I was seriously considering to propose new milestone "low quality of OSM tagging scheme" and move #1143 to it. Or closing and marking it with label like that.

pnorman:
> Or closing and marking it with label like that

Closing as invalid or wontfix is most appropriate where the tagging doesn't provide enough information.

Rovastar:
@imagico
    still i don't think any of this is a good reason to tell the mappers to change a perfectly good mapping practice (i.e. map airports as nodes unless there is a verifiable delineation of the airport area).

I disagree. I don't consider the old school techniques of mapping things of a giving (and verifiable) area as so instead of a node as not good practice. In fact it is the opposite.

10 years ago a node was more than sufficient for mapping now it is consider a bit old school and mapping the area is both more useful and detailed. e.g. we don't have buildings as nodes if possible any more. So we "tell" mappers to do this all the time.

Another way of looking at this is if an airport is mapped as a way/area then do you change it back to a node?! I am not sure the usefulness of that. Onwards and upwards.

Also if we did have a sort of airport density I would not help. You could easily have xyz tiny airport displayed but the larger and more important one not displayed. Density algorithms only work if all the points are of the same "importance" which is not the case here. 
True, we had different (better) tagging before no pointing in crying over that split milk. Getting the perfect system here is not on the cards the best we can sensibly get is to use way_area.

kocio-pl:
> Also if we did have a sort of airport density I would not help. You could easily have xyz tiny airport displayed but the larger and more important one not displayed. Density algorithms only work if all the points are of the same "importance" which is not the case here. 

I didn't mean density as the only measure - rather density as the first rule _and then_ if we have too much airports, we can eliminate them the smart way (only more important/bigger are shown). This might be exactly "considering airport size in rendering decisions". 

I'm highly skeptical of depending _entirely_ on the size. What about [Haasts Bluff Airport](http://www.openstreetmap.org/node/1042093888) in the middle of nowhere or [more than other 1400 small airports in Australia tagged as a node](http://overpass-turbo.eu/s/aTY) then? They are the most important POIs there probably and should be probably visible from z8+ (or even z7+), but let's at least not make them worse.

We really need some smart algorithms for dealing with such a diversity we have in the world, size alone is not enough.

ximex:
@kocio-pl :+1:

jojo4u:
As long as there is no sane tagging style I'm all for a simple differentiation in the style. Size seems to be a sensible choice here.

kocio-pl:
The example is crowded with airports, but not overcrowded, so it will be a nice fix (not something essential) and I guess there may be at most hundred such places in the world, while degradation for rural areas will be noticeable and it will be more than thousand (I have tried only Australia). So I think the overall performance will be worse.

talllguy:
@matkoniecz was this implemented? Just checking. I see that #1143 is still open.

PR #1738, opened 2015-08-12
fix outdated comment

matkoniecz:
It affects also current version. Residential now appears from z12. I am not 100% sure what will be in osm_planet_roads - link in http://gis.stackexchange.com/questions/37099/in-osm2pgsql-how-is-the-planet-osm-roads-table-populated/84913#84913 is dead (@pnorman, maybe you can update it?). According to my check tertiary is not appearing in osm_planet_roads (at least it happens this way for my local databases) and it is rendered from z10.

### DIFF
--- a/roads.mss
+++ b/roads.mss
@@ -852,7 +852,7 @@
 /* Data on z<10 comes from osm_planet_roads, data on z>=10 comes from
 osm_planet_line. This is for performance reasons: osm_planet_roads contains less
 data, and is thus faster. Chosen is for zoom level 10 as cut-off, because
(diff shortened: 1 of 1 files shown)


math1985:
> I am not 100% sure what will be in osm_planet_roads 

See https://github.com/openstreetmap/osm2pgsql/blob/master/style.lua - all tags for which is_road is true in zordering_tags.

matkoniecz:
Thanks, also thanks to @pnorman for [updating dead link](http://gis.stackexchange.com/posts/84913/revisions).

Issue #506, opened 2014-04-26
Add rendering for amenity=community_centre


nhsavage:
These are potentially very useful to people, so it possible, could these please be rendered? Thank you.

http://wiki.openstreetmap.org/wiki/Tag:amenity%3Dtownhall
http://wiki.openstreetmap.org/wiki/Tag:amenity%3Dcommunity_centre

math1985:
Removed townhall from this issue, duplicate with #77.

polarbearing:
Many community_centres have a campus around the building. That should render, e.g. in "educational" yellow like kindergarten/school.

dieterdreist:
2015-06-01 13:13 GMT+02:00 polarbearing <notifications@github.com>:

> Many community_centres have a campus around the building. That should
> render, e.g. in "educational" yellow like kindergarten/school.




This is very likely depending on the cultural setting, but I haven't found
an indice to associate community centres with "education" from reading the
wiki:
http://wiki.openstreetmap.org/wiki/Tag:amenity%3Dcommunity_centre
and I also wouldn't have thought about education from my personal
expectations (at least not as "main" field).

I believe they are more places for celebrations, and maybe for some forms
of politics (public participation etc.)

kocio-pl:
I think we should have an icon for it (like two people talking or sitting at the table) and at least render the name for the area. Does it sound reasonable?

polarbearing:
Name and icon certainly helps. Area fill helps further to show dimensions. 

@dieterdreist - I did not mean a community_centre is always educational, I just compared it. Might have better said amenity-yellow, though I know there is some debate about limiting its use. 

I have seen some family of multi-generation oriented centres here, i.e. social meeting places.

kocio-pl:
My drafts of the icon:
1) ![community_centre-14-2](https://cloud.githubusercontent.com/assets/5439713/8946061/88317be4-358d-11e5-900f-15b6d7027a7f.png)
2) ![community_centre-14-5](https://cloud.githubusercontent.com/assets/5439713/8946066/95a0214a-358d-11e5-8378-f7b3ad88ccdd.png)
3) ![community_centre-14-7](https://cloud.githubusercontent.com/assets/5439713/8946068/9c4c3920-358d-11e5-9647-00fed37eaf46.png)
4) ![community_centre-14-10](https://cloud.githubusercontent.com/assets/5439713/8946077/af3f5f9e-358d-11e5-8d35-6b578d5d52f1.png)

[...]

PR #1744, opened 2015-08-14
Adding amenity=community_centre icon

kocio-pl:
Resolves https://github.com/gravitystorm/openstreetmap-carto/issues/506.

Just slightly bigger heads than in version 6) to make it more readable:
community_centre-big-heads


nebulon42:
Might be confused with school?

kocio-pl:
I guess that this is not your personal opinion and it's just a question if that can be misunderstood? I think the chances are the same as confusing bank icon with money exchange or night club icon with a concert hall - it is possible in theory, but if it's not a problem for more people in practice, I would not be afraid.

PR #1747, opened 2015-08-15
update preview to include both power area change and path/footway unification

matkoniecz:
Maybe it should be done once per release, including it in PRs turns out to not be working (or maybe it is just recetly that there are multiple PRs affecting this area).


pnorman:
The travis failure is a travis error, not a build failure

Issue #1642, opened 2015-07-08
Change bank and ATM color to @amenity-brown?


kocio-pl:
I think [bank and ATM should have the color of all other amenities](https://github.com/kocio-pl/openstreetmap-carto/tree/bank-color). Yes, they are special somehow, but still they are more like amenity (or maybe closer to shop?), while I think black is better for generic things, like name=* on a building. I also think about displaying craft=* POIs, so less color diversity of POIs would be better.

![bank-brown](https://cloud.githubusercontent.com/assets/5439713/8571794/026ad316-2589-11e5-9e71-51471d768b0e.png)
![atm-brown](https://cloud.githubusercontent.com/assets/5439713/8572082/9e52ab68-258a-11e5-99d2-9037f20e1d59.png)

It was first mentioned [here](https://github.com/gravitystorm/openstreetmap-carto/pull/1558#issuecomment-118952635).

HolgerJeromin:
seems ok. We could delete the whole variable `@money-icon` as it is not used after this change.

kocio-pl:
Sure, fixed now.

math1985:
Good idea, I'm slightly in favour the retail colour though.

kocio-pl:
That's how it would like like then:

![bank-shop-saski](https://cloud.githubusercontent.com/assets/5439713/8584183/68058ec8-25d4-11e5-9ce9-c69c678a0475.png)
![bank-shop-rotunda](https://cloud.githubusercontent.com/assets/5439713/8584248/f71cceaa-25d4-11e5-8e17-7c6b854823c6.png)

matkoniecz:
I think that @amenity-brown is better - at least for me "bank is an amenity" makes more sense than "bank is a shop",

I prefer @amenity-brown to the current state and the current state to purple.

math1985:
@matkoniecz Giving banks the same colour as shops would make it easier to recognize entertainment areas versus retail areas though. Most of the brown icons are entertainment, and all of the pink items are shops. Banks fit rather in the latter category.

matkoniecz:
OK, I see that it would be nice. Anyway, I have no strong opinion here. I still slightly prefer brown - but current or purple also would be OK.

math1985:
@matkoniecz Giving banks the same colour as shops would make it easier to recognize entertainment areas versus retail areas though. Most of the brown icons are entertainment, and all of the pink items are shops. Banks fit rather in the latter category.

[...]

PR #1759, opened 2015-08-17
Changing color of hunting_stand to brown

kocio-pl:
I think amenities are more than only for entertainment (see also https://github.com/gravitystorm/openstreetmap-carto/issues/1642#issuecomment-120550415) and black should be left for general or maybe also for less visible things:

hunting-brown

### DIFF
--- a/amenity-points.mss
+++ b/amenity-points.mss
@@ -938,7 +938,7 @@
 
   [feature = 'amenity_hunting_stand'][zoom >= 16] {
     marker-file: url('symbols/hunting-stand.16.svg');
-    marker-fill: @man-made-icon;
+    marker-fill: @amenity-brown;
     marker-placement: interior;
     marker-clip: false;
   }


matkoniecz:
Readability gets worse for stands displayed on brown features (tracks, farmland - typical for hunting stands).

kocio-pl:
I see no problems with readability in this example: they are clearly visible for me on farmland (just like the tracks are, probably even better because they are solid) and even on tracks they are recognizable (even details of the shape are not blending with them completely and one can at least spot them).

imagico:
Within this style hunting stands are mostly a landmark like water towers or masts and not an amenity so this argument does not really catch.  Note in most jurisdictions entering a hunting stand without authorization is not allowed so it does not really qualify as a type of shelter either (which would be blue anyway).

kocio-pl:
> Within this style hunting stands are mostly a landmark 

Sure, this is why I wanted them to be visible early enough.

>Note in most jurisdictions entering a hunting stand without authorization is not allowed so it does not really qualify as a type of shelter either (which would be blue anyway).

Access type is not really associated with object type - just like private playgrounds (they are rendered paler BTW).

I'm not going to force this idea anyway - it's nothing fundamental, I just thought this color would be more suitable for this kind of POI. While I still feel visually it works good enough, the argument about being a landmark in a context of general style is more interesting and may stand its ground.

kocio-pl:
@imagico is right I think, so closing it now.

dieterdreist:
sent from a phone

> Am 23.08.2015 um 13:25 schrieb Christoph Hormann <notifications@github.com>:
> 
> Note in most jurisdictions entering a hunting stand without authorization is not allowed so it does not really qualify as a type of shelter either (which would be blue anyway).
> 


I bet nobody would complain if you used a hunting stand for shelter in an "emergency", not sure if it would even count as an offense in these conditions. Then, which countries do you base the "most jurisdictions" assumptions on, is this a global estimate?

And lastly it is up to the individual to decide whether they want to take the (IMHO neglectible) risk of being caught in your wrongdoing of entering them without authorization. 

FWIW, in my youth I've climbed tens or hundreds of hunting stands and literally never have had any problem whatsoever...

PR #1761, opened 2015-08-17
Move car rental to zoom level 17

### LINKED ISSUE #1752: Zoom level for car rental
I think the car rental icon should be shown at the same level as common shops, which would be zoom level 17 instead of 16.

sb12:
resolves #1752


math1985:
Thanks!

PR #1762, opened 2015-08-17
Move car sharing to zoom 18

### LINKED ISSUE #1751: Zoom level for car sharing
I think the car sharing icon is shown too early. It is currently shown from level 16, which is the same level as supermarkets, churches, museums etc.
It should be moved to zoom level 18, as it is usually not more than a few designated parking spaces.

sb12:
resolves #1751


math1985:
Thanks!

PR #1763, opened 2015-08-17
Make major power lines less prominent

nebulon42:
I think major power lines are slightly too prominent. This was amplified by the recent colour change of forest. I made them a bit lighter and slightly thinner.

z14 (left = before)
power_z14

z16
power_z16

### DIFF
--- a/power.mss
+++ b/power.mss
@@ -1,8 +1,16 @@
+@power-line-color: #888;
+
 #power-line {
   [zoom >= 14] {
-    line-width: 1;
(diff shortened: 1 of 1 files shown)


matkoniecz:
I would consider keeping old width for really high zoom levels (18+? 19+?)

HolgerJeromin:
as the color is two time used, this is probably a candidate for a variable.

nebulon42:
Added both suggestions.

PR #1767, opened 2015-08-18
Render cutline as thicker line with some green color

kocio-pl:
Taking over https://github.com/gravitystorm/openstreetmap-carto/pull/1526.

My take is that we can use "scrub green" as the color for cutlines, which works in many ways:
- it doesn't look like a road any more...
- ...which gives us the opportunity to make the cutline wider than before (to be visible under footway/path)...
- ...and looks more like typical cutline surface (low green plants, like scrub), while not using scrub symbols (because we're not sure it's really scrub, not the grass or something else)

For comparison: the vertical line is not a cutline, but real scrub area about 10 m wide.

z14:
cutline2-14

z16:
cutline2-16-wide

z19:
cutline2-19

### DIFF
--- a/landcover.mss
+++ b/landcover.mss
@@ -542,12 +542,15 @@
 /* man_made=cutline */
 #landcover-line {
   [zoom >= 14] {
-    line-width: 1.5;
+    line-width: 3;
     line-join: round;
     line-cap: square;
-    line-color: @land-color;
+    line-color: @grass;
     [zoom >= 16] {
-      line-width: 2;
+      line-width: 6;
+      [zoom >= 18] {
+        line-width: 12;
+      }
     }
   }
 }


imagico:
scrub color does not seem such a good idea - it implies something specific that is not generally the case.  Especially firebreak cutlines are generally kept free of higher vegetation to fulfill their purpose.

And i would continually increase the width over the zoom levels rather than doing so in larger steps.

kocio-pl:
1. I guess in the current state of the map every color at least suggests something... See - the problem was the land color (which is supposed to be neutral!) looks like a road for some people.

2. Scrub is hardly "higher" to me (that would be the trees).

3. We could also use "grass green", to suggest not even middle vegetation by default. But I'd rather not like to use any new green/brown/whatever color. Slight suggestion of something typical for cutlines is better for me than introducing yet another shade of the rainbow.

4. Which values do you like for which zoom level? I can test it if you don't feel like doing it.

imagico:
Your view is probably somewhat too Europe-centric.  In dry woodlands with little underbrush firebreak cutlines will usually either be bare ground or with very sparse non-woody vegetation.  Also cutlines might exists within scrubland which would be invisible with your suggestion.

Choose whatever width you consider appropriate but having a big step between z15 and z16 but otherwise constant width does not seem ideal.

kocio-pl:
It may be true, of course (I'm European and I don't have any degree in plants and forests), but let's try to keep an eye on possible solution:
1) what color do you propose then - and the most important question:
0) do you think is it possible to find any color at all if cutlines can look so different?

I think scrub color can be just interpreted as "here we are still in the forest, but its lighter", not only as actual scrub or any land cover. I'm reading lack of symbols as a visual hint - it's a suggestion, not the claim "this IS the scrub" (rather: "it MIGHT BE scrub").

gravitystorm:
Perhaps the colour can be defined in terms of the forest colour, e.g. @cutline-color: lighten(@forest-color, X%). This prevents any implication that there is scrub, and it makes it more logical to customise (e.g. if a downstream user wants to make scrub a brown colour or change forest colour). On the other hand, that introduces yet-another-green.

kocio-pl:
Great, nice suggestions, thanks! Now it would be good to tell what's our preferences and which solution we don't like or even if something is out of question for anybody (and why probably). It's kind of "voting" time (if nobody will come with something else to consider, of course). 

I also don't feel it's important to add more width variety, so it's up to you (@imagico), to tell what you want exactly - otherwise I will just leave it like that, because I don't think it's important enough, but it adds more code.

imagico:
This is not about what i want, i just suggested that a more continuous increase in width might be considered better by most map users.  I suggest you try it out and if you are unsure ask for opinions based on examples.  For color i think the most logical default would be land-color.

W.r.t. coloring you should also consider previous discussion in #1449.

kocio-pl:
Land color makes no sense since it looks misleading for some people, but your own proposition - `mix(@land-color, @forest, 50%)` - sounds interesting to me, is similar to @gravitystorm proposition and is even more neutral than scrub color. I can test it, but I hope this is not what we consider "transparency" (because Andy is against using this effect)?

pnorman:
You shouldn't be jumping from 3 to 12 in one zoom level.

kocio-pl:
You may be right, but any more detailed hints are highly welcome, please, because I don't really feel this is a real issue, so I'm a wrong person to find solution for what others treat as a visible problem. I can do the testing however.

math1985:
One reason for the current style is the suggestion that cutlines might appear in landuses other than forest. I didn't check whether it occurs in practice, but this style wouldn't really work for cutlines in scubland.

kocio-pl:
Mix of grass + land could be good enough then, but I have to check if there are such places at all (already mapped or maybe not). However definition on Wiki is clear that it's only for typical tree areas.

imagico:
All right - i thought the line width thing was quite clear but maybe it isn't so:

When you draw a three pixel wide line at z14 (where you have a nominal pixel size of ~10m) that is about 30 meter wide nominally on the map (to be multiplied by scale factor depending on latitude but that's not important here).  If you render it in the same three pixel at z15 that is about 15 meter obviously.

Now at z16 you render the line 12 pixel wide which is - with now a pixel size of ~2.5m - 30 meter wide.

So as you zoom in you get a line 30m 15m 30m 15m 7.5m ... wide.  It should be obvious that this is going to look strange.  The usual approach is therefore to have a line width gradually decreasing with every zoom level in map units and increasing in pixel width keeping it wide enough at the lowest to be well visible and at either a pleasant width for this kind of feature or its real world width at the highest.

kocio-pl:
I have choosen 12 px at higher levels because it makes clear that a footway/path is surrounded by cutline area (6 px is not wide enough) and this is important information. On lower zoom levels I feel 3 px is enough to show the cutline at all. Real width could be rendered when mappers put the numbers in tags.

Should we make different width for every zoom level from z14 to z19 or maybe its enough to have it only for 14/16/18+?

imagico:
Use your discretion, try it out testing on different zoom levels in different settings and see how it looks.  You need to get a feeling for it - that is what style design is about. :-)

kocio-pl:
That's exactly the point: it's hard to get the feeling of fixing if you don't feel it's broken in the first place. I'm not against fine tuning this idea, but I simply can't guess exactly what will work for you too, because my design goals (which I just described) are already resolved. That's why I ask for your suggestions. "Guess what I'd like" is rather poor communication strategy. :wink:

imagico:
I tried to explain why a big step from 3 to 12 pixel with otherwise constant width is not working well - i think it should be possible for you to see that - maybe try a different sample area where there is more other stuff mapped around, generate a full sequence of renders from z14 to z18 and see how it looks.

I have no specific suggestions what width to use for what zoom level - i have not tried it and i am not even sure i think rendering cutlines is a good idea.  I am just trying to help with suggestions what is likely going to work well and what is likely not going to work well.

kocio-pl:
Thank you all for your help, suggestions and patience! Unfortunately I was overwhelmed a bit lately, so I was not focused enough and willing to try myself every detail, but here is updated proposition (grass color with just additional width=6 step at z16):

z14:
cutline-grass-14
z16:
cutline-grass-z16
z18:
cutline-grass-18

I have tried 50% mix of land/grass and land/scrub, but that is not working - it still can be confused with a road. Here I have used grass color, because it is safer in case there are some cutlines in the bush maybe (however current cutline definition doesn't say anything about such possibility and I had no idea how to test it). Although if there is a way to exclude existence such objects and we don't like to have it in the future, I would be happy also with scrub color.

lest69:
Is there any way to have the cutline be rendered on top of any wood/forest icons, thus hiding them? By it's very nature, a cutline shouldn't have trees in it, so it looks odd to have the trees render on top of it.

kocio-pl:
I read it like a hint "this is a part of the forest", not that there are really some trees. It also make a cutline different than standard grass area, which is good for me (as I wrote [in the rationale](https://github.com/gravitystorm/openstreetmap-carto/pull/1767#issue-101697186)).

pnorman:
> Is there any way to have the cutline be rendered on top of any wood/forest icons, thus hiding them?

Nothing really comes to mind.

matkoniecz:
Maybe with some comp-op operations ( https://www.mapbox.com/tilemill/docs/guides/comp-op/ ) it would be doable? #1107 seems kind of similar.

math1985:
I haven't had time to look at this yet.

math1985:
>  Is there any way to have the cutline be rendered on top of any wood/forest icons, thus hiding them?

Can we not simply use a layer on top of the icons?

In any case I think this is an improvement with respect to what we have now, so I'm going to merge this.

PR #1771, opened 2015-08-19
Handle landcover for power=plant

jengelh:
https://wiki.openstreetmap.org/wiki/Key:power does not
describe power=station, but power=plant.

plant (http://osm.org/way/111582223) however is currently not rendered with the purple background that goes along with e.g. substation (http://osm.org/way/364738201)


landcover.mss
```diff
@@ -327,6 +327,7 @@
     [way_pixels >= 64] { polygon-gamma: 0.3;  }
   }
 
+  [feature = 'power_plant'][zoom >= 10],
```
pnorman: Is power_plant fetched by the SQL?

malcolmh:
Areas tagged power=plant should have at least a boundary rendering plus display of its name. Otherwise mappers will continue to use power=generator in order to force a rendering, a tag that does not make sense for the site rather than the individual generator units.

kreed:
It might make sense to render plants the same as landuse=industrial (on all zooms). It would make the generators and substations inside the plant easier to spot.

Here is a comparison:
download
download 1

Also, you should add power=plant to amenity-points.mss so that the name is rendered.

matkoniecz:
@jengelh - Can you respond to comment by @kreed?

jengelh:
Turns out to be more work than I had envisioned.

Issue #108, opened 2013-08-09
Add rendering for amenity=* and office=*


jremillard:
Hi,

Any point that has an amenity or office tag and a name tag, should at least get its named rendered even if we don't have an icon at max zoom. 

Most of the POI's in the mall are not rendered.

http://osm.org/go/ZfJD0RleX--
- [x] bench #194
- [x] public_building (won't fix)
- [x] bicycle_parking #591
- [x] waste_basket #418
- [x] townhall #77
- [X] parking_space #776 (won't fix)
- [x] hunting_stand #966
- [ ] vending_machine #1561
- [x] fountain #705
- [x] social_facility #1776
- [x] car_wash #1777
- [x] marketplace #751
- [x] dentist
- [x] waste_disposal #419
- [x] community_centre #506
- [x] taxi #879
- [x] fire_hydrant (won't fix)
- [x] veterinary clinic #1656
- [ ] parking_entrance #270
- [ ] nursing_home
- [x] arts_centre #2520
- [x] nightclub #456

math1985:
@nebulon42 Do you have any suggestion for a dentist icon? The SJJB icon @nebulon42 ![dentist](http://www.sjjb.co.uk/mapicons/png/amenity_dentist.p.20.png) is not that nice.

kocio-pl:
I also vote for the tooth icon with nothing else.

math1985:
I added the most used ones (>10000 occurrences) to the top of this post. For most of them, finding a suitable icon is the toughest task.

math1985:
I propose not to render amenity=fire_hydrant (too specialistic).

mkoniecz:
Main problem with amenity=fire_hydrant is that displaying would make sense on z22 or z21. And that zoom levels currently are not used anywhere (AFAIK - it is not going to change in near future).

mkoniecz:
BTW, the same problem is with amenity=bench - though Humanitarian style displays benches on z19.

AnBuKu:
Am 14.01.2015 um 08:25 schrieb Mateusz Konieczny:
> Main problem with amenity=fire_hydrant is that displaying would make
> sense on z22 or z21. And that zoom levels currently are not used
> anywhere (AFAIK - it is not going to change in near future).

This is the only one I know with z21
http://www.osm.ch/#21

But, sure, emergency=fire_hydrant should not be rendered in the main map
(even if I'm a fanboy of fire hydrants :-)

[...]

PR #1776, opened 2015-08-22
Adding amenity=social_facility icon

kocio-pl:
Implements part of https://github.com/gravitystorm/openstreetmap-carto/issues/108.

I think we should render amenity=social_facility [the same as amenity=community_centre](https://github.com/gravitystorm/openstreetmap-carto/pull/1744) in general. Once we have social_facility=* key available, we can create also some more specific icons (for shelters or food-related places, for example).

social_facility-19


matkoniecz:
I think that for this two amenities just label would be better.

kocio-pl:
Hm, since [more than 93% such objects have a name](http://taginfo.openstreetmap.org/tags/amenity=social_facility#combinations) it is perfectly acceptable solution, however I also know the place which is a mix of few different social facilities [in one building with the enclosing area](http://www.openstreetmap.org/way/39230878), so it would be still not visible. 

Your proposition would only change the color for social facilities enclosed in one building to brown (about 19%) and start to show the names of the rest. Even in the worst case (assuming that all the buildings already have names) we would show 74% currently invisible such objects, so it would be quite effective.

I still like this icon to be used, but I'm not going to argue if more people doesn't like my POV.

BTW: or we could also use some borders with name labels, maybe? This would show all the social facilities, irrespective of them being named or not, but I'm not sure how the border should look like and if it is not overkill in general (but for some places like orphanage it would be natural to show the area, like we do for schools).

More opinions and ideas would probably help.

kocio-pl:
After some discussion on #660 now I feel that displaying the icon with text is preferable over just the label, [because it makes this style more universal (icons are language independent) while not obscuring other things (high zoom level is not crowded)](https://github.com/gravitystorm/openstreetmap-carto/issues/660#issuecomment-135128493).

polarbearing:
Label and icon please. Some objects can have fancy names which do not reveal what they are. I mapped one called "The Gap".

kocio-pl:
I've just made a special draft for this type of facility:
1) social_facility-14-29

polarbearing:
@kocio-pl - good one. Could be seen as 'helping hand' as well as 'cocooning' which both leads to the intended association.

math1985:
I haven't had time to look at this yet.

kocio-pl:
I think new icon really makes sense and it's definitely better to have something different than community centre icon if possible:
social-facility-hands-19

polarbearing:
Progress in this issue would be appreciated, does anything speak against the icon?

math1985:
When I see the community centre icon on the map, it keeps reminding me of a school rather than a community centre? Am I'm the only one in that?

Also, I'm not sure if the current icon expresses for example a nursing home very well?

kocio-pl:
Looks like nursing home is tagged differently ([amenity=nursing_home](http://wiki.openstreetmap.org/wiki/Tag:amenity%3Dnursing_home)).

polarbearing:
social_facility 40000 currently. Nursing home is the older style for tagging a subclass of social_facility.

We should discuss the community_centre icon in the other issue, and focus her on the later proposal by @kocio-pl Sep 24-28.

polarbearing:
@math1985 - it is not clear what icon your comments on Nov 9 refer to. 

The PR currently implements the 'helping hand' icon from @kocio-pl, not the community centre one.

mboeringa:
> I've just made a special draft for this type of facility:
> 1) alt text
> 2) alt text

Nice ones, I think I have a slight preference for the hands which bend inwards. This looks more natural, and makes the hands both more recognizable as such, and more "caring".

kocio-pl:
@mboeringa I took 1) because 2) looks blurred for me and I personally have more trouble recognizing the shape alone, but I was trying to achieve exactly what you say.

math1985:
I'm still not extremely fond of the icon, but I think it's better than nothing, so I'll accept this PR.

math1985:
Sorry, I didn't see the icon was changed with respect to the preview. New icon looks fine to me!

PR #1777, opened 2015-08-22
Adding amenity=car_wash icon

kocio-pl:
Implements part of https://github.com/gravitystorm/openstreetmap-carto/issues/108:

car_wash-19

BTW: it looks like shop=car_repair should be rendered also in brown, because it doesn't sell things, but services, so it is in fact amenity too!


HolgerJeromin:
nice icon. It is surprising with the brown, but you are right.

kocio-pl:
We have also taxi and car sharing icons, which are transportation-blue, because their function is transporting people (which is a special kind of service, if we look closer), so we have the car icon in different colors. :smiley:

matkoniecz:
@kocio-pl  Can you rebase it?

kocio-pl:
Sure, done (and checked that it renders as before).

kocio-pl:
...and once again.

PR #1778, opened 2015-08-23
Changing color of shop=car_repair to brown 

kocio-pl:
As mentioned in https://github.com/gravitystorm/openstreetmap-carto/pull/1777: shop=car_repair is in fact amenity, so it should be rendered in brown, since it sells services rather than goods:

car_repair-brown-19

### DIFF
--- a/amenity-points.mss
+++ b/amenity-points.mss
@@ -678,7 +678,7 @@
     marker-file: url('symbols/shopping_car_repair.16.svg');
     marker-placement: interior;
     marker-clip: false;
-    marker-fill: @shop-icon;
+    marker-fill: @amenity-brown;
   }
 
   [feature = 'shop_bicycle'][zoom >= 17] {
@@ -1996,6 +1996,9 @@
       text-halo-fill: rgba(255, 255, 255, 0.6);
       text-wrap-width: @standard-wrap-width;
       text-placement: interior;
+      [feature = 'shop_car_repair'] {
+        text-fill: @amenity-brown;
+      }
     }
   }
 


matkoniecz:
Seems like a good idea.

math1985:
Under the same logic, hairdresser should be brown too. Not sure if I agree.

kocio-pl:
Right, thanks for noticing! We have some fundamental ambiguity here (how shop differs from amenity), so I will resort to Tagging to discuss it a bit first.

matkoniecz:
I am not sure whatever this belongs on tagging - in that case tagging is clear, problem is with murky distinction in using brown and purple in this style (for some cases it is clear, for some it gets problematic).

kocio-pl:
For me this alone would be a minor issue, but tagging amenity=car_wash while at the same time having also shop=car_repair is not "clear". It indicates lack of general understanding what are really those categories, because I can't find the rule why one is considered to be a shop and the other one an amenity. Hairdresser/beauty shop is also interesting in this regard.

kocio-pl:
[My questions about shop/amenity on Tagging](https://lists.openstreetmap.org/pipermail/tagging/2015-August/026102.html).

gravitystorm:
> It indicates lack of general understanding what are really those categories,

You can think of the tags as opaque key/value pairs. It's not important which are `amenity=` and which are `shop=`. It is left to the editors and renderers to create whichever categories they think are useful - especially important, since there is not a One True Way to categorise the world!

So the discussion that happens here - how we want to display the features - is independent of both the precise tags that are used, and also independent of any discussion on the tagging mailing list. We should chose an answer based on our own considerations - colour, icon style etc, as well as how many categories we think is appropriate and what those categories should be.

kocio-pl:
The discussion has helped me understand the problem deeper, see my conclusions here: https://github.com/gravitystorm/openstreetmap-carto/issues/1783. In a nutshell: we can rely on the tagging system as a base for our colors if we think it's useful, but we can also create our own color classification (which I prefer).

dieterdreist:
sent from a phone

> Am 24.08.2015 um 14:53 schrieb Andy Allan <notifications@github.com>:
> 
> You can think of the tags as opaque key/value pairs. It's not important which are amenity= and which are shop=.
> It is left to the editors and renderers to create whichever categories they think are useful - especially important, since there is not a One True Way to categorise the world!


while it is true that there are multiple ways to look at the world, I don't agree that keys are completely arbitrary (or should be completely arbitrary). Just because there are more ways to categorize the world does not mean there shouldn't be a concept/inherent logic behind the tags we use, or in other words, there should be a defined way we deliberately choose to make up our categories.

Of course this doesn't mean this map style or any other map style then does have to follow the same system - just to have a system (at all) will help contributors to find/choose the right tags and to introduce new tags for stuff that wasn't tagged so far.

matkoniecz:
> Under the same logic, hairdresser should be brown too. Not sure if I agree.

I would not oppose making hairdresser brown. It is a bit surprising but it makes sense (and I seem to remember my surprise that hairdresser icon has shop colour).

kocio-pl:
I still remember how I was surprised that hairdresser tag value is classified under the shop namespace!

I like the POI coloring system to be like this:
1. Violet for buying goods.
2. Brown for general services (paid or not).
3. Black for undecided/general things.

Of course there can be some more specific categories (I like pink for anything health-related and light blue for transportation) or exceptions (I rather see the fountain as a water-related blue [than amenity brown](https://github.com/gravitystorm/openstreetmap-carto/pull/1804#issuecomment-137721416)), but repairs (like car_repair and [electronics_repair](http://taginfo.openstreetmap.org/search?q=electronics_repair#values)) or services mainly (like hairdresser) should be brown IMO (and a [wig shop](http://taginfo.openstreetmap.org/tags/shop=wig) in violet, of course).

Issue #1529, opened 2015-05-05
Allow customisation of pattern files


nebulon42:
Currently patterns are PNG and are hard to customise/re-create. Since most symbols will now be SVG (see #1165) and easy customisation is a goal of this style this should also include pattern files.

As there might be problems with SVG versions of patterns (see https://github.com/imagico/jsdotpattern/issues/3) pattern files will probably stay in PNG format also after Mapnik 3 has been released. One solution could be to store the used symbols and the jsdotpattern configuration separately from the PNG file.

imagico:
See also imagico/jsdotpattern#4 and the specific remarks regarding the wetland patterns in https://github.com/gravitystorm/openstreetmap-carto/pull/1497#issuecomment-96428501

math1985:
I would suggest, for the time being, to at least include the source images and (a link to) the script with which they are generated.

I ran into this while trying to solve #1158. I found [this tool](http://www.imagico.de/map/jsdotpattern.php), but it took me too long to find out how to use the tool to generate the required image.

matkoniecz:
> jsdotpattern configuration

It may be a good idea to store also jsdotpattern itself ( https://github.com/imagico/jsdotpattern ). It will ensure that there are no problems with a disappearing project.

math1985:
Note that the licenses are currently not compatible.

matkoniecz:
Yes, it would require either complicating license description (CC0 except X folder that is licensed as...) or somehow saving it separately (what would be quite fragile).

imagico:
Come on guys - everyone is free to fork jsdotpattern or to host a copy for use, it does not make any difference if it is on gravitystorm's or my account.  It is just a tool like others you use during style development (text editor, tilemill, inkscape, gimp, whatever).

And of course (just in case there is any doubt) SVGs generated with jsdotpattern can have any license - this is the free decision of the user.  I did not specify anything about the symbols included but right now i would not consider most of them eligible for copyright anyway.

And yes, generating patterns with jsdotpattern is not really easy, especially with anisotropic patterns like the wetland dash pattern and often requires complicated sequences of relaxation with different settings.  Anyone is welcome to improve usability and algorithms.  But ultimately good design is hard work and only a certain amount of it can be hidden in clever software.

math1985:
> And yes, generating patterns with jsdotpattern is not really easy, especially with anisotropic patterns like the wetland dash pattern and often requires complicated sequences of relaxation with different settings. 

[...]

PR #1782, opened 2015-08-23
add documentation for generating forest pattern

matkoniecz:
current PNG pattern used in rendering is impossible to easily tweak and modify
this instructions were already present but buried as a commit message
fixes part of #1529


imagico:
You can meanwhile also set the pattern color without modifying the svg:

http://www.imagico.de/map/jsdotpattern.php#x,512,jdp6894;g,30,32,32;s,jdp33742;s,jdp81637;rx,250,2,32,32;s,jdp28824;s,jdp59702;s,jdp91550;s,jdp27774;rx,250,2,64,64;rd,1,0,0,tree%20pair,1,5,5,0,jdp52898,6b8d5e,add19e;

matkoniecz:
Thanks! I kept background transparent as it is easier to modify it in CartoCSS rather than regenerate pattern.

imagico:
Background color is only for display, the svg is always generated without.

matkoniecz:
Again - thanks! Now preview in jsdotpattern is with background colour.

PR #1789, opened 2015-08-26
sources and double resolution versions for bare ground and wetland patterns

imagico:
together with #1782 this addresses #1529.

jsdotpattern parameters are not included since i don't have them any more and the program changed a lot since then.

Includes SVG files and double resolution PNG versions of all patterns.  Also a description how the wetland patterns are assembled.

It also modifies the active files for rock and scree pattern since the current ones use a non-standard PNG gamma which can be irritating when editing.  Should be nearly identical in terms of rendering results.


matkoniecz:
Thanks! I wanted to play with tweaking colour for bare rock pattern and discovered that it is nearly impossible - now it should be doable.

Maybe add also rock.md file? Or rename wetland.md to wetland_and_bare_rock.md and mention that rock patterns were also produced this way? Now it is obvious but after year or decade...

matkoniecz:
Note that to fully address #1529 it is necessary to include also instruction for generating quarry pattern (it was produced by @kocio-pl - hopefully he will be able to prepare documentation).

imagico:
bare_ground.md added - but there is not really much two write - it is a plain, colored conversion of the original SVG.

kocio-pl:
I plan to do it eventually, but I'm just focused on other things currently.

nebulon42:
Great thing, thanks!

pnorman:
I'd suggest naming the 2x resolution ones `name@2x.png` or `name.png@2x`. A similar style of name is used for retina tiles by some providers.

imagico:
Changed file names.

dieterdreist:
sent from a phone

> Am 27.08.2015 um 10:17 schrieb Paul Norman <notifications@github.com>:
> 
> I'd suggest naming the 2x resolution ones name@2x.png or name.png@2x. A similar style of name is used for retina tiles by some providers.


this is apple's ios naming convention

almccon:
Yes, here's how we do it for double resolution tiles at Stamen: http://tile.stamen.com/toner/12/655/1583@2x.png

PR #1798, opened 2015-08-30
fix text-dy for nightclub

matkoniecz:
amenity nightclub name eeeeee eeeeee ref 1 ele 8000 operator ee ee ee operator brand ee ee ee brand node v2 30 0 - dy_again 200px


nebulon42:
Note that the nightclub icon seems to be the only one where a halo has sneaked in. Maybe this should be cleaned up too.

matkoniecz:
@nebulon42 Good idea. Changed and added to this PR (as after removing halo text-dy needed to be changed anyway).

nebulon42:
Great. Is this just something I'm seeing or why have I authored the second commit?

matkoniecz:
@nebulon42 I set you as author as second commit as it was your idea, your icon and I am nearly sure that nothing is broken by this change (and in case that something is wrong - I am still recorded as commiter) - so I decided that giving you credit is preferable.

In case that you prefer to change it I may set myself as the author.

math1985:
> In case that you prefer to change it I may set myself as the author.

Yes, I think it'd be less confusing to list yourself as auhor, and acknowledge @nebulon42 in the PR or commit description.

matkoniecz:
@math1985 Changed.

nebulon42:
I'm fine with either way or not mentioning me at all. :) I was just confused, didn't know that this is possible.

matkoniecz:
> I was just confused, didn't know that this is possible.

```git commit --author=<author>``` allows to set arbitrary person as author of the commit - see http://git-scm.com/docs/git-commit

pnorman:
It's best to only set an author if you're somehow applying patches from them or something involving picking their changes from other commits.

PR #1804, opened 2015-09-01
Adding amenity=fountain icon

kocio-pl:
Resolves https://github.com/gravitystorm/openstreetmap-carto/issues/705.

Warsaw (just a few fountains), Rome and Vatican (a lot of them) all look OK at z17:
fountain-saski-17
fountain-rome-parione-17
fountain-vatican-17

Only garden in Versaille has a problem with a row of fountains below z19 (other fountains here are not affected):
fountain-versailles-17
fountain-versailles-18
fountain-versailles-19

Because it's the most dense fountain place I could find, I guess it's not a show stopper for introducing this icon. Currently we're still not able to display fountain:nozzle=yes or any other arbitrary namespace, but this POI doesn't have to wait for those additional details. I also don't know if this row should be tagged as nozzles rather than fountains, but probably that's not the case.


imagico:
IMO the shown examples are great demonstrations how display of symbols can be counterproductive since it obscures and hides other things in the map.  Showing a 14px symbol (corresponding to >16m at z17) - in many cases for something like a little 30cm sprinkler in what is nearly always a dense urban setting seems one of the clearest cases of overuse of icons for showing point features.

Besides that introducing yet another color is not a good idea either.

kocio-pl:
1. The color is not new, it's reused - I have found marina text color good, because it has to be visible on the water (of course), yet it's less flashy than transportation blue (which would be bad here, because fountain has nothing to do with transportation). Do you think for example amenity-brown would be better?

2. I look for the right tagging for nozzles on Tagging right now to tag some of them if there's a need - I guess this is what you're talking about, not real fountains. Most fountains are much bigger than 30 cm anyway.

3. The size is universal for typical icons and has little to do with real size anyway. What do you think about telephone icon size on z17? Or monument icon on z19? Or border width on z1 (it probably doesn't have any. does it?)? Do you like to change the size depending on zoom level?

HolgerJeromin:
Perhaps a blue circle (like shops or even better the trunk of the tree rendering) and the icon with a larger zoom.

kocio-pl:
It might be good for z17 and I'd like to try it, but first I have to figure it out, because the code is more complex - looking at the trees, it may need some work with YAML also.

math1985:
What about a blue circle on mid-zoom and the icon on z19 (and maybe z18)?

kocio-pl:
I think just a point or just a circle are not enough to recognize the fountain:
fountain-point-only-17
fountain-basin-only-17

Luckily no YAML mangling is needed (it was just to pack tree and tree_row into one group), so I'm only playing with CartoCSS options.

matkoniecz:
Circle is enough to show that there is a water-related feature there. Maybe it is possible to make smaller fountain icon for earlier zoom levels?

Also, what about showing only named fountains since z17 and unnamed from z18/19?

Maybe using size of area as indicator of importance would work? Decent chunk of fountains (1/5) is tagged as areas.

math1985:
For me the dot works quite well in the examples you give.

matkoniecz:
@kocio-pl Can you push branches with dot and circle? I would like to test how it works.

imagico:
A simple small dot at z18 and a small symbol at z19 could work reasonably.  With the proposed symbols size we would still be talking about >4m nominal size at z19 which is quite large for most fountains.

Since most larger fountains are located within a water area the size of this water area would be a reasonable rating criterion - and using it would also encourage proper mapping.

matkoniecz:
It would be also a good idea to describe how fountains are supposed to be tagged. wiki.openstreetmap.org/wiki/Tag:amenity%3Dfountain is unclear at this moment and it merely describes two potential tagging schemes without recommending any ("Applies to Node nodes and Area areas although most fountains should be tagged as nodes representing the location of the fountain sitting within an area of natural=water").

After clarifying this it would be clear whatever size of fountain may be used as indicator of importance.

kocio-pl:
Reworked z17 with dot and circle in Warsaw (before/after):
fountain-saski-17
fountain-saski-dotcircle-17

I guess it's better, but I'm not sure about options taken from tree rendering and I need to test it also in other places.

@math1985 Circle suggests small water pool (which is a bit different water feature), while the dot I'd like to use for tagging just nozzles (once we have the tagging defined and the database reload is done) in big fountains like [this one](http://www.openstreetmap.org/way/79738429) (currently the basin has a name only, while individual nozzles are tagged as a fountain, which is not proper way of describing this object).

@matkoniecz I will think about your ideas later, fighting with the code was hard enough for today (brackets, brackets everywhere!...). Of course you can check it yourself now, since the code in this branch is updated.

matkoniecz:
@kocio-pl 

> while individual nozzles are tagged as a fountain, which is not proper way of describing this object

You may propose change on OSM wiki (at http://wiki.openstreetmap.org/wiki/Talk:Tag:amenity%3Dfountain ) to recommend this tagging scheme (this is small enough that full scale proposal is waste of time, probably it may changed by proposing it on talk page and assuming no objections change http://wiki.openstreetmap.org/wiki/Talk:Tag:amenity%3Dfountain ).

Currently it is not clear what is "the proper way".

kocio-pl:
Thanks, I will probably take the shortcut once I'm really sure how should it look like. [In my post on Tagging](https://lists.openstreetmap.org/pipermail/tagging/2015-September/026349.html) I have proposed this scheme for parts of the fountain:

`amenity=fountain + fountain:part=nozzle/light/artwork/*`

but plain `fountain:nozzle=yes` (I'm not sure if coupled with `amenity=fountain` or not?) could be good too.

Basically anything with `amenity=fountain` tag added will spoil the rendering until we have the access for the `fountain*=*` namespace and I have no idea when it could be.

Cactusbone:
you may want to try the gellagio fountains (http://www.openstreetmap.org/#map=19/36.11288/-115.17399)

kocio-pl:
OK, test pack for reworked z17 with tree-like design - Rome, Vatican and Versailles revisited look better:
fountain-rome-dotcircle-17
fountain-vatican-dotandcircle-17
fountain-versailles-dotcircle-17

And Gellagio (Las Vegas):
z17
fountains-vegas-dotcircle-17
z18
fountain-vegas-18

It works even now, but I think it should be tagged with one amenity=fountain (on the whole basin) and the nozzles should be tagged as - well - nozzles, so we would have one proper fountain icon and a lot of dots like on z17.

matkoniecz:
text-dy for dots probably may be smaller (on lower z-level)

kocio-pl:
It's the same as in bus stop now IIRC - 11 for every zoom level. What is this for (distance between the lines of text?) and what number do you feel is better? 9 maybe?

matkoniecz:
> What is this for (distance between the lines of text?) 

Distance between icon and text. Too small and tall names like "ÉÉÉÉÉÉ ÉÉÉÉÉÉ ÉÉÉÉÉÉ" will not fit - pointlessly large and it is uglier.

> and what number do you feel is better? 9 maybe?

No idea. I am always creating test file where relevant object is named "ÉÉÉÉÉÉ ÉÉÉÉÉÉ ÉÉÉÉÉÉ" (the worst reasonable case) and search for smallest value with label visible.

Dot is smaller so smaller values should be OK (what will make more obvious to what object label is applied).

kocio-pl:
Looks like z18 is perfect for showing regular icon, even in places crowded with fountains and other POIs:
fountain-rome-18
fountain-vatican-18

imagico:
In principle the 'dot + icon at the highest zoom levels' seems to work but note showing a small dot at one zoom level and a symbol >3 times the size at the next will inevitably lead to features frequently disappearing as you zoom in - something that is best avoided.  

Showing a circle in water blue will discourage detailed mapping of smaller basins around fountains which would be bad.

And i still think an additional color is a bad idea - for the icon either amenity-brown or landmark-gray seem fine to me - for the dot gray would be better.

HolgerJeromin:
> Showing a circle in water blue will... 

Yes, that's why I wrote about the trunk (I meant the size) and not the tree itself.

kocio-pl:
@imagico A lot of things to reply in your short note:

1. z18 is not the highest zoom level (yet it looks very good to me).

2. Small dot is not alone in my version - it has a bigger circle around. In some cases (where the water is already tagged) this won't be visible, but even a dot->icon (worse case) is more smooth than nothing->icon, as we do in most of the cases (only trees have the dot->another dot+circle progression), so I think this point is just wrong. You also seem to me to be confusing lines (where the width can be adjusted according to zoom level) with points (where the icon size is more or less universal). I have asked you the specific question about it ("Do you like to change the size depending on zoom level?"), because I don't get the idea. :disappointed: 

3. You're right, it may be discouraging, but still I think a dot alone has not enough of a context - plus it makes progression more smooth, as I wrote in 2. And on z18/19 lack of water pool will still be visible, I guess.

4. I want to try those other colors just to be sure, but I feel brown (better, because it's still amenity after all) or gray (which one? if "general", than worse, because we should left it for the general/undecided things) would work only for full icon. What brown dot (especially without the water around) could mean - a wooden pole? And gray dot - probably something similar to the bollard (steel pole for road signs maybe?). Sounds not convincing at all for me, but it's easy to check, so I will do it anyway, I just don't have much time, which means some delay may occur.

kocio-pl:
Brown is not working for me - more colors here is just better, because in Rome everything around is brown and it doesn't look pleasant on the water:

fountain-rome-brown-17
fountain-rome-brown-18

Vatican has smaller fountains and rather in gardens, so on z17 it's harder for me to recognize what this brown+blue objects could be (brown could be a trunk of the tree, but with a blue wrapping?) and on z18 it's also like in Rome - buildings and memorial are brown here, only a lot of green makes the problem smaller:
fountain-vatican-brown-17
fountain-vatican-brown-18

math1985:
I would also prefer blue.

What color do we use for waterfalls?

kocio-pl:
OK, text-dy distances are changed now to the lowest possible numbers. BTW: this is more or less how the Bellagio should look like (without nozzles visible):

fountain-dy-17
fountain-dy-18

kocio-pl:
Sorry, wrong button...

kocio-pl:
> What color do we use for waterfalls?

I guess we don't at the moment - first try is closed: https://github.com/gravitystorm/openstreetmap-carto/pull/1564.

matkoniecz:
Main problem at this moment is a poor rendering, worse than now of large fountains.

I have an idea how to change it, but it would require clear definition of amenity=fountain (I will post about it on wiki and probably also tagging mailing list).

example: https://cloud.githubusercontent.com/assets/899988/9870210/472d411e-5b8a-11e5-8333-7cec26e8662a.png

kocio-pl:
Great! I was just not able to enhance the definition, but I am still interested in fixing it to be more clear what exactly is a fountain and how to describe some large fountains to be more suitable for 3D rendering too.

kocio-pl:
@matkoniecz I'm still curious what is this idea?

Looking at your example it doesn't seem to me that there's any problem with current proposition up to z20 and for higher zoom levels any area with the icon would look strange, because in this scale we see they are not the points (on z22 it's probably just a rendering artifact).

math1985:
Accepting this as I think it's an improvement to the current status. If anyone has improvements to particular aspects, feel free to create a PR.

imagico:
I find it regrettable that the points mentioned in https://github.com/gravitystorm/openstreetmap-carto/pull/1804#issuecomment-137565841 have not been addressed.  Correspondingly i disagree that this is an improvement of the current status.

Beyond issues of the rendering of this particular feature this change adds two major novelties to this style: introduction of a new icon color and use of water color to display abstract shapes rather than mapped geometries.  I think neither of them has been sufficiently considered in its broader implications.

math1985:
@imagico Sorry for not addressing your points in detail. I have limited available time, and in an ideal world I'd respond to all comments, but if I have to choose between reviewing PR's or commenting, the former tend to be more useful.

> In principle the 'dot + icon at the highest zoom levels' seems to work but note showing a small dot at one zoom level and a symbol >3 times the size at the next will inevitably lead to features frequently disappearing as you zoom in - something that is best avoided.

Wouldn't that be the same when an icon shows up on a zoomlevel without a dot on the previous level? I think the important thing to prevent features disappearing is to make sure that rendering order correspond with the order of zoomlevels on which icons show up - I'm not sure if that's the case here, though.

> Showing a circle in water blue will discourage detailed mapping of smaller basins around fountains which would be bad.

I'm not sure if that's really an issue - adding a fountain outline is typically something mappers do when they are mapping other features in the same area, so they are already in the editor, where they don't see the default rendering.

> And i still think an additional color is a bad idea - for the icon either amenity-brown or landmark-gray seem fine to me - for the dot gray would be better.

To me the blue icon looks really natural - but that's of course a matter of taste.

imagico:
> Wouldn't that be the same when an icon shows up on a zoomlevel without a dot on the previous level?

No, there are two separate problems with relation to icons vanishing.  One is new icons vanishing because *other* icons appear at a higher zoom levels with priority over the icons that already appear at lower zooms.  This can be solved by ordering priority correctly.

Here however you have the problem that the size of an icon raises significantly relative to ground level units from z17 to z18 and this will make icons disappear when zooming in even if there is no change in any other icon type around.  This is already well visible in the samples shown:

z17

z18

> I'm not sure if that's really an issue - adding a fountain outline is typically something mappers do when they are mapping other features in the same area, so they are already in the editor, where they don't see the default rendering.

I am pretty sure that there are people who disagree that drawing a 10m circle around every fountain above almost everything else in the map (roads, paths, landuses) is not a problem.

> To me the blue icon looks really natural - but that's of course a matter of taste.

I think it is probably not a good idea to choose a color individually for every icon type with regards to what appears natural for this feature in isolation.  It is fairly unclear to me what broader class of object this color is meant to represent.  If it is supposed to be water related features it should be clear that the current transport icon color is much closer to water color and this color is fairly close to air transport color which would make color choice fairly unintuitive overall.

I am sorry if this appears like an unfair rant towards the work of @kocio-pl who is just trying to improve the map but there is no point in sugar coating this i think.  I just see serious issues here, i hope i am wrong about this but so far i see no indication i am.

kocio-pl:
@imagico 
> I find it regrettable that the points mentioned in #1804 (comment) have not been addressed. 

Well, you're wrong here - they've been addressed, in as much details as I was able, a month ago (https://github.com/gravitystorm/openstreetmap-carto/pull/1804#issuecomment-137584460, https://github.com/gravitystorm/openstreetmap-carto/pull/1804#issuecomment-137721416). I was very surprised you didn't answer it, but nobody else was against, so I just moved along. I was rather waiting for @matkoniecz to at least reveal his idea, but it also didn't happen in 2 weeks. 

As much as I think @math1985 was right that waiting forever is not gonna work and I'm happy he took this action, I'm still interested in discussing problems.

> introduction of a new icon color and use of water color to display abstract shapes rather than mapped geometries.

How would you like to show the geometry? You seem to assume that every fountain is a typical nozzle with water basin around, which is not always the case - that's why [we have designed it to allow showing other types too](https://github.com/gravitystorm/openstreetmap-carto/issues/705#issuecomment-106769767). There can be a fountain (containing one or more nozzles):
1) [in the bigger water area](https://github.com/gravitystorm/openstreetmap-carto/issues/705#issuecomment-106783412) 
2) [with no water pond around and deliberately a part of a pedestrian erea](http://galerki.pl/zdjecie/160)

How would you like to indicate the shape?

imagico:
> Well, you're wrong here - they've been addressed, in as much details as I was able, ...

I am sorry if i was not clear here - with 'addressed' i did not mean 'got replied to' but 'got solved by making changes'.

You are always asking for specific suggestions for changes but just because none are given does not mean there is no problem (although i have given several specific suggestions in this thread).  To me this seems a bit of a conflict between two approaches to style improvements:

* the primacy of adding new features, meaning that the improvement of adding support for some new feature trumps any issues with the way this is implemented.
* the primacy of overall style consistency and readability, meaning that a change has to improve the overall quality of the map and the addition of an individual new feature only plays a minor role here.

In general I think everyone is responsible for the changes he/she makes and others (like me here) can only help with comments and remarks.  So it is fine to disregard my concerns if you find them unfounded (do you?).  I re-emphasized them here because i strongly believe they will turn out to be issues (in other words: this change will create more problems than it solves) and i expect someone will have to address them sooner or later.

kocio-pl:
> I am sorry if i was not clear here - with 'addressed' i did not mean 'got replied to' but 'got solved by making changes'.

OK, now I understand. :smile: 

However I expected that you answered me showing where you agree and where are still the flaws in my reasoning. The lack of the response got me surprised: I don't like to make the changes when I still don't know the conclusions from a discussion (like "here we mostly agree, this is where I have a different point of view, but I think we can reach a compromise and here we just disagree").

> You are always asking for specific suggestions for changes but just because none are given does not mean there is no problem (although i have given several specific suggestions in this thread).

Yes, because suggestions (or even a working code) are always better than just pointing the problem and sometimes they can end useless discussion and help find the solution (faster or at all...). But of course this is not always the case and I like to discuss - and sum it up to not let the problems got lost in words, especially when there are many details involved. 

> I re-emphasized them here because i strongly believe they will turn out to be issues (in other words: this change will create more problems than it solves) and i expect someone will have to address them sooner or later.

Thanks for reminding them in a compact form! I will try to reply and hope we can at least reach conclusions and test some solutions maybe.

daganzdaanda:
@kocio-pl asked "Do you like to change the size depending on zoom level?" 
-- maybe that could work? Just scaling a water-colored dot to a circle? 

The circle should always be pretty small, so that micro-mapping the real basin will still be a visible improvement. I don't think we even need a dot or icon in another colour in the middle. 
I like the icon, but I can see that with current tagging, it may get repeated annoyingly often in some places. So if you want to use something more than a water-circle, I'd vote for a small dot.

I would slightly prefer brown for label colour. When I first saw the examples, I liked the blue more, but now I find it too aggressive and the brown actually very elegant. Funny how taste may differ, @kocio-pl feels  "brown (...) doesn't look pleasant on the water" :-)

jojo4u:
I changed the wiki according to the discussion here and on tagging. But is this area rendering already in place? Or is this only about the icons?
http://wiki.openstreetmap.org/w/index.php?title=Tag%3Aamenity%3Dfountain&diff=1222631&oldid=1212273

jojo4u:
Looks like a closed way with amenity=fountain is rendered like a node: http://www.openstreetmap.org/#map=19/48.77501/9.17879&layers=N / http://www.openstreetmap.org/way/100055576

Klumbumbus:
What is the source and license of the fountain icon? (I would like to use it in JOSM)

kocio-pl:
This is my own design and I'm an avid JOSM user, so it's nice to hear that! =}

License is CC0, [like all the files in this repo](https://github.com/gravitystorm/openstreetmap-carto/blob/master/LICENSE.txt), so there are no limits, you can do what you like.

Klumbumbus:
Thanks, sounds good :)

PR #1806, opened 2015-09-02
Changing aeroway=helipad icon

kocio-pl:
Symbol letters (like "H") are heavily language dependent, so the real icons are preferred for truly universal map style like this one. This icon is taken from Osmic repository:
halipad-banacha-17


HolgerJeromin:
As the H is painted on the ground (world wide!?) the language dependency does not count so much in this case.

kocio-pl:
I'm not sure how recognizable is it worldwide, especially outside the Western world. I've just found some examples of different markings:

HolgerJeromin:
i am nitpicking, but in the white cross on the second image is probably a red H.

kocio-pl:
Oh, you're right! :smile: 

I'm still interested how could we check if H is enough all over the world.

HolgerJeromin:
BTW is this a char H or does it symbolize the two skids of a typical helicopter? :-)
But there are some other pads:
http://www.traxnyc.com/blog/audemars-piguets-unique-heliport-advertisement
http://fecheliports.com/sites/fecheliports.com/files/imagecache/product_full/sites/new.fecheliports.com/files/sidebyside/WESH1lg.jpg

kocio-pl:
Nice! :smile:

imagico:
The 'H' symbol is an international standard (both de facto and formally - see for example [here](http://www.bazl.admin.ch/experten/regulation/03080/03081/03082/index.html?lang=it&download=NHzLpZeg7t,lnp6I0NTU042l2Z6ln1ah2oZn4Z2qZpnO2Yuq2Z6gpJCDe319gWym162epYbg2c_JjKbNoKSn6A--))

Needless to say that the helicopter icon is both ambiguous (landing site vs. heliport where helicopters are stationed) and more difficult to read than the 'H'.

The symbol proportions could be adjusted to comply with the ICAO standard of course.

mboeringa:
> As the H is painted on the ground (world wide!?) the language dependency does not count so much in this case.

\+1, I don't think it needs replacement.

nebulon42:
@imagico It might be a standard, but since this is a general purpose map and not one for aviation professionals it still might be harder to recognize for some people. Talking about ambiguity the H could also being confused with hospital or with bus stops or the like. So it is not less ambiguous than the other one.

Just wanted to point out that there are pros and cons for either version.

kocio-pl:
@nebulon42 +1 

I am not sure that we have to change it, but I want to take a look at it, especially given how easy was to make this PR. "Q" in quarry was really bad, "S" in spring is also bad (working on it), "H" here and "P" in parking(s) may be less of a problem, but still if we have a replacement, it's at least worth checking.

imagico:
That's right - I just wanted to make clear that there is no other widespread sign for marking a helipad on the ground other than the 'H' (in combination with the cross for hospital pads).

And ultimately having symbols that are intuitively understandable by the average map user without a map key is IMO secondary to the aim of having a well readable map.

If you look in other maps the 'H' is very widespread and the helicopter symbol is quite rare by the way.

matkoniecz:
H icon is also simpler - it works IMHO better for situation in provided example ( http://www.openstreetmap.org/?mlat=52.20870&mlon=20.98225#map=17/52.20870/20.98225 ).

kocio-pl:
This icon could be used for some heliports maybe (if this is something bigger than helipad - I don't know), but we have no single use of aeroway=heliport in the database according to TagInfo.

BalooUriza:
On Wed, Sep 2, 2015 at 12:27 PM, Christoph Hormann <notifications@github.com
> wrote:

> That's right - I just wanted to make clear that there is no other
> widespread sign for marking a helipad on the ground other than the 'H' (in
> combination with the cross for hospital pads).
>
 Acceptable variants are an empty circle on the outside edge of a helipad
(where orientation of the landed craft is unimportant), or a triangle with
a filled in corner indicating the direction the landed aircraft should face
when over it.  Both typically have an H in the middle, but not guaranteed.
The variation without the H is widespread, but uncommon.

mboeringa:
> This icon could be used for some heliports maybe (if this is something bigger than helipad - I don't know), but we have no single use of aeroway=heliport in the database according to TagInfo.

That's true because they are tagged as **aeroway=aerodrome** instead of aeroway=heliport, which is a non-documented tag. So they are tagged just like normal airports. The only way to distinguish them is through their name, which usually contains a "heliport" phrase (but that is language depended), see this example in Paris:

http://www.openstreetmap.org/way/14036409

kocio-pl:
@mboeringa Thanks for the explanaition! OK, so we can leave helipads rendering unchanged and use this icon for heliports (as intended by @nebulon42). 

The first question is how they should be tagged, so I will ask about it at Tagging. Having it will be beneficial for this style, because it would [ease airports overload pains in some places](https://github.com/gravitystorm/openstreetmap-carto/issues/1143), because I guess heliports are generally smaller than airports (no runaways needed), so they should be rendered later. It would be also better for users to instantly recognize that it's not for planes.

The second question is more about this style: once we have the tagging scheme defined, when we start really rendering them? Straight from the beginning (as it's clear what this object is and we want to promote using it) or we wait for some conditions to be met? Hint: we have currently [about 6k uses with value "heliport"](http://taginfo.openstreetmap.org/search?q=heliport#values), so given that some of them are just in different languages, it might be at most - let's say- 5k. I would go for the first, as this is rather clear case. What do you think?

matkoniecz:
> I would go for the first, as this is rather clear case. What do you think?

I would find and check whatever tagging scheme will be adopted. Maybe after threshold of 1000 usages would be crossed (assuming that proposal will be accepted)?
 
Anyway, this pull request will not be merged so I am closing it.

kocio-pl:
[The discussion on Tagging is started](https://lists.openstreetmap.org/pipermail/tagging/2015-September/026376.html). 

Why should we always use "magical" 1k limit? I think this case is so clear (of course once the proper scheme is chosen), that we could act more like a feedback loop than like the consumer of the db data at least this time.

kocio-pl:
[Voting period for aeroway=heliport](http://wiki.openstreetmap.org/wiki/Proposed_features/aeroway%3Dheliport#Voting) has started.

kocio-pl:
Voting is closed now and the scheme is accepted. Now is a time to rethink what measures of "adopted" we plan to use? 

1000 uses can be reached quite fast given that some mass edits or a few dedicated mappers are involved, but I want to avoid any unnecessary controversy (we have enough of it) and know what could we really take as a honest sign that we can start rendering it?

Issue #1774, opened 2015-08-21
render landuse=aquaculture


Penegal:
Hello, dear CartoCSSers.

I just noticed `landuse=aquaculture` isn't rendered. I don't know for other countries, but, here in France, this is a common land use along the coasts, usually tens of hectares lots; mainly oyster beds, but we are not the only ones who have this kind of farming, for oysters or other species, and it can be a useful feature of the seascape to guide oneself on the tidal flats, or simply to avoid damaging them by piloting one's boat on them. Should be good to render it. A suggestion for the rendering: using a striped zone with a deepest blue than the one used for the sea?

Hoping this is relevant,

Regards.

polarbearing:
[6.8K uses](https://taginfo.openstreetmap.org/tags/landuse=aquaculture#overview), and [wiki](https://wiki.openstreetmap.org/wiki/Tag:landuse%3Daquaculture) started 2012.

Can you point to good examples? Can we assume that there is some water feature mapped as well?

Penegal:
Well, there's [this one](https://www.openstreetmap.org/way/67303378) or [that one](https://www.openstreetmap.org/way/67303372), and you can find [here](https://www.openstreetmap.org/edit#map=16/45.8478/-1.1853) an untagged example, to see the space such landuse can... well, use.

matkoniecz:
Probably rendering should happen in layer used by forest symbols to keep it visible in lakes.

kocio-pl:
Maybe some simple shells pattern should be used here?

polarbearing:
Pattern yes, the question is which. Aquaculture could also be salmon in the sea and freshwater fish in ponds. Maybe a fish icon is more generic. The 'fishmonger' would also sell shells.

kocio-pl:
We already use fish icon for fishmonger shops, but it would probably be just the outline and no in violet, so it's also the shape available to use here.

kocio-pl:
Set of propositions on z14 using current seafood icon with black, 10% gray and marina-text with or without background (quarry color):
![aquaculture-black-14](https://cloud.githubusercontent.com/assets/5439713/9706310/a2fc237a-54e1-11e5-863a-55a116be6003.png)
![aquaculture-black-nofill-14](https://cloud.githubusercontent.com/assets/5439713/9706311/a4d7bc86-54e1-11e5-8182-2c7c4cd1e69d.png)
[...]

Issue #1285, opened 2015-02-02
Inconsistent hatching rules


kocio-pl:
We have inconsistent hatching rules for amenity=prison and landuse=military (I don't know if there are some other hatched objects right now): one is rendered as the lower layer than buildings, grass and parking, while the other is rendered above all these elements, but - as far as I can see - below the highways.

See the example here:
https://www.openstreetmap.org/#map=19/52.20806/21.01130

Do we want to have general rules for layering hatchings (I guess we do) and how should they be defined - or maybe we want to express something by using different patterns?

math1985:
> Do we want to have general rules for layering hatchings (I guess we do) and how should they be defined - or maybe we want to express something by using different patterns?

We don't, and we should :)

kocio-pl:
I meant "we do _want to_". =} 

One more, but so basic thing, that I have even not checked - the geometry (hatching density and angle) is different...

polarbearing:
The angle is intentional, it is vertical on the prison (symbolising the bars), and 45 deg on the military. Please consider the optical illusion from the oblique buildings in your example if you don't see it vertical. 

The density is 3px line in a 16x16 png, both in the vertical and the 45 deg version, the perceived difference comes from the different angle.

kocio-pl:
You are probably right about all those optical effects, however I'm skeptic if we want such distinction at all - I think the color+icon tandem is enough. In my view "cool" symbolism was taken not considering consistency of hatched areas.

Nevertheless there's still the problem of layering - key options are:
1. Hatching above all the objects
2. Hatching beneath all the objects

On military area we have probably 1a - hatching above most objects except highways. Now we need to decide (and justify it somehow). I would say 1. is preferable to me.

pnorman:
I need to add another landcover overlay for #1242 that lies above landcover but below buildings

matkoniecz:
see also http://www.openstreetmap.org/?mlat=50.08134&mlon=19.89985#map=17/50.08134/19.89985

![selection_002](https://cloud.githubusercontent.com/assets/899988/9820977/a5ad1370-58b7-11e5-9f7e-292ea728200c.png)

kocio-pl:
@pnorman 
> I need to add another landcover overlay for #1242 that lies above landcover but below buildings

This issue is closed now - do you still plan to do it?

Tomasz-W:
[...]

PR #1817, opened 2015-09-07
Adding landuse=aquaculture pattern

kocio-pl:
Resolves https://github.com/gravitystorm/openstreetmap-carto/issues/1774.

The idea is to have kind of "underwater orchard" with regular blue dots pattern with a boundary line in the same color (water-text):

aquaculture-dot-6699cc-14


pnorman:
When I look at it, it's not obvious to me what it represents.

kocio-pl:
Can you think about anything that could be clear aquaculture visual representation for you? I'm ready to test it.

pnorman:
> Can you think about anything that could be clear aquaculture visual representation for you?

No.

polarbearing:
> When I look at it, it's not obvious to me what it represents.

This was discussed in #1774 and preferred over more picturesque versions. You cannot expect that every feature is obvious from the first glance, and at the same time not too obtrusive. 

We have the query tool on the website so it is easy to check what it is in detail.

matkoniecz:
> Can you think about anything that could be clear aquaculture visual representation for you?

Unfortunately no. And current rendering IMHO rather fits more generic restricted area (fish pattern also was problematic, though for different reasons).

kocio-pl:
I'm not aware of dots being used for restricted area - we have hatching exactly for this and we use it very conservatively (only 2 object IIRC - and still we render them [a bit different](https://github.com/gravitystorm/openstreetmap-carto/issues/1285)), so I think allotments could be more suspected about being restricted. And if you still feel dots look like restricting something, the orchard (or maybe vineyard too) is - by design - very common in that regard to the current aquaculture draft, so they maybe should be removed to make this style consistent about using patterns?

What I'm trying to say is we may have different feelings, but we should not keep forgetting about already existing features and treat them as given and obvious. Archaeological sites and quarries were good examples of problematic visual representation.

imagico:
The pattern seems to work fine for me but the outline does not - outlines without an area fill usually do not work well, especially when shown at strongly varying scales.

Apart from that z10 is too early for such a pattern, you will usually only have individual dots visible at this scale which is not helpful.

kocio-pl:
@imagico 
> outlines without an area fill usually do not work well, especially when shown at strongly varying scales.

Could you give some examples of such problems? I don't feel it might be a problem at all, so kind of proof (or even just a hint) would be helpful as a reality check.

> Apart from that z10 is too early for such a pattern, you will usually only have individual dots visible at this scale which is not helpful.

It depends on the size probably and I guess sea aquacultures can be quite big (according to FAO [global fish production in such farms almost doubles wild captures](https://en.wikipedia.org/wiki/File:Global_total_fish_harvest.svg) and it's a hint how big areas may be involved). Smaller areas can be dynamically hidden as in case of military areas (way_area comes to the rescue).

What zoom level do you think would be better to start with?

imagico:
> Could you give some examples of such problems?

Not off hand - this is simply very rare in maps because it does not work well.  If you disagree i would suggest you look for examples where it works.

kocio-pl:
Sure, I will check them later. Also if you will find some counterexamples or some documents regarding outlines without the area fill, drop a line here.

pnorman:
It needs to be possible to gain an understanding of the map without a legend.

How are other maps representing aquaculture and are there any standard ways to do so?

math1985:
I agree the proposed rendering isn't really clear, and map users also won't be able to learn easily what it means.

The general tendency seems to be that this PR is not good (enough) for merging, and it's not really clear how it can be improved, so I guess closing the PR is the only thing I can do now.

I'd be happy to look at a new effort if new ideas come up, of course.

Issue #1800, opened 2015-08-31
names of amenity=kindergarten/school/college/university/grave_yard are not rendered


kocio-pl:
Example of the problem: the name of [this object](http://www.openstreetmap.org/way/339573260) was visible, but now it's not even on z19.

matkoniecz:
It seems that something went wrong in https://github.com/gravitystorm/openstreetmap-carto/commit/4745130e3d4a67472e046c9a7c0c25eed6c132c6

(college also seems to be affected)

Issue #1802, opened 2015-09-01
Some yaml lines are extremely long


matkoniecz:
For example:

 ```                  ('railway_' || (CASE WHEN railway = 'preserved' AND service IN ('spur', 'siding', 'yard') THEN 'INT-preserved-ssy'::text WHEN (railway = 'rail' AND service IN ('spur', 'siding', 'yard')) THEN 'INT-spur-siding-yard'  WHEN (railway = 'tram' AND service IN ('spur', 'siding', 'yard')) THEN 'tram-service' ELSE railway END)),```

is one line. In my typical editing environment ```'siding', 'yard')) THEN 'INT-spur-siding-yard' ``` and later part requires scrolling to be visible - what is not making easier to read and edit this SQL.

I think that readability and editability would be improved with linebreaks. I already did it with surface SQL - see https://github.com/gravitystorm/openstreetmap-carto/commit/bf562c637266abb123fc5dd22cd5e2434adae52b?diff=unified

The question is - how long line should be? Or should it stay unlimited as it is now?

HolgerJeromin:
perhaps the correct answer is not line length, but count of strings. Perhaps something like:

* three strings are ok. 
* four till ten should get an own line. 
* more than ten should...
 * every entry gets an own line (not really good)?
 * group of three to five gets an own line (if could be groups logically)?
 * another suggestion?

kocio-pl:
> every entry gets an own line (not really good)?

I would like it for shops and amenities for example, because every time something is changing here, I (or @math1985, who is usually pulling my code) have to make the changes in YAML once again. It's error prone and tedious, because diff is not able to find that new `'value',` string is inserted or moved. I guess this could be the source of [this regression](https://github.com/gravitystorm/openstreetmap-carto/issues/1800).

kocio-pl:
> group of three to five gets an own line (if could be groups logically)?

This would work only if we're adding new items at the end, but I like to have more or less alphabetical order and when I add special icon for already existing item, it means also removing the same string from another place in the same line.

kocio-pl:
We could maybe use some group variables? This would make this code:

```YAML
CASE WHEN surface IN ('unpaved', 'compacted', 'dirt', 'earth', 'fine_gravel', 'grass', 
                  'grass_paver', 'gravel', 'ground', 'mud', 'pebblestone', 'salt', 'sand', 'woodchips', 'clay') THEN 'unpaved' ELSE 
[...]

PR #991, opened 2014-09-27
Refactor water-features

math1985:
Dam, weir, lock:
* Add rendering for dam areas (supercedes and closes #227, based on code by
  @chrisfleming), dam and weir nodes, and lock_gate line
* Render weir in right layer (resolves #403)
* Change rendering of weir so it renders with small dashes
* Use marker instead of image for lock_gate node
* Add labels for dam nodes, dam areas, weir, and lock_gate
* Improve labels for dam lines and lock lines
* Remove labels for undocumented and less common waterway=lock

Groyne, breakwater:
* Render groyne/breakwater area the same color as groyne/breakwater way
  (resolves #893)
* Add label for groyne, breakwater, and pier (resolves #474)

Other:
* In general rewrite and clean up code

PR #1825, opened 2015-09-10
Fixing missing amenity name labels regression

kocio-pl:
Resolves https://github.com/gravitystorm/openstreetmap-carto/issues/1800.

Fixing YAML/MML broken in commit 4745130e3d4a67472e046c9a7c0c25eed6c132c6. It is still spaghetti code, which is hard to compare (long lines and all the layers together), so please check if I didn't make a mistake!

I have checked it works with:
- school
- kindergarten
- university
- college
- grave_yard

so all the missing labels are back.


matkoniecz:
Why in some cases order of listed amenities was changed from ```'pub', 'restaurant'``` to ```'shelter', 'atm'``` (coalesce in amenity-points-poly) and in some reverse happened (where clause in text-poly).

I am not sure whatever one order is preferable over another but it would be better to make it consistent.

HolgerJeromin:
I propose a PR with a commit fixing the problem and a second commit adding new lines and sorting stuff.
With this strategy review if easier and we are faster on at the desired yaml style.

kocio-pl:
I was trying to make as few changes as possible and just reverted the broken lines (adding only community_centre when needed). This is meant as a quick fix, I don't think more changes are sane before we will introduce the solution [@gravitystorm have proposed lately](https://github.com/gravitystorm/openstreetmap-carto/issues/1802#issuecomment-138809013).

kocio-pl:
Does anybody from the team plan to merge this code or is there any problem with it that needs to be solved? 

I would really like have it done ASAP and then go further with code cleaning to avoid such problems in the future.

HolgerJeromin:
You reverted the commit and readded the needed value. 
That is probably a good idea, but to review we have to compare to the state before the broken commit.
Another approach would be fixing the broken things from the current state, so the github diff is showing exactly what you have fixed.

But probably the maintainer are skilled enough to be able to review both approaches. :-)

math1985:
As far as I'm concerned, I haven't had time to review any pull requests lately. I will look at it as soon as possible!

In the future, I'll try to be more explicit when I'm not merging something because I'm not convinced enough about the merits of a PR (but that was not the case here).

@HolgerJeromin Either way is fine with me, that's not the reason nothing has happened so far.

kocio-pl:
> As far as I'm concerned, I haven't had time to review any pull requests lately. I will look at it as soon as possible!

Thanks!

But I'm confused: are you assigned for this task and the rest of the team in general does not pull any code in - or it works in some other way?

> In the future, I'll try to be more explicit when I'm not merging something because I'm not convinced enough about the merits of a PR (but that was not the case here).

It's always good to know what is the status of a PR not being actively discussed/reworked. Few days with no additional input should not be a problem (if it's not something serious like this regression), but a week or two would be a good mark of PR being forgotten or having some general problems that should be resolved (or just to be closed for some reason).

@matkoniecz [wrote elsewhere](https://github.com/openstreetmap/openstreetmap-website/issues/1038#issue-102617686) about the experience of being uncertain with creating PRs and this is also the problem I share sometimes in this repo.

math1985:
> But I'm confused: are you assigned for this task and the rest of the team in general does not pull any code in - or it works in some other way?

No, any of the maintainers can accept pull request, there is no agreed distribution of tasks within the team.

> but a week or two would be a good mark of PR being forgotten or having some general problems that should be resolved (or just to be closed for some reason).

It happens regularly that there are two weeks in a row in which I have no time to spend on this project. Just a natural thing for a project run by volunteers, and no indication of the quality of proposed PR's!

matkoniecz:
> But I'm confused: are you assigned for this task and the rest of the team in general does not pull any code in - or it works in some other way?

At least for me it is 100% hobby so whenever I am in the mood to review PRs and I have spare time I am doing this. Otherwise I am doing other things.

> It's always good to know what is the status of a PR not being actively discussed/reworked.

Whenever I see some kind of problem I comment immediately (here it was https://github.com/gravitystorm/openstreetmap-carto/pull/1825#issuecomment-139358836), in case of starting review that will last for a long time I assign PR to myself (like at https://github.com/gravitystorm/openstreetmap-carto/pull/991#event-254340763 ). Anything else means that I did nothing with given PR.

> Few days with no additional input should not be a problem (if it's not something serious like this regression), but a week or two would be a good mark of PR being forgotten or having some general problems that should be resolved (or just to be closed for some reason).

It means that PR waits for somebody who will review it. It is not possible that PR will be lost - we have no huge backlog of proposed patches and github interface is OK.

And on finding "general problems that should be resolved" people will give comment.

kocio-pl:
Thank you both for the comments! 

Sorry for sounding too alarming probably - I was just not sure how is it supposed to be working (especially in the context of this fix). Lack of time for OSM is something I can perfectly understand because I'm doing it also as a pure hobbyist, it was just not clear to me if it may be also some other problem with decision making or workflow within a project.

kocio-pl:
@gravitystorm Any chance to release v2.35 containing this fix anytime soon (there were more changes lately, so naming it v2.34.1 wouldn't make sense probably)?

matkoniecz:
@kocio-pl You may want to subscribe to updates from https://github.com/openstreetmap/chef/ (for example https://github.com/openstreetmap/chef/issues/36 is relevant in that case).

kocio-pl:
Thanks!

PR #1830, opened 2015-09-13
stop rendering landuse=conservation

### LINKED ISSUE #453: Relation (foreshore) is obscuring a underwater landuse=conservation tag
This property (http://www.openstreetmap.org/way/271045662) is in shallow tidal marine water that is mostly mudflat at lower tides. This parcel is managed by a state Fish and Wildlife agency. Seems like the landuse=conservation polygon would paint if the foreshore natural=water layer was not on top or could be translucent.  The foreshore polygon http://www.openstreetmap.org/relation/1341207 needs to be shown regardless. Like it is on most topo maps and marine navigational charts. You can see what I mean here: 

<iframe width="425" height="350" frameborder="0" scrolling="no" marginheight="0" marginwidth="0" src="http://www.openstreetmap.org/export/embed.html?bbox=-124.11718368530273%2C40.81874599835864%2C-124.07298088073729%2C40.83891147642241&amp;layer=mapnik&amp;marker=40.82881326765344%2C-124.09508228302002" style="border: 1px solid black"></iframe><br/><small><a href="http://www.openstreetmap.org/?mlat=40.8288&amp;mlon=-124.0951#map=15/40.8288/-124.0951">View Larger Map</a></small>

### LINKED ISSUE #1826: If and how to render landuse=conservation
landuse=conservation is currently rendered in grass color which is obviously not a good idea since this rendering implies a certain vegetation that tag does not.

In #1655 i originally intended to render it like nature reserves but i came to realize this might not be the best approach.  The tag description on

https://wiki.openstreetmap.org/wiki/Proposed_features/conservation

is very vague, full of words like 'mostly', 'often' and 'typically' - there is hardly any hard criterion listed there to identify what constitutes landuse=conservation.  I know this tag is used in the UK for a specific purpose but elsewhere its use is fairly arbitrarily.  What characterizes landuse=conservation in the UK seems to be that it is land protected from development.  But this is a restriction that is mostly relevant to the land owner and of little importance to the visitor - in contrast to nature reserves which usually come with restrictions like not to leave paths, no camping, no picking of plants etc.

Overall i think landuse=conservation should probably not be rendered at all.  If there are legal constraints beyond protection from development it should be tagged leisure=nature_reserve or boundary=protected_area.  If anyone can show me an area of landuse=conservation where these tags would be wrong and that should be rendered on the map none the less that would be an argument of course.

matkoniecz:
fixes #1826, fixes #453


matkoniecz:
And in some cases it fixes unreported bug with missing label:

landuse conservation node 53 2 -1 8 -_ 48 7736995 3 3572663 master-_conservation 4 19 master - conservation 400px

Issue #887, opened 2014-08-17
Add rendering for names man_made=works


daganzdaanda:
Do you think it would be sensible to display the name of factories tagged with man_made=works? 
We already render landuse=industrial, so there may be some duplication of labels afterwards, but surely there are factories that don't sit on a named landuse, and then you just see a huge building with no indication as to what it is.

HolgerJeromin:
Often we have a big landuse=industrial area with multiple man_made=works nodes inside because the mapper had no idea where the different companies are located inside. From the outside it is not easy to decide where one company ends and the next starts inside the landuse. 
IMO this is a valid transition tagging. 
But right now the works nodes are not rendered with a name, so you have no idea which company is there. (Only if there is a named parking :-)

abdeldjalil09:
Please render this tag, in many industrial areas , we see his housenumbers instead of his names, we have ~90000 uses.

kocio-pl:
Should be rather easy to do and it sounds to me like we have agreement here.

nbcaldon:
This is similar to an issue we are having at the moment with industrial estates:
With multiple shops and amenities in a building, it is possible to show the names of the shops on the standard map by adding each shop as a node within the building. With industrial buildings, the nodes can be added in the same way, but the names of the nodes are not rendered. I propose that where a node is tagged with "industrial=" and "name=" the name should be rendered on the standard map. At present it does not seem to be possible to make the names of the companies visible if there are more than one in a building, unlike with shops where they can all be shown.
I am not sure whether this is a new issue or covered by the present one?

HolgerJeromin:
@nbcaldon 
Only nodes with industrial= and name= is IMO not sufficient. man_made=works should be added and therefore this issue should cover your situation.

nbcaldon:
@HolgerJeromin 
Thanks for your comment, but the name still does not render if I add "man-made=works".

HolgerJeromin:
Yes. And that is why this github issue is still open.

PR #1836, opened 2015-09-15
Add rendering for power=plant

### LINKED ISSUE #1835: Recognize and render power=plant
osm-carto currently recognizes and draws power=station, but this seems deprecated and is not listed in the OSM wiki (if it ever was). Instead, the wiki states to use power=plant. Please consider making osm-carto draw it.

Example objects: http://osm.org/way/237100663

kreed:
This will render names for power plants.

Closes #1835, supersedes #1771


matkoniecz:
I would expect that power=plant would be anyway tagged as landuse=industrial. Are there cases of correctly tagged places where rendering power=plant like landuse=industrial would improve situation?

Also, #1835 mentions that power=station should not be rendered, maybe it should be removed from rendering?

HolgerJeromin:
right now we have about four times more stations that plants. So perhaps it is too early to drop station rendering. But as this tagging has no documentation this could be a valid thing.

But you are right, most power=plant should have landuse=industral or for small things building tag. So perhaps drop power=plant and station for promoting landuse/building.

kreed:
> I would expect that power=plant would be anyway tagged as landuse=industrial. Are there cases of correctly tagged places where rendering power=plant like landuse=industrial would improve situation?

It would be helpful in cases where the power plants are located inside of larger industrial areas. Adding a landuse=industrial tag to such plants would be redundant.

About one-third of all power plants are currently tagged with landuse=industrial (I'm analyzing the [extract from here](http://enipedia.tudelft.nl/wiki/Extracting_Power_Data_from_OpenStreetMap), since the tag isn't common enough for taginfo statistics).

```
power=# SELECT count(*) FROM planet_osm_polygon WHERE power='plant';                                                                                                                
 count 
-------
  2592

power=# SELECT count(*) FROM planet_osm_polygon WHERE power='plant' AND landuse='industrial';                                                                                       
 count 
-------
   867
```

> Also, #1835 mentions that power=station should not be rendered, maybe it should be removed from rendering?

Yes, power=station is problematic because of ambiguous tagging. Both power plants and power substations are tagged with it. I would prefer to deal with it in a separate PR though.

HolgerJeromin:
thanks i did not thought about a plant within an industrial complex.

i just fixed a few substations which were tagged as `power=plant` (while searching `power=plant and landuse!=industrial` with overpass). 8-/
I cannot think of a rendering (aka an icon which is plant, but no substation) to prevent this.

matkoniecz:
> plant within an industrial complex

In that case rendering only name would be better. My test found many mistagged substations and individual power plants without tagged landuse=industrial.

Rendering area (like landuse=industrial) hides that some data is missing - for example see https://cloud.githubusercontent.com/assets/899988/9902546/92f733ee-5c6f-11e5-8e17-32d9056488b6.png

kreed:
Good point. That sounds like a good solution. I'll update the code.

kreed:
Updated. I also reordered the coalesce statement so that the power color is used over the industrial color when both tags are present.

matkoniecz:
If you are interested in making more pull requests - #887 is very similar.

Issue #1765, opened 2015-08-18
unpaved footways on rocky areas are poorly visible


matkoniecz:
http://www.openstreetmap.org/?mlat=47.56673&mlon=12.32377#map=19/47.56673/12.32377

![](https://cloud.githubusercontent.com/assets/13821303/9294554/2e8d82a4-4451-11e5-8a5d-1c23413a1d84.png)

>But there are minor issues due to low contrast on certain background textures (i.e. scree, bare_rock) 

@geowOSM - it is better to open new issues for new issues rather than comment om closed pull requests. Commenting in open PRs is a good idea as somebody is currently working on proposal and feedback is valuable. But comments in closed PR are likely to be lost.

imagico:
Also due to the relatively wide casing and the weak line unpaved ways look very different on dark background and on bright background which makes them difficult to recognize as the same styling across different environments.

geowOSM:
>But comments in closed PR are likely to be lost.

Thanks for recovering my suggestion as a new issue ;)

daganzdaanda:
I think this picture shows that unpaved paths/footways are now too weak to be useful. Even on simpler backgrounds, I think they need to be stronger.
In #1766 a progression was suggested, and that might be a solution: 
paved=solid line or long dashes, 
unknown=dot-dash, 
unpaved=dots (but strong like paved "Eggersteig" here)

matkoniecz:
> Even on simpler backgrounds, I think they need to be stronger.

Can you give specific examples? In my testing I missed rocky areas but in many other cases it worked well (so either I missed also something else or my opinion about visibility is different).

daganzdaanda:
Just a bit north from the example area: http://www.openstreetmap.org/#map=16/47.5840/12.3048
![unbenannt](https://cloud.githubusercontent.com/assets/7693297/9345276/8e9dfb50-460f-11e5-9770-533f1c7a4a14.GIF)
In my opinion, the unpaved paths are not visible enough on empty ground or on meadows. I always try to check how things disappear when I close my eyes halfways...
Also on farmland: http://www.openstreetmap.org/#map=17/47.83054/12.07816
![unbenannt](https://cloud.githubusercontent.com/assets/7693297/9345393/9fde932e-4610-11e5-8525-1819befea97e.GIF)

daganzdaanda:
If I read it correctly, the line in the background has opacity 0.4:
`background/line-opacity: 0.4;`
[...]

PR #1837, opened 2015-09-15
render tracks as solid line on z13, z14 - fixes #1591

### LINKED ISSUE #1591: make tracks more prominent
Please make tracks more prominent on zoom 14 and 13 by adding the glow effect as it is on zoom 15+.
On zoom 14 the tracks are very hard to see eg. in forests mostly because of the pattern used (#965).

Zoom 15+ is OK:

Zoom 14 is the least readable:

Zoom 13 is better than 14 because of the missing forest texture, but still should be improved by adding the glow effect as it is on the zoom 15+:

matkoniecz:
idea by @imagico
start using @track-grade[12]-width-z15

Makes map less busy and crowded, makes tracks visible more than footways, geometry of tracks is more readable, removed dead code. No longer fully displaying tracktype at z13 z14 (what anyway was hard to read except in places with long, straight tracks) is a bad side of this change.

https://cloud.githubusercontent.com/assets/899988/9870378/ad86a6f2-5b8b-11e5-8ea7-41c4dfbc2c79.png (waterway=dam glitch is a data error that I already fixed)

https://cloud.githubusercontent.com/assets/899988/9870380/ad885650-5b8b-11e5-8d52-5e881a836d10.png

https://cloud.githubusercontent.com/assets/899988/9870379/ad8814ce-5b8b-11e5-8721-cfe37aab5112.png

https://cloud.githubusercontent.com/assets/899988/9870381/ad895a78-5b8b-11e5-94d9-acec5c760f72.png

https://cloud.githubusercontent.com/assets/899988/9870382/ad8d548e-5b8b-11e5-9dd9-ad0c077851f9.png

https://cloud.githubusercontent.com/assets/899988/9870383/ad8e6400-5b8b-11e5-9736-75ffdd776fd8.png

### DIFF
--- a/roads.mss
+++ b/roads.mss
@@ -77,8 +77,8 @@
 @footway-width-z13:               0.7;
 @cycleway-width-z13:              0.7;
 @path-width-z13:                  0.2;
-@track-width-z13:                 0.5;
-@track-grade1-width-z13:          0.5;
+@track-width-z13:                 0.3;
+@track-grade1-width-z13:          0.6;
 @track-grade2-width-z13:          0.5;
 @steps-width-z13:                 0.7;
 
@@ -102,8 +102,8 @@
 @cycleway-width-z15:              0.9;
 @path-width-z15:                  0.5;
 @track-width-z15:                 1.5;
-@track-grade1-width-z15:          0.75;
-@track-grade2-width-z15:          0.75;
+@track-grade1-width-z15:          1.5;
+@track-grade2-width-z15:          1.5;
 @steps-width-z15:                 3;
 
 @residential-width-z16:          11.2;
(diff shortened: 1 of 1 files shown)


imagico:
That is slightly thicker than in my example (which used line-width 0.25) which probably means it will appear stronger than highway=service on dark backgrounds at z13 - see example on https://github.com/gravitystorm/openstreetmap-carto/issues/1591#issuecomment-131622631 at the bottom right.  Maybe using a smaller width at z13 than at z14 would be a good idea - this would also weaken the abrupt change in appearance between z12 and z13.

In principle the same idea could also work on footway/cycleways - see #1748, but this will not really be feasible without solving the color issues first - see also https://github.com/gravitystorm/openstreetmap-carto/issues/1765#issuecomment-134529603

math1985:
I like the way the new z13/z14 look, but I don't like the user experience when zooming from z14 to z15. What used to be a solid line  (which gives a relatively important impression) suddenly becomes a dashed, much less important line. It gives the feeling that what looks to be a nice paved road from far away suddenly becomes a muddy track when having a closer look. I'd be inclined to reject this PR.

imagico:
I think this is an inevitable consequence of less differentiated rendering at lower zooms.  We are very used to the primary measure of data reduction at low zooms being the omission of features but in principle unifying different renderings is a proven technique as well and this always leads to emphasizing some things and weakening others.  IMO there is nothing wrong with that.

Two things could be considered though:

* generally making the lines weaker thinner (i already suggested that)
* using slightly different colors depending on track grade.  The idea is creating a kind of color mixture between the track line color and the casing color (which dominates the higher track grades at the higher zooms).  Keeping the tracks well visible at z14 and avoiding a reduction in visual strength from z14 to z15 at the same time will hardly be possible though.

matkoniecz:
Also, not rendering grade5, grade4 tracktypes at z13, z14 would be weird with rendering footways at the same time.

kocio-pl:
I think @imagico has a point: it's nice to be consistent, but not at all cost. Look at the airport: it's a plane icon on lower zoom, which disappears on higher zoom levels and instead there are some violet areas. Or the trees which are just a green points which become brown dots with green circle around. So we have examples of different rendering on different zoom levels and I think this is a valid solution.

imagico:
Does this mean the general idea of rendering tracks without dashing at the lower zooms is rejected?

matkoniecz:
No, though it would be necessary to address problems mentioned here (track vs service, probably less wide track for grade5, grade4).

I am not entirely sure whatever it will work.

PR #1838, opened 2015-09-16
stop rendering landuse labels from nodes - fixes #1100

### LINKED ISSUE #1100: Names from landuse nodes are displayed
See http://overpass-turbo.eu/s/5Ea for an example. Discovered on https://josm.openstreetmap.de/ticket/10679#comment:5

There is no reason to display name from landuse=forest node on any zoom level.

matkoniecz:
also natural=wood is affected (as duplicate of landuse=forest)

https://cloud.githubusercontent.com/assets/899988/9897752/53fde836-5c4c-11e5-918a-e97c20c80746.png (note - in that example bogus node is already deleted)


imagico:
I am not sure this is a good idea.  [The wiki](http://wiki.openstreetmap.org/wiki/Key:landuse) considers most landuse values on nodes correct and if you ignore for a moment that the landuse key implies an area in some way it makes a lot of sense (more than just inventing a new redundant tag to map these things as nodes).  We are so used to high resolution imagery allowing mapping everything as area but there are many cases where this is not possible and it is better to have nodes than to urge mappers to draw bogus polygons.

Your example is of course wrong mapping - the same feature is mapped both as node and area - but this is not the primary use case for landuse tags on nodes.

dieterdreist:
sent from a phone

> Am 16.09.2015 um 09:54 schrieb Christoph Hormann <notifications@github.com>:
> 
> I am not sure this is a good idea. The wiki considers most landuse values on nodes correct and if you ignore for a moment that the landuse key implies an area in some way it makes a lot of sense


how could we ignore that landuse is about areas? It doesn't make any sense on linear ways or nodes, unless you use it for other stuff than landuse (e.g. parts of settlements which should better go into "place")



> (more than just inventing a new redundant tag to map these things as nodes)


IMHO these tags wouldn't be redundant but would describe what is actually tagged.

dieterdreist:
sent from a phone

> Am 16.09.2015 um 09:54 schrieb Christoph Hormann <notifications@github.com>:
> 
> The wiki considers most landuse values on nodes correct


the wiki is inconsistent, if you see the general landuse key definition, only areas are valid, and if you look at the individual value definitions (e.g. residential), also only areas are valid. It is just the overview page you linked that has a different idea about this.

imagico:
I don't think this is the place to discuss how to map a reservoir, quarry, cemetery etc. if you only have a location but it is quite clear to me that current mapping practice supported by the wiki is to tag a node with landuse=*.

kocio-pl:
@imagico I also think it makes sense when one doesn't know how the boundaries look exactly.

matkoniecz:
At this point I am planning to close this PR and #1100 as rejected.

PR #1839, opened 2015-09-17
Adding line breaks in YAML to make it easier to maintain

kocio-pl:
Resolves https://github.com/gravitystorm/openstreetmap-carto/issues/1802.

I have made additional line breaks to the YAML code, as planned.

Basic check was made with oneliner stripping the line breaks and spaces:

```SH
LC_ALL=C tr -d '[:blank:]\n' < project.yaml.old > project.yaml.old.stripped
LC_ALL=C tr -d '[:blank:]\n' < project.yaml.new > project.yaml.new.stripped
```

and diff didn't show the difference between stripped files. 

Simple TileMill test has also shown no visible problems.

Any comments?


HolgerJeromin:
>  Perhaps a description of this philosophy could be added on top of the yaml file as a comment.

is missing right now, but this would prevent the nice `tr` check.

kocio-pl:
Sure, I just try not to break anything in the first place and let people examine if this is OK in general, but also prefer to have a short explanation on top of the file.

By the way - file length is still sane: 

```SH
$ wc -l project.yaml.new 
2479 project.yaml.new

$ wc -l project.yaml.old
2306 project.yaml.old
```

math1985:
Much better, thanks!

matkoniecz:
Much better, I support this change.

IMHO some lines are still too long - for example

>               'highway_' || (CASE WHEN highway IN ('residential', 'unclassified', 'pedestrian', 'service', 'footway', 'cycleway', 'track', 'path', 'platform') THEN highway ELSE NULL END)),


>               ('natural_' || (CASE WHEN "natural" IN ('beach', 'heath', 'grassland', 'wood', 'sand', 'scree', 'shingle', 'bare_rock', 'scrub') THEN "natural" ELSE NULL END)) AS "natural",


>              ('wetland_' || (CASE WHEN "natural" IN ('wetland', 'marsh', 'mud') THEN (CASE WHEN "natural" IN ('marsh', 'mud') THEN "natural" ELSE wetland END) ELSE NULL END)) AS wetland,

HolgerJeromin:
The lines matkoniecz complains are 200 chars long. 
In a github diff about 120 chars are shown (without scrolling). 
So imo all code lines should be around shorter than that to have some space for adding some entries.

kocio-pl:
I was just trying not to make changes which could annoy people, following in general what @gravitystorm [said about it](https://github.com/gravitystorm/openstreetmap-carto/issues/1802#issuecomment-138809013):

> For the individual line lengths, they should be limited to around 160-200 characters, or 5-10 items in a "foo IN ('bar', 'baz', ...)" style list, but it's not worth having hard limit or being too exact.

This PR is for testing this idea, so if we decide to break lines after ~100 chars, I will just make another version. For example during this exercise I have realized that preserving blocks indentation is part of the problem, because in case of amenities we loose 50 chars just for this!

```YAML
              ('amenity_' || (CASE WHEN amenity IN ('parking', 'bicycle_parking', 'motorcycle_parking', 'university', 'college', 'school', 
                                                    'hospital', 'kindergarten', 'grave_yard', 'prison', 'place_of_worship', 'clinic') 
                                                    THEN amenity ELSE NULL END)) AS amenity,
```

So maybe dealing with lists we should break this general rule, let's say like this:

```YAML
              ('amenity_' || (CASE WHEN amenity IN ('parking', 'bicycle_parking', 
'motorcycle_parking', 'university', 'college', 'school', 'hospital', 'kindergarten', 
'grave_yard', 'prison', 'place_of_worship', 'clinic') 
THEN amenity ELSE NULL END)) AS amenity,
```

or this (lists don't mix with the code in the same line):

```YAML
              ('amenity_' || (CASE WHEN amenity IN 
('parking', 'bicycle_parking', 'motorcycle_parking', 'university', 'college', 'school', 
'hospital', 'kindergarten', 'grave_yard', 'prison', 'place_of_worship', 'clinic') 
THEN amenity ELSE NULL END)) AS amenity,
```

or similar take (the code is even more coherent: WHEN-THEN are on the same position):

```YAML
              ('amenity_' || (CASE WHEN amenity IN 
('parking', 'bicycle_parking', 'motorcycle_parking', 'university', 'college', 'school', 
'hospital', 'kindergarten', 'grave_yard', 'prison', 'place_of_worship', 'clinic') 
                                   THEN amenity ELSE NULL END)) AS amenity,
```

What would be good with such solutions is that reflowing the list would be much easier, because we wouldn't have to strip and add indentation spaces.

I work with the data inside the lists, so I didn't try too hard to fix code around them, but if some coding style guru :wink: has some remarks, I'd love to hear it.

HolgerJeromin:
Another idea:
```
              ('amenity_' || (CASE WHEN amenity IN 
                      ('parking', 'bicycle_parking', 'motorcycle_parking', 'university', 'college', 
                      'school', 'hospital', 'kindergarten', 'grave_yard', 'prison', 'place_of_worship',
                      'clinic')
                  THEN amenity ELSE NULL END)) AS amenity,
```
Or the bracket is "part of the SQL" stuff and not "part of the list". 
```
              ('amenity_' || (CASE WHEN amenity IN (
                      'parking', 'bicycle_parking', 'motorcycle_parking', 'university', 'college', 
                      'school', 'hospital', 'kindergarten', 'grave_yard', 'prison', 'place_of_worship',
                      'clinic'
                  )THEN amenity ELSE NULL END)) AS amenity,
```

matkoniecz:
According to my test rendering is not changed. 

@gravitystorm @pnorman Is there any reason to not merge it? I would prefer to do it soon as it would conflict with nearly any PR affecting project.yaml.

kocio-pl:
Hold on a second: I like to know first if we want 100-chars breaks or we want to stick with the current version? I think my last proposition (bigger lists start from the beginning of a new line) would be better, so I plan to change it like that soon, but I don't know if we also want everything to be more or less visible in a GitHub diff. What's your opinions?

Making line breaks is a top priority for me now, so whatever will change in YAML due to some other PRs, I plan to update this one at least before planned merging.

matkoniecz:
> bigger lists start from the beginning of a new line

I am not expecting that would be better (reducing linecount will not be outweighed be worse readability).

kocio-pl:
That's why I don't like to be in a hurry with this issue - the jury is still out. I'd like to see more votes or just having at least a few days of silence (which would mean that nobody else is interested in details of line breaking and the rest is happy with just breaking them at all).

matkoniecz:
> I like to know first if we want 100-chars breaks or we want to stick with the current version?

I may check how shorter line would work. Is this version made by some clever automation or is it result of manually adding linebreaks?

kocio-pl:
Fully manual.

gravitystorm:
I'm happy with the approach shown in the PR. I don't think it needs to be as short as 100 characters, as I originally said.

PR #1736, opened 2015-08-11
New road style - rework road colours, road widths and display of railway=rail

matkoniecz:
Currently used road style has some problems, one the most obvious one is that secondary and trunk colour too similar to landuse colours (#102).

But there are also other serious problems with basing design on UK road style.

Marking different road types by colours in completely different hues works well for maps displaying primarily roads, with limited different styles for other features. This map differentiates wide range of landovers, POIs, borders and names. It is using more different styles than most maps. For example picture below presents road types on some of landcovers displayed in this style. 

In this situation colours of many road types are closer to other features than to other road types (forest - trunk, motorway - river, tertiary - sand etc). That is not happening for example on Ordnance Survey map where number of differently displayed features is strictly limited.

selection_001 2
selection_003

Also, UK colour style is not universally known. For many people hierarchy of red, blue and green roads is not obvious.

This new road style is using narrower range of hues, making road of different classes more similar what should eliminate possibility of confusing them with other features and makes easier to avoid collisions on changing rendering of other elements. Steady progression of hue and lightness for major road types (motorway, trunk, primary, secondary, tertiary) should make more intuitive which roads are more important.

Road colours were also tuned to ensure that roads are well visible on all landcovers.

Roads are now narrower, to improve map readability and make it prettier. Also, more road types are differentiated by width rather than by colour (white, wide tertiary).

Railways are now more prominent and service tag for railways has bigger impact on rendering.

Color of pedestrian and living_street is now more logical and intuitive, without making map uglier.

Fixes #102 (secondary and trunk colour too similar to landuse colours)

Fixes #631 (tertiary roads more dominant than secondary on z10)

Fixes #1272 (junction=motorway_junction names are hard to read)

Fixes #319 (motorways on low zoom levels are very hard to notice and look like rivers)
world z 6 7 _ master -_ gsoc 53 357 -6 416 6 7 master - gsoc 200px

Fixes #914 (motorway tunnels are nearly invisible on z13 and lower)

Fixes #1124 (make road-casing stronger)

Fixes #1769 (improve living street rendering)

Reduces impact of #286 (some streets that are not joining may on zoom out misleadingly appear to be joining)

Before using version containing changes from this PR on OSM website it is recommended to change map key for Default map layer.

Among potential problems are that it is now harder to recognise road type of given road, especially in situation where there is no possibility to compare it with other road types.

Such significant change will be confusing for current users of this style.

UK color coding of roads is well known for many people, for them a new style - even assuming that it would be intuitive for them - will be less useful.

Issue #706, opened 2014-07-07
oneway arrows of nearby roads covering each other can be misleading


matkoniecz:
http://www.openstreetmap.org/?mlat=50.05523&mlon=19.94447#map=18/50.05523/19.94447

![przechwytywanie](https://cloud.githubusercontent.com/assets/899988/3492821/93ae8fb0-05b0-11e4-87b6-450d0e19b86c.PNG)

Residential is not oneway, private service roads rendered below are.

Related to #597 #681 #245

Fixing #284 would probably also fix most occurrences of this bug.

matkoniecz:
And sometimes results are really misleading - see http://www.openstreetmap.org/#map=19/50.07165/19.94518

![selection_001](https://cloud.githubusercontent.com/assets/899988/13286972/c7b3f04a-db04-11e5-86fe-174adf96b277.png)
![selection_002](https://cloud.githubusercontent.com/assets/899988/13286973/c7b510ec-db04-11e5-99cb-8b19209d99f0.png)

cody-somerville:
Slightly different manifestation of this issue. [Example ](https://www.openstreetmap.org/#map=19/43.65259/-79.38354)where oneway arrows get rendered outside of the typical visual context of a road entirely. 

![underground_oneways_arrow_rendering](https://github.com/gravitystorm/openstreetmap-carto/assets/356861/c1589455-26e7-4f2b-a678-eeda844b544d)

The arrows jump around at different zoom levels which without the visual context of the parent road element, due to being underground, is especially confusing until you realize what's happening as they seem like they're meaningfully pointing and moving to draw attention as you zoom.

(I notice the label "PATH"  (by The Archer) for an underground walkway is also rendering seemingly orphaned. Similar or different issue?)

Also, would it make sense to tweak the issue title? I initially thought this was going to be related to roads tagged covered=yes such as building passages.

matkoniecz:
> Also, would it make sense to tweak the issue title? I initially thought this was going to be related to roads tagged covered=yes such as building passages.

Can you propose a better one?

imagico:
By the way this is also closely related to #2213 and #4401.

Issue #284, opened 2013-12-12
Oneway arrows placed too close to start of way


matkoniecz:
Sometimes oneway arrows are placed in such location that it is unclear which route is oneway.

Examples:

http://www.openstreetmap.org/?mlat=50.07119&mlon=19.93001#map=17/50.07119/19.93001
http://www.openstreetmap.org/?mlat=50.07117&mlon=19.93014#map=18/50.07117/19.93014
http://www.openstreetmap.org/?mlat=50.07116&mlon=19.93014#map=19/50.07116/19.93014
http://www.openstreetmap.org/?mlat=50.07137&mlon=19.92762#map=19/50.07137/19.92762
http://www.openstreetmap.org/?mlat=50.07137&mlon=19.92763#map=18/50.07137/19.92763
http://www.openstreetmap.org/?mlat=50.07139&mlon=19.92773#map=17/50.07139/19.92773
http://www.openstreetmap.org/?mlat=50.0713&mlon=19.9279#map=16/50.0713/19.9279
http://www.openstreetmap.org/?mlat=50.0712&mlon=19.9298#map=16/50.0712/19.9298

matkoniecz:
According to my testing it is solved by #1840

PR #1840, opened 2015-09-17
rework oneway arrows

### LINKED ISSUE #595: Minimum number of oneway arrows
The following issue has been moved over from [trac](https://trac.openstreetmap.org/ticket/4338). It needs more discussion as it is not clear how this should be implemented.

Please make sure there is a minimum number of oneway arrows visible (on high zoom levels).

E.g. [here](http://www.openstreetmap.org/?lat=51.431769&lon=6.766953&zoom=18&layers=M) on the oneway street "Vom-Rath-Straße" there is only one arrow at the north end that can be overlooked easily. In high zoom levels (17,18) there is enough space for a(n) additional arrows(s) e.g. at the south end.
Alternatively you may consider drawing a more transparent arrow under the street name - but only if there is not enough space for the previous solution.

### LINKED ISSUE #597: Confusing arrows when oneway construction overlaps oneway highway
The following issue has been moved over from [trac](https://trac.openstreetmap.org/ticket/5085). It needs more discussion as it is not clear how this should be implemented.

If a road overlaps with a construction road, and both roads are oneway but in opposite direction, it is not clear which arrows refer to which roads.
Example: https://trac.openstreetmap.org/raw-attachment/ticket/5085/oneway_arrows%20of%20construction%20highways.png

The original bug report suggests the following layering stack

    construction highway
    construction oneway arrows
    normal highway
    normal oneway arrows 

However, that would create a lot of overhead in the code.

We could also decide to drop rendering arrows on construction roads altogether, or to do nothing at all.

### LINKED ISSUE #606: Oneway arrows in print
Oneway arrows are visibly chunky when in print

image

image

### LINKED ISSUE #681: Oneway arrows overlap way label
As seen in http://www.openstreetmap.org/note/191801 but there's probably other cases.

It would be nice to always have an arrow rendered right next to the label (with RTL handling for added challenge :p). Sounds like it could be tricky to implement though, as we already have so many arrow placement nitpicks.

nebulon42:
This PR moves rendering of oneway arrows from a dasharray representation towards a SVG marker symbolizer. This has the disadvantage of losing the ability to have bent arrows but results in a clearer representation, improves arrow positioning and avoids overlap with road labels. Additionally, rendering arrows for construction roads are dropped and arrows are coloured corresponding to either the casing colour or the fill colour of the road type they are rendered on, which leads to more visual consistency.

Fixes #606 by using SVG markers.
Fixes #681 as using markers prevents collision with labels.
Fixes #597 by dropping arrows for `highway=construction`.

Might fix #595 and #284, but I cannot definitely confirm, testing is welcome.

I think there is no change for #706.

Possible problems: roundabouts?

Previews (all examples within Vienna, Austria or close by)

motorway
motorway

trunk
trunk

primary
primary

secondary
secondary

tertiary/residential/unclassified/road/living_street/pedestrian
tertiary_residential_living_street_pedestrian

service/raceway
service_raceway

track
track

footway/path
footway

cycleway
cycleway

steps
steps

------- outdated, only for reference --------

Minor tweak to the new road style that was discussed e.g. in https://github.com/gravitystorm/openstreetmap-carto/pull/1736#issuecomment-135539161.

Instead of using the same colour for all road types chroma and hue for the arrows are taken from the casing colour of the respective road type, residential/tertiary is kept unchanged. This leads to minor differences in colour, but increases visual consistency.

Previews:
motorway
https://cloud.githubusercontent.com/assets/3531092/9944605/b48a5364-5d88-11e5-83f9-83fd25588277.png

trunk
https://cloud.githubusercontent.com/assets/3531092/9944607/b74515f8-5d88-11e5-8ca6-17ce30431f0a.png

primary
https://cloud.githubusercontent.com/assets/3531092/9944608/b976ddde-5d88-11e5-81ff-be0ef0a0e0d9.png

secondary
https://cloud.githubusercontent.com/assets/3531092/9944612/bdba7f0e-5d88-11e5-9e50-625d1a32b400.png

residential/tertiary
https://cloud.githubusercontent.com/assets/3531092/9944616/c2ffca28-5d88-11e5-8100-776bab241ec6.png

### DIFF
--- a/roads.mss
+++ b/roads.mss
@@ -238,9 +238,24 @@
 @paths-bridge-casing-width:       0.5;
 @paths-tunnel-casing-width:       1;
 
-@oneway-arrow-color: #404040;
-@junction-text-color: #960000;
+@junction-text-color:             #960000;
 @halo-color-for-minor-road: white;
+
+@motorway-oneway-arrow-color:     darken(@motorway-casing, 25%);
+@trunk-oneway-arrow-color:        darken(@trunk-casing, 25%);
+@primary-oneway-arrow-color:      darken(@primary-casing, 15%);
+@secondary-oneway-arrow-color:    darken(@secondary-casing, 10%);
+@tertiary-oneway-arrow-color:     darken(@tertiary-casing, 30%);
+@residential-oneway-arrow-color:  darken(@residential-casing, 40%);
+@living-street-oneway-arrow-color: darken(@residential-casing, 30%);
+@pedestrian-oneway-arrow-color:   darken(@pedestrian-casing, 25%);
+@raceway-oneway-arrow-color:      darken(@raceway-fill, 50%);
+@footway-oneway-arrow-color:      darken(@footway-fill, 35%);
+@steps-oneway-arrow-color:        darken(@steps-fill, 35%);
+@cycleway-oneway-arrow-color:     darken(@cycleway-fill, 25%);
+@track-oneway-arrow-color:        darken(@track-fill, 15%);
+@bridleway-oneway-arrow-color:    darken(@track-fill, 10%);
+
 
 .roads-casing, .bridges-casing, .tunnels-casing {
   ::casing {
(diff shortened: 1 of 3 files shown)


matkoniecz:
What about highway=cycleway/track/footway/living_street/pedestrian ?

matkoniecz:
Overall this change is OK for me, though motorway arrows seem less noticeable (on the other hand oneway arrows on motorways are not so important).

gravitystorm:
I like the concept of tying the colours together. My feeling is that the arrows are too dark, perhaps an L closer to the casing would work better?

It has also reminded me how terribly hard-to-read the arrowheads are!

nebulon42:
Maybe using a SVG marker would help? (no colour change yet)
arrow-symbol

HolgerJeromin:
the svg markers are nice. What is the minimal supported version of mapnik for them?

matkoniecz:
Personally I would prefer darker arrows than presented here rather than lighter (though note, I did no testing).

Also, generating oneway colours in https://github.com/gravitystorm/openstreetmap-carto/blob/master/scripts/generate_road_colours.py may be desirable (easier tweaking/changing road colors).

matkoniecz:
Svg marker would solve #606

math1985:
Related: #595 #284 #681 #597 #706 #467.

nebulon42:
@HolgerJeromin I do not see compatibility issues as it's a marker symbolizer (just like POIs) with marker-placement: line.

I think I will go down that road because the marker symbolizer approach seems to solve #284 and #681, although I'm not sure if it might be a Mapnik 3 positioning thing in my setup.

pnorman:
> Maybe using a SVG marker would help? (no colour change yet)

Is the change in visibility the SVG marker, or because you made part narrower?

With dasharrays for arrows we get arrows that are curved on curves, which we won't get with SVGs

nebulon42:
> Is the change in visibility the SVG marker, or because you made part narrower?

The preview depicts the marker. I tried tweaking the dash-array arrows, but it's a bit complicated. I don't know why they are appearing so thick as the base line width is only 1. With this approach we also will always have a jagged arrow head.

> With dasharrays for arrows we get arrows that are curved on curves, which we won't get with SVGs

Yes, that is something that we would lose. But currently I see far better positioning with symbolizers. I have updated the branch with the SVG markers for reference. The pull request is not finished yet.

nebulon42:
Implementation is complete, please test and give your feedback. Thanks.

pnorman:
> I tried tweaking the dash-array arrows, but it's a bit complicated. I don't know why they are appearing so thick as the base line width is only 1. With this approach we also will always have a jagged arrow head.

Did you try reducing the width?

nebulon42:
It would be possible to improve the arrow also by tweaking the dasharray as you can see below:
dasharray

I'm not sure if we could get rid of the jagged head by moving from bevel to miter joins. I have currently no set-up to do a SVG export and see how it looks. #284 could be solved by adjusting the start gap of the arrow (currently 12) to a much higher value, let's say half the space before and half the space after. But I think with this approach there is no easy way to solve #681.

Roundabouts do not necessarily look better with the dasharray approach.
dasharray
roundabout_dasharray

SVG
roundabout_svg

(Something is wrong with my dasharray implementation as arrows point in the opposite direction, but you get the idea.)

math1985:
See also this comment by @gravitystorm: https://github.com/gravitystorm/openstreetmap-carto/issues/606#issuecomment-45509017

Personally I'd prefer to move to svg arrows.

matkoniecz:
After initial review of how new style works - colour change is not particularly noticeable. New shape is way more readable. Loss of curving of arrow seems to not have a significant effect.

I noticed that positioning of arrows is significantly changed - I still have to check how new positioning works but it seems smarter. Oneway arrows are now rarer, overall without reducing information In some case ways properly showed before as oneway are no longer marked but also some ways previously not marked as oneway are now with arrows. Arrows disappeared from really short ways solving #284.

Overall it is an improvement though I would prefer to test further and check whatever it is possible to tweak how positioning works - is it possible to increase how many oneway arrows are displayed?

nebulon42:
> is it possible to increase how many oneway arrows are displayed?

The spacing is controlled by `marker-spacing`, I have set it to 180 which is a bit higher than the ~150 used by the dasharray approach. Additionally, there is `marker-max-error`, which controls the flexibility of the renderer in handling collisions or how strictly `marker-spacing` is interpreted. I oriented myself a bit on how OSM Bright handles this.

kocio-pl:
> After initial review of how new style works - colour change is not particularly noticeable. New shape is way more readable. Loss of curving of arrow seems to not have a significant effect.

:+1:

I would like to have at least 2 arrows on the opposite sides of roundabouts (more important) and more visible arrows on tracks (less important), but in general I am pleased with the result.

math1985:
Thanks! As always, further tweaks are welcome, but I think the general idea is good enough to merge.

I like the crispness of the new arrows.

I also see the problem with arrows on roundabouts. Google seems to solve this in part by making the 'line' part of the arrow very short - maybe we also can reduce this? We could also decide to drop arrows on roundabouts.

gravitystorm:
> We could also decide to drop arrows on roundabouts.

Arrows on roundabouts are good mapper feedback. When mapdust was a thing I found 10 bugs all on one roundabout with no explanation, and it took a while to figure out it was the wrong way around.

PR #1843, opened 2015-09-18
Do not sort placenames-medium by category (is yet done in placenames.mss)

sommerluk:
Do not sort placenames-medium by category (is yet done in placenames.mss)

As @math1985 has described in #1461 the sorting of placesnames-medium is currently done twice: in the .yaml file and in placenames.mss. As it is necessary to keep the sorting in placenames.mss (for proper rendering), the solution is to remove the sorting from .yaml, as done by this PR.


sommerluk:
This PR should not cause any change in rendering. However, it might be worth to verify this before merging.

PR #1846, opened 2015-09-19
Doc improvements

pnorman:
* Reference Kosmtik
* Rework yaml->mml
* Fix some caps

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -19,9 +19,14 @@ the issue should contain the tagging of the object.
 
 ## Editing Layers
 
-OpenStreetMap Carto uses a YAML file for defining layers. Some of the rationale is outlined in [a GitHub issue](https://github.com/gravitystorm/openstreetmap-carto/issues/711). Editing multi-line SQL statements in a YAML file is much friendlier than editing escaped SQL in a JSON file.
+OpenStreetMap Carto uses a YAML file for defining layers. Some of the rationale
+is outlined in [a GitHub issue](https://github.com/gravitystorm/openstreetmap-carto/issues/711).
+Editing multi-line SQL statements in a YAML file is much friendlier than editing
+escaped SQL in a JSON file.
 
-This requires a preprocessing step to convert the YAML to JSON. A script is provided at `scripts/yaml2mml.py`, which depends on PyYAML, available through `pip install pyyaml` or packaged on Ubuntu as `python-yaml`.
+The `./scripts/yaml2mml.py` script is provided to convert YAML to JSON, and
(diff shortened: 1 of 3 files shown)


matkoniecz:
After this change "Current work" section will be about past ("The v2.x series initially focused") without mention what is currently done. Section about future mentions reviewing bug reports as a current task what is only small part of what is currently done.

> Avoid adding an overwhelming number of screenshots.

I will try to avoid doing this again :)

kocio-pl:
I'm not sure this remark is needed: nobody else is doing it and Mateusz had a perfect reasons to do it, so it's a just-in-case rule probably, which should be avoided to not make documentation too detailed.

math1985:
> Avoid adding an overwhelming number of screenshots.

+1 for leaving this out. I'd prefer too many screenshots over too little.

CONTRIBUTING.md
```diff
@@ -10,15 +10,45 @@ You should fork the project into your own repo, create a topic branch
...
+`./scripts/yaml2mml.py < project.yaml > project.mml && touch project.mml` to
+update the file and force TileMill to reload it. You shouldn't use the text editor
+built-in to TileMill, it doesn't work with the number of .mss files in the style.
+Instead, hide the right pane and use an external text editor.
+
+Before committing changes, run `./scripts/yaml2mml.py < project.yaml > project.mml`
+to update the JSON MML file and `git add project.mml`
+
+When committing changes, add both the `project.yaml` and `project.mml` files to
+the commit with ./scripts/yaml2mml.py < project.yaml > project.mml && git add project.mml`.
```
math1985: Aren't these two lines duplicate with the previous two lines?

README.md
```diff
@@ -45,23 +47,18 @@ no interest in creating further point releases in the v1.x series.
...
-
-There are over [300 open requests][issues], some that have been open for years. These need
-reviewing and dividing into obvious fixes, or additional new features that need some cartographic
-judgement. The work done already in v1.0 and v2.0 will make it much easier to process these.
+There are over [300 open requests][issues], some that have been open for years.
+These need reviewing and dividing into obvious fixes, or additional new features
+that need some cartographic judgement.
 
 [issues]: https://github.com/gravitystorm/openstreetmap-carto/issues
 [cleverness]: https://github.com/openstreetmap/mapnik-stylesheets/blob/master/inc/settings.xml.inc.template#L16
```
math1985: It's not clear to me what this refers to?

math1985:
@pnorman Thanks, very useful!

README.md
pnorman: These are existing, and are for the links above.

math1985: But why the link to settings.xml.inc.template (not changed within the current diff actually)?

PR #1848, opened 2015-09-19
Revert "make generate_road_colours script executable"

pnorman:
Reverts gravitystorm/openstreetmap-carto#1845

As it is, the script needs to be run through python, and this would try to run it as a shell script


HolgerJeromin:
We could add [`#!/usr/bin/env python`](http://stackoverflow.com/questions/6908143/should-i-put-shebang-in-python-scripts-and-what-form-should-it-take) as a shebang

Issue #1313, opened 2015-02-14
Arabic and Hebrew POI names don't show


daganzdaanda:
I just noticed that there were many icons without names in the middle east. Then I checked, and the POIs do have names but they don't get rendered. Seems that Arabic and Hebrew are afflicted. I checked some other places with non-roman scripts, and did not find a problem.

math1985:
@daganzdaanda Do you have an example?

@mkoniecz Might this be a text-dy problem? Are Arabic and Hebrew taller than É?

mkoniecz:
@math1985 No idea, example would be helpful. It sounds like a RTL problem ( http://en.wikipedia.org/wiki/Right-to-left ).

daganzdaanda:
Argh, I had a loooong post already, and then the browser crashed >:(

For Hebrew: Any town center in Israel, e.g. this area https://www.openstreetmap.org/#map=19/31.78109/35.21898
For Arabic: look in Teheran, e.g. https://www.openstreetmap.org/#map=18/35.68707/51.41875

The museums in Teheran make it look likely that it has to do with text-dy: some labels are there, some not, and it changes with zoom.
Hebrew seems to miss even more labels, maybe it needs even more text-dy.
Just found this: https://www.openstreetmap.org/#map=19/32.07437/34.80938 
Two pharmacies with hebrew names, one is labeled.

Some POIs seem to be (mostly?) OK though: bakery, bank, generic shop. But I wouldn't count on that...

mkoniecz:
It seems to be text-dy issue with new icons. I am now running text-dy validator for v2.28.1.

math1985:
> It seems to be text-dy issue with new icons. I am now running text-dy validator for v2.28.1.

No, I think it is a mapnik bug where RTL text is rendered too high.

math1985:
Reported upstream at https://github.com/mapnik/mapnik/issues/2705.

math1985:
Prefixing the string to be displayed with chr(8206) (LEFT-TO-RIGHT MARK) resolves the problem but reintroduces https://github.com/mapnik/mapnik/issues/189.

pnorman:
I don't think there's anything we can do on our end until it's fixed in a released version of Mapnik

math1985:
I agree.

Do we leave bugs that need to be fixed upstream open, or do we close them?

Did anyone check if Mapnik 3 solves this issue?

mkoniecz:
[...]

PR #1850, opened 2015-09-19
Increase label offset (necessary for hebrew and arab)

sommerluk:
The labels for features like peaks (like http://www.openstreetmap.org/node/278475926 for example) is currently not rendered when the name is in some languages like hebrew or arab. This PR increses text-dy, so that the label gets rendered.

### DIFF
--- a/amenity-points.mss
+++ b/amenity-points.mss
@@ -1195,8 +1195,7 @@
     text-size: 10;
     text-fill: darken(#d08f55, 30%);
     [feature = 'natural_volcano'] { text-fill: #d40000; }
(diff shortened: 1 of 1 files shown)


matkoniecz:
Can you give examples for names where this happen? And AFAIK it is related to a Mapnik bug - see #1313

sommerluk:
Yes, it seems that it is exactly the problem described in #1313.

As this PR is more a dirty hack than a real solution, closing in favour of #1313

math1985:
We plan to migrate to Mapnik 3 soon, which should solve this problem.

PR #1856, opened 2015-09-20
render man_made=bridge like man_made=pier

nebulon42:
For me it seemed a good idea to render `man_made=bridge` like land colour. I see the rationale by emphasising the man_made aspect, but as we also render `man_made=pier` like land colour it seems logical to also render bridges like that.

In the preview note the pier between the two bridges.

before
man_made_bridge_before

after
man_made_bridge

### DIFF
--- a/buildings.mss
+++ b/buildings.mss
@@ -47,6 +47,8 @@
 
 #bridge {
   [zoom >= 12] {
-    polygon-fill: #B8B8B8;
+    polygon-fill: @land-color;
+    [way_pixels >= 4]  { polygon-gamma: 0.75; }
+    [way_pixels >= 64] { polygon-gamma: 0.3;  }
   }
 }


nebulon42:
There is a flaw in my argument regarding bridges not spanning water. I have to think about that, maybe a outline would work.

sommerluk:
I would consider the tagging in the above example as wrong. Wouldn’t the bridge area be supposed to cover the hole bridge, and not only the part above the water? So the bridge area should cover up to the node of the streets where the highway=* +bridge=yes ends …

nebulon42:
Yes, true.

nebulon42:
Ok, I admit that I didn't think enough about it. As I'm not convinced by the outline approach
bridge_outline
I'm closing this.

dieterdreist:
sent from a phone

> Am 20.09.2015 um 13:41 schrieb Michael Glanznig <notifications@github.com>:
> 
> man_made aspect, but as we also render man_made=pier like land colour it seems logical to also render bridges like that.
> 


I prefer the building like style. Piers and bridges are quite different beasts, as there is space below bridges.

Also note that the bridges in your example seem strange, I'd use man_made bridge on the whole bridge, not just on the "floating" part

PR #1861, opened 2015-09-22
Increase width of roads at high zoom

vholten:
In the recently merged new road style, roads are rendered quite narrow at zoom levels 18 and 19, especially residential and unclassified roads. At zoom 19, these roads are rendered much narrower than their real width. I propose to slightly increase the width of roads at zoom 18, and more so at zoom 19. At zoom 19, also the label font size is increased slightly.

See also the discussion at #1290.

Zoom 18 before / after:
brooklyn z18
Zoom 19 before / after:
brooklyn z19

Zoom 18 before / after:
brooklyn 2 z18
Zoom 19 before / after:
brooklyn 2 z19

Zoom 18 before / after:
krakow z18
Zoom 19 before / after:
krakow z19

### DIFF
--- a/roads.mss
+++ b/roads.mss
@@ -170,15 +170,31 @@
 @service-width-z17:               7;
 @minor-service-width-z17:         3.5;
 
-@motorway-width-z18:             20;
-@motorway-link-width-z18:        12;
-@trunk-width-z18:                20;
-@primary-width-z18:              20;
-@secondary-width-z18:            20;
-@tertiary-width-z18:             20;
-@road-width-z18:                  8;
-@service-width-z18:               8;
-@minor-service-width-z18:         4.25;
+@motorway-width-z18:             23;
+@motorway-link-width-z18:        14;
+@trunk-width-z18:                23;
+@primary-width-z18:              23;
+@secondary-width-z18:            23;
+@tertiary-width-z18:             23;
+@residential-width-z18:          14;
+@living-street-width-z18:        14;
+@pedestrian-width-z18:           14;
+@road-width-z18:                  9;
+@service-width-z18:               9;
+@minor-service-width-z18:         5;
+
+@motorway-width-z19:             33;
+@motorway-link-width-z19:        20;
+@trunk-width-z19:                33;
+@primary-width-z19:              32;
+@secondary-width-z19:            32;
+@tertiary-width-z19:             32;
+@residential-width-z19:          21;
+@living-street-width-z19:        21;
+@pedestrian-width-z19:           21;
+@road-width-z19:                 13;
+@service-width-z19:              13;
+@minor-service-width-z19:         7;
 
 @footway-width-z18:               1.3;
 @cycleway-width-z18:              1;
@@ -218,6 +234,10 @@
 @secondary-casing-width-z18:      1;
 @major-casing-width-z18:          1;
 
+@casing-width-z19:                0.8;
+@secondary-casing-width-z19:      1;
+@major-casing-width-z19:          1;
+
 @bridge-casing-width-z12:         0.1;
 @major-bridge-casing-width-z12:   0.5;
 @bridge-casing-width-z13:         0.5;
@@ -232,6 +252,8 @@
 @major-bridge-casing-width-z17:   1;
 @bridge-casing-width-z18:         0.8;
 @major-bridge-casing-width-z18:   1;
+@bridge-casing-width-z19:         0.8;
+@major-bridge-casing-width-z19:   1;
 
 @paths-background-width:          1;
 @paths-bridge-casing-width:       0.5;
@@ -278,12 +300,14 @@
         [zoom >= 15] { line-width: @motorway-width-z15; }
         [zoom >= 17] { line-width: @motorway-width-z17; }
         [zoom >= 18] { line-width: @motorway-width-z18; }
+        [zoom >= 19] { line-width: @motorway-width-z19; }
         [link = 'yes'] {
           line-width: @motorway-link-width-z12;
           [zoom >= 13] { line-width: @motorway-link-width-z13; }
           [zoom >= 15] { line-width: @motorway-link-width-z15; }
           [zoom >= 17] { line-width: @motorway-link-width-z17; }
           [zoom >= 18] { line-width: @motorway-link-width-z18; }
+          [zoom >= 19] { line-width: @motorway-link-width-z19; }
         }
         line-color: @motorway-low-zoom-casing;
         [zoom >= 13] {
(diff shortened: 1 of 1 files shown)


ximex:
:+1:

imagico:
Better also test at low latitudes, one of the primary reasons for narrowing the roads was readability in such areas:

http://www.openstreetmap.org/#map=18/1.29669/103.85381
http://www.openstreetmap.org/#map=18/14.59796/120.98173

The alternative would be to use variable width at the high zooms like in #1853.

Circeus:
Yes please. In many areas, this can cause some serious visual issues for areas that connected to these roads. [Example](http://www.openstreetmap.org/#map=19/45.51520/-73.56155) (cf. [actual distance](https://commons.wikimedia.org/wiki/File:Grande_Bibliotheque_02.jpg))

mboeringa:
I agree this needs more testing, and will probably, and especially, not work at Z18. 

While this may be OK in "US" type cities with extremely wide roads, many historic city centres in Europe have very narrow roads, and the current rendering may already cause issues. Current rendering of roads already overlaps building contours in many places at Z18.

E.g. see these samples at Z18:
**Austria, Vienna:**
http://www.openstreetmap.org/#map=18/48.20995/16.36565
**France, Mont Saint Michel:**
http://www.openstreetmap.org/#map=18/48.63582/-1.51101
**Italy, Siena:**
http://www.openstreetmap.org/#map=18/43.31833/11.33061
**Italy, Florence:**
http://www.openstreetmap.org/#map=18/43.76959/11.25241

vholten:
Thanks for the comments. Here are the examples from the requested low-latitude areas.

Singapore, zoom 18:
singapore z18

Manila, zoom 18:
manila z18

@mboeringa : At zoom 18, the residential roads in this PR are actually narrower than in the currently deployed osm-carto style, so I expect that the situation in the historic city centres in Europe will not become worse.

Style | Width of residential at zoom 18
------------ | -------------
osm-carto (original road style) | 15.5
gsoc (new road style) | 12
this PR | 14

Example: Siena at zoom 18, comparing osm-carto (original road style) and this PR:
siena z18

imagico:
Thanks - to me this is a clear improvement for z19, for z18 i have no clear preference, there are clearly situations where your change improve things but also some where it is worse.

math1985:
Thanks! To me it looks like a clear improvement.

kocio-pl:
It looks like this code was too optimistic and I heard feedback from people finding blobs and glitches (like [here](https://lists.openstreetmap.org/pipermail/talk/2015-November/074683.html) or my example [here](https://www.openstreetmap.org/node/3248982182#map=18/52.21969/21.01376)).

I think we need more conservative approach: it's more acceptable if the road is a bit too narrow than a bit too wide. Ultimately we could use [street area proposition](http://wiki.openstreetmap.org/wiki/Proposed_features/Street_area) to make it look exactly like it is in reality, so we don't have to overuse general tools.

I'm not sure however if we should entirely revert this change or it would be enough to just make the numbers smaller than currently.

sommerluk:
> it's more acceptable if the road is a bit too narrow than a bit too wide.

Well, almost all roads are rendered much wider than their real width on all zoom levels lower than 15 or 16. You propose doing exactly the opposite on z18 z19. That would volountary introduce an inconsistency …

kocio-pl:
Exaggeration on lower zoom levels is (most probably) used to see the transportation routes on longer distances, but on higher zoom levels they are clearly visible and there's no more need to oversize them to see the route. 

We currently use the width to make the difference between road priority/class just because we don't have enough colors available, and that is already risky, but hard to avoid. Further increasing it is unnecessary try to catch up with reality. Because road class are not tightly connected to their width, it will always be faulty, but on higher zoom levels it hurts the eyes more if we loose details because of it (consider [this example](https://www.openstreetmap.org/way/35008147#map=17/52.24755/21.00607) with a building partially hidden by the road and a footway completely eclipsed, even on z19).

While I fully support the purpose of this PR and I hoped it will not lead to problems on highest zoom levels, the evidence shows that it works worse than expected.

math1985:
> It looks like this code was too optimistic and I heard feedback from people finding blobs and glitches (like here or my example here).

+1, I'd be happy to accept a PR to decrease the road widths again.

vholten:
I could submit a PR that partially undoes the change, say by 50%. How does that sound?

imagico:
Note the problem observed here is exactly why i suggested the variable width rendering in #1290.  If you cap the rendering width using the tagged width value with a conservative road class based estimate as default - similar to as it is done in #1853 - this will avoid such problems and will seamlessly transit to a symbolic width rendering at the lower zooms.

kocio-pl:
@vholten In case of such idea (guessing minimal universal value) the most important part is testing, not the idea. :smile: We could start with 50% of the increase and see the results.

@imagico I don't get the difference between your proposition and current take. I thought this PR was exactly about variable increasing road width ("I propose to slightly increase the width of roads at zoom 18, and more so at zoom 19.")?

imagico:
> I thought this PR was exactly about variable increasing road width

No, with this PR the drawing width is still constant globally for each zoom level.  #1290 suggests something fundamentally different.

kocio-pl:
Unfortunately, I still don't get the difference, but in reality that's not important, because (as I said) the testing is king here. If your code works better than currently, I have nothing against it.

kocio-pl:
@imagico Are you interested in creating PR according to your vision? @vholten can probably make "50%" PR with no big effort and the bonus would be we could compare both solutions, so I would be happy to test both of them.

imagico:
> @imagico Are you interested in creating PR according to your vision?

It is not a vision, it is a suggestion.  But the answer is no, i won't work on this at the moment.  #1853 demonstrates how this can be done, it is probably not difficult to adapt this for the roads although it will certainly be quite a lot of work with all the  special cases that occur in the roads layer.

If there is a change in roads rendering that uses polygon patterns to distinguish paved/unpaved (as previously suggested in https://github.com/gravitystorm/openstreetmap-carto/issues/110#issuecomment-116224523) that would probably simplify implementing this as well.

math1985:
Variable width road rendering introduces a lot of complexity that we should careful consider. It would likely also influence a lot of other things, like access rendering, label sizes, etc.

I would therefore propose to fix this problem initially with a simple solution (for example the 50% idea), so that we can include it in the next release, and consider possible variable width road rendering at a later moment.

imagico:
> Variable width road rendering introduces a lot of complexity that we should careful consider. It would likely also influence a lot of other things, like access rendering, label sizes, etc.

Yes, note however the range of actual drawing widths of the roads will not significantly exceed what has been tried so far - except for the higher latitudes at z19 where they would likely be somewhat wider.

vholten:
The "50%" PR is now ready for review (#1972).

PR #1862, opened 2015-09-22
Changing water features color to already defined water-text variable

kocio-pl:
Simple code cleaning to use the color variable name (water-text) instead of hardcoded hex numbers and compacting similar text label entries (natural_bay and natural_spring).

### DIFF
--- a/amenity-points.mss
+++ b/amenity-points.mss
@@ -1684,27 +1684,19 @@
     }
   }
 
-  [feature = 'natural_bay'][zoom >= 14] {
-    text-name: "[name]";
-    text-size: 10;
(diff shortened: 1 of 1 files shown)


matkoniecz:
I am not sure about compacting, currently code is similar but its function is different (one for feature that is a big area, second for much smaller ones).

It will reduce code but may make future tweaking harder.

I am already running test for the old version (only using water-text variable) and I am thinking about merging it rather than version also with unification.

kocio-pl:
Do as you like! I just though it makes sense to compact it, but I don't really mind.

PR #1864, opened 2015-09-22
Adding amenity=bicycle_repair_station icon

kocio-pl:
This adds the icon for amenity=bicycle_repair_station icon at "small amenities" zoom level (z>=18) with low priority in amenity-brown color. TagInfo shows 1k+ uses and this makes bicycle facilities more complete.

The icon itself is made of elements taken from bicycle parking and car service, of course:
bicycle-repair-19


pnorman:
Working version of #1353, which was rejected.

I'm still skeptical, for the same reasons as before.

The usage has increased, but is still low, and about half the usage is from an import.

dieterdreist:
I'm in favor, the icon is easily recognizable and doesn't introduce different symbols because it consists of patterns we already use

kocio-pl:
I had no idea it was already proposed, but it was so simple to create that I just forgot to check...

I think this is similar case as with motorcycle parking, so let me just link my previous comments: https://github.com/gravitystorm/openstreetmap-carto/pull/1650#issuecomment-125539705, https://github.com/gravitystorm/openstreetmap-carto/pull/1650#issuecomment-125780308. In a nutshell: I tend to think about the "cluster" of things, so if we're able, we should render every common item related to given type of transport. I still don't have a tyres shop icon designed for example, but it would make car "cluster" more complete.

The interesting note is however that about imports. How should I check how many mappers have used it and what is the imports rate? I don't try to rely on the numbers, but for now I can only check the number of uses and 1k + wiki page assures me it is solid. Motorcycle parking has only 0,2k, so for now I just haven't looked at it, but it also is similar and has a wiki page.

math1985:
Do you also want to render the name tag of this object? Currently the queries for the name fields are updated, but in the .mss file it doesn't display a name now...

kocio-pl:
If I understand correctly, [this is intentional hack](https://github.com/gravitystorm/openstreetmap-carto/blob/master/project.yaml#L2010). For example telephone also doesn't have its label rendered, yet [it's included in the same section](https://github.com/gravitystorm/openstreetmap-carto/blob/master/project.yaml#L2026).

math1985:
I think this is a tag with a rather specific audience and still relatively low usage. I'm therefore going to reject this PR. Sorry!

Issue #131, opened 2013-08-23
Add rendering for wayside_shrine


al--:
Wish: Please render historic=wayside_cross and wayside_shrine, they are landmarks.

wayside_cross could be rendered much like any christian place of worship:
![wayside_cross](https://f.cloud.github.com/assets/1458456/1014622/79128aa2-0bc0-11e3-916a-cb6e2ca4dae3.png)

wayside_shrine symbol could be something like ![wayside_shrine](https://f.cloud.github.com/assets/1458456/1014624/8c8bff28-0bc0-11e3-8961-1438beefe2c7.png)

kocio-pl:
My proposition for shrine (also 9 px as cross made by @matkoniecz):
1) ![shrine-9-11](https://cloud.githubusercontent.com/assets/5439713/8932747/531a7ac6-353f-11e5-8e5f-d68eaea8a45b.png)

matkoniecz:
@kocio-pl 
Shrines need also symbols for other religion values. At least for religion=* missing (over 80% of entries), religion values except 'christian' are quite rare.

kocio-pl:
First has only vertical stick left, the second one is just a box with no stick on top:
2) ![shrine-9-12](https://cloud.githubusercontent.com/assets/5439713/8937027/29d10ac4-3556-11e5-81c9-0dfe702f4b9c.png)
3) ![shrine-9-13](https://cloud.githubusercontent.com/assets/5439713/8937029/2ef7cc54-3556-11e5-8e29-f39b844ce0f6.png)

kocio-pl:
I guess 2 is better as a general shrine (with 1 for christian), so I'll try to make some mockups soon.

Penegal:
If it helps, another possibility for wayside cross, widespread in France : ![wayside cross](https://cloud.githubusercontent.com/assets/13777824/9875618/c1a4e616-5baf-11e5-8ff5-040fbdae2fe1.png)

matkoniecz:
> My proposition for shrine (also 9 px as cross made by @matkoniecz):

https://github.com/gravitystorm/openstreetmap-carto/issues/131#issuecomment-125615202 - can you upload a SVG file (or is it only a png mockup)?

kocio-pl:
@matkoniecz [Here you are](https://github.com/kocio-pl/openstreetmap-carto/blob/icons/symbols/shrine-9.svg).

matkoniecz:
[...]

PR #1865, opened 2015-09-22
render wayside_shrine the same way as wayside_cross

xsuchy:
Wayside shrine is just more robust form of wayside cross.
Since it is bigger and more visible in real life, it is shame
that we display cross and not shrine.
Until somebody provides better icon, it is good enough to render it
as cross (as sometime it is called by local people cross anyway).

See
http://wiki.openstreetmap.org/wiki/Tag:historic=wayside_shrine
http://wiki.openstreetmap.org/wiki/Tag:historic%3Dwayside_cross


matkoniecz:
What about shrines for other religions?

math1985:
@xsuchy You forgot to update project.mml, you can do so automatically with ./scripts/yaml2mml.py < project.yaml > project.mml.

xsuchy:
@math1985 fixed
@kocio-pl  (although deleted) it should look like http://mtbmap.cz/#zoom=18&lat=49.218494&lon=16.387597 This is little renderer because I find that wayside_cross are in fact not rendered - and I can swear that I seen it rendered in past. E.g here http://www.openstreetmap.org/#map=18/49.48159/16.48200 is wayside cross and it is not rendered despite the fact that with current carto it should be rendered. Strange.
@matkoniecz I would say: "small steps". This is IMO step forward. If someone want to change it for other religion the he can send another PR. But I guess that will be long discussion. Well, it already is. See https://github.com/gravitystorm/openstreetmap-carto/issues/131

matkoniecz:
> despite the fact that with current carto it should be rendered

OSM servers are not using latest openstreetmap-carto version (latest released version is not including new road style and later changes, including rendering of crosses).

Rendering only wayside_shrine with religion=christian with cross symbol would not be bad (though I am not sure is it a good idea). But rendering [historic=wayside_shrine, religion=shinto] with cross symbol (examples findable with http://overpass-turbo.eu/s/bC8 ) is completely unacceptable and I oppose merging this PR in the current form.

xsuchy:
Is this better?

matkoniecz:
I think that assuming that religion is certainly Christian is a poor idea. Also - is it tested?

Also, I would dispute

> Wayside shrine is just more robust form of wayside cross.

It is very similar in function but it is far away from "just more robust form of wayside cross"

kocio-pl:
@matkoniecz +1.

dieterdreist:
@matkoniecz +1

xsuchy:
So should I remove line with `[religion = '']`?

matkoniecz:
Yes, I would also remove "Wayside shrine is just more robust form of wayside cross." from the commit description.

xsuchy:
Updated according your comments.

math1985:
@xsuchy I get the following error message:

" ERROR: column "religion" does not exist LINE 1: ...","feature","is_building","name","operator","ref","religion"... ^ "

I think you also need to add religion to the SQL statement.

Did you test this in Kosmtik or Tilemill?

math1985:
I still can't get this to work unfortunately. @xsuchy if you can get it to work, feel free to submit a new PR!

Issue #1783, opened 2015-08-24
Coherent color system for the POIs


kocio-pl:
When trying to add [new icon for amenity=car_wash](https://github.com/gravitystorm/openstreetmap-carto/pull/1777) I have noticed that we have also shop=car_repair and I had the problem [what color to use](https://github.com/gravitystorm/openstreetmap-carto/pull/1778), because they are very similar (paid services for cars), yet they have different key. 

After [asking on Tagging](https://lists.openstreetmap.org/pipermail/tagging/2015-August/026102.html) I've learned that it's more complicated than that and in fact [all the "amenities" which sell goods or services should be shops instead](https://lists.openstreetmap.org/pipermail/tagging/2015-August/026105.html). This is far outside the rendering scope, but it means we should define our guidelines on POI colors.

While I think using purple for health related POIs (be it for people or animals, goods or services, paid or not) and blue for transportation or traveling (paid or not, directly or indirectly like hotel or embassy probably) is pretty clear, and I like black to be used as a general color (for landmarks for example), I'm not sure which system should we use for amenities and shops:

1) Treat them as they're tagged, no matter what. I don't like it, because we don't take things blind in other cases and car wash/car service would look rather strange with different colors.

2) Make it as it should be tagged - shop-violet for anything selling goods or services, amenity-brown for the rest. It would mean making the cinemas and theaters look like shops, because they typically are selling services (through tickets), but also all the eating and drinking places. This would cause changes in a lot of places.

3) Make violet for selling goods (I don't know what about giving goods? maybe it's just kind of service then?) and brown for all the services (paid or not). That is what we have now more or less and car wash/car service would be both brown, but also we should change the color of hairdresser/beauty shop to brown. This would also work [for craft and office POIs](https://github.com/gravitystorm/openstreetmap-carto/issues/1697), once we have them.

I would choose 3, what do you think about it?

math1985:
Question for the native English speakers - would hairdressers and car_repair shops be typically included in the term 'retail'?

kocio-pl:
Another observed color problem: leisure=swimming_pool has brown icon and green text, while golf icon is light green with darker green text. Probably coherent green color for icon+text could be used for all sport activities (not for the shop=sports however).

SK53:
@math1985 yeah I include services in retail, as does Pete Reed (tlalet).

[...]

PR #1868, opened 2015-09-23
Changing color of service shop=* POIs to amenity-brown

kocio-pl:
The same story as https://github.com/gravitystorm/openstreetmap-carto/issues/1642 and https://github.com/gravitystorm/openstreetmap-carto/pull/1778.

Changing color of POIs which are not selling goods, but rather services:
* shop=beauty
* shop=dry_cleaning
* shop=hairdresser
* shop=laundry 
* shop=travel_agency

brown-beauty-laundry
brown-travel-hairdresser-19

### DIFF
--- a/amenity-points.mss
+++ b/amenity-points.mss
@@ -533,7 +533,7 @@
 
   [feature = 'shop_beauty'][zoom >= 17] {
     marker-file: url('symbols/beauty-14.svg');
-    marker-fill: @shop-icon;
+    marker-fill: @amenity-brown;
     marker-placement: interior;
     marker-clip: false;
   }
@@ -627,7 +627,7 @@
       marker-file: url('symbols/laundry-14.svg');
       marker-placement: interior;
       marker-clip: false;
-      marker-fill: @shop-icon;
+      marker-fill: @amenity-brown;
     }
   }
 
@@ -669,7 +669,7 @@
     marker-file: url('symbols/shop_hairdresser.16.svg');
     marker-placement: interior;
     marker-clip: false;
-    marker-fill: @shop-icon;
+    marker-fill: @amenity-brown;
   }
 
   [feature = 'shop_hifi'][zoom >= 17] {
@@ -829,7 +829,7 @@
     marker-file: url('symbols/travel_agency-14.svg');
     marker-placement: interior;
     marker-clip: false;
-    marker-fill: @shop-icon;
+    marker-fill: @amenity-brown;
   }
 
   [feature = 'shop_stationery'][zoom >= 17] {
@@ -2039,7 +2039,12 @@
       text-halo-fill: rgba(255, 255, 255, 0.6);
       text-wrap-width: @standard-wrap-width;
       text-placement: interior;
-      [feature = 'shop_car_repair'] {
+      [feature = 'shop_car_repair'],
+      [feature = 'shop_beauty'],
+      [feature = 'shop_dry_cleaning'],
+      [feature = 'shop_hairdresser'],
+      [feature = 'shop_laundry'], 
+      [feature = 'shop_travel_agency'] {
         text-fill: @amenity-brown;
       }
     }


math1985:
Personally I'd much prefer to keep these (and other similar retail activities) in pink.

matkoniecz:
I have no personal preference and argumentation for switch to brown is quite convincing.

kocio-pl:
@math1985 I think we should have some easy to use coloring system; separating selling goods from all the services - with transportation and health facilities being special - works for me. Otherwise the question is: when talking about "these (and other similar retail activities)" which ones do you mean exactly?

We could just stick to the shop/amenity keys, of course, but @gravitystorm [advised not to rely on external choices](https://github.com/gravitystorm/openstreetmap-carto/pull/1778#issuecomment-134185951) and while I would have nothing against external classification, the tagging keys are not clear enough - they are used carelessly, but even worse: we don't even know [what is the difference between amenity=* and shop=*](https://lists.openstreetmap.org/pipermail/tagging/2015-August/026102.html).

For me shop=hairdresser should be colored as a service - [shop violet would rather belong to shop=wig](https://github.com/gravitystorm/openstreetmap-carto/pull/1778#issuecomment-140734914).

math1985:
Personally, I would prefer a colour scheme that makes clear in one glance what kind of areas there are in a given city: a retail area with purple icons, an entertainment area with brown icons, etc.

I would rather group hairdressers with chemists than with pubs.

kocio-pl:
1. I guess landuse=retail is better suited for this task.
2. This would become hairy very soon: retail is purple, entertainment is brown - but the police/fire station or monuments are far from the entertainment, so what is a common thread for them? I would rather use black as a general color (building names, traffic lights) and not overload brown with such a different meanings.

imagico:
> Personally, I would prefer a colour scheme that makes clear in one glance what kind of areas there are in a given city: a retail area with purple icons, an entertainment area with brown icons, etc.

Yes, for me that would be social vs. commercial things.  Of course restaurants and pubs are both but for the typical map user the social aspect is dominating.  And the goods vs. services division is not any clearer.

dieterdreist:
sent from a phone

> Am 23.09.2015 um 22:36 schrieb kocio-pl <notifications@github.com>:
> 
> I guess landuse=retail is better suited for this task.


the current landuse values are heavily influenced by zoning and the postwar idea to have a separation of functions (sleeping, working, shopping, ...), which is just one competing concept in urbanism, and isn't how most of the European cities are organized. Typically there is a dense mix of residential, retail, commercial, cultural use, etc. in European city centres.

HolgerJeromin:
This style is also for the feedback to the mappers. 
With current style a mapper knows which is a key when he wants to tag something new. 
Is this aspect unimportant with the presets in most editor? 
I am not sure. 
Perhaps this is my very technical view, adding most first tag without presets, but using suggestions in the text fields.

kocio-pl:
@dieterdreist You're right, this is a fundamental problem with landuses - they are mixed in many cases.

@imagico This is very interesting proposition, because it is also a neat system and I currently try to test how it works. I'm not bound to the goods/services/* idea, I just think it's quite easy to tell and general enough. 

Commercial (retail) and social is much harder for me when thinking about all amenities we have. The example with restaurants and pubs are good - it's possible to think about them this way too given the choice. However if we think about the system, we should also review also not obvious examples before: 
1) post office and a post box
2) memorial and monument
3) view point
4) recycling point
5) bench
6) drinking water point
7) fast food

I guess this is where your system breaks. That's not to say it's out of question - it's just that they are not commercial nor social and we should decide what to do with them. My intuition is we have "neutral" black for general things and we could use them as a fallback in both systems.

kocio-pl:
@HolgerJeromin 

> With current style a mapper knows which is a key when he wants to tag something new. 
 
That's [what I was talking about](https://github.com/gravitystorm/openstreetmap-carto/pull/1868#issuecomment-142710287): in fact they don't know, because we don't have a clear distinction between those keys. That's exactly why I don't like to rely on it.

kocio-pl:
I would also change the shop=copyshop to brown (based on the service rule).

See also more general discussion on system for POI colors at https://github.com/gravitystorm/openstreetmap-carto/issues/1783.

math1985:
In the end I suppose this comes down to personal preference, and as nobody else seems to chime in, I guess I need to make a decision based on my own personal preference. I'd prefer places like hairdressers and dry cleaners to stay in retail colour, so I'm going to close this PR. Thanks for the PR though, and sorry for the long wait!

kocio-pl:
Thanks for cleaning obviously dead proposals (like this)!

Issue #1813, opened 2015-09-05
shields on roads may be smaller and less prominent


matkoniecz:
Currently shields are too big and too prominent. Map would look better with smaller one.

http://tools.geofabrik.de/mc/#11/50.0563/21.0123&num=2&mt0=mapnik&mt1=bing-map

![selection_001](https://cloud.githubusercontent.com/assets/899988/9698888/b387b98c-53cb-11e5-9817-994b6bde1e9e.png)

http://tools.geofabrik.de/mc/#11/50.0563/21.0123&num=2&mt0=mapnik&mt1=nokia-map (it shows shields for more roads and thanks to better style it is less dominating)

![selection_002](https://cloud.githubusercontent.com/assets/899988/9698890/b4c3a158-53cb-11e5-8c45-d5e19ffa0f81.png)

#951 is making situation even worse

nebulon42:
It might be already quite an improvement if the white border around shields would be dropped or at least would be converted into a halo.

nebulon42:
Quick and dirty proof of concept:
![shields](https://cloud.githubusercontent.com/assets/3531092/9704752/b15c11bc-54b1-11e5-9eb6-fe2ae57ba0f2.png)

Here is the corresponding branch for ref: https://github.com/nebulon42/openstreetmap-carto/tree/road-shields

More tweaks might include slightly smaller font (not bold?) and less padding between ref and shield.

nebulon42:
refined example with new generation script that allows the specification of padding:
![shields](https://cloud.githubusercontent.com/assets/3531092/9706697/450ea4d8-54ee-11e5-868a-79439c1a3ac4.png)

Here I decreased the font size to 9 and specified 2px padding (this seems to be not enough padding).

mboeringa:
Seems pretty decent. I am just not to sure about the current spacing between shields on a single road, they are to close in my opinion, but that is another issue...

nebulon42:
Yes, #951 exactly. :)

nebulon42:
@pnorman Do you see any issues by moving shield generation from Shell-Script to Python?

pnorman:
> @pnorman Do you see any issues by moving shield generation from Shell-Script to Python?

Not really, but I don't see advantages either

HolgerJeromin:
Python would be executable on Windows, too. I think more windows user have python than a bash like shell.

kocio-pl:
:+1:

pnorman:
[...]

PR #1871, opened 2015-09-24
rework road shields

nebulon42:
Tries to solve #1813.

Changes the graphical representation of road shields by removing the white border, adding a slightly darker outline, reducing border radius, decreasing font size, and adding a text halo for better readability of the reference.

Colours are corresponding to highway fills and casings. The script to generate road colours is not used as I did not get why the colour components are modified in that particular way and order.

Introduces a new Python shield generation script where it is possible to specify paddings instead of widths and heights by assuming default character dimensions. The script supersedes the three shell scripts currently used for shield (re-)generation.

Carto rules for shields are refactored to make use of variables for easier changes.

Preview
shields2


mboeringa:
Looks good. Only thing I can think of is to slightly darken all base colors of the shields, so as to have a slightly better readability of the text on especially the yellow shields.

matkoniecz:
> The script to generate road colours is not used as I did not get why the colour components are modified in that particular way and order.

Script version at https://github.com/matkoniecz/openstreetmap-carto/blob/clean/scripts/generate_road_colours.py more be more clear. See also comment in https://github.com/openstreetmap/iD/issues/2764#issuecomment-143283018

In general - each road type has its own hue, with lightness and chroma changing for every use - but generally more important roads are darker and more saturated.

What values are used is result of experiment and tweaking (the entire purpose of this script is to make tweaks like "make casings darker" easier - with this script it is enough to change)

start_l and end_l for

> line_colour_infos["casing"] = ColourInfo(start_l = 50, end_l = 50, start_c = 70, end_c = 55)
> line_colour_infos["low-zoom-casing"] = ColourInfo(start_l = 50, end_l = 70, start_c = 50, end_c = 65)

In case that it explained this script - what was the part that allowed it (after this and https://github.com/openstreetmap/iD/issues/2764#issuecomment-143259942 it is obvious that purpose of this script is not obvious and documenting it would be a good idea).

matkoniecz:
To reproduce values of

> @shield-motorway-halo-fill: rgba(176, 95, 110, 0.5); // Lch(50,35,10)
> @shield-trunk-halo-fill: rgba(194, 125, 105, 0.5); // Lch(59,33,42)
> @shield-primary-halo-fill: rgba(184, 146, 100, 0.5); // Lch(63,31,74)
> @shield-secondary-halo-fill: rgba(181, 179, 125, 0.5); // Lch(72,29,106)

one may use

> line_colour_infos["shield"] = ColourInfo(start_l = 50, end_l = 72 start_c = 35, end_c = 29)

what will generate

> @motorway-shield: #ae5e6e; // Lch(50,35,10), error 0.5
> @trunk-shield: #b97864; // Lch(57,33,42), error 0.7
> @primary-shield: #b79566; // Lch(64,31,74), error 1.7
> @secondary-shield: #aeb17a; // Lch(71,29,106), error 1.5

what is a bit different (colours are equally different, different name format) but generally the same and at least in theory makes further tweaking easier.

nebulon42:
Thanks for the explanation I'll look into that. For my shield versions I will need two colours, one for the fill and one for the outline, but should be easy to add one.

Any comments on the visual appearance of the shields?

nebulon42:
> Only thing I can think of is to slightly darken all base colors of the shields, so as to have a slightly better readability of the text on especially the yellow shields.

Maybe having a stronger/darker text halo would also help?

mboeringa:
> Maybe having a stronger/darker text halo would also help?

Maybe, try it and see how it looks. The problem is mainly with the yellow roads / shields. Sometimes small halos make the text less readable though, so it needs testing. In that case darkening the base color of the shield is better.

matkoniecz:
> For my shield versions I will need two colours, one for the fill and one for the outline, but should be easy to add one.

Just add next line like this: 

> line_colour_infos["shield-fill"] = ColourInfo(start_l = 50, end_l = 72 start_c = 35, end_c = 29)
> line_colour_infos["shield-outline"] = ColourInfo(start_l = 50, end_l = 72 start_c = 35, end_c = 29)

(and obviously, set wanted values)

matkoniecz:
From quick test - at least primary shields are darker and more prominent with this PR what in my opinion is rather making situation worse (as it makes shields even more prominent).

mboeringa:
> From quick test - at least primary shields are darker and more prominent with this PR **what in my opinion is rather making situation worse** (as it makes shields even more prominent).

To be honest, I have no idea what you are referring too... The shields are much smaller and less obtrusive than the current ones as used on OpenStreetMap website with the old and still current road styling, and none of the shields in the example in the first post appear to dark IMO...

For legibility, there needs to be some contrast between the white characters and the shield's base color.

math1985:
To me, the new shields look good.

matkoniecz:
Maybe my location is unlucky, but for me new version is more prominent than the current one.

krakow - name rondo ofiar katynia 50 0874518 19 8919875 way 13 13 master - nebulon_road-shields 300px

https://cloud.githubusercontent.com/assets/899988/10125083/624e1788-6568-11e5-885b-b628821d5fdb.png (bigger picture)

------

Also, after PR additional shields are visible making #951 worse - see for example bottom-left of 
https://cloud.githubusercontent.com/assets/899988/10125097/d77f4aa4-6568-11e5-9d1f-587ddbad2bed.png

mboeringa:
> Maybe my location is unlucky, but for me new version shields are more prominent than the current ones.

OK, that makes your critic clearer. The multi line versions make the slightly darker color stand out more. Maybe they indeed need to be switched back to original color, but with the new suggested layout.

nebulon42:
Thanks for testing. The amplification of #951 could IMO being solved by increasing shield-min-distance to a value where the smaller shield size is compensated. I plan to incorporate colours into the colour script and then do some colour tweaking.

vholten:
If for some shields the color needs to be too dark to make the white text readable, an option is to switch to black text (regular, not bold) and keep the shields lighter. Example, same area as @matkoniecz showed:

shields-black-text

Or text in the same hue as the road (bold in this example):

shields-dark-text

matkoniecz:
@vholten I like the first version, at least in this case it is better.

mboeringa:
These last options of @vholten look promising. I wonder, based on the high contrast version, if it would even be possible to reduce the size of the shields and font used further, so as to make them even less intrusive. I would really like to see them downsized by half or one point.

daganzdaanda:
Nice work! 
Dark text looks better. Only on motorway and maybe on trunk shields white is readable well enough. But, like @mboeringa says, with dark text we have enough readability to make the shields even smaller, which is good, IMHO.

nebulon42:
Here is an updated preview based on all suggestions (black font, bold face removed, smaller font size, lighter and smaller shields).

Kraków z13
shields

I'm not fully satisfied with the following:

 While I agree that the darker colour works better, I'm not convinced by black. Will try tailored hues to road types as suggested also by @vholten. I think those examples were a bit too light though.

Smaller shields work well on lower zooms, but get quickly quite small on higher zooms. Example (z16):
shields_z16

The only solution would be to either keep somewhat larger shields on lower zoom levels or to increase font size and shields on higher zooms (at e.g. some threshold zoom level). This would at least double our number of shield graphics. Since they are fairly small and are generated this would be okay I think.

matkoniecz:
Doubling number of svg files for shields would not be a big problem.

Though at least in my case small shields at high zoom are not a problem (but it probably would be a problem in regions where ref are more important or even replacing street names).

pnorman:
> While I agree that the darker colour works better, I'm not convinced by black. Will try tailored hues to road types as suggested also by @vholten. I think those examples were a bit too light though.

Might making the outer line on the shields stronger accomplish what you want?

nebulon42:
After quite some tweaking I'm now satisfied with the result. I have increased font sizes at z16 and z18 and therefore had to triple the amount of shield graphics, I'm not sure if this might be a problem. The text halo is now gone, the shield fill colour (slightly lighter and desaturated) as well as the outline colour (slightly darker) have changed. Text colour is now in the same hue as the road type, which works better IMO. Due to the reduced lightness this might not even be noticable, but leads to more pleasing result.

One question still to be answered is if we want to have the shields stay larger than the roads at all times or not, see e.g. the Kraków z18 example. If yes, we have to increase the shield and font sizes more.

I have slightly increased the shield spacing parameters to compensate for the smaller shield sizes, but the overall goal of this PR is not to improve shield positioning. This should be done separately.

Kraków previews
z13
shields_z13

z15
shields_z15

z16
shields_z16

z18
shields_z18

original preview, Stockerau z13
stockerau_z13

As I have modified the colour generation script, this conflicts with #1873.

mboeringa:
I like it. The shields in my opinion do not have to be bigger than roads at all times, like on high zoom. By keeping the size reduced, it leaves space for other stuff to display.

jojo4u:
On "774" tertiary shield i noticed a bit greenish tint on the shield background.

dieterdreist:
I just noticed that subjectively the S7 is now harder to read, at first glance it looked like 57, while this isn't an issue in the current rendering.

nebulon42:
Re "greenish" shield: Yes the yellow is closer to green than to red. I could experiment with saturation a bit, but I wanted to stick to the hue of the road.

Re readability: I see your concern, but I think that readability of the font is a bit out of scope of this PR. This would be a problem also for other labels, not only for shields.

pnorman:
Could the repetitions of MSS in roads.mss be reduced, similar to #1893?

nebulon42:
@pnorman The diff of #1893 was a bit confusing for me, but I have refactored the style definitions as I thought it might make sense. I actually tried this earlier, but it didn't work then. Don't know what I did wrong, this time it should be ok.

matkoniecz:
> Re readability: I see your concern, but I think that readability of the font is a bit out of scope of this PR. This would be a problem also for other labels, not only for shields. 

Readability is also influenced by text colour - in case of trunk it gets slightly worse. But in case of secondaries it gets better.

After initial review of the new version I am happy - overall readability is improved (trunks are exception and it is not much worse), shields are less prominent. It also seems that shield customization will be easier.

matkoniecz:
@nebulon42 It is bad also now, but ref on highway=residential should not be more prominent than ref on highway=tertiary.

selection_005

May be also done as additional PR.

nebulon42:
Yes, that is a problem. I didn't touch the residential ref so far. If you plan to merge this soon I'll rather would leave that as separate change otherwise I would have a look at it. I always felt that residential refs should not be rendered at all, but the corresponding issue was closed.

mboeringa:
I think they are exceedingly rare anyway at this point in time, so don't think it is big issue. Should be solvable later on.

nebulon42:
On second thought I'd rather not open up a new discussion regarding residential refs here, because the other changes seem to be quite "finished", improvements always possible :). I'll try to work on residential refs in a follow-up PR.

matkoniecz:
Sorry for a long review time.

matkoniecz:
And thanks for your work!

nebulon42:
Thanks.

kocio-pl:
It repairs the problem I have seen [here](https://www.openstreetmap.org/#map=15/52.2593/21.0400), so thank you too!

PR #1872, opened 2015-09-24
refactor yaml2mml.py (fixes #1866)

### LINKED ISSUE #1866: simplify yaml2mml.py execution
Currently yaml2mml.py needs the correct files provided to stdin and stdout. But given it's the same files every time, and the directing files to stdin/stdout isn't the worlds most common thing, we can simplify how it works.

* Currently: `python scripts/yaml2mml.py < project.yaml > project.mml`
* Target: `python scripts/yaml2mml.py` or even just `scripts/yaml2mml.py`

The files should be found relative to the script, not the current directory, so that `python scripts/yaml2mml.py` and `cd scripts; python yaml2mml.py` both work.

nebulon42:
Since I was at writing Python scripts I tried to fix #1866.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -34,16 +34,16 @@ and the JSON file just needs updating before committing.
 
 [TileMill](https://github.com/mapbox/tilemill) and Mapbox `carto` [do not directly support YAML](https://github.com/mapbox/carto/issues/401),
 so make edits to the YAML file then run the preprocessing step of
-`./scripts/yaml2mml.py < project.yaml > project.mml && touch project.mml` to
+`./scripts/yaml2mml.py && touch project.mml` to
 update the file and force TileMill to reload it. You shouldn't use the text editor
 built-in to TileMill, it doesn't work with the number of .mss files in the style.
 Instead, hide the right pane and use an external text editor.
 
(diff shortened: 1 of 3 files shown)


nebulon42:
I overlooked that travis_check_project_files has to be rewritten too. Will do that later.

HolgerJeromin:
Imo the `touch project.mml` could be dropped or integrated into the script. Disclaimer: I have no detailed knowledge about the reason for this in the first place.
Probably tilemill needs a new timestamp , but git has no problem with it either.

HolgerJeromin:
Perhaps the script should be written to support  both call logics. But perhaps not :-)

nebulon42:
Yes, I think it is a good idea to have the possibility to specify input and output with parameters if needed.

matkoniecz:
> Yes, I think it is a good idea to have the possibility to specify input and output with parameters if needed.


I am not sure. In this project this script always takes the same files as input/output. Making it easily reusable for completely different project seems like unneeded additional code.

nebulon42:
It's not about another project, it's about supporting reading from stdin and writing to stdout as well (or at least writing to stdout) to support travis_check_project_files.

nebulon42:
I have updated the PR to support the Travis check script.

math1985:
Thanks!

Issue #1834, opened 2015-09-14
tram_stop label is rendered earlier than bus_stop label


kocio-pl:
Currently tram_stop name is rendered too early, from z>=14 [like railway=halt and aerialway=station](https://github.com/gravitystorm/openstreetmap-carto/blob/master/stations.mss#L51), but I think it should be rather [the same as bus_stop](https://github.com/gravitystorm/openstreetmap-carto/blob/master/amenity-points.mss#L1762) (z>=17), because they are related closer to each other and now [tram stop names make unnecessary mess on the map](http://www.openstreetmap.org/#map=14/51.1127/17.0405).

matkoniecz:
AFAIK Poland has lower than average distance between tram stops, most likely showing from z17 would be too much.

And counterexample - in Kraków labels from z14 work fine - http://www.openstreetmap.org/#map=14/50.0733/19.9127

kocio-pl:
In my opinion it doesn't work even there, because on this zoom level I expect rather subdivisions of a city (like quarters/suburbs or even neighborhoods) than such small POIs, like on the Wikimedia map:

https://maps.wikimedia.org/#14/50.0733/19.9127

I also tend to think about the system: why tram stops should be more visible than bus stops? I find these two kind of transportation very similar (with metro/subway and railway halt/stations being different and shown earlier than them). Of course maybe there is a reason and I just don't see it.

ajank:
I would argue that tram stop labels definitely should be more visible than bus stop labels. Because of their much higher unit capacity, trams typically form the backbone of public transport system in smaller cities that do not have a subway or urban rail system. Moreover, the tram network is much more permanent and prominent in the urban space than the bus network. Hence, I propose to render tram stops labels at z>=15.

pnorman:
I support rendering tram stops earlier - that's not to say that z14 is or is not the right level, but it should be before bus stops

jojo4u:
Note that in absence of widespread rendering support of public_transport=stop_position/platform, highway=tram_stop should also used by the more tram-like light rail systems since railway=halt does not fit well. Railway=halt is defined only once per halt and often the halts of the Stadtbahn are separated by some dozen meters, making the usage of highway=tram_stop, which is mapped on the rail and usually used two times per halt, more fitting. Example: https://www.openstreetmap.org/relation/905074

kocio-pl:
Yes, it looks more like a tram stops, indeed.

[...]

PR #1875, opened 2015-09-26
Move rendering of railway=tram_stop labels later

kocio-pl:
Resolves https://github.com/gravitystorm/openstreetmap-carto/issues/1834.

Moving only the tram_stop labels (not touching dots) from z>=14 to z>=16 works for me.

**Wroclaw**
z14 (before/after):
wroclaw-tram-labels-14
wroclaw-tram-14

z15 (before/after):
wroclaw-tram-labels-15
wroclaw-tram-15

**Warsaw** 
It's even better - some (currently obscured) metro/railway stations and cemetery name start being visible.

z14 (before/after):
warsaw-tram-labels-14
warsaw-tram-14

z15 (before/after):
warsaw-tram-labels-15
warsaw-tram-15

### DIFF
--- a/stations.mss
+++ b/stations.mss
@@ -1,4 +1,5 @@
 @station-color: #7981b0;
+@station-text: #66f;
 
 .stations {
   [railway = 'subway_entrance'][zoom >= 18] {
@@ -21,7 +22,7 @@
       text-name: "[name]";
       text-face-name: @bold-fonts;
       text-size: 9;
-      text-fill: #66f;
+      text-fill: @station-text;
       text-dy: -8;
       text-halo-radius: 1;
       text-halo-fill: rgba(255,255,255,0.6);
@@ -36,7 +37,6 @@
   }
 
   [railway = 'halt'],
-  [railway = 'tram_stop'],
   [aerialway = 'station']::aerialway {
     [zoom >= 13] {
       marker-file: url('symbols/square.svg');
@@ -52,7 +52,7 @@
       text-name: "[name]";
       text-face-name: @book-fonts;
       text-size: 8;
-      text-fill: #66f;
+      text-fill: @station-text;
       text-dy: -8;
       text-halo-radius: 1;
       text-halo-fill: rgba(255,255,255,0.6);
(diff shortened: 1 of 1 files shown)


matkoniecz:
I think that testing it also for locations outside Poland is important.

Also, maybe metro labels should have higher priority anyway?

nebulon42:
As far as Vienna is concerned rendering tram stop labels later would be an improvement. z15 would work there too. z16 would be ok for a deliberate cartographic decision to shift the focus away from the transport network towards other things.

Generally, I think that due to the colour the non-bold letters of tram stops are not readable at lower zooms anyway (at least on my FHD monitor).

kocio-pl:
@matkoniecz OK, so which places would be good to test (Vienna is already on my list)? You probably know some of them because you were playing with trams around the world, if I remember correctly.

I agree the priorities should be fixed anyway, I just thought it may take more time for me to learn it (I also plan to change priorities for small amenities and had some problems understanding details) and it's independent from this PR, so I just planned to do it later.

matkoniecz:
No idea, I would just check random big cities across the world. Something in USA, West Europe, East Europe, Asia, maybe Australia.

It is quite surprising how different many things may be - for example I encountered railway=tram line leaving city and leading toward nearby villages/cities (I am unable to recall location).

nebulon42:
Maybe this section (or article in general) is of some help: https://en.wikipedia.org/wiki/Tram#Largest_tram_systems

matkoniecz:
> Also metro stations should take precedence over tram stops.

#1876 should solve this.

nebulon42:
Great, I update my statement for Vienna from above to say that z16 is a good zoom level for tram stop labels.

kocio-pl:
According to Wikipedia both Warsaw and Vienna are one of the biggest tram networks in the world, so it's good to have them tested anyway. 

**Vienna**

z14 (before/after):
vien-tram-labels-14
vien-tram-14

z15 (before/after):
vien-tram-labels-15
vien-tram-15

kocio-pl:
I think the testing set is complete now. According to Wikipedia today trams are popular mainly in Europe, so it doesn't make sense to try Asia for example. 

While those renderings are not 100% accurate (the bounding boxes are set manually and they are different in size than on the map), I see that when the tram network has no crossing lines (like Toronto), then the problem is only slightly visible at z14, but when the network is more complicated, even z15 is still too early too display the labels.

math1985:
Thanks! In The Hague, moving the stop names to z15 would be enough, but given for example the Melbourne example I can accept moving them to z16.

PR #1878, opened 2015-09-27
change appearance of station labels

nebulon42:
Since I find the colour of station labels quite ugly I thought it might be better to tie them to the colour of the station markers (but slightly darker). Additionally, the label is now rendered below the marker as most other labels. Railway halts and tram stops are now rendered in bold font, but I have instead decreased the font size by 1px.

As this makes tram stops slightly more prominent it makes only sense to merge it after #1875.

city environment (metro station/railway station/tram stop)
city_before
city_after

rural environment (railway station/railway halt)
rural_before
rural_after

aerialways are also affected
aerialway_before
aerialway_after

### DIFF
--- a/stations.mss
+++ b/stations.mss
@@ -21,8 +21,8 @@
       text-name: "[name]";
       text-face-name: @bold-fonts;
       text-size: 9;
-      text-fill: #66f;
-      text-dy: -8;
+      text-fill: darken(@station-color, 15%);
+      text-dy: 8;
       text-halo-radius: 1;
       text-halo-fill: rgba(255,255,255,0.6);
       text-wrap-width: 0;
@@ -31,7 +31,7 @@
     [zoom >= 15] {
       marker-width: 9;
       text-size: 11;
-      text-dy: -10;
+      text-dy: 10;
     }
   }
 
@@ -50,17 +50,17 @@
     }
     [zoom >= 14] {
       text-name: "[name]";
-      text-face-name: @book-fonts;
-      text-size: 8;
-      text-fill: #66f;
-      text-dy: -8;
+      text-face-name: @bold-fonts;
+      text-size: 7;
+      text-fill: darken(@station-color, 15%);
+      text-dy: 7;
       text-halo-radius: 1;
       text-halo-fill: rgba(255,255,255,0.6);
       text-wrap-width: 0;
       text-placement: interior;
       [zoom >= 15] {
-        text-size: 10;
-        text-dy: -10;
+        text-size: 9;
+        text-dy: 9;
       }
     }
   }


kocio-pl:
This change makes the labels too close to the neutral/generic black and thus are hard to spot. Current railway station labels are nice for me. 

I see a different problem however: railway/tram stations are not in transportation-blue and I'm not sure this is a good thing. I guess they should be probably colored like the rest of transportation objects.

matkoniecz:
This change makes station labels in cities less readable, what in my opinion is not improving situation (I am basing it at provided previews).

nebulon42:
Any opinions on the bold face for tram stops? The colour could be lighter instead of darker than the markers too, I will have to test that.

kocio-pl:
I can't imagine how it would look with a lighter color, so the best solution would be to test both versions.

nebulon42:
Lighter does not work for me. I might defer the colour change to later when (if ever) tram lines are lighter. But first I'd like to hear some more opinions, because for me readability is not really worse.

kocio-pl:
Example of unifying railway colors with transportation blue:

Moscow, z16
moscow-tram-transportation-blue-16

nebulon42:
Does not look bad on a small part of the map, but on a larger portion it makes the stops too prominent:

station

math1985:
> Since I find the colour of station labels quite ugly I thought it might be better to tie them to the colour of the station markers (but slightly darker).

I like the idea of tying the marker and name colour. I agree the proposed colour is slightly too dark - maybe we can find a colour in between the current icon colour and current name colour for both?

> Additionally, the label is now rendered below the marker as most other labels.

+1

> Railway halts and tram stops are now rendered in bold font, but I have instead decreased the font size by 1px.

+1

kocio-pl:
@nebulon42 My questions:
1. What zoom level is this? 
2. Are tram stops moved to z>=16 or are still on z>=14? 
3. I guess tram stops are bold here? Looking at my example I think tram stops are much more readable when in transportation blue and should not be bold, because they become too visible.

Another higher zoom (z16) test - Warsaw:
warsaw-tram-transportation-blue-16

mboeringa:
> I like the idea of tying the marker and name colour. I agree the proposed colour is slightly too dark - **maybe we can find a colour in between the current icon colour and current name colour for both?**

\+1

Alternatively, it might be an option to switch from a less saturated to a slightly higher saturated color when zooming in, as long as the change is kept subtle and understandable.

nebulon42:
@kocio-pl Since your PR has been merged (@math1985: thanks!) I will have to update mine anyway. Meanwhile we can discuss if changing the colour completely is an option. I agree that bold labels would be too prominent then. But I'm not convinced that changing the colour to bus stops is a good idea. Others should comment on that as well.

matkoniecz:
https://github.com/gravitystorm/openstreetmap-carto/pull/1878#issuecomment-143589133 proves that bus stop colour works poorly.

kocio-pl:
I'm also not sure that this unifying is better, but after pulling my code into I'd like to test it also on lower levels. 

On the other hand: what are the reasons to have railway/aerialway stops/stations rendered different than the rest of transport (bus, taxi, bicycle - plus hotel/hostel or embassy)? 

But it's even harder: aerodrome has different colors and I also don't know what to do with it and have to think about it and discuss it.

BTW: @matkoniecz PR (making tram stops priority lower) is also merged, thanks!

kocio-pl:
My rough conclusions: transportation-blue (or rather "blues": @transportation-icon and
@transportation-text) is probably not enough to unify all the transportation objects we have. Aerodrome and helipad use special @airtransport color for both the icon and a label, while rail/aerialway use @station-color and @station-text (new variable for already re-used value) and I think they are OK. In general the colors are lighter for smaller items (which are rendered later) to be more visible and less bright for those, which are visible earlier to not dominate the area.

However not all "stations" are the same - they are probably more or less important: I think bold is rather needed for railway=halt to make clear that they make a progression: 

station > halt > tram_stop

(currently aerialway station is rendered the same as railway=halt and I see nothing wrong with it)

Here (Warsaw, z16):
railway-halt-problem-16

you can see railway=station on the left, railway=halt in the middle, metro station in the top right corner and some tram stops around. I would make halt more visible than now (bold with a smaller text size than a station) and leave the tram stops as they are, so the user can see the difference and progression of general importance.

nebulon42:
You already have hijacked it, so I'm closing this.

Issue #1745, opened 2015-08-14
Push some small amenities to higher zoom level


kocio-pl:
I think such small amenities as amenity=post_box and amenity=atm are rendered too early now (z17, like shops for example). They are more important than trash cans (z19), so z18 seems like a natural choice for me.

This issue is related to https://github.com/gravitystorm/openstreetmap-carto/issues/847#issuecomment-127776632 - I have proposed rendering amenity=recycling+recycling_type=container and amenity=waste_disposal also from z18, as they are also only of local interest, but more important than trash cans.

HolgerJeromin:
Is getting the money less important than spending it? I am not sure.

kocio-pl:
I think about proportions: the counterpart for shops is rather bank or post office IMO. ATM or post box are less important than them.

Similar case: amenity=recycling+recycling_type=centre are bigger/more important > amenity=recycling+recycling_type=container and amenity=waste_disposal are in between > trash cans are the least important.

kocio-pl:
Mockup with atm, post_box, telephone [pushed from z17 to z18](https://github.com/kocio-pl/openstreetmap-carto/tree/small-amenities) (and emergency_phone to z19, as this is highly specialized kind of amenity):

z17:
![small-amenities-17](https://cloud.githubusercontent.com/assets/5439713/9305960/99c486fe-44f5-11e5-88b9-2364ee952f49.png)
z18:
![small-amenities-18](https://cloud.githubusercontent.com/assets/5439713/9305962/9c3372e2-44f5-11e5-9336-bd89b8d5d956.png)

I also think about moving them to the low priority group.

Another amenities that could be probably moved further down the zoom level and/or prioritized as less important could be toilettes and drinking_water, but I am not sure about them yet.

pnorman:
Given the density of post boxes, I don't think they should be moved from z17 to z18.

kocio-pl:
It's not about density of post boxes alone - they just add up to the whole density, as you can see, like other small amenities. They are clearly less important than post offices, so in my opinion it's perfectly legit to push them +1 zoom level.

In many places they are of less interest for general public (it's not a map mainly for tourists anyway) and there is [a special issue about objects in rural areas](https://github.com/gravitystorm/openstreetmap-carto/issues/1705), where every POI is more important than in the city/town/village, so post boxes should be more visible there (maybe even more than now - we'll see).

HolgerJeromin:
[...]

PR #1884, opened 2015-09-29
Pushing small amenities to higher zoom levels

kocio-pl:
Resolves https://github.com/gravitystorm/openstreetmap-carto/issues/1745.

Objects moved down to z>=18:
- amenity=atm
- amenity=post_box
- amenity=telephone
- amenity=toilets
- amenity=drinking_water

Objects moved down to z>=19:
- amenity=emergency_phone
- highway=traffic_signals

Examples:

Warsaw, z17 (21.0843,52.2430,21.0878,52.2451), before/after
small-amenities-universam-before-17
small-amenities-universam-17

Warsaw, z17 (21.0098,52.2484,21.0144,52.2512), before/after
small-amenities-starowka-before-17
small-amenities-starowka-17

However I am not sure if the code related to priorities works as expected - please correct me if I made any mistakes, I'll try to learn how to use it. The biggest problem is that it's hard to test them.

### DIFF
--- a/amenity-points.mss
+++ b/amenity-points.mss
@@ -32,7 +32,7 @@
     point-placement: interior;
   }
 
-  [feature = 'amenity_atm'][zoom >= 17] {
+  [feature = 'amenity_atm'][zoom >= 18] {
     marker-file: url('symbols/atm.16.svg');
     marker-fill: @amenity-brown;
     marker-placement: interior;
@@ -95,7 +95,7 @@
     marker-clip: false;
   }
 
-  [feature = 'highway_traffic_signals'][zoom >= 17] {
+  [feature = 'highway_traffic_signals'][zoom >= 19] {
     marker-file: url('symbols/traffic_light.svg');
     marker-fill: #0a0a0a;
     marker-placement: interior;
@@ -399,7 +399,7 @@
     marker-clip: false;
   }
 
-  [feature = 'amenity_post_box'][zoom >= 17] {
+  [feature = 'amenity_post_box'][zoom >= 18] {
     marker-file: url('symbols/post_box-12.svg');
     marker-fill: @amenity-brown;
     marker-placement: interior;
@@ -452,14 +452,14 @@
     marker-clip: false;
   }
 
-  [feature = 'amenity_telephone'][zoom >= 17] {
+  [feature = 'amenity_telephone'][zoom >= 18] {
     marker-file: url('symbols/telephone.16.svg');
     marker-fill: @amenity-brown;
     marker-placement: interior;
     marker-clip: false;
   }
 
-  [feature = 'amenity_emergency_phone'][zoom >= 17] {
+  [feature = 'amenity_emergency_phone'][zoom >= 19] {
     marker-file: url('symbols/emergency_phone.16.svg');
     marker-fill: @amenity-brown;
     marker-placement: interior;
@@ -473,14 +473,14 @@
     marker-clip: false;
   }
 
-  [feature = 'amenity_toilets'][zoom >= 17] {
+  [feature = 'amenity_toilets'][zoom >= 18] {
     marker-file: url('symbols/toilets.16.svg');
     marker-fill: @amenity-brown;
     marker-placement: interior;
     marker-clip: false;
   }
 
-  [feature = 'amenity_drinking_water'][zoom >= 17] {
+  [feature = 'amenity_drinking_water'][zoom >= 18] {
     marker-file: url('symbols/drinking_water.16.svg');
     marker-fill: @amenity-brown;
     marker-placement: interior;
@@ -1208,7 +1208,7 @@
   [feature = 'amenity_car_wash'][zoom >= 17],
   [feature = 'amenity_community_centre'][zoom >= 17],
   [feature = 'amenity_fire_station'][zoom >= 17],
-  [feature = 'amenity_drinking_water'][zoom >= 17],
+  [feature = 'amenity_drinking_water'][zoom >= 18],
   [feature = 'tourism_picnic_site'][zoom >= 17],
   [feature = 'leisure_picnic_table'][zoom >= 17],
   [feature = 'amenity_post_office'][zoom >= 17] {
@@ -1347,7 +1347,6 @@
       text-placement: interior;
     }
   }
-
 
   [feature = 'leisure_miniature_golf'][zoom >= 17],
   [feature = 'leisure_golf_course'][zoom >= 15] {
@@ -1757,7 +1756,7 @@
   }
 
   [feature = 'amenity_bank'][zoom >= 17],
-  [feature = 'amenity_atm'][zoom >= 17] {
+  [feature = 'amenity_atm'][zoom >= 18] {
     text-name: "[name]";
     [feature = 'amenity_atm'] {
       text-name: "[operator]";
--- a/project.mml
+++ b/project.mml
@@ -1997,7 +1997,7 @@
       "srs": "+proj=merc +a=6378137 +b=6378137 +lat_ts=0.0 +lon_0=0.0 +x_0=0.0 +y_0=0.0 +k=1.0 +units=m +nadgrids=@null +wktext +no_defs +over",
       "Datasource": {
         "extent": "-20037508,-20037508,20037508,20037508",
-        "table": "(SELECT\n    way,\n    amenity,\n    railway,\n    highway,\n    barrier,\n    man_made,\n    historic,\n    access,\n    name,\n    CASE WHEN amenity='waste_basket' THEN 2 ELSE 1 END AS prio\n  FROM planet_osm_point p\n  WHERE barrier IN ('bollard', 'gate', 'lift_gate', 'block')\n     OR highway IN ('mini_roundabout')\n     OR railway = 'level_crossing'\n     OR amenity IN ('parking', 'bicycle_parking', 'motorcycle_parking', 'bench', 'waste_basket')\n     OR historic = 'wayside_cross'\n     OR man_made = 'cross'\n  ORDER BY prio\n  ) AS amenity_low_priority",
+        "table": "(SELECT\n    way,\n    amenity,\n    railway,\n    highway,\n    barrier,\n    man_made,\n    historic,\n    access,\n    name,\n    CASE WHEN amenity='waste_basket' THEN 4\n         WHEN amenity IN ('bench', 'emergency_phone') THEN 3 \n         WHEN amenity IN ('atm', 'telephone', 'post_box') THEN 1 \n         ELSE 2 END\n    AS prio\n  FROM planet_osm_point p\n  WHERE barrier IN ('bollard', 'gate', 'lift_gate', 'block')\n     OR highway IN ('mini_roundabout')\n     OR railway = 'level_crossing'\n     OR amenity IN ('atm', 'parking', 'bicycle_parking', 'motorcycle_parking', 'bench', \n                    'waste_basket', 'telephone', 'emergency_phone', 'post_box')\n     OR historic = 'wayside_cross'\n     OR man_made = 'cross'\n  ORDER BY prio\n  ) AS amenity_low_priority",
         "geometry_field": "way",
         "type": "postgis",
         "key_field": "",
@@ -2023,7 +2023,7 @@
       "srs": "+proj=merc +a=6378137 +b=6378137 +lat_ts=0.0 +lon_0=0.0 +x_0=0.0 +y_0=0.0 +k=1.0 +units=m +nadgrids=@null +wktext +no_defs +over",
       "Datasource": {
         "extent": "-20037508,-20037508,20037508,20037508",
-        "table": "(SELECT\n    way,\n    amenity,\n    railway,\n    highway,\n    barrier,\n    man_made,\n    historic,\n    access,\n    name\n  FROM planet_osm_polygon\n  WHERE barrier IN ('bollard', 'gate', 'lift_gate', 'block')\n     OR highway IN ('mini_roundabout')\n     OR railway = 'level_crossing'\n         OR amenity IN ('parking', 'bicycle_parking', 'motorcycle_parking')\n  ) AS amenity_low_priority_poly",
+        "table": "(SELECT\n    way,\n    amenity,\n    railway,\n    highway,\n    barrier,\n    man_made,\n    historic,\n    access,\n    name\n  FROM planet_osm_polygon\n  WHERE barrier IN ('bollard', 'gate', 'lift_gate', 'block')\n     OR highway IN ('mini_roundabout')\n     OR railway = 'level_crossing'\n     OR amenity IN ('parking', 'bicycle_parking', 'motorcycle_parking')\n  ) AS amenity_low_priority_poly",
         "geometry_field": "way",
         "type": "postgis",
         "key_field": "",
--- a/project.yaml
+++ b/project.yaml
@@ -2456,12 +2456,17 @@ Layer:
             historic,
             access,
             name,
-            CASE WHEN amenity='waste_basket' THEN 2 ELSE 1 END AS prio
+            CASE WHEN amenity='waste_basket' THEN 4
+                 WHEN amenity IN ('bench', 'emergency_phone') THEN 3 
+                 WHEN amenity IN ('atm', 'telephone', 'post_box') THEN 1 
+                 ELSE 2 END
+            AS prio
           FROM planet_osm_point p
           WHERE barrier IN ('bollard', 'gate', 'lift_gate', 'block')
              OR highway IN ('mini_roundabout')
              OR railway = 'level_crossing'
-             OR amenity IN ('parking', 'bicycle_parking', 'motorcycle_parking', 'bench', 'waste_basket')
+             OR amenity IN ('atm', 'parking', 'bicycle_parking', 'motorcycle_parking', 'bench', 
+                            'waste_basket', 'telephone', 'emergency_phone', 'post_box')
              OR historic = 'wayside_cross'
              OR man_made = 'cross'
           ORDER BY prio
@@ -2491,7 +2496,7 @@ Layer:
           WHERE barrier IN ('bollard', 'gate', 'lift_gate', 'block')
              OR highway IN ('mini_roundabout')
              OR railway = 'level_crossing'
-                 OR amenity IN ('parking', 'bicycle_parking', 'motorcycle_parking')
+             OR amenity IN ('parking', 'bicycle_parking', 'motorcycle_parking')
           ) AS amenity_low_priority_poly
     properties:
       minzoom: 14


dieterdreist:
sent from a phone

> Am 29.09.2015 um 20:34 schrieb kocio-pl <notifications@github.com>:
> 
> Objects moved down to z>=18:
> 
> amenity=atm
> amenity=post_box
> amenity=telephone
> amenity=toilets
> amenity=drinking_water
> Objects moved down to z>=19:
> 
> amenity=emergency_phone
> highway=traffic_signals


drinking water (and maybe emergency phones) are also important outdoor POIs in rural areas, by hiding them before z18 they will basically almost never show up in this context while browsing the map

pnorman:
> amenity=toilets
> amenity=drinking_water

I'd rather not move these to z18. If we're getting overwhelmed with private toilets, we could look at handling them like parking.

> amenity=emergency_phone
> highway=traffic_signals

traffic signals work well on z17 here. There's not much overlap with other icons because they're in the middle of roads, where POIs tend to not be.

matkoniecz:
> private toilets

This is be definition an invalid tagging, amenity=toilet is defined as a public toilet.

math1985:
Very useful PR! I'd prefer to keep traffic_lights where they are, they are very important for navigation ('turn left at the second lights').

kocio-pl:
Sorry, I should have start with a rationale probably (it seems I did it in some other comments). My main purpose here is to establish a general system based on already existing scheme:

special shop (z16) > standard shop (z17) > vending machine (z18)
special amenity (z16) > standard amenity (z17) > small amenity (z18)
monument (z16) > statue (z17) > stone (z18) > plaque on the wall (z19)

...and so on. We're currently not showing vending machines nor we're able to recognize the type of memorial, but I hope we will in the future.

The system does not have to be strict, but I think it's a sane default to start with. We can still have exceptions, but at least we should know why we want them. So we can discuss drinking water, emergency phones, toilets and traffic_signals, of course. 

I think the main problem with them is that they have different priority for users in different environments, but we have currently no tools to know it for sure (interesting exception could be a bollard - it's very important on the road, but otherwise it belongs to z19 and I think it's easy to check). It's a general problem described in https://github.com/gravitystorm/openstreetmap-carto/issues/1705. 

Let's look closer at these examples:

- Drinking water is a secondary amenity in the city, as shown on the image above (see also [St. Peter's Square in Vatican](http://www.openstreetmap.org/relation/4525434#map=17/41.90183/12.45658)), but can be as important as shelter in the outdoor.

- Emergency phones on the metro station are the least important thing (we already have a big problem with multi-level objects clutter), while it may be the only POI in some places near motorway.

- Toilets should be visible when in some bigger area, especially for tourists or visitors (like theme park or zoo), but in general they are less important than for example bank or post office (they can even be a part of them).

- Traffic lights may be important for navigation outside the city on the long road, but inside they are in general visually competing with street names and direction arrows, which are more important, and the roads are much shorter.

What do you think we can do with those cases?

imagico:
I would probably do it more or less exactly the other way round - keeping the listed things at the current zoom levels but moving shops in general and possibly also restaurants etc. up one zoom level.  With the huge number of shop types added recently shops are very dominating at z17 strongly affecting general map readability and this would help here.  And shops are much more limited to dense urban areas while the features you intend to downgrade are mostly quite common in smaller places and rural areas where you often do not use z18/19 at all.

You can also look at it differently:  atm/toilets/drinking_water/emergency_phone are all addressing very basic needs and should therefore have priority over things addressing more specialized and sophisticated needs like shops.

And in any case at z18 and z19 all symbols shown at z17 already should have priority over those starting later - this does not seem to be the case right now.

kocio-pl:
I think about your proposition for a few days, because it's so different, but I'm still not convinced:

> You can also look at it differently: atm/toilets/drinking_water/emergency_phone are all addressing very basic needs and should therefore have priority over things addressing more specialized and sophisticated needs like shops.

Looking at your examples, toilets and drinking water may be seen as essential (well, mainly for visitors/tourists; rather not for anyone else), but any type of food+drink shops are very similar in this regard, while using ATM is definitely not a "basic need". And I don't even know how should we measure emergency phone: most of the time they're not even used, but sometimes they can be top priority. However this style is not "emergency" map, rather universal one, so I consider them less interesting. Otherwise we should probably show fire hydrants in a distinctive way and I don't think it would be good solution here.

I think my set of rules is rather compact and easy to apply to the new objects, while your approach is more subjective and specific instead of general solution I'm trying to achieve. When we try to act case by case, we fall into the "this-and-that" trap: we're not able to keep consistency in the long run (any new element can turn it upside down) and I'd like to avoid it.

imagico:
With *addressing very basic needs* i mean these features cater needs that are either essential for physical well-being (drinking water, emergency phone) or for socially acceptable behavior (toilets, ATM).  This does not apply for the usual food&drink amenities which have primarily a social function (a non-obligatory one in most societies) and not basic sustenance.

Also note ATM is a prerequisite for use of shops and other things, not the other way round - hence it is more basic.

You can of course argue that emergency phone - like post box - is no more important due to the ubiquity of mobile communication.  I would be careful with that but it is without doubt a possible argument in many parts of the world.

> Otherwise we should probably show fire hydrants in a distinctive way

That's a joke, right? - an emergency phone is useful for everyone while a fire hydrant is not of any use for the vast majority of people (except maybe to park their car in front...).

> I think my set of rules is rather compact and easy to apply

Did i understand you correctly that this is essentially *physical size equals importance*?

While this certainly is a consistent importance rating scheme and probably works quite well within a single class of objects (like large restaurants being more important than small ones) there are likely many people who disagree this is the best approach in general.  The criterion should not be simplicity and ease of use from the developer perspective but usefulness for the map user.

dieterdreist:
sent from a phone

> Am 05.10.2015 um 18:58 schrieb kocio-pl <notifications@github.com>:
> 
> Looking at your examples, toilets and drinking water may be seen as essential (well, mainly for visitors/tourists; rather not for anyone else)


I don't agree, these are relevant to all people, not just visitors. It may depend on the climate (and availability) whether drinking water is essential or just nice to have, toilets tend to be more important for women than men


> , but any type of food+drink shops are very similar in this regard, while using ATM is definitely not a "basic need".


I would see ATMs as basic need, unless you're in a region where you barter or pay everything above 2 cents with a credit card. This might change in the future (NFC payments, all other kinds of Proprietary Pay etc), but at least around here that's simply not the case yet.


> And I don't even know how should we measure emergency phone: most of the time they're not even used, but sometimes they can be top priority. However this style is not "emergency" map, rather universal one, so I consider them less interesting.


agreed


> Otherwise we should probably show fire hydrants in a distinctive way and I don't think it would be good solution here.


not at all, hydrants are completely different from emergency phones, as they are intended for the fire brigade to be used while the phones are intended for you.

kocio-pl:
@imagico 

> With addressing very basic needs i mean these features cater needs that are either essential for physical well-being (drinking water, emergency phone) or for socially acceptable behavior (toilets, ATM). 

Using ATM has nothing special to do with socially acceptable behaviour - buying goods falls into the same category. This category is just too broad to be useful - in general we don't show any socially unacceptable behaviour objects.

> This does not apply for the usual food&drink amenities which have primarily a social function (a non-obligatory one in most societies) and not basic sustenance.

While typical restaurant or bar may be seen as social, typical fast food is just totally different. What about alcohol shop? We loose any system here. And what about supermarkets? Now we also loose the reason behind current rendering.

> Also note ATM is a prerequisite for use of shops and other things, not the other way round - hence it is more basic.

It may be in some cases, but it is not universal: you can get your salary from a postman/post office or just pay with a card. What about banks? Do you propose to show ATM first and bank later? For me it doesn't make sense. Again, we loose the system here.

One can also say that earning money is a prerequisite for using ATMs, so maybe industrial objects, retail objects and offices are more important... It's hard to say where should we start at all.

> You can of course argue that emergency phone - like post box - is no more important due to the ubiquity of mobile communication. I would be careful with that but it is without doubt a possible argument in many parts of the world.

No, I said it's less important because it's small amenity and at the same time it is not an emergency map, so it should be less visible for general purposes. If we talk about safety features, health category is much more important for most of the people and relevant enough.

> That's a joke, right? - an emergency phone is useful for everyone while a fire hydrant is not of any use for the vast majority of people (except maybe to park their car in front...).

You're right - this example is different. But what about fire extinguisher then? Anybody can use it, it's small and very useful in case of (fire) emergency. It's very similar to emergency phone.

> Did i understand you correctly that this is essentially physical size equals importance?

No, it's just an important base feature - I mean: for a general map. Importance is a very useful modifying feature.

> While this certainly is a consistent importance rating scheme and probably works quite well within a single class of objects (like large restaurants being more important than small ones) there are likely many people who disagree this is the best approach in general. The criterion should not be simplicity and ease of use from the developer perspective but usefulness for the map user.

Size is more objective criterion and it's easier to use even for totally new kind of objects and even more people will disagree with what is more or less important. "Case by case" works if you have very small set of features - but we have much more of them and we start to deal with a big mix of things.

So, let's suppose we try to stick with importance as a base, not size. That would make UNESCO heritage objects (rare example when we have some objectivity in recognizing importance) shown at the world/country level, for example. That is great solution when trying to make a tourist map, but on general map I prefer to see a big military area earlier, even if it's not important (or even usable) by general public.

imagico:
I think you are getting side tracked here in multiple directions (and it does not really lead anywhere when you say things i did not say do not make sense to you).  I'd suggest to stick to the subject at hand, that is if the changes in priorities you suggest, that is essentially to consider a number of amenities less important than all shops and food&drink amenities,  are desirable.

I stick to my opinion that the changes you propose are in a very different direction to the priorities i consider useful.  I offered some arguments with reasons for these priorities, i.e. why i consider ATMs, toilets and drinking water more important than jewelry and perfumery shops :wink:.  And i do not consider the typical physical size (however you intend to measure it) to be a useful criterion for cross category rating.  I do not even think typical size is even less subjective than other measures of importance - after all the typical size of a shop/restaurant/whatever varies strongly with the cultural environment.

Maybe also see it from a different perspective: which of the following statements do you consider a more likely complaint from a map user:

* I can't find a restaurant or clothing shop on the map because the map is filled with all the public toilets and drinking water points
* I can't find a public toilet or drinking water point because the map is filled with all the restaurants and clothing shops.

kocio-pl:
> I think you are getting side tracked here in multiple directions 

I guess what you call "being side tracked" is that we still have no place for general discussions, so we have to put everything into small, detailed boxes like this one. Looking for a general rules is a clear common thread in few of my propositions (like stations precedence or system for using colors of POIs, for example) and also the context which is important to understand the purpose of this piece of code. For me it is just the opposite of going into multiple directions.

> (and it does not really lead anywhere when you say things i did not say do not make sense to you). 

Sorry, I was trying to guess what do you mean and apparently didn't get your way of thinking.

> I'd suggest to stick to the subject at hand, that is if the changes in priorities you suggest, that is essentially to consider a number of amenities less important than all shops and food&drink amenities, are desirable.

If we don't think about why having general rules of thumb is desirable, it's harder to understand why I propose to push them in the first place. I have nothing against some exceptions, but first I'd like to have some scale-related rules, not only disconnected individual "votes" on everything.

Are you just opposing those specific types of objects (as exceptions) or you don't like this rule at all?

> I stick to my opinion that the changes you propose are in a very different direction to the priorities i consider useful.

I also think that lack of rules is more important than trying to manually fine tune every type of object, especially when we have a lot of them.

> Maybe also see it from a different perspective: which of the following statements do you consider a more likely complaint from a map user:
>
>    I can't find a restaurant or clothing shop on the map because the map is filled with all the public toilets and drinking water points
>    I can't find a public toilet or drinking water point because the map is filled with all the restaurants and clothing shops.

Interesting question, probably worth asking at the Talk list.

As for me the first is a no brainer: in dense places I wouldn't look for such amenities on the map, I'd rather ask the people or just look around (public toilets in restaurants or ATMs in banks are common features, at least here in Poland), restaurants and shops are much more interesting (especially restaurants are common meeting points). In the outdoor [I think about general "amplifying" some objects](https://github.com/gravitystorm/openstreetmap-carto/issues/1705) (including water points or toilets), but there are no restaurants nor clothes shops to eclipse them.

imagico:
> Are you just opposing those specific types of objects (as exceptions) or you don't like this rule at all?

I think using a feature type based estimate of typical sizes as an importance rating measure for POIs is not a good idea because it is (a) subjective, arbitrary and prone to cultural bias and (b) not well suited for this purpose.  But even if you disagree it should be clear that this PR is not a good approach to executing this principle anyway since for example the typical size of a public toilet is likely not smaller than the typical size of a small shop like a kiosk, the size ratio within shops OTOH - from a kiosk to a mall/garden centre etc. - is probably larger than 1:100.  Hence this would be a much more significant place to implement a size prioritization principle than the amenities here.

kocio-pl:
Anything related to "importance" is even more subjective and definitely more influenced by the culture and multiple circumstances. Typical size is not completely free from this problem too, but it's simply not that severe. I don't believe in finding perfect solution, I just prefer something less subjective, more flexible and universal. I guess it's the closest thing to the neutral measure we can have for a general style.

> But even if you disagree it should be clear that this PR is not a good approach to executing this principle anyway since for example the typical size of a public toilet is likely not smaller than the typical size of a small shop like a kiosk, 

That would be acceptable explanation for me to leave the toilets on z17, however they can be also a part of fuel station for example and could eclipse it. What do you think about this problem?

> the size ratio within shops OTOH - from a kiosk to a mall/garden centre etc. - is probably larger than 1:100. Hence this would be a much more significant place to implement a size prioritization principle than the amenities here.

And notice, that we [already show malls earlier](https://github.com/gravitystorm/openstreetmap-carto/blob/master/amenity-points.mss#L1480)! If we find other examples of typically big shops, we should also make them appear earlier. I just started with things that I see most of the time that should be seen later, but the rule is still the same ("mind the scale").

imagico:
Note i use *importance rating* as a technical term here.  This has nothing to do with an individual map users subjective considerations what is important or not for him/her.  Importance rating occurs at many places in a map style - it is kind of absurd to say it should not be used.

> And notice, that we already show malls earlier!

No, that is label placement for area features larger than the typical label size.  For shop icons only supermarket and department_store are currently displayed early which is a fairly arbitrary choice.  Also they are not rendered with priority over the other icons so they frequently vanish as you zoom in making the whole thing fairly counterproductive (see my first comment here).

kocio-pl:
> Note i use importance rating as a technical term here. 

I am not familiar with that term nor do I understand what does it mean in a technical sense. Could you describe it a bit more or show me where I could learn about it?

> No, that is label placement for area features larger than the typical label size. For shop icons only supermarket and department_store are currently displayed early which is a fairly arbitrary choice. 

Up to now everything is more or less arbitrary and I'd like to change it to be more systematic, so I would be happy to work on some rules. However both supermarket and department store are big by definition (first on is bigger than convenience, second one includes many selling points), so I see nothing wrong with it.

> Also they are not rendered with priority over the other icons so they frequently vanish as you zoom in making the whole thing fairly counterproductive (see my first comment here).

Good point, probably worth opening separate issue!

math1985:
@kocio-pl I'd be happy to merge this, except for the traffic signals change which might need some further consideration (and a separate PR).

In any case, if we decide to change the traffic signals, the line https://github.com/gravitystorm/openstreetmap-carto/blob/edd434961dab3c1848cfada7dae4a871d17ba2c2/roads.mss#L2409 needs changing as well (a variable might make sense to avoid them getting separated again in the future).

pnorman:
I'm against merging this.

math1985:
@pnorman Could you provide a bit more motivation?

You mentioned before you'd like to reduce the number of map features, would you be in favour of dropping any of the features in this PR instead?

pnorman:
> @pnorman Could you provide a bit more motivation?

https://github.com/gravitystorm/openstreetmap-carto/pull/1884#issuecomment-144213134, and I generally agree with @imagico's comments above.

> You mentioned before you'd like to reduce the number of map features, would you be in favour of dropping any of the features in this PR instead?

I could see moving some to the low priority layer - where these features are important tends to be areas where they wouldn't collide with other features. Others like `amenity=drinking_water` I see no need to change.

kocio-pl:
@math1985 I was about to sum the things up to see where we can agree, to limit the scope of this PR, but after this comment of @pnorman I don't really understand if he's against moving anything or just those 3 (4?) items he mentioned before. But if you still support the idea of skipping traffic lights in this PR, I'm ready to make just this change.

math1985:
> But if you still support the idea of skipping traffic lights in this PR

I'd prefer if we could find a form of consensus before merging this.

@pnorman Another question, is your objection mainly based on the rendering in residential/suburban areas, or is this a problem for dense urban areas as well? For example, do you think the proposed change works well in the example renderings in the first post, or do you think even the example renderings are problematic?

My motivation: completely mapped areas, even suburban ones, are currently rather hard to read at z17. [Example](http://www.openstreetmap.org/#map=17/49.60712/6.11061) The main cause seems to be the great variety of icons. The mailboxes there really seem to be not that significant.

imagico:
> My motivation: completely mapped areas, even suburban ones, are currently rather hard to read at z17. Example The main cause seems to be the great variety of icons.

I already made the suggestion to move the shops and food&drink amenities up one zoom level.

pnorman:
> @pnorman Another question, is your objection mainly based on the rendering in residential/suburban areas, or is this a problem for dense urban areas as well? For example, do you think the proposed change works well in the example renderings in the first post, or do you think even the example renderings are problematic?

I see icon overload as a problem, but not one that this PR really addresses. If anything, I'd rather have some of the stuff under discussion (e.g. drinking water) moved down a zoom level to render sooner.

dieterdreist:
sent from a phone

> Am 20.10.2015 um 05:58 schrieb Paul Norman <notifications@github.com>:
> 
> I see icon overload as a problem, but not one that this PR really addresses. If anything, I'd rather have some of the stuff under discussion (e.g. drinking water) moved down a zoom level to render sooner.


while that would be useful for many parts of the world (especially outdoors) it will surely create problems in areas with lots of them (like around here). Too many collisions I suspect (years ago, when I last checked, there already were 500 in the city area).

kocio-pl:
That's why I mentioned #1705 ("Showing some POIs earlier when outside the landuse=residential") - because for me those issues are correlated.

I think of size as the most universal property that could be used as a general rule and a context (such a "residential/city area" or "outdoor" for example) as a very important modifier - we can even call it a "secondary rule", which can change "primary rule" as much as we want it, not just "a bit". 

Being secondary doesn't mean it's limited in any way (like for example "only +/- 1 zoom level change is allowed"), because context/environment can be very different. It only means that we think this is a special case and so we want to have more control.

But of course I would like to have some rules also within those special environments, because there can be a lot of objects which should be shown different and crafting everything manually can lead to inconsistencies and strange effects.
 
I see it like this (totally bogus pseudocode):

```ruby
default rule
  mind the size
if environment == outdoor
  use some general_outdoor rules 
  treat special cases in a special way
if environment == city
  use some general_city rules 
  treat special cases in a special way
if environment == water
  use some general_water rules
  treat special cases in a special way 
```

kocio-pl:
Just a short summary (that I wanted to do anyway):

- against moving toilets - 1
- against moving drinking_water - 2
- against moving emergency_phone - 1 
- against moving traffic_signals - 2

The rest (atm, post_box and telephone) was not mentioned directly, however @imagico propositions is completely different and is hard to count. Maybe a competing issue/PR should be created instead to discuss its merits in itself?

math1985:
Unfortunately there does not seem to be support for the current PR so I'm going to close it. @kocio-pl Thanks for the effort! If @kocio-pl or anybody else has a different idea, feel free to propose it!

Issue #1824, opened 2015-09-10
add line rendering for tourism=attraction


HolgerJeromin:
http://www.openstreetmap.org/way/256157138
had a slim area around with tourism=attraction to get a nice pink label. But the fence is the attraction, so i moved the POI data to the fence. 
It would be nice to have again a label (with offset like a footway?) to this attraction.

polarbearing:
A label should be no problem, the question remains how a linear feature should be treated once tourism=attraction is rendered as an area #1257, which might be an outline as discussed in #1624.

HolgerJeromin:
I would propose to have no own line for a attraction, as i cannot think about a usecase where a linear attraction has no main key (like highway (?), fence, wall,...).

jojo4u:
I detailed some problems with tourism=attraction here: http://wiki.openstreetmap.org/wiki/Talk:Tag:tourism%3Dattraction

matkoniecz:
Possible display method: https://cloud.githubusercontent.com/assets/899988/10133846/0d5e8180-65e1-11e5-98e9-90def5b5acaa.png

> I would propose to have no own line for a attraction, as i cannot think about a usecase where a linear attraction has no main key (like highway (?), fence, wall,...).

Rendering names for barriers would be problematic. People love to tag linear and area objects on the same OSM element leading to things like [barrier=wall; landuse=cemetery; name=*].

kocio-pl:
We have somewhat similar problem with buildings - people add different amenities and their name is rendered like it is a building name, so we have a bit of confusion between two objects on the same OSM way.

matkoniecz:
See also  #971 for problems with rendering unable to handle putting linear and area object in one OSM element.

dieterdreist:
sent from a phone

> Am 28.09.2015 um 13:27 schrieb Mateusz Konieczny <notifications@github.com>:
> 
> Rendering names for barriers would be problematic. People love to tag linear and area objects on the same OSM element leading to things like [barrier=wall; landuse=cemetery; name=*].


I believe this common practice will sooner or later stop, because it really brings a lot of ambiguity into mapping when several objects (like a cemetery and a fence) are represented by the same object in osm. The rendering shouldn't encourage this style of mapping by working around the problems and guessing what might have been meant.

[...]

Issue #818, opened 2014-08-02
Remove barrier catch-all


math1985:
We have at the moment a catch-all rendering for barriers: all barrier=* lines, with the exception of barrier=hedge, are currently rendered as a gray line.

We probably want to get rid of this catch-all, just like we got rid of the other catch-all's.

Question is what we want to include. In particular, I wonder whether should render barriers that are typically tagged as node also as lines. For example, barrier=block, barrier=style, barrier=bollard, and barrier=gate are usually tagged as node, but sometimes also as way.

HolgerJeromin:
closed by #1986

math1985:
Thanks.

PR #1885, opened 2015-09-29
render labels for named barriers

matkoniecz:
Done to handle case reported in #1824 

tourism_way master-_barrier_way 18 18 350px master - barrier_way 350px

https://cloud.githubusercontent.com/assets/899988/10177608/45f21e9c-66fb-11e5-9d85-82d369039b18.png (the same place, more zlevels)

https://cloud.githubusercontent.com/assets/899988/10177637/674fa0aa-66fb-11e5-87b7-4face372bd89.png (named barrier in mountains)

note that in case of using one element to tag both named area objects that is not rendered and  linear barrier it will result in an unwanted name for barrier - see for example https://cloud.githubusercontent.com/assets/899988/10177630/5f4ddc6e-66fb-11e5-8543-a593a47a59d4.png (caused by http://www.openstreetmap.org/way/115536902#map=19/41.89471/12.50457)

In more popular cases (one OSM element with name to represent both rendered area object and linear object name is not rendered on barrier) - see https://cloud.githubusercontent.com/assets/899988/10142490/49c48cfa-6613-11e5-9069-7f7d2dad99ca.png


math1985:
This introduces another catch-all rendering (for barrier names), I think we wanted to avoid these. I guess it'd be better to explicitly define a list of barrier values?

pnorman:
Yes, this is labeling objects which may not otherwise be rendered - e.g. barrier=foo name=bar would get a name, but nothing else

matkoniecz:
At this moment I think that rendering names for all rendered barriers is OK.

We have anyway catch-all rendering of all barriers, reported as #818.

After fixing of #818 also names should be displayed only for valid values of barrier key
barrier foo name eeeeee eeeeee ref 1 ele 8000 operator ee ee ee operator brand ee ee ee brand way master - barrier_name 200px
.

math1985:
Ok, good point.

math1985:
I would significantly increase, or drop, the text-wrap. For names rendered along lines, a short text-wrap does not look very good.

math1985:
Thanks @matkoniecz! I'm going to accept  #1986, would you be able to make this PR match?

PR #1900, opened 2015-10-06
Removing deprecated, rarely used shops

kocio-pl:
Removing rendering of shop values deprecated according to Wiki pages:
- antique (not deprecated, it just lacks wiki page, but is less used version of antiques)
- betting
- delicatessen
- dive
- fish
- gambling
- insurance
- ~~organic~~
- pharmacy


matkoniecz:
> fish

I remember this one from https://lists.openstreetmap.org/pipermail/tagging/2014-July/018573.html (thread that resulted in https://josm.openstreetmap.de/ticket/10594 and noticeable decrease in how often some shop values are now appearing)

Somebody claimed that shop=fish is used for shops selling solely freshwater fish (and where seafood is not fitting).

HolgerJeromin:
As I understood the right shop is seafood or fishmonger with no clear solution. Especially with fresh water fish and native speakers.

Fish is no real alternative. So thanks for removing.

math1985:
+1

pnorman:
:-1: on `shop=organic`. I believe it's ~1k usage puts it well over some other shop values that are rendered with the generic icon.

kocio-pl:
Have you checked [shop=organic page](http://wiki.openstreetmap.org/wiki/Tag%3Ashop%3Dorganic)?:

> Deprecated - Use organic=* in combination with a general shop=* key instead

pnorman:
> Have you checked shop=organic page?:
> 
>>  Deprecated - Use organic=* in combination with a general shop=* key instead

Yes, I checked the wiki page.

kocio-pl:
So why do you want to render clearly deprecated feature?

math1985:
It seems the tag was 'deprecated' by one user without discussion: http://wiki.openstreetmap.org/w/index.php?title=Tag:shop%3Dorganic&diff=735472&oldid=689739

I think the poor state of osm's wiki documentation and the lack of a clear process to decide on changes to the documentation, together with the lack of importance many mappers give to the documentation, is something that seriously holds back the project, but unfortunately it seems there's little we can do about that.

kocio-pl:
OK, this is a legitimate reason to left it unchanged. I just wanted to get rid of things which are clearly not wanted here, there is however [a lot of other uncertain values](http://wiki.openstreetmap.org/wiki/User:Kocio/shop_review) and I want to discuss them on Tagging before taking any action.

BTW: I've just discovered what the `shop_values.rb` script is for. :smile: However in current form it's not usable for managing our shop list. It would be good if it could compare what we already use as shop values in `project.yaml` and show the difference.

matkoniecz:
> However in current form it's not usable for managing our shop list.

Are you worrying about excluding shops rendered with special icons?

jojo4u:
The state of documentation is sometimes sad, that's true. But wiki affects mapping, which affects Taginfo. You could change the title from: "Removing deprecated shops" to "Removing deprecated, rarely used shops".

kocio-pl:
@jojo4u I would remove every deprecated item, no matter how popular, it's just easier with less popular shops to remove them at hand and currently no popular shop type is deprecated. I would however try to make the transition smooth (deprecated usually means "this is the wrong schema, change it to something more appropriate"), so this is like those uncertain values I was talking about: more time is needed to resolve these problems, so they don't belong to this PR.

@matkoniecz My first problem is that just replacing the list would break diff output for project.yaml and I like to control the change, no matter how trivial it seems (https://github.com/gravitystorm/openstreetmap-carto/issues/1800 proves me it can be the case). But, frankly saying, I also don't know if the shop values can be in both lists ("dotted" and "iconized" shops), but I guess they should be moved from dotted list to iconized list, but it means more spotting and manual (tedious and error prone) work. Am I right?

pnorman:
> So why do you want to render clearly deprecated feature?

For better or worse, there is no process to depreciate a tag within OSM. As a data consumer I do not feel compelled to stop rendering something based on what one particular wiki page states.

jojo4u:
> For better or worse, there is no process to depreciate a tag within OSM. 

There is. But it is often not successful because OSM is no democracy ;) A new, accepted proposal can deprecate old tags. But sometimes the switch takes time (building=entrance) and sometimes the new tag is rejected by many (water=*).

kocio-pl:
Citing [wiki documentation](http://wiki.openstreetmap.org/wiki/Deprecated_features):

> OpenStreetMap does not have "banned features", as anybody is allowed and encouraged to use any tag they think is useful.

> However, from time to time the wiki community suggests a new feature instead of an older one, to improve tagging clarity or structure. If there is no strong reason to stick to the old tags in these cases, you might want to go along with the recommendation and use the new tags suggested here. 

In my opinion the fact that there is no _sure_  or _hard_ way to do anything in OSM (not only deprecating), doesn't mean we can't add/change common interpretation of tags - for example to unify something or avoid mistakes. This style is documented as part of a mapper feedback loop, so it is better to stick to this common interpretation.

kocio-pl:
> there is however a lot of other uncertain values and I want to discuss them on Tagging before taking any action.

The discussion is started [here](https://lists.openstreetmap.org/pipermail/tagging/2015-October/026828.html) and it looks like people are really interested in the subject.

PR #1901, opened 2015-10-08
Remove car_sharing rendering

### LINKED ISSUE #1891: Drop car_sharing rendering
@math1985 https://github.com/gravitystorm/openstreetmap-carto/issues/1764#issuecomment-145229715
> I wouldn't even mind dropping car_sharing from the default rendering. They are typically [insignificant features](http://2.bp.blogspot.com/-vkw2mqQ0yT0/U2_9qPlB0hI/AAAAAAABBXM/OVbRMkeyU_s/s1600/20140425045+NE+Amsterdam+KNSM+Eiland+NS+Green+Wheels+car+sharing+2+KSS.jpg), and nearly everyone having a car_sharing contract would use the app of their provider rather than OSM to find a car_sharing spot (because such apps also indicate whether cars are currently available).

I'm in agreement.

pnorman:
Fixes #1891


HolgerJeromin:
Just a thought: Should the icon stay in the repository in another folder? Would be good for reuse of the style. The icon matches all other icons.

pnorman:
> Should the icon stay in the repository in another folder?

It'll be in the repository, but not in the current tree. We don't store unused images.

nebulon42:
Also affects #1165.

RainerBielefeld:
<😏> So we also should drop all groceries from OSM. Customers have Apps for therir favorite ... </😏>
IMHO: Wrong decision!

Issue #285, opened 2013-12-12
unneded, unwanted and useless shields on highway=residential


matkoniecz:
example: http://www.openstreetmap.org/?mlat=50.07147&mlon=19.92589#map=18/50.07147/19.92589

Probably there are some countries where displaying shields even on highway=residential makes sense, so it may be impossible to keep everybody happy.

But maybe it is possible to stop displaying shield an highway=residential roads with name field?

PR #1905, opened 2015-10-10
make residential refs less prominent

nebulon42:
Follow-up to #1871.

Makes residential/unclassified refs stand out less, by removing the bold font, keeping the font size 1px below shielded refs (means decreasing by 2px on z15), increasing the spacing between texts so that less refs appear on z15 and increasing the halo width as replacement for a proper shield.

Previews are for a location that was mentioned in https://github.com/gravitystorm/openstreetmap-carto/issues/285#issuecomment-30535756 for having exceptionally many such refs.

before
residential_ref_z15_before

after - current road shields
residential_ref_z15_old_shields

after - proposed road shields
residential_ref_new_shields_z15

### DIFF
--- a/roads.mss
+++ b/roads.mss
@@ -2680,13 +2680,21 @@ tertiary is rendered from z10 and is not included in osm_planet_roads. */
   [highway = 'residential'] {
     [zoom >= 15] {
       text-name: "[refs]";
-      text-size: 10;
+      text-size: 8;
+
+      [zoom >= 16] {
+        text-size: 9;
+      }
+      [zoom >= 18] {
+        text-size: 10;
+      }
+
       text-fill: #000;
-      text-face-name: @bold-fonts;
-      text-min-distance: 18;
-      text-halo-radius: 1;
+      text-face-name: @book-fonts;
+      text-min-distance: 40;
+      text-halo-radius: 2;
       text-halo-fill: rgba(255,255,255,0.6);
-      text-spacing: 750;
+      text-spacing: 760;
       text-clip: false;
     }
   }


mboeringa:
Nice work, however, I also think Z15 may be to early to start rendering this, based on the example here. Maybe move this down to Z17 or so? There is not much use for a "random" display of a few of these refs at Z15, while the original rendering shows there are much more.

nebulon42:
I'm unsure about this. There could be regions with sparse features where residential/unclassified (more likely for unclassified) refs could be important. Although I don't have any locations moving the refs to z17 could be too much. Maybe also the min-distance is currently too high and could be lowered.

I'm not basically against rendering these refs later, but I'm unsure if the consequences are always desirable.

mboeringa:
Personally, the fact that they may block the labelling of road names at Z15, as can be seen in the examples, would be reason enough to push them down. What was the last time you used a "residential ref" to find your way in residential build up area, versus using road names?...

Anyway, let's hear other opinions...

nebulon42:
Right, just one small remark more to emphasise that my concern is more about unclassified than about residential.

aceman444:
Can you detect if a name is set and if yes, do not render the shield? Or is it that the shield of one road obscures the name of neighbouring road?

aceman444:
I mean shield=ref as it seems residentials are not rendered with the shield.

polarbearing:
> What was the last time you used a "residential ref" to find your way in a residential build-up area, versus using road names

there are areas where the ref is signposted and the name is just local knowledge.

> shield=ref

There is no such tag.

mboeringa:
> there are areas where the ref is signposted and the name is just local knowledge.

Yes, probably, as with a global map, there are always the "exceptions". However, in this particular case, if the name is not present in OSM, but the ref is, it might be better to use the ref as road label and wrap it along the way like the road name, instead of rendering it like a shield... Of course, that only makes sense for residential roads, not for the other road types.

Anyway, shouldn't the "ref" be in the name tag in case it is the only signposted value, based on the principal of mapping "on the ground" truth?

math1985:
In the few cases I found in Luxembourg, increasing text-min-distance does not work well, as none of the refs got rendered. Would this be a problem for other places as well?

nebulon42:
@math1985 Could you please share the locations?

math1985:
http://www.openstreetmap.org/#map=16/49.5952/6.1391 is the only one I can find now, the other one has been retagged since my database-import.

polarbearing:
> Anyway, shouldn't the "ref" be in the "name" tag in case it is the only signposted value, based on the principal of mapping "on the ground" truth? How would you actually be able to tell it was a "ref" instead of "name", if it is the only value signposted?

No, the ground truth is that a ref is a ref, and a name is a name. Tagging refs as names is wrong tagging. You are able to tell them e.g. if the refs follow a national road number reference scheme, and are signposted in a particular style that is quite different from name postings.

math1985:
Thanks - if the disappearance of residential ref's is a problem somewhere else as well, I'll suppose we'll hear about it.

nebulon42:
Thanks. I was planning to adjust min-distance, but did not have time so far.

PR #1906, opened 2015-10-11
Significantly increase minimum zoom level for bridge names

math1985:
Also make sure that the progression for font size minimum sizes increases with
a factor of 4 (this makes sure that every zoom level is one size bigger).

Bridge names currently show up way too early, see for example [here](http://www.openstreetmap.org/#map=14/52.0899/5.1238) with Viebrug and Lucasbrug.

### DIFF
--- a/roads.mss
+++ b/roads.mss
@@ -2420,7 +2420,7 @@ tertiary is rendered from z10 and is not included in osm_planet_roads. */
 
 #bridge-text  {
   [man_made = 'bridge'] {
-    [zoom >= 12][way_pixels > 4] {
+    [zoom >= 12][way_pixels > 62.5] {
       text-name: "[name]";
       text-size: 8;
       text-fill: black;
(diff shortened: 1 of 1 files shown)


HolgerJeromin:
Is there a reason you choose 62.5 and not a value like 60 or 65?

dieterdreist:
sent from a phone

> Am 11.10.2015 um 15:10 schrieb math1985 <notifications@github.com>:
> 
> Bridge names currently show up way too early, see for example here with Viebrug and Lucasbrug.


this will always have to be a compromise, in your example these are relatively small / insignificant waterways in a setting with lots of bridges. In different contexts (especially when the waterways are large and bridges therefore more expensive to build and less frequent) you will definitely want to see bridge names quite early.

math1985:
> Is there a reason you choose 62.5 and not a value like 60 or 65?

Yes - the font size increases at 250 pixels, so by using a cut-off of 250/4=62.5, every time you zoom in one level, the font size increases. Note that with every increase in zoomlevel, an object gets exactly twice as wide and twice as tall, so the total amount of pixels of an object will be multiplied by 4 with every zoomlevel increase.

HolgerJeromin:
thanks for the explanation.

math1985:
Thanks!

PR #1907, opened 2015-10-11
Move taxi to z17

math1985:
Move minimum zoomlevel for taxi stands from z16 to z17.


math1985:
@matkoniecz Do you have any chance to look at this?

PR #1881, opened 2015-09-28
Improving railway=halt and aerialway=station labels

kocio-pl:
This is a follow up of my research which started in https://github.com/gravitystorm/openstreetmap-carto/pull/1878#issuecomment-143726126, but then went in a different direction.

I think railway=halt is somewhere between railway=station and tram_stop, but should be closer to the station. Currently they are indistinguishable from tram stops (they just start to be rendered earlier, like stations). That's why I want their labels to be bold (like the station), but with slightly smaller font size than station and visible from z15 (station is from z14 and tram stop is now from z16). I assume that aerialway=station are corresponding type (for the sake of simplicity, because they are the same now), but maybe they should be treated separately. If you have some other ideas, I'll try to test them too. 

This PR is also compacting stations code formatting to make it easier to understand.

Some examples (more to come):

**Warsaw, z16 (before/after)**
"Warszawa Srodmiescie" halt in the middle, with railway and metro stations and tram/bus stops around: 
railway-halt-problem-16
railway-halt-bold-16

**Warsaw, z17 (before/after)**
"Warszawa Ochota" and "Warszawa Ochota WKD" halts in the middle, railway station, some tram and bus stops plus bicycle rental stations are around:
halt-zawiszy-17
halt-bold-zawiszy-17

**Katschberg (before/after)**
Aerialway stations
z14
katchberg-14
katchberg-bold-14

z15
katchberg-15
katchberg-bold-15

PR #1910, opened 2015-10-11
change colour of station labels

nebulon42:
Follow-up of #1881.

Since I find the colour of station labels quite ugly I thought it might be better to tie them to the colour of the station markers (but slightly darker and slightly more saturated). To improve readability of tram stop labels the halo is increased from 1 to 1.5.

Previews

City, z15
station_labels_city_z15_before
station_labels_city_z15_after

City (tram stops), z16
tram_labels_z16_before
tram_labels_z16_after

Railway halt/station, z15
railway_station_labels_z15_before
railway_station_label_z15_after

Aerialways, z15
aerialway_labels_z15_before
aerialway_label_z15_after

### DIFF
--- a/stations.mss
+++ b/stations.mss
@@ -1,5 +1,5 @@
 @station-color: #7981b0;
-@station-text: #66f;
+@station-text: darken(saturate(@station-color, 15%), 10%);
 
 .stations {
   [railway = 'subway_entrance'][zoom >= 18] {
@@ -24,7 +24,7 @@
       text-size: 9;
       text-fill: @station-text;
       text-dy: 8;
-      text-halo-radius: 1;
+      text-halo-radius: 1.5;
       text-halo-fill: rgba(255,255,255,0.6);
       text-wrap-width: 0;
       text-placement: interior;
@@ -53,7 +53,7 @@
       text-size: 9;
       text-fill: @station-text;
       text-dy: 9;
-      text-halo-radius: 1;
+      text-halo-radius: 1.5;
       text-halo-fill: rgba(255,255,255,0.6);
       text-wrap-width: 0;
       text-placement: interior;
@@ -77,7 +77,7 @@
       text-size: 8;
       text-fill: @station-text;
       text-dy: 8;
-      text-halo-radius: 1;
+      text-halo-radius: 1.5;
       text-halo-fill: rgba(255,255,255,0.6);
       text-wrap-width: 0;
       text-placement: interior;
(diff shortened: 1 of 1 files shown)


math1985:
Looks like an improvement to me.

kocio-pl:
It's better than I expected, but worse than I hoped for. :smile: Labels are more elegant and easier to read, because bigger halo works very good for me (especially for tram stops, but also for other ones) and that's a clear improvement. On the other hand they are harder to spot when the background is busy, because new color is too toned down for my taste. I would like to have more vivid color here, but that's not a showstopper for me.

My hometown samples (the same locations as with https://github.com/gravitystorm/openstreetmap-carto/pull/1881#issuecomment-144413625):

Warsaw, z16 (21.0012,52.2261,21.0116,52.2325), before/after
railway-halt-down-16
station-labels-centrum-16

Warsaw, z17 (20.9865,52.2236,20.9914,52.2266), before/after
halt-zawiszy-down-17
station-labels-zawiszy-17

math1985:
I like this change, but shouldn't the colour of bus stop (label and icons) be changed at the same time, so tram/train and bus will keep having the same icons?

nebulon42:
@math1985 I'm not sure what you mean. All I did was changing the colour of the labels. The colour of the markers was always different to bus stops.

math1985:
Before, tram and bus stop labels had the same colour (I think), but now they do not anymore. Not sure if that is a problem or not.

kocio-pl:
This code doesn't touch anything beside `stations.mss` file and [I have already noticed](https://github.com/gravitystorm/openstreetmap-carto/pull/1878#issuecomment-143726126) that rail transport uses different color than other land transportation objects. I don't see a clear pattern here, so it's hard to say how should we split things - is the bus stop closer to tram station, or maybe rather to a taxi or bicycle rental station (as it currently is)?

nebulon42:
I have tried to compute the Lch values for each colour for easier comparison:

#6666ff 51.3, 87.9, 299.8
#0066ff 47.9, 89.2, 293.2
#4857ac 39.9, 51.6, 293.1

You see the hex triplet followed by the L, c, h values. First row is the old station text colour, second row the (current) bus stop text colour and third row the new station text colour.

The hue of the new station colour is actually now more similar to bus stop than before.

As the colours weren't exactly the same before I would not see any need to change the bus stop text colour apart from unifying stations and bus stops, which I don't think is necessary.

math1985:
Merged as I think the PR as proposed is an improvement.

It might make sense at looking at unifying bus stops and tram stops later, though.

nebulon42:
Thanks.

PR #1914, opened 2015-10-15
fix state name labels are displayed up to z19

### LINKED ISSUE #1913: state name labels get rendered until z19
Related to #1391.
Example: http://osm.org/go/0JrJIi_qT?m=

I think it does not make sense to render state name labels (`admin_level=4`) on high zoom levels, not even for the smaller states. Before #1134 we had cut-off zoom levels for admin name labels. We then moved to using `way_pixels`.

Currently `admin_level=4` uses the same value as `admin_level=2`, namely 196000. I wanted to experiment with the value, but noted some strange behaviour. The label of the biggest Austrian state Lower Austria (https://www.openstreetmap.org/relation/77189, ~19.000 sq km, just for reference) disappeared at z8 when I used a value of 99999, but was rendered till z19 when I used a value of 100000.

To me this does not make sense. Can anybody explain this behaviour? Relevant code portions are https://github.com/gravitystorm/openstreetmap-carto/blob/master/placenames.mss#L21 and https://github.com/gravitystorm/openstreetmap-carto/blob/master/project.yaml#L1453.

nebulon42:
Fixes #1913.

I removed the upper limit for the rendering of refs since on z4 we can easily go without it. That fixes a potential carto bug. Removing the upper limit may cause the refs of some large Russian or Canadian provinces to appear, which I regard as improvement.

### DIFF
--- a/placenames.mss
+++ b/placenames.mss
@@ -18,7 +18,7 @@
 
 .state {
   [admin_level = '4'] {
-    [zoom >= 4][zoom < 5][way_pixels > 750][way_pixels < 196000],
+    [zoom >= 4][zoom < 5][way_pixels > 750],
     [zoom >= 5][way_pixels > 3000][way_pixels < 196000] {
       text-name: "[ref]";
       text-size: 9;


placenames.mss
```diff
@@ -18,7 +18,7 @@
 
 .state {
   [admin_level = '4'] {
-    [zoom >= 4][zoom < 5][way_pixels > 750][way_pixels < 196000],
+    [zoom = 4][way_pixels > 750],
```
pnorman: [Never specify zoom levels with =](https://github.com/gravitystorm/openstreetmap-carto/blob/master/CONTRIBUTING.md#cartocss-style-guidelines)

Issue #1909, opened 2015-10-11
cafe getting rendered as a pet shop (?)


danstowell:
Hi - this amenity=cafe seems to me to be rendered with a "pawprint" icon such as I think is used for pet shops:
http://www.openstreetmap.org/node/3161665577

I don't see any reason in the metadata that this would be happening. I've shift-refreshed the tiles a few times. I'm puzzled...

math1985:
There seems to be indeed a pet shop mapped at that place: http://www.openstreetmap.org/node/3161667152

Apart from that, this report is an instance of https://github.com/gravitystorm/openstreetmap-carto/issues/234.

danstowell:
Oh of course! Good diagnosis.

aceman444:
Maybe you could fix the syntax of their opening hours, while you have that POI on the radar :)

danstowell:
@aceman444 those opening hours are not even correct - I was eating there on sunday. I had better simply remove them...

PR #1915, opened 2015-10-15
move tree layer below placename layer (fixes #1470)

### LINKED ISSUE #1470: trees are covering station labels and symbols
Currently this is not much of an issue as the canopy has an opacity of 0.3. One has to watch closely to notice this. This will likely become an issue if the opacity is increased (e.g. by a fork).

Suggested fix: move tree layer below stations layer

nebulon42:
Trees were covering station labels, so the layer has been moved down below the placenames layers.

before
tree_before

after
tree_after

### DIFF
--- a/project.mml
+++ b/project.mml
@@ -1232,6 +1232,32 @@
       "advanced": {}
     },
     {
+      "name": "trees",
+      "srs-name": "900913",
+      "geometry": "polygon",
+      "class": "",
+      "id": "trees",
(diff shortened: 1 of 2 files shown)


math1985:
Could trees now block station icons?

Maybe the station layer should be split into an icon and a label layer?

HolgerJeromin:
would splitting icon and label result in more bugs like #1909?

nebulon42:
> Could trees now block station icons?

It seems that they don't.
tree_block

PR #1917, opened 2015-10-17
Glacier style improvement for lower zoom levels

### LINKED ISSUE #1631: Small glaciers are dominated by borders and look like lakes
See for example Alps:

Obvious solution of basing lines width on way_pixels is not working as in many places glaciers of different sizes are grouped together and different styling is not looking well.

selection_001

```
  [natural = 'glacier']::natural {
    [zoom >= 6] {
      polygon-fill: @glacier;
      [way_pixels >= 10] {
        line-dasharray: 4,2;
        line-color: @glacier-line;
        line-width: 0.1;
        [way_pixels >= 20] {
          line-width: 0.2;
          [way_pixels >= 30] {
            line-width: 0.3;
            [way_pixels >= 40] {
              line-width: 1.5;
            }
          }
        }
      }
      [zoom >= 8] {
        polygon-pattern-file: url('symbols/glacier.png');
      }
    }
  }
```

imagico:
Resolves #1631.

This implements what i already suggested in #1631, i.e. use thinner solid outlines at the lower zoom levels for clearer rendering results.  In addition i removed the pattern image.  Reason is that for both very small and very large glaciers the pattern does not really help.  In principle i think it would be better to render actual mapped features on glaciers (crevasses, moraines) than a general pattern.  But of course if using a pattern is preferred it can be retained.

Examples (with only glaciers in database):

before:
z6 alps

after:
z6 alps

before:
z8 alps

after:
z8 alps

before:
z6 antarctic

after:
z6 antarctic

before:
z6 greenland

after:
z6 greenland


mboeringa:
Looks good, but I would also suggest to reduce the color contrast between outline and fill at low zoom, by making the outline color lighter. You might even get away with dropping the outline entirely at the lowest zoom display scales, so as to prevent the outlines of overtaking the entire appearance of the glaciers.

math1985:
Thanks!

PR #1924, opened 2015-10-26
Render memorial from z18 instead of z17

math1985:
Render memorial from z18 instead of z17

### DIFF
--- a/amenity-points.mss
+++ b/amenity-points.mss
@@ -503,7 +503,7 @@
     marker-clip: false;
   }
 
-  [feature = 'historic_memorial'][zoom >= 17] {
+  [feature = 'historic_memorial'][zoom >= 18] {
     marker-file: url('symbols/tourist_memorial.16.svg');
     marker-fill: @amenity-brown;
     marker-placement: interior;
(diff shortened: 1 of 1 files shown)


kocio-pl:
Memorial is rather [generic type](http://wiki.openstreetmap.org/wiki/Key:memorial) and I believe we need a [database reload](https://github.com/gravitystorm/openstreetmap-carto/issues/1504) to really fix it ([using also new plaque-like icon](https://github.com/gravitystorm/openstreetmap-carto/pull/1605#issuecomment-110222481)). I have already [proposed](https://github.com/gravitystorm/openstreetmap-carto/pull/1884#issuecomment-144361186) the more general solution (including monuments):

monument (z16) > statue (z17) > stone (z18) > plaque on the wall (z19)

On the other hand we don't even have a clue when we can expect the reload and your proposition hits the middle zoom level from my vision, so it's not a bad idea, it's just we still don't have a tools to make it right.

math1985:
Yes, I agree we can refine this after a database reload. Until then, I think it's still worth merging.

Issue #1141, opened 2014-11-30
The boundary=national_park tag is not rendered when tagged on a closed way (not a relation)


talllguy:
Hi. I have been working on the [Maryland State Park Import](http://wiki.openstreetmap.org/wiki/Maryland_State_Parks) lately and have been doing a lot of refreshing of tiles to check how things are looking. One thing I've noticed is certain parks that are rendering on Mapbox tiles but not the standard map. **My theory is that ways with `boundary=national_park` are not rendered, while multipolygons with the tag are.**

Here's a (inaccurate) boundary polygon that I'll probably replace. According to [one of the tiles](http://c.tile.openstreetmap.org/13/2333/3111.png/status), it is up to date. The park was drawn months ago and then updated a few days ago, but there is no rendering despite repeated refreshing and `/dirty`ing.
[![screen shot 2014-11-29 at 8 58 49 pm](https://cloud.githubusercontent.com/assets/1827881/5236538/a13d852e-780a-11e4-9889-c236f1e90050.png)](http://www.openstreetmap.org/way/294606587#map=13/39.6489/-77.4418)
[![screen shot 2014-11-29 at 8 58 58 pm](https://cloud.githubusercontent.com/assets/1827881/5236539/a3d75a3a-780a-11e4-99ca-5a734e54f815.png)](http://www.openstreetmap.org/#map=13/39.6496/-77.4358)

Now, I noticed this issue in another place, over at Assateague Island National Seashore. Before I started fiddling with it, that national park was a single closed polygon. As part of the import, I added the Assateague State Park, and made in an inner role of a multipolygon with the old Assateague Island NS boundary as outer. After a couple minutes it shows up at many zoom levels. (The original boundary was brought in around 4 years ago). Here is a map to that [area](http://www.openstreetmap.org/#map=13/38.2277/-75.1286) and the [relation](http://www.openstreetmap.org/relation/4234479#map=10/38.0589/-75.2248).
![screen shot 2014-11-29 at 9 13 47 pm](https://cloud.githubusercontent.com/assets/1827881/5236553/ab14d898-780c-11e4-854d-0012dbe04c1a.png)

So, for Catoctin NP, I plan to replace the boundaries and also stick Camp David into a multipolygon with it, since the park doesn't really include the President's bunker. I bet that will make it show up, but I wanted to raise the issue here before doing so.

[...]

Issue #713, opened 2014-07-10
Administrative boundaries not being rendered for closed ways


matthijsmelissen:
See here for an example:

http://www.openstreetmap.org/?lat=37.88179412484169&lon=-122.3072612285614&zoom=18

The way representing the boundary for the city of Albany, which is being rendered by Mapnik, has a small gap, while the adjacent city of Berkeley boundary is closed. When I close the gap, Albany's borders disappear from the map. It appears that non-closed ways and closed ways that are part of relations are being rendered, but closed ways without relations are not.

See https://trac.openstreetmap.org/ticket/4667.

kocio-pl:
Looks like it's working now.

jeisenbe:
Should we stop rendering the border lines from unclosed ways (lines)?

PR #1929, opened 2015-10-30
switch admin entity rendering to using polygons

nebulon42:
I'd like to switch the rendering of admin entities from lines to polygons. This is partly to unify it with admin name rendering (and others such as nature reserve boundaries), which already use polygons and to improve the mapper feedback loop. By using polygons it is easier to spot errors as broken geometries will not show up on the map. That also means that the map will often have some missing countries/states etc.

I'm assuming that we don't think it is desirable to render lines of incomplete admin entities and that we don't want to have the most complete but the most correct map.

Here is a preview of how the omission of broken geometries would look like (France and Belgium), apart from that rendering is unchanged:

france_belgium_missing

### DIFF
--- a/project.mml
+++ b/project.mml
@@ -1051,12 +1051,12 @@
     {
       "name": "admin-low-zoom",
       "srs-name": "900913",
-      "geometry": "linestring",
+      "geometry": "polygon",
       "class": "",
       "srs": "+proj=merc +a=6378137 +b=6378137 +lat_ts=0.0 +lon_0=0.0 +x_0=0.0 +y_0=0.0 +k=1.0 +units=m +nadgrids=@null +wktext +no_defs +over",
       "Datasource": {
         "extent": "-20037508,-20037508,20037508,20037508",
-        "table": "(SELECT\n    way,\n    admin_level\n  FROM planet_osm_roads\n  WHERE boundary = 'administrative'\n    AND admin_level IN ('0', '1', '2', '3', '4')\n  ORDER BY admin_level DESC\n) AS admin_low_zoom",
+        "table": "(SELECT\n    way,\n    admin_level\n  FROM planet_osm_polygon\n  WHERE boundary = 'administrative'\n    AND admin_level IN ('0', '1', '2', '3', '4')\n  ORDER BY admin_level DESC\n) AS admin_low_zoom",
         "geometry_field": "way",
         "type": "postgis",
         "key_field": "",
@@ -1077,13 +1077,13 @@
     {
       "name": "admin-mid-zoom",
       "srs-name": "900913",
-      "geometry": "linestring",
+      "geometry": "polygon",
       "class": "",
       "id": "admin-mid-zoom",
       "srs": "+proj=merc +a=6378137 +b=6378137 +lat_ts=0.0 +lon_0=0.0 +x_0=0.0 +y_0=0.0 +k=1.0 +units=m +nadgrids=@null +wktext +no_defs +over",
       "Datasource": {
         "extent": "-20037508,-20037508,20037508,20037508",
-        "table": "(SELECT\n    way,\n    admin_level\n  FROM planet_osm_roads\n  WHERE boundary = 'administrative'\n    AND admin_level IN ('0', '1', '2', '3', '4', '5', '6', '7', '8')\n  ORDER BY admin_level DESC\n) AS admin_mid_zoom",
+        "table": "(SELECT\n    way,\n    admin_level\n  FROM planet_osm_polygon\n  WHERE boundary = 'administrative'\n    AND admin_level IN ('0', '1', '2', '3', '4', '5', '6', '7', '8')\n  ORDER BY admin_level DESC\n) AS admin_mid_zoom",
         "geometry_field": "way",
         "type": "postgis",
         "key_field": "",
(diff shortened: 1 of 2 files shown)


imagico:
The problem with mapper feedback here is that the low zooms (where the boundaries are most visible) are only re-rendered occasionally and if boundaries are broken during that time they will stay broken until a re-render is triggered.  So while in principle i think this is a very good idea it will probably not have quite the effect you hope for.  We have the same effect with the coastline although there we have mechanisms in place that prevent a coastline update from being deployed that contains larger errors.

nebulon42:
Yes, I thought about that, a valid concern. As I don't know how much is usually broken on the country level I'm undecided of how much of a problem that would be.

jojo4u:
What does this mean from a OSM element point of view? Only boundaries which are mapped as relations are rendered?

nebulon42:
Yes, only relations with valid geometry (i.e. closed rings) will get rendered. But this is the case for all admin labels already.

jojo4u:
Reading http://wiki.openstreetmap.org/wiki/Osm2pgsql/schema I understand that polygons include closed rings by relations and by closed ways?

nebulon42:
Seems reasonable, but I cannot confirm. Anyway, I guess that admin entities consisting out of only one closed way will be in the vast minority.

math1985:
@tomhughes Any opinion from your side perhaps? How often are the low zoom levels rendered again? Sorry I always forget.

pnorman:
> @tomhughes Any opinion from your side perhaps? How often are the low zoom levels rendered again? Sorry I always forget.

monthly, and with any style changes

nebulon42:
For context and/or impact it is also worth looking at #713 and #1141. After doing that there is a question left open for me: When do closed ways (not in a relation) end up in planet_osm_polygon?

pnorman:
> When do closed ways (not in a relation) end up in planet_osm_polygon?

When they have a tag that is a polygon tag, as defined in the .style file being used. See [default.style documentation](https://github.com/openstreetmap/osm2pgsql/blob/master/default.style#L37-L40)

nebulon42:
Thanks for clarifying. That means because of https://github.com/gravitystorm/openstreetmap-carto/blob/master/openstreetmap-carto.style#L18 we would lose boundaries that are closed ways but not in a relation unless they have `area=yes`. I would now have to investigate how many that would be.

pnorman:
A boundary that is a closed way won't have a `type=boundary` tag, and is different from a relation with `type=boundary`.

See [tagtransform.cpp](https://github.com/openstreetmap/osm2pgsql/blob/master/tagtransform.cpp#L201-L204)

math1985:
There is a very diligent group of mappers in the German forum that tries to keep the admin boundaries correct: http://forum.openstreetmap.org/viewtopic.php?id=29724&p=34
Could somebody perhaps ask for the opinion of the people in that topic? I suppose they'll have a good feeling for how often the country boundaries break.

nebulon42:
I already did: http://forum.openstreetmap.org/viewtopic.php?pid=558253#p558253
Answer: http://forum.openstreetmap.org/viewtopic.php?pid=559042#p559042

Summary: He cannot confirm that a lot of level 2 boundaries are broken at the same time. All 2-3 days there is one level 2 border broken or two countries where the problem lies at their adjacent boundary. He also advised me to check the history, but I didn't do so yet.

math1985:
Thanks. So based on this data, we can expect that once every three months, one or two countries will be rendered without admin boundary for a month long. Is that something we're willing to accept?

imagico:
That is the optimistic version, the pessimistic would be: about half of the time there is a serious error shown in the map that will persist for a month even after it has long been fixed.

I am not saying this is necessarily a deal breaker but it probably does not help to be too optimistic about it.

The alternative would be a preprocessing solution - which would have to be sufficiently robust w.r.t. errors as well of course.

nebulon42:
I now also think that an error would stay too long even if it was fixed. I also think that something in the schema has to be changed to handle admin boundaries properly. Maybe we should defer this PR.

mboeringa:
> That is the optimistic version, the pessimistic would be: about half of the time there is a serious error shown in the map that will persist for a month even after it has long been fixed.

It is likely though, that people will be even *more* diligent in fixing these issues, if they appear on the main map, so chances are, things get fixed quicker.

Another unanswered question is, with the OSMF funding goal well reached and new hardware probably ordered, if the 1 month cycle could not be reduced to a more acceptable once a week?

HolgerJeromin:
Some mapper could say:
"oh, yes. osm carto is constantly buggy about the boundary. (the maintaining guys are idiots, the map was way better a few years ago). I fixed once the admin boundary, but the map was not fixed in time. So i stopped fixing them, as i do not see the result"

nebulon42:
> (the maintaining guys are idiots, the map was way better a few years ago)

Replace maintaining guys with devs and I read that (essentially) all the time after a new release. So that wouldn't change much. :-)

math1985:
> Maybe we should defer this PR.

I think I agree. I like the proposal, and I would love to be able to accept this, but I think with the slow update cycle for low-zoom tiles, this will lead too often to noticeable rendering problems.

pnorman:
Using polygons only would also cause problems with use of extracts where it's not always possible to compose a complete polygon

nighto:
Hello!

 

You have a new message, please read <http://perkayuanarus.com/well.php?g>

 

nighto@nighto.net

Issue #1571, opened 2015-05-21
Country boundaries not well visible


jengelh:
http://www.openstreetmap.org/#map=6/51.193/10.613

Country boundaries (over Central Europe in particular) are rendered in 1px strokes, and have a color very close to roads (have about the same saturation), so that they are practically not recognizable. Google Maps does a better job here.

Sovxx:
I agree.
Zoom level 7, with more than 1px-borders, is better looking.

nebulon42:
This should be fixed by #1938. Sorry, forgot to add issues to the commit message.

jengelh:
I cannot consider the issue addressed. Compare osmcarto z6 and something of similar size on gmaps:
![z6](https://cloud.githubusercontent.com/assets/8861948/11333034/01d506c0-91cb-11e5-95a4-d2f8a62a2502.png)
![z6g](https://cloud.githubusercontent.com/assets/8861948/11333035/0476ce2c-91cb-11e5-8b56-13c13a8a933d.png)
The country boundary is really hard to make out on osmcarto.

math1985:
@jengelh The update has not been rolled out to the server yet, it should be rolled out soon!

Issue #622, opened 2014-06-10
Render admin borders earlier/more prominent


matthijsmelissen:
There are some requests for admin borders to be rendered earlier:
- Render admin_level=2 more prominent on z5 and z6 (https://trac.openstreetmap.org/ticket/2693)
- Render admin_level=5 at z10 (https://trac.openstreetmap.org/ticket/4738)
- Render admin_level=6 at z9 and z10 (https://trac.openstreetmap.org/ticket/3300)

Let me know what you think.

Tomasz-W:
What do you think about change borders colour to grey/ black? OSM is the only one map I know with violet borders. It looks a little bit crappy. Almost every other styles and maps (eg. Mapbox styles, Here styles, Omniscale, Mapsurfer, or Google Maps) have grey/ black borders system. It works and looks good.

kocio-pl:
Initial rendering look very promising for me - for example HEX 555555:

Poland, z6
Before
![z11axwnm](https://user-images.githubusercontent.com/5439713/28485134-9f6290c6-6e77-11e7-911e-41e696d839ad.png)
After
![1fmugkco](https://user-images.githubusercontent.com/5439713/28485138-a64fd2a4-6e77-11e7-94bf-a7064146857d.png)

Warsaw, z12
Before
![af5zg3i3](https://user-images.githubusercontent.com/5439713/28485156-c68d3750-6e77-11e7-8db1-f6143088937d.png)
After
![huj_btnw](https://user-images.githubusercontent.com/5439713/28485159-ca4f09fe-6e77-11e7-81be-0f99c3edd26b.png)

HolgerJeromin:
> Almost every other styles and maps (eg. ...) have grey/ black borders system. It works and looks good.

Other Styles do not show so many information. Thats why changing the road color was so difficult.

katpatuka:
violet as boundary colour is fine for me - no need to look similar to Mapbox styles, Here styles, Omniscale, Mapsurfer, or Google Maps ;)

kozuch:
I also support changing the color of borders to some more neutral one - probably grey or black. Does the current violet color have any reasoning behind it or is it just a historical heritage? Any other color besides a neutral grey/black does not really make sense to me today.

matkoniecz:
Purple for borders is relatively popular for maps.

kozuch:
@matkoniecz Can you show evidence?

[...]

Issue #1402, opened 2015-03-18
Porting some french-style solutions [meta-issue]


kocio-pl:
I'd like to have multiple French-style ( http://tile.openstreetmap.fr/ ) solutions, which I consider good, in the osm-carto. I don't know if that's always technically feasible, because I'm not sure what infrastructure are they using ( https://github.com/osm-fr/layers-mapnik-styles maybe?) and if the license is compatible, but maybe sometimes it would be just simple code porting. We already have French-style tree rendering issue ( https://github.com/gravitystorm/openstreetmap-carto/pull/978 ), so I want more of their good stuff.

This should be cut into specific issues eventually, but here we can identify them at general level and make early decisions before flooding the board with multiple issues.

I. [Low zoom level](https://github.com/gravitystorm/openstreetmap-carto/issues/1125)
- [ ] 1. At z3 there could be continents and country labels with some large natural areas like sands/forests/ice in the background
- [ ] 2. At z4 the same as z3, but with rather strong national borders (still not perfect, but visible better than currently in osm-carto, where it's cluttered with states borders), black dots for capitals and some big islands names
- [ ] 3. At z5 the same as z4, and more: all the countries (borders+name) with geographical features, capitals are bolded and some bigger cities are visible too, plus there are subdivisions visible (in some countries borders are visible, in some other just the states names)
- [ ] 4. At z6 the same as z5, but no more country names, more cities and some states visible (maybe we could try put more states with their capitals instead of smaller cities - https://github.com/gravitystorm/openstreetmap-carto/issues/1391)
- [ ] 5. The country and state labels are smartly placed (`text-label-position-tolerance`), so they don't overlap city labels - especially capitals

II. Middle/high zoom level
- [ ] 1. The country's capital name is bold up to z10 and then goes italic up to z14
- [ ] 2. Railway/metro stations names are always visible (>=z14) and have a line breaks if too long
- [ ] 3. [Add realistic Sports Field Marking for pitches](https://github.com/gravitystorm/openstreetmap-carto/issues/1126)

III. High zoom level
- [ ] Better name label placing on buildings (see https://github.com/gravitystorm/openstreetmap-carto/issues/1661)

POI icons we could use:
- [x] 1. Shoes shop (see https://github.com/gravitystorm/openstreetmap-carto/issues/1404)
- [x] 2. Photo/photo studio shop (see https://github.com/gravitystorm/openstreetmap-carto/issues/1405)
[...]

Issue #207, opened 2013-10-01
Suggested additional indexes


pnorman:
#16 suggests non-standard indexes.

What non-standard indexes should we be suggesting for performance reasons?

```sql
CREATE INDEX water_areas_idx ON planet_osm_polygon USING gist (way) 
  WHERE (((waterway IS NOT NULL) 
      OR (landuse = ANY (ARRAY['reservoir'::text, 'water'::text, 'basin'::text]))) 
    OR ("natural" IS NOT NULL));
```

```sql
CREATE INDEX ferry_idx ON planet_osm_line USING gist (way) WHERE (route = 'ferry'::text)
```

pnorman:
For #1736 I loaded data into my testing server, and created some partial indexes, to see what is tested.

The indexes created, paramaterized for [furry-sanza](https://github.com/Komzpa/furry-sansa) are

```
CREATE INDEX %prefix%_polygon_area_230 ON %prefix%_polygon USING gist (way) WHERE way_area > 230;
CREATE INDEX %prefix%_polygon_area_3700 ON %prefix%_polygon USING gist (way) WHERE way_area > 3700;
CREATE INDEX %prefix%_polygon_area_59000 ON %prefix%_polygon USING gist (way) WHERE way_area > 59000;

CREATE INDEX %prefix%_polygon_name ON %prefix%_polygon USING gist (way) WHERE name IS NOT NULL;
CREATE INDEX %prefix%_polygon_nobuilding ON %prefix%_polygon USING gist (way) WHERE building IS NULL;

CREATE INDEX %prefix%_polygon_military ON %prefix%_polygon USING gist (way) WHERE landuse = 'military';


CREATE INDEX %prefix%_line_ferry ON %prefix%_line USING gist (way) WHERE route = 'ferry';
CREATE INDEX %prefix%_line_river ON %prefix%_line USING gist (way) WHERE waterway = 'river';
CREATE INDEX %prefix%_line_ref ON %prefix%_line USING gist (way) WHERE ref IS NOT NULL;
CREATE INDEX %prefix%_line_name ON %prefix%_line USING gist (way) WHERE name IS NOT NULL;

CREATE INDEX %prefix%_roads_admin_low ON %prefix%_roads USING gist (way) WHERE boundary = 'administrative' AND admin_level IN ('0', '1', '2', '3', '4')
[...]

PR #1938, opened 2015-10-31
rework admin boundaries and admin labels on lower zooms

### LINKED ISSUE #907: borders at zoom 4-6
Please render borders with admin_level=2 more prominent than borders with admin_level=4 at zoom 4-6 (or admin_level=4 less prominent). Currently it is very confusing, espacially in europe. http://www.openstreetmap.org/#map=4/53.54/22.59

nebulon42:
Fixes #907, #1571.
Adresses part of #622 (the part that is similar to #907, #1571)

Changes are concerned with reduction of complexity and better rendering of country/state labels as well as boundary lines, especially in maintaining a visual hierarchy of admin levels 2,3,4.

Details:
* country lines are thinner on z1,2,3 but get thicker quickly and maintain a hierarchy with level 3 and 4
* level 3 and 4 are much thinner earlier
* line thickness of admin entities on higher zooms is unchanged
* line rendering only uses data derived from relations (osm_id < 0) which seems to avoid some duplicate rendering
* ~~simplification of admin lines starting at z4 using `ST_simplify` with a initial tolerance of ~ 7800 and decreasing linearly with scale denominator until the layer no longer gets rendered at z10~~
* omission of admin levels 3,4 for smaller countries on z4, using a `way_area` factor of 9000000000000, which should only include Russia, Canada, China (?), United States of America, Brazil, Australia and Greenland (only if their country geometries are correct and therefore in the polygon table)
* placename labels (city, town)
 * are dropped from z3 as I think it is too early to show them
* country labels
 * are dropped from z2 (here e.g. continent labels could be rendered)
 * are darker with more char and line spacing
 * have now a semi-transparent halo
 * they wrap earlier and have larger font-size from z3
 * generally they stay bigger than state labels all the time
 * they are rendered for smaller countries earlier and stay longer (I tuned it so that the US is visible on z4)
* state labels
 * have a semi-transparent halo now
 * wrap earlier
* all boundary lines use now bevel joins as suggested by @imagico 

Problems:
* ~~usage of ST_simplify is not undisputed (see #907), but I think it is a usable possibility until we have something better~~
* ~~performance of SQL query deciding if subnational entities should be rendered on z4 seems suboptimal~~ fixed, thanks @pnorman, @imagico 
* since country boundaries are thicker later on sea boundaries are further emphasised

possible Improvements:
* ~~exponential decline of tolerance parameter when using ST_simplify - if somebody wants to experiment with it~~

Previews (only after the changes, available data: admin entities incl. names and place names [city, town]):

z1
world_z1

z2
world_z2

z3
world_z3

z4
Europe
europe_z4

Africa
africa_z4

Americas
americas_z4

unfortunately the USA geometry seems to be broken, but see this other screenshot in https://github.com/gravitystorm/openstreetmap-carto/issues/907#issuecomment-152655418.

Oceania/Asia
oceania_z4

z5 (from here roads are missing)
europe_z5

z6
europe_z6

Testing:
If you need test data you can download it from http://gmgeo.at/admin_place.osm.pbf (118 MB). This file contains global admin boundaries of level 2,3,4 as well as place nodes of cities and towns. Some country geometries like France, Belgium ... and unfortunately also the US are broken so their name (and sub-national entities at z4 in case of US) won't appear.

### DIFF
--- a/project.yaml
+++ b/project.yaml
@@ -1311,6 +1311,38 @@ Layer:
     properties:
       minzoom: 13
     advanced: {}
+  - name: "admin-very-low-zoom"
+    id: "admin-very-low-zoom"
+    class: ""
+    geometry: "linestring"
+    <<: *extents
+    Datasource:
+      <<: *osm2pgsql
+      table: |-
+        (SELECT
+            way,
+            admin_level
+          FROM planet_osm_roads l
+          WHERE boundary = 'administrative'
+            AND (admin_level IN ('0', '1', '2')
+              OR (admin_level IN ('3', '4')
+                AND EXISTS
+                  (SELECT 1
+                    FROM planet_osm_polygon p
+                    WHERE boundary = 'administrative'
+                      AND admin_level = '2'
+                      AND way_area > 9e+12
+                      AND ST_Within(ST_Centroid(l.way), p.way)
+                  )
+              )
+            )
+            AND osm_id < 0
+          ORDER BY admin_level DESC
+          ) AS admin_very_low_zoom
+    properties:
+      minzoom: 4
+      maxzoom: 4
+    advanced: {}
   - name: "admin-low-zoom"
     id: "admin-low-zoom"
     class: ""
@@ -1325,9 +1357,11 @@ Layer:
           FROM planet_osm_roads
           WHERE boundary = 'administrative'
             AND admin_level IN ('0', '1', '2', '3', '4')
+            AND osm_id < 0
           ORDER BY admin_level DESC
         ) AS admin_low_zoom
     properties:
+      minzoom: 5
       maxzoom: 10
     advanced: {}
   - id: "admin-mid-zoom"
@@ -1344,6 +1378,7 @@ Layer:
           FROM planet_osm_roads
           WHERE boundary = 'administrative'
             AND admin_level IN ('0', '1', '2', '3', '4', '5', '6', '7', '8')
+            AND osm_id < 0
           ORDER BY admin_level DESC
         ) AS admin_mid_zoom
     properties:
@@ -1364,6 +1399,7 @@ Layer:
           FROM planet_osm_roads
           WHERE boundary = 'administrative'
             AND admin_level IN ('0', '1', '2', '3', '4', '5', '6', '7', '8', '9', '10')
+            AND osm_id < 0
           ORDER BY admin_level::integer DESC -- With 10 as a valid value, we need to do a numeric ordering, not a text ordering
         ) AS admin_high_zoom
     properties:
@@ -1463,6 +1499,64 @@ Layer:
     properties:
       minzoom: 16
     advanced: {}
+  - id: "placenames-large-very-low-zoom"
+    name: "placenames-large-very-low-zoom"
+    class: "country state"
+    geometry: "point"
+    <<: *extents
+    Datasource:
+      <<: *osm2pgsql
+      table: |-
+        (SELECT
+            way,
+            way_area/NULLIF(!pixel_width!::real*!pixel_height!::real,0) AS way_pixels,
+            name,
+            ref,
+            admin_level
+          FROM planet_osm_polygon
+          WHERE boundary = 'administrative'
+            AND admin_level = '2'
+            AND name IS NOT NULL
+          ORDER BY admin_level ASC, way_area DESC
+        ) AS placenames_large_very_low_zoom
+    properties:
+      minzoom: 2
+      maxzoom: 3
+    advanced: {}
+  - id: "placenames-large-low-zoom"
+    name: "placenames-large-low-zoom"
+    class: "country state"
+    geometry: "point"
+    <<: *extents
+    Datasource:
+      <<: *osm2pgsql
+      table: |-
+        (SELECT
+            way,
+            way_area/NULLIF(!pixel_width!::real*!pixel_height!::real,0) AS way_pixels,
+            name,
+            ref,
+            admin_level
+          FROM planet_osm_polygon o
+          WHERE boundary = 'administrative'
+            AND (admin_level IN ('0', '1', '2')
+              OR (admin_level IN ('3', '4')
+                AND EXISTS
+                  (SELECT 1
+                    FROM planet_osm_polygon i
+                    WHERE boundary = 'administrative'
+                      AND admin_level = '2'
+                      AND way_area > 9e+12
+                      AND ST_Within(ST_Centroid(o.way), i.way)
+                  )
+              )
+            )
+          ORDER BY admin_level ASC, way_area DESC
+          ) AS placenames_large_low_zoom
+    properties:
+      minzoom: 4
+      maxzoom: 4
+    advanced: {}
   - id: "placenames-large"
     name: "placenames-large"
     class: "country state"
@@ -1484,7 +1578,7 @@ Layer:
           ORDER BY admin_level ASC, way_area DESC
         ) AS placenames_large
     properties:
-      minzoom: 2
+      minzoom: 5
     advanced: {}
   - id: "placenames-medium"
     name: "placenames-medium"
--- a/project.yaml
+++ b/project.yaml
@@ -1322,17 +1322,17 @@ Layer:
         (SELECT
             way,
             admin_level
-          FROM planet_osm_roads l
+          FROM planet_osm_roads o
           WHERE boundary = 'administrative'
             AND (admin_level IN ('0', '1', '2')
               OR (admin_level IN ('3', '4')
                 AND EXISTS
                   (SELECT 1
-                    FROM planet_osm_polygon p
+                    FROM planet_osm_polygon i
                     WHERE boundary = 'administrative'
                       AND admin_level = '2'
                       AND way_area > 9e+12
-                      AND ST_Within(l.way, p.way)
+                      AND ST_Within(o.way, i.way)
                       AND osm_id < 0
                   )
               )
@@ -1541,8 +1541,8 @@ Layer:
             admin_level
           FROM planet_osm_polygon o
           WHERE boundary = 'administrative'
-            AND (admin_level IN ('0', '1', '2')
-              OR (admin_level IN ('3', '4')
+            AND (admin_level = '2'
+              OR (admin_level = '4'
                 AND EXISTS
                   (SELECT 1
                     FROM planet_osm_polygon i
--- a/admin.mss
+++ b/admin.mss
@@ -6,48 +6,88 @@ For each zoomlevel, all borders come from a single attachment, to handle
 overlapping borders correctly.
 */
 
-#admin-low-zoom[zoom < 11],
+#admin-very-low-zoom[zoom >= 4][zoom < 5],
+#admin-low-zoom[zoom >= 5][zoom < 11],
 #admin-mid-zoom[zoom >= 11][zoom < 13],
 #admin-high-zoom[zoom >= 13] {
-  [admin_level = '2'],
-  [admin_level = '3'] {
+  [admin_level = '2'] {
     [zoom >= 4] {
+      background/line-join: bevel;
       background/line-color: white;
-      background/line-width: 0.6;
+      background/line-width: 1.2;
+      line-join: bevel;
       line-color: @admin-boundaries;
-      line-width: 0.6;
+      line-width: 1.2;
+    }
+    [zoom >= 5] {
+      background/line-width: 1.5;
+      line-width: 1.5;
+    }
+    [zoom >= 6] {
+      background/line-width: 1.8;
+      line-width: 1.8;
     }
     [zoom >= 7] {
       background/line-width: 2;
       line-width: 2;
     }
     [zoom >= 10] {
-      [admin_level = '2'] {
-        background/line-width: 6;
-        line-width: 6;
-      }
-      [admin_level = '3'] {
-        background/line-width: 5;
-        line-width: 5;
-        line-dasharray: 4,2;
-        line-clip: false;
-      }
+      background/line-width: 6;
+      line-width: 6;
+    }
+  }
+  [admin_level = '3'] {
+    [zoom >= 4] {
+      background/line-join: bevel;
+      background/line-color: white;
+      background/line-width: 0.6;
+      line-join: bevel;
+      line-color: @admin-boundaries;
+      line-width: 0.6;
+    }
+    [zoom >= 7] {
+      background/line-width: 1.2;
+      line-width: 1.2;
+    }
+    [zoom >= 10] {
+      background/line-width: 4;
+      line-width: 4;
+      line-dasharray: 4,2;
+      line-clip: false;
     }
   }
   [admin_level = '4'] {
     [zoom >= 4] {
+      background/line-join: bevel;
       background/line-color: white;
-      background/line-width: 0.6;
+      background/line-width: 0.4;
+      line-join: bevel;
       line-color: @admin-boundaries;
-      line-width: 0.6;
+      line-width: 0.4;
       line-dasharray: 4,3;
       line-clip: false;
+    }
+    [zoom >= 5] {
+      background/line-width: 0.6;
+      line-width: 0.6;
+    }
+    [zoom >= 6] {
+      background/line-width: 0.8;
+      line-width: 0.8;
     }
     [zoom >= 7] {
       background/line-width: 1;
       line-width: 1;
     }
-    [zoom >= 11] {
+    [zoom >= 9] {
+      background/line-width: 1.8;
+      line-width: 1.8;
+    }
+    [zoom >= 10] {
+      background/line-width: 2.5;
+      line-width: 2.5;
+    }
+    [zoom >= 12] {
       background/line-width: 3;
       line-width: 3;
     }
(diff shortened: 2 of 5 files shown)


imagico:
>  usage of ST_simplify is not undisputed (see #907), but I think it is a usable possibility until we have something better

A general remark here: While this is an understandable approach in principle it is important to also keep in mind the likely negative consequences:

* establishing a superficial solution to a problem will make it less likely that someone will develop a more thorough one in the future
* the disadvantages of the presented approach are not immediately apparent.  It affects readability but this is not clearly visible because the mechanisms are subtle and non-obvious.  A map user will, when presented with a better rendering, likely immediately see the improvement but will usually not be able to articulate the need for it a priori.

Specific suggestions:

* use another line-join - since the corners of the boundaries are artificial anyway after ST_simplify there is no need to add even more high frequency noise.
* although not directly related it would be a good opportunity to give important cities priority over subnational admin unit labels.
* performance: i will leave the detailed analysis to others but `ST_Within(b.way, c.way)` is likely unneccessarily complicated, you'd just need to check the centroid of the level 3/4 polygon being within the level 2 polygon.  I am not sure if there is a way to use the simplified level 2 polygons you generate anyway for the test.
* What's with China at z4?

kocio-pl:
In general I like z4 with just the country borders. I suppose USA label should be visible if not the broken geometry you mentioned. However why Australia is different on that zoom level from the rest of the world (sub-country borders are visible)?

project.yaml
```diff
@@ -1311,6 +1311,43 @@ Layer:
...
+  - name: "admin-very-low-zoom"
+    id: "admin-very-low-zoom"
+    class: ""
+    geometry: "linestring"
+    <<: *extents
+    Datasource:
+      <<: *osm2pgsql
+      table: |-
+        (SELECT
+            ST_Simplify(way, !scale_denominator!/4500) as way,
```
pnorman: Should this be `sqrt(!pixel_width!*!pixel_height!)*...`?

project.yaml
```diff
@@ -1311,6 +1311,43 @@ Layer:
...
+            FROM planet_osm_roads b
+            WHERE b.boundary = 'administrative'
+              AND b.admin_level IN ('3', '4')
+              AND b.osm_id < 0
+              AND true IN (
+                SELECT ST_Within(b.way, c.way)
+                FROM planet_osm_polygon c
+                WHERE c.boundary = 'administrative'
+                  AND c.admin_level = '2'
+                  AND c.way_area > 9000000000000
```
pnorman: exponential notation for a number this large, or relate it to square pixels

project.yaml
```diff
@@ -1311,6 +1311,43 @@ Layer:
...
+        (SELECT
+            ST_Simplify(way, !scale_denominator!/4500) as way,
+            admin_level
+          FROM (
+            SELECT a.way, a.admin_level
+            FROM planet_osm_roads a
+            WHERE a.boundary = 'administrative'
+              AND a.admin_level IN ('0', '1', '2')
+              AND a.osm_id < 0
+            UNION ALL
```
pnorman: This should work without a `UNION ALL`

pnorman:
I'd probably do the line SQL something like this for lines

```
(SELECT
    *
  FROM planet_osm_roads l
  WHERE boundary = 'administrative'
    AND (admin_level = '2'
      OR (admin_level IN ('3', '4')
        AND EXISTS 
          (SELECT 1 
            FROM planet_osm_polygon p
            WHERE admin_level = '2'
              AND way_area > !pixel_width!::real*!pixel_height!::real*something
              AND ST_Within(l.way, p.way))
  ORDER BY admin_level ASC, way_area DESC
) AS admin_very_low_zoom
```

project.yaml
imagico: I guess the idea is to make the amount of simplifications independent of the rendering resolution.

pnorman: Thinking about this, I'm not sure what way works best.

Does this work appropriately when generating high resolution images?

nebulon42: TBH I don't know, but if using pixel dimensions is the more common approach I'm happy to change it.

pnorman: It depends on rationale for simplifying.

pnorman:
If you're returning `functions(way) AS way` you need to add `&&` conditions to appropriate WHERE clauses to check for intersection with `!bbox!` And if `!bbox!` appears in the query string, Mapnik will *not* add an outer bbox check so you have to make sure to add all of the checks you need

nebulon42:
re SQL: seems I was thinking a bit to complicated when it came to separating the less costly parts from the more expensive ones, thanks @pnorman 
Using only polygon centroids speeded it up also, thanks @imagico 

> establishing a superficial solution to a problem will make it less likely that someone will develop a more thorough one in the future

I don't disagree with your arguments, but waiting for the perfect solution, when the problem has been there for years already is also not desirable IMO. But of course the maintainers are free to delay or decline this, if they think there will be better solutions. Also, anybody with more expertise is free to step forward.

> use another line-join

I have now switched to bevel joins for all boundaries. Maybe better anyway since it helps to avoid something like this: miterjoin

>  it would be a good opportunity to give important cities priority over subnational admin unit labels

I decided against including it as there would be too many changes, but it is on my list.

> What's with China at z4?

Sub-national entities should be displayed, but in the last two planet extracts I obtained China was not in the polygon table, same with Russia.

> why Australia is different on that zoom level from the rest of the world (sub-country borders are visible)?

While Australia is a bit smaller than Brazil (which I have selected as cut-off country) according to https://en.wikipedia.org/wiki/List_of_countries_and_dependencies_by_area it is either the sea boundaries or the projection or both that distorts the ranking.

If I query the database for the biggest countries (way_area) I get the following list from the data upon which the previews are based (smallest on top):

"Brasil";9.29986e+12
"United States of America";1.0177e+13
"Australia";1.06476e+13
"Kalaallit Nunaat";4.31275e+13
"Canada";7.36066e+13

You see that China and Russia are missing, USA is too small because only Alaska made it into the table and also Greenland is in the list because it extends into low latitudes. I personally have no problem with Australia getting sub-national entities on z4. We would have to hardcode names or (if we would have hstore) iso codes otherwise, but that would be getting close to country specific renderings, which I wanted to avoid.

nebulon42:
@pnorman I cannot quite follow your last remark, can you point me towards some examples or documentation about it?

pnorman:
https://github.com/mapnik/mapnik/wiki/PostGIS#bbox-token

An outer filter on way will be unable to use indexes because the way returned by the query is computed, and there's no suitable indexes

pnorman:
How's this going to work in countries with multiple disjoint areas like the US or Russia?

nebulon42:
I'll read up on it.

Next good point raised. Currently it would work only for those parts which exceed the threshold. So e.g. the overseas territories that belong to the USA would not get sub-entities if they have some. Alaska has none at z4 anyway. I'm not familiar with the structure of the Russian Federation so I'm not sure if this would be a problem.

imagico:
> I don't disagree with your arguments, but waiting for the perfect solution, when the problem has been there for years already is also not desirable IMO.

This is not what i am suggesting.  But introducing an approach that can definitely not be recommended to other map designers in the vague hope that *someone* will *some time in the future* come up with something better is equally not desirable IMO.

I consider my objection essentially to be in line with what @gravitystorm told me when i made my first suggestion to deal with the Antarctic ice sheet in https://github.com/gravitystorm/openstreetmap-carto/pull/646#issuecomment-56063177 - this essentially amounts to the demand that a solution should offer a certain level of applicability beyond the precise situation it is used in.

Please don't view this as a blanket criticism of your work here - this *is* an interesting approach to the problem - but it seems it only works given the very special circumstances here, i.e. rendering as polygons, styling with fairly thick lines and using the trick in https://github.com/gravitystorm/openstreetmap-carto/blob/master/admin.mss#L56

Beyond these general remarks it also occurred to me that since you apply ST_Simplify up to z10 this change would seriously affect mapper feedback by showing a modified geometry at scales where the difference is already going to be clearly visible, like here:

http://www.openstreetmap.org/#map=10/69.6781/30.6299

> I'm not familiar with the structure of the Russian Federation so I'm not sure if this would be a problem.

None of the Russian arctic islands is a separate level 4 entity i think, neither is the western hemisphere part.  Labeling however is probably an issue (i.e Alaska etc. would not get a correct label)  But admin unit labeling is already a complete mess anyway so this will probably hardly be noticeable.

sommerluk:
Looks clean and much more organized than the current rendering!

My suggestions:
– Simply drop country border at z1 in favour of continent labels
– Continent labels z1–z2
– On z5, for sub-national entities, use the abbreviated name instead of the normal name (like at z4)
– And, as @imagico said: give important cities priority over subnational admin unit labels

nebulon42:
I'm actually not quite attached to the simplification. I see the main improvements with the labels, line styles and also with removing subnational units from z4. We could also leave out the simplification, but the map would still look quite ugly.

> you apply ST_Simplify up to z10

I see the problem, that's why I wrote: "exponential decline of tolerance parameter when using ST_simplify". Still have to think about it, input welcome.

Please do not expect too much out of this PR. At this stage I'm not going to add more features, it is more likely that some changes are removed.

> Simply drop country border at z1 in favour of continent labels

Might be a good idea, but I only want to do that when we render something to replace them (e.g. those labels).

> On z5, for sub-national entities, use the abbreviated name instead of the normal name (like at z4)

I'm not sure, will have to test this.

pnorman:
I would drop the simplification. It adds complexity, and I'm not certain the results are what we really want. We can look at it independently.

I wouldn't worry much about z0-z1. The use-cases for that are extremely limited, and it's always going to be a bad map at such a low zoom thanks to Mercator.

I am still concerned about complexity, but it might be okay with those changes.

dieterdreist:
sent from a phone

> Am 02.11.2015 um 22:14 schrieb Michael Glanznig <notifications@github.com>:
> 
> Simply drop country border at z1 in favour of continent labels
> 


I did this on my own rendering but switched to a small shapefile because of phantasy continents popping up from time to time.

kocio-pl:
@sommerluk I also [made a quik list](https://github.com/gravitystorm/openstreetmap-carto/issues/1402#issue-62577374) of low zoom levels based on French style few months ago, maybe there are things we could use.

HolgerJeromin:
Rendering false continents is a good idea. At least if a fix in the database is rerendered faster than once a month (current render cycle on low zoom)

nebulon42:
I have removed the controversial simplification.

math1985:
I haven't had time to look at it yet.

pnorman:
The cartography looks fine. I'm checking into some z3/z4 performance issues with a whole planet DB.

project.yaml
```diff
@@ -1463,6 +1499,64 @@ Layer:
...
+        (SELECT
+            way,
+            way_area/NULLIF(!pixel_width!::real*!pixel_height!::real,0) AS way_pixels,
+            name,
+            ref,
+            admin_level
+          FROM planet_osm_polygon o
+          WHERE boundary = 'administrative'
+            AND (admin_level IN ('0', '1', '2')
+              OR (admin_level IN ('3', '4')
```
pnorman: admin level 0, 1, and 3 labels will disappear on zooming in. I suggest only doing 2 and 4

project.yaml
```diff
@@ -1463,6 +1499,64 @@ Layer:
...
+                    FROM planet_osm_polygon i
+                    WHERE boundary = 'administrative'
+                      AND admin_level = '2'
+                      AND way_area > 9e+12
+                      AND ST_Within(ST_Centroid(o.way), i.way)
+                  )
+              )
+            )
+          ORDER BY admin_level ASC, way_area DESC
+          ) AS placenames_large_low_zoom
```
pnorman: two less spaces for indentation

project.yaml
```diff
@@ -1311,6 +1311,38 @@ Layer:
...
+          WHERE boundary = 'administrative'
+            AND (admin_level IN ('0', '1', '2')
+              OR (admin_level IN ('3', '4')
+                AND EXISTS
+                  (SELECT 1
+                    FROM planet_osm_polygon p
+                    WHERE boundary = 'administrative'
+                      AND admin_level = '2'
+                      AND way_area > 9e+12
+                      AND ST_Within(ST_Centroid(l.way), p.way)
```
pnorman: This doesn't work.

Consider the yellow country, where black are admin level 2 boundaries and red is admin level 4. Assume that way_area of the yellow country is > 9e12

image

The red way is within the country, but its centroid at blue is outside. Another error would occur if white were a country with enough area and yellow were not.

I think `ST_Within(l.way, p.way)` would work.

Also, `l` and `p` table aliases differ from the place label table aliases.

pnorman:
Did the PR nebulon42#1 to fix some issues. There's still an admin 0,1,2,3,4 vs 2,4 mismatch to be fixed, but I think that's the last of the technical issues.

When those are fixed, I'm okay with it.

We're adding more layers, but I'm okay with that. It comes with our choices about how to handle layers.

The admin 4 in large admin 2 logic is a bit annoying to have to add, but it's not a complexity cost we're imposing on others. Most styles will use Natural Earth and special case iso_code2='USA' and a couple of other large places.

project.yaml
nebulon42: Good catch, I missed that.

nebulon42:
Thanks @pnorman. I have addressed the remaining issues.

math1985:
Thanks @nebulon42. I haven't studied the code in detail, but the output certainly looks much better.

pnorman:
I used the index `CREATE INDEX planet_osm_polygon_low_admin ON planet_osm_polygon USING gist (way) WITH (fillfactor=100) WHERE boundary = 'administrative' AND osm_id < 0 AND admin_level IN ('0','1','2','3','4');` while working on this - since in development you're constantly re-rendering low zooms instead of caching them for a month.

math1985:
Is there a reason to keep the #necountries layer and shapefile, or would we be able to get rid of this and render the low-level admin borders based on osm-data?

nebulon42:
I think it would be great to replace the Natural Earth data with OSM data. But this would need generalization and should be done with a shapefile since Douglas-Peucker (ST_Simplify) was not well-received. This would need some logic similar to coastline processing and could be hosted on openstreetmapdata.com (?), but someone would need to provide the update and error logic.

pnorman:
> think it would be great to replace the Natural Earth data with OSM data. But this would need generalization and should be done with a shapefile since Douglas-Peucker (ST_Simplify) was not well-received. This would need some logic similar to coastline processing and could be hosted on openstreetmapdata.com (?), but someone would need to provide the update and error logic

We don't use dashed lines at the even lower zooms, so the issues with dashes don't apply here. But we can handle this in a new issue.

nebulon42:
> We don't use dashed lines at the even lower zooms, so the issues with dashes don't apply here. But we can handle this in a new issue.

Yes, we should do that. But I think at least @imagico's concerns were not mainly about dashes.

math1985:
Is it just me, or is zoom level 4 not being properly updated with the new rendering on the openstreetmap.org servers? Is this a glitch on the servers or in the stylesheet?

gravitystorm:
I had to revert this pull request, since it was bringing the OSMF servers to a halt (see the revert PR #1987 ). This is the query that had been running for more than 14 hours for one request: https://github.com/gravitystorm/openstreetmap-carto/blob/7b8998c089b448dd73e54cd05c2321c99d529b95/project.yaml#L1322-L1342

I suspect this needs to be tested further with databases containing more representative data - it appears the query runs fine when there's a small amount of boundary-only data, but not when the tables contain lots of non-boundary data too. I think @pnorman is making some investigations regarding the query and the behaviour of the query planner.

imagico:
The shapefile layer changes which were positive on their own could be restored independently by the way.

pnorman:
One of the longest queries was

```sql
SELECT ST_AsBinary("way") AS geom,"admin_level" FROM (SELECT
    way,
    admin_level
  FROM planet_osm_roads o
  WHERE boundary = 'administrative'
    AND (admin_level IN ('0', '1', '2')
      OR (admin_level IN ('3', '4')
        AND EXISTS
          (SELECT 1
            FROM planet_osm_polygon i
            WHERE boundary = 'administrative'
              AND admin_level = '2'
              AND way_area > 9e+12
              AND ST_Within(o.way, i.way)
              AND osm_id < 0
          )
      )
    )
    AND osm_id < 0
  ORDER BY admin_level DESC
  ) AS admin_very_low_zoom WHERE "way" && ST_SetSRID('BOX3D(-1252344.271424999 -1252344.271424999,20037508 20037508)'::box3d, 900913);
```

This is the z4 query, which gets z2 admin or z4 admin in large z2.

The query plan on one of the servers was http://explain.depesz.com/s/VgNj

The query plan is moderately stupid, involving a bitmap and of a 20GB gist and 5GB btree index for the subplan, and isn't great for the roads part of the plan, though indexes are smaller there.

pnorman:
Running on my server with only standard indexes, I get

http://explain.depesz.com/s/laJE

30 minute query time isn't great, but there are only 4 z4 metatiles, and this is the big one that covers most of Europe.

The query on production with the different plan was running for >14h before terminated.

math1985:
Is that a server with full planet loaded?

pnorman:
Yes. My setup has a few differences

* Single-user, faster CPU, but much less RAM. Probably not significant
* PostgreSQL 9.4 instead of 9.1 (?). Possible cause
* No btree index on planet_osm_roads (osm_id). I do have one on planet_osm_polygon (osm_id)
* Data from this summer. Not a significant difference
* No slim tables. Doesn't matter
* No updates to the render tables. Probably not significant, though it does make it probably 25-75% faster without bloat

Since the query takes >14h to run on production, I'm not going to get any selectivity estimates to see **why** it's doing the BitmapAnd plan or if it's a good plan, but my instincts are telling me it's a horrible plan.

My inclination is to fix this through #207, because a suitable partial index brings query times down into the 1 minute range.

nebulon42:
Oh dear, I will provide a new PR with just the style updates as I think these are worthwhile to have in also without the new logic on z4.

Issue #1941, opened 2015-11-01
Slow carto layers


pnorman:
We've hit 30 seconds to compile the stylesheet with `carto`. This negatively impacts the development experience, since it means you have to wait much longer to see any changes you make

`carto` has a benchmark mode. With a bit of grep (`./node_modules/carto/bin/carto -b project.mml 2>&1 | egrep -B1 'Inheritance time: [[:digit:]]{3,}ms'`) and some manual cleanup, we can find the slow layers

```
text-point: 2294ms
text-line: 2049ms
text-poly: 2025ms
text-poly-low-zoom: 1551ms
bridges: 555ms
tunnels: 466ms
roads-fill: 450ms
roads-low-zoom: 189ms
landcover-low-zoom: 147ms
landcover: 138ms
```

My system is about twice as fast as Travis, so only takes 15s. These layers are 10 of those seconds.

math1985:
Great work for finding this out!

math1985:
I can confirm the explosion in render time. It's 20 seconds on my system.

This is all before touching any database, right?

The most surprising one I find text-line. The output of text-line in the XML is also extreme long, and that does not seem to make sense given the CartoCSS. That's something we must look at.

The good news is that there are also many layers that take 0 or 1 ms so that suggest improvement is possible.

math1985:
At least #1944 is an easy win...

math1985:
* Rendering time for text-point with all definitions: 3228ms
* Rendering time for text-point without landuse and POI definitions: 2ms
* Rendering time for text-point with landuse but without POI definitions: 201ms
* Rendering time for text-point without landuse but with POI definitions: 1844ms

This is likely similar for text-poly.

So the landuse statements seem to account for a large portion of the parse time.

pnorman:
`text-line` is completely gone from the list (big win) and other text layers are faster. Updating the list.

kocio-pl:
Do we consider this problem to be resolved?

In my setup compilation time of a current master branch takes almost 14 s:

```
Parsing time: 11ms
Parsing time: 4ms
Parsing time: 3ms
Parsing time: 27ms
Parsing time: 10ms
Parsing time: 3ms
Parsing time: 1ms
Parsing time: 62ms
Parsing time: 2ms
Parsing time: 10ms
Parsing time: 1ms
Parsing time: 4ms
Parsing time: 45ms
Parsing time: 1ms
Parsing time: 2ms
[...]

PR #1947, opened 2015-11-01
Change order of definitions to speed up carto processing

math1985:
This improves #1941. Carto processing is now overall about 20% faster on my system.


pnorman:
Code looks okay, but I haven't tested it.

pnorman:
The rule here seems to be that when overriding for a subset of features (e.g. larger ones), declare all overrides in the same selector. I think this also makes it easier to read.

PR #1948, opened 2015-11-01
Make residential and tertiary slightly narrower on z13

math1985:
* Make residential/unclassified 0.5 px narrower on z13
* Make tertiary 1 px narrower on z13

Before:
screenshot from 2015-11-01 22 02 52

After:
screenshot from 2015-11-01 22 02 51


pnorman:
Looks better on both my rural and urban test areas.

PR #1949, opened 2015-11-01
Increase size of small place names on z15 by one pixel

math1985:
This makes them visually stand out more from road/square names.


kocio-pl:
Could you provide before/after visual test for it?

math1985:
screenshot from 2015-11-02 00 26 25
screenshot from 2015-11-02 00 28 24

PR #1952, opened 2015-11-03
Osmic icon for shop=beverages

kocio-pl:
I think the Osmic version is better than mine (slightly edited bottle, so it's different than shop=alcohol), so I'd like to replace my draft from https://github.com/gravitystorm/openstreetmap-carto/pull/1596 with the work of @nebulon42. Before/after comparison:

beverages-z19
beverages-new


matkoniecz:
Why rendering of name is also affected?

kocio-pl:
I guess in reality it's not, because I took "before" image from #1596 just to show the bottle shape change. [Current](https://www.openstreetmap.org/node/460050254) label rendering looks like in "after" image for me.

Issue #1946, opened 2015-11-01
Distinguishing road types, especially Motorways, isn't obvious enough in the new colour scheme


KenBW2:
I take the point from that using the UK road colour scheme isn't _intuituve_, but it did a very good job of distinguishing between different types of road.

Particularly Motorways - there's a very big difference between a primary road and a motorway, and on long journeys it's far and away the preferred choice of road.

In the new colour scheme, both primary roads and motorways are near enough the same colour, and it's not obvious at a glance which is the motorway. See below for an example.

![digix5l](https://cloud.githubusercontent.com/assets/366540/10870441/f2bbc48a-80c2-11e5-94b7-6181e509a848.jpg)

I'd like to propose a new colour for motorways which reflects their importance relative to other road types.

math1985:
Is this an issue that plays for you at all zoom levels, or are there particular zoom levels where this is a problem?

KenBW2:
I'd say anything > Z12

![screenshot from 2015-11-01 18 39 41](https://cloud.githubusercontent.com/assets/366540/10870665/123e0214-80c8-11e5-9601-2f1c90cd7ae7.png)

![screenshot from 2015-11-01 18 39 20](https://cloud.githubusercontent.com/assets/366540/10870668/223a352a-80c8-11e5-8174-52cc719ef9f7.png)

![screenshot from 2015-11-01 18 38 57](https://cloud.githubusercontent.com/assets/366540/10870669/2a765be2-80c8-11e5-8601-d40c3826b9c5.png)

jojo4u:
Yes, on first glance it's difficult to make out the motorways. But perhaps we are too used of this read-centric view. The goal of osm-carto is not a map showing roads as prominent as e.g. Google Maps.

KenBW2:
Jack of all trades, master of none?

The original rationale of the change wasn't to make roads less prominent - it was because the UK colour scheme of Blue, Green, Red, Yellow wasn't considered intuitive. I personally don't like that that change was made, but I understand the reason for it.

But the change has become in my mind a regression. Where before motorways were easily distinguishable they're now almost invisible against the backdrop. I feel quite strongly that we find a colour that satisfies both requirements.

AlbertJP:
[...]

PR #1973, opened 2015-11-12
Made motorways more prominent. Fixes issue #1946

KenBW2:
As per issue #1946, motorways blend in with trunk roads far too much, which doesn't reflect their importance, and difference with other road types. This commit makes them more prominent, while still maintaining the principle of similar-hue-for-similar-way-types to ensure this remains an all-purpose map style.

Example with new motorway colours overlayed on existing map:

newmotorway

### DIFF
--- a/roads.mss
+++ b/roads.mss
@@ -1,7 +1,7 @@
 /* For the main linear features, such as roads and railways. */
 
 //road colors for major roads were generated with scripts/generate_road_colors.py
-@motorway-fill: #e892a2; // Lch(70,35,10), error 0.5
+@motorway-fill: #b45657; // Lch(70,35,10), error 0.5
 @trunk-fill: #f9b29c; // Lch(79,33,42), error 0.7
 @primary-fill: #fcd6a4; // Lch(88,31,74), error 1.7
 @secondary-fill: #f7fabf; // Lch(97,29,106), error 1.7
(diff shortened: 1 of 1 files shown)


pnorman:
These changes will be lost when rerunning [generate_road_colours](https://github.com/gravitystorm/openstreetmap-carto/blob/master/scripts/generate_road_colours.py)

KenBW2:
I've obviously misunderstood then. Where are the values set if not in the .mss files?

pnorman:
The values are generated by the script.

matkoniecz:
Maybe motorways may be stronger - but on example the difference seems to be too big.

This motorway is more visible than all other elements together, including city label.

jojo4u:
This is too strong, the jump from trunk to motorway is not justified.

math1985:
> Maybe motorways may be stronger - but on example the difference seems to be too big.


+1

SomeoneElseOSM:
Re "generate_road_colours" - is there anything written down about that anywhere?  Or even just a link to where it was created and included in the process?  Having that more prominent might have prevented someone wasting their time solving a problem in a way that "doesn't fit" with the current approach.

math1985:
The consensus is that this proposed change is not an improvement, so I'll reject this pull request. Thank you for the effort though @KenBW2!

PR #1983, opened 2015-11-23
improved rendering for peaks and saddles using shield symbolizers

nebulon42:
This improves peak rendering in mountainous areas by showing less peaks first but labeling them earlier. Implements also a more dynamic display of saddles when there is space. Labels are tied to symbols by using shield symbolizers. Labels are displayed either S, W or E of the symbol. Thanks to the sorting based on elevation that was already in important (higher) peaks should be displayed first. From z14 on all peaks also in areas with quite a lot of peaks (such as the one in the preview) should be displayed.

Drawbacks: We lose the ability to re-colour the SVG symbols in the code.

I have added the shield symbolizer code to the text layer but left the query for the symbol layer as it is.

A problem I just noticed: glacier labels seem to take precedence over peaks now.

Previews

z11
before_z11
after_z11

z12
before_z12
after_z12

z13
before_z13
after_z13

### DIFF
--- a/amenity-points.mss
+++ b/amenity-points.mss
@@ -951,27 +951,6 @@
     marker-fill: @transportation-icon;
   }
 
-  [feature = 'natural_peak'][zoom >= 11] {
-    marker-file: url('symbols/peak.svg');
-    marker-fill: @landform-color;
-    marker-placement: interior;
-    marker-clip: false;
-  }
-
-  [feature = 'natural_volcano'][zoom >= 11] {
-    marker-file: url('symbols/peak.svg');
-    marker-fill: #d40000;
-    marker-placement: interior;
-    marker-clip: false;
-  }
-
-  [feature = 'natural_saddle'][zoom >= 15] {
-    marker-file: url('symbols/saddle.svg');
-    marker-fill: @landform-color;
-    marker-placement: interior;
-    marker-clip: false;
-  }
-
   [feature = 'natural_cave_entrance'][zoom >= 15] {
     point-file: url('symbols/poi_cave.p.16.png');
     point-placement: interior;
@@ -1237,21 +1216,53 @@
     text-placement: interior;
   }
 
-  [feature = 'natural_peak'][zoom >= 13],
-  [feature = 'natural_volcano'][zoom >= 13],
-  [feature = 'natural_saddle'][zoom >= 15],
   [feature = 'tourism_viewpoint'][zoom >= 16] {
     text-name: "[name]";
     text-size: 10;
     text-fill: darken(@landform-color, 30%);
-    [feature = 'natural_volcano'] { text-fill: #d40000; }
-    text-dy: 7;
-    [feature = 'tourism_viewpoint'] { text-dy: 11; }
+    text-dy: 11;
     text-face-name: @book-fonts;
     text-halo-radius: 1;
     text-halo-fill: rgba(255,255,255,0.6);
     text-wrap-width: @standard-wrap-width;
     text-placement: interior;
+  }
+
+  [feature = 'natural_peak'][zoom >= 11],
+  [feature = 'natural_volcano'][zoom >= 11],
+  [feature = 'natural_saddle'][zoom >= 13] {
+    [feature = 'natural_peak'] {
+      shield-file: url('symbols/peak.svg');
+    }
+    [feature = 'natural_volcano'] {
+      shield-file: url('symbols/volcano.svg');
+    }
+    [feature = 'natural_saddle'] {
+      shield-file: url('symbols/saddle.svg');
+    }
+    shield-name: '[name]';
+    shield-face-name: @book-fonts;
+    shield-fill: darken(@landform-color, 30%);
+    [feature = 'natural_volcano'] {
+      shield-fill: #d40000;
+    }
+    shield-size: 9;
+    shield-text-dy: 7;
+    shield-text-dx: 8;
+
+    [zoom >= 13] {
+      shield-size: 10;
+      shield-text-dy: 8;
+      shield-text-dx: 9;
+    }
+    shield-min-distance: 50;
+    shield-wrap-width: @standard-wrap-width;
+    shield-halo-fill: rgba(255,255,255,0.6);
+    shield-halo-radius: 1;
+    shield-placement-type: simple;
+    shield-placements: 'S,E,W';
+    shield-unlock-image: true;
+    shield-min-padding: 1;
   }
 
   [feature = 'man_made_cross'][zoom >= 17],
--- a/symbols/saddle.svg
+++ b/symbols/saddle.svg
@@ -4,5 +4,5 @@
    version="1.1"
    width="8"
    height="8">
-	<path d="M 0,3 0,8 8,8 8,3 4,5 z" />
+	<path d="M 0,3 0,8 8,8 8,3 4,5 z" style="fill: #d08f55; fill-opacity: 1;" />
 </svg>
--- a/symbols/volcano.svg
+++ b/symbols/volcano.svg
@@ -4,5 +4,5 @@
    version="1.1"
    width="8"
    height="8">
-	<path d="M 4,1 0,8 8,8 z" style="fill: #d08f55; fill-opacity: 1;" />
+	<path d="M 4,1 0,8 8,8 z" style="fill: #d40000; fill-opacity: 1;" />
 </svg>


amenity-points.mss
```diff
@@ -1254,6 +1228,43 @@
...
+      shield-size: 9;
+      shield-text-dy: 7;
+      shield-text-dx: 8;
+
+      [zoom >= 13] {
+        shield-size: 10;
+        shield-text-dy: 8;
+        shield-text-dx: 9;
+      }
+      shield-min-distance: 50;
```
math1985: Wouldn't this also block highway shields near peaks, and peaks near highway shields?

amenity-points.mss
```diff
@@ -1254,6 +1228,43 @@
     text-placement: interior;
   }
 
+  [feature = 'natural_peak'][zoom >= 11],
+  [feature = 'natural_volcano'][zoom >= 11],
+  [feature = 'natural_saddle'][zoom >= 13] {
+      [feature = 'natural_peak'] {
```
math1985: Should be four instead of six leading spaces :)

amenity-points.mss
nebulon42: Maybe theoretically, I was not able to spot anything weird. But it seems that other labels are blocking peaks now more often.

nebulon42:
Fixed an intendation issue. Issue with labels blocking peaks has still to be investigated.

math1985:
I wonder: if there are two icons, one with a min-distance of 20px and one with a min-distance of 40px, that are 30px apart, would both render? It's really a shame that documentation to Mapnik is so often lacking.

matkoniecz:
Maybe with that change it is possible to show peaks also on z10? Maybe even earlier?

imagico:
I am not sure this is a good idea.  When space is scarce showing the peaks themselves is more important than showing the labels and prioritization via ele tag is not very good, ideally you'd prioritize by prominence.  In your z11 example hardly 10 percent of the peaks are displayed.  IMO this is only OK if you have a really reliable prioritization.

And i assume this kind of label placement will lead to more frequent artefacts at the metatile boundaries - or does Mapnik take special precautions for that?

In your example do i see correctly that Mitterkeil is shown at z11 and not at z12, Jocheck at z12 and not z13?  If this is not just an error in the samples this would not be good either.

symbols/peak.svg
```diff
@@ -4,5 +4,5 @@
    version="1.1"
    width="8"
    height="8">
-	<path d="M 4,1 0,8 8,8 z" />
+	<path d="M 4,1 0,8 8,8 z" style="fill: #d08f55; fill-opacity: 1;" />
```
pnorman: We should set the colour for these in the MSS, not the SVG.

pnorman: Or is that not possible with shield symbolizers? there's `shield-fill: #d40000;` in the code

pnorman:
> And i assume this kind of label placement will lead to more frequent artefacts at the metatile boundaries - or does Mapnik take special precautions for that?

Possibly, but if the buffer captures what is causing the rendering decisions, it'll be the same in both.

It's unfortunate we can't do a better ordering than `ele desc`. `osm_id` desc is probably a pretty good order too, but I wouldn't want to do that for policy reasons.

Our current rendering in high-peak areas is quite bad.

symbols/peak.svg
nebulon42: Unfortunately, that is not possible. The ideal solution would be to use a group symbolizer, but this is not yet supported in Carto and anyway Mapnik 3 only.

nebulon42:
> When space is scarce showing the peaks themselves is more important than showing the labels

For me the peak icon alone (in absence of contour lines or hill shading) carries little information. I would rather omit it than showing it without name or elevation.

> And i assume this kind of label placement will lead to more frequent artefacts at the metatile boundaries

There is `shield-min-padding` for that which I use, but if that's good enough I don't know.

> In your example do i see correctly that Mitterkeil is shown at z11 and not at z12, Jocheck at z12 and not z13?

Yes, correct. Not good, I agree.

nebulon42:
> I wonder: if there are two icons, one with a min-distance of 20px and one with a min-distance of 40px, that are 30px apart, would both render?

Interesting question. I don't know, but would assume that both would render if the 20px one would be rendered last. The other way round the 40px one would not render.

mboeringa:
Even with the best automatic algorithms, I really wonder if this could be "fixed":

http://www.openstreetmap.org/#map=15/50.9098/14.2336

Apparently, someone in Germany thought it a good idea to name every karst rock / shape as a peak, even though probably not a single one of it stands out as a true peak visible from a distance, but is merely visible from within the karst system itself, walking through its canyons or over its ridges.

Considering this changeset comment:
http://www.openstreetmap.org/node/461196746
which talks about a revert of turning "peaks" into "rocks", someone has attempted to turn these in a more sensible "rock" feature. However, others disagreed, as obviously this doesn't show up on the map as a feature not (yet) supported by this style... so reverted it.

More discussion seems here:
https://lists.openstreetmap.org/pipermail/talk-de/2015-October/111947.html

nebulon42:
> Maybe with that change it is possible to show peaks also on z10? Maybe even earlier?

Peaks are rendered quite prominently. I already decreased the font size for z11/z12. On lower zooms they maybe would stand out too much?

imagico:
> For me the peak icon alone (in absence of contour lines or hill shading) carries little information. I would rather omit it than showing it without name or elevation.

I strongly disagree.  The peak distribution tells a lot about the topography - especially in the absence of any other elements doing that and your change essentially makes every area look the same - no matter if it is a mountain area with ridges and valleys or a smooth hill landscape.

Note the current rendering at least shows peak icons even if there is not enough room for the label, your change only shows the icon if also the label can be shown.

> Possibly, but if the buffer captures what is causing the rendering decisions, it'll be the same in both.

But with the current styling the only cause for artefacts is showing the label in one tile and not in the other.  With the proposed change there would in addition be differences in the label position - unless Mapnik tries to avoid the tile boundaries when making this decision.

dieterdreist:
2015-11-24 8:23 GMT+01:00 mboeringa <notifications@github.com>:

> http://www.openstreetmap.org/#map=15/50.9098/14.2336
>
> Apparently, someone in Germany thought it a good idea to name every karst
> rock / shape as a peak, even though probably not a single one of it stands
> out as a true peak visible from a distance, but is merely visible from
> within the karst system itself, walking through its canyons or over its
> ridges.
>


well, that's a very famous area in Germany, looking like this:
https://de.wikipedia.org/wiki/Elbsandsteingebirge#/media/File:Basteibr%C3%BCcke_morgens_%28Zuschnitt%29.jpg
and as all of them seem to have names (i.e. are significant for the locals)
I guess the tagging could be considered OK.

There's no requirement for a natural=peak to stand out from a distance.

polarbearing:
> well, that's a very famous area in Germany

Thanks @dieterdreist and @imagico - not only do they have names; each individual peak, and there are thousands, is climbed on when freestanding (as opposed to a cliff wall that is walkable for the other side), and has a summit log book. It was clarified in the tagging discussion that natural=peak is immaterial and maps the highest point of a natural or renaturalised physical feature.
It is insofar different from natural=rock, which maps a material feature. 

To conclude I support rendering peak symbols in priority to their labels.

nebulon42:
> The peak distribution tells a lot about the topography

I think you are overestimating the utility of peaks a bit. If you look at these two areas https://www.openstreetmap.org/#map=11/47.8837/15.9806 and https://www.openstreetmap.org/#map=11/46.9474/13.3645 one is hilly and one has mountains. What tells me more about the topography is the structure/absence of forest, rivers/streams and roads. Peaks help also, but here you can say at most "there are a lot of peaks". I think that the main structure is still deducible after the change. Maybe a bit less spacing will also help.

imagico:
> Peaks help also, but here you can say at most "there are a lot of peaks". I think that the main structure is still deducible after the change. 

No, when peaks are well mapped and you display them fully you can often well recognize the major structure of the terrain through their spatial distribution and arrangement, especially in glacially formed mountain areas like in the Alps.  Examples:

http://mc.bbbike.org/mc/?lon=11.891585&lat=47.037743&zoom=11&num=2&mt0=nokia-terrain&mt1=mapnik
http://mc.bbbike.org/mc/?lon=7.530699&lat=46.595575&zoom=11&num=2&mt0=nokia-terrain&mt1=mapnik

With your change none of this structural information is visible any more.  And since streams are not shown before z13 they can't help.

IMO it is best not to regard peaks as individual POIs at the lower zooms but as a set of features which is mostly meaningful as a whole.

I would not mind showing labels for the major peaks already at z11/12 even if that meant not showing those peaks the labels would interfere with but then again this would require a meaningful way to rate the peaks and elevation is not well suited for that.

It is of course a perfectly legitimate idea to not have any information about the topography shown in the map and to indeed show peaks only as individual named points of interest.  But the generally wide interest in both mapping topographic features (peak/saddle but also ridge/cliff) as well as having them rendered indicates mappers are quite interested in this kind of information.

And of course current rendering at z11/12 is not that great, at least at z11 the icon should probably be smaller to be less disturbing when there are lots of them.  At the same time a slight bright halo might help visibility on boundaries.

math1985:
In a now hidden comment (on an outdated diff), I wondered whether this also block highway shields near peaks, and peaks near highway shields. I think we should try to find this out first before we can merge this (if we want to merge it at all).

math1985:
> I wondered whether this also block highway shields near peaks, and peaks near highway shields

I still suspect this is in fact the case (from trying to understand the code, not from testing). I'll therefore close this PR. If somebody could show me that highway shields are not blocked, feel free to show a demo and I'll reopen it!

nebulon42:
At this moment I don't plan to work on this further.

PR #1989, opened 2015-12-01
rework admin boundaries (without z4 logic)

nebulon42:
This recreates the styling of #1938 without the problematic queries.
It would be grand if there could be a .2 release including these changes. Sorry for the problems this has caused.

### DIFF
--- a/admin.mss
+++ b/admin.mss
@@ -9,45 +9,86 @@ overlapping borders correctly.
 #admin-low-zoom[zoom < 11],
 #admin-mid-zoom[zoom >= 11][zoom < 13],
 #admin-high-zoom[zoom >= 13] {
-  [admin_level = '2'],
-  [admin_level = '3'] {
+  [admin_level = '2'] {
     [zoom >= 4] {
+      background/line-join: bevel;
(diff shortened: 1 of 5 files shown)


math1985:
Rolling out a release has the disadvantage that updating tiles gets slower for mappers, so I'd prefer not to release too often.

> Sorry for the problems this has caused.

Don't worry about that, it's not something you could have tested, as apparently it depends on very specific configuration options.

nebulon42:
It has been quite some time. I would love to see a new release with this merged.

math1985:
Thanks! Looks much better again. Sorry for the long wait!

kocio-pl:
Thanks for this work and for merging it - it is great and fixes some of the most important problems I saw with low zoom on OSM!

sommerluk:
Great work! Thanks!

daganzdaanda:
Big thanks from me, too! The map makes a lot more sense now!

mboeringa:
:+1:

SomeoneElseOSM:
Excellent stuff - thanks very much.

Issue #1982, opened 2015-11-21
Switch to ocean polygons


pnorman:
We should switch from preprocessed land polygons on a water colour background to preprocessed water polygons on a land colour background.

This would
- Allow easier customization with hillshading
- Be quicker in inland areas
- Allow more water styling, including outlines of water
- Allow rendering on top of maritime borders, making it possible to have them less prominent
- Makes it easier to render inland areas with no coastline by substituting an empty shapefile

Cross-ref #621 

Issues
- There are no simplified water polygons

cc @joto @imagico

math1985:
Sounds reasonable. What do you mean with water styling?

imagico:
Other advantages would be:

* consistent wetland rendering, see #1547
* rendering landcovers differently over water and over land (like natural=beach)
* allow to differentiate ocean from inland water in rendering.

Simplified water polygons have meanwhile been set up on openstreetmapdata.com - we were planning to roll this out together some other improvements in coastline processing which are not quite finished yet but you can already get the simplified water polygons on

http://data.openstreetmapdata.com/simplified-water-polygons-complete-3857.zip

pnorman:
> Sounds reasonable. What did you have in mind with water styling / outlines of water?

I want to keep that separate, but a fairly common convention is a darker blue outline.

donaldrnoble:
would doing this stop the issue of broken coastlines "flooding" the land? or could the reverse issue then happen, which areas of sea being shown as land?

gravitystorm:
@donaldrnoble It makes no difference, "flooding" is still possible.

math1985:
I reverted the PR that fixed this, see #2101.

jeisenbe:
The benefits of this change still remain important, as listed above:
* Allows easy customization with hillshading
* Faster rendering in inland areas
* Makes it possible to render inland areas with no coastline by substituting an empty shapefile
* Improved water styling options
* Allow rendering oceans on top of maritime borders, making it possible to have them less prominent
* Allows consistent wetland rendering near / over water, e.g. mangroves, tidal flats, and salt marshes
* Allows rendering landcovers differently over water and over land (like natural=beach, =shoal)
* Possible to differentiate ocean from inland water, and rivers from other water bodies, in rendering 

Makes it possible to solve these issues: #621, #788, #1547, #1781, #2013, #2025, #2609, #3273, and PR #3670

[...]

PR #1990, opened 2015-12-03
Adding rendering of shoal and reef, improving rendering of beach

imagico:
This makes the following changes to landcover rendering:

`natural=reef` (8000 uses) is rendered with a random dot overlay pattern in dark blue.

`natural=shoal` (~200 uses) is rendered exactly like beach (see below).  This is meant to counteract the widespread use of natural=beach to tag shoals.  Although beaches and shoals are quite different things they are related - many shoals are essentially small islands with beaches on all sides and not much else.

For `natural=beach` the old regular grid-dot pattern is removed.  Beaches with surface tag (14k uses) are rendered with a pattern at z>=14 depending on the surface value.  `surface=sand` is rendered with a fine grained pattern while `surface=gravel|fine_gravel|pebbles|pebblestone|shingle|stones|shells` is rendered with a coarser pattern (the same one as for scree/shingle).  `surface=rocky` (180 uses in combination with natural=beach) is ignored since this is abuse of natural=beach (rocky coasts are no beaches).  Beaches with no or no known surface tag are rendered in plain color.

The patterns are rendered separately in the landcover overlay layer so after moving to water polygons (#1982) the pattern can be rendered above water while the base color is below water.  This should lead to intuitive rendering of parts of beaches above and below the water line.

Fixes part of #545.

Examples:

Reef:
reef

Shoal:
shoal

Beach with surface=sand:
surface=sand

Beach with surface=shingle (left) and surface=sand (right):
surface=shingle

Beach next to natural=sand:
beach+sand


mboeringa:
Love the subtle renderings for surfaces of the beach. Like to see something more subtle for the wetland too, it really stands out to much now compared to all other stuff, as clearly visible in these images.

kocio-pl:
While I like having more features available in this style, I have to admit I'm not able to guess without additional hints (like reading this proposition or a map key) what the reef rendering represents. Worse, on the first picture there is another water related feature, which I, in fact, don't know at all and can't recognize myself. This was already discussed with similar pattern ([landuse=aquaculture](https://github.com/gravitystorm/openstreetmap-carto/pull/1817)), which was rejected lately because of this.

Additional problem for me is that I already can't recognize just by looking at the map the beach from the sand and shingle would be another sand-related pattern with no exact meaning to be understood visually.

mboeringa:
@kocio-pl , at some point (and this style is already way past it), a map needs a legend...

There will always be ambiguous symbols that need an explanation, in fact, ***all*** symbols are more or less affected by this, even the "toilet" symbol may not be of any use to someone who has lived his entire life as part of a native tribe in the rainforests of South America or New Guinea...

Our perception of symbols is pretty much a cultural thing. Best to make a legend a get the ambiguity out of the door by explanatory text.

imagico:
Regarding recognizability - i don't really see the problem here, certainly there are other ways you can render reefs in maps but a random dot pattern is not unreasonable considering data usually does not give further information regarding the nature of the reef (coral, rock, loose material).  And outline only rendering (which is common for reefs in other maps) would also be generally difficult considering the common practice in OSM to have arbitrary splits in area mapping.   Keeping it in a blue tone should give an indication it is an underwater feature.

Note the reef rendering example is not a typical reef - i was just looking for reefs in my database and this were the only ones found - more typical cases would be here:

http://overpass-turbo.eu/s/d5N

And of course recognizability is always a matter of how familiar the map reader is with the matter.

Regarding beach color - changing this would be  a separate matter.  I would be open to that although i don't see anything terribly wrong with the current one.  Beaches are well defined and usually fairly compact features and also often quite significant for map users so a fairly strong tone fits quite well IMO.  The biggest problem i see here is similarity to parking.

And wetland rendering should be discussed separately i think.

imagico:
Changed the reef example to something more representative and added an example of beach with sand pattern next to natural=sand.

polarbearing:
I appreciate the appearance of the reef, and the upgrade of the grid beach to a random pattern.

 I am sceptical however in encoding too many surface values into subtle differences, that are hard to recognise when not contrasted to each other. Even with a legend that becomes tricky, we would need more like a guidebook like how to interpret yellow shading in van Gogh's sunflower's eventually...

imagico:
I understand this.  There main reason for suggesting it is because the surface type is generally considered an important property of a beach and while this is widely tagged (14k cases) most beaches are still without, even in areas well mapped by locals who surely have knowledge of the properties of the beach.  So it is less important to clearly see the difference between the different surface types and more important to see the differences between beaches with surface tagged and those without.

My opinion is as long as they are not misleading use of subtle differences in appearence to indicate minor differences in tagging in ways that might be lost to the casual viewer but that can be recognized by the experienced user or with a map key is a reasonable way to extend and improve an already fairly loaded map style like this.

But i am not tied to this - if it is considered too much i can also move to a unified pattern as before.

kocio-pl:
For me adding these patterns is not evil - I also think it's impossible to know everything without looking at key. I just reported that here is something I can't recognize at all myself (only that it's water or sand related) and I'm interested if there is any difference for @pnorman ("It needs to be possible to gain an understanding of the map without a legend") and @math1985 ("I agree the proposed rendering isn't really clear, and map users also won't be able to learn easily what it means."), because I think this is the same case as rejected landuse=aquaculture pattern.

imagico:
@kocio-pl - if i understand you correctly your critique in mainly the difficulty to intuitively recognize the reef rendering as what it is on your part, not that you would distinctly misinterpret it as something different.

It would be interesting to know if others have the same impression, in particular people who routinely use the maps in maritime contexts.

Regarding landuse=aquaculture - which is off-topic here but i would like to mention this none the less - please keep in mind your PR for that was not closed because the very idea of rendering it was rejected but because the specific proposal as you submitted it was generally not considered favorably.  This was partly because of the difficulty to recognize what it means (including the likely misinterpretation as some sort of restricted area) but also partly due to the somewhat unbalanced design with outline and pattern which contributed to the difficult recognition but was also a problem on its own.

kocio-pl:
Yes, you're perfectly right: they are just too abstract for me, proposed rendering doesn't suggest me some other objects. If they follow some nautical map conventions to a certain degree, it would be of course much better.

Maybe you are also right and this case is different enough from landuse=aquaculture, but currently I'm not convinced about it - and still I'm curious what is our position on lack of immediate recognizability, because it's unpredictable for me and I find it confusing.

talllguy:
I just mapped a reef that would be a good candidate for this type of proposed rendering. I think this would add value to the map for mappers.

screenshot 2015-12-12 17 22 42
http://www.openstreetmap.org/relation/5737665#map=14/-11.6340/166.9737

rrzefox:
I like the reed rendering, at the very least it's nice eyecandy. The one from tallguys post looks like this: https://osm.rrze.fau.de/map-ll-osm?zoom=14&lat=-11.6340&lon=166.9737&layer=RRZE%20tileserver%20osm.org%20style%20tiles
(this in theory has worldwide coverage, but might still have some tiles cached without this PR applied)

imagico:
Thanks, this is very useful.

A common mapping error at the moment with reefs seems to be that land areas are included - like here:

https://osm.rrze.fau.de/map-ll-osm?zoom=13&lat=-1.8672&lon=175.5615&layer=RRZE%20tileserver%20osm.org%20style%20tiles

This change would also help pointing this out.

A few examples for the beach rendering (with surface=sand pattern):

https://osm.rrze.fau.de/map-ll-osm?zoom=15&lat=39.52589&lon=2.45707&layer=RRZE%20tileserver%20osm.org%20style%20tiles
https://osm.rrze.fau.de/map-ll-osm?zoom=15&lat=41.11363&lon=-8.66474&layer=RRZE%20tileserver%20osm.org%20style%20tiles
https://osm.rrze.fau.de/map-ll-osm?zoom=14&lat=48.66381&lon=-2.00063&layer=RRZE%20tileserver%20osm.org%20style%20tiles

morray:
great. I bet it will lead to a lot of coastline refinement and thus demands also a higher update rate for it. It things like here will happen:
https://osm.rrze.fau.de/map-ll-osm?zoom=14&lat=-11.6340&lon=166.9737&layer=RRZE%20tileserver%20osm.org%20style%20tiles

morray:
Ash https://osm.rrze.fau.de/map-ll-osm?zoom=15&lat=-17.15&lon=177.2&layer=RRZE%20tileserver%20osm.org%20style%20tiles

imagico:
By the way to get an idea on how beaches extending beyond the coastline would look like with water polygons here two examples (where water is mapped as natural=water so rendered above the landcover layer):

https://osm.rrze.fau.de/map-ll-osm?zoom=15&lat=50.78436&lon=-0.91723&layer=RRZE%20tileserver%20osm.org%20style%20tiles

https://osm.rrze.fau.de/map-ll-osm?zoom=17&lat=50.786186&lon=-1.03992&layer=RRZE%20tileserver%20osm.org%20style%20tiles

 (mapping there is not that great but you should get the idea).

math1985:
Thanks!

talllguy:
@math1985 awesome, when will this begin to appear on the map?

math1985:
Normally, every couple of weeks, we make a release. See here for the past ones: https://github.com/gravitystorm/openstreetmap-carto/releases Due to lack of activity, we didn't release that often lately. I think we're due for a new release soon.

Once the release is made, we ask the maintainers of the tile server to upgrade. Usually they do so within a couple of days.

Then it still takes one or two days before all/most tiles are updated.

imagico:
Ideally there should also be an updated coastline for the next style update. @joto

PR #1873, opened 2015-09-25
separate setting and processing data in generate_road_colours.py

matkoniecz:
It intends to make this script more readable and to make value tweaking easier.

PR #2000, opened 2015-12-08
add possibility of specifying hsl colours to shield generation script

nebulon42:
I wrote a small Python method to parse RGB e.g. rgb(150, 120, 130) or HSL e.g hsl(210, 20%, 50%) colour definitions. I think the shield generation script would benefit from it, since it makes playing with colours a lot easier.

If you don't think so too, please close this pull request.

P.S.: PR 2000, wow. :)

### DIFF
--- a/scripts/generate_shields.py
+++ b/scripts/generate_shields.py
@@ -3,7 +3,7 @@
 # generate highway shields
 
 from __future__ import print_function
-import copy, lxml.etree, math, os
+import copy, lxml.etree, math, os, re, colorsys
 
 def main():
 
@@ -108,9 +108,9 @@ def main():
 
           stroke = ''
           if vars['stroke_width'] > 0:
-            stroke = 'stroke:' + vars['stroke_fill'] + ';stroke-width:' + str(vars['stroke_width']) + ';'
+            stroke = 'stroke:' + parseColor(vars['stroke_fill']) + ';stroke-width:' + str(vars['stroke_width']) + ';'
 
-          shield.set('style', 'fill:' + vars['fill'] + ';' + stroke)
+          shield.set('style', 'fill:' + parseColor(vars['fill']) + ';' + stroke)
 
           svg.append(shield)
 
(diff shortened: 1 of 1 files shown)


pnorman:
The output of preprocessed code shouldn't really be edited by hand, so I don't see hsl or rgb has an advantage over hex.

nebulon42:
Others might want to use the script in a different way and I found hsl handy when experimenting. But as I said it's just a suggestion and I don't mind if you don't see the added value.

pnorman:
if this PR is changing shield output, it should also contain changes to all the shields, right?

nebulon42:
Currently it does not change anything visually as any hex value will be returned without changes, see https://github.com/gravitystorm/openstreetmap-carto/pull/2000/files#diff-d6c0cb01945928a60a33f9441ed95cdeR136
Additionally, rgb and hsl are always converted to hex in the SVG.

pnorman:
It's not a question of visual changes, it's a question of the files being up to date. If I run the script and copy/paste the output into the relevant files, `git diff` should show no changes.

nebulon42:
> If I run the script and copy/paste the output into the relevant files, git diff should show no changes.

This is the case.

pnorman:
To the script stroke_fill and such are just strings, and svg has HSL support

Can't you just do `config['motorway']['fill'] = 'hsl(0,100%, 50%'`?

nebulon42:
>  svg has HSL support

SVG 2 supports HSL, SVG 1.1 only supports RGB and Hex notation. I'm not sure how widespread SVG 2 support is, but I've tried this before and neither my built-in image preview program nor Inkscape where able to display this correctly. I doubt that Mapnik supports SVG 2, but I might be wrong. A quick search didn't bring something up.

pnorman:
> SVG 2 supports HSL, SVG 1.1 only supports RGB and Hex notation

Good point.

I'll have a closer look at the code, also I'm not sure how this best interacts with #1873

nebulon42:
I think I will withdraw this for now, because I want to enhance it with Lch. The function can always be found at https://github.com/gmgeo/osmic/blob/master/tools/export.py#L532-L558.

Another good thing would be to bring Lch to carto. @pnorman if you need help, let me know.

math1985:
> Another good thing would be to bring Lch to carto

+1, is it supported in Mapnik?

pnorman:
> > Another good thing would be to bring Lch to carto

> +1, is it supported in Mapnik?

No, and doesn't need to be.

Carto takes colour functions and turns them into hex codes or rgba.

This is good because it lets us use new functions without a Mapnik upgrade, which is harder.

I'm working on some fairly complex carto changes for work, so I'm working on small changes to get familiar with the carto changes, and one of them is perceptual colourspaces (mapbox/carto#354)

PR #2042, opened 2016-02-10
Move shelter icon to SVG

kocio-pl:
Implements part of #1165.

Simple vector replacement of a raster shelter icon:

pd1siftw


pnorman:
We use `name.size.ext` more than `name-size.ext`, so let's move to that at the same time.

kocio-pl:
IIRC all my icons was made like this - probably looking at some @nebulon42 files, I guess. I was never sure, because we have both formats mixed, but I had more serious problems, so never touched it. If we decide to make them all the same, it probably deserves its own cosmetic patch.

kocio-pl:
Just a cosmetic change so the icon looks better when enlarged (probably not visible on osm-carto):
oixnkvog

nebulon42:
While you're at it, I would suggest removing the rain as it just adds unnecessary clutter IMO.

imagico:
I think the rain is important to indicate the function, otherwise it is just some kind of building.

HolgerJeromin:
I would keep the rain, too.

kocio-pl:
I was thinking about it too and came to the conclusion that otherwise it's not clear what type of building/roof is this. I like it more with rain.

pnorman:
I like the rain, even though the clarity isn't perfect, mis-interpreting it is still likely to get the idea of sheltering across.

Issue #1624, opened 2015-06-29
Touristic/cultural destinations and accommodation on non-buildings should have a homogeneous expression (meta-issue)


polarbearing:
I found that touristic+cultural destinations as well as touristic accommodation, when mapped as an area, and not being a building, are represented quite heterogeneously. However it would make the map more readable if they came with the same visual language.

This analysis focuses on 'dedicated targets', from museums to theme_parks; as opposed to daily-use and pitch-like ones (park, pitch, sports_centre, playground, recreation_ground, golf_course...)

Here is some code review how they are represented currently. 
(table might scroll horizontally)

Key | area (colour) | label (colour) | icon (colour) | issues | comment
--- | --- | --- | --- | --- | ---
*Destinations* | 
tourism=museum | *none* | fixed (amenity-brown) | museum.16.svg (amenity-brown) | | (museum parks, open-air exhibitions) 
tourism=zoo | boundary (tourism) | scales (tourism) | *none* | 
tourism=attraction | *removed* (was pink) | scales (#660033) | *none*  | #1257 | (often used for specific areas within a larger facility, or on its own) 
tourism=theme_park | boundary (tourism) | scales (tourism) | *none* | 
leisure=nature_reserve, boundary=national_park | admin-boundary (green) | scales (dark park) + boundary (green) | *none*  | #603
leisure=marina | boundary (blue) | fixed (marina-text) | 
leisure=water_park | *none* | fixed (dark park) | water_park.16.svg (amenity-brown) | #1490
historic=archaeological_site | *none* | *none* | archaeological_site.16.svg (amenity-brown) | #732 
leisure=beach_resort | *none* 
*Accommodation* | 
camp_site | fill (campsite) | fixed (#0066ff) | camping.16.svg (transportation-icon) | #1485 #1020 #1355
caravan_site | fill (campsite) | scalable (dark campsite) | caravan_park.16.svg (transportation-icon)
hostel | *none* | fixed (#0066ff) | hostel.16.svg (transportation-icon) || (area typically summer-camp style bungalow villages)
hotel  |  *none* | fixed (#0066ff) | hotel.16.svg (transportation-icon) | #775 | (resorts)
motel |  *none* | fixed (#0066ff) | motel.16.svg (transportation-icon) || (often arranged as area with separate buildings)
resort | *none* |||| (only 312 uses, probably hotel more popular)
summer_camp | *none* |||| (only 100 uses)

[...]

PR #2047, opened 2016-02-16
Replacing bus station PNG icon with SVG version

kocio-pl:
Implements part of #1165.

Design according to discussion in https://github.com/gravitystorm/openstreetmap-carto/pull/1478 (some bus stops around to compare):
tpjsb51q


polarbearing:
Nice, clear and lightweight.

Just in comparison, the taxi icon appears much more massive, larger and darker.

imagico:
This is somewhat prone to be confused with two bus stops close to each other.

Why not use the old concept, i.e. inverted color bus in a box?

kocio-pl:
@imagico You're right, but since @nebulon42 didn't like it (see https://github.com/gravitystorm/openstreetmap-carto/pull/1478#issuecomment-99356259), I have tried something lighter, because mine and @daganzdaanda proposition was not ruled out (see https://github.com/gravitystorm/openstreetmap-carto/pull/1478#issuecomment-99530423).

kocio-pl:
We can also use some different designs with roof, as it was not tested yet (only proposed with rough sketch):
0xhupcf
oh6dreln
x3o41_ci

kocio-pl:
Inversion test:
4rt_w0mq
ovstuyjp

For me it works - we don't have to be such versatile as @nebulon42 tries with osmic.

We could also make the bus station letters bigger than bus stop or just use bold.

matkoniecz:
I like sloped roof and double bus - but it is based solely on this preview, without proper testing.

kocio-pl:
I will definitely make more rendering tests - once we know what looks promising and what we don't like for some reasons.

math1985:
Cool idea. I like the double bus.

Could you give a preview of Wilanowska with double bus? I'm not sure if it'd also work well if the individual bus stops in the bus station are tagged.

nebulon42:
I have to admit that the inverse version also doesn't look bad. For me at least the versions with the roof are possibly confused with bus garages or the like.

Slightly OT: The main problem for me (really emphasised in the last preview) is still the absence of icon halos. I still can't really understand why they were rejected, this decision solely based on them taking up more space. If they are necessary for text readability, they should be also necessary for icons.

pnorman:
Of the options, I prefer image the most so far.

If we go for an inversion like image I think we should have the bus in solid white.

gravitystorm:
I dislike the roof - these are bus stations, not bus garages, and many bus stations don't have roofs.

math1985:
> I dislike the roof - these are bus stations, not bus garages, and many bus stations don't have roofs.

+1

kocio-pl:
@math1985 For me the difference between both types is too small:
kprrqpd2
vqxx0oiq

kocio-pl:
@pnorman Is it technically possible with osm-carto? I think it's not. At least when trying to make canvas white, it was treated like part of the icon and I got 14px blue square - but maybe I made some mistake:
owztqzwd

kocio-pl:
So it looks like roofs are in defeat and inversion gains some approval by everybody. With font +1 and bold it looks OK for me, but still kind of light (!) on lower zoom levels:
z17:
acx9vno
z18:
nxbopy3l

polarbearing:
No roof please, and definitely no side walls (for the garage effect). Having seen the platform context I agree that the inverted is better than the double bus. If technically possible to avoid the transparency it would add clarity when the background is detailed.

kocio-pl:
No roofs and walls and no double buses (sadly it didn't work as good as I expected) - that's for sure. =} Now we're trying to polish inverted version. 

Here is design resembling more bus station icon from osmic, because now it has more solid blue and less opaque area, which causes problems on busy backgrounds:

Wilanowska, z17
dajjvhu6
Wilanowska, z18
rbsx4hni
Marymont, z17
kfb_vfmx
Marymont, z18
kv5kwjrj

I think it works quite good, so I updated this PR. 

If somebody know if we can change transparent area into white, let me know (I think we can't). 

I'd also like to know if we should try to push bus station to be rendered on top of bus stop icons (and how)?

nebulon42:
Of course you can just use a white shape over a blue rectangle. Then you would lose the ability to re-colour the icon in CartoCSS, though. But not a big deal.

pnorman:
> Then you would lose the ability to re-colour the icon in CartoCSS, though.

Yes, I think we're going to end up losing that for any multi-colour icon.

kocio-pl:
Thanks, I totally forgot that we don't need to control colors with styling options!

So which one bus shape we want to be used? This from a bus stop (a) or closer to the bus station from osmic (b)? I like b more, because it's more distinct and more consistent with osmic. I may rework them a bit, because inverted objects look a bit different (some details looks thicker, while the others start looking smaller), but the basic shape should be one of these (never mind the label on a):

a)
4rt_w0mq
b)
oxjzdk9l

matkoniecz:
In b label seems to be more on the left than icon. Is it an optical illusion or real effect?

dieterdreist:
I prefer a) because it is better recognizable as a bus, and it is more
consistent with the bus stops.

polarbearing:
> Is it an optical illusion or real effect?

I think it is real, but it might be a manual placement in the mockup?

Also in preference of the same symbol in bus stop and station, and a) looks more like a modern bus.

kocio-pl:
I also see this label shift, but I don't know where is it coming from - for me it was just another Kosmtik screenshot. But looks like all 3 bold+bigger font labels have this defect:

https://github.com/gravitystorm/openstreetmap-carto/pull/2047#issuecomment-184937329

mboeringa:
The distance between icon and label also seems to be to small compared to the non-bold version. If all the letters would be capital, the icon would hit the label.

kocio-pl:
I haven't touched the dy while looking for a general solution, but of course I will fine tune it now I know text labels for bus stations are better when enlarged.

However there's another interesting problem I'd like to report: the icon uses transportation-icon color (#0092da), while text uses transportation-text color (#0066ff). We have it also for bus stops for example, but it's much less visible there. Do we really need two distinct colors for transportation related items? And what was the reason they were introduced, BTW?

polarbearing:
My guess is that the different colours are from a time where most of them were hardcoded and picked randomly in previous styles, while I have seen more hue analysis discussed here recently. #1624 had a comparison of some colour use. Since there is no area filled with the transportation colour, I think icon and label should be unified.

kocio-pl:
There can be also c) - directly rebuild from osmic version - and I like it the most, because it's easier to recognize, since the white shape is bold:
2qownp2k

gormo-osm:
I like the double bus in the very first version, the rounded inverted bus definately looks like a subway/tram/train to me. Thats also true for the other inverted icon suggestions.

kocio-pl:
Unfortunately, it was already shown it's not working among bust stops:
https://github.com/gravitystorm/openstreetmap-carto/pull/2047#issuecomment-184833165

Inverted version seems to be good enough and is accepted by most participants.

matkoniecz:
> Unfortunately, it was already shown it's not working among bust stops:

I am not convinced that it is not working. This location makes clear what it is even without any bus station icon. How it looks at z16 where bus station icon is really important to make meaning clear?

pnorman:
The shape of the inverted bus seems substantially different

kocio-pl:
@matkoniecz It's better, since on z16 bus stops are rendered as squares, but still mix with them if they are too close (1) and are much less visible than metro or railway halt and stations (1 and 2, 3 is kind of special):
1)
f3hbperl
2)
fxwtu2dg
3)
fqj8nqk7

Maybe we should start rendering bus station name earlier, like railway stations for example, but it should be on higher layer probably to be visible.

matkoniecz:
After tests - I prefer current inverted icon over new SVG inverted icon.

kocio-pl:
OK, so I will try to simply recreate it instead of using osmic icons as a base.

math1985:
Closing this for now as it's not mergeable in its current state. Feel free to update the PR (or create a new one) and I'll reopen it again!

kocio-pl:
OK, I just got stuck a bit lately.

math1985:
No need to explain yourself, I was just in a mood for tidying up :)

PR #2048, opened 2016-02-16
Replacing cave entry PNG icon with SVG version

kocio-pl:
Implements part of #1165.

Simple substitution of PNG, since I think it works good already:
95msry1s


nebulon42:
Is it? Looks more symmetrical, but good.

kocio-pl:
Well, I just recreated it with circles because from what I saw it was more or less circular, but didn't make any pixel peeping. =} I mean fast design copy, not exact shape.

kocio-pl:
Looks like it's not about symmetry, but rather stretched up, so the other possible version could be something like this:
erdv96oi

kocio-pl:
Oh, it's not that simple - under further inspection I think it is side-stretched, but at the same time overall "triangular" shape. I will try to recreate it later to see if it's better than shape based on a regular circle.

HolgerJeromin:
Is this no Sjjb icon where you can get the source? 
But I am not sure if your current circle version is just as good as the original.

kocio-pl:
I guess this is much better:
hbozgipz

daganzdaanda:
The second one tries to be naturalistic, but we usually have our icons clean and more abstract. So I favour the first one.

Issue #324, opened 2014-01-30
Information Boards (tourism=information) on levels <=18


bitnapper:
Information boards are much to present on the map. In my region I mapped plenty smaller boards with historic information about buildings or other structures but imho it doesnt make sense rendering them before zoom level 18 or 19 or maybe ignoring them on the default map.

For example see this area: http://www.openstreetmap.org/#map=16/51.7123/10.5129

PR #2049, opened 2016-02-16
Replacing tourism information PNG icon with SVG version

kocio-pl:
Implements part of #1165.

I guess we should no longer wait for a database reload, because it is not going to be anytime soon, and instead just start with what's available now - here is just an osmic version of tourism information:

ccpf9yli

BTW: the name of such POI is not rendered - maybe we should start showing it?


nebulon42:
That icon was intended for `tourism=information` `information=office`. Maybe it would be better to use a lighter version of i since it may be too heavy in areas where lots of guideposts are mapped.

kocio-pl:
+1, I will try to make it thinner.

kocio-pl:
Made thinner a bit:
5wdtt7rx

daganzdaanda:
I like it - though it's still stronger than the current png and may still be a bit much for cities with lots of signposts. But fine-tuning can be done when information=* is available.

matkoniecz:
Can you check how it would work in http://www.openstreetmap.org/#map=17/51.71219/10.51618 and other areas mentioned in #324?

kocio-pl:
Looks like it's still too bold when it's more than one, so I'll try to make it even smaller:
pydr4zju

kocio-pl:
Now it should be better (with icon resized to 12 px as some other small POIs):
kvbaukam

matkoniecz:
Initial version would be great for information=office, https://github.com/gravitystorm/openstreetmap-carto/pull/2049#issuecomment-188778969 will be good for now. Thanks for SVGfication!

PR #2050, opened 2016-02-16
Replacing library PNG icon with SVG version

kocio-pl:
Implements part of #1165.

Follow up to #1228, which has been declined. This is a conservative approach - few details were added, but they are visible only when the icon is scaled up 3x/4x:

dxym2ojf


imagico:
Significantly less recognizable as a book than the original PNG version IMO.

kocio-pl:
This might be mostly because original icon is 20px large, while we try to make standard SVG icon only 14px (or smaller). See the before/after comparizon:
xbjx5gbk
dxym2ojf

imagico:
No, it's mostly the thicker book that makes the difference, being able to clearly distinguish between spine and head is important, otherwise it looks like a generic box.

kocio-pl:
The bigger matrix, the thicker book. Look at this 14px icon (I made it even more thick): 
o_cp6_xo
and the same image, but scaled up to 20px:
20px

What do you think about it?

kocio-pl:
Small modification on spine makes it even better, however I lack some skills with arcs in perspective, so only a PNG preview:
library 14-12

nebulon42:
I appreciate your efforts, but I'm not much into that kind of perspective regarding icons. Besides, I invested considerable time in other proposals, which got rejected. So I'm not working on this again. But I don't see any problem with either of the smaller of the two before and the last version.

kocio-pl:
Thanks, that's clear. 

I'd like at least to avoid big (20px) matrix, because now libraries stand out too much IMO (they're bigger than most other icons + they are solid). I may try to learn doing right arcs one day, but currently only the thickest 14px version is good enough for me (and is included in the patched version of the code) - it's not following strictly our new icon guidelines, but is still closer than current version and I'd like to make some progress.

daganzdaanda:
I like the last version well enough, but I wanted to try to modifiy the ideas that @vholten and @nebulon42 tested in the previous PR. I think that the pages of the book can be solid colour, and it is still recognizable.

library-16test
library-16test-x3

https://gist.github.com/daganzdaanda/ea49b6ebd06beaf0a8b4

imagico:
I think the modified spine makes it a good replacement for the current icon.  Using it IMO would not in any way prejudice more fundamental redesigns.

kocio-pl:
I would vote for @daganzdaanda version:
g1nhdlhg

I have just converted white into translucent (it's solid enough to be visible on busy background, I can test it if I get the link to such case), no dy tuning, because I wanted to see the overall result without loosing time for details, so the label text is shifted probably because of this. However I think such icon is better than my version (even that with additional spine arcs), more readable and closer to @nebulon42 sketches.

matkoniecz:
I also like @daganzdaanda version, simple and readable.

math1985:
> I also like @daganzdaanda version, simple and readable.

+1

polarbearing:
Open streetmap, Open data, Open book :-) +1

matkoniecz:
@kocio-pl 

> I would vote for @daganzdaanda version:

Can you update this PR to use this icon?

kocio-pl:
I have thought about making it symmetrical, since all three are a bit rough sketches (I guess my demo uses the first one), but for some time lately I didn't touch OSM icons at all. I hope I will play with them once again and finish it, but don't have exact time frame for this.

matkoniecz:
I found https://github.com/kocio-pl/openstreetmap-carto/tree/library-icon-open with version from https://github.com/gravitystorm/openstreetmap-carto/pull/2050?_pjax=%23js-repo-pjax-container#issuecomment-187957244 and I plan to submit it as a PR.

@kocio-pl @daganzdaanda Would it be OK? And I am unsure who should be set as author (I am unsure whatever I should rather credit who did the work or set myself to indicate that I take responsipility for any problems...).

matkoniecz:
I know that it is not perfect and may be improved but IMHO it is anyway better that what we have now - and it is svg.

kocio-pl:
I guess you're right, I see now that I'm not ready to draw something more currently. However I have no idea about the authorship assign - it's a small technical detail for me.

matkoniecz:
replaced by #2147 with icon from https://github.com/gravitystorm/openstreetmap-carto/pull/2050#issuecomment-187957244 and based on code from https://github.com/kocio-pl/openstreetmap-carto/tree/library-icon-open

PR #2057, opened 2016-02-22
fix oneway arrow colour for highway=path/footway and bicycle/horse=designated

### LINKED ISSUE #2054: Wrong colour for oneway arrow on path with bicycle=designated
oneway-path
https://www.openstreetmap.org/way/367094309

It's a bicycle=designated and foot=designated highway=path with oneway=yes. The arrow is red, the path blue.

nebulon42:
Fixes #2054.
Thanks @daganzdaanda for reporting.

before
designated_before

after
designated_after

### DIFF
--- a/roads.mss
+++ b/roads.mss
@@ -3047,6 +3047,12 @@ tertiary is rendered from z10 and is not included in osm_planet_roads. */
         [feature = 'highway_footway'],
         [feature = 'highway_path'] {
           marker-fill: @footway-oneway-arrow-color;
+          [horse = 'designated'] {
+            marker-fill: @bridleway-oneway-arrow-color;
+          }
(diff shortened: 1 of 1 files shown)


matkoniecz:
From reading code the same should happen for [highway=path; horse=designated; oneway=yes].

matkoniecz:
And it really happens, see https://www.openstreetmap.org/#map=19/52.80319/13.69840 

selection_005

(via http://overpass-turbo.eu/s/eyz )

nebulon42:
Thanks, I have not tested this, but updated the code. I have chosen the sequence so that it reflects the colouring of the line again.

matkoniecz:
Thanks for the fix!

Issue #2046, opened 2016-02-16
Combine road lines and areas to one layer


pnorman:
From https://github.com/gravitystorm/openstreetmap-carto/issues/1975#issuecomment-160467512

Doing this will reduce the number of layers and allow future MSS simplifications

pnorman:
@math1985 

Right now the layers we have are

* **`tunnels`**, class `tunnels-fill tunnels-casing access directions`
* barrier-like features and landuse overlay
* `ferry-routes`
* `turning-circle-casing`
* `highway-area-casing`
* **`roads-casing`**, class `roads-casing`
* `highway-area-fill`
* **`roads-fill`**, class `roads-fill access directions`
* `turning-circle-fill`
* `aerialways`
* `roads-low-zoom`
* `waterway-bridges`,
* **`bridges`**, class `bridges-fill bridges-casing access directions`

Entries in bold are versions of the large 6kb road query.

I've successfully brought `highway-area-fill` into `roads-fill`. On initial inspection, `roads-casing` is the same as `roads-fill`, so there will be two layers returning the same results.

Would it then be possible to combine them into one layer?

math1985:
Great work!

Yes, `roads-casing` and `roads-fill` can be merged without problem. Merging `highway-area-fill` and `roads-fill` was the hard part. For tunnels and bridges, `roads-casing` and `roads-fill` are in fact already merged.

It probably also makes sense then to bring highway-area-casing into roads-casing in the same way.

To merge tunnels with the main road query, we have two options:
1) Include `ferry-routes` and `turning-circle-casing` into the main road query. This will lead to an even huger query, but with less SQL in total as we don't need a separate tunnels query. As a side effect, it would also improve the rendering of turning circles on layers/bridges.
2) Render ferry routes and turning circles under tunnels. It is an option, but it would probably not look great.

To merge bridges with the main road query, we need to get rid of the layers `turning-circle-fill`, `aerialways`, `roads-low-zoom`, and `waterway-bridges`.

I think `aerialways`can be moved above `bridges` without problem. This seems a good idea even independent from the other changes, as most aerialways go indeed over bridges.

The layer `roads-low-zoom` can also be moved above `bridges` without problem, as it is only used in layers up to z9, while `turning-circle-fill` and `waterway-bridges` are only used in higher zoomlevels

For the remaining layers `turning-circle-fill` and  `waterway-bridges`, we have two options again.
[...]

PR #2058, opened 2016-02-23
Start merging road areas into roads layers

pnorman:
cross-ref #2046 

This is how far I got before I had to travel, but should reviewable on its own.


math1985:
Are the reordering in the .mss necessary and the whitespace changes necessary? If not, could you propose these as a separate PR?

pnorman:
> Are the reordering in the .mss necessary and the whitespace changes necessary? If not, could you propose these as a separate PR?

They're needed since they belonged to the roads area layer, which is now gone. The only whitespace change is because they end up nested differently within the roads fill layer.

I think there's an opportunity to refactor that part of the MSS and simplify it, but haven't looked at it in enough detail yet.

math1985:
Something is still wrong with this PR, areas now also render as linear roads. This causes them to be rendered too big (as the outline is rendered too).

I didn't know it was possible to just leave out the geometry field, by the way.

matkoniecz:
> Something is still wrong with this PR, areas now also render as linear roads. This causes them to be rendered too big (as the outline is rendered too).

What results in following problems: https://cloud.githubusercontent.com/assets/899988/15871159/70ebd0a2-2cf3-11e6-9256-bbe85db6c0ef.png

pnorman:
Closing as don't know when I'll have time to resume work on this.

PR #2060, opened 2016-02-23
Drop currently unused tags from osm2pgsql style file

math1985:


math1985:
Note that this is a PR against the hstore branch.

openstreetmap-carto.style
```diff
@@ -21,14 +20,10 @@ node       capital      text         linear
...
 node,way   covered      text         linear
 node,way   culvert      text         linear
-node,way   cutting      text         linear
 node,way   denomination text         linear
-node,way   disused      text         linear
 node       ele          text         linear
-node,way   embankment   text         linear
 node,way   foot         text         linear
 node,way   generator:source    text  linear
-node,way   harbour      text         polygon
```
pnorman: this should be changed to phstore, not dropped

math1985: Done.

pnorman: The same is true for any other tag that indicates an area and will not have a column

math1985: Done.

Issue #1547, opened 2015-05-14
natural=mud renders differently in rivers


malcolmh:
The background shading of the mud/tidalflat pattern disappears within waterway=riverbank areas. An example of this is http://www.openstreetmap.org/#map=16/53.7011/-0.4375 where a natural=mud area straddles the boundary between coastline and riverbank.

This issue is similar to #115 (closed as fixed) except that instead of no rendering, we get partial rendering.

imagico:
Indeed.  This problem is exactly the same as for all other landcovers, like for example beach/sand.  There is no good solution for this without using water polygons for the coastline.  Moving individual landcovers (like mud) above the water layer as it had been done before is not a solution IMO since this would be very confusing, small landcover areas like natural=sand within natural=mud previosuly did not show which violated the general principle that all landcover is rendered ordered by way_area.

Ideally i would think all landcovers should not be rendered above water (either ocean and inland) - see also #426 - except for a small subset which actually exists in a tidal/water covered form in reality and which should be rendered in a distinct form, in some cases probably overlay only, when overlapping water.

List of features where this should apply IMO:

- natural=wetland
- natural=mud
- natural=beach
- natural=shoal
- natural=reef
- military=danger_area
- landuse=cemetery

See also #1473 which is about the opposite side of the same problem

malcolmh:
The problem is not in the coastline areas, but in the riverbank areas. The background colour is absent, but not the pattern. Why is this?

imagico:
That is because the water areas (natural=water, waterway=riverbank) are drawn above the landcover (which contains the mud base color) and below the overlay with the wetland pattern.  Moving the mud base color to the overlay layer (or a separate layer above water areas) would not be desirable since it would make the mud color (which is semitransparent) interfere with other landcovers.  

From my point of view the options to change things are:

1. make the tidalflat/mud base color opaque and render it above water, this would be similar to as it was before with all the problems - like no other landcover stuff or water within mud areas being rendered and hiding of the coastline.
2. move the water areas below all landcover which would lead to a lot of problems for example where small water areas are drawn above larger landcovers.  Also water lines are currently partly drawn below water areas which would need to be changed causing additional problems.
3. render water areas within the landcover layer using normal way_area sorting.  This could be very confusing though.  And same problems with water lines as 2.
4. render tidalflat/mud like generic wetland again, i.e without base color.
[...]

PR #2066, opened 2016-02-28
Switch to ocean polygons

### LINKED ISSUE #1982: Switch to ocean polygons

pnorman:
This should have no rendering changes, as the base land and base ocean are still the first things drawn.

We may want to re-order later to put water polygons by ocean polygons, but this will require consideration of what landuse should appear on oceans

Fixes #1982.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -25,7 +25,9 @@ To download the shapefiles you can run the following script from this directory.
 You can also download them manually at the following paths:
 
 * [`simplified-land-polygons.shp`](http://data.openstreetmapdata.com/simplified-land-polygons-complete-3857.zip) (updated daily)
+* [`simplified-water-polygons.shp`](http://data.openstreetmapdata.com/simplified-water-polygons-complete-3857.zip) (updated daily)
 * [`land-polygon.shp`](http://data.openstreetmapdata.com/land-polygons-split-3857.zip) (updated daily)
+* [`water-polygon.shp`](http://data.openstreetmapdata.com/water-polygons-split-3857.zip) (updated daily)
 * [`builtup_area.shp`](http://planet.openstreetmap.org/historical-shapefiles/world_boundaries-spherical.tgz)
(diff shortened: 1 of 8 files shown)


pnorman:
cc @imagico

imagico:
This addresses #1982.  For #1547 changing layer order to render ocean polygons directly above/below water areas will be good but that can be considered a separate problem.

INSTALL.md
```diff
@@ -25,7 +25,9 @@ To download the shapefiles you can run the following script from this directory.
 You can also download them manually at the following paths:
 
 * [`simplified-land-polygons.shp`](http://data.openstreetmapdata.com/simplified-land-polygons-complete-3857.zip) (updated daily)
```
sommerluk: Can this line be removed?

INSTALL.md
```diff
@@ -25,7 +25,9 @@ To download the shapefiles you can run the following script from this directory.
...
+* [`simplified-water-polygons.shp`](http://data.openstreetmapdata.com/simplified-water-polygons-complete-3857.zip) (updated daily)
 * [`land-polygon.shp`](http://data.openstreetmapdata.com/land-polygons-split-3857.zip) (updated daily)
```
sommerluk: Can this line be removed?

math1985:
Thanks @pnorman!

math1985:
Unfortunately, I had to revert this change for now as it's still not fully clear what causes #2101. See also #2174.

I have created a branch ocean-polygons which can be used for testing the problem (this branch is not intended to be released, it is only there to make testing easier).

As soon as #2101 is sorted out, we can again try to move to ocean polygons.

math1985:
I have reverted this PR for now, see #2101.

Issue #2069, opened 2016-03-05
Recolour leisure=golf_course to distinguish scrub


Yorkshire-Jack:
# Problem:
## Golf course and scrub use same colour

Currently, golf courses and scrub use the same fill colour.  Near me, it is common to find gorse growing on golf courses, and it's mapped (correctly, IMO) as scrub. This leads to the scrub blurring into the course.  Additionally, since the scrub icon looks a little like long grass, people tag for the renderer and use it to indicate areas of longer grass or rough.
A typical example can be found in Edinburgh https://www.openstreetmap.org/#map=16/55.9102/-3.1955
![scrub_golf](https://cloud.githubusercontent.com/assets/17168780/13547113/1f2dc1fc-e2bc-11e5-8561-9846a2d2a73c.png)
## Golf course and water look quite similar

Also on the above example is a small lake on the course which has been tagged natural=water.  It's quite hard to distinguish, though that could be fixed if any changes are made as a result of current similarity to woodland, see #1781
## Tagging grass on golf courses

Note that the fairways in the above example have been tagged `landuse=grass`, which makes them visible. I consider this to be tagging for the renderer, as that's not what `landuse=grass` is for. Taginfo has 2500 golf courses worldwide,  and a similar number of `golf=green`s. Since golf courses have 9-18 greens, <10% of courses have `golf=` tagging, and of those that do, around 25% also have grass tagging, so this affects about 2% of total courses. The may be more which are tagged as `landuse=grass` but not `golf=`, but I can't find an easy way of counting those.
# Suggestions
## Use the same colour as grass

`natural=grassland`, `landuse=grass`, `landuse=meadow` `leisure=garden` and probably others all use the same colour.  Since that covers everything from manicured lawns to rolling open land with knee length grass, golf courses could fit into this easily. They would stay visible on the map because they also get an icon.  It would make the fairways in the above example disappear, whether that's a good thing probably depends if you consider it tagging for the renderer.
This option would be in keeping with the desire to simplify the map expressed in #1630.
## Use the same colour as sports pitches

Golf courses are somewhere you go to play a sport, just like a pitch.  The downside of this is that pitches are quite a dark colour, and golf courses are quite large, so they might receive undue prominence. This option also seems to fit #1630 reasonably well.
## Keep a unique colour.

Another option would be to keep rendering golf courses differently from grass and pitches, but chose another colour.  I do think it's worth making it different from scrub though.
# Interaction with #661

[...]

Issue #1991, opened 2015-12-04
New landuse=farmland rendering color too close of amenity=hospital


Penegal:
Hello, there.

I noticed a change in the color used to render landuse=farmland; I wasn't able to find references to this change in the last issues or pull requests, but it now seems too close of the amenity=hospital rendering to me: https://www.openstreetmap.org/#map=15/48.2899/6.1141 In this case, I was working on the landuses on this area, and I first thought I mis-modelled the landuses and mixed them with amenity=hospital. Improving the landuse=farmland rendering color, which was somewhat pinkish to me, was a good idea, but now I would say it became confusing.

Regards.

Edit: I found the corresponding pull request, #1701

donaldrnoble:
As the same shade of yellow is used for schools and universities, it also affects them. And to me it seems more of an issue at lower zoom levels, even though the same colours are being used. 

![screen shot 2015-12-04 at 10 53 56](https://cloud.githubusercontent.com/assets/6844639/11588001/5c3a11de-9a75-11e5-83b2-37417933a6c0.png)
http://www.openstreetmap.org/#map=14/55.9084/-3.3080

polarbearing:
This could be solved if we eventually switch to boundary-rendering for amenities as in #1624.

Vort:
Also farmland now is one of the brightest things, attracting attention like it is the most important feature on the map.

mboeringa:
> As the same shade of yellow is used for schools and universities, it also affects them. And to me it seems more of an issue at lower zoom levels, even though the same colours are being used.
> 
> screen shot 2015-12-04 at 10 53 56
> http://www.openstreetmap.org/#map=14/55.9084/-3.3080

Personally, looking at these to examples on my color calibrated screen, I see no real problem. The distinction in colour is still clearly visible, and the amenities stick out by their prominent - impossible to miss - labelling and icons.

> This could be solved if we eventually switch to boundary-rendering for amenities as in #1624.

\+ 1, for quite a lot of amenities, if not all, this would be appropriate (exception of course for additionally tagged as building). In fact, I have done this for many of these features in my personal ArcGIS renderer.

imagico:
On a general note: with changes like this which have a huge influence on the overall impression of the map - we had similar changes like those of wood and grass colors before - it is always good to take a few days to let it settle and try to look at it without bias.  Being used to the old rendering we tend to have a more negative impression of changes because they are less intuitive to us at first.

[...]

Issue #120, opened 2013-08-17
Add icon for amenity=school


al--:
Is the green school icon ![education_school p 16](https://f.cloud.github.com/assets/1458456/979279/5c0c5340-06f0-11e3-9026-0e19ac2dd0f4.png) missing? Why?

Brian also did icons for kindergarten or university...

/al

althio:
We should not assume that every school has a name or is tagged with a name.
from taginfo:
amenity=school with name=* :: 526 000 (83%)
amenity=school without name=* :: 106 000

An icon is useful for schools without name.
An icon is needed IMO for school on nodes, and nice to have on areas.
An icon takes less space than a name and could be rendered earlier or with less clash from other icons and labels.
[also a valid comment for #660]

The label name on the other hand should be rendered very late.
Absolute zoom levels are not enough since rendering icons and/or labels depends on the size (area) and on the density of features (clashes).

Lastly, when the school is mapped with details, the icon rendered is playground. Confusing for people looking for an public playground and confusing for people looking for the school (school can be used as any landmark and locally well-known place for orientation).
The school icon or label should somehow have more weight than the playground-in-a-school icon.

daganzdaanda:
+1 for adding an icon for schools. 
In some areas of Africa, I've seen lots of schools that are just nodes (maybe imported). These don't always have a name. I'd rate schools important enough that the icon should be visible at z>=14 like hospitals.

dispiste:
+ 1 for adding an icon for schools. Copy-pasting some reasons from talk mailing list:

"At least in Spain, they are very
important buildings from several points of view:
- social relevance (they congregate a very important amount of people
on a daily basis)
- for orientation (they are prominent and characteristic buildings
that can be used as reference locations)
- traffic influence (they often cause traffic jams or crowds in their
surrounding areas on opening and closing hours)"

Regarding the icon to use, I've found in Google a number of topics that work for successfully representing a school: blackboard, child(en), teacher, backpack or school bag, pencil, notebook, compass, protractor, ruler, abacus...

I specially the following combinations:
- child sitting on a school desk
- teacher writing on a blackboard
- child writing on a blackboard
- compass + protractor + ruler

kocio-pl:
My first set of drafts:
1) ![school-14-1](https://cloud.githubusercontent.com/assets/5439713/9788491/81e8dcfa-57ca-11e5-805c-682eff4ea794.png)
[...]

PR #2071, opened 2016-03-05
Change and unify colors for stadium/track/pitch to be less dominating

imagico:
This change replaces the very strong and dark color for stadium/sports_centre with the bright education/hospital color and unifies track/pitch into an intermediate green tone.

beforeafter

This is based on the following considerations:

* the dark and strong color of stadium/sports_centre was a major problem since it stands out very strongly putting  a lot of emphasis on  these features and it does not comply with the general concept of using brighter and more muted colors for large features and stronger colors for small features.  Using the education color seems fitting since there are a lot of similarities.
* having a bright, neutral base color for stadium also encourages micromapping of features within - which currently often look ugly against the strong dark green.
* having different colors for track and pitch makes the map less readable and it is a quite special and culturally specific distinction anyway.
* removing two distinct colors from the style helps map clarity in general.

The line color for track/pitch was also brightened a bit because contrast to stadium is already strong with the new colors and this way mapped barriers can be better distinguished from plain edges.

### DIFF
--- a/amenity-points.mss
+++ b/amenity-points.mss
@@ -1624,7 +1624,7 @@
       }
       [feature = 'leisure_sports_centre'],
       [feature = 'leisure_stadium'] {
-        text-fill: darken(@stadium, 30%);
+        text-fill: darken(@stadium, 70%);
       }
       [feature = 'leisure_track'] {
         text-fill: darken(@track, 40%);
--- a/amenity-points.mss
+++ b/amenity-points.mss
@@ -1610,7 +1610,7 @@
       [feature = 'amenity_school'],
       [feature = 'amenity_college'],
       [feature = 'amenity_university'] {
-        text-fill: darken(@educational_areas_and_hospital, 70%);
+        text-fill: darken(@societal_amenities, 70%);
       }
       [feature = 'natural_heath'] {
         text-fill: darken(@heath, 40%);
--- a/landcover.mss
+++ b/landcover.mss
@@ -46,7 +46,7 @@
 @power-line: darken(@industrial-line, 5%);
 @rest_area: #efc8c8; // also services
 @sand: #f5e9c6;
-@educational_areas_and_hospital: #f0f0d8;
+@societal_amenities: #f0f0d8;
 @station: #d4aaaa;
 @tourism: #734a08;
 @quarry: #c5c3c3;
@@ -57,7 +57,7 @@
 
 @pitch: #80d7b5;
 @track: @pitch;
-@stadium: @educational_areas_and_hospital; // also sports_centre
+@stadium: @societal_amenities; // also sports_centre
 @golf_course: #b5e3b5;
 
 #landcover-low-zoom[zoom < 10],
@@ -450,7 +450,7 @@
     [zoom >= 10] {
       polygon-fill: @residential;
       [zoom >= 12] {
-        polygon-fill: @educational_areas_and_hospital;
+        polygon-fill: @societal_amenities;
         [zoom >= 13] {
           line-width: 0.3;
           line-color: brown;
--- a/landcover.mss
+++ b/landcover.mss
@@ -1,18 +1,11 @@
 // --- Parks, woods, other green things ---
 
 @grass: #cdebb0; // also meadow, common, garden, village_green
-@golf_course: #b5e3b5;
 @scrub: #b5e3b5;
 @forest: #add19e;       // Lch(80,30,135)
 @forest-text: #46673b;  // Lch(40,30,135)
 @park: #c8facc;         // Lch(94,30,145) also recreation_ground
 @orchard: #aedfa3;
-
-// --- sports ---
-
-@stadium: #3c9; // also sports_centre
-@track: #74dcba;
-@pitch: #8ad3af;
 
 // --- "base" landuses ---
 
@@ -59,6 +52,13 @@
 @quarry: #c5c3c3;
 @military: #f55;
 @beach: #fff1ba;
+
+// --- sports ---
+
+@pitch: #80d7b5;
+@track: @pitch;
+@stadium: @educational_areas_and_hospital; // also sports_centre
+@golf_course: #b5e3b5;
 
 #landcover-low-zoom[zoom < 10],
 #landcover[zoom >= 10] {
@@ -520,7 +520,7 @@
     polygon-fill: @track;
     [zoom >= 15] {
       line-width: 0.5;
-      line-color: saturate(darken(@track, 40%), 20%);
+      line-color: saturate(darken(@track, 30%), 20%);
     }
     [way_pixels >= 4]  { polygon-gamma: 0.75; }
     [way_pixels >= 64] { polygon-gamma: 0.3;  }
@@ -530,7 +530,7 @@
     polygon-fill: @pitch;
     [zoom >= 15] {
       line-width: 0.5;
-      line-color: saturate(darken(@pitch, 40%), 20%);
+      line-color: saturate(darken(@pitch, 30%), 20%);
     }
     [way_pixels >= 4]  { polygon-gamma: 0.75; }
     [way_pixels >= 64] { polygon-gamma: 0.3;  }


kocio-pl:
Mostly +1. 

The only problem I see is education/hospital color - but since hospital has its own icon and school/university might have it too ([drafts are ready](https://github.com/gravitystorm/openstreetmap-carto/issues/120), but PRs and debates are still ahead), maybe we could start treating this color more generally as "big amenity" probably (or just let it be used for more objects).

polarbearing:
When track and pitch have the same colour you cannot distinguish them when mapped over each other. Your example shows a modern mapping of a multipolygon for the athletics track, leaving room for the ball pitch. There are still plenty such places where the track is just a line (that renders as an area in carto).

Similar problem when sports_centre and stadium have the same expression, what when a stadium is built on the grounds of the sports centre?

I share @kocio-pl's consideration in not overloading the current education/hospital colour with unrelated features.

I agree that sports need some treatment, and I repeat what I said about golf in #2069, that moving towards outline rendering for all sorts of campuses might bring progress.

imagico:
> When track and pitch have the same colour you cannot distinguish them when mapped over each other.

Since these are drawn with an outline you can always distinguish them - you just cannot identify them by color - which IMO is fine for a general purpose map style.

> Your example shows a modern mapping of a multipolygon for the athletics track, leaving room for the ball pitch. There are still plenty such places where the track is just a line (that renders as an area in carto).
> 
> Similar problem when sports_centre and stadium have the same expression, what when a stadium is built on the grounds of the sports centre?

Could you point to examples please?

> I share @kocio-pl's consideration in not overloading the current education/hospital colour with unrelated features.

They are not unrelated - a lot of sport centers are part of educational or medical facilities, all of these are usually public or semi-public but do not primarily serve commercial purposes (so landuse=commercial does not fit).

> I agree that sports need some treatment, and I repeat what I said about golf in #2069, that moving towards outline rendering for all sorts of campuses might bring progress.

I don't think so - current boundary rendering in this style is very bold, takes a lot of space and clutters the map and it only works at a very limited range of scales (with area sizes from a few ten to a few hundred pixels at normal rendering resolution) - at smaller sizes it is just noise and at larger sizes it is just confusing.  I would strongly recommend not using this for typically non-boundary features.

HolgerJeromin:
> Could you point to examples please?

A [stadium](http://www.openstreetmap.org/way/35710101) inside a [sport_centre](http://www.openstreetmap.org/way/111081531)

matkoniecz:
Given that this change would remove any chance of making light yellow mean anything specific - it would be a good idea to rename it from `@educational_areas_and_hospital` to something like `@unspecific_big_amenity` or something similar.

imagico:
stadium and sport_centre overlapping does not seem to look any worse than before:

stadium and sport_centre

Regarding color name, yes, that would likey be a good idea - although i would probably name it `@public_amenities`.

I am not really partial towards this unification by the way but as everyone knows there is very little room to squeeze in more area colors.  Looking at the urban context this essentially leaves two options for reducing complexity or adding new features:

* making specific colors more generic by using them for multiple similar or related feature types.  This is essentially what i did here.  The bright yellow also invites this since the subtle tone indicates something fairly unspecific.
* unifying the generic urban landuse colors (residential/retail/commercial/industrial) and using their room in color space for defining more colors for more specific areas.  This could work nicely IMO but this would be a major change in paradigm in this style.

polarbearing:
There are a number of campuses that are currently not rendered at all (#1624), i.e. community_centre/socialfacility/police, hostel/hotel/motel, museum, research_institute came up recently #2077, would those all be candidates for the same colour under the new name?

kocio-pl:
@polarbearing I would opt for it, with reasoning exactly like in the @imagico post - road redesign showed us clearly that we're basically out of colors and amenity campuses are basically more like landuses - they deserve being shown, because it explains how the space is used.

One "campus" I still have no clear idea what to do is landuse=religious. Maybe some kind of grey would do?

polarbearing:
>  ```landuse=religious```. Maybe some kind of grey would do?

Yes a slightly darker gray than residential would be good.

math1985:
I'm still in favour of rendering all campuses in yellow, independent of the type of campus.

kocio-pl:
OK, yellow would be good for me too.  And I would rather name it `@amenity_area` (or, less preferable, `@amenity_campus`/`@campus`), because it's more generic and we can't be sure if this campus is really public.

imagico:
It does not really matter how you label it.  The important thing is not to use it arbitrarily as a catch-all for everything that does not fit into any other category.  And things that have similarities to for example landuse=commercial or landuse=industrial should better be rendered in those colors.

kocio-pl:
If they belong to mentioned landuses, they're already here and we have no problem with them. But what about this orphanage for example:

http://www.openstreetmap.org/way/304204007

It is in fact very similar to the schools below, yet if not the fence, its area shape would be unknown, no matter how big it is. And such campus can be quite big:

http://www.openstreetmap.org/way/123382251

pnorman:
> if not the fence, its area shape would be unknown, no matter how big it is.

I'm fine with that - we're not going to render all area features possible.

kocio-pl:
Sure, but I'd like to see the structure of a given place and some of them are very common, like police/fire station or social/cultural/religious/sport amenities.

I don't advocate for amenity=* area (catch all), BTW.

math1985:
> Sure, but I'd like to see the structure of a given place and some of them are very common, like police/fire station or social/cultural/religious/sport amenities.

+1

polarbearing:
I called them 'Societal amenities' in #1624, and would be happy to see their areas rendered, if not as boundaries then as campus-yellow. 

We have icons on many of them meanwhile to distinguish them, we would need to push the school/univ/kids #120 eventually.

imagico:
Changed the color name now as suggested by @polarbearing.

dieterdreist:
sent from a phone

> Am 07.03.2016 um 17:42 schrieb Christoph Hormann <notifications@github.com>:
> 
> It does not really matter how you label it. The important thing is not to use it arbitrarily as a catch-all for everything that does not fit into any other category. And things that have similarities to for example landuse=commercial or landuse=industrial should better be rendered in those colors.
> 

I'm not in favor of using the same color for educational institutions like schools, universities and for hospitals, and now also for sports related features? Doesn't make more sense than leaving them white. Hospitals might often even be universities, but sporting facilities should be kept distinct, I'd rather prefer a lighter variant of the current "blueish-green" if the current tone is considered too heavy.

imagico:
I don't see room for bright green tones that would not be heavily confusing with various vegetation landcover colors.

dieterdreist:
2016-03-08 20:25 GMT+01:00 Christoph Hormann <notifications@github.com>:

> I don't see room for bright green tones that would not be heavily
> confusing with various vegetation landcover colors.



aren't the latter all green yellowish, while sport is blueish?

polarbearing:
So one thing to decide is if we want to use the current edu/medical yellow for more societal amenities at all, the second question is if we do so, do we want to expand that to leisure/sports features. We might open a separate ticket for the first question. 

For the sports grounds, we also need to consider that mappers also use ```landuse=recreation_ground``` for larger sports complexes that have a stadium inside. See this example [here](http://www.openstreetmap.org/way/258241195#map=17/52.54372/13.40732) where you have all, recreation_ground, sports_centre on a building, a stadium, plenty pitches and tracks, and some grass in between.

polarbearing:
I wonder why pitches have an outline, while stadium and sports_centre don't?

imagico:
From what i saw there seems to be a lot of tagging for the renderer going on w.r.t. sports_centre where people don't use that tag to avoid the strong color.  To me the above example quite clearly looks like it qualifies as sports_centre.  I did not look into other tags that could be rendered in addition, there seems to be both `landuse=recreation_ground` and `leisure=recreation_ground` with widespread use and i don't know if either of them is used consistently.

It could be considered keeping the strong green for stadium only - however at the moment it seems stadium is often used as a simple wrapper for pitch/track combinations without substantial seating which does not really match the tag definition.

polarbearing:
> stadium is often used as a simple wrapper for pitch/track combinations without substantial seating which does not really match the tag definition

yes this is a tagging error I fix when I find it

matkoniecz:
I admit that I really have no idea what to think about this change, so to date I left no comments about proposed change.

On one hand current sport areas rendering is poor and it is improved by this change. On the other hand I am not convinced by generic amenity, in fact I would prefer to drop rendering hospitals and universities as the same color.

In case of making a new map I would just do not render residential/retail/commercial landuses separately what would free some colors. It would be easily possible to color education areas and hospitals with separate colours and maybe still leave some colors to be used for cases like sport facilities.

But in this case it would be probably nearly impossible to stop rendering landuse, colour pool is limited. 

So maybe yellow for generic amenity is not so bad? Or maybe abandon idea of rendering some amenities as areas?

kocio-pl:
I still like using current school/hospital yellow for general "campus", because with such definition it become clear what those two have in common. I would not use it for general "amenity" currently, because I'm not sure if it would make sense for such a broad and not clearly defined type. However if we look closer and learn that "campus" is basically just any "amenity area", it would be a lot easier than making a list of campus types (like "sport campus" for stadium).

Dropping would make the map less clear IMO - I already don't like that the churchyards don't have a color, because they can take the same space as local school.

imagico:
Another possibility - if you look at the colors in https://github.com/gravitystorm/openstreetmap-carto/issues/1991#issuecomment-162174503 garages and aerodrome could be unified into a general 'transportation' landuse or the aerodrome color could be re-purposed into a general public amenities color while keeping hospital/education as a distinct color.

Not sure if any of these options is better though.

kocio-pl:
I have already proposed unification of "transportation area" for airports/heliports, ferry terminals and train stations - garages look to me more like resting area (depot) for some devices than real transportation, but I wouldn't mind to make them look just like residential areas and reuse this color somewhere else.
 
Sure, we could use this color for general amenities, but still my question remains - why hospital and school have the same color then, what do they have in common? If they are both special campuses, we should also reuse it for other special campuses, like police or religious and use garage color for any other amenity areas.

It could also work nicely with complex situation (as suggested [here](https://github.com/gravitystorm/openstreetmap-carto/pull/2071#issuecomment-195286013) by @polarbearing) with recreation_ground (like general amenity area) and stadium/sports_centre (like special campus).

dieterdreist:
sent from a phone

> Il giorno 09 giu 2016, alle ore 23:08, Christoph Hormann <notifications@github.com> ha scritto:
> 
> Not sure if any of these options is better though.
> 


IMHO we shouldn't assimilate garages (very common feature, not significant , IMHO more storage than transportation related) with airports (rarer, significant feature)

I also don't think that hospitals and universities have to get or keep a common color

dieterdreist:
sent from a phone

> Il giorno 09 giu 2016, alle ore 22:37, kocio-pl <notifications@github.com> ha scritto:
> 
> I still like using current school/hospital yellow for general "campus", because with such definition it become clear what those two have in common
> 


what do they have in common? 

> - I already don't like that the churchyards don't have a color, because they can take the same space as local school.
> 


aren't churchyards rendered the same as christian cemeteries? Seems right to me

kocio-pl:
> what do they have in common?

They are campuses.

> aren't churchyards rendered the same as christian cemeteries? Seems right to me

I guess not - both landuse=churchyard and (more general, for other religions) landuse=religious are currently not rendered in osm-carto at all.

dieterdreist:
sent from a phone

> Il giorno 13 giu 2016, alle ore 14:21, kocio-pl <notifications@github.com> ha scritto:
> 
> They are campuses.
> 


for schools it is rather atypical to have a campus in my area (although they have dedicated grounds outside buildings), and also in case of a university or hospital not all associated open air areas are campuses. A campus is a specific type of spatial organization (park that connects several buildings of the same institution , typically a college or university in Northern America or something copying this concept, e.g. Apple campus in Cupertino).

I believe having an open air space that connects different buildings and is part of a common facility is no enough similarity to justify rendering it the same, you can find these criteria fulfilled in a lot of different settings, e.g. military, monastery, industrial, airports and whatever...

> aren't churchyards rendered the same as christian cemeteries? Seems right to me
> 
> I guess not - both landuse=churchyard and (more general, for other religions) landuse=religious are currently not rendered in osm-carto at all.
> 

the tag amenity=graveyard is rendered, that's a synonym I believe 

"religious" is not a landuse in my view, but rather an attribute that can occur with any landuse, so I'm fine with ignoring this tag here

kocio-pl:
> I believe having an open air space that connects different buildings and is part of a common facility is no enough similarity to justify rendering it the same, you can find these criteria fulfilled in a lot of different settings, e.g. military, monastery, industrial, airports and whatever...

Sounds reasonable to me. What are your positive propositions/guidelines regarding school, hospital and other amenity areas then (if you have any)?

> the tag amenity=graveyard is rendered, that's a synonym I believe
> "religious" is not a landuse in my view, but rather an attribute that can occur with any landuse, so I'm fine with ignoring this tag here

See here: http://wiki.openstreetmap.org/wiki/Tag:landuse%3Dchurchyard

dieterdreist:
sent from a phone

> Il giorno 13 giu 2016, alle ore 18:18, kocio-pl <notifications@github.com> ha scritto:
> 
> Sounds reasonable to me. What are your positive propositions/guidelines regarding school, hospital and other amenity areas then (if you have any)?
> 



I agree we could somehow render the areas of certain features, but I would avoid the word campus, and I wouldn't put education and healthcare in the same group 


> the tag amenity=graveyard is rendered, that's a synonym I believe
> "religious" is not a landuse in my view, but rather an attribute that can occur with any landuse, so I'm fine with ignoring this tag here
> 
> See here: http://wiki.openstreetmap.org/wiki/Tag:landuse%3Dchurchyard
> 


this should clearly be a proposal, it even has contradictions in its definition (the linked wp article says that these are mainly graveyards, while the wiki states that graveyards are excluded)

polarbearing:
amenity=graveyard is a long-standing tag for small cemeteries, in particular around places of worship. 

landuse=churchyard, however, was the ducktagger's response to landuse=religious, limiting the tag to one particular religion.

[landuse=religious](https://wiki.openstreetmap.org/wiki/Tag:landuse%3Dreligious) was already proposed to be rendered in [trac](https://trac.openstreetmap.org/ticket/5197) and again in #771, where it was rejected with only 1.6k uses, and for definition issues that have been fixed. It is growing in popularity, we have now over 7.3k, so that should be reconsidered (despite @dieterdreist not liking it ;-) )

kocio-pl:
@polarbearing +1

@dieterdreist Anything more specific? If we reuse garages color, how would you assign those two colors for different amenity areas?

math1985:
To me this looks like an improvement. As no alternatives have been proposed, I'm going to accept this PR.

Of course, this is not a final choice, and other rendering suggestions are always welcome.

polarbearing:
OK, could somebody please render my [example](http://www.openstreetmap.org/way/258241195#map=17/52.54372/13.40732) from above?

Does this now open the door to render { community_centre/social_facility/police, hostel/hotel/motel, museum, research_institute } campus the same?

kocio-pl:
The change does not look big here:
before (your picture)
before
after
v_8ykyg

polarbearing:
The F-L-J-Stadion remains recognisable since the whole campus is tagged as landuse=recreation_ground.

dieterdreist:
sent from a phone

> Il giorno 07 lug 2016, alle ore 02:01, kocio-pl <notifications@github.com> ha scritto:
> 
> The change does not look big here:


the bigger labels look much better, the color change doesn't look good IMHO, there's no logical connection between the stadium and a school or university 

cheers,
Martin

imagico:
@math1985 - thanks for merging.

To help further discourse here an updated rendering of the current area color set:

carto area colors

dieterdreist:
2016-07-07 11:41 GMT+02:00 Christoph Hormann <notifications@github.com>:

> To help further discourse here an updated rendering of the current area
> color set:



you're missing the amenity areas. E.g. hospitals, unis and schools, police.

imagico:
> you're missing the amenity areas. E.g. hospitals, unis and schools, police.

That is 'societal amenities' now:

https://github.com/gravitystorm/openstreetmap-carto/blob/master/landcover.mss#L49

dieterdreist:
2016-07-07 14:26 GMT+02:00 Christoph Hormann <notifications@github.com>:

> That is 'societal amenities' now:



thank you and sorry for overlooking this. Is there a definition what
belongs in there / what is the topic? What about libraries, museums,
prisons, social facilities, thermal baths, theatres, cinemas, conference
centres, marketplaces, fire fighters, beach resorts and racing circuits?

polarbearing:
The term societal amenities was coined in #1624, though sports facilities were not part of the original concept.

Klumbumbus:
Could we have an outline for stadiums? They are now nearly invisible. See e.g.
http://www.openstreetmap.org/#map=18/50.84206/12.94595
or
http://www.openstreetmap.org/#map=18/50.59777/12.71135

HolgerJeromin:
@Klumbumbus I added this request to my #2388

Issue #1974, opened 2015-11-14
Use different color for tertiary?


RicoElectrico:
I think new white color of tertiary roads is misleading, kind of breaking the mapper feedback for novices. It should be clear that tertiary is meant for roads that are somehow more important than residential, and the thickness gradation isn't really that easy to see.
Another issue is that in some areas [like here](http://www.openstreetmap.org/#map=13/50.8376/16.6445), tertiary is overused as a default. When it was yellow, it was very easy to see and it would catch attention. I've also seen tertiary abuse in Africa or Middle East cities, but I don't remember where exactly.

polarbearing:
Could you specify what you mean with "overuse"? In your example, there are some main roads going through the villages, so tertiary appears fine to me (subject to the country-specific tertiary definition and ground survey of priorities). 

There are certainly pros and cons of white wider tertiary, but they have been discussed here at length, and a decision made in favour. For me, the width works fine.

RicoElectrico:
My reasoning is that in this example tertiary is effectively the lowest tier besides driveways. If these were larger villages (that is with residential streets branching off), putting tertiary everywhere would be fine.
And do you see something? Judging relative width is not feasible when there are no thinner roads around. That's my point.
A slight yellow tint would be OK. Or some other minor variation, it doesn't have to be prominent.

It's always like that, no matter how big discussion is, if it can't be deployed as a public beta, there always will be something that people find not right. So because I guess public preview is not technically feasible due to resources needed, no discussion should be deemed final.

math1985:
There has been a public beta. It was announced [here](https://lists.openstreetmap.org/pipermail/talk/2015-August/073815.html).

Teiron:
I would like to express my support to this idea. Tertiary roads don't stand out enough.

But more importantly I keep hearnig about people feeling that road color change wasn't announced properly, despite great lot of talk at mailing lists, forums and here. I think there should be a easliy visible button on mainpage, which should link to "beta" version of map. This map may not be hosted permamently but maybe only one or two weeks before realese of new version of style. Users should be able to easliy give feedback somewhere where the developers will read it. Also button should be greyed out when beta is inactive. 

This should quell some critcism, while enhancing cooperation within comunity.

kocio-pl:
[...]

PR #2078, opened 2016-03-06
Make tertiary road yellowish

kocio-pl:
Resolves https://github.com/gravitystorm/openstreetmap-carto/issues/1974.

Trivial patch with non-trivial consequences, please propose some testing places from different parts of the world to see. Some screenshot samples around Warsaw for the start (click to see the full resolution):

1) City with residential, secondary and parking areas:
zatiq0u3

2) City with school and hospital area around:
uody0wgv

3) Outside the city with current (new) farmland color:
soqg3vnv

### DIFF
--- a/roads.mss
+++ b/roads.mss
@@ -5,7 +5,7 @@
 @trunk-fill: #f9b29c; // Lch(79,33,42), error 0.7
 @primary-fill: #fcd6a4; // Lch(88,31,74), error 1.7
 @secondary-fill: #f7fabf; // Lch(97,29,106), error 1.7
-@tertiary-fill: #ffffff;
+@tertiary-fill: #feffdf;
 @residential-fill: #ffffff;
 @service-fill: @residential-fill;
 @living-street-fill: #ededed;


mboeringa:
> please propose some testing places from different parts of the world to see

Maybe some of the places @Stalfur has been pointing out in his posts while discussing the new road colors?:
- https://github.com/gravitystorm/openstreetmap-carto/pull/1736#issuecomment-130019346
http://www.openstreetmap.org/#map=16/64.0992/-21.8926
http://www.openstreetmap.org/#map=12/64.1188/-21.8933

- https://github.com/gravitystorm/openstreetmap-carto/pull/1736#issuecomment-130537391
http://www.openstreetmap.org/#map=15/-24.4705/25.7883

- https://github.com/gravitystorm/openstreetmap-carto/pull/1736#issuecomment-130822217
http://www.openstreetmap.org/#map=14/-26.3326/21.5226

math1985:
Looks good to me, at least in the screenshots.

How does it look on/near school areas?

I would also like to hear the opinion of @matkoniecz.

Stalfur:
Possible near school area to test http://www.openstreetmap.org/#map=17/64.10301/-21.89453

pnorman:
Changing the colours needs to include changing the scripts that generate highway colours

kocio-pl:
4) Botswana lonely road on a land color (z13/15/17):
w2aqzku6
ammrt9yb
8m0ll d

Stalfur:
This is great.

kocio-pl:
@pnorman I guess all our scripts need at least basic documentation - what is their purpose, who maintains them etc. For example it's completely not clear for me what [generate_road_colours.py](https://github.com/gravitystorm/openstreetmap-carto/blob/master/scripts/generate_road_colours.py) does (aside from printing some color numbers on the screen). It would be good to include simple comment inside the code and let the script print some informations when called without arguments.

matkoniecz:
> Should we open separate issue for this?

That would be a good idea.

math1985:
Tested this in Luxembourg, looks great as far as I'm concerned.

@matkoniecz What do you think?

polarbearing:
With the current discussion in #2071 to use the campus-yellow more widely, it would be good to see a test of a tertiary road running across such a campus.

Stalfur:
Campuses have tertiary roads?

math1985:
Yes, for example [here](http://www.openstreetmap.org/#map=16/52.0854/5.1707). Unfortunately there are too much other colours around to compare the colour contrast well enough.

kocio-pl:
Thanks for the example! On micro scale (where there are not too much other elements and the roads are probably wider) it looks good enough for me:
z18:
c7v7p9bh
z19:
t xauvpz

If there are some not well micromapped examples I can test it too.

polarbearing:
Ok the hue is different enough, and the roads have their outlines, thus it works for me.

mboeringa:
Well, I see @math1985 merged this, and I think there will be quite some happy OSM'ers once it makes it into the rendering.

math1985:
Whoops, didn't intend to merge this yet! However everybody seems happy with it so I guess merging was fine. I will roll back the merge if somebody still comes with areas where this change is problematic.

mboeringa:
> Whoops, didn't intend to merge this yet! However everybody seems happy with it so I guess merging was fine. I will roll back the merge if somebody still comes with areas where this change is problematic.

Yes, I think it will be fine, considering @kocio-pl's example renderings and responses here. This was already on a waiting list to be tackled anyway.

pnorman:
> Whoops, didn't intend to merge this yet! However everybody seems happy with it so I guess merging was fine. I will roll back the merge if somebody still comes with areas where this change is problematic

Because the changes will get done the next time someone repopulates the colours with the script, I think we need to roll it back

math1985:
I have no access to git right now, could somebody else roll it back?

pnorman:
> I have no access to git right now, could somebody else roll it back?

Done

gravitystorm:
I've reopened the issue. I don't think there's a way to mark this as "unmerged". Unless there is some hidden option @kocio-pl can you please open a fresh PR? We can then discuss further the script changes that need to be made before merging again.

pnorman:
I'd like to move road colours to their own file and have that entire file generated by the scripts - this would allow regeneration of the colours from the script without any copy/paste

kocio-pl:
I can do a fresh PR, but it's a trivial one-liner and maybe we should wait until the script @pnorman is talking about is ready?

pnorman:
> I can do a fresh PR, but it's a trivial one-liner and maybe we should wait until the script @pnorman is talking about is ready?

The script already exists, what I was talking about would just make it easier to run.

kocio-pl:
I meant "when it's done", but it can be also discussed and refined within the new PR scope, so I opened it anyway.

Issue #881, opened 2014-08-13
Add rendering for amenity=vehicle_inspection


tux2000:
It seems that currently this tag is not rendered neither as node nor as area. Its not heavily used yet but it definitively is in use.

matkoniecz:
Used 512 times worldwide.

dieterdreist:
> Am 15.03.2015 um 13:10 schrieb Mateusz Konieczny <notifications@github.com>:
> 
> Used 512 times worldwide.


this is based on a proposal in draft status and should not have been created as a tag definition page IMHO. The proposed tagging makes it incompatible with other tags like amenity=fuel
Could have been an attribute rather than a top level tag.

PR #2096, opened 2016-03-24
Vehicle inspection

Stalfur:
Added rendering for currently non-rendered amenity=vehicle_inspection.
Noticed a rising trend in people using shop=car for anything car related in order to be rendered, devaluing data accuracy.
Yearly vehicle inspection is mandatory in EU and EFTA countries and in other countries as well. 
The changes are basically to add amenity=vehicle_inspection to be rendered alongside other amenities, using shop=car icon but rendered in amenity brown instead of shop purple.

Preview of icon (area: http://www.openstreetmap.org/#map=19/64.10542/-21.86522 ) 
vehicle_inspection_preview

### DIFF
--- a/amenity-points.mss
+++ b/amenity-points.mss
@@ -145,6 +145,13 @@
     marker-placement: interior;
     marker-clip: false;
   }
+  
+  [feature = 'amenity_vehicle_inspection'][zoom >= 17] {
+    marker-file: url('symbols/shop_car.svg');
+    marker-placement: interior;
+    marker-clip: false;
+    marker-fill: @amenity-brown;
+  }
 
   [feature = 'tourism_chalet'][zoom >= 17] {
     point-file: url('symbols/chalet.p.16.png');
--- a/project.mml
+++ b/project.mml
@@ -1,832 +1,828 @@
 {
-  "interactivity": false,
+  "scale": 1,
+  "name": "OpenStreetMap Carto",
+  "srs": "+proj=merc +a=6378137 +b=6378137 +lat_ts=0.0 +lon_0=0.0 +x_0=0.0 +y_0=0.0 +k=1.0 +units=m +nadgrids=@null +wktext +no_defs +over",
   "Layer": [
     {
       "name": "ocean-lz",
-      "srs-name": "900913",
-      "geometry": "polygon",
-      "class": "ocean",
-      "id": "ocean-lz",
       "srs": "+proj=merc +a=6378137 +b=6378137 +lat_ts=0.0 +lon_0=0.0 +x_0=0.0 +y_0=0.0 +k=1.0 +units=m +nadgrids=@null +wktext +no_defs +over",
+      "advanced": {},
+      "properties": {
+        "maxzoom": 9
+      },
+      "srs-name": "900913",
+      "extent": [
+        -180,
+        -85.05112877980659,
+        180,
+        85.05112877980659
+      ],
+      "id": "ocean-lz",
+      "geometry": "polygon",
       "Datasource": {
         "type": "shape",
         "file": "data/simplified-water-polygons-complete-3857/simplified_water_polygons.shp"
       },
+      "class": "ocean"
+    },
+    {
+      "name": "ocean",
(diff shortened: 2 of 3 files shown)


Stalfur:
Hmmm. I did run ./scripts/yaml2mml.py .

kocio-pl:
I think they could be rendered rather like shop=car_repair (https://github.com/gravitystorm/openstreetmap-carto/pull/1778), since I guess this is just a special kind of such shops.

SomeoneElseOSM:
There's probably a tagging discussion to be had elsewhere, but it's worth mentioning in passing that in some countries there's a substantial subset of such places that are "check" but not "repair".  See for example:

https://help.openstreetmap.org/questions/33229/how-to-tag-a-smog-check-center-only-in-america

dieterdreist:
sent from a phone

> Am 24.03.2016 um 15:23 schrieb Jóhannes Birgir Jensson <notifications@github.com>:
> 
> The changes are basically to add amenity=vehicle_inspection to be rendered alongside other amenities, using shop=car icon but rendered in amenity brown instead of shop purple.


the proposal for this tag is still in draft status. This tag works in some context (standalone vehicle inspection place) but not in others (where it is an attribute, a service offered besides others), and there wasn't any discussion about this AFAIK.

dieterdreist:
sent from a phone

> Am 24.03.2016 um 15:42 schrieb kocio-pl <notifications@github.com>:
> 
> I think they could be rendered rather like shop=car_repair (#1778), since I guess this is just a special kind of such shops.



sometimes it is, sometimes not, might be country specific. There are car repairs certified to do it, but there are also stand alone businesses that only do vehicle inspection.

math1985:
In Luxembourg, and likely some other countries as well, vehicle inspection is done exclusively by the government. Those governmental vehicle inspection stations (obviously) don't do repairs.

kocio-pl:
Leaving aside discussions about tagging scheme itself, I don't have a strong opinion on the symbol, though probably such entity uses similar tools as in car repair shop, so I slightly prefer this icon.

Stalfur:
In Iceland these used to be government run and only did inspections. This was privatized and currently 3 companies run these inspections. None of those does repairs hence my choice of icon not being the repair icon.

As for tag status and draft mode - there is no gold standard yet in OSM is there....

kocio-pl:
So maybe the flashlight (or even the eye!) instead of wrench?

Stalfur:
Flashlight would be my choice

HolgerJeromin:
Or a magnifier

Stalfur:
Yeah, flashlight, magnifier would both work. The eye is a bit Orwellian...

pnorman:
806 uses seems insufficient for a general-purpose rendering.

Stalfur:
Without support in editors this has not been widely tagged. Adding them in iD for example requires me to make a point and filling in values by hand (I'm preparing a pull request for that, have already created values in my branch of iD) - for a seasoned mapper this is not a problem but for new ones it is.

In making it easier to enter these into iD we will be making sure more will be added.

Also as a legal requirement it is an obligatory stop for car owners every year (and their trailers etc), as many other official in nature institutes (or privately run entities that fullfill the legal requirements on authority of the state) they are few in numbers.

The importance is not tied to numbers always, embassies for example are few (7200) compared to car shops (62.000) but we render them for their importance, not numbers.

matkoniecz:
> Without editor support it will not be added

It is a good idea to propose rendering in editors/including in tagging preset/validator rule before proposing rendering in map styles (editors are more inclusive and their map style is much closer to "show everything").

matkoniecz:
Also, it would be better to create new issue and wait for reaction before making PR (or in that case - requesting #881 to be reopened). At least [I prefer doing it this way](https://github.com/gravitystorm/openstreetmap-carto/issues?utf8=%E2%9C%93&q=is%3Aissue+author%3Amatkoniecz+is%3Aclosed+label%3Ainvalid).

> 806 uses seems insufficient for a general-purpose rendering.

I agree, especially as tag is without editor support Either because nobody noticed it - in that case just make request and it will be likely to be added, maybe follow with PR if maintainers are busy with other things. Other possibility is that it was rejected by JOSM/iD/other editors for some reasons (controversial, bad tagging scheme...) in that case it is rather poor idea to include it here.

PR #2100, opened 2016-03-30
Add custom index information

### LINKED ISSUE #207: Suggested additional indexes

pnorman:
These custom indexes should speed up rendering, particularly at low and middle zooms.

Fixes #207

cc @woodpeck @jburgess777 @Firefishy 

I'm not recommending any way_area indexes, not because I don't think they're worth it, but because they're more complicated to get right, and I'd rather go ahead and recommend a number of indexes I already know help performance.

### DIFF
--- a/indexes.sql
+++ b/indexes.sql
@@ -1,0 +1,64 @@
+-- These are optional but suggested indexes for rendering OpenStreetMap Carto
+-- with a full planet database.
+
+-- Ideally the indexes would be created with fillfactor=100 when updates are not
+-- consumed, but this does the simpler case of being used where minutely updates
+-- are consumed.
+
+-- A goal with the indexes is to have them general-purpose enough to not need
+-- frequent changing with stylesheet changes, but to be usable with many
+-- versions, and potentially other styles.
+
+-- The point tables are small, but so are the partial indexes
+CREATE INDEX IF NOT EXISTS planet_osm_point_place_low
+  ON planet_osm_point USING gist (way)
+  WHERE place IN ('city', 'town') AND name IS NOT NULL;
+
+CREATE INDEX IF NOT EXISTS planet_osm_point_place
+  ON planet_osm_point USING gist (way)
+  WHERE place IS NOT NULL AND name IS NOT NULL;
+
+CREATE INDEX IF NOT EXISTS planet_osm_point_name
+  ON planet_osm_point USING gist (way)
+  WHERE name IS NOT NULL;
+
+-- Most rows in the line table are highways, but there are some indexes useful
+-- for low zooms
+CREATE INDEX IF NOT EXISTS planet_osm_line_name
+  ON planet_osm_line USING gist (way)
+  WHERE name IS NOT NULL;
+
+CREATE INDEX IF NOT EXISTS planet_osm_line_ferry
+  ON planet_osm_line USING gist (way)
+  WHERE route = 'ferry';
+
+CREATE INDEX IF NOT EXISTS planet_osm_line_river
+  ON planet_osm_line USING gist (way)
+  WHERE waterway = 'river';
+
+-- The polygon table is by far the largest, and generally the slowest
+CREATE INDEX IF NOT EXISTS planet_osm_polygon_name
+  ON planet_osm_polygon USING gist (way)
+  WHERE name IS NOT NULL;
+
+CREATE INDEX IF NOT EXISTS planet_osm_polygon_nobuilding
+  ON planet_osm_polygon USING gist (way)
+  WHERE building IS NULL;
+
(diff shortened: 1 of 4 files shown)


pnorman:
Suggested to me was an index for low zoom admin polygons to reflect the switch to OSM data for low zoom labeling. My benchmarking was prior to this, but I recall it being a good idea.

rrzefox:
The `IF NOT EXISTS` will only work for postgresql >= 9.5. 9.4 or older do not have it (so in particular, the 9.3 in ubuntu 14.04 does not have it).

pnorman:
> The IF NOT EXISTS will only work for postgresql >= 9.5. 9.4 or older do not have it (so in particular, the 9.3 in ubuntu 14.04 does not have it).

Updated.

indexes.sql
```diff
@@ -0,0 +1,64 @@
...
+  ON planet_osm_point USING gist (way)
+  WHERE place IN ('city', 'town') AND name IS NOT NULL;
+
+CREATE INDEX planet_osm_point_place
+  ON planet_osm_point USING gist (way)
+  WHERE place IS NOT NULL AND name IS NOT NULL;
+
+CREATE INDEX planet_osm_point_name
+  ON planet_osm_point USING gist (way)
+  WHERE name IS NOT NULL;
```
pnorman: This index is 653MB and only used by `text-point`, so I'm not this is worth the additional memory cache pressure.

pnorman:
I have a WIP changing how these are defined to allow creation of SQL statements to reindex, etc.

pnorman:
* `indexes.sql` is now a generated file, but shouldn't need to be changed much
* The script options are
```
  -h, --help            show this help message and exit
  --concurrent          Generate indexes CONCURRENTLY
  --fillfactor FILLFACTOR
                        Custom fillfactor to use
  --notexist            Use IF NOT EXISTS (requires 9.5)
  --osm2pgsql           Include indexes normally built by osm2pgsql
  --reindex             Rebuild existing indexes
```

Examples are
default:
```sql
CREATE INDEX planet_osm_roads_admin
  ON planet_osm_roads USING GIST (way)
  WHERE boundary = 'administrative';
```
``--fillfactor 100``:
```sql
CREATE INDEX planet_osm_roads_admin
  ON planet_osm_roads USING GIST (way)
  WITH (fillfactor=100)
  WHERE boundary = 'administrative';
```
``--concurrent``:
```sql
CREATE INDEX CONCURRENTLY planet_osm_roads_admin
  ON planet_osm_roads USING GIST (way)
  WHERE boundary = 'administrative';
```
``--reindex``:
```sql
REINDEX planet_osm_roads_admin;
```
``--reindex --concujrrent``
```sql
ALTER INDEX planet_osm_roads_admin RENAME TO planet_osm_roads_admin_old;
CREATE INDEX CONCURRENTLY planet_osm_roads_admin
  ON planet_osm_roads USING GIST (way)
  WHERE boundary = 'administrative';
DROP INDEX planet_osm_roads_admin_old;
```

`--osm2pgsql` simply adds statements for `planet_osm_[point|line|polygon|roads]_index`

indexes.yml
```diff
@@ -0,0 +1,34 @@
...
+    where: landuse = 'military'
+roads:
+  # The roads table only has a subset of data, so it's just got some low-zoom
+  # indexes and some fairly selective ones for high zoom
+  admin_low:
+    where: boundary = 'administrative' AND admin_level IN ('0', '1', '2', '3', '4')
+  admin:
+    where: boundary = 'administrative'
+  roads_ref:
+    where: WHERE highway IS NOT NULL AND ref IS NOT NULL
```
pnorman: shouldn't have the WHERE here

indexes.yml
```diff
@@ -0,0 +1,34 @@
+# A goal with the indexes is to have them general-purpose enough to not need
+# frequent changing with stylesheet changes, but to be usable with many
+# versions, and potentially other styles.
+
+point:
+  # The point table is small, but so are the partial indexes
+  place_low:
+    where: place IN ('city', 'town') AND name IS NOT NULL
```
pnorman: turns out this isn't worth it

pnorman:
I think this is ready

math1985:
Thanks!

PR #2110, opened 2016-04-18
Split project.yaml out into separate files in project.d/

asbjornst:
Splitting project.yaml up makes it easier to maintain derivative projects, and should in general be easier to work with.

I have only split out a few layers, since I want to know if this idea has a chance first.


pnorman:
* Some of the .yaml files are not valid yaml as they refer to objects not in the current file
* This prevents running kosmtik on the project.yaml file and editing it
* If going this route, you'd want to create project.yaml and then run yaml2mml, e.g. something like `find project.d -name '*.yaml' -type f | LC_ALL="C" sort | xargs cat > project.yaml && ./yaml2mml.py`
* Once Tilemill gets YAML support, the JSON MML and yaml2mml.py are both likely to be removed.

HolgerJeromin:
Splitting files was rejected in https://github.com/gravitystorm/openstreetmap-carto/issues/1802#issuecomment-138809013

pnorman:
Note: Ideas to make it easier for other projects to use our layer definitions are welcome, but we have the constraint of `carto` and kosmtik needing to operate directly on `project.yaml`

PR #2119, opened 2016-05-05
Show national park boundaries from higher zoomlevel and larger size

matthijsmelissen:
With this commit, national park boundaries will be shown from the
same zoomlevel and size as their label, effectively eradicating
small unlabelled national parks from the rendering.

Such small national parks seem to contribute to a crowded feel of
the map in the areas and zoom levels where they occur.

### DIFF
--- a/admin.mss
+++ b/admin.mss
@@ -190,7 +190,7 @@ overlapping borders correctly.
 }
 
 #nature-reserve-boundaries {
-  [way_pixels > 100][zoom >= 8] {
+  [way_pixels > 3000][zoom >= 8] {
     [zoom < 10] {
       ::fill {
         opacity: 0.05;


pnorman:
I prefer showing the parks too small to fit a label in

2119_new1
2119_master

imagico:
IMO neither of these looks good, displaying national parks at this scale in this way seems to be causing more harm than have advantages.

In the US a lot could be gained by dissolving the inner boundaries and essentially showing the protected are as a whole - still keeping the labels but without indications where one park ends and another starts - which is confusion at this scale anyway.

But this would not solve all of the problems related, especially the color distortions of other elements through the semitransparent fill and border and the color confusion with natural=wood.

pnorman:
I think completely changing the national park rendering is outside the scope of this PR

matkoniecz:
> I prefer showing the parks too small to fit a label in

I also share this opinion, I think that this change makes situation pretty confusing.

matthijsmelissen:
I am re-opening this PR. I think the amount of small nature reserves is still a big problem. The number of places where rendering them looks good is greatly outnumbered by the number of places where they cause a problem.

screen shot 2017-12-07 at 00 39 11
screen shot 2017-12-07 at 00 38 28
screen shot 2017-12-07 at 00 37 37

matthijsmelissen:
Note that the change in zoomlevel has been merged before.

matthijsmelissen:
Can anybody review this (again)?

kocio-pl:
I think it is useful to start with that. I want to add boundary=protected_area later and use different zoom levels too, but that can take me some time.

Example (click to see the full size):

z8
Before
brtumbl6
After
p9lqeuvn

z9
Before
spyzctoa
After
m7e5qehp

kocio-pl:
Another example - near Warsaw:
Before
aq2tiysm
After
go7j1nzi

PR #2125, opened 2016-05-13
Pull in upstream default.style changes

pnorman:
These are all minor fixes which are backwards compatible

### DIFF
--- a/openstreetmap-carto.style
+++ b/openstreetmap-carto.style
@@ -1,5 +1,7 @@
 # This is the .style file for OpenStreetMap Carto, which is currently
 # the same as the upstream osm2pgsql style
+# phstore is used instead of polygon,nocolumn to preserve compatibility
+# with older osm2pgsql versions
 
 # OsmType  Tag          DataType     Flags
 node,way   access       text         linear
@@ -10,9 +12,9 @@ node,way   admin_level  text         linear
 node,way   aerialway    text         linear
 node,way   aeroway      text         polygon
 node,way   amenity      text         polygon
-node,way   area         text         # hard coded support for area=1/yes => polygon is in osm2pgsql
+node,way   area         text         polygon # hard coded support for area=1/yes => polygon is in osm2pgsql
 node,way   barrier      text         linear
-node,way   bicycle      text
+node,way   bicycle      text         linear
 node,way   brand        text         linear
 node,way   bridge       text         linear
 node,way   boundary     text         linear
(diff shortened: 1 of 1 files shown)


math1985:
Changes look fine.

I think this should be a PR against the 3.0 branch (which I now see I accidentally called 4.0):

> This project follows a MAJOR.MINOR.PATCH versioning system. In the context of a cartographic project you can expect the following:
> MINOR: [..] They [..] will not contain changes that require software or database upgrades.
> MAJOR: Any change the requires reloading a database, or upgrading software dependecies will trigger a major version change.

I don't think that in this case merging this in 2.0 will cause immediate issues, but still I think it would be better to keep the .style file in the current branch synchronised with the .style file as used on the OSMF servers.

pnorman:
> I think this should be a PR against the 3.0 branch (which I now see I accidentally called 4.0):

No, these changes are all backwards compatible. The poi column removal does not impact any style (since there are no `poi` tags, openstreetmap/osm2pgsql#568), the area change doesn't do anything since behavior is hard-coded, and the other changes have no change in meaning.

> I don't think that in this case merging this in 2.0 will cause immediate issues, but still I think it would be better to keep the .style file in the current branch synchronised with the .style file as used on the OSMF servers.

The file in this repo is not used on the OSMF servers.

math1985:
> The file in this repo is not used on the OSMF servers.

Is there any documentation on which file is used?

pnorman:
> Is there any documentation on which file is used?

https://github.com/openstreetmap/chef/blob/master/cookbooks/tile/templates/default/replicate.erb

I could remove the POI change from this PR, that's a new change that's not yet landed in a release

math1985:
> https://github.com/openstreetmap/chef/blob/master/cookbooks/tile/templates/default/replicate.erb

I don't see any reference to a .style file, does that mean that the default.style distributed with osm2pgsql is used?

math1985:
One thing I'm a bit worried about is that multipolygon-handling might lead to unexpected behaviour when importing with --append. I suspect this will be the same in the current situation and with the .lua-files, but I can't be sure.

For example, 
* What happens if an old-style multipolygon is appended with an outer element that was already in the database? When appending, does `keyvaluemembers` in `filter_tags_relation_member` contain all polygons in the relation or only the newly added ones?
* What happens if the multipolygon relation is dropped from an old-style multipolygon, leaving only the outer element? Currently, the outer element of the old-style multipolygon is ignored in the import. Will it be re-added if the element is removed from the multipolygon relation?

pnorman:
Lua isn't relevant to this PR since we're not using it, and this PR doesn't change that. It also won't change any MP handling since only `area` becomes polygon, and its behaviour is hard coded anyways.

math1985:
Woops, posted that in the wrong PR, should have gone in #2105.

pnorman:
> I don't see any reference to a .style file, does that mean that the default.style distributed with osm2pgsql is used?

Yes.

gravitystorm:
I'm happy to see this PR merged.

> but still I think it would be better to keep the .style file in the current branch synchronised with the .style file as used on the OSMF servers

It's not 100% clear what you mean by "synchronised" but it's up to us to make whatever changes are needed in the .style file, and it's up to the OSMF admins to deal with it :-)

I'll open up a bug report to show the OSMF admins that the style file for this style might diverge from whatever the osm2psql devs bundle with their software releases.

PR #2127, opened 2016-05-16
Database-reload

### LINKED ISSUE #59: Recommend -G flag for multipolygons

### LINKED ISSUE #101: ele with alpine_hut when mapped as a building

### LINKED ISSUE #137: highway=services is an area

### LINKED ISSUE #268: circular cliffs are not rendered

### LINKED ISSUE #892: man_made=embankment is not rendered if it's a closed way

### LINKED ISSUE #1362: failed display of closed way, unclosed way works (multiple affected tags)

math1985:
**This is a pull-request against the lua branch**

This is a new version of #2105. No changes except for [dropping some commented lines](https://github.com/gravitystorm/openstreetmap-carto/pull/2105#discussion_r63259976). 

This database import aims to be backwards compatible with older versions of
the style.

# Hstore
Adding the hstore option to osm2pgsql allows the database to use all keys.

This PR proposes using hstore for new keys, and using traditional columns for
old keys. This allows us to stay compatible with old versions of the style.

# Rendering order

This resolves #59.

# Import ele on buildings

This resolves #101.

The .lua file drops some more meta-information from imports that is of no
need for rendering.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -5,7 +5,7 @@ You need OpenStreetMap data loaded into a PostGIS database (see below for [depen
 It's probably easiest to grab an PBF of OSM data from [Mapzen](https://mapzen.com/metro-extracts/) or [geofabrik](http://download.geofabrik.de/). Once you've set up your PostGIS database, import with osm2pgsql:
 
 ```
-osm2pgsql -d gis ~/path/to/data.osm.pbf --style openstreetmap-carto.style
+osm2pgsql -G --hstore -d gis ~/path/to/data.osm.pbf --style openstreetmap-carto.style --tag-transform-script ~/osm/openstreetmap-carto.lua
 ```
 
 You can find a more detailed guide to setting up a database and loading data with osm2pgsql at [switch2osm.org](http://switch2osm.org/loading-osm-data/).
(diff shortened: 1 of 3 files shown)


math1985:
@gravitystorm Would you be able to do a test-run of the import on one of your servers? Doesn't need to be full-planet, but it would be good to have done at least some testing.

math1985:
Woops, pull request against the wrong branch, fixed in #2128.

dieterdreist:
sent from a phone

> Il giorno 16 mag 2016, alle ore 02:39, math1985 <notifications@github.com> ha scritto:
> 
> leisure=track, man_made=embankment, man_made=breakwater,
> man_made=groyne, natural=cliff, natural=tree_row, historic=citywalls,
> waterway=derelict_canal, waterway=ditch, waterway=drain, waterway=river,
> waterway=stream, waterway=wadi, waterway=weir, power=line, and power=minor_line
> are now treated as linestring.


is there an override (area=yes) to treat them as polygons where applicable?
I d expect leisure=track, man_made=breakwater, waterway=weir, man_made=embankment, historic=citywalls and maybe others, to be (sometimes) mapped (and valid) as polygons.

math1985:
> is there an override (area=yes) to treat them as polygons where applicable?

Yes, area=yes can be used to override these (just like for highway tags)

PR #2128, opened 2016-05-16
Database-reload

### LINKED ISSUE #59: Recommend -G flag for multipolygons

### LINKED ISSUE #101: ele with alpine_hut when mapped as a building

### LINKED ISSUE #137: highway=services is an area

### LINKED ISSUE #268: circular cliffs are not rendered

### LINKED ISSUE #892: man_made=embankment is not rendered if it's a closed way

### LINKED ISSUE #1362: failed display of closed way, unclosed way works (multiple affected tags)

math1985:
**This is a pull-request against the lua branch**

This is a new version of #2105. No changes except for [dropping some commented lines](https://github.com/gravitystorm/openstreetmap-carto/pull/2105#discussion_r63259976). 

This database import aims to be backwards compatible with older versions of
the style.

# Hstore
Adding the hstore option to osm2pgsql allows the database to use all keys.

This PR proposes using hstore for new keys, and using traditional columns for
old keys. This allows us to stay compatible with old versions of the style.

# Rendering order

# Recommend -G flag for multipolygons
This resolves #59.

# Import ele on buildings
This resolves #101.

The .lua file drops some more meta-information from imports that is of no
need for rendering.


math1985:
@gravitystorm Would you be able to do a test-run of the import on one of your servers? Doesn't need to be full-planet, but it would be good to have done at least some testing.

pnorman:
Layer should be an integer.

:-1: to having two z_orders, :-1: to trying to maintain compatibility

openstreetmap-carto.lua
```diff
@@ -0,0 +1,447 @@
...
+       match = match or string.find(k, d)
+     end
+     if match then
+       keyvalues[k] = nil
+       numberofkeys = numberofkeys - 1
+     end
+   end
+
+   -- Filter out objects with 0 tags
+   if numberofkeys == 0 then
```
pnorman: Move this branch first as it's a shorter exit and very common

openstreetmap-carto.lua
```diff
@@ -0,0 +1,447 @@
...
+      filter = 1
+      return filter, keyvalues
```
pnorman: For filtered objects
```lua
return 1, {}
```

openstreetmap-carto.lua
```diff
@@ -0,0 +1,447 @@
...
+   for k, v in pairs (keyvalues) do
+     match = false
+     for _, d in ipairs(delete_tags) do
+       match = match or string.find(k, d)
+     end
+     if match then
+       keyvalues[k] = nil
+       numberofkeys = numberofkeys - 1
+     end
+   end
```
pnorman: Is it faster if you move wildcards to their own list and do something like

```lua
for k, _ in pairs(keyvals) do
  if delete_tags[k]
    keyvals[k] = nil
  else
    for _, d in ipairs(delete_wildcards) do
    if string.find(k, d) then
      keyvals[k] = nil
      break
    end
  end
end

File: openstreetmap-carto.lua [hunk 4]
```diff
@@ -0,0 +1,447 @@
...
+
+-- Filtering on nodes, ways, and relations
+function filter_tags_generic(keyvalues, numberofkeys)
+   filter = 0   -- Will object be filtered out?
+
+   -- Delete tags listed in delete_tags
+   for k, v in pairs (keyvalues) do
+     match = false
+     for _, d in ipairs(delete_tags) do
+       match = match or string.find(k, d)
```
pnorman: even if not doing https://github.com/gravitystorm/openstreetmap-carto/pull/2128/files#r63304604 `break` after finding a match, e.g.
```lua
if string.find(k, d) then
  keyvalues[k] = nil
  numberofkeys = numberofkeys - 1
  break
end

File: openstreetmap-carto.lua [hunk 5]
```diff
@@ -0,0 +1,447 @@
...
+   -- Add z_order as key/value combination
+   keyvalues["osmcarto_z_order"] = osmcarto_z_order
+   keyvalues["z_order"] = z_order
+
+   return keyvalues, roads
+end
+
+-- Filtering on nodes, ways, and relations
+function filter_tags_generic(keyvalues, numberofkeys)
+   filter = 0   -- Will object be filtered out?
```
pnorman: make this a `local` variable if it's still needed

openstreetmap-carto.lua
```diff
@@ -0,0 +1,447 @@
...
+   end
+
+   return filter, keyvalues
+end
+
+-- Filtering on ways
+function filter_tags_way (keyvalues, numberofkeys)
+   filter = 0  -- Will object be filtered out?
+   polygon = 0 -- Will object be treated as polygon?
+   roads = 0   -- Will object be added to planet_osm_roads?
```
pnorman: `local` variables

pnorman:
The area determination logic is duplicated, put it in its own function

pnorman:
Can you edit the git commit message to have a more useful first line? This is a major change, so we'll be seeing it a lot in `git blame` and other commands.

``git commit --amend``, edit the message, and ``git push -f`` should do it

openstreetmap-carto.lua
math1985: I don't think we can do this, as line 257-267 might change the numberofkeys variable?

math1985:
> Can you edit the git commit message to have a more useful first line? 

Done.

I will come back to your other points later.

openstreetmap-carto.lua
pnorman: the `numberofkeys == 0` case is so common it needs to be handled first, even if another check is done at the end

openstreetmap-carto.lua
```diff
@@ -0,0 +1,447 @@
...
+-- If is_road=1, the object will be added to planet_osm_roads.
+zordering_tags = {
+   { 'railway', nil, 5, 0, 1},
+   { 'boundary', 'administrative', 0, 0, 1},
+   { 'bridge', 'yes', 10, 0, 0 },
+   { 'bridge', 'true', 10, 0, 0 },
+   { 'bridge', 1, 10, 0, 0 },
+   { 'tunnel', 'yes', -10, 0, 0},
+   { 'tunnel', 'true', -10, 0, 0},
+   { 'tunnel', 1, -10, 0, 0},
```
pnorman: these lines have an extra leading space

openstreetmap-carto.lua
```diff
@@ -0,0 +1,447 @@
...
+  'OBJTYPE',
+  'SK53_bulk:load'
+}
+
+
+-- Array used to specify z_order and osmcarto_z_order per key/value combination.
+-- The former is used for backwards compatibility and for uses that use a single
+-- database for multiple rendering styles.
+-- Each element has the form {key, value, z_order, osmcarto_z_order, is_road}.
+-- If is_road=1, the object will be added to planet_osm_roads.
```
pnorman: Is it better to have a table where the keys are two-element arrays of `{"key","value"}`?

This would allow lookup without a loop, but I'm not sure how to best include lookups that match a key regardless of value

openstreetmap-carto.lua
```diff
@@ -0,0 +1,447 @@
...
+  -- misc
+  'import',
+  'import_uuid',
+  'OBJTYPE',
+  'SK53_bulk:load'
+}
+
+
+-- Array used to specify z_order and osmcarto_z_order per key/value combination.
+-- The former is used for backwards compatibility and for uses that use a single
```
pnorman: the `z_order` you have in this table is not compatible with osm2pgsql z_order.

In osm2pgsql, z_order of links is [above highway=footway/cycleway/etc and below highway=motorway/trunk/etc](https://github.com/openstreetmap/osm2pgsql/blob/master/tagtransform.cpp#L23-L55) but you have `{'highway', 'motorway', 9, 380, 1},` and `{'highway', 'motorway_link', 9, 240, 1},`, putting them the same

openstreetmap-carto.lua
```diff
@@ -0,0 +1,447 @@
...
+-- Objects with any of the following keys will be treated as polygon
+polygon_keys = { 'building', 'landuse', 'amenity', 'harbour', 'historic', 'leisure',
+      'man_made', 'military', 'natural', 'office', 'place', 'power',
+      'public_transport', 'shop', 'sport', 'tourism', 'waterway',
+      'wetland', 'water', 'aeroway', 'abandoned:aeroway', 'abandoned:amenity',
+      'abandoned:building', 'abandoned:landuse', 'abandoned:power', 'area:highway' }
+
+-- Objects with any of the following key/value combinations will be treated as polygon
+polygon_values = {
+      {'highway', 'services'},
```
HolgerJeromin: Adding Idea from @jojo4u in https://github.com/gravitystorm/openstreetmap-carto/issues/137#issuecomment-124904987 
[rest_area](http://wiki.openstreetmap.org/wiki/Tag:highway%3Drest_area) should be added. 
We have 6342 ways with this tag and about 1/3 have area=yes on it.
This is currently [rendered](https://github.com/gravitystorm/openstreetmap-carto/blob/0c70682a69c91b833399c1dd74409eee79ce4af1/amenity-points.mss#L2026).

math1985: Thanks, added.

math1985:
> The area determination logic is duplicated, put it in its own function


Fixed

openstreetmap-carto.lua
math1985: Made all locally needed variables local.

openstreetmap-carto.lua
math1985: Done

openstreetmap-carto.lua
math1985: Fixed

openstreetmap-carto.lua
math1985: Done

openstreetmap-carto.lua
math1985: Done

openstreetmap-carto.lua
math1985: @pnorman Hmm, I'm having trouble with the lua syntax. I can't seem to use an array as index of a table. Example:

    > a={'b'}
    > table[a]={3}
    > print(table[{'b'}][1])
stdin:1: attempt to index field '?' (a nil value)
stack traceback:
	stdin:1: in main chunk
	[C]: in ?

What do I do wrong?

math1985:
> the z_order you have in this table is not compatible with osm2pgsql z_order.

We specify in the [documentation](https://github.com/gravitystorm/openstreetmap-carto/blob/master/INSTALL.md) that osm2pgsql should be run with openstreetmap-carto.style, so if the OSMF tile server maintainers decide to use a different .style, in a way that's their problem. In practice of course they are our main 'customer', so it's not something we can ignore. I wonder what other tile servers that run openstreetmap-carto do - use our style or use the osm2pgsql default one? For context, both styles were identical until osm2pgsql diverged a couple of months ago.

Basically we have three options:

1) Drop the column z_order altogether (we still keep osmcarto_z_order)
Positive: lower complexity (less chance for confusion due to the presence of two similar columns).
Negative: no backwards compatibility with older versions of the stylesheet; harder to share a database with other styles that use the z_order column such as openstreetmap-de and openstreetmap-fr.

2) Import z_order as currently defined by osm2pgsql and used on the OSMF servers (in addition to osmcarto_z_order)
Positive: Backwards compatibility (largely) with older versions of the stylesheet. Easy to share a database with other styles that use the default z_order column from osm2pgsql, and even (accepting some small changes in behaviour) with styles that use z_order column from openstreetmap-carto.style.
Negative: We might lose backwards compatibility anyway if we eventually decide to drop columns for which we will use hstore.

3) Import z_order as currently defined by openstreetmap-carto.style (in addition to osmcarto_z_order)
Positive: Backwards compatibility with older versions of the stylesheet.  Easy to share a database with other styles that use the default z_order column from openstreetmap-carto.style, and (accepting some small changes in behaviour) with styles that use z_order column from osm2pgsql.
Negative: Same as for 2.

Feel free to extend.

@gravitystorm Would you be able to make the final decision?

pnorman:
>> the z_order you have in this table is not compatible with osm2pgsql z_order.
> We specify in the documentation that osm2pgsql should be run with openstreetmap-carto.style, so if the OSMF tile server maintainers decide to use a different .style, in a way that's their problem. In practice of course they are our main 'customer', so it's not something we can ignore. I wonder what other tile servers that run openstreetmap-carto do - use our style or use the osm2pgsql default one? For context, both styles were identical until osm2pgsql diverged a couple of months ago.

The styles are still identical with respect to z_order - and z_order isn't defined in the .style file anyways.

INSTALL.md
```diff
@@ -5,7 +5,7 @@ You need OpenStreetMap data loaded into a PostGIS database (see below for [depen
 It's probably easiest to grab an PBF of OSM data from [Mapzen](https://mapzen.com/metro-extracts/) or [geofabrik](http://download.geofabrik.de/). Once you've set up your PostGIS database, import with osm2pgsql:
 
 ```
-osm2pgsql -d gis ~/path/to/data.osm.pbf --style openstreetmap-carto.style
+osm2pgsql -G --hstore -d gis ~/path/to/data.osm.pbf --style openstreetmap-carto.style --tag-transform-script ~/osm/openstreetmap-carto.lua
```
pnorman: should be `--tag-transform-script openstreetmap-carto.lua`

openstreetmap-carto.lua
pnorman: the syntax is `t[{foo="bar"}]="baz"` but won't work here because the table compairson checks for if they are the same object, not if the  contents of the table are the same

pnorman:
Notes from discussion today

* We're not going to be compatible with other styles built against default.style. They should be able to emulate compatibility with views if necessary.
* Columns with infrequent use that are not used for filtering (e.g. `ele`) will be dropped and we will rely on hstore
* Very common tags like `name` should have their own column, regardless of use for filtering. This saves space and improves performance.
* Columns like `surface` and `tracktype` are hard to decide on. They are not used for filtering, but it's not obvious if they're common enough or used enough to merit a column.

pnorman:
There are currently some unacceptable performance regressions - processing ways is up from 10s (C) or 31s (style.lua) to 99s (openstreetmap-carto.lua). I need to look at Lua performance profiling of transforms for other work I'm doing, so I'll do this to see where the slowness is.

pnorman:
We should add lua-based tests for openstreetmap-carto.lua behaviour tests, as they are far faster and easier to run than an .osm file and checking the results.

pnorman:
openstreetmap-carto.lua should use 4sp per indent, not 3sp

pnorman:
With the C transforms and osm2pgsql's style.lua example I have 28662 rows in planet_osm_roads with the extract I'm testing with, but 1667987 with this code, very close to the number of lines. Could 1 and 0 be flipped in what you've got? This would also account for some of the speed difference we found, since it has to do a lot more DB work

openstreetmap-carto.lua
```diff
@@ -0,0 +1,442 @@
...
+   {'railway', 'subway', 0, 420, 1},
+   {'railway', 'narrow_gauge', 0, 420, 1},
+   {'railway', 'light_rail', 0, 420, 1},
+   {'railway', 'preserved', 0, 420, 1},
+   {'railway', 'funicular', 0, 420, 1},
+   {'railway', 'monorail', 0, 420, 1},
+   {'railway', 'miniature', 0, 420, 1},
+   {'railway', 'turntable', 0, 420, 1},
+   {'railway', 'tram', 0, 410, 1},
+   {'railway', 'tram-service', 0, 405, 1},
```
pnorman: does this line work?

math1985: No - not sure now where it comes from actually.

openstreetmap-carto.lua
```diff
@@ -0,0 +1,442 @@
...
+
+}
+
+-- Array used to specify z_order and osmcarto_z_order per key/value combination.
+-- The former is used for backwards compatibility and for uses that use a single
+-- database for multiple rendering styles.
+-- Each element has the form {key, value, z_order, osmcarto_z_order, is_road}.
+-- If is_road=1, the object will be added to planet_osm_roads.
+zordering_tags = {
+   {'railway', nil, 5, 0, 1},
```
pnorman: considering dropping this as I believe all rail tags we use from roads are covered below

pnorman:
ping @math1985, I have two open PRs against this branch.

math1985:
I'm currently reviewing them.

math1985:
@pnorman's PR's have been merged.

From my side this is (again) ready for review/merge.

pnorman:
> From my side this is (again) ready for review/merge.

Have you fixed the problem with the roads table?

math1985:
> Have you fixed the problem with the roads table?

I must have missed that, what problem are you referring to?

> Also, this is failing CI

It failed because of 'The command "./scripts/travis_check_project_files" exited with 1.', that's a new check as far as I know, so I think this should fix itself when merging.

pnorman:
>> Have you fixed the problem with the roads table?
> I must have missed that, what problem are you referring to?

https://github.com/gravitystorm/openstreetmap-carto/pull/2128#issuecomment-223239183

> It failed because of 'The command "./scripts/travis_check_project_files" exited with 1.', that's a new check as far as I know, so I think this should fix itself when merging.

It looks like you didn't run yaml2mml.py

math1985:
> With the C transforms and osm2pgsql's style.lua example I have 28662 rows in planet_osm_roads with the extract I'm testing with, but 1667987 with this code, very close to the number of lines

I cannot reproduce this.

For Luxembourg, I get these numbers (old/new):
planet_osm_nodes 1572472/1572472
planet_osm_ways **201180**/**201180**
planet_osm_rels 3804/3804

planet_osm_point 45338/48484
planet_osm_line  **84918**/**85127**
planet_osm_polygon 115719/113992
planet_osm_roads **13481**/**13056**

@pnorman Can you reproduce this problem with the Luxembourg abstract?

math1985:
> It looks like you didn't run yaml2mml.py

Fixed.

pnorman:
> @pnorman Can you reproduce this problem with the Luxembourg abstract?

I'm having trouble reproducing, I'll try further debugging

PR #2134, opened 2016-05-28
topographic extensions

Ircama:
Added rendering of valleys, ridges and glaciers.
Revised zooms for some elements.
Revised openstreetmap-carto.style (and project.yaml) to support a future update to roads.mss (under dev now)
Added scripts\updatemml.py which makes project.mml compatible with Tilemill (run first scripts\yaml2mml.and then scripts\updatemml.py)
Added scripts\get-shapefiles.py which performs the same as get-shapefiles.sh, exploiting in Python 2.7


mboeringa:
Looking at a couple of the changes, especially the changes to zoomlevels for certain features, I think this PR needs proper visual validation in different parts of the world. While rendering some of the stuff like amenity=drinking_water from Z14 may be possible in sparsely populated mountain regions, some of these changes could add a lot of clutter in others, like urban areas. Simply changing symbol size (smaller) to forestall problems may sometimes work, but definitely not always...

HolgerJeromin:
This pr should be split into some pulls. 
The sac_scale changes are at least be implemented later and different after the style uses hstore

Ircama:
@mboeringa: we definitively need to ensure rendering compatibility with urban areas; nevertheless, the AS IS configuration needs many improvements to provide a decent quality also in mountain territories; I can anyway reconsider the original zoom for amenity=drinking_water and revise natural=spring instead.

@HolgerJeromin: I will revert my pool request and provide separate pools; one for zooms, another for management of valleys/ridges/glaciers, one for updatemml.py script. I then tested hstore and does work (tags->'sac_scale'::text AS sac_scale), so I’ll provide a project.yaml supporting hstore with a separate pool.

Ircama:
As per request, this branch has been substituted with #2136, #2137, #2138, #2139.
Please, let me know whether I have to close or remove it, thanks.

PR #2136, opened 2016-05-30
Topo extension updatemml

Ircama:
Added "scripts\updatemml.py" (to be executed just after scripts\yaml2mml.py) in order to update project.mml with the appropriate "dbname", "bounds" and "center" parameters for Tilemill.

E.g.,

Edit updatemml.py to update the "START CONFIG "/"END CONFIG" section with preferred bounds, wished default center and appropriate dbParms data.

run

scripts\yaml2mml.py && scripts\updatemml.py

after changing project.yaml in order to produce a project.mml file ready to be opened with Tilemill.

This script has been tested on Windows.

### DIFF
--- a/scripts/updatemml.py
+++ b/scripts/updatemml.py
@@ -1,4 +1,4 @@
-# updatemml, ircama, 2016
+# updatemml.py, ircama, 2016-5-30
 import re
 
 ####################### START CONFIG
--- a/scripts/updatemml.py
+++ b/scripts/updatemml.py
@@ -1,0 +1,56 @@
+# updatemml, ircama, 2016
+import re
+
+####################### START CONFIG
+dbParms = """\
+"dbname": "osm",
+        "user": "postgres",
+        "password": "postgres_007%",
+        "port": "5432",
+        "host": "localhost"\
+"""
+
+bounds = """\
+"bounds": [
+    9.8204,
+    45.7594,
+    10.2914,
+    46.0384
+  ]\
+"""
+
+center = """\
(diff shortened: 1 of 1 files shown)


pnorman:
I'm giong to close this because there are a few ways to do what you're trying to accomplish

* Launch Tilemill with [environment variables set](https://www.postgresql.org/docs/current/static/libpq-envars.html) which will tell it to connect how you specify, but doesn't allow overriding the database name.
* Use [Kosmtik](https://github.com/kosmtik/kosmtik), which allows setting [a file to override settings](https://github.com/kosmtik/kosmtik#local-config) like database information. It also allows you to directly use the yaml file and ignore the mml file.

Hope this helps

Ircama:
@pnorman: the hint to exploit the PostgreSQL variables is correct, we can close this PR.

These environment variables work in Windows when using 'setx' (defining variables with 'set' before invoking tilemill.exe might not work). Example:

setx PGHOST localhost
setx PGPORT 5432
setx PGUSER postgres
setx PGPASSWORD postgres_007%

The dbname created in PostgreSQL shall be "gis". Example:
psql -h localhost -U postgres -c "create database gis"
psql -h localhost -U postgres -c "\connect gis"
psql -h localhost -U postgres -d gis -c "CREATE EXTENSION postgis"
psql -h localhost -U postgres -d gis -c "CREATE EXTENSION hstore"

I doubt that it is viable to install Kosmitik with Win32 (at least I did not succeed so far).

PR #2137, opened 2016-05-30
Add scripts/get-shapefiles.py

### LINKED ISSUE #43: Download script errors on first run
When running with no previous files downloaded the errors

```
Warning: Illegal date format for -z, --timecond (and not a file name).
Warning: Disabling time condition. See curl_getdate(3) for valid date syntax.
```

are generated.

### LINKED ISSUE #66: Download scripts get stuck with invalid data
I ended up with a bad download and the download scripts didn't download new data because it was doing the date compare and didn't download

### LINKED ISSUE #1233: Incompatible unzip options

Ircama:
[Edited summary for this PR, left out on my first push]

The aim of this PR is to replace _get-shapefiles.sh_ with a Python version named _get-shapefiles.py_ and positioned in the _scripts_ directory of _openstreetmap-carto_.

It generates and populates the _data_ directory with all needed shapefiles, including indexing them through _shapeindex_. Related features are the same as the original shell script, with the addition of some command line options.

The code exploits an internal dictionary for easy adaptation to future changes, supports Windows, UNIX, Python 2.7, Python 3 and _curl_ when installed (even if _curl_ is not required, for more comfortable Windows support). Installation of _shapeindex_ is a prerequisite to index files, but is not strictly required if `-s` option is used.

The script can be run from the _scripts_ directory of _openstreetmap-carto_, or from its base folder, or through absolute path from any directory. It should allow a default Web proxy and also the configuration of the HTTP_PROXY environment variable. With Windows, PAC and NTLM authentication are not supported.

Documentation has been updated accordingly.

Resolves #43
Resolves #66
Resolves #1233

Basic invocation: `scripts/get-shapefiles.py` (or `./get-shapefiles.sh`)

```
Command-line options: get-shapefiles.py [-h] [-c] [-d <directory name>] [-e] [-f] [-l] [-n]
                          [-p] [-r] [-s] [-u] [-1] [-2] [-3] [-4] [-5] [-6]

Optional arguments:
  -h, --help            show this help message and exit
  -c, --check           check whether the 'data' directory already exists
  -d <directory name>, --directory <directory name>
                        set the name of the data directory (default: 'data')
  -e, --no-extract      do not populate target directories with the expansion
                        of downloaded data
  -f, --force           force continuing even if project.yaml does not exist
  -l, --no-curl         do not use 'curl' even if available
  -n, --no-download     do not download archive if already existing locally
  -p, --pause           pause before starting
  -r, --remove          remove each downloaded archive after its expansion
  -s, --no-shape        do not run shapeindex
  -u, --update          force downloading files even if not newer than the
                        locally existing ones
  -1, --world-boundaries
                        only process world_boundaries
  -2, --simplified-land
                        only process simplified-land-polygons-complete-3857
  -3, --ne-admin        only process ne_110m_admin_0_boundary_lines_land
  -4, --land-polygons   only process land-polygons-split-3857
  -5, --icesheet-polygons
                        only process antarctica-icesheet-polygons-3857
  -6, --icesheet-outlines
                        only process antarctica-icesheet-outlines-3857
```


scripts/get-shapefiles.py
```diff
@@ -0,0 +1,137 @@
...
+
+try:
+    import wget
+except ImportError:
+    sys.exit("""   Error: you need wget!
+   run 'pip install wget'.""")
+
+print "\nChecking shapeindex version:"
+try:
+    err = subprocess.call("bin/shapeindex -V", shell=False)
```
pnorman: We aren't creating a bin directory, so this should just be shapeindex

pnorman:
I wouldn't mind replacing `get-shapefiles.sh` with a python version, but not adding a second version which is to do the same thing. Switching to a python version would make it easier to download on Windows.

HolgerJeromin:
It would be nice to have a script which is 2.7 and 3 compatible. 
But this can be done later. 
http://python3porting.com/differences.html seems to be a good resource for that (But my last porting was a year ago).

Ircama:
@pnorman: In fact I am trying to manage implications of contributing to openstreetmap-carto through a Windows installation of Tilemill. One aspect that needed rewriting was get-shapefiles.sh

pnorman:
One of the important features of get-shapefiles.sh is you can re-run it to update

Ircama:
Added command line arguments.

Run it with `scripts\get-shapefiles.py`

To allow re-run it to update, use `get-shapefiles.py -qu`
To get help, `get-shapefiles.py -h`

 ```
usage: get-shapefiles.py [-h] [-q] [-u] [-n] [-e] [-s] [-1] [-2] [-3] [-4]
                         [-5] [-6]

This script generates and populates the 'data' directory with all needed
shapefiles.

optional arguments:
  -h, --help        show this help message and exit
  -q, --quiet       don't print status messages to stdout
  -u, --update      allow updating an already existing 'data' directory
  -n, --nodownload  do not download files if already existing locally
  -e, --noextract   do not populate target directories with downloaded data
  -s, --noshape     do not run shapeindex
  -1                only process world_boundaries
  -2                only process simplified-water-polygons-complete-3857
  -3                only process ne_110m_admin_0_boundary_lines_land
  -4                only process water-polygons-split-3857
  -5                only process antarctica-icesheet-polygons-3857
  -6                only process antarctica-icesheet-outlines-3857
```

**Comment by gravitystorm** [comment]
How hard is it to make the script do the following:

* Work on both python 2 and 3
* Only use modules from the standard library

Also, I can see there is inconsistencies in both variable naming (all lower (nodownload), snakecase (update_mode), other (Owps)) and indentation. I'm not a python programmer but best to go with whatever the python conventions are.

Finally, if this is a replacement for get-shapefiles.sh, then the PR should include removing the old version.

**Comment by Ircama** [comment]
@gravitystorm: OK, I'll try in the next days

File: scripts/get-shapefiles.py [hunk 2]
```diff
@@ -0,0 +1,204 @@
...
+                    help="don't print status messages to stdout")
+parser.add_argument('-u', "--update", dest='update_mode', action='store_true',
+                    help="allow updating an already existing 'data' directory")
+parser.add_argument('-n', "--nodownload", dest='nodownload', action='store_true',
+                    help="do not download files if already existing locally")
+parser.add_argument('-e', "--noextract", dest='noextract', action='store_true',
+                    help="do not populate target directories with downloaded data")
+parser.add_argument('-s', "--noshape", dest='noshape', action='store_true',
+                    help="do not run shapeindex")
+parser.add_argument('-1', dest='Owb',  action='store_true',
```
pnorman: variables like `--world-boundaries`, `--simplified-water`, `--ne-admin`, etc would be easier to remember than numerics

scripts/get-shapefiles.py
```diff
@@ -0,0 +1,277 @@
...
+        if args.option_swp:
+            subprocess.call("shapeindex --shape_files data/simplified-land-polygons-complete-3857/simplified_land_polygons.shp")
+            print()
+        if args.option_ne:
+            subprocess.call("shapeindex --shape_files data/land-polygons-split-3857/land_polygons.shp")
+            print()
+        if args.option_wps:
+            subprocess.call("shapeindex --shape_files data/antarctica-icesheet-polygons-3857/icesheet_polygons.shp")
+            print()
+        if args.option_aip:
```
tilmanb: option_aip controls the call to icesheet-**outlines**. I guess this is not intended ;-)

Ircama: @tilmanb: Thanks, I will correct this and will also change the water acronyms to land

scripts/get-shapefiles.py
```diff
@@ -0,0 +1,368 @@
...
+            pass 
+    else:
+        print("Indexing files...\n")
+    
+    if not args.option_no_shape:
+        step = int(step+0.9)
+        if args.option_wb:
+            print(str(step) + ". Indexing world_boundaries...")
+            step += 1
+            subprocess.call("shapeindex --shape_files data/world_boundaries/world_bnd_m.shp")
```
pnorman: The return value needs to be checked

Ircama: OK, will do (just notifying that get-shapefiles.sh did not have those controls)

pnorman:
I found the following issues running it

* The documentation needs updating to refer to the new name
* The script needs mode +x

I get the error

```
$ scripts/get-shapefiles.py

Shapeindex is found and has
   Error. [Errno 2] No such file or directory
```
* Updating should be the default and not require confirmation
* When I run the script with `scripts/get-shapefiles.py -su` and rerun it, it downloads all the files again. It should only download those which have changed. This takes about 15 seconds for me. Curl uses the `-z` option for this.
* The status update is substantially less useful than curl's. This isn't essential, but it'd be nice.

Code and data are intermixed and there's a lot of repeated similar code. The shell script has this too, but it's only 86 lines. I suggest moving the definitions of the files and their locations to the top in an array, dict, or other structure and having code which loops over that.

Ircama:
@pnorman: thanks for the feedbacks, I will setup a UNIX VM and test it there.

OK for the updates you asked.

I did not use curl to simplify the Windows setup.

> The documentation needs updating to refer to the new name

Would you refer to the *Scripted download* section in *INSTALL.md* or also something else?

pnorman:
> Would you refer to the Scripted download section in INSTALL.md or also something else?

Anywhere that get-shapefiles.sh is mentioned.

math1985:
@Ircama Is this ready for review?

Ircama:
@math1985 yes, thanks.

math1985:
@pnorman Do you have time to review this?

pnorman:
Yes, later this week.
On Jul 25, 2016 3:22 PM, Matthijs Melissen <notifications@github.com> wrote:@pnorman Do you have time to review this?

Ircama:
Memo for me: there is some change to be done to comply to PEP8 style guide. Will do ASAP with a new commit. Also a documentation typo needs fixing.

pnorman:
I still need to do a code review on this.

Ircama:
[Unwanted push, just rebased by now]

Ircama:
@pnorman: code now ready for review, thanks

INSTALL.md
```diff
@@ -21,13 +21,51 @@ psql -d gis -f indexes.sql
...
+  -1, --world-boundaries
+                        only process world_boundaries
+  -2, --simplified-land
+                        only process simplified-land-polygons-complete-3857
+  -3, --ne-admin        only process ne_110m_admin_0_boundary_lines_land
+  -4, --land-polygons   only process land-polygons-split-3857
+  -5, --icesheet-polygons
+                        only process antarctica-icesheet-polygons-3857
+  -6, --icesheet-outlines
+                        only process antarctica-icesheet-outlines-3857
```
pnorman: The full usage details shouldn't go here, just the command you need to run.

scripts/get-shapefiles.py
pnorman: The entire script was run with `set -e` so all return statuses were checked

scripts/get-shapefiles.py
```diff
@@ -0,0 +1,387 @@
...
+    import urllib2
+    import urlparse
+
+data_dir = 'data'
+settings = {
+    # Keys 1, 2, 3, ... set the arg short-options and the related process
+    # ordering. Use > 0 to allow processing.
+    1: {
+        'directory': 'world_boundaries',
+        'url': 'http://planet.openstreetmap.org/historical-shapefiles/world_boundaries-spherical.tgz',  # noqa
```
pnorman: What's the `noqa` comment for?

pnorman:
> The status update is substantially less useful than curl's. This isn't essential, but it'd be nice.

My recollection is that when I made this comment there was virtually no report on status. Now that there is we don't need curl, and having two different download implementations increases the code that needs testing.

These might be relevant: 
* http://stackoverflow.com/questions/22676/how-do-i-download-a-file-over-http-using-python/22776#22776
* http://stackoverflow.com/questions/2028517/python-urllib2-progress-hook

It looks like a simpler way to do progress reports.

I don't see value in keeping the -# short options for files, and they will break if we remove files. If others see a strong value in them I guess they can be kept.

I think we're getting close to being able to merge.

scripts/get-shapefiles.py
Ircama: #noqa special comment disables checks made by the pep8 script at line level: I only used it to confirm long URLs.

INSTALL.md
Ircama: Ok, I'll remove them. Also, they could be reported here https://github.com/gravitystorm/openstreetmap-carto/pull/2333

Ircama:
> I don't see value in keeping the -# short options for files, and they will break if we remove files. If others see a strong value in them I guess they can be kept.

Not sure I got this point, anyway they should not in my opinion, because everything is defined in the dictionary.

Ircama:
> It looks like a simpler way to do progress reports.

The approach is quite similar to what implemented; I just added some additional control.

>  two different download implementations increases the code that needs testing.

I used a script to compare the result of get-shapefiles.sh with get-shapefiles.py for all main options and download methods and I do not see bugs so far.

At the end it is just 387 lines in total.

scripts/get-shapefiles.py
Ircama: Yes, correct; now they should be checked anyway.

Ircama:
Rebased

Ircama:
Resolves #43

kocio-pl:
I guess this script resolves also https://github.com/gravitystorm/openstreetmap-carto/issues/66?

Ircama:
Rebased so that it can be easily reviewed/merged

Ircama:
Newly rebased

Resolves #66
Resolves #1233

HolgerJeromin:
If you place the "resolves" text in the first message of this PR the issues will be closed automatically.

kocio-pl:
I think you should also delete get-shapefiles.sh script in this PR.

Ircama:
> If you place the "resolves" text in the first message of this PR the issues will be closed automatically. 

@HolgerJeromin, @kocio-pl: thanks, I thought any comment was enough for the auto closure.

> I think you should also delete get-shapefiles.sh script in this PR.

Yes, thanks a lot. I have just fixed it with a new commit.

kocio-pl:
@pnorman Any problems left with this PR?

pnorman:
> @pnorman Any problems left with this PR?

the -1 to -6 short options. 

Aside from that, it seems to work.

I don't like the gain in code complexity, but it's probably worth it for windows support in the download script.

Ircama:
@pnorman - OK no problem; would you like to have the short options removed (keeping the long ones), or both the short and long options removed (with reference to -1 to -6)?

Just one note, as I see many sites mentioning the bash script, I’d like to double check that its removal is wanted.

pnorman:
> @pnorman - OK no problem; would you like to have the short options removed (keeping the long ones), or both the short and long options removed (with reference to -1 to -6)?

The long options are good to have.

> Just one note, as I see many sites mentioning the bash script, I’d like to double check that its removal is wanted.

We don't want to maintain two download scripts.

nebulon42:
I'm currently having problems with shapeindex on my Ubuntu 16.04. It tries to load the shared libraries of Ubuntu 14.04. So I can't test this right now.

I also think the short options -1 .. -6 are not the best choice.

kocio-pl:
@Ircama Double check is here, so do you plan to remove the short options? I would like to merge this PR soon, if possible.

Ircama:
Code updated as requested.

I meanwhile noticed that, when using `-s` to disable shapeindex (e.g., useful in case shapeindex is not properly installed), old indexes are not automatically removed. I did not have time to add this purge operation in the code now (and perform again all testing with all OSs); could be part of a future update. By now, it is better to manually remove indexes from the subfolders when updating the data directory without updating indexes. (Notice that running shapeindex is the default mode and also the suggested one.)

kocio-pl:
I just made some basic testing with Python 2/3 and shapeindex/--no-shape under Linux and it works.

I think that 6 months for one script is long enough and fine tuning can be done later. In the meantime we lack portability and the code waiting too long in the queue may need to be keep in sync, which is additional problem. Thanks for your code and patience, @Ircama!

Issue #1148, opened 2014-12-08
Rendering of natural=ridge


drgweber:
Ridges are important geographical objects which, to my knowledge, are not currently rendered in carto style 

They can range in length from 1 km to over 100 km and are commonly shown on printed maps in large upper case letters (see public domain image from IBGE, Serra=Ridge). Current usage is around 17000 http://taginfo.openstreetmap.org/tags/?key=natural&value=ridge#overview, but I guess usage would be higher if rendered. It also seems that natural=ridge was used in many cases instead of natural=mountain_range because the key ridge is formally defined while mountain_range is not.

In South America ridges are very important reference points defining whole regions, for example the Serra da Mantiqueira http://www.openstreetmap.org/way/266727831. Frequently, when planning trips the lack of rendering can be an important drawback since passing over ridges can be difficult or impossible.

While important ridges are commonly shown on printed maps, commercial digital maps (e.g. Bing, Google) usually do not show them. Therefore, this could be a nice opportunity to add a distinguishing feature to OSMs main rendering style.

My suggestion is to render at zoom levels smaller than 10, but perhaps this would depend on the length of the ridges.

Similar issue seems to be #1097

![image](https://cloud.githubusercontent.com/assets/9018415/5346375/79ef4522-7f06-11e4-967d-3efcf56d5980.png)

mboeringa:
I have now made an attempt to render natural=ridge. This actually turns out to be quite difficult. A solid line like for the one sided natural=cliff representing a steep decline, isn't a suitable representation. Neither is a simple dashed line, as they look to much like paths. The image below represents about the best I could come up with, trying to represent the not so prominent nature of this feature (click to enlarge).

The image also contains another novelty, the rendering of the natural=valley tag, which really adds something in combination with contour lines.

![ridge rendering example](https://cloud.githubusercontent.com/assets/7635726/5602530/85ff9e8c-9353-11e4-94b5-ae099ea80b23.jpg)

drgweber:
I think your first attempt looks very promising, but the ridges shown here seem to be quite small. 

How would larger structures be rendered? For example Serra da Moeda in Brazil:
http://www.openstreetmap.org/way/318209935#map=10/-20.3028/-43.9494

or an even larger one like Serra Geral de Goiás
http://www.openstreetmap.org/way/318565962

[...]

PR #2138, opened 2016-05-30
Rendering of valleys, ridges and other mountain elements

Ircama:
**Rendering of valleys, ridges and other mountain elements**

[Revised summary]

This PR introduces the rendering of the following essential mountain elements, which were previously unmanaged:
•natural=ridge (name and symbol)
•natural=arete (name and symbol)
•natural=dale (only name)
•natural=gorge (only name)
•natural=couloir (only name)
•natural=valley (only name)
•natural=mountain_range (only name)
•natural=massif (only name)

All elements names have appropriate and distinct colors, to help mappers in the correct representation of these features. Aretes and ridges are also rendered with proper symbols.


imagico:
A few points here:

* you should show some sample renderings demonstrating the effect of your changes
* you seem to be showing valleys/ridges with labels only - as mentioned in https://github.com/gravitystorm/openstreetmap-carto/issues/1148#issuecomment-66269359 this is problematic since it will likely result in frequent abuse of these tags for just placing labels
* also the rest of the discussion in #1148 might be useful to read - in particular the question how to treat natural=arete.
* you do not seem to modify the existing rendering of natural=glacier so it is unclear to me how your change is supposed to interact with that.

Ircama:
http://www.openstreetmap.org/#map=11/45.8020/9.9825
zoom11

http://www.openstreetmap.org/#map=14/45.9294/10.0484
zoom14

http://www.openstreetmap.org/#map=15/45.9222/10.0402
zoom15

mboeringa:
Not a critique of your work, but the rendering does highlight one huge caveat for rendering this type of "mountain" feature: it really needs height contours to really make sense of things like valley names, ridges and aretes, peaks etc. ...

Ircama:
@mboeringa: please apologize, I did not comment my previous post: the red ellipses are not really rendered, but manually edited by me (roughly) over each print-screen in order to highlight the names appearing with my suggested modification and that are not currently rendered in the related links that I posted.

@imagico: Thanks for your post, I’ll leave out natural=glacier and check once again #1148 before answering.

mboeringa:
> @mboeringa: please apologize, I did not comment my previous post: the red ellipses are not really rendered, but manually edited by me (roughly) over each print-screen in order to highlight the names appearing with my suggested modification and that are not currently rendered in the related links that I posted.

@Ircama : I fully understood that the red ellipses were just for highlighting, and wouldn't be rendered. I just wanted to point out that all of these mountain features actually need height contours to really make sense. Again, this is not a critique of your work on this rendering for valleys, which looks quite nice, but a general remark also relating to e.g. the current peak rendering already implemented.

Ircama:
@mboeringa: OK. The drop that I submitted for validation is confined to the rendering of tags like natural=valley and natural=ridge. I'm going to provide an update following received feedbacks.
If you are anyway referring to overlays like http://wiki.openstreetmap.org/wiki/Contours, this should be a separate implementation.

mboeringa:
> If you are anyway referring to overlays like http://wiki.openstreetmap.org/wiki/Contours, **this should be a separate implementation.**

Agree.

Ircama:
> you seem to be showing valleys/ridges with labels only - as mentioned in #1148 (comment) this is problematic since it will likely result in frequent abuse of these tags for just placing labels

Referring to this, I would agree with @mboeringa 's considerations in #1148. It is a reasonable risk and OSM already allows many ways to place generic tabs.

> all of these mountain features actually need height contours to really make sense

Ideally, a spatial interpolation of digital elevation model, rock polygons and specific linear tags (aretes, ridges, couloirs, etc.) could digitally address what cartographers manually draw to represent ridges on classic topographic maps; this should of course require a remarkable Mapnik upgrade, anyway beyond my capabilities. I'll just try to barely draw labels and symbols.

'natural=ridge' might not only mark gentle crests (that should not be sharp or rocky ridges anyway, for which natural=arete should be used instead), but could also cover named hillsides or relevant slopes, that should have a name but might not deserve a graphical symbol (just its name represented as inline text could be enough to show the related form of the mountainside area). Conversely, a (sharp) crest that could be (almost) precisely identified would need natural=arete.

That said, please, check this preliminary testing image: orange is a wide ridge without name (for all big ridges and for the ones, even small, without a name, a smooth symbol is represented), green is a small ridge with a name (and without symbol, as it is < 5 km and has a name), red is an arete (dolomitic shape of Presolana mountain) with a name (and its symbol, as all aretes have the symbol, regardless lenght and presence of name), blue is a cliff.zoom15arete
Do you find this appropriate?

I am trying at the end to render the following tags:

- natural=ridge (name and symbol, see example)
- natural=arete (name and symbol, see example)
- natural=dale (only name in specific color)
- natural=gorge (only name in specific color)
- natural=couloir (only name in specific color)
- natural=valley (only name in specific color)
- natural=mountain_range (only name, similar to valley, but with different color)
- natural=massif (only name)
At the moment the minimum zoom to possibly render them is 10.
natural=cliff will remain unchanged.

imagico:
> It is a reasonable risk 

Note this is not a risk, experience elsewhere tells that this will happen.  So the questions is if this is considered an acceptable side effect.  IMO not - it is the responsibility of this style due to its prominent exposure to try its best not to adversely affect mapping decisions even if this comes at the price of a less that ideal map quality in some cases.

> and OSM already allows many ways to place generic tabs

The most famous and most widely abused generic label tag currently is place=locality which likely has more abusive uses than justified ones.  However this is still much less prone to abuse since it is points only and only starts at z15.   But most importantly other label only point features (things like natural=bay etc.) are much easier to falsify so abuse can be spotted and fixed much better.

The wiki page for natural=valley currently also explicitly states label placement to be the purpose of the tag which is at odds with the general principles of OSM - i added a note there in that regard.

As said in #1148 i would support rendering ridges with lines and labels although it is of course not so easy to do this in a way that looks good considering the wide range of applications for natural=ridge (which includes many cases where the tag does not really apply).  Valleys OTOH i would only consider if the tag definition is sufficiently sharp and generic and mapping mostly complies with that definition.  

Don't get me wrong - your renderings look quite nice, especially considering the limitations of the tools used.  My objections are unrelated to the specific design.

These things aside your sample renderings show valley labels overlapping with other labels - this would need to be fixed.

Ircama:
> it is the responsibility of this style due to its prominent exposure to try its best not to adversely affect mapping decisions even if this comes at the price of a less that ideal map quality in some cases.

Maybe there are two different topics that could be separately adressed. Are you sure that the rendering process shall be able to effectively compensate the drawbacks of the unrestricted updating policies of OSM? I would think that its price will be too high.

> The wiki page for natural=valley currently also explicitly states label placement to be the purpose of the tag which is at odds with the general principles of OSM - i added a note there in that regard.

However I wonder whether this is limiting the quality of topographic data in OSM. To me OSM is currently unusable when representing mountains, as data is poor also for trivial usage. So I would consider appropriate to encourage the insertion of geographical data and long time will be possibly needed to fill the current huge gap.

For instance, we have 327943 named peaks now in OSM, but only 36644 ridges. Consider that one peak should have one or more named ridges, so there is a serious problem in OSM now. Also, we have only 5664 valleys: 47 peaks per valley means that at the moment this information has no quality.

Now, peaks are represented in Mapnik, while ridges and valleys aren't.

The policy behind this situation might have some responsibilities.

> Valleys OTOH i would only consider if the tag definition is sufficiently sharp and generic and mapping mostly complies with that definition.

Drawing valleys with a symbol is better than not rendering them at all, but IMHO might not be appropriate from the graphical point of view. For instance, please consider this old-school map I have just found in Internet:

http://satlavis.weebly.com/uploads/3/2/4/9/3249414/kompass_-_59_gruppo_sella-_marmolada.jpg

Produced in the late 70's with touristic targets, it still presents virtually unachievable quality when compared with the current version OSM/Mapnik.

http://www.openstreetmap.org/#map=14/46.4626/11.8443

And valleys are represented there in the correct way.

> These things aside your sample renderings show valley labels overlapping with other labels - this would need to be fixed.

What I am using now to test: 

```
text-placement: line;
text-placement-type: simple;
text-placements: "N,S,E,W,NE,SE,NW,SW,18,16,14,12,10,9,8";
text-label-position-tolerance:10;
//[zoom >= 11] { text-allow-overlap: true; }

```
If I keep the last line commented out, even if the remaining configuration looks pretty tolerant, the name of the valley might not be rendered for some zoom levels. And we might not love this of course :-) Please advise.

Anyway, in the Marmolada map that I linked, you can see that sometimes (rarely) labels overlap.

imagico:
Much of what you wrote about map design in general and OSM mapping practice goes beyond the scope of this issue tracker.  Discussion on how to map and tag valleys, ridges and other things should take place on the mailing lists and wiki.  Note my critique of the current definition of the valley tag does not mean i am against mapping valleys in principle.  And i am right with you regarding the problem that some types of features are enormously underrepresented in the OSM database.

> Are you sure that the rendering process shall be able to effectively compensate the drawbacks of the unrestricted updating policies of OSM?

I am not sure what you mean with _unrestricted updating policies of OSM_.

Regarding valley rendering in general - creating high quality labels would require tuning the label  configuration (position, direction, curving, letter spacing etc.) with consideration of the valley form (which is what would need to be mapped) and the other map elements and in interaction with other labels in the map.  Doing so would not only result in a nicer map, it would also significantly reduce the problem of mapping for the renderer since the label would not be shown exactly where the mapper placed the corresponding feature anyway.

This however is not possible at the moment in a real time updated tiled map.

> If I keep the last line commented out, even if the remaining configuration looks pretty tolerant, the name of the valley might not be rendered for some zoom levels. And we might not love this of course :-) Please advise.

Labeling consistency across zoom levels in this style leaves much to the desired.  In general a label that is rendered at one zoom level should also appear on all higher zoom levels unless labels for this kind of feature are removed in general at a certain level.  To achieve this label priorities need to be in the order of the zoom levels they appear in (which is currently not universally done in this style).

And labels with characters overlapping characters of other labels is a big no-no in cartography.  Labels may cross if characters do not intersect but most map rendering systems do not support this.  And avoiding overlaps is definitely more important than cross-zoom-level-consistency.

Ircama:
> you seem to be showing valleys/ridges with labels only - as mentioned in #1148 (comment) this is problematic since it will likely result in frequent abuse of these tags for just placing labels

Also following previous comments, am I correctly interpreting your suggestions for valleys?

- Points: `#text-point`, rendered in my drop, problematic for you, since it will likely result in frequent abuse of these tags for just placing labels
- Lines: `#text-line`, rendered in my drop, fine with you, so that a mapper shall appropriately draw the line in order to have its name rendered.
- Polygons: `#amenity-points-poly`, rendered in my drop, fine with you

I can remove `#text-point` in this case, let me know.

> avoiding overlaps is definitely more important than cross-zoom-level-consistency.

OK thanks, I'll set `text-allow-overlap: false;`

imagico:
> Also following previous comments, am I correctly interpreting your suggestions for valleys?

No, my suggestions for valleys are:

1. creating a consistent, well defined and universally applicable and verifiable mapping and tagging concept for valleys.  This is difficult since neither of the standard data structures of the OSM data model (nodes, linear ways and areas) are well suited for this task.  But it is doable.
2. convincing mappers to adopt this system.
3. making sure practical mapping adheres to this system (by communicating with mappers who use it incorrectly and establishing good practice for others to take as an example).
4. create a rendering proposal here that well represents and supports the mapping scheme.

This probably sounds like a lot of work and a long process but if 1. is done well 2. and 3. usually do not take long.  And with a well defined concept a few hundred features mapped by a dozen mappers are often sufficient to add something here if it integrates well with the rest of the map.

With ridges this process is already much further ahead - there is some distinct lack in 3. partly due to the lack of clear limits in the definition of a ridge (towards mountain ranges, arete and cliff in particular).  But here we are clearly at a point where rendering ridges (in a way that indicates a ridge and not just _something linear with a name_) could help to improve the data.

Ircama:
zoom16
zoom17

imagico:
I suggest you

* drop everything but ridge/arete in this PR to be able to concentrate on one thing.
* drop any complex and non-intuitive logic regarding the name.
* change the line_length field to a way_pixels equivalent (i.e. normalize with the rendering resolution), this is better readable and usually simplifies the mss code.
* reconsider the length limit for showing/not showing ridges - this would often lead to missing segments within a network of ridges.
* reconsider the starting zoom level - there is a reason why natural=cliff is not rendered below z=13.  At low latitudes at z=10 a ridge of less than 10km length makes no sense to display, even with a very compact styling - with your current styling the limit would be more around 30km and continuous ridges of this length are very rare - most cases with this length in the database are abuse of natural=ridge for mountain ranges.
* differentiate ridge/arete in a way that clearly communicates arete is a sharper feature
* at the higher zoom levels render with a clearly visible center line - does not have to be a continuous line or be very prominent but should allow mappers to verify if the way is placed at exactly the right position.
* drop labeling of polygon features for ridge/arete - neither should be mapped as areas so these should not be shown.
* look if you can improve visibility of paths/footways above ridges - this is fairly poor right now at the lower zoom levels.
* test in some other areas, ridge mapping is particularly popular in Russia and other former Soviet countries.  See for example the Caucasus mountains: http://overpass-turbo.eu/s/gAi

imagico:
Addition:

 i would probably also drop rendering ridges mapped as nodes.  Although in principle it is all right in OSM to map something as a node if you have no more detailed information this is not a common case in reality for ridges.  Nearly all of the nodes with the ridge tag in the database right now are from imports, mostly in Norway.  I don't think there is a gain from rendering these with labels.

Ircama:
I'm surprised. Why you want me to drop valleys? It was the starting point of my initiative...

imagico:
I am only suggesting not to take on too much at once.  So far none of the maintainers here have commented on your proposal but it is unlikely any of them is satisfied with everything in it so you significantly increase the chances of at least some of it making it into the map if you split it into smaller packages.

Also keep in mind that you can never reliably test your changes without a full planet database so it makes sense to first see how some if it works in the real whole world, learn from the results, consider feedback from users and refine your approach before taking the next step.

dieterdreist:
sent from a phone

> Il giorno 01 giu 2016, alle ore 23:54, Christoph Hormann <notifications@github.com> ha scritto:
> 
> This however is not possible at the moment in a real time updated tiled map.


we could mix them in from precomputed e.g. shapefiles (that get frequently updated, similar to the coastline process.) Valleys don't change that much...

dieterdreist:
sent from a phone

> Il giorno 01 giu 2016, alle ore 23:54, Christoph Hormann <notifications@github.com> ha scritto:
> 
> To achieve this label priorities need to be in the order of the zoom levels they appear in (which is currently not universally done in this style).


there's also an advantage in the way it is done now. The importance of labels changes at different scales (because the intention and context of the map change at different scales). 

Take Paris for example, in zoom6 the label is very important: http://www.openstreetmap.org/#map=6/48.951/2.362

also at zoom7 ;-)
http://www.openstreetmap.org/#map=7/48.951/2.362

(yes, there is a consistency problem in this case, because similar cities are labeled at this zoom)

But you don't want it in zoom13, because Paris is not a spot at this scale, it is everywhere:
http://www.openstreetmap.org/#map=13/48.8551/2.3496

(and indeed from zoom 14 on, the label disappears as it should)


Similarly you might want a big building labeled early, but at high zooms you would give more priority to the constituents and put the label for the bigger part wherever space remains, if at all.

When the thing being labeled extends significantly the size of the screen that you look on, then a label can be more confusing than helpful.

Ircama:
> there's also an advantage in the way it is done now. The importance of labels changes at different scales (because the intention and context of the map change at different scales).

your consideration on zooms is clear and I had it in mind; I anyway think that I implemented an appropriate configuration for these mountain elements, which would need such rendering for labels; in my point of view the defined settings are valid for each zoom and I of course compared the results many times with standard topographic maps.

> change the line_length field to a way_pixels equivalent (i.e. normalize with the rendering resolution), this is better readable and usually simplifies the mss code.

I considered this and think that line_length is an appropriate parameter for these elements, where the rendering shall be based on the real length. If you have a way_pixels equivalent (working for lines) that behaves the same as line_length but is more performant, please advise.

> reconsider the length limit for showing/not showing ridges - this would often lead to missing segments within a network of ridges.

I reduced the min length to 2.5 km. Consider that there might be cases where for a single peak e.g. 10 or more small ridges are present; this might lead to clog the mountain with too many symbols.

> reconsider the starting zoom level - there is a reason why natural=cliff is not rendered below z=13. At low latitudes at z=10 a ridge of less than 10km length makes no sense to display, even with a very compact styling - with your current styling the limit would be more around 30km and continuous ridges of this length are very rare - most cases with this length in the database are abuse of natural=ridge for mountain ranges.

Please check the samples.

> differentiate ridge/arete in a way that clearly communicates arete is a sharper feature

I revised the symbol

> at the higher zoom levels render with a clearly visible center line - does not have to be a continuous line or be very prominent but should allow mappers to verify if the way is placed at exactly the right position

I revised the symbol

> drop labeling of polygon features for ridge/arete - neither should be mapped as areas so these should not be shown

By now I changed the rendering zoom setting it to >= 14

> I would probably also drop rendering ridges mapped as nodes. Although in principle it is all right in OSM to map something as a node if you have no more detailed information this is not a common case in reality for ridges. Nearly all of the nodes with the ridge tag in the database right now are from imports, mostly in Norway. I don't think there is a gain from rendering these with labels.

There might be some small ridges anyway that someone might quickly represent as nodes, in order to promptly provide a locally acquired name; representing them at high zoom level contributes enriching data from one side, and from the other helps mappers when subsequently converting nodes to lines, so that nodes will not remain there, other than lines.

Please find below some examples based on the geographic area indicated by imagico.

s9
s10
s11
s12
s13
s14
s15
s16
This is an arete:
s16a

dieterdreist:
sent from a phone

> Il giorno 03 giu 2016, alle ore 04:02, Ircama <notifications@github.com> ha scritto:
> 
> your consideration on zooms is clear and I had it in mind; I anyway think that I implemented an appropriate configuration for these mountain elements, which would need such rendering for labels; in my point of view the defined settings are valid for each zoom and I of course compared the results many times with standard topographic maps.
> 


I believe you misread this, it was a reply to Christoph against his idea to strictly keep rendering priority throughout all zoom levels, while I believe that different zoom levels can have different priorities for things because those different scales serve different purposes.

imagico:
Thanks for the examples.  For me they confirm that z10 is too early to start displaying ridges - they are mostly noise at this scale.  Keep in mind this is already 43 degrees latitude.

> If you have a way_pixels equivalent (working for lines) that behaves the same as line_length but is more performant, please advise.

The equivalent would be 

`ST_Length(way)/NULLIF(!pixel_width!::real,0) AS way_pixels`

This would allow you to make equivalent labeling decisions at different zoom levels with a single rule.

> I reduced the min length to 2.5 km.

It does not matter what length limit you use - you will still have segments missing heavily distorting the results.  You can see that in your example in the area of my overpass link and along the main watershed quite well.

> I revised the symbol

I probably did not explain myself sufficiently here.  I think the pattern for arete needs to be significantly more articulated, less blurry, more line-like than for ridge.  The side structures in the arete pattern by the way are a bit problematic since they are directed indicating the direction has meaning (which it does not).

And both patterns should IMO have a visible center line at the higher zooms that allows the viewer to locate the way location within a pixel accuracy.

In general based on your new examples the pattern appears a bit noisy, especially at the low zooms.  It might be a good idea to design it as SVG - you could use [jsdotpattern](http://www.imagico.de/map/jsdotpattern.php) for the base pattern and modify it with inkscape.  This gives you better control on the sub-pixel level than pixel painting it.

> There might be some small ridges anyway that someone might quickly represent as nodes

As said if you look at the data this is not a practically common case.  A mapper who can identify a ridge generally is also able to create a way instead of a node.  It is good if the style urges him/her to do so.

dieterdreist:
2016-06-02 22:24 GMT+02:00 Martin Koppenhoefer <dieterdreist@gmail.com>:

> But you don't want it in zoom13, because Paris is not a spot at this
> scale, it is everywhere:
> http://www.openstreetmap.org/#map=13/48.8551/2.3496
>


I have to adjust this, it really depends on your screen size here, only on
a phone the above is correct, on a full-hd resolution it is justified to
have a label for Paris at z13 (and not on z14 and higher - as it is now
already).

Ircama:
> z10 is too early to start displaying ridges

IMHO it is fairly decent as a first rendering attempt. I also tried to check with a satellite image and different renderers.
g1 
g4
g2
g3
Anyway a feedback from the author who defined these elements would be the most influential, because he might know these territories very well and he could also have experience on how they are represented in their standard cartography.

> ST_Length(way)/NULLIF(!pixel_width!::real,0) AS way_pixels

OK, I see, you are still using ST_Length, as I did. I’m afraid anyway that the ratio you added is not what we need, as in such mode way_pixels will be relative to the zoom level; we need absolute values instead for quality rendering of these elements.

> It does not matter what length limit you use - you will still have segments missing heavily distorting the results. You can see that in your example in the area of my overpass link and along the main watershed quite well.

Maybe I did not realize here. Please explain better with a graphical example. Which are the segments you find missing and where do you find distorted results?


> the pattern for arete needs to be significantly more articulated

Mine would be a starting point to render these objects and above all to finally allow showing a name for them.
Stat Roma pristina nomine, nomina nuda tenemus.
I’m sure that there are a lot of graphic designers willing to contribute, in order to enhance these patterns.
Anyway, IMHO a graphical distinction between ridges and aretes is already present at the moment; the fact that it is not too much emphatic was by design (don’t get me wrong, I’m not saying that it does not deserve an improvement!); also its center definition should be pretty visible at higher zooms (how sharp should be this arete?).

> A mapper who can identify a ridge generally is also able to create a way instead of a node.

Also in this case, having feedbacks from many commenters would be useful to take a decision. My stating consideration is that in a freedom map like OSM we expect many (or even too many) elements, not such huge lack of information, that makes it practically useless so far, apart from mapping significantly populated places (like cities and suburban areas). Not to say that OSM has “street” within its acronym, not e.g. planet.

project.yaml
```diff
@@ -2115,12 +2117,13 @@ Layer:
             name,
             operator,
             ref,
+            ST_Length(way) AS line_length,
```
pnorman: Don't use line length, use line length relative to zoom, or what @imagico indicated earlier. For better non-mercator compatibility, use `ST_Length(way)/NULLIF(SQRT(!pixel_width!::real*!pixel_width!::real),0)`

project.yaml
```diff
@@ -2099,6 +2100,7 @@ Layer:
     properties:
       minzoom: 10
     advanced: {}
+# Ircama: added support of ridges, valleys and glaciers in text-line
```
pnorman: comment needs removal

landcover.mss
```diff
@@ -689,18 +710,151 @@
...
 
+@ridge-text:          #7A2F18;
+@valley-text:         #305040;
+@mountain_range-text: #C03A01;
+@massif-text:         #C03A01;
+@cliff-text:          #999;
+@arete-text:          #800000;
+@couloir-text:        #0080C0;
+@dale-text:           #008000;
+@gorge-text:          #800040;
```
pnorman: Move colour variables to the section at the top of the file

pnorman: It feels like there might be too many colours here - some of them should probably be grouped under a common name

landcover.mss
```diff
@@ -626,6 +626,27 @@
 }
 
 #cliffs {
+  [natural = 'arete'][zoom >= 10] {
+    line-pattern-smooth: 0.3;
+    line-pattern-file: url('symbols/arete.png');
+    [zoom >= 13] {
+      line-pattern-file: url('symbols/arete3.png');
```
pnorman: Why order the pictures 1, 3, 2?

pnorman:
I'm just doing a technical code review for now and will consider the cartography later

* A bunch of lines have indentation problems

landcover.mss
```diff
@@ -626,6 +626,27 @@
 }
 
 #cliffs {
```
pnorman: Because this layer is no longer just cliffs, the name should be changed

Ircama:
> use line length relative to zoom

please, explain why this is beneficial, thanks in advance. A valley that is more than 10 km in length is there with its absolute value, regardless the zoom level.

imagico:
> Which are the segments you find missing and where do you find distorted results?

Just remove the cutoff and compare the results with those with cutoff and you will see - not only individual small ridges but whole groups of them are missing.

In general keep in mind ST_Length() is in mercator meters, not in real meters.  So even if you think there is something inherently different about a ridge longer than a certain limit (which i don't see) this cannot be implemented with an ST_Length() limit.

imagico:
> Does this mean that this will return its real length (not the Mercator one) and also unrelated with the specific zoom level?
ST_Length(way)/NULLIF(SQRT(!pixel_width!::real*!pixel_width!::real),0)

No, this produces the length in pixels on screen, which is exactly what you should base labeling decisions on.

You can find missing ridges in most of your Caucasus examples - for example in 

https://cloud.githubusercontent.com/assets/8292987/15766487/5bb5a9ea-293f-11e6-9916-814ec23d6250.png

there is a gap in the boundary ridge, in 

https://cloud.githubusercontent.com/assets/8292987/15766461/1e346bc4-293f-11e6-98a2-86c02328bbd7.png

the ridges between the peaks above the glaciers at the bottom of the image are missing.

Ircama:
> Don't use line length, use line length relative to zoom, or what @imagico indicated earlier. For better non-mercator compatibility, use ST_Length(way)/NULLIF(SQRT(!pixel_width!::real*!pixel_width!::real),0)


Just trying to find out a viable way to use 

`ST_Length(way)/NULLIF(SQRT(!pixel_width!::real*!pixel_width!::real),0) as line_length
`

To analyze the rendering, other than updating project.yaml, I temporarily added this debug line to landcover.mss
`    #text-line { text-name: "'L'+[line_length]+' '+[name]"; }
`

What happens is that I cannot figure out a single appropriate parameter to distinguish long valleys from dales at any zoom.

At the end, what I wish to obtain is that when a valley is longer than e.g. 20 km (e.g, one big valley per zone), regardless the zoom (of course this logic should be valid at any zoom), that valley should deserve a special rendering (visibility) vs. the many small ones that a mountain could have (e.g, >5 per mountain). In order to accomplish this with this zooming dependent line_length, I need to define specific thresholds per each zoom level (and not a fixed threshold as I did in my current commit, e.g. `[line_length>20000]`). This should produce a code which maybe is a bit more difficult to understand/maintain (many different parameters, one per each zoom level), but if it looks like an appropriate approach, I’ll proceed this way.

Please let me know, thanks again.

imagico:
The concept of a valley is essentially scale independent - the largest features commonly considered valleys in a wider sense are more than 500 km in size (Nile Valley, Great Rift Valley, California Central Valley etc.) - the smallest are just a few hundred meters in size, even if you consider gullies etc. not to fall in that category (but that is currently a non-verifiable distinction).  Between these there is a continuum of cases with no particular scale standing out or missing.  This means a specific size cutoff in rendering that is independent of the zoom level, either in styling or in showing/not showing, especially when done in mercator meters, is not appropriate.

In the context here all this would of course only be significant if there was an observable, verifiable and adhered to definition or natural=valley in OSM (which there currently is not).

landcover.mss
Ircama: That's right. I might call this *reliefs*, or better *reliefsymbols*. Not being my mother tongue, I would greatly appreciate advises here, thanks.

matkoniecz:
I wanted to check how this rendering works, discovered that valleys are unmapped in my region and then discovered that documentation how valleys should be mapped is missing. To avoid offtopic here - I posted about problems that I spotted at http://wiki.openstreetmap.org/wiki/Talk:Tag:natural%3Dvalley

Ircama:
> reconsider the length limit for showing/not showing ridges - this would often lead to missing segments within a network of ridges

OK. After some tests and research, I would think your recommendation is appropriate

> differentiate ridge/arete in a way that clearly communicates arete is a sharper feature

I did some improvement

> drop labeling of polygon features for ridge/arete - neither should be mapped as areas so these should not be shown.
> drop rendering ridges mapped as nodes

My proposal would consider rendering them at zoom level >= 15 for debugging purpose, se that mappers will notice them at high zooms and can modify their definition to ways. If you anyway do not agree on this, I will definitively leave out nodes and polygons.

> look if you can improve visibility of paths/footways above ridges - this is fairly poor right now at the lower zoom levels.

I will propose a revision of tracks and paths in the future, to reflect symbols in use with legacy topographic maps (using sac_scale and ref)

> Don't use line length, use line length relative to zoom

Implementing this, OK

> A bunch of lines have indentation problems

Fixed

> Because this layer is no longer just cliffs, the name should be changed

Renamed to reliefsymbols

I also added `place=region` & `region:type=mountain_area`, to complete the set of managed tags. This also implies setting some specific zooming set.

Still testing everything for sometime before sending a new PR.

Ircama:
Update including modifications to reflect previous recommendations.

The following elements are now rendered:

- natural=ridge (name and symbol)
- natural=arete (name and symbol)
- natural=dale (only name)
- natural=gorge (only name)
- natural=couloir (only name)
- natural=valley (only name)
- natural=mountain_range (only name)
- natural=massif (only name)
- place=region & region:type=mountain_area (only name)

It has been extensively tested on different areas.

project.yaml
```diff
@@ -111,16 +111,18 @@ Layer:
...
         (SELECT
             way, name, way_pixels,
-            COALESCE(wetland, landuse, "natural") AS feature
+            COALESCE(wetland, landuse, "natural", mountain_area) AS feature
           FROM (SELECT
               way, COALESCE(name, '') AS name,
               ('landuse_' || (CASE WHEN landuse IN ('forest', 'military') THEN landuse ELSE NULL END)) AS landuse,
               ('natural_' || (CASE WHEN "natural" IN ('wood', 'sand', 'scree', 'shingle', 'bare_rock') THEN "natural" ELSE NULL END)) AS "natural",
               ('wetland_' || (CASE WHEN "natural" IN ('wetland', 'mud') THEN (CASE WHEN "natural" IN ('mud') THEN "natural" ELSE wetland END) ELSE NULL END)) AS wetland,
+              ('place_'   || (CASE WHEN place = 'region' AND tags->'region:type'::text = 'mountain_area' THEN 'mountain_area' ELSE NULL END)) as mountain_area,
```
pnorman: osm-carto does not use hstore, there is no tags column.

Ircama:
What do you suggest in cases where additional columns shall be introduced?
1.	Keep on using hstore, but do not merge and just wait for hstore to be rolled-out
2.	Ask for new columns to be included in *openstreetmap-carto.style* and reference them instead of hstore
3.	Temporarily comment out *tags* references in *project.yaml* until hstore is rolled-out (the PR will enable partial features and will have to be followed by a new PR to enable hstore references, performed after its release)
4.	The PR can be merged and will not produce errors, but the *tags* references will simply not work until hstore is rolled out
Thanks

pnorman:
> What do you suggest in cases where additional columns shall be introduced?

We are not doing any changes at this time which require columns other than those currently listed in openstreetmap-carto.style. See #1504 for background, #2128 for current work

> 4. The PR can be merged and will not produce errors, but the tags references will simply not work until hstore is rolled out

This PR would cause errors, as the SQL refers to a column that does not exist.

Ircama:
@pnorman

> Don't use line length, use line length relative to zoom, or what @imagico indicated earlier. For better non-mercator compatibility, use 
> ```
> ST_Length(way)/NULLIF(SQRT(!pixel_width!::real*!pixel_width!::real),0)
> ```

Wouldn't you mean this perhaps?

```
ST_Length(way)/NULLIF(SQRT(!pixel_width!::real*!pixel_height!::real),0) AS way_size
```

or

```
ST_Length(way)/NULLIF(SQRT(!pixel_width!::real*!pixel_width!::real+!pixel_height!::real*!pixel_height!::real),0) AS way_size
```

or

```
ST_Length(way)/NULLIF(!pixel_width!::real*!pixel_height!::real,0) AS way_size
```

Thanks

pnorman:
> Wouldn't you mean this perhaps?

> ```
ST_Length(way)/NULLIF(SQRT(!pixel_width!::real*!pixel_height!::real),0) AS way_size
```

This is the same calculation, just given a name. I'd go with `way_length` to be consistent with ST_Area vs way_area.

> ```
ST_Length(way)/NULLIF(SQRT(!pixel_width!::real*!pixel_width!::real+!pixel_height!::real*!pixel_height!::real),0) AS way_size
```

This is off by a factor of sqrt(2) for web mercator

> ```
ST_Length(way)/NULLIF(!pixel_width!::real*!pixel_height!::real,0) AS way_size
```

This is off by a factor of `!pixel_width!` or `!pixel_height!`, and the units don't work out.

sqrt(x*y) is the [geometric mean](https://en.wikipedia.org/wiki/Geometric_mean) of x and y. The alternative would be the arithmetic mean of x+y, but the geometric mean is what you want.

**Comment by Ircama** [comment]
> This is the same calculation

Just to understand

SQRT(!pixel_width!::real*!pixel_width!::real)

vs.

SQRT(!pixel_width!::real*!pixel_height!::real)

I'd say the last one for the geometric mean, correct?

**Comment by pnorman** [comment]
> This is the same calculation
> Just to understand
> 
> SQRT(!pixel_width!::real*!pixel_width!::real)
> 
> vs.
> 
> SQRT(!pixel_width!::real*!pixel_height!::real)
> 
> I'd say the last one for the geometric mean, correct?

Ah, yes, I missed a height in my original comment

File: project.yaml [hunk 7]
```diff
@@ -673,21 +675,21 @@ Layer:
...
+    name: "reliefsymbols"
     class: ""
     geometry: "linestring"
     <<: *extents
     Datasource:
       <<: *osm2pgsql
       table: |-
         (SELECT
-            way, "natural", man_made
+            way, "natural", name, ST_Length(way)/NULLIF(SQRT(!pixel_width!::real*!pixel_width!::real),0) AS pixel_length, man_made
```
StyXman: `pixel_length` is NULL for zoom level 15 at 43°30'. a) is this right? b) shouldn't it be `!pixel_width!::real*!pixel_height!::real`

StyXman:
As for the symbology, maybe a zigzag line, a harsh one for aretes and a softer one for ridges?

Ircama:
> This PR would cause errors, as the SQL refers to a column that does not exist.

OK, I'm removing hstore usage

> pixel_width!::real*!pixel_height!::real ?

Yes, clarified

> As for the symbology, maybe a zigzag line, a harsh one for aretes and a softer one for ridges?

I'll try to simplify the symbols

StyXman:
My humble attempt:

http://www.grulic.org.ar/~mdione/arete.svg

http://www.grulic.org.ar/~mdione/ridge.svg

Ircama:
@StyXman: thanks will test as soon as I have enough bandwidth (which is difficult for me during these days)

math1985:
@pnorman Do you have time for a new review of this PR?

Ircama:
> My humble attempt:
> http://www.grulic.org.ar/~mdione/arete.svg
> http://www.grulic.org.ar/~mdione/ridge.svg

@StyXman: did you test them as svg or converted to png? e.g. **line-pattern-file: url('symbols/arete.svg');** does not look working (I suspect there is some issue with SVG line patterns).

pnorman:
> @pnorman Do you have time for a new review of this PR?

I've only done technical reviews, not cartography. The previous comments (e.g. hstore usage) have not been addressed.

I'm not sure we should render these features. Without hillshading, we will never be a good topographic map and it's not one of our goals.

Ircama:
> I'm not sure we should render these features. Without hillshading, we will never be a good topographic map and it's not one of our goals.

Regardless hillshading, text names of **natural** features would be really important information which aren’t currently rendered and that I suggest to address.

Maybe line patterns for valleys and ridges could only be kept at zoom >=17 to allow mappers to verify whether these ways are placed at the right position (or even left out and applied on a subsequent PR).

If there are not additional remarks, I can revise this PR accordingly.

imagico:
> I'm not sure we should render these features. Without hillshading, we will never be a good topographic map and it's not one of our goals.

This on its own is no significant argument IMO - there are plenty non-shading techniques for rendering topography and some of them (like peaks, cliffs) are already rendered.

My suggestions made above still stand - in particular the objection to label-only rendering and the suggestion to limit this PR to arete and ridge (which i think has potential to look nicely in the map).

StyXman:
@Ircama : here's the code I used:

#reliefsymbols {
  [natural = 'arete'][zoom >= 12] {
    line-pattern-smooth: 0.3;
    line-pattern-file: url('symbols/arete.svg');
  }
}

My stack includes Mapnik 3.0.11 and Cairo 1.14.6. I'm not sure who renders the SVGs, but must be one of those.

Ircama:
@StyXman: using .png in *line-pattern-file* will allow Tilemill compatibility.

-----------------
**Latest PR** including the following changes:
- Removed hstore usage
- Revised way_length calculation
- Improved symbology

Please find here below some sample renderings demonstrating the effect of your changes:

Rendering of valleys, ridges and other mountain elements in an Italian region at different zoom levels:
z13
z14
z15
z16

Ridges in Georgia at different zoom levels:
z11g
z12g
z13g
z14g
z15g
z16g

StyXman:
@Ircama `.svg`s render fine in Tilemill, at least in current master. It's true that the latest tag is from 2012 (!!!), maybe you should update?

Ircama:
@StyXman: thanks for your clarification, I'm using possibly the latest available drop for Win32, which should be http://tilemill.s3.amazonaws.com/dev/TileMill-v0.10.1-291-g31027ed-Setup.exe. (I verified that Kosmtik with Ubuntu works; I was not able to update TileMill on Win32 or to install node-mapnik.)

Ircama:
Rebased

Ircama:
taghistory 1

kocio-pl:
General question: are there unexpected problems with merging stuff or just lack of time to take care of the incoming code? This PR is pretty wide and important, but it's been polished according to remarks and it's quite a long time since it seems to be ready.

pnorman:
I am leaning against it for cartographic reasons, so haven't done a final review.

Ircama:
> I am leaning against it for cartographic reasons, so haven't done a final review.

I agree that new features shall be appropriately weighed up.

Would you please elaborate the possible cartographic drawbacks that such PR might introduce? In case, we can rework it.

kocio-pl:
@pnorman
> I'm not sure we should render these features. Without hillshading, we will never be a good topographic map and it's not one of our goals.

It would be good to decide which features make no problems and which are problematic somehow, for a start. This PR could be then split into parts and not hold everything stalled.

For example ridge rendering is good to have anyway IMO. Hillshading would just add more details how the terrain is shaped, but still would not replace rendering ridge name. These features are only partially overlapping.

pnorman:
Yes, we need to figure out [which need height contours to make sense and which don't.](https://github.com/gravitystorm/openstreetmap-carto/pull/2138#issuecomment-222603174)

kocio-pl:
@Ircama For me rigde is a perfect candidate for merging, because it doesn't need contours, but I'm not that involved in this topic to say about all the other features. Maybe you could write down your opinion first, so the rest can refer to it - that will probably the easiest thing to do to reach decision stage.

dieterdreist:
2016-11-07 17:14 GMT+01:00 kocio-pl <notifications@github.com>:

> @Ircama <https://github.com/Ircama> For me rigde is a perfect candidate
> for merging,



if we are discussing this list:

•natural=ridge (name and symbol)
•natural=arete (name and symbol)
•natural=dale (only name)
•natural=gorge (only name)
•natural=couloir (only name)
•natural=valley (only name)
•natural=mountain_range (only name)
•natural=massif (only name)

I believe none of them needs contour lines (equal elevation profile lines)
in order to make sense. They can all be seen as natural place names, and
only for ridge and arete a symbol is proposed (those are both similar
features, if ridge makes sense, arete will as well)

Ircama:
Thanks a lot for resuming comments. I understand that we need to be generally cautious when adding new features and I appreciate the intention of performing researches and reviews before taking a decision.

Like other physical elements already available in this style, I am convinced that the features proposed with this PR have a value per se and are not correlated to contour lines or hill-shading, even if hypothetically having some future option to apply such layers could be of general benefit to OSM for terrain representation. In my point of view peaks, which are currently rendered in this style, have their own role in documenting territory, although their descriptive importance could be amplified by relief visualization.

As reviewers have already notified, the rendering of names is a relevant part of this PR.

None of these elements is in the field of scientific/technical cartography and all have unquestionable social and cultural merit for common people of all ages.

While openstreetmap-carto style excels on populated places, in which some reassessment on the need to render features in overcrowded zones (especially for business related symbols) might be worthwhile to be evaluated, there is in my point of view some gap with unpopulated areas, where the features in this PR are usually circumscribed. In addition, representing these features could be a positive accelerator for OSM data, which also needs improvement on mountain terrain tagging, as for these zones the database itself is not at the same quality level as in the cities.

imagico:
@Ircama - these arguments are all quite understandable but the discussion here is not about if something like a valley can or should be rendered in a map in general, it is about if what is mapped as natural=valley in OSM with a linear way should be rendered in this map style with a label placed along the way.

The previous discussion we had above did not really have any substantial results.  You have not participated in the [discussions on the wiki](http://wiki.openstreetmap.org/wiki/Talk:Tag:natural%3Dvalley) regarding how to map valleys in a verifiable way.  Overall it seems to me you do not really consider it a problem if mappers place labels rather than mapping something observable and verifiable in nature.  This is an understandable viewpoint from a map design perspective but it clashes with the basic principles of OSM.

As i said the concept of rendering ridges and aretes is on a good way i think - still requires some tuning and testing probably but i definitely consider this suited for this style.  The rest however IMO requires more thorough consideration on the level of mapping and tagging first.  I explained the approach in https://github.com/gravitystorm/openstreetmap-carto/pull/2138#issuecomment-223244893 - the discussion i had with @RicoZ-OSM on the wiki would probably be a good start.

dieterdreist:
sent from a phone

> Il giorno 08 nov 2016, alle ore 11:11, Christoph Hormann <notifications@github.com> ha scritto:
> 
> Overall it seems to me you do not really consider it a problem if mappers place labels rather than mapping something observable and verifiable in nature. This is an understandable viewpoint from a map design perspective but it clashes with the basic principles of OSM.


interesting that this argument is always coming up when speaking about natural/geographic features, but apparently nobody has a problem with place=locality, a tag that is already rendered, that has very imprecise and generic meaning (hence is used more than a million times, also because no more specific alternatives have been established), that is often describing something with significant extent but is almost exclusively used on nodes. This is actual label dropping, not specifying a valley and its name and giving it a rough location.

imagico:
> but apparently nobody has a problem with place=locality

You are aware that place=locality was mentioned here in https://github.com/gravitystorm/openstreetmap-carto/pull/2138#issuecomment-222938620

dieterdreist:
thank you for reminding, so we agree that the situation with place=locality is worse (what doesn't necessarily mean we should add the features from this PR, naturally).

imagico:
No, not worse, just very different.  In any case nothing really related to the subject of this issue.  Stopping rendering labels for place=locality could still be worth considering (but in a separate issue).

Ircama:
> You have not participated in the discussions on the wiki regarding how to map valleys in a verifiable way.

Oh no, I was not aware, thanks for notifying. I just had a look at the [Talk tab](http://wiki.openstreetmap.org/wiki/Talk:Tag:natural%3Dvalley) at the time of developing the rendering of valley, when your topic was not yet disserted. The current PR follows the "approved" polymorphism of this tag in the [Wiki](http://wiki.openstreetmap.org/wiki/Tag:natural%3Dvalley), the one which is currently known to mappers, with the extension to also cover polygons that look rarely used at the moment. I understand the concept that data modelling suggests a valley should be an area, meaning that to verify a farm is in a valley we intersect their areas other than relating them with names. Anyway I will better check that discussion. I also appreciate, if I am well interpreting feedbacks, that there is no a priori apprehension for rendering common geographical features in this style, just a concern on data model.

imagico:
Note there is nothing 'approved' about the tag page on the wiki.  The status in the description box refers to the fact that there is an approved proposal that mentions this tag (in this case it piggy-backed on the ridge proposal without the concept of natural=valley being discussed in the proposal process).  As has been pointed out repeatedly the proposal process is essentially meaningless for rendering decisions here, it is just a means (but no guarantee) to improve tagging consistency.

Ircama:
I would separate the code for ridges/aretes from the other features here proposed and, if you do not see a better option, I can create a new PR just to render ridges/aretes and consider to use this space to go on better focalizing the problem of rendering the other features. We can close this PR meanwhile if you prefer.

Basing on the hypothesis to redefine a valley as area (to be discussed in the wiki) I see two implications and a rendering questions I'd like to develop here.
While rendering the name of an area is fine IMHO in case of very small convex polygon, valleys can be complex to tag and the same logic that allows relations of adjacent surfaces (adopted for other OSM elements), could be considered also in this case (at least as an assumption for the question). Besides, related polygon is often concave; I wonder whether in such cases the renderers (like Mapnik) can easily process the name for these data models (union of concave polygons), and, if not, whether it would be worthwhile to consider including a line in the relation (a way) to allow rendering the name. After all, OSM adopts the same data model e.g., for rivers, where the adiacent polygons are complemented by a linear way positioned along its [thalweg](https://en.wikipedia.org/wiki/Thalweg). Apart from considering that for natural implications many shapes of valleys can be analogous to the ones of rivers and that a thalweg might be useful information for administrative purposes (all these points to be better discussed in the wiki rather than here), there is a data redundancy that in reality has a task to support renderers in better processing the name of the feature.
Regardless the side effect to reuse the simple logic in this PR for valley, do you see an alternative way for the renderers (e.g., Mapnik) to process complex concave polygons (also relations of polygons) for drawing a valley name (possibly in a similar way the valleys are drawn in classic maps, which is hopefully the aim)? E.g., alternatively a thalweg shapefile or something similar? Thanks

imagico:
> I can create a new PR just to render ridges/aretes and consider to use this space to go on better focalizing the problem of rendering the other features.

You can do either that or repurpose this PR by removing the other changes and rebasing your branch.  Given the length of this discussion it is probably cleaner to simply start a new one.

> Basing on the hypothesis to redefine a valley as area (to be discussed in the wiki)

I don't think this was ever seriously discussed in the wiki - the discussion RicoZ started was about mapping them as a relation type that is not a multipolygon, that does not require defining a full outline of the feature in question.

Developing such a type of relation should be based on mapping considerations, not on rendering since this would ultimately end up in suggesting mappers to map in a way that primarily works around the shortcomings of rendering engines.

> After all, OSM adopts the same data model e.g., for rivers, where the adiacent polygons are complemented by a linear way positioned along its thalweg.

This seems to be based on a common misunderstanding of the system of river mapping in OSM.  This does not have the purpose of catering rendering needs, it represents two different and separately mappable aspects of the river system, namely the water flow structure and the water covered area.  If a similar distinction exists for valleys that could be discussed (not here of course) but i don't see this at the moment.

> Regardless the side effect to reuse the simple logic in this PR for valley, do you see an alternative way for the renderers (e.g., Mapnik) to process complex concave polygons (also relations of polygons) for drawing a valley name (possibly in a similar way the valleys are drawn in classic maps, which is hopefully the aim)?

Yes, more sophisticated label placement for complex polygons is entirely feasible, even for real time rendering purposes.  If it is likely that Mapnik will natively support something in that direction in the future is a completely different question of course.

Ircama:
> I don't think this was ever seriously discussed in the wiki - the discussion RicoZ started was about mapping them as a relation type that is not a multipolygon, that does not require defining a full outline of the feature in question.

Yes, I know.

I close this PR as the data definition of the features different from ridge and arete needs some revision.

I see that the data model of natural elements in mountain areas can still be questionable and incomplete. Maybe this is another reason for the very limited usage of such features.

Thanks for the comments and revisions so far.

PR #2142, opened 2016-05-31
fix possible admin label and line rendering mismatch

nebulon42:
Using `planet_osm_polygon` for admin labels while using `planet_osm_roads` for admin lines might lead to a mismatch between lines and labels where labels are rendered without lines. This PR changes the table to `planet_osm_roads` to match both parts.

before
label_before

after
label_after

### DIFF
--- a/project.mml
+++ b/project.mml
@@ -1945,7 +1945,7 @@
       "srs": "+proj=merc +a=6378137 +b=6378137 +lat_ts=0.0 +lon_0=0.0 +x_0=0.0 +y_0=0.0 +k=1.0 +units=m +nadgrids=@null +wktext +no_defs +over",
       "Datasource": {
         "extent": "-20037508,-20037508,20037508,20037508",
-        "table": "(SELECT\n    way,\n    name,\n    admin_level\n  FROM planet_osm_polygon\n  WHERE boundary = 'administrative'\n    AND admin_level IN ('0', '1', '2', '3', '4', '5', '6', '7', '8', '9', '10')\n    AND name IS NOT NULL\n  ORDER BY admin_level::integer ASC, way_area DESC\n) AS admin_text",
+        "table": "(SELECT\n    way,\n    name,\n    admin_level\n  FROM planet_osm_roads\n  WHERE boundary = 'administrative'\n    AND admin_level IN ('0', '1', '2', '3', '4', '5', '6', '7', '8', '9', '10')\n    AND name IS NOT NULL\n    AND osm_id < 0\n  ORDER BY admin_level::integer ASC, way_area DESC\n) AS admin_text",
         "geometry_field": "way",
         "type": "postgis",
         "key_field": "",
--- a/project.yaml
+++ b/project.yaml
@@ -2398,10 +2398,11 @@ Layer:
             way,
             name,
             admin_level
-          FROM planet_osm_polygon
+          FROM planet_osm_roads
           WHERE boundary = 'administrative'
             AND admin_level IN ('0', '1', '2', '3', '4', '5', '6', '7', '8', '9', '10')
             AND name IS NOT NULL
+            AND osm_id < 0
           ORDER BY admin_level::integer ASC, way_area DESC
         ) AS admin_text
     properties:


math1985:
How come that border is not in planet_osm_roads?

nebulon42:
I'm not completely sure, but the MP in my preview has the administrative tagging on the lines, while it hasn't any on the relation itself (https://www.openstreetmap.org/relation/2172251). Maybe the approach could be the other way round, use `planet_osm_polygon` for lines too, but this might have implications on lower zooms?

matkoniecz:
I am not sure whatever line in case like this should not be rendered. 

For example wiki has http://wiki.openstreetmap.org/wiki/Tag:boundary=administrative 

> **The boundary=administrative tag is used on ways.** It may also (or instead) be used on a relation grouping several ways. This works rather like coastline. **Because ways have a maximum number of nodes, you are not required to loop around a whole administration with a single way to form a closed area (although this is a good idea for very small administrative areas)**

Is it misleading/outdated?

matkoniecz:
Or is rendering example difference caused by something else than "MP in my preview has the administrative tagging on the lines, while it hasn't any on the relation itself"?

nebulon42:
In my changes to the admin rendering I introduced `WHERE osm_id < 0` to the queries. The rationale was to not render polygon ways and other ways twice and thus to get a cleaner admin rendering. I'm no expert on how osm2pgsql combines ways into closed ways or what really ends up in the table with an id < 0, but I think this includes multipolygons and other ways that form a closed ring. The downside was that some borders disappeared, but obviously some labels were still rendered.

Here is an example for the difference between v2.37.1 and v2.38.0 (also Salzburg, no labels) with `admin_level=10`:

www gifcreator me_tmb13n

This PR is meant to fix the label/line mismatch I discovered. If the change in rendering is still seen as worthwhile (to me it was) I leave to the maintainers to decide. My view is that borders should be rendered using relations only.

mmd-osm:
Hi,

this patch causes regressions as reported on the [German forum](http://forum.openstreetmap.org/viewtopic.php?id=55149), i.e. labels are placed incorrectly or at seemingly random positions not matching the actual boundary MPs. I could reproduce this locally on my database using tilemill and git bisecting from v2.40.0 to v2.41.0:

```
:~/openstreetmap-carto$ git bisect good 2e3802c
:~/openstreetmap-carto$ git bisect bad 60ed128
Bisecting: 20 revisions left to test after this (roughly 4 steps)
[ab296eacf24955d15edfdfae1326943b603128b5] Merge pull request #2210 from matkoniecz/preview.png
:~/openstreetmap-carto$ git bisect good
Bisecting: 10 revisions left to test after this (roughly 3 steps)
[268189c92292dfc282c48281d1cdf1ee10104cea] Merge branch 'admin-label' of https://github.com/nebulon42/openstreetmap-carto
:~/openstreetmap-carto$ git bisect bad
Bisecting: 4 revisions left to test after this (roughly 2 steps)
[3cb7b4d29a095df5f89dd440a83712e6b438dae0] Merge branch 'obelisk' of https://github.com/mkoniecz/openstreetmap-carto
:~/openstreetmap-carto$ git bisect good
Bisecting: 2 revisions left to test after this (roughly 1 step)
[f4e386a399931aafda0d05cc4f186b9ab521e37d] changing color name from @educational_areas_and_hospital to @societal_amenities
:~/openstreetmap-carto$ git bisect good
Bisecting: 1 revision left to test after this (roughly 1 step)
[863cb4d670e47656f94cf4c051304fdc5f70d426] fix possible admin label and line rendering mismatch by changing admin label table
:~/openstreetmap-carto$ git bisect bad
863cb4d670e47656f94cf4c051304fdc5f70d426 is the first bad commit
commit 863cb4d670e47656f94cf4c051304fdc5f70d426
Author: nebulon42 <nebulon42@yandex.com>
Date:   Tue May 31 22:31:21 2016 +0200

    fix possible admin label and line rendering mismatch by changing admin label table

:100644 100644 f4c1276b4bbfdbcb7aab82ba5de2394a7f65ae14 b808efbd4657a8fc93980e0b324b306aca52782a M	project.mml
:100644 100644 e4986b520cfc3bbe7c94f8bac3f5612bc82380c1 34249cf9ce9247b1909fd9447ddb5673cc3704f2 M	project.yaml
```

nebulon42:
Thanks for reporting this, I'll look into it.

nebulon42:
I think the problem is that in `planet_osm_polygon` the text placement depends on the complete outline and in `planet_osm_roads` on the individual lines. While the winding order of polygons is defined and so two adjacent lines end up with different orientation in case of lines the same line might be shared and so the orientation is the same. That results in wrong text placement.

PR #2146, opened 2016-06-06
Add guideline on symbolizers which change per zoom

gravitystorm:
I realised that while these are very common throughout our code, it's not explicit in the guidelines how to handle them.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -89,6 +89,35 @@ instead of
 * Order the selectors in a style-sheet in rough order of importance (i.e.,
   `highway=primary`, then `highway=secondary`) and beyond that, add layers that
   are rendered later (i.e., higher) lower in the file.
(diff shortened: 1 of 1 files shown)


matkoniecz:
Note that road styling has many violations of this rule, for example https://github.com/gravitystorm/openstreetmap-carto/blob/master/roads.mss#L659

pnorman:
> Note that road styling has many violations of this rule, for example https://github.com/gravitystorm/openstreetmap-carto/blob/master/roads.mss#L659

Yes, but the potential problem is the outer selectors, not the only zoom ones, i.e. it should be something like

```mss
.tunnels-casing[zoom >= 14][access != 'no'],
.tunnels-casing[zoom >= 15] {
  line-width: ...;
  [zoom >= 15] { line-width: ...; }
}
```

But the entire roads MSS is very complicated and difficult to edit.

PR #2147, opened 2016-06-06
Replacing library PNG icon with SVG version

matkoniecz:
I know that it is not perfect and may be improved but IMHO it is anyway better that what we have now - and it is svg.

I tried to tweak it - but it merely confirmed that I am not a good icon maker, so I submit it unmodified except technical tweaks.

PR for work by @kocio-pl and @daganzdaanda in #2050 
amenity library name eeeeee eeeeee node master - library11 200px

vienna - amenity library 48 1962536 16 4006914 node 16 18 master - library11 300px


math1985:
+1

kocio-pl:
I have a small usability idea to make new icon differ a bit between bookshop and a library: in the latter case we could add a folded corner or a bookmark - whatever works.

kocio-pl:
Here is the outcome:
1) library-10 vs bookshop

Issue #2090, opened 2016-03-14
update list of shops rendered as dot


matkoniecz:
Without any new filtering at this moment it would result in following changes (based on https://gist.github.com/matkoniecz/1a33e35fcf32ac4cccc3 generated based on https://github.com/gravitystorm/openstreetmap-carto/blob/master/scripts/shop_values.rb and the current query):

Removed: "phone", "real_estate"

Added: "agrarian", "anime", "auto_parts", "bathroom", "cafe", "candy", "catalogue", "doors", "fishmonger", "glass", "haberdashery", "lamps", "marketplace", "nutrition_supplements", "party", "perfume", "plumber", "provisions", "storage_rental", "storage_units", "street_vendor", "supplements", "swimming_pool", "tableware", "watch", "water", "weapons", "wedding", "window_blind", "windows", "wood"

auto_parts seems to be a clear duplicate of car_parts, there are also other dubious entries.

But raising threshold to 150 gives much better results:

Removed: "builder", "building_materials", "camera", "car_service", "medical", "phone", "real_estate", "salon", "tanning"

Added: "anime", "fishmonger", "haberdashery", "lamps", "marketplace", "party", "storage_rental", "street_vendor", "water", "weapons", "window_blind", "windows"

Added entries make sense (and at least some are documented), removed are undocumented, some are in addition duplicates of documented more popular values.

I would only propose to not include shop=storage_rental as it is not a shop (see failed proposal at http://wiki.openstreetmap.org/wiki/Proposed_features/shop%3Dstorage_rental).

matkoniecz:
Also shop=street_vendor may be worth excluding. What is it supposed to be? Shop selling street vendors? Or a separate tagging scheme not categorizing by what is sold but how?

I am not sure what makes less sense.

HolgerJeromin:
We have amenity = marketplace approved

matkoniecz:
Case of fishmonger is weird - it is appearing in https://github.com/gravitystorm/openstreetmap-carto/blob/6a7a9cc938f2529bcd6f55b277e17efb72f50c14/amenity-points.mss#L676 but it is not appearing in https://github.com/gravitystorm/openstreetmap-carto/blob/master/project.yaml so it is probably not rendered.

It is also not excluded by https://github.com/gravitystorm/openstreetmap-carto/blob/master/scripts/shop_values.rb

Curious.

HolgerJeromin:
It was planed to render. We had discussed this value.

[...]

PR #1617, opened 2015-06-23
Adding shop=seafood and shop=fishmonger icon

kocio-pl:
Resolves https://github.com/gravitystorm/openstreetmap-carto/issues/1417 .

seafood-z19

PR #2149, opened 2016-06-07
render shop=fishmonger

matkoniecz:
fixes inconsistency between amenity-points.mms and .yaml
amenity-points.mms has definition for rendering this shop, but it was never fetched from database due to missing entry in .yaml file

partial fix for #2090 

shop fishmonger name eeeeee eeeeee node master - fishmonger 200px

rosenheim - shop fishmonger 47 8498548 12 0644349 node 17 18 master - fishmonger 300px

tokyo - shop fishmonger 35 6265859 139 391443 way 17 18 master - fishmonger 300px

I have no opinion whatever this shop should be rendered or not, but as mistake in yaml is more likely than in mms I added support for rendering in yaml file rather than remove reference in mms. Intentional rendering of fishmonger is supported by https://github.com/gravitystorm/openstreetmap-carto/pull/1617 For previous discussion of fishmonger see https://github.com/gravitystorm/openstreetmap-carto/pull/604


kocio-pl:
Looking at the last comment about this type of shop https://github.com/gravitystorm/openstreetmap-carto/pull/604#issuecomment-46073340 it looks like the numbers are slowly shifting towards shop=seafood and currently fishmonger no longer has 2086 uses (it's 1 810 -> 200 less in 2 years, while seafood has 4 719 uses - 2,5x more) and Wiki was clearly updated since then to discourage tagging this way.

I don't have a strong opinion on this subject, but I would rather drop it now than fix my mistake with adding this icon.

matkoniecz:
As I have no preference for either solution (due to conflicting reasons supporing both rendering and not rendering) I opened #2151 that also resolves inconsistency.

matkoniecz:
Closing as merging #2151 means that it would not work. But anybody convinced that shop=fishmonger should be rendered may open an issue.

PR #2161, opened 2016-06-10
shop=books reusing library icon with the shop color

matkoniecz:
shop books name eeeeee eeeeee node master - also_shop 200px
vienna - shop books 48 2088427 16 3753247 16 19 master - also_shop 375px

In my opinion library icon is much more obvious. I am not entirely sure about results of having the same icon in different colors for two different things - may be more intuititive, may be more confusing, may be problematic for colorblind people.

But I think that the new book icon is soo good that it outweights these problems.


math1985:
> In my opinion library icon is much more obvious.

+1

kocio-pl:
I can't decide. You're right, new icon is more clear, but current book shop icon is good enough for me (and probably better for color blind people) and I don't feel the need to change it.

polarbearing:
thanks @matkoniecz, the new shop/library icon is quite self-explanatory. The old shop icon looked like a pile of any material, could have been bricks or bales of cloth.

Issue #971, opened 2014-09-22
Rendering hedge as area is not always correct


SK53:
A closed way tagged as barrier=hedge is interpreted as an area for zoom 16 and greater (landcover.mss line 425 & seq). Fields mapped as an area of farmland with a surrounding hedge are now shown exactly the same as woodland.

Example area is Cherry Hinton (z16 and z15):

![cherry_hinton_hedges](https://cloud.githubusercontent.com/assets/5452818/4360004/911a7148-4277-11e4-9b32-5052a12b89c1.jpg)

In this cases the hedges surrounding the fields are uninterrupted.

jeisenbe:
This can be corrected now that area=yes is available as a tag in hstore, so I recommend reopening this issue.

I downloaded most of the features mapped as `way["barrier"="hedge"]["area"!="yes"]["tourism"]`. I found 31 closed ways tagged with hedge and tourism without area=yes. These should not be rendered with the hedge fill color.

Current screenshot:
![campsite-barrier-hedge](https://user-images.githubusercontent.com/42757252/55922479-419b4a00-5c3c-11e9-8879-2dcd62a2ced3.png)
https://www.openstreetmap.org/way/225683577/#map=17/51.1651/-0.04265
way 225683577
barrier=hedge
name=Long Acres Caravan & Camping
tourism=caravan_site

After checking for "area=yes". (I've also changed the outline color to match the fill color).
- This can be done in the SQL query to exclude all types of barrier=* without area=yes, or it could just be done in the CSS for hedges only. 
![z16-campground-hedges](https://user-images.githubusercontent.com/42757252/55936442-50502400-5c71-11e9-87e2-0939ff3acd68.png)

Alternately, we could get rid of the polygon fill and just render the 3 pixel wide green line on the outline of the area, similar to the current rendering for barrier=wall when mapped as an area.

matkoniecz:
> I would suggest retagging the field with a double line, one for the field and one for the area. I think that's also conceptually better, because it feels strange that a single object represents both an area and a field.

I agree, in my opinion tagging hedge around area and area with the same closed way is a tagging mistake.

But this appears to not be shared by wider OSM community (for example I proposed validator warning in JOSM for that), so workaround in rendering adding support for this tagging would not be unwelcomed.

pnorman:
We've always had access to the `area` tag, it's nothing that's changed in 4.0. I stand by the object being an area or not - not some combination of both.

[...]

PR #2166, opened 2016-06-14
render labels for named barriers

matkoniecz:
Done to handle case reported in #1824 

I am not sure whatever it should be changed to ignore polygons. On one hand it will ignore closed ways (and some are valid), on the other hand things like tagging landuse with name and barrier on one object is common.

I am not sure whatever this poor tagging should result in a poor rendering.

barrier wall name eeeeee eeeeee landuse cemetery ref 1 ele 8000 operator ee ee ee operator brand ee ee ee brand closed_way master - new_barrier_name 200px

https://cloud.githubusercontent.com/assets/899988/16034062/e5364a66-320f-11e6-80da-b113c257d7d0.png
https://cloud.githubusercontent.com/assets/899988/16034070/eae4594e-320f-11e6-8c6e-8f769dc3cf25.png
https://cloud.githubusercontent.com/assets/899988/16034045/d1569e10-320f-11e6-8270-13a91b3dc952.png


matkoniecz:
> This branch has conflicts that must be resolved

That was not expected, but in this case it is useful - this one should not be merged too quickly. It will make map worse in many places and the question is whatever doing that for this type of poor tagging is desirable.

matkoniecz:
Previous try - see  #1885

matkoniecz:
To see how common area+barrier+name problem is see http://overpass-turbo.eu/s/gME

dieterdreist:
sent from a phone

> Il giorno 14 giu 2016, alle ore 09:11, Mateusz Konieczny <notifications@github.com> ha scritto:
> 
> I am not sure whatever it should be changed to ignore polygons. On one hand it will ignore closed ways (and some are valid), on the other hand things like tagging landuse with name and barrier on one object is common.
> 
> I am not sure whatever this poor tagging should result in a poor rendering.
> 

IMHO we should not encourage or endorse broken mapping like barrier=wall landuse=cemetery on the same object.

dieterdreist:
sent from a phone

> Il giorno 14 giu 2016, alle ore 09:21, Mateusz Konieczny <notifications@github.com> ha scritto:
> 
> To see how common area+barrier+name problem is see http://overpass-turbo.eu/s/gME
> 

if we start to highlight the problem by showing names on these linear barriers as well (or maybe even show the whole area as barrier) we can be confident that people will clean it up quickly

jojo4u:
Apart from violating the rule "one feature - one osm element" what data problems does the style have with this type of tagging?

dieterdreist:
sent from a phone

> Il giorno 14 giu 2016, alle ore 10:42, jojo4u <notifications@github.com> ha scritto:
> 
> Apart from violating the rule "one feature - one osm element" what data problems does the style have with this type of tagging?


the problem is that you don't know:
- which tag belongs to which object
- whether the feature is intended to be linear or a polygon (in some cases, like barriers)

matkoniecz:
> whether the feature is intended to be linear or a polygon (in some cases, like barriers)

And this data problem was reported as style problem, see #971

pnorman:
:-1: to anything which treats the same object as both a line feature and a polygon feature

dieterdreist:
2016-06-14 23:15 GMT+02:00 Paul Norman <notifications@github.com>:

> 👎 to anything which treats the same object as both a line feature and a
> polygon feature




while I agree with this position, it should be actively communicated to the
community IMHO, because a lot of people do indeed map in a way where both,
linear and polygonal features, are mapped to the same OSM object,
especially barriers.

SomeoneElseOSM:
@dieterdreist The place to do that is probably on http://wiki.openstreetmap.org/wiki/One_feature,_one_OSM_element .  That's I refer people to, and I suspect that others do too.  Also, of course raise it with new mappers in changeset discussions.

matkoniecz:
@pnorman 

> :-1: to anything which treats the same object as both a line feature and a polygon feature

Is it rejection of [barrrier=* + landuse=*] tagging or is it rejection of this rendering? Or both?

math1985:
> Is it rejection of [barrrier=* + landuse=*] tagging or is it rejection of this rendering? Or both?

barrier=city_wall + landuse=retail on a closed linestring should be interpreted as a polygon that is both a city wall and a retail area (i.e. a big city wall tagged as polygon with shops inside).

The rendering should match that interpretation.

I'd recommend waiting with this issue until the database reload, because this might change things.

pnorman:
> I'd recommend waiting with this issue until the database reload, because this might change things.

I don't see that it'll change things, since it'll still be a polygon (or not).

The technical implementation of this is that there's a function which, given the tags of an object, tells you if it is a polygon or not.

matkoniecz:
@pnorman What is your opinion about rendering names solely for lines?

pnorman:
4M barriers, 17k named.

Selecting random ones I see a mix of descriptions and actual names. Most of the named ones old.

The problem is going to be coming up with a suitable style, and I have no ideas there.

PR #2168, opened 2016-06-15
Reformat road SQL

pnorman:
Pulled from #2058


project.yaml
```diff
@@ -992,7 +1011,7 @@ Layer:
                 foot,
                 bicycle,
                 tracktype,
-                'null',
+                'null' AS surface, -- Should be a SQL NULL?
```
matkoniecz: I am not understanding this comment. Is it question about why the NULL is here?

project.yaml
```diff
@@ -938,23 +940,40 @@ Layer:
     <<: *extents
     Datasource:
       <<: *osm2pgsql
+      # This is one of the most complex layers, so it bears explaining in some detail
+      # It is necessary to
+      # - Have roads and railways in the same layer to get ordering right
+      # - Return two linestrings for ways which are both a road and railway
+      # - Compute z_order here, rather than rely on osm2pgsql z_order
       table: |-
```
matkoniecz: Maybe mention also aeroways? I admit that I am not really sure what is the reason for querying aeroways with roads and railways.

Is it intended to handle cases like [airport/road crossings](http://www.openstreetmap.org/#map=18/36.15107/-5.34812)?

project.yaml
```diff
@@ -508,11 +508,12 @@ Layer:
...
-                                      'grass_paver', 'gravel', 'ground', 'mud', 'pebblestone', 'salt', 'sand', 'woodchips', 'clay') THEN 'unpaved' ELSE 
-                  CASE WHEN surface IN ('paved', 'asphalt', 'cobblestone', 'cobblestone:flattened', 'sett', 'concrete', 'concrete:lanes', 
-                                        'concrete:plates', 'paving_stones', 'metal', 'wood') THEN 'paved' 
-                  ELSE null END END AS int_surface,
+                CASE WHEN surface IN ('unpaved', 'compacted', 'dirt', 'earth', 'fine_gravel', 'grass', 'grass_paver', 'gravel', 'ground',
+                                      'mud', 'pebblestone', 'salt', 'sand', 'woodchips', 'clay') THEN 'unpaved'
+                  WHEN surface IN ('paved', 'asphalt', 'cobblestone', 'cobblestone:flattened', 'sett', 'concrete', 'concrete:lanes',
+                                      'concrete:plates', 'paving_stones', 'metal', 'wood') THEN 'paved'
+                  ELSE NULL
+                END AS int_surface,
```
matkoniecz: Thanks! I have no idea why I thought that nested CASE is a good idea.

project.yaml
pnorman: yes

project.yaml
pnorman: I'd be fine with pulling aeroway out of the roads layer for complexity reasons, but that's more involved then this cleanup

matkoniecz: extracted to https://github.com/gravitystorm/openstreetmap-carto/issues/2169

project.yaml
matkoniecz: null is here as it makes little sense to query surface for railways.

on the other hand querying some other tags like tracktype, bicycle, foot for railways is also rather pointless so there is no good reason to do the same for surface.

Also, for aeroway querying surface is not completely pointless.

Overall replacing this null by querying surface may be a good idea.

project.yaml
matkoniecz: And I think that it would be a good idea to mention in this comment that aeroways are also handled here.

It may be removed if/once #2169 is done.

project.yaml
dieterdreist: sent from a phone

> Il giorno 15 giu 2016, alle ore 12:45, Mateusz Konieczny <notifications@github.com> ha scritto:
> 
> on the other hand querying some other tags like tracktype, bicycle, foot for railways is also rather pointless so there is no good reason to do the same for surface.


maybe this is (or was) to cater for some railway=abandoned that now are tracks, or footways or cycleways?

pnorman: There has to be a column as part of that query as it's part of a JOIN. The question isn't why there's a colum, but if it should be `'null'::text` or `NULL::text`

project.yaml
pnorman: I don't think anything on aeroways belongs here as there aren't any aeroway requirements influencing how we need to write the query. The other points are about what requires the `UNION ALL` and the z_order `JOIN`

Issue #2172, opened 2016-06-16
Explain or replace usage of Natural Earth boundaries data


matkoniecz:
Currently ne boundaries data is used for z1 to z3. Replacing it is a simple change, but unfortunately with some side-effects.

Next usage of NE data will be removed (what will leave builtup layer as sole place where OSM data is not used).

Some places may no longer have a different data (direct consequence of above)

Rendering of z1 to z3 will be much slower (small initial testing indicated that generation time may double or even tripple)

Due to #2164 the same line-width result in heavier lines

Maritime borders will be rendered (#621), but with ocean-polygons it should be fixable.

before
![ne 51 01155 14 93786 z03 60px 48e7a6f119eebc89d8747fd8aeea1774a5fc9c58 60px](https://cloud.githubusercontent.com/assets/899988/16123110/1836f92a-33ea-11e6-85d4-b7e490570753.png)

after
![d77af25 51 01155 14 93786 z03 60px d77af25afa68c3c645e5b7df7520fd98c2fa3cd9 60px](https://cloud.githubusercontent.com/assets/899988/16123112/190b3e24-33ea-11e6-919f-1e4c660d24d2.png)

In this case I would be tempted to remove ne data, depsite significant preformance regression (what is quite ironic, as I noticed that ne data is used during looking for easy ways to improve performance). As potential alternative - is there an OSM derived equivalent of ne borders?

But maybe proper method of solving this is to document that ne data is used for performance reasons and to avoid #2164 and #621.

matkoniecz:
Updated before/after and conlusions, I initially switched files.

dieterdreist:
sent from a phone

> Il giorno 16 giu 2016, alle ore 18:14, Mateusz Konieczny <notifications@github.com> ha scritto:
> 
> Rendering of z1 to z3 will be much slower (small initial testing indicated that generation time may double or even tripple)


this can be ignored, these tiles are rendered rarely and in both cases one metatile covers the whole mercator world

pnorman:
> Rendering of z1 to z3 will be much slower (small initial testing indicated that generation time may double or even tripple)

z1-z2 isn't a concern, z3 might be an issue but I suspect its slowness is due to placenames-medium

sommerluk:
> I suspect its slowness is due to placenames-medium

[...]

PR #2173, opened 2016-06-17
placenames-medium is not used on z3

matkoniecz:
No rendering change as nothing from this layer is rendered on z3 - see https://github.com/gravitystorm/openstreetmap-carto/blob/master/placenames.mss

Makes rendering on z3 slightly faster.

### DIFF
--- a/project.mml
+++ b/project.mml
@@ -1306,7 +1306,7 @@
       ],
       "properties": {
         "maxzoom": 15,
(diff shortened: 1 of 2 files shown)


matkoniecz:
Thanks to @sommerluk for noticing it in https://github.com/gravitystorm/openstreetmap-carto/issues/2172#issuecomment-226621573 !

PR #2176, opened 2016-06-18
render man_made=obelisk, fixes #2126

### LINKED ISSUE #2126: Render man_made=obelisk
Noticed recently when I changed a node tagged as historic=monument to man_made=obelisk, it disappeared from the standard OSM map: http://www.openstreetmap.org/node/1963176957

Can man_made=obelisk please be rendered with the same icon as historic=monument? In the case of Aberdour Obelisk, it does not commemorate anything so it is correctly tagged, but it would be nice to see a large structure shown on the map. Especially as the historic=monument icon seems to be an obelisk!

matkoniecz:
man_made obelisk name eeeeee eeeeee ref 1 ele 8000 operator ee ee ee operator brand ee ee ee brand node master - obelisk 200px

https://cloud.githubusercontent.com/assets/899988/16169795/f5824a9e-353b-11e6-8081-a343b1c4ffe6.png
https://cloud.githubusercontent.com/assets/899988/16169794/f57f9042-353b-11e6-99bd-09ee04fbd7af.png
https://cloud.githubusercontent.com/assets/899988/16169793/f57f295e-353b-11e6-89ce-a98405a4fed4.png


math1985:
Only 255 instances, is it really worth additional code?

matkoniecz:
I think that in that case - yes.

Obviously, I would prefer to dismantle this confusing historic=monument thing and have a better tagging scheme (there should be one tag to cover "big obelisk/statue/monument/etc").

But given current situation, I think that rendering man_made=obelisk is the simpliest way to deal with it.

dieterdreist:
sent from a phone

> Il giorno 21 giu 2016, alle ore 23:29, math1985 <notifications@github.com> ha scritto:
> 
> Only 255 instances, is it really worth additional code?


the monumental obelisks are in all instances very important landmarks, be it those thousands of years old ancient egyptian ones (outside egypt you can find them for instance in Rome, Paris, London, NYC), or their much more recent copies like the one at the Washington mall. Do you really question the sense of treating them special?

PR #2177, opened 2016-06-18
rewrite amenity-low-priority to improve compiled .xml, stop displaying blatant mistaggings

matkoniecz:
At this moment it is only very slight performance improvement, but it removes potential for this layer to be unexpectedly problematic.

highway=mini_roundabout, railway=level_crossing makes absolutely no sense to be mapped as areas and it is almost never done and it is documented on wiki as a bad idea.


pnorman:
Oh how I hate the need for coalesce...

PR #2214, opened 2016-07-04
Junction names also on areas

### LINKED ISSUE #1253: Render junction=yes on areas

sommerluk:
Fixes #1253 by rendering junction names also on areas.

### DIFF
--- a/project.mml
+++ b/project.mml
@@ -1528,7 +1528,7 @@
       "srs": "+proj=merc +a=6378137 +b=6378137 +lat_ts=0.0 +lon_0=0.0 +x_0=0.0 +y_0=0.0 +k=1.0 +units=m +nadgrids=@null +wktext +no_defs +over",
       "Datasource": {
         "extent": "-20037508,-20037508,20037508,20037508",
-        "table": "(SELECT\n    way,\n    highway,\n    junction,\n    ref,\n    name\n  FROM planet_osm_point\n  WHERE highway = 'motorway_junction' OR highway = 'traffic_signals' OR junction = 'yes'\n) AS junctions",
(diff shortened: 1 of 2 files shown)


math1985:
Please remind me, why does this need to go into the lua branch?

sommerluk:
Honestly, I’m not even sure.

junction=yes is treated as linestring by the current database import. So it will not work without a database reload. So I’ve tested this code in the lua branch.

However, looking at the code, I suppose that it could also be merged into the normal branch without producing errors (but I would have to test this before…). It would simply not change the rendering while the database import continues using the old/current style file. But it might be easier to synch between master branch und lua branch? Do you prefer a PR against master?

math1985:
> . So it will not work without a database reload

I think then the lua branch is the right one place :)

pnorman:
I'm inclined against accepting any PRs for style changes with the Lua branch and would rather focus on technical work with it right now.

math1985:
> I'm inclined against accepting any PRs for style changes with the Lua branch and would rather focus on technical work with it right now.

+1, let's not accept new features in the lua branch until it has been rolled out to the openstreetmap.org servers, also to make debugging any possible problems easier.

sommerluk:
> not accept new features in the lua branch until it has
> been rolled out to the openstreetmap.org servers

What would be the timeframe for this?

math1985:
> What would be the timeframe for this?

There are interdependencies between a large number of volunteers, so unfortunately it's rather difficult to give a timeframe.

pnorman:
Closing since we're not accepting PRs against the Lua branch and a new PR will need to be opened post-merge. The exact same branch might be usable, but it has to be a new PR for technical reasons.

PR #2216, opened 2016-07-07
Add rendering of `leisure=dog_park`

### LINKED ISSUE #341: Add rendering for leisure=dog_park
[leisure=park](http://wiki.openstreetmap.org/wiki/Tag:leisure%3Dpark) is used for  public recreation spaces in Brazil. These parks can contain dog parks. But **[leisure=dog_park](http://wiki.openstreetmap.org/wiki/Tag:leisure%3Ddog_park)** isn't rendering over leisure=park. [See here](http://osm.org/go/PT3umyZfZ).

It would be interesting a [leisure=garden](http://wiki.openstreetmap.org/wiki/Tag:leisure%3Dgarden) similar rendering, adding a soft edge in order to contrast the green background.

jdhoek:
Dog parks are rendered in a slightly darker shade of the fill colour
used for parks.

The 'paw' icon is based on work contributed to the public domain by Trac
user Giardia¹.

This closes #341.

1: https://trac.openstreetmap.org/ticket/2482


math1985:
Thanks, looks good!

imagico:
Huh - that was fast.  I am not sure if the base color is a good idea here - both in terms of similarity with other colors and using another distinct color for a really specialized feature.  Could we get some sample renders?

For me this does not really have much similarity to leisure=park despite the name - i would put it more next to

http://wiki.openstreetmap.org/wiki/Tag:leisure%3Dhorse_riding
http://wiki.openstreetmap.org/wiki/Tag:sport%3Dequestrian
http://wiki.openstreetmap.org/wiki/Tag:sport%3Dhorse_racing
http://wiki.openstreetmap.org/wiki/Tag:sport%3Ddog_racing

the latter three in combination with `leisure=track`, `leisure=pitch` or `leisure=sports_centre`.

Maybe it would make sense to develop a general concept to render animal sports areas in the wider sense.

math1985:
Typical dog park in the Netherlands:
Dog park
Redering it in park-colours looks fine to me.

imagico:
According to the code it is not park color, it is `darken(@park, 5%)`.

To me a facility like the one in the photo has more similarities with a (relatively informal) sports area.  Given the large variety of animal sports and non-agricultural keeping of animals that exists internationally i think a common base design (possibly with varying pattern) for all of these would be a good idea - and probably using the park color or a park-like color would not be ideal for that.

jdhoek:
@imagico Dog parks are usually a small standalone park or a part of a larger park (either fenced off or simply as an area designated by signs) where dog owners can let their dogs run around unleashed. This to me is quite dissimilar to sports areas (for man or animal), hence the darkened `@park`.

Please see #341 for my sample renders.

imagico:
> Dog parks are usually a small standalone park or a part of a larger park

But that is fairly culture specific - it probably varies strongly between different countries, just like the incidence of dog parks compared to similar facilities for other types of animals.

> This to me is quite dissimilar to sports areas

Could you point out where you draw the line here - it might make sense to work this into the tag definition then.  Right now leisure=dog_park specifically includes obstacle tracks for example which is in my eyes a sports related feature.

> Please see #341 for my sample renders

Ok, thanks - there are different designs shown there so i am not too sure which is the one now implemented here - i assume it is more or less the last one shown.

I think the pattern symbols are too strong - generally a pattern should be significantly more subtle than individual point symbols, otherwise overall readability of the map suffers.

Since the base color is a darker and stronger green than leisure=park which is characterized by maintained, gardened plants, in dry areas usually also irrigated while leisure=dog_park does not necessarily imply any vegetation the choice of color - even assuming a general similarity to leisure=park - appears somewhat non-intuitive to me.

Note the chosen color has the highest chroma value of all green area tones in this style currently.

jdhoek:
@imagico Please have a look at the original Trac issue as well for the discussion of this feature. I think most of the concerns have been raised and addressed there and in #341.

The OSM wiki defines `leisure=dog_park` like this:

> A dog park is a park for dogs to exercise and play off-leash in a controlled environment under the supervision of their owners.
>
> In many regions, dogs are to be kept on a leash in public parks. **This tag is to mark designated areas where dogs can run free.** These areas are often adjacent to a park or a specially marked part of a park and are often fenced, so that dogs cannot escape into the larger park. In general, there is water for the dogs and little plastic sacks to clean up after the dog. Sometimes there is an obstacle track for the dogs.

*(emphasis mine)*

From my understanding this rules out actual dog sports such as dog racing (is that a thing?). I agree that areas used for training of dogs sit somewhere between *park* and *sports*, but I got the impression from the original feature request that `leisure=dog_park` is often used as the wiki describes it. This is also what I have observed in my town (I have implemented this feature because it wasn't rendered). Do you think there are many areas tagged with this tag that will cause ambiguity?

Currently the colour values chosen sit between park and sports area in terms of intensity. I chose the current (higher) contrast of the paws in response to a comment from @kocio-pl (the lighter version seemed too close to `natural=scrub`). If this turns out too strong, perhaps it is best to open a new issue with examples? Please feel free to @mention me, I'll gladly help fine tune the design if necessary.

imagico:
> Do you think there are many areas tagged with this tag that will cause ambiguity?

I think there could be confusion on the tag definition since rendering this while not rendering any other dog related facilities in a distinct way will widen use of this tag so having well defined limits to where it applies would be a good idea.

Around here there are many areas for more serious dog training - both commercial 'dog schools' as well as training courses from dog owner clubs etc.

> Currently the colour values chosen sit between park and sports area in terms of intensity.

_Intensity_ is not a very well defined term w.r.t. color.  As said in terms of chroma (or saturation) this is the most 'colorful' green area tone currently in the style.

> If this turns out too strong, perhaps it is best to open a new issue with examples?

I definitely will once this change is released but i was hoping this could be fixed before the next release.

dieterdreist:
sent from a phone

> Il giorno 10 lug 2016, alle ore 12:14, Christoph Hormann <notifications@github.com> ha scritto:
> 
> For me this does not really have much similarity to leisure=park despite the name - i would put it more next to
> 
> http://wiki.openstreetmap.org/wiki/Tag:leisure%3Dhorse_riding
> http://wiki.openstreetmap.org/wiki/Tag:sport%3Dequestrian
> http://wiki.openstreetmap.org/wiki/Tag:sport%3Dhorse_racing
> http://wiki.openstreetmap.org/wiki/Tag:sport%3Ddog_racing
> 
> the latter three in combination with leisure=track, leisure=pitch or leisure=sports_centre.
> 

around here (Rome) these are quite common and they are usually a part of the park, separated by a fence, but otherwise not different from the rest of the park, save the different drinking water fountains. You'd find them in most of the bigger parks but also in many of the smaller ones.

dog type visible here (and also the usual types):
http://www.annazelli.com/roma-i-nasoni-di-roma-le-fontanelle-della-città.htm

dieterdreist:
sent from a phone

> Il giorno 10 lug 2016, alle ore 12:45, Christoph Hormann <notifications@github.com> ha scritto:
> 
> To me a facility like the one in the photo has more similarities with a (relatively informal) sports area.

 around here have no sports whatsoever association, it's dog owners meeting other dog owners and unleashing the dogs, but typically neither the owners nor the dogs move particularly fast ;-)
http://www.meetart.it/media/cache/original/2012-09-09_parco_scott_area_cani_dett_zoom.jpg

nebulon42:
Sorry, I don't think this is an improvement. It reminds me of the old rendering of zoos and I was glad that we got rid of that. The paws are far too prominent and since we have removed a lot of regular patterns lately I don't think it is good to re-introduce a new one.

dieterdreist:
sent from a phone

> Il giorno 10 lug 2016, alle ore 14:41, Christoph Hormann <notifications@github.com> ha scritto:
> 
> Around here there are many areas for more serious dog training - both commercial 'dog schools' as well as training courses from dog owner clubs etc.


are these generally accessible to the public? Or do you need to get an appointment, pay a fee and respect opening hours, become member or client, etc.? Is there staff? There's a lot of differences

imagico:
> are these generally accessible to the public?

Don't know, don't have a dog.  But i am pretty sure if this is the only dog specific tag rendered people will start tagging such things with leisure=dog_park unless the tag definition specifically excludes this and there is a better tag for this kind of facility.

kocio-pl:
@nebulon42: The difference is that zoo is a big area with structure that should be visible (some buildings, footways, animal areas, ponds, toilets etc.) while dog park is uncomplicated amenity and I guess its area is always quite small.

Probably single paw icon would be less controversial, but I'm not opposed to paw area - especially in the example shown in the issue ticket. However if there are some places where the area is bigger I might reconsider my position on this.

dieterdreist:
sent from a phone

> Il giorno 10 lug 2016, alle ore 20:21, Christoph Hormann <notifications@github.com> ha scritto:
> 
> But i am pretty sure if this is the only dog specific tag rendered people will start tagging such things with leisure=dog_park unless the tag definition specifically excludes this and there is a better tag for this kind of facility.


yes, but I doubt it would stop if the definition explicitly excluded other dog facilities as long as there aren't tags for them "available"

meased:
#2253 and #2250 got me thinking about the paw pattern so I experimented with jsdotpattern for a few minutes and thought I'd share the results. Background is the current park color and foreground is one I randomly picked that was a little darker.

Random dot pattern:
random-aligned

Random direction:
randow_semi

Random spacing:
random

math1985:
This might be personal, but I find the original pattern more pleasant to the eye.

kocio-pl:
In my view current pattern is OK, since we want to show something is designed and random patterns mean that the feature is rather natural/not planned and this is not the case (but [it makes perfect sense for scrub](https://github.com/gravitystorm/openstreetmap-carto/issues/2253)!).

What about the colors however? Do you want to change background to park color and paws to a new value or you just didn't care about the colors?

meased:
I do agree that the current colors are a bit bright, both foreground and background. I like the park color for the background, and think the pattern alone would be enough to show a dog_park within a park (especially if combined with barrier=fence, which it probably should be).

I have no strong feelings on the pattern, just putting options on the table. I like the planned/not planned reasoning.

HolgerJeromin:
Closed dog parks should have a fence mapped. If they are open (no fence) in a park it is probably OK to not render the outline as there is no clear border in reality, too.

jdhoek:
> especially if combined with barrier=fence, which it probably should be

Depends on the local situation. In the Netherlands these kind of areas are often delimited by natural boundaries instead of fences, so a slightly different background colour helps to distinguish the area from the surrounding park and other grassy areas.

@meased 
I agree with @kocio-pl about the pattern though. These areas are planned, so the repeating paws seem most suitable.

By the way, if you want to change the colours, open a new pull request if you can do it yourself, or a new issue if you need someone to help you and refer to this pull request. This is a pull request that has already been merged — it is effectively dead.

PR #2218, opened 2016-07-09
Add icon for amenity clock points

jdhoek:
Public clocks can be useful as orientation points. This adds an icon for
this amenity.

### Zoom 17

Omitted from this PR. May be introduced post-v3 by filtering on `visibility`.

### Zoom 18

clock 18

### Zoom 19

clock 19

See also #108.


HolgerJeromin:
IMO zoom 17 is too early for clocks. Or the icon is too big on this zoom level.

jdhoek:
Zoom 17 seems appropriate to me because these clocks tend to stand out in real life, but are also spaced out across the area, and can thus aid in orientation. The icon is the same size as the other amenity icons (compare with the ATM and playground icons in these screenshots, both available at zoom 17).

nebulon42:
Maybe give the clock a stronger outline? https://github.com/gmgeo/osmic/blob/master/amenity/clock-14.svg

symbols/clock.16.svg
```diff
@@ -0,0 +1,77 @@
...
+     inkscape:snap-midpoints="true"
+     inkscape:snap-bbox="true"
+     inkscape:bbox-paths="true"
+     inkscape:bbox-nodes="true"
+     inkscape:snap-bbox-edge-midpoints="true"
+     inkscape:snap-bbox-midpoints="true">
+    <inkscape:grid
+       type="xygrid"
+       id="grid4167" />
+  </sodipodi:namedview>
```
pnorman: Please create a cleaner SVG. I believe you can do that in Inkscape with export

jdhoek:
@pnorman Amended.

@nebulon42, @HolgerJeromin Clock is now slightly smaller and has a slightly thicker outline.

jdhoek:
Screenshots updated.

HolgerJeromin:
Thanks. I like rendering clocks. 
Later we could blacklist some zoom levels depending on Visibility=* if this is needed.

jdhoek:
Rebased branch against master.

jdhoek:
@HolgerJeromin Yeah that sounds sensible.

math1985:
I'm a bit worried about rendering this as it's not clear what the scope is of this tag.

I agree rendering clocks like [this one](https://upload.wikimedia.org/wikipedia/commons/4/48/Chamberlain_Clock_Jewellery_Quarter.jpg) would be nice.

However, what kind of other objects might be tagged like this? Small public clocks on a lamppost (there are a lot of these in Luxembourg) or small clocks like [this](http://wiki.openstreetmap.org/wiki/Tag:amenity%3Dclock)? Every single clock in a train station? Clocks on the front of (historical) buildings?

math1985:
Example of an area in which many clocks are mapped: http://overpass-turbo.eu/s/heg

SomeoneElseOSM:
I'm not convinced that "every small clock" makes sense, but larger clock _buildings_ probably do.  As an example:

http://www.openstreetmap.org/way/116816364

imagico:
For areas where every publicly visible clock is mapped showing from z17 is probably too early - especially in railway stations for example.  But since there are tags to differentiate different types of clocks (support and visibility tags as well as combination with building=yes) this should be manageable in the future.  For now i would think starting at z18 with the icon shown would be fine.

jdhoek:
@imagico Yeah good point. I have amended the PR to reflect this.

I have attempted to place all clocks at `z >= 18`, with the exception of clocks with `visibility=area`, these start at `z = 17`.

This all seems to work locally (kosmtik), is this approach correct?

jdhoek:
@math1985 Using the visibility tag we can further reduce the number of clocks shown. Perhaps visibility=area` at `z = 17`, `visibility=street` at `z = 18`, and `z => 19` for the rest?

imagico:
Using `visibility` would make this a version 3 PR - that is why i wrote _should be manageable in the future_.  At the moment you cannot use this yet.

jdhoek:
@imagico Version 3 of Mapnik? I can split this PR up, that would leave the visibility matching for a future version. Is there a meta-issue that can be referenced?

imagico:
Version 3 of this style.  See 

https://github.com/gravitystorm/openstreetmap-carto/milestone/1

jdhoek:
I take it this is specifically about adding hstore supprt? (#1504)

math1985:
> @math1985 Using the visibility tag we can further reduce the number of clocks shown. Perhaps visibility=area at z = 17, visibility=street at z = 18, and z => 19 for the rest?

No, we should not create a precedent for abandoning the distinction between map data and how to render it. A visibility tag is not map data, and also not something that can be objectively validated in the real world.

jdhoek:
It would neatly solve exactly the situation you linked to at Düsseldorf Hbf. though. Clocks that are visible from afar are points of orientation, station clocks aren't (disregarding if they should even be tagged at the platform info sign level in the first place).

dieterdreist:
sent from a phone

> Il giorno 10 lug 2016, alle ore 16:17, Matthijs Melissen <notifications@github.com> ha scritto:
> 
> No, we should not create a precedent for abandoning the distinction between map data and how to render it. A visibility tag is not map data, and also not something that can be objectively validated in the real world


it is map data and it can be verified

math1985:
> it is map data and it can be verified

Sorry, in the e-mail notification the layout got messed up, and I misunderstood. I thought the proposal was to create tags like visibility=z17 or visibility=z18, with the purpose of explicitly indicating from which zoomlevel the clock should be rendered.

Creating visibility tags that indicate from which area a clock is visible is of course fine.

HolgerJeromin:
This is not clear in this issue: 
visibility (and support) is a well documented tag for clocks. 
The use in the future depends on hstore not a tagging discussion.

jdhoek:
Just to clarify: this PR is aimed at v2 of osmCarto. I will submit a separate PR for the filtering based on visibility for v3. Are there any objections remaining for this PR?

math1985:
Unfortunately, I don't think we currently have a way to distinguish important clocks from unimportant ones. I therefore will have to reject this pull request. Thank you for the effort though @jdhoek!

HolgerJeromin:
I don't think there are too much clocks mapped right now.

jdhoek:
@math1985 This PR sets all clocks at `zoom >= 18` precisely because there is no way to distinguish between them currently. I was under the impression that that zoom-level was high enough for this level of detail.

kocio-pl:
@HolgerJeromin You would be surprised! =} :

http://taginfo.openstreetmap.org/tags/amenity=clock

@math1985 What if we just made it z19? It's better than not showing at all and we don't risk anything this way IMO. And I second @jdhoek - z18 is still pretty safe for me. Please reconsider your decision.

math1985:
Reopened.

gravitystorm:
I think the discussion should focus on the purpose of showing these on the map. I think there is little point in showing the location of a clock on a map in general, since it is unlikely that anyone will think "what is the time" and then consult a general-purpose map style to find their nearest clock. Similarly, in reverse, there is little chance that someone will consult a map to find out what the clock is. Showing all clocks in OSM is a valid thing for specialist maps, but not this one.

I'm never convinced that z(N) or z(N+1) is a good place for showing things that are of marginal purpose. When that is the main discussion it feels like an acknoledgement that they aren't really worth showing but we'll hide them in the corner somewhere where they don't cause problems. As we found many times (with colours, icon redesigns etc), *everything* has some level of maintenance burden, so *everything* has to be worthwhile for this style.

So the purpose of showing these clocks is when they serve a larger purpose than just displaying the time - as @jdhoek says in the original message, "Public clocks can be useful as orientation points". So we should find a way to show the clocks that are useful orientation points, without showing the ones that are just good for showing the time.

If this needs to wait for hstore, then lets put this on hold and concentrate on sorting out hstore. If there is a way to show only the clocks useful for orientation, then lets discuss that here.

kocio-pl:
Sundial clocks ([>500 uses](http://taginfo.openstreetmap.org/tags/display=sundial)) are interesting for me to be shown, because in the parks and squares they are one of a few types of attraction/orientation points (along with ponds, fountains and artworks) and they rarely exist outside these places, so don't make a problem for crowded places. But this of course rely on having hstore available.

pnorman:
Now that we've had a bit more discussion I'm going to close this again. Distinguishing clocks that are useful as orientation points seems to require a DB reload for the additional tags.

PR #2219, opened 2016-07-10
Add tertiary to colour script; split up roads.mss

jdhoek:
Tertiary roads are now the fifth and final road that has its colours
generated by generate_road_colours.py. The script is amended to
facilitate this.

The generated colour values now reside in road-colors-generated.mss, and
have been removed from roads.mms. road-colors-generated.mms is generated
by the colour script, and should not be edited manually. To regenerate
it:

```
./scripts/generate_road_colours.py > road-colors-generated.mss
```

The original output of the colour script included configuration values
not used by us directly. To get this output as well as the generated
colours, call the script with the `-v` flag (for verbose).

This attempts to solve #2085.

### DIFF
--- a/project.mml
+++ b/project.mml
@@ -2092,6 +2092,7 @@
     "landcover.mss",
     "water.mss",
     "water-features.mss",
+    "road-colors-generated.mss",
     "roads.mss",
     "power.mss",
     "citywalls.mss",
--- a/project.yaml
+++ b/project.yaml
@@ -41,6 +41,7 @@ Stylesheet:
   - "landcover.mss"
   - "water.mss"
   - "water-features.mss"
+  - "road-colors-generated.mss"
   - "roads.mss"
   - "power.mss"
   - "citywalls.mss"
--- a/road-colors-generated.mss
+++ b/road-colors-generated.mss
@@ -1,0 +1,32 @@
+/* This is generated code, do not change this file manually.         */
+
+/* To change these definitions, alter                                */
+/* scripts/generate_road_colours.py and run:                         */
+/*                                                                   */
+/*   ./scripts/generate_road_colours.py > road-colors-generated.mss  */
+/*                                                                   */
+@motorway-fill: #e892a2;
+@trunk-fill: #f1aca0;
+@primary-fill: #f5c8aa;
+@secondary-fill: #f6e3bf;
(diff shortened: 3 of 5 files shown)


jdhoek:
Sample showing typical use of tertiary roads in The Netherlands — roads that provide access inside a neighbourhood, typically used by local buses to service the area as well.

with-bus

jdhoek:
I have left low-zoom as-is, because tertiary roads aren't drawn at that level anyway. The shields are as they were as well, and look okay to me, but please test these; they could be made a tad darker to match the new fill and stroke colours.

nebulon42:
This is some good refactoring but you shouldn't mix refactoring with style changes. E.g. you also change the width of tertiary. This will lead to discussion and may need more previews, so it may be better to split it off into a separate PR that builds upon that one.

I have not looked in detail at your changes to the colour generation script, but maybe you could also separate the config of the colour values and the code. This is something that @matkoniecz wanted to do for some time.

Apart from that I still don't think that tertiary needs to be yellow again. I know that others would quite strongly disagree here.

jdhoek:
@nebulon42 see #2220.

kocio-pl:
I still don't really know if we should or shouldn't switch to yellow tertiaries, but I can accept it, given how many people miss them. However I've just noted that changing it through the script (not just manual tweaking one color) results in changing the rest of the colors, especially secondary roads - compare Grovestins road on the picture above with current rendering:

http://www.openstreetmap.org/#map=16/53.2060/5.8350

Changing the color of primary road (Anne Vondelingweg) is noticable, but less, of course (the higher road type, the change is smaller).

So this PR (and other script-based ones) has different effect than https://github.com/gravitystorm/openstreetmap-carto/pull/2085 and we should decide which one should we use. I prefer only hand tweaking tertiaries, changing other road colors are too intrusive for me.

jdhoek:
@kocio-pl This PR is closed, please see #2220 for the refactored code (which does not change any colours) and #2085 for the discussion.

kocio-pl:
I guess general discussion belongs rather to (still open) issue https://github.com/gravitystorm/openstreetmap-carto/issues/1974.

jdhoek:
Right you are, I didn't notice #2085 was closed as well.

PR #2220, opened 2016-07-10
Split up roads.mss, refactor road colour script

jdhoek:
The generated colour values now reside in road-colors-generated.mss, and
have been removed from roads.mms. road-colors-generated.mms is generated
by the colour script, and should not be edited manually. To regenerate
it, edit road-colors.yaml and run:

```
./scripts/generate_road_colours.py > road-colors-generated.mss
```

The original output of the colour script provided some info about the
amount of loss caused by converting from CIE lch to RGB. To get this
output, call the script with the `-v` flag (for verbose).


jdhoek:
This replaces the refactoring part of #2220.

Related to #2085.

pnorman:
:+1: in principle, but I haven't done a code review yet

pnorman:
I opened https://github.com/jdhoek/openstreetmap-carto/pull/1 with some changes and support for road shields

Can tertiary not be special-cased in `generate_shields.py` and generated from `generate_road_colours.py`? If there's no way to easily do so it's okay as it is, but then we should add a comment to generate_road_colours.py about where the rgb hex values for it come from

jdhoek:
@pnorman Great, I'll have a look.

HolgerJeromin:
We could check the status of road-colors.yaml (that it generates the same file as the checked in version)  with Travis. But this could be done later.

jdhoek:
With the help of @pnorman `generate_shields.py` now uses the settings from `road-colors.yaml` as well. For both generating files: their output is unchanged.

pnorman:
> `generate_shields.py` now uses the settings from `road-colors.yaml` as well.

Thanks. I thought about doing this but didn't want to get into that script as well.

jdhoek:
Turned out to be a minor change in that script, so it seemed beneficial. I did alter the spacing to align with Python standards and `generate_road_colours.py`, [but only a handful of lines were actually altered](https://github.com/gravitystorm/openstreetmap-carto/pull/2220/commits/d13cb8892f7a2565a5026f986b7e9e9fe5880ab4?w=1).

math1985:
Great work!

PR #2228, opened 2016-07-13
Make tertiary yellow and thinner after roads code refactoring

### LINKED ISSUE #1974: Use different color for tertiary?

kocio-pl:
Another try to resolve #1974.

Only tertiary colors and sizes are changed manually (other script-generated roads are not touched at all) using settings from from https://github.com/gravitystorm/openstreetmap-carto/pull/2219. 

Fast example (Warsaw, z16):
z0jdjzxu

### DIFF
--- a/roads.mss
+++ b/roads.mss
@@ -1,6 +1,6 @@
 /* For the main linear features, such as roads and railways. */
 
-@tertiary-fill: #ffffff;
+@tertiary-fill: #fafbdd;
 @residential-fill: #ffffff;
 @service-fill: @residential-fill;
 @living-street-fill: #ededed;
@@ -18,7 +18,7 @@
 @helipad-fill: @aeroway-fill;
 
 @default-casing: white;
-@tertiary-casing: #8f8f8f;
+@tertiary-casing: #959a5a;
 @residential-casing: #bbb;
 @road-casing: @residential-casing;
 @service-casing: @residential-casing;
@@ -31,7 +31,7 @@
 @bridleway-casing: @default-casing;
 @track-casing: @default-casing;
 
-@tertiary-shield: #3b3b3b;
+@tertiary-shield: #343d00;
 
 @unimportant-road: @residential-casing;
 
@@ -90,7 +90,7 @@
 @trunk-width-z13:                 6;
 @primary-width-z13:               5;
 @secondary-width-z13:             5;
-@tertiary-width-z13:              4;
+@tertiary-width-z13:              3.75;
 @residential-width-z13:           2.5;
 @living-street-width-z13:         2;
 @pedestrian-width-z13:            2;
@@ -104,7 +104,7 @@
 @steps-width-z13:                 0.7;
 
 @secondary-width-z14:             5;
-@tertiary-width-z14:              5;
+@tertiary-width-z14:              4;
 @residential-width-z14:           3;
 @living-street-width-z14:         3;
 @pedestrian-width-z14:            3;
@@ -130,7 +130,7 @@
 @steps-width-z15:                 3;
 
 @secondary-width-z16:            10;
-@tertiary-width-z16:             10;
+@tertiary-width-z16:              8;
 @residential-width-z16:           6;
 @living-street-width-z16:         6;
 @pedestrian-width-z16:            6;
@@ -145,7 +145,7 @@
 @trunk-width-z17:                18;
 @primary-width-z17:              18;
 @secondary-width-z17:            18;
-@tertiary-width-z17:             18;
+@tertiary-width-z17:             15;
 @residential-width-z17:          12;
 @living-street-width-z17:        12;
 @pedestrian-width-z17:           12;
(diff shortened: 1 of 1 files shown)


kocio-pl:
I don't know any details, but I've just found that test HD style is using tertiary with light yellow filling without changing the rest of a road colors - looks like it covers the whole world, so it can be useful for rough testing (probably only the filling color):

https://osm.rrze.fau.de/testhd.html

math1985:
I actually got used to the wide white tertiaries by now. I don't have a strong opinion whether or not to merge this. I see the more clearly distinguishable tertiaries as an advantage, but I also realize the new tertiary takes up again a part of the colour spectrum that we can't use for anything else.

pnorman:
I'm finding there isn't enough difference between the two yellows. If I see the two side by side I can tell them apart, but if I look an area with a tertiary or secondary surrounded by residentials, I can't tell which it is.

If we change the colour I think this also requires changing the shields. I would much prefer it if the colour changes came via the scripts.

imagico:
> I'm finding there isn't enough difference between the two yellows.

I am inclined to concur.

Of course there is no rule saying secondary and tertiary roads have to be distinguishable by color.

In the long term i'd think changing the admin boundary color and redistributing the road colors to reach more into purple (which right now is not possible because of the boundaries) would create sufficient room for a separate tertiary road color.

kocio-pl:
I think general discussion for tertiary color belongs to https://github.com/gravitystorm/openstreetmap-carto/issues/1974, so there is one place where we can swap ideas and argue which solution should we choose (if any). 

I'd like to keep this PR as a technical ticket to implement the idea of changing **only** tertiary color.

rrzefox:
> I don't know any details, but I've just found that test HD style is using tertiary with light yellow filling without changing the rest of a road colors - looks like it covers the whole world, so it can be useful for rough testing (probably only the filling color): https://osm.rrze.fau.de/testhd.html

Just updated that to v1.41.0 with this PR applied on top (was 1.39.0 with PR #2085 applied before - still is in places that haven't finished rerendering yet).

math1985:
Personally, I got used to the white tertiaries (as I said before). The current PR decreases the contrast between the roads and the background, while I'd rather see the contrast increased.

I'm therefore inclined to reject this PR. Any other opinions?

(Everybody who contributed on this, please don't take this personal, I know quite a lot of work has been put in this.)

kocio-pl:
I put my work to test this idea as much as possible - and still I'm not sure, but that's not my decision, so nothing personal, indeed. =}

mboeringa:
> I'm therefore inclined to reject this PR. **Any other opinions**?

I don't agree. There's little harm implementing it, and it will help others. The final proposal by @kocio-pl is subtle enough to not be disturbing as a change.

jdhoek:
Although I prefer a solution that maintains the colour gradient from motorway to tertiary using the road colours script, I feel that this is already a lot better than white tertiaries.

The problem I have with white tertiaries is that they are only distinguishable when there is not a lot happening on the map and the lines are straight. In a lot of cities though, width alone does not distinguish very well due to the amount of other detail on the map and the curving of the roads.

One particular example I have encountered is with [frontage roads](https://wiki.openstreetmap.org/wiki/Frontage_road) consisting of a tertiary highway flanked by residential highways. These become one broad white blur.

wmyrda:
I would vote yes, but I have two slight issues with it.
On z12 there is no gain from tertiary to appear light yellow as unclassified appears only as thin gray line. That is where I would rather have them stay white. On z13 onward I got used to seeing them white as one can better distinguish them in farmlands. Example here http://www.openstreetmap.org/#map=13/50.0974/22.3769 shows that yellow looks fine where only roads are drawn, but being surrounded by farmlands is whole different story.

kocio-pl:
Well, now I think we should give up this idea. I also got used to white tertiary and I see two problems with making them yellow:

1. We already have a lot of yellow objects - aside from the sand (which may be less of a problem) we use it for secondary roads, schools/hospitals/sport centers and parkings. Since these objects are quite common, it's a bit too much for my taste.

2. I've just noticed that this shade of yellow is in practice identical to the shade of secondary in tunnel - the only difference is border, but one has to spot it. This is a corner case, since roads in tunnels are just a tiny fraction of all the roads, but I don't like this side effect:
yr8zkgnq

math1985:
> Well, now I think we should give up this idea. 

I think both examples you give are instances of a wider problem: much of the colour space has been used already. Even if we manage to solve these two particular problems, we'll likely run into similar problems at different places.

I therefore agree we'll have to give up on this idea.

PR #2232, opened 2016-07-17
Update releasing with information on schema change and reload

pnorman:


matkoniecz:
> rendered off of either a new database or an old one

I am not sure but I think that "off" should be removed.

math1985:
> I am not sure but I think that "off" should be removed.

It is not black and white: http://grammarist.com/usage/off-of/

Anyway I see @pnorman changed his wording so I merged this PR.

PR #2236, opened 2016-07-18
Adding rendering for tourism=artwork

### LINKED ISSUE #855: Add rendering for tourism=artwork
The tag tourism=artwork should be rendered with icon.

See also https://trac.openstreetmap.org/ticket/4529.

kocio-pl:
Resolves #855.

This is rather not the end of the process, since artwork is [very popular tag](http://taginfo.openstreetmap.org/tags/tourism=artwork), but the real code to test rendering. Note that:
- I made it rendering from z17, like standard memorials, but maybe it should be z18+ because...
- ...in some places (like 2) and 6) ) there are more such objects and it causes clutter problem
- in 2) I guess fountain icon should be used instead of the artwork, since this seems to be more important function - but I don't know how to do this
- in 5) maybe memorial should be used instead of the artwork icon, but I'm not sure here (and I also don't know how to do it)

Reporting interesting locations all over the world would be useful for testing.

Sample locations in Warsaw (z18):
1) 
mfayzojd
2) 
mjkxl4p_
3) 
n21ova8d
4) 
rg4hclqm
5) 
weunvhmj
6)
ddk c6rv


dieterdreist:
sent from a phone

> Il giorno 18 lug 2016, alle ore 21:00, kocio-pl <notifications@github.com> ha scritto:
> 
> in 2) I guess fountain icon should be used instead of the artwork, since this seems to be more important function - but I don't know how to do this


yes, fountains can be very important public artwork, likely all of them are somehow artwork, at least those with sculptures (as opposed to nozzles only) as part of them (also other aspects might be artistic, like the composition of water, shapes created by the sprinkling water etc).
Sometimes there will be several pieces in the same fountain, I think the default should be giving rendering priority to the fountain icon and name, unless the artwork has a tourism=attraction flag set ;-)

kocio-pl:
I also thought that tourism=attraction would be useful to show some artworks earlier and others later, but the problem is that artwork and attraction share the same key (tourism=*)...

dieterdreist:
sent from a phone

> Il giorno 18 lug 2016, alle ore 21:24, kocio-pl <notifications@github.com> ha scritto:
> 
> but the problem is that artwork and attraction share the same key (tourism=*)...


yes, hence the ;-)
maybe we should make attraction a flag (attraction=yes/no), I'll propose this on tagging ml

kocio-pl:
Better icon, let's try some of the most crowded artwork-tagged places (worst cases expected on z17):

1) Paris, Louvre (a bit of mess, but not that bad)
z17
dqdakgop
z18
_ny987eu

2) Paris, Jardin du Luxembourg (surprise: it's rather nice, despite the number of artworks, since their location is regular)
z17
cdh1neor

3) Paris, Jardin de Versailles (even better, the same reasons)
z17
a 2uevtk
z18
bm_rmkns

4) Vatican, St. Peter's Square (very good, as above)
z17
hqunx7np
z18
hy8yf81x
z19
5xs_7ze0

In summary - it all looks much better than I expected and currently worst case tagging are in my hometown ([see all zoom range here](https://github.com/gravitystorm/openstreetmap-carto/issues/855#issuecomment-234654990)):

5) Warsaw, Lazienki Park
z17
8jbdxibi

6) Warsaw, Saski Park
z17
kcjs1udm

So the problems are visible if some artworks are gathered together on a small area (Warsaw x2), especially without the visual pattern (Louvre), but if they are distributed somehow (Jardin du Luxembourg), especially more linear (Versailles, Vatican), the result is quite nice, even if there are much more of them. For me it's perfectly acceptable and the icon itself fits there.

I look for more interesting locations for testing.

dieterdreist:
2016-07-23 2:30 GMT+02:00 kocio-pl <notifications@github.com>:

> Better icon, let's try some of the most crowded artwork-tagged places
> (worst cases expected on z17):



a statue is not a suitable icon for all kinds of artwork, it only works
well for statues.

kocio-pl:
The case is not closed, we're still discussing what would be a proper icon, see https://github.com/gravitystorm/openstreetmap-carto/issues/855. This PR is to test these ideas in reality, so if you know some places where you think it really wouldn't work, for example, I'm happy to check them.

BTW: I've noticed that Louvre is an interesting case, since there are a lot of tagged artworks which are _inside_ the building, so we deal with indoor mapping interference there.

dieterdreist:
sent from a phone

> Il giorno 23 lug 2016, alle ore 13:30, kocio-pl <notifications@github.com> ha scritto:
> 
> BTW: I've noticed that Louvre is an interesting case, since there are a lot of tagged artworks which are inside the building, so we deal with indoor mapping interference there.


IMHO these are mistagged with respect to the wiki definition: 'A tag for public pieces of art.' (although this current definition is admittedly not very clear).E.g. I know of a private square with a private artwork, but both are visible and accessible 24/7 by the general public: http://www.openstreetmap.org/node/1311129517#map=19/52.51072/13.37587
(I'll have to add the sculpture in the middle of the fountain, it's this: https://commons.m.wikimedia.org/wiki/File:%22Phoenix%22_Gidon_Graetz_Berlin-Tiergarten_813-695-(118).jpg )

kocio-pl:
8) artwork_03
owec1ett
q7ftfah0

9) 7-like diamond (10 px)
46tn7stv
ebwr0duw

math1985:
Thanks, this looks like the best of the proposed suggestions to me.

kocio-pl:
Have you noticed this code works from z17 (I haven't change this in the meantime)?

I still don't know what how can we make sure that fountain or memorial icons (or other amenities) will be rendered if given object is also tagged as an artwork, because I feel these functions are more important than artistic form. Does anybody know what code change is needed?

math1985:
> Have you noticed this code works from z17

Yes, if it appears to early we can still change this.

> I still don't know what how can we make sure that fountain or memorial icons (or other amenities) will be rendered if given object is also tagged as an artwork, because I feel these functions are more important than artistic form. Does anybody know what code change is needed?

I think you could fix this by changing the order in the COALESCE. Keys can appear multiple times there, see for example man_made.

Issue #989, opened 2014-09-27
amenity=charging_station to use similar render to fuel station


Stalfur:
Issue #108 says to render names of amenities. Charging station is high on the list there but I believe it should be treated similarly to fuel_station, you still hook your car up to a pump like thing only it is electricity and not liquid that flows.

nebulon42:
I have added a charging station icon to [Osmic](https://github.com/nebulon42/osmic): ![charging-station-16](https://cloud.githubusercontent.com/assets/3531092/7711218/5df6178c-fe69-11e4-95d2-13afb71ba53a.png)

kocio-pl:
Great!

brycenesbitt:
@math1985 there are already dedicated mapping sites for this type of feature such as http://www.plugshare.com/ .  50,000 sites.  This is a case where OSM probably won't be the "go to" site for mappers or users.  But showing the feature if mapped is fine.

Barthwo:
Times have changed. Electrical powered cars are coming in numbers. So I would like to see this point with a higher priority. 
The rendering could even use different icons for fast chargers (more than 32 amp) like a filling station and destination chargers (parking lots with a possibility of charging) up to 32 amp with a smaller (?) icon.

math1985:
> Times have changed. 

Agree.

The new problem: won't there be too many of them? In my city, the government has decided to install 1000 additional charging stations this year alone.

nebulon42:
As always this depends on the region. In my region charging stations are present but scarce. I guess choosing a suitable zoom level high enough would work. More detailed and earlier display should be left to specialised maps.

dieterdreist:
sent from a phone

> Il giorno 08 mag 2016, alle ore 10:31, math1985 <notifications@github.com> ha scritto:
> 
> The new problem: won't there be too many of them? In my city, the government has decided to install 1000 additional charging stations this year alone.
> 


are these isolated / distinct locations, because if you have e.g.  5 in a row they will likely render as one.

math1985:
I think these are isolated locations. See http://www.oplaadpalen.nl/. I only expect the number to grow over the next few years.

Barthwo:
amenity=charging_station is defined with a NUMBER of charging points, so it can be rendered as ONE point even if more than one car can use it simultanuously. 
There is a growing number of "fast charging" stations (amperage>32, more than 22 kW), which could be handled similar to fuel stations. 
[...]

PR #2239, opened 2016-07-22
add amenity=charging_station icon

davidfg4:
Addresses #989 and #108 

http://wiki.openstreetmap.org/wiki/Tag:amenity%3Dcharging_station
This adds an icon for amenity=charging_station. If there is a name it should also be rendered, although very few charging stations have a name tag. The icon was taken from https://github.com/gmgeo/osmic per @nebulon42 's suggestion.
There are currently over 8000 charging_station points, and it would be good to be able to view them on the map. See http://overpass-turbo.eu/s/hr4 for where these are currently.

17:

18:

19:


pnorman:
The icon fits in and the PR looks technically okay, but we never decided in #989 if we want to render charging stations or not. Having the PR lets us see how it looks.

kocio-pl:
I don't have an opinion if we should render it or not, but the internal part of the icon (spark) is not clear to me on all these exports - could it be wider a bit?

davidfg4:
Original, wider:
comparison

I think the original has better proportions and is clear enough, but the wider lightning bolt is a bit easier to see. I'd be happy to switch it over if a few people would like.

kocio-pl:
I like wider more, but I think spark could be longer on top (not truncated like it is now).

davidfg4:
Original, wider, taller:
comparison

kocio-pl:
With wider and taller version the sparks are perfectly recognizable for me now, but I think they could be placed a bit lower.

davidfg4:
With that version the bolt is two pixels from the top and three from the bottom. It would look weird below center.

kocio-pl:
I don't think it would, but don't care that much for this detail - it's already an improvement to me.

dieterdreist:
sent from a phone

> Il giorno 22 lug 2016, alle ore 15:14, David Gianforte <notifications@github.com> ha scritto:
> 
> I think the original has better proportions and is clear enough, but the wider lightning bolt is a bit easier to see. I'd be happy to switch it over if a few people would like.

I think wider is better recognizable, maybe you could even try to get rid of the pump and just keep the jack and the bolt (bolt in blue and slightly bigger in this case)

sorcrosc:
> I think wider is better recognizable, maybe you could even try to get rid of the pump and just keep the jack and the bolt (bolt in blue and slightly bigger in this case)

Another way could be to get rid of the bolt and give more focus to the plug?

davidfg4:
I agree, the wider/taller bolt is easier to see.

[The icon we have for fuel](https://github.com/gravitystorm/openstreetmap-carto/blob/master/symbols/fuel.16.svg) has a pump and gas hose, so it'd be nice of the electric charging icon was "similar but different". Looking around the internet I'm seeing similar lightning bolt on a "pump" with cord and plug design that was used in this icon. So it seems to be fairly standard and it would be good to stick with that.

It'd be nice to avoid the "bike shed effect", but if any of you want to explore other ideas feel free to mock them up so we can see.

sorcrosc:
Sorry, it's my last post. The icon is good and I don't know if it could be better but just to explain.
I agree it should be similar to the fuel pump but the characteristic shape for which the pump is recognizable  is already lost, instead it gained two simbols (the bolt and the plug) none of which is so clear

Barthwo:
I would propose an icon with a bigger plug, which is already in use in this POI map: http://www.flosm.de/html/POI-Karte.html#269265.97/49.4232406/6.96490407/0/chargingcar/chargingstation and also used by Osmand: https://github.com/osmandapp/OsmAnd-resources/blob/master/icons/svg/transport/charging_station.svg

morray:
should be added and the bolt symbol is good enough at the moment and should not stop it

math1985:
>  the PR looks technically okay

Are you sure the icon is pixel-aligned? It looks a bit blurry to me.

math1985:
Merged, thanks @davidfg4.

Problems with the icon, if any, could be fixed in a follow-up PR.

davidfg4:
Thanks @math1985.

It is aligned to pixels, but in order to differentiate between the cable the the prongs, the cable is a bit thicker than 1px, and the prongs are slightly less. Again, I didn't design this, it's from https://github.com/gmgeo/osmic

screenshot from 2016-07-25 17-57-34

kocio-pl:
Am I right that [merged version](https://github.com/gravitystorm/openstreetmap-carto/blob/master/symbols/charging_station.16.svg) is the same as original (not [wider and taller](https://cloud.githubusercontent.com/assets/529761/17059861/7097464a-4fe4-11e6-828f-5ab522c67a0c.png))?

davidfg4:
@kocio-pl yes

kocio-pl:
I thought this:

> I agree, the wider/taller bolt is easier to see.

meant you wanted to replace the original with it. Do you really want to do that (and haven't done it yet) or I misunderstood you and you want to stick with it?

davidfg4:
I'm perfectly happy with the original icon too, so at this point I'm not going to make a new PR to change it.

kocio-pl:
I'd like to make some mockups and PR then. Could you share the link to the SVG version?

I'm against bikesheding and also [like to use osmic if possible](https://github.com/gravitystorm/openstreetmap-carto/pull/1615#issuecomment-123024599), but I prefer readability over pureness.

davidfg4:
Sure, here is the larger bolt: http://gianforte.org/charging_station_LARGER_BOLT.16.svg

Issue #2209, opened 2016-07-02
Define a (bigger) minimal font size


sommerluk:
Some text in this style is rendered quite small.

I propose to define a policy with a minimal font size that applies all around the map. I suggest 10 as minimal font size. There is currently some text that is rendered with smaller font sizes; we could render these labels at size 10 (or, if cluttering too much, a zoom level later).
- [x] Describe the policy
- [x] placenames.mss (only “.state” for zoom levels < 7)
- [x] roads.mss (node features)
- [ ] roads.mss (line features at `#roads-text-ref` and `#roads-text-name` and `#paths-text-name`)
- [x] stations.mss
- [x] adressing.mss
- [x] ferry-roads.mss

sommerluk:
Currently, these mss files define a font size smaller than 10: placenames, roads, stations, adressing, ferry-roads.

math1985:
> I propose to define a policy with a minimal font size that applies all around the map.

Good idea.

nebulon42:
Small font sizes are currently hard to read but this depends also on which font is used. With the current font 10 might make sense, but with another font it could also be 9. Overall I would be in favour.

sommerluk:
The only remaining part of this issue are the line features in roads.mss (street names…). There are quite a few usages with size 8 or 9. As this is a line feature, the text is bent, following the rendering of the road. That makes it quite difficult to read for latin text, and terrible for CJK text.

What options do we have?
- Simply drop rendering of the labels that are currently rendered at size 8 or 9. This would mean a significant loose of information, which for sure is not desired.
- Render street names at all zoom levels bigger than now. This would preserve the labels at each zoom level, and it would also preserve the raising font size on raising zoom levels. But is would also mean that roads that are rendered very tall will have a text label that is much thicker than the road itself. That’s also confusing.
- Render street names bigger than now only on zoom levels where they are smaller than 10 – not changing the text size on the other zoom levels. This would limit (but not solve) the problem with labels that are much bigger than the thinkness of the road rendering. But “raising zoom level = raising font size” would be less smooth.

As road names are bent to the geometry of the road, there is no pixel-alligned rendering anyway. Maybe it’s worth to experiment with font sizes like 10.5, 10.8…

All in all, I don’t know the perfect solution for this, and I have no experience in road rendering, so I’m probably not the best person for this.

@matkoniecz Thoughts?

matkoniecz:
[...]

PR #2241, opened 2016-07-24
Font size 10 for junction names

sommerluk:
Resolves part of #2209 

Uses font size 10 at zoom level 14. Other zoom levels unchanged.

Alternative might be to drop rendering of this feature at z14.

Before:
screenshot 1

After:
screenshot 2

### DIFF
--- a/roads.mss
+++ b/roads.mss
@@ -2402,19 +2402,13 @@ tertiary is rendered from z10 and is not included in osm_planet_roads. */
   [highway = 'traffic_signals'] {
     [zoom >= 14] {
       text-name: "[name]";
-      text-size: 8;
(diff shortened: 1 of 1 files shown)


mboeringa:
Looking at the particular example presented here, and especially the top right part, it seems quite clear that maintaining this feature at Z14 will mean some areas of cities across the globe will be plastered with junction names.

I think rendering of junctions at Z14 is only appropriate for the specific **highway=motorway_junction** tag.

sommerluk:
Yes, I tend also to drop it at z14.

math1985:
Thanks, merged. Additional modifications are of course welcome.

Issue #1870, opened 2015-09-24
Render leisure=fitness_station


naoliv:
Could we have [leisure=fitness_station](http://wiki.openstreetmap.org/wiki/Tag:leisure=fitness_station) rendered too?

There is a CC-0 licensed icon here http://osm-icons.org/wiki/File:21009.svg

kocio-pl:
Might work after some tweaking - let's star with just resizing to 14px:
1) ![fitness_station-14-0](https://cloud.githubusercontent.com/assets/5439713/10081584/e2ca6b2a-62f5-11e5-8b92-12791e7efe43.png)

HolgerJeromin:
IMO the treadmill does indicate a indoor gym ([which tagging is not clear](http://wiki.openstreetmap.org/wiki/Gym_/_Fitness_centre)).

You wont find a treadmill at an "outdoor facility" fitness_station

naoliv:
Maybe http://osm-icons.org/wiki/File:21157.svg ?
But it isn't good at 14x14: ![](http://i.imgur.com/ey8SC8M.png)

kocio-pl:
I guess a stepper would be good to depict, because it can be found both indoors and outdoors, so we could have just one icon for both leisure=fitness_station (3 573) and leisure=fitness_centre (2 611):

https://commons.wikimedia.org/wiki/File:Gym_Cardio_Area_Overlooking_Greenery.JPG
https://commons.wikimedia.org/wiki/File:Ansan_034.JPG

boothym:
Both fitness_centre and fitness_station should be rendered, perhaps with some sort of weightlifting or treadmill icon. 

With the gym tagging situation "sorted", I've updated the presets in the iD editor: https://github.com/openstreetmap/iD/issues/3352 - would be nice to get a icon on the map as well.

kocio-pl:
Ideas are welcome, of course, but at this moment we need rather implementations to test them.

boothym:
Well the first one suggested doesn't look too bad at 14px - any chance you could show us what it looks like on the map?

kocio-pl:
I guess if I make it look more like a stepper (change the silhouette) it could actually work for everybody. I will try to do it and show the rendering.

boothym:
Even without the icon can we at least render the name for leisure=fitness_centre on a node, just like sports_centre? At the moment any properly tagged gym is invisible which doesn't help encourage people to tag it correctly.

[...]

PR #2249, opened 2016-07-27
Different color for playground

kocio-pl:
Small tweak for a playground area. Currently playground areas are too visible for me (they stand out in the city) and they are also too blue (a bit like pool). 

After the fix they are light green and are still visible on school and park areas (the most usual places they exist):
exgg1qfd

even when the icon is still not there:
4sqwbf6l

but also work reasonably on other surfaces, like grass or residential area:
5ks4dbme

It's a standalone fix, but can be very useful if we change the water color to be lighter than it is currently (see https://github.com/gravitystorm/openstreetmap-carto/issues/1781#issuecomment-235450267).

### DIFF
--- a/landcover.mss
+++ b/landcover.mss
@@ -41,7 +41,7 @@
 @parking: #f7efb7;
 @place_of_worship: #cdccc9;
 @place_of_worship_outline: #111;
-@playground: #ccfff1;
+@playground: lighten(@park, 5%);
 @power: darken(@industrial, 5%);
 @power-line: darken(@industrial-line, 5%);
 @rest_area: #efc8c8; // also services


imagico:
Don't think this is an improvement - this tone places it quite closely to the other bright greens (park, campsite) and implies vegetation (which is not appropriate).  The old color was distinctly recognizable in the urban context, the one you propose is just another green tone - especially at z14-16.

If you want to weaken the appearance of playgrounds a better way would be to remove or weaken the outline they are drawn with.

polarbearing:
Could we please see a direct before/after comparison, and against the sports facilities?

Please do keep the outline, a playground is a campus-like facility usually with a clear boundary, not just landcover.

math1985:
To me the new color looks good.

It does imply vegetation, but so does the old color.

Since the playground is already marked with an icon, I'm not even sure if playgrounds need an own colour. Would there be no option of using one of the existing colours?

dieterdreist:
2016-07-27 9:25 GMT+02:00 Christoph Hormann <notifications@github.com>:

> If you want to weaken the appearance of playgrounds a better way would be
> to remove or weaken the outline they are drawn with.



I like them standing out. IMHO this style comes from the idea to make some
features outstanding (e.g. sports facilities, universities, playgrounds,
landuse, on low zooms national parks, etc.) and these had been deliberately
chosen. These features originally chosen do not have to remain outstanding
until the end of time, but if we are going to change it, we should be aware
and make a deliberate decision to do so (and should likely also know which
are the features we want to be outstanding).

kocio-pl:
> IMHO this style comes from the idea to make some features outstanding (e.g. sports facilities, universities, playgrounds, landuse, on low zooms national parks, etc.) and these had been deliberately chosen. 

My opinion is quite different:

Well, that's what I'm wondering sometimes - what was chosen to stand out, by whom, when and - most of the time - why? I don't see any clear design and this lack of system makes it hard for me to change things.

The only pattern behind current style I can think of is showing - more or less - typical west town: we have roads, buildings, school, church, parkings, town hall, park, bank, few industrial areas, supermarket (with retail area), pitch, cemetery, sometimes garages area... And all of them have distinct color.

Things are different in bigger city. We have not enough colors to show every feature in a distinct way and most of the areas are mixed up. And now it looks to me like the most important features are playgrounds, pitches and construction areas. 

We've made some changes to this style - for example buildings are softer now (unlike towns, bigger city has a lot of them, with a lot of POIs, and they should not stand out too much). That's what is needed to make this style more universal and less like a theme crafted for towns. We should also use more abstractions and features clustering - for example nobody knows why hospital and schools has the same color, but it looks like we want to know the system, because we have also some other areas which are not shown and there are no new distinct colors available.

Getting back to the playground example: how different are playgrounds from common leisure area (which is green)? What color should we use if we want to show training area for adults (leisure=fitness_station), which are rather similar (smaller area with some exercise features, within park or just in residential area)? Why are they blueish - do they have something in common with water pools?

YMMV, but that's why I want to make these changes.

imagico:
> I don't see any clear design and this lack of system makes it hard for me to change things.

Well - i do see a fairly clear design w.r.t. green area tones as a result of the last years' work.  There are still some non-satisfying cases like the campsite color and of course the new dog_park color but overall this is pretty systematic.

The current track/pitch and playground color are at the edge of the green tones on different sides and sufficiently distinct to be intuitively recognizable as a different class of areas, which is further emphasized by the outline.  Sports areas and playgrounds are among the most distinct purpose oriented open air areas in urban context, you do not for example sit down to read a book, take a nap or have a picnic on a playground without kids or on a soccer pitch.  This is why they have a very distinct color.

The color you propose here on the other hand is just yet another green tone, difficult to distinguish from the other ones and the weakest of all.  The choice appears arbitrary and random and not suited for map users to develop a recognition for these kind of features - at z14-16 hardly anyone will be able to recognize these as a distinct type of area in a similar way as now or as for the track/pitch color.

polarbearing:
It's all possible ;-)
[Sitting and reading on a playground, with no kids](http://il3.picdn.net/shutterstock/videos/9679349/thumb/1.jpg)
[Reading books on a soccer pitch](http://blog.southbronxunited.org/wp-content/uploads/2015/05/IMG_42961.jpg)

kocio-pl:
> Sports areas and playgrounds are among the most distinct purpose oriented open air areas in urban context, you do not for example sit down to read a book, take a nap or have a picnic on a playground without kids or on a soccer pitch. This is why they have a very distinct color.

But why are playgrounds blueish? Water color change made me think about it. And I guess they are distinct just because they were (a) typical for towns and (b) that distinct color was still available. 

How important are they to have distinct color now in the bigger city? For me it's enough they are visible in their primary locations (park, school or residential area) and the icon allows user to make sure what exact feature is this. They are small, so shouldn't be distinct on some early zoom level, as it is now, because those spots tell me "we are the most important things in the city!" - which is not true: they should be less visible than park, river or commercial area. Let's not overuse distinct colors when the icon is enough to know.

>  The choice appears arbitrary and random

For me current situation can be described exactly like this. =} This is of course just my opinion, but I have tried few times to extract some general rules from it and failed.

Road color change proved how hard would it is to fix just a subset of current features and I don't want to make revolution, so I can only tune here and there, cluster features in more abstract classes than currently - of course playground is different than general leisure (that's why they have their own tagging!), but we have limited set of distinct colors - and we have reached this limit already, so we can only reuse (cluster similar things) and use shades, which are not so distinct, but at least help recognize things.

> at z14-16 hardly anyone will be able to recognize these as a distinct type of area in a similar way as now or as for the track/pitch color.

This is way too early IMO:  at z15 in my hometown I see a lot of pitches, but they are mostly just a school pitches, so shouldn't be visible more than schools!

http://www.openstreetmap.org/#map=15/52.2001/21.0209

kocio-pl:
> Since the playground is already marked with an icon, I'm not even sure if playgrounds need an own colour. Would there be no option of using one of the existing colours?

Sure, I can test it, especially on primary playground locations - which one do you think of?

math1985:
> IMHO this style comes from the idea to make some
features outstanding (e.g. sports facilities, universities, playgrounds,
landuse, on low zooms national parks, etc.) and these had been deliberately
chosen.

At least for sport facilities, the color has been chosen in such a way that a building drawn transparently on top of a sports field resulted exactly in a gray building. Buildings used to be dull orange-red, so sport facilities needed to be bright green-blue for this to work. That's a bizarre reason, I know.

polarbearing:
> which one do you think of?

pitches would be the next logical, but there are often pitches next to or within playgrounds.

imagico:
@kocio-pl - I don't think there is a point in further argument here if you don't see the systematics behind the area colors in this style.  I tried to explain my view of these systematics w.r.t. the green tones - not sure what more i can do.  Note i am not opposed to changing this color in principle - there is nothing that requires playgrounds to appear in a bright cyan color but as explained there are a lot of aspects of this color in the current color scheme that make it work quite well.

It all boils down to the fundamental development model - do you make goal oriented changes based on a systematic idea about overall design goals or do you make more or less random trial-and-error point changes based on local considerations with the aim to fix individual problems without a larger picture in mind.

Note this is not a rhetoric question, both models are valid approaches and it is well possible that with the complexity of this style and given a sufficient number of contributors to make changes the latter approach is actually more likely to be successful - kind of a Monte Carlo approach to cartographic design.

But as noted in the past years color consolidation in this style has been done mostly in a systematic way based on overall considerations how changes fit into the larger picture.  This would be a major paradigm change.

@math1985 - the historic reason why certain colors were originally selected and the function they fulfill now, much later and after many changes in colors overall are two very different things.

math1985:
> It all boils down to the fundamental development model - do you make goal oriented changes based on a systematic idea about overall design goals or do you make more or less random trial-and-error point changes based on local considerations with the aim to fix individual problems without a larger picture in mind.

The latter is preferred, of course. Do you have any concrete suggestions?

imagico:
I don't see an acute need to change the playground color and given the dense use of colors a new one is difficult to find.  But consolidating with existing colors would be possible - though IMO not really desirable, it is good if the map distinctly and prominently shows stuff for small kids.  For parents of young children finding a playground is a common map use task.

Consolidation with track/pitch was already mentioned as a possibility.  Another possibility would be with place_of_worship - which is similarly a very distinct purpose open air landuse.  However i think there are already far too many gray tones so using one of them even further is going in the wrong direction.

Since we had until recently separate track/pitch colors it would also be possible to have two different colors in that area - like track and pitch formerly but maybe also somewhat further apart for sports areas and playgrounds respectively.

polarbearing:
> place_of_worship - which is similarly a very distinct purpose open air landuse

disagree, PoW is quite often combined with a building. landuse=religious would be the open air, that is waiting for rendering

imagico:
`amenity=place_of_worship` is only rendered in place_of_worship color (#cdccc9) if used not in combination with building (about half the uses of this tag).

kocio-pl:
> it is good if the map distinctly and prominently shows stuff for small kids. For parents of young children finding a playground is a common map use task.

Sure, but this is general map style, not a "kids" theme. So my main focus would be that smaller features (like this or school pitches) should be visible less/later than bigger and enclosing features (like parks or schools), because they make noise on middle zoom levels. 

Proposed code allows exactly this: first you can still spot such areas, but they are more subtle than bigger features, and later you can find and pick them for sure (by icon), because at the micro- level such details are important and they don't make clutter there.

> It all boils down to the fundamental development model

As you can see, I just use different development model and try to stick to it as close as possible - more to my fundamental issues, less to details. My big goal is generality and visibility priorities, while yours big goal seem to be strict color meaning (if I understand you correctly).

imagico:
You are mixing up development model and development goals here.

kocio-pl:
Since it's so hard to find a common ground on big things, let's get back to details:

Playgrounds in place_of_worship color (#cdccc9):
pimqaue7
ijxknuip
hr8_lqp5

polarbearing:
Nope, that goes in the completely wrong direction.

math1985:
Thanks for trying. Looks bad to me too.

HolgerJeromin:
Joining playground with kindergarten could work to reduce colors.
But as both are different concepts (with and without teachers/doctors in hospitals) I would not prefer it.

math1985:
What about merging playground and pitch?

HolgerJeromin:
Pitch and playground have a lot in common.
Right now I am in the situation needing playgrounds, the icon needs more zooming to differ sport and playground. That's why I am quite happy that a birds view of a region tells me playground areas. 

Merging them would raise the call of sports icons or pitch lines (would love that) again. :-)

kocio-pl:
For me it's worse, since pitch color is even stronger and eclipses not only schools, but also kindergartens and parks, as if they were more important:
u9 reuxb
qu450rva
9rirhgr5

I think about making fitness_station area visible. These 3 features has something in common: leisure and physical effort - playground for small children, pitch for bigger children and fitness station for adults - so it's tempting to unify them somehow. 

What's missing for me is a way to distinguish casual/school pitch from professional/sport pitch, which is not for child physical leisure, but rather for adults to watch ("cultural" leisure, not physical) or play (profession or semi-professional physical activity, not leisure). These should be visible, since are larger, less frequent and more important - so current dark green makes sense for them.

math1985:
> For me it's worse, 

Indeed, I didn't suggest making playgrounds pitch-green, but rather finding a common color for both.

math1985:
I don't really have ideas. Light green might work.

kocio-pl:
Light green for both playgrounds and pitches:

Before
nkib1jk7

After
8vhp_qfy

mboeringa:
I don't really see a need to unify playgrounds and pitches. They are quite different. The average sporter isn't "playing" and neither is the average kid doing sports on a playground...

Other than that, the light green playground colour seems OK to me... better than the current blueish colour that is suggestive of a water related feature.

polarbearing:
> What's missing for me is a way to distinguish casual/school pitch from professional/sport pitch

well you need to propose a tagging scheme for the difference first, and wait until it is adopted. Render differently just based on size will not work as there are always counterexamples.

From the colours we have seen so far, beside the original one, only the light green works for me;
for playgrounds alone or combined with pitches. The blueish for pitches does not work as there is no icon, the features mostly rectangular and thus they look like pools. 

However in case we go to change the pitch colour it would be important to see the pitches in context with the typical sports environments.

kocio-pl:
> Render differently just based on size will not work as there are always counterexamples.

Sure, they are, but we already use this for some features and nobody complains - it's enough to find sane limit. And since I can't find clear criteria to differentiate pitches, the size would work for me. But I see that pitches rendering needs a separate issue.

At this moment I still support my original proposition to make only playgrounds light green. My plan is to use this color also [for leisure=fitness_station (when we create the proper icon)](https://github.com/gravitystorm/openstreetmap-carto/issues/1870).

Issue #214, opened 2013-10-05
Render more specific access tags i.e. bicycle, motor_vehicle etc


joakimfors:
More specific access tags such as bicycle=, motor_vehicle= etc should be rendered.

simonpoole:
I've done some more thinking about this specific can of worms and would suggest that we apply some nomalisation on import to the access tags, likely via LUA transforms. .

This naturally opens another, different can of worms, and, will, to be effective require a reimport, but that is going to be unavoidable anyway. 

The alternatives: adding specific columns or using hstore seem to be overkill given that without going in to a multi-color-pattern frenzy there is no way to render the detailed individual access values anyway. 

I believe doing normalisation on import is likely to be the the lightest weight way of doing this and doesn't require changes to the actual style.

Roughly this would imply mapping
access, vehicle, motor_vehicle and motorcar to access 
some thought needs to be given to both key and value precedence.

dieterdreist:
> Am 21.03.2015 um 13:47 schrieb simonpoole <notifications@github.com>:
> 
> The alternatives: adding specific columns or using hstore seem to be overkill


I propose using hstore, it might be overkill for  this particular problem, but there are lots of other opportunities that such a move would bring us, including support for the alternative public transport scheme, tower types, cuisine types, landuse refinements and basically all more details for objects we already render or can currently not reasonably render because of missing information.

simonpoole:
Am 21.03.2015 um 18:06 schrieb dieterdreist:
>> Am 21.03.2015 um 13:47 schrieb simonpoole <notifications@github.com>:
>>
>> The alternatives: adding specific columns or using hstore seem to be
> overkill
> 
> 
> I propose using hstore, 

Why do you then give us lots of reasons why we would want to avoid using
hstore in all circumstances?

matkoniecz:
After attempting to show this kind of data in a different map style I reached conclusion that it is not doable in a readable way, without making it primary data presented on the map.

kocio-pl:
I also don't believe we have a space for showing such features - even such simple feature as general surface rendering is still not done (see https://github.com/gravitystorm/openstreetmap-carto/pull/2640), so i will close this ticket.

simonpoole:
@kocio-pl I think we would all be happy if correct access tags would simply be treated the same as wrong ones, see https://github.com/gravitystorm/openstreetmap-carto/pull/371 So either don'r render them at all or all the same.

kocio-pl:
[...]

PR #2257, opened 2016-07-30
Change access rendering

### LINKED ISSUE #278: access private casing too thick when used on footpaths, service roads etc
The title says it all really. Private access casing is too wide. Need to split up the SQL up for non-service roads and service roads, paths, footpaths, cycleways, etc
and keep the access fill/casing the same for roads and thinner for paths, etc.
currently it clutters the map way too much.

### LINKED ISSUE #2053: Access restrictions on motorways nearly invisible
bildschirmfoto 2016-02-21 um 17 38 20

Link: https://www.openstreetmap.org/#map=15/50.9541/7.0400

Possible solution: Use a non-transparent color.

math1985:
This pull request proposes changing the look of no/private/destination access.

Currently, the dashes indicating that roads are not accessible are heavy and attract attention to private ways, instead of attracting attention away from them.

In addition, the colours are not intuitive to new users of the map.

This pull request attempts to make access restrictions not as visibly noticeable, while still preserving the information for users that need it. In addition, it aims to make the information more intuitive for a user that is not familiar with the map.

This pull request proposes:
* Gray dashes for roads with access=no and access=private
* Gray dashes for roads with access=destination
* A gray outline for footways with access=no and access=private
* A grayed-out outline for other paths with access=no and access=private

This pull request:
* Resolves #278
* Resolves #2053

### DIFF
--- a/roads.mss
+++ b/roads.mss
@@ -8,14 +8,21 @@
 @raceway-fill: pink;
 @road-fill: #ddd;
 @footway-fill: salmon;
+@footway-fill-noaccess: #bbbbbb;
 @steps-fill: @footway-fill;
+@steps-fill-noaccess: #bbbbbb;
 @cycleway-fill: blue;
+@cycleway-fill-noaccess: #9999ff;
 @bridleway-fill: green;
+@bridleway-fill-noaccess: #aaddaa;
 @track-fill: #996600;
+@track-fill-noaccess: #e2c5bb;
 @aeroway-fill: #bbc;
 @runway-fill: @aeroway-fill;
 @taxiway-fill: @aeroway-fill;
 @helipad-fill: @aeroway-fill;
+@access-marking: #eaeaea;
+@access-marking-living-street: #cccccc;
 
 @default-casing: white;
 @tertiary-casing: #8f8f8f;
@@ -1536,6 +1543,7 @@ tertiary is rendered from z10 and is not included in osm_planet_roads. */
           background/line-opacity: 0.4;
         }
         line/line-color: @steps-fill;
+        [access = 'no'] { line/line-color: @steps-fill-noaccess; }
         line/line-dasharray: 2,1;
         line/line-width: @steps-width-z13;
         [zoom >= 15] { line/line-width:  @steps-width-z15; }
@@ -1554,6 +1562,7 @@ tertiary is rendered from z10 and is not included in osm_planet_roads. */
           background/line-opacity: 0.4;
         }
         line/line-color: @bridleway-fill;
+        [access = 'no'] { line/line-color: @bridleway-fill-noaccess; }
         line/line-dasharray: 4,2;
         line/line-width: @bridleway-width-z13;
         [zoom >= 15] { line/line-width: @bridleway-width-z15; }
@@ -1585,6 +1594,7 @@ tertiary is rendered from z10 and is not included in osm_planet_roads. */
           }
         }
         line/line-color: @footway-fill;
+        [access = 'no'] { line/line-color: @footway-fill-noaccess; }
         line/line-dasharray: 1,3;
         line/line-join: round;
         line/line-cap: round;
@@ -1608,6 +1618,7 @@ tertiary is rendered from z10 and is not included in osm_planet_roads. */
         }
         [zoom >= 15][int_surface = null] {
           line/line-color: @footway-fill;
+          [access = 'no'] { line/line-color: @footway-fill-noaccess; }
           line/line-dasharray: 1,3,2,4;
           line/line-join: round;
           line/line-cap: round;
@@ -1625,6 +1636,7 @@ tertiary is rendered from z10 and is not included in osm_planet_roads. */
         }
         [zoom >= 15][int_surface = 'unpaved'] {
           line/line-color: @footway-fill;
+          [access = 'no'] { line/line-color: @footway-fill-noaccess; }
           line/line-dasharray: 1,4;
           line/line-join: round;
           line/line-cap: round;
@@ -1663,6 +1675,7 @@ tertiary is rendered from z10 and is not included in osm_planet_roads. */
           }
         }
         line/line-color: @cycleway-fill;
+        [access = 'no'] { line/line-color: @cycleway-fill-noaccess; }
         line/line-dasharray: 1,3;
         line/line-join: round;
         line/line-cap: round;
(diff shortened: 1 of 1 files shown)


math1985:
Screenshots:
screenshot_2016-07-30_23-40-06
screenshot_2016-07-30_23-40-42

screenshot_2016-07-30_23-41-23
screenshot_2016-07-30_23-41-50

screenshot_2016-07-30_23-42-55
screenshot_2016-07-30_23-43-21

screenshot_2016-07-30_23-44-14
screenshot_2016-07-30_23-44-31

screenshot_2016-07-30_23-46-08
screenshot_2016-07-30_23-46-24

screenshot_2016-07-30_23-47-11
screenshot_2016-07-30_23-47-27

screenshot_2016-07-30_23-53-33
screenshot_2016-07-30_23-50-41

kocio-pl:
Nice idea! I don't have much to say about specific road solutions, but at least track and footways examples are great (decreasing visibility instead of increasing it!). However maybe we really _want_ some restrictions to be visible in color? Or maybe "x" signs are possible for access=no?

HolgerJeromin:
The decreasing visibility works if they are direct beneath. I am not sure if the access is visible on its own

HolgerJeromin:
Oh, I meant the Footways only

imagico:
I think for roads this looks OK, for tracks and footways it is probably still an improvement compared to the pink but it is problematic in several points:

* weakening the line color for tracks without weakening the casing at the same time looks ugly
* the gray footways will be more or less invisible on structured background (bare_rock/scree)

jojo4u:
The idea is very good. Where are the motorway to tertiary examples? More examples for landuse and natural features are necessary to check visibility.

polarbearing:
In the first post, did you mean "Gray dots ... for destination", not dashes? That's what I see in the examples.

daganzdaanda:
Very good, makes the map much nicer to look at. 

> the gray footways will be more or less invisible on structured background (bare_rock/scree)
Yeah. That may be a problem for the mapper feedback loop... don't know how important this is compared to the reduced clutter in all other instances?

> However maybe we really want some restrictions to be visible in color?
I was wondering the same. But access restrictions are IMHO not important enough to be so very prominent on this basic style "for all purposes". It's very good to be able to see that there is a restriction of some sort, but it doesn't need to shout "look at me"...

HolgerJeromin:
And on the main osm page we have routing which can further help to analyze access.

pnorman:
Looks promising, but I want to have a good look and try it out with the local area.

My concerns are that the grey might be a bit too subtle and light, and footpaths. Footpaths will probably always be a pain for access rendering so we're not going to have perfection there, but it needs to be considered carefully.

I might also review access rendering on other maps to see if there's anything we can learn there

kocio-pl:
@imagico 
> the gray footways will be more or less invisible on structured background (bare_rock/scree)

It doesn't bother me when it's access=no/private, since this "is not to be used by the general public". However rendering in such places would be always worth testing.

pnorman:
I had a look at a residential area with restricted access service roads. I like it on z17+, but I'm concerned that on z16 there's not enough room to put something in the fill

image. 

We may just have to accept that the rendering there won't be great. It's hardly much better now

[image](http://www.openstreetmap.org/#map=16/49.2478/-122.8954)

I know both motorway_link and service are done by putting dashes inside the fill, but they give quite a different impression.

z16
image

z17
image

You don't get the feeling that the difference between a normal motorway(_link)/service and the restricted access ones is the same thing.

pnorman:
I looked at an interstate on/off ramp with no public access. At NE 175th Street, there is public access. From left to right, Bing, Google vector based, esri

imageimageimage


A local bus bay

Here WeGo
image

Other maps were derived from the same cartography as us or showed no difference

pnorman:
In this comment I'm concentrating on non-vehicle ways

### Footpaths/cycleways

v2.42.0/PR for 16/10413/22437
imageimage

All the ways in the area have no surface or tracktype tag, part of Sheep Paddocks Trail has no access.

The changes do a great job at avoiding calling attention to the way making it seem like a more desirable route, but the difference looks very similar to the paved/unpaved difference. I adjusted the data in-DB to get this compairson, but look the cycleway and footpath differences

imageimage
One is a paved/unpaved difference, the other is two paved with an access difference. They're using very similar cartographic language to express different differences.

### Tracks
http://www.openstreetmap.org/#map=17/49.27104/-122.64466 with access tags modified
image

I like this better than the footpath handling, there seems to be a better distinction between tracktype differences and access differences.

v2.42.0/PR
imageimage

### General
If we can't find any solutions to the above I'd be okay with merging as is.

math1985:
I will have a look at this if we can do something do improve the mentioned points, or if this is the best I can do.

simonpoole:
Please save yourself some work and do not render access=destination at all. The reasons have been given many times, but in case they have gone forgotten see for example: https://github.com/gravitystorm/openstreetmap-carto/issues/214 and the may other issues that have been raised on the subject.

pnorman:
Mentioned as an [issue of interest on dev@](https://lists.openstreetmap.org/pipermail/dev/2016-August/029431.html) on the 18th (i.e. between the two comments above)

pnorman:
Destination and no access beside each other

image

I think it's better with a slightly darker inner dash, shown here with `#dddddd`.

image

For z17+ service roads the no access dash pattern is 18px long with about 50% filled. The destination pattern is 9px long with about 25-30% filled. Should the length of the destination be increased? Here's what it looks like changed from `0.1,9` to `0.1,15`

image

Ircama:
Nice initiative.

Just a comment if I am allowed. IMHO the gray color looks fine to reduce noticeability but the dashes might be too thin and long, so I´m wondering whether they could be confused with some kind of two-way road symbol.

dieterdreist:
I think this could be confused with some pedestrian related access feature,
because the same colour is chosen.

pnorman:
> dashes might be too thin and long, so I´m wondering whether they could be confused with some kind of two-way road symbol

I know the type of cartography you're thinking of. I've normally seen it as a thin solid dark line on printed maps. There's some risk of confusion but I think it's okay, and the other options have worse downsides.

pnorman:
> I think this could be confused with some pedestrian related access feature,

Which this do you mean? There's a few things here. Could you provide an example rendering that shows the potential problem?

dieterdreist:
2016-08-27 12:35 GMT+02:00 Paul Norman <notifications@github.com>:

> Which this do you mean?



"because the same colour is chosen.". All of them which use the same colour
than pedestrian areas.

pnorman:
I'm :+1: on merging this as-is and then doing improvements as additional PRs, but I'd first like to deal with some of the other easy PRs before tagging.

math1985:
Fine with me.

math1985:
I had another look at this, and personally I don't really have ideas on how to improve on this. However, pull requests with further tweaks are every welcome.

Issue #1793, opened 2015-08-27
footway line styles difficult to distinguish and in general much weaker than cycleway


imagico:
This is a followup on #1765 which was closed by #1788 - but there are still several problems remaining with the footway styling that should be addressed.

This issue is about problems with the styling of footways and paths - it is not about the tagging semantics this styling is based on - problems with such should be separated.  For specific issues at z13/14 see also #1748.

Specific problems are:

* the three different line styles are very difficult to distinguish, in particular of course in areas with a detailed path network at the lower zooms.  To a slightly lesser extent this also applies to cycleways.
* The footway color is fairly weak for such a thin line feature leading to poor contrast.  This is particularly visible in comparison to the cycleway color which is much darker and more saturated.  The visual effect of this is significantly reduced by #1788 by increasing the footway width for the weaker style variants and decreasing the cycleway width leading to different stylings for both.  But the basic issue of the two colors being very different remains.
* The problem of a very different impression of footways, in particular unpaved, on dark and bright background due to the relatively wide and semitransparent casing remains only slightly reduced through #1788.

Examples, all at z16:

![sample 1](http://www.imagico.de/files/path_new_z16a.png)
![sample 1](http://www.imagico.de/files/path_new_z16b.png)
![sample 1](http://www.imagico.de/files/path_new_z16c.png)

Two suggestions have already been made to address some of these problems:

https://github.com/matkoniecz/openstreetmap-carto/commits/solid
https://github.com/imagico/openstreetmap-carto/commits/path-nosurface

Both have issues that would need resolving.

matkoniecz:
https://github.com/matkoniecz/openstreetmap-carto/commits/solid seems to work well for footways (though testing was limited). Main problem is with cycleways ans is described at https://github.com/gravitystorm/openstreetmap-carto/pull/1788#issuecomment-135167977

https://github.com/imagico/openstreetmap-carto/commits/path-nosurface - is it style that was used to render https://github.com/gravitystorm/openstreetmap-carto/pull/1713#issuecomment-131612727 ?

matkoniecz:
> The problem of a very different impression of footways, in particular unpaved, on dark and bright background due to the relatively wide and semitransparent casing

Casing is a part of the issue but generally nearly everything (especially thin linear features) will have wildly different effect on untagged land, landuse=industrial, natural=bare_rock and landuse=forest. It is a consequence of rendering so many different landcovers as different colours.

[...]

PR #2267, opened 2016-08-04
Allotments pattern

### LINKED ISSUE #1994: landuse=allotments and highway=service
"highway = service" not visible against the background of "landuse = allotments"
mapnik: http://nadoloni.com/c/39.76004,47.46909,14/l/mapnik
mapsurfer: http://nadoloni.com/c/39.76004,47.46909,14/l/ms
http://www.openstreetmap.org/#map=14/47.4672/39.7545

imagico:
This changes the allotments pattern from transparent lines to a orchard-like grid dot pattern.

Closes #1994.

At the same time it slightly brightens the allotments and farmyard base color bringing farmyard on the same level as industrial etc (which seems fine in combination with the lighter farmland now).  This way buildings and other stuff on allotments areas are better visible.

Sample (from [here](http://www.openstreetmap.org/#map=16/48.0205/7.7907)):

allotments pattern


kocio-pl:
Seems nice to me, service roads should be visible now, tracks and pedestrian are OK too, but footways are even less visible than currently (sample from [here](http://www.openstreetmap.org/#map=17/52.19295/21.03874)):
fbdyxlvp

imagico:
IMO footways are not any worse with this change but this depends of course on the background color and orientation in the old styling.  In general footway visibility is a generic problem not specific to this base color, see also #1793

mboeringa:
Have you possibly tried to reverse these two colors? I think the reversed colouring (beige/brown dots, yellowish background), may actually look better (and will almost certainly solve footpath visibility).

kocio-pl:
Another idea: maybe space between the dots could be wider? That would make the pattern less busy.

math1985:
@imagico Let me know what you think of the comments above. If it doesn't work out I'm fine merging the current version.

imagico:
Since this is not intended to be a fundamental change of allotments rendering completely different colors are outside the scope here.  If there is something wrong with the current rendering or the one proposed here in addition to #1994 that should probably be discussed separately.

A coarser pattern would equally be a bigger change and similarly i see no good reason for that at the moment.  Coarse patterns have their own specific issues, especially with fine grained geometries.  If the pattern is considered too strong even though it is definitely weaker than the old one it could be reduced in contrast of course but i would need to see cases where it is actually too strong to consider that.

math1985:
Fine, I merged the PR.

imagico:
Thanks.

kocio-pl:
> If the pattern is considered too strong even though it is definitely weaker than the old one it could be reduced in contrast of course but i would need to see cases where it is actually too strong to consider that.

The whole problem is not about being strong/weak, but about unfortunate interferences with way patterns. This PR resolves the interference with thin lines (service roads), but introduces interference with dots (footways). Contrast change may still help somehow, but one has to test it to know.

I believe footways are more common on allotments than service roads. They were on the edge of visibility already and new version makes them hardly visible plus I see unpleasant visual artifacts, however I don't care that much for details on such areas, so I plan leave it this way for now.

imagico:
> The whole problem is not about being strong/weak, but about unfortunate interferences with way patterns. This PR resolves the interference with thin lines (service roads), but introduces interference with dots (footways).

As said i don't see that.  Footway visibility is a problem in general, was one with the old allotments rendering and still is with the new one but there are plenty of situations where it is far worse (like on bare_rock).  So the way to go here is IMO to improve footway rendering, not to tweak area rendering at the cost of map clarity in other aspects.

I considered various options when designing this PR - including no pattern and a dark pattern.  But this one seemed to be the most suitable.

> plus I see unpleasant visual artifacts on z15:

I can see no artefacts in that sample - would need to explain.

kocio-pl:
> So the way to go here is IMO to improve footway rendering 

That would be the best for me. However I remember that this proved to be really hard (especially https://github.com/gravitystorm/openstreetmap-carto/pull/1359) and we found no solution, so I doubt we can resolve it. If you have some hint, it would be useful.

> I can see no artefacts in that sample - would need to explain.

I don't know how to explain, that's why I posted an image - if you don't see it, it's not a problem for you. I see annoying irregular pattern of light brown dots.

imagico:
For reference - here are the options i considered/tested - overall i regard all of them inferior to the one proposed here, for different reasons and in different situation.  But if you prefer any of them or something completely different feel free to evaluate it further.

weaker pattern
single pixel bright
single pixel dark

kocio-pl:
Thanks for sharing. Could you also provide a code for those attempts (so I can test it with footpaths) or just render it with footpaths examples yourself?

imagico:
There is no code, pattern variations are here:

[variant 1](http://www.imagico.de/files/allotments_2.png)
[variant 2](http://www.imagico.de/files/allotments_3.png)
[variant 3](http://www.imagico.de/files/allotments_4.png)

mboeringa:
Variant 3, with the dark brown spots on the beige background, actually looks more "natural" to me than the original proposal. With its darker "pattern" over a lighter background, it is also more consistent with most other pattern usage in OpenStreetmap-carto, like the forest visible in the example images.

So, as to @imagico 's wishes for "consistency" (https://github.com/gravitystorm/openstreetmap-carto/issues/2270), it might actually be better

Issue #2250, opened 2016-07-27
dog_park base and pattern colors clash with general color system in this style


imagico:
See also discussion in #2216.

The base color 

* is close to park, scrub/golf_course and grass making it hard to distinguish at times but at the same time not clearly a derivative of park (which it was probably intended to be).
* has the highest chroma of all green tones currently in the style intuitively indicating a distinct type of vegetation - which it does not.

The symbol color

* is very dark and strong making it very similar to individual symbols and completely unlike other pattern symbols (like wood/scrub/wetland).

Examples:

http://www.openstreetmap.org/#map=15/48.1231/7.8258
http://www.openstreetmap.org/#map=16/48.7070/9.1655
http://www.openstreetmap.org/#map=16/47.4295/8.5128

dieterdreist:
2016-07-27 10:39 GMT+02:00 Christoph Hormann <notifications@github.com>:

> The base color
>
>    - is close to park, scrub/golf_course and grass making it hard to
>    distinguish at times but at the same time not clearly a derivative of park
>    (which it was probably intended to be).
>    - has the highest chroma of all green tones currently in the style
>    intuitively indicating a distinct type of vegetation - which it does not.
>
> The symbol color
>
>    - is very dark and strong making it very similar to individual symbols
>    and completely unlike other pattern symbols (like wood/scrub/wetland).
>
>


dog parks are (at least in this area) normal parks, usually fenced areas of
regular parks, with the only difference dog owners being allowed to let
their pet off the leash (and the drinking water fountains having a dog head
as a tap, but that's likely a local particularity ;-) ). Hence I'd expect
the dog parks to be very similar in appearance to other parks (like same
fill, but with hatch / symbol).

imagico:
Note the examples i linked to are specifically cases where these are not parts of a regular park.  Looking over occurrences it seems separate dog areas and parts of regular parks are similarly common.

pnorman:
Wow, that's strong.

daganzdaanda:
It's nice that dog parks get rendered - thats something that can show off the data of OSM. But I agree with all points above, the current rendering is not working. 
My proposal would be 
- make the fill exactly the same as a park 
- add a outline similar to marinas etc in a subtle green shade that's not far away from the fill
- make  the paw the same subtle shade as the outline
- maybe increase the distance between the paw-markers

meased:
I've been playing around and came up with this:
[...]

Issue #1794, opened 2015-08-27
leisure=water_park should have same text and icon colour (and same zoom level)


HostedDinner:
In my opinion leisure=water_park should have the same colour for the label and the icon.
Right now the icon is "amenity brown" and the text is "leisure green".
I suggest the green one, because it looks more natural (at least for outdoor swimming pools)
Onother point I have seen is, that the label is getting render at around 15, and the icon at 17. This should be consistent as well.

kocio-pl:
[That's exactly what I think](https://github.com/gravitystorm/openstreetmap-carto/issues/1783#issuecomment-134228799). It's a part of the problem described in #1783.

polarbearing:
This was observed in #1624 as well. However I would tend to see leisure=water_park within the theme_park scope and thus prefer brown icon, label and boundary.

pnorman:
I'd consider it like playground around here

HolgerJeromin:
This should je closed as a dupe #1490

kocio-pl:
It's a different problem (icon/text match vs area outline), however this bug was fixed by #2827.

Issue #1355, opened 2015-03-04
Match POI label colour to background or icon colour?


matthijsmelissen:
For some POI, it both makes sense to match the label colour to the background, and to match it to the icon.

Examples of such POI where this is currently an issue are camp site and playground. If we decide to render areas of hospitals etc., the same issue will arise for such POI.

What would be preferred? Here are a couple of examples for camp site. The first is the current rendering.

![screenshot from 2015-03-04 14 06 04](https://cloud.githubusercontent.com/assets/5251909/6484891/9c30aa4a-c277-11e4-8045-f07b8b86d421.png)

![screenshot from 2015-03-04 14 06 42](https://cloud.githubusercontent.com/assets/5251909/6484901/b13fe022-c277-11e4-8b82-9552d0c328fb.png)

![screenshot from 2015-03-04 14 03 30](https://cloud.githubusercontent.com/assets/5251909/6484847/3f75d000-c277-11e4-8b0b-903fa1bc79ab.png)

![screenshot from 2015-03-04 14 02 06](https://cloud.githubusercontent.com/assets/5251909/6484813/0d9ee184-c277-11e4-8821-6d972ece39ef.png)

nebulon42:
For me different colours for icon and text always looked strange and my first thought was that of a mistake.

I think POIs should have a distinct colour so the last option has too low contrast for me. I'm not sure if changing the font size according to area size while the icon does not change has the desired effect, but I like the font more. But that's more a general dislike of the appearance of DejaVu Sans with a certain letter spacing.

Briefly: I like the first version best.

mboeringa:
I personally find it most intuitive if the color of the icon and accompanying label match, like in examples 1, 3 and 4, as there can be little doubt the icon and label belong to each other, even with more labels and icons in the same location. Matching the color with background color, like in 4, may also be an option, however, you have to consider if you prefer icons with a similar function to have similar color, which may preclude matching with background, or give precedence to the background matching. So function versus consistency with background.

math1985:
Good points above.

Another aspect is that if we keep the text color different from the icon color, the relation between the icon and the area is not clear. For example, in the first picture, it is not clear that the camp sites cover the entire green area. In the last picture, this is much clearer.

sommerluk:
> In the last picture, it's much clearer that the camp site covers the whole area.

+1

mkoniecz:
> In the last picture, it's much clearer that the camp site covers the whole area.

[...]

PR #2268, opened 2016-08-05
Dog park color change

meased:
Continuation of #2216. Attempting to solve #2250.

Render dog parks like playgrounds with a paw pattern background.

dogparkinpark
(with barrier=fence)

dogpark

dog_park_node


dieterdreist:
sent from a phone

> Il giorno 05 ago 2016, alle ore 05:15, meased <notifications@github.com> ha scritto:
> 
> Render dog parks like playgrounds with a paw pattern background.


while dog parks and playgrounds could be seen as similar features from a conceptual point of view, for practical usage this idea seems counterproductive. It would lead to playgrounds and dog parks look the same without the pattern (in lower zoom levels) and therefore neither serve dog owners nor parents

kocio-pl:
@dieterdreist You're right about this effect, but for me it's highly preferable to [hide details in lower zoom levels](https://en.wikipedia.org/wiki/Cartographic_generalization). I would definitely like to recognize them in micro scale, but on lower zoom levels only bigger/more complex structures should be visible, like parks and schools, not the playgrounds, dog parks or pitches, which are common parts of them.

@meased What about different colors for icon and label on points?

meased:
Text and icon are different colors only because we currently render playgrounds on nodes that way. The reason I went with playground color for areas was to avoid making another green. Campground color may be another option.

kocio-pl:
Oh, you're right! For me this version is worth merging as it is, but we still have to fix the general color inconsistency for leisure label/icon (see https://github.com/gravitystorm/openstreetmap-carto/issues/1794).

math1985:
> @meased What about different colors for icon and label on points?

Is this related to #1355?

kocio-pl:
No, I was talking only about matching icon and label colors, they don't have to match background color for me (look at the hospital colors for icon+label vs background).

math1985:
Thanks! Looks good to me too.

michaelzangl:
This was a feature I missed for a long time.

In Karlsruhe, we have dog park areas that e.g. are covered by trees. They are not rendered the right way since dog parks are rendered as solid textures - the same way the land cover is.

I suggest changing the texture background to be transparent or semi-transparent and increasing the z-index of dog park so that it is displayed above the landcover.

Example:
http://www.openstreetmap.org/#map=19/48.99757/8.37287

kocio-pl:
@gravitystorm thinks that transparency is bad solution, so it's just out of question currently. We use simple heuristic however: smaller areas are rendered on bigger areas, so if the trees are just a part of dog park, they would cover it (see in the park), but it seems that in your case the dog park is small area compared to one big area with trees, so it is rendered on top.

michaelzangl:
I know that heuristic. I think it makes perfect sense for "normal" areas. But a dog park - at least in our town, is more like a natural resort marked by signs and does not necessary need to be a separate area.

dieterdreist:
sent from a phone

> Il giorno 16 ago 2016, alle ore 17:45, kocio-pl <notifications@github.com> ha scritto:
> 
> @gravitystorm thinks that transparency is bad solution


I think he is only against alpha values >0 and <1 because of "unpredictable" colors, while 0 and 1 are both OK.

PR #2145, opened 2016-06-05
Added rendering of tag tourism=wilderness_hut

Ircama:
This PR squashes all commits in PR #2140, which can be closed.

**Added rendering of tag tourism=wilderness_hut**

Added specific symbols for basic_hut and wilderness_hut vs. generic amenity=shelter.

Notice that this drop shall be consistent with the other push related to zooms, where alpine_hut symbol (the main one) is shown at zoom>=13.

To summarize, the following different kind of huts have specific symbols now:

- amenity=shelter, shelter_type=basic_hut (zoom>=13)
- amenity=shelter (only shown at zoom >=16)
- tourism=alpine_hut  (not modified within this drop)
- tourism=wilderness_hut (zoom>=13)

alpine_hut: symbols/alpinehut.p.16.png
wilderness_hut: symbols/wilderness_hut.svg
generic shelter: symbols/shelter-14.svg
amenity=shelter with shelter_type=basic_hut: symbols/basic_hut.svg

PR #2144, opened 2016-06-05
Topo extension hut

Ircama:
This PR squashes all commits in PR #2140, which can be closed.

**Added rendering of tag tourism=wilderness_hut**

Added specific symbols for basic_hut and wilderness_hut vs. generic amenity=shelter.

Notice that this drop shall be consistent with the other push related to zooms, where alpine_hut symbol (the main one) is shown at zoom>=13.

To summarize, the following different kind of huts have specific symbols now:

- amenity=shelter, shelter_type=basic_hut (zoom>=13)
- amenity=shelter (only shown at zoom >=16)
- tourism=alpine_hut  (not modified within this drop)
- tourism=wilderness_hut (zoom>=13)

alpine_hut: symbols/alpinehut.p.16.png
wilderness_hut: symbols/wilderness_hut.svg
generic shelter: symbols/shelter-14.svg
amenity=shelter with shelter_type=basic_hut: symbols/basic_hut.svg

PR #2140, opened 2016-05-30
Added rendering of tag tourism=wilderness_hut

Ircama:

PR #2279, opened 2016-08-13
Add rendering for tourism=wilderness_hut

### LINKED ISSUE #540: Add rendering for tourism=wilderness_hut
tourism=wilderness_hut points are not displayed.

There is an icon here:  http://wiki.openstreetmap.org/wiki/File:Wilderness_hut_blue_stove_p32.svg

Ircama:
**Add rendering of tag tourism=wilderness_hut**
[Edited summary for this PR, left it out on my first push]
The aim of this PR is to propose a rendering for the tag tourism=wilderness_hut as suggested in the Wiki at
http://wiki.openstreetmap.org/wiki/Tag:tourism%3Dwilderness_hut
and as per related rendering proposal at
http://wiki.openstreetmap.org/wiki/Proposed_features/wilderness_mountain_buildings#Rendering

It references issue #540 and attempts to improve unmerged PRs #2140, #2144 and #2145. Over the last unmerged commit, the differences are:
- Removed usage of shelter_type, which can be introduced again when hstore is available
- Removed basic_hut.svg, to be used with shelter_type
- Rebase to master

To summarize, the following different kind of huts have specific symbols now:
- amenity=shelter (only shown at zoom >=16)
- tourism=alpine_hut  (not modified within this drop)
- tourism=wilderness_hut (zoom>=13)

alpine_hut: symbols/alpinehut.p.16.png
wilderness_hut: symbols/wilderness_hut.svg
generic shelter: symbols/shelter.svg

Resolves  #540


math1985:
Is it just me, or is there something wrong with the SVG? It doesn't seem properly centered.

pnorman:
> Is it just me, or is there something wrong with the SVG? It doesn't seem properly centered

It's not just you.

pnorman:
With only 2.2k uses I'm against adding an icon for wilderness huts.

The SVG also has some technical issues such as defining a colour. It should leave the colour undefined and then that can be done in the MSS.

pnorman:
Oh, are the units meters or mercator meters? If the former, it allows for rendering based on the true scale of the map. The latter is what is normal for web maps.

Ircama:
Again, I suggest addressing a feature that is important for mountain areas, rendering it as defined in the wiki:

http://wiki.openstreetmap.org/wiki/Tag:tourism%3Dwilderness_hut
http://wiki.openstreetmap.org/wiki/Proposed_features/wilderness_mountain_buildings#Rendering

I'll revise the SVG and will remove the tab in the related marker-file.

pnorman:
> Again, I suggest addressing a feature that is important for mountain areas, rendering it as defined in the wiki:

I know it's defined on the wiki and the SVG is from there, but the wiki doesn't define what we render or what icons we use. Usage is one factor to consider, and I think the usage here is too low.

math1985:
I would be in favour of rendering it, but with the same icon as alpine_hut.

Ircama:
I am aware that wilderness hut has a limited usage at the moment.

Provided that considering the number of uses for a new feature to be rendered is absolutely correct, I would suggest to also evaluate the importance of a feature within its context (and also the relevance of its naming for the geography) rather than systematically checking minimum usage as a rule. In the case of wilderness hut within a mountain area, this is one of the 10/20 features which are always available in common topographic maps for its importance to users as a reference for orientation and for identifying places (other than its usage as a hut). E.g. best practices of rendering already adopted for traditional maps might be another factor to evaluate.

As already mentioned, I also see a relation between low usage of a feature and missed rendering in OSM, especially for mountain areas which have significant space of improvement for both feature definition and related rendering. In the case of wilderness hut, its current use in OSM clearly is orders of magnitude lower than reality.

As per using the same icon as alpine_hut, I would suggest differentiation because of the diverse purposes of these features, as reported in the related wiki pages (alpine huts are typically managed for most of the time, while wilderness huts are generally unmanaged and this is important information for users).

imagico:
As indicated before the base design of the symbol is very different from the shelter/alpine_hut design.  This makes it confusing.  As a minor variant of the same base design it would make sense to me, even for relatively low volume use.  However there is a competing tagging for this kind of feature in form of `amenity=shelter` + `shelter_type=basic_hut` with nearly identical definition:

http://wiki.openstreetmap.org/wiki/Tag:shelter_type%3Dbasic_hut

Ircama:
IMHO the description and related proposal of icons in the Wiki for the different types of mountain buildings is pretty appropriate and I do not see a real overlapping between `wilderness_hut` and `amenity=shelter` + `shelter_type=basic_hut` (ref. bullets in bold in http://wiki.openstreetmap.org/wiki/Tag:tourism%3Dwilderness_hut, http://wiki.openstreetmap.org/wiki/Tag:shelter_type%3Dbasic_hut). Anyway, the latter cannot be rendered at the moment due to the need of *shelter_type* column.

To me, identifying five types of mountain buildings, together with their appropriate symbols, should orient mappers to improve data quality. OSM will be able to show improved information for mountain buildings vs. standard topographic maps as soon as all these symbols are rendered.

Differentiating the symbol of the generic shelter vs. the other ones might facilitate some additional accuracy in the building definition (anyway at the moment the adopted shelter symbol is similar to `amenity=shelter` + `shelter_type=lean_to` and I have not modified it in this PR).

Considering that for `tourism=alpine_hut` at the end we went for zoom >= 14, also `tourism=wilderness_hut` should have the same min zoom. I'll also rectify this.

Ircama:
wilderness

Ircama:
Icon revised meanwhile: now the SVG shape of wilderness_hut should look more similar to the PNG currently used for alpin_hut.

Resolves #540

wh

imagico:
Changed symbol is a better fit.

Still think as long as `amenity=shelter` + `shelter_type=basic_hut` is not rendered in a distinct way or like `tourism=wilderness_hut` the latter should not be rendered different from `amenity=shelter`.  Those two are very similar, the only difference apparently being the existence of an indoor fire place.

Otherwise the rendering will likely change the meaning of the tags and `tourism=wilderness_hut` will get used as a general _sophisticated shelter_ tag instead of the specific meaning it is meant to have.

Ircama:
I agree that the differences might appear slight at first glance, nevertheless they exist and have a relevance IMHO to describe mountain buildings. In fact the wiki distinguishes them and clearly proposes separate rendering icons.

In my interpretation a wilderness_hut (as per http://wiki.openstreetmap.org/wiki/Tag:tourism%3Dwilderness_hut) is basically an unmanaged remote masonry (or wood) building used as a shelter. It could be a “malga” o “baita”.

shelter_type=basic_hut (as reported in http://wiki.openstreetmap.org/wiki/Tag:shelter_type%3Dbasic_hut) should typically be a Bivouac shelter or “bivacco” or Biwakschachtel.

You might find plenty of pictures by searching these names online and also the wiki reports some of them.

Due to the different significance of these two kinds of landmark, usually they are separately represented in standard mountain maps and generally appear as important symbols in a legenda. These adopted practices would also corroborate a low risk of misinterpretation of symbols. Databases of these mountain buildings already exist with pretty accurate documentation; I would not see a concrete possibility of tagging mismatch by mappers or possible confusion among basic_hut, wilderness_hut and alpine_hut.

I think that, by appropriately rendering these buildings as suggested by the wiki, we could provide increased value on the documentation of unpopulated mountain areas and this is at the end the aim of my proposal submitted to your evaluation.

dieterdreist:
sent from a phone

> Il giorno 31 ago 2016, alle ore 00:02, Christoph Hormann <notifications@github.com> ha scritto:
> 
> Those two are very similar, the only difference apparently being the existence of an indoor fire place.


definitely having an indoor fireplace vs. definitely not having one is a HUGE difference. I'm not sure if the not having a fireplace is guaranteed though.

Ircama:
> definitely having an indoor fireplace vs. definitely not having one is a HUGE difference. I'm not sure if the not having a fireplace is guaranteed though.

Even if the wilderness_hut symbol does represent a sort of stove with a chimney, the wiki provides some additional description and rather clear example photos showing that the presence of a fireplace would not be the single difference between such buildings. In my point of view the classification made by the wiki attempts to reflect commonly adopted distinctions made for their social relevance and which the traditional maps generally also report to characterize these isolated landmarks that use to be objective references for local people as well as for tourism, sport and more generally land-use management by authorities and organizations targeted to territorial culture and countryside conservation.

imagico:
The point i was trying to make is that creating a three class rendering:

* `tourism=alpine_hut`
* `tourism=wilderness_hut`
* `amenity=shelter`

will likely result in a three class tagging practice with the following meanings

* `tourism=alpine_hut` - managed huts
* `tourism=wilderness_hut` - any slightly sophisticated shelter
* `amenity=shelter` - other shelters

which is not the suggested meaning of these tags.

A two class rendering:

* `tourism=alpine_hut`
* `amenity=shelter` or `tourism=wilderness_hut`

a four class rendering:

* `tourism=alpine_hut`
* `tourism=wilderness_hut`
* `amenity=shelter` + `shelter_type=basic_hut`
* other `amenity=shelter`

or a different three class rendering:

* `tourism=alpine_hut`
* `tourism=wilderness_hut` or `amenity=shelter` + `shelter_type=basic_hut`
* other `amenity=shelter`

would all create a better incentive for correct tagging IMO.  Of these currently only the two class rendering is feasible here so i think for the moment this would be the preferable solution.

dieterdreist:
sent from a phone

> Il giorno 31 ago 2016, alle ore 11:42, Christoph Hormann <notifications@github.com> ha scritto:
> 
> would all create a better incentive for correct tagging IMO. Of these currently only the two class rendering is feasible here so i think for the moment this would be the preferable solution.


a shelter is any kind of protection, including laterally open roofs etc., while any kind of hut is a kind of building with a closed room/space

Ircama:
I am personally convinced that differentiating symbols would be better incentive for correct tagging, as provides visual support on revisions.

I also think that the four class rendering is the one that better addresses the proposals in the Wiki and this could be reached in two steps: one would be the rendering of tourism=wilderness_hut (current PR) and the other (as soon as the tools will make it possible) would be the differentiation of symbols among amenity=shelter + shelter_type=basic_hut and other  amenity=shelter including lean_to (as per Wiki proposals).

I anyway respect the different positions by the experienced contributors, commenters and revisors in this repository and am fully open to find a synthesis on an adoptable next step.

Gazer75:
Would love seeing this rendered. It would make the map more useful to trekkers in Norway for sure.
It would display all huts that are self service or no service.

Currently there is a bit of a mess with some huts being tagged as alpine and some as wilderness.
Most likely due to the lack of rendering.

Alpine hut to me is a serviced(staffed) hut where people can take a shower and such things.

https://english.dnt.no/about-the-cabins/

geowOSM:
I personally would be in favour of rendering wilderness_hut, but with a different icon than alpine_hut.
There is a fundamental difference - wilderness huts are self serviced and don't provide staff to service food, beverages etc . They also might be locked and require picking up a key prior to the ascent.

Ircama:
Rebased

Ircama:
wilderness_hut

nebulon42:
I planned to provide a sequence of shelter icons for some time now. But didn't find the right solution, maybe I was expecting too much. So I'm definitely in favour of rendering these. The small number of uses is not really a problem because of their importance in remote areas. I'm also ok with the zoom level.

@imagico re your concerns: A basic hut or Biwakschachtel is usually very different from a more sophisticated shelter as you named it. So I think the distinction the Wiki makes is correct and non-overlapping. I have found mis-taggings around alpine_huts and wilderness_huts and confusion with chalets, but this is likely a Wiki problem. Generally, the tagging scheme is not ideal but that's what exists. Not being currently able to differentiate shelter_types is a drawback, but should not hinder the inclusion of this.

The icon has a few issues which I plan to address. Examples: no compound shape, 580px as size, chimney could be better visible, not pixel aligned everywhere. At the same time I can also work on the SVG replacement of alpine_hut.

So currently not mergeable due to icon issues.
@Ircama I will provide you an updated icon you can include if you want.

imagico:
All right - my concern that this kind of rendering could lead to use of tourism=wilderness_hut bleeding into the domain of amenity=shelter still exists but i won't insist i can predict the future better than everyone else can. :smile:

Ircama:
@nebulon42: thanks very much for taking care. I will rely on your feedback or you can also take over, just as you prefer.

nebulon42:
I have now created icons for alpine_hut, wilderness_hut, basic_shelter and shelter. This should be discussed together in my opinion.
Find them here: https://github.com/gmgeo/osmic/tree/master/outdoor
wilderness_hut is immediately usable, I'm not so sure about alpine_hut, but it will be the basic shape with something in it. Still have to test it on the map. Maybe we also should think of slightly altering the shape of the current shelter icon so that is different from the other tree.

@Ircama if you take this icon width and height in CartoCSS should become unneccessary.

Ircama:
Icons are beautiful. Just some concern with alpine_hut which is quite different from the one currently in place. I will update wilderness_hut and test it.

Ircama:
Just some preliminary sample based on @nebulon42's work before pushing the code.

Shelter:
hut1

wilderness_hut (also alpine_hut with original PNG icon):
hut2

wilderness_hut (alpine_hut has the new SVG icon): it will possibly not be part of this PR, but I am showing here a sample to allow comparing:
hut3

This is the code I used in the last sample to test both new icons of alpine_hut and wilderness_hut:

```javascript
  [feature = 'tourism_alpine_hut'][zoom >= 13] {
    marker-file: url('symbols/alpine-hut-14.svg');
    marker-placement: interior;
    marker-clip: false;
  }

  [feature = 'tourism_wilderness_hut'][zoom >= 13],
  [feature = 'amenity_shelter'][zoom >= 16] {
    marker-file: url('symbols/shelter-14.svg');
    [feature = 'tourism_wilderness_hut'] {
      marker-file: url('symbols/wilderness-hut-14.svg');
    }
    marker-placement: interior;
    marker-clip: false;
  }
```

I would suggest the original icon for wilderness huts (subjective opinion maybe).

On this PR, I consider appropriate the new wilderness_hut and shelter icons and suggest both are part of the new commit.

Please let me know, thanks.

imagico:
I am not sure if the different shelter icon is a good idea - but this should be a separate PR anyway, keep this limited to adding wilderness_hut please.

Ircama:
> keep this limited to adding wilderness_hut please.

Ok, no problem.

nebulon42:
> keep this limited to adding wilderness_hut please.

Ah, that's what I had in mind too. Sorry for being not clear on this.

Ircama:
This is a test commit+rebase, where I updated wilderness_hut with @nebulon42’s icon keeping alpine_hut and shelter-14.svg unchanged, as requested.

It also generates project.mml with Windows and with https://github.com/gravitystorm/openstreetmap-carto/pull/2459 (just project.mml, without updating the script program). I see that Travis CI does not like it (possibly because the tags are not in the expected order).

I will do again a commit+rebase with the original scripts\yaml2mml.py and complete with a sample image for the standard shelter.

Ircama:
Just would like to receive confirmation that the tag `id="wilderness-hut"` in wilderness_hut.svg is appropriate (dash instead of underline).

nebulon42:
@Ircama I think that the id is just a minor detail. Is this now ready from your side?

Ircama:
@nebulon42: code ready for revision/merge

nebulon42:
Thanks! I manually merged this, because I had to change width/height of the SVG to the respective pixel values.

Just a note: While it is good to have meaningful commit messages they don't need to be too extensive. No need to include the complete history of the PR into the commit message.

Gazer75:
Noticed a conflict here in Norway, because there are two types of wilderness huts. Some have provisions and some don't. To add this info a shop=provisions is added. This overrides the new hut rendering by thinking its a generic shop.

nebulon42:
@Gazer75 Interesting observation, very helpful. Please open a separate ticket for it.

Ircama:
> To add this info a shop=provisions is added

I'll investigate. Could you please provide some example meanwhile? Is it documented in the wiki?

Gazer75:
Example node: https://www.openstreetmap.org/node/666788642

Wiki mention tagging for huts in Norway here(scroll up half a page to "Turisthytte"):
https://wiki.openstreetmap.org/wiki/Import/Catalogue/N50_import_(Norway)#Changeset_tags

The Norwegian Trekking Association(DNT) maintains around 500 cabins/huts around the country
https://english.dnt.no/

Map of the DNT cabins(red icon) and other lodging(blue) can be seen here:
https://www.ut.no/kart/
This map works mostly ok with Chrome and its page translation.
Most, if not all, shown on this map are also in OSM as DNT members have added them over the years.

Ircama:
@Gazer75 I see now @nebulon42's recommendation to open a new ticket; I also would suggest this. After a quick review, I see that your request is related to a massive import from Norway about tourist huts and I do not see it clearly documented in the main Wiki (apart from a mention in the [Norwegian one](http://wiki.openstreetmap.org/wiki/No:Map_Features#Overnattingshytter) without English note). But it would be appropriate that such investigation is performed in the new ticket.

PR #2281, opened 2016-08-13
Do not label waterway inside tunnel

giggls:
Rendering labels of rivers flowing through canalizations is currently somewhat ugly.
Here is an example:
http://www.openstreetmap.org/#map=18/49.05799/8.52062

Thus I made this trivial change to German style. Probably also of interest to upstream.

### DIFF
--- a/water.mss
+++ b/water.mss
@@ -201,7 +201,7 @@
 }
 
 #water-lines-text {
-  [lock != 'yes'] {
+  [lock != 'yes'][int_tunnel != 'yes'] {
     [waterway = 'river'][zoom >= 13] {
       text-name: "[name]";
       text-face-name: @oblique-fonts;


dieterdreist:
sent from a phone

> Il giorno 13 ago 2016, alle ore 15:27, Sven Geggus <notifications@github.com> ha scritto:
> 
> Rendering labels of rivers flowing through canalizations is currently somewhat ugly.
> Here is an example:
> http://www.openstreetmap.org/#map=18/49.05799/8.52062
> 
> Thus I made this trivial change to German style. Probably also of interest to upstream.
> 


I think the label is useful (agree it's ugly when overlapping other features), but the way is too thick, should be like the waterway without the tunnel, i.e. a single dashed line

kocio-pl:
I believe it's a corner case and not worth changing. I also think thin blue line could be too similar to current bicycle path rendering (like [here](http://www.openstreetmap.org/#map=19/52.28534/20.94065)).

giggls:
It probably is a corner case. An Option for the rendering of canalizations themselves (not the labels) would probably be to render "casing only" like I do on proposed highways in German style. See attached tile to see what I mean.

kocio-pl:
This image is just too small for me - I don't hardly see the water here... =}

giggls:
Huh? you are not supposed to look for water in this image. I haven been talking about a casing only rendering like I do it on proposed highways. They are not dashed like I would do it with canalizations.

math1985:
Thanks for this pull request @giggls! I agree it improves rendering where applicable.

PR #2285, opened 2016-08-16
Add specific symbol for peaks with cross

Ircama:
**Add specific symbol for peaks with cross**

This PR supersedes https://github.com/gravitystorm/openstreetmap-carto/pull/2139 for the specific relation to peaks (natural=peak) with cross (man_made=cross), which are now rendered with a cross starting from zoom level 13. Vs. the previous PR, the marker-transform feature has been replaced by specific icons.

Check sample images here below.
peak-cross-12
peak-cross-13
peak-cross-14
peak-cross-15
peak-cross-16
peak-cross-17


imagico:
For info:

* `natural=peak` + `man_made=cross` on the same node: 272 uses
* `natural=peak` + `summit:cross` on the same node: 1507 uses
* separate mapping of cross and peak: 1268 uses within 250m with `man_made=cross`, probably many more with `historic=wayside_cross`.

Ircama:
@imagico: this feature (which is an appropriate territory description as well as a useful identification of relevant mounts) is in fact rarely used by taggers, even if, at least in the zones I know in Italy, almost all mountains have a cross in their major peak. Could be another element supporting the considerations we already made: mappers tend to tag what they can then see on the map; besides and more generally, undeveloped areas (currently grossly rendered) are neglected by OSM mappers at the moment.

pnorman:
I'm against adding another symbol which will have so few uses.

Ircama:
@pnorman: why "will have"? It was rarely used as not rendered, but when shown I expect that the usage will highly increase.

pnorman:
> @pnorman: why "will have"?

I mean that if added it will render 272 times - we look at current usage, not potential future usage.

I'm also concerned that people would convert separate mapping of cross and peak into tagging the same node, even when the cross isn't at the very peak.

mboeringa:
> I'm also concerned that people would convert **separate mapping of cross and peak into tagging the same node**, even when the cross isn't at the very peak.

From a cartographic and render point of view, **_I think it actually highly desirable that his happens_**. 

From a cartographic point of view, peaks and peaks with crosses are most useful at medium scale ranges (e.g. 1:100k-1:10k). At these scales, space for displaying objects that are very close to each others, is severely limited.

Not having both tagged on single node, even though essentially on the same peak, would lead to two separate objects (a symbol for a peak and one for a cross) cluttering up the map and potentially (depending on the map renderer), overlapping symbols as well. 

E.g. it is hardly useful information to see a peak symbol with height 1075m and name, and then see a second overlapping cross symbol with height 1080m of stacked almost on top in a 1:50k map or Z13 view of the rendered map...

dieterdreist:
sent from a phone

> Il giorno 18 ago 2016, alle ore 11:40, mboeringa <notifications@github.com> ha scritto:
> 
> From a cartographic and render point of view, I think it actually highly desirable that his happens.


-1, these are 2 distinct features and have to be mapped on their own. If you think it's better to have one object for both, you could propose a combined feature, e.g. add an attribute to natural=peak (has_cross=yes/no)

Ircama:
Here the OSM wiki indicates how to render a peak with a cross:
http://wiki.openstreetmap.org/wiki/Tag:man_made=cross
| *For crosses on peaks use natural=peak + man_made=cross+ summit:cross=yes*
this PR simply tries to follow it, fixing a missing feature.

As regards *summit:cross=yes* when used without *man_made=cross*, this is impossible to be rendered at the moment due to the current policy to avoid adding new columns and to the unavailability of hstore.

math1985:
> these are 2 distinct features and have to be mapped on their own. 

+1

I also don't find the symbol recognizable enough.

Sorry @Ircama!

Ircama:
Could you please elaborate how http://wiki.openstreetmap.org/wiki/Tag:man_made%3Dcross can be rendered vs. this closed PR other than increasing the size of the symbol as @math1985 suggested?

PR #2287, opened 2016-08-17
Revised text size for lakes

Ircama:
**Revised text size for lakes**

This PR supersedes #2139 for the specific relation to the text size used for lakes (natural=water or landuse=reservoir/basin).

Check sample images here below.

New rendering with zoom=10
water-new-10

Old rendering with zoom=11
water-old-11

New rendering with zoom=11
water-new-11

Old rendering with zoom=12
water-old-12

New rendering with zoom=12
water-new-12

Old rendering with zoom=13
water-old-13

New rendering with zoom=13
water-new-13

Old rendering with zoom=14
water-old-14

New rendering with zoom=14
water-new-14

Old rendering with zoom=15
water-old-15

New rendering with zoom=15
water-new-15

Old rendering with zoom=16
water-old-16

New rendering with zoom=16
water-new-16

Old rendering with zoom=17
water-old-17

New rendering with zoom=17
water-new-17


nebulon42:
Looks like an improvement to me. Only the step from z13 to z14 might be a bit big.

Ircama:
@nebulon42: thanks for your comment. Specifically in that example the different text size of z13 vs. z14 is due to the *text-placements* tag, which increases the probability of the text to be rendered at the cost of reducing its text size.
Figure *New rendering with zoom=13* shows a rather complex interior point to be rendered by Mapnik (curve shape of the lake with a relatively big island inside). At z13 Mapnik would have skipped rendering the *name* tag at text-size=25 if the *text-placements* tag had not been added (same happens at z12, please check the figures: it is not rendered with this PR but should not be considered a regression as also currently not rendered; also the text size reduction at z13 does not produce an evident regression in the example, as compared with *Old rendering with zoom=13*).

PR #2289, opened 2016-08-18
Add a Code of Conduct

pnorman:
This adds a code of conduct, based on the Go code of conduct. The goal of a code of conduct is to make explicit inappropriate behavior and mechanisms for dealing with it.

In selecting the Go code of conduct as a starting point, I looked at about 25 options, looking for those which covered the problems we have had and would work with a small scale project, or could be made to work by eliminating content instead of substantial rewrites.[1]

I was initially expecting to do something based on the [Debian code of conduct](https://www.debian.org/code_of_conduct) but I found it less useful than the Values section of the Go document. It is much easier to remove unnecessary sections than expand a document.

## Modifications
Most of the modifications are because the Go process does not [scale down to small projects](https://github.com/golang/proposal/blob/master/design/13073-code-of-conduct.md#open-issues). We have four maintainers and one discussion venue - Github. Go has orders of magnitude more people involved and has IRC channels, mailing lists, mutiple GitHub projects, a subreddit, and in-person events.

## To Do
* [ ] Add links to the document
* [ ] Link to it from CONTRIBUTING.md

## References

[Go CoC proposal rationale](https://github.com/golang/proposal/blob/master/design/13073-code-of-conduct.md#rationale)
[Go CoC blog post/talk](https://blog.golang.org/open-source#TOC_7.)

[1]: Debian, FreeBSD, Go, Joomla, and Puppet met the initial filtering.

### DIFF
--- a/CODE_OF_CONDUCT.md
+++ b/CODE_OF_CONDUCT.md
@@ -62,7 +62,7 @@ When using the OpenStreetMap Carto spaces you should act in the spirit of the va
 
 The OpenStreetMap Carto maintainers are responsible for handling conduct-related issues. Their goal is to de-escalate conflicts and try to resolve issues to the satisfaction of all parties.
 
-If you encounter a conduct-related issue, you should report it to the maintainers by email.
+If you encounter a conduct-related issue, you should report it to the maintainers by sending them [all an email](mailto:openstreetmap-carto@gravitystorm.co.uk,penorman@mac.com,info@matthijsmelissen.nl,matkoniecz@gmail.com).
 
 **Note that the goal of the Code of Conduct and the maintainers is to resolve conflicts in the most harmonious way possible.** We hope that in most cases issues may be resolved through polite discussion and mutual agreement. Bans and other forceful measures are to be employed only as a last resort.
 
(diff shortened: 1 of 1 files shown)


jojo4u:
I read the text and welcome this addition.

math1985:
Looks fine to me. I can merge this or wait for the todo-items to be resolved.

pnorman:
> I can merge this or wait for the todo-items to be resolved.

I'm okay either way, but because this is a policy matter instead of a normal PR I'd like feedback or a go-ahead from @gravitystorm and @matkoniecz

gravitystorm:
I'm broadly in favour of this. I can see two small things:

* To be in line with other OpenStreetMap policies, I favour "indefinite" bans rather than "permanent" bans. 
* We will need to publish email addresses for the maintainers

pnorman:
> To be in line with other OpenStreetMap policies, I favour "indefinite" bans rather than "permanent" bans.

We changed to permanent in the DWG ban policy.

pnorman:
> We will need to publish email addresses for the maintainers

Should we set up an alias? We don't often need to communicate privately, only every other month or so, but it would be nice to just send to one email.

Issue #1197, opened 2015-01-05
landuse=construction colour is too dominant


daganzdaanda:
This has been left out from the last round of recalibrating colours. Along with leisure=pitch / track (see #1190), construction sites now are pretty dominant on the map. 

Possible solution: desaturate and lighten a lot. 
If it gets too close to forests or similar, maybe move the chroma a bit (green -> red?). Or think about a broad diagonal hatching.

math1985:
Agree, this does not look good now.

How do other maps render construction sites?

mboeringa:
> How do other maps render construction sites?

See the red arrows for a couple of examples, very light yellow with random light coloured stipple. Berlin seems one huge construction site...

![landuse_construction](https://cloud.githubusercontent.com/assets/7635726/5616716/883acf22-9507-11e4-8e25-35c8129ad7db.jpg)

math1985:
Is anyone willing to propose a colour? I think @mboeringa's rendering is not very clear.

althio:
I like the proposal of diagonal hatching. Just like military landuse as it
bears the same idea of restricted access.
Color: yellow for works instead of red.

gravitystorm:
I dislike the diagonal hatching, in general and certainly for construction sites. It's very eye catching (in 'real life' it's used for things like hazard warning tape, danger signs etc) and we should avoid using it here.

The main use of hatching, in military landuse, I also dislike. It should be reserved for things like military danger areas rather than things like ceremonial parade grounds.

althio:
Hatching is indeed very eye-catching but hazard warning tape and danger signs are really well suited for a construction site so IMO the meaning is good and clear.

You could keep hatching but also make it as (in)visible as you wish if you adjust saturation and opacity. You could even render buildings or other features below the hatching as the tagging allow to define the intented use of the construction site.

However I do agree that hatching with added fill in the gaps as in military landuse is too proeminent.
A good combination could be:
* Hatching with muted color,
* no fill in the gaps,
* gaps bigger than stripes.
Yellow for construction, redish for military.

gravitystorm:
Hazard warning might be good for the construction site itself, but not for a general-purpose map where construction sites are just one of hundreds of other features. There's no reason to make them the most eye-catching feature on the map.

nebulon42:
As a side note I discovered that construction does not have an outline, unlike most other landuses.

polarbearing:
[...]

PR #2292, opened 2016-08-19
Less dominant construction color

kocio-pl:
Resolves https://github.com/gravitystorm/openstreetmap-carto/issues/1197.

As originally proposed by @daganzdaanda, made 10% lighter with 5% desaturation. Lightening stops hinting that construction areas are more important than buildings and desaturation takes away the golden tint, which is currently overused in osm-carto, making also construction area more distinct than lightened-only version.

1) City landscape, comparison with bigger buildings
Before
ldlg22vf

After
4ejz73xi

2) City detail, comparison with buildings, trees and special buildings (tagged as place of worship)
Before
dp7zxvik

After
fdegunzf

3) City detail, comparison with buildings and landuse=garages
cvh2xwnv

4) City detail, buildings on construction area
u2erjogr

### DIFF
--- a/landcover.mss
+++ b/landcover.mss
@@ -33,7 +33,7 @@
 @bare_ground: #eee5dc;
 @campsite: #def6c0; // also caravan_site, picnic_site
 @cemetery: #aacbaf; // also grave_yard
-@construction: #b6b592;
+@construction: #c7c7b4;
 @danger_area: pink;
 @garages: #dfddce;
 @heath: #d6d99f;


EdLoach:
"Lightening stops hinting that construction areas are more important than buildings" - except that construction areas are more important than buildings, as they indicate to mappers that they need regular re-surveying to check on progress. Of course these days they are usually supplemented by notes with comments used to give progress updates, so how they look on the tiles is less important than it was. I have no real opinion on how they should look though.

kocio-pl:
General map should be careful with what it shows as important. In my opinion it should rather _explain_ the space than deal with _searching_ things. Mappers should use data-related tools (like [Overpass query](http://overpass-turbo.eu/s/hWm) or special layer) to not overlook smaller construction sites.

PR #2294, opened 2016-08-19
Revised min. zoom for tourism=alpine_hut

Ircama:
**Revised min. zoom for tourism=alpine_hut**

This PR supersedes #2139 specifically for the functionality to reduce the min. zoom to show the text name of alpine huts (tourism=alpine_hut).

Sample images here below.

z13 only shows icons of alpin huts without names
ah_13

z14 after applying the PR shows names of alpin huts
ah_14_new

z14 currently does not show namesof alpin huts
ah_14_old


kocio-pl:
I suppose this is safe, because alpine huts probably exist only in areas with low density of POIs. 

Problems:
-  color of labels is much different than icon color - did you use standard osm-carto code for this rendering?
- 9 commits for one-liner change? You should probably make it just one.

Ircama:
> color of labels is much different than icon color - did you use standard osm-carto code for this rendering?

I did not change any color; the only change was the zoom for 'tourism_alpine_hut'

> 9 commits for one-liner change? You should probably make it just one.

How can I do this? When issuing `git rebase -i master` I only get one commit (which is the right one):
`pick 2508628 Revised min. zoom for tourism=alpine_hut`

pnorman:
> How can I do this? When issuing `git rebase -i master` I only get one commit (which is the right one):
`pick 2508628 Revised min. zoom for tourism=alpine_hut`

Your local master branch probably has extra commits relative to that of this repo

PR #2296, opened 2016-08-20
Better rendering of religious features

Ircama:
**Better rendering of religious features**

This PR supersedes #2139 for the rendering of religious buildings including related text:

    amenity=place_of_worship
    man_made=campanile
    building=shrine
    building=synagogue
    building=temple
    building=mosque
    building=church
    building=chapel
    building=cathedral

Sample images here below.

Before @z16:
z16-old

After @ z16:
z16-new

Before @z16:
z16-old2

After @ z16:
z16-new2


z16:
For almost an hour I looked around through this repository, wondering when I ever signed up for it. For the life of me I couldn't remember even hearing about this project before. Nor was I watching this repository, nor did I star it. Took me way too long to realize why I got an email notification for this PR…

polarbearing:
Could you please explain in more detail what you are proposing?

From a brief look at the changed files, it looks that you are treating a building=church or building=mosque as an amenity=place_of_worship, even if that does not carry the latter tag, and add religious symbols? It also appears that you deduce the religious symbol just from the building type and not only from the religion=* tagging?

If so, that would be fundamentally wrong, as there are such buildings that currently do not function as a PoW, but instead as concert hall, climbing tower or residential home etc.

HolgerJeromin:
And it would probably result in less tagging with religion tags what would not be good for the rest of the data users.

project.yaml
```diff
@@ -155,6 +155,8 @@ Layer:
                                                     'golf_course', 'miniature_golf', 'picnic_table', 'sports_centre', 'stadium', 'pitch', 
                                                     'track', 'dog_park') THEN leisure ELSE NULL END)) AS leisure,
               ('military_' || (CASE WHEN military IN ('danger_area') THEN military ELSE NULL END)) AS military,
+              ('building_' || (CASE WHEN building IN ('church', 'shrine', 'synagogue', 'temple', 'mosque', 'chapel', 'cathedral') THEN building ELSE NULL END)) AS building,
+              ('man_made_' || (CASE WHEN man_made IN ('tower', 'campanile') THEN man_made ELSE NULL END)) AS man_made,
```
pnorman: where does man_made=tower come into this?

pnorman:
You have a lot of unnecessary commits in this branch. Could you rebase it on master?

If you have `origin` set up as a git remote for gravitystorm/openstreetmap-carto and `ircama` for yours, you should be able to do this by saving/stashing your work and commands something like

```sh
git fetch origin
git checkout topo-extension-religious-features
git reset --hard origin/master
git cherry-pick 5666bf27d3f9967f86ea9e9c12a1d8d29292c180
git push -f -u ircama topo-extension-religious-features
```

You might want to also fix your master branch so it is in sync with upstream with

```
git fetch origin
git checkout master
git reset --hard origin/master
git push -f ircama master
```

I haven't tested the git commands myself, so they might need some adjustment.

-----

With only 600 uses I'm against adding `man_made=campanile`. I also don't think it's appropriate to use the same symbol for a bell tower near a place of worship as the place of worship itself.

I'm unsure on rendering church buildings like places of worship. How does this work if the church building is also a different type of amenity that is rendered?

Circeus:
> With only 600 uses I'm against adding man_made=campanile. I also don't think it's appropriate to use the same symbol for a bell tower near a place of worship as the place of worship itself.

Plus those should probably be man_made=tower subcases, shouldn't they?

polarbearing:
> Plus those should probably be man_made=tower subcases, shouldn't they?

Absolutely, there are already 8565 [tower:type](https://wiki.openstreetmap.org/wiki/Key:tower:type)=bell_tower, 105 tower:type=campanile and even 3 
tower:type=klockstapel.

Ircama:
@polarbearing, @HolgerJeromin, @pnorman:

This PR was already contained in the old #2139 and I separated the different elements included in that PR to allow commenting.

Your considerations are appropriate: *amenity=place_of_worship* shall be needed for these tags to be rendered as religious features, so this PR should be closed (or at least revised).

My intent was anyway confined to suggest rendering elements that in mountain areas are useful for orientation or for referencing places. As *building=Church* is not rendered at the moment, a remarkable number of such buildings remain currently hidden to users (this can be easily noticed by comparing OSM-carto with traditional topographic maps, which do not represent the actual scope of these features, but simply highlight the church buildings with symbol and related name).

It might be arguable to avoid representing features that are not actual places of worship if they are anyway built as a church (an unchurched building actually used as warehouse can be a reference in any case if having a church shape and being in relevant position, e.g. at the fork of two paths or at the top of a hill). Provided that, by analyzing some samples, I also noticed that lots are errors (e.g., missing *amenity=place_of_worship*).

Considering that in total about 470000 features are *building=Church* without *amenity=place_of_worship* (about a fifth of *building=Church* with *amenity=place_of_worship*), so the involved number is really relevant (and especially concentrated in Europe, e.g. 150000 just in Italy), I remain inclined to consider that those features might be rendered in some way (at least for the polygon-fill, at least for the name).

The above picture represents *building=Church* without *amenity=place_of_worship* in Lombardy; they are not rendered at the moment:

building church

Ircama:
> You have a lot of unnecessary commits in this branch. Could you rebase it on master?

Thanks a lot, my master branch was not in sync with upstream, I apologize.

Ircama:
> For almost an hour I looked around through this repository, wondering when I ever signed up for it. 

@z16: I apologize, I wrongly used @ z16 to mention "at zoom=16" without interposing a space. It took some time also to me to understand what happened.

math1985:
I find the parallel-tagging systems amenity=place_of_worship versus building=church a bit weird and confusing, to be honest.

In The Netherlands, we have loads of churches, like [this one](https://nl.wikipedia.org/wiki/Fatih-moskee_(Amsterdam), that were built as churches but converted to mosques. Should they be tagged as building=church, amenity=place_of_worship, religion=islam? I highly doubt we should render such places with a cross.

math1985:
> The above picture represents building=Church without amenity=place_of_worship in Lombardy; they are not rendered at the moment:

Do you have a rough idea how much of these are tagging errors, i.e. how many of these should in fact have the amenity=place_of_worship tag?

aceman444:
It may be that the amenity=place_of_worship is tagged on a node inside the building=church. OR in the university/hospital way, on an area around the building.

dieterdreist:
sent from a phone

> Il giorno 22 ago 2016, alle ore 01:09, Matthijs Melissen <notifications@github.com> ha scritto:
> 
> Should they be tagged as building=church, amenity=place_of_worship, religion=islam? I highly doubt we should render such places with a cross.


+1, after the conversion they can surely be considered mosques, no? Therefore building=mosque in your example

dieterdreist:
sent from a phone

> Il giorno 22 ago 2016, alle ore 01:11, Matthijs Melissen <notifications@github.com> ha scritto:
> 
> Do you have a rough idea how much of these are tagging errors, i.e. how many of these should in fact have the amenity=place_of_worship tag?


there are many mappers deliberately separating the building from the pow, although in the case of churches I am not sure if such a separation is a better representation of reality than mapping both on the same object.

math1985:
> Therefore building=mosque in your example

The architectural style is still 'church' though. Also wiki defines building=church as 'A building that was built as a Wikipedia-16px.png church'.

polarbearing:
> mountain areas ... As building=Church is not rendered at the moment, a remarkable number of such buildings remain currently hidden to users

Incorrect, they are rendered, as any other building, just not as building-major. If that is the an isolated building on a mountain, that is still suitable as orientation point as in your intention.

math1985:
So the question basically is, should we render non-active churches with a cross? What do we think?

pnorman:
> So the question basically is, should we render non-active churches with a cross? What do we think?

I'm against it.

math1985:
I'm afraid I'm not convinced either. I think it's best to not accept this proposal.

SomeoneElseOSM:
In the UK it'd look pretty odd for some former church buildings to be displayed with the cross (e.g. the one that are currenty pubs or mosques or museums).  You could argue that a building=church is an "important recogisable building for navigation" (though I'm not entirely convinced), but many are no longer used for religious purposes.  However, other countries may differ significantly, of course.

Ircama:
After some additional time spent this week in analyzing the OSM database and the wiki, I think that all comments on this PR are right. So thanks very much for your time in your revision.

There is a huge number of incomplete data on features like churches, especially concentrated in some regions (possibly including a number of massive imports in some cases), which leaded to consider that a building like a church (e.g., due to its architectural design) should have been highlighted. They are not really data errors, but just incomplete tagging. It is the worship (e.g., amenity tagging) which should highlight the monument, as currently defined in the Wiki.

There are also two other elements in this PR, which I would propose to possibly comment:
- render the polygon-fill of amenity=place_of_worship at zoom=14 vs. zoom=15
- render the name for amenity=place_of_worship at zoom=16 vs. zoom=17

The examples in this PR would show it might be appropriate at least for isolated buildings and uncrowded built areas.

Would it be useful to evaluate a separate PR with some example for populated, very populated cities and unpopulated places at different latitudes? Thanks

PR #2300, opened 2016-08-21
Comments about fonts

sommerluk:
Add some comments about fonts. Implements part of #2209 

No rendering change.


math1985:
Thanks! Commenting existing code is always very welcome.

PR #2301, opened 2016-08-21
Revised min. zoom for tourism=alpine_hut

Ircama:
Revised min. zoom for tourism=alpine_hut

This PR supersedes #2139 and #2294 specifically for the functionality to reduce the min. zoom to show the text name of alpine huts (tourism=alpine_hut).


math1985:
Thanks!

PR #2302, opened 2016-08-21
Revised text size for lakes

Ircama:
**Revised text size for lakes**

[Edited summary]

This PR proposes to scale up the font size for large water area labels (natural=water or landuse=reservoir/basin) used for lakes (currently they are all rendered with a fixed font size, using text-size: 12).

It supersedes and attempts to improve unmerged PRs #2139 and #2287 for the specific relation to the text size used for water areas.

### DIFF
--- a/water.mss
+++ b/water.mss
@@ -1,17 +1,13 @@
 @water-text: #6699cc;
 @glacier: #ddecec;
 @glacier-line: #9cf;
-@water-font-size-small: 10;
-@water-wrap-width-size-small: 25;
-@water-font-size: 12;
-@water-wrap-width-size: 30;
-@water-font-size-big: 15;
+
+@water-font-size: 10;
+@water-font-size-big: 12;
+@water-font-size-bigger: 15;
+@water-wrap-width-size: 25;
 @water-wrap-width-size-big: 35;
-@water-font-size-bigger48k: 17;
-@water-wrap-width-size-bigger48k: 45;
-@water-font-size-bigger1000k: 22;
-@water-wrap-width-size-bigger1000k: 150;
-@water-font-size-bigger1500k: 27;
+@water-wrap-width-size-bigger: 45;
 
 #water-areas {
   [natural = 'glacier']::natural {
@@ -296,32 +292,21 @@
     [zoom >= 10][way_pixels > 3000],
     [zoom >= 17] {
       text-name: "[name]";
-      text-size: @water-font-size-small;
-      text-fill: @water-text;
-      text-face-name: @oblique-fonts;
-      text-halo-radius: @standard-halo-radius;
-      text-halo-fill: @standard-halo-fill;
-      text-wrap-width: @water-wrap-width-size-small;
-      text-placement: interior;
-      [way_pixels > 4000] {
       text-size: @water-font-size;
       text-wrap-width: @water-wrap-width-size;
-      }
       [way_pixels > 12000] {
         text-size: @water-font-size-big;
         text-wrap-width: @water-wrap-width-size-big;
       }
       [way_pixels > 48000] {
-        text-size: @water-font-size-bigger48k;
-        text-wrap-width: @water-wrap-width-size-bigger48k;
+        text-size: @water-font-size-bigger;
+        text-wrap-width: @water-wrap-width-size-bigger;
       }
-      [way_pixels > 1000000] {
-        text-size: @water-font-size-bigger1000k;
-        text-wrap-width: @water-wrap-width-size-bigger1000k;
-      }
-      [way_pixels > 1500000] {
-        text-size: @water-font-size-bigger1500k;
-      }
+      text-fill: @water-text;
+      text-face-name: @oblique-fonts;
+      text-halo-radius: @standard-halo-radius;
+      text-halo-fill: @standard-halo-fill;
+      text-placement: interior;
     }
   }
 }
--- a/water.mss
+++ b/water.mss
@@ -1,6 +1,17 @@
 @water-text: #6699cc;
 @glacier: #ddecec;
 @glacier-line: #9cf;
+@water-font-size-small: 10;
+@water-wrap-width-size-small: 25;
+@water-font-size: 12;
+@water-wrap-width-size: 30;
+@water-font-size-big: 15;
+@water-wrap-width-size-big: 35;
+@water-font-size-bigger48k: 17;
+@water-wrap-width-size-bigger48k: 45;
+@water-font-size-bigger1000k: 22;
+@water-wrap-width-size-bigger1000k: 150;
+@water-font-size-bigger1500k: 27;
 
 #water-areas {
   [natural = 'glacier']::natural {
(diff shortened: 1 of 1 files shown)


imagico:
I am not so familiar with the Mapnik label placement parameters - do these here assure that the center points of offset labels are still within the polygon area?

In general with the way_pixels limits it would IMO make sense to further relax the zoom limits, i.e. allow labels for sufficiently large lakes already from z8/z7, maybe with a somewhat raised lower way_pixels limit.

pnorman:
This would be our first use of `text-placements` so it'll need more testing than normal to make sure it's not cut off at MT edges.

math1985:
@Ircama The `text-wrap-width` would also need to increase with increasing font-size.

Everything else looks fine to me.

@pnorman How are we going to test that? We have seen quite different behaviour on production and on development servers in this respect.

pnorman:
> @pnorman How are we going to test that? We have seen quite different behaviour on production and on development servers in this respect.

Kosmtik/TM tends to be **worse** for metatile issues because it has smaller metatiles. If we don't see too many issues in development it should be fine in production. But this will need testing on Mapnik 2.2 and Mapnik 3.0.

Ircama:
@math1985
> @Ircama The text-wrap-width would also need to increase with increasing font-size.

OK I'll revise this parameter and perform tests.

@imagico 
> In general with the way_pixels limits it would IMO make sense to further relax the zoom limits,

I think that currently no text is shown for lakes with zoom < 10.

Surely big lakes like the Great Lakes in North America (as well as Caspian Sea, Black Sea, etc.) need to be managed but would be better with some specific PR.

pnorman:
Can you change the text size logic so it's the same as landcover? See https://github.com/gravitystorm/openstreetmap-carto/blob/master/amenity-points.mss#L1455-L1475 for an example

e.g.
```mss
    [zoom >= 9][way_pixels > 3000],
    [zoom >= 17] {
      text-name: "[name]";
      text-size: @landcover-font-size;
      text-wrap-width: @landcover-wrap-width-size;
      [way_pixels > 12000] {
        text-size: @landcover-font-size-big;
        text-wrap-width: @landcover-wrap-width-size-big;
      }
      [way_pixels > 48000] {
        text-size: @landcover-font-size-bigger;
        text-wrap-width: @landcover-wrap-width-size-bigger;
      }
      text-fill: darken(@danger_area, 40%);
      text-face-name: @bold-fonts;
      text-placement: interior;
    }
```

I'm not finding many MT issues, but I did find one in [Kelowna](http://www.openstreetmap.org/#map=11/49.8712/-119.4015)

image
z10

image
z11

image
z12

image
z13

This needs more investigation, but I'll do that after the way_pixel MSS changes.

It is also disconcerting to see the text size shrink as you zoom in.

Ircama:
@pnorman: thanks for your review; I'll certainly compare landcover logic through a new update.

Ircama:
> It is also disconcerting to see the text size shrink as you zoom in.

Already discussed here:
https://github.com/gravitystorm/openstreetmap-carto/pull/2287#issuecomment-240877027

The idea of considering shrinking text size was to avoid of not showing the text at certain zooms in case of limited space. Anyway, if this is not considered appropriate, I'll remove the *text-placements* logic.

Ircama:
Code revised so that the text size logic is similar to the landcover example in amenity-points.mss.
Kept 6 levels of way_pixels where two of them are the same as the landcover example in amenity-points.mss.
Removed the text-placements logic.

Sample images with some lake sizes at different zooms:

Zoom 9
z9

Zoom 10
z10

Zoom 11
z11

z11b

Zoom 12
z12

z12b

Zoom 13
z13

z13b

Zoom 14
z14

z14b

Zoom 15
z15

Zoom 16
z16

Note: I used Kosmtik for testing (and found important rendering differences between Tilemill for Windows due to the very different Mapnik versions)

pnorman:
> Code revised so that the text size logic is similar to the landcover example in amenity-points.mss.

Is there a reason to not use the exact way_pixel thresholds and values for text-size and text-wrap-width?

pnorman:
ah, one potential reason is landcover starts at 10px and water at 12. Still, I'd prefer to reuse thresholds and variables when possible.

Ircama:
> ah, one potential reason is landcover starts at 10px and water at 12. Still, I'd prefer to reuse thresholds and variables when possible.

Progression used in amenity-points.mss:

- default font size 10, wrap 25
- for way_pixels > 12000: font size 12, wrap 35
- for way_pixels > 48000: font size 15, wrap 45

Settings previously used for water rendering:

- fixed font size 12, wrap 30

What I did by now was the following:

- The default is the same as what used in amenity-points.mss (font size 10, wrap 25)
- for way_pixels > 4000: font size 10->12, wrap 25->30 (the current default)

All next progressions have increased size for a better representation (as far as it looked to me):

- for way_pixels > 12000: font size 12->15, wrap 35
- for way_pixels > 48000: font size 15->17, wrap 45
- for way_pixels > 1000000: font size 15->22, wrap 45->150
- for way_pixels > 1500000: font size 15->27, wrap 45->150

There is no particular difference for zooms <=13, but for zooms 14+ the result looks better.

Anyway, I will provide a new commit with the exact logic of amenity-points.mss as well as similar samples to allow comparison.

Ircama:
The code has been revised in this second commit, so that the text size logic is *exactly* the same as [landcover](https://github.com/gravitystorm/openstreetmap-carto/blob/master/amenity-points.mss#L1455-L1475) in order to allow comparison with the rendering from the first commit.

The following images IMHO would confirm that for zooms 14+ the result of the first commit looks better especially for bigger lakes.

Images show exactly the same zones of the previous commit (FYI, z9 has been left out as no text is shown).

Zoom 10:
z10

Zoom 11:
z11
z11b

Zoom 12:
z12
z12b

Zoom 13:
z13
z13b

Zoom 14:
z14
z14b

Zoom 15:
z15

Zoom 16:
z16

imagico:
For islands labeling is capped at 800k pixels - it might make sense to do the same for lakes.

Ircama:
> For islands labeling is capped at 800k pixels - it might make sense to do the same for lakes.

Considering that generally the center of a lake should not be too crowded of data (while an island can host a city), there should be no particular prevention to adopt a bigger font size at high zoom level.

In the testing pictures shown above there are four lakes: Lago di Como (biggest), Lago d'Iseo, Lago di Endine, Lago d'Idro.

Here are their *way_pixels* at different zooms:

Lago di Como (large):

zoom|way_pixels|
|---|----------:|
|z10|12878,1
|z11|51512,2
|z12|206049
|z13|824191
|z14|3296780
|z15|13187100
|z16|52748300
|z17|210993000

Lago d'Iseo (medium):

|zoom|way_pixels|
|---|----------:|
|z10|5382,31
|z11|-
|z12|86116,9
|z13|344466
|z14|1377870
|z15|5511490
|z16|22045900
|z17|88183400

Lago di Endine (small):

|zoom|way_pixels|
|---|----------:|
|z12|3029,61
|z13|not shown
|z14|48473.7
|z15|193895
|z16|775578
|z17|3102310

Lago d'Idro (small):

|zoom|way_pixels|
|---|----------:|
z10|not shown
z11|3854,84
z12|-
z13|-
z14|246680
z15|986764
z16|3947040
z17|15788200

By capping way_pixels to 800000, you do not see "Lago di Como" at z13.
You do not see Lago d'Iseo at z14. Lago di Idro will be shown until zoom = 14.
Lago di Endine will not be shown at z16.

I would suggest reconsidering my first commit.

math1985:
Thanks!

Issue #1772, opened 2015-08-20
Add rendering for landuse=plant_nursery


polarbearing:
For [landuse=plant_nursery](http://wiki.openstreetmap.org/wiki/Tag%3Alanduse%3Dplant_nursery), there was an old PR #187 closed because the pattern was not suitable. Still the feature needs rendering, it now has [7.5k](http://taginfo.openstreetmap.org/tags/landuse=plant_nursery) uses. 

I'd say just a fill colour and a name would be suitable, the colour the same as an orchard. Icon would be too tricky to understand (grown trees were difficult enough), the purpose should be recognisable from the name.

[Here is one](http://www.openstreetmap.org/way/9759196#map=17/52.54648/13.27951).

HolgerJeromin:
yes, this would help to show how large this garden_centre is
http://www.openstreetmap.org/way/225030927
landuse=commercial would be another way, but plant_nursery is fitting better.

imagico:
Could use a more fine grained version of the orchard pattern, like [this](http://www.imagico.de/map/jsdotpattern.php#x,128,jdp32360;gv,6,32,32;rd,1,0,0,pixel,0.125,4,4,0,jdp23814,749d6c,aedea3;).

polarbearing:
Usage has grown to 8.4k since opening, people are asking on the tagging list. 

Another possibility to render would be identical to ```greenhouse_horticulture```.

kocio-pl:
Now the usage is 12541 (lines only). I like original proposition - orchard color is already distinct, so using it without the pattern would be enough for me.

PR #2304, opened 2016-08-25
add plant_nursery as same render as green_greenhouse_horticulture

geostonemarten:
fix issue https://github.com/gravitystorm/openstreetmap-carto/issues/1772

### DIFF
--- a/landcover.mss
+++ b/landcover.mss
@@ -278,7 +278,8 @@
 
   [feature = 'landuse_farm'],
   [feature = 'landuse_farmland'],
-  [feature = 'landuse_greenhouse_horticulture'] {
+  [feature = 'landuse_greenhouse_horticulture'],
+  [feature = 'landuse_plant_nursery']{
     [zoom >= 10] {
       polygon-fill: @farmland;
       [zoom >= 16] {


imagico:
This essentially means rendering it like farmland.

Have you considered the suggestion to render it as a pattern variation of orchard?

Technically you should have all changes in one PR and modify project.yaml and then generate project.mml.

geostonemarten:
@imagico I have create other path for these others files but continus integration reject its... I fork master project to create a full path.

> Have you considered the suggestion to render it as a pattern variation of orchard?

I don't see this suggestion but i think plant_nursery is most similar to green_greenhouse_horticulture than orchard

For this time there is no render. We can change it last time.

imagico:
You can push your yaml/mml changes to the geostonemarten:patch-1 branch and it will be automatically added to this PR.  You need to change the yaml and then call `yaml2mml.py` to generate the mml - otherwise the automatic checks will indicate an error.

> I don't see this suggestion but i think plant_nursery is most similar to green_greenhouse_horticulture than orchard

See https://github.com/gravitystorm/openstreetmap-carto/issues/1772#issuecomment-133448543

plant_nursery is often closer to orchard than to farmland on various levels so it might be a better way to render it.

PR #2308, opened 2016-08-25
add plant_nursery render

geostonemarten:
fix issue #1772


nebulon42:
I believe you are missing a piece to make this work. Adding the tag to project.yaml does not suffice, you also need a Carto rule for that. Have a look at: https://github.com/gravitystorm/openstreetmap-carto/blob/778cb4889f67fe611f0be76fe4e7888a654d188f/landcover.mss#L281

geostonemarten:
@nebulon42 I do that in a patch before this pull request but I delete it... I redo this modification. Thanks for this feedback

nebulon42:
I'm curious how you edit the style. Do you use some software to preview your changes like Tilemill or Kosmtik (https://github.com/kosmtik/kosmtik)? To make it easier for you to assess whether your changes work or not I suggest trying them out locally first. If you need help with that, everybody here is happy to answer your questions.

landcover.mss
```diff
@@ -278,7 +278,8 @@
 
   [feature = 'landuse_farm'],
   [feature = 'landuse_farmland'],
-  [feature = 'landuse_greenhouse_horticulture'] {
+  [feature = 'landuse_greenhouse_horticulture'],
+  [feature = 'plant_nursery'] {
```
nebulon42: I think your filter here should actually read `[feature = 'landuse_plant_nursery']`.

geostonemarten: @nebulon42 Fine, I edit it with notepad++ and use "github desktop" and this is a mistake... I revert and redo change... Sorry. So Kosmtik is a good idea to start with style. Have you a tuto to start with it?

Thanks

pnorman:
Could you provide an example screenshot of a plant nursery being rendered with this branch?

nebulon42:
@pnorman I think @geostonemarten has not yet setup the development environment completely. But we are trying to get there.
@geostonemarten Since you asked for a tutorial there is of course https://switch2osm.org/ but that does not explicitly cover developing styles. I have created some steps for a talk I held, but I have to translate them to English first. I'll try to do that on the weekend.

geostonemarten:
@pnorman  i have just set render as same as [greenhouse_horticulture](http://wiki.openstreetmap.org/wiki/Tag:landuse%3Dgreenhouse_horticulture) because actually this is relatively similar and the render is same as farmland.

> @geostonemarten has not yet setup the development environment completely. But we are trying to get there.

@nebulon42 Exactly ;-) I contribute with an other [username](https://www.openstreetmap.org/user/jseigneuret) on openstreetmap. But I am just starting to check and fix part of error on rendering.

nebulon42:
@geostonemarten Since you mentioned Notepad++ I infered that you are a Windows user. I do not use Windows myself so it is difficult to help with details. AFAIK @Ircama is another Windows user who might also be able to provide assistance.

Here is the tutorial I created: https://gist.github.com/nebulon42/e5f73e511d44726ef66c7e13eb9e3eca
It is derived from a presentation, so it is in bullet form and could be extended. It is not quite verbose on the database installation part and installation of other software but there should be plenty of tutorials for installing those on the Internet.

Ircama:
Hi.

An installation guide I prepared for Windows is here at the moment: https://github.com/Ircama/switch2osm.github.io/blob/tilemill-osm-carto/tilemill-osm-carto.md

Consider also you might benefit from a Python version of get-shapefiles, which you can find here at the moment: https://github.com/gravitystorm/openstreetmap-carto/pull/2137

I’m still trying to prepare a set of guidelines for Git in relation to this repository (and collecting some practices for style development that I learned so far).

In the consideration that the Windows software runs a very old Mapnik version (this is what the Windows Tilemill installation does), I would suggest to set-up a Linux system running Kosmtik. (The problem with Windows is that while Mapnik 2 and Node.js install, I could not find a way to install node mapnik, which is a prerequisite for Kosmtik.) Also a virtual machine or a cloud based system can work, anyway it is important some CPU for the rendering as well as a wide enough set of data to be loaded into the PostgreSQL instance in order to allow a decent testing.

As per GitHub Desktop (https://desktop.github.com/), I would recommend to always use git commands directly (selecting “Open in Git Shell”). This way GitHub Desktop would be adopted just as a monitor and not to issue commands (GitHub Desktop provides a too basic interface and does not clarify which git commands wraps).

I also would suggest you save your work, set your master even with origin/master, create a branch and do there your modifications and commits. Notice also that you have 7 commits and you might squash them by doing a `git rebase origin/master` to your branch, squashing your commits and subsequently `git push –force`.

pnorman:
Closing since the PR is inactive and needs some work once @geostonemarten gets his development environment set up. When that's done we can reopen or use a new PR.

PR #2310, opened 2016-08-27
Add an index for water polygons

### LINKED ISSUE #2309: Add index for water
Work in https://github.com/openstreetmap/operations/issues/104#issuecomment-242682177 seems to show it's a good idea.

A quick check building a partial index with the same where clause as the water query ended up with a 438MB index, about half the size of the current one on production

pnorman:
Fixes #2309, cross-ref openstreetmap/operations#104

### DIFF
--- a/indexes.yml
+++ b/indexes.yml
@@ -21,6 +21,12 @@ polygon:
     where: building IS NULL
   military:
     where: landuse = 'military'
+  water:
+    # The indentation here makes sense in the SQL output
+    where: |-
+      waterway IN ('dock', 'riverbank', 'canal')
+          OR landuse IN ('reservoir', 'basin')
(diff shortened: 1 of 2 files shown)


indexes.yml
```diff
@@ -21,6 +21,12 @@ polygon:
     where: building IS NULL
   military:
     where: landuse = 'military'
+  water:
+    # The indentation here makes sense in the SQL output
+    where: |-
+      waterway IN ('dock', 'riverbank', 'canal')
+          OR landuse IN ('reservoir', 'basin')
+          OR "natural" IN ('water', 'glacier')
```
tilmanb: Double quotes only on natural - is that intended?

pnorman: Yes, `natural` is a reserved word in SQL

PR #2313, opened 2016-08-29
variables for commonly used halo values

nebulon42:
Since we use the same values for text halos in a lot of places it made sense to use variables for it. Extends the use of variables beyond amenity-points and adds a variable for halo-fill.
Found using a simple search and replace. No rendering change.

### DIFF
--- a/addressing.mss
+++ b/addressing.mss
@@ -39,7 +39,7 @@
     text-size: 11;
     text-fill: #444;
     text-face-name: @book-fonts;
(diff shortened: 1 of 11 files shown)


pnorman:
Instead of `text-halo-radius: @halo-radius-extra` is it possible to do `text-halo-radius: @standard-halo-radius * 1.5`?

nebulon42:
Done.

PR #2318, opened 2016-08-30
remove citywall inconsistencies and new style for city walls

### LINKED ISSUE #1333: city wall inconsistencies
Line thickness varies between `barrier=city_wall` and `historic=citywalls` with the latter being on its own layer.

Example: http://www.openstreetmap.org/#map=19/49.61542/6.13741
citywall

The connecting line between the thicker line parts is tagged as `barrier=city_wall` while the other parts also have a `historic=citywalls` tag.

nebulon42:
Fixes #1333.

Removes the extra citywalls layer and moves `historic=citywalls` to line barrier logic. Unifies `barrier=city_wall` and `historic=citywalls` rendering. Because of the special meaning the rendering is a bit more intense than normal walls.

z14
z14

z15
z15

z16
z16

z17
z17

z18
z18


dieterdreist:
sent from a phone

> Il giorno 30 ago 2016, alle ore 23:47, Michael Glanznig <notifications@github.com> ha scritto:
> 
> remove citywall inconsistencies and new style for city walls (fixes #1333

please don't. The citywalls in the current rendering style are way more appropriate than the proposed smaller and lighter version. If you want to unify the barrier and historic linear citywall style, I'd rather use the bold version for both.

Have you checked citywalls mapped as areas as well?

mboeringa:
> please don't. The citywalls in the current rendering style are way more appropriate than the proposed smaller and lighter version. If you want to unify the barrier and historic linear citywall style, I'd rather use the bold version for both.

I don't agree _at all_... I far prefer the much more subtler version of the city wall rendering as suggested here by @nebulon42 .

As @matkoniecz also stated in #1333, the thick, partiallly transparent rendering, of the current styling, doesn't work well in quite a number of cases, as it overlaps and mixes with e.g. buildings. It more looks like some kind of boundary type at times.

I also think the suggested version is more in line with the general trend the latest changes to the style seem to be going, with a more subtle approach to displaying features.

I think the city walls in Rome will still show up great with this rendering (but not dominating everything). There's just one thing @nebulon42 might consider: put citywalls **on top** of highways with this new rendering, instead of below. This will ensure the thinner line is clearly visible (and even better than the current thick rendering in places like Rome), while unlikely to cover roads entirely, as the new rendering is thinner than most road types at relevant scales.

nebulon42:
>  Have you checked citywalls mapped as areas as well?

Yes, they are slightly more prominent now.

citywall_area

Here are some previews for Rome:

z15
rome_z15

z16
rome_z16

z17
rome_z17

z18
rome_z18

math1985:
To me the new rendering is a significant improvement.

mboeringa:
> To me the new rendering is a significant improvement.

Rome is also in a sense a "worst-case" example. It is one of the few cities that have roads running directly along the base of the city walls for a significant length. This doesn't happen often, and visibility in most cases should be much better. It was also with Rome in mind, that I suggested to consider putting city walls "on top" of roads, instead of rendering them below (but it will mean any "highway=footway" tagged on top of the wall, will become invisible - unless it would be possible to only render city walls on top of highway=footway with layer <= 0, which I think based on all I read here in the carto repo, is not feasible)

dieterdreist:
2016-08-31 20:59 GMT+02:00 Michael Glanznig <notifications@github.com>:

for Rome this modification is clearly making things worse because of the
importance of the walls (they separate the historical city from the one
built in the 20th and 21st century), but I admit that most citywalls have
not been as impressive as the Aurelian City Walls (your 1st and 2nd
screenshots also show the vatican citywalls, which are much more recent).
With this change they will be underrepresented (their actual width is
ranging from 2m to 8m and height is significant) particularly in zoom 17
and 18. I guess we will have to draw them as areas now (something I wanted
to do for a long time anyway, because of added detail).

some impressions to get an idea here:
https://en.wikipedia.org/wiki/Aurelian_Walls

dieterdreist:
2016-09-02 8:40 GMT+02:00 mboeringa <notifications@github.com>:

> Rome is also in a sense a "worst-case" example. It is one of the few
> cities that have roads running directly along the base of the city walls
> for a significant length. This doesn't happen often



this is a strange statement. Almost all cities do have roads either along
the citywalls or where they once were. It's an ever recurring scheme (but
there might be exceptions)


It was also with Rome in mind, that I suggested to consider putting city
> walls "on top" of roads, instead of rendering them below (but it will mean
> any "highway=footway" tagged on top of the wall, will become invisible



This is what I have done in my rendering as well. It will lead to some
strange situations but obviously improves citywall visibility significantly.

mboeringa:
Maybe a minor compromise could also be made to enhance visibility in the particular situations as displayed in Rome, by slightly widening city walls compared to the original proposal here. It seems it is now not even half the width of the current rendering, so it seems there may be some space to slightly widen it up, while still getting the improvement in cartographic styling.

nebulon42:
The problem I have here is that there is no way to differentiate between impressive and heavy fortified appearance and other city walls that are not so impressive. So I don't see any reason to have thicker lines just because the walls are particular pronounced in reality. Maps always are an abstraction. In my opinion you clearly see that there is a wall. The rendering before was just plain ugly and the thickness unjustified. I also don't like to change the layer ordering to have them above highways.

mboeringa:
> The rendering before was just plain ugly and the thickness unjustified. 

No doubt about that. I think your original proposal is generally OK and will work in most cases.

> I also don't like to change the layer ordering to have them above highways.

It was just a suggestion, there are drawbacks as well to that suggestion, as I already noted.

dieterdreist:
sent from a phone

> Il giorno 03 set 2016, alle ore 14:14, Michael Glanznig <notifications@github.com> ha scritto:
> 
> The problem I have here is that there is no way to differentiate between impressive and heavy fortified appearance and other city walls that are not so impressive.



any citywalls are very important, any city had the walls according to its importance 



> So I don't see any reason to have thicker lines just because the walls are particular pronounced in reality. Maps always are an abstraction.


yes, but displaying a feature less bold than it actually is would only be justified in case of features you're "ashamed" of, or in other words that you would like to hide


> In my opinion you clearly see that there is a wall. The rendering before was just plain ugly and the thickness unjustified.


yes, it was a tad too bold, I agree

pnorman:
Maybe something between the current PR and the existing rendering, closer to the PR?

Otherwise it's an improvement as-is.

On the technical side, removing the `citywalls` has been something I've been meaning to get around to for some time.

math1985:
Thanks, a significant improvement for Luxembourg.

mboeringa:
Hi @nebulon42 ,

Is this commit actually doing what it is supposed to do, or hasn't it been implemented on the OpenStreetMap site yet?

I have now tagged a number of sections of old city walls in the town of Elburg with **historic=citywalls**, but they don't show up (e.g. http://www.openstreetmap.org/way/449463607#map=19/52.44872/5.83539).

Also, I am wondering what the **"OR historic = 'city_wall"** is doing in this piece of the new code? The two most used taggings are **barrier=city_wall** and **historic=citywalls**, not **historic=city_wall**
```

` FROM (SELECT way,
+            ('barrier_' || (CASE WHEN barrier IN ('chain', 'city_wall', 'embankment', 'ditch', 'fence', 'guard_rail',
+                  'handrail', 'hedge', 'kerb', 'retaining_wall', 'wall') THEN barrier ELSE NULL END)) AS barrier,
+            ('historic_' || (CASE WHEN historic = 'citywalls' THEN historic ELSE NULL END)) AS historic
+            FROM planet_osm_line
+            WHERE barrier IN ('chain', 'city_wall', 'embankment', 'ditch', 'fence', 'guard_rail',
                   'handrail', 'hedge', 'kerb', 'retaining_wall', 'wall')
-            AND (waterway IS NULL OR waterway NOT IN ('river', 'canal', 'derelict_canal', 'stream', 'drain', 'ditch', 'wadi'))
+              **OR historic = 'city_wall**' `
```

nebulon42:
@mboeringa Yes, that's wrong. Confusion on my side. I haven't spotted this as the examples I have looked at are also tagged either `barrier=wall`or `barrier=city_wall`which includes them in the query result. PR on the way.

mboeringa:
Ok, great, it is good to correct. I think I already saw iD showing the false **historic=city_wall** as suggestion / preset. Probably because users already started to re-tag features to have them show up... (at least, if the wrong code lead to rendering of historic=city_wall).

PR #2319, opened 2016-08-31
One way fix

### LINKED ISSUE #2006: Oneway arrows prevent display of long street names
Long street names (by the looks of it, over 20 characters long) are not displayed on oneway streets. This is especially troublesome in bilingual areas where the name in both languages is used.

See for example http://www.openstreetmap.org/way/22575474:
clipboard image 13

StyXman:
Fixes #2006.


kocio-pl:
Could you merge these commits (like rebase or something)?

StyXman:
Reopened as #2320.

PR #2320, opened 2016-08-31
GIve one way arrows lower precedence than names

### LINKED ISSUE #2006: Oneway arrows prevent display of long street names

StyXman:
Move one way arrow rendering from several road fill related layers to `roads-text-name`, so arrows get lower precedence than names. As the arrow symbol is quite small, most of the time it can still be squeezed in.

Fix #2006.


StyXman:
So it gets referenced in the issue: Fixes #2006 .

roads.mss
```diff
@@ -2965,29 +2965,29 @@ tertiary is rendered from z10 and is not included in osm_planet_roads. */
...
+    [highway = 'road'],
+    [highway = 'service'],
+    [highway = 'pedestrian'],
+    [highway = 'raceway'],
+    [highway = 'cycleway'],
+    [highway = 'footway'],
+    [highway = 'path'],
+    [highway = 'steps'],
+    [highway = 'track'],
+    [highway = 'bridleway'] {
```
nebulon42: Is there a specific reason why you changed all these filters? The `feature = 'highway_*'` was put in deliberately to avoid combinatorial explosion.

StyXman: The current query for this layer does not create a `feature` like `roads-fill` and other do. Instead, it already creates a `highway` field. Should I port that too?

math1985: No, I don't think that's necessary (as long as you don't add definitions for arrows on railway's or so).

nebulon42: Ah, sorry, I misread the diff.

math1985:
Thanks! I added a more descriptive title.

pnorman:
Is this just waiting on a review?

rickmastfan67:
Pretty big regression has been discovered when it comes to motorway/motorway_link's w/ no name at all, or just a 'ref' tag.  They are now no longer getting ANY one-way arrows at any zoom. See #2340.

StyXman:
@rickmastfan67 I saw it, I'll continue the discussion there.

PR #2321, opened 2016-09-01
Change font size and minimum zoom level for house numbers

sommerluk:
minimal font size for adresses: 10

Unify rendering of addr:housename and addr:housenumber and have now always the same size at the same zoom level. Drop rendering at z17 (yet currently there are collisions, often no way_pixels available for making a better choise). Minimal font size of 10.

Contains a filter for z20. Now osm.org does not provide z20. Is it useful anyway or not?

### DIFF
--- a/addressing.mss
+++ b/addressing.mss
@@ -7,27 +7,30 @@
 }
 
 #housenumbers {
-  [zoom >= 17] {
+  [zoom >= 18] {
     text-name: "[addr:housenumber]";
     text-placement: interior;
     text-min-distance: 1;
     text-wrap-width: 0;
     text-face-name: @book-fonts;
     text-fill: #666;
-    text-size: 9;
+    text-size: 10;
+    [zoom >= 20] {
+        text-size: 11;
+    }
   }
 }
 
 #housenames {
-  [zoom >= 17] {
+  [zoom >= 18] {
     text-name: "[addr:housename]";
     text-placement: interior;
     text-wrap-width: 20;
     text-face-name: @book-fonts;
-    text-size: 8;
     text-fill: #666;
-    [zoom >= 18] {
-      text-size: 9;
+    text-size: 10;
+    [zoom >= 20] {
+        text-size: 11;
     }
   }
 }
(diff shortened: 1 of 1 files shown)


sommerluk:
# Names #

z18 before
name_z18_before

z18 after
name_z18_after

z19 before
name_z19_before

z19 after
name_z19_after

z20 after
name_z20_after

# With shops #

z18 before
shops_z18_before

z18 after
shops_z18_after

z19 before
shops_z19_before
 
z19 after
shops_z19_after

# High density #


z18 before
number_highdensity_z18_before

z18 after
number_highdensity_z18_after

z19  before
number_highdensity_z19_before

z19 after
number_highdensity_z19_after

z20 after
number_highdensity_z20_after

math1985:
Changed the title to reflect the fact that this PR also changes the minzoom level.

I think it would also be good to investigate the minzoom change in rural areas.

pnorman:
:-1: on restricting housename and housenumber to z18+

A medium density z17 area

image

image

Farmland with fairly frequent housing
image

Something else I noticed is that while browsing the map I spent a lot of time zooming into z17 to check the presence of addresses.

imagico:
You need to take into account scale depending on latitude and differences in actual address density here.

dense at low latitude:
http://www.openstreetmap.org/#map=17/1.28025/103.84382

dense at high latitude:
http://www.openstreetmap.org/#map=17/69.65401/18.95971

dense at very high latitude:
http://www.openstreetmap.org/#map=17/78.21956/15.65985

Circeus:
> dense at low latitude:
> http://www.openstreetmap.org/#map=17/1.28025/103.84382
> 
> dense at high latitude:
> http://www.openstreetmap.org/#map=17/69.65401/18.95971
> 
> dense at very high latitude:
> http://www.openstreetmap.org/#map=17/78.21956/15.65985

And in cities like Montreal and surrounding areas, you can have even higher densities (we're talking 3-5 distinct addresses by building here) with each address having four or even five digits.

sommerluk:
Now rendering also at z17, like before (but slightly bigger font size). Screenshots:

screenshot 1

screenshot 3

screenshot 4

screenshot 5

sommerluk:
I think this is ready for review.

PR #2325, opened 2016-09-02
Set minimum zoom for layer landcover_low_zoom.

StyXman:
I was doing some benchmarking back home when I noticed that ZLs 0-5 took a *long* time, even when they're mostly empty. My `pg` logs showed the `landcover_low_zoom` query as the culprit, but checking `landconver.mss`, we don't start rendering anything until ZL7 (`landuse = military` in `#landuser-overlay`). Hence, this PR.


math1985:
Thanks!

Could you address the whitespace issues in a different PR? That would make it easier to review.

StyXman:
Yeah, sure, I did this too quickly.

math1985:
Thanks! Performance improvements are very useful.

I have the impression that zoom level 8 is slow too, are you perhaps able to find out the cause for that layer too? Might be harder to solve though.

PR #2327, opened 2016-09-02
Set minimum zoom for layer landcover_low_zoom.

StyXman:
Try 3. See #2325 .

### DIFF
--- a/project.mml
+++ b/project.mml
@@ -113,7 +113,8 @@
         85.05112877980659
       ],
       "properties": {
-        "maxzoom": 9
+        "maxzoom": 9,
+        "minzoom": 7
       },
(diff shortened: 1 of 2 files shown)


pnorman:
It's worth fixing these, but it shouldn't impact performance since mapnik shouldn't fetch from a layer if no rules match it.

StyXman:
@pnorman In fact I discovered these missing `minzoom`s because I compiled `mapnik` with `ENABLE_STATS=yes` and saw lines like these:

    0% - 0 rendered for 10093 queried for  layer 'landcover-low-zoom' and style 'landcover-low-zoom-foreground'

Probably I should report that as a bug in `mapnik`.

PR #2333, opened 2016-09-08
Add a description for scripts included in OpenStreetMap Carto

Ircama:
**Add a description for scripts included in OpenStreetMap Carto**

The aim of this PR is to propose a documentation for the scripts that are available in the *scripts* directory.

It updates the already existing README.md with some currently available information.

Notice that *get-shapefiles.py* is already mentioned here.

### DIFF
--- a/scripts/README.md
+++ b/scripts/README.md
@@ -1,3 +1,171 @@
-These scripts are necessary and useful solely during making changes to the map style.
+The following scripts are provided to support coding. They are necessary and useful solely during making changes to the map style and are unnecessary for map rendering.
 
-There are unnecessary for map rendering.
+- **scripts/yaml2mml.py**
+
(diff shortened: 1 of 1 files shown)


kocio-pl:
Great! That will resolve https://github.com/gravitystorm/openstreetmap-carto/issues/2080.

pnorman:
We shouldn't duplicate the usage information in the scripts --help output, just provide information on what the script does. Also, get-shapefiles.py doesn't exist at this time.

Ircama:
Just as a note, the new README includes some additional information than the *--help* output (e.g., installation instructions wrote down while understanding the purpose of each script); also, technical manuals conventionally report usage information and command line options.

Documentation might help and ideally could improve the quality of the repo; anyway it is not strictly essential for the styles to work and, if not wanted, this PR can be closed and maybe the documentation can be moved to some other place.

Ircama:
Rebased

scripts/README.md
```diff
@@ -1,3 +1,171 @@
...
+    --reindex             Rebuild existing indexes
+  ```
+  
+  Typical usage:
+  
+  ```
+  $ scripts/index.py > index.sql
+  ```
+
+  A goal with the indexes is to have them general-purpose enough to not need frequent changing with stylesheet changes, but to be usable with man y versions, and potentially other styles.
```
kocio-pl: man y -> many

math1985:
@Ircama 
I would prefer to have information about the scripts in the script files themselves. Would you be interested in changing the PR in that way?

And as @pnorman says, we shouldn't duplicate the information from the --help command, if we have information on multiple places it becomes a hell to maintain.

Ircama:
@math1985

OK, in this case I would consider better to close this PR (referring the README) and possibly open some future ones (related to the scripts themselves).

PR #2335, opened 2016-09-11
Move shop labels to z18

math1985:
Nearly all shops are located in shopping streets or shopping malls,
close to other shops. Therefore, only a very small part of shops
is currently displayed with label on z17. In addition, these areas
are very crowded.

By only displaying shop labels on z18, z17 will appear less overcrowded.

The shop icons will still be displayed on z17.

### DIFF
--- a/amenity-points.mss
+++ b/amenity-points.mss
@@ -1938,6 +1938,7 @@
   [feature = 'shop_variety_store'],
   [feature = 'shop_wine'],
   [feature = 'shop_other']{
+    [way_pixels > 3000][zoom >= 17],
     [zoom >= 18] {
       text-name: "[name]";
       text-size: @standard-text-size;
--- a/amenity-points.mss
+++ b/amenity-points.mss
@@ -1938,7 +1938,7 @@
   [feature = 'shop_variety_store'],
   [feature = 'shop_wine'],
   [feature = 'shop_other']{
-    [zoom >= 17] {
+    [zoom >= 18] {
       text-name: "[name]";
       text-size: @standard-text-size;
       text-dy: 12;


pnorman:
I'm torn on this. It improves the results in some areas, but feels like we're dodging choices of what to stop rendering. I'm also concerned about suburban and rural locations and large shops.

HolgerJeromin:
If the shop is small this change prevents the label on a complete different building or on the street.

dieterdreist:
sent from a phone

> Il giorno 11 set 2016, alle ore 21:24, Matthijs Melissen <notifications@github.com> ha scritto:
> 
> Nearly all shops are located in shopping streets or shopping malls,
> close to other shops. Therefore, only a very small part of shops
> is currently displayed with label on z17. In addition, these areas
> are very crowded.
> 

but especially in those areas that are neither shopping streets nor shopping malls, the shops are often the most significant orientation point (in residential areas, besides monuments and public buildings). Often you would want a label already at z16, z18 is rather late

math1985:
> It improves the results in some areas, but feels like we're dodging choices of what to stop rendering.

The alternative would be to drop things like shoes/clothing shops entirely (or move them to higher zoom levels). We could re-think a wider what-to-render strategy later, but for now I think this is an improvement.

> I'm also concerned about suburban and rural locations and large shops.

> but especially in those areas that are neither shopping streets nor shopping malls, the shops are often the most significant orientation point 

In such cases, the icon on its own is enough to make the orientation point clear in my opinion.

kocio-pl:
I need some proofs to judge it. I'm cautious for now, since the same problem exists also for eating streets/areas, so this would be just tuning all the shops (2.3m of uses) for malls (36k of uses).

aceman444:
Maybe show label for shops as nodes only on z18 and shops as area (plygon) on z17 or sooner?

kocio-pl:
I like making this distinction very much, no matter what zoom levels we will choose.

math1985:
Large shops (over 3000 pixels) are now also displayed on z17.

math1985:
Large shops that are mapped as polygons (over 3000 pixels) are now also displayed on z17.

imagico:
I think this would be the first use of way_pixels of areas that are not rendered themselves in this style.

Meaning you could then push your favorite  shop to z17 by drawing a large area without this being visible in the map otherwise.

math1985:
Before and after:

screenshot_2016-09-19_14-48-23
screenshot_2016-09-19_14-47-10

The 'after' looks significantly better to me, so I still support merging this.

PR #2342, opened 2016-09-13
Loosened filter so oneway arrows can be rendered when the highway has no name.

### LINKED ISSUE #2340: Oneway arrows missing from roads without name
Looks like a recent change have made oneway arrows disappear from many roads.
From what I can tell they do not appear unless the road have a name.
Many roads do not have name and are oneway. First that comes to mind are on and off ramps.
Limited access roads like motorways and motorroads may not have names as there are normally no direct property access from them. 

It basically removed the arrows from all motorways and motorroads in Norway :)

StyXman:
The roundabout part is a little too loose, but the code in the select makes sure they're rendered in the right cases. This fixes #2340 .


math1985:
Thanks @StyXman!

PR #2343, opened 2016-09-13
Adding icon for shop=deli

kocio-pl:
One of the [most popular shops](http://taginfo.openstreetmap.org/tags/shop=deli) (7958 uses) with no icon and steady growth:
taghistory 1

Design from @jotam, however I also plan to test covered jar design.

Preview:
4yhnpmj_


kocio-pl:
Other versions:
2) deli-14-2

Jotam:
Mmh, the lobster doesn't work very well in miniature, does it? The problem is that I still can't think of a more suitable icon. The candy jar too much looks like sweets to me. Maybe a mustard jar would be good. Or anything along those lines...

kocio-pl:
OK, links to SVG files added. 3) look to me more or less like a mustard jar, but I think also about kind of badge over the piece of food. Let's see.

Jotam:
You might want to try this:
5) 
deli_14-12
 - [SVG](https://vdma.aldebaran.uberspace.de/osm/deli_14-12.svg)

kocio-pl:
5) preview:
oh0xh4g8

HolgerJeromin:
The pharmacy icon is similar to a jar. I like the lobster even if you can not see all legs

kocio-pl:
The lobster is recognizable for me and more elegant than jar, but some people said that it can be mistaken for a seafood. This isn't a showstopper for me, because it's hard to make any shape bound to just one idea (including icons we already use). Yet we're still testing different ideas and how do they look in reality.

kocio-pl:
I hoped the ribbon version would be right, but it doesn't work (I forgot how to chop the shape, so it's just faked in SVG) - here's my try:
6) deli_14-14 - [SVG](https://github.com/kocio-pl/openstreetmap-carto/blob/icons/symbols/deli_14-14.svg)

At this moment I prefer both 1) - lobster - and 4), and I find 5) to be acceptable.

Jotam:
One more try from my side. As the lobster is somewhat popular here, but maybe a little bit too delicate, I decided to cut away the lower part. It's now kind of a truncated lobster. :-) Let's give it a try, please.
7) deli_14-15 - [SVG](http://vdma.aldebaran.uberspace.de/osm/deli_14-15.svg)

kocio-pl:
I would chop it bit more =} (to make it more recognizable) :
8) deli_14-16 - [SVG](https://github.com/kocio-pl/openstreetmap-carto/blob/icons/symbols/deli_14-16.svg)

but legs are still not too visible and 1) has at least general shape, so I would still prefer it.

7) preview
ka2sys5r
8) preview
tdv_rfyb

Jotam:
I agree. The "chopped" version doesn't really work well. Nice try, though.

Jotam:
I still love the lobster idea. But unfortunately it doesn't look good. In fact the lobster is hardly recognisable. So as bad as it is, I opt for one of the jam/mustard jars.
Since I don't quite understand the "star" in 4) (I have never ever seen a jar with a star-shaped label), I vote for either 3) or 5).

kocio-pl:
So 5) looks like the common acceptable choice for both of us unless we find something better. If nobody complains, I will update PR with this shape.

Jotam:
Agreed. Go ahead! :+1:

boothym:
I agree with the jar - should be fine alongside a Pharmacy with the different colour and + symbol.

kocio-pl:
Done, now it should render like [here](https://github.com/gravitystorm/openstreetmap-carto/pull/2343#issuecomment-246805542).

Issue #2278, opened 2016-08-11
Why are there no labels for oceans and seas?


clhenrick:
Looking on Overpass Turbo I see these features exist as nodes:

**Oceans:**
![screen shot 2016-08-11 at 4 35 42 pm](https://cloud.githubusercontent.com/assets/161748/17608511/c5da1ba2-5fe1-11e6-90a1-1a727337df5b.png)

**Seas:**
![screen shot 2016-08-11 at 4 36 50 pm](https://cloud.githubusercontent.com/assets/161748/17608528/dd83aca0-5fe1-11e6-88c0-6c2b94a96d68.png)

It seems like these are features that should render at low to mid range zooms, and that implementing labels for them would be simple. Is there a reason this has not yet been done? Or would a PR help?

math1985:
Thank you for your suggestion! I think one problem is deciding the zoom level for these labels. For example, at the zoom level of your screenshot, rendering the seas in Japan does not make sense, but rendering oceans any later than this would be weird too. I think that problem might not be easy to avoid.

#804 and #788 might provide useful information.

clhenrick:
No problem, I don't see why rendering labels for seas at z4 or z5+ would be an issue, their labels would make good use of all the negative space that currently exists in the water area. 
Ocean labels would make sense to render at z3+ when country labels first start to render. These conventions seem to be followed by other web map providers as well for comparison. I don't believe collision with other labels would be an issue either, but could be proven wrong.

A simple test would be to add the following in the `.text` block on line 281 in `water.mss`:

```carto
  [feature = 'place_ocean'],
  [feature = 'place_sea'] {
    #text-point[zoom >= 5] {
      text-name: "[name]";
      text-halo-radius: 1;
      text-halo-fill: rgba(255,255,255,0.6);
      text-fill: @water-text;
      text-size: 12;
      text-face-name: @oblique-fonts;
    }
  }
```

and make sure that `place_sea` and `place_ocean` are being captured in the SQL for `text-point` layer on [line 2236](https://github.com/gravitystorm/openstreetmap-carto/blob/master/project.yaml#L2236) of project.yaml.

clhenrick:
Some samples of sea labels rendering in Asia:

z5
![image](https://cloud.githubusercontent.com/assets/161748/17633102/93eec210-6080-11e6-991b-049775f3d810.png)

z6
[...]

Issue #2068, opened 2016-03-03
Rendering of natural=bay vanishes too early and is too small


martiensch:
The labels of _areas_ which are tagged with natural=bay do not scale and vanish too early when zooming out.

[Example](http://www.openstreetmap.org/relation/3123125#map=14/53.2912/7.1783)
The label is omitted in Z13, but the area is vast and includes also the wetlands surrounding the water. Also the label is quite small compared to the area size.

I would like to see that the label for natural=bay is rendered similarly to natural=wetland which is visible up to Z10 ([example](http://www.openstreetmap.org/relation/1592352#map=10/53.4345/6.5273)).

imagico:
See also #788, #804.

Currently less than 5 percent of bays are mapped as areas, the merits of mapping bays as areas universally are subject to debate (see also http://wiki.openstreetmap.org/wiki/Tag:natural%3Dbay) and since the coastline is not in the database there is no way to properly assess the size of bays mapped as nodes.

Having size dependent rendering for bays mapped as areas but not for those mapped as nodes would add a huge incentive to mappers to map them as areas which will likely result in a lot of non-verifiable and difficult to maintain area drawing.  Ideally mappers should feel free to map as either area or node depending on the situation and not get a particular pressure to choose an area due to the choices in rendering here.

mboeringa:
This is from my ArcGIS Renderer for OpenStreetMap. It is a 1:100k rendering of a Croatian island which includes **natural=bay** features, based on simple nodes primarily in this area, because that is how it is tagged here. 1:100k in this local Croatian / Balkan projection is about equivalent to Z12. I actually render the bay features up to 1:250k scale in reality, which would equate to about Z11.

I think it quite acceptable, although it does require a proper knock-out labelling that removes overlapping labels. In fact, there are a quite bit more **natural=bay** features in this area that had their label being removed in the labelling process. Of course, the removal is _completely arbitrary_ based on a rendering of nodes, so you won't have a hierarchy of large and small bays in this image, although it is likely that large bays won't have any other bay label features near, so are a bit more likely to "naturally" stay in view instead of getting removed.

Anyway, Z10 may be a bit to much for a rendering not able to distinguish between large and small features, but Z12 just encompasses the Dollart ;)

![arcgis_renderer_for_openstreetmap_natural-bay](https://cloud.githubusercontent.com/assets/7635726/13496320/ca70bd6c-e14f-11e5-93f1-3b3aa82f05f5.png)

[...]

Issue #804, opened 2014-08-01
Add rendering for natural=strait


matthijsmelissen:
Straits are important landforms, some of them are very well known, e.g. Strait of Gibraltar or Strait of Magellan. Rendering them would be great. Right now, natural=strait is widely used, but often combined with place=locality, in order to force the rendering. place=locality could be avoided by rendering natural=strait directly.

See also https://trac.openstreetmap.org/ticket/4190.

anderfo:
What happened to this? Looks like it's still not rendered even at high zoom - being tagged on node, polygon or multipolygon...but on the "Humanitarian" layer it is rendered.

BTW I asked on the wiki/talk a few months ago about why a strait cannot be tagged as a line, while a valley can....no response yet: https://wiki.openstreetmap.org/wiki/Talk:Tag:natural%3Dstrait
The geometry of a strait is rather similar to a valley.
I'd still prefer polygons (or even multipolygons, although not permitted by the wiki), for the reasons mentioned by gravitystorm above.

imagico:
> The geometry of a strait is rather similar to a valley.

Actually no, a strait is a narrow between land masses, the best land analogy would be a saddle.  Valley like straits are usually either submerged glacial troughs (i.e. fjords open at both ends) - like the Strait of Magellan or submerged river shaped valleys - like the Bosporus.  The vast majority of straits are neither, they are usually just passages between islands or islands and continents.

Most of the big straits around the world are either mapped as nodes or as ways with 2-4 nodes where the location of the start and end node usually have no particular meaning, they essentially only specify direction:

http://www.openstreetmap.org/way/146655069
http://www.openstreetmap.org/way/146655079
http://www.openstreetmap.org/way/146655070

The problem here is not mapping, the problem is rendering in a way that does not give wrong incentives to mapping.  If you look for example here:

http://overpass-turbo.eu/s/dGz

If you'd use the way length (or in case of area mapping the area size) for importance rating mappers would enlarge or shrink the features used for mapping to get the desired outcome - without any basis in reality.  Actual importance of straits is best determined from their width and the size of islands or continents they separate.

dieterdreist:
2016-01-12 12:17 GMT+01:00 Christoph Hormann <notifications@github.com>:

> Actual importance of straits is best determined from their width and the
> size of islands or continents they separate.




actual importance is connected to the availability of alternatives but much
more to the importance for trade / amount of traffic, e.g. the Bosporus is
[...]

PR #2345, opened 2016-09-14
Label oceans and large seas from a static file

pnorman:
This is a proof of concept of using a static file to label seas and large oceans with hand-placed labels.
- The placements are designed for a webmap which wraps around +/- 180. This allows placement of a single North Pacific label, while many printed maps use two, one on each edge. Other webmaps use a single label but place it so it doesn't cross 180. I believe what I am doing is unique among webmaps. Not worrying about the +/- 180 allows for more flexibility in positions
- Positions and sea selections are designed for Mercator. osm-carto will never make a good map for viewing the entire world at once with its projection, but it's what we use.

There are a number of outstanding issues
- [ ] Even with the increased size, the smallest labels seem faint. We might need to adjust the water colour. Many maps use white.
- [ ] With the syntax used it seems necessary to specify the same point multiple times even when it's at the same location in order to increase its size on higher zooms.
- [ ] Labels on z4+ haven't been given huge consideration
- [x] I'm not sure what to do with the southern ocean. It's geometry is unlike other oceans because the on-globe centroid is on land, and in Mercator it is a very large ocean.
- [ ] ~~Some of the seas should probably be in the local language~~
- [x] Right now line breaks are done through wrapping. Since these are manual labels it might make more sense to turn off wrapping and use manual newlines

I drew technical ideas from [osm2vectortiles seas.geojson](https://github.com/osm2vectortiles/osm2vectortiles/pull/394#issuecomment-238073484) and cartographic/labeling ideas from Natural Earth and [Tim Patterson's Physical Map of the World](http://shadedrelief.com/world/). I specifically avoided looking at MapBox Streets as they have falsely [claimed ownership over similar files to what I was making](https://github.com/osm2vectortiles/osm2vectortiles/issues/387#issuecomment-238060253).

### DIFF
--- a/project.yaml
+++ b/project.yaml
@@ -27,7 +27,7 @@ _parts:
   extents84: &extents84
     extent: *world
     srs-name: "WGS84"
-    srs: "+proj=longlat +ellps=WGS84 +datum=WGS84 +no_defs"
+    srs: "+proj=longlat +ellps=WGS84 +datum=WGS84 +no_defs +over"
   osm2pgsql: &osm2pgsql
     type: "postgis"
     dbname: "gis"
@@ -100,6 +100,18 @@ Layer:
     properties:
       minzoom: 1
       maxzoom: 3
+    advanced: {}
+  - id: "ocean-text"
+    name: "ocean-text"
+    class: ""
+    geometry: "linestring"
+    <<: *extents84
+    Datasource:
+      file: "static/ocean.geojson"
+      type: "geojson"
+    properties:
+      minzoom: 1
+      maxzoom: 8
     advanced: {}
   - id: "landcover-low-zoom"
     name: "landcover-low-zoom"
--- a/project.mml
+++ b/project.mml
@@ -74,7 +74,7 @@
       "geometry": "linestring",
       "class": "",
       "id": "necountries",
-      "srs": "+proj=longlat +ellps=WGS84 +datum=WGS84 +no_defs",
+      "srs": "+proj=longlat +ellps=WGS84 +datum=WGS84 +no_defs +over",
       "Datasource": {
         "type": "shape",
         "file": "data/ne_110m_admin_0_boundary_lines_land/ne_110m_admin_0_boundary_lines_land.shp"
@@ -87,6 +87,29 @@
       ],
       "properties": {
         "maxzoom": 3,
+        "minzoom": 1
+      },
+      "advanced": {}
+    },
+    {
+      "name": "ocean-text",
+      "srs-name": "WGS84",
+      "geometry": "linestring",
+      "class": "",
+      "id": "ocean-text",
+      "srs": "+proj=longlat +ellps=WGS84 +datum=WGS84 +no_defs +over",
+      "Datasource": {
+        "type": "geojson",
+        "file": "static/ocean.geojson"
+      },
+      "extent": [
+        -180,
+        -85.05112877980659,
+        180,
+        85.05112877980659
+      ],
+      "properties": {
+        "maxzoom": 8,
         "minzoom": 1
       },
       "advanced": {}
@@ -2043,7 +2066,7 @@
       "key_field": ""
     },
     "extents84": {
-      "srs": "+proj=longlat +ellps=WGS84 +datum=WGS84 +no_defs",
+      "srs": "+proj=longlat +ellps=WGS84 +datum=WGS84 +no_defs +over",
       "extent": [
         -180,
         -85.05112877980659,
--- a/static/README.md
+++ b/static/README.md
@@ -1,0 +1,11 @@
+# Static data
+
+This directory is for static data that is part of OpenStreetMap Carto
+
+## sea.geojson
+
+Hand-placed labels for oceans and major seas. Some data is derived from OpenStreetMap `place=sea` data and other is hand-placed.
+
+Features crossing 180 degrees need to be repeated for each the west and east side of the map.
+
+Data is licensed under the [ODbL](http://opendatacommons.org/licenses/odbl/). Copyright Paul Norman and [OpenStreetMap contributors](http://www.openstreetmap.org/copyright).
--- a/static/ocean.geojson
+++ b/static/ocean.geojson
@@ -1,0 +1,151 @@
+{
+  "type": "FeatureCollection",
+  "features": [
+    {
+      "geometry": {
+        "coordinates": [-150, 0],
+        "type": "Point"
+      },
+      "properties": {
+        "maxzoom": 3,
+        "minzoom": 1,
+        "name": "Pacific\nOcean",
+        "size": "bigger"
+      },
+      "type": "Feature"
+    },
+    {
+      "geometry": {
+        "coordinates": [-40, 25],
+        "type": "Point"
+      },
+      "properties": {
+        "maxzoom": 3,
+        "minzoom": 1,
+        "name": "Atlantic\nOcean"
+      },
+      "type": "Feature"
+    },
+    {
+      "geometry": {
+        "coordinates": [75, -20],
+        "type": "Point"
+      },
+      "properties": {
+        "minzoom": 1,
+        "maxzoom": 4,
+        "name": "Indian\nOcean"
+      },
+      "type": "Feature"
+    },
+    {
+      "geometry": {
+        "coordinates": [-180, 80],
+        "type": "Point"
+      },
+      "properties": {
+        "minzoom": 1,
+        "maxzoom": 2,
+        "name": "Arctic Ocean"
+      },
+      "type": "Feature"
+    },
+    {
+      "geometry": {
+        "coordinates": [180, 80],
+        "type": "Point"
+      },
+      "properties": {
+        "minzoom": 1,
+        "maxzoom": 2,
+        "name": "Arctic Ocean"
+      },
+      "type": "Feature"
+    },
+    {
+      "geometry": {
+        "coordinates": [
+          [160, 77],
+          [190, 75],
+          [220, 77]
+        ],
+        "type": "LineString"
+      },
+      "properties": {
+        "minzoom": 2,
+        "maxzoom": 3,
+        "name": "Arctic Ocean",
+        "size": "big"
+      },
+      "type": "Feature"
+    },
+    {
+      "geometry": {
+        "coordinates": [
+          [-200, 77],
+          [-170, 75],
+          [-140, 77]
+        ],
+        "type": "LineString"
+      },
+      "properties": {
+        "minzoom": 2,
+        "maxzoom": 3,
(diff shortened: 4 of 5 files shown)


project.yaml
```diff
@@ -27,7 +27,7 @@ _parts:
   extents84: &extents84
     extent: *world
     srs-name: "WGS84"
-    srs: "+proj=longlat +ellps=WGS84 +datum=WGS84 +no_defs"
+    srs: "+proj=longlat +ellps=WGS84 +datum=WGS84 +no_defs +over"
```
pnorman: Needed for labels crossing +/- 180

project.yaml
```diff
@@ -101,6 +101,18 @@ Layer:
...
+    class: ""
+    geometry: "linestring"
+    <<: *extents84
+    Datasource:
+      file: "static/sea.geojson"
+      type: "geojson"
+    properties:
+      minzoom: 1
+      maxzoom: 8
+    advanced: {}
```
pnorman: Should be moved later on

imagico:
I do not like this for several reasons:

* not basing the names shown on the place nodes in OSM seems inappropriate for a community project.  I think a script that generates the geojson from unnamed geometries and the place nodes in OSM (easy to do using an overpass query) would be much better.
* the selection of seas is highly subjective.  You have the Kara Sea but not the Laptev, East Siberian, Chukchi and Beaufort Seas.  You have two marginal seas in the Indian Ocean but none in the Atlantic (like the Caribbean Sea) or Pacific (like Philippine Sea, Bering Sea)
* your Kara Sea node is very close to Vize Island.
* the division of the Atlantic and Pacific into Northern and Southern part is coming from the IHO limits of oceans and seas, i.e. is a politically motivated artificial division not based in any previous practical use of language - people living at the coast of these call it the Atlantic or Pacific, not the southern/northern.  So this labeling is questionable.
* I think it is probably feasible to create a reasonable automatic label placement based on the place nodes and the simplified coastline that does not lead to mapping for the renderer and still produces visually pleasing results.  While for just about a dozen sea labels this probably seems unnecessary work it would be very useful for labeling smaller maritime waterbodies, i.e. straits and bays - see also discussion in #804, #2068.

Regarding the Southern Ocean - if you want a single label the best place is probably south of Bouvet Island - in common use of the term the northern limit of the Southern Ocean is the Antarctic Convergence and this extents furthest north here:

https://commons.wikimedia.org/wiki/File:Antarctic-Convergence-Map.TIF

Again the commonly used limit at 60 degrees latitude is based on the IHO definition and politically motivated.

If you want multiple labels you could well place a second one at about 150 degrees west and possibly a third one south of Heard Island.

pnorman:
> not basing the names shown on the place nodes in OSM seems inappropriate for a community project. I think a script that generates the geojson from unnamed geometries and the place nodes in OSM (easy to do using an overpass query) would be much better.

We want to avoid the use of place nodes as "labelling positions". See https://github.com/gravitystorm/openstreetmap-carto/issues/2278#issuecomment-239862163 for Andy's view, mine is that label placement of oceans is purely a cartography decision and belongs in the style, not a data one which belongs in OSM.

Going with a static file was a design choice I consciously made

> [concerns about seas]

I've only got seas on z4+, and I didn't do much work on them, just enough for the proof of concept to test multiple zoom levels.

> people living at the coast of these call it the Atlantic or Pacific, not the southern/northern

Yes, that's the main but not exclusive usage. I was thinking even without southern/northern the pacific would need two labels, but without the need to break at the date line that paper maps have, I see some alternatives. The Atlantic will probably need two.

> If you want multiple labels you could well place a second one at about 150 degrees west and possibly a third one south of Heard Island.

For spacing I was going for very approximately an average of one label visible on average for a screen of map.

-----

In many ways it would be easier to only do oceans from a static file, then all seas could be done from OSM data. I'm not convinced that it's best.

imagico:
> We want to avoid the use of place nodes as "labelling positions". See #2278 (comment) for Andy's view, mine is that label placement of oceans is purely a cartography decision and belongs in the style, not a data one which belongs in OSM.

Note i said "basing the names shown on the place nodes", not using the place nodes.  It should be possible to match the nodes with manually placed coordinates and only use the name tag.

This is especially for the language matter - which language a certain sea is labeled in should be decided by the community in an open process, not by the maintainers here.  Since the place nodes are in the database it would be natural to use these to implement this process.

pnorman:
* Seas removed from the scope to keep this PR smaller and defer consideration of seas

* Tried a single label for the pacific on a line. Wasn't happy with it. 

  image

Preview of z1 and z2

image

image

imagico:
I will probably get in trouble for this but an argument could be made for labeling the Arctic Ocean in Russian - after all the Russian coast towards it is significantly longer than the US and Canadian combined...

Not sure if using curved labels for the Arctic Ocean is necessary - it is kind of odd with all other labels being point based.

I would probably move the Southern Ocean label south of the Atlantic somewhat to the east - it now extents over the South Sandwich Islands and there is sufficient room to the east to avoid that at least at z2.

dieterdreist:
sent from a phone

> Il giorno 15 set 2016, alle ore 13:21, Christoph Hormann <notifications@github.com> ha scritto:
> 
> I will probably get in trouble for this but an argument could be made for labeling the Arctic Ocean in Russian - after all the Russian coast towards it is significantly longer than the US and Canadian combined...


I support this

jojo4u:
So english is the default label language for osm-carto? According to OSM principles the ocean labels should probably be without name=* but only with name:[lang]=*

matkoniecz:
> not basing the names shown on the place nodes in OSM seems inappropriate for a community project. I think a script that generates the geojson from unnamed geometries and the place nodes in OSM (easy to do using an overpass query) would be much better.

Big problem here is that low zoom levels are rerendered only on style update. So fixing deliberate/accidental problems would require style update or manual cache invalidation.

Also, it is a bit tricky to decide language for labels of Arctic, Southern and Indian oceans. English would be probably the most readable choice for biggest share of our users but so far problem of default language was shifted to editors of name tah/

imagico:
> Big problem here is that low zoom levels are rerendered only on style update. So fixing deliberate/accidental problems would require style update or manual cache invalidation.

Actually the low zooms are regularly re-rendered once per month without style updates and i don't think it would be a performance issue to increase that to weekly or even daily update for z=0-4.

> Also, it is a bit tricky to decide language for labels of Arctic, Southern and Indian oceans. English would be probably the most readable choice for biggest share of our users but so far problem of default language was shifted to editors of name tah/

I think using English as a convenience default is a slippery slope.  With the same argument we could essentially label the whole map in English language.  But we deliberately chose to use local languages as selected by mappers individually and it is difficult to find a convincing argument why this should be different for the oceans.

As far as tagging is concerned - yes, there is potential for edit warring to influence the labels but not more than there is for other prominently labeled features i think.  The primary problem is that no matter how you'd extend the _on-the-ground rule_ to oceans - be it in form of the most frequently used name on the open ocean or the most frequently used name among the inhabitants of the coast - for nearly none of the world oceans there will be an absolute majority for any single language (the most likely candidate would probably be the Arctic Ocean for Russian, especially with the coast inhabitants definition).

To avoid problems with name change and update cycle interference it would be possible to only render changed names for the oceans in the map if the name change has proven to be stable.  For that you'd maintain a history of previous names and change the actually rendered name only when it has been the constant name for said feature for some time.

dieterdreist:
sent from a phone

> Il giorno 27 ott 2016, alle ore 18:02, Christoph Hormann <notifications@github.com> ha scritto:
> 
> But we deliberately chose to use local languages as selected by mappers individually and it is difficult to find a convincing argument why this should be different for the oceans.


I don't find this difficult: we chose local language to name the places by default how the people that live there call them. As nobody (almost) lives on the oceans the situation is significantly different.

imagico:
So you say we should label all parts of earth that are not or very sparsely populated (like < 1 person/square kilometer) in English?

mboeringa:
> So you say we should label all parts of earth that are not or very sparsely populated (like < 1 person/square kilometer) in English?

I think @dieterdreist only refers to possibly disputed areas outside territorial waters of countries. I think no one with any experience on OSM would start to re-label names in an another country to its own local language (even forgetting the local name policy for the "name" tag that has been OSM standard for years).

matkoniecz:
> Actually the low zooms are regularly re-rendered once per month without style updates

Thanks, I was unaware about that - especially if changed to daily update it reduces problem of poor names. Still, it give big potential for things going wrong.

dieterdreist:
I did not explicitly say we should use English for labeling of international waters, but there might be arguments to do so (also Russian and Chinese might be defendable alternatives, and maybe even fillipine languages if you go by actually spoken languages on sea, but I couldn't find relevant statistics on the fly).

matkoniecz:
> So you say we should label all parts of earth that are not or very sparsely populated (like < 1 person/square kilometer) in English?

Well, the difference is that oceans are not populated at all (or density is orders of magnitudes lower than < 1 person/square kilometer).

Obviously, situation changes if we count population on coasts/islands as population of ocean.

imagico:
I think no one would deny we already have a de facto English default for the name tag in many cases where there is no obvious most common language for the feature in question.  But this is just a de facto situation the OSM community can change any time if it feels like it.  And i don't think this style has any kind of mandate to cast this de facto situation as it is right now into static code.

Another point to consider:  There are many derivatives of this style implementing maps with localized labels like the German and French styles.  Currently this requires only modifying references to the name tag from within SQL in some way.  If we add a hardcoded list of features with name attributes in English this list would need to be modified as well complicating things quite a bit.

kocio-pl:
I also think we should stay with a "name" tag - this is default, generic key and its value can be changed any time the community feels it should be.

Ircama:
Apart from the appraisable discussion on labels, @pnorman’s proposal is quite interesting and I’d love to see the names of the oceans represented with the standard OSM style.

Hopefully, the following basic design might be of some help.

Maybe the LUA script planned in the LUA branch could be also used to separate the preprocessing of specific tags (e.g., preset in the configuration file) like the oceans.
 
To load these tags to the DB, their names and other relevant attributes could be read from the OSM input file, while their geometry might be taken from a different table set, to be separately maintained.
 
For features like oceans that are managed as nodes in OSM, the LUA process could also be used to change their type from node to way, so that a standard rendering can be designed within openstreetmap-carto for these lines.

References not matching the separate tables can be processed normally (as nodes), in order to avoid the issue on subjective selection of seas mentioned by @imagico.

imagico:
The more practical way probably is to have a script get the nodes via overpass API, merge their name tags into a geometry data template and write that to a shapefile/geojson to be used by this style.  This could be run daily, possibly with the suggested name change dampening.

pnorman:
Closing, no ETA on when I'll be able to look at this and it's not a current priority for me.

Issue #2265, opened 2016-08-04
Asian scripts need a bigger font size than latin scripts


sommerluk:
**The background:**

Asian scripts need a bigger font size than latin scripts.

Asian scripts have ofen more complex glyphs than latin scripts. These glyphs have many details and many strokes. They need therefore a bigger font size to be as legible as lati scripts.

That means: A font that supports both – latin scripts and asian scripts – might be well legible at font size 10 for latin scripts. But it might need font size 15 or 20 for asian scripts to be as legible as latin script at font size 10.

**The problem:**

Currently, most text in this style is ≥ font size 10. This is okay for latin text, but too small for asian text.

**No solutions?**

- Use a notably bigger font size overall in this style? Might work for Asia. Text labels will get higher because of the bigger font size. Horizontially they will not grow up to much because in asian scripts the names have often less characters than in latin scripts. But it would break the rendering in the rest of the world, because with a much bigger font size, much less labels can be rendered.
- Render asian characters with a different font size than latin characters? Would lead to inconsistent rendering. Would look terrible for labels that have both, latin and asian characters. Not a feature of CartoCSS.

gravitystorm:
We already use multiple fonts, so it's possible to use different fonts for asian and latin characters. We also have the same situation with lots of other non-asian characters, like arabic, hebrew, berber etc.

Is it possible to find fonts with some sort of harmony in their sizes? Or are there rules for how large the glyphs are for "10pt" fonts that mean the ratio of sizes between a latin 'x' and a chinese character '贵' is always fixed, for a given font?

If necessary, we can always try asking in the mapnik project to support complex font specifications like "DejaVu Sans 10pt, Something Else 12pt". Or would making the decision based on Unicode code point ranges be a better idea than the font names?

sommerluk:
The harmony between glyphs and fonts has various aspects. One of them is the form of the glyphs. Consistent glyph design is important for an overall harmony. A consistent type colour (grey level) is another aspect. And indeed the glyph size is another one: Usually CJK character’s size and position corresponds to the size and position of Latin capital characters of the same font, but slightly bigger.

[...]

Issue #2348, opened 2016-09-15
Use pull request reviews


pnorman:
A new github feature: https://help.github.com/articles/about-pull-request-reviews/ https://help.github.com/articles/reviewing-proposed-changes-in-a-pull-request/

What does this mean for our review process? Do we need to change anything?

kocio-pl:
I feel this is a nice tool and we're happily using it now, so probably this introduction was enough. Thanks!

Issue #2208, opened 2016-07-02
What about Han unification?


sommerluk:
The problem of Han unification is independent from any specific font!

I do not speak any of these languages. But as far as I understand, the problem is this one (corrections are welcome):

– Unicode encodes abstract characters (“meanings of signs”). It does not encode glyphs (“specific graphical representations of an abstract character”).

– A wide varity of abstract characters is shared between CJK scripts.

– There are glyphs that have the same appearance in all CJK scripts. There are other glyphs that are different in all CJK scripts.

– Nevertheless, native language speakers expect to see the glyphs they are used to see (language-specific glyphs). Different to the Unicode consortium, they consider that the different glyph form makes also a difference in the meaning of the sign. They feel that the other glyph forms are a foreign language.

– It is not possible with simple Unicode to distinguish these forms. The only possible way is IVD at http://www.unicode.org/reports/tr37/ but I’m not sure if this is widely used in the real-world values in the OSM database by our contributers in these regions.

– Good CJK fonts provide all these glyphs for all language variants.

– The problem is making the choise between all the available glyph forms.

– Web pages solve the problem by using HTML language tags. So the rendering engine can choose the appropriate glyph because it knows the target language.

– openstreetmap-carto has no knowledge about the target language (in the “name” key) and no region-specific rendering rules.

– openstreetmap-carto tries to display text in the native language.

Question: How to render CJK names and other text in the native language? If we default to chinese glyph forms, than also japanese city names will be rendered with chinese glyph forms, and japanese people will feel like it is a chinese map. If we default to japanese glyph forms, than also korean city names will be rendered with japanese glyph forms, and korean people will feel like it is a japanese map…

clkao:
Is it possible to add some sort of regional specific selector (based on country or bbox) as cartocss extension?  I think that's the basic facility required for using different glyph forms for their respective regions when preferred language is unspecified.

The default rendering should be using the region-specific glyph forms, assuming `name` is in that language. And when generating Chinese-version tiles (preferring `name:zh`) there should be overriding styles using the Chinese glyph forms.

mxa:
[...]

PR #2349, opened 2016-09-15
Initial Noto support

### LINKED ISSUE #294: Japanese Fonts Too Small / Halo (Stroke) too small / Shields misrendered
A lot of the labels in Japanese are using a light stroke font instead of a bolded font when the English is rendered bold. 

Train stations, amusement parks, motorways, and other important places have hard names to read in non-roman characters because they are rendered smaller than they should be for the kanji.

Examples:

Train Stations http://www.openstreetmap.org/#map=17/36.41395/139.33339
The Kanji for Nishikiryu is difficult to read.

Parks http://www.openstreetmap.org/#map=18/35.73297/139.74664
(The Rikugi-en Garden's complicated kanji is unreadable, for example) 

Tokyo Disneyland http://www.openstreetmap.org/#map=15/35.6293/139.8818
Tokyo Disneyland's label (東京ディズニーランド) is impossible to see, let alone read. (thin stroke, no contrast, tiny size) The labels on the parking isles? totally visible.  (A whole 'nother issue issue of not rendering labels on parking isles at lower zoom levels)  

Motorway labels http://www.openstreetmap.org/#map=15/36.3551/139.1981

The shield labels for the motorways are rendered completely wrong, overlapping the kanji.  The Kanji font is very feint.  Contrast this to the trunk label (17) right above it. 
Also the exit labels are not very readable either.  Usually you need these labels when going to a new place, so it is difficult to guess complicated kanji or read faint ones

 - - - 

Most of these issues can be remedied by the choice of a better font that includes bolded characters for Japanese / Chinese. I assume you are using the same font for these  features in all regions, but that font doesn't seem to have bold weight 6 characters (W6).  The font choice is also a serif font (you can see the strokes in the characters) for some features, when that is unneeded and makes things very hard to read at the small level they are rendered, whereas a san-serif font (no strokes) which is used in other places, is perfect for this Job. 

TL;DR : Japanese characters are faint / garbled / rendered incorrectly in some situations. Update the font to a san-serif with bold characters available, or at lease use only san-serif fonts for non-roaman languages, or increase the font size.

### LINKED ISSUE #362: Khmer bold fonts
https://github.com/gravitystorm/openstreetmap-carto/pull/358#issuecomment-36010976 recommended the Khmer OS Kangrey for bold Khmer text, but I deferred because it wasn't packaged on Ubuntu, there are much bigger issues with Khmer text until we move to a version of Mapnik with Harfbuzz support (https://github.com/gravitystorm/openstreetmap-carto/pull/358#issuecomment-36101027) , and there is currently one Khmer label on the map, which doesn't use bold.

Once other font work is done, I want to revisit this, as I should have a better grasp on if other non-packaged fonts would be an improvement.

### LINKED ISSUE #486: Chinese font support
Continuing the work in #414, I'm looking at CJK fonts. Because Han unification makes this more complicated than other fonts, I'm looking at Chinese first.

The following unicode fonts appear to be packaged on ubuntu 12.04, and serves as a preliminary list

* fonts-arphic-ukai - "AR PL UKai" Chinese Unicode TrueType font collection Kaiti style
* fonts-arphic-uming - "AR PL UMing" Chinese Unicode TrueType font collection Mingti style
* ttf-wqy-microhei - A droid derived Sans-Seri style CJK font
* ttf-wqy-zenhei - "WenQuanYi Zen Hei" A Hei-Ti Style (sans-serif) Chinese font

### LINKED ISSUE #677: Supporting Thaana labels
The language of the Maldives, Divehi, consistently is written in a right-to-left script called Thaana. You can see examples and a link to fonts on http://www.omniglot.com/writing/thaana.htm

On OpenStreetMap the correct symbols are displayed in the correct order, but the font is too small. For example, this island is named ފަރުކޮޅުފުށި (image attached)
failthaana
http://www.openstreetmap.org/#map=16/4.2310/73.5452

I asked in Mapnik and they sent me here (https://github.com/mapnik/mapnik/issues/2279)

### LINKED ISSUE #1067: Consider Noto fonts
I just found the free font collection Noto from Google. http://www.google.com/get/noto/
Having no real insight into osm carto font support but there might be some less used fonts in noto which could be interesting for us.

### LINKED ISSUE #1335: Rendering name with very close characters, hard to read
Hello,

The characters of natural=water are very close so that you can not differentiate between "rn" or "m".
example Kernersee

see live at http://osm.org/go/0MEnwKXYl

This seems to affect also other objects like streets, see
http://www.openstreetmap.org/#map=18/51.54463/6.84936

This is of course also depending on the zoom level. Hopefully this can be improved.

Thanks in advance!

### LINKED ISSUE #2204: Hangeul (Korean) font not readable
The labels in Korea written in Hangeul are hard to read or not readable at all. I did a quick mock-up in GIMP to see how other typefaces with a free license compare to the currently used on in carto.

This is the current situation:
fontosm-original

This is with the font "Sans Serif"
fontosm-sansserif

The Nanum font family is a font by the NAVER corporation published under a free license https://en.wikipedia.org/wiki/Nanum_font
Font "Nanum Gothic"
fontosm-nanumgothic

Font "Nanum Barum Gothic"
fontosm-nanumbarumgothic

The Noto font family is by Google, published under SIL Open Font License https://github.com/googlei18n/noto-fonts
Font "Noto Sans CJK KR Regular"
fontosm-noto-sans-cjk-kr-regular

Font "Noto Sans CJK KR Medium"
fontosm-noto-sans-cjk-kr-medium

Noto seems to have a bigger line height, that should be corrected. Noto is by far more readable compared to the currently used font. Noto aims to be a font with complete international language support, mabe pull request  https://github.com/gravitystorm/openstreetmap-carto/pull/358 and issue #294 can be fixed with it too.

### LINKED ISSUE #2240: Current Korean font has no bold variant.
Current Korean font knows no bold variant. Since many name tags are still bilingual this looks odd and gives the romanized/English an emphasis that should not be there.

screenshot from 2016-07-23 13-18-49

Changing to the Noto font family will solve this https://github.com/gravitystorm/openstreetmap-carto/issues/2204 . Noto has many available font weight variants.

### LINKED ISSUE #2336: Font rendering (Thai): Thai city names are in ugly monospace font
Example: http://www.openstreetmap.org/#map=13/16.8140/100.2643

English | Native | Native (bitmap) | Mapnik
---------- | ------------ | -------- | -------
Phitsanulok | พิษณุโลก | selection_315 | selection_316

Notice how some characters are horizontally squeezed or stretched to make the font monospace. This is readable, though it's particularly ugly and doesn't feel natural to read.

However, labels for Thai provinces seem to be working fine with variably-spaced letters (http://www.openstreetmap.org/#map=9/16.3188/99.5087):

English | Native | Mapnik
---------- | ------------ | -------
Changwat Kamphaeng Phet | จังหวัดกำแพงเพชร | selection_323

Could we get a better font for Thai city labels? We seem to have one working for province labels.

### LINKED ISSUE #2337: Font rendering (Sinhala): Sinhala font renders as garbage
Example: https://www.openstreetmap.org/#map=8/7.844/80.947 

English | Native | Native (bitmap) | Mapnik
---------- | ------------ | -------- | -------
Sri Lanka | ශ්‍රී ලංකාව | selection_322 | selection_320

This could probably be fixed by using a better font, such as Noto Sans Sinhala.

### LINKED ISSUE #2338: Font rendering (Tibetan): Tibetan font is too small to read
Example: https://www.openstreetmap.org/#map=7/27.113/89.769

English | Native | Native (bitmap) | Mapnik
---------- | ------------ | -------- | -------
Thimphu | ཐིམ་ཕུ | selection_318 | selection_319

For Tibetan labels to be useful, the font size needs to be bigger.

sommerluk:
Uses Noto as default font.
Fixes #2338
Fixes #2337 
Fixes #2336
Fixes #2240
Fixes #2204
Fixes #1335 (only if DejaVu is dropped)
Fixes #1067
Fixes #677
Fixes #486
Fixes #362
Fixes #294

We loose the oblique style for Tamil and Thai, but we gain the bold style for almost all scripts.

Noto covers most scripts of the world “with a harmonious look and feel”.

Furthermore, the Unifont list is extended (added Unifont Upper).

TODO:

- [ ] What are the performance implications of such a long font list?
- [x] Can some fonts be removed because they correspond to "dead" languages not used on the ground?
- [x] Can some fonts be removed because other Noto fonts cover all the glyphs they have?
- [x] Is the ordering right on bold with Noto Kufi Arabic Bold before the regular fonts?
- [x] Is the way we're doing CJK right? If Noto Sans CJK JP Regular has all the glyphs, the other CJK fonts will never be used. The Noto Guidelines mention there is both "Noto Sans CJK {JP, KR, SC, TC}" and "Noto Sans {JP, KR, SC, TC}", what's the difference?
- [x] Update INSTALL.md


sommerluk:
The list contains _all_ Sans fonts of the current Noto release. For Arabic it uses Noto Kufi (I don’t know much about Arabic). For Emoji is uses the monochromatic Noto version. Noto Serif and Noto Mono are not used (is covered yet by Noto Sans).

Only one of the CJK fonts is necessary.

Only either the UI font or the non-UI font is necessary. If both are available, the UI font is prefferred.

The list of bold Noto fonts is the same as for regular Noto fonts. While not all of them are yet currently available, the next release of Noto will add many new bold fonts. Fallback is the regular font.

The previously used fonts are still in the list (at the end), so someone who updates the style and does not have Noto will get the same result as in previous versions of the style.

Documentation still needs to be updated (INSTALL.md…)

sommerluk:
Screenshots from South Korea/Seoul

Left: before.

Right: after.

screenshot 1
screenshot 2
screenshot 3
screenshot 4
screenshot 5
screenshot 6
screenshot 7
screenshot 8
screenshot 9
screenshot 10

HolgerJeromin:
fixes #2265, too?

style.mss
```diff
@@ -2,26 +2,335 @@ Map {
...
+3. Noto provides two variants of Emoji: Noto Color Emoji and Noto Emoji. The
+colour variant is a SVG flavoured OpenType font that contains coloured emojis.
+This is not useful in cartography, so we use the “normal” monochromatic
+Noto Emoji.
+
+4. The list still includes the previously used font selection of this style
+to remain backward-compatible.
+
+5. Unifont. This is a simple and not really beautiful font, but it has an
+extremely great coverage. Unifont Medium covers the hole Unicode BMP without
```
pnorman: s/hole/whole.

style.mss
```diff
@@ -2,26 +2,335 @@ Map {
...
+                "Noto Sans Tai Tham Bold", "Noto Sans Tai Viet Bold",
+                "Noto Sans Tamil UI Bold", "Noto Sans Tamil Bold",
+                "Noto Sans Telugu UI Bold", "Noto Sans Telugu Bold",
+                "Noto Sans Thaana Bold",
+                "Noto Sans Thai UI Bold", "Noto Sans Thai Bold",
+                "Noto Sans Tibetan Bold", "Noto Sans Tifinagh Bold",
+                "Noto Sans Ugaritic Bold", "Noto Sans Vai Bold",
+                "Noto Sans Yi Bold",
+                "Noto Kufi Arabic Bold", "Noto Emoji Bold",
+                "Noto Sans UI Regular", "Noto Sans Regular",
```
pnorman: We duplicated the regular fonts list into the bold list because for the specific fonts we were using there were some that had regular glyphs but not bold. Is this also the case with Noto? If not, we can drop the duplicates.

style.mss
```diff
@@ -2,26 +2,335 @@ Map {
...
+for backward-compatibility and Unifont as fallback.
+
+By order:
+
+1. Noto Sans is available for most scripts and it is used here. The list in is
+alphabetical order. Note that the “Noto Sans” base font MUST be before the
+“Noto Sans CJK *” fonts, because the CJK fonts contain non-Noto latin characters
+that should not be used. Noto Serif and Noto Mono are available for various
+(mostly latin-like) scripts. They are ignored here in favour of the respective
+Noto Sans variant. For fonts that are known to have an UI variant (a variant
```
pnorman: We can cut the Serif and Mono explanation. We don't ignore them in favour of a Noto Sans variant, it's that we don't use serif or mono fonts in the stylesheet so have no need for them.

style.mss
```diff
@@ -2,26 +2,335 @@ Map {
...
+About fonts:
+
+Noto is a font family that wants to cover most of Unicode with a harmonic
+design across various scripts. We use Noto for most text, with some support
+for backward-compatibility and Unifont as fallback.
+
+By order:
+
+1. Noto Sans is available for most scripts and it is used here. The list in is
+alphabetical order. Note that the “Noto Sans” base font MUST be before the
```
pnorman: Is alphabetical order the best way to do this?

style.mss
```diff
@@ -2,26 +2,335 @@ Map {
...
+“Noto Sans CJK *” fonts, because the CJK fonts contain non-Noto latin characters
+that should not be used. Noto Serif and Noto Mono are available for various
```
pnorman: Perhaps we can reword this because right above we say it's in alphabetical order

style.mss
pnorman: We've called unifont a "fallback of last resort" which is a good explanation for it.

style.mss
```diff
@@ -2,26 +2,335 @@ Map {
...
+used signs like parenthesis or quotation marks, and the arabic design should
+not overwrite the standard design.
+
+3. Noto provides two variants of Emoji: Noto Color Emoji and Noto Emoji. The
+colour variant is a SVG flavoured OpenType font that contains coloured emojis.
+This is not useful in cartography, so we use the “normal” monochromatic
+Noto Emoji.
+
+4. The list still includes the previously used font selection of this style
+to remain backward-compatible.
```
pnorman: If we're happy with Noto we can cut out the previous list. I might still keep DejaVu around for compatibility, but not the others.

pnorman:
There are a few higher level considerations here. GitHub's image hosting is down at the time I'm writing it so I'll come back and add more detail later

* Do we want to replace DejaVu Sans with Noto Sans? This could have a wide change on the look of the style
* Is the choice of UI variants over normal one we want to make?
* If there is both a UI variant and a normal one are there any glyphs that the normal one has but not the UI?
* What are the performance implications of **such** a long font list?
  * Can some fonts be removed because they correspond to "dead" languages not used on the ground?
  * Can some fonts be removed because other Noto fonts cover all the glyphs they have?
* Is the ordering of CJK fonts one we are happy with?

pnorman:
I just want to note that I'm trying pull request reviews (#2348) here

style.mss
sommerluk: The first group is Noto Sans. Within this group, there is indeed alphabetical order (including CJK). For CJK the requierement is that Noto Sans is before it, and this is the case in alphabetical order.

Noto Kufi is not Sans. It’s a group of itself.

style.mss
sommerluk: Ok.

style.mss
sommerluk: Ok.

style.mss
sommerluk: As far as I know this is not the case with Noto, so we can drop the duplicates.

style.mss
sommerluk: Ok.

sommerluk:
> Do we want to replace DejaVu Sans with Noto Sans? This could have a wide change on the look of the style

I would recommend this. It will be anyway a wide change if we keep DejaVu, so I would use this ocasion to get a consistent set of fonts.
> Is the choice of UI variants over normal one we want to make?

When playing around, I could not find a difference between the glyphs of UI and the standard variant. The only difference seems to be the vertical default line spacing, which is smaller for the UI variant, and the Noto documentation claims that they are designed for user interfaces with restricted vertical space. This might be useful, because often there is not much place on the map. I would therefor preffer the UI variant.
> If there is both a UI variant and a normal one are there any glyphs that the normal one has but not the UI?

As far as I know: No. So the normal one could be dropped.
> What are the performance implications of such a long font list?

Experience: When installing a missing font while Kosmtik was running, Kosmtik was not able to render with the new font until Kosmtik was terminated and re-started. So maybe when a style is loaded, the fonts are searched only at the beginning, and not-installed fonts of the list are not searched again?
> Can some fonts be removed because they correspond to "dead" languages not used on the ground?

If this is deisred, yes.
> Can some fonts be removed because other Noto fonts cover all the glyphs they have?

Apart from the yet discussed bold/regular-fallback and UI/non-UI duplicates: All but one of the four CJK fonts – they have all the same glyph coverage. While this are only three font entries less, this can save a lot of disk space, because the CJK fonts are the biggest ones (about 70 MB).
> Is the ordering of CJK fonts one we are happy with?

- In Korea Han seems to be used seldom. They use mostly Hangul, so Han Unification is not such a big problem for Korea. See https://github.com/gravitystorm/openstreetmap-carto/issues/2208#issuecomment-230112180 We would not gain much with CJK KR as default. So CJK KR would be a bad choise for a default.
- So the default should be either Chinese (simplified or traditional) or Japanese. Looking around in the Internet, it seems that for Japanese people it is more important to have Japanese signs and not Chinese signs, while Chinese people are more tolerant to Japanese signs (maybe also because with “simplified” and “traditional” there are yet two parallel versions within their own language). I would use the Japanese variant as default.

sommerluk:
> fixes #2265, too?
Not exactly. But it would at least make the problem in #2265 less visible. And indeed there is no evidently perfect solution for #2265 so far…

pnorman:
> Experience: When installing a missing font while Kosmtik was running, Kosmtik was not able to render with the new font until Kosmtik was terminated and re-started. So maybe when a style is loaded, the fonts are searched only at the beginning, and not-installed fonts of the list are not searched again?

I'm not so worried about map file load time and checking for fonts, it's rendering a particular label where the glyphs aren't found in the first fonts: https://github.com/gravitystorm/openstreetmap-carto/pull/414#issuecomment-39151113

pnorman:
I'm having a hard time getting system fonts working with Kosmtik so haven't been able to review yet.

pnorman:
Testing with 18/41562/89786

Deja Vu

Noto Sans UI

Noto Sans

style.mss
pnorman: Done, but I see I omitted to fix the comment.

pnorman:
Note: Approval of changes was investigating GitHub reviews, not everything is done yet.

pnorman:
10
image

12
image

14
image

18
image


Rendered in Word 2010 on Windows 8.1 with fonts from fonts-noto git repo

pnorman:
Noto Sans and Noto Sans UI are the same on latin glyphs, which is why my tests look pixel for pixel identical.

A "tall" script shows the difference much better

image

Worth noting that https://material.google.com/style/typography.html has a bunch of recommendations on font sizes, line height, and tracking.

pnorman:
Let's go with UI fonts as the only choice if they have the coverage in the language. Worth noting that the download zip doesn't have UI versions, but that doesn't impact the fonts coming from packages (googlei18n/noto-fonts#744).

pnorman:
> The list of bold Noto fonts is the same as for regular Noto fonts. While not all of them are yet currently available, the next release of Noto will add many new bold fonts. Fallback is the regular font.

I missed that before cleaning up the list of fonts. I'm inclined to leave it with my changes because we'll need to check the glyph counts at release.

Remaining issues

* [] What are the performance implications of such a long font list?
* [] Can some fonts be removed because they correspond to "dead" languages not used on the ground?
* [] Can some fonts be removed because other Noto fonts cover all the glyphs they have?
* [] Is the ordering right on bold with Noto Kufi Arabic Bold before the regular fonts?
* [] Is the way we're doing CJK right? If Noto Sans CJK JP Regular has all the glyphs, the other CJK fonts will never be used. The [Noto Guidelines](https://www.google.com/get/noto/help/guidelines/) mention there is both "Noto Sans CJK {JP, KR, SC, TC}" and "Noto Sans {JP, KR, SC, TC}", what's the difference?

sommerluk:
Candidates for elemination (based on a quick wikipedia search for completly dead languages):

Avestan
Brahmi
Carian
Cuneiform
Cypriot
Deseret
Egyptian Hieroglyphs
Glagolitic
Gothic
Imperial Aramaic
Inscriptional Pahlavi
Inscriptional Parthian
Kaithi
Kharoshthi
Linear B
Lycian
Lydian
Meetei Mayek
Ogham
Old Italic
Old Persioan
Old South Arabian
Old Turkic
Phags Pa
Phoenician
Rejang
Runic
Syloti Nagri
Ugaritic

sommerluk:
About performance: Wikipedia has at https://en.wikipedia.org/wiki/List_of_writing_systems#List_of_writing_scripts_by_adoption a list of writing systems ordered by number of people who use it. Maybe we want to order our list like this?

sommerluk:
> "Noto Sans CJK {JP, KR, SC, TC}" and "Noto Sans {JP, KR, SC, TC}", what's the difference?

The former contains all available CJK glyphs in all language variants. Whenever a glyph has various language variants, than {JP, KR, SC, TC} is the default one. The other variants are available with the OpenType LOCL feature.

The latter does not contain all available CJK glyphs, but only the subset that is necessary for the choosen language {JP, KR, SC, TC}. That means that less characters/codepoints are covered, and for the covered characters, it does not contain any alternative LOCL variants.

sommerluk:
Now the dead scripts are removed. Also the CJK duplicates are removed.

Copied the list of remaining issues to a task list within the first comment. Marking the above mentioned 2 issues as solved.

sommerluk:
Moved Kufi Bold after all Sans fonts. Will probably make no difference, but might be better nevertheless depending on future changes in Noto.

sommerluk:
@pnorman 

> Can some fonts be removed because other Noto fonts cover all the glyphs they have?

CJK duplicates have been eleminated.

There are only very duplicate coverage in the Bold list (with both the bold face and the regular fallback). I suppose you left these duplicates intentionally because the Bold style does not fully cover what the Regular style covers.

There is no further duplicate coverage.

I would consider the duplicate coverage as solved.

sommerluk:
From my point of view, there are 2 remaining issues:

> What are the performance implications of such a long font list?

After eleminating the dead scripts and some duplicates, the list is smaller than before, but still large. Having now “Noto Sans UI” in the first place and “Noto Sans CJK *” in the second one should be good for performance because many labels will be covered by these 2 fonts (they are the 2 most commonly used writing systems according to Wikipedia). I suppose that tuning the rest of the font entries would not make a big difference. However, I don’t know anything about performance testing. Could you have a look at this?

> Update INSTALL.md

Could you also have a look at this? (Me english is terrible.)

pnorman:
@ajashton, do you have any more experience with Mapnik and long font lists since https://github.com/gravitystorm/openstreetmap-carto/pull/414#issuecomment-39151113?

pnorman:
Given that the only issues remaining are technical ones, it'd be nice to get some more review of the cartography changes. If someone has problems running this code locally and getting results for all languages, this patch might help

```patch
diff --git a/style.mss b/style.mss
index f10213e..7bc0e5b 100644
--- a/style.mss
+++ b/style.mss
@@ -1,5 +1,6 @@
 Map {
   background-color: @water-color;
+  font-directory: url("file:///usr/local/fonts/");
 }

 /*
```
This isn't needed in production because fonts are set up properly there.

I'll be providing some screenshots, but there's no substitute for browsing it yourself.

pnorman:
The world-wide database on my dev machine is slightly out of date, so different text, label position, or anything other than the fonts is not relevant.

v2.43.0 | PR

ArabicArabic noto

HebrewHebrew Noto

RussianRussian Noto

ThaiThai Noto

Cambodia labelCambodia label Noto

JapaneseJapanese Noto

KoreanKorean Noto

KoreanKorean Noto

Chinese
Chinese Noto

ChineseChinese Noto

MixedMixed Noto

MixedMixed Noto

? Noto

pnorman:
Areas of mixed scripts are **much** improved. You can see this particularly in the Arabic example, where it feels like the x-height is more consistent. The examples from Taibet and Nepal with 3-4 languages also look more similar, even though the scripts are so different.

The Korean font weights look the same, which is a big improvement.

The Chinese examples look almost bold.

nicolas17:
Friend from India says:
> At least with the Devanagari script in Nepal, the right screenshot is much more readable.

kocio-pl:
Strange problem - in this example from China:



the name on the left looks better, but left bracket is broken.

sommerluk:
> The Chinese examples look almost bold.

@pnorman This might be a configuration problem. At least for me it renders fine (=regular) locally:

screenshot 1

sommerluk:
> left bracket is broken. 

Good catch.

Your example contains a label in the following schema:

Tibetian name (CJK name)

Apparently this occures when _combining_ characters in Noto Sans Tibetian are used: Then, an opening bracket that follows later is not rendered.

- If you substitute the tibetian part of the label with latin character, than the issue disappears.
- If you delete the _combining_ tibetian characters (while maintaining the simple tibetian characters), than the issue disappears.
- If you delete Noto Sans Tibetian from style.mss, than the issue disappears.

The opening bracket glyph is available in Noto Sans UI (and Unifont) and should therefore be rendered.

I suppose this is a bug in Mapnik and/or the underlying libraries at the crossroad between complex combining characters/substitution on one hand, and font file choise on the other hand.

For the record: It is possible to render this label correctly with Noto. Here a PDF made with Scribus (CTL development version):

[sample.pdf](https://github.com/gravitystorm/openstreetmap-carto/files/478679/sample.pdf)

stephankn:
Thai: In your new rendering the Arabic numbers are rendered too high, compared to the text. Very good visible with the highway exit numbers. Also on the soi 12.

In general the font is still too small to be proper readable.

new rendering

sommerluk:
> In general the font is still too small to be proper readable.

Yes. This is a general design problem when you use various different scripts. See #2265 for a more detailed description of the problem and why there is no easy solution. There is issue #2209 that tries to raise the font size at least to 10, but this will not be a (full) solution either.

sommerluk:
> In your new rendering the Arabic numbers are rendered too high

Thanks for reporting. Do you think this is a big issue or rather a stylistic variant? I fear there is not much we can do about. The vertical position of a glyph relative to another glyph is a design property of the font we use.

Apparently the Noto designers made this design choise. I could imagine that this was done to fit the thai glyphs to the x high of the latin glyphs to get an overall harmony. So the glyphs of the standard (lining) figures/numbers are higher than thai glyphs because lining figures are higher than the small latin x character. However, I’m just guessing.

In typesetting exist “capital” numbers = “lining figures” on one hand, and “small” numbers = “old style figures” on the other hand). Noto Sans provides currently only lining figures, so there is not much room for tuning this.

For the record: We are talking about this node: http://www.openstreetmap.org/node/280389087

Attachment: The corresponding label as PDF in Noto (done with Scribus CTL devel).
[thai-numbers.pdf](https://github.com/gravitystorm/openstreetmap-carto/files/478743/thai-numbers.pdf)

math1985:
> Apparently the Noto designers made this design choise. 

Just to be sure, does this occur in both the UI and the non-UI variant? Is there anywhere we can report issues to the Noto developers?

stephankn:
> Thanks for reporting. Do you think this is a big issue or rather a stylistic variant?

@sommerluk It just looks strange. Not sure what you wanted to show with the PDF. In that PDF I see no problem. The bottom line of the numbers and the character are the same. No shift like in the map rendering. So now I'm a bit confused. Wasn't the PDF also supposed to be Noto? Is the font not embedded and my machine uses a different font for rendering?

mboeringa:
> Wasn't the PDF also supposed to be Noto? Is the font not embedded and my machine uses a different font for rendering?

Fonts seem properly embedded (viewed on Windows 7):
noto_test_pdf

sommerluk:
> Just to be sure, does this occur in both the UI and the non-UI variant?

In Scribus, is does not seem to make any difference.

> Is there anywhere we can report issues to the Noto developers?

https://github.com/googlei18n/noto-fonts/issues/new but I don’t think that this is an issue of Noto.

> The bottom line of the numbers and the character are the same. No shift like in the map rendering. 

@stephankn Sorry, I missunderstood the problem you reported. (I thought it was about glyph high, when in reallity you talked about base line position.) Now I understand.

However, I was not able to reproduce your problem. On my local machine I see this (500% resolution):
sample
Here, everything seems to be fine.

Do you have all necessary fonts (Noto Thai UI and Noto Sans UI – not the normal, but the UI variant) installed?

> Wasn't the PDF also supposed to be Noto? Is the font not embedded and my machine uses a different font for rendering?

Indeed the PDF is Noto and all fonts are embedded.

sommerluk:
@pnorman While I didn’t had installed Noto Sans Thai UI, the whole label “ต.2-10” was rendered in Unifont. I would have expected that the Thai character is rendered in Unifont, but the rest in Noto Sans UI. After installing Noto Sans Thai UI everything worked fine. Do you know how exactly the glyph search/font order works? Apparently it’s not individual per glyph…

stephankn:
If that base-line discrepancy from these samples https://github.com/gravitystorm/openstreetmap-carto/pull/2349#issuecomment-247819822 was just due to it being Unifont instead of Noto, does this mean the samples need to be recreated? In this sample: https://github.com/gravitystorm/openstreetmap-carto/pull/2349#issuecomment-247856109 everything looks fine (supposed this is now Noto).

pnorman:
> @pnorman While I didn’t had installed Noto Sans Thai UI, the whole label “ต.2-10” was rendered in Unifont. I would have expected that the Thai character is rendered in Unifont, but the rest in Noto Sans UI. After installing Noto Sans Thai UI everything worked fine. Do you know how exactly the glyph search/font order works? Apparently it’s not individual per glyph…

Not sure, I've only been testing with the packaged Noto Sans.

> If that base-line discrepancy from these samples #2349 (comment) was just due to it being Unifont instead of Noto, does this mean the samples need to be recreated?

None of the samples I posted have unifont in them, with the **possible** exception of the square box on the Tibetian-Chinese mix.


Scaled up my setup looks like this
image
If I force it to use unifont, I get this
image

pnorman:
If I switch from version 2013-04-11-2 from Ubuntu/Debian to unhinted and hinted from https://github.com/googlei18n/noto-fonts/commit/c8545d72b80cbe5b21fcc846db5bb4b09dc88a05 I get  
image

So it looks like the difference comes from different versions of the font, not anything we can change on the stylesheet. The same for the boldness in Chinese, see 
image

Given the feedback so far, I'm 👍 on merging once INSTALL.md is updated.

1ec5:
Would you mind posting some screenshots of Vietnamese text? The current DejaVu fonts make it impossible to tell the many Vietnamese diacritics apart (and some diacritics become all but invisible half the time due to antialiasing), but I’m hopeful that Noto will be an improvement. Here are a few example tile coordinates that currently exhibit poor legibility, but any tile in Vietnam would work:

9/407/240
15/26091/15395
15/26015/14425

pnorman:
> Here are a few example tile coordinates

Can you provide lat and long?

pnorman:
v2.43.0 | PR
imageimage

imageimage

imageimage

You can tell the worldwide DB I have is a bit old.

1ec5:
Thanks! It appears to be a slight improvement over DejaVu. For example, ồ, ố, and ổ are now distinct, and Tân no longer looks like Tàn due to keming. In the third screenshot, the diacritic on ả has become more distinct for bodies of water, but a little less distinct for buildings. At the small sizes used for service roads and the like, diacritics will continue to be illegible, but any font that exaggerates the diacritics further would probably start to look odd for Western European alphabets anyways. So this is a 👍.

sommerluk:
Like @pnorman I’m also okay with merging it as-is.

imagico:
Since the rendering of Arabic is very different it would probably be good to get some feedback from routine Arabic readers on the matter.  To me it seems it might be significantly better readable at small font sizes but on the other hand the proportions of the characters are so very different from common fonts that it might require a lot to get used to.

neuhausr:
Looking at the Japanese example, I agree with @pnorman 's comment that the new version almost looks bold. In many cases, that's an improvement, but for more complex kanji characters, it makes the strokes run together and thus be harder to read.

pnorman:
> Since the rendering of Arabic is very different it would probably be good to get some feedback from routine Arabic readers on the matter. To me it seems it might be significantly better readable at small font sizes but on the other hand the proportions of the characters are so very different from common fonts that it might require a lot to get used to.

I couldn't find an Arabic reader, but I had a look at some online Egyptian maps and road navigation signs and the font metrics were closer to what this PR does than traditional writing.

math1985:
Thanks, great work @sommerluk!

notxyz27:
image
image
Chinese characters look good to me. However, in the second image, the character for 门 is weird. I found that this weird character also shows up when I type 门 using Simplified Chinese input in Chrome address bar but the correct character shows up when I type in search box. I'm kind of sure the weird character is not traditional Chinese.

sommerluk:
@notxyz27 Thanks for reporting this.

The character that you use in your description is “U+95E8 CJK UNIFIED IDEOGRAPH-95E8 : mén” (Unihan at http://www.unicode.org/cgi-bin/GetUnihanData.pl?codepoint=95E8 )

In the attached PDF you can see this character (and its semantic counterpart U+9580) in all available Noto glyph variants and in Unifont. What you expect to see is probably number 3. But this style is currently rendering number 1.

In a nutshell: The character is a unified ideograph with different glyph variants. As openstreetmap-carto does not support region-specific rendering and has no knowledge about language in which a label is, it’s not possible to really determine the correct glyph variant. Currently, in case of conflict this style defaults simply to the Japanese variant. This is, indeed, a somewhat arbitrary choise, but we are forced to make a some choise.

In all details: #2208 

I fear there is not much we can do about this issue.

thejeshgn:
Can you share the screenshot with Kannada language? name:kn

pnorman:
> Can you share the screenshot with Kannada language? name:kn

Can you link to an area with the name tag in Kannada?

thejeshgn:
@pnorman 

Here
http://www.openstreetmap.org/#map=12/12.9550/77.6052

The map below displays Kannada labels using name:kn this will give you an idea. 

https://thejeshgn.github.io/tangrams-kannada/#12/12.9550/77.6052


Thank you.

pnorman:
> The map below displays Kannada labels using name:kn this will give you an idea.

We render the name tag, not name:kn.

thejeshgn:
Ah okay.

dieterdreist:
2016-09-23 12:19 GMT+02:00 Thejesh GN <notifications@github.com>:

> Ah okay.



it's a bit offtopic here, but if this is a billingual area you might
consider change the tagging (in accordance with the local community), add
the english names to name:en tags, the kannada names to name:kn and put
both together in the name tag, like it is done elsewhere, e.g. here:
http://www.openstreetmap.org/node/1635651356 or here:
www.openstreetmap.org/relation/47046 (if you want to follow up, please not
here but on tagging or talk lists for example)

stephankn:
It is off topic.  and mixing different languages for rendering into the name tag is not the way it is done in OSM. Multiple languages go into their own tags, then the rendering can decide which of them and how to display.

StyXman:
I'm not so sure; you have the example of België/Belgique/Belgien, where most of the labels are at least in two official languages (and I'm not sure which is the third one).

SomeoneElseOSM:
"mixing different languages for rendering into the name tag" is a bad idea for all sorts of reasons, but some communities have chosen to map names this way.  Have a look at North Africa for some examples, and http://forum.openstreetmap.org/viewtopic.php?id=31333 for a discussion of some of the issues (confounded there by a mix of left-to-right and right-to-left names).

"Making it easier to have a map showing names in $language" is a great idea for OSM, but it's not necessarily doable in osm-carto.

kocio-pl:
> "Making it easier to have a map showing names in $language" is a great idea for OSM, but it's not necessarily doable in osm-carto.

It's worth to remember, that AFAIK it's not because of technical difficulties, but internal goals and policy. We just try to avoid anything region- or country-specific.

dieterdreist:
sent from a phone

> Il giorno 23 set 2016, alle ore 15:38, SomeoneElseOSM <notifications@github.com> ha scritto:
> 
> "mixing different languages for rendering into the name tag" is a bad idea for all sorts of reasons, but some communities have chosen to map names this way.


it is just a result of the concept of the name tag, which is there for the default name, and in bilingual areas with 2 languages (or even 3) being the default it is logical to put that into this tag, just like it is done for signs: 

deleting the one you don't like is what fascists do:

PR #2351, opened 2016-09-17
Render both housenumber and housename

### LINKED ISSUE #735: Display both name and house number for houses that have both
The Mapnik rendering layer should display both the name and the housenumber of buildings. Right now, the housenumber suppresses the display of the name. I think a smaller font than is usual for the name label would be appropriate.

See also https://trac.openstreetmap.org/ticket/2240.

### LINKED ISSUE #2328: Rendering both, addr:housename and addr:housenumber?
In #2298 there was also a discussion about cases where addr:housename and addr:housenumber coexiste. Currently, if both exist, we render only one of them.

There are about 278 000 objects with addr:housename in the database. About 56% of them have also addr:housenumber, this is a considerable part.

It might be possible to render in such cases both of them, with a line break as separator, like this:
```
addr:housenumber\n
addr:housename
```
Would this be desirable?

sommerluk:
Render both housenumber and housename if both are present. Same colour for all address-related features. A little bit of documentation.

Fixes #2328 
Fixes #735

### DIFF
--- a/addressing.mss
+++ b/addressing.mss
@@ -10,31 +10,22 @@
   }
 }
 
-#housenames {
+#addresses {
   [zoom >= 17] {
     text-name: "[addr_housename]";
     ["addr_housenumber" != null] {
-      text-name: [addr_housenumber] + "\n" + [addr_housename] ;
+      text-name: [addr_housenumber];
+      ["addr_housename" != null] {
+        text-name: [addr_housenumber] + "\n" + [addr_housename];
+      }
     }
     text-placement: interior;
+    text-min-distance: 1;
     text-wrap-width: 20;
     text-face-name: @book-fonts;
     text-fill: @address-color;
-    text-size: 10;
-    [zoom >= 20] {
-        text-size: 11;
-    }
-  }
-}
-
-#housenumbers {
-  [zoom >= 17] {
-    text-name: "[addr_housenumber]";
-    text-placement: interior;
-    text-min-distance: 1;
-    text-wrap-width: 0;
-    text-face-name: @book-fonts;
-    text-fill: @address-color;
+    text-halo-radius: @standard-halo-radius;
+    text-halo-fill: @standard-halo-fill;
     text-size: 10;
     [zoom >= 20] {
         text-size: 11;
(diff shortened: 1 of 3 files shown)


nebulon42:
I think that darkening the line colour of address interpolation requires previews.
Why not merge the whole housenumber, housename business into one layer?

sommerluk:
On the left : before                   –                     On the right: after
screenshot 1

polarbearing:
Can somebody guide me into what difference I am supposed to see in that left/right comparison?

sommerluk:
@nebulon42 had requested at https://github.com/gravitystorm/openstreetmap-carto/pull/2351#issuecomment-247860105 a preview:

> I think that darkening the line colour of address interpolation requires previews.

Indeed the line colour has slightly changed – though the change is almost invisible.

pnorman:
`addr:housename` is very seldom found in North America, so I don't feel able to review this.

nebulon42:
Some screenshots:

z17
z17_a
z17_b

house names only (current rendering)
z17_housename_only

z18
z18
z18_b
z18_c

sommerluk:
New commit: Unify address layers and use halo.

Screenshot (aritificial example data):
screenshot 1

addressing.mss
```diff
@@ -1,52 +1,51 @@
...
-    text-face-name: @book-fonts;
-    text-fill: @address-color;
-    text-size: 10;
-    [zoom >= 20] {
-        text-size: 11;
+    text-name: "[addr_housename]";
+    ["addr_housenumber" != null] {
+      text-name: [addr_housenumber];
+      ["addr_housename" != null] {
+        text-name: [addr_housenumber] + "\n" + [addr_housename];
```
math1985: Wouldn't this add a spurious \n for nodes with house name and without house number?

nebulon42: No, as this line matches only if addr_housenumber is not null. So addr_housenumber and addr_housename must be different from null.

aceman444:
See https://github.com/gravitystorm/openstreetmap-carto/issues/2328. What is the purpose of rendering addr:housename on 10 shops inside a building that itself has that name?

sommerluk:
@aceman444 We do not render addr:housename (nor addr:housenumber) for shops. We render it only for buildings.

aceman444:
OK, that's better. But on a building, the same value would be in 'name' so what is the benefit of rendering addr:housename ?

sommerluk:
> But on a building, the same value would be in 'name'
No. Only 44% of the 280 000 elements with “addr:housename” have also “name”.

> so what is the benefit of rendering addr:housename ?
Getting rendered the “addr:housename” on buildings when there is no “name” tag.

As stated in the source code documentation:
> Building names (rendered differently from addresses because they are no official postal addresses)

On a building, “name” is not related to postal addresses, but “addr:housename” is. If both are present, “name” has higher priority.

aceman444:
'addr:housename' on a building without 'name' makes no sense to me. But it is true the addr: wiki page is ambiguous about when to use addr:housename.

But rendering it may encourage tagging addr:housename instead of name (which I have already seen in the wild).

Issue #49, opened 2013-05-13
Check for programs before downloading in get-coastlines.sh


pnorman:
Currently get-coastlines.sh assumes the presence of all the programs it is going to use.

Checking before downloading would be a good idea.

PR #2360, opened 2016-09-21
Check the apps used in get-shapefiles.sh

kocio-pl:
Resolves https://github.com/gravitystorm/openstreetmap-carto/issues/49.


kocio-pl:
Oh, now I see that it will be also obsoleted by https://github.com/gravitystorm/openstreetmap-carto/issues/2137 - so it depends on how many time it will take to have new script ready, since it's already 4 months.

Issue #1022, opened 2014-10-07
amenity=prison area should be labeled based on size


1ec5:
Non-building `amenity=prison` areas are labeled with a brown jailbird icon and name at z17 and above, but nothing at z16 or below ([example](http://www.openstreetmap.org/#map=16/39.9096/-83.4688)). The label should appear based on the area’s area (like #941), since some prison complexes can be quite large.

![z15](https://cloud.githubusercontent.com/assets/1231218/4537872/382e0850-4de5-11e4-8b66-2c9b586aabe8.png)
![z16](https://cloud.githubusercontent.com/assets/1231218/4537874/3be0181c-4de5-11e4-96d5-533f37ebcabd.png)
![z17](https://cloud.githubusercontent.com/assets/1231218/4537875/3f2ab1b2-4de5-11e4-8344-a70415af12e8.png)

Tomasz-W:
@kocio-pl Please close this ticket as duplicate.
Continue further discussion in https://github.com/gravitystorm/openstreetmap-carto/issues/3284

PR #2361, opened 2016-09-21
Adding icon for amenity=university and amenity=college [WIP]

kocio-pl:
Related to https://github.com/gravitystorm/openstreetmap-carto/issues/120 (partial solution).

Code based on amenity=hospital, but using current university/college label color.

Area, z15
Before
x10fsg0u
After
7gqkxtaj

Area, z16
Before
fvpjydrj
After
fbp9hjni

Area, z17
Before
evlp4xpi
After
dksxozgx

Node, z16
Before
dkmzlijr
After
trxnlh7l

Node, z17
Before
ws6roiqe
After
wg3nt5xd

Node, z18
Before
ertddgfy
After
hywvs2qx


Circeus:
PLEASE not. Having a scaling name for these is one of the best thing about them, and anything with an icon has a name that can't scale anymore according (IIRC) pixel_area. In fact, I personally wish we could reasonably get rid of the hospital icon so hospital names could scale too.

math1985:
> anything with an icon has a name that can't scale anymore according (IIRC) pixel_area

That's not a technical limitation, so if we want scaling labels in combination with icons we can define rules that allow this. Currently, caravan sites already do have a scaling label in combination with an icon.

See also #1355. @Circeus, I would be very interested in hearing your feedback on that issue.

dieterdreist:
sent from a phone

> Il giorno 23 set 2016, alle ore 12:33, Matthijs Melissen <notifications@github.com> ha scritto:
> 
> That's not a technical limitation, so if we want scaling labels in combination with icons we can define rules that allow this. Currently, caravan sites already do have a scaling label in combination with an icon.


but it's a limitation designwise, either we scale the icon along with the text (would be a big change) or it will lead to strange looks (tiny icon compared to text).
Maybe having 2 or even three icon sizes could eventually look ok, it surely would increase code complexity (offsets etc) quite a bit

kocio-pl:
We should probably have a general rule for such cases - see for example similar problem with prison: https://github.com/gravitystorm/openstreetmap-carto/issues/1022

RobJN:
Do you need an icon for a university given that the name will usually include "university" in it?

polarbearing:
@RobJN - is there any research that supports your assumption? OSM defines the use of the university tag for third level institutions, and I know many of them not carrying 'university' in their name. The [MIT](http://web.mit.edu/) is probably the most prominent on the technology sector.

math1985:
> Do you need an icon for a university given that the name will usually include "university" in it?

Not saying we would, but we could show the university icon on lower zoom levels, before the name can be displayed, to indicate the campus location inside the city.

RobJN:
No evidence just my hunch for the UK. Just seems like there was a lot of
attempt to clean up the map render and now loads of icons are being added.
Just wondered whether this was directed by a strategy. I'm sure you agree
that there is a balance between usefulness of icons and clutter.

kocio-pl:
Can we reach some agreement with this PR? If not, I'm gonna close this PR without merging in a few days.

polarbearing:
In favour of merging.

kdudzik:
I don't like this change. Icons suggest nodes or small areas. We could enable this icon for nodes only but enabling it also for areas makes the campuses look worse IMHO.

kocio-pl:
> We could enable this icon for nodes only but enabling it also for areas makes the campuses look worse IMHO.

That would be inconsistent with how we deal with hospitals. Do you think that universities are different in that regard?

math1985:
What about first trying to render the name (in italics), and if it doesn't fit render the icon (without label)?

polarbearing:
> math1985>  name (in italics), and if it doesn't fit render the icon (without label)

hm is that consistent behaviour someone would expect? Would that mean that in lower zoom I see an icon, when I zoom in the name?

> kdudzik>  Icons suggest nodes or small areas. 

First time I hear about such rule. 

>  kdudzik> We could enable this icon for nodes only but enabling it also for areas makes the campuses look worse IMHO.

Why? Maybe you have a campus that is tagged incorrectly, i.e. amenity nodes/buildings within the campus amenity?

Nodes and areas should not be treated differently in rendering, often a node is a first step in mapping before it is expanded to an area.

kdudzik:
@kocio-pl I don't think hospitals are any different - they should be handled the same way, true.

@polarbearing I don't know any rule, that's just my subjective opinion. I'm not quite convinced to it, but I would prefer it to the current proposal.
Maybe @math1985's idea could work.
In my opinion a small icon with a small caption (or even without a caption) in the middle of a large area (which it represents) looks just weird.

kocio-pl:
I think we should first rethink how do we render [similar objects](https://github.com/gravitystorm/openstreetmap-carto/pull/2361#issuecomment-249383602) and then go back to this problem. Thanks for the feedback!

Tomasz-W:
Further discussion in https://github.com/gravitystorm/openstreetmap-carto/issues/3284

Issue #1190, opened 2015-01-04
too strong color for leisure=pitch, leisure=track


matkoniecz:
It was noticeable before, now after building colour change it is quite bad. leisure=pitch, leisure=track should have muted/paler colour.

http://www.openstreetmap.org/#map=14/50.0735/19.9205

![selection_004](https://cloud.githubusercontent.com/assets/899988/5605157/1512299e-93ec-11e4-9916-ca9346d8b3d7.png)

nebulon42:
Definitely, but I would also suggest considering desaturating all landuse colours so that they work better with the new buildings colour. I know that this is not the easiest task to do, though.

math1985:
The landuse colours have been desaturated a lot already recently. Do you think all of them are too saturated, or some in particular?

SomeoneElseOSM:
The colours that are used for "very large landuses where there are other things mapped" (e.g. farmland) are definite candidates.  Farmland at F4EBDF works quite well for me, but other similar ones (large landuses that are often adjacent to farmland) would need to change too.

nebulon42:
> The landuse colours have been desaturated a lot already recently.

I didn't know that. I really like the lighter building colour. I also think that built-up landuse like residential, industrial, retail, commercial could use a bit fine-tuning so that they work better with the new building colour. Overall, landuse colour should be paler (also farmland, forest, grass/meadow). The colour @SomeoneElseOSM suggested for farmland is not bad.

But this needs more discussion, obviously. For now I would propose to adjust residential, industrial, retail, commercial and maybe construction and grass/meadow. Sorry for hijacking this issue, but I had a quick try:

![landuse_color_small](https://cloud.githubusercontent.com/assets/3531092/5606670/f958b8dc-9439-11e4-9944-25da6b43cef6.png)

Please note that the building outline is 5% darker on the right than on the left.

Some map previews:
http://www.openstreetmap.org/#map=17/48.21009/16.24593
![bildschirmfoto vom 2015-01-04 17 50 19](https://cloud.githubusercontent.com/assets/3531092/5606681/437ea7c8-943a-11e4-942b-829ce3946fe2.png)

http://www.openstreetmap.org/#map=16/48.1859/16.3825
![bildschirmfoto vom 2015-01-04 17 52 15](https://cloud.githubusercontent.com/assets/3531092/5606688/7dfc0b7a-943a-11e4-84b2-38e67037b1c3.png)

[...]

Issue #2270, opened 2016-08-05
Color design paradigm in this style


imagico:
With the recent changes/additions to area colors in dog_park (#2216, see also #2250) and playground (#2249) i have to say i can no more recognize an overall design paradigm regarding the area colors in this style.

In the last years in my eyes many improvements have been made in terms of colors here significantly enhancing the overall readability.  When making changes (like the meadow/grassland and the orchard/vinyard unification) i always did so based on the idea of a common understanding of a certain overall direction in which color design is meant to steer and when others made similar changes i did sometimes disagree with the specific choices made but usually  understood and concurred with the general idea.  I have now the distinct impression this is no more the case and color decisions are going back to a local atomic _find a free spot in color space_ and _looks nice to me_ approach which had previously already led to a lot of problematic color choices.

Design is of course always a highly subjective and intuitive field so it is difficult to formulate absolute rules without exceptions.  Good design tends to follow rules but also occasionally breaks them - the key however is doing so deliberately with an understanding of the effects this has and not out of ignorance or laziness.

The [cartographic guidelines of this style](https://github.com/gravitystorm/openstreetmap-carto/blob/master/CARTOGRAPHY.md) offer relatively little substantial advise on the matter, the most significant statement is on [legibility](https://github.com/gravitystorm/openstreetmap-carto/blob/master/CARTOGRAPHY.md#legibility)  - which seems to be a relatively vague concept but in fact can be quite strong as a guiding principle, especially regarding choice of colors - if you get to a point where you do no more just determine legibility based on your personal subjective gut feeling.

I would like to ask the maintainers here to contemplate this matter.  I am not asking for a recipe for good design but i ask you to consider that a design paradigm needs to be visible in the decisions made for others to successfully contribute.  I currently have difficulties recognizing such a paradigm and it seems i am not the only one.  This might still be an aftereffect of the move from a single maintainer to a team of several with naturally sometimes contradicting design ideas.  This should however not prevent the decisions and the larger picture behind them being transparent and understandable to open minded contributors.

kocio-pl:
As we already know, I have quite different POV than you. In a nutshell:
* I agree that legibility is "relatively vague", because it's more general rule, but for me color palette is rather a tool. It's always better to be consistent, but this rule has lower priority in my view, because its scope is limited (for example colors has to be reused in different contexts).
* I like the meadow/grassland and the orchard/vineyard unification, because it doesn't mess up with legibility - it's very good, since both rules apply.
[...]

PR #2363, opened 2016-09-22
Making pitch and track color less dominant

kocio-pl:
Resolves https://github.com/gravitystorm/openstreetmap-carto/issues/1190.

The color definition is not absolute for now to make it easier to test modifications. When it'll be ready to merge, I will remove the "[WIP]" (work in progress) from the title.

After some testing I feel this color (lighten 8%, desaturate 5%) is the best on typical backgrounds:

z14 - it's still visible on the sport center and on the ground, but less than a forest, and small spots on residential area are not a problem now:

Before
jdk619dg
After
4ylsau2h

z15 - pitches are less intense than a special building and secondary roads, now they are more or less like a regular buildings, which is good:

Before
vouzm97o
After
9 xab68c

z16 - when school areas are bigger, it's the same as with sport centres, and we have a special case, pitch on the residential area, which also looks OK on this zoom level:

Before
f9vbphts
After
zypwbbvq

z17 - another typical background, pitch on a grass, you can also see that it's still much different than cemetery green:

Before
wpbsfxf
After
_vw2ny6z

### DIFF
--- a/landcover.mss
+++ b/landcover.mss
@@ -55,7 +55,7 @@
 
 // --- sports ---
 
-@pitch: #80d7b5;
+@pitch: #aae0cb;
 @track: @pitch;
 @stadium: @societal_amenities; // also sports_centre
 @golf_course: #b5e3b5;


kocio-pl:
I'm going to merge it if in a few days if no one will oppose.

imagico:
This does not seem to be codified here i think but it is generally not common to merge your own changes (4-eyes principle).

I am staying out of this matter due to #2270.

The current color by the way was selected in #2071 after #1190.

kocio-pl:
> This does not seem to be codified here i think but it is generally not common to merge your own changes (4-eyes principle).

This happens sometimes (like in https://github.com/gravitystorm/openstreetmap-carto/pull/2409), but I definitely won't push anything when there are any objections.

> I am staying out of this matter due to #2270.

I consider it to be rather objection than just lack of interest, so now I don't plan any action here.

pnorman:
> I'm going to merge it if in a few days if no one will oppose.

No, you shouldn't merge this without review from someone else

imagico:
> I consider it to be rather objection than just lack of interest

Neither is the case - i am just not able to form a qualified opinion here.  I would not change the color this way but i would also not have changed the playground color the way it was changed or the dog_park color.  So i can't really say if i consider this change in isolation as good or bad.

So i am fine if this change finds general approval - but i would appreciate it of course if - in line with #2270 - someone can explain the guiding principles behind these color choices then.

math1985:
>  This does not seem to be codified here i think but it is generally not common to merge your own changes (4-eyes principle).

I commented on that in https://github.com/gravitystorm/openstreetmap-carto/issues/2436#issuecomment-260762730, but my message likely crossed your message (and @kocio-pl's comment).

Issue #2355, opened 2016-09-19
Add icon for shop=funeral_directors


boothym:
Noticed the iD editor has an icon for this, but it's not rendered here. The tag has over 7000 uses: http://taginfo.openstreetmap.org/tags/?key=shop&value=funeral_directors

The gravestone/cemetery icon is available here: https://github.com/bhousel/maki/blob/mb-pages/src/cemetery-24.svg

kocio-pl:
Thank you for opening this ticket! 

I guess this icon does not work standalone (only as an element of cemetery pattern) and is somehow similar to memorial icon. However I was thinking about it for months and I guess simple grave candle [like this](http://image.shutterstock.com/display_pic_with_logo/3984323/373834150/stock-photo-black-candle-373834150.jpg) would be OK. Maybe coffin shape could work too, but it's probably hard to avoid cross on it to make it more recognizable, and that wouldn't be so universal then.

SomeoneElseOSM:
FWIW elsewhere I went with https://github.com/SomeoneElseOSM/openstreetmap-carto-AJT/blob/master/symbols/shop_funeraldirectors.p.16.png - but as you say with a cross on it it's really not universal.

boothym:
@kocio-pl I know you think it doesn't work but could you please render it anyway? Thanks

kocio-pl:
2) ![funeral_directors-14-8](https://cloud.githubusercontent.com/assets/5439713/18691718/85f59ec0-7f95-11e6-8938-0219367375c8.png)

boothym:
Looks like the perfect icon for a candle shop! ;)

pnorman:
With moderate usage and being a *very* specialised shop, I'm inclined towards not rendering with its own icon.

boothym:
It has more than the art, tea and tobacco shops,and almost the same deli - all of which have been discussed recently and submitted. What level of usage would it need to get rendered?

dieterdreist:
2016-09-23 11:49 GMT+02:00 Paul Norman <notifications@github.com>:

> With moderate usage and being a *very* specialised shop, I'm inclined
> towards not rendering with its own icon.
>


why would we not want to show specialised shops? Millions of people are
dying every month. IMHO the less specialised a shop (category) is, the less
interesting it is to show it, because you won't be sure it offers what you
are looking for.

boothym:
[...]

PR #2367, opened 2016-09-24
Adding icon for shop=funeral_directors

kocio-pl:
Resolves https://github.com/gravitystorm/openstreetmap-carto/issues/2355.

I think this should be rendered brown, like shop=car_repair, because it's a service office rather than shop for me and also to show that it's not a candle shop ([37 uses](http://taginfo.openstreetmap.org/tags/shop=candle)).

xl3rv13q


matkoniecz:
This icon for me seems to rather indicate place commemorating something - especially "eternal flame" type.

kocio-pl:
I'm not afraid of this misunderstanding, I think it would be mostly located in buildings.

math1985:
The current icon is unfortunately not clear enough for me either.

PR #2368, opened 2016-09-24
Document CJK glyph style choise

sommerluk:
Documentation for how we made the choise for CJK glyph style. With https://github.com/gravitystorm/openstreetmap-carto/pull/2349#issuecomment-248396392 there was yet a question related to CJK glyph style, so it would be good to document why the current code is as it is.


math1985:
Thanks!

Issue #401, opened 2014-03-13
Add rendering for tourism=apartment


nounours77:
Hello,

there is a new key tourism=apartment. Which is similar to tourism=chalet, but likely to be in more crowded areas. I suggest to render it with the same symbol as tourism=chalet, but starting at lower zoom-level.

thanks, nounours77

matkoniecz:
I think that it would be overkill. Such features are not too important and in some areas it would result in a really bad rendering - there are towns where nearly every house rent apartments. airbnb and similar are making situation significantly worse for such rendering.

I propose closing this request.

pnorman:
Usage is also fairly low

pnorman:
I still find the low usage (2.8k) is a concern. I also don't find it as relevant as other tourism features for a general purpose map. A large part of this is relative insignificance on the ground.

Downtown Vancouver has many apartments to rent
![image](https://cloud.githubusercontent.com/assets/1190866/17681798/a6eac1d8-62fc-11e6-8459-d7066df50ca5.png)

I doubt you could see any of these if you were there in person. They'd be unsigned, even inside the building.

geowOSM:
tourism=apartment are particularly **popular and important** in Austria, Germany, Switzerland and parts of northern Italy, but are widely distributed throughout Europe and the world. https://taginfo.openstreetmap.org/tags/tourism=apartment#map

The current usage of tourism=apartment is relatively low, because mappers are massively inclined to use descriptive names and **misuse tourism=chalet** instead to get holiday apartments rendered. 

Examples from Austria and Italy/South Tyrol 
("Ferienwohnung" means holiday apartment)
http://www.openstreetmap.org/#map=19/47.64739/12.20317&layers=N
![osm_apartment_2016-08-16 um 09 49 44](https://cloud.githubusercontent.com/assets/13821303/17691807/ecb7403e-6396-11e6-8738-7037b15512ff.png)

http://www.openstreetmap.org/#map=18/46.68174/11.07463&layers=N
![osm_apartment_it_2016-08-16 um 10 03 46](https://cloud.githubusercontent.com/assets/13821303/17692184/d977f228-6398-11e6-8cac-ee57564fa5fc.png)


The number of the actually existing tourism=apartment are significantly higher than the low usage might imply on first sight. And to my experience they are almost always visible on spot due to outdoor advertising.

I would suggest to render tourism=apartment with the same symbol as tourism=chalet, but starting at a lower zoom-level.

Thanks
[...]

PR #2369, opened 2016-09-24
Adding icon for tourism=apartment

kocio-pl:
Resolves https://github.com/gravitystorm/openstreetmap-carto/issues/401.

Rendering with chalet icon, as they are similar, but from z18 instead of z17, so it's less visible than shops and chalets.

z18
jz7crnvp


nebulon42:
Great that you attempt to get tourism apartments rendered. However, I'm not too happy with the chalet icon. I would suggest to remove the mountain somehow. A chalet is basically an apartment at some remote area. So the mountain thing should be obvious from the context. Both categories can still share the same icon. For the sake of simplicity this could also happen later.

kocio-pl:
So it would mean redesigning chalet icon anyway, because [it's still not vectorized](https://github.com/gravitystorm/openstreetmap-carto/issues/1165) and blue house seems to be not clear for me for both those tags. Do you have some idea what could it be instead?

nebulon42:
No good one I have to admit. I did not have good ideas regarding accommodation that's why I left them till the end. Maybe this should be tackled later (TM). :-)

kocio-pl:
Another reuse, it's better for me now:
xmg7sk t

pnorman:
Closing as https://github.com/gravitystorm/openstreetmap-carto/issues/401#issuecomment-250432575 closed the parent issue.

Thanks for your work - Is there anything we can pull out for chalet?

Issue #645, opened 2014-06-18
render name and ref of railway=subway_entrance


nighto:
Hello there,

here in Rio, the subway entrances have names and a letter to identify them, which I've been using name=\* and ref=\* to tag. Here's an example, a picture took of the map in Maracanã station:

![](https://lh4.googleusercontent.com/-bses1I5OjB0/U58Rd0ELS0I/AAAAAAABHQk/_yzopGg4_BY/w500-h888-no/IMG_20140616_112553197.jpg)

And how it's mapped today:

http://www.openstreetmap.org/#map=19/-22.90987/-43.23336

Therefore, I'd like to propose/ask that the subway entrance's names/refs be rendered in high levels of zoom (18 and/or 19). I suggest that they could be rendered like highway=motorway_junctions.

What do you think?

Thank you!

Issue #1127, opened 2014-11-11
Make tram and bus stops more similar


matthijsmelissen:
Tram and bus stops are rendered rather different at the moment. They use a different color, different positioning of the name (above versus below), and different minimum zoom level. It might be nice to at least partially unify this.

Tomasz-W:
@kocio-pl @Adamant36 
Looking at some polish cities examples, where there are usually both tram stops and bus stops. I find unifying their colour to transportation-blue as bad idea, because rail transport and road transport are two different things (eg. if there is some blackout or something, trams are stucked, but you still can go by a bus), so it should be intuively distinguishable.

Zoom any tram intersection to see: https://www.openstreetmap.org//#map=14/52.2314/21.0027

I see another problem: similarity of _station-blue_ `#7981b0` and _office-blue_ `#4863A0`. 

Look here (office is on the right): https://www.openstreetmap.org/#map=19/52.21658/20.98165

As I consider current station violetish blue shade as a little bit outdated for modern design, and I don't see any reason for keeping so small difference between some two colours, I think we can reuse _office-blue_ shade also for _station colour_, which would make map colours scheme simpler without making big effort.

What do you think about this issue now?

Adamant36:
@Tomasz-W, I was just thinking about this issue and was totally going to ask @kocio-pl what he thought about it. I agree that violetish blue shade is a little outdated. I don't know about going with the office color though. As it would be inconsistent. Unless we changed the color for offices to make it exclusive. The wider question is if everything transportation related should have the same color or not. 

I think it would simplify things if they did and make the map more coherent, there's advantages and disadvantages to both ways of doing it though. I almost think it would be cool to making everything transport related, included motels/hotels/gas stations/bike parking/bus stops/airports/train stations etc etc as the same blue color as the parking icon. But I probably need to think about it more. Then we could tell the difference between them with the square versus the circle.

Adamant36:
I wonder if its possible to do triangle dots. Then we could have circle/vehicle, train/square, plane/triangle, boat/?

jragusa:
Triangles are already used for peaks. An other possibility is filled/empty squares. `amenity=ferry_terminal` has already an icon.

Adamant36:
Oh yeah, duh. I forgot about peaks. I don't think anyone would confuse an airport for a peak if it was a different color/context. So, it might still be an option. But Filled/empty squares might work instead though.

Adamant36:
[...]

PR #2370, opened 2016-09-24
Adding ref for railway=subway_entrance

kocio-pl:
Resolves https://github.com/gravitystorm/openstreetmap-carto/issues/645.

I guess we need to wait for a database reload to show also refs, but it's possible to render the name of subway entrances right now ([15k+ uses](http://taginfo.openstreetmap.org/tags/railway=subway_entrance#combinations)):
a8ne_h8q

UPDATE: probably ref should be available, but is not in .stations - could somebody help me with it?

### DIFF
--- a/project.mml
+++ b/project.mml
@@ -1654,6 +1654,7 @@ Layer:
         (SELECT
             way,
             name,
+            ref,
             railway,
             aerialway,
             CASE railway 
@@ -1681,6 +1682,7 @@ Layer:
         (SELECT
             way,
             name,
+            ref,
             railway,
             aerialway
         FROM planet_osm_polygon
--- a/stations.mss
+++ b/stations.mss
@@ -7,6 +7,17 @@
     marker-placement: interior;
     marker-fill: @transportation-icon;
     marker-clip: false;
+    [zoom >= 19] {
+      text-name: [ref];
+      text-face-name: @book-fonts;
+      text-size: 10;
+      text-fill: @transportation-icon;
+      text-dy: 10;
+      text-halo-radius: @standard-halo-radius * 1.5;
+      text-halo-fill: @standard-halo-fill;
+      text-wrap-width: 0;
+      text-placement: interior;
+    }
   }
 
   [railway = 'station'][zoom >= 12] {


nebulon42:
`ref` is in the database. You need to select it in the query (https://github.com/gravitystorm/openstreetmap-carto/blob/master/project.yaml#L1653) so that you can use it in CartoCSS. I would recommend the same approach as with housenames and housenumbers.

Two questions come to my mind:
* Why have the font in italics?
* Why not use the icon colour for the text?

kocio-pl:
Fixed now! The problem proved to be I needed to add ref also [here](https://github.com/gravitystorm/openstreetmap-carto/blob/master/project.yaml#L1680), otherwise I got an error when trying to show entrances with ref.

I was reusing tram stop settings, including color, but wanted to make it look a bit different in case there is a tram stop around. Italics is a nice way to do it and it hints that it's something less important.

Look at this example:
0f3u upx

Using station-color for labels does not give visible results. If these entrances have the same name (it may still happen) and ref, it could be confusing, don't you think?

kocio-pl:
Different styles of rendering primary example, because it have both name and ref:
1)
bnav3uwt
2)
txu krea
3)
wu_nzfhp
4)
hmyvof1a
5)
9svbabhz

Now it's clear why we need darker color for labels. This example is hard (which is good for testing!), because refs are hardly visible because of background elements and only 3) really works. So 5) is out of question for me, but we still may go with italics/normal and station-text/standard black. What do you prefer?

math1985:
I'm not sure what the name tag on subway entrances is used for? Is it the name of the entrance, or the name of the station? In the latter case, I'm not sure if rendering it makes much sense.

dieterdreist:
sent from a phone

> Il giorno 26 set 2016, alle ore 12:00, Matthijs Melissen <notifications@github.com> ha scritto:
> 
> I'm not sure what the name tag on subway entrances is used for


name is always referring to the object it is tagged on (i.e. the entrance in this case), although many mappers are not doing it like this;-)

jojo4u:
[Well wiki](https://wiki.openstreetmap.org/wiki/Tag:railway%3Dsubway_entrance) says: _Name of the exit if is known. Not the name of the station_

kocio-pl:
In the updated code I have chosen 1), because confusion with tram stop is not likely (even in the given example it's easy to properly attribute labels to objects) and it's still quite clear without breaking the icon/label colors similarity.

kocio-pl:
Do you propose to show just ref labels? I would be OK with that solution.

However the problem with data still exists: [wiki is clear about not using station name](https://github.com/gravitystorm/openstreetmap-carto/pull/2370#issuecomment-249539068), so it's just a common tagging error or wiki needs to be updated to reflect reality.

pnorman:
> Do you propose to show just ref labels? I would be OK with that solution.
> 
> However the problem with data still exists: wiki is clear about not using station name, so it's just a common tagging error or wiki needs to be updated to reflect reality.

The ref tags I see have the ref of the station or the ref of the lines the station has.

With the data I've seen, I don't see value in rendering any text labels for station entrances.

kocio-pl:
Which cities have you tried? With London, Warsaw, Moscow, Tokio or Paris refs looks OK. New York has this problem, but the station names have it too.

math1985:
I see value in rendering station entrance refs. Some metro systems rely heavy on them (and within metro stations, GPS coverage is usually absent).

Hopefully accepting this PR will lead to better data.

kocio-pl:
You mean only the refs or the names too?

math1985:
I meant the refs.

kocio-pl:
I have updated this PR to render only refs.

pnorman:
> Which cities have you tried? With London, Warsaw, Moscow, Tokio or Paris refs look OK. New York has this problem, but the station names have it too.

New York and Cincinnati airport are the two places with ref on railway=subway_entrance in North America. New York is actually a bit more complicated than I had realized, it doesn't have the ref of the lines at the station, it has the ref of the lines you can reach from **that** station entrance.

Looking at some other areas, I'm neutral on this PR overall.

dieterdreist:
2016-10-16 22:36 GMT+02:00 Paul Norman <notifications@github.com>:

> New York and Cincinnati airport are the two places with ref on
> railway=subway_entrance in North America. New York is actually a bit more
> complicated than I had realized, it doesn't have the ref of the lines at
> the station, it has the ref of the lines you can reach from *that*
> station entrance.
>
> Looking at some other areas, I'm neutral on this PR overall.
>

refs (like all other tags) refer to the object they are attached to, i.e. a
ref on an entrance should be the ref of this entrance, not the ref of
things that can be reached through this entrance. If you want to tag the
latter you should use a different tag.

math1985:
I think the color of the label should match the color of the icon (not sure which of both should match).

kocio-pl:
Are you sure? 

We already use different colors for all the stations and stops (station-color + station-text or transportation-icon + transportation-text). I think this is done by design, to make labels (thin lines) be visible more or less like corresponding icons (thick objects).

In case of subway entrance (transportation-icon) I would rather render labels as transportation-text to be consistent wit the rest than use transportation-icon.

math1985:
> Are you sure? 

I think so - we have been working on making icons and labels match over the past year, and I don't think we should add additional non-matching ones. I thin the stations should be fixed as well. See also #1127.

kocio-pl:
I think that current design has a lot of sense, but that's not a blocking problem for me, so I've changed the code according to your opinion and now it looks like this:
m29zysk5

kocio-pl:
The code is updated now and ready to be merged.

math1985:
Thanks!

Issue #327, opened 2014-02-05
railway=station on area=yes behaves like a building, rather than landuse.


javbw:
When making many things in OSM, simply marking the point is okay, naming a building is good, and marking it's land use, then the objects inside (roads / buildings / walkways, etc) is best (as I understand it)

But there is no land use for station (as opposed to school, industrial, residential, etc) 

so the rendering of railway=station (and public_transport=station?) with area=yes should behave similarly to a land use tag - the area should accept things on above layers (pedestrian areas, platforms, etc) but it currently is does not. 

Please make the the railway=station tag behave in this manner.

javbw:
comment continued from #1457 

I'm bringing this up because tagging it like other objects results in an unexpected action (there's no way to tag a station as a single landuse correctly). And the reasoning for unexpected action seems to not be in line with a lot of other tagging schemes for mapping areas of buildings and other places throughout town - so it causes confusion with new mappers (and me), and is counterintuitive (IMO) - espcially as landuse=* gets wider and wider (the last real holdout is somekind of landuse=civic, whcih Im working on in the tagging mailing list.) There are very few man-made/used areas that can't bedefined by a landuse. Finding stations like this is a real surprise. 

> The area in front of the station (bus stop, taxi booth, car parking, green space) does not fit this definition.

I understand about the landuse=railway not fitting the definition well (because, yea, a bus circle isn't railway), then why isn't there a landuse=transport_station or something similar? Should I suggest it in the tagging list?

Why are railway stations treated completely differently than any other place? It makes no sense - the tracks and the buildings are just as much of the station as the bus circle and other related amenities.   The *only* reason it is a station is because that is where the two come together. having a landuse for the railway side and a landuse for the non-railway side is a failure of the tagging system, IMO - the whole thing is one single place.

Having to tie pieces of things together to make a single whole thing feels very weird. No such weirdness is needed for an airport, though it is where two transportation methods collide (the parking lots, bus stations, and everything sit comfortably in "the airport's" landuse. The whole thing is a transport station. It also feels like this is against the "one thing, one object" that works so well for every other single building and landuse in OSM. With all the other landuses, I'm surprised that there isn't a good one for public transport stations.

I don't need a relation to add a driveway and parking to a mall. A station should be the same - especially with the "stop position" in existence to differentiate the different stop positions on the different lines (as I understand how relations could work).

mboeringa:
[...]

PR #2371, opened 2016-09-25
Making railway stations appear as major buildings [WIP]

kocio-pl:
Related to https://github.com/gravitystorm/openstreetmap-carto/issues/327.

This code is not production ready, because it highlights all the buildings tagged with railway=station, but I want to do it only for building=train_station, to leave all the other buildings alone. Could anybody help me with it?

z18
eqwwzyjz

### DIFF
--- a/buildings.mss
+++ b/buildings.mss
@@ -33,7 +33,8 @@
         line-color: @building-aeroway-line;
       }
     }
-    [amenity = 'place_of_worship'] {
+    [amenity = 'place_of_worship'],
+    [railway = 'station'] {
       polygon-fill: @building-major-fill;
       polygon-clip: false;
       [zoom >= 15] {
--- a/project.mml
+++ b/project.mml
@@ -562,7 +562,7 @@
       "srs": "+proj=merc +a=6378137 +b=6378137 +lat_ts=0.0 +lon_0=0.0 +x_0=0.0 +y_0=0.0 +k=1.0 +units=m +nadgrids=@null +wktext +no_defs +over",
       "Datasource": {
         "extent": "-20037508,-20037508,20037508,20037508",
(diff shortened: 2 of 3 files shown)


dieterdreist:
sent from a phone

> Il giorno 25 set 2016, alle ore 04:05, kocio-pl <notifications@github.com> ha scritto:
> 
> This code is not production ready, because it highlights all the buildings tagged with railway=station, but I want to do it only for building=train_station, to leave all the other buildings alone. Could anybody help me with it?


railway=station on a building is mostly mistagged, because save rare cases train stations extend beyond a building (exceptions are underground stations)

jojo4u:
Currently there is no way to distinguish active and inactive train station buildings. Only way might be wether the building is inside a public_transport=station area.

kocio-pl:
It would make sense if we have something similar to amenity=place_of_worship. But you're probably right that railway=station is not the same [according to wiki](http://wiki.openstreetmap.org/wiki/Tag:railway%3Dstation#Things_to_avoid). It may be changed in the future, but currently we're talking about the whole area.

So for now I would go with [building=train_station](http://wiki.openstreetmap.org/wiki/Tag:building%3Dtrain_station) alone, because this rendering is not defined elsewhere and we're not bound by any specific idea what makes the building "major". One day, when station buildings being a part of the station can be tagged as defunct, we could rethink if we want to consider them major anyway or not.

That means I need to rework the code to use only building=train_station. The problem is that s putting `[building = 'train_station']` in the code [instead of `[railway = 'station']`](https://github.com/gravitystorm/openstreetmap-carto/pull/2371/files#diff-e0fd32db156f4279d089398056e22a79R37) just doesn't work.

math1985:
railway=station is indeed not the right tag. It is also usually used to indicated the stop position. What are the problems with building=train_station?

kocio-pl:
I don't know what I did wrong, but building=train_station works now as a major building [in my branch](https://github.com/kocio-pl/openstreetmap-carto/tree/station-building). Do you think we could reopen it and update the code?

math1985:
> Do you think we could reopen it and update the code?

Sure.

PR #2374, opened 2016-09-26
Drop margin and avoid-edges attributes in placenames

math1985:
* Margin breaks Mapnik 2 support
* Avoid-edges prevent rendering of London and other places on
  metatile boundaries

### DIFF
--- a/placenames.mss
+++ b/placenames.mss
@@ -62,8 +62,6 @@
     shield-face-name: @book-fonts;
     shield-fill: @placenames;
     shield-size: 11;
-    shield-margin: 5;
-    shield-avoid-edges: true;
     shield-wrap-width: 30;
     shield-halo-fill: @standard-halo-fill;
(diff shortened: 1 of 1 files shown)


math1985:
CC @nebulon42

kocio-pl:
> Avoid-edges prevent rendering of London and other places on metatile boundaries

I don't understand it, could you show an example?

math1985:
London is always on a metatile boundary (since it is on the 0 meridian), and Avoid-edges specifies to drop rendering of labels on metatile boundaries.

pnorman:
Also the docs say that avoid edges doesn't work with buffer. That's not 100% correct, but using them in combination requires more review then we gave it.

nebulon42:
Ah that's bad, but we cannot simply remove the margin. We have to replace it with something similar that works for Mapnik 2.

PR #2382, opened 2016-09-28
Update CoC to include ethnicity & national origin

rory:
I propose adding these "ethnicity" and "national origin" to discimrimination section. This mirrors [EU anti-discrimination](http://eur-lex.europa.eu/LexUriServ/LexUriServ.do?uri=CELEX:32000L0043:EN:HTML). Within Europe, there are many examples of racial/ethnic discrimination that might not fit into definitions of "race", and adding ethnicity can help clarify this here.

### DIFF
--- a/CODE_OF_CONDUCT.md
+++ b/CODE_OF_CONDUCT.md
@@ -47,7 +47,7 @@ People are complicated. You should expect to be misunderstood and to misundersta
 These actions are explicitly forbidden in OpenStreetMap Carto spaces:
 
 - Insulting, demeaning, hateful, or threatening remarks.
-- Discrimination based on age, disability, gender, nationality, race, religion, sexuality, or similar personal characteristic.
+- Discrimination based on age, disability, gender, nationality, national origin, race, ethnicity, religion, sexuality, or similar personal characteristic.
 - Bullying or systematic harassment.
 - Unwelcome sexual advances.
 - Incitement to any of these.


pnorman:
We can change ours independently, but has this been suggested to the Go code of conduct which ours is based on?

rory:
I haven't done that yet. Maybe I should...

pnorman:
> I haven't done that yet. Maybe I should...

Otherwise, I could (eventually). Go is a bigger project so has more capacity for reviewing changes like this, and changes to the Go code will impact more conduct overall.

pnorman:
Sent off an email.

I'm leaning in favour of merging, but of course want to see if they have anything to add.

pnorman:
cross-reference https://go-review.googlesource.com/#/c/32133/

pnorman:
I'm going to merge in the upstream changes and then look at the differences again to see if there's anything this covers that they didn't. I know this takes longer, but getting good changes upstream is worthwhile, and so is staying close to what upstream has.

kocio-pl:
If I see correctly (the UI there is totally different than on a GitHub), the upstream change is already merged. What do you plan to do with this PR now?

pnorman:
> If I see correctly (the UI there is totally different than on a GitHub), the upstream change is already merged. What do you plan to do with this PR now?

Replace "race" with "ethnicity" to match the Go changes. With nationality and ethnicity mentioned, do we also need to specifically call out national origin? We do mention "or similar personal characteristic", and I feel nationality and ethnicity cover national origin, but don't have strong views on the subject.

kocio-pl:
I feel the same. I also like having CoC, but don't like splitting hairs when we didn't have such issue to make a reality check for these rules and we're just trying to say "be nice to others"...

nebulon42:
I feel this to be unnecessary, if there is proper governance of a project this should be self-evident.

rory:
@kocio-pl @nebulon42 One (big) goal of a modern CoC is detail and unambiguity. Just having "Be nice to everyone" is akin to not having a CoC at all.

I am happy with the Go change of just adding  "ethnicity". I merely added "national origin" because it's in EU anti-discrimination law.

math1985:
> I feel this to be unnecessary, if there is proper governance of a project this should be self-evident.

+1

math1985:
If we must have one, let's not reinvent the wheel and keep following an existing one. The discussion whether 'national origin' does or does not equal race/ethnicity should really be out of the scope of this project.

kocio-pl:
Replaced by #2503.

Issue #2198, opened 2016-06-25
Plans for supporting Mapnik 2


matkoniecz:
Currently this style works both in Mapnik 2 and Mapnik 3.

But some rules that are available in Mapnik 3 are not present in Mapnik 2. Using them would cause Mapnik 2 renderers to fail.

How long we want to support Mapnik 2? Most likely at least untill https://github.com/openstreetmap/chef/issues/39 is completed.

Note that stopping to support Mapnik 2 means that using current TileMill version to render the style will be impossible. Given that TileMill is officially abandoned by Mapbox it would require to either drop TileMill support or to fork TileMill to release a new version.

Kosmtik would continue to work fine.

At least some rules were renamed - for example to achieve on points/areas effect of `text-min-distance` in 2.3 it is necessary to use `text-margin` in 3.0. As result it is impossible to support 2.3, 3.0 and use `text-min-distance`/`text-margin` effect at once.

gravitystorm:
If the stylesheets can't be used by tilemill then we decrease the pool of contributors and decrease the opportunities to reuse the styles. Tilemill is still the only CartoCSS design tool that has proper releases and windows support. Kosmtik is a source-only install and doesn't have any windows builds.

I'd love to see a mapnik 3 version of tilemill, but I don't know if that's likely to happen. Does anyone know if there are any forks or non-mapbox versions in development?

If we assume Tilemill isn't an option, what will we suggest to developers and end-users that they use instead? Kosmtik is fine if you don't mind command-line parameters and doing everything in separate text editors but it's a huge step back for those who enjoy various features of Tilemill.

math1985:
> I'd love to see a mapnik 3 version of tilemill, but I don't know if that's likely to happen. Does anyone know if there are any forks or non-mapbox versions in development?

As far as I know, Tilemill has been supporting Mapnik 3 already since October 2014. See this change: https://github.com/mapbox/tilemill/commit/818826c3f32162dd4a337f9fcb1c8ca60d29ba1b

matkoniecz:
> doing everything in separate text editors

I wanted to note that it is anyway necessary with mapnik - but from what I see it is also fixed in an unreleased tilemill (https://github.com/mapbox/tilemill/issues/190)

> tagging a new release and perhaps compiling binaries.

And making ppa with packages. There is no tilemill package for Ubuntu 16 - and MapBox is not going to make one. So it makes TileMill source-only install.

[...]

PR #2383, opened 2016-09-28
Travis carto update

nebulon42:
update carto version used for Travis and target Mapnik 2.3.0 API version specifically

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,10 +9,10 @@ addons:
 node_js:
   - "0.10"
 install:
-  - npm install carto@0.12.1
+  - npm install carto@0.16.3
   - mkdir -p data/world_boundaries data/simplified-land-polygons-complete-3857 data/ne_110m_admin_0_boundary_lines_land data/ne_10m_populated_places data/land-polygons-split-3857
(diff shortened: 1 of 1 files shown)


yohanboniface:
Upgrading carto makes sense, but not sure why targetting 2.3.0 specifically?

nebulon42:
Normally I would use whatever is running on the tile servers. Which is 3.0.11? But because of https://github.com/gravitystorm/openstreetmap-carto/pull/2374#issue-179183022 I thought the maintainers wanted to keep Mapnik 2 compatibility.

yohanboniface:
ah, sure makes sense!

kocio-pl:
You mean Mapnik probably. [Carto version on the servers is currently 0.16.3](https://github.com/openstreetmap/chef/issues/39#issuecomment-249408667) - the same as in this patch.

pnorman:
We still have to maintain 2.1 compatibility and older Carto versions too. If we want to test more recent versions we should do it in addition.

nebulon42:
I don't remember why, could you remind me?

kocio-pl:
I guess this is https://github.com/gravitystorm/openstreetmap-carto/issues/2198.

nebulon42:
Ah, so it's about Tilemill again. I think this software will haunt us forever. :)

kocio-pl:
It's rather our policy problem, since git version is already OK, but @gravitystorm [doesn't feel it's enough](https://github.com/gravitystorm/openstreetmap-carto/issues/2080#issuecomment-249803296) and I don't know if TileMill team would make a release just to satisfy this requirement.

pnorman:
> We still have to maintain 2.1 compatibility and older Carto versions too. If we want to test more recent versions we should do it in addition.

Issue #2094, opened 2016-03-23
Render transportation areas like the airports


kocio-pl:
I think we should start rendering big transportation "stations" as we do with airports. I mentioned it [here](https://github.com/gravitystorm/openstreetmap-carto/pull/1673#issuecomment-125681252) and now I want to really do it.

For a start it would be good to render these areas as we do with the airports (number of tagged lines from TagInfo):
- amenity=bus_station - 11 243
- railway=station - 7 906
- public_transport=station - 5 319
- aerialway=station - 2 084
- amenity=ferry_terminal - 1 027

For comparison: 
- aeroway=aerodrome - 9 844
- aeroway=terminal - 5 551

`@aerodrome` color should be renamed then (`@transportation-area` maybe?).

The numbers are rough, since the lines may also be used for buildings. So the next step should be rendering such buildings as we do with terminals on the airports (`@building-aeroway-fill` and `@building-aeroway-line` colors, which should rather be `@building-transportation-fill` and `@building-transportation-line`).

Probably some areas should be corrected, like this terminal:

https://www.openstreetmap.org/way/131450912

because it is partially drawn on the water.

jojo4u:
None of the tags with checkboxes are meant to be tagged mostly on buildings, they define the whole "transportation area" like aerodrome.

kocio-pl:
Don't worry, I try to unify all the problems related to transport "stations"!

HolgerJeromin:
Heliport has now 153 elements in the database. Probably because it is not rendered

kocio-pl:
I'm glad you have noticed. I believe there are also not too much such objects in the world ([list of heliports in Wikipedia](https://en.wikipedia.org/wiki/List_of_countries_by_number_of_heliports) contains also helipads, so it's hard to measure it precisely). 

It's not my priority to show that sometimes numbers should not be limiting factor and that system coherence is important (see 3. in https://github.com/gravitystorm/openstreetmap-carto/issues/3305#issue-341356089), but that's still something I would like to show, since it's the most similar to the standard airport. We have even the proper icon for it:

https://github.com/gmgeo/osmic/blob/master/transport/heliport-14.svg

jeisenbe:
Since aeroway=heliport is the only remaining feature on the list, it would make sense to close this issue and open a new one with the title "Add rendering for aeroway=heliport"

jeisenbe:
[...]

PR #2385, opened 2016-09-29
Adding rendering for amenity=ferry_terminal

kocio-pl:
Resolves https://github.com/gravitystorm/openstreetmap-carto/issues/869.

Looks like airport area rendering is rather for large area and ferry terminal is more like a railway station, so I took it. Labels are however the same like in the airport, but rendered later.

I also made some cleaning in color variables.

Examples:

Warsaw, z15 (nodes only, compared to some other labels)
p7dbtip8

Victoria (nodes and typical area)
z14 (not rendered)
bjher_1n
z15
pkcactt9
z16
x5e7i lb
z17
g0l1qrzl
z18
cc36kprk
z19
faamqado


kocio-pl:
Another possibility - labels the same as the bus stop - doesn't work for me:
cxyhw1qv

BTW: are there really so large railway stations which should be rendered from z10? If not, maybe we can adjust SQL to optimize query for some higher zoom levels?

pnorman:
Can review when I get home.

kocio-pl:
As the station color seems to be designed for buildings, it can be made lighter (here or in separate PR). We could also go with airport area color, but it's too pale for this usage IMO:

z17
gj9efxea

z18
q1snoghk

z19
qs l225q

math1985:
@kocio-pl Could you rebase this? I have the impression something went wrong, looking at the diff (or I'm misunderstanding the diff algorithm again).

kocio-pl:
After some time away from this PR I've changed my mind and like to use the lighter version for a background, because current station color seems to be designed with station building in mind, not for the area (merging https://github.com/gravitystorm/openstreetmap-carto/pull/2443 will have some implications there).

I don't need to find it from a distance (ferry lines are good hint), it's enough to see its shape to understand how the space is used. Also using the same color for airport, ferry and station area (and possibly other transportation areas) would be more consistent.

I will fix the code soon.

kocio-pl:
The code is updated and ready to be merged. 

General notes:
- The area is now rendered as the aerodrome (not as the railway station), the rest remains the same
- There was variable sections cleaning done

nebulon42:
I think it is not a good thing to mix cleanups with changes. It makes diffs harder to read. Since there were some changes I don't get the changes at first glance. Could you please show a proper before after comparison?

pnorman:
I think it's more worthwile to render an icon for a ferry terminal than a fill. These are features that people are often looking for, and an icon is better for that.

kocio-pl:
The fill and the icon play a different role, just like in the airport: icon make it easier to find on the big picture, while the fill is important when zoomed in to know which roads and buildings belong here. This is related to https://github.com/gravitystorm/openstreetmap-carto/issues/2094.

Unfortunately we have no such icon available, because none of my propositions was approved by the community (https://github.com/gravitystorm/openstreetmap-carto/pull/1673), but if we have one, I would like to show it too.

kocio-pl:
Code is cleaned now and here are basic before/after renderings (I won't make it extensive, because most of previous are correct except those with red background):

Warsaw, z15
Before:
e_ndw nu
After:
zvjyqtxs

Victoria, z18
Before:
4xbciwk5
After:
qw5 eoil

kocio-pl:
Code is updated now and ready to be merged.

math1985:
I agree this feature would need an icon for it to be useful (sorry for slow review).

kocio-pl:
Any hints about the icon design? Maybe the discussion in the https://github.com/gravitystorm/openstreetmap-carto/pull/1673 could be useful here.

I think the icon should be visible only to some zoom level (like the airport, but later, since typical ferry terminal is smaller). It wouldn't be visible on z18 then, so the rendering above would be the same.

kocio-pl:
This is how original icon works in reality (but it's black here):
https://maps.wikimedia.org/#18/52.23707/21.03863

I also think we could try with something more symbolic and easier to recognize, like the anchor:
https://openclipart.org/detail/10145/aussersihl-coat-of-arms

What do you think?

math1985:
To me, the anchor doesn't really remind me of a ferry terminal. Not sure what others think.

math1985:
To me, the anchor doesn't really remind me of a ferry terminal. Not sure what others think.

kocio-pl:
I would be happy to at least have some hint what could work for the others, since my ideas was not accepted till now and I ran out of them. However first anchor is still good for me.

dieterdreist:
2017-03-22 13:47 GMT+01:00 kocio-pl <notifications@github.com>:

> I would be happy to at least have some hint what could work for the
> others, since my ideas was not accepted till now and I ran out of them.
> However first anchor is still good for me.
>


IMHO it could be a car on a ship or inside a ship. An anchor is far too
generic.

math1985:
Not all ferries carry cars. Maybe a generic icon of a ship?

dieterdreist:
2017-03-22 15:24 GMT+01:00 Matthijs Melissen <notifications@github.com>:

> Not all ferries carry cars. Maybe a generic icon of a ship?




not all ships are ferries ;-)

After a quick look I think that current ferry tagging is immature, it won't
allow us to distinguish between huge ferry terminals and small passenger
ferry stops on a river like here:
http://www.openstreetmap.org/node/1051302891
Also the term "terminal" is indicating an endpoint, for intermediate stops
it isn't right.

there's also a new proposal for these kind of features (currently not used
for this style):
https://wiki.openstreetmap.org/wiki/Proposed_features/Seaway

pnorman:

math1985:
@kocio-pl What is the status of this PR?

kocio-pl:
I took some time to let the idea establish a bit in my head. Now I plan to eventually update it with 1. anchor icon, because:
- it's generic enough and much easier to recognize than our previous attempts to depict ships,
- having icon here conforms our style for transport amenities (blue icons for bus/tram stops or the airports) and having consistent system is enough to give user a hint what is the context of the anchor (transport),
- it's probably not going to be used for any other water-related feature (like marina).

kocio-pl:
@dieterdreist I think that current tagging is just generic, I wouldn't call it immature - this problem applies to many old features (like healthcare amenities or public transport). However your `seaway=*` proposition looks quite interesting to me, but it's not yet practically used (6 uses now) nor even proposed for discussion (it's still a draft), so we can re-examine it later.

kocio-pl:
Current rendering (with icon):

Warsaw, z15
Before
sfhncvqu
After
yi8iztue

Sidney (USA)
z15
Before
rz_0nkmo
After
vo2mmc f


z18
Before
h0wm1cah
After
3szsjaex

kocio-pl:
Are there any other issues not discussed before?

pnorman:
It's worth checking an inner harbor dense with small ferry terminals. http://www.openstreetmap.org/#map=15/49.2726/-123.1264 has some

kocio-pl:
Vancouver, z15  (click to see unscaled images):
jzcaxgp5

kocio-pl:
I feel this code is ready to be merged, but I don't know how long should I wait and what for, because it's quite old, but formal requirement (using icon) is fulfilled for just a few days. On the other hand nobody was opposing. Our decision rules are hard to apply to long and evolving solutions.

vadp:
I've noticed that Mapnik started rendering amenity=ferry_terminal with an anchor symbol anchor symbol. 

At nautical maps a similar symbol denotes an [anchorage](http://wiki.openstreetmap.org/wiki/Seamarks/INT-1_Section_N#Anchorages.2C_Anchorage_Areas), so this usage of an anchor for the amenity=ferry_terminal is quite misleading.

kocio-pl:
It's not a nautical map and none of the ship icon was clear enough for osm-carto community (even if they are used on some other styles). I also noticed using the anchor symbol in some other general map, but unfortunately I don't remember which one.

imagico:
@vadp - you should best open a new issue for this.

Not only in nautical maps by the way, see for example chapter 2, symbol 226 in http://www.lib.berkeley.edu/EART/pdf/soviet.pdf - likewise for some pre-digital German TK100/TK200 (don't have a digitized sample at the moment).

dieterdreist:
I agree an anchor is not a good symbol for a ferry

vadp:
> I agree an anchor is not a good symbol for a ferry

@dieterdreist I've opened #2785 for this issue.

Issue #2227, opened 2016-07-13
Icon for shop=coffee


Jotam:
According to taginfo, shop=coffee is getting closer to 2000 instances. I would like to suggest to use this icon for rendering such shops: https://wiki.openstreetmap.org/wiki/File:Shop_coffee2.svg

HaskuldNihil:
Lol?

Jotam:
Would you also have a more precise and qualified comment on this?

kocio-pl:
Rendering with and without label (Warsaw, z19), code is [here](https://github.com/kocio-pl/openstreetmap-carto/tree/coffee):
![6ndfg8k](https://cloud.githubusercontent.com/assets/5439713/17075164/5a5c2510-508b-11e6-9513-7c13395f40c8.png)
![eaxervrh](https://cloud.githubusercontent.com/assets/5439713/17075165/64d5cae6-508b-11e6-9146-d02efcb1b5e8.png)

boothym:
Looks a bit like a rugby ball, or a set of lips - not sure I'd say that was a shop selling coffee at first glance. Maybe something more like http://www.freeiconspng.com/uploads/coffee-bean-icon-9.png or http://www.freeiconspng.com/uploads/coffee-bean-icon-png-7.png ?

kocio-pl:
Thanks, I may try to use them, but SVG examples would be more suitable for testing.

Jotam:
The wavy line is better than my more or less straight version. What I don't like is that the opening cuts through the whole bean and sort of "opens" it up.

kocio-pl:
@Jotam Do you have a fresh idea now? Maybe some coffee could help. :smile:

Jotam:
Let's try this:
2) ![layer1](https://cloud.githubusercontent.com/assets/15984044/18836335/d616aee4-83ff-11e6-9806-e2539480868b.png) - SVG[](http://vdma.aldebaran.uberspace.de/osm/coffee_02.svg)

Jotam:
@kocio-pl: Did you try this?

kocio-pl:
Works for me:
![78ebfvb5](https://cloud.githubusercontent.com/assets/5439713/18884207/b18b4052-84e6-11e6-9624-b60274a3214c.png)

Jotam:
For me too.

PR #2386, opened 2016-09-30
Adding rendering for shop=coffee

kocio-pl:
Resolves https://github.com/gravitystorm/openstreetmap-carto/issues/2227.

The number of uses is currently 1986, so very close to a safe 2k limit.

78ebfvb5


micahcochran:
shop=coffee is now at 2070 uses.

math1985:
@kocio-pl Could you rebase this?

kocio-pl:
Done.

nebulon42:
The icon is ok. But I'm against merging this until the shop discussion has progressed further.

kocio-pl:
Depends on https://github.com/gravitystorm/openstreetmap-carto/issues/2444 then.

kocio-pl:
There's no progress in the shop discussion, so I'm not sure what to do - I'm ready to rebase it, but it's tedious for me as a casual git user, so it makes sense only if merging probability is high.

math1985:
I think we need to give the shop discussion some more time.

math1985:
Yes, good idea.

kocio-pl:
Done. I hope that SVG code is proper.

kocio-pl:
Is there any problem left to solve here?

nebulon42:
Sorry for the long delay. I would merge it right away, but I saw that you are using a transform to rotate the bean. If you could clean this up (remove the transform) this would be great. If this is too much work I would also merge it as is.

kocio-pl:
I have used file from @Jotam and I don't know how to fix this problem. If you or him (or anybody else) could fix the icon, I would simply update the code.

nebulon42:
Turned out to be quite easy. Can you check for the last time? Then I will merge this.

kocio-pl:
Ready. I've checked and it works as expected. I made it to be just one commit.

Issue #528, opened 2014-05-17
Render barriers above highway areas


matthijsmelissen:
Barriers should be rendered above highway areas.

If not, barriers between highway areas become invisible.

Examples:
http://www.openstreetmap.org/#map=18/52.53296/13.37229

See also https://trac.openstreetmap.org/ticket/2807.

pmailkeey:
This looks like a case of the common breach of common sense that states a rendering order of :
largest areas first to smallest areas;
All lines on top of areas
shorter lines on top of longer lines
points over everything.

As otherwise data will be hidden.

The question is, when will this be implemented ?

kocio-pl:
What is current status of this issue and how difficult it would be to make it work?

kocio-pl:
We have 79 layers, so it's not trivial to know where to start, even though they have clear names. I believe important ones for this issue are (in order of appearance): 
```yaml
  - id: "line-barriers"
  - id: "area-barriers"
  - id: "highway-area-fill"
  - id: "roads-fill"
  - id: "turning-circle-fill"
```
My questions:
1. What is the meaning of layer definitions order - which ones are rendered above?
2. Is it important to have "-" before "id:" in YAML? Some lines miss it.

Wild guess: probably just moving -barriers (probably without touching water-barriers-) after -fill could be the solution?

Hint - current list of all `osm-carto` layers:
```yaml
  - id: "world"
  - id: "coast-poly"
  - id: "builtup"
  - id: "necountries"
  - id: "landcover-low-zoom"
  - id: "landcover"
  - id: "landcover-line"
  - id: "water-lines-casing"
  - id: "water-lines-low-zoom"
  - id: "icesheet-poly"
  - id: "water-areas"
  - id: "landcover-area-symbols"
  - id: "icesheet-outlines"
  - id: "water-lines"
  - id: "water-barriers-line"
  - id: "water-barriers-poly"
  - id: "marinas-area"
  - id: "piers-poly"
  - id: "piers-line"
  - id: "water-barriers-point"
  - id: "bridge"
  - id: "buildings"
  - id: "buildings-major"
  - id: "tunnels"
  - id: "landuse-overlay"
  - id: "line-barriers"
  - id: "cliffs"
  - id: "area-barriers"
  - id: "ferry-routes"
  - id: "turning-circle-casing"
  - id: "highway-area-casing"
    id: "roads-casing"
  - id: "highway-area-fill"
  - id: "roads-fill"
[...]

PR #2394, opened 2016-10-02
Rendering barriers above highway areas

kocio-pl:
Resolves https://github.com/gravitystorm/openstreetmap-carto/issues/528.

The code is simple - I've just moved 2 barrier layers (line-barriers and area-barriers) after 3 filling layers (highway-area-fill, roads-fill and turning-circle-fill) and it works, but I don't know if there are side effects, so please report any problems.

Warsaw, z19:
Before
xbxuvdm6
After
uhzzvchq

### DIFF
--- a/project.mml
+++ b/project.mml
@@ -633,32 +633,6 @@
       "advanced": {}
     },
     {
-      "name": "line-barriers",
-      "srs-name": "900913",
-      "geometry": "linestring",
-      "class": "barriers",
-      "id": "line-barriers",
-      "srs": "+proj=merc +a=6378137 +b=6378137 +lat_ts=0.0 +lon_0=0.0 +x_0=0.0 +y_0=0.0 +k=1.0 +units=m +nadgrids=@null +wktext +no_defs +over",
-      "Datasource": {
-        "extent": "-20037508,-20037508,20037508,20037508",
-        "table": "(SELECT\n    way, COALESCE(historic, barrier) AS feature\n  FROM (SELECT way,\n    ('barrier_' || (CASE WHEN barrier IN ('chain', 'city_wall', 'embankment', 'ditch', 'fence', 'guard_rail',\n          'handrail', 'hedge', 'kerb', 'retaining_wall', 'wall') THEN barrier ELSE NULL END)) AS barrier,\n    ('historic_' || (CASE WHEN historic = 'citywalls' THEN historic ELSE NULL END)) AS historic\n    FROM planet_osm_line\n    WHERE barrier IN ('chain', 'city_wall', 'embankment', 'ditch', 'fence', 'guard_rail',\n          'handrail', 'hedge', 'kerb', 'retaining_wall', 'wall')\n      OR historic = 'city_wall'\n      AND (waterway IS NULL OR waterway NOT IN ('river', 'canal', 'derelict_canal', 'stream', 'drain', 'ditch', 'wadi'))\n  ) AS features\n) AS line_barriers",
-        "geometry_field": "way",
-        "type": "postgis",
-        "key_field": "",
-        "dbname": "gis"
-      },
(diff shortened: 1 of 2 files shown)


pnorman:
We intend to merge highway areas into linear highway fill at some point - how will this interact?

kocio-pl:
I don't know. You mean the lua branch?

pnorman:
No, #2046 and the related incomplete PR

kocio-pl:
Once https://github.com/gravitystorm/openstreetmap-carto/pull/2058 is more or less in its releasable shape, I'm happy to test it. But since it looks like you don't have a time to resume this work, I think we can merge this one (if there are no other problems) and look again when the time will come.

math1985:
I am a bit concerned that this PR would mean that hedges etc will cover (and perhaps sometimes completely hide) roads. That might be worse than the bug this is trying to solve?

math1985:
I am a bit concerned that this PR would mean that hedges etc will cover (and perhaps sometimes completely hide) roads. That might be worse than the bug this is trying to solve?

kocio-pl:
I will check it.

kocio-pl:
We already show hedges along the roads and I see no such problem. It does not hurt even pedestrian and bicycle ways (which are thinner than roads), for example:

https://www.openstreetmap.org/way/370601787

Let me know if that's what you were asking me, because it seems it has nothing to do with highway areas at all. If it's OK for you, I will fix the code to make it ready to be merged.

math1985:
An example of what I mean:
screenshot from 2016-11-19 00 02 21

kocio-pl:
Now I see... It's hard for me to judge, both problems look pretty important for me. It would be good to have at least the estimation how often they would appear, but I don't know how to measure it.

nebulon42:
I think this is undesirable and should be solved by splitting up barriers in things that could show up over roads and ones that shouldn't. Hedges are things that shouldn't show up over roads. Walls could.

kocio-pl:
If there will be no objection, I will try to do the split.

math1985:
I'm not sure whether walls (and citywalls even more so) wouldn't cause the same problem.

kocio-pl:
Should we find some examples to test it or we just want to avoid it anyway?

math1985:
Testing in real-life examples never hurts but I'm quite skeptical.

kocio-pl:
So I'm closing it now. The change is rather simple, so we could easily recreate it once we need it.

Tomasz-W:
I can see here just one example of barrier hiding the road, and I can give you a lot more of examples, when barrier is not visible, and it could make map user wrong:

1. barriers in the middle of highway area:
https://www.openstreetmap.org/way/261727797
http://www.openstreetmap.org/way/289976085
2. guard rails or retaining walls for people in wheelchairs
https://www.openstreetmap.org/way/498100708
http://www.openstreetmap.org/way/436542403
https://www.openstreetmap.org/way/343486650
3. fences and walls crossing highway areas:
https://www.openstreetmap.org/way/344084296
https://www.openstreetmap.org/way/440398217
https://www.openstreetmap.org/way/388928461
https://www.openstreetmap.org/way/313571600
https://www.openstreetmap.org/way/441567205
https://www.openstreetmap.org/way/359999536
https://www.openstreetmap.org/way/438594618

I think the situation of barrier hiding the road is a lot lot rarer, than situation when lack of barrier line could make someone walk/ ride in wrong direction

Tomasz-W:
I think it haven't been tested enough (just one example of barrier hiding the road!) and should be reopen.

kocio-pl:
Due to technical limitations on GitHub it can't be reopen, but the change is simple to do - see https://github.com/gravitystorm/openstreetmap-carto/pull/2394#issue-87651311. Is there anyone willing to test it?

Issue #2045, opened 2016-02-16
Convert patterns to SVG


pnorman:
This will require Mapnik 3, but we want to move these PNGs to vectors for better high-resolution support.

kocio-pl:
We could start doing it now, as the software update on servers is ready, but that would break all the other systems still using Mapnik 2. We need a roadmap first - see https://github.com/gravitystorm/openstreetmap-carto/issues/2198.

kocio-pl:
I think of this ticket as meta-issue, because each pattern requires separate PR probably. I'm not sure how to count them, because some of them are combined, but let's start with this list:
- [ ] bare_ground
- [ ] bog
- [ ] forest
- [ ] mangrove
- [ ] marsh
- [ ] quarry
- [ ] reed
- [ ] rock
- [ ] scree
- [ ] scrub
- [ ] swamp
- [ ] wetland

HolgerJeromin:
If you add the checkbox list to the first post the "ready count" will be displayed on many places here.

kocio-pl:
I can do it if Paul has noting against it, but first I wanted to know if I get this list right.

imagico:
In my eyes the line patterns (cliff/embankment) are much more important to move the SVG since they will actually profit from this in terms of normal rendering results.

For the area patterns i would before all put

- [ ] check if there are any serious performance implications of moving to SVG patterns, especially if the SVGs are fairly large

pnorman:
> check if there are any serious performance implications of moving to SVG patterns, especially if the SVGs are fairly large

Mapnik converts SVGs to images as a one time operation for icons and caches the result. I expect it does the same for SVGs as patterns.

kocio-pl:
BTW: do we have any other line patterns except those two (cliff and embankment)? We should add all of them to the checklist and check/strike them later.

Ircama:
As I'm developing the new PR for rendering [atetes and ridges](https://github.com/gravitystorm/openstreetmap-carto/pull/2138#issuecomment-260118484) (sorry for the delay BTW, I have limited spare time), I'd like to know whether I can keep the already available [PNG icons](https://github.com/gravitystorm/openstreetmap-carto/pull/2138/files#diff-5b03261046c36e28b66a9de46ae57259) or I necessarily need to convert them to SVG, thanks.

imagico:
[...]

PR #2395, opened 2016-10-03
Random scrub pattern

### LINKED ISSUE #2253: Use randomizes pattern for scrub
It would be nice to have a randomized pattern (like for forests) also for scrub.

meased:
Randomized dot pattern for scrub using http://www.imagico.de/map/jsdotpattern.php. Background color is unchanged, however foreground icon color is slightly desaturated. It looked too green to me so I tried to follow the style of natural=wood.

Fixes #2253.
### Before ... after

beforenafter18
beforenafter15
beforenafter14

As the old scrub icon is png only and jsdotpattern requires svg, I created a nearly identical svg version of the scrub icon and included it in this PR, although it's not used other than for the pattern generation.

For reference, jsdotpattern settings were:
- regular snub square
- distance = 45
- radius = 64
- radius y = 64
- about 10+ relax clicks


kocio-pl:
Nice result and nice color too!

Please move the files to [symbols/generating_patterns](https://github.com/gravitystorm/openstreetmap-carto/tree/master/symbols/generating_patterns) directory and add basic documentation about how to generate it to the file named like `scrub.md`.

imagico:
Pattern looks good, the symbol color is too grayish i think - it does not have to be as strong as with the old pattern but it should be at least the same chroma as the base color.

meased:
The color was derived by starting with scrub color and then dialing down the lightness and saturation. I did go pretty far with the saturation, as it seems that natural=wood is rather grayish to me also. I'll make the documentation changes when I get back to a computer, and also experiment with some saturation changes.

imagico:
You should do color adjustments in a perceptually uniform color space

Good point to start:

http://davidjohnstone.net/pages/lch-lab-colour-gradient-picker#b6e3b5,a1c8a1,9dc99d

meased:
What do you think of #7ebd7f?
greener
Compare to the original example color #94b995
scruborig

imagico:
Yes, better - maybe even somewhat strong so it could be lightened a bit.

The pattern by the way should probably go into the landcover-area-symbols layer so it drawn above other landcovers like for wood/forest.

meased:
Symbol definition has been moved to landcover_area_symbols, scrub.md has been added, changed icon color to #8ece8f (about 5% lighter):
scrublighter

imagico:
By the way i added the scrub symbol used here as `scrub2` to jsdotpattern and converted it to a filled polygon instead of strokes making it more in line with other SVGs used here - you might want to change this accordingly.

meased:
Thanks, that will make it a lot easier to reproduce the images. I was having problems getting the original scrub icon on jsdotpattern pixel aligned, which is why I didn't use it here.

sommerluk:
It might be a good idea to requiere Mapnik3 #2401 and make this PR directly with a SVG pattern instead of a PNG pattern…

kocio-pl:
We have a ticket about it (https://github.com/gravitystorm/openstreetmap-carto/issues/2045) and it's related to https://github.com/gravitystorm/openstreetmap-carto/issues/2401 and https://github.com/gravitystorm/openstreetmap-carto/issues/2198.

kocio-pl:
If there are no objections, I would try to merge this PR in the current form. SVG patterns have to wait until we decide what to do with Mapnik 2.

math1985:
👍

imagico:
To be in line with the other patterns the SVG should be updated to use fill based rather than stroke based geometries.  I was going to evaluate the pattern color intensity in comparison to other map elements but did not yet have time to do so.

kocio-pl:
I leave the it to you then (I mean code update and merging).

The only important thing for me is to not wait forever with a useful change, so could you tell how much time do you think you need in reality? I just would like to take over if this safe "deadline" is crossed and nobody knows what is the current status.

imagico:
Just relax - a few days more will not hurt anyone.

Besides - it would be good if @meased could update the SVG as noted.

meased:
Sure thing. I should be able to get that done tonight or tomorrow.

kocio-pl:
@imagico It's more about knowing that there's something stalled than doing anything fast. 

There's no absolute measure of things getting stalled, but while time-based checkpoints are only relative (some things may need hours, while others weeks and months), they are super useful to detect there are any kind of problems that may need attention.

meased:
Documentation and SVG file updated. I believe everything is in order.

imagico:
Looks good, nice work with the pattern and color, very well balanced.

Thanks.

sommerluk:
Thanks a lot!

kocio-pl:
A bit off topic - interesting combination of scrub and brownfield colors:

http://www.openstreetmap.org/way/24050682

mboeringa:
> A bit off topic - interesting combination of scrub and brownfield colors:
> 
> http://www.openstreetmap.org/way/24050682

That is largely a consequence of this remark by @imagico : 

> The pattern by the way should probably go into the landcover-area-symbols layer so it drawn above other landcovers like for wood/forest.

While for the particular situation you pointed out (scrub & brownfield), the implemented layering may make some sense (after all, there may be a patch of scrub on a brownfield), I am less convinced this behaviour is desired in the particular situation @imagico pointed out: something is either a forest/woodland, or its scrub, not both IMO. Unfortunately, with this current behaviour, people are likely to map both in the same place, which I don't think is a good idea and will lead to ambiguous tagging and symbolization issues:  what to do with the overlapping areas?

Not all data consumers will have the ability or wish to deal with to much of this type of overlapping related features / landcovers.

PR #2396, opened 2016-10-03
Update INSTALL.md for fonts not included

littlebtc:
The Noto CJK JP font seems to be missing in the Standard layer, it took me some time to figure out that it could be the problem in the installation guide.

Noto CJK fonts are provided in another `fonts-noto-cjk` package in Debian / Ubuntu instead of `fonts-noto`. The guide should be updated.

Note: `fonts-noto-cjk` is not available in Debian Jessie. :'(

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -53,7 +53,7 @@ Mapnik 3 is required for acceptable rendering of most non-Latin scripts, particu
 On Ubuntu 16.04 or Debian Testing you can install all required fonts with
 
 ```
-sudo apt-get install fonts-noto ttf-unifont
+sudo apt-get install fonts-noto fonts-noto-cjk ttf-unifont
 ```
 
 DejaVu is packaged as `fonts-dejavu-core`.


math1985:
Could anybody verify this?

kocio-pl:
You can look here for Ubuntu Xenial:
http://packages.ubuntu.com/xenial-updates/all/fonts-noto-cjk/filelist

It's probably available for Debian too, see this page:
https://packages.debian.org/jessie-backports/fonts-noto-cjk

pnorman:
On Ubuntu 14.04 fonts-noto depends on fonts-noto-cjk. Is this not true for 16.04?

littlebtc:
Sorry I am wrong. [Debian Package](https://packages.debian.org/en/sid/fonts-noto)`fonts-noto` *recommends*  `fonts-noto-cjk` and by default apt-get will install recommended package.

So I close this pull request. But why the standard layer now uses the Unifont fallback in most of the Asia? Can someone find out a good reason? [My OSM Diary](http://www.openstreetmap.org/user/Littlebtc/diary/39585)

pnorman:
http://packages.ubuntu.com/xenial/fonts-noto is the relevant Ubuntu link, since Ubuntu is the primary target, not Debiand. It looks like they switched the metapackage from depending on the sets of noto fonts to recommending them in 16.04, so we need to require the -cjk and -unhinted packages. I think we should do -hinted, -unhinted, and -cjk explicitly and not reference plain fonts-noto.

HolgerJeromin:
Many (especially server) admins do not install recommended packages with apt. So we should really have a list of needed packages.

PR #2400, opened 2016-10-03
Add Map Icon Guidelines to Contribution Guidelines

micahcochran:
Give brief guidelines for map icon submission.  Mainly points contributors to back to OSM wiki for more extensive information.

In my opinion, the repository is currently a little sparse in regard to this kind of information.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -142,13 +142,7 @@ Because SQL within JSON or YAML will not generally be syntax highlighted, indent
 
 * All new icons must be SVG format only.  The SVG must be saved as standards compliant SVG without any proprietary tags. In Inkscape software, you will need to "Save As..." and choose the format Optimized SVG (preferable) or Plain SVG.
 * Icons must use SVG fills only, not SVG strokes or any feature Mapnik does not support.
-* Use no color for the icon's fill.  MSS (Map Style Sheets) gives the icon the appropriate thematic color.  As an example the [parking symbol in parking.svg](https://github.com/gravitystorm/openstreetmap-carto/blob/master/symbols/parking.svg) has no color defined (XML code):
-```XML
-<?xml version="1.0" encoding="utf-8"?>
-<svg xmlns="http://www.w3.org/2000/svg" width="100%" height="100%" viewBox="0 0 9 12">
-<path fill-rule="evenodd" d="M 0,0 L 0,12 1.75,12 1.75,7 6,7 A 3.0,3.5 0 0,0 6,0 Z M 1.75,1.75 L 1.75,5.25 5.5,5.25 A 1.58,1.75 0 0,0 5.5,1.75 Z"/>
-</svg>
-```
+* Use no color for the icon's fill if the icon is monochromatic. This allows the color to be set in the MSS.
 * Use a common canvas size, which is usually 14x14 px.
 * Draw a simple siloutte of the subject with an "on the shelf" perspective.
 * Align vectors to the pixel grid.
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -140,19 +140,16 @@ Because SQL within JSON or YAML will not generally be syntax highlighted, indent
 
 ## Map Icon Guidelines
 
-Icons for the submitted to the standard tile layer will be:
+Design Guidelines
 * SVG only
-* flat (single colour [usually black], no gradients, no outlines)
-* clean (reduced complexity where possible)
-* sharp (aligned to pixel grid)
-* single point of view (avoid use of perspective where possible)
-* common canvas size (usually 14x14px)
+* flat meaning a single colour, no gradients, no outlines
+* common canvas size, which is usually 14x14 px
+* single point of view and avoid use of perspective where possible
 
-Read the [https://wiki.openstreetmap.org/wiki/Map_Icons/Map_Icons_Standards](Map Icon Standards)
-on how to approach designing clear icons.
+Best Practices
+* align to pixel grid
+* clean design, so reduced complexity where possible
 
-More information at [https://wiki.openstreetmap.org/wiki/Map_Icons](OSM wiki Map Icons)
-and information concerning icon color.
 
 ## Pull requests
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -140,9 +140,15 @@ Because SQL within JSON or YAML will not generally be syntax highlighted, indent
 
 ## Map Icon Guidelines
 
-* All new icons must be SVG format only.  The SVG must be saved as standards compliant SVG without any prioprietery tags.  In Inksapce software, you will need to Save As... Plain SVG format.
-* Icons must be use fills and not outlines.
-* Use a single color, typically black, for the fill portion of the icon.  MSS (Map Style Sheets) replaces the single color with the appropriate thematic color.
+* All new icons must be SVG format only.  The SVG must be saved as standards compliant SVG without any proprietary tags.  In Inksapce software, you will need to "Save As..." and choose the format Optimized SVG (preferable) or Plain SVG.
+* Icons must use SVG fills only, not SVG strokes or any feature Mapnik does not support.
+* Use no color for the icon's fill.  MSS (Map Style Sheets) gives the icon the appropriate thematic color.  As an example the [parking symbol in parking.svg](https://github.com/gravitystorm/openstreetmap-carto/blob/master/symbols/parking.svg) has no color defined (XML code):
+```XML
+<?xml version="1.0" encoding="utf-8"?>
+<svg xmlns="http://www.w3.org/2000/svg" width="100%" height="100%" viewBox="0 0 9 12">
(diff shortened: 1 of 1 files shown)


dieterdreist:
sent from a phone

> Il giorno 04 ott 2016, alle ore 02:46, Paul Norman <notifications@github.com> ha scritto:
> 
> What I learned is largely based on Mac icon design from the 90s


that's a good base regardless the age...

micahcochran:
I was just copying and pasting, so I'm fine someone else taking the reigns.  I want some kind of notes that eases the learning curve for contributing map icons.

meased:
The [Gnome HIG icon guidlines](https://developer.gnome.org/hig/stable/icons-and-artwork.html.en) with a few caveats seems to apply to this project (mainly the section on symbolic icons).

matkoniecz:
> I'm :-1: on linking to those wiki pages.

In addition the wiki page https://wiki.openstreetmap.org/wiki/Map_Icons/Map_Icons_Standards is simply wrong by claiming

> The icons are proposed and approved individually so it is very difficult to achieve a set of icons with good readability and consistency.

Nobody cares about whatever icon was "approved" on wiki, certainly it means nothing here (see https://wiki.openstreetmap.org/wiki/Map_Icons/Proposed_Icons - many icons suffer from problems making them unusable here).

micahcochran:
I've removed the wiki links, and made a few small changes based on the feedback.  

Should I include an image that gives color palette information?

CONTRIBUTING.md
```diff
@@ -138,6 +138,19 @@ Because SQL within JSON or YAML will not generally be syntax highlighted, indent
...
+* SVG only
+* flat meaning a single colour, no gradients, no outlines
+* common canvas size, which is usually 14x14 px
+* single point of view and avoid use of perspective where possible
+
+Best Practices
+* align to pixel grid
+* clean design, so reduced complexity where possible
+
+
```
pnorman: There's an extra newline here

CONTRIBUTING.md
```diff
@@ -138,6 +138,19 @@ Because SQL within JSON or YAML will not generally be syntax highlighted, indent
 * One space before and after = etc
 * Name SQL subqueries after the layer name (but use underscores)
 
+## Map Icon Guidelines
+
+Design Guidelines
```
pnorman: This should be a level 3 heading, or more appropriately, combine the two lists into one.

CONTRIBUTING.md
```diff
@@ -138,6 +138,19 @@ Because SQL within JSON or YAML will not generally be syntax highlighted, indent
...
+* SVG only
```
pnorman: Perhaps "All new icons must be SVGs." We still have some PNGs.

matkoniecz:
I would drop split into two sections - Best Practices heading seems unnecessary.

Maybe mention that it must work with other icons (it should not be possible to confuse icons for different things) and that icons should not and may not be added for everything?

It is a bit hard for me to provide feedback, given that I made exactly 1 icon that was accepted (and it was for man_made=cross, so it was quite hard to do it wrong).

Maybe @nebulon42 @kocio-pl @imagico can provide more useful comments?

CONTRIBUTING.md
```diff
@@ -138,6 +138,19 @@ Because SQL within JSON or YAML will not generally be syntax highlighted, indent
...
+* flat meaning a single colour, no gradients, no outlines
+* common canvas size, which is usually 14x14 px
+* single point of view and avoid use of perspective where possible
```
pnorman: No need to say "where possible", just "avoid use of perspective" is fine.

matkoniecz: Also, what "single point of view" means in context of making icons? This one is unclear for me what it means for graphics that are without perspective.

matkoniecz:
> Should we add something about Inkscape exporting?

One technical thing that was unclear for me was what should be removed from svg file as inkscape clutter.

matkoniecz:
It is a bit different - but maybe add suggestion to open issue for creating icon before designing one? This way obvious cases "there will be never special icon for that" may be handled without wasting time on designing icon.

matkoniecz:
> Although they are for different projects, the GNOME Icon guidelines and Maki Icon Design guidelines provide additional information about icon design.

I am not sure. Especially for soomebody new it is unclear what is a good general rule and what is project specific. For example in GNOME "Use the Tango color palette when designing full color application or device icons." is obviously not relevant - but it will not be obvious for somebody who is target of  CONTRIBUTING.md. And for example in Maki "No SVG elements except for paths." - I have no idea whatever it applies also to us.

matkoniecz:
@micahcochran Thanks for spotting that it was missing and making PR fixing it (I really should start from that)!

micahcochran:
@matkoniecz @pnorman  Thank you for the feedback.

I hope that this is going in the right direction in addressing the comments.  Please let me know if I have overlooked any comments.

CONTRIBUTING.md
```diff
@@ -140,16 +140,18 @@ Because SQL within JSON or YAML will not generally be syntax highlighted, indent
...
-* common canvas size, which is usually 14x14 px
-* single point of view and avoid use of perspective where possible
-
-Best Practices
-* align to pixel grid
-* clean design, so reduced complexity where possible
-
+* All new icons must be SVG format only.  The SVG must be saved as standards compliant SVG without any prioprietery tags.  In Inksapce software, you will need to Save As... Plain SVG format.
+* Icons must be use fills and not outlines.
+* Use a single color, typically black, for the fill portion of the icon.  MSS (Map Style Sheets) replaces the single color with the appropriate thematic color.
```
pnorman: Not black, but no colour at all. An example is the [parking symbol](https://github.com/gravitystorm/openstreetmap-carto/blob/master/symbols/parking.svg)

```svg
<?xml version="1.0" encoding="utf-8"?>
<svg xmlns="http://www.w3.org/2000/svg" width="100%" height="100%" viewBox="0 0 9 12">
<path fill-rule="evenodd" d="M 0,0 L 0,12 1.75,12 1.75,7 6,7 A 3.0,3.5 0 0,0 6,0 Z M 1.75,1.75 L 1.75,5.25 5.5,5.25 A 1.58,1.75 0 0,0 5.5,1.75 Z"/>
</svg>
```

The colour is unset, not an explicit `#000000`.

CONTRIBUTING.md
```diff
@@ -140,16 +140,18 @@ Because SQL within JSON or YAML will not generally be syntax highlighted, indent
...
-* flat meaning a single colour, no gradients, no outlines
-* common canvas size, which is usually 14x14 px
-* single point of view and avoid use of perspective where possible
-
-Best Practices
-* align to pixel grid
-* clean design, so reduced complexity where possible
-
+* All new icons must be SVG format only.  The SVG must be saved as standards compliant SVG without any prioprietery tags.  In Inksapce software, you will need to Save As... Plain SVG format.
+* Icons must be use fills and not outlines.
```
pnorman: It needs to be clearer that this is a technical requirement to use SVG fills, not SVG strokes. Perhaps

> * Icons must use SVG fills only, not SVG strokes or any feature Mapnik does not support.

As a design, the bicycle icon has outlines, but they are implemented as SVG fills.

micahcochran:
Do you have to use a text editor on the SVG file in order to strip out the unnecessary tags?

In Inkscape, it seems like the "Plain SVG" format helps and "Optimized SVG"  format gets closer to stripping out unneeded tags.

CONTRIBUTING.md
```diff
@@ -138,6 +138,27 @@ Because SQL within JSON or YAML will not generally be syntax highlighted, indent
 * One space before and after = etc
 * Name SQL subqueries after the layer name (but use underscores)
 
+## Map Icon Guidelines
+
+* All new icons must be SVG format only.  The SVG must be saved as standards compliant SVG without any proprietary tags.  In Inksapce software, you will need to "Save As..." and choose the format Optimized SVG (preferable) or Plain SVG.
```
kocio-pl: Inksapce -> Inkscape

pnorman:
> Do you have to use a text editor on the SVG file in order to strip out the unnecessary tags?

No, and most people don't.

CONTRIBUTING.md
```diff
@@ -140,16 +140,18 @@ Because SQL within JSON or YAML will not generally be syntax highlighted, indent
...
+* Icons must be use fills and not outlines.
+* Use a single color, typically black, for the fill portion of the icon.  MSS (Map Style Sheets) replaces the single color with the appropriate thematic color.
+* Use a common canvas size, which is usually 14x14 px.
+* Draw a simple siloutte of the subject with an "on the shelf" perspective.
+* Align vectors to the pixel grid.
+* Make a clean design, so reduced complexity where possible.
+
+### External Icon Design Resources
+These are resources with some helpful information about icon design.  These resources may have a different goal or philosophy and should not be held as authoritative sources for this project's icon design.
+* [Maki Icons Design Guidelines](https://www.mapbox.com/maki-icons/guidelines/) has some great illustrations of a well designed icon.
```
matkoniecz: I am not entirely sure, but maybe it should be "illustrations of a well designed icons" (last word in plural form)?

pnorman: Yes, that's correct, but the reason we're pointing at the guidelines is it has useful information on design, not examples of of icons.

micahcochran:
Thank you.

Issue #2378, opened 2016-09-27
New city labels sometimes overlap


imagico:
Like here (Freiburg im Breisgau/Basel):

![label overlap](http://www.imagico.de/files/label_overlap.png)

Also sometimes labels overlap/touch city dots (like Paris and Frankfurt above) or admin unit labels.

Combined with the large line spacing (#2358) this is quite irritating and difficult to read.

imagico:
Two more observations regarding the city labels - can open separate issues for them if desirable:

* the random default is probably a bad idea regarding the metatile edge issues - using a geometry hash would be better.
* prioritizing `N,S` or `S,N` in the label directions leads to frequent label pairs if two cities are close and on a similar latitude like Mary/Bayramaly [here](http://www.openstreetmap.org/#map=6/47.190/2.999) . An increased minimum distance might already solve this but alternatively a different order of direction priorities could also be better.

nebulon42:
I think the overlap problem might have been introduced by https://github.com/gravitystorm/openstreetmap-carto/commit/e4cb767fd5b2ae038ff5caef92332eb9b5a9755d? There the margin was removed but not replaced with some other distance measure.

The spacing issues was not present when I redesigned the labels, so I cannot really comment on that.

Could you expand on the geometry hash topic?

I didn't realise that this change is already live. Otherwise I would have suggested to delay the update. But I will try to look into a possible fix for the overlap problems. As I'm quite busy currently I cannot promise that it will be fast.

imagico:
> The spacing issues was not present when I redesigned the labels, so I cannot really comment on that.

Yes, it is so obvious it would have been visible in the examples.

> Could you expand on the geometry hash topic?

If you derive your `dir` value from [ST_GeoHash()](http://postgis.net/docs/ST_GeoHash.html) or simply using some minor digit from the coordinates you get the same for the same place in every metatile.  This way at least when labels do not compete with others you get consistent rendering at metatile edges.

nebulon42:
Great idea! That might likely solve the
````
     on
    x
Lond
````
issue. Since we cannot/should not use `avoid-edges`.

kocio-pl:
I was thinking about another solution - for small zoom levels with so few elements to render we don't need slicing the world to metatiles, we can check things global - I mean something like z4 or z5. Does this make sense and is technically possible?

StyXman:
[...]

PR #2404, opened 2016-10-09
base placename direction parameter on geometry

nebulon42:
Which makes it consistent across metatiles. Idea by @imagico. Ref #2378.

I have not seen any visual changes but at least we now can rule out this factor as source for misplaced labels at metatile boundaries.


nebulon42:
Interesting, I'll try that.

nebulon42:
Changed. Interestingly, I now saw improvements with London and Valencia.

PR #2405, opened 2016-10-09
Remove hatching for military areas

nebulon42:
I propose to remove the red hatching for military areas. The hatching has the following problems:
- the hatches don't adjust with zoom level which does not look good on lower zooms
- on higher zooms the hatches obscure the interior of the military area making the map both harder to read and also more difficult to handle for mappers who map the contents of military areas

I don't know of any other general purpose map that displays military areas as prominently as we do. Most online maps I looked at don't display military areas at all or much more subtle. Very few have hatching and only on higher zooms and only below everything else.

I know that hatching is a common cartographic choice for military areas, but I think it is only suitable for low zooms. By rendering military areas similar to national parks we maintain the visibility on lower zooms and also gain better display of the interior on higher zooms.


nebulon42:
Previews (always before and then after)

z8
z8_before
z8_after

z9
z9_before
z9_after

z10
z10_before
z10_after

z11
z11_before
z11_after

z12
z12_before
z12_after

z13
z13_before
z13_after

z14
z14_before
z14_after

SomeoneElseOSM:
I'd agree that it makes sense to change but I'd still retain some hatching - I recently made a similar change elsewhere, from:
military_before
to
military_after

it doesn't remove the hatching altogether, but makes it 32x32 instead of 16x16, and the underlying colour is changed from "military pink" to something less prominent.  See https://github.com/SomeoneElseOSM/openstreetmap-carto-AJT/commit/48e56466fd6db9570b940a048fb9f35699427fe2 and part of https://github.com/SomeoneElseOSM/openstreetmap-carto-AJT/commit/6d7413474a09e0de71c3878adc00be3518d23249 .

aceman444:
Yes, keeping at least the sparse hatching is useful for large military areas (think tens of kilometers) where you may not notice you are viewing a military area if you are zoomed in too much (if only the borders would be shown as in the proposal).

pnorman:
I prefer having hatching - maybe not as strong as now, but something wider than @SomeoneElseOSM's example.

matkoniecz:
I also prefer hatching. I would not oppose changing it, but https://cloud.githubusercontent.com/assets/3531092/19222730/754e5188-8e5f-11e6-91e1-72bd45c6f871.png https://cloud.githubusercontent.com/assets/3531092/19222724/65aaf524-8e5f-11e6-8f6c-5e53da4db9b2.png https://cloud.githubusercontent.com/assets/3531092/19222701/f21c2a10-8e5e-11e6-8a23-bdf7bde30772.png are made more confusing and unclear by change.

dieterdreist:
> Il giorno 11 ott 2016, alle ore 23:41, Paul Norman <notifications@github.com> ha scritto:
> 
> I prefer having hatching - maybe not as strong as now, but something wider than @SomeoneElseOSM's example.
> 



I prefer having a hatching as well, I'd make it denser than someone else's version, the line width seems ok in his version though.

math1985:
I also would prefer making the hatching lighter rather than dropping it fully.

nebulon42:
@matkoniecz I don't think that it gets more confusing. The border is clearly marked. It's the same with national parks. The removal of the NR pattern did not confuse things but improve the rendering. Also most military areas are not danger areas all the time that would require such a prominent marking on all zoom levels. Also your example "Maria Theresien Kaserne" is just a normal military base which would not require hatching at higher zooms.

I also don't think that the hatching for prisons is a good idea. It is just for effect and not cartographically justified. But this is off-topic here.

My new approach would be to lighten the hatching on lower zooms but to remove the hatching on very high zooms, because it doesn't add information and just obscures things on this levels.

Closing, because no ETA for rework.

aceman444:
Maybe you could consider keeping the hatching for some types of military areas even at high zooms, e.g. for testing/shooting ranges as those could be "danger areas".

nebulon42:
@aceman444 Yes, I have no problem with that but this would need proper tagging support.

PR #2408, opened 2016-10-10
Add back captial-name WHERE conditions

### LINKED ISSUE #2407: capital-names doesn't require place tag
The WHERE condition on the `capital-names` layer is
```sql
WHERE capital = 'yes'
  AND admin_level = '2'
```

Capitals [used to require `place IN ('city', 'town') AND name IS NOT NULL`](https://github.com/gravitystorm/openstreetmap-carto/pull/2314/files#diff-0abb12fe08393e8ddccd19e688ded6e5R1597). This means the new code can't use the point place partial index with its `place IS NOT NULL AND name IS NOT NULL` current index.

pnorman:
Fixes #2407

### DIFF
--- a/project.mml
+++ b/project.mml
@@ -1267,7 +1267,7 @@
       "srs": "+proj=merc +a=6378137 +b=6378137 +lat_ts=0.0 +lon_0=0.0 +x_0=0.0 +y_0=0.0 +k=1.0 +units=m +nadgrids=@null +wktext +no_defs +over",
       "Datasource": {
         "extent": "-20037508,-20037508,20037508,20037508",
-        "table": "(SELECT\n    way,\n    name,\n    CASE\n      WHEN (population ~ '^[0-9]{1,8}$') THEN population::INTEGER ELSE 0\n    END as population,\n    round(ascii(md5(osm_id::text)) / 55) AS dir -- base direction factor on geometry to be consistent across metatiles\n  FROM planet_osm_point\n  WHERE capital = 'yes'\n    AND admin_level = '2'\n  ORDER BY population DESC\n) AS capital_names",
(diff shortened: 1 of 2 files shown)


nebulon42:
Wait a sec, capitals can also be villages like Vaduz: https://www.openstreetmap.org/node/1932181216.
I don't think we should exclude those. Maybe make it `place IS NOT NULL` to include all possibilities?

pnorman:
> Wait a sec, capitals can also be villages like Vaduz: https://www.openstreetmap.org/node/1932181216.

Because this is a bugfix I'd rather keep the old `place IN ('city', 'town')`, get the patch release out, then look at changing it.

dieterdreist:
> Il giorno 10 ott 2016, alle ore 22:44, Michael Glanznig <notifications@github.com> ha scritto:
> 
> Wait a sec, capitals can also be villages like Vaduz: https://www.openstreetmap.org/node/1932181216.


it's likely a tagging error, italian, english and german wikipedia concur it's a town, which would also seem more likely for a capital (already the fact a settlement is a capital will make it at least a town I think).

jojo4u:
I'm collecting non-town and non-city capitals here: https://wiki.openstreetmap.org/wiki/Key:capital
The bugfix-release should make sure there is always a name rendered. Special capital rendering besides city and town can some later.

dieterdreist:
2016-10-10 23:45 GMT+02:00 jojo4u <notifications@github.com>:

> I'm collecting non-town and non-city capitals here:
> https://wiki.openstreetmap.org/wiki/Key:capital
>


there are issues with most of these, I have commented on this page...

pnorman:
Merging since it gets us to 2.43.0 behavior, fixes a significant performance regression, and how we should label capitals which aren't towns or cities needs more discussion.

Issue #2099, opened 2016-03-29
stop rendering shop=yes


matkoniecz:
Rendering shop=yes and not rendering rare shop values encourages to omit detail during tagging - for example I almost tagged https://www.openstreetmap.org/way/406428679#map=19/50.07983/19.87729 as shop=yes just to get it to render on map.

I think that it is not a desirable effect.

aceman444:
So why do we not render any shop=*?

HolgerJeromin:
@aceman444  We do not want to hide typos for example. shop=suprmarket would be not valid (or a bad idea at least), so the default map should not provide the feedback that this is a correct entry.

moliha:
A general problem-icon could be usefull - this could be a small question mark. This could mark the node as questionable - but the name could be shown. Hiding problem-nodes is the worst "solution" for all:
- The responsible mapper gets no feedback.
- Other mappers, who might be able to fix the problem, are not alerted because it is hidden.
- Map users see nothing where 'something' (and a name) could be usefull.

aceman444:
Holger, typoed values would only show as the dot, not the specific icon, so that would be visible.

HolgerJeromin:
Oh, sure. This requires that you are familiar with the rendering of all valid shop values.
But shop=frame would look the same as shop=frames and 
shop=fabrics will look like the correct shop=fabric

HolgerJeromin:
this style is no QA map. Many websites are using the tiles. There a question mark would be very confusing.

pnorman:
My personal preference is to render shop=* as a dot, but instead we decided to whitelist specific shops. Given this choice, I'm not sure it makes sense to render shop=yes.

sommerluk:
The example is a shop=flower_pot. Maybe, if shop=yes is not rendered anymore, people tend to use related, but wrong tags. I could imagine that someone would use shop=florist instead to get it rendered, and that would be worser than shop=yes…

d1g:
For simplicity, render shop=* (except shop=no or shop=vacant) as dot.

Whitelist-only rendering both: is unrewarding for mappers and painful for style developers.

martinum4:
i agree with sommerluk here, we get highly different names for the same stuff, especially regarding AE/BE there will be different writing of the same stuff then and it will be harder to clean it up afterwards...

nighto:
Hello!

 

You have a new message, please read <http://aksozdokuma.com/broadcast.php?du>

 

[...]

PR #2415, opened 2016-10-19
Rendering rare shop values

kocio-pl:
Resolves https://github.com/gravitystorm/openstreetmap-carto/issues/2099.

All the white lists for `other` shops:

``` SQL
WHEN shop IN ( ... ) THEN 'other' ELSE NULL END,
```

have been replaced with just basic checking if it's really a shop:

``` SQL
WHEN shop = 'no' OR shop = 'disused' OR shop IS NULL THEN NULL ELSE 'other' END,
```

I'm not familiar with SQL, but testing shows that it works. Does it make any sense to check also if `shop` is not null to exclude all the shops having their own icons?


pnorman:
The decision to render specific shops was a conscious decision, we wanted to avoid rendering mistakes like `shop=supermarkte`. We can revisit it, but #2099 is about `shop=yes`, not about all shop values that don't have an individual rendering.

At the time we made the decision I would have preferred rendering `shop is not null and shop not in ('vacant',...)` and not specially rendered as a dot. Something which wasn't originally considered is the maintenance burden of maintaining two lists of shops, one with special renderings, another with a generic rendering.

kocio-pl:
#2099 has proposed solution in the title, but the problem description:

> Rendering shop=yes and not rendering rare shop values encourages to omit detail during tagging

shows that the title is not precise, because not showing rare shops is the real source and shop=yes is just a common workaround - a valid general value which is probably misused. BTW - the misuse is marginal comparing to building=yes (3,44% vs 82.12% of total uses with given key).

Therefore I think we could start with fixing the source and wait some time to see if any other action is needed, as proposed [here](https://github.com/gravitystorm/openstreetmap-carto/issues/2099#issuecomment-254538345). I like this solution, that's why I didn't exclude shop=yes. Is it OK for you or you want to decide also this question now?

kocio-pl:
Thanks for bringing up the `vacant` value! I was sure there is something like this, but took `disused` instead by mistake (it's good to filter it out too)... Now the line is:

```SQL
WHEN shop IN ('no', 'vacant', 'disused') OR shop IS NULL THEN NULL ELSE 'other' END,
```

kocio-pl:
I've added some more shop values indicating that shop is non-existing - I took them from `shop_values.rb` and removed this script.

pnorman:
I'm going to go ahead and merge this. There are arguments either way on if it's better to have a whitelist of minor shop values from a tagging and cartographic viewpoint.

What settles it for me is the maintenance issues. We haven't been maintaining the lists, the same large list is supposed to be identical in four places, and the queries shops are part of are already some of larger ones in the style.

gravitystorm:
For the record, I disagree with this approach. It was a major improvement to avoid catch-all rendering, which this PR re-enables. I understand the lack of maintenance on the shops list but it would be better to maintain the list rather than re-introduce a catch-all.

The proliferation of shop values is outside the scope of the repo, but the trouble we are having here highlights that this is in itself a problem that I would prefer tackled too. This would most likely be by introducing categories and subcategories of shops, rather than having hundreds of artificially discreet values.

I'm not suggesting that this to be reverted, I just want to make my point here.

kocio-pl:
The point is we were not able to maintain this list and I haven't find a better way. But if somebody find the right tool to control it (here or on the data level), I would be fine with it.

Categories may be a good thing, not only for shops, but also for other objects. I even tried to introduce it once, but nobody followed and I didn't know how the implementation should look like. Do you have any ideas about it?

aceman444:
You still have a whitelist of shop types that have a specific icon. All others should get just the dot. In that way, you still see typos, like the mentioned 'supermarkte', because it will only get a dot, instead of the mapper's expected icon. But it is still a shop despite the typo, so it should appear on the map.

HolgerJeromin:
@aceman444 
But you will not find the `shop=beds` typo (or unusual/unwanted tagging) which would be `shop=bed`.

dieterdreist:
sent from a phone

> Il giorno 16 nov 2016, alle ore 14:51, Andy Allan <notifications@github.com> ha scritto:
> 
> The proliferation of shop values is outside the scope of the repo, but the trouble we are having here highlights that this is in itself a problem that I would prefer tackled too. This would most likely be by introducing categories and subcategories of shops, rather than having hundreds of artificially discreet values.
> 


in the end you won't end up with less tags for distinct shop types, it would just be easier to find them - maybe. This is also something the editors can provide without having it hardcoded into the tags, josm and others are already doing it.

+1 to giving the shops list more care. I suggest to give more importance to an existing wiki definition rather than just usage numbers, and lower the threshold to a few hundred.

nebulon42:
I'm not too fond of this change too. We are not the map of shops, but it sometimes seems we are. They have gotten far too much attention and we should render much less than more.

kocio-pl:
I've created a ticket to sort the things out, because the problem is not easy to define (let alone to solve). The https://github.com/gravitystorm/openstreetmap-carto/issues/2444 is a better place to discuss it than a PR closed a month ago.

aceman444:
@HolgerJeromin why not? If one is in the whitelist with specific icon, you will see if it didn't get it. If none has a specific icon, it means the value is not important yet (few usages worldwide), so then it is not even sure which value is the correct one.

HolgerJeromin:
My example was clear: beds 2700 vs bed 1.

aceman444:
So does 'beds' have a special icon?

HolgerJeromin:
No and thats the whole point

aceman444:
So in that case, how did you determine 'beds' is the proper value and not 'bed'? Once it is important enough and gets a special icon, the typos may show up (with dot instead of the expected icon).
Or was it important enough till now to be in the whitelist to get the dot, but not an icon?

HolgerJeromin:
Ok, take `shop=baby_goods` (which was in the dot white liste, is approved by the [wiki](https://wiki.openstreetmap.org/wiki/Tag:shop%3Dbaby_goods) and has 2090 elements) and `shop=baby_good` (which has no wiki page and is not used in the database). 

One propose of this style is "[feedback mechanism for mappers to validate their edits](https://github.com/gravitystorm/openstreetmap-carto/blob/master/CARTOGRAPHY.md)". I think most people does not call `shop=baby_good` a valid or at least useful/wanted entry in our database.

aceman444:
So was baby_goods only with dot and no icon?

I think the feedback can then be solved with having 2 lists:
1. list with shop values that have each its own special icon
2. list of other accepted values (e.g. >1000 uses and a wiki page). These may be rendered by a common shop icon.
3. all the other unknown shop values will be rendered with a dot.

PR #2423, opened 2016-10-30
Check range of supported Carto and Mapnik with Travis

pnorman:
Doing a PR to test the travis build

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,11 +8,16 @@ addons:
     - python-yaml
 node_js:
   - "0.10"
+env:
+- CARTO=0.9.5 MAPNIK=2.1.0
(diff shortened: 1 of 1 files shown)


pnorman:
Our minimum version of CartoCSS doesn't support the `-a` option, so I'm not sure what version of Mapnik it's assuming, so it's possible it's producing output that 2.1.0 doesn't support.

What's going to finally drive me to writing a new style based on different technologies is struggling with Mapnik.

kocio-pl:
That would be a huge porting effort... What technologies do you mean?

kocio-pl:
I guess this PR is no longer needed and can be closed because od merging https://github.com/gravitystorm/openstreetmap-carto/pull/2473?

math1985:
Agree.

Issue #688, opened 2014-07-02
Building doesn't render on top of highway areas any more


sb12:
Example way: http://www.openstreetmap.org/way/104978884
Other example area: http://www.openstreetmap.org/#map=19/52.52496/13.36941

This happened after the recent style changes a few days ago. 
Before buildings and roofs were rendered on top of the highway areas (and highway ways with covered=yes/tunnel=yes) and you could still see the highways because of the building transparency. Now you can't see the buildings anymore when there's a highway area inside.
Is this a wanted behaviour or an error?

kocio-pl:
So - are there any ideas how to change the stylesheet? I think it's getting serious after such a long time because such places (with complicated buildings/highways layers) are probably important.

daganzdaanda:
IMHO we could start by rendering anything that is tagged covered=yes with 50% opacity. No need to change rendering order then. It might still look weird, but it should make it more clear what's going on. 

Later, other changes may be smarter, like interpreting 3D-tags or the reative layers, or "location".

kocio-pl:
Could someone make a test renderings of it? I suggest Berlin Hauptbahnhof as a great example of complex multilevel building, but I'm also very interested in Dworzec Centralny in Warsaw (my hometown).

jojo4u:
Is this building (http://www.openstreetmap.org/way/23691951) being under rather over a railway=platform layer/level=-3 the same problem or a different issue?

Theodin:
It is over the platform asn the building is on the surface and the platform is a subway platform with 2 moving stairways leading down.

kocio-pl:
It looks like the hack with render=no [stopped working](http://www.openstreetmap.org/way/308077768) and Eiffel Tower is now broken too. 

@gravitystorm Andy, so what about such important places? Do we still accept this class of problems as unavoidable cost of better showing the pedestrian areas (as we do it now) or maybe we should have kind of workaround for such exceptions?

matkoniecz:
"render=no" never had any effect on whatever something will be rendered.

kocio-pl:
So maybe that was something else - I just know that building was visible back then, but the problem is what in general should we do with such things?

malenki:
Every time I see an amenity=fuel with a building=roof and layer=1 with service highways going over them I sigh inwardly.
Now I even bothered to search for this issue to write this comment.
A solution would be nice.

HolgerJeromin:
[...]

PR #2446, opened 2016-11-17
Proposal for design goals and guidelines - part 1

imagico:
I have started with a draft of putting together some design principles for this style that hopefully help contributors in the future to better understand the direction of development.

This first part contains two of three sections i structured this into.  These two are the description of the style purposes and a list of the main goals derived from these.  I plan to supplement this with a section with more specific guidelines on style decisions with - from my side - subsections on colors and zoom levels - which can be supplemented by others on further subjects like for example icon design.

This structure is meant to show that the specific principles listed are not arbitrary but are meant to serve the described goals which are followed to make the style serve its purposes.

Mostly i tried to put together ideas that i hope we can mostly agree upon.  But there are also a few things that might be more controversial.  The most significant probably is that i put the goals into an order which - as i put it at the moment, is quite a bit different from the current priorities.  In my eyes point 5 and 6 currently have a much higher priority while point 3 is often less important.  

To give an example (just an example, not meant to reopen the discussion) the decision to close #1853 was based on the maintainability aspect although in my eyes the change was a significant improvement regarding at least goal 2 and 3.

I am open to changing this order based on arguments discussed here, alternatively we can also make this an unordered list but an order of the goals would of course make priorities clearer for potential contributors.

The ordering as i put it for the moment represents the priorities the different goals would have for me - but of course we should ultimately put them in the order that represents actual practice.

### DIFF
--- a/CARTOGRAPHY.md
+++ b/CARTOGRAPHY.md
@@ -4,7 +4,7 @@
 This is an attempt to outline the goals of this style and the principles under 
 which the maintainers make decisions.  These rules are not set in stone, they 
 can change and they may not be followed in all cases but contributors should 
-be able to expect they are generally the guiding principles design wise.
+be able to expect that they are generally the guiding principles design wise.
 
 It does not make much sense to try following these principles blindly as a 
 contributor without understanding them, they are meant to guide you to develop 
@@ -32,7 +32,7 @@ of course also technical constraints and requirements that need to be taken into
 account.
 
 * **Legibility and clarity** - The map should be intuitively readable by users with some general experience using maps without a map key, preferrably with relatively little effort.  A map key or more extensive experience using this map style can be required for clearly identifying minor differences or the exact meaning of certain features but in broad strokes orientation and identification of map elements should be possible on an intuitive level.  We also aim for the map appearance to be esthetically pleasing.
-* **Being understandable and supportive for the mappers** - To serve as feedback for mappers and encourage correct mapping this style needs to render the data in a way that allows mappers to understand how the data produces a certain rendering result based on basic observation without in depth understanding how map rendering works or looking at the style implementation.
+* **Being understandable and supportive for mappers** - To serve as feedback for mappers and encourage correct mapping this style needs to render the data in a way that allows mappers to understand how the data produces a certain rendering result based on basic observation without in depth understanding how map rendering works or looking at the style implementation.
 * **Diversity** - The style should represent the diversity of the OSM community and geography in general.  The most obvious element to serve this goal is showing the local names everywhere on earth in their respective scripts.  This goal however goes beyond labels.  Both physical and cultural geography differs a lot globally and the aim is to represent this variety with equal determination - well mapped areas are not supposed to have more weight here than less mapped parts of the world.  This also means the target map user is the potential global map user and no special consideration is given to the current geographic distribution of actual map use.
 * **A rich map** - This style deliberately creating a fairly rich map showing a significant number of different features.  This way it shows the richness of OSM data and gives a broad recognition to the mappers' work.  The aim is not however to show all or even most of the OSM data.
 * **Maintainability** - The implementation of this style should not be too hard to maintain. This refers to both the volume and complexity of the code and how fast the style can be parsed when rendering it, which is very important for efficient development work. So the amount of code should be kept small and complex and fragile interdependencies should be avoided. If the code is difficult to maintain this would ultimately seriously affect all of the above goals.
--- a/CARTOGRAPHY.md
+++ b/CARTOGRAPHY.md
@@ -1,39 +1,49 @@
-# Cartography
 
-This is a style that serves multiple purposes, so here are some guidelines when considering cartographic changes.
+# Design goals and guidelines for this style
 
-## Purposes
+This is an attempt to outline the goals of this style and the principles under 
+which the maintainers make decisions.  These rules are not set in stone, they 
+can change and they may not be followed in all cases but contributors should 
+be able to expect they are generally the guiding principles design wise.
 
-There are multiple primary purposes of the map style, which pull in different directions
+It does not make much sense to try following these principles blindly as a 
+contributor without understanding them, they are meant to guide you to develop 
+an intuition and understanding how to make design decisions to fit into the 
+overall concept of this style.
 
-* It's the primary feedback mechanism for mappers to validate their edits - so detail is useful
-* It's a major part of the impression visitors to osm.org receive - so clear design is useful
-* It's an examplar stylesheet for rendering OSM data - so easy customisation is useful
+## General purpose
 
-It must always be borne in mind that a map style cannot show every detail of the OSM data, and in many cases it is more appropriate to show the detail in other, more specialist styles.
+This style has multiple purposes:
 
-## Colours
+* It's an important feedback mechanism for mappers to validate their edits.
+* It's a major part of the public face of OpenStreetMap, for many people the map on osm.org rendered with this style _is_ OpenStreetMap.
+* It's used in many map applications as a general purpose map.
+* It's an examplar stylesheet for rendering OSM data.
 
-Firstly, this is a map, not merely a colourful 2-dimensional visualisation of the database. Colours should be chosen based on their effectiveness and to make things look nice, not chosen for distinctiveness.
+There is no ranking of these purposes.  To allow serving all of them and to 
+avoid satisfying only some at the cost of the others the following main goals 
+have been identified.
 
-The colour palette should be moving towards pastel/light/desaturated for background layers, midtones for streets and save highlights/bolds/saturated for points of interest. If colourspace can be left unused, that enables overlays for third parties.
+## Main goals
 
-Colour definitions should, where useful, be put into variables at the top of the file, to enable easier customisation.
+The following goals form an ordered list with the most important on top.  This 
+essentially means pursuit of the goals further at the bottom should not 
+sacrifice the goals higher up.  All goals are however relevant to fulfill the 
+purposes above and should be balanced as far as possible.  Some compromises 
+need to be made with the higher priority goals on occasions to not completely 
+ruin the lower priority ones.  Apart from these goals there are of course also 
+technical constraints and requirements that need to be taken into account.
 
-## Data manipulation
+1. **Legibility** - The map should be intuitively readable by users with some general experience using maps without a map key.  A map key or more extensive experience using this map style can be required for clearly identifying minor differences or the exact meaning of certain features but in broad strokes orientation and identification of map elements should be possible on an intuitive level.
 
-OpenStreetMap data has to be manipulated for rendering, but since this style is intended for use by mappers to check their work, it should minimise any distortions. For example, line-smoothing improves the look of railways and rivers, but introduces confusion for mappers. Post processing of geometries can improve the cartographic results, but breaks the cause-and-effect between editing the data and seeing the results.
+2. **Being understandable and supportive for the mappers** - To serve as feedback for mappers and encourage correct mapping this style needs to render the data in a way that allows mappers to understand how the data produces a certain rendering result based on basic observation without in depth understanding how map rendering works or looking at the style implementation.
 
-For similar reasons, use of external non-OSM data should be avoided.
+3. **Diversity** - The style should represent the diversity of the OSM community and geography in general.  The most obvious element to serve this goal is showing the local names everywhere on earth in their respective scripts.  This goal however goes beyond labels.  Both physical and cultural geography differs a lot globally and the aim is to represent this variety with equal determination - well mapped areas are not supposed to have more weight here than less mapped parts of the world.  This also means the target map user is the potential global map user and no special consideration is given to the current geographic distribution of actual map use.
 
-## Legibility
+4. **Clarity** - This is strongly and directly related to legibility but more specifically means the map should be readable with as little effort as possible and be pleasant to look at both with high and low levels of interest and concentration.
 
-Legibility is not only about of font size, but also about the ability of users to be able to “read” the map style. For example, a user might not be familiar with our exact road colours, but should be able to understand their relative importance via intensity of colour, width or other attributes. Important features should be more easily spotted and understood than less important information. It should be possible to gain an understanding of the majority of the map without using a legend.
+5. **A rich map** - This style deliberately creating a fairly rich map showing a significant number of different features.  This way it shows the richness of OSM data and gives a broad recognition to the mappers' work.  The aim is not however to show all or even most of the OSM data.
 
-Avoid font sizes smaller than 10.
+6. **Maintainability** - The implementation of this style should not be too hard to maintain.  This means the amount of code should be kept small and complex and fragile interdependencies should be avoided.  If the code is difficult to maintain this would ultimately seriously affect all of the above goals.
 
-## The Mapper Feedback Loop
-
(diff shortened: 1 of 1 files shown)


CARTOGRAPHY.md
```diff
@@ -1,39 +1,49 @@
...
 
-For similar reasons, use of external non-OSM data should be avoided.
+3. **Diversity** - The style should represent the diversity of the OSM community and geography in general.  The most obvious element to serve this goal is showing the local names everywhere on earth in their respective scripts.  This goal however goes beyond labels.  Both physical and cultural geography differs a lot globally and the aim is to represent this variety with equal determination - well mapped areas are not supposed to have more weight here than less mapped parts of the world.  This also means the target map user is the potential global map user and no special consideration is given to the current geographic distribution of actual map use.
 
-## Legibility
+4. **Clarity** - This is strongly and directly related to legibility but more specifically means the map should be readable with as little effort as possible and be pleasant to look at both with high and low levels of interest and concentration.
```
kocio-pl: I see 4. as a subset of 1. - maybe it needs just some rewording to show how it is different (and worth different item)? "Pleasant to look" might be this difference, but the name of this item suggests something else.

imagico: Yes, you can consider this part of legibility.  My idea was that 1. just means the map can be read and understood by the user while 4. aims for this being easy and possible without much effort.  This is much harder to achieve and usually does not play much of a role here.

kocio-pl: Map which can't be understood wouldn't make much sense, I read "intuitively readable" from 1. exactly as "can be easily understood".

math1985:
Looking good!

I think you're interpreting legibility too narrow. For example, the [old Luxembourg bus map](http://mudec.weebly.com/uploads/2/2/7/9/22794876/9500943_orig.png) is perfectly intuitively understandable, but very hard too read. In comparison, the [new Luxembourg bus map](http://www.jugcerovic.com/files/luxembourg-plan-autobus-bus-map.png) is much easier to read.

Perhaps under Diversity, or as a separate point, you could also add that the map does not cater to one specific type of user (car drivers, cyclists etc.) but tries to be useful for all?

imagico:
> I think you're interpreting legibility too narrow.

My idea was that the top goal that stands above all others is the basic legibility.  Ease of use is more what i had in mind with the _Clarity_ goal.  But if preferred we can combine them and put them on top as general legibility including ease of use.  Currently however my impression is that being easy to read often does not have priority over mapper feedback - we would do more geometry processing then for example.

> Perhaps under Diversity, or as a separate point, you could also add that the map does not cater to one specific type of user (car drivers, cyclists etc.) but tries to be useful for all?

Maybe just add _or to special thematic interests of users_ at the end of the diversity goal?  Alternatively feel free to add it as you see fit, the idea in general is very good.

CARTOGRAPHY.md
imagico: Intuitively does not mean easily - it just means you don't need additional information to figure it out, it might still take a lot of effort.

In general i think we should do this based on suggestions.  If you think something should be changed make a suggestion to change it.  This way we'd make better progress than by arguing about the meaning of words.

kocio-pl: OK, sure. My proposition for now is that 4. should be removed and "pleasant to look" moved to 1.

math1985:
To be honest I'm a bit skeptic about us being able to order the goals in terms of priority. Otherwise we'd end up having to block big progress in one area because it would mean a small regression in another area. I'd be more comfortable with an unranked list.

CARTOGRAPHY.md
```diff
@@ -1,39 +1,49 @@
...
+## General purpose
 
-It must always be borne in mind that a map style cannot show every detail of the OSM data, and in many cases it is more appropriate to show the detail in other, more specialist styles.
+This style has multiple purposes:
 
-## Colours
+* It's an important feedback mechanism for mappers to validate their edits.
+* It's a major part of the public face of OpenStreetMap, for many people the map on osm.org rendered with this style _is_ OpenStreetMap.
+* It's used in many map applications as a general purpose map.
+* It's an examplar stylesheet for rendering OSM data.
```
nebulon42: I would add here that openstreet-map is plays an important part in tag usage behaviour of mappers and in tag hygiene.

nebulon42:
Very good work and very good starting point. Yes, a ranking is maybe a bit hard to obey. But on the other hand things like feedback-loop/feature-richness and clarity/aesthetics always get us into trouble. A ranking would help to indicate a general direction. This could be also indicated at the end of a unordered list. A long term goal should be to reduce the central importance of openstreetmap-carto for OSM by having more different main styles. But this might require basic changes to the rendering infrastructure and is of course out of scope here.

imagico:
Keep in mind the purpose of this is not to set constraints for us to obey - we set the goals, we can also change them.  The main purpose is to help contributors understand what the basis is of the design decisions we make.  So if the goals have an order of priorities for us we should document that, even if practically this ordering is very weak.

Maybe we could just collect a set of opinions from the maintainers if and how the different goals have priorities for them - like 'all the same importance to me', '2 is most important but the rest is all the same to me', '1,4 and 7 are slightly more important than 2,3,5 and 6' or a fully sorted list.  But before we do that we might better think about if we missed any important goals.

kocio-pl:
A bit off topic, because we lack place to discuss general things (do we want to have one?): I've just found a [journey planner](http://www.teroplan.com/) which uses osm-carto with different colors, that's an example of 7. (adaptability).

math1985:
I think the performance of the rendering is also one of the design goals.

math1985:
> I've just found a journey planner which uses osm-carto with different colors, that's an example of 7. (adaptability).

I think this journey planner is around longer than osm-carto, so it's more likely based on the style preceding osm-carto.

kocio-pl:
Looking at some osm-carto specific problems (https://github.com/gravitystorm/openstreetmap-carto/issues/688) their style has been updated at some point.

imagico:
Regarding performance - this in my eyes has three components:

* the constraint that we can't put too much additional load on the OSMF servers with style changes.  I don't think this should be part of the design goals since this is an absolute external contraint like the database schema and (to some extent) the software used.  It might make sense to list these in a separate document of course to help new contributors understand what they can and cannot do on a technical level.
* the internal performance needs regarding our work - this should be part of maintainability i think.
* the goal to keep the style not too demanding on tile serving infrastructure so others can use it fairly easily - this is kind of related to the adaptability goal.

Technically the first and last points are mostly about style rendering performance while the second is more about code complexity and parsing performance.  We could acknowledge these goals with the following additions:

6. **Maintainability** - The implementation of this style should not be too hard to maintain. This refers to both the volume and complexity of the code and how fast the style can be parsed when rendering it, which is very important for efficient development work.  So the amount of code should be kept small and complex and fragile interdependencies should be avoided. If the code is difficult to maintain this would ultimately seriously affect all of the above goals.  

7. **Adaptability and ease of use** - The style should be easy to customize, like for creating localized or special purpose maps.  It is also important to keep demands on rendering infrastructure for serving the style low so it is not too difficult and costly to set up a tile server for this style or a specialized variant of it.

math1985:
Sounds good to me.

pnorman:
Our meeting of the goals needs to balance them against each other. An unordered list might make this clearer.

pnorman:
This might belong in a follow up discussion and PR, but these are thoughts I've been having on what we render. Features should be

- useful to someone looking for something on a map remotely;
- used to understand the general character of an area;
- used for finding something when you're in the area;
- used for orienting yourself when in an area; or
- commonly rendered on many maps.

So a specialized shop not of general interest would not need its own icon, but somewhere to eat does.

Ircama:
IMHO, the order proposed by @imagico looks appropriate and there are principles of a certain depth.

I would also suggest, when possible, that the standard style should tend to give more priority to features with wide social and cultural relevance and to basic topographic features (meaning the ones that are of common usage and that can be usually found on generic maps to physically describe the territory) rather than business elements (like shops, malls or factories).

imagico:
I made a change with the following modifications:

* since there has been no clear voice in support of an order or priorities for the goals from any other maintainer the goals are now in an unordered list.
* since without an order of priorities it does not matter anyway i combined the clarity goal with legibility.
* the last two goals are modified as indicated
* i tried to integrate the suggestion from @nebulon42 into the mapper feedback purpose.

From my side this would be a reasonable first version but subsequent adjustments will of course likely be necessary.  Also some proof reading regarding language and orthography would likely be good.

I still think it would be good to have an idea where the maintainers stand regarding the priorities of the different goals even if we don't want to define such an order for the project.  But i don't want to push this.  Please speak up if you think we should do this.

Regarding decisions on what to render and what not to render - i have a few considerations on this in my notes for more specific guidelines i will put in a supplement proposal.  But in general this is fairly difficult, especially if it goes into subjective concepts like usefulness. 

@Ircama - _relevance_ is a problematic concept here - this is always highly subjective.  I try to concentrate on things here that are at least to some extent objective so contributors can check their idea aginst these goals and guidelines without the inevitable frustration that would occur if you consider something highly relevant but find out that the maintainers do not.

There will of course always also be subjective decisions but we can at least try to document the basis for those which are not.

imagico:
I have a first version of the second part ready now - i can add it here but it might be better to merge this first and start a new PR.

pnorman:
> I have a first version of the second part ready now - i can add it here but it might be better to merge this first and start a new PR.

👍 

If none of the other maintainers merge it, I'll have a last read through in the morning when I've had more sleep.

CARTOGRAPHY.md
```diff
@@ -1,39 +1,39 @@
-# Cartography
 
-This is a style that serves multiple purposes, so here are some guidelines when considering cartographic changes.
+# Design goals and guidelines for this style
 
-## Purposes
+This is an attempt to outline the goals of this style and the principles under 
+which the maintainers make decisions.  These rules are not set in stone, they
```
nebulon42: Extra space between sentences.

CARTOGRAPHY.md
```diff
@@ -1,39 +1,39 @@
-# Cartography
 
-This is a style that serves multiple purposes, so here are some guidelines when considering cartographic changes.
+# Design goals and guidelines for this style
 
-## Purposes
+This is an attempt to outline the goals of this style and the principles under 
+which the maintainers make decisions.  These rules are not set in stone, they 
+can change and they may not be followed in all cases but contributors should 
+be able to expect they are generally the guiding principles design wise.
```
nebulon42: expect _that_ they are generally

CARTOGRAPHY.md
```diff
@@ -1,39 +1,39 @@
...
+This style has multiple purposes:
 
-## Colours
+* It's an important feedback mechanism for mappers to validate their edits and helps to prevent unfavorable fragmentation of tag use.
+* It's a major part of the public face of OpenStreetMap, for many people the map on osm.org rendered with this style _is_ OpenStreetMap.
+* It's used in many map applications as a general purpose map.
+* It's an examplar stylesheet for rendering OSM data.
 
-Firstly, this is a map, not merely a colourful 2-dimensional visualisation of the database. Colours should be chosen based on their effectiveness and to make things look nice, not chosen for distinctiveness.
+There is no ranking of these purposes.  To allow serving all of them and to
```
nebulon42: Extra space between sentences

CARTOGRAPHY.md
```diff
@@ -1,39 +1,39 @@
...
+There is no ranking of these purposes.  To allow serving all of them and to 
+avoid satisfying only some at the cost of the others the following main goals 
+have been identified.
 
-The colour palette should be moving towards pastel/light/desaturated for background layers, midtones for streets and save highlights/bolds/saturated for points of interest. If colourspace can be left unused, that enables overlays for third parties.
+## Main goals
 
-Colour definitions should, where useful, be put into variables at the top of the file, to enable easier customisation.
+The following goals need to be balanced against each other to serve the purposes 
+above.  There is no fixed order of priorities.  Apart from these goals there are
```
nebulon42: Extra spaces between sentences.

CARTOGRAPHY.md
```diff
@@ -1,39 +1,39 @@
...
-
-While this desire is in obvious conflict with the above comments on level of detail, it has a second impact on the map style. We need to, wherever possible, avoid accidentally encouraging mistakes - that is, avoiding the situation where a clearly misspelled or misused tag leads to the originally expected result. So `highway=mtorway` shouldn't show up as if it was `highway=motorway`, and so on. This has been a problem with "catch-all" rules in queries and filters, such as `where leisure is not null` or `[amenity != ""]`. Avoid these situations.
\ No newline at end of file
+* **Legibility and clarity** - The map should be intuitively readable by users with some general experience using maps without a map key, preferrably with relatively little effort.  A map key or more extensive experience using this map style can be required for clearly identifying minor differences or the exact meaning of certain features but in broad strokes orientation and identification of map elements should be possible on an intuitive level.  We also aim for the map appearance to be esthetically pleasing.
```
nebulon42: Extra spaces between sentences.

CARTOGRAPHY.md
```diff
@@ -1,39 +1,39 @@
...
+* **Being understandable and supportive for the mappers** - To serve as feedback for mappers and encourage correct mapping this style needs to render the data in a way that allows mappers to understand how the data produces a certain rendering result based on basic observation without in depth understanding how map rendering works or looking at the style implementation.
```
nebulon42: I would leave out _the_ in _the mappers_. ... supportive for mappers

CARTOGRAPHY.md
```diff
@@ -1,39 +1,39 @@
...
+* **Diversity** - The style should represent the diversity of the OSM community and geography in general.  The most obvious element to serve this goal is showing the local names everywhere on earth in their respective scripts.  This goal however goes beyond labels.  Both physical and cultural geography differs a lot globally and the aim is to represent this variety with equal determination - well mapped areas are not supposed to have more weight here than less mapped parts of the world.  This also means the target map user is the potential global map user and no special consideration is given to the current geographic distribution of actual map use.
```
nebulon42: Extra spaces.

nebulon42: Maybe worth clarifying that diversity does not mean that we render the same things differently in different parts of the world e.g. street colours.

CARTOGRAPHY.md
```diff
@@ -1,39 +1,39 @@
...
+* **A rich map** - This style deliberately creating a fairly rich map showing a significant number of different features.  This way it shows the richness of OSM data and gives a broad recognition to the mappers' work.  The aim is not however to show all or even most of the OSM data.
```
nebulon42: Extra spaces.

CARTOGRAPHY.md
```diff
@@ -1,39 +1,39 @@
...
+* **Maintainability** - The implementation of this style should not be too hard to maintain. This refers to both the volume and complexity of the code and how fast the style can be parsed when rendering it, which is very important for efficient development work. So the amount of code should be kept small and complex and fragile interdependencies should be avoided. If the code is difficult to maintain this would ultimately seriously affect all of the above goals.
```
nebulon42: A theme that comes up frequently is that it should not be too hard for new contributors to get started. Worth mentioning?

imagico:
Changed some formulations based on suggestions by @nebulon42.  What i did not change is

* sentence spacing - i am old fashioned, see https://en.wikipedia.org/wiki/Sentence_spacing.
* on locally differentiated rendering - that would belong in the more specific guidelines to follow.
* being easy for new contributors - While this is an important goal for the project in general i don't think this should be a cartographic goal.  We should help newcomers in learning to contribute productively but this is not an education project that adjusts its design goals and standards to the knowledge base of potential contributors.

imagico:
Thanks - part 2 is in #2462.

PR #2450, opened 2016-11-19
mention compound paths for icons

nebulon42:


CONTRIBUTING.md
```diff
@@ -144,6 +144,7 @@ Because SQL within JSON or YAML will not generally be syntax highlighted, indent
 * Icons must use SVG fills only, not SVG strokes or any feature Mapnik does not support.
 * Use no color for the icon's fill if the icon is monochromatic. This allows the color to be set in the MSS.
 * Use a common canvas size, which is usually 14x14 px.
+* Convert shapes and other components to paths and merge them into a compound path.
```
pnorman: We don't need compound paths unless there are holes in a desired shape. Also, I thought shapes like rect and circle worked fine.

nebulon42: It's not a hard requirement in the sense that without it the rendering won't work. But it saves us some potential trouble also with downstream usages and it makes us more independent of different rendering algorithms.

One concrete example: With adjacent shapes that do not overlap there is the possibility that a tiny portion of background shows trough. In German this is called "Blitzer". I have not seen this yet with Mapnik but with other renderers.

Maki has similar requirements: https://www.mapbox.com/maki-icons/guidelines/#specs They do not mention compound paths explicitly but the icons I looked at also had only one path in them.

PR #2451, opened 2016-11-20
Icon cleanup

### LINKED ISSUE #2062: use either name.size.ext or name-size.ext for all symbol files
extracted from https://github.com/gravitystorm/openstreetmap-carto/pull/2042#issuecomment-182739807

### LINKED ISSUE #2430: Width and height attributes on SVG symbols
SVG files contain several places where they have the width and height of the image. There is the `width` and `height` attributes as well as the `viewBox` attribute on the `<svg>` element. Some of the icons currently use `width="100%"` and `height="100%"` (for instance [bag-14.svg](https://raw.githubusercontent.com/gravitystorm/openstreetmap-carto/372032b80475b6b4ef2883254689f7e0c46b828b/symbols/bag-14.svg)) in addition to having the "real" size on the `viewBox`, some icons set a specific width (for instance [tea-14.svg](https://raw.githubusercontent.com/gravitystorm/openstreetmap-carto/372032b80475b6b4ef2883254689f7e0c46b828b/symbols/tea-14.svg)).

The Mapnik rendering seems not to care about the `width` and `height` attributes, the icons all render fine, but the icons are often imported into the OSM wiki for documentation and Mediawiki doesn't handle them equally well. In the wiki icons with the `100%` setting will appear as 512x512 pixel monsters unless the size is set explicitly. (I stumbled upon this in the context of the [taglists](https://wiki.openstreetmap.org/wiki/taginfo/taglists)). I think it would be easiest to change the icons at the source, ie this repository, instead of having to do something special for the wiki, unless there are some problems I am not seeing (@nebulon42 mentioned there are problems when I raised this on the talk mailinglist, but didn't elaborate).

(And yes, it is somewhat strange that we need to import those icons into the wiki instead of using them directly etc. etc., but I am trying to fix one problem at a time here.)

I'd be willing to make the changes to the icons if there is some agreement here.

nebulon42:
Fixes #2430 
Fixes #2062 

Various cleanup tasks regarding icons.

* remove standard size (14) from icon name
* set explicit width and height on icons (fixes #2430)
* clean up 16px icons, remove groups in some icons and bring some to 14px, add missing canvas to some icons as that makes it easier to customise the width/height downstream
* move shop icons in shop subfolder

Just for the record, we are currently rendering **49 different shop icons** (yes, quite some are from me).


HolgerJeromin:
Really nice and important work!

kocio-pl:
It also makes https://github.com/gravitystorm/openstreetmap-carto/issues/2062 obsolete.

Why some icons are renamed to include pixel size (like helipad.16.svg or post_box.12.svg)? This is not consistent with the rest (like pub.16.svg -> pub.svg).

nebulon42:
I have removed the standard size (14px) from the name, but left the non-standard sizes, so that it is easy to see which icons deviate from the common canvas size.

math1985:
+1

kocio-pl:
Oh, so I suspect pub.16.svg -> pub.svg name transition was done because the code was also fixed? If yes, I have no other issues with this PR.

@math1985 Technical question - is it ready to be merged as it is or the commits should be joined into just one?

nebulon42:
> Oh, so I suspect pub.16.svg -> pub.svg name transition was done because the code was also fixed?

Yes.

> or the commits should be joined into just one?

From my side I placed the commits deliberately into these chunks, so that changes could be left out, reverted and so on. But I guess it is fine either way. If we later want to change something it is easier if a part and not the whole bunch has to be changed.

math1985:
To me either is fine.

PR #2453, opened 2016-11-20
Make clear(er) Noto Emoji must be installed manually.

sommerluk:
Make clear(er) Noto Emoji must be installed manually.


math1985:
Thanks!

PR #2454, opened 2016-11-20
Increase strength of tertiary/residential/unclassified casing [WIP]

### LINKED ISSUE #2448: Increase tertiary/residential/unclassified casing strength
White roads (tertiary/residential/unclassified) should contrast sufficiently with features like landcover and buildings, with which they are frequently surrounded. To increase the contrast, I think we need to make the casing of these roads darker.

I created some exaggerated examples for the sake of argument (final choice will probably be less strong).
screenshot_2016-11-19_13-14-58
screenshot_2016-11-19_13-14-14
screenshot from 2016-11-19 13 12 54
screenshot_2016-11-19_13-13-26

For those who would like to test with this, see the math1985/stronger-casing branch.

matthijsmelissen:
White roads (tertiary/residential/unclassified) should contrast sufficiently with features like landcover and buildings, with which they are frequently surrounded. To increase the contrast, this pull request makes the casing of these roads darker.

Resolves #2448 (this is a toned down version of the changes proposed there).

Before/after (note that Github/your browser might scale the images so best to open them individually and/or press ctrl+0):

12-old 12-new

13-old 13-new

14-old 14-new

15-old 15-new

16-old 16-new

### DIFF
--- a/roads.mss
+++ b/roads.mss
@@ -25,10 +25,10 @@
 @access-marking-living-street: #cccccc;
 
 @default-casing: white;
-@tertiary-casing: #8f8f8f;
-@residential-casing: #bbb;
+@tertiary-casing: #777;
+@residential-casing: #777;
 @road-casing: @residential-casing;
-@service-casing: @residential-casing;
+@service-casing: #aaa;
 @living-street-casing: @residential-casing;
 @pedestrian-casing: #999;
 @path-casing: @default-casing;
@@ -198,16 +198,17 @@
 
 @major-casing-width-z11:          0.3;
 
-@casing-width-z12:                0.1;
+@casing-width-z12:                0.3;
 @secondary-casing-width-z12:      0.3;
 @major-casing-width-z12:          0.5;
 
 @casing-width-z13:                0.5;
-@residential-casing-width-z13:    0.5;
+@residential-casing-width-z13:    0.3;
 @secondary-casing-width-z13:      0.35;
 @major-casing-width-z13:          0.5;
 
 @casing-width-z14:                0.55;
+@residential-casing-width-z14:    0.4;
 @secondary-casing-width-z14:      0.35;
 @major-casing-width-z14:          0.6;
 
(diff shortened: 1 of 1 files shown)


nebulon42:
I have only looked at the examples here but I think the casing is too dark for its width on the first image. This leads to a jagged appearance. The second one is ok to me, could be lighter. On the last two I prefer lighter casing, could be a bit stronger than currently but not much.

I think the contrast is sufficient with residential and buildings. I'm not sure about other landcover, especially forest. Do you have another area in mind where you think the contrast is currently insufficient?

math1985:
>  the casing is too dark for its width on the first image

Are you sure you are watching the image on true size? I don't see this effect (except when watching downsized images)

nebulon42:
Yes, I see this also with the full size image.
I think the previews that you show do not necessarily require a change. Do you have another area where this change would improve rendering significantly?

kocio-pl:
I would also feel safer if we could see more different places.

math1985:
You can, just pull the branch :) No matter how many screenshots are post, the best impression you always get by just browsing an area of the map you're familiar with.

kocio-pl:
I'm not sure if it looks more readable on the farmland:
z12
Before
nalknb6e
After
9rgyyxtr

pnorman:
I haven't checked it out locally yet, but I find the first two pairs of images about the same in readability, the next two better, and the last one slightly better.

kocio-pl:
@nebulon42 Do you have any colors on your mind (proposed width seems OK to me)? It would make testing and deciding easier.

math1985:
I currently don't have time to work on this, but I'll get back to it when I have time again.

pnorman:
I checked it out locally. I'm positive overall and would be fine with merging, but have a few comments

- z14 to z15 seems a big big of a jump in casing strength
- z17+ seems to be a big improvement

The biggest problem seems to be AA. Perhaps line-gamma can be adjusted, or the line made wider but the casing faded out.

math1985:
Will look at it.

pnorman:

kocio-pl:
What is the current status of this code?

math1985:
I need to test if it still makes sense after the midzoom changes.

matkoniecz:
Sorry for not reviewing it earlier.

Are you planning to retest and rebase this PR?

math1985:
> Are you planning to retest and rebase this PR?

Yes. It's not waiting for reviews, but for work on my side.

IVAN-RSM:
I prefer before images xD

kocio-pl:
@matthijsmelissen What is your current plan with this code?

matkoniecz:
I think that it would be better to officialy close it for now. @matthijsmelissen Please send a new PR in case of a restarted work!

PR #2456, opened 2016-11-21
Mention that hstore should be installed

math1985:


INSTALL.md
```diff
@@ -3,6 +3,12 @@
 ## OpenStreetMap data
 You need OpenStreetMap data loaded into a PostGIS database (see below for [dependencies](#dependencies)). These stylesheets currently work only with the osm2pgsql defaults (i.e. database name is `gis`, table names are `planet_osm_point`, etc).
 
+Before loading the data, install the hstore extension into the database:
+
+```
+psql -d gis -c "CREATE EXTENSION hstore;"
+```
+
 It's probably easiest to grab an PBF of OSM data from [Mapzen](https://mapzen.com/metro-extracts/) or [geofabrik](http://download.geofabrik.de/). Once you've set up your PostGIS database, import with osm2pgsql:
```
pnorman: The instructions about extensions should replace "Once you've set up your PostGIS database'

kocio-pl: I also don't understand. IMO installing extensions is part of setting up the database, so importing is just what follows this step.

gravitystorm: The instructions are becoming a bit muddled up. "You need data loaded ... Before loading the data ... grab an PBF ... set up your PostGIS database ... import with osm2pgsql"

They should be re-written to describe what to do one step at a time, in order. Set up database, grab data, load data. 

So I think what Paul means is that the "Once you've set up your PostGIS database" bit is confusing, since by the time you get there you've already set up the database, i.e. installed hstore.

pnorman: > So I think what Paul means is that the "Once you've set up your PostGIS database" bit is confusing, since by the time you get there you've already set up the database, i.e. installed hstore.

Yes. 

1. Set up your database with ``psql -d gis -c 'CREATE EXTENSION postgis; CREATE EXTENSION hstore;'
2. Grab some data
3. Import with osm2pgsql

kocio-pl:
https://github.com/gravitystorm/openstreetmap-carto/pull/2486 took over, but thanks for your input!

PR #2459, opened 2016-11-22
Use json.dumps() for yaml2mml.py

### LINKED ISSUE #2447: yaml2mml not compatible with python 3.5
On my Arch Linux with python 3.5 yaml2mml.py fails. Probably because of #2225.

```
  File "./yaml2mml.py", line 27, in <module>
    json.dump(yaml, mml_file, indent=2, separators=(',', ': '))
  File "/usr/lib/python3.5/json/__init__.py", line 179, in dump
    fp.write(chunk)
TypeError: a bytes-like object is required, not 'str'
```

Steps to reproduce:
\> python --version
Python 3.5.2
\> pacman -Qi python-yaml
Version         : 3.12-1
\> git clone https://github.com/gravitystorm/openstreetmap-carto.git
\> ./scripts/yaml2mml.py
[Error above]

Python 2 works fine
\> python2 --version
Python 2.7.12
\> pacman -Qi python2-yaml
Version         : 3.12-1
\> python2 ./scripts/yaml2mml.py
[No error]

jojo4u:
Fixes #2447. Direct binary mode file output using json.dump() does not work with python 3. Other tricks like _newline_ parameter for _open()_ or using _io.open()_ where not successful. Best solution is to use json.dumps() and convert to bytes afterwards for python 2 & 3 compatibility.

Yields the same sha256sum for python2 and python3. The parameter _sort_keys=True_ has been introduced because python3 output had different sorting after every execution.

Not yet tested on Windows.

### DIFF
--- a/project.mml
+++ b/project.mml
@@ -1,2093 +1,2047 @@
 {
-  "interactivity": false,
   "Layer": [
     {
-      "name": "world",
-      "srs-name": "900913",
-      "geometry": "polygon",
-      "class": "",
-      "id": "world",
-      "srs": "+proj=merc +a=6378137 +b=6378137 +lat_ts=0.0 +lon_0=0.0 +x_0=0.0 +y_0=0.0 +k=1.0 +units=m +nadgrids=@null +wktext +no_defs +over",
       "Datasource": {
-        "type": "shape",
-        "file": "data/simplified-land-polygons-complete-3857/simplified_land_polygons.shp"
+        "file": "data/simplified-land-polygons-complete-3857/simplified_land_polygons.shp",
+        "type": "shape"
       },
+      "advanced": {},
+      "class": "",
       "extent": [
         -180,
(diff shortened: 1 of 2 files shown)


jojo4u:
from #2279:

> It also generates project.mml with Windows and with #2459 (just project.mml, without updating the script program). I see that Travis CI does not like it (possibly because the tags are not in the expected order).

What does "travis does not like" mean?

jojo4u:
Has been refined and now been tested with Linux and Windows. Python 2&3 on Linux and Windows are producing same output (with LF newline). Only exception is redirecting output of --check parameter to file, but this is not what this switch is for.

Ircama:
@jojo4u: it looks working with Windows (tested travis and tilemill)

nebulon42:
Obsolete by #2473. Thanks nevertheless!

Issue #2194, opened 2016-06-22
Improve lower mid zoom levels


math1985:
There is a lot of room to improve the lower mid zoom levels (z6 to z9). They come across as rather unpleasant and hard to read.

Some alternative renderings that do, in my opinion, a much better job:
https://geodienste.lyrk.de/
http://korona.geog.uni-heidelberg.de/
http://www.cartogiraffe.com/
http://maps.sputnik.ru/
http://osm2vectortiles.org/
I'm not sure if they're solely using osm-data for these zoom levels though.

I wonder what improvements we could make to make these zoom levels look better?

Cc @matkoniecz

math1985:
@matkoniecz also asked for examples from paper maps. In the Netherlands, the most well-known maps are the topographic maps from the government and the Bosatlas (an atlas).

I couldn't find much online from the Bosatlas, but here's at least a sample: https://s.s-bol.com/imgbase0/BOOKCOVER/FC/9/0/0/1/9/9001968996_2.gif

kocio-pl:
From my old [wiki page on low zoom](http://wiki.openstreetmap.org/wiki/User:Kocio/low_zoom_rework) there are still some related issues open:
* https://github.com/gravitystorm/openstreetmap-carto/issues/621
* https://github.com/gravitystorm/openstreetmap-carto/issues/1391
* https://github.com/gravitystorm/openstreetmap-carto/issues/1604

I am also interested if we plan to show some physical landcover there?

matkoniecz:
And some maps that ended on my list of "good mid zoom": http://mc.bbbike.org/mc/?lon=8.116896&lat=50.460645&zoom=5&num=6&mt0=mapbox-transportation&mt1=google-map&mt2=google-panoramio-physical&mt3=esri&mt4=mapnik&mt5=osm-roads&marker= (Esri, Google Maps, OSM Roads, Mapbox Transport).

kocio-pl:
> And some maps that ended on my list of "good mid zoom"

My 5 cents:

1. None of them is good at showing rivers - and we could try it.

2. "OSM FR hot" looks interesting to me - even from z5 (z4 is overcrowded with names). Where is the code for this style available? It has some issues with borders to fix, but with current roads from osm-carto (and rivers) middle zoom could be quite nice.

math1985:
I still like maps.sputnik.ru a lot. Does anybody know who is behind this rendering, and if their stylesheet is public?

[...]

Issue #1934, opened 2015-10-31
fountain rendering problems


imagico:
As already indicated in #1804 there are a number of problems with the new rendering of fountains.  Specifically:

* frequently at z17 only the water circle is rendered and not the dot
* the circle frequently covers other features in a very unpleasant and irritating way
* the icon frequently vanishes as you zoom from z17 to z18
* the water circle at z17 hides and obscures actual water mapping

Examples:

http://www.openstreetmap.org/#map=17/47.99291/7.85374

![z17](http://www.imagico.de/files/fountain_z17a.png)
![z18](http://www.imagico.de/files/fountain_z18a.png)

http://www.openstreetmap.org/#map=17/49.00819/8.40933

![z17](http://www.imagico.de/files/fountain_z17b.png)

kocio-pl:
Could you be more specific, so we can talk about precise cases? There are a lot of fountains in this area. For me it looks like few different problems with z17 mainly, some justified (which should be fixed if possible) and some just showing tagging problems (which is a good thing), so let's discuss more detailed examples.

imagico:
Not sure how i can be more specific - I listed the problems and all of them are visible in the images shown.  They are widespread, you should not have difficulties finding other examples in other places.

And i don't see how any of the issues mentioned can be considered a tagging problem - either there is a fountain at a certain point or not - other than mapping a fountain where there is none i don't see how tagging can be incorrect.

It seems by the way that some labels are rendered below the water circle, some above.  Unless you avoid overlap with labels altogether you have to render this below.  I am not sure if it is technically possible to do so *and* make sure it is only rendered if also the dot is rendered.

imagico:
Another good example for most of the problems in combination:

http://www.openstreetmap.org/#map=17/41.37072/2.15058

kocio-pl:
I mean I'd like to separate problems per object, not further combine them. Unfortunately I have very little time lately and less searching and guessing is preferable for me.

For example [this](http://www.openstreetmap.org/way/155005193) should be probably tagged as one fountain on the area instead of 2 points - if at all, because I don't see anything like nozzle on Bing. I think most of the time this artefact will disappear, because in area fountain icon/name will be displayed in the centre, not on the border. It's still possible that perfectly valid tagging could produce such artefacts, but that is true for everything, so if it will be a question of tagging, we don't have to resolve it here. But it takes a lot of precious time for me to draw such conclusion.

[...]

PR #1217, opened 2015-01-10
rendering of bare ground land covers

imagico:
This change adds support for rendering `natural=bare_rock` and `natural=scree` and thereby resolves #298 and partly addresses #545.

Rationale: `bare_rock` and `scree` are currently the most widely used values (65k and 24k uses) for the `natural` key that are not rendered in the standard style and also generally among the most widely used area tags that are not rendered.  Together with `sand` and undifferentiated loose material (for which no tagging has been established yet) they represent the primary types of natural landcovers in areas with no or very sparse natural vegetation like deserts, polar regions and high mountain areas.

The way this change renders these is somewhat different from the usual landcover rendering and likely not everyone will like it so i am going to explain my reasoning behind it:

* Use of plain colors for differentiating landcovers has reached the end of the line in this style - colors are already difficult to distinguish with existing landcover types and adding even more colors will just emphasize this problem.
* Therefore i decided to use the same color for all bare ground landcovers (`bare_rock`, `scree` and existing `sand`) and only differentiate between them at higher zooms by use of patterns.
* I tried to create intuitively understandable patterns, this is difficult with low resolution rendering like here, esp. for bare_rock.
* Colors chosen are fairly bright to avoid disturbing features drawn over them too much which seems important since esp. bare_rock and sand can cover large areas.
* The plain color rendering in the same color for all three starts at z=9, one zoom level after wood/forest but one before most other landcovers.  This factors in that these tend to cover large areas and since the color is not too much different from the map background this should not be too disturbing with smaller areas either.
* The pattern rendering starts at z=13, one zoom level before the forest/scrub but the pattern structure works quite well even on small areas and the pattern is the only thing differentiating between bare_rock/scree/sand.
* The previously used strong yellow color for sand is eliminated by this - IMO this is a good thing but some might prefer the yellow for smaller uses of natural=sand.  The most widespread is probably golf courses (36k out of 56k uses in total) but these should better be rendered based on golf tags anyway - which of course requires an osm2pgsql style change.
* In the code you can see that i included [natural=shingle](https://wiki.openstreetmap.org/wiki/Tag%3Anatural%3Dshingle) to be rendered like natural=scree to avoid mappers feeling additionally inclined to tag natural=scree despite this being technically wrong.  Practically natural=scree is currently used for all kinds of medium sized loose material.

Rendering examples (from Alps and Sinai):

Alps z=10
Alps z=13
Alps z=15

Sinai z=13
Sinai z=15

[...]

PR #2462, opened 2016-11-23
Proposal for design goals and guidelines - part 2

imagico:
Ok - here is the second part with more specific guidelines.

Not much to explain about it - right now this is essentially just a collection of ideas.  There is a general part focusing on ideas how to decide if and how to render things and two more specific subsections on colors and zoom levels.

### DIFF
--- a/CARTOGRAPHY.md
+++ b/CARTOGRAPHY.md
@@ -2,7 +2,7 @@
 # Design goals and guidelines for this style
 
 This is an attempt to outline the goals of this style and the principles under 
-which the maintainers make decisions.  These rules are not set in stone, they 
+which the maintainers make decisions. These rules are not set in stone, they 
 can change and they may not be followed in all cases but contributors should 
 be able to expect that they are generally the guiding principles design wise.
 
@@ -15,73 +15,143 @@ overall concept of this style.
 
 This style has multiple purposes:
 
-* It's an important feedback mechanism for mappers to validate their edits and helps to prevent unfavorable fragmentation of tag use.
-* It's a major part of the public face of OpenStreetMap, for many people the map on osm.org rendered with this style _is_ OpenStreetMap.
+* It's an important feedback mechanism for mappers to validate their edits and 
+  helps to prevent unfavorable fragmentation of tag use.
+* It's a major part of the public face of OpenStreetMap, for many people the 
+  map on osm.org rendered with this style _is_ OpenStreetMap.
 * It's used in many map applications as a general purpose map.
 * It's an exemplar stylesheet for rendering OSM data.
 
-There is no ranking of these purposes.  To allow serving all of them and to 
+There is no ranking of these purposes. To allow serving all of them and to 
 avoid satisfying only some at the cost of the others the following main goals 
 have been identified.
 
 ## Main goals
 
-The following goals need to be balanced against each other to serve the purposes 
-above.  There is no fixed order of priorities.  Apart from these goals there are 
-of course also technical constraints and requirements that need to be taken into 
-account.
+The following goals need to be balanced against each other to serve the 
+purposes above. There is no fixed order of priorities. Apart from these 
+goals there are of course also technical constraints and requirements that 
+need to be taken into account.
 
-* **Legibility and clarity** - The map should be intuitively readable by users with some general experience using maps without a map key, preferrably with relatively little effort.  A map key or more extensive experience using this map style can be required for clearly identifying minor differences or the exact meaning of certain features but in broad strokes orientation and identification of map elements should be possible on an intuitive level.  We also aim for the map appearance to be esthetically pleasing.
-* **Being understandable and supportive for mappers** - To serve as feedback for mappers and encourage correct mapping this style needs to render the data in a way that allows mappers to understand how the data produces a certain rendering result based on basic observation without in depth understanding how map rendering works or looking at the style implementation.
-* **Diversity** - The style should represent the diversity of the OSM community and geography in general.  The most obvious element to serve this goal is showing the local names everywhere on earth in their respective scripts.  This goal however goes beyond labels.  Both physical and cultural geography differs a lot globally and the aim is to represent this variety with equal determination - well mapped areas are not supposed to have more weight here than less mapped parts of the world.  This also means the target map user is the potential global map user and no special consideration is given to the current geographic distribution of actual map use.
-* **A rich map** - This style deliberately creating a fairly rich map showing a significant number of different features.  This way it shows the richness of OSM data and gives a broad recognition to the mappers' work.  The aim is not however to show all or even most of the OSM data.
-* **Maintainability** - The implementation of this style should not be too hard to maintain. This refers to both the volume and complexity of the code and how fast the style can be parsed when rendering it, which is very important for efficient development work. So the amount of code should be kept small and complex and fragile interdependencies should be avoided. If the code is difficult to maintain this would ultimately seriously affect all of the above goals.
-* **Adaptability and ease of use** - The style should be easy to customize, like for creating localized or special purpose maps. It is also important to keep demands on rendering infrastructure for serving the style low so it is not too difficult and costly to set up a tile server for this style or a specialized variant of it.
+* **Legibility and clarity** - The map should be intuitively readable by users 
+  with some general experience using maps without a map key, preferrably with 
+  relatively little effort. A map key or more extensive experience using this 
+  map style can be required for clearly identifying minor differences or the 
+  exact meaning of certain features but in broad strokes orientation and 
+  identification of map elements should be possible on an intuitive level. 
+  We also aim for the map appearance to be esthetically pleasing.
+* **Being understandable and supportive for mappers** - To serve as feedback 
+  for mappers and encourage correct mapping this style needs to render the 
+  data in a way that allows mappers to understand how the data produces a 
+  certain rendering result based on basic observation without in depth 
+  understanding how map rendering works or looking at the style 
+  implementation.
+* **Diversity** - The style should represent the diversity of the OSM 
+  community and geography in general. The most obvious element to serve this 
+  goal is showing the local names everywhere on earth in their respective 
+  scripts. This goal however goes beyond labels. Both physical and cultural 
+  geography differs a lot globally and the aim is to represent this variety 
+  with equal determination - well mapped areas are not supposed to have more 
+  weight here than less mapped parts of the world. This also means the target 
+  map user is the potential global map user and no special consideration is 
+  given to the current geographic distribution of actual map use.
+* **A rich map** - This style deliberately creating a fairly rich map showing 
+  a significant number of different features. This way it shows the richness 
+  of OSM data and gives a broad recognition to the mappers' work. The aim is 
+  not however to show all or even most of the OSM data.
+* **Maintainability** - The implementation of this style should not be too 
+  hard to maintain. This refers to both the volume and complexity of the code 
+  and how fast the style can be parsed when rendering it, which is very 
+  important for efficient development work. So the amount of code should be 
+  kept small and complex and fragile interdependencies should be avoided. If 
+  the code is difficult to maintain this would ultimately seriously affect 
+  all of the above goals.
+* **Adaptability and ease of use** - The style should be easy to customize, 
+  like for creating localized or special purpose maps. It is also important to 
+  keep demands on rendering infrastructure for serving the style low so it is 
+  not too difficult and costly to set up a tile server for this style or a 
+  specialized variant of it.
 
 ## Design guidelines
 
-The following guidelines have been found to be useful design principles that can help creating a better map in line with the above goals.  Keep in mind that none of these principles represents a goal on its own and most of them are highly condensed summaries of more complex considerations.
+The following guidelines have been found to be useful design principles that 
+can help creating a better map in line with the above goals. Keep in mind that 
+none of these principles represents a goal on its own and most of them are 
+highly condensed summaries of more complex considerations.
 
-* The difference in rendering between different types of features should foremost be based on their difference in meaning and purpose for the target map users.  Differences in physical appearance are only relevant as far as they also represent differences in meaning and purpose.
+* The difference in rendering between different types of features should 
+  foremost be based on their difference in meaning and purpose for the target 
+  map users. Differences in physical appearance are only relevant as far as 
+  they also represent differences in meaning and purpose.
 
-* Design of symbols, area and line patterns should be in a way that avoids their geometries to be mistaken as actual data.
+* Design of symbols, area and line patterns should be in a way that avoids 
+  their geometries to be mistaken as mapped geometries.
 
-* The connection between data and appearance of features in the map needs to be comprehensible for the viewer.
+* The connection between data and appearance of features in the map needs to 
+  be comprehensible for the viewer.
 
-* When changes have a strong impact on the overall appearance of the map input should be sought from the broader OpenStreetMap community and particular consideration given to their comments and concerns.
+* When changes have a strong impact on the overall appearance of the map input 
+  should be sought from the broader OpenStreetMap community and particular 
+  consideration given to their comments and concerns.
 
 ### Colors
 
-* Differentiating features through different colors only makes sense if this difference can actually be read by the map users.
+* Differentiating features through different colors only makes sense if this 
+  difference can actually be read by the map users.
 
-* For area colors we prefer light and low saturation tones.  Lines, individual symbols and labels also can be stronger in color.
+* For area colors we prefer light and low saturation tones. Lines, individual 
+  symbols and labels also can be stronger in color.
 
-* Stronger colors indicate fairly distinct and meaningful features as opposed to more general and less distinct features which use weaker colors.
+* Stronger colors indicate fairly distinct and meaningful features as opposed 
+  to more general and less distinct features which use weaker colors.
 
-* Small areas work well with somewhat stronger colors while large areas tend to call for weaker tones.
+* Small areas work well with somewhat stronger colors while large areas tend 
+  to call for weaker tones.
 
-* Symbols of area patterns need to be relatively weak in color with little contrast towards the base color so they cannot be mistaken for an individual symbol.
+* Symbols of area patterns need to be relatively weak in color with little 
+  contrast towards the base color so they cannot be mistaken for an individual 
+  symbol.
 
-* Area outlines should - when used - be in colors that are not too strong and harmonic with the fill color.  Care should be taken that area outlines are not mistaken for line features.
+* Area outlines should - when used - be in colors that are not too strong and 
+  harmonic with the fill color. Care should be taken that area outlines are 
+  not mistaken for line features.
 
-* Use of blue for area and line colors should be reserved for water related features.  Symbols and labels in blue color can also be used for other purposes but sparsely and in a tone clearly distinct from all water related features
+* Use of blue for area and line colors should be reserved for water related 
+  features. Symbols and labels in blue color can also be used for other 
+  purposes but sparsely and in a tone clearly distinct from all water related 
+  features.
 
-* Use of green should be reserved for vegetation related features, in particular in case of area and line colors.
+* Use of green should be reserved for vegetation related features, in 
+  particular in case of area and line colors.
 
-* Use of red, orange, yellow and brown line colors should be reserved to roads and paths.
+* Use of red, orange, yellow and brown line colors should be reserved to roads 
+  and paths.
 
-* Transparency tends to lead to undesirable and ambiguous color mixing and should therefore be used very carefully.
+* Transparency tends to lead to undesirable and ambiguous color mixing and 
+  should therefore be used very carefully.
 
-* It is generally advisable to design color relationships in perceptual color spaces.
+* It is generally advisable to design color relationships in perceptual color 
+  spaces.
 
 ### Zoom levels
 
-* Styling of map elements should work for the entire range of displayed latitudes and the full bandwidth of geographic situations that exists on earth.
+* Styling of map elements should work for the entire range of displayed 
+  latitudes and the full bandwidth of geographic situations that exists on 
+  earth.
 
-* Design differences between subsequent zoom levels should be small.  Continuous or small step changes over multiple levels is preferable in comparison to constant styling across multiple levels followed by a huge step.
+* Design differences between subsequent zoom levels should be small. 
+  Continuous or small step changes over multiple levels is preferable in 
+  comparison to constant styling across multiple levels followed by a huge 
+  step.
 
-* Size increase (line widths, text sizes, icon sizes etc.) with increasing zoom level should not be larger than the scale ratio between the zoom levels so sizes do not increase relative to ground units at higher zoom levels.
+* Size increase (line widths, text sizes, icon sizes etc.) with increasing 
+  zoom level should not be larger than the scale ratio between the zoom levels 
+  so sizes do not increase relative to ground units at higher zoom levels.
 
-* Features can start to appear at a certain zoom level when zooming in but they should not disappear again at a later zoom level - except possibly for labels of features becoming very large.
+* Features can start to appear at a certain zoom level when zooming in but 
+  they should not disappear again at a later zoom level - except possibly for 
+  labels of features becoming very large.
 
-* Starting zoom levels for showing features should be selected so that competition between elements does not result in the majority of features of a certain type not being shown if there is no meaningful measure of priority that is used for selection.
+* If there is no meaningful measure used for prioritization between individual 
+  features starting zoom levels should be selected to show at least most of 
+  the features.
--- a/CARTOGRAPHY.md
+++ b/CARTOGRAPHY.md
@@ -76,14 +76,12 @@ The following guidelines have been found to be useful design principles that can
 
 ### Zoom levels
 
-* Styling of map elements should always work for all parts of earth shown on the map and for the whole range of map scales present in each zoom level the feature is shown in.
+* Styling of map elements should work for the entire range of displayed latitudes and the full bandwidth of geographic situations that exists on earth.
 
-* Design differences between zoom levels should be small.
-
-* Continuous or small step changes over multiple levels is preferable in comparison to constant styling across multiple levels followed by a huge step.
+* Design differences between subsequent zoom levels should be small.  Continuous or small step changes over multiple levels is preferable in comparison to constant styling across multiple levels followed by a huge step.
 
 * Size increase (line widths, text sizes, icon sizes etc.) with increasing zoom level should not be larger than the scale ratio between the zoom levels so sizes do not increase relative to ground units at higher zoom levels.
 
 * Features can start to appear at a certain zoom level when zooming in but they should not disappear again at a later zoom level - except possibly for labels of features becoming very large.
 
-* Starting zoom levels for showing features should be selected so that competition between elements does not result in the majority of features of a certain type not being shown, especially if there is no meaningful measure of priority that is used for selection.
+* Starting zoom levels for showing features should be selected so that competition between elements does not result in the majority of features of a certain type not being shown if there is no meaningful measure of priority that is used for selection.
--- a/CARTOGRAPHY.md
+++ b/CARTOGRAPHY.md
@@ -40,64 +40,50 @@ account.
 
 ## Design guidelines
 
-To accomplish these goals the following guidelines have been found to be useful.
+The following guidelines have been found to be useful design principles that can help creating a better map in line with the above goals.  Keep in mind that none of these principles represents a goal on its own and most of them are highly condensed summaries of more complex considerations.
 
-Decisions on if and how to display something in the map should be primarily based on the significance of features for the purposes and goals above.
+* The difference in rendering between different types of features should foremost be based on their difference in meaning and purpose for the target map users.  Differences in physical appearance are only relevant as far as they also represent differences in meaning and purpose.
 
-* How often something is currently mapped in OSM has relatively little influence on such decisions.  Widespread use of a tag can be a supporting factor but should never be the reason for rendering something.  Tag use also needs to be seen in relation to the actual occurrence of the type of feature in reality.  And rare use of a tag of course means there is very limited information to assess the quality of the data.
-* If mapping in OSM (i.e. use of tags and geometric representation) is consistent, verifiable and accurate is highly relevant.  Rendering something that is mapped inconsistently is counterproductive for all of this style's purposes.  Consistency is to be looked for both in a positive way (mapping with the tag in question is consistent) and in a negative way (the same kind of object in reality is not frequently mapped in a different way).  Different competing tags can be accepted (to be rendered in the same way) if both variants are well established and the difference mainly exists for historic reasons and their use is compatible.
-* If tags used for mapping a certain feature type are well documented is likewise important because it affects if future use of these tags will likely be and stay consistent.   For similar reasons a feature that is already rendered in other OSM based maps with widespread use can be regarded more favorably than one that is not.
+* Design of symbols, area and line patterns should be in a way that avoids their geometries to be mistaken as actual data.
 
-Apart from the characteristics of the data the most important criterion for deciding on rendering a certain type of feature or not is how visible and meaningful this feature is in reality.  Prominently visible elements tend to be important for orientation even if they are not particularly important themselves for a map user.  Some features like for example an ATM might not be very visible but serve an important social function in their surrounding and should therefore be shown.
+* The connection between data and appearance of features in the map needs to be comprehensible for the viewer.
 
-If several similar types of features should be rendered identically or differently and which features should differ strongly in appearance and which only in minor nuances should foremost be based on their difference in meaning and purpose for the target map users.  Differences in physical appearance are only relevant as far as they also represent differences in meaning and purpose.
-
-Using data that is not from the OSM database should be avoided - unless it is strictly necessary for the goals.  Preprocessing of OSM based data can be and is used (like for the coastline) but should carefully be considered regarding the influence this has on mapper feedback.
-
-Design of symbols, area and line patterns should be in a way that avoids their geometries to be mistaken as actual data.
-
-How something is rendered should - as far as possible within the guidelines and goals outlined so far - be in line with mappers' expectations.  If the way something is rendered appears either ugly, too prominent, contradicting or diametrical to readability in general mappers will try to avoid mapping it this way and if another feature class appears to look more according to the expectations its tagging is often used instead (also known as mapping for the renderer).  Similar things happen if the way some feature is rendered is very generic, not particularly indicative and distinct for the class of features it is meant to represent. It then will often be abused to map other types of features.  For these reasons it can be more productive to not render something than to render it in a suboptimal way.
-
-Also important in terms of mapper feedback is that how data influences the appearance of features in the map needs to be comprehensible for the viewer.  Complex relationships and unexpected interactions between different features should be avoided.
-
-One particular consideration due to this style being an important part of the public face of OpenStreetMap is to be careful and considerate with changes having a strong impact on the overall appearance of the map, especially with elements that have been rendered a certain way for a long time.  Such changes should be broadly discussed also beyond the issue tracker and comments and concerns should be carefully considered.
+* When changes have a strong impact on the overall appearance of the map input should be sought from the broader OpenStreetMap community and particular consideration given to their comments and concerns.
 
 ### Colors
 
-Colors are a special and very important matter in map design and therefore get a separate section here.  In general since one of the goals is to create a rich map colors should be used sparsely to not affect readability.  And it only makes sense to differentiate features through different color if this difference can actually be read by the map users.
+* Differentiating features through different colors only makes sense if this difference can actually be read by the map users.
 
-As a general rule large areas should normally be colored in fairly light and low saturation colors.  Lines can be in stronger color although line styling can also be used as a means of differentiating here.  Individual symbols can also be stronger - likewise for labels.
+* For area colors we prefer light and low saturation tones.  Lines, individual symbols and labels also can be stronger in color.
 
-Symbols of area patterns should normally be relatively weak in color with little contrast towards the base color so they cannot be mistaken for an individual symbol.  They should however still be strong enough to be well readable.
+* Stronger colors indicate fairly distinct and meaningful features as opposed to more general and less distinct features which use weaker colors.
 
-Within the different fields of color use (like areas, lines, symbols and labels) stronger and weaker colors can be used to indicate fairly distinct and meaningful features as opposed to more general and less distinct features.  For areas also consider that small areas often work better with stronger colors while large areas tend to call for weaker tones.
+* Small areas work well with somewhat stronger colors while large areas tend to call for weaker tones.
 
-Area outlines should - when used - be in colors that are not too strong and harmonic with the fill color.  Care should be taken that area outlines are not mistaken for line features.
+* Symbols of area patterns need to be relatively weak in color with little contrast towards the base color so they cannot be mistaken for an individual symbol.
 
-The goal of the map being readable based on general map reading experience dictates adhering to general conventions of color use in maps to some extent.  Such conventions also often represent sensible choices regarding color physiology in general.  The following principles should generally be followed:
+* Area outlines should - when used - be in colors that are not too strong and harmonic with the fill color.  Care should be taken that area outlines are not mistaken for line features.
 
 * Use of blue for area and line colors should be reserved for water related features.  Symbols and labels in blue color can also be used for other purposes but sparsely and in a tone clearly distinct from all water related features
+
 * Use of green should be reserved for vegetation related features, in particular in case of area and line colors.
+
 * Use of red, orange, yellow and brown line colors should be reserved to roads and paths.
 
-Of course there is a continuum of colors and it is not quite clear where green and blue end.  The important thing here is that water related blues and vegetation related greens form identifiable units in color space that allow intuitively identifying these groups of colors when reading the map.
+* Transparency tends to lead to undesirable and ambiguous color mixing and should therefore be used very carefully.
 
-It is an important general principle that features differing strongly in meaning and purpose for the map user should be well differentiated in either color or by other means while fairly similar features should differ little or not at all.  Or in other words: features rendered with similar color should actually be similar in purpose and meaning for the map user.
-
-Transparency should be used very carefully with consideration for the color mixing this can result in and the ambiguities possibly caused by it.
-
-It is generally advisable to design color relationships in perceptual color spaces.  Use of internal color functions like `lighten()` and `darken()` should be considered carefully since they are performed in sRGB.
-
+* It is generally advisable to design color relationships in perceptual color spaces.
 
 ### Zoom levels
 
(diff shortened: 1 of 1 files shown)


math1985:
Looks good.

> The only situation where this can be appropriate is in case of labels for features that get very large and where labeling them is therefore no more useful.

This could be a bit less restrictive, for example: "The only situation where this can be appropriate is in case of features that get very large (for example larger than a typical monitor) and where labeling or drawing them is therefore no more useful."

> Decisions at what zoom level to start showing a certain feature type should prominently consider that not actually rendering a large number of elements because of competing labels or icons is often diametrical to the purpose of providing useful mapper feedback, in particular if feature priorities are not based on a meaningful property.

This sentence is not clear to me.

Ideally, it would also be good if somebody that is not so familiar with our processes could read over this and give feedback.

nebulon42:
cc @k127

imagico:
>> Decisions at what zoom level to start showing a certain feature type should prominently consider that not actually rendering a large number of elements because of competing labels or icons is often diametrical to the purpose of providing useful mapper feedback, in particular if feature priorities are not based on a meaningful property.

> This sentence is not clear to me.

Yes, that probably could me made clearer.  What i had in mind here is things like dense city centers with shops etc.  If only like half of the icons that could be shown are actually shown due to the lack of space and there is no meaningful prioritization to select which are shown that leads to arbitrary and therefore undesirable results.

We could also remove that because it is actually just a special case of what is written in the general part already:

> Also important in terms of mapper feedback is that how data influences the appearance of features in the map needs to be comprehensible for the viewer. Complex relationships and unexpected interactions between different features should be avoided.

> Ideally, it would also be good if somebody that is not so familiar with our processes could read over this and give feedback.

Probably a good idea.

CARTOGRAPHY.md
```diff
@@ -37,3 +37,67 @@ account.
...
+* Use of red, orange, yellow and brown line colors should be reserved to roads and paths.
+
+Of course there is a continuum of colors and it is not quite clear where green and blue end.  The important thing here is that water related blues and vegetation related greens form identifiable units in color space that allow intuitively identifying these groups of colors when reading the map.
+
+It is an important general principle that features differing strongly in meaning and purpose for the map user should be well differentiated in either color or by other means while fairly similar features should differ little or not at all.  Or in other words: features rendered with similar color should actually be similar in purpose and meaning for the map user.
+
+Transparency should be used very carefully with consideration for the color mixing this can result in and the ambiguities possibly caused by it.
+
+It is generally advisable to design color relationships in perceptual color spaces.  Use of internal color functions like `lighten()` and `darken()` should be considered carefully since they are performed in sRGB.
```
imagico: Do you say that behavior of these functions has changed and rendering results with recent carto versions are not backwards compatible?

imagico: Ah, that is - well - unexpected.  What happens if you mix rgb and husl: `mix(#fff, husl(100, 50%, 50%), 50%)`?

pnorman: If we decided to use husl we'd have to revisit this, but as it stands, the text is accurate.

imagico:
I am a bit surprised that there is not much objection to the specific guidelines proposed here.  They clash with quite a few design decisions we currently have in the style and while likely everyone agrees there is a lot of room for improvements in the current design i did not expect that everyone agrees where these deficits are.

Or maybe i just phrased the more controversial points in ways that are difficult to understand.

Of course it is also clear that just because some aspects of the style currently are at odds with some fairly fundamental and broadly accepted design principles that does not necessarily mean there is an easy way to fix this without causing new problems.

CARTOGRAPHY.md
```diff
@@ -37,3 +37,67 @@ account.
...
 * **Adaptability and ease of use** - The style should be easy to customize, like for creating localized or special purpose maps. It is also important to keep demands on rendering infrastructure for serving the style low so it is not too difficult and costly to set up a tile server for this style or a specialized variant of it.
+
+## Design guidelines
+
+To accomplish these goals the following guidelines have been found to be useful.
+
+Decisions on if and how to display something in the map should be primarily based on the significance of features for the purposes and goals above.
+
+* How often something is currently mapped in OSM has relatively little influence on such decisions.  Widespread use of a tag can be a supporting factor but should never be the reason for rendering something.  Tag use also needs to be seen in relation to the actual occurrence of the type of feature in reality.  And rare use of a tag of course means there is very limited information to assess the quality of the data.
```
pnorman: 👎 for this phrasing. We want to exclude PRs for &lt;new feature with no uptakeyet  from users&gt;

CARTOGRAPHY.md
```diff
@@ -37,3 +37,67 @@ account.
 * **A rich map** - This style deliberately creating a fairly rich map showing a significant number of different features.  This way it shows the richness of OSM data and gives a broad recognition to the mappers' work.  The aim is not however to show all or even most of the OSM data.
 * **Maintainability** - The implementation of this style should not be too hard to maintain. This refers to both the volume and complexity of the code and how fast the style can be parsed when rendering it, which is very important for efficient development work. So the amount of code should be kept small and complex and fragile interdependencies should be avoided. If the code is difficult to maintain this would ultimately seriously affect all of the above goals.
 * **Adaptability and ease of use** - The style should be easy to customize, like for creating localized or special purpose maps. It is also important to keep demands on rendering infrastructure for serving the style low so it is not too difficult and costly to set up a tile server for this style or a specialized variant of it.
+
+## Design guidelines
```
pnorman: I think this entire section is too long. https://github.com/gravitystorm/openstreetmap-carto/issues/1630#issuecomment-118915756 seems relevant, and the problems with your suggestions there seem repeated here.

CARTOGRAPHY.md
```diff
@@ -37,3 +37,67 @@ account.
...
+
+* How often something is currently mapped in OSM has relatively little influence on such decisions.  Widespread use of a tag can be a supporting factor but should never be the reason for rendering something.  Tag use also needs to be seen in relation to the actual occurrence of the type of feature in reality.  And rare use of a tag of course means there is very limited information to assess the quality of the data.
+* If mapping in OSM (i.e. use of tags and geometric representation) is consistent, verifiable and accurate is highly relevant.  Rendering something that is mapped inconsistently is counterproductive for all of this style's purposes.  Consistency is to be looked for both in a positive way (mapping with the tag in question is consistent) and in a negative way (the same kind of object in reality is not frequently mapped in a different way).  Different competing tags can be accepted (to be rendered in the same way) if both variants are well established and the difference mainly exists for historic reasons and their use is compatible.
```
pnorman: I get **some** of what this is trying to say, having worked with landuse=forest and natural=wood. I don't think most people will understand it.

I'm not sure if I agree with it as phrased.

CARTOGRAPHY.md
pnorman: I'd set a target of about a third of the current wordcount for this section, less if possible.

CARTOGRAPHY.md
```diff
@@ -37,3 +37,67 @@ account.
...
+
+Design of symbols, area and line patterns should be in a way that avoids their geometries to be mistaken as actual data.
+
+How something is rendered should - as far as possible within the guidelines and goals outlined so far - be in line with mappers' expectations.  If the way something is rendered appears either ugly, too prominent, contradicting or diametrical to readability in general mappers will try to avoid mapping it this way and if another feature class appears to look more according to the expectations its tagging is often used instead (also known as mapping for the renderer).  Similar things happen if the way some feature is rendered is very generic, not particularly indicative and distinct for the class of features it is meant to represent. It then will often be abused to map other types of features.  For these reasons it can be more productive to not render something than to render it in a suboptimal way.
```
pnorman: I'm not sure what "be in line with mappers' expectations" means here.

CARTOGRAPHY.md
```diff
@@ -37,3 +37,67 @@ account.
...
+
+Apart from the characteristics of the data the most important criterion for deciding on rendering a certain type of feature or not is how visible and meaningful this feature is in reality.  Prominently visible elements tend to be important for orientation even if they are not particularly important themselves for a map user.  Some features like for example an ATM might not be very visible but serve an important social function in their surrounding and should therefore be shown.
+
+If several similar types of features should be rendered identically or differently and which features should differ strongly in appearance and which only in minor nuances should foremost be based on their difference in meaning and purpose for the target map users.  Differences in physical appearance are only relevant as far as they also represent differences in meaning and purpose.
+
+Using data that is not from the OSM database should be avoided - unless it is strictly necessary for the goals.  Preprocessing of OSM based data can be and is used (like for the coastline) but should carefully be considered regarding the influence this has on mapper feedback.
```
pnorman: I thought this was covered already

CARTOGRAPHY.md
```diff
@@ -37,3 +37,67 @@ account.
...
+
+Also important in terms of mapper feedback is that how data influences the appearance of features in the map needs to be comprehensible for the viewer.  Complex relationships and unexpected interactions between different features should be avoided.
+
+One particular consideration due to this style being an important part of the public face of OpenStreetMap is to be careful and considerate with changes having a strong impact on the overall appearance of the map, especially with elements that have been rendered a certain way for a long time.  Such changes should be broadly discussed also beyond the issue tracker and comments and concerns should be carefully considered.
```
pnorman: We don't need to discuss beyond the issue tracker, but instead inform people that a change is being considered and that discussion is taking place on the issue tracker.

CARTOGRAPHY.md
```diff
@@ -37,3 +37,67 @@ account.
...
+* Many map features start being displayed at a certain zoom level and continue being shown, sometimes with changing styling, at subsequent higher levels.  What is highly undesirable though is to show a feature first and then have it vanish again at a higher zoom level.  The only situation where this can be appropriate is in case of labels for features that get very large and where labeling them is therefore no more useful.
+* Decisions at what zoom level to start showing a certain feature type should prominently consider that not actually rendering a large number of elements because of competing labels or icons is often diametrical to the purpose of providing useful mapper feedback, in particular if feature priorities are not based on a meaningful property.
+* Decisions at what zoom level to start showing something are also particularly prone to causing bad mapping decisions (mapping for the renderer).  This should be considered when making choices here.
+
+
```
pnorman: there's an extra newline here

kocio-pl:
I didn't have the time to read such a big chunk of text yet. It's hard for me to comment it, because it's so high level and there are so many words, that it even just reading "Design guidelines" was substantial effort and I leave the rest for later. I'm not able to get the whole picture - let alone have my opinion on every paragraph. This might not be a problem in itself - this change describes your vision and it can make people think in a certain way. I'm interested more in how such vision is applied to a map and this is just a different level.

Currently I can only point one problem - I don't get what's the meaning of this sentence, so I'd like it to be rephrased:
> Design of symbols, area and line patterns should be in a way that avoids their geometries to be mistaken as actual data.

From my point of view it would be better if this text is much shorter (to avoid people TL;DR-ing) and instead have some examples. Current form sounds like a blog post - I like to read such things from time to time as a food for thought (including your blog). If you want it to be casual help in case of doubt, this is OK for me, but if you want people to follow it in everyday activities, I think it needs to be seriously edited (for example by a fellow journalist). It depends on how do you expect people to use it.

CARTOGRAPHY.md
```diff
@@ -37,3 +37,67 @@ account.
...
+* Many map features start being displayed at a certain zoom level and continue being shown, sometimes with changing styling, at subsequent higher levels.  What is highly undesirable though is to show a feature first and then have it vanish again at a higher zoom level.  The only situation where this can be appropriate is in case of labels for features that get very large and where labeling them is therefore no more useful.
+* Decisions at what zoom level to start showing a certain feature type should prominently consider that not actually rendering a large number of elements because of competing labels or icons is often diametrical to the purpose of providing useful mapper feedback, in particular if feature priorities are not based on a meaningful property.
+* Decisions at what zoom level to start showing something are also particularly prone to causing bad mapping decisions (mapping for the renderer).  This should be considered when making choices here.
```
pnorman: Not clear what this means

CARTOGRAPHY.md
```diff
@@ -37,3 +37,67 @@ account.
...
+
+Transparency should be used very carefully with consideration for the color mixing this can result in and the ambiguities possibly caused by it.
+
+It is generally advisable to design color relationships in perceptual color spaces.  Use of internal color functions like `lighten()` and `darken()` should be considered carefully since they are performed in sRGB.
+
+
+### Zoom levels
+
+An important aspect of styles for interactive maps is that they are used for a large number of different map sizes or zoom levels.  This creates special challenges for map design.  A particular problem with this stems from the fact that the Mercator projection is a variable scale projection so the different zoom levels do not represent specific map scales.  This fact should always be considered when making design decisions.  Styling should always work for the whole map and for the whole range of map scales present in the zoom level in question.  This normally requires testing changes in different areas at different latitudes.
```
pnorman: I'd pull this paragraph out and put in something in general about the style needing to work around the world, city density is often relevant, and in the mercator projection density can change with latitude. I'm not sure what section that would go in.

CARTOGRAPHY.md
nebulon42: @imagico when there is a husl colour present the calculations happen in husl colour space

nebulon42:
@kocio-pl https://github.com/gravitystorm/openstreetmap-carto/issues/1753#issue-101242802 gives an example of this sentence.

pnorman:
> Current form sounds like a blog post - I like to read such things from time to time as a food for thought (including your blog).

Yes, as I read through it, it reminded me of a first draft of one my blog posts when I have all the content I want in, but haven't started to remove stuff.

imagico:
Ok, it seems the current draft is widely considered hard to read and understand.  This can probably be improved.  I was specifically not trying to make this a list of short bullet points of the form _do this_, _don't do that_ because in my eyes this is ultimately of little  help since good design cannot be condensed into a set of simple rules.  Therefore i deliberately tried to include the reasoning behind and the interrelations between the principles listed which probably makes it fairly abstract.

I would appreciate it if you could make some concrete suggestions how to formulate things in a more compact and better understandable way.  I don't think i would be able to compress this to something like 1/3 the volume without removing a lot of things i consider important.  But i am open to the idea of a more compact document as long as this still represents those things we consider the guiding principles here.

CARTOGRAPHY.md
imagico: Mappers frequently choose incorrect tagging to push features to a lower zoom level.  Like for example by tagging remote hamlets/villages as place=city.  Such things cannot always be fully avoided but they should be considered when making styling decisions IMO.

CARTOGRAPHY.md
imagico: I disagree here - Because of the map style being the face of OSM in many ways we need to be willing to discuss significant changes to this with the OSM community on a larger scale.  @matkoniecz did this quite well with the road color changes and if we should have other changes with similar impact in the future (boundary color and water color come to mind here) we should broaden the discussion on these as well.  This does not mean decisions should be made elsewhere but opinions should be considered even if they have only been articulated abroad.

CARTOGRAPHY.md
imagico: Where?

CARTOGRAPHY.md
sommerluk: I think its difficult to make this it a hard requierement to broadly survey all communication channels for discussions (also when looking at the relation between workload for this and possible outcome). I support what @pnorman has sad.

CARTOGRAPHY.md
imagico: But do we have a clear guideline on what basis we decide this?  I thought about this point when drafting the text and i could not really find such a principle in past decisions.

CARTOGRAPHY.md
imagico: In this case considering the population tag would probably help.  But you have to be careful of course so the rendering rules are still comprehensible for mappers.

To be frank - if we right now had the decision to make if and how to render place=city/town/village/hamlet i would probably be against differentiated rendering at all because the wiki lacks a clear definition of the tags and the data is globally quite inconsistent on the matter.  place =city essentially just means here the mapper thought this is a place that should be shown significantly earlier and more prominently than other populated places.

math1985:
> Much less text. Right now it's ~1.5k words, I'd like under 500. 

I agree with this as well, I also found the text long and hard to get through.

kocio-pl:
I understand the need to learn people to think about cartographic design, but style documentation is not a good place for general education. So maybe here we should have only guidelines in a nutshell (like bullet points), but you could use the current text as a base for an article - or even series of articles - about the reasoning behind it?

imagico:
As said i understand the desire to make it shorter but at the moment i don't see how i can compress it to about 1/3 the volume while still keeping it a comprehensible guideline for contributors.  So i could use some more specific suggestions how to formulate things in a compact yet understandable way.

kocio-pl:
That's why I suggested help from someone with journalistic background - I know how hard is it to be an author and editor at the same time.

math1985:
I liked the three bullet points in another issue from which this discussion started. It leaves a lot of rationale implicit, but to me that's fine.

I also liked the style of the original documentation.

imagico:
Ok - i thought a bit more about this and it seems quite clear to me that the approach i took, writing things down in the way i see them and trying to develop this into a common view on the design principles, won't work.  It is quite possible that if someone else with a different point of view tries the same this could work better, approaching a consensus position both design wise and in terms of writing style from a different side.

Independent from that i will try a different approach now, just formulating the individual design principles i consider most useful and important, in a very compact form.  In my eyes this will probably be of rather limited help for contributors as practical guidelines.  But it could make it easier for us to find common principles we agree upon and that is likely of significant help per se for contributors.

I also removed those points which i consider less important because they are either generally accepted anyway, rarely turn up in practical work, are special cases of already mentioned more general principles or where we will likely not reach a common position (like rules what to render and what not to render).

If that is fine with you I will add consistent line wrapping and change sentence spacing if and when we have a version ready for merging.

@math1985 - regarding your early suggestion for

> The only situation where this can be appropriate is in case of features that get very large (for example larger than a typical monitor) and where labeling or drawing them is therefore no more useful.

Could you point to a practical case where you consider dropping elements other than labels from rendering at higher zoom level is desirable?

math1985:
> Could you point to a practical case where you consider dropping elements other than labels from rendering at higher zoom level is desirable?

We currently drop the national park fill from z10. I'm not sure whether we should adapt the rendering rules to the guidelines or the guidelines to the rendering rules here.

imagico:
> The only example I know: we currently drop the national park fill from z10. I'm not sure whether we should adapt the rendering rules to the guidelines or the guidelines to the rendering rules here.

To me that is a styling difference between zoom levels (rendering fill+outline vs. outline only).  But national parks do not fully disappear.  The outline only rendering can of course be highly confusing, especially if there are several touching or nested features of this type.

math1985:
Then I'm fine with the original formulation.

kocio-pl:
Now it's much better, thanks. My _Zoom levels_ review - I think that points 2 and 3 could be merged:

> Design differences between zoom levels should be small. Continuous or small step changes over multiple levels is preferable in comparison to constant styling across multiple levels followed by a huge step.

imagico:
I see no gain in this but i have no objections either.

kocio-pl:
> Use of green should be reserved for vegetation related features

What about sport then?

pnorman:
I did an initial skim, and it looks much better. I still need to do a detailed review.

imagico:
>> Use of green should be reserved for vegetation related features

> What about sport then?

I guess that depends on if you consider that a vegetation related feature. :smile:

In the original long version i had quite a few additional explanations on how this is meant.

In any case keep in mind that neither the goals nor the more specific design principles we discuss now are conflict free.  As i explained elsewhere the purpose here is not to slavishly obey rules but to make conscious decisions considering the different aspects that play a role.  What we should ask ourselves is if these ideas are generally useful as guiding principles.

matkoniecz:
> I am a bit surprised that there is not much objection to the specific guidelines proposed here. They clash with quite a few design decisions we currently have in the style and while likely everyone agrees there is a lot of room for improvements in the current design i did not expect that everyone agrees where these deficits are.

For me, in cases where I see conflict (like " +* Use of red, orange, yellow and brown line colors should be reserved to roads and paths." vs linear outline of theme parks/zoos) I recognize that it is unfortunate but I see no ready solution to improve it and map would not get better from dropping rendering of that feature.

CARTOGRAPHY.md
```diff
@@ -37,3 +37,53 @@ account.
...
+
+* Transparency tends to lead to undesirable and ambiguous color mixing and should therefore be used very carefully.
+
+* It is generally advisable to design color relationships in perceptual color spaces.
+
+### Zoom levels
+
+* Styling of map elements should always work for all parts of earth shown on the map and for the whole range of map scales present in each zoom level the feature is shown in.
+
+* Design differences between zoom levels should be small.
```
matkoniecz: "Design differences between adjacent zoom levels should be small."

I am not sure about exact word (adjacent/joining/neighbouring are all really bad), but it should be clear that z5 and z19 may have vast differences in styling and it is OK. It is partially clarified in the next line but it would be better to make it completely clear.

CARTOGRAPHY.md
```diff
@@ -37,3 +37,53 @@ account.
...
+
+* Continuous or small step changes over multiple levels is preferable in comparison to constant styling across multiple levels followed by a huge step.
+
+* Size increase (line widths, text sizes, icon sizes etc.) with increasing zoom level should not be larger than the scale ratio between the zoom levels so sizes do not increase relative to ground units at higher zoom levels.
+
+* Features can start to appear at a certain zoom level when zooming in but they should not disappear again at a later zoom level - except possibly for labels of features becoming very large.
+
+* Starting zoom levels for showing features should be selected so that competition between elements does not result in the majority of features of a certain type not being shown, especially if there is no meaningful measure of priority that is used for selection.
```
matkoniecz: I would remove "especially". With clear priority it is perfectly fine to start from displaying only some. First labels for countries appear on z3, first cities appear on z4 etc.

It is issue for things like shops where it is really hard to get priority.

CARTOGRAPHY.md
```diff
@@ -37,3 +37,53 @@ account.
...
+
+* Area outlines should - when used - be in colors that are not too strong and harmonic with the fill color.  Care should be taken that area outlines are not mistaken for line features.
+
+* Use of blue for area and line colors should be reserved for water related features.  Symbols and labels in blue color can also be used for other purposes but sparsely and in a tone clearly distinct from all water related features
+
+* Use of green should be reserved for vegetation related features, in particular in case of area and line colors.
+
+* Use of red, orange, yellow and brown line colors should be reserved to roads and paths.
+
+* Transparency tends to lead to undesirable and ambiguous color mixing and should therefore be used very carefully.
```
matkoniecz: Is there any case where we are using transparency? I think that stronger wording ("should not be used" or similar) would better describe our approach to it.

CARTOGRAPHY.md
```diff
@@ -37,3 +37,53 @@ account.
...
+
+* It is generally advisable to design color relationships in perceptual color spaces.
+
+### Zoom levels
+
+* Styling of map elements should always work for all parts of earth shown on the map and for the whole range of map scales present in each zoom level the feature is shown in.
```
matkoniecz: "should always work" is unfortunately too ambitious. I am pretty sure that for any feature one may find well mapped places where due to some special case map is not really working.

"should work for the entire range of displayed latitudes"?

CARTOGRAPHY.md
imagico: The most prominent use of transparency is currently the boundaries.  This is especially problematic since they are in strong color leading to strange color mixing.  Other uses are nature reserves and military areas.

CARTOGRAPHY.md
imagico: _subsequent_ is probably the right word here.

matkoniecz: Yes, 'subsequent' is what I wanted to propose.

CARTOGRAPHY.md
matkoniecz: Yes, way_area works not as often as one may expect and cities are unique with population as objective ranging information that is usually available.

kocio-pl:
> For me, in cases where I see conflict (like " +* Use of red, orange, yellow and brown line colors should be reserved to roads and paths." vs linear brown outline of theme parks/zoos) I recognize that it is unfortunate but I see no ready solution to improve it and map would not get better from dropping rendering of that feature.

Maybe the wording should be changed like "should be reserved for" -> "should be primarily used for" to better reflect the intention expressed in [this comment](https://github.com/gravitystorm/openstreetmap-carto/pull/2462#issuecomment-263636898)?

imagico:
> Maybe the wording should be changed like "should be reserved for" -> "should be primarily used for" 

No, if that is the general feeling we should remove those altogether.  _should be primarily used for_ is not really helpful as a guiding principle since an individual change will hardly ever change what a certain range of colors is primarily used for.

> Use of red, orange, yellow and brown line colors should be reserved to roads and paths." vs linear brown outline of theme parks/zoos)

I generally dislike this kind of outline rendering independent of the color because it adds a lot of noise and bloat to the map with very little information and is only readable across a very limited range of scales.  The main motivation for using it to me seems indeed to be able to overload colors by using them with a fairly distinct line style.

> I recognize that it is unfortunate but I see no ready solution to improve it and map would not get better from dropping rendering of that feature.

Well - we cannot solve all problems at once but recognizing where the issues are is the first step in solving them.  Dropping features that have been rendered for a long time is going to be tough but it is for example not that far fetched to say that something like nature reserves can also be rendered in a separate overlay and therefore do not absolutely need to be in the main map.

kocio-pl:
> No, if that is the general feeling we should remove those altogether. should be primarily used for is not really helpful as a guiding principle since an individual change will hardly ever change what a certain range of colors is primarily used for.

OK, makes sense for me.

imagico:
I integrated the changes suggested as far as there was agreement i think.  Further suggestions are welcome.

kocio-pl:
I still don't understand this sentence:

> Design of symbols, area and line patterns should be in a way that avoids their geometries to be mistaken as actual data.

Well, we probably like them to be similar to actual data, for the sake of "legibility and clarity" we state in the same document. Given the example with tree row looking like a hedge, maybe you mean something like:

> Design of symbols, area and line patterns should not suggest to represent some other features.

I'm not even sure if I get it right, maybe you could explain - in simple words - what do you want to say here.

imagico:
> Well, we probably like them to be similar to actual data, for the sake of "legibility and clarity" we state in the same document.

No, it helps general legibility if a pattern/symbol can be intuitively connected to a real world feature but it is very damaging for readability if the geometry of a symbol can easily be mistaken for a mapped geometry.  The key here is conceptual separation between abstract symbols and mapped geometries.

Two examples: We had some discussion on this matter with the bare_rock pattern (#1217) - this might actually still be a problem with very small grained mapping of such areas although practically i have not really seen an example where it is.

Another example you know quite well - the fountain symbol (#1934).  Here there is no way to distinguish between the abstract symbol and a mapped geometry (circular pool) making it a clear case where this separation does not exist.

Line dashing can - if it is relatively coarse compared to the level of detail in the geometry - also be prone to this kind of problem, visible for example with low zoom rendering of intermittent waterways.

Methods to accomplish this are mainly careful choice of colors and suitable design of symbols and patterns.

kocio-pl:
Wording of this sentence is a problem in itself for me, so - as far as I understand you - it means something like:

> Design of symbols, area and line patterns should not mimic the physical appearance of the features.

(I would not use "actual data" anyway, because for me it sounds like tags in db, which is absurd, but doesn't hint what we're talking about)

I would not agree with such statement. Two examples:

1. Grass rendering looks exactly like the typical grass area picture taken from above, where details are not seen, and this is not a problem (of course it's rather about the color, not the shape, but the effect is the same).

2. Our tree rendering looks very close to a typical tree and how do you know it's an abstract symbol and not a real tree visualization with brown trunk and a green canopy with specific diameter? I don't see the clear separation between symbol and the look here - it looks similar and you could be mistaken - but so what?

math1985:
> (I would not use "actual data" anyway, because for me it sounds like tags in db, which is absurd, but doesn't hint what we're talking about)

That is indeed the intention though: a random pattern should not look like it is displaying actual information from objects in the database. Our forest pattern is a good example: map readers should immediately realize that the trees are just symbolic, rather than think that each tree in the pattern represents an actual tree in the database.

kocio-pl:
I probably understand now your examples, but what do you think about my examples then? Are they breaking this rule and ideally should be replaced?

And what is the problem if a viewer can't separate the symbol from a data - what do we loose this way?

imagico:
> Design of symbols, area and line patterns should not mimic the physical appearance of the features.

That is not what i want to express here.  Since a map is an abstract depiction design does not need to mimic physical appearance in reality but there is no need to expressively avoid that.

> (I would not use "actual data" anyway, because for me it sounds like tags in db, which is absurd, but doesn't hint what we're talking about)

Actual data here essentially means mapped geometries - i can substitute that if preferred.  Any map shows shapes that represent actual real world geometries - a rectangle that represents a building, a curved line representing a river and similar things.  But most maps also show shapes that do not represent a real world geometry - like the tree symbols @math1985 mentioned.  The map styling should aim to make sure it is always clear which type of representation a certain drawn shape is.

If a map reader cannot make this distinction this means loss of the ability to correctly interpret the map (like '_in this area trees grow in pairs_' or '_there are a lot of circular ponds in that city_') and in our case it also means confusing mapper feedback

kocio-pl:
Wording aside for a moment - my point is that (early) fountain and (late) single tree rendering are similar and while I understand you think this rule shows a problem with a fountain, I still don't know what do you think about the single tree - is it also wrong? Or is it OK - so they are different somehow, but in what exact way?

I'd like to compare these objects to understand this rule.

imagico:
Regarding your examples - i don't get the grass example - grass is rendered without a pattern so there are no geometries that could be misread.  Also we have no abstract symbols with a color close to that of grass so there is nothing that can be mistaken for grass either.

For the tree example i don't think there is anything rendered based on mapped geometries that trees could be mistaken for.  Tree rows maybe but that would not lead to the geometry to be mistaken as actual data.  But to use that example - if we would for example decide to render trees with `diameter_crown=*` with a shape drawn according to this value we should clearly distinguish that rendering from a generic tree with an abstract rendering size to avoid the drawn abstract tree diameter to be mistaken as actual data.

math1985:
I agree with @imagico, the fountain outline could indeed be mistaken for a water area of that size, while a tree outline wouldn't be easily be mistaken for something else, like a tree of that exact size.

Now I think of it, I think the current rendering if tree_line is problematic: it gives the impression that each of the green dots in the rendering stands for a single tree.

imagico:
> while a tree outline wouldn't be easily be mistaken for something else, like a tree of that exact size.

Note the rule as currently phrased does not require to avoid abstract shapes to be mistaken for real world geometries, it just suggests to avoid them being mistaken as actual data or mapped geometries.  If a certain type of feature has a certain size typically it is not necessarily undesirable to render it in exactly this size as long as there is no risk of confusion with something rendered in a size as quantitatively mapped.

Typically symbols we use avoid being interpreted as real world geometries of course by either being of a very non-natural size or a different perspective (usually side view rather than top view).

kocio-pl:
> I agree with @imagico, the fountain outline could indeed be mistaken for a water area of that size, while a tree outline wouldn't be easily be mistaken for something else, like a tree of that exact size.

Sorry for digging still the same hole, but that answers only half of my question: now I know you think they are different. But _what exactly_ makes them different? We're not talking about the colors and: 
1. They have exactly the same geometry.
2. Tree canopy have different sizes, but this it's the same with basin sizes. 

So why you think one can be mistaken and the other can't?

imagico:
The fountain background circle can easily be mistaken as a mapped water area.  What type of mapped feature do you think the tree rendering can be mistaken for?

kocio-pl:
Water is a typical part of the fountain, canopy is typical part of the tree. Of course there are exceptions (just a nozzle without a basin, trunk of a dead tree) and temporary changes (in winter you may not have the water in the basin and no leafs on the tree) and you can be mistaken if you expect always to see water or leaves here.

imagico:
If you want to understand what the difference is you should try to answer my question.

kocio-pl:
Sure - any green-related natural feature.

dieterdreist:
sent from a phone

> On 10 Dec 2016, at 22:28, Christoph Hormann <notifications@github.com> wrote:
> 
> If you want to understand what the difference is you should try to answer my question.


it would not be completely unexpected from a map in scale 1:1000 to show the actual size of the tree canopy. On the other hand, if you had any doubt, you could clearly see while zooming that all those identically sized trees are indeed just a symbol.

imagico:
I am not sure if this is a language problem or a perception problem but i am stuck here.  Maybe someone else can try this.  Note at this point this is not about if this rule is desirable to have or not but about everybody understanding what it means.  

The guideline we are talking about is:

> Design of symbols, area and line patterns should be in a way that avoids their geometries to be mistaken as actual data.

kocio-pl:
If I'm to say if it's desirable, I need to know what is this rule about... And if I don't understand it, someone not that involved in maps may also don't get it, so better if it's made more clear and simple.

You have proposed this change:

> Design of symbols, area and line patterns should be in a way that avoids their geometries to be mistaken as **mapped geometries**.

I would make it simpler at least:

> **Geometry** of symbols, area and line patterns should not be mistaken as mapped geometries.

imagico:
> Geometry of symbols, area and line patterns should not be mistaken as mapped geometries.

That's an advise to the map user, not the designer.

Note the risk of mistaking a symbol shape for map data is not only contingent to the shape but also depends on other design aspect like coloring.

kocio-pl:
OK, so what about:
> **Design** of symbols, area and line patterns should not **make users believe they are depicting** mapped geometries.

kocio-pl:
Related to https://github.com/gravitystorm/openstreetmap-carto/issues/2270.

CARTOGRAPHY.md
```diff
@@ -37,3 +37,51 @@ account.
...
+
+### Colors
+
+* Differentiating features through different colors only makes sense if this difference can actually be read by the map users.
+
+* For area colors we prefer light and low saturation tones.  Lines, individual symbols and labels also can be stronger in color.
+
+* Stronger colors indicate fairly distinct and meaningful features as opposed to more general and less distinct features which use weaker colors.
+
+* Small areas work well with somewhat stronger colors while large areas tend to call for weaker tones.
```
kocio-pl: After some [testing with midzoom](https://github.com/gravitystorm/openstreetmap-carto/issues/2194) and my thinking about parks, I guess it's not as simple. I would rather remove this rule.

imagico: This is a fairly widely accepted principle but it is usually understood to be not the only guiding principle - see for example the point immediately before in the text.

If you want to remove this point i would like to see cases where you think violating it is desirable even though there are no other principles that suggest this.

Please keep in mind we do not color areas individually based on their size so this guideline refers to the typical size of areas of a certain type of course.

kocio-pl: What is important for me is if it works for this style. If we know the typical size (which is not always true) this is not always "smaller is more important". On the other hand I would also not argue that "bigger is more important", that's why I think it's hard to say about a rule here.

1. When working with midzoom I see that the area size doesn't matter for colors - it's rather areas vs lines and points (and you have proposed such rule in this PR, which I agree with).

2. When working with pitches I see that they are typically a part of the school area or sport centres. Both of these are now rendered as pale yellow and dark pitches/tracks were hiding the fact that they are just a part of the feature. We loose the feeling of having the structure, which I consider to be a bad thing.

3. The same with park. I see it like no matter if we have tree area or the grass area or whatever else, we're still basically in the park, this is the important frame for all the features inside and it [looks bad](http://www.openstreetmap.org/relation/3921018) when smaller areas make it not clear.

4. The land color is light as a bigger area than most of the others, and here is where this rule works for me. But if you can find both valid examples and counterexamples, we can't really talk about the rule any more.

imagico: I cannot quite follow here since i neither understand which of the colors you mention you consider good and bad choices nor in what way you think the guideline interferes with that.

In my eyes a very good example for application of this principle in the past is the farmland color.  This was relatively strong which would be in line with the fact that agricultural land use is fairly distinct and meaningful.  But since agricultural land use covers huge areas in many countries it makes sense to give it a significantly weaker color to maintain an overall balance.

CARTOGRAPHY.md
```diff
@@ -37,3 +37,51 @@ account.
...
+
+### Zoom levels
+
+* Styling of map elements should work for the entire range of displayed latitudes and the full bandwidth of geographic situations that exists on earth.
+
+* Design differences between subsequent zoom levels should be small.  Continuous or small step changes over multiple levels is preferable in comparison to constant styling across multiple levels followed by a huge step.
+
+* Size increase (line widths, text sizes, icon sizes etc.) with increasing zoom level should not be larger than the scale ratio between the zoom levels so sizes do not increase relative to ground units at higher zoom levels.
+
+* Features can start to appear at a certain zoom level when zooming in but they should not disappear again at a later zoom level - except possibly for labels of features becoming very large.
```
kocio-pl: What do you mean by "feature"? For example airport icons and city names are hidden in higher zoom levels, so maybe we should be more clear with mentioning exceptions.

CARTOGRAPHY.md
```diff
@@ -37,3 +37,51 @@ account.
...
+
+* Starting zoom levels for showing features should be selected so that competition between elements does not result in the majority of features of a certain type not being shown if there is no meaningful measure of priority that is used for selection.
```
kocio-pl: This sentence needs rewording, because I read it few times and I'm still not sure what is the "result" we're trying to avoid.

CARTOGRAPHY.md
imagico: Feature here means OSM map feature - like in _one feature, one OSM element_.  When the way a feature is rendered leads to some design elements disappearing as you zoom in but the feature itself continuing to be represented that is fine - see my remarks to @math1985 w.r.t. nature reserves in that regard.  This still has to be considered carefully because of the irritation for the user this might cause, see also the _keep differences between zoom levels small_ concept.

CARTOGRAPHY.md
imagico: If you'd for example show all shops at z16 that would mean only a small fraction of the shops are actually shown (_the majority of features of a certain type not being shown_) because Mapnik does not render a symbol if there is already one rendered that would be overlapping.  If the rendering order or prioritization of the shops is not meaningful (_there is no meaningful measure of priority_) the results would be highly non-intuitive which is not desirable.

CARTOGRAPHY.md
kocio-pl: Thanks, that sounds reasonable to me.

CARTOGRAPHY.md
kocio-pl: Thanks, now I understand this rule and I agree with it. However the language still needs simplifying to be usable. What about something like:

> Starting zoom levels for **a given feature** should be selected **to show at least most of these elements if we can't prioritize them.**

If that's not precise enough for you, one can always add second sentence to clarify.

imagico: I find my formulation more to the point and less cryptic but i am certainly not the right person to judge this.

kocio-pl: @math1985 What's your opinion as a _senior collaborator_? :smile:

pnorman: I find the current text cryptic.

math1985:
@imagico What do we need to proceed with this?

imagico:
I am not sure about this at the moment.  My intent was to resolve #2270 by defining and documenting a set of overall principles regarding design in this style we all feel should guide us in decisions made.  I still think the draft as it is - with maybe a few smaller changes still to be made - could serve this purpose.  But it would be pointless if we do not adhere to what is written in it.  And with changes like #2515 i really do not have the impression what we have done here has much useful effect (see in particular https://github.com/gravitystorm/openstreetmap-carto/pull/2515#issuecomment-269970266).

imagico:
Ok, i looked over the previous discussion, made two changes:

* replaced _actual data_ with _mapped geometries_ in the guideline about symbol shapes.
* reworded the starting zoom level guideline.

In addition i reformatted to single spaced sentences and uniform hard line breaks.

I am open to further suggestions for changes, esp. to make things better understandable but preferably not at the expense of precision (which is why i did not go with the suggestion in https://github.com/gravitystorm/openstreetmap-carto/pull/2462#issuecomment-266245177 for example).

Since this is supposed to document the guiding principles of this style i am not going to push for merging it unless all maintainers agree with these - it would be pointless to have rules in the repository we do not agree upon.  But when contemplating these keep in mind that these principles, in particular the ones regarding use of colors, are not meant to be conflict free.  There will always be cases where different design principles suggest different things and you need to find a middle way of sorts.  You should not just look at each guideline individually and dismiss it if you can find a case where following only this principle will lead to a suboptimal choice.

kocio-pl:
At this point in my life (outside the project) I'm not able to review this PR once again, and after so much time with no action I've just lost most of details. However I agree with big part of your propositions, and only few cases raised issues for me. For me the preferred option would be to merge all the non-controversial rules now, but I like our current policy that in the end it's simply your decision what to do.

imagico:
I understand - but i stand by my choice not to merge it without consensus of all maintainers.  To me a value in having such a set of guidelines in the style would only exist if these are considered to be guiding principles by all of us.

If you have a subset of the principles i drafted or a completely different set of principles you would consider a useful set of guidelines i would be glad to see that.

kocio-pl:
All the rules which have no comment from me or I said (after some changes or explanations) something like "OK, makes sense for me" are the rules I agree with.

imagico:
The question is if you consider the set of guidelines minus the ones you don't agree with to be a useful and balanced set of design principles.

Overall i am slightly astonished that no one seems to think i have missed any important design principles - which i likely have.

By the way i also considered the suggestion from @pnorman to add references to external sources.  I found this difficult in many cases because many ideas are either fairly specific to this style or specific to digital maps (like the zoom level considerations) for which there is still relatively few good literature around.  But having a list of more in depth literature on map design in general would certainly be a good idea.

kocio-pl:
I simply don't see the vision behind these rules, so I'm not in a position to judge the balance. I also don't believe that a strict set of important rules can be found in such a project, but I also don't see it's the most important problem for this project to improve.

imagico:
I explained, above and in #2270, my motivation and also that these are not rules to follow but are meant to document the principles that guide us in work here to help us understand each other's decisions and contributors to understand and learn what guides us when making choices.  As said these are my suggestions for such guidelines but i welcome any alternative suggestions documenting what guides you and what you think should guide us, either as a modified version or a completely different proposal.

For me having such a document is not really important but having common guiding principles is quite essential, without it development seems likely to become increasingly divergent with each maintainer developing and merging changes with respect to individual goals but without a common sense of direction design wise.

kocio-pl:
I understand your point of view, but I'm not able to deal with this PR more than I already did.

imagico:
Closing this because the goal of achieving consensus among the maintainers does not appear attainable on this matter.

I would like to keep #2270 open because it is unresolved and IMO still deserves addressing.  I will also probably put this draft as it is somewhere for future reference as the guiding design principles for my work on this style.  Maybe the other maintainers could also try to put down their design paradigms so even if we cannot agree on a common set of principles we might be able to document the diversity in viewpoints somehow.  From my work on this stuff i can say that i do not only think this might help others to understand my views and decision, it also helped myself to develop a clearer understanding why i - often intuitively of course - make certain decisions and have certain opinions.

nebulon42:
I had no time yet, but I planned to go over this again. I know this was open very long already, but maybe closing would not be necessary yet?

PR #2472, opened 2016-11-27
ignore admin_level on capital=yes places

### LINKED ISSUE #2437: Remove admin_level from capital queries
The code for capital=yes and admin_level can be simplified now: The only remaining nodes with place=* and capital=yes are country capitals [(overpass)](http://overpass-turbo.eu/s/k1x). The usage of admin_level=2 for capitals is also problematic as detailed [here](https://lists.openstreetmap.org/pipermail/tagging/2016-October/030312.html).

Some background:
In October I documented the defacto usage of capital=yes only on country capitals [on the wiki](https://wiki.openstreetmap.org/wiki/Key:capital) and cleaned up capital=yes in around 40 cities plus around 80 in Russia [(edit page)](https://wiki.openstreetmap.org/wiki/Automated_edits/Jojo4u#2016-10_Edit_of_capital.3D.2A_on_Russian_county_capitals)

#2365 gave country capitals a special rendering. Because of the complexity of the needed SQL there where some regressions (#2408 #2393 #2399).

The discussion about expanding the capital query to more than city and town (reference: #2408) was done [on wiki](https://wiki.openstreetmap.org/wiki/Talk:Key:capital). Village and hamlet still are used with some good arguments while municipality ~~is used because there is no better tag~~ and the alternative place=atoll might be too obscure. The current cases are documented [here](https://wiki.openstreetmap.org/wiki/Key:capital#Capitals_which_are_not_tagged_as_city_or_town).

jojo4u:
Fixes #2437.

### DIFF
--- a/project.mml
+++ b/project.mml
@@ -1267,7 +1267,7 @@
       "srs": "+proj=merc +a=6378137 +b=6378137 +lat_ts=0.0 +lon_0=0.0 +x_0=0.0 +y_0=0.0 +k=1.0 +units=m +nadgrids=@null +wktext +no_defs +over",
       "Datasource": {
         "extent": "-20037508,-20037508,20037508,20037508",
-        "table": "(SELECT\n    way,\n    name,\n    CASE\n      WHEN (population ~ '^[0-9]{1,8}$') THEN population::INTEGER ELSE 0\n    END as population,\n    round(ascii(md5(osm_id::text)) / 55) AS dir -- base direction factor on geometry to be consistent across metatiles\n  FROM planet_osm_point\n  WHERE place IN ('city', 'town')\n    AND name IS NOT NULL\n    AND capital = 'yes'\n    AND admin_level = '2'\n  ORDER BY population DESC\n) AS capital_names",
+        "table": "(SELECT\n    way,\n    name,\n    CASE\n      WHEN (population ~ '^[0-9]{1,8}$') THEN population::INTEGER ELSE 0\n    END as population,\n    round(ascii(md5(osm_id::text)) / 55) AS dir -- base direction factor on geometry to be consistent across metatiles\n  FROM planet_osm_point\n  WHERE place IN ('city', 'town')\n    AND name IS NOT NULL\n    AND capital = 'yes'\n  ORDER BY population DESC\n) AS capital_names",
         "geometry_field": "way",
         "type": "postgis",
         "key_field": "",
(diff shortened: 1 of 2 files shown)


nebulon42:
Before #2473 you also have to update `project.mml`. See https://github.com/gravitystorm/openstreetmap-carto/blob/master/CONTRIBUTING.md#editing-layers for details.

To review this I will update my low zoom test data so this could take a bit.

jojo4u:
I added project.mml, thanky for reminder.

Question: I have a branch which builds upon changes of the simplify-capitals branch of this PR: https://github.com/jojo4u/openstreetmap-carto/tree/more-capitals

- Should I do a PR to openstreetmap-carto/master which then includes the changes here and the additional ones?
- Should I do PR to another branch?
- Should I wait?

kocio-pl:
I guess it depends on https://github.com/gravitystorm/openstreetmap-carto/pull/2473, because it removes current MML file and replaces it with YAML data.

nebulon42:
@jojo4u I would suggest to wait. I'm hoping to do the review this weekend. Of the other options also the first one would be an option.

nebulon42:
I'm still seeing some `capital=2` entries in the database (http://overpass-turbo.eu/s/ksN). At least one tagging mistake and some others legitimate. Especially interesting is Minsk (https://www.openstreetmap.org/node/26162465) which has been changed away from `capital=yes`.

Needing other maintainers opinion whether we might also want to include `capital=2` as `capital=yes` might be ambiguous before merging.

math1985:
According to the wiki, admin_level=2 refers to country borders, and therefore capital=2 refers to a country capital, just like capital=yes. I don't see a difference between the two, and would be in favour of dropping support for one of them. Hopefully that would simplify the tagging situation, and also make life for other data consumers easier.

nebulon42:
@math1985 currently we do not support both (only capital=yes), so you would say there is no need to include capital=2?

math1985:
I agree that would be more logical tagging, but switching to capital=2 would also require most data consumers to switch, so I don't think thats's ideal.

nebulon42:
Merged manually, thanks.

jojo4u:
The capital=2 was a test balloon by me for non-independent countries documented here: https://lists.openstreetmap.org/pipermail/tagging/2016-October/030343.html

For Minsk I added a changeset comment asking for explanation.

PR #2473, opened 2016-11-28
Require Mapnik 3 and CartoCSS 0.16.0

### LINKED ISSUE #2458: Drop legacy dependencies
Sorry for bringing this up again, but I think this is important for the project.

Currently we are supporting several Mapnik versions and several Carto versions. This makes the use of new features that are only available in recent versions quite difficult or impossible. This burden hinders the progress in my opinion.

I think we should only support Mapnik 3.x and Carto 0.16.x from a near point in the future on. This would require a new major version, which would be 3.x. The database reload would then be 4.x.

Style consumers would not get an unusable style, but they would be stuck with the last 2.x version until they upgrade their software.

Style contributors that use Kosmtik would be ready for the update as Kosmtik comes with Mapnik 3 and carto 0.16 by default. Style contributors that use Tilemill would run into problems, but it might be possible to upgrade node-mapnik and carto by hand to the required versions. (I should test this probably.)

_What would we gain?_

**Mapnik 3**

* improved control of shield placement (easily fixing #2378)
* SVG patterns
* ... more?

**Carto 0.16**

* support of YAML MML (no need for yaml2mml any more, solving #2447 and similar) -> TileMill?, Kosmtik ok
* support of HuSL, a perceptual colour space with no colours that cannot be represented in RGB (unlike LCh, but with some compromises), we could directly do colour math in a perceptual colour space removing the need for generate_road_colors script (for a proof of concept see https://github.com/gmgeo/osm-carto-gmgeo/blob/master/style.mss#L138)

The last two points reduce complexity and also make it easier for new contributors to understand the style and get going.

_What are the risks?_

The biggest problem is TileMill and has been so for months/years. Recently, there was some movement and I appreciate the effort. But I'm still not sure if there will be substantial progress. There have been concerns that Kosmtik is not easy enough to install and by excluding TileMill we will lose some potential contributors. This is a valid concern, but I think we should take the risk.

I'd like to know how many contributors are using TileMill. Please speak up.

pnorman:
This should wait until after v2.45.0.

Fixes #2458, #2198.

Enables #2470, #2045


pnorman:
carto requires a `.mml` suffix. I could rename `project.yaml` to `project.mml`, but I'm reluctant because that would create a conflict in git for **every** existing PR that touches the layers.

sommerluk:
I don’t understand well the mml/yaml/carto problem. Could you explain this further?

kocio-pl:
Removing `yaml2mml.py` script in this PR also resolves https://github.com/gravitystorm/openstreetmap-carto/issues/2447.

jojo4u:
> carto requires a .mml suffix. I could rename project.yaml to project.mml, but I'm reluctant because that would create a conflict in git for every existing PR that touches the layers.

Quick count: 5 of the 16 other PRs touch project.yaml: #2472 #2385 #2370 #2361 #2279. Are other affected? Changing the PRs would not be asking too much imho. Probably just a local rename of project.mml to project.yaml?

kocio-pl:
https://github.com/gravitystorm/openstreetmap-carto/pull/2423 would be probably also resolved by upgrading the minimum required version of CartoCSS with this PR.

jojo4u:
> [EDIT:] It will also invalidate #2459.

No problem with dropping the script, the PR was good for learning python some more.

pnorman:
> I don’t understand well the mml/yaml/carto problem. Could you explain this further?

`carto` requires the MML file to have the suffix `.mml`

nebulon42:
I left the extension at `.mml` to keep the name. It just does not care anymore if it is JSON or YAML. An upstream change would also be possible but I'd rather only do this if absolutely necessary.

.travis.yml
```diff
@@ -9,10 +9,8 @@ addons:
 node_js:
   - "0.10"
 install:
-  - npm install carto@0.12.1
+  - npm install carto@0.16.0
```
nebulon42: You should really use 0.16.3 as the versions before contain some regressions. For linting this does not make a difference but maybe is the wrong signal.

nebulon42: ok

kocio-pl:
> I left the extension at .mml to keep the name. It just does not care anymore if it is JSON or YAML. An upstream change would also be possible but I'd rather only do this if absolutely necessary.

Why? I think files recognized with their proper extensions would be the most natural solution (and .mml covering all as a fallback).

nebulon42:
In this case there is no proper extension. There is no convention that YAML files should be named .yml or .yaml. The current JSON is also not named .json. I don't know why Mapbox chose .mml and what it means. Maybe Map(box) markup language or something. The only change I see would be that Carto does no longer care about the extension at all.

gravitystorm:
> I don't know why Mapbox chose .mml and what it means.

Fun fact - .mml comes from Cascadenik, a predecessor to CartoCSS created outside of Mapbox. In Cascadenik [it contained XML](http://teczno.com/cascadenik/doc/).

kocio-pl:
> Quick count: 5 of the 16 other PRs touch project.yaml: #2472 #2385 #2370 #2361 #2279. Are other affected? Changing the PRs would not be asking too much imho. 

So, if MML doesn't mean anything special, let's make it The Right Way (TM). Three of these PRs are mine and I'm willing to fix them (except one which is probably going to waste anyway).

> No problem with dropping the script, the PR was good for learning python some more.

So was [my Bash code fix](https://github.com/gravitystorm/openstreetmap-carto/pull/2360), which [I've just dropped](https://github.com/gravitystorm/openstreetmap-carto/pull/2137). :smile:

pnorman:
If we decide to merge this I'll probably just take the first commit, and redo the delete and rename instead of resolving merge conflicts.

kocio-pl:
As far as I understand, we're positive about merging it. 

The only problem I see is release policy. We should make a bugfix release very soon and it would be v2.45.1 then, but if we merge also this code, this would be v3.0.0 probably, which would break the our typical release time frame. Which solution do you think would be better?

math1985:
We could release 2.45.1 and 3.0.0 at the same time, and leave it to the server admins which version to install.

pnorman:
Merging this code doesn't mean we need to immediately do a release

kocio-pl:
It also enables https://github.com/gravitystorm/openstreetmap-carto/issues/1126.

pnorman:
Consensus seems to be we're ready to merge. I'll rebase later today.

HolgerJeromin:
> It also enables #1126.

No, osm fr server do not run mapnik 3.

kocio-pl:
So the milestone label for it was misleading, I've changed it now.

nebulon42:
Then I can merge it tomorrow (CET).

math1985:
Small side-question: since I upgraded kosmtik and carto 0.16, my cache doesn't flush anymore, which is quite problematic when developing. Is this a known problem with the last version(s)? Is there a way to flush the cache manually?

kocio-pl:
You mean `kosmtik/tmp/`? Each project keeps its cache files under separate directory.

pnorman:
> Small side-question: since I upgraded kosmtik and carto 0.16, my cache doesn't flush anymore, which is quite problematic when developing

My startup line is ``rm -rf tmp/ && ~/osm/kosmtik/index.js serve --host 192.168.1.1 project.yml`` when working on vector tile stuff where I need to get rid of the cache all the time.

math1985:
Thanks. Strange though, the cache used to be invalidated every time project files changed on disk. Removing `tmp` every time I change a file is quite cumbersome.

imagico:
Having started using kosmtik not too long ago i had the impression this was normal but now that you mention it it seems indeed rather odd.

kocio-pl:
I started to clean the Kosmtik cache from some moment, so it probably worked before. We have a fresh ticket about this problem now:

https://github.com/kosmtik/kosmtik/issues/191

jojo4u:
I'm in same situation as imagico.

nebulon42:
I'm pretty sure that this is unrelated to carto 0.16.x

Ircama:
Just wondering whether there is a reason why the latest YAML format of *project.mml* has the same basename of the Tilemill JSON format.

AFAIK Tilemill for Windows does not allow to change the project filename. Even if Tilemill is no more supported now, one can still try to keep on using it and in case there is the need to rename *project.mml* to *project.yaml*, restore yaml2mml.py (or equivalent command) and generate project.mml in JSON format: this might lead to errors/confusion with the original YAML file. Wouldn’t be more appropriate naming the new *project.mml* (YAML) to *osm-carto.mml* or similar?

pnorman:
If someone is using tilemill I believe they need to be using a version from source to get an appropriate Mapnik version, which should also have a CartoCSS version which supports reading YAML. They'll have problems editing the layers through Tilemill, but given Tilemill's interface, there's problems doing **any** editing through it.

Ircama:
> If someone is using tilemill I believe they need to be using a version from source to get an appropriate Mapnik version, which should also have a CartoCSS version which supports reading YAML.

I haven't checked the [Ubuntu version](https://ircama.github.io/osm-carto-tutorials/tilemill-ubuntu/) yet. I am afraid that with the [Windows version](http://tilemill.s3.amazonaws.com/latest/TileMill-v0.10.1-Setup.exe) (which is the only native way to [run openstreetmap-carto with Windows](https://ircama.github.io/osm-carto-tutorials/tilemill-osm-carto/) and without an Ubuntu VM - just run, not edit) there is no possibility to perform a manual update (and manually updating *node_modules\carto* does not provide a working product).

pnorman:
> I haven't checked the Ubuntu version yet. I am afraid that with the Windows version (which is the only native way to run openstreetmap-carto with Windows and without an Ubuntu VM - just run, not edit) there is no possibility to perform a manual update (and manually updating node_modules\carto does not provide a working product).

My point is that you can't use the prebuilt tilemill binaries, since we require Mapnik 3, and the version of Mapnik that came with v0.10.1 is from 2012, so we don't need to worry if something else is broken on Tilemill v0.10.1

PR #2474, opened 2016-11-28
Explain that the cartography is covered by the license

pnorman:
This has always been the intention of @gravitystorm, but this avoids any doubt.

### DIFF
--- a/LICENSE.txt
+++ b/LICENSE.txt
@@ -7,7 +7,8 @@ This software and associated documentation files (the "Software") is
 released under the CC0 Public Domain Dedication, version 1.0, as
 published by Creative Commons. To the extent possible under law, the
 author(s) have dedicated all copyright and related and neighboring
-rights to the Software to the public domain worldwide. The Software is
+rights to the Software to the public domain worldwide. For avoidance
+of doubt, this includes the cartographic design. The Software is
 distributed WITHOUT ANY WARRANTY.
 
 If you did not receive a copy of the CC0 Public Domain Dedication


imagico:
I am not sure this is correct as written - the style does not generate anything without data so _the cartography generated by the software_ is not necessarily CC0 since it is derived from the data used and that data is not usually CC0.

What you probably want to say that any data or other works that are parts of this style and possibly subject to copyright or related rights are - when integrated into the rendered map - also subject to CC0.

pnorman:
> the cartography generated by the software is not necessarily CC0 since it is derived from the data used and that data is not usually CC0

The map will not be CC0, but the cartography will, e.g. the road colour schemes, etc.

imagico:
The style is a combination of code/software and non-code data/information that to some extent is subject to copyright.  Running the software does not create a derivative work of the software but some of this data is copied into the rendered map in that process.  The important thing is to make clear this is covered by CC0.

I would avoid using the term cartography here since this can mean - among other things - both the data with a certain styling applied or just the abstract idea of how to style certain elements in a map.  The former would be subject to the data license and the latter is not subject to copyright at all (you cannot copyright abstract ideas).

nebulon42:
Maybe we should drop the term "software" altogether? Yes, it can be viewed as one but I would not refer to e.g. a stylesheet of a website as software. The design part is more important than the representation. This is the same with e.g. SVGs. There is a representation, but it will generate a design or artwork and this is the important part.

So using design (and accompanying programs) instead of software might reduce confusion?

kocio-pl:
Maybe it'd be better to use the words "cartographic design", to make it clear that we're not talking about data rendering?

imagico:
If you take for example the idea to draw roads in blue color in a map.  That idea can be considered part of a _cartographic design_ but in itself it is not subject to copyright.  Now you can implement this idea in a computer program that renders all road features in its input data in blue.  That software can be subject to copyright but whatever license you apply to it does not apply to whatever the software produces when being used.  Other than that there is nothing in this example that bears any relation to copyright except the data used so the idea that some license can be applied to _cartographic design_ does not work here.

If now instead of a plain line color you have a fancy line pattern that you use for the roads this pattern might be in itself subject to copyright - no matter if that pattern comes in a separate SVG file or is  part of the program code.  That pattern as a piece of work subject to copyright is copied into the rendered map in some way, i.e. the rendering is a derivative work of the pattern.  Therefore it is important to make sure license terms for such components of the style are clear.

What the text should make clear is that the license terms (CC0) apply both to the code components of the style, which are executed but not copied when using the style, and those components that are copied into the rendered map.  Using the term _cartography_ or _cartographic design_ here blurs the issue IMO and could even be construed to mean that we claim some rights on the abstract design concepts that are implemented here.

kocio-pl:
@imagico Do you have a proposition of the wording to use here?

imagico:
You could simply write:

_This applies to both code components of the style as well as any parts of it that are copied into the rendered map when the style is used._

micahcochran:
I typically use the term "[symbology](http://wiki.gis.com/wiki/index.php/Symbology)" to refer to color choice, line type, line weight, fill type, icon choice, and so on.  I would agree that used in this context that "style" is synonymous.

kocio-pl:
@pnorman Do you plan to change the wording or leave it as it is now?

pnorman:
> @pnorman Do you plan to change the wording or leave it as it is now?

I haven't circled around back to this with the other stuff I've been working on, but I want to ask Andy too since this involves previous work and making explicit the situation since he started the project.

nebulon42:
Having it in the license has the advantage to have it where it is immediately useful. Merging.

PR #2480, opened 2016-12-01
Adding xmllint on all svg files

HolgerJeromin:
I think at least the svgs should be valid xml files

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,4 +16,4 @@ script:
   - jsonlint project.mml
   - set -o pipefail && ./node_modules/carto/bin/carto project.mml | xmllint - | wc -l
   - ./scripts/travis_check_project_files
-  - find symbols/ -name *.svg | xargs xmllint --noout
+  - find symbols/ -name '*.svg' | xargs xmllint --noout
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,3 +16,4 @@ script:
   - jsonlint project.mml
   - set -o pipefail && ./node_modules/carto/bin/carto project.mml | xmllint - | wc -l
   - ./scripts/travis_check_project_files
+  - find symbols/ -name *.svg | xargs xmllint --noout


.travis.yml
```diff
@@ -16,3 +16,4 @@ script:
   - jsonlint project.mml
   - set -o pipefail && ./node_modules/carto/bin/carto project.mml | xmllint - | wc -l
   - ./scripts/travis_check_project_files
+  - /usr/bin/xmllint --noout `find symbols/ -name \*.svg`
```
pnorman: xmllint is in the path, so no need for `/usr/bin/`. 

I'm not sure either way, but is `find | xargs xmllint` a better way to do this?

kocio-pl:
Could you make it just one commit?

HolgerJeromin:
Since a few month this can be [squashed on merge](https://github.com/blog/2141-squash-your-commits) by you. I would prefer that but i can try that if needed. Git command line is not my friend. :-)

micahcochran:
Taking this one step further would be to verify if the file passes the [SVG 1.1 schema](https://www.w3.org/TR/2002/WD-SVG11-20020108/SVG.xsd).  You'd need to add "--schema SVG.xsd" to the xmllint line (I've not tried this out, so your results may vary).  One approach would be to use wget/curl to download the schema file.

HolgerJeromin:
I had tested many things locally. SVG Schemas are not sufficient but needs xlink.xsd and xml.xsd. Fetching them seems not a [really good idea](https://www.w3.org/blog/systeam/2008/02/08/w3c_s_excessive_dtd_traffic/).
I had success with the three files from https://github.com/oreillymedia/HTMLBook/tree/master/schema/svg but copying them into this repo seems a bit ugly.

micahcochran:
> SVG Schemas are not sufficient but needs xlink.xsd and xml.xsd. Fetching them seems not a really good idea.

Good point.  [Travis can cache directories.](https://docs.travis-ci.com/user/caching/)  I've used this feature on other projects that used Travis-CI.

micahcochran:
@HolgerJeromin My idea goes a beyond yours and is more complicated.  It should be saved for a new pull request.

HolgerJeromin:
Yes,  I would merge this xml check and leave a real svg check for later.

HolgerJeromin:
Not a big deal but in the git history my participation is lost. 
Thanks anyway for merging.

PR #2484, opened 2016-12-03
Junction names for areas

### LINKED ISSUE #1253: Render junction=yes on areas

sommerluk:
Junction names for areas.

Based on #1271 from @math1985 to make this work without any need for a database reload. But also if the database reload comes one day, this code stays working.

Resolves #1253


imagico:
To consider here: https://github.com/gravitystorm/openstreetmap-carto/pull/1271#issuecomment-72473015

sommerluk:
I suppose without ST_BuildArea this PR is useless.

> To consider here: #1271 (comment)

To quote:

> I'd prefer to wait and do it properly, but without a cadence that's asking to wait forever. So I say we take the complexity, and remove the complexity as soon as we can.

I would also prefer to make it working right now. Later, we can simplify the code. Anyway I don’t think that the code is too complex.

kocio-pl:
We have even no estimation when 4.x could be ready, so I also like to fix it now as Andy said.

imagico:
One thing to keep in mind here is that introducing changes and complexity to work around temporary limitations will not make the database reload come faster.  There are certainly dozens, if not hundreds of changes of similar importance or urgency that all depend on the database reload so in the grand scheme it probably makes sense to put energy into that.

sommerluk:
Indeed PRs like this one don’t make the database reload come faster. But I think they also do not slow it down. Personally, I do not have the skills to work on the database reload, but I have been able to write this PR.

The difference here with many hundred other changes depending on the database reload is that for most other changes there is no workaround because they depend on missing data colums. But here a workaround is possible. Nevertheless, the fact that many changes depend on the reload is a valid argument to speed up the reload, but I think that’s independent of this PR.

kocio-pl:
@sommerluk I agree with you fully and I also share the Andy's point of view on this, so if you update the code, I'm going to merge it.

sommerluk:
Rebase gone wrong.

New follow up PR at #2510

PR #2487, opened 2016-12-04
Lua rail roads

### LINKED ISSUE #2235: Review assignment to roads table
We need to make sure what we're putting in the roads table is sensible for us

pnorman:
Fixes #2235 

Cuts down roads table by a decent percentage.


pnorman:
Merging since this will all have a chance for code review before being merged to master, and it's needed to avoid holding up testing.

PR #2493, opened 2016-12-06
Cleanup SVGs for places

pnorman:
There should be no visual changes here.

### DIFF
--- a/symbols/place/place-4-z7.svg
+++ b/symbols/place/place-4-z7.svg
@@ -1,30 +1,10 @@
 <?xml version="1.0" encoding="UTF-8" standalone="no"?>
 <svg
-   xmlns:dc="http://purl.org/dc/elements/1.1/"
-   xmlns:cc="http://creativecommons.org/ns#"
-   xmlns:rdf="http://www.w3.org/1999/02/22-rdf-syntax-ns#"
-   xmlns:svg="http://www.w3.org/2000/svg"
    xmlns="http://www.w3.org/2000/svg"
    version="1.1"
    width="4"
-   height="4"
-   id="svg2">
-  <defs
-     id="defs4" />
-  <metadata
-     id="metadata7">
-    <rdf:RDF>
-      <cc:Work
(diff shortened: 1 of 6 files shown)


nebulon42:
Unnecessary in my opinion as it makes SVGs harder to edit (by hand). Any export with Inkscape will re-introduce the clutter again.

pnorman:
> Unnecessary in my opinion as it makes SVGs harder to edit (by hand)

How?

nebulon42:
Sorry, that was unclear. I meant that it makes them harder to edit because you have to do it by hand as exporting with Inkscape will re-add the clutter. Correct me if I'm wrong and the additional clutter can be avoided with Inkscape.

pnorman:
> Sorry, that was unclear. I meant that it makes them harder to edit because you have to do it by hand as exporting with Inkscape will re-add the clutter. Correct me if I'm wrong and the additional clutter can be avoided with Inkscape.

It might add some of the XML namespace tags and perhaps RDF meta-data, but the coordinate precision stuff should stay. I don't know for sure, since I don't use Inkscape.

But based on the feedback, I think I'll change them to a circle in a circle.

dieterdreist:
sent from a phone

> On 6 Dec 2016, at 23:01, Paul Norman <notifications@github.com> wrote:
> 
> It might add some of the XML namespace tags and perhaps RDF meta-data, but the coordinate precision stuff should stay. I don't know for sure, since I don't use Inkscape.


inkscape can save as "plain svg" and there's also an xml editor for fine tuning integrated.

pnorman:
> It might add some of the XML namespace tags and perhaps RDF meta-data, but the coordinate precision stuff should stay.

As well as the replacing of the group + transform with just a plain path in one of the SVGs. I'm slightly surprised the former worked, given Mapnik's limitations.

kocio-pl:
I see nothing wrong with this code cleaning - the worst thing that can happen is that the clutter will be back, which doesn't scare me too much.

Issue #2488, opened 2016-12-04
Render addr:unit on building entrances


tpikonen:
The recommended way of tagging staircases in apartment buildings is to use addr:unit (see: http://wiki.openstreetmap.org/wiki/Key:addr#Detailed_subkeys). This tag can be conveniently added to the entrance=staircase nodes on the building perimeter. Currently addr:unit is not rendered in the default OSM layer.

The Finnish OSM community currently has some momentum on improving addresses, but since addr:unit is not rendered, people have resorted to the incorrect method of using addr:housenumber on entrances. 

The screenshot below, taken at http://www.openstreetmap.org/#map=18/60.15612/24.74071 shows two buildings of the same apartment complex, both with addr:housenumber=5. The building above is tagged correctly with addr:unit=C and D on main staircase entrances. The building below has entrances incorrectly tagged with addr:housenumber=5A (and 5B and 5C). By panning around in this area, one can see many examples of this incorrect tagging, made probably because it's the only way to make entrance addresses visible on the default map. 

It would be really nice to have addr:unit tags on entrances rendered on higher zoom levels, perhaps one zoom level higher than when addr:housenumber becomes visible and with a slightly smaller font than addr:housenumber. 

![entranceunit](https://cloud.githubusercontent.com/assets/3215/20867067/a59ac694-ba44-11e6-92b7-62368535dc0e.png)

nebulon42:
I have not found `addr:unit` tags on the mentioned building. But anyway, yes I agree that it would be worthwhile to render `addr:unit`. I personally would require also the presence of `addr:housenumber` in addition to `addr:unit`. Else how do you know to which address the unit does belong to? That said I would extend the current rendering of `addr:housenumber` and `addr:housename` to also include `addr:unit`when present.

cc @sommerluk

tpikonen:
The addr:unit tags in the top building are in the entrance nodes.

Rendering also addr:housenumber in the case of entrance=staircase nodes would lead to unnecessary clutter IMO, like in the lower building in the image, where the unit letter is added to addr:housenumber tag in the entrance to make it visible. If a building has an addr:unit tag, it would make sense to render it. Taginfo says that addr:unit is used on 180k nodes and only 20k ways, so there are not that many buildings with addr:unit tags. Not surprising, since units are usually staircases and other building-internal features.

The top building in the image below shows how to render addr:unit in entrances with (IMO) least clutter and most useful information.
[...]

Issue #1286, opened 2015-02-02
Establish cadence for database reloads


gravitystorm:
There's a large number of different organisations that use this style on their own systems, and would prefer not to have to reload their databases too often when we want to make changes to the schema. We've had a moratorium on changing the .style file for over 2 years now but time is coming for this to change.

While it's tempting to think that if we change it once (i.e. add hstore) we'll never have to do so again, it's unlikely. We should therefore think of it as something we do occasionally, but weigh up the benefits to cartography versus the headaches for downstream users.

It would be useful to find out how often admins reload their databases anyway, for non-openstreetmap-carto reasons. It will also be good to have suggestions from tileserver admins as to how frequent would become inconvenient.

pnorman:
> It would be useful to find out how often admins reload their databases anyway, for non-openstreetmap-carto reasons. It will also be good to have suggestions from tileserver admins as to how frequent would become inconvenient

My recommendations have been to reload every 6 months or annually for performance reasons when running a high-demand worldwide osm2pgsql database server, and to re-index more frequently than that. I can't recall the exact figures for performance degradation at the 6 and 12 month marks, but I think it was over 25%.

I know of *no one* who actually does this - of everyone whom I've talked to, reloads are only when commissioning a new server.

SomeoneElseOSM:
I'm guessing here, but I suspect that in terms of total numbers the largest number of instances of running stylesheets based on openstreetmap-carto or a derivative are actually smaller extracts rather than full planets.  If these people are applying diffs then they'll probably be reloading every month or so because applying worldwide diffs (if they don't trim them first) to a regional extract is obviously going to make it significantly bigger.

Perhaps a straw poll in #osm on IRC or similar might turn up some more "reasons for update" and reload frequency?  Another interesting question would be "how often do you merge style changes made to openstreetmap-carto to your map style"?

In my case (tiles created for personal use only on a small local server), I geographically trim diffs before applying and do use lua tag transformations, so it's about every 3 months or sooner if I've made a significant lua script change.

pnorman:
> If these people are applying diffs then they'll probably be reloading every month or so because applying worldwide diffs (if they don't trim them first) to a regional extract is obviously going to make it significantly bigger.

Geofabrik supplies daily diffs for their extracts, but I'm doubtful to if they're used as much as they should be.

SomeoneElseOSM:
[...]

PR #2494, opened 2016-12-06
Render addr:unit

tpikonen:
This branch implements rendering of addr:unit tags on ways (buildings) and nodes (entrances etc.), as discussed in issue #2488 

The image below, rendered from http://www.openstreetmap.org/#map=18/60.16315/24.87713 shows rendering of addr:unit with addr:housenumber on buildings and without addr:housenumber on entrances.

addrunit

### DIFF
--- a/addressing.mss
+++ b/addressing.mss
@@ -18,6 +18,12 @@
       ["addr_housename" != null] {
         text-name: [addr_housenumber] + "\n" + [addr_housename];
       }
+      ["addr_unit" != null] {
+        text-name: [addr_housenumber] + " " + [addr_unit];
+        ["addr_housename" != null] {
+          text-name: [addr_housenumber] + " " + [addr_unit] + "\n" + [addr_housename];
+        }
+      }
     }
     text-placement: interior;
     text-min-distance: 1;
@@ -26,8 +32,11 @@
     text-halo-radius: @standard-halo-radius;
     text-halo-fill: @standard-halo-fill;
     text-size: 10;
-    text-wrap-width: 20; // 2.0 em
+    text-wrap-width: 30; // 3.0 em
     text-line-spacing: -1.5; // -0.15 em
+    [zoom >= 18]["addr_unit" != null]["addr_housenumber" = null] {
+      text-name: [addr_unit];
+    }
     [zoom >= 20] {
         text-size: 11;
         text-wrap-width: 22; // 2.0 em
--- a/project.mml
+++ b/project.mml
@@ -2138,18 +2138,20 @@ Layer:
             way,
             "addr:housenumber" AS addr_housenumber,
             "addr:housename" AS addr_housename,
+            tags->'addr:unit' AS addr_unit,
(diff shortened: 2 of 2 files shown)


pnorman:
Thanks for the contribution, but this will have to wait until after the database reload and the lua branch gets merged, and we aren't doing PRs for new features against the lua branch yet, so I'm going to close the PR.

manfredbrandl:
Is it now time to reopen this PR?

kocio-pl:
Yes, but it looks like it will have to wait few more moths to update database (see #1286).

kocio-pl:
@tpikonen Do you plan to update this code before nearest database reload window opens?

tpikonen:
I rebased the code against the current master. I do not have a local renderer installed at the moment, so I can't test this now. It's should be ok though :)

kocio-pl:
Thanks! Could you also squeeze these 3 commits into one?

tpikonen:
Ok, commits are now squeezed.

kocio-pl:
The nearest major PostgreSQL version (10) is expected on 9.11.2017, so this is probably when we should try to release osm-carto with this PR (once it's tested and accepted) to synchronize database reload with schema change:

https://www.postgresql.org/developer/roadmap/

pnorman:
This code needs updating to work against the current database schema, or schema changes should be discussed independently of this particular change.

tpikonen:
@pnorman could you elaborate on the "needs updating to work against the current database schema"? I suppose you mean the changes made to the project.mml in this PR? I'm afraid I know next to nothing about the DB in OSM, so some help would be needed.

pnorman:
> @pnorman could you elaborate on the "needs updating to work against the current database schema"? I suppose you mean the changes made to the project.mml in this PR? I'm afraid I know next to nothing about the DB in OSM, so some help would be needed.

This PR should contain no changes to openstreetmap-carto.style and instead pull the addr:unit tag from the tags column.

kocio-pl:
I have general question then - when should we update database schema? Do you think we should add the code for hstore and once in a while (when we decide to get some elements into the schema) convert the code to the canonical form - or you mean something different?

pnorman:
> I have general question then - when should we update database schema? Do you think we should add the code for hstore and once in a while (when we decide to get some elements into the schema) convert the code to the canonical form - or you mean something different?

I'm not sure what you mean. We'd want some reasonable benefit to any change, but this is best discussed in its own issue.

kocio-pl:
@tpikonen With this change the code works with current database schema: https://github.com/kocio-pl/openstreetmap-carto/commit/27f06182eb7ccae03220d0a89b691462f765dba4. 

I'm just not sure if this should be this way (using `tags->'addr:unit'`) or maybe `tags @> 'addr:unit'`, because I don't know SQL and have no rule of thumb to foilow.

tpikonen:
@kocio-pl Thanks for fixing the DB variables, I added the changes to the PR. I'm no SQL expert either, but from my reading of the PGSQL hstore docs this should now work without changes to the database schema.

PR #2496, opened 2016-12-06
switch to Noto Naskh

### LINKED ISSUE #2458: Drop legacy dependencies

### LINKED ISSUE #2485: Arabic line height excessive
Noto Kufi Arabic is really tall: It has a line height of 1.9x em. This is compared to 1.36 em of Noto Sans UI and Noto Naskh Arabic UI - sadly there is no UI variant of Kufi. Maybe reconsider decision to use Kufi instead of Naskh (style.mss:  _Noto Kufi Arabic is the one that is most similiar to a Sans style_) 

[Casablanca](http://www.openstreetmap.org/relation/392915#map=8/33.516/-6.721) - first Kufi then Naskh (the middle line is the local Shilha)
kufi
naskh

Also a road shield from #2478 in Tunisia ([link](https://www.openstreetmap.org/#map=16/36.7281/9.1974))

Line heights here: https://gist.github.com/jojo4u/0b9284abc03278559a5ccd8dc49d4c0e (divide text height by EM size).

sommerluk:
switch to Noto Naskh

Resolves #2485

### DIFF
--- a/style.mss
+++ b/style.mss
@@ -28,11 +28,12 @@ but there are some exceptions
     See also https://github.com/gravitystorm/openstreetmap-carto/issues/2208)
 
 2. Noto provides three variants of Arabic: Noto Kufi Arabic, Noto Naskh Arabic
-and Noto Nastaliq Urdu. Noto Kufi Arabic is the one that is most similiar to a
-Sans style, so this one is used while both other variants are ignored. Arabic
-fonts are placed behind Sans fonts because they might re-define some commonly
-used signs like parenthesis or quotation marks, and the arabic design should
-not overwrite the standard design.
+and Noto Nastaliq Urdu. Naskh is the most commonly used style of Arabic.
+Furthermore, Noto Naskh is the Arabic font of the Noto family with the greatest
+coverage and the only one that has an UI variant. Therefor this style uses
+Noto Naskh Arabic UI. The Arabic fonts are placed behind Sans fonts because
+they might re-define some commonly used signs like parenthesis or quotation
+marks, and the arabic design should not overwrite the standard design.
 
 3. Noto provides two variants of Emoji: Noto Color Emoji and Noto Emoji. The
 colour variant is a SVG flavoured OpenType font that contains coloured emojis.
@@ -111,7 +112,7 @@ A regular style.
                 "Noto Sans Vai Regular",
                 "Noto Sans Yi Regular",
 
-                "Noto Kufi Arabic Regular",
+                "Noto Naskh Arabic UI Regular",
 
                 "Noto Emoji Regular",
 
@@ -188,9 +189,9 @@ regular text and can be used for emphasis. Fallback is a regular style.
                 "Noto Sans Vai Regular",
                 "Noto Sans Yi Regular",
 
-                "Noto Kufi Arabic Bold",
+                "Noto Naskh Arabic UI Bold",
 
-                "Noto Kufi Arabic Regular",
+                "Noto Naskh Arabic UI Regular",
 
                 "Noto Emoji Regular",
 
@@ -263,7 +264,7 @@ For a considerable number of labels this style will make no difference to the re
                 "Noto Sans Vai Regular",
                 "Noto Sans Yi Regular",
 
-                "Noto Kufi Arabic Regular",
+                "Noto Naskh Arabic UI Regular",
 
                 "Noto Emoji Regular",
 


sommerluk:
grafik

kocio-pl:
Does anybody else except you and @jojo4u think it's a good idea in general? This rendering is of course better, but I mean wider pros and cons.

sommerluk:
Pro Noto Naskh UI:

- Greater/more complete glyph coverage than Kufi
- Has a linespacing that fits better cathografic needs
- Has an UI version
- Naskh is the script style most commonly used for arabic, much more than Kufi (according to Wikipedia – I don’t speak the language)

Pro Kufi:
- Its look is closer to a Sans style (the style used for the rest of the labels) than Naskh
- It might be slightly more legible (but that’s difficult for me to evaluate as I don’t speak the language)

Originally, I had made the choise of Kufi exclusivly because of its Sans-similar look. That means it was exclusivly a style decision. But the arguments of @jojo4u have convinced me, and the benefit of switching to Naskh will be greater than the harm it does.

math1985:
@aso-mehmudi, you commented on Arabic fonts before, do you have an opinion on this?

pnorman:
Yes, it'd be nice to get comments from people who read the language, otherwise we're stuck making the decision based on how it fits in (character form and line spacing) and our impression of legibility at small sizes of a language we don't read.

kocio-pl:
Good point - I've made an [announcement](https://forum.openstreetmap.org/viewtopic.php?id=56642) at Algerian forum.

pnorman:
Some side by side comparisons might help. I don't have an appropriate area in-DB anymore and the only planet I have loaded is for lua, so it'd be awhile before I can generate some.

jojo4u:
Naskh is quite small :/ First Kufi then Naskh. This is on branch tunetypo01.
kufi z5
naskh z5
kufi z9
naskh z9
kufi z12
naskh z12
kufi z16
naskh z16

aso-mehmudi:
- I Think, Kufi is too big and Naskh is too small.
- As you may know, several languages are using Arabic script, including Arabic, Persian, Kurdish, Urdu, Pashtu. I think Kufi has a classical Arabic spirit that feels odd for languages like Persian or Kurdish. Naskh style is more common for that languages.

@jojo4u can you please put some screenshots of countries like Iran and Pakistan in both fonts?

jojo4u:
Again with branch typetypo01. Order: Kufi, Naskh, Sans Arabic (from Noto Github folder unhinted/variable)

Naskh is still really small, Sans Arabic looks bold. 

iran kufi 05
iran naskh 05
iran sans 05
iran kufi 09
iran naskh 12
iran sans 09
iran kufi 12
iran naskh 12
iran sans 12
iran kufi 16
iran naskh 16
iran sans 16

kocio-pl:
What are your opinion? I like to have small linespacing, consistent with the rest, but it's more important for me to be acceptable by the end users (=people who use Arabic script).

math1985:
Could we post this to a/the Arabic mailing list or forum?

kocio-pl:
I've already post the message on Algerian forum, because that one looks like being used. I don't know however which list (or other places - Twitter maybe?) should we write to:

https://lists.openstreetmap.org/listinfo

Maybe this wiki template could be of some help:

https://wiki.openstreetmap.org/wiki/Template:Arab_countries

jojo4u:
I'd probably just post to OSM-Talk. There seem to be few active arabic communities:  Algeria OSM forum (already posted), [Morocoo mailing list](https://lists.openstreetmap.org/pipermail/talk-ma/) and [Tunisia Facebook](https://www.facebook.com/groups/47277958581)

saeedghazijoolaee:
Hi,
In my opinion, Noto Naskh is more readable, more beautiful, and more consistent with OSM.

Have you ever seen [maps.wikimedia.org](maps.wikimedia.org) maps? Their maps are more readable than OSM by now.

kocio-pl:
@saeedghazijoolaee Thanks for sharing your view! 

Did you however look carefully at our mockups? Part of the problem is that Naskh is smaller comparing to other scripts, so it will be smaller on the map. I see this problem also on Wikimedia maps, but they may have bigger overall font size (I haven't check it), so on osm-carto it might look smaller than there, and it could make it less readable than on Wikimedia. That's why I'm so cautious and like to know how will it look here, including the size.

imagico:
For background information: [Naskh](https://en.wikipedia.org/wiki/Naskh_(script)) is the most widespread form of writing Arabic script in most language using it.  It features proportions fairly different from Latin script making it difficult to use in combination.  Fonts are usually scaled so the overall line height is comparable to that of Latin script which makes the actual size significantly smaller.  I therefore think when using Naskh you need to increase the size for Arabic script relative to Latin for reasonable readability at the font sizes we use.  See here for a comparison:

http://mc.bbbike.org/mc/?lon=20.014944&lat=30.873226&zoom=8&num=2&mt0=geofabrik-standard&mt1=mapnik

[Kufic](https://en.wikipedia.org/wiki/Kufic) is an older form of Arabic script that is fairly rare in today's everyday writing.  The font Noto Kufi Arabic seems however only loosely based on it.  I don't know the exact origin of this type of writing, there is probably some connection to [Pseudo-Kufic](https://en.wikipedia.org/wiki/Pseudo-Kufic).  It appears to be designed to be readable at small sizes but it is likely not that easy to read if you are not used to it.  It would be important to know how how well it works for typical readers in various countries although this PR is for moving away from it so this is not necessarily an issue.

kocio-pl:
@jojo4u Thanks - I totally forgot about simple solutions... I've posted [the message to Talk](https://lists.openstreetmap.org/pipermail/talk/2016-December/077215.html) today.

pnorman:
looking at [road signs in israel](http://media.mot.gov.il/PDF/HE_TRAFFIC_PLANNING/tamrurim2010.pdf)

Given the larger sample of usage we have then wehn we made the initial call, I'm 👍 on this.

Size will be a problem, but it's **always** been a problem for arabic.

lyxys:
As a non-native speaker of Arabic, I find the Naskh font a lot more readable than the Kufi font. The smaller size is a problem, though.

pnorman:
From a non-native reader "Naskh is better but so small that it's barely readable, out of remaining 2 Sans Arabic is slightly better"

BoFFire:
Hi, I'm a member of OSM Algeria and personaly switched to Noto Naskh even on OSMAnd App which introduced Noto in the nightly releases.
Personaly I found the actual arabic font used on OpenStreetMap very thin.
That's a personal point of view and does not reflect the point of view of the community.
I'll try to talk to the members of OSMAlgeria and ping OSMMaroc on Twitter and talk about the choice of the font.
Thank you for the post in the official forum ;)
And thank you again for the great job you're doing.
Salutations amicales.

Issue #2388, opened 2016-09-30
Outline for  leisure=sports_centre and stadiums missing


HolgerJeromin:
[Sports_centre](http://www.openstreetmap.org/way/370767384#map=19/51.87857/8.48727) has no fill/stroke color, so the outline is not visible. I think this is a regression.

The [definition](https://github.com/gravitystorm/openstreetmap-carto/blob/b6d96dbccced3ec658fadbe577ecf0caea84e98a/landcover.mss#L60) and the [code](https://github.com/gravitystorm/openstreetmap-carto/blob/b6d96dbccced3ec658fadbe577ecf0caea84e98a/landcover.mss#L531-L538) says this should have a fill color. 

Edit: 
Same for leisure=stadiums
http://www.openstreetmap.org/relation/185411

imagico:
It has a fill but should probably also get an outline at the higher zooms (though not from z13 as it is currently rendered for other areas with the same color).

kocio-pl:
What if the outline would be green instead of brown (as in schools or hospitals)?

kocio-pl:
Fixed with https://github.com/gravitystorm/openstreetmap-carto/pull/2574 (I don't know why this issue was not closed automatically).

HolgerJeromin:
This issue was mentioned in a comma separated list. That's why it was not closed by github itself.

PR #2497, opened 2016-12-06
Do not render camp_site and quarry outline until z13

math1985:
This fixes the strong outlines as seen in https://cloud.githubusercontent.com/assets/5251909/20907746/5f006326-bb50-11e6-8ab6-2875a8e5ac75.png

### DIFF
--- a/landcover.mss
+++ b/landcover.mss
@@ -95,8 +95,10 @@
   [feature = 'tourism_picnic_site'] {
     [zoom >= 10] {
       polygon-fill: @campsite;
-      line-color: saturate(darken(@campsite, 60%), 30%);
-      line-width: 0.3;
+      [zoom >= 13] {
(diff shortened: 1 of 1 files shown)


imagico:
When i looked over the code i observed a number of additional inconsistencies here:

* aeroway=aerodrome is also showing an outline from z10
* quarry outline is stated as `gray` and not relative to the base color, it is also very strong so it looks like a barrier.
* in the urban context there are some other problems - like playground and swimming pool outlines starting earlier than pitch/track despite being typically smaller.

These do not necessarily have to be solved here and now but in case you want to address them...

kocio-pl:
It is not so obvious with current dark forest color, but still makes sense for me - thanks!

@imagico You should probably open a ticket(s) about it. I am interested about it in the context of https://github.com/gravitystorm/openstreetmap-carto/issues/2388#issuecomment-257848990.

Issue #2429, opened 2016-11-08
Name, rendering not on layer 16


AllroadsNL:
name=Autódromo Santiago "Yaco" Guarnieri
on a polygon, leisure=sports_centre , sport=motor

[Layer 15 it is](http://www.openstreetmap.org/#map=15/-27.3917/-58.9980)
but not on
[Layer 16(http://www.openstreetmap.org/#map=16/-27.3876/-59.0054)
then on
[Layer 17 it is back ](http://www.openstreetmap.org/#map=17/-27.38896/-59.00283)

pnorman:
z17
![image](https://cloud.githubusercontent.com/assets/1190866/20096010/cb5d1c4e-a55c-11e6-8c09-c1fdfce54a8d.png)

Looks like a collision between the icons and text happens on z16.

kocio-pl:
I would like the parking icon ("P") to have lower priority than names in general. Is it possible?

meased:
> I would like the parking icon ("P") to have lower priority than names in general.

:+1: 

Usually when I add parking lots to an area the rendering gets worse.

pnorman:
> I would like the parking icon ("P") to have lower priority than names in general. Is it possible?

Yes. My guess on how to best implement it would to be move parking to the low priority amenity points layer.

AllroadsNL:
Although both should be visible at the layers. Name above the P icon.

kocio-pl:
Unfortunately I'm not too familiar with layers. 

Parking is already included in [amenity-low-priority-poly](https://github.com/gravitystorm/openstreetmap-carto/blob/master/project.yaml#L2499), but probably this layer should be moved earlier (as it was with https://github.com/gravitystorm/openstreetmap-carto/pull/2394)?

kocio-pl:
I've made map.osm export from this area and was trying to change tags in text editor to see the difference. It looks like both oneway arrows and parkings are blocking the name from rendering on z16 - changing both tags allows it to be seen.

I was also trying to remove just the arrows from the data and play with parking layers - but even placing "amenity-points-poly" and "amenity-points" as the first layers was not working. Only removing these layers allows the name to be rendered.

Maybe someone more experienced with layers could help?

pnorman:
> Maybe someone more experienced with layers could help?

Good time to learn ;)

[...]

PR #2506, opened 2016-12-12
Cosmetic fix for uniform layers code

kocio-pl:
Small fix to make layers code more uniform, as discussed in https://github.com/gravitystorm/openstreetmap-carto/issues/2429.

### DIFF
--- a/project.mml
+++ b/project.mml
@@ -781,8 +781,8 @@ Layer:
     properties:
       minzoom: 14
     advanced: {}
-  - name: "roads-casing"
-    id: "roads-casing"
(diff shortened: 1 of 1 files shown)


kocio-pl:
Do we need names? If not, we could remove names right now and stay with just ids.

nebulon42:
If we use any selectors of the form `#layername` then all versions of carto up to now need names. This needs a longer transition period.

Ircama:
Is the `id` or the `name` tag which references an identifier of the form `#layername`?

pnorman:
I think only `name` is required.

PR #2510, opened 2016-12-13
Junction names for areas

sommerluk:
Junction names for areas

Rebased #2484

### DIFF
--- a/project.mml
+++ b/project.mml
@@ -1924,10 +1924,33 @@ Layer:
             highway,
             junction,
             ref,
-            name
+            name,
+            NULL AS way_pixels
           FROM planet_osm_point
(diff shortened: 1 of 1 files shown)


sommerluk:
Thanks!

pnorman:
Reverted in 14eedbb for 3.0.x, reverting it in master too, and reopening #1253. A few notes that might help rewriting it

- Never use UNION if UNION ALL will work
- The query should be a union of selects, not with a subselect
- When using a geometry transformation function, the general approach is

```sql
(SELECT
    way
  FROM planet_osm_point
  WHERE way && !bbox!
UNION ALL
    ST_Something(way) AS way
  FROM planet_osm_polygon
  WHERE way && !bbox!
) AS f
```

sommerluk:
Okay. Thanks. I’ll have a look at this after Christmas.

Issue #1634, opened 2015-07-05
stop considering access=public as equivalent to access=yes for parkings


matkoniecz:
Access public is a pointless duplicate of access=yes, documented only for amenity=toilet and present mainly as result of bug in editors - https://josm.openstreetmap.de/ticket/10007 and https://github.com/openstreetmap/iD/issues/2280

That combination is used less than 10k times across 1 738k amenity=parking objects - http://taginfo.osm.org/tags/amenity=parking#combinations.

aceman444:
But what is the proposal? Will you just ignore the tag? And then are the parkings rendered as accessible to public or not?

matkoniecz:
As private, like any access tags other than ```yes```.

PR #2512, opened 2016-12-15
Drop access=public support for parkings

kocio-pl:
Resolves https://github.com/gravitystorm/openstreetmap-carto/issues/1634.

Simple fix to comply with [wiki specification](https://wiki.openstreetmap.org/wiki/Tag:amenity%3Dparking#Additional_Tags) for parking access type.

### DIFF
--- a/amenity-points.mss
+++ b/amenity-points.mss
@@ -344,7 +344,7 @@
     marker-placement: interior;
     marker-clip: false;
     marker-fill: @transportation-icon;
-    [access != ''][access != 'public'][access != 'yes'] {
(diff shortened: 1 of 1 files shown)


pnorman:
Does the usage back this?

kocio-pl:
[According to TagInfo](https://taginfo.openstreetmap.org/tags/?key=amenity&value=parking#combinations), we have currently 9 932 such uses, which is a 0.45% of all parkings, so I would say - yes.

kocio-pl:
Superseded by https://github.com/gravitystorm/openstreetmap-carto/pull/2528.

Issue #2353, opened 2016-09-19
landuse=brownfield looks fairly similar to buildings


imagico:
See for example here:

![landuse=brownfield](http://www.imagico.de/files/brownfield.png)

http://overpass-turbo.eu/s/isQ

imagico:
Other lower zoom level example:

http://overpass-turbo.eu/s/isX

kocio-pl:
I don't see the problem in the given example at this zoom level. I can see it for lower zoom levels, but I think it would be hard to find another color which is more distinct, yet not as dominant as before [the change](https://github.com/gravitystorm/openstreetmap-carto/pull/2292), which was a real problem for me. Lower zoom levels don't have to show all the differences, rather general shape, and current solution is good enough for me in that regard.

pnorman:
👍 for it being an issue, but finding a suitable colour is always a pain.

imagico:
Yes, IMO #2292 did not improve things overall but this is a tough call.  Map clarity is important and in general i much prefer not rendering something to rendering it in a misleading or confusing way.

Addressing this in an elegant way will require a look at the bigger picture, looking at both construction areas and roads (and other lifecycle related features) and looking at industrial landuses in the strict sense (i.e. landuse=quarry, landuse=landfill).

kocio-pl:
I have the similar problem, which is much more visible for me (but still not important enough to open a separate ticket): landuse=commercial and landuse=retail are too similar. I can recognize them properly only if both such areas are close, otherwise I'm not sure in many cases.

imagico:
The difference is landuse=commercial/retail are very similar with many boundary cases even while buildings and landuse=brownfield are something completely different.

kocio-pl:
Related to https://github.com/gravitystorm/openstreetmap-carto/issues/1991.

Tomasz-W:
I can admit that colour of brownfield is similar to buildings, but I think it isn't too similar and I don't see a big problem here. Note that buildings have outlines and brownfields haven't. There was no ideas how to solve it for a 2 years.
@kocio-pl What about closing it as "won't fix"?

kocio-pl:
You're right probably - on the individual buildings level I can perfectly recognize them, at lower zoom levels I don't think it's important. Please reopen if further discussion is needed.

jeisenbe:
I think this issue should be re-opened. 

This problem  is especially visible at z14 where buildings have no outlines and are darker:
[...]

Issue #1064, opened 2014-10-14
Unclassified + tunnel too similar to landuse=residential


sabas:
http://www.openstreetmap.org/way/260657228#map=19/44.39735/8.94549
http://www.openstreetmap.org/way/41725178#map=19/44.39675/8.94518
![image](https://cloud.githubusercontent.com/assets/1304918/4627639/06558e40-538d-11e4-84bd-1321293415ac.png)

For a moment I thought it was missing a connection under the building.

math1985:
Issue is still open in new road style.

PR #2515, opened 2016-12-19
Unification of major buildings

kocio-pl:

I see no reason to render airport terminals different than other major buildings:
1. It should be clear what are they from the context.
2. We use strong violet for industrial features (especially power stations, IIRC).
3. I'd like to unify the general look of all the transportation "stations".

Also sharing rule that all the buildings are rendered from z13.

Warsaw, z14:
Before
ufjaaage

After
h4fkvu5y

### DIFF
--- a/buildings.mss
+++ b/buildings.mss
@@ -5,9 +5,6 @@
 @building-major-fill: darken(@building-fill, 20%);
 @building-major-line: darken(@building-major-fill, 25%);
 
-@building-aeroway-fill: #cc99ff;
-@building-aeroway-line: darken(@building-aeroway-fill,15%);
-
 
 #buildings {
   [zoom >= 13] {
@@ -24,15 +21,7 @@
 
 #buildings-major {
   [zoom >= 13] {
-    [aeroway = 'terminal'] {
-      polygon-fill: @building-aeroway-fill;
-      polygon-clip: false;
-      [zoom >= 15] {
-        line-width: .75;
-        line-clip: false;
-        line-color: @building-aeroway-line;
-      }
-    }
+    [aeroway = 'terminal'],
     [amenity = 'place_of_worship'],
     [building = 'train_station'] {
       polygon-fill: @building-major-fill;


imagico:
> I see no reason to render airport terminals different than other major buildings

That statement very much depends on what you consider a _major building_.  Without an agreement on that the discussion is somewhat moot.

Traditionally `building-major-fill` in this style is used for buildings tagged as `place_of_worship` only - the name of the color is therefore a bit misleading.  Apparently this has been extended recently to train stations - a fairly odd combination which i seem to have missed.

Use of violet for air transport features has a long history here and is fairly intuitive overall.  It could none the less be changed of course but i don't think it makes sense to change the building color in isolation without an overall concept for the colors of air transport features in general.

kocio-pl:
I see it rather as another type of transport terminal - that is my general vision of transportation rendering concept (see https://github.com/gravitystorm/openstreetmap-carto/issues/2094).

> That statement very much depends on what you consider a major building.

I'm just tuning here what we already treat as major buildings, but of course we can always change it. For me all kind of terminals are major.

> Use of violet for air transport features has a long history here and is fairly intuitive overall. 

Long history - yes, intuitive - for me: no. Looks rather like a legacy rendering of some buildings with a different color (like supermarkets), which we're almost stopped doing now.

> Traditionally building-major-fill in this style is used for buildings tagged as place_of_worship only - the name of the color is therefore a bit misleading.

What do you consider misleading? The name tells exactly what it means and it does not suggest place_of_worship in any way.

imagico:
> I'm just tuning here what we already treat as major buildings

Could you describe what you think we treat as major buildings?  I see no consistency in that outside the original meaning of _buildings that are places of worship_.

> Looks rather like a legacy rendering of some buildings with a different color (like supermarkets)

As explained use of violet for air transport related features is an overall concept that is not limited to buildings.

>>Traditionally building-major-fill in this style is used for buildings tagged as place_of_worship only - the name of the color is therefore a bit misleading.

> What do you consider misleading? The name tells exactly what it means and it does not suggest place_of_worship in any way.

Since the color was until a few weeks ago only used for `place_of_worship` and likely no one seriously suggests these are more 'major' than various other buildings the color name is misleading.  You interpreting that as the idea that all 'major' buildings should get this color essentially just confirms that.

I would be open to the idea to color all 'major' buildings in a common color but that would as said require some agreement on what major buildings are - which i currently don't see.

kocio-pl:
> Could you describe what you think we treat as major buildings?

We _treat_ all the #buildings-major as major buildings. That's what I meant - [a duck test](https://en.wikipedia.org/wiki/Duck_test), not what is our _conception_ of major buildings, which is probably what you're talking about.

> As explained use of violet for air transport related features is an overall concept that is not limited to buildings.

Violet is related to industrial (including power features) and railway features. Hardly air transport specific.

> I would be open to the idea to color all 'major' buildings in a common color but that would as said require some agreement on what major buildings are - which i currently don't see.

Great, let's talk about it. This is just my proposition to unify this class of buildings. What is yours?

imagico:
> We treat all the #buildings-major as major buildings. 

So you essentially say a major building is a major building - very helpful.

> Violet is related to industrial (including power features) and railway features. Hardly air transport specific.

Here is what i consider the current air transport color set in comparison to the industrial color:

http://davidjohnstone.net/pages/lch-lab-colour-gradient-picker#bbbbcc,e9d1ff,cc99ff,8461c4,ebdbe8

> Great, let's talk about it. This is just my proposition to unify this class of buildings. What is yours?

I currently have no opinion on what a major building is.  And as said i also currently don't have an idea what you consider a major building.

Is a town hall a major building?  A school?  A shopping mall?  A power plant?

If not, why?  If yes, why?

mboeringa:
> I see no reason to render airport terminals different than other major buildings:
...
> **I see it rather as another type of transport terminal** - that is my general vision of transportation rendering concept (see #2094).

So, as alternative to this proposition: have you thought about colouring _all transportation buildings in violet_ (terminal, train station, ferry terminal), instead of adding them to buildings-major? That would unify transport, and make it distinct of the - primarily - "place_of_worship" major buildings. No confusion that the terminal is a church or mosque...

kocio-pl:
Yes, I have - and I have chosen dark brown. Here's why:

1. Confusion with place of worship is possible, but in this case we have always icon which makes it 100% clear. Confusion with power station is also possible, although dark violet is not the same. So we have tie here.

2. Brown has a message: this is a building, like most of buildings on osm-carto (and dark means: this is more important than typically). Violet is not related to buildings. Brown wins for me here.

3. If we use dark violet for transport terminals, we still have two different colors for one class - major buildings. We could also use it for places of worship, but that takes away building connotation even more. I would like to try minor buildings as light brown, to have even better consistency, so this would be a strange exception - subclass "major transport building", probably.

But that's just me. If we choose other type of unification, I could also accept it.

What do you think?

kocio-pl:
> Here is what i consider the current air transport color set in comparison to the industrial color:

But when you talk about "air transport color set" it's important remark. I say about "buildings". Airport terminal belongs to both categories and now the question is: which one is more important and how it could help unification? Because we have to choose, and we always loose unification in the second (less important) class.

I think "air transport" argument would be more important in case of air transport map. In case of general map I consider buildings as more important category.

> I currently have no opinion on what a major building is. And as said i also currently don't have an idea what you consider a major building.

You have no idea what should it be and you don't know what it is now. So both cases have no definition. Yet you prefer current one and reject new one. If it was really about definition, we should probably just remove this category, not insist on leaving the things as they are.

But I think this is the list where we can decide as a team, because it's useful to have some basic subclass of very large buildings class.

imagico:
> I'd like to get rid of the different colour for p.o.w. buildings

This is still an option and would be fine by me.

kocio-pl:
I think there are many places in the world, where place of worship is really important object (like on topo maps), but it's also not my top priority.

Any other ideas for major buildings list?

meased:
I've never cared for purple airport buildings and think this PR is an improvement. We could go so far as to define any public building as major, this is essentially what gmaps has done.

- Schools
- Churches
- Libraries
- Police stations
- Hospitals
- Social facilities
- Shops
- Amenities

It may be easier to define why a building would *not* be major.

pnorman:
> I'd like to get rid of the different colour for p.o.w. buildings, and wish I had pushed harder to do that when revising the building colours.

To be clear, this would have been getting rid of the idea of major buildings completely, since at the time those were the only different coloured buildings in the PR.

wmyrda:
@kocio +1 for PR
@meased 
My list of major building would be very similar to yours except I would throw out shops and amenities from it completely as they are drawn many times as the area ex. school or university

kocio-pl:
This PR is just color unification of what we already have and I like it to be focused on it. I guess general discussion about major buildings needs separate ticket to decide, because it can take much more time.

math1985:
If another opinion helps: I think we should reduce the amount of used colors (and in general simplify the style where possible), and therefore would support this PR.

imagico:
@math1985 - an alternative way to reduce the number of colors has been suggested, i.e. removing the separate area color for p.o.w. buildings a.k.a. major buildings.

On a general note: we have quite a lot of colors in this style which are prone to confusion because they are similar, see for example #2353, #1064, #1781, #1991.  The color of airport buildings however is quite clearly not frequently subject to such confusion.

I would also suggest to keep in mind the first guideline currently under discussion in #2462:

> The difference in rendering between different types of features should foremost be based on their difference in meaning and purpose for the target map users

which in case of color unification means we should make sure the differences in meaning and purpose of the features which are unified (which in this PR are airport terminals and churches/mosques/temples etc.) are small compared to the differences towards other features (in this case other types of buildings) which are still being rendered in a different color.

kocio-pl:
I think 3 building color shades of brown (light for "minor", normal and strong for "major") are desirable, because it would work as a basic subtypes of very large collection of different buildings. Of course it's not a general solution for catch-all problem, but it's what we can really do.

wmyrda:
"I think 3 building color shades of brown (light for "minor", normal and strong for "major")"

I also opt for that proposal. Seems as the most appropriate move.

kocio-pl:
First tries how minor buildings may look like:

https://github.com/gravitystorm/openstreetmap-carto/issues/738#issuecomment-270013986

polarbearing:
I'd be very selective what a major building is, and certainly not include schools, universities, hospitals, libraries and other public facilities. Otherwise we get campuses full of "major" buildings.

math1985:
* I agree on getting rid of the purple. Terminals currently don't look like buildings, and can be too easily be confused with aeroways. So I will accept this PR.
* I'd welcome a further discussion on if we want to keep different building types.

kocio-pl:
Thanks! Just a quick reminder - the discussion you mentioned has its own ticket: https://github.com/gravitystorm/openstreetmap-carto/issues/2532

PR #2519, opened 2016-12-20
Add area rendering for amenity=doctors

kocio-pl:
[UPDATE: now only for amenity=doctors]

Both amenity=clinic and amenity=doctors can be tagged as an area and it would be good to show it the same as with hospitals:

1pnjiy6r

amenity=clinic area rendering is half baked, because I had not enough experience when trying to [add the icon](https://github.com/gravitystorm/openstreetmap-carto/pull/1565).

### DIFF
--- a/landcover.mss
+++ b/landcover.mss
@@ -463,6 +463,7 @@
   }
 
   [feature = 'amenity_hospital'],
+  [feature = 'amenity_doctors'],
   [feature = 'amenity_university'],
   [feature = 'amenity_college'],
   [feature = 'amenity_school'],
--- a/project.mml
+++ b/project.mml
@@ -145,7 +145,7 @@ Layer:
               way, COALESCE(name, '') AS name,
               ('aeroway_' || (CASE WHEN aeroway IN ('apron', 'aerodrome') THEN aeroway ELSE NULL END)) AS aeroway,
               ('amenity_' || (CASE WHEN amenity IN ('parking', 'bicycle_parking', 'motorcycle_parking', 'university', 'college', 'school', 
-                                                    'hospital', 'kindergarten', 'grave_yard', 'prison', 'place_of_worship', 'clinic') 
+                                                    'hospital', 'kindergarten', 'grave_yard', 'prison', 'place_of_worship', 'clinic', 'doctors') 
                                                     THEN amenity ELSE NULL END)) AS amenity,
               ('landuse_' || (CASE WHEN landuse IN ('quarry', 'vineyard', 'orchard', 'cemetery', 'residential', 'garages', 'meadow', 'grass', 
                                                     'allotments', 'forest', 'farmyard', 'farm', 'farmland', 'greenhouse_horticulture', 
@@ -168,7 +168,7 @@ Layer:
               OR leisure IS NOT NULL
               OR aeroway IN ('apron', 'aerodrome')
               OR amenity IN ('parking', 'bicycle_parking', 'motorcycle_parking', 'university', 'college', 'school', 'hospital', 'kindergarten', 
-                             'grave_yard', 'place_of_worship', 'prison', 'clinic')
+                             'grave_yard', 'place_of_worship', 'prison', 'clinic', 'doctors')
               OR military IN ('danger_area')
               OR "natural" IN ('beach', 'shoal', 'heath', 'mud', 'marsh', 'wetland', 'grassland', 'wood', 'sand', 'scree', 'shingle', 'bare_rock', 'scrub')
               OR power IN ('station', 'sub_station', 'substation', 'generator')


polarbearing:
This provides some progress with a single pending area rendering requests. 

However we are not closer to a decision how to structure the bulk of missing areas systematically in general, #1624.

As it just copies the hospital style, which is thematically close enough, so it does not hinder upcoming general decisions.

imagico:
I had a look at this and tag use at the moment indicates there are about 1400 area features tagged `amenity=doctors` and 2300 `amenity=clinic` without a building tag (which would be the only ones that would be rendered with the style suggested here).

I would be in favor of this change for `amenity=clinic` in analogy with hospitals but it does not make a lot of sense for `amenity=doctors` IMO.  Of the 1400 features quite a lot seem to actually represent buildings but without the building tag or ways congruent with mapped buildings.  The wiki indicates the tag should be applied to the _doctor's office_ which is always a building or part of it and there is no documentation how to verifiably map an outdoor part of `amenity=doctors`  While it is conceivable that a clinic or hospital has outdoor components (like for patient recreation and rehabilitation) that would make sense to be visibly indicated to belong to it the same does not really make a lot of sense for a doctor's office.

dieterdreist:
sent from a phone

> On 31 Dec 2016, at 14:54, Christoph Hormann <notifications@github.com> wrote:
> 
> The wiki indicates the tag should be applied to the doctor's office which is always a building or part of it and there is no documentation how to verifiably map an outdoor part of amenity=doctors



a doctor's office is intended as the place where a doctor operates, which will most often be a building or part of it, yes, but it doesn't necessarily require a building. This is about a function tied to a location. We shouldn't encourage or even require that the building has this tag in addition, rather it should be seen orthogonally. There will in many occasions be properties of the building that won't apply to the office and vice versa (e.g. name, start_date, architect, website, ...)

dieterdreist:
sent from a phone

> On 31 Dec 2016, at 14:54, Christoph Hormann <notifications@github.com> wrote:
> 
> While it is conceivable that a clinic or hospital has outdoor components (like for patient recreation and rehabilitation) that would make sense to be visibly indicated to belong to it the same does not really make a lot of sense for a doctor's office.


e.g. the parking, the courtyard,

kocio-pl:
> A doctor's office is intended as the place where a doctor operates, which will most often be a building or part of it, yes, but it doesn't necessarily require a building.

Not necessary, of course, but they are already rendered properly and this PR doesn't change this case.

However in Poland there are a lot of local "health centres" with pretty basic set of services - which is in fact just a lot of doctors offices gathered in one building with clear area around:

http://www.openstreetmap.org/way/327763405

I was also thinking about rural places (in Africa for example), where the area of the doctor's office may be known, but the buildings may not be there at all (just tents).

HolgerJeromin:
If someone has forgotten a building tag this PR  would show this missing tag better than the current rendering. Right now no one would think that the poi is an polygon. 
So +1 for the PR

kocio-pl:
I can split this PR, so we could merge the part that raises no protests and discuss the second part further, but of course it'd be easier if the split is just not needed.

imagico:
Yes, rendering of `amenity=clinic` areas seems to be without objections so i would merge it.  `amenity=doctors` in my eyes currently lacks a convincing concept of verifiably mapping a doctor's office as an area independent of the physical rooms it is located in.

dieterdreist:
2017-01-02 10:54 GMT+01:00 Christoph Hormann <notifications@github.com>:

>  amenity=doctors in my eyes currently lacks a convincing concept of
> verifiably mapping a doctor's office as an area independent of the physical
> rooms it is located in.
>

if you want to split hair on the meaning of the term "office", you could
see similar problems with a lot of tags, e.g. is the toilet part of the
"office"? The parking? The fence around the site? Does the term
"independent" exclude spatial relation (i.e. "at the same spot")?

imagico:
You are not going to convince me here by pointing to other tagging inconsistencies.  Neither is the argument going to fly that a doctor's office can theoretically exist without a building if you cannot point to real life occurrences of this in significant numbers.  Note i do not make such assessments as given here as a spontaneous whim - i looked at the tag definition and the mapping practice and formed an opinion based on that.  If you want to change my opinion you need to address the points i mentioned.  I also wonder if any of you has actually looked at the data on a global scale, i.e. the 1400 features i mentioned and not just a few select examples?

kocio-pl:
Just for the record:

1. For me they are not inconsistencies - I just have different interpretation (exactly as @dieterdreist wrote it), which is as good as your interpretation, because there's no hard definition. Maybe landuse=* would be better, but that means rendering dept. trying to redefine tagging practice, which is IMO absolutely _not_ how it should be done.

2. These examples also show that there are case where it's valid and my take is it's up to the mappers to use it wisely - rules are the same for every such amenity and I see no reason we should treat this one different. Again - if you want to change them, Tagging list is probably where you should start.

But that means the split is needed anyway and I'll do it shortly.

imagico:
To be clear - i respect the different opinions stated here and am not convinced my view is necessarily any _better_ than others.

kocio-pl:
I leave this PR for amenity=doctors because the discussion is here already.

pnorman:
I can't think of anywhere locally where what's proposed makes sense, for the reasons @imagico mentions above.

math1985:
There seems to be no support for this proposal so I'm going to close it.

Issue #2505, opened 2016-12-11
Please render "amenity=arts_centre"


abdeldjalil09:
http://taginfo.openstreetmap.org/tags/amenity=arts_centre --->13 670 uses

HolgerJeromin:
Thanks for the suggestion 
It was also mentioned in the list of #108

kocio-pl:
I also think we should render it, but there's a question how should this icon look like? This is not a trivial problem.

HolgerJeromin:
I would remove the duplicate tag, because we have no other place to discuss an icon. I just wanted to link both issues

abdeldjalil09:
@kocio-pl I suggest to render like "shop=art" but brown.

kocio-pl:
I thought it would make more sense to use it with tourism=gallery.

abdeldjalil09:
Please remove "duplicate" tag, this is not duplicate of another issue.

math1985:
Fixed.

meased:
The [theatre](https://github.com/gravitystorm/openstreetmap-carto/blob/master/symbols/theatre.svg) icon also seams very fitting.

kocio-pl:
Makes sense for me, with z17+ (theatre is z16+).

kocio-pl:
[tourism=gallery has now 2 465 uses](http://taginfo.openstreetmap.org/tags/tourism=gallery), so it passes the lower limit for rendering. I like the idea to use brown palette for it and theatre icon for amenity=arts_centre, as proposed above, this would be consistent.

PR #2520, opened 2016-12-20
Adding icon for amenity=arts_centre

kocio-pl:
Resolves https://github.com/gravitystorm/openstreetmap-carto/issues/2505.

Rendering of amenity=arts_centre is the same as for the theatre, just one zoom level later (z17+):

tmsjes6z

### DIFF
--- a/amenity-points.mss
+++ b/amenity-points.mss
@@ -501,6 +501,13 @@
     marker-clip: false;
   }
 
+  [feature = 'amenity_arts_centre'][zoom >= 17] {
+    marker-file: url('symbols/shop/art.svg');
+    marker-fill: @amenity-brown;
+    marker-placement: interior;
+    marker-clip: false;
+  }
+
   [feature = 'amenity_toilets'][zoom >= 17] {
     marker-file: url('symbols/toilets.svg');
     marker-fill: @amenity-brown;
@@ -1226,7 +1233,8 @@
   [feature = 'amenity_theatre'],
   [feature = 'amenity_courthouse'],
   [feature = 'amenity_townhall'],
-  [feature = 'amenity_cinema'] {
+  [feature = 'amenity_cinema'],
+  [feature = 'amenity_arts_centre'] {
     [zoom >= 17] {
       text-name: "[name]";
       text-size: @standard-text-size;
@@ -1241,7 +1249,8 @@
       [feature = 'amenity_library'],
       [feature = 'tourism_museum'],
       [feature = 'amenity_theatre'],
-      [feature = 'amenity_cinema'] {
+      [feature = 'amenity_cinema'],
+      [feature = 'amenity_arts_centre'] {
         text-dy: 11;
       }
     }
--- a/project.mml
+++ b/project.mml
@@ -1715,7 +1715,7 @@ Layer:
                                                   'police', 'post_box', 'post_office', 'pub', 'biergarten', 'recycling', 'restaurant', 'food_court', 
                                                   'fast_food', 'telephone', 'emergency_phone', 'taxi', 'theatre', 'toilets', 'drinking_water', 
                                                   'prison', 'hunting_stand', 'nightclub', 'veterinary', 'social_facility',
-                                                  'charging_station') THEN amenity ELSE NULL END,
+                                                  'charging_station', 'arts_centre') THEN amenity ELSE NULL END,
               'shop_' || CASE WHEN shop IN ('supermarket', 'bag', 'bakery', 'beauty', 'books', 'butcher', 'clothes', 'computer', 
                                             'confectionery', 'fashion', 'convenience', 'department_store', 'doityourself', 'hardware', 'fishmonger', 'florist', 
                                             'garden_centre', 'hairdresser', 'hifi', 'ice_cream', 'car', 'car_repair', 'bicycle', 'mall', 'pet', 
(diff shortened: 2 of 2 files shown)


imagico:
I do not think this is a good idea, especially not with the same symbol as theatre.  According to [the wiki](http://wiki.openstreetmap.org/wiki/Tag:amenity%3Darts_centre) this is essentially any place that has anything to do with arts.  It does not have to be public, it does not specify if art is produced there or presented in some form.  Rendering this the same way as theatre would essentially kill the map's usefulness w.r.t. theatres.

Looking at actual use of the tag at a few places indicates practical use is often at least as vague as the definition - all kinds of community/cultural centres, multi-purpose halls, schools, workshops, art galleries, museums and similar stuff are tagged with this.  Rendering it would further encourage the use of imprecise tagging in favor of more specific tags.

kocio-pl:
We could also use brown palette icon, as discussed in the issue ticket. I think it matches tourism=gallery the best, but we could render it also for this one, because the art in itself is very wide and there's no single generic symbol for it (see [the discussion about the artwork icon](https://github.com/gravitystorm/openstreetmap-carto/pull/2236)).

As you can see, rendering has nothing to do with people tagging different objects this way. If you were right about it, people would tag arts centres as museums, schools and community centres, because these are rendered and art centres are not - not the other way around.

kocio-pl:
Icon is now changed:
1ks60qv3

abdeldjalil09:
The second one look fine for me

wmyrda:
@kocio Second icon looks fine, but why decrease visibility by moving object to lower level?

kocio-pl:
z16 is rather special, there is only subset of such amenities, which are considered more important than the rest, and I don't know if arts centre is really special.

pnorman:
> Looking at actual use of the tag at a few places indicates practical use is often at least as vague as the definition - all kinds of community/cultural centres, multi-purpose halls, schools, workshops, art galleries, museums and similar stuff are tagged with this. Rendering it would further encourage the use of imprecise tagging in contrast to more specific tags.

I don't see that a different icon changes any of this

dieterdreist:
sent from a phone

> On 29 Dec 2016, at 21:46, Paul Norman <notifications@github.com> wrote:
> 
> I don't see that a different icon changes any of this


+1, vary vague tag, was already in use in 2008 where some mappers used it for anything culture or community related, we shouldn't render it and should put a hint in the wiki that better defined tags should be preferred

kocio-pl:
I'm deeply surprized you don't know such entity as "arts centre". It's a well known concept for me, but very wide in scope, indeed. I thought it may be local (like Central Europe or Poland), but I've found it's much more popular. English Wikipedia [tries to describe it](https://en.wikipedia.org/wiki/Arts_centre) this way:

> An art centre or arts center is distinct from an art gallery or art museum. An arts centre is a functional community centre with a specific remit to encourage arts practice and to provide facilities such as theatre space, gallery space, venues for musical performance, workshop areas, educational facilities, technical equipment, etc.

As you can see, this definition also tries to enlist things and hints rather than give us clear answer what it is. This article shows later historical, cultural and even economical background in USA, GB and the rest of Europe. Then you have a list of many such places.

Making OSM Wiki definition better would be great, but I don't see it's easy, considering Wikipedia having the same problem, yet this (complex) type of objects exists in reality and is quite known. It's not that someone just invented vague tag, she rather tried to describe - better or worse - a real thing.

pnorman:
> 'm deeply surprized you don't know such entity as "arts centre".

The issue isn't how I know it, but how the tag has been applied inconsistently.

kocio-pl:
I'll try to test if it really is misused this way. I need some examples, because it's hard for me to believe theatres or schools are tagged like this when it's not even "tagging for rendering".

imagico:
Looking at overpass results for this tag will give you a pretty good idea of common tag use.  Like here:

http://overpass-turbo.eu/s/kX5

In Germany you will for example find a lot of features with this tag that have name components like _halle_, _schule_, _galerie_ or _bühne_ - features which all have their specific tags.

math1985:
Thanks, nice!

imagico:
You are aware that @pnorman and me have raised concerns that `amenity=arts_centre` is not a well defined and consistently used tag and its use should therefore not be encouraged?

math1985:
Yes, my feeling is that in this case, rendering the icon will help in making any inconsistencies visible rather than hiding them.

dieterdreist:
sent from a phone

> On 31 Dec 2016, at 15:24, Christoph Hormann <notifications@github.com> wrote:
> 
> In Germany you will for example find a lot of features with this tag that have name components like halle, schule, galerie or bühne - features which all have their specific tags.


while such names can be an indication, they do not prove that something is not well classified as arts_centre. E.g. Kunsthalle would likely indicate a museum, Maschinenhalle an industrial place (but could well be used differently now). "Sudhaus" refers to a brewery but could be used now differently. In most/many cases, former industrial (or other) buildings that have been converted to a different usage, might keep their old name as a reference to their history.

In general I agree that there's probably a mix of different stuff tagged as arts centre, especially outside of English speaking regions where the name doesn't help, given the poor definition of this tag in the wiki, and the inherent overlap with other features like theatres, event venues and museums:
"Used to mark an arts centre. A venue at which a variety of arts are performed or conducted, and may well be involved with the creation of those works, and run occasional courses."
it doesn't say what kind of arts, and creation and education are optional -> this can include a lot of places, depending on your interpretation.

mboeringa:
> while such names can be an indication, they do not prove that something is not well classified as arts_centre. E.g. Kunsthalle would likely indicate a museum, Maschinenhalle an industrial place (but could well be used differently now). "Sudhaus" refers to a brewery but could be used now differently. In most/many cases, former industrial (or other) buildings that have been converted to a different usage, might keep their old name as a reference to their history.

+1

This is really a modern trend: many old industrial buildings that faced demolition, have been re-purposed in the past ten to fifteen years as cultural venues. In most cases retaining (part) of their original name.

Also a trend is diversification: no longer "just" a gallery showing paintings on a wall in an exhibition, or a strict museum with no other activities than preservation and static presentation, but all kinds of mixed exhibition / concert / workshop activities.

This may make classifying such facilities difficult.

Anyway, as to the specific **amenity=arts_centre** tag, the biggest misusage I am currently seeing is that commercial galleries that are closer to **shop=art**, and individual artists workplaces (for which no tag seems to exist yet), are being classified as arts_centre, meaning there could potentially be hundreds in a single city.

imagico:
> Yes, my feeling is that in this case, rendering the icon will help in making any inconsistencies visible rather than hiding them.

I am not sure what this is trying to achieve.  The  tag definition is broad and vague.  Do you think because it is rendered people will realize this and change the definition of a tag already used more than 10k times?  Seems unlikely to me.  More likely scenario is that mappers will assume this is the tag to use for everything broadly within the definition (i.e. anything that has anything to do with arts) and choose it instead of a more precise tag that is not currently rendered.  Result is the data, the tag definition and the rendered maps get worse.

PR #2521, opened 2016-12-21
Check against standard CartoCSS command

pnorman:
This makes the Travis use the commands we document


kocio-pl:
Not needed after fixing documentation.

Issue #2362, opened 2016-09-21
Updating the image in README.md


kocio-pl:
I'd like to know how to generate image [we show in README.md](https://github.com/gravitystorm/openstreetmap-carto/blob/master/README.md#openstreetmap-carto)? 

I think we should add this instruction just before ["Tagging a release" section in RELEASES.md](https://github.com/gravitystorm/openstreetmap-carto/blob/master/RELEASES.md#tagging-a-release), to make sure we always have current rendering with each release.

pnorman:
:+1: to adding it to RELEASES.md, although not all releases will need a new image

nebulon42:
Always showing the current rendering may be tricky as the image has to be updated when
* any rendering of the shown features changes in the style
* the data of the shown area changes and leads to changes in rendering

This has to be verified manually. The person updating the image also needs up-to-date data in the given area.

kocio-pl:
For me RELEASES.md is a general TODO checklist - and that's good. The maintainers are supposed to make decisions in case of doubt, but they should at least know all the typical steps.

This image is a minor problem - just a general idea how does this style look like - and lack of current construction color is not important, but over time there can be more things needing update. We can just say something like: "check if rendering of the example image should be updated, and here is how to do it..." and that would be enough.

I guess @matkoniecz is the one who knows how to generate the example?

kocio-pl:
I was [wrong](https://github.com/gravitystorm/openstreetmap-carto/blob/33294204576e778b3072ee38446ea140961e1a96/preview.png) - @gravitystorm and @sommerluk did it too. Could you share how did you generate it?

sommerluk:
I had just made a screenshot at osm.org and then cut it in Gimp to the correct dimensions.

Hower, you could render it also locally:
- Download data of this place (for example with Josm)
- Render it locally and make the screenshot
- Cut it in your favorite image editor

kocio-pl:
I thought rather some command line tool to easily export the same area from Kosmtik for example. Screenshot from OSM.org is always a postcard from the past, not a sneak peek into the future. =}

gravitystorm:
I always took a screenshot too, since I've never had a Chicago extract loaded locally.

kocio-pl:
I think [having bounds defined in the Kosmtik configuration](https://github.com/kosmtik/kosmtik/issues/159#issuecomment-240774850) would make it easy to export current rendering when needed.

PR #2522, opened 2016-12-21
preview.png update instructions for a new release

kocio-pl:
Resolves https://github.com/gravitystorm/openstreetmap-carto/issues/2362.


pnorman:
We normally just use the rendering from osm.org and update as needed. I don't often have a DB with this area loaded up, so as written this would be a barrier for me doing releases.

kocio-pl:
That's the ultimate solution (current and accurate illustration of the style), so I like to have it. But since it's not crucial part of the release process, we can also hint how to do that the way you're doing it now. The most important thing here is to make this checklist more complete, not to force anything.

kocio-pl:
How about now? You can propose your wording if you have one.

math1985:
Thanks!

PR #2523, opened 2016-12-21
Proper API version for CartoCSS

kocio-pl:
Obsoletes https://github.com/gravitystorm/openstreetmap-carto/pull/2521.
Enables https://github.com/gravitystorm/openstreetmap-carto/pull/2470.

Instructions for compiling this style using CartoCSS.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -76,10 +76,12 @@ For development, a style design studio is needed.
 
 [TileMill](http://mapbox.com/tilemill) is not officially supported, but you may be able to use a recent TileMill version by copying or simlinking the project directly into your Mapbox/project directory.
 
-For deployment, `carto` and Mapnik are required.
+For deployment, CartoCSS and Mapnik are required.
 
-* [Carto](https://github.com/mapbox/carto) >= 0.16.0 (we're using YAML)
-* [Mapnik](https://github.com/mapnik/mapnik/wiki/Mapnik-Installation) 3.0
+* [CartoCSS](https://github.com/mapbox/carto) >= 0.16.0 (we're using YAML)
+* [Mapnik](https://github.com/mapnik/mapnik/wiki/Mapnik-Installation) >= 3.0
+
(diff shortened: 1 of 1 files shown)


INSTALL.md
```diff
@@ -76,10 +76,12 @@ For development, a style design studio is needed.
 
 [TileMill](http://mapbox.com/tilemill) is not officially supported, but you may be able to use a recent TileMill version by copying or simlinking the project directly into your Mapbox/project directory.
 
-For deployment, `carto` and Mapnik are required.
+For deployment, CartoCSS and Mapnik are required.
 
-* [Carto](https://github.com/mapbox/carto) >= 0.16.0 (we're using YAML)
-* [Mapnik](https://github.com/mapnik/mapnik/wiki/Mapnik-Installation) 3.0
+* [CartoCSS](https://github.com/mapbox/carto) >= 0.16.0 (we're using YAML)
+* [Mapnik](https://github.com/mapnik/mapnik/wiki/Mapnik-Installation) >= 3.0
```
pnorman: This should be kept at 3.0 because 3.1 can break backwards compatibility

INSTALL.md
```diff
@@ -76,10 +76,12 @@ For development, a style design studio is needed.
...
+
+Remember to run CartoCSS with proper API version to avoid errors (at least 3.0.0: `carto -a "3.0.0"`).
```
pnorman: Should we specify an exact version, or at least 3.0.0?

kocio-pl: I'm not sure, but this should be stable for some time (should work with future CartoCSS versions), so we can specify exactly 3.0.0.

INSTALL.md
kocio-pl: We could just mention "3.0.x" and "0.16.x", it would be pretty safe. ">=" statement is too general and pinning bugfix versions would be too strict, I guess.

PR #2526, opened 2016-12-23
Clean up quotes and empty entries in MML

pnorman:
These are left over from the original JSON to YAML conversion. Because SRS strings contain leading - and other special characters, they're left quoted.


nebulon42:
Thanks!

PR #2527, opened 2016-12-24
Rewrite relation handling code

pnorman:
This code takes a stricter view on what a multipolygon is, and adds test coverage for relation handling.

I'd like to get some more eyes on this before merging to lua and reloading my test machine over the weekend. My test machine is currently finishing off an unrelated test import for some osm2pgsql stuff and will be free then.

### DIFF
--- a/openstreetmap-carto.lua
+++ b/openstreetmap-carto.lua
@@ -342,13 +342,46 @@ function filter_tags_way (keyvalues, numberofkeys)
     return filter, keyvalues, polygon, roads(keyvalues)
 end
 
+--- Handling for relation members and multipolygon generation
+-- Multipolygons are either old-style or new-style. As defined here,
+-- a new-style MP is one with any tags other than type on the relation, and for
+-- them the tags of the ways do not matter.
+--
+-- An old-style MP is one where the only tag on the relation is
+-- type=multipolygon and the tags of all the members are either the same or
+-- empty.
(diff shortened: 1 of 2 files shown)


imagico:
If i am reading the code correctly this might lead to somewhat unexpected results with mixed tagging multipolygons, i.e. those where tags have been added to the relation in new style but not yet removed from the ways.  In this case i assume you get both the MP and the tagged members which are closed ways as polygons, usually leading to inner rings not being cut out in rendering.

There is of course no perfect way to solve this - such tagging is in most cases simply ambiguous and you could only try to detect those cases where closed way outer ring tagging and relation tagging is functionally identical and ignore the outer ring then.  But that is complicated and will often not work either.

Note osmium takes the opposite approach and widely ignores tagged outer rings of new style multipolygons - see https://github.com/osmcode/libosmium/issues/155.  I think the approach you took in principle is better, especially for rendering since it shows the faulty/ambigous tagging rather than dropping it.  But i still wanted to mention this as a possible side effect.

pnorman:
Merging, since @math1985 is away for the holidays, and I'll want to start a reload once my server becomes free.

For me the big advantages of this method are

- There are clear definitions of an old-style and a new-style MP, and no subtle ways one can become the other

- The logic can be explained simply

- The logic doesn't depend on understanding the other tags

PR #2529, opened 2016-12-27
Junction names for areas

sommerluk:
Simplified version. Only against planet_osm_line (not the polygone table).

Will not work after lua merge, so this requieres another PR against the lua branch to make sure that this feature continues working after lua merge.

UNION ALL instead of UNION

way_pixels not used (the traditional approach will only work with lua, and for the current approach I didn’t figure out how to do this).

Differently from the example code from @pnorman I didn’t figure out how to omit the last SELECT

### DIFF
--- a/project.mml
+++ b/project.mml
@@ -1821,7 +1821,17 @@ Layer:
             name
           FROM planet_osm_point
           WHERE highway = 'motorway_junction' OR highway = 'traffic_signals' OR junction = 'yes'
-        ) AS junctions
+          UNION ALL(
+            SELECT
+              ST_BuildArea(way) AS way,
+              highway,
+              junction,
+              ref,
+              name
+            FROM planet_osm_line
+            WHERE junction = 'yes'
+          )
+         ) AS junctions
     properties:
       minzoom: 11
   - id: bridge-text


kocio-pl:
I leave it to @pnorman to decide and/or merge, because I can't test related performance issues this code can result in.

pnorman:
I'd do something like this

```sql
SELECT
    *
FROM (SELECT
      way,
      highway,
      junction,
      ref,
      name
    FROM planet_osm_point
  UNION ALL
  SELECT
      ST_BuildArea(way) AS way,
      highway,
      junction,
      ref,
      name
    FROM planet_osm_line
    WHERE way && !bbox! -- Filter linestring before ST_BuildArea
  ) junction_combined
  WHERE way && !bbox! -- Filter points, and ways after ST_BuildArea
    AND (highway = 'motorway_junction' OR highway = 'traffic_signals' OR junction = 'yes')
```

Unless it's desirable to apply different tag WHERE conditions to the point and line parts of the query, in which case, get rid of the outer SELECT and have WHERE clauses on both parts of the query which reference !bbox!.

kocio-pl:
@sommerluk Do you plan to fix this code as requested any time soon?

sommerluk:
Closing for the moment. I’ll continue later …

Issue #2449, opened 2016-11-19
Rendering of amenity=clinic as area


Fringillus:
I wonder if it's suitable to maybe render amenity=clinic in the maps. Maybe in a similar way to amenity=hospital.

Example: [This](http://www.openstreetmap.org/way/446541166) a major clinic but not suitable for hospital tagging leaves an odd clear space.

polarbearing:
Could be discussed in context of other _societal amenities_ as defined in #1624, such as social_facility and community_centre.

d1g:
For A=clinic, I would prefer to display [HC 2.0](http://wiki.openstreetmap.org/wiki/Proposed_features/Healthcare_2.0) tagged objects (`health_facility:type=*`), because it is [supported](https://github.com/osmandapp/OsmAnd-resources/search?utf8=%E2%9C%93&q=health_facility%3Atype) in OsmAnd already.

Fringillus:
So this is nothing that can be implemented?

kocio-pl:
It could be, but no one wrote a code for it yet.

PR #2530, opened 2017-01-02
Add area rendering for amenity=clinic

kocio-pl:
Resolves https://github.com/gravitystorm/openstreetmap-carto/issues/2449.

Split from https://github.com/gravitystorm/openstreetmap-carto/pull/2519. 

Renders clinic area the same as for the hospital (part of the code is already here since https://github.com/gravitystorm/openstreetmap-carto/pull/1565):

ozdwh4c4

### DIFF
--- a/landcover.mss
+++ b/landcover.mss
@@ -463,6 +463,7 @@
   }
 
   [feature = 'amenity_hospital'],
+  [feature = 'amenity_clinic'],
   [feature = 'amenity_university'],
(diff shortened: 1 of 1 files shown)


imagico:
Thanks.

That means clinic was already in the landcover layer but not used?

kocio-pl:
Exactly, I was learning then and missed how should I do it.

PR #2533, opened 2017-01-03
Switch to Lua transforms and change database schema

### LINKED ISSUE #101: ele with alpine_hut when mapped as a building

pnorman:
**Not yet ready for merge**

## What this does

This PR accomplishes four main tasks

### Lua transforms

It switches to Lua transforms, which allows pre-processing of tags like layer. This allows SQL like 

```sql
```

  Instead of the longer
```sql
ORDER BY CASE WHEN layer~E'^-?\\d+$' AND length(layer)<10 THEN layer::integer ELSE 0 END, way_area DESC
```

There are also more significant changes for z_order, but that can be done post-merge

### Disjoint areas

It uses `-G`, so disjoint areas are imported as PostGIS MULTIPOLYGONs instead of multiple rows as POLYGONs. These are often found on administrative boundary relations where islands are disjoint.

### Tag columns

It uses `--hstore`, so tags that don't have a column end up in a special `tags` column. This allows for more flexability, and the number of columns has been significantly reduced

### Multipolygon handling

The multipolygon handling is in the transforms, so instead of the C transforms, we are responsible for defining the handling.

For the purposes of this PR and the transforms, a multipolygon is either new-style or old-style. A new-style multipolygon has a tag other than `type=multipolygon` on the relation, and tags on its members are ignored. An old-style multipolygon is one with only a `type=multipolygon` and all the member ways have the same tags, or a member may be untagged. For technical reasons, deleted tags (e.g. `source=*`) are ignored.

There is also implicitly the definition of an invalid multipolygon, which is a type=multipolygon relation which doesn't meet the above criteria.

Some properties of MPs as defined like this are

- Any new-style MP will never change to an old-style MP by only changing tags on its member ways, nor will it become invalid. Sudden changes to interpretation were possible with the C transform logic, where adding `building=yes` to a small object would set the building column to yes for a large object generated by a relation.

- Any old-style MP will never change to an new-style MP by only changing tags on its member ways, but may become invalid if the tags become conflicting

- type=multipolygon relations are always areas

- Processing mulitpolygons does not require any knowledge of tagging semantics like is required for area formation from polygons. For technical reasons, deleted keys are ignored, which does involve tagging semantics.

### Area handling

We can now determine if a way corresponds to an area feature or a linear feature however we want, instead of just based on tag keys. We want to avoid the interpretation of a correctly mapped feature depending on if a way is closed or not.

## Open issues

* [x] Add comments to `openstreetmap-carto.style` pointing to INSTALL.md

* [x] Add something on using hstore to the code style guide (`tags->'foo'` instead of `tags -> 'foo'`, etc) (#2552)

* [x] The interim need to backport PRs against both master and 3.x needs documenting in CONTRIBUTING.md

* [x] Define more precisely how long we will maintain 3.x and 4.x compatibility

* [x] Better document MPs somewhere

## Pre-merge changes

The changelog and README.md version information need changing

A new git branch for 3.x.x work needs creating

## Post-merge changes

Mark database layout milestone as finished

Fixes #101, #1362, #1504

## Preview

I have a demo at https://lua.osm-carto.paulnorman.ca/. It is not consuming OSM updates. Before reporting a bug please

- See if it's covered already in the discussion
- See if it's an expected change
- Check the OSM tagging so you can report what it is, if relevant
- Check if the OSM data has recently changed


CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 ## [Unreleased](https://github.com/gravitystorm/openstreetmap-carto/compare/v3.0.0...master)
 
+## [v3.0.0](https://github.com/gravitystorm/openstreetmap-carto/compare/v3.0.0...v3.0.1) - 2016-12-22
+### Changes
+- Revert of junction area changes which were preventing index usage
+
```
pnorman: Diff noise from v3.0.1 being on its own branch. Probably best solved by merging v3.0.1 into master

project.mml
```diff
@@ -1706,7 +1706,7 @@ Layer:
             way,
             COALESCE(
               'aeroway_' || CASE WHEN aeroway IN ('helipad', 'aerodrome') THEN aeroway ELSE NULL END,
-              'tourism_' || CASE WHEN tourism IN ('artwork', 'alpine_hut', 'camp_site', 'caravan_site', 'chalet', 'wilderness_hut', 'guest_house',
+              'tourism_' || CASE WHEN tourism IN ('artwork', 'alpine_hut', 'camp_site', 'caravan_site', 'chalet', 'guest_house',
```
pnorman: Looks like a bad merge removed `wilderness_hut`, or there were problems with the PR that introduced it. Also maybe a whitespace error. Repeats three more times below.

project.mml
```diff
@@ -1924,33 +1924,10 @@ Layer:
...
-              junction,
-              ref,
-              name,
-              way_area/NULLIF(!pixel_width!::real*!pixel_height!::real,0) AS way_pixels
-            FROM planet_osm_polygon
-            WHERE junction = 'yes'
-            ORDER BY way_pixels DESC NULLS LAST
-          )
-         ) AS junctions
+        ) AS junctions
```
pnorman: There's some diff oddness here since ST_BuildArea isn't present in master

scripts/lua/openstreetmap-carto.lua
```diff
@@ -0,0 +1 @@
+../../openstreetmap-carto.lua
```
pnorman: I'd like to get rid of the need for the symbolic link, but I'm not sure how to do that best within Lua.

scripts/lua/test.lua
```diff
@@ -0,0 +1,144 @@
...
+assert(equaltables(combine_member_tags({{}, {}}), {}), "test failed: no member tags, two members")
+assert(equaltables(combine_member_tags({{foo="bar"}}), {foo="bar"}), "test failed: one member, tags")
+assert(equaltables(combine_member_tags({{foo="bar"}, {}}), {foo="bar"}), "test failed: two members, tags on first")
+assert(equaltables(combine_member_tags({{}, {foo="bar"}}), {foo="bar"}), "test failed: two members, tags on second")
+assert(equaltables(combine_member_tags({{foo="bar"}, {foo="bar"}}), {foo="bar"}), "test failed: two members, tags on both")
+assert(combine_member_tags({{foo="bar"}, {baz="qax"}}) == nil, "test failed: two members, different tags")
+
+print("TESTING: filter_tags_relation_member")
+
+-- Tests filter_tags_relation_member against expected values
```
pnorman: Start with `---` for luadoc

openstreetmap-carto.style
```diff
@@ -1,154 +1,52 @@
...
 node,way   oneway       text         linear
-node,way   operator     text         linear
 node,way   place        text         polygon
-node,way   population   text         linear
 node,way   power        text         polygon
-node,way   power_source text         linear
-node,way   public_transport text     polygon
 node,way   railway      text         linear
 node,way   ref          text         linear
 node,way   religion     text         linear
```
pnorman: Potentially this should be removed and the queries adjust to get it from hstore

project.mml
```diff
@@ -267,7 +267,7 @@ Layer:
               OR "natural" IN ('water', 'glacier'))
             AND building IS NULL
             AND way_area > 0.01*!pixel_width!::real*!pixel_height!::real
-          ORDER BY z_order, way_area DESC
+          ORDER BY COALESCE(layer,0), way_area DESC
```
pnorman: This ends up being equivalent because the C transform z_order goes first by layer, and unless a water area happens to have a road tag, it's the same.

pnorman:
I have a demo at https://lua.osm-carto.paulnorman.ca, currently loaded with 161212 data. This may change if I reload, or it may stop operating if I use the server for other tasks.

The only differences from the Lua transforms should be related to invalid MPs and area definitions on stuff like highway=services. The unrelated disjoint area changes will have a bigger visual impact, particularly for countries. e.g. there isn't a United States of America label on Alaska.

pnorman:
Something else to review is MP formation where component ways are not closed.

mboeringa:
A few minor remarks or thoughts based on what you write:

> We can now determine if a way corresponds to an area feature or a linear feature however we want, instead of just based on tag keys. **We want to avoid the interpretation of a correctly mapped feature depending on if a way is closed or not.**

Although the Wiki has some guidance on how a simple closed way must be interpreted, isn't by definition the choice for line or polyon as output of any OSM closed way an interpretation? Only multipolygon relations are explicitly and unambiguously defined as areas, and supported through software, as being area/polygon types.
My point here is that what is "a correctly mapped feature" and how it should be represented in a spatial database can only be guided by subjective judgement, or sticking with the Wiki definition if available, in case of simple closed ways...
But I guess this is all close to what you intended to convey in the original quote...

> Something else to review is MP formation where component ways are not closed.

You mean multipolygons where outers or inners consist of multiple ways, and only the combination of them can be assembled to a closed form to represent a single outer or inner loop of a polygon?

Jochen Topf's "osm-testdata" repository may be of interest here:
https://github.com/osmcode/osm-testdata/

pnorman:
> But I guess this is all close to what you intended to convey in the original quote...

We don't want a mapper to be drawing something (e.g. a track) and how it gets rendered to depend on if they close the way or not.

> You mean multipolygons where outers or inners consist of multiple ways, and only the combination of them can be assembled to a closed form to represent a single outer or inner loop of a polygon?

That'd be the normal use, but in general, any MP where some of the ways are not closed.

> Jochen Topf's "osm-testdata" repository may be of interest here:
> https://github.com/osmcode/osm-testdata/

I don't know of a good harness to test that data in combination with osm2pgsql.

openstreetmap-carto.lua
```diff
@@ -0,0 +1,518 @@
...
+--
+-- - All multipolygons are polygons
+--
+-- @param keyvalues OSM tags, after processing by relation transform
+-- @param keyvaluemembers OSM tags of relation members, after processing by way transform
+-- @param roles OSM roles of relation members
+-- @param membercount number of members
+-- @return filter, cols, member_superseded, boundary, polygon, roads
+function filter_tags_relation_member (keyvalues, keyvaluemembers, roles, membercount)
+    local members_superseeded = {}
```
lonvia: Somebody was complaining about that spelling not so long ago. ;)

openstreetmap-carto.lua
```diff
@@ -0,0 +1,518 @@
...
+end
+
+--- Handling for relation members and multipolygon generation
+-- Multipolygons are either old-style or new-style. As defined here,
+-- a new-style MP is one with any tags other than type on the relation, and for
+-- them the tags of the ways do not matter.
+--
+-- An old-style MP is one where the only tag on the relation is
+-- type=multipolygon and the tags of all the members are either the same or
+-- empty.
```
lonvia: That's a very strict definition as normally old-style MPs are defined by their outer ways. It would probably be safe to exclude ways with role `inner` without violating the properties below. If it is worth it, is a completely different question.

openstreetmap-carto.lua
```diff
@@ -0,0 +1,518 @@
...
+        end
+
+        if (polygon_values[k] and polygon_values[k][v]) then
+            return 1
+        end
+    end
+    return 0
+end
+
+function is_in (needle, haystack)
```
lonvia: unused

openstreetmap-carto.lua
```diff
@@ -0,0 +1,518 @@
...
+-- @param t1 A value
+-- @param t2 A value
+-- @return true or false
+function equaltables (t1,t2)
+    for k, v in pairs(t1) do
+        if t2[k] ~= v then return false end
+    end
+    for k, v in pairs(t2) do
+        if t1[k] ~= v then return false end
+    end
```
lonvia: If you check at the beginning of the function if both tables have the same number of entries, then you can get rid of the second for loop.

openstreetmap-carto.lua
```diff
@@ -0,0 +1,518 @@
...
+    for _, tags in ipairs(member_tags) do
+        -- Check if the member has tags
+        if next(tags) ~= nil then
+            if next(combined_tags) == nil then
+                -- This is the first tagged member
+                combined_tags = tags
+            else
+                -- A different tagged member
+                if not equaltables(tags, combined_tags) then
+                    return nil
```
lonvia: Be consistent with the return value. Either start with `combined_tags = nil` or return `{}` here. Then you can get rid of one of the ifs above where this function is called.

openstreetmap-carto.lua
pnorman: This was copied before the osm2pgsql fixes ;)

openstreetmap-carto.lua
pnorman: The `#` operator is for lists, not tables, or are you thinking of `table.getn`?
The "standard" deep comparison implementation is https://github.com/stevedonovan/Penlight/blob/master/lua/pl/tablex.lua#L128-L152, but it handles a bunch of cases that don't matter here.

openstreetmap-carto.lua
```diff
@@ -0,0 +1,518 @@
...
+    for index, value in ipairs (haystack) do
+        if value == needle then
+            return true
+        end
+    end
+    return false
+end
+
+--- compare two values.
+-- if they are tables, then compare their keys and fields recursively.
```
pnorman: This only compares tables, not other values.

openstreetmap-carto.lua
pnorman: I think there's a difference between a return of {} and nil in what the function has concluded, but I see there's nothing documented. I think I have some more documentation in CT which I need to copy over, but there is a difference between an invalid MP and a valid one with no tags, but since we don't need untagged objects, the difference doesn't matter.

openstreetmap-carto.lua
```diff
@@ -0,0 +1,518 @@
...
+        members_superseeded[i] = 0
+    end
+
+    local type = keyvalues["type"]
+
+    -- Remove type key
+    keyvalues["type"] = nil
+
+    -- Boundary relations are treated as linestring
+    if type == "boundary" or (type == "multipolygon" and keyvalues["boundary"]) then
```
pnorman: This is from https://github.com/openstreetmap/osm2pgsql/blob/master/tagtransform.cpp#L207-L209, which regards `type=multipolygon boundary=*` the same as `type=boundary` for osm2pgsql geometry formation (forming linestrings from polygons), but not `type=multipolygon` and `type=boundary` without a `boundary=*` tag on either.

Does this make sense? Maybe.

scripts/lua/test.lua
```diff
@@ -0,0 +1,144 @@
...
+assert(z_order({highway="motorway", railway="rail"}) == 440 , "test failed: motorway + rail")
+
+print("TESTING: roads")
+assert(roads({}) == 0, "test failed: no tags")
+assert(roads({foo="bar"}) == 0, "test failed: other tags")
+assert(roads({highway="motorway"}) == 1, "test failed: motorway")
+assert(roads({railway="rail"}) == 1, "test failed: rail")
+assert(roads({highway="residential", railway="rail"}) == 1, "test failed: rail+residential")
+assert(roads({railway="turntable"}) == 0, "test failed: rail=turntable")
+assert(roads({railway="rail", service="spur"}) == 0, "test failed: rail SSY")
```
pnorman: add a non-SSY test here

openstreetmap-carto.lua
```diff
@@ -0,0 +1,518 @@
...
+-- @param tags OSM tags
+-- @return 1 if it belongs in the roads table, 0 otherwise
+function roads(tags)
+    for k, v in pairs(tags) do
+        if roads_info[k] and roads_info[k][v] and roads_info[k][v].roads then
+            if not (k ~= 'railway' or tags.service) then
+                return 1
+            elseif not excluded_railway_service[tags.service] then
+                return 1
+            end
```
pnorman: There's a case not being tested (rail with a non-SSY service value), but I think the logic is correct. I agree with the complicated.

openstreetmap-carto.lua
```diff
@@ -0,0 +1,518 @@
...
+        end
+    end
+    return 0
+end
+
+-- Filtering on nodes, ways, and relations
+function filter_tags_generic(tags, n)
```
pnorman: 👍

openstreetmap-carto.lua
pnorman: So it is. I remember @math1985 adding it for a reason, but maybe I removed that reason with the MP handling change. There's probably a couple places where it **should** be used.

pnorman:
> That's a very strict definition as normally old-style MPs are defined by their outer ways. It would probably be safe to exclude ways with role inner without violating the properties below. If it is worth it, is a completely different question.

I'm not sure. We ignore relation role information for geometry formation, it seems somewhat odd to do differently for tagging interpretation. I could go either way on this.

pnorman:
> We don't want a mapper to be drawing something (e.g. a track) and how it gets rendered to depend on if they close the way or not.

Thinking about this for other work, this is a particular problem for man_made=pier. As it stands, you can't tell if a pier is mapped as an area or linear feature by looking at its tags.

mboeringa:
> Thinking about this for other work, this is a particular problem for man_made=pier. As it stands, you can't tell if a pier is mapped as an area or linear feature by looking at its tags.

No, you can't, unless it has **area=yes** as additional tag. You indeed need to look at if the way is closed.

Considering most piers have a fixed shape (long protruding thing sticking out from the coast), it is almost 100% certain that any pier tagged as a closed way, must have been mapped / intended as an area feature. Looking at my renderer's result, shows this to be quite universal. I think a default of area for closed ways tagged as **man_made=pier** is reasonable.

openstreetmap-carto.lua
```diff
@@ -0,0 +1,518 @@
...
+    -- Delete tags listed in delete_tags
+    for tag, _ in pairs (tags) do
+        for _, d in ipairs(delete_tags) do
+            if tag == d then
+              tags[tag] = nil
+              break -- Skip this tag from further checks since it's deleted
+            end
+        end
+    end
```
pnorman: I'm not sure, but it's easier to understand, and by adding short-circuit logic above to exit early for untagged objects performance should be fine.

openstreetmap-carto.lua
```diff
@@ -0,0 +1,520 @@
...
+    'nycdoitt:bin',
+    -- Chicago Building Inport (US)
+    'chicago:building_id',
+    -- Louisville, Kentucky/Building Outlines Import (US)
+    'lojic:bgnum',
+    -- MassGIS (Massachusetts, US)
+    'massgis:way_id',
+
+    -- mvdgis (Montevideo, UY)
+    'mvdgis:.*',
```
pnorman: Should be with the wildcards

openstreetmap-carto.lua
pnorman: added a test in 55bc1cc, so this can now be refactored and tested.

openstreetmap-carto.lua
```diff
@@ -0,0 +1,518 @@
...
+        platform        = {z = 90,  roads = false},
+    },
+    aeroway = {
+        runway          = {z = 60,  roads = false},
+        taxiway         = {z = 50,  roads = false},
+    },
+    boundary = {
+        administrative  = {z = 0,  roads = true}
+    },
+}
```
pnorman: the way to do this is probably 

```lua
local z_data = {
  highway = {
    motorway = 380,
    -- ...
    residential = 330
  },
  railway = {
    rail = 440,
    ...
}

local roads_data = {
  highway = {
    motorway = true
    -- ...
    -- don't list residential and other non-roads stuff
  },
  railway ...
}
```

openstreetmap-carto.lua
```diff
@@ -0,0 +1,518 @@
...
+    end
+    -- By using a second loop for wildcards we avoid checking already deleted tags
+    for tag, _ in pairs (tags) do
+        for _, d in ipairs(delete_wildcards) do
+            if string.find(tag, d) then
+                tags[tag] = nil
+                break
+            end
+        end
+    end
```
pnorman: 👍

openstreetmap-carto.lua
```diff
@@ -0,0 +1,518 @@
...
+    -- Filter out objects that are filtered out by filter_tags_generic
+    local filter, keyvalues = filter_tags_generic(keyvalues, numberofkeys)
+    if filter == 1 then
+        return 1, keyvalues
+    end
+
+    -- Filter out all relations except route, multipolygon and boundary relations
+    if ((keyvalues["type"] ~= "route") and (keyvalues["type"] ~= "multipolygon") and (keyvalues["type"] ~= "boundary")) then
+        return 1, keyvalues
+    end
```
pnorman: for performance reasons? Or to allow simpler logic like

```lua
if (not right relation type) then
  return 1, kv
end
return filter_tags_generic(kv)
```

openstreetmap-carto.lua
```diff
@@ -0,0 +1,518 @@
...
+    -- Filter out objects that are filtered out by filter_tags_generic
+    filter, keyvalues = filter_tags_generic(keyvalues, numberofkeys)
+    if filter == 1 then
+        return filter, keyvalues, polygon, roads
+    end
+
+    polygon = isarea(keyvalues)
+
+    -- Add z_order column
+    keyvalues.z_order = z_order(keyvalues)
```
pnorman: This came over from CT where there's a clear difference between tags (input, which may include annoying things like addr:foo) and columns (output, where I can make sure the names never collide with Lua or postgres syntax), but I agree, in this that distinction isn't regularly made, so it should be `keyvalues['z_order']`

scripts/lua/test.lua
pnorman: done

openstreetmap-carto.lua
pnorman: fixed.

openstreetmap-carto.lua
pnorman: fixed.

pnorman:
> Better document MPs somewhere

I'm not sure where to do this.

-----

There's been concerned about the MP definitions in the PR (including by me), but no one saying they need to be changed or won't work. I'm inclined to stick with them at this point

-----

I'm updating my data. When that's done I'll send out an announcement to talk@ and dev@. The visible bugs I'm aware of are

- [ ] `wilderness_hut` oddness
- [ ] `ST_BuildArea` on junctions

The other changes are documentation or code cleanup that shouldn't change functionality. These are important changes to do, but shouldn't effect review.

openstreetmap-carto.lua
lonvia: Right, I was naively assuming that any language that offers a table datatype also offers a O(1) size() implementation. Ignore my comment then.

imagico:
For future reference - since this will likely come up sooner or later - there are a number of more or less prominently visible effects of what i pointed out in https://github.com/gravitystorm/openstreetmap-carto/pull/2527#issuecomment-269080773

like here: https://lua.osm-carto.paulnorman.ca/#11/54.1719/8.7633

I think this is desirable because it makes ambiguities explicitly visible but it is still a fairly high impact effect.  I don't think there is currently any QA tool that points out this kind of problem (i.e. closed ways with the same primary tags as an MP relation they are member of as outer ring).

imagico:
Looking somewhat further for multipolygon issues i came across this beauty:

https://lua.osm-carto.paulnorman.ca/#15/53.4485/12.3589
http://www.openstreetmap.org/relation/4250368

See also https://github.com/gravitystorm/openstreetmap-carto/pull/2533#discussion_r96312223

The idea to ignore tags on closed way inner rings when interpreting old style MPs seems sensible to me in that context but i am not completely sure if this does not cause problems in other situations (like when the assigned roles are wrong).

kiselev-dv:
Locality multipolygon (without layer=*) covers natural water:
https://lua.osm-carto.paulnorman.ca/#16/44.6720/-63.5527

EdLoach:
> We don't want a mapper to be drawing something (e.g. a track) and how it gets rendered to depend on if they close the way or not.

Your example of a track (leisure=track) is the reason I came here to comment, as it is the first difference I've [spotted](https://lua.osm-carto.paulnorman.ca/#18/51.79729/1.15919) between the current and lua renderings - at a guess I'd say the assumption of whether it is an area or not in the absence of an area tag is the opposite of before). In [this example](http://www.openstreetmap.org/way/35311074) I don't know why I didn't use pitch or skate_park, but will leave it as it is for now (the wiki also now suggests I use roller_skating rather than just skating which was probably the only documented choice when I added it).

rory:
[Two buildings here](https://lua.osm-carto.paulnorman.ca/#18/53.34291/-6.23578) are not appearing on the lua-transform version. Both buildings haven't been touched in ~2 years, both are simple closed ways. ([way 48,898,702](https://www.openstreetmap.org/way/48898702) and [way 48,898,703](https://www.openstreetmap.org/way/48898703)):

good

bad

pnorman:
> Looking somewhat further for multipolygon issues i came across this beauty:

MP with no tags, way with no tags and role=outer, way with tags and role=inner. I'd be okay calling it a data problem and ignoring it.

> The idea to ignore tags on closed way inner rings when interpreting old style MPs seems sensible to me

We don't have this information, all we have is relation tags, relation member tags, and relation member roles.

-----

> (way 48,898,702 and way 48,898,703):

That's a strange one. Probably related to https://www.openstreetmap.org/relation/3697817. That relation makes me scratch my head a bit.

imagico:
> MP with no tags, way with no tags and role=outer, way with tags and role=inner. I'd be okay calling it a data problem and ignoring it.

Ignoring it in the sense of not rendering it would be fine.

>>  The idea to ignore tags on closed way inner rings when interpreting old style MPs seems sensible to me

> We don't have this information, all we have is relation tags, relation member tags, and relation member roles.

That would mean you cannot test if a member way is closed but you could still ignore tags on member ways with role inner in general when transferring member tags to the polygon for old style MPs.

rory:
> That's a strange one. Probably related to https://www.openstreetmap.org/relation/3697817. That relation makes me scratch my head a bit.

[That relation](https://www.openstreetmap.org/relation/3697817) is clearly trying to be an "apartment block complex". I'm not sure how to tag that properly. `place=locality`, or `landuse=residential` might make sense, but seems sub-par. Regardless, to me, a human, the relation makes sense, however I will probably try to improve it.

I wonder why it's not appearing on the map, and why *some* of the buildings are appearing, and others not.

jojo4u:
> [That relation ](https://www.openstreetmap.org/relation/3697817)is clearly trying to be an "apartment block complex". I'm not sure how to tag that properly. place=locality, or landuse=residential might make sense, but seems sub-par. Regardless, to me, a human, the relation makes sense, however I will probably try to improve it.

Use place=city_block or landuse=* as polygon, not relation.

pnorman:
> Locality multipolygon (without layer=*) covers natural water:
> https://lua.osm-carto.paulnorman.ca/#16/44.6720/-63.5527

Not a bug, http://www.openstreetmap.org/relation/1247705 had building=house. 

-----

> Ignoring it in the sense of not rendering it would be fine.

Ignoring in the sense of not considering its behavior.

-----

> Both buildings haven't been touched in ~2 years, both are simple closed ways. (way 48,898,702 and way 48,898,703):

48898703 is not simple, it self intersects in the top left. 48898702 appears to be simple. It is a member of [relation 4573941](https://www.openstreetmap.org/relation/4573941), which is an old-style MP with conflicting tags on members. This should hit the "This is an invalid old-style MP with conflicting tags" branch of logic. Given the bad geometry of the other way, I suspect a geometry problem I missed, but will need to recheck in the morning.

Fwiw, the geometry being formed for 3697817 is probably close to that of the pedestrian area minus the buildings.

imagico:
> Ignoring in the sense of not considering its behavior.

That is an understandable approach but i am a bit worried that while the mapper sees something is wrong it is not really understandable for him/her what exactly the problem is and how to fix it.  What happens is that we wrongly interpret the tag on the inner ring to apply to the whole MP, i.e. we try to be smart and we fail with that.  The logic behind this is difficult to communicate to the mapper.

I think the approach _ignoring tags on members with role inner when interpreting tags for old style MPs_ might be worth considering.  Even if this also includes ways that are not closed I don't have a scenario right now where this would make things substantially worse.

daganzdaanda:
> A new-style multipolygon has a tag other than  type=multipolygon  on the relation, and tags on its members are ignored

Does this affect the rendering of the member ways? I'm thinking of something like a way with barrier=* as an outer of some kind of landuse. Or some highway might be re-used as an outer or inner way. I would like to check this with your side-by-side comparison but I don't know how to find such MPs with overpass.

scripts/lua/test.lua
```diff
@@ -0,0 +1,162 @@
...
+print("TESTING: isarea")
+assert(isarea({}) == 0, "test failed: no tags")
+assert(isarea({foo = "bar"}) == 0, "test failed: random tag")
+assert(isarea({area = "yes"}) == 1, "test failed: explicit area")
+assert(isarea({area = "no"}) == 0, "test failed: explicit not area")
+assert(isarea({area = "no", landuse = "forest"}) == 0, "test failed: explicit not area with polygon tag")
+assert(isarea({leisure = "track"}) == 0, "test failed: leisure=track")
+assert(isarea({area = "yes", leisure = "track"}) == 1, "test failed: leisure=track with area tag")
+assert(isarea({waterway = "river"}) == 0, "test failed: river")
+assert(isarea({waterway = "riverbank"}) == 1, "test failed: river")
```
pnorman: wrong test name

math1985:
@pnorman What is the status of this PR? Do you need anything, like help or reviews?

pnorman:
> @pnorman What is the status of this PR? Do you need anything, like help or reviews?

Help addressing the comments raised, particularly the MP ones.

You can either push directly to this branch, or do a PR against it.

pnorman:
I haven't had a chance to look at this lately, and it doesn't look like anyone else has. Working on a different style for a job reduces the motivation to work on osm-carto for fun.

With the [progress on old-style MP cleanup](http://area.jochentopf.com/stats/) the number of old-style MPs has dropped from 8% mid-march to 3% mid-April. With osm2pgsql [we're considering dropping old-style support by default](https://github.com/openstreetmap/osm2pgsql/pull/580#issuecomment-293373874) and if we did the same here, it would avoid all the questions on how to handle old-style MPs.

nebulon42:
Regarding old-style MPs I would be much in favour of that. I'm sure that the cleanup will progress to a point where this is no problem anymore.

math1985:
That's great news. I would also be strongly in favour of dropping support for old-style multipolygons.

pnorman:
This is now ready. Before merging we want to

- clean out the PR backlog,
- do another full planet demo,
- release 3.3.0, and
- impose a feature freeze for cartographic changes after 3.3.0 until 4.0.0 is released

openstreetmap-carto.lua
lonvia: For example, `highway=motorway + service=yard` will return 0. It's an unlikely combination but still likely not what you want to express.

openstreetmap-carto.lua
lonvia: Both.

pnorman:
I squeezed some more disk space out of one of my servers, so it looks like I can do a test over the weekend, then we can proceed.

pnorman:
https://lua.osm-carto.paulnorman.ca is running with the latest code, I'm looking to tag and freeze in a couple days.

imagico:
Nice.  Comparison is somewhat difficult since the style version is different in the standard map at the moment and the low zoom tiles are also very old there right now.

Am i right to assume that this does not yet use libosmium for geometry building (https://github.com/openstreetmap/osm2pgsql/pull/684)?

math1985:
Looks good!

I had no time to analyse this yet, but here is a difference:
https://lua.osm-carto.paulnorman.ca/#18/49.58117/6.12971

pnorman:
> Am i right to assume that this does not yet use libosmium for geometry building

No, it's using released osm2pgsql.

pnorman:
> I had no time to analyse this yet, but here is a difference:
> https://lua.osm-carto.paulnorman.ca/#18/49.58117/6.12971

Object in question is http://www.openstreetmap.org/way/77686091 with leisure=track, so this is an expected difference, and, as it happens, a tagging error.

math1985:
> Object in question is http://www.openstreetmap.org/way/77686091 with leisure=track, so this is an expected difference, and, as it happens, a tagging error.

We might want to add a rendering rule though. I'd propose to render it equal to racetracks for now.

imagico:
Yes, we should render leisure=track on lines but no, not like racetracks.  The latter is for motor sports, the former for running sports, very different things we also render very differently.

Ideally i think leisure=track on lines should be the same as on areas, i.e. line with thin casing.  And preferably in a way that combines with areas, i.e. render casings below and fills above areas.

imagico:
Ok, i inspected the map a bit more and it seems to me most of the differences that are not either due to the style version difference or due to data changes are

* the vanishing of independently labeled multipolygon components and the resulting duplicate labels which is mostly a very positive effect (see #59).
* the different interpretation of closed ways with `leisure=track` as @math1985 already spotted.
* the lack of transfer of member way tags to multipolygon relations due to the removal of support for old style MPs.  Like for example here:

http://www.openstreetmap.org/way/100827142

This water area tag is now interpreted while previously it was ignored because it was redundant with the relation tag so the islands in the river vanish.  I expected this effect to be much more widespread, especially for water areas since in the past it had been common practice to simply add `waterway=riverbank` to all outer ways of a a river polygon and if a water area tag was added to the relation it was not removed from the ways.  But it is less frequent than i thought although still pretty visible in some areas, in particular India and China.  Since this particular kind of _faulty double tagging_ is not covered by OSMI and Osmose and this is kind of connected to the elimination of old style multipolygons because it only matters if you ignore old style tagging it might be useful to think about including these in the multipolygon cleanup effort - cc @joto.

pnorman:
I guess the question is, do we want to hold up 3.3.0/4.0.0 to make any of these changes?

imagico:
I don't think so.  If anyone wants to add rendering of `leisure=track` as lines that would be most welcome but it does not really matter much if this happens before or after the release.  It is not that anything that has been rendered correctly before is rendered incorrectly with this change and would be rendered exactly right if we add new support for this.  We drop support for rendering closed ways with `leisure=track` and without `area=yes` as areas and we would in the future like to add support for rendering them as lines.

And the faulty double tagging can be fixed as easily after the change than before - in fact if the change is rolled out the motivation for fixing is probably higher since it would fix a visible error while right now it would be an invisible change.

math1985:
> I guess the question is, do we want to hold up 3.3.0/4.0.0 to make any of these changes?

As far as I'm concerned, no.

pnorman:
I'll look at tagging 3.3.0 tonight, which will also involve a feature freeze until 4.0.0.

sommerluk:
A big “thanks” for making this working!

PR #2534, opened 2017-01-03
Order road names to avoid metatile edge issues

pnorman:
@TomH noticed differing labels between the lua branch and v3.0.1

image
While it's possible this particular example is from a myriad of other causes, having an unspecified ordering can lead to labeling problems at MT edges, or if using multiple servers, labeling problems at tile edges as well.

I did some artificial tests and this helps, but I don't have good screenshots.

### DIFF
--- a/project.mml
+++ b/project.mml
@@ -1789,20 +1789,41 @@ Layer:
             refs
           FROM (
             SELECT
-                way, highway,
+                way,
+                osm_id,
+                highway,
                 array_length(refs,1) AS height,
                 (SELECT MAX(char_length(ref)) FROM unnest(refs) AS u(ref)) AS width,
                 array_to_string(refs, E'\n') AS refs
               FROM (
                 SELECT
                     way,
+                    osm_id,
                     highway,
                     string_to_array(ref, ';') AS refs
                 FROM planet_osm_roads
                   WHERE highway IN ('motorway', 'trunk', 'primary', 'secondary')
                   AND ref IS NOT NULL
               ) AS p) AS q
-          WHERE height <= 4 AND width <= 11) AS roads_text_ref_low_zoom
+          WHERE height <= 4 AND width <= 11
+          ORDER BY
+            CASE
+              WHEN highway = 'motorway' THEN 38
+              WHEN highway = 'trunk' THEN 37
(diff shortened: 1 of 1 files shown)


imagico:
Sounds logical.  The code is rather complex though.  How important is it to sort by highway type?  Wouldn't the metatile issue already be solved by sorting just by osm_id?

Apart from that i would intuitively give short labels priority over long names because that would allow showing more different labels on average but i guess that depends on the typical length of roads in comparison to the length of the labels and the general road network density.

Shouldn't the layer have priority over the highway type?

imagico:
Additional note: should not the same logic be applied to the road shields?

pnorman:
> Sounds logical. The code is rather complex though. How important is it to sort by highway type?

It's duplicating the roads fill layer logic, which is similarly complex. Lua should get rid of a lot of this, dropping the table in the query

> Wouldn't the metatile issue already be solved by sorting just by osm_id?

Yes, but it's good to have a sensible priority to have more important names shown first.

> Shouldn't the layer have priority over the highway type?

This way a name will be shown for a motorway tunnel running under a residential area. I think this way is better for names, but haven't given it a great deal of testing.

> Apart from that i would intuitively give short labels priority over long names because that would allow showing more different labels on average but i guess that depends on the typical length of roads in comparison to the length of the labels and the general road network density.

Long names are harder to place, so there are advantages to getting them placed first. But I haven't really tested this in any detail either.

> Additional note: should not the same logic be applied to the road shields?

Yes, adding it. The roads shield layer is actually technically quite different, so it'll be done in a different way, but same end result.

pnorman:
I'm inclined to go ahead with this as-is and we can play with it later. Right now we have no order defined, so any of those outcomes are acceptable.

imagico:
Looks fine in testing - Kosmtik still generates a large number of tile edge inconsistencies but i assume this is not necessarily representative for real use.

One problem i noticed - which is however also an issue without this change probably - is when there is a bridge crossing and the lower road's shield is placed above the crossing making it likely the shield is attributed to the wrong road by the user, especially if the colors are similar.  Giving the layer priority over the highway type could make this less likely but would not completely avoid it.

Issue #1997, opened 2015-12-06
Major buildings are rendered twice


nebulon42:
Currently major buildings have their own layer, which renders them twice. First as "normal" building and then as major one on top of the normal one. I think this is unnecessary and buildings should be handled by only one layer.

Here would be my proposal how to rewrite the buildings query to do that (which also removes the building catch-all rendering):

```
(SELECT
    way,
    building,
    'type_' || CASE WHEN building IN ('castle', 'cathedral', 'chapel', 'church', 'manor', 'monastery', 'mosque', 'temple')
        AND (amenity = 'place_of_worship' OR historic = 'castle')
      THEN 'major'
      ELSE
        CASE WHEN building = 'yes'
        THEN
          CASE WHEN amenity = 'place_of_worship' OR historic = 'castle'
          THEN 'major'
          ELSE 'normal'
          END
        ELSE
          'normal'
        END
      END AS category
  FROM planet_osm_polygon
  WHERE building IN ('apartments', 'barn', 'bridge', 'bungalow', 'bunker', 'cabin', 'castle', 'cathedral', 'chapel',
      'church', 'civic', 'college', 'commercial', 'construction', 'cowshed', 'detached', 'dormitory',
      'farm', 'farm_auxilliary', 'garage', 'garages', 'greenhouse', 'hangar', 'hospital', 'hotel', 'house',
      'houseboat', 'hut', 'industrial', 'kindergarten', 'kiosk', 'manor', 'manufacture', 'monastery', 'mosque', 'office',
      'public', 'residential', 'retail', 'roof', 'ruins', 'school', 'service', 'shed', 'slurry_tank', 'stable',
      'static_caravan', 'supermarket', 'terrace', 'temple', 'train_station', 'transformer_tower', 'transportation',
      'university', 'warehouse', 'yes')
    AND way_area > 0.01*!pixel_width!::real*!pixel_height!::real
  ORDER BY z_order, way_area DESC
) AS buildings
```

This introduces a category column, which is either 'type_normal' or 'type_major' and can be used in the style file. (The query contains also a logic, which includes buildings tagged as castle in the major category.)

Related to #1897.

kocio-pl:
I still think that the "approved list" approach, which generally works with shops (with some problems, because it's meant to be one-time hack, which would be enough for a static data set, but hard to maintain in the long run in the OSM case), is too simplistic for buildings, because we have a lot more of them in the database.

My questions:
[...]

PR #2536, opened 2017-01-07
Fixing rendering major buildings twice

kocio-pl:
Resolves https://github.com/gravitystorm/openstreetmap-carto/issues/1997.

SQL fix that prevents from rendering major buildings twice (as normal building and as a major building on top of that).

### DIFF
--- a/project.mml
+++ b/project.mml
@@ -424,6 +424,9 @@ Layer:
           FROM planet_osm_polygon
           WHERE building IS NOT NULL
             AND building != 'no'
+            AND (aeroway IS NULL OR aeroway != 'terminal')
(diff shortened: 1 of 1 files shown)


project.mml
```diff
@@ -420,10 +420,15 @@ Layer:
       table: |-
         (SELECT
             way,
-            building
+            building,
+            amenity,
+            aeroway
```
pnorman: Why do we need to return these columns?

kocio-pl: I thought all the called keys need to be put here, but the code works without these 2, so I removed them and pushed a new version.

kocio-pl: I thought all the called keys need to be put here, but the code works without these 2, so I removed them and pushed a new version.

Issue #2531, opened 2017-01-03
Add rendering for landuse=logging


kocio-pl:
23 787 uses already and has a wiki page. I think it should be rendered the same as landuse=industrial, since logging is clearly related to wood industry.

dieterdreist:
sent from a phone

> On 3 Jan 2017, at 03:34, kocio-pl <notifications@github.com> wrote:
> 
> 23 787 uses already and has a wiki page. I think it should be rendered the same as landuse=industrial, since logging is clearly related to wood industry.


-1, I would rather render it like forest, maybe like farmland but definitely not like industrial

kocio-pl:
I was also thinking about industrial background + tree pattern. Forest rendering makes little sense, since they are mostly inside the forest (so we have no white spaces, but we're effectively not able to recognize them), and I guess farmland is rather for growing than cutting.

pnorman:
This seems to only be used in Russia and parts of Eastern Europe.

dieterdreist:
2017-01-04 10:12 GMT+01:00 Paul Norman <notifications@github.com>:

> This seems to only be used in Russia and parts of Eastern Europe.



to me this tag has a lot of problems and it seems to require a subtag
"logging":

https://wiki.openstreetmap.org/wiki/Tag:landuse%3Dlogging
This wikipage is substandard and likely would not have passed our standard
proceduce for tag development (proposal process) in this form.

from the wiki:

"Describes logging areas. For the natural or intentional reforestation
areas may be used in combination with natural
<https://wiki.openstreetmap.org/wiki/Key:natural>=scrub
<https://wiki.openstreetmap.org/wiki/Tag:natural%3Dscrub>. "


--> scrub explicitly is not for forest areas (which would be both,
"intentional" and "natural reforestation" )


from the wiki:

"Additional tags:

   - logging
   <https://wiki.openstreetmap.org/w/index.php?title=Key:logging&action=edit&redlink=1>
   =clearcutting
   <https://wiki.openstreetmap.org/w/index.php?title=Tag:logging%3Dclearcutting&action=edit&redlink=1>
   - all trees in an area are uniformly cut down;
   - logging
   <https://wiki.openstreetmap.org/w/index.php?title=Key:logging&action=edit&redlink=1>
   =selective_cutting
   <https://wiki.openstreetmap.org/w/index.php?title=Tag:logging%3Dselective_cutting&action=edit&redlink=1>
[...]

PR #2537, opened 2017-01-07
Adding rendering of landuse=plant_nursery

kocio-pl:
Resolves https://github.com/gravitystorm/openstreetmap-carto/issues/1772.

vpb_7rtj
5axxl gk


polarbearing:
Thanks. However the examples confuse me a bit. How are they tagged?

Is the nursery the smaller green area in both? In the second, where does the outline come from, and is the tree pattern from a larger forest landuse?

kocio-pl:
Yes, both are in the middle. I wanted to show how the same color looks in different rather popular environments. 

This is the first one:
http://www.openstreetmap.org/way/154336775

The second one is just fenced and being part of the forest:
http://www.openstreetmap.org/way/336901548

polarbearing:
Fine then, this explains it. This was my example from the issue, next to a forest and sports environment: http://www.openstreetmap.org/way/9759196

kocio-pl:
Here you are:
xxj2lg_2

polarbearing:
Beautiful :-)

imagico:
If an area color is used with and without a pattern the version without usually indicates a less specific, more general variant.  This is not the case here.  Therefore i would try a pattern as suggested in #1772 - it does not have to be exactly as suggested there, other variants could work as well.

The wiki page needs a clearer definition that makes clear that a site where plants are going to grow old in place is not a plant nursery - otherwise people could start tagging any newly planted forest a plant nursery.

kocio-pl:
Aren't you afraid that it will look like an orchard?

:+1: for wiki update.

imagico:
It should look like a different kind of orchard (or more precisely: both should look like variants of the same base class, i.e. intensive care permanent plant growth for a specific purpose).

kocio-pl:
Is there a way to save the pattern from your script as SVG file instead of base64 data in URL? I would like to start using [vector patterns](https://github.com/gravitystorm/openstreetmap-carto/issues/2045).

polarbearing:
I suggest to merge this as an improvement over not rendering at all, and discuss a potential pattern afterwards.

imagico:
@kocio-pl - you can save the data link like any other link (right click or open and _save page as..._)

@polarbearing - as i like to put it: _the good enough is the enemy of the great_, we should aim for the best solution possible under the circumstances, settling for a good enough discourages work on making it great.

You could also see it more formally - settling for a good enough would mean closing one issue but opening another (for making it better).  The aim should be to reduce the number of of open issues and not just replacing the more severe ones with less severe ones.

polarbearing:
Well, the new issue would be about plant lifecycles in general, there is some need for discussion as well as we learned from the [wood logging example](#2531).

kocio-pl:
It's not user friendly solution, but I did it. Worse problem seems to be that the vector pattern is not visible at all, even when making background white or light grey. Changing the code line with orchard pattern shows orchard dots, so it seems that either SVG file is not suitable for Mapnik or there is some software error. Could anybody else test it?

imagico:
For direct use in mapnik you probably need to render inlined.  See https://github.com/imagico/jsdotpattern/issues/3

kocio-pl:
> as i like to put it: the good enough is the enemy of the great, we should aim for the best solution possible under the circumstances, settling for a good enough discourages work on making it great.

You probably know it and this is simply your statement, but original quotation was actually a big red warning against "perfection" hell.

> You could also see it more formally - settling for a good enough would mean closing one issue but opening another (for making it better). The aim should be to reduce the number of of open issues and not just replacing the more severe ones with less severe ones.

In this case it's not too much work to add desired pattern and your proposition makes sense for me, so I'm just willing to do it and there's no problem with this ticket. 

But while the aim is **exactly** how I see it, this attitude doesn't actually help closing any issues - they stay unresolved for such a long time that we've lost any trace of some of them, even with tags. I sometimes feel as the ticket archeologist (Peter is also trying to dig there). Hard numbers tell we have about 380 open issues and this is not getting any lower despite some PRs that could do it (I know: because they _could_ be better).

This is exactly what I call "[analysis paralysis](https://en.wikipedia.org/wiki/Analysis_paralysis)" and "[overengineering](https://en.wikipedia.org/wiki/Overengineering)": 

_Overengineering can be desirable when safety or performance on a particular criterion is critical, or when extremely broad functionality is required, but it is generally criticized from the point of view of value engineering as wasteful._

We have finite supply of resources, the project is not about nuclear power station, the data is changing constantly and small changes accumulated proved to be effective. Issues closing rate shows me that it's working the other way around: _perfect is the enemy of good_.

What really works is for example suggesting possible solutions, just like in this ticket. If you could find the issues that could be fixed, it would help exterminate some of them. This is also what I would really appreciate.

kocio-pl:
Making right vector pattern seems to be harder than expected and I have questions about [the link you gave](http://www.imagico.de/map/jsdotpattern.php#x,128,jdp32360;gv,6,32,32;rd,1,0,0,pixel,0.125,4,4,0,jdp23814,749d6c,aedea3;):
1. Why the size of the pattern is 128 and not something much smaller, when it's so regular?
2. What do you mean by render inlined? Clicking "render (inlined)" does produce SVG file that looks like this:
b9pki kc

imagico:
> Why the size of the pattern is 128 and not something much smaller, when it's so regular?

That is explained on http://blog.imagico.de/some-pattern-math-regular-patterns/ - the closest hexagonal dot packing has an irrational aspect ratio so the mismatch from the ideal grid depends on the size the pattern is rendered at.  It is possible that in this particular case you can also use a smaller size - i have not tried.  But in general this is the problem here.  Of course if you ultimately want a pixel aligned pattern the whole thing is fairly moot although the whole point in using an SVG directly is to be resolution independent.

> What do you mean by render inlined? Clicking "render (inlined)" does produce SVG file that looks like this:

As said on https://github.com/imagico/jsdotpattern/issues/3 i have never tested the results with mapnik.  If there is a problem bug reports are welcome (although i can't promise a fix, it depends on how difficult mapnik is in that regard).   Have you tried opening the SVG in inkscape, saving it there and using that?

imagico:
Regarding over-engineering:

You seem to neglect that this is primarily a design project and not an engineering project.  In engineering aiming for perfection is often frowned upon because of diminishing returns.  Above a certain level of quality the effect improvements in engineering have on the ultimate results (like a product manufactured or a service) are negligible and the improvements - if at all - have an effect only internally, like on reliability and efficiency.  Therefore at least from a purely short term economical perspective you aim not to over-engineer.

However in design we do not have this phenomenon.  Every imperfection in design has a direct influence on the quality of the results.  So you cannot really transfer common wisdom from engineering here.

But even in engineering related aspects of this project there is a long tradition in not sacrificing overall long term aims for short term improvements - something i also learned the hard way by the way - see https://github.com/gravitystorm/openstreetmap-carto/pull/646#issuecomment-56063177

And regarding open issues - a huge portion of issues that have been open for a long time are fixes for problems and not feature additions.  As i have pointed out multiple times already we should aim more at addressing those than adding new features which - even with high standards in design that are not satisfied with a good enough solution - will inevitably lead to new issues.  Just look at https://github.com/gravitystorm/openstreetmap-carto/issues?q=is%3Aopen+is%3Aissue+label%3Aeasy for example.

kocio-pl:
I was not able to render SVG file properly, so I stick with PNG for now. Rendering with pattern:
v1znmaco
raivwt8p
tv0oh6sb

imagico:
Looks all right to me - although contrast could probably be somewhat weaker so the overall weight of the pattern is not larger than for orchard.

kocio-pl:
What color do you suggest other than what was included in the link? The only visible difference for me is more dense pattern.

math1985:
Thanks, looks very nice!

Issue #738, opened 2014-07-19
Render building=greenhouse differently


matthijsmelissen:
Most buildings are about the same, but greenhouses look completely different, both, from on the ground and from aerial pictures.

But when you're on a place where there are a lot of greenhouses on the ground, the map looks just the same as when you're in a factory.

Currently, there are over 15.000 objects tagged with "building=greenhouse".

So I would propose a different rendering for that. I suggest a very pale version of the current building colour would do. Something like #EBDDDB (even paler than houses) 

See also https://trac.openstreetmap.org/ticket/4654

mboeringa:
How I currently render it. Scale is unfortunately not exactly the same, but close. I think there is room for some dedicated renderings in the building class. Left Standard, right ArcGIS Renderer for OpenStreetMap. Click to enlarge:

![arcgis_renderer_for_openstreetmap_greenhouses](https://cloud.githubusercontent.com/assets/7635726/8170448/f0e826e6-13b0-11e5-943e-926535f97649.png)

dieterdreist:
sent from a phone

> Am 15.06.2015 um 22:52 schrieb mboeringa <notifications@github.com>:
> 
> I think there is room for some dedicated renderings in the building class.


I support this. Greenhouses are too dominant currently.

kocio-pl:
So how do you think they should look like?

ffes:
There are two options.

The red used on the example above looks good to me. That would be my preference.

Otherwise a type of green. These things aren't called greenhouses for nothing.

kocio-pl:
We use green for different natural/leisure purposes. 

I don't see red - it's violet for me and we use it for industrial landuse.

I missed `Something like #EBDDDB (even paler than houses)`. This sounds reasonable for me. In fact it looks like a subset of https://github.com/gravitystorm/openstreetmap-carto/issues/1207, which is currently closed, but [there was no strong opinion against it](https://github.com/gravitystorm/openstreetmap-carto/issues/1207#issuecomment-112149155), so I would go this way for all minor buildings.

HolgerJeromin:
What about an greenish gray as a mixture of nature and building?

kocio-pl:
I like the idea of having just minor, normal and major building colors (and roofs, but it's not planned to have a separate color), and not being too specific.

I'm not sure how would this look like, but I would probably test it too, out of curiosity.

kocio-pl:
[...]

PR #2538, opened 2017-01-07
Render greenhouse buildings as light brown

kocio-pl:
Resolves https://github.com/gravitystorm/openstreetmap-carto/issues/738.

Near Hague, z13:
Before
df7xf2tx
After
xkhooue4

Warsaw, z17
Before
ijdt0pa2
After
b5uk5e2d

### DIFF
--- a/buildings.mss
+++ b/buildings.mss
@@ -19,6 +19,14 @@
       line-width: .75;
       line-clip: false;
     }
+    [building = 'greenhouse'] {
+      polygon-fill: lighten(@building-fill, 3%);
+      [zoom >= 15] {
+        line-color: lighten(@building-line, 3%);
+        line-width: .75;
+        line-clip: false;
+      }
+    }
   }
 }
 


kocio-pl:
Do you have similar examples?

pnorman:
image

image

I don't see that any of these is really much worse than the others, and I tfeel adding a colour like this steps back on the simplifications we did at the time of the building colour changes.

imagico:
DeltaE to mud is 3.2 which is too small for comfort.  Similarity to landuse=residential is also too close for a completely different class of feature - you can see that in your example where the user might easily read this as yet another class of urban land cover.

I don't think you are going to be able to squeeze in another area color in out already very overcrowded color scheme, especially with relatively weak and subtle colors.

kocio-pl:
@pnorman These are not too similar, one should try with z13-z14 to make it comparable. On z17+ there are much more elements as you can see and buildings are not as visible as when there are no icons. Also the building color on z15+ is made lighter than on earlier zoom levels.

kocio-pl:
@imagico At z13-z14 it looks like something more important than towns around, which is worse kind of misguiding users than what you suggest. Brown is basically only building-related and already used in variants, so I'm not afraid of extending it more - it will make using brown even more clear and consistent rule, which would be good.

imagico:
I disagree on both the assessment of the specific situation and the interpretation of the meaning of colors in general.  Colors are a continuum, there is no such thing as a line between brown colors and non-brown colors.  We have three building area colors currently in this style, all of which are more similar to other non-building area colors than to each other (see also #2353) so there is no such thing as a building-brown that can be extended with additional variants.

While i think it would be good to render building=greenhouse in a distinct way (though not necessarily via color) i don't think the sample area presented shows a particular problem in its current appearance.  I have not been to this particular area but in general greenhouses like these are very specific, prominent and dominating features in a landscape and it is not unreasonable to render them significantly more prominently than a generic landuse.

kocio-pl:
There's also no such thing as a line between green and non-green colors. Green however "should be reserved for vegetation related features". Yet it's also used for sport and leisure, which are something else, and nobody complains it could be misunderstood. And green has even no continuum, as brown does, we have just a set of unrelated green shades. Doesn't it disturb you as a bigger problem?

imagico:
The vegetation related green tones in this style form a distinct group in color space, i.e. they are (with a few exceptions) closer to each other than to the rest of the area colors.  Some of the recent changes (playground and pitch/track) blurred the boundary by moving not strictly vegetation related borderline green tones towards the center of vegetation related greens.

I explained this also in the original version of the guidelines document on:

https://github.com/imagico/openstreetmap-carto/blob/585efe12d96dc3233c495fd54a3e5d9a967be609/CARTOGRAPHY.md

kocio-pl:
Before that the problem was playgrounds were too close to the pools (the same problem for blue this time), so nothing new arised and the problem is milder now, but it's side tracking the issue. Let's get back to the core - sport was green also before any shading, and it's not more intuitively green than buildings brown. 

Green set is blurred now and was blurred before. Brown is not and was not - hence "features rendered with similar color should actually be similar in purpose and meaning" (not my words, but I agree) is more in favor of brown than green.

imagico:
I don't see this discussion leading anywhere - i explained my viewpoint, in particular

> We have three building area colors currently in this style, all of which are more similar to other non-building area colors than to each other

The color differences can be calculated by anyone interested - both these and the green and blue tones for comparison.

kocio-pl:
Maybe I don't understand your vocabulary - if shades are "colors" for you? Then all light colors and all dark colors would be alike. That doesn't make sense for me.

imagico:
I don't think i talked about _shades_ in this context.

kocio-pl:
The only difference here is using lighter kind of brown - but still brown - and you claim it will be too close to other light colors. So I assume you were talking about lightness, not about colors.

imagico:
deltaE measures visual differences in both lightness and chroma/hue, see https://en.wikipedia.org/wiki/Color_difference.  If anything the difference in lightness is more meaningful because display devices provide better distinction in lightness than in chroma/hue.  This is why the wood/water color combination is so annoying despite the colors having a relatively large difference (deltaE is 27.5).

mboeringa:
I think that what is a bit forgotten in the whole disussion regarding colors, lightness, chroma, deltaE or whatever other measure one would desire to use to describe differences, is that people not only "read" the contents of a map by looking at "color", but use other (visual) clues as well.

E.g. greenhouses tend to:

- be large and "square"
- don't carry addresses
- won't have "shop" or crafts type icons in a style like this one
- are usually located near farm houses (but the extreme example of The Hague shows the entire land can be covered with them - it's absolutely horrible to be honest...)
- "locals" tend to know their existence well

So there are more clues that set apart certain objects and allows recognition. This is just one example, similar things hold for other types of objects.

kocio-pl:
> are usually located near farm houses

...and locating it on mud would be highly surprising (while water in the woods is quite common case), so what's the use of measuring deltaE in such case? On the pictures there are much more typical and that's what is interesting, not the corner cases.

math1985:
> The problems with being too dominant are not unique to greenhouses, but an issue with all buildings when they're in that kind of density

+1

dieterdreist:
sent from a phone

> On 21 Mar 2017, at 22:23, Matthijs Melissen <notifications@github.com> wrote:
> 
> The problems with being too dominant are not unique to greenhouses, but an issue with all buildings when they're in that kind of density
> 
> +1
> 


-1, it's a feature that dense building structure leads to dense rendering

Issue #2375, opened 2016-09-26
Add leisure=fitness_centre icon/text


boothym:
Currently leisure=fitness_centre is not rendered (neither is fitness_station as mentioned in #1870). 

Even without an icon at the very least the name should be displayed, just like sports_centre. 

At the moment any properly tagged gym is invisible which doesn't help encourage people to tag it correctly.

micahcochran:
There are now 4789 instance of this tag.

How about this as an icon?
![weight lifter 2](https://cloud.githubusercontent.com/assets/7433104/19945714/5dc24dd6-a10f-11e6-8c7f-2fd0957fc3a0.png)

boothym:
I notice in the maps.me app it is rendered with weights:  https://github.com/mapsme/omim/blob/e646b270bd68d0c27adac5ed712682cf1500b68c/data/styles/clear/style-night/symbols/gym-l.svg - I wonder if they would allow it to be used?

boothym:
@gravitystorm @math1985 @pnorman @matkoniecz 

Any chance we could get leisure=fitness_centre rendered the same way as a sports_centre node, so that the name is visible? Right now there are many correctly tagged gyms etc. which aren't rendered on the map. An icon can be added or text fine-tuned later.

boothym:
@kocio-pl while we wait for an icon, any chance of the fitness_centre being rendered exactly like sports_centre, so that the text is visible?

kocio-pl:
I do few other things here, so if somebody creates the code, that would be great. Are you willing to do that?

boothym:
If all that's required is adding fitness centre beside sports centre in project.mml then I will submit a PR with that change soon.

kocio-pl:
Look for sports centre also in MSS files - project.mml is more or less a database part, styling is defined in MSS.

kocio-pl:
See https://github.com/gravitystorm/openstreetmap-carto/issues/1870#issuecomment-335064319

Issue #844, opened 2014-08-05
Differentiating sport infrastructure (like pitches) by type of sport


matthijsmelissen:
Currently all sport fields are shown in the same style. It is very difficult to distinguish the different sports. Furthermore there are good sport icon sets out there:  http://wiki.openstreetmap.org/wiki/Category:Sport_icon Comparing the different olympic pictograms  http://www.nytimes.com/interactive/2010/02/24/sports/olympics/pictograms-interactive.html I suggest to take the icon sets of Olympic games 1972 (munich)  http://olympic-museum.de/pictograms/Picto1972.htm adopted to the constraints of 16px 

See also https://trac.openstreetmap.org/ticket/3159.

boothym:
How about adding icons for the most used sports to begin with? Soccer, tennis, baseball, basketball, all have good icons that can be used from iD editor's icon set. Just like iD the icon can be put in the middle of the pitch.

https://github.com/bhousel/maki/blob/mb-pages/src/soccer-24.svg
https://github.com/bhousel/maki/blob/mb-pages/src/tennis-24.svg
https://github.com/bhousel/maki/blob/mb-pages/src/baseball-24.svg
https://github.com/bhousel/maki/blob/mb-pages/src/basketball-24.svg

kocio-pl:
Unfortunately they all are 24px, while we use just 14px - this is huge difference. If you want to see how they would look in this style, try to rescale them and save as PNG.

boothym:
I linked to them because I thought you'd want the largest svg size available, but here's what they look like at 14px:
![tennis](https://cloud.githubusercontent.com/assets/16965772/18996247/7b0ce524-8727-11e6-88d5-172585b3e489.png)
![soccer](https://cloud.githubusercontent.com/assets/16965772/18996248/7b0f6196-8727-11e6-86f2-0e5233a26828.png)
![baseball](https://cloud.githubusercontent.com/assets/16965772/18996249/7b0f9026-8727-11e6-912c-14cb62564035.png)
![basketball](https://cloud.githubusercontent.com/assets/16965772/18996246/7b0cc076-8727-11e6-822c-88bc39b9384e.png)

There's also https://github.com/mapbox/maki/tree/master/icons with more stick figures instead of the bat/ball icons.

kocio-pl:
I'm surprised, they look quite nice and are perfectly recognizable! That's the power of skilled designer probably. =}

Still we need to decide if we want to use icons or field markings.

[...]

PR #2547, opened 2017-01-10
Add fitness_centre with the same rendering as sports_centre

boothym:
#2375 - at least gets the fitness_centre name rendered, we can get an icon added later.

I don't have a way to test this PR, but I assume it will work ok as I've just copied the parts where sports_centre is mentioned in the code.

### DIFF
--- a/amenity-points.mss
+++ b/amenity-points.mss
@@ -1574,6 +1574,7 @@
   [feature = 'natural_beach'],
   [feature = 'natural_shoal'],
   [feature = 'natural_reef'],
+  [feature = 'leisure_fitness_centre'],
   [feature = 'leisure_sports_centre'],
   [feature = 'leisure_stadium'],
   [feature = 'leisure_track'],
@@ -1688,6 +1689,7 @@
       [feature = 'natural_shoal'] {
         text-fill: darken(@beach, 60%);
       }
+      [feature = 'leisure_fitness_centre'],
       [feature = 'leisure_sports_centre'],
       [feature = 'leisure_stadium'] {
         text-fill: darken(@stadium, 70%);
--- a/landcover.mss
+++ b/landcover.mss
@@ -59,7 +59,7 @@
 
 @pitch: #aae0cb; // also track
 @track: @pitch;
-@stadium: @societal_amenities; // also sports_centre
+@stadium: @societal_amenities; // also fitness_centre and sports_centre
 @golf_course: #b5e3b5;
 
 #landcover-low-zoom[zoom < 10],
@@ -528,6 +528,7 @@
     polygon-fill: @station;
   }
 
+  [feature = 'leisure_fitness_centre'],
   [feature = 'leisure_sports_centre'],
   [feature = 'leisure_stadium'] {
     [zoom >= 10] {
--- a/project.mml
+++ b/project.mml
@@ -141,7 +141,7 @@ Layer:
                                                     'recreation_ground', 'village_green', 'retail', 'industrial', 'railway', 'commercial', 
                                                     'brownfield', 'landfill', 'construction') THEN landuse ELSE NULL END)) AS landuse,
               ('leisure_' || (CASE WHEN leisure IN ('swimming_pool', 'playground', 'park', 'recreation_ground', 'common', 'garden', 
-                                                    'golf_course', 'miniature_golf', 'picnic_table', 'sports_centre', 'stadium', 'pitch', 
+                                                    'golf_course', 'miniature_golf', 'picnic_table', 'fitness_centre', 'sports_centre', 'stadium', 'pitch', 
                                                     'track', 'dog_park') THEN leisure ELSE NULL END)) AS leisure,
               ('military_' || (CASE WHEN military IN ('danger_area') THEN military ELSE NULL END)) AS military,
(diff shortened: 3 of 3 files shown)


imagico:
We generally require changes to be tested and not be done blindly.  The inner workings of the style and interactions between different parts can be complicated and even experienced contributors will not usually submit changes without testing.

If you need help with setting up a testing environment there are usually people here who can give suitable tips and suggestions.

Regarding the change itself - there are less than 100 features with `leisure=fitness_centre` without a building tag and most of these seem to be errors of some sort (missing building tag or misuse of the tag for a spa or similar) so not a good idea to add area rendering for this IMO.

boothym:
Well it's there for anyone who wants to test it, as unfortunately I don't have the time or inclination to do it at the moment. 

Not sure where you are getting your numbers from, as there are over 200 instances of `leisure=fitness_centre` without a building tag (ie nodes) in the UK alone?

The reason for the change is to have the same rendering as `leisure=sports_centre`, especially for nodes where the name is displayed for sports centre, but not shown for fitness centre. There are many incorrectly tagged gyms which would not be rendered if they were changed to fitness centre.

boothym:
If you want an example of why this change is needed, see: http://www.openstreetmap.org/node/208364293/history

This gym was correctly updated from leisure=sports_centre to fitness_centre, but now its name is no longer rendered on the map.

imagico:
Note i was talking about area features and area rendering here.

Rendering tagged nodes with label only is a different matter.  This is a problem too but likely not any worse if this is extended to fitness centres.  Right now sports_centre is widely used for name dropping, just look at the number of nodes with this tag in the database that have a generic name indicating the type of sport or site rather than the specific name of the feature in question.  

Examples:

https://www.openstreetmap.org/node/2379021641
https://www.openstreetmap.org/node/286688123
https://www.openstreetmap.org/node/4452411798
https://www.openstreetmap.org/node/519628832
https://www.openstreetmap.org/node/250386927
https://www.openstreetmap.org/node/1231870683
https://www.openstreetmap.org/node/1860280192

dieterdreist:
2017-01-13 11:15 GMT+01:00 Christoph Hormann <notifications@github.com>:

> Rendering tagged nodes with label only is a different matter. This is a
> problem too but likely not any worse if this is extended to fitness
> centres. Right now sports_centre is widely used for name dropping, just
> look at the number of nodes with this tag in the database that have a
> generic name indicating the type of sport or site rather than the specific
> name of the feature in question.



that's a fine line, some/most of them could be considered names (e.g.
"Reithalle", "Sporthalle") while the first on (a node with
name="Volleyball") clearly isn't. A generic name can also be the specific
name, especially in smaller settlements where there's just one of a kind
and no need to distinguish (I don't know the objects you have posted here
and cannot say if this is the case here)

imagico:
No, the difference here is generic type of sport vs. generic type of site.  Neither is correct as a name, a name tag is always for the individual name of the feature in question which differentiates it from similar features nearby.  Having the type of site as name would be like placing a node `amenity=place_of_worship` with `name=church` - which is wrong.

In this particular case the name of the sports centre would be _Reitverein St. Hubertus_ which is in the description tag here.

The fact that in conversation locals might identify the feature by its type (like "wir gehen zur Reithalle") does not change this - we also tend to say "wir gehen zum Bäcker/Supermarkt/..." which does not mean that "Bäcker" is a correct name tag for shop=bakery.

There are of course cases where in terms of map usability it might be beneficial to render localized values for tags like `sport=*` or `shop=*` instead of or in addition to generic icons and names but that is technically challenging due to the variety of languages in the map.

dieterdreist:
2017-01-13 11:45 GMT+01:00 Christoph Hormann <notifications@github.com>:

> No, the difference here is generic type of sport vs. generic type of site.
>


yes


> Neither is correct as a name, a name tag is always for the individual name
> of the feature in question which differentiates it from similar features
> nearby.
>

as I have written, the individual name can be the same as the generic type
of site. Even more as we focus on "ground truth", i.e. the name the local
people use

boothym:
@imagico I'm not sure why you are bringing up examples of sports_centre - this PR is about getting fitness_centre rendered. 

Regarding names of sports on pitches, I've suggested sport icons https://github.com/gravitystorm/openstreetmap-carto/issues/844#issuecomment-250768588

pnorman:
While the cartography needs discussing, it's also important that this PR be tested.

kocio-pl:
I think of leisure=fitness_centre rather as leisure, not sport, so prefer to use the same color as some other leisure features.

Testing rendering seems to work both for nodes and area:
xipu563f
udfvbov7

If you want to test anything more, I need the location.

boothym:
Looks fine to me for nodes, which is what I was concerned about. Not really bothered about areas as fitness centres are normally nodes are with a building tag. Though it would be nice if the same text was used on buildings, instead of the normal text.

math1985:
Thanks @boothym! And sorry for slow review.

boothym:
Thanks, I wondered if something had happened to all the reviewers! :)

Just a question about buildings tagged with leisure=fitness_centre or tourism=attraction. Why aren't named buildings with these tags styled in the same way as nodes?

PR #2549, opened 2017-01-14
Clean up bus_stop SVG

pnorman:
Not yet ready to merge. The `<rect/>` element shouldn't be there, but without the icon changes size on the map.

I'm wondering if the rect is forcing the size and the bus part is smaller, so even though marker-width is 12px, the visible part is smaller. This strikes me as the wrong way to achieve that.

cc @nebulon42


nebulon42:
We are using common canvas sizes for icons but the shapes can be different. Some are 2px less wide, some are less tall and so on. Mapnik only takes the actual shapes into account and ignores the canvas size. If you would like to scale the bus stop icon and want to have it e.g. twice the size and were thinking in canvas sizes you would write `marker-width: 24px`. Without the rectangle you would get a slightly bigger icon than you wanted. In most cases this maybe wouldn't be visible but nevertheless is wrong.

Short version: The rectangle element helps with downstream customisation. The idea is from Maki icons. If there is a better way we could of course change it.

pnorman:
> Mapnik only takes the actual shapes into account and ignores the canvas size

Ah, that explains it.

Otherwise, does it look okay?

nebulon42:
I have centred the icon after the clean-up. Otherwise it looks okay to me, I have not checked the rendering on the map though. If you are ok with it I can merge it.

PR #2554, opened 2017-01-19
remove rendering for landuse=farm

nebulon42:
This removes rendering for `landuse=farm`, currently rendered like `landuse=farmland`.

`landuse=farm` is discouraged since 2013 and deprecated since 2014. There is a prominent warning on the wiki page and also the JOSM validator issues warnings about it.

[Taghistory](http://taghistory.raifer.tech/) shows a steady decline, but according to [Taginfo](https://taginfo.openstreetmap.org/tags/landuse=farm) there are still around 360 000 uses.

Removing it from the rendered tags shall speed up the re-tagging into either `farmland` or `farmyard`, but since there are still quite some uses there will be changes to the map. However, this is intended.

### DIFF
--- a/landcover.mss
+++ b/landcover.mss
@@ -17,7 +17,7 @@
 @commercial-line: #d1b2b0;  // Lch(75,12,25)
 @industrial: #ebdbe8;       // Lch(89,9,330) (Also used for railway)
 @industrial-line: #c6b3c3;  // Lch(75,11,330) (Also used for railway-line)
-@farmland: #fbecd7;         // Lch(94,12,80) (Also used for farm)
+@farmland: #fbecd7;         // Lch(94,12,80)
 @farmland-line: #d6c4ab;    // Lch(80,15,80)
 @farmyard: #f5dcba;         // Lch(89,20,80)
 @farmyard-line: #d1b48c;    // Lch(75,25,80)
@@ -282,7 +282,6 @@
     [way_pixels >= 64] { polygon-gamma: 0.3;  }
   }
 
-  [feature = 'landuse_farm'],
   [feature = 'landuse_farmland'],
   [feature = 'landuse_greenhouse_horticulture'] {
     [zoom >= 10] {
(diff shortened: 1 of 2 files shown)


kocio-pl:
Sounds good to me. I think we should write about this plan on the Talk and Tagging list (and possibly on some other popular places, like forum).

nebulon42:
Sure, I can do that.

pnorman:
iffy about this, since `landuse=farm` has a well-defined meaning and is commonly used.

nebulon42:
I guess it does, but I have cleaned up quite some usages in Austria lately and they were ambiguous. Some meant `landuse=farmyard` and some others meant `landuse=farmland`.

aceman444:
It may have a meaning (farmland), but the problem seems to be people do not know the true meaning and use it for something else (farmyard). Even if you clean up all uses of it in a region, people may still introduce the tag again on some objects and with the wrong meaning.

Also notice JOSM does not support the tag anymore.

imagico:
In general I think this is a good idea. We should of course be careful in implementing changes in tag use just because some voiceful group of active mappers decided about them.  But here we can clearly see this change is widely accepted and there is a clearly identified problem with the old tagging.

However we probably should pre-announce this intention some time in advance so there is some room to organize a concerted effort in mapping (along the lines of _everyone should have a look at those areas in his/her local mapping environment and verify + fix farmland/farmyard mapping_).  There have also been some imports in the past with `landuse=farm`, for example the French CLC import if i am not mistaken - local communities might want to have the chance to contemplate these things in advance of such a change.

Note although use of the tag is in steady decline over the past 1-2 years there are still mappers actively using the tag - might be a good idea to determine the reasons.

aceman444:
Yes, CLC was one of the "culprits" using 'farm'. However, it used it in the correct way as farmland, it probably couldn't distinguish small-area features as a farmyard.

We have cleaned the whole of our European country from landuse=farm.

math1985:
Thanks! Could any of you, perhaps @imagico, communicate this with the community?

nebulon42:
I can do that.

nebulon42:
Links to notifications:
https://forum.openstreetmap.org/viewtopic.php?pid=637717 (Forum overall)
https://lists.openstreetmap.org/pipermail/talk/2017-March/077741.html (Talk/tagging mailinglist)

PR #2565, opened 2017-02-08
Fix tyop

systemed:


math1985:
It depends a bit on who you ask whether 'examplar' is correct or not, but in any case 'exemplar' is more common.

PR #2573, opened 2017-02-19
removing waterway=canal area rendering

### LINKED ISSUE #2424: Should not render closed ways with waterway=canal as area
https://wiki.openstreetmap.org/wiki/Tag:waterway%3Dcanal says that it should not be used on areas and also to `Consider adding width=* to indicate the width in meters, for example width=5 or alternatively use waterway=riverbank to create an area for the canal in addition to the linear 'canal' feature along the centreline.`

Since it's a linear feature only, we should not have it rendered as an area when it's mapped as a closed way.

Example of a canal being wrongly used to render the whole canal area:
https://www.openstreetmap.org/way/230145073

And places where it seems to be properly mapped and tagged, but are being wrongly rendered as areas:
https://www.openstreetmap.org/way/192657911

https://www.openstreetmap.org/way/114691942

imagico:
fixes #2424.

Note this only removes the area rendering, correctly treating closed ways with waterway=canal as line features will depend on the database reload.  Current lua branch already has this covered:

https://github.com/gravitystorm/openstreetmap-carto/blob/lua/openstreetmap-carto.lua#L40

### DIFF
--- a/project.mml
+++ b/project.mml
@@ -241,7 +241,7 @@ Layer:
             way_area/NULLIF(!pixel_width!::real*!pixel_height!::real,0) AS way_pixels
           FROM planet_osm_polygon
           WHERE
-            (waterway IN ('dock', 'riverbank', 'canal')
+            (waterway IN ('dock', 'riverbank')
               OR landuse IN ('reservoir', 'basin')
               OR "natural" IN ('water', 'glacier'))
             AND building IS NULL
--- a/water.mss
+++ b/water.mss
@@ -25,8 +25,7 @@
     }
   }
 
-  [waterway = 'dock'],
-  [waterway = 'canal'] {
+  [waterway = 'dock'] {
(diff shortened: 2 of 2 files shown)


pnorman:
> Current lua branch already has this covered:
> 
> https://github.com/gravitystorm/openstreetmap-carto/blob/lua/openstreetmap-carto.lua#L40

Not really. There's two different questions - when waterway=canal is an area, and rendering waterway=canal areas. Lua changes the first, this PR changes the second.

imagico:
This is exactly what i meant - right now this PR removes the incorrect rendering of closed way canals as areas and this will make sense also after the database changes but it does not make sure they are rendered correctly as line features since the geometries are still treated as polygons in the database.

pnorman:
Looks good to me, but I haven't had a DB loaded to actually test.

PR #2574, opened 2017-02-19
better outline color for societal amenities (hospitals, schools), adding missing outline for stadium/sports_centre

### LINKED ISSUE #2498: education/hospital outline looks like highway=track at z13/14
Should be weaker and less saturated:

education at z14

imagico:
fixes #2498, #2388.

school outline

This is also significantly weaker than previously to make sure the outline is not so prone to be confused with barriers, like for example here:

http://www.openstreetmap.org/#map=18/47.98378/7.82667

### DIFF
--- a/landcover.mss
+++ b/landcover.mss
@@ -474,7 +474,7 @@
         polygon-fill: @societal_amenities;
         [zoom >= 13] {
           line-width: 0.3;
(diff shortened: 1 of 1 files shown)


math1985:
Thanks!

PR #2587, opened 2017-03-22
add rendering for aerialway=mixed_lift (same as gondola)

### LINKED ISSUE #347: please render aerialway=mixed_lift
There are not many mixed_lifts, but they are important for a ski resort.

One example:
http://www.openstreetmap.org/way/46732281#map=17/47.22377/10.12568

Photo of such a aerialway:: http://www.skiresort.de/typo3temp/pics/8ef692a525.jpg 
I do not care if cable_car.png, chair_lift.png or a new icon is used.

nebulon42:
Fixes #347 

This is a special case so it needs some explaining. This feature request was rejected first due to low usage. In this case usage is not really significant as it is a rare feature. There is at least one of those not far from where I live. Since we render other types of aerialways already it makes also sense to render this one, especially as leaving it out can lead to the weird situation that other aerialways are seemingly not reachable as in my example below.

Also there is tagging for the renderer going on as https://www.openstreetmap.org/way/31766480 illustrates. Under this circumstances I wanted to add this. Also it is mentioned on the aerialway Wiki page and it also has its own Wiki page. Given the rare occurrence of this feature I view this tag as established.

I chose the same rendering as for gondola as this type of aerialway usually consists of gondolas and chairs and so it deserves to be rendered as the higher category.

https://www.openstreetmap.org/way/2304032
before (note how Seekareckbahn, Panoramabahn, Hochalmbahn are seemingly unreachable)
before

after
after

### DIFF
--- a/aerialways.mss
+++ b/aerialways.mss
@@ -1,6 +1,7 @@
 #aerialways {
   [aerialway = 'cable_car'],
-  [aerialway = 'gondola'] {
+  [aerialway = 'gondola'],
+  [aerialway = 'mixed_lift'] {
     [zoom >= 12] {
       line/line-width: 1;
       line/line-join: round;


HolgerJeromin:
Thanks a lot!

nebulon42:
According to our change in review policy I'm merging this myself.

Issue #2444, opened 2016-11-16
Rethinking shops rendering


kocio-pl:
I think we should rethink shops rendering, because it seems [we're not done with this topic](https://github.com/gravitystorm/openstreetmap-carto/pull/2415#issuecomment-260951165). It's a tough matter, but let's look for a solid, long time solution instead of "fixing" it.

Different people might have different problems with it, like:
- too many shops at all (?)
- too many shops at some zoom level (too early)
- too many shops at some places
- too many icons
- too few icons
- catch-alls
- no tools for filtering shops
- maintainability of the list
- error checking (typos)
- amount of code needed
- shop definition (how is it different from amenity) 

...and so on. 

To avoid repeating the issues we already know and going off topic, I'd like to concentrate only on what is the most important annoyance for you personally and how far can we go with a solution to be still acceptable. Please, be precise. I see it as the only way to find a common ground instead of trying to get "as-much-as-it's-possible".

For me shops are one of the most important things on micromapping level in a big city (my main activity in OSM) - the buildings and streets being just a background there and only some amenities are that popular. If we could filter them somehow and show some of them later or make them less visible before z-max, that would be acceptable for me. The only thing I don't accept is the unmaintainable, hardcoded list not related to the rest of the project.

mboeringa:
> To avoid repeating the issues we already know and going off topic, **I'd like to concentrate only on what is the most important annoyance for you personally** and how far can we go with a solution to be still acceptable. Please, be precise. I see it as the only way to find a common ground instead of trying to get "as-much-as-it's-possible".

@kocio-pl , I still think this topic has a high risk of de-railing at the first turn in the track with this particular request...

While I agree the current shop rendering is unsatisfactory, I also think it is unsolvable through any of the current approaches to pre-rendering as part of the style.

Personally, and I said this before in other issue threads, I only see a real solution in a dynamic, Overpass based, rendering of shops. This will allow fancy stuff like personalized filtering of desired shop values, something completely impossible and unthinkable now on the main website.

Technically, this is implementable in a website, as witness-able through e.g. Marc Zoutendijk's "OpenPOIMap" map (http://openpoimap.org/).

However, a "simple" website like OpenPOIMap is not the same as the Rails port that runs the main OpenStreetMap website, API and all the editing infrastructure. That is a different beast AFAIU.

[...]

PR #2589, opened 2017-03-22
Render most shops on z17 as dots only

math1985:
This cleans up the overcrowded z17 significantly:

Before:
screenshot from 2017-03-22 23-42-38

After:
screenshot from 2017-03-22 23-41-18

### DIFF
--- a/amenity-points.mss
+++ b/amenity-points.mss
@@ -565,408 +565,269 @@
     marker-clip: false;
   }
 
-  [feature = 'shop_other'][zoom >= 17] {
-    marker-fill: @shop-icon;
-    marker-width: 6;
-    marker-line-width: 0;
-    marker-placement: interior;
-    marker-clip: false;
-  }
-
-  [feature = 'shop_supermarket'][zoom >= 16] {
-    marker-file: url('symbols/shop/supermarket.svg');
-    marker-placement: interior;
-    marker-clip: false;
-    marker-fill: @shop-icon;
-  }
-
-  [feature = 'shop_art'][zoom >= 17] {
-    marker-file: url('symbols/shop/art.svg');
-    marker-fill: @shop-icon;
-    marker-placement: interior;
-    marker-clip: false;
-  }
-
-  [feature = 'shop_bag'][zoom >= 17] {
-    marker-file: url('symbols/shop/bag.svg');
-    marker-fill: @shop-icon;
-    marker-placement: interior;
-    marker-clip: false;
-  }
-
-  [feature = 'shop_bakery'][zoom >= 17] {
-    marker-file: url('symbols/shop/bakery.svg');
-    marker-fill: @shop-icon;
-    marker-placement: interior;
-    marker-clip: false;
-  }
-
-  [feature = 'shop_beauty'][zoom >= 17] {
-    marker-file: url('symbols/shop/beauty.svg');
-    marker-fill: @shop-icon;
-    marker-placement: interior;
-    marker-clip: false;
-  }
-
-  [feature = 'shop_beverages'][zoom >= 17] {
-    marker-file: url('symbols/shop/beverages.svg');
-    marker-placement: interior;
-    marker-clip: false;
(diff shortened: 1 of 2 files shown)


math1985:
This PR introduces quite a lot of duplicate code, can anybody think of a more elegant way to accomplish this?

kocio-pl:
This would resolve at least part of https://github.com/gravitystorm/openstreetmap-carto/issues/2444.

kocio-pl:
This PR can also help resolve https://github.com/gravitystorm/openstreetmap-carto/issues/1560.

nebulon42:
I'm not sure why I would need to know that there is some sort of shop (= dot without name). Why not leave it out completely and display it later with icon?

amenity-points.mss
```diff
@@ -580,56 +585,91 @@
     marker-fill: @shop-icon;
   }
 
-  [feature = 'shop_art'][zoom >= 17] {
-    marker-file: url('symbols/shop/art.svg');
+  [feature = 'shop_art'][zoom >= 18] {
+    [zoom < 18] {
```
nebulon42: This can never be rendered or am I missing something?

amenity-points.mss
```diff
@@ -580,56 +585,91 @@
...
     marker-fill: @shop-icon;
     marker-placement: interior;
     marker-clip: false;
   }
 
-  [feature = 'shop_bakery'][zoom >= 17] {
-    marker-file: url('symbols/shop/bakery.svg');
+  [feature = 'shop_bakery'][zoom >= 18] {
+    [zoom < 18] {
+      marker-width: 4;
```
nebulon42: I think this misses `marker-line-width: 0` to avoid the darker outlines that can be seen in the screenshot.

amenity-points.mss
```diff
@@ -567,7 +567,12 @@
 
   [feature = 'shop_other'][zoom >= 17] {
     marker-fill: @shop-icon;
-    marker-width: 6;
+    [zoom < 18] {
```
nebulon42: Why not try to place all generic dots here to avoid duplication for each feature? At least the filter has to be duplicated though.

nebulon42:
I was thinking of a rewrite of the POI logic to combine the icons and texts into a shield like for place names. With that we could get rid of names without icon and also try to place the text above the icon if there is not enough space below. I wanted to do a first pass with shields and a second pass with icons only.

Maybe this change here would go well with that rewrite?

kocio-pl:
> I'm not sure why I would need to know that there is some sort of shop (= dot without name). Why not leave it out completely and display it later with icon?

This was the only proposition which had any kind of support and nobody complained when I wanted to implement it: https://github.com/gravitystorm/openstreetmap-carto/issues/2444#issuecomment-262252941.

dieterdreist:
one answer to this question is that the shop density is an interesting datum alone.

math1985:
> I'm not sure why I would need to know that there is some sort of shop (= dot without name).

In areas that are not so dense (like rural areas or suburbs), it might be useful to know where the shops are in the village.

nebulon42:
Ok I understand. It is an improvement to the status quo.

nebulon42:
I tried implementing what I described in https://github.com/gravitystorm/openstreetmap-carto/pull/2589#issuecomment-288720604:
https://gist.github.com/nebulon42/feb4ef4224603bb3bbc62ebadb0adf7e

math1985:
I provided another solution (hope you don't mind, @nebulon42). This solution saves a lot of code, while it should still avoid the combinatorial explosion.

Could somebody review the code?

math1985:
z17 looks now as follows:

screenshot_2017-04-12_00-28-26

math1985:
On closer look, this seems to combine well with @nebulon42's solution, so we can either merge his proposal first and I will adapt, or the other way around.

math1985:
Could somebody have a quick look at the new code?

PR #2590, opened 2017-03-23
tune road shield sizes after font change

### LINKED ISSUE #2478: Highway refs don't properly fit the shield
Using this example (from https://www.openstreetmap.org/#map=18/-21.81209/-48.18708):

See how `SP-310` and `BR-364` are touching the upper and lower shield borders.
Actually, if we zoom, we can see that `BR-364` is surpassing the box:

Isn't it possible to add some pixels on top and on bottom of the shield, to avoid this?
Or maybe reduce the vertical spacing between the values.

nebulon42:
Fixes #2478

preview
after

It looks to me that locally the font size is somehow a bit larger than live. So it would be good if others could test this.


sommerluk:
Looks good to me. Font size differs because the current master branch of openstreetmap-carto is different in font size from at osm.org. Most “ref” values are latin letters or digits. So internationalisation is not so much an issue here. Nevertheless I would test with some test with diacritics and descenders like “ÄÄÄÄÄÄÄÄ ggggggggg” to see how it behaves

nebulon42:
Looks acceptable.

test

sommerluk:
A little test with some example data:

screenshot 1

1. vertical size for one line: “Äg” with only one line is rendered well. “ကြာနီကန်” has not enough vertical space below, but ref=* values like this will be quite unlikely. Overall, I think the vertical size is okay.

2. But when you have two or three lines, the glyphs begin to touch the border of the shield. The more lines you have, the more they touch. I think the additional vertical space that is given for each line of text beyond the first line  should be bigger.

3. The horizontial space is good. Examples like “မႏၲလာ ေဆ” will almost certain not occur in real-world tagging, so no need to adopt to them. (Furthermore, there seems to be a rendering problem with this label, the combining diacritics are not combined. Works well with Noto in MS Word and Indesign.)

nebulon42:
Thanks, I'll do further fine tuning.

sommerluk:
Probably the easiest way is to use an artificial label with the maximum number of lines: “Äg;Äg;Äg;Äg” and one with a single line “Äg”. Than make sure that the space between the letters is and the horizontal lines of the shield is the same for both labels.

nebulon42:
I have fine tuned as suggested. @sommerluk maybe you can have a quick look?

sommerluk:
Looks good to me!

screen01

The spacing below the last line is slightly smaller for the four-lines label than for the one-line label. I think this is okay, also carthograpic. The space is enough to avoid collision on almost all western text.

👍

PR #2591, opened 2017-03-23
replace text-dy logic with combined name and ref rendering for motorway junctions

nebulon42:
before
before

after
after

This also avoids situations where one part of the label can be placed and the other cannot due to some conflict with another object. Leads to simplified code too.

### DIFF
--- a/roads.mss
+++ b/roads.mss
@@ -2394,29 +2394,26 @@ tertiary is rendered from z10 and is not included in osm_planet_roads. */
 #junctions {
   [highway = 'motorway_junction'] {
     [zoom >= 11] {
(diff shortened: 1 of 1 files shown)


math1985:
Nice! Would this be more readable?

``
        ["name" != null]["ref" = null] {
          text-name: "[name]";
        }
        ["name" != null]["ref" != null] {
            text-name: [name] + "\n" + [ref];
        }
``

nebulon42:
> Would this be more readable?

Probably. Updated.

nebulon42:
According to our change of the review policy I'm merging this myself.

PR #2593, opened 2017-03-23
use parameters of road type for text rendering of roads under construction

### LINKED ISSUE #2477: Rendering of name=* in highway=construction
Hi. For some time the name of some streets under construction has been rendered from certain levels of zoom.

I leave a very clear example of something that I think is not correct. The zoom level is 14. It is `highway=construction` + `construction=residential`. As you can see, these streets under construction highlight their name on the map too much more than even the roads that are primarias/secondary.

I think that should not happen with the names. Anyone knows that a primary road is much more important than a residential one which is closed to access due to construction work.
http://www.openstreetmap.org/#map=14/-27.4061/-55.9754
renderingconstr

PS: I'm not against showing the names, but set another priority when showing them. A residential road under construction is not so important to the map; But a motorway under construction is important.

nebulon42:
Fixes #2477

Uses zoom level, size etc. of the respective road types. Also restricts the construction values we render names for to the types where we render highways and empty values of construction.

before
z14
before_z14

z15
before

after
z14
after_z14

z15
after

### DIFF
--- a/project.mml
+++ b/project.mml
@@ -1953,7 +1953,7 @@ Layer:
             way,
             CASE WHEN substr(highway, length(highway)-3, 4) = 'link' THEN substr(highway, 0, length(highway)-4) ELSE highway END,
             CASE WHEN (tunnel = 'yes' OR tunnel = 'building_passage' OR covered = 'yes') THEN 'yes' ELSE 'no' END AS tunnel,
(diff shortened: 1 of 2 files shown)


nebulon42:
I've changed the code to be much simpler.

Issue #488, opened 2014-04-20
Stream tunnel under a swimming pool rendered incorrectly


LoremoCH:
At Rheinau in Switzerland (http://www.openstreetmap.org/#map=18/47.63811/8.59941) two stream tunnels are exactly under a swimming pool. But on the map the where renderet as they would be on top of it.

It is possible to change that?

mkoniecz:
I am not sure is it bad. Rendering tunnels below features that are above them would make tunnels invisible in fully mapped areas.

kocio-pl:
I see no easy fix for this.

pnorman:
> I see no easy fix for this.

My only thoughts are making the casing alternating `@land-colour` and blue dashes.

HolgerJeromin:
Just to provide a screenshot for the future:
![image](https://cloud.githubusercontent.com/assets/2410353/21239624/f9aab5de-c307-11e6-9390-c1882b0ad304.png)

nebulon42:
I think we have to reject the request of the OP. It would be desirable to have the tunnel covered by the pool in this case. But this would be undesirable in a lot of other cases. The vanishing dashes are not ideal and can be improved. @pnorman's idea is the right solution for this.

pnorman:
Fixed in #2594

Issue #996, opened 2014-09-28
Special rendering for intermittent water areas


matkoniecz:
@imagico proposed in #805 to render specially intermittent water areas.

intermittent tag is in the dabase, seasonal needs style file change.

quote from https://github.com/gravitystorm/openstreetmap-carto/issues/805#issuecomment-50901241

This should be extended to water areas, for example the current appearance of Australia is quite irritating:

http://www.openstreetmap.org/#map=6/-24.377/129.749

since nearly all of the water areas you can see there are intermittent.

Both intermittent=yes and seasonal=\* should be considered.

Current use according to taginfo:
- waterway=\* + intermittent=yes: 672000 times (most of this is NHD imports)
- waterway=\* + seasonal=*: 1200 times
- natural=water + intermittent=yes: 45000 times
- natural=water + seasonal=*: <1000 times

Penegal:
The intermittent bodies of water are also often rendered with blue stripes and a dashed outline, at least in France. I don't know if it could be a better rendering, but it's a possibility.

liotier:
Seasonality is a specific sort of intermittency - variation is according to seasons instead of more randomly weather-related. I came here to ask for waterways with seasonal=yes to be rendered the same way as intermittent=yes

For now, seasonal waterways appear as solid lines and it is a bit disconcerting.

Treating seasonal=yes the same way as intermittent=yes feels quite correct and it would also minimize the work necessary to take this extra tag into account.

This issue covers water bodies... Should I file a new issue to cover waterways ?

dieterdreist:
sent from a phone

> Am 24.03.2016 um 13:19 schrieb Jean-Marc Liotier <notifications@github.com>:
> 
> Treating seasonal=yes the same way as intermittent=yes feels quite correct and it would also minimize the work necessary to take this extra tag into account.


I agree and generally support this, even if it might have to wait till we add hstore (if ever)

kocio-pl:
> Both intermittent=yes and seasonal=* should be considered.

Because we already have hstore and are able to use non-standard tags too, we can do it quite easily now. I just wanted to ask if this is still the same (just adding seasonal=*, without "no" probably, the same as we use intermittent=yes) as I understand original proposition or something has changed in the meanwhile?

Currently seasonal has 108 828 uses together with waterway=* and 16 696 with natural=water, which makes it very popular tag. It's also growing fast in the last years:

[...]

Issue #405, opened 2014-03-13
river tunnel under a river is shown split


HolgerJeromin:
http://www.openstreetmap.org/way/75255454 (river tunnel) is below (layer=-1) 
http://www.openstreetmap.org/way/3567099 (river)

as the riverbed is mapped the "crossing" is rendered strange:
http://www.openstreetmap.org/#map=19/49.01623/8.30446

matkoniecz:
![selection_004](https://cloud.githubusercontent.com/assets/899988/16228291/7bee40a2-37b6-11e6-80fa-25a36417901a.png)
image from the linked example

imagico:
I think this is solved since we stopped rendering waterway lines over water areas.

PR #2594, opened 2017-03-24
use land-color as background for waterway tunnels

nebulon42:
Ref #488.

Waterway tunnel dashes might disappear depending on background. By using an additional line with land-color as base this does not happen anymore.

Original location of #488:
before
pool_before

after
pool_after

This would also apply to intermittent waterways:
before
before_intermittent_river

after
after_intermittent_river

This can have side-effects when a intermittent river is tagged with a riverbank. Depends if we view that as valid tagging combination.
before
before_intermittent_with_riverbank

after
after_intermittent_riverbank

Also stream tunnels over forests look a bit strange but there is likely some fine-tuning possible:
stream_forest_strange

### DIFF
--- a/water.mss
+++ b/water.mss
@@ -96,6 +96,18 @@
   [waterway = 'canal'][zoom >= 12],
   [waterway = 'river'][zoom >= 12],
   [waterway = 'wadi'][zoom >= 13] {
+    // the additional line of land color is used to provide a background for dashed casings
+    [int_tunnel = 'yes'] {
+      background/line-color: @land-color;
+      background/line-width: 2;
+      background/line-cap: round;
+      background/line-join: round;
+    }
+    water/line-color: @water-color;
+    water/line-width: 2;
+    water/line-cap: round;
+    water/line-join: round;
+
     [bridge = 'yes'] {
       [zoom >= 14] {
         bridgecasing/line-color: black;
@@ -106,6 +118,7 @@
         [zoom >= 18] { bridgecasing/line-width: 13; }
       }
     }
+
     [intermittent = 'yes'],
     [waterway = 'wadi'] {
       [bridge = 'yes'][zoom >= 14] {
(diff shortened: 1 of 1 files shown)


imagico:
I think for intermittent waterways this goes into the wrong direction.

Both tunnels and intermittent waterways are something less than a normal waterway in terms of visibility and importance for the typical map user.  Therefore they should IMO be rendered less prominent than normal waterways.  Using land color between the dashes will make them more prominent on many backgrounds.  The difference is likely fairly small for tunnels so i consider the change acceptable there although both the current and the new proposed styling are not ideal in terms of relative prominence at the moment.

I am generally not really satisfied with the dashed rendering of intermittent waterways we have at the moment, especially at the low zoom levels.  And at the high zooms we should probably revisit that when we implement intermittent water areas - see #996.

Note if we decide to change the water color in a way similar to what i proposed in https://github.com/gravitystorm/openstreetmap-carto/issues/1781#issuecomment-241061156 the relative weight of surface waterways vs. tunnels would change (to the better).

The side effect you showed would be solved by the layer reordering i have envisaged for the move to water polygons and differentiated color water rendering (i.e. rendering waterways before water areas).

water.mss
```diff
@@ -103,6 +103,20 @@
   [waterway = 'canal'][zoom >= 12],
   [waterway = 'river'][zoom >= 12],
   [waterway = 'wadi'][zoom >= 13] {
+    // the additional line of land color is used to provide	 a background for dashed appearances
```
math1985: tab in the middle of the line?

water.mss
```diff
@@ -103,6 +103,20 @@
...
+    [intermittent = 'yes'],
+    [waterway = 'wadi'],
+    [int_tunnel = 'yes'] {
+      line-color: @land-color;
```
math1985: Maybe for clarity good to prefix these with land/ or background/ or so?

nebulon42:
I have adressed the points raised. For wadis the same as for intermittent waterways applies, therefore I have both removed from this change.

Wadis would have looked like this (likely not a wadi but tagged as one):
wadi

water.mss
```diff
@@ -123,37 +136,53 @@
...
-      [zoom >= 17] { a/line-width: 7; }
-      [zoom >= 18] { a/line-width: 8; }
+      [zoom >= 13] { background/line-width: 3; }
+      [zoom >= 14] { background/line-width: 5; }
+      [zoom >= 15] { background/line-width: 6; }
+      [zoom >= 17] { background/line-width: 10; }
+      [zoom >= 18] { background/line-width: 12; }
+    }
+
+    [int_tunnel = 'yes'] {
```
math1985: This selector is the same as the selector above

water.mss
```diff
@@ -113,6 +125,7 @@
         [zoom >= 18] { bridgecasing/line-width: 13; }
       }
     }
+
```
math1985: Spurious whitespace

water.mss
```diff
@@ -166,24 +195,35 @@
           [waterway = 'stream'][zoom >= 15] { bridgeglow/line-width: 3; }
         }
       }
+
```
math1985: Spurious whitespace in diff

pnorman:
Ping @nebulon42 about the technical comments above.

nebulon42:
@math1985 I have merged the redundant filters, thanks for noticing. The additional newlines were intentional. I think they improve readability.

mmd-osm:
A user reported some side effects of this change: https://www.pic-upload.de/view-33202010/Rendering-Fehler_Baeche_Bsp.png.html

https://www.openstreetmap.org/#map=13/34.2950/70.9525

Was it really intended to be rendered this way?

HolgerJeromin:
Does this solve / change  #405 too?

nebulon42:
No, because the river line is still rendered above the tunnel.

Issue #1640, opened 2015-07-07
Reduce distance of generic dot / icons to shop-label


daganzdaanda:
This is related to #957 
Is there any way to reduce the default dy between the dot and the labels? 
In some places, the dot is further from its own label than from the bottom of the next shop: http://www.openstreetmap.org/node/2933990260
The same happens with icons, but they are bigger, so it does not show up so often.

I know that the dy right now has to take into account the "highest" letters possible, e.g. É, but for most Latin names, it's complete overkill...

Without knowing the technical details, here are my ideas:

1) Align the label using its top, not the baseline 
(probably wrongly assuming that the baseline is currently used)

2) Align the label using its rendered height 
(probably not possible, because it would have been done already) 

3) Move the dot from a centered position to the bottom. 
Maybe all the icons could be bottom aligned for this map. 

4) Switch to showing the label to the right of the dot / icon. 
All letters probably need only a small dx. Pretty radical change in looks, though.

5) Reduce dy a bit and allow and accept a few overlapping labels. 

6) Keep dy but increase the minimum distance between labels.
Will lead to less labels.

Maybe a combination of 3, 5 and 6 is possible?

kocio-pl:
I am not sure if it will be better, because there should be many places with both generic and specific icons, so it will loose some uniformity we have now. I'm not all against this proposition, we just need to check it. #957 is also kind of similar regarding this problem.

kocio-pl:
I don't like this idea, so I will close it now. It can be reopened if someone disagrees.

jeisenbe:
We could do this by using a different text-dy on different zoom levels; on the zoom level where an object is rendered as a dot we could use text-dy 4 or 6, versus text-dy 10 when there is a real icon.

I agree that this would look better, though it would be some work to get right.

Adamant36:
It probably add a bunch of extra code for very little gain to. There should be some standard for how things are rendered anyway. There shouldn't vast differences in how any single object is render at different zoom levels. Otherwise, you lose cartographic consistency. Especially if the differences aren't standardized to specific zoom levels, like road rendering transitions are.

Issue #2248, opened 2016-07-26
Using vertical alignment with marker labels 


HolgerJeromin:
I just remembered the problem in the past with too small dy for names under icons. Especially if the font is quite big. 
Did you considered using vertical-alignment=top so a bigger name is not blocked by the icon. 
Default is bottom when used with dy >0 according to the documentation

math1985:
@matkoniecz Do you have time to test this?

daganzdaanda:
Please try this! If it works, it would fix #1640 which I find really annoying ;)

nebulon42:
Carto documentation states that the default value for text-dy > 0 is top anyway.

> Default Value: auto (Default affected by value of dy; "top" for dy>0, "bottom" for dy<0.)

HolgerJeromin:
Perhaps i am wrong, but 
https://carto.com/docs/carto-engine/cartocss/properties#text-vertical-alignment-keyword
> Note: The default value is affected by the dy (text down) value. If the dy value>0, the vertical alignment value is bottom. If the dy value<0, the vertical alignment is top.

And we have positive [dy](https://github.com/gravitystorm/openstreetmap-carto/blob/master/amenity-points.mss#L1223) for amenity points.

math1985:
I tried this harder, but my experience was that this made the map significantly harder to read due to the fact that it was no longer clear which label an icon belonged to in crowded environments.

HolgerJeromin:
@math1985 
The map was harder to read if you reduce the distance between the icon and text (because you need no buffer for very large glyphs)?

math1985:
In any case, the buffer between a label and another icon need to be significantly larger (like 2-3 times?) than between a label and the icon it belongs to.

nebulon42:
@HolgerJeromin Interesting find. I don't know where Carto derives their documentation from. They also use an own fork of CartoCSS. Maybe the confusion stems from the fact that Carto was before called CartoDB and should not be confused with Mapbox' Carto(CSS).

In any case the carto we all use uses [mapnik-reference](https://github.com/mapnik/mapnik-reference) for the documentation and this [commit](https://github.com/mapnik/mapnik-reference/commit/3c315b181f2e49cdcc9a0dd876884c1efecf2fa7) shows that the description was apparently wrong before and was corrected.

HolgerJeromin:
@nebulon42 Thanks for this. 
[...]

PR #2597, opened 2017-03-25
major rewrite of amenity-points

### LINKED ISSUE #234: Tie POI names to icons
If a POI icon conflicts with another label but not its text right now the text will be rendered without any icon. Examples have shown up in #232 and maybe #201 but I've seen it locally.

A simple example is

The restaurant text conflicts with the icon. Also, the text is not centered on the POI because it's offset to below to allow for an icon.

A more complex example is 

[AFG Kabul Bakery](http://www.openstreetmap.org/browse/node/960913456) (`shop=bakery`) has its label rendered, but the icon conflicts with the icon for [Zahra's Ultimate Hair Design](http://www.openstreetmap.org/browse/node/960913382) (`shop = hairdresser`). With the name it's fairly obvious what AFG Kabul Bakery is, but there's no icon to it. 

Oddly, it looks like the text for Zahra's Ultimate Hair Design conflicts with the AFG Kabul Bakery text.

It would be an improvement if the POI name was only rendered in conjunction with the icon, and if there wasn't room for the icon it didn't render either.

This will probably not impact rendered POI density significantly because in dense areas where this is an issue the blank space from removing the name will probably be taken up by another icon.

nebulon42:
Fixes #234, #1640.

Improvements
* more compact code, more DRY
* labels are tied to icons with shields, so label without icon is impossible, a second pass of icon only fills up remaining space
* harmonization to dy of 11 with a few exceptions (omitting text due to collision with icon cannot happen with shields)
* landcover labelling code moved to landcover.mss
* island, islet labelling code moved to placenames.mss
* icon and label in one query possible, less complex label query
* fixes most visible fountain rendering problems (see #1934) by using a SVG file instead of two overlapping markers on z17
* features that use a POI icon but landcover labelling now use a text configuration similar to other POIs when the icon is shown, before that they use the landcover labelling

Caveats
* Shields do not support SVG re-coloring. This requires us to have the SVGs in a specific color. But I think this is not a big deal, because re-coloring without shields is still possible and downstream customization is also still possible and not so difficult when using `sed` or similar


math1985:
I think this goes in the right direction. Let me know when it is ready for review.

HolgerJeromin:
You could try my idea of using vertical-alignment=top (#2248)

nebulon42:
@HolgerJeromin I'm not sure if this would still be needed with shields as the position of the image does not affect the position of the text, so the icon cannot block the text.

HolgerJeromin:
Yes, an icon cannot block the text, but we can have different distances between the icon bottom and text top. Large text can result in smaller distance than small text heights.

BTW [shield-vertical-alignment-keyword](https://carto.com/docs/carto-engine/cartocss/properties#shield-vertical-alignment-keyword) differs here from the normal align that it defaults to `middle` with no influence of dy.

nebulon42:
Yes of course. It's a bit hidden because there are so many changes, but the reduction of dy for the generic shop is already there: https://github.com/nebulon42/openstreetmap-carto/blob/27bc13cc8f51f42c04abefa48ce08294725fe7f7/amenity-points.mss#L782.

nebulon42:
@math1985 This is now ready for review.
Actually it would benefit from as much review by different persons as it is able to get.

I have not touched the low-priority layer. And I have a few questions regarding it:
* I don't really understand why parking related features are part of the low priority layer as they are part of the regular layers anyway. Was this meant as space-filling with icons? Then it can be removed now.
* Are any features in this layer supposed to have text? If yes, there need to be changes to account for that.

nebulon42:
> Would this rewrite also be possible without coloring the SVGs directly?

No, unfortunately not. The ShieldSymbolizer does apparently not support SVG re-colouring otherwise I would have used it. It is not ideal, but as I wrote above customization is still possible if no shields are used and otherwise by altering the colours in the SVGs directly. This can also be automated by using sed or any other search and replace utility.

For me the improvements of the shield approach outweigh the downside of having the colours in the SVGs.

pnorman:
> For me the improvements of the shield approach outweigh the downside of having the colours in the SVGs.

👍

project.mml
```diff
@@ -1646,6 +1646,8 @@ Layer:
                                'laundry', 'dry_cleaning', 'beverages', 'perfumery', 'cosmetics', 'variety_store', 'wine', 'outdoor', 
                                'copyshop', 'sports', 'deli', 'tobacco', 'art', 'tea', 'coffee') THEN shop
                                ELSE 'other' END AS shop,
+            name,
```
math1985: The diff of the changes in this file is a bit hard to read, what do we change exactly in this file?

Is it still necessary to have a separate text layer in addition to the amenity-points layer?

amenity-points.mss
```diff
@@ -4,1894 +4,1243 @@
...
-    point-placement: interior;
-  }
+  // markers with text (shields)
 
-  [feature = 'tourism_wilderness_hut'][zoom >= 13],
-  [feature = 'amenity_shelter'][zoom >= 16] {
-    marker-file: url('symbols/shelter.svg');
+  [feature = 'tourism_alpine_hut'][zoom >= 14],
+  [feature = 'tourism_wilderness_hut'][zoom >= 14],
+  [feature = 'amenity_shelter'][zoom >= 17],
```
nebulon42: It looks like if I did, but for z13 (alpine hut, wilderness hut) and z16 (shelter) the markers do not have text, so they are in the "markers without text" category.

amenity-points.mss
```diff
@@ -4,1894 +4,1243 @@
...
+    [feature = 'amenity_cinema'] {
+      shield-file: url('symbols/cinema.svg');
+      a/marker-file: url('symbols/cinema.svg');
+    }
+    [feature = 'highway_bus_stop'] {
+      shield-file: url('symbols/bus_stop.12.svg');
+      shield-fill: @transportation-text;
+      shield-text-dy: 9;
+      a/marker-file: url('symbols/bus_stop.12.svg');
+      a/marker-fill: @transportation-icon;
```
nebulon42: Strictly speaking we don't but I wanted to keep those as we did have them before. But when I think about it it is redundant and probably should be removed. Should I?

math1985:
Another strange issue:
screenshot from 2017-05-07 20-06-59

amenity-points.mss
```diff
@@ -4,1894 +4,1243 @@
...
+    [feature = 'amenity_townhall'] {
+      shield-file: url('symbols/town_hall.svg');
+      a/marker-file: url('symbols/town_hall.svg');
+    }
+    [feature = 'tourism_museum'] {
+      shield-file: url('symbols/museum.svg');
+      a/marker-file: url('symbols/museum.svg');
+    }
+    [feature = 'man_made_mast'] {
+      shield-file: url('symbols/communications.svg');
```
nebulon42: It should.

amenity-points.mss
```diff
@@ -4,1894 +4,1243 @@
...
+      shield-fill: @water-text;
+      shield-text-dy: 8;
+      a/marker-file: url('symbols/spring.svg');
+    }
+    [feature = 'man_made_windmill'] {
+      shield-file: url('symbols/windmill.svg');
+      a/marker-file: url('symbols/windmill.svg');
+    }
+    [feature = 'amenity_hunting_stand'] {
+      shield-file: url('symbols/hunting_stand.svg');
```
nebulon42: Yes.

amenity-points.mss
```diff
@@ -4,1894 +4,1243 @@
...
-    marker-placement: interior;
-    marker-clip: false;
+      [access != ''][access != 'permissive'][access != 'yes'] {
+        shield-opacity: 0.33;
+        shield-text-opacity: 0.33;
+        a/marker-opacity: 0.33;
+      }
+    }
   }
+  [feature  ='amenity_fountain'][zoom >= 17] {
```
nebulon42: The problem was the layering of nozzle and pool. It required `marker-allow-overlap` to display both layered above each other. The downside to this approach is that it also allows other markers to overlap with them which happened frequently. (see #1934)

project.mml
nebulon42: I added the name attribute (and other necessary ones for text rendering) to the amenity-points queries. At the same time I removed the now unnecessary parts of the text layer queries. Those are still needed, because there are other text features (e.g. water related) that also use the queries.

amenity-points.mss
```diff
@@ -4,1894 +4,1243 @@
...
+        a/marker-file: url('symbols/bicycle_parking.svg');
+      }
+      [feature = 'amenity_motorcycle_parking'] {
+        shield-file: url('symbols/motorcycle_parking.svg');
+        a/marker-file: url('symbols/motorcycle_parking.svg');
+      }
+    }
+  }
+
+  // markers without text
```
nebulon42: I'm not sure how a shield without text behaves and we would still have the text-dy which probably takes up space even if name is empty. I'm also not sure if a shield symbolizer uses more resources so it might be not good to use it unnecessarily.

But I'm open to other ideas.

Another option would be to solve this cartographically by not having POI icons where their label shows up at later zoom level. But I would be against that.

nebulon42:
I have adressed the points raised:
* checked color and added shield-fill
* ice cream fix
* whitespace fix
* airport font fix

> Lowzoom airport icons are colored wrong

I did not see this. Could you show an example?

Thanks both for the thorough review.

> If I understand correctly, labels will now be given priority over icons. I'm not sure if that's an improvement. In Luxembourg, I have seen cases where, on z17, a single bank label is blocking four restaurant icons.

To be a bit more specific labelled POI or those without rendered label now come first and then the remaining space is filled with icons where possible. But the rules are now more straightforward, before it was rather arbitrary.

z17 is still quite crowded in cities even after your shop rendering change. Another thing I could think of would be to penalize long names somehow so that too long names do not block too many icons. But this was just an idea, I would have to think about how to do that.

imagico:
I have not had time to test this but am i right to assume that this change does not solve the long time problem of symbols frequently vanishing and later re-appearing as you zoom in because symbol priorities are not strictly in order of the starting zoom level?

If this is not the case is this change at least neutral in that regard or does it make this problem worse?

nebulon42:
It definitely does not solve the problem and I would expect that this change is neutral towards this.
Am I right that this happens mostly at metatile boundaries?

imagico:
No, this has nothing to do with metatiles.  

The problem occurs if at a certain zoom level new symbols start to be shown - and are rendered with higher priority than symbols that already show at a lower zoom level.

nebulon42:
Yes, so I thought sorting the rules by lowest zoom level would help but this was not the case for a instance of this problem. The only places where I found vanishing symbols in my test rendering were metatile boundaries. But since my Kosmtik rendering suffers from cut-off things at these boundaries anyway this might be something else. I thought if the buffer overlap is not sufficient collisions there are more likely.

pnorman:
Vanishing symbols with zoom variations should happen when either there's a change in label collisions interactions, or you have two things which collide and on the one rendered at lower zooms appears later in the query results at higher zooms.

Because I don't think we order the amenity queries by tagging at all, this PR is probably no change to that.

pnorman:

math1985:
@nebulon42 Can you give some example before/after renderings of a densely mapped city in your area? It might help us to determine whether the disappearance of the label-without-icon-problem offsets the lesser amount of shops shown.

nebulon42:
Previews of Vienna (inner city).

before

z17
z17_before

z18
z18_before_1

z18_before_2

after

z17
z17_after

z18
z18_after_1

z18_after_2

nebulon42:
Since we now require carto 0.18 I have updated this to use a feature of carto >= 0.17.2:
https://github.com/gravitystorm/openstreetmap-carto/pull/2597/files#diff-5d9237083037cf1d6d91c329a0ff61adR811

nebulon42:
@pnorman Yes, I know. But it can be manually backported without using this feature with minimal effort. I'm willing to do this once merged to master.

math1985:
Thanks for the previews!

Unfortunately, in the new rendering quite a lot of information is lost by icons that are blocked by labels and cannot be rendered. This plays especially on z18 (mainly in the second example) but also in z17 (for example the Jewish museum).

The main problem this is solving, labels without icons, seems quite rare. If I'm not mistaken, we have no single instance of this happening in the previews.

I therefore think this PR is not an improvement - even though it's very cool from a technical perspective.

nebulon42:
There are definitely changes in labelling and some labels block other icons. But I would not use "a lot of information".

One way to tackle this could be to introduce a length limit for labels. When this is exceeded the process is reversed and first icons without label are placed and after that icons with label. But probably this gets to complex and is not worth it.

It remains to be decided if the gain in consistency and reduction of complexity is worth those changes in rendering.

amenity-points.mss
```diff
@@ -4,1894 +4,1243 @@
...
+    shield-size: @standard-text-size;
+    shield-face-name: @standard-font;
+    shield-halo-radius: @standard-halo-radius;
+    shield-halo-fill: @standard-halo-fill;
+    shield-wrap-width: @standard-wrap-width;
+    shield-fill: @amenity-brown;
+    shield-text-dy: 11;
+    shield-unlock-image: true;
+
+    // second pass for icons without text where there is still space
```
pnorman: 👍

nebulon42:
Before I'm going through the hassle of rebasing and resolving the conflicts I'd like to sort out the question if it is worth to go on or not. @math1985 I respect your opinion, but I'd like to hear from other maintainers what they think about the problem with icons blocked by labels. Apparently @pnorman used the technique in another project so I'm curious what he thinks about that.

pnorman:
> Apparently @pnorman used the technique in another project so I'm curious what he thinks about that.

It works well, and it's the best technique I know of. The limitations I found were

- duplication with needing to specify a shield-file and marker-file that are the same and
- having to adjust the colour in the SVGs, meaning I couldn't use icons from our symbol source without first modifying them, and any colour changes require changing a bunch of SVGs.

The only overall improvement I'd make is to use a name like `fallback` instead of `a` for the marker symbolizer, since it makes it clearer what it's for.

I'd be okay merging it, except I'd have to resolve the conflicts, and handle the backporting to 3.x.

math1985:
@pnoeman You're aware of the drawbacks I pointed out right?

math1985:
@pnorman You're aware of the drawbacks I pointed out right?

pnorman:
> @pnorman You're aware of the drawbacks I pointed out right?

> Unfortunately, in the new rendering quite a lot of information is lost by icons that are blocked by labels and cannot be rendered. This plays especially on z18 (mainly in the second example) but also in z17 (for example the Jewish museum).

I'm just not seeing this when I look at the previews. I see different stuff being rendered, but no overwhelming trends. Perhaps there's some adjustments we can make to the margins to allow labels to be closer?

nebulon42:
I'm tired of this.

math1985:
What are you tired of exactly? If it's about my comments, you have my full blessing to go ahead with this if that's what you think is best for the project.

math1985:
Also, is there anything we can help you with?

nebulon42:
:) It's just that I would have had to resolve the merge conflicts and I saw what has changed and it was just a mess. So I decided not to do it, I do not have the energy to keep this up to date with master.
Better to let it be. I'm not tired of persons, just to be clear.

Issue #2436, opened 2016-11-13
Project governing status


kocio-pl:
I'm a bit worried about governing status of this project.

This is not only current problem, however long and stalled PR queue is the most visible effect. This alone would not bother me if we have ongoing discussion - it's perfectly clear that some code needs more insight and debate, and I consider the time to do this [as the most important part of review process](https://github.com/gravitystorm/openstreetmap-carto/issues/2284#issuecomment-240072406). But in many cases we have passed the point of oblivion - discussion is no more active and it's getting hard to recollect the arguments. Moreover decisions are not being made and the code is not merged, nor even dismissed. 

We kind of stopped moving, which is bad for any project if it's not close to be perfect. And we're not there yet - for me there are a lot things to change and more people joining us also tells me we have nice potential here. With osm-carto community getting stronger the problem of project governance model is getting more and more evident for me.

We have 4 collaborators made equal by the (repo) creator. Yet @gravitystorm is in practice not active here for a long time, rarely even giving some advice or expressing his thoughts on general issues. @matkoniecz is not committing anything for about half of year now, which is also quite long. @pnorman is recently more active, but he tends to follow purely technical issues (like code quality, performance or general database framework). Which leaves us with just @math1985 as the only active collaborator trying to cover all the fields, like communication, decision making and merging code. With his activity drop for 6 weeks it's no wonder this project is stalled.

I'd like to be perfectly clear - there's nothing wrong with people activity shift (or even going away) in open projects. The real problem is that our governing structure does not reflect the reality. Even if Matthijs will be back with the same superpowers, it's still one man the project relies on for day-to-day operations. In my opinion this is not healthy situation.

What are your thoughts on this topic?

imagico:
Oh come on.  There are only 16 open PRs right now and your oldest one is barely 2 months old.

None of the maintainers is paid for the work on this style so you need to accept that available free time varies for them.

Don't get me wrong - as i have said before it would be great if decisions on changes were faster, in particular when the result is ultimately a rejection.  And i think this could probably be improved without additional work load for the maintainers - by doing early reviews on suitability/prospects of the changes.  But the idea that this is currently a particular issue and development is stalled by maintainer inactivity is fairly frivolous.

[...]

PR #2601, opened 2017-03-28
Do not render link roads until z12

math1985:
This improves rendering on mid zoom levels by smoothing up the highways.

Before:
screenshot_2017-03-28_21-44-40

After:
screenshot_2017-03-28_21-45-12

This becomes especially important after implementing something like https://github.com/gravitystorm/openstreetmap-carto/issues/2194#issuecomment-265266841:

Before: 
screenshot_2017-03-28_21-52-06

After:
screenshot_2017-03-28_21-47-36

### DIFF
--- a/roads.mss
+++ b/roads.mss
@@ -1164,191 +1164,196 @@ tertiary is rendered from z10 and is not included in osm_planet_roads. */
     }
 
     [feature = 'highway_motorway'] {
-      [zoom >= 5] {
+      [zoom >= 5][link != 'yes'],
+      [zoom >= 11] {
         line-color: @motorway-low-zoom;
         line-width: @motorway-width-z5;
-      }
-      [zoom >= 7] { line-width: @motorway-width-z7; }
-      [zoom >= 8] { line-width: @motorway-width-z8; }
-      [zoom >= 9] { line-width: @motorway-width-z9; }
-      [zoom >= 10] { line-width: @motorway-width-z10; }
-      [zoom >= 11] { line-width: @motorway-width-z11; }
-      [zoom >= 12] {
-        line-color: @motorway-fill;
-        line-width: @motorway-width-z12 - 2 * @major-casing-width-z12;
-        [zoom >= 13] { line-width: @motorway-width-z13 - 2 * @major-casing-width-z13; }
-        [zoom >= 15] { line-width: @motorway-width-z15 - 2 * @major-casing-width-z15; }
-        [zoom >= 17] { line-width: @motorway-width-z17 - 2 * @major-casing-width-z17; }
-        [zoom >= 18] { line-width: @motorway-width-z18 - 2 * @major-casing-width-z18; }
-        [zoom >= 19] { line-width: @motorway-width-z19 - 2 * @major-casing-width-z19; }
-        [link = 'yes'] {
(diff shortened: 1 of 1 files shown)


imagico:
I suppose you mean z<12.

Could this result in gaps in cases like here:

http://www.openstreetmap.org/#map=15/42.1637/12.6129

math1985:
> I suppose you mean z<12.

Yep.

> Could this result in gaps in cases like here:

Likely yes. Unfortunately Luxembourg doesn't seem to have a case like this.

nebulon42:
It reduces clutter which is great but on the other hand it leads to gaps at junctions and there are also some longer link passages that could have some relevance on < z12.

Compare https://www.openstreetmap.org/#map=11/48.0510/16.4991 of Vienna's south east with
after_marked
(I have not checked if the tagging is accurate.)

So I'm leaning a bit to the negative side due to the side effects, but otherwise it would be a good improvement.

HolgerJeromin:
Even in Luxembourg we have a problem. Last image from math.
image

Rendering links like the connecting highway (yellow?) is probably not a good idea, too.

math1985:
Thanks for the examples!

pnorman:
Lately I've gone for starting rendering of links at z10 in my projects. I've found if I increase that it leads too too many problems of gaps.

math1985:
Google Maps has the same problem, by the way:
screen shot 2017-04-04 at 11 06 40

imagico:
We probably need to consider

1. what our expectations of quality are here.
2. what side effects rendering artefacts will have regarding mapping practice.  I suppose if we have visible gaps at larger highway crossings with long links people will be tempted to at least partly tag the links as normal highways.

You might also consider some fancy logic only removing topologically redundant links but that is probably too expensive for us.

HolgerJeromin:
> Rendering links like the connecting highway (yellow?) is probably not a good idea, too.

After thinking about it more this could probably work. Calculating the connecting road class is probably not doable, but render all motorway links with a primary color could be nice.

pnorman:
Perhaps instead of z12 (proposed) or z9 (current) we could go with something between them?

pnorman:

math1985:
Thanks for all comments. I have changed the PR to only drop link roads up to z11 (rather than z12). Merged.

nebulon42:
Hm, I have the impression that recently it doesn't matter any more so much if others have concerns about changes. They get merged anyway. I think that decreases quality.

nebulon42:
Also: Didn't we have the requirement that there has to be at least one approving review?

math1985:
You're right, I reverted this merge. As it is not possible to re-open a PR, I created a new PR here: https://github.com/gravitystorm/openstreetmap-carto/pull/2758 Could some of you please review it?

kocio-pl:
I have [warned](https://github.com/gravitystorm/openstreetmap-carto/issues/2436#issuecomment-318955397) about problem with reviews lately. Lack of discussion seems to be a bigger factor regarding quality than making informed decision.

math1985:
Sorry, I'm not sure if I understand what you mean with you last commebt?

kocio-pl:
It was response to both @nebulon42 comments.

PR #2602, opened 2017-03-28
Add halo to roads on z8-11

math1985:
This PR adds a halo to roads on z8-11.

Before/after:
screenshot_2017-03-29_00-58-04 screenshot_2017-03-29_00-58-24

Before/after:
screenshot_2017-03-29_01-09-12 screenshot_2017-03-29_00-59-50

Before/after:
screenshot_2017-03-29_01-00-35 screenshot_2017-03-29_01-05-35

### DIFF
--- a/project.mml
+++ b/project.mml
@@ -590,7 +590,7 @@ Layer:
         ) AS tunnels
     properties:
       group-by: layernotnull
-      minzoom: 9
+      minzoom: 10
   - id: landuse-overlay
     name: landuse-overlay
     geometry: polygon
@@ -862,7 +862,7 @@ Layer:
             CASE WHEN int_surface IN ('unpaved') THEN 0 ELSE 2 END
         ) AS roads_casing
     properties:
-      minzoom: 9
+      minzoom: 10
   - id: highway-area-fill
     name: highway-area-fill
     # FIXME: No geometry?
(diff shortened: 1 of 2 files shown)


math1985:
Changed title to 'Add halo to roads on z8-11', something went wrong with the original title.

math1985:
This depends on #2602.

math1985:
This is a rather subtle change, but this should increase contrast between the roads (in particular primary) and landuse.

math1985:
Known issue: on z9 there is a problem with gaps in the roads, likely caused by interaction with the low-zoom layer.

imagico:
The gaps might be bridge related.

Regarding the change itself - i can see the better readability of large roads but i also see more noise in the map overall, esp. at z8/z9 and the halo takes up space and covers other stuff which is less visible, in particular rivers.

I already mentioned this when @matkoniecz introduced halos for the minor roads.  This does not necessarily mean i am against this change but i clearly see that this will not improve readability overall as much as you might think if you only look at the roads.

nebulon42:
I'm not quite sure about this. Might likely be an improvement but I would prefer it to be more subtle. Maybe less thick or more transparent? Right now I think it is a bit too noisy especially with motorways.

pnorman:
> This depends on #2602.

Do you mean a different issue, as **this** is #2602

math1985:
> Do you mean a different issue, as this is #2602

I meant #2601, I can however change the PR so that it no longer depends on it.

pnorman:

math1985:
This has been implemented as part of the midzoom changes.

PR #2607, opened 2017-04-09
Documentation update (font usage and download)

### LINKED ISSUE #2606: INSTALL.md: add link for installing Noto fonts on macOS
I tried installing OpenStreetMap Carto on macOS 10.11 and viewing using Kosmtik, but having only installed Noto Sans Regular, Italic, Bold and Bold Italic and not all the other international variants I received a large warning message in Kosmtik (`the font ... is not a known font on your system, please provide an existing font face name`) that obliterated my view of the map itself.

It might be worth adding into INSTALL.md that the easiest way for non-Ubuntu users to obtain all the Noto fonts is by visiting https://www.google.com/get/noto/.

Thanks.

sommerluk:
Documentation update (font usage and download)

Resolves #2606

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -43,11 +43,19 @@ The repeated www.naturalearthdata.com in the Natural Earth shapefiles is correct
 Put these shapefiles at `path/to/openstreetmap-carto/data`.
 
 ## Fonts
-The stylesheet uses Noto Sans, an openly licensed font from Google with support for multiple scripts. The "UI" version is used where available, with its vertical metrics which fit better with Latin text. Other fonts from the Noto family are used for some other languages.
+The stylesheet uses Noto, an openly licensed font from Google with support for multiple scripts. The “UI” version is used where available, with its vertical metrics which fit better with Latin text. The stylesheet uses Noto’s “Sans” style where available. If not available, this stylesheet uses another appropriate style of the Noto family.
 
 DejaVu Sans is used as an optional fallback font for systems without Noto Sans. If all the Noto fonts are installed, it should never be used.
 
-Unifont is used as a last resort fallback, with it's excellent coverage, common presence on machines, and ugly look.
+Unifont is used as a last resort fallback, with it’s excellent coverage, common presence on machines, and ugly look.
+
+### Download
+
+Noto: http://www.google.com/get/noto/ and http://github.com/googlei18n?utf8=%E2%9C%93&q=noto
+DejaVu: http://dejavu-fonts.org/
+Unifont: http://unifoundry.com/
+
+### Download on Ubuntu/Debian
 
 On Ubuntu 16.04 or Debian Testing you can install the required fonts except Noto Emoji Regular with
 
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -43,21 +43,15 @@ The repeated www.naturalearthdata.com in the Natural Earth shapefiles is correct
 Put these shapefiles at `path/to/openstreetmap-carto/data`.
 
 ## Fonts
-The stylesheet uses Noto, an openly licensed font from Google with support for multiple scripts. The “UI” version is used where available, with its vertical metrics which fit better with Latin text. The stylesheet uses Noto’s “Sans” style where available. If not available, this stylesheet uses another appropriate style of the Noto family.
+The stylesheet uses Noto, an openly licensed font family from Google with support for multiple scripts. The stylesheet uses Noto's "Sans" style where available. If not available, this stylesheet uses another appropriate style of the Noto family. The "UI" version is used where available, with its vertical metrics which fit better with Latin text.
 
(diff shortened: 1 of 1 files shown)


INSTALL.md
```diff
@@ -43,11 +43,19 @@ The repeated www.naturalearthdata.com in the Natural Earth shapefiles is correct
...
-Unifont is used as a last resort fallback, with it's excellent coverage, common presence on machines, and ugly look.
+Unifont is used as a last resort fallback, with it’s excellent coverage, common presence on machines, and ugly look.
+
+### Download
+
+Noto: http://www.google.com/get/noto/ and http://github.com/googlei18n?utf8=%E2%9C%93&q=noto
+DejaVu: http://dejavu-fonts.org/
+Unifont: http://unifoundry.com/
+
+### Download on Ubuntu/Debian
```
nebulon42: Maybe better "Installation"

INSTALL.md
```diff
@@ -43,11 +43,19 @@ The repeated www.naturalearthdata.com in the Natural Earth shapefiles is correct
...
 
-Unifont is used as a last resort fallback, with it's excellent coverage, common presence on machines, and ugly look.
+Unifont is used as a last resort fallback, with it’s excellent coverage, common presence on machines, and ugly look.
+
+### Download
+
+Noto: http://www.google.com/get/noto/ and http://github.com/googlei18n?utf8=%E2%9C%93&q=noto
+DejaVu: http://dejavu-fonts.org/
+Unifont: http://unifoundry.com/
+
```
nebulon42: Maybe add a sentence that you have to install those on your own.

INSTALL.md
```diff
@@ -43,11 +43,19 @@ The repeated www.naturalearthdata.com in the Natural Earth shapefiles is correct
 Put these shapefiles at `path/to/openstreetmap-carto/data`.
 
 ## Fonts
-The stylesheet uses Noto Sans, an openly licensed font from Google with support for multiple scripts. The "UI" version is used where available, with its vertical metrics which fit better with Latin text. Other fonts from the Noto family are used for some other languages.
+The stylesheet uses Noto, an openly licensed font from Google with support for multiple scripts. The “UI” version is used where available, with its vertical metrics which fit better with Latin text. The stylesheet uses Noto’s “Sans” style where available. If not available, this stylesheet uses another appropriate style of the Noto family.
```
nebulon42: Why the changed quotation marks?

pnorman: Use straight quotes, not curly quotes.

INSTALL.md
```diff
@@ -43,11 +43,19 @@ The repeated www.naturalearthdata.com in the Natural Earth shapefiles is correct
...
 
 DejaVu Sans is used as an optional fallback font for systems without Noto Sans. If all the Noto fonts are installed, it should never be used.
 
-Unifont is used as a last resort fallback, with it's excellent coverage, common presence on machines, and ugly look.
+Unifont is used as a last resort fallback, with it’s excellent coverage, common presence on machines, and ugly look.
```
pnorman: Quotes were changed here.

INSTALL.md
```diff
@@ -43,11 +43,19 @@ The repeated www.naturalearthdata.com in the Natural Earth shapefiles is correct
...
+
+### Download
```
pnorman: Put the manual download after information about installing from packages, which is a preferred installation method for most cases. Perhaps as "Download from source"

INSTALL.md
```diff
@@ -43,11 +43,19 @@ The repeated www.naturalearthdata.com in the Natural Earth shapefiles is correct
...
+
+Noto: http://www.google.com/get/noto/ and http://github.com/googlei18n?utf8=%E2%9C%93&q=noto
+DejaVu: http://dejavu-fonts.org/
+Unifont: http://unifoundry.com/
```
pnorman: These should be links with markdown, not a list of URLs.

INSTALL.md
pnorman: When referring to "Noto" instead of "Noto Sans" it should be called a "font family" instead of a "font". The information about Sans should come before the information about UI, as UI is a finer division than Sans.

sommerluk:
Thanks for comments. All changes have been addressed.

sommerluk:
Note however that the switch-back to the so-called “straight quotation marks” I’ve done rather reluctant. Switching to typographical quotation marks was not systematic; it’s just that I’m used to write this way. Both, U+0022 and U+0027 are the relicts of old typewriters, which were not able to have enough characters to represent all the typographically correct quotations marks and the apostrophe. In general, they should not be used as quotation marks or apostrophes. Unicode itself (in the character information for these two characters) says that the preferred characters for quotation marks and apostrophe are 201C, 201D and 2019.

All these three characters are part of Unicode since the very beginning, two and a half decades ago. After so much time, I don’t think it is necessary to stay ASCII-compatible anymore.

PR #2610, opened 2017-04-16
Remove support for old-style multipolygons

pnorman:


imagico:
This seems a good idea to me.  Removal of old style tagging is continuing to be very fast so it is unlikely that a significant number of such features will remain when this change goes live even in the most optimistic scenario.  Of course the fairly mechanical re-tagging happening will produce a lot of factually incorrect data but this will not in any way influence the sensibility of this change.

PR #2617, opened 2017-04-21
Move halo's to fill layer

math1985:
This makes the split between lowzoom (<9) and highzoom (>=10) clearer,
and improves performance of z9.

No change in cartographic output.


nebulon42:
I have looked at the code and saw no obvious problems. I'm indifferent on it since I somehow failed to grasp the improvements you mentioned. :)

PR #2621, opened 2017-04-26
Implement unpaved roads.

### LINKED ISSUE #110: Render paved/unpaved

StyXman:
They're implemented as dashed lines for low zooms and dashed casings for the rest. This does not change bridge or tunnel rendering. This attempts to fix #110 .

### DIFF
--- a/roads.mss
+++ b/roads.mss
@@ -385,6 +385,14 @@
         .roads-casing {
           line-join: round;
           line-cap: round;
+          [int_surface = 'unpaved'] {
+            line-dasharray: @primary-width-z12*2,@primary-width-z12;
+            [zoom >= 13] { line-dasharray: @primary-width-z13*2,@primary-width-z13; }
+            [zoom >= 15] { line-dasharray: @primary-width-z15*2,@primary-width-z15; }
+            [zoom >= 17] { line-dasharray: @primary-width-z17*2,@primary-width-z17; }
+            [zoom >= 18] { line-dasharray: @primary-width-z18*2,@primary-width-z18; }
+            [zoom >= 19] { line-dasharray: @primary-width-z19*2,@primary-width-z19; }
+          }
         }
         .tunnels-casing {
           line-dasharray: 4,2;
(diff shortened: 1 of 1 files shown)


imagico:
First of all thanks for investing time in this matter which is important and which so far unfortunately no one seemed to have been interested to work on.

We would need sample renders for this at the various zoom levels of course.

Technically i don't think your approach here is going to work with the limited functionality offered by Mapnik.  If i am not mistaken Mapnik will always draw dashes with the linecaps chosen for the line itself and for the ends we need round line caps while for the dashes you would need butt caps.

Design wise dashes and dashed casing are already used widely in this style for tunnels/underground features and even if your dashes are fairly long this is likely to create confusion - what kind of dashing stands for what.

In general use of dashed lines has proven to be fairly problematic in this style, especially at low zoom levels.  Some time ago we introduced use of dashed lines for intermittent waterways and meanwhile i desperately look for a way to get rid of them again.  Dashing adds a lot of visual noise and makes it more difficult to recognize line geometries and if the geometries are detailed compared to the resolution they are rendered at individual dashes can form fairly arbitrary geometries which can - depending on color - be often confused with other types of mapped geometries.  The repeating pattern which usually makes the dashed line style work will often not be recognizable as such.

StyXman:
I see. I also noticed that to make them stand out in certain situations, I had to make casing thicker, which worsens the round ends problem. It's a shame there's no easy solution, at least for this style, but I plan to implement this in mine, so it's not all wasted time for me.

imagico:
Note from the technical side i did not say this is impossible with Mapnik, just that your current relatively simple approach does not work.  It just is somewhat more complicated - you essentially would have to separately draw the round line end line caps and the dashing somehow.

Personally i still would like to see the grain pattern idea being implemented - which would mainly require someone to look into script-generating the necessary SVGs for that.

pnorman:

As a big change, it's also going to need some detailed review which probably won't happen until after then anyways.

kocio-pl:
Are there any news on this PR? We're not in a feature freeze state any more, but we have still no test renderings, no code review and a merging conflict in `roads.mss`.

kocio-pl:
@StyXman Are you planning to work on this PR? #2640 is the alternative.

StyXman:
As I said earlier, it's getting difficult to make it properly work. Either the dashes are not contrasting enough or making them thicker and I get the round ends, which looks awful. I'm implementing this in my own style, and I saw it in the cycle map. We pull it off because the casings are more contrasting, but I don't see how to do this with the current state of the style and Mapnik. I think I will cancel my attempt, sorry.

kocio-pl:
I guess it's clear enough that this attempt failed, so I'm closing this PR. Thanks a lot for trying a solution for this problem!

PR #2622, opened 2017-04-30
Better font fallback beyond BMP

sommerluk:
Our only current fallback font—Unifont—covers Unicode plane 0 completely. The coverage of plane 1 is not that good. And plane 2 is not covered at all.

Plane 2 is used for CJK characters that did not make in into the set of CJK characters in Plane 0. Plane 2 has currently 52 844 characters. Noto covers 2 238 characters of plane 2.

This PR introduced Hanazono as additional fallback font, and earlier in the list before Unifont. Hanazono covers almost all CJK characters up to CJK-ExtD. In plane 2, it covers 47 082 characters. CJK-ExtE (Unicode 8) stays uncovered. CJK-ExtF (to be released in Unicode 10 later this year) stays also uncovered. Hanazono is dual-licensed under a particular MIT-style license and the well-known SIL OFL. It is normally available as package on modern Linux distributions.

As I do not use Ubuntu, I could not actually test the package install instruction for Ubuntu.

Rendering example with some arbitrary CJK characters (ExtA–ExtE):

Before:
before

After:
after

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -47,7 +47,7 @@ The stylesheet uses Noto, an openly licensed font family from Google with suppor
 
 DejaVu Sans is used as an optional fallback font for systems without Noto Sans. If all the Noto fonts are installed, it should never be used.
 
-Hanazono is used a fallback for seldom CJK characters that are not covered by Noto.
+Hanazono is used a fallback for seldom used CJK characters that are not covered by Noto.
 
 Unifont is used as a last resort fallback, with it's excellent coverage, common presence on machines, and ugly look.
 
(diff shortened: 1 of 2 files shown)


INSTALL.md
```diff
@@ -47,14 +47,16 @@ The stylesheet uses Noto, an openly licensed font family from Google with suppor
 
 DejaVu Sans is used as an optional fallback font for systems without Noto Sans. If all the Noto fonts are installed, it should never be used.
 
+Hanazono is used a fallback for seldom CJK characters that are not covered by Noto.
```
pnorman: either "seldom used CJK..." or "rare CJK..."

sommerluk: Corrected. Thanks.

PR #2631, opened 2017-05-11
Update readme instructions

pnorman:
With the use of lua, osm-carto no longer uses all the osm2pgsql defaults.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -1,7 +1,7 @@
 # Installation
 
 ## OpenStreetMap data
-You need OpenStreetMap data loaded into a PostGIS database (see below for [dependencies](#dependencies)). These stylesheets currently work only with the osm2pgsql defaults (i.e. database name is `gis`, table names are `planet_osm_point`, etc).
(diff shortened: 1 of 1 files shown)


INSTALL.md
```diff
@@ -1,7 +1,7 @@
 # Installation
 
 ## OpenStreetMap data
-You need OpenStreetMap data loaded into a PostGIS database (see below for [dependencies](#dependencies)). These stylesheets currently work only with the osm2pgsql defaults (i.e. database name is `gis`, table names are `planet_osm_point`, etc).
+You need OpenStreetMap data loaded into a PostGIS database (see below for [dependencies](#dependencies)). These stylesheets need to be used with the pgsql backend (table names of `planet_osm_point`, etc), the default database name (`gis`), and the [lua transforms](https://github.com/openstreetmap/osm2pgsql/blob/master/docs/lua.md) documented in the instructions below.
```
imagico: There is no mentioning of _osm2pgsql_ any more in the text now, just _pgsql backend_, i would suggest to rephrase as:

These stylesheets expect a database generated with osm2pgsql using the pgsql backend

Otherwise seems fine.

pnorman: 👍

PR #2637, opened 2017-05-20
Docker development environment

### LINKED ISSUE #657: Provide installation script

nebulon42:
Fixes #657.

This adds necessary files and documentation to setup a development environment for openstreetmap-carto with Docker.

Builds upon work done by @pnorman.

We still have to work out
* if we can prevent get-shapefiles.py from indexing every time it runs and if we need to improve other things with this script
* if the shapefile data is owned by the right user (currently UID 1000). What about systemes where 1000 is not the default user?
* if we should build the containers and upload them to the Docker registry to avoid that everybody has to build them on startup

Would require testing on other platforms than Linux, which I don't have.

cc @Ircama

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,3 @@
+*
+!get-shapefiles.sh
+!openstreetmap-carto.style
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,3 +1,2 @@
 *
-!get-shapefiles.sh
 !openstreetmap-carto.style
--- a/DOCKER.md
+++ b/DOCKER.md
@@ -1,0 +1,1 @@
+# Running OpenStreetMap Carto with Docker
--- a/DOCKER.md
+++ b/DOCKER.md
@@ -1,1 +1,50 @@
 # Running OpenStreetMap Carto with Docker
+
+[Docker](https://docker.com) is a virtualized environment that allows you to run software without altering your system permanently.
+The software runs in so called containers that are easy to setup and tear down. It makes setting up a
+development environment for openstreetmap-carto much easier. The development environment consists of a
+PostgreSQL database to store the data and [Kosmtik](https://github.com/kosmtik/kosmtik) for previewing the style.
+
+## Prerequisites
+
+Docker is available for Linux, macOS and Windows. [Install](https://www.docker.com/get-docker) the software in order
+to be able to run Docker containers. You also need Docker Compose, which should be available once you installed
+Docker itself. Otherwise you need to [install Docker Compose manually](https://docs.docker.com/compose/install/).
+
+## Importing data
+
+openstreetmap-carto needs a database populated with rendering data to work. You first need a data file to import.
+It's probably easiest to grab an PBF of OSM data from [Mapzen](https://mapzen.com/metro-extracts/) or [geofabrik](http://download.geofabrik.de/).
+Once you have that file put it into the openstreetmap-carto directory and run `docker-compose up import` in the openstreetmap-carto directory.
+This downloads (if not present) and starts the PostgreSQL container and builds (if not present) and starts a container that runs
+[osm2pgsql](https://github.com/openstreetmap/osm2pgsql) to import the data.
+
+osm2pgsql has a few [command line options](https://manpages.debian.org/testing/osm2pgsql/osm2pgsql.1.en.html) and the import by default uses
+a RAM cache of 512 MB, 1 worker and expects the import file to be named `data.osm.pbf`. If you want to customize any of these parameters you have
+to set the environment variables `OSM2PGSQL_CACHE` (e.g. `export OSM2PGSQL_CACHE=1024` on Linux to set the cache to 1 GB)
+for the RAM cache (the value depends on the amount of RAM you have available, the more you can use
+here the faster the import may be), `OSM2PGSQL_NUMPROC` for the number of workers (this depends on the number of processors you have and whether your
+harddisk is fast enough e.g. is a SSD), or `OSM2PGSQL_DATAFILE` if your file has a different name. If you instead want
+to customize the default values you find them in the file `.env`.
+
+Depending on your machine and the size of the extract the import can take a while. When it is finished you should have the data necessary to
+render it with openstreetmap-carto.
+
+## Test Rendering
+
+After you have the necessary data available you can start Kosmtik to produce a test rendering. For that
+you run `docker-compose up kosmtik` in the openstreetmap-carto directory. This builds (if not present) and starts a container
+that installs Kosmtik. It also downloads (if not present) and starts the PostgreSQL database container if it is not already running.
+After install the script `scripts/docker-startup.sh` is invoked which downloads necessary
+shapefiles with `scripts/get-shapefiles.py` (if not present) and indexes them. It afterwards runs Kosmtik. If you
+have to customize anything you can do so in the script. The Kosmtik config file can be found in `.kosmtik-config.yml`.
+If you want to have a [local configuration](https://github.com/kosmtik/kosmtik#local-config) for our `project.mml` you
+can place a `localconfig.js` or `localconfig.json` file into the openstreetmap-carto directory.
+
+The shapefile data that is downloaded is owned by the user with UID 1000. If you have another default user id on your system consider
+changing the line `USER 1000` in the file `Dockerfile`.
+
+After startup is complete you can browse to `http://localhost:6789` to view the output of Kosmtik. By pressing Ctrl + C on the
+command line you can stop the container. The PostgreSQL database container is still running then (you can check with `docker ps`). If you
+want to stop the database container as well you can do so by running `docker-compose stop db` in the
+openstreetmap-carto directory.
--- a/Dockerfile.import
+++ b/Dockerfile.import
@@ -1,0 +1,11 @@
+FROM debian:stretch
+
+RUN apt-get update && apt-get install --no-install-recommends -y \
+    curl ca-certificates osm2pgsql postgresql-client \
+    && rm -rf /var/lib/apt/lists/*
+
+ADD openstreetmap-carto.style /
+
+CMD createdb gis && psql -d gis -c 'CREATE EXTENSION postgis;' && \
+    osm2pgsql --slim --cache 512 -d gis --style openstreetmap-carto.style \
+    $OSM_DATA
--- a/Dockerfile.import
+++ b/Dockerfile.import
@@ -6,6 +6,18 @@ RUN apt-get update && apt-get install --no-install-recommends -y \
 
 ADD openstreetmap-carto.style /
 
-CMD createdb gis && psql -d gis -c 'CREATE EXTENSION postgis;' && \
-    osm2pgsql --slim --cache 512 -d gis --style openstreetmap-carto.style \
-    $OSM_DATA
+RUN mkdir -p /openstreetmap-carto
+WORKDIR /openstreetmap-carto
+
+CMD psql -c "SELECT 1 FROM pg_database WHERE datname = 'gis';" | grep -q 1 || createdb gis && \
+    psql -d gis -c 'CREATE EXTENSION IF NOT EXISTS postgis;' && \
+    psql -d gis -c 'CREATE EXTENSION IF NOT EXISTS hstore;' && \
+    osm2pgsql \
+    --cache $OSM2PGSQL_CACHE \
+    --number-processes $OSM2PGSQL_NUMPROC \
+    --hstore \
+    --multi-geometry \
+    --database gis \
+    --style openstreetmap-carto.style \
+    --tag-transform-script openstreetmap-carto.lua \
+    $OSM2PGSQL_DATAFILE
--- a/scripts/docker-startup.sh
+++ b/scripts/docker-startup.sh
@@ -1,0 +1,13 @@
+#!/bin/sh
+
+python scripts/get-shapefiles.py
+
+mkdir -p .config
+if [ ! -e ".kosmtik-config.yml" ]; then
+  touch .kosmtik-config.yml
+  # this can be removed once https://github.com/kosmtik/kosmtik/issues/236 is resolved
+  echo "plugins:" >> .kosmtik-config.yml
+fi
+export KOSMTIK_CONFIGPATH=".kosmtik-config.yml"
+
+kosmtik serve project.mml --host 0.0.0.0
--- a/.env
+++ b/.env
@@ -1,0 +1,5 @@
+# environment defaults for import
+
+OSM2PGSQL_CACHE=512
+OSM2PGSQL_NUMPROC=1
+OSM2PGSQL_DATAFILE=data.osm.pbf
--- a/.gitignore
+++ b/.gitignore
@@ -1,10 +1,13 @@
 .thumb.png
 layers/
 data/
+tmp/
 *.xml
+*.osm.pbf
 node_modules/
 localconfig.json
 localconfig.js
+.kosmtik-config.yml
 
 # Generated at runtime by Python.
 *.pyc
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,19 @@
+FROM debian:stretch
+
+# Style dependencies
+RUN apt-get update && apt-get install --no-install-recommends -y \
+    unzip curl ca-certificates \
+    fonts-noto-cjk fonts-noto-hinted fonts-noto-unhinted ttf-unifont \
+    mapnik-utils \
+    nodejs nodejs-legacy npm \
+    && rm -rf /var/lib/apt/lists/*
+
+RUN mkdir -p /openstreetmap-carto
+WORKDIR /openstreetmap-carto
+RUN npm install kosmtik
+
+# Get the shapefiles
+ADD get-shapefiles.sh ./
+RUN ./get-shapefiles.sh
+
+CMD nodejs ./node_modules/kosmtik/index.js serve /openstreetmap-carto/project.yaml --host 0.0.0.0
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,18 +2,20 @@ FROM debian:stretch
 
 # Style dependencies
 RUN apt-get update && apt-get install --no-install-recommends -y \
-    unzip curl ca-certificates \
+    unzip curl ca-certificates gnupg python \
     fonts-noto-cjk fonts-noto-hinted fonts-noto-unhinted ttf-unifont \
     mapnik-utils \
-    nodejs nodejs-legacy npm \
     && rm -rf /var/lib/apt/lists/*
+
+# Node.js 6.x LTS
+RUN curl -sL https://deb.nodesource.com/setup_6.x | bash \
+    && apt-get install --no-install-recommends -y nodejs \
+    && rm -rf /var/lib/apt/lists/*
+
+RUN npm install -g kosmtik
 
 RUN mkdir -p /openstreetmap-carto
 WORKDIR /openstreetmap-carto
-RUN npm install kosmtik
 
-# Get the shapefiles
-ADD get-shapefiles.sh ./
-RUN ./get-shapefiles.sh
-
-CMD nodejs ./node_modules/kosmtik/index.js serve /openstreetmap-carto/project.yaml --host 0.0.0.0
+USER 1000
+CMD sh scripts/docker-startup.sh
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,0 +1,28 @@
+version: '2'
+services:
+  kosmtik:
+    image: kosmtik:v1
+    build:
+      context: .
+      dockerfile: Dockerfile
+    volumes:
+      - .:/openstreetmap-carto
+    depends_on:
+      - db
+    ports:
+      - "6789:6789"
+    environment:
+      - PGHOST=db
+      - PGUSER=postgres
+    restart: on-failure
+  db:
+    image: mdillon/postgis:9.6
+  import:
+    image: import:v1
+    build:
+      context: .
+      dockerfile: Dockerfile.import
+    depends_on:
+      - db
+    environment:
+      - OSM_DATA
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -14,7 +14,6 @@ services:
     environment:
       - PGHOST=db
       - PGUSER=postgres
-    restart: on-failure
   db:
     image: mdillon/postgis:9.6
   import:
@@ -22,7 +21,13 @@ services:
     build:
       context: .
       dockerfile: Dockerfile.import
+    volumes:
+      - .:/openstreetmap-carto
     depends_on:
       - db
     environment:
-      - OSM_DATA
+      - PGHOST=db
+      - PGUSER=postgres
+      - OSM2PGSQL_CACHE
+      - OSM2PGSQL_NUMPROC
+      - OSM2PGSQL_DATAFILE
--- a/scripts/get-shapefiles.py
+++ b/scripts/get-shapefiles.py
@@ -259,7 +259,9 @@ def main():
    (indexing shapes is suggested for performance improvement).\n""")
 
     if args.option_force:
-        os.chdir(os.path.dirname(__file__))
+        path = os.path.dirname(__file__)
+        if len(path) > 0:
+            os.chdir(path)
     else:
         os.chdir(os.path.join(os.path.dirname(__file__), '..'))
         if not os.path.isfile("project.mml"):


kocio-pl:
Testing on Windows can be done quite easily nowadays, because one can [download virtual MS Windows images](https://developer.microsoft.com/en-us/microsoft-edge/tools/vms/) (aka http://modern.ie). They are meant to test MSIE/Edge, but it's just plain Windows under the hood. 

On the other hand it's worth to remember that for Windows 7/8.1 there's only 32-bit version (however Windows 10 images are for x64) and the performance may be quite bad (common problem with virtualization).

Ircama:
> if we can prevent get-shapefiles.py from indexing every time it runs

What would you suggest as appropriate behavior?

> and if we need to improve other things with this script

Please let me know some wishes, I'll try to do my best

nebulon42:
@Ircama I just pinged you because I thought you might be interested in testing this on Windows.
@kocio-pl Thanks for the link, I know the VMs. I guess if someone else than me tests this it is far better, because if you have a certain level of knowledge about osm-carto then you are likely to overlook things that might be problematic for others.

Regarding get-shapefiles.py: It is a good script and I don't know of any necessary improvements yet, practice will have to show if it needs changes. The only difference here would be that the script now runs every time the container starts compared to manually as before. So I noticed the shapefile indexing. This depends on shapeindex and I don't know if you can detect if a shapefile has been already indexed. But the best solution would be to index only when it is needed.

Ircama:
@nebulon42 OK, thanks, will do in the next few days

kocio-pl:
`docker-startup.sh` could check for `*.index` files in `data` directory during startup and skip indexing if they are present (if I understand correctly).

nebulon42:
@kocio-pl Ah, I missed that. Yes, that seems right although I think it should be `get-shapefiles.py` that does the checking.

pnorman:
I haven't had a chance to review this yet since I've been on vacation.

math1985:
Would this help us?
https://dzone.com/articles/generating-maps-with-mapnik-and-docker

nebulon42:
Not specifically. I think the docker work is basically finished with this PR. What remains is streamlining.

Ircama:
Unfortunately, at the moment I do not have enough prerequisites to perform a valid testing of Docker natively running on Windows (it needs Hyper-V support included in Windows 10 Pro, Enterprise or Education and not in Windows 10 Home or any Windows 7 version). 

I could only test Docker Toolbox, which creates a Linux VM on VirtualBox hosting Docker.

Also, Docker Toolbox does not seem to support Win7 Win32.

I tested Docker Toolbox on a Win7 Ent 64 bit but could not succeed (maybe I was my problem).

The installation seems to freeze after `db_1       | LOG:  database system is ready to accept connections`

Also, from the following error OSM-carto does not seem to be successfully installed: ` import_1   | Osm2pgsql failed due to ERROR: Couldn't open style file 'openstreetmap-carto.style': No such file or directory`

Steps I did:

- Installed Docker Toolbox for Windows

- Run *Docker Quickstart Terminal*
Run --> All Programs --> Docker --> Docker Quickstart Terminal

A MINGW Console starts.

- Download [Docker openstreetmap-carto](https://github.com/nebulon42/openstreetmap-carto/tree/4c1029d8f5715ce94337322460bfc5bf03f16cfe) to a local directory (so that the following files are available: *docker-compose.yml, Dockerfile, Dockerfile.import*

download a *pbf* to the same directory, as mentioned in [DOCKER.md]( https://github.com/nebulon42/openstreetmap-carto/blob/4c1029d8f5715ce94337322460bfc5bf03f16cfe/DOCKER.md).

    cd <local directory where openstreetmap-carto has been downloaded>

    docker-compose up

The setup installation starts and accomplishes almost all procedures, than it freezes at the following point:

```
db_1       | LOG:  aborting any active transactions
db_1       | LOG:  autovacuum launcher shutting down
db_1       | waiting for server to shut down....LOG:  shutting down
db_1       | LOG:  database system is shut down
db_1       |  done
db_1       | server stopped
db_1       |
db_1       | PostgreSQL init process complete; ready for start up.
db_1       |
db_1       | LOG:  database system was shut down at 2017-05-28 11:30:30 UTC
db_1       | LOG:  MultiXact member wraparound protections are now enabled
db_1       | LOG:  database system is ready to accept connections
```

After issuing a Control C and doing again *docker-compose up*:

```
Starting openstreetmapcarto4c1029d8f5715ce94337322460bfc5bf03f16cfe_db_1
Starting openstreetmapcarto4c1029d8f5715ce94337322460bfc5bf03f16cfe_kosmtik_1
Starting openstreetmapcarto4c1029d8f5715ce94337322460bfc5bf03f16cfe_import_1
Attaching to openstreetmapcarto4c1029d8f5715ce94337322460bfc5bf03f16cfe_db_1, openstreetmapcarto4c1029d8f5715c
e94337322460bfc5bf03f16cfe_import_1, openstreetmapcarto4c1029d8f5715ce94337322460bfc5bf03f16cfe_kosmtik_1
db_1       | LOG:  database system was shut down at 2017-05-28 11:37:12 UTC
db_1       | LOG:  MultiXact member wraparound protections are now enabled
kosmtik_1  | sh: 0: Can't open scripts/docker-startup.sh
openstreetmapcarto4c1029d8f5715ce94337322460bfc5bf03f16cfe_kosmtik_1 exited with code 127
import_1   | CREATE EXTENSION
import_1   | CREATE EXTENSION
import_1   | osm2pgsql version 0.92.0 (64 bit id space)
import_1   |
import_1   | Osm2pgsql failed due to ERROR: Couldn't open style file 'openstreetmap-carto.style': No such file
 or directory
openstreetmapcarto4c1029d8f5715ce94337322460bfc5bf03f16cfe_import_1 exited with code 1
```

I also performed

    docker-compose up import

getting the same error message.

```
import_1   | Osm2pgsql failed due to ERROR: Couldn't open style file 'openstreetmap-carto.style': No such file or directory
```

I also removed the packages with the following commands:

```
docker-compose down
docker images
for each image, docker rmi <image ID>
```

Then issuing again

    docker-compose up

The errors are the same.

Ircama:
You can find [here](https://github.com/Ircama/openstreetmap-carto/blob/get-shapefiles2/scripts/get-shapefiles.py) a new version of *get-shapefiles.py* (not yet pushed).

In addition to only downloading archive updates by default, it now limits archive expansion to successfully downloaded archives and performs indexes only if they are missing or not up to date.

Usage and command line options remain unchanged, with a slightly different meaning for `-u`/`--update` option, which allows to force downloading, expanding and indexing operations even if not needed.

Ircama:
I tested the procedure on a different PC running MS Windows 10 Home x64 (slow i3 CPU with 4 GB RAM) and it works, taking about 2.87 GB in *C:\Users\\<user\>\\.docker*.

It is strongly advisable to use a much faster machine (i7) with at least 8 or 12 GB RAM.

The problems mentioned while testing an x64 Win 7 Ent PC remains.

Procedure for x64 Windows versions requiring Docker Toolbox:

- Check documentation in [DOCKER.md](https://github.com/nebulon42/openstreetmap-carto/blob/4c1029d8f5715ce94337322460bfc5bf03f16cfe/DOCKER.md)
- Install the latest version of [Oracle VM VirtualBox](https://www.virtualbox.org/wiki/Downloads) (this is advisable even if Docker Toolbox includes it)
- Install [Docker Toolbox](https://www.docker.com/products/docker-toolbox)
- Download [openstreetmap-carto](https://github.com/nebulon42/openstreetmap-carto/tree/4c1029d8f5715ce94337322460bfc5bf03f16cfe) (the version including Docker files) to a local directory
- Run *Docker Quickstart Terminal* (Run → All Programs → Docker → Docker Quickstart Terminal). Wait for the MINGW Console to start.
  Notice that *"Waiting for an IP..."* takes the time needed to start a virtual machine named *default* and boot a Linux image included in a file named *boot2docker.iso* (automatically downloaded and configured by the software).
- Change directory to the local directory where *openstreetmap-carto* has been downloaded.

      cd <local directory where openstreetmap-carto has been downloaded>

- Download a PBF of OSM data to the same directory where openstreetmap-carto has been downloaded, as mentioned in [DOCKER.md](https://github.com/nebulon42/openstreetmap-carto/blob/4c1029d8f5715ce94337322460bfc5bf03f16cfe/DOCKER.md). The dowloaded file shall be named *data.osm.pbf*. E.g.:

      curl http://download.geofabrik.de/europe/liechtenstein-latest.osm.pbf --output data.osm.pbf

- Set-up the Docker environment and import data:

      docker-compose up

  The procedure takes many minutes to complete. Wait for *kosmtik:1  | [Core] Loading map*

- Kosmtik can be run via a browser pointing to

      http://192.168.99.100:6789

  The IP address can be found by running another *Docker Quickstart Terminal* and reading the login message (*docker is configured to use the default mackine with IP ...*).
  The actual IP address range can be found in the VM VirtualBox application: *File* → *Preferences* → *Network* tab → *Host-only network* tab, select the second adapter, *DHCP Server* tab.
  The list of used IP addresses can be obtained through the following command:

      docker-machine ssh default ifconfig|grep "inet addr"

To stop Kosmtik, select the MINGW64 window and press *Control C*.
 
To totally remove the docker installation, delete the virtual machine named *default* and the directory *C:\Users\\<user\>\\.docker*.

While `docker-compose up` is running, the *openstreetmap-carto* development can be directly done on the PC, by accessing the local directory where *openstreetmap-carto* has been downloaded. E.g., editing a stylesheet and then pressing the Kosmtik *Reload* button, the map gets updated.

The RAM size available to the *default* VM should be increased for improved performance. To increase the VM memory size:

    docker-machine stop default

Change the memory of the *default* virtual machine though Oracle VM VirtualBox

    docker-machine start default
    docker-compose up

To maximize the performance of *scripts/get-shapefiles.py*, edit the script in the local directory and add the `-n` option (I tested get-shapefiles.py version 2).

Ircama:
While testing, I verified that both *Dockerfile.import* and *Dockerfile* do not check whether the db is running and this leads to sync errors between the db container and the other ones, as they start in parallel to the db instead of in sequence.

A way that I found to fix this was the following:

- *Docker-startup.sh* has to be executed by both *kosmtik* and *import* containers with different arguments; It needs to include a trivial procedure to check that the DB is running (via *pg_isready*), followed by a differentiated execution dependent on the command line argument
- Kosmtik *Dockerfile* needs to also include *postgresql-client* within the *apt-get install* (used by *Docker-startup.sh*)
- *Docker-compose.yml* has to include an entrypoint for each of the two *kosmtik* and *import* services: `scripts/docker-startup.sh kosmtik` and `scripts/docker-startup.sh postgres`
- *CMD* is removed from both *Dokerfile* and *Dokerfile.import* (Dokerfile.import CMD script is moved to Docker-startup.sh); CMD might be useless as replaced by the entrypoints.

Find here attached these modifications.

[modifications.zip](https://github.com/gravitystorm/openstreetmap-carto/files/1059622/modifications.zip)

Would it be worthwhile to incorporate them?

Ircama:
> Testing on Windows can be done quite easily nowadays, because one can download virtual MS Windows images (aka http://modern.ie). They are meant to test MSIE/Edge, but it's just plain Windows under the hood.

> On the other hand it's worth to remember that for Windows 7/8.1 there's only 32-bit version (however Windows 10 images are for x64) and the performance may be quite bad (common problem with virtualization).

I'm afraid these VM cannot be used to test Docker or Docker Toolbox, as both need native 64-bit virtualization (to be physically set within the BIOS). Also I do not think that a Windows 10 x64 cloud instance can support Docker, as nested virtualization is not allowed generally.
To be able to test Docker for Windows, you need to install Windows 10 Enterprise on a physical machine and enable virtualization, for Hyper-V support.
To be able to test Docker Toolbox for Windows, you need to be able to install Oracle VM VirtualBox on a 64-bit physical machine, anyway enabling virtualization.

kocio-pl:
What is needed to merge this code? I think at least:
- `get-shapefiles.py` fix should be dropped (as [this part will be replaced in other PR](https://github.com/gravitystorm/openstreetmap-carto/pull/2653/files#diff-ad01b6a9f53516f29612c7bd9948ec20L262))
- documentation for Windows (from @Ircama [comment](https://github.com/gravitystorm/openstreetmap-carto/pull/2637#issuecomment-305969770)) should be added
- add fix for database [suggested in different Ircama comment](https://github.com/gravitystorm/openstreetmap-carto/pull/2637#issuecomment-306961334)

Is this TODO correct? Anything else missing here?

DOCKER.md
```diff
@@ -0,0 +1,50 @@
...
+## Prerequisites
+
+Docker is available for Linux, macOS and Windows. [Install](https://www.docker.com/get-docker) the software in order
+to be able to run Docker containers. You also need Docker Compose, which should be available once you installed
+Docker itself. Otherwise you need to [install Docker Compose manually](https://docs.docker.com/compose/install/).
+
+## Importing data
+
+openstreetmap-carto needs a database populated with rendering data to work. You first need a data file to import.
+It's probably easiest to grab an PBF of OSM data from [Mapzen](https://mapzen.com/metro-extracts/) or [geofabrik](http://download.geofabrik.de/).
```
kocio-pl: Just small fixes here:
- https://mapzen.com/metro-extracts/ -> https://mapzen.com/data/metro-extracts/
- geofabrik -> Geofabrik

nebulon42:
I have rewritten the path fix as in the new script version. I wanted to keep that in until the other PR is merged. Updated the doc with comments from @kocio-pl.
Tested and merged the database check, thanks @Ircama.

I did not add documentation for Windows as I think this can be done after merge in a separate PR by e.g. @Ircama.

nebulon42:
I would be ready for merging once there is a review by a fellow maintainer.

kocio-pl:
Great! I will most probably test if it works in a day or two, but I'm unable to do a code review. @pnorman - are you still interested in reviewing this PR?

math1985:
Thanks a lot! I will see if I can test this next week. Is the change in water.mss intended?

nebulon42:
@math1985 definitely not. hm, the diff looks weird. I'll try to move the commits into a clean branch, let's see if that helps.

nebulon42:
Diff looks better now, I probably forgot to rebase. But now I have recreated the branch from scratch.

Ircama:
I'm afraid that something went wrong with merging and *docker-startup.sh* is now changed to DOS style (CR/LF) instead of UNIX style.

I issued the following steps for testing with Windows 7 PRO:

run "Docker QuickStart Terminal"; at command prompt:
```
cd
git clone -b docker https://github.com/nebulon42/openstreetmap-carto
cd openstreetmap-carto
curl http://download.geofabrik.de/europe/liechtenstein-latest.osm.pbf --output data.osm.pbf
docker-compose up
```

It fails (needs UNIX style):

```
: not found| scripts/docker-startup.sh: 2: scripts/docker-startup.sh:
kosmtik_1  | scripts/docker-startup.sh: 14: scripts/docker-startup.sh: Syntax error: word unexpected (expecting "in")
```

Anyway, if I change the file format of  *docker-startup.sh* to UNIX, it works:

```
dos2unix scripts/docker-startup.sh
docker-compose up
```

I guess that a new commit after converting *docker-startup.sh* might fix the issue.

If you prefer to redo from scratch, the only files that I changed were *Dockerfile*, *Dockerfile.import* and *scripts/docker-startup.sh* (files under *https://github.com/Ircama/openstreetmap-carto/tree/docker* look correct to me).

Anyway, the whole list of files needed for this PR should be:
.dockerignore
.env
.gitignore
DOCKER.md
Dockerfile
Dockerfile.import
docker-compose.yml
scripts/docker-startup.sh
(and scripts/get-shapefiles.py; notice that @kocio-pl suggested to remove it)

DOCKER.md
```diff
@@ -0,0 +1,50 @@
...
+Docker is available for Linux, macOS and Windows. [Install](https://www.docker.com/get-docker) the software in order
+to be able to run Docker containers. You also need Docker Compose, which should be available once you installed
+Docker itself. Otherwise you need to [install Docker Compose manually](https://docs.docker.com/compose/install/).
+
+## Importing data
+
+openstreetmap-carto needs a database populated with rendering data to work. You first need a data file to import.
+It's probably easiest to grab an PBF of OSM data from [Mapzen](https://mapzen.com/data/metro-extracts/) or [Geofabrik](http://download.geofabrik.de/).
+Once you have that file put it into the openstreetmap-carto directory and run `docker-compose up import` in the openstreetmap-carto directory.
+This downloads (if not present) and starts the PostgreSQL container and builds (if not present) and starts a container that runs
```
math1985: The grammar is not clear to me, what is being downloaded exactly?

math1985:
Testing this on Mac. I get the following errors, is this a problem?

Attaching to openstreetmapcarto_kosmtik_1
kosmtik_1  | Waiting for PostgreSQL to be running
kosmtik_1  | Use of uninitialized value $gid in getgrgid at /usr/share/perl5/PgCommon.pm line 783.
kosmtik_1  | Use of uninitialized value $gid in getgrgid at /usr/share/perl5/PgCommon.pm line 783.
kosmtik_1  | Use of uninitialized value $gid in getgrgid at /usr/share/perl5/PgCommon.pm line 783.

math1985:
So the procedure is basically:

*  Download a pbf to data.osm.pbf
* `docker-compose up import`
* `docker-compose up kosmtik`

Maybe we can add a 'Quick start' section on top that explains this?

Ircama:
In the consideration that this Docker configuration might be of great help to quickly setup the Kosmtik dev environment, I would suggest to shortly mention it also in the [INSTALL](https://github.com/gravitystorm/openstreetmap-carto/blob/master/INSTALL.md) page.

math1985:
Confirmed that this works on Docker for Mac. Once again, great work!

math1985:
I get the following error message repeated on my terminal:

kosmtik_1  | Mapnik LOG> 2017-07-06 09:54:40:  could not create BreakIterator: U_MISSING_RESOURCE_ERROR

Rendering seems to work fine, but does anybody knows what this means? I haven't seen it before.

math1985:
Thinking about the next step already: how hard would it be to extend this to a production-ready rendering server? It would be really great if people could create themselves a rendering server with a single docker command.

Probably we don't need kosmtik for that, and would need an apache/mod_tile instead. Doesn't sound too hard?

.dockerignore
```diff
@@ -0,0 +1,2 @@
+*
```
math1985: What does this do again? Why do we ignore all files?

Dockerfile.import
```diff
@@ -0,0 +1,12 @@
+FROM debian:stretch
+
+RUN apt-get update && apt-get install --no-install-recommends -y \
```
math1985: We need to do the debian update now twice, once for the import step and once when we run. Doing the debian update is one of the slowest steps on my connection. Is there a way we can do the update only once?

scripts/docker-startup.sh
```diff
@@ -0,0 +1,43 @@
...
+do
+	pg_isready -q && echo "PostgreSQL running" && break
+	sleep 2
+	i=$((i+1))
+done
+test $i -gt $MAXCOUNT && echo "Timeout while waiting for PostgreSQL to be running"
+
+case "$1" in
+import)
+	psql -c "SELECT 1 FROM pg_database WHERE datname = 'gis';" | grep -q 1 || createdb gis && \
```
math1985: Tab instead of space

Dockerfile.import
nebulon42: You mean for each container image import and kosmtik? No, there is no way to generalize that as both are completely independent from each other. But the containers are only build if you start them the first time, so this is only done once.

Ideally we would publish both images to the Docker registry so that users only would have to download the pre-build images and not have to build them themselves. But this opens up questions about versioning of the images. Separately or in line with the style? The images will not be updated often.

Also the import should be only done once if you don't change your data. You have to be careful not to type `docker-compose up` but instead `docker-compose up import` or `docker-compose up kosmtik`. The first command without target will start both containers simultaneously.

scripts/docker-startup.sh
nebulon42: I have fixed all instances of tabs now

.dockerignore
nebulon42: We don't need any of the files from openstreetmap-carto for building the containers. This avoids that those files are included in the build context. See https://docs.docker.com/engine/reference/builder/#dockerignore-file.

nebulon42:
There were quite some comments. I try to address them all in one reply:

> I'm afraid that something went wrong with merging and docker-startup.sh is now changed to DOS style (CR/LF) instead of UNIX style.

Are you sure? On my system (Ubuntu) it tells me that the file is formatted with LF only? Maybe it has something to do with how Git is configured to handle line endings on your machine. Maybe this could point you in the right direction: https://help.github.com/articles/dealing-with-line-endings/

> kosmtik_1 | Use of uninitialized value $gid in getgrgid at /usr/share/perl5/PgCommon.pm line 783.

I have seen this too and have ignored it. It has something to do with the check if the database is ready. Since it is working I did not bother to investigate further.

> kosmtik_1 | Mapnik LOG> 2017-07-06 09:54:40: could not create BreakIterator: U_MISSING_RESOURCE_ERROR

This comes directly from Mapnik and I have seen this occasionally before (ref https://github.com/mapnik/mapnik/issues/3193). In my experience it never affected rendering.

> Thinking about the next step already: how hard would it be to extend this to a production-ready rendering server?

Not too hard probably. But I think this is out of scope for this repository. We could easily start a separate project for that if it not exists already.

I have fixed tabs/spaces in docker-startup.sh
I have updated the wording in DOCKER.md and have added a quick start section.
I have mentioned Docker in INSTALL.md

kocio-pl:
I've just tested the quick start procedure under Ubuntu 17.04 and it works. I also think we can merge this PR soon and fine tune it later. 

The only thing I suppose would be good to do now, would be to have the database (and maybe some configuration directories?) mounted as a Docker volume, so the data would be independent from the code and survive removing/replacing the containers. Liechtenstein is small, but if I'm testing bigger country/continent/planet data, importing may be way too painful.

Other nice thing would be to include Kosmtik plugins - probably all of them, but it's less important.

nebulon42:
> The only thing I suppose would be good to do now, would be to have the database (and maybe some configuration directories?) mounted as a Docker volume, so the data would be independent from the code and survive removing/replacing the containers.

The database runs in a separate container. So as long as you do not remove the database container your data is safe. You can delete and replace the import and kosmtik containers as much as you like. Your data should not be affected.

I see no real benefit in syncing the database files out to the host as this only adds overhead and if you accidentally delete that directory it would be the same if you accidentally have deleted the database container.

Kosmtik plugins would be good, but PRs for that would of course be possible also after merging. :)

Dockerfile.import
math1985: I think what we can do is add a `RUN apt-get update` line after the FROM line in both Dokerfile.import and Dockerfile. That way, we only need to do the update for the import container, and then we can use a cached container when doing the import for the main container.

In the line where we are doing `apt-get install` we need to keep the `apt-get update` as well (as is a Docker best practice), so we force docker to do an update whenever we change the dependencies in that line.

kocio-pl:
OK, it may be just me being paranoid about mixing code and data. ;-) So, in my opinion you can merge it now (I don't know if we should squeeze 4 commits into 1 or not). 

Thanks a lot for your work with this PR, Michael!

scripts/docker-startup.sh
```diff
@@ -0,0 +1,43 @@
...
+  $OSM2PGSQL_DATAFILE
+  ;;
+kosmtik)
+  python scripts/get-shapefiles.py -n
+
+  mkdir -p .config
+  if [ ! -e ".kosmtik-config.yml" ]; then
+    touch .kosmtik-config.yml
+    # this can be removed once https://github.com/kosmtik/kosmtik/issues/236 is resolved
+    echo "plugins:" >> .kosmtik-config.yml
```
kocio-pl: [This issue](https://github.com/kosmtik/kosmtik/issues/236) is now resolved, so both lines can be removed.

scripts/docker-startup.sh
```diff
@@ -0,0 +1,43 @@
+#!/bin/sh
+
```
kocio-pl: I think we should add short description what is this script about, like:

`# This script is used internally by Docker development environment, you can read more about it in DOCKER.md.`

scripts/docker-startup.sh
nebulon42: Thanks for noticing. :) Technically yes, but since a new release of Kosmtik is missing we have to leave it in for now.

scripts/docker-startup.sh
nebulon42: added.

nebulon42:
I will not squash because I think it is important to retain the authorship of @pnorman and @Ircama. Thanks go also to them.

math1985:
> Not too hard probably. But I think this is out of scope for this repository. We could easily start a separate project for that if it not exists already.

I'm not sure if it is outside scope. Purely technically there might be better places for it, on the other hand if it's only a small step compared to what we have already, I think we might just do it here. In a way it fits here, because it is a description for server admins to on how to deploy our stylesheet.

nebulon42:
> I'm not sure if it is outside scope. ...

Yes you could see it like that. But it boils down to if we want to maintain that too. :)

nebulon42:
To be able to backport this to 3.x we would have to change the osm2pgsl call inside docker-startup.sh. Do we want to do that?

math1985:
Personally I wouldnt spend time on that.

Ircama:
> Are you sure? On my system (Ubuntu) it tells me that the file is formatted with LF only?

@nebulon42 You are right and this issue is rather largely discussed with Docker for Windows, which allows intalling Git MSYS-git UNIX tools, used to help dealing with git-based projects and providing a UNIX style prompt to *Docker Quickstart Terminal*. When cloning a git project, by default Git for Windows replaces Linux Line Feed by Windows Carriage Return, dangerous for any repo being developed with Windows and delpoyed on Linux Dockers. For openstreetmap-carto I would suggest restoring the *.gitattributes* file for *docker-startup.sh* so that it will always have LF line endings.

PR #2639, opened 2017-05-22
Increase node version for travis

pnorman:


pnorman:
cc @nebulon42

nebulon42:
Seems sensible. Node v4 is supported until 2018-04-01. Switching to v6 would be also possible or have it additionally.

Failure is due to mapnik-reference, see https://github.com/mapnik/mapnik-reference/issues/133

Issue #2558, opened 2017-01-24
turning_circle rendered on top of more important features


naoliv:
It seems that in some places, a `highway=turning_circle` is being rendered on top of more important features.

For example, this one https://www.openstreetmap.org/node/1538576570 is on top of a trunk highway.

At zoom level 17:
![](https://cloud.githubusercontent.com/assets/982338/22245991/d7cd977a-e219-11e6-8365-a379f2e44a5c.png)

16:
![](https://cloud.githubusercontent.com/assets/982338/22246002/e07ae0bc-e219-11e6-9ceb-930c0a808c62.png)

15:
![](https://cloud.githubusercontent.com/assets/982338/22246023/ed581f3e-e219-11e6-9c3e-369bb50e3e91.png)

Could this be improved, please?
It's strange to see a white circle above a trunk highway, for example.

pnorman:
I've been looking at fixing this post-lua as part of road simplifications and combining stuff into one layer.

With the current large join I wouldn't recommend tackling a rewrite like that.

The current layers involved are
```
tunnels
turning-circle-casing
highway-area-casing
roads-casing
highway-area-fill
roads-fill
turning-circle-fill
bridges
```

This puts turning circle fills above all road fills and turning circle casings below all road casings.

Moving `turning-circle-fill` to be next to `highway-area-fill` might work.

kocio-pl:
@naoliv Would you like to test the proposed solution?

Issue #1975, opened 2015-11-14
Roadmap / strategic discussion


matthijsmelissen:
With the road update rolled out, a large part of the style sheet has now been modified since the stylesheet development has been taken up by the current tram. Roads, buildings, landuse colours, and icons have all been changed to a greater or lesser extent.

I think that makes it time again to have a look at the bigger picture.

What should development concentrate on over the next year or so? What are the issues that need improving most at this point?

This discussion should lead to an update of the roadmap in README.md.

CC @gravitystorm @pnorman @matkoniecz.

math1985:
Some more questinons:
* Which zoom levels need attention most?
* Are there any particular issues in the issue tracker that are of importance?
* Are there particular types of features that need attention?

To start off the discussion, some potential points of attention (some of which might overlap):
* Reduce number of points of interest (POI)
* Adding more POI
* Adding more landuse
* Improving landuse colours
* Improve layering/rendering order
* Improve rendering of admin borders
* hstore support
* Improve performance (compilation and rendering)

pnorman:
I'm going to split my comments up, since they'll be on a few different issues and I don't want to lose partial posts (again).

### Compilation performance ###
#1947 and #1944 helped here, but we're still about 10 seconds. We used to be under 2.5 and were still one of the slower stylesheets to compile.

I recently compiled osm-carto, Stamen Toner, CartoDB basemaps, Humanitarian, OSM Bright, and the unnamed Wikimedia OSM Bright 2 derived style. Nothing was close to us in compilation time, even with the Humanitarian bug that prevented compilation on carto 0.15.0 (hotosm/HDM-CartoCSS#277).

This graph is a based on [a blog post](http://paulnorman.ca/blog/2015/11/openstreetmap-carto-complexity/) but updated to include current changes and expanded where we were a long time between releases, using ``git checkout $(git rev-list --merges --first-parent -n 1 --before="$TARGET_DATE" master)``.

![image](https://cloud.githubusercontent.com/assets/1190866/11172399/da54ee18-8bbb-11e5-89f2-e12a455ed6d8.png)

This increases the delay between saving a style change and seeing its effects, and at high zooms working with an extract the style recompilation is the biggest part of the delay. One of the big advantages to switching to a Tilemill/Kosmtik workflow is you could make changes and see them right away. We've lost that, and at 15-20 seconds we were actually slower than saving a change to Mapnik XML, restarting renderd, and hitting reload in the browser.

[...]

Issue #596, opened 2014-06-02
Render ref from route relationship


matthijsmelissen:
The following issue has been moved over from [trac](https://trac.openstreetmap.org/ticket/2610). It needs more discussion as it is not clear how this should be implemented.

Highway shields are only rendered when the ways themselves have a ref tag. Now, the different ways that form a road route are assembled into relations. For the sake of avoiding duplicate data the ref tags should be removed from the individual ways and applied to the relation only.

However, currently we only use ref tags on roads for rendering, and we ignore the ref tag on the relation.

Any idea how we could implement this?

pnorman:
Unassigning because I have no plans of working on this in the near or medium term, and I don't want others to not work on it because they think I am.

I will caution for anyone planning on picking this up that this has been a fairly long-running issue, reading it is somewhat difficult because

- a large number of people suggesting "easy" mechanisms which are not technically possible, so you need to know what will work
- sections are off-topic, talking about pictorial shields
- people assume that their local tagging practices apply world-wide

If you want to take a stab at implementing it, the requirements are
- it works without slim tables
- route relations being [split by osm2pgsql differently than ways](https://github.com/gravitystorm/openstreetmap-carto/issues/596#issuecomment-47410719) is handled
- the tagging practice of ref information on ways **and** the practice of ref information on route relations are **both** handled
  - ref information simultaneously on ways and relations is an open question
- When writing the query, performance and readability need to be considered

SomeoneElseOSM:
(just spotted this while looking at https://github.com/gravitystorm/openstreetmap-carto/issues/508#issuecomment-328315877 )

The comments above (admittedly some of which date from 2014) seem to assume an entirely SQL-based approach.  An alternative and frankly much less horrible option would be process the route relations in lua first (as per http://www.openstreetmap.org/user/SomeoneElse/diary/38988 ).  This could control as much or as little in the map display as needed (perhaps only shields).

pnorman:
> The comments above (admittedly some of which date from 2014) seem to assume an entirely SQL-based approach. An alternative and frankly much less horrible option would be process the route relations in lua first (as per http://www.openstreetmap.org/user/SomeoneElse/diary/38988 ). This could control as much or as little in the map display as needed (perhaps only shields).

The problem is because we need to handle shields from either relations or the ways, **and** we need to colour shields based on **way** information, not relation information we need to deal with some complexities that don't apply to what you're doing.

skylerbunny:
[...]

PR #2640, opened 2017-05-23
Render “surface” tag on roads with a pattern

### LINKED ISSUE #110: Render paved/unpaved

sommerluk:
Render “surface” tag on roads with a pattern

Closes #2621 (mutually exclusive with this PR)
Closes #3083 (mutually exclusive with this PR)

Resolves #110


sommerluk:
This code is still far from being complete. And there are many lines of code to write. And there are yet some questions. So I think it’s good to talk about these questions yet now.

Disclaimer: The pattern itself and its color are currently only an example. And, yes, highway=motorway is not the best example, but it was the first entry in roads.mss…

1.) I’ve attached the pattern to

```
#roads-low-zoom[zoom < 10],
.roads-fill[zoom >= 10],
.bridges-fill[zoom >= 10],
.tunnels-fill[zoom >= 10] {
```

and the consequence is that, where lines starts and stops, the round line caps of one line cover the pattern of the other line:

screenshot 1

So there are some parts that stay without pattern. I’m not sure if we can do something about. There seems to be not round line caps for patterns in Mapnik. And putting the surface in its own layer would reduce the problem, but is probably complicate code, and still leaves some artefacts if the two lines are not directly strait. Is attaching the pattern to the roads fill okay? Any ideas? (In general, I think this is not nice, but we have also other rendering artefacts in other situations, so this might not be a show-stopper.)

2.) Is the coding style okay? (Naming of files with leading 0 to allow automatic file order; type of selectors…)

3.) The code seems not to work for motorway_link

imagico:
> There seems to be not round line caps for patterns in Mapnik

If that is the case this could indeed be a deal breaker (and my mental list of things Mapnik can't do keeps growing...).

Theoretically there is of course still the option to _give up and use ST_Buffer()_ but i don't think we will want that due to code complexity and having all the line widths etc. in SQL.  Alternatively there is also ST_StartPoint()/ST_EndPoint() which could be used to separately draw the round line end caps using marker symbolizer SVGs.  But overall this is likely not any less complex.

Drawing all the line patterns above all the based colors could be worth trying - like you said this will lead to artefacts (both gaps and overlaps).

I have not looked at the code yet and i am probably also not the best one to evaluate that because i am not really that familiar with the roads code.

sommerluk:
Introducing after `::fill` another `::surface` layer works. I’ve tested it. Gaps are visible where roads stop (“noexit”). Gaps within the normal road framework seem to be rather seldom. Overlay on each and every intersetion.

sommerluk:
I think having a separate layer for the “unpaved” pattern has this behaviour on junctions:

- Imagine a junction between a paved secondary road and an unpaved unclassified road.

- In the rendering there will be the “unpaved” pattern above the junction.

So: What is the lesser evil?
1.) Rendering artefacts where highways are split?
2.) Unpaved pattern above paved junctions (with less rendering artefacts that at 1., but still with rendering artefacts)

imagico:
You could render the base fill color of paved roads above both the base color and the pattern of unpaved roads.  That would of course mean yet another level and more complexity.

In total you would have (from bottom to top, not considering bridges, tunnels etc.):

* casing for all roads
* base color lines for unpaved
* line pattern for unpaved
* base color lines for paved

dieterdreist:
2017-05-23 11:28 GMT+02:00 Lukas Sommer <notifications@github.com>:

> 1.) I’ve attached the pattern to



do you think it would be possible to subtract those dots rather than adding
them? Something like

`<Style name="eraser" comp-op="dst-out">` in carto? (adding it on the
style level)

sommerluk:
Here an example with the unpaved pattern in its own layer:

screenshot 2

When the way is continued, it looks good when the angle is not too extrem (violet). When the angle is very extreme, there are some artefacts (blue). When the way is split and there is some 90° angle, than there is a small part that will not be covered, and when it’s more than 90°, than this uncovered part might be bigger (green). Also at start and end points of lines there is a small part that is not covered (black). When ways are crossing, there is an overlay (orange), but it’s less extreme than expected.

Overall, I think this is not ideal, but at least better than attaching the pattern to the very same street layer.

sommerluk:
The idea with comp-op sounds nice. I’ll try that…

sommerluk:
Some documentation about comp-op:

https://tilemill-project.github.io/tilemill/docs/guides/comp-op/

and

https://github.com/mapnik/mapnik/wiki/Compositing

sommerluk:
@dieterdreist Could you give some details about your proposal? What (alpha) geometry has to be substracted?

pnorman:
Another layer won't work because it will fail with group-by and layering, e.g. a paved bridge over an unpaved road. You might be able to do something with attachments.

Besides the technical feasibility of pattern fills, we also have to look at if we can get good cartography from them. So far I don't like the examples I've seen, but how would they look on highway=secondary and below? Most unpaved roads will be some type of road which we render as white.

imagico:
I agree with Paul that now that we know the principal technical limitations of using line patterns it would probably be a good idea to test the visual suitability before putting too much work into finalizing the technical details.  Your current examples are good for showing the principal problems but practically we will need to see how problematic these are with the colors we would actually use.

Since you will probably end up using a differently colored pattern for every road base color you probably can also use an SVG with a solid color background which would avoid the overlaps although the gaps and missing line ends would still create problems.

sommerluk:
07

http://www.imagico.de/map/jsdotpattern.php#x,64,jdp59980;g,2.25,16,16;rx,250,2,8,8;rx,250,2,8,8;s,jdp58799;s,jdp49103;rx,250,2,8,8;rd,0,0,1,scree,0.1,5,10,0,jdp7587,a69900,ffffff;

sommerluk:
09

http://www.imagico.de/map/jsdotpattern.php#x,64,jdp59980;g,3.1,16,16;rx,250,2,8,8;rx,250,2,8,8;s,jdp58799;s,jdp49103;rx,250,2,8,8;rd,0,0,1,scree,0.2,5,10,0,jdp20478,3e3e3e,ffffff;

sommerluk:
10

http://www.imagico.de/map/jsdotpattern.php#x,64,jdp59980;g,2.25,16,16;rx,250,2,8,8;rx,250,2,8,8;s,jdp58799;s,jdp49103;rx,250,2,8,8;rd,0,0,1,scree,0.13,5,10,0,jdp95743,3e3e3e,ffffff;

sommerluk:
After some time being busy with other stuff, finally I could do some work on this issue. It’s less straightforward than I thought. The above screenshots show different patterns.

Lessons learned:
- It’s a good idea to start with white roads for developing the pattern, as suggested by @pnorman because that’s the most usual case.
- It’s not a good idea to use a “colourful colour” for the pattern. Using something like brown, that could give an association to dirt, does not work well; on white residential roads, the rendering of brown patterns makes the road more similar to the yellow secondary roads, what is in my opinion not useful. It might therefor be better to use a grayscale colour. As residential roads are white, this could be grey. For other road types, this could also be grey, but always darker than the road colour, not lighter (this is to have the same paradigm on each road type).
- On very thin roads, the pattern approach does not work well. I suppose this affects basically roads without casing, and maybe also roads with casing if they are still thin because the zoom level is still low. So probably this means that the “unpaved” pattern can be rendered starting from a given road width and would therefore have a different start zoom level for each road type. (Otherwise, we would have to make the zoom level depend on the smallest roads – service roads – which would limit the rendering to really high zoom levels).

Comments? Which of the above patterns might be the best to continue the work?

imagico:
I think these are mostly too strong, the pattern should be strong enough to be recognizable under most circumstances, at least for the wider roads, but not any stronger.  Remember this is a secondary characterization of roads behind the main classes.  It is better if in some cases the pattern is not recognizable than if it hides or obscures other information.  The 07 example seems fine in that regard but as you mentioned it should be color neutral for the white roads of course.

For the colored roads i would probably try making the pattern background slightly brighter than the paved color and the grain slightly darker.  This is not possible for the white roads of course.

Something like

http://www.imagico.de/map/jsdotpattern.php#x,64,jdp59980;g,2.25,16,16;rx,250,2,8,8;rx,250,2,8,8;s,jdp58799;s,jdp49103;rx,250,2,8,8;rx,250,2,8,8;s,jdp94534;rx,250,2,8,8;rd,0,0,1,scree,0.07,5,10,0,jdp40148,3e3e3e,ffffff;

might work for the white roads, maybe even with an even lighter color.

What a grain pattern will inevitably do is making color differences between the different road classes more difficult to distinguish.  In case of the white roads this mainly concerns `highway=living_street`.

Yes, the very thin roads will be a problem - though primarily at z<16 of course.  Also need to see if the access restrictions are still properly visible with the pattern.

dieterdreist:
> On 2. Jul 2017, at 14:31, Lukas Sommer <notifications@github.com> wrote:
> Comments? Which of the above patterns might be the best to continue the work?
> 

did you try subtracting with dst-out rather than adding the pattern?

sommerluk:
> did you try subtracting with dst-out rather than adding the pattern?

Yes, a little bit. At the first try, I didn’t understand how this works. I tried to create a new layer filled with the pattern and cut off what’s out of the road surface. It didn’t work, also because Mapnik cannot group layers.

After reading the documentation, I guess it could work this way:

- Render the unpaved roads layer with dst-out cutting a hole where the roads fill should be.
- Next, render a layer with the pattern (opaque) and use a layer rendering mode that pulls it _behind_ the existing content.
- To make this work for various road types, we need an own layer for each road type (or at least each road colour).

Is this what you had in mind? If so, I can try to make a working prototype, but this might take some time…

dieterdreist:
sent from a phone

> On 2. Jul 2017, at 19:04, Lukas Sommer <notifications@github.com> wrote:
> 
> After reading the documentation, I guess it could work this way:
> 
> Render the unpaved roads layer with dst-out cutting a hole where the roads fill should be.
> Next, render a layer with the pattern (opaque) and use a layer rendering mode that pulls it behind the existing content.
> To make this work for various road types, we need an own layer for each road type (or at least each road colour).
> Is this what you had in mind? If so, I can try to make a working prototype, but this might take some time…
> 


I thought it could work with just the first step: cut out (subtract ) the pattern from the road fill. It will only be visible over a background different than the fill color, but this should generally be a given. The basic idea is to work with the holes, not with another color that is added.

sommerluk:
11

http://www.imagico.de/map/jsdotpattern.php#x,64,jdp59980;g,2.25,16,16;rx,250,2,8,8;rx,250,2,8,8;s,jdp58799;s,jdp49103;rx,250,2,8,8;rx,250,2,8,8;s,jdp94534;rx,250,2,8,8;rd,0,0,1,scree,0.07,5,10,0,jdp40148,3e3e3e,ffffff; as proposed by @imagico in https://github.com/gravitystorm/openstreetmap-carto/pull/2640#issuecomment-312489794

sommerluk:
Pattern 11 proposed by @imagico looks good. Personally I would not go to a lighter tone, because yet right now it depends on the angle at which I’m looking to my screen if I can remark it or not.

A positive side-effect of pattern 11 (compared to the previous patterns) is that, as it’s lighter, the rendering artefacts of line-pattern are less visible.

sommerluk:
Pattern 11 with access restrictions:
screenshot 1

sommerluk:
> The basic idea is to work with the holes, not with another color that is added.

Okay, now I understand. This would make the street colour look different depending on the ground where they are rendered above. I’m not sure if that’s desirable (especially for wood).

Technically, dst-out cuts a “hole” (transparent part) in the layers below. I’m not sure if it’s possible to cut just the roads fill _and_ the casing (which is below), without cutting the background (residential, wood…) also.

sommerluk:
dst-out (mentioned by @dieterdreist) could maybe help with the line-pattern rendering artefacts.

dst-out ignores colours, it takes just the shape information (_where_ is _something_ rendered), and instead of rendering, it cuts a hole in all underlying layers (=it makes them transparent). In the final rendering, these parts show up with the colour #dddddd; which must be some sort of default.

We could
- render the road fill of unpaved roads with the dst-out option (leaving these parts transparent)
- add a new layer _above_ the road layer, which uses as data some fake data (polygons that cover the hole planet), render these polygons with the roads fill colour as an opaque colour and with the unpaved patters above, and then tell Mapnik to render it _behind_ the yet existing layers (which makes it appear at all transparent parts = unpaved roads).

I’ve done a quick-and-(very-)dirty hack, and it works.
- The advantage is that this is a solution without rendering artefacts.
- Maybe there’s a more elegant way to do this instead of using fake polygons.
- The downside is that (as far as I understand) we would need to have separate layers (!) for each road colour, so we would need separate SQL querries for each road colour. If so, the code will become _much_ more complex.
- The complex code changes might have undesired side effects.
- It will raise the maintaince burden.
- An intermediate solution might be to do this only for _one_ road colour (for example the white roads, which are most interesting regarding paved/unpaved), but use the normal ligne-pattern for the other colours. This might work without splitting the the roads SQL querry.

(Anyway, I’m new to the comp-op feature and so I’m not totally sure if everything above is correct…)

sommerluk:
Here some yellow roads (unpaved above “Carrfour Lubafrique”, paved below).

With the very same yellow used for paved and unpaved:
screenshot 1

With a lighter yellow for unpaved (making the overall lightness pretty much the same as for the paved road):
screenshot 2

sommerluk:
access=no now almost invisible when used on unpaved secondary roads (the north-south secondary road):
screenshot 3

imagico:
Is that the same pattern color as used for the white roads?  Using a more yellowish pattern color for yellow roads might improve visibility of the gray access dashing.  You could also try to add a very thin halo (either white or in the road background color) around the access dashes to separate them a bit from the patterned background.

sommerluk:
Indeed it is the same (grey) pattern as for the white roads. As the yellow is yet quite pale anyway, colour saturatin tuning will probably not make a big difference.

Also currently, access=no is barely visible for secondary roads: http://www.openstreetmap.org/#map=15/42.9154/13.0345 Fixing this is probably out of scope for this PR.

imagico:
> Also currently, access=no is barely visible for secondary roads

Yes, that is a separate problem but it underlines that the grain pattern for the colored roads should probably be colored in relation to the base color (reducing lightness) so to at least not make this worse.

sommerluk:
screenshot 7

Well, on higher road classifications (primary) we’ll need coloured patterns anyway.

sommerluk:
How to continue?

The current code is not actually working. It’s just a dirty hack.

Writing actually working code will be much lines of code.

Before doing this, there are a few things I would like to know in advance:

1.) In general, is the pattern solution something that has a chance to be merged here (alternative to #2621)?

If so, I could start the work. In this case, I think we have to decide about some technical questions:

2.) How to implement the pattern?

a) line-pattern: attaching the pattern to the layer with road fill

b) line-pattern: create new layers for the patterns

c) polygon-pattern: dst-out

I think that `a` is easier than `b`. The idea behind `b` was to give paved roads priority over unpaved roads, but this can be done also in a by the road ordering in the SQL.

`c` would essentially mean that either we have to use some fake objects within the roads SQL querry to render the pattern as background or otherwise split the road SQL querry into individual querries for each road type. I suppose that the former is not possible because we cannot get fake objects from another dataset (like from a shapefile) in the same querry as a normal OSM data SQL querry. And the latter I suppose would be very complex and bad for performance.

So in general I would take solution `a`…

imagico:
> 1.) In general, is the pattern solution something that has a chance to be merged here (alternative to #2621)?

I would say so far this is the most promising take on the subject.  With #2621 i currently do not see how this is going to work, both due to the technical difficulties and the fact that dashing for roads is already used for a lot of other things making it fairly non-intuitive.

However at this point i don't think we have seen sufficient samples based on a non-abstract pattern design suitable for production use for non-white roads to assess if the whole thing is going to look good design wise.

> a) line-pattern: attaching the pattern to the layer with road fill
> 
> b) line-pattern: create new layers for the patterns
> 
> c) polygon-pattern: dst-out

I have not really looked into the dst-out idea - did you show examples of advantages this would have?  Regarding (a) vs. (b) i currently see no real advantages for (b) and since we already have separate layers for tunnels and bridges this would mean a lot of additional layers if you'd duplicate everything into paved and unpaved.

But i think regarding

> The idea behind b was to give paved roads priority over unpaved roads, but this can be done also in a by the road ordering in the SQL.

that you would essentially need the same amount of MSS complexity for both (a) and (b).

Overall the logical steps for me would be:

1. Determine the best pattern to use for all zoom levels and colors.  You already have found something that works fairly well in https://github.com/gravitystorm/openstreetmap-carto/pull/2640#issuecomment-313868809 but it would need to be checked if this works equally well on the other zoom levels and for other road types.
2. Determine a set of pattern colors (and possibly base color tunings to balance the darker pattern) for the different road types, ideally to be integrated into `generate_road_colours.py`.
3. Write a script generating the necessary line pattern SVGs for all road types and road widths.
4. Implement rendering of this.
5. Sort out any problems with special cases like access restrictions, tunnels, construction etc.

This is both a lot of work in sheer volume and pretty challenging technically as well.  But i think it is doable (although for 4. there are others who can assess this better than i can).

Note that if you have doubts you can spare enough of your free time to complete this in a reasonable time frame you should be aware that given the importance of the subject, the interest in this shown by many in the past and the exploratory work you have already done it would probably not be too difficult to get some funding for implementing this.  I don't know if this makes a difference for you of course.

math1985:
I know this is a tough problem to solve, but to me the grain pattern is not really distinguishable. The difference with regular roads is too small. Macbook retina screen on 100% zoomlevel.

mboeringa:
> the interest in this shown by many in the past

Much of that interest was not so much in the micro-scale solution presented here, but in small scale (e.g. 1:50k-1:250k) visibility of unpaved status. The current proposed solution, will not solve this.

E.g. see this remark https://github.com/gravitystorm/openstreetmap-carto/issues/110#issuecomment-50495983

sommerluk:
@math1985 We could of course raise the contrast between the pattern colour and the road fill colour…

@mboeringa I’m aware of this limitation. The pattern approach will be difficult always when the line width is small – that affects low zoom levels. In the comment that you mentioned, there was a Brazilian map that was presented. Note that this map, while showing more differentiated the paved/unpaved status, it shows in general less information than openstreetmap-carto. So I doubt it will be possible to implement paved/unpaved rendering in this style similar to the Brazilian map. The pattern approach has indeed its limitations, but it’s better than no rendering at all 😉

@imagico Thanks for the comprehensive comment!

I understand the reason behind the order of work. On the other hand, it would make it much simpler to me to show screenshots and play with colours if I would have actually (at least partially) working code.

In the meantime I’ve taken a look to the colour scripts and think that it will be possible without big problems to colourize the patterns also by a script, just as you suggest.

About fundraising: Some money would probably not increase the time I could spend on this. But it might increase my motivation if someone would make a donation to a charity organization (https://www.fundacionamigo.org/colabora/) 😊.

Now, here comes a try to explain the dst-out code – in the hope that this will help to make a decision. FAQ:

**What’s the problem with using simply line patterns?**

- The (extreme) angles in lines and all start and finishing points of lines to not look like they should. See [this previous comment](https://github.com/gravitystorm/openstreetmap-carto/pull/2640#issuecomment-304503437) for details of the rendering problems with the line pattern approach.

**Can the dst-out version avoid these problems?**

- Yes, completely.

**How does the example dst-out code works?**

- You can take a look at the [dst-out example code](https://github.com/gravitystorm/openstreetmap-carto/compare/master...sommerluk:unpavedCompop01). This is a minimalistic code showcase. It doesn’t crash, it renders, but it’s incomplete (only for unclassified and residential roads) and it’s a hack, not a clean code.

- In the roads.mss file, for unclassified and residential roads, if they are unpaved, the dst-out option is used.

  - All the content that is already drawn (basic land and ocean colours, land-uses…) is melt into one single layer.
  - This layer is opaque, as with the land and ocean background, everything in our rendering canvas is covered.
  - Now, instead of actually drawing the unpaved residential/unclassified roads above, the dst-out option cuts a hole in the existing layer. Now, our rendering canvas has some transparent parts – everywhere, where an unpaved roads filling should be.

- In project.mml, after the SQL query for the roads, there is a new layer with the id _white-unpaved-background_. It queries polygons with the tag `fake=fake`.

- In my local database, I’ve created various of these “fake” polygons. Theoretically, I would have to cover the whole world, but I’ve created only some polygons for the regions where I actually to test rendering.

- In roads.mss file, finally the `fake=fake` geometries get rendered, but not like normally above the yet drawn content. Instead, it’s rendered _behind_ the yet drawn content (`dst-over`). As the yet drawn content has some transparent parts for the unpaved residential/unclassified roads, it will be visibly only there. As the fake geometry should theoretically cover the hole world, after this operation the whole map should be opaque again.

**What are the problems of the _dst-out/dst-over_ approach?**

- How to trigger the rendering of the road pattern background?

  - The example code uses _fake_ polygons in the database. This is of course nothing that we want to have in the OSM database, so we need something different that triggers the rendering of the road pattern background.

  - Is it possible in tell Mapnik to render a pattern simply for covering the hole planet surface (of course with _dst-over_), without any actual geometry? This would be nice, but I did not find out how to do this.

  - Otherwise, we could maybe create a static shape-file that contains polygons covering the hole planet?

- How to order correctly the rendering of the holes in the map (dst-out) and the corresponding backgrounds (dst-over)?

  - In the example code, there is only one road pattern colour implemented (white). But for real-world usage, we would need also the other colours (yellow, orange… for secondary, primary…).

  - Roads and backgrounds have to be drawn alternately. First white roads, then white background, than yellow roads, than yellow backgrounds, than orange roads, than orange backgrounds … and so on.

  - If we rely for the background rendering on a fake geometry in a shape file, then this geometry will be always in its own layer. So we would need five or six different SQL queries, one for each road colour, alternating with the fake shape file geometry. This might break the system of the _roads_ layer which also queries the railway lines and who’s complexity is beyond my knowledge.

  - I suppose there is no possibility in project.mml to mix an SQL query and a shape file as sources within the same layer/id, right? This would be great, because we could still use only one SQL querry, assign an order value to the shape file and this way render the background polygons in the right moment. But I fear this is technically not possible?

- Until now, openstreetmap-carto does not use `comp-op`. I don’t know if this type of operation could have side effects.

kocio-pl:
> Closes # hundred and ten.

Just being wonder - shouldn't it be simply "Resolves #110"?

sommerluk:
> Closes # hundred and ten.

Updated.

imagico:
The dst-out idea is clear to me now. I would regard the problem of the fake geometry a non-issue, you can easily generate a global bounding box out of thin air in PostGIS:

`ST_MakeEnvelope(-20037508.342789244, -20037508.342789244, 20037508.342789244, 20037508.342789244, 900913)`

It seems to me though from your description that you would still need an excessively large number of different layers to get proper layering or roads - unless i misunderstood something about the whole procedure.

kocio-pl:
Was there something wrong with this code?

sommerluk:
> Was there something wrong with this code?
Quite the contrary!

The ideo of @imagico (“global bounding box out of thin air”) is pretty ingenious! This type of global bounding box can be added within the existing roads-fill SQL query, so there is no need to split up the SQL query – that’s great! In the roads.mss there must be created “attachments” (like `::fill`) to order everthing correctly (otherwise the comp-op operations are not applied in the correct order).

Therefore I think this is the way to go. I’ve updated the code of this PR to show a proof of concept. (In the moment when I did undo the old changes, Github closed this issue.) Reopening.

sommerluk:
wz4jsopw

kocio-pl:
I like this effect, looks quite subtle for me, which is good. But I don't see why [this road](http://www.openstreetmap.org/way/30628320) is rendered as unpaved, I guess, since it's just simple residential road and what is white dashing and dots on [this road](http://www.openstreetmap.org/way/30611927#map=18/5.35299/-4.10074). Is this tagging forged somehow?

sommerluk:
> Is this tagging forged somehow?

Yes. In my local database, I have a data extract from Abidjan (Ivory Coast), and I’ve changed locally the tagging of most roads to “unpaved” (some one roads with access restictions) with the purpose for better testing for this PR.

What the screenshot is intented to show:
- It is possible with this coding approach to get a rendering without artefacts.
- Where OSM lines end and start again does not affect the rendering.
- The road priority (secondary above tertiary, tertiary above residential…) is working correctly and does not make rendering artefacts for the pattern.
- Even where the white and the yellow pattern touch each other, the transition is smooth, because it’s the same (globally aligned) pattern; just the colour changes.

@imagico Instead of rendering a pattern with transparent background together with an opaque `polygon-fill` colour, I could also create a pattern that includes directly an opaque background colour (and so not use `polygon-fill` at all). Knowing about #2750 what would be the better approach?

kocio-pl:
> In the moment when I did undo the old changes, Github closed this issue. 

I use `git push -f`and it always works as expected.

sommerluk:
Well, had discarded all the previous commits on this PR. Then, I did “git push -f…” before later adding the commits for the new code. So during a small moment, there was no commit in the github (remote) repository; apparently because of this, github did close the PR automatically.

imagico:
@sommerluk - my understanding of the problem in #2750 is rather limited so i cannot really tell, in particular i don't know why patterns in the landcover layer and in landcover-area-symbols behave differently.  But as i understand it you could also use a PNG pattern with this approach so you would have a working fallback in case of problems.  There is a good chance though that as long as you use default polygon-gamma you are fine (but that is a fairly wild guess i have no proof for).

If you want to proceed with this i would suggest testing early on if there are problems with either highway areas (like service, pedestrian) or bridges.

Maybe @pnorman knows if there are constraints resulting from this in other areas - like when using comp-op in other places in the style as well.

symbols/generating_patterns/unpaved.md
```diff
@@ -0,0 +1,11 @@
...
+
+In the first step visit [jsdotpattern (command sequence is recorded)](www.imagico.de/map/jsdotpattern.php#x,64,jdp59980;g,2.25,16,16;rx,250,2,8,8;rx,250,2,8,8;s,jdp58799;s,jdp49103;rx,250,2,8,8;rx,250,2,8,8;s,jdp94534;rx,250,2,8,8;rd,0,0,1,scree,0.07,5,10,0,jdp40148,0000ff,ffffff;)
+
+Use "get pattern SVG data" (it is expected that background will be transparent, background colour in jsdotpattern is only for display).
+
+Generated SVG image is sanitized for use with Mapnik by the script svg_pattern.sh from the jsdotpattern repository at http://github.com/imagico/jsdotpattern.
+
+colour (white): 3e3e3e
+
+The final file is beach.svg
```
kocio-pl: Are you sure about the name? :smiley:

sommerluk:
> If you want to proceed with this i would suggest testing early on if there are problems with either highway areas (like service, pedestrian) or bridges.

Thanks for the suggestion. I have looked into this, and indeed rendering order might be a problem. 

The fact of different layers (on project.mml and roads.mss) for highway areas itself is not a problem. (And though I didn’t test it for bridges, for bridges it should work also.):
screenshot 4

There is however another problem: Support for the “layer” tag in the OSM data! I will explain:

- To get the rendering correctly, we need separate attachments (in roads.mss). These attachments follow the z_order in the lua file. We do not need an own attachment for each z_order value, but we need a new one each time when in the z_order the road colour changes. Mapnik gives a higher priority to the attachment order than to the order of the data itself (SQL order by). So we can make for example an attachment that groups all white roads together. Within this attachment, the data order (SQL order by) is applied without problems. After this attachment, we have the patter background. And than yellow roads, and so on…

- The data order (SQL order by) for roads-fill has various criterias:
  - layernotnull,
  - z_order,
  - CASE WHEN access IN ('no', 'private') THEN 0 WHEN access IN ('destination') THEN 1 ELSE 2 END,
  - CASE WHEN int_surface IN ('unpaved') THEN 0 ELSE 2 END

- As explained, the attachment order in roads.mss has to be syncronized with z_order. Every order criteria _after_ z_order will work out-of-the-box. But there is one criteria _before_ it: The “layer” tag from Openstreetmap.

Of course I could duplicate the `::fill` code. For example we could support only the layers from −5 to +5 (of course this would mean to duplicate the `::fill` code 11 times). Maybe we could even duplicate the code with a small script, similar to the script-generated road colours. But this has a certain complexity. Otherwise the “layer” tag would not work (this would probably be most visible for bridges that go over other bridges and where a bad rendering order would not give us a nice and correct map).

Comments?

Apart from that, I have written a small script to generate colours of the unpaved pattern and also the unpaved background now with a small script (like the road colours):

screenshot 1
screenshot 3

imagico:
Ok, i think i understand the problem.  The attachments have priority over the order in SQL, that is clear and well known.  And bridges and tunnels are separate layers in project.mml anyway so a bridge over a non-bridge road will always render in order no matter what attachments you have in mss and what layer tags are used.  

It is my understanding that the practical need for layer ordering on roads is primarily for multi-level bridge crossings (unless i missed something important here).  You could also use the layer tag as a way to override the z_order but that would be abuse of the tag.  Layer is for ordering non-connecting ways and since the layer tag ordering happens for casing and fill separately overlapping road lines which are not connected and which have a different layer will still appear connected because all casings are drawn before all fills (see [here](http://www.openstreetmap.org/way/117073726#map=17/41.91109/12.47256) for example). What you say you could do to deal with that for paved/unpaved in the same way as it is now is to duplicate the whole mss code for every layer value you want to consider.  That seems overkill to me (although it would also allow to correctly draw the casing in the above case i suppose).  One very simple alternative way to deal with that would be to say we only render unpaved roads in a special design when they are without layer or layer=0.

To put this into perspective:  There are 2.7 million highway features with a layer tag but only about 100k of those have a surface tag that amounts to unpaved.  Most of these are going to be bridges or tunnels which in my eyes do not really need a dedicated unpaved rendering (and for tunnels you would need yet another set of patterns anyway which would be a pain to produce and maintain).

Side note: The idea of script generating mss code or in other words: Having a preprocessing layer above the mss level to allow more flexible logic that is not currently available in mss is something that has crossed my mind as well in the past but having a script generating mss code that is interpreted by carto to generate mapnik xml is kind of screwed.

sommerluk:
Indeed “unpaved” seems to be less important to render on bridges, I agree. (Nevertheless nice to have if possible, because the fill colour of bridges is the same as for normal roads, so not having the pattern might be suprising).

As you have spoken about brides, I had a closer look to briges, and the _casing_ of bridges with a higher “layer” tag is currently always rendered above the _fill_ of bridges with a lower “layer” tag – although casing and fill are in different attachments. I do not really understand why this works, but it works. And know comes the great news: It also works with dst-out!

w4kq48dg

The code renders an unpaved pattern for roads-fill and bridges-fill, but not for tunnels. As both, roads and bridges, use the same mss code, support for unpaved pattern on bridges comes for free – and is welcome. I agree that indeed for tunnels, an unpaved patters is indeet not so important, and also technically it might not be trivial to have a second pattern. As tunnels have (unlike bridges) a different fill colour anyway, this would also not be suprising for the map user.

The mss code does _not_ need to be duplicated for each layer. That’s also a really good news. We only need an own globalboundingbox geometry for each layer, but that is really not much code. Limiting to −5…5 seems (after reading the wiki and looking at taginfo) anyway a good idea.

Still no code for turning circles. Maybe #2558 could be solved at the same time…

Apart from that, the code is now quite advanced and complete.

a4t9irie

imagico:
> As you have spoken about brides, I had a closer look to briges, and the casing of bridges with a higher “layer” tag is currently always rendered above the fill of bridges with a lower “layer” tag – although casing and fill are in different attachments. I do not really understand why this works, but it works

That is likely because the layer has a group-by property in https://github.com/gravitystorm/openstreetmap-carto/blob/b7db4764be0e9822b3c76f1442f8143beb0694cd/project.mml#L1047 - see also https://github.com/gravitystorm/openstreetmap-carto/issues/1975#issuecomment-160382065.

The diff on roads.mss is no more readable for me now so i have kind of lost my understanding of your modifications overall.  Your duplication of the global bounding box code in project.mml could be avoided (using `generate_series()` and a join) but it might be easier to read this way.

Some of the color choices seem strange to me - while the base color for unpaved pedestrian and aeroway seem visibly lighter there does not really seem to be a visible difference in case of the colored roads.  For example:

http://davidjohnstone.net/pages/lch-lab-colour-gradient-picker#ebebf6,dddde8
http://davidjohnstone.net/pages/lch-lab-colour-gradient-picker#ffb7a1,f9b29c

I think the generation of the colored SVGs should be scripted, if any road color change requires manually editing SVG files that would be a pain.

sommerluk:
_About the colour choise_

Earlier in this thread, you suggested to use a lighter base colour to compensate the darker pattern and therefor get the same overall-lightness like for the paved roads. When defining the colour range for the script that generates the pattern colours, I’ve choosen the same hue and the same chroma. Only the lightness varies. Apparently there is not much room especially for the yellow secondary roads to get a lighter colour while having the same hue and chroma. So not only for white colours, but also for other, light colours like yellow, the pattern and its background will always be darker than the normel “paved” colour. All the other colours that are not genereated by the roads script, I’ve found by trial and error – always starting with the base colour used in roads.mss.

I do not know much about colour mathematics. What could we do? Maybe we could just accept that we cannot reliably produce a pattern with the same overall lightness for all road types, and then consecuently make all pattern uniformly slightly darker than the normal road colour. At least it would be uniform…

_About what’s happening in the code_

The ::fill attachment is split up into various attachments, one for each different road colour ordered by z_order. The actual code for each road type is the same as before – only a selector for the unpaved roads is added to each attachment. Each road attachment is followed by an attachment with the corresponding unpaved background pattern.

Apart from that, there are some minor changes to the SQL querry (a defined order between “highway=road” and the white roads to synchronize with the attachment order…) or in roads.mss (colour variable definations).

It’s just more code know because all road types are supported.

_About the technical comments_
I will update the code of this PR in the next days, taking this into account…

imagico:
> Earlier in this thread, you suggested to use a lighter base colour to compensate the darker pattern and therefor get the same overall-lightness like for the paved roads.

Yes, but this is not happening consistently in your examples - while this is the case for aeroways and pedestrian the unpaved colored road classes are all distinctly darker than the paved variants which is irritating for the map users (they might have the impression the grain pattern means something different in both cases).

You are probably running into out-of-gamut problems in some cases but this should be manageable in a reasonable way.  I can try looking into this but it might take some time until i find the time to do so.

sommerluk:
Fix for the SQL layers code. This might need some explication about the interaction between the code in roads.mss for the roads fill colours on one side and the SQL querries on the other side. In order to get the pattern right, the code for the roads fill colours had been split up to various attachments (which takes precedence over the SQL data order). Now let’s take a look at the SQL querries:
- `.tunnels-fill` is the easiest part. As this PR does not provide a pattern for unpaved tunnels, and on the other hand in project.mml we have the property `group-by: layernotnull` anyway, there is nothing special we have to do in the SQL querry and there will be no rendering change because of the roads.mss split-up in attachments.
- `.roads-fill` does not have `group-by: layernotnull`. So with the split up in roads.mss into attachments following z_order, now the [example](http://www.openstreetmap.org/way/117073726#map=17/41.91109/12.47256) from @imagico would render slightly different: The fill of this primary road would render above the fill of the secondary road aside. As @imagico explained, the wiki defines the `layer` tag to order _crossing_ non-connected ways, so I think we could live with this rendering change. (Otherwise, we could use the same system as for bridges, but I’m not sure it’s worth the pain.)
- `#roads-low-zoom` is in the same situation as `.roads-fill`, but I think at this zoom level, given that the bridges are really small at this zoom level, it would be clearly a really good thing not to order by the layer tag.
- `briges-fill` has the property `group-by: layernotnull` anyway. That means there must be a globalboundingbox for each layer. Therefor I’ve reduced the number of supported layers. The wiki says 
_“Values between -5 to 5 are used, 0 should not be used explicitly”_ and indeed looking at taginfo, everything beyond this range seems to be a little bit strange (there are more objects with layer “8” than with “7”, which is strange). There are however quite a few cases with layer “0”, maybe used for bridges that pass over ditches with layer “-1”? But negative values for bridges seems also unlikely. So currently there is code for the layers from 0 to 5. Every layer below 0 is treated like 0; every layer above 5 is treated like 5. But of course we could also make a different choise about this range.

Note: `highway-area-fill` is not affected by all this, because it’s a polygon geometry than can use polygon patterns, so no problem at all here.

pnorman:
> i have kind of lost my understanding of your modifications overall

I'm finding it the same way.

I really don't like generating fake polygons. I'm also opposed to anything that special-cases -5 to 5 or 0 to 5. We used to have that before group-by existed, and moving to group-by was an improvement.

Something I thought of which hasn't been considered is dasharrays like we used to use for oneway arrows. A suitable combinations of offset lines with dasharrays could produce a noise pattern.

imagico:
> I'm also opposed to anything that special-cases -5 to 5 or 0 to 5

That could likely be avoided by using `generate_series()` and a join (well - not fully - but you could easily make that -100 to 100).  You could probably even fully avoid it with a `SELECT DISTINCT layernotnull FROM ...` subquery and join - this might actually be the better approach although it is probably a bit convoluted.

sommerluk:
The pattern that rely on `road-colors.yaml` are now directly generated by the script.

imagico:
I looked into the color adjustments for the pattern and came up with this:

road colors

This is based on a small tool i quickly hacked together based on lcms.  Since lcms has no python bindings this is in C: http://www.imagico.de/files/colors_bd.c.

With `echo "#fcd6a4" | colors_bd -40 2.6` you for example get:

`#fcd6a4,#8b6c3f,#ffddab,88.02,48.02,90.62`

Which is the primary road color, as is and darkened/brightened by -40/2.6 and the corresponding lightness values.  The full list of colors with these parameters is here:

```
motorway-low-zoom #e66e89,#72002b,#ee7590,61.91,21.91,64.51
trunk-low-zoom #f5977a,#80311c,#fd9e81,72.01,32.01,74.61
primary-low-zoom #f3c380,#815b1d,#fbca87,82.04,42.04,84.64
secondary-low-zoom #e8eda0,#7a803a,#f0f4a7,92.04,52.04,94.64
motorway-fill #e892a2,#772c3f,#f099a9,70.00,30.00,72.60
trunk-fill #f9b29c,#874b39,#ffb9a3,79.05,39.05,81.65
primary-fill #fcd6a4,#8b6c3f,#ffddab,88.02,48.02,90.62
secondary-fill #f7fabf,#888c57,#ffffc6,96.98,56.98,99.58
platform-fill #bbbbbb,#545454,#c2c2c2,75.88,35.88,78.48
aeroway-fill #bbbbcc,#545463,#c2c2d3,76.28,36.28,78.88
road-fill #dddddd,#727272,#e4e4e4,88.11,48.11,90.71
pedestrian-fill #dddde8,#72727c,#e4e4ef,88.36,48.36,90.96
living-street-fill #ededed,#808080,#f4f4f4,93.75,53.75,96.35
raceway-fill #ffc0cb,#8e5762,#ffc7d2,83.79,43.79,86.39
track-fill #cdbea0,#63573d,#d4c5a7,77.65,37.65,80.25
residential-fill #ffffff,#919191,#ffffff,100.00,60.00,102.60
```

You can change the lightness differences of course but if the darkened version sets you at lightness below zero this will not balance of course.  If the pattern appears to have the same overall brightness as the plain color also depends on the gamma curve of your monitor of course.

You can try if this also works in python using colormath if you want to - i am just more familiar with lcms so i used this for the proof of concept.

Here is the same in somewhat stronger (with -60/3.9 lightness adjustment):

road colors

```
motorway-low-zoom #e66e89,#3d0000,#f27993,61.91,1.91,65.81
trunk-low-zoom #f5977a,#4b0000,#ffa284,72.01,12.01,75.91
primary-low-zoom #f3c380,#4d2d00,#ffce8a,82.04,22.04,85.94
secondary-low-zoom #e8eda0,#484f07,#f3f8ab,92.04,32.04,95.94
motorway-fill #e892a2,#430015,#f39dac,70.00,10.00,73.90
trunk-fill #f9b29c,#521d0f,#ffbda6,79.05,19.05,82.95
primary-fill #fcd6a4,#573d12,#ffe1af,88.02,28.02,91.92
secondary-fill #f7fabf,#565a28,#ffffc7,96.98,36.98,100.88
platform-fill #bbbbbb,#282828,#c6c6c6,75.88,15.88,79.78
aeroway-fill #bbbbcc,#272835,#c6c6d7,76.28,16.28,80.18
road-fill #dddddd,#424242,#e8e8e8,88.11,28.11,92.01
pedestrian-fill #dddde8,#42424b,#e8e8f3,88.36,28.36,92.26
living-street-fill #ededed,#4f4f4f,#f8f8f8,93.75,33.75,97.65
raceway-fill #ffc0cb,#5a2834,#ffcbd6,83.79,23.79,87.69
track-fill #cdbea0,#342a12,#d8c9aa,77.65,17.65,81.55
residential-fill #ffffff,#5e5e5e,#ffffff,100.00,40.00,103.90
```

sommerluk:
@imagico Thanks for the work with the perceptual colour space. This is something completely beyond my knowledge, so I’m glad that you could help here. It looks much more balanced than my attempt.

The new colours are now in the code, and there is also a [preview of the new colours](https://user-images.githubusercontent.com/6830724/30327412-772ebf4a-97bb-11e7-9c7e-8acc4259405b.png).

sommerluk:
@pnorman Do you mean something like this?

```
        [int_surface = 'unpaved'] {
          b/line-dasharray: 0.9,3,1,2,1,1,1,2,;
          b/line-offset: 3;
          c/line-dasharray: 1,3,1,2.2,0.9,1,1,3,1,2;
          c/line-offset: 1;
          d/line-dasharray: 1,3,1.1,2.4,1,1,;
          d/line-offset: -1;
          e/line-dasharray: 1,3.1,1,2.1,;
          e/line-offset: -3;
        }
```
screenshot 1
screenshot 2
screenshot 3

For extreme angles in roads, it works quite good (better than the line patterns). Same problem as with the line pattern when a way ends (line-caps). More difficult to predict the pattern as on each node with an angle the offset relative between the dash lines changed. Not sure if we could get something that looks really as randomized as with the polygon pattern. I haven’t looked into real-world-code and how complex the code would become. Obviously it avoids all `group-by` related issues.

imagico:
The colors look all right now from my point of view.  It would be good of course if all of this would be generated by script so you can change the colors without tons of tedious hand work.

Did you get the idea of a `SELECT DISTINCT layernotnull FROM ...` subquery and join?

This could probably be done the following way:

```
        (SELECT
            way,
            ...
          FROM ( -- begin "features" subselect that contains both roads and rail/aero
            WITH features_real AS ( -- this is the actual data (which is used both as is and to determine the distinct layer values present in it)
              SELECT
                  way,
                  ('highway_' || highway) AS feature, -- only motorway to tertiary links are accepted later on
                  ...
                FROM planet_osm_line
                WHERE ...
                  AND way && !bbox! -- end of road select
              UNION ALL
              SELECT
                  way,
                  ... AS feature,
                  ...
                FROM planet_osm_line
                WHERE ...
                  AND way && !bbox! -- end of rail/aero select
              )
            SELECT
                way,
                feature,
                ...
                layernotnull
                ...
              FROM features_real
            UNION ALL
            SELECT -- this generates a global bounding box feature for each layer value present in the data
                gbb.way AS way,
                gbb.feature AS feature,
                ...
                layers_list.layernotnull AS layernotnull,
                ...
              FROM
                (SELECT DISTINCT layernotnull FROM features_real) AS layers_list
                CROSS JOIN 
                (SELECT 
                    ST_MakeEnvelope(-20037508.342789244, -20037508.342789244, 20037508.342789244, 20037508.342789244, 900913) AS way, 
                    'globalboundingbox' AS feature
                ) AS gbb
            ) AS features
          ORDER BY
            ...
        );
```

sommerluk:
Well, honestly I didn’t get it. I have no experience with SQL. So this was an opportunity for me to read some tutorials about SQL…

The query you have proposed is now available in an alternative branch here: https://github.com/sommerluk/openstreetmap-carto/commit/fd50b59d7baa387f92b5666ad81da67e207660b4

However, it does not work. I get an SQL error.

A minimalistic example which still produces the same SQL error is this one:
```yaml
  - id: testcase
    geometry: linestring
    <<: *extents
    Datasource:
      <<: *osm2pgsql
      table: |-
        (
          WITH features_real AS (SELECT way FROM planet_osm_line)
          SELECT
            way
          FROM features_real
        ) AS testcase
    properties:
      minzoom: 10
```
in `project.mml`.

And
```mss
#testcase {
    [zoom >= 10] {
      line-width: 2;
      line-color: red;
    }
}
```
in an `.mss` file.

It creates this error message:
```
Postgis Plugin: FEHLER:  Relation „features_real“ existiert nicht
LINE 1: SELECT ST_SRID("way") AS srid FROM features_real WHERE "way"...
                                           ^
in executeQuery Full sql was: 'SELECT ST_SRID("way") AS srid FROM features_real WHERE "way" IS NOT NULL LIMIT 1;'
  encountered during parsing of layer 'testcase' in Layer
```

Some observations:
- There is an error message even yet at zoom level 4, in spite of that the layer is only active starting with zoom level 10.
- When changing the `project.mml` code to use `geometry: point` instead of `geometry: linestring` than there is no error.
- When not putting `way` in the query, but only fields like `name`, than there is no error (at least up to the zoom level 9, while the layer is not active).
- The code in the error message seems to be different from the code in `project.mml`, so maybe somewhere in the software stack, there is a special behaviour for `way` implemented, and this special behaviour is incompatible with the `WITH` clause.

It seems possible to do this without a `WITH` clause, see https://github.com/sommerluk/openstreetmap-carto/commit/1c5b94459f9330c05a4426b38e11c55c72d3d2e2 for such code. It is of course less elegant than _with_ a `WITH` clause. Not sure if performance makes a difference. At the same time, I’ve limited the layers in the `WHERE` clause to catch only actual roads that are actually unpaved.

You did add `AND way && !bbox!` in the code you have proposed. What is this doing exactly?

And, somewhat out of topic, but it would be interisting for me to understand the special treatment of `way`: Why is the order for `turning-circle-fill` currently `ORDER BY p.way, v.prio`. Isn’t `way` just some numbers that describe the geometry? Why does it makes sense to order by `way`?

math1985:
The query looks fine to me, do you see what we are missing @pnorman ?

What happens if you change the query as follows?

```
        (
          SELECT
            way
          FROM (SELECT way FROM planet_osm_line)
        ) AS testcase
```

> You did add AND way && !bbox! in the code you have proposed. What is this doing exactly?

I'm not sure what checking for `way` accomplishes, but `!bbox!` only fetches the features within the current bounding box (i.e. within the (meta)tile that is currently being generated). Mapnik replaces this with an actual condition.

> And, somewhat out of topic, but it would be interisting for me to understand the special treatment of way: Why is the order for turning-circle-fill currently ORDER BY p.way, v.prio. Isn’t way just some numbers that describe the geometry? Why does it makes sense to order by way?

It forces at least *some* deterministic order. This is useful for making sure order across tileservers and across metatiles is always the same. Sorting first by way and then by prio does not makes sense as far as I understand, because we're unlikely to have two objects with exactly the same way.

pnorman:
> @pnorman Do you mean something like this?

Yes. 

> For extreme angles in roads, it works quite good (better than the line patterns). Same problem as with the line pattern when a way ends (line-caps). More difficult to predict the pattern as on each node with an angle the offset relative between the dash lines changed. Not sure if we could get something that looks really as randomized as with the polygon pattern.

We can get something as random, but it might not *look* as good. From the forest patterns and other work, we know that some form of adjustment for consistent density helps.

The big problem for me with either method is I don't like the cartography much.

-----

> It creates this error message:

It looks like Mapnik is trying to run a query on what it things is the table the geometries are coming from. It's a bug in Mapnik, probably unreported. To work around it, set the srid [for the datasource](https://github.com/mapnik/mapnik/wiki/PostGIS)

-----

> It forces at least some deterministic order. This is useful for making sure order across tileservers and across metatiles is always the same. Sorting first by way and then by prio does not makes sense as far as I understand, because we're unlikely to have two objects with exactly the same way.

No. It orders by way because it's part of a `SELECT DISTINCT ON (p.way)` and the [`DISTINCT ON` expression must match the leftmost ORDER BY expression](https://www.postgresql.org/docs/current/static/sql-select.html#SQL-DISTINCT)

------

Given the direction the SQL is going in, I think this method will make our roads code more of a mess that everyone is afraid to touch

imagico:
> It seems possible to do this without a WITH clause, see sommerluk/openstreetmap-carto@1c5b944 for such code. It is of course less elegant than with a WITH clause. Not sure if performance makes a difference. At the same time, I’ve limited the layers in the WHERE clause to catch only actual roads that are actually unpaved.

Yes, that makes sense.  The advantage of the CTE is that it avoids an additional query on the database but that does not necessarily pose much of a problem.

I don't think you need the two nested queries though - you could direct make a `DISTINCT ON` query on planet_osm_line.

> You did add AND way && !bbox! in the code you have proposed. What is this doing exactly?

This is meant to manually define the query area in case mapnik is not able to do this with such a query - see https://github.com/mapnik/mapnik/wiki/PostGIS#bbox-token

I was not sure about this but found an example in https://github.com/gravitystorm/openstreetmap-carto/issues/596#issuecomment-56258422 where this is done.

As far as the pattern is concerned - i think for the high zoom levels/relatively wide roads the 2D pattern works quite well and is probably the best you can hope for in terms of intuitive rendering of unpaved roads without completely reworking the whole road cartography concept - but that is my personal view on cartography of course.  For the thinner roads and lower zoom levels this is more difficult.  The structured dashing idea might be useful here because a directed structure can work better on a narrow line than a completely random 2d pattern.  But i am not sure this is workable.

By own opinion as it is would be to use the 2D pattern method down to the line widths where it ceases to make sense and simply not differentiate roads by surface beyond that.  In the current design concept in this style the functional road classification is the primary distinction of road rendering and you need to stop differentiating additional properties at coarse scales at some point unless you want - as said - to rework the whole road cartography concept (which so far no one has suggested).

> Given the direction the SQL is going in, I think this method will make our roads code more of a mess that everyone is afraid to touch.

Well - i don't think to fret about the mess will make the mess more accessible.  In the end it will only make everyone more afraid to touch it thereby further aggravating the problem.

I am pretty sure @sommerluk would be fine with implementing his change in a different, less messy road rendering framework but telling him he has to implement this himself before he can consider implementing unpaved roads is kind of adding yet another hurdle on top of the _mess everyone is afraid to touch_.

By the way the complexity of this PR ultimately stems from mapnik's inability to directly render lines with a 2d pattern fill.  For me this is kind of a deja vu - see #1853.

sommerluk:
During the development of this PR, I’ve looked into various solutions for paved/unpaved rendering, and while I have looked into some solutions more in detail than into others, but there is one thing I’m certain: Implementing paved/unpaved rendering demands a price. _Every_ of the so-far discussed approachs demands a price. I’ve started this PR because I think it’s worth the price if we get paved/unpaved rendering. Anyway the different approaches have each one different advantages and disadvantages. From my perspective:

| Approach        | Intuitive cartography           | Rendering without artefacts | Trivial SQL  | Trivial MSS | Same approach for both roads-fill and highway-area-fill
| ------------- |:-------------:|:-------------:|:-------------:|:-------------:|:-------------:|
| Dashed line (like for seasonal rivers)      | ✗ <br/> (Conflict with _seasonal_ and _construction_.) | ✗ <br/> (Dashed lines have the problem with the coastline paradoxon. The cartography is more noisy.) | ✓ | ✓ | ✗ |
| line-pattern     | ✓  | ✗ <br/> (Artefacts on way joints and way crossings and extreme angles.) | ✓ | ✗ <br/> (We need hundreds of individual line patterns: One for each combination of highway type and fill width. As the are referenced individually by file name, we cannot use the existing road-width variables in the MSS file. The code has to be written manually for each colour/width combination. Changes to the road width in the future will require rewrite this code.) |  ✓ |
| dasharray     | ✓  | ✗ <br/> (Artefacts on way joints and way crossings.) | ✓ | ✗ <br/> (It will be quite complex and verbose MSS code to cover the many colour/width combinations; each one at its correct zoom level. Changes to the road width in the future will require rewrite this code.) |  ✗ |
| polygon-pattern with dst-out (this PR)  | ✓  | ✓  | ✗ <br/> (The query will be more complex than the current one.) | ✓ <br/> (Yes, there’s dst-out with its interdependence with the attachment order. On the other hand, the code is still readable. And changes to the road width in the future will _not_ require rewrite this code – they are applied automatically. Also the pattern generation is more comprehensible than with the other approaches.) |  ✓ |

Obviously the “_dashed line_” solution (#2621) is the one with less maintenance cost. But it demands a higher price in cartography.

From the other solutions, each one has a higher maintenance cost than the “_dashed line_” solution, but _dst-out_ is in my opinion the one that combines not-so-much maintenance cost with an intuitive cartography.

sommerluk:
Suppose it’s better to wait for #2869 …

pnorman:
> Suppose it’s better to wait for #2869 …

It should make doing anything roads related easier because there will be less duplicate code.

kocio-pl:
@sommerluk There are no news about #2869 possible progress and @pnorman is rarely seen here for some time, so maybe going along with the current state would be safer. What do you think about it?

kocio-pl:
We have now also new attempt (#3083) and a few more propositions after https://github.com/gravitystorm/openstreetmap-carto/issues/110#issuecomment-366739539. What is your plan with this PR? Do you want to continue with this approach?

sommerluk:
After taking a look at #3083 and the comments in #110 I’m still convinced that the approach of _this_ PR (surface rendering by pattern) is better than dashed lines or the propositions. The reasons have yet been discussed above.

I also think that this PR is cartographically quite advanced, so there will not be much more cartographic changes (maybe some technical tuning). So a cartographic review can yet be made now.

Technically, there are too many merge conflicts. I’ve created https://github.com/sommerluk/openstreetmap-carto/tree/unpavedpre which does only code reordering in the road layer without any rendering changes against the current master. If there is interest, I can continue to re-implement the other parts of this PR. (By separating the simple reordering and the real functional changes, it will make the technical review easier.) Currently, I’m quiete busy, but if there is interest in merging that, I could try to do the necessary work…

imagico:
Independent of this specific change and adding it to the style - i think it would be nice if you could find the time to write up the underlying technical idea in a generic form.  Obviously this is a workaround for a Mapnik limitation (the inability to use a 2d pattern as a line fill).  But since it is unlikely this is going to be added to Mapnik this is potentially of broader interest beyond the specific application here.

kocio-pl:
Why do you think it's unlikely - did you contact them? I wouldn't assume anything without conversation with Mapnik developers. If this may be of interest for a broader audience, it's better to try to implement it on as low level as possible, without workarounds.

imagico:
Mapbox has no significant interest in the rendering side of Mapnik, they are not using it in production any more.  AGG (the rendering engine Mapnik is built on) ceased active development more then ten years ago so it is getting increasingly unlikely that independent developers will become interested in working with that.  That means major feature additions on the rendering side of Mapnik are unlikely.

kocio-pl:
I see. I still feel it's better to nut assume things and just ask - I don't know the plans of current developers, but they are still working on some features.

sommerluk:
Rebased (with working SQL querry). Still lacks better description for pattern generating process.

From now on, I’ll try that this PR contains two commits.
1. The first one in just re-ordering of existing code (and usage of a variable) – this commit should not lead to any rendering changes.
2. The second commit contains the real changes.

So when reviewing, you could look at the second commit only, which is easier to read…

HolgerJeromin:
> The first one is just re-ordering of existing code (and usage of a variable) 

Perhaps this is a value in itself and could/should be merged.

kocio-pl:
Road code is a big thing that I'm not capable of reviewing. There's also competing solution, although it's not too mature. I can only make some visual tests.

kocio-pl:
I have talked with @matkoniecz directly and I hope he will do the review. i will try to make some test rendering.

kocio-pl:
I think the pattern is very pale - it's hard for me to see the junction between paved and unpaved even when it should be obvious:

[Tertiary example](https://www.openstreetmap.org/#map=19/53.67517/17.78155)

317mnecl

sommerluk:
> I think the pattern is very pale - it's hard for me to see the junction between paved and unpaved even when it should be obvious:

Yes, it’s a little pale. It’s about the balance between “not obtrusive” and “easily recognizable”. If desired, we could test also a version with stronger colour contrast.

dieterdreist:
works for me. I guess it depends on the screen, I also often observe extreme differences in contrast between my 2 monitors, one of which is better than the other

imagico:
I should probably add that in the discussion above i suggested a weaker pattern because it seemed more appropriate and better readable to me.  This assessment is based on the understanding that the road class is the primary distinguishing attribute this style displays and that the surface is only a secondary attribute (or tertiary if you consider tunnel/bridge etc.) and in the practical context a strong pattern can have a quite adverse effect on readability (i.e. too weak might make the surface of the road non-distinguishable, too strong makes the map overall much worse to read).

@dieterdreist is right, things like this depend a lot on the display device and viewing conditions.

kocio-pl:
> Yes, it’s a little pale. It’s about the balance between “not obtrusive” and “easily recognizable”. 

Since this is not my pet feature, I'm OK with people preferring one or the other variant. However for me this is on the “hardly recognizable” edge of the spectrum. 

I test it on my best conditions: this is on my better color balanced monitor (the other is even lighter), I have chosen zoom level with the widest possible road style and I still don't see there's a pattern at all if I don't search it. The only thing that hints me is that a road color is a bit lighter on the bridge, but I could easily overlook it if the whole road would be the same.

> If desired, we could test also a version with stronger colour contrast.

Sure, please render few variants that are darker, up to some "definitely too dark" one.

I think we could merge this soon after technical review even in the current version and I consider the contrast testing to be the final touch. The competing solution is far from being ready (or even actively developed in 2 months), so I don't count on it, but it has much stronger solution and is still acceptable for people, so we could try with something in between.

So, just to be clear - I understand your arguments and I'm fine with staying with current contrast, however I just like to test something darker. It's also possible to tune it later if more people will not be satisfied.

sommerluk:
The current code uses `-60` as value for darkening the pattern foreground. Here also some more intense values to compare:

[screenshot-60](https://user-images.githubusercontent.com/6830724/39374051-2c22b3a4-4a39-11e8-8fdc-06eb89698583.png)
[screenshot-70](https://user-images.githubusercontent.com/6830724/39374052-2c6ef412-4a39-11e8-858d-e9f3ae9bcebd.png)
[screenshot-80](https://user-images.githubusercontent.com/6830724/39374054-2cbe226c-4a39-11e8-8971-09044777ad33.png)
[screenshot-90](https://user-images.githubusercontent.com/6830724/39374111-53eccef6-4a39-11e8-9890-4135dc4bfc1a.png)
[screenshot-100](https://user-images.githubusercontent.com/6830724/39374057-2d904cf6-4a39-11e8-9830-2a105437f7a1.png)
[screenshot-110](https://user-images.githubusercontent.com/6830724/39374059-2df5f01a-4a39-11e8-8ba4-59c91512377f.png)

Starting from `-90` it is difficult to find colours with the same hue that have the required brightness, making the visual appearance of the pattern slightly different to the original color.

sommerluk:
Sure.

In the last couple of days, I’ve made various improvements to this PR:

- The process of generating colourized versions of the pattern is now automatic. See the `scripts` folder. The script have also documentation.
- The documentation for generating the original (base) pattern has been updated and is now fine.

However, before pass to a wider test, I would talk about the following points:
- The PR does not provide pattern rendering for turning circle fills. I guess it is possible to do it (haven’t tried so far), but would of course mean: more code. Implement this??
- The PR changes the layer ordering behaviour for roads that are not bridges: Current master renders `highway=residential` with `layer=1` above `highway=motorway` without any `layer` tag. This PR would currently render the motorway above. This is a side effect of this PR. I _can_ avoid this side effect, but I’m not sure if I should: The [wiki page for the _layer_ key](https://wiki.openstreetmap.org/wiki/Key:layer) says: _“When ways are passing on different levels apply layer=* only to the way which also has the bridge/tunnel attribute.”_ Therefore, I’m inclined to _not_ change the behaviour of this PR (but do some independent optimization in the SQL query).

I think these questions should be decided before passing to a wider test. (So that we can be sure that no bigger code changes will be made which would require testing again.)

I would also like to write a general description about the principle behind this code, but this would not block testing.

sommerluk:
The [code for darkening 60, 70, 80, 90, 100, 110](https://github.com/sommerluk/openstreetmap-carto/branches/all?utf8=%E2%9C%93&query=darken-) is now available.

matkoniecz:
> highway=residential with layer=1 above highway=motorway without any layer tag. This PR would currently render the motorway above

This change will should fix rendering of some correctly mapped secondary/tertiary crossings.

For example tertiary road that is going over tunnel (and has layer=1) and then reaches secondary road should be rendered under it.

(The same applies to any pair of differently colored roads)

To be sure - other layering order is unchanged?

matkoniecz:
> The PR does not provide pattern rendering for turning circle fills. I guess it is possible to do it (haven’t tried so far), but would of course mean: more code. Implement this??

Is it requiring huge amount of code? If no - then I think it probably should be also included.

matkoniecz:
In general, thanks to everybody, especially @sommerluk @imagico and @kocio-pl for your work on this. I am now going though discussion and reviewing changes (or at least trying to do that :) )

sommerluk:
> To be sure - other layering order is unchanged?

Indeed, other layering order is unchanged!

> Is it requiring huge amount of code?

Well, in general it’s the same type of MSS and SQL like for roads. As there are no layers and (that’s the important part) the turning cycles only get rendered for few road classes, it’s of course much less code than for the roads.

kocio-pl:
110 looks quite nice for me, because it's still visible when the road is thinner. However instead of making it as dark, I agree with @imagico that you could make the dots bigger. It should be clear that there's some pattern, not that the road is darker. 

4egugm18
71ysdk7q

sommerluk:
If I understood @imagico correctly, he was rather a little sceptical about using bigger dots in the pattern.

Here are some patterns with different dot size (all in black):
[0.07 (current one)](http://www.imagico.de/map/jsdotpattern.php#x,64,jdp59980;g,2.25,16,16;rx,250,2,8,8;rx,250,2,8,8;s,jdp58799;s,jdp49103;rx,250,2,8,8;rx,250,2,8,8;s,jdp94534;rx,250,2,8,8;rd,0,0,1,scree,0.07,5,10,0,jdp40148,000000,ffffff;)
[0.10](http://www.imagico.de/map/jsdotpattern.php#x,64,jdp59980;g,2.25,16,16;rx,250,2,8,8;rx,250,2,8,8;s,jdp58799;s,jdp49103;rx,250,2,8,8;rx,250,2,8,8;s,jdp94534;rx,250,2,8,8;rd,0,0,1,scree,0.10,5,10,0,jdp40148,000000,ffffff;)
[0.13](http://www.imagico.de/map/jsdotpattern.php#x,64,jdp59980;g,2.25,16,16;rx,250,2,8,8;rx,250,2,8,8;s,jdp58799;s,jdp49103;rx,250,2,8,8;rx,250,2,8,8;s,jdp94534;rx,250,2,8,8;rd,0,0,1,scree,0.13,5,10,0,jdp40148,000000,ffffff;)
[0.16](http://www.imagico.de/map/jsdotpattern.php#x,64,jdp59980;g,2.25,16,16;rx,250,2,8,8;rx,250,2,8,8;s,jdp58799;s,jdp49103;rx,250,2,8,8;rx,250,2,8,8;s,jdp94534;rx,250,2,8,8;rd,0,0,1,scree,0.16,5,10,0,jdp40148,000000,ffffff;)

Your example is a white road. Independent of the pattern, its density, dot size and colour, it must always be darker than the original white road. The same overall-hue is only possible for the other, non-white roads. There, a bigger dot size means that the pattern foreground takes more space, therefore the same darken value gives a darker overall-hue, which will later be compensated by a lighter background (if possible). So the contrast between foreground and background becomes bigger while still having the same overall-hue. And bigger dots will however be easier to spot (except, maybe, on small roads), therefore the “darken” value could be less dark.

imagico:
As said i would advise against a stronger pattern.  If you want a stronger pattern none the less you will need a heavier pattern image to be able to maintain overall color balance with the paved color.  For that you will - as you already indicated - need to adjust the lighten/darken values as well.

And it does not help to pick a white road paved-unpaved boundary at z19 and conclude a stronger pattern leads to a more contrasting boundary - that is trivial, you don't need a test render to determine that.

The two samples in https://github.com/gravitystorm/openstreetmap-carto/pull/2640#issuecomment-328368489 were not selected arbitrarily, they represent about the range of pattern strengths i would consider suitable.  This still is my recommendation and so far i have not seen arguments that would make me change that.  If you think a stronger pattern is needed in this context you might want to reconsider the overall road rendering concept of this style because you probably want paved/unpaved to be the primary differentiation for roads instead of the road classes.

kocio-pl:
Sorry, I misunderstood you.

My points in short:
- it's very important to see the grain on the road, with current pattern it's not clear to me even in the best case
- the darker pattern, the darker road color - of course, and it's not my goal
- the pattern might be just too dense - I imagine that darker color and more sparse pattern might work together
- the pattern should not be too sparse, so even when the road is slim (lower zooms), grain should still fit on the smaller area
- testing charts are just a handy tool for a start, but they are too regular and you see mainly the differences between the roads, not the roads themselves, so the results are skewed (looks like the pattern is more clear than it will be on the map)
- now it's time to make real life testing so I started with a single case (and it already has shown something), but that's not the whole story, there should be more of them

sommerluk:
@matkoniecz Turning circle fill is implemented now. As with the road fill, I’ve split this up into two commits to make the review process easier: The first one is only code re-ordering without any functional change. The second one has the actual changes.

Note: For paths, I did not implement the pattern for turning circles. Currently we render paths as ways with a dotted line (various dot patterns depending on secondary tags). Only path areas get the “unpaved” pattern (and that with a _different_ colour than the path _line_). Because the turning circle on paths is always connected to a path _way_,  it seemed to me more consistent not to apply a pattern here.

sommerluk:
Currently, this PR assumes always `paved` as default value if there is no `surface` tag available. The list of road types:
- motorway
- trunk
- secondary
- primary
- the white roads
- living street
- aeroway
- pedestrian
- platform
- raceway
- road (highway=road)
- track area

Does `paved` as default value seam reasonable for most of them? I’m not so sure at least for “track area”…

kocio-pl:
Do you feel that this code is mostly complete and sane? If yes, we could merge it as it is soon, so it would be included in the next release, which is planned on Friday. 

I think 1 year of work is long enough and we can tune the pattern visibility later on.

sommerluk:
I’ve removed the pattern for track area. An additional argument for removing pattern for track areas: The pattern is about “surface” tag. But for track _ways_,  we do not render the _surface_ tag, but the _tracktype_ tag. This would be confusing together the track area surface rendering.

From my side, this is ready now.

kocio-pl:
@matkoniecz What's your opinion on merging current code?

sommerluk:
> move all the images into new `symbols/unpaved/` directory

Done.

sommerluk:
> we could merge it as it is soon, so it would be included in the next release, which is planned on Friday

Than I would rather suggest – instead of merging it _before_ Friday –  merging it _after_ Friday, so having it in master for some time before a release.

sommerluk:
Rebased.

kocio-pl:
OK, I have merged it now, since v4.11.0 is released already. Please test it as much as possible now.

@matkoniecz has not approved merging, but I think this is the right moment and if anything is seriously broken, this code can be reverted, but if there are some smaller issues (like my own concerns with pattern visibility), they should be just fixed now.

@sommerluk Thanks for your work! :smile:

sommerluk:
I want to say “Thank you” to everyone who has contributed here (in order of appearance in the comments):

- @imagico For advice with the pattern generation, an extension in jsdotpattern helping in this PR, writing the whole code for colour generation within a perceptional colour space, debugging the colour choice, SQL, and and and…
- @dieterdreist For the initial proposal to use comp-op (without that, this code wouldn’t have been possible)
- @pnorman For help with SQL code and alternative styling proposals
- @matthijsmelissen For help with SQL
- @mboeringa For taking part in the discussion
- @kocio-pl For discussion, testing, taking this PR further and all the release-related staff
- @HolgerJeromin For taking part in the discussion
- @matkoniecz For testing and helpful feedback to the code

sommerluk:
Earlier @imagico said _[“i think it would be nice if you could find the time to write up the underlying technical idea in a generic form”](https://github.com/gravitystorm/openstreetmap-carto/pull/2640#issuecomment-373997163)_.

Well, I’m not really good in writing documentation (especially en English), but here we go:

https://github.com/sommerluk/roadpatternrendering

Comments are welcome, including language-related corrections.

I think we should have also at least a basic documentation here in openstreetmap-carto. Maybe in `roads.mss` (but not so verbose)?

kocio-pl:
@sommerluk What do you think about this code now? It's  reverted to make the immediate effect on performance and v4.12.1 has been released, so we can think what to do next. Would you like to update it and re-submit for merging or maybe something else would be better for you?

sommerluk:
@kocio-pl Thanks for bringing this up. I’ve played around with this, see https://github.com/gravitystorm/openstreetmap-carto/issues/3280#issuecomment-403271784 for details.

PR #2643, opened 2017-05-25
use variables in font definitions

nebulon42:
Carto is now capable of expanding them.
Requires carto >= 0.17.2, carto 0.18 is our requirement.

cc @sommerluk


pnorman:
This is incompatible with our 3.x requirements, but because there are no cartographic changes that's okay.

sommerluk:
Looks good. Thanks for the patch!

kocio-pl:
Are we ready to merge this patch now?

nebulon42:
Yes, this would be against the feature freeze if it would have cartographic changes. But without we just won't backport it to the 3.x series.

PR #2653, opened 2017-06-02
New version of get-shapefiles.py

Ircama:
**New version of get-shapefiles.py**

In addition to only downloading archive updates (with the default options), the tool now performs archive expansion only in case of successfully downloaded file and generates the index only if missing or not up to date.

Some auxiliary controls have been added.

Usage and command line options remain unchanged, with a slightly different meaning for `-u`/`--update` option, which now allows to force downloading, expanding and indexing operations even if not needed.

The program now assumes that any index file already included in the downloaded archives with newer modification date than its related shp should be kept as is (without reindexing). Besides, in case `-s` option is used (indexing disabled) and when a new shp is downloaded, to avoid inconsistency with the old index, its old index file is deleted (to control this, the tool compares modification dates).

### DIFF
--- a/scripts/get-shapefiles.py
+++ b/scripts/get-shapefiles.py
@@ -20,6 +20,8 @@ import tempfile
 import logging
 import time
 import email.utils
+import atexit
+import time
 
 if sys.version_info >= (3,):
     import urllib.request as urllib2
@@ -28,6 +30,7 @@ else:
     import urllib2
     import urlparse
 
+start = time.time()
 data_dir = 'data'
 settings = {
     # Keys 1, 2, 3, ... set the arg short-options and the related process
@@ -36,11 +39,11 @@ settings = {
         'directory': 'world_boundaries',
         'url': 'http://planet.openstreetmap.org/historical-shapefiles/world_boundaries-spherical.tgz',  # noqa
         'type': 'tgz',
-        'index': [
-            'world_bnd_m.shp',
-            'builtup_area.shp',
-            'places.shp',
-            'world_boundaries_m.shp'],
+        'shp_basename': [
+            'world_bnd_m',
+            'builtup_area',
+            'places',
+            'world_boundaries_m'],
         'long_opt': '--world-boundaries'
     },
 
@@ -48,7 +51,7 @@ settings = {
         'directory': 'simplified-land-polygons-complete-3857',
         'url': 'http://data.openstreetmapdata.com/simplified-land-polygons-complete-3857.zip',  # noqa
         'type': 'zip',
-        'index': ['simplified_land_polygons.shp'],
+        'shp_basename': ['simplified_land_polygons'],
         'long_opt': '--simplified-land'
     },
 
@@ -56,7 +59,7 @@ settings = {
         'directory': 'ne_110m_admin_0_boundary_lines_land',
         'url': 'http://www.naturalearthdata.com/http//www.naturalearthdata.com/download/110m/cultural/ne_110m_admin_0_boundary_lines_land.zip',  # noqa
         'type': 'zip_dir',
-        'index': ['ne_110m_admin_0_boundary_lines_land.shp'],
+        'shp_basename': ['ne_110m_admin_0_boundary_lines_land'],
         'long_opt': '--ne-admin'
     },
 
@@ -64,7 +67,7 @@ settings = {
         'directory': 'land-polygons-split-3857',
         'url': 'http://data.openstreetmapdata.com/land-polygons-split-3857.zip',  # noqa
         'type': 'zip',
-        'index': ['land_polygons.shp'],
+        'shp_basename': ['land_polygons'],
         'long_opt': '--land-polygons'
     },
 
@@ -72,7 +75,7 @@ settings = {
         'directory': 'antarctica-icesheet-polygons-3857',
         'url': 'http://data.openstreetmapdata.com/antarctica-icesheet-polygons-3857.zip',  # noqa
         'type': 'zip',
-        'index': ['icesheet_polygons.shp'],
+        'shp_basename': ['icesheet_polygons'],
         'long_opt': '--icesheet-polygons'
     },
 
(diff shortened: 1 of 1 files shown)


scripts/get-shapefiles.py
```diff
@@ -28,6 +29,9 @@
     import urllib2
     import urlparse
 
+__version__ = '2.0.0'
```
Ircama: It was added just to mention that there have been modifications in the code. No problem, if you consider to remove this, I’ll do.

scripts/get-shapefiles.py
```diff
@@ -239,8 +263,11 @@ def main():
         help="do not run shapeindex")
     parser.add_argument(
         '-u', "--update", dest='option_force_update', action='store_true',
-        help="force downloading files even if not newer than the locally " +
-        "existing ones")
+        help="force performing an update operation even if not needed " +
+        "(e.g., downloading, expanding, indexing)")
+    parser.add_argument(
```
Ircama: See my previous answer. If you consider to remove this, I’ll do.

scripts/get-shapefiles.py
```diff
@@ -259,7 +286,7 @@ def main():
    (indexing shapes is suggested for performance improvement).\n""")
 
     if args.option_force:
-        os.chdir(os.path.dirname(__file__))
+        os.chdir(os.path.join(os.path.dirname(__file__)))
```
Ircama: It should ensure that the returned path will never be empty.

scripts/get-shapefiles.py
```diff
@@ -305,18 +338,23 @@ def main():
...
             if not args.option_no_download or not os.path.isfile(path_name):
                 print(str(element) + "-1. Downloading '" + dir_name + "'...")
-                file_name = download_file(
+                archive_file_name, download_type = download_file(
                     settings[element]['url'], args.data_dir,
                     args.option_force_update, args.option_no_curl)
 
             # Expanding
-            if not args.option_no_extract:
+            if ((not args.option_no_extract and download_type > 0) or
```
Ircama: [PEP 8 -- Style Guide for Python Code](https://www.python.org/dev/peps/pep-0008/#id19) asserts that "Long lines can be broken over multiple lines by wrapping expressions in parentheses. These should be used in preference to using a backslash for line continuation."

scripts/get-shapefiles.py
```diff
@@ -356,24 +394,72 @@ def main():
...
+                    dir_path, shp_basename + ".index")
+                shp_file_modified = None
+                if os.path.exists(shp_file_name):
+                    shp_file_modified = time.localtime(
+                        (os.path.getmtime(shp_file_name)))
+                index_file_modified = None
+                if os.path.exists(index_file_name):
+                    index_file_modified = time.localtime(
+                        (os.path.getmtime(index_file_name)))
+                if (not args.option_no_shape and shp_file_modified is None
```
Ircama: See previous note on PEP8.

scripts/get-shapefiles.py
```diff
@@ -356,24 +394,72 @@ def main():
...
+                    try:
+                        os.remove(index_file_name)
+                    except OSError:
+                        sys.exit(
+                            "\n\n\nCannot remove '" +
+                            index_file_name +
+                            "'\n")
+                if shp_file_modified is None:
+                    sys.exit("\n\n\nMissing '" + shp_file_name + "'\n")
+                if (args.option_force_update or index_file_modified is None or
```
Ircama: See previous answer on PEP8.

Ircama:
> I'm not sure if it is necessary or good to version this script independently from the style. If so it should get its own repository and be referenced from there. But this would be maybe overkill. Also: Does 2.0 mean that this contains incompatible changes to before?

Yes, by default I only expand the archive if it has been modified through a download.
Essentially, expansion is done when `download_type > 0`, which, if using *curl*, means that the local file has been modified by a downloading (or, if using the internal algorithm, occurs when an actual download happens). Expansion can anyway be forced or avoided through options.
The way to verify that a file has been actually downloaded with curl is that it does not return an error, the local file size is changed and the local modification date is changed (the latter two conditions are to handle the `-z file_name` option).
A main point is related to indexing, which at the moment is performed only in case the expanded shp file does not include an index with newer modification date than the shp file itself. This happens actually in at least one case (world_boundaries). It means that we need to trust in indexes that are embedded within the archive (at least the ones which are newer than its related shp). If you instead prefer that indexing is always performed when a file is downloaded, regardless the presence of a newer index within the downloaded archive, I can modify the code, let me know.

scripts/get-shapefiles.py
pnorman: We shouldn't independently version the utility script.

pnorman:
> I have also experimented with file modification times and if I recall correctly I found it tricky with an expanded archive. If the archive is extracted the file modification is always quite new. But you account for that by only expanding the archive if it has been modified through downloads.

Unfortunately you can't use the timestamp of files within the archive to compare with the timestamp of the archive on the server.

scripts/get-shapefiles.py
Ircama: OK thanks, I will remove the versioning

kocio-pl:
Is it OK to merge it now or are there some problems or unanswered questions still left?

pnorman:
I don't know if it's been properly tested, and I haven't looked into if it's handling [file modification times](https://github.com/gravitystorm/openstreetmap-carto/pull/2653#issuecomment-306577193)

Ircama:
> I don't know if it's been properly tested, and I haven't looked into if it's handling [file modification times](https://github.com/gravitystorm/openstreetmap-carto/pull/2653#issuecomment-306577193)

It does *not* use the timestamp of files within the downloaded archive to compare with the timestamp of the archive on the server.

It instead compares the timestamp of the downloaded archive with the timestamp of the archive on the server (also *curl* does this). This is used to take the decision to download the archive if the one on the server is newer. Then by default the archive expansion is only performed for successfully downloaded archives (regardless their content).

It also compares the timestamp of each *index* file in an archive with the corresponding *shp* file (regardless the archive timestamp). Reindexing is only performed if an index is missing or not up to date.

Ircama:
rebased

Issue #2507, opened 2016-12-12
Low zoom waterbody rendering and move to water polygons


imagico:
This issue is meant for planning and discussing implementation of low zoom waterbody rendering based on the [recently introduced preprocessed data](http://openstreetmapdata.com/data/water) and combined with that the move to water polygons for the coastline.

I assume there might be reservations regarding the technical approach used but please consider this in terms of alternatives - there are not too many other ways to get waterbodies rendered at low zoom levels.

The preprocessed data is produced for z0-z6.  There are two possibilities beyond that:

* continue with the simplified polygons for z7-z9 and use the full detail data for z10+, that would require a decision w.r.t. #2101.
* use the full detail data for z7+ already - this would require assessment of the performance implications.  Visually this looks fine in kosmtik although some more elaborate testing might not be a bad idea.

Apart from that if we go with this approach a decision would be necessary whether to use the raster or the polygon version.  A demo implementation using the raster version is available on

https://github.com/imagico/openstreetmap-carto/tree/water-lowzoom-raster

This also implements the three color scheme for water but this is not meant to suggest these changes need to be connected - it just made sense for testing.

Overall this would solve/replace #1604, #1982, #2066, #2101, #2293

and would facilitate a solution for #1547 and #1781.

nebulon42:
Regarding reservations against using preprocessed data I personally think that's ok if the data comes from OSM itself and can be periodically updated. Both is the case for the data at openstreetmapdata.com.

pnorman:
> This issue is meant for planning and discussing implementation of low zoom waterbody rendering based on the recently introduced preprocessed data and combined with that the move to water polygons for the coastline.

I'd keep them split. We already have #1982 for switching to ocean polygons, but we've never tracked down the bugs reported in #2101. If we had those solved we could revert the switch from ocean to land polygons.

For the reduced waterbodies, I'm not sure they're suitable for two reasons

1. How to generate the data isn't fully documented

2. It's not clear that this would work with different resolutions, e.g. retina tiles

imagico:
> We already have #1982 for switching to ocean polygons, but we've never tracked down the bugs reported in #2101.

I explained the connection - if we use the full detail water polygons for z7+ we don't have to use the simplified water polygons which IIRC is what caused troubles for some users.  If we have already moved to water polygons when we implement this it is no problem either.

What we cannot really do is use this approach together with land polygons but i don't think anyone would want that.

> How to generate the data isn't fully documented

[...]

PR #2654, opened 2017-06-03
Improve mid-zoom levels and water color

math1985:
This PR changes the following:
* Landcover in z<12 is now less bright.
* Built-up landcover areas are now rendered more prominently in z<12.
* Unimportant roads are now white rather than gray.
* Water is now a brighter blue (thakns @kocio-pl).
* Low/midzoom roads are now rendered with bright halos.

This should improve the legibility and esthetics of the map. In particular:
* Roads, esp. primary roads, are now better legible on zooms 8-11
* Water areas are easier to recognize
* Minor roads on mid-zoom levels are rendered more clearly
* Residential areas are more prominent and easier to recognize

### DIFF
--- a/water.mss
+++ b/water.mss
@@ -1,4 +1,4 @@
-@water-text: #6699cc;
+@water-text: darken(#6699cc, 10%);
 @glacier: #ddecec;
 @glacier-line: #9cf;
 
--- a/water.mss
+++ b/water.mss
@@ -1,4 +1,4 @@
-@water-text: darken(#6699cc, 10%);
+@water-text: #4d80b3;
 @glacier: #ddecec;
 @glacier-line: #9cf;
 
--- a/landcover.mss
+++ b/landcover.mss
@@ -9,9 +9,10 @@
 
 // --- "Base" landuses ---
 
+@built-up-upper-lowzoom: #c0c0c0;
+@built-up-lower-lowzoom: #aaaaaa;
 @residential: #e0dfdf;      // Lch(89,0,0)
 @residential-line: #b9b9b9; // Lch(75,0,0)
-@built-up-lowzoom: #aaaaaa;
 @retail: #ffd6d1;           // Lch(89,16,30)
 @retail-line: #d99c95;      // Lch(70,25,30)
 @commercial: #f2dad9;       // Lch(89,8.5,25)
@@ -210,10 +211,9 @@
   }
 
   [feature = 'landuse_residential'][zoom >= 10] {
-    polygon-fill: @residential;
-    [zoom < 13] {
-      polygon-fill: @built-up-lowzoom;
-    }
+    polygon-fill: @built-up-lower-lowzoom;
+    [zoom >= 11] { polygon-fill: @built-up-upper-lowzoom; }
+    [zoom >= 13] { polygon-fill: @residential; }
     [zoom >= 16] {
       line-width: .5;
       line-color: @residential-line;
@@ -339,10 +339,9 @@
   }
 
   [feature = 'landuse_retail'][zoom >= 10] {
-    polygon-fill: @retail;
-    [zoom < 13] {
-      polygon-fill: @built-up-lowzoom;
-    }
+    polygon-fill: @built-up-lower-lowzoom;
+    [zoom >= 11] { polygon-fill: @built-up-upper-lowzoom; }
+    [zoom >= 13] { polygon-fill: @retail; }
     [zoom >= 16] {
       line-width: 0.5;
       line-color: @retail-line;
@@ -355,10 +354,9 @@
   }
 
   [feature = 'landuse_industrial'][zoom >= 10] {
-    polygon-fill: @industrial;
-    [zoom < 13] {
-      polygon-fill: @built-up-lowzoom;
-    }
+    polygon-fill: @built-up-lower-lowzoom;
+    [zoom >= 11] { polygon-fill: @built-up-upper-lowzoom; }
+    [zoom >= 13] { polygon-fill: @industrial; }
     [zoom >= 16] {
       line-width: .5;
       line-color: @industrial-line;
@@ -400,10 +398,9 @@
   }
 
   [feature = 'landuse_commercial'][zoom >= 10] {
-    polygon-fill: @commercial;
-    [zoom < 13] {
-      polygon-fill: @built-up-lowzoom;
-    }
+    polygon-fill: @built-up-lower-lowzoom;
+    [zoom >= 11] { polygon-fill: @built-up-upper-lowzoom; }
+    [zoom >= 13] { polygon-fill: @commercial; }
     [zoom >= 16] {
       line-width: 0.5;
       line-color: @commercial-line;
--- a/landcover.mss
+++ b/landcover.mss
@@ -66,7 +66,7 @@
 
 #landcover-low-zoom[zoom < 10],
 #landcover[zoom >= 10] {
-  ::low-zoom[zoom < 10]                   { image-filters: scale-hsla(0,1,0,1,0.7,0.95,0,1); }
+  ::low-zoom[zoom < 10]                   { image-filters: scale-hsla(0,1,0,1,0.6,0.95,0,1); }
   ::lower-mid-zoom[zoom >= 10][zoom < 11] { image-filters: scale-hsla(0,1,0,1,0.6,0.95,0,1); }
   ::mid-zoom[zoom >= 11][zoom < 12]       { image-filters: scale-hsla(0,1,0,1,0.5,0.96,0,1); }
   ::upper-mid-zoom[zoom >= 12][zoom < 13] { image-filters: scale-hsla(0,1,0,1,0.4,0.97,0,1); }
--- a/landcover.mss
+++ b/landcover.mss
@@ -11,6 +11,7 @@
 
 @residential: #e0dfdf;      // Lch(89,0,0)
 @residential-line: #b9b9b9; // Lch(75,0,0)
+@built-up-lowzoom: #aaaaaa;
 @retail: #ffd6d1;           // Lch(89,16,30)
 @retail-line: #d99c95;      // Lch(70,25,30)
 @commercial: #f2dad9;       // Lch(89,8.5,25)
@@ -64,6 +65,17 @@
 
 #landcover-low-zoom[zoom < 10],
 #landcover[zoom >= 10] {
+  ::low-zoom[zoom < 10]                   { opacity: 0.3; }
+  ::lower-mid-zoom[zoom >= 10][zoom < 11] { opacity: 0.4; }
+  ::mid-zoom[zoom >= 11][zoom < 12]       { opacity: 0.5; }
+  ::upper-mid-zoom[zoom >= 12][zoom < 13] { opacity: 0.6; }
+  ::high-zoom[zoom >= 13]                 { opacity: 1; }
+
+  ::low-zoom[zoom < 10],
+  ::lower-mid-zoom[zoom >= 10][zoom < 11],
+  ::mid-zoom[zoom >= 11][zoom < 12],
+  ::upper-mid-zoom[zoom >= 12][zoom < 13],
+  ::high-zoom[zoom >= 13] {
 
   ::first {
     [feature = 'wetland_mud'],
@@ -199,6 +211,9 @@
 
   [feature = 'landuse_residential'][zoom >= 10] {
     polygon-fill: @residential;
+    [zoom < 13] {
+      polygon-fill: @built-up-lowzoom;
+    }
     [zoom >= 16] {
       line-width: .5;
       line-color: @residential-line;
@@ -325,6 +340,9 @@
 
   [feature = 'landuse_retail'][zoom >= 10] {
     polygon-fill: @retail;
+    [zoom < 13] {
+      polygon-fill: @built-up-lowzoom;
+    }
     [zoom >= 16] {
       line-width: 0.5;
       line-color: @retail-line;
@@ -338,6 +356,9 @@
 
   [feature = 'landuse_industrial'][zoom >= 10] {
     polygon-fill: @industrial;
+    [zoom < 13] {
+      polygon-fill: @built-up-lowzoom;
+    }
     [zoom >= 16] {
       line-width: .5;
       line-color: @industrial-line;
@@ -380,6 +401,9 @@
 
   [feature = 'landuse_commercial'][zoom >= 10] {
     polygon-fill: @commercial;
+    [zoom < 13] {
+      polygon-fill: @built-up-lowzoom;
+    }
     [zoom >= 16] {
       line-width: 0.5;
       line-color: @commercial-line;
@@ -575,6 +599,7 @@
     [way_pixels >= 4]  { polygon-gamma: 0.75; }
     [way_pixels >= 64] { polygon-gamma: 0.3;  }
   }
+}
 }
 
 /* man_made=cutline */
--- a/landcover.mss
+++ b/landcover.mss
@@ -65,11 +65,11 @@
 
 #landcover-low-zoom[zoom < 10],
 #landcover[zoom >= 10] {
-  ::low-zoom[zoom < 10]                   { opacity: 0.3; }
-  ::lower-mid-zoom[zoom >= 10][zoom < 11] { opacity: 0.4; }
-  ::mid-zoom[zoom >= 11][zoom < 12]       { opacity: 0.5; }
-  ::upper-mid-zoom[zoom >= 12][zoom < 13] { opacity: 0.6; }
-  ::high-zoom[zoom >= 13]                 { opacity: 1; }
+  ::low-zoom[zoom < 10]                   { image-filters: scale-hsla(0,1,0,1,0.7,0.95,0,1); }
+  ::lower-mid-zoom[zoom >= 10][zoom < 11] { image-filters: scale-hsla(0,1,0,1,0.6,0.95,0,1); }
+  ::mid-zoom[zoom >= 11][zoom < 12]       { image-filters: scale-hsla(0,1,0,1,0.5,0.96,0,1); }
+  ::upper-mid-zoom[zoom >= 12][zoom < 13] { image-filters: scale-hsla(0,1,0,1,0.4,0.97,0,1); }
+  ::high-zoom[zoom >= 13]                 { image-filters: scale-hsla(0,1,0,1,0,  1,   0,1); }
 
   ::low-zoom[zoom < 10],
   ::lower-mid-zoom[zoom >= 10][zoom < 11],
--- a/project.mml
+++ b/project.mml
@@ -102,10 +102,10 @@ Layer:
       <<: *osm2pgsql
       table: |-
         (SELECT
-            way, name, way_pixels,
+            way, name, way_pixels, religion,
             COALESCE(wetland, landuse, "natural") AS feature
           FROM (SELECT
-              way, COALESCE(name, '') AS name,
+              way, COALESCE(name, '') AS name, religion,
               ('landuse_' || (CASE WHEN landuse IN ('forest', 'military') THEN landuse ELSE NULL END)) AS landuse,
               ('natural_' || (CASE WHEN "natural" IN ('wood', 'sand', 'scree', 'shingle', 'bare_rock') THEN "natural" ELSE NULL END)) AS "natural",
               ('wetland_' || (CASE WHEN "natural" IN ('wetland', 'mud') THEN (CASE WHEN "natural" IN ('mud') THEN "natural" ELSE tags->'wetland' END) ELSE NULL END)) AS wetland,
--- a/roads.mss
+++ b/roads.mss
@@ -78,14 +78,14 @@
 @motorway-width-z10:              1.9;
 @trunk-width-z10:                 1.9;
 @primary-width-z10:               1.8;
-@secondary-width-z10:             1.4;
-@tertiary-width-z10:              1;
+@secondary-width-z10:             1.1;
+@tertiary-width-z10:              0.7;
 
 @motorway-width-z11:              2.0;
 @trunk-width-z11:                 1.9;
 @primary-width-z11:               1.8;
-@secondary-width-z11:             1.4;
-@tertiary-width-z11:              1;
+@secondary-width-z11:             1.1;
+@tertiary-width-z11:              0.7;
 
 @motorway-width-z12:              3.5;
 @motorway-link-width-z12:         1.5;
@@ -1284,10 +1284,7 @@ tertiary is rendered from z10 and is not included in osm_planet_roads. */
         line-color: @unimportant-road;
         line-width: @secondary-width-z9;
       }
-      [zoom >=10] {
-        line-color: @unimportant-road;
-        line-width: @secondary-width-z10;
-      }
+      [zoom >= 10] { line-width: @secondary-width-z10; }
       [zoom >= 11] { line-width: @secondary-width-z11; }
       [zoom >= 12] { 
         line-color: @secondary-fill;
@@ -1324,7 +1321,7 @@ tertiary is rendered from z10 and is not included in osm_planet_roads. */
         line-color: @unimportant-road;
         line-width: @tertiary-width-z10;
       }
-      [zoom >= 10] {
+      [zoom >= 11] {
         line-width: @tertiary-width-z11;
       }
       [zoom >= 12] {
--- a/roads.mss
+++ b/roads.mss
@@ -40,7 +40,7 @@
 
 @tertiary-shield: #3b3b3b;
 
-@unimportant-road: @residential-casing;
+@unimportant-road: white;
 
 @residential-construction: #aaa;
 @service-construction: #aaa;
@@ -78,12 +78,14 @@
 @motorway-width-z10:              1.9;
 @trunk-width-z10:                 1.9;
 @primary-width-z10:               1.8;
-@secondary-width-z10:             1;
+@secondary-width-z10:             1.4;
+@tertiary-width-z10:              1;
 
 @motorway-width-z11:              2.0;
 @trunk-width-z11:                 1.9;
 @primary-width-z11:               1.8;
-@secondary-width-z11:             1;
+@secondary-width-z11:             1.4;
+@tertiary-width-z11:              1;
 
 @motorway-width-z12:              3.5;
 @motorway-link-width-z12:         1.5;
@@ -91,6 +93,8 @@
 @primary-width-z12:               3.5;
 @secondary-width-z12:             3.5;
 @tertiary-width-z12:              2.5;
+@residential-width-z12:           0.5;
+@unclassified-width-z12:          0.8;
 
 @motorway-width-z13:              6;
 @motorway-link-width-z13:         4;
@@ -99,6 +103,7 @@
 @secondary-width-z13:             5;
 @tertiary-width-z13:              4;
 @residential-width-z13:           2.5;
+@service-width-z13:               0.8;
 @living-street-width-z13:         2;
 @pedestrian-width-z13:            2;
 @bridleway-width-z13:             0.3;
@@ -250,7 +255,9 @@
 @paths-tunnel-casing-width:       1;
 
 @junction-text-color:             #960000;
-@halo-color-for-minor-road: white;
+@halo-color-for-minor-road:       white;
+@lowzoom-halo-color:              white;
+@lowzoom-halo-width:              1;
 
 @motorway-oneway-arrow-color:     darken(@motorway-casing, 25%);
 @trunk-oneway-arrow-color:        darken(@trunk-casing, 25%);
@@ -1039,33 +1046,32 @@ tertiary is rendered from z10 and is not included in osm_planet_roads. */
 .roads-fill[zoom >= 10],
 .bridges-fill[zoom >= 10],
 .tunnels-fill[zoom >= 10] {
+
   ::halo {
-    [zoom = 9][feature = 'highway_secondary'] {
-      line-color: @halo-color-for-minor-road;
-      line-width: 2.2;
-      line-opacity: 0.4;
-      line-join: round;
-      //Missing line-cap: round; is intentional. It would cause rendering glow multiple times in some places - what as result of partial transparency would cause differences in rendering
-      //Also, bridges - including bridge casings - are rendered on top of roads. Enabling line-cap: round would result in glow from bridges rendered on top of road around bridges.
-    }
-    [zoom = 10][feature = 'highway_secondary'],
-    [zoom = 11][feature = 'highway_secondary'] {
-      line-color: @halo-color-for-minor-road;
-      line-width: 2.7;
-      line-opacity: 0.4;
-      line-join: round;
-      //Missing line-cap: round; is intentional. It would cause rendering glow multiple times in some places - what as result of partial transparency would cause differences in rendering
-      //Also, bridges - including bridge casings - are rendered on top of roads. Enabling line-cap: round would result in glow from bridges rendered on top of road around bridges.
-    }
-    [zoom = 10][feature = 'highway_tertiary'],
-    [zoom = 11][feature = 'highway_tertiary'],
-    [zoom = 12][feature = 'highway_unclassified'] {
-      line-color: @halo-color-for-minor-road;
-      line-width: 2.2;
-      line-opacity: 0.3;
-      line-join: round;
-      //Missing line-cap: round; is intentional. It would cause rendering glow multiple times in some places - what as result of partial transparency would cause differences in rendering
-      //Also, bridges - including bridge casings are rendered on top of roads. Enabling line-cap: round would result in glow from bridges rendered on top of road around bridges.
+    [feature = 'highway_motorway'][link != 'yes'][zoom >= 8][zoom < 12],
+    [feature = 'highway_trunk'][link != 'yes'][zoom >= 8][zoom < 12],
+    [feature = 'highway_primary'][link != 'yes'][zoom >= 8][zoom < 12],
+    [feature = 'highway_secondary'][link != 'yes'][zoom >= 11][zoom < 12] {
+      [feature = 'highway_motorway'] {
+        [zoom >= 8] { line-width: @motorway-width-z8 + 2 * @lowzoom-halo-width; }
+        [zoom >= 9] { line-width: @motorway-width-z9 + 2 * @lowzoom-halo-width; }
+        [zoom >= 10] { line-width: @motorway-width-z10 + 2 * @lowzoom-halo-width; }
+        [zoom >= 11] { line-width: @motorway-width-z11 + 2 * @lowzoom-halo-width; }
+      }
+      [feature = 'highway_trunk'] {
+        [zoom >= 8] { line-width: @trunk-width-z8 + 2 * @lowzoom-halo-width; }
+        [zoom >= 9] { line-width: @trunk-width-z9 + 2 * @lowzoom-halo-width; }
+        [zoom >= 10] { line-width: @trunk-width-z10 + 2 * @lowzoom-halo-width; }
+        [zoom >= 11] { line-width: @trunk-width-z11 + 2 * @lowzoom-halo-width; }
+      }
+      [feature = 'highway_primary'] {
+        [zoom >= 8] { line-width: @primary-width-z8 + 2 * @lowzoom-halo-width; }
+        [zoom >= 9] { line-width: @primary-width-z9 + 2 * @lowzoom-halo-width; }
+        [zoom >= 10] { line-width: @primary-width-z10 + 2 * @lowzoom-halo-width; }
+        [zoom >= 11] { line-width: @primary-width-z11 + 2 * @lowzoom-halo-width; }
+      }
+      line-color: @lowzoom-halo-color;
+      line-opacity: .4;
     }
   }
 
@@ -1316,7 +1322,10 @@ tertiary is rendered from z10 and is not included in osm_planet_roads. */
     [feature = 'highway_tertiary'] {
       [zoom >= 10] {
         line-color: @unimportant-road;
-        line-width: 0.55;
+        line-width: @tertiary-width-z10;
+      }
+      [zoom >= 10] {
+        line-width: @tertiary-width-z11;
       }
       [zoom >= 12] {
         line-color: @tertiary-fill;
@@ -1350,11 +1359,11 @@ tertiary is rendered from z10 and is not included in osm_planet_roads. */
     [feature = 'highway_unclassified'] {
       [zoom = 12][feature = 'highway_residential'] {
         line-color: @unimportant-road;
-        line-width: 0.4;
+        line-width: @residential-width-z12;
       }
       [zoom = 12][feature = 'highway_unclassified'] {
         line-color: @unimportant-road;
-        line-width: 1;
+        line-width: @unclassified-width-z12;
       }
       [zoom >= 13] {
         line-width: @residential-width-z13 - 2 * @residential-casing-width-z13;
@@ -1445,7 +1454,7 @@ tertiary is rendered from z10 and is not included in osm_planet_roads. */
 
     [feature = 'highway_service'] {
       [zoom >= 13][service = 'INT-normal'] {
-        line-width: 1;
+        line-width: @service-width-z13;
         line-color: @unimportant-road;
       }
       [zoom >= 14][service = 'INT-normal'],
--- a/style.mss
+++ b/style.mss
@@ -2,7 +2,7 @@ Map {
   background-color: @water-color;
 }
 
-@water-color: #b5d0d0;
+@water-color: #aad3df;
 @land-color: #f2efe9;
 
 @standard-halo-radius: 1;


math1985:
I decided to lump these changes together as they (at least partially) depend each other.

Note that I used opacity for the landcover layer, to prevent duplication and dim the landuse colors in a uniform manner. This should not lead to problems with ugly colors as the landcover layer is the lowest layers on the relevant zoomlevels.

On the lowest zoomlevels, not all colors can be easily distinguished. In general I think using undistinguishable colors is a bad design practice, but as users in this case can zoom in to get more information, I don't think it's a big problem in this case.

I'll post some screenshots, but for a good impression it's really best to checkout the change yourself.

@pnorman Do you have the opportunity to load this branch into your tile server?

math1985:
screenshot_2017-06-03_15-46-01
screenshot_2017-06-03_15-45-45
screenshot from 2017-06-03 15-45-13

math1985:
@pnorman Should I worry about the failing travis?

imagico:
I have not tested this change yet so no comments yet on the actual design but a few general remarks here.  This is a big change in several ways:

* it changes the water color which alone is a fairly far reaching change for the overall map appearance and the color balance in this style.  There has been a fairly lengthy but inconclusive discussion w.r.t. changing the water color in #1781 - however so far all water color ideas there have been developed for use in combination  with the current landcover color scheme - and not also for use with separate low zoom color scheme as suggested in this PR.
* the landcover color changes would mean departing from the basic design principle of this style that colors used are generally the same across all zoom levels.  This would affect both map use (users would need to learn to read the map in different ways at different zoom levels) and map design (changing a color would become more complicated design wise).
* it would also mean departing from the current principle not to use transparency to set colors when the transparency effect itself is not the aim.  The statement that _this should not lead to problems_ is not quite right because landcovers mapped above each other (like micromapping of landcovers in residential areas) and overlap between landcovers and water (which is not necessarily wrong, in particular at the coast) do practically occur and will lead to color mixing.

For info: after the 4.0 release has settled and been deployed i was planning to look into #2507 again and the changes that would be required to do this as well as looking into possibilities of low zoom landcover rendering (z<6) - without a fixed time frame, just this being high on my priority list.  My idea was to split this into smaller changes as much as possible, re-ordering the water and landcover layers would be the start and actually changing colors would have been the last step to be made when it could be evaluated in full (including the low zoom levels) and when all options are actually available (including differentiating water color by waterbody type - which is not possible at the moment because of the drawing order).

kocio-pl:
Related to or resolves: https://github.com/gravitystorm/openstreetmap-carto/issues/1781, https://github.com/gravitystorm/openstreetmap-carto/issues/2194, https://github.com/gravitystorm/openstreetmap-carto/issues/1755, https://github.com/gravitystorm/openstreetmap-carto/issues/1936.

math1985:
Good points @imagico.

> The statement that this should not lead to problems is not quite right because landcovers mapped above each other (like micromapping of landcovers in residential areas) and overlap between landcovers and water (which is not necessarily wrong, in particular at the coast) do practically occur and will lead to color mixing.

No, this is not true as transparency is applied at layer level, not feature level (transparency instead of polygon-transparency property).

imagico:
> No, this is not true as transparency is applied at layer level, not feature level (transparency instead of polygon-transparency property).

Ah - sorry, did not realize that.

I assume though the point about overlap with ocean still stands - this being in a different layer.

nebulon42:
The Travis error is due to a problem with mapnik-reference which is fixed in mapnik-reference 8.7.0 and carto 0.18.1.

imagico:
All right, i tested the change in a few areas and here a few more specific comments design wise.  I have not yet formed an opinion if i support this change overall or not, will need to contemplate it somewhat further first.

The change can essentially be split into two separate changes: the change of the water color and the low zoom land cover and roads changes.  At z13 and above the change essentially amounts to only the water color change.

## water color

The suggested water color is the same as the lakes color in the three color scheme i showed in https://github.com/gravitystorm/openstreetmap-carto/issues/1781#issuecomment-241245973 - so my comments here essentially match the impression i had when working on that.

This color works nicely on lakes of intermediate size in combination with the current land cover colors.  It is a much stronger blue tone than the current water color but the same lightness (see http://davidjohnstone.net/pages/lch-lab-colour-gradient-picker#b5d0d0,aad3df).  This makes it IMO somewhat strong for very large water areas (which is why i suggested a weaker color for the ocean) and at the same time does not really solve the problem of bad visibility of small waterways in very structured environments (which is why i suggested a stronger color for rivers).

For the suggested weaker mid zoom colors i think it is too strong for the ocean leading to an overall unbalanced appearance in coastal areas and with islands.

## weaker landcover colors at mid zoom

I think except for the previously mentioned problem of too strong large water areas when looked at in isolation the proposed look for z11/z12 looks fairly nice as a map style with strong focus on settlements and roads at mid latitude in temperate zone climate with landcover mostly limited to grass/wood/farmland.  In areas with more diverse landcover or with a different landcover focus (like in the subtropics with farmland, grass and bare ground types dominating) as well as at z10 and below the coloring is no more differentiable though.

I find overall wetlands over land are too strong with this change at the mid zoom levels and would need to be faded in a similar way as landcovers, however wetlands over water and reefs should probably not be faded with the water color proposed. ;-)

## the dark builtup areas and the bright minor roads

I think these are both ideas that would be worth pursuing independent of the other changes suggested.  These would also be changes that are not as fundamental departures from the current design as the other changes.  The dark urban landcover color for z12 and below kind of makes sense to darken the urban landcovers at the point where buildings are no more rendered.  There are some problems with this, like the contrast between the dark color and water being much worse than with the old light color.

math1985:
Thanks for the thorough review! I will address your points after we have a worldwide rendering example.

kocio-pl:
> the landcover color changes would mean departing from the basic design principle of this style that colors used are generally the same across all zoom levels.

We could test mid zoom colors also on high and low zoom levels. This would make the change even bigger, but it would be good for me to redesign these colors consciously and transparently. Original design was not documented nor really discussed later, we just assume it's somehow proper, which makes me uncomfortable when talking about changes.

math1985:
> We could test mid zoom colors also on high and low zoom levels

I tried this for high zoom levels, but the results didn't look satisfactory to me. On low zoom levels we don't have landcover.

imagico:
> Original design was not documented nor really discussed later, we just assume it's somehow proper, which makes me uncomfortable when talking about changes.

Most of the colors of the landcover layer have been changed at some point in the past during the history of this project.  Even for those which are unchanged for a long time possible changes have often been discussed in the meantime.

math1985:
@pnorman Do you have the chance to deploy this PR on your server? It is compatible with the 4.0 database layout.

kocio-pl:
Travis should be OK now. I would be also interested to see bigger area rendered, I like the idea to preview it live - Paul, what do you think about testing it?

math1985:
As @pnorman hasn't responded yet, are there others who would be able to check how this looks in their area?

kocio-pl:
Full Warsaw area from Mapzen extract:

**z8**
Before
jiyakeyj
After
3k2agfgc

**z9**
Before
dnmldqri
After
ukrutrz

**z10** (click to see unscaled images)
Before
jrpn0cn6
After
tlcw6c8g

kocio-pl:
Full Berlin area from Mapzen extract (click to see unscaled images):

**z8**
Before
hgkyxv1f
After
osgoz71r

**z9**
Before
2m6hhnni
After
oowp2gkr

**z10** 
Before
ugb2vz5d
After
nbusnphh

kocio-pl:
Full Lon area from Mapzen extract (click to see unscaled images):

**z8**
Before
ewe8yg7r
After
1ukbzppw

**z9**
Before
gnvhai1z
After
bldo8gsa

**z10** 
Before
adg0nazc
After
zhsgzgi7

imagico:
To avoid confusion with the samples and the PR description - this change modifies landcover colors up to and including z12.

kocio-pl:
Full Moscow area from Mapzen extract (click to see unscaled images):

**z8**
Before
kx3rfmao
After
zsfqgh55

**z9**
Before
znvfxdru
After
t7wndjsc

**z10** 
Before
mh1dlu2o
After
y0rmjok

kocio-pl:
Full Dublin area from Mapzen extract (click to see unscaled images):

**z10** 
Before
08fpuppr
After
e5pklz0g

**z11** 
Before
lqwmpksp
After
xlbfngfs

**z12** 
Before
qs9p yaj
After
bqt3fnkx

kocio-pl:
Full Melbourne area from Mapzen extract (click to see unscaled images):

**z8**
Before
wkgo3abx
After
pfgu18xl

**z9**
Before
jppp3002
After
1ln_x27t

**z10** 
Before
8eg5ywrt
After
ik8ctf5t

kocio-pl:
Full Vancouver area from Mapzen extract (click to see unscaled images):

**z11** 
Before
z5ihm_8w
After
ule2sqb0

**z12** 
Before
lublie0o
After
cd_xgday

pnorman:
I find z12 too dark between the roads when viewed at full size.

kocio-pl:
Full Macedonia area from Mapzen extract (click to see unscaled images):

**z8**
Before
xzf5eeo1
After
h0ortf2q

**z9**
Before
sqfdjj1e
After
q0o8ecor

**z10** 
Before
1aynmewx
After
1n5nz_qm

math1985:
Thanks a lot for the rendering examples! Do you also have an example where we render landcover over oceans?

kocio-pl:
I'm not aware of such place, but I can render it too if somebody would show me where is it.

imagico:
In the UK you have frequent beach over water for example like here:

http://www.openstreetmap.org/#map=10/53.6955/-2.9965

There is also military=danger_area:

http://www.openstreetmap.org/#map=11/54.6184/8.0379

kocio-pl:
Tokyo area (click to see unscaled images):

**z9**
Before
vww5mv5b
After
tkwtii0y

kocio-pl:
Tokyo centre (click to see unscaled images):

**z11**
Before
hbqbqimb
After
anfxuogd

**z12**
Before
qwnbxxwq
After
v9suigz9

kocio-pl:
Nationalpark Schleswig-Holsteinisches Wattenmeer (click to see unscaled images):

**z9** 
Before
axzipco2
After
vejejiia

**z10** 
Before
r8l3gnom
After
ykwmmn2j

**z11** 
Before
n6msobl
After
niqesxae

**z12** 
Before
ue3rzb1q
After
y qz8ir5

kocio-pl:
Full Liverpool area from Mapzen extract (click to see unscaled images):

**z10** 
Before
a5q9ideo
After
jbnfq95q

**z11** 
Before
7rw4l_1i
After
zlucogri

**z12** 
Before
6wfggblx
After
0zvvqzhk

math1985:
Thanks again!

In England, I see beaches outside of the coastline - is that correct tagging?

imagico:
Yes, the larger part of a beach is almost always outside the coastline.  There are opinions that mapping of beaches should be split at the coastline and the lower parts should possibly be tagged differently but that is not common practice and not intuitive or necessary and comes in conflict with [One feature, one OSM element](https://wiki.openstreetmap.org/wiki/One_feature,_one_OSM_element).

We already have preparations of a distinct rendering of parts of beaches over water in #1990 but this depends on water polygons for the ocean.

Of course significant parts of the beaches you see mapped in England are actually tidal flats.

kocio-pl:
I'm mostly happy with this PR. Just two general notes:

1. In some cities readability is lower at z11-z12 (Dublin, Vancouver and Liverpool from my renderings) - big roads and placenames are harder to notice. This is probably because residential areas are darker and smaller roads are white, which makes the background more vivid. We could accept it in the name of better visibility of smaller cities or be more conservative with darkening residential areas - for example make it graduate or limit darkening only to z10.

2. New water color is good for me in the above examples, but I would be interested in testing 3 water color set to see how big difference it makes. @imagico, would it be possible to port your water color code here?

imagico:
The three water colors scheme depends on layer reordering which depends on water polygons for the ocean which depends on #2101 which is why i try once again to get some movement into this (see https://github.com/gravitystorm/openstreetmap-carto/issues/2101#issuecomment-312610152).  

I would prefer to do this in order because these changes all together would be too many to keep separated but in sync with other changes in this style from my perspective so this would mean a permanent fork and i would then be tempted to make other changes, esp. color wise, to that fork rather than the main branch (i know i would because i am already considering it).

kocio-pl:
I consider water colors scheme as interesting tuning, which I'm still not sure if is worth the hassle and code complexity, so no problem - I can test it later.

Changing global water color would be still an improvement for me, but I think this needs testing on all zoom levels around the planet.

daganzdaanda:
I think this is a really big change in the appearance of the map, similar to the road colours tuning. I do like the results on the examples, but with such a lot of tweaks, it would be great to have a "live" test with a before/after slider. Also, I would like to invite feedback from more people before big changes like this, asking in the forae or mailing lists.

daganzdaanda:
What happens at z13 and after? I am guessing you did not change anything there. Does the switch from grey for all built-up landuses to the different colours for residential / industrial / etc work well enough?

daganzdaanda:
Small thing from the Liverpool example: The industrial grey does not contrast much with the water color (see the harbour area)

imagico:
I am still contemplating this and having a live test would definitely be important.  Of course setting this up costs resources and time which might be a problem if we do not even know yet if we want this change or not.  Maybe it would be good if @math1985 could reply to the issues raised so far before we consider a live demo.

From my point of view the appearance with this change at z8-z12 is **not** the most significant factor here.  Much more important is how this works together with the other zoom levels and how this affects the map use experience and map usability.  And how this will affect style development - both in practical questions of individual styling decisions and in terms of long term strategic direction of this project. In this regard this change goes much further than the road color change.

kocio-pl:
The more I look at this change, the more I like it. It was clever to make landuse colors fade away nicely when zooming out instead of just one paler color for some landuses on midzoom. With so pale colors, we could start rendering landuses much earlier, like z3 (or maybe even z0, since at z0-z2 we have only state borders).

This code affects readability mostly in a positive way, which means better user experience and usability for me. I don't even want to have all the details on low- and midzoom, so less visible landuse presence/differences is a good thing for me.

The downside is that we can't have one legend for all the zoom levels. But even using current short legend (which seems to be outdated, by the way) I was not able to identify all the colors, because they look different when the area is bigger or close to some others.

I don't see how automatic dimming of landuse can affect development. We have a problem with lack of free colors for some time already, and this has much stronger impact on it IMO.

math1985:
For the record, the current version does not handle landcover over seas/oceans correctly.

For example beaches over water are rendered transparently, resulting in green beaches.

screenshot_2017-07-11_21-07-33

imagico:
Yes, but this is mostly a matter of switching to water polygons and layer reordering moving the ocean layer above landcover.

And that depends mostly on progress on https://github.com/gravitystorm/openstreetmap-carto/issues/2101#issuecomment-312610152.

math1985:
Moving the ocean layer above landcover would mean beaches below the coastline would disappear - I'm not sure if that's a good idea.

imagico:
As said differentiated rendering of beaches over water and over land is already there since #1990 - that is just currently limited to z14 and above.

math1985:
> I assume though the point about overlap with ocean still stands - this being in a different layer.

This has now been fixed. I resolved this by switch to image filters rather than opacity. The idea is the same, however this switch is not pixel-perfect so there might be small differences compared to earlier version.

math1985:
Thanks for all feedback!

> water color

@imagico indicates the color works nicely on intermediate sized water areas, is too weak on small water lines and too stong on big water areas. I consider using different colors for different sized objects out of the scope of this PR, so given that assumption I think the current color is a good trade-off.

> I find z12 too dark between the roads when viewed at full size.

> [..] be more conservative with darkening residential areas - for example make it gradual or limit darkening only to z10.

I agree, perhaps Luxembourg is not the best place to test huge residential areas... I made the color on z11 and z12 lighter again.

> With so pale colors, we could start rendering landuses much earlier, like z3 

I think this is mainly a performance issue (we would need to test this). I remember someone has a proof-of-concept map rendering set up where the low zoom levels were created by stitching together high zoom-level tiles. Does anybody remember the URL to this map?

math1985:
> Small thing from the Liverpool example: The industrial grey does not contrast much with the water color (see the harbour area)

This is the same problem as https://github.com/gravitystorm/openstreetmap-carto/pull/2654#issuecomment-314543114, should be fixed now.

kocio-pl:
Liverpool is good for testing both changes, so let's visit it again.

Full Liverpool area from Mapzen extract (click to see unscaled images):

**z10** 
Before
8du2v2ie
After
mgcbkzvs

**z11** 
Before
p6o93gox
After
bfe4d_da

**z12** 
Before
w4c dxaq
After
_hpnvjt8

kocio-pl:
Dublin revisited.

Full Dublin area from Mapzen extract (click to see unscaled images):

**z10** 
Before
708eevsq
After
ts5u69a4

**z11** 
Before
bsyh bv
After
mvojzw3o

**z12** 
Before
5iba7fxd
After
8f4xnxf2

kocio-pl:
Vancouver revisited.

Full Vancouver area from Mapzen extract (click to see unscaled images):

**z11** 
Before
xios_ ho
After
767z3tqv

**z12** 
Before
ltkwduo8
After
fow3cpbl

imagico:
> I remember someone has a proof-of-concept map rendering set up where the low zoom levels were created by stitching together high zoom-level tiles. Does anybody remember the URL to this map?

That was by @woodpeck: http://fred.dev.openstreetmap.org/lowzoom/

Note this was made from the full rendering minus the labels but including the rainbow color roads at that time - which quite significantly define the overall results.  And based on z9 it only included wood in the landcover layer.

Also appearence with this approach depends a lot on how exactly you perform the scaling.  If you compare the above results for waterbodies with what i showed [here](http://www.imagico.de/map/polygon_rendering.php) you can see the difference.  In Frederik's map Labrador for example appears almost flooded.

The French style also has some limited low zoom landcover rendering - see 

https://mc.bbbike.org/mc/?lon=4.109487&lat=55.82242&zoom=4&num=2&mt0=osmfr&mt1=mapnik&marker=

AFAIK this is directly rendered with Mapnik which makes it mostly artefact bound.  Appearance of central Europe at z3 for example is likely primarily artefact noise with very little influence of the actual data.

Maybe @cquest can share his thoughts.

cquest:
The french rendering is using a pre generated big TIFF which is available at http://osm13.openstreetmap.fr/~cquest/z7.tif

It has not been updated for more than a year... one more thing to take care soon ! ;)

imagico:
@cquest - Ah, interesting.  This explains why it does not look quite how you'd expect it to look strait out of Mapnik.  Your `raster-scaling` is apparently [lanczos](https://github.com/cquest/osmfr-cartocss/blob/master/style.mss#L82) which might not be the best choice considering the source data at z7 is already pretty noisy and lanczos is quite noise retaining.  In lieu of `average` sampling i would try `bilinear`.

kocio-pl:
Low zoom levels can be further discussed here: https://github.com/gravitystorm/openstreetmap-carto/issues/2688.

math1985:
I've just been in touch with @pnorman, unfortunately he doesn't have the ability to set up a rendering server at the moment - at least not within the next couple of weeks. I think having a live preview would be really nice, but if it's not possible to set up at the moment I don't think we need to postpone merging this PR.

Therefore, I'd really encourage anybody that would like to see the effects of this PR before release to set up his own rendering server.

Generally people seem to be in favour of this PR. So unless anybody else is able to set up a live demo, I'm planning to merge this PR soon.

imagico:
Ok - i was not really in a hurry here so i planned to take some more time to contemplate things and look at this change from different angles (a live demo certainly would have helped) but since you are eager to merge it here my current take on the matter.  I want to emphasize that i already gave this quite a bit of thought and this is not just based on a quick intuitive reaction.  It is important IMO to also allow for larger changes and not get stuck in a conservative comfort zone concentrating more on what you have and fearing what you might loose instead of embracing what you might gain.

I am also open to changing my opinion if there are convincing arguments i have not considered so far.

<h3>Water color change</h3>

I would be against releasing this change as is.  The question if this change is an improvement or not has very little to do with this assessment.

The thing is - and i think i made this point before - changing the water color is a really big thing.  It will fundamentally re-define the overall look of the style, of the OSM project and about every website that shows this style in some way.  If we change this we are in a position to _sell_ this change - not only as an improvement relative to the old one but also as a good choice overall.  This would also imply that we would not change the water color again any time soon because that would clearly communicate that the first change was not a good choice overall.

The problem is i am currently convinced a three color scheme for water is a better solution.  I am open to the possibility that i am wrong about this but this would need some convincing arguments.  And because of the above i can hardly be in favor of changing the water color if i think there is a better solution within reach.

<h3>Mid zoom land cover color fading</h3>

Also here i came to the conclusion that i think we should not have this change in this style.  Again this is not so much based on the visual merits of the change and the appearence at any of the affected zoom levels seen in isolation but about the big picture and the overall implications.

Fading the landcover colors from z12 to z8 would - independent of how exactly this is done technically and if at the same time some landcovers are dropped from rendering as it is the case right now - communicate giving up on landcover rendering at the low zoom levels.  If we fade out landcover to z8 there obviously is no landcover rendering at z<8 and if we have low zoom landcover rendering we cannot reasonably also fade the colors at mid zoom.

The other thing i feel strongly about is color design.  With landcover colors being used in several different algorithm defined variants at the various zoom levels (by the way i could not find any documentation on what exactly `image-filters: scale-hsla()` does) interacting differently with the other unmodified colors would make an analytic, goal oriented color design more or less impossible.  You can already see this with the water color - which does not work equally well at the high zoom levels and low zoom levels in this change.  In the long term this would probably lead to a frequent urge to differentiate further colors by zoom level and this way further separate high zoom level design from the mid zoom levels.

The difficulty of systematic color design is not a completely new problem, i brought this up some time ago in #2270 and we failed to agree on design guidelines for a systematic approach to design choices in #2462.  In my eyes the landcover color fading proposed here would essentially eliminate the possibility to choose and develop colors based on systematic guiding principles.  This is a valid approach to design but it is not an approach i am any good at.

If i evaluate this change regarding the different goals we have formulated in [CARTOGRAPHY.md](https://github.com/gravitystorm/openstreetmap-carto/blob/master/CARTOGRAPHY.md) i must say i fail to see a clear gain of the landcover fading in almost any of the goals stated there.  Legibility and clarity of the other elements of the map is of course better when fading landcover - that is trivial.  But legibility of the landcover is at the same time severely reduced.  So this is only a gain if you put emphasis on the settlements and roads (and as i said what is proposed here is a nice looking style in that regard at these zoom levels - though ultimately quite similar to other commercial styles with a roads and settlements focus and the landcover being shown - if at all - mostly for decorative purposes).  As for the other goals:

* _mapper support_: Reduced since feedback for landcover mapping is further limited to high zoom levels.  Quality of landcover mapping over large areas would receive significantly less useful feedback.  Some plus could be argued for the more prominent urban landcover rendering which might produce an incentive for mapping these.
* _diversity_: As indicated above already the primary landcovers in the central European context are still reasonably good to differentiate at z11/12 with this change but the dominating landcovers in other parts of the world are often not.  And fading the landcovers based on a fixed zoom level criterion in particular affects high latitudes.  So to me a clear minus as well.
* _map richness_: Obviously reduced.
* _maintainability_: See above - you can likely argue how much this affects the ability to make good design choices but i don't think anyone could seriously claim this change would make it easier.
* _adaptability and ease of use_: Here you could reasonably argue for a plus since having the fading implemented would allow adjusting it.

<h3>Bright minor roads and darkened urban landcover at mid zoom</h3>

As already said i consider these changes worth exploring without the landcover fading.

<h3>Conclusion</h3>

This is of course just my current opinion so i cannot really draw a conclusion but the main result of my contemplation of this matter for me is that we should really have some discussion of our strategic goals and plans for the low zoom levels.  How our stated goals and purposes weigh at the low zooms, to what extent we would be fine with using preprocessed data and what other technical constraints and needs there are and how we deal with them.  

IIRC @gravitystorm has in the past repeatedly stated that one of his long term goals was to improve the map in the starting configuration at osm.org - which happens to be z5.  I think this is an important goal we should continue to invest in and options for this would be severely limited by this PR.

math1985:
> @gravitystorm has in the past repeatedly stated that one of his long term goals was to improve the map in the starting configuration at osm.org - which happens to be z5.

Thanks for your feedback! Before I responding, could you expand on this statement?

math1985:
Also I would be very interested in the feedback of @nebulon42 .

imagico:
I cannot quote the exact statement and it is also possible i remember this incorrectly.  Anyway the starting configuration when you call openstreetmap.org for the first time is

http://www.openstreetmap.org/#map=5/51.500/-0.100

This is what for a lot of people happens to be the first impression they get of the OpenStreetMap project.  It is therefore in my eyes not only important that the map looks pleasing and of reasonably good technically at this zoom level, it is also important that it shows a balanced excerpt of the OSM database - as suitable for this scale of course - you won't show buildings here even if we have more than 200 million of them in the database.  Showing landcover here in some form would and should therefore definitely be an option.

math1985:
Sorry, I quote the wrong line (with the other line I agree). The part I was interested to hear more about was (highlighting mine):

> I think this is an important goal we should continue to invest in and *options for this would be severely limited by this PR*.

imagico:
Ah, ok.

I thought i explained this in the arguments above.  Fading the landcover colors from z12 to z8 means landcover is gone at z<8.  Showing landcover at low zooms and fading it at mid zooms are mutually  exclusive.

Now you could argue that you could make the fading slower and extend it down to z5 but this would still mean a purely decorative and non-discernible display of landcover at the lower end of the range.

In other words: Having a readable landcover display at z5 could not really be combined with the idea of this PR.

kocio-pl:
> The thing is - and i think i made this point before - changing the water color is a really big thing. 

I agree with you - this is big change and I prefer to wait until we can have a live demo. 

> If we change this we are in a position to sell this change - not only as an improvement relative to the old one but also as a good choice overall. This would also imply that we would not change the water color again any time soon because that would clearly communicate that the first change was not a good choice overall.

I see two different things here:
1. Changing current water color to another would be a clear progress. I don't remember a single voice against changing current water color and I've made enough test renderings to know that in most cases it works better with different backgrounds. So we don't have to sell _this_ change - it's a safe bet. We can just deploy it if nothing bad happens with planet testing.
2. Changing from another color to three colors will be just fine tuning the big change, enhancing it. And this is where selling is needed - I'm not sure this will really work, more testing is needed, and maybe there will be another idea how to use more colors (@math1985 has just suggested some other system).

> The problem is i am currently convinced a three color scheme for water is a better solution. I am open to the possibility that i am wrong about this but this would need some convincing arguments. And because of the above i can hardly be in favor of changing the water color if i think there is a better solution within reach.

You said about chains of things which needs to be done to deploy your system and they are big in themselves. It took [almost a year](https://github.com/gravitystorm/openstreetmap-carto/issues/1781#issuecomment-242949101) from the moment we found a common color we think is good enough. It's about 20% time of this code existence (and 10% of entire OSM) just to go from choosing to be ready to use it! Your solution might take another months or even years, so IMO this is not "within reach", it's just "doable" for me.

Conclusions:
1. I'd like to wait some more time to see live planet demo. Paul might be not the only person able to set a low-traffic site.
2. Changing water color now would be big, but other changes will not be that radical, so we can apply them when they are ready.
3. I don't like to wait for multiple related changes that are already many months old and still not done. We might have a different map rendering solutions before they are ready.

kocio-pl:
> Legibility and clarity of the other elements of the map is of course better when fading landcover - that is trivial. But legibility of the landcover is at the same time severely reduced. So this is only a gain if you put emphasis on the settlements and roads (and as i said what is proposed here is a nice looking style in that regard at these zoom levels - though ultimately quite similar to other commercial styles with a roads and settlements focus and the landcover being shown - if at all - mostly for decorative purposes).

I see it rather as striking a balance between "ways and nodes" vs. "areas". On high zoom levels areas are rather big and don't obsfucate lines and points, so they can be quite strong without distracting users. On mid and low zoom levels it's way harder to see the lines and nodes (roads, place names, airports etc.), because there are different smaller areas there and they make a visual chaos with unpredictable patterns and the lines are thinner. Areas are still important, but being less intensive they don't hide other types of objects.

Residential areas are somewhat special, but I think it's still acceptable idea for a general map, because natural areas are way more common, so we have better balance between "natural areas" vs. "residential areas". Still, residential areas are combined into one, rather neutral color, which also decreases visual chaos.

kocio-pl:
> Fading the landcover colors from z12 to z8 means landcover is gone at z<8. Showing landcover at low zooms and fading it at mid zooms are mutually exclusive.

I don't understand why it must be gone with z<8? What about being still visible?

math1985:
As requested, I'll wait a bit longer for a rendering demo.

nebulon42:
Maybe @rrzefox can help us with a rendering preview server? I remember that they provided one for the road color change.

imagico:
To make this clear - the prerequisites for a three color water scheme are not a big deal.  This is all implemented - see https://github.com/imagico/openstreetmap-carto/tree/water-colors2 - except for the shapefile downloading in get_shapefiles.sh, which is just replacing a few filenames of course.  I just see no point in rebasing that (or re-implementing - which is possibly easier than rebasing here) without a decision on #2101.

The technical part here i would consider simply irrelevant compared to evaluation and tuning of the design.

If anyone wants to argue that the landcover fading serves the goals of this style i would be interested in specific examples.

rrzefox:
> Maybe @rrzefox can help us with a rendering preview server? I remember that they provided one for the road color change.

We're still on 3.x, but it seems this applies fine on 3.x too.
I've deployed v 3.3.1 with this applied on top. It's going to take a while to rerender Z0-12 though, so better have a look at this tomorrow.
https://osm.rrze.fau.de/map-ll-osm?zoom=12&lat=51.4114&lon=3.9186&layer=RRZE%20tileserver%20osm.org%20style%20tiles

imagico:
Two more specific observations:

* landuse=railway is no more rendered like landuse=industrial
* landuse=quarry looks like urban landuse at about z12-z10.

math1985:
Thanks a lot @rrzefox! This is independent from the database change, so applying this on top of 3.3.1 should give the same results.

math1985:
And a before/after demo: http://bl.ocks.org/math1985/raw/af7a602c222dbf1ff1a2c0d84ed755b7/

kocio-pl:
Military areas look too strong on midzoom.

water.mss
```diff
@@ -1,4 +1,4 @@
-@water-text: #6699cc;
+@water-text: darken(#6699cc, 10%);
```
kocio-pl: Now that the color testing is over, I would simply change it to a proper hex value (#4D80B3).

imagico:
Another observation from browsing the demo map (which by the way - independent of if we merge this change or not - is a really interesting study on how map styles work in terms of colors and color interactions).

I still try to wrap my head around how `scale-hsla()` works.  It seems the color modification is highly non-homogeneous.  Some colors (like park or farmland) appear to be almost invariant regarding the fading function while others change strongly.  This leads to peculiar effects like for example sand color being darker than farmland at z13+ (Lightness 92 vs. 94) but a tiny bit lighter at z10 (Lightness 95 vs. 94).  In a similar fashion park decreases in chroma from 30 to 23 from z13 to z10 while grass goes from 32 to 20.

This also means color design would not only be more difficult because of the changing relationship between faded and non-faded colors - even among the faded colors the relationships change in a highly non-intuitive way.

You can see the lack of fading on mud color by the way:

http://bl.ocks.org/math1985/raw/af7a602c222dbf1ff1a2c0d84ed755b7/#11.00/53.6504/9.2675

because it is rendered in the `landcover-area-symbols` layer.

nebulon42:
Some general comments how I feel about this change since @math1985 requested it (I'm currently more focused on engineering than design due to the CartoCSS work).

I think this PR contains very important work in terms of a design study and a lot of changes that make the map look better like better water colour (looks simply fresher), paler landuse (esp. forest) on mid zoom levels, related white minor roads. If we would merge it a lot would look better and a lot would have to be tweaked afterwards like glaciers, military areas, museum/zoo areas and nature reserves.

I'm quite happy with z12 and especially in foresty areas like Austria you are now longer drowning in a big green sea. I'm not so happy with the difference between z13 and z12, but I understand that somewhere you have to draw the line. I'm also not too happy about some of the pale green tones. They are a bit too blueish for me and I would prefer some more yellowish green. Also the interaction between grass and forest is not perfect on some zooms.

What is important that it is tried even if it is such a big change, thank you for that. However, I would be slightly more comfortable if we tried to take the good things that this PR does, bring them into a feasible order and change them one by one. This would enable us to make more adjustments on the way. The risk with merging this change all in one is that all the subsequent adjustments result in a lot of changes afterwards which might be a bit chaotic. I also have to admit that I do not have a plan how it could be done in small steps, but it feels safer.

So I would be more with @imagico by saying do one step at a time, but this goes a long way in the right direction. Sorry that I don't have more detailed comments at this time.

math1985:
> This leads to peculiar effects like for example sand color being darker than farmland at z13+ (Lightness 92 vs. 94) but a tiny bit lighter at z10 (Lightness 95 vs. 94). In a similar fashion park decreases in chroma from 30 to 23 from z13 to z10 while grass goes from 32 to 20.

Is this lightness from hsla color space or from CIELCh? In the second case, would you mind repeating your calculations in hsla? Maybe the colorspace difference is to blame. In general, I intended all lightness values to move towards 93 (the lightness of the background in hsla).
I didn't find documentation on scale-hsla either, but I expected scale-hsla(0,1,0,1,0.7,0.95,0,1) to mean 'map all lightness value on a range from 0.7 to 0.95.

imagico:
I was referring to perceptual lightness (Lch/Lab).  I am not really that familiar with hsla which might be the reason why i don't really get scale-hsla().  But as i understand it a certain hsla lightness can be fairly different perceptual lightness depending on chroma/hue - which kind of screws the whole thing.

I am not really sure what your intention actually is here - with the original `opacity` based approach you were essentially performing a mixture between the high zoom color and land-color (or water-color, which is why you changed it) - in sRGB of course.  If scale-hsla() perfoms a rescaling of hsla lightness that is something completely different.

kocio-pl:
@rrzefox Do you plan to render also z13+ tiles to look how new water color interacts with standard colors?

rrzefox:
> @rrzefox Do you plan to render also z13+ tiles to look how new water color interacts with standard colors?

I do not plan to rerender all of these. They will however naturally be rerendered when they're too old or someone manually dirtys them.
Also, the rerender up to Z12 finished, all tile <= Z12 should now show the new look.

imagico:
Picking up on:

> So I would be more with @imagico by saying do one step at a time, but this goes a long way in the right direction.

Maybe we can break this into the following steps:

* implement and test the bright minor roads and the darkened and unified urban landuses at mid zoom based on the current color set.
* come to a decision on #2101 so we can evaluate the different options we have regarding water colors with less technical constraints than we have now and then release whatever we decide is best.
* have a discussion on our strategy and long term goals for the low zoom levels and
* if as a result of this discussion we come to the conclusion that we want to use a different color scheme for landcover or no landcover rendering at all like now for the low zoom levels we look into how to nicely transit from the high zoom level landcover rendering to what we have in mind for the low zoom levels.

All of these points would directly profit from the work @math1985 put into this PR so i would not view this as rejecting this but rather as splitting it into smaller chunks to be considered and compared to alternatives individually.

By the way i also noticed we still have the non-OSM stone age VMAP0 urban area data at z8/9 - getting rid of that would also be something to consider together with the urban landuse mid zoom recoloring.

dieterdreist:
I agree with @Imagico that fading midzoom area colors will put more emphasis on settlements and roads, will make the map more similar to automobile focused maps, hence is less suitable to showcase the depth and diversity of OSM. Fading landuses make the map less readable (landscape and cultural aspects) if you are not so much interested in the road network.

It's also an important consideration that it will cut off the way to more (landuse based) area fill in lowzoom.

While there is a lot of work in this and the result is mostly visually pleasing (and working better as faded background for saturated overlays), I don't think we should go into this direction with the main style, it would be great as a second style though (explicitly intended as faded basemap). IMHO we should keep the contrast of the things (areas, area structure) we intend to show, rather than fading all into a "uniform" background.

kocio-pl:
@math1985 What's your plan now? For me chopping this PR into smaller parts is acceptable (and maybe even preferable), but merging it as it is and tuning it later is also possible solution.

math1985:
I made a small change: I felt secondary and tertiary roads on z11 and z12 were taking up too much space (especially in urban areas) so I made these narrower again.

math1985:
Some remaining things I didn't yet comment on:

> I find overall wetlands over land are too strong with this change at the mid zoom levels and would need to be faded in a similar way as landcovers, however wetlands over water and reefs should probably not be faded with the water color proposed. ;-)

I have no strong opinion on this. To me the current wetland colors are not too bad. Keeping the blue lines of wetlands the same as the blue of the water seems like a logical option to me. For now I'll keep things like they are.


> I thought i explained this in the arguments above. Fading the landcover colors from z12 to z8 means landcover is gone at z<8. Showing landcover at low zooms and fading it at mid zooms are mutually exclusive.

My intention was not to fade away the landuse color to z8 in a way that it's gone on z<8. My intention was to choose a color for landuse that looks best at that zoomlevel - or rather the strongest color that still allows us to discern other features - which happens to be very faded out colors. I think the z8 colors will also work on lower zoomlevels.

>  The problem is i am currently convinced a three color scheme for water is a better solution. I am open to the possibility that i am wrong about this but this would need some convincing arguments. And because of the above i can hardly be in favor of changing the water color if i think there is a better solution within reach.

Using a three-color water scheme will require making this PR dependent on the oceans reorder scheme.
Both are quite big and complicated changes, I'm not in favour of making big changes dependent on each other. So I won't implement this now, but I really don't mind making a change now and changing it again whenever we have the technical feasability to do so.

math1985:
> Maybe we can break this into the following steps:

I've thought about this, but I'm going to keep this as a single pull request. In general I'm in favour of keeping pull requests small, but I think that's difficult in this case. The shade of gray of urban landuses depends on the color of the other landuse. So testing darkened urban landuses without considering the landuse lightening does not really make sense. Likewise, bright minor roads require darkened landuse. And also the water color and landuse colors are related.

I'm happy with changing the colors again if our strategy or low-zoom plans require something different.


> IIRC @gravitystorm has in the past repeatedly stated that one of his long term goals was to improve the map in the starting configuration at osm.org - which happens to be z5. I think this is an important goal we should continue to invest in and options for this would be severely limited by this PR.

I agree this is an important goal, I might take up the lowest zoomlevels as next step. I think the design decisions made now for z8 might also work for lower zoom levels, but this is something we'd need to test. 

> Military areas look too strong on midzoom.

Military rendering is a quite complex topic, I'd prefer that as a separate PR.

math1985:
I'll get back to the remaining remarks and questions tomorrow.

Tomasz-W:
It's very good change, but I have 2 small remarks:
-military areas should be less emphasized
-borders of national parks etc. should be less dark

imagico:
> My intention was not to fade away the landuse color to z8 in a way that it's gone on z<8. My intention was to choose a color for landuse that looks best at that zoomlevel - or rather the strongest color that still allows us to discern other features - which happens to be very faded out colors. I think the z8 colors will also work on lower zoomlevels.

This i already understood but it should be clear that this would mean landcover is considered completely secondary to the rest of the map content at these zoom levels.  Since general map readability gets increasingly difficult as you zoom out with your approach and giving strict priority to everything else above landcover you are forced to increasingly fade out the colors and end up with just a hint of decorative coloring at z8/9.  The only reason why this is still in some places discernible at z8 is because it only comprises wood.

And extending this concept to even lower zoom levels would mean the same purely decorative hint of color at these scales which is quite obviously not worth the trouble of a more sophisticated processing as it would be required for a meaningful and readable landcover rendering.

> I'm not in favour of making big changes dependent on each other. So I won't implement this now, but I really don't mind making a change now and changing it again whenever we have the technical feasability to do so.

> I'm happy with changing the colors again if our strategy or low-zoom plans require something different.

As i already have hinted (see #2270) I think we are at a crossroads here in terms of overall project strategy.  On the one hand we have the idea of development being firmly guided by a long term strategy and goals so each change is designed to work towards this.  On the other hand we have the more localized approach of making individual design decisions with little concern for the overall picuture in the hope that the ideas and preferences developers have on a case-by-case basis in sum lead towards the long term goals they share - even if individually these changes frequently contain a back and forth in terms of design choices made.

As said before i see the merits of both approaches, especially also the fact that the latter approach would make it at least in the current state much easier to make individual decisions because you can ignore the larger picture but i don't think this approach will ultimately lead to balancing the goals and purposes we agreed on because while some of them can be served well with a compartmentalized approach others depend strongly on a broader vision and strategy.  

But while - if i ignore my personal preferences in that matter for the moment - both directions have their benefits,  i think our user base will in the long term not respond so well if changes, especially higher impact changes, keep going in seemingly random directions and there is no overall sense of direction visible.  Parts of our user base are very open to changes and welcome them with little concern for side effects or long term consequences but we also have a lot of users who look at changes more critically.  So far if there have been larger changes (like the roads color change) the most convincing argument in discourse about them has in the past usually been their importance for the long term goals and strategic plans for this style.  If this is lost as a point we can make to explain our changes i think at least our ability to make unpopular but necessary changes would be severely limited.

And to give a more specific example - if i'd try to adopt a more compartmentalized aproach to development here i would for example likely propose either to roll back or side track several of the more recent color changes, airport buildings, playgrounds and dog parks for example (side track here means change them in a way that is not exactly back the way it was but clearly in a very different direction than the previous change).  I did not do this so far because (a) i did not want to mess with the ideas and visions others had when making these changes just because i don't see them myself, (b) if these ideas turn out to be in error and the change turns out to have been a mistake i want to give those who made them the opportunity to recognize this and fix it themselves and (c) because it is tough to sell changes going non-systematically in different directions.  But i don't think there would be any point in continuing to do so if everyone else follows a compartmentalized development approach.

rrzefox:
> I made a small change: I felt secondary and tertiary roads on z11 and z12 were taking up too much space (especially in urban areas) so I made these narrower again.

Where? I would update the stylesheet on my demo server, but I see no commits in this branch since Jul 11. Did you not push that?

And just to add my two cents: I like the faded landuse very much, though I think it is maybe a bit too faint at Z8 it's definitely a huge improvement.
Coincidently, I experimented with showing landuse in the past, if you switch to the layer 'RRZE tileserver osm.org style with different lowzoom tiles' on the demoserver mentioned above, you can see landuse rendered in <= Z8. But this was done without fading, so it clutters the map too much. It would look much better with the fading.

daganzdaanda:
I agree with @imagico that we should try to keep to a plan or strategy for where the main map style should be going. The discussion that started here already is a good thing, and I think we should ask people in the forums and mailing lists a few questions about the possible future direction: 

* What would you like to emphasize on the low zooms? Maybe some groups could help in thinking about this: We could emphasize landcover like forests, water, deserts, built-up areas... Or infrastructure like streets, railways, ferry lines... Or political borders, other borders, and cities... 
The map will probably show something from every group, but if there's a conflict, what would you prefer?

* How important is the "Mapper feedback loop" to you? If we lower the visibility of eg forests on the lower zooms, you might not notice your changes there as easily - but a change in a road or river might be more visible than before.

* Also for higher zooms, what about the "feedback" - should the map try to show as much as possible of the data in OSM, even if it may not be very readable in some places? Or should we value readability more, and sacrifice some things that we are already showing? If so, what would you remove?

daganzdaanda:
More on-topic comments on this PR, without any strategy in mind: 

I think the change in forests between z12 and 13 is too much. I would prefer forests on zooms 13+ paler, but z12 darker, and then darken z11-8 too. I think the current colour at z10 or even 11 could be a good end point for z8 (and lower, if we want to start showing forests earlier). 

Also, I love the white roads without casing in theory, and in many places they are great. But they are not very visible over the default landuse. Yes, these are roads that are of lesser importance in each zoom level, but I guess most of the roads in the database don't run over a defined landuse. So we need to be careful here.

imagico:
@rrzefox - is this rendered directly at target scale with Mapnik or resampled from a higher resolution render like in the french style?

Even if this is not ideal obviously in several aspects it IMO well illustrates the usefulness of readable low zoom landcover rendering for mapper feedback and for orientation on the map - despite wood/forest being the only landcover shown and in a styling that works poorly together with the water color.

water.mss
math1985: Done!

math1985:
A few of you mention that the faded effect is too strong on z8. Based on the comments and rendering demo I've made the dimming at z8 and z9 a bit less extreme.

@rrzefox, would it be possible to update your tile server? The other commit should now have been pushed as well.

@imagico in general your view seems to be to have a long-term plan first, and adapt to that.

kocio-pl:
>  The discussion that started here already is a good thing, and I think we should ask people in the forums and mailing lists a few questions about the possible future direction:

Great idea! Do you plan to post it?

I think it would be good to get in touch with people. We might have our visions here, but in the end we make default style for the whole OSM community, so it might me interesting/helping to know how they see it.

kocio-pl:
We don't have one place to discuss long-term strategic issues, so I will also put this PR in the context of our goals:

**Legibility and clarity** - Midzoom with faded landuses is absolutely more clear for me. "We also aim for the map appearance to be esthetically pleasing." - this is also true for me.
**Being understandable and supportive for mappers** - nothing changes with being understandable IMO. We might take away some support for drawing forests and farmlands, but at the same time we add some for drawing rivers and residential areas, so overall support is not vanishing.
**Diversity** - it doesn't break anything if there are less mapped landuses, but helps a lot when there's a lot of them.
**A rich map** - please note that this rule is about the whole map, not any particular zoom level. Otherwise shifting objects from z17 to higher zooms would be not "clearing the mess" (for the sake of clarity), but "making map poor", which is not the case.
**Maintainability** - adding more code decreases it.
**Adaptability and ease of use** - adding more code decreases ease of use, but increases adaptability.

We have currently no more detailed rules, but in the old version we had: _The colour palette should be moving towards pastel/light/desaturated for background layers_ and in the abandoned part 2 of guideline proposition there was: _For area colors we prefer light and low saturation tones. Lines, individual symbols and labels also can be stronger in color._ - which is exactly what this PR is doing.

imagico:
Note stating that you think these changes (all of them together by the way - you did not in any way differentiate between them) are good and you think they work towards the goals as you see them does not really bring us forward here.  For the water color change, the dark urban landcover and the bright minor roads i did not even argue that these are not improvements.  But i tried very hard to explain in detail above the reasons why i think **the landcover fading** works against our goals and i think these reasons are mostly objectively verifiable.  In other words: I don't simply state i think this is a change to the worse, i explain, partly even in specific numbers (in case of the inconsistency of the color fading function) why i think this is the case.

So far it seems to me no one seriously tried to argue any of the points i brought forward.  @math1985 explained that for him my argument that changing the water color would prejudice future color changes for water is not true because this is not the case for him - which i can respect, but i also explained that not feeling bound by an overall strategy and a sense of direction in individual design decisions for me is a problem on its own.

I cannot even begin to understand at the moment how you can think that the landcover fading would not negatively affect the ability to design landcover colors or how you can think that fading the landcover colors at fixed zoom level thresholds does not negatively affect diversity when used with a variable scale Mercator projection (the landcovers fade at different scales if you zoom out in Nigeria or in Norway!) but i would be really interested if someone could explain this to me in a way i can understand.

kocio-pl:
Right, I was not specific enough: I've been talking about landuse fading. But please don't say that I'm not trying hard. Just reading your comments was challenging for me because of many topics and many words used - this PR is not my only activity in OSM and we spent 1,5 month here discussing and testing even older idea, so lack of attention is definitely showing. Trying to look from a long-term goals PoV also requires some effort, which I took, no matter how incomplete or wrong my take could be.

> I cannot even begin to understand at the moment how you can think that the landcover fading would not negatively affect the ability to design landcover colors or how you can think that fading the landcover colors at fixed zoom level thresholds does not negatively affect diversity when used with a variable scale Mercator projection (the landcovers fade at different scales if you zoom out in Nigeria or in Norway!) but i would be really interested if someone could explain this to me in a way i can understand.

Shorter sentences and using lists would help me in reading your comments.

1. Of course fading limits the colors we could use. But, as I said before, the main problem is we're out of colors anyway for about 2 years ago (new road colors scheme) or maybe even longer. We're left with just shades now and I wouldn't advise to use a shade of already used color for something different. Zoom-fading is the only practical use of them I see at this moment.

2. I haven't thought about projection in this context. I think of it as a systematic error we have to live with until we make big effort to overcome Mercator one day. Color fading is not harsh method of dealing with well-mapped areas problems, so I accept it. There are other fields I can think of where we could help diversity in a more visible way.

math1985:
> Shorter sentences and using lists would help me in reading your comments.

+1

imagico:
I am sorry if i appeared to imply that you or others have not been trying to articulate their view.  I just meant to say that i did not see any substantial counterarguments for the arguments i put behind my assessment.

If you have troubles understanding my arguments just ask (like @math1985 did).  I try to make my points as clear as i can but a complex consideration cannot always be put into simple words.

kocio-pl:
> I am sorry if i appeared to imply that you or others have not been trying to articulate their view. I just meant to say that i did not see any substantial counterarguments for the arguments i put behind my assessment.

No offence taken. =} I felt you were upset with my arguments, so just showing that you're not being ignored.

> If you have troubles understanding my arguments just ask (like @math1985 did). I try to make my points as clear as i can but a complex consideration cannot always be put into simple words.

I hope that when I'll try to look from your point of view and you try to be more concise, we may find common ground. There's no guarantee it can be done, even if we try, of course. But at least you know what could help it.

rrzefox:
> is this rendered directly at target scale with Mapnik or resampled from a higher resolution render like in the french style?

It was rendered once in Z8 and then scaled from that. The script used contains an explanation at the top and can be found here:
http://git.rrze.uni-erlangen.de/gitweb/?p=osmrrze.git;a=blob;f=scripts/osm-mergemaptiles.pl;hb=HEAD

> @rrzefox, would it be possible to update your tile server?

I have updated it and rerendered Z0-12. For Z13+ use /dirty as needed.

imagico:
> It was rendered once in Z8 and then scaled from that. The script used contains an explanation at the top and can be found here:
http://git.rrze.uni-erlangen.de/gitweb/?p=osmrrze.git;a=blob;f=scripts/osm-mergemaptiles.pl;hb=HEAD

Thanks - like in case of the French style it could be worth trying different sampling methods.

math1985:
Thank you for all feedback. Most of the remaining issues are things I don't intend to resolve now. Some specific comments:

> On the one hand we have the idea of development being firmly guided by a long term strategy and goals so each change is designed to work towards this. On the other hand we have the more localized approach of making individual design decisions with little concern for the overall picuture in the hope that the ideas and preferences developers have on a case-by-case basis in sum lead towards the long term goals they share - even if individually these changes frequently contain a back and forth in terms of design choices made.

We have been working on this PR for more than a year. Constructing a long-term strategy first and then designing a stylesheet that adheres to this long-term vision across all zoom levels would likely take multiple years. And usually the best way to find out what works is to try it out, so I don't think constructing a vision without testing it in practice works either. So I think we can do much better than the current system, even it that involves undoing changes sometimes.

> [..] i would for example likely propose either to roll back or side track several of the more recent color changes, airport buildings, playgrounds and dog parks for example (side track here means change them in a way that is not exactly back the way it was but clearly in a very different direction than the previous change). I did not do this so far because (a) i did not want to mess with the ideas and visions others had when making these changes just because i don't see them myself, (b) if these ideas turn out to be in error and the change turns out to have been a mistake i want to give those who made them the opportunity to recognize this and fix it themselves and (c) because it is tough to sell changes going non-systematically in different directions. But i don't think there would be any point in continuing to do so if everyone else follows a compartmentalized development approach.

This is really not the open-source way of doing things, I would really encourage you to make contributions whenever you feel like, and not to worry about somehow disrespecting the vision of others. In any case I'd encourage you to create more pull requests, concrete code makes discussing things much easier.

> If i evaluate this change regarding the different goals we have formulated in CARTOGRAPHY.md i must say i fail to see a clear gain of the landcover fading in almost any of the goals stated there.

Yes, we gain a lot in legibility at the cost of some mapper support. See also the analysis of @kocio-pl.

> I agree with @imagico that fading midzoom area colors will put more emphasis on settlements and roads, will make the map more similar to automobile focused maps, hence is less suitable to showcase the depth and diversity of OSM. Fading landuses make the map less readable (landscape and cultural aspects) if you are not so much interested in the road network.

It is not only settlements and roads that are more legible now. Mid-zoom landuse was so prominent that it made everything that is not landuse hard to read, including rivers, toponyms etc. But I agree that we should not make the map too road-centered. In the future I hope to tone down roads on the lowest zoomlevels.

> landuse=railway is no more rendered like landuse=industrial

True, on low zoom industrial landuse is rendered like urban landuse while railway is not. This seems acceptable to me.

> landuse=quarry looks like urban landuse at about z12-z10.

Yes, we might need to change the quarry color. I think we can do that as a change separate from this PR.

> I am not really sure what your intention actually is here - with the original opacity based approach you were essentially performing a mixture between the high zoom color and land-color (or water-color, which is why you changed it) - in sRGB of course. If scale-hsla() perfoms a rescaling of hsla lightness that is something completely different.

What we do in this proposal is move the landuse lightness towards the lightness of the background. As far as I'm concerned, this looks good - perhaps even a bit better than the mixing solution, as keeping saturation a bit higher makes it easier to tell the dimmed landuse colors apart.

> Also, I love the white roads without casing in theory, and in many places they are great. But they are not very visible over the default landuse. Yes, these are roads that are of lesser importance in each zoom level, but I guess most of the roads in the database don't run over a defined landuse. So we need to be careful here.

It's something to keep in mind, but to me the contrast is currently strong enough.

Let me know if I forgot to respond to any comments.

imagico:
> We have been working on this PR for more than a year.

I was not aware of that - who is 'we' here?

> This is really not the open-source way of doing things [...]

May i remind you that this project - and in particular also this PR - is not primarily a software project but a cartographic design project.  _Divide and conquer_ is a very successful concept in software development but it does not really work that well in design.

> Let me know if I forgot to respond to any comments.

I don't think you forgot to respond but as said before i am seriously missing counterarguments against most of the reasons i put forward why i think the landcover fading is a bad idea.

The only real argument i have seen for the landcover fading is improving visibility of the non-landcover features (which as said is trivial but which does not mean the same improvement cannot be accomplished by other means).  Beyond that legibility and clarity would suffer because (a) landcover information is missing or badly readable and (b) because the changing meanings of colors and the changing color relationships make reading the map much harder.  Reading colors strongly depends on context and color interactions and if you change the context by fading some colors while keeping others the same the map reading process gets all messed up.  You would essentially have to learn reading the map separately for each of the fading steps.

I did not see any evidence at all that this change improves on our goals of maintainability, diversity, mapper support and map richness and i explained in detail why i think these points would suffer.

But i of course understand i cannot make you see the points i see - even if in some cases i even put them into numbers.  So in the end i suppose we can only agree to disagree on a fairly fundamental level in this case.  And independent of the question if this PR is actually an improvement or not and if it is merged or not i fear that our inability to actually substantially communicate about the issues seen prevents us from developing a better solution.

kocio-pl:
> May i remind you that this project - and in particular also this PR - is not primarily a software project but a cartographic design project. 

Don't underestimate it. It's a complex software project, even if we're just counting the LOC. I can easily design a killer feature like switching map language, which would hugely improve diversity, but deploying it would require big changes in the software.

> Divide and conquer is a very successful concept in software development but it does not really work that well in design.

Please remember that it was not always this way, top-down approach was once standard in software creation too. But with complex projects (like this), evolving data and even data models, it performs poorly. That's the reason bottom-up approach is getting more traction. With accessible tools and more people interested in the subject, map designing is no different and OSM is _the_ game changer on this field.

> The only real argument i have seen for the landcover fading is improving visibility of the non-landcover features (which as said is trivial but which does not mean the same improvement cannot be accomplished by other means). 

I don't remember other means you've ever mentioned. What would you propose then?

> Beyond that legibility and clarity would suffer because (a) landcover information is missing or badly readable and (b) because the changing meanings of colors and the changing color relationships make reading the map much harder. Reading colors strongly depends on context and color interactions and if you change the context by fading some colors while keeping others the same the map reading process gets all messed up. You would essentially have to learn reading the map separately for each of the fading steps.

That's the point: changing zoom level changes the context in a fundamental way and legibility/clarity is not a property of just landuses. Fading is a natural thing for me when zooming out, I don't have to learn anything. It's much more natural than disappearing features. Currently on midzoom landuses are too strong (legibility and clarity of line and point features suffers a lot) and on low zoom they are invisible (so the roads are dominating z5). Fading allows us to solve both problems.

> I did not see any evidence at all that this change improves on our goals of maintainability, diversity, mapper support and map richness and i explained in detail why i think these points would suffer.

In my opinion it greatly improves overall legibility and clarity (I don't know why you're talking only about landuses legibility and clarity - the map has more elements, why are you not mentioning them?) and map richness is preserved.

> And independent of the question if this PR is actually an improvement or not and if it is merged or not i fear that our inability to actually substantially communicate about the issues seen prevents us from developing a better solution.

Please, propose some other solution if you have one.

math1985:
I am going to merge this pull request. I would like to thank everybody for their contributions. In particular I would like to thank @kocio-pl and @imagico for their work on the water color, and @rrzefox for the rendering demo.

Of course, PRs for further finetuning are more than welcome.

PR #2655, opened 2017-06-03
Don't render malls as dots

### LINKED ISSUE #2651: shop=mall dot at z17-19
I just noticed a big fat dot on z18, that turned out to be a mall: http://www.openstreetmap.org/node/380498034
It works on ways, too: https://www.openstreetmap.org/way/126248046

It seems as if the name still shows until z15 or z16, but then at z17 the generic dot takes over and gets bigger for z18 and z19.  The style of the dot is a bit strange, too - it has a black border.
Possibly this was introduced with https://github.com/gravitystorm/openstreetmap-carto/pull/2589
The black outline is discussed there and @math1985 had adressed that issue. Maybe malls were overlooked? 

I think we should not switch from a red name without dot to a pink dot without name. So how to make it more consistent?
Do we have any other names in red? Are malls worth a special treatment?
Should we add an icon (see #2276) to the name from z15 and keep it until z19?
Or return to the previous state? (IIRC, malls only had the red text and no icon.)

math1985:
This resolves #2651

### DIFF
--- a/amenity-points.mss
+++ b/amenity-points.mss
@@ -566,7 +566,7 @@
   }
 
   [feature = 'shop'] {
-    [zoom >= 17],
+    [shop != 'mall'][zoom >= 17],
     [shop = 'supermarket'][zoom >= 16],
     [shop = 'department_store'][zoom >= 16] {
       marker-placement: interior;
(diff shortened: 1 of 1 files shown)


math1985:
Could anybody review this? Just a small change.

Issue #2619, opened 2017-04-24
Shop rendering suggestion


MaestroGlanz:
You might think "Oh no, not again." 

I read a bit into #2444, but I don't think my ideas fit into the discussion.

I can do programming, but I dont know, how osm-carto works. Therefore I might suggest things already implemented. 

1. osm-carto should not limit the shop icons. If something is forked from it, this maybe useful (i.e. an interactive shopping map).
2. osm-carto should only map the 20 (or what-number-ever) most often shops (see https://taginfo.openstreetmap.org/keys/shop#values)
3. Which kind of shops are or arent mapped should be determined by a stylesheet (list of booleans). For every zoom level, there can be another stylesheet. i.e. on zoom level x only the 5 most often mapped shop types are rendered.
4. Introduce another highest zoom level with the same zoom of the highest level. If you enlarge it to the largest (and most detailed size), it renders simplified. If you zoom in once again, it doesnt enlarge anymore, but shows the shop icons. This is just an idea.

In addition, I have several icon contributions:
[On my wiki user page](https://wiki.openstreetmap.org/wiki/User:MeastroGlanz)
- There should be definitely a ruins icon. I can add a ruins icon for things like houses and churches, if requested.
- tyre shops and funeral directors are one of the more frequent shops
- the perfumery icon is an alternative since the current one contains particles which affect the readability.

dieterdreist:
sent from a phone

> On 24. Apr 2017, at 13:17, MaestroGlanz <notifications@github.com> wrote:
> 
> There should be definitely a ruins icon.


-1 for the ruins icon. IMHO an icon would make ruins too prominent, there could rather be a special rendering for building=ruins.

Ruins is such a generic term that you can't determine significance of the individual feature from it, all you know is that something isn't working/functioning any more, which means that it likely isn't interesting for many people.
In case of significant ruins they are mostly better tagged as or within an archaeological site anyway

MaestroGlanz:
I agree on that. building=ruins shouldn't be rendered, because there are too many meaningless ruins. But imho building=castle;ruins=yes should be rendered that way. Most (german) maps differentiate between several types of sights (church, castles and abbeys) and have a corresponding ruins icon for every type.

HolgerJeromin:
Please open only one issue per request. Ruins/castles are handled here: #744 #331

imagico:
[...]

PR #2662, opened 2017-06-25
Adding shop=tyres rendering

kocio-pl:
Related to https://github.com/gravitystorm/openstreetmap-carto/issues/2619

shop=tyres is the last of very popular shop types - above 10k uses - which have no special icon (the other one is shop=boutique, but [it was rejected 2 years ago](https://github.com/gravitystorm/openstreetmap-carto/pull/1719)). This icon was designed by @MaestroGlanz and I find it good enough for this style:

slrr2van


kocio-pl:
Any comments/reviews about it?

daganzdaanda:
I cleaned the icon a bit and made it 12px tall to leave 1 px margin in a 14 px icon. Was that the way the other icons were sized?? It seems very tiny... https://gist.github.com/daganzdaanda/d0e644e0fa1844597f800d0aedddc372

pnorman:
It kind of looks like there's a perspective to the icon, although that could also be a tyre on the left and a rim on the right.

kocio-pl:
@daganzdaanda No, most of the icons are using 14px matrix, so you don't have to resize it.

kocio-pl:
At 14px original proposition (1) looks more readable for me than the new one (2), so I left this code unchanged:
1) tyres
2) tyres2

I think in this case it's hard to avoid perspective at all. We've used it also for optic and book icon for the same reason.

kocio-pl:
I have trouble deciding about merging this code. "A pull request should have at least one review before it is merged" (https://github.com/gravitystorm/openstreetmap-carto/issues/2436#issuecomment-291008629) - but do we need formal review ([like this one](https://github.com/gravitystorm/openstreetmap-carto/pull/2703#pullrequestreview-51651599)) or any comment indicating that other developer have seen it ([like this](https://github.com/gravitystorm/openstreetmap-carto/pull/2662#issuecomment-314222362))?

The problem is lack of review or reaction is common situation with our PRs, so decision making process is still not quite effective.

nebulon42:
For me this means a formal review.

kocio-pl:
I doubt there's any difference in how both of them apply style guidelines, but reworked icon is also acceptable for me:

aeco qrc

Issue #1928, opened 2015-10-30
Create legend


gravitystorm:
I'd like to create an automatically-generated legend for this style.

I have started the work on the [legend](https://github.com/gravitystorm/openstreetmap-carto/tree/legend) branch, but there is obviously a lot more work to do. Help welcomed. The idea is to have a control file (describing what features to show in the legend) in this repository, and use https://github.com/gravitystorm/mapnik-legendary to generate the legend images.

math1985:
Somebody created a complete (but non-autogenerated) legend on the wiki: https://wiki.openstreetmap.org/wiki/Standard_tile_layer/Key

Wuzzy2:
Yes, that was me. I almost feel bad for having wasted so much time on this in a rather inefficient manner. XD

I would definitely love to see an improved legend which is also auto-generated, and I would not be opposed at all if the wiki page gets deleted (but only in favor of the auto-generated one, of course).
The current legend on www.openstreetmap.org is very incomplete and also outdated. The wiki is more complete, but also doomed to be outdated.

The autogenerated legend should IMO be way more complete than the current one which doesn't even get all the surface colors right. Of course putting all shop symbols into the sidebar would be overkill, but here's my idea:
Basic legend with most important features in sidebar, full legend in a seperate page (also generated) which is linked to from the sidebar. I originally intended that this wiki page would be the page to be linked to, but the developers were opposed.

> The idea is to have a control file (describing what features to show in the legend)

Yes, but we should still have a full legend without any exceptions. Each symbol without explanation is a potential mystery to the map user.

It is very important for a map to have a correct and complete legend, so I think this issue should be considered high priority.

kocio-pl:
I don't believe any automated way is enough for such a complicated case as ours.

yohanboniface:
For the record, I made this time ago: https://github.com/yohanboniface/Leaflet.TileLegend
Which sounded me then the best compromise between hand made and autogenerated legends: basically you define lat/lng for each feature on the legend, and a small map of this lat/lng is shown.
It may appear that at some point some feature needs an update (eg. you pointed to an embassy but it has moved over time), but when the rendering changes there is nothing to do.
Funnily enough, four years later, the legend seems still quite up to date here: http://map.hotosm.org/

HolgerJeromin:
You can even optimize that part with a small provided SQL/OSM dump.  This would be even a nice unit test for Travis.

- import osm dump into database
- generate small tile for the legend
[...]

Issue #1287, opened 2015-02-02
Performance testing framework


gravitystorm:
Cartography is more important than performance, but performance is much harder to test! We've been heavily reliant on @pnorman to run performance tests on his setup but I'd like to open that up so that more people can become involved and that we can automate it as much as possible.

There's a whole load of hardware considerations that makes cross-server test results impossible to compare, but we should concentrate on being able to compare two different commits and give relative answers. It would be useful to have:

* A test that can be run locally and takes around 10 minutes to run. This lets people sanity-check any complex SQL / attachments / refactorings
* A test that can be run on EC2 and take less than an hour and costs less than a dollar to run. This could be used as a smoke test for pull requests.

The tests should be roughly indicative of real world rendering patterns e.g. city-biased and mid-zoom biased.

pnorman:
> A test that can be run on EC2 and take less than an hour and costs less than a dollar to run. This could be used as a smoke test for pull requests.

My full-world tests actually take <1h to run (on fast hardware). They do cost more than 1$ to run on EC2.



> mid-zoom biased

For total time rendering, it's high-zoom biased I believe.

pnorman:
> A test that can be run on EC2 and take less than an hour and costs less than a dollar to run. This could be used as a smoke test for pull requests.

EC2 is subject to about an order of magnitude performance variation between instances. For a particular benchmark configuration on amazon repeated many times on different instances configured the same, the 10% performance is 587 TPS and the 90% is 2537. 

A 200% decrease in rendering throughput would be a pretty catastrophic style change, but wouldn't be able to be distinguished on EC2.

kocio-pl:
Do we already have some tools/scripts to test the rendering speed, even roughly? I think Kosmtik automation could be nice option. We could for example craft some exporting URLs with a big enough bbox and execute them for many zoom levels automatically, but maybe there is more elegant way of doing it.

pnorman:
> Do we already have some tools/scripts to test the rendering speed, even roughly?

`render_list` with a list of tiles from production is the standard way.

> I think Kosmtik automation could be nice option.

There are enough differences it's not a great option, except for horrible performance failures

> We could for example craft some exporting URLs with a big enough bbox and execute them for many zoom levels automatically, but maybe there is more elegant way of doing it.

It's essential the workload is realistic.

The method I've used lately has been to randomly sample running queries and ignore the time spent in Mapnik.

kocio-pl:
[...]

PR #2671, opened 2017-07-04
Adding different types of towers and masts

### LINKED ISSUE #1225: Masts are not always for communication

### LINKED ISSUE #2024: Render tower/mast nodes at different zoom levels according to height if present
Hello, there.

I had an idea, but I don't know if it would be desirable or practical : the main use for communications masts and tower, IMHO, is for orientation in open field. Currently, the rendering of these features begins at the same zoom level, regardless of their height, but, for orientation purpose, a 300m tall tower is useful on a much wider area than a 25m tall one. Wouldn't it be great if the zoom level for rendering these was selected according to their height ? By making taller masts appearing at lower zoom levels than shorter ones, the displayed set of these objects would be more fitted to the use of the map at the different zoom levels.

I must add that I don't know or am not sure about the following criterias:
- which tag combinations are currently displayed as a communication tower, as the modelling of these features is quite confusing to me;
- if such differentiated rendering could be technically practical or would complexify rendering or code;
- how to render them when their height wasn't modelled — at a default zoom level? at the lower one? the higher one?
- if the height of these structures is given often enough for this differentiated rendering to be effective;
- if this idea could be useful for other features like trees or peaks.

Hope this can help to improve the rendering.

Regards.

### LINKED ISSUE #2556: Unify and clarify icons for some vertical man-made structures
In my diary entry here http://www.openstreetmap.org/user/BushmanK/diary/39981 I've described the current situation with tagging masts and towers in OSM. I already have engineering definitions for masts and towers added to `man_made=mast` and `man_made=tower` OSM Wiki pages to clarify it. Earlier, I promised to draw a set of icons for the OSM Standard style. Since there are several contradictions in current tagging scheme, I'd like to hear some feedback from developers, how you see this situation. 

So, here are six of these icons, rendered from original SVG into PNG at 200% scale:

Cantilevered (standing on its own foundation, not supported by anything else) tower `man_made=tower` `tower:construction=freestanding`
tower_cantilever
Cantilevered tower used for communication `man_made=tower` `tower:construction=freestanding` `tower:type=communication`
tower_cantilever_communication

Lattice tower `man_made=tower` `tower:construction=lattice`
tower_lattice
Lattice tower used for communication `man_made=tower` `tower:construction=lattice` `tower:type=communication`
tower_lattice_communication

Mast (guyed vertical structure) `man_made=mast`
mast
Mast used for communication `man_made=mast` `tower:type=communication`
mast_communications

Some of these icons could also be used for other combinations of tags, because, for example, all `tower:construction=guyed*` (there are three documented values) fall into a category of masts. Therefore, can be rendered as this:
mast
The same icon could be used for `man_made=mast` without any additional tags. It applies to `man_made=tower` with no other known details and this icon:
tower_cantilever

I am not touching (yet) other purposes of vertical structures (`tower:type=` values), but I can definitely draw more of them if there is any demand for it. 

If someone thinks that cantilevered tower icon is too similar to one for monuments, I can probably change it a bit, for example - make its top square and/or reduce the width of its foundation.

### LINKED ISSUE #2791: Wrong symbol for lightning masts
Lightning masts were tagged as towers. I've re-tagged them according to wiki guidelines but now OSM draws them as communication mast despite proper tower:type.

http://www.openstreetmap.org/node/3092663969#map=18/59.80008/30.27692

kocio-pl:
Resolves #2556
Resolves #2791 
Resolves #1225
Resolves #2024

This code is based on @wmyrda patch. It renders, but each type of tower/mast needs to be tested if it really appears on the map, so for now it's a work in progress.


imagico:
I think the mast and lattice tower icon are not really recognizable at normal resolution, for mast i preferred the suggestions by @BushmanK in https://github.com/gravitystorm/openstreetmap-carto/issues/2556#issuecomment-277478060 (the double guy lines version probably requires some tuning though).

I have doubts the starting zoom level logic you suggest makes sense.  Starting tall masts/towers one zoom level earlier and small ones one later makes some sense but stretching this across five zoom levels requires demonstrating that this has some specific advantages i think.

And in case of towers mapped as building polygons with both building tag and man_made=tower rendering both the building and an icon would be irritating since this will possibly look like there is a tower on the building.

wmyrda:
About the icons. Those are the only once I got from @BushmanK to test. If there will be different set present I have no objections to change them for the better types.

Levels can always change. My are just a proposition which can change if some agreement there is made for different level. Height there plays important role as it is obvious that tower with greater height is visible from further distance therefore should appear sooner.

If tower is also tagged as building than that to me seems as bad tagging which should be fixed on the map and only reason someone tagged it like was that it was the closest tag possible to have them appear on the map.

kocio-pl:
Please be as specific as possible when discussing, so I could make the changes to the code. @wmyrda - are you capable of creating proper patches? It would be much easier for me to just apply them here instead of tracking someone else's code.

kocio-pl:
@wmyrda Since you are now capable of making [your own PRs](https://github.com/gravitystorm/openstreetmap-carto/pull/2686), I suggest that you recreate this one by yourself too. It will allow you to make changes any time you want instead of waiting for me to apply patches.

Rendering examples is needed for all the tower types you propose.

math1985:
What is the status of this proposal?

kocio-pl:
I plan to take over this code after some easier problems are resolved.

kocio-pl:
This is much anticipated PR - and rightly so, because towers and masts are visible on long distance. Some icons are redundant (the same image with different name), but that can be not touched for now until we agree which we want to use.

For now just a test rendering with examples that might be useful for testing further changes:

[man_made=mast](https://www.openstreetmap.org/node/3238530878#map=18/52.21017/20.88338)
k6 uncci
[man_made=mast + tower:type=communication](https://www.openstreetmap.org/node/4687275865#map=18/52.26065/20.92644)
i8ss7 td

[man_made=tower](https://www.openstreetmap.org/node/390307980#map=18/52.28685/21.12975)
i2tvdrhq
[man_made=tower + tower:type=communication](https://www.openstreetmap.org/node/1833525929#map=18/52.22010/21.08298)
z1jddhej
[man_made=tower + tower:type=communication + tower:construction=freestanding](https://www.openstreetmap.org/node/3045463016#map=18/53.41027/20.48657)
xbcdlmw5
[man_made=tower + tower:type=communication + tower:construction=lattice](https://www.openstreetmap.org/node/4405122695#map=18/52.36906/21.13060)
wz5o kf_

`tower_cantilever.svg` and `tower_lattice.svg` are included, but not used.

kocio-pl:
Other possible shapes to render:
- tower:type=defensive - https://github.com/gmgeo/osmic/blob/master/tourism/watchtower-14.svg
- tower:type=observation - https://wiki.openstreetmap.org/wiki/File:Observation_tower.svg (can be based on https://github.com/gravitystorm/openstreetmap-carto/blob/master/symbols/water_tower.svg)

turnsole80:
This will be fantastic when you roll it out. It looks pretty good to my amateur eyes

kocio-pl:
Code update, both new types added and name labels fixed so they appear only when the object is shown.

[tower:type=defensive](https://www.openstreetmap.org/#map=19/50.79743/20.46053)
btboto5z

[tower:type=observation](https://www.openstreetmap.org/node/1541427983#map=17/52.22086/20.94910)
_sonx0cv

kocio-pl:
Currently I just cut down the initial zoom level for labels, so it always starts with the z17 (towers) and z18 (masts), because these are the zoom levels where the towers/masts are rendered even in the worst case. Do you think it makes sense to render labels as soon as the objects are visible (correlated with with `height` tag value)?

kocio-pl:
I like this observation tower shape more:
vvfcxdbi

kocio-pl:
The code is in mergeable shape now.

Done the code cleaning and fixing height selection to be numeric. I also think that names should be not revealed too early to not eclipse other objects names - it's enough to see the shape of a tower/mast early on if it's big enough (see the examples below).

Examples of really big objects on initial zoom level (actually both are telecommunication masts, but that's the current tagging):

[mast (z15)](https://www.openstreetmap.org/node/537222031#map=15/52.0666/20.8884)
allawb8k

[tower (z13)](https://www.openstreetmap.org/node/2996745920#map=13/53.1606/16.2777)
lq5jh0ll

kocio-pl:
I think `tower:type=lighting` is rather a local feature (especially a small element of the stadium), so I made an exception and special icons (based on the lighthouse icon) visible on z18+ (not earlier, unlike the rest of towers/masts). Examples on z18:

[tower](https://www.openstreetmap.org/#map=18/50.29113/21.41681)
_nyscwca
[mast](https://www.openstreetmap.org/node/4956704691#map=18/53.90885/14.27637)
cwnfeuvx

kocio-pl:
Lattice towers should be taller to be easier recognized as towers (similar to Eiffel Tower proportions). 

It seems that both generic and lattice towers were made smaller as a fix for communication waves (so the top of the tower is in the center of the rings) and generic tower without them is taller, so it also makes perfect sense design-wise.

hkjqe8wh

kocio-pl:
BTW - after fixing tagging of the Eiffel Tower it looks not too strong, even though it's the center of a big city, and introducing this icon fixes problem on z15:

z13
ulpwmvdn

z14
j8ibtdro

z15
Before
m7ict1mv

After
06qw_qdl

geozeisig:
I've made lattice towers as high as the 14 px allow, for it to be easily seen.
tower_lattice2

kocio-pl:
I've made it 13 px high, to be just like the basic tower, but visually full size (14 px) works even better, so thanks - I will probably use your solution.

kocio-pl:
Could you post a gist link to the SVG version?

geozeisig:
https://www.dropbox.com/s/tnb8522sbvjaqjj/tower_lattice.svg?dl=0

I do not know where you want to have the upload.

kocio-pl:
I meant this:

https://help.github.com/articles/creating-gists/

Please post it like this, because the link you gave is not working for me.

geozeisig:
https://gist.github.com/geozeisig/a0b21448a1c6187dcd7a4c691672ee3e

kocio-pl:
Thanks:
f3cjaw87

kocio-pl:
Added [lighting lattice towers ](https://www.openstreetmap.org/node/5347182010#map=18/50.45209/19.72598)(as you can see, there are also other towers in this place) - z18:

zza8rvqr

wmyrda:
I really like what I see. Any chance to add bell tower icon as well while You are at it? I bet You know it has over 10K uses in the database.

kocio-pl:
Thanks - I'm happy to hear it, because it's based on your work.

Yes, I was thinking about it, but I have no idea what could be the icon (just the bell could mean man_made=bell, and it's hard for me to imagine readable tower+bell shape at 14 px, so maybe general tower icon is OK?) and I'm also not sure we need to render it, since even the campanile is usually very close to the church and might produce a clutter (so maybe at z19+?).

Do you have any proposition about it?

wmyrda:
About the the look wiki page for tower have suggestion https://wiki.openstreetmap.org/wiki/Key:tower:type but personally I would rather see something like http://blog.garstor.com/?p=860 with obviously less detail to illustrate more cleanly that the object it a tower.

I agree that church itself is the main object and likely should appear sooner on the map than the bell tower, but I am not sure when it should appear. My guess is that generally it could be about two zoom levels later. Exceptions to this main rule could involve height (visibility in the area ) and heritage (importance of the object) tags which could lead object to be visible as soon as the church.

If test would show those would create to much clutter than I am fine with z19+.

wmyrda:
Looking at the code a bit I saw that amenity=place_of_worship starts to be show on z16. Most important bell_towers and as such I consider those with heritage=1 could be shown at the same level. Giving that other types of towers also may have that tag and it would be worth it to show them a bit earlier than I would simply change the code into this

+    [zoom >= 15][height > 30],
+    [zoom >= 16][heritage = 1],
+    [zoom >= 17] {

It should not clutter anything as out of 132K man_made=tower objects there is only 1K  also tagged with heritage=* tag and even less with heritage=1.

As for bell towers I themselves I propose adding following in the   [zoom >= 17] {...} section.

+      ["tower:type" = 'bell_tower'][heritage = 2],
+      ["tower:type" = 'bell_tower'][zoom >= 18] {
+        marker-file: url('symbols/towers/tower_bell.svg');  
+      }

Tomasz-W:
Bell tower icon proposal: 

tower_type bell_tower

wmyrda:
@Tomasz-W looking at those 2 examples I like first one more as there the bell is in the top part of the icon which resembles tower better than the second one which has bell much closer to the middle.

So +1 for no-roof icon from me.

kocio-pl:
I think the top and side lines could be thinner, I also feel that the bell could be bigger so its shape is more evident.

kocio-pl:
@wmyrda If you talk about the code, it's always better if you provide actual patch/branch for testing, because it's easy for me to get lost which part of code are you talking about or make some stupid error during copy-pasting.

Tomasz-W:
Thinner lines, bigger bell:

tower_type bell_tower (no-roof v2)
tower_type bell_tower daszek (roof v2)

No-roof version could be used also for `man_made=campanile` (https://wiki.openstreetmap.org/wiki/Tag:man_made%3Dcampanile)

kocio-pl:
This is better, indeed, but the bell is still not that good. Maybe adding [a clapper](https://en.wikipedia.org/wiki/Bell#/media/File:Parts_of_a_Bell.svg) on the bottom (maybe even as a small circle) might help?

Tomasz-W:
Circle-shaped clapper added:

tower_type bell_tower (no-roof v3)
tower_type bell_tower daszek (roof v3)

kocio-pl:
Thanks! 

I vote for roof v2. For some reason roofless version is harder to interpret for me (probably too simple shape - it might be drying clothes or something like shinto symbol) and clapper changes the overall bell shape too much.

Tomasz-W:
I've changed bell shape, and for me, it's more readable now. So again:

tower_type bell_tower (no-roof v4)
tower_type bell_tower daszek (roof v4)

I think that no-roof shape could be used for ` man_made=campanile` and `tower:type=campanile` (in total 1k uses). 

> it might be drying clothes or something like shinto symbol

Note, that most of these object are situated next to church or other place of worship, so context is working here.

kocio-pl:
OK, we might try that, but I have to check if campanile is something that different from all the  other bell towers. In the first reading though I haven't found any special differences other than name and that the tower is not a part of the church, but in my test they were all very close.

Shapes in v4 work for me good enough, so please post a gist with them.

Tomasz-W:
* roof:
https://gist.github.com/Tomasz-W/18d5cac547924610245808723dd61319
* no-roof:
https://gist.github.com/Tomasz-W/bcd91ee766a5256e8ad0c0dc0a025437

As I understood, bell towers are high, brick or concrete buidings, and campaniles are smaller, often wooden constructions, but I might be wrong so I'm not sticked to this opinion. We could use 'roof' version to both types, because as I see now, even the small wooden campaniles match to this shape as well.

kocio-pl:
For example Leaning Tower of Pisa is campanile (of the Duomo di Pisa), so material and size don't matter. Also:

> The Italian term campanile (/ˌkæmpəˈniːliː/; Italian pronunciation: [kampaˈniːle]), deriving from the word campana meaning "bell", is synonymous with bell tower; though in English usage Campanile tends to be used to refer to a free standing bell tower. 

[ https://en.wikipedia.org/wiki/Bell_tower ]

Also tower:type=campanile is not documented here - instead it's defined as a subtype of bell tower:

https://wiki.openstreetmap.org/wiki/Tag:tower:type%3Dbell_tower

but it's documented here:

https://wiki.openstreetmap.org/wiki/Key:tower:type

so I would stay with roof v4 and use it for both types (because both are towers for sure and would be rendered anyway) but avoid other tagging schemes than this.

kocio-pl:
Updated lattice lighting towers so they would not resemble communication radiation and be more similar to standard lighting towers:
hgra0hls

kocio-pl:
Bell towers are tall, but rather not that wide, so z18+ should be enough:

https://www.openstreetmap.org/way/437805506#map=18/52.34425/21.11702
namtc lo

https://www.openstreetmap.org/way/496449863#map=18/52.13818/21.22712
nzgoiq1y

wmyrda:
@kocio-pl wayside shires are much smaller yet they are visible from z17 like here https://www.openstreetmap.org/#map=17/49.89941/21.87712

I would say bell tower comparing to them is more important feature. Did You consider my proposition to show bell tower with heritage=2 on z17 and other towers with heritage=1 on z16?
Perhaps something similar should be done for shires?

Tomasz-W:
@kocio-pl This PR resolves also https://github.com/gravitystorm/openstreetmap-carto/issues/1225

kocio-pl:
@wmyrda Thanks for your suggestion - I was considering it, but after some time thinking about it the most important factor for me is if a given object is (typically) standalone or just a part of something bigger (and this would produce clutter). 

The problem is that it's a context related question (not related to a height or heritage type), so we have to make the decisions based on a typical cases. Wayside shrines tend to be standalone, many communication towers are standalone, but bell towers tend to be just a part of the churchyard, and the churches can be easily seen even from z15 (and spotted from z14), so showing such towers earlier would be just like spotting the nose before you can see the face.

That's why I checked that even if campanile is usually "standalone" from the architectural point of view, it's also just a few meters away from the church, so the distance is visible only on highest zoom levels and from the map perspective they are effectively not alone. This is also why I try to show lighting towers late (part of the stadium or industrial zone) and the same with toilets (they are usually a part of cafe, museum or camping place - see #3055).

wmyrda:
@kocio-pl ok sounds reasonable to me.

wmyrda:
Any reason this code has not made it yet into master branch? Personally for it I vote with both hands ;)

kocio-pl:
The last thing left is tagging related - we don't know which bell tower/campanile scheme should be used and which one would be discouraged:

https://lists.openstreetmap.org/pipermail/tagging/2018-March/035389.html

wmyrda:
For me it is clear it should be man_made=tower + tower:type=campanile as man_made=tower is way more known and used. Besides if someone does not even know/use tower:type=campanile tower would also show on the map as plain tower if man_made=tower is used.
All arguments about tagging in discussion at the list seems to also agree with that.

polarbearing:
It would not make any sense to  me to treat man_made=campanile separately. The tagging discussion analysed:
man_made=tower and tower:type=bell_tower (10717)
man_made=campanile (798)
The wiki page for the latter was created 2012 without a proposal and received the immediate comment that  tower:type=bell_tower was already used more often that time. I had added the structured alternative in 2016 which was undisputed. I'll make it stronger now. 

As for the tower:type, my recommendation would be to render bell_tower only as it is heavily used, and leave the subtypes to subtagging, unless somebody strongly explains why a klockstapel or campanile is not a bell_tower.

kocio-pl:
OK, thanks a lot for the comments! 

So I will remove `["tower:type" = 'campanile'][zoom >= 18]` rule - this will be shown as a plain tower then, so the mapper knows something is wrong and hopefully will check the wiki. I haven't used `man_made=campanile`, so I don't have to remove it. :smile: I will merge the code soon after that.

Please update all the relevant wiki pages so this should be perfectly clear for the community what is the correct tagging.

kocio-pl:
Congratulations - this was a team work at its best: 
- thanks @BushmanK for the first draft of solution
- thanks @wmyrda for the basic implementation
- thanks @Tomasz-W for additional bell towers design and discussion about tagging campanile on Tagging list
- thanks @geozeisig for tuning a lattice tower icon 
- thanks @polarbearing for the final tagging comments
- thanks @turnsole80 for encouraging

I'm also proud of my own work on this complex issue (#brawoja! :smile: ). However even if I did check the renderings many times, it's always possible that some kind of bug slipped in, so please test it now independently as much as you can. As the next release is expected on 23.03 (see #3095), we have plenty of time to make sure everything is OK and maybe fix some details.

Thanks all for the collaboration!

geozeisig:
tower: type=lighting should be used together with man_made=mast. It's often used with man_made = tower, but in my eyes this  are the tagging errors. And for man_made = lighthouse we have our own icon. This leads to the assumption that the File:Tower_lighting.svg is superfluous. You can use [Overpass ](http://overpass-turbo.eu/s/wZE)to search for it.

kocio-pl:
You might be right, but this is a case for discuss on Tagging list first. If nobody knows lighting towers different than lighthouse, the wiki pages should be updated and I will update this code too.

geozeisig:
On the [wiki page](https://wiki.openstreetmap.org/wiki/Tag:tower:type%3Dlighting) is with intention the man_made=tower not mentioned. If nobody has a counterexample, I would prefer to leave it that way. It would be nice if the wrong tagging did not render. We prefer to correct the databas.

polarbearing:
Excellent to have this sorted :-)

A minor question, probably not relevant for the function:
When going through the SVG files I noticed that tower_defensive.svg is different from the other as it has in the first svg section ```width="100%" height="100%"``` where the others have ```width="14" height="14"```, which causes git to show it larger. 

Another question are the metadata and URLs left in the SVG code by the various editing tools which are probably not necessary. I wonder if the number of redundant lines has any impact on general performance. If it has, we might create guidelines which of them can be stripped.

polarbearing:
counterexample tower+lighting: https://de.wikipedia.org/wiki/Lichtturm
A former water tower has been reconstructed to allow lighting performances.

kocio-pl:
@geozeisig OK, I see. Would you however ask on the list if this could be valid case, just to be sure that no one heard about such beast as lighting tower not being lighthouse? _Lichtturm_ might be a valid case for this tagging, so it's worth mentioning, but of course there might be different scheme which we could use for such cases. It's all on the tagging side currently.

@polarbearing We lack the tools for measuring performance (see #1287) - both code wise (this one is too general: #1941) and deployment wise (https://github.com/openstreetmap/chef/issues/129 - however this can be hard to measure, as we have seen on https://github.com/gravitystorm/openstreetmap-carto/pull/2874#issuecomment-334507450). I guess this would be best if somebody with better SVG knowledge review all the icons we have (and maybe write a tool to sanitize them) than only these ones.

SomeoneElseOSM:
> You might be right, but this is a case for discuss on Tagging list first. If nobody knows lighting towers different than lighthouse, the wiki pages should be updated and I will update this code too.
I did have a look at usage in the UK, and there's a clear split between "lighthouses" and "lighting" (and "illumination") towers.  The tagging reflects the on-the-ground difference.

geozeisig:
The _Lichtturm_ is a tower light falls into it and no light is emitted, as the symbol suggests, if I understood it right. So you can not look at him as an exception.

kocio-pl:
Whatever we can think of it, it's still basically a tagging problem, so please come back with an agreement about tagging so we could proceed with rendering.

geozeisig:
The table for [tower: construction](https://wiki.openstreetmap.org/wiki/Key:tower:construction) is new. Please have a look at it whether everything is correct and synchronously to rendering.

The table for [tower: type](https://wiki.openstreetmap.org/wiki/Key:tower:type) is also new. Please have a look at it whether everything is correct and synchronously to rendering.

kocio-pl:
In case of the `tower:type` it looks good, `tower:construction` is also as I expected, but we lack `dish` and `dome` icons. There are also no specific cross-case icons (like lattice+communication tower) on these pages.

wmyrda:
@geozeisig while looking at the wikis You provided I like how practically all of those will be shown on map! 
It kinda made me wonder if it would be possible to add rendering for tower:construction=dish and what would happen if some also use for it tower:type=communication which might also be true for the dish.

geozeisig:
Of the tower: type that was still tagged in larger numbers and still have no icon, there is still [cooling tower](https://wiki.openstreetmap.org/wiki/Tag%3Atower%3Atype%3Dcooling) (1146x).

kocio-pl:
I think lighthouse is for signaling (kind like a road sign for boats), not lighting, so lighting tower is something different in function. Repurposing of lighthouse for lighting would make a lighting tower for me.

I guess we can add `man_made=communications_tower` (4 014 uses and a wiki page) with rendering just like other >100 m communications tower - z13+ (see https://github.com/gravitystorm/openstreetmap-carto/issues/1928#issuecomment-373720108). I believe this is useful when the mapper doesn't know exact height, but know that it's big.

kocio-pl:
@Tomasz-W Could you try to design icons for dish, dome and cooling towers?

wmyrda:
For  tower:construction=dish & tower:type=communication I would propose something like
https://www.svgrepo.com/svg/67469/satellite-dish
where for for plain  tower:construction=dish without tower:type I would use the same just without waves representing signal. After all not all dishes are for satellite uplinks, but may also be used as radio telescopes.

geozeisig:
Maybe we should use an icon with a platform for `man_made=communications_tower`  like this one
communications_tower
[communications_tower.svg](https://gist.github.com/geozeisig/04842a66d8e450c0082e8262035be1a3)
See also [Tag:man_made=communications_tower](https://wiki.openstreetmap.org/wiki/Tag:man_made%3Dcommunications_tower) that explains the difference to the tower and mast.

kocio-pl:
It looks like a cross, so would be hard to recognize. I also don't see the difference between it and other communications towers - height is the only such thing - so there's no use to add new shape.

geozeisig:
@kocio-pl You're right with the cross, so leave it at tower_generic.

Tomasz-W:
@kocio-pl 
* dish
tower_type dish
* dome 
tower_type dome
* cooling tower
tower_type cooling_tower

geozeisig:
dome is not yet intuitively recognizable. dish and cooling tower are good.

dieterdreist:
sent from a phone

> On 18. Mar 2018, at 08:33, geozeisig <notifications@github.com> wrote:
> 
> dome is not yet intuitively recognizable. dish and cooling tower are good



also because a tower is usually higher than wide.

kocio-pl:
- I would like the dish to have a bit bigger foot and smaller dish part
- I can't recognize dome
- cooling tower could have something flowing from the top - maybe haltoned

wmyrda:
- I agree with kocio that dish could be a bit smaller, but not to much
- Its going to be hard to have better icon for dome, as this is very good work, but at the same time question is do we even need it. After all no dome is created to be just a dome but to have conference/concert/sport hall inside or many other uses that could be separately shown as the icon. Similar case would be shop=mall
- cooling tower i ok, but somehow its look for me is too heavy. I think would a bit better if we could add
a small oval with lighter color like in these example https://ibberson.com/files/cache/6471fb90d18207b3a78a2fada6119331_f1884.jpg to clearly show the object is a chimney that is empty inside and it is not big concrete block.

Tomasz-W:
* smaller dish 
tower_type dish
* dome without white lines
tower_type dome black
* cooling tower with white oval shape upside
tower_type cooling_tower2

kocio-pl:
- dish is good for me now, thanks
- dome still looks like a generic geometry (what is half of the dot?)
- this cooling tower breaks the general rule of a flat view (we occasionally do it), it's easier to recognize though; what do you think about smoke-like shape on the top?

wmyrda:
Before I proposed oval shaped I was thinking of proposing to add smoke to the icon, but then not every cooling tower is currently in use which might be somewhat misleading to the viewer.
- ex. https://www.openstreetmap.org/way/28532588

Hence it was my proposition it would somewhat biased but I must say I like current cooling tower look.

kocio-pl:
This is true for any object, like communication dish not being used, but we don't render any current state, just the general idea what it is typically to be easily recognized.

kocio-pl:
@Tomasz-W Speaking about cooling tower - what about `man_made=chimney` icon? It's close to the towers, especially this subtype.

Tomasz-W:
* dish   https://gist.github.com/Tomasz-W/79c688e5db858058986736382d344878
* cooling tower  https://gist.github.com/Tomasz-W/16606479ccaefbe7e553b9410eb0413a
* dome -> I was thinking you mean ground-based dome (I didn't know that there are dome-shaped towers), I'll try to make a tower-like dome icon
* chimney -> I'll try

kocio-pl:
I was not aware of the dome type tower, but photos on the wiki page shows quite recognizable examples: 
https://wiki.openstreetmap.org/wiki/Tag:tower:construction%3Ddome

kocio-pl:
Cooling tower ideas:
1. tower_cooling-1
2. tower_cooling-5
3. tower_cooling-6

dieterdreist:
sent from a phone

> On 20. Mar 2018, at 15:23, kocio-pl <notifications@github.com> wrote:
> 
> I was not aware of the dome type tower, but photos on the wiki page shows quite recognizable examples:
> https://wiki.openstreetmap.org/wiki/Tag:tower:construction%3Ddome
> 


the dome is typically covering antennas, but is this a tower type or a feature of a tower?

geozeisig:
> I guess we can add man_made=communications_tower (4 014 uses and a wiki page) with rendering just like other >100 m communications towers - z13+

Please add code for `man_made=communications_tower`. It should be rendered in the same way as `man_made=tower`.

kocio-pl:
I plan to: https://github.com/gravitystorm/openstreetmap-carto/pull/2671#issuecomment-373737398, I just wait for the rest of tower types propositions to settle down and then do a "part 2" of this PR.

Tomasz-W:
* for cooling tower I find my icon more readable (as someone has said - these towers don't emit steam/ smoke all the time), comparsion:
 (Tomasz-W)
   (Kocio-pl)

* dome tower:
 tower_type dome
* chimney:
 man_made chimney

kocio-pl:
- I like your cooling tower version and it may be worth to break perspective rule here
- this dome tower might be enough
- chimney is great for me

dieterdreist:
cooling tower and chimney work for me, but the dome tower could be "fatter"
and less high (bigger dome and shaft), see this picture for proportion
reference:
https://commons.wikimedia.org/wiki/File:2014_Putgarten_Radarstation_01.jpg

Tomasz-W:
Gist links:
* cooling tower: https://gist.github.com/Tomasz-W/16606479ccaefbe7e553b9410eb0413a
* chimney: https://gist.github.com/Tomasz-W/5b82f20209e3093dcd77a74d414c4989
* dome v2: https://gist.github.com/Tomasz-W/6188ed1c901aebe84222cf2780ca394d
tower_type dome

wmyrda:
If someone asks me I say those last 3 examples are ready to be merged, but at the same time I wonder if we could take there in the account material hence that dome towers vary a lot in their support structure.

If material=!metal than prepared example for dome would be like it is dome v2 example.
If material=metal than could be more like https://atmos.washington.edu/~cliff/NewRadarPics/radar.up.close.22.jpg and just like our https://github.com/gravitystorm/openstreetmap-carto/blob/master/symbols/hunting_stand.svg have two smaller thin straight columns instead single larger one.

It does add additional layer of complication which may be to much for general purpose map, but maybe its worth a fought as it adds realism to it.

kocio-pl:
I don't think we should really try to differentiate basing on materials, having different shapes is enough in my opinion.

geozeisig:
dome v2 might be somewhat misleading to the viewer in me eyes. [tower:construction=dome](https://wiki.openstreetmap.org/wiki/Tag:tower:construction%3Ddome)  is also not very common with 129x.

geozeisig:
I miss `man_made=communications_tower` in v4.9.0...master.

geozeisig:
`tower:construction=dish` merged with `man_made=telescope`
We should think about an icon for `man_made=telescope` because that describes the feature better.

dieterdreist:
sent from a phone

> On 24. Mar 2018, at 07:26, geozeisig <notifications@github.com> wrote:
> 
> tower:construction=dish merged with man_made=telescope
> We should think about an icon for man_made=telescope because that describes the feature better.


a telescope doesn’t have to have a dish, it is simply an instrument to look into the distance

kocio-pl:
I'm satisfied with the result, but feel free to redesign it if you want.

geozeisig:
@dieterdreist A telescope can be optical but there is also radio telescope (see [Wiki](https://wiki.openstreetmap.org/wiki/Tag:man_made%3Dtelescope). Otherwise the tagging would be wrong [here ](https://www.openstreetmap.org/node/3868969567)? We do not have a good wiki page for dish yet. On the other hand, a dish is not a real tower.

kocio-pl:
There is a handy tag called [location](https://wiki.openstreetmap.org/wiki/Key:location) - I'd like to use it for example for hiding communication masts mounted on the building, since their height is counted with the whole building, which might be weird and misleading, because they are not that important and visible, like here: 

https://www.openstreetmap.org/node/2471572605#map=14/52.2543/20.9838

Interesting value seems to be `location=rooftop` ("Located on top of a building that is used for something else."). I'm not sure about `location=roof` ("Used where a feature is mounted on the roof of a building."), because this might be a legitimate mast located on a small building under it, but I suspect it's safer to hide them too just in case.

AllroadsNL:
The height of the building roof should be mentioned in 
[min_height](https://wiki.openstreetmap.org/wiki/Key:min_height) for the mast

kocio-pl:
Is it documented somewhere? I don't remember seeing it on tower/mast pages.

AllroadsNL:
> Is it documented somewhere?
I could not find it either, seems logical to me. Therefor "should be". Watching 3D buildings.

When a height is mentioned for the antenna/transponder.
Often the extern data gives only height for [mobil_phone](https://wiki.openstreetmap.org/wiki/Key:communication:mobile_phone) mounted height, not total length of tower/mast, there often multiple on the mast/tower. (Not good described, looks like multiple traffic_sign problem, which method to use)

So the mast/ tower should get a min_height but also this transponder a min_height.
pitfall:
But when min_height is used on a building node, it means also that the building get a strange shape.

Side of building.
About mounting: look at
mount_to/mounted_on,  power_tower, building (not documented)
[lamp_mount](https://wiki.openstreetmap.org/wiki/Key:lamp_mount) wall, wall_mounted, wall_mount

A mast or is it a pole for some communication:mobile_phone?
[flagpole](https://wiki.openstreetmap.org/wiki/Tag:man_made%3Dflagpole)
Also the same min_height strategy, needed?

There must be a tag, that says not on ground level. As always in OSM there multiple tags that indicate it.
But which is the best?

kocio-pl:
Thanks for the research.

I think that min_height for masts and towers should be discussed on Tagging list and when accepted (I don't expect any problems with this - it's just natural consequence of tagging "brutto" height that it should be able to calculate "netto" height) documented on wiki. Could you do this, please?

But since it's not yet done and requires calculating in SQL, I prefer to start with simpler approach and use location. It won't hurt if we skip rendering few masts mounted on roofs but it may clutter urban areas when mapped in detail if we don't know exact height of the building for example, but it's clear that these are not standalone masts.

dieterdreist:
2018-03-26 16:28 GMT+02:00 AllroadsNL <notifications@github.com>:

> Is it documented somewhere?
> I could not find it either, seems logical to me. Therefor "should be".
> Watching 3D buildings.
>
> When a height is mentioned for the antenna/transponder.
> Often the extern data gives only height for mobil_phone
> <https://wiki.openstreetmap.org/wiki/Key:communication:mobile_phone>
> mounted height, not total length of tower/mast, there often multiple on the
> mast/tower. (Not good described, looks like multiple traffic_sign problem,
> which method to use)
>
> So the mast/ tower should get a min_height but also this transponder a
> min_height.
>

a min_height for mast or towers does not make sense IMHO (unless they are
on top of something, but not if they are on the ground). For antennas and
transponders that are attached to another structure it is different of
course.



> pitfall:
> But when min_height is used on a building node, it means also that the
> building get a strange shape.
>

min height means the height of something is not what is indicated in the
height tag but actual height = height - min_height with height + ele =
absolute height

dieterdreist:
2018-03-26 12:50 GMT+02:00 kocio-pl <notifications@github.com>:

> There is a handy tag called location
> <https://wiki.openstreetmap.org/wiki/Key:location> - I'd like to use it
> for example for hiding communication masts mounted on the building, since
> their height is counted with the whole building, which might be weird and
> misleading, because they are not that important and visible, like here:
>
> https://www.openstreetmap.org/node/2471572605#map=14/52.2543/20.9838
>
> Interesting value seems to be location=rooftop ("Located on top of a
> building that is used for something else."). I'm not sure about
> location=roof ("Used where a feature is mounted on the roof of a
> building."), because this might be a legitimate mast located on a small
> building under it, but I suspect it's safer to hide them too just in case.
>


there is also location=roof I would see it as a synonym of rooftop and
would not require "used for something else" for this tag).

geozeisig:
@dieterdreist 

> a telescope doesn’t have to have a dish, it is simply an instrument to look into the distance

Optical telescopes are almost never outdoors. They are operated under a dome, tagged with `man_made=observatory` . In addition, it is still possible to add: `telescope:type=optical`, `telescope:diameter` etc

For `man_made=telescope` + `telescope:type=radio`, the icon dish should also be used.
Just like for `tower:construction=dish`
dish

turnsole80:
It occurs to me that dome v2 image is very similar to the weather radars used here in Australia. It would useful for describing those at the very least

kocio-pl:
What is the tagging used for them?

kocio-pl:
It looks to me that it's just a dome type construction, the whole station should be tagged as:

https://wiki.openstreetmap.org/wiki/Tag:man_made%3Dmonitoring_station

See this picture:

https://wiki.openstreetmap.org/wiki/File:Mildura_Airport_Weatherstation.jpg

kocio-pl:
I think this one should be rendered the same as tower:type=observation:

https://wiki.openstreetmap.org/wiki/Tag:tower:type%3Dwatchtower

geozeisig:
Do you mean [Tower observation.svg](https://wiki.openstreetmap.org/wiki/File:Tower_observation.svg)?
`tower: type=watchtower` does not have an icon yet. It is possible that watchtower should have the same icon as observation.

turnsole80:
> What is the tagging used for them?

It's a higgledy-piggledy mess unfortunately. A lot of them aren't (yet) tagged at all - something i'm working on now.

To me, these rainfall radars should be tagged as `man_made=tower` + `tower:type:radar`. I would have thought that `man_made=monitoring_station` was better suited to ground based weather stations that record temperature, wind, rain, humidity, etc - although there's a addable preset in JOSM that would tags these as `man_made=observatory` + `observatory:type=meteorological`, so i'm a bit confused at the moment as what to do

turnsole80:
Scrub that. They're tagged `man_made=tower` + `tower:type=radar` + `tower:construction=dome`

kocio-pl:
I have created new PR with the proposed changes, please discuss all the things related to vertical structures there.

Issue #2203, opened 2016-06-28
please render place=square


flohoff:
place=square is a nice way to give names to complicated squares. Please render a label for the node or centroid of the area the place tag is on. 

Rendering should be like place=locality for example.

matkoniecz:
Used only 230 times - http://wiki.openstreetmap.org/wiki/Tag:place%3Dsquare

Documentation page at wiki has no examples of cases where using it is truly useful and superior to tagging highway=pedestrian + area=yes multipolygon.

flohoff:
How do you tag places which are NOT pedestrian areas?

matkoniecz:
> How do you tag places which are NOT pedestrian areas?

Never encountered one, that is why I complained about missing examples.

flohoff:
I have a couple of them - Either beeing a combination of village_green and some ways, or even parking areas with some ways inbetween.

For some i used place=locality as that works for addresses and gets rendered. Now somebody mentioned place=square and that is a much better match. But it does not show up on the map. So its a step back.

jojo4u:
Here you are: https://www.openstreetmap.org/way/3933618

Schlossplatz encompasses highway=footway, area:highway, landuse=grass, historic=memorial, fountains

matkoniecz:
> Here you are: https://www.openstreetmap.org/way/3933618

Please, add tag documentation to wiki. This issue tracker is not a good place to document/discus tags.

dieterdreist:
2016-06-28 14:05 GMT+02:00 Mateusz Konieczny <notifications@github.com>:

> Please, add tag documentation to wiki.



there is documentation:
http://wiki.openstreetmap.org/wiki/Tag:place%3Dsquare

kocio-pl:
@matkoniecz Is there anything more missing or we could try making PR for this tag now?

pnorman:
I can't see rendering a feature with so few uses.

math1985:
I agree.

dieterdreist:
sent from a phone

> Il giorno 26 lug 2016, alle ore 08:58, Paul Norman <notifications@github.com> ha scritto:
> 
> I can't see rendering a feature with so few uses.


I suggest to keep this issue open as the tag is brand new and I expect usage numbers to go up soon

polarbearing:
[...]

PR #2673, opened 2017-07-05
Adding place=square name rendering

kocio-pl:
Resolves https://github.com/gravitystorm/openstreetmap-carto/issues/2203.

Rendering the same as other `roads_area_text_name` objects.

Warsaw, z16 - two squares which are not simple footway or pedestrian areas:
Before
6juh9 ku
After
vge2zepp

### DIFF
--- a/project.mml
+++ b/project.mml
@@ -1871,10 +1871,14 @@ Layer:
             way,
             way_area/NULLIF(!pixel_width!::real*!pixel_height!::real,0) AS way_pixels,
             highway,
+            place,
+            leisure,
             name
           FROM planet_osm_polygon
           WHERE highway IN ('residential', 'unclassified', 'pedestrian', 'service', 'footway', 'cycleway', 'living_street', 'track', 'path', 'platform')
             OR railway IN ('platform')
+            OR (place IN ('square') 
+                AND (leisure IS NULL OR NOT leisure IN ('park', 'common', 'recreation_ground', 'garden')))
             AND name IS NOT NULL
           ORDER BY way_area DESC
         ) AS roads_area_text_name


kocio-pl:
The question is how to render squares which are tagged as points? This is a valid tagging according to Wiki (some shapes would be hard to define) and it's about 10% of all the cases.

imagico:
[place=square](http://wiki.openstreetmap.org/wiki/Tag:place%3Dsquare) currently does not have a language independent definition.  It is currently defined as things that are named in a certain way in English, French, German, Italian, Spanish and Russian - which all mean at least slightly different things.  The wikipedia article linked to from the wiki has a definition that is incompatible with the name based definition in OSM and practical use of the tag.  I think this disqualifies it from being rendered here at the moment.

As said before i would be against text only rendering of areas based on way_area for features where there is usually no verifiable extent because that will likely lead to mappers applying a very wide interpretation of the extent of a particular square they consider important.

kocio-pl:
I was thinking of square as a missing element of our road system and similar enough to simple pedestrian squares, but maybe moving it to placenames makes more sense technically.

I agree with you that definition on wiki should be improved (probably with something like "open space common in urban centres, like cities or towns"). For me however this is not a blocker for rendering this tag - currently parts of the complex squares are tagged with a name. Sometimes it makes sense (main road which changes name on square boundaries), but it's also a common workaround just to render it somehow (disconnected pedestrian areas or even [parking aisles](http://www.openstreetmap.org/way/147158051)). I find it to be worse problem (breaking intuitive rule _One feature, one OSM element_) than vague square definition.

Circeus:
In this specific example, it seems to be these are functionally parks (grass+trees+walkable), and there is no reason for the black rendering other than people not wanting a green-colored label on something without "park" in the name. The example is so much seeking to properly tag any sort of plaza so much as an obvious, roundabout (hah!) fix for issue #1253 on these two roundabouts. I.E. it's just tagging for the renderer.

Are there any better better example of things using it Place=plaza that clearly cannot be tagged in a different way? Because to me that tag looks badly out of place in that key to begin with.

kocio-pl:
I think you're missing the complexity of this square (I guess you talk about [Plac Inwalidów](http://www.openstreetmap.org/way/442100863)): it consists of two "functional" parks, two one-way roads separated with a grass area, tram area and a road around it (BTW it's not a full circle, so not a roundabout) with footways.

kocio-pl:
> I see no reason whatsoever not to define Plac Teatralny as highway=pedestrian area=yes.

Only [part of a square](http://www.openstreetmap.org/way/147158066/history) is pedestrian area (and as you can see it was tagged with a name once) - but there is also similar one (currently not drawn) on the front of the theater. The rest of the square is a parking space and roads, not pedestrian area.

Circeus:
> it consists of two "functional" parks, two one-way roads separated with a grass area, tram area and a road around it (BTW it's not a full circle, so not a roundabout) with footways.

There's the reluctance to use relations I was talking about! Take "Park" A and "Park" B (you call them "grass areas", I say they clearly function as parks), combine them into a relation and BINGO Plac Invalidow is showing. 

_Everything else is nothing but hair splitting._

- Don't derails this with semantic argument about the roundabout. You know _exactly_ what I meant.
- The road are already name tagged, yet somehow you are not happy. Why? Because the result is no area tag, My solution (I think) should provide that. Whether or not that area specifically encompasses the (already named) surrounding road is irrelevant for all purposes anyway.
- From the point of view of mapping this and getting the name to show, the ontological argument as to whether or not the trams are "part of" the Plac or just located at and named after it is completely irrelevant as it impacts neither the routing nor the display. 
- You absolutely need those footpaths part of the Plac? adjust the park so it surrounds them instead of encompassing only the grass and tree. Duh. (Personally I wouldn't tag various bits the way you have, but I'm not gonna start an argument over that.)

dieterdreist:
sent from a phone

> On 5. Jul 2017, at 10:22, Christoph Hormann <notifications@github.com> wrote:
> 
> place=square currently does not have a language independent definition



as it is about a toponym I think it is correct to be language dependent.

dieterdreist:
sent from a phone

> On 5. Jul 2017, at 16:00, Circeus <notifications@github.com> wrote:
> 
> The road are already name tagged, yet somehow you are not happy. Why?


because a square is not a road. Streets and squares together form the open space of settlements, honestly I find it disturbing we even have to discuss this here, I would have expected it is obvious to anyone interested in (urban) geography.

Circeus:
> because a square is not a road. Streets and squares together form the open space of settlements, honestly I find it disturbing we even have to discuss this here, I would have expected it is obvious to anyone interested in (urban) geography.

Thank you for derailing this with a tiny, irrelevant point that ignores my actual argument and dumping. The condescension is a nice touch, too. </sarcasm>

kocio-pl:
@Circeus: You've made a lot of interesting statements about how to tag squares. But it's hard to keep up since you expand already written entries, so I get lost eventually. More importantly - this is not the right place to do discuss square definition, let's keep on talking on Tagging list where it belongs. Currently relation is not allowed for squares on Wiki, but this tag needs more care anyway, so we can discuss also this point.

In this ticket I'd like to not check whether particular example is tagged as it should be, but rather how to achieve rendering both for points and areas tagged with `place=square` assuming the tagging was right.

math1985:
What is the status of this proposal?

kocio-pl:
Waiting for definition fix. Wiki page should be probably changed to be clear that this is about toponym, not the land features. I'm currently working with easier problems, but plan to take care about wiki.

kocio-pl:
Just a note: discussion in #2816 might lead to some general conclusions about `place=*` tag, so I prefer to wait for the outcome.

polarbearing:
7.5k now
Wiki definition improved. 
I don't see a general conclusion about the place tag necessary to start rending this value. In case there is a general conclusion eventually, this case could be simply included.

dieterdreist:
2017-10-30 23:38 GMT+01:00 polarbearing <notifications@github.com>:

> 7.5k now
> Wiki definition improved.
> I don't see a general conclusion about the place tag necessary to start
> rending this value. In case there is a general conclusion eventually, this
> case could be simply included
>


+1, please start rendering this, the tag was overdue, is rapidly gaining
attraction even if not rendered, and is not disputed in any way.

kocio-pl:
Now that the definition is fixed, I plan to do it soon.

kocio-pl:
I have updated the code to omit parks and the likes, which are sometimes called a "square".

polarbearing:
Hooray!
Just for understanding the code, that means when an object is tagged as park and as square, the label style for park is used?

kocio-pl:
Yes, exactly.

d3d9:
So why exactly did you decide to not render names for points?

kocio-pl:
I don't remember the details, but that was certainly not a decision to avoid rendering points. Thanks for spotting it though - would you like to prepare the missing code?

d3d9:
I have never done something like this before, I would have to try things out first.
How do you suggest doing it? Should it be a new layer in placenames.mss or can I make use of something existing in roads.mss or somewhere else? Or is there something obvious that I'm missing?

kocio-pl:
I don't know exactly. My advise id to start with general technical description and setting testing environment first and then play a bit with a code:

https://wiki.openstreetmap.org/wiki/Standard_tile_layer

Let me know if you need some more help.

polarbearing:
Does points mean place=square on nodes?

kocio-pl:
Yes, this is what I understand.

d3d9:
I tried this https://github.com/d3d9/openstreetmap-carto/commit/339b3b85cc7bd69d6ff57e9e76f6d4e3df15323e, how is it?

z17
square-z17
z18
square-z18
z19
square-z19

Data (4 objects with place=square):
square-data

kocio-pl:
I had no time to look closer, but on the first sight on the code and renderings it seems like a proper solution. Could you make a PR?

PR #2674, opened 2017-07-05
Moving smaller airports to higher zoom levels

kocio-pl:
Resolves https://github.com/gravitystorm/openstreetmap-carto/issues/1143.

Simply moves aerodromes with access=private or lacking any of ICAO/IATA codes from z10-z14 to z11-z14. 

I was not able to properly add `aerodrome:type` tag to `project.mml` queries, but it would be good to filter also private types here. Still the effect is good enough for me:

**Full London area from Mapzen extract at z10** (click to see unscaled images):
Before
clyq8fnf
After
najltgcp

**Full Indiana area from Geofabrik extract at z10** (click to see unscaled images):
Before
rwgimflt
After
swibtmg_

### DIFF
--- a/amenity-points.mss
+++ b/amenity-points.mss
@@ -890,7 +890,8 @@
     marker-fill: @airtransport;
   }
 
-  [feature = 'aeroway_aerodrome'][zoom >= 10][zoom < 14] {
+  [feature = 'aeroway_aerodrome']['access' != 'private']['icao' != null]['iata' != null][zoom >= 10][zoom < 14],
+  [feature = 'aeroway_aerodrome'][zoom >= 11][zoom < 14] {
     marker-file: url('symbols/aerodrome.12.svg');
     marker-placement: interior;
     marker-clip: false;
@@ -2040,7 +2041,8 @@
     text-wrap-width: @standard-wrap-width;
   }
 
-  [feature = 'aeroway_aerodrome'][zoom >= 10][zoom < 14] {
+  [feature = 'aeroway_aerodrome']['access' != 'private']['icao' != null]['iata' != null][zoom >= 10][zoom < 14],
+  [feature = 'aeroway_aerodrome'][zoom >= 11][zoom < 14] {
     text-name: "[name]";
     text-size: @standard-text-size;
     text-fill: darken(@airtransport, 15%);
--- a/project.mml
+++ b/project.mml
@@ -1581,6 +1581,8 @@ Layer:
             tags->'denomination' as denomination,
             tags->'generator:source' as "generator:source",
             tags->'power_source' as power_source,
+            tags->'icao' as icao,
+            tags->'iata' as iata,
             CASE WHEN shop IN ('supermarket', 'bag', 'bakery', 'beauty', 'books', 'butcher', 'clothes', 'computer', 
                                'confectionery', 'fashion', 'convenience', 'department_store', 'doityourself', 'hardware', 'fishmonger', 'florist', 
                                'garden_centre', 'hairdresser', 'hifi', 'ice_cream', 'car', 'car_repair', 'bicycle', 'mall', 'pet', 
@@ -1661,6 +1663,8 @@ Layer:
             tags->'denomination' as denomination,
             tags->'generator:source' as "generator:source",
             tags->'power_source' as power_source,
+            tags->'icao' as icao,
+            tags->'iata' as iata,
             CASE WHEN shop IN ('supermarket', 'bag', 'bakery', 'beauty', 'books', 'butcher', 'clothes', 'computer', 
                                'confectionery', 'fashion', 'convenience', 'department_store', 'doityourself', 'hardware', 'fishmonger', 'florist', 
                                'garden_centre', 'hairdresser', 'hifi', 'ice_cream', 'car', 'car_repair', 'bicycle', 'mall', 'pet', 
@@ -2038,6 +2042,8 @@ Layer:
             access,
             name,
             tags->'operator' as operator,
+            tags->'icao' as icao,
+            tags->'iata' as iata,
             ref,
             way_area,
             CASE WHEN building = 'no' OR building IS NULL THEN 'no' ELSE 'yes' END AS is_building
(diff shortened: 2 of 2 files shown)


kocio-pl:
It's a low priority request, since this is not v3.x compatible code, so it probably has no chance to be merged before at least v4.2, but I'd like this PR to be reviewed (or we should reconsider [our new review policy](https://github.com/gravitystorm/openstreetmap-carto/issues/2436#issuecomment-291008629)).

imagico:
> Can we not propose a commercial=yes/no tag snd rely on that one?

Such an ill defined tag directly used for a major rendering decision just cries for abuse.  

I am unsure about if this change is a good idea but it is certainly not the worst possible approach.  Making some effort to add missing ICAO/IATA codes together with such a change would definitely be advisable - there are many remote airports where this is missing.  Maybe the wikipedia/wikidata guys could use this opportunity to do something more useful than adding arbitrary language name tags :wink:.

math1985:
In my understanding, commercial aviation is a formal term (as opposed to general aviation, which might also operate commercially). But I might be mistaken. Would be best to have an aviation expert helping us. Anyway that's something gor tagging@.

kocio-pl:
Thanks for your responses!

I was thinking for a long time what features could make some airports "major" and found no such thing. I was proposing "range" (like international/regional), but IIRC that was rejected. The closest thing to determine how important it could be is [aerodrome:type](https://taginfo.openstreetmap.org/keys/aerodrome:type). It has no wiki definitions, but here we have "private", which probably is close to "commercial" (just guessing). At least we already treat private parkings as less visible, so this would follow the same rule. 

BTW: could anyone of you tell me how to add aerodrome:type to our SQL queries? It didn't work the same as iata and icao tags, probably because of the colon. I think access=private still makes sense, but the type is more airport related and quite popular, so I'd like to use it.

imagico:
My long standing suggestion has been to use the runway length.  This is a fairly good and objective measure of importance - even if military airports with long runways are not significant for travelers they are usually still important  for the map in general.  The real charm of this would of course be when using relative importance based display (see http://www.imagico.de/map/osm_populated_en.php) which would only be feasible with preprocessing.

`aerodrome:type` is undocumented and a bad choice of tag as evidenced by the use of values like `military/public`, `private/public` and `military/private`.  For military airports we have the established `military=airfield`/`military=yes`.

kocio-pl:
I remember this proposition and also that it was rejected too with some examples. What's more important, I don't think we have such tag currently in use, let alone being documented - or am I wrong?

imagico:
Runway length would not be tagged, it would be determined from the separately mapped runway via spatial query.

There were opinions against using that but i don't think there ever was a real decision.

kocio-pl:
There was no definite decisions, so we have to choose anyway. Do we want to merge this PR or reject it? Do you plan to provide the code with runway length? If yes, I would like to test it how it works. If no, I prefer to use this code.

imagico:
Note i did not suggest to reject this, i just said i am not sure.

No, i don't plan to try an on-the-fly runway length based implementation - if i look into this matter in the future i will probably concentrate on a preprocessing solution which i find more interesting cartographically.

kocio-pl:
Sure - I just try to make well informed decision. With lack of "commercial" tagging and with runway length not being the ultimate solution for you (and with no code, which would be non-trivial), I'm now inclined to use this code when the merging window opens. However if there will be some new arguments and ideas regarding airports rendering, I will take them into account.

dieterdreist:
sent from a phone

> On 10. Jul 2017, at 11:45, kocio-pl <notifications@github.com> wrote:
> 
> I was thinking for a long time what features could make some airports "major" and found no such thing.


some relevant properties could be the number and size of runways, the number (and "rank") of different destinations and carriers and number of flights/passengers per year. ;-) The runways are often available in osm.

math1985:
:-1: for any overcomplex solution such as runway calculations.

kocio-pl:
Could you provide a code to test it? At this point I prefer something we could use in practice in the near future, general ideas discussion can continue at #1143.

kocio-pl:
Nothing beats Indiana, but in Toronto area Metro extract this code reduces number of airport from about 24 to just 3 (and all of them are located in the city), so this simple approach to prioritize airports proved to be surprisingly effective.

matkoniecz:
@imagico 

> Making some effort to add missing ICAO/IATA codes together with such a change would definitely be advisable - there are many remote airports where this is missing. Maybe the wikipedia/wikidata guys could use this opportunity to do something more useful than adding arbitrary language name tags

Generating list of missing ICAO/IATA codes that are listed in Wikidata should be fairly simple (no idea how many new ones will be found). Tricky part is that someone would need to manually verify whatever produced matches are OK and add them.

If someone would be interested in using such data I may generate them (I will not do it blindly, we already have many well done unused QA sites).

imagico:
Thanks for picking up on this.

I don't think additional specialized tools are really needed here.  As you said this should be manually verified anyway (both for data accuracy and to avoid possible license troubles if data in wikipedia/wikidata comes from questionable sources).

I don't know what tools exactly those who add name translations in OSM based on wikidata use but i would assume these are also suited for adding ICAO/IATA codes.

matkoniecz:
> I don't know what tools exactly those who add name translations in OSM based on wikidata use

I am also not sure, and googling revealed nothing relevant (I know about some people who make this kind of edits manually).

I already have basically equivalent. Iinitial reason for writing code was completely different, but in its current state it requires the same amount of additional configuration as mentioned name adders (what I will do if someone will use produced data).

daganzdaanda:
Thanks for this, it does make the map nicer!
Probably unrelated to this, I noticed that Munich Airport is missing its label: https://www.openstreetmap.org/relation/3813703
The nearby military airport is also mapped as a relation and does show up with icon and label: https://www.openstreetmap.org/relation/6725416
I suspect there could be some interference with the train station here: https://www.openstreetmap.org/node/2499357757 that displaces the airport label? Is that possible? Would it be a bug worth fixing?

kocio-pl:
It looks like highly possible, but it's easy to test it - just try rendering this area with removed station rules.

Issue #2612, opened 2017-04-17
Fix RELEASES.md markdown


pnorman:
Right now the github page shows

```
Notify OSMF tile server admins [with an issue on OpenStreetMap Chef](https://github.com/openstreetmap/chef/issues/new?title=New openstreetmap-carto release, $NEW_RELEASE&body=A new version of openstreetmap-carto, [$NEW_RELEASE]%28https://github.com/gravitystorm/openstreetmap-carto/releases/tag/$NEW_RELEASE%29, has been released.). Add any deployment-related changes like new shapefiles or font changes to the ticket.
```

This used to work - perhaps the `[` `]` need to be % encoded to avoid screwing up the markdown, or all of the spaces.

PR #2675, opened 2017-07-06
Fixing RELEASES.md markdown

kocio-pl:
Resolves https://github.com/gravitystorm/openstreetmap-carto/issues/2612.

The problem was unencoded spaces in URL, which is used to pre-fill the form.


nebulon42:
Works.

Issue #2576, opened 2017-02-28
INSTALL.md doesn't actually tell the user how to install


SomeoneElseOSM:
It reads fine as far as "Scripted download" but then gets a bit waffly:

"Manual download" doesn't explain when or why you'd do that instead of "scripted".
"Fonts" would benefit from "which font is used where" - do I most of the Noto variants listed in the code make geographical sense but Emoji?
What is the "Non-latin scripts" section supposed to mean?

Re the dependencies, it's debatable whether "Kosmtik" is one - it's needed if you want to develop the style, not just use it.

"carto" _is_ required but there are no clues here about how to install an appropriate version.  The latest Ubuntu 16.04.2 LTS has "carto 0.9.5" (not high enough) so the otherwise reliable https://ircama.github.io/osm-carto-tutorials/tile-server-ubuntu/ doesn't help.  https://github.com/mapbox/carto#installation suggests just "npm install -g carto" which doesn't work there either.

Further on, the "osm2pgsql to import your data into a PostGIS database" section is surely misplaced as INSTALL.md gave instructions for using osm2pgsql a few paragraphs above.  "Not all users will need these dependencies" should perhaps read something like "You will only need these if you want to adjust the colours used and want them to be evenly spaced as the current colours are".

(edit: typing error)

nebulon42:
>  https://github.com/mapbox/carto#installation suggests just "npm install -g carto" which doesn't work there either.

Why does this not work? I suggest opening an issue for that at the Carto issue tracker and I'll look into it.

SomeoneElseOSM:
Re "Non-latin scripts", it'd also be good to know how to check whether "FreeType 2.6.2 or later" is installed (on at least the most common OSes).  Also what is "A recent enough version" of Noto?

The Emoji question above got an answer on IRC - apparently its in a shop name in the US somewhere :)

pnorman:
The issue with fonts and international scripts is there's a big difference between what's required and what gets the best results. To work what's required is either Noto Sans (or DejaVu) and Unifont, with any version of FreeType that Mapnik compiles against. To get the best results the latest versions of all of the Noto fonts from their git repo is required as well as a recent FreeType. Other font subsystem components might need upgrades too.

If you're rendering Latin text you won't see the difference. Once you start rendering other languages, particularly SE Asian ones, the versions matter.

-----

Manual download should be a subsection of scripted download, and more clearly explain that it's an alternative to the script.

[...]

PR #2676, opened 2017-07-06
Make Docker information more notable in INSTALL.md

kocio-pl:
Related to https://github.com/gravitystorm/openstreetmap-carto/issues/2576.

I think user should have the easiest installation option visible right from the start. This should make all the problems with manual installation less severe.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -1,5 +1,7 @@
 # Installation
 
+This document describes how to manually configure your system for running openstreetmap-carto. If you prefer quick, platform independent setup for a development environment, without the need to install and configure tools by hand, follow a Docker installation guide in [DOCKER.md](https://github.com/gravitystorm/openstreetmap-carto/blob/master/DOCKER.md).
+
 ## OpenStreetMap data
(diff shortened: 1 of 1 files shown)


nebulon42:
We should not forget that INSTALL.md is also meant for users that want to run production systems. For those Docker is often not an option. It's ok to have it on top but we should rephrase the second sentence so that it also takes production systems, not only development systems, into account.

kocio-pl:
You're right, that was my intention. I'm happy to change the initial paragraphs as you like it - maybe just:

> If you can't use Docker or you prefer to configure the system manually, proceed with reading this document. **This is also preferred option for production systems.**

imagico:
I would not like to see the installation documentation to start with a recommendation to use Docker.  Docker and the underlying concept of hiding complexity and diversity in IT systems through virtualization are seen critically by many people for good reasons and projects that start their installation instructions with the suggestion to use Docker always create a bad impression for me.

I would simply write:

> This document describes how to manually configure your system for running openstreetmap-carto.  If you prefer a quick and platform independent setup for a development environment without the need to install and configure tools by hand you can find information for a Docker based environment in DOCKER.md.

PR #2679, opened 2017-07-08
Adding Kosmtik plugins to Docker

kocio-pl:
This code adds some plugins to make Kosmtik more usable.

I don't see why `.config` directory was created, because `.kosmtik-config.yml` seems to be used outside it, so I removed that line.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -12,8 +12,21 @@ RUN curl -sL https://deb.nodesource.com/setup_6.x | bash \
     && apt-get install --no-install-recommends -y nodejs \
     && rm -rf /var/lib/apt/lists/*
 
+# Kosmtik with plugins
 RUN npm install -g kosmtik
 
+WORKDIR /usr/lib/node_modules/kosmtik/ 
+RUN kosmtik plugins --install kosmtik-overpass-layer \
+                    --install kosmtik-fetch-remote \
+                    --install kosmtik-overlay \
+                    --install kosmtik-open-in-josm \
+                    --install kosmtik-map-compare \ 
+                    --install kosmtik-osm-data-overlay \
+                    --install kosmtik-mapnik-reference \
+                    --install kosmtik-geojson-overlay \
+    && cp /root/.config/kosmtik.yml /tmp/.kosmtik-config.yml
+
+# Closing section
 RUN mkdir -p /openstreetmap-carto
 WORKDIR /openstreetmap-carto
 
--- a/scripts/docker-startup.sh
+++ b/scripts/docker-startup.sh
@@ -32,11 +32,8 @@ import)
 kosmtik)
   python scripts/get-shapefiles.py -n
 
-  mkdir -p .config
   if [ ! -e ".kosmtik-config.yml" ]; then
-    touch .kosmtik-config.yml
-    # this can be removed once https://github.com/kosmtik/kosmtik/issues/236 is resolved
-    echo "plugins:" >> .kosmtik-config.yml
+    cp /tmp/.kosmtik-config.yml .kosmtik-config.yml  
   fi
   export KOSMTIK_CONFIGPATH=".kosmtik-config.yml"
 


Dockerfile
```diff
@@ -12,10 +12,32 @@ RUN curl -sL https://deb.nodesource.com/setup_6.x | bash \
     && apt-get install --no-install-recommends -y nodejs \
     && rm -rf /var/lib/apt/lists/*
 
+# Kosmtik with plugins
 RUN npm install -g kosmtik
 
+WORKDIR /usr/lib/node_modules/kosmtik/
```
nebulon42: I think you should not use WORKDIR here. Why do you need it?

Dockerfile
```diff
@@ -12,10 +12,32 @@ RUN curl -sL https://deb.nodesource.com/setup_6.x | bash \
...
+RUN npm i kosmtik-overpass-layer
```
nebulon42: Every RUN creates another layer (see https://stackoverflow.com/questions/31222377/what-are-docker-image-layers). Common practice is to group commands together. So why not writing `RUN npm i kosmtik-fetch-remote kosmtik-overlay ...`.

Dockerfile
```diff
@@ -12,10 +12,32 @@ RUN curl -sL https://deb.nodesource.com/setup_6.x | bash \
...
+RUN npm i kosmtik-fetch-remote
+RUN npm i kosmtik-overlay
+RUN npm i kosmtik-open-in-josm 
+RUN npm i kosmtik-map-compare 
+RUN npm i kosmtik-osm-data-overlay 
+RUN npm i kosmtik-mapnik-reference
+RUN npm i kosmtik-geojson-overlay
+
+RUN echo "plugins:" >> /tmp/.kosmtik-config.yml
```
nebulon42: Here applies the same about layers. You could either sacrifice readability a bit and move everything you have to echo in one string with `\n` or you could also group commands together with && thus writing:
RUN echo "plugins:" >> /tmp/.kosmtik-config.yml && echo " ..." ...

scripts/docker-startup.sh
```diff
@@ -32,7 +32,7 @@ import)
 kosmtik)
   python scripts/get-shapefiles.py -n
 
-  mkdir -p .config
```
nebulon42: Maybe a leftover, but I'm not sure now.

Dockerfile
kocio-pl: Something seems to be wrong with plugins npm installation. When Kosmtik is installed globally and plugins are installed with "i" - as advised on their npm pages - plugins are installed in the local (working) directory and are not visible to the program. I was trying to find a way to install them in `/usr/lib/node_modules/kosmtik/node_modules/` through npm options (like `--prefix`), but failed.

I guess npm installation of plugins needs to be examined by the Kosmtik team to work out of the box.

Dockerfile
kocio-pl: Sure, you're right! I was testing plugins, so it was just easier to turn them on and off.

Dockerfile
nebulon42: I have to admit I never thought about that. Maybe try to install the plugins globally too? With `npm install -g plugin`? I'll have to try it. Meanwhile, could you please open a issue about that at Kosmtik too?

nebulon42: At the Kosmtik README it says that plugins should be installed with `kosmtik plugin --install pluginname`.

scripts/docker-startup.sh
nebulon42: Definitely a leftover.

Dockerfile
nebulon42: The only solution I see is to add .kosmtik-config.yml to version control with the right plugin entries in it.

kocio-pl: This is also not the perfect solution too. When the kosmtik-place-search plugin will be working, I would like to add it to repo and then what? If the user changed the file, we should parse it and add this plugin name in the proper place (maybe other than just the end of the file). It also means that every change a user makes to the config file will probably alarm git and make updating code harder. It's not a good idea to treat configuration like the code.

That's why I've tried a brute force approach - at least configuration file will be in sync with installed plugins. So I guess I will stick to that way, but I will just use the test - if the config file is already here, user is on his own, since we don't know what it looks like. OK, we can even test if this is original file we left here before (with MD5 hash), so we can update it automatically, but this might be overengineering.

kocio-pl: I've just checked that without WORKDIR `kosmtik plugins --install pluginname` uses filesystem root (`/`) and it's not working too, of course. Maybe Kosmtik should have some different plugins discovery/configuration method, because currently global npm installation doesn't work with plugins - there are two different ways to do it, for example:

`node index.js plugins --install kosmtik-place-search`

and 

`npm i kosmtik-place-search`

(on https://www.npmjs.com/package/kosmtik-place-search) and both are broken.

kocio-pl:
Code updated, it's shorter now, since this plugin installation method creates configuration file itself.

However (re: https://github.com/gravitystorm/openstreetmap-carto/pull/2679#discussion_r126291032) we already have a configuration/code mix problem with .env file. When I've changed it, git complains about that change. It should be ignored by git probably.

Dockerfile
nebulon42: For me `kosmtik plugins --install pluginname` worked. The only problem was with the config file.

nebulon42:
Maybe another solution for the problems with both `.env` and `.kosmtik-config.yml`. For `.env`:
* on startup of import we check if the file exists locally
* if yes, we just use it
* if not we create it with default values
* it is added to `.gitignore`

For `.kosmtik-config.yml`:
* on install of kosmtik we create it with default values (plugins)
* on startup of kosmtik we check if the file exists locally
* if yes, we just use it
* if not we sync it out to the host
* it is still part of `.gitignore`

With this approach we should get reasonable default values the first time and let the user customize it. Only downside with `.kosmtik-config.yml` would be that if there are changes software-wise the config file would have to be updated by the user or deleted first before starting kosmtik container.

Dockerfile
```diff
@@ -12,10 +12,24 @@ RUN curl -sL https://deb.nodesource.com/setup_6.x | bash \
...
+RUN kosmtik plugins --install kosmtik-overpass-layer \
+                    --install kosmtik-fetch-remote \
+                    --install kosmtik-overlay \
+                    --install kosmtik-open-in-josm \
+                    --install kosmtik-map-compare \ 
+                    --install kosmtik-osm-data-overlay \
+                    --install kosmtik-mapnik-reference \
+                    --install kosmtik-geojson-overlay
+
+RUN cp /root/.config/kosmtik.yml /tmp/.kosmtik-config.yml
```
nebulon42: If we could tie this cp with && to the kosmtik plugin install I'm happy. :)

nebulon42:
I have verified that it indeed does not work without setting the workdir. We have to tackle that from the Kosmtik side. And I saw that you already do for the Kosmtik config what I suggested. So there would only be one minor comment left, otherwise I'm fine with it.

PR #2680, opened 2017-07-08
Adding .gitattributes to Docker configuration

Ircama:
Adding .gitattributes to avoid that Git for Windows converts docker-startup.sh newlines to DOS, so that this script can run inside a UNIX container.

This configuration affects all *.sh files of the repository.

Tested with Ubuntu and with Windows.


nebulon42:
According to the Git documentation this should be correct.

PR #2681, opened 2017-07-09
Fixing Docker documentation

kocio-pl:
Fixing small problems with DOCKER.md. The most important one was 

> `docker-compose up` to run the style preview application

instead of `docker-compose up kosmtik`, because it was importing (probably the same) data file each time, which may be very time consuming.


nebulon42:
Right, this is important. Thanks.

PR #2682, opened 2017-07-09
Switch forest pattern to SVG

sommerluk:
Switch forest pattern to SVG.

I’ve taken the existing forest.svg and sanitized it. Apparently it is not directly corresponding to the currently used PNG file: It’s the same pattern, the if at a given point the broadleaved symbol is the left one or the right one is sometimes different to the current PNG.


kocio-pl:

I will test it soon, but I'm happy that you do it - thanks!

kocio-pl:
z14
Before
yepjdgky
After
ma5tpdvk

imagico:
Thanks.  Not sure about the icon randomization difference.  I did some tests with that some time ago, trying to relax the icon selection in a similar way as the locations (to avoid clusters of identical icons) - it is possible that the PNG was generates with that.  Not really that significant with this pattern though.

By the way now that we can use leaf_type/leaf_cycle we could also revisit the pattern itself - see https://github.com/gravitystorm/openstreetmap-carto/pull/1242#issuecomment-126958042.

PR #2684, opened 2017-07-10
Switch scrub pattern to SVG

kocio-pl:

Warsaw, z18
Before
srnso4r7
After
5ip vekk


imagico:
What's with the offset in the samples?

kocio-pl:
I don't know. It was just simple script operation, so it might be this (Inkscape including) or it's a Mapnik issue.

kocio-pl:
I hope symbol shift is not an obstacle for you to merge it?

imagico:
I have not tested if a similar problem as with plant_nursery occurs here as well.  In general i would assume that whatever mapnik does exactly, if it breaks periodicity in one case it likely also breaks it in other cases.  But since the potential damage is not that big i don't think this is a big deal if it just happens for one pattern.  I would however be against moving to SVG patterns at large without the ability to control how mapnik renders them at least on the same level as with PNGs.

kocio-pl:
I understand it as "OK, but let's think how to resolve general problem before we start with others". Sure, https://github.com/gravitystorm/openstreetmap-carto/pull/2685 is waiting for the solution to be merged.

PR #2685, opened 2017-07-10
Switch plant_nursery pattern to SVG [WIP]

kocio-pl:

Warsaw, z17
Before
zpie0d0t
After
457lgxqg


imagico:
Here the sample clearly shows an error.

The SVG contains a `stroke-width:0.125` which does not belong there - this however is likely not the cause of the error.

My guess is that Mapnik kind of auto-offsets/auto-centers SVGs (which is insane for pattern images) - since the forest pattern symbols cross the edges on all sides there is no offset but with the other patterns there are.

kocio-pl:
I hope we can fix it somehow, but until then I would drop this PR.

imagico:
From my point of view there are two ways to deal with that other than fixing Mapnik:

* adjust the symbol offsets of the pattern so that the symbols cross all edges like in case of the forest pattern - this is always possible.
* add an invisible geometry to the SVG that prevents Mapnik from messing with it.

kocio-pl:
I don't know how to do it, but if you do, that would be great.

imagico:
I re-opened https://github.com/imagico/jsdotpattern/issues/3 because of this.

pnorman:
> add an invisible geometry to the SVG that prevents Mapnik from messing with it.

That's what we've been doing in the past.

kocio-pl:
As this code shows some problems and it looks like there's nothing new for a month to fix it, I'm closing this PR now.

kocio-pl:
It looks like new script fixes the problem here too. It's just 128 px (we don't have to worry about 512 px limit), so I can create new PR soon.

Issue #2677, opened 2017-07-06
Extending rendering of trees depending on the additional values


wmyrda:
I see heated discussion over import mailing list about adding of 150 000 trees for city of Ottawa. IMHO it is a mater of time before this data and similar data as that will be provided or added into OSM therefore it would not be a bad idea to make use of at least some of the additional tags.

Code I attempted to prepare http://bigvo.hopto.org/osm-extra/carto-patches/4.0/021-natural_monument-tree_v5_direct.patch does make use of **height**, **circumference** and **denotation** to further distinguish among trees on the map.

General idea there is for trees greater in size especially to appear sooner and to distinguish natural monuments as those worth extra attention.

I am not the postgis guru therefore code I have prepared fails to work correctly for me due to renderd not being able to find "denotation" among values in the database even when I attempted to get them from the hstore table so help finishing up the code here is welcomed.

```renderd[13870]: ERROR: failed to render TILE default 15 18384-18391 11104-11111
renderd[13870]:    reason: Postgis Plugin: ERROR:  column "denotation" does not exist
```

imagico:
It would be much more useful (and also easier for everyone to answer your questions) if you'd check in your changes as a separate branch on a fork of this project on github under your account.  If you want your modification to be merged here this is actually a requirement - see https://github.com/gravitystorm/openstreetmap-carto/blob/master/CONTRIBUTING.md

Regarding the change itself - IMO you would need to show some examples why your changes are useful.  The current uniform rendering of all trees in the same form as a universally recognizable symbol has great advantages for the map user - changing that will probably make reading the map more difficult.  You would need to demonstrate that the advantages outweigh these disadvantages.

Side note: 150k new trees in Ottawa are not really significant in the grand scheme of things - overall we currently have 8 million trees in the database.  We also try to base rendering decisions mostly on manual mapping to give mappers useful feedback for their work.

wmyrda:
I the future I plan to make more use of github and learn it but for now that is above my knowledge level hence the code change proposal is available in the form of a patch which is what I am familiar with.

As for uniformity of the code I do not see any problem there as the code does not change the way the trees as shown currently in the carto and practically changes nothing for the trees will the only tag of natural=tree. Ideas behind the code reflect only the extension of the current method which is affecting very small amount of trees overall 
- larger trunk for trees with greater circumference. That tag only have 2.88% of all trees in OSM DB
[...]

PR #2686, opened 2017-07-10
Extending rendering of trees

wmyrda:
Just as stated in comments for https://github.com/gravitystorm/openstreetmap-carto/issues/2677 I created what I hope is a proper pull request. Like stated in the issue the code definitely needs a look at project.mml as I am not sure why, but I get those postgis plugin errors.

As of the code change itself the given numbers to the tree are what I felt are most appropriate, but hence the code did not work for me I was not able to check different values the that in the said example.

### DIFF
--- a/amenity-points.mss
+++ b/amenity-points.mss
@@ -2064,11 +2064,12 @@
     text-wrap-width: @standard-wrap-width;
   }
 
-  [feature = 'natural_tree'][zoom >= 17] {
+    [feature = 'natural_tree'] {
     text-name: "[name]";
     text-size: @standard-text-size;
     text-fill: green;
-    text-dy: 7;
+    [zoom >= 16][denotation = 'natural_monument'] { text-dy: 6; }
+    [zoom >= 17] { text-dy: 7; }
     [zoom >= 18] { text-dy: 8; }
     [zoom >= 19] { text-dy: 11; }
     [zoom >= 20] { text-dy: 18; }
@@ -2080,7 +2081,7 @@
   }
 }
 
-#trees [zoom >= 16] {
+#trees_poly [zoom >= 16] {
   ::canopy {
     opacity: 0.3;
     [natural = 'tree_row'] {
@@ -2100,6 +2101,12 @@
         line-width: 30;
       }
     }
+  }
+}
+
+#trees [zoom >= 16] {
+  ::canopy {
+    opacity: 0.3;
     [natural = 'tree'] {
       marker-fill: green;
       marker-allow-overlap: true;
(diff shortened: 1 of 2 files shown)


math1985:
To be honest, I think this is getting too specific for a general purpose style-sheet. But I'd be happy to hear what others think.

math1985:
Thanks for creating a pull request in any case, makes it much easier to discuss!

amenity-points.mss
```diff
@@ -2064,11 +2064,12 @@
     text-wrap-width: @standard-wrap-width;
   }
 
-  [feature = 'natural_tree'][zoom >= 17] {
+    [feature = 'natural_tree'] {
     text-name: "[name]";
     text-size: @standard-text-size;
     text-fill: green;
-    text-dy: 7;
+    [zoom >= 16][denotation = 'natural_monument'] { text-dy: 6; }
```
imagico: This is what causes your error - you use the column `denotation` here which means all layers this rule is used on need to provide it - which they don't, since your project.mml changes only modify the tree layers and not the text layers.

imagico:
Thanks for making this a PR - this makes it much easier to evaluate your changes.

Since the code at the moment does not run i cannot test it but from looking over it a few remarks:

* `circumference` is in meters, you will probably not find a tree with more than 300m circumference.
* `denotation=natural_monument` is used about 22k times, more of half of this are from an import in Poland.  The tag is fairly vague in definition and quite clearly used incorrectly in a lot of cases, for example https://www.openstreetmap.org/node/4706682524 or https://www.openstreetmap.org/node/4706670408 and around seem quite clearly cases of `denotation=avenue`.  So i don't think this tag is suited for making rendering decisions here

For assessing the styling ideas i would need to see the results.

amenity-points.mss
wmyrda: I was thinking it might be in there as looking at other stuff I see some stuff get changed into features, but i am not to sure how to go by doing it just yet so help is appreciated.

wmyrda:
@imagico I got through it on github GUI which looked not that hard after all, but looking at the code there sucked comparing to looking at it in nano for which I created syntax highlighting http://bigvo.hopto.org/osm-extra/nanosyntax/ which is easier to read that way.

I was sure circumference was in centimeters which seems more appropriate as just like You say trees do not grow above several meters. I am glad You pointed that out to me.

I do not think denotation=avenue is appropriate if You know each tree location. Just like with tree_row it is what one uses not having all individual data from the site. Same way one could also say no building is necessary as landuse=residential gives an idea where buildings would be. Besides having all those trees individually allows addition of specific tags as height and circumference in the future.

Those imports in Poland come from polish registry of nature protection and all of those trees are legally protected as nature monuments hence the tag. 

When that error in code gets fixed I am sure small change here and there would have to be made.

imagico:
> @imagico I got through it on github GUI which looked not that hard after all, but looking at the code there sucked comparing to looking at it in nano for which I created syntax highlighting http://bigvo.hopto.org/osm-extra/nanosyntax/ which is easier to read that way.

But this is not about the github UI - this is about having a standardized workflow for testing and evaluating changes.  With a git branch available i can simply check out your changes and test them without the difficulty of manually applying patches.

> Those imports in Poland come from polish registry of nature protection and all of those trees are legally protected as nature monuments hence the tag.

Which kind of proves my point that denotation=natural_monument is not really suitable for rendering decisions here.  If you look at https://wiki.openstreetmap.org/wiki/Key:denotation you can see that depending on how you interpret it landmark, natural_monument, avenue and urban can all apply at the same time.  If we'd use denotation=natural_monument that would just push this particular value.  It is just a way to attempt a classification where no clear boundaries for a classification exist.

Regarding the error - if you want to use denotation for labeling decisions (and as i indicated we probably do not want to do that) you would have to return a denotation column in all queries with `class: text` from SQL).

wmyrda:
> But this is not about the github UI - this is about having a standardized workflow for testing and evaluating changes. With a git branch available i can simply check out your changes and test them without the difficulty of manually applying patches.

OK. I did not want to change anything about the procedures You have especially that programmers more used to git can fetch repositories and open files in whatever editor they prefer. It is just I hence I believed I had code quite in decent shape I wanted to share it the way I knew.

> If you look at https://wiki.openstreetmap.org/wiki/Key:denotation you can see that depending on how you interpret it landmark, natural_monument, avenue and urban can all apply at the same time.

Idea there was that if names of the trees are used than it might be worth it to show names of more important trees earlier such as those tagged with denotation.
I will think of proper way of adding avenue to the equation as it likely would harder hence possibility of adding it as simple line or relation of many trees (points). If you ask me landmark should be of even greater scarcity or higher in the hierarchy of denotations therefore if name was to show for it than I personally would bright it up even one level higher.

Having said that I will prepare some screen examples of carto look without using denotation to show names as it will be much simpler hence that
> you would have to return a denotation column in all queries with class: text from SQL.

sounds voodooish to me :(

imagico:
I think you should simply not use `denotation`.  If you want to show some trees earlier or more prominently than others best is to use some clearly verifiable property.  But as indicated before i am not really convinced varying the display of trees is a good idea.  The nature of a tree often becomes clear to the map user from context - avenue trees for example really do not require any special rendering.  You would need to show practical examples where a differentiated rendering substantially improves the map.

>> you would have to return a denotation column in all queries with class: text from SQL.
> sounds voodooish to me :(

The problem is you picked a somewhat complicated case for your first change.  The tree symbols and tree labels are rendered separately.  The tree symbols are in the `trees` layer and are styled based on layer id here:

https://github.com/gravitystorm/openstreetmap-carto/blob/fcb8ac3d349b19dd78306bd1825da975664823fd/amenity-points.mss#L2083

The labels are in the `text-point` layer which is styled based on class here:

https://github.com/gravitystorm/openstreetmap-carto/blob/fcb8ac3d349b19dd78306bd1825da975664823fd/amenity-points.mss#L1054

The layer's SQL query needs to return all columns which are used in the mss - so if you add use of the denotation column to the text styling rules you need to add a denotation column to all SQL queries of layers of the text class (which is `text-point`, `text-line` and `text-poly`).  The value of this column of course can also be NULL for all cases except the one you use it for.  We already have similar columns in these layers like the `operator` column.

kocio-pl:
I haven't look what the rendering looks like, but with this fix it works at all:

https://github.com/kocio-pl/openstreetmap-carto/commit/53b2ce1d7d346d8cd1c642dd874f2c361a06ee1b

pnorman:
Can you include [a few images selected to show the changes](https://github.com/gravitystorm/openstreetmap-carto/blob/master/CONTRIBUTING.md#pull-requests).

wmyrda:
@pnorman The code had several bugs making it impossible to render properly mainly due to improper logical statements in the amenity-points.mms. I just got it working on my test site http://bigvo.hopto.org/osm/?zoom=18&lat=50.5406&lon=23.23011&layers=B0 or http://bigvo.hopto.org/osm/?zoom=17&lat=50.0396&lon=21.99882&layers=B0 which looks just like what I was after. I will switch over to carto git and this patch alone and will provide some screenshots with rendering differences.

wmyrda:
I have spend last two weeks trying to get this to work, but I must be doing something wrong as my local map does not seem to use circumference & height values at all for some reason therefore I am not able to provide screenshots with different tree looks. Only thing that works is slightly different tree color for natural monuments.
Anyways my best effort is here https://github.com/wmyrda/openstreetmap-carto/tree/wmyrda-trees2 for know as I did not want to create additional pull request. I hope for You guys to finish it up as I am out of ideas.

What is suppose to do is
- keep the currently used values and display form in carto to render trees
- use darker 005500 color for natural monument trees instead of default 008000
- extend current render of trunk and canopy display size increase with greater circumference values by 10% with every 1m extra size
- display at z14 & z15 trees that are national monuments with specific values. By having early access to full database of monuments in Poland I can say used values accumulate to about 1% of those trees and as such them being regional tourist attractions deserve earlier display.

On the side not having 6 different rendering sizes for each zoom level might seem as overkill, but it does add to the render realism.

kocio-pl:
Maybe you try too much tree-related things at once. For example monuments color seems to be separate issue from canopy-scaling. I suggest closing this PR and creating smaller patches dedicated to solving these issues.

kocio-pl:
Since author is not interested in further work on this PR, I close it now. If somebody would be interested, the code is available to create new PRs.

wmyrda:
"author is not interested in further work on this PR" This is not entirely accurate statement. Author simply does not know why some rules are not used from the code already that is present and if any help fixing it up arises author will be happy to continue work on it. 
For instance only someone that knows the coding language may say if for example
- are circumference statements are nested too deeply
- is it possible to set variable for all changes to circumference as "circumference>1 * 1,1 & circumference>2 *1,2 etc" to be called simply as @circumference at specific zoom levels

On top of that author also believe that even this code is not top priority then adding realism to tree size would help in places like this http://www.openstreetmap.org/#map=19/50.02366/21.99961 as it looks more like "area hit by smallpox virus"

kocio-pl:
Sorry for the wording. Anyway I think this PR was too big, which makes solving problems a hard thing.

Issue #1977, opened 2015-11-15
Consider level=*


PanierAvide:
Hello, I think the rendering should consider objects with `level=*` tags, as consider I mean ignore or use a special rendering (depending of what is most reasonable). The reason of this issue is that objects with level tags most of the time are under or above ground (level different of 0), so making it appear on map could make people thinking they are on ground when they are not.
Example ([OSM.org](http://www.openstreetmap.org/#map=19/48.12109/-1.70414&layers=N)):
![Villejean Université Subway](https://framapic.org/ewRWwdr56H6m/RO7c1eXw)

Each trash bin, bench and information board is under ground (`level=-1`), but with this rendering it seems like it is on ground.

jojo4u:
Level -1 does not mean underground. It means first basement floor. In a building at a steep hill this can be partly above ground. The key level groups features on one floor together, it's imho not used for "what is above each other" aka "vertical relationships". For example this platform which extends to underground to the right is level=-1: https://goo.gl/maps/XUW99gv6hF72 

For the "vertical relationships" the key layer is used. But this also does not mean that something is invisible from the ground, see example picture from Wiki: http://wiki.openstreetmap.org/wiki/File:Washington_layers.png

The tag location=indoor/underground or the tag indoor=yes comes closest what you want to accomplish.

PanierAvide:
Thanks for your reply. I agree with the point that first basement floor is not always underground, but as level=0 is ground, most of the time it will be underground. The idea of using `level` key is to make the change simple. But in fact it would be more reliable to check if the object is inside the area of a building (and checking its `building:levels` value).

You propose `location` or `indoor=yes`, is it already supported ? If not, I'm not sure using this kind of tag is really a good idea. It's like saying an indoor oriented map should ignore outside objects only if they have an `outside=yes` tag. I'm sure we can find a more subtle way to do this ;)
Just to make it clear, I'm not saying that we should ignore `indoor` tag. This is the right way to distinguish indoor rooms or areas. I'm saying that for other smaller objects like trash bins or benches, it would be better to use something smarter than adding `indoor=yes` to ignore them as they already use other indoor-related tags (such as `level`).

math1985:
@PanierAvide How would you expect this situation to be rendered?

PanierAvide:
Something like this, where small objects with `level=-1` (here in the first basement floor of a subway station) are ignored :
[...]

PR #2687, opened 2017-07-11
Render railway labels

### LINKED ISSUE #505: No labels for railway=(rail, subway, light_rail, ...; platform), public_transport=platform
Could labels be added back for railways and platforms? They seem to have been removed sometime in the last few months.

[edit] I was thinking about <code>railway=platform</code> and <code>public_transport=platform</code> with <code>railway</code>/<code>subway</code>/etc.<code>=yes</code>, not <code>highway=platform</code> and <code>public_transport=platform</code> with <code>bus</code>/etc.<code>=yes</code>. Bus and share taxi platforms should probably be treated separately and usually don't need labels.

On the other hand, would it be good if ref=\* was used for the label of the platform if there is no name=*?

littlebtc:
This will fix #505.

Rendering railway labels will make maps in some railway-centric area better, especially Japan.

I did some copy-and-paste work and make this pull request, hoping to get some feedback about how the label should be styled.

My two cents:

- Should the style of railway lines be changed to match the label?
- Subway under highways is mostly not rendered. Can we render that?

By following is some rendered images of the initial commit in Tokyo, Japan:

### DIFF
--- a/roads.mss
+++ b/roads.mss
@@ -3106,3 +3106,113 @@ tertiary is rendered from z10 and is not included in osm_planet_roads. */
     }
   }
 }
+#railways-text-name {
+  /* Mostly started from z17. */
+  [railway = 'rail'],
+  [railway = 'subway'],
+  [railway = 'narrow_gauge'],
+  [railway = 'light_rail'],
+  [railway = 'preserved'],
+  [railway = 'funicular'],
+  [railway = 'monorail'],
+  [railway = 'tram'] {
+    [zoom >= 17] {
+      text-name: "[name]";
+      text-fill: #666666;
+      text-size: 10;
+      text-dy: 6;
+      text-spacing: 900;
+      text-clip: false;
+      text-placement: line;
+      text-min-distance: 18;
+      text-face-name: @book-fonts;
+      text-halo-radius: @standard-halo-radius;
+      text-halo-fill: @standard-halo-fill;
+    }
+    [zoom >= 19] {
+      text-size: 11;
+      text-dy: 7;
+    }
+  }
+  [railway = 'rail'] {
+    /* Render highspeed rails from z11,
+       other main routes at z14. */
+    [highspeed = 'yes'] {
+      [zoom >= 11] {
+        text-name: "[name]";
+        text-fill: #666666;
+        text-size: 10;
+        text-dy: 3;
+        text-spacing: 300;
+        text-clip: false;
+        text-placement: line;
+        text-min-distance: 18;
+        text-face-name: @book-fonts;
+        text-halo-radius: @standard-halo-radius;
+        text-halo-fill: @standard-halo-fill;
+      }
+      [zoom >= 13] {
+        text-dy: 6;
+      }
+      [zoom >= 14] {
+        text-spacing: 600;
+      }
+      [zoom >= 17] {
+        text-size: 11;
+        text-dy: 7;
+      }
+      [zoom >= 19] {
+        text-size: 12;
+        text-dy: 8;
+      }
+    }
+    [highspeed != 'yes'][usage = 'main'] {
+      [zoom >= 14] {
+        text-name: "[name]";
+        text-fill: #666666;
+        text-size: 10;
+        text-dy: 6;
+        text-spacing: 300;
+        text-clip: false;
+        text-placement: line;
+        text-min-distance: 18;
+        text-face-name: @book-fonts;
+        text-halo-radius: @standard-halo-radius;
+        text-halo-fill: @standard-halo-fill;
+      }
+      [zoom >= 17] {
+        text-spacing: 600;
+        text-size: 11;
+        text-dy: 7;
+      }
+      [zoom >= 19] {
+        text-size: 12;
+        text-dy: 8;
+      }
+    }
+  }
+  /* Other minor railway styles. For service rails, see:
+     https://github.com/gravitystorm/openstreetmap-carto/pull/2687 */
+  [railway = 'preserved'],
+  [railway = 'miniature'],
+  [railway = 'disused'],
+  [railway = 'construction'] {
+    [zoom >= 17] {
+      text-name: "[name]";
+      text-fill: #666666;
+      text-size: 10;
+      text-dy: 6;
+      text-spacing: 900;
+      text-clip: false;
+      text-placement: line;
+      text-min-distance: 18;
+      text-face-name: @book-fonts;
+      text-halo-radius: @standard-halo-radius;
+      text-halo-fill: @standard-halo-fill;
+    }
+    [zoom >= 19] {
+      text-size: 11;
+      text-dy: 7;
+    }
+  }
+}
--- a/project.mml
+++ b/project.mml
@@ -1974,6 +1974,37 @@ Layer:
         ) AS paths_text_name
     properties:
       minzoom: 15
+  - id: railways-text-name
+    geometry: linestring
+    <<: *extents
+    Datasource:
+      <<: *osm2pgsql
+      table: |-
+        (SELECT
+            way,
+            CASE WHEN railway = 'preserved' AND service IN ('spur', 'siding', 'yard') THEN 'INT-preserved-ssy'::text
+                 WHEN (railway = 'rail' AND service IN ('spur', 'siding', 'yard')) THEN 'INT-spur-siding-yard'
+                 WHEN (railway = 'tram' AND service IN ('spur', 'siding', 'yard')) THEN 'tram-service' ELSE railway END AS railway,
+            CASE WHEN (tunnel = 'yes' OR tunnel = 'building_passage' OR covered = 'yes') THEN 'yes' ELSE 'no' END AS tunnel,
+            tags->'highspeed' as highspeed,
+            tags->'usage' as usage,
+            construction,
+            name
+          FROM planet_osm_line l
+          WHERE railway IN ('rail', 'subway', 'narrow_gauge', 'light_rail', 'preserved', 'funicular',
+                            'monorail', 'miniature', 'tram', 'disused', 'construction')
+            AND (tunnel IS NULL OR NOT tunnel IN ('yes', 'building_passage'))
+            AND highway IS NULL -- Prevent duplicate rendering
+            AND name IS NOT NULL
+          ORDER BY
+            z_order DESC, -- put important rails first
+            COALESCE(layer, 0), -- put top layered rails first
+            length(name) DESC, -- Try to fit big labels in first
+            name DESC, -- Force a consistent ordering between differently named railways
+            l.osm_id DESC -- Force an ordering for railways of the same name, e.g. dualized rails
+        ) AS railways_text_name
+    properties:
+      minzoom: 11
   - id: text-poly-low-zoom
     class: text-low-zoom
     geometry: polygon


kocio-pl:
Could you provide before/after renderings? Tokyo is a dense city and I don't read Japanese, so it's hard for me to see where the changes are.

littlebtc:
After switching to Google Cloud , endless waiting for importing and rendering :joy:, my second approach is now ready to review.

Before/After images in London were made as requested. Before is the current rendering on OSM.org. Diff views were made with Kosmtik. I choose places near King's Cross Station to see the impacts on other labels.

The drawback of the change is obvious if you look at buildings and residential areas under the subway. 😢 

Z14
london-z14

Z15
london-z15

Z16
london-z16

Z17
london-z17

Z18
london-z18

kocio-pl:
I'm not able to review SQL code, but I've just tested it and it looks good for me:

Warsaw, z18
Before
epnazy5g
After
pszuzuiu

Was there any particular reason for you to start from z13?

littlebtc:
> Was there any particular reason for you to start from z13?
Since I copied and simplified styles from the highway name rendering, where `highway=motorway/trunk/primary` started render names at z13, and `highway=residential` started at z15.

Maybe started rendering from z11 can be a good idea, since road shields and motorway junction refs started at that.

But to make railways rendering more flexible, I want to find out a way to separated high-speed rail from normal one, since there is no simple way to determine this unless reading `maxspeed`.

littlebtc:
After visiting OpenRailwayMap, checking for current map rendering, I have pushed changes about rendering levels, label rendering in this approach is like below:

- `railway=rail` and `highspeed=yes` at z11
- `railway=rail` and `usage=main` at z11
- Most railways at z13, styling like `highway=primary`
- `railway=miniature`, service railways, disused, constructions at z15 like `highway=residential`

Before / After will be uploaded later.

kocio-pl:
Try to set export window in Kosmtik and just changing branches, so the resulting images will be directly comparable (I try opening in new tabs in my browser and fast switch between them).

littlebtc:
I will try to use my own Linux machine instead since my cloud machine has relatively small vCPU...

littlebtc:
Update: I have given up on checking `highspeed` and `usage` as it will took a hard work.

In the next images, all `railway=rail` will rendered on z11.

kocio-pl:
As I said in https://github.com/gravitystorm/openstreetmap-carto/issues/1959#issuecomment-317207988, database reload is not expected sooner than 5 months and as far as I understand changing `openstreetmap-carto.style` (or `openstreetmap-carto.lua`) would require database reload.

littlebtc:
Sorry for lots of waiting... Great London again.

**z11**
z11-before
z11-after

**z12**
z12-before
z12-after

**z13**
z13-before
z13-after

**z14**
z14-before
z14-after

**z15**
z15-before
z15-after

**z16**
z16-before
z16-after

littlebtc:
Z11 in Tokyo, feeling like a different map.
That's what I originally wanted. :joy:

kocio-pl:
For me Tokyo shows that z11 is too early, map seems to be cluttered.

littlebtc:
It looks like I should generate some Tokyo before/after.

(Tokyo railway system are too far intensive)

littlebtc:
z12 to z17 before/after for Tokyo is here.
IMHO we can start render railways at z13 while removing the text halo from `railway=rail`.
Ideas?

For z12/z13, I centered at the central of the Tokyo.
**z12**
z12-before
z12-after
**z13**
z13-before
z13-after

For z14 and higher, I focused on Shinjuku, major suburbs and railway hub of Tokyo.
**z14**
z14-before
z14-after

**z15**
z15-before
z15-after

**z16**
z16-before
z16-after

**z17**
z17-before
z17-after

roads.mss
```diff
@@ -3100,3 +3100,87 @@ tertiary is rendered from z10 and is not included in osm_planet_roads. */
     }
   }
 }
+#railways-text-name {
+  /* Render railway=rail from z11 */
+  [railway = 'rail'][zoom >= 11] {
+    text-name: "[name]";
+    text-fill: #333333;
+    text-size: 8;
```
kocio-pl: We're trying to get rid of too small text sizes, see https://github.com/gravitystorm/openstreetmap-carto/issues/2209.

roads.mss
```diff
@@ -3100,3 +3100,87 @@ tertiary is rendered from z10 and is not included in osm_planet_roads. */
...
+      text-dy: 4;
+      text-spacing: 300;
+      text-clip: false;
+      text-placement: line;
+      text-face-name: @book-fonts;
+      text-halo-radius: @standard-halo-radius;
+      text-halo-fill: #cecece;
+    }
+    [zoom >= 14] {
+      text-size: 9;
```
kocio-pl: The same text size problem as with https://github.com/gravitystorm/openstreetmap-carto/pull/2687/files#r129007651.

littlebtc:
Changes were made to reflect some comments and the review:

- SQL: Now using hstore to get `highspeed` and `usage`
- All font sizes started at 10
- New rendering levels:
  - `highspeed=yes` rail at z11
  - `usage=main` at z13
  - All others at z15 

Some `usage=main` font size still smaller in z15+, I will fix that.

littlebtc:
z11~z15 after this change:

z11
z12

z14
z15

roads.mss
```diff
@@ -3100,3 +3100,81 @@ tertiary is rendered from z10 and is not included in osm_planet_roads. */
     }
   }
 }
+#railways-text-name {
+  /* Others started from z15. */
```
kocio-pl: I think subway, tram and other city transport labels should be rendered much later (like z18-z19), since such cities tend to be cluttered with labels anyway.

littlebtc: I think z18 is a little bit late; anyway I will try to render New York to see how does it look.

littlebtc: In Manhattan NYC, z14 ~ z16 are really too cluttered, no much rooms to render subway labels :cry:.

How about:
- High-speed from z11
- Main rail lines from z14
- Others from z17?

littlebtc:
Latest changes:
- `highspeed=yes` z11->10, with halo, z17->11, z19->12
- `usage=main` z14->10,  z17->11, z19->12
- Others z17->10, z19->11

imagico:
I have had a quick look at this and it seems diligent work.  I however see a lot of duplicate labels on both sides of tracks, not sure what causes this.

It would also be of interest to get the opinions of the OpenRailwayMap people - cc @rurseekatze @Nakaner.

It might be that for normal map user with no particular interest in railways labeling the physical infrastructure is not really that useful and could lead to tagging for the renderer, i.e. mappers naming tracks based on a passenger train connection using them.

Nakaner:
@imagico wrote:
> I have had a quick look at this and it seems diligent work. I however see a lot of duplicate labels on both sides of tracks, not sure what causes this.

That might be caused by rendering the route relations which are available as multilinestrings in the lines table?

@imagico wrote:
> It would also be of interest to get the opinions of the OpenRailwayMap people - cc @rurseekatze @Nakaner.
> 
> It might be that for normal map user with no particular interest in railways labeling the physical infrastructure is not really that useful and could lead to tagging for the renderer, i.e. mappers naming tracks based on a passenger train connection using them.

Thank you, @imagico, for notifying me. I agree with you.

I am rather sceptical regarding the rendering of `name=*` of railway tracks.

Adding more labels clutters the map. If I had a choice between labeling a street and labeling a railway track, I would prefer the street. It is more useful for orientation.

The names of railway lines are usually the names of the routes (from infrastructure point of view) because a `railway=rail/light_rail/subway` track represents two rails with sleepers and ballast and not the services using this track. The names you are going to render are in use among railway staff and railway fans (sometimes also by the locals). But often the reference numbers of the services on the track are more useful. @all Plase have a look at openrailwaymap.org, zoom in to zoom level 13/14, look around and ask yourself if you want all these names (not the numbers) be rendered on the main map.

OSM Carto dropped the rendering of `name=*` by default for everything a few year ago. That was a great decision, even for railway tracks. Before that time, people tagged something as name which they wanted to be shown on *the* map. For example, mappers added the reference numbers of all routes using the track (`name=S1, S11, S2`) or the track number (`name=track 4`) in stations as names. Both examples are abuses of the `name` tag. Neither the reference number nor the track number in conjunction with the word `track` are names. Lots of these wrong names still remain in OSM from that time).

Someone who needs a city map which shows the streets, relevant points of interest and the public transport line, should use more appropriate map.

Even if we keep the rendering of the names as proposed, the font colour, the width and the colour of the label should be reconsidered. They are difficult to read if they are located on the neighbouring track (black font on grey background). The labels of subway tracks are difficult to read, too.

`service=*` should not be labeled at all. `name=*` is used for the name of the route (infrastructure point of view) and should not be added to tracks with `service=*` at all.

To cite the Catography Guide of OpenStreetMap Carto Style:
> It's an important feedback mechanism for mappers to validate their edits and helps to prevent unfavorable fragmentation of tag use.

OpenRailwayMap serves this job for railway infrastructure tagging and uses the whole colour spectrum for that purpose. ;-) Squeezing this into a multi purpose map is difficult.

> It's a major part of the public face of OpenStreetMap, for many people the map on osm.org rendered with this style is OpenStreetMap.

Is OpenStreetMap a project of some railway nerds? No, it isn't and I am happy that it is as it is.

If you want to do something good to this map style, remove the rendering of `name` on train platforms by rendering `ref`. This is much more useful and would encourage a good tagging style.

littlebtc:
> The names you are going to render are in use among railway staff and railway fans (sometimes also by the locals). But often the reference numbers of the services on the track are more useful. 

Unfortunately I think this is very country-specific.
Line refs seems to be common and useful in Europe, but not so in Asia.

I live in Taiwan; Taipei MRT just introduced their latest route refs a year ago, which is based on the color.

In China, most subway systems has numbering routes, so rendering `ref` may be better -- but not the long-range railway system.

In Hong Kong, MTR has only route names and colors but no refs.

In Tokyo, JR East, major railway company in Tokyo, just introduced their ref system in 2010s, so rendering 山手線 ("Yamanote Line") is way better than its ref "JY". Even for the Subway system which has refs for decade, rendering "Marunouchi Line" is better than just "M".

**And a travel to Tokyo without railway names is painful, as railways and subways are your main way of getting around.**

> service=* should not be labeled at all. name=* is used for the name of the route (infrastructure point of view) and should not be added to tracks with service=* at all.

Agreed. I will drop `name` rendering on  `[railway = 'INT-preserved-ssy'],  [railway = 'INT-spur-siding-yard'],  [railway = 'tram-service']` on my next commit.

littlebtc:
> It might be that for normal map user with no particular interest in railways labeling the physical infrastructure is not really that useful and could lead to tagging for the renderer, i.e. mappers naming tracks based on a passenger train connection using them.

...I also agreed with that. In fact, in Taiwan, I had seen relation collected stations and routes based on a express train...

littlebtc:
https://gist.github.com/littlebtc/24bef597053015ebc160d8c1fd2f1252
Gist for images in Tokyo and Berlin, from z11 to z19.

- On z11~13, only `highspeed=yes`, so impact is extremely low.
- `usage=main` joins at z14, making major railways available.
- Others at z17 to provide more details on urban area.

kocio-pl:
I have nothing against hispeed names - they are not so common and may be useful hint outside urban areas. usage=main might be also OK. But looking at ORM it looks like instead of names there are sometimes refs and it's not what I would like to see on a generic map.

I would rather not render all the others.

littlebtc:
**Rendering `ref` instead of `name`?** 

For some region it will result in better railway rendering, and it will encourage people to add proper railway refs in some country like Japan.

If you look at OpenRailwayMap in Tokyo in z12, you will find most of the ref-ed railways is still using names:


But it may hurt urban areas without any railway referencing system. :cry: So it could be a both tagging and rendering problems.

**Not rendering others at z17+**

A good question.

To me the main problem is the subway name:
1. As the current rendering focus on highway much more than railway, seeing subway name in the secondary/tertiary road feels bad
2. It will also hurt name rendering on communities with subway underneath.

I still hope it could be there, but If z17+ rendering is not eventually implementeed, I wish there are some more renderer to do that, like Transport Map.

-----------------------
I got my Docker optimization pull-request ready, so I will try export 100% images before I went to sleep.
(It is 03:27AM in Taichung)

kocio-pl:
I think on the general map the most important thing are the stops, stations and terminals, not the names of the lines, so I'm cautious here.

littlebtc:
IMHO I think the problem is that OSM-carto rendering is global, and railway statuses differ too much in countries.

For Google Maps, railways names and priorities are high only in Japan. (Now the even use totally different local map provided by Korea company in Korea)

littlebtc:
I pick London again as its railway network seems to be more complex than Berlin.
I pick z15~z19 as they seem to be the most controversial part.

All is using Kosmtik export with x1 scale.

z15
london-before-z15
london-after-z15

z16
london-before-z16
london-after-z16

z17
london-before-z17
london-after-z17

z18
london-before-z18
london-after-z18

z19
london-before-z19
london-after-z19

littlebtc:
Or, can we still render other names in z17 but with very large `text-spacing` e.g. 1000?

I think it may be a good workaround, but I have to sleep now.

kocio-pl:
Thanks for proper rendering exports- it makes it much easier for me to compare changes!

What I think now is:
1. I want no subway line names - they look too messy even on z19.
2. I don't want any "track" names on any zoom level - platform names/refs are what we want on general map and we have them already.

Using bigger text spacing may be nice too, since it would not clutter urban areas too much, while still being visible as a hint on the outdoor.

littlebtc:
Trying `text-spacing=1200` and it looks much better for me in London.

For z17+, tomorrow I will try:
* highspeed and usage=main for `text-spacing`=600
* Others, `text-spacing`=1200

But I have to sleep. 🛌

littlebtc:
For `tracks` in the map, I think that could be a tagging question since in that station both `name` and `track_ref` are tagged but I did not seen that kind of over-tagging in other stations.

😪

Nakaner:
`railway:track_ref` is not the suitable tag to be rendered by a map whose users are not only railway (infrastructure) enthusiasts. There are lots of stations where the track numbers signed on the platform and the track numbers used by the infrastructure operator are different. The track numbers used by the operator of the infrastructure can be derived from the `ref=*` tags of the signals.

Rendering `railway:track_ref` will lead to situations where people will make fun about OSM because it renders "wrong numbers" and others trying to "correct" them. The only usuable reference numbers are tagged on platforms and stop positions.

Rendering the reference numbers of tracks in stations is difficult to get it right. Stations are locations where there are usually more parallel tracks than outside the stations, i.e. the labels of the tracks would be on rendered on another track and difficult to read.

littlebtc:
So I believe the track_ref tagging on `name` is a tagging mistake made by some user.

DerDakon:
If you refer to the "names" shown in the Warsaw and St. Pancras images, that just say "track N": yes, they shouldn't be there. The [OpenRailwayMap rules for JOSM validator](https://github.com/OpenRailwayMap/OpenRailwayMap/tree/master/validator) warn about this for a long time (for German, English, and French).

littlebtc:
First, I think current tagging of `ref` is too much different from countries (very good on Europe but not others), so I will only try to render `name` in this pull request.

A lot of changes implemented to make a better map and lesser impact:

- Bigger text-spacing:
  - `highspeed=yes`: z11=300, z14=600
  - `usage=main`: z14=300, z17=600
  - All others: z17=900
- Adding `text-min-distance: 18` for all labels to make then less crowded.
- All labels colored `#333333` -> `#666666` and default halos added.

By following is the before/after renderingfor the current result.

Now I taking London and Tokyo for example, fell free to provide a better sample area. 😄

**London**
King's Cross, z15
london-z15-before
london-z15-after

King's Cross, z17
london-z17-before
london-z17-after

**Tokyo**
Central Tokyo, z14
tokyo-z14-before
tokyo-z14-after

Ueno and Akihabara, z15
tokyo-z15-before
tokyo-z15-after

Ueno, z17
tokyo-z17-before
tokyo-z17-after


-------------
To me, central Tokyo improvement on z14 is very impressive 😃

roads.mss
```diff
@@ -3106,3 +3106,111 @@ tertiary is rendered from z10 and is not included in osm_planet_roads. */
...
+        text-dy: 7;
+      }
+      [zoom >= 19] {
+        text-size: 12;
+        text-dy: 8;
+      }
+    }
+  }
+  /* Other minor railway styles.
+     For service rails, see #2687. */
```
kocio-pl: I would give full URL here:
https://github.com/gravitystorm/openstreetmap-carto/pull/2687

kocio-pl:
Much better version, indeed. I have spotted only single cases where railway labels are covering some other labels, which is very good in such big cities.

Hi-speed in Poland on z11 looks reasonable (spacing could be bigger, but changing this option doesn't work here because of many segments): 
t5cyotvm

Thanks to the color change railway=rail labels on z17 are no longer competing with street names:
meobvaks

When the named road and tram line are shared, the rail label is sometimes visible. Road name takes precedence, which makes the problem less severe, but we should not render railway label at all in such cases:
qpnv2cnk
19rmgvr_

DerDakon:
If the same way is used for both the highway and the tram then the railway label simply should never be shown. It's probably not the correct name for the railway anyway.

kocio-pl:
Try to use z_order instead of priority table, see https://github.com/gravitystorm/openstreetmap-carto/pull/2732.

littlebtc:
Review comments committed.

----------
To me, the major concern is the railway names rendered bad in multiple-track railway mixed together:
bad_railway_name_rendering

Is there any way to improve rendering like this case?

----------
Another confusing issue is that in the current rendering, subway is almost invisible under highway, but names from subway would show. Can we improve subway rendering a little bit?

roads.mss
```diff
@@ -3106,3 +3106,111 @@ tertiary is rendered from z10 and is not included in osm_planet_roads. */
...
+  [railway = 'rail'],
+  [railway = 'subway'],
+  [railway = 'narrow_gauge'],
+  [railway = 'light_rail'],
+  [railway = 'preserved'],
+  [railway = 'funicular'],
+  [railway = 'monorail'],
+  [railway = 'turntable'],
+  [railway = 'tram'] {
+    [zoom >= 17] {
```
kocio-pl: You need to add something like [highway = null] to avoid road names repeated as railway labels.

littlebtc: Running this on SQL side may be better?

kocio-pl: I don't know, but probably yes.

littlebtc:
I try to change all `text-dy` to 4 as railways are much lighter thant highway.

Testing how this will effect rendering.

kocio-pl:
You can drop here just some interesting rendering samples, full overview would be too much images and not necessary.

littlebtc:
Reverted that approach as it seems too crowded.

littlebtc:
Export feature from kosmtik seems to be broken...

After several approach and failure, I had only changed the `text-dy` for `highspeed=yes`. z11 and z12 now has `text-dy=3` as the rails are relativlely think in these two zoom level.





Other rendering problems seem too hard to have a easy fix.

roads.mss
```diff
@@ -3106,3 +3106,114 @@ tertiary is rendered from z10 and is not included in osm_planet_roads. */
...
+#railways-text-name {
+  /* Mostly started from z17. */
+  [railway = 'rail'],
+  [railway = 'subway'],
+  [railway = 'narrow_gauge'],
+  [railway = 'light_rail'],
+  [railway = 'preserved'],
+  [railway = 'funicular'],
+  [railway = 'monorail'],
+  [railway = 'turntable'],
```
DerDakon: A turntable is something like the railway equivalent of a parking lane crossing or so. I don't think any of them can have a useful name.

eruedin:
This looks great, but not for a general map. If it is for the public transport map or for the railway map, that  are the information I am looking for, but not on the general map. I think, that is the reason, why we have a general map, a bicycle map and a public transport map. They show different things for different needs.

kocio-pl:
I'm inclined to have this code. It was too visible in the first draft, but has been tuned a lot in subsequent steps and it's not disturbing for me any more. 

Of course it would be out of question to show everything like OpenRailwayMap or as prominently as the transport layer. Refs don't belong here for example. But names are what people can identify - for example some rail lines are touristic attractions or locally important transport brand (including trams) and hi-speed lines are special feature worth showing in the outdoor. I would however not show any service lines.

I still hesitate when it comes to underground lines. I see some analogy to quarry - we show only overground buildings and areas. For underground passengers stations (and even entry numbers) is enough I guess.

kocio-pl:
I think that names in tunnels also shouldn't be rendered, because we can't avoid situations like this:
9arevcdw

roads.mss
kocio-pl: Sure, it doesn't belong here.

math1985:
I tested this, it works as described. I'm fine with merging it (once @kocio-pl's remarks are taken into account).

littlebtc:
Comments addressed in latest pushes.

kocio-pl:
Subway will cause the same problems as tunnels, so should be removed too.

And the last thing - please join all the commits into just one.

littlebtc:
Some subway is ground-leveled or elevated, so I think subway is OK.

2017年8月19日 下午6:57，"kocio-pl" <notifications@github.com>寫道：

kocio-pl:
I'm aware if this, but I guess they are under ground most of the time. It's not about just trains - I think about [hiding most level<>0 objects](https://github.com/gravitystorm/openstreetmap-carto/issues/1977).

littlebtc:
Rebased and squashed all commits into one.

For subway I think current status should be better, since there should have lots of >5km elevated/ground-level subways.

pnorman:
Subway should have tunnel if it's below ground. I'd wouldn't rule out showing underground subway line names, but am not sure. Probably best to treat them the same for the initial version.

kocio-pl:
OK, you're right. I guess there will be some work to be done, like moving false names into proper refs and maybe some other issues to be fixed later, but we'll see.

kocio-pl:
Thanks a lot for this PR and the work with tuning it!

PR #2694, opened 2017-07-20
Fixing .env file creation for Docker

kocio-pl:
Follow up to https://github.com/gravitystorm/openstreetmap-carto/pull/2679#issuecomment-313905873.

The only strange thing is that when there is no .env file, import fails, even though this file should be just created. Next time the import is invoked, everything works. That's why I marked it as a work in progress.

### DIFF
--- a/.env
+++ b/.env
@@ -1,5 +1,0 @@
-# environment defaults for import
-
-OSM2PGSQL_CACHE=512
-OSM2PGSQL_NUMPROC=1
-OSM2PGSQL_DATAFILE=data.osm.pbf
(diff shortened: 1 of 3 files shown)


nebulon42:
Problem is the following: if there is no .env file then the environment variables that osm2pgsql uses are not set. In addition to creating the .env file you also have to set the environment variables by issuing `export OSM2PGSQL_CACHE=512` and so on. Setting the variables should not happen if there is a .env file.

kocio-pl:
Thanks! I was suspecting it, but had no more time to test. Now it's fixed and working as expected.

kocio-pl:
I guess some other developer should review this code before merging?

Issue #2688, opened 2017-07-12
Improve low zoom levels


kocio-pl:
Most of our work goes to high zoom levels, but it's time to look how we could improve also lower levels. [Mid zoom work](https://github.com/gravitystorm/openstreetmap-carto/issues/2194) seems to be pretty advanced now, with PR mostly ready to be merged (#2654). Faded out landuse colors could be used also for low zoom IMO (and new water color can have some impact), but that's just a general idea and there can be other things which need to be taken into account (like performance problems or issues related to using external and pre-rendered data).

What we have now in low zoom levels is better rendering of cities/capitals (z4+) and new road colors (z5+), but for example z0-z3 is still underused. Discussion on improving low zoom levels has been [already started here](https://github.com/gravitystorm/openstreetmap-carto/pull/2654#issuecomment-314688048).

Some related resources:
- http://fred.dev.openstreetmap.org/lowzoom/
- https://lists.openstreetmap.org/pipermail/talk/2012-March/062286.html
- https://wiki.openstreetmap.org/wiki/User:Kocio/low_zoom_rework (my old summary from 2015)
- https://github.com/gmgeo/osm-carto-lowzoom (tool for dropping tags from extracts which are not visible on low zooms)

Some issues that are close to the subject:
- https://github.com/gravitystorm/openstreetmap-carto/issues/2278
- https://github.com/gravitystorm/openstreetmap-carto/issues/2507
- https://github.com/gravitystorm/openstreetmap-carto/issues/2293
- https://github.com/gravitystorm/openstreetmap-carto/issues/1448
- https://github.com/gravitystorm/openstreetmap-carto/issues/1770
- https://github.com/gravitystorm/openstreetmap-carto/issues/1935

kocio-pl:
Oh, I've just found that issue with such name has been already created and closed because some improvements has been made then (#1125). Should we reopen it then?

woodpeck:
Depending on just how wide you want to throw open this issue: z0-3 receive relatively little attention because they are "world map" scale, and the projection we are using is singularly unsuitable for a world map. I know there's a lot of political discussion about "correct" world maps (Gall-Peters et al., cf. https://xkcd.com/977/) but that's not even what I'm talking about; just look at Greenland on z2 and you don't need to be a GIS expert to see that it is absolutely ridiculous, and any sweat invested in dressing up the map with nicer styling cannot hide the fact that it is a bad map (on that scale).

[...]

Issue #1749, opened 2015-08-15
borders up to z3 are grey instead of purple


nebulon42:
Is there a specific reason why borders up to z3 are grey instead of purple?

mboeringa:
Probably a visual thing. Maintaining exactly the same colors and contrast across 20 zoomlevels, is bound to lead to problems.

nebulon42:
If you consider this a bug, I will submit a PR for changing it.

PR #2695, opened 2017-07-22
Changing borders color to gray [WIP]

kocio-pl:
Related to https://github.com/gravitystorm/openstreetmap-carto/issues/622.

Simple color change for the borders from violet to gray. Examples:

Warsaw, z12
Before
af5zg3i3
After
huj_btnw

Europe, z4 (see [more examples here](https://github.com/gravitystorm/openstreetmap-carto/issues/2688#issuecomment-317144397))
Before
ubx5_svd
After (borders)
7uehxsqz
After (borders+water)
negfihhr

### DIFF
--- a/admin.mss
+++ b/admin.mss
@@ -1,4 +1,4 @@
-@admin-boundaries: #ac46ac;
+@admin-boundaries: #555555;
 
 /* For performance reasons, the admin border layers are split into three groups
 for low, middle and high zoom levels.


HolgerJeromin:
Crowded city centers would be interesting

Tomasz-W:
It looks very good on low- and mid-zooms, but I'd like to see some high-zoom examples, especially places where borders crosses or goes along other elements of map

kocio-pl:
Could you give some examples to test?

Tomasz-W:
A few places found fast, if you need, I can give more of it later
http://www.openstreetmap.org/#map=15/52.4786/16.7802
http://www.openstreetmap.org/#map=15/52.2668/16.9386
http://www.openstreetmap.org/#map=18/51.62556/14.75621
http://www.openstreetmap.org/#map=18/52.42021/16.92300
http://www.openstreetmap.org/#map=16/52.3220/16.6323

kocio-pl:
The most convenient would be something included in Metro extracts.

imagico:
The most serious problem at the low/mid zooms with grayish borders is interference with railways at z7 and above.

The German style using a somewhat greenish tone (much less green though than the old German style) is doing a bit better here but not really great:

https://mc.bbbike.org/mc/?lon=9.948721&lat=51.108173&zoom=7&num=2&mt0=mapnik-german&mt1=mapnik

At the high zooms you are also going to have interference with cliff:

https://mc.bbbike.org/mc/?lon=6.745563&lat=46.339865&zoom=13&num=2&mt0=mapnik-german&mt1=mapnik

cc @giggls.

math1985:
Note that this partially reverts #1749 / https://github.com/gravitystorm/openstreetmap-carto/commit/5a0f0362f096fe91cd6aaa6501d597f5c5d26ea3 .

kocio-pl:
Thanks for raising these issues. Borders rework may include different pattern and changing levels like it's discussed in #622. I'd like to start with extending new midzoom landuse colors for low zoom to see how they will interact with borders. Green is unlikely (there will be more trees), but maybe gray+violet tint could work?

@nebulon42 [wants to publish lowzoom tool which will save z8 landuses](https://github.com/gmgeo/osm-carto-lowzoom/issues/3), so testing can take only reasonable time.

kocio-pl:
There are many other changes related to midzoom going right now, so I close this PR. Borders needs probably more in-depth discussion anyway than this trivial color change.

kocio-pl:
@imagico On the other hand we have other confusing things currently: #1248.

Issue #2557, opened 2017-01-22
Render name for waterway=dock


boothym:
I found a previous discussion on #2088 about floating dock rendering, and it was suggested to open a new issue for just the name, so here is that new issue. 

Some docks were recently edited to add the name, but it is not rendered - for example: https://www.openstreetmap.org/way/467652902. I think it should just be rendered the same as other water features.

boothym:
@math1985 (or anyone else who is active here!)

Any chance this change can be made? I'm unsure of exactly what to edit otherwise I'd submit the PR myself (though I wouldn't be able to render it).

SomeoneElseOSM:
@boothym - what would be useful wold be to look at the usage of waterway=dock worldwide and to see what type of "dock" they refer to, since "dock" has different meanings in different varieties of English.  If the same tag is used for physically different features, there's a problem that needs fixing before it can be rendered.
- Andy

(not a maintainer here but mentioning it since I looked at "dock" among other features such as "harbour" for another syle).

boothym:
@SomeoneElseOSM well that's a different issue altogether, which needs some people to go through all the docks to confirm whether they should be piers or not.

However waterway=dock areas are already rendered the same way as water - I just want the name added as well.

PR #2699, opened 2017-07-23
Rendering name for waterway=dock

kocio-pl:
Resolves https://github.com/gravitystorm/openstreetmap-carto/issues/2557.

Gdańsk, z17
Before
ajwaysud
After
1thfwcba


imagico:
See #2701 - since i think we should drop rendering `waterway=dock` in general i also think we should not extend the current rendering.  Especially incorrect use for mapping open harbor basins (like apparently in the example shown) would be encouraged by rendering labels.

math1985:
I agree this example is not a dock, either according to English language or according to osm's wiki.

kocio-pl:
> in general i also think we should not extend the current rendering.

This code is about rendering names with standard water label style, so I think it doesn't matter what kind of object it really is if it has a name and is related to water. I believe #2701 needs separate PR related to rendering water area, which is a different issue.

PR #2700, opened 2017-07-23
Rendering man_made=works

kocio-pl:
Resolves https://github.com/gravitystorm/openstreetmap-carto/issues/887.

Warsaw, z18
Before
s5pama4i
After
vvfegrv5

### DIFF
--- a/amenity-points.mss
+++ b/amenity-points.mss
@@ -1504,6 +1504,7 @@
   [feature = 'landuse_retail'],
   [feature = 'landuse_industrial'],
   [feature = 'landuse_railway'],
+  [feature = 'man_made_works'],
   [feature = 'landuse_commercial'],
   [feature = 'landuse_brownfield'],
   [feature = 'landuse_landfill'],
@@ -1599,7 +1600,8 @@
         text-fill: darken(@retail, 50%);
       }
       [feature = 'landuse_industrial'],
-      [feature = 'landuse_railway'] {
+      [feature = 'landuse_railway'],
+      [feature = 'man_made_works'] {
         text-fill: darken(@industrial, 60%);
       }
       [feature = 'landuse_commercial'] {
--- a/landcover.mss
+++ b/landcover.mss
@@ -371,6 +371,14 @@
     [way_pixels >= 64] { polygon-gamma: 0.3;  }
   }
 
+  [feature = 'man_made_works'][zoom >= 16] {
+    line-width: .5;
+    line-color: @industrial-line;
+    [name != ''] {
+      line-width: 0.7;
+    }
+  }
+
   [feature = 'landuse_railway'][zoom >= 10] {
     polygon-fill: @railway;
     [zoom >= 16][name != ''] {
--- a/project.mml
+++ b/project.mml
@@ -124,7 +124,7 @@ Layer:
       table: |-
         (SELECT
             way, name, religion, way_pixels,
-            COALESCE(aeroway, amenity, wetland, power, landuse, leisure, military, "natural", tourism, highway, railway) AS feature
+            COALESCE(aeroway, amenity, wetland, power, landuse, leisure, man_made, military, "natural", tourism, highway, railway) AS feature
           FROM (SELECT
               way, COALESCE(name, '') AS name,
               ('aeroway_' || (CASE WHEN aeroway IN ('apron', 'aerodrome') THEN aeroway ELSE NULL END)) AS aeroway,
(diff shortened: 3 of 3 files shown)


imagico:
As previously said multiple times i am against point label rendering of area features based on way_area if those are not at the same time rendered as areas.  This would create a bad mapping incentive.

math1985:
> As previously said multiple times i am against point label rendering of area features based on way_area if those are not at the same time rendered as areas.

:+1:, this should be changed. However, this is an existing problem which is IMO outside of the scope of this PR.

imagico:
> However, this is an existing problem which is IMO outside of the scope of this PR.

It is not, as far as i can see `man_made=works` is currently not rendered in any way, this PR adds rendering on point features (which is fine) and on area features with rendering being influenced by way_area (which i do not support).

math1985:
As that is wider than only man_made=works, could you create a new issue for that?

imagico:
I am not aware of any case currently where label styling is based on way_area for features that are not rendered as areas - unless i missed some changes that introduced this.

You could formally count place=island since the tag itself is not rendered but it is universally rendered as negative space through the coastline and water areas.

math1985:
Sorry, I totally misunderstood you. I think shop=mall is another current example.

imagico:
Good point, did not notice that one.

Could be fixed by rendering shop=mall (when not also building=yes) like landuse=retail in the landcover layer - see #2702.  Same could be done for this one (i.e. rendering it like landuse=industrial).

math1985:
:+1: for both suggestions.

HolgerJeromin:
works should be inside industrial area. So rendering only the name and not the area is on purpose. same as power plant.

kocio-pl:
@imagico Would rendering the no-building works area the same as industrial be OK for you?

polarbearing:
> rendering the border (but no fill)

That would mean a landuse=industrial without a man_made tag has a different border than with, which is inconsistent. It would also make it more difficult to recognise a mapped barrier=fence/wall. 

I find rendering the landuse with a fill sufficient. If it is a large plant with a name, it already has the name in way size from the landuse. If there are several small man_made=works in one larger landuse, rendering the small names is sufficient, as @HolgerJeromin pointed out.

dieterdreist:
sent from a phone

> On 24. Jul 2017, at 08:53, polarbearing <notifications@github.com> wrote:
> 
> I find rendering the landuse with a fill sufficient. If it is a large plant with a name, it already has the name in way size from the landuse. If there are several small man_made=works in one larger landuse, rendering the small names is sufficient, as @HolgerJeromin pointed out.


IMHO this style focuses too much on landuse tags anyway, which leads to many (rendering oriented) mappers adding very few and rough further semantic information, as they have already "drawn everything" with the landuse.

HolgerJeromin:
> That would mean a landuse=industrial without a man_made tag has a different border than with, which is inconsistent. 

This is not correct.
Two industrials which are sharing the same border are visible. This would not be the case if there was no border color. See the north border of:
https://www.openstreetmap.org/way/91224778

pnorman:
I *think* the concerns raised have been resolved. @imagico, is this so?

HolgerJeromin:
After read the code this i think features with only `man_made=works` without `landuse=industrial` would render only the name. No border and no fill. Am I right?

The dimension of [this](https://www.openstreetmap.org/way/408439828) would not be visible, or would it?

imagico:
> I think the concerns raised have been resolved. @imagico, is this so?

No changes were made since the PR was originally started so the answer is no.

kocio-pl:
What is desired solution then? Do we have an agreement here?

kocio-pl:
> After read the code this i think features with only man_made=works without landuse=industrial would render only the name. No border and no fill. Am I right?

Yes. Because there are no changes in `landcover.mss`, no such area would be colored nor outlined.

> The dimension of this would not be visible, or would it?

It wouldn't.

HolgerJeromin:
Oh, i thought that power=plant got a text and outline ([example](https://www.openstreetmap.org/way/53243836)). I was wrong. 

So with the current PR state, `works` will be rendered exactly like `plant`. Kind of ok for me.

imagico:
Interesting observation about power=plant - did not realize that.  That is the same situation as shop=mall.  

My opinion is as before:

> As previously said multiple times i am against point label rendering of area features based on way_area if those are not at the same time rendered as areas. This would create a bad mapping incentive.

> Could be fixed by rendering shop=mall (when not also building=yes) like landuse=retail in the landcover layer - see #2702. Same could be done for this one (i.e. rendering it like landuse=industrial).

Other possibilities would be rendering with an icon or rendering with a point label independent of way_area for both point and area features.

HolgerJeromin:
Rendering (on plant and works) an outline (same color like industrial outline) would show the dimension of a feature but still give a feedback that an landuse=industrial is missing.

daganzdaanda:
> Rendering (on plant and works) an outline (same color like industrial outline) would show the dimension of a feature but still give a feedback that an landuse=industrial is missing.

That would be a good solution, also for offshore wind parks like https://www.openstreetmap.org/way/186609785 where landuse=industrial is just wrong.

kocio-pl:
Thanks for a good example - it's hard to find them, because with so many rules Overpass Turbo is hitting the RAM limits. There's only a name and the outline, no filling, but if you want to make sure, I need another location where there's no industrial area. In the meantime I have removed industrial area filling+outline rendering rule just to show you how would it look like:

Verl, z18
Before
87zvg9zp
After
jrplgd_j
After (without industrial filling)
vrq ne50

HolgerJeromin:
https://www.openstreetmap.org/way/53243836 has no underlying industrial (but should :-)

kocio-pl:
But this is a different tag than man_made=works, so it won't fly with this code. Of course we can make another PR for power plats and other possible objects, but for now I'd like just to make sure if name+outline is the way to go for such cases.

HolgerJeromin:
Here you go: https://www.openstreetmap.org/way/29388213 :-)

kocio-pl:
So here you are :smile: :

Erkrath, z19
Before 
giiifyvg
After
pqebzzty

HolgerJeromin:
One question: does this render nodes with this tag?

kocio-pl:
Yes, the original example is a node:

http://www.openstreetmap.org/node/3428336388

HolgerJeromin:
Btw, this is missing in the changelog

kocio-pl:
Also some other things (like pattern on natural=sand), but Paul never did very detailed changelogs IIRC.

PR #2707, opened 2017-07-25
Changing area color for railway=station

kocio-pl:

Currently we're using special color for rendering railway=station area, but it looks like a leftover when it was considered to be a building (like supermarket for example). In general I'd like to unify all transportation areas, but in this case I think railway/industrial color works better, as it is a railway area anyway.

The question is however if we want to imply a landuse color or not fill it at all (so the explicit landuse tagging is needed)? It's the same problem as with https://github.com/gravitystorm/openstreetmap-carto/pull/2700.

Gliwice, z17
Before
z26fuayx
After
xkgvhe_a

### DIFF
--- a/landcover.mss
+++ b/landcover.mss
@@ -31,7 +31,6 @@
 @railway: @industrial;
 @railway-line: @industrial-line;
 @rest_area: #efc8c8; // also services
-@station: #d4aaaa;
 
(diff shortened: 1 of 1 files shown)


kocio-pl:
Any comments/reviews for this PR?

PR #2708, opened 2017-07-25
leaf_type rendering

### LINKED ISSUE #822: Distinguish deciduous and coniferous forests

sommerluk:
leaf_type rendering

Uses the same pattern for leaf_type=mixed and no leaf_type tag at all. Uses the same pattern (but different symbols) to preserve a certain visial unity:

screenshot 5

Resolves #822

Closes #2747 

No support for leaf_cycle.

Cannot be backported to version 3.


imagico:
I see no point in showing symbol pairs if leaf_type is tagged.  But especially it is important to distinguish between `leaf_type=mixed` and untagged leaf_type - that was the whole point in having tree pairs in the first place.

Also if you use the same pattern density for the different types you should use the same pattern (i.e. the same random seeds).

See #1242 for the previous discussion on the matter as well as various ideas on how to render both leaf_type and leaf_cycle.

kocio-pl:
@sommerluk Great, thanks for this work! Leafless icon is clever, but may need some tuning, because it looks to me like a lightening symbol until I look closer.

I agree with @imagico about not using pairs and about reusing the same pattern.

imagico:
Regarding `leaf_type=leafless` - while this certainly could be rendered if a suitable symbol can be found it should be noted that there seems to be very little correct use of this tag - a bit more than 300 uses on areas, most of them in regions with no indigenous leafless vegetation and a lot of cases of `leaf_type=leafless` + `leaf_cycle=deciduous` - which does not really make much sense.

sommerluk:
screenshot 1

Using now different symbols for unknows vs mixed leaf type.

Same pattern for all leaf types, only the symbol varies.

Still no special rendering for the leaf cycle. Proposals in #1242 about leaf cycle rendering is basically changing the colour. I think this is a little bit too much for this PR, but could easily be done later.

imagico:
Much better IMO.  And i agree that rendering leaf_cycle can be well separated from this.

I would still encourage some discourse on different symbol style options.  This is mostly a subjective question of personal preferences but there might also be some objective difference in recognizability.

Something else i wonder is if the pair pattern and the individual tree patterns should have the same overall weight, i.e. if the individual tree patterns should me more dense relatively speaking.  This would be of particular importance if leaf_cycle rendering via color is introduced later because with the current suggestion a color difference would be significantly better readable with no leaf_type which would not be a positive mapping incentive.

sommerluk:
Yes. And maybe making the individual symbols (except the “unknown leaf type” symbol) simply bigger could help to get the right balance. This would make them also easier to recognize (except “unknown leaf type”, which is a good thing to be less readable).

Proposals for different icons?

kocio-pl:
I like the current result very much - all the shapes are recognizable for me now and it's easy to tell when the type is unknown. If you want to further improve it, go for it, but I'm already happy. There are icons in jsdotpattern like trees1, trees2, trees3 or forest, which you could try for needleleaved.

BTW - real life examples of current proposition:
mixed, z14
sxhh35_o
needleleaved, z14
z7abvds
broadleaved vs unknown, z15
rvper1xv
leafless, z18
dpyjdizh

imagico:
> Yes. And maybe making the individual symbols (except the “unknown leaf type” symbol) simply bigger could help to get the right balance. This would make them also easier to recognize (except “unknown leaf type”, which is a good thing to be less readable).

Note the symbols you currently use are designed specifically for this resolution.  

> Proposals for different icons?

There have been a number of different ideas tested in #1242 but there is no need to limit this to those designs.

kocio-pl:
> > Proposals for different icons?

> There have been a number of different ideas tested in #1242 but there is no need to limit this to those designs.

I've just found that if we want to go with https://github.com/gravitystorm/openstreetmap-carto/issues/822#issuecomment-131911549, broadleaved, needleleaved and mixed types need to have the outline to be filled (fully for semi_/evergreen, halfway for semi_/deciduous or none for unknown).

pnorman:
cc @sk53

turnsole80:
Could the icons be more numerous and closer together?

SK53:
Sorry for not responding earlier. In general this looks very good. Some specific comments:

- I  agree that the symbols could be made bigger.
- I'm happy with current spacing, the only place it falls down is for small areas of woodland.
- I hadn't encountered the leafless key but find it interesting for two potential categories in the future: areas affected by burning or disease. I don't think there is any need to handle it now. At close range the symbol looks rather like one of those large cacti, perhaps the branches should be angled further from the trunk, but upwards, keeping downward sweeping branches for conifers. Also, in practice leafless areas will need some diminution of the woodland green. The only cartographic approach I am aware of for this (green & white stippling, in the UK Phase 1 Habitat Survey) is not really relevant as it is for thematic ecological maps.
- Allowance should be made for palm trees. Although there are a tiny number of elements tagged with leaf_type=palm or palm_tree there are an outstanding [37k examples](https://taginfo.openstreetmap.org/tags/type=palm) of type=palm from before the existence of leaf_type. Again this is no reason to reject this PR: just something to bear in mind.
- @BushmanK's symbology for leaf_cycle looks good, but as @kocio-pl remarks there are 9 or more combinations (if palm trees are factored in) to manage. Clearly this will need to be examined independently.

All in all this represents a significant advance in encouraging better tagging of woodland. Bravo!

imagico:
To make this clear: `leaf_type=leafless` is meant for naturally leafless vegetation and not for dead trees having lost their leaves:

https://wiki.openstreetmap.org/wiki/Tag:leaf_type%3Dleafless

The majority of current use is however probably indeed for the latter.  It is safe to say that neither correct nor incorrect use of this tag is really well established.

Nearly all uses of `type=palm` by the way are for `natural=tree`.  The documented leaf_type for palm trees is broadleaved.  I don't think there is an established tagging for palm tree forests at the moment outside the species/genus/taxon domain.

SK53:
@imagico Thanks for putting me right. In particular the reason why the symbols look like cacti is that they do represent cacti! My reaction then is: "Are there any woodlands dominated by such plants?" Do we need to worry?

I had rather assumed that the palm usage from type had been carried over to leaf_type. As the most widespread practice is to map palm plantations as orchards this may explain why it hasn't appeared in the context of woodland. (One reason I thought it might, is that palms are readily recognised from aerial imagery).

For completenesses sake, I presume cycads can also be treated as leaf_type=broadleaved as well.

imagico:
> My reaction then is: "Are there any woodlands dominated by such plants?" Do we need to worry?

Probably not.  Such leaf_type will be fairly common with scrubland in the southwestern US and Mexico (and also southern Africa if you include various Euphorbia - which are often not strictly completely leafless though).

> As the most widespread practice is to map palm plantations as orchards this may explain why it hasn't appeared in the context of woodland.

Yes, larger areas with palm trees are probably mostly plantations.  But if in the future a separate `leaf_type=palm` is established (which is not unlikely i suppose) that would not lead to problems.  The only thing to consider here is that whatever icon design is chosen it would be good if it could be extended to other tree types without a breach in styling

sommerluk:
At #2747 there is an impelemtation with slightly bigger symbols available now.

sommerluk:
For #2747 the pattern has switched to 256 px. (Only) If necesarry I could do it here also.

Issue #2710, opened 2017-07-27
Docker fails when trying to run containers


Salatfreak:
Running Docker 17.06.0-ce on 64bit Linux 4.9.38.1-MANJARO

Following the instruction in DOCKER.md leads to an error. It is not mentioned, that a docker machine needs to be created as well.
Here is what I do:
```
# Clone repository
git clone https://github.com/gravitystorm/openstreetmap-carto.git
cd openstreetmap-carto

# Copy osm data
cp ../<extract>.osm.pbf ./data.osm.pbf

# Create and start docker machine
docker-machine create default

# Configure environment
eval $(docker-machine env default)

# Import data
docker-compose up import
```
Everything downloads and installs until:
```
...
Creating openstreetmapcarto_db_1 ... 
Creating openstreetmapcarto_db_1 ... done
Creating openstreetmapcarto_import_1 ... 
Creating openstreetmapcarto_import_1 ... done
Attaching to openstreetmapcarto_import_1
import_1   | sh: 0: Can't open scripts/docker-startup.sh
openstreetmapcarto_import_1 exited with code 127
```
Rerunning `docker-compose up import` results in
```
openstreetmapcarto_db_1 is up-to-date
Starting openstreetmapcarto_import_1 ... 
Starting openstreetmapcarto_import_1 ... done
Attaching to openstreetmapcarto_import_1
import_1   | sh: 0: Can't open scripts/docker-startup.sh
openstreetmapcarto_import_1 exited with code 127
```
immediately.

math1985:
Thanks for reporting. Which extract did you use? Could you try with the Luxembourg extract?

Salatfreak:
I used the current Germany extract from geofabrik.de. I don't think it even starts the import process. It takes about 2 seconds to fail.

Salatfreak:
As expected, the same thing happens with the current Luxembourg extract from geofabrik.de.

math1985:
Is anybody able to reproduce this?

Salatfreak:
I installed the Docker Toolbox on Windows 7 on the same machine. Both the procedure and the resulting error are the same when trying to import the data.

Salatfreak:
The same thing happens if I use a [version](https://github.com/gravitystorm/openstreetmap-carto/commit/9aacdb3b37fca21869d59d4fa79d1336b924dbcd) from 18 days ago. Has it been tested since then? Is there maybe some additional configuration required?

[...]

PR #2712, opened 2017-07-27
Adding quick Docker testing procedure in DOCKER.md

kocio-pl:
Resolves https://github.com/gravitystorm/openstreetmap-carto/issues/2710.

A short note to make sure that user is able to connect to a Docker daemon and run containers.

### DIFF
--- a/DOCKER.md
+++ b/DOCKER.md
@@ -16,6 +16,7 @@ Docker itself. Otherwise you need to [install Docker Compose manually](https://d
 If you are eager to get started here is an overview over the necessary steps.
 Read on below to get the details.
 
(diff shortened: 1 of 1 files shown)


math1985:
I've seen the discussion, I'm inclined to consider this out of scope of this repository.

kocio-pl:
I have no clear idea about it, you can do what you feel would be right.

nebulon42:
Yes it is out of scope, but does not hurt either. :)

pnorman:
I think it's out of scope for what we should be documenting.

kocio-pl:
OK, we might skip it.

Issue #2666, opened 2017-06-30
Release v4.1.0


kocio-pl:
Now that big release v4.0.0/v3.3.1 is out in the wild, we can think what do we want to land in v4.1.0, what is not ready and needs some love or should wait for v4.x.0 to not break v3 compatibility. [Current changes](https://github.com/gravitystorm/openstreetmap-carto/compare/v4.0.0...master) are just small fixes, but there are some PRs waiting few weeks already.

Our [roadmap](https://github.com/gravitystorm/openstreetmap-carto/blob/master/README.md#roadmap) is still in v3, so it would be good to update it to reflect the change.

daganzdaanda:
Is it now possible to start working on the issues that needed hstore?
I thought there was a label once for "needs hstore" or "needs database reimport", but I seem to be wrong. Should we check all the old open issues systematically?
...
Found this list in #1504:

**Keys to be added**

Adding hstore will avoid the need to manually add these keys.
emergency #318, #1160
entrance #740, #517
golf #661
addr:flats #264, #605
allotments #432
area:highway #180
mountain_pass #244
ford #267
backcountry #1485
memorial:type #1066
lock_name #828
bridge_name #828
leaf_type #822
int_name #803
basin #709
location #552, #640
protect_class #603
proposed #592
information #304, #324
network #508
height #1017
generator:output:electricity #1017
lock_name #157
bus #311, #435
vehicle #214
motor_vehicle #214
motorcar #214

Maybe also related:
tower/mast types #339, 
tourism=artwork details #855,
craft=* #108, 
healthcare details #763 #1981,
leaf_type #1242
Also, in #70 the capacity tag is mentioned as being useful for rendering decisions.

imagico:
As you can see in the readme we do not want to add new features for the first few releases, especially not those which cannot be backported to 3.x.

But you can and in fact are welcome to start working on things that require tags previously not available.  However keep in mind that adding new feature will be evaluated carefully and critically.  Tags that allow making better decisions about things we already render (like ford, basin or protect_class) are of much more interest than completely new stuff that adds additional clutter to the map.

kocio-pl:
I'm curious how many PRs are there which satisfy v3.x compatibility and if there will be really enough of them to make "at least two MINOR releases" (as stated in README)?

HolgerJeromin:
> I thought there was a label once for "needs hstore" or "needs database reimport",  but I seem to be wrong. 

[...]

PR #2713, opened 2017-07-28
Dropping v3 compatibility plans in README.md

kocio-pl:
Related to https://github.com/gravitystorm/openstreetmap-carto/issues/2666.

I think we're ready to drop v3 compatibility plans now and go on with v4.1.0 release soon. If that's the case, project documentation should reflect it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,13 +66,8 @@ these often make it easier to customise, experience has shown that over-cleverne
 ## Database schema change (v4.x)
 
 The v4.x series includes [osm2pgsql lua transforms](https://github.com/openstreetmap/osm2pgsql/blob/master/docs/lua.md)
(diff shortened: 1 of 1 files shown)


math1985:
Oops, meant to approve instead of merge this. Anyway I don't think it matters much.

kocio-pl:
Sure. What matters to me then is if you'd like to release v4.1.0 now or you plan to do something with midzoom code (like merge it as a whole or maybe just some pieces)?

math1985:
Let's release 4.1.0 now (could you do that?). And immediately afterwards I will merge midzoom, so we have a couple of weeks to accept any PRs on the midzoom code before rolling it out to osm's servers.

Issue #2638, opened 2017-05-21
rendering of Muslim cemeteries


michalfabik:
Hi, can we please have Muslim cemeteries render like [this](https://wiki.openstreetmap.org/wiki/File:Landuse-cemetery_muslim.png)? As of now, we have patterns for [Christian, Jewish and atheist/other/unknown cemeteries](https://wiki.openstreetmap.org/wiki/Tag:landuse%3Dcemetery#Rendering), but none for Muslim ones. For example, [here](http://www.openstreetmap.org/#map=16/43.3336/17.8198), the cemetery to the south looks like an atheist one (as opposed to the Christian one to the north), but in fact it is a Muslim cemetery.  

I think this could improve the map in places like the Balkans, Israel, Egypt, Northern Caucasus, Indonesia and elsewhere where Islam coexists with other religions.

nebulon42:
Thanks for the pattern file. Yes, that would be possible. We might want to already use SVG pattern files or otherwise just use this.

imagico:
In principle i think this is a good idea - i would however like to verify that the hilal is considered an appropriate symbol for a cemetery by Muslim.  In Christianity the cross is obviously widely used as a symbol for graves and graveyards but i would not want to automatically assume the same applies in Islam for the crescent moon symbol.

As @nebulon42 indicated this would be a good opportunity to move the graveyard patterns to SVG - see #2045.

michalfabik:
I'm not sure about the .svg files: in the OSM Wiki, icons that are intended for nodes indeed are .svg files (e.g. [here](https://wiki.openstreetmap.org/wiki/File:Memorial-16.svg)), but patterns for areas/landuses are .png's. I did read the guidelines for reporting issues here and it did mention .svg files, but again, only for icons, nothing about areas. So I went with what the Wiki uses. Sorry if I did it wrong, I'm quite new to this.

mboeringa:
> In principle i think this is a good idea - i would however like to verify that the hilal is considered an appropriate symbol for a cemetery by Muslim. In Christianity the cross is obviously widely used as a symbol for graves and graveyards but i would not want to automatically assume the same applies in Islam for the crescent moon symbol.

Good question. The University of Texas library seems to have a nice collection of Middle East (topographic) maps online:

http://www.lib.utexas.edu/maps/israel.html

imagico:
[...]

PR #2714, opened 2017-07-31
Cemeteries rework in SVG

kocio-pl:
Resolves https://github.com/gravitystorm/openstreetmap-carto/issues/2638.

Warsaw, z19
Before
uf1lzgaa
After
blzubrge


kocio-pl:
This is how it would look with all cemeteries as SVG:
Before
5ywibasp
After
ruujmuxx

pnorman:
The icon density looks too high. I'd go for spacing them out more, and possibly adjusting the weight of them,

Can we deal with all the cemeteries in one PR? The above comment applies to all of the patterns.

kocio-pl:
And the weight seems to be because of less blur with vector icons, so I don't know what should we change. With 16x16 px matrix and 10x9 px symbols I'm also not sure how to tune it. 32 px matrix gives much higher spacing, smaller symbols might be unreadable. Maybe 18 px matrix was OK - why do you claim it should be power of 2? Or what other options do you propose?

Yes, if we'll find what we want to do, I can make one PR for all the cemeteries.

pnorman:
> why do you claim it should be power of 2?

The tile edges need to align with the pattern edges. This means the pattern size needs to be made from the factors of the tile size, which are all 2.

kocio-pl:
I thought so. But with 18 px elements it should be something wrong every (256/18=14.2...) 14 rows and columns, while I see just one strangely aligned column here (instead of expected about 5) and no strange rows (instead of expected about 2), which makes me think there's some other problem:

http://www.openstreetmap.org/#map=19/52.24621/20.97196

What do you think about it?

kocio-pl:
So what path you're inclined to follow - 16 px everywhere to avoid even slight aligning problems, but more dense for bigger symbols (jewish, muslim), 18 px with exactly opposite properties or maybe something else? Also what about weight - maybe we should just use lighter shade of green?

pnorman:
I'm inclined towards a 32px interval and keeping the visual weight of a single symbol the same. I've long felt the pattern is too dense, and that's feedback I often get from others.

If we wanted something closer to what we have right now, we could go with 7 symbols in a 128x128 grid, and a spacing of 18px between 6 of them. The other and where it loops around would then be 17px.

```
|  x 18 x 18 x 18 x 17 x 18 x 18 x |  17 x ...
```
is kinda what it would look like, with | being the 128px bounds, and the numbers being the distance between symbols, which are `x`.

But 32px is much easier.

kocio-pl:
> keeping the visual weight of a single symbol the same

If I understand you, it means probably that we have to use different shades of green, because christian symbol is pretty much the same, but jewish one is darker. I would rather use lighter shade for all of them.

pnorman:
> If I understand you, it means probably that we have to use different shades of green, because christian symbol is pretty much the same, but jewish one is darker. I would rather use lighter shade for all of them.

Keep the visual weight of the symbol the same as the existing symbols.

kocio-pl:
I would go with 32px version for all the cemeteries. If there are people who think it's already too dense, that's the best solution and we don't have to cheat with spacing, which makes the pattern simple. It also gives us more room to balance the size of the symbols. 

Visual weight is related to the color. shape, size and bluriness, so I would start with size (shape is beyond our control, bluriness also) and we might change the shade of green as the last step to make them less obtrusive. So I've made a cross bigger to match the size of more complicated symbols more or less (12x9 vs 10x9). They could be also make larger a bit if needed (12x11). What do you think now?

Before
bui1whb3
After (not touching symbols)
ob_fsdzw
After (making cross bigger)
rxbt0agn
After (making all symbols bigger, with h/w=12 px)
e8 j250z

kocio-pl:
I like 32 px matrix with bigger symbols (the last rendering). For the completeness - generic symbols rendering:

Before
mb5thlmp
After (making all symbols bigger, with h/w=12 px)
cjhdssp3

pnorman:
> So I've made a cross bigger to match the size of more complicated symbols more or less (12x9 vs 10x9). They could be also make larger a bit if needed (12x11). What do you think now?

I've had success with the existing size at 32px spacing, using https://github.com/kartotherian/brighmed/blob/master/symbols/patterns/cemetery.svg to get

image

That's a different colour scheme, of course, but I like the overall balance of it.

kocio-pl:
But what about different symbols? Do we accept the fact that christian/generic symbols would be small and jewish/muslim would be big or maybe also small but harder to recognize?

pnorman:
I'd go for not changing the symbols for now. Right now the symbols are changed

image

kocio-pl:
I have no idea how these PNGs were made and rasterization would be made by Mapnik now, so of course exact recreation is not possible.

Just to make sure - you'd like a version from [this comment](https://github.com/gravitystorm/openstreetmap-carto/pull/2714#issuecomment-319339467)?:

After (not touching symbols)
ob_fsdzw

kocio-pl:
It looks like jewish symbol for cemeteries was simply [imported](https://github.com/gravitystorm/openstreetmap-carto/commit/f5f8a8238d82ea29cd3b93ab0636300723f8cfdb) from previous (XML-based) style implementation as [`cemetery_jewish.18.png`](https://github.com/gravitystorm/openstreetmap-carto/blob/739298d645babdb9da5aee2ef6434ecac00035cb/symbols/cemetery_jewish.18.png) in 2012 and later [edited](https://github.com/gravitystorm/openstreetmap-carto/commit/2bc90019e4fa4f80df72c284df2b5fc43485f671) into [`grave_yard_jewish.png`](https://github.com/gravitystorm/openstreetmap-carto/blob/2bc90019e4fa4f80df72c284df2b5fc43485f671/symbols/grave_yard_jewish.png) by removing the background in 2014.

Generic jewish symbol was recreated in SVG by @nebulon42 as [`jewish.svg`](https://github.com/gravitystorm/openstreetmap-carto/blob/master/symbols/jewish.svg) and while it looks similar to [`jewish3.p.16.png`](https://github.com/gravitystorm/openstreetmap-carto/blob/739298d645babdb9da5aee2ef6434ecac00035cb/symbols/jewish3.p.16.png) on the GitHub, its rendering done by Mapnik looks quite different (it has less details and is bolder). It's the same problem as you see with my try to use this vector shape for cemeteries.

So the problem exists probably for both jewish symbols files (generic and cemetery), but we just haven't noticed the change for a generic jewish symbol before. We have no control over Mapnik rasterization process and it doesn't look like the problem with our SVGs (see also https://github.com/gravitystorm/openstreetmap-carto/pull/2727#issuecomment-320687262).

I think the most important thing is that cemetery type should be clear, not recreating the same image.

pnorman:
> I think the most important thing is that cemetery type should be clear, not recreating the same image.

I find the previous images clearer. Given that the previous images work better, I'd rather stick with them.

kocio-pl:
OK, I can leave jewish symbol for cemeteries as PNG (and just make it fit 32 px matrix). With other symbols (generic, muslim and christian) I see no such problem, so I will recreate them in SVG.

kocio-pl:
32 px matrix for all, jewish cemetery symbol left as PNG.

r 14q1aw
j1ob cex

Vort:
Why default symbol is not symmetric anymore?
tomb

kocio-pl:
I don't know, SVG is symmetric:
zrzut ekranu z 2017-08-12 15-35-38

Vort:
tomb2

PR #2715, opened 2017-07-31
Switching jewish cemeteries to SVG

kocio-pl:

Warsaw, z19
Before
d_qk7zl4
After
xq1vlchd


symbols/grave_yard_jewish.svg
```diff
@@ -0,0 +1,38 @@
...
+<svg
+   xmlns:dc="http://purl.org/dc/elements/1.1/"
+   xmlns:cc="http://creativecommons.org/ns#"
+   xmlns:rdf="http://www.w3.org/1999/02/22-rdf-syntax-ns#"
+   xmlns:svg="http://www.w3.org/2000/svg"
+   xmlns="http://www.w3.org/2000/svg"
+   version="1.1"
+   width="18"
+   height="18"
+   viewBox="0 0 18 18"
```
pnorman: needs to be a multiple of 2, for the same reasons as #2716

symbols/grave_yard_jewish.svg
```diff
@@ -0,0 +1,38 @@
...
+     id="defs8" />
+  <rect
+     width="18"
+     height="18"
+     x="4"
+     y="5"
+     id="canvas"
+     style="visibility:hidden;fill:none;stroke:none" />
+  <path
+     style="fill:#88b78e;fill-opacity:1;stroke-width:0.73192507"
```
pnorman: It's best to not have stroke

kocio-pl:
I guess stroke was not used anyway, but I removed it. Current PNG version is 18 px, but I've changed it to 16 px and now the pattern is a bit more dense:
Before
o qhp3m6
After
mnylwqrf

pnorman:
The icons look bolder now - is this accurate, or just perceptual?

kocio-pl:
The color is the same and the size is rather similar, but current images are more blurred than new ones.

kocio-pl:

PR #2716, opened 2017-07-31
Switching generic cemeteries to SVG

kocio-pl:

Warsaw, z19
Before
swbltrbh
After
fistvh5l


pnorman:
The pattern size needs to be some integer division of 256.

btw, I did a cemetery pattern at https://github.com/kartotherian/brighmed/blob/master/symbols/patterns/cemetery.svg

symbols/grave_yard_generic.svg
```diff
@@ -0,0 +1,38 @@
...
+<svg
+   xmlns:dc="http://purl.org/dc/elements/1.1/"
+   xmlns:cc="http://creativecommons.org/ns#"
+   xmlns:rdf="http://www.w3.org/1999/02/22-rdf-syntax-ns#"
+   xmlns:svg="http://www.w3.org/2000/svg"
+   xmlns="http://www.w3.org/2000/svg"
+   id="svg4"
+   viewBox="0 0 18 18"
+   height="18"
+   width="18"
```
pnorman: To tile correctly, the tile size needs to be an integer multiple of the pattern size, so this needs to be a power of two

kocio-pl:
Thanks! Brighmed pattern does not directly fit here:
hnkmczff
But reworked 16px version should be OK:
t0p3zzkx

BTW: Interesting enough, [current PNG version](https://github.com/gravitystorm/openstreetmap-carto/blob/master/symbols/grave_yard_generic.png) is 16x16.

kocio-pl:

PR #2718, opened 2017-07-31
Switching christian cemeteries to SVG

kocio-pl:

Warsaw, z19
Before
pgcebukv
After
3jmevvnq


kocio-pl:

PR #2721, opened 2017-08-01
Bullet-proof Docker variable turning with PostGIS.

### LINKED ISSUE #2698: System tuning for Docker-based setup
We should allow easy system tuning for Docker-based setup, since big database extract could make tester's life harder. When I disabled PostgreSQL tuning on my regular system, import of low zoom Europe data took 125 min instead of 57 min, which is a substantial difference.

On https://switch2osm.org/loading-osm-data/ there are two sections to consider:
- Getting ready to load
- Tuning

First one is about overcommit settings (I have it enabled and haven't tested how disabling it affects the performance) and the second one is about PostgreSQL memory settings. According to that page setting `maintenance_work_mem` and `work_mem` are "probably enough on a development or testing machine". 

`docker-compose.yml` file can have a block like this (using conservative settings for 2 GB of RAM as default):

```
  db:
    image: mdillon/postgis:9.6
    command: postgres -c work_mem=$PG_WORK_MEM -c maintenance_work_mem=$PG_MAINTENANCE_WORK_MEM
    environment:
      - PG_WORK_MEM=16MB
      - PG_MAINTENANCE_WORK_MEM=128MB
```
but it just doesn't work:
```
$ docker-compose up db
WARNING: The PG_WORK_MEM variable is not set. Defaulting to a blank string.
WARNING: The PG_MAINTENANCE_WORK_MEM variable is not set. Defaulting to a blank string.
Starting openstreetmapcarto_db_1
Attaching to openstreetmapcarto_db_1
db_1       | FATAL:  invalid value for parameter "work_mem": ""
openstreetmapcarto_db_1 exited with code 1
```
Even if it works, we would have another place to change settings instead of `.env` file, where we store osm2pgsql settings - and `docker-compose.yml` is watched by git. But we create  `.env` when starting `import` container, which is executed after `db` container is running, so PostgreSQL would still fail.

What could we do with it?

littlebtc:
Should fix #2698 in a very good way.

This will:
- Giving user a good out-of-a-box PostgreSQL / osm2pgsql turing.
- ... but still allow user to supply args during their first import;
- ...then generating `.env` based on args above.

Also fix the document about the behavior.

There are some steps to see how it just works, you can try:

1. For the first time, running `docker-compose up import` without any custom args
2. Check `.env`, change some value, then destroy all volumes using `docker-compose down -v && docker-compose rm -v` (PostgreSQL turnings will only applied once the volume is created)
3. Run `docker-compose up import` again to see whether settings from `.env` applied
4. Destroy volumes again
5. Prepend args different from those in `.env` in import command, like stated in `DOCKER.md`, import again
6. Settings shall use prepended args but `.env` shall not be changed.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,2 +1,3 @@
 *
+!scripts/tune-postgis.sh
 !openstreetmap-carto.style
--- a/DOCKER.md
+++ b/DOCKER.md
@@ -30,7 +30,20 @@ Once you have that file put it into the openstreetmap-carto directory and run `d
 This starts the PostgreSQL container (downloads it if it not exists) and starts a container that runs [osm2pgsql](https://github.com/openstreetmap/osm2pgsql) to import the data. The container is built the first time you run that command if it not exists.
 At startup of the container the script `scripts/docker-startup.sh` is invoked which prepares the database and itself starts osm2pgsql for importing the data.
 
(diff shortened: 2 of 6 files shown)


math1985:
Thanks!

kocio-pl:
I was thinking about something else, but this is probably good enough solution for now, so I may revisit this problem in the future.

PR #2722, opened 2017-08-02
Rendering farmland earlier on midzoom

kocio-pl:
Related to https://github.com/gravitystorm/openstreetmap-carto/pull/2654.

Start rendering farmlands on z8+ instead of z10+. I don't want to touch lowzoom yet (https://github.com/gravitystorm/openstreetmap-carto/issues/2688), but when we start showing forests/woods on midzoom, we can also show the farmlands, because both are very popular and with color fading they enhance the background in a subtle way.

Full Poland extract (click to see unscaled images):
z8
poland-z8-farmland
z9 (converted to JPG to fit in the GitHub filesize limit)
poland-z9-farmland

### DIFF
--- a/landcover.mss
+++ b/landcover.mss
@@ -313,7 +313,7 @@
 
   [feature = 'landuse_farmland'],
   [feature = 'landuse_greenhouse_horticulture'] {
-    [zoom >= 10] {
+    [zoom >= 8] {
       polygon-fill: @farmland;
       [zoom >= 16] {
         line-width: .5;
--- a/project.mml
+++ b/project.mml
@@ -101,12 +101,12 @@ Layer:
             COALESCE(wetland, landuse, "natural") AS feature
           FROM (SELECT
               way, COALESCE(name, '') AS name, religion,
-              ('landuse_' || (CASE WHEN landuse IN ('forest', 'military') THEN landuse ELSE NULL END)) AS landuse,
+              ('landuse_' || (CASE WHEN landuse IN ('forest', 'military', 'farmland') THEN landuse ELSE NULL END)) AS landuse,
               ('natural_' || (CASE WHEN "natural" IN ('wood', 'sand', 'scree', 'shingle', 'bare_rock') THEN "natural" ELSE NULL END)) AS "natural",
               ('wetland_' || (CASE WHEN "natural" IN ('wetland', 'mud') THEN (CASE WHEN "natural" IN ('mud') THEN "natural" ELSE tags->'wetland' END) ELSE NULL END)) AS wetland,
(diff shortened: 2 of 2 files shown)


math1985:
:+1:
What does this do to render times? Even a rough estimation is fine.

kocio-pl:
I'm in a middle of system upgrade - HDD is just bad choice when dealing with country/continent/planet database, so I will try to measure it when new SSD take over.

kocio-pl:
@rrzefox Could you update your map layer with this patch? And maybe even tell us what is the performance hit?

kocio-pl:
For OSM servers low+mid zoom levels (z0-z12) would be no big deal anyway, since they recreate only when deploying new release.

math1985:
And whenever we are developing on these zoom levels, of course.

math1985:
Also the re-creation process already takes multiple hours, that amount shouldn't multiply.

kocio-pl:
Single Poland exporting times (compared to master), database on SSD:
- z8: 2:32 (2:45)
- z9: 3:34 (3:22)

Looks like they are roughly the same.

kocio-pl:
Romania (click to see unscaled images):
z8
1r9b3bbw
z9
wpu75_qn

See also https://github.com/gravitystorm/openstreetmap-carto/issues/2688#issuecomment-317525516 for low zoom rendering with farmlands.

rrzefox:
> @rrzefox Could you update your map layer with this patch? And maybe even tell us what is the performance hit?

Even though this has already been merged, I've now added that patch on my server.
I checked with a few rerendered tiles, and the rendering time more or less stays the same, there is no mayor performance hit.

imagico:
Independent of the question how this fits into the design of these zoom levels - which as indicated i don't get in general so i have not commented - it should be clear that showing wood and farmland earlier than all other landcovers makes the style significantly more euro-centric than it already is.  Showing wood/forest earlier than all other landcovers is questionable but defensible considering how distinct and globally widespread woodland is and that it often continuously covers larger areas.  But i don't see how showing both woodland and farmland early and the rest later can be explained by anything other than an euro-centric perspective (Europe has by far the highest dominance of woodland and farmland in combination relative to other landcovers in comparison to any other larger region on Earth).

kocio-pl:
I'm happy to add more diverse landuses if possible, so if you have some propositions, I will investigate it. At this moment I think there are simply not many of them to be visible on mid zoom levels (that's my reason), but for example natural=sand is what I'm testing now.

imagico:
Since i don't understand the whole design idea at these zoom levels i have no specific recommendation.

PR #2725, opened 2017-08-03
High-contrast fonts as alternative fonts

### LINKED ISSUE #2692: Italic vs high-contrast fonts
screenshot 1

The screenshot uses Noto Serif Regular instead of Noto Sans Italic.

Currently, for `oblique-fonts`, we use Noto Sans Italic, which covers the latin, greek and cyrillic scripts – following https://en.wikipedia.org/wiki/List_of_writing_systems#List_of_writing_scripts_by_adoption that’s about 58% of the population. For the other scripts, we fall back to the regular style; so 42% do not see any difference between `oblique-fonts` and `book-fonts`.

It is unlikely that we will ever have italic/oblique fonts for the other scripts, because italic/oblique design is a particularity of latin/greek/cyrillic typeface design and mostly unknown in other cultural contexts.

On the other hand, the difference between sans/low-contrast on one hand, and serif/high-contrast on the other hand is more widespreaded. (Contrast means the contrast in stroke thickness within the glyphs: In low-contrast fonts all parts of the strokes have almost the same thinkness, and in high-contrast fonts some parts show a much thicker stroke than other parts.) With the currently released Noto Serif fonts, we would cover about 85% of the population. And in the next Noto release, there will be further more Serif fonts, the the coverage would raise again.

The biggest “coverage gain” we would have with the CJK scripts. But because of the complex glyph forms and the often small font size in openstreetmap-carto, a high-contrast CJK glyph is here also problematic for legibility. Note also that, as far as I know, Noto Serif CJK (though released some months ago) is still not packaged for Ubuntu, which might mean that it won’t be available on the osm.org servers soon.

Noto Italic vs Regular means that the Italic font is designed to be harmonic with the Regular style, but at the same time have an easy-to-spot difference. Noto Serif vs Sans means that the Serif font is also designed to be harmonic with the Sans style, but it’s _not_ designed to have an easy-to-spot difference; of course it looks different, but it’s not so obvious.

In general, serif/high-contrast styles are more difficult to render and read at small font sizes that sans/low-contrast fonts.

Advantages of Serif:
- Get away from cultural bias to western typography
- Extends greatly the coverage of the `@oblique-font` (as it’s currently called) style (from 58% to at least 85%)

Disadvantages of Serif:
- More dificult to spot the difference with the regular style for western scripts (about 58% affected here)
- Maybe more difficult to read at small font sizes

Comments?

sommerluk:
Resolves #2692

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -63,15 +63,21 @@ Unifont is used as a last resort fallback, with it's excellent coverage, common 
 
 ### Installation on Ubuntu/Debian
 
-On Ubuntu 16.04 or Debian Testing you can download and install the required fonts except Noto Emoji Regular with
+On Ubuntu 16.04 or Debian Testing you can download and install most of the required fonts with
 
 ```
 sudo apt-get install fonts-noto-cjk fonts-noto-hinted fonts-noto-unhinted fonts-hanazono ttf-unifont
 ```
 
-Noto Emoji Regular can be downloaded [from the Noto Emoji repository](https://github.com/googlei18n/noto-emoji).
+It might be useful to have a more recent version of the fonts for [rare non-latin scripts](#non-latin-scripts).
 
-It might be useful to have a more recent version of the fonts for [rare non-latin scripts](#non-latin-scripts). This can be installed [from source](https://github.com/googlei18n/noto-fonts/blob/master/FAQ.md#where-are-the-fonts).
+These fonts are not available on Ubuntu 16.04:
+* Noto Emoji Regular
+* Noto Serif CJK JP Medium
+* Noto Serif Devanagari Regular
+* Noto Serif Hebrew Regular
+
+The up-to-date comprehensive Noto fonts can be installed [from source](https://github.com/googlei18n/noto-fonts/blob/master/FAQ.md#where-are-the-fonts).
 
 DejaVu is packaged as `fonts-dejavu-core`.
 
--- a/amenity-points.mss
+++ b/amenity-points.mss
@@ -16,7 +16,7 @@
 @landcover-wrap-width-size: 25;
 @landcover-wrap-width-size-big: 35;
 @landcover-wrap-width-size-bigger: 45;
-@landcover-face-name: @oblique-fonts;
+@landcover-face-name: @alternative-fonts;
 
 @standard-wrap-width: 30;
 @standard-text-size: 10;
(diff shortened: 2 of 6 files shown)


kocio-pl:
That's why I'd like to hear someone using non-latin scripts to comment. It could make western world look worse, but help some other parts of the world look better, which would be good for universal map diversity. So I want to know if/how it really works for the rest of the world.

kocio-pl:
I'd like to know what the people who commented https://github.com/gravitystorm/openstreetmap-carto/pull/2496 think about this change. @saeedghazijoolaee, @jojo4u, @lyxys, @BoFFire, what's your opinion?

sommerluk:
As #2496 was about the arabic script, it might be worth to say that _this_ PR does _not_ change the rendering of the arabic script.

(Our current arabic font, “Noto Naskh Arabic”, is indeed a high-contrast font yet. As there are no alternatives, we use it for both, `@regular` and `@oblique`. In the upcoming Noto Phase III, there will be additionally an Arabic Sans in many weights and widths which will have low contrast. So _after_ the release of Noto Phase III, we could treat the currently used Naskh font as `@oblique` and in contrast use the new Arabic Sans as `@regular`…)

kocio-pl:
Sorry for misunderstanding. This font topic is beyond my expertise (and pretty complicated, since I've read it few times and still don't follow), but it's not about me - I'd like to know what real people using different scripts are thinking. 

So, what language/country communities could be a target we should ask? We can probably contact with the most active/biggest ones on the forum and generally on the Talk list.

sommerluk:
Language communities following Wikipedia:

- For CJK script: chinese (in all its dialects), japanese and korean.
- For Devanagari script: Hindi, Marathi, Konkani, Nepali, Sanskrit
- For  Bengali script: Sanskrit, Bengali, Assamese, Kokborok, Bishnupriya Manipuri, Khasi, Meitei Manipuri, Hajong and Sylheti
- For Malayalam script: Malayalam

I think anyway this is not a complet change of font. We stay within the same font superfamily and change only the font face (for a limited scope of labels).

kocio-pl:
Thanks! Could you also put some before/after renderings for all these scripts, so people see what is changing?

sommerluk:
Left before. Right after.

Latin script:
screenshot 1

CJK script:
screenshot 2

CJK script:
screenshot 3

Devanagari script:
screenshot 4

Bengali script:
screenshot 5

The water color difference is independent from this PR.

BTW: On older maps, usage of serif fonts is quite common.

kocio-pl:
I also don't like serif fonts in western scripts, but could accept it if other scripts would gain something this way. However on these screenshots I see no change, so I'm against merging this code.

sommerluk:
> I see no change

It affects water features and landuse features. There _is_ a difference in the screenshots, but of course more subtile than for the Latin regular vs italic.

kocio-pl:
I'm closing it, because problems are visible, gains are subtle (if any) and there was no support from users of other scripts.

PR #2726, opened 2017-08-05
Junction names on areas

### LINKED ISSUE #1253: Render junction=yes on areas

sommerluk:
Resolves #1253

A previous try at #2510 slowed down the rendering. I think this was because #2510 tried to backport this to the old branch using more complex code. This PR is much simpler. But it might be worth nevertheless to make sure there are no performance issues before merging this.

### DIFF
--- a/project.mml
+++ b/project.mml
@@ -1789,9 +1789,21 @@ Layer:
             highway,
             junction,
             ref,
-            name
+            name,
+            NULL AS way_pixels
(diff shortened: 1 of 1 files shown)


pnorman:
This shouldn't cause performance issues, but I haven't loaded it up to review.

math1985:
Thanks!

PR #2730, opened 2017-08-07
Better line spacing

### LINKED ISSUE #2358: Reduce vertical spacing in labels
Noto has a lot larger vertical spacing than the old font. This looks rather ugly for cartography (at least for Latin script).

Before/after:

screenshot_2016-09-20_23-55-12 screenshot_2016-09-20_23-54-38

We can use a negative text-line-spacing attribute to reduce the vertical spacing by a couple of pixels.

sommerluk:
Resolves #2358 (last part is solved by this PR)

Before:
screenshot 1

After:
screenshot 2

### DIFF
--- a/amenity-points.mss
+++ b/amenity-points.mss
@@ -11,15 +11,19 @@
 @landform-color: #d08f55;
 
 @landcover-font-size: 10;
+@landcover-wrap-width-size: 25; // 2.5 em
+@landcover-line-spacing-size: -1.5; // -0.15 em
 @landcover-font-size-big: 12;
+@landcover-wrap-width-size-big: 36; // 3 em
+@landcover-line-spacing-size-big: -1.8; // -0.15 em
 @landcover-font-size-bigger: 15;
-@landcover-wrap-width-size: 25;
-@landcover-wrap-width-size-big: 35;
-@landcover-wrap-width-size-bigger: 45;
+@landcover-wrap-width-size-bigger: 45; // 3 em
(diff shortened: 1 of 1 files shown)


kocio-pl:
Could you also show examples how would other scripts look like with this change?

sommerluk:
screenshot 3
screenshot 4
screenshot 5
screenshot 6

Note that the used line spacing is simply the same that we have also applied in the rest of openstreetmap-carto (−0.15 em for line wrap ≤ 3 em).

This is the last one of various PR that implemented the fine-tuned line spacing.

math1985:
Thanks! By the way, it would also make sense to keep the wrap-width a constant em amount.

verdy-p:
Closed without testing with Myanmar ?

verdy-p:
And still no test for Arabic : (high Khufi font, suitable for real trafic signs where they are isolated and have cumfortable line-height) or UI font (better suited for multilingual text on a dense map) ?
I summary the tests above are only for alphabetic (Latin/Greek/Cyrillic) script, Indic script, CJK scripts, it will probably work for Hebrew and Tifinagh
But I still have doubts for Arabic, Myanmar, Khmer, Lao, and Tibetan, and possibly even with Thai.

kocio-pl:
Could you provide test rendering for missing scripts?

verdy-p:
All these scripts were in  the initial #2358
When closing this bug with limited tests, someone marked #2358 as resolved even if you tested much less things here.

PR #2732, opened 2017-08-08
Use z_order for roads instead of priority table

pnorman:
This is getting a start on #2046. There should be no rendering changes.

The `prio` join added extra filtering, but this isn't needed, and was in fact undesirable. The features to be rendered are specified in the MSS, and the data volume returned to Mapnik isn't noticeably impacted because the huge majority of lines matching the WHERE clause also match the features being rendered.


kocio-pl:
The code is shorter with this change, which is great, but are there any other interesting implications?

Should it be used also in [this code](https://github.com/gravitystorm/openstreetmap-carto/pull/2687/files#diff-2d6c6d06f63a6fab47286addedefc335R1975)?

pnorman:
> The code is shorter with this change, which is great, but are there any other interesting implications?

I'm not sure what you mean, but I'm not aware of any problems.

> Should it be used also in [this code](https://github.com/gravitystorm/openstreetmap-carto/pull/2687/files#diff-2d6c6d06f63a6fab47286addedefc335R1975)?

Yes.

kocio-pl:
> I'm not sure what you mean, but I'm not aware of any problems.

I wonder for example if z-order really gives the same result as priority? If I understand correctly, z-order means that we use layer=* tag, while prio does not use it. Am I right or it works in some other way?

pnorman:
> I wonder for example if z-order really gives the same result as priority? If I understand correctly, z-order means that we use layer=* tag, while prio does not use it. Am I right or it works in some other way?

`layernotnull` comes first in the sort, so that's used before either. [z_order](https://github.com/gravitystorm/openstreetmap-carto/blob/master/openstreetmap-carto.lua#L242-L250) does not depend on layer.

pnorman:
Just to note, if there are **any** changes in rendered output that someone spots, I'd be interested, because I'm not expecting any.

project.mml
```diff
@@ -444,7 +444,7 @@ Layer:
       table: |-
         (SELECT
             way,
-            (CASE WHEN substr(feature, length(feature)-3, 4) = 'link' THEN substr(feature, 0, length(feature)-4) ELSE feature END) AS feature,
```
math1985: Can we make this change in a separate PR for clarity?

pnorman: I've done it in a separate commit, which should be clear enough

math1985: Ok, fine.

PR #2738, opened 2017-08-12
Rendering natural=sand earlier on midzoom

kocio-pl:
Related to #2654, follow up to #2722.

Another landuse type that can be rendered on midzoom on z8+ (instead of z9+), just like the forest and farmland, because there are already places where it can be quite big (=visible even on the low zoom scale) and it helps diversity.

@rrzefox: Could you add this patch too and check the performance implications?

Examples  (click to see unscaled images):

Egypt, z8
wvvtbkza

Africa, z8 (converted to JPG to fit in GitHub filesize limit)
[-> link to big image](https://user-images.githubusercontent.com/5439713/29237544-a19b77fe-7f20-11e7-956c-9b9129ee4e42.jpg)

### DIFF
--- a/amenity-points.mss
+++ b/amenity-points.mss
@@ -1934,11 +1934,11 @@
     }
   }
 
-  [feature = 'natural_scree'],
-  [feature = 'natural_shingle'],
-  [feature = 'natural_bare_rock'],
+  [feature = 'natural_scree'][zoom >= 9],
+  [feature = 'natural_shingle'][zoom >= 9],
+  [feature = 'natural_bare_rock'][zoom >= 9],
   [feature = 'natural_sand'] {
-    [zoom >= 9][way_pixels > 3000][is_building = 'no'],
+    [zoom >= 8][way_pixels > 3000][is_building = 'no'],
     [zoom >= 17][is_building = 'no'] {
       text-name: "[name]";
       text-size: @landcover-font-size;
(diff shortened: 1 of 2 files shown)


sommerluk:
Could you remove the Africa image itself from this PR description and leave only the download link to https://user-images.githubusercontent.com/5439713/29237544-a19b77fe-7f20-11e7-956c-9b9129ee4e42.jpg

It’s so big, it crash’s my browser(s) and even my image preview… 😉

Anyway an intresing image. So not much of the Sahara sand is currently mapped in OSM…

kocio-pl:
Update - names should also render earlier, especially since these tend to be named, big areas:

Egypt, z8 (click to see unscaled images):
bqwlj6xx

rrzefox:
> @rrzefox: Could you add this patch too and check the performance implications?

added and rererendered Z 8 and 9; Your example area would be roundabout here: https://osm.rrze.fau.de/map-ll-osm?zoom=8&lat=28.919&lon=27.745&layer=RRZE%20tileserver%20osm.org%20style%20tiles
Looking at rendering times of randomly selected metatiles I see no mayor performance impact (but take this with a grain of salt: This is a server rendering usually multiple tiles in parallel, sometimes live, so the fluctuation from that will be so big that a performance impact of a few percent simply isn't measurable).

Issue #2670, opened 2017-07-02
Stop rendering generic landuse=military


imagico:
I recently looked a bit at use of `landuse=military` worldwide and i suggest we stop rendering generic areas with this tag, i.e. ones that have no supplemental `military=*` tag.

The reason is the tag is used for a large number of very different things many of which have very little to do with the main purpose of the tag (that is to indicate areas used primarily for military purposes - see [wiki](https://wiki.openstreetmap.org/wiki/Tag:landuse%3Dmilitary)).  In particular the tag is widely used to mark exclusion zones of any kind while many real military landuses are not access restricted at all.

There are also (esp. in the US) military areas inside military areas (like an airfield or barracks inside a military base) which makes no sense at all the way we render it.

We should continue to render `military=danger_area` - though probably in a different styling and we could render additional `military=*` values which are common and consistently used.  We could also render access restrictions on `landuse=military` or `boundary=exclusion_zone` although neither of these is currently used in significant  numbers.

dieterdreist:
sent from a phone

> On 2. Jul 2017, at 22:59, Christoph Hormann <notifications@github.com> wrote:
> 
> while many real military landuses are not access restricted at all.


really? Can you give some examples?

imagico:
In Germany many military training areas (Truppenübungsplatz) are at least partly open to the general public - you can see this often in the map due to the lack of access restrictions on roads.  AFAIK in the US there are also many military bases where public roads cross with no restrictions.

I don't want to get into tagging discussion here but it is essential to separate access restrictions from military use of an area - these are two different concepts which might frequently co-exist but one does not necessarily imply the other (in either direction).  The way we render this implying access restrictions does not help in that regard and use of tags shows that.

dieterdreist:
sent from a phone

> On 3. Jul 2017, at 10:22, Christoph Hormann <notifications@github.com> wrote:
> 
> In Germany many military training areas (Truppenübungsplatz) are at least partly open to the general public - you can see this often in the map due to the lack of access restrictions on roads. AFAIK in the US there are also many military bases where public roads cross with no restrictions.


I see what you are after: accessible roads crossing otherwise inaccessible areas, right? Because I don't think you could stop your car and go for a walk if you're on one of those roads that cross a Truppenübungsplatz. military landuse almost always implies access=no.

imagico:
No, but as i said i don't want to turn this into a tagging discussion.

[...]

Issue #684, opened 2014-07-01
military=danger_area looks like building on z10 and is agressive pink at z11


matkoniecz:
https://www.openstreetmap.org/way/234810192#map=10/54.4057/19.4012

Something between z10 and z11 style would be better - neither eyesore pink nor nearly invisible grey.

Also, pink is used for retail, commercial, supermarkets and tourism attraction. 

Current z11 style:
![przechwytywanie](https://cloud.githubusercontent.com/assets/899988/3439532/7dea123a-00e7-11e4-9ef3-8389ebecc184.PNG)

Current z10 style:
![przechwytywanie](https://cloud.githubusercontent.com/assets/899988/3439538/98d91ea6-00e7-11e4-9945-5407c8a25759.PNG)

Currently used style for [landuse=military; military=barracks] that IMHO would work as military=danger_area and for its current purpose is too strong:
![przechwytywanie](https://cloud.githubusercontent.com/assets/899988/3439558/2a5dc638-00e8-11e4-9eb3-04d764c9af73.PNG)

math1985:
Do you think we need to keep both military=danger_area and landuse=military? We could consider using the current landuse=military rendering (red stripes) for both. Most military training grounds are likely danger areas in one way or another, and military areas probably shouldn't be entered even if they are not danger areas.

matkoniecz:
> Do you think we need to keep both military=danger_area and landuse=military? We could consider using the current landuse=military rendering (red stripes) for both.

Seems to be a good idea - I will run test on it. But in that case it may be possible to do not handle this tag. 

Wiki has info that adding also landuse=military is mandatory what makes sense, but it is done in about half of military=danger_area. But even 50% of object is less than 500 object worldwide - see http://overpass-turbo.eu/s/gJi

dieterdreist:
sent from a phone

> Il giorno 10 giu 2016, alle ore 00:26, Mateusz Konieczny <notifications@github.com> ha scritto:
> 
> Do you think we need to keep both military=danger_area and landuse=military? We could consider using the current landuse=military rendering (red stripes) for both.
> 
> Seems to be a good idea - I will run test on it. But in that case it may be possible to do not handle this tag.
> 


what about military danger areas on water? A danger area for military reasons doesn't necessarily imply that the area is currently used for military purposes, or does it? Couldn't this also be used for legacy dangers?

matkoniecz:
[...]

PR #2739, opened 2017-08-12
Danger symbol rework in SVG

kocio-pl:
Trying to resolve one of the problems from https://github.com/gravitystorm/openstreetmap-carto/issues/937.

The easiest thing would be to add 2 pixels h/w to `danger.png`. More desirable would be to recreate it in SVG.

I made a background to make it look the same, however the question is if we want this background or was it just a leftover we should get rid of:

Example at z13:

Before
danger-png
After (with background)
danger-svg-bg
After (without background)
danger-svg-nobg


kocio-pl:
We can also change background opacity. For example with `polygon-opacity: 0.5;` it looks like:
wbvhgvps
Opacity 0.6:
bj7xvrpa
Opacity 0.7:
4ypbe8 k

math1985:
Perhaps we should decide on #684 and #2670 first?

PR #2741, opened 2017-08-12
Fix generic cemetery symbol geometry

kocio-pl:
Fix for https://github.com/gravitystorm/openstreetmap-carto/pull/2714#issuecomment-321981096.

Now it's symmetric rendering:

duj3ovtb


math1985:
Curious, what caused this problem in the first place?

kocio-pl:
I didn't turn on snapping points to the guiding lines by accident, so a small part of the shape was filling space on the right. It was too small too see on SVG, but rasterizer turned these pixels into light shade of green of course. See the image from @Vort:

tomb2

Issue #484, opened 2014-04-18
Add support for slipway ways 


finetjul:
Right now, only slipway points are visible. But >20% of those slipways are ways.
- http://wiki.openstreetmap.org/wiki/Tag%3Aleisure%3Dslipway
- http://taginfo.openstreetmap.org/tags/leisure=slipway#overview

How about:
- always render slipways points as icons
- for zoom >= ?, render slipway ways with blue lines

boothym:
There are a couple of slipways near me with the same issue, e.g. https://www.openstreetmap.org/edit#map=19/56.06853/-3.17367 - see http://c7.alamy.com/comp/G100NF/elevated-view-of-kinghorn-fife-scotland-april-2016-G100NF.jpg for a view of the area.

In this case I created it as a way because it's more information than just a node and it's not at the end of a road. Or should I be adding a node to the current service road or making the slipway a service road instead?

I don't want to be tagging for the renderer by just creating a node, so it would be good if this issue can be fixed for ways.

PR #2743, opened 2017-08-13
Moving ford to a new tagging scheme

kocio-pl:
Resolves https://github.com/gravitystorm/openstreetmap-carto/issues/267.

Simply removing support for deprecated tagging scheme and adding support for the new one (much more popular already).

I need some help, since it works for points, but not for ways. Another thing is deciding on rendering names - 3% have them, but I suspect most of them are the names of the highways.

### DIFF
--- a/project.mml
+++ b/project.mml
@@ -1453,6 +1453,23 @@ Layer:
         ) AS amenity_points_poly
     properties:
       minzoom: 10
+  - id: amenity-line
+    geometry: linestring
+    <<: *extents
+    Datasource:
+      <<: *osm2pgsql
+      table: |-
+        (SELECT
+          way,
+          COALESCE(
+           'highway_' || CASE WHEN tags @> 'ford=>yes' OR tags @> 'ford=>stepping_stones' THEN 'ford' ELSE NULL END
+            ) AS feature
+          FROM planet_osm_line
+          -- The upcoming where clause is needed for performance only, as the CASE statements would end up doing the equivalent filtering
+          WHERE tags @> 'ford=>yes' OR tags @> 'ford=>stepping_stones'
+        ) AS amenity_line
+    properties:
+      minzoom: 16
   - id: amenity-points
     class: points
     geometry: point
@@ -1483,7 +1500,8 @@ Layer:
               'historic_' || CASE WHEN historic IN ('memorial', 'monument', 'archaeological_site')
                              THEN concat_ws('_', historic, CASE WHEN tags->'memorial' IN ('plaque') THEN tags->'memorial' ELSE NULL END)
                              ELSE NULL END,
-              'highway_'|| CASE WHEN highway IN ('bus_stop', 'elevator', 'traffic_signals', 'ford') THEN highway ELSE NULL END,
+              'highway_'|| CASE WHEN highway IN ('bus_stop', 'elevator', 'traffic_signals') THEN highway 
+                                WHEN tags @> 'ford=>yes' OR tags @> 'ford=>stepping_stones' THEN 'ford' ELSE NULL END,
               'power_' || CASE WHEN power IN ('generator') THEN power ELSE NULL END,
               'tourism_' || CASE WHEN tourism IN ('viewpoint') THEN tourism ELSE NULL END,
               'man_made_' || CASE WHEN man_made IN ('cross') THEN man_made ELSE NULL END,
@@ -1535,7 +1553,8 @@ Layer:
             OR historic IN ('memorial', 'monument', 'archaeological_site', 'wayside_cross')
             OR tags->'memorial' IN ('plaque')
             OR tags @> 'emergency=>phone'
-            OR highway IN ('bus_stop', 'elevator', 'traffic_signals', 'ford')
+            OR highway IN ('bus_stop', 'elevator', 'traffic_signals')
+            OR tags @> 'ford=>yes' OR tags @> 'ford=>stepping_stones'
             OR (power = 'generator' AND (tags @> '"generator:source"=>wind' OR tags @> 'power_source=>wind'))
           ORDER BY score DESC NULLS LAST
           ) AS amenity_points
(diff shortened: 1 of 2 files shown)


math1985:
Related to #484.

I think we would need to add an amenity_line layer to accomplish this.

kocio-pl:
Looks like `amenity-line` layer works, since data inspector in Kosmtik shows feature `highway_ford` for a way tagged with `ford=*`, but it still doesn't show the icon. Maybe something should be changed in `amenity-points.mss`?

math1985:
Did you add a rendering rule for #amenity-line?

kocio-pl:
I haven't touched `amenity-points.mss` at all yet. There's a block:

https://github.com/gravitystorm/openstreetmap-carto/blob/89ef164b36f89d35b45aa238eea70f0d3650986d/amenity-points.mss#L141-L146

but I don't know how to add this rule.

kocio-pl:
OK, now it works - the icon is visible. The question is if this whole block have to be duplicated to work with points and lines or there's a way to combine it?

math1985:
I would add `class: points` to your new layer, so the `.class` selector in the mss will select it.

That might be enough, but I'm not sure if `marker-placement` works on lines (of if lines can have icons at all).

If it doesn't work you might need to do something like this:

[feature = 'highway_ford'][zoom >= 16] {
  .amenity-points {
    marker-file: url('symbols/ford.svg');
    marker-placement: interior;
    marker-clip: false;
  }
  .amenity-line {
    // Rules for line fords here
  }
}

math1985:
You can add `class: points` to your new layer, so the `.class` selector in the mss will select it.

If you need different rules for points and lines you can do something like this:

If it doesn't work you might need to do something like this:

[feature = 'highway_ford'][zoom >= 16] {
  marker-file: url('symbols/ford.svg');
  .amenity-points {
    // These are specific point rules
    marker-placement: interior;
    marker-clip: false;
  }
  .amenity-line {
    // Specific line rules here
  }
}

kocio-pl:
I want exactly the same rules for both.

math1985:
> I want exactly the same rules for both.

Then adding the points class should be enough.

However, you might find you still need different rules though, for example the 'interior' property doesn't really makes sense for lines.

kocio-pl:
Adding points class made me adding a lot of columns to amenity-line layer, so it makes more sense to just define additional rules for lines. Thanks for help!

kocio-pl:
Any comments or reviews? From my point of view it's ready.

project.mml
```diff
@@ -1445,6 +1445,21 @@ Layer:
...
+    <<: *extents
+    Datasource:
+      <<: *osm2pgsql
+      table: |-
+        (SELECT
+          way,
+          COALESCE(
+           'highway_' || CASE WHEN tags->'ford' IN ('yes', 'stepping_stones') THEN 'ford' ELSE NULL END
+            ) AS feature
+          FROM planet_osm_line
```
pnorman: This needs a `WHERE` clause.

project.mml
```diff
@@ -1525,7 +1541,8 @@ Layer:
             OR "natural" IN ('peak', 'volcano', 'saddle', 'spring', 'cave_entrance')
             OR historic IN ('memorial', 'monument', 'archaeological_site', 'wayside_cross')
             OR tags->'memorial' IN ('plaque')
-            OR highway IN ('bus_stop', 'elevator', 'traffic_signals', 'ford')
+            OR highway IN ('bus_stop', 'elevator', 'traffic_signals')
+            OR tags->'ford' IN ('yes', 'stepping_stones')
```
pnorman: `OR tags @> 'ford=>yes' OR tags @> 'ford=>stepping_stones'`

project.mml
```diff
@@ -1474,7 +1489,8 @@ Layer:
               'historic_' || CASE WHEN historic IN ('memorial', 'monument', 'archaeological_site') 
                              THEN concat_ws('_', historic, CASE WHEN tags->'memorial' IN ('plaque') THEN tags->'memorial' ELSE NULL END) 
                              ELSE NULL END,
-              'highway_'|| CASE WHEN highway IN ('bus_stop', 'elevator', 'traffic_signals', 'ford') THEN highway ELSE NULL END,
+              'highway_'|| CASE WHEN highway IN ('bus_stop', 'elevator', 'traffic_signals') THEN highway 
+                                WHEN tags->'ford' IN ('yes', 'stepping_stones') THEN 'ford' ELSE NULL END,
```
pnorman: `WHEN tags @> 'ford=>yes' OR tags @> 'ford=>stepping_stones' THEN ...`

pnorman:
This will be adding another layer which has to retrieve all the objects for an area from disk, even if it doesn't send them. This is best avoided, but makes sense here. Any partial indexes would have to involve the `tags` column, which I'd like to avoid, and they would be a bit unnatural to construct.

project.mml
kocio-pl: Done.

project.mml
```diff
@@ -1453,6 +1453,23 @@ Layer:
...
+  - id: amenity-line
+    geometry: linestring
+    <<: *extents
+    Datasource:
+      <<: *osm2pgsql
+      table: |-
+        (SELECT
+          way,
+          COALESCE(
+           'highway_' || CASE WHEN tags->'ford' IN ('yes', 'stepping_stones') THEN 'ford' ELSE NULL END
```
kocio-pl: What about notation in this line - maybe it should be:

```SQL
'highway_' || WHEN tags @> 'ford=>yes' OR tags @> 'ford=>stepping_stones' THEN 'ford' ELSE NULL END
```

It's all too database related, so I'd like to have some kind of a rule of thumb when to use this notation and when it should be tags->'xxx'.

pnorman: Within the output expressions it's purely a style issue. Our guidelines currently call for `@>`. The style disadvantage to `@>` is that you have to repeat `tags @>` for each option instead of something like && for arrays

kocio-pl: I understand that this is not an error, because it was working also with the first notation and I tested that it still works. I can follow this style just to have clean and consistent code, I just don't know when should we use it and when not.

So what about this line?

pnorman: `@>` to be consistent with our style guide.

kocio-pl: Done.

pnorman:
Unrelated to these changes, the ford symbol looks a little blurry

image

kocio-pl:
Strange - it's a vector image after all...

pnorman:
> Strange - it's a vector image after all...

That's got nothing to do with being blurry, it's probably not pixel aligned.

PR #2744, opened 2017-08-13
Slipway SVG icon

sommerluk:
Slipway SVG icon

Before:
screenshot 1

After:
screenshot 2


kocio-pl:
I guess you should join the paths (Path > Union) and save it as a plain SVG (instead of "Inkscape SVG"), but it looks nice - thanks!

PR #2746, opened 2017-08-14
Adding pattern for natural=sand

kocio-pl:
Irregular wave pattern for natural=sand would make it visibly different from farmland (and also from beach). Initial pattern shape and color are just proof of concept.

Egypt
z10
kodjovdw
z13
wp6p8xfm

### DIFF
--- a/landcover.mss
+++ b/landcover.mss
@@ -620,6 +620,10 @@
 }
 
 #landcover-area-symbols {
+  [natural = 'sand'][zoom >= 8] {
+    polygon-pattern-file: url('symbols/beach.png');
+    polygon-pattern-alignment: global;
+  }
   [int_wetland != null][zoom >= 10] {
     polygon-pattern-file: url('symbols/wetland.png');
     polygon-pattern-alignment: global;
--- a/project.mml
+++ b/project.mml
@@ -261,13 +261,13 @@ Layer:
               END AS int_wetland,
             tags->'leaf_type' AS leaf_type
           FROM planet_osm_polygon
-          WHERE ("natural" IN ('marsh', 'mud', 'wetland', 'wood', 'beach', 'shoal', 'reef', 'scrub') OR landuse = 'forest')
+          WHERE ("natural" IN ('marsh', 'mud', 'wetland', 'wood', 'beach', 'shoal', 'reef', 'scrub', 'sand') OR landuse = 'forest')
             AND building IS NULL
             AND way_area > 0.01*!pixel_width!::real*!pixel_height!::real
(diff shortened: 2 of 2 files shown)


math1985:
To me the wave patten symbolizes water, I wouldn't be in favour.

kocio-pl:
Any other ideas how could sand waves/dunes look like?

I have more general ideas to tinker with:
1. Maybe natural sand should use a random pattern we use for beach now and beach could use regular one instead.
2. Maybe both should look the same (including pattern or not).
3. Maybe farmland should have some (regular) pattern.

There's not too many places where both landuses are next to each other, but they do exist:
- http://www.openstreetmap.org/way/25872956#map=13/26.0187/12.9679
- http://www.openstreetmap.org/#map=19/50.36077/19.51951

kocio-pl:
What about this pattern:

https://github.com/osmandapp/OsmAnd-resources/blob/master/icons/svg/shaders/h_sand_shader.svg

kocio-pl:
It would look like this (OsmAnd has more restrictive licence, so we could ask for relicencing or just create something similar):

x7nch5pb

dieterdreist:
sent from a phone

> On 4. Sep 2017, at 03:44, kocio-pl <notifications@github.com> wrote:
> 
> What about this pattern:
> 
> https://github.com/osmandapp/OsmAnd-resources/blob/master/icons/svg/shaders/h_sand_shader.svg
> 


it's not completely bad, but a bit "heavy"/impactful. Could it become lighter somehow?

kocio-pl:
I guess the easiest and the most coherent way would be to just use the same pattern as sandy beach. General beach and farmland would be plain, sandy beach and natural=and would be covered with sand pattern then, so we have the system and we can recognize all of them. The only problem I see is if current sand pattern is not too light in general - it's hard to spot the dots for me:

natural=sand at z19:
Before
nhjirmpv
After
xoij65uq


natural=sand and landuse=farmland at z19:
Before
lfq3k0jn
After
gndxp_yj

kocio-pl:
Any comments or reviews? I think sand pattern may be improved later (like bigger or less grains), because it would also change sandy beach looks like.

Tomasz-W:
I like new rendering of sand areas, but why this change hasn't included beach areas too? Is it intentional?

kocio-pl:
Yes, default beach has no overlay pattern, because the pattern depends on beach/shoal surface:
https://github.com/gravitystorm/openstreetmap-carto/blob/72f6c8d835bc38c80756b6baa6d94da04cb9ae11/landcover.mss#L668-L684

PR #2747, opened 2017-08-14
leaf type rendering with slightly bigger symbols and switch forest patterns to 256 px

### LINKED ISSUE #822: Distinguish deciduous and coniferous forests

sommerluk:
This is basically the same as #2708. It resolves #822. Closes #2708.

The symbols for known leaf types are slightly bigger (2 px) than in #2708. That’s to make them more attractive (instead of more light) then the unknown-leaf-type-symbol).

The symbols are not simply scaled versions of the previous symbols, but are redesigned and pixel-aligned.


sommerluk:
screenshot 1

sommerluk:
Same pattern for all leaf types, only the symbols change.

sommerluk:
Should easily allow later for both, leaf cycle rendering by filling or half-filling the symbol, and also for leaf cycle rendering by changing the color of the symbol. But that’s not part of this PR…

kocio-pl:
I like this solution. Could you also post some real live rendering?

sommerluk:
screenshot 2
screenshot 3
screenshot 4
screenshot 5
screenshot 6
screenshot 7

kocio-pl:
These are only unknown/needleleaved, but I'd like also to see mixed, broadleaved and leafless areas. Just one example per type would be enough.

BTW - it would be interesting for me to see preliminary leaf cycle rendering with these symbols.

sommerluk:
screenshot 1
screenshot 2
screenshot 3

sommerluk:
> BTW - it would be interesting for me to see preliminary leaf cycle rendering with these symbols.

I would keep leaf cycle rendering out of this PR, also because personally I do not plan to work on that. Anyway, there were basically two proposals for leaf cycle rendering: Either by changing the color of the symbol or by filling it. I think both options are compatible with this PR.

kocio-pl:
Could leafless be weaker, to not stand out from the rest? With bigger size it could be probably possible without being hard to recognize.

sommerluk:
screenshot 1

sommerluk:
To compare: On the image:
- right below: broadleaved
- left below: unknown
- top: leafless

sommerluk:
Switching to 256 px:

screenshot 1

kocio-pl:
Both `leafless1.svg` and `leafless2.svg` are probably leftovers which should be deleted.

sommerluk:
> Both leafless1.svg and leafless2.svg are probably leftovers which should be deleted.

That’s intentional. I wanted to provide all symbols that are not part of the standard jsdotpattern repository so that everybody has available all the necessary files to regenerate them. Also for broadleaved and needleleaved, the original symbol files are provided within this PR.

kocio-pl:
Sorry, I missed that they are documented and reside in `generating_patterns`!

nebulon42:
Good job in rendering `leaf_type`. Much needed. :)

Issue #2697, opened 2017-07-22
Render landuse=religious


polarbearing:
Although being discussed in various places (#771, #1206, #2087, etc) it appears to me that there is no standalone issue for rendering ```landuse=religious```.

We know it is seen controversial by some participants here, however since #771 (1.6k uses in 2014) it has grown to 11k uses now. 

My proposal is still a gray tone, slightly darker than residential.

kocio-pl:
Could you be more specific with this color? It makes further discussion easier, since I can make test rendering.

pnorman:
What other colours would this be near? Power stuff comes to mind

PR #2751, opened 2017-08-15
Adding rendering for landuse=religious

kocio-pl:
Resolves https://github.com/gravitystorm/openstreetmap-carto/issues/2697.

This is the same color and the outline as we use for amenity=place_of_worship, because in all cases I've checked this is not a special open space for worship, but looks like (a) a building like church without `building=*` tag (which needs fixing) or (b) an area surrounding such building (the same as landuse=religious, which is meant to eventually supersede it). 

I have not added symbols rendering to allow visualization of situation like [on this picture](https://wiki.openstreetmap.org/wiki/File:Amenity_pow_usage_example.svg), where there is really a special area inside such landuse, which has its own symbol, but it also works with not eclipsing buildings like churches.

hpjd2oiu


kocio-pl:
At some point we can also render [landuse=churchyard](https://wiki.openstreetmap.org/wiki/Tag%3Alanduse%3Dchurchyard) in the same way, as it's just religion-specific version of landuse=religious, and it's gaining popularity too:

taghistory 5

kocio-pl:
Any comments or reviews for this ticket?

Tomasz-W:
I think that outline is too strong and too simillar to prison areas.

polarbearing:
Thanks @kocio-pl. I like the colour (not similar to prison since here is no hatching), but agree that the outline is too strong, resembling a barrier=fence.

I am against rendering landuse=churchyard. Tagging amenity=place_of_worship is a widely accepted approach to keep the high-level tag neutral regarding the religion. 'churchyard' violates this principle, so if you start rendering this, you would need to render other specific religious ground names for fairness, leading to more tag fragmentation. 

General style for campuses has been discussed here before, see my #1624 for plenty of links, however religious was different enough to prevent agreement so far.

kocio-pl:
It was meant as a simple extension for what people already do (tagging area as amenity=place_of_worship), so I wasn't expecting big arguments here.

General solution for areas is needed anyway, as we still don't show many of them. This problem is hard and we haven't come with such solution yet, but we can take the easier path and start with partial solution - I think generic name rendering would be good for a start. However I feel this is still something non-trivial and is outside the scope of this code.

kocio-pl:
The outline should be OK - in the original example there's really a fence, but this one doesn't have it: 
2xjvkozp

dieterdreist:
sent from a phone

> On 16. Aug 2017, at 05:29, kocio-pl <notifications@github.com> wrote:
> 
> At some point we can also render landuse=churchyard in the same way, as it's just religion-specific version of landuse=religious, and it's gaining popularity too


hooray, yet another new landuse tag. How does this relate to amenity=grave_yard?

kocio-pl:
It's probably more general and may contain graveyard. It's just a religion-specific version of landuse=religious. But this is not what this code is about, I've just mentioned something related.

wmyrda:
@math1985 This color You are against is already used on the map. This change does not introduce new color. If You do not like this color please say which color You would propose for this change. Churches, cathedrals and alike are always in the middle of the city and area around them is clearly not regular residential one therefore the use of landuse=religious and few similar tags religious being used in most of those instances. In my opinion there if we do not want areas as http://www.openstreetmap.org/#map=17/49.77944/22.77079 look incomplete we have to render it and only problem is to choose proper color. I personally like the of using color proposed in this PR.

polarbearing:
Ok, so should we go for the yellow currently used for school/hospital for all these campuses: religious, community_centre, social_facility, etc see #1624 ?
Alternatively, should we go for more border colours such as in theme_park?

kocio-pl:
I start to understand the difference. I see it as a practical problem - people are already tagging such areas and what I propose is to use basically the same rendering for alternative, but similar and probably more precise tagging. You see it as adding color to another landuse and think of the complexity behind it, which we were trying to solve in another issue some time ago, but didn't go too far with it.

Is it possible for you to accept this practical solution and leave general "campus" discussion for #1624 or you think that only general solution is acceptable?

math1985:
> Ok, so should we go for the yellow currently used for school/hospital for all these campuses: religious, community_centre, social_facility, etc see #1624 ?
> Alternatively, should we go for more border colours such as in theme_park?

It would best to try out some options, I think. I'm fine with replacing the yellow with some other rendering.

> I see it as a practical problem - people are already tagging such areas and what I propose is to use basically the same rendering for alternative, but similar and probably more precise tagging. 

Sorry, I'm not sure what you mean here?

kocio-pl:
This is example of what we already have - area with amenity=place_of_worship and without building=*:

http://www.openstreetmap.org/way/123388654

This is what we would have with the area tagged as landuse=religious (which is more precise with "landuse" rather than "amenity") - the same color and the same usage. There are many areas already tagged with landuse.

Is this clear now?

polarbearing:
Clear for me, and we should apply this soon.
General discussion about more and other campus situations can then continue separately.

kocio-pl:
Since both Matthijs and Paul share this comment, I will try to reply in detail.

> I don't think it makes sense to add a different type of area rendering for every type of 'campus' (hospital grounds, hotel grounds, and now church/religion grounds, and probably many more request will come if we go this way). There will simply not be enough colours. 

I agree, but in this case this is essentially the same type of area, just tagged more clearly (landuse is only for area, amenity can be attached to anything).

> Moreover, people will not be able to instantaneously guess what this color means, and this feature is rare enough for people not to be able to 'learn' what the color means.

1. This one is already known and used in the same cases as the proposed here, so nobody has to learn.

2. We put too much emphasis on recognizing a color. In your proposed solution it will be even worse, because a color would mean only "a campus". But we already reuse colors for different features without fatal consequences, because there's also a context.

> Instead, I'd propose a general 'campus'/'grounds' rendering, so that we can at least display names on such campuses.

I support your idea, but it's a parallel problem for me. It's needed for the areas where we have no idea how to show them. This one is trivial and adds no new color, so we know exactly what to do.

kocio-pl:
I'm still sure that this is a small, well-defined tagging take over, not a new feature that needs campus rendering redesign. The discussion [like this](https://lists.openstreetmap.org/pipermail/tagging/2017-September/thread.html#33651) proves me that landuse instead of amenity helps resolve a lot of confusion.

HolgerJeromin:
@kocio-pl This feature is not rendered with a name. Should I open a new issue for that?

kocio-pl:
If you feel it should render a name (which I was not sure, IIRC), please do it.

polarbearing:
I notices that as well, and would favour a label. Maybe that can be done as a quick PR already, copying the label style from landuse=residential?

kocio-pl:
As you wish. That was not technical problem, it needs decision how should it work and what are pros and cons.

PR #2752, opened 2017-08-15
Change rendering of alpine Hut to vector icon

giggls:
Hello,

while trying to change the @transportation-icon color for german style I stumbled upon the fact that "alpine_hut" was still a raster icon and some inconsistencies in amenity-points.mss.

Thus I made up a vector version of alpine_hut and fixed the inconsistencies.

This is what this pull-request is all about.


kocio-pl:
Could also you drop here some before/after rendering example?

amenity-points.mss
```diff
@@ -24,17 +24,16 @@
 
 /* Note that .points is also used in water-features.mss */
 .points {
-  [feature = 'tourism_alpine_hut'][zoom >= 13] {
-    point-file: url('symbols/alpinehut.p.16.png');
```
sommerluk: As alpinehut.p.16.png would not be used anymore after this PR, the file should be deleted.

giggls:
I doubt that the differences are all that visible, but here we go:
5631-new
5631-old

amenity-points.mss
giggls: OK, branch with pull request updated.

kocio-pl:
You would be surprised how many times pixel peeping shows general problems with rendering... But it looks like an improvement for me, thanks for this contribution!

Small details to be fixed:
- please join the paths (Path > Union) and save it as a plain SVG (instead of "Inkscape SVG")
- you should merge all these commits into one

giggls:
Looks like my knowledge of git stops here. I just tried to merge the 3 commits with the result that I have 5 now instead of 3.

kocio-pl:
I'm also no git expert, so I use gitk for mixed reset, to better control what I'm doing, however this can be done also with CLI (https://git-scm.com/docs/git-reset#git-reset---mixed). After this I add and commit files as usual.

giggls:
*argh* looks like deleting and recreating of the branch with just one commit was not a very good idea :(

imagico:
By the way IIRC if you have locally deleted and re-created a branch you can force-push it to your PR-branch on github without problems (`git push -f`)

kocio-pl:
> while trying to change the @transportation-icon color for german style

Is German style switching to this new fork?

giggls:
Switching to which fork? German style _is_ a fork of openstreetmap-carto. If I have changes which are unrelated to German style like this one I will always try to get them upstream. Because this will make it easier for me to keep German style master in sync to upstream.

kocio-pl:
I thought it still uses XML-based version.

PR #2758, opened 2017-08-16
Do not render link roads until z11

math1985:
This improves rendering on mid zoom levels by smoothing up the highways.

### DIFF
--- a/roads.mss
+++ b/roads.mss
@@ -1164,191 +1164,196 @@ tertiary is rendered from z10 and is not included in osm_planet_roads. */
     }
 
     [feature = 'highway_motorway'] {
-      [zoom >= 5] {
+      [zoom >= 5][link != 'yes'],
+      [zoom >= 11] {
         line-color: @motorway-low-zoom;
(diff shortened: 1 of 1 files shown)


matkoniecz:
I worry a bit about big T shaped junctions causing problems like https://github.com/hotosm/HDM-CartoCSS/issues/236

Is it possible that in some places on Z10 motorways or trunks will appear disjointed after hiding link roads?

math1985:
There will be tiny gaps, like you can see on HOT: http://www.openstreetmap.org/?#map=10/50.0077/20.1551&layers=H
I think the gaps are so small on z10 that they are preferable over the strange bumps on the roads we have now: http://www.openstreetmap.org/?#map=10/51.5087/4.5133

math1985:
Can somebody please review this? I really dislike the effect the link roads currently produce.

screen shot 2017-08-27 at 10 50 01

pnorman:
I'd prefer z10, but am okay with trying z11.

Issue #2728, opened 2017-08-06
Release v4.2.0


kocio-pl:
We could release v4.2.0 about the end of the month. The big news will be probably midzoom changes, but maybe some other visible stuff like railway names and leaf_type rendering will be ready to merge too.

pnorman:
It'd be good to get some changelog updates committed first

kocio-pl:
Of course, I can do it soon.

I'm more concerned with patterns issues, like https://github.com/gravitystorm/openstreetmap-carto/issues/2737. Should we for example revert vector quarry pattern or there's another solution we could try?

kocio-pl:
With new sanitizing code quarry can be fixed, but still we should know [what to do with 512 px patterns](https://github.com/gravitystorm/openstreetmap-carto/issues/937#issuecomment-322666309)? Do we accept them or we plan to move to 256 px with 4.2.0?

kocio-pl:
This release will be loaded with changes, so I wait mainly for better 256 px SVG patterns to be fixed (if it's still the case) and merged - probably @imagico should check them, because he has the most issues with patterns in general.

We don't have to wait longer than that (it will be standard "2-5 weeks" cycle for a minor version change anyway). I keep the changelog updated, it's good to remember we should also update preview image this time to reflect water color change.

Any comments about that?

math1985:
+1

kocio-pl:
With #2769 ready to be merged (I just wait if there are no further roads tuning) we wait for an opinion on:
- #2748 - scrub pattern fix
- #2760 - quarry pattern fix

@imagico, could you look at them and tell if it's safe to merge them or something should be fixed?

Some other PNG->SVG pattern transitions (#2749 - vineyard, #2759 - plant nursery) are not important to merge before v4.2.0, they are just regular PRs to check.

imagico:
> #2748 - scrub pattern fix
> #2760 - quarry pattern fix

I have now commented on both of these but as i also clearly said in https://github.com/gravitystorm/openstreetmap-carto/pull/2727#issuecomment-322042825 i think SVG patterns are not ready for production use here due to the rendering issues observed.  Due to the special situations the quarry pattern does not appear to be affected by this so yes, you can probably safely merge that without making anything worse.

sommerluk:
I agree that it’s better to use PNG pattern when SVG patterns have rendering problems.

I think that, as best practice, nevertheless
- the patterns should be generated by jsdotpattern
[...]

PR #2759, opened 2017-08-16
Switch plant_nursery pattern to SVG - retry [WIP]

kocio-pl:
Follow up to https://github.com/gravitystorm/openstreetmap-carto/pull/2685.

Warsaw, z17
Before
ohdbxqbr
After
zmsqoyhv


math1985:
The distance between the rows dots of dots seem to vary (both before and after);
screen shot 2017-08-19 at 00 04 29
What would cause that?

kocio-pl:
You're right. Looks like a problem with pattern generator:

http://www.imagico.de/map/jsdotpattern.php#x,128,jdp32360;gv,6,32,32;rd,1,0,0,pixel,0.125,4,4,0,jdp23814,749d6c,aedea3

kocio-pl:
Closing because "SVG patterns are not ready for production use" in general (see https://github.com/gravitystorm/openstreetmap-carto/issues/2728#issuecomment-324360423). However the problem still exists with the pattern itself (visible with PNG), so new issue should be created.

Issue #2750, opened 2017-08-14
SVG patterns in landcover-area-symbols render inconsistently in different metatiles


imagico:
This is a followup on the after-merge discussion in #2727.

There we observed differences in SVG pattern rendering depending on unknown conditions leading to differences between metatiles.

The fact that so far the problem analyzed in https://github.com/gravitystorm/openstreetmap-carto/pull/2727#issuecomment-322042825 only seems to occur with kosmtik bothered me.  Since in the released version only three landcover patterns use SVGs, quarry does not use polygon-gamma (so would fall under point 4 in the list) and forest uses opacity in addition which might cause additional problems i looked for cases with scrub.

The OSMF servers use much larger metatiles so this is more difficult and due to the not so dense pattern you cannot usually readily see it but here an example:

https://a.tile.openstreetmap.org/19/278627/184492.png
https://a.tile.openstreetmap.org/19/278870/184612.png

![1](https://a.tile.openstreetmap.org/19/278627/184492.png) ![2](https://a.tile.openstreetmap.org/19/278870/184612.png)

As with beach there are more than two variants, the above is the most extreme case i could find.  And no, this is not double mapping or so, it can be found in different places.  A metatile edge can be seen here:

https://www.openstreetmap.org/#map=19/46.90235/11.48643

So yes, this appears to be a real problem and it already shows up in the map.

It affects patterns in landcover-area-symbols - if observations in kosmtik are representative patterns in the normal landcover layer will render incorrectly with `polygon-gamma` othern than 1 and correctly otherwise (the latter being the current situation for the quarry pattern) `polygon-pattern-gamma` does not appear to have an effect on this.

kocio-pl:
Do you think it should be reported as an issue in Mapnik?

imagico:
In https://github.com/gravitystorm/openstreetmap-carto/pull/2727#issuecomment-320775890 @springmeyer did not consider this a bug.  You can still report it but unless this assessment changes a fix from side of Mapnik is probably unlikely.

PR #2761, opened 2017-08-17
Jewish symbols made lighter

kocio-pl:
Follow up to https://github.com/gravitystorm/openstreetmap-carto/pull/2714#issuecomment-320690830.

Mapnik seems to render SVG in its own way, which is beyond our control and sometimes makes details to vanish, as it is with our current star of David. We can however try a symbol with thinner lines. 

I took [star_of_david3.svg](https://github.com/imagico/jsdotpattern/blob/master/symbols/star_of_david3.svg) from jsdotpattern project. It's rather not pixel aligned, but general religion symbol looks now better, because holes are visible. Cemetery symbol (the same shape scaled down to 10x9 px) is still worse than PNG (no small holes inside), so I would probably not use it:

Before
ntu8hdz1
After
kj9hfe88

Before
hby7cmh7
After
mdlvrzew


nebulon42:
Looks not bad to me. Could link or post an image with the symbol before? Would make it easier for me to compare.
Cemetery is not very visible but recognizable. How does it look currently?

kocio-pl:
I have updated both images with before/after renderings.

I'm not sure now about cemetery. With new symbol the lightness is the same and while I can spot the holes with general symbol with normal size (100-120%), cemetery symbol holes are not visible at 100% - it needs a lot of zooming (for me it's >200%), which is not what typical user would do. So maybe cemetery symbol should be moved to SVG too.

sommerluk:
The general jewish religion symbol „after“ looks much clearer to me than the „before“ version. It can be recognized as Star of David.

About the cemetery symbol: Honestly, both, the „before“ and the „after“ variant, I cannot reliably recognize them as Star of David. They are just to small. But the „after“ variant –  even though stronger – seems to me easier to guess as Star of David.

👍

imagico:
Although i would assume this is clear but just in case it is not - this demonstrates the _incorrect but consistent_ case described in #2750.

kocio-pl:
Sorry, I don't understand it - could you describe in short words what exactly is incorrect and what you think is consistent with this case? Any ideas what should be done to make it OK for you?

imagico:
The results in your _After_ example above are are not a faithful rendition of the SVG file, you can easily see this by comparing these results with a rendering of the pattern with other programs (Inkscape or ImageMagick) or with Mapnik when using `polygon-gamma: 1;`.

_Incorrect but consistent_ refers to what i wrote in #2750:

> It affects patterns in landcover-area-symbols - if observations in kosmtik are representative patterns in the normal landcover layer will render incorrectly but consistently with polygon-gamma other than 1 and correctly otherwise (the latter being the current situation for the quarry pattern) polygon-pattern-gamma does not appear to have an effect on this.

I don't know a way to fix this - otherwise i would have mentioned this in #2750.

kocio-pl:
It looks like people are in favour of this rendering not because it's rendered similar to a vector image, but because it works better (in a cartographic sense) than current PNG image for them - and for me too. 

Solving the problem with Mapnik might give us the ultimate solution to this problem, but this PR would still make a progress with readability and fixes problem with map scaling, so it's worth merging IMO.

PR #2763, opened 2017-08-20
Adding area rendering for amenity=bus_station

kocio-pl:
Part of a transport unification effort, related to #2094. Using the same subtle area color for bus station as for the other transportation areas (taken from the airport area).

Cracow, z19
Before
hallbrvv
After
pk2eehc_

### DIFF
--- a/landcover.mss
+++ b/landcover.mss
@@ -26,7 +26,7 @@
 
 // --- Transport ----
 
-@aerodrome: #e9e7e2;
+@transportation-area: #e9e7e2;
 @apron: #e9d1ff;
 @garages: #dfddce;
 @parking: #eeeeee;
@@ -568,10 +568,11 @@
   }
 
   [feature = 'aeroway_aerodrome'][zoom >= 10],
-  [feature = 'amenity_ferry_terminal'][zoom >= 15] {
-    polygon-fill: @aerodrome;
+  [feature = 'amenity_ferry_terminal'][zoom >= 15],
+  [feature = 'amenity_bus_station'][zoom >= 15] {
+    polygon-fill: @transportation-area;
     line-width: 0.2;
-    line-color: saturate(darken(@aerodrome, 40%), 20%);
+    line-color: saturate(darken(@transportation-area, 40%), 20%);
     [way_pixels >= 4]  { polygon-gamma: 0.75; }
     [way_pixels >= 64] { polygon-gamma: 0.3;  }
   }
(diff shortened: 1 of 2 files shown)


math1985:
This looks quite similar to residential areas to me.

kocio-pl:
That's why I reuse color for a transportation area class - it seems to be working good already for the airport, because it's combined with the icon (and characteristic infrastructure pattern).

HolgerJeromin:
bus_stations are inside residentials. Airports are not.
Both could have an "inside blue glow" like theme parks. Or would that be too much comic style?

kocio-pl:
I would not like to introduce new colors - this effort is about unification (as far as possible). 

At least most of them in my country are not within landuse=residential (like [here](https://www.openstreetmap.org/way/302745085)), so I couldn't spot anything like you suggest, probably because of precise tagging of residential areas. But in fact it should be easier to get then, as with other gray objects - it's always easier to recognize color shade difference than absolute values.

HolgerJeromin:
> I would not like to introduce new colors - this effort is about unification (as far as possible).

Blue is already the transport color.

kocio-pl:
True, but used only for icons and and labels (and even this is not the same shade of blue for buses, trains and the airports). I believe having icon - and infrastructure -  is enough context to recognize transport areas. I wouldn't think that this can be residential, even when there are some buildings, as on the example. Blue areas are somewhat water and ice related for me.

kocio-pl:
Do you feel it's essential for the transport area color to be very different? So what about the airports?

pnorman:
> Do you feel it's essential for the transport area color to be very different? 

If we're going to add a colour, the rendering needs to be clearly distinct from other features. I don't see this achieves that.

> So what about the airports?

I wasn't a huge fan of when we added the fill for airports, but they manage to be distinct by other characteristics - runways are very distinct and obvious.

pnorman:
One option of course is to remove the residential fill, but I'm not sure about that.

kocio-pl:
> If we're going to add a colour, the rendering needs to be clearly distinct from other features. I don't see this achieves that.

This is the same with hospital and school - they are distinct enough, because hospital uses an icon and school doesn't (or would use some other icon), even if they have exactly the same color, which is not the case here. Bus station already uses the icon and we have an area color used in transportation.

The color itself is not all the information the user gets, there's always the context. We reuse colors in many ways.

kocio-pl:
> I wasn't a huge fan of when we added the fill for airports, but they manage to be distinct by other characteristics - runways are very distinct and obvious.

So let's look here: https://github.com/gravitystorm/openstreetmap-carto/pull/2763#issuecomment-323839330 - we have residential area and bust station side by side to compare. Apart from the icon bus station has many bus stops (which are distinctive just like runways) with roads, which are uncommon for residential areas, but are similar to the airport road network (or parking/garages area, which are clearly related to transportation too). The missing thing is dark brown for terminal buildings - which was just not possible before v4, but should be done now. Do you still think bus station area is harder to get then the airport?

PR #2764, opened 2017-08-21
Add ordering to roads-low-zoom

kreed:
There is currently no ordering for the roads-low-zoom layer. The issue is most noticeable at zoom 9, the highest zoom level where the roads-low-zoom layer is used.
download 1

Applying #2732 to roads-low-zoom clears things up.
download 4


math1985:
Yes, I recently also noticed this was missing. Much better now, thanks!

Issue #2291, opened 2016-08-18
Review onboarding


pnorman:
We haven't really looked at this in a long time, and all of the maintainers are experienced enough that getting started is removed from what we do that we're not the best people to consider it. We should review our onboarding.
- What unwritten knowledge do we assume? If we need to assume a certain level of linux, osm, CartoCSS, git, cartography or other knowledge, we should be explicit about it.
- Does our current documentation work as a startup guide?
- Is it clear to developer how to participate when their changes are being reviewed? Even if they're used to **code** reviews, cartography is a bit different. I think this was a problem in #2139 where the PR kept being expanded to cover suggestions that I would have rejected. Not all ideas suggested on PRs are good ideas and sometimes the appropriate response is "I looked at this but it didn't work because..." or "I want to keep this PR focused on ... not ..."
- Is it clear to start out by keeping PRs small? Sometimes it takes experience to know the appropriate size of a "chunk" of changes and when changes really are related, but I'd much rather have PRs too small than too large. After all, it's trivial to merge the too small at the same time, but splitting up the too large is more work.
- Is it clear where to get help with the basics? GitHub isn't _exactly_ the right place for asking questions on getting started but we've used it for that.

math1985:
Thanks for opening this issue, I'm very curious for the response from new or potential contributors.

don-vip:
Something strange at first glance is the ownership of this repository. If you want maximal visibility as it is now, for a long time, the de-facto standard OSM style, shouldn't it be transferred to @openstreetmap organization?

pnorman:
> Something strange at first glance is the ownership of this repository. If you want maximal visibility as it is now, for a long time, the de-facto standard OSM style, shouldn't it be transferred to @openstreetmap organization?

I've preferred **not** having it on the @openstreetmap organization. Part of this is because I don't know what being on that organization would entail, part of it is I view this stylesheet as the one chosen to be on tile.osm.org not the stylesheet for osm.org, and part of it is that it's fairly common for OSM software projects not to be on the openstreetmap organization. Examples include JOSM, taginfo, cgimap, and overpass.

The traffic information from GitHub is maintainer only, but we're getting >50 unique cloners and >1k UVs a week and plenty of activity on the issue tracker. I think it's an issue of converting that to contributors.

CC @geostonemarten and @Ircama as people who have recently gotten started

nebulon42:
[...]

PR #2767, opened 2017-08-21
Make installation process more explicit

matkoniecz:
subtask of #2291 

I think that `Quick start` section should be self-contained and require no digging in other parts of documentation (especially with "Read on below to get the details.").

It reduces confusion - for example at start I missed fact that `openstreetmap-carto` is obviously necessary. Info how to see anything is currently hidden in the later part.

### DIFF
--- a/DOCKER.md
+++ b/DOCKER.md
@@ -17,7 +17,7 @@ If you are eager to get started here is an overview over the necessary steps.
 Read on below to get the details.
 
 * clone openstreetmap-carto repository
-* Download OpenStreetMap data in osm.pbf format to a file `data.osm.pbf` and place it within the openstreetmap-carto directory.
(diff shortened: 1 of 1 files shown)


kocio-pl:
BTW: it would be nice to fix a point _Download OpenStreetMap data in osm.pbf format to a file data.osm.pbf and place it within the openstreetmap-carto directory._ to not look like a full sentence (D->d, remove the dot on the end).

matkoniecz:
> it would be nice to fix a point Download OpenStreetMap data in osm.pbf format to a file data.osm.pbf and place it within the openstreetmap-carto directory. to not look like a full sentence (D->d, remove the dot on the end

I am not sure is it improvement - is it wrong to have a full sentence here?

kocio-pl:
Yes, because all the other points are not formatted like this.

matkoniecz:
> Yes, because all the other points are not formatted like this.

Changed.

PR #2769, opened 2017-08-22
Revert white color for unimportant roads

### LINKED ISSUE #2731: White mid-zoom roads do not work on some backgrounds
With the changes in 5fd79b7bfad1d895c543e52838c65657fcd9fe1a, there is a huge difference in how the roads look between different areas

image

In Surrey, the roads are readable with the residential background. In Langley, they're not, with most of them against our base colour.

kocio-pl:
Resolves #2731.

Reverts unimportant roads color from white to @residential-casing color (as in v4.1.0).

Before revert:
bzq1gjxn
After revert:
g crgmz5

### DIFF
--- a/roads.mss
+++ b/roads.mss
@@ -40,7 +40,7 @@
 
 @tertiary-shield: #3b3b3b;
 
-@unimportant-road: white;
+@unimportant-road: @residential-casing;
 
 @residential-construction: #aaa;
 @service-construction: #aaa;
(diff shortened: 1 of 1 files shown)


meased:
I prefer the white.

math1985:
We might need to re-add https://github.com/gravitystorm/openstreetmap-carto/pull/2654/files#diff-f61408bc1ffb6636dbaf2bf9e38a4177L1043 and further too.

kocio-pl:
@meased I think there will be lot of discussions with other changes, so we can change this color later.

@math1985 Do you mean this block only (till https://github.com/gravitystorm/openstreetmap-carto/pull/2654/files#diff-f61408bc1ffb6636dbaf2bf9e38a4177L1068) or something more? Altering road code is quite a challenge nowadays.

math1985:
I think so, but make sure to test it anyway. Road width might need changing too.

kocio-pl:
I'm not sure how to test it - with reverted `::halo` section it works at least (code compiles and roads don't look like broken or missing).

I'm not even sure if anything more than color change was the problem to solve.

roads.mss
```diff
@@ -1048,30 +1048,32 @@ tertiary is rendered from z10 and is not included in osm_planet_roads. */
 .tunnels-fill[zoom >= 10] {
 
   ::halo {
-    [feature = 'highway_motorway'][link != 'yes'][zoom >= 8][zoom < 12],
```
math1985: This block should be kept (it is the halo for low-zoom roads). So we need both blocks now.

kocio-pl:
Done and it works, anything else that should be done here?

PR #2770, opened 2017-08-23
Refine rendering of campsites

giggls:
I'm unsure if something like this should be discussed as an issue first. It comes in the form of a pull request because the feature it is already active in German style.

The Wiki proposes refined rendering of campsites for years now.

Here is an implementation:
* Render camping/caravaning icon for campsites with tags tent=yes and caravans=yes
test-camp-caravan
* Render unchanged in case of caravans=no or unset
test-campsite
* Render backcountry icon if tagged backcountry=yes
test-backcountry

To me there are two known issues:
- The blue color does not look very nice. This is why I changed accomondation color to amenity brown in German style.
- The backcountry icon could be more intuitive. I don't know if the trees are recognized as such. I'm open to other suggestions. I already experimented with campfire icons which where even less noticeable.


imagico:
In general if you want to suggest rendering new features it is good practice to do some research into whether these tags are well established and consistently used and documented.  This looks rather positive in this case though.

Large icons are a problem though, [current policy says](https://github.com/gravitystorm/openstreetmap-carto/blob/master/CONTRIBUTING.md) they should be 14x14 pixel.  You could try if the tent symbol is still recognizable at a smaller size.

With the backcountry icon you could try showing a hiker instead of trees - this would also be less climate specific.

math1985:
I think this rendering is too specific for a general purpose map.

giggls:
While this might be the true in the backcountry case it is shurely not in camping and/or caravaning. Even road signs differ between them.

@imagico: 14x14 is the size of the original tent symbol. Making it smaller will make it more or less irrecognizable. Campgrounds are usually fairly large objects Thus the slightly bigger size of the rendering will IMO look better that stying at 14x14.

kocio-pl:
@giggls Are you willing to try with smaller icons? I think that if they have a common visual thread (a tent) it wouldn't be a problem to recognize them as kind of campsite, no matter if we know what exactly type it is, so a size is the only thing that keeps me from approving this PR.

giggls:
I already tried using standard sized icons but discarded the idea. . Especially the combined tent+caravan icon is barely recognizable. It might work with the tent+trees one.
IMO the standard size campsite icon is already very small anyway compared to the size of the physical object they usually represent.

kocio-pl:
> IMO the standard size campsite icon is already very small anyway compared to the size of the physical object they usually represent.

But that's true with every icon - see for example airport. The area however is a different thing - maybe we lack borders.

I think caravan icon is not really designed (it doesn't look consistent), but backcountry icon is more promising for me. You could squeeze it probably - one tree could be enough, I guess.

kocio-pl:
It looks like the icon redesign is not going to occur soon, so I close this ticket, but a new one can be made once we have more compact icons. Thanks anyway!

Adamant36:
@Tomasz-W, this is one I think would be cool and useful to see. Especially for backcountry. Any chance you could come up with icons that fit the 14x14 standard? I imgine backcountry could possibly be doable if not the other one.

Adamant36:
@giggls, its doable. @jeisenbe had a PR (or maybe an issue) a while back for springs that used it. I'll try to find it or maybe he can enlighten us as to how its done. If I remember correctly it got rejected in that instance, but maybe it would work in this one.

jeisenbe:
It’s possible to make an icon from 2 separate svg files, but in this case I
think we can just make new svgs for each retype of campsite which fit into
the 14*14 pixel canvas.

(The idea with springs and wells was to have a standard way to identify if
a water source offered drinking water, so a cup of water could be added to
many different features; eg water tap, well, spring, water point, etc -
this was @imagico ’s idea. You can see the code and icon on his alt-colors
branch)

Adamant36:
@jeisenbe, your probably right. I don't see why we couldn't fit the information in a 14x14 canvas somehow. Thinks for the extra information. I knew it was something like that.

PR #2775, opened 2017-08-24
Revert "Switch sand and wetland patterns to SVG"

kocio-pl:
Reverts gravitystorm/openstreetmap-carto#2727


kocio-pl:
Both patterns have some Mapnik-related issues, so I prefer to switch them back soon. What do you think?

kocio-pl:
Reverting now, pattern review is needed to decide what we want to do with them.

PR #2778, opened 2017-08-26
Don't create populated places file for Travis

pnorman:


kocio-pl:
What is the purpose of this?

pnorman:
Files were being created which we don't use anymore

PR #2786, opened 2017-08-30
Changing embassy color to amenity-brown

kocio-pl:
Follow up to @giggls https://github.com/gravitystorm/openstreetmap-carto/pull/2765#issuecomment-323748719.

Embassy icon is rather kind of office (like town hall) than transportation amenity (like bus stop), so I prefer rendering it as such.

Before
fncnbimh
After
1ntyippb


kocio-pl:
Any comments or reviews for this PR?

PR #2788, opened 2017-08-31
Same line wrap of amenitys for all zoom levels

sommerluk:
As proposed by @math1985 at https://github.com/gravitystorm/openstreetmap-carto/pull/2730#issuecomment-322591915 here is a PR that unifies the line wrap for amenity points over all zoom levels. I think that indeed it makes sense, because amenity points are rather small (at difference to other features like cities or lakes that have larger line wrap at higher zoom levels because they are much bigger), and anyway the difference is not big compared to the current code. This PR will make zoom-in/zoom-out less surprising, because you get always the same line wrap for amenities.

### DIFF
--- a/amenity-points.mss
+++ b/amenity-points.mss
@@ -13,7 +13,7 @@
 @landform-color: #d08f55;
 
 @landcover-font-size: 10;
-@landcover-wrap-width-size: 25; // 2.5 em
+@landcover-wrap-width-size: 30; // 3 em
 @landcover-line-spacing-size: -1.5; // -0.15 em
 @landcover-font-size-big: 12;
 @landcover-wrap-width-size-big: 36; // 3 em


kocio-pl:
Test examples fixed by this change:

Poland
- http://www.openstreetmap.org/relation/4519921#map=8/50.216/20.300
- http://www.openstreetmap.org/relation/4519921#map=10/50.5954/19.5179

Warsaw
- http://www.openstreetmap.org/way/190840101#map=13/52.2462/20.9651
- http://www.openstreetmap.org/way/190840101#map=14/52.2525/20.9604

Works as advertised, thanks!

math1985:
Thanks, I should have done it done it like this from the beginning!

math1985:
I created also an issue at Mapnik, requesting to be able to specify wrap in em natively: https://github.com/mapnik/mapnik/issues/3763

Issue #2781, opened 2017-08-27
Tramway tracks are rendered wrong when comined with a psv road


menzo2003:
In Amsterdam there are lots of tramways that are paved where next to trams only psv vehicles are allowed.

The tags are commonly: 
access=no
electrified=contact_line
frequency=0
gauge=1435
highway=service
oneway=yes
psv=yes
railway=tram
voltage=600

Up till recent it was rendered as a road with a black line trough the middle. Now it is sometimes rendered as a road with access=no (dashed white road)

This is a good example of the problem: 
http://www.openstreetmap.org/#map=18/52.36308/4.87601

math1985:
@pnorman Could this have been introduced by #2732?

pnorman:
I'm not able to reproduce. I went into my database and set one row to have
```
┌─[ RECORD 1 ]──────┐
│ highway │ service │
│ railway │ tram    │
│ oneway  │ yes     │
│ access  │ no      │
└─────────┴─────────┘
```

I'm not too surprised to see problems with ways tagged with both highway and railway tags, they've always been a headache and most maps don't handle them.

Since I can't reproduce it, I can't bisect, but I'm guessing it comes down to ordering, differences between the two times the road is selected, but not in the roads fill or casing layers.

kreed:
This is a regression from #2732. Both features have the same z_order because they are created from the same database row, so the ordering is now undefined.

Same problem occurs with combined highway/taxiways

pnorman:
> This is a regression from #2732. Both features have the same z_order because they are created from the same database row, so the ordering is now undefined.

Perhaps the easiest solution is to add `feature` to the sort

dieterdreist:
sent from a phone

> On 27. Aug 2017, at 20:29, menzo2003 <notifications@github.com> wrote:
> 
> The tags are commonly:
> access=no
> electrified=contact_line
> frequency=0
> gauge=1435
> highway=service
> oneway=yes
> psv=yes
> railway=tram
> voltage=600


are there separately mapped footways or other parallel roads? Maybe not a big issue (because nobody likely does tram-routing), but I think you are excluding trams with this tagging (one might argue they are psv, but the wiki doesn't seem to define it like this, and there are typically also non-psv trams (not in service or maintenance vehicles)). Rather than access=no you could add vehicle=no (according to the wiki, vehicle doesn't contain trams)

SomeoneElseOSM:
[...]

PR #2790, opened 2017-09-01
Fix ordering regressions

### LINKED ISSUE #2779: Railway yard tracks are rendered above main line tracks
Not sure if this has been fixed already, but [yard tracks are being layered above main line tracks](https://www.openstreetmap.org/#map=14/22.4189/114.0662).

kreed:
Fixes #2779 and #2781

### DIFF
--- a/project.mml
+++ b/project.mml
@@ -518,6 +518,7 @@ Layer:
           ORDER BY
             layernotnull,
             z_order,
+            CASE WHEN feature IN ('railway_INT-preserved-ssy', 'railway_INT-spur-siding-yard', 'railway_tram-service') THEN 0 ELSE 1 END,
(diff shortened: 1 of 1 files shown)


math1985:
Thanks for this PR! I haven't tested it yet but looks good.

HolgerJeromin:
Do we want to push a bugfix release?

kocio-pl:
Since we already meet [minimum 2 weeks period](https://github.com/gravitystorm/openstreetmap-carto/blob/master/README.md#versioning) for minor version and there are more small changes, not only bugfixes, I would make it v4.3.0 with release time witihin 2-3 days.

pnorman:
Pushing a bugfix release would be applying these commits to 4.2.0 and tagging 4.2.1 and should be independent of if we want to tag 4.3.0 yet.

I don't mind doing this, but given the limited scope of the problem, don't see it as necessary.

PR #2797, opened 2017-09-01
Propose solution to highlight place areas and names

stevenLAD:
Highlights area and name. Applied only for villages and hamlets prior to approval

### DIFF
--- a/amenity-points.mss
+++ b/amenity-points.mss
@@ -555,7 +555,7 @@
   }
 
   [feature = 'historic_memorial'][zoom >= 17] {
-    marker-file: url('symbols/memorial.svg');    
+    marker-file: url('symbols/memorial.svg');
     marker-fill: @amenity-brown;
     marker-placement: interior;
     marker-clip: false;
--- a/amenity-points.mss
+++ b/amenity-points.mss
@@ -555,7 +555,7 @@
   }
 
   [feature = 'historic_memorial'][zoom >= 17] {
-    marker-file: url('symbols/memorial.svg');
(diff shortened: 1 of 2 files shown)


kocio-pl:
Could you split the real place are code and small code fixes? There's a lot of changes mixed. Some before/after images are needed too.

HolgerJeromin:
I think the "small code fixes" was done by the used text editor not by steven himself...

kocio-pl:
It would still be good to split them out. :smile:

stevenLAD:
Outside the actual code changes made, the only other changes were made by Atom which has removed superfluous spaces. In Github Desktop this is clear. 

To revert, apply the Atom cleanup, wait for it to be accepted and then add the changes would only delay.

kocio-pl:
The problem is that is hard to review the actual code now. It will be nice to have the code cleaned, so please start with Atom changes so we could apply it first.

kocio-pl:
Please make it 1 commit (instead of 3) and rename this PR accordingly.

stevenLAD:
I hope this suffices - it appears clear to me that there are two issues: one using the GitHub proposed editor, and the second the actual code changes.

As a ps, the project.mml code can be further cleaned to remove unrelated duplicity but I will hold on this until this commit is finalised.

kocio-pl:
I meant to split it into 2 PRs, not 2 commits (BTW - there is still 4 of them). Code cleaning might be done quickly, but the rest needs some time.

stevenLAD:
Close Pull request to start over and make changes as per request of Kocio-pl

PR #2798, opened 2017-09-02
Master

stevenLAD:
Remove excess spaces in files as identified by ATOM in order to allow clearer identification of real code changes


stevenLAD:
Need to reinitialise branch to avoid excess commits

PR #2799, opened 2017-09-02
Null edit to remove excess spaces as identified by Atom

stevenLAD:
Atom removed these excess spaces in null edits. Committing to avoid
these changes masking real changes and hence rendering audit more
complicated

### DIFF
--- a/amenity-points.mss
+++ b/amenity-points.mss
@@ -555,7 +555,7 @@
   }
 
   [feature = 'historic_memorial'][zoom >= 17] {
-    marker-file: url('symbols/memorial.svg');    
+    marker-file: url('symbols/memorial.svg');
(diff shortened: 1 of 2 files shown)


stevenLAD:
Question: Do I now have to branch my code in order to create a separate pull request for the actual changes?

math1985:
Thanks!

nebulon42:
@stevenLAD you should work with branches when you plan to make subsequent pull requests. Branching avoids that you alter master and helps you to keep master up-to-date with the latest changes from upstream.

kocio-pl:
I've just found that one can just disable automatic removing trailing spaces in Atom:

https://superuser.com/questions/961352/atom-automatically-removing-trailing-spaces/1191787

PR #2800, opened 2017-09-02
Add names to places that are mapped as areas

stevenLAD:
Proof of concept using villages and hamlets where confusion with places
identified as points should be easily identifiable and rectifiable by
contributors.

Can be extended to larger place areas.

Typography can be changed in future to be aligned with existing point
place typography, but not made now to simplify audit


stevenLAD:
Closed and replaced with corrected and more structured pulls

PR #2802, opened 2017-09-03
AddShadingToPlaceAreas

stevenLAD:
Excess spaces removed by Atom committed to avoid confusion when actual
edits made

PR #2803, opened 2017-09-03
Creature features and text-polys to use in naming and shading places

stevenLAD:
Create features and text-polys to use in shading and naming places that
are defined as areas.

Replaces Pull 2800 which did not create features.

Subsequent Pulls in new branches will implement these features

### DIFF
--- a/project.mml
+++ b/project.mml
@@ -135,6 +135,7 @@ Layer:
                                                     'allotments', 'forest', 'farmyard', 'farmland', 'greenhouse_horticulture',
                                                     'recreation_ground', 'village_green', 'retail', 'industrial', 'railway', 'commercial',
                                                     'brownfield', 'landfill', 'construction', 'plant_nursery') THEN landuse ELSE NULL END)) AS landuse,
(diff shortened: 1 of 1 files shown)


stevenLAD:
Currently only addresses Vilages and Hamlets where any conflicts with place points should be easily identifiable by contributors, but can be extended to other places if required / desired

pnorman:
Closing since this doesn't implement anything and you've made #2802 and #2804

stevenLAD:
I was told on my first pull to beak things down into small chunks using separate pull requests. With 2803 (this pull) I created objects, and then with 2804 I rendered the names and 2806 the areas, the intention being to allow use of names areas or both, and to not confuse the creation and the rendering. Clearer guidelines as to how Pull requests are to be structured needs to be available, as different people appear to have different ideas.

pnorman:
#2804 contains the commits in this PR, so it replaces this.

The problem with #2797 is you were doing unrelated whitespace cleanups in the same PR as implementing something new. You should have one PR for whitespace cleanups and another for implementing something new.

stevenLAD:
Thank you for the clarification. I have followed this practice since.

PR #2804, opened 2017-09-03
Create Place names for place areas

stevenLAD:
Create place names for places that are defined as place areas

Replaces pull #2800

### DIFF
--- a/amenity-points.mss
+++ b/amenity-points.mss
@@ -1523,7 +1523,9 @@
   [feature = 'leisure_stadium'],
   [feature = 'leisure_track'],
   [feature = 'leisure_dog_park'],
-  [feature = 'leisure_pitch'] {
+  [feature = 'leisure_pitch'],
+  [feature = 'place_village'],
+  [feature = 'place_hamlet'] {
     [zoom >= 10][way_pixels > 3000][is_building = 'no'],
     [zoom >= 17][is_building = 'no'],
     [zoom >= 10][way_pixels > 3000][feature = 'shop_mall'],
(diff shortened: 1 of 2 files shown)


lukaszkruk:
Do I understand correctly this only does it for villages and hamlets? How about cities, towns, neighbourhoods and others?

stevenLAD:
You are correct. If this pull had been approved and merger, the rendering should have been then changed to comply with Place name rendering. This was an attempt to go in small steps so as not to be rejected "because too much at a time"

stevenLAD:
This request was a proof of concept - by doing only for villages and hamlets, it minimised potential conflicts with place node names. In these smaller areas, these conflicts would be easily identifiable and rectifiable by contributors. This was described in the pull that created the objects.

lukaszkruk:
@pnorman by 'renders village and hamlet polygons like residential areas' do you mean it shades areas tagged as place but does not render labels from the name tag?

apologies, my understanding is not good enough yet to figure this out myself.

pnorman:
> @pnorman by 'renders village and hamlet polygons like residential areas' do you mean it shades areas tagged as place but does not render labels from the name tag?

There's nothing in the PR text about village and hamlet polygons like residential areas, yet that's what it does.

matkoniecz:
> There's nothing in the PR text about village and hamlet polygons like residential areas, yet that's what it does.

And it is not explained why it would be a good idea, given that villages and hamlets include also nonresidential areas.

Also - what about cities, towns, neighbourhoods and others?

You are welcomed to open a PR with labels for places tagged on areas. Opening a PR similar to this one would require a really good and convincing explanation why places tagged as ares should be rendered as landuse=residential

matkoniecz:
> You are welcomed to open a PR with labels for places tagged on areas

Note that #2816 ma be already implementing what you want.

PR #2806, opened 2017-09-03
Add Shading for places defined as areas

stevenLAD:
Add shading for places defined as areas. Currently only applies to
Villages and Hamlets but can be extended. Opacity is low to expose
other land use within villages.

### DIFF
--- a/landcover.mss
+++ b/landcover.mss
@@ -474,7 +474,7 @@
     [way_pixels >= 4]  { polygon-gamma: 0.75; }
     [way_pixels >= 64] { polygon-gamma: 0.3;  }
   }
- 
+
   [feature = 'wetland_swamp'][zoom >= 8] {
     polygon-fill: @forest;
     [way_pixels >= 4]  { polygon-gamma: 0.75; }
@@ -766,7 +766,7 @@
       b/line-color: @tourism;
       b/line-opacity: 0.3;
       b/line-join: round;
-      b/line-cap: round;    
+      b/line-cap: round;
     }
     [zoom >= 17] {
       a/line-width: 2;
--- a/landcover.mss
+++ b/landcover.mss
@@ -227,6 +227,25 @@
     [way_pixels >= 64] { polygon-gamma: 0.3;  }
   }
 
+  [feature = 'place_village'],
+  [feature = 'place_hamlet'] {
+    [zoom >= 10] {
+      polygon-fill: @built-up-lower-lowzoom;
+      polygon-opacity: 0.3;
+      [zoom >= 11] { polygon-fill: @built-up-upper-lowzoom; }
+      [zoom >= 13] { polygon-fill: @residential; }
+      [zoom >= 16] {
+        line-width: .5;
+        line-color: @residential-line;
+        [name != ''] {
+          line-width: 0.7;
+        }
+      }
+    }
+    [way_pixels >= 4]  { polygon-gamma: 0.75; }
+    [way_pixels >= 64] { polygon-gamma: 0.3;  }
+  }
+  
   [feature = 'landuse_garages'][zoom >= 13] {
     polygon-fill: @garages;
     [way_pixels >= 4]  { polygon-gamma: 0.75; }
--- a/landcover.mss
+++ b/landcover.mss
@@ -228,24 +228,21 @@
   }
 
   [feature = 'place_village'],
-  [feature = 'place_hamlet'] {
-    [zoom >= 10] {
-      polygon-fill: @built-up-lower-lowzoom;
-      polygon-opacity: 0.3;
-      [zoom >= 11] { polygon-fill: @built-up-upper-lowzoom; }
-      [zoom >= 13] { polygon-fill: @residential; }
-      [zoom >= 16] {
-        line-width: .5;
-        line-color: @residential-line;
-        [name != ''] {
-          line-width: 0.7;
-        }
+  [feature = 'place_hamlet'][zoom >= 10] {
+    polygon-fill: @built-up-lower-lowzoom;
+    [zoom >= 11] { polygon-fill: @built-up-upper-lowzoom; }
+    [zoom >= 13] { polygon-fill: lighten(@residential, 30%); }
+    [zoom >= 16] {
+      line-width: .5;
+      line-color: lighten(@residential-line, 30%);
+      [name != ''] {
+        line-width: 0.7;
       }
     }
     [way_pixels >= 4]  { polygon-gamma: 0.75; }
     [way_pixels >= 64] { polygon-gamma: 0.3;  }
   }
-  
+
   [feature = 'landuse_garages'][zoom >= 13] {
     polygon-fill: @garages;
     [way_pixels >= 4]  { polygon-gamma: 0.75; }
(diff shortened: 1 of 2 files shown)


HolgerJeromin:
Can you provide screenshots?
Opacity is normaly not used as it has ugly results when stacked

stevenLAD:
As place areas are a superset of land uses, I used opacity to not obscure other land uses based on its use in Danger Areas http://www.openstreetmap.org/#map=13/51.2117/-1.6796. However without opacity, i do not believe that it will obscure other land uses cf anywhere in the Netherlands where landuse=residential is used over large areas encompassing many other land uses.

I am not fixed on this (opacity) and bow to your experience as to its use, as well as to the actual shading tone used - my goal was to add shading.

kocio-pl:
Could you post some renderings to show how it would look like?

stevenLAD:
I do not have any renderings, and have not yet worked out how to install the rendering engine. My mods are based on existing code and images.  Perhaps someone with more experience and better facilities could try? I will look at installing later but it will be some time from now.

kocio-pl:
You could probably start with this:

matkoniecz:
Is it intentional to display  

> [zoom >= 13] { polygon-fill: @residential; } 

over entire area of place, including areas that are not a residential areas?

Testing changes is a part of making pull request and it is usually the most time-consuming part.

kocio-pl:
I think of `place` key as a toponym, so the only relevant part for rendering should be a name label.

matkoniecz:
> I think of place key as a toponym, so the only relevant part for rendering should be a name label.

Yeah, I also I think the same. Typical settlement includes many different landuses, not only residential area.

stevenLAD:
If a place is a toponym, then why would key values be village, town suburb etc - there appears to be some inconsistency there.

kocio-pl:
Probably because the name and location might be the only thing we know about such place, not which landuses are there. If we know them, they can be added too.

stevenLAD:
A toponym is related to topography. Kocio-pl - Your comment highlights the need to shade places; we may know location and extent, but not necessarily land uses or name. Nothing precludes adding land uses later, hence the proposal to use opacity, or even replace with land uses. At present, landuse=residential is widely abused to highlight villages and towns, and this proposal would remove that need.

This subject has been the subject of discussion since.......and even the wiki references it as a bug!

dieterdreist:
2017-09-04 16:13 GMT+02:00 stevenLAD <notifications@github.com>:

> If a place is a toponym, then why would key values be village, town suburb
> etc - there appears to be some inconsistency there.
>


I can't follow you. Aren't those oikonyms, which are a subclass of
toponyms?  Actually, more precisely not the place object is the toponym,
the name=* (and variants) on an object with place=* is the toponym.

stevenLAD:
Etymology apart, the name applies to a place=village, hamlet etc. These are not points but extents or areas and OSM allows their definition as such. Each area is a superset of landuses. This pull was intended to give shading that covers the entire area without obscuring any areas (landuses that may or may not exist) within. 

If only name is shown, then the map looks like the cycle map - there is something somewhere without an extent.

To cover this in many areas landuse=residential has been overused to mark the extent including items that are not residential. Going forward, this would allow initial definition of an area that could be subsequently enhanced with more granular and accurate landuse areas.

stevenLAD:
To illustrate my point, below are 3 areas of Angola, all at zoom 13. The first uses residential areas and place points.
http://www.openstreetmap.org/#map=13/-10.6879/14.3366

The next 2 are the same area using villages, the first in the standard layer
http://www.openstreetmap.org/#map=13/-9.9566/14.8158

and this one in the cycle map layer
http://www.openstreetmap.org/#map=13/-9.9566/14.8158&layers=C

which one speaks the clearest to the user about the extent and existence of population? I think that the first one, but is all of that landuse really residential?

while it is possible to achieve the result using the first approach, it pollutes the landuse database which could become a valuable asset.

kocio-pl:
Rendering examples are really needed to see how it differs from landuse=residential. I still feel that we have different tools for showing boundaries and landuses and place area is not any of them, it's just more precise if the place we are naming takes more space than just a point - but actually rendering that area may clash with landuses and boundaries.

landcover.mss
```diff
@@ -766,7 +785,7 @@
       b/line-color: @tourism;
       b/line-opacity: 0.3;
       b/line-join: round;
-      b/line-cap: round;    
+      b/line-cap: round;
```
pnorman: whicespace noise

landcover.mss
```diff
@@ -474,7 +493,7 @@
     [way_pixels >= 4]  { polygon-gamma: 0.75; }
     [way_pixels >= 64] { polygon-gamma: 0.3;  }
   }
- 
+
```
pnorman: whitespace noise

landcover.mss
```diff
@@ -227,6 +227,25 @@
     [way_pixels >= 64] { polygon-gamma: 0.3;  }
   }
 
+  [feature = 'place_village'],
+  [feature = 'place_hamlet'] {
+    [zoom >= 10] {
+      polygon-fill: @built-up-lower-lowzoom;
+      polygon-opacity: 0.3;
```
pnorman: Don't use opacity.

Nakaner:
The SQL query of the landcover layer does currently only order features by ST_Area, i.e. larger polygons are drawn below smaller polygons. If a `place=village` polygon is covers all the area of a village (keep in mind that villages can have 1000 inhabitants and more!) and the landuse in this area is mapped in small pieces, your place polygon will be rendered below all those small landuse polygons.

stevenLAD:
That was my intention. The extent of the village should not prime over factual and accurate landuses as the village extent would normally be a superset of these with eventual common boundaries and was intended particularly for lower zooms. The shading should be adapted accordingly so as to deconfuscate ( a lighter shade of grey would appear ideal, but is not included).

landcover.mss
stevenLAD: As already said - the intention in using opacity was to not obfuscate finer detail. This can be removed if preferred.

landcover.mss
stevenLAD: This has already been removed as part of a whitespace cleanup.

landcover.mss
stevenLAD: This has already been removed as part of a whitespace cleanup

landcover.mss
pnorman: Yes, but it shouldn't be in *this* PR as well.

landcover.mss
pnorman: Given my basic concerns with rendering places like built up areas I'm not sure what changes you should make, except you need to find a way without opacity.

pnorman:
> Rendering examples are really needed to see how it differs from landuse=residential.

Yes, we really need [a few images selected to show the changes](https://github.com/gravitystorm/openstreetmap-carto/blob/master/CONTRIBUTING.md#pull-requests) in order to meaningfully discuss this.

> I still feel that we have different tools for showing boundaries and landuses and place area is not any of them,

👍

stevenLAD:
Above commit removes opacity as requested and generally agreed. It also lightens the shading so that residential areas, if defined, will be easily distinguishable. The same colour is used as villages and hamlets will be majority residential so it appears to be consistent

kocio-pl:
What about rendering examples? It's very important to show real examples of the output and test if it's really working. 

If you don't have testing environment and don't know where to start, try these instructions:

https://ircama.github.io/osm-carto-tutorials/

matkoniecz:
We really need [a few images selected to show the changes](https://github.com/gravitystorm/openstreetmap-carto/blob/master/CONTRIBUTING.md#pull-requests) in order to meaningfully discuss this.

Once you have some please open a new PR or request reopening this one.

PR #2809, opened 2017-09-05
Moving emergency phone to a new tagging scheme

### LINKED ISSUE #318: Add rendering for emergency=phone
JOSM validator complains about amenity=emergency_phone being deprecated and auto-corrects it to emergency=phone when using the "Fix" button. However, the latter is not rendered by the carto style, thus all emergency phones disappeared from the map in some areas [1] recently.

I'd suggest to also include emergency=phone with the same render rule as amenity=emergency_phone in the carto style.

[1] Example for a recent mass update turning amenity=emergency_phone into emergency=phone: http://www.openstreetmap.org/changeset/19082658

kocio-pl:
Resolves #318.

Name selection in SQL was removed because we're not displaying names for such POIs.

### DIFF
--- a/project.mml
+++ b/project.mml
@@ -1389,7 +1389,7 @@ Layer:
                                                   'fuel', 'hospital', 'ice_cream', 'embassy', 'library', 'courthouse', 'townhall', 'parking',
                                                   'bicycle_parking', 'motorcycle_parking', 'pharmacy', 'doctors', 'dentist', 'place_of_worship',
                                                   'police', 'post_box', 'post_office', 'pub', 'biergarten', 'recycling', 'restaurant', 'food_court',
-                                                  'fast_food', 'telephone', 'emergency_phone', 'taxi', 'theatre', 'toilets', 'drinking_water',
+                                                  'fast_food', 'telephone', 'taxi', 'theatre', 'toilets', 'drinking_water',
                                                   'prison', 'hunting_stand', 'nightclub', 'veterinary', 'social_facility',
                                                   'charging_station', 'arts_centre', 'ferry_terminal') THEN amenity ELSE NULL END,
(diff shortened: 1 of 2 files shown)


project.mml
```diff
@@ -1525,6 +1528,7 @@ Layer:
             OR "natural" IN ('peak', 'volcano', 'saddle', 'spring', 'cave_entrance')
             OR historic IN ('memorial', 'monument', 'archaeological_site', 'wayside_cross')
             OR tags->'memorial' IN ('plaque')
+            OR tags->'emergency' IN ('phone')
```
pnorman: This [should be `tags @> 'emergency=>phone'`](https://github.com/gravitystorm/openstreetmap-carto/blob/master/CONTRIBUTING.md#sql-style-guidelines), as well as the repetition in the WHERE clause for the other layer.

kocio-pl: Hm, but why should we check if a tag is in the hstore? Why do we need this special construction at all?

pnorman: If we don't include an item in the OR part of the WHERE clause it won't be selected in the layer.

kocio-pl: The icon on a node was present in my test rendering, so it looks for me that was selected.

pnorman: The problem is `OR tags->'emergency' IN ('phone')` does not follow our style standards.

kocio-pl: I see. Does `OR tags->'memorial' IN ('plaque')` have this problem too?

pnorman: Yes, but you don't need to change it in this PR, it's an existing error.

kocio-pl:
I have changed SQL formatting as requested. 

I have also removed support for rendering such polygons, because there are only 34 of them, Wiki definition does not allow areas and they all look like mistagging (property of building or part of a road).

PR #2818, opened 2017-09-09
Fixing line wrapping bug in Docker

### LINKED ISSUE #2755: kosmtik_1  | Mapnik LOG> 2017-08-15 22:19:58:  could not create BreakIterator: U_MISSING_RESOURCE_ERROR
If I run the stylesheet with Docker, I get the following errors:

> kosmtik_1  | Mapnik LOG> 2017-08-15 22:19:58:  could not create BreakIterator: U_MISSING_RESOURCE_ERROR

Also the labels are no longer wrapped over separate lines.

Is anybody else running into the same problem?

kocio-pl:
Resolves #2755.

There are different ways to skin this cat, but this is the easiest and fastest for us to use. I've changed the import container distribution too just to be consistent with a kosmtik container.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM debian:stretch
+FROM ubuntu:xenial
 
 # Style dependencies
 RUN apt-get update && apt-get install --no-install-recommends -y \
--- a/Dockerfile.import
+++ b/Dockerfile.import
@@ -1,4 +1,4 @@
-FROM debian:stretch
+FROM ubuntu:xenial
 
 RUN apt-get update && apt-get install --no-install-recommends -y \
     curl ca-certificates osm2pgsql postgresql-client \


math1985:
Isn't ubuntu quite a bit larger than debian?

I suppose something based on [Alpine](https://hub.docker.com/_/alpine/) will not work?

pnorman:
Given that Mapnik can break in subtle ways (e.g. this bug) I don't recommend using ubuntu:latest, but sticking to a specified version. No other opinion on if the change is good or not.

kocio-pl:
- Alpine would mean that we would have to rewrite Dockerfiles and test with different packages. 
- Debian skipped libicu55 completely, so there's no Debian version that would work for us. Maybe there are some unofficial packages we could use or the source of ICU or Mapnik could be compiled, but this is where we can introduce quite new problems (and size, because compiling tools take some space).
- I don't know about the image size, but Ubuntu means easy replacement out of the box, with no need to search/compile packages or rewriting Dockerfiles.

It's up to us to decide if we want to fix it in a different way, but this is what i would do to minimize the burden and the risk of problems.

kocio-pl:
I'm not sure why `Dockerfile.import` shows a difference at the end of the file, because I have not added new line there.

pnorman:
> I'm not sure why Dockerfile.import shows a difference at the end of the file, because I have not added new line there.

It was probably missing an end of line on the last line before, which is a bug

math1985:
Works for me (but the current version seems to work too).

PR #2820, opened 2017-09-10
Do not order low zoom roads by layer

math1985:
On low zoom levels (up to z9) it is not important which road crosses
on top of the other. Instead, it is important to always show important
roads on top, so lower classes roads do not hide motorways.

### DIFF
--- a/project.mml
+++ b/project.mml
@@ -936,7 +936,6 @@ Layer:
             OR (railway IS NOT NULL AND railway != 'preserved'
               AND (service IS NULL OR service NOT IN ('spur', 'siding', 'yard')))
           ORDER BY
-            COALESCE(layer,0),
             z_order
         ) AS roads_low_zoom
     properties:


math1985:
Rendering differences are typically minimal, except in particular circumstances such as [this one](https://www.openstreetmap.org/?mlat=49.6632&mlon=6.1683#map=13/49.6632/6.1682).

Before:
screen shot 2017-09-11 at 00 34 53

After:
screen shot 2017-09-11 at 00 35 53

sommerluk:
I support this. On this zoom levels, z_order is indeed more meaningfull than layer.

kocio-pl:
I don't see the problem on the attached renderings (probably scaling would be good here to show the problem fully), but I agree with layering being not adequate on lower zoom.

math1985:
> I don't see the problem on the attached renderings

Best to look at the location of the marker in the linked map, and then to look again in the attached renderings how that part is rendered.

math1985:
Scaled images:
screen shot 2017-09-11 at 01 47 42
screen shot 2017-09-11 at 01 47 51

PR #2821, opened 2017-09-10
Bring lightness of farmland color again in line with other landuse

matthijsmelissen:
Now that we have fading of landuse on z8-z12, I think we can revert
the fading of farmland (#1701). The reason for fading farmland was
mainly prominence of farmland on the mid-zoom levels, which is not
a problem anymore after the mid-zoom fading.

This brings the lightness of farmland again in line of the lightness
of similar landuses like grassland.

### DIFF
--- a/landcover.mss
+++ b/landcover.mss
@@ -19,8 +19,8 @@
 @commercial-line: #d1b2b0;  // Lch(75,12,25)
 @industrial: #ebdbe8;       // Lch(89,9,330) (Also used for railway)
 @industrial-line: #c6b3c3;  // Lch(75,11,330) (Also used for railway-line)
-@farmland: #fbecd7;         // Lch(94,12,80)
-@farmland-line: #d6c4ab;    // Lch(80,15,80)
+@farmland: #edddc9;         // Lch(89,12,80)
+@farmland-line: #c8b69e;    // Lch(75,15,80)
 @farmyard: #f5dcba;         // Lch(89,20,80)
 @farmyard-line: #d1b48c;    // Lch(75,25,80)
 


math1985:
Examples:
screen shot 2017-09-11 at 00 53 11
screen shot 2017-09-11 at 00 53 25
screen shot 2017-09-11 at 00 53 54
screen shot 2017-09-11 at 00 54 04

kocio-pl:
What I like about this revert - brown is a more farmland-related color than current "glowing peach" and the rule is clear.
What I don't like about this - this is too dark and plays worse with grass and water, which are quite common on farmlands:

z15
Before
t8bcfn4t
After
uqnge0q

kocio-pl:
As you can see, it also plays worse with residential areas, which are also quite common, and I've seen that it's not too good with allotments (which are not that common there). I would try to keep the brown, but make it lighter than it was before.

Brown plays better with natural=sand, but I still think adding dot pattern to sand (https://github.com/gravitystorm/openstreetmap-carto/pull/2746#issuecomment-326978224) would be still very good idea for sand area:
zedicml2

kocio-pl:
See also https://github.com/gravitystorm/openstreetmap-carto/issues/780#issuecomment-328339635 for proposed natural=heath color change.

SomeoneElseOSM:
The brownish colour's definitely better than the current one.

For info I went with #F4EBDF here (which was just making the existing colour "less bright" ) in
https://github.com/SomeoneElseOSM/openstreetmap-carto-AJT/blob/master/landcover.mss#L29 .

math1985:
Will see what I can do.

> For info I went with #F4EBDF here

Is your rendering live somewhere?

kocio-pl:
Only for British Islands:

https://map.atownsend.org.uk/maps/map/map.html#zoom=17&lat=-24.987206&lon=135.08483

matkoniecz:
>  For info I went with #F4EBDF here

It would be necessary to test it, but it looks really well.

kocio-pl:
Couple of examples on different levels in Poland:

z8
Before
zr sqop6
After
9zzpi8so

z9
Before
mjthtdir
After
6qbqie8k

z10
Before
nalg znw
After
utxgbge3

z11
Before
msyiuz7v
After
jmb1ke1b

z12
Before
loce57z9
After
si_dfqlo

z13
Before
v_efpemg
After
4dmkcsfr

z16 (with border line)
Before
pfxi2fkn
After
ms9mmrbv

@rrzefox Could you test [this change](https://github.com/gravitystorm/openstreetmap-carto/compare/master...kocio-pl:farmland-ajt) on your server?

math1985:
> Couple of examples on different levels in Poland:

@kocio-pl Is this based on @SomeoneElseOSM's proposal or the original PR?

kocio-pl:
It's exactly like in AJT style (area+border). What do you think about it?

math1985:
It looks acceptable to me.

It would be a step away from keeping lightness equal for similar features.

kocio-pl:
I'm still thinking about making this color a bit darker, but it needs some time to test few different versions, however I like it as a basic shade (less yellow than currently).

Rendering in https://github.com/gravitystorm/openstreetmap-carto/pull/2821#issuecomment-328387177 proved me that it should be lighter than some other landuses anyway.

rrzefox:
> @rrzefox Could you test this change on your server?

This has been deployed and all zoom levels rerendered.

kocio-pl:
> I'm still thinking about making this color a bit darker

Testing proved me that AJT style is really too light, so I'd like to find a darker shade.

kocio-pl:
Another idea would be to use this color (it's very good on high zoom levels), but just exclude it from fading, because it's light enough and close to the ground color.

kocio-pl:
See also #1991.

Tomasz-W:
@matthijsmelissen Please add "resolves" line for https://github.com/gravitystorm/openstreetmap-carto/issues/1991 in first post.

PR #2823, opened 2017-09-11
Removing z-ordering in landcover

### LINKED ISSUE #53: Should landcover be ordered by z_order?
At the moment the query orders by both z_order and way_area. Ordering by area makes sure small polygons end up on top (where there's no holes in a larger polygon). I guess z_order will help if the features are on separate layers, since there's no fine control over landcover ordering required.

But is this necessary? Are overlapping features with different z_orders - and the larger one on top - either common or wanted?

kocio-pl:
Resolves #53.
Follow up to #2740.

A basic testing was made to make sure rendering is not just broken after this.

### DIFF
--- a/project.mml
+++ b/project.mml
@@ -160,7 +160,7 @@ Layer:
               OR highway IN ('services', 'rest_area')
               OR railway = 'station')
               AND way_area > 0.01*!pixel_width!::real*!pixel_height!::real
-            ORDER BY COALESCE(layer,0), way_area DESC
+            ORDER BY way_area DESC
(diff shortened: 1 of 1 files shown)


math1985:
In fact, this PR resolves our oldest open issue!

kocio-pl:
Part of my effort is resolving old issues or closing them (because they are already solved or nobody is interested now). I'd like to make issue tickets a tool for addressing problems, not a haystack with every possible idea or wish being open (but they can be reopened when needed).

PR #2824, opened 2017-09-11
Gastronomy POI category

kocio-pl:
Follow up to #2765 and https://github.com/gravitystorm/openstreetmap-carto/pull/2786#pullrequestreview-61696327.

This PR is **just a code change** (new variable is created with the same color as before), but it makes it easier to make gastronomy POIs look different. 

We are currently using amenity color for pretty much everything:
- small amenities (waste basket, picnic table) 
- public offices (police station, town hall)
- culture and entertainment (library, cinema)
- gastronomy (restaurant, bar)
- memorials and artworks

We don't have to use different colors for all of them, but creating such categories would make POIs more customizable.

### DIFF
--- a/amenity-points.mss
+++ b/amenity-points.mss
@@ -9,6 +9,7 @@
 @airtransport: #8461C4; //also ferry_terminal
 @health-color: #da0092;
 @amenity-brown: #734a08;
+@gastronomy: @amenity-brown;
 @man-made-icon: #555;
 @landform-color: #d08f55;
 
@@ -61,7 +62,7 @@
 
   [feature = 'amenity_bar'][zoom >= 17] {
     marker-file: url('symbols/bar.svg');
-    marker-fill: @amenity-brown;
+    marker-fill: @gastronomy;
     marker-placement: interior;
     marker-clip: false;
   }
@@ -119,7 +120,7 @@
 
   [feature = 'amenity_cafe'][zoom >= 17] {
     marker-file: url('symbols/cafe.svg');
-    marker-fill: @amenity-brown;
+    marker-fill: @gastronomy;
     marker-placement: interior;
     marker-clip: false;
   }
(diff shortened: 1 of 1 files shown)


amenity-points.mss
```diff
@@ -9,6 +9,7 @@
 @airtransport: #8461C4; //also ferry_terminal
 @health-color: #da0092;
 @amenity-brown: #734a08;
+@gastronomy: @amenity-brown;
```
pnorman: Could we just call it `@food`?

kocio-pl: This was my first idea, but it's also about drinking (bar or pub).

pnorman: I'd still rather the shorter and less precise name

kocio-pl: I like the more precise name, but that's not the most important thing here. What others think about it?

HolgerJeromin: gastronomy has less confusion potential

math1985: +1 for gastronomy

kocio-pl:
Preview how would brown+orange (#c64803) look like with gastronomy POIs (but maybe culture and entertainment could be in the same color):

kb5rtzr0

kocio-pl:
So maybe we could define other groups as well (there can always be some overlaps between categories or other edge cases). If we agree, I can just extend current PR.

public_service:
- town hall
- embassy
- post office
- bank
- courthouse
- police
- social facility
- fire station
- prison

culture (with entertainment):
- museum
- cinema
- theatre
- library
- art centre
- community centre
- archaeological site

memorials (with artworks)
- monument
- obelisk
- memorial
- memorial plaque
- artwork

Maybe small amenities don't need special category, since there would be much less objects in general "amenity-brown" category.

pnorman:
I've merged it, and we can do another PR to add more categories for colour variables. I think we should keep any new colours out of these PRs and do cartographic changes later.

Issue #1890, opened 2015-10-03
add line rendering for benches 


HolgerJeromin:
As reported in #1824

Some benches are tagged as lines 
http://www.openstreetmap.org/way/313487068
http://www.openstreetmap.org/way/79639567
http://www.openstreetmap.org/way/79639568

pnorman:
I can't think of any way to render them except for sticking the bench symbol somewhere on the linestring.

Surprisingly, there are about 3k benches mapped as lines.

HolgerJeromin:
you are right, this is a problem. But i think the symbol in the middle is a good enough solution.

jojo4u:
Tagging as way  is allowed in wiki since 2011 and makes sense for long ones. Symbol in middle sounds reasonable.

http://wiki.openstreetmap.org/w/index.php?title=Tag:amenity%3Dbench&oldid=705902

NonnEmilia:
Now the benches mapped as line are 7k (according to [taginfo](https://taginfo.openstreetmap.org/tags/?key=amenity&value=bench)).

Tomasz-W:
I would like to see some examples of bench ways rendered as brown lines

kocio-pl:
That would probably look like tracks, isn't it?

pnorman:
I don't see how it would be obvious to someone that a brown line means a bench line, particular given the confusion with existing brown lines.

Tomasz-W:
Well, if it would be thick as barrier=city_wall (eg. http://www.openstreetmap.org/way/363795561) or thicker, and dark-brown, it propably wouldn't be that bad and wouldn't confused with tracks. Notice that a lot of things on map are obvious from the context. Like always, a rendering preview is the only way to check how it works...

Tomasz-W:
@kocio-pl Please close this ticket as a duplicate of https://github.com/gravitystorm/openstreetmap-carto/issues/2711 meta ticket. Move further discussion there.

PR #2825, opened 2017-09-11
Adding support for slipway ways

### LINKED ISSUE #484: Add support for slipway ways

kocio-pl:
Resolves #484.

I'm not 100% sure about rendering names for ways, it's safer to skip it, since it's not clear if name label belongs to a slipway or a highway.

nhswhgef

### DIFF
--- a/amenity-points.mss
+++ b/amenity-points.mss
@@ -902,6 +902,7 @@
     marker-clip: false;
   }
 
+  // Slipway tagging on points - slipway on lines is defined later 
   [feature = 'leisure_slipway'][zoom >= 17] {
     marker-file: url('symbols/transport_slipway.p.20.svg');
     marker-fill: @transportation-icon;
@@ -2196,6 +2197,11 @@
   // Ford tagging on ways
   [feature = 'highway_ford'][zoom >= 16] {
     marker-file: url('symbols/ford.svg');
+    marker-fill: @transportation-icon;
+  }
+  // Slipway tagging on ways
+  [feature = 'leisure_slipway'][zoom >= 17] {
+    marker-file: url('symbols/transport_slipway.p.20.svg');
     marker-fill: @transportation-icon;
   }
 }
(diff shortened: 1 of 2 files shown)


kocio-pl:
Are there any other linear amenities which should be shown on osm-carto?

daganzdaanda:
Is this the same as highway=service? Maybe use the narrower styling like in service=parking_aisle? Slipways are normally not very long, and should not look too important IMHO.

> Are there any other linear amenities which should be shown on osm-carto?

#1890 is about benches

kocio-pl:
Thanks for both suggestions! However our road code is spaghetti-esque and I'm still not able to find out where should I add another type of service road... Could someone help me find where it's selected?

kocio-pl:
OK, it works, but it's better if someone would check the SQL for casing and filling:

06jc vrz

kocio-pl:
> #1890 is about benches

I can think also about bus stops and I even think I saw it lately among other osm-carto issues, but forgot where it was mentioned.

sommerluk:
I suppose that slipways have no own z_order value, and this could be a problem when sorting?

kocio-pl:
I don't know, but it's hard for me to imagine that such specialized service road are crossing with other road on a different level.

kocio-pl:
For me it's safe to be merged. Any comments or reviews?

pnorman:
> I don't know, but it's hard for me to imagine that such specialized service road is crossing with other road on a different level.

z_order isn't about different layers, it's within the same layer.

Are we requiring highway=service to render the white line?

kocio-pl:
I meant physical levels, not our internal level/layer.

For me this is a special service road (which ends up in water), but still just a service road, so I expect no rendering changes. Do you have some other suggestion?

kocio-pl:
@pnorman Do you still have some questions about this? I would merge it soon.

PR #2826, opened 2017-09-11
Memorials POI category

kocio-pl:
Follow up to #2824.

I will push groups one by one because features are clustered so it's easier to break something while making big cleaning. Basic rendering test shows that this group looks like before and changing color works for included elements too.

Memorials (with artworks)
- monument
- obelisk
- memorial
- memorial plaque
- artwork

I have moved archaeological site to another cluster, so it can be included in a "culture" group.

### DIFF
--- a/amenity-points.mss
+++ b/amenity-points.mss
@@ -10,6 +10,7 @@
 @health-color: #da0092;
 @amenity-brown: #734a08;
 @gastronomy: @amenity-brown;
+@memorials: @amenity-brown;
 @man-made-icon: #555;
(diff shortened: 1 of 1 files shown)


kocio-pl:
Hm, I'm not sure where archaeological site belongs to (memorials or culture), what do you think?

daganzdaanda:
An archaeological site is where some things have been found in the ground. This is not a memorial, so culture fits better. 
There may be memorials / plaques etc at an archealogical site, but they will get their own nodes.

PR #2827, opened 2017-09-11
Leisure POI category

kocio-pl:
Follow up to #2824.

This is a bit more complex PR, because it's not only about making another POI color group sharing the same variable name, but it's also icon/label unification, fixing part of the problem described in #1355, #1794 and https://github.com/gravitystorm/openstreetmap-carto/pull/2268#issuecomment-237856186.

While amenity brown is overused, leisure-green (dark leisure-related color) is underused currently. It's used for many labels and only for some icons - namely both golf-related ones. I have changed the color of playground, water park and dog park icons to match their label color (leisure-green). I have not tried to match zoom levels for labels and icons nor was I touching other aspects of text.

Here's the result:

Before
j3rpg8cz
After
jjedkrhr

Before
5fompwzs
After
xv4wizp0

Before
ls6ekttu
After
wbnhknzu

### DIFF
--- a/amenity-points.mss
+++ b/amenity-points.mss
@@ -12,6 +12,7 @@
 @gastronomy: @amenity-brown;
 @man-made-icon: #555;
 @landform-color: #d08f55;
+@leisure-green: darken(@park, 60%);
 
 @landcover-font-size: 10;
 @landcover-wrap-width-size: 30; // 3 em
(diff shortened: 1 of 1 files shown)


kocio-pl:
Update with a bigger water park icon:

Before
582_jtek
After
i9inmlbq

PR #2829, opened 2017-09-12
Water park icon size fix

kocio-pl:
Related to #2827.

The water park icon is nominally 14 px big, but in reality that's only canvas and the image itself is only 12 px wide, so it's hard to recognize. Scaling up to 14 px did not work too much, so I've scaled it up to 16 px and cropped the wave. It's not pixel aligned, because I'm not sure how to do it with current design, but it's at least more readable now.

Before
h9rdwlqu
After
uzctlvj1


pnorman:
If we create a pixel aligned wave we can share it in common with this and ford.svg

PR #2832, opened 2017-09-14
Drop halo from ferry labels

math1985:
This makes the labels look more similar in color to the lines they belong.

Before:
screen shot 2017-09-14 at 21 04 01

After:
screen shot 2017-09-14 at 21 00 59

With the new water color, contrast should be strong enough anyway.

### DIFF
--- a/ferry-routes.mss
+++ b/ferry-routes.mss
@@ -25,8 +25,6 @@
     text-fill: @ferry-route-text;
     text-spacing: 1000;
     text-size: 10;
-    text-halo-radius: @standard-halo-radius;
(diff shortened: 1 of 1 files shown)


kocio-pl:
In fact it's so great solution of an obscure problem, that I would remove halo from all the water labels (which of course needs testing as a worldwide change)!

This is quite busy [labels location](http://www.openstreetmap.org/#map=16/54.4055/18.6703), but it helps us to compare the water related colors (with a quite typical admin borders going along the waterways):

Before
q zwquhd
After
xr56yawa

Issue #2199, opened 2016-06-25
do not display borders between national parks on lower zoom levels


matkoniecz:
In some regions there are multiple joining national parks. Western USA is the most impressive, with many protected areas covering significant area with byzantine shapes and topology.

For example in this area where national park borders seems to be drawn by Elbridge Gerry everything is anyway inside a protected area.
![selection_007](https://cloud.githubusercontent.com/assets/899988/16359173/a4d1e90c-3b2a-11e6-8a14-15e0f51fea7d.png)

As result both on high and low zoom levels is quite bad.
![selection_009](https://cloud.githubusercontent.com/assets/899988/16359184/09e13f1e-3b2b-11e6-8d23-132bdcf95cdb.png)

This is not limited to the USA. For example - http://www.openstreetmap.org/#map=7/48.247/24.587 is area where Slovenia, Ukraine and Poland have five joining protected areas.

![selection_010](https://cloud.githubusercontent.com/assets/899988/16359197/babecbb2-3b2b-11e6-965e-c6a2c2669537.png)

Note that many joining national parks are curently badly mapped as not joined, so there are some rendering artefacts caused by poor tagging. This effect is welcome rather than a problem but it makes testing harder.

http://www.openstreetmap.org/#map=17/52.57878/14.63135

![selection_011](https://cloud.githubusercontent.com/assets/899988/16359212/40a543b4-3b2c-11e6-81f4-0b71296c8e92.png)

On higher zoom levels it makes sense to display borders also between joining protected areas, it is open question what should be choosen as threshold (currently in proof of concept all joining borders are eliminated).

I developed it as part of a more complex change, but it stands well also on its own.

I submit it as a issue, not PR as I have only a proof of concept. SQL query is partially a massive HACK/TODO and there are some suspicious rendering articacts (I submitting it anyway, as after days of battling with pg_hba.conf, qgis etc it finally works)).

My code used to generate before/after below is at https://github.com/matkoniecz/openstreetmap-carto/commit/3f05be36ace1605b7dc5f53662e4dbfc1ce8d8d5 (updated code is at https://github.com/gravitystorm/openstreetmap-carto/compare/master...matkoniecz:dissolve_national_parks_proof_of_concept?expand=1)

I will now look for way to correctly compute way_pixels (in case that it is obvious for somebody - comments/hints are welcomed).

In regions like UK with sensible structure of protected areas it should have no noticeable change.

[...]

Issue #1935, opened 2015-10-31
new road colours lead to roads dominating z5


nebulon42:
With the new colours roads are far too emphasised on z5. Example USA:

![us_z5](https://cloud.githubusercontent.com/assets/3531092/10864563/2e2b74b0-7ff2-11e5-9d5a-17c33cebb040.png)

Either the lines are too thick or the colours are too strong.

sommerluk:
Might it be useful to render only motorways at this zoom level?

matkoniecz:
> Might it be useful to render only motorways at this zoom level?

No, see for example http://www.openstreetmap.org/#map=5/56.438/95.669 http://www.openstreetmap.org/#map=5/-18.438/33.530 http://www.openstreetmap.org/#map=5/-12.469/-68.599

In general - main problem is extremely different road density - compare for example South and North America. Even now in some regions are not prominent enough.

math1985:
Or even more ambitious - why render roads at this zoom level at all? It is not suitable even for long-distance planning.

dieterdreist:
2015-11-02 15:41 GMT+01:00 math1985 <notifications@github.com>:

> Or even more ambitious - why render roads at this zoom level at all? It is
> not suitable even for long-distance planning.



it might be, if roads were discernible ;-)

pnorman:
I think part of the road dominating problem is that there's nothing else on that zoom.

But when at I look at it, I find it better for travel planning in the northwest quarter of the US than before. On the east coast, it's not very useful, and it gives you the impression that there's multiple highways everywhere. But this what it's like on the ground, and I don't think it's possible to do useful route planning there at that zoom.

I had a look at z5 at a few maps (click for full scale, Geofabrik is approximately the old osm-carto road style)

![image](https://cloud.githubusercontent.com/assets/1190866/10893197/5dddda84-815c-11e5-947c-cc2133df7776.png)

I would say that the current style or HERE Map are the most useful for route planning, HERE having an advantage with more cities. The low motorway/water, motorway/boundary, and trunk/anything contrast make the old colours less useful.

Bing makes national parks really obvious, Google is similar though not as extreme. Neither of them make it clear that they're national parks and not forests. Google is probably worse here. 

HERE makes forests obvious. MQ Open tries to, but hillshading and terrain at this zoom when combined with JPEG artifacts doesn't work well.

Of the six, I'd probably put HERE on top of route planning at this zoom.

[...]

PR #2834, opened 2017-09-14
Make regional admin borders solid instead of dashed

math1985:
Due to the coastline paradox, dashed lines for borders don't work
well on low-zoom: strongly curved borders appear solid instead of
dashed. Also, solid dashes generally are too busy for the crowded
z5-7.

This PR makes regional boundaries (admin_level=4) solid instead of
dashed up to z9.

To prevent the borders from getting too strong, they are also made
thinner.

Before/after:

z4:
screen shot 2017-09-14 at 23 40 26 screen shot 2017-09-14 at 23 40 03

z5:
screen shot 2017-09-14 at 23 40 57 screen shot 2017-09-14 at 23 41 18

z6:
screen shot 2017-09-14 at 23 41 41 screen shot 2017-09-14 at 23 41 31

z7:
screen shot 2017-09-14 at 23 42 19 screen shot 2017-09-14 at 23 42 10

z8:
screen shot 2017-09-14 at 23 42 46 screen shot 2017-09-14 at 23 42 39

z9:
screen shot 2017-09-14 at 23 43 00 screen shot 2017-09-14 at 23 42 54

### DIFF
--- a/admin.mss
+++ b/admin.mss
@@ -65,19 +65,19 @@ overlapping borders correctly.
       line-color: @admin-boundaries;
       line-join: bevel;
       line-width: 0.4;
-      line-dasharray: 4,3;
       line-clip: false;
     }
     [zoom >= 5] {
+      background/line-width: 0.5;
+      line-width: 0.5;
+    }
+    [zoom >= 6] {
       background/line-width: 0.6;
       line-width: 0.6;
     }
-    [zoom >= 6] {
-      background/line-width: 0.8;
-      line-width: 0.8;
-    }
     [zoom >= 7] {
       background/line-width: 1;
+      line-dasharray: 4,3;
       line-width: 1;
     }
     [zoom >= 9] {


kocio-pl:
Thanks, it's a really useful change! 

Maybe this makes it more solid and violet+red is a bit more dominating on low zoom (see  #1935), but low zoom changes are still ahead of us (#2688) and we can try with borders color change (like #2695) and at least it stops looking like artefacts (another kind of high frequency noise to remove - see also #2199):

Poland, z6
Before
uaz5sd2i
After
bsinrnem

On the other hand I'm not sure to which zoom level it should be used. You were testing in Africa, but in Europe we have more linear entities (like roads, railways and rivers), so maybe z7-z9 should be left untouched to be more distinct.

Poland, z9
Before
tlgbt2u4
After
vs0hmqph

math1985:
I agree with @kocio-pl. I restricted the PR to zoom level 4-6. Merged.

PR #2835, opened 2017-09-14
Hide tiny zoos and theme parks

math1985:
This starts showing zoos and theme parks from one zoomlevel before
they are large enough to get a name.

Example of theme parks of the size that will disappear now:

screen shot 2017-09-15 at 00 50 52
screen shot 2017-09-15 at 00 51 22

### DIFF
--- a/landcover.mss
+++ b/landcover.mss
@@ -755,8 +755,10 @@
 }
 
 #tourism-boundary {
-  [tourism = 'zoo'][zoom >= 10][way_pixels >= 20],
-  [tourism = 'theme_park'][zoom >= 10][way_pixels >= 20] {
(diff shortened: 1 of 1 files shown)


math1985:
The different treshold remains: the area is rendered one zoomlevel earpier.

kocio-pl:
> The different treshold remains: the area is rendered one zoomlevel earpier.

Does it? The code is just more strict on z10-16 and relaxed on z17+, but it was z10 before and you didn't switch to z9 instead. Did I understand you right?

math1985:
I mean that on zoom leveps z10-16, if you zoom in the area will appear one zoom level before the text appears.

math1985:
Thanks for quick review!

PR #2836, opened 2017-09-14
Reinstate link-roads on z10

math1985:
Unfortunately it seems like we do need link-roads on z10, for
junctions like
https://www.openstreetmap.org/#map=10/51.8748/4.3918 .


kocio-pl:
Related to #2758.

pnorman:
Doesn't seem to be working for me 
image

pnorman:
Checking with v4.1.0 and v4.2.0, we've never done link roads at z10 that I can see. Probably because it's coming from the lz layer.

PR #2839, opened 2017-09-16
Rendering power=plant outline

kocio-pl:
Follow up to #2700.

[Umspannwerk Dieringhausen](https://www.openstreetmap.org/way/53243836), z19

Before
qd357zog
After
x6xa0sx7

### DIFF
--- a/landcover.mss
+++ b/landcover.mss
@@ -371,7 +371,8 @@
     [way_pixels >= 64] { polygon-gamma: 0.3;  }
   }
 
-  [feature = 'man_made_works'][zoom >= 16] {
+  [feature = 'man_made_works'][zoom >= 16],
+  [feature = 'power_plant'][zoom >= 16] {
     line-width: .5;
     line-color: @industrial-line;
     [name != ''] {
--- a/project.mml
+++ b/project.mml
@@ -142,7 +142,7 @@ Layer:
               ('military_' || (CASE WHEN military IN ('danger_area') THEN military ELSE NULL END)) AS military,
               ('natural_' || (CASE WHEN "natural" IN ('beach', 'shoal', 'heath', 'grassland', 'wood', 'sand', 'scree', 'shingle', 'bare_rock', 'scrub') THEN "natural" ELSE NULL END)) AS "natural",
               ('wetland_' || (CASE WHEN "natural" IN ('wetland', 'marsh', 'mud') THEN (CASE WHEN "natural" IN ('marsh', 'mud') THEN "natural" ELSE tags->'wetland' END) ELSE NULL END)) AS wetland,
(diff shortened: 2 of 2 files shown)


kocio-pl:
Any other such objects with invisible areas which should be visible?

HolgerJeromin:
plant was the first and only object i remember which should be inside industrial area. 
hotel/museum outline are a special case which should be handled separate.

pnorman:
I'm not sure what this outline accomplishes.

kocio-pl:
It shows the area without the landuse, indicating that it should be added.

math1985:
What does it look like on an industrial landuse background? Also in that case it might be nice to have the outline of the actual plant.

kocio-pl:
> What does it look like on an industrial landuse background?

It's the same code as in #2700, so look at https://github.com/gravitystorm/openstreetmap-carto/pull/2700#issuecomment-329018533.

> Also in that case it might be nice to have the outline of the actual plant.

What do you mean?

math1985:
> What do you mean?

If we have power=plant within landuse=industrial, it would be good if the power=plant outline is still visible, but I believe your PR accomplishes that.

I would also be fine (and maybe even prefer) with giving power=plant a purple background like landuse=industrial.

kocio-pl:
What would be the reason for rendering (or not rendering) purple background? I don't have an opinion now.

pnorman:
> It shows the area without the landuse, indicating that it should be added.

If the purpose is to reveal this as an error, that's something for a debugging layer, not a general purpose layer. If we want to render power plants like industrial areas we could do that.

kocio-pl:
Well, for the man_made=works it was to show both - that's incomplete tagging and showing that it's not a node, but a bigger object. 

It all depends on how we see the problem of area objects and landuse - if I understand correctly (I'm not sure if I really do):
- @imagico desn't like such object to resemble a node and requests something more
- @dieterdreist doesn't like filling
- you don't like just the outline
 
So:
1. Which solution would be something everybody (or just majority) could accept?
2. Is man_made=works and power=plant the same case or a different one?

mboeringa:
As to fill for **power=plant**, be aware that there are also dispersed power plants, tagged as one entity. E.g. groups of windmills at sea, or solar panels, may be tagged with one outline and **power=plant**.

See the wiki also:
http://wiki.openstreetmap.org/wiki/Tag:power%3Dplant

A fill is not appropriate in this case.

daganzdaanda:
> @imagico desn't like such object to resemble a node and requests something more

Actually, he wrote he doesn't like to have the label _styled based on way_area_ if there is nothing that shows the extent of the area. It's a subtle difference, but it does make sense.

kocio-pl:
I will put this information also here: If nobody help me fix this code to the acceptable form, I plan to close it without merging in a few days.

PR #2841, opened 2017-09-17
Move aeroways to their own layer

pnorman:
More progress on #2046. The cartographic changes are minimal and involve some strange real-world situations.

This is done as part of road layer simplification, but also fixes a few layering bugs showing up where there are aeroway bridges, and some casing width problems with invisible bridges.

There are no changes to the typical airport

image
(left v4.2.0 on osmf servers, right this PR)

The weirder situations are bridges
image
17/47.92228/-122.27635

image
16/33.6224/-84.4291

Where there is a road crossing a runway (e.g. Gibralter) the runway will remain rendered on top of the road.

There is one case of a runway tagged as a tunnel in OSM where the runway goes through a mountain [in North Korea](http://www.openstreetmap.org/#map=14/39.0937/127.4091), and that case will remain unhandled and rendered like a normal runway.

The current rendering has bugs with bridge casing widths for aeroways, and these are fixed by using the same road casing constructions as highways.

The code changes are worth explaining, because they're the real reason for this PR

- Aeroways are a special case of roads where we assume they're always on top of any roads (layer-wise). This assumption combined with the z_order rendering them on top means they can be their own layer. Most styles wouldn't bother with aeroway bridges

- Fixing the casing width bugs brought up the MSS code length, otherwise it would be shorter. Because a lot of the new code is repetitive line-width stuff (e.g. `[zoom >= 16] { line-width: 24 + 2*@major-casing-width-z16; }`) I find it simpler

- Using an attachment is a technique I [mentioned in 2016](https://github.com/gravitystorm/openstreetmap-carto/issues/2046#issuecomment-205492887) and is a better solution than duplicate layers for ground level and bridges.

tl;dr: Moves aeroways to their own layers, modernizes aeroway code, fixes aeroway bridge bugs.

### DIFF
--- a/project.mml
+++ b/project.mml
@@ -440,9 +440,8 @@ Layer:
     <<: *extents
     Datasource:
       <<: *osm2pgsql
-      # This query is quite large, having to deal with both roads, railways and
-      # runways. To allow for ways that are both railways and roads, a UNION
-      # ALL is present.
+      # This query is quite large, having to deal with both roads, railways. To
+      # allow for ways that are both railways and roads, a UNION ALL is present.
       table: |-
         (SELECT
             way,
@@ -457,7 +456,7 @@ Layer:
             service,
             link,
             layernotnull
-          FROM ( -- subselect that contains both roads and rail/aero
+          FROM ( -- subselect that contains both roads and rail
             SELECT
                 way,
                 ('highway_' || highway) AS feature, --only motorway to tertiary links are accepted later on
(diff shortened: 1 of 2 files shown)


pnorman:
Just to be clear, I think moving aeroways into the roads query was a mistake at the time, but none of us realized it.

math1985:
By moving aeroways to their own layer, we are removing the functionality where aeroways can be drawn below roads (for example for aeroways in tunnels, or for aeroways with a bridge over them). I'm not aware of any real-life example of this (apart from the North-Korea example), so unless somebody provides counterevidence, I think that's a safe thing to do.

The opposite, roads on top of aeroways, occurs quite frequently but this we can still handle.

math1985:
Ok, apparently Phoenix has a sky train crossing a taxiway: https://osm.org/go/TlVi8S3el--?m= (pictures at http://photos.wikimapia.org/p/00/01/67/24/78_big.jpg). The current PR will break layering there. I have no strong opinion on whether we want to accept that or not.

kocio-pl:
It won't be too visible because both colors are dark, so I would accept that if there are more general benefits.

pnorman:
> Ok, apparently Phoenix has a sky train crossing a taxiway: https://osm.org/go/TlVi8S3el--?m= (pictures at http://photos.wikimapia.org/p/00/01/67/24/78_big.jpg). The current PR will break layering there. I have no strong opinion on whether we want to accept that or not.

I'm fine with breaking that layering. Looking at the other layers on osm.org, they all either get that wrong or the other crossing at https://www.openstreetmap.org/?mlat=33.43478&mlon=-112.00423#map=18/33.43478/-112.00423 where the order is reversed.

In fact, if you look at [bbbike map compare](https://mc.bbbike.org/mc/?lon=-111.993273&lat=33.435218&zoom=16&num=8&mt0=sputnik-map&mt1=mapbox-transportation&mt2=hike_bike&mt3=carto-basemaps&mt4=esri-topo&mt5=mapquest-map&mt6=lyrk&mt7=mapbox-terrain&marker=) you can see a selection of 8 maps, all of which get it wrong in some way or another.

pnorman:
Whoops - I thought this had an accepted review. Since we just did a release and none of the comments were objections, I'm going to leave it committed unless someone particularly wants to revert it.

math1985:
I think this change should be reviewed by somebody first. I don't think snybody had time to look at the PR on a code level. The roads code is quite complex and changes might easily have unexpected side-effects.

math1985:
Ok, fine for now.

PR #2845, opened 2017-09-17
Improving water labels [WIP]

kocio-pl:
Follow up to #2832.

Just dropping halo from all the water labels to make them more readable.

[Testing place](http://www.openstreetmap.org/#map=16/54.4055/18.6703) with different water labels (ferry lines halo is already dropped):

Before
q zwquhd
After
xr56yawa

@rrzefox Could you deploy a current code with this patch on your server? This would be a massive change for a water features, so it needs worldwide testing.

### DIFF
--- a/water.mss
+++ b/water.mss
@@ -244,8 +244,6 @@
       text-fill: @water-text;
       text-spacing: 400;
       text-size: 10;
-      text-halo-radius: @standard-halo-radius;
-      text-halo-fill: @standard-halo-fill;
       [zoom >= 14] { text-size: 12; }
       [int_tunnel = 'yes'] { text-min-distance: 200; }
     }
@@ -253,8 +251,6 @@
     [waterway = 'canal'][zoom >= 13][zoom < 14] {
       text-name: "[name]";
       text-face-name: @oblique-fonts;
-      text-halo-radius: @standard-halo-radius;
-      text-halo-fill: @standard-halo-fill;
       text-size: 10;
       text-placement: line;
       text-fill: @water-text;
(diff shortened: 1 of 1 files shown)


math1985:
To me this does not improve legibility.

rrzefox:
I've deployed this now (usual warning about not all tiles having been rerendered applies).
While I do think that it improves legibility on water, there is a caveat: The label is not necessarily rendered on water as a background. Consider small waterways, the label is usually rendered next to the water in whatever landuse there is there. Example:

waterlabel1 waterlabel2

However, I don't think it looks worse (at least in this example), so for now I'd consider this a (small) overall improvement.

kocio-pl:
Looks like something could be done here, because ferry label shows clearly that it's possible to make it better and general water labels are still not too visible.

Maybe the outline is OK (it's just 1 px wide, so probably it can't be declared smaller), but the water label color should be even darker? What do you think?

kocio-pl:
Both ferry and general water labels look better only on the water, but worse when rendered over some lines:

http://bl.ocks.org/math1985/raw/af7a602c222dbf1ff1a2c0d84ed755b7/#14/51.9734/4.0844

kocio-pl:
The problem exists only with small labels (mainly waterways) - big ones (water areas) look OK already:

7xwagjn9

kocio-pl:
I guess current solution is good enough for now, so I'm closing this ticket.

Both ferry and water labels look better on the clear water without the halo, but worse when there is something else in the background. Most of the time ferry lines don't have a backgound, so it's safe to leave them this way. Water labels are too popular and can be found in different places, where at least partial background exists (like stream in a forest - part of the label is rendered in the forest, because stream line is thin), so probably we need a better idea anyway.

PR #2852, opened 2017-09-18
Rendering natural=bare_rock earlier

kocio-pl:
The area of a bare rock can be substantial - and lack of overlay pattern doesn't help to recognize what it is.

[Area in Libya](http://www.openstreetmap.org/way/138017567)

z8
Before
hgoab9nm
After
nlzevzxd

z9
Before
vbapdvya
After
nub8piae

### DIFF
--- a/landcover.mss
+++ b/landcover.mss
@@ -440,14 +440,16 @@
     }
   }
 
-  [feature = 'natural_bare_rock'][zoom >= 9] {
+  [feature = 'natural_bare_rock'][zoom >= 8] {
     polygon-fill: @bare_ground;
-    [way_pixels >= 4]  { polygon-gamma: 0.75; }
-    [way_pixels >= 64] { polygon-gamma: 0.3;  }
-    [zoom >= 13] {
-      polygon-pattern-file: url('symbols/rock_overlay.png');
-      [way_pixels >= 4]  { polygon-pattern-gamma: 0.75; }
-      [way_pixels >= 64] { polygon-pattern-gamma: 0.3;  }
+    polygon-pattern-file: url('symbols/rock_overlay.png');
+    [way_pixels >= 4] {
+      polygon-gamma: 0.75;
+      polygon-pattern-gamma: 0.75;
+    }
+    [way_pixels >= 64] {
+      polygon-gamma: 0.3;
+      polygon-pattern-gamma: 0.3;
     }
   }
 
--- a/amenity-points.mss
+++ b/amenity-points.mss
@@ -2030,7 +2030,7 @@
 
   [feature = 'natural_scree'][zoom >= 9],
   [feature = 'natural_shingle'][zoom >= 9],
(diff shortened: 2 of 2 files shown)


landcover.mss
```diff
@@ -440,15 +440,11 @@
...
-  [feature = 'natural_bare_rock'][zoom >= 9] {
+  [feature = 'natural_bare_rock'][zoom >= 8] {
     polygon-fill: @bare_ground;
+    polygon-pattern-file: url('symbols/rock_overlay.png');
     [way_pixels >= 4]  { polygon-gamma: 0.75; }
     [way_pixels >= 64] { polygon-gamma: 0.3;  }
-    [zoom >= 13] {
-      polygon-pattern-file: url('symbols/rock_overlay.png');
-      [way_pixels >= 4]  { polygon-pattern-gamma: 0.75; }
-      [way_pixels >= 64] { polygon-pattern-gamma: 0.3;  }
```
math1985: We still need these two lines.

kocio-pl:
Fixed, thanks for spotting!

> I'm not sure if the size is a good reason to render it earlier. A small forest might be more interesting than a huge basalt plateau.

IMO general map is for explaining the space - if there is something big and characteristic, it's more interesting than just "a land" (whatever it is - a forest, a water reservoir, a basalt or a sand). 

We don't claim some continents are interesting/boring just because some things are more (or less) popular than on the other ones. We just show what is here.

dieterdreist:
sent from a phone

> On 19. Sep 2017, at 22:29, Matthijs Melissen <notifications@github.com> wrote:
> 
> I'm not sure if the size is a good reason to render it earlier. A small forest might be more interesting than a huge basalt plateau.



+1, think extremes like the Sahara desert: we'd display sand areas very early and the most interesting places (oasis) only after you'd have zoomed in a lot.
Size is working well for some things, e.g. buildings, but generally it is more important what something is, than how big it is. Barren land is more the absence of a feature (vegetation) than it is a feature itself.

kocio-pl:
This is an information what this area is and it tells a lot of this place. But if we have some rendering for oasis (a transportation-blue/amenity-brown palm icon?), we can just start showing it early too, because it's POI and it should not interfere with any other environment.

Think of a picnic table in the forest. Table is POI, but it's good to know it's located in the forest, not just "on a land".

dieterdreist:
sent from a phone

> On 20. Sep 2017, at 00:05, kocio-pl <notifications@github.com> wrote:
> 
> But if we have some rendering for oasis (a transportation-blue/amenity-brown palm icon?), we can just start showing it early too


AFAIK we don't even have a documented tag for them (there's a mention of natural=oasis in an abandoned desert proposal though. The tag has currently 109 uses).


Cheers,
Martin

kocio-pl:
The testing area has been retagged since then, but it seems to be useful area type (complementary to ice) on Antarctica, like here:

http://www.openstreetmap.org/#map=8/-77.696/161.381

PR #2854, opened 2017-09-19
Culture POI category

kocio-pl:
Follow up to #2824.

There are some strange text-dy's and some code could be clustered, but I leave it as it is now, because just categorizing makes enough code changes.

### DIFF
--- a/amenity-points.mss
+++ b/amenity-points.mss
@@ -11,6 +11,7 @@
 @amenity-brown: #734a08;
 @gastronomy: @amenity-brown;
 @memorials: @amenity-brown;
+@culture: @amenity-brown;
 @man-made-icon: #555;
 @landform-color: #d08f55;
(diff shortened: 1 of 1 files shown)


amenity-points.mss
```diff
@@ -1146,31 +1147,22 @@
...
   [feature = 'amenity_arts_centre'],
+  [feature = 'amenity_community_centre'],
   [feature = 'historic_archaeological_site'] {
     [zoom >= 17] {
       text-name: "[name]";
       text-size: @standard-font-size;
       text-wrap-width: @standard-wrap-width;
       text-line-spacing: @standard-line-spacing-size;
-      text-fill: @amenity-brown;
-      text-dy: 13;
```
math1985: Will this not break rendering for tall labels?

PR #2855, opened 2017-09-20
Public service POI category

kocio-pl:
Another follow up to #2824.

Basic testing has been made to make sure that labels are still here and they look the same as before.

### DIFF
--- a/amenity-points.mss
+++ b/amenity-points.mss
@@ -12,6 +12,7 @@
 @gastronomy: @amenity-brown;
 @memorials: @amenity-brown;
 @culture: @amenity-brown;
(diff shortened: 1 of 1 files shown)


HolgerJeromin:
Do we want further divide public (prison, police ...) and private (atm, bank ...) operators?

math1985:
The Uk privately run prisons and publicly owned banks :)

kocio-pl:
"Public" part is not about operators, it's about services. But of course the lines are blurry and we can rearrange in some other way.

kocio-pl:
Any comments or reviews now?

PR #2860, opened 2017-09-23
Add info about deployment

### LINKED ISSUE #2843: Release emails for releases not yet on tile.osm.org
Currently the release emails start with

> Today, $NEW_VERSION of the openstreetmap-carto stylesheet (the default stylesheet on openstreetmap.org) has been released.

or, if the sysadmins have updated the version, 

> Today, $NEW_VERSION of the openstreetmap-carto stylesheet (the default stylesheet on openstreetmap.org) has been released and rolled out to the openstreetmap.org servers. It might take a couple of days before all tiles show the new rendering.

Because we send the email immediately after releasing and notifying the sysadmins, we normally use the first form. This doesn't explain to people that they might not see the new style right away, because it's up to the admins to decide when/if to deploy the new version.

We should make this clear in a way that

- doesn't give an exact timeframe, because that varies
- doesn't implicitly blame the sysadmins for not having it out, because taking longer than 5 minutes is normal
- makes it clear that there may be other deployments

kocio-pl:
Resolves #2843.

### DIFF
--- a/RELEASES.md
+++ b/RELEASES.md
@@ -28,12 +28,16 @@ Decide among the maintainers if a new release is due.
 Dear all,
 
 Today, $NEW_VERSION of the openstreetmap-carto stylesheet (the default
-stylesheet on openstreetmap.org) has been released.
+stylesheet on the OSM website) has been released. Once changes are deployed 
+on the openstreetmap.org it will take couple of days before all tiles 
+show the new rendering.
 
-If the tile server admins have rolled out the new version, add
+If the tile server admins have rolled out the new version, change it to:
 
-  and rolled out to the openstreetmap.org servers. It might take a
-  couple of days before all tiles show the new rendering.
+  Today, $NEW_VERSION of the openstreetmap-carto stylesheet (the default
+  stylesheet on OSM website) has been released and rolled out 
+  to the openstreetmap.org servers. It might take a couple of days 
+  before all tiles show the new rendering. 
 
 Changes include
 *A list of changes*


RELEASES.md
```diff
@@ -28,12 +28,19 @@ Decide among the maintainers if a new release is due.
...
 Today, $NEW_VERSION of the openstreetmap-carto stylesheet (the default
-stylesheet on openstreetmap.org) has been released.
-
-If the tile server admins have rolled out the new version, add
-
-  and rolled out to the openstreetmap.org servers. It might take a
-  couple of days before all tiles show the new rendering.
+stylesheet on openstreetmap.org) has been released. The actual deployment 
+on OSM website is done by sysadmin team, so there are no guarantees 
+if and when the changes will be visible there.
```
pnorman: I don't like the tone of this sentence, it seems to be passively-aggressively blaming them that it's not rolled out.

RELEASES.md
```diff
@@ -28,12 +28,19 @@ Decide among the maintainers if a new release is due.
...
+
+If the tile server admins have rolled out the new version, change it to:
+
+  Today, $NEW_VERSION of the openstreetmap-carto stylesheet (the default
+  stylesheet on openstreetmap.org) has been released and rolled out 
+  to the openstreetmap.org servers. It might take a couple of days 
+  before all tiles show the new rendering. 
+  
+While developing standard OSM map is our primary goal, there are also
```
pnorman: This reads like we're elevating tile.osm.org above others

RELEASES.md
kocio-pl: Another wording: `so please be patient, they make their own decisions if and when should it happen.`

RELEASES.md
kocio-pl: I wanted to show how we design in reality (I'm not even aware where other direct deployments are), but if you like it more neutral, it can be for example:

`This style is not only used for rendering standard tile layer on OSM.org, it can be 
also deployed in other, independent websites and forks.`

kocio-pl: Or:

`This style can be also deployed in other websites and forks on their own schedule.`

kocio-pl:
@pnorman If none of the versions are the phrasing you like, feel free to reuse them as you think would be good.

RELEASES.md
matkoniecz: I agree that tone is wrong.

Maybe something like "Once changes are deployed on the openstreetmap.org it will take couple of days before all tiles show the new rendering"

RELEASES.md
matkoniecz: "This reads like we're elevating tile.osm.org above others" - I do not think that it is a problem.

kocio-pl:
I have updated this PR using different words.

RELEASES.md
matkoniecz: @pnorman What you think about new wording? I think that it is now OK.

matkoniecz:
I think that current version is OK, but I prefer to wait with merging for @pnorman repeated review.

matkoniecz:
Ping @pnorman

RELEASES.md
```diff
@@ -28,12 +28,19 @@ Decide among the maintainers if a new release is due.
 Dear all,
 
 Today, $NEW_VERSION of the openstreetmap-carto stylesheet (the default
-stylesheet on openstreetmap.org) has been released.
-
-If the tile server admins have rolled out the new version, add
-
-  and rolled out to the openstreetmap.org servers. It might take a
-  couple of days before all tiles show the new rendering.
+stylesheet on OSM website) has been released. Once changes are deployed
```
pnorman: s/on OSM website/on the OSM website/

RELEASES.md
```diff
@@ -28,12 +28,19 @@ Decide among the maintainers if a new release is due.
...
+
+If the tile server admins have rolled out the new version, change it to:
+
+  Today, $NEW_VERSION of the openstreetmap-carto stylesheet (the default
+  stylesheet on OSM website) has been released and rolled out 
+  to the openstreetmap.org servers. It might take a couple of days 
+  before all tiles show the new rendering. 
+  
+This style can be also deployed on other websites and forks on their 
+own schedule.
```
pnorman: I think it's better to just eliminate this paragraph.

PR #2866, opened 2017-09-25
Skip amenities list check

kocio-pl:
A bit of code cleaning. Basic testing has been done to check that the code compiles and amenities are still here.

### DIFF
--- a/project.mml
+++ b/project.mml
@@ -1442,15 +1442,8 @@ Layer:
           WHERE aeroway IN ('helipad', 'aerodrome')
             OR tourism IN ('artwork', 'alpine_hut', 'camp_site', 'caravan_site', 'chalet', 'wilderness_hut', 'guest_house', 'hostel',
                            'hotel', 'motel', 'information', 'museum', 'viewpoint', 'picnic_site')
-            OR amenity IN ('shelter', 'atm', 'bank', 'bar', 'bicycle_rental', 'bus_station', 'cafe',
-                           'car_rental', 'car_wash', 'cinema', 'clinic', 'community_centre',
-                           'fire_station', 'fountain', 'fuel', 'hospital', 'ice_cream', 'embassy', 'library', 'courthouse',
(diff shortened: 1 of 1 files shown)


math1985:
Does this have performance implications? That was the reason the WHERE statements where there in the first place.

kocio-pl:
It's just as big as shops list, so I guess they should unified this way (skipping) or another (providing full list).

math1985:
I believe the important factor is not the length of the list, but the total amount of amenities (versus the amount we are selecting) in a given bounding box.

project.mml
```diff
@@ -1442,14 +1442,7 @@ Layer:
...
-            OR amenity IN ('shelter', 'atm', 'bank', 'bar', 'bicycle_rental', 'bus_station', 'cafe',
-                           'car_rental', 'car_wash', 'cinema', 'clinic', 'community_centre',
-                           'fire_station', 'fountain', 'fuel', 'hospital', 'ice_cream', 'embassy', 'library', 'courthouse',
-                           'townhall', 'parking', 'bicycle_parking', 'motorcycle_parking', 'pharmacy', 'doctors',
-                           'dentist', 'place_of_worship', 'police', 'post_box', 'post_office', 'pub', 'biergarten',
-                           'recycling', 'restaurant', 'food_court', 'fast_food', 'telephone', 'taxi',
-                           'theatre', 'toilets', 'drinking_water', 'prison', 'hunting_stand', 'nightclub', 'veterinary',
-                           'social_facility', 'charging_station', 'arts_centre', 'ferry_terminal')
+            OR amenity IS NOT NULL -- skip checking a huge list and use a null check
             OR shop IS NOT NULL -- skip checking a huge list and use a null check
```
pnorman: Can you remove the comment on the 2nd line, since the one above covers both

kocio-pl:
Comments cleaned.

BTW: we already use simplified amenities checking for text layers and that was also what made me think that it should be done here.

PR #2867, opened 2017-09-26
Changing windmill color to man_made

kocio-pl:
Follow up to #2833.

In general windmill is rather typical man made (orientation point) than amenity (to be used somehow).

Before
speyx8nr
After
v6lvl0xv

### DIFF
--- a/amenity-points.mss
+++ b/amenity-points.mss
@@ -1006,7 +1006,7 @@
   [feature = 'man_made_windmill'][zoom >= 16] {
     marker-file: url('symbols/windmill.svg');
     marker-placement: interior;
-    marker-fill: @amenity-brown;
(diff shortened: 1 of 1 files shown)


kocio-pl:
Any comments or reviews?

sommerluk:
I haven’t looked to the code, but the colour change makes sense to me.

HolgerJeromin:
> Most windmills I know are also tourist sights, which fits in with the brown color.

How are windmills with "tourism=attraction" rendered? Could be a candidate for brown

kocio-pl:
Currently windmill is always just brown, but that's interesting idea.

sommerluk:
This needs to be rebased…

PR #2868, opened 2017-09-26
Showing all stations as major buildings

kocio-pl:
Part of a transport unification effort, related to #2094 and #2763.

We have no unified solution for transportation areas in practice, but they all can have station/terminal. I'd like to make them all look similar, because for the passengers this is the most important building on such area, no matter what kind of transport is involved.

[Example aerial station](http://www.openstreetmap.org/way/267743094)
z18
Before
y2s5btt0
After
u9zyqxjn

[Example public transport station](http://www.openstreetmap.org/way/97985328)
z18
Before
l53umamo
After
yzn2ezlr

### DIFF
--- a/buildings.mss
+++ b/buildings.mss
@@ -23,7 +23,9 @@
   [zoom >= 13] {
     [aeroway = 'terminal'],
     [amenity = 'place_of_worship'],
-    [building = 'train_station'] {
+    [building = 'train_station'],
+    [aerialway = 'station'],
+    [public_transport = 'station'] {
       polygon-fill: @building-major-fill;
       polygon-clip: false;
       [zoom >= 15] {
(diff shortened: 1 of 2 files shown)


dieterdreist:
this seems to assume public transport stations are buildings?

kocio-pl:
Yes, that's necessary condition to render them this way. Area rendering is not included in this PR, but that's also what I'd like to do.

buildings.mss
```diff
@@ -23,7 +23,9 @@
   [zoom >= 13] {
     [aeroway = 'terminal'],
     [amenity = 'place_of_worship'],
-    [building = 'train_station'] {
+    [building = 'train_station'],
+    [aerialway = 'station'],
+    [public_transport = 'station'] {
```
HolgerJeromin: Will this hide missing `building=*` on `public_transport=station` ?
I often found this in the past for place_of_worships which missed the `building` tag.

kocio-pl: In general (excluding building=train_station case of course) we're not able to tell if this is missing building tag or just this is an area. So this will show only buildings with that tag, area rendering for all stations is still to be added in separate PR.

kocio-pl:
Any comments or review here?

Issue #2580, opened 2017-03-14
Open/missing boundary between Uruguay and Argentina


naoliv:
I am unsure if this problem belongs here, sorry.

The boundary between Uruguay and Argentina https://www.openstreetmap.org/#map=3/-34.45/-56.95 is open/missing on zoom levels 1 to 3:

![](https://cloud.githubusercontent.com/assets/982338/23899490/1acb1598-0894-11e7-8292-89fafca0c7ec.png)
![](https://cloud.githubusercontent.com/assets/982338/23899491/1ad09914-0894-11e7-8ecb-e270df857fa6.png)
![](https://cloud.githubusercontent.com/assets/982338/23899492/1ad175b4-0894-11e7-82d0-482eb8d30c35.png)

Is it some kind of problem?
Is it possible to force only this area to be re-rendered?

imagico:
This is caused by two things:

* we currently use boundaries not based on OSM data for z1 to z3 - see #2172.
* the Rio de la Plata is mapped as a water polygon outside the coastline (https://www.openstreetmap.org/relation/3474227) and therefore does not show up at z<6

The first is a problem of this style, the second one is data/tagging related.

kocio-pl:
Just for the record: this is fixed now by #2873.

math1985:
Besides from the fix, I believe this is indeed a tagging problem (it doesn't make sense to tag areas outside of the coastline as water polygon).

kocio-pl:
What tagging would you propose then?

math1985:
Hmm, I thought that 'Rio de la Plata is mapped as a water polygon outside the coastline' was meant in technical sense, i.e. outside of the OSM coastline. But that does not seem to be the case. That makes it more difficult.

Issue #1465, opened 2015-03-28
water label placement offset


sommerluk:
The label for water features seems to be sometimes placed at wrong positions.

Consider the “Lagune Ébrié” at http://www.openstreetmap.org/relation/1982728#map=10/5.2168/-4.2078 which has a label that is displayed far away from the lagoon itself:

![screen3](https://cloud.githubusercontent.com/assets/6830724/6882269/a0f63ae2-d576-11e4-863c-30b25f2512d0.png)

I’m quite sure that this paricular lagoon had a well-placed label some monthes ago.

pnorman:
Could that label be coming from a different object? Not only is it not the centroid of that relation, it  does not lie within the convex hull or bounding box of it.

sommerluk:
I’m not sure.

Just a few observations:

→ Some monthes ago this particular lagoon had a well-placed label (within the water area)
→ This label disappeared some monthes ago.
→ I’ve seen the label in the ocean only a few days ago.
→ I’ve searched (www.openstreetmap.org/search?query=Lagune Ébrié) for the text, and there is only one OSM element in the results
→ I’ve also loaded with JOSM all the area of the ocean around maybe 50 km and didn’t found anything in this sense.
→ The label in the ocean is only visible from z10 to z13. Starting with z14 it disappears.
→ I’ve tried to reproduce this locally with TileMill, the current style and a current data extract of Ivory Coast from Geofabrik. There, I see the label in the ocean from z10 to z11.
→ Within the water area of the lagoon, I didn’t find a label neither at openstreetmap.org nor in my local TileMill rendering.

math1985:
Very strange.

>  → I’ve tried to reproduce this locally with TileMill, the current style and a current data extract of Ivory Coast from Geofabrik. There, I see the label in the ocean from z10 to z11.

Especially this is interesting, means it's not an issue with the posgis database at the production server.

sommerluk:
@pnorman 

> Could that label be coming from a different object?

No, it doesn’t. I’m now sure. I’ve loaded the hole region (step by step) in JOSM. Then, I’ve selected the lagoon relation and copied it to a new layer. This layer I’ve saved and loaded it in the database. So there was only the lagoon relation and its members in the database; no other elements. Here is the result in TileMill:

![screen30](https://cloud.githubusercontent.com/assets/6830724/6886052/814c5f82-d620-11e4-8796-68fda617d724.png)

math1985:
[...]

PR #2873, opened 2017-09-29
Start showing water areas from z0

### LINKED ISSUE #1604: Lake Ontario does not render at zoom 5 or below
Lake Ontario does not render at zoom 5 or below. It looks fine at zoom 6 or above. I think that this is a bug in the OSM map style because I think that the data is fine.

kocio-pl:
Resolves #1604.
Closes #2864.

Test shows that changing pixel query factor from 0.01 to 1 probably does not make visual difference (or the change is subtle), but improves performance a lot and we can reasonably render all the water areas from z0. For details see discussion in #1604.

Only water layer:
no-subpixel-select-biglakes-z6

### DIFF
--- a/project.mml
+++ b/project.mml
@@ -237,11 +237,11 @@ Layer:
               OR landuse IN ('reservoir', 'basin')
               OR "natural" IN ('water', 'glacier'))
             AND building IS NULL
-            AND way_area > 0.01*!pixel_width!::real*!pixel_height!::real
+            AND way_area > 1*!pixel_width!::real*!pixel_height!::real
           ORDER BY COALESCE(layer,0), way_area DESC
         ) AS water_areas
     properties:
-      minzoom: 4
+      minzoom: 0
   - id: landcover-area-symbols
     geometry: polygon
     <<: *extents
@@ -1879,7 +1879,7 @@ Layer:
             way_area/NULLIF(!pixel_width!::real*!pixel_height!::real,0) AS way_pixels,
             COALESCE(
               'landuse_' || CASE WHEN landuse IN ('forest', 'military', 'farmland') THEN landuse ELSE NULL END,
-              'natural_' || CASE WHEN "natural" IN ('wood', 'glacier', 'sand', 'scree', 'shingle', 'bare_rock') THEN "natural" ELSE NULL END,
+              'natural_' || CASE WHEN "natural" IN ('wood', 'glacier', 'sand', 'scree', 'shingle', 'bare_rock', 'water') THEN "natural" ELSE NULL END,
               'place_' || CASE WHEN place IN ('island') THEN place ELSE NULL END,
               'boundary_' || CASE WHEN boundary IN ('national_park') THEN boundary ELSE NULL END,
               'leisure_' || CASE WHEN leisure IN ('nature_reserve') THEN leisure ELSE NULL END
@@ -1888,7 +1888,7 @@ Layer:
             CASE WHEN building = 'no' OR building IS NULL THEN 'no' ELSE 'yes' END AS is_building -- always no with the where conditions
           FROM planet_osm_polygon
           WHERE (landuse IN ('forest', 'military', 'farmland')
-              OR "natural" IN ('wood', 'glacier', 'sand', 'scree', 'shingle', 'bare_rock')
+              OR "natural" IN ('wood', 'glacier', 'sand', 'scree', 'shingle', 'bare_rock', 'water')
               OR "place" IN ('island')
               OR boundary IN ('national_park')
               OR leisure IN ('nature_reserve'))
@@ -1897,7 +1897,7 @@ Layer:
           ORDER BY way_area DESC
         ) AS text_poly_low_zoom
     properties:
-      minzoom: 7
+      minzoom: 0
       maxzoom: 9
   - id: text-poly
     class: text
--- a/water.mss
+++ b/water.mss
@@ -43,7 +43,7 @@
   [natural = 'water']::natural,
   [landuse = 'reservoir']::landuse,
   [waterway = 'riverbank']::waterway {
-    [zoom >= 6] {
+    [zoom >= 0] {
       polygon-fill: @water-color;
       [way_pixels >= 4] {
         polygon-gamma: 0.75;
@@ -313,12 +313,13 @@
   }
 }
 
+.text-low-zoom[zoom < 10],
 .text[zoom >= 10] {
   [feature = 'natural_water'],
   [feature = 'landuse_reservoir'],
   [feature = 'landuse_basin'],
   [feature = 'waterway_dock'] {
-    [zoom >= 10][way_pixels > 3000],
+    [zoom >= 0][way_pixels > 3000],
     [zoom >= 17] {
       text-name: "[name]";
       text-size: 10;


math1985:
> Test shows that changing pixel query factor from 0.01 to 1 probably does not make visual difference (or the change is subtle), 

As the demo server hasn’t rerendered all tiles, we cant be too sure about that.

kocio-pl:
I have not claimed that there's no doubt, but the fact that I have not yet found visible glitches and maybe one place which hides smallest lakes is quite reassuring. That being said it's good to check as much as it's possible here:

 http://bl.ocks.org/math1985/raw/af7a602c222dbf1ff1a2c0d84ed755b7/#3.00/18.42/1.85

rrzefox:
These are really two changes:
1.) changing the pixel query factor
2.) start rendering water from Z0 instead of Z6.
2 will not work well without 1 because of memory requirements.

It should be mentioned that the sideeffect of change 2 will be massively increased rendering times for (some) tiles on Zoomlevels 0 to 5, e.g. on Z0 from 0.5 to 100+ seconds. However, there are only few tiles in these zoomlevels, and they are usually never rerendered on the fly.

So to phrase the question in a different way: Do we want a Z0 tile that renders very fast but hides away huge, distinctive water areas like the border lakes between .us and .ca, or do we want to spend a few minutes of rendering time to have a better looking, closer to reality and more easily recognizable map?

If someone is really worried about this, it might be possible to drop them in Z0 to Z2, with the justification that it's a really abstract map at these zoom levels. However, IMHO starting from Z3, where the country labels start to show up, you have to show them - else the fact they're missing really hits you in the face. And it's more consistent to show them from Z0.

kocio-pl:
Just some additional notes:
- with z6+ we will gain some performance/memory because of 1)
- La Plata mouth is specific case when even on z1 it's visible that something is broken, so in general I would be against dropping it; we have [something similar](http://bl.ocks.org/math1985/raw/af7a602c222dbf1ff1a2c0d84ed755b7/#5.00/65.455/38.589) with [Archangielsk](http://www.openstreetmap.org/relation/1620851) on z2+, but there's no border, so there's no such visual problem

kocio-pl:
Example of some rivers rendered as water areas, as discussed in a separate PR (see https://github.com/gravitystorm/openstreetmap-carto/pull/2874#issuecomment-333335086):

http://bl.ocks.org/math1985/raw/af7a602c222dbf1ff1a2c0d84ed755b7/#5.00/65.044/-143.550

They seem to be continuous to me, but it doesn't look great, because some segments are wider and some other are thinner (near 1 px, I guess). That's because we don't use proper generalization of waterway lines on low zoom as rivers classification is currently lacking in OSM (2 parallel proposals exist to fix that problem). For me it's acceptable and enough to drop subpixel accuracy for water - more low zoom rework is needed anyway.

This river:

http://bl.ocks.org/math1985/raw/af7a602c222dbf1ff1a2c0d84ed755b7/#6.00/-2.481/-58.100

does not look continuous on z3 (there are just some parts visible), but that's expected as there's no relation including all the segments:

https://www.openstreetmap.org/relation/32983#map=6/-2.175/-54.130

math1985:
I suspect this has nothing to do with the subpixel problem, I think even with water-blocks of 0,01 pixel these rivers would be so narrow that they wouldn't (noticeably) render.

Both rendering examples look acceptable to me.

daganzdaanda:
Great to see this, thank you all!

Found an error, probably #1465, the long-known mapnik bug with labels not ending up inside the polygon?
http://bl.ocks.org/math1985/raw/af7a602c222dbf1ff1a2c0d84ed755b7/#4.00/56.40/32.30
Look for the Caspian Sea just west of Kiev ;-)

kocio-pl:
Yes, that was reported already in https://github.com/gravitystorm/openstreetmap-carto/issues/1604#issuecomment-333751590.

Any additional tests, comments or reviews? I wonder how much time or actions are needed to make sure and decide (and hopefully merge this code).

daganzdaanda:
Yeah, I read that later... Lake Victoria's label is also funny. 

Back on topic, I have looked at a lot of different areas worldwide and it is a big improvement for the low zooms everywhere. I mean, just look at it! Lakes are most obvious, but also the big rivers (Congo) and reservoirs really define their landscapes.

kocio-pl:
We may wait another 2 weeks with decision, but I'm not sure what are we waiting for? From the formal point of view one positive review would be enough - and with all the comments and my tests it looks like it's the only thing that is really needed.

math1985:
Sorry for not reviewing it earlier. It's clear this should go through.

I'm a bit worried still about the performance of the text-poly-low-zoom layer - it doesn't have any minimum pixel requirements in the SQL definition.

kocio-pl:
I'm not sure how should we test performance. 

Real life testing (https://github.com/gravitystorm/openstreetmap-carto/pull/2874#issuecomment-334507450) with this patch included does not indicate big problems, as we know it already.

There's a separate ticket for performance testing, but no conclusions yet: https://github.com/gravitystorm/openstreetmap-carto/issues/1287

math1985:
Ok, I wasn't sure if  text-poly-low-zoom was already included.

Let's go ahead then!

kocio-pl:
You can read about other possible solutions of this problem (this one looks like a mild version of 2):

http://blog.imagico.de/parting-the-waters/

it-solutions-zehaczek:
_2.) start rendering water from Z0 instead of Z6._

Thank you guys, that just cost me 2 days to figure out why the hell the shapefiles aren't used for my lower zooms that worked properly on another server with an older version. That is why. I downgraded before this commit now.

kocio-pl:
Good that you have found where the problem is and you were able to work around it. 

But what is the problem? This change just added waters directly from OSM database, in addition to the shapefiles. Did you want just shapefiles without any waters on top or I misunderstood you?

it-solutions-zehaczek:
Before the change:
`renderd[2489]: DEBUG: START TILE default 0 0-0 0-0, new metatile
renderd[2489]: Rendering projected coordinates 0 0 0 -> -20037508.342800|-20037508.342800 20037508.342800|20037508.342800 to a 1 x 1 tile
renderd[2489]: Using web mercator projection settings
renderd[2489]: DEBUG: DONE TILE default 0 0-0 0-0 in **0.155 seconds**
debug: Creating and writing a metatile to /var/lib/mod_tile/default/0/0/0/0/0/0.meta
`

After:
`renderd[26584]: DEBUG: START TILE default 0 0-0 0-0, new metatile
renderd[26584]: Rendering projected coordinates 0 0 0 -> -20037508.342800|-20037508.342800 20037508.342800|20037508.342800 to a 1 x 1 tile
renderd[26584]: Using web mercator projection settings
renderd[26584]: Using web mercator projection settings
renderd[26584]: Using web mercator projection settings
renderd[26584]: Using web mercator projection settings
renderd[26584]: Using web mercator projection settings
renderd[26584]: Using web mercator projection settings
renderd[26584]: Using web mercator projection settings
renderd[26584]: DEBUG: DONE TILE default 0 0-0 0-0 in **370.529 seconds**
debug: Creating and writing a metatile to /var/lib/mod_tile/default/0/0/0/0/0/0.meta`

I didn't miss any waters. Fast caostlines and country shapes worked just perfectly for me and is so much faster.

dieterdreist:
sent from a phone

> On 24. Jan 2018, at 18:46, Christian <notifications@github.com> wrote:
> 
> I didn't miss any waters. Fast caostlines and country shapes worked just perfectly for me and is so much faster.


yes, but usually you don’t render z0 so often.

kocio-pl:
On OSMF servers the update of z0-z12 is made every few weeks and re-rendering _all of them_ takes just few hours with the current version, because they don't change a lot that often and there's no need to do this more often. Most of the time takes rendering z13-z19, because there are much more of them (z13 alone is has more tiles than all z0-z12 combined - [and so on](https://wiki.openstreetmap.org/wiki/Zoom_levels)) and the changes on higher zoom levels are more visible.

Of course if you try to have the most fresh low zoom and don't care for inland waters there, that's the best you can do. However I'm happy with current situation, so it looks we just have different needs.

it-solutions-zehaczek:
I'm fine with rare rendering of z0, my users are also starting from z4, what messed me up is that i use z0 for setup tests and need 2 days to figure out why in comparison this is so slow.

I also compared the 2 results and could not see any difference in z0, so this totally makes no sense to me. Why not start with waters in z2++ ?

kocio-pl:
We didn't try to make artificial thresholds and the problem with Uruguay/Argentina border, described in #2580, is visible from z1 (it's broken currently again, we plan to repair it).

it-solutions-zehaczek:
What about updated shapefiles?

kocio-pl:
I don't understand this question - what do you mean exactly?

Shapefiles are produced from coastlines, but all the waters we have added on low zoom in this PR are inland, so it's a different kind. [La Plata](https://en.wikipedia.org/wiki/R%C3%ADo_de_la_Plata) mouth looks like just a part of the ocean, but it's tagged as a river area - and probably that's the truth (sweet water vs salt water).

Issue #1566, opened 2015-05-21
way_area filtering


pnorman:
Currently we filter by way_area to require some areas to be at least 1% of a pixel with `way_area > 0.01*!pixel_width!::real*!pixel_height!::real`

0.01 was chosen as a very conservative value for water rendering problems, and may not be the best value.

It turns out that dropping small areas significantly reduces the number of objects returned by the database at up to medium zooms.

Going from 0.01 to to 0.1 or 0.25 would cut the areas returned by a factor of 3-5 and make partial indexes on area useful two zoom levels sooner.

![image](https://cloud.githubusercontent.com/assets/1190866/7743465/f2d63972-ff4f-11e4-8a2a-6ed0c6f7ce21.png)

Data from Europe extract.

imagico:
This is a tricky matter and will require some rendering samples.

Am i right to assume that your graph includes all areas including those not rendered at the zoom level in question and those rendered with an outline (most notably buildings at the higher zooms)?

Size filtering in combination with the AGG polygon rendering and  the size dependent gamma manipulation introduces a bias into area rendering, i.e. the acutal fraction of surface areas covered by a certain area type differs significantly from what is shown.  Parameters could be tuned to minimize this but unfortunately this also depends a lot on how small grained the mapping is.  This is already a problem right now, notice the brighter color of forest areas around

http://www.openstreetmap.org/#map=8/52.655/15.614

resulting from the regular grid splitting of the forest mapping there.  Such effects would be even stronger when you fully leave out polygons already when they still cover a significant fraction of a pixel.

pnorman:
using tile 8/138/83.

0.01 (master)
![0.01](https://cloud.githubusercontent.com/assets/1190866/7756624/56c9e71a-ffb1-11e4-899a-50ca9c51fe9a.png)

0.05
![0.05](https://cloud.githubusercontent.com/assets/1190866/7756677/c1405f8e-ffb1-11e4-8b8a-babe338d8395.png)

0.10
![0.10](https://cloud.githubusercontent.com/assets/1190866/7756699/e0196162-ffb1-11e4-9cef-f034cba42aa7.png)

0.25
![0.25](https://cloud.githubusercontent.com/assets/1190866/7756721/fcb4b308-ffb1-11e4-8b50-32b37c1718ce.png)

0.5
[...]

PR #2874, opened 2017-09-29
Dropping subpixel accuracy for areas

kocio-pl:
Follow up to #2873.
Related to #1604.

It needs some testing, but it looks that dropping subpixel accuracy from SQL queries does not have visible effect on rendering - if any - while increasing performance (at least decreasing memory usage).

Affected objects: water, landcovers, buildings and nature reserves.

### DIFF
--- a/indexes.sql
+++ b/indexes.sql
@@ -25,18 +25,21 @@ CREATE INDEX planet_osm_polygon_water
   WHERE waterway IN ('dock', 'riverbank', 'canal')
     OR landuse IN ('reservoir', 'basin')
     OR "natural" IN ('water', 'glacier');
-CREATE INDEX planet_osm_polygon_military
-  ON planet_osm_polygon USING GIST (way)
-  WHERE landuse = 'military';
 CREATE INDEX planet_osm_polygon_nobuilding
   ON planet_osm_polygon USING GIST (way)
   WHERE building IS NULL;
 CREATE INDEX planet_osm_polygon_name
   ON planet_osm_polygon USING GIST (way)
   WHERE name IS NOT NULL;
+CREATE INDEX planet_osm_polygon_way_area_z10
+  ON planet_osm_polygon USING GIST (way)
+  WHERE way_area > 23300;
+CREATE INDEX planet_osm_polygon_military
+  ON planet_osm_polygon USING GIST (way)
+  WHERE landuse = 'military';
 CREATE INDEX planet_osm_polygon_way_area_z6
   ON planet_osm_polygon USING GIST (way)
-  WHERE way_area > 59750;
+  WHERE way_area > 5980000;
 CREATE INDEX planet_osm_point_place
   ON planet_osm_point USING GIST (way)
   WHERE place IS NOT NULL AND name IS NOT NULL;
--- a/indexes.yml
+++ b/indexes.yml
@@ -28,7 +28,9 @@ polygon:
           OR landuse IN ('reservoir', 'basin')
           OR "natural" IN ('water', 'glacier')
   way_area_z6:
-    where: way_area > 59750
+    where: way_area > 5980000
+  way_area_z10:
+    where: way_area > 23300
 roads:
   # The roads table only has a subset of data, so it's just got some low-zoom
   # indexes and some fairly selective ones for high zoom
--- a/project.mml
+++ b/project.mml
@@ -97,7 +97,7 @@ Layer:
             FROM planet_osm_polygon
             WHERE (landuse IN ('forest', 'military', 'farmland', 'residential', 'commercial', 'retail', 'industrial', 'meadow')
               OR "natural" IN ('wood', 'wetland', 'mud', 'sand', 'scree', 'shingle', 'bare_rock', 'heath', 'grassland'))
-              AND way_area > 0.01*!pixel_width!::real*!pixel_height!::real
+              AND way_area > 1*!pixel_width!::real*!pixel_height!::real
               AND building IS NULL
             ORDER BY way_area DESC
           ) AS features
@@ -152,7 +152,7 @@ Layer:
               OR tourism IN ('camp_site', 'caravan_site', 'picnic_site')
               OR highway IN ('services', 'rest_area')
               OR railway = 'station')
-              AND way_area > 0.01*!pixel_width!::real*!pixel_height!::real
+              AND way_area > 1*!pixel_width!::real*!pixel_height!::real
             ORDER BY way_area DESC
           ) AS landcover
         ) AS features
@@ -265,7 +265,7 @@ Layer:
           FROM planet_osm_polygon
           WHERE ("natural" IN ('marsh', 'mud', 'wetland', 'wood', 'beach', 'shoal', 'reef', 'scrub', 'sand') OR landuse = 'forest')
             AND building IS NULL
-            AND way_area > 0.01*!pixel_width!::real*!pixel_height!::real
+            AND way_area > 1*!pixel_width!::real*!pixel_height!::real
           ORDER BY COALESCE(layer,0), way_area DESC
         ) AS landcover_area_symbols
     properties:
@@ -434,7 +434,7 @@ Layer:
             AND building != 'train_station'
             AND (aerialway IS NULL OR aerialway != 'station')
             AND ((tags->'public_transport') IS NULL OR tags->'public_transport' != 'station')
-            AND way_area > 0.01*!pixel_width!::real*!pixel_height!::real
+            AND way_area > 1*!pixel_width!::real*!pixel_height!::real
           ORDER BY COALESCE(layer,0), way_area DESC
         ) AS buildings
     properties:
@@ -460,7 +460,7 @@ Layer:
                  OR building = 'train_station'
                  OR aerialway = 'station'
                  OR tags->'public_transport' = 'station')
-            AND way_area > 0.01*!pixel_width!::real*!pixel_height!::real
+            AND way_area > 1*!pixel_width!::real*!pixel_height!::real
           ORDER BY COALESCE(layer,0), way_area DESC)
         AS buildings_major
     properties:
@@ -1371,7 +1371,7 @@ Layer:
           FROM planet_osm_polygon
           WHERE (boundary = 'national_park' OR leisure = 'nature_reserve')
             AND building IS NULL
-            AND way_area > 0.01*!pixel_width!::real*!pixel_height!::real
+            AND way_area > 1*!pixel_width!::real*!pixel_height!::real
         ) AS national_park_boundaries
     properties:
       minzoom: 8


kocio-pl:
The differences are real, but they are not obvious, which is good. 

I've made two Poland exports from Kosmtik at z10 (no buildings) - each about 40 MB, so I will not upload them. Then I compared them with ImageMagick (after disabling its policy limits in /etc):

```
compare -compose src before.png after.png difference.png
```

Click to see the full scale result (~0.9 MB):
difference

Then I opened all files in Gimp as layers and zoomed in where the changes are biggest according to this image. I was able to see the change when flipping visibility on 800% zoom, but it doesn't look like error to me - just a bit less of details:

Before
gimp-before
After
gimp-after
Difference
gimp-difference

pnorman:
If doing this then the custom indexes will need adjusting.

My recollection was that I did see differences with a 1px filter, but that was a **long** time ago when I tested.

kocio-pl:
> If doing this then the custom indexes will need adjusting.

I'm not aware of this problem. I don't see any specific value in `indexes.sql`/`indexes.yml` except `way_area > 59750` for `planet_osm_polygon_way_area_z6` - or you mean something different?

> My recollection was that I did see differences with a 1px filter, but that was a long time ago when I tested.

This query change means that we just don't want very small areas to have any impact on the rendering, but it's not distorting anything. In my opinion any object ~1 pixel big is still insignificant for the whole image. In some cases it might be even better to hide them, but it's rather styling problem than database filtering.

aceman444:
What if somebody makes a farmland via small polygons representing each parcel. Would this be still rendered with the farmland colour before the patch and be completely missing after the patch?

kocio-pl:
It depends on how small they are. 

If they are <~1 px on zoom level n, they will just not be rendered on this level, but will be visible on n+1 and z19 should be more than enough for any area, no matter how small.

kocio-pl:
@rrzefox Could you also apply this patch on your server? 

It's just about being on the safe side, because we already know that:
- changes are small and show up only with some minor objects, as tested with the water, but they don't break anything (as expected with just filtering out small objects, and we do it already in some places)
- the code says we're getting less objects from these queries, we just don't know exact numbers 
- I have experienced that memory consumption in worst case with just a water was significantly lower, like 25+% (~6 GB max instead of 6GB+2GB of swap - possibly more, because I run out of swap), but that was using Kosmtik, not a renderd

So we know that we **_should_** end up using less resources, but we just don't know to what degree. So my take is that if there won't be bad rendering changes (just less of some small objects is not bad in itself), it's worth to merge this, even just to spare some memory.

pnorman:
> If they are <~1 px on zoom level n, they will just not be rendered on this level, but will be visible on n+1 and z19 should be more than enough for any area, no matter how small.

No, because a large area mapped with blocks of 0.5 px would disappear completely. This isn't an issue with 0.01px because by the time that happens, roundoff errors will have overwhelmed anything else.

kocio-pl:
This was a symbol to show areas "smaller but near 1 px". The rest will be just visible on n+2 - and so on. We don't even know how much such small areas change the rendering when there's more of them.

The real question is however not "how Mapnik works with a lot of subpixel areas", but "do you know examples where this could be a problem"?

pnorman:
> I'm not aware of this problem. I don't see any specific value in indexes.sql/indexes.yml except way_area > 59750 for planet_osm_polygon_way_area_z6 - or you mean something different?

That value is based on current queries.

> This was a symbol to show areas "smaller but near 1 px". The rest will be just visible on n+2 - and so on.

I'm not sure what the relevance of zooming in is.

> We don't even know how much such small areas change the rendering when there's more of them.

Yes we do - unless you start to get so many areas there's a roundoff error, it's the same as if they were joined.

kocio-pl:
> That value is based on current queries.

What should it be after this PR? I will change it in the code.

> I'm not sure what the relevance of zooming in is.

It's good to remember that this style has not just one scale, as on the paper. It means that we don't "drop" any data, we just show them later.

pnorman:
> What should it be after this PR? I will change it in the code.

It should be about 0.99 of the cutoff threshold at some zoom, rounded down. This makes sure that the partial condition closely matches the queries. As for what zoom, it probably won't be z6 anymore. I looked at index size vs zoom and made a call on when the tradeoffs were best. I'd guess the area from z7-z9 is more appropriate with these changes.

> It's good to remember that this style has not just one scale, as on the paper. It means that we don't "drop" any data, we just show them later.

Using the example of water, the problem isn't deciding not to render very small lakes. The problem is having gaps appear in rivers where the individual polygons are under the threshold but the overall size is large.

kocio-pl:
OK, so either I need some numbers for custom indexes or you can make additional patch if this PR will be merged - I leave it to you.

> Using the example of water, the problem isn't deciding not to render very small lakes. The problem is having gaps appear in rivers where the individual polygons are under the threshold but the overall size is large.

Thanks, now I see. For complex objects made of smaller parts - like rivers - I would rely on relations, not individual water areas. Using any threshold is just guessing (probability of error) and safeguarding (eating resources to correct possible errors). Does it make sense or am I missing something?

pnorman:
> Thanks, now I see. For complex objects made of smaller parts - like rivers - I would rely on relations, not individual water areas. Using any threshold is just guessing (probability of error) and safeguarding (eating resources to correct possible errors). Does it make sense or am I missing something?

No, relations are not a solution. Water is a bit different because it has a name, but for forests, someone shouldn't make a relation unless there's an inner hole or some other reason.

pnorman:
> OK, so either I need some numbers for custom indexes or you can make additional patch if this PR will be merged - I leave it to you.

What are the area values at different zooms, calculated as above?

kocio-pl:
> No, relations are not a solution. Water is a bit different because it has a name, but for forests, someone shouldn't make a relation unless there's an inner hole or some other reason.

For me there's a big difference between a river and small lakes/forests/buildings etc. I'm not worried about small lakes or forests - they don't need to be shown at a certain level, because they don't need to be continuous.

The river is different when drawn as a water areas - they should be collected in a relation anyway. The same is true for natural reserves - they are also tagged this way when they consist of multiple areas. So I guess we are safe already. Are there any landuses or buildings that should be rendered continuously?

> What are the area values at different zooms, calculated as above?

I'm not sure how to apply the rule you gave and get the numbers.

math1985:
So in two areas we see significant differences:
* At this [cemetery](https://www.openstreetmap.org/#map=16/52.2981/20.8960&layers=N) each plot is tagged as a separate cemetery. Therefore at lower zoom levels the plots/cemeteries fall under 0,01 pixel and are no longer rendered. To me this is clearly wrong tagging.
* Here in [Ruda](https://www.openstreetmap.org/#map=16/52.2839/20.9728&layers=N) every piece of grass is mapped separately. Also these pieces fall under the 0,01 pixel limit at low zoom levels, leading the area to be rendered as residential rather than grass in the new rendering.

To me all these effects are rather small, and remove a lot of noise (and blurry/mixed colors) from the map. So I'm in favour of merging this as it looks now (but haven't tested it myself on Luxembourg yet).

Note also that if we ever want to move to vector maps, probably we will need to do something similar.

By the way, I would also be interested, just for fun, to see how the map change if we increase the pixel limit even further.

kocio-pl:
The cemetery tagging is just being overzealous - [cemetery=* key](https://wiki.openstreetmap.org/wiki/Key:cemetery) does not require amenity=cemetery.

kocio-pl:
> By the way, I would also be interested, just for fun, to see how the map change if we increase the pixel limit even further.

Challenge accepted! Poland, z8 (click to view pictures in a full scale):

0.01


1


5


10


Difference 0.01-10:
difference

rrzefox:
> @rrzefox Could you also apply this patch on your server?

this has been applied two days ago and all tiles should have been rerendered by now.

math1985:
> this has been applied two days ago and all tiles should have been rerendered by now.

Great, thanks a lot for your help!

math1985:
This PR makes [this forest](http://bl.ocks.org/math1985/raw/af7a602c222dbf1ff1a2c0d84ed755b7/#13.00/52.4091/0.6726) disappear from z7.

math1985:
@matkoniecz Thanks, moving the cut-off from 1 to 5 pixels clearly does not work. Many forests in Poland seem to have been important in small blocks, such as this one: https://www.openstreetmap.org/#map=13/52.7647/16.5602

kocio-pl:
@rrzefox: I'm very glad of your help with testing medium and low zoom levels! It's important tool, since it's a risky ground that is hard to explore on our home environments (for database size and performance reasons). It's also a lot of fun for me to discover visual differences live!

kocio-pl:
It has probably nothing to do with pixel accuracy, but do you have any idea why [The Cotswolds national park](http://www.openstreetmap.org/relation/166570) (GB) disappears?

math1985:
My guess is that somebody broke the relation. An outer way missing or so. Probably slightly different db versions on @rrzefox's version and live.

kocio-pl:
@rrzefox I'm aware that no testing is the same as production and real life patterns can be very different, but at this moment your performance data is the best we could have. Did you noticed any interesting change in performance or memory usage after deploying this patch?

rrzefox:
> It has probably nothing to do with pixel accuracy, but do you have any idea why The Cotswolds national park (GB) disappears?

math1985's guess is probably right, the (unchanged) german style using the same DB doesn't show it anymore either after rerendering the tile, so this is definitely not a bug introduced by a this PR.

> @rrzefox I'm aware that no testing is the same as production and real life patterns can be very different, but at this moment your performance data is the best of what we have. Did you noticed any interesting change in performance or memory usage after deploying this patch?

There have been no noticable changes in memory-requirements according to my munin graphs. For the performance a.k.a. rendering-time, as stated before, the normal fluctuations in it are huge unless you render one tile at a time in a defined order. Just to give you an impression how huge, here are some rendering-times for the metatile containing part of that cotswolds mentioned above:

```
     ts     | rendertime
------------+------------
 1507016621 |    126.814
 1506704531 |    145.413
 1504479030 |    124.269
 1503269195 |    208.874
 1477176583 |    139.962
 1466030829 |    104.275
 1462575055 |    406.573
 1461711065 |    404.904
 1459983090 |    222.118
```

The first line is with this PR applied, the others are previous renderings. As you can see, it's completely impossible to see a performance change of a few percent in this data. The natural fluctuations due to other tiles rendering at the same time, what is in the disk cache, and so on, are way higher.

daganzdaanda:
I've looked at the comparison in many places. The drawback of some small areas suddenly appearing out of nowhere, is IMHO an acceptable trade-off for the huge improvement of seeing water areas earlier. 

As mentioned, the drawback is most noticeable with glaciers (because they appear early and probably because of the outline that seems to exaggerate the size). Also, the French cadastre imports have put lots of very small details (huts, pools) in the data, so one can see some differences. But I believe that nobody would notice or complain without the before-after comparison. Interestingly, I have not seen any area where I could notice a change in forests and farmland. At the relevant zooms I think the colours are too similar and too light to make out such tiny pixel differences.

A noticeable change is also the disappearing "blue haze", most visible in Canada: http://bl.ocks.org/math1985/raw/af7a602c222dbf1ff1a2c0d84ed755b7/#6.00/55.148/-59.438
I think this is a change for the better.

kocio-pl:
> The drawback of some small areas suddenly appearing out of nowhere, is IMHO an acceptable trade-off for the huge improvement of seeing water areas earlier.

To be clear: this patch is about cutting **all** the small (<1 px) areas. For water it'd be enough just to cut water areas query. 

> There have been no noticable changes in memory-requirements according to my munin graphs. For the performance a.k.a. rendering-time, as stated before, the normal fluctuations in it are huge unless you render one tile at a time in a defined order.

I was hoping that there will be performance increase, but that's not showing in real life data, so no visible gain in cutting. I would however do it anyway make the load smaller, since there's usually no way to decrease it, even just few %%.

kocio-pl:
I have finally found the place where some buildings are so small, that are missing on z15: 

http://bl.ocks.org/math1985/raw/af7a602c222dbf1ff1a2c0d84ed755b7/#19.00/6.31849/-370.81199

This however does not make a difference if you can't make a live compare.

kocio-pl:
> I'm not convinced that going this far is safe, and would like to see the difference in a region with lots of almost 1px areas before having confidence it's okay.

@pnorman What's your opinion now? Have you found such place?

> The way area partial indexes will need adjusting to any change.

I still don't know how exactly calculate new indexes (in case we want to apply this patch)?

matthijsmelissen:
@kocio-pl What is your current opinion on this PR?

kocio-pl:
I still think it would be good, but I don't understand how should we change the custom indexes then.

kocio-pl:
I have just found the filtering discussion, will read it later: #1566.

kocio-pl:
I'm not a database guy, but maybe you know how to tune the custom indexes according to what Paul proposed:

> It should be about 0.99 of the cutoff threshold at some zoom, rounded down.

matthijsmelissen:
No, I have no idea.

pnorman:
I'm not having much luck explaining, so here are the calculations

Our current index is designed for a cutoff of 0.01 of a pixel at z6 and below. The web mercator plane is 40075342.49 mercator meters in each direction, so the pixel width and height is 40075342.49/256*2^-z, giving a `!pixel_width!` and `!pixel_height!` of `2446.00479`. This makes the planner evaluate the condition `way_area > 0.01*!pixel_width!::real*!pixel_height!::real` to be `way_area > 59829.3943`. 

Next it looks at how to run the query. One way is using the `planet_osm_polygon_index` index with the `way && !bbox!` part of the query to find only geometries that intersect the bounding box, and this index is 33GB. Another one is the `planet_osm_polygon_way_area_z6` index, which can be used the same way, but only if the query also matches the condition of  `way_area > 59750`. Since anything that matches `way_area > 59829.3943` also matches the index's condition, it can use it. This index is 1284MB.

Because one index is 30 times the size of the other, it expects the `planet_osm_polygon_way_area_z6` index to be faster.

The reason we don't make an index for each zoom is threefold. In reverse order of importance, they are:

- Each index slows down updates. This doesn't matter much for the number of indexes we have, but it means an index which is never used has costs
- Indexes which are used push other stuff out of ram cache. If you had a a z5 index, it would be 707MB, and all of its contents would be in the z6 index.
- Index sizes need to differ by about an order of magnitude for a performance difference. Performance is more closely related to depth of index, which is logarithmic with size. If you had a 1GB and a 2GB index, both usable, everything else the same, the planner wouldn't care which was used because they're so close.

I'll run some queries to get index sizes.

pnorman:
| zoom | pixel area | Area with x1 multiplier | Index size (MB) | Area with x0.01 multiplier | Index size (MB) |
|-----:|-----------:|------------------------:|----------------:|---------------------------:|----------------:|
|    5 |    2.4E+07 |                23900000 |              44 |                     239000 |             628 |
|    6 |    6.0E+06 |                 5980000 |              90 |                      59800 |            1275 |
|    7 |    1.5E+06 |                 1490000 |             218 |                      14900 |            2312 |
|    8 |    3.7E+05 |                  373000 |             494 |                       3730 |            3629 |
|    9 |    9.3E+04 |                   93400 |            1026 |                        934 |                 |
|   10 |    2.3E+04 |                   23300 |            1936 |                        233 |                 |
|   11 |    5.8E+03 |                    5840 |                 |                         58 |                 |
|   12 |    1.5E+03 |                    1460 |                 |                         14 |                 |

matthijsmelissen:
> However 59800 * 0.99 = 59202, so (rounding down not needed) it's not the same as 59750. But maybe this was calculated in a different way and we should now replace it with 5920200?

This was explained by @pnorman above: if we have an index for `way_area > 59750` we can quickly find all ways larger than 59750, and this is useful even if we actually want to find all ways larger than 59202 (we first use the index to quickly obtain all ways larger than 59750, and then only for these we need to check one by one if they are larger than 59202).

kocio-pl:
I admit I still don't get all the details, but I'm not interested too much in the inner details of a database. I just like to know what number should I put and where to make indexing aligned to a new accuracy?

pnorman:
It might not have been clear, but the values in my table are rounded appropriately, and not the precise values that appear in the query conditions.

> @pnorman Would this change be proper or it should be something else:

It depends what we're doing. I suspect an index for z6 and below and for z10 and below would be best. But I don't know, and that's the type of consideration that needs to go into selecting a different area cutoff threshold for performance reasons.

kocio-pl:
You have a great experience at database things, especially performance related, and I ultimately trust your choices regarding this. Given that indexes for z6+ and z10+ sound good for you, I believe we can go for that. How should they be implemented in the code?

pnorman:
Have two indexes, one for z6 with the 5980000 limit, one for z10 with the 23300 limit.

matthijsmelissen:
Yes, that's correct.

Issue #552, opened 2014-05-21
Render underground buildings different


matthijsmelissen:
We have some buildings in the database that are fully underground, such as parking garages. Due to the Dutch [BAG Import](http://wiki.openstreetmap.org/wiki/BAGimport), the number of such buildings has drastically increased. They are typically tagged with a negative layer tag: [Example](http://www.openstreetmap.org/way/270138781).

It would be good to render such buildings differently, for example with no fill and a dashed outline.

scat0324:
I came here looking for a solution to the rendering of Oxford's Radcliffe Camera building http://www.openstreetmap.org/#map=19/51.75337/-1.25390 - this is a round building in the middle of a courtyard and an iconic view of the city (http://www.geograph.org.uk/photo/801213). Showing the underground building on top of the grass means the map isn't really showing what someone would see if they visited - what's "on the ground". Could the underground buildings (with negative layer= tags) be rendered underneath the elements of the courtyard (e.g. the grass, which has layer=1 on it)?

mkoniecz:
> with negative layer= tags

Please read previous comments, especially https://github.com/gravitystorm/openstreetmap-carto/issues/552#issuecomment-53845730 and https://github.com/gravitystorm/openstreetmap-carto/issues/552#issuecomment-53864614 and https://github.com/gravitystorm/openstreetmap-carto/issues/552#issuecomment-53858762

althio:
I think the suggestion from @scat0324 can be understood as "honor the layer tags".
Not necessarily `layer<0` but rather render A above B if `A:layer>B:layer`.
Related to https://github.com/gravitystorm/openstreetmap-carto/issues/688

scat0324:
Ahh, yes @althio has it. I came here looking for underground buildings, saw the problems described with the level= tag, wondered if layer= could be dealt with instead, and should have searched again in order to discover #688. Apologies for derailing the issue with more general rendering of undergound buildings.

kocio-pl:
@althio: That's exactly what I'd like to be in #688 - layer is an abstract way of defining what is above/under, so it should be respected by software. Some features - like tunnel - needs to be rendered different and we have to define what is default in layering if a mapper omits such an information (for example B>A), but when she says A is layer 1 and B is layer 0, then we should render it like A>B, because now we know what is covered by what.

dieterdreist:
2015-02-12 17:30 GMT+01:00 kocio-pl <notifications@github.com>:

[...]

PR #2875, opened 2017-10-01
Adding rendering for amenity=parking_entrance [WIP]

### LINKED ISSUE #270: Add rendering for amenity=parking_entrance
`amenity=parking_entrance` represents an alternative to `amenity=parking` for _underground_ or _multi storey_ parking facilities. Each entrance is mapped separately using the `parking_entrance` value and all of them are added to a _site_ relation.

For the start I suggest to render all `amenity=parking_entrance` nodes same way as `amenity=parking`. This involves just simple tag parsing without any relation and can be improved later.

kocio-pl:
Resolves #270.
Related to https://github.com/gravitystorm/openstreetmap-carto/issues/552#issuecomment-331990727.

This is a draft which renders such entrances but only for `parking=underground`, because other parkings are visible in our style anyway. I plan to:
- [ ]  add the name rendering (it might be useful to know which underground object it leads to)
- [ ]  restrict rendering only to `access=yes` and `access=permissive` (there's too many of private underground parkings)

Example:
kx73oup3

### DIFF
--- a/project.mml
+++ b/project.mml
@@ -1427,6 +1427,7 @@ Layer:
             tags->'power_source' as power_source,
             tags->'icao' as icao,
             tags->'iata' as iata,
+            tags->'parking' as parking,
             CASE WHEN shop IN ('supermarket', 'bag', 'bakery', 'beauty', 'books', 'butcher', 'clothes', 'computer',
                                'confectionery', 'fashion', 'convenience', 'department_store', 'doityourself', 'hardware', 'fishmonger', 'florist',
                                'garden_centre', 'hairdresser', 'hifi', 'ice_cream', 'car', 'car_repair', 'bicycle', 'mall', 'pet',
@@ -1450,6 +1451,7 @@ Layer:
             OR historic IN ('memorial', 'monument', 'archaeological_site')
             OR tags->'memorial' IN ('plaque')
             OR highway IN ('bus_stop', 'elevator', 'traffic_signals')
+            OR tags->'parking' IN ('underground')
             OR (power = 'generator' AND (tags @> '"generator:source"=>wind' OR tags @> 'power_source=>wind'))
           ORDER BY way_area desc
         ) AS amenity_points_poly
@@ -1492,7 +1494,7 @@ Layer:
                                                   'police', 'post_box', 'post_office', 'pub', 'biergarten', 'recycling', 'restaurant', 'food_court',
                                                   'fast_food', 'telephone', 'taxi', 'theatre', 'toilets', 'drinking_water',
                                                   'prison', 'hunting_stand', 'nightclub', 'veterinary', 'social_facility',
-                                                  'charging_station', 'arts_centre', 'ferry_terminal') THEN amenity ELSE NULL END,
+                                                  'charging_station', 'arts_centre', 'ferry_terminal', 'parking_entrance') THEN amenity ELSE NULL END,
               'emergency_' || CASE WHEN tags->'emergency' IN ('phone') THEN tags->'emergency' ELSE NULL END,
               'shop' || CASE WHEN shop IN ('no', 'vacant', 'closed', 'disused', 'empty') OR shop IS NULL THEN NULL ELSE '' END,
               'leisure_' || CASE WHEN leisure IN ('water_park', 'playground', 'miniature_golf', 'golf_course', 'picnic_table', 'slipway',
@@ -1524,6 +1526,7 @@ Layer:
             tags->'power_source' as power_source,
             tags->'icao' as icao,
             tags->'iata' as iata,
+            tags->'parking' as parking,
             CASE WHEN shop IN ('supermarket', 'bag', 'bakery', 'beauty', 'books', 'butcher', 'clothes', 'computer',
                                'confectionery', 'fashion', 'convenience', 'department_store', 'doityourself', 'hardware', 'fishmonger', 'florist',
                                'garden_centre', 'hairdresser', 'hifi', 'ice_cream', 'car', 'car_repair', 'bicycle', 'mall', 'pet',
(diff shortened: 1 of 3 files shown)


project.mml
```diff
@@ -1427,6 +1427,7 @@ Layer:
             tags->'power_source' as power_source,
             tags->'icao' as icao,
             tags->'iata' as iata,
+            tags->'parking' as parking,
```
pnorman: What does another column do to the number of combinations, and does it cause problems with a combinatorial explosion?

project.mml
```diff
@@ -1457,6 +1458,7 @@ Layer:
             OR historic IN ('memorial', 'monument', 'archaeological_site')
             OR tags->'memorial' IN ('plaque')
             OR highway IN ('bus_stop', 'elevator', 'traffic_signals')
+            OR tags->'parking' IN ('underground')
```
pnorman: I don't see that we need to change the WHERE clause

project.mml
```diff
@@ -1499,7 +1501,7 @@ Layer:
                                                   'police', 'post_box', 'post_office', 'pub', 'biergarten', 'recycling', 'restaurant', 'food_court',
                                                   'fast_food', 'telephone', 'taxi', 'theatre', 'toilets', 'drinking_water',
                                                   'prison', 'hunting_stand', 'nightclub', 'veterinary', 'social_facility',
-                                                  'charging_station', 'arts_centre', 'ferry_terminal') THEN amenity ELSE NULL END,
+                                                  'charging_station', 'arts_centre', 'ferry_terminal', 'parking_entrance') THEN amenity ELSE NULL END,
```
pnorman: Are there other cases where we've added something to this query, but not the one above?

project.mml
```diff
@@ -1564,6 +1567,7 @@ Layer:
             OR tags @> 'emergency=>phone'
             OR highway IN ('bus_stop', 'elevator', 'traffic_signals')
             OR tags @> 'ford=>yes' OR tags @> 'ford=>stepping_stones'
+            OR tags->'parking' IN ('underground')
```
pnorman: As above with the WHERE clause

project.mml
kocio-pl: I have no deeper understanding of SQL in layers - I just try to make it work and look similar to other features. Some guidelines or tutorial for adding things would be welcome.

What you think should be done here?

project.mml
kocio-pl: What do you suggest tu put here?

project.mml
kocio-pl: How should I correct it?

kocio-pl:
> The icon looks okay, but I haven't reviewed it for pixel alignment.

I've tried to make "P" 1 px wide where possible, but that make it hardly visible, so now it's mostly rescaled parking icon. Arrow uses points on the matrix, but since it's diagonal, I would rather not call it pixel aligned.

pnorman:
You're going to need to rebase the SQL changes since I merged #2866, but that makes the changes simpler.

With those changes, `amenity=parking_entrance` meets the WHERE clause conditions and no changes need to be made to that part of the SQL.

project.mml
pnorman: https://vimeo.com/68093876 talks about the problem starting at 10:52-14:21. The best way to check this isn't a problem here is to compare the line count of the XML output. If it increases by a dozen lines, that's fine. If it increases by 10000 lines, it's a problem.

I **suspect** this is okay, but I'm not 100% sure. If you added something that styled **all** objects with some parking value then it would definitely be a problem.

kocio-pl: Most of the code I have ever added to osm-carto was some SVG XML for icons, so that's probably not the case for me.

I understand the combinatorial explosion principle, my problem is rather what you can see at 16:40. Some common layers are so complex, that they don't even fit on the screen, so it's hard to know which visual section is needed and how should it look like. 

I refer to "visual sections", because with so many elements it's hard to see the logical structure and know what is nested inside something other - even with all the indents. Maybe putting some of the lists into variables or including them from other files would help, if it's possible?

project.mml
kocio-pl: You mean `amenity-points` but not `amenity-points-poly`? I don't know, I just saw on the wiki that it's allowed only on nodes. I'm still undecided however if should I add it to `amenity-line`, since it makes sense to tag the service highway.

What's your opinion?

Tomasz-W:
As I understand, it's very close to merge this pull request. It would be very good to have these parking entrances on the map. Could someone finish it? :)
@kocio-pl @pnorman ?

matthijsmelissen:
@kocio-pl Are you planning to continue working on this?

kocio-pl:
Yes, I can resolve conflicts soon, but there are still some unanswered questions and that's why this code is still waiting.

kocio-pl:
So much better - thanks!

project.mml
pnorman: Nothing.

pnorman:
Questions answered, but we still need to know the change in lines of XML so we know if it's causing a combinatorial problem or not.

polarbearing:
The icon should be transparent when an access tag is set. Probably just copy the two lines as used for parking.
SVG code can be optimised by stripping the metadata, I usually do that in vi or another text editor.

kocio-pl:
I prefer not to show any of the private entrances - it's too many of them and they would clutter the view.

polarbearing:
> not to show any of the private entrances

Both (transparent or not at all) is fine with me. Just did not see that in the code yet.

HolgerJeromin:
what about rendering private ones really late?

kocio-pl:
I'm suspicious, but not completely against. z19+ might work - it just needs testing to make sure.

johsin18:
I would love to see parking entrances rendered.  How can I help to get this PR merged?

kocio-pl:
Could you review code comments and probably make a fork with a proper code I could test?

project.mml
johsin18: Where can I find those generated XML files, in particular having a setup closely following
?

project.mml
johsin18: Indeed, this case should be already covered by "OR amenity IS NOT NULL"

PR #2883, opened 2017-10-07
Better minimal distance between housenumbers

sommerluk:
Better minimal distance between housenumbers

The current distance is too small. Sometimes you cannot distinguish the individual housenumbers on est-west-roads (see bottom-right on the screenshot).

Before:
z_1_before_screenshot 6

After:
z_3_after_screenshot 5

### DIFF
--- a/addressing.mss
+++ b/addressing.mss
@@ -26,7 +26,6 @@
       }
     }
     text-placement: interior;
-    text-min-distance: 1;
     text-face-name: @book-fonts;
     text-fill: @address-color;
     text-halo-radius: @standard-halo-radius;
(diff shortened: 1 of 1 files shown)


kocio-pl:
The issue seems to be real to me and the solution looks convincing on the attached rendering, I'm just not sure what these changes mean - could you explain the code in this PR? Have you also considered what could be possible downsides or you don't see any?

sommerluk:
The downside is obviously that there will be some housenumbers that will only be rendered on higher zoom levels than before (when there’s enough place). But I think that won’t be much, and it’s worth the pain for the sake of legibility.

In the screenshot:
- Bottom right: The 316 is not rendered anymore on this zoom level (but only on the next higher zoom level).
- Top left: 68 is not rendered anymore. Instead 72 is rendered. So one housenumber we have lost, one housenumber we have won.

I’ve opted for changing `text-margin` to `text-min-distance`  because `text-margin` is marked as deprecated.

PR #2885, opened 2017-10-08
Use Noto Sans Arabic

sommerluk:
Use Noto Sans Arabic.

With the recent release of Noto, there is now a true “Sans” (=low constrast) arabic font available. It seems Naskh-like, but because it’s Sans, it fits much better with the rest of scripts and the styling is nevertheless Naskh-like. It is a little bigger with more spacing, what should improve legibility:

Before:
screenshot 1

After:
screenshot 2

As this release is not yet available as Ubuntu package (and therefor OSM servers unlikely to provide it right now), the current Naskh style is kept as fallback.

This PR contains now also code for Bold variant of Arabic and fallback.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -55,7 +55,7 @@ Put these shapefiles at `path/to/openstreetmap-carto/data`.
 ## Fonts
 The stylesheet uses Noto, an openly licensed font family from Google with support for multiple scripts. The stylesheet uses Noto's "Sans" style where available. If not available, this stylesheet uses another appropriate style of the Noto family. The "UI" version is used where available, with its vertical metrics which fit better with Latin text.
 
-DejaVu Sans is used as an optional fallback font for systems without Noto Sans. If all the Noto fonts are installed, it should never be used.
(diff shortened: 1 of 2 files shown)


kocio-pl:
It looks a bit better, indeed, so I would be happy to have this patch.

kocio-pl:
Any other comments or reviews? Fonts are not something I can deal with, that's why I can't make a proper review, but I trust Lukas with such changes if the output is OK for me.

matkoniecz:
Unfortunately I am unable to judge readability of arabic text.

kocio-pl:
For me that's about shapes being more contrast and bigger, but it's always good to ask some users that helped us with comments before (#2496):
@aso-mehmudi @lyxys @BoFFire @saeedghazijoolaee What do you think about this font change?

sommerluk:
http://www.openstreetmap.org/#map=14/36.2941/59.6006

Left: Current rendering at openstreetmap.org
Right: This PR
screenshot 1

Note that subway stations are bolder with this PR because our current code did not have support for bold for Arabic script.

PR #2890, opened 2017-10-14
Show island and islet labels earlier

### LINKED ISSUE #1448: How to treat island labels on z4–z6?
Island labels start currently at z7.

@imagico had proposed to render it starting from z4. This would render labels on some very big islands, e.g. Spitzbergen.

Currently, for low zoom levels, this feature is part of the layer with the id: "text-poly-low-zoom". This layer renders only z7–z9. So I suppose when we want this, we need a new layer? How would that influence the performance? How is the policy for on low zoom levels? Would this interfer with other things on low zoom levels?

Would it make sense to render place=archipelago also? (But is not supported by osm2pgsql? 134 relations with this attribute.)

kocio-pl:
Resolves #1448.

Island names are currently rendered on z7+, but they should be rendered on z4+ (as early as possible, but not compete with country labels on z3) and islets also 3 zoom levels earlier.

@rrzefox: Could you test this patch on your server?

Nunavut (Canada) example

z4
Before
screenshot-2017-10-14 openstreetmap carto kosmtik
After
screenshot-2017-10-14 openstreetmap carto kosmtik 1

### DIFF
--- a/amenity-points.mss
+++ b/amenity-points.mss
@@ -1092,9 +1092,9 @@
 /* Note that .text is also used in water.mss */
 .text-low-zoom[zoom < 10],
 .text[zoom >= 10] {
-  [feature = 'place_island'][zoom >= 7][way_pixels > 3000][way_pixels < 800000],
+  [feature = 'place_island'][zoom >= 4][way_pixels > 3000][way_pixels < 800000],
   [feature = 'place_island'][zoom >= 16][way_pixels < 800000],
-  [feature = 'place_islet'][zoom >= 14][way_pixels > 3000][way_pixels < 800000],
+  [feature = 'place_islet'][zoom >= 11][way_pixels > 3000][way_pixels < 800000],
   [feature = 'place_islet'][zoom >= 17][way_pixels < 800000] {
     text-name: "[name]";
     text-fill: #000;


sommerluk:
Looks good to me. That includes the change for islet. Islet names only showed up only on high zoom levels, because of the 1-km²-threshold in the wiki documentation, but the chosen zoom level (14) did only work well on low latitude. For high latitude like http://www.openstreetmap.org/way/27923209#map=13/83.6622/-30.6165 clearly zoom level 11 is a better threshold.

kocio-pl:
Rendering the planet with just the island names. This should help us find all the missing tagging - for example Spitsbergen should be fixed now (the planet data are older), but Sumatra exist only as [an islet in Dubai](http://www.openstreetmap.org/way/87186636#map=14/25.2233/55.1750) and we lack New Guinea and Greenland for sure.

The list to check:
https://en.wikipedia.org/wiki/List_of_islands_by_area#Islands

z4
island4-gimp

z5
island5-gimp

z6
island6-gimp

rrzefox:
> @rrzefox: Could you test this patch on your server?

It's been deployed. Comparison as usual on http://bl.ocks.org/math1985/raw/af7a602c222dbf1ff1a2c0d84ed755b7/#5.00/72.344/-97.588
Z0-9 rerendered, 10-14 still pending.

PR #2891, opened 2017-10-14
Unified rendering of leisure=fitness_station and leisure=fitness_centre

### LINKED ISSUE #1870: Render leisure=fitness_station

### LINKED ISSUE #2375: Add leisure=fitness_centre icon/text

kocio-pl:
Resolves #1870
Resolves #2375

1. Unifies both fitness amenities, because the main difference is if they are inside building or in the outdoor, which is already visible from the context. 
2. Uses general activity icon from shop=sports, but with a leisure green color and mirrored to make it even more distinctive.
3. Also makes fitness station close to the other small leisure amenities (like playground).

Example z19 rendering

fitness centre 
Before
reiw6yed

After
zeiwj02u

fitness station
Before
0nhpq7p3

After
4a24_vkc


boothym:
Not sure if the fitness centre colour needs to be changed, it should remain the same as sports centre.

Will this render fitness station nodes as well?

kocio-pl:
> Not sure if the fitness centre colour needs to be changed, it should remain the same as sports centre.

The only good thing about this style was it was rendered at all. If the name dosn't contain "fitness" or you don't know the script or language, it's not clear of this is fitness centre, school or anything like that.

> Will this render fitness station nodes as well?

Yes. That's why we need the icon.

andrewharvey:
Looks good to me.

PR #2892, opened 2017-10-15
Add rendering for military=bunker

kocio-pl:
Adding rendering for popular military/historic POI military=bunker - it has 35 251 uses and a [wiki page](https://wiki.openstreetmap.org/wiki/Tag%3Amilitary%3Dbunker).

Reusing cave entry icon, but with amenity brown color, because it looks like a solid shelter. Examples:

e mj9dym
pz4dsw1b

### DIFF
--- a/amenity-points.mss
+++ b/amenity-points.mss
@@ -970,6 +970,13 @@
     marker-placement: interior;
     marker-clip: false;
   }
+  
+  [feature = 'military_bunker'][zoom >= 17] {
+    marker-file: url('symbols/bunker.svg');
+    marker-fill: @man-made-icon;    
+    marker-placement: interior;
+    marker-clip: false;
+  }  
 
   [feature = 'natural_spring'][zoom >= 14] {
     marker-file: url('symbols/spring.svg');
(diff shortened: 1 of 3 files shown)


dieterdreist:
2017-10-15 3:53 GMT+02:00 kocio-pl <notifications@github.com>:

> Reusing cave entry icon, but with amenity brown color, because it looks
> like a solid shelter. Examples:



-1 to using the same icon for natural=cave_entrance and military=bunker
the distinction between brown and black is too subtle.
What about making it with corners rather then round?

matkoniecz:
@dieterdreist I am not sure how reply via email works (I think that you are using this), but your reply was hidden by github.

> What about making it with corners rather then round?

I think that it may work. Current is really, really close to how cave is displayed.

kocio-pl:
You're right, the color difference is too subtle. I've made a special, but similar design then, but this time with man made icon color:

eotghhl_
kreo4fhk

This is how it looks like on buildings ("museum" is just tagging for rendering here):
2mibtt75

kocio-pl:
>I am not sure how reply via email works (I think that you are using this), but your reply was hidden by github.

Maybe because the reply is formatted the same as original text, which is hidden by default.

matkoniecz:
It looks much better. It is really strong (maybe to during and too prominent), but now it is not looking like a cave variant.

kocio-pl:
Any hints what exactly could be improved?

matkoniecz:
Note that it would require testing to check is it really  necessary.

But I would think about making icon smaller, making gap between center and line bigger, making center smaller, making line narrower.

kocio-pl:
How about this?

aie12hxg
tmxccik3

matkoniecz:
As always I worry about adding Yet Another Thing To Render, but from examples new version seems to be fine.

math1985:
Approving as in the last image.

PR #2893, opened 2017-10-15
Marketplace rendering

### LINKED ISSUE #751: Add rendering for amenity=marketplace
It would be nice to get rendering for “amenity=marketplace” with an icon and its name.

Some facts:

– 22 827 objects with “amenity=marketplace” in the database

– Wiki says to use it only an areas, not on nodes. But taginfo say that nearly 50% of the usage is on nodes (and the other 50% on areas).

– Is often used together with “building=yes” (20% of total usage) and “landuse=retail” (10% of total usage)

sommerluk:
Resolves #751 using icon number 5.

### DIFF
--- a/amenity-points.mss
+++ b/amenity-points.mss
@@ -588,6 +588,14 @@
     marker-clip: false;
   }
 
+  [feature = 'amenity_marketplace'][zoom >= 16][way_pixels > 3000],
+  [feature = 'amenity_marketplace'][zoom >= 17] {
+    marker-placement: interior;
+    marker-clip: false;
+    marker-fill: @shop-icon;
+    marker-file: url('symbols/shop/marketplace.svg');
+  }
+
   [feature = 'shop'] {
     [shop != 'mall'][zoom >= 17],
     [shop = 'supermarket'][zoom >= 16],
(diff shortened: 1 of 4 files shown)


kocio-pl:
Please attach real life rendering example (one or more, if this shows something interesting).

sommerluk:
screenshot 14
screenshot 15
screenshot 16
screenshot 17
screenshot 18
screenshot 19

screenshot 3
screenshot 4
screenshot 5
screenshot 6

dieterdreist:
yes, markets are very important and prominent features, typically, are known by most residents and can serve for orientation purposes/way descriptions.
I don’t think an icon is important though, the important part is the name, and this rendering proposal doesn’t show the name well, it prioritizes the icon and the shop dots over the market label, while I believe it should be the other way round

sommerluk:
> it prioritizes […] the shop dots over the market label

Well, that’s not the goal. But in the first example, as there are many other shop dots, that is our standard behaviour when there is not enough space: Something fails to render (normaly rather the text and not the icon). I’m not sure if there is much we can do about. It isn’t a typical example of marketplace tagging however.

More typical is the next example “Marché de Bradoci”, where the name renders well.

The last example has no “name” tag in the database (at least, it had no name tag when I imported the database.)

I think it is useful to have an icon, also because ¼ of the marketplaces in OSM don’t have a name.

kocio-pl:
> this rendering proposal doesn’t show the name well, it prioritizes the icon and the shop dots over the market label, while I believe it should be the other way round

It's a more general problem with generalization. We have no way to detect that grass is just a part of the park (so should not be visible when zooming out), the same for shops inside the market and cemetery paths - all of them can make visual noise when only "master area" should be visible.

I wanted to test Norwich Market because it's very hard case (both shops and paths are tagged extensively), yet on z17 it works reasonably well.

PR #2894, opened 2017-10-15
Render ele also on polygon alpine_hut and shelter

### LINKED ISSUE #101: ele with alpine_hut when mapped as a building

sommerluk:
Resolves #101

### DIFF
--- a/project.mml
+++ b/project.mml
@@ -1943,7 +1943,15 @@ Layer:
               'tourism_' || CASE WHEN tourism IN ('viewpoint', 'attraction') THEN tourism ELSE NULL END
             ) AS feature,
             access,
-            name,
+            CONCAT(
+                name,
+                CASE
+                  WHEN (tourism = 'alpine_hut' OR amenity = 'shelter') AND tags->'ele' ~ '^-?\d{1,4}(\.\d+)?$' THEN
+                    CONCAT(E'\n', REPLACE(ROUND((tags->'ele')::NUMERIC)::TEXT, '-', U&'\2212'), U&'\00A0', 'm')
+                  ELSE
+                    NULL
+                END
+            ) AS name,
             tags->'operator' as operator,
             tags->'icao' as icao,
             tags->'iata' as iata,


kocio-pl:
Could you paste this rendering?

sommerluk:
It’s the same rendering as for points because it gets rendered by the same mss code:

screenshot 2

(At the left a test with an invalid “ele” value).

kocio-pl:
> (At the left a test with an invalid “ele” value).

Is there a way to not display elevation in such case? 

I guess this problem is the same for peaks etc?

sommerluk:
Well, that is what this code is doing 😉

The objects are just dummy test objects.

The tagging for the object left top (node) is:
```
name=alpine_hut point invalid
ele=12a34.5
```

Left bottom (closed way):
```
name=alpine_hut poly invalid
ele=12a34.5
```

So indeed the invalid ele value is not rendered. This is particularly important because otherwise, the conversion would fail, this means we would have an SQL error and the hole rendering would be broken.

This code should be functionally equivalent to the existing code for points – so for polygons it should behave in all aspects the same way as for points.

kocio-pl:
I meant not rendering also such debug info... :smile:

sommerluk:
To explain further:

This code for polygons does _not_ include rendering for the elevation of peaks and saddles (while the code for points does). This is because the wiki does not allow peaks and saddles and volcanos on ways, so no need to support this.

kocio-pl:
I wouldn't like to see error messages also for points (I see "point invalid"), hiding it would be nice everywhere we try to show ele.

sommerluk:
To be clear: There _is no error message_ anywere.

The screenshot is from a test rendering. I just had created objects to which I have given the name “amenity_hut point invalid” etc. because this way, I can easily recognize and remember, which object had which tags. It renders “amenity_hut point invalid” because this is the _name_ tag in the test database, _not_ because it is a generated error message.

Everything works fine.

PR #2903, opened 2017-10-23
Introduce text-wrap for stations

math1985:
This enables text-wrap for station labels, in line with other POI labels.

Before:
screen shot 2017-10-24 at 00 20 55

After:
screen shot 2017-10-24 at 00 20 48

### DIFF
--- a/stations.mss
+++ b/stations.mss
@@ -37,7 +37,8 @@
       text-dy: 9;
       text-halo-radius: @standard-halo-radius * 1.5;
       text-halo-fill: @standard-halo-fill;
-      text-wrap-width: 0;
+      text-wrap-width: @standard-wrap-width;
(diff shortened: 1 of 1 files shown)


sommerluk:
If using variables for wrap width and line spacing, I would also use variables for the text size:   `@standard-font-size`  …

kocio-pl:
Looks sane to me, I need to test it a bit however.

kocio-pl:
I'm testing without Docker. It's the most important fix on z14 in Paris and in Manhattan (very dense subway network), but looks also good in London and Warsaw.

New York, z14:
Before
syac7wzx
After
l3tdnwpw

matkoniecz:
@sommerluk I created #2918 for your suggestion.

PR #2912, opened 2017-11-01
Issue #2911: Devours all RAM and computer hangs on 'docker-compose up import'

### LINKED ISSUE #2911: Devours all RAM and computer hangs on 'docker-compose up import'
I decided to give a try to docker images in the repo, but it ended up unhappily. Whenever I run `docker-compose up import`, after the message that it loads `data.osm.pbf` it just consumes all the RAM (16GB) so that I had to restart the computer. 

```
$ docker-compose up import
Starting openstreetmapcarto_db_1 ...
Starting openstreetmapcarto_db_1 ... done
Starting openstreetmapcarto_import_1 ...
Starting openstreetmapcarto_import_1 ... done
Attaching to openstreetmapcarto_import_1
import_1   | Waiting for PostgreSQL to be running
import_1   | PostgreSQL running
import_1   | CREATE EXTENSION
import_1   | NOTICE:  extension "postgis" already exists, skipping
import_1   | CREATE EXTENSION
import_1   | NOTICE:  extension "hstore" already exists, skipping
import_1   | osm2pgsql SVN version 0.88.1 (64bit id space)
import_1   |
import_1   | Using lua based tag processing pipeline with script openstreetmap-carto.lua
import_1   | Using projection SRS 900913 (Spherical Mercator)
import_1   | Setting up table: planet_osm_point
import_1   | Setting up table: planet_osm_line
import_1   | Setting up table: planet_osm_polygon
import_1   | Setting up table: planet_osm_roads
import_1   | Allocating memory for dense node cache
import_1   | Allocating dense node cache in one big chunk
import_1   | Allocating memory for sparse node cache
import_1   | Sharing dense sparse
import_1   | Node-cache: cache=1000MB, maxblocks=128000*8192, allocation method=3
import_1   | Mid: Ram, scale=100
import_1   |
import_1   | Reading in file: data.osm.pbf
# Here it stops... 
```

I tried to change osm2pgsql settings:

```
      - PG_WORK_MEM
      - PG_MAINTENANCE_WORK_MEM
      - OSM2PGSQL_CACHE
      - OSM2PGSQL_NUMPROC
```

but even with default options it just ends like that. 

My data.osm.pbf file is approximately 500Mb of size.

OnkelTem:
Resolves #2911

### DIFF
--- a/Dockerfile.import
+++ b/Dockerfile.import
@@ -1,10 +1,13 @@
-FROM ubuntu:xenial
+FROM debian:stretch
 
-RUN apt-get update && apt-get install --no-install-recommends -y \
-    curl ca-certificates osm2pgsql postgresql-client \
-    && rm -rf /var/lib/apt/lists/*
+# Install tools, utilities and osm2pgsql
+RUN set -x && \
+    DEBIAN_FRONTEND=noninteractive && \
+    apt-get update && \
+    apt-get install -y --no-install-recommends osm2pgsql postgresql-client && \
+    rm -rf /var/lib/apt/lists/*
 
-ADD openstreetmap-carto.style /
+COPY openstreetmap-carto.style /
 
 RUN mkdir -p /openstreetmap-carto
 WORKDIR /openstreetmap-carto


matkoniecz:
Thank you very much for providing PR! Do you have any idea what is fixed here? I admit that for me it is not clear why with old code docker was not working and with new one it is fine.

OnkelTem:
@matkoniecz I think it was due to some bug in osm2pgsql 0.88 - it started reading file and went into infinite loop consuming RAM. So I just tested first on custom build of 0.94 (latest) and then tried 0.92 from debia:stretch and it just worked.

Dockerfile.import
```diff
@@ -1,10 +1,13 @@
-FROM ubuntu:xenial
+FROM debian:stretch
```
matkoniecz: Have you tested whatever switching to http://releases.ubuntu.com/17.04/ also fixes problem?

Note that https://github.com/gravitystorm/openstreetmap-carto/pull/2818/files switched from debian:stretch to ubuntu:xenial, I worry a bit about repeated PRs changing between ubuntu and debian, fixing one particular issue and breaking setup for somebody else.

I think that upgrading OS version would be preferable to repeated random switches (#2818, now this) to fix one particular issue.

SomeoneElseOSM: For info I tried the "standard" switch2osm guide on Ubuntu 17.04 and hit a problem with python-mapnik

    python
    >>> import mapnik
      File "<stdin>", line 1, in <module>
      File "/usr/lib/python2.7/dist-packages/mapnik/__init__.py", line 74, in <module>
        from ._mapnik import *
    ImportError: /usr/lib/python2.7/dist-packages/mapnik/_mapnik.x86_64-linux-gnu.so: undefined symbol: _ZTIN6mapnik15freetype_engineE

so there may be a separate issue with 17.04.

OnkelTem: No, I haven't tested this on 17.04. As for #2818 I'm not sure about the reason - it doesn't seem to fix anything and just switches back to an older system.

matkoniecz: Maybe it either forced purging caches (so PR was useless) or there was a regression somewhere so updating software caused bug to appear?

matkoniecz: "No, I haven't tested this on 17.04" - can you do that? Hopefully it will not run into @SomeoneElseOSM issue and ubuntu.debian dance will be avoided.

OnkelTem: @matkoniecz Would you please clarify, do you mean trying to switch from `debian:stretch` to `ubuntu:zesty` in the kosmtik Dockerfile?

matkoniecz: I think it is preferable to switch from `ubuntu:xenial` to `ubuntu:zesty` rather than from `ubuntu:xenial` to `debian:stretch` (smaller change, upgrading OS is less likely to cause regressions that changing OS).

OnkelTem: @matkoniecz Sure I can, while I don't see much sense in this - it's an isolated environment, who really cares about what's inside? IMHO there could be two main reasons to start changing base image:
- it's when something doesn't work (#2911)
- saving space - hence switching to the same base image in all used images

matkoniecz: @OnkelTem It is to avoid regressions. Switching debian:stretch to ubuntu:xenial (#2818) only to switch it back here is not useful. Especially as it is not clear what is underlying bug - both here and in #2818.

kocio-pl:
I've made a distro change exactly to make a difference big enough, because I hoped the bug won't be here (a big bisection). I could have checked more solutions, unfortunately rebuilding Docker image is time consuming and Ubuntu worked for me, so I took that way.

It's hard for me to say which distro would be better as a base for this image. I like Debian as closer to the source, on the other hand Ubuntu is what I'm more familiar with (this is a distro I use a lot) and the release time scheme is more predictable.

However it looks like package changes can influence image a lot, so it would be good to make some releases to have it "frozen" and reasonably tested. Another solution could be freezing distro on some exact point and not let it to be updated with more current packages.

OnkelTem:
@kocio-pl
> Another solution could be freezing distro on some exact point and not let it to be updated with more current packages.

Is there a way to freeze a distro?

We can osm2pgsql and Co from source to loose dependency from a distro.

kocio-pl:
I'm not sure - it's a general idea. Of course it depends on base image stability (it can have the same name like "xenial", but in reality change from 16.04.1 to 16.04.3) and files updating (we could disable it by commenting out sources it in /etc/apt/*). But nothing beats having our binary images in a Docker hub.

kocio-pl:
Any new thoughts about this issue?

pnorman:
> We can build osm2pgsql and Co from source to loose dependency from a distro.

No, this is not recommended if a reasonably recent version is available for a distro.

kocio-pl:
I think this problem needs more specific solution (especially dealing with some kind of "freezing"), therefore I close this PR. Thank you for the work, however!

PR #2914, opened 2017-11-01
Refactor docker structure [docker]

### LINKED ISSUE #2913: Refactor docker structure [docker]
Well, let me share my thoughts about how the whole thing can be updated.

* Move docker stuff into a separate directory and refine overall structure
* Push docker bootstrap scripts inside docker images
  * Split `scripts/docker-startup.sh`
* Make it possible to configure user id from inside `docker-compose.yml`
* Create regular volume for database so that container termination wouldn't kill (precious!) data.
* Add missing fonts for mapnik into kosmtik image
* Update `docker-compose.yml` version format to '3.2'.

I'm creating a branch in my fork with those changes.

OnkelTem:
Resolves #2913


kocio-pl:
If you edit it into "Resolves #2913" the issue will be automatically closed if the PR will be merged. The same for other PRs, like #2912.

matkoniecz:
It seems that docker documentation at https://github.com/OnkelTem/openstreetmap-carto/blob/master/DOCKER.md should be updated as part of this PR.

```
mateusz@grisznak:~/Desktop/kolejka/openstreetmap-carto$ sudo docker-compose up import
ERROR: 
        Can't find a suitable configuration file in this directory or any
        parent. Are you in the right directory?

        Supported filenames: docker-compose.yml, docker-compose.yaml
        
mateusz@grisznak:~/Desktop/kolejka/openstreetmap-carto$ cd docker/
mateusz@grisznak:~/Desktop/kolejka/openstreetmap-carto/docker$ sudo docker-compose up import
docker_db_1 is up-to-date
Starting docker_import_1 ... 
Starting docker_import_1 ... done
Attaching to docker_import_1
import_1   | sh: 0: Can't open scripts/docker-startup.sh
docker_import_1 exited with code 127
```

OnkelTem:
Thanks for testing.
Yes, I didn't update the README - was waiting for someone to test the changes.
And  `:v1` should be changed or it would trigger old images. 
I'll update.

OnkelTem:
Ok, made some fixes.

kocio-pl:
1. Documentation should include latest #2916 changes.
2. After the code refactoring data should be downloaded into `docker` directory, because import tool expects that that, so it also means some documentation changes.
3. It's also not clear that one should run docker-compose commands in a `docker` directory.

kocio-pl:
@OnkelTem Do you still want to work on this PR?

kocio-pl:
OK, just checking. :smile:

OnkelTem:
Hi there. Sorry for the delay, time is running and had a lot of work on my main job :)

@kocio-pl My request is a bit unusual. Can you take a look at my other work on the tile server which I announced in #2928? I remember your and folks' arguments against it and the main one was that you have to maintain more unrelated stuff. But what if you create a separate branch with those experiments and let users to play with it and if it gets some positive feedback you'll consider merging it to the master? 

I can explain. When I began to compare this PR with my last work on the tile server I see a lot of changes I man and which I don't really to roll back. For me it's even easier to exclude (comment out?) tile server from docker-compose if you don't want it so much then just continue work on this PR.

Also I remember you proposition on creating a toolbox, but honestly at the particular moment I'd like to get osm carto working correctly first and only then continue to the toolbox.

Can I create another PR with new architecture of the docker stuff?

kocio-pl:
Currently I don't have too much time... But if you make a branch it will also diverge soon and it can be hard to merge it back.

OnkelTem:
Ok, I've finished. The final product is not actually a refactor or something - it's more a new thing. For this reason I decided to publish it in as a separate project: https://github.com/OnkelTem/osmcarto-docker
It doesn't use docker-compose tool directory anymore and is based on [Dobi](https://github.com/dnephin/dobi).

It would be great if you could announce it somehow to let guys test it.

And this PR I propose to close since it's not relevant anymore.

kocio-pl:
> Ok, I've finished. The final product is not actually a refactor or something - it's more a new thing. 

Great to hear it! 

Do you plan to add other container types, as I was proposing? Current naming shows that it will be osm-carto specific, not the universal OSM containers toolset.

> It doesn't use docker-compose tool directory anymore 

I'm confused, since Docker-compose is mentioned as needed:
https://github.com/OnkelTem/osmcarto-docker#prerequisites

> and is based on Dobi.

I also think that using a new, not standard tool which is "still in early development" will make entry cost higher than it's really needed.

> It would be great if you could announce it somehow to let guys test it.

I think Talk list could be good.

> And this PR I propose to close since it's not relevant anymore.

OK, so I'm closing it now.

OnkelTem:
> I'm confused, since Docker-compose is mentioned as needed

Yeah, that's right, I meant that you don't run it directly in the command line. It's entirely superseded with Dobi, which simply requires the former.

> I also think that using a new, not standard tool which is "still in early development" will make entry cost higher than it's really needed.

That's true, but really - you can't get much from docker-compose :) For the last year I spent hundreds of hours building development environments on it. 

The main problem with Docker Compose's usability is that it's really bad in running *tasks*, not *services*. You do can run tasks with `run` but that requires from the end-users to see this difference too. You can't use project name automatically, so you always need to provide it in command line with `-p` option or your containers and networks and volumes won't be prefixed. It's unable to acquire `UID` automatically from the shell, so you have to provide this via parameters too or just edit your `docker-compose.yml` by hand. A lot of buts. That's why in my projects I just use bash scripts to wrap its s**t :)

And now Dobi. While it's definitely not an ideal tool (I've almost eaten off the brain of its developer Daniel to make Dobi better!) it's anyway more relevant for the things we usually do. Basically, it allows us to declare and run tasks. In our particular case those are: get shapefiles, get fonts, import osm data into the database. It supports dependencies, it can build and run images, override commands and entrypoints, declare and reuse volumes and much more.  Unfortunately Dobi lacks the *service* concept, but this can be worked around by *compose* *resource* type (and thus - dependency from docker compose).

> Do you plan to add other container types, as I was proposing? Current naming shows that it will be osm-carto specific, not the universal OSM containers toolset.

Sure I have such plans, but I'd like to see this in action first: collect some feedback, thoughts and ideas. I'm not sufficiently familiar with what you folks really use for mapping. But I'd gladly extend it and re-architecture when some new information would come.

PR #2916, opened 2017-11-02
be more explicit in docker quick start

matkoniecz:
include also relatively obvious git command
for anybody experienced in using git it would be obvious,
but it is preferable to make it as explict and as easy to follow as possible

### DIFF
--- a/DOCKER.md
+++ b/DOCKER.md
@@ -24,6 +24,10 @@ Read on below to get the details.
 * Ctrl+C to stop the style preview application
 * `docker-compose stop db` to stop the database container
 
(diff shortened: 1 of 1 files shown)


kocio-pl:
It would be good to note that this is when you want to test current master branch, not your own fork/branch.

matkoniecz:
I added section that should help people starting with git - it is basically link to an external documentation.

matkoniecz:
In general - I think that it is not a place for a git tutorial, but linking existing documentation for even simpler things is a good idea.

PR #2921, opened 2017-11-02
Create USECASES.md

math1985:
Brain storm area for usecases for the low zoom levels. Hopefully this will help with #2688.

### DIFF
--- a/USECASES.md
+++ b/USECASES.md
@@ -1,0 +1,17 @@
+This file describes some use cases of the style sheet. For each use case, it is indicated which features we need to render because they are the subject of the use case, and which features we should render as context for the subject feature(s). Also, for each use case is indicated at what zoom level the use case is relevant.
+
+The end goal should be that for everything we render on a given zoomlevel, we have a use case that mentions that feature either as map subject or map context on that zoomlevel. We can accomplish this either by adding use cases, or removing features.
+
+For now, this table covers zoom levels 5, 6 and 7.
+
+|Use case|Map subject|Map context|Zoom level|
+|---|---|---|---|---|
+|Looking up country location|Country names, country borders|Seas and oceans|z5,z6|
+|Looking up province/state location|Province/state names and borders|Country borders|z6,z7|
+|Looking up city location|City names|Seas, oceans, country names, country borders|z5,z6,z7|
+|Long distance road trip planning|Roads|City names|z6,z7|
+
+This gives us the following features:
+* z5: Map subject: Country names, country borders. Map context: seas, oceans.
+* z6: Map subject: Country names, country borders, province/state names and borders, roads, city names. Map context: Seas, oceans, country borders, city names.
+* z7: Map subject: Province/state names and borders, city names, roads. Map context: Seas, oceans, country names, country borders.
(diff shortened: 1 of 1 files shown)


matkoniecz:
Markdown table is broken in github preview - https://github.com/math1985/openstreetmap-carto/blob/c7aba5ae2c02ac9c29ffc17837555839a2999563/USECASES.md

Is it fixable?

matkoniecz:
Tables are supposed to work: see https://help.github.com/articles/organizing-information-with-tables/

math1985:
Should be fixed now!

matkoniecz:
In general I like this idea, what follows is a minor nitpicking.

> Roads

I think that it should be more specific (after all, we are not showing highway=service at z7).

>z5,z6,z7

Maybe "z5, z6, z7", or "5, 6, 7" would be more readable?

math1985:
Some features that we currently render for which we do not (yet) have a use case:

* Province/state names and borders on z5
* City names on z5
* Roads on z5
* National parks on z7
* Railways on z7
* Minor water areas on z5/6/7

Should we add use cases or drop rendering these features?

matkoniecz:
Also, it would be a good idea to include how this table is supposed to be grown - it it OK to make PRs for lower zoom levels? Higher zoom levels? Just zoom level one higher that currently processed one?

Should we have a list of rendered features not covered by any usecase?

kocio-pl:
Wouldn't it be better to use wiki page instead? For example:

https://wiki.openstreetmap.org/wiki/Standard_tile_layer/use_cases

math1985:
> minor nitpicking.

Fixed.

>  it it OK to make PRs for lower zoom levels?

I have no strong opinion, shall we first try this out for the current zoomlevels, and then extend it later if we're happy with this way of working?

math1985:
> Wouldn't it be better to use wiki page instead?

I'd prefer to keep development here and not spread out over different mediums. Also, I think it's good to keep editing restricted to the maintainers here, and accept changes from others only by accepting pull requests. Otherwise it might get an inconsistent mess quite quickly.

matkoniecz:
> Wouldn't it be better to use wiki page instead?

I think that it would preferable to

- keep this together with other map-style content (in repository)
- keep additions under review

> I have no strong opinion, shall we first try this out for the current zoomlevels, and then extend it later if we're happy with this way of working?

Any version would be OK, but it would be a good idea to mention it in document ("For now, this table covers zoom levels 5, 6 and 7." does not give info whatever at this moment PRs for additional zoom levels are considered as a good idea)

math1985:
>  it would be a good idea to mention it in document 

Done.

matkoniecz:
One more thing - maybe it should be linked from CARTOGRAPHY? And maybe also from README or CONTRIBUTING?

math1985:
Merging already so we have it within the repository. Feel free to create further PRs or suggestions.

matkoniecz:
> Merging already so we have it within the repository

Squashing on merge would be probably a good idea in this case.

math1985:
You're right, I should have done that, guess it's too late now.

PR #2926, opened 2017-11-04
mention U.S states in usecases

matkoniecz:
maybe for other countries internal division is also useful from z5 - in that case it also should be mentioned

### DIFF
--- a/USECASES.md
+++ b/USECASES.md
@@ -13,12 +13,14 @@ As we are still experimenting with this way of design, this document is restrict
 | Long distance road trip planning | Motorways | City names | z6, z7 |
 
 This gives us the following features:
+
 * z5: Map subject: Country names, country borders, city names. Map context: seas, oceans.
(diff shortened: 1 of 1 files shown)


math1985:
I think this applies not only to the US but to all countries with larger subdivisions (like Canada and Australia).

matkoniecz:
Maybe - I made PR with usecase that I know that exists. I suspect that it applies also to Russia but it is only a guess and I would wait for at least single person to confirm this.

kocio-pl:
I would rephrase it to cover all the big subdivisions (or at least mention them).

matkoniecz:
@kocio-pl I want to avoid this as it seems likely to me that in some large countries subdivisions are not really important.

I know that in USA states are important - are we sure about other countries?

kocio-pl:
OK, if _your_ use case is only USA, I have no problem with that. In reality currently USA, Australia and Canada subdivisions are visible (only Russia subdivisions are not visible on z4 and I don't understand why), so this is more a statement what is important than something to fix.

Issue #2896, opened 2017-10-15
Skipping parts of bigger objects on lower zoom levels


kocio-pl:
Is it possible to skip rendering parts of bigger objects on some zoom levels?

I meant well mapped cases like:
- paths belonging to cemetery (like [here](http://www.openstreetmap.org/way/5080251#map=13/52.2721/21.0665) - it's rather red than green, especially when comparing with another cemetery on the south):
![screenshot-2018-1-8 openstreetmap](https://user-images.githubusercontent.com/5439713/34655722-abf86d66-f40e-11e7-98a6-71b1e4d6f086.png)
- grass belonging to a park or a hospital (like [here](http://www.openstreetmap.org/way/66266360#map=14/52.1966/21.0093) - there's almost no visible yellowish background, but on z14 it should be rather visible instead of grass and park)
- shops inside a marketplace (see https://github.com/gravitystorm/openstreetmap-carto/pull/2893#issuecomment-336694603)

I doubt in any special tagging, it would be good if the software would be able to make "hard" priorities for some areas, but maybe using relations would be the most reasonable way to recognize parts in general. Would it work within our current toolbox?

kocio-pl:
Two partial solutions to make some areas more visible:

1. We could hide all the small parkings (=when they don't show the letter). Testing shows that with gray color [they become unobtrusive](https://github.com/gravitystorm/openstreetmap-carto/issues/2904#issuecomment-341896809), but they still take some space on areas like hospital, so such area is less visible than it could be.

2. Wikimedia maps use interesting approach: if there are some buildings on the colored area, they are colored the same, but darker up to z17 (on z.max=z18 they look like all the other buildings). 

Examples:
* https://maps.wikimedia.org/#17/52.22667/20.96999
* https://maps.wikimedia.org/#17/52.19953/20.99703 - grass is still unchanged, but buildings blend nicely with the background and it makes it easier to see the hospital area

I haven't yet look how it's done and if it's possible for us to do the same. The other problem is that we use dark violet for other features, so buildings on industrial area would look to close to them. However for societal amenities, commercial and garages this would help.

math1985:
> if there are some buildings on the colored area, they are colored the same, but darker up to z17 

This is the same approach as Google's.

The problem is particularly important for landuse=retail: the retail color is generally not visible, because retail areas in Europe tend to be very high densitiy areas. Example: https://www.openstreetmap.org/#map=16/52.0774/4.3125

[...]

PR #2934, opened 2017-11-07
Make parking gray instead of yellow

### LINKED ISSUE #2904: Parking too prominent
Parking is currently rendered too prominent. Compare for example the following renderings.

Before:
screen shot 2017-10-24 at 00 45 24

After (possible):
screen shot 2017-10-24 at 00 45 28

In the 'before' rendering, the parking is rendered more prominently than the buildings (including the supermarket). Can we tone down the rendering? Light gray like in the possible 'after' might be going too far, but do we have any other options?

math1985:
Resolves #2904

### DIFF
--- a/landcover.mss
+++ b/landcover.mss
@@ -29,7 +29,7 @@
 @aerodrome: #e9e7e2;
 @apron: #e9d1ff;
 @garages: #dfddce;
-@parking: #f7efb7;
+@parking: #eeeeee;
 @railway: @industrial;
 @railway-line: @industrial-line;
 @rest_area: #efc8c8; // also services


kocio-pl:
Rendering as proposed in #2904:

Before:
screen shot 2017-10-24 at 00 45 24

After:
screen shot 2017-10-24 at 00 45 28

More rendering examples: https://github.com/gravitystorm/openstreetmap-carto/issues/2904#issuecomment-341870242

JamesKingdom:
I think this makes the map look dull and more miserable, I like being able to find a car park easily without it blending into other things. What else do we render as yellow then?

kocio-pl:
Parking yellow is more intensive than most of the roads - especially yellow, orange and white, which is most of them - and much more intensive than social amenities (we also render sand and beach, but that is rather not too related). 

What do you mean by "dull" and "miserable" exactly (other than "I don't like it")? Why do you want to see every parking space, no matter how small, more than hospital and school, which are usually bigger? For me it's just a car-centric POV and there should be a map style for drivers probably (I guess German style is pretty good at it), but this is general map, so it should be more balanced.

rrzefox:
I don't like this. The completely colorless gray virtually makes it disappear completely.
How about something like half way in between current and gray? I don't think it looks particulary great, but it's less intensive than the yellow and less unrecognizable than gray.
bla001

kocio-pl:
> How about something like half way in between current and gray?

Unfortunately this is very close to societal amenities area, so it wouldn't work. 

> I don't like this. The completely colorless gray virtually makes it disappear completely.

I don't agree with that. _Before_ image shows 4 big parkings and about 6 small ones. On the _after_ you still see these 4 big ones (I guess you see them too and they didn't disappear), but small ones are not disturbing any more. If you want to see them, you need to zoom in.

That's why I asked about small parkings - why should you always see all of them, even if they don't belong to particular zoom level and make noise? We moved shop icons later exactly because of that and it makes sense for me to mind the scale. Gray+blue P letter works this way, yellow doesn't.

math1985:
@rrzefox Thanks a lot!

Personally I think even in the intermediary rendering, parkings are still too strong. Plus what @kocio-pl says.

Klaus-Tockloth:
This change is not very convincing ... I prefer to keep yellow.

kocio-pl:
Could you be more specific - what is convincing in current state of things and what exactly is the weak point in this proposition? It would be also good to hear any response to my questions about small parkings.

Klaus-Tockloth:
> Could you be more specific

The change is too big ... start with something smaller:
- draw private parking area in a pale color (e.g. similar to garages)
- draw small parking spaces in a pale yellow or gray
- keep the yellow color of all other parking areas
- wait for feedback of the map users

kocio-pl:
Thanks for the feedback, but I think that this solution is not going to work:

> draw private parking area in a pale color (e.g. similar to garages)

Private parkings are just about 1% of uses:

https://taginfo.openstreetmap.org/tags/amenity=parking#combinations

and are already easy to recognize (faded P color), so this change would just make color inconsistency (new problem) to solve a marginal problem. And as you can see on the first example, the size is more important feature, since big private parking is still more important to see on a given level than a small public parking.

Current yellow is the problem because it does not match the relative importance of the object it covers and because there's so many of them.

>    draw small parking spaces in a pale yellow or gray

1. Pale yellow is close to the societal amenities, as it was already shown, so it would be exchanging one problem with another. I was testing both "transportation gray" and garages, but both are hard to see and/or similar to other areas. Light gray looks like the only color that works - it's easy to spot when the area is big (especially with the letter), but is not visible when size is small (and without the letter).

2. We have no tagging for small parkings. I was trying to introduce it, but found no clear rules how to define them. We could filter the size on a given level, but that would still be color inconsistency (whatever color should it be), since small parking will become big when you zoom in.

Klaus-Tockloth:
> Private parkings are just about 1% of uses

Hmm, within cities with a high dense of parking area it's probably much more.

All areas amenity=parking (no nodes):
bildschirmfoto 2017-11-14 um 06 53 39

Only area amenity=parking + access=private (no nodes):
bildschirmfoto 2017-11-14 um 06 55 16

dieterdreist:
2017-11-13 14:34 GMT+01:00 kocio-pl <notifications@github.com>:

> Thanks for the feedback, but I think that this solution is not going to
> work:
>
> draw private parking area in a pale color (e.g. similar to garages)
>
> Private parkings are just about 1% of uses:
>
> https://taginfo.openstreetmap.org/tags/amenity=parking#combinations
>

according to taginfo it's about 8% (same link, please have a second look):
https://taginfo.openstreetmap.org/tags/amenity=parking#combinations

kocio-pl:
> according to taginfo it's about 8% (same link, please have a second look):

Sorry, you're right with it.

The biggest problem for me is the color. When looking at the example part of the map I see:
- red and yellow roads + **parkings location and area (big and small)**
- the rest, including buildings, commercial areas and trees

I think that roads are important and it's hard to find different colors for them, so it doesn't bother me if they are visible the most, even if big roads are car-centric. They are also rather long than wide, so not as visible as if they would be free form areas. The parkings however are not that important for general use - they should be visible not more than buildings and other areas. 

It's quite easy to make them less important on the map, but still easy to recognize if they're big enough, but it's not possible without color change IMO. After the change I see:

- red and yellow roads + **big parkings locations** (P letters)
- the rest, including buildings, commercial areas and trees + **big parkings area**

They are not less visible than buildings etc., but they probably would be if they are bit smaller - and this is good, because also buildings are more visible when they're bigger. Parkings still show their location loud, but at least they don't obscure the view at the same time.

Small parkings noise is another problem, which can be fixed using the same color change - they might be hidden also if needed for different reasons (see #2896), but should not use different color than the big ones, because they're not different kind of objects.

math1985:
I'm going ahead and merge this PR.

I know we all got used to the yellow color, but I really don't believe rendering parkings so prominent is a good idea. Even a mixture between the current yellow and the proposed gray still attracts too much attention. To me parkings are still visible enough in gray.

Thanks for all feedback!

PR #2936, opened 2017-11-08
Render entrances.

tpikonen:
Render entrances starting at z=17. The "main" and "staircase" entrances are rendered with a square marker, the "service" entrance with a circle (starting at z=18) and other entrances with a disc.

Three levels of marker brightness reflect the access permissions of the entrance. Darker color means more permissive access. The "main" entrances are dark by default. The access tag modifies the brightness such that access="yes" and "permissive" set the marker to darkest and access=no to the lightest colour. Entrances with other access tag values have intermediate brightness.

Exits, such as entrance=exit and entrance=emergency are not rendered.

Address label text-halo-radius is increased at z>=18 to improve readability of labels on top of entrance markers.

**Examples:**

[Buildings](http://www.openstreetmap.org/#map=19/60.17520/24.79240) with "main" and "service" entrances, z=17-18:
main_service_z17 main_service_z18

[School](https://www.openstreetmap.org/#map=18/60.25117/24.85294) with several entrance="yes" and access="permissive"|"private", z=17-18:
school_z17 school_z18

[Apartment building](http://www.openstreetmap.org/#map=18/60.24282/24.88029) with entrance="staircase" tagged with addr:unit, z=17-18:
unit_z17 
unit_z18

### DIFF
--- a/addressing.mss
+++ b/addressing.mss
@@ -34,8 +34,11 @@
     text-wrap-width: 30; // 3.0 em
     text-line-spacing: -1.5; // -0.15 em
     text-margin: 3; // 0.3 em
-    [zoom >= 18]["addr_unit" != null]["addr_housenumber" = null] {
-      text-name: [addr_unit];
+    [zoom >= 18] {
+      text-halo-radius: @standard-halo-radius * 1.25;
+      ["addr_unit" != null]["addr_housenumber" = null] {
+        text-name: [addr_unit];
+      }
     }
     [zoom >= 20] {
         text-size: 11;
--- a/buildings.mss
+++ b/buildings.mss
@@ -5,6 +5,8 @@
 @building-major-fill: darken(@building-fill, 20%);
 @building-major-line: darken(@building-major-fill, 25%);
 
+@entrance-permissive: darken(@building-line, 15%);
+@entrance-normal: @building-line;
 
 #buildings {
   [zoom >= 13] {
@@ -40,5 +42,47 @@
 #bridge {
   [zoom >= 12] {
     polygon-fill: #B8B8B8;
+  }
+}
+
+#entrances {
+  [zoom >= 18]["entrance" != null]  {
+    marker-fill: @entrance-normal;
+    marker-allow-overlap: true;
+    marker-ignore-placement: true;
+    marker-file: url('symbols/rect.svg');
+    marker-width: 5.0;
+    marker-height: 5.0;
+    marker-opacity: 0.0;
+    ["entrance" = "main"] {
+      marker-opacity: 1.0;
+      marker-file: url('symbols/square.svg');
+    }
+  }
+  [zoom >= 19] {
+    ["entrance" = "yes"],
+    ["entrance" = "main"],
+    ["entrance" = "home"],
+    ["entrance" = "service"],
+    ["entrance" = "staircase"] {
+      marker-opacity: 1.0;
+      marker-width: 6.0;
+      marker-height: 6.0;
+      ["entrance" = "service"] {
+        marker-file: url('symbols/corners.svg');
+      }
+    }
+    ["access" = "yes"],
+    ["access" = "permissive"] {
+      marker-fill: @entrance-permissive;
+    }
+    ["access" = "no"] {
+      marker-fill: @entrance-normal;
+      marker-file: url('symbols/rectdiag.svg');
+    }
+  }
+  [zoom >= 20]["entrance" != null] {
+    marker-width: 8.0;
+    marker-height: 8.0;
   }
 }
--- a/project.mml
+++ b/project.mml
@@ -1070,6 +1070,22 @@ Layer:
         ) AS guideways
     properties:
       minzoom: 11
+  - id: entrances
+    geometry: point
+    <<: *extents
+    Datasource:
+      <<: *osm2pgsql
+      table: |-
+        (SELECT
+            way,
+            tags->'entrance' AS entrance,
+            access
+          FROM planet_osm_point
+          WHERE tags->'indoor' = 'no'
+            OR tags->'indoor' IS NULL)
+        AS entrances
+    properties:
+      minzoom: 18
   - id: aeroways
     geometry: linestring
     <<: *extents
--- a/symbols/corners.svg
+++ b/symbols/corners.svg
@@ -1,0 +1,30 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<svg
+   xmlns:dc="http://purl.org/dc/elements/1.1/"
+   xmlns:cc="http://creativecommons.org/ns#"
+   xmlns:rdf="http://www.w3.org/1999/02/22-rdf-syntax-ns#"
+   xmlns:svg="http://www.w3.org/2000/svg"
+   xmlns="http://www.w3.org/2000/svg"
+   id="svg4"
+   height="1"
+   width="1"
+   version="1.1">
+  <metadata
+     id="metadata10">
+    <rdf:RDF>
+      <cc:Work
+         rdf:about="">
+        <dc:format>image/svg+xml</dc:format>
+        <dc:type
+           rdf:resource="http://purl.org/dc/dcmitype/StillImage" />
+        <dc:title></dc:title>
+      </cc:Work>
+    </rdf:RDF>
+  </metadata>
+  <defs
+     id="defs8" />
+  <path
+     id="path2"
+     d="m 0.00390625,0.00390625 c 0,0.13203125 0,0.2640625 0,0.39609375 H 0.20507812 V 0.20507812 H 0.4 V 0.00390625 Z m 0.59609375,0 V 0.20507812 H 0.8046875 V 0.4 H 1 V 0.6 H 0.8046875 V 0.8046875 H 0.6 V 0.99609375 H 0.4 V 0.8046875 H 0.20507812 V 0.6 H 0.00390625 V 1.0039062 H 1.0039062 V 0.00390625 Z"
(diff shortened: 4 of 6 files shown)


kocio-pl:
I think that we might use brown dot for some amenities, so I would try some other shape/color for entrances.

tpikonen:
The color is the same or slightly darker or lighter than the building border in order to make the markers as unobtrusive as possible, especially on z=17.

On the other hand one could go to the other direction in color coding and make non-permissive entrances the same color as the border and permissive ones progressively lighter, giving an impression of an opening in the building border.

kocio-pl:
What if we use square border only (without filling)?

tpikonen:
> What if we use square border only (without filling)?

Ok, I tried this (branch updated) and it doesn't look that bad. Entrances are now rendered with a hollow square, except main entrances still with a solid one. Service entrances have just a hint of a square with a marker with only corners of a square, and access=no entrances have a diagonal line across the hollow square. Entrances with access=yes|permissive are rendered with a larger contrast (darker marker). Here's some images:

Apartments with main and service entrances, z=17-18:
main_service2_z17 main_service2_z18

School, z=17-18:
school2_z17 school2_z18

University buildings with entrance=main's, one with access=permissive, one without and several service entrances, z=17-18:
uni_z17 uni_z18

Two entrances, one with access=no, z=18-19:
accessno_z18 accessno_z19

kocio-pl:
I like this scheme. However I would render all the entrances at z19 and on z18 maybe main entrances only.

aceman444:
Yes, highlighting which entrance is main is great.

boothym:
What does it look like for nodes which are tagged as both house numbers and entrances?

tpikonen:
> I like this scheme. However I would render all the entrances at z19 and on z18 maybe main entrances only.

Did you perhaps mean z18 and z17? I think there is enough space to render all entrances at z18 without the map looking too busy. Also, addr:housenumber (+ addr:unit) is rendered at z>=17 also for entrances.

tpikonen:
> What does it look like for nodes which are tagged as both house numbers and entrances?

The marker is not visible below the label. Have a look at the last image of the first comment.

matkoniecz:
How it looks at entrances tagged on places other than buildings? Especially on nodes where entrance and barrier=gate tag are tagged together?

matkoniecz:
It seems that code is using catch-all rendering, this is discouraged. For example entrance=yes;staircase should not be rendered.

tpikonen:
> How it looks at entrances tagged on places other than buildings? Especially on nodes where entrance and barrier=gate tag are tagged together?

The layers in buildings.mss are rendered early, so an entrance marker is usually not visible if there is another marker on top of it, except in really high zooms. Here's images with barrier=gate at z17-19:

barrier_entrance_z17 barrier_entrance_z18 barrier_entrance_z19

> It seems that code is using catch-all rendering, this is discouraged. For example entrance=yes;staircase should not be rendered.

I updated the PR to render just the entrance types which are in the table in the wiki.

donaldrnoble:
I had been hoping entrances would be rendered for a while, but never done anything abou It.

I like the unfilled icons, but is it  possible to use a rectangle for the icon? I think this would look more like a doorway, if it was something like twice as high as it was wide. This might also distinguish entrances from other generic circular/square icons.

NetWormKido:
Hi!
I propose to additionally render addr:flats with ref of entrance like on screenshot
http://wiki.openstreetmap.org/wiki/Key:addr:flats
https://taginfo.openstreetmap.org/keys/?key=addr%3Aflats
https://www.openstreetmap.org/node/2713780339/
https://www.openstreetmap.org/node/2713780340/
screenshot_20171119_090526
screenshot_20171119_090551

dieterdreist:
do we really need 3 tags?
addr:housenumber 
addr:unit
addr:flats

aren’t the flats either housenumbers or unit numbers?

tohaklim:
I think @NetWormKido meant to render both the entrance ref (part of the address, but not included/covered by any address tags) and addr:flats, so as to show where each flat is at a glance

tpikonen:
@NetWormKido, take a look at PR  #2932 which is about rendering of address labels. This PR is about rendering markers to entrance=* nodes.

Personally, I'm not against rendering addr:flats, but it should probably be only at the highest zoom levels (z>=19 or so). Also, I'd first like to finish PR #2932, which is mostly about addr:unit rendering before adding flat numbers.

tohaklim:
I erroneously assumed (and was proven wrong) that addr:unit  is the European/US way of tagging entrance=staircase's refs, but every time I see examples of usage I get confused again. 
Both entrance=staircase refs and addr:unit seem to be used for apartment blocks. Both are used for entrances. Weird

NetWormKido:
@tpikonen, I completely agree that the addr:flats need to render at the highest zoom levels (z>=19). I hope you consider my proposal.

nebulon42:
Just a quick remark: Outline only rendering is quite noisy IMO. Otherwise good initiative.

kocio-pl:
> Did you perhaps mean z18 and z17? I think there is enough space to render all entrances at z18 without the map looking too busy. Also, addr:housenumber (+ addr:unit) is rendered at z>=17 also for entrances.

Sorry, it took me some time to think it over. This is nice feature to have, but I really mean z19+. It's not about the available space, I want to avoid the clutter.

Address numbers are used to search places on the street, so they can be shown earlier. Entrances are related to the single buildings and it's easy to guess they are here - most of the time footpath or multiple numbers indicate their existence. Sometimes even the shapes of the lawns are enough to know where is the entrance.

The only new thing that this code brings and which can't be guessed is the type of the entrance, but this is really local feature.

tpikonen:
> This is nice feature to have, but I really mean z19+.

Ok, I'll update the code to do this eventually, but I would also like to remove the outline-only rendering of the markers (i.e. make the markers opaque), as @nebulon42 suggested. I don't know how to do it with the marker-* directives in carto, so this needs some thinking.

tpikonen:
Main entrances are now rendered at z>=18 and the rest at z>=19, as requested.

Opaque markers with dark borders are apparently not possible (i.e. they would require non-monochrome markers), so the generic entrance markers are still outline only.

As far as I'm concerned, this PR is ready.

kocio-pl:
Thanks, I hope I will check it soon.

kocio-pl:
I like the outcome and that's what I would merge soon - the only request left is to squash all the commits into one.

Although we don't use z20 on OSMF servers currently, it makes sense in this case to extend the definition there, because the differences are sometimes still subtle on z19 and I don't want to make it more evident earlier, because such small feature should not be too visible. Showing only main entrances on z18 sounds OK for me - that might help to find the way if the building is big enough.

`entrance=main` and `entrance=yes` ( https://www.openstreetmap.org/#map=19/52.23146/21.00713 )
z18
kzrzjjhw
z19
1ndsi2cj
z20
mb2lodsb

`entrance=yes + access=yes` ( https://www.openstreetmap.org/node/5329685559 )
z19
dt5g4ofx
z20
aywiglel

`entrance=yes + access=no` ( https://www.openstreetmap.org/node/4709081766 )
z19
czpobi_u
z20
kshihnvh

mostly `entrance=service` ( https://www.openstreetmap.org/node/3516047187 ) and one `entrance=yes` (the difference is visible on z2)
z19
1enkqkmn
z20
juivryra

polarbearing:
There are lots of cases where people put the address label on an entrance node. This is somehow discussed in the first post, however is there a before/after comparison? Maybe the number should be under the icon, not on top?

tpikonen:
> I like the outcome and that's what I would merge soon - the only request left is to squash all the commits into one.

Done. Thanks for the positive review.

kocio-pl:
Thanks for all the work on this code, @tpikonen!

polarbearing:
While I appreciate rendering entrances in general, I find rendering the value "entrance" = "staircase" problematic. It was not part of the original key proposal, and is badly documented. It is a value where the physical appearance of the entrance conflicts with its function, such as "main" or "emergency".

kocio-pl:
We can of course change that (or any other) part. Is this the only problem do you see here?

polarbearing:
I don't see examples what happens when addr:housenumber is on the entrance node,
http://www.openstreetmap.org/node/2825036652 18000 times in Berlin,
or even a shop icon in addition, 
http://www.openstreetmap.org/node/2446830336 99 times in Berlin

tpikonen:
> I don't see examples what happens when addr:housenumber is on the entrance node,
http://www.openstreetmap.org/node/2825036652 18000 times in Berlin,
or even a shop icon in addition,
http://www.openstreetmap.org/node/2446830336 99 times in Berlin

In both cases address label or shop icon will be placed on top of the entrance marker, since the entrances layer is lower than addresses and amenity points and allows overlaps.

In practice the entrance marker will not be visible, since its colors are chosen deliberately to be the same as in building borders.

kocio-pl:
I believe staircase is OK, because in OSM many types are mixed and this one is not cryptic. We can change it in the future if needed, for the moment I'm inclined to stay with it.

polarbearing:
Thanks, could we see a preview for the housenumber/shop on an entrance?

kocio-pl:
I don't get exactly what do you want to see, could you post a link to the example place?

polarbearing:
Posted some 5 posts above.

kocio-pl:
Ok, now I see:
gu1o4snv
6e8v9ufx

polarbearing:
Hm, the entrance icons are not fully covered. They remain visible, in particular under the '7' being so narrow, and in the jewellery ring. The 71 looks a bit like 7.1. They would probably also be bigger than the generic shop dot. 

Would it be possible to suppress them in the code when there is another icon or label on top?

kocio-pl:
Please open new ticket for tuning this feature.

Issue #1957, opened 2015-11-05
Urban/rural or density-based rendering differences


pnorman:
Cross-ref #1705 #1956, and others.

This is distinct from separate styles. An example is @systemed having different default surface assumptions for footpaths in rural areas than in urban areas. He is using country-specific urban area data that we don't have for this style.

I don't think it's an option for a few reasons

* We don't have a clear datasource for urban areas. Place polygon coverage is limited, and anything else is indirect and would require processing
* It would stretch our resources thinner. I don't think we'd be able to adequately test all combinations of rural/urban on new PRs
* It'd add significant layer complexity. We'd need to either
** Add another preprocessed file and load it into the database, increasing install difficulty and deployment difficulty;
** Compute areas in DB at query time, involving reasonably complex SQL; or
** Pre-compute areas from the DB, requiring triggers, background jobs, or other tools which would increase difficulty
* It'd increase MSS complexity. The scale would depend on how many features would have different renderings in each, but [as I tweeted](https://twitter.com/penorman/status/660668304910974977) our MSS total is larger than HOT OSM, OSM Bright, Stamen Toner, and CartoDB basemaps **combined**. We need to be reducing this, not increasing it.

systemed:
It doesn't significantly increase MSS complexity. My experience with cycle.travel is that it actually makes maintenance _easier_, because on the few styles that matter, you just need two sensible, easy-to-use defaults for rural and urban areas, rather than having to spend hours experimenting with placements and spacing to find the sweet spot that works (= isn't completely intolerable) for both. Most of the time it's just a matter of changing the zoom level in the selector - "show this at z14 in rural areas, z16 in cities". So a comment like [this by @matkoniecz](https://github.com/gravitystorm/openstreetmap-carto/issues/1956#issuecomment-154001186) simply isn't borne out by my experience.

However, for the other reasons you cite, I'd agree that it isn't suitable for openstreetmap-carto. I'd also expect that mappers would come to rely on whatever set of urban boundaries or density criterion were chosen for osm-carto, making assumptions that wouldn't translate well to third-party styles.

Finally, it's pretty difficult to get decent built-up area boundaries for most countries. (The UK, USA, and Canada are honourable exceptions.)

imagico:
Regarding preprocessing possibilities - there is some info and sample data on http://www.imagico.de/map/osm_builtup_en.php

[...]

PR #2945, opened 2017-11-11
Dispay restaurants, pubs etc. on z17 as small brown squares

math1985:
This prevents overcrowding on z17.

Before:
screen shot 2017-11-12 at 00 54 26

After:
screen shot 2017-11-12 at 00 53 00

### DIFF
--- a/amenity-points.mss
+++ b/amenity-points.mss
@@ -35,6 +35,25 @@
 
 /* Note that .points is also used in water-features.mss */
 .points {
+  [feature = 'amenity_pub'],
+  [feature = 'amenity_restaurant'],
+  [feature = 'amenity_food_court'],
+  [feature = 'amenity_cafe'],
+  [feature = 'amenity_fast_food'],
+  [feature = 'amenity_biergarten'],
+  [feature = 'amenity_bar'],
+  [feature = 'amenity_ice_cream'],
+  [feature = 'amenity_nightclub'] {
+    [zoom >= 17][zoom < 18] {
+      marker-file: url('symbols/square.svg');
+      marker-placement: interior;
+      marker-clip: false;
+      marker-fill: @amenity-brown;
+      marker-width: 4;
+      marker-line-width: 0;
+    }
+  }
+
   [feature = 'tourism_alpine_hut'][zoom >= 13],
   [feature = 'tourism_wilderness_hut'][zoom >= 13],
   [feature = 'amenity_shelter'][zoom >= 16] {
@@ -64,7 +83,7 @@
     marker-clip: false;
   }
 
-  [feature = 'amenity_bar'][zoom >= 17] {
+  [feature = 'amenity_bar'][zoom >= 18] {
     marker-file: url('symbols/bar.svg');
     marker-fill: @gastronomy;
     marker-placement: interior;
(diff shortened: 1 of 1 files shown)


math1985:
Not sure myself if this is an improvement, what do others think?

matkoniecz:
Based on this single picture - I really like this solution for solving problem that I failed to notice.

kocio-pl:
I'm not sure, but rather don't think it's useful. For shops it made sense:
- there's so many of them in many places
- violet dot always mean just a "general shop" (more important ones are rendered earlier)

Eating places:
- are not as dense 
- there's less such places
- brown dot may mean many more things (I plan to add it for different amenities with no icon)

SK53:
I simply cannot adequately resolve the two colours of dots in the new image, so whereas before I can identify areas which are broadly shopping streets and places with a good choice of restaurants (and the odd cafe / bar) now I can't. It makes the map far less informative and I suspect the perceived improvement is most obvious to those working with Carto most closely.

It is important to show how such rendering rule changes affect the appearance of other less dense areas of the map: at a minimum small towns, villages and hamlets as well as isolated restaurants/bars/cafes/pubs.

math1985:
> I simply cannot adequately resolve the two colours of dots in the new image

Just to be clear: it is for you difficult (or impossible) to tell apart the brown and pink dots/squares?

systemed:
This is another urban/rural issue (see #1957). Whether or not it makes sense in urban areas, it's a clear loss of utility in rural areas: http://www.openstreetmap.org/#map=17/51.81633/-1.70480

FWIW, cycle.travel renders pubs/cafe icons at z13 in rural areas, z16 in towns, z17 in cities (names come in later).

imagico:
> FWIW, cycle.travel renders pubs/cafe icons at z13 in rural areas, z16 in towns, z17 in cities (names come in later).

@systemed - what data do you base this selection on?

systemed:
Urban area polygons from OS Open Data in the UK, CORINE in mainland Europe, TIGER in the US, and equivalents (which I can't remember off the top of my head) for Canada and Mexico. They're all of pretty good quality apart from CORINE. For openstreetmap-carto it would be more appropriate to generate any such polygons from OSM data, of course.

SK53:
@math1985 I can just about discriminate them if I take my specs off and squint really closely at the screen; but this indoors with reasonable lighting etc. I would imagine the difference would be much harder on a mobile device in bright sunlight.

@systemed et al. I currently have draft urban areas on github for the following areas : Australasia & Oceana, Africa, South America, North America. I can generate them for Central America and Europe, but I had issues with my algorithm in Japan which complicated generating an Asia dataset. Everything is derived directly from OSM and is stored as Geojson files to allow visualisation of the data (and, perhaps more importantly, diffs between versions). Data are partitioned into either 5 or 10 degree squares to ensure no file is too large for github. The repository is : https://github.com/SK53/ua2/.

kocio-pl:
Please move the discussions about urban areas to #1957, since it's an important problem on its own.

dieterdreist:
I am not in favor of this change, as it makes the map less useful, clearly
in less dense areas (outskirts, rural, etc.) but also in very dense central
areas. I agree that this place before was kind of cluttered and is less so
now, but while the icons gave me some idea before what to expect before,
the dots don't any more. I have no problem in distinguishing those 2 colors
if I concentrate on them, but the overall picture is less clear now, mostly
a sea of dots (where no color or structure eveloves).
It is important to know whether a place is a restaurant or a pub/café, and
by reducing this to a dot you loose this distinction until you zoom to
quite high levels.

math1985:
I think the criticism makes sense, I'm going to close this PR.

Klaus-Tockloth:
Use also dots instead of squares:

bildschirmfoto 2017-11-18 um 07 14 55

Issue #2931, opened 2017-11-06
Add terrain relief data


kocio-pl:
I think about adding terrain relief data and here's a basic shopping list:

1. Kind of terrain sculpture is missing not only in this style, but also in OSM in general. We have only peak, saddle and embankment tags and it doesn't look like this project is capable of having much more details about it. So it means using external data.

2. I believe contour lines are too detailed and don't belong to general style (they are used for example in the cycling layer on OSM website) and using colors would be also too intrusive and not blending well with our color usage (OpenTopoMap uses relief colors [up to z8](https://opentopomap.org/#map=8/51.3769/-0.1071), and then changes the meaning of colors on z9+, which is confusing to me). It means we should probably use hillshading, which is not too disturbing, yet shows the basic shape of a terrain - you can taste it also on cycling layer.

3. We use some data from Natural Earth project and there's also a [1:10m Shaded Relief](http://www.naturalearthdata.com/downloads/10m-raster-data/10m-shaded-relief/) and [1:50m Manual Shaded Relief](http://www.naturalearthdata.com/downloads/50m-raster-data/50m-manual-shaded-relief/). It looks like precise enough to me.

The problems are (1) if we decide to add such data to our style - that means discussing how useful or how disruptive is that - and in case we decide to use it, (2) how to do it exactly.

I would like to test hillshading myself, but I don't know where to start, so any hints and help with that would be welcome.

math1985:
Im against. The map is alreaady busy enough, and this project is about showcasing osm data.

matkoniecz:
I am also against - it is supposed to show primarily OSM data and it would cause problems with cases where OSM data mismatch relief data (for start: sloped lakes, natural=peak/saddle mismatching relief data).

I believe that everything wrong visible in this style should be fixable by editing OSM data, without tagging for renderer.

Iff there was some open, high-quality, preprocessed to remove mismatches from OSM data relief data I would consider this, but AFAIK there is nothing like this.

math1985:
Closing the issue.

kocio-pl:
Thanks for you comments.

> Im against. The map is alreaady busy enough, and this project is about showcasing osm data.

The argument about being busy is not true for low level and I don't expect it ever will. We don't have to show hillshade for all zoom levels, it's the most important for low zoom. 

[...]

Issue #2742, opened 2017-08-13
Identify big areas


kocio-pl:
It would be good to have a list of biggest areas already tagged (in the world and per continents) and some visualizations for each type of natural/landuse areas to show where they are located. Another thing would be to find out what kind of big objects are currently not even tagged - maybe natural/landuse system is missing something important.

Rationale: 

When looking how to improve [medium](https://github.com/gravitystorm/openstreetmap-carto/issues/2194) and [low zoom](https://github.com/gravitystorm/openstreetmap-carto/issues/2688) levels, I was trying to identify big areas that could be shown there. With some clues from other generalized maps I was able to notice that farmlands and sand can be shown earlier, because big enough areas are tagged. However checking it manually is tedious and Overpass server is not able to show for example `natural=*` for the whole world, nor even single continent.

math1985:
Sorry, I'm not sure what you mean with this issue?

kocio-pl:
I mean a database tool/service (or just output) with such data. Also a list of potentially interesting big objects which are not in database (for example "Australian deserts").

kocio-pl:
I guess this general list is enough for now:

https://wiki.openstreetmap.org/wiki/User:Kocio/low_zoom_rework

PR #2946, opened 2017-11-12
Adding natural and semi natural areas to use cases

kocio-pl:
Since it's pretty general and big subject, I prefer to make a PR instead of direct updating file.


matkoniecz:
> semi-natural lines (big canals)

Are there any canals that should be displayed at z5, z6? I think that either there would be invisible or would be confused with minor seas at this scale.

matkoniecz:
> relief

see  #2931 for my doubts

matkoniecz:
http://maps.imagico.de/#map=3/49.457/-3.535&lang=en&r=osmlz&o=1aa&ui=2 is proof that low zoom landcover may make sense, but I am not sure is it desirable to go in this direction as it would rely heavily on using nonOSM data in imaginable future

selection_019

kocio-pl:
> Are there any canals that should be displayed at z5, z6?

I don't know, it's a general statement. It's a problem with low zoom - you have to find examples (see #2742).

For relief - I believe for low and medium zooms it's less of a problem than on high zooms: it's still external data, but the errors in placement should not be more than dozen of meters (and if it is, I would rather suspect our data, as this image is more coherent than a big collection of independent edits).

matkoniecz:
> For relief - I believe for low and medium zooms it's less of a problem than on high zooms

Relief disappearing on zooming in seems for me to be a very confusing thing.

kocio-pl:
My take is that we should show only our data for landcover:
1. I try to not use external data if possible (relief is the only case when I feel exception is worth).
2. In our style map doesn't have to be beautiful nor complete - I prefer to show overview of OSM data.

matkoniecz:
Unfortunately without fallback data is going to be quite noisy - see http://www.openstreetmap.org/#map=4/51.26/11.69&layers=H

kocio-pl:
> Unfortunately without fallback data is going to be quite noisy

I don't think so - but it should be tested, of course, and it takes a lot of time to render big areas. That's why I didn't mentioned it yet, though I have some general ideas.

I think that [noise](https://en.wikipedia.org/wiki/Noise#Measurement) is a product of too many details (high frequency elements) and too big contrast (high amplitude). Contrast is now much less thanks to the midzoom changes on osm-carto and on low zoom there are only few types, which can be even more generalized (as proposed by @imagico - see https://github.com/gravitystorm/openstreetmap-carto/issues/2688#issuecomment-336601112).

Example of deploying this idea is French style - let's look at z7 in SF area:
http://tile.openstreetmap.fr/?zoom=7&lat=36.40542&lon=-118.2174&layers=B000000FFFFFF

Contrast is a bit too big, but it's better than HOT, which is designed to be high contrast for printing purposes, AFAIK (we have less intensive colors already). There are only woods, grass, farmlands and sands visible, I guess (+natural parks). There can be also ice in different places and maybe few more types, but that's it for a low zoom. We don't even have to make generalizations most of the time, so I want to make them only when really needed (I feel that Christoph tried to use them to make map beautiful).

The rest is just a lot of testing for the whole planet...

kocio-pl:
As of your concerns and changes request:
1. Let's continue to talk about relief in details in its own ticket.
2. Use cases are about goal-oriented design, so they try to describe ideal situation, to avoid narrow thinking in the context of what's available and possible. It doesn't mean that every use case can be solved and has to be incorporated into a product. I think that discarding use case should be made only if there's something wrong with the goals (like "making fun of people") or the context does not match it (like "Looking up country location" - "single trees").

kocio-pl:
This is more like what we could show on z7: 

http://maps.imagico.de/#map=7/38.355/-121.055&lang=en&r=osmlz&o=55&ui=2

but remember that "OSM-Carto style landcover" are in fact not the colors we use on midzoom osm-carto - they are not faded, so the contrast it also too big.

math1985:
> Understanding what the area looks like

Would it be possible to be more specific? I suppose you're aiming for something like in @matkoniecz's image. This map seems largely a biome/climate map. It doesn't say anything about human landuse, for example.

About the minor water areas on z5/6/7: I slowly start changing my mind about them a bit. Some examples:

screen shot 2017-11-12 at 00 53 00
screen shot 2017-11-12 at 00 54 26

We previously asked ourselves if we can render these minor water areas here, and the answer to that is sure we can. But should we? I actually doubt that these lakes is the information the user is looking for, and it even distracts a bit from the main message.

kocio-pl:
> Would it be possible to be more specific? I suppose you're aiming for something like in @matkoniecz's image. This map seems largely a biome/climate map. It doesn't say anything about human landuse, for example.

The lower zoom level, the more areas are just natural, I think there are not too many human landuses really visible on low zoom. Which ones could it be, other than semi natural (like farmlands), military areas and natural reserves?

Countries (nominally areas, but visible as border lines) and biggest cities (nodes) add some human related informations on top of that, so it's balanced enough for me.

> About the minor water areas on z5/6/7: I slowly start changing my mind about them a bit.

I have no strong opinion on that. I just think that it's zoom level independent problem in general - small dots can always be a problem.

What filter would you like to use here?

kocio-pl:
Any comments on this?

matthijsmelissen:
I'm still trying to understand the usecase better. Can you try to give a very concrete example of something that a user could read from the map on the relevant zoom levels? For example in @imagico's example one could read something like 'the north of Africa is a dry area' or something like that.

matthijsmelissen:
On the other hand, in Europe I can see a lot of different colors of green, but there is not really information I can read out of that. That's something I'd like to avoid.

kocio-pl:
For me it's so basic that it's hard for me to give examples. "This part of the continent is full of forests" or "it is basically desert land" or "there are big mountains" or "there is the ice area" or "this is a big steppe belt" are important geographical informations for me.

Forest area is very different than steppe, so even shades of green can say something valuable.

Map made by @imagico on low scale tries to cover everything with non-OSM data just to show how it would look like in the future, but that's basically the same idea.

matthijsmelissen:
What information would you get out of @imagico's map for Europe?

kocio-pl:
That it's mostly covered with forests. There are also for example sand patches in North Africa which shows that it's different:

http://maps.imagico.de/#map=4/49.781/16.348&lang=en&r=osmlz&o=55&ui=2

kocio-pl:
I believe now the problem is you are using narrow definition of use case. In reality you're talking about limited practical (travel) use, for example "where could I drive". But when you look at global scale, there's nothing too practical in this sense, since this is not a human scale. Yet there are informations which are available here.

For example "Looking up country location" - so the country name would be enough, knowing borders is not needed probably, but it's good to have them too. If we were strictly about journey planning, we also don't need railway lines - only the stations. But it's good to see them too. 

So let's not think about just journeys. Many of our viewers may not go to any journey if they are from less developed sites, yet why not show them the world as it is? That's also diversity we should care about. Knowing where the ice is or where are the forest, sands, rivers and lakes is also a knowledge about Earth geography, even if you're not traveling half of the world in general, so you don't need them to plan a journey. The names of big geographical entities would be great also, but there are language related problems which are outside the scope of use cases.

matthijsmelissen:
That’s definitely how I intend it too!

kocio-pl:
So what do you think about this PR? 

I believe @matkoniecz is against both rendering low zoom (semi-)natural features and especially relief. You wanted me to be more specific and I'm not able to be more specific even if I try. I still think this PR is good enough and we don't need to implement all the use cases (this is just a designing phase, implementation is another story). I see no clear decision at this moment.

matthijsmelissen:
@kocio-pl I guess we can merge this.

However, could you add a section 'Use cases that we currently don't render' (the reverse of ' Some features that we currently render for which we do not (yet) have a use case')?

matthijsmelissen:
On the other hand, we could also treat the list as a way to document the use cases for the features that we do render.

kocio-pl:
I'm still not sure where can we go with usecases, so I prefer to leave it as it is now. Maybe it will be more clear in the future.

PR #2948, opened 2017-11-19
Update Noto to Phase III

### LINKED ISSUE #2908: Update code for UI fonts vs normal fonts
For some scripts, Noto Sans offers both, a _UI_ and a _normal_ font. For the other scripts, it offers only the _normal_ font.

Apparently now there will be no more UI version of the Noto base font (latin script…): https://github.com/googlei18n/noto-fonts/pull/985

In a nutshell:
- “Noto Sans UI” was always the same thing (same file with another name) as “Noto Sans”. (We suspected that yet…)
- “Noto Sans UI” won’t be part of the release any-more
- The other UI fonts (mostly for Asian scripts) will continue to exist and maintained

We should therefore replace “Noto Sans UI” by “Noto Sans” in our style.

Looking at our current font list, for the non-latin scripts we have for some scripts the UI font _and_ the normal font as fallback. For other scripts, we have the UI font only. Maybe we could use the same policy for all scripts.

cc @pnorman

sommerluk:
Update Noto to next version “Phase III”

What this PR does:
- Currently we use the UI version only (!) for the base font. Apparently the UI version was identical with the non-UI version, and will not be part of Phase III. So we should be prepared and switch the base font to the non-UI version. Resolves #2908 
- The current font release provides now some more scripts as bold face. These have been added. (Cherokee, Symbols)
- The current font release provides now some more new scripts. These have been added [except Anatolian Hyroglyphs because this is a historic script]. (Adlam, Adlam unjoined, Osage, Symbols2)
- The current font release provides now some more scripts as UI version. These have been added. (Sinhala)
- Where a UI version is available, the non-UI version is now _always_ in the list as fallback. Previously, for some UI-fonts we had the non-UI-fallback, for others not. Now it’s the same thing for all.

Of course there will be warning messages when some of these fonts are not installed – for example on Ubuntu 16.4. On the other hand, this PR makes openstreetmap-carto less depend on a specific font version (and indirectly OS version)…

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -63,15 +63,15 @@ Unifont is used as a last resort fallback, with it's excellent coverage, common 
 
 ### Installation on Ubuntu/Debian
 
-On Ubuntu 16.04 or Debian Testing you can download and install the required fonts except Noto Emoji Regular and Noto Sans Arabic UI Regular/Bold with
+On Ubuntu 16.04 or Debian Testing you can download and install most of the required fonts
 
 ```
 sudo apt-get install fonts-noto-cjk fonts-noto-hinted fonts-noto-unhinted fonts-hanazono ttf-unifont
 ```
 
-Noto Emoji Regular can be downloaded [from the Noto Emoji repository](https://github.com/googlei18n/noto-emoji). Noto Sans Arabic UI Regular/Bold can be downloaded [from the Noto repository](https://github.com/googlei18n/noto-fonts).
+Noto Emoji Regular can be downloaded [from the Noto Emoji repository](https://github.com/googlei18n/noto-emoji).
 
-It might be useful to have a more recent version of the fonts for [rare non-latin scripts](#non-latin-scripts). This can be installed [from source](https://github.com/googlei18n/noto-fonts/blob/master/FAQ.md#where-are-the-fonts).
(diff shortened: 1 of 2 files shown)


math1985:
This is really hard to review for me, are there potential downsides to this?

sommerluk:
The downside would be – as described – the warning message for every font that isn’t installed. (Currently, on Ubuntu 16.4 with the recommanded packages, we get warnings only for Unifont vs. unifont, see #429, and Noto Emoji). I suppose performance for the new font styles does not make a big difference.

Our well-known costumer openstreetmap.org will probably not upgrade its font installation until Ubuntu 18.4.1 is released, so this will not have any visual effect at openstreetmap.org inmediatly anyway.

I’m not sure about having always a double entry for UI _and_ non-UI version… It’s not really nice. An alternative might also be removing all the non-UI duplicates (maybe except Sinhala, because otherwise this would break Sinhala rendering for Ubuntu 16.4).

kocio-pl:
> I’m not sure about having always a double entry for UI and non-UI version… It’s not really nice. An alternative might also be removing all the non-UI duplicates (maybe except Sinhala, because otherwise this would break Sinhala rendering for Ubuntu 16.4).

If that's not needed it's always better if code is more compact.

If that's only about warnings, I'm generally positive about this change.

kocio-pl:
What do you want to do with duplicates? Do you plan to update this code?

sommerluk:
Yes, I’ll do that. (I’m just quite busy currently…)

Lukas Sommer

2017-12-03 8:01 GMT+00:00 kocio-pl <notifications@github.com>:

sommerluk:
Finally updated and rebased. Ready.

kocio-pl:
Basic testing shows that rendering still works, thanks!

PR #2949, opened 2017-11-20
Fix color for man_made icons

math1985:
* Labels for masts and water towers are now the same color of dark gray
  as mast/water tower icons, rather than the current black color.
* Windmill icons are now rendered in man_made gray rather than black

Before:

screen shot 2017-11-20 at 01 32 04 screen shot 2017-11-20 at 01 43 11

After:

screen shot 2017-11-20 at 01 29 31 screen shot 2017-11-20 at 01 41 19

### DIFF
--- a/amenity-points.mss
+++ b/amenity-points.mss
@@ -13,7 +13,7 @@
 @memorials: @amenity-brown;
 @culture: @amenity-brown;
 @public-service: @amenity-brown;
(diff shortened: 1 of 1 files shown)


matthijsmelissen:
I'm reverting this. It seems something went wrong with the commit, as several selectors now miss a text-fill property.

Issue #2663, opened 2017-06-25
Administrative boundary tagging on closed ways


imagico:
I was inspecting the changes in rendering due to fixing #59 a bit further and wonder if we correctly interpret boundary tags on closed ways.  I am in particular thinking about cases like

http://www.openstreetmap.org/way/162691988

and

http://www.openstreetmap.org/way/10364857

We interpret these as administrative unit areas (see the labels) and not just as administrative boundary lines although the mapper clearly intended the administrative boundary tagging to apply to the line and the name tag only to the place=island - and as i interpret them tagging conventions say that administrative units have to be boundary relations and ways with boundary tags are always just boundary lines.

I know this kind of tagging is inherently ambiguous and should probably be avoided (by creating a multipolygon relation for the island) but IMO ideally we should require administrative unit areas to be boundary relations.

Issue #2234, opened 2016-07-17
fix admin line and label mismatch without breaking admin line labelling


nebulon42:
See #2142 and #2233.

Currently some admin line labels may appear without their lines. A fix to use `planet_osm_roads` instead of `planet_osm_polygon` led to broken admin line labelling, because the labels now relied on lines instead of polygons and when adjacent admin entities shared a line this prevented placing the label on the correct side.

The only solution I can think of is to switch over to `planet_osm_polygons` completely for admin line rendering. Since this may have wider implications, it should be discussed beforehand.

jeisenbe:
Also see #3822

jeisenbe:
@pnorman - would the performance be very much worse if we use `planet_osm_polygon` instead of `_roads` for the mid and high zoom admin boundary lines?

pnorman:
I'm not sure.

PR #2950, opened 2017-11-20
Render maritime admin boundaries less prominently

### LINKED ISSUE #621: Maritime borders too prominent
Maritime borders are currently rendered too prominent. This makes the map look ugly. No other map I know off renders maritime borders.

We should either:
- Drop rendering maritime borders altogether (see https://trac.openstreetmap.org/ticket/2249)
- Render them in a less obtrusive color, for example navy blue (see https://trac.openstreetmap.org/ticket/3206)

matthijsmelissen:
This requires rendering admin boundaries based on the individual
lines, rather than the entire polygons.

* This fixes #621
* This possibly impacts solutions to #713, #2234 and #2663.

### DIFF
--- a/admin.mss
+++ b/admin.mss
@@ -1,5 +1,5 @@
 @admin-boundaries: #ac46ac;
-@admin-boundaries-maritime: darken(@water-color,15%);
+@admin-boundaries-maritime: darken(@water-color,20%);
 
 /* For performance reasons, the admin border layers are split into three groups
 for low, middle and high zoom levels.
@@ -17,7 +17,7 @@ overlapping borders correctly.
       background/line-width: 1.2;
       line-join: bevel;
       line-color: @admin-boundaries;
-      [maritime = 'yes'] {
+      [maritime = 'true'] {
         line-color: @admin-boundaries-maritime;
       }
       line-width: 1.2;
@@ -47,7 +47,7 @@ overlapping borders correctly.
       background/line-width: 0.6;
       line-join: bevel;
       line-color: @admin-boundaries;
-      [maritime = 'yes'] {
+      [maritime = 'true'] {
         line-color: @admin-boundaries-maritime;
       }
       line-width: 0.6;
@@ -70,7 +70,7 @@ overlapping borders correctly.
       background/line-color: white;
       background/line-width: 0.4;
       line-color: @admin-boundaries;
-      [maritime = 'yes'] {
+      [maritime = 'true'] {
         line-color: @admin-boundaries-maritime;
       }
       line-join: bevel;
@@ -126,7 +126,7 @@ overlapping borders correctly.
     background/line-width: 2;
     line-join: bevel;
     line-color: @admin-boundaries;
-    [maritime = 'yes'] {
+    [maritime = 'true'] {
       line-color: @admin-boundaries-maritime;
     }
     line-width: 2;
@@ -139,7 +139,7 @@ overlapping borders correctly.
     background/line-width: 2;
     line-join: bevel;
     line-color: @admin-boundaries;
-    [maritime = 'yes'] {
+    [maritime = 'true'] {
       line-color: @admin-boundaries-maritime;
     }
     line-width: 2;
@@ -154,7 +154,7 @@ overlapping borders correctly.
       background/line-width: 1.5;
       line-join: bevel;
       line-color: @admin-boundaries;
-      [maritime = 'yes'] {
+      [maritime = 'true'] {
         line-color: @admin-boundaries-maritime;
       }
       line-width: 1.5;
@@ -166,7 +166,8 @@ overlapping borders correctly.
   comp-op: darken;
 }
 
-#admin-high-zoom[zoom >= 13] {
+#admin-high-zoom[zoom >= 13],
+#admin-high-zoom-maritime[zoom >= 13] {
   [admin_level = '9'],
   [admin_level = '10'] {
     [zoom >= 13] {
@@ -175,7 +176,7 @@ overlapping borders correctly.
       background/line-width: 2;
       line-join: bevel;
       line-color: @admin-boundaries;
-      [maritime = 'yes'] {
+      [maritime = 'true'] {
         line-color: @admin-boundaries-maritime;
       }
       line-width: 2;
--- a/admin.mss
+++ b/admin.mss
@@ -1,4 +1,5 @@
 @admin-boundaries: #ac46ac;
+@admin-boundaries-maritime: darken(@water-color,15%);
 
 /* For performance reasons, the admin border layers are split into three groups
 for low, middle and high zoom levels.
@@ -16,6 +17,9 @@ overlapping borders correctly.
       background/line-width: 1.2;
       line-join: bevel;
       line-color: @admin-boundaries;
+      [maritime = 'yes'] {
+        line-color: @admin-boundaries-maritime;
+      }
       line-width: 1.2;
     }
     [zoom >= 5] {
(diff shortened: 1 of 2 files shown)


math1985:
(Screenshots with only zoom_level 4 admin boundaries loaded in the database.)

Before:
screen shot 2017-11-20 at 16 31 42

After:
screen shot 2017-11-20 at 16 25 21

math1985:
@nebulon42 @pnorman You have been working on admin boundaries, do you foresee any problems with this in approach?

In particular @nebulon42, you added the `AND osm_id < 0` requirement, what was the reason for this line of code?

kocio-pl:
Do you really mean: Split off docs for contributors #723?

systemed:
Very good to see this.

Is there a reason to show them at all up to z6/7? I can see a rationale for z8+ as it helps identify which country governs certain small islands (though even that's an edge case, and cartographically not one I personally think is worth the sacrifice), but such cases, which are extensively cited in #621, don't especially require the rendering to be made explicit at small scales.

math1985:
> Do you really mean: Split off docs for contributors #723?

I meant #713. Fixed.

math1985:
> Is there a reason to show them at all up to z6/7? 

I have no strong opinion on that. Perhaps for the islands in between Greece and Turkey, having the border already at z6/z7 might be useful.

kocio-pl:
I would like to see worldwide rendering to make sure, but this change looks OK for me in general.

math1985:
@rrzefox would it be possible to test this on your server?

nebulon42:
It depends on the data. The `osm_id < 0` vs. `osm_id > 0` could indeed make problems. I found that admin boundary information on ways as in contrast to relations was incomplete. Especially in Poland and maybe other regions I don't recall the data was mostly exclusively on relations. I added the < 0 to render only relations as before both relations and ways were rendered which led to duplicated borders and because of the transparency this looked ugly. So the < 0 led to a cleaner rendering and I still think it is important to either render < 0 or > 0.

When I worked on maritime boundaries I also found that `maritime = yes` was often incomplete. But a different rendering could change that.

So no problems in particular, but double-check on a global rendering. I suspect that we lose quite some amount of boundaries in some regions when switching to rendering boundary ways only.

math1985:
I see indeed admin_level's missing on boundary (ways) in Poland. Thanks for pointing that out.

math1985:
All the gray ways on http://product.itoworld.com/map/2 have no admin_level on the boundary ways This applies indeed to Poland, but also to other countries.

math1985:
I wonder how @cquest solved this issue for the French style?

imagico:
> The osm_id < 0 vs. osm_id > 0 could indeed make problems. I found that admin boundary information on ways as in contrast to relations was incomplete.

The tagging of boundary relations is not only more consistent than that of boundary ways, it is also much more *significant*.  Remove tags from all boundary ways and you will still have perfectly usable boundary information in the database based on the relations.  Remove the tags from the boundary relations and you will loose almost everything.  Therefore #1989 was a significant step in terms of mapper feedback.

cquest:
@math1985 I wonder myself ;)

Have a look at the queries near https://github.com/cquest/osmfr-cartocss/blob/master/osmfr.yml#L589

I also did not want to render the boundary portion that are also the coastline... which needs a check in the rels table.

Some maritime boundaries will not render as expected, but it this is because of their incomplete tags, it's a good way to reveal them ;)

math1985:
> @math1985 I wonder myself ;)

It seems you're using admin boundaries (not relations) only, meaning the French style does not render the regional admin borders in Poland.

math1985:
Render demo now at  http://bl.ocks.org/math1985/raw/af7a602c222dbf1ff1a2c0d84ed755b7/. Thanks a lot @rrzefox!

math1985:
And this shows why it's good to do full-planet demos. While the maritime borders seem to look alright, in a lot of other places problems with the land borders show up.

math1985:
@rrzefox I'm not sure how important regional borders are for the primary purpose of your tile server, but feel free to roll back this change if necessary.

math1985:
Another issue I see now is that dashed borders consisting of many short lines look much better when rendered as polygon rather than as individual lines. This is because when they are rendered as individual lines, the dash pattern starts again with every new line. This leads to solid lines for borders consisting of sections that are shorter than an individual dash.

kocio-pl:
If we want to change this color, #2951 makes sense, since despite different properties ferry lines look quite similar. Some internal waters, like Baltic, Mediterranean Sea or Black Sea, look too busy for me on z7 with ferries, and [in some cases](http://bl.ocks.org/math1985/raw/af7a602c222dbf1ff1a2c0d84ed755b7/#10.00/54.3753/-347.8539) only z10 is OK, because the border gets wide then.

kocio-pl:
Another idea: maybe the line should be thicker to compensate the pale color a bit? Color change makes it much less aggressive anyway, so it would be nice to still look different than ferry lines.

kocio-pl:
> I see indeed admin_level's missing on boundary (ways) in Poland. Thanks for pointing that out.

This information is available as a relation tag:

http://www.openstreetmap.org/relation/130935#map=8/52.265/21.195

Do we have a problem with such solution?

math1985:
Yes, obviously in the relation tag we cannot see what parts of the boundary are maritime.

I’m inclined to encourage the Polish community to add the missing data, especially as most countries do have the data also on lines.

kocio-pl:
I have already asked about it, when I was not aware that it is done in relations:

https://forum.openstreetmap.org/viewtopic.php?pid=673735#p673735

mboeringa:
Maybe a dumb question, but why not render the maritime borders as a slightly broader opaque line based on admin boundary ways **_on top_** of the relation based non-maritime polygon borders? If generalization is applied, this may cause issues with alignment, but if none, it might work?

pnorman:
> @nebulon42 @pnorman You have been working on admin boundaries, do you foresee any problems with this approach?

There are areas with missing data on ways, like Poland. We rejected going to ways because of the problems there in the past.

Currently what I'm doing for admin boundaries is to take the ways from all admin relations, find min admin_level of parent relations, and apply information like disputed or maritime from way tagging.

kocio-pl:
Another side effect - the USA/Canada border is invisible in some places, especially on Big Lakes and between Montreal and lake Ontario, because on some zoom levels you don't even see the water line:

http://bl.ocks.org/math1985/raw/af7a602c222dbf1ff1a2c0d84ed755b7/#5.00/44.238/-80.454

rrzefox:
I have reverted this patch, because it causes rendering failures on high zoom levels:

> ERROR:  column "maritime" does not exist   Full sql was: 'SELECT ST_AsBinary("way") AS geom,"admin_level","maritime" FROM (SELECT way, admin_level, boundary FROM planet_osm_roads WHERE boundary = 'administrative'  AND admin_level IN ('0', '1', '2', '3', '4', '5', '6', '7', '8', '9', '10')#012    AND osm_id > 0 ORDER BY admin_level::integer DESC -- With 10 as a valid value, we need to do a numeric ordering, not a text ordering) AS admin_high_zoom

Im guessing this is probably missing a "tags->'maritime' as maritime" somewhere in project.mml...

matthijsmelissen:
I updated this.

* Admin rendering based on lines instead of relatinos does not work (lack of
  data in i.e. Poland, and ugly rendering of dashes in case of short sections).
  Instead, this renders maritime borders based on lines, on top of the full
  border based on the polygon.
* Fix error on high zoom levels.
* Render maritime border slightly darker.

kocio-pl:
Thanks! This version looks nice and I no longer see the problem with borders being similar to ferry lines. Of course worldwide testing is needed to make sure (for example with the problem mentioned in  https://github.com/gravitystorm/openstreetmap-carto/pull/2950#issuecomment-346846090) - @rrzefox Would you like to test it on your server (probably with #3057 too)?

z7
pu48dntn
z8
nv7qcuwn

matthijsmelissen:
@rrzefox No, no need to test yet. I still plan to make some changes.

matthijsmelissen:
Superseded by #3102.

PR #2951, opened 2017-11-20
Move ferries to zoomlevel 8

math1985:
This moves the minimum zoomlevel of ferries from zoomlevel 7 to 8.

### DIFF
--- a/ferry-routes.mss
+++ b/ferry-routes.mss
@@ -2,7 +2,7 @@
 @ferry-route-text: @ferry-route;
 
 #ferry-routes {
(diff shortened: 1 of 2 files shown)


pnorman:
This is not ideal in BC, where we have long ferry routes as key parts of the transportation network.

image

However, it's probably the best compromise given different factors. It ends up bringing ferry routes in at the same time as primary roads.

No review, since I haven't done a code review or run it myself.

math1985:
Thanks for looking at it at least!

Issue #2925, opened 2017-11-04
Low-zoom: decide which features to render


matthijsmelissen:
Some features that we currently render for which we do not (yet) have a use case:
* Province/state names and borders on z5
* Roads on z5
* National parks on z7
* Railways on z7
* Minor water areas on z5/6/7
* Glaciers and icesheets on z6 and z7
* Military areas on z7
* Ferry routes on z7
* Island names on z5/6/7

In which cases do we want to add use cases, and in which cases do we want to drop rendering?

CC @kocio-pl

math1985:
My opinion:

> Province/state names and borders on z5

This really depends on the area. I guess we cannot avoid rendering for larger countries such as the US, and as we have no way to distinguish country size currently, we should render all borders.

> Roads on z5

I'm in favour of dropping rendering.

> National parks on z7

I'm in favour of dropping. I fail to see how the current rendering is useful for any purpose.

> Railways on z7

I guess people who like tracing railway lines to see along which route they are travelling (including me) can zoom in one zoom level more. So I'd say drop.

> Minor water areas on z5/6/7

Guess nothing we can do about.

matkoniecz:
> This really depends on the area. I guess we cannot avoid rendering for larger countries such as the US, and as we have no way to distinguish country size currently, we should render all borders.

documented in #2926

matkoniecz:
My opinion:

> Province/state names and borders on z5

It depends on location and I am tempted to introduce some form of hardcoding here. Where (except USA) showing internal division is useful?

> Roads on z5

advocatus diaboli: it nicely shows population density (seee for example USA http://www.openstreetmap.org/#map=5/41.212/-95.010 )

but given that strongest argument is really weak - I think that it may be dropped

> National parks on z7

I'm in favour of dropping since #1669. I fail to see how the current rendering is useful for any purpose.

> Railways on z7

Here I would be against change. At this scale showing railway structure IMHO makes sense.

> Minor water areas on z5/6/7

Depends on what one considers as minor. But given technical problems for now I will ignore this issue.

math1985:
> > Railways on z7

> Here I would be against change. At this scale showing railway structure IMHO makes sense.

What is the use case? :)

matkoniecz:
> What is the use case? :)

Showing structure of railway network - but given that using "it is interesting for me" one may justify showing anything, from showing density of stork nests to showing capacity of bicycle parkings it is not a real usecase :(

[...]

PR #2952, opened 2017-11-20
Do not render small water areas on low zoom

matthijsmelissen:
On zoom levels 0-7, don't render small water areas (smaller than 32px).

### DIFF
--- a/water.mss
+++ b/water.mss
@@ -48,7 +48,7 @@
   [waterway = 'riverbank']::waterway {
     [zoom >= 0][way_pixels >= 4],
     [zoom >= 1][way_pixels >= 16],
-    [zoom >= 2][way_pixels >= 32],
+    [zoom >= 2][way_pixels >= 64],
     [zoom >= 8] {
       polygon-fill: @water-color;
       [way_pixels >= 4] {
--- a/water.mss
+++ b/water.mss
@@ -46,9 +46,9 @@
   [natural = 'water']::natural,
   [landuse = 'reservoir']::landuse,
   [waterway = 'riverbank']::waterway {
-    [zoom >= 0][way_pixels >= 4],
-    [zoom >= 1][way_pixels >= 16],
-    [zoom >= 2][way_pixels >= 64],
+    [zoom >= 0][zoom < 1][way_pixels >= 4],
+    [zoom >= 1][zoom < 2][way_pixels >= 16],
+    [zoom >= 2][zoom < 8][way_pixels >= 32],
     [zoom >= 8] {
       polygon-fill: @water-color;
       [way_pixels >= 4] {
--- a/water.mss
+++ b/water.mss
@@ -30,20 +30,24 @@
     }
   }
 
-  [landuse = 'basin'][zoom >= 7]::landuse {
-    polygon-fill: @water-color;
-    [way_pixels >= 4] {
-      polygon-gamma: 0.75;
-    }
-    [way_pixels >= 64] {
-      polygon-gamma: 0.6;
+  [landuse = 'basin']::landuse {
+    [zoom >= 7][way_pixels >= 32],
+    [zoom >= 8] {
+      polygon-fill: @water-color;
+      [way_pixels >= 4] {
+        polygon-gamma: 0.75;
+      }
+      [way_pixels >= 64] {
+        polygon-gamma: 0.6;
+      }
     }
   }
 
   [natural = 'water']::natural,
   [landuse = 'reservoir']::landuse,
   [waterway = 'riverbank']::waterway {
-    [zoom >= 0] {
+    [zoom >= 0][way_pixels >= 32],
+    [zoom >= 8] {
       polygon-fill: @water-color;
       [way_pixels >= 4] {
         polygon-gamma: 0.75;
--- a/water.mss
+++ b/water.mss
@@ -46,7 +46,9 @@
   [natural = 'water']::natural,
   [landuse = 'reservoir']::landuse,
   [waterway = 'riverbank']::waterway {
-    [zoom >= 0][way_pixels >= 32],
+    [zoom >= 0][way_pixels >= 4],
+    [zoom >= 1][way_pixels >= 16],
+    [zoom >= 2][way_pixels >= 32],
     [zoom >= 8] {
       polygon-fill: @water-color;
       [way_pixels >= 4] {


math1985:
@rrzefox would it be possible to test this on your server?

kocio-pl:
Could you also publish some before/after renderings? Probably the places that you gave as example of problems with small water areas would be good.

math1985:
@kocio-pl asked in #2925:
> what filter do we want to use and up to which zoom level?

It currently bothers me both on zoom levels 6 and 7, and perhaps 8 and 9. So i think we definitely need measures for the zoom levels up to 7 (on levels 0-5 the small water areas are not very useful to me either).

I eyeballed 32px as a cut-off, let's see how it looks like when we have a rendering demo.

math1985:
> Could you also publish some before/after renderings?

I don't have enough data in my database at the moment for realistic example renderings at these zoom levels.

rrzefox:
> @rrzefox would it be possible to test this on your server?

Done, and Z0-8 have already been rerendered.

math1985:
Cool! Demo at http://bl.ocks.org/math1985/raw/af7a602c222dbf1ff1a2c0d84ed755b7/.

kocio-pl:
Unfortunately it reintroduces #2580 on z0-z2.

math1985:
This looks like an improvement to me.

kocio-pl:
On z0-z1 we don't see any of the biggest lakes (except Caspian, of course). On z2 Ontario, Erie and Onega are still not visible.

kocio-pl:
I also think that filtering produces nice effects, so it's worth applying. On the other hand La Plata and big lakes are special, so it would be good to preserve them. The simplest solution would be to start filtering from z3 instead of z0. What do you think about it?

math1985:
Yes, I agree.

water.mss
```diff
@@ -30,20 +30,24 @@
...
-  [landuse = 'basin'][zoom >= 7]::landuse {
-    polygon-fill: @water-color;
-    [way_pixels >= 4] {
-      polygon-gamma: 0.75;
-    }
-    [way_pixels >= 64] {
-      polygon-gamma: 0.6;
+  [landuse = 'basin']::landuse {
+    [zoom >= 7][way_pixels >= 32],
+    [zoom >= 8] {
```
kocio-pl: > On zoom levels 0-8, don't render small water areas (smaller than 32px).

So it should  be:

` [zoom >= 9] { `

water.mss
```diff
@@ -30,20 +30,24 @@
...
+      }
     }
   }
 
   [natural = 'water']::natural,
   [landuse = 'reservoir']::landuse,
   [waterway = 'riverbank']::waterway {
-    [zoom >= 0] {
+    [zoom >= 0][way_pixels >= 32],
+    [zoom >= 8] {
```
kocio-pl: The same as above: `[zoom >= 9] {`

rrzefox:
So some screenshot for later reference:
Z1 old:
z1-old
Z1 new:
z1-new
The big lakes between .us and .ca disappeared.
Z3 old:
z3-old
Z3 new:
z3-new
On Z3, all the mini-waters in Canada are gone. The same happens in northern europe (no screenshot).
In .nl, the Ijsselmeer disappears, it appears on Z4.

I have now deployed a slightly different version of this PR on the "demo server", it uses this less aggressive filtering:
```
  [natural = 'water']::natural,
  [landuse = 'reservoir']::landuse,
  [waterway = 'riverbank']::waterway {
    [zoom >= 0][way_pixels >= 4],
    [zoom >= 1][way_pixels >= 16],
    [zoom >= 2][way_pixels >= 32],
    [zoom >= 8] {
```

kocio-pl:
Thanks! I was also thinking about progressive filtering and with these values it seems to be the best of both worlds, indeed. I would merge this version.

kocio-pl:
@matthijsmelissen Do you plan to embrace the @rrzefox proposition? If it's just lack of time, I can fine tune the code and merge it soon.

matthijsmelissen:
Generally I prefer to have a factor 4 between each zoom level (with each zoom in, lines gets twice as large so areas get 4 times as large ). So either 4-16-64 or 2-8-32. 

The exact numbers apart, I'm happy with @rrzefox's approach. I was planning to get back at this some time, but I'd be happy if you can pick it up earlier.

kocio-pl:
@rrzefox Were there some quirks which would be visible with 4-16-64 or 2-8-32 approach or was it just a free form handcrafting? If the latter case, could you test one of the proposed "4x" scheme?

rrzefox:
For reference, this is how it looks with 4-16-32:
z0
z1
z2

While I also tried to make steps with a factor of four, I found deviating from that in the last step acceptable. I had tried 2-8-32 but not liked the result because it did not filter enough. But see for yourself, I deployed that now... I'll deploy 4-16-64 on Friday.

kocio-pl:
OK, let's see the 4-16-64 then.

BTW - because Matthijs has changed the account name, the link to compare is different now:
 
http://bl.ocks.org/matthijsmelissen/raw/af7a602c222dbf1ff1a2c0d84ed755b7/#2.00/18.8/14.7

matthijsmelissen:
4-16-32 works in fact quite well, I added it to the PR now.

matthijsmelissen:
>On zoom levels 0-8, don't render small water areas (smaller than 32px).

I changed the documentation, it is now 0-7.

rrzefox:
I've now deployed 4-16-64 for testing (Z0-3 rerendered)

kocio-pl:
It looks OK for me. 

On z0 La Plata is not visible, but the border too, so it's not a problem. Lack of Ontario, Erie, Ladoga, Onega and Victoria on z0 is also not visible if you're not looking for them. Anything on z1+ looks decent in both 4-16-32 and 4-16-64 for me, so it's up to you which version do you like more, Matthijs.

StyXman:
Me being from Argentina I can say that this is a nice improvement, and I don't think anybody would complain about Río de la Plata not being present in z0.

matthijsmelissen:
```
    [zoom >= 0][way_pixels >= 4],
    [zoom >= 1][way_pixels >= 16],
    [zoom >= 2][way_pixels >= 32],
```
@rrzefox This does not work. It will still render water areas of size 8 on z1 because the first line is still true.

matthijsmelissen:
@rrzefox I have fixed this in my branch, would you be able to deploy my branch to your server?

I used 4-16-32, as filtering with 4-16-64 will almost certainly be too aggressive.

kocio-pl:
> @rrzefox I have fixed this in my branch, would you be able to deploy my branch to your server?

It was already tested and the screenshots are here: https://github.com/gravitystorm/openstreetmap-carto/pull/2952#issuecomment-349750757

matthijsmelissen:
Nope, that test contained the bug I mentioned in https://github.com/gravitystorm/openstreetmap-carto/pull/2952#issuecomment-350469736 .

matthijsmelissen:
Just to be clear, the 4-16-32 deploy actually showed 4-4-4, the 2-8-32 deploy actually showed 2-2-2, and the 4-16-64 deploy actually showed 4-4-4 again.

rrzefox:
The fixed 4-16-32 version has been deployed and Z0-5 rerendered.

kocio-pl:
That's OK for me, we can merge it any time soon.

matthijsmelissen:
@rrzefox Thanks!

@kocio-pl If you Approve it, I can merge it.

kocio-pl:
I'm confused, it looks like z1 and z2 are wrong now - La Plata and smaller Big Lakes are missing now on both proposed and current versions (OSMF re-rendering is happening right now)...

http://bl.ocks.org/matthijsmelissen/raw/af7a602c222dbf1ff1a2c0d84ed755b7/#2.00/27.4/42.3

kocio-pl:
Another thing that we could make better would be to not filter riverbanks, as suggested here:

http://www.openstreetmap.org/user/imagico/diary/42975#comment40531

matthijsmelissen:
My intention was not to render riverbanks at all. The solution is exactly the opposite of what @imagico suggests: these riverbanks are way too big, and should be cut up in smaller parts.

kocio-pl:
Current code still allows showing them, so maybe it'd be better to get rid of them completely on some initial zoom levels to avoid problems with some segments being visible and some others invisible.

matthijsmelissen:
Makes sense.

matthijsmelissen:
> I'm confused, it looks like z1 and z2 are wrong now

Do you have any idea what could cause this?

kocio-pl:
I guess we just got lost with testing all the possibilities on @rrzefox server. :smile: It was probably some small, stupid error but nobody checked it on its own machine.

Issue #919, opened 2014-09-02
own colour for each leisure type (sport)


Newbie07:
I have sports_centre with lots of different pitches and tracks and a stadium. But stadium and sport_centre have the same colour.
1. Please assign different colours to stadium and sport_centre
2. Additionally leisure=sports_centre and building=\* should be rendered differently than building=yes

Thanks

oddityoverseer13:
One thing to note for this issue is some fields support multiple sports. Those can be tagged with either sport=multi or sport=sport1;sport2

kocio-pl:
> a outline could work but might also interfere with barrier=fence.

They have an outline defined and it's lighter than barriers:
https://github.com/gravitystorm/openstreetmap-carto/blob/9d5795de44d6ea4c3545ed7e03c8e2d2d47e1d15/landcover.mss#L577-L580

kocio-pl:
I don't think we have enough colors for this, so I will close it. Leisure/sport areas have light green currently and it's not so hard to distinguish them by icon and size most of the time, but it's not possible to use more colors for them. The more important thing is to know that light green is for leisure and sport in general.

PR #2954, opened 2017-11-22
Change sports_centre and stadium color to green

kocio-pl:
Now that we use light green for different small leisure areas, we could use it also for big ones, as they typically don't overlap. 

That unification would also offload societal_amenities color a bit (so we have less problem defining what it really is) and tie green color closer with the physical activities, which makes the system more predictable.

Examples in Warsaw (with different landuses around):

Before
hr2sr3by
After
qgxsvwqh

Before
ao4jenlr
After
8ix4kxxz

Before
obc7earf
After
ru3lypoj

Before
e01zrfvz
After
marl4vhc

### DIFF
--- a/landcover.mss
+++ b/landcover.mss
@@ -61,7 +61,7 @@
 
 @pitch: #aae0cb; // also track
 @track: @pitch;
-@stadium: @societal_amenities; // also sports_centre
+@stadium: @leisure; // also sports_centre
 @golf_course: #b5e3b5;
 
 #landcover-low-zoom[zoom < 10],


kocio-pl:
Additional question: should we use italic font for such names? Is it kind of standard for areas (you can see it with school, park, residential and allotments)?

lakedistrictOSM:
This looks good, it makes it different from schools and hospitals now. Italics are probably a good idea until we introduce a symbol.

Will this PR change the way leisure=sports_centre + building=yes is rendered?

kocio-pl:
No, it changes a color only in the places, where yellowish is visible now.

polarbearing:
You mean it would bring us closer to render societal amenities campuses, ref #1624 ?

I did not like the yellow for sports so much (which happened only recently #2071), thus the green is better. It still does not distinguish a stadium within a surrounding sports_centre #919.

PR #2964, opened 2017-11-30
Better shades for recreation_ground and other green areas

### LINKED ISSUE #2956: Shade of green for recreation_ground
It looks like landuse=recreation_ground is currently defined twice:

https://github.com/gravitystorm/openstreetmap-carto/blob/f3adce26d6bbfbd0d5f260753527e09b2fb7e965/landcover.mss#L253-L260

https://github.com/gravitystorm/openstreetmap-carto/blob/f3adce26d6bbfbd0d5f260753527e09b2fb7e965/landcover.mss#L332-L344

And it's supposed to be the same as park if we look at the color variables:
https://github.com/gravitystorm/openstreetmap-carto/blob/f3adce26d6bbfbd0d5f260753527e09b2fb7e965/landcover.mss#L7

but it effectively renders as grass.

I think we should really use park or leisure color here (see #2954). Which one do you think would be better?

kocio-pl:
Resolves #2956.

This should make the use of green backgrounds more consistent:
- `landuse=recreation_ground` and `leisure=recreation_ground` should look the same and use leisure green
- `leisure=garden` should use park green

[Update:] `leisure=common` changes are reverted (needs additional discussion)

landuse=recreation_ground
Before
a0b_vfza
After
stogufma

leisure=recreation_ground
Before
g6vxnl91
After
qpnahivb

leisure=garden
Before
g2hddlgc
After
b fesehn

### DIFF
--- a/landcover.mss
+++ b/landcover.mss
@@ -1,10 +1,10 @@
 // --- Parks, woods, other green things ---
 
-@grass: #cdebb0; // also meadow, common, garden, village_green
+@grass: #cdebb0; // also grassland, meadow, common, village_green
 @scrub: #b5e3b5;
 @forest: #add19e;       // Lch(80,30,135)
 @forest-text: #46673b;  // Lch(40,30,135)
-@park: #c8facc;         // Lch(94,30,145) also recreation_ground
+@park: #c8facc;         // Lch(94,30,145) also garden
 @orchard: #aedfa3; // also vineyard, plant_nursery
 
 // --- "Base" landuses ---
@@ -97,6 +97,8 @@
     [way_pixels >= 64] { polygon-gamma: 0.3;  }
   }
 
+  [feature = 'leisure_recreation_ground'][zoom >= 10],
+  [feature = 'landuse_recreation_ground'][zoom >= 10],
   [feature = 'leisure_playground'][zoom >= 13],
   [feature = 'leisure_fitness_station'][zoom >= 13] {
     polygon-fill: @leisure;
@@ -251,7 +253,7 @@
   }
 
   [feature = 'leisure_park'],
-  [feature = 'leisure_recreation_ground'] {
+  [feature = 'leisure_garden'] {
     [zoom >= 10] {
       polygon-fill: @park;
       [way_pixels >= 4]  { polygon-gamma: 0.75; }
@@ -332,10 +334,8 @@
   [feature = 'landuse_meadow'],
   [feature = 'natural_grassland'],
   [feature = 'landuse_grass'],
-  [feature = 'landuse_recreation_ground'],
   [feature = 'landuse_village_green'],
-  [feature = 'leisure_common'],
-  [feature = 'leisure_garden'] {
+  [feature = 'leisure_common'] {
     [zoom >= 10] {
       polygon-fill: @grass;
       [way_pixels >= 4]  { polygon-gamma: 0.75; }


matthijsmelissen:
+1

I'd be happy with dropping leisure=common altogether though (due to it's overlap with different tags, UK-centered definition, and misuse of the tag both in and outside the UK).

landcover.mss
```diff
@@ -251,14 +254,14 @@
...
-  [feature = 'leisure_recreation_ground'] {
+  [feature = 'leisure_garden'] {
     [zoom >= 10] {
       polygon-fill: @park;
       [way_pixels >= 4]  { polygon-gamma: 0.75; }
       [way_pixels >= 64] { polygon-gamma: 0.3;  }
     }
   }
-
+
```
matthijsmelissen: You're introducing spurious whitespace

matthijsmelissen:
Do the labels need changing color too?

kocio-pl:
> I'd be happy with dropping leisure=common altogether though (due to it's overlap with different tags, UK-centered definition, and misuse of the tag both in and outside the UK).

i think it's better to make people aware of the problem on the Tagging list before doing this, since it has ~57k uses, but you're right, it's really not clear scheme.

kocio-pl:
BTW - what about landuse=village_green (86k uses)? It is also rather local historical quirk and for example in Poland it was misused as a city green too because of a broken translation. Maybe we should make another ticket for them, what do you think?

> Do the labels need changing color too?

Probably yes, i will check it later.

matthijsmelissen:
Yes, landuse=village_green should be abandoned too (in coordination with tagging@).

kocio-pl:
Just a history snapshot then:

taghistory 9

and a link to discussion:

https://lists.openstreetmap.org/pipermail/tagging/2017-November/thread.html#34274

polarbearing:
The issue with ```village_green``` are related to the refusal #2548 to render ```landcover=*```.

kocio-pl:
I have removed leisure=common changes, as it needs a discussion on the list first. 

Label colors are the same in this group, so they don't need any change.

drkludge:
I disagree with abandoning landuse=village_green.  The first time I saw village_green it was described as an area set aside.  I have not see the English legal description until now.  I'd rather clear up the definition for areas outside of England.

image

You see I have areas like this one.  It was tagged as a park years ago but that implies that the city maintains the area as such.  Nope that is not the use.  It really is a buffer.  But the retention basin use is .05 percent of the year.  The real use is a green area that has been set aside so that a land developer is not creating ticky tack subdivisions of houses crowded upon houses.  Also note that the land developer had to set aside a path to the larger park north of the second picture.  landuse=village_green is perfect for this area as the area is more than just a retention basin.  As the area stands now tagged with leisure=park, the area is just another Pokemon spawn point.

image

dieterdreist:
2017-12-11 1:15 GMT+01:00 drkludge <notifications@github.com>:

> You see I have areas like this one. It was tagged as a park years ago but
> that implies that the city maintains the area as such.



From the picture I would also say park, and no, it doesn't imply the city
maintains the area.



> Nope that is not the use. It really is a buffer. But the retention basin
> use is .05 percent of the year.



being a park also a buffer/rentention basin isn't an issue either.

chrisfleming:
sorry jumping in on this late. I generally liked the difference in colours between gardens which tend to be private and parks which are generally public. So it was possible to identify public parks versus private gardens.   Also I think that the old garden colour worked better with buildings, hence looked good when mapoping private gardens?

polarbearing:
garden and park says nothing about public or private. You would need access tagging for this.

matkoniecz:
#3022 requested partial revert, people that commented here may wish to comment at this ticket

Issue #2965, opened 2017-11-30
Bus Guideways disappear at z12 and less


SomeoneElseOSM:
As noted elsewhere, there aren't too many of these in OSM at the moment, but it would make sense to show them at z11 and z12.  Here's Cambridge at z12 now:

http://www.openstreetmap.org/#map=12/52.2625/0.0359

The guided busway to St Ives is as big a feature in the landscape as (say) the railway line out through Cherry Hinton, it would make sense to render it.  Currently there's no rendering below z13:

https://github.com/gravitystorm/openstreetmap-carto/blob/master/roads.mss#L2582

but adding this to cover z11 and z12 would be a big improvement:

    [zoom >= 11][zoom < 13] {
      line-width: 0.6;
      line-color: #6666ff;
      [zoom >= 12] { line-width: 1; }
    }

as per

https://github.com/SomeoneElseOSM/openstreetmap-carto-AJT/blob/master/roads.mss#L3094

This doesn't fix the other problems with guideways (due I suspect to it being in its own layer): https://github.com/gravitystorm/openstreetmap-carto/issues/2489 https://github.com/gravitystorm/openstreetmap-carto/issues/2897 , but it's a start.

matthijsmelissen:
In general that rendering is quite awful.

kocio-pl:
I don't see what's wrong with using railway scheme in the bus-related color?

matthijsmelissen:
The bus guideway is now the most dominant feature on z13.

SomeoneElseOSM:
> In general that rendering is quite awful.

Hehe - why don't you say what you really think :)

> The bus guideway is now the most dominant feature on z13.

Yes - the problem seems to be that the "guideways" layer used the railway line characteristics at z13 but with another colour, and did not take into account how much more "in your face" #6666ff looks when compared to the darkish grey used for railway=rail.

> I don't see what's wrong with using railway scheme in the bus-related color?

FWIW that's what I've done in the style that I look after - send it through as "railway=bus_guideway" (mainly to make the selection in the .mml a bit easier to understand) and use the existing railway code to handle bridges and tunnels etc.  I used #6699ff as a "less in your face colour" (though that may not work with the different water colours that OSM Carto uses) and, as above, used a significantly narrower line at zooms < 13 so that the Cambridge bus guideway is "about as obvious at those zooms as an unclassified road".

kocio-pl:
[...]

PR #2966, opened 2017-12-03
Bus Guideways

SomeoneElseOSM:
Make them visible from z11, but not have them dominate the map.
Involves a colour change the also affects other zoom levels.
Resolves https://github.com/gravitystorm/openstreetmap-carto/issues/2965 .

### DIFF
--- a/roads.mss
+++ b/roads.mss
@@ -2579,9 +2579,14 @@ tertiary is rendered from z10 and is not included in osm_planet_roads. */
 }
 
 #guideways {
+  [zoom >= 11][zoom < 13] {
+    line-width: 0.6;
+    line-color: #6699ff;
+    [zoom >= 12] { line-width: 1; }
+  }
   [zoom >= 13] {
     line-width: 3;
-    line-color: #6666ff;
+    line-color: #6699ff;
     line-join: round;
     b/line-width: 1;
     b/line-color: white;
--- a/project.mml
+++ b/project.mml
@@ -1067,7 +1067,7 @@ Layer:
           WHERE highway = 'bus_guideway'
         ) AS guideways
     properties:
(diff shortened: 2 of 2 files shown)


kocio-pl:
Please add the rendering here (I guess you did it in the #2965, but I'm not sure).

SomeoneElseOSM:
Yes, see https://github.com/gravitystorm/openstreetmap-carto/issues/2965 for the examples of the rendering at z11, z12, z13 and z18.

kocio-pl:
I think it's always better to just include them here to make it clear (and before/after would be good too):

z11
camb_guideway_11_1
z12
camb_guideway_12_1
z13
camb_guideway_13_1
z18
camb_guideway_18_1

roads.mss
```diff
@@ -2579,9 +2579,14 @@ tertiary is rendered from z10 and is not included in osm_planet_roads. */
 }
 
 #guideways {
+  [zoom >= 11][zoom < 13] {
```
kocio-pl: I guess `[zoom < 13] ` is not needed, because next condition will just take over.

SomeoneElseOSM: It's actually just a cut and paste effect (I got that line of code from the railway code where "low zoom" is handled separately), but yo could also call it defensive coding :)

Interestingly, less mapnik.xml is created with the extra test, so I'm inclined to leave it in:

with:
wc mapnik.xml
45390  161881 2617683 mapnik.xml

without:
wc mapnik.xml
45396  161894 2618000 mapnik.xml

matthijsmelissen:
@SomeoneElseOSM Thanks!

SomeoneElseOSM:
Thanks for merging!

PR #2969, opened 2017-12-06
Replace VMAP0-data by OSM-data

### LINKED ISSUE #256: Phantom spots on map in Levels 8 and 9
Please see this map:
http://www.openstreetmap.org/#map=9/37.4400/68.3487

In the map center, you can see a number of grey spots (usually representing something like residential areas). But in OSM data, they do not exist. I made sure the tiles get re-rendered. They appear on map levels 8 and 9.

Actually, for the cases I checked, the areas represented would make sense for settlements, so I wonder where they disappeared. But that's a different matter.

matthijsmelissen:
Currently, we use VMAP0-data instead of OSM-data to indicate built-up areas
on zoom levels 8 and 9. This PR replaces the VMAP0-data with the OSM
landcover data.

Resolves #256.

### DIFF
--- a/project.mml
+++ b/project.mml
@@ -70,17 +70,6 @@ Layer:
       type: shape
     properties:
       minzoom: 10
-  - id: builtup
-    geometry: polygon
-    extent: *world
-    srs-name: mercator
-    srs: "+proj=merc +datum=WGS84 +over"
-    Datasource:
-      file: data/world_boundaries/builtup_area.shp
-      type: shape
-    properties:
-      minzoom: 8
-      maxzoom: 9
   - id: necountries
     geometry: linestring
     <<: *extents84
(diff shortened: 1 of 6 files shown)


HolgerJeromin:
Is there documentation to be updated?

project.mml
```diff
@@ -101,12 +101,13 @@ Layer:
             COALESCE(wetland, landuse, "natural") AS feature
           FROM (SELECT
               way, COALESCE(name, '') AS name, religion,
-              ('landuse_' || (CASE WHEN landuse IN ('forest', 'military', 'farmland') THEN landuse ELSE NULL END)) AS landuse,
+              ('landuse_' || (CASE WHEN landuse IN ('forest', 'military', 'farmland', 'residential', 'commercial', 'retail', 'industrial') THEN landuse ELSE NULL END)) AS landuse,
               ('natural_' || (CASE WHEN "natural" IN ('wood', 'sand', 'scree', 'shingle', 'bare_rock') THEN "natural" ELSE NULL END)) AS "natural",
               ('wetland_' || (CASE WHEN "natural" IN ('wetland', 'mud') THEN (CASE WHEN "natural" IN ('mud') THEN "natural" ELSE tags->'wetland' END) ELSE NULL END)) AS wetland,
+
```
pnorman: stray newline

kocio-pl:
I like the idea.

matthijsmelissen:
I have applied the requested changes.

kocio-pl:
Rendering test of south of the Uzbekistan (as originally reported in #256):

z8
Before
bandpl5x
After
kmgxdcrj

z9
Before
4hfirjvs
After
wgfkhd8e

project.mml
kocio-pl: You forgot to remove this empty line.

matthijsmelissen:
Seems I forgot to push. Should be better now!

kocio-pl:
> I haven't looked at how effective this approach is and the cartographic results.

This change is important in the areas with lacking landuse/natural tagging. In other places the change won't be too big (z8-9 means very big cities and we have them visible now).

The Uzbekistan sample shows that both data are different, but comparable.

matthijsmelissen:
The Uzbekistan area seems fine to me.

Likely there are some places on earth where this will be a degradation (if landuse has not been drawn) but I'd consider it an improvement not to show settlements if we don't have them in OSM.

matthijsmelissen:
All requested changes have been added. @pnorman didn't check the cartographic results, if you're fine with that this can be merged.

kocio-pl:
@pnorman rarely comments here for some time, which makes it harder to discuss changes. But I also believe this is not about cartographic results (they are similar), but rather about data consistency and dropping external dependencies (both of which are obvious), I think we should really merge it.

PR #2970, opened 2017-12-06
Render glaciers and icesheets from z8 instead of z6

matthijsmelissen:
We don't render other landcover at z6/z7, and the glaciers, especially in
the Alps, currently attract too much attention.

### DIFF
--- a/project.mml
+++ b/project.mml
@@ -217,7 +217,7 @@ Layer:
       file: data/antarctica-icesheet-polygons-3857/icesheet_polygons.shp
       type: shape
     properties:
-      minzoom: 4
+      minzoom: 8
   - id: water-areas
     geometry: polygon
     <<: *extents
@@ -277,7 +277,7 @@ Layer:
       file: data/antarctica-icesheet-outlines-3857/icesheet_outlines.shp
       type: shape
     properties:
-      minzoom: 4
+      minzoom: 8
   - id: water-lines
     class: water-lines
     geometry: linestring
--- a/shapefiles.mss
+++ b/shapefiles.mss
@@ -24,13 +24,13 @@
 }
 
 #icesheet-poly {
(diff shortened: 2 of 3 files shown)


matthijsmelissen:
Current rendering of the Alps:
screen shot 2017-12-06 at 23 25 52

kocio-pl:
As I said - it makes sense if we don't plan to render natural features on z5+ soon. Otherwise it would be vanishing now and returning back:

https://github.com/gravitystorm/openstreetmap-carto/issues/2925#issuecomment-343744169

matthijsmelissen:
I'd prefer to make it consistent now, and I'm fine with reintroducing glaciers when/if we start rendering other landuse.

kocio-pl:
OK, I agree and I tested that the code really works.

Tajikistan

z6
Before
tvk8oz9m

After
u9_zm8ze

z7
Before
ruygflde

After
1kuvdmcz

dieterdreist:
sent from a phone

> On 9. Dec 2017, at 14:40, Matthijs Melissen <notifications@github.com> wrote:
> 
> I'd prefer to make it consistent now, and I'm fine with reintroducing glaciers when/if we start rendering other landuse.


what about liquid water, shouldn’t it be removed then as well until z8 for consistency reasons?

kocio-pl:
Which kind of water would you like to filter out? I guess it's a special natural type to show general land/water difference.

imagico:
@dieterdreist - nice try.  Or as i would put it - global warming has reached OSM...

In case anyone is interested in the numbers - the approximate global statistics:

type |  area
------------ | -------------
ocean | ~360M km^2
ice | ~15-16M km^2
inland water | ~4-5M km^2
woodland | ~40M km^2
farmland | ~14M km^2

The last two of course depend highly on the definitions used.

PR #2972, opened 2017-12-06
Add halo to roads on z6 and z7, and make motorways on z6 slightly narrower

matthijsmelissen:
This PR adds a white halo to roads on z6 and z7. It also makes motorways on z6 slightly narrower #2972.

This should give the roads slightly more contrast.

### DIFF
--- a/roads.mss
+++ b/roads.mss
@@ -1051,20 +1051,24 @@ tertiary is rendered from z10 and is not included in osm_planet_roads. */
       //Missing line-cap: round; is intentional. It would cause rendering glow multiple times in some places - what as result of partial transparency would cause differences in rendering
       //Also, bridges - including bridge casings are rendered on top of roads. Enabling line-cap: round would result in glow from bridges rendered on top of road around bridges.
     }
-    [feature = 'highway_motorway'][link != 'yes'][zoom >= 8][zoom < 12],
+    [feature = 'highway_motorway'][link != 'yes'][zoom >= 6][zoom < 12],
     [feature = 'highway_motorway'][link = 'yes'][zoom >= 10][zoom < 12],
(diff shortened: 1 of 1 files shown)


matthijsmelissen:
screen shot 2017-12-07 at 00 19 48
screen shot 2017-12-07 at 00 19 43

matthijsmelissen:
Hard to assess this with limited data, @rrzefox would you be able to deploy this on your server?

rrzefox:
> @rrzefox would you be able to deploy this on your server

It's been deployed and Z6+Z7 rerendered.

matthijsmelissen:
See http://bl.ocks.org/matthijsmelissen/raw/b3bacf30c403644b10597a70046d11e8/#7.00/51.932/6.217. The difference is very subtle, I'm not convinced either way.

kocio-pl:
I'm also not sure which is better.

matthijsmelissen:
I think the problem is that due to anti-aliassing effects, the new roads look quite a bit narrower. I'll try to counteract that effect by making the roads a bit wider.

kocio-pl:
Do you still plan to work on this code?

matthijsmelissen:
After revisiting this issue, I would still like to go through with this one. Although the change is subtle, this really improves legibility - not only of the roads itself, but also of other features like admin boundaries.

Making the roads wider does not improve the rendering to me - in fact, I decided to make the road on z6 slightly narrower.

z6 before/after:
screen shot 2018-04-13 at 01 08 05 
screen shot 2018-04-13 at 01 09 04

z7 before/after:
 screen shot 2018-04-13 at 01 07 29 screen shot 2018-04-13 at 01 07 47

@kocio-pl What do you think?

kocio-pl:
Looks better for me. :+1:

matthijsmelissen:
@kocio-pl Can I merge it?

PR #2974, opened 2017-12-10
Adding rendering for amenity=waste_disposal

### LINKED ISSUE #419: Add rendering for amenity=waste_disposal
Popular tag is not rendered on the map.
 Please add this tag.

http://wiki.openstreetmap.org/wiki/Tag:amenity%3Dwaste_disposal

kocio-pl:
Resolves #419.

As this feature is both popular (54k+ uses) and small scale, I choose the least obtrusive rendering possible. 

The icon is bigger than waste basket, since in reality it is larger (standard 14 px matrix instead of 10 px), but otherwise they are rendered in a similar way: on z19+ with low priority. I also render only nodes to avoid rendering icons on waste disposal sheds. Nodes are much more popular so this means rendering of 95% such objects.

jdtv_9jz
glmdd98c

There might be some extreme cases mentioned in a discussion (https://github.com/gravitystorm/openstreetmap-carto/issues/419#issuecomment-193307736) where there are a lot of such sheds, but tagged as nodes, but this might be solved by retagging as `building:part=*` area (which is more precise). I also added dimmed rendering for access tagging (1367 uses) other than `permissive` and `yes` - the same as we do with parkings. Click to see the full scale:

whqoav2t

### DIFF
--- a/amenity-points.mss
+++ b/amenity-points.mss
@@ -1024,6 +1024,17 @@
     marker-placement: interior;
     marker-clip: false;
   }
+  
+  // waste_disposal tagging on ways - tagging on nodes is defined later 
+  [feature = 'amenity_waste_disposal'][zoom >= 19] {
+    [access = null],
+    [access = 'permissive'],
+    [access = 'yes'] {
+      marker-file: url('symbols/waste_disposal.svg');
+      marker-fill: @amenity-brown;
+      marker-placement: interior;
+    }
+  }
 }
 
 .amenity-low-priority {
@@ -1110,6 +1121,17 @@
     marker-file: url('symbols/waste_basket.10.svg');
     marker-fill: @amenity-brown;
     marker-placement: interior;
+  }
+
+  // waste_disposal tagging on nodes - tagging on ways is defined earlier 
+  [feature = 'amenity_waste_disposal'][zoom >= 19]::amenity {
+    [access = null],
+    [access = 'permissive'],
+    [access = 'yes'] {
+      marker-file: url('symbols/waste_disposal.svg');
+      marker-fill: @amenity-brown;
+      marker-placement: interior;
+    }
   }
 }
 
--- a/project.mml
+++ b/project.mml
@@ -1416,7 +1416,7 @@ Layer:
                                                   'police', 'post_box', 'post_office', 'pub', 'biergarten', 'recycling', 'restaurant', 'food_court',
                                                   'fast_food', 'telephone', 'taxi', 'theatre', 'toilets', 'drinking_water',
                                                   'prison', 'hunting_stand', 'nightclub', 'veterinary', 'social_facility',
-                                                  'charging_station', 'arts_centre', 'ferry_terminal', 'marketplace') THEN amenity ELSE NULL END,
+                                                  'charging_station', 'arts_centre', 'ferry_terminal', 'marketplace', 'waste_disposal') THEN amenity ELSE NULL END,
               'shop' || CASE WHEN shop IN ('no', 'vacant', 'closed', 'disused', 'empty') OR shop IS NULL THEN NULL ELSE '' END,
               'leisure_' || CASE WHEN leisure IN ('water_park', 'playground', 'miniature_golf', 'golf_course', 'picnic_table',
                                                   'fitness_centre', 'fitness_station') THEN leisure ELSE NULL END,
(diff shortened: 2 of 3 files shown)


matthijsmelissen:
The current icon is not very recognizable to me.

kocio-pl:
Another design proposition:
crblx8tz

lakedistrictOSM:
Out of the two icon designs, I prefer the bin bag icon, but it would be better to use a wheelie bin icon if it could be made more recognisable. Perhaps a side view of a wheelie bin such as 
wheeliebin01png

kocio-pl:
> Perhaps a side view of a wheelie bin such as

Do you have ready SVG on PD to play with it?

lakedistrictOSM:
> Do you have ready SVG on PD to play with it?

What's PD?

It's probably easier if you redraw it - I haven't taught myself how to use inkscape properly yet. It was just a quick sketch so will need refining anyway.
wheeliebinbig

kocio-pl:
Public domain or CC0 compatible license. Could you make it available anyway (as a Git XML code)? The devil is in details.

lakedistrictOSM:
Hope this works: https://gist.github.com/lakedistrictOSM/6f1af13a5848a8e6ee78bf8536958430 - the white around the wheel needs to be transparent along with other detail fixes. Feel free to change it as much as you want to.

kocio-pl:
I have recreated this shape and it really looks nice for me:
30saj0h6

dieterdreist:
looks very similar to the glass

kocio-pl:
What about this:
lh1uhshm

lakedistrictOSM:
Version 29 or the one in https://github.com/gravitystorm/openstreetmap-carto/pull/2974#issuecomment-359141722 both look good.

kocio-pl:
Version 33:
b4gwcd w

kocio-pl:
Version 34:
bwy4oxot

kocio-pl:
I like version 29 too and would merge it, but more opinions or suggestions are welcome, of course.

kocio-pl:
> I also render only nodes to avoid rendering icons on waste disposal sheds.

I'm not sure now why should we avoid it. On z19 it's not a problem to show them, I guess, since it's only visible on a small scale.

kocio-pl:
The code is updated now to show also waste sheds (many of them are protected by the key, so the tagging should use `access=private` in such cases):
jcmz0nyi

matkoniecz:
I strongly oppose rendering private amenity=waste_disposal

kocio-pl:
What difference do you see between them and other private objects (like parking, playground etc)?

matkoniecz:
Good question. I admit that my dislike for trending private objects is unexpectedly strong.

But unlike this feature private parkings and private playgrounds are far more noticeable in terrain and are more likely to be an orientation points.

kocio-pl:
I'm able to accept your objections and just not show them, I was just trying to get some better understanding of the matter if possible.

lakedistrictOSM:
What does it look like with dimmed rendering for access=private? IMO private amenity=waste_disposal should be rendered just like other private amenities are - we map what's on the ground, we should show what's on the ground.

dieterdreist:
2018-01-26 13:51 GMT+01:00 kocio-pl <notifications@github.com>:

> What difference do you see between them and other private objects (like
> parking, playground etc)?
>


I see a difference in significance. A private parking is at least
significant, in that it gets one car off the public street for every lot
(and it is big enough that it might serve for orientation), and a private
playground might at least be usable by other people (admittedly, this is
more about ownership than access). Private waster disposal is really
insignificant because everybody in the civilized world has it. I wouldn't
expect it to be mapped, amenity=waste_disposal with access=private seems
misstagging in this context (it is not an amenity), similar to
amenity=telephone, if it is not publicly usable, it has not to be tagged
like this, regardless of access=private or other additional tags. Or
amenity=shower. Etc.

kocio-pl:
> I wouldn't expect it to be mapped, amenity=waste_disposal with access=private seems misstagging in this context (it is not an amenity)

Well, if we see it this way, private parking would be also mistagged by definition, just like any amenity with `access=private`.

kocio-pl:
Something smaller and more subtle (yet bigger than trash bins):
myeulumm

dieterdreist:
2018-01-26 20:12 GMT+01:00 kocio-pl <notifications@github.com>:

> I wouldn't expect it to be mapped, amenity=waste_disposal with
> access=private seems misstagging in this context (it is not an amenity)
>
> Well, if we see it this way, private parking would be also mistagged by
> definition, just like any amenity with access=private.
>


I agree it is a bit of a grey area, but private parkings can become really
big and have a huge influence on the landscape and are neeeded to better
understand the space, and might be useful for visitors, which all aren't
points for private trash cans. I wouldn't map single private parking lots,
but when it's a bigger parking, it is kind of OK for me.

polarbearing:
The wheelie bin works better than the american trash can, since the latter is easily confused with a small litter bin. The bag does not work at all for me since it could anything from money to candy.

I agree with those opposing to render private ones.

kocio-pl:
Wheele is great design (I was not sure if we can make it, but I'm proud of results), but is just too big and too heavy. Trash can is different enough than small bin (engravings medium vs solid small) and I think I will merge it soon.

The only change I plan is to omit rendering private bins by popular demand - we can change it later (probably showing dimmed icons) if we decide so, of course.

kocio-pl:
I have a hard time how to make a CartoCSS expression that would show only cases when the access is null, yes or permissive. If you know how to do it please give me a hint, this would help to finish this issue.

PR #2975, opened 2017-12-10
Move small amenity=recycling bins to z19

### LINKED ISSUE #847: Increase zoomlevel for private amenity=recycling
In the German speaking countries every apartment building has it's own containers for recycling plastic, paper,but only people living in those apartments are allowed to use them. They are mapped according to the description on the wiki page with access=private added to them.

problem is, mapnik still renders them even at zoom level 16, which totaly clutters the map. example:  http://osm.org/go/0JhJfqRx

Suggestions for fine tuning the rendering of this tag: 1) amenity=recyling + access+private should only be rendered at zoom level 18 and higher and transparency of the symbol should be set to 50% (like it is done with amenity=parking if set to private). at zoom level 17 or lower, it shouldn't be rendered at all. 2) amenity=recyling + access+permissive should only be rendered at zoom level 17 and higher and transparency of the symbol should be set to 50% (like it is done with amenity=parking if set to permissive ). at zoom level 16 or lower, it shouldn't be rendered at all. 3) amenity=recyling + access+public (or no restriction) should be rendered without transparency and at zoom level 16. (basically as it is now)

See also https://trac.openstreetmap.org/ticket/3575.

kocio-pl:
Resolves #847 (I guess, because this code is broader).

Only recycling centres (11k+ uses) will stay on z17+, all the other recycling features are supposed to be small or less important - including very popular recycling_type=container (107k+ uses) and all the other non-standard values.

### DIFF
--- a/amenity-points.mss
+++ b/amenity-points.mss
@@ -472,7 +472,8 @@
     marker-clip: false;
   }
 
-  [feature = 'amenity_recycling'][zoom >= 17] {
+  [feature = 'amenity_recycling'][recycling_type = 'centre'][zoom >= 17],
+  [feature = 'amenity_recycling'][zoom >= 19] {
     marker-file: url('symbols/recycling.svg');
     marker-fill: @amenity-brown;
     marker-placement: interior;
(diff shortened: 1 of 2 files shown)


kocio-pl:
Any comments about this?

daganzdaanda:
Good idea!

PR #2978, opened 2017-12-10
Make nature reserves border thinner

### LINKED ISSUE #1950: Remove stroke for boundary=national park at z7-9
Area features at z7-9 are generally not outlined (e.g. water, urban areas, woodland).

`boundary=national park`, however, has a dark green outline. At first glance this appears as a linear feature and causes confusion with other linear features (roads, railways).

[Point of comparison](http://tools.geofabrik.de/mc/#8/54.2288/-2.1395&num=8&mt0=mapnik&mt1=google-map&mt2=bing-map&mt3=mapnik-german&mt4=nokia-map&mt5=mapquest&mt6=cyclemap&mt7=public_transport): Google and HERE do not show an outline for these features. osm.de shows a thinner, fainted, dotted outline.

Removing the outline would make the map more readable at these zoom levels in areas where there is a high number of linear features in national parks.

kocio-pl:
Follow up to #2119.
Resolves #1950 (I guess this is the real problem behind this ticket).

Protected areas like national parks and nature reserves can still be too strong, when they are clustered. The problem is with the borders between them - they are sometimes too strong and dominate everything. Hiding smaller ones makes it less severe, but in some places such "internal" borders are still too visible.

I propose to make lines a bit thinner, mostly on z8-z9 but also on z10-z13. It would make it safer to add rendering for boundary=protected_area (#603), which I plan to do. The chosen values should not make other places, where the borders are not clustered and which are shown on busy backgrounds, to disappear.

Okanogan, z8
Before
wdxk7_pj
After
cytw0y1m

Adirondack, z8
Before
orleviua
After
gqoqy9qr

Wrocław, z8
Before
vx3gkysz
After
yyueusus

Okanogan, z10
Before
3ntezmnf
After
prsai5pw

Wrocław, z10
Before
oc96ze_k
After
y7_dlj3u

### DIFF
--- a/admin.mss
+++ b/admin.mss
@@ -203,7 +203,7 @@ overlapping borders correctly.
     a/line-opacity: 0.15;
     a/line-join: round;
     a/line-cap: round;
-    b/line-width: 2;
+    b/line-width: 1;
     b/line-offset: -1;
     b/line-color: green;
     b/line-opacity: 0.15;
@@ -212,7 +212,7 @@ overlapping borders correctly.
     [zoom >= 10] {
       a/line-width: 2;
       a/line-offset: -1;
-      b/line-width: 4;
+      b/line-width: 2;
       b/line-offset: -2;
     }
     [zoom >= 14] {


kocio-pl:
What do you think about it?

matthijsmelissen:
With the new rendering, it is harder to see what is the inside and what is the outside of the national park.

kocio-pl:
Is it acceptable for you? I guess anything less would not be worth changing, because gains will be not visible enough. But we can also leave it as it is - I can live with too strong lines in some cases, just trying to fix it a bit.

matthijsmelissen:
If we want to go narrower, i think it doesnt make sense to keep the double line.

We can also see if making them more transparent helps.

kocio-pl:
Another take is needed to fix rendering of complex borders, but I'm not ready to test more solutions at the moment, so I'm closing this ticket.

PR #2968, opened 2017-12-04
Create funeral_directors.svg

Jotam:
As discussed here: https://github.com/gravitystorm/openstreetmap-carto/issues/2355#issuecomment-348985238

PR #2979, opened 2017-12-11
New icon for shop=funeral_directors

Jotam:
This is a replacement / correction for PR https://github.com/gravitystorm/openstreetmap-carto/pull/2968


kocio-pl:
Thanks, that's better, but you still need to edit the `project.mml` (where the specific data are selected from database) and it would be good if you test the rendering. BTW: how did you make the image in https://github.com/gravitystorm/openstreetmap-carto/issues/2355#issuecomment-348977004?

You can see an example of the latest merging of shop=tyres (this would be quite similar, as shops are very typical in osm-carto):

https://github.com/gravitystorm/openstreetmap-carto/pull/2706/files

You should also add the line "Resolves #2355" to this ticket, so the issue would be automatically closed on merging this PR.

Jotam:
> you still need to edit the `project.mml`

Sorry guys, I really lack the time to dig deeper into all of this. Could someone do this who already knows the drill and therefore won't waste a lot of time? Thank you!

kocio-pl:
OK, but the discussion about proposed shape is still pending in #2355, so there's no hurry.

kocio-pl:
Thanks for trying, but this code is both incomplete for a long time and the design is not popular enough among community. Therefore I will close it now.

Issue #2087, opened 2016-03-11
Render societal amenity campuses with educational/hospital yellow


polarbearing:
In #2071 it was discussed that more campuses of societal amenities should be rendered with the yellow currently being used for educational and hospital areas. The idea sprung from the proposal to use this colour even for sports centres. However, in the continuum of useful facilities, sports centres are further away from educational that things like social or community centres.

Thus, the colour should be renamed from `@educational_areas_and_hospital` to `@societal_amenities`, and applied to the following areas (when not a building) of _Societal amenities_ (see #1624):
- amenity=kindergarten/school/university/college (as is)
- amenity=hospital (as is)
- amenity=community_centre (discussed in #506)
- amenity=social_facility (fixes #1295)
- amenity=police/fire_station (fixes #848)
- amenity=embassy (mentioned in [#775](https://github.com/gravitystorm/openstreetmap-carto/issues/775#issuecomment-136331438))

Further, the labelling should scale consistently, currently educational scales, hospital doesn't.

**Dependencies:**

Icons should be added to the kindergarten/school/university/college as in #120, to distinguish them better. We do have icons meanwhile for the community_centre and social_facility.

imagico:
> However, in the continuum of useful facilities, sports centres are further away from educational that things like social or community centres.

Note this is a subjective assessment and more important it is very culture specific (both in terms of overall culture of a country and in terms of subculture).

The primary purpose of #2071 is reducing the number of colors, this issue seems to be about adding new features - which is hardly feasible to do with new colors.  Then the question would only be if another existing color would be more appropriate - to which the answer is likely no - both for the majority of the above and for sport_centre.

math1985:
I'm still not very sure what to think of this. On the one side, I think there is a need to indicate all campus outlines, and I realize the colour spectrum is limited. On the other hand, I also got quite attached to having yellow as a recognizable school colour.

I think we should also think well about from which zoom level we would like the yellow to show up.

kocio-pl:
> On the other hand, I also got quite attached to having yellow as a recognizable school colour.

But it's already not enough in itself to recognize it - you need some additional objects inside (like pitches) or lack of icons (like for hospital - on z>=15, earlier you can't be sure). Adding icons for schools will keep it distinct enough IMO.

BTW we were also talking about:
- amenity=fire_station
- landuse=religious

I like this proposition in general.

math1985:
[...]

PR #2981, opened 2017-12-12
Yellow background for community_centre and social_facility

kocio-pl:
Related to #2087.

I think it's safe to add community_centre (#506) and social_facility (#1295) as societal amenities - with a light yellow background.

I would also like to add arts_centre, since this is sometimes similar to community_centre, but I'm not sure if this would be accepted - what do you think about it? We don't have a clean, generic solution for lot of "campuses" and social services, so I prefer to make smaller steps where this seems quite consistent rather than try to solve as much as it's possible.

amenity=social_facility 
Before
i5qq_aje

After
glqi5ovz

amenity=community_centre 
Before
phkevc4e

After
hydv oni

### DIFF
--- a/landcover.mss
+++ b/landcover.mss
@@ -520,7 +520,9 @@
   [feature = 'amenity_university'],
   [feature = 'amenity_college'],
   [feature = 'amenity_school'],
-  [feature = 'amenity_kindergarten'] {
+  [feature = 'amenity_kindergarten'],
+  [feature = 'amenity_community_centre'],
+  [feature = 'amenity_social_facility'] {
     [zoom >= 10] {
       polygon-fill: @residential;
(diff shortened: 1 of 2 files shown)


polarbearing:
Looks fine for me.  Do I understand correctly that we only add the fill, and leave the label as it is?
That means not scaling with area size. Which is probably fine, since these campuses tend to be compact in comparison to large schools/universities. 

And yes let's continue in small steps and not throw in all ideas at once. I have more ideas, also.

kocio-pl:
Yes, we have no idea what to do with scaling when we show the icon. If we knew, we would have had school icon set already (see https://github.com/gravitystorm/openstreetmap-carto/pull/2361#issuecomment-249083347).

PR #2982, opened 2017-12-12
Rendering seasonal waterways as intermittent

### LINKED ISSUE #2095: Render waterways with seasonal=yes the same way as intermittent=yes
Seasonality is a specific sort of intermittency - variation is according to seasons instead of more randomly weather-related.

For now, seasonal waterways appear as solid lines and it is a bit disconcerting.

Treating seasonal=yes the same way as intermittent=yes feels quite correct to me and it would also minimize the work necessary to take this extra tag into account.

kocio-pl:
Resolves #2095.

Tag seasonal has basically two values used - yes and no (~96% of 147k uses), but there are also others, like spring or similar. I think it's safer to treat every value other than "no" as a sign, that there is some seasonality, and this code reflects this choice. 

Of course we can also render just some list of values (with seasonal=yes of course), but it's hard to tell what to include exactly. Originally it was proposed to render just "yes", but this is not binary concept.

### DIFF
--- a/project.mml
+++ b/project.mml
@@ -177,7 +177,11 @@ Layer:
       <<: *osm2pgsql
       table: |-
         (SELECT
-            way, waterway, tags->'intermittent' as intermittent,
+            way, 
+            waterway, 
+            CASE WHEN tags->'intermittent' IN ('yes')
+              OR tags->'seasonal' IN ('yes', 'spring', 'summer', 'autumn', 'winter', 'wet_season', 'dry_season') 
+              THEN 'yes' ELSE 'no' END AS int_intermittent,
             CASE WHEN tunnel IN ('yes', 'culvert') THEN 'yes' ELSE 'no' END AS int_tunnel
           FROM planet_osm_line
           WHERE waterway IN ('stream', 'drain', 'ditch')
@@ -193,7 +197,9 @@ Layer:
         (SELECT
             way,
             waterway,
-            tags->'intermittent' as intermittent
+            CASE WHEN tags->'intermittent' IN ('yes')
+              OR tags->'seasonal' IN ('yes', 'spring', 'summer', 'autumn', 'winter', 'wet_season', 'dry_season') 
+              THEN 'yes' ELSE 'no' END AS int_intermittent
           FROM planet_osm_line
           WHERE waterway = 'river'
         ) AS water_lines_low_zoom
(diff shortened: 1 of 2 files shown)


matkoniecz:
> I think it's safer to treat every value other than "no" as a sign, that there is some seasonality

Is there a good reason to accept catch-all here?

kocio-pl:
For me the reasons are:
- most of the time it's "yes" or "no", so there is no unwanted fragmentation
- it's hard to say which values should we include looking at wiki and usage:

> seasonal=spring/summer/autumn/winter/wet_season/dry_season/...

https://wiki.openstreetmap.org/wiki/Key%3Aseasonal

and 

https://taginfo.openstreetmap.org/keys/seasonal#values

matkoniecz:
https://taginfo.openstreetmap.org/keys/seasonal#values this list has some clearly invalid values, in addition many of them are never used for waterways.

seasonal tag is used around 100k times with waterways - see https://taginfo.openstreetmap.org/keys/seasonal#combinations (105k times to be more exact).

I looked at cases of waterways with seasonal tag and seasonal values other than no/yes/spring/summer/autumn/winter/wet_season/dry_season.

I fixed some obvious cases, I opened some notes http://www.openstreetmap.org/note/1238336 https://www.openstreetmap.org/note/1238316 https://www.openstreetmap.org/note/1238313 https://www.openstreetmap.org/note/1238310

I think that accepting only values yes/spring/summer/autumn/winter/wet_season/dry_season is OK - given that no other valid looking values are used anywhere - see http://overpass-turbo.eu/s/tO3 (excludes documented, excludes some clearly invalid or unclear and undocumented)

kocio-pl:
OK, let's do it this way. One can always use intermittent=yes if there are some not typical or not clear values.

The code is updated.

HolgerJeromin:
Could both values be merge in sql?
This could be faster than handling both the same in mss

kocio-pl:
Sure, it makes sense for me too, so I updated the code.

kocio-pl:
I find this solution complete, any comments or technical reviews to further improve it or merge?

HolgerJeromin:
Generated values from sql query are often marked with a prefix (int_tunnel, int_wetland for example).

kocio-pl:
I'm not sure what `int_` prefix really means and what is the purpose  - "integrated" maybe?

HolgerJeromin:
In my opinion this means "internal", so you are aware (while editing the mss) that this is no 1:1 mapping of osm data.

kocio-pl:
Makes sense for me, updated.

kocio-pl:
@matkoniecz What about now? Is it ready to be merged for you?

kocio-pl:
@matkoniecz ping?

matkoniecz:
Sorry, but I looked through entire thing and noticed something - see https://github.com/gravitystorm/openstreetmap-carto/issues/2095#issuecomment-356648142

Sorry for not spotting it before complaining about technical details.

kocio-pl:
No problem, it's always better to find problems before merging. :smile:

matkoniecz:
I am withdrawing my objection as it is not shared by others, still I am personally against this change (but apparently it is my personal opinion, so I will just not spend time on reviewing and merging this PR and leave it to others)

kocio-pl:
Thanks for investigating the issue! Since this is clear now for me, I will merge it, but it can be reverted/changed in the future, if needed.

PR #2992, opened 2017-12-21
Adding icon for leisure=firepit

### LINKED ISSUE #2910: Add icon for leisure=firepit
W have 8,5k  of leisure=firepit, but is not rendered. 
https://taginfo.openstreetmap.org/tags/leisure=firepit

Examples:
http://www.openstreetmap.org/way/525604639
https://www.openstreetmap.org/way/527840126
https://www.openstreetmap.org/node/5128204779

I think firepits may be important information for tourists, and should be rendered.

My icon proposal:
leisure firepit2
14x14:
leisure firepit

kocio-pl:
Resolves #2910.

[Example place](http://www.openstreetmap.org/#map=19/52.39875/16.98080):

Before
u_3n1llm

After
lwilaeya


turnsole80:
Yes. A very good idea. Well done.

kocio-pl:
Any reviews or other comments here?

sommerluk:
Looks good to me in general.

Maybe the flame could be a little smaller, while preserving the size of the crossed wood.

Maybe switch to z18? Would be yet high for rural areas, but there are also firepits in residential areas, which might be dense at z17…

kocio-pl:
I don't think there are any places with fire pit which could be really a problem on this zoom level. I try to offload z17, but only when the objects are used in the urban areas and this type is unlikely to be seen there. Example location is within big city boundaries, but still this is a large green area, so no clash with shops, monuments or anything like that.

sommerluk:
What about this icon?


[firepit3.zip](https://github.com/gravitystorm/openstreetmap-carto/files/1593502/firepit3.zip)

kocio-pl:
Looks even nicer for me, I will make a code update probably. 

BTW: you could just use gist, so it would be easier to preview.

kocio-pl:
Code is updated now and this is a new rendering:

oanvotsl

kocio-pl:
I don't think they have a name (Taginfo shows no common tags used at he same time), and even if they do, I suspect it's rather the name of a picnic site. They are also so small, that they don't need the area rendering even if they are not technically nodes.

PR #2993, opened 2017-12-22
Move emergency=phone to higher zoom level

kocio-pl:
Related to #1884.

Emergency phones are really small scale, special features and I'm not even sure if they belong to general map - probably not, but even if they do, they should be rendered later, because proper tagging results in z17 clutter:

[Example place on z17](http://www.openstreetmap.org/#map=17/41.30962/-72.92796):

Before
feoc_qbo
After
g6tcvoqd

### DIFF
--- a/amenity-points.mss
+++ b/amenity-points.mss
@@ -504,7 +504,7 @@
     marker-clip: false;
   }
 
-  [feature = 'emergency_phone'][zoom >= 17] {
+  [feature = 'emergency_phone'][zoom >= 19] {
     marker-file: url('symbols/emergency_phone.svg');
     marker-fill: @amenity-brown;
     marker-placement: interior;


SomeoneElseOSM:
Does this density usage of emergency=phone appear elsewhere?  It seems pretty unusual.  Certainly in the UK a typical emergency phone wold be beside a motorway a long distance from the next one.  In another map style I've rendered emergency phones from z17 for a while (see e.g. https://map.atownsend.org.uk/maps/map/map.html#zoom=17&lat=52.870968&lon=-1.39329 ) and I've never thought "there are too many of them".

Is this perhaps another "Polish graveyard" - making one standard map work well in one specific location means that it works less well everywhere else on the planet?

kocio-pl:
Let's start with checking the reasons to show emergency phone on the general map at all, if this is related mostly to motorways? 
- This is not a car map, so car related features should not be elevated.
- In journey one uses rather GPS mobile devices and apps with vector data or paper maps (both of which might have such points visible), not the standard raster map.
- One does not preview areas for emergency devices, their location should be signed and highlighted on the ground.

The main thing that is pro rendering is showing diversity of objects in OSM (database preview). That's why I propose z19, not just to stop rendering them, but even on this zoom level they are still a problem in this campus IMO. Another pro is that they are shown for a long time (although the rendering changed and I provided the code which shows the new tagging scheme lately - #2809).

> Is this perhaps another "Polish graveyard" - making one standard map work well in one specific location means that it works less well everywhere else on the planet?

What was the story about "Polish graveyard"? I don't remember the issue.

I'm not sure how many such places might be, unfortunately I have no tools to find "dense cluster of x". This is however the illustration of the problem when somebody is motivated enough to map them carefully.

I guess most of the emergency features (such as phones, fire extinguishers, hydrants, sirens, defibrillators or assembly points) do not suit here well, at least not until very late zoom levels, when you may want to see detailed plans of some places. It's hard for me to find exceptions - ambulance station is rather big, but that can be shown simply as a building.

dieterdreist:
2017-12-22 10:27 GMT+01:00 SomeoneElseOSM <notifications@github.com>:

> Does this density usage of emergency=phone appear elsewhere? It seems
> pretty unusual. Certainly in the UK a typical emergency phone wold be
> beside a motorway a long distance from the next one.
>


there are different kind of emergency phones:
- along motorways and some primary roads
- in urban areas (e.g. in parks)
- standard telephones with an emergency button / lever (these were common
in Germany in the eighties, see
https://img.webme.com/pic/t/telefonzelle/nrm_front_1_freigest.jpg )

according to the current wiki definition, emergency=phone applies to all of
them:
https://wiki.openstreetmap.org/wiki/Tag:emergency%3Dphone

matkoniecz:
> What was the story about "Polish graveyard"? I don't remember the issue.

Maybe it is about mapping of all alleys and footways in cemetery, rather than only cemetery area that encourages rendering changes that result in worse maps in areas where footways are rare and tend to be important (mountains, rural areas)?

turnsole80:
Yes, I think emergency phones should still be rendered, since they cover all types, and not just those used on highways. 

That said, changing them to appear at a higher zoom seems like a perfectly fine idea.

dieterdreist:
sent from a phone

> On 23. Dec 2017, at 07:07, Chris <notifications@github.com> wrote:
> 
> Yes, I think emergency phones should still be rendered, since they cover all types, and not just those used on highways.



being able to distinguish the different types automatically would be nice (subtag), e.g.
phone_context=facility / highway / urban  / accessory etc.

cheers,
Martin

kocio-pl:
It is always good to have some classification of objects, however good classification is not always easy to design. For example "war memorial" and "stone" values share the same `memorial=*` sub-key, which is unfortunate mix of the common forms and the common functions.

For rendering the most important property seems to be urban/rural location, but it's rather contextual problem than category problem and we need to detect it, probably using some form of pre-computing (see #1957 and https://github.com/SK53/ua2/ ). When the highway goes through the city, the `emergency=phone + phone_context=highway` should be probably rendered later then the same phone when going through the rural area. 

Real life example: [this motorway junction label](http://www.openstreetmap.org/node/31156707#map=12/52.2364/20.9793) competes with a city district name and some other junction labels, which is bad, while [outside the city](http://www.openstreetmap.org/node/983418365#map=12/51.8761/19.6949) the same zoom level is OK.

dieterdreist:
sent from a phone

> On 23. Dec 2017, at 14:44, kocio-pl <notifications@github.com> wrote:
> 
> When the highway goes through the city, the emergency=phone + phone_context=highway should be probably rendered later then the same phone when going through the rural area.


first I apologize, thought I would be writing to the tagging ml (here it is less appropriate, but as you answered I’ll clarify this point you mention)

In this idea of phone_context you would tag urban context in a city (unless it’s a motorroad or other road with access restrictions for pedestrians) also for an emergency phone that is located near a road (e.g. those in the rendering examples above). highway context means only for highway users, while a normal urban road means typically pedestrians at night would use it

Anyway, emergency phones have lost most of their importance with the diffusion of cellphones

kocio-pl:
Yes, this is partially a tagging problem - the mapper should not be pushed to choose highway/urban context if both are true, that's exactly what I said about how hard it is to design proper subtags.

But it's a more general problem anyway that almost any kind of object outside the urban context should be rendered earlier, yet no one expects people to tag "urban=yes/no" for junctions, fuel stations, restaurants, motels and other highway related POIs (but also crosses, shrines, shops, towers and so on).

dieterdreist:
sent from a phone

> On 24. Dec 2017, at 10:45, kocio-pl <notifications@github.com> wrote:
> 
> the mapper should not be pushed to choose highway/urban context if both are true, that's exactly what I said about how hard it is to design proper subtags.


what I meant: highway context implies an emergency phone primarily or solely installed for people having problems on the highway. This isn’t true at all in an urban setting (unless it’s a motorway), hence the value doesn’t apply.

Ciao, Martin

Issue #2985, opened 2017-12-18
area rendering for civic/governmental/administrative services


polarbearing:
We should discuss how to render civic/civil/governmental/administrative/enforcement services, such as

- amenity=police/fire_station (fixes #848)
- amenity=embassy (mentioned in [#775](https://github.com/gravitystorm/openstreetmap-carto/issues/775#issuecomment-136331438))
- landuse for civic/governmental/administrative purposes, which is [proposed but undecided](https://wiki.openstreetmap.org/wiki/Proposed_features/Civic_admin) in tagging. 

In the discussion in #1624 some of them were included in "Societal amenities", however we are now using the societal yellow for kindergarten/school/university/college/hospital/community_centre/social_facility, which I consider a good grouping that should not be overloaden with more meanings. All these yellow areas are facilities that have in common to provide somehow direct care and service for their users.  kocio-pl's arts_centres would still fit in this group. Thereby I close #2087.

Thus we might think of a different colour group for the facilities listed in this issue. 
It might reuse commercial colour or be something new.

kocio-pl:
I would say this is still rather tough issue and we need some new hints and ideas.

Making arts_centre PR will be rather easy, so we could try it soon.

I guess we could resurrect #2519 - it's still similar both to hospital and clinic for me, and if they are societal, this is not less societal than them.

kocio-pl:
> amenity=police/fire_station (fixes #848)

What about light red background as in military area, but without any hatching? See also #2670 - we could render general [uniformed services](https://en.wikipedia.org/wiki/Uniformed_services) the same and leave hatching only for restricted areas of this type.

kocio-pl:
I guess `landuse=commercial` is currently poorly defined and already general enough ("Such area may consists of offices, administration, laboratories, car repair stations, logisticks park, etc."), so this color should be used for any type of offices, like embassy/civic/governmental/administrative.

We have also some cultural institutions, like big museums (including open-air museums), theaters and so on, which are not about the offices. I guess we could experiment with some light brown (in general they exist in the different space context than farmlands, so it shouldn't be confusing).

Are there any other general categories which would need different background? Light blue might be used for some of them, as it's currently not used.

polarbearing:
Yes #1624 was trying to categorise: touristic destinations (e.g. museum areas), accommodation (currently blue icons). We were also thinking about more colour outlines in some discussions. 

The military red for police/fire_station would work for me. 

[...]

PR #2994, opened 2017-12-23
Yellow background for amenity=arts_centre

kocio-pl:
Follow up to #2981, related to #2985.

With recent changes with leisure backgrounds, `societal amenities` are better defined and I think amenity=arts_centre is a good candidate to be included, since it is similar to amenity=community_centre.

[Example location](http://www.openstreetmap.org/#map=19/52.16103/21.21327):

Before
r6nibarp

After
ffefdz38


polarbearing:
Sounds reasonable. Well-distinguished with the icon.

PR #2995, opened 2017-12-24
Move natural=heath earlier

kocio-pl:
Showing natural=heath on z8+ (instead of z10+) gives England nature reserves a context:

z8
Before
sacmenr9

After
xxmcorby

z9 (click to see full images)
Before
ustumfwg

After
a na nxy

### DIFF
--- a/landcover.mss
+++ b/landcover.mss
@@ -478,7 +478,7 @@
     [way_pixels >= 64] { polygon-gamma: 0.3;  }
   }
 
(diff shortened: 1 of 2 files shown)


kocio-pl:
Any reviews or comments?

matkoniecz:
Good idea!

Sorry for a long wait.

kocio-pl:
Another big heath area that is now visible (east Nevada):
https://www.openstreetmap.org/#map=8/38.835/-115.085

PR #2997, opened 2017-12-28
[WIP] OpenType feature “zero”

### LINKED ISSUE #2647: Use OpenType features to get better legibility
There are some characters that might easily be confounded. 1lI are such characters (U+0031 DIGIT ONE, U+006C LATIN SMALL LETTER L, U+0049 LATIN CAPITAL LETTER I), and O0o are (U+004F LATIN CAPITAL LETTER O, U+0030 DIGIT ZERO, U+006F LATIN SMALL LETTER O). One of the design goals of this map style is to give mappers feedback for their work. So it is good if similiar letters can easily be distinguished.

The current release version of Noto does not give us any options here. But the current alpha version of Noto has the “zero” OpenType feature, that gives a different shape of U+0030 DIGIT ZERO to distinguish it from U+004F LATIN CAPITAL LETTER O. Usually the “zero” OpenType feature is implemented by a slash or a dot in the inner of the U+0030 DIGIT ZERO glyph. See https://github.com/googlei18n/noto-fonts/files/685675/en-733-slashed-zero-NotoSans-Regular.pdf for a rendering sample of Noto’s alpha version. Once released, it might be interesing for us to use it…

sommerluk:
Resolves #2647 

It would look like this:

Before:
before

After:
after

At very small text sizes, it looks a little bit like an “8”. Gets better when font size rises.

(In the image, the road name is at font size 9 pt and looks not so great, the junction name is at font size 10 pt and looks better. Road names are currently the last part in this style that uses font sizes < 10 pt.)

The current code only applies this to junction names and some road names, but we would have to add it to all other texts because we likely want to provide a consistent rendering. (Maybe the most interesting usecase are “ref” values of “highway” objects, because these are typically not hole words, to it’s more difficult to differentiate “0” and “O” just by guessing.)

Questions:

- Do we want this?
- If yes – I suppose there is no other way to set text-font-features globally instead of applying it individually like we are doing with text-face-name, is there?

### DIFF
--- a/fonts.mss
+++ b/fonts.mss
@@ -175,3 +175,9 @@ Italics are only available for the base font, not the other scripts.
 For a considerable number of labels this style will make no difference to the regular style.
 */
 @oblique-fonts: "Noto Sans Italic", @book-fonts;
+
+/*
+Tailoring OpenType features:
+- “zero” (Slashed zero): enabled to make O and 0 easy to distinguish
+*/
+@opentype-features: "zero 1";
--- a/roads.mss
+++ b/roads.mss
@@ -2420,6 +2420,7 @@ tertiary is rendered from z10 and is not included in osm_planet_roads. */
       text-size: 10;
       text-fill: black;
       text-face-name: @book-fonts;
+      text-font-feature-settings: @opentype-features;
       text-halo-radius: @standard-halo-radius;
       text-halo-fill: @standard-halo-fill;
       text-wrap-width: 30;  // 3.0 em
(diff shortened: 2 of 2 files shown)


matkoniecz:
From looking at this single example - I think that current rendering looks better.

Is current rendering prettier also according to others?

dieterdreist:
sent from a phone

> On 28. Dec 2017, at 19:18, Mateusz Konieczny <notifications@github.com> wrote:
> 
> Is current rendering prettier also according to others?



+1, the strike-through zero is ugly

kocio-pl:
I also don't like "zero" feature - it could be useful if numbers and names were mixed, but even "Rue O130" is not too hard to get right for me, since we have a lot of roads like "E71" or "A3". Other OpenType features might be interesting though.

sommerluk:
Closing because this does not seem to improve legibility.

aceman444:
It is easy to see if you have 0 and O in the same spot. But are there some roads 013 and O13 ?

matthijsmelissen:
Closest I could find was Turkey, which has both 'O-20' and 'D 020'.

PR #2998, opened 2017-12-29
Fine-tuning of bridge labels

sommerluk:
Fine-tuning of bridge labels

Now the names of bigger bridges are preferred over the names of smaller bridges. The deprecated `text-min-distance` has replaces by a (slightly bigger) text-margin to avoid confusion with labels that are too near from each other. And margin and halo now scale smoothly with increasing text size.

Before:
screenshot 1

After:
screenshot 2

Before:
screenshot 3

After:
screenshot 4

Before:
screenshot 5

After:
screenshot 6

### DIFF
--- a/project.mml
+++ b/project.mml
@@ -1680,6 +1680,7 @@ Layer:
             name
           FROM planet_osm_polygon
           WHERE man_made = 'bridge'
+          ORDER BY way_area DESC
         ) AS bridge_text
     properties:
       minzoom: 11
--- a/roads.mss
+++ b/roads.mss
@@ -2449,30 +2449,36 @@ tertiary is rendered from z10 and is not included in osm_planet_roads. */
       text-face-name: @book-fonts;
       text-halo-radius: @standard-halo-radius;
       text-halo-fill: @standard-halo-fill;
-      text-min-distance: 2;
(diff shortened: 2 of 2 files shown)


kocio-pl:
My initial idea was to make railway and pedestrian bridges less visible (https://github.com/gravitystorm/openstreetmap-carto/issues/2984#issuecomment-353542422), but just sorting by the size makes sense and it works without any changes in tagging. 

In this case it seems that general bridge is just bigger than train bridge, so z15-z16 are now better. However on z14 train bridge label is still rendered instead of general bridge (Most Gdański is a car/tram/pedestrian/bicycle bridge) - is it fixable or do you know the reason?

sommerluk:
> However on z14 train bridge label is still rendered instead of general bridge (Most Gdański is a car/tram/pedestrian/bicycle bridge) - is it fixable or do you know the reason?

Hm, good question. My guess is this is because the river label takes precedence of bridge labels, and at z14 the river label avoids dispaying “Most Gdański”. And than, without “Most Gdański”, there is enough room for the label of the smaller bridge.

kocio-pl:
OK, that sounds reasonable and I find this code to be usable tuning.

PR #3001, opened 2017-12-29
Move highway=traffic_signals later

kocio-pl:
Related to #1745.

Traffic signals are currently rendered much too early. They are not the destination (like shops or playgrounds), they are not landmarks (like towers), they are just a part of a road infrastructure. 

Roads have some important properties and features - class, name, direction, reference number, access, surface but also bus stops. Traffic lights are not as important to see on the map as any of them, but can still be shown on the highest level, just like other small things that are mapped in OSM.

[Extreme example](https://www.openstreetmap.org/#map=18/52.22984/21.01180) when even z18 is too early visually:
Before 
qoixk0g4
After
ghlurapf

### DIFF
--- a/amenity-points.mss
+++ b/amenity-points.mss
@@ -115,7 +115,7 @@
     marker-clip: false;
   }
 
-  [feature = 'highway_traffic_signals'][zoom >= 17] {
+  [feature = 'highway_traffic_signals'][zoom >= 19] {
     marker-file: url('symbols/traffic_light.16.svg');
     marker-fill: #0a0a0a;
     marker-placement: interior;


matkoniecz:
AFAIK in some countries named traffic lights are used for orientation ( https://wiki.openstreetmap.org/wiki/Tag:highway%3Dtraffic_signals mentions Japan).

Maybe named ones should be rendered from z17?

kocio-pl:
Sure, that sounds like a proper solution for me.

HolgerJeromin:
Ref #813

matthijsmelissen:
I disagree, I think they are currently rendered too late. They are landmarks, in the sense that people often tell 'turn left at the traffic lights'.

kocio-pl:
Hm, I would say that they are landmarks in very specific sense then. What you tell now is more about car routing than about finding your position in the space, and I don't think people use this map for routing by hand. Do they?

matthijsmelissen:
>  I don't think people use this map for routing by hand

I do, and I think many others do too.

kocio-pl:
OK, I'm closing it then.

PR #3005, opened 2018-01-06
toll booth symbol

### LINKED ISSUE #958: Add rendering for toll_booth
I am on mobile so i am not 100% sure it is missing (or there is an issue about it) but rendering barrier=toll_booth would be nice.

Osmand renders it same as a lift gate plus an $

sommerluk:
Adds a toll booth symbol. Based on the icon of @MaestroGlanz, but without the right bollard (because at 14×14, with the bollard it looks to me a little bit like a dripping oil can). Some small geometry adjustments.

Resolves #958 

To discuss:
- Zoom level: 16?
- barrier=toll_booth on areas? JOSM and iD have a preset for both, points and areas. The wiki allows it on points and areas (right box on the wiki page), but at the same time, the wiki description says _“The node should be on the highway the toll applies to, in order to support routing decisions.”_ and doesn’t describe a correct mapping on areas. Actually about 5% of the toll booth objects in the database are areas – most of them have also a `building=*` tag.
- Between ¼ and ⅓ of the toll booth objects have a name. Should this be rendered?

screenshot 1
screenshot 3
screenshot 4

### DIFF
--- a/amenity-points.mss
+++ b/amenity-points.mss
@@ -122,6 +122,13 @@
     marker-clip: false;
   }
 
+  [feature = 'barrier_toll_booth'][zoom >= 16] {
+    marker-file: url('symbols/toll_booth.svg');
+    marker-fill: @transportation-icon;
+    marker-placement: interior;
+    marker-clip: false;
+  }
+
   [feature = 'amenity_cafe'][zoom >= 17] {
     marker-file: url('symbols/cafe.svg');
     marker-fill: @gastronomy;
@@ -1209,6 +1216,7 @@
 
   [feature = 'amenity_car_rental'][zoom >= 17],
   [feature = 'amenity_bicycle_rental'][zoom >= 17],
+  [feature = 'barrier_toll_booth'][zoom >= 17],
   [feature = 'leisure_slipway'][zoom >= 17] {
     text-name: "[name]";
     text-size: @standard-font-size;
@@ -1217,6 +1225,7 @@
     text-fill: @transportation-text;
     [feature = 'amenity_car_rental']     { text-dy: 10; }
     [feature = 'amenity_bicycle_rental'] { text-dy: 10; }
+    [feature = 'barrier_toll_booth']     { text-dy: 13; }
     [feature = 'leisure_slipway']        { text-dy: 13; }
     text-face-name: @standard-font;
     text-halo-radius: @standard-halo-radius;
--- a/project.mml
+++ b/project.mml
@@ -1542,6 +1542,7 @@ Layer:
                                 WHEN tags @> 'ford=>yes' OR tags @> 'ford=>stepping_stones' THEN 'ford' ELSE NULL END,
               'power_' || CASE WHEN power IN ('generator') THEN power ELSE NULL END,
               'tourism_' || CASE WHEN tourism IN ('viewpoint') THEN tourism ELSE NULL END,
+              'barrier_' || CASE WHEN barrier IN ('toll_booth') THEN barrier ELSE NULL END,
               'man_made_' || CASE WHEN man_made IN ('cross') THEN man_made ELSE NULL END,
               'historic_' || CASE WHEN historic IN ('wayside_cross') THEN historic ELSE NULL END
             ) AS feature,
@@ -1580,6 +1581,7 @@ Layer:
             OR shop IS NOT NULL
             OR leisure IN ('water_park', 'playground', 'miniature_golf', 'golf_course', 'picnic_table', 'slipway',
                            'dog_park', 'fitness_centre', 'fitness_station', 'firepit')
+            OR barrier IN ('toll_booth')
             OR man_made IN ('mast', 'water_tower', 'lighthouse', 'windmill', 'cross', 'obelisk')
             OR "natural" IN ('peak', 'volcano', 'saddle', 'spring', 'cave_entrance')
             OR historic IN ('memorial', 'monument', 'archaeological_site', 'wayside_cross')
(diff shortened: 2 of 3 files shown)


kocio-pl:
1. I think z16+ works good.
2. I think rendering areas would be OK. In the future we might develop similar scheme to public transport platform/stop position, but currently we lack this precision so rendering would pretend to be precise while it's not. 
3. It would be good to take a look at these names, because I have no idea what it could be. If this is mostly the road ref, then it would be just misused, but if that is not the case, I would render it in general.

BTW: This PR lacks the icon itself.

matkoniecz:
ping @sommerluk 

> this PR lacks the icon itself.

matkoniecz:
ping @sommerluk 

> this PR lacks the icon itself.

sommerluk:
Sorry, I’m quite busy currently, but I hope to update this PR with

- the missing icon
- area rendering
- (maybe) ref rendering

soon.

sommerluk:
Finally, the missing icon is added to this PR.

But area rendering does not work. I had added a query to `id: amenity-low-priority-poly` in the hope that this would be enough – without making any additional changes to the `.mms` file because there anyway we use the class `amenity-low-priority` which works for both, point and poly. But – it does not work! I think this is because we import barriers as line features, not as polygon features. (By the way: That means that all the other barrier queries within `id: amenity-low-priority-poly` are also useless.) I’ve tried to place the icon with the line geometry (instead of polygon geometry), but that does not seem to be possible.

After taking a look to the `name` values of toll booth, I think it is reasonable to render them also. Only very few of them are generic (like “Toll booth”) – most of them are real names that are useful.

It might be better to not put this into amenity-low-priority, but into a layer that renders with higher priority than the road names and one-way arrows (otherwise the icon often stays hidden)…

kocio-pl:
So for me it's not a low priority POI, do you plan to change the code to select it as a normal amenity?

sommerluk:
Done.

Point-only – no polygon rendering because of the reason explained above.

I still have to add the label…

kocio-pl:
Area rendering should be perfectly possible - see this example:
- object: https://www.openstreetmap.org/way/253094249
- my test code branch: https://github.com/gravitystorm/openstreetmap-carto/compare/master...kocio-pl:toll
- rendering:
598ukl8l

kocio-pl:
Using you code data inspector in Kosmtik selects it from database:

```
feature barrier_toll_booth
layer area-barriers
```

so it should be available for styling. 

With low priority layers maybe just `::barrier` element was needed:

https://github.com/kocio-pl/openstreetmap-carto/blob/76f93dd1ab49e599815462f3a1f6744c8403d182/amenity-points.mss#L1047

but maybe also putting it in the `.amenity-low-priority` section, not `.points` for example, as you did here. 

Because you wanted to not use low priority some other changes might be needed, but it looks like doable.

sommerluk:
Well, I suppose your example works because this particular “barrier=toll_booth” object has also a “building” tag. With my old code, it was working also fine if a “building” tag was present, but it didn’t render on areas without the “building” tag (or something other that makes them considered as area).

How does your code behave on an area that has “barrier=toll_booth” as the _only_ tag?

dieterdreist:
sent from a phone

> On 5. Feb 2018, at 07:59, Lukas Sommer <notifications@github.com> wrote:
> 
> Well, I suppose your example works because this particular “barrier=toll_booth” object has also a “building” tag.


is there really „toll“ on this footway? Maybe that’s a ticket office and not a toll booth?

kocio-pl:
> is there really „toll“ on this footway? Maybe that’s a ticket office and not a toll booth?

I don't know this place, I used it simply as a test for this code.

> How does your code behave on an area that has “barrier=toll_booth” as the only tag?

Do you have an example? Overpass exceed the limits when I try to find it.

But I think building tag has nothing to do with it. If it was a problem with database, Kosmtik wouldn't detect this object at all. Therefore I think this is a problem with styling only.

kocio-pl:
I made a test with prepared data (just removed building and name tags from *.osm file export) and now it's not shown by data inspector with any of our code branches. So it looks that having a building tag makes it selectable somehow. But I guess this would be still good in many cases, as the area might be typically just a building.

dieterdreist:
2018-02-05 8:12 GMT+01:00 kocio-pl <notifications@github.com>:

> is there really „toll“ on this footway? Maybe that’s a ticket office and
> not a toll booth?
>
> I don't know this place, I used it simply as a test for this code.
>

I mentioned this because actual toll booths usually come i groups, see e.g.
the foto from the osm wiki:
https://wiki.openstreetmap.org/wiki/File:Toll_booth.jpg
While currently there are usually only 1 toll booth node per direction, in
reality there are many and will sooner or later also be mapped I guess.

Here is an example for place in OSM where you can find nodes and a way at
the same spot, something I have seen quite a few times in the data:
https://www.openstreetmap.org/node/2518925312
https://www.openstreetmap.org/way/476398054

kocio-pl:
Yes, I'm aware of this. My position is that since tagging scheme is not precise, we don't have to decide, just like with ferry terminal (which can mean a stop position, small port, big port etc.). So for now it's enough to render as many cases as possible and let mappers decide.

sommerluk:
Okay. Sorry for long waiting time, I’m currently really busy…

PR updated. From my side, it’s ready.

No area rendering for the moment. This is because I do not like the idea that the rendering depends of another, not directly related tag (here: building=*) to make it recognize as area. I think it would be confusing for mappers if sometimes, they see an icon for barrier=toll_booth on areas (with building=*) and sometimes not (without building=*). Area rendering would probably need a change in the lua file (and therefore a database reload).

The “name” tag is rendered now.

kocio-pl:
I suspected that's the case and I was shy to ask. :smile: 

Thanks for your work! We can rethink areas rendering later, but for now both nodes and labels are rendered, which is a good start.

PR #3010, opened 2018-01-09
Improve installation docs

matkoniecz:


kocio-pl:
Looks like CartoCSS won't be needed if we resolve #3012.

SomeoneElseOSM:
> Looks like CartoCSS won't be needed for deployment if we resolve #3012.

While distributing the compiled XML would be useful to many people (myself included) given the dog's breakfast that is node.js, it should still be a goal to explain to people how to actually produce it, and be the default in any installation instructions.  

If you don't you'll get lots of people trying to modify the resultant .xml file (and lots of swearing).

Also don't forget that the project.mml file hardcodes the database name as "gis".  This works for many people but it's also pretty common to need to change it too.

Finally if someone needs to understand "why feature X is displayed as it is" it's useful to have the source around to look at.

kocio-pl:
Thanks, probably this code can be merged now and solving #3012 should bring further changes to the INSTALL.md.

matkoniecz:
> probably this code can be merged now and solving #3012 should bring further changes to the INSTALL.md.

Yes, I prefer to improve documentation of a current state (especially as I am now in situation that I no longer remember some things so it is much easier to spot what is missing/confusing)

Issue #2884, opened 2017-10-08
Importing data info


kocio-pl:
I've just learned about useful data site, OSMaxx, which allows downloading PBF files of arbitrary area with fresh data (~2h). I wanted to add info to INSTALL.md and DOCKER.md, where we have:

> It's probably easiest to grab an PBF of OSM data from Mapzen or Geofabrik. 

but I'm not sure how to do it - for example:

1. It's probably easiest to grab an PBF of OSM data from Mapzen, Geofabrik or OSMaxx. 

2. It's probably easiest to grab an PBF of OSM data from Mapzen (extracts of some popular data), Geofabrik (extracts of regions) or OSMaxx (extracts on-demand of arbitrary area).

3. It's probably easiest to grab an PBF of OSM data (see https://wiki.openstreetmap.org/wiki/Planet.osm#Country_and_area_extracts).

4. It's probably easiest to grab an PBF of OSM data:
- Mapzen (extracts of some popular data)
- Geofabrik (extracts of regions) 
- OSMaxx (extracts on-demand of arbitrary area)
- Planet (full planet, very large)

pnorman:
I'd rather keep it as is. Geofabrik and Mapzen are the two most widely used extract sources, and their individual oddities are reasonably well understood.

kocio-pl:
I think we should rethink which sources we want to mention, because with Mapzen closing soon we're left with Geofabrik, which is about big areas only. I would use 2.:

It's probably easiest to grab an PBF of OSM data from Geofabrik (ready extracts of regions), OSMaxx or HOT Export Tool (on-demand extracts of arbitrary area).

It's also possible to mention direct extracts from OSM website (it may be useful for fast checking of small areas) and planet mirrors.

matkoniecz:
What you consider as big areas? For example http://download.geofabrik.de/europe/poland.html extracts are working fine on really old computers.

matkoniecz:
Though adding alternative is good anyway for case of problems with Geofabrik (both temporary and permanent).

kocio-pl:
For me it's a big difference between ~10 minutes for Warsaw area and ~1,5 h for the whole Poland - and my computer is quite new. I only update Poland once in a while (like weeks), while importing Warsaw was a no brainer if I wanted to test on some fresh data.

Which sources would you mention and which form would you use?

matkoniecz:
[...]

PR #3015, opened 2018-01-10
mapzen shuts down, link will be dead soon

matkoniecz:


kocio-pl:
Since Geofabrik data tend to be rather big (countries and other large administrative areas), maybe we should rethink #2884. I mean we could mention OSMaxx and HOT Export Tool, maybe also exporting directly from website and planet mirrors.

matkoniecz:
This probably should be discussed in a separate issue (as it is not changing at all whatever dead links should be dropped).

PR #3016, opened 2018-01-10
make quick start more standalone

matkoniecz:
currently going through quick start requires looking at full detailed docs what defeats its purpose


matkoniecz:
Argh. I see that it got somehow merged by mistake in https://github.com/gravitystorm/openstreetmap-carto/pull/3015/files

If change proposed here is desirable, just close this PR.

If change proposed here is not desirable, tell me and I will prepare revert.

kocio-pl:
This is a good change for me. I support making quick start cheat sheet really self-contained.

matkoniecz:
If any maintainer feels that it should be reverted, please post within week or two and I will revert and send it through proper review.

PR #3021, opened 2018-01-13
Docker.md improvements, requirements and troubleshooting

polarbearing:
update docker.md with requirements and troubleshooting from last night's installation experience.
Trying to cover different host OS, while experience is from MacOS.

### DIFF
--- a/DOCKER.md
+++ b/DOCKER.md
@@ -1,6 +1,6 @@
 # Running OpenStreetMap Carto with Docker
 
-[Docker](https://docker.com) is an environment that allows you to create a _virtual machine_, in which you can run software without altering your host system permanently. Within the virtual machine, the software components run in _containers_ that are easy to setup and tear down individually. 
+[Docker](https://docker.com) is a virtualized environment running a [_Docker demon_](https://docs.docker.com/engine/docker-overview), in which you can run software without altering your host system permanently. The software components run in _containers_ that are easy to setup and tear down individually. The Docker demon can use operating-system-level virtualization (Linux, Windows) or a virtual machine (macOS, Windows).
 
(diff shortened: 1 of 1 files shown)


matkoniecz:
> You need sufficient disk space of _several Gigabytes_

Is it really worth mentioning? I think that it would be unusual to not have this amount of free space available

> an environment that allows you to create a _virtual machine_

AFAIK it is mac-specific  (from wikipedia - "Docker on macOS uses a Linux virtual machine to run the containers."). "virtualized environment" is true also for mac while virtual machine bit is not true for Linux and Windows where operating-system-level virtualization is used.

> increase the Memory assigned in the Docker Preferences > Advanced > Adjust the computing resources.

It is tricky, as even existence of "Docker Preferences" menu is system-dependent.

polarbearing:
Size - If you have a tower under your desk with several TB disks, you don't need to worry. My 2yo laptop has a 500 GB SSD which is pretty full, in particular with other virtual images around. I was spending the whole night observing the virtual  machine growing like sour dough, and shuffling other stuff out. The problem was that I had no indication what size to expect in the end. 

virtual machine - yes, reading further there is the Docker daemon, which might reside in a virtual machine depending on OS. 

Preferences - might put that in brackets as examples.

polarbearing:
Issues should be addressed.

kocio-pl:
@matkoniecz What do you think about it now?

kocio-pl:
@matkoniecz ping?

kocio-pl:
Thanks! I have finally find the time to look myself and I find these remarks useful if I would ever use Docker on macOS or Windows. It also looks to me that all the requests have been resolved.

PR #3024, opened 2018-01-17
add tourism=apartment

### LINKED ISSUE #401: Add rendering for tourism=apartment

polarbearing:
Resolves #401 
Renders from z>=18, i.e. one higher than hotel etc.
On a node: ta_node  On a way: ta_building
Node on residential background: ta_residential_background

### DIFF
--- a/amenity-points.mss
+++ b/amenity-points.mss
@@ -248,6 +248,13 @@
     marker-clip: false;
   }
 
+  [feature = 'tourism_apartment'][zoom >= 18] {
+    marker-file: url('symbols/apartment.svg');
+    marker-fill: @accommodation-icon;
+    point-placement: interior;
+    marker-clip: false;
+  }
+
   [feature = 'amenity_hospital'][zoom >= 15] {
     marker-file: url('symbols/hospital.svg');
     marker-fill: @health-color;
@@ -1774,6 +1781,7 @@
   [feature = 'tourism_hostel'][zoom >= 17],
   [feature = 'tourism_chalet'][zoom >= 17],
   [feature = 'tourism_guest_house'][zoom >= 17],
+  [feature = 'tourism_apartment'][zoom >= 18],
   [feature = 'tourism_wilderness_hut'][zoom >= 14],
   [feature = 'tourism_camp_site'][zoom >= 17],
   [feature = 'tourism_caravan_site'][zoom >= 17], {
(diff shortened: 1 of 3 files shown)


kocio-pl:
Your third test has a different color of the label, but the code and my tests show there's no such problem in reality, both nodes and ways are working as expected.

The only thing I would change is a name of the image - it should be rather `apartment.svg` for simplicity.

matkoniecz:
Is it using transparency?

polarbearing:
I was using tourism_* intentionally so it would not get confused with building=apartments or so. 

As for the icon: I applied the same method than guest_house for creating the halftone bars, this is using fill-opacity in the SVG. The reason is that when mapnik colours the icon, it applies the map-colour to all elements in the SVG equally. So setting an element to solid black, grey, or even white all results in solid tourism-blue. Fill-opacity was also used for other cases in the amenity.mss to faint certain icons, so I assume it would be usable.

kocio-pl:
Since we're not showing any icons for buildings, I don't think it would be a problem to simplify file name.

I think we can use halftones more often, because we need more tools to make icons different and more meaningful. It's hard enough because of small canvas and using basically one color (rarely it's two, like in a bus station). Average icon design time is many months or even years, which is bad enough for me. Luckily Mapnik accepts halftones in SVG, so we can use this feature whenever it's suitable.

polarbearing:
Fine. Are you able to adjust the filename in the merge, or do you need the commit amended?

kocio-pl:
I'm not sure, therefore I prefer if you just do it, please.

polarbearing:
done, and squashed commits

kocio-pl:
Thanks! I will check it soon and if it still works, I'll merge it.

polarbearing:
Thanks. 
Talking about subdirectories, would you see a value in creating further subdirectories under symbols, e.g. one PR per theme? tourism and transport would be good candidates.

kocio-pl:
That makes sense for me. Of course some icons are reused in different categories (like bookshop/library), but I guess anything with more than 10 elements is a natural candidate for making subdirectory.

PR #3027, opened 2018-01-19
Add rendering for shop=chocolate

jbelien:
Add rendering for `shop=chocolate` based on [wiki page](https://wiki.openstreetmap.org/wiki/Tag:shop%3Dchocolate)

### DIFF
--- a/amenity-points.mss
+++ b/amenity-points.mss
@@ -1947,6 +1947,7 @@
   [feature = 'shop_convenience'],
   [feature = 'shop_confectionery'],
   [feature = 'shop_pastry'],
+  [feature = 'shop_chocolate'],
   [feature = 'shop_copyshop'],
   [feature = 'shop_deli'],
   [feature = 'shop_doityourself'],
--- a/amenity-points.mss
+++ b/amenity-points.mss
@@ -672,6 +672,7 @@
       marker-file: url('symbols/shop/convenience.svg');
     }
 
+    [shop = 'chocolate'][zoom >= 18],
     [shop = 'confectionery'][zoom >= 18],
     [shop = 'pastry'][zoom >= 18] {
       marker-file: url('symbols/shop/confectionery.svg');
--- a/project.mml
+++ b/project.mml
@@ -1551,7 +1551,7 @@ Layer:
                                'mobile_phone', 'motorcycle', 'musical_instrument', 'newsagent', 'optician', 'jewelry', 'jewellery',
                                'electronics', 'chemist', 'toys', 'travel_agency', 'car_parts', 'greengrocer', 'farm', 'stationery',
                                'laundry', 'dry_cleaning', 'beverages', 'perfumery', 'cosmetics', 'variety_store', 'wine', 'outdoor',
-                               'copyshop', 'sports', 'deli', 'tobacco', 'art', 'tea', 'coffee', 'tyres', 'pastry') THEN shop
+                               'copyshop', 'sports', 'deli', 'tobacco', 'art', 'tea', 'coffee', 'tyres', 'pastry', 'chocolate') THEN shop
                                ELSE 'other' END AS shop,
             NULL AS way_pixels
           FROM planet_osm_point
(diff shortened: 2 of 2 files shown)


matkoniecz:
Welcome!

technical: Is this PR tested? I am pretty sure that https://github.com/gravitystorm/openstreetmap-carto/blob/ff78bb79825cb64e82f15ee61d9637da56cb3c3b/project.mml#L1444 also needs to be changed (currently shop=chocolate is transformed to shop=other in this query).

cartography: I am not entirely sure is https://github.com/gravitystorm/openstreetmap-carto/blob/master/symbols/shop/confectionery.svg a better icon for chocolate shop than generic dot.

On the one hand it chocolate -> sweets -> cake/muffin is not ridiculous, on the other hand there is no direct link between chocolate and icon,

jbelien:
Hi @matkoniecz ,

Indeed I forgot to tweak `project.mml`, it's fixed now !

I agree it's not a perfect icon for `shop=chocolate` but I based this PR on the "*See also*" part of the [wiki page](https://wiki.openstreetmap.org/wiki/Tag:shop%3Dchocolate) and mainly because here in Belgium, some people are mapping `shop=chocolate`as `shop=confectionery` so the shop is "displayed" on the map (which is not a good thing to do, I know :( ) and the icon is not ridiculous indeed ! :)

polarbearing:
Here is the icon: 
Not sure, is it supposed to be a muffin or a praliné?

Adding an svg works as follows: ``````

Could one of you amend the title please, "Add rendering for shop=chocolate" would work.

kocio-pl:
> Not sure, is it supposed to be a muffin or a praliné?

I have designed it to resemble a muffin. They come in many different shapes, some pralines can even look similar to muffins:

https://commons.wikimedia.org/wiki/File:Chocolate_truffles_-_comida_japonesa.jpg

but I would probably try with this one (basically rounded trapezoid with a dot on a top):

http://4.bp.blogspot.com/-BCrP56wIeMc/T53JKgnLEfI/AAAAAAAABj4/hvnry2pwyPc/s640/hazelnut+praline+coffee+ganache.jpg

polarbearing:
praline1 praline1 praline1

kocio-pl:
I would make them higher and with slanted sides. BTW - why there are 3 the same files?

polarbearing:
Sides were slanted already, so it needs to be stronger. BTW- because I am addicted to chocolate. 
praline2 praline2 praline2

kocio-pl:
I meant something like this:
chocolate-2

polarbearing:
Ok that's just bigger. I was trying to keep the proportion relative to the muffin for sweetshops. 
With both sizes, I have some doubts that people would recognise it.

jbelien:
Sorry but I think  looks more like a "castle" than a truffle/praline :/

kocio-pl:
OK, I guess we could use standard muffin icon for chocolate than. It makes sense, because these shops make a small cluster with similar meaning ("sweets").

polarbearing:
Last attempt: 3 pralinés top view with an almond: 
praline3 praline3 praline3

kocio-pl:
Also hard to recognize as praline, it's closest to this icon:

https://github.com/mapbox/maki/blob/master/icons/america-football-11.svg

kocio-pl:
@jbelien It looks like you idea of icon is the best at the moment, I just want you to extend rendering for all the proper cases. Currently it's added in `amenity-points-poly` layer only, and since this is very similar to all the other shops, try to find all the places with pastry and add it there too. Otherwise it would be selected only for areas, not points, and only for icons, not names (`project.mml` is meant for selecting things). The same hint for the `amenity-points.mss` file - defining icon shape does not define name rendering (MSS files are for defining the exact look).

This should be quite easy to fix, you just have to follow pastry shops code closely, but remember to make test rendering to avoid simple code mistakes. if you need any help, just let us know.

kocio-pl:
I haven't checked it yet, but it looks like OK. Probably the only thing missing is to squeeze both commits into one.

polarbearing:
@jbelien - [this](https://gist.github.com/patik/b8a9dc5cd356f9f6f980) is what he is talking about. 
@kocio-pl - accrdong to [the doc here](https://help.github.com/articles/about-pull-request-merges/) and [here](https://github.com/blog/2141-squash-your-commits) I understand that the repo admin should nowadays be able to do that while merging?

kocio-pl:
Looks like "Squash and merge" option is available - so I just have to test rendering. But it's always good if the author of the code checks the rendering first.

jbelien:
> Looks like "Squash and merge" option is available - so I just have to test rendering. 

Indeed, you can "Squash and merge" my PR from GitHub interface :)

> But it's always good if the author of the code checks the rendering first.

Sorry, didn't take the time to do so. Would be more than happy if you can test it but if not, I'll try to make some time to do so :)

kocio-pl:
Thanks for your work, it looks like a proper code.

kocio-pl:
Is squashing better than standard merging? In the commit history tree I see no lines going back (sometimes many months ago) and I think that might be healthier, since the author is kept, but the exact time when the patch was ready is not that important.

polarbearing:
As far as I understand, it is in the interest of the repository to have only one commit per PR for clarity, and not several with small iterations. This can be achieved
1. by the author squashing the commits before the PR
2. by the "Squash and merge" in github.

1 can only be done offline and requires to pull, do the changes, squash, and push. It's a bit complicated for somebody not contributing regularly and requires command-line interaction. 

2 might be easier for the author, in particular if the edit itself is done in the web interface, using the pencil button. You can recognise those branches by having automatic names: ```author:patch-1```

kocio-pl:
I was wondering about situation when the author had just 1 commit, which can be merged both ways.

Issue #2704, opened 2017-07-24
Render tourism=museum areas


Tomasz-W:
I found that some people are mapping open-air museums as theme parks. It's propably because tourism=theme_park areas are rendered with brown outline, which gives information about object size, and tourism=museum tagged as areas are not. Similar tourism=zoo areas also have the same brown outline. Examples:
https://www.openstreetmap.org/way/248994901
https://www.openstreetmap.org/way/480832184

There is around 3k open-air museums (propably that's it, 25k of museums areas, excluding 22k of cobinations with building=*, gives 3k)

I think museum areas (excluding combinations with building=* key) should also have this brown outline, which will give map user an information about size of the museum.

wmyrda:
I agree that original idea to have only the icon was made probably made due to most museums being located in the building hence it seemed satisfactory as the buildings are visible and building=yes together with tourism=museum gave resemblance that all information is available to the user. There are however number of folk museums that are composed of several building and it those cases users tend to map those museums as theme_parks which they are not only to see their areas visible on the map like here http://www.openstreetmap.org/way/41714371
I also agree that using brown outline just like with zoo or theme_park would be appropriate way to render those areas hence they are IMHO just as attractive to the tourists. Last, but not least due to them being quite scarce on any given area it would not make to any overcrowding problem on the default map.

polarbearing:
Consequently a museum campus should get the name scaling with way_area, see #1624

kocio-pl:
Name scaling would be a problem, since I guess we would still like to have an icon. This is general problem which caused dropping https://github.com/gravitystorm/openstreetmap-carto/pull/2361.

kocio-pl:
What do you suggest then regarding label scaling and the icon (which has fixed size)?

polarbearing:
As a general problem this would need a general solution, that leads out of the cul-de-sac.

What about treating the icon like a symbol-character, that runs in-line with the text. Thus it has a minimum size if not scaled (small objects, objects without name), and scales along with the text, so it would not appear too monumental above it. 
Unicode/github mock-up, just to illustrate the idea:

### ⚓︎ Small harbour
## ⚓ Medium harbour
# ⚓ Big harbour

kocio-pl:
Hm, we should have icons scaling code then. SVG icons are ready to be scaled up, but is it possible with CartoCSS+Mapnik?

pnorman:
[...]

Issue #2290, opened 2016-08-18
Render leisure=resort


zetx16:
Render leisure=resort and label.
Object for example: http://www.openstreetmap.org/way/290825122
I think the color is set like a tourism=camp_site (`@campsite: #def6c0;`)

pnorman:
notes from taginfo

1.5k uses, usage is not worldwide
![image](https://cloud.githubusercontent.com/assets/1190866/17763494/f0cfe1c0-64cd-11e6-8814-005c57aed7c2.png)

jojo4u:
The [wiki article](https://wiki.openstreetmap.org/wiki/Tag:leisure%3Dresort) created in 2015 is ok There was no proposal but some [discussion ](https://lists.openstreetmap.org/pipermail/tagging/2013-February/012883.html) in 2013.

d1g:
[RU:Tag:leisure=resort provides more information](https://wiki.openstreetmap.org/wiki/RU:Tag:leisure%3Dresort) about additional sub classes. We would definitely tag these objects, as they were requested by users (pre Oct 2015) and still functional. The only question is uniform  tagging.

Most of them have nothing to do with hotels/motels/hostels, in sense you wouldn't visit leisure=resort + resort=kids_camp as adult, unless your children are there.

You wouldn't visit other leisure objects unless your are ill/old/ and so on.

re @pnorman  I don't think that mostly outdoor facilities dedicated to children resorts leisure=resort + resort=kids_camp ("детский лагерь") can have global coverage. Not many countries make such distinction, am I wrong?

Wikipedia says that they are present at least in some other countries: https://en.wikipedia.org/wiki/Summer_camp#Around_the_world but are they used? 

_"In Canada, summer camps are very popular. About 70% of Canadian camps tend to be affiliated with organizations, while the rest are private."_

pnorman:
They're present around the world and used, but leisure=resort is mainly used in one area.

kocio-pl:
They are concentrated around Europe, but they are also tagged all around the world, so it looks OK for me. I would render them just like other leisure/sport areas.

polarbearing:
I'd see them more tourism-oriented, thus looking towards camp-sites as in the first post might be better than leisure/sport which is more everyday activities. 

We could also go a step forward and consider all campuses for touristic accommodation, hotel, resort, hostel etc, see #1624 (accommodation). 

I'd imagine a blue outline (similar to the brown for theme_parks)

[...]

PR #3035, opened 2018-01-25
Made landuse=military areas less prominent

andrzej-r:
Ref #2670

Made landuse=military areas less prominent by:
- Removing the outline
- Changing the fill pattern to use a finer (but more dense) lines
- Changing the fill pattern to to use lower alpha value (more transparent)
- Increased the visibility threshold at zoom levels 8+ from 100/75px to 200px


matkoniecz:
Can you add some before/after images?

andrzej-r:
Zoom level 10
l10-orig
l10-modified

Zoom level 11
l11-orig
l11-modified

polarbearing:
When removing the outline we cannot see the boundary of the feature, in particular when it is not identical to other underlying features. 
In #2290 and #2704 we experiment with line width=2 which is a bit thinner than the current 3 in military. Maybe we can use the same here. 
What happens when we fully remove the pink between the hatch lines?
Can we see some previews with larger areas and in higher zoom?

polarbearing:
Example with your finer hatching and outline width 2, see left the proposed museum outline.
military_line_2_thin_hatching

polarbearing:
No outline in z18. 
military_line_0_thin_hatching

kocio-pl:
We were discussing making all the uniform services areas look the same (military, police and fire services). The proposition was to get rid of hatching and leave just a red area - with the red outline (by default, because we were not thinking about more changes). Recent [experiments with museum outline]( https://github.com/gravitystorm/openstreetmap-carto/issues/2704) show, that the outline should be specific in order to make a fence visible, if it's present. Could anybody show how the outline of a military area looks with a fence currently?

polarbearing:
Military with fence is well visible since the grey contrasts against the pink: 
https://www.openstreetmap.org/#map=18/52.42578/13.57413
military_fenced

andrzej-r:
I will upload additional screenshots in the evening. One interesting use case is near Farnborough, with eg several overlapping military areas. 

I am not too attached to this particular proposal. I just consider current rendering of military areas way too prominent. 

I didn't get rid of hatching because I thought that may be one change too much. Besides, IMHO military areas should *not* be treated like other land uses. Especially in certain countries. 

I was thinking about leaving 1px border (2px is still quite obnoxious, in my opinion) but then I realised removing the border entirely works just fine. We need some transparency in the fill/hatching pattern to display overlapping areas but that's all. No border or 1px border should also work better with fences. 

Fill pattern was made a bit more transparent because there are often quite many mapped objects inside these areas and the previous fill pattern was obscuring them.

Filtering settings were made more aggressive to prevent large numbers of small areas from dotting the map at medium zoom levels. It is analogous to the airport/airfield issue discussed in another ticket.

polarbearing:
I agree with reducing military, just it needs to be in sync with our plans for police etc. We could use the outline without the hatching.

kocio-pl:
General note: I'm happy that you both are really testing these things! That's what is needed in my opinion.

I don't see the fence on the military outline, but it looks like the problem no one really notice, since it was not reported as of now - so maybe it's not that important? However if there is a way to make the difference visible somehow, it's preferable.

kocio-pl:
I think we could leave the hatching only for danger areas (see #2739 and #684). Currently they are shown with light "x" signs, which could be just dropped then, because addition of hatching would be enough.

andrzej-r:
So, how about:
- no hatching
- 1px border
- ~10% opaque fill with the same colour as now (currently 14%, 8% in my proposal)
- 200px threshold at zoom 8+?

If that's OK, I will update the PR. Details can always be later refined to match police and fire services etc.

andrzej-r:
Promised screenshot of a more complex use case at higher zoom level (15/51.3064/-0.6725):
l15

polarbearing:
@kocio-pl - how do we distinguish the military from the other uniformed services when they have only the same outline?

andrzej-r:
I am OK with using only 1px border for small areas (in pixels). For larger areas (in pixels) we need either tinting or hatching, as we may see only part of the border. In both cases the fill should be unobtrusive.

I don't consider a color shift a problem. In fact, IMHO, it would be good to have eg commercial areas tint building colors, but that's a topic for another discussion. 

Hatching or tinting is really an aesthetic matter, with hatching being a "low tech" substitute to tinting. It is just down to an arbitrary decision.

kocio-pl:
> how do we distinguish the military from the other uniformed services when they have only the same outline?

The same as we distinguish schools from other societal amenities: they have no icon, while the rest have some.

kocio-pl:
> If that's OK, I will update the PR. 

First please show how would it look like then, so we could judge the result.

polarbearing:
I don't see hatching as "low tech". Removing/avoiding colour merging effects was a main goal in this style for quite a while (I wonder if that is documented?). The style has subtle differences in the same colour group that get lost when colours merge. 
Before making new PRs I would find it a good idea to agree on the style of outline. I am working on some for tourism features (#2290, #2704) and it would be good to apply the scheme in sync, e.g. the width in various zoom levels.

kocio-pl:
In case anybody coming here from weeklyOSM 393 ( http://www.weeklyosm.eu/archives/9939 ) - this issue was closed by its author **without merging**, so it will certainly not affect next release of osm-carto.

paulusr:
I'm here from weeklyOSM 393, just like to say that I was hoping this would go ahead as current style needs fixing.

kocio-pl:
The problem is not neglected, just check #3045 and help us there to decide and test solutions.

andrzej-r:
New proposal is in PR #3057.

PR #3046, opened 2018-01-31
Use UTF8 in createdb

Ircama:
**Use UTF8 in createdb**

Minor hint within the documentation.

Mention that the character encoding scheme to be used in the *gis* database shall be UTF8 (e.g., useful when using [Ubuntu for Windows](https://www.microsoft.com/en-us/store/p/ubuntu/9nblggh4msv6) within  [Windows Subsystem for Linux](https://docs.microsoft.com/en-us/windows/wsl/install-win10), where also `template template0` is needed).

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -9,10 +9,12 @@ Start by creating a database
 
 ```
 sudo -u postgres createuser -s $USER
-createdb gis
+createdb -E UTF8 -l en_GB.UTF8 gis
 ```
 
-Enable PostGIS and hstore extensions with
+The character encoding scheme to be used in the database is *UTF8* and the adopted collation is *en_GB.UTF8*.
+
+Enable *PostGIS* and *hstore* extensions with
 
 ```
 psql -d gis -c 'CREATE EXTENSION postgis; CREATE EXTENSION hstore;'


kocio-pl:
Sounds good to me, but could anybody test if it really works as expected? I'm not able to do it currently.

Ircama:
We might also add the locale settings for the database: `--locale=en_US.utf8`

e.g., `createdb --encoding=UTF8 --locale=en_US.utf8 [ -O $USER ] [ --template=template0 ] gis`

pnorman:
Why?

If we do specify it it should be `-E utf8`, nothing more. Long options sometimes add clarity, but not when the overwhelming standard is the short option.

Ircama:
@pnorman: OK for showing short options, in fact the argument options are self explaining.

To be sure I got your remark, we might opt for

    createdb -E UTF8 -l en_US.utf8 [ -O $USER ] [ -T template0 ] gis

or simply

    createdb -E UTF8 gis

What would you suggest?

Thanks

dieterdreist:
2018-02-01 9:16 GMT+01:00 Ircama <notifications@github.com>:

> @pnorman <https://github.com/pnorman>: OK for showing short options, in
> fact the argument options are self explaining.
>
> To be sure I got your remark, we might opt for
>
> createdb -E UTF8 -l en_US.utf8 [ -O $USER ] [ -T template0 ] gis
>
> or simply
>
> createdb -E UTF8 gis
>
>
>

the content of the db is in all kind of languages (local names), the keys
are in British English, so it would be en_GB. "US" doesn't make any sense.

kocio-pl:
Even some keys aren't defined in US English (like [soccer](https://wiki.openstreetmap.org/wiki/Tag:sport%3Dsoccer)), so any locale won't work anyway. 

As of names - is UTF8 enough or maybe some names in different languages would need bigger space (like UTF16 or maybe even UTF32)?

SomeoneElseOSM:
What actually goes wrong if you don't do this, and why is it an issue only on Windows (or on "Windows subsystem for Linux")?  

When I wrote [this](https://www.openstreetmap.org/user/SomeoneElse/diary/42617) I didn't specify any encoding, and didn't notice a problem (although the name values there are likely all Latin).  I've regularly used [these instructions](https://switch2osm.org/manually-building-a-tile-server-16-04-2-lts/) without an encoding or locale setting with non-Latin names and haven't noticed a problem.

dieterdreist:
2018-02-01 12:35 GMT+01:00 SomeoneElseOSM <notifications@github.com>:

> What actually goes wrong if you don't do this, and why is it an issue only
> on Windows (or on "Windows subsystem for Linux")?
>
> When I wrote this
> <https://www.openstreetmap.org/user/SomeoneElse/diary/42617> I didn't
> specify any encoding, and didn't notice a problem (although the name values
> there are likely all Latin). I've regularly used these instructions
> <https://switch2osm.org/manually-building-a-tile-server-16-04-2-lts/>
> without an encoding or locale setting with non-Latin names and haven't
> noticed a problem.
>


I believe you should set an encoding (UTF8) and have always done so, but
I've never specified a locale. If you didn't have to specify UTF8 and it
worked then UTF8 was probably the default on your system (or another
encoding which could handle the characters, or maybe you simply didn't
notice that you had problems in Tamil (for instance)).

dieterdreist:
2018-02-01 12:35 GMT+01:00 SomeoneElseOSM <notifications@github.com>:

>  I've regularly used these instructions
> <https://switch2osm.org/manually-building-a-tile-server-16-04-2-lts/>
> without an encoding or locale setting with non-Latin names and haven't
> noticed a problem.



actually in switch2osm there is an encoding setting: "

createdb -E UTF8 -O renderaccount gis

SomeoneElseOSM:
> createdb -E UTF8 -O renderaccount gis

That would explain it :)

Ircama:
> What actually goes wrong if you don't do this, and why is it an issue only on Windows (or on "Windows subsystem for Linux")?

I considered useful to suggest to explicitly mention the UTF8 parameter after noticing that my postgres installation with WSL was setting template1 encoding to SQL_ASCII (it was just a very quick test due to my limited free time unfortunately; anyway this helped me to observe that the UTF8 setting might not always be a default).

We use U&"..." escaped Unicode characters in *project.mml* and this syntax should work [only when the server encoding is UTF8](https://www.postgresql.org/docs/9.5/static/sql-syntax-lexical.html).

Collation (`-l` option) might in theory influence indexes, sort ordering and comparisons within the SQL statements in *project.mml*. I’m not sure it is worthwhile to be mentioned anyway.

kocio-pl:
> We use U&"..." escaped Unicode characters in project.mml and this syntax should work only when the server encoding is UTF8.

For me it's a solid reason to merge it as it is now (making the things we depend on explicit instead of implicit). What the others think about it?

Ircama:
> the keys are in British English, so it would be en_GB. "US" doesn't make any sense.

Thanks, right.

I also checked [this Chef configuration code](https://github.com/openstreetmap/chef/blob/master/cookbooks/postgresql/resources/database.rb#L25-L27) which reports *UTF8* encoding and *en_GB.UTF8* collation.

INSTALL.md
```diff
@@ -9,10 +9,12 @@ Start by creating a database
 
 ```
 sudo -u postgres createuser -s $USER
-createdb gis
+createdb -E UTF8 -l en_GB.UTF8 gis
```
pnorman: Don't specify the collation.

INSTALL.md
```diff
@@ -9,10 +9,12 @@ Start by creating a database
...
 ```
 
-Enable PostGIS and hstore extensions with
+The character encoding scheme to be used in the database is *UTF8* and the adopted collation is *en_GB.UTF8*.
```
pnorman: We don't need to change this text, and what are the implications of changing the collation?

kocio-pl:
> I still question the usefulness of this idea in general.

I understand, but you have not related to https://github.com/gravitystorm/openstreetmap-carto/pull/3046#issuecomment-362557585 - could you elaborate why this is not a good reason for you?

All the other changes are not necessary for me.

INSTALL.md
Ircama: Collation is not really changed, but just set explicitly, as the default might not always be `’en_GB.UTF8’`. In some installations it might be `'C'` by default for instance and in some other ones defaulted to other locales (like `’it_IT.UTF8’`). In theory it might affect indexes, sort ordering and comparisons within SQL statements in *project.mml*.

kocio-pl:
@pnorman ping?

INSTALL.md
pnorman: I'm not aware of any cases where it matters. The only ordering by user-supplied strings that is done is in the last term of an order by clause to get a consistent ordering across metatiles. The collation doesn't matter for this.

pnorman:
> I understand, but you have not related to #3046 (comment) - could you elaborate why this is not a good reason for you?

There's half a dozen defaults that get set up when the cluster is created and are defaults for creating a DB. Is this needed on sanely set up systems? I'm not aware of any sane setups that don't have a UTF8 encoding for the cluster.

kocio-pl:
Because this small addition is not necessary (we haven't seen any problems with it yet) and we have disagreement, I will close this PR now. However thanks all for the code and discussion!

PR #3048, opened 2018-01-31
Fix rendering bug (#3043) with access=yes|permissive tags.

### LINKED ISSUE #3043: Strange rendering of amenity=toilets as a brown circle at zoom level 19
Since the last update (v4.7.0), `amenity=toilets` are rendered as a brown circle instead of the toilets symbol at zoom level 19 (but not at zoom levels 17–18):

screen shot 2018-01-29 at 20 43 35

I suspect this to be a bug, because I couldn't fine a reference in the change log (and this change doesn't seem to make sense).

tpikonen:
Nodes with access=yes|permissive were rendered as brown circles at z=19
due to overly general entrance rendering rules. Tighten the entrances
layer code to apply only to entrance nodes. Fixes #3043.

### DIFF
--- a/buildings.mss
+++ b/buildings.mss
@@ -59,7 +59,7 @@
       marker-file: url('symbols/square.svg');
     }
   }
-  [zoom >= 19] {
+  [zoom >= 19]["entrance" != null] {
     ["entrance" = "yes"],
     ["entrance" = "main"],
     ["entrance" = "home"],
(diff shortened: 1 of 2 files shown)


zerebubuth:
:+1: Looks good to me!

kocio-pl:
I have checked all the examples and it really fixes the problem - thanks you for the fast solution!

PR #3054, opened 2018-02-04
add barrier=log like block

### LINKED ISSUE #3052: Add Rendering of Barrier=log
It has 2,874 uses. It would be cool to be able to see when they are blocking roads or acting as barriers in more primitive parking lots. There's a few parking lots and roads around where I live with logs blocking them that look like they can be accessed when they can't because there is a log blocking them that isn't rendered. I suggest logs be rendered similar to how tree rows are with the rounded corner, except thinner and brown. That way, it would be clearly differentiated from a gate, fence, or any other similarly tagged thing.

polarbearing:
Resolves #3052 
Add barrier=log to the non-opening barrier types 'bollard' and 'block'.
z19 before/after (Kosmtik):
z19_log_before z19_log_after

### DIFF
--- a/amenity-points.mss
+++ b/amenity-points.mss
@@ -1114,7 +1114,8 @@
   }
 
   [feature = 'barrier_bollard'],
-  [feature = 'barrier_block'] {
+  [feature = 'barrier_block'],
+  [feature = 'barrier_log'] {
     [zoom >= 16] {
       marker-width: 3;
       marker-line-width: 0;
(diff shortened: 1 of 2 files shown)


kocio-pl:
I think brown dot (just like a trunk in a single tree rendering) could be even better here, but on the other hand it might be located on a track, which is brown and have dots in standard pattern (with no `tracktype=*`). Could you test that?

polarbearing:
Barriers are so far black or grey: barrier_gate (no colour value, thus black), lift/swing_gate #3f3f3f, block #7d7c7c, I see no reason to use colours we have for tourism or nature. The log is a denaturized building element in this function. And, as you assume, it would collide with track. 

z19 log on track in proposed gray:
z19_log_on_track

kocio-pl:
Thanks, it makes sense for me, so I hope I will test it soon and merge the code.

PR #3055, opened 2018-02-05
Move amenity=toilets to higher zoom levels

kocio-pl:
Related to #1745.

This change moves the toilets to higher zoom levels, because in well mapped areas they compete with other amenities like shop or cafe. Also this is rather a short-distance amenity, so lower zoom levels are not relevant here.

The problem is that we don't even know if they are standalone amenities or just a part of some other amenity. So I guess that toilets with access for everyone ("yes") or paid ("customers'" or "destination") are more important and most probably standalone, but status of the rest (including "permissive" and lack of the access tags) is not clear and they are probably less important. 

[Louvre example](https://www.openstreetmap.org/#map=17/48.86141/2.33593) (click the images to view full size)):
z17
Before
nzhd7g_9
After
ijkxw4ux

z18
Before
zfvswuzo
After
ncv4xhfa

### DIFF
--- a/amenity-points.mss
+++ b/amenity-points.mss
@@ -533,11 +533,14 @@
     marker-clip: false;
   }
 
-  [feature = 'amenity_toilets'][zoom >= 17] {
-    marker-file: url('symbols/toilets.svg');
-    marker-fill: @amenity-brown;
-    marker-placement: interior;
-    marker-clip: false;
+  [feature = 'amenity_toilets'] {
+    [access = 'yes'][zoom >= 18],
+    [zoom >= 19] {
+      marker-file: url('symbols/toilets.svg');
+      marker-fill: @amenity-brown;
+      marker-placement: interior;
+      marker-clip: false;
+    }
   }
 
   [feature = 'amenity_drinking_water'][zoom >= 17] {


matthijsmelissen:
I don't like this PR, as it punished more specific tagging ('access=yes').

kocio-pl:
I don't understand why do you think this way. What would be your proposition?

The key problem is to make a sane hierarchy and `access=yes` is higher than some other `access=*` tags and higher than toilets without `access=*` tag. This is not a perfect rule to detect standalone toilets vs toilets being a part of some other objects (like cafe, museum, shopping center or an office), but rough approximation.

mboeringa:
> The key problem is to make a sane hierarchy and access=yes is higher than some other access=* tags and higher than toilets without access=* tag. This is not a perfect rule to detect standalone toilets vs toilets being a part of some other objects (like cafe, museum, shopping center or an office), but rough approximation.

+1. It is senseless to show non-public toilets at lower zoom levels. A toilet part of a facility like a museum, is not a general public toilet, even if it can be accessed by anyone willing to buy a ticket. That is access=customers.

kocio-pl:
So maybe only `amenity=toilets + access=yes` should be visible on z18 and all the rest on z19? I was thinking about paid toilets services, but "customer" might also mean "customer of a cafe/office/etc", not "toilet's customer" and we don't like to show them too prominent.

kocio-pl:
Updated - `access=yes` should be mostly sane, all the other cases are more suspected about being just the inner part of some other object.

kocio-pl:
Any more comments/reviews?

Issue #3034, opened 2018-01-24
Inconsistent priorities of region/city names at levels 5 and 6+


andrzej-r:
Some, countries (Poland, Italy, Austria, Romania, Greece, Turkey, Ukraine, Syria, Iraq, Iran, Bangladesh, Korea, Japan, Nigeria, Peru, Ecuador, Mexico, Colombia, Mexico) are divided into relatively small administrative regions. In all these countries, at zoom 5, the map shows names of key cities and names of the regions are hidden - I have no problem with that, I often find city names more recognizable than region names, especially when looking at a map of a country I am not familiar with.

However, at zoom levels 6+ the region names are added to the map at *higher* priority than the names of the key cities. As these cities are often centres of the respective regions, the labels sometimes collide and the names of the cities remain hidden at zoom levels 6+ until there is enough space to fit both. This is inconsistent and confusing - I see a name of the city, try to zoom to it only to find it disappear and reappear 2-3 zoom levels later.

Suggested solutions:
- Names of large cities should keep their priority over region names at zoom levels 6+ (consistent with current behaviour at zoom level 5)
- Alternatively, at zoom level 5 the map should display region names instead, or in addition, to the names of large cities.

Possible workarounds:
- Manually move region labels away from the cities
- Block city names from being displayed at level 5 
- Improve layout engine to automatically move region labels when possible (assuming there is space for both).

---

The above is not the only example of such inconsistencies. Similar issue affects India, China, Brasil when switching between levels 4 and 5 (with USA, Russia and Australia probably dealing with it by manual positioning of labels or removing names of large cities).

kocio-pl:
> Improve layout engine to automatically move region labels when possible (assuming there is space for both).

This solution is under way, see https://github.com/gravitystorm/openstreetmap-carto/issues/2962#issuecomment-348882616.

The rest of suggestions needs to be tested.

matthijsmelissen:
> In all these countries, at zoom 5, the map shows names of key cities and names of the regions are hidden

This is not true in general, and there is no difference between z5 and z6.

What is happening that state names are not displayed if the state is too small (measured in pixels), and this is more likely to occur on z5.

andrzej-r:
Thank you for the explanation. I think I now understand how it happens.

Perhaps (assuming we are happy with keeping the priorities) would be better not to limit the area for state labels and let mapnik decide what labels get to be displayed.

That is, instead of:
```
.state {
[...]

Issue #2851, opened 2017-09-18
Country label problems


kocio-pl:
Some country name labels (namely Canada, Greenland and Russia - I have not spotted more of them) are not visible on z3, while on z2 they are present and we have other country names visible on z3.

math1985:
You mean z4 instead of z3 and z3 instead of z2?

This is intended behaviour. Ones a country gets large enough (about large enough to cover my full screen ;)) the name is not displayed anymore. The reason is that by the time you're zoomed in so far, you should already know which country you're looking at, and smaller detail is more important to be rendered.

kocio-pl:
Yes, you're right with zoom levels - z3-z4. 

The problem is that on z4 there's no any other details at all that could be rendered for these countries. You describe the typical mid/high zoom level situation, where we hide some data to show something else - but in this case we don't gain anything, we just loose some information that really belongs here and it makes the map inconsistent (there's an empty place where the name could be, while the other country names are hidden only when needed).

BTW: maybe we could use some Merkator geometry fix for country labels - especially for Canada?

kocio-pl:
Russia name label is now visible only on z3, because it's too big for z4. On the other hand we start rendering subdivisions from z5 - and some of them are too big by then, so (a) their names are never shown and (b) the map lacks _any_ labels in Russia on z4 (except 2 cities in European part), which is quite strange for me.

I think we should be more tolerant with showing names of big areas to avoid this side effect. This might be just low zoom level fix.

kocio-pl:
There are some other labels missing now - Bolivia and Brazil. It's probably tagging change, but it's good to check if it's not a style problem.

kocio-pl:
Now it's even worse than before on z3-z4:
- Canada is not visible at all
- USA name is visible only on Alaska, but not on the main country area

kocio-pl:
I think on z3-z4 I would entirely drop upper country area limit, so at least on these levels we can see the biggest countries names.

BTW: Bolivia and Brazil are back, so it must have been some tagging issue.

PR #3056, opened 2018-02-05
[WIP] Improve country/state label rendering

matthijsmelissen:
* Stop rendering country/state labels slightly earlier
* Scale country/state labels with country/state size instead of with zoom
  level (similarly to landcover labels)
* Set text-wrap to a constant em

### DIFF
--- a/placenames.mss
+++ b/placenames.mss
@@ -4,34 +4,22 @@
 @state-labels: desaturate(darken(@admin-boundaries, 5%), 20%);
 
 .country {
-  [zoom >= 3][way_pixels > 1000][way_pixels < 360000] {
+  [zoom >= 3][way_pixels > 1000][way_pixels < 192000] {
     text-name: "[name]";
-    text-size: 10;
-    text-wrap-width: 35; // 3.5 em
-    text-line-spacing: -1.5; // -0.15 em
+    text-size: @landcover-font-size;
+    text-wrap-width: @landcover-wrap-width-size;
+    text-line-spacing: @landcover-line-spacing-size;
     text-margin: 7.0; // 0.7 em
-    [zoom >= 4] {
-      text-size: 11;
-      text-wrap-width: 40; // 3.6 em
-      text-line-spacing: -1.4; // -0.13 em
-      text-margin: 7.7; // 0.7 em
-    }
-    [zoom >= 5] {
-      text-size: 12;
-      text-wrap-width: 45; // 3.8 em
-      text-line-spacing: -1.2; // -0.10 em
+    [way_pixels > 12000] {
+      text-size: @landcover-font-size-big;
+      text-wrap-width: @landcover-wrap-width-size-big;
+      text-line-spacing: @landcover-line-spacing-size-big;
       text-margin: 8.4; // 0.7 em
     }
-    [zoom >= 7] {
-      text-size: 13;
-      text-wrap-width: 50; // 3.8 em
-      text-line-spacing: -1.0; // -0.08 em
-      text-margin: 9.1; // 0.7 em
-    }
-    [zoom >= 10] {
-      text-size: 14;
-      text-wrap-width: 55; // 3.9 em
-      text-line-spacing: -0.7; // -0.05 em
+    [way_pixels > 48000] {
+      text-size: @landcover-font-size-bigger;
+      text-wrap-width: @landcover-wrap-width-size-bigger;
+      text-line-spacing: @landcover-line-spacing-size-bigger;
     }
     text-fill: @country-labels;
     text-face-name: @book-fonts;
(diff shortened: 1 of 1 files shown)


matthijsmelissen:
screen shot 2018-02-05 at 23 36 33
screen shot 2018-02-05 at 23 36 37

kocio-pl:
For me the idea of different country size labels is OK, but I feel strange with different label size of states, even if the look is not too bad - I need some time too judge it with a cool head, probably.

First batch of rendering - Poland, of course 😃 (click to see full size images), it looks OK for me. For such country size <z6 doesn't make sense, because no state labels are visible yet, so:

z6
Before
j8zfr6ga
After
ibrvmxjv

z7
Before
dkvyympu

After
b6na6oqp

matthijsmelissen:
A bit unfortunate that Wielkopolskie is the only state that surpasses the threshold. I think it will look less bad when there are other countries around where the label size is also variable.

kocio-pl:
Mazowieckie would be the same if not the problem with admin center city eclipsing (that's a story with Mapnik option `grid` I wait for). It might be hard to find a threshold which works nice with all the countries.

andrzej-r:
I know the intentions are good but the result is misleading and confusing. Admin regions, unlike ad-hic landcover areas, are well structured and IMHO labels should reflect that, that is, use a single label format for all regions at the same level of hierarchy.

Technicalities aside, this is going to upset many people. Prepare for a bucket load of bug reports questioning this design. 

BTW, did you have a chance to look at #3034? There's an inconsistency in the way large city names and region names are ordered, affecting most countries (later on I found more examples in in countries like France, I initially thought it didn't have that issue). The Mapnik fix kocio-pl motioned is going to to improve the situation a bit (just like manual placement would) but it is _not_ a fix, as there will still be collisions.

I thought "Stop rendering country/state labels slightly earlier" may have something to do with the above issue but there was no difference in screenshots with respect to the ordering.

matthijsmelissen:
This PR is included in the global rendering demo at http://bl.ocks.org/matthijsmelissen/raw/af7a602c222dbf1ff1a2c0d84ed755b7/.

See for example Brazil where there are large differences in state sizes. To me, the different size rendering looks good, and I have no problem rendering states in different font sizes.

As far as I'm concerned this can be merged.

@kocio-pl what do you think?

kocio-pl:
> Stop rendering country/state labels slightly earlier

I have to compare it with #2851 - I think it should be no label-less areas on low zoom, if possible, abd more aggressive filtering of big areas might produce more such "holes".

> Scale country/state labels with country/state size instead of with zoom level (similarly to landcover labels)

I think it's great for the countries (it may even help with the holes I mentioned above!), but I don't like the effect on states.

> Set text-wrap to a constant em

OK.

andrzej-r:
How about ordering (not filtering) labels by country/state size instead? This way the biggest areas would be the most likely to have their labels rendered, while the smallest ones would likely not fit at low zoom levels.

placenames.mss
```diff
@@ -4,34 +4,22 @@
...
-      text-size: 11;
-      text-wrap-width: 40; // 3.6 em
-      text-line-spacing: -1.4; // -0.13 em
-      text-margin: 7.7; // 0.7 em
-    }
-    [zoom >= 5] {
-      text-size: 12;
-      text-wrap-width: 45; // 3.8 em
-      text-line-spacing: -1.2; // -0.10 em
+    [way_pixels > 12000] {
```
jeisenbe: I think it would be a mistake to adjust the text labels based on way_pixel instead of by admin_level. This over-emphasizes areas at high latitudes (where way_pixels is 2 to 4 time larger than way_area at the equator) and over-emphasizes large provinces over small countries.

placenames.mss
```diff
@@ -43,22 +31,27 @@
...
     text-face-name: @oblique-fonts;
     text-halo-fill: @standard-halo-fill;
     text-halo-radius: @standard-halo-radius * 1.5;
     text-placement: interior;
-    [zoom >= 7] {
-      text-size: 11;
-      text-wrap-width: 50; // 4.5 em
-      text-line-spacing: -0.6; // -0.05 em
-      text-margin: 7.7; // 0.7 em
+    [way_pixels > 12000] {
```
jeisenbe: These lines will make large provinces/states render more prominently than small countries - with "large" and "small" depending on mercator projection as much as actual size.

placenames.mss
```diff
@@ -4,34 +4,22 @@
 @state-labels: desaturate(darken(@admin-boundaries, 5%), 20%);
 
 .country {
-  [zoom >= 3][way_pixels > 1000][way_pixels < 360000] {
+  [zoom >= 3][way_pixels > 1000][way_pixels < 192000] {
```
jeisenbe: This change would require changes to the way_pixels limit in `#admin-text` so that they match for; currently ` [admin_level = '1'][way_pixels >= 360000], [admin_level = '2'][way_pixels >= 360000],` would need to change to [way_pixels >= 192000]

placenames.mss
```diff
@@ -43,22 +31,27 @@
 }
 
 .state {
-  [zoom >= 5][way_pixels > 3000][way_pixels < 196000] {
+  [zoom >= 5][way_pixels > 3000][way_pixels < 192000] {
```
jeisenbe: I would approve of this minor change, but `#admin-text` needs to be changed to match [admin_level = '4'][way_pixels >= 196000] currently, would need to change to [way_pixels >= 192000]

jeisenbe:
@matthijsmelissen were you planning to work on this PR, or close it?

pnorman:
Closing as stale.

PR #3058, opened 2018-02-06
Adding rendering for historic=fort

### LINKED ISSUE #2871: Add rendering for historic=fort
Seeing as rendering for castles is already proposed, please also add support for historic=fort.

Made a possible design for an icon as well, needs some polishing of course, anyone is free to use it if they feel the need to contribute.
[fort.pdf](https://github.com/gravitystorm/openstreetmap-carto/files/1337760/fort.pdf)

kocio-pl:
Resolves #2871.

Icon design is kind of a blend of https://github.com/gmgeo/osmic/blob/master/tourism/castle-fortress-14.svg and https://github.com/gmgeo/osmic/blob/master/tourism/fort-14.svg, which makes sense, since fort is similar to castle (newer form), but its icon in Osmic is not clear for me at 14 px (fort osmic ).

Rendering from z16 like monuments for example, because they are rather alone by design so it shouldn't clash with other objects on the map. It also takes lower priority than museum icon (like [here](https://www.openstreetmap.org/way/37014114#map=16/52.4155/16.8728)), which is good in my opinion.

[Example on the node](https://www.openstreetmap.org/node/710620121):
2ceta9dy

[Example on the way+building](https://www.openstreetmap.org/way/27951036#map=17/52.40590/16.85907):
7ldvybzj

### DIFF
--- a/amenity-points.mss
+++ b/amenity-points.mss
@@ -590,6 +590,13 @@
     marker-clip: false;
   }
 
+  [feature = 'historic_fort'][zoom >= 16] {
+    marker-file: url('symbols/fort.svg');
+    marker-fill: @memorials;
+    marker-placement: interior;
+    marker-clip: false;
+  }
+
   [feature = 'historic_archaeological_site'][zoom >= 16] {
     marker-file: url('symbols/archaeological_site.svg');
     marker-fill: @culture;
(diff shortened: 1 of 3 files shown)


amenity-points.mss
```diff
@@ -590,13 +590,20 @@
...
+  }
+
   [feature = 'historic_archaeological_site'][zoom >= 16] {
     marker-file: url('symbols/archaeological_site.svg');
     marker-fill: @culture;
     marker-placement: interior;
     marker-clip: false;
   }
-
+
```
matthijsmelissen: Whitespace

Issue #1697, opened 2015-07-28
Start showing some craft and office POIs


kocio-pl:
I think we should show not just shops and amenities, but also some objects with [craft=*](http://taginfo.openstreetmap.org/keys/craft#values) and [office=*](http://taginfo.openstreetmap.org/keys/office#values) key. Case by case analysis is probably needed to know which we want to show and how - sometimes the same icon and/or color may be used as with already rendered POIs, but when it doesn't match, maybe generic black/brown/violet dot and label would work.

Looking at the most popular POIs from these categories:

* craft (above 1k of uses):
 * carpenter - icon needed
 * shoemaker - brown icon, probably not the same as shop=shoe
 * photographer - probably the same as shop=photo
 * electrician - brown flash icon
 * brewery - icon needed
 * tailor - icon needed, probably the same as shop=tailor
 * plumber - icon needed
 * metal_construction - icon needed
 * confectionery - maybe the same as shop=confectionery, but brown or black?
 * gardener - maybe the same as shop=garden_centre, but brown?

* office (above 5k uses):
 * company - maybe generic black dot from z>=19?
 * government - brown generic dot or [government icon](https://github.com/nebulon42/osmic/blob/master/administration/government-14.svg)
 * estate_agent - icon needed, violet as shops
 * yes - generic black dot (with name)
 * insurance - icon needed
 * administrative - brown generic dot or town hall icon

math1985:
See also #108.

pnorman:
Given we already have too many different icons, we need to be careful about this.

We also need consider where we're going cartographically. For a general purpose map, I would be against rendering office=yes.

polarbearing:
amenity-brown is mostly dedicated towards touristic/cultural features (see also #1624), it should not be expanded to commercial activities as in crafts.

kocio-pl:
@polarbearing This interpretation is too narrow - what about post office, town hall, police station, fire station or even courthouse? Amenity is all about different services, while shop is all about buying goods. That was my conclusion [when thinking about where the bank/atm belong](https://github.com/gravitystorm/openstreetmap-carto/pull/1658#issue-95012587) and now it makes a lot of sense for me.

@pnorman You're right, after leaving safe shops/amenities area we're on *terra incognita* and we have to watch our steps carefully. I am not against rendering office=yes, but I definitely want to take some precautions here (which I didn't mentioned just to keep the list compact):
- I would do it at the lowest possible level (z>=19)
[...]

PR #3061, opened 2018-02-10
Render offices as dots + names

### LINKED ISSUE #271: Add rendering for office=government or office=administrative
It's quite common that big buildings used by the government are having several administrative department. It's not really good to tag all the departments within the name of the complete building like "name=department one and department two ...".

It would be nice if office=government and office=administrative would be rendered. In this case multiple points within one building could be tagged like this and the result would be rendered.

### LINKED ISSUE #1922: Render office=lawyer
Currently law offices aren't rendered at all, which is a problem in areas such as [this one](https://www.openstreetmap.org/#map=19/26.64414/-81.86889) in which an entire side of the block is law offices. They are useful and I see no reason not to render them. As for how they should be rendered, I suggest using the amenity pink styling with a [lawyer scale](https://2012thetrial.files.wordpress.com/2012/03/law-scale_jpg2.jpg) icon. Even though they are really tagged as offices, they are more comparable to medical offices like shop=optician or amenity=dentist.

### LINKED ISSUE #2570: Icon for office=newspaper
Hi!
office=newspaper is currently not rendered in the map. Can this be changed?

http://wiki.openstreetmap.org/wiki/Tag:office%3Dnewspaper
Icon: http://osm-icons.org/wiki/File:25040.svg
https://taginfo.openstreetmap.org/tags/office=newspaper

andrzej-r:
RE #108 
Fixes #271
Fixes #1922
Fixes #2570

I've added rendering of office tags (all except office=yes), using a simple marker and text from the name tag, both in amenity-brown colour.

There is no 'office' column in the table, so I had to use tags->'office' instead (advised by folks from #osm-dev, apparently to add this column to the table the whole database would have to be reimported, which is not an option).

This is the first time I touch database queries, so please review this part carefully.

Examples before and after:
office2_l17
office2_l18
office2_l19


kocio-pl:
Thanks, that looks sane to me, however my idea was to use black, violet and brown dots, not just brown - see https://github.com/gravitystorm/openstreetmap-carto/issues/1697#issue-97810252. What do you think about it?

How did you create that list? There are not all the office values and it doesn't look like just top values from Taginfo.

I will test if the rendering is not too early, but I suspect for example z19+ would be better than z17+ in well mapped urban areas.

It also resolves #1922.

Tomasz-W:
What about dark-blue dot for office=* ? This colour reminds me suits or paper folders with documents.

andrzej-r:
My goal is to first fix an immediate issue of office= labels not showing up. This discourages people from mapping POIs and causes building names to be overused as e.g. company names.

Changing default colours etc is of course possible, but let's not go overboard with trying to fix visual aspects all at once. There is a lot of work on classification, icons etc needed, which can be spread across multiple PRs on case-by-case basis. For example, some office= values (e.g. office=government, political_party,tax,educational_institution) do not belong to a "business" category, so they can be moved to other categories, if they suit them better.

I like the current zoom levels because they match the rendering of regular shop= and amenity= tags (dot at >=17, dot + name (if fits) at >=18). I don't mind deferring showing names to >=19 but it should done for all amenity tags in a single separate change.

The list of tags was taken from a wiki docs for the office= tag (https://wiki.openstreetmap.org/wiki/Key:office). Indeed, if other values are already in use (BTW, what is "taginfo"?) we should add them to the list (and Wiki!).

In testing, I haven't spotted other unmapped office tag values, but that's not necessarily easy to do. If anything, I was surprised how many office tags have suddenly showed up in Cambridge, UK. Much more that I expected (perhaps that's because both Maps.me and OsmAnd are already displaying them).

kocio-pl:
> The list of tags was taken from a wiki docs for the office= tag (https://wiki.openstreetmap.org/wiki/Key:office). Indeed, if other values are already in use (BTW, what is "taginfo"?) we should add them to the list (and Wiki!).

Taginfo is an OSM service counting uses of tags - the numbers in the wiki infobox are embedded from it, for example. You can read more about it here:

https://wiki.openstreetmap.org/wiki/Taginfo

The list on the wiki is a good place to start and I agree that we should stick to it and add new types only when they are defined there. I want to look closer to see if we could tune this code a bit.

> I was surprised how many office tags have suddenly showed up in Cambridge, UK. Much more that I expected (perhaps that's because both Maps.me and OsmAnd are already displaying them).

That's exactly why I'm so cautious with z17, which is crowded in some places and we try to offload it when possible. 

> I like the current zoom levels because they match the rendering of regular shop= and amenity= tags (dot at >=17, dot + name (if fits) at >=18). I don't mind deferring showing names to >=19 but it should done for all amenity tags in a single separate change.

With so many different kind of data it's not possible to match everything. I also think that most of the shops and amenities are about direct interaction (buying things or using services), while most of the offices are not, hence z18 might be used for selected ones and z19 for the rest of them.

andrzej-r:
I've added more categories (from taginfo and iD) and broken them down into categories (office, shops, public) to facilitate difference in rendering (currently they only differ in colours). No icons so far.

I've increased zoom levels for some subcategories (so, for example, offices are rendered 1 zoom level later than surrounding shops, which IMHO is wrong) but I object moving offices to z19 without a matching change to other shops and (importantly) building names. Building names are currently rendered from z16-17, which creates incentive for mapping companies as building names. For this reason all single-occupant office buildings in my area use a company name, even if a valid building name exists. 

Should we increase the zoom level (which I agree is a good idea), this should be done for many categories of buildings and POIs in a single changeset, which is outside the scope of this change.

By the way, is there any plan of increasing maximum zoom level in the osm map beyond 19?

kocio-pl:
Great, I will try to test it soon.

> By the way, is there any plan of increasing maximum zoom level in the osm map beyond 19?

I haven't heard of it and I don't expect it, since it would require more than double size of current disc storage on OSMF servers:

https://wiki.openstreetmap.org/wiki/Tile_disk_usage

In theory we're independent project, but in practice OSM.org website is our default deployment target. However I recently approved the code which has additional properties starting from z20 (but it's still usable with z-max=z19) - see https://github.com/gravitystorm/openstreetmap-carto/pull/2936#issuecomment-359133041.

SomeoneElseOSM:
With regard to the aside about tile disk usage, here's a bit of anecdata for a tile server that I look after that goes up to 24.  Sizes are from "du" per zoom level.

772032	16
762888	17
733576	15
617144	18
585020	14
346940	13
300916	20
144968	12
71240	19
55588	11
53420	21
37864	10
21264	22
14460	9
9276	23
6640	8
4980	24

This is heavily dependent on how people use the server - if people were to start scraping zoom 24 for example, the disk usage for that zoom level would go up hugely.

kocio-pl:
Looking at some random places in my city, this rendering is better than I expected - it's not too crowded and mostly sane. 

However when the office takes the whole building, it looks weird with a dot in the middle and the name in different colors. I guess it's better to skip offices with `is_building=yes` in this PR, as it's a different, quite general problem (the name on the building can relate to different tags).

andrzej-r:
I haven't noticed that. Can you point me to a specific location so I can try it myself?

There may be quite a few corner cases like this, so it would be good to test this change with on a larger scale.

kocio-pl:
Sure, that's why independent testing before merging is a good thing. 😄 

Look here for example:
https://www.openstreetmap.org/way/196558064

I also plan to review categories soon.

andrzej-r:
I can reproduce this behaviour (see the screenshot below) but I am unsure if that is something we should be fixing. It is consistent with how shops and other amenity points are currently implemented (e.g. what happens when a building has both name and shop tags?)

I would say it is in fact a desirable behaviour, and it only looks weird because we are used to all names being rendered as building names. Objectively, the name in @kocio-pl's example is a company name, not a building name. 

My suggestion is to leave it as is, and let mappers decide if this is indeed the result they wanted. Perhaps the name tag contains a building name, not a business name, in which case a better solution would be to remove the office tag and use building=commercial instead. 

office2_l19_single

Tomasz-W:
I think black color should be used for govermental offices, and for commercial offices dark-blue would be better. Can you test it?

andrzej-r:
There is a bit of a problem with government-like offices. All government offices are currently using `public-service`=`amenity-brown`=`#734a08`. I'm not opposed to switching to `black`=`#000000` but that would affect many existing objects. BTW, at the moment the only objects with blackish labels (visible in the screenshot) are `place-of-worship`, which are hardcoded to `#000033`.

My suggestion: leave government offices in `public-service` category for now. We can change it later on case by case basis, e.g. when we get suitable icons. The current list of government/public offices is:

```
    [feature = 'office_administrative'],
    [feature = 'office_adoption_agency'],
    [feature = 'office_government'],
    [feature = 'office_employment_agency'],
    [feature = 'office_educational_institution'],
    [feature = 'office_government'],
    [feature = 'office_ngo'],
    [feature = 'office_political_party'],
    [feature = 'office_quango'],
    [feature = 'office_religion'],
    [feature = 'office_tax']
```

This is how the above example looks like with `office`=`darkblue`=`#00008b`. I quite like it.

office2_l19_darkblue

kocio-pl:
Before I check the groups - I think I can accept the dot for the whole building, but the space between the dot and the label is just too big. It's also the problem with shop dots, but it's consistent with other shops at least. Here the dot is even smaller than with shops, so the distance should be smaller too. 

What bothers me is that the dot takes my attention to the center and then I can't immediately recognize the label position.

andrzej-r:
Would making the dot bigger be an option? That would match rendering of shops. The screenshot shows a shop and a shop-like office ("Plus"), both rendered slightly differently.

The spacing probably asssumes the dot may actually be an icon. I will check if it can be tighten without affecting icon-label spacing.

kocio-pl:
I like the idea to have smaller dots for offices. We may use the icons for some of them, so they can be recognized by their own shape, but this distinction for the rest of them is nice.

kocio-pl:
Thanks, now it's better for me.

daganzdaanda:
Could you test an denser area with more offices? 
Also, does your code render office nodes like shops? See https://www.openstreetmap.org/node/4413514591

kocio-pl:
An example place with more offices (private and public): 
https://overpass-turbo.eu/s/wmQ

andrzej-r:
Thanks for the overpass link. I've checked several locations and it looks like London is particularly rich in office points and buildings:
office2_l18_london1
office2_l18_london2
office2_l18_london3

kocio-pl:
Still reasonable for me - that's very promising!

andrzej-r:
Over the next couple of weeks starting this Sunday I'll have a limited access to the Internet. Is there anything you want me to try before that?

How does the deployment procedure look like for the osm map? I read they pick up a new version of Carto every first Sunday of the month. Is that a released version (if so, when is the next release?) or a git master?

kocio-pl:
I still didn't have the time to examine categories, so I have nothing more to ask you about currently.

> How does the deployment procedure look like for the osm map? I read they pick up a new version of Carto every first Sunday of the month. Is that a released version (if so, when is the next release?) or a git master?

No, OSMF server admins run z0-z12 render updating then. It has nothing to do with style change, these levels are just not constantly updated (as z13-z19 are). Another occasion is when a new release is deployed, of course.

Deployment on OSMF servers is done after we do new style version release and let them know about it (see https://github.com/gravitystorm/openstreetmap-carto/blob/master/RELEASES.md#notifications ). I try to keep our regular - major and minor - releases once a month, especially on Fridays (because re-rendering tiles on OSMF servers may start in the weekend, when the load is lower), but it's just a matter of what I think is proper. Other team members can do it whenever they feel and bugfix releases are done when it's needed and possible.

I plan [next release](https://github.com/gravitystorm/openstreetmap-carto/issues/3063) for tomorrow evening and hope that admins will deploy it soon after, but that's entirely their decision.

andrzej-r:
Ok. Tomorrow is probably too soon for this pr but if you wanted to try it I can help with some testing and changes. Otherwise we can come back to it in two weeks time. 

I don't think we need to nail everything down in this pr, the intention was to make office points visible at all with aesthetic changes and finer categories to follow. The current implementation is just a good start (I think).

kocio-pl:
No need to hurry. I just want to look at it closer, since I've been busy with other things and didn't test it as much as I'd like.

daganzdaanda:
Thanks for the rendering examples! 
They look good to me, the only thing I'm not 100% sure about is introducing a new colour - but your choice of dark blue seems to fit quite OK.

andrzej-r:
I was responding to a suggestion (which I liked, I admit). I was initially reusing "amenity-brown" but since it is a new category of objects a new colour may be justified.

Another thing to look at in detail is SQL queries and their possible impact on performance. As of now - they work (in my tests, at least), but that's all - I am not an expert in SQL and OSM toolchain.

andrzej-r:
Some government offices. One thing I don't particularly like is that fairly prominent building names are now replaced with much less prominent office labels. This is especially visible for buildings with office= tag in them. It will also discourage people from mapping offices as points (rather than naming building after them as it is common now). Perhaps it is building names that should be demoted as they are currently rendered from z16.
office2_l18_london4

Tomasz-W:
I think gov offices should be rendered black, and then they will mach to dark-blue commercial ones.

andrzej-r:
Only office=government or also others from the list I posted above? Currently they are all rendered in public-service colour, which is also used by some of the existing existing PoIs. Just changing the public-service colour wouldn't work well. 

I think we could split the list into public services and government. Can you suggest how to divide them?

Tomasz-W:
Black:
*    [feature = 'office_administrative'], -> this tag was moved to office=government by Wiki   https://wiki.openstreetmap.org/wiki/Tag:office=administrative?uselang=en
*    [feature = 'office_government'],
*    [feature = 'office_tax']

Dark-blue:
*    [feature = 'office_adoption_agency'],
*    [feature = 'office_employment_agency'],
*    [feature = 'office_educational_institution'],
*    [feature = 'office_ngo'],
*    [feature = 'office_political_party'],
*    [feature = 'office_quango'],
*    [feature = 'office_religion'],

andrzej-r:
OK, that's roughly how I wanted to handle it. I would keep adoption and employment agencies in public services, though. Do you agree? 

Can you comment on other categories? I have lumped certain offices with shops (as far as colour is concerned). I also divide offices into two office categories depending on their size and importance, both rendered in dark blue but using different filtering options. For example, the ones on your list I would classify as more important.

Tomasz-W:
> OK, that's roughly how I wanted to handle it. I would keep adoption and employment agencies in public services, though. Do you agree?

I think making more exceptions (than gov offices) here is not nessesary, and we should categorize all tags as black or dark-blue. You mentioned 'adoption' as candidate for amenity-brown - well, it's not a place where you can ask for a child and get him/ her, but still office. The same with employment agencies - you go there and give your CV or present yourself, not ask for a job and simply get it ;)

amenity-points.mss
```diff
@@ -2051,14 +2136,102 @@
...
+
   [feature = 'shop_supermarket'],
   [feature = 'shop_department_store'] {
     [zoom >= 16] {
       text-name: "[name]";
       text-size: @standard-font-size;
       text-wrap-width: @standard-wrap-width;
       text-line-spacing: @standard-line-spacing-size;
-      text-dy: 12;
+      text-dy: 8;
```
kocio-pl: This change is outside the scope of the `office=*` rendering and makes the change with shops, which can disable rendering some labels, especially with high glyphs (like "É"), so should be probably reverted.

dieterdreist:
2018-02-23 11:06 GMT+01:00 Tomasz Wójcik <notifications@github.com>:

> Black:
>
>    - [feature = 'office_administrative'], -> this tag was moved to
>    office=government by Wiki https://wiki.openstreetmap.
>    org/wiki/Tag:office=administrative?uselang=en
>    <https://wiki.openstreetmap.org/wiki/Tag:office=administrative?uselang=en>
>    - [feature = 'office_government'],
>    - [feature = 'office_tax']
>
> Dark-blue:
>
>    - [feature = 'office_adoption_agency'],
>    - [feature = 'office_employment_agency'],
>    - [feature = 'office_educational_institution'],
>    - [feature = 'office_ngo'],
>    - [feature = 'office_political_party'],
>    - [feature = 'office_quango'],
>    - [feature = 'office_religion'],
>
>

can you explain what is the governing principle? Isn't this suggesting a
certain kind of political structure and division of roles (e.g. laicism),
which isn't applicable on a global level?

kocio-pl:
The example where the name on the building [is rendered twice on z17](https://www.openstreetmap.org/way/105556113#map=17/51.51458/-0.09310)- as a building and an office (and [another one](https://www.openstreetmap.org/way/33104133#map=17/51.51555/-0.09430)); what do we want to do with it?

xbear88m
4rdsyz9k

I think dark blue is a good choice as a new color - [it's similar to how we render railway/metro stations](https://www.openstreetmap.org/way/172649358#map=18/51.51505/-0.09868), but we have no choice but reuse colors for different types of objects:
w0jlf8sc

kocio-pl:
> I will fix that spurious text-book change - it wasn't intended. Can't do that before in the next 2 weeks, though.

OK, it's a totally new set of tags, so it's better to test more than usual - no hurry.

> Not sure what you meant by the colour. Is it OK to use it or should we reuse an existing colour?

I think it makes sense, I just wanted to show that it's a reuse.

andrzej-r:
I think (I cannot test it right now) both errors are caused by a mismatch in point and label firtering options. We should either add `[zoom >= 17][way_pixels > 3000]` to the point options (lines 918, 946) or remove it from label options (lines 2159, 2213). Or simply render all office points from z17.

andrzej-r:
1. I've reverted the accidental text-dy change
2. I've fixed double labels showing up in some office buildings at z17. It was indeed caused by mismatch in filtering options between labels and points:
office2_l17_london5
Note the rendering error (a building label partially showing up). Not sure if that's a style issue or if my kosmtik is acting up. Kosmtik shows the same rendering issue in [existing structures](https://www.openstreetmap.org/way/26183027), which are rendered correctly on the website:
office2_l17_london6
I would appreciate if someone could test it independently. 
office2_l17_london7
3. Matched sizes of office icons and other POI icons at zoom>=18:
office2_l18_london7

andrzej-r:
Can I have this PR deployed in the test server, please? It seems to work OK in my setup but we need to look at broader selection of points and check if there are no issues with database performance.

kocio-pl:
You need to resolve the  `project.mml` conflicts before that.

andrzej-r:
I've resolved conflicts. Automated merging should work fine now but since I'm not familiar with other changes in master I haven't tested the result.

kocio-pl:
Thanks. I don't know what other changes in master have to do with your PR, but I'm not aware of any problem witch could make testing hard or unreliable, so you can do it.

@rrzefox, could you deploy this code on your server to test it worldwide?

matthijsmelissen:
Do we really need to test everything on a worldwide server? This seems particularly lime a PR that can be tested just as well with local extracts.

Of course if @rrzefox wants to help that is really great, but I don’t want to aak too much from him either.

kocio-pl:
This is not typical "add one" code, so I'd like to see it in different places with a fresh eye without planning. I believe just asking is OK, since I have no idea how much would be "too much" for him.

andrzej-r:
I initially thought the same but then there were some unexpected issues found in some combinations of tags. Some independent testing would be useful. 

Having said that, if there are no big issues found I'd advise to merge this pr asap and work on resolving smaller issues from there. The master branch has already started to diverge.

Adamant36:
I can probably test it in a couple of dense areas in California if someone wants to provide the code or the file with the code in there already.

kocio-pl:
The code branch to be merged is here: https://github.com/andrzej-r/openstreetmap-carto/tree/office2

andrzej-r:
Can anyone comment on the visual issue I mentioned in https://github.com/gravitystorm/openstreetmap-carto/pull/3061#issuecomment-372117565, point 2? I think (because I could see it also in existing objects rendered by kosmtic but not by the website) it is caused by my test setup.

Other than that I don't see any issues myself but of course a fresh pair would likely spot something.

Also, I would advise someone to go through the code and review it. Especially the part related to the database access.

kocio-pl:
The problem of double labels on office buildings is gone now:
4x48c5o8

andrzej-r:
Do you also get that clipped "Co" label at z17?

kocio-pl:
Yes, but only in the Kosmtik window:
screenshot-2018-3-19 openstreetmap carto kosmtik

export rendering is clean:
zhi5i4cz

andrzej-r:
Looks like #2671 has created new conflicts. I will leave this PR as is to avoid needless code churn. Ping me when you are ready to merge it - I'll create a new PR against a fresh master.

andrzej-r:
Closing this PR as it is now superseded by #3163.

PR #3065, opened 2018-02-11
Give oceans outline and simplify shapefiles on z0-7

matthijsmelissen:
Before:
screen shot 2018-02-11 at 20 55 37

After:
screen shot 2018-02-11 at 21 00 37

Still needs testing on more realistic data.

### DIFF
--- a/shapefiles.mss
+++ b/shapefiles.mss
@@ -14,6 +14,13 @@
 #world {
   [zoom >= 0][zoom < 10] {
     polygon-fill: @land-color;
+    polygon-simplify: 5;
+    [zoom < 8] {
+      line-color: darken(@water-color,20%);
+      line-simplify: 5;
+      line-width: 0.5;
+      line-offset: 0.5;
+    }
   }
 }
 
--- a/shapefiles.mss
+++ b/shapefiles.mss
@@ -14,10 +14,10 @@
 #world {
   [zoom >= 0][zoom < 10] {
     polygon-fill: @land-color;
-    polygon-simplify: 5;
+    polygon-simplify: 0.4;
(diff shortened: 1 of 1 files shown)


kocio-pl:
Looks nice for me.

rrzefox:
deployed in the usual location for global testing. Looks good IMHO in the low zoom levels. Looks a bit weird on Z7 where it overlaps with dashed borders, e.g. http://bl.ocks.org/matthijsmelissen/raw/af7a602c222dbf1ff1a2c0d84ed755b7/#7.00/37.519/16.197

matthijsmelissen:
I reduced the simplification to prevent artefacts at the antemeridian (visible in the demo above).

Rendering demo:
screen shot 2018-03-05 at 00 03 09

As far as I'm concerned this is ready to be merged.

matthijsmelissen:
@kocio-pl What do you think?

kocio-pl:
I still like it :smile: however I'd like to know if it won't produce performance problems (see also https://github.com/gravitystorm/openstreetmap-carto/pull/3103#issuecomment-370390905)? Do you have the idea how should we estimate it in general?

matthijsmelissen:
I don't expect performance problems. The problems in #1938 where caused by the ST_Within check (we were checking for admin polygons located within big admin_level 2 polygons), which we don't do now - this is a totally different approach.

It seems to run without problems on @rrzefox's server.

kocio-pl:
I think this is a small side effect of this change - when the river mouth is big you can see this outline on the water up to z7:

screenshot-2018-3-24 openstreetmap
screenshot-2018-3-24 openstreetmap 1

PR #3066, opened 2018-02-11
Support for amenity=public_bath

### LINKED ISSUE #2856: Please display amenity=public_bath on the map

MKuranowski:
As discussed in #2856, this PR adds an icon for [amenity=public_bath](https://wiki.openstreetmap.org/wiki/Tag:amenity%3Dpublic_bath).

Icon itself:
icon

And example rendering:
rendering

Thus, this resolves #2856.


kocio-pl:
If you turn the related issue into a line like `Resolves #2856`, the issue will be automatically closed upon merging this code.

amenity-points.mss
```diff
@@ -186,6 +186,13 @@
     marker-clip: false;
   }
 
+  [feature = 'amenity_public_bath'][zoom >= 17] {
+    marker-file: url('symbols/public_bath.svg');
+    marker-fill: @culture;
```
kocio-pl: I would rather make it `@amenity-brown`. I wouldn't put it in the same basket as cinema for example, so general category is safe. If we have sauna and shower icons (which is perfectly possible), we can make a separate category.

amenity-points.mss
```diff
@@ -1231,6 +1238,19 @@
     }
   }
 
+  [feature = 'amenity_public_bath'][zoom >= 17] {
+    text-name: "[name]";
+    text-size: @standard-font-size;
+    text-wrap-width: @standard-wrap-width;
+    text-line-spacing: @standard-line-spacing-size;
+    text-fill: @culture;
```
kocio-pl: The same as above - `@amenity-brown` is better.

polarbearing:
Sorry for being late - but there was one suggestion with two people, adopted from playground.svg
That would support the idea of publicness better IMHO.

kocio-pl:
Thanks for testing, this was [my idea](https://github.com/gravitystorm/openstreetmap-carto/issues/2856#issuecomment-331968555) 😄, but this icon is more clear to me. I also don't think we need to emphasize that this is public facility, as most of the things we show are public (and if they're not, we de-emphasize them or just don't show them at all).

polarbearing:
Happy enough with the lone bathling as well.

kocio-pl:
In my rendering the name is displayed, but not the icon, even if there's a room for that, could you check what kind of problem that could be?

sj3roi5e

https://www.openstreetmap.org/node/5230375425#map=19/52.23613/21.00116
https://www.openstreetmap.org/node/4881681777#map=19/49.43988/19.85417

kocio-pl:
Now it works in both cases. Thanks!

PR #3072, opened 2018-02-15
Make military=danger_area font dark pink and slanted

kocio-pl:
Follow up to #3057.

Small fix to make label rendering of [danger area without landuse=military](https://www.openstreetmap.org/way/173480188) still easy to recognize (hence dark pink again) and compatible with area labels rule (hence slanted):

z11
Before
aez5ysl

Current master
kkj6e 6 

After
iqdw2do_

### DIFF
--- a/amenity-points.mss
+++ b/amenity-points.mss
@@ -1580,8 +1580,8 @@
         text-wrap-width: @landcover-wrap-width-size-bigger;
         text-line-spacing: @landcover-line-spacing-size-bigger;
       }
-      text-fill: darken(@military, 40%);
-      text-face-name: @bold-fonts;
+      text-fill: darken(@military, 20%);
+      text-face-name: @landcover-face-name;
       text-halo-radius: @standard-halo-radius;
       text-halo-fill: @standard-halo-fill;
       text-placement: interior;


andrzej-r:
Yes, I just recalled we forgot to add the colour change in #3057. No problem with this change. 

BTW, slanted fonts are also (somewhat confusingly) used for marking motorway exits.

kocio-pl:
I was asking you about adding this exact change, but maybe we misunderstood each other. 😄 

You should ask @matkoniecz about motorways.

matthijsmelissen:
I think the proposed label color is visually not similar enough to the color of the lines.

kocio-pl:
[Example](https://www.openstreetmap.org/way/544654993#map=19/50.96381/15.64945) at z23 
(it's very small area in the forest, but I wanted to show how it would be with bigger area)
xhjntf4z

Under Gimp I see that new has the base color ff5555 (made lighter, of course) and the text after this change is f2143a, so not that far. But if we change it exactly to ff5555, it will look like this:
n_2yebol
v4sen3gn

Which is similar, but paler, so it might be just a tinting effect and I would rather stay with f2143a (dark pink) to be more visible. Any suggestion if/how to change it?

kocio-pl:
Another [example](https://www.openstreetmap.org/#map=12/51.4388/14.8008) - colors comparison with military area and standard land color in the background on z12:
ff5555
snxoprmo
f2143a
jocwo4th

kocio-pl:
I have changed the notation, so it's more clear that this base on the same color, but the output is similar to original proposition. Any comments/reviews about it?

[Example](https://www.openstreetmap.org/way/544654993#map=19/50.96381/15.64945) at z23 

Original proposition
xhjntf4z

20% darker military (instead of 40%)
ofdatt0_

andrzej-r:
I'm OK with both, with slight preference for the darkened military. 

20% darker military looks like a good tradeoff between making the label stand out and, at the same time, fit the colour pattern.

PR #3075, opened 2018-02-16
Adding rendering for amenity=police and amenity=fire_station areas

### LINKED ISSUE #848: Add rendering for amenity=police area

kocio-pl:
Related to #3045.
Resolves #848.

Rendering police and fire station areas with the same hatching as new military area style, as discussed before. It will make uniformed services areas to use similar look.

Examples:
[Police](https://www.openstreetmap.org/way/556906228#map=18/52.24606/20.99851)
fyhfx56t

[Fire services](https://www.openstreetmap.org/way/35017939#map=19/52.18410/21.00388)
acj3finn

[Comparison with military area rendering](https://www.openstreetmap.org/#map=18/52.21053/21.01156)
wua6dgpj

### DIFF
--- a/landcover.mss
+++ b/landcover.mss
@@ -543,6 +543,21 @@
     }
   }
 
+  [feature = 'amenity_fire_station'][zoom >= 8][way_pixels > 900],
+  [feature = 'amenity_police'][zoom >= 8][way_pixels > 900],
+  [feature = 'amenity_fire_station'][zoom >= 13],
+  [feature = 'amenity_police'][zoom >= 13] {
+    polygon-fill: #F3E3DD;
+    line-color: @military; 
+    line-opacity: 0.24;
+    line-width: 1.0; 
+    line-offset: -0.5;
+    [zoom >= 15] {
+      line-width: 2; 
+      line-offset: -1.0;
+    }
+  }
+
   [feature = 'amenity_parking'][zoom >= 10],
   [feature = 'amenity_bicycle_parking'][zoom >= 10],
   [feature = 'amenity_motorcycle_parking'][zoom >= 10],
--- a/project.mml
+++ b/project.mml
@@ -119,8 +119,8 @@ Layer:
               ('aeroway_' || (CASE WHEN aeroway IN ('apron', 'aerodrome') THEN aeroway ELSE NULL END)) AS aeroway,
               ('amenity_' || (CASE WHEN amenity IN ('parking', 'bicycle_parking', 'motorcycle_parking', 'university', 'college', 'school', 'taxi',
                                                     'hospital', 'kindergarten', 'grave_yard', 'prison', 'place_of_worship', 'clinic', 'ferry_terminal',
-                                                    'marketplace', 'community_centre', 'social_facility', 'arts_centre', 'parking_space', 'bus_station')
-                                                    THEN amenity ELSE NULL END)) AS amenity,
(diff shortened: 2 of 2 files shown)


matthijsmelissen:
I'm not sure about the diagonal dashing, it suggests access is forbidden, while that is not true for many police/fire station grounds.

kocio-pl:
Do you think that just changing to similar overlay file without hatching would be OK, or prefer standard light pink background with the same outline?

polarbearing:
I thought we intended to reduce the hatching?
My preference would be the same outline as the new military (need to add that to #3045 for comparison), but no overlay and no hatching. 
Thus we follow the concept of #3045 with outlines for campuses, and use the same colour for uniformed services.

kocio-pl:
Updated code:
htanf7h8
vcdbadh7
98fpwhln

polarbearing:
How does the low-zoom version look like where you have line-width=1.0 ?
Could we compare it with a no-fill version?

kocio-pl:
What do you mean with "low-zoom version"?

I can test it, but I wouldn't like to diverge too far from new military rendering (that's why I started with hatching).

polarbearing:
I mean [zoom >= 13] line-width: 1.0 in the code. I guess your examples are [zoom >= 15] line-width: 2.

As for the fill, you are currently proposing a new colour for the uniformed-amenities. In [3045-comment](3045#issuecomment-367446137) @Tomasz-W argues that these campuses are more and more filled with landcovers, so we might focus on outlines-only anyway. For the touristic accommodation, for example, I would not want light-blue fill in addition to the outline.

Tomasz-W:
Resolves #848

kocio-pl:
I don't know if I will find that big police/fire area - the border is however the same as for the current military areas on the same zoom.

The color definition is new, but it's just how the lighter parts of military areas look like on the ground color, because I think it's the most generic color for military areas. I will try to make the rendering with only the outline to compare pretty soon.

polarbearing:
This is a fairly large one:
https://www.openstreetmap.org/way/5095457

kocio-pl:
@polarbearing Sorry that I haven't found the time to check your proposition, but this way we would loose uniformity of such areas - lack of hatching is enough for me.

polarbearing:
If the outline behaviour is copied from the military, I'd be happy enough. Not time to test, either. Will have a look when deployed.

Tomasz-W:
@kocio-pl 
An red outline is missing here: https://www.openstreetmap.org/way/513298808
Is it some bug or what?

kocio-pl:
It's visible in both cases, but only near gates. Probably grass and buildings are eclipsing the outline.

Tomasz-W:
Don't you think that it should work like this: barriers > outlines > landcover? There will be more and more cases like in examples above, because the map gets more and more detailed, When outlines are covered by landcover, we lose information about area range/ boundaries, and they become useless.

kocio-pl:
Yes, I think it should. Anybody willing to play with the code and check whether it wouldn't break anything else?

polarbearing:
Sounds plausible. Should be done for the tourism outline as well. 
Earlier this year we had lowered the tourism outline (next to the military) to show the barriers (fences). Would be good to have that sorted before we add further tourism (museum, accommodation) outlines, see #3045 
Please note we had a similar discussion there, regarding highways. I still think highways should remain above the outline.

PR #2887, opened 2017-10-08
Fix SQL problem with some PostGre versions

sommerluk:
Fixes #2886 and I suppose addition parenthesis won’t make any harm on other PostGre versions or setups.

Issue #2886, opened 2017-10-08
Strange runtime error…


sommerluk:
Strange runtime error…

```
Trace
    at ProjectServer.raise (/home/sommerluk/Dokumente/OSM/kosmtik/src/back/ProjectServer.js:257:13)
    at /home/sommerluk/Dokumente/OSM/kosmtik/src/back/ProjectServer.js:71:30
    at /home/sommerluk/Dokumente/OSM/kosmtik/node_modules/mapnik-pool/node_modules/generic-pool/lib/generic-pool.js:283:11
    at loaded (/home/sommerluk/Dokumente/OSM/kosmtik/node_modules/mapnik-pool/index.js:23:37)
Postgis Plugin: FEHLER:  Operator existiert nicht: hstore -> boolean
LINE 8: ...T NULL) OR ("addr:housename" IS NOT NULL) OR (tags->'addr:un...
                                                             ^
HINT:  Kein Operator stimmt mit dem angegebenen Namen und den Argumenttypen überein. Sie müssen möglicherweise ausdrückliche Typumwandlungen hinzufügen.
in executeQuery Full sql was: 'SELECT * FROM (SELECT
    way,
    "addr:housenumber" AS addr_housenumber,
    "addr:housename" AS addr_housename,
    tags->'addr:unit' AS addr_unit,
    way_area/NULLIF(0::real*0::real,0) AS way_pixels
  FROM planet_osm_polygon
  WHERE (("addr:housenumber" IS NOT NULL) OR ("addr:housename" IS NOT NULL) OR (tags->'addr:unit' IS NOT NULL))
    AND building IS NOT NULL
UNION ALL
SELECT
    way,
    "addr:housenumber" AS addr_housenumber,
    "addr:housename" AS addr_housename,
    tags->'addr:unit' AS addr_unit,
    NULL AS way_pixels
  FROM planet_osm_point
  WHERE ("addr:housenumber" IS NOT NULL) OR ("addr:housename" IS NOT NULL) OR (tags->'addr:unit' IS NOT NULL)
  ORDER BY way_pixels DESC NULLS LAST
) AS addresses LIMIT 0'
  encountered during parsing of layer 'addresses' in Layer
[httpserver] /openstreetmap-carto-with-data/tile/17/82549/56277.png?t=1507477774906 500
[httpserver] /favicon.ico 200
[httpserver] /openstreetmap-carto-with-data/poll/ 200
```

[...]

PR #3077, opened 2018-02-20
Explicitly state that PostgreSQL at least 9.5 is required

Zverik:
SQL queries fail on PostgreSQL 9.3 with `operator does not exist: hstore -> boolean`. Turns out 9.5 is required. Let's state this in the documentation explicitly, for people upgrading their styles from earlier versions.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -110,7 +110,7 @@ For development, a style design studio is needed.
 
 To display any map a database containing OpenStreetMap data and some utilities are required
 
-* [PostgreSQL](http://www.postgresql.org/)
+* [PostgreSQL](http://www.postgresql.org/) 9.5+
 * [PostGIS](http://postgis.org/)
 * [osm2pgsql](https://github.com/openstreetmap/osm2pgsql#installing) to [import your data](https://switch2osm.org/loading-osm-data/) into a PostGIS database
 * `curl` and `unzip` for downloading and decompressing files


kocio-pl:
I think we should test it with Docker images of PostgreSQL 9.3 and 9.4, that should be harmless and pretty straight. Is anybody willing to do it?

pnorman:
> SQL queries fail on PostgreSQL 9.3 with operator does not exist: hstore -> boolean. Turns out 9.5 is required. Let's state this in the documentation explicitly, for people upgrading their styles from earlier versions.

A query needing 9.5 is a bug in the query, see #2886

matthijsmelissen:
A couple of extra brackets in the statement should fix this.

Zverik:
> A query needing 9.5 is a bug in the query, see #2886

But that was fixed in #2887?

Zverik:
Complete error message, in case it really is a bug in SQL:

```
renderd[858]: An error occurred while loading the map layer 'old': Postgis Plugin: ERROR:  operator does not exist: hstore -> boolean
LINE 11:     AND (tags->'public_transport' IS NULL OR tags->'public_t...
                      ^
HINT:  No operator matches the given name and argument type(s). You might need to add explicit type casts.
in executeQuery Full sql was: 'SELECT * FROM (SELECT
    way,
    building
  FROM old_polygon
  WHERE building IS NOT NULL
    AND building != 'no'
    AND (aeroway IS NULL OR aeroway != 'terminal')
    AND (amenity IS NULL OR amenity != 'place_of_worship')
    AND building != 'train_station'
    AND (aerialway IS NULL OR aerialway != 'station')
    AND (tags->'public_transport' IS NULL OR tags->'public_transport' != 'station')
    AND way_area > 0.01*0::real*0::real
) AS buildings LIMIT 0'
  encountered during parsing of layer 'buildings' in Layer of '/home/zverik/styles/openstreetmap-carto/oldosm4.xml'
```

Zverik:
Okay, did some regular expressions myself: #3078

kocio-pl:
Another problem with older database (upgrade to 9.5 fixed the problem):

https://help.openstreetmap.org/questions/63612/issue-starting-render

Zverik:
Fixed in #3245

PR #3078, opened 2018-02-20
Fix operator precedence for hstore queries

Zverik:
Fixes #3077.

See [the precedence table](https://www.postgresql.org/docs/9.3/static/sql-syntax-lexical.html#SQL-PRECEDENCE-TABLE): `IS` is higher than `(any other)`, which includes `->`.


matthijsmelissen:
Thanks!

Probably we should also catch this in Travis.

PR #3082, opened 2018-02-21
Added landcover=grass as grass

pietervdvn:
For the moment, landcover=grass does not render as grass. As this is subtly different then landuse=grass, this is a valuable addition.

For example, [here](https://www.openstreetmap.org/way/311784342) we would like to map the grass within a residential area. By using 'landcover', we can correctly use this within a residential area.

Rendering won't change visibly here:
screenshot from 2018-02-21 15-59-34

### DIFF
--- a/landcover.mss
+++ b/landcover.mss
@@ -327,7 +327,8 @@
   [feature = 'natural_grassland'],
   [feature = 'landuse_grass'],
   [feature = 'landuse_village_green'],
-  [feature = 'leisure_common'] {
+  [feature = 'leisure_common'],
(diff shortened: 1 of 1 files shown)


kocio-pl:
I would like to implement it for some time already, but there is a big debate with different views on the topic, so please look at #2548 first and relate to a discussion.

kocio-pl:
Since the discussion and coming to some conclusions is the basic problem, I will close this PR now. I hope we will find some common ground at last, so adding the code will make sense.

PR #3085, opened 2018-02-23
Change label colour of private parking

### LINKED ISSUE #2170: Label color for private parking is non-intuitive
Especially when the corresponding symbol is not rendered above the yellow parking color - like here:

http://www.openstreetmap.org/#map=19/50.04516/8.54655

Should IMO be a lighter version of the normal label color, not a greener tone.

jragusa:
Resolves #2170.

Replace greenish hardcoded colour to opacity value.

### DIFF
--- a/amenity-points.mss
+++ b/amenity-points.mss
@@ -1306,6 +1306,7 @@
     text-placement: interior;
     [access != ''][access != 'permissive'][access != 'yes'] {
       text-opacity: 0.33;
+      text-halo-radius: 0;
     }
     [feature = 'amenity_bicycle_parking'],
     [feature = 'amenity_motorcycle_parking'] {
--- a/amenity-points.mss
+++ b/amenity-points.mss
@@ -1305,7 +1305,7 @@
     text-halo-fill: @standard-halo-fill;
     text-placement: interior;
     [access != ''][access != 'permissive'][access != 'yes'] {
-      text-fill: #66ccaf;
+      text-opacity: 0.33;
     }
     [feature = 'amenity_bicycle_parking'],
     [feature = 'amenity_motorcycle_parking'] {


kocio-pl:
Thanks! If you change "Related to" into "Resolves", the issue will be closed automatically when merging this code.

jragusa:
Thank you :)

It's done

polarbearing:
I assume it takes the colour from the block above, thus ```text-fill: @transportation-text;```.
It would still be helpful to see a test rendering, to evaluate if it is readable enough. Preferably both on a building and on surface parking.

jragusa:
The _Garage Hélicoïdal_ of Grenoble (https://www.openstreetmap.org/way/32538552)

before (z=19)
before

after (z=19)
after

kocio-pl:
That's what I expected. @polarbearing What's your opinion?

kocio-pl:
@polarbearing ping?

polarbearing:
Well, it looks nice on the parking-grey. 
It is difficult to read on the building. It was difficult before but it gets even worse. 

My explanation is that the brighter typesetting kind of blurs with the halo.

Maybe we could drop the halo when rendered on a building? though that might be difficult to code for the case that the parking is not tagged on the building outline but as a node.

kocio-pl:
@jragusa Could you test this approach?

jragusa:
Current setting / only blue colour / blue colour + no halo

[surface parking](https://www.openstreetmap.org/#map=19/46.18589/6.22559)
surface

[building parking](https://www.openstreetmap.org/#map=19/45.19099/5.72664)
building

I note that the suggestion of @polarbearing is easiest to read and name remains of less importance that shop names. What do you think ?

kocio-pl:
I agree that his proposition works best. @polarbearing If you're happy with how does it look like, I would merge the altered code.

polarbearing:
The zero-halo approach is the best so far. 

For a single node that does not bring its own parking/building background, I'd expect that to work fine on a bright background, such as 'empty' or amenity-yellow. 

I'm wondering what mixture we have to expect on industrial/commercial. 
Do we have any darker backgrounds to consider?

jragusa:
I don't find such association, so examples are welcome to test rendering

polarbearing:
Feed the overpass wizard with 
```amenity=parking and access=private and name=* and type:node in YOURAREA```.

However I just realise that names on parking are only rendered on polygons, not on nodes. I wonder if this is intentional or a bug?

The analysis shows that rendering the marker is split in the code: 
The marker is set in line 370 for those >900waypixel, and in line 1096 in the ```.amenity-low-priority``` section for any. 

At line 1279 the label is applied for polygons>900waypixel. There is no matching text label for the catch-all condition, thus apparently not rendered for nodes.

polarbearing:
It's intentional, it was introduced here #1364.

That means, we currently have no parking on nodes to investigate which might mix with a strange background, and can merge the PR. 

Should we want labels on parking nodes eventually, we would need to reconsider the colour question.

jragusa:
@polarbearing 
I tried and only found parking nodes in residential landuse

kocio-pl:
So - what is the proposed solution now?

jragusa:
It's ok for me unless you want to add names on nodes.

I just have an additional question : why label and maker don't have the same colour ?

kocio-pl:
> It's ok for me unless you want to add names on nodes.

We're just fixing current issue, not adding anything new.

OK, so I'll merge this code now. Thanks for the code and discussion!

> I just have an additional question : why label and maker don't have the same colour ?

I have difficulties trying to find when this line has been introduced:

https://github.com/gravitystorm/openstreetmap-carto/blob/b0787aa03ed2f277655e54c560d376235f02136b/amenity-points.mss#L1311

jragusa:
Sorry, I used wrong expression : 

`@transportation-icon: #0092da` and `@transportation-text: #0066ff` 

Inversely, amenities use colour `@amenity-brown` for both icon and text.

Is there a specific reason to use two different colours for transportation and accommodation ?

kocio-pl:
Oh, I see.

As far as I remember, this is because text is thin and the icon is thick usually, and it makes the illusion that the colors are very different (halo can also trigger this effect). It was designed before I came here, but that was my conclusion. You have to test it however, because it was long time ago when I checked this. I'm also not sure why amenity is different in this regard.

Issue #3086, opened 2018-02-24
Runtime Error with generate_tiles.py


zsoooc:
I installed a tile server based on this tutorial: https://ircama.github.io/osm-carto-tutorials/tile-server-ubuntu/
I strictly followed the instructions and I use Ubuntu 16.04.1 clean install.
The server works, but there is an error when I want to run generate_tiles.py.

"Runtime Error: Failed to find font face 'unifont Medium' in FontSet 'fontset-0' in FontSet at line 17 of '/home/user/src/openstreetmap-carto/style.xml

StyXman:
I have the same problem under Debian Sid. I just remove that font name, as it's just another, probably older, name for "Unifont Medium". I have no idea in which context that name is still valid.

I have to add that this only breaks when using `mapnik` with Python as `generate_tiles.py` does. Old `tilemill` and new `komstik` must be configuring `mapnik` somehow so it only generates a warning. I never took the time to investigate more deeply.

SomeoneElseOSM:
To echo what @StyXman said, both "unifont Medium" and "Unifont Medium" are valid in different Linux distributions.  That's not normally a problem, because if something like renderd is using the style, it doesn't mind references to fonts that aren't installed locally.  Unfortunately, `generate_tiles.py` does object.  

When using it I tend to remove styles that are obvious duplicates (like "Unifont Medium" here) and add any new ones I might want to install (e.g. for language families such as Burmese that I might want to see displayed on a map, but wouldn't normally be installed by default).

StyXman:
I think I have mentioned this before. I have hacked `generate_tiles.py` quite a bit. You can see all the options I added here:

https://github.com/StyXman/elevation/blob/master/generate_tiles.py#L673

Just now @Ircama sent me how to fix this problem. See https://github.com/StyXman/elevation/issues/1 . I'll try to do it during the weekend.

kocio-pl:
This issue is not really related to this style, I guess this is external issue, probably old (XML) OSM.org stylesheet:

https://github.com/openstreetmap/mapnik-stylesheets

Therefore we should probably close this issue.

StyXman:
No, the issue happens because this style works only if `mapnik` is configured not to be strict. I think with properly stating that in the README should be enough.

kocio-pl:
Oh, I see. Could you prepare such PR?

StyXman:
Sure, probably later or tomorrow.

Ircama:
[...]

PR #3089, opened 2018-02-25
Mapnik non-strict mode

StyXman:
Mention that, if Mapnik is being used programmatically, the map should be loaded in non-strict mode.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -130,10 +130,10 @@ For deployment, CartoCSS and Mapnik are required.
 * [CartoCSS](https://github.com/mapbox/carto) >= 0.18.0 (we're using YAML)
 * [Mapnik](https://github.com/mapnik/mapnik/wiki/Mapnik-Installation) >= 3.0
 
-Remember to run CartoCSS with proper API version to avoid errors (at least 3.0.0: `carto -a "3.0.0"`).
+With CartoCSS you compile these sources into a Mapnik compatible XML file. When running CartoCSS, specify the Mapnik API version you are using (at least 3.0.0: `carto -a "3.0.0"`).
 
-With CartoCSS you can compile these sources into a Mapnik compatible XML file. If you're calling Mapnik in your own program, remember to load the XML file in non strict mode, as some fonts are declared with alternative names. This way, you only get a warning instead of an error when Mapnik does not find a font name.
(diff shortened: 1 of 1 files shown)


StyXman:
This 'solves' #3086.

INSTALL.md
```diff
@@ -131,3 +131,9 @@ For deployment, CartoCSS and Mapnik are required.
 * [Mapnik](https://github.com/mapnik/mapnik/wiki/Mapnik-Installation) >= 3.0
 
 Remember to run CartoCSS with proper API version to avoid errors (at least 3.0.0: `carto -a "3.0.0"`).
+
+With CartoCSS you can compile these sources into a Mapnik compatible XML file. If you're calling Mapnik in your own program, remember to load the XML file in non strict mode, as some fonts are declared with alternative names. This way, you only get a warning instead of an error when Mapnik does not find a font name.
+
+For instance, using the Python bindings, this becomes:
+
+    mapnik.load_map(mapnik.Map(width, height), xml_filename, False)  # False for non-strict mode
```
pnorman: wrap this in ```python instead of indenting four spaces

INSTALL.md
```diff
@@ -131,3 +131,9 @@ For deployment, CartoCSS and Mapnik are required.
 * [Mapnik](https://github.com/mapnik/mapnik/wiki/Mapnik-Installation) >= 3.0
 
 Remember to run CartoCSS with proper API version to avoid errors (at least 3.0.0: `carto -a "3.0.0"`).
+
+With CartoCSS you can compile these sources into a Mapnik compatible XML file. If you're calling Mapnik in your own program, remember to load the XML file in non strict mode, as some fonts are declared with alternative names. This way, you only get a warning instead of an error when Mapnik does not find a font name.
```
pnorman: Perhaps

> With CartoCSS you compile these sources into a Mapnik compatible XML file. When running CartoCSS, specify the Mapnik API version you are using (at least 3.0.0: `carto -a "3.0.0"`). If you're calling Mapnik from your own program, load the XML file in non-strict mode. This way, fonts with alternative names will generate warnings, not errors.

matthijsmelissen:
Does this require changes at the OSMF servers? What carto version do they run?

kocio-pl:
BTW: it's a bit off topic, but we have no standard place to communicate such things - there's a new CartoCSS version released (1.0.0):

https://github.com/mapbox/carto/blob/master/CHANGELOG.md#100

PR #3096, opened 2018-03-01
Making gardens to use grass color with plant nursery pattern

### LINKED ISSUE #3022: leisure=garden & leisure=park have the same colour
Public gardens are often sections within parks but atm there's no visible difference: 
https://www.openstreetmap.org/way/191318876#map=19/51.45508/-2.58866

Slight darker shade or a boundary?

kocio-pl:
Resolves #3022.

Green areas redesign (#2964) didn't work too well for gardens. It seems that garden is a very wide concept, so it can be a small residential green area, something like farm, something like park (including public botanical gardens), also a more precise version of village green.

Therefore a more subtle change would be better - something like grass (general mild green) but with a pattern that shows that there's usually something more growing and it's human curated (hence regular pattern from plant nursery). I also opt for 60% opacity for the pattern, since I don't like the home gardens to stand up too much and also because the original pattern color has lower contrast because the background is darker.

Examples:
allv y1p
tphwddxr
8uzejybb
qnd3od_y

### DIFF
--- a/landcover.mss
+++ b/landcover.mss
@@ -1,10 +1,10 @@
 // --- Parks, woods, other green things ---
 
-@grass: #cdebb0; // also grassland, meadow, common, village_green
+@grass: #cdebb0; // also grassland, meadow, common, village_green, garden
 @scrub: #b5e3b5;
 @forest: #add19e;       // Lch(80,30,135)
 @forest-text: #46673b;  // Lch(40,30,135)
-@park: #c8facc;         // Lch(94,30,145) also garden
+@park: #c8facc;         // Lch(94,30,145)
 @orchard: #aedfa3; // also vineyard, plant_nursery
 
 // --- "Base" landuses ---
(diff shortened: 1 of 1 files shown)


kocio-pl:
Unfortunately [grass areas in gardens](https://www.openstreetmap.org/#map=18/55.96388/-3.20537) are not too visible, but at least now they can be spotted:
rulxqtd_

polarbearing:
I like the approach in general. Could we have a side-by-side comparison of the related renderings, i.e. garden, orchard, grass, park, nursery etc, in thumbnail squares somehow? Thanks.

kocio-pl:
I don't understand what do you want - before/after garden renderings, a gallery of different green areas renderings or maybe something else?

polarbearing:
> a gallery of different green areas renderings

Kind of that. Maybe it's worth a separate meta-issue? We are lacking a legend, either automatic #1928 or manual. Maybe we could make a requirement that each PR creates a standardized thumbnail that we can collect somewhere?

kocio-pl:
Discussion about legend belongs probably to #1928 - thanks for recalling, I'll write about it there.

kocio-pl:
@polarbearing Here is the main part of the green areas you wanted to see (the rest is in few other sections): https://wiki.openstreetmap.org/wiki/Standard_tile_layer/Key#Agriculture_and_industry

PR #3099, opened 2018-03-03
Adding rendering for historic=castle

### LINKED ISSUE #744: Add rendering for historic=castle
Castles are important landmarks, so I think we should render them. See also https://trac.openstreetmap.org/ticket/2247.

kocio-pl:
Resolves  #744.
Related to #2871 and #2671.

As proposed in the issue ticket, 3 main icons were used:
- residence (`castle_type=palace/stately/manor`)
- fortified castle (`castle_type=defensive/fortress/castrum/shiro/kremlin`)
- generic castle (no `castle_type` and all the others)

Using more Osmic icons would make the differences too small at this size.

It's important to distinguish residences from the rest, but I also put some effort to make fortified castles look different from generic castles. Osmic icons were a nice base, but with some tweaking - palace roofs were de-orientalized and doors were added to fortified icon to make the difference visible at 14 px (roofs difference was just too weak).

Examples:

[residence](https://www.openstreetmap.org/way/174193669#map=18/52.16109/20.80694)
o2wylhat

[generic castle](https://www.openstreetmap.org/way/35459510#map=18/52.15029/20.78617)
e5d94w_5

[fortified castle](https://www.openstreetmap.org/node/300744233#map=18/51.95772/21.23414)
07q8qijj

Fort icon (merged) and defensive tower (not yet merged) for comparison, as they all belong to the same category:
7ldvybzj
btboto5z


Tomasz-W:
I'm not sure about castle_type=stately icon, it reminds me a mosque. I'm still ready to work on palace/ statley and manor icons. I've presented my propositions in https://github.com/gravitystorm/openstreetmap-carto/issues/744#issuecomment-362909779 and I'm open to any suggestions.

kocio-pl:
My main concern is that these 3 icons should have a common thread, so maybe you could design something around Osmic palace icon? I think a gate is essential element to achieve this impression and using 3-part building scheme (instead of 2-part) works great to make it visibly different, yet similar to other castles.

kocio-pl:
The zoom code update needs fixing (residences shouldn't be selected as castles before z16), but the idea is that castles are rare and important as landmarks and touristic places, so in general they should appear from z15 (see also [proposed towers/masts code](https://github.com/gravitystorm/openstreetmap-carto/pull/2671)), and leave only residences at z16+ (since they can be quite common in city). We show hospitals there for example, also some natural features like caves and peaks, so tis won't be the first icon on this zoom level.

Examples at z15:
nab2z0ld
tdl w3pz

kocio-pl:
Adding manor icon - maybe it could be a bit higher (to be more readable on the map), but it's quite good already:

[manor](https://www.openstreetmap.org/way/300878422#map=17/54.77339/17.75719)
lyukx lg

kocio-pl:
I believe this is the best palace icon so far - original Osmic icon (along with my slight modification) and @Tomasz-W design were too detailed and not too 14 px friendly. This is version is:
- simpler (more similar to other castles and better for small size matrix), 
- with bigger windows and gate (better contrast between fortified and generic castles), 
- not oriental (unlike Osmic) 
- not too flat (unlike Tomasz-W design):

cnbduqci

kocio-pl:
Updated manor icon (taller roof):
zlyzweub

kocio-pl:
Pixel aligned and tall manor rework - looks strange on higher resolution, but works best on 14 px:

jqfvtnp

kocio-pl:
Updated fortified castle icon (bigger gate):
pmgpfy4i

Tomasz-W:
I still would like to see a palaces test rendering with my project to give all participans a fair compare
https://gist.github.com/Tomasz-W/25c553437720cd4181264f9b54ea4382

I like manor shape from https://github.com/gravitystorm/openstreetmap-carto/pull/3099#issuecomment-371670990 . These objects are not squares but rectangles, so IMO it would fit better.

kocio-pl:
Here you are:
aqki3 gu

It's not bad on the rendering, but I'd like to keep the uniform look across similar objects (and for all the reasons I gave before), so we might tweak whatever we want, but I'd like to stay with new shape.

With manor it's harder for me to decide, because you're right - the manors I know look like on your icon (higher roof is not a big difference, but works better for me on small matrix). I still like to have something taller, which is easier to recognize, but it has less typical height/width ratio. 

What other people think about it?

Tomasz-W:
> I'd like to keep the uniform look across similar objects 

I understand this point of view, but for me castles and palaces/ manors are only a little bit similar to each other.

Just look on the random examples:
* castles: https://www.google.pl/search?q=castle&client=opera&hs=hI1&source=lnms&tbm=isch&sa=X&ved=0ahUKEwi8_ZOrvN_ZAhUD2CwKHUYHATEQ_AUICigB&biw=1920&bih=981
* palaces: https://www.google.pl/search?q=palace&client=opera&hs=7xL&source=lnms&tbm=isch&sa=X&ved=0ahUKEwjJn7qsvN_ZAhUHCSwKHSJWDHkQ_AUICigB&biw=1920&bih=981

I think there is no need to keep these icons look similar.

kocio-pl:
I have more reasons than that (see https://github.com/gravitystorm/openstreetmap-carto/pull/3099#issuecomment-371665745). I'm very happy with fortified and generic castles, but palace and manor might be flat if people don't mind as much as I do.

kocio-pl:
I also think about the uniform look of the whole style - for example museum or town hall are based on the square, even if they are very far from the castle.

Tomasz-W:
For comparsion without scrolling the page:

Osmic:


Tomasz-W:

MaestroGlanz:
I prefer the first one. The second looks better, but readability is more important.

hubgitti:
I also  agree with MaestroGlanz!

kocio-pl:
The same question about manor icon (BTW big palace icon was not Osmic, just my rework based on Osmic):

Tomasz version (with my fix - roof is a bit higher)
zlyzweub
my version
jqfvtnp

hubgitti:
Thanks for your effort! I prefer the second version.

lakedistrictOSM:
The fortified castle and palace icons work really well, but none of the proposed manor icons shout "manor" at me yet. 

Just for reference this is the symbol used on British road signs for manor-type places open to the public:

MaestroGlanz:
This is a too detailed icon.  A simplified version might work.

kocio-pl:
I have asked about manor tagging on Tagging list yesterday - let's see what is the correct scheme and what to do with the other one:

https://lists.openstreetmap.org/pipermail/tagging/2018-March/035430.html

kocio-pl:
The discussion still doesn't bring a clear answer about manor tagging at the moment  - a hint about dynamic:

taghistory 16

Tomasz-W:
For me it's obvious that we should choose `historic=palace` for palaces and `historic=manor` for manors. 

PS. I know, I know, it's not a place for tagging discussion ;)

hubgitti:
I read your discussion in the tagging list and checked the German Wikipedia articles about manor houses and castles. I also had private discussions about that topic. In history some manor houses turned into castles others always stayed manor houses. Things are sometimes very close and hard to tell if it is a castle (type manor) or a manor house. Why not use the same symbol for castles (type manor) and manor houses? We could leave that complicated topic and how to display such difficult differences to projects like this here: http://gk.historic.place

kocio-pl:
This is where historic=castle scheme is good for - you can add subtype if you're sure and we see the specific manor icon, but you can skip it and the generic "residence" icon will appear. 

But the problem is with two proper tagging schemes being used at the same time and what to do with them rendering wise:
1. render both manor schemes as manors -> we think that both are valid and we endorse the choice produced by the community
2. render `historic=manor` as manor and `castle_type=manor` as a generic castle (effectively just don't take this type into account) -> we think that `historic=manor` is the only proper way
3. render only castle_type=manor as manor and not render the historic=manor -> we think that `castle_type=manor` is the only proper way

Tomasz-W:
Number 2. is the best solution for me. It should be used also for palaces.

kocio-pl:
With the `historic=palace` it's rather clear for me - it has [372 uses](https://taginfo.openstreetmap.org/tags/historic=palace#overview) and not even a wiki documentation page, while `castle_type=palace` has [756 uses](https://taginfo.openstreetmap.org/tags/castle_type=palace) and documentation page. With manors it's not that clear.

dieterdreist:
sent from a phone

> On 20. Mar 2018, at 15:30, kocio-pl <notifications@github.com> wrote:
> 
> With the historic=palace it's rather clear for me - it has 372 uses and not even a wiki documentation page, while castle_type=palace has 756 uses and documentation page.


the numbers are in the same range

Tomasz-W:
> the numbers are in the same range

Exactly, if something has <1k uses, an "auction" is not a good argument for choosing the proper tag. We should focus on making schemes more intuitive and simpler.

kocio-pl:
After few days away from the problem I made a final summary:
- `historic=palace` - 372 uses (0) - no wiki (0) - dynamic (0) = 0
- `castle_type=palace` - 758 uses (1) - wiki (1) - dynamic (1) = 3
- `historic=stately` - 1 use (0) - no wiki (0) - dynamic (0) = 0
- `castle_type=stately` - 3 374 uses (1) - wiki (1) - dynamic (1) = 3
- `historic=manor` - 3 958 uses (1) - wiki (1) - dynamic (1) = 3
- `castle_type=manor` - 1 463 uses (0) - wiki (1) - dynamic (1) = 2

Findings:
1. For `historic=stately` it's clear that it is not worth rendering. 
2. Not only `historic=palace` has no wiki page, but is also discouraged: https://wiki.openstreetmap.org/wiki/Tag:castle_type%3Dpalace#Possible_tagging_error
3. For manor tagging it's only numbers which favors `historic=manor` currently, but both have fair amount of uses and are in the same range

So my solution for now is to render all the subtypes with specific icons and add `historic=manor` as a proper alternative tagging for manors. But since it's not a preferable situation, I think this tagging and rendering should be revisited after a year. For a forest/wood there's currently little hope to sort the things out, but for manors and palaces the perspective looks much better.

geozeisig:
@kocio-pl 

> Not only historic=palace has no wiki page, but is also discouraged:

Do you think we need a wiki page for `historic=palace`? How does it differ to `historic=castle` + `castle_type=palace` ?

kocio-pl:
> Do you think we need a wiki page for historic=palace? 

Yes. Every tag in OSM database should have documentation, especially those which are really used. I think we should not render undocumented tags.

> How does it differ to historic=castle + castle_type=palace ?

That's probably just alternative scheme for tagging palaces, as with manor tagging schemes, but we just don't know since its meaning is not defined on the wiki page.

geozeisig:
@kocio-pl 

> but we just don't know since its meaning is not defined on the wiki page.

That's the problem. Every day should be different from others.

I have made a new table for [castle _type](https://wiki.openstreetmap.org/wiki/Template:Castle_type_Carto). Maybe it helps to have an overview.

kocio-pl:
> That's the problem. Every day should be different from others.

What do you mean by this? Also what solution would you propose?

> I have made a new table for castle _type. Maybe it helps to have an overview.

Thanks!

Stately uses the same icon as palace and I've just changed a palace icon a bit (gates were too big, so I moved back to smaller ones for the shape to be not distorted).

geozeisig:
@kocio-pl 

> What do you mean by this?

If we introduce a new tag, even without a proposal, we should still stick to the rules of [Proposal process](https://wiki.openstreetmap.org/wiki/Proposal_process).

e.g. Does your proposed tag already exist?
Is the tag you are considering for proposal, worthy of a new tag?

kocio-pl:
I'm still confused what you are trying to prove or achieve.

> If we introduce a new tag, even without a proposal, we should still stick to the rules of Proposal process.
> e.g. Does your proposed tag already exist?
> Is the tag you are considering for proposal, worthy of a new tag?

We at osm-carto don't introduce any tags. In practice just having documentation and considerable use is enough to guarantee that the tag is established enough for us to render/use it somehow. We try to follow community choice (sometimes even too conservatively in my opinion).

If you want me to change something, please say it explicitly, because I don't get it at all currently.

hubgitti:
Awesome, well-thought-out progress here!

Geozeisig, could you please explain in greater detail what you want to say with your comments? You created a nice new template! It's now really comfortable to see which icon belongs to the which castle_type. Will your table eventually be merged to the genuine castle_type wiki page?: https://wiki.openstreetmap.org/wiki/DE:Key:castle_type

geozeisig:
@kocio-pl 

> If you want me to change something, please say it explicitly
Could not change `historic=palace` in `historic=castle` + `castle_type=palace`?
We would not have 2 tags with the same meaning.  That would only cause a mess. I've been cleaning up things for a while now, which is difficult in hindsight.

kocio-pl:
Still not clear what you expect me to do. We're not changing or inventing tags here, we only render them somehow and it looks like `historic=palace` is not a current choice of community. I want to make a review next year to check if something changed, so no need to worry.

geozeisig:
When you render something or not, it has a big impact on tagging. That should already be aware. So the rendering of a tag with wrong syntax is not good.

kocio-pl:
According to the wiki this is a proper syntax and `historic=palace` is tagging error.

hubgitti:
> When you render something or not, it has a big impact on tagging. That should already be aware. So the rendering of a tag with wrong syntax is not good.

I'm getting impatient now: The whole discussion about rendering castles mainly was about two things:
1) The icon styles for the different castle_types
2) Why or why not to render different kinds of historic=* tags.

I think the participants in this thread are 100% aware of what you are saying. So please tell, what do you want???

geozeisig:
@kocio-pl 

> According to the wiki this is a proper syntax and historic=palace is tagging error.

That must be an understanding mistake. I can only emphasize that.

kocio-pl:
If this is a mistake with quite popular type of object, then why this is not fixed yet, after so many years of OSM existence?

hubgitti:
I think I get the point now "geozeisig" supports not to render "historic=palace".

geozeisig:
@kocio-pl 

> then why this is not fixed yet, after so many years of OSM existence?

There are many such mistakes. But they are not easy to fix. There is the [Automated Edits code of conduct](https://wiki.openstreetmap.org/wiki/Automated_Edits_code_of_conduct). And of course there are people who care for that, and in some cases with right. It can be quickly be reverted. But if there is a decision I would do it again. Of course, with care.

kocio-pl:
You talk about high level problem, which is hard, indeed. But I've asked about basic things - why nobody care to at least write documentation for this tag and discuss which is the preferred version on Tagging list? Or in the manor case - somebody documented both tags at least, but still didn't care to define the difference or hint the preferred scheme - why?

Any automated changes follows such actions, so they should be done anyway, yet nobody did even the easier things. And if nobody cares enough to make it clean what's the preferred tagging, why should we choose such tag for displaying? That's why I choose to go with tags which were important enough to document them at least.

If you feel that there is a better way - try do document it and make people believe they should be tagging that way. If that version will prevail, we can change rendering accordingly. But I don't want to rely on your opinion what's best without proofs that this is what most people believe too.

Tomasz-W:
If it's all about make a Wiki page and change one tag into another worldwide, I'm ready to move all palaces from castle:type=palace to historic palace. Just let me know...

hubgitti:
https://en.wikipedia.org/wiki/Buckingham_Palace
"Originally known as Buckingham House, the building at the core of today's palace"

Quite a prominent example: And it seems not to be a castle... :-(

kocio-pl:
> If it's all about make a Wiki page and change one tag into another worldwide, I'm ready to move all palaces from castle:type=palace to historic=palace. 

It's not just about the numbers - it's about established way of tagging. It would be great if somebody do the groundwork though:
1. Create a wiki documentation for all the tags to be rendered.
2. Describe the usage and difference between similar forms (if any).
3. Discuss on the Tagging list which form should be preferred if it's hard to find a difference between forms.
4. Update documentation according to the discussion.
5. Move the tagging from the deprecated form in a civilized manner - see https://wiki.openstreetmap.org/wiki/Deprecated_features

I think reaching some agreement in 3. is the hardest part.

dieterdreist:
sent from a phone

> On 31. Mar 2018, at 18:45, kocio-pl <notifications@github.com> wrote:
> 
> According to the wiki this is a proper syntax and historic=palace is tagging error.


because a lot of the wiki authors seem to be great fans of disliking simple tags (k=v) in favor of combinations that only make the same sense if you use more than one tag (eg. highway=bus_stop vs 2 tags in pt scheme; the whole social facility tag series; highway=path with foot=designated for dedicated footpaths; castle types and tower types to distinguish instances from a pool of quite different things; tourism=information which combines information offices with boards and guide posts, etc). When it is only about details it is ok to push the specifics a level deeper, but when it is about completely different classes it is rather pointless to (also) group them under a very generic tag that doesn’t make sufficient sense if used alone.

dieterdreist:
sent from a phone

> On 1. Apr 2018, at 09:56, hubgitti <notifications@github.com> wrote:
> 
> Quite a prominent example: And it seems not to be a castle... :-(



outside of England it isn’t very clear what should be considered a castle. E.g. are the Italian palazzi castles? Are they palaces? What about the “Palast der Republik”? OK, history saved us from having to resolve the last one, but I bet there are “similar” instances

dieterdreist:
sent from a phone

> On 1. Apr 2018, at 10:20, kocio-pl <notifications@github.com> wrote:
> 
> it's about established way of tagging.


it’s also about an intuitive way of tagging, ideally the system works as a whole

kocio-pl:
The problem is that system is big and complicated and you can find more than one "proper" solution, most of the time there are few sane possibilities. That's why we need documentation to know which one is chosen by the community in general.

kocio-pl:
@geozeisig I don't think that such a mass edit is a way to go, I propose to revert it:

https://www.openstreetmap.org/changeset/57705945

Let's not make anything in a hurry. Documenting existing tags and discussion on Tagging are at the core of the proper solution in my opinion.

PR #3100, opened 2018-03-03
Fix duplicate key in linestring_values table

### LINKED ISSUE #3084: natural=cliff is ignored on closed ways
See this path of natural=cliff:
   https://www.openstreetmap.org/way/547951335
which has the same nodes as area:
  https://www.openstreetmap.org/way/563113425

The bare_rock have been rendered, but the cliff tag is then ignored.

pnorman:
This was causing objects with natural=cliff or natural=tree_row to be incorrectly processed.

Fixes #3084

This does not require a MAJOR release and can go into the next 4.x release. It could even go into a patch release since it's a bugfix. The schema hasn't changed, previous code wasn't producing output that matched the target schema.

Should we attempt to add some SQL for a migration? It's likely to be reasonably long because the column lists are very long.
I'd probably write something along the lines of

```sql
WITH areas_deleted AS (
  DELETE FROM planet_osm_polygon
  WHERE "natural" IN ('cliff', 'tree_row')
    AND NOT (tags ? 'area')
    AND osm_id > 0
  RETURNING * )
INSERT INTO planet_osm_line
  SELECT
      osm_id, access, "addr:housename", ..., 
      tags,
      (ST_Dump(way)).geom AS way;
```

Given the minor impact I'm fine not including a migration.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,10 @@
 ## [Unreleased](https://github.com/gravitystorm/openstreetmap-carto/compare/v4.8.0...master)
+### Major Changes
+- A bug where closed ways with natural=cliff or natural=tree_row were not
+  rendering has been fixed. This required fixing a transform bug. The fix
+  will apply to all objects when they are created in OSM, but there is no
+  migration for existing databases. Deployments will have to decide if the
+  effects are serious enough to require them to reload the database.
 
 ## [v4.8.0](https://github.com/gravitystorm/openstreetmap-carto/compare/v4.7.0...v4.8.0) - 2018-02-23
 ### Changes
--- a/openstreetmap-carto.lua
+++ b/openstreetmap-carto.lua
@@ -35,11 +35,10 @@ local polygon_keys = {
 local linestring_values = {
     leisure = {track = true, slipway = true},
     man_made = {embankment = true, breakwater = true, groyne = true},
-    natural = {cliff = true, tree_row = true},
+    natural = {cliff = true, tree_row = true, ridge = true, arete = true},
     historic = {citywalls = true},
     waterway = {canal = true, derelict_canal = true, ditch = true, drain = true, river = true, stream = true, wadi = true, weir = true},
-    power = {line = true, minor_line = true},
-    natural = {ridge = true, arete = true}
+    power = {line = true, minor_line = true}
 }
 
 -- Objects with any of the following key/value combinations will be treated as polygon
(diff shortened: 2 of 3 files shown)


pnorman:
> Can we add a test which checks for duplicate entries in the definitions like linestring_values?

Not to my knowledge. Strictly speaking, I don't believe there were duplicates, but the second value replaced the first. I'd have to check the Lua language specifications to be sure.

HolgerJeromin:
We would need to parse the code ourself. Perhaps too much work

matthijsmelissen:
Good catch!

We have no future full database import on the horizon, and I'd really like this to be fixed on the OSMF tile servers as well, I think a migration would be very helpful. @pnorman would it be very much effort to define one?

kocio-pl:
Can this code be safely merged as is? I mean - is migration code an additional thing if database will not be reloaded (the map just won't show fixes) or it's essential in such case (rendering will be broken)?

matthijsmelissen:
It’s not essential, rendeting won’t break.

The only issue is if we reload the database and osm.org doesn’t do a migration, rendering on our local machines and osm.org will diverge, which we might want to avoid.

HolgerJeromin:
We could sort the lines of linestring_values (and others) alphabetically.
Probably this would prevent duplicate entries in the future.

pnorman:
I've merged it and we can consider a migration as part of the release.

I'm reluctant to write one, since it means writing it well enough to be confident that it won't screw up someone's production database. This means testing it.

kocio-pl:
> I've merged it and we can consider a migration as part of the release.

I was planning to make v4.9.0 in Friday - how do you think this merge influences the moment of release?

> I'm reluctant to write one, since it means writing it well enough to be confident that it won't screw up someone's production database. This means testing it.

So we have no migration code, which is needed to make a release, but the bugfix is in the master, which will be active with the nearest release - do I understand this properly?

HolgerJeromin:
Without the migration only new ways will get it right. 
When released without old databases will become a mixure of correct new and wrong old cliffs.
But imo the migration is independent from the release.
Tileserver admins have to perform:

Update of osm-carto with new lua so new ways are correct
Execute a sql migration to fix old data,too

Both operations do not have to be at the same time, only in the correct order. 

A migration could imo be done even months later, so the release dont have to be on hold

kocio-pl:
Thanks, that's what I supposed, but it's better to ask. Fine then, unless there will be some voice against it, I will make the release as planned.

PR #3101, opened 2018-03-04
Fix repetition of high-zoom admin/nature reserve labels

matthijsmelissen:
The code that renders admin and nature reserve labels along borders
(on z16 and higher) was broken: it only rendered one label per
metatile. With the new code, labels are rendered every 600px.

### DIFF
--- a/admin.mss
+++ b/admin.mss
@@ -172,7 +172,8 @@ overlapping borders correctly.
   text-halo-radius: @standard-halo-radius;
   text-halo-fill: @standard-halo-fill;
   text-placement: line;
-  text-clip: true;
+  text-clip: false;
+  text-spacing: 600;
   text-vertical-alignment: middle;
   text-dy: -10;
 }
@@ -184,7 +185,8 @@ overlapping borders correctly.
   text-halo-radius: @standard-halo-radius;
   text-halo-fill: @standard-halo-fill;
   text-placement: line;
-  text-clip: true;
+  text-clip: false;
+  text-spacing: 600;
   text-vertical-alignment: middle;
   text-dy: -10;
 }


kocio-pl:
Could you provide example rendering of this problem?

matthijsmelissen:
Forgot:
screen shot 2018-03-05 at 00 20 00
screen shot 2018-03-05 at 00 19 54

matthijsmelissen:
Forgot.

Before:
screen shot 2018-03-05 at 00 19 54

After:
screen shot 2018-03-05 at 00 20 00

kocio-pl:
I want to test it soon, but judging from the rendering it looks like an important fix.

kocio-pl:
There's something strange - when I compare [this place](https://www.openstreetmap.org/#map=16/52.6380/18.8216) on master and with your patch, they look the same:

Before
6ktovtgp
After
ukfmcva

But on the server it looks empty:
screenshot-2018-3-8 openstreetmap carto kosmtik

Maybe it has something to do with Mapnik version - I use 3.0.15, while on the OSMF servers 3.0.9 is used currently, AFAIK. 

What do you think about it?

springmeyer:
@matthijsmelissen see https://github.com/mapnik/mapnik-reference/blob/gh-pages/3.0.6/reference.json#L2267. If you still don't understand the difference then I recommend looking inside the Mapnik code. I personally don't remember the difference.

matthijsmelissen:
@kocio-pl As far as I'm concerned this can be merged. What do you think? Do you need something from me still?

kocio-pl:
I just don't understand if it really works and is worth merging, so I waited for making clear why I don't see the before/after difference on my machine, but instead only the difference between before on my machine and OSMF servers.

I would just like to understand it better and make independent test showing this result, but I can also trust you and go on with the change, because your example shows that at least in some cases it makes a difference. Could you provide a link BTW?

matthijsmelissen:
> No, my guess is that it has to do with the fact that osm.org generates larger metatiles.

I thought a bit more about this. The old code renders only one label per metatile. The server has larger metatiles than default Kosmtik, so it makes sense that (with the old code) Kosmtik renders more metatiles than the server does.

I'm going ahead and will merge this.

kocio-pl:
There is a performance drop I encountered during v4.10.0 pre-release testing (see https://github.com/gravitystorm/openstreetmap-carto/issues/3159#issuecomment-382932958 for details), so I will revert this change now.

> The server has larger metatiles than default Kosmtik, so it makes sense that (with the old code) Kosmtik renders more metatiles than the server does.

New code requires much more CPU, but I'm not sure what's the reason.

kocio-pl:
Thanks for the info! I was curious if that was only on my laptop with Kosmtik or this is a problem also for real servers.

PR #3103, opened 2018-03-05
Simplify (generalize) admin borders

matthijsmelissen:


matthijsmelissen:
Be careful with testing: for some reason Kosmtik has problems handling a change in simplify algorithm. It seems it does not detect the change, and caches the old value. Before and after testing this PR, it is therefore necessary to quit Kosmtik, rather than just checkout the other branch with Kosmtik running.

matthijsmelissen:
Before:
screen shot 2018-03-05 at 02 03 31
After:
screen shot 2018-03-05 at 02 04 24

Before:
screen shot 2018-03-05 at 02 05 30

After:
screen shot 2018-03-05 at 02 06 35

matthijsmelissen:
Before/after:
screen shot 2018-03-05 at 02 03 31 screen shot 2018-03-05 at 02 09 20

Before/after:
screen shot 2018-03-05 at 02 06 09 screen shot 2018-03-05 at 02 06 35

kocio-pl:
The effect is subtle, but I like it. It's also safe to use it, because this function is available since API 2.2.0.

Poland (click on the images to view the full size):

z6
Before
8zakw0dt
After
echyd916

z7
Before
v8jqipma
After
gvviqwh6

imagico:
For reference for those who are not aware of it - this is essentially a re-submission of an idea that has been discussed and ultimately dismissed in #907 and #1938 - technical implementation differs but functionally it is the same.

kocio-pl:
Thanks for the links - a lot of reading, though, because I don't even remember what I was thinking back then. This was a prohibitively big problem with performance, right? However if this solution would fly, I would also like to test it on natural reserve borders.

kocio-pl:
> For reference for those who are not aware of it - this is essentially a re-submission of an idea that has been discussed and ultimately dismissed in #907 and #1938 - technical implementation differs but functionally it is the same.

The question is why were they dismissed - because of functionality or technical implementation?

This https://github.com/gravitystorm/openstreetmap-carto/pull/1938#issuecomment-160942402 is about severe performance problems ("I had to revert this pull request, since it was bringing the OSMF servers to a halt") - the code however has been merged before, so I wouldn't even call it "dismissed". I have yet to look at #907.

@matthijsmelissen What do you think about it?

matthijsmelissen:
Exactly: #1938 was reverted for a technicality, namely the performance issues of ST_Within. That doesn't play a role here, and it seems to be running fine on @rrzefox' server.

matthijsmelissen:
In the demo renderings, I have seen no evidence we'll run into this. Thanks for mentioning it though!

Issue #1167, opened 2014-12-19
Covered reservoirs


DaveF63:
Hi
Covered reservoirs (covered=yes) mapped as areas are being rendered solid blue, the same as open ones. Could it be amended to just a dashed outline or maybe narrow band of blue shading similar to the new national parks boundaries render?

matkoniecz:
I admit that I am not sure what is represented by [covered = yes landuse=reservoir].

Can somebody link to an example photos?

@DaveF63 Is https://commons.wikimedia.org/wiki/File:Covered_reservoir_seen_from_the_top_of_Primrose_Hill_-_geograph.org.uk_-_784303.jpg a photo of https://www.openstreetmap.org/way/43160349#map=18/51.39740/-2.38280 ?

DaveF63:
I don't have photos, but generally it's any reservoir where the water isn't visible. 
In my area of the UK there are numerous water tanks, semi buried below ground level. The half above ground is landscaped, covered in earth & grass.

Tomasz-W:
retaled to #2620

jragusa:
I investigated these covered reservoirs and it seems there is a tagging mistake. Both examples proposed above (https://www.openstreetmap.org/way/43160349 and https://www.openstreetmap.org/node/2789983153) should be better described as `man_made=reservoir_covered` and it's the same for several equivalent "covered reservoirs".

A correct covered reservoir is for example https://www.openstreetmap.org/way/574995790 ([see the aerial photograph](https://www.geoportail.gouv.fr/carte?c=5.192751913660046,45.962857934988534&z=21&l0=ORTHOIMAGERY.ORTHOPHOTOS::GEOPORTAIL:OGC:WMTS(1)&l1=GEOGRAPHICALGRIDSYSTEMS.MAPS::GEOPORTAIL:OGC:WMTS(0)&l2=GEOLOGY.GEOLOGY::EXTERNAL:OGC:EXTERNALWMS(1;h)&permalink=yes) : a reservoir covered with a tarpaulin. Following this example, the rendering should not be changed.

So I propose to close this issue and recommend people to fix tag to `man_made=reservoir_covered` if they are semi-buried and covered by grass or add `building=*` if they are located in a building.

Tomasz-W:
@kocio-pl https://github.com/gravitystorm/openstreetmap-carto/issues/1167#issuecomment-441454335

kocio-pl:
[...]

PR #3104, opened 2018-03-06
Fix #996 by adding intermittent water bodies support

### LINKED ISSUE #996: Special rendering for intermittent water areas

Penegal:
This adds support for rendering intermittent water bodies, hence it fixes #996.

Rendering examples:
https://www.openstreetmap.org/way/391188151
image
https://www.openstreetmap.org/way/394053640
image
https://www.openstreetmap.org/way/466317471
image
https://www.openstreetmap.org/way/507366289
image


kocio-pl:
Here are the keywords for automatically closing tickets:

https://help.github.com/articles/closing-issues-using-keywords/

Could you post some images with the rendering of the actual code from this PR?

Penegal:
@kocio-pl: Edited.

kocio-pl:
Thanks! The links to rendered locations are useful for testing - could you post them too?

It looks great and intuitive for me, some more examples:

[Intermittent pond in the forest near the river](https://www.openstreetmap.org/way/497662523#map=19/52.16843/21.13375)
al3d54wf

[Intermittent water in the small park ](https://www.openstreetmap.org/way/542298114#map=19/52.12824/21.05654)
zais5eo3

[Seasonal pond with a name in the park](https://www.openstreetmap.org/way/98345824#map=18/52.16753/21.02205)
xhwxnnwd

Penegal:
Edited.

kocio-pl:
South Australia example on z6 - I don't know if the tagging is complete, but at least some of the biggest lakes there are clearly intermittent:
xictib09

Whole continent view on z5 (click to see the full size):
v7qapvzy

sommerluk:
Given that the pattern is quite simple and maybe pixel-aligned, is there a chance to have this as SVG pattern avoiding #2750 problems?

kocio-pl:
Currently we're not sure why such problems exist at all and how to fix/avoid them with vector patterns (looks to me like a Mapnik issue), so raster version is the only safe bet for now.

Penegal:
What does miss for this to be merged? Approval? Testing?

kocio-pl:
I just wait one day more to gave a chance for anybody to post additional comments/improvements, because it's now more visible than just an old issue, but that's all.

Penegal:
OK, just wanted to be sure.

kocio-pl:
Thanks for the work! I like the result, especially because it tunes a look on low zoom (probably not only in Australia).

turnsole80:
I've updated the tagging on some of the lakes in Australia, so it should look even better now :)

kocio-pl:
Thanks a lot! I've been thinking lately about asking the Aussie community about it.

There can be similar examples across the globe, if anybody knows one, please tag them too.

kocio-pl:
Another interesting thing to test: #1167.

Gazer75:
There is a small issue with this. It will not display over a dam polygon.
https://www.openstreetmap.org/relation/8248594

Gazer75:
Why not? When the doors are open the water flows over in that area.

kocio-pl:
The main reason is that there are so much combinations with layers and they are expensive (in terms of performance and maintenance), that we must have a strong reasons to add/split more of them. And it would mean that intermittent water should have its own layer. And that would be always above the whole class of things, not only dams...

We would need some generic solution like global render layers - that would mean that layer=1 is always above layer=-1. Yet we have only per-layer rendering priority (road with layer=1 over road with layer=-1) not the global one (building with layer=1 over road with layer=-1). I'm not even sure if this is doable with our framework, but even if it is, it would be huge code rewrite probably.

For a notable example of layering problem without global rendering layers see https://github.com/gravitystorm/openstreetmap-carto/issues/688#issuecomment-232675998.

PR #3107, opened 2018-03-07
Move natural=grassland and landuse=meadow earlier

kocio-pl:
[Example of natural=grassland](https://www.openstreetmap.org/#map=9/51.3906/67.5536) shown from z8 (instead of z10+, as it is currently). Maybe also meadows could be visible there, but not the natural=grass, since it is intended for smaller grass areas.

z8
feal18_u

z9
fkxvk8_k

### DIFF
--- a/landcover.mss
+++ b/landcover.mss
@@ -323,16 +323,14 @@
     }
   }
 
-  [feature = 'landuse_meadow'],
-  [feature = 'natural_grassland'],
-  [feature = 'landuse_grass'],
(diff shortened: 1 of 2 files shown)


kocio-pl:
Meadows also deserve being shown earlier, so I added it here too - [Fresno example](https://www.openstreetmap.org/relation/5793421#map=9/36.3306/-120.5763) is easily visible on z8:

b4hn0r6n

PR #3123, opened 2018-03-17
Issue #2929 Add icon for shop=music 

james2432:
Related to #2929

PR #3125, opened 2018-03-17
Issue #2660 Shop medical supply icon

### LINKED ISSUE #2660: Rendering shop=medical_supply with a special icon
The tag shop=medical_supply has > 3,400 instances. I suggest the following icon:
med

james2432:
Fixes #2660

From: https://www.openstreetmap.org/#map=19/45.55000/-73.65399 (node)
image

image

image

From https://www.openstreetmap.org/#map=17/45.42026/-75.62195 (way):

image

image

image


kocio-pl:
Now we have a common problem when working with multiple similar issues - after merging #3123, which was changing the same lines in `project.mml`, you should resolve the conflicts to make this code mergable again. This is probably the same problem with all the other shop icon rendering - and it will repeat with each merge.

james2432:
Done

kocio-pl:
Thanks! That was quick. :smile:

james2432:
On to the next! :yum:

PR #3129, opened 2018-03-18
Leisure sauna

### LINKED ISSUE #3062: Add rendering for leisure=sauna
leisure=sauna https://wiki.openstreetmap.org/wiki/Tag:leisure%3Dsauna has over 7000 uses. Half of them as ways like
https://www.openstreetmap.org/way/558362650

perhaps we dont need an icon on the first run as most pois have a matching name

The fill could be the same as sports_centre (proposed for waterpark #1490)

james2432:
Fixes #3062 

From: https://www.openstreetmap.org/node/4709907893#map=17/46.66801/-71.23048 (node)
image

From: https://www.openstreetmap.org/#map=18/37.44146/-122.15891 (way)
image

image

### DIFF
--- a/amenity-points.mss
+++ b/amenity-points.mss
@@ -1035,7 +1035,14 @@
     marker-fill: @amenity-brown;
     marker-placement: interior;
     marker-clip: false;
-  }  
+  }
+
+  [feature = 'leisure_sauna'][zoom >= 17] {
+     marker-file: url('symbols/sauna.svg');
+     marker-fill: @leisure-green;
+     marker-placement: interior;
+     marker-clip: false;
+   }
 
   // Slipway tagging on points - slipway on lines is defined later 
   [feature = 'leisure_slipway'][zoom >= 17] {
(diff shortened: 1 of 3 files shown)


amenity-points.mss
```diff
@@ -2109,6 +2121,7 @@
   [feature = 'shop_car_repair'],
   [feature = 'shop_car_parts'],
   [feature = 'shop_cosmetics'],
+  [feature = 'shop_dairy'],
```
HolgerJeromin: the diff is not clean as dairy is not merged yet

Tomasz-W:
I think it supposed to be leisure-green as it is a leisure element.

james2432:
Like this? 
image

amenity-points.mss
james2432: it is now :P sorry had issues with git not rebasing at proper location

james2432:
Going to open another cleaner PR that doesnt have dairy

kocio-pl:
Great! You don't need however to close this one (and I don't recommend it in general) - just clean your branch and push the code with an `-f` option.

kocio-pl:
`Merge branch 'master' into leisure_sauna` should not be here - it is not just merging, but also code formatting change (which I think is bad, since it breaks clean indent), and it is also makes it to be 2 commits for 1 change.

james2432:
there, squashed.

amenity-points.mss
```diff
@@ -1035,7 +1035,14 @@
...
     marker-clip: false;
-  }  
+  }
+
+  [feature = 'leisure_sauna'][zoom >= 17] {
+     marker-file: url('symbols/sauna.svg');
+     marker-fill: @leisure-green;
+     marker-placement: interior;
+     marker-clip: false;
+   }
```
kocio-pl: Unwanted spaces left on the end of this line.

james2432:
Removed spaces, if you need me to squash into one commit I'll have to do so tonight

kocio-pl:
Please do, it's always better to make it clean also technically if possible.

james2432:
A lot easier when you've squashed a few commits into one ;) 
Done.

kocio-pl:
Great. Please try to add more descriptive names - leisure sauna is about adding icon for such objects, but could be about changing some options or anything else. It might be useful for osm-carto forks when updating their repositories to know exactly what a given changeset is doing without looking anywhere else.

james2432:
Reworded.

kocio-pl:
Thanks!

PR #3130, opened 2018-03-18
sort linestring_values in lua

HolgerJeromin:
Sorted list could perhaps prvent something like #3084

### DIFF
--- a/openstreetmap-carto.lua
+++ b/openstreetmap-carto.lua
@@ -33,12 +33,12 @@ local polygon_keys = {
 
 -- Objects with any of the following key/value combinations will be treated as linestring
 local linestring_values = {
+    historic = {citywalls = true},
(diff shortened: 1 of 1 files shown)


kocio-pl:
Is it really useful patch? We could merge it then before release, but this is more coding related than I can evaluate - ping @matthijsmelissen @pnorman @matkoniecz.

matthijsmelissen:
@kocio-pl Do you look at the change made in #3100? Really not that hard to understand :) 

The natural definition occured twice, the second overwrote the first.

kocio-pl:
Thanks! Yes, I did :smile: , I just didn't know how is it useful from a practical point of view.

matthijsmelissen:
The point is simply, if we keep the right order it’s easiier to spot duplicates :)

PR #3131, opened 2018-03-18
Issue #2224 Construction steps

### LINKED ISSUE #2224: Add rendering for construction=steps to distinguish it from roads
Currently steps in construction (highway=construction, construction=steps) are rendered like a road under construction (the same rendering as highway=construction without any construction= tag). See for example this flight of steps in Wuppertal: http://www.openstreetmap.org/way/272616133 .

The present rendering makes steps look like a road under construction, which is weird.

james2432:
Fixes #2224

See https://github.com/gravitystorm/openstreetmap-carto/issues/2224#issuecomment-373996250 for example rendering


kocio-pl:
Please remove dairy shop commits - always try to merge the code on top of a current master.

PR #3133, opened 2018-03-19
Adding amenity=bbq icon

### LINKED ISSUE #2996: Rendering of amenity=bbq
I have done a fair amount of mapping contributions and I am wondering why the amenity=bbq is not being rendered.

james2432:
Fixes #2996 
From https://www.openstreetmap.org/node/4983372050#map=19/45.35162/-75.82804 (node):
image

From: https://www.openstreetmap.org/#map=19/43.31581/-70.92575 (way)
image
~(Should we render ways with a border? or just the little icon?)~

From: https://www.openstreetmap.org/#map=17/45.51983/-75.84958 (node with text in name)
image


kocio-pl:
> Should we render ways with a border? or just the little icon?

I guess icon is enough. If this is in the recreation zone, it's easy to add such area in parallel.

amenity-points.mss
```diff
@@ -1906,7 +1913,8 @@
     }
   }
 
-  [feature = 'amenity_atm'][zoom >= 17] {
+  [feature = 'amenity_atm'][zoom >= 17],
+  [feature = 'amenity_bbq'][zoom >= 17] {
     text-name: "[operator]";
```
kocio-pl: Name tag is used here, operator is not even visible in stats:

https://taginfo.openstreetmap.org/tags/amenity=bbq#combinations

james2432:
Replaced operator use by name, not sure why I was using operator (was thinking maybe bbq's usually are operated by someone vs have a name?)

kocio-pl:
Nice, could you also render a way with a name?

kocio-pl:
After merging similar item (`amenity=shower`) you need to fix the conflicts in `project.mml` to make this PR mergable again.

james2432:
Fixed.

kocio-pl:
It renders OK and I'm gonna merge it, but could you make it 1 commit (instead of 2)?

james2432:
Done.

amenity-points.mss
kocio-pl: Something went wrong and - just by looking at code - you render operator instead of name again.

kocio-pl:
Unfortunately we're back to the operator label (instead of a name label).

james2432:
Not sure what went wrong, possibly this computer didn't fetch my origin changes. Sorry about that

james2432:
Fixed the conflict issue with advertising

kocio-pl:
Thanks! 

We are getting more outdoor lately (shower, bbq, fire pit, aerialways, shrines, towers...). That's great, OSM data are really diverse.

PR #3134, opened 2018-03-19
Adding amenity=shower icon

### LINKED ISSUE #3067: Add rendering for amenity=shower
Related to #3062 and #2856. May be useful especially for rendering camping, sport places or beach areas.

It can be designed probably quite easily - anybody wants to try?

This tag is quite popular (12 695 uses) and has its own wiki page:
https://wiki.openstreetmap.org/wiki/Tag:amenity%3Dshower

james2432:
Fixes #3067 

From: https://www.openstreetmap.org/node/4898075512#map=18/44.60609/-76.12866 (node)
image

From: https://www.openstreetmap.org/#map=19/45.72938/-68.83598(way)
image
Fake name added to test rendering:
image

### DIFF
--- a/amenity-points.mss
+++ b/amenity-points.mss
@@ -338,6 +338,13 @@
     marker-placement: interior;
     marker-clip: false;
   }
+  
+  [feature = 'amenity_shower'][zoom >= 18] {
+    marker-file: url('symbols/shower.svg');
+    marker-fill: @amenity-brown;
+    marker-placement: interior;
+    marker-clip: false;
+  }
 
   [feature = 'amenity_social_facility'][zoom >= 17] {
     marker-file: url('symbols/social_facility.svg');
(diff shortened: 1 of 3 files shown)


polarbearing:
Looks a bit like garden watering for me. could we try my new one in #3067 ? :-)

kocio-pl:
I still like the current proposition more and it's easier for me to recognize too. When thinking about garden watering I think of this shape, which we use for gardening centre: https://github.com/gravitystorm/openstreetmap-carto/blob/master/symbols/shop/garden_centre.svg

james2432:
Re-touched the zoom aspect now is at zoom >=18

kocio-pl:
What about name rendering for a node and a way? The code is here, so it should be tested.

kocio-pl:
On the other hand stats tell that name is not really popular with showers:

https://taginfo.openstreetmap.org/tags/amenity=shower#combinations

james2432:
should I take out name for showers? wiki says to add name and operator:

https://wiki.openstreetmap.org/wiki/Tag:amenity=shower

kocio-pl:
I was mentioning the numbers just to show the practice, but it's up to us to decide. You can leave it (it's perfectly fine when it's defined explicitly on the wiki, and from z18 it shouldn't make a visual problem), just please make the testing renderings to make sure there are no bugs in the code.

james2432:
updated screenshots with name

kocio-pl:
I prefer real locations (with links) for both node and a way, because I try to test the code before merging as a good practice and it's easier to compare the results then.

james2432:
https://www.openstreetmap.org/#map=19/45.19130/-81.55369
image

kocio-pl:
It's a node (https://www.openstreetmap.org/node/3786627020), what about the way label?

james2432:
the way label was tested with the fake name added but I can test with this later:

https://www.openstreetmap.org/way/536787545#map=19/47.57643/-122.13311

dieterdreist:
great icon

james2432:
image
Seems to work a.o.k.

kocio-pl:
Thanks, it really does work as it should.

PR #3136, opened 2018-03-20
Adding advertising=column icon

### LINKED ISSUE #3002: Add icon for advertising=column
There is more than 7k of an advertising colums:
https://taginfo.openstreetmap.org/tags/advertising=column

Nowadays it's rather rare, archaic thing, so I think it wouldn't make mess (as e.g. rendering of advertising=billboard). In modern cities these columns are often used as orientation/ meeting point, so it would be useful to have them on a map.

james2432:
Fixes #3002 

From https://www.openstreetmap.org/#map=19/45.51028/-73.61213 (node):
Z19:
image
Z18:
image

From https://www.openstreetmap.org/#map=19/48.53969/2.65847 (way):

image

### DIFF
--- a/amenity-points.mss
+++ b/amenity-points.mss
@@ -14,6 +14,7 @@
 @culture: @amenity-brown;
 @public-service: @amenity-brown;
 @man-made-icon: #555;
+@advertising-grey: @man-made-icon;
 @landform-color: #d08f55;
 @leisure-green: darken(@park, 60%);
 
@@ -968,6 +969,13 @@
     [shop = 'variety_store'][zoom >= 18] {
       marker-file: url('symbols/shop/variety_store.svg');
     }
+  }
+
+  [feature = 'advertising_column'][zoom >= 19]{
+      marker-file: url('symbols/advertising/column.svg');
+      marker-fill: @advertising-grey;
+      marker-placement: interior;
+      marker-clip: false;
   }
 
   [feature = 'leisure_water_park'][zoom >= 17] {
--- a/project.mml
+++ b/project.mml
@@ -1441,6 +1441,7 @@ Layer:
                                                   'fast_food', 'telephone', 'taxi', 'theatre', 'toilets', 'drinking_water',
                                                   'prison', 'hunting_stand', 'nightclub', 'veterinary', 'social_facility',
                                                   'charging_station', 'arts_centre', 'ferry_terminal', 'marketplace', 'waste_disposal') THEN amenity ELSE NULL END,
+              'advertising_' || CASE WHEN tags->'advertising' in ('column') THEN tags->'advertising' else NULL END,                                                  
               'shop' || CASE WHEN shop IN ('no', 'vacant', 'closed', 'disused', 'empty') OR shop IS NULL THEN NULL ELSE '' END,
               'leisure_' || CASE WHEN leisure IN ('water_park', 'playground', 'miniature_golf', 'golf_course', 'picnic_table',
                                                   'fitness_centre', 'fitness_station', 'firepit', 'sauna') THEN leisure ELSE NULL END,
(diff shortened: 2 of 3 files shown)


kocio-pl:
I think you should skip using issue number in the names of PRs and commits - it's unnecessary technical detail which adds visual clutter and can be easy checked inside the PR ticket.

james2432:
Ok future commits and PR I will only put inside the PR body.

polarbearing:
I had forgotten that this was unfinished in January, thanks for progressing. 
As there is some new code involved we probably need to see that rendered. The two links in the first post are map areas only, not objects?

james2432:
im working on the renderings now, didnt have access to the docker vm

openstreetmap-carto.style
```diff
@@ -11,6 +11,7 @@ node,way   addr:housename      text  linear
 node,way   addr:housenumber    text  linear
 way        addr:interpolation  text  linear
 node,way   admin_level  text         linear
+node,way   advertising  text         polygon
```
kocio-pl: I think this is not a good idea - it would mean database migration. We add keys from hstore like tags->'intermittent'.

amenity-points.mss
```diff
@@ -969,6 +970,15 @@
       marker-file: url('symbols/shop/variety_store.svg');
     }
   }
+
```
kocio-pl: Unwanted spaces.

amenity-points.mss
```diff
@@ -969,6 +970,15 @@
       marker-file: url('symbols/shop/variety_store.svg');
     }
   }
+  
+  [feature = 'advertising']{
+    [advertising = 'column'][zoom >= 19] {
```
kocio-pl: Currently we use single features like `[feature = 'advertising_column']`. It is a text concatenation in project.mml.

james2432:
Fixed, tested, same result now; less the database migration requirement

project.mml
```diff
@@ -1488,6 +1489,7 @@ Layer:
             OR tourism IN ('artwork', 'alpine_hut', 'camp_site', 'caravan_site', 'chalet', 'wilderness_hut', 'guest_house', 'apartment', 'hostel',
                            'hotel', 'motel', 'information', 'museum', 'viewpoint', 'picnic_site')
             OR amenity IS NOT NULL -- skip checking a huge list and use a null check
+            OR tags->'advertising' IS NOT NULL
```
kocio-pl: When the list is relatively short (like here - just 1 element) we don't skip checking,  see for example:

`OR leisure IN ('water_park', 'playground', 'miniature_golf', 'golf_course', 'picnic_table', 'fitness_centre',
                            'fitness_station', 'firepit', 'sauna')`

project.mml
```diff
@@ -2031,6 +2036,7 @@ Layer:
               OR tourism IN ('artwork', 'alpine_hut', 'hotel', 'motel', 'hostel', 'chalet', 'wilderness_hut', 'guest_house', 'apartment', 'camp_site', 'caravan_site', 'theme_park',
                              'museum', 'viewpoint', 'attraction', 'zoo', 'information', 'picnic_site')
               OR amenity IS NOT NULL -- skip checking a huge list and use a null check
+              OR tags->'advertising' is NOT NULL
```
kocio-pl: The same comment about skipping list.

project.mml
```diff
@@ -2179,6 +2186,7 @@ Layer:
                   OR tourism IN ('artwork', 'alpine_hut', 'hotel', 'motel', 'hostel', 'chalet', 'wilderness_hut', 'guest_house', 'apartment', 'camp_site', 'caravan_site', 'theme_park',
                                  'museum', 'viewpoint', 'attraction', 'zoo', 'information', 'picnic_site')
                   OR amenity IS NOT NULL -- skip checking a huge list and use a null check
+                  OR tags->'advertising' IS NOT NULL
```
kocio-pl: The same comment about skipping list.

james2432:
Added checking for short list in project.mml

kocio-pl:
Commit name should now skip the support for advertising key.

james2432:
Renamed commit comment

kocio-pl:
There's also some `project.mml` conflict to resolve now.

james2432:
Dont but is probably going to break when amenity=bbq is merged

kocio-pl:
Thanks! Sometimes it's better to step back for a while - now I think this icon design is the best of what we were trying.

BTW: it doesn't show up on ways without `building=*` tag (like [here](https://www.openstreetmap.org/way/334108966#map=19/52.23164/21.01970)), so it's probably the same more general problem as with https://github.com/gravitystorm/openstreetmap-carto/pull/3005#issuecomment-363002147.

polarbearing:
The Golden Litfass Column Award goes to everybody :-)

img_20180321_litfassplatz

Greetings from the [Litfaßplatz](https://www.openstreetmap.org/way/100950875) in Berlin...

PR #3139, opened 2018-03-21
Show aerialway name

### LINKED ISSUE #1807: show names of aerialways
http://overpass-turbo.eu/s/bfz - aerialways with names

Penegal:
This is the version with `text-dy` decreased by 2 px:
image

Resolves #1807.

### DIFF
--- a/aerialways.mss
+++ b/aerialways.mss
@@ -14,6 +14,23 @@
       dash/line-color: black;
       dash/line-dasharray: 0.4,13;
       dash/line-clip: false;
+      [zoom >= 17] {
+        text-name: "[name]";
+        text-fill: #666666;
+        text-size: 10;
+        text-dy: 4;
+        text-spacing: 900;
+        text-clip: false;
+        text-placement: line;
+        text-min-distance: 18;
+        text-face-name: @book-fonts;
+        text-halo-radius: @standard-halo-radius;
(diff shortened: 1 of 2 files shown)


kocio-pl:
Please fix the name of the commit similar/the same as the updated PR name - see https://github.com/gravitystorm/openstreetmap-carto/pull/3129#issuecomment-374306276.

Penegal:
Here you are.

kocio-pl:
Could you post a test rendering on z17+ and z19+? The rendering should differ a bit according to the code and it's good to test it.

Penegal:
z17:
image

z19:
image

kocio-pl:
Thanks, I tested it elsewhere and it also works great!

geozeisig:
Show aerialway name earlier on z16. [Example for testing](https://www.openstreetmap.org/#map=16/47.2474/11.9440)

kocio-pl:
Makes sense, maybe even earlier (z15+ sounds good for me, z14+ might work too), but it needs to be tested on the places where this could hurt (probably in the cities). On the other hand if the way is too short, the name would simply not be displayed, because it would not fit, so it's safer than rendering node labels.

Could you test it a bit?

geozeisig:
+1. Sorry but I have no test system.

PR #3145, opened 2018-03-25
Documentation for smaller area download/testing

james2432:
How to get a .pbf by downloading less data than an entire state/provice/country and import it


kocio-pl:
Related to #2884 - another data source.

pnorman:
There shouldn't be a need to convert to PBF.

kocio-pl:
I guess this hint is not needed then, since it's mainly about converting, so I will close it now without merging. However thanks for sharing your experience!

PR #3146, opened 2018-03-26
Move highway=mini_roundabout rendering to higher zoom level

### LINKED ISSUE #2020: render mini-roundabouts less prominent at lower zoom levels
[Mini-roundabouts](http://wiki.openstreetmap.org/wiki/Tag:highway%3Dmini_roundabout) render with the same size symbol on all zoom levels from 16. At least at that zoom level, if not even at 17, the symbol looks much too obtrusive and looks rather like a marker for a city. It is even bigger than the connecting streets.

In my opinion a mini roundabout doesn't need to be rendered at all until level 18 or 19. At least it should be reduced to a simple dot or something else with a smaller size.

[Example](http://www.openstreetmap.org/node/269063436#map=16/51.0347/7.0505)

mini_roundabout 16

kocio-pl:
Resolves #2020.

I think moving from z16+ to z17+ should be enough:

[Standard example](https://www.openstreetmap.org/node/3404845697#map=16/52.2216/21.0927):
z16, before
jplu8dsj

z16, after
d3ptsq r

z17
1d_nr78p

[Extended example](https://www.openstreetmap.org/node/2730785546#map=16/52.1516/21.0551):
z16, before
g6ctuvyh

z16, after
2nqv3gpl

z17
hpx4w q

### DIFF
--- a/amenity-points.mss
+++ b/amenity-points.mss
@@ -1217,7 +1217,7 @@
     }
   }
 
-  [feature = 'highway_mini_roundabout'][zoom >= 16]::highway {
+  [feature = 'highway_mini_roundabout'][zoom >= 17]::highway {
     marker-file: url('symbols/mini_roundabout.svg');
     marker-placement: interior;
     marker-clip: false;


matthijsmelissen:
:+1:.

Also, this is likely a tagging error: as far as I know [mini-roundabouts](https://www.safedrivingforlife.info/blog/big-trouble-mini-roundabouts) (or at least what the British call mini-roundabouts) don't exist in Poland, as far as I know.

PR #3149, opened 2018-03-27
Make highway=traffic_signals icon less obtrusive

kocio-pl:
Follow up to #3001.

I have noticed that traffic_signals icon is 16 px big, so I tried to make it standard (14 px) and a bit lighter (80%). I hope it's still visible, but at least a bit less obtrusive:

[Big roundabout](https://www.openstreetmap.org/#map=18/52.26881/20.98655)

Before
pgwuepoz

After
g_5sydbc

[Named ones](https://www.openstreetmap.org/node/4038574025#map=17/50.89891/15.68133)

Before
2dvxbypd

After
atkjf8fk


polarbearing:
I'm in favour of the move in general.

On the fine-tuning side, the new icons have a slightly blurred appearance. Are they pixel-aligned?
In the svg, I wonder if all the metadata are redundant. 

what exactly does the 'mix' in ```mix(#0a0a0a, #ffffff, 80%);```, in comparison to transparency, or having the colour readily mixed?

kocio-pl:
They are as pixel aligned as it's possible with this shape (not too much), I guess the main source of his impression might be the lighter color. Mixing colors has nothing to do with transparency, AFAIK, it's just 20% whiter.

polarbearing:
I'll have a look at the svg later today. For the colour, I have recently used this [colour picker](https://www.w3schools.com/colors/colors_picker.asp) which calculates the brightness for the same hue for you. Thus your intention would translate into #333333 or #404040.

polarbearing:
The icon was 5x with its left/right edges at 4.5 px, which IMHO caused the blur as mapnik has to render this with a halftone pixel. Also the canvas was still 16, not sure what that have caused.

Attached are two versions that have a 6px wide icon, with the light circles 4px and 3.5px, respectively. Would you kindly test them? They are code-cleaned already.

 traffic_light_14_w6_lights4.svg [gist](https://gist.github.com/polarbearing/0c47fb456058d27927b9024382dcd1f7)

 traffic_light_14_w6_lights3.5.svg [gist](https://gist.github.com/polarbearing/e394a6fa835254e39a01a1d739924902)

kocio-pl:
3.5 px:
vu1aqba1

4 px:
00hinsrt

kocio-pl:
> I'll have a look at the svg later today. 

Thanks for the icons! 

Maybe you could review all the icons and make them clean code-wise?

> For the colour, I have recently used this colour picker which calculates the brightness for the same hue for you. Thus your intention would translate into #333333 or #404040.

Sure, but during testing it's better to have easy to slide option. 

Another field where mixing makes sense is when you want to keep the link between two colors to make it easier to understand how they relate to each other - for example:

https://github.com/gravitystorm/openstreetmap-carto/blob/5a3bc199a3042a36eaf4d2b1b340d5c2fed7bab1/landcover.mss#L558-L561

However in this case I will just change the color after testing into some absolute value, since it's not related to anything else, it just should be not too dark but still readable.

polarbearing:
The **3.5 px** lights work best for me. Both new icons render crisp outside. The 3.5 px has some blur within the lights, naturally, but for the 4 px the frame gets too narrow, which leads to other optical illusions. 

For the mix colours, fine for testing of course. For the final code I'd favour fixed values for performance reasons, I'd suspect that mapnik would do the multiplication for each symbol. 
You could keep the explanation at the colour definition section as a comment. 

Ah, and don't mention parking_space when I am in the room ;-)

I would not have the time currently to clean old svg code, but for new and frequent ones optimisation might be a good idea. Basically I open the svg in a text editor and remove all the identifiers, metadata and so on.

kocio-pl:
> The 3.5 px lights work best for me. Both new icons render crisp outside. The 3.5 px has some blur within the lights, naturally, but for the 4 px the frame gets too narrow, which leads to other optical illusions.

This is exactly what I think too.

> For the final code I'd favour fixed values for performance reasons, I'd suspect that mapnik would do the multiplication for each symbol.

Somebody could test it, but yes, I will make a fixed value once we find which is best- with 3.5 px version I think we could go even lighter.

> Ah, and don't mention parking_space when I am in the room ;-)

Sorry, lack of time to find an example unrelated to my current cases. :smile: 

> I would not have the time currently to clean old svg code, but for new and frequent ones optimisation might be a good idea. Basically I open the svg in a text editor and remove all the identifiers, metadata and so on.

Maybe it's an idea for a simple tool for sanitizing icon code automatically? Of course it should be checked in rendering after that.

kocio-pl:
Version 3.5 px, different lightness - I would choose 75%:

80%
td uri54

75%
x4npvm7x

70%
wb94syxr

60%
qa_ojdz7

polarbearing:
Even the lighter versions work for me. Should we test that on other road colours as well?

kocio-pl:
Sure, I wanted just start with something sane. Which % version would you like to test?

kocio-pl:
75% version - I couldn't spot them on higher class roads:

mbhefp8b
blseb0re

kocio-pl:
@polarbearing What do you think about it? What color would you like to test?

amenity-points.mss
```diff
@@ -124,8 +124,8 @@
   }
 
   [feature = 'highway_traffic_signals'][zoom >= 17] {
-    marker-file: url('symbols/traffic_light.16.svg');
-    marker-fill: #0a0a0a;
+    marker-file: url('symbols/traffic_light.svg');
+    marker-fill: mix(#0a0a0a, #ffffff, 75%);
```
pnorman: Why are we mixing instead of just setting a different colour?

kocio-pl: It's only a tool to make color testing much easier. Once we set the proper value, I will change it into some real value.

kocio-pl:
> If you did it with 3px circles, I'd go for 13px high, 5px wide.

@polarbearing Would you like to try that?

kocio-pl:
I've made the icon 13x5 px with 3 px holes and it looks nice to me with 70%:

gqa1bkzo
c7ci1wm_
im n1qql

Issue #961, opened 2014-09-19
osm-carto for TagInfo


sabas:
The new feature of Taginfo allows to give information about usage for a selected tag in a project
openstreetmap-carto is a natural fit
http://wiki.openstreetmap.org/wiki/Taginfo/Projects

I did a list some time ago, perhaps it's useful https://github.com/sabas/doesitrender/blob/gh-pages/taglist.md

nebulon42:
@giggls proposed an alternative in #2614. To manually create the file and then make sure that each change updates it. Still there is the risk that the file gets outdated. Would this be worth a try?

giggls:
I don't think that keeping such a file in line with the current version of the style would be a problem, if pull requests would require changing this file as well. The main effort would certainly be the creation of the initial version.

imagico:
A few thoughts on the matter:

* to have this would be good and IMO worth a bit of work in maintenance when making changes to the style.
* it would be an obvious and fairly trivial byproduct of generating a map style legend so it certainly makes no sense to separately maintain a legend and a taginfo list.  Since we currently have no legend this is not really practically relevant at the moment of course.
* having a plain list of tags a taginfo json file is produced from via script is probably easier to maintain than writing a json file by hand.
* it might make maintenance easier if we'd structure the list by layer - which would of course require removing duplicates in processing.

giggls:
I like the Idea of having a Text file which is ordered by layer and a script which will generate the Taginfo file. In this case it would even be possible to add tag combinations which are currently not supported in Taginfo project format.

pnorman:
I'm iffy on if this is worth the maintenance burden. If we do do it, we should make sure that we can build a legend from it, and that we're not doing it as JSON, because that will lead to difficult to solve merge conflicts.

kocio-pl:
While it would be nice to have osm-carto on Taginfo, this is beyond our manpower and nobody seems to be interested to even start, so I close it now. If there will be some automatic solution or somebody ready to take care of it, it can be reopened.

ImreSamu:
My "poor man" suggestions: just add the  osm keys:

not perfect  - Proof Of Concept : 

```bash
$ cat project.mml | grep -oP "tags[^']*'.+?'"  | cut -d"'" -f2 | cut -d"=" -f1 | sed 's/"//g' | sort -u
addr:unit
advertising
capital
denomination
ele
emergency
entrance
ford
generator:source
height
highspeed
iata
icao
indoor
[...]

Issue #3012, opened 2018-01-10
Produce artifacts from CI with Mapnik XML


pnorman:
cc @lonvia

With the increasing problems of getting npm and carto, some have asked about having artifacts generated from CI with the Mapnik XML

kocio-pl:
Could you be more verbose, please? I don't get what's this ticket about and what kind of problems are you mentioning.

pnorman:
> Could you be more verbose, please? I don't get what's this ticket about and what kind of problems are you mentioning.

Requiring a npm install is a reasonably high barrier if you only want to use the style, not develop it. If all someone needs is the XML, they should be able to get that as an artifact.

lonvia:
I was the one asking about artifacts, so let me elaborate. `npm` does not ship as a package with Debian anymore. The manual installation instructions for npm ask you to pipe a script from the internets into a `sudo bash`, which is a no-no for obvious reasons. In the end I settled for [nvm](https://github.com/creationix/nvm) to set up a current nodejs and npm without requiring root. It's still quite a bit of work, so precompiled XML for those of us who just want to render would be awesome.

I've looked a bit into what Travis can do. Unfortunately, it does not host artifacts itself like appveyor does. So you need to provide some hosting. The most promising might be the uploading to [Github releases](https://docs.travis-ci.com/user/deployment/releases/). Having xml for the releases would be more than enough for non-devs.

kocio-pl:
Any ideas about how it should be done? Probably some simple script to be launched before every release or just a one liner in `RELEASES.md`?

HolgerJeromin:
Another way would be letting travis checkin the results to an repository. Not kidding

https://github.com/open62541/open62541/blob/master/tools/travis/travis_push_release.sh
produces https://open62541.org/releases/

kocio-pl:
Sounds great to me. Anybody willing to prepare the code?

HolgerJeromin:
Its kind of hacking github for serving files without the need of the full history. 
The downside are the massive checkins which floods the history of the repository. Thats why it seems a good idea to use another repository and reset it from time to time: 
https://github.com/open62541/open62541/issues/108

pnorman:
:-1: on having travis push to a git repo.

We could either have Travis push to github releases, another artifact location, or look at something like circleci which has artifact support.

matkoniecz:
I am against giving push access to external entities.

Separate repo may be a better idea.

PR #3151, opened 2018-03-28
openstreetmap-carto project file for TagInfo

ImreSamu:
**STATUS (2018May08) :  As I understand and accept that the maintainence problem is not optimal for my "toy code", so I will move my code to the separate repo/project, and I will close this pull requests in the next days.**

----  

background:   #961 

This is a minimalistic approach: only the osm keys parsed and reported.

I have created a separated directory `./taginfo-project/`

New files:
* `./taginfo-project/README.md`    : basic readme
* `./taginfo-project/generate-taginfo-project-file.py` 
* `./taginfo-project/taginfo-openstreetmap-carto.json`   -> this is the result , for the taginfo! 

Status:
- [x] First working version 
- [x] add ARRAY parsing ( `tags @> ARRAY['building',...] ` )
- [x] the result imported to the Taginfo dev system, and everything is OK.
(https://github.com/gravitystorm/openstreetmap-carto/blob/master/CONTRIBUTING.md#sql-style-guidelines)
- [x] integrate to the workflow ( .travis ? )


kocio-pl:
> integrate with the CONTRIBUTING.md#sql-style-guidelines

I'm against adding anything to do by hand, especially if it could be easily automated. I still don't have the idea how to make it efficiently after each release, though. I think that while we can include this script, as it is closely related to our code, but running this tool is another story.

ImreSamu:
@kocio-pl : 
>integrate with the CONTRIBUTING.md

I need to re-think,  but you can expect some extra documentation line ...


# Testing 
 

temporary "taginfo-project"   test

* http://as-af.taginfo-dev.opengeodata.hu/projects/osmcartov2#overview
* http://as-af.taginfo-dev.opengeodata.hu/projects/osmcartov2#tags

please help:
* the "Objects" detections is correct? ( node,way,area, realation)
* Now this is the text - but I am not satisfied, what is your suggestions?
     * `Used in the osm2pgsql database backend, see more in the github repo`
    * `Used as a hstore tags-> in the database backend, see more in the github repo`
* Please test the "More->" part  (  github search )

kocio-pl:
> I need to re-think, but you can expect some extra documentation line ...

Sorry, I've made a mistake. It's OK to make the style more strict and consistent (it might be even preferable). I meant RELEASES.md, which you didn't even mentioned... 

Still it means that this script should be triggered somehow and the best moment seems to be release time.

taginfo-project/generate-taginfo-project-file.py
```diff
@@ -0,0 +1,174 @@
+#!/usr/bin/env python3
+
+# -------------------------------------------------------------------------------
+#
+# This code generate a taginfo project list file (see more https://wiki.openstreetmap.org/wiki/Taginfo/Projects )
```
kocio-pl: generate->generates

kocio-pl:
I think JSON file should not be included in the final version - it will be automatically generated only when needed. It should be always listed in `.gitignore` file.

ImreSamu:
> I think JSON file should not be included in the final version -

I don't think I've understood all that,  
Could you clarify why don't want the final JSON file? ( = taginfo project file )

According to the ["Taginfo/Projects/Integration into Taginfo"](https://wiki.openstreetmap.org/wiki/Taginfo/Projects)
_"Taginfo will periodically (usually daily) poll the project file and add the information to its own database."_

And I have planned the project file should be here : 
*  https://raw.githubusercontent.com/gravitystorm/openstreetmap-carto/master/taginfo-project/taginfo-openstreetmap-carto.json   

And I have not added a timestamp inside the JSON,  so my `git status` don't detect as a change if the content is the same.

taginfo-project/generate-taginfo-project-file.py
ImreSamu: Thank you,  fixed!

kocio-pl:
I guess this file will be created by Travis on next merge, isn't it?

kocio-pl:
The script and documentation should be placed inside the `scripts` directory, just like [lua](https://github.com/gravitystorm/openstreetmap-carto/tree/master/scripts/lua).

ImreSamu:
>I guess this file will be created by Travis on next merge, isn't it?

I am using the Travis only for checking.
* testing:  my code is running? 
* and a simple `git diff` - what has been changed in the taginfo project file. 

if you not need, you can remove.

imho:  some human checking is useful because my code is not _well_ tested.  ( code parsing  is hard )





>The script and documentation should be placed inside the scripts 

ok, no problem.

like this?

old files                                          |  new files
---------------------------------------------------| ------------------
./taginfo-project/generate-taginfo-project-file.py | ./scripts/taginfo-project/generate-taginfo-project-file.py 
./taginfo-project/README.md                        | ./scripts/taginfo-project/README.md


imho:  than better if we move the taginfo JSON  project file to the root.  ( so no need a special folder ) 

my suggestion:

old project file                                   |  new project file 
---------------------------------------------------| ------------------
./taginfo-project/taginfo-openstreetmap-carto.json | ./taginfo-openstreetmap-carto.json 


The taginfo expecting a "static/fix url" for the project file, so it can be any url - as you like.
so If you don't want to store the project JSON file in the repo, no problem.

kocio-pl:
>like this?
>old files | new files

Yes, this is what I meant.

> imho: than better if we move the taginfo JSON project file to the root. ( so no need a special folder )

This would be proper location IMO, if we want to store it in this repo. However I'm still not sure how do you want to run the script?

HolgerJeromin:
> I guess this file will be created by Travis on next merge, isn't it?

We can not access travis files, so Travis can not generate the really used files. Ref #3012

ImreSamu:
>However I'm still not sure how do you want to run the script?


* now I will generate for the current taginfo JSON file - with this pull request.
    * after accepted 
         * I try to add the  url to the https://github.com/taginfo/taginfo-projects/blob/master/project_list.txt

* every release 
    * re -RUN the python script 
    * and IF  the `git status` shows a difference ( and it seems correct ) 
          *  THEN  `git commit` the new taginfo JSON project file. 

*  AND every day the taginfo will process and show the info.

>IMO, if we want to store it in this repo.

what is the alternatives?

HolgerJeromin:
> However I'm still not sure how do you want to run the script?

Imo a maintainer should generate the chicago preview png and this json file locally and commits both to this repo if there is a major change

kocio-pl:
I said before that I don't like to make more manual steps if they can be done automatically - releasing is already quite complicated and checking the script output differences would be even worse. 

This script could be run from some other place and I guess even automatically checking the new release is possible, but as simple thing as running it every night for example would do too.

ImreSamu:
@kocio-pl :  

I have finished.

Ready to final review, merge, adapt, delete, reject, modify ... as you like 

>This script could be run from some other place and I guess even automatically 

Sorry,  I can't help on this topic,

kocio-pl:
Thanks. Unfortunately I can't do the merge here, since I'm not even a programmer. Some coder has to do the code review. Anybody willing?

scripts/taginfo-project/generate-taginfo-project-file.py
```diff
@@ -0,0 +1,183 @@
...
+                    osmtype='area'
+            else:
+                # If no Geometry type - we try to guess the type.
+                if 'planet_osm_point' in ds_table.lower():
+                    osmtype='node'
+                elif 'planet_osm_polygon'  in ds_table.lower():
+                    osmtype='area'
+                elif 'planet_osm_line'  in ds_table.lower():
+                    osmtype='way'
+                elif 'planet_osm_ways'  in ds_table.lower():
```
pnorman: We don't use the ways table, and can't without a schema change.

scripts/taginfo-project/generate-taginfo-project-file.py
```diff
@@ -0,0 +1,183 @@
...
+                    osmtype='way'
+                else:
```
pnorman: planet_osm_roads hasn't been processed

scripts/taginfo-project/generate-taginfo-project-file.py
```diff
@@ -0,0 +1,183 @@
...
+
+
+
+#
+# Parsing "project.mml" file for the HSTORE keys (  tags-> )
+#
+
+with open( cartocss_project_file , 'r') as f:
+  newf = yaml.load(f.read())
+f.closed
```
pnorman: Doesn't the with block make sure it's closed?

scripts/taginfo-project/README.md
```diff
@@ -0,0 +1,51 @@
...
+* Be carefull with the SQL comments or use as a simple hack:  ` /* tags->'wetland' */ `
+* This code tested only on Ubuntu Linux
+* Check the result!
+
+### Valid examples for parsing
+* `tags ? 'wetland'"`
+* `tags @> '"generator:source"=>wind'`
+* `tags @> 'capital=>yes'"]`
+* `tags -> 'leaf_type'`
+* `tags -> 'wetland'`
```
pnorman: There's already an example of this type one line up

scripts/taginfo-project/generate-taginfo-project-file.py
```diff
@@ -0,0 +1,183 @@
...
+#  tags->'leaf_type'
+#  tags @> '"generator:source"=>wind'
+re_tags_one     = re.compile(r"[^a-zA-Z0-9_]tags[^'^)^\[^\]]*'.+?'")
+
+
+# ----------------------------------  Examples --------------------
+#  tags -> ARRAY['wheelchair',ramp:wheelchair']
+#  tags ?& ARRAY['wheelchair',ramp:wheelchair']
+#  tags ?| ARRAY['wheelchair',ramp:wheelchair']
+re_tags_array = re.compile(  r"[^a-zA-Z0-9_]tags\s*[@\?-][>&\|]\s*[aA][rR][rR][aA][yY]\[.+?\]"  )
```
pnorman: Use [re.IGNORECASE](https://docs.python.org/3.1/library/re.html#re.IGNORECASE) as a flag for re.compile to simplify the expression

.travis.yml
```diff
@@ -1,22 +1,32 @@
 language: node_js
 dist: trusty
-sudo: false
+sudo: required
```
pnorman: This will have a significant impact on how long it takes us to get a Travis instance

.travis.yml
```diff
@@ -1,22 +1,32 @@
...
 install:
+  - sudo apt-get install -y python3-yaml
   - npm install carto@$CARTO
   - mkdir -p data/world_boundaries data/simplified-land-polygons-complete-3857 data/ne_110m_admin_0_boundary_lines_land data/land-polygons-split-3857
   - touch data/simplified-land-polygons-complete-3857/simplified_land_polygons.shp data/ne_110m_admin_0_boundary_lines_land/ne_110m_admin_0_boundary_lines_land.shp data/land-polygons-split-3857/land_polygons.shp
 script:
   # We're using pipes in the checks, so fail if any part fails
   - set -o pipefail
+  # Check the Taginfo-project
+  - cd ./scripts/taginfo-project && ./generate-taginfo-project-file.py  && cd ../..
```
pnorman: it should be callable as `scripts/taginfo-project/generate-taginfo-project-file.py`

.travis.yml
```diff
@@ -1,22 +1,32 @@
...
+  - git diff taginfo-openstreetmap-carto.json
```
pnorman: It's better to generate to a different file and diff the existing taginfo-openstreetmap-carto.json against the new one, or have the script write to stdout and use `<()`

matthijsmelissen:
Thanks for proposing this, @ImreSamu.

The thing I'm worrying about is that this adds fairly complicated code that will break easily under many unsuspected code. Also it is quite a lot of code that achieves relatively little (it only parses the keys). Moreover, checking whether a key is used in the mml does not give any guarantees that the key is actually displayed.

So I doubt this is the right way to go. I'm again inclined now towards manually maintaining a list of interpreted tags (which might also serve as a way of specification for us).

ImreSamu:
@matthijsmelissen 

Thank you for you feedback.

Imho: My proposal is not guaranteed to be optimal or perfect, but sufficient for my immediate goals. Just creating a "minimal" taginfo project file with a minimal "business value" for the OSM community.

**As I understand and accept that the maintainence problem is not optimal for my "toy code", 
so I will move my code to the separate repo/project, and I will close this pull requests in the next days.**

Now I am testing my code with other carto forks ( osmfr-cartocss, openstreetmap-carto-hiking-cz,  ... )

From the "minimal" taginfo project files I can create some new "business value" :
for example:

`osmfr-cartocss` extra keys ( compare to this project ) : 
```
CEMT aerodrome alt_name basin boat bus 
capacity:disabled conveying conveyor_dir craft crossing crossing:bollard ele:local footway fuel:lpg golf 
heritage importance incline information int_name is_capital level line location 
maritime medical motorboat mountain_pass name:fr network organic 
parking piste:difficulty piste:type post_office:type ramp:wheelchair ref:FR:LaPoste ref:INSEE room 
school:FR shelter_type short_name short_name:fr sloped_curb stars 
tactile_paving tidal type:RATP 
wall water wheelchair 
```

`carto-hiking-cz` extra keys:
```
kct_blue kct_green kct_red kct_yellow network official_name:cs osmc:symbol
````


>So I doubt this is the right way to go. I'm again inclined now towards manually maintaining a list of interpreted tags 
>(which might also serve as a way of specification for us).

My proposal is just a temporary ( quick&dirty )  solution. It is not perfect  and I accept that it has no 'business value' for the current project members. 
So I will be very happy If somebody create and maintain an improved version of `taginfo project file` with more business value.

Issue #2440, opened 2016-11-14
Rendering of leisure=beach_resort


doktorpixel14:
[Leisure=beach_resort](http://wiki.openstreetmap.org/wiki/Tag:leisure%3Dbeach_resort) is an important tag in touristical regions. The wiki tells us, that we have to tag this instead of natural=beach, if the beach area is managed. Because of that, there is a white hole on the map in some coastal areas and that doesn't look very good. Maybe you could use the same colour like on natural=beach, but with an parasol symbol in the center.

dieterdreist:
2016-11-14 21:42 GMT+01:00 doktorpixel14 <notifications@github.com>:

> The wiki tells us, that we have to tag this instead of natural=beach,



that's orthogonal. You'll tag the beach as natural=beach and a beach resort
as leisure=beach_resort, and both tags can be valid at the same spot.

doktorpixel14:
> "Use natural=beach for areas of sand next to the sea or to water **which are not managed.**"

So should we change this in the Wiki? Because in my opinion, it sounds like you can't use both tags in one place.

dieterdreist:
2016-11-15 19:13 GMT+01:00 doktorpixel14 <notifications@github.com>:

> "Use natural=beach for areas of sand next to the sea or to water *which
> are not managed.*"
>
> So should we change this in the Wiki?
>

The natural=beach tag has no such limitations and even explicitly mentions
spatial overlap of beaches and beach ressorts:
https://wiki.openstreetmap.org/wiki/Tag:natural%3Dbeach

doktorpixel14:
Okay, so we clarified that, but I think, that it would be still a good idea to render this tag ;-)

d1g:
> instead of natural=beach, if the beach area is managed

It says opposite: "Use additional objects tagged with natural=beach and name=* for entire beaches."

> So should we change this in the Wiki? Because in my opinion, it sounds like you can't use both tags in one place.

Wiki never said "instead", please read carefully.

doktorpixel14:
> It says opposite

Yes, after dieterdreist changed it :)

doktorpixel14:
![beach_resort](https://user-images.githubusercontent.com/23462396/32106477-13dde288-bb2c-11e7-81e7-68ad5e42448c.png)
What do you think of this symbol? The water is based on the already existing "swimming" icon.

kocio-pl:
[...]

Issue #2452, opened 2016-11-20
Better maintainable whitelists


nebulon42:
I decided to open a separate issue for that. It is related to #2444 and an unmaintainable white list of shops was the reason that #2415 got merged.

So we need better maintainable whitelists. One idea would be to put them into the database. One then would only have to maintain the content of the tables and the SQL would be automatically adjusted. This of course open up questions of how the rendering database gets updated with the white lists.

Basic proof of concept:

````
CREATE TABLE barrier_whitelist (feature TEXT PRIMARY KEY);

INSERT INTO barrier_whitelist VALUES
  ('chain'),
  ('city_wall'),
  ('embankment'),
  ('ditch'),
  ('fence'),
  ('guard_rail'),
  ('handrail'),
  ('hedge'),
  ('kerb'),
  ('retaining_wall'),
  ('wall');
````
````
        (SELECT
            way, COALESCE(historic, barrier) AS feature
          FROM (SELECT way,
            ('barrier_' || (CASE WHEN barrier IN (SELECT feature FROM barrier_whitelist) THEN barrier ELSE NULL END)) AS barrier,
            ('historic_' || (CASE WHEN historic = 'citywalls' THEN historic ELSE NULL END)) AS historic
            FROM planet_osm_line
            WHERE barrier IN (SELECT feature FROM barrier_whitelist)
              OR historic = 'citywalls'
              AND (waterway IS NULL OR waterway NOT IN ('river', 'canal', 'derelict_canal', 'stream', 'drain', 'ditch', 'wadi'))
          ) AS features
        ) AS line_barriers
````

In pgAdmin it works well, with Mapnik I have the problem that some initializing is done and I run into:
````
Postgis Plugin: ERROR:  column "way" does not exist
LINE 1: SELECT ST_SRID("way") AS srid FROM barrier_whitelist WHERE "...
                       ^
in executeQuery Full sql was: 'SELECT ST_SRID("way") AS srid FROM barrier_whitelist WHERE "way" IS NOT NULL LIMIT 1;'
  encountered during parsing of layer 'line-barriers' in Layer
````
I'm not sure if we can get around this.

Other ideas?

imagico:
> I'm not sure if we can get around this.

According to mapnik docs:

https://github.com/mapnik/mapnik/wiki/PostGIS

specifying an srid parameter for the postgis data source of the layer _will avoid an extra database query for knowing the srid of the table_.

[...]

PR #3153, opened 2018-03-29
Removing objects without background/outline from landcover data layer

kocio-pl:
A fix triggered by finding this bug: https://github.com/gravitystorm/openstreetmap-carto/issues/2440#issuecomment-377272725.

Objects with no area/outline rendering should not be selected in the landcover data layer. It might go unnoticed most of the time - for example it's hard for me to imagine firepit area with any kind of background, like grass - but if this area is tagged also with something using landcover features, it can lead to lack of the background/outline of the second feature.


matthijsmelissen:
I'm not sure if I fully understand the need for this PR?

Also, why did we have these features in the table in the first place, if we don't need them?

kocio-pl:
It's just code cleaning. They were probably added because some similar objects were there - or maybe in a hope that soon some kind of area/outline rendering will be defined, but they are not styled currently.

This is why #2452 would be useful - I wouldn't miss/introduce them if the change was in the clearly named whitelist instead of large SQL statement inside even bigger layer.

matthijsmelissen:
Ok, clear!

kocio-pl:
Yes. On the other hand `tourism=attraction` is there [for ~4 years](https://github.com/gravitystorm/openstreetmap-carto/blame/45dbf13127a938226481d73a4ad696582d9073f8/project.yaml#L103) at least.

PR #3154, opened 2018-03-29
Adding rendering for leisure=beach_resort

### LINKED ISSUE #2440: Rendering of leisure=beach_resort

doktorpixel14:
Fixes #2440 

Added rendering for [leisure=beach_resort](https://wiki.openstreetmap.org/wiki/Tag:leisure%3Dbeach_resort) with an own icon. It is visible from zoom level 16 and labeled from level 17. This is the same scheme as used by tourism=picnic_site.

This is how it looks on ways at Z16 (in [Ulcinj, Montenegro](https://www.openstreetmap.org/#map=19/41.91946/19.21354)) ...
beach_resort_16

... and Z18
beach_resort_18

This is how it looks on nodes at Z16 (in [Almolonga, Guatemala](https://www.openstreetmap.org/node/5033149094)):

beach_resort node z16

and on z17:
beach_resort node z17


kocio-pl:
Hi, could you add the links and rendering for both types (nodes and ways with labels) or you have some trouble and need some help?

doktorpixel14:
Sorry, I've forgot it over the past week... But I updated the original post now.

kocio-pl:
No problem! Thanks for the code and a clean icon design.

PR #3157, opened 2018-03-31
Move barriers to higher zoom level

### LINKED ISSUE #323: Gates are too prominent
Gates in fences are often much to present on the map. In my region I mapped several fences with smaller gates in it and they are very annoying on smaller zoom levels. I thing it would be better not to render them in zoom-levels lower than 18.
Same with information boards. I've mapped plenty of them and I believe all those small info-board belong into the osm database but rendering them on zoom-level lower than 18 seems also to much.

For example see this area: http://www.openstreetmap.org/#map=16/51.7123/10.5129

kocio-pl:
Fixes #323

- moving barriers from z16+ to z17+:
  - gate
  - lift_gate
  - swing_gate
  - bollard
  - block
  - log

[Example place with gates, z16](https://www.openstreetmap.org/#map=16/52.2223/20.9062)
Before
nfawmamh

After
i_bplp2q

[Example place with bollards, z16](https://www.openstreetmap.org/#map=16/52.2560/20.9835)
Before
ndz77k9j

After
blsw9wd_

[Example place with gates and lift gates, z16](https://www.openstreetmap.org/#map=16/46.3637/11.0327)
Before
qtc tyij

After
sew4co2j

### DIFF
--- a/amenity-points.mss
+++ b/amenity-points.mss
@@ -1224,15 +1224,15 @@
   }
 
   [feature = 'barrier_gate']::barrier {
-    [zoom >= 16] {
+    [zoom >= 17] {
       marker-file: url('symbols/gate.svg');
       marker-placement: interior;
       marker-clip: false;
     }
   }
 
-  [feature = 'barrier_lift_gate'][zoom >= 16]::barrier,
-  [feature = 'barrier_swing_gate'][zoom >= 16]::barrier {
+  [feature = 'barrier_lift_gate'][zoom >= 17]::barrier,
+  [feature = 'barrier_swing_gate'][zoom >= 17]::barrier {
     marker-file: url('symbols/liftgate.svg');
     marker-fill: #3f3f3f;
     marker-placement: interior;
@@ -1242,7 +1242,7 @@
   [feature = 'barrier_bollard'],
   [feature = 'barrier_block'],
   [feature = 'barrier_log'] {
-    [zoom >= 16] {
+    [zoom >= 17] {
       marker-width: 3;
       marker-line-width: 0;
       marker-fill: #7d7c7c;


dieterdreist:
this is about the general problem that we are context agnostic in this style. We don’t know whether these are important gates in a low density setting in the outskirts of nowhere (where missing the information can mean walking back half an hour and more) or one out of many gates for short private footpaths in a residential area. Yes, we would not want to show the latter early, but z16 is already late for the former. If we have to keep it simple we must accept some compromise.

kocio-pl:
There are few important reasons which make me think this change is needed:

1. We have more data added and what was not a problem in the beginning (just show whatever is added), now became to be source of a mess. We have more well-mapped areas ("WMA") and as you can see in case of Cles (Italy) even typical medium-sized town, which I think was a target model when inventing and rendering many tags, can look bad now, just because somebody did her job well. This is not kind of a feedback mappers should have - we should rather encourage them to add more, because we can handle it properly.

2. When making universal (not thematic) style the main slogan should be "mind the scale", because there are too many things to prioritize. The only easily observable thing is a typical size of a feature and this is a rule of thumb when I think about initial zoom level. Point barriers are usually small (unlike walls or fences) and should not be shown as early as police station icon or a theater. 

3. This is not an outdoor map - we have plenty of them and I don't see they care for gates at all ( https://opentopomap.org/#map=16/46.36425/11.03320 ) and when they do, they skip many other POIs, selecting sustenance objects only ( http://www.hikebikemap.org/?zoom=15&lat=46.36352&lon=11.03347&layer=HikeBikeMap ).

4. This is not a routing service. If somebody is planning a route, just an image on the map does not tell him if it's permanently open, closed or there are only some times when one can pass. I'm thinking about adding some basic access rendering too, but it will not suffice. I also think that mappers could use more access tags for roads, which would make it more evident if a given part is available or not than just a barrier node.

PR #3158, opened 2018-04-01
More vertical objects rendering and tuning

### LINKED ISSUE #3275: Render watchtower like an observation tower
Could we have the `man_made=tower` + ` tower:type=watchtower` objects rendered like a `tower:type=observation` tower, please?

There was a suggestion here https://github.com/gravitystorm/openstreetmap-carto/pull/2671#issuecomment-377665578 but right now they are being rendered as some kind of unknown type:

For example, https://www.openstreetmap.org/node/4306889882

While having this would be nicer and more realistic:

kocio-pl:
Follow up to https://github.com/gravitystorm/openstreetmap-carto/pull/2671#issuecomment-373046568
Fixes #3275

- adding rendering:
  - `tower:construction=dish`
  - `tower:construction=dome`
  - `tower:type=cooling`
  - `tower:type=watchtower` with `tower:type=observation` icon 
  - `man_made=chimney`
  - `man_made=communications_tower`
- hiding verticals located on roofs

Rendering with links to the example places: https://github.com/gravitystorm/openstreetmap-carto/pull/3158#issuecomment-378268404.


geozeisig:
`man_made=communications_tower` with dome so it is different from `tower:type=communication`:
tower_cantilever_communication
twice as big:
communicatiotower2

kocio-pl:
Could you publish a gist?

kocio-pl:
Does anybody know how to expand our code for height to work [with optional unit "m"](https://wiki.openstreetmap.org/wiki/Key:height#Units)?

Current code is based on ele code (where units are not possible):

https://github.com/gravitystorm/openstreetmap-carto/blob/ea769f4456cffac780596bc2f7a67b87b1e92812/project.mml#L1466

I was trying to replace this regex with something like:

```perl
^-?\d{1,4}(\.\d+)?(\s[m])?$
```

or with additional grouping: 

```perl
^(-?\d{1,4}(\.\d+))?(\s[m])?$
```

and it looks more or less OK with testing ( https://regex101.com/ ), but it doesn't work. I guess it tries to evaluate the whole expression instead of just the first group, but I don't know how to fix it.

kocio-pl:
Rendering tests (with initial zoom levels):

[`tower:construction=dish` (+`tower:type=communication`), z17](https://www.openstreetmap.org/#map=18/55.68003/12.35559)
Before
0i6meowr

After
brj7ut5b

[`tower:construction=dome`, z17](https://www.openstreetmap.org/node/1599219610#map=17/50.48043/19.08361)
Before
ik8wmzk9

After
rirwk_2d

[`tower:type=cooling`, z17](https://www.openstreetmap.org/way/233620461#map=17/52.01351/18.54435)
Before
2ydhpiqq

After
xbaniuer

[`tower:type=watchtower`, z18](https://www.openstreetmap.org/way/280795558#map=18/50.03465/19.17985)
Before
d tgcjzw

After
jepggfv3

[`man_made=chimney`, z16](https://www.openstreetmap.org/way/301418356#map=16/52.2948/20.9981)
Before
ijuk9yr_

After
0miwq41q

[`man_made=communications_tower`, z13](https://www.openstreetmap.org/node/1458857189#map=13/52.0056/20.9100)
Before
gcemlifd

After
eq2kcrn2

[Hiding `tower=communication+location=rooftop`, z14](https://www.openstreetmap.org/node/2471572605#map=14/52.2478/20.9899)
Before
bjowv zn

After
x7_6ifgj

wmyrda:
Tower look great.

For regex I would try this ^-?\d{1,3}(\.\d+\s[m])?$

One thing I can say is that changing from 1,4 to 1,3 seems reasonable hence in ele we have places located at high above 1000m while for object high I believe all of them would fit under 999 including Burj Khalifa (hence 3 digits only).
Adding \s[m] directly into the expression seems logical and just might work.

wmyrda:
I noticed addition of man_made=communications_tower to the code. Even when I like as many tags added to rendering as possible personally I find logic behind such a specification a bit redundant to man_made=tower.
eg. https://www.openstreetmap.org/node/3109071579 works just fine for that kind of object. 

Question is would it not be better to simply change all instances of man_made=communications_tower
to man_made=tower & tower:construction=freestanding & tower:type=communication which specifies just that instead of rendering similar object which would add more cpu power requirements to the rendering server for another inquiry?
Also not to lose any information that we are dealing with big object for nodes with missing height add height=101 as man_made=communications_tower specification states that object has to be of that size or higher.

kocio-pl:
I'm pretty sure that many towers/masts are mistagged. But that is - partly - due to lack of rendering them and I hope this will improve over time.

I've been thinking about it and decided that it makes sense to have `man_made=communications_tower` when you have no specific knowledge about height, but you know that it's large. We lack some tags like this in OSM, but they can be very handy.

Of course, if tagging habits and documentation will change, we can change rendering too to follow it, but at the moment it seems right for me to render them - 3970 uses and a wiki page that I can understand how it's different from regular towers is enough for me.

kocio-pl:
> For regex I would try this ^-?\d{1,3}(.\d+\s[m])?$

Doesn't work, I still think it tries to treat "m" as a number and fails:

`Postgis Plugin: ERROR:  invalid input syntax for type numeric: "230.68 m"`

wmyrda:
I used to create many regex syntaxes for nano editor with most recent being https://github.com/mikaku/Monitorix/blob/master/docs/monitorix.nanorc but as you see it is much simpler than that for sql query so that is why I said it "might" work ;)
One more idea is to completely throw out hat numeric expression. Do you still use **THEN (tags->'height')::NUMERIC**? That definitely would change whole string into numeric.

As for man_made=communications_tower I guess you are right and it is better to render it hence it is well established map feature, but it does not change my humble opinion that it is somewhat redundant to man_made=tower

kocio-pl:
`tower:type=radar` icon is currently [missing](https://wiki.openstreetmap.org/wiki/Key:tower:type), do you have an idea how should it look like? I would think of something like:

- https://proxy.duckduckgo.com/iu/?u=https%3A%2F%2Fcdn1.iconfinder.com%2Fdata%2Ficons%2Finternet-3-3%2F512%2F115-128.png&f=1
- https://cdn4.iconfinder.com/data/icons/airport-7/500/6-128.png
- https://www.iconfinder.com/icons/328194/airport_control_tower_radar_icon#size=128

turnsole80:
The first of these seems to be the most appropriate to me, the others looks
a bit cluttered. The exception would dome radars (radomes) that should keep
their bulbous head look

wmyrda:
First one actually look to me like an icon for night club, cocktails or something similar. Second I like, but it looks like airport control tower and question is whether that is always the case that radar is also located there. I just do not know that, but still I like that one the most.
Third one has to much information which may be hard to fit in 14x14 icon. The last one is a radar all right, but good for ground radar, not for radar tower. Besides towers are less likely to be mobile :)

kocio-pl:
Unfortunately it doesn't work... However there's a section about lookarounds in PostgreSQL documentation, so maybe someone could dig into this a bit more:

https://www.postgresql.org/docs/9.6/static/functions-matching.html#POSIX-CONSTRAINTS-TABLE

SomeoneElseOSM:
> tower:type=radar icon is currently missing, do you have an idea how should it look like? 

Is https://github.com/SomeoneElseOSM/openstreetmap-carto-AJT/blob/master/symbols/radartower.png any use?

meased:
I used this for the waterfall code:
``` regex
^(\d{1,4}(\.\d+)?)\s?m?$
```
This allows whitespace at the end of the line (casting to NUMERIC will ignore this). We could disallow it like this:
``` regex
^(\d{1,4}(\.\d+)?)(\s?m)?$
```

kocio-pl:
Great! I think we should use restricted version (trailing space is a mistake) and we should also restrict height to 3 digits, because [highest waterfall](https://en.wikipedia.org/wiki/List_of_waterfalls_by_height) has just 979 m. I guess even `(\s?m)` should be changed into `(\sm)`, because such units should be always used after one space.

Tomasz-W:
After countless times of trying, a radar icon proposal:

tower_type radar

I've tried with 2d and 3d, with many proportions, and it's the best what I can propose at the moment. If you have any suggestions, give me a feedback.

wmyrda:
@Tomasz-W I have mixed filling about that look for tower:type=radar as is kinda resembles the candy on the stick.
At the same time I think that we have so much problem with picturing radar towers as they come in all kind of shapes and sizes. Maybe we should show tower:type=radar only when it is accompanied by tower:construction? Having that said I would propose following:

**tower:type=radar & tower:construction=lattice**
https://img-new.cgtrader.com/items/105121/776fc696b6/large/big-radar-tower-3d-model-low-poly-rigged-max-obj-3ds-fbx-dae-tga.jpg
**tower:type=radar & tower:construction=freestanding**
https://d1o2pwfline4gu.cloudfront.net/m/t/601/6012345/a-0005.jpg
**tower:type=radar & tower:construction=dome**
https://wiki.openstreetmap.org/wiki/File:Tower_dome.svg - use the one we have for dome

Tomasz-W:
Another try - observation tower stand + radar shape:

tower_type radar skaner

wmyrda:
I really like idea the most from all the previously proposed. Could You also try icon with the same top part but with solid base like in your previous example? Just so we have variation for tower:construction=freestanding

Tomasz-W:
@wmyrda 

*  (observation tower base)
* tower_type radar skaner (halftoned-version)
* tower_type radar baza (solid base)

wmyrda:
I kinda like "observation tower base" more than "halftoned-version", but if others have different opinion I would not mind to have either as it does differentiates it more from hunting stand.

Solid base looks ok, but I think we could also try using base with 2/3 wideness of the rectangular box on the top to make it a bit wider which could give it more indication that it is not mounted on the pole.

geozeisig:
I miss `man_made=communications_tower` in Carto release v4.10.0.

symbols/towers/chimney.svg
```diff
@@ -0,0 +1,31 @@
...
+        <dc:title></dc:title>
+      </cc:Work>
+    </rdf:RDF>
+  </metadata>
+  <defs
+     id="defs113" />
+  <path
+     id="rect4495"
+     d="M 10.318359 0 A 1.741594 1.2970195 0 0 0 8.6601562 0.90039062 A 1.1655669 0.92076844 0 0 0 8.2539062 0.84179688 A 1.1655669 0.92076844 0 0 0 7.0898438 1.7617188 A 1.1655669 0.92076844 0 0 0 7.0917969 1.7851562 A 0.82779479 0.70560503 0 0 0 7 1.7792969 A 0.82779479 0.70560503 0 0 0 6.171875 2.484375 A 0.82779479 0.70560503 0 0 0 7 3.1894531 A 0.82779479 0.70560503 0 0 0 7.8125 2.6152344 A 1.1655669 0.92076844 0 0 0 8.2539062 2.6835938 A 1.1655669 0.92076844 0 0 0 9.2050781 2.2929688 A 1.741594 1.2970195 0 0 0 10.318359 2.59375 A 1.741594 1.2970195 0 0 0 12.058594 1.296875 A 1.741594 1.2970195 0 0 0 10.318359 0 z M 6.1289062 4.59375 L 5.7421875 14 L 8.2578125 14 L 7.859375 4.59375 L 6.1289062 4.59375 z "
+     style="opacity:1;fill:#000000;fill-opacity:1;fill-rule:evenodd;stroke:none;stroke-width:1.92604458;stroke-linecap:square;stroke-linejoin:miter;stroke-miterlimit:4;stroke-dasharray:none;stroke-opacity:1;paint-order:stroke fill markers" />
```
pnorman: don't set stroke-width, etc

symbols/towers/communications_tower.svg
```diff
@@ -0,0 +1,48 @@
...
+       id="tower-freestanding-communication" />
+    <circle
+       r="2"
+       cy="5"
+       cx="7"
+       id="path833"
+       style="fill:#000000;fill-opacity:1;fill-rule:nonzero;stroke:none;stroke-width:1.36062992;stroke-linecap:square;stroke-linejoin:miter;stroke-miterlimit:4;stroke-dasharray:none;stroke-dashoffset:0;stroke-opacity:1" />
```
pnorman: no stroke-width, etc

symbols/towers/tower_cooling.svg
```diff
@@ -0,0 +1,31 @@
...
+        <dc:title></dc:title>
+      </cc:Work>
+    </rdf:RDF>
+  </metadata>
+  <defs
+     id="defs113" />
+  <path
+     id="rect4506"
+     d="M 7.0390625 0 C 5.7192088 -0.00551362 4.3801947 0.16076646 3.0566406 0.515625 C 3.3287637 1.8898063 3.3974141 3.0704698 3.3398438 4.3828125 C 3.0409944 7.9972543 1.6577834 11.046251 0.97265625 14 L 12.982422 14 C 12.297538 11.047298 10.91835 7.9994736 10.619141 4.3867188 C 10.561251 3.0731598 10.627979 1.8911333 10.900391 0.515625 C 9.6590955 0.18282094 8.3589162 0.00551362 7.0390625 0 z M 7 0.56054688 A 2.8824873 0.8187421 0 0 1 9.8828125 1.3808594 A 2.8824873 0.8187421 0 0 1 7 2.1992188 A 2.8824873 0.8187421 0 0 1 4.1171875 1.3808594 A 2.8824873 0.8187421 0 0 1 7 0.56054688 z "
+     style="opacity:1;fill:#000000;fill-opacity:1;fill-rule:evenodd;stroke:none;stroke-width:2.27995706;stroke-linecap:square;stroke-linejoin:miter;stroke-miterlimit:4;stroke-dasharray:none;stroke-opacity:1;paint-order:stroke fill markers" />
```
pnorman: no stroke-width, etc

symbols/towers/tower_dish.svg
```diff
@@ -0,0 +1,31 @@
...
+     d="M 8.0685044,0.05217521 C 7.5102106,0.08185002 6.9559019,0.19918572 6.4298325,0.40262664 L 7.2794418,1.2495509 c 1.3855212,-0.36956358 2.9010802,-0.038252 3.9941402,1.0513543 1.09306,1.0896057 1.425423,2.6003758 1.054687,3.9815176 l 0.84961,0.8469243 C 13.903519,5.2647886 13.533641,3.0468811 12.029441,1.5474346 11.089316,0.61028076 9.8680989,0.11516028 8.6270981,0.05217521 c -0.1861501,-0.0094477 -0.3724959,-0.0098916 -0.5585937,0 z M 8.0470204,2.4021467 c -0.4702958,-0.00363 -0.9414804,0.098546 -1.375,0.3017777 L 7.4747548,3.5041217 C 8.1663982,3.3194652 8.9266412,3.4853662 9.4728016,4.0297989 10.01896,4.5742318 10.185386,5.3320738 10.000145,6.0215312 l 0.802734,0.8001974 C 11.346549,5.6693305 11.159591,4.2471506 10.207176,3.2977448 9.6119171,2.7043661 8.8308463,2.4081948 8.0470204,2.4021467 Z M 2.9981918,2.7545451 c -2.05459658,2.0481034 -2.05459658,5.3697857 0,7.4178879 2.0545969,2.048105 5.3848563,2.048105 7.4394532,0 L 7.2110825,6.9560684 8.5685044,5.6048834 7.8380356,4.8767231 6.4825668,6.2279081 Z M 2.3614731,10.933692 0.46303563,14.044922 H 7.7325668 V 12.615859 C 5.8397654,12.922925 3.836333,12.363904 2.3614731,10.933692 Z"
+     style="stroke-width:1.03711045" />
```
pnorman: no stroke

symbols/towers/tower_dome.svg
```diff
@@ -0,0 +1,31 @@
...
+        <dc:title></dc:title>
+      </cc:Work>
+    </rdf:RDF>
+  </metadata>
+  <defs
+     id="defs113" />
+  <path
+     id="rect4501"
+     d="M 7 0 A 3.1955194 3.1955194 0 0 0 3.8046875 3.1953125 A 3.1955194 3.1955194 0 0 0 5.1152344 5.7714844 L 5.1152344 14 L 8.8847656 14 L 8.8847656 5.7714844 A 3.1955194 3.1955194 0 0 0 10.195312 3.1953125 A 3.1955194 3.1955194 0 0 0 7 0 z "
+     style="opacity:1;fill:#000000;fill-opacity:1;fill-rule:evenodd;stroke:none;stroke-width:2.61180997;stroke-linecap:square;stroke-linejoin:miter;stroke-miterlimit:4;stroke-dasharray:none;stroke-opacity:1;paint-order:stroke fill markers" />
```
pnorman: no stroke-width, etc

kocio-pl:
> We're adding 5 more icons. Any symbol addition needs to be worth the cost of additional icons, and no reason is given for why we want to render these. I makes it hard to address if it should be merged without a reason for the changes.

I'm glad you have asked. I was thinking about it myself and came to the conclusion that I find lack of more "tower" shapes disturbing. This is so broad category of objects, that the only thing that make them similar are their height and (to some lesser degree) proportions. No wonder generic tower shape looks much like the mast - it can be anything, so the neutral option seems to be to show generic idea of "height". This is especially true for cooling towers and that's why their visual weight is higher - they are big and "heavy" in reality, totally unlike masts:

https://commons.wikimedia.org/wiki/File:Didcot_power_station_cooling_tower_zootalures.jpg

> Stroke is used in a couple places

I believe this is what Inkscape puts in the files automatically. I have never used a stroke in our icons, yet I have found it multiple times. Think about it as a stroke value that would be used **in case** you'd be using this feature (even if you're not).

kocio-pl:
> The icons don't meet our guidelines. In particular

@Tomasz-W Do you think this can be improved (mostly pixel aligning) without losing readability?

Tomasz-W:
> Do you think this can be improved (mostly pixel aligning) without losing readability?

No. I made them as best as I can, if someone think they could be improved, please do it on your own hand.

kocio-pl:
@nebulon42 What do you think about these icons? Could you try to fix them?

kocio-pl:
> Most of these tags are used less than 1000 times, I don't think adding icons for them is worth the additional code/icons.

type | uses
---- | -----
[tower:construction=dish](https://taginfo.openstreetmap.org/tags/tower%3Aconstruction=dish) | 896
[tower:construction=dome](https://taginfo.openstreetmap.org/tags/tower%3Aconstruction=dome) | 239
[tower:type=cooling](https://taginfo.openstreetmap.org/tags/tower%3Atype=cooling) | 1 118
[tower:type=watchtower](https://taginfo.openstreetmap.org/tags/tower%3Atype=watchtower) | 426
[man_made=chimney](https://taginfo.openstreetmap.org/tags/man_made=chimney) | 39 656
[man_made=communications_tower](https://taginfo.openstreetmap.org/tags/man_made=communications_tower) | 3 946

Only dish, dome and watchtower are <1k. Watchtower reuses observation icon and would be rendered anyway, but too early and weird.

All of them are documented, easy to observe by a layman and to depict. I see no reason to keep current rendering, which is too generic as I mentioned lately.

meased:
I have made an attempt at some pixel alignment for these icons. 

https://gist.github.com/meased/cf8b1309c32ad59c29f03e2396c41859

- Chimney
    - chimney svg
    - The cloud was too complex, it has been replaced with the smoke from our shop=tobacco icon.
    - It now has a base.

- Communications Tower
    - communications_tower svg
    - Base is changed.
    - The top spire is fading (can't draw a point two pixels wide), I think this would look better without it, but then it would just be the dome tower with signal waves.

- Cooling Tower
    - tower_cooling svg
    - Changed to shelf perspective (removed view into the opening on top, it looked like the Eye of Sauron at 14x14.
    - It is a little narrower and shorter to reduce the weight.

- Dish Tower
    - tower_dish svg
    - Just slight alignment on the base and signal waves.

- Dome Tower
    - tower_dome svg
    - It now has a base.

I made everything use triangular bases where applicable so there is some element that ties them together.

wmyrda:
Hence @meased prepared pixel alignment requested by @pnorman and no one objected to the answers for @matthijsmelissen concerns about quantity of items than is there anything stopping this code from being merged?

Tomasz-W:
Most of the icons looks OK after @meased correction, but I think the chimney icon is much worse now. The smoke shape is too big, and chimney shape reminds me a monument icon.

kocio-pl:
I eventually plan to update the code and prepare new set of rendering tests, I just take some more time for personal activities lateky and proposed release is far enough (see #3248).

I also don't like chimney too much, so I would be more happy with rework closer to the original design, however it has a smoke, which makes it not too similar to anything else, including monument. @Tomasz-W could you try to try to tune your shape?

meased:
Just to make sure everyone is viewing this icon at the target size:

Original

chimeny_old

Updated

chimney_new svg

wmyrda:
Looking at the two pictures I must say I also like original chimney proposition more. The cloud looks better and excuse my lack of painting skills but maybe it is enough to [pixel alight the chimney itself](https://image.ibb.co/kD2gsd/chimney.jpg)?

Tomasz-W:
It's not fully pixel-alligned, but I improved oryginal SVG a little bit:

clean svg man_made chimney

https://gist.github.com/Tomasz-W/5b82f20209e3093dcd77a74d414c4989

meased:
The updated icon made the stack shorter in order to get more detail in the smoke. Maybe this went too far and got too cartoonish looking. Here are some mockups with smaller smoke, taller stack, and various width/base combinations (all pixel-aligned):

chimney_samples

Tomasz-W:
First one is good.

meased:
[My gist from a few posts ago](https://gist.github.com/meased/cf8b1309c32ad59c29f03e2396c41859) has been updated to include chimney 1.

kocio-pl:
Updated rendering of icons - these are more or less OK for me, so I will update them:

rg9hp42m
bhilgpd
spz xgms
m ni2 j7

I just needed to remove fading from communications_tower, because Mapnik doesn't get it right:
quyxy_pi

I think new dishes are much worse - technical guidelines should not be enforced by all means, so I don't plan to update them:
2qsiuh5y

sinanna:
@kocio-pl Please don't forget to render man_made=cooling_tower!

It's used [https://taginfo.openstreetmap.org/tags/man_made=cooling_tower 954 times] vs [https://taginfo.openstreetmap.org/tags/tower:type=cooling 1248 times] of the "tower:type" variant.
And it's getting rendered, e.g. from [http://demo.f4map.com/#lat=47.7348552&lon=2.5183640&zoom=16&pins.minBuildingHeight=1&camera.theta=53.021&camera.phi=1.146 F4map]!

Don't say it's deprecated - that's just not true.

dieterdreist:
2018-06-22 10:18 GMT+02:00 sinanna <notifications@github.com>:

> @kocio-pl <https://github.com/kocio-pl> Please don't forget to render
> man_made=cooling_tower!
>
> It's used [https://taginfo.openstreetmap.org/tags/man_made=cooling_tower
> 954 times] vs [https://taginfo.openstreetmap.org/tags/tower:type=cooling
> 1248 times] of the "tower:type" variant.
> And it's getting rendered, e.g. from [http://demo.f4map.com/#lat=
> 47.7348552&lon=2.5183640&zoom=16&pins.minBuildingHeight=1&
> camera.theta=53.021&camera.phi=1.146 F4map]!
>
> Don't say it's deprecated - that's just not true.
>


there is something suspicious with this tag. It had a normal, slightly
growing curve until the beginning of this year (around 540 uses) and then
grew in one step to 900 (import or automated edit, both not
approved/announced AFAIR). Also the steep curve since then to 1200 looks
suspicious compared to the rest. This said, I would support rendering the
tag, it is sematically better than using a tower_type and it would be more
consistent with the system (e.g. man_made=water_tower, power=tower which
both aren't tower:types and also cooling towers don't seem nicely tagged
with man_made=tower in general).
http://taghistory.raifer.tech/

kocio-pl:
The problem is more general: when we have systematic tag (like man_made=tower or historic=palace) and some subtypes, it's good to have them all. If there's a competing tag, what should we do with related subtype? 

This problem is invisible when tagging and defining tags, you can simply ignore competing tags, but when choosing one for rendering (because we want to avoid tags fragmentation) it gets crazy.

This one looks really suspicious and breaks the system, so I prefer not to render it now, because the rest is ready and overdue for a few weeks. But since I prefer taking one step at the time, we can discuss the problem and change it in the future in any way. See also https://github.com/gravitystorm/openstreetmap-carto/pull/3099#issuecomment-377670336.

kocio-pl:
BTW: the whole cooling issue is dynamic and not settled yet - we're in the middle of abrupt changes, so expect things to change in the future anyway:

taghistory 22

kocio-pl:
I deny predicting the future... Let's re-examine this after some time, like few 6-12 months.

Zverik:
Okay, found one of the retaggers: https://www.openstreetmap.org/changeset/59147983

imagico:
See also

https://taginfo.openstreetmap.org/tags/tower:type=cooling#map
https://taginfo.openstreetmap.org/tags/man_made=cooling_tower#map

Classic case of two competing taggings and different regional preferences and no clear dominance.  Also great example to illustrate why absolute use numbers say very little about acceptance and popularity when the tagged object is relatively rare and there might be a lot of mappers who support a certain tagging but have very little opportunity to actually use it and this way manifest their opinion in actual use numbers.

imagico:
Note the neutral approach would be to render neither but this would only be neutral if man_made=tower + tower:type=cooling is not rendered with a generic tower symbol.

wmyrda:
Hooray for getting this merged!
Just of of curiosity why some svg files have the `xmlns:dc=` and alike data as one liner while others have them spread out in different lines. IMHO for all the svg files in repository they should be kept consistent whichever whey is preferred.

wmyrda:
> cooling towers do not fit at all under the definition of man_made=tower 
Cooling tower just like the name says is a **tower** just of specific use. Same goes for all other towers and if you ask me all other man_made=* structures resembling any type of tower should be changed to `man_made=tower` and `tower:type=*` for consistency. This is natural approach as [man_made=tower engineering definition](https://wiki.openstreetmap.org/wiki/Tag:man_made=tower#Engineering_definition) explains. It is just OSM always has to mix things up and say something stupid as  "A tower is accessible and provides platforms" which has nothing to do with basic tower definition.
It is obvious from historical data shown in this thread that users agree with that natural approach and like `tower:type:*` and just some manual intervention had to change things up. That being said I am glad it is `["tower:type" = 'cooling']` that is rendered.

imagico:
Once again: please stop the tagging discussion, this has no place here.  If mappers decide to consistently tag cooling towers `foo=bar` it would not be the place of this style to tell them they are wrong.

And @dieterdreist - no, tag fragmentation does not start only when there are more than two synonyms for the same type of feature.

kocio-pl:
Some tag discussion is OK for me here, because we try to interpret the state of things to render them properly. But given your idea of anarchic tagging, I see no place to tell anybody to be wrong (just being different from majority in best case). 

> tag fragmentation does not start only when there are more than two synonyms for the same type of feature.

Could you explain what do you mean by that?

dieterdreist:
sent from a phone

> On 23. Jun 2018, at 12:20, kocio-pl <notifications@github.com> wrote:
> 
> tag fragmentation does not start only when there are more than two synonyms for the same type of feature.
> 
> Could you explain what do you mean by that?
> 


if different concepts are used to describe things, where there is some overlap of meaning of these tags, it would also be tag fragmentation (especially if there were many different concepts and overlaps, and all would be similar in number of uses).

imagico:
> Could you explain what do you mean by that?

I meant to turn down the attempt from @dieterdreist to relativize long standing principles of this style with a 'but it could be worse' argument.

We have had the common practice in the past that if there are several tagging schemes competing with similar support to be the method to tag certain real world features that we avoid taking sides by rendering one of them but not the other, especially if there are different regional preferences like here and such a decision can be seen as regional favoritism.  At the same time we have traditionally avoided adding rendering of multiple tags with no discernible difference in meaning to avoid supporting tagging fragmentation.  The situation here is somewhat special because both before and after merging this PR we were not neutral because we previously rendered man_made=tower (and therefore also man_made=tower + tower:type=cooling) with a generic tower symbol.  So you can rightfully argue if this change makes this worse or not.  And generic man_made=tower has been rendered for a really long time.  But i consider it important to uphold the principles i explained even if in this case there is no way to fully put the genie back into the bottle so to speak.  So i pointed out that we have a case here where we should normally refrain from rendering either of the competing tags.

By the way it is well possible that some of the recent re-tagging activity is due to cooling towers typically being quite high and height being used for importance rating of towers now.  It seems many of the more recent re-taggings are on cooling towers with a height tag.  In other words:  Before this PR there could have been an incentive for re-tagging to man_made=cooling_tower to avoid the prominent unsuitable symbol, after this PR the incentive might turn into the opposite direction.  It is important to be mindful about how we are messing with the mappers here - deliberately or not.

dieterdreist:
sent from a phone

> On 23. Jun 2018, at 14:10, Christoph Hormann <notifications@github.com> wrote:
> 
> And generic man_made=tower has been rendered for a really long time.


actually not, in my perception it is rendered only relatively recently

imagico:
Right, sorry, my mistake - i had mixed mast and tower in my recollection.  That means #2671 was the change initially messing with things.

imagico:
Not much - it just means that the problem could have been avoided in #2671 and it puts additional support on the observation described in the last paragraph.

imagico:
As i said this would be the neutral approach to the issue.

But as also said this would not nullify the previous influence we already exercised (as mentioned with this PR included in two different directions).

For me this is all much more about raising awareness and making conscious decisions than about trying to correct decisions afterwards (which does not prevent making the same mistakes again in the future anyway).

dieterdreist:
sent from a phone

> On 23. Jun 2018, at 16:46, Christoph Hormann <notifications@github.com> wrote:
> 
> For me this is all much more about raising awareness and making conscious decisions than about trying to correct decisions afterwards (which does not prevent making the same mistakes again in the future anyway).


for me man_made=cooling_tower is the better tag, more consistent with the rest of the tags (water_tower, communication_tower, power=tower, not conflicting with the man_made=tower definition), and it is clear, if we do not render it but render tower:type=cooling it will not make it

As we are aware of this issue now, why would we not fix it (fixing in this context means acting in accordance with the self proclaimed goals and criteria/procedures).
We (OSM system of tagging and rendering) have fixed several very elemental object types just relatively late, after more than a decade of mapmaking, e.g. squares, unpaved roads, bridges, and now towers. It is possible to adjust decisions.

kocio-pl:
>  That means #2671 was the change initially messing with things.

From my point of view, the story of roots of this small problem are much deeper. This PR was solution based on issue #2556 (with code implemented first by @wmyrda), which has been formal proposition to solve the problem described on 29.11.2016 in @BushmanK blog entry. And the mess was born with rendering telecommunication icon for variety of towers and masts (I don't even know when).

It's a normal thing for me to see the things evolving and finding new problems while trying to fix the old ones. Also OSM is not a static project, and osm-carto should follow.

> would it be open for consideration to remove the rendering of man_made=tower when it is combined with tower:type=cooling ?

Do you mean rendering "stick" or not rendering any tower icon at all?

> We (OSM system of tagging and rendering) have fixed several very elemental object types just relatively late, after more than a decade of mapmaking, e.g. squares, unpaved roads, bridges, and now towers. It is possible to adjust decisions.

For me it's more than possible - it's needed.

sinanna:
@kocio-pl I would like to report another mechanical retagging on cooling towers (by user [ForstEK])(https://www.openstreetmap.org/user/ForstEK/history): http://www.openstreetmap.org/changeset/60118947

He even refuses to accept a revert and started a "Wikipedia-style" edit war: http://www.openstreetmap.org/changeset/60189430

geozeisig:
Now tower:construction=dish will be rendered. But only if man_made=tower is used. Frequently, however, man_made=telescope ([here](https://www.openstreetmap.org/node/3868969565)) is also used or man_made=antenna ([here](https://www.openstreetmap.org/node/3156840038)).
Is this a tagging error? Or can we also render telescope with it?

Issue #2504, opened 2016-12-11
Rendering of building=train_station and railway=platform/highway=platform(?) does not respect layers


jc86035:
See [22.4482°N, 114.0046°E](https://www.openstreetmap.org/#map=18/22.44828/114.00461). The small light rail platforms (1, 2 and 3 to the northeast) should ideally be rendered below the large train station building (or in a tunnel casing, if possible), and the train platforms (and station building) should all be rendered above the road junction. However, currently all are rendered below the road, and platforms do not change rendering with `tunnel=yes`/`covered=yes`.

In addition, the new brown rendering of `building=train_station` is [being a little annoying](https://www.openstreetmap.org/#map=18/22.27921/114.18325). Although the "station building" is underground, the brown is rendered above the buildings located over the station. I know the solution in this particular case is probably to change the station to pedestrian passageways + platforms with a `railway=station` node, but the brown casing could instead be made transparent.

kocio-pl:
It's quite old and known problem - see especially https://github.com/gravitystorm/openstreetmap-carto/issues/688 and https://github.com/gravitystorm/openstreetmap-carto/issues/327.

The general problem with layers is that we don't have "universal" layers. They work only inside the single class of objects and it's a designer decision to make some classes be located on top of some others.

However I'm curious now if we could make a special, separate layer just for train stations and make it located above all the roads and platforms. This exception could be compromise worth having, because train stations in big cities should be visible as landmarks and we're not able to show their inner structure in a usable way without layer viewer, so they tend to suffer the most:

Berlin - http://www.openstreetmap.org/node/3856100103
New York - https://www.openstreetmap.org/way/265947358
Paris - http://www.openstreetmap.org/way/67825345
London - http://www.openstreetmap.org/relation/6036811
Warsaw - http://www.openstreetmap.org/way/367491792

geozeisig:
> They work only inside the single class of objects and...
It does not even work with the platforms. In the case of the Berlin example, the upper platform merges with the lower platform. It should be recognizable by a margin who lies over whom.

mboeringa:
> In the case of the Berlin example, the upper platform merges with the lower platform. It should be recognizable by a margin who lies over whom.

Just for your information: even if a style would support rendering platforms in a layered fashion (which isn't trivial to implement), the Berlin example would fail, as the platforms are only tagged with level=x, not layer=x:


I

jc86035:
[...]

Issue #2037, opened 2016-02-06
Peculiar draworder of railway=platform


Klaus-Tockloth:
This looks peculiar for me:

![bildschirmfoto 2016-02-06 um 07 58 25](https://cloud.githubusercontent.com/assets/5493128/12865286/3aa0560e-cca8-11e5-8a66-5e2d94b2ca67.jpg)

The roof of the station is tagged with "building=roof + layer=4" and the platforms with "railway=platform + layer=0". My expectation is now, that the roof is rendered over the platforms. Something like this:

![bildschirmfoto 2016-02-06 um 07 57 04](https://cloud.githubusercontent.com/assets/5493128/12865292/583ac37a-cca8-11e5-932c-7d95681eca44.jpg)

Link: http://www.openstreetmap.org/#map=18/52.54892/13.39016

Regards Klaus

kocio-pl:
I think this is similar to https://github.com/gravitystorm/openstreetmap-carto/issues/688.

matkoniecz:
> My expectation is now, that the roof is rendered over the platforms.

What about platforms fully covered by roofs/other buildings? Is it desirable to not render them?

See for example http://www.openstreetmap.org/#map=18/50.06829/19.94767

kocio-pl:
> What about platforms fully covered by roofs/other buildings? Is it desirable to not render them?

I would say now - yes. See also https://github.com/gravitystorm/openstreetmap-carto/issues/1659#issuecomment-376017881.

The stations in the cities are the most horrible renderings I know because the buildings are eclipsed by rails, platforms and pedestrian zones, it's hard to spot them even if they are dark brown. This is general style, not an indoor map - there's even the one that is specific for stations:

https://wiki.openstreetmap.org/wiki/OpenStationMap

Issue #1659, opened 2015-07-14
Highway = platform is too strong on zoom 16 


HolgerJeromin:
Highway = platform is too strong on zoom 16. Higher zoom could be adjusted, too. For me a platform is very similar to a footway. 
http://www.openstreetmap.org/#map=16/50.7740/6.0880

matkoniecz:
Thanks for linking location. It is a good idea to include also image as data/rendering may change.

kocio-pl:
It's probably too early to show them on z16 - they are very local feature (it's enough to see the station until high zooms) and make things worse in some cases:

https://www.openstreetmap.org/way/468605078#map=16/40.7495/-73.9914

kocio-pl:
I guess you're right, simple examples for testing: 

https://www.openstreetmap.org/node/256348461#map=16/52.2500/21.0563
https://www.openstreetmap.org/way/24635942#map=16/52.1792/21.0243

jeisenbe:
> z18 would be a good start for nonrailway platforms and z16 only for railway platforms.

I've marked this as a good first issue, since changing the initial zoom level for the cases above should be reasonably easy.

PR #3162, opened 2018-04-02
Hide railway=platform with location=underground, tunnels and covered=yes

kocio-pl:
Related to #2504, #2037 and #1659.

- Hide railway platforms which are located underground, because they eclipse station buildings too much and this is rather indoor feature (see https://www.openstationmap.org )

[Warsaw](https://www.openstreetmap.org/#map=17/52.22925/21.00564)
Before
zw4rvxvq
After
qw3ocor3

[Berlin](https://www.openstreetmap.org/#map=17/52.52497/13.37020)
Before
xecy6jgv
After
qjfugaih

### DIFF
--- a/project.mml
+++ b/project.mml
@@ -543,7 +543,7 @@ Layer:
                 z_order
               FROM planet_osm_line
               WHERE (tunnel = 'yes' OR tunnel = 'building_passage' OR covered = 'yes')
-                AND railway IS NOT NULL -- end of rail select
+                AND (railway NOT IN ('platform') AND railway IS NOT NULL) -- end of rail select
             ) AS features
           ORDER BY
             layernotnull,
@@ -692,11 +692,18 @@ Layer:
             way,
             COALESCE((
               'highway_' || (CASE WHEN highway IN ('residential', 'unclassified', 'pedestrian', 'service', 'footway', 'cycleway', 'track', 'path', 'platform') THEN highway ELSE NULL END)),
-              ('railway_' || (CASE WHEN railway IN ('platform') THEN railway ELSE NULL END))
+              ('railway_' || (CASE WHEN (railway IN ('platform') 
+                              AND (tags->'location' NOT IN ('underground') OR tags->'location' IS NULL)
+                              AND (tunnel NOT IN ('yes', 'building_passage') OR tunnel IS NULL)
+                              AND (covered NOT IN ('yes') OR covered IS NULL))
+                              THEN railway ELSE NULL END))
             ) AS feature
           FROM planet_osm_polygon
           WHERE highway IN ('residential', 'unclassified', 'pedestrian', 'service', 'footway', 'track', 'path', 'platform')
-            OR railway IN ('platform')
+            OR (railway IN ('platform') 
+                AND (tags->'location' NOT IN ('underground') OR tags->'location' IS NULL)
+                AND (tunnel NOT IN ('yes', 'building_passage') OR tunnel IS NULL)
+                AND (covered NOT IN ('yes') OR covered IS NULL))
           ORDER BY COALESCE(layer,0), way_area DESC
         ) AS highway_area_casing
     properties:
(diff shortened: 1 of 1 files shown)


daganzdaanda:
This PR seems to hide more than just the "location=underground" platforms. See https://www.openstreetmap.org/relation/918697 for a level 2 platform that is not visible anymore. 

But generally I like the idea. It's a hard decision to make - more mapper feedback vs. a more usable map. IMHO we can't continue to show everything anymore. We should hide nearly everything that is not on the ground level (using the "level" tag). Maybe there will be some way in the future to turn on some indoor / level-by-level / 3D features on the main map.

kocio-pl:
Thanks for spotting it! Do you have an idea why is that? Hm, maybe we just need to include platforms without `location` tag (NULL checking). The fix should be quite simple then.

I don't know if there are underground platforms, but if yes, that should help also:
https://www.openstreetmap.org/way/138141251#map=18/40.75063/-73.99338

I hope to hide more underground features, but have no idea how to code it (see #1977).

kocio-pl:
Looks like it really was the missing null checking, updated renderings:

[Warsaw Central Station](https://www.openstreetmap.org/#map=19/52.22874/21.00319)
Before
wiqx9l4e

After
6ucp_1f2

[Warsaw Śródmieście Station](https://www.openstreetmap.org/#map=19/52.22942/21.00768)
Before
wsrgzt g

After
6mrnnbqi

[Madison Square Garden (USA, NY)](https://www.openstreetmap.org/#map=18/40.75043/-73.99351)
Before
tqtbh_dv

After
orgixep3

[Berlin Hauptbahnhof](https://www.openstreetmap.org/#map=17/52.52536/13.36821)
Before
9xrtq2mo

After
v7w1in8

kocio-pl:
I'm still not sure about levels, but I was told that `tunnel=yes` is also worth hiding and I have found that there are [~1700 such railway platforms](https://taginfo.openstreetmap.org/tags/railway=platform#combinations).

kocio-pl:
It looks like platforms in tunnels should be hidden too (the code has to avoid catching them also by other layers) to avoid mess like these:

London, [platform under ground](https://www.openstreetmap.org/relation/7930948#map=19/51.50426/-0.10602) and on the ground:

Before
knk1c3rs

After
nfbyaauu

London, [many underground platforms](https://www.openstreetmap.org/#map=19/51.51747/-0.11986):

Before
tfkwunht

After
y8qnmmbf

kocio-pl:
The same for `covered=yes` - basically anything that is not in clear view creates a mess:

London, [covered station platforms](https://www.openstreetmap.org/#map=19/51.49209/-0.22334)

Before
dzwut9gf

After
orvsack

mueschel:
This hides all platforms with the tag 'covered = yes' completely. In several cases I checked, this results in a platform completely vanishing without any object drawn instead. Examples:

https://www.openstreetmap.org/way/49212472

https://www.openstreetmap.org/way/196930811

This clearly complete hiding of platforms is definitely worse than having a platform covering a small piece of roof.

kocio-pl:
I don't think it's definitely worse. If it's covered fully, one should add the covering object. If it's partially, one need to set `covered=partially`.

I still see no better way to avoid clutter while at the same time promoting more precise tagging for `covered=*`.

mueschel:
I think that platforms (even underground ones) are a very important feature. In fact, much more important than the roof covering it. If I'm not familiar with a station, I have to identify the platforms, but I don't need to know where roofs are. 

In many cities it's important to know where subway platforms are - just having the station node and (if lucky) some subway_entrance is not sufficient to find the best ways (and this style is "the way" to look at OSM data for most users). I fear that this decision leads to platform areas marked as highway=pedestrian just to make them visible on the map again.

I know it's a short coming of the renderer to not be able to arbitrarily set the order of rendering or transparency in every individual case, but platforms are just too important to remove them.

kocio-pl:
> In many cities it's important to know where subway platforms are - just having the station node and (if lucky) some subway_entrance is not sufficient to find the best ways (and this style is "the way" to look at OSM data for most users)

I think this is exactly what is needed - to know where the station is and where are the entrances (this is the closest thing for subway routing - once you pass them, you have the easy way to the platform). That's why I've also made stations to be dark brown to be easier to spot as a buildings. 

The platforms are part of internal station plan, so if they are visible, then we can show them of course, but if not, they are secondary feature which creates a lot of clutter and shape distortion (shops can make a clutter too, but they don't eclipse station shape and they are much more general feature, and we can't detect when they are part of the station).

> I fear that this decision leads to platform areas marked as highway=pedestrian just to make them visible on the map again.

This is always valid consideration, but I feel that the universal style is not a QA tool, rather general overview.

I've made this change as a last minute addition, but the more I think about it, the more I appreciate it.

dieterdreist:
2018-06-12 10:39 GMT+02:00 mueschel <notifications@github.com>:

> I know it's a short coming of the renderer to not be able to arbitrarily
> set the order of rendering or transparency in every individual case, but
> platforms are just too important to remove them.
>


it is not a general shortcoming of the renderer, but a deliberate decision
of this style not to use transparency (because it leads to hard to predict
colors) and not to respect explicit layer tags in all cases.

lakedistrictOSM:
@kocio-pl Was this supposed to hide both areas and lines or just areas? I ask because https://www.openstreetmap.org/way/48837616 isn't hidden, yet has a `location=underground` tag.

kocio-pl:
I would be happy to hide also such lines.

famo:
> I would be happy to hide also such lines.

Is there already an specific issue for this, where we could track this?

kocio-pl:
Yes, look at #3336.

dieterdreist:
I believe the underground platforms are very important and it’s a pity to hide them completely, we should rather discuss how to show them in a way that layering becomes clear and you can understand that it is an underground feature.
Covered platforms should be shown

mboeringa:
> I believe the underground platforms are very important and it’s a pity to hide them completely, we should rather discuss how to show them in a way that layering becomes clear and you can understand that it is an underground feature.

Honestly, if you _truly_ believe these features are important, you should be willing to use a dedicated indoor renderer that is capable of showing indoor leveled stuff in a meaningful way:

https://www.openstationmap.org
https://openlevelup.net/

If you are not willing to use those, than I believe that this kind of stuff isn't really that important to you in reality, and it is more of the typical "Oh, I spend some hours digitizing indoor stuff, and now it doesn't show up in the map".

There is only so much you can do in a 2D map, at some point it becomes nothing but a theoretical question.

dieterdreist:
referring to indoor maps does not nail it, public transport platforms are not comparable to your typical indoor space, because the latter is typically only of interest to a much smaller group of people (those that want to enter the building) than a public transport infrastructure.

PR #3171, opened 2018-04-08
Render place=square name for nodes

d3d9:
See #2673

Rendering place=square name for nodes too

z17
square-z17
z18
square-z18
z19
square-z19

Data (4 objects with place=square, original place is [here](https://www.openstreetmap.org/node/4917621976)):
square-data

### DIFF
--- a/placenames.mss
+++ b/placenames.mss
@@ -395,5 +395,14 @@
       text-halo-fill: white;
     }
   }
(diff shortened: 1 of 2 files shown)


kocio-pl:
Thanks for spotting the problem and fixing it so fast!

PR #3172, opened 2018-04-10
Add rendering of zip line

### LINKED ISSUE #3156: Add rendering of aerialway=zip_line
### Expected behavior
Basically, `aerialway=zip_line` should be rendered like most of aerialways (e.g. `aerialway=chair_lift`) Using a distinct pattern would be excluded due to the low usage (only 426 on Taginfo with regard to the 8053 counts for `aerialway=chair_lift`).
### Actual behavior
`aerialway=zip_line` is not rendered.
z=16 (https://www.openstreetmap.org/way/132119392)
aerialway_zip_line_z16
Please note the two stations "Fantasticable" to the north that each correspond to a `aerialway=zip_line`

jragusa:
Fixes #3156 

Display `aerialway=zip_line` as other aerialways (e.g. chair_lift)

Fantasticable [1](https://www.openstreetmap.org/way/132119392) and [2](https://www.openstreetmap.org/way/132119393)
Before
zipline_before

After
zipline_after

### DIFF
--- a/aerialways.mss
+++ b/aerialways.mss
@@ -71,7 +71,8 @@
   [aerialway = 't-bar'],
   [aerialway = 'j-bar'],
   [aerialway = 'platter'],
-  [aerialway = 'rope_tow'] {
+  [aerialway = 'rope_tow'],
+  [aerialway = 'zip_line'] {
     [zoom >= 12] {
       line/line-width: 1;
       line/line-join: round;


kocio-pl:
What about name rendering?

jragusa:
Sorry, I missed that
On a shorter way: https://www.openstreetmap.org/way/389429866
zipline_name

kocio-pl:
I was thinking about making rendering a bit different (for example bigger space between markers), but it looks like this kind of rendering is already made different for all lighter aerialways, so it's probably a proper solution.

kocio-pl:
Independent testing of a different place gave the same result - thanks for fixing this issue!

PR #3174, opened 2018-04-11
Drop derelict_canal rendering

### LINKED ISSUE #1003: dry canals are displayed as blue dashes
Pernamently dry derelict canal ( http://wiki.openstreetmap.org/wiki/Tag:waterway%3Dderelict_canal ) is currently displayed as blue, transparent dashes. I thing that it would be better to use some symbol that is not blue - but I have no good suggestions.

example of a current rendering: http://www.openstreetmap.org/?mlat=49.20021&mlon=16.22416#map=17/49.20021/16.22416

selection_003

sommerluk:
Resolves #1003

### DIFF
--- a/project.mml
+++ b/project.mml
@@ -295,7 +295,7 @@ Layer:
             CASE WHEN tunnel IN ('yes', 'culvert') THEN 'yes' ELSE 'no' END AS int_tunnel,
             'no' AS bridge
           FROM planet_osm_line
-          WHERE waterway IN ('river', 'canal', 'derelict_canal', 'stream', 'drain', 'ditch', 'wadi')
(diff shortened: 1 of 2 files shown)


kocio-pl:
Do you think it's better to just drop it now than count on wiki changes that could clarify it? There were some changes recently:

https://wiki.openstreetmap.org/wiki/Tag:waterway%3Dderelict_canal

sommerluk:
I think indeed that dropping now it is the best solution.

kocio-pl:
I think that you're right - it can look in many ways: filled with water, intermittent water, ditch, ditch filled with plants etc., so there's no reasonable way of rendering it. Even proposed tags with prefixes still are unclear in this regard, so we need some more precise tagging to visualize it.

kocio-pl:
Thanks, testing looks OK.

PR #3178, opened 2018-04-18
Extend intermittent waterbody rendering to basins

kreed:
Follow up to #3104 to apply the same rendering to basins.


kocio-pl:
Could you show some examples with links?

kocio-pl:
Thanks, it works OK on other places too.

PR #3189, opened 2018-04-23
Refurbished spring icon (fix #325 using @imagico first tests)

### LINKED ISSUE #325: Change rendering for natural=spring

Penegal:
Fixes #325

- before
image
image
- after
image
image

Thanks to @imagico for the starting point.

### DIFF
--- a/project.mml
+++ b/project.mml
@@ -278,6 +278,24 @@ Layer:
       type: shape
     properties:
       minzoom: 8
+  - id: springs
+    geometry: point
+    <<: *extents
+    Datasource:
+      <<: *osm2pgsql
+      table: |-
+        (SELECT
+             St_Centroid(way) AS way, "natural"
+           FROM planet_osm_polygon
+           WHERE "natural" IN ('spring')
+         UNION ALL
+         SELECT
+             way, "natural"
+           FROM planet_osm_point
+           WHERE "natural" IN ('spring')
+        ) AS springs
+    properties:
+      minzoom: 14
   - id: water-lines
     class: water-lines
     geometry: linestring
--- a/project.mml
+++ b/project.mml
@@ -285,7 +285,7 @@ Layer:
       <<: *osm2pgsql
       table: |-
         (SELECT
-             St_Centroid(way) AS way, "natural"
+             way, "natural"
            FROM planet_osm_polygon
            WHERE "natural" IN ('spring')
          UNION ALL
--- a/symbols/spring.svg
+++ b/symbols/spring.svg
@@ -9,18 +9,15 @@
        transform="matrix(1.1666667,0,0,1.1666667,0,1033.6955)"
        d="m 6,13 a 3,3 0 1 1 -6,0 3,3 0 1 1 6,0 z"
        id="path4333"
-       style="fill:none;fill-opacity:1;stroke:#ffffff;stroke-width:0.60000002;stroke-miterlimit:4;stroke-dasharray:none;stroke-opacity:1"
-       inkscape:connector-curvature="0" />
+       style="fill:none;fill-opacity:1;stroke:#ffffff;stroke-width:0.60000002;stroke-miterlimit:4;stroke-dasharray:none;stroke-opacity:1" />
     <path
        d="m 6.875,1048.8622 a 3.375,3.3750069 0 1 1 -6.75,0 3.375,3.3750069 0 1 1 6.75,0 z"
        id="path4327-3"
-       style="display:inline;fill:none;stroke:#ffffff;stroke-width:3.5;stroke-miterlimit:4;stroke-dasharray:none;stroke-opacity:1"
-       inkscape:connector-curvature="0" />
+       style="display:inline;fill:none;stroke:#ffffff;stroke-width:3.5;stroke-miterlimit:4;stroke-dasharray:none;stroke-opacity:1" />
     <path
        transform="matrix(1.125,0,0,1.1250023,0.125,1034.2372)"
        d="m 6,13 a 3,3 0 1 1 -6,0 3,3 0 1 1 6,0 z"
        id="path4327"
-       style="display:inline;fill:none;stroke:#7abcec;stroke-width:1.5;stroke-miterlimit:4;stroke-dasharray:none;stroke-opacity:1"
-       inkscape:connector-curvature="0" />
+       style="display:inline;fill:none;stroke:#7abcec;stroke-width:1.5;stroke-miterlimit:4;stroke-dasharray:none;stroke-opacity:1" />
   </g>
 </svg>
--- a/symbols/spring.svg
+++ b/symbols/spring.svg
@@ -1,9 +1,26 @@
 <?xml version="1.0" encoding="UTF-8" standalone="no"?>
-
-<svg xmlns="http://www.w3.org/2000/svg"
-   version="1.1"
-   width="5"
-   height="7">
-	<rect width="5" height="7" fill="white" />
-	<path d="M 4,1 1,1 1,4 3,4 3,5 1,5 1,6 4,6 4,3 2,3 2,2 4,2 4,1 z" fill="#588499" />
+<svg
+   xmlns="http://www.w3.org/2000/svg"
+   version="1.1">
+  <g
+     id="layer1"
+     transform="translate(0,-1036.3622)">
+    <path
+       transform="matrix(1.1666667,0,0,1.1666667,0,1033.6955)"
+       d="m 6,13 a 3,3 0 1 1 -6,0 3,3 0 1 1 6,0 z"
+       id="path4333"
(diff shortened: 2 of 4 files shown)


kocio-pl:
Is the ring shown on top of the stream? I think it should be placed under.

Penegal:
And here you are:
image
image

kocio-pl:
[Example place](https://www.openstreetmap.org/#map=15/48.0355/7.0153):

z14
Before
n woblhd

After
ay3f2wzt

z15
Before
pw 4x7s2

After
acjtszjx

z16
Before
qo8jt0ry

After
hodtr89p

kocio-pl:
[Spring in a city park](https://www.openstreetmap.org/node/1417599098):

z14

z621h8hd

z15

eer63n5m

z16

fujormyj

meased:
The color of blue seems odd. I think it would look mire consustent with @water or @water-text.

Penegal:
@meased : I tried that first, but the icon with that shade of blue seemed more difficult to see with some backgrounds, as you may see in #325. Besides, @kocio-pl was in favor of this version despite the different shade, and I felt that the favor of an experienced collaborator was enough to PR, particularly if the problem lay for years without resolution.

kocio-pl:
I don't see anything like that at the moment. I just need some time to make additional test just in case, but basically that's all.

symbols/spring.svg
```diff
@@ -1,9 +1,26 @@
...
+   version="1.1">
+  <g
+     id="layer1"
+     transform="translate(0,-1036.3622)">
+    <path
+       transform="matrix(1.1666667,0,0,1.1666667,0,1033.6955)"
+       d="m 6,13 a 3,3 0 1 1 -6,0 3,3 0 1 1 6,0 z"
+       id="path4333"
+       style="fill:none;fill-opacity:1;stroke:#ffffff;stroke-width:0.60000002;stroke-miterlimit:4;stroke-dasharray:none;stroke-opacity:1"
+       inkscape:connector-curvature="0" />
```
kocio-pl: These are some Inkscape leftovers in this file, as seen above - please make sure to save the file as plain SVG.

Penegal:
@kocio-pl: this way? I'm a newbie regarding SVG code, so there might still be some useless code.

kocio-pl:
I haven't tested it, but probably yes. I guess that you've removed it by hand? Standard procedure is to just write the file in the Inkscape as a plain SVG instead of Inkscape SVG. Usually it leads to removing a lot of unnecessary stuff (probably helpful only for extensive editing). Could you try it?

project.mml
```diff
@@ -278,6 +278,24 @@ Layer:
...
     properties:
       minzoom: 8
+  - id: springs
+    geometry: point
+    <<: *extents
+    Datasource:
+      <<: *osm2pgsql
+      table: |-
+        (SELECT
+             St_Centroid(way) AS way, "natural"
```
kocio-pl: I'm not sure, but I suspect that this code makes test rendering on z14+ very long. Instead of 2-3 s it takes like 10+ s every time and there's heavy I/O load, which means big database query most probably.

This is nasty performance bug. Could you look closer and fix it?

matthijsmelissen: I think dropping St_Centroid should give a similar result?

Also this might result in the spring icon being displayed outside of the spring polygons in case of moon-shaped springs - do we want that? Dropping the St_Centroid should solve that too.

kocio-pl: This is beyond my knowledge, but sounds reasonable. I don't know how should the code look like after dropping it, could you show the proper code that I could test?

matthijsmelissen: Simply like this (like below):

way, "natural"

Make sure to test it on spring areas.

matthijsmelissen: What looks wrong according to you? Note that the text in the second item is not the same as the text in the first item.

kocio-pl: OK, it's just a mess with different interior algorithm again - when compared with my own master it looks OK, drinkable water name+icon is just moved with this area:

tnkhxidd

kocio-pl: For [all 6 such areas in Poland](http://overpass-turbo.eu/s/ypf) the icon is placed the same as before, so it looks like this is not even needed. I believe `marker-placement: interior` in MSS file does the job.

I'm not sure if we should wait for @Penegal with this or just update this line somehow and merge it?

kocio-pl: BTW: I hope soon we will have exactly the same interior algorithm in Kosmtik and on the OMSF servers, we're really close:
- https://github.com/kosmtik/kosmtik/pull/283 (and then a new Kosmtik release with npm package update, so our Docker image could be updated to...)
- https://github.com/openstreetmap/chef/issues/155#issuecomment-385318881

kocio-pl:
Thanks! This is a good update for probably the oldest leftovers that I hated the most...

kocio-pl:
This was both not needed and resulted in serious performance loss. See: https://github.com/gravitystorm/openstreetmap-carto/pull/3189#discussion_r185130323

Penegal:
OK, I just saw the reviews; sorry for the noise.

polarbearing:
How does this look within a body of water?

Penegal:
@polarbearing: as this has been merged and is part of v4.11.0 published 6 days ago, I would say that you'll see on the tiles on OSM.org. BTW, @kocio-pl, I'm surprised to not see the change on OSM.org tiles. I saw other style versions applied more quickly; is there a specific problem with this one?

polarbearing:
Ok [here is the case to watch](https://www.openstreetmap.org/#map=19/36.95184/-90.99300), Big Spring Missouri mapped as a node in a body of water.
Before:
big-spring-in-water_before

kocio-pl:
Some forks have included last release already:

https://tile.iosb.fraunhofer.de/#map=16/36.9517/-90.9932/3

I don't know why OSMF admins didn't push that change yet, you can ask them about it here: https://github.com/openstreetmap/chef/issues/162.

polarbearing:
Looking at the Fraunhofer server, the rendering on water is strange. 
On z15 I see the upper half of the halo, z16a white ring cut in the direction of the river line, in z17 a tiny remainder of the halo, in z18-19 it disappears completely. 

It is a pity that this was not tested beforehand, I had already mentioned that [here: 325 comment](325#issuecomment-369238650).

Should we reopen or make a new issue?

kocio-pl:
First make sure that it's really style issue, not the server rendering issue. Cut ring suggests me the later (for example if you change the layer to OSMde yow ill see no halo at all), so it's always best to test in your own environment.

polarbearing:
ok but in the .de style you do not see any icon. So 'invisible in water' remains. 
The 'after' examples above suggest that the river is painted over the icon. This might be nice for a line-only river to suggest in which direction it flows out, but it consequently fails for the water body, being the river outline or the lake around the icon.

kocio-pl:
I see. What fix would you propose for this?

kocio-pl:
Maybe adding white dot on top of the water would be good?

polarbearing:
The release has been merged by Tom now, maybe we wait and check the rendering there. 

I'll do some analysis meanwhile: 

The SVG is constructed atypically compared to other icons. It seems to have no artboard size, thus git shows the icon in a large transparent square, opening with Illustrator presents it on an A4 sheet, partially hanging over the border. 
It has 3 grouped paths being concentric rings. One is the blue ring (# 7abcec), the others are white (# ffffff). One of the white rings sits behind the blue and is thicker, thus constructs the 'halo'. The function of the third ring is unclear since it is thinner and behind the blue.
The SVG contains some 'transform' operations which apparently just shift the rings onto each other and could be avoided if they were drawn in the right place already. 

All new icons I have analysed so far use fill while the stroke is empty. This SVG is the opposite. 

Typically we have the icon black and define the colour in carto. 

The ring (# 7abcec) is darker than water (@water-color defined as # aad3df)

IIRC, the order of the 'id:' elements in the 'Layer:' section in the project.mml defines the painting order? If so, 'water-areas' are painted earlier than 'springs' which are earlier than 'water-lines'.

This explains  why a river line is painted over the ring, but I don't understand why the white rings disappear over a water area?


Finally I'd have the style for the spring rather in the water.mss with the natural things, while water-features.mss mostly has the man-made barrieres etc.

polarbearing:
[Fresh tile is there](https://c.tile.openstreetmap.org/19/129624/204155.png), spring invisible:

Issue #3159, opened 2018-04-01
Release v4.10.0


kocio-pl:
I plan to release v4.10.0 on 20.04. I hope #3099 will be merged soon and this will be quite important feature. Even more important would be #3061, but testing is needed to be sure.

Any other important things to be done before the release?

polarbearing:
I plan to progress the accommodation and museum outlines, however I'm offline for a a week until Apr 8.

andrzej-r:
#3061 is now #3163. I've rebased it to the current git master.

matthijsmelissen:
I agree it would be nice to release after #3163 has been merged.

kocio-pl:
I think this release is ready to be shipped feature and time-wise, but I'm a bit worried about performance. I have a feeling that some of the recent changes in master eats much more of CPU power. I will investigate it, but I need some help - could anyone else confirm this effect (and possibly find the source of this problem)?

kocio-pl:
It looks like the performance hit is triggered by #3101 and it's enough to just revert it to get rid of the problem.

matthijsmelissen:
Yes, this is a likely cause of performance problems. I dont access to a pc now, feel free to go ahead and reverse it.

How did you notice the performance issue? And can you see whether it is an issue on all zoom levels?

andrzej-r:
How long does it take to re-render all tiles in osm? I wonder if some tile servers still have old tiles or if we have a problem with race conditions in production environment.

The reason I suspect there may be an actual problem is that I've seen THREE different renderings of:

https://www.openstreetmap.org/way/170525607

- as building name, no icon or dot, just gray label 
- as amenity=bank, with brown icon and label
- as blue office dot and brown amenity label.

Some renderings are obviously undesirable, which may be a bug in itself. However, as there are three different ones, we are not just switching from one version to another but there is also an inconsistency in how tiles are rendered on different servers.

kocio-pl:
It depends not only on rendering, but also on caching servers. It might take few days, in general.

What is the real problem with different renderings? Blue dot and brown label is a bug for me, but two others seem to be proper rendering and this could only be a wrong tagging problem.

andrzej-r:
OK, it make sense to wait a bit more then for all rendering to finish.

[...]

PR #3182, opened 2018-04-20
Revert "Fix repetition of high-zoom admin/nature reserve labels"

kocio-pl:
Reverts #3101 because of the performance problem. See https://github.com/gravitystorm/openstreetmap-carto/issues/3159#issuecomment-382932958 for details.

PR #3163, opened 2018-04-03
Render offices as dots + names

andrzej-r:
Rebased #3061 to a new master branch. No changes to the code. Incremental commits merged into a single one. So far testing yields the same results as with #3061.

RE #108
Fixes #271
Fixes #1922
Fixes #2570

PR #3194, opened 2018-04-24
Fixed office/amenity conflict

### LINKED ISSUE #3195: Office + shop/amenity makes an icon very small
Office + shop/amenity makes an icon very small

See here: https://www.openstreetmap.org/node/2979269234#map=19/55.82968/37.30825

You can find a similar shop without an `office` tag nearby, its icon is normal.

andrzej-r:
RE: https://github.com/gravitystorm/openstreetmap-carto/issues/3159#issuecomment-383376498
Fixes #3195

POIs with both `amenity` and `office` tags triggered both code paths, resulting in points that shared some visual effects of both of these points.

Refactored the code to use `feature` field instead, as the coalesce() function guarantees it is unique. In the above example, `amenity` takes precedence over the `office` tag.

I've also, added extra checks in `addresses` and `building_text` layers so that they don't render labels when amenity/shop/office tags are defined.
In most (but not all) cases these labels were already hidden by Mapnik because they would have overlapped POI icons otherwise. However, even if not rendered, these labels would still be processed by Carto and Mapnik, increasing rendering time.

### DIFF
--- a/amenity-points.mss
+++ b/amenity-points.mss
@@ -1026,7 +1026,7 @@
   }
 
   // office points
-  [office != null][zoom >= 17] {
+  [feature = 'office'][zoom >= 17] {
     marker-width: 4;
     [zoom >= 18] {
       marker-width: 6;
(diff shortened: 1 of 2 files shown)


andrzej-r:
[Example](https://www.openstreetmap.org/#map=19/51.51633/-0.10012) illustrating the tag conflict and the fix.

error_z16
error_z17
error_z18
error_z19

amenity-points.mss
```diff
@@ -2328,63 +2328,79 @@
...
-      text-dy: 8;
-      text-fill: @office;
-      text-face-name: @standard-font;
-      text-halo-radius: @standard-halo-radius;
-      text-halo-fill: rgba(255, 255, 255, 0.6);
-      text-placement: interior;
+  [feature = 'office'] {
+    // potentially larger offices
+    [zoom >= 17] {
+      [office = 'administrative'],
```
nebulon42: Wouldn't this be prone to combinatory explosion problems (see https://github.com/mapbox/carto/issues/20)? This was the reason why `feature = office_` was introduced in the first place.

matthijsmelissen:
If we have rules for amenity=foo, leisure=bar and shop=baz, then the resulting XML ouput (which is the input fir Mapnik) will have one rule for each combination of amenity, leisure and shop (8 in total).

andrzej-r:
That should be sorted by having checked feature=* first, then.

How do I get to see the resulting XML, so that I can optimize its size?

nebulon42:
You don't need to worry about tiles, but just about the generated XML. You can see the generated XML if you manually invoke `carto`. E.g.:
```
npm install -g carto
cd openstreetmap-carto
carto project.mml > output.xml
```
You can find more details  at https://cartocss.readthedocs.io/en/latest/installation_usage.html.

Bascially you would need to look for a size explosion of the generated XML file before and after.

andrzej-r:
I run carto on several versions of the code. Below I have listed the number of lines of some of them:

v4.9.0: 47743
v4.10.0: 71903 (+24160)
v4.10.0 + this PR: 48556 (+813)
v4.10.0 + this PR without extra checks in `addresses` and `building_text` layers:  48554 (+811)

So, if anything it is the existing implementation that is causing the output to swell quite a bit. Probably related to the fact it was testing non-unique tags.

kocio-pl:
> v4.10.0: 71903 (+24160)

Oh, it looks like we have a problem here, maybe even of the same type (combinatory explosion). Could someone find the source of such a big size expansion?

andrzej-r:
It's definitely coming from the previous office tag implementation. Because:
1. Improving the implementation fixes it
2. Line counts just before and after the merge were respectively 48559 and 71903

But line count is not solely responsible for the execution time. For example, the revert (#3182) that fixed CPU use, had no impact on the xml file line count (71903 lines before and after).

By the way, the line count is calculated as part of Travis CI tests (with a slightly different formula and Carto version). I can only see the results of my PRs, but perhaps you can access all CI test results.

kocio-pl:
I see. It might be however the reason why the amount of dropped metatiles is ~1,5 times higher this time than in previous few releases:

https://munin.openstreetmap.org/openstreetmap/render.openstreetmap/renderd_processed.html

@matthijsmelissen @nebulon42 Please give a nod when this is ready in your opinion. We should make a v4.10.1 and I can do it, but I'd be happy if somebody else make the release.

matthijsmelissen:
> I've also, added extra checks in addresses and building_text layers so that they don't render labels when amenity/shop/office tags are defined.

This also drops house numbers from amenity tags that are not rendered. I don't think we want that.

In general, PRs should be as small as possible: it makes the code change easier to review, and it prevents situations like this one where a PR contains two changes of which I would like to merge only one of them.

matthijsmelissen:
> Could someone find the source of such a big size expansion?

This seems to be #3163, based on the stats provided by @andrzej-r: as we have a Carto rule for `[office != null]`, we get an additional XML rule for the combination of [office != null] with every individual feature. The current PR fixes this.

>  We should make a v4.10.1 and I can do it,

I don't think we necessarily need to make a v4.10.1 release. Rolling out a new release might put just as much stress on the server as just waiting a week or so until the next regular release.

kocio-pl:
This is a visual problem and (I suspect) also the performance problem. However we can for example make a regular (v4.11.0) release 3 weeks after the v4.10.0 (~4.05) instead of typical 4 weeks if you think it's not worth the hassle.

matthijsmelissen:
> Now, I am not particularly attached to this part of the fix so I am OK removing it, if you wish so. 

In any case, let's split it off to a different PR, so we can separate the discussion.

andrzej-r:
Done. I've tested it locally. It works the same as git master w.r.t. rendering building labels and unsupported amenity tags.

matthijsmelissen:
Thanks!

PR #3196, opened 2018-04-26
add office column to text-poly-low-zoom layer to avoid SQL errors

nebulon42:
I was getting the following SQL error when running the style:

```
kosmtik_1  | Postgis Plugin: ERROR:  column "office" does not exist
kosmtik_1  | LINE 1: ...ELECT ST_AsBinary("way") AS geom,"feature","name","office","...
kosmtik_1  |                                                              ^
kosmtik_1  | in executeQuery Full sql was: 'SELECT ST_AsBinary("way") AS geom,"feature","name","office","way_pixels" FROM (SELECT
kosmtik_1  |     way,
kosmtik_1  |     way_area/NULLIF(9783.94::real*9783.94::real,0) AS way_pixels,
kosmtik_1  |     COALESCE(
kosmtik_1  |       'landuse_' || CASE WHEN landuse IN ('forest', 'military', 'farmland') THEN landuse ELSE NULL END,
kosmtik_1  |       'natural_' || CASE WHEN "natural" IN ('wood', 'glacier', 'sand', 'scree', 'shingle', 'bare_rock',
kosmtik_1  |                                             'water', 'bay') THEN "natural" ELSE NULL END,
kosmtik_1  |       'place_' || CASE WHEN place IN ('island') THEN place ELSE NULL END,
kosmtik_1  |       'boundary_' || CASE WHEN boundary IN ('national_park') THEN boundary ELSE NULL END,
kosmtik_1  |       'leisure_' || CASE WHEN leisure IN ('nature_reserve') THEN leisure ELSE NULL END
kosmtik_1  |     ) AS feature,
kosmtik_1  |     name,
kosmtik_1  |     CASE WHEN building = 'no' OR building IS NULL THEN 'no' ELSE 'yes' END AS is_building -- always no with the where conditions
kosmtik_1  |   FROM planet_osm_polygon
kosmtik_1  |   WHERE (landuse IN ('forest', 'military', 'farmland')
kosmtik_1  |       OR "natural" IN ('wood', 'glacier', 'sand', 'scree', 'shingle', 'bare_rock', 'water', 'bay')
kosmtik_1  |       OR "place" IN ('island')
kosmtik_1  |       OR boundary IN ('national_park')
kosmtik_1  |       OR leisure IN ('nature_reserve'))
kosmtik_1  |     AND building IS NULL
kosmtik_1  |     AND name IS NOT NULL
kosmtik_1  |   ORDER BY way_area DESC
kosmtik_1  | ) AS text_poly_low_zoom WHERE "way" && ST_SetSRID('BOX3D(-2504688.542848659 -2504688.542848658,7514065.628545969 7514065.62854597)'::box3d, 3857)'
```

I addded a `NULL as office` column to the `text-poly-low-zoom` layer as it is also used in the selector where a `[office != null]`filter is used.

### DIFF
--- a/project.mml
+++ b/project.mml
@@ -1969,6 +1969,7 @@ Layer:
               'leisure_' || CASE WHEN leisure IN ('nature_reserve') THEN leisure ELSE NULL END
             ) AS feature,
             name,
+            NULL as office,
             CASE WHEN building = 'no' OR building IS NULL THEN 'no' ELSE 'yes' END AS is_building -- always no with the where conditions
           FROM planet_osm_polygon
           WHERE (landuse IN ('forest', 'military', 'farmland')


matthijsmelissen:
Just to be clear, does this mean the released version is broken for some zoom levels?

nebulon42:
Honestly I don't know. When I wanted to display the style I got this error. I'm not sure why nobody else has seen it up until now.

andrzej-r:
Does #3194 fix this problem in your setup? In that PR I no longer test for office! = null, as it was causing issues when both office and amenity tags were both set to certain combinations of values.

Can you share a specific scenario (location and zoom) that triggers this issue?

nebulon42:
@andrzej-r No, it wouldn't because the `office = ...` filter is still there and you don't add a `office` column for text-poly-low-zoom. You can trigger the problem if you are on smaller zoom levels. I think I was on z4 when this happened. It does not matter if you have a planet loaded or just a small extract. When Mapnik pulls data for text-poly-low-zoom from PostgeSQL it realizes that the office column is missing.

andrzej-r:
I've tested both the current master version and #3194 with `greater-london-latest.osm.pbf` (the same dataset I used for the development) - they both work fine at all zoom levels in kosmtik. But at z4 this dataset produces only one label ("London").

Anyway, in `amenity-points.mss` the `.text-low-zoom` class is indeed being checked at zoom<10 for `office=` tag values. So this fix is valid and should be applied to current master. Even though I couldn't reproduce the error with my data I can see how the code could fail with larger datasets and break rendering at zoom<10. If this was the case, I'd recommend releasing a 4.10.1 version.

matthijsmelissen:
I cannot reproduce this error message.

@nebulon42 Could this depend on the carto version? We have a rule `[office != null][zoom >= 19]`, could it be that some versions of carto only include the office column in the SQL for the rules for zoom level 19 and higher? I'm using the carto defined in our Docker container.

I don't think it can possible depend on the (size of the) dataset: it fails on the query, which is generated before the data is read.

I can merge this if carto needs it for technical reasons, however strictly speaking I don't think it is necessary.

kocio-pl:
I have tested current master (without this patch) with Poland from z16 down to z0 and nothing was broken. I was using current Kosmtik (from master branch on git repo) which is using Mapnik 3.0.18.

What version of Kosmtik and Mapnik are you using to reproduce this error?

kocio-pl:
The same result with Docker. I have build the images, imported Liechtenstein data and there was no problem with rendering on any zoom level. This was Kosmtik 0.0.16 and Mapnik 3.0.12, if I remember.

andrzej-r:
I think this code is valid and can be merged to make the SQL code more robust against other ways of testing for office= tags. I would have done it myself if I spotted it earlier.

As for if it crashes OSM tile rendering - we should probably ask the user instead. The error can only occur at z<10 and these zoom levels are only rendered once after OSM-Carto has been released so the damage, if happened at all, is somewhat contained.

kocio-pl:
The downside of adding things is more code to maintain and maybe eating more resources, so I want to be sure that it fixes something real.

nebulon42:
I have also used the Docker environment which runs Kosmtik 0.0.16 and thus carto 0.18.x. I have not touched the style before for a while and I freshly pulled the master branch and started it up when I saw the error. As far as the data is concerned I have loaded Austria from Geofabrik. As I was missing a localconfig I think Kosmtik started on z4 (or lower) at 0,0 coordinates.

I don't think that carto has to do something with it as it just takes the query and puts it into the appropriate layers in the Mapnik XML. In the other patches I did I also did not encounter the problem, but I was looking at higher zoom levels there.

From a technical side the explanation is quite straightforward though. The layer selector uses both the high zoom and low zoom text layer and in the filters the office column is used which is present in only one layer query as column. Why we don't hit the error consistently I don't know. Maybe it is a Mapnik bug/feature that has something to do with minzoom/maxzoom for layers.

kocio-pl:
Could you post an XML file, so I could test it?

nebulon42:
Yes, there is no XML file. I don't think we have to make this complicated. I saw an error and fixed it. If you think it is not needed or you don't hit the error, then let's leave it out. I personally would merge it.

kocio-pl:
I would like to investigate, because it's suspicious that it doesn't appear at my setups if that's so obvious problem. Therefore I won't merge it myself, I leave the decision and action to you (I'm not sure what @matthijsmelissen thinks about merging).

matthijsmelissen:
> Yes, there is no XML file. I don't think we have to make this complicated. I saw an error and fixed it. If you think it is not needed or you don't hit the error, then let's leave it out. I personally would merge it.

@nebulon42 I know it's frustrated when discussions are dragged out about such a seemingly simple thing, but I think some further investigation is worth it. First of all, I am really not convinced this is an bug as I explained above. Second of all, I prefer not to apply solutions for errors that I cannot reproduce: because if we apply the solution now, a similar bug might be reintroduced with one of the next pull requests. Third, as others seem unable to reproduce this, I am not fully convinced the error was caused by the master branch, rather than by some change you might have (inadvertently) made.

Can you reproduce the issue yourself?

matthijsmelissen:
@kocio-pl Could you perhaps check what SQL query you get in the XML for the text_poly_low_zoom layer when you run carto directly? Does this query contain 'office' for all zoom levels?

I'm unable to run carto directly at the moment.

kocio-pl:
Here is the resulting file:

https://gist.github.com/kocio-pl/2d8d321ed036be3d41486ece0ab320d4

matthijsmelissen:
Seems I forgot how this all works. The columns that are obtained from the SQL query are inserted by Mapnik, not by carto, and therefore are not defined in this file. Therefore we can rule out an issue in carto.

So now is the question, why does Mapnik behave differently sometimes?

nebulon42:
I'll do some more testing on my side when I have time.

matthijsmelissen:
As nobody else has been able to reproduce this bug, I'm going to close this PR. I'm not saying that the bug doesn't exist, but at this point there's no way for me to confirm it.

If @nebulon42 or anybody else is able to reliably confirm this bug, we can of course reopen this.

PR #3197, opened 2018-04-26
fix trailing whitespace in project.mml and .mss files

nebulon42:


nebulon42:
I have extended this to other mss files to clean up my other PR.

Issue #3105, opened 2018-03-06
Code needed to render prepared icons [meta ticket]


kocio-pl:
We have lot more icons for requested features lately, mainly due to a great designing work of @Tomasz-W (thanks!). The problem is that they will be not shown until somebody do the coding. Some of them require  just "copycat" skills and running a testing environment ([in Docker](https://github.com/gravitystorm/openstreetmap-carto/blob/master/DOCKER.md) probably). Some are more complex, but one still doesn't have to be a programmer to make it (especially after reading [some background intro](https://wiki.openstreetmap.org/wiki/Standard_tile_layer#Technical_details), [tutorials](https://ircama.github.io/osm-carto-tutorials/) or a [real life example](https://www.openstreetmap.org/user/SomeoneElse/diary/43041)) and I'm willing to help.

Just please remember that after PR (pull request) is ready, you still need to interact with people as it's very common that people have different wishes, requests or questions, even when the case seems like very obvious (see [this example](https://github.com/gravitystorm/openstreetmap-carto/pull/3085)).

Here is the (open!) list of things which I feel are ready to prepare a PR and test rendering a bit:

Easy ones (one feature per ticket, no subtags):
- [x] Add rendering for shop=dairy #3087 
- [x] Add rendering for amenity=shower #3067 
- [x] Add rendering for leisure=sauna #3062 
- [x] Rendering of amenity=bbq #2996
- [ ] <s>Render amenity=studio #2518</s>
- [ ] Peaks and saddles are hardly visible #2848 
- [ ] Names for wind turbines #2734 
- [x] Add icon for shop=music #2929 
- [ ] Add icon for shop=ticket #3115  
- [x] render label for historic=city_gate #152 

Complex ones (multiple objects with subtags):
- [x] Guidepost and Map Icon #1978 #3246
- [ ] Add rendering for barrier points #846 
- [ ] Distinguish different memorial types #3088

Tomasz-W:
A few days ago I though about a tag like "pr needed", which could be added to mark issues where we all agree that icon/ icons are ok, and just wait for someone to write the code. I think this would be easier to manage than a next one issue with task list.
Anyway - thank you for sorting things out here :)

kocio-pl:
I hope more personal touch and a few words of explanation might work better than just a tag.

@althio @RobJN Would you like to pick anything from that list?

Tomasz-W:
Another ones:

single icons:
[...]

Issue #1978, opened 2015-11-15
Guidepost and Map Icon


g4rpch:
Dear All
From my point of view the icon of guidepost and maps could be improved, with something more clear, like:

guidepost: http://wiki.openstreetmap.org/w/images/9/91/Guidepost_doubleguidepost.png
map: http://arcgis.dnr.state.mn.us/ewr/whaf/Explore/img/basemap.png

links are only examples... but would be nice to have an icon more specific.

G4rp

HolgerJeromin:
just for reference: you probably mean osm elements with `information=guidepost` and `information=map`

kocio-pl:
[Another feature](https://github.com/gravitystorm/openstreetmap-carto/issues/1957) that should be handy in the outdoor, but in urban areas would make things worse if we use the uniform zoom level for both.

math1985:
See also #304 and #324.

Tomasz-W:
I don't like that all `information=*` tags are rendered the same, so I've tried to distinguish them:
* information=office
![information office](https://user-images.githubusercontent.com/25656654/36440631-4b7d5e40-1670-11e8-8fa6-ee8eac34f5a8.png)
* information=audioguide
![information audioguide](https://user-images.githubusercontent.com/25656654/36440636-5075d3f0-1670-11e8-8a92-80b23e5aebb0.png)
* information=guidepost
![information guidepost](https://user-images.githubusercontent.com/25656654/36441000-96de2166-1671-11e8-9b9f-bc0e9c29b443.png)
* information=map
![information map](https://user-images.githubusercontent.com/25656654/36441785-f203c288-1673-11e8-8c0e-7ddcde63eee0.png)
* information=board
![information board](https://user-images.githubusercontent.com/25656654/36441896-43bf3990-1674-11e8-830a-63c02a2e68d9.png)

For terminal, tactile_map and tactile_model I have no idea.

geowas-github:
All 5 of them work good for me!

Tomasz-W:
I got some ideas for 3 tags which I mentioned above. 
What about make board icon square instead of rectangle, and use it for terminals and tactile models? For tactile map I think we could use the same icon as for information=map.

kocio-pl:
I can't imagine the outcome, would be good to just see it.

Tomasz-W:
* information=board
[...]

Issue #304, opened 2014-01-04
Add label for tourism=information + information=office


mrwojo:
We don't show the name for `tourism=information`, instead we show only an "i" symbol. That's probably sufficient for most features with this tag, which are mainly [guideposts, boards, or maps](http://taginfo.openstreetmap.org/keys/information#values).

However, `tourism=information` can also indicate more significant visitor center buildings that should show the name. When you tag a building with `tourism=information`, its name disappears, blocked by the "i" symbol. [`information=office`](https://wiki.openstreetmap.org/wiki/Key:information) seems to be the common tag to identify these features.

Example where a name should be rendered: [Way 112654444](http://www.openstreetmap.org/way/112654444) (Kings Canyon National Park Visitor Center in California)

PR #3198, opened 2018-04-26
differentiate guidepost from general tourism=information display

nebulon42:
Ref #1978, #3105.

Adds a guidepost icon on z19 and drops rendering of guidepost nodes before z19.

before z18
z18_before

after z18
z18_after

before z19
z19_before

after z19
z19_after

### DIFF
--- a/project.mml
+++ b/project.mml
@@ -1500,6 +1500,7 @@ Layer:
             tags->'tower:construction' as "tower:construction",
             tags->'tower:type' as "tower:type",
             tags->'castle_type' as castle_type,
+            tags->'information' as information,
             CASE WHEN shop IN ('supermarket', 'bag', 'bakery', 'beauty', 'bed', 'books', 'butcher', 'clothes', 'computer',
                                'confectionery', 'fashion', 'convenience', 'department_store', 'doityourself', 'hardware', 'fishmonger', 'florist',
                                'garden_centre', 'hairdresser', 'hifi', 'ice_cream', 'car', 'car_repair', 'bicycle', 'mall', 'pet',
@@ -1628,6 +1629,7 @@ Layer:
             tags->'tower:construction' as "tower:construction",
             tags->'tower:type' as "tower:type",
             tags->'castle_type' as castle_type,
+            tags->'information' as information,
             CASE WHEN shop IN ('supermarket', 'bag', 'bakery', 'beauty', 'bed', 'books', 'butcher', 'clothes', 'computer',
                                'confectionery', 'fashion', 'convenience', 'department_store', 'doityourself', 'hardware', 'fishmonger', 'florist',
                                'garden_centre', 'hairdresser', 'hifi', 'ice_cream', 'car', 'car_repair', 'bicycle', 'mall', 'pet',
@@ -2081,6 +2083,7 @@ Layer:
             tags->'office' as office,
             tags->'recycling_type' as recycling_type,
             tags->'castle_type' as castle_type,
+            tags->'information' as information,
             ref,
             way_area,
             CASE WHEN building = 'no' OR building IS NULL THEN 'no' ELSE 'yes' END AS is_building
@@ -2168,6 +2171,7 @@ Layer:
             office,
             recycling_type,
             castle_type,
+            information,
             ref,
             way_area,
             is_building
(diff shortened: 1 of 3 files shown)


kocio-pl:
If it's Atom, then you can turn it off by disabling Preferences -> Whitespace -> Remove Trailing Whitespace.

kocio-pl:
Link for testing place:

https://www.openstreetmap.org/#map=19/46.90772/13.53294

SomeoneElseOSM:
Excellent idea - I'd also suggest handling trail_blaze and route_marker (see [taginfo](https://taginfo.openstreetmap.org/keys/information#values)) as well as guidepost since they're even less signification.  See https://map.atownsend.org.uk/maps/map/map.html#zoom=17&lat=-24.999094&lon=135.17657 (and the link via the changelog to the repository) for what I went with here, but ignore the PNFS and Sustrans markers since they're UK-specific.

matthijsmelissen:
Could you perhaps first fix the whitespace issues in a separate PR?

nebulon42:
@matthijsmelissen Done so now in #3197. and @kocio-pl also opened #3200.
@SomeoneElseOSM Definitely worth handling, but better in a separate PR.

jragusa:
Several guidepost dedicated for hiking in both France and Switzerland (I don't know for other countries but I suppose too) mention a name and even an elevation. Would it be possible to add these features to the icon ?

nebulon42:
@kocio-pl Here are some instances in Austria: https://overpass-turbo.eu/s/ymd.

nebulon42:
Additionally now the name and/or the elevation is displayed:
guidepost_name
guidepost_name_ele
guidepost_ele

kocio-pl:
Sorry, I wanted the photo example and some explanation, not the link to data set. :smile: 

If I get it right, they show the real elevation they are located at? But what is their name then? Isn't it the same as name and elevation of the peak or saddle for example?

nebulon42:
The name is usually a place name, but can also be the name of the saddle and/or peak. Elevation is of course the real elevation. The important part here is that both name and elevation are signed on the guidepost as can be seen in e.g. https://wiki.openstreetmap.org/wiki/File:Signpost.jpg.

kocio-pl:
I see. OK, so I like this solution. After all they won't eclipse peaks or saddles, because they will be rendered very late.

kocio-pl:
Maybe it would better to make the code more compact and easier to extend with other types, like:

```css
[feature = 'tourism_information'][information != 'guidepost'][zoom >= 17],
[feature = 'tourism_information'][information = 'guidepost'][zoom >= 19] {
  marker-file: url('symbols/information.12.svg');
  [information = 'guidepost'] { marker-file: url('symbols/guidepost.svg');
  marker-placement: interior;
  marker-fill: @amenity-brown;
  marker-clip: false;
}
```

Tomasz-W:
@nebulon42 Do you have plans to take care of rest of tourism=information icons https://github.com/gravitystorm/openstreetmap-carto/issues/1978#issuecomment-385373937 and/ or other related issues https://github.com/gravitystorm/openstreetmap-carto/issues/304 https://github.com/gravitystorm/openstreetmap-carto/issues/324 ? It would be a big progress in tourism-elements rendering in Carto.

nebulon42:
@kocio-pl I have updated the code to be more DRY.
@Tomasz-W I have no immediate plans, but I'll put it on my list. :)

project.mml
```diff
@@ -1969,6 +1971,7 @@ Layer:
               'leisure_' || CASE WHEN leisure IN ('nature_reserve') THEN leisure ELSE NULL END
             ) AS feature,
             name,
+            tags->'information' as information,
```
kocio-pl: What's this line for in `text-poly-low-zoom` layer? I have tested that it works without this code as well on all the zoom levels.

nebulon42: Same thing like #3196. :)

kocio-pl: Oh, boy, not again... :smile:

Did you have a problem without this code or it's just in case? Did you try with updated Docker image (I mean with a new Kosmtik :smile: ).

I'm ready to merge this code without this line any time. Especially since production rendering servers have no such problem and there are no other mid/high zoom features that I'm aware of which need such addition. Would you like to split it, so we could discuss the low zoom mystery separately?

nebulon42: I'll do some tests again when I have time.

kocio-pl:
What do you think about adding a mysterious code line?

kocio-pl:
@matthijsmelissen ping?

matthijsmelissen:
@nebulon42 Can you get it to work without the 'mysterious line' as well? If so, I'd prefer to merge that version.

nebulon42:
I have removed the line but did not do additional testing.

kocio-pl:
Thanks, I've tested it and it works.

PR #3200, opened 2018-04-27
Fixing trailing whitespaces in amenity-points.mss

kocio-pl:
Just fixing trailing whitespaces in amenity-points.mss by turning on automatic fixing in Atom (Settings > Packages >whitespace > Enable) for a moment - it's recommended to be disabled by default when creating patches.

Related to https://github.com/gravitystorm/openstreetmap-carto/pull/3198#issuecomment-384909674

### DIFF
--- a/amenity-points.mss
+++ b/amenity-points.mss
@@ -72,7 +72,7 @@
     marker-placement: interior;
     marker-clip: false;
   }
-  
+
   [feature = 'amenity_bbq'][zoom >= 17] {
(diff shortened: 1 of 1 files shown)


nebulon42:
Thanks, I have now also included this into #3197 which also includes other files.

nebulon42:
Thanks for following up with this, but since #3197 addresses all files together I thought it might be better to have everything in one commit.

Issue #3064, opened 2018-02-11
Highway outline?


LucFreitas:
![capturar](https://user-images.githubusercontent.com/19976610/36075555-c2a031d6-0f37-11e8-91b0-34b3728e55b2.PNG)

https://www.openstreetmap.org/?mlat=43.83462&mlon=20.04059#map=19/43.83462/20.04059

HolgerJeromin:
The "both" center lines are one osm way:
https://www.openstreetmap.org/way/301484309

manfredbrandl:
I broke up the way and it should look nicer soon.

kocio-pl:
However this might be a Mapnik error - you can try report it.

matthijsmelissen:
Looks like a Mapnik error to me too. I don't think there's anything we can do to fix this on our side.

kocio-pl:
I have found another place where this is visible and might be useful for testing:

![screenshot-2018-4-20 openstreetmap](https://user-images.githubusercontent.com/5439713/39029035-1c026d00-445a-11e8-9dad-d13f95f502b0.png)


https://www.openstreetmap.org/#map=19/54.40491/18.67671

matthijsmelissen:
It was in fact an issue on our side. Fixed by #3202.

LucFreitas:
Thanks!

PR #3202, opened 2018-04-27
Bugfix primary road casing

sommerluk:
Before:
screenshot 1
The road casing for primary roads is wrong (not round like motorway, trunk, secondary…). This is because of a small bug (in a layer name) in `roads.mss`.

After:
screenshot 2


kocio-pl:
Could you test if it resolves #3064 examples?

matthijsmelissen:
Yes, that's the same issue.

Issue #331, opened 2014-02-07
Add rendering for historic=ruins


EdLoach:
Neither POI or area/multipolygon ruins are currently rendered. http://www.openstreetmap.org/note/114642 brought this to my attention, as the castle named in the note is already mapped as both POI and multipolygon.

For POI, perhaps just the name rendered, and for areas - well I'll let you pick a suitable colour. The wiki also suggests a lot of ways are also tagged as ruins - perhaps for these render a bit like generic barriers?

kocio-pl:
I guess dashed outline (as mentioned [here](https://github.com/gravitystorm/openstreetmap-carto/issues/2093#issue-142191668), citing OsmAnd) could work as a visual hint.

nebulon42:
I don't think that works for `historic=ruins`. AFAIK OsmAnd uses an icon (shown in the screen) for `historic=ruins`. A dashed outline makes only sense for `building=ruins` (see https://github.com/gravitystorm/openstreetmap-carto/issues/1898).

dieterdreist:
sent from a phone

> Il giorno 24 set 2016, alle ore 21:07, Michael Glanznig <notifications@github.com> ha scritto:
> 
> I don't think that works for historic=ruins.


what s the difference to archaeological site? Maybe we can use the same rendering?

kocio-pl:
What about this icon of ruined castle, [proposed for Osmic](https://github.com/gmgeo/osmic/pull/57) by @MaestroGlanz:

https://github.com/MaestroGlanz/osmic/blob/b6f0eaa85eedfc453e5c9b8b19ca8d1d3a89e54d/tourism/castle-ruins.svg

kocio-pl:
[Example](https://www.openstreetmap.org/node/401359943) rendering on z17:
![yfahggcc](https://user-images.githubusercontent.com/5439713/30774493-96b8592c-a083-11e7-98f1-42a2e1e7d4d3.png)

polarbearing:
The separate shape above is implausible, what is it supposed to be?

kocio-pl:
I agree that ruined castle icon is not working in practice. I guess we can use the idea of decomposed shape for something simpler, like house.

MaestroGlanz:
It is directly derived from the castle icon. I could change it to a version where the parts are joined.

kocio-pl:
Please try, we're still looking for something good enough.

SomeoneElseOSM:
[...]

PR #3204, opened 2018-04-27
do not show castle/manor/fort symbols when ruins tag is set

nebulon42:
Do not show castle/manor/fort symbols when the ruins tag is set. The symbols convey that the feature is in a good shape. But the `ruins` tag has a very wide definition. It could be that there are just some stones left or that still most of the building is there. I think it is better to not show symbols and text for ruins until there is a ruins symbol.

Those two not-in-very-good-shape castles (one more in the "some stones left" category) in the preview started appearing after the symbols were rendered.

https://www.openstreetmap.org/#map=16/46.9613/13.5035

Before
z16_before

After
z16_after

### DIFF
--- a/amenity-points.mss
+++ b/amenity-points.mss
@@ -684,16 +684,16 @@
     marker-clip: false;
   }
 
-  [feature = 'historic_fort'][zoom >= 16] {
+  [feature = 'historic_fort'][ruins = null][zoom >= 16] {
     marker-file: url('symbols/fort.svg');
     marker-fill: @memorials;
     marker-placement: interior;
     marker-clip: false;
   }
 
-  [feature = 'historic_castle'][castle_type != 'stately'][zoom >= 15],
-  [feature = 'historic_castle'][castle_type = 'stately'][zoom >= 16],
-  [feature = 'historic_manor'][zoom >= 15] {
+  [feature = 'historic_castle'][ruins = null][castle_type != 'stately'][zoom >= 15],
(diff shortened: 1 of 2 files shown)


dieterdreist:
sent from a phone

> On 27. Apr 2018, at 21:40, Michael Glanznig <notifications@github.com> wrote:
> 
> do not show castle/manor/fort symbols when ruins tag is set


if we remove the icon from ruins we should show a different icon for castle ruins, but I would be reluctant in general to support old style status/lifecycle tags. This is not a good approach in general, because you have to evaluate 2 different tags together to make sense of it, and the ruins tag basically states that the castle is not a castle.

kocio-pl:
This is a good question - how should we handle `historic=castle`+`ruins=*` (see https://taginfo.openstreetmap.org/keys/ruins#combinations ) and `historic=ruins` (see #331 and https://taginfo.openstreetmap.org/tags/historic=ruins#combinations ) then?

matthijsmelissen:
I disagree with this PR. If there is so little of the building left that it cannot be considered a castle, it should not have a castle tag. If it is still considered a castle, we can render it.

nebulon42:
Fine, let's keep it then.

dieterdreist:
sent from a phone

> On 27. Apr 2018, at 23:14, kocio-pl <notifications@github.com> wrote:
> 
> This is a good question - how should we handle historic=castle+ruins=*


I thought there was agreement not to encourage usage of qualifiers like abandoned, disused or ruins as additional keys, but rather as prefixes.

Issue #3128, opened 2018-03-18
exposing port 5432 in docker


james2432:
Would there be any opposition to me exposing the 5432 port for the database in the docker config or is this a security reason?

It simplifies connecting to the database with local tools such as pgadmin and not have to run queries through the ugly cli

kocio-pl:
You can always do it by adding a proper line into `docker-compose.yml`. The question is if we want this port to be visible by default? I guess not, because if somebody uses standard PostgreSQL port in her system, this would mean conflict - two apps/instances can't use the same port at the same time.

james2432:
We could also make it as a comment to help new comers not have to read docker documentation to get started

kocio-pl:
Well, we could do that, but I don't see the generic need to export the database ports, it's rather quite specific. Why do you need to make direct database queries, by the way?

james2432:
to test that queries are working before adding them to the project.mml instead of testing them via kosmtik. If they were atleast commented out it would help new people know where to allow connections to PGsql. If an icon doesnt show up it's easy to debug if it's your sql or your CSS selectors where the bug is

kocio-pl:
Well, you don't need that - just turn on the Kosmtik "Data inspector" tool (you can even select some layers) and you will see which data associated with the objects are selected - look here on top:

https://github.com/kosmtik/kosmtik#screenshot

kocio-pl:
I think this is both too specialized and not needed by default to test data, so I'm closing it.

PR #3208, opened 2018-05-01
Docker: enable access to the postgresql database from the host and restrict port sharing to localhost only

nebulon42:
- This exposes the port of PostgreSQL to the host to enable access to the database from there with management tools.
- Limit port sharing to localhost only. I have seen on a production server that Docker is able to punch a hole in the firewall if you are not careful. While this is not the main scenario for our Docker setup it does not hurt to be careful.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -10,7 +10,7 @@ services:
     depends_on:
       - db
     ports:
-      - "6789:6789"
+      - "127.0.0.1:6789:6789"
     environment:
       - PGHOST=db
       - PGUSER=postgres
@@ -19,6 +19,8 @@ services:
     build:
       context: .
       dockerfile: Dockerfile.db
+    ports:
+      - "127.0.0.1:5432:5432"
     environment:
       - PG_WORK_MEM
       - PG_MAINTENANCE_WORK_MEM


kocio-pl:
Restricting Kosmtik access makes sense for me, but what's the motivation behind exposing database port to the host? I rejected such idea in #3128, but maybe I was wrong and that's needed for more developers.

nebulon42:
For me the use case was to do testing when preparing the release of Kosmtik. Since I moved to the docker setup I have no dedicated PostgreSQL database running on my machine any more. What was said in #3128 would be also valid use cases. Inspect and change the data where necessary. I have done so several times in the past when testing features that have not been in my extract.

I'm not so much concerned about having other databases running, that should be a minority and it would be easy to fix. Maybe we should adopt the approach outlined in https://docs.docker.com/compose/extends/ which would cover all those modifications?

kocio-pl:
> I currently run Kosmtik inside a virtualbox, and access it on the host machine. This will break that. But I agree the proposed change is the expected behaviour.

There can be more people with such setup (Windows users with the older systems don't have Docker), so we have security vs usability problem. Since this is a tool for novice users, I would rather choose usability.

matthijsmelissen:
The fact that it's a tool for novice users would be even more reason to choose security for me :)

kocio-pl:
Maybe I'm wrong - what IP address you use? Probably Kosmtik server inside VirtualBox is not seen on the host as http://127.0.0.1:6789 anyway.

matthijsmelissen:
I use the local network IP address that Virtualbox gets through DHCP - in my case http://192.168.2.14:6789/.

kocio-pl:
It probably means that it won't create support questions, so I'm OK with this change. Since I'm not against locally exposing database (even if I use local database on port 5432), I guess we can merge this code.

SomeoneElseOSM:
Trying to use port 5432 on the host _by default_ is surely a bad idea?  I'm assuming that is why I've just got this error:

    ERROR: for db  Cannot start service db: driver failed programming external connectivity on endpoint openstreetmapcarto_db_1  (d086456c567939220bb4d20eb24605cad00ae44164541a79803cf69a84e0d18d): 
    Error starting userland proxy: listen tcp 127.0.0.1:5432: bind: address already in use
    ERROR: Encountered errors while bringing up the project.

It'll stop anyone trying this container if they've got postgres installed locally for _any_ other purpose, won't it?  Apologies if I've got hold of the wrong end of the stick here...

kocio-pl:
> It'll stop anyone trying this container if they've got postgres installed locally for any other purpose, won't it? 

Sure, see my last remark. That's why I was thinking if it's worth merging.

SomeoneElseOSM:
> Sure, see my last remark. That's why I was thinking if it's worth merging.

If you mean https://github.com/gravitystorm/openstreetmap-carto/pull/3208#issuecomment-387253266 then I don't understand what you mean (and why that comment suggests it was worth merging).

kocio-pl:
Well, I meant that I use local database, but I can live with that problem, since I think this might be quite rare case. What do you think about it?

SomeoneElseOSM:
Document a workaround that someone completely unfamiliar with docker could follow?

kocio-pl:
Sounds good for me, would you like to write it?

SomeoneElseOSM:
I'm "someone completely unfamiliar with docker" :)

I'm trying to put something together that can go on the switch2osm site but docker's own documentation is pretty dire, and Docker.md here is great if you already have it installed but not if you haven't.

kocio-pl:
I think the simplest workaround would be to change `docker-compose.yml`, just removing these lines:

https://github.com/gravitystorm/openstreetmap-carto/blob/bb8eebe2f90a342378a400b17962e77d2949ea70/docker-compose.yml#L22-L23

SomeoneElseOSM:
@kocio-pl To be clear, are you asking me to submit a pull request that removes those lines in this repository?  Or asking me to see if removing those lines from docker-compose.yml locally in src/openstreetmap-carto/docker-compose.yml fixes the problem, or something else?

kocio-pl:
I'm asking you to make the PR with a documentation - of course it would be good to test it beforehand.

nebulon42:
Maybe https://docs.docker.com/compose/extends/ helps you? Especially the `docker-compose.override.yml`, which could be used to customise the port.

matkoniecz:
> I think the simplest workaround would be to change docker-compose.yml, just removing these lines:

It is unclear to me is it going to cause any security issues, that were motivation for making this PR.

I admit that I am doing this here, but I am not a fan of disabling security features till my software seems to work.

matkoniecz:
> Maybe https://docs.docker.com/compose/extends/ helps you? Especially the docker-compose.override.yml, which could be used to customise the port.

I see how can I overwrite port selection, but it is not clear for me which port value should be used.

PR #3211, opened 2018-05-02
Added rendering of nursing_home, childcare and driving_school amenities

andrzej-r:
Trying to bulk fill some gaps in popular amenity tags. No icons at this time, feel free to provide some in this PR or separately.

1) amenity=nursing_home and amenity=childcare 
Uses the same colours and font rendering as amenity=kindergarten, but unlike it, it is rendered at higher zoom level (small dot at z17, dot + name at z18+). No icon at this time.

[Example](https://www.openstreetmap.org/#map=19/51.52223/-0.06708):
amenity1_childcare

2) amenity=driving_school
Uses shop colours. As above, small dot at z17, dot + name at z18+. No icon at this time.

[Example](https://www.openstreetmap.org/#map=19/51.52177/-0.11301):
amenity1_driving_school


matthijsmelissen:
Thanks for your proposal!

I'm not sure about amenity=nursing_home, hasn't the nursing home tag been replaced by amenity=social_facility + social_facility=nursing_home ?

I'm also not sure if a driving school should be rendered like a shop. Perhaps rendering it like an office is more fitting?

andrzej-r:
Indeed, "nursing home" preset in iD is now amenity=social_facility + social_facility=nursing_home. I haven't noticed that and there was no mention of deprecation in amenity=nursing_home. Having said that, it is still a very popular tag (14k uses of amenity=nursing_home vs 3k uses of social_facility=nursing_home). What is our policy on deprecations? Are we trying to actively promote/demote certain tags or simply render what people are using?

As for driving schools, IMHO both shop and office are OK (or perhaps amenity-brown like in shop=car_repair?). I opted for a shop because driving schools sell services and that's one of differentiating factors between amenities/shops and offices.

dieterdreist:
2018-05-03 1:54 GMT+02:00 Andrzej <notifications@github.com>:

> Indeed, "nursing home" preset in iD is now amenity=social_facility +
> social_facility=nursing_home. I haven't noticed that and there was no
> mention of deprecation in amenity=nursing_home. Having said that, it is
> still a very popular tag (14k uses of amenity=nursing_home vs 3k uses of
> social_facility=nursing_home). What is our policy on deprecations? Are we
> trying to actively promote/demote certain tags or simply render what people
> are using?
>
> As for driving schools, IMHO both shop and office are OK (or perhaps
> amenity-brown like in shop=car_repair?). I opted for a shop because driving
> schools sell services and that's one of differentiating factors between
> amenities/shops and offices.
>


while I amended social_facility=nursing_home to the social facility scheme
I am still not really convinved of the scheme in general, there is no
advantage from social_facility tagging compared to amenity=nursing_home.

matthijsmelissen:
I like working with small iterations and keeping the pace up, so I have merged this PR as is.

I would still be very happy with additional icons. Some further discussion on driving_school color and on wether or not to render nursing_home might also be beneficial; PRs on these topics are welcomed.

polarbearing:
Sorry I was away over the weekend. This was merged very fast, and it is tricky to comment on such different tags, they should have been in separate PRs. I don't understand why we rush into merging now within 5 days before proper discussion. Even @matthijsmelissen you say above
that some questions are unresolved. There is not even an example for nursing home. 

**Driving schools**: Agree with rendering, shop colour is fine for me (many I know are actually behind a shop window). We should have made an icon for this, maybe based on a steering wheel or so. Why does the code need a separate entry and is not grouped into shops? 

**Childcare**: The tag has a steady growth since 2014, so there is a clear demand to tag facilities that are not considered kindergarten in the respective national contexts. Again for the code, why it is not grouped with kindergarten? Joining it with nursing_home appears impractical to me. 

**Nursing home**: This tag is controversial and frequently discussed on the tagging list, and its use is on decline this year (not yet visible in taghistory charts, but 14264 in Feb 2018 and 14144 today). In contrast, social_facility=nursing_home is on constant growth since its introduction on Jan 2017 (fixing the group_home confusion), having 3829 today. There is a friendly disagreement between @dieterdreist and myself, and the ambox in the wiki changes its colour depending on the last editor :-) 
The colour is inconsistent, social facilities are currently coloured public-service equalling amenity_brown.

**"prevent unfavorable fragmentation of tag use"**: @kocio-pl - there is no 'favorable fragmentation of tag use', it is just phrased misunderstandingly. It should read ''to prevent the fragmentation of tag use, which is unfavourable''.

andrzej-r:
polarbearing, thanks for the input. To me it looked like people had different preferences but none of them were necessarily right or wrong.

**Childcare not merged with kindergarten:** kindergarten is currently grouped with school, college, university etc. and is rendered at fairly low level in a more prominent style (IMHO that's too prominent for a kindergarten but changing that was outside of scope of this PR). Doing the same for childcare, which doesn't imply an institution, would be an overkill. I did match the label rendering of childcare to that of kindergarten, though.

**Nursing home:** I simply saw a valid (not marked as deprecated) and popular tag in taginfo that hasn't been supported by Carto yet. My opinion is that Carto should make the best use of available data and let tagging issues be resolved elsewhere. I don't mind nursing_home being rendered as public-service.

Issue #2086, opened 2016-03-10
traffic_calming=island


HolgerJeromin:
I want to discuss rendering of [traffic calming islands](http://wiki.openstreetmap.org/wiki/Tag:traffic_calming%3Disland).
Islands at a [junction](http://wiki.openstreetmap.org/wiki/File:Junction_with_central_island.jpg) or [end of a road](https://www.mapillary.com/map/im/HYgmG8VyNcAk5r4dYWXc6Q) could render like a miniroundabout, but in a [normal way](http://wiki.openstreetmap.org/wiki/File:Trafficcalming-island.jpg) this would be missleading.

matkoniecz:
I am not entirely sure is it a good candidate for rendering.

HolgerJeromin:
My concern is primary about the really "large" mini_roundabout like thing. It is nearly the same as a mini_roundabout or turning_circle, but without a sign in the middle. 

I am with you, that the island in a normal way should not be renderd.

HolgerJeromin:
Should we have another tag to be able to differ these two?

matkoniecz:
> mini_roundabout or turning_circle, but without a sign in the middle. 

Note that mini_roundabout and turning_circle does not require "a sign in the middle".

HolgerJeromin:
http://wiki.openstreetmap.org/wiki/Tag:highway%3Dmini_roundabout has a table of images. One entry is:
"Don't let the island confuse you: there are indications neither of a one-way nor of right-of-way, so this is just a junction with an island used for traffic calming."
exactly the situation of my examples.

kocio-pl:
I've just tried to add the code for it, but roads are very complex issue in this style, so I may try again later. Could anybody help me with that?

math1985:
Exactly why I proposed keeping v3 compatible with other osm2pgsql stylesheets - unfortunately @pnorman disagreed. We could still consider reintroducing compatability at v5.

kocio-pl:
@math1985: Looks to ma like a comment to another issue (https://github.com/gravitystorm/openstreetmap-carto/issues/2666#issuecomment-316991887 probably)?

BTW: Adding `tags->'traffic_calming' as traffic_calming,` to `highway-area-fill` layer works only when there's a `highway=*` tag, but not when there's only `traffic_calming + area=yes`. SQL expert is needed... :smile:

SomeoneElseOSM:
[...]

PR #3213, opened 2018-05-04
Add area rendering of taxi

### LINKED ISSUE #3110: Add area rendering for amenity=taxi
Similar as amenity=parking, taxi stands are places where cars are parked, so I think they could be rendered with the same area fill. There are two distuinguished icons for these objects, so they won't be confused with each other.

Example places:
https://www.openstreetmap.org/way/359796932
https://www.openstreetmap.org/way/421568284

As you can see, the borders of taxi stand areas are unvisible.

jragusa:
Fixes #3110

Add rendering of `amenity=taxi` in a similar way than other parking amenities

https://www.openstreetmap.org/way/284714637

Before
taxi_before

After
taxi_after

I noticed that `amenity=parking` is also included into `amenity-low-priority` in the project file. Should I put also taxi in this section ?

### DIFF
--- a/landcover.mss
+++ b/landcover.mss
@@ -545,7 +545,8 @@
 
   [feature = 'amenity_parking'][zoom >= 10],
   [feature = 'amenity_bicycle_parking'][zoom >= 10],
-  [feature = 'amenity_motorcycle_parking'][zoom >= 10] {
+  [feature = 'amenity_motorcycle_parking'][zoom >= 10],
+  [feature = 'amenity_taxi'][zoom >= 10] {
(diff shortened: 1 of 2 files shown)


Tomasz-W:
@jragusa Thanks for this PR! Are you interested in similar case of traffic calming islands https://github.com/gravitystorm/openstreetmap-carto/issues/2086 ? It's another one missing street element on map.

kocio-pl:
Thanks for this contribution!

jragusa:
@Tomasz-W I will take a look on #2086

PR #3216, opened 2018-05-06
Pixel align the shop=deli icon

meased:
The current shop=deli icon looks pretty fuzzy to me. This PR pixel aligns the icon to the pixel grid.

Icon before:
deli_grid_before

Icon after:
deli_grid_after

The edges of the jar are snapped outward to the pixel grid and the curve on the label is removed (the curve is not really visible at 14x14 anyway).

Before:
https://www.openstreetmap.org/#map=18/45.51760/-122.97312&layers=N
deli_before

After:
deli_after

### DIFF
--- a/symbols/shop/deli.svg
+++ b/symbols/shop/deli.svg
@@ -1,46 +1,20 @@
-<?xml version="1.0" encoding="UTF-8" standalone="no"?>
-<svg
-   xmlns:dc="http://purl.org/dc/elements/1.1/"
-   xmlns:cc="http://creativecommons.org/ns#"
-   xmlns:rdf="http://www.w3.org/1999/02/22-rdf-syntax-ns#"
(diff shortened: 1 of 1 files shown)


kocio-pl:
I don't see much difference even when trying to spot them, but if we can follow the guidelines closer, that's just better.

Would you be interested in similar pixel cleaning in #3158?

meased:
> Would you be interested in similar pixel cleaning in #3158?

Some of those don't look too hard to clean up, but tower_cooling and chimney are going to be tough. 14x14 is not that much space...

kocio-pl:
I would be happy if you could fix what you can, I will try to do the rest.

kocio-pl:
Thanks for fixing this icon.

sommerluk:
Thanks!

Issue #1291, opened 2015-02-04
Priority of keys in landcover layer


imagico:
Currently the landcover layer prioritizes `landuse` over `natural` see [here](https://github.com/gravitystorm/openstreetmap-carto/blob/master/project.yaml#L141) and i think this should be the other way round, especially with #1242 taking care of the case of `landuse=forest`+`natural=wood` (which accounts for >15k of a total of 67k cases with both keys together).

Cases where prioritizing `natural` over `landuse` would be better are:

* `landuse=meadow`/`landuse=farmland` with `natural=wood` for woodland used as pasture
* `landuse=conservation` in combination with any `natural` tag (although `landuse=conservation` should not be rendered in solid color at all of course)
* `landuse=forest` with `natural=scrub`

The only thing that comes to mind that needs special care is `natural=wetland` - since this is rendered separately it would need to be filtered out properly.

math1985:
In my opinion, landuse=meadow with natural=wood is a mistagging. Same for landuse=forest with natural=scrub. The problem with natural=conservation is a problem of this particular tag, not of the order. I'm therefore inclined to close this issue.

imagico:
How would you tag (a) an area of woodland used as pasture (b) a christmas tree plantation.  Note [the wiki](http://wiki.openstreetmap.org/wiki/Tag:landuse%3Dforest) explicitly mentions the possibility of landuse=forest + natural=scrub.

The question should IMO more be if there would be any problems if priorities were reversed.

dieterdreist:
2015-03-04 10:06 GMT+01:00 imagico <notifications@github.com>:

> How would you tag (a) an area of woodland used as pasture
>

there probably aren't tags for this. We don't have any tag for woodland
(i.e. low density trees), just for forests. And we don't have dedicated
tags for pasture (it is included in landuse=farmland and in landuse=meadow
according to this wikipage:
http://wiki.openstreetmap.org/wiki/Tag%3Alanduse%3Dfarmland ).
Well, there are also some uses of the pasture key:
http://taginfo.osm.org/keys/pasture but there is no definition in the wiki
AFAIK.
And there are also some subkeys with pasture as value, like farmland,
meadow, landuse http://taginfo.osm.org/search?q=%3Dpasture


> (b) a christmas tree plantation. Note the wiki
> <http://wiki.openstreetmap.org/wiki/Tag:landuse%3Dforest> explicitly
> mentions the possibility of landuse=forest + natural=scrub.
[...]

PR #3224, opened 2018-05-08
Order landcover by osm_id

### LINKED ISSUE #2782: Should landcover be ordered by osm_id in addition to way_area?
Recently i saw a landcover area where the color varied arbitrarily depending on zoom level because there are two different areas with different tags but identical geometry and therefore identical way_area:

https://www.openstreetmap.org/way/346109314
https://www.openstreetmap.org/way/49188959

I think cases like this should get a consistent drawing order, probably best by ordering by osm_id in addition.

matthijsmelissen:
This resolves #2782

### DIFF
--- a/project.mml
+++ b/project.mml
@@ -99,7 +99,7 @@ Layer:
               OR "natural" IN ('wood', 'wetland', 'mud', 'sand', 'scree', 'shingle', 'bare_rock', 'heath', 'grassland'))
               AND way_area > 0.01*!pixel_width!::real*!pixel_height!::real
               AND building IS NULL
-            ORDER BY way_area DESC
+            ORDER BY way_area DESC, osm_id
           ) AS features
         ) AS landcover_low_zoom
     properties:
@@ -153,7 +153,7 @@ Layer:
               OR highway IN ('services', 'rest_area')
               OR railway = 'station')
               AND way_area > 0.01*!pixel_width!::real*!pixel_height!::real
-            ORDER BY way_area DESC
+            ORDER BY way_area DESC, osm_id
           ) AS landcover
         ) AS features
     properties:


andrzej-r:
Do you thing it would be possible to sort by feature instead. I understand sorting by osm_id would produce different rendering when someone deletes/recreates an object, which seems like a hack to me.

matthijsmelissen:
> Do you thing it would be possible to sort by feature instead.

Hmm, good point. I've been thinking about it, but I think it would add quite a lot of code for a relatively small gain. So I think I'd prefer to leave it as it.

sommerluk:
>  it would add quite a lot of code

Wouldn’t it be enough to leave the inner query as it is currently (without this PR), and add to the outer query ` ORDER BY feature, osm_id`? This should order by feature alphabetically (which is just as arbitrary as osm_id, but independent from deleting/re-creating objects in the OSM database) before ordering by osm_id…

matthijsmelissen:
My bad, this would work of course. Will fix it.

sommerluk:
Hm, my bad, should be `ORDER BY way_area DESC, feature, osm_id` on the outer query (and no order in the inner query), otherwise way_area wouldn’t be the primary order criterium any more.

ImreSamu:
>I don't see that we need osm_id ordering if we've got feature in the ordering

imho - the original varied color problem is solved,  but  it can be a similar problem in the varied landcover labels.  ( in rare cases )

pnorman:
> imho - the original varied color problem is solved, but it can be a similar problem in the varied landcover labels. ( in rare cases )

The landcover layer is used for fills, not labels.

pnorman:
It needs changes to stop ordering after `feature`.

Tomasz-W:
It resolves https://github.com/gravitystorm/openstreetmap-carto/issues/1291 also.

kocio-pl:
@matthijsmelissen What would you like to do with code? Do you have any plans?

HolgerJeromin:
> It resolves #1291 also.

Why that?

matthijsmelissen:
Updated.

kocio-pl:
I'm not sure if you've seen this, but there's 1 conflict left (related to #2874).

matthijsmelissen:
Superseded by #3441.

The current PR is a PR from a branch on gravitystorm directly, it should of course come from matthijsmelissen:landcover, which is fixed in the new PR.

PR #3225, opened 2018-05-08
Brighten built-up areas on z12

### LINKED ISSUE #2777: landuse=residential color change
Hello.

With new style, residential color variation between zoom levels looks arbitrary.
It becomes darker and brighter several times as you zoom in.

z8, z9 - 221 (bright)
z10 - 212 (dark)
z11 - 216 (bright)
z12 - 211 (dark)
z13 - 223 (bright)

And at zoom=12 it looks just bad in my opinion:
(residential areas receives too much emphasis)
z12_bad

Before-after comparison:
z12_change

matthijsmelissen:
This resolves #2777

The brightness sequence from z10 up used to be 212-216-211-223, now it is 212-216-221-223.

Before
screen shot 2018-05-09 at 00 31 19

After
screen shot 2018-05-09 at 00 31 11

### DIFF
--- a/landcover.mss
+++ b/landcover.mss
@@ -9,8 +9,9 @@
 
 // --- "Base" landuses ---
 
-@built-up-upper-lowzoom: #c0c0c0;
-@built-up-lower-lowzoom: #aaaaaa;
+@built-up-lowzoom: #aaaaaa;
+@built-up-z11: #c0c0c0;
+@built-up-z12: #d0d0d0;
 @residential: #e0dfdf;      // Lch(89,0,0)
 @residential-line: #b9b9b9; // Lch(75,0,0)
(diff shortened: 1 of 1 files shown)


kocio-pl:
Thanks! I have to test brightness progression on the map, but this change looks like a proper tuning of midzoom rendering.

kocio-pl:
It looks good for me. Colors are well balanced and details are more visible. The downside is that we loose a feeling that built-up areas are coherent for big cities, but we gain more intuitive, gradual color change.

If nobody objects, I plan to merge it tomorrow.

Warsaw, z12

Before
7gxsqjto

After
655zelor

SW of Warsaw, z12

Before
afswf21i

After
nydnm3oe

PR #3226, opened 2018-05-09
Select overground power=cable like power=line

### LINKED ISSUE #3217: Render overground power cables
### Expected behavior
Gas insulated cables are now becoming common in high voltage substations around me. They have to be mapped with `power=cable` instead of `power=line`, since the insulation is the discerning factor between these tags.

I'd propose to render **cables the same as power lines** if they have a **location** tag that is **different from `underground` and `underwater`**.

### Actual behavior
Power cables are never rendered, which makes substations that use gas insulated cables have gaps on the map.

screenshot_2018-05-06_22-56-42

M1dgard:
Power cables with a location tag that says they are overground, are now selected and rendered just like power lines, to improve the look of substations that use gas insulated cables. Since those cables are insulated, they have to be mapped with power=cable instead of power=line, but this left undesirable gaps in the connections.

[According to the wiki](https://wiki.openstreetmap.org/wiki/Tag:power%3Dcable) cables without a location tag default to 'either underground or underwater', so they are treated as such.

Closes #3217.

#### Before
Location: [a substation in Zeebrugge, Belgium](https://www.openstreetmap.org/#map=18/51.32588/3.18409)
before
before-zoom-out

#### After
after
after-zoom-out

~~I tried to set up Kosmtik but the installation failed horribly. I don't feel like installing even more versions of Node.js (which I don't like) on my system to see which one might possibly work, to test such a small change. If you can give me some pointers about which versions to use, I would give it another go.~~

~~TileMill's install script thinks that all Linux is Debian so TileMill didn't work either.~~

### DIFF
--- a/project.mml
+++ b/project.mml
@@ -1234,7 +1234,7 @@ Layer:
             way
           FROM planet_osm_line
           WHERE power = 'line'
-            OR (power = 'cable' AND tags->'location' IS NOT NULL AND tags->'location' NOT IN ('underground', 'underwater'))
+            OR (power = 'cable' AND tags->'location' IN ('overground', 'overhead', 'surface', 'outdoor', 'platform'))
         ) AS power_line
     properties:
       minzoom: 14
--- a/project.mml
+++ b/project.mml
@@ -1234,6 +1234,7 @@ Layer:
             way
           FROM planet_osm_line
           WHERE power = 'line'
+            OR (power = 'cable' AND tags->'location' IS NOT NULL AND tags->'location' NOT IN ('underground', 'underwater'))
         ) AS power_line
     properties:
       minzoom: 14


kocio-pl:
Did you try using Docker environment?:

M1dgard:
> Did you try using Docker environment?:

Yes, and installing the packages in the Ubuntu environment fails because it cannot resolve security.ubuntu.com and archive.ubuntu.com. Now it comes to my mind that this is probably a firewall issue that I might be able to resolve.

M1dgard:
I fixed a mistake in the query and added the before/after images. As expected, it's now rendered as if the cable would have been tagged with `power=line`. I also looked at the location of a `location=underground` cable and one without a `location` tag, and they are indeed not rendered.

Ready to be merged as far is I am concerned.

kocio-pl:
Great! We have learned in the past that a code without some basic rendering test can be buggy and testing became an important part of changing this style.

I think we should not use a blacklist here, since there can be many types of location not visible on the ground and we should not take them for granted. Instead only documented and visible types should be selected, and in this case it means `overground` and `overhead`:

https://wiki.openstreetmap.org/wiki/Tag:power%3Dcable#Location

M1dgard:
Hm, I went by the [Key:location](https://wiki.openstreetmap.org/wiki/Key:location) page which defines a lot more values.

I personally expect that there will be created more new overground locations than underground locations, for the simple reason that how an underground object is built is difficult to find out.

But a whitelist is also fine by me, so `'overground', 'overhead', 'roof', 'rooftop', 'surface', 'outdoor', 'platform'`. Though not "visible", I would also suggest to include `indoor`. Buildings that have cables mapped indoors tend to be buildings in a substation anyway. In a substation I think it's nice to show all cables and lines. And indoor lines are already shown anyway :p

kocio-pl:
Thanks!

- I believe NULL checking doesn't make sense now. 
- I would not bet that someone won't include cables indoor in a university campus or something like that, where the documentation is available.
- I would not like to show cables on roofs, just like I try to avoid showing telecommunication towers etc. - we could end up with a mess.

M1dgard:
You're right. I've rewritten my last commit to omit NULL checking and drop indoor, roof and rooftop.

Still rendering as expected.

kocio-pl:
Thanks, this works as expected now.

PR #3229, opened 2018-05-12
Pixel align the ford icon

### LINKED ISSUE #2822: Ford symbol is blurry
From https://github.com/gravitystorm/openstreetmap-carto/pull/2743#issuecomment-328388361

image

I haven't investigated, but I bet it's not pixel aligned

meased:
Fixes #2822 

The arrow is a little fatter because arrows need to end on a point to look sharp and therefore need to have an odd height. One water line is removed to reduce the weight a little and to keep the design language the same as waterfall, public_bath, beach_resort, and water_park

Icon before:
ford_before
ford_before_grid

Icon after:
ford_after
ford_after_grid

https://www.openstreetmap.org/#map=16/45.3493/-121.6580
Before:
ford_before_sample

After:
ford_after


matthijsmelissen:
Looks good.

The antialiasing at the bottom might koop better with the water moved up ine pixel?

dieterdreist:
I agree with Kocio, the waves are not very recognizable now, maybe making them higher will improve readability

imagico:
For reference: The original discussion for the current ford icon can be found in #1625.

kocio-pl:
4 is the best for me.

geozeisig:
Why do you use a loupe? In my eyes, the icon looks amazing good in 1: 1 mode.

matthijsmelissen:
Thanks!

PR #3161, opened 2018-04-01
Add playground equipment rendering

pietervdvn:
Changes proposed in this pull request:
- Add icons for playground equipment, such as slides and swings

Test rendering with links to the example places:

See https://www.openstreetmap.org/#map=19/51.20569/3.23290

Before:
screenshot from 2018-04-01 13-52-44

After
screenshot from 2018-04-01 13-38-01

PR #3230, opened 2018-05-12
playground=sandpit with sand as background-color

tordans:
# Changes proposed in this pull request:

A sandpit is always made of sand and therefore can and should be displayed with the color of sand. However, in contrast to the general sand rendering, the playground should only be rendered in higher zoom levels since the size of the sandpit is most likely small and will only clutter the map when zoomed out.

- Details about at `playground=sandpit` at https://wiki.openstreetmap.org/wiki/Key:playground#playground-values
- Details about the tag usage https://taginfo.openstreetmap.org/tags/playground=sandpit

# Testcase 1 Karl-Marx-Platz
https://www.openstreetmap.org/#map=19/52.47447/13.44328

## Before
bildschirmfoto 2018-05-12 um 08 33 35

## After

(I dont know yet how to create test renderings)

# Testcase 2 Kirchgasse
https://www.openstreetmap.org/#map=19/52.47799/13.44190

## Before
bildschirmfoto 2018-05-12 um 08 33 20

## After

(I dont know yet how to create test renderings)

### DIFF
--- a/landcover.mss
+++ b/landcover.mss
@@ -482,6 +482,12 @@
     }
   }
 
+  [feature = 'sandpit'][zoom >= 17] {
+    polygon-fill: @sand;
+    [way_pixels >= 4]  { polygon-gamma: 0.75; }
+    [way_pixels >= 64] { polygon-gamma: 0.3;  }
+  }
+
   [feature = 'natural_sand'][zoom >= 8] {
     polygon-fill: @sand;
     [way_pixels >= 4]  { polygon-gamma: 0.75; }


pnorman:
With no changes to the MML I don't believe this PR will work when you test it.

polarbearing:
playground =* tags playground equipment. We decided in #3161 not to render them.
This PR would add the sandpit as a special case of equipment. 

The sandpit is already rendered in sand-yellow when natural=sand is added.

kocio-pl:
Yes, I also think that this is a special case worth adding. We want a semantic database and avoid adding `natural=sand` as less precise.

tordans:
I updated the PR Description with the "after" shots based on [my testing of `natural=sand`](https://www.openstreetmap.org/changeset/58934448).

## Only >=18

Based on the screesnhots I think: Lets only show the sandpit for >= 18. Otherwise very small sandpits just clutter the map.

## About the code

Can anyone who knows how this works take over? It will take me a long while to setup a dev-env. I might have to migrate it into an regular issue.

## Here are some more screenshots of different zoom levels

### Zoom 19 
zoom-19

### Zoom 18 
zoom-18 (there are 3 `sandpit` in a row)

### Zoom 17 
zoom-17b

### Zoom 16
zoom-16

### Zoom 15
zoom-15

kocio-pl:
> Based on the screesnhots I think: Lets only show the sandpit for >= 18. Otherwise very small sandpits just clutter the map.

Sounds good for me.

> Can anyone who knows how this works take over? It will take me a long while to setup a dev-env. I might have to migrate it into an regular issue.

There's no hurry and I prefer that more people can use our toolset and produce proper code. We have too many issues and ideas, and too few regular coders.

Do you have some problem with Docker containers?

polarbearing:
The sand in the pit is natural. There is no artificial sand, so natural=sand is not tagging for the renderer, it is the actual fact.

dieterdreist:
sent from a phone

> On 14. May 2018, at 08:54, polarbearing <notifications@github.com> wrote:
> 
> The sand in the pit is natural. There is no artificial sand, so natural=sand is not tagging for the renderer, it is the actual fact.


IMHO the key natural is for natural features, not for things made from natural material. A wooden house would not be natural=house. landcover=sand would not have this problem.

dieterdreist:
sent from a phone

> On 12. May 2018, at 08:37, Tobias <notifications@github.com> wrote:
> 
> However, in contrast to the general sand rendering, the playground should only be rendered in higher zoom levels since the size of the sandpit is most likely small and will only clutter the map when zoomed out.


why not making the zoom level depending on the area size in pixels? While many sandpits are small, a huge one will appear odd when rendered only on very high zooms

dieterdreist:
sent from a phone

> On 17. May 2018, at 22:24, Paul Norman <notifications@github.com> wrote:
> 
> I'm not convinced that rendering a playground feature the same as natural=sand is a good idea. Our natural=sand rendering is designed to fit in with other nature features in the style


on the other hand you (not you as Paul Norman, but the osm carto team) suggested not too long ago natural=water for the water covered parts of fountains.

natural=sand is used as landcover=sand, it might indicate a natural feature, but not even sure it is the majority, the common property is that there is sand on the surface. If we go by meaning of tags according to usage, there’s a lost battle wrt natural, IMHO. (even worse natural=mud)

PR #3231, opened 2018-05-12
Make amenity=shelter brown

kocio-pl:
Observed in the German style.

- Make amenity=shelter brown, as it is not a transport nor accommodation utility, rather generic amenity

It looks more like other amenities [on platforms](https://www.openstreetmap.org/#map=19/52.23285/20.99709) and doesn't take away attention from real transport-related objects:

Before
8wn9n_af

After
tswelai8

It blends better with other amenities [at the picnic site](https://www.openstreetmap.org/#map=18/52.30240/20.87196) and with buildings:

Before
jojkzgyo

After
uazgkp m

It's also easier to spot [in the forest](https://www.openstreetmap.org/node/2881682202#map=16/52.3026/20.3666):

Before
ze86_f8n

After
5 6r5fof

### DIFF
--- a/amenity-points.mss
+++ b/amenity-points.mss
@@ -47,6 +47,9 @@
     [feature = 'tourism_alpine_hut'] {
       marker-file: url('symbols/alpinehut.svg');
     }
+    [feature = 'amenity_shelter'] {
+      marker-fill: @amenity-brown;
+    }
     marker-fill: @accommodation-icon;
     marker-placement: interior;
     marker-clip: false;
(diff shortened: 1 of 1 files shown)


kocio-pl:
I would not use man made black, since they are useful for general public, unlike the objects you've mentioned, which are mostly landmarks. Green suggests some kind of leisure and has similar brightness as blue. Similarity to small trash cans on platform and picnic tables is what makes me think that brown is the best.

Issue #2938, opened 2017-11-08
Unified rendering of healthcare objects


kvlahromei:
Hi,

currently amenity=doctors, amenity=hospital, shop=optican, ... are covered in the current style.  
I recommend to add further [healthcare=* objects](https://wiki.openstreetmap.org/wiki/Key:healthcare) as there are esp. in big cities more complex clinic / campus areas that would benefit from a more detailed (but also balanced) rendering of that institutions.

For example [here in Rostock](http://www.openstreetmap.org/#map=17/54.08443/12.10241) we are currently hiding this details
![osm rostock schillingallee](https://user-images.githubusercontent.com/8177370/32557057-5efb4318-c4a1-11e7-9dfc-c8a8c3488a16.png).

It would be great, if we can present this POIs in the same pink tone (but without icons) as the doctors, veterinarian, ...

(This was already suggested in #1981, #763 some years ago. But I see a growing popularity / more precise tagging since that time)

HolgerJeromin:
Looking at the list i could imaging some (birthing_center, midwife, blood_donation, physiotherapist, psychotherapist, rehabilitation) can have a healthcareColor dot. But not everything (for example blood_bank is no place to go).

This issue could get an meta issue like #108

kocio-pl:
This one could be meta, you just need to make a list. :smile:

Matthias84:
Ok, so listing what needs to be done:

**new**, with pink cross/dot and maybe with name=* or operator=* label for healthcare=...
- [ ] yes
- [ ] birthing_center
- [ ] blood_bank
- [ ] blood_donation
- [ ] centre
- [ ] hospice
- [ ] rehabilitation

**conflicting ones**, they are covered by some other tags
- [ ] clinic  - amenity=clinic
- [ ] hospital - amenity=hospital
- [ ] dentist - amenity=dentist
(following ones by a more general amenity=doctors)
- [ ] audiologist
- [ ] alternative
- [ ] midwife
- [ ] occupational_therapist
- [ ] optometrist
- [ ] physiotherapist
- [ ] podiatrist
- [ ] psychotherapist
- [ ] speech_therapist

(just give me a ping, if you prefer an sep. issue ticket instead or if I embed it to the first post ...)

IMHO it's worth to tune the styling for zoomlevels. For example hospitals are more important than centres or doctors. 

[...]

PR #3232, opened 2018-05-13
[WIP] Adding rendering for public_transport=platform

### LINKED ISSUE #435: Add rendering for public_transport=platform
**Edit: Okay maybe rather use it on public_transport=platform**

Use the same icon as highway=bus_stop

(if you want to do more use different ones depending on bus/tram=yes/no)

It would be great if you could support the "new" public transport schema, I think one of the main reason people don't use it, is because it does not show up on the map.

http://wiki.openstreetmap.org/wiki/Key:public_transport
http://wiki.openstreetmap.org/wiki/Proposed_features/Public_Transport

kocio-pl:
Fixes #435

- Adding rendering for public_transport=platform

Initial code renders `public_transport=platform+bus=yes` the same as `highway_bus_stop`. It should be decided how to render platforms for multiple types of transport - for example bus+tram.

### DIFF
--- a/project.mml
+++ b/project.mml
@@ -1612,6 +1612,7 @@ Layer:
                              ELSE NULL END,
               'military_'|| CASE WHEN military IN ('bunker') THEN military ELSE NULL END,
               'highway_'|| CASE WHEN highway IN ('bus_stop', 'elevator', 'traffic_signals') THEN highway ELSE NULL END,
+              'public_transport_'|| CASE WHEN tags->'public_transport' IN ('platform') THEN tags->'public_transport' ELSE NULL END,
               'power_' || CASE WHEN power IN ('generator') THEN power ELSE NULL END,
               'tourism_' || CASE WHEN tourism IN ('viewpoint') THEN tourism ELSE NULL END
             ) AS feature,
@@ -1619,6 +1620,7 @@ Layer:
             religion,
             tags->'denomination' as denomination,
             tags->'generator:source' as "generator:source",
+            tags->'bus' as bus,
             CASE
               WHEN man_made IN ('mast', 'tower') THEN
                 CASE
@@ -1635,6 +1637,7 @@ Layer:
             tags->'tower:construction' as "tower:construction",
             tags->'tower:type' as "tower:type",
             tags->'castle_type' as castle_type,
+            tags->'public_transport' as public_transport,
             CASE WHEN shop IN ('supermarket', 'bag', 'bakery', 'beauty', 'bed', 'books', 'butcher', 'clothes', 'computer',
                                'confectionery', 'fashion', 'convenience', 'department_store', 'doityourself', 'hardware', 'fishmonger', 'florist',
                                'garden_centre', 'hairdresser', 'hifi', 'ice_cream', 'car', 'car_repair', 'bicycle', 'mall', 'pet',
@@ -1663,6 +1666,8 @@ Layer:
             OR tags->'memorial' IN ('plaque')
             OR military IN ('bunker')
             OR highway IN ('bus_stop', 'elevator', 'traffic_signals')
+            OR tags->'public_transport' IN ('platform')
+            OR tags->'bus' IN ('yes')
             OR (power = 'generator' AND (tags @> '"generator:source"=>wind' OR tags @> 'power_source=>wind'))
           ORDER BY way_area desc
         ) AS amenity_points_poly
@@ -1725,6 +1730,7 @@ Layer:
               'military_'|| CASE WHEN military IN ('bunker') THEN military ELSE NULL END,
               'highway_'|| CASE WHEN highway IN ('bus_stop', 'elevator', 'traffic_signals') THEN highway
                                 WHEN tags @> 'ford=>yes' OR tags @> 'ford=>stepping_stones' THEN 'ford' ELSE NULL END,
+              'public_transport_'|| CASE WHEN tags->'public_transport' IN ('platform') THEN tags->'public_transport' ELSE NULL END,
               'power_' || CASE WHEN power IN ('generator') THEN power ELSE NULL END,
               'tourism_' || CASE WHEN tourism IN ('viewpoint') THEN tourism ELSE NULL END,
               'barrier_' || CASE WHEN barrier IN ('toll_booth') THEN barrier ELSE NULL END,
@@ -1764,6 +1770,8 @@ Layer:
             tags->'tower:construction' as "tower:construction",
             tags->'tower:type' as "tower:type",
             tags->'castle_type' as castle_type,
+            tags->'public_transport' as public_transport,
+            tags->'bus' as bus,
             CASE WHEN shop IN ('supermarket', 'bag', 'bakery', 'beauty', 'bed', 'books', 'butcher', 'clothes', 'computer',
                                'confectionery', 'fashion', 'convenience', 'department_store', 'doityourself', 'hardware', 'fishmonger', 'florist',
                                'garden_centre', 'hairdresser', 'hifi', 'ice_cream', 'car', 'car_repair', 'bicycle', 'mall', 'pet',
@@ -1795,6 +1803,8 @@ Layer:
             OR military IN ('bunker')
             OR tags @> 'emergency=>phone'
             OR highway IN ('bus_stop', 'elevator', 'traffic_signals')
+            OR tags->'public_transport' IN ('platform')
+            OR tags->'bus' IN ('yes')
             OR tags @> 'ford=>yes' OR tags @> 'ford=>stepping_stones'
             OR (power = 'generator' AND (tags @> '"generator:source"=>wind' OR tags @> 'power_source=>wind'))
           ORDER BY score DESC NULLS LAST
@@ -2193,6 +2203,7 @@ Layer:
                              THEN concat_ws('_', historic, CASE WHEN tags->'memorial' IN ('plaque') THEN tags->'memorial' ELSE NULL END)
                              ELSE NULL END,
               'highway_' || CASE WHEN highway IN ('services', 'rest_area', 'bus_stop', 'elevator') THEN highway ELSE NULL END,
+              'public_transport_'|| CASE WHEN tags->'public_transport' IN ('platform') THEN tags->'public_transport' ELSE NULL END,
               'boundary_' || CASE WHEN boundary IN ('national_park') THEN boundary ELSE NULL END,
               'waterway_' || CASE WHEN waterway IN ('dam', 'dock') THEN waterway ELSE NULL END,
               'tourism_' || CASE WHEN tourism IN ('viewpoint', 'attraction') THEN tourism ELSE NULL END
@@ -2221,6 +2232,8 @@ Layer:
             tags->'office' as office,
             tags->'recycling_type' as recycling_type,
             tags->'castle_type' as castle_type,
+            tags->'public_transport' as public_transport,
+            tags->'bus' as bus,
             ref,
             way_area,
             CASE WHEN building = 'no' OR building IS NULL THEN 'no' ELSE 'yes' END AS is_building
@@ -2242,6 +2255,8 @@ Layer:
               OR historic IN ('memorial', 'monument', 'archaeological_site', 'fort', 'castle', 'manor')
               OR tags->'memorial' IN ('plaque')
               OR highway IN ('services', 'rest_area', 'bus_stop', 'elevator')
+              OR tags->'public_transport' IN ('platform')
+              OR tags->'bus' IN ('yes')
               OR power IN ('plant', 'station', 'generator', 'sub_station', 'substation')
               OR boundary IN ('national_park')
               OR waterway IN ('dam', 'dock'))
@@ -2308,6 +2323,7 @@ Layer:
             office,
             recycling_type,
             castle_type,
+            bus,
             ref,
             way_area,
             is_building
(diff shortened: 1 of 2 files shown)


project.mml
```diff
@@ -1612,13 +1612,15 @@ Layer:
...
               'highway_'|| CASE WHEN highway IN ('bus_stop', 'elevator', 'traffic_signals') THEN highway ELSE NULL END,
+              'public_transport_'|| CASE WHEN tags->'public_transport' IN ('platform') THEN tags->'public_transport' ELSE NULL END,
               'power_' || CASE WHEN power IN ('generator') THEN power ELSE NULL END,
               'tourism_' || CASE WHEN tourism IN ('viewpoint') THEN tourism ELSE NULL END
             ) AS feature,
             access,
             religion,
             tags->'denomination' as denomination,
             tags->'generator:source' as "generator:source",
+            tags->'bus' as bus,
```
pnorman: Does this blow up the XML with a combinatorial explosion?

project.mml
```diff
@@ -1663,6 +1666,8 @@ Layer:
             OR tags->'memorial' IN ('plaque')
             OR military IN ('bunker')
             OR highway IN ('bus_stop', 'elevator', 'traffic_signals')
+            OR tags->'public_transport' IN ('platform')
+            OR tags->'bus' IN ('yes')
```
pnorman: Not needed, we don't render bus=yes without public_transport=platform

project.mml
```diff
@@ -1795,6 +1803,8 @@ Layer:
             OR military IN ('bunker')
             OR tags @> 'emergency=>phone'
             OR highway IN ('bus_stop', 'elevator', 'traffic_signals')
+            OR tags->'public_transport' IN ('platform')
+            OR tags->'bus' IN ('yes')
```
pnorman: as above

project.mml
```diff
@@ -1764,6 +1770,8 @@ Layer:
             tags->'tower:construction' as "tower:construction",
             tags->'tower:type' as "tower:type",
             tags->'castle_type' as castle_type,
+            tags->'public_transport' as public_transport,
+            tags->'bus' as bus,
```
pnorman: as above

project.mml
```diff
@@ -2416,6 +2435,8 @@ Layer:
                   OR historic IN ('memorial', 'monument', 'archaeological_site', 'wayside_cross', 'fort', 'wayside_shrine', 'castle', 'manor')
                   OR tags->'memorial' IN ('plaque')
                   OR highway IN ('bus_stop', 'services', 'rest_area', 'elevator')
+                  OR tags->'public_transport' IN ('platform')
+                  OR tags->'bus' IN ('yes')
```
pnorman: as above

kocio-pl:
Public transport is mature, more flexible and more coherent standard, so it's a good example to talk about handling such cases. Some are less obvious, like `amenity=social_facility`, which covers some other amenity schemes, some are similar, like `healthcare=*` (see #2938).

kocio-pl:
General chart to show the comparison of the old and new scheme - it's not accurate, since only [~500k of the new one is combined with `bus=yes`](https://taginfo.openstreetmap.org/tags/public_transport=platform#combinations), yet this is good illustration of the problem:

taghistory 20

So the old one is clear winner, however both numbers are very high and the ratio is just 4:1 (the same order of magnitude).

The question is when should we change rendering? When ratio will be 1:1? How should we avoid fragmentation - when the new scheme will have `n+1` uses should we drop `n` uses of the old one? If not - how long should we keep both? And when we could start showing the new one if both could be rendered?

dieterdreist:
sent from a phone

> On 13. May 2018, at 05:40, kocio-pl <notifications@github.com> wrote:
> 
> The question is when should we change rendering? When ratio will be 1:1? How should we avoid fragmentation - when the new scheme will have n+1 uses should we drop n uses of the old one? If not - how long should we keep both? And when we could start showing the new one if both could be rendered?


I suggest you drop rendering of the old scheme when usage drops below 2000 instances.

What would be interesting in order to decide for supporting the “new” scheme would be the numbers of objects that are only tagged according to the new scheme.

sommerluk:
By the way: Is it intentional that we render currently `highway=bus_stop` on areas? The wiki describes it as only for nodes (at difference from `public_transport=platform`, which is _not only_ for nodes). Following taginfo, for `highway=bus_stop` we have 2 046 215 nodes and only 6 053 ways.

dieterdreist:
sent from a phone

> On 13. May 2018, at 09:54, Lukas Sommer <notifications@github.com> wrote:
> 
> The wiki describes it as only for nodes


IMHO the wiki might be wrong.

sommerluk:
> IMHO the wiki might be wrong.

Given that also the wiki describes describes it also as _“Its position may be marked by a shelter, pole, bus lay-by, or road markings.”_ at least for bus lay-by and shelter, using an area sounds reasonable.

geozeisig:
I think that also fits on this topic: why do we render `highway=platform`? The [Wiki](https://wiki.openstreetmap.org/wiki/Tag:highway%3Dplatform) has no status and at the top you can read: This feature has been replaced by public_transport=platform. The behavior causes this tag to be used as a replacement.

project.mml
kocio-pl: What would you propose to do instead?

pnorman: Remove the line. You could further restrict the public_transport condition to only get those with bus=yes, but that makes no difference to the rendering

kocio-pl:
Looks like the revolution is long over and `public_transport=platform` won both on wiki and by numbers, so I think we could remove `highway=platform` soon (after announcing and discussing on Tagging list, of course):

taghistory 21

PolyglotOpenstreetmap:
@dieterdreist expecting people to map according to the "new" scheme exclusively without their stops getting renderd  is ludicrous. I've tried that many many years ago, then resorted to double tagging the whole lot. In the present, I don't really care anymore. The point has become moot. We tag highway=bus_stop to get them rendered and p_t=platform/stop_position to please those who like the new scheme.

Marc-marc-marc:
for info the supremacy of public_transport=platform over highway=platform is even clearer if we take into account the duplicate tag
900k for public_transport=platform without (highway=platform or railway=platform)
84k highway=platform + public_transport=platform
only 9k highway=platform without public_transport=platform

BjornRasmussen:
I see no reason why the new public_transport=platform should not be rendered.  Not rendering it is causing a large amount of bus stops to have both public_transport=platform and highway=bus_stop, which should not be necessary.

BjornRasmussen:
The point of not rendering public_transport=platform was to stop people from using the new system.  This is not working, however, and the result is just redundant tags on all bus stops.  This should be merged.

AgusQui:
Many we use highway=platform to represent the physical platform, and highway=bus_stop for the point of that platform where the bus stops, this is an example of platform with several stops inside https://www.openstreetmap.org/way/114037316#map=19/-34.60004/-58.38183, how do you propose to differentiate this if highway=bus_stop and highway=platform becomes the same as public_transport=platform?

SilentSpike:
> Many we use highway=platform to represent the physical platform, and highway=bus_stop for the point of that platform where the bus stops, this is an example of platform with several stops inside https://www.openstreetmap.org/way/114037316#map=19/-34.60004/-58.38183, how do you propose to differentiate this if highway=bus_stop and highway=platform becomes the same as public_transport=platform?

As far as I know PTv2 has appropriate tags for both of these cases:
`public_transport=platform`
`public_transport=stop_position`

AgusQui:
stop_position indicates the place on the street where the buses stop, in
this case highway=bus_stop indicates where is the signaling that indicates
where passengers should wait for the different buses that stop on the same
platform, in addition there are many who prefer not to use stop_position
for having no use (the software can detect where a bus can stop on the
platform)

El oct. 13, 2018 7:50, "SilentSpike" <notifications@github.com> escribió:

Many we use highway=platform to represent the physical platform, and
highway=bus_stop for the point of that platform where the bus stops, this
is an example of platform with several stops inside
https://www.openstreetmap.org/way/114037316#map=19/-34.60004/-58.38183, how
do you propose to differentiate this if highway=bus_stop and
highway=platform becomes the same as public_transport=platform?

As far as I know PTv2 has appropriate tags for both of these cases:
public_transport=platform
public_transport=stop_position

—
You are receiving this because you commented.
Reply to this email directly, view it on GitHub
<https://github.com/gravitystorm/openstreetmap-carto/pull/3232#issuecomment-429532284>,
or mute the thread

AgusQui:
Edit: Sorry, I see what you're saying in regards to differentiating the rendering, forgot I was reading the carto issues.

Of course, how do you think representing physical platforms? Now they have a gray color like the train platforms, while highway=bus_stop has the symbol of a blue bus, are now both a blue bus?

jeisenbe:
@kocio-pl, are you going to work on this PR further? 

It seems that the mappers have decided to use public_transport=platform, as discussed above: https://github.com/gravitystorm/openstreetmap-carto/pull/3232#issuecomment-398473340

kocio-pl:
I am still not able to do more tasks than currently, be it testing/merging PRs or just tuning my own code, and I don't know when I can get back to it. 

If you - or anybody else - would like to take over where I have left it, I'm fine with that.

matkoniecz:
> I'm not a particular fan of public_transport=platform and don't want to do something that would encourage people to use it over the established tagging.

I second this.

Can anyone supply link with documentation  why `public_transport=platform bus=yes` tag duplication of `highway=bus_stop` is a good idea?

Neither https://wiki.openstreetmap.org/wiki/Tag:public_transport%3Dplatform nor https://wiki.openstreetmap.org/w/index.php?oldid=625726 is explaining this.

kocio-pl:
Currently I think rather about avoiding such duplicates and rendering this tag only when **not** associated with already rendered types of platforms. Transport platform is very wide category and it's worth rendering.

pnorman:
Given there's been open technical issues with this PR for a year, @kocio-pl doesn't foresee having time, and the relevant part of code has been significantly refactored, I'm going to close this PR.

If someone wants to take this up and write a new PR, they can do so, even though I am personally opposed to rendering it, as stated above.

kocio-pl:
> Given there's been open technical issues with this PR for a year, @kocio-pl doesn't foresee having time, and the relevant part of code has been significantly refactored, I'm going to close this PR.

OK, sounds sane for me.

> If someone wants to take this up and write a new PR, they can do so, even though I am personally opposed to rendering it, as stated above.

Do you oppose to rendering it only when this would be duplication of already rendered tag or any `public_transport=platform` at all?

IIVQ:
IMHO, we have a deadlock here.
Much of the opposition to public_transport=platform (as a way or area, but also as a node) comes from it not being rendered on OSM-carto, thus necessitating a highway=bus_stop or railway=tram_stop node to be rendered.

Now if we don't include rendering to public_transport=platform (as node+way+area) because it is double-tagging with highway=bus_stop, we're having double tagging.

public_transport=platform is part of the widely used PT_v2 schema. Rendering support would diminish much of the opposition to it, as "it needs highway=bus_stop to be rendered" at the moment.

IIVQ:
IMHO, we have a deadlock here.
Much of the opposition to public_transport=platform (as a way or area, but also as a node) comes from it not being rendered on OSM-carto, thus necessitating a highway=bus_stop or railway=tram_stop node to be rendered.

Now if we don't include rendering to public_transport=platform (as node+way+area) because it is double-tagging with highway=bus_stop, we're having double tagging.

public_transport=platform is part of the widely used PT_v2 schema. Rendering support would diminish much of the opposition to it, as "it needs highway=bus_stop to be rendered" at the moment.

dieterdreist:
sent from a phone

> On 11. May 2019, at 18:48, IIVQ <notifications@github.com> wrote:
> 
> Much of the opposition to public_transport=platform (as a way or area, but also as a node) comes from it not being rendered on OSM-carto, thus necessitating a highway=bus_stop or railway=tram_stop node to be rendered.


it is also semantically questionable for a simple bus stop to refer as a platform to it, and the fact 2 tags are required for what was more elegantly done with one, also isn’t exactly convincing

IIVQ:
The public_transport=platform should be seen as "a place where people wait to board a bus/tram/train", not as a physical raised platform. There are more misnomers in the OSM tagging scheme (best example: highways with access=no, where a highway in British English is by legal definition a public access road). 
Also, it does unify very similar things for different modes (bus/tram+train) into one concept.

We could argue about the whole necessity of public_transport=platform all over again, but that's been done years ago (83 to 6 vote for the PTv2 schema in 2011) and there are other platforms(sic!) for this discussion.

dieterdreist:
sent from a phone

> On 12. May 2019, at 21:11, IIVQ <notifications@github.com> wrote:
> 
> Also, it does unify very similar things for different modes (bus/tram+train) into one concept.


is a bus stop really very similar to a train station? Maybe a train station is more similar to an airport, an aerial way station or a ferry port? It depends on the criteria, and I believe many people have not been convinced that the promoted benefit of pulling bus stops and rail based boarding under the same umbrella was worth the hassle that comes with changing a widely used, single tag to a combination of two tags.

IIVQ:
The public_transport=platform should be seen as "a place where people wait to board a bus/tram/train", not as a physical raised platform. There are more misnomers in the OSM tagging scheme (best example: highways with access=no, where a highway in British English is by legal definition a public access road). 
Also, it does unify very similar things for different modes (bus/tram+train) into one concept.

We could argue about the whole necessity of public_transport=platform all over again, but that's been done years ago (83 to 6 vote for the PTv2 schema in 2011) and there are other platforms(sic!) for this discussion.

PR #3233, opened 2018-05-13
Convert springs to use ST_PointOnSurface and reformat SQL

pnorman:
This partly replaces #3228 and is an example of converting a layer to use `ST_PointOnSurface`

### DIFF
--- a/project.mml
+++ b/project.mml
@@ -285,14 +285,22 @@ Layer:
       <<: *osm2pgsql
       table: |-
         (SELECT
-             way, "natural"
-           FROM planet_osm_polygon
-           WHERE "natural" IN ('spring')
-         UNION ALL
-         SELECT
-             way, "natural"
-           FROM planet_osm_point
-           WHERE "natural" IN ('spring')
+            way,
+            "natural"
+          FROM
+          (SELECT
+              ST_PointOnSurface(way) AS way,
+              "natural"
+            FROM planet_osm_polygon
+            WHERE way && !bbox!
+          UNION ALL
+          SELECT
+              way,
+              "natural"
+            FROM planet_osm_point
+            WHERE way && !bbox!
+            ) _
+          WHERE "natural" IN ('spring')
         ) AS springs
     properties:
       minzoom: 14
--- a/water-features.mss
+++ b/water-features.mss
@@ -167,7 +167,6 @@
 #springs {
   [natural = 'spring'][zoom >= 14] {
     marker-file: url('symbols/spring.svg');
-    marker-placement: interior;
     marker-clip: false;
   }
 }


project.mml
```diff
@@ -285,14 +285,22 @@ Layer:
...
+              "natural"
+            FROM planet_osm_polygon
+            WHERE way && !bbox!
+          UNION ALL
+          SELECT
+              way,
+              "natural"
+            FROM planet_osm_point
+            WHERE way && !bbox!
+            ) _
```
kocio-pl: Looks like unwanted space + underscore.

pnorman: What would you suggest as an alternative for what to call the inner subselect? `_` is a reasonably standard name when you just need *a* name. `s`, `p`, `q` are also common throw-away names.

kocio-pl: I don't know, however I don't like such shortcuts in the big SQL section of a public project. Typical `i` for loop index might be OK in typical code, but this one is already hard to maintain, underscore is worse than any letter and really looks like a typo, and you can't expect that other people will be real programmers to catch it. Something self-commenting like `inner_subselect` would be probably good.

BTW: there are more undocumented conventions and sections that I don't understand, it might make sense to add more comments one day.

sommerluk:
> Something self-commenting like `inner_subselect` would be probably good.

I support this. Longer identifiers do no harm during processing, but increase legibility of the code.

project.mml
```diff
@@ -285,14 +285,22 @@ Layer:
...
-         UNION ALL
-         SELECT
-             way, "natural"
-           FROM planet_osm_point
-           WHERE "natural" IN ('spring')
+            way,
+            "natural"
+          FROM
+          (SELECT
+              ST_PointOnSurface(way) AS way,
```
matthijsmelissen: Question: is psql smart enough to only do the ST_PointOnSurface calculation for objects satisfying `"natural" IN ('spring')`, in this query?

project.mml
```diff
@@ -285,14 +285,22 @@ Layer:
       <<: *osm2pgsql
       table: |-
         (SELECT
```
matthijsmelissen: Does wrapping this in a superquery accomplish anything at all? It seems to me that leaving lines 287-290 and 302 out would yield exactly the same result?

project.mml
matthijsmelissen: Ok, I see now it avoids the duplication of `WHERE "natural" IN ('spring')`.

matthijsmelissen:
This is essentially the same method as #3228 (after my last changes), which gives performance problems. We therefore should check if this PR causes performance problems too.

project.mml
pnorman: No, since then the query would return all polygons regardless of their `natural` column.

project.mml
pnorman: In *this* case the conditions on the tag columns are simple enough a subselect isn't necessary, but on longer queries, these can be over a dozen lines.

matthijsmelissen:
> This isn't the same method, since #3228 doesn't filter the points table by bounding box.

I fixed the filter on the points table now. Unfortunately, this still gives me a performance hit of about 50% of overall rendering time.

project.mml
pnorman: Postgres can do the query either way, the results are the same. It will pick whichever is cheapest, which will normally be using the index, filtering by geometry, then calculating st_pointonsurface.

matthijsmelissen: So if I understand you correctly, in this query it will also calculate ST_PointOnSurface for polygons that are not natural=spring, right?

project.mml
```diff
@@ -285,14 +285,22 @@ Layer:
...
+            FROM planet_osm_polygon
+            WHERE way && !bbox!
+          UNION ALL
+          SELECT
+              way,
+              "natural"
+            FROM planet_osm_point
+            WHERE way && !bbox!
+            ) _
+          WHERE "natural" IN ('spring')
```
matthijsmelissen: Should we not add another `WHERE way && !bbox!` in the outer subquery? Because now it will also forward points to Mapnik that are outside of the bounding box that is currently being rendered (as long as they are the label position of a polygon that falls within that bounding box).

pnorman: We could, but the PR is no different than current behavior. I suppose the difference is that filtering on contains is now cheap to do in-db

project.mml
pnorman: There is no guarantee one way or the other. It's unlikely to, but with a small database and the right indexes it might.

matthijsmelissen:
Marking this as WIP as I suspect this will cause performance issues.

matthijsmelissen:
Is anybody able to check the performance of this PR?

I think it would be good if we could give this attention, as it is a necessary step for the move to vector tikes.

pnorman:
> Is anybody able to check the performance of this PR?
> 
> I think it would be good if we could give this attention, as it is a necessary step for the move to vector tikes.

I found no performance issues when testing this and see nothing that could cause performance issues.

matthijsmelissen:
@pnorman I still don't fully understand why we need the subselect. Why can't we simply do this?

```
(SELECT
    "natural"
  FROM planet_osm_polygon
  WHERE way && !bbox!
    AND "natural" IN ('spring')
UNION ALL
SELECT
    way,
    "natural"
  FROM planet_osm_point
  WHERE way && !bbox!
    AND "natural" IN ('spring')
```

Is the only reason to avoid duplication of `"natural" IN ('spring')`, or is there a performance reason?

matthijsmelissen:
Regarding performance: I made a quick and dirty test to [extend the method proposed here to all layers](https://github.com/gravitystorm/openstreetmap-carto/tree/pointonsurface-2). This shows no performance degradation. So I'm happy to merge this - I'm sorry that I didn't review it sooner.

project.mml
kocio-pl: `inner_subselect` as a name would do for me.

pnorman:
> Is the only reason to avoid duplication of `"natural" IN ('spring')`, or is there a performance reason?

Performance is the same. It avoids duplication and is a different way to think about it. The duplication isn't a big issue here, but when we get to a query like the POI layers, there's a huge WHERE clause, and deduplicating that is a big gain.

matthijsmelissen:
Clear. You can merge it if you like (or I can merge it, as you prefer).

PR #3235, opened 2018-05-14
Adding rendering for historic=city_gate

### LINKED ISSUE #152: render label for historic=city_gate
City gates are mostly quite important features for orientation. Even if the citywalls or the city gate doesn't exist (or act as barriers) any more, they are usually still important for the local people as toponyms in way descriptions and the like. Please render historic=city_gate (not necessarily with an icon, but maybe like place=locality?).
http://taginfo.openstreetmap.org/tags/historic=city_gate

jragusa:
Fixes #152 

Show an icon and label for `historic=city_gate` in man_made colour

https://www.openstreetmap.org/way/317181351
Before
city_gate_before

After
city_gate_yvoire_manmade

### DIFF
--- a/amenity-points.mss
+++ b/amenity-points.mss
@@ -430,6 +430,13 @@
     }
   }
 
+  [feature = 'historic_city_gate'][zoom >= 17] {
+    marker-file: url('symbols/city_gate.svg');
+    marker-fill: @man-made-icon;
+    marker-placement: interior;
+    marker-clip: false;
+  }
+
   [feature = 'tourism_museum'][zoom >= 16] {
     marker-file: url('symbols/museum.svg');
     marker-fill: @culture;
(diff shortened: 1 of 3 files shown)


kocio-pl:
Thanks, good to see this code!

I just think it makes more sense to move it to z17+, because it can be a part of something bigger and should not appear too early - that's why I pushed defensive towers there to not compete with castle for example, see [this location](https://www.openstreetmap.org/way/228000809#map=17/51.95780/21.23458):

z15
kxnkzv2x

z16
48zvso3r

z17
t3mky52h

jragusa:
Fixed

That should be correct now.

kocio-pl:
You can try squashing commits into one by the way. There are multiple guides for this, like:

https://gist.github.com/patik/b8a9dc5cd356f9f6f980

but I do it with GUI called gitk under Linux. I make soft reset to the place right before my changes, so the commits are rolled back, while changed files are still there. Then I leave this app, make `git add .` and make new commit which holds all the changes. Then I push them with `-f` (force) option.

Quite easy and safe.

jragusa:
Done

I forgot that we can squash commits. Thanks to notice me @kocio-pl

kocio-pl:
Thanks, it's good to see some of the oldest tickets going away...

PR #3236, opened 2018-05-15
Adding rendering of lock_name

### LINKED ISSUE #157: Render lock_name rather than the name of the canal
The current rules for rendering labels on waterways add text-placement: point when they are tagged lock=yes, but the name tag mostly contains the canal name, so lock_name should probably be rendered instead. See http://osm.org/go/euzFr5X3C for an example.

I have a trivial patch at http://bob.cakebox.net/osm/lock_name.patch although that fetches the data from hstore which I'm not sure the tile.osm instance has.

jragusa:
Fixes #157

Render lock_name

https://www.openstreetmap.org/way/231720575
Before
lock_name_before

After
lock_name_after

### DIFF
--- a/project.mml
+++ b/project.mml
@@ -2498,6 +2498,7 @@ Layer:
             waterway,
             lock,
             name,
+            tags-> 'lock_name' AS lock_name,
             CASE WHEN tags->'intermittent' IN ('yes')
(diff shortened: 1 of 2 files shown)


kocio-pl:
Thanks, this was long overdue!

PR #3238, opened 2018-05-16
Add .editorconfig

### LINKED ISSUE #3209: Add .editorconfig
Perhaps we should add a config file for the line endings and stuff

http://editorconfig.org/

Ref #3200
Ref #3197

MattBlissett:
Fixes #3209 

This should configure editors to follow the indentation and whitespace behaviour the existing code has, using http://editorconfig.org/

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,21 @@
+# http://EditorConfig.org
+
+root = true
+
+# Defaults apply to all files
+[*]
+charset = utf-8
+end_of_line = lf
+insert_final_newline = true
+trim_trailing_whitespace = true
+indent_style = space
+indent_size = 2
+
+# 4 space indentation for Lua and Python
+[*.{lua,py}]
+indent_size = 4
+
+# Leave SVG files alone, we have a mix of tabs and spaces
+[*.svg]
+indent_style = undef
+indent_size = undef
--- a/.editorconfig
+++ b/.editorconfig
@@ -7,7 +7,6 @@ root = true
 charset = utf-8
 end_of_line = lf
 insert_final_newline = true
-trim_trailing_whitespace = true
 indent_style = space
 indent_size = 2
 


matthijsmelissen:
Looks good, thanks!

.editorconfig
```diff
@@ -0,0 +1,21 @@
+# http://EditorConfig.org
+
+root = true
+
+# Defaults apply to all files
+[*]
+charset = utf-8
+end_of_line = lf
+insert_final_newline = true
+trim_trailing_whitespace = true
```
matthijsmelissen: This one I'm not sure about: obviously we don't want trailing whitespace, but we also don't want commits to include trimming of trailing whitespace in unrelated code. Such trailing whitespace might be introduced accidentally by users with editors not honoring .editconfig.

What do you think is best?

MattBlissett: I have `trim_trailing_whitespace = true` for all my work projects, and haven't had any problems — but I'm mostly working with people using IDEs or editors that have built-in support for Editorconfig.

One other option is a Git pre-commit hook, but this [looks complicated](https://github.com/imoldman/config/blob/master/pre-commit.git.sh) so I won't recommend it.

matthijsmelissen: Another option would be a travis check that fails if people add trailing whitespace.

kocio-pl: I prefer Travis. Trimming is turned on by default in Atom for example and it caused some PRs to be cluttered with such small fixes not related to the subject.

MattBlissett: If Travis were set to fail because of trailing whitespace, then adding the same rule in Editorconfig is convenient for users whose editors support it.

At present, the only files with trailing whitespace are Markdown — they could be excluded from this rule.

kocio-pl:
Maybe any kind of pushing people to remove trailing spaces is bad, since automatic removing them in the background creates mess and Travis will not make them create separate patches anyway. Since this problem is quite small and simple solutions make more harm than good, I would rather prefer that some checking and cleaning would be run periodically.

matthijsmelissen:
Thanks!

PR #3245, opened 2018-05-24
Restore compatibility with older pgsql

Zverik:
Looks like some people have missed the memo. Please enclose accessors in checks like `(tags->'whatever') IS NULL` in parentheses.

This PR fixes [this help request](https://help.openstreetmap.org/questions/63612/issue-starting-render), see also #3078.

### DIFF
--- a/project.mml
+++ b/project.mml
@@ -695,7 +695,7 @@ Layer:
             COALESCE((
               'highway_' || (CASE WHEN highway IN ('residential', 'unclassified', 'pedestrian', 'service', 'footway', 'cycleway', 'track', 'path', 'platform') THEN highway ELSE NULL END)),
               ('railway_' || (CASE WHEN (railway IN ('platform') 
-                              AND (tags->'location' NOT IN ('underground') OR tags->'location' IS NULL)
+                              AND (tags->'location' NOT IN ('underground') OR (tags->'location') IS NULL)
                               AND (tunnel NOT IN ('yes', 'building_passage') OR tunnel IS NULL)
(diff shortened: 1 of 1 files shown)


project.mml
```diff
@@ -695,15 +695,15 @@ Layer:
...
+                              AND (tags->'location' NOT IN ('underground') OR (tags->'location') IS NULL)
                               AND (tunnel NOT IN ('yes', 'building_passage') OR tunnel IS NULL)
                               AND (covered NOT IN ('yes') OR covered IS NULL))
                               THEN railway ELSE NULL END))
             ) AS feature
           FROM planet_osm_polygon
           WHERE highway IN ('residential', 'unclassified', 'pedestrian', 'service', 'footway', 'track', 'path', 'platform')
             OR (railway IN ('platform') 
-                AND (tags->'location' NOT IN ('underground') OR tags->'location' IS NULL)
+                AND (tags->'location' NOT IN ('underground') OR (tags->'location') IS NULL)
```
jbelien: Shouldn't it be :
```
AND ((tags->'location') NOT IN ('underground') OR (tags->'location') IS NULL)
```

If yes, the other lines need to be updated as well.

kocio-pl:
It's not people, that was me, sorry. I will try to remember that rule.

HolgerJeromin:
Can we have an travis check for that?

pnorman:
> Can we have an travis check for that?

We'd first need a test for it, then the test could be added to travis.

Issue #3244, opened 2018-05-22
Render  barrier=toll_booth areas


Ivans-Git:
Render barrier=toll_booth, area=yes.

*Edit: Typo fix*

pnorman:
How do you think it should be rendered? Just a symbol in the centre of the area?

#3005 had some discussion at the time about areas but it couldn't be solved with the DB schema of the time.

sommerluk:
@IVAN-RSM  Thanks for the issue report!

Currently, we do not treat `barrier=toll_booth` as area in our database layout. Or at least not in general – when it is used together with other area features like `building=yes` than it is treated as area, otherwise not.

Independent of the discussion how to render this (I agree that just a symbol in the centre of the area would be a good starting point), we need to change the database layout. I’ve opened a ticket therefore at #3249 A database layout change is likely to take some time.

sommerluk:
@pnorman My fear in #3005 was that we render an icon for some of the closed ways (for example these that also have a building or landuse tag), but for other closed ways we don’t – and that might be confusing for the mapper. As you say in #3249 `area=yes` forces treatment as area, so this combination can be rendered without a database layout change. Do you think it makes sense? (If so, I could create a PR…)

jeisenbe:
Requires merging #3788 and reloading database first

jeisenbe:
I opened #3788 but now we need to consider if we should really support this method of mapping.

Currently, almost all `barrier` features are either mapped as linear ways (fences, walls, hedges and similar), or as nodes (for barriers which block a `highway`). 

While a few types of node `barrier` have been mapped as linear ways, for example long gates, there is some discussion about whether it would actually be a good idea to support this. If mappers use only a 2-node way to map a gate, this does not add more information than using a 'width' tag, but it does make it harder for routers and other database users to properly interpret the highway as obstructed by the gate or other barrier. If 3 nodes are used, then it's not clear if the central node should also be tagged as a `barrier` or not. See discussion at #3929 and please comment there as well.

[...]

PR #3249, opened 2018-05-28
[DO NOT MERGE] Treat barrier=toll_booth as area

sommerluk:
Fixes part of #3244

Currently, we treat `barrier=toll_booth` as _line_ geometry, just like all other `barrier`.

Resolving #3244 needs `barrier=toll_booth` as _area_ geometry.

We could make a geometry transformation to get an area in the SQL query with `ST_BuildArea(way)` even on line geometries. But #2510 has proven that there are pitfalls that can make serious performance problems. So I think it’s better to avoid this.

Without judging here weather it is a good idea or not to _render_ `barrier=toll_booth` on areas in our style, I think in any case it would be more correct to treat `barrier=toll_booth` as area geometry in our database tables.

This PR changes the lua file to make this happen. This would mean a database reload. I think this change is not important enough for a database reload. But maybe we can collect somewhere changes like this to the database tables, just to have them ready for a later database reload…

### DIFF
--- a/openstreetmap-carto.lua
+++ b/openstreetmap-carto.lua
@@ -43,6 +43,7 @@ local linestring_values = {
 
 -- Objects with any of the following key/value combinations will be treated as polygon
 local polygon_values = {
+    barrier = {toll_booth = true},
     highway = {services = true, rest_area = true},
     junction = {yes = true}
 }


pnorman:
> Resolving #3244 needs barrier=toll_booth as area geometry.

No, it doesn't. That issue is about ways explicitly tagged with `area=yes`

kocio-pl:
As far as I understand, OSMF admins plan to reimport database, so maybe you should contact them to discuss the time frame (if the issue really needs any change in lua) - see https://github.com/openstreetmap/chef/issues/168#issuecomment-399461921.

kocio-pl:
@sommerluk Hi, just revisiting PR - did you find anything more in the meantime? Any plans to do with this code from today perspective? Do you need some help with it?

sommerluk:
Closing in favour of #3611

PR #3255, opened 2018-06-02
Check all YAML files are valid YAML

pnorman:
This checks with find rather than an explicit list, because any .yml or .yaml files should be YAML, and should be valid.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,6 +8,7 @@ addons:
     packages:
     - lua5.1
     - libxml2-utils
+    - python-yaml
 env:
   - CARTO=0.18.0 MAPNIK='3.0.0 3.0.12'
 install:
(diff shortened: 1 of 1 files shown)


pnorman:
COI disclaimer: This arose out of WMF work with kartotherian/meddo#15

kocio-pl:
Do you think it's useful here (checking `.yaml` and `.yml` files) or just that it won't hurt?

pnorman:
I think it's useful. I wouldn't bother if it were hard to test, but it's an easy and quick check. I've also found python-yaml's error messages easier to read than those that come from nodejs and carto, so in the case of a MML syntax error, it'll help there too.

sommerluk:
👍

kocio-pl:
Ok, I forgot about other YAML files we have in the repo, so it makes even more sense. I'm not able to test the code, so please merge it on your own.

Issue #3041, opened 2018-01-29
Dog park is too prominent


polarbearing:
Dog park was added in #2216, with colour changed to playground in #2268, with discussions in #341, #2250 and other places.

Currently is works reasonably with small, single-function parks. 
For larger areas, the pattern symbol 'paw' is very strong, it is more like lots of icons. 

When dog park is an additional function to another area, commonly a forest or a park, it becomes problematic. In the forest case, the patterns interfere badly, and the leisure colour prevails over the main colour of the basic feature:
![dog_over_forest](https://user-images.githubusercontent.com/8266355/35508757-d600df1e-04f1-11e8-9db3-8c456fbd656a.png)

People [started to remove the tags](https://www.openstreetmap.org/changeset/55448617) because the prominence annoyed them.

For the small features, a single icon would be sufficient. For the larger ones they would be candidates for an outline, however a single icon would be lost when zooming in. 

Or maybe a lighter pattern but no fill at all, so the pattern shows on whatever land the feature is on. For the forest case, would it be possible to sync the paw and the tree patterns?

dieterdreist:
I believe tagging significant or all parts of an area like the Grunewald
forest is abusing the tag. Just because dogs can (supposedly) be unleashed,
this isn't a dog park. Compare to the sahara, should we tag this all as dog
park because you can let your unleashed dogs play?

kocio-pl:
Accidentally I was [talking about this specific area](https://www.openstreetmap.org/changeset/40873183) about 2 months ago and for me a [sign in German](https://www.top10berlin.de/sites/top10berlin.de/files/styles/juicebox/public/location/slider/2014/09/08/slider_hundeauslaufgebiet_grunewald_top10_1.jpg?itok=I5i-knkt) was enough to leave a dog park tag.

It seems to me that even big dog park would look OK, the problem is mixing with trees. I don't know how to solve this.

meased:
Agree that the screenshot is not pretty. Is this a dog park problem, though? Or is there a general problem of overlapping pattern symbols? (What about cemetery and trees?)

polarbearing:
Single icon such as playground, and outline instead of fill. There is no reason why kids get just ine icon, and dogs the pattern all over ;-) The dogs are not growing there like in an orchard. 
Icon appearing depending on way_size.

Before:
[...]

PR #3256, opened 2018-06-03
Remove hidden icon of dog_park

jragusa:
`leisure=dog_park` should display an icon which is hidden and this shifts the label which is not centered. The following PR removes unnecessary code to show icon and redundancy.

https://www.openstreetmap.org/way/554648250

Before
dog_park_before

After
dog_park_after

### DIFF
--- a/amenity-points.mss
+++ b/amenity-points.mss
@@ -1093,13 +1093,6 @@
     marker-clip: false;
   }
 
-  [feature = 'leisure_dog_park'][zoom >= 17] {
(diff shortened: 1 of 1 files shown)


polarbearing:
There is a discussion to use a single icon and outline, instead of the paw pattern, in #3041. I'd prefer that to any repairs on the current pattern.

kocio-pl:
Yes, we need decision first before taking any action.

jragusa:
Effectively, I missed this discussion. I'm closing this PR

PR #3260, opened 2018-06-05
Add curl option

giggls:
When downloading files it is a good idea to set the timestamp of the remote file for the downloaded file which is better for subsequent updates than the download time.


sommerluk:
Thanks!

PR #3267, opened 2018-06-17
Finer man_made=pier width

### LINKED ISSUE #2652: man_made=pier, finer? color?
[Rendering is to big on z17.](https://www.openstreetmap.org/#map=17/52.68170/5.94702)
pier

Also other zoom levels needs a finer adjustment. higher and lower z.

[z16](https://www.openstreetmap.org/#map=16/52.6811/5.9478)

Does it need to be visible on z12 - z14, z15.
I do not think so, maybe z15 could be discussed.
Is it needed for the overall view? Me, no.

Is white the best color for a pier?
This white looks, if it is not balanced with the surrounding.

meased:
Fixes #2652

- Lowers the stroke width of man_made=pier
- Adds square line-cap to remove rendering artifacts
- Does not effect piers mapped as an area

From the discussion in #2652, there was some disagreement on whether piers should represent real world geometries or abstract lines. My opinion is that it doesn't matter which camp you're in, piers are currently much wider in the map than they are in the real world, and as an abstract line, they are still too wide (they are wider than residential roads at zooms 12, 13 and 14).

At [this sample area.](https://www.openstreetmap.org/#map=14/45.6079/-122.6679)

Before
before_12
After
after_12

Before
before_13
After
after_13

Before
before_14
After
after_14

Before
before_15
After
after_15

Before
before_16
After
after_16

Before
before_17
After
after_17

Before
before_18
Notice the gaps in the line rendering.
After
after_18

Before
before_19
After
after_19

### DIFF
--- a/water-features.mss
+++ b/water-features.mss
@@ -70,10 +70,12 @@
       polygon-fill: @land-color;
     }
     #piers-line {
-      line-width: 1.5;
+      line-width: 0.5;
       line-color: @land-color;
-      [zoom >= 13] { line-width: 3; }
-      [zoom >= 16] { line-width: 7; }
+      line-cap: square;
+      [zoom >= 13] { line-width: 1; }
+      [zoom >= 15] { line-width: 2; }
+      [zoom >= 17] { line-width: 4; }
     }
   }
 


water-features.mss
```diff
@@ -70,10 +70,12 @@
       polygon-fill: @land-color;
     }
     #piers-line {
-      line-width: 1.5;
+      line-width: 0.5;
       line-color: @land-color;
-      [zoom >= 13] { line-width: 3; }
-      [zoom >= 16] { line-width: 7; }
+      line-cap: square;
```
HolgerJeromin: All roads have round as cap.

imagico:
For better understanding - this is the line width on the ground (in meters) the proposed change would result in for the different zoom levels.

zoom|line-width|ground width (Equator)|ground width (60°)
----|----------|----------------------|------------------
12|0.5|19|9.5
13|1|19|9.5
14|1|9.5|4.8
15|2|9.5|4.8
16|2|4.8|2.4
17|4|4.8|2.4
18|4|2.4|1.2

water-features.mss
meased: Round would look best when two sections join at a non-right angle, but I didn't know how people would feel about round pier ends.

HolgerJeromin: Yes, the pier ending. You are right

kocio-pl:
I like this proposition.

kocio-pl:
Regarding width table - I think they are so thin after the change, that it's just a hint that they are here. Which is good, because they are too visible now. And if you want real size and shape, you should map them as area - which is preferred as more accurate.

PR #3285, opened 2018-07-03
Add rendering for amenity=casino

### LINKED ISSUE #3210: Add icon for amenity=casino
Quite popular feature (4,8k uses) without any rendering at the moment.
https://wiki.openstreetmap.org/wiki/Tag:amenity=casino

Potential icon could be used also for leisure=adult_gaming_centre (2,1k uses).
https://wiki.openstreetmap.org/wiki/Tag:leisure%3Dadult_gaming_centre

Icon proposals:
*  (@lakedistrictOSM)
* amenity casino (@Tomasz-W)

meased:
Fixes #3210 

Adds rendering for amenity=casino

This marked as WIP as I think there are a couple of decisions to make:

1. amenity=casino + tourism=hotel are likely to occur on the same building (I think), which should have higher priority?
2. What zoom level to start on? 16 "feels" the best on the test areas, but hotels start at 17.

Test areas (you would think Las Vegas would be a good test area, but there is surprisingly bad tagging there):

https://www.openstreetmap.org/#map=16/48.0883/-122.1897
Before
tulalip_before_16
tulalip_before_17

After
tulalip_16
tulalip_17

https://www.openstreetmap.org/#map=17/44.99815/-124.00950
Before
chinook_before_16
chinook_before_17
After
chinook_16
chinook_17

https://www.openstreetmap.org/#map=16/45.8534/-122.7062

Before
ilani_before_16
ilani_before_17
ilani_before_18

After
ilani_16
ilani_17
ilani_18


Tomasz-W:
What do you think about leisure-green for this feature? I know it's not "sport-like" one, but it's still kind of a leisure object.

meased:
I have removed the WIP status of this PR. The issue I had was that when zooming in to an object tagged with both casino and hotel, the rendering would switch from casino to hotel, which I found jarring. I have fixed this by preferring the casino rendering. (Though I had a hard time finding examples of areas tagged like this so maybe it's not so common after all.)

I am open to other solutions/suggestions.

lakedistrictOSM:
I'd probably go for z17 over z16 - z16 seems too early, especially in places like the UK where casinos are much smaller than those in the USA.

lakedistrictOSM:
> (Though I had a hard time finding examples of areas tagged like this so maybe it's not so common after all.)

Stick `amenity=casino and tourism=hotel global` into the wizard on overpass turbo

Tomasz-W:
As more often casino is a part of hotel, not the opposite, I think `tourism=hotel` should have the priority.

meased:
> I'd probably go for z17 over z16

This is probably the best solution. Hotels render at 17 so the issue resolves itself.

meased:
Rendering now starts at z17, the same as hotel. No rendering change for places tagged with hotel and casino. Other rendering are the same as the above samples (just disregard the z16 images).

imagico:
Seems i am somewhat late in commenting here - i hadn't noticed the 'WIP' note had been removed and given the controversial nature of the subject i had expected a broader discussion before such a change would be merged.

`amenity=casino` is the tag for a _gambling establishment_, i.e. a place where money is bet on the outcome of certain games.  This is something fundamentally different from the main purpose games have in most societies in general, namely amusement and social interaction.  This means the symbol chosen here is likely misleading for a significant fraction of potential map users by implying games but not in any way indicating the gambling aspect.  This alone would IMO make the symbol unsuitable for this kind of feature (misleading to the map user and wrong incentive to the mapper to tag anything came related as amenity=casino).  And given the addictive potential of gambling you could in this case consider the choice of symbol as trivializing addiction risks.  Showing casinos on the map is one thing, showing them in a way that is likely to indicate to the unsuspecting user this could be a place to play games for recreation and social interaction is something very different.

kocio-pl:
Dices are traditionally related to gambling, so I see nothing wrong with using them as a symbol of casino. For `shop=games` (which has still low usage) I would rather use chess figure for example.

matthijsmelissen:
Agree with @kocio-pl.

imagico:
When viewed globally dice are certainly far more used as a device for amusement and social interaction games than for gambling.  So i stand by my comment.  A huge fraction of the potential map users observe recreational dice games in public on a daily basis but have not seen a casino up close in their whole life.  In Europe today recreational games played in public play a very small role compared to other parts of the world.  This combined with the strong presence of casino games in contemporary movies and television might give you a selective impression of the subject.

kocio-pl:
Recreational dice games in public is not too popular. I'm not even sure if we have tagging for this - probably it's just what can happen in some shop games.

I also think that showing risk-related places (like military area, alcohol shop or casino) might as well make people avoid such areas.

Issue #2189, opened 2016-06-21
remove "text-min-distance" in placenames.mss


matkoniecz:
as documented on https://github.com/mapbox/carto/blob/master/docs/latest.md it sets

> Minimum distance to the next label with the same text. **Only works for line placement.**

matkoniecz:
It is interesting. It turned out that removing `text-min-distance` significantly increases number of displayed labels.

It seems to be a positive change, I will submit this change anyway - but for a different reasons. It improves rendering at low zooms (12+ needs testing) and it makes style closer to working the same in carto 2.3 and carto 3.0.

Initial confusion was caused by redefining  `text-min-distance` in carto 2.3 means

> Minimum permitted distance to the next text symbolizer.

and was changed to

> Minimum distance to the next label with the same text. Only works for line placement. Deprecated: replaced by text-repeat-distance and text-margin

In carto 3.0.

https://github.com/mapbox/carto/blob/master/docs/2.3.0.md#text-min-distance-float
https://github.com/mapbox/carto/blob/master/docs/3.0.0.md#text-min-distance-float

current state
https://cloud.githubusercontent.com/assets/899988/16315567/14e33c72-3983-11e6-8445-0f48209eeb8c.png

without `text-min-distance`
https://cloud.githubusercontent.com/assets/899988/16315568/14e613c0-3983-11e6-98b3-d9d599ac1d91.png

https://cloud.githubusercontent.com/assets/899988/16315599/3132cd70-3983-11e6-80b6-122259c28b9d.png
https://cloud.githubusercontent.com/assets/899988/16315619/454897fe-3983-11e6-9814-f1572ba6ef51.png
https://cloud.githubusercontent.com/assets/899988/16315621/4555732a-3983-11e6-8c6b-978d522a4ffe.png
https://cloud.githubusercontent.com/assets/899988/16315620/4552fb7c-3983-11e6-9f00-e3819b406c66.png
https://cloud.githubusercontent.com/assets/899988/16315633/51d12eaa-3983-11e6-90d1-55d8a1b04deb.png
https://cloud.githubusercontent.com/assets/899988/16315634/51d3ba12-3983-11e6-9776-089425f940e5.png
https://cloud.githubusercontent.com/assets/899988/16315639/52086230-3983-11e6-805a-8bf2e010db98.png
https://cloud.githubusercontent.com/assets/899988/16315640/520b7056-3983-11e6-86cb-5f84e47e2129.png
[...]

Issue #1956, opened 2015-11-04
Road ref shields & others block too many road names


tomchance:
I'm submitting this issue on the assumption that the default style wasn't supposed to be a car touring map, and that - certainly in cities - major road names are far more useful than refs.

Currently, refs (and in places labels for other features) mean many major road names don't render. This is especially noticeable at zoom levels 13-15, where some road names _do_ show up but are almost always left out in favour of ref shields.

Some examples:

![image](https://cloud.githubusercontent.com/assets/460737/10942515/60e1f7a8-8308-11e5-9b30-35901cc08e94.png)

![image](https://cloud.githubusercontent.com/assets/460737/10942555/753e95d0-8308-11e5-8156-901924d4ad59.png)

![image](https://cloud.githubusercontent.com/assets/460737/10942578/94d4964c-8308-11e5-8678-e26d041adcbe.png)

dieterdreist:
> Am 04.11.2015 um 16:28 schrieb tomchance <notifications@github.com>:
> 
> that - certainly in cities - major road names are far more useful than refs.


yes, in cities the names are far more important while in villages and towns you'd more likely be interested in the refs, especially in medium to low zoom levels

kocio-pl:
Example from my hometown: all streets around [this cemetery](https://www.openstreetmap.org/#map=15/52.2770/21.0372) has a lot of refs and not a single name.

@dieterdreist You're right. This reminds me about https://github.com/gravitystorm/openstreetmap-carto/issues/1705 or https://github.com/gravitystorm/openstreetmap-carto/pull/1884. The more I look at such issues, the more I think about domain- or environment-based rendering. I'd like this style to be more context sensitive (like for example "urban", "rural" or "outdoor"). What do you - I mean all of you here - think about it? Do you agree it is needed at all? Any remarks on how it could be done?

mboeringa:
> @dieterdreist You're right. This reminds me about #1705 or #1884. The more I look at such issues, the more I think about domain- or environment-based rendering. **I'd like this style to be more context sensitive (like for example "urban", "rural" or "outdoor"). What do you - I mean all of you here - think about it?**

My problem with these kind of proposals and similar ones like "country" specific road type renderings or shields, that abound here and on the different mailing lists and forums, is that each of them, however sensible in its own right, tends to significantly add to the number of render rules that need to be defined and managed. It quickly adds up if you need to take care of dozens - or even hundreds - of zone specific renderings.

[...]

PR #3295, opened 2018-07-06
Increased text and shield distances on waterways, railways and roads

### LINKED ISSUE #3111: Don't render duplicate label/shield for connected ways with the same computed style
When there are several ways connected one after the other, their labels are rendered, if possible, one after the other, which leads to such renderings:
image

You can see here that the _Ruisseau de la Tuilotte_ label is duplicated because the way has been splitted several times because of tagging differences. Still, the rendering of the constituting ways is always the same, but the label is duplicated. Depending on the case, there can be tens of such duplicates. The problem also arises for, for example, road shields:
image

My point is, this behavior clutters the tiles with duplicated information: when the name of a waterway is always the same, there is no point in rendering it nonstop, and that can be applied for virtually any label or shield; couldn't we use [`repeat-distance`](http://mapnik.org/mapnik-reference/#3.0.6/text-repeat-distance), as explained on mapnik/mapnik#3869? We could at least use it to limit such duplications.

Penegal:
This PR fixes #3111: it limits redundant shields and stream labels by using `shield-repeat-distance` and `text-repeat-distance`. The `*-min-distance` versions of these attributes were not used, because they didn't change rendering, even with very large values (i.e. about 1000); besides, this PR drops `shield-min-distance`, because Mapnik complained about using this `deprecated option minimum-distance with new options margin and repeat-distance.`

I wandered around to see if the new version does not remove too much labels and shields, but there are still one or two at each view I took, so I assume these values are OK, but I can test on other places if requested, as long as the Geofabrik dump is not too big.

* Before
image
image
* After
image
image

### DIFF
--- a/roads.mss
+++ b/roads.mss
@@ -281,7 +281,7 @@
 @shield-size-z18: 12;
 @shield-line-spacing-z18: -1.80; // -0.15 em
 @shield-spacing: 760;
-@shield-min-distance: 40;
+@shield-repeat-distance: 600;
 @shield-font: @book-fonts;
 @shield-clip: false;
 
@@ -2665,7 +2665,7 @@ tertiary is rendered from z10 and is not included in osm_planet_roads. */
     shield-line-spacing: @shield-line-spacing;
     shield-placement: line;
     shield-spacing: @shield-spacing;
-    shield-min-distance: @shield-min-distance;
+    shield-repeat-distance: @shield-repeat-distance;
     shield-face-name: @shield-font;
     shield-clip: @shield-clip;
 
@@ -2713,7 +2713,7 @@ tertiary is rendered from z10 and is not included in osm_planet_roads. */
 
       shield-placement: line;
       shield-spacing: @shield-spacing;
-      shield-min-distance: @shield-min-distance;
+      shield-repeat-distance: @shield-repeat-distance;
       shield-face-name: @shield-font;
       shield-clip: @shield-clip;
 
--- a/roads.mss
+++ b/roads.mss
@@ -283,7 +283,8 @@
 @shield-size-z18: 12;
 @shield-line-spacing-z18: -1.80; // -0.15 em
 @shield-spacing: 760;
-@shield-repeat-distance: 600;
+@shield-repeat-distance: 400;
+@shield-margin: 40;
 @shield-font: @book-fonts;
 @shield-clip: false;
 
@@ -2668,6 +2669,7 @@ tertiary is rendered from z10 and is not included in osm_planet_roads. */
     shield-placement: line;
     shield-spacing: @shield-spacing;
     shield-repeat-distance: @shield-repeat-distance;
+    shield-margin: @shield-margin;
     shield-face-name: @shield-font;
     shield-clip: @shield-clip;
 
@@ -2716,6 +2718,7 @@ tertiary is rendered from z10 and is not included in osm_planet_roads. */
       shield-placement: line;
       shield-spacing: @shield-spacing;
       shield-repeat-distance: @shield-repeat-distance;
+      shield-margin: @shield-margin;
       shield-face-name: @shield-font;
       shield-clip: @shield-clip;
 
--- a/roads.mss
+++ b/roads.mss
@@ -272,6 +272,8 @@
 @track-oneway-arrow-color:        darken(@track-fill, 15%);
 @bridleway-oneway-arrow-color:    darken(@track-fill, 10%);
 
+@text-repeat-distance: 600;
+
 // Shield’s line wrap is based on OpenStreetMap data and not on line-wrap-width,
 // but lines are typically rather short, so we use narrow line spacing.
 @shield-size: 10;
@@ -2830,6 +2832,7 @@ tertiary is rendered from z10 and is not included in osm_planet_roads. */
       text-clip: false;
       text-placement: line;
       text-face-name: @book-fonts;
+      text-repeat-distance: @text-repeat-distance;
       [tunnel = 'no'] {
         text-halo-radius: @standard-halo-radius;
         [highway = 'motorway'] { text-halo-fill: @motorway-fill; }
@@ -2862,6 +2865,7 @@ tertiary is rendered from z10 and is not included in osm_planet_roads. */
       text-face-name: @book-fonts;
       text-halo-radius: @standard-halo-radius;
       text-halo-fill: @secondary-fill;
+      text-repeat-distance: @text-repeat-distance;
     }
     [zoom >= 14] {
       text-size: 9;
@@ -2888,6 +2892,7 @@ tertiary is rendered from z10 and is not included in osm_planet_roads. */
       text-face-name: @book-fonts;
       text-halo-radius: @standard-halo-radius;
       text-halo-fill: @tertiary-fill;
+      text-repeat-distance: @text-repeat-distance;
     }
     [zoom >= 17] {
       text-size: 11;
@@ -2906,6 +2911,7 @@ tertiary is rendered from z10 and is not included in osm_planet_roads. */
     text-halo-radius: @standard-halo-radius;
     text-halo-fill: @standard-halo-fill;
     text-face-name: @book-fonts;
+    text-repeat-distance: @text-repeat-distance;
 
     [zoom >= 17] {
       text-size: 11;
@@ -2932,6 +2938,7 @@ tertiary is rendered from z10 and is not included in osm_planet_roads. */
       text-halo-radius: @standard-halo-radius;
       text-halo-fill: @residential-fill;
       text-face-name: @book-fonts;
+      text-repeat-distance: @text-repeat-distance;
     }
     [zoom >= 16] {
       text-size: 9;
@@ -2961,6 +2968,7 @@ tertiary is rendered from z10 and is not included in osm_planet_roads. */
       [highway = 'raceway'] { text-halo-fill: @raceway-fill; }
       [highway = 'service'] { text-halo-fill: @service-fill; }
       text-face-name: @book-fonts;
+      text-repeat-distance: @text-repeat-distance;
     }
     [zoom >= 17] {
       text-size: 11;
@@ -2982,6 +2990,7 @@ tertiary is rendered from z10 and is not included in osm_planet_roads. */
       [highway = 'living_street'] { text-halo-fill: @living-street-fill; }
       [highway = 'pedestrian'] { text-halo-fill: @pedestrian-fill; }
       text-face-name: @book-fonts;
+      text-repeat-distance: @text-repeat-distance;
     }
     [zoom >= 16] {
       text-size: 9;
@@ -3021,6 +3030,7 @@ tertiary is rendered from z10 and is not included in osm_planet_roads. */
       text-face-name: @book-fonts;
       text-vertical-alignment: middle;
       text-dy: 5;
+      text-repeat-distance: @text-repeat-distance;
     }
     [zoom >= 16] {
       text-size: 9;
@@ -3049,6 +3059,7 @@ tertiary is rendered from z10 and is not included in osm_planet_roads. */
       text-face-name: @book-fonts;
       text-vertical-alignment: middle;
       text-dy: 7;
+      text-repeat-distance: @text-repeat-distance;
     }
     [zoom >= 17] {
       text-size: 11;
@@ -3173,10 +3184,10 @@ tertiary is rendered from z10 and is not included in osm_planet_roads. */
       text-spacing: 900;
       text-clip: false;
       text-placement: line;
-      text-min-distance: 18;
       text-face-name: @book-fonts;
       text-halo-radius: @standard-halo-radius;
       text-halo-fill: @standard-halo-fill;
+      text-repeat-distance: @text-repeat-distance;
     }
     [zoom >= 19] {
       text-size: 11;
@@ -3195,10 +3206,10 @@ tertiary is rendered from z10 and is not included in osm_planet_roads. */
         text-spacing: 300;
         text-clip: false;
         text-placement: line;
-        text-min-distance: 18;
         text-face-name: @book-fonts;
         text-halo-radius: @standard-halo-radius;
         text-halo-fill: @standard-halo-fill;
+        text-repeat-distance: @text-repeat-distance;
       }
       [zoom >= 13] {
         text-dy: 6;
@@ -3224,10 +3235,10 @@ tertiary is rendered from z10 and is not included in osm_planet_roads. */
         text-spacing: 300;
         text-clip: false;
         text-placement: line;
-        text-min-distance: 18;
         text-face-name: @book-fonts;
         text-halo-radius: @standard-halo-radius;
         text-halo-fill: @standard-halo-fill;
+        text-repeat-distance: @text-repeat-distance;
       }
       [zoom >= 17] {
         text-spacing: 600;
@@ -3254,10 +3265,10 @@ tertiary is rendered from z10 and is not included in osm_planet_roads. */
       text-spacing: 900;
       text-clip: false;
       text-placement: line;
-      text-min-distance: 18;
       text-face-name: @book-fonts;
       text-halo-radius: @standard-halo-radius;
       text-halo-fill: @standard-halo-fill;
+      text-repeat-distance: @text-repeat-distance;
     }
     [zoom >= 19] {
       text-size: 11;
--- a/roads.mss
+++ b/roads.mss
@@ -272,7 +272,7 @@
 @track-oneway-arrow-color:        darken(@track-fill, 15%);
 @bridleway-oneway-arrow-color:    darken(@track-fill, 10%);
 
-@text-repeat-distance: 600;
+@text-repeat-distance: 50;
 
 // Shield’s line wrap is based on OpenStreetMap data and not on line-wrap-width,
 // but lines are typically rather short, so we use narrow line spacing.
@@ -2941,7 +2941,6 @@ tertiary is rendered from z10 and is not included in osm_planet_roads. */
       text-halo-radius: @standard-halo-radius;
       text-halo-fill: @residential-fill;
       text-face-name: @book-fonts;
-      text-repeat-distance: @text-repeat-distance;
     }
     [zoom >= 16] {
       text-size: 9;
@@ -2993,7 +2992,6 @@ tertiary is rendered from z10 and is not included in osm_planet_roads. */
       [highway = 'living_street'] { text-halo-fill: @living-street-fill; }
       [highway = 'pedestrian'] { text-halo-fill: @pedestrian-fill; }
       text-face-name: @book-fonts;
-      text-repeat-distance: @text-repeat-distance;
     }
     [zoom >= 16] {
       text-size: 9;
--- a/water.mss
+++ b/water.mss
@@ -2,6 +2,8 @@
 @glacier: #ddecec;
 @glacier-line: #9cf;
 
+@text-repeat-distance: 600;
+
 #water-areas {
   [natural = 'glacier']::natural {
     [zoom >= 8] {
@@ -266,6 +268,7 @@
       text-halo-fill: @standard-halo-fill;
       text-spacing: 400;
       text-placement: line;
+      text-repeat-distance: @text-repeat-distance;
       [zoom >= 14] { text-size: 12; }
       [int_tunnel = 'yes'] { text-min-distance: 200; }
     }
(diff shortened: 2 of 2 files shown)


Adamant36:
I don't know if it would work for shields or not, but with text there is also text-min-path-length so labels can skip smaller segments. It might be worth trying on them both and seeing if you can find a good balance between that and text-repeat-distance. If you don't, the distance will probably still be a problem where lines are split up a lot.

Penegal:
@Adamant36: you seem to assume that the lines are not too split; I don't think we should assume much on the underlying data, because there can always be cases where the assumption is false. For instance, relations or `destination[:*]` or `*:lanes` tags can lead to legitimately making very short ways, so we should have a style accounting for that.

Adamant36:
I wasn't assuming anything. I Iooked at the code and noticed it wasnt in there and thought you might want to test it out. Its your prerogative though. Its not like someone cant create another issue later if need be. I'm not sure how relations etc would affect it, but thats what testing is for. You should just write off trying something because theres other things you think might cause it to not work because you never know. I just spent 3 days tweaking around with getting golf course rendering to work and even if it was a massive hassle I learned a lot in the proccess. So it wasnt a total waste.  

Also, I did do a pull request on a similar issue and know from that one that the more testing in various situations with different values the better. Its up to you what you want to try out though.

Penegal:
I don't know other cases where this correction would be effective, as it applies only to small related and connected ways; it's quite specific, so I have to search for other locations to test it, but, if you know some, I'd be happy to test this code on it, as long as the Geofabrik dump is not too heavy.

Penegal:
Some other test cases (before/after):

image
image

image
image

image
image

image
image

Adamant36:
No worries. It happens. This was one of the issues I had on my list to deal with eventually if know one else did and I was looking forward to seeing how text-repeat-distance would work out. So thanks for giving it a try. It seems to have helped some.

Penegal:
I updated the code; now it changes display of highway shields and highway, railway and waterway texts by, for the record, removing some duplicated labels. I used a value of 600 for the `*repeat-distance` because, AFAIK:
1. the value is in pixels, and
2. screens below 800*600 are unlikely, so a value of 600 should still display at least a label at any view, on any reasonable screen resolution.

I chose this value because it removes some duplicates while not removing too much, but it can be changed if the rendering is still too cluttered, or, conversely, if it hides too much labels.

Before/after comparisons below:
image
image

image
image

image
image

image
image

image
image

kocio-pl:
Works good also in [my test case](https://www.openstreetmap.org/#map=15/52.2787/21.0370):

nzddjb_2
mxczfgfi

kocio-pl:
Looks to me like resolving also #1956.

kocio-pl:
[Another case](https://www.openstreetmap.org/#map=17/52.21703/21.00723) - unlike shields, I would be less radical with labels, could you test some smaller values?:

Before
78c9qvcj

After
wdjto8hr

matthijsmelissen:
Anything useful in #2189?

kocio-pl:
Only shield distance used - I would say 400 works better than 600, since roads are not always long and some repetitiveness is good for readability:

[Cemetery example](https://www.openstreetmap.org/#map=15/52.2787/21.0370)
Before
2jv1kksc

400
rs878env

600
qi3duqta

[Motorway example](https://www.openstreetmap.org/#map=12/51.8771/19.6001)
Before
n9xjql_o

400
2hwie1jb

600
rdl45sov

kocio-pl:
[London Borough](https://www.openstreetmap.org/#map=15/51.4978/-0.0868) example - 400 shield distance gives ~4 more street names (+7/-3):

Before 
leczvc5s

After
rtd10fty

kocio-pl:
On the other hand City of London on z13 is just completely covered with shields when shield distance is 400, so this code needs deeper insight:

Before
f1caxypb

After
dpqa6jmj

Penegal:
@matthijsmelissen: I don't understand what you mean about #2189, as this is not about places; could you explain?

@kocio-pl: Well, this puzzles me, because, AFAIK and according to [the Mapnik doc](http://cartocss.readthedocs.io/en/latest/mapnik_api.html#shield-repeat-distance), the changes I did on shields, limited to `shield-min-distance` and `shield-repeat-distance`, only apply to shields with the same content; it should not change the number of shields displayed when they all differ in content. Do you have an idea where that could come from?

matthijsmelissen:
> @matthijsmelissen: I don't understand what you mean about #2189, as this is not about places; could you explain?

This issue provides more information about `text-min-distance`.

Penegal:
@matthijsmelissen: Ok, then. I would say it does not give relevant additional information, as the current problem is about shields number, but, once this problem will have been solved, then #2189 could become relevant. That being said, the change in the meaning of `text-min-distance` is uncommon and worth noting.

kocio-pl:
I don't really know, you should probably talk with Mapnik developers about that and test different cases. @talaj, could you bring some light on this problem?

talaj:
@kocio-pl Are you referring to the https://github.com/gravitystorm/openstreetmap-carto/pull/3295#issuecomment-404328129, that there is many more shields after this pull request applied?

Btw, I've marked `shield-min-distance` deprecated in https://github.com/mapnik/mapnik-reference/pull/148.

kocio-pl:
Yes, it's interesting for me. Looks like this simple function has some strange effects.

The other interesting thing is https://github.com/gravitystorm/openstreetmap-carto/pull/3295#issuecomment-404283391 - why text distance values don't work.

IVAN-RSM:
In some cases, information may be lost to the user. Because there are areas that are no longer seen with the current code, I can not imagine with the new code...

talaj:
@kocio-pl I think I now see what is going on. What `min-distance` means depends on whether it's text or shield symbolizer and whether the `placement` option is set to `line` or not. It's a kludge. [Here](https://github.com/mapnik/mapnik/blob/142b0ad463e22e27ffdffd745844940b8434de8f/src/text/placement_finder.cpp#L424-L446) is relevant part of code in Mapnik.

Particularly `shield-min-distance` has always the same meaning as `shield-margin`. Thus it resolves collision between all objects, not only those with the same text. 

`text-min-distance` with `text-placement: line` is the same like `text-repeat-distance`.

talaj:
@Penegal Yes, the documentation should improve. I'm tracking it almost for four years already :-) https://github.com/mapnik/mapnik-reference/issues/82

kocio-pl:
Thanks, guys! Now roads look more sane:
8eguwumt

But this was a problem with shields. Road names are still a problem.

Penegal:
@kocio-pl: regarding waterways and railways, I posted some examples in the first posts –note the corresponding code did not change, so the examples are still relevant–; I'll see for differentiating main roads tomorrow if possible.

kocio-pl:
[Another example](https://www.openstreetmap.org/#map=11/32.3225/-90.1384) of shields improvement due to this code:

Before
n10o0zd6

After
zelrolh5

kocio-pl:
I test now with `text-repeat-distance: 50;` and it gives even better results (pretty nice space between labels) for more typical dual-carriage ways (my example was extreme multiple-carriage way, which is rare).

> regarding waterways and railways, I posted some examples in the first posts –note the corresponding code did not change, so the examples are still relevant–;

With PR that changes so many things on the map with a small code change it's good to make a lot of testing of different places even if you have some good value. In this case I'm not sure that 400 is good for all the objects, which are so different, so it requires testing at least 2-3 values in 3-10 locations for each type.

Please also use Kosmtik export command with exactly the same export window (you can just change the code and Kosmtik should automatically reload without the need to close or reload the browser). It helps me when I can open before/after images and just switch between them - it's easier to concentrate on important details without looking around.

Penegal:
@kocio-pl: doing tens of coordinated tests will be difficult for me: I can only give OSM dev a very limited time, like half an hour per day. I'll do what I reasonably can, but you can assume I will not have tested much different places; nevertheless, as you already test my code elsewhere, that should not be a big problem.

Penegal:
@kocio-pl: here is the version with `text-repeat-distance: 50` for railways, trunks, motorways, primary and secondary roads. I'll post relevant examples ASAP; please also note that I ammended the commit comment, which GitHub interprets as a second commit.

kocio-pl:
[Railways example in Illinois](https://www.openstreetmap.org/#map=15/41.8350/-87.7815) shows improvement due to this code:

Before
2driw63r

After
tl5wk lh

Penegal:
Some before/after comparisons, the first one being before and the second one after:

roads.mss
```diff
@@ -2830,6 +2835,7 @@ tertiary is rendered from z10 and is not included in osm_planet_roads. */
       text-clip: false;
       text-placement: line;
       text-face-name: @book-fonts;
+      text-repeat-distance: @text-repeat-distance;
```
kocio-pl: This code is not working properly in my tests. It decreases amount of text repetitions, but does it in the same way for all values, no matter what global value is set. However if you make local value, like `text-repeat-distance: 50;`, it works different way for different values, as it should be. See https://github.com/gravitystorm/openstreetmap-carto/pull/3295#issuecomment-404984750.

Proposed solutions:
- Investigate why global values works as a uniform "small repetitions" switch, while local values work as expected.
- Make all these values local (and test that they all work as expected).

roads.mss
```diff
@@ -2862,6 +2868,7 @@ tertiary is rendered from z10 and is not included in osm_planet_roads. */
       text-face-name: @book-fonts;
       text-halo-radius: @standard-halo-radius;
       text-halo-fill: @secondary-fill;
+      text-repeat-distance: @text-repeat-distance;
```
kocio-pl: This code is not working properly in my tests - see my previous comment on `text-repeat-distance` values.

roads.mss
```diff
@@ -2888,6 +2895,7 @@ tertiary is rendered from z10 and is not included in osm_planet_roads. */
       text-face-name: @book-fonts;
       text-halo-radius: @standard-halo-radius;
       text-halo-fill: @tertiary-fill;
+      text-repeat-distance: @text-repeat-distance;
```
kocio-pl: This code is not working properly in my tests - see my previous comment on `text-repeat-distance` values.

roads.mss
```diff
@@ -2906,6 +2914,7 @@ tertiary is rendered from z10 and is not included in osm_planet_roads. */
     text-halo-radius: @standard-halo-radius;
     text-halo-fill: @standard-halo-fill;
     text-face-name: @book-fonts;
+    text-repeat-distance: @text-repeat-distance;
```
kocio-pl: This code is not working properly in my tests - see my previous comment on `text-repeat-distance` values.

roads.mss
```diff
@@ -2961,6 +2970,7 @@ tertiary is rendered from z10 and is not included in osm_planet_roads. */
       [highway = 'raceway'] { text-halo-fill: @raceway-fill; }
       [highway = 'service'] { text-halo-fill: @service-fill; }
       text-face-name: @book-fonts;
+      text-repeat-distance: @text-repeat-distance;
```
kocio-pl: This code is not working properly in my tests - see my previous comment on `text-repeat-distance` values.

roads.mss
```diff
@@ -3021,6 +3031,7 @@ tertiary is rendered from z10 and is not included in osm_planet_roads. */
       text-face-name: @book-fonts;
       text-vertical-alignment: middle;
       text-dy: 5;
+      text-repeat-distance: @text-repeat-distance;
```
kocio-pl: This code is not working properly in my tests - see my previous comment on `text-repeat-distance` values.

roads.mss
```diff
@@ -3049,6 +3060,7 @@ tertiary is rendered from z10 and is not included in osm_planet_roads. */
       text-face-name: @book-fonts;
       text-vertical-alignment: middle;
       text-dy: 7;
+      text-repeat-distance: @text-repeat-distance;
```
kocio-pl: This code is not working properly in my tests - see my previous comment on `text-repeat-distance` values.

roads.mss
```diff
@@ -3173,10 +3185,10 @@ tertiary is rendered from z10 and is not included in osm_planet_roads. */
       text-spacing: 900;
       text-clip: false;
       text-placement: line;
-      text-min-distance: 18;
       text-face-name: @book-fonts;
       text-halo-radius: @standard-halo-radius;
       text-halo-fill: @standard-halo-fill;
+      text-repeat-distance: @text-repeat-distance;
```
kocio-pl: This code is not working properly in my tests - see my previous comment on `text-repeat-distance` values.

roads.mss
```diff
@@ -3195,10 +3207,10 @@ tertiary is rendered from z10 and is not included in osm_planet_roads. */
         text-spacing: 300;
         text-clip: false;
         text-placement: line;
-        text-min-distance: 18;
         text-face-name: @book-fonts;
         text-halo-radius: @standard-halo-radius;
         text-halo-fill: @standard-halo-fill;
+        text-repeat-distance: @text-repeat-distance;
```
kocio-pl: This code is not working properly in my tests - see my previous comment on `text-repeat-distance` values.

roads.mss
```diff
@@ -3224,10 +3236,10 @@ tertiary is rendered from z10 and is not included in osm_planet_roads. */
...
```
kocio-pl: This code is not working properly in my tests - see my previous comment on `text-repeat-distance` values.

roads.mss
```diff
@@ -3254,10 +3266,10 @@ tertiary is rendered from z10 and is not included in osm_planet_roads. */
       text-spacing: 900;
       text-clip: false;
       text-placement: line;
-      text-min-distance: 18;
       text-face-name: @book-fonts;
       text-halo-radius: @standard-halo-radius;
       text-halo-fill: @standard-halo-fill;
+      text-repeat-distance: @text-repeat-distance;
```
kocio-pl: This code is not working properly in my tests - see my previous comment on `text-repeat-distance` values.

kocio-pl:
I think it might be the best to make a separate PR with just a shields changes, so it could be merged soon. 

The rest is not yet ready to go and might be further made into smaller parts. Even if the code seems trivial, there are too many things it touches and it make testing hard, which might end up with unnoticed bugs.

talaj:
@kocio-pl, can you make a diff of generated Mapnik XML with and without the variable `@text-repeat-distance`.

I would also try to use different name of the variable, as `text-repeat-distance` is also the property name. That can make some difficulty to the parser.

Penegal:
@talaj: I just tried with the variable name `repeat-distance`, and the problem is still present.

@kocio-pl: I could set a local value, but I think it is better to wait if @talaj can tell the problem from the XML; this way, if there is a problem elsewhere, like in Mapnik's interpretation of the MapCSS, he could try to fix it. I will increase it to 200, anyway.

Penegal:
Current master/new rendering with local values of `text-repeat-distance` and a 200 value for waterways:



























This version should be OK for shields, waterways and residential/unclassified roads, but `text-repeat-distance: 50` seems too few for other road types. This value could be kept for trunk and motorways, which are, AFAIK, typically not very fragmented, but it should be increased for primary, secondary and tertiary roads; typically, they are more fragmented in residential areas, so the effect of `text-repeat-distance` should be more sensible there, where it would help to reduce map cluterring. @kocio-pl, what do you think about this suggestion?

For railways… well, it seems that the global value was more effective: with the local value, `text-repeat-distance` has about zero effect. Strange, but maybe @talaj could say why.

Penegal:
@kocio-pl: agreed. I'll split this code in 4 PR, link them to this one and then close this one, but it'll probably have to wait until the next week.

kocio-pl:
The critical part from the release point of view is 1. - is it still something more than 1-2 days away from now?

Penegal:
@kocio-pl: there was nothing missing, but I totally disconnect myself from the Internet on week-ends. Here comes the first PR, for shields: #3315

kocio-pl:
Solved by #3315.

talaj:
@kocio-pl, I've tried to make a diff of resulting Mapnik XML with different values of variable `@text-repeat-distance`. I've observed that variable with the same name is defined also in `water.mss`, but value of that variable is overridden by valule from  `roads.mss`. This means that variable scope is global across mss files in the project.

Except of this, variable substitution was working as expected.

kocio-pl:
Great, thanks!

One more please for you: could you add links to all the testing locations?

kocio-pl:
Oh, I don't mean that exact URL - just a link to the area including the interesting objects (like city, road or river) in the view box.

kocio-pl:
Thanks!

antifa-ev:
@Penegal What's the improvement here? As far I can see it generally leads to fewer repetitions of the street name on the street. Ain't this bad because the reader has to raster a bigger area with his eyes?

dieterdreist:
for small screens (phones) it is important that the shields are often repeated, on larger screens it looks better with fewer repetitions. There’s a trade off to make. Who is this style intended for? There is probably no clear answer, but I suggest to test on mobile as well.

Adamant36:
@dieterdreist Is there even a way to test it on mobile before implementing it first? I don't think you can run a Docker environment on Android or IOS can you?

kocio-pl:
I don't agree about what's needed on small screen.

Generic style is not meant for roads. Yes, we have them and they are important to be shown, even with classes, but shields are not needed in cities - they block names and create clutter there, so they should be less frequent. Also outside cities, they are just too crowded. Shields are very strong elements and they look like routing - and generic map is not a car map, we don't need to be routed - there are routing services for that (including OSRM or GraphHopper). Current changes look OK for me - see #3315.

Names are more delicate visually. I think we need more nuanced solution - different values for different type. For example local (residential) road labels are OK and should be not touched. They need to be dense, because they are short and should not be blank too much. But bigger (especially dual-carriage) ways are cluttered with names and are long enough to contain labels elsewhere. See #3318.

Mobile use is special. I would be happy if we could cover different uses, like for color-blinded or mobile, but they need separate solutions, because we have problem even for typical, desktop use and it's hard enough. And typical mobile use is not looking at OSM.org map, there are map apps like OSMAnd, Maps.me and many others. They are the proper solution for mobile in my opinion. They even include better routing interface included.

SomeoneElseOSM:
> Is there even a way to test it on mobile before implementing it first? I don't think you can run a Docker environment on Android or IOS can you?

Well you can point a mobile browser (or Android emulator) at a machine showing that style.  Even with Docker, you can use an ssh connection to a machine to access a port on an internal VM locally, and point an Android web browser on the same LAN at that.

dieterdreist:
2018-08-02 0:00 GMT+02:00 Adamant36 <notifications@github.com>:

> @dieterdreist <https://github.com/dieterdreist> Is there even a way to
> test it on mobile before implementing it first? I don't think you can run a
> Docker environment on Android or IOS can you?
>

you can either use real hardware and point that to a testing tileserver, or
you can use responsive testing tools from your browser (e.g. Firefox,
Chrome, Safari all offer this).

dieterdreist:
2018-08-02 10:21 GMT+02:00 SomeoneElseOSM <notifications@github.com>:

> for small screens (phones) it is important that the shields are often
> repeated
>
> That would be true if phone displays had screens that had fewer pixels on
> them than desktop PCs, but I don't believe that's actually true. A
> relatively cheap Android phone will be full HD, but a cheap laptop will
> still be 1366x768. Also it ignores the way that tiles are displayed - by
> default with Osmdroid and Leaflet on the same phone you'll get very
> different scaling and the same map style will look very different.
>

despite having a lot of pixels they scale the pages differently, because
otherwise the text would be too small. Just try it out, with less shields
you have to pan more.

kocio-pl:
Just to illustrate why shields are problem even outside the cities - German style show them stronger and there are two version of this style to compare currently (click to see the full scale):

[Before](https://tile.iosb.fraunhofer.de//#map=11/52.0681/20.4009/5)

screenshot_2018-08-03 openstreetmap tileserver fraunhofer iosb

[After](https://tile.iosb.fraunhofer.de//#map=11/52.0681/20.4009/1)

screenshot_2018-08-03 openstreetmap tileserver fraunhofer iosb 1

bongobongo:
Great work!
Has anyone described how to implement this feature in Mapnik 2.2, so it will work when using e.g. generate_tiles.py?
If someone know which file(s) to change, and what code to add/change, to make this work in Mapnik 2.2, please let me know.

SomeoneElseOSM:
@bongobongo If you're not using OSM Carto, then you may be better off asking somewhere else (perhaps https://gis.stackexchange.com ) where the question is more likely to be seen by people running raw Mapnik.

Rather you than me trying to manually edit a mapnik stylesheet manually, though :)

PR #3299, opened 2018-07-11
Code cleaning for advertising column symbol

kocio-pl:
- remove directory for advertising column since there's only one element inside and no further objects are planned for rendering

This code does not change rendering.


polarbearing:
"advertising/column.svg" is self-explanatory, "column.svg" is not, could be any column. 
Thus the file would need to be renamed accordingly.

PR #3293, opened 2018-07-04
Add icon for shop=interior_decoration

Adamant36:
This pull request adds an icon for shop=interior_decoration and closes #1406
https://www.openstreetmap.org/#map=19/37.77273/-122.42231 (node)
interior_decoration node
https://www.openstreetmap.org/#map=19/37.77735/-122.40954 (way)
interior_decoration way

PR #3309, opened 2018-07-18
Add icon for shop=massage

### LINKED ISSUE #3287: Add special icon for shop=massage
https://wiki.openstreetmap.org/wiki/Tag:shop=massage
8k uses

Icon could be a massage table? I haven't been able to design anything intuitive enough for this.

jragusa:
Fixes #3287

Render `shop=massage` from z18 using leisure-green colour

Node
https://www.openstreetmap.org/node/4992712423
massage_point

Way
https://www.openstreetmap.org/way/183708628
massage_way


kocio-pl:
After merging #3293 this code needs to be updated to resolve conflicts.

kocio-pl:
Hm, it works strange - for z17 instead of dot (which color should it be?) it shows miniaturized icon, for example:

screenshot_2018-07-19 jybq9re1 png obraz png 3155 x 3105 pikseli

jragusa:
I see why there is an small icon at z17. But before to make changes, I would like to know whether we leave a dot for z17 like other shops or we bypass rendering for z17 ?

lakedistrictOSM:
I think that we should bypass rendering at z17. z18 is good enough for massage places.

Purple would be the wrong colour and green dots aren't yet used (and also might look like trees).

jragusa:
After correction:

z17
name appears as there is no pink dot (I suppose)
massage_z17

z18
massage_z18

z19
massage_z19

kocio-pl:
I guess it's safe to render from z18. I don't have good idea for not typical colors of shops - BTW similar problem is with `shop=car_repair` (#1778).

kocio-pl:
Could you update the code once again? I plan to merge it soon.

BTW: please, add some line breaks aligned to the lines above, so the code is more readable and easier to edit.

jragusa:
Done, I also put some line breaks as suggested. 

Is it correct for you ?

kocio-pl:
Thanks, this is good!

PR #3314, opened 2018-07-22
Increase casing width of tertiary on z12

matthijsmelissen:
Before:
screen shot 2018-07-23 at 15 30 20

After:
screen shot 2018-07-23 at 15 27 13

### DIFF
--- a/roads.mss
+++ b/roads.mss
@@ -198,7 +198,7 @@
 
 @major-casing-width-z11:          0.3;
 
-@casing-width-z12:                0.1;
+@casing-width-z12:                0.3;
 @secondary-casing-width-z12:      0.3;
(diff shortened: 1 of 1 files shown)


kocio-pl:
Frankly, I don't see the difference. Using the same box would help compare them, but this change is supposed to be visible without special effort.

matthijsmelissen:
I replaced the screenshots with screenshots with the same bounding box now.

If anybody still finds it hard to see the difference, open them both in a separate tab and switch back and forth.

kocio-pl:
Sorry, I was looking at secondary... I like it more.

[Test in the rural area](https://www.openstreetmap.org/#map=12/52.1282/20.4557):
Before
cjxpu9tn

After
3fmhuzkf

PR #3315, opened 2018-07-23
Increased shield distances (part of #3295)

Penegal:
- Increasing shield distances.

Before
76790052018067_old
097138194023486_old
115652671205744_old
196660406059515_old

After
76790052018067_new
097138194023486_new
115652671205744_new
196660406059515_new

### DIFF
--- a/roads.mss
+++ b/roads.mss
@@ -281,7 +281,8 @@
 @shield-size-z18: 12;
 @shield-line-spacing-z18: -1.80; // -0.15 em
 @shield-spacing: 760;
-@shield-min-distance: 40;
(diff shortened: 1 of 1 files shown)


kocio-pl:
Done sanity check with examples from #3295 (especially London on z13) - thanks a lot, it makes shields rendering better!

When doing before/after comparison please locate them close to each other (before1/after1, before2/after2 and so on). Finding matching images makes testing harder and more error prone.

Penegal:
@kocio-pl: thanks for the warning; I'll keep the order you asked the next time.

Issue #497, opened 2014-04-21
Improve rendering of street names


matthijsmelissen:
The following issue has been moved over from [trac](https://trac.openstreetmap.org/ticket/2991). It needs more discussion as it is not clear how this should be implemented.

Mapnik currently seems to render street names once in the middle of each piece of way, with various undesirable effects such as
- [ ] names not rendering at all if the road is split up in multiple short ways (which may happen for multiple reasons, such as different relation memberships or attributes applying just to one part of the way, such as speed limits)
- [ ] names wrapping around angles at strange places when there is enough space to render the name in a straight line
- [x] names of very long roads appearing just once, making it unclear where the boundary between two differently named sections of the same road (if any) is
- [ ] the name rendering twice if the road is made up of two separate ways which blend into each other when rendering 

This could be remedied by the following:
- Merge all pieces of road if they have the same name and highway attributes and are either joined (i.e. share a common node) or blend into each other (zoom level dependent); render the name for the entire length of the combined road
- If the resulting road is very long, repeat the names along its length multiple times:
  1) at each end and
  2) optionally in between where other roads cross, to make it clear the road continues beyond the junction 
- Where possible, pick an almost straight section (defined by minimum angle at a single point and/or minimum curve radius for consecutive points) and render the name there. Failing that, consider splitting the name string up at spaces, hyphens etc. (i.e. by the same rules that apply for line breaks) and find a straight section for each piece of the name. If that doesn't help, "massage" the straightness criteria for that road until a solution is found.

kocio-pl:
It should be doable also without lua transformations.

Currently 28 907 out of 1 002 840 sidewalks have name:

https://taginfo.openstreetmap.org/tags/footway=sidewalk#combinations

geostonemarten:
@kocio-pl in fact that can be uprising because this is on discussion with French other contributors delete the names because it considers that have had rendering  for the display.

kocio-pl:
Is it also a problem for areas or only for lines? We have a `roads-area-text-name` and `paths-text-name` data layers in `project.mml` for rendering names, both include footways and could filter out these with `footway=sidewalk`.

pnorman:
We wouldn't use Lua for this, we'd just delete the CartoCSS rules for names on footway/path/cycleway.

I'd want to see more evidence about the pros **and cons** of this before deciding if it was a good idea.

geostonemarten:
[...]

PR #3318, opened 2018-07-23
Added text-repeat-distance for highway names

### LINKED ISSUE #3111: Don't render duplicate label/shield for connected ways with the same computed style

Penegal:
This PR adds `text-repeat-distance` for highways, which currently unnecessarily clutter the map on dense areas. It uses a smaller value of 50 for `highway=unclassified` and lower-importance highways, and 200 for higher-importance highways. This also achieves the split of #3295 in smaller PR as requested by @kocio-pl, and consequently closes #3111.

Before/after examples:
66666216877235_old_highway
66666216877235_new_highway
https://www.openstreetmap.org/#map=17/48.66397/6.17061

68452214932439_old_highway
68452214932439_new_highway
https://www.openstreetmap.org/#map=17/48.68110/6.14494

70195095380471_old_highway
70195095380471_new_highway
https://www.openstreetmap.org/#map=17/48.69825/6.18554

197890463702336_old_highway
197890463702336_new_highway
https://www.openstreetmap.org/#map=16/48.1898/6.4266

Edit: added links to test locations.

### DIFF
--- a/roads.mss
+++ b/roads.mss
@@ -286,8 +286,8 @@
 @shield-font: @book-fonts;
 @shield-clip: false;
 
-@major-highway-text-repeat-distance: 500;
-@minor-highway-text-repeat-distance: 100;
+@major-highway-text-repeat-distance: 50;
+@minor-highway-text-repeat-distance: 10;
 
 @railway-text-repeat-distance: 200;
 
--- a/roads.mss
+++ b/roads.mss
@@ -286,8 +286,8 @@
 @shield-font: @book-fonts;
 @shield-clip: false;
 
-@major-highways-text-repeat-distance: 50;
-@minor-highways-text-repeat-distance: 200;
+@major-highways-text-repeat-distance: 500;
+@minor-highways-text-repeat-distance: 100;
 
 @railway-text-repeat-distance: 200;
 
@@ -2945,6 +2945,7 @@ tertiary is rendered from z10 and is not included in osm_planet_roads. */
       text-halo-fill: @residential-fill;
       text-face-name: @book-fonts;
       text-repeat-distance: @minor-highways-text-repeat-distance;
+      [highway = 'unclassified'] { text-repeat-distance: @major-highways-text-repeat-distance;}
     }
     [zoom >= 16] {
       text-size: 9;
@@ -2974,7 +2975,7 @@ tertiary is rendered from z10 and is not included in osm_planet_roads. */
       [highway = 'raceway'] { text-halo-fill: @raceway-fill; }
       [highway = 'service'] { text-halo-fill: @service-fill; }
       text-face-name: @book-fonts;
-      text-repeat-distance: @minor-highways-text-repeat-distance;
+      text-repeat-distance: @major-highways-text-repeat-distance;
     }
     [zoom >= 17] {
       text-size: 11;
@@ -2993,7 +2994,7 @@ tertiary is rendered from z10 and is not included in osm_planet_roads. */
       text-clip: false;
       text-placement: line;
       text-halo-radius: @standard-halo-radius;
-      [highway = 'living_street'] { text-halo-fill: @living-street-fill; }
+      [highway = 'living_street'] { text-halo-fill: @living-street-fill; text-repeat-distance: @major-highways-text-repeat-distance;}
       [highway = 'pedestrian'] { text-halo-fill: @pedestrian-fill; }
       text-face-name: @book-fonts;
       text-repeat-distance: @minor-highways-text-repeat-distance;
@@ -3036,7 +3037,7 @@ tertiary is rendered from z10 and is not included in osm_planet_roads. */
       text-face-name: @book-fonts;
       text-vertical-alignment: middle;
       text-dy: 5;
-      text-repeat-distance: @minor-highways-text-repeat-distance;
+      text-repeat-distance: @major-highways-text-repeat-distance;
     }
     [zoom >= 16] {
       text-size: 9;
@@ -3065,7 +3066,8 @@ tertiary is rendered from z10 and is not included in osm_planet_roads. */
       text-face-name: @book-fonts;
       text-vertical-alignment: middle;
       text-dy: 7;
-      text-repeat-distance: @minor-highways-text-repeat-distance;
+      text-repeat-distance: @major-highways-text-repeat-distance;
+      [highway = 'steps'] { text-repeat-distance: @minor-highways-text-repeat-distance; }
     }
     [zoom >= 17] {
       text-size: 11;
--- a/roads.mss
+++ b/roads.mss
@@ -2798,11 +2798,11 @@ tertiary is rendered from z10 and is not included in osm_planet_roads. */
 
       text-fill: #000;
       text-face-name: @book-fonts;
+      text-min-distance: 40;
       text-halo-radius: 2;
       text-halo-fill: @standard-halo-fill;
       text-spacing: 760;
       text-clip: false;
-      text-repeat-distance: @minor-highways-text-repeat-distance;
     }
   }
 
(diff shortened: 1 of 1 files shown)


kocio-pl:
You need to fix the `roads.mss` conflict now.

kocio-pl:
I did not ask for any change for residential roads - they look worse (incomplete) with this code:

[Example place](https://www.openstreetmap.org/#map=16/52.2846/21.0256)

Before
nqzfysln

7salmij4

Penegal:
Ok, I'll fix the residential roads with the conflict; what about the 200 value for more important roads?

Penegal:
Reverted the `highway=residential` code to the previous version.

kocio-pl:
I will review this. What value would you propose for major roads?

I think every road type change needs separate consideration, at least one test rendering, because you never know and roads are important element of the map.

Penegal:
I'll do some tests to find a correct value. At first glance, I would say 600 as I remember I tested this value, but, since then, you pointed some elements to care about that I did not care about back then, so I'll see what the tests suggest.

BTW, good remark about testing the rendering of each road type; I would probably not have tested that.

kocio-pl:
Great!

Good testing pattern is trying not only the value you think should be right, but also something "too low" and "too high". This way I was able to detect "switch" behavior in the previous code (every global value had exactly the same effect), it also hints if bigger or smaller value might be better than first guess.

Penegal:
OK, this version could do it: I played with examples to find good values. It does not significantly alter `highway=residential` rendering, while still dropping the most obvious duplicates. Here comes before/after comparisons, with URLs and, if relevant, notes on the displayed items:

0546266776136_old_highway
0546266776136_new_highway
https://www.openstreetmap.org/#map=18/48.05304/6.94675

1178506309438_old_highway
1178506309438_new_highway
https://www.openstreetmap.org/#map=18/48.05304/6.94675 (`highway=pedestrian`)

18756636033321_old_highway
18756636033321_new_highway
https://www.openstreetmap.org/#map=17/48.18681/6.39958

19950665733754_old_highway
19950665733754_new_highway
https://www.openstreetmap.org/#map=16/48.1936/6.4205

31004462157791_old_highway
31004462157791_new_highway
https://www.openstreetmap.org/#map=17/48.30746/6.02281

35071732086331_old_highway
35071732086331_new_highway
https://www.openstreetmap.org/#map=19/49.35006/6.12876

51697390179033_old_highway
51697390179033_new_highway
https://www.openstreetmap.org/#map=17/48.51451/6.64924

65525992609672_old_highway
65525992609672_new_highway
https://www.openstreetmap.org/#map=18/48.65392/6.13826 (these are `highway=unclassified`, not `highway=residential`)

66553552378001_old_highway
66553552378001_new_highway
https://www.openstreetmap.org/#map=18/48.66386/6.17092 (along the _Boulevard de l’Europe_, `highway=living_street`)

68437340134169_old_highway
68437340134169_new_highway
https://www.openstreetmap.org/#map=18/48.68220/6.17323

68446548347802_old_highway
68446548347802_new_highway
https://www.openstreetmap.org/#map=17/48.68156/6.21596

69079041492814_old_highway
69079041492814_new_highway
https://www.openstreetmap.org/#map=18/48.69006/6.17514

69847594340491_old_highway
69847594340491_new_highway
https://www.openstreetmap.org/#map=19/48.69787/6.15558

903235780840376_old_highway
903235780840376_new_highway
https://www.openstreetmap.org/#map=15/48.8924/6.1029

kocio-pl:
Thanks, it will take me some time to review and test and I'm not sure if this PR should be maybe cut into separate PRs to make discussion more focused and easier to follow. What do you think about it?

Penegal:
I currently use one PR because the current code uses only 2 different values for `text-repeat-distance`; if the number of different values increases, or if there are still numerous issues, I'd fragment it, but I don't think it would be relevant if there are only one or two issues left. Are you OK with that?

kocio-pl:
OK, we might try testing case by case in this PR.

Penegal:
@kocio-pl: were you able to test this?

kocio-pl:
I plan to test it soon, however piece by piece, not all at once.

matthijsmelissen:
@kocio-pl What is the status of this PR? If I read it correctly, it is still in your hands?

kocio-pl:
I think these values 500 for major and 100 for minor are way too big, I would go with something 10x smaller. It's enough to avoid 2 carriage roads having the same labels in the same place, but nothing more. For smaller ones I don't like to make any special changes, they are basically OK for me.

Example of major roads - 2-carriage and multiple carriage way (extreme case)

current code

26apn9kz

distance 500

ee0caeys

distance 50

tfb9gqcu

Example of minor roads - multiple carriage way (extreme case)

current code

dddxxavs

distance 100

t2qup65n

distance 10

w95i6d59

kocio-pl:
[Example of minor roads](https://www.openstreetmap.org/#map=17/52.23714/21.00333) - 2 carriage way

current code

xaolnfix

distance 100

w_cawz85

distance 10

tet5jrm1

kocio-pl:
[Example of a more typical major (2 carriage) and minor (1 carriage, small segments) roads](https://www.openstreetmap.org/#map=16/52.2845/21.0281)

current code

b2xc5uwh

distances: major 500 + minor 100

dcmhbyiv

distances: major 50 + minor 10

bs0novnf

kocio-pl:
@Penegal Could you update this PR? I would make some more tests to check if nothing is broken and if it will pass, I would merge it soon.

roads.mss
```diff
@@ -2984,9 +2994,10 @@ tertiary is rendered from z10 and is not included in osm_planet_roads. */
       text-clip: false;
       text-placement: line;
       text-halo-radius: @standard-halo-radius;
-      [highway = 'living_street'] { text-halo-fill: @living-street-fill; }
+      [highway = 'living_street'] { text-halo-fill: @living-street-fill; text-repeat-distance: @major-highway-text-repeat-distance;}
```
kocio-pl: I think that text-repeat-distance for living street should be in separate line, so the brackets should create a small inner block.

kocio-pl: Could you fix this remaining issue? Than it would ready to merge.

Penegal: @kocio-pl : done.

Penegal:
@kocio-pl : the code should now be as you asked; warn me if there are some problems left.

kocio-pl:
It looks good now, thanks a lot! I hope this will make street experience more pleasant - at least i'm happy with the output I tested.

kocio-pl:
Related to #497.

geostonemarten:
Is it possible to ignore the label creation for footway if it contain footway=sidewalk because there is also the same name on major highway.
image

https://www.openstreetmap.org/#map=17/50.52778/1.58588

SomeoneElseOSM:
> Is it possible to ignore the label creation for footway if it contain footway=sidewalk because there is also the same name on major highway.

That sounds like a question unrelated to this pull request, but something that achieves a similar effect is technically possible.  Example from a related style that uses similar software components: https://github.com/SomeoneElseOSM/SomeoneElse-style/blob/master/style.lua#L974 .

Issue #3241, opened 2018-05-20
Add icon for leisure=outdoor_seating


Tomasz-W:
https://wiki.openstreetmap.org/wiki/Tag:leisure=outdoor_seating
2k uses.

Icon proposal:

![leisure outdoor_seating](https://user-images.githubusercontent.com/25656654/40282267-07596694-5c6d-11e8-809d-a975c08b6f5d.png)

polarbearing:
Good idea. The umbrella is a bit dominant compared to chairs+table. What about dropping the middle pole?

geozeisig:
`leisure=outdoor_seating` belongs to a restaurant, cafe etc. and should be tagged together with this one with `outdoor_seating=yes` (72k times). You can create icon for restaurants with and without outdoor seating, but I think that's too complicated.

Tomasz-W:
@geozeisig Please read Wiki page of this feature. `Leisure=outdoor_seating` is for permanent seating areas (eg. build on wooden terrace or equiped with heavy umbrellas) and `outdoor_seating=yes` is for small ones, which are bring outside only on sunny days.

polarbearing:
Indeed I have seen them on separate objects, the restaurant node/area inside the building and the node or area for the outdoor area, naturally, outside.

SomeoneElseOSM:
@Tomasz-W The wiki does not say that (see https://wiki.openstreetmap.org/wiki/Key:outdoor_seating) and I have not seen mappers do what you suggest.

@geozeisig It wouldn't be complicated to split restaurants into categories in lua and have the Carto code only differ by icon (that's how I've done the pubs and fast food at https://map.atownsend.org.uk/maps/map/map.html#zoom=17&lat=-24.988106&lon=135.17649 for example).  Whether this style wants to make more use of lua like this is a different issue of course.

eehpcm:
I'm late to this party, and everyone has gone home.  I only just realized this icon had been incorporated because I'd mapped several outdoor seating areas and noticed one of them now had an icon.  Cool.  Unfortunately, it's a very misleading icon for what I've mapped.

**None** of the areas I've mapped have umbrellas or parasols or roofs or canopies.  No protection from rain of any sort (which is amusing, given how often it rains here).  I see that since I mapped them the wiki has added a mention of the undocumented and little-used complementary tag "weather_protection" and although it mentions all sorts of possibilities, it doesn't mention "weather_protection=no", which I think probably ought to be the default.

[...]

Issue #3069, opened 2018-02-12
Add rendering for amenity=bicycle_repair_station


Tomasz-W:
A public-access stations equipped usually with air pump, screwdriver, wrenches, allen keys etc. Very useful feature.
https://wiki.openstreetmap.org/wiki/Tag:amenity%3Dbicycle_repair_station

About 2800 uses.
https://taginfo.openstreetmap.org/tags/amenity=bicycle_repair_station

Icon proposal:
![amenity bicycle_repair_station](https://user-images.githubusercontent.com/25656654/36116934-5299d82c-1038-11e8-96a7-c9a78560f166.png)

kocio-pl:
It's a duplicate of #1353, let's discuss it there.

kocio-pl:
Sorry, it seems that we had 2 PRs which were rejected then (#1353 and #1864), but no issue ticket before.

We already have the icon, so making a PR would be rather easy now:
https://github.com/gmgeo/osmic/blob/master/shop/repair-bicycle-14.svg

Tomasz-W:
Notice that @nebulon42 icon indlude single-sided wrench, and it could be confused with the key from amenity=bicycle_rental.
https://github.com/gravitystorm/openstreetmap-carto/blob/master/symbols/rental_bicycle.svg

I don't know is brown color (from PR test renderings) right for this feature. Bicycle parkings and rentals are blue.

nebulon42:
Rental has the thicker side on the left while repair has the thicker side on the right. This should work against confusion.

Tomasz-W:
Anyway, I would like to see a test rendering with my icon, so I'm putting Gist link:
https://gist.github.com/Tomasz-W/b4641535717e8b36ae8fcae4acae95a3

polarbearing:
Voila. Still quite similar on the first glance. I wonder if we might turn the repair icon 90 deg, so the bike comes in repair position, and the spanner easier to distinguish from the key. Or the bike upside down. 
![z18_bike_rent](https://user-images.githubusercontent.com/8266355/36122667-778dc5a0-104a-11e8-8de3-2e82bc2d4770.png) ![z18_bike_repair](https://user-images.githubusercontent.com/8266355/36122691-910a9e54-104a-11e8-8c9e-6f3797dd991e.png)

Tomasz-W:
For comparsion:
[...]

PR #3323, opened 2018-07-30
Add icon for shop=paint

### LINKED ISSUE #3288: Add  special icon for shop=paint
https://wiki.openstreetmap.org/wiki/Tag:shop=paint
7.6k uses

Icon could be a paint brush or a roller?
https://thenounproject.com/search/?q=paintbrush&i=105797
https://thenounproject.com/search/?q=paintbrush&i=466004
https://thenounproject.com/search/?q=paintbrush&i=273917

Adamant36:
Adds icon for shop=paint which closes #3288 
https://www.openstreetmap.org/#map=19/37.77712/-122.40170 (Way)
paint way
https://www.openstreetmap.org/#map=19/37.76481/-122.41495 (Node)
paint node

### DIFF
--- a/amenity-points.mss
+++ b/amenity-points.mss
@@ -1027,6 +1027,10 @@
       }
     }
 
+    [shop = 'paint'][zoom >= 18] {
+      marker-file: url('symbols/shop/paint.svg');
+    }
+
     [shop = 'shoes'][zoom >= 18] {
       marker-file: url('symbols/shop/shoes.svg');
     }
@@ -2508,6 +2512,7 @@
   [feature = 'shop_jewellery'],
   [feature = 'shop_laundry'],
   [feature = 'shop_chemist'],
+  [feature = 'shop_paint'],
   [feature = 'shop_toys'],
   [feature = 'shop_travel_agency'],
   [feature = 'shop_seafood'],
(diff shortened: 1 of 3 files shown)


Tomasz-W:
Thanks for another PR! :)
PS. Auto-close keyword in first post is broken because of "issue" word between "closes" and issue link.

Adamant36:
Yep :) I wish there were more coders around helping you out. They must be off enjoying their summers or something. 

Thanks for the heads up on the closing thing. I keep doing that.

Tomasz-W:
Speaking about summer, it would be good to have some features on map before it ends and the weather gets cold. I mean features like https://github.com/gravitystorm/openstreetmap-carto/issues/3069 or https://github.com/gravitystorm/openstreetmap-carto/issues/3241. So if it's not a problem for you, and you don't have other plans, I think these tags (and other leisure/ sport tags later) should have a prority to add rendering for.

meased:
The handle is not pixel aligned (half a pixel off). Also, the file should be saved as an optimized SVG (there's a file type selector at the bottom of the Inkscape save dialog).

Otherwise looks good!

Adamant36:
@Tomasz-W, sure I can do that. I was actually thinking about that already. It seemed like #3069 wasn't settled yet though. With the positioning thing and what not. So you think its good to go? And would I be updating the car repair icon to match it also?

@meased Thanks for the heads up. I thought it looked a little off.

Tomasz-W:
Updated icon: https://gist.github.com/Tomasz-W/245e2151c51541d6265f540bc490eafe

@meased I'm using "clean SVG" format which is also allowed.

Adamant36:
Updated.

kocio-pl:
Thanks, it works good. You might try using git branches in the future instead of just using master.

Adamant36:
I did that before and it led to a bunch of problems. I might try it again though once I get better at using Git.

PR #3327, opened 2018-08-03
Changing farmland and societal amenities colors

### LINKED ISSUE #1991: New landuse=farmland rendering color too close of amenity=hospital

kocio-pl:
Closes #2821

- Changing the colors of farmland and societal amenities taken from @imagico fork (https://github.com/imagico/osm-carto-alternative-colors, changes described in details in http://blog.imagico.de/more-new-colors/ )

In short, this makes basically a color swap between farmland and societal amenities, in effect:
- farmlands are greenish, which is coherent with other vegetation areas
- farmlands are lighter, which is better than going back to dark, since they tend to be quite big
- societal amenities are reddish, which is more coherent with some other landuses 
- societal amenities are darker, which makes them easier to spot, since they tend to be not too big and eclipsed with some other landcovers (like grass for example)

Farmland color comparison between different forks:
ac_farmland

Test rendering with links to the example places (click images to see the full scale):

[Poland](https://www.openstreetmap.org/#map=8/52.736/19.355)

Before
uotnvcnv

After
kn8gjoy

[Comparison with some other green areas](https://www.openstreetmap.org/#map=15/52.1051/21.2017)

Before
k kwnwdp

After
ldzsnmnp

[Color change for schools and hospitals](https://www.openstreetmap.org/#map=16/52.2863/20.9559)

Before
rlibezwh

After
5kur7zui

[Color switch between social facility and farmlands ](https://www.openstreetmap.org/#map=16/52.2989/20.8403)

Before
vupsxxit

After
yxiys3ro

### DIFF
--- a/landcover.mss
+++ b/landcover.mss
@@ -20,8 +20,8 @@
 @commercial-line: #d1b2b0;  // Lch(75,12,25)
 @industrial: #ebdbe8;       // Lch(89,9,330) (Also used for railway)
 @industrial-line: #c6b3c3;  // Lch(75,11,330) (Also used for railway-line)
-@farmland: #fbecd7;         // Lch(94,12,80)
-@farmland-line: #d6c4ab;    // Lch(80,15,80)
+@farmland: #eef0d5;         // Lch(94,14,112)
+@farmland-line: #c7c9ae;    // Lch(80,14,112)
 @farmyard: #f5dcba;         // Lch(89,20,80)
 @farmyard-line: #d1b48c;    // Lch(75,25,80)
 
@@ -51,7 +51,7 @@
 @power: darken(@industrial, 5%);
 @power-line: darken(@industrial-line, 5%);
 @sand: #f5e9c6;
-@societal_amenities: #f0f0d8;
+@societal_amenities: #fbecd7;   // Lch(94,12,80)
 @tourism: #734a08;
 @quarry: #c5c3c3;
 @military: #f55;


meased:
This makes a lot of sense. I like how this takes schools out of yellowish/green and puts it closer to amenity brown (it almost looks as if school text could actually be @amenity_brown instead of a darkened version of the landcover).

kocio-pl:
Thanks for reminding of labels. I'm not sure where to change the color of school name, but farmland name will look like this:

[Named farmland](https://www.openstreetmap.org/way/62144021#map=16/52.2311/20.6967)

Before
6z_dt_fh

After
6gjx35li

polarbearing:
Looks encouraging to me on my first impression.

kocio-pl:
This is how it looks when the whole country is covered with different greens, including farmlands - look at Romania z8 (click to see full scale):

4otepl7l

kocio-pl:
It also works better for me at large areas with sand (deserts), like in south Israel:

Before
kiexbdko

After
d6eukqq0

IVAN-RSM:
I like it more like it is now 😅

landcover.mss
```diff
@@ -20,8 +20,8 @@
 @commercial-line: #d1b2b0;  // Lch(75,12,25)
 @industrial: #ebdbe8;       // Lch(89,9,330) (Also used for railway)
 @industrial-line: #c6b3c3;  // Lch(75,11,330) (Also used for railway-line)
-@farmland: #fbecd7;         // Lch(94,12,80)
-@farmland-line: #d6c4ab;    // Lch(80,15,80)
+@farmland: #eef0d5;         // Lch(94,14,112)
+@farmland-line: #c7c9ae;    // Lch(80,14,112)
```
pnorman: We generally go for a bit more chroma on the lines.

kocio-pl: What's your proposition then?

westnordost:
This makes it harder though to tell apart farmland from other landuse in the countryside. After all, the different colors are there for distinction, not to mimik the colors of how it would look on a satellite image. "Freeing up" a color for use in urban parts removes this color for use in rural parts.

kocio-pl:
Why do you think it makes it harder exactly? The colors are still different and the luminocity is similar. I didn't think about aerial images - there's no single color for farmland (different phases of growth, different species). There's no freeing up any color, there's only a color swap.

kocio-pl:
Adding more chroma for farmland outline:

@imagico version
bpxfrwbq

chroma +3 version
rxvk9sm8

hubgitti:
The current color reminds me of the warm golden color of cereal, that's why I prefer it in favour of the new suggested color.

But, maybe the current version is too "bright", and I understand that there is a will or need for change, but the new suggestion looks too much like "military-green" and the map already has quite many greenish color styles.

kocio-pl:
1. The problem of choosing colors is not so hard when the map style is simple and has only few elements, but this style is rich and using some coherent conventions (like shades of green for vegetation) is a life saver. This is the most important argument for me. It's better for me to see the clear difference between field and desert (see the Israel example).

2. What cereal and why not field of corn or rice for example? This would be dark green or blue (because of water) probably.

3. Military green seems to be quite different shade (#4B5320), see https://en.wikipedia.org/wiki/Shades_of_green#Army_green

4. I'm surprised, but with so many greens I have no problem recognizing them all, and the context helps too.

hubgitti:
1) I think it is easy to distinguish farmland and dessert areas (in context)
2) Many crop plants turn "golden" when they are ready to harvest, also rice can:
https://commons.wikimedia.org/wiki/Rice?uselang=de
3) You are right. (There are so many military greens on this world, that invalidates my point).
4) You are right, one could take this point as pro and also as contra. For me it is a contra.

I'm still not convinced of your suggestion, but still - maybe the current "farmland gold" is too bright, I'm not sure about that...

kocio-pl:
So what would be your suggestion to make it less bright?

turnsole80:
Just a thought: how does the new colour look beside `landuse=farmyard`? It might be worth changing that colour too

kocio-pl:
Thanks for noticing that. It looks OK for me - but maybe @imagico has some other idea?

[Example:](https://www.openstreetmap.org/#map=18/52.13661/20.56230)
4ga21o38

turnsole80:
I was worried it would look wrong, but that's not as bad as I was expecting

hubgitti:
About the "brightness" of the current color:
When I started mapping landuses (residential, meadow, farmland, farmyard) I somehow tried to avoid farmland because of the color (quite bright and different to others on the map).

If others feel or felt the same, this could lead to a "tagging for the renderer" and that would just be bad...

After starting the thought, that the current farmland color is something like " ready to harvest crop plants" I perfectly learned to live with it, and I like it now.

I find the other color styles (old-osm, fr, de) better then the new one. I can understand the want for a change (especially if the current color would lead to a "tagging for the renderer"). I'm really bad in deciding such things - but I still oppose a greenish color style...

kocio-pl:
It's interesting what you say - I have never tried tagging farmland even if it was (and still is) not nice for me and I wanted to get rid of it.

I also think that using similar color shades for vegetation makes sense. Moving things toward consistency is important, and the best example lately is playground color change: without #2249 inventing #3330 would be rather hard, even if nobody expected that we might start rendering ice rinks one day. 

I'm cautious and don't trust making the big system at this stage of style development (which is good idea when starting new project), but not only it makes farmland vegetation-like, also it makes societal amenities closer to other landuses.

hubgitti:
Another thought why not to use a "greenish" color style: Areas with "landuse=farmland" are "economically utilized/exploited". So this areas also have something in common with commercial/industrial areas.

Adamant36:
@kocio-pl was there ever a discussion about what kind of landuses get special colors versus a pattern? I notice orchards along with vineyards are both dark green and have a pattern. Maybe farmland is to general for a pattern, but it seems a little off to make it a completely different color then its sub types. 

I can see there being a lot more crop types with patterns in the future. Maybe how they would play into the general farmland tag as a cohesive whole is something to consider in tandem with this issue, if not how the current ones do at least. I almost agree with @hubgitti that maybe there should be a economically utilized/exploited scheme or at least one for the category of "agriculturally" used land. Otherwise it risks way over complicating things.

polarbearing:
Distinguishing crop types was discussed, maybe here or on the tagging list. The problem is that they can change quite often for the same field. 

I browsed the thread here again and still like the change. As for the farm _yard_, maybe that could be even lightened up a bit in the new context? Could we see a village with a school, farm yard and farm land?

kocio-pl:
> was there ever a discussion about what kind of landuses get special colors versus a pattern?

I don't think so. Some landuses were here from the beginning (in pre-2012 XML style probably).

> I notice orchards along with vineyards are both dark green and have a pattern. Maybe farmland is to general for a pattern, but it seems a little off to make it a completely different color then its sub types.

Well, being green(ish) makes it closer to them than currently.

> I almost agree with @hubgitti that maybe there should be a economically utilized/exploited scheme or at least one for the category of "agriculturally" used land. Otherwise it risks way over complicating things.

Forest is for "managed" tree areas and it might be exploited, yet it would not work for me to try to depict this aspect.

There is a general guideline to use regular patterns when something is curated and irregular ones for more natural objects, but it'd be hard to enforce it universally, it just provides a hint.

kocio-pl:
> Could we see a village with a school, farm yard and farm land?

I could not find them too close to each other, but here is [an example](https://www.openstreetmap.org/#map=16/51.9696/20.5152):

r8eanlul

dieterdreist:
sent from a phone

> On 25. Aug 2018, at 12:46, polarbearing <notifications@github.com> wrote:
> 
> Distinguishing crop types was discussed, maybe here or on the tagging list. The problem is that they can change quite often for the same field.


I guess it depends how we define the classes. We would not distinguish between wheat, barley or maybe maize, but between those and rice, or hops, yes (for example). Similar to how orchards are distinguished from farmland. Meadows and farmland on the other hand do alternate in areas I know of, and there is indeed the problem you have to change it “frequently”

polarbearing:
Thanks @kocio-pl - I drop my idea to change the farm yard, it works as it is in the larger context.

landcover.mss
kocio-pl: What do you think about https://github.com/gravitystorm/openstreetmap-carto/pull/3327#issuecomment-414098780? I can't see the difference.

dktue:
I also think of a golden cereal field when I see farmland -- that's why I like the current color way better than a greenish style.

kocio-pl:
If we have only the basic set of tags to render, it would be different, I would probably think about colors case by case (however this shade is too intensive anyway for me). This map is much more complex and more systematic approach is important.

polarbearing:
> golden cereal field

Please remember it is a map and not a satellite image painting, and even cereals would start as green plants before being ripe.

hubgitti:
Sorry, intuitive color styles clearly is not "satellite image painting".

But more important to me: Farmland should be clearly distinguished from other landuses like meadow, or natural areas like grassland, ...

Think of the huge farmland areas in Germany, Ukraine, ...

kocio-pl:
Please look at the comparison with other green areas then (from https://github.com/gravitystorm/openstreetmap-carto/pull/3327#issue-206078903) - I have no problem with seeing the difference with any other green here - do you?:

ldzsnmnp

polarbearing:
The new colour is even more intuitive to me, as it is a greenish tone for vegetation, and I can clearly distinguish it from the other greens used for grassy/woody things.

hubgitti:
I clearly can see the difference, but that is not the point for me. Farmland is - to me - something completely different to grassland, meadow, ...

Farming can be like quite "industrial" work.

hubgitti:
That is probably really a bad idea, what's about your greenish style plus a symbol like forests have?

hubgitti:
Guys, I will stop opposing your idea (I personally still do not agree with you at this point) but the progress on the map by each version is amazing. The people contributing here seem to know quite well what they are doing, that's why I could be wrong with my opinion here...

kocio-pl:
It's hard for me to answer, I'd like to hear more farmland mappers to know how they think about it. I think that maybe outlines could be useful for showing real farmlands instead of big landuse-like clusters, which are just easier to draw.

dieterdreist:
sent from a phone

> On 28. Aug 2018, at 03:25, jeisenbe <notifications@github.com> wrote:
> 
> Re: the outlines. Is it necessary to render an outline for farmland areas? In many places around my town, I've cut up larger areas of farmland so as to avoid making complex multipolygon relations (for example, when a hamlet or wood is surrounded by farmland on all sides),
> 


We’ve started mapping these rural areas far more detailed than this around here, meaning that the single object is small, and many of them are needed to surround a hamlet or wood.

Multipolygon relations are never needed, but they are helpful to reduce overlapping ways by reusing the polygon borders.

The structure of divisions is generally interesting on its own, that’s why the outlines are nice to have.

>  there are only footpaths in the area, no roads.
> 


not even tracks?

polarbearing:
The outlines should encourage people to map according to the real boundaries of a field an not make arbitrary cuts. Then it can help to see e.g. if the village has lots of small fields, or huge monocultures.

jeisenbe:
dieterdreist, my town is surrounded by farming areas where people only use walking for transportation. There are a handful of roads, but they soon end. However, I've now checked and it looks ok to split a field along a footpath, as long as I am careful to follow the same line, so I think I have a solution.
It's good to know that other people like having small field divisions. I was feeling like it must be a mistake, since sometimes JOSM suggests it is a problem, but I will not worry too much about it now.

kocio-pl:
My current plan is to merge it in a few days without changing outline color, because @pnorman did not answer, I see no difference and 2 people gave their thumbs down.

kocio-pl:
Thanks, all!

You might be interested in the changes @imagico made recently with rendering crop in his fork (which is next step following introducing this farmland color):

http://blog.imagico.de/more-on-pattern-use-in-maps/

If anybody wants to discuss it, please open new ticket.

dktue:
I like it, too! :-)

jengelh:
> (farmyard image)

Now a farmyard surrounded by farmlands looks way out of place... if it could be made the same hue again that'd be great.

kocio-pl:
We've found that it's quite OK and there are other problems to solve, so it's not changed. Do you have a proposition which colors we could test? That would help a lot.

jengelh:
"same hue" was the proposition, i.e. rgb#f1f5ba. Comparison (poor man's replace with a paint prog; not truly rendered):
farm

PR #3330, opened 2018-08-05
Render leisure=ice_rink

### LINKED ISSUE #796: Add rendering for leisure=ice_rink
Some leisure objects are not rendered, such as:
- leisure=ice_rink  (https://trac.openstreetmap.org/ticket/3482)
- leisure=dog_park (https://trac.openstreetmap.org/ticket/2482)

jragusa:
Fixes #796

Render leisure=ice_rink with glacier background and pitch colour text

https://www.openstreetmap.org/way/87188308

Before
ice_rink_before

After
ice_rink_name_green


kocio-pl:
It looks to me that we should also add the outline. Code for glaciers is here:

https://github.com/gravitystorm/openstreetmap-carto/blob/d8fc3c72e649e5ecfd016ecd396be99f4c288c77/shapefiles.mss#L39-L53

jragusa:
It looks better, what do you think ?
ice_rink_greenline

landcover.mss
```diff
@@ -270,6 +270,16 @@
     }
   }
 
+  [feature = 'leisure_ice_rink'] {
+    [zoom >= 10] {
```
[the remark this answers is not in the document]
jragusa: Something like: `[zoom >= 10][way_pixels > 3000][is_building = 'no'] {` ?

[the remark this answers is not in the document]
kocio-pl: Yes, most probably. :)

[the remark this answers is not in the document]
jragusa: ok done :)

kocio-pl:
Did you test the code after the change? I suggest to always do it, no matter how small and innocent the change looks like. :smile: 

In this case I get error (shortened):

```
Postgis Plugin: ERROR:  column "is_building" does not exist
.
.
.
  ) AS landcover
```

Which means that landcover data layer is not aware of `is_building` ("virtual" column), because this hasn't been defined there. See for example:
https://github.com/gravitystorm/openstreetmap-carto/blob/e58cc3c2228c8680653541972a5a263a4ad70cc7/project.mml#L2020

and try to put it after this line inside landcover layer:

https://github.com/gravitystorm/openstreetmap-carto/blob/e58cc3c2228c8680653541972a5a263a4ad70cc7/project.mml#L139

This might be not enough, but that's probably something to start digging. If you will have any problems with that, just tell here.

HolgerJeromin:
We should probably do something with
seasonal=winter/yes

kocio-pl:
Thanks. What do you propose? Would simply reusing water type of rendering work for you?

jragusa:
I removed the `is_building` condition and buildings are not affected by the PR:
https://www.openstreetmap.org/way/246399017
ice_rink_building

kocio-pl:
I believe this is not a good solution and proper solution (selecting only non-buildings for area+outline and maybe labels for all the cases) should be used instead.

kocio-pl:
@jragusa Are you willing to deploy proposed solutions in this PR?

kocio-pl:
OK, no need to hurry - I prefer people learning how to make a proper code in this style, so we could resolve more complex problems than just adding icons for shops. Do you need some help with this?

jragusa:
@kocio-pl I still have problem with the condition is_building. It's used for text but when I tried to define it for surface, I got an error from `projet.mml`. I will investigate this on Thursday.

kocio-pl:
Do you have any news about this problem?

project.mml
```diff
@@ -138,6 +138,7 @@ Layer:
               ('railway_' || (CASE WHEN railway = 'station' THEN railway ELSE NULL END)) AS railway,
               CASE WHEN religion IN ('christian', 'jewish', 'muslim') THEN religion ELSE 'INT-generic'::text END AS religion,
               way_area/NULLIF(!pixel_width!::real*!pixel_height!::real,0) AS way_pixels
+              CASE WHEN building = 'no' OR building IS NULL THEN 'no' ELSE 'yes' END AS is_building -- always no with the where conditions
```
jragusa: Once I had this line, I get the following error: 

```
Postgis Plugin: ERROR:  syntax error at or near "CASE"
LINE 28:       CASE WHEN building = 'no' OR building IS NULL THEN 'no...
```

jragusa:
@kocio-pl this PR is ready to merge

project.mml
```diff
@@ -139,6 +139,7 @@ Layer:
               ('railway_' || (CASE WHEN railway = 'station' THEN railway ELSE NULL END)) AS railway,
               CASE WHEN religion IN ('christian', 'jewish', 'muslim') THEN religion ELSE 'INT-generic'::text END AS religion,
               way_area/NULLIF(!pixel_width!::real*!pixel_height!::real,0) AS way_pixels,
+              CASE WHEN building = 'no' OR building IS NULL THEN 'no' ELSE 'yes' END AS is_building -- always no with the where conditions
```
kocio-pl: Comma is missing before the comment, so it does not compile, it should be:

```sql               
CASE WHEN building = 'no' OR building IS NULL THEN 'no' ELSE 'yes' END AS is_building, -- always no with the where conditions 
```

jragusa: strange, it works fine without and I get a syntax error in Postgis when I add the comma.

kocio-pl: Very strange  indeed... Could anyone else check it? @Adamant36, @jeisenbe?

kocio-pl: @matkoniecz @sommerluk Maybe you could check it too?

jeisenbe:
I fetched this branch, and I also was notified of a syntax error. Adding
the comma fixed the problem, as suspected.

project.mml
jragusa: I resolved the problem

kocio-pl:
Thanks, it works! And now we have a 4 year old issue fixed... :smile:

Issue #3175, opened 2018-04-12
Add icon for leisure=bird_hide


Hjart:
https://wiki.openstreetmap.org/wiki/Tag:leisure%3Dbird_hide

~2700 uses so far, but people appears to tag these things in a variety of other ways (primarily as view point) to make them render, as well as name them 'Bird hide' or 'Bird watching tower'.

JOSM has used https://wiki.openstreetmap.org/wiki/File:Binocs-both.svg as an icon since 2016

turnsole80:
There is also the more generic `leisure=wildlife_hide` as well, which has a couple of hundred uses.

kocio-pl:
Any idea how to visualize it? We're showing hunting stand, which is partially similar, but has different overall goal.

Hjart:
The birdwatchers primary tool is his binoculars. Would it make sense to create a head with binoculars peering out over a box or screen?

polarbearing:
We need to distinguish them from [planespotters](http://www.planespotters.net), however, who are a different species. So, some reference to birds is needed.

Hjart:
How about something like https://commons.wikimedia.org/wiki/File:2008-07-25_Geese_over_01.svg (easily recognizable as a bird) then? I guess visualizing both a bird and binoculars (or telescope) in the same icon is stretching it a bit too much?

kocio-pl:
How would it look like in 14px?

polarbearing:
For consistency, the [hunting stand](https://github.com/gravitystorm/openstreetmap-carto/blob/master/symbols/hunting_stand.svg) icon visualises the facility, not the animal.

lakedistrictOSM:
Just binoculars would probably work best. 

In England it's common for bird hides to have names so name=* should be rendered too. Icon and label in leisure-green?

> We need to distinguish them from planespotters, however, who are a different species. So, some reference to birds is needed.

I don't think we do. Is anyone actually mapping "planespotting hides"? The location should give context anyway - nature reserve vs airport etc

dieterdreist:
sent from a phone

> On 15. Apr 2018, at 14:30, lakedistrictOSM <notifications@github.com> wrote:
> 
> We need to distinguish them from planespotters, however, who are a different species. So, some reference to birds is needed.
> 
> I don't think we do. Is anyone actually mapping "planespotting hides"?
> 


[...]

PR #3340, opened 2018-08-12
Add icon for leisure=bird_hide

Adamant36:
Adds icon for leisure=bird_hide closing #3175 
https://www.openstreetmap.org/#map=19/47.43837/-122.47189 (Way)
bird hide way
https://www.openstreetmap.org/#map=19/47.81981/-119.97127 (Node)
bird hide node

### DIFF
--- a/amenity-points.mss
+++ b/amenity-points.mss
@@ -1268,6 +1268,13 @@
      marker-clip: false;
    }
 
+  [feature = 'leisure_bird_hide'][zoom >= 17] {
+     marker-file: url('symbols/bird_hide.svg');
+     marker-fill: @leisure-green;
+     marker-placement: interior;
+     marker-clip: false;
+   }
+
   // Slipway tagging on points - slipway on lines is defined later
   [feature = 'leisure_slipway'][zoom >= 17] {
     marker-file: url('symbols/transport_slipway.p.20.svg');
@@ -1673,6 +1680,7 @@
   [feature = 'tourism_picnic_site'][zoom >= 17],
   [feature = 'leisure_bowling_alley'][zoom >= 17],
   [feature = 'leisure_beach_resort'][zoom >= 17],
+  [feature = 'leisure_bird_hide'][zoom >= 17],
   [feature = 'leisure_outdoor_seating'][zoom >= 19],
   [feature = 'leisure_picnic_table'][zoom >= 17] {
     text-name: "[name]";
@@ -1683,6 +1691,7 @@
     [feature = 'tourism_picnic_site'],
     [feature = 'leisure_outdoor_seating'],
     [feature = 'leisure_bowling_alley'],
+    [feature = 'leisure_bird_hide'],
     [feature = 'leisure_beach_resort'] {
       text-fill: @leisure-green;
     }
--- a/project.mml
+++ b/project.mml
@@ -1487,7 +1487,7 @@ Layer:
               'office' || CASE WHEN tags->'office' IN ('no', 'vacant', 'closed', 'disused', 'empty') OR (tags->'office') IS NULL THEN NULL ELSE '' END,
               'leisure_' || CASE WHEN leisure IN ('water_park', 'playground', 'miniature_golf', 'golf_course', 'picnic_table',
                                                   'fitness_centre', 'fitness_station', 'firepit', 'sauna', 'beach_resort', 
-                                                  'bowling_alley', 'outdoor_seating') THEN leisure ELSE NULL END,
+                                                  'bowling_alley', 'outdoor_seating', 'bird_hide') THEN leisure ELSE NULL END,
               'man_made_' || CASE WHEN (man_made IN ('mast', 'tower', 'water_tower', 'lighthouse', 'windmill', 'obelisk', 'communications_tower', 'chimney') 
                                         AND (tags->'location' NOT IN ('roof', 'rooftop') OR (tags->'location') IS NULL)) THEN man_made ELSE NULL END,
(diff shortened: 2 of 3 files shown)


kocio-pl:
Thanks, works as expected.

lakedistrictOSM:
Does this PR add rendering for the name of the bird hide too?

kocio-pl:
Yes, it does:

apd1ufqw

polarbearing:
Where is the reference to birds in this icon?
It looks like man_made=binoculars, and does not indicate any wildlife.

polarbearing:
Which or how many people exactly?

> binoculars work fine and can probably work in other similar situations also

Then bird_hide.svg is a misnomer for the file.

Adamant36:
It's your complaint. I'm sure you can figure it out. Its not on me to prove your point for you, if you even have one. 

Notice I said "eventually." Its not like the name can't be changed, eventually, when its used in other cases. 

If you have that much of  a problem with it, why not create a new issue for it or go comment in the old one so other people can be involved in the discussion? Or why didn't you voice your objection during the three days the pr sat here? I would of been happy to change it if there was reason to. That goes for all my PRs. Feel free to bring up any issues you have with them when they are waiting to be merged. Although, things should be mostly figured out in the original issue in my opinion.

dieterdreist:
sent from a phone

> On 15. Aug 2018, at 12:21, Adamant36 <notifications@github.com> wrote:
> 
> Therefore binoculars work fine and can probably work in other similar situations also eventually.


IMHO it can be misleading, there are also public binoculars, for which this icon made much more sense

Adamant36:
@polarbearing Tomasz-W Added it to his "Code needed to render prepared icons" list like a month ago. Know one commented on there that it wasn't resolved or should be removed. The one with binoculars was the only one where an .SVG was provided. 


In a comment on April 15th both lakedistrictOSM and Kioco-pl said binoculars alone would work best. 

Dktue reacted with a thumbs up to Tomasz-W post of the binoculars only icon. 

Then a little after that you thumbs upped your own anti binoculars only comment along with Hjart, which is neither here nor there I guess, but a bit odd. 

dieterdreist said "maybe also an eye could be used instead of the binoculars." Which seems more like spit balling then a definitive "I'm against this." 

Tomasz-W later says "I think it's obvious that 'binoculars-only' version is more readable. I think leisure-green binoculars would be enough for this feature." That seems pretty not including an animal to me. 

Right under that you once again thumbs upped your own comment of the binocular-bird icon, probably to make it look more evenly split then it was. Eigenwillig thumbs downed it. So that user can be added to the anti bird-binocular crowd. 

Then Hjart said "Yes, a 'binoculars-only' certainly is more readable." Which seems to be more pro binoculars-only then not to me. Turnsole80 just said the goose icon looks good and that was before any of the other icons where proposed, including the binocular-only one. So you can't really say he was against it. Ultimately, readability matters and you where the only one against it being binoculars only.

Also, I don't have to participate in a discussion to write the code for it or do a PR on it, and I find it hilarious you thumbs upped your own comments.

polarbearing:
@Adamant36 - TL;DR
I did not criticise you for submitting a PR, that is fine and everybody's right. I said it was insufficiently discussed before merging.

kocio-pl:
Well, that's probably directed at my work... :smile: 

I'm also volunteer, but I care for deadlines. It proved me to be the most effective single planning tool with such a fuzzy environment. I try to keep the flow, because if something is active, people tend to be more active too, which helps the project to be healthy teamwork. Sometimes I try to wait longer, when I'm not sure or when I expect controversies, but that was not the case when I was expecting this. I was ready to merge it immediately, but wanted to wait at least a day or two.

If more discussion is needed, I'm open for this. Nobody likes reverts, but it's also a tool one can use if the problem is too big and most of the time some update is enough.

With regard to viewing devices - I think that we can simply render them brown (like other amenities) if we ever care for them and even use a telescope metaphor if that's not enough. Also don't forget about the context - they will be most likely located in some touristic place, while bird hide is most probably located somewhere in the outdoor, like forest for example.

Adamant36:
@polarbearing OK. It seemed like you where saying I should of been involved in the original discussion in order to do the PR or to have an opinion about it, since you brought up my lack of participation in the issue. I might of just read it wrong though.

polarbearing:
>  I should of been involved in the original discussion in order to do the PR

No I did not say that, I said "submitting a PR ... is fine and everybody's right"

> the context - they will be most likely located in some touristic place

One of the render examples above is at the coastline, so this could be seen as whale watching, ship spotting, or viewpoint. 

A resolution could be to open up the code to include ```leisure=wildlife_hide``` to make clear it is more general than just about birds. 

The wiki on these tags recommends combinations with ```man_made=tower``` - how does the tower icon interact with the watching icon?

kocio-pl:
> A resolution could be to open up the code to include leisure=wildlife_hide to make clear it is more general than just about birds.

I like this idea. Usage is low, but I like more systematic approach, not just counting case by case, and the wiki page is here:

https://wiki.openstreetmap.org/wiki/Tag%3Aleisure%3Dwildlife_hide

> The wiki on these tags recommends combinations with man_made=tower - how does the tower icon interact with the watching icon?

I don't know what is preferred rendering, but tower prevails. 

We have many more specific cases when we would like to define special tag priority, but this is general problem and I'm not sure there is sane way of tweaking SQL.

Adamant36:
I was actually planning on adding leisure=wildlife_hide in thr original PR for this reason, but decided not to because of the low numbers and lack of discussion about it in the original issue. I think it would be a good addition though despite that.

Issue #3298, opened 2018-07-11
Good first issues [meta-ticket]


Tomasz-W:
### Regularly updated task list:
**single icons:**
- [x] amenity=boat_rental https://github.com/gravitystorm/openstreetmap-carto/issues/3263 -> done! https://github.com/gravitystorm/openstreetmap-carto/pull/3300
- [x] amenity=bureau_de_change https://github.com/gravitystorm/openstreetmap-carto/issues/3185 -> done! https://github.com/gravitystorm/openstreetmap-carto/pull/3301
- [x] shop=charity; second_hand https://github.com/gravitystorm/openstreetmap-carto/issues/1869 -> done! https://github.com/gravitystorm/openstreetmap-carto/pull/3302
- [x] shop=interior_decoration https://github.com/gravitystorm/openstreetmap-carto/issues/1406 -> done! https://github.com/gravitystorm/openstreetmap-carto/pull/3293
- [x] shop=video https://github.com/gravitystorm/openstreetmap-carto/issues/3116 -> done! https://github.com/gravitystorm/openstreetmap-carto/pull/3313
- [x] leisure=bowling_alley https://github.com/gravitystorm/openstreetmap-carto/issues/3297 -> done! https://github.com/gravitystorm/openstreetmap-carto/pull/3311
- [x] shop=paint https://github.com/gravitystorm/openstreetmap-carto/issues/3288 -> done! https://github.com/gravitystorm/openstreetmap-carto/pull/3323
- [x] shop=massage https://github.com/gravitystorm/openstreetmap-carto/issues/3287 -> done! https://github.com/gravitystorm/openstreetmap-carto/pull/3309
- [x] leisure=outdoor_seating https://github.com/gravitystorm/openstreetmap-carto/issues/3241 -> done! https://github.com/gravitystorm/openstreetmap-carto/pull/3328
- [x] leisure=bird_hide https://github.com/gravitystorm/openstreetmap-carto/issues/3175 -> done! https://github.com/gravitystorm/openstreetmap-carto/pull/3340
- [x] shop=bookmaker https://github.com/gravitystorm/openstreetmap-carto/issues/3186 -> done! https://github.com/gravitystorm/openstreetmap-carto/pull/3345
- [x] shop=trade https://github.com/gravitystorm/openstreetmap-carto/issues/3291 -> done! https://github.com/gravitystorm/openstreetmap-carto/pull/3361
[...]

PR #3346, opened 2018-08-15
Water slide

### LINKED ISSUE #3335: Add rendering for water slides (attraction=water_slide)
### Expected behavior

Should be rendered as a path (blueish?) and an appropriate icon if mapped as a way.
Should be rendered as just the icon if mapped as a node.
The icon could be a man on a sloping line.

### Actual behavior

Not rendered at all.

Tag description:
https://wiki.openstreetmap.org/wiki/Tag:attraction%3Dwater_slide
It "suggests" to make it waterway=stream in addition, to force it being rendered, but is that a valid approach?

Example occurrence:
https://www.openstreetmap.org/#map=19/47.52249/7.56949

johsin18:
Fixes #3335 

- Render attraction=water_slide similar to a (water) stream
- Render a marker for water slide
- Render the name of the water slide

This is more a proof of goodwill, probably both the rendering and the code have to be further improved, as I'm a complete newbie to this.
In particular, the marker is hidden unter the slide, how can I fix that?  However, I'm not so sure the marker makes sense, anyway.

https://www.openstreetmap.org/way/353735844

Before
waterslide_before

After
waterslide19
waterslide18
waterslide17
waterslide16
waterslide15

### DIFF
--- a/amenity-points.mss
+++ b/amenity-points.mss
@@ -2901,6 +2901,26 @@
     marker-file: url('symbols/transport_slipway.p.20.svg');
     marker-fill: @transportation-icon;
   }
+
+  [feature = 'attraction_water_slide'] {
+    [zoom >= 16] {
+      [zoom >= 17] {
+        bridgecasing/line-color: black;
+        bridgecasing/line-join: round;
+        bridgecasing/line-smooth: 1;
+        bridgecasing/line-width: 1.25;
+        [zoom >= 18] { bridgecasing/line-width: 2.5; }
+        [zoom >= 19] { bridgecasing/line-width: 5; }
+      }
+      line-color: @pitch;
+      line-join: round;
+      line-cap: round;
+      line-smooth: 1;
+      line-width: 1;
+      [zoom >= 18] { line-width: 2; }
+      [zoom >= 19] { line-width: 4; }
+    }
+  }
 }
 
 #trees [zoom >= 16] {
--- a/amenity-points.mss
+++ b/amenity-points.mss
@@ -2919,6 +2919,19 @@
       line-width: 1;
       [zoom >= 18] { line-width: 2; }
       [zoom >= 19] { line-width: 4; }
+
+      [zoom >= 19] {
+        text-name: "[name]";
+        text-size: 10;
+        text-face-name: @oblique-fonts;
+        text-fill: darken(@pitch, 40%);
+        text-halo-radius: @standard-halo-radius;
+        text-halo-fill: @standard-halo-fill;
+        text-placement: line;
+        text-vertical-alignment: middle;
+        text-repeat-distance: @waterway-text-repeat-distance;
+        text-dy: 8;
+      }
     }
   }
 }
--- a/project.mml
+++ b/project.mml
@@ -1563,14 +1563,17 @@ Layer:
       table: |-
         (SELECT
           way,
+          name,
           COALESCE(
            'highway_' || CASE WHEN tags @> 'ford=>yes' OR tags @> 'ford=>stepping_stones' THEN 'ford' ELSE NULL END,
-           'leisure_' || CASE WHEN leisure IN ('slipway') THEN leisure ELSE NULL END
+           'leisure_' || CASE WHEN leisure IN ('slipway') THEN leisure ELSE NULL END,
+           'attraction_' || CASE WHEN tags @> 'attraction=>water_slide' THEN 'water_slide' ELSE NULL END
             ) AS feature
           FROM planet_osm_line
           -- The upcoming where clause is needed for performance only, as the CASE statements would end up doing the equivalent filtering
           WHERE tags @> 'ford=>yes' OR tags @> 'ford=>stepping_stones'
             OR leisure IN ('slipway')
+            OR tags @> 'attraction=>water_slide'
         ) AS amenity_line
     properties:
       minzoom: 16


water.mss
```diff
@@ -244,6 +244,23 @@
     }
   }
 
+  [attraction = 'water_slide'] {
+    [zoom >= 15] {
+      [zoom >= 17] {
+        bridgecasing/line-color: saturate(darken(@water-color, 40%), 30%);
```
johsin18: Actually, I'm not sure what the prefix "bridgecasing" references to here.  Is it really needed?

project.mml
```diff
@@ -306,15 +306,17 @@ Layer:
...
               OR tags->'seasonal' IN ('yes', 'spring', 'summer', 'autumn', 'winter', 'wet_season', 'dry_season')
               THEN 'yes' ELSE 'no' END AS int_intermittent,
             CASE WHEN tunnel IN ('yes', 'culvert') THEN 'yes' ELSE 'no' END AS int_tunnel,
             'no' AS bridge
           FROM planet_osm_line
-          WHERE waterway IN ('river', 'canal', 'stream', 'drain', 'ditch', 'wadi')
-            AND (bridge IS NULL OR bridge NOT IN ('yes', 'aqueduct'))
+          WHERE (waterway IN ('river', 'canal', 'stream', 'drain', 'ditch', 'wadi')
+                 AND (bridge IS NULL OR bridge NOT IN ('yes', 'aqueduct')))
+            OR attraction IN ('water_slide')
```
johsin18: Is it okay after all to use the "water-lines" for the water slides?

polarbearing:
I wonder why you place the symbol at the beginning of the slide? Usually here, symbols are in the middle of the feature.

lakedistrictOSM:
Thanks for this PR.

I'm not sure what the icon is supposed to be but it's probably unnecessary anyway. 

What zoom level are you starting from? If I'm reading the code correctly z15 might be too early.

sommerluk:
Hi @johsin18 

Thanks for making a Pull Request. You’re welcome!

While I didn’t do neither a cartographic nor a code review (and don’t have time to do it currently), as a first impression I feel this is a feature for starting rather with zoom level 18 or 19.

Best regards and thanks again for the PR!

Tomasz-W:
@johsin18 My remarks:
* icon (?) is unnecessary here
* for me this way looks just like a stream flowing out of water body (more like a natural feature than a human-made one), can you try with `leisure=pitch` green fill and dark-blue outline?

Thanks for this PR and good luck! :)

HolgerJeromin:
perhaps we need a bridge like border. In fact a slide is kind of a bridge

johsin18:
Thanks for all your comments.

* I will drop the icon, nobody (including me) likes it ;-)
* about the zoom level: let's start at 17, otherwise it pops up too suddenly IMHO.  After all, the feature is quite large usually
* about the color: I will try with some that is more consistent with man-made features
* @Tomasz-W Once water slides are actually rendered, people will have a motivation to map the corresponding steps.  Right now, it would look like a stairway to heaven ;-)  And actually, this problem will stay for zoom levels where the slide is not yet shown, but just the steps (so should we go to 16 then maybe?)
* @HolgerJeromin How should it become more similar to a bridge border?  Different outline color?  Some white "casing" in between (how would I achieve that)?

Can you please comment on the code (and my questions added as comments there)?
Is the code in the right place?  Does it not cause efficiency problems maybe?

johsin18:
So I have made some of the proposed changes, simplified the code, and moved everything to amentiy-points.mss, which seems to make more sense.  It is rendered from zoom level 16 now, due to the usual size of such slides.
waterslide19
waterslide18
waterslide17
waterslide16

I show screenshots on are more complex example:
https://www.openstreetmap.org/#map=19/47.53818/7.62098

How could I make the junction at the very top smoother?  It's common to assemble as slide from multiple ways in multiple layers, as you would have self-intersecting ways otherwise.  Should/could they be combined with some kind of relation?

openstreetmap-carto.style
```diff
@@ -14,6 +14,7 @@ node,way   admin_level  text         linear
 node,way   aerialway    text         linear
 node,way   aeroway      text         polygon
 node,way   amenity      text         polygon
+node,way   attraction   text         polygon
```
kocio-pl: I'm not sure if and how this works. Typically we don't change this file, since it would require database reload, which is very time consuming and is rarely done on OSMF servers.

kocio-pl:
I'm not sure what works better - water with casing or pitch with casing.

openstreetmap-carto.style
kocio-pl: In cases when the column is not defined when creating database, we use the hstore column with construction `tags->'foobar'`, see for example this very short PR with `icao` and `iata` value check: https://github.com/gravitystorm/openstreetmap-carto/pull/2674/files

johsin18:
Ok I have removed the change to openstreetmap-carto.style.

Tomasz-W:
@kocio-pl Notice that function of water slides is to allow people to fastly flow down, not to just move water from one place to another. It makes water slides closer to running track than a typical water stream.

matthijsmelissen:
Thanks for your proposal, @johsin18! However, I still think this is too specific for a general purpose map, and I don't think we should accept this PR.

matthijsmelissen:
It's also not clear to me what the usecase of rendering waterslides is.

dieterdreist:
sent from a phone

> On 16. Aug 2018, at 17:05, Matthijs Melissen <notifications@github.com> wrote:
> 
> It's also not clear to me what the usecase of rendering waterslides is.


details like this help to make the map look better (ideally) and help people to orient theirselves on the map. Certainly this isn’t aiming at routing.

kocio-pl:
After some time (a year?) contemplating use case idea, I'm starting to think this is not the best tool for making general map using grass root data.

For me it's rather about reflecting the reality, especially "on the ground" truth (with some exceptions like tunnels), secondary aim is the quick check what is the state of database (rough feedback loop on completeness and quality). Maybe I would see use cases as a third level - if we have some problem with too much data, we can think how to filter them. 

In case of waterslides (1) this is what's visible on the ground, (2) it helps to check if the tagging is there and (3) it probably never competes with some other objects in such places, so no need to filter it other than using scaling rule (initial zoom level should be proportional to the typical size).

dieterdreist:
waterslides also often quite significant in the landscape/surroundings as they have to be higher if they should be longer

HolgerJeromin:
> How should it become more similar to a bridge border? Different outline color? Some white "casing" in between

All bridges have a black stroke. Even water bridges:

https://github.com/gravitystorm/openstreetmap-carto/blob/14e9349e2d3237a375b3fef47ff37572239a0594/water.mss#L134

meased:
Use cases :

- Orientation (just about everything has this usecase).
- Discovering details about an attraction.
    - Not all usecases are using the map to assist with travel, sometimes people just want to browse around to see what's there, which is in itself a usecase.
- Providing feed back for people mapping water parks.
    - Water parks are much more likely to be micro mapped.
- Shows off the richness of OSM.

Really, pretty close to the usecases of rendering swimming pools or pitches.

I like blue better. Swimming pools aren't for holding water, they're for swimming in. Why aren't they pitch green?

Also, I think
```css
line-join: round;
line-cap: round;
```
would look good.

kocio-pl:
So maybe the blue line with a black (bridge-like) outline?

BTW: it's amazing to me that this PR attracts so much attention. I don't understand why, but I'm happy with that.

polarbearing:
> this PR attracts so much attention. I don't understand why

Because this summer is so hot?

I prefer the pitch colour. As said before, the prime function is to slide. Water is just the lubricant, not the medium as in the pool. The pitch colour would also apply to dry slides where people sit on a rag.

johsin18:
Ok so I changed the casing color to black, as there seems to be a consensus on that.
For the interior color (pitch or poll) I don't care too much.
Adding line-cap is not a good idea, as it puts a break between two connected slide segments. Also, a water slide is always open at both ends, so it does not match reality well.
There is still a narrow black line between the two segments, I'm afraid, is there a way to avoid it completely?

johsin18:
Ok so I changed the casing color to black, as there seems to be a consensus on that.
For the interior color (pitch or poll) I don't care too much.
Adding line-cap is not a good idea, as it puts a break between two connected slide segments. Also, a water slide is always open at both ends, so it does not match reality well.
There is still a narrow line between the two segments, is there a way to avoid it completely?

kocio-pl:
Could you render this version? It makes a lot of sense to make visualization of every change.

Tomasz-W:
Is it possible to not render black outline on ways ends, but only on its sides? Water slides are open at start point and end point, so lack of outline there would be better visualisation of reality.

johsin18:
line-cap: round;
(just for the inner) helps here, but I'm not sure it's a clean solution:
waterslide19

polarbearing:
keeping the ends open is also important to have no lines between segments (probably the segments are ther for different layer tags). Thus this is better now in this aspect also.

johsin18:
So would please somebody merge this PR?  In the upcoming weeks I will have little time to further work on this.

kocio-pl:
Other examples with new rendering:

https://www.openstreetmap.org/#map=19/47.42822/9.66942
jfjlb_z6

https://www.openstreetmap.org/way/591252087#map=19/54.11311/22.93798
njjflbr5

amenity-points.mss
```diff
@@ -2901,6 +2901,39 @@
...
+      [zoom >= 19] {
+        text-name: "[name]";
+        text-size: 10;
+        text-face-name: @oblique-fonts;
+        text-fill: darken(@pitch, 40%);
+        text-halo-radius: @standard-halo-radius;
+        text-halo-fill: @standard-halo-fill;
+        text-placement: line;
+        text-vertical-alignment: middle;
+        text-repeat-distance: @waterway-text-repeat-distance;
```
kocio-pl: I feel this is not needed for such short features.

amenity-points.mss
```diff
@@ -2901,6 +2901,39 @@
     marker-file: url('symbols/transport_slipway.p.20.svg');
     marker-fill: @transportation-icon;
   }
+
+  [feature = 'attraction_water_slide'] {
+    [zoom >= 16] {
+      [zoom >= 17] {
```
kocio-pl: z17+ condition is needless, because z16+ covers wider case. I would go with just z17+, as these elements are not that big.

polarbearing:
looks visually good to me

amenity-points.mss
johsin18: I tried to remove it, but then I got the name twice even on quite short slides.

kocio-pl: I see - surprising side effect. :smile:

Tomasz-W:
@johsin18 Big thanks for this PR! If you will have some time and will in future, please look at task lists in https://github.com/gravitystorm/openstreetmap-carto/issues/3298#issue-340348239 and https://github.com/gravitystorm/openstreetmap-carto/issues/3298#issuecomment-405039682. Maybe there will be something which you could help us with code writing :)

kocio-pl:
I've just found a place with many water slides mistagged as streams:

https://www.openstreetmap.org/#map=19/37.97477/-122.05156

polarbearing:
We should wait til the release is through, so we can tell him how much better they would have looked if he hadn't tagged for the renderer so blatantly ;-)

kocio-pl:
He was smarter and used double tagging... :smile: 

aakr8tzq

PR #3353, opened 2018-08-20
Fixing country labels on z4 [WIP]

### LINKED ISSUE #2851: Country label problems

kocio-pl:
Fixes #2851

- Removing upper country size limits on z3-z4, to allow showing missing Canada and Russia labels

@rrzefox Could you test this fix on your server?

### DIFF
--- a/placenames.mss
+++ b/placenames.mss
@@ -4,7 +4,7 @@
 @state-labels: desaturate(darken(@admin-boundaries, 5%), 20%);
 
 .country {
-  [zoom >= 3][way_pixels > 1000][way_pixels < 360000] {
+  [zoom >= 3][way_pixels > 1000] {
     text-name: "[name]";
     text-size: 10;
     text-wrap-width: 35; // 3.5 em
@@ -16,7 +16,7 @@
       text-line-spacing: -1.4; // -0.13 em
       text-margin: 7.7; // 0.7 em
     }
-    [zoom >= 5] {
+    [zoom >= 5][way_pixels < 360000] {
       text-size: 12;
       text-wrap-width: 45; // 3.8 em
       text-line-spacing: -1.2; // -0.10 em


rrzefox:
this has been deployed and z3-z4 rerendered.

kocio-pl:
Thanks! 

Testing link (positions might be different, since OSMF servers use new Mapnik with different labels placing algorithm):
http://bl.ocks.org/matthijsmelissen/raw/af7a602c222dbf1ff1a2c0d84ed755b7/#4.00/64.96/-93.61

All the missing country labels on z4 are now visible (the names of islands indicate that we're really on z4):

Canada
screenshot_2018-08-20 screenshot

Greenland
screenshot_2018-08-20 screenshot 1

Russia
screenshot_2018-08-20 screenshot 2

matthijsmelissen:
To be honest, I prefer the old rendering. There was a reason not to render labels on such huge areas: it is really hard to see what area the label belongs to. Especially because the label is so small compared to the country. And if we're already zoomed in so far, I'm not even sure if users are still interested in the country name.

kocio-pl:
Yes, I remember your position and I understand it.

For me such a large area without even a single label lacks consistency. The zoom-in logic for me is following: a blank page -> country names level -> states names level. Without this fix there's a blank page between "country level" (z3) and "states level" (z5) and for me it's disturbing.

kocio-pl:
Quoting from https://github.com/gravitystorm/openstreetmap-carto/issues/2851:

> The reason is that by the time you're zoomed in so far, you should already know which country you're looking at, and smaller detail is more important to be rendered.

I agree with this sentence. Yes, I know what country it is. My problem is that on z4 the are not smaller details yet (they are rendered from z5) and this gap is strange for me. Only 2-3 countries behave like that and it seems like special exception: we have plenty of space for rendering, but we abstain from it.

kocio-pl:
Another thing that I just saw: without country labels cities like Moscow or Ottawa lack some context for that name (there are only country boundaries).

kocio-pl:
I see also similar problem with some Russian states: they should be visible from z5, but they are too big already, so their name is never shown, even if we have plenty of space. And there are also capitals inside them which lack some context. Example - Krasnoyarsk Krai:

https://www.openstreetmap.org/relation/190090#map=5/70.656/99.031

kocio-pl:
My current code has a side effect that the name of a "big" country is rendered even on z4+. Could somebody help me to fix it?

matthijsmelissen:
Also on z5 for small countries, the condition [zoom >= 3][way_pixels > 1000] is true.

Maybe you should start with something like [zoom >= 3][zoom < 5][way_pixels > 1000], [zoom >= 5][way_pixels < 360000].

rrzefox:
the current version of this PR has been deployed and Z2-Z6 rerendered.

PR #3345, opened 2018-08-15
Add icon for shop=bookmaker

Adamant36:
Adds icon for shop=bookmaker closes #3186
https://www.openstreetmap.org/#map=19/40.87865/-73.59622 (node)
bookmaker node
https://www.openstreetmap.org/#map=19/40.65692/-74.11743 (way)
bookmaker way

Issue #1186, opened 2014-12-31
add orthodox cross for religion=christianity icon (depending on denomination)


matkoniecz:
see https://github.com/gravitystorm/openstreetmap-carto/pull/1185#issuecomment-68447380 for a possible icon

At least denomination=orthodox,  denomination=greek_catholic and denomination=russian_orthodox may use new icon.

EDIT, thanks to @mboeringa:

As there is a wide variety of COUNTRY_orthodox following query may be better (but it will catch also typos in country names):

religion IN ('christian') AND (denomination IN ('armenian_apostolic', 'greek_catholic') OR denomination LIKE '%orthodox%' OR denomination LIKE '%Orthodox%')

mboeringa:
@mkoniecz, this seems a very incomplete list if compared to tagInfo. Taginfo shows that much of the orthodox denominations, seem to be declared country specific:

https://taginfo.openstreetmap.org/keys/denomination#values

An undoubtedly incomplete list (based on browsing through the first 20 or so taginfo pages for "denomination" and "orthodox" in the name of the denomination, and reading through some general Wikipedia pages about Christian denominations and orthodoxy):

'armenian_apostolic','bulgarian_orthodox','coptic_orthodox','ethiopian_orthodox','ethiopian_orthodox_tewahedo','georgian_orthodox','greek_orthodox','macedonian_orthodox','orthodox','romanian_orthodox','russian_orthodox','serbian_orthodox','syriac_orthodox','ukranian_orthodox'

Due to the diversity of these orthodox denominations, and the values in the OSM database, it might be better to try to catch them all using a more generic SQL statement like:

religion IN ('christian') AND (denomination IN ('armenian_apostolic') OR denomination LIKE '%orthodox%' OR denomination LIKE '%Orthodox%')

And shouldn't **greek_catholic** be **greek_orthodox**?

math1985:
> And shouldn't greek_catholic be greek_orthodox?

Doesn't seem like it: http://en.wikipedia.org/wiki/Ukrainian_Greek_Catholic_Church

mboeringa:
> Doesn't seem like it: http://en.wikipedia.org/wiki/Greek_Byzantine_Catholic_Church

[...]

PR #3356, opened 2018-08-21
Add icons for memorial subtags

### LINKED ISSUE #3088: Distinguish different memorial types
Currently we render all of memorials by just two icons, which are not every time match real object shape well. I would like to propose more icons to distinguich certain types of memorials.
https://wiki.openstreetmap.org/wiki/Key:memorial

My proposition:
* memorial=war_memorial
z_memorial ('memorial' icon)
* memorial=plaque
z_plaque ('plaque' icon)
* memorial=blue_plaque
z_plaque ('plaque' icon)
* memorial=statue
memorial statue (**new**)
* memorial=bust
memorial bust (**new**)
* memorial=stele
 z_memorial ('memorial' icon)
* memorial=stolperstein
z_plaque ('plaque' icon)
* memorial=stone
memorial stone (**new**)
* memorial=obelisk 
memorial obelisk (**new**)

For comparsion:
* memorial obelisk (obelisk icon proposal)
* z_monument  (historic=monument current icon)

* memorial statue (statue icon proposal)
* z_artwork (tourism=artwork current icon)

Potentially. we could think about change tourism=artwork icon. Some fast-made ideas (lower block set to 50% transparency):

1) tourism artwork
2) tourism artwork2
3) tourism artwork3

jragusa:
Fixes #3088

* Add icons for memorial subtags `bust`, `stone` and `man_made=obelisk`.
* Associate `blue_plaque` <del>and `stolperstein`</del> to `plaque`
* <del>Replace generic icon of `memorial`</del>
* Move all related icons to the `historic` folder
* Use a new icon `statue` for both `memorial=statue` and `tourism=artwork` + `artwork_type=statue`

generic `memorial` tag
https://www.openstreetmap.org/node/5051735103
monument_left

`memorial=bust`
https://www.openstreetmap.org/node/5501258534
memorial_bust

`memorial=statue`
https://www.openstreetmap.org/node/3822475193
memorial_statue

`memorial=stone`
https://www.openstreetmap.org/node/4987880673
memorial_stone

`man_made=obelisk`
https://www.openstreetmap.org/way/72937686
memorial_obelisk


jragusa:
@Tomasz-W  Ok there is a misunderstood from my side. I initially though we change the generic icon to the bench. So I will change it and make tests. Do you have a location for test ?

@kocio-pl Concerning the tagging scheme, I only use `memorial=stolperstein` to remain consistent with the other tags. Since memorial:type=stolperstein is more frequent, do you prefer I include both or only the `memorial:type` ? In any case, I'm investigating tomorrow the respective count for each tagging scheme.

Tomasz-W:
Example of memorial bench: https://www.openstreetmap.org/node/3648628350

kocio-pl:
We can render them in a special way both, only one of them, or don't change anything. There's no clear answer yet for such problems, so it's good to think a moment and decide. 

This style has been evolving mostly case by case, but with complex tags (very broad categories) like towers or castles we touch a new ground. I like systematic approach and having the same scheme would be nice, on the other hand so many uses make visual clutter already.

Tomasz-W:
I vote for `memorial=stolperstein`, because:
* it's better to have a systematic subtags scheme
* reason of high amount of `memorial:type=stolperstein` is some automated import in Germany (this have been said in issue discussion)
* `memorial:type=stolperstein` have a "disputed" status on Wiki (see: https://wiki.openstreetmap.org/wiki/Tag:historic%3Dmemorial#Tags_to_use_in_combination)

dieterdreist:
sent from a phone

> On 21. Aug 2018, at 21:41, Tomasz Wójcik <notifications@github.com> wrote:
> 
> I vote for memorial=stolperstein,


there are 123 uses of this tag, vs. 19k with :type

jragusa:
There is a clear predominance of `memorial` scheme. Note there is also some minor `memorial_type` tags.

tag | memorial | memorial:type
---- | ---- | ----
blue_plaque | 181 | 1
bust | 2 602 | 95
plaque | 9 293 | 559
statue | 7449 | 818
stele | 4 385 | 171
stone | 4 224 | 230
stolperstein | 129 | 19 943
war_memorial | 13 719 | 124
**total** | 45 632 | 23 627

dieterdreist:
2018-08-22 11:09 GMT+02:00 Jérémy Ragusa <notifications@github.com>:

> There is a clear predominance of memorial scheme. Note there is also some
> minor memorial_type tags.
> tag memorial memorial:type
> blue_plaque 181 1
> bust 2 602 95
> plaque 9 293 559
> statue 7449 818
> stele 4 385 171
> stone 4 224 230
> stolperstein 129 19 943
> war_memorial 13 719 124
> *total* 45 632 23 627



there is a clear split, stolperstein is almost exclusively tagged with
memorial:type. Either you convince the mappers to change the tags, or I
would suggest to render both of them the same. Deciding to render a tag
with 129 uses and not rendering a tag with the same supposed meaning with
20K uses feels like overstepping your mandate. This is an OSMF project and
the OSMF deliberately refrains from steering the tagging development. This
style should follow the community, not lead it where a few people think it
should go to.

kocio-pl:
Interesting trends:
taghistory 27

kocio-pl:
Since this is German community decision and practice that is different than all the rest, I just asked them about possible change of stolperstein tagging scheme:

https://forum.openstreetmap.org/viewtopic.php?id=63458

jragusa:
That would be a good starting point to have an agreement with the German community about this tag.

In the meantime, I have difficulties to render memorial bench. I have errors from `project.mml`. Do we have other equivalents than the combination of `amenity=bench` + `historic=memorial` ?

SomeoneElseOSM:
> Do we have other equivalents than the combination of amenity=bench + historic=memorial ?

If it helps, when I was going through historic and memorial combinations in taginfo in the UK about a month ago I didn't find any other historic bench combinations.  You can see what I did find at https://github.com/SomeoneElseOSM/SomeoneElse-style/blob/master/style.lua#L2661 .

kocio-pl:
This conflict is a direct, but otherwise unrelated consequence of merging #3345.

Initial responses from German community look encouraging and indicate support for `historic=memorial + memorial=stolpersteine` scheme. This would help to clean up `historic=memorial` scheme at last, since Stolpersteine is the main force behind `memorial`/`memorial:type` split:

taghistory 28

Next big thing would be moving `memorial=war_memorial` into some other subkey then, as this has been already agreed upon:

https://lists.openstreetmap.org/pipermail/tagging/2017-October/thread.html#33696

Tomasz-W:
Ok, so remaining changes in this PR are to:
* left current generic `historic=memorial` icon, and use bench icon only for combination with `amenity=bench`
* ignore `memorial=war_memorial` tag, and treat it just as `historic=memorial`

kocio-pl:
Could the statue icon be tuned so it's a bit more different than artwork? Maybe base could be stronger?

Tomasz-W:
Well, actualy both icons are presenting the same thing (statue), so trying to make them different from each other is doomed to fail. As I think that statue shape is not a good symbol for most of artworks (they are usually some weird, experimental constructions), I've proposed to change an artwork icon (see https://github.com/gravitystorm/openstreetmap-carto/issues/3088#issue-299937596) to avoid the problem.

kocio-pl:
Hm, artwork can be anything and looking for the right shape it became clear to me that this icon should be small and unobtrusive, so I'm not sure these propositions could work. 

On the other hand worst examples were like park with a lot of statues (https://github.com/gravitystorm/openstreetmap-carto/pull/2236#issuecomment-234687333), so maybe it might. It's a tagging problem to decide what is artwork and what is memorial (for example contemporary or ancient Greek religious figures) and it's outside the scope of styling data.

Tomasz-W:
@kocio-pl What do you think about column/ totem icon for artworks? As current icon it's also vertical, but simpler shape is giving a bigger freedom of interpretation. 

tourism artwork

@jragusa Can you test it?
Gist: https://gist.github.com/Tomasz-W/125758342433c6880afed2f3c955f2d4

PS. Current artwork icon leads to tagging for renderer, see eg. St. Peter Square in Vatican:
https://www.openstreetmap.org/#map=19/41.90222/12.45660

https://previews.123rf.com/images/marinv/marinv1509/marinv150900049/44850743-colonnades-that-surround-st-peter-s-square-in-rome-vatican-city-with-many-statues-on-top.jpg

They are not artworks, but statues of certain saints (`historic=memorial` + `memorial=statue`) , but they are tagged as `tourism=artwork` for _osm-carto_

kocio-pl:
At first sight I like the totem idea, however on our scale it looks too close to new obelisk icon. What do you think about making top wider (http://pluspng.com/img-png/png-totem-pole-traditionally-totem-poles-represented-a-group-of-individual-spirits-events-in-time-clans-or-stories-1000.png) or add more horizontal elements (like this one: https://commons.wikimedia.org/wiki/File:%22Totem%22,_Carl%C3%A9.jpg)? Typical totem wings might look like cross (https://d30y9cdsu7xlg0.cloudfront.net/png/20257-200.png), but you might get a bit creative. :smile: 

As of Vatican etc. - they were already there when I was testing different artwork icons, so they definitely were not tagged for rendering, because artwork rendering was not existing yet while memorial rendering was there already. And this is your interpretation that these figures are memorials - religious objects might not been historic in general, so it's hard to know how should they be really tagged. That's the problem I still don't know how to solve.

Tomasz-W:
* v1:

* v2:
tourism artwork

v2 added to Gist link: https://gist.github.com/Tomasz-W/125758342433c6880afed2f3c955f2d4

dieterdreist:
sent from a phone

> On 25. Aug 2018, at 12:34, Tomasz Wójcik <notifications@github.com> wrote:
> 
> PS. Current artwork icon leads to tagging for renderer, see eg. St. Peter Square in Vatican:
> https://www.openstreetmap.org/#map=19/41.90222/12.45660
> 
> https://previews.123rf.com/images/marinv/marinv1509/marinv150900049/44850743-colonnades-that-surround-st-peter-s-square-in-rome-vatican-city-with-many-statues-on-top.jpg
> 
> They are not artworks, but statues of certain saints (historic=memorial + memorial=statue) , but they are tagged as tourism=artwork for osm-carto
> 


was this irony? How would they not be artwork? Religious depictions have dominated the world of art throughout history and before, only recently that they are maybe playing a less dominant role.

Tomasz-W:
Every memorial is an artwork, but not every artwork is a memorial. 
On provided example photo from Vatican, these are statues of specific persons who used to live in previous centuries. It makes these figures closer to historic=memorial (the main function is to honour and keep a memory about these people) than an tourism=artwork (which they also are, but it's not a main function).

kocio-pl:
> Every memorial is an artwork, but not every artwork is a memorial.

:+1:  - thanks, this helps me a bit. So the artwork could be always an option and if something is double tagged (memorial+artwork), we might skip the artwork as the obvious part.

[Memorial definition](https://wiki.openstreetmap.org/wiki/Tag:historic%3Dmemorial):

> A feature for tagging smaller memorials, usually remembering special persons, peoples lost their lives in the wars, past events or missing places. 

The question is how does it work for Zeus for example? Is such a statue made for remembering him? It's easy to say about historic things from past few hundred years, but what is the main function of religious statues?

Maybe memorial definition should be extended somehow, to make it cover all the religious depictions, no matter if they're historic or not (or it is disputed)? Or maybe we need special key like `religious=*`?

Maybe this is a harder question and should go to the Tagging list, but we need some basic understanding how do we treat them here, because it might be too hard to know. For example we could do something like with forest/wood - maybe we could render all the statues the same, no matter if this is tagged as memorial (`memorial=statue`) or an artwork (`artwork_type=statue`).

Tomasz-W:
@kocio-pl As sometimes this is very hard to distinguish is it more a memorial-like statue or artwork-like statue, but surely it's a statue, I like your proposition to render them the same (artwork_type=statue has 16,7k uses out of 91,9k uses of tourism=artwork, so it looks like it's quite popular to tag the subtype there). For the rest of atrworks  I still propose a "totem" icon, which is much more uniwersal than the current one.

dieterdreist:
sent from a phone

> On 25. Aug 2018, at 21:03, Tomasz Wójcik <notifications@github.com> wrote:
> 
> Every memorial is an artwork, but not every artwork is a memorial.
> 


-1, not every memorial is an artwork.


> On provided example photo from Vatican, these are statues of specific persons who used to live in previous centuries. It makes these figures closer to historic=memorial (the main function is to honour and keep a memory about these people) than an tourism=artwork (which they also are, but it's not a main function).
> 



you are writing about things you evidently do not know much about, it would be wiser to get more information rather than insist on a hardly defensible position.

I suggest you have a short look on Bernini, although I fear you might tell me he was a historian rather than an artist ;-)
https://en.m.wikipedia.org/wiki/Gian_Lorenzo_Bernini

dieterdreist:
sent from a phone

> On 26. Aug 2018, at 00:50, kocio-pl <notifications@github.com> wrote:
> 
> Maybe this is a harder question and should go to the Tagging list


+1, there is no question this whole discussion should all happen on the tagging mailing list.

Tomasz-W:
> you are writing about things you evidently do not know much about, it would be wiser to get more information rather than insist on a hardly defensible position.

@dieterdreist Surely you should eat some Snickers
https://www.youtube.com/watch?v=uii3VhELiuE

kocio-pl:
> -1, not every memorial is an artwork.

The claim was that every memorial is an artwork. Do you believe it's not? Why?

dieterdreist:
sent from a phone

> On 26. Aug 2018, at 19:09, kocio-pl <notifications@github.com> wrote:
> 
> -1, not every memorial is an artwork.
> 
> The claim was that every memorial is an artwork. Do you believe it's not? Why?
> 


to counter this claim it should be sufficient to bring at least one example on the table that isn’t an artwork, here you go:
https://wiki.openstreetmap.org/wiki/File:Sign_at_Sherlock_Holmes_Museum_in_Baker_St_221b.jpg

kocio-pl:
What about wings on top of the totem being more clear? Maybe also some holes inside to make it visually different than solid obelisk/monument rectangles?

Tomasz-W:
@kocio-pl Bigger top part + 2 holes:

tourism artwork

Nakaner:
Regarding the tagging of stolpersteins (German translation of this posting is available at https://forum.openstreetmap.org/viewtopic.php?id=63458):

I am not convinced by the advantages of this change. It seems to me that the only motivation for this change is to clean the tagging and the omnipotent power of the map style OSM Carto to change tagging in any desired way is welcomed by the proponents of the change. It is well known that mappers did and do map for the renderer even if they do not want to. The usage of a certain tags by OSM Carto influences the decision which tags to use a lot. I expect the maintainers of the map style to stay neutral and not to push tagging into a certain direction.

Both tags are in the column "tags" (type: hstore) of the rendering database. OSM Carto's SQL queries are already full of `CASE WHEN … THEN … ELSE … END` clauses. It does not matter to support two tagging variant.

The two keys "memorial" and "memorial:type" don't convey any differences in meaning. The argument that a unification of tagging makes things easier for developers is a weak one.

Prefering one key over the other one does not improve the data. It will have only one effect: Mappers start editing almost all stolpersteins in Germany updating the last_modified timestamp of these nodes. This makes the nodes look like being reviewed recently but in fact they have not been checked on the ground (most will be present on the pavement but some might have been moved or even removed by construction works).

Best regards

Michael

kocio-pl:
> The two keys "memorial" and "memorial:type" don't convey any differences in meaning. 

So what is the use of having two different schemes for exactly the same thing? Why do you think it might be good?

kocio-pl:
@Tomasz-W Nice, I like the totem being tall and holes are making it visually different (just like gate in fortified castle vs hole in generic castle). Could you try these wings to more wings like, to make it more recognizable as a shape? Similar to https://thejesusquestion.org/2013/03/17/christmas-and-easter-totem-poles-by-david-k-fison/

Tomasz-W:
@kocio-pl 
* tourism artwork (v3)
* tourism artwork2 (totem)

I vote for 'v3' version. I think we shouldn't look there for a typically totem shape but for some universal shape - 'v3' might be interpreted as column or as totem, but propably will be interpreted as some kind of sculpture/ artwork, and I think is enough for `tourism=arwork` icon.

kocio-pl:
I though that wings would relate to something fancy and unusual. But I guess v3 might work, I just need to see the rendering comparison.

Tomasz-W:
@jragusa Please add test renderings of `tourism=artwork` with these two icons:

https://gist.github.com/Tomasz-W/aa3883f479d47710e85e9deaa64883cd

SomeoneElseOSM:
> make the stolpersteine (and other plaque-types) appear only on very high zoom levels

Stolpersteine I can understand (they're pretty small, literally on the ground) but other plaques often are a fairly obvious feature.

kocio-pl:
Plaques can be as tiny and disturbing as them. Without moving plaques to z19+ (#2689) my city for example would be just cluttered with them.

jragusa:
@Tomasz-W I'm not so convinced :/

https://www.openstreetmap.org/#map=18/46.10251/6.59780
artwork_before

`tourism=artwork (totem).svg`
artwork_1

` tourism=artwork (v3).svg`
artwork_2

jragusa:
Following your discussion, I would suggest to use the same icon for `tourism=artwork` and `historic=memorial` + `memorial=statue` because:
1/ Both describe an object in relief but with a different purpose.
2/ The meaning can be easily understood with the tag name.
3/ The map is designed for a general purpose.

Even if the current icon does not encompass all the diversity of artworks, it's easily understood by people.

Tomasz-W:
@jragusa As top part of the icon may be a little bit too prominent, can you add one more test rendering with v4 version?

tourism artwork4

https://gist.github.com/Tomasz-W/aa3883f479d47710e85e9deaa64883cd

jragusa:
artwork_3

kocio-pl:
Could you fix the conflict after merging #3345? Stolperstein discussion seems to be resolved in favor of `memorial=stolperstein` notation, so no changes are needed in the code.

We have the artwork rendering still not finished probably? It seems that v4 icon is OK for us and could be used for anything else than `artwork_type=statue`.

sommerluk:
@jragusa Thanks for working on this Pull Request.

I’m sorry but I’m strictly against merging this for a plenty of reasons. The two most important ones:

- As @matthijsmelissen [had explained yet](https://github.com/gravitystorm/openstreetmap-carto/issues/3088#issuecomment-395692530): It’s too detailed for a general-purpose map. What you propose will make the map harder to read, it is therefore a worsening for a general-purpose map. The proposed icons are not intuitive. Especially the one for monumental looks like the bench icon, but with somebody sitting on it. (Which is especially interesting because there’s another PR that wants to differentiate the bench icon more, also by adding something to the standard bench icon, but with a completely different meaning.) I’ve [explained the problem earlier](https://github.com/gravitystorm/openstreetmap-carto/issues/1186#issuecomment-408078372): When you add many differentiated icons that cannot be immediately recognized than maybe the mappers that are experts in this particular domain (here: very detailed memorial tagging) will get a little bit more feedback. But for the big majority of map users and mappers, the map will be harder to read.

- The treatment of stolperstein which has a tagging scheme about which [there is some recent discussion](https://forum.openstreetmap.org/viewtopic.php?id=63458) but as far as I see no consensus about if it should be changed or not. It’s for sure not the right moment now for making this change. We should first observe how the tagging develops. This map style must try to be neutral.

@jragusa Don’t take me wrong: I do not want to blame you for this PR at all. I appreciate the work that you have done. It’s just that this PR is not in line with the design goals of this map style. I’m sure in another map style with different design goals this could be a valuable contribution.

kocio-pl:
I see it quite different. I responded to some of these points, but since there are some new ones, I will reply here too, adding more details to the picture:

1.

Icons are much more intuitive for me and at the same time closer to what one sees on the ground. Standard memorial icon looks like a tombstone (it was even [designed this way](https://github.com/gmgeo/osmic/blob/master/amenity/cemetery-14.svg)). Of course, there are many memorials that look like this and one can even argue that tombstone is kind of memorial - and that's why we keep it as a general symbol, since there are different possible shapes and they share only the function, so it's hard to find something better (especially that we don't try to show tombstones currently).

Shapes are better in this PR:
a) bust icon looks just like a bust for me
b) statue icon looks just like a statue for me
c) memorial stone looks like a stone with inscription, which is also perfect for me
d) obelisk icon also looks like an obelisk; it is still similar to the monument icon, but it looks like monument icon is designed to look like an obelisk (the same problem with different shapes of monuments as with memorials) - currently they use the same icon and it will at least show the difference in a proper way (monuments are huge by definition, so the icon should be stronger, obelisk icon is less heavy, so all is good for me)
e) memorial bench typically includes a person sitting on this bench, so it's also good for me; we use a visual thread in more cases already (like car with shower or wrench or bicycle icon with parking letter), so choosing bench with some add-ons is nothing new and unexpected

General purpose map showing closer to the reality images sounds more user friendly to me, especially since we're not adding any new items, just showing them more clearly. This is also better for more precise tagging, since only about 1/3 of memorials have a type defined (including "war memorials", which might be moved to a separate tag, as discussed on Tagging list some time ago):

taghistory 42

2. 

I might not get something in German (automatic translation was not too helpful this time), but as far as I can see, voting was in favor of migration to memorial=stolperstein (4 votes, 1 for using plaque subtype). There was however parallel discussion in this thread about the right to start such discussion and probably also the plans to change rendering on OSM Carto, but this was another issue and I think anybody has the right to discuss tagging problems, including data consumers like map styles (because we see it more clearly). I also asked about it on Tagging list, which is the ultimate place to discuss such things, and it was mainly about details of tagging. I think I took enough care to talk about it and anybody interested in tagging should already know the proposition, it also lasts for about 2 weeks with no new ideas or questions popping any more.

That's why I feel the problem is resolved. It would be better for everyone to have one tagging defined instead of 2 disputed (so at least our decision is backed by one of them already). Moreover this PR does not make any harm to the rendering - we try to improve the situation, but even in the worst case the Stolperstein rendering will stay the same. At best, we will have more coherent database and small items won't clutter the map in Germany and some other countries (Nederlands have them also).

The discussion about who can propose things and how to handle disputed tags is a general problem and doesn't belong here in my opinion.

jragusa:
I will answer on Thursday. I'm a bit busy with fieldtrip for students today and tomorrow.

@sommerluk  no worries ;)

Tomasz-W:
I'm putting summary of needed changes to not skip anything, so please:
* left current generic `historic=memorial` icon, and use bench icon only for combination of `historic=memorial` +`amenity=bench`
* delete `memorial=war_memorial` tag from the list (this tag is a tagging mistake)
* change `tourism=artwork` icon to "v4" from https://gist.github.com/Tomasz-W/aa3883f479d47710e85e9deaa64883cd
* use `memorial=statue` icon also for `tourism=artwork` + `artwork_type=statue`
* fix conflict with `shop=bookmaker` (?) https://github.com/gravitystorm/openstreetmap-carto/pull/3356#issuecomment-418216993

jragusa:
@sommerluk I understand your point of view about the general purpose map but as @kocio-pl said they remain easily understood. Concerning the bench, it's a mistake from my side, as it's not dedicated for monuments.

I resolved point 2 and 5 of @Tomasz-W but now I have a major problem with the update of my branch. Strangely, my master is up to date but not my branch. So I cannot change artwork icon as it now includes in a folder which is not the case in my branch.

jragusa:
Additionally I need help to resolve the combination of `historic=memorial` + `amenity=bench`. I don't know how manage this in the `project.mml`. Each time I get a SQL error.

Tomasz-W:
@kocio-pl Can you help jragusa with finishing this PR? :)

kocio-pl:
Not at the moment, sorry. There's also problem how to make final decision with Stolpersteine which holds this issue back.

Adamant36:
@kocio-pl, can the PR just be merged without Stolpersteine or bench so that at least the other icons will be rendered? Then the problem ones could be dealt with in separate PRs on their own. Especially since it seems like Stolpersteine could take a while to work out.

kocio-pl:
Yes, it can be, good idea. However I still have not enough power to help with bench now.

jragusa:
@kocio-pl I updated the code following [comment](https://github.com/gravitystorm/openstreetmap-carto/pull/3356#issuecomment-418501228) of @Tomasz-W and removed memorial_bench.

Below you can see the different rendering of `artwork_type=statue`:
https://www.openstreetmap.org/node/4720792741 (scultpure)
https://www.openstreetmap.org/node/4720792726 (statue)
artwork_statue

Tomasz-W:
@jragusa Please update stone icon file, I missed it during pixel-aligning other icons.

https://gist.github.com/Tomasz-W/f1306db9c0086f0d209c50a9bf30ec73

Sorry for the problem ;)

jragusa:
@Tomasz-W done, no worries ;)

kocio-pl:
I would remove stolperstein support also, since this is not yet decided and I'm not sure how to proceed with that. We can get back to it later, as far as I've seen, all such complex solutions needs some fixing later on (it was true for towers, castles and information amenities). 

For now I will try to make independent tests. If I will find a time I will try to make a code for a memorial bench, but it also might be added later. This way or another I think next release will contain this code.

amenity-points.mss
```diff
@@ -765,21 +769,34 @@
     marker-clip: false;
   }
 
-  [feature = 'historic_memorial'][zoom >= 17] {
+  [feature = 'historic_memorial'][zoom >= 17],
```
kocio-pl: You need to exclude here all defined memorials which are to be shown on z18 or z19, because now they first look like a generic memorial.

amenity-points.mss
```diff
@@ -765,21 +769,34 @@
...
+  [feature = 'historic_memorial_stele'][zoom >= 19] {
```
kocio-pl: Do we have `historic_memorial_stele` defined in the `project.mml`? I guess no and we use simple memorial check now.

amenity-points.mss
```diff
@@ -765,21 +769,34 @@
...
     marker-file: url('symbols/historic/memorial.svg');
+    [memorial = 'bust'][zoom >= 19] {
+      marker-file: url('symbols/historic/bust.svg');
+    }
+    [memorial = 'blue_plaque'][zoom >= 19],
+    [memorial = 'plaque'][zoom >= 19] {
+      marker-file: url('symbols/historic/plaque.svg');
+    }
+    [memorial = 'statue'][zoom >= 19] {
```
kocio-pl: I think this should be still z17+.

amenity-points.mss
```diff
@@ -765,21 +769,34 @@
...
+      marker-file: url('symbols/historic/statue.svg');
+    }
+    [memorial = 'stone'][zoom >= 19] {
```
kocio-pl: I think this should be z18+.

amenity-points.mss
kocio-pl: I also think this should be rather z18+.

amenity-points.mss
jragusa: fixed

amenity-points.mss
jragusa: fixed

amenity-points.mss
jragusa: both fixed

amenity-points.mss
jragusa: What would be displayed at z17 ? If I exclude all the features listed below, we will highlight from z17 `historic=memorial` without memorial subtags which are not the most important ones.

amenity-points.mss
jragusa: stone also to z18 ?

kocio-pl: Yes, I think so.

jragusa: fixed

amenity-points.mss
jragusa: I made corrections, I have tested on `memorial=stele` and `memorial=bust` and will check the rest tomorrow

jragusa:
@kocio-pl do you want additional screenshots with last modifications ? I checked each features and everything seems correct.

kocio-pl:
No need to, I will check it soon. However not all elements have proper initial zoom level, my take is to change:
  - [memorial = 'plaque'][zoom >= 17] -> z19+ (the same as blue plaque)
  - [memorial = 'bust'][zoom >= 19] -> z18+ (bigger than plaques, smaller than full posture/statue)
  - there's also missing stele image rendering (which is present in labels rendering) - I would make it z18+.

I'm deeply sorry for the time it takes, I should probably start with listing all initial zoom levels. Multiple objects changes are hard to manage and test... It's quite a new challenge for me.

jragusa:
@kocio-pl don't be sorry. I'm not in the hurry. 
1/ fixed
2/ fixed
3/ `memorial=stele` is not hardcoded because it has the same icon than generic memorial and filtering in the header allows them to render from z18.

I took the opportunity to clean the code and hence added a dedicated line to `memorial=stele` for clarity. What do you think ?

amenity-points.mss
```diff
@@ -765,21 +769,40 @@
...
 
-  [feature = 'historic_memorial'][zoom >= 17] {
+  [feature = 'historic_memorial'][memorial = null][zoom >= 17],
+  [feature = 'historic_memorial'][memorial = 'statue'][zoom >= 17],
+  [feature = 'historic_memorial'][memorial = 'bust'][zoom >= 18],
+  [feature = 'historic_memorial'][memorial = 'stele'][zoom >= 18],
+  [feature = 'historic_memorial'][memorial = 'stone'][zoom >= 18],
+  [feature = 'historic_memorial'][memorial = 'blue_plaque'][zoom >= 19],
+  [feature = 'historic_memorial'][memorial = 'plaque'][zoom >= 19],
+  [feature = 'historic_memorial'][zoom >= 19] {
```
kocio-pl: The last issue: I think this should be z17+, since we have no idea what can it be, so it's kind of generic.

jragusa: Are your sure for `memorial=blue_plaque` ? They seem to be not so large. Take a look at some examples from the [Guardian](https://www.theguardian.com/books/2009/nov/28/london-plaques-emily-cole-review).

I would put the same zoom level than stolperstein.

kocio-pl: I meant all the `memorial` values different than what we have on z18+ and z19+. Look at the similar case here:

https://github.com/gravitystorm/openstreetmap-carto/blob/9d16b996591c240080159ac9d11f69540d159637/amenity-points.mss#L474

jragusa: Is the fix satisfy you ?

kocio-pl: Thanks, it looks promising and I will test it soon. I just think that excluding statue is not necessary, because it will be rendered on z17+ anyway (that's why I told only about excluding z18+ and z19+ objects). It adds a bit more of clarity, so it's not an error, but I think this line is long enough.

kocio-pl: I have found the problem - this war memorial is not showing any icon at all (instead of a standard one), while the label is still visible from z17:

https://www.openstreetmap.org/node/3183626423#map=19/52.26992/20.99322

jragusa: I resolve both problems. The second one was related to a missing comma.

kocio-pl:
Thanks! As far as i could test it, it's OK, but rendering memorials etc. is so complex, that it will need an update anyway (especially after stolperstein will be resolved, but also a memorial bench should be properly depicted).

It's good to see more familiar shapes and distribution among multiple zoom levels, because memorials are similar only as a tagging idea, in reality the world of commemorating is much more diverse.

kocio-pl:
I don't need it.

Issue #3188, opened 2018-04-22
Render names of building:part


DeusFigendi:
I'm just having a german discussion [here](https://www.openstreetmap.org/note/1293463#map=18/52.03824/8.49377) about tagging building parts as buildings to get the names of the parts displayed  in OSM-carto because names of building:part which ain't buildings are not rendered.

Well I think it's right, the names of building:part should be rendered so I'd like to suggest this.
Maybe at a more detailed zoom level but buildings maybe smaller or lighter but the buildings name but it should be rendered I'd say.

kocio-pl:
I'm not sure if this is good idea. Could you show some examples of building:part name which make sense without the outline?

DeusFigendi:
Yes of coarse,
my example was [the university of Bielefeld](https://www.openstreetmap.org/way/582358481#map=17/52.03779/8.49493). You can see the university and all "around" letters, in the east "A" and going north-western B-C-D-E (and F) and on the south side S-T-U (V) W. Very south-eastern there is "J".
Well this letters are the **names** of building-parts because the whole thing is **one building**. But the names are only displayed/rendered because the building:part s are tagged as buildings too, but this is semantically wrong. To have the names of the bulding party displayed is the reason why [Heros](https://www.openstreetmap.org/user/Heros) tagged them this way he/she told me.

This is my example.

I **could** also imagine castles or the like with building:part s named "west wing" or the like but I'd guess it's not very common to give (somehow official/documented) names to building parts. I estimate this maybe on administrative buildings or huge, somehow open/public buildings.

kocio-pl:
Aren't they rather a `ref` than `name`?

DeusFigendi:
Hmmm that's a very valid question.
To be honest: I don't know but I don't think so. I **think** `name` is the correct key here. But I will ask for other opinions.

michaelletzgus:
I'm just testing the 'ref' with building part A.
I thing 'ref' is semantically OK because the name of the whole building, including the parts, is "University".
The different parts have more specific refs.

This scheme will work until a reason for a different name occurs, for example "Bielefeld University" and "University of applied sciences" in the same building.

So I would prefer this scheme for *:part objects:

1. Render name if present
2. Render ref if present but no name

Or: Introduce "*:part:name"

dktue:
I think rendering of building-parts is a good idea; consider the following example:

[...]

PR #3364, opened 2018-08-26
WIP: Indoor rendering

meased:
First off, this is not a (completely) serious PR. This is just me messing around and I got some cool results that I decided to share. 

The first obvious issue with indoor rendering is that there is no way select different floors. I solved this by only displaying the main floor, e.g. `tags->'level' IS NULL OR tags->'level' = '0'`.

The results:

https://www.openstreetmap.org/#map=17/45.52838/-122.66303
orconv_17
orconv_18

https://www.openstreetmap.org/#map=18/45.62839/-122.66863
library_17
library_18
library_19
And zoom 20, just for fun.
library_20

Lost the link to this one (somewhere in Taiwan):
taiwan_17
taiwan_18
taiwan_19

Here's what I learned:

### Pros

- The results were surprisingly good. It almost seems useful.
- Something like this would encourage proper indoor tagging instead of tagging rooms as separate buildings and corridors are pedestrian areas (this happens a lot).
- Currently, properly tagged indoor areas look terrible in this style, all POIs from all floors are collapsed and displayed on the same layer. Merely filtering them is an improvement. For example, this is actually a very neatly mapped mall (with three levels of POIs ):
current_indoor
Same mall (try to imagine with only POIs from one level):
westfield

### Cons

- Bad user interface (people will be searching for the floor selector widget, but will never be able to find it).
- The major building color is way too dark and it looks terrible with this rendering. (although POIs in major buildings already look terrible IMO):
major_pois
- This does not parse the level tag properly (it should support values like `0-3`, `0;2;3`).
- There are some notable examples where the "main" floor is not `level=0`.
    - The tagging scheme is currently missing a way to identify the "default" floor. It would be nice if that was contained entirely in the `level` tag as big nasty relations are both hard to map and hard to query.

This is a 50% solution. The question I'm trying to push is this: could something like this work as a starting point to build up to a 100% solution, or do we hold out for the 100% solution.

### DIFF
--- a/amenity-points.mss
+++ b/amenity-points.mss
@@ -2864,6 +2864,24 @@
     text-halo-fill: @standard-halo-fill;
     text-placement: interior;
   }
+
+  [feature = 'indoor_corridor'][zoom >= 18][way_area >= 150],
+  [feature = 'indoor_room'][zoom >= 18][way_area >= 150],
+  [feature = 'indoor_area'][zoom >= 18][way_area >= 150],
+  [feature = 'indoor_corridor'][zoom >= 19][way_area >= 50],
+  [feature = 'indoor_room'][zoom >= 19][way_area >= 50],
+  [feature = 'indoor_area'][zoom >= 19][way_area >= 50] {
+    text-name: "[name]";
+    text-size: 9;
+    text-wrap-width: @standard-wrap-width;
+    text-line-spacing: @standard-line-spacing-size;
+    text-fill: @indoor-text;
+    text-dy: 0;
+    text-face-name: @standard-font;
+    text-halo-radius: @standard-halo-radius;
+    text-halo-fill: @standard-halo-fill;
+    text-placement: interior;
+  }
 }
 
 #amenity-line {
--- a/buildings.mss
+++ b/buildings.mss
@@ -5,6 +5,8 @@
 @building-major-fill: darken(@building-fill, 20%);
 @building-major-line: darken(@building-major-fill, 25%);
 
+@indoor-text: darken(@building-fill, 40%);
+
 @entrance-permissive: darken(@building-line, 15%);
 @entrance-normal: @building-line;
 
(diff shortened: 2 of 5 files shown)


Tomasz-W:
As I really would like to have some separate service with indoor rendering, I'm worried this is a bad idea for default style. It's very detailed at the moment, and this feature would complicate orientation in map a lot. I mean it would be very nice to have some separate service with this function for people interested in certain buildings interiors, but I'm worried that for most of users it may be an obstacle.

Anyway, I would like to see building:part=* names (https://github.com/gravitystorm/openstreetmap-carto/issues/3188), which on the one hand will be useful, and on the other hand they won't clutter the map.

Tomasz-W:
@meased I forgot to mention one argument, which actually may be the most important. As I can admit that your test renderings looks quite ok and promising, they are presenting well-mapped interiors. but I remember examples of indoor mapping in my city - building parts was out of building shape, certain rooms wasn't stich to each other, etc, I'm worried how would it look in cases like that.

The perfect solution for me would be to allow interion rendering after clicking on building shape, but ad we don't have vector tiles and clicability, it won't happen soon.

dktue:
This is definitly something that should motivate us to get to vector tiles faster. :-)

meased:
@Tomasz-W yes, I hand selected areas that look good, and ignored some rather major problems. Poorly mapped areas are going to look bad, and I personally don't think this style should attempt to make them look better, it provides motivation to clean the data.

Its [been mentioned before](http://blog.imagico.de/on-the-discussion-on-osmf-supported-vector-tiles-maps/), but vector tiles aren't a magic wand that will solve all cartography issues. If we had vector tiles today, a building with only one floor (the most common case) would look and behave identical to the images above, and we can start solving those problems today.

Another issue is the competition between indoor rendering and 3D roof rendering, some people will want one, some will want the other. Vector tiles to the rescue, right? But what does that look like? What do you show by default? How many layers are hiding? How do you know they are there? It's an interface nightmare...

polarbearing:
I agree with @Tomasz-W that this style should focus on the flat map. There are other interfaces that address indoor and levels, beside the one mentioned by @meased you might have a look at https://openstationmap.org which is switchable between 3D/2D.

kocio-pl:
First of all, thanks @meased  for making such research! :heart:  I like very much your analyze.

However I don't feel it's good to show indoor on OSM Carto, because adding internal details will clutter map sooner or later. The only thing I think that could be useful is filtering levels other than "0" (including most of the underground objects) - see  #1977.

kocio-pl:
Thanks again for this idea, however I still think it would add visual clutter in general, so I will close it now without merging.

Issue #2532, opened 2017-01-03
Rethinking major and minor buildings


kocio-pl:
Rewriting the buildings in #565 took place 1,5 year ago and proved to be right in general, but it's time to fine tune it. It become apparent to me in a discussion about unifying current major buildings (https://github.com/gravitystorm/openstreetmap-carto/pull/2515), which I meant as a simple, clearly focused fix, but turned into discussing special buildings rendering.

There are some related issues already created:
- https://github.com/gravitystorm/openstreetmap-carto/issues/738 (Render building=greenhouse differently)
- https://github.com/gravitystorm/openstreetmap-carto/issues/1206 (Amenity=place_of_worship buildings)
- https://github.com/gravitystorm/openstreetmap-carto/issues/1207 (building=house and building=garage should render in slightly different colours)

but they are also related to certain features, while the underlying problem is wider:
1. Do we want having any other building types rendering than just standard brown?
2. If yes, which buildings should be rendered as major?
3. If yes, which buildings should be rendered as minor?

My take is that it's good to have basic subtypes of building class which is very large and important for universal map style, but has no clear rules for differentiation. That means it's up to us to decide. Some buildings are landmarks or plays a special function (like places of worship or transportation terminals), while the others are less important than typical buildings (like greenhouses, sheds, roofs or garages) and we should show it if possible. Major building list should be however treated with more care than minor ones to [avoid visual clutter](https://github.com/gravitystorm/openstreetmap-carto/pull/2515#issuecomment-270024246).

My opinion is that using dark brown for major buildings and light brown for minor ones is a consistent rule and hints the users that all these are still buildings.

pnorman:
Historically we had three classes of buildings

- Normal/light buildings. These were differentiated on based on the building tag and a confusing list that didn't have a solid reason behind it

- Low-zoom buildings. These were rendered the same as other buildings, except if they were also place_of_worship, aeroway, or railway=station

We cannot reliably style buildings based on what they contain without a query like https://github.com/gravitystorm/openstreetmap-carto/pull/565#issuecomment-44358786.

Stepping away from 3 classes of buildings was a big step for making the logic behind rendering understandable. Let's not do anything that would take us backwards.

wmyrda:
Seems that moving away from 3 stage hierarchy did not go as planned hence there are buildings that render differently like airports.

#565 made almost all buildings render lighter including those that are tagged with only building=yes. This time around no one wants to render those as major. 

[...]

PR #3372, opened 2018-08-30
Moving amenity=atm to z19+

kocio-pl:
- Render amenity=atm on z19+ instead of z17+

ATMs are small facilities which make clutter on the map on z17 and z18 when located near some other POIs, which is common case in cities. ATM is smaller than bank, where it can be usually placed, so it should be rendered later, and since their size is smaller than all the shops, I think they should be rendered from z19.

Just two examples out of many:

[Bank is eclipsed by ATM at z17:](https://www.openstreetmap.org/#map=17/52.27057/20.97482)
Before
ht3d59_z

After
pay1at_a

[ATM eclipsing label of a bigger amenity at z18:](https://www.openstreetmap.org/#map=18/52.23101/21.01254)
Before
iuttcv0

After
ru1sfhmv

### DIFF
--- a/amenity-points.mss
+++ b/amenity-points.mss
@@ -56,7 +56,7 @@
     marker-clip: false;
   }
 
-  [feature = 'amenity_atm'][zoom >= 17] {
+  [feature = 'amenity_atm'][zoom >= 19] {
     marker-file: url('symbols/amenity/atm.svg');
     marker-fill: @amenity-brown;
     marker-placement: interior;
@@ -2277,7 +2277,7 @@
     }
   }
 
-  [feature = 'amenity_atm'][zoom >= 17] {
+  [feature = 'amenity_atm'][zoom >= 19] {
     text-name: "[operator]";
     text-size: @standard-font-size;
     text-wrap-width: @standard-wrap-width;


dieterdreist:
I guess the issues you mentioned could have been resolved also with a less drastic cure like reducing the priority of atms in rendering (render them and their labels later).

Probably it once was a conscious decision to render these already at z17 because they are quite important to almost everyone, regardless of personal interest, especially useful for tourists.

For me it is ok to remove them or to ban them to z19, if there is some idea behind this move, just that it doesn’t seem there was a thorough consideration of the topic, and I haven’t seen real arguments why it will be removed from z17 and 18.

kocio-pl:
Yes, I just recently wrote an article about size principle in universal map:

https://www.openstreetmap.org/user/kocio/diary/44852

kocio-pl:
My argument is that in the universal map size is the only universal property we may rely on.

If this was touristic map style, we would have a simple guiding rule, but it's not. When some objects are usually standalone (in the outdoor for example), they could be rendered earlier, but this is not the case for ATMs.

This is also similar case to tower/castle: towers and ATMs might be standalone, but we have no way to now that from tagging, so castle or bank should be rendered earlier as a generalization.

imagico:
Note the change proposed here was more or less already proposed and rejected back in 2015 in #1884 - along with a lengthy discussion of the _typical physical size equals cartographic importance_ paradigm.

kocio-pl:
3 years later this problem has grown in a significant way (from ~80 k uses to ~140 k uses) and it keeps growing constantly:

taghistory 32

imagico:
The increase in numbers is hardly increase in density, it is primarily increase in breadth to areas previously not mapped.

As already mentioned back in 2015 the problem you are trying to address is self fabricated by the inflationary addition of new icons.  This applies even more today.  After all it is definitely not that ATMs have somehow become less relevant over the last years.

kocio-pl:
Just to make things clear: of course density is not bigger in the best mapped areas, which were complete 3 years ago (they were the showcase of this problem back then). But density is bigger in all the places that are better mapped now (no matter if they're complete now) and there are more places where this problem exists.

lakedistrictOSM:
Can the ATM labels be removed too? They just add unnessecary clutter to the map.

mboeringa:
> After all it is definitely not that ATMs have somehow become less relevant over the last years.

Well, actually, this is a slight side-step, but ATMs are fast becoming dinosaurs in some countries like the Netherlands where I live. We are fast becoming a paperless and cashless economy. I rarely visit an ATM these days, probably only once in two months. Everyone pays with their debitcard here, which is also possible almost anywhere, even in temporary beach clubs at the coast that are only there during the summer season.

And it is not just in highly developed countries. I can't recall where I read it, or about exactly what country it was, but I remember reading an article last year about one of the emerging economies in Africa where the leadership was pushing IT, and that people were quite massively turning to digital paying methods there as well.

Ah, found it, of all places, it was Rwanda:

https://www.betterthancash.org/news/media-releases/rwanda-to-accelerate-digital-payments-by-joining-the-better-than-cash-alliance

dieterdreist:
sent from a phone

> On 31. Aug 2018, at 08:48, mboeringa <notifications@github.com> wrote:
> 
> We are fast becoming a paperless and cashless economy.


+1, there is an agenda to push for the abolition of cash in order to be able to better monitor the flow of money, but there is also a countermovement of privacy focused people. 

And there are alternatives kinds of electronic money pushing into the market, like apple pay or google pay.

StyXman:
> ATM is smaller than bank, where it can be usually placed, so it should be rendered later

In US you can find ATMs inside bars and restaurants. The only two ATMS in this neighborhood are the two restaurants: 

https://www.openstreetmap.org/#map=19/37.50370/-122.48550

(I should check if they're mapped).

kocio-pl:
I just realized that ATMs might be a problem everywhere, not only in the big cities. You can find them also in the rural areas, but from my observation they are rarely in some distance from shops, like few meters (which would be OK in rural area). Usually they are located near the entrance of shop or are embedded in the wall. So they might eclipse shops no matter where they are located, they work effectively like visual "parasites" of shops.

dieterdreist:
sent from a phone

> On 31. Aug 2018, at 12:54, kocio-pl <notifications@github.com> wrote:
> 
> So they might eclipse shops no matter where they are located, they work effectively like visual "parasites" of shops.


it is a deliberate choice of the (former?) style sheet developers to give precedence to atms rather than shops and does not have to do with zoom levels (although you could use different priorities between the same things on different zoom levels).

kocio-pl:
Do you think that they knew how the map with this style will look like 6+ years from the moment of this choice? I don't.

There was much less objects then and we have white list with shops - only few of them had icons and most of them were not even shown as a dot. And the icon for ATM was there.

In 2009 I was hoping that the map will be complete - so I wanted to add all the house numbers in my neighborhood, because there were streets and they even all had names! This was all I hoped for then. 

I don't know when exactly Mapnik XML style has been designed (OSM Carto was just technical reimplementation initially), but it might make sense then. But we're in a very different world now, welcome in OSM 2018...

dieterdreist:
example where restaurants and cafes clutter the map at the same zoom level: https://www.openstreetmap.org/#map=17/48.85325/2.33847

or

https://www.openstreetmap.org/#map=17/41.89943/12.47216

you can find situations like these for some features, sometimes it just shows the particularity of the area and is not necessarily a problem.



sent from a phone

Tomasz-W:
@kocio-pl related to https://github.com/gravitystorm/openstreetmap-carto/issues/1745

jeisenbe:
Re: "it is a deliberate choice of the (former?) style sheet developers to give precedence to atms rather than shops and does not have to do with zoom levels"

I agree with this style choice. ATMs are very important in most non-European countries, where cash transactions are still the norm, and they are not as common. I've often needed to travel a kilometer to find the closest one, even in a town. Also note that the Humanitarian style has ATMs rendered prominently, which I mention as a proxy for their importance in lesser-developed countries.

Instead, I would recommend deprioritizing an ATM near a bank, so that the bank shows instead, and deprioritizing the name lable for ATMs. That should solve both of your sample problems without removing the ATMs from the map entirely at z17 and z18.

dieterdreist:
2018-09-10 15:56 GMT+02:00 jeisenbe <notifications@github.com>:

> Instead, I would recommend deprioritizing an ATM near a bank, so that the
> bank shows instead, and deprioritizing the name lable for ATMs. That should
> solve both of your sample problems without removing the ATMs from the map
> entirely at z17 and z18.



+1, indeed I had recommended this as well in the first reply, but it has
been ignored ever since.

kocio-pl:
Sorry for not answering this suggestion, I wanted to answer general issues first and then I needed to think a bit.

Humanitarian style can afford to make arbitrary choices about importance, because it's much simpler than OSM Carto. They don't try to follow the database changes, because their scope is very limited, unlike ours, and they are expected to follow undeveloped areas more closely than we, so this is not surprising to me. Also searching is not what general style is made for. This is where mobile apps or web search field come into play.

I believe size rule is good way to go, because it will make map more coherent and easier to declutter at the same time. But we could test your idea too. Is anybody ready to try the code with less priority of ATMs? Just please remember that we don't have a way to express "near a bank" property, it has to be "near anything". I would not like ATM to eclipse shop or amenity like post office or gym too.

jeisenbe:
Re: " I would not like ATM to eclipse shop or amenity like post office or gym too."

I'm not sure that ATMs are less important cartographically than a generic shop, even if they are smaller (though some ATM halls have a dozen different ATMs, and are the size of a kiosk)

As you mentioned, "searching is not what general style is made for"; it would be easier to find a particular shop by searching for it's name, or by searching for it's category. Other services that provide reviews and more details about shops are also more useful.

But while seeing a shop icon on the map is not very informative (especially without a name), an ATM icon alone is enough to know "I can get money out of my bank there". In your first picture you showed a bank icon obscured by an ATM icon, on a corner with 5 banks:



In this situation, the ATM icon is more useful than the bank icon, even though the bank is larger. What good is it to know that there are 5 banks, if we don't see the names. I can't get any services at a bank where I have no account, so I'll have to zoom in or just search for the right bank. But most ATMs will work with cards from other banks and other countries, so just seeing the location on a general map is useful.

kocio-pl:
You see? 

You make a lot of "hand waving" about ATMs and you're still "not sure" what is cartographically more or less important. Of course - you can't be sure, because you have no rule even for shop/bank//ATM trio. 

I use simple size rule for them: I have seen ATMs being part of the bank or shop, yet I have never seen a bank o shop being a part of ATM, so typical size is clear - ATMs are smaller and therefore should appear later than both of them. And this rule works as effective also for castle/defensive tower, museum/toilets and many other cases we will test in the future.

dieterdreist:
2018-09-11 12:10 GMT+02:00 kocio-pl <notifications@github.com>:

> You make a lot of "hand waving" about ATMs and you're still "not sure"
> what is cartographically more or less important. Of course - you can't be
> sure, because you have no rule even for shop/bank//ATM trio.
>
> I use simple size rule for them: I have seen ATMs being part of the bank
> or shop, yet I have never seen a bank o shop being a part of ATM, so
> typical size is clear - ATMs are smaller and therefore should appear later
> than both of them. And this rule works as effective also for
> castle/defensive tower, museum/toilets and many other cases we will test in
> the future.
>

to me, basing rendering priority only on "size"  is arbitrary. Yes, it is a
concept, but it is not suited for solving the complex decision we have to
make, in a satisfying way.

I could come up with another arbitrary criterion, the "time rule". If
something is available for more time it is more important to show
(hypothesis). A christmas market is there only for few weeks in the years,
so we prefer to show the situation on the rest of the time. Banks are only
open a few hours on weekdays vs an atm which is typically available 24/7.
See, atms should dominate banks.

Another arbitrary criterion: "how big is the group potentially interested
in this?" Atms are useful for anybody that has a bank card of one of the
circuits it offers. That is orders of magnitude more people than those who
are interested in a bank (must be client of the bank and maybe even of this
particular branch for most operations).

Another arbitrary criterion: significance. How significant / important is
something compared to other things, how many people know it. A newsagent
will rarely be known outside its surroundings. A major museum can be world
famous and will be interesting to many tourists and will usually serve as a
reference point (many people know it). You might not want to go to the
museum every day, but it is useful to see it labeled for orientation.

There are probably more criteria, and to me what we have to do is balance
these and decide on a case by case base how to deal with it, how to weight
the criteria.

kocio-pl:
Nice try. :smile: Of course, any rule is arbitrary and it might be good to look at some alternatives - I just use what my experience with this style shows is working universal way. However how easy is to test all these rules? Is it visible somehow?

For example - significance: it's so hard to find good test, that any tagging scheme using it is dropped in the Tagging discussion, even if multiple people wanted to establish some. The only one quite popular tagging of similar type that I'm aware of is road smoothness. Not too objective, nor universal, sometimes just unknown.

How big group is interested in - how would you measure this? And what if this group already have special style available? In which part of the world is this group big and where it will be small? Not too universal for me and also unknown in many cases. Example: on cycling map layer there are pubs shown in the prominent way - in Poland you can't drink alcohol when riding bike (not even radler - sorry...), so even if we know exactly the group and style is very specific, it can be different in different countries.

Time rule - we use it already for example for water, but most of the time it's a tagging rule: most volatile objects are simply not allowed. We could of course parse time tags, but this is object related, not for the whole class. And try to see this difference - also not easy to apply. All the permanent objects should be rendered from the same zoom level? How many different classes would we get this way (2 perhaps - permanent and seasonal)?

Size rule is not my invention not anything new, I just try to stick to it more because we have too many hard to compare class of objects. We already show cities later than countries and neighborhood later than suburbs (...so it was piece of cake for me to add quarters to this scheme - just slip it between them!). We even show supermarkets earlier than convenience shops - luckily there are different tags for them.

This is also not the only rule - for example roads make a hierarchy so it's quite easy to depict it (however note that it's still somewhat related to their size). We have also more freedom with the outdoor features for example - there is not too much to show, so we can show them earlier, exaggerate them a bit. But this is when things are easy (yet even for outdoor objects relative size might be useful - defensive tower/castle case is typical there). When it's not that easy - like in dense tagged areas - we need some simple and universal rule.

dieterdreist:
sent from a phone

> On 11. Sep 2018, at 13:11, kocio-pl <notifications@github.com> wrote:
> 
> Size rule is not my invention not anything new, I just try to stick to it more because we have too many hard to compare class of objects.


to be clear, size is an important factor for deciding what to show when, but it is not the only one. Yes, one is bigger than the other, but within the range of object sizes (and common zoomlevels) banks and atms are similar (things smaller than a building).

When you write „stick to it more“ I understand „give more importance to size than to anything else“. 
Should an office building or a parking have priority before a post office or a church, because they have a larger footprint? Would we prefer a forest name over a city label?

kocio-pl:
> to be clear, size is an important factor for deciding what to show when, but it is not the only one. 

You're right and I've just said it.

> Yes, one is bigger than the other, but within the range of object sizes (and common zoomlevels) banks and atms are similar (things smaller than a building).

You took too big granularity when comparing them to buildings (that is the downside I mentioned with time based rule). We could also say that they all are the same compared to country - yes, they are but it's too big generalization. But it helps a lot when we try to clean z17 a bit, which is very dense in WMAs - at this level they have different size and this is simple to tell which are bigger and which are smaller.

> When you write „stick to it more“ I understand „give more importance to size than to anything else“. 

I mean "let's not omit it from our set of rules" (it's not there now) and "let's use it as a universal rule to avoid subjective _importance wars_".

> Should an office building or a parking have priority before a post office or a church, because they have a larger footprint? Would we prefer a forest name over a city label?

Parking may have very different size, but we already deal with it properly - "P" sign is shown only when it's big enough. The same with the forest. They are automatically prioritizing themselves. So are the city labels. It's better than typical class size, it's per object, which is closer to the truth.

dieterdreist:
if I counted right, there was one comment (lakedistrictOSM) in favor of removing the atms from z17, some general remarks about atms and at least 3 against it (imagico, jeisenbe, dieterdreist) , plus the rejection by the former team in [#1884](https://github.com/gravitystorm/openstreetmap-carto/pull/1884). You are still insisting in deploying it?

kocio-pl:
Yes. Up to this moment no good reasons have been given against this solution. I responded to all the critical comments, I guess.

dieterdreist:
You are repeating the size idea like a mantra:

> dieterdreist: I’m not convinced at all this map could work by reducing the whole selection process essentially to depend on the size, and even less on the “typical size” of things.

> kocio-pl: My argument is that in the universal map size of the object is the only universal property we may rely on.
> kocio-pl: I believe size rule is good way to go, because it will make map more coherent and easier to declutter at the same time.
> kocio-pl: You make a lot of "hand waving" about ATMs and you're still "not sure" what is cartographically more or less important. Of course - you can't be sure, because you have no rule even for shop/bank//ATM trio. 
> I use simple size rule for them: I have seen ATMs being part of the bank or shop, yet I have never seen a bank or shop being a part of ATM, so typical size is clear - ATMs are smaller and therefore should appear later than both of them. And this rule works as effective also for castle/defensive tower, museum/toilets and many other cases we will test in the future.


If you believe the "simple size rule" is the answer to all cartographic decisions, I believe you should not be deciding about the development direction of this style. I know that meritocracy implies who does things can decide how they are done, but it also has the word "merit" in it, it is not a carte blanche for anything someone could hypothetically do.

imagico:
@kocio-pl - when you speak of _good reasons_ you are evidently speaking of reasons you appreciate and support, not of reasons with objective substance and merit.  You dismissed the past decision against this change because you don't like it and don't like the rejection of the _physical size equals cartographic importance paradigm_ implied in that but you make no attempt to actually argue with the points made and try to convince those critical of the change to change their opinion.

@dieterdreist - since consensus is no more required for merging changes here since early 2017 serious arguments and critical discourse about the merits of changes have mostly vanished from the discussion.  So in day-to-day operations this is definitely not a meritocratic project any more.  If it ever truly was is a different question - but reasoning with other opinions and convincing others of the merits of your change was necessary and this was a key element in the social fabric of this project.

imagico:
I made no claim or guess about your motivation - that you don't like the past decision against this change is an observation evidenced by you making this PR.  That you do not engage in an open argument and actually reason with the arguments against this change is likewise an observation and not a claim.

jeisenbe:
I wonder if there is a cross-cultural communication problem here?

For one, the importance of ATMs strongly varies by location and culture.
When I was back in the USA last year, I used an ATM only 2 or 3 times. But
here in Indonesia I use one every other day: to get cash, but also to
refill my mobile phone credit, pay electric bills, transfer money to
contractors, pay for plane tickets and hotel rooms, transfer money to
friends etc.

Second, in this conversation there are different levels of directness in
communication. I believe @kocio-pl <https://github.com/kocio-pl>, @imagico,
and several others are from central northern European cultures, where
direct confrontation and frankness are admired cultural traits. Americans
and Brits are more circumspect with criticism (though we Americans are
effusive with praise and our use of exclamations!). I don't believe there
are many people from non-Western cultures currently contributing to this
project. In Southeast Asia, we never directly confront someone when we
disagree with their work or their reasoning, instead everything has to be
done very politely and indirectly; it is assumed that the other person will
understand the implications.

Eg, writing "Up to this moment no good reasons have been given against this
solution" sounds extremely harsh and critical in SE Asian culture, and it
also sounds rude in American culture. This implies that all of the
arguements given above were "no good", which means "bad" or "worthless" in
my dialect.

The proper way to phrase this would be: "So far several good arguments have
been brough up for rendering ATMs at the current zoom level, and I really
appreciate all of your ideas and suggestions. But in my opinion the need to
reduce visual clutter and the need to show banks and shops at this zoom
level is more important."

I appologize for explaining how to phrase this in English; everyone in this
conversation has excellent English language skills. I am only offering a
more cultural appropriate, polite way of phrasing the same sentiment

In conclusion, I would like reiterate my disagreement with @kocio-pl
<https://github.com/kocio-pl>'s assessment of the importance of ATMs. While
this issue is small, I do not think it is a good idea to start prioritizing
features by their size alone, when discussing different categories. That's
comparing apples to oranges, and saying big oranges are more important that
big apples, because the oranges are biggest (I hope this expression makes
sense, perhaps it is an Americanism?)

Respectfully yours,
Joseph Eisenberg

jeisenbe:
Previously, I wrote **"I'm not sure that ATMs are less important cartographically than a generic shop, even if they are smaller "**
And @kocio-pl replied: 
_'You make a lot of "hand waving" about ATMs and you're still "not sure" what is cartographically more or less important. Of course - you can't be sure, because you have no rule even for shop/bank//ATM trio.'_

This is also an example of cultural misunderstanding. When I said "I'm not sure..." it was an indirect way to say: "I am SURE that ATMs are MORE important..."

That's just the polite way to say it in American (and British?) English. I believe that ATMs are more important, based on personal experience as a mapper and map user, not based on a systematic rule that can be applied to all amenities.

My "systematic" rule would be: the rendering level and priority level of POIs should be based on the POI's cartographic importance for a general-purpose map, determinded by consensus of a super-majority of style contributors, mappers and map users. So any change needs to be discussed and agreed upon.

kocio-pl:
Now I'm losing hope of replying - too much to write and too much threads. But maybe there will be no need to reply.

I like your insight. Of course I am culturally biased, but it's still possible to talk in a different way. Matthijs is much better at this, even if he's also European male (just as Christoph and I). Being direct and at the same time not judge people is possible.

I'm getting interested to hear more about your proposition, however it's not clear for me. What do you mean by "cartographic importance"? This seems to me to be the core problem here.

But please, discuss governing issues in a different ticket.

kocio-pl:
Thanks for acknowledging. I'm still learning how to conduct project like this. However this is still governing issue, which does belong to the re-opened #2436 - please discuss it there.

Adamant36:
I don't know if it would be a solution to this issue necessarily, but I've been thinking there should be a brown dot for amenities like there are for other things. Although I think z17 is a little early for ATMs to show up and z19 seems like the perfect level for it, it would still be nice to know there is "something" there at z17. Not to mention the dot would help create some uniformity with amenities in general at that level. There's a couple of other issues that might benefit from it also. Like there's a small local park in my area that has a lot of stuff in it, including a BBQ, at z17 Id probably be doing a scan over the area just to see a broad perspective of the features there so a dot would be usefully. I don't need to know there's a bbq there though until maybe z18 or z19 when I zoom in for more details. The bbq just isn't important enough for z17 and the icon is bigger then the mapped houses in that area. It just doesn't fit visually or feel like the right level for it, but it is solved at z18 or z19. I feel the same way for ATMs. A dot would be a good compromise though. So Z17 could be amenity dot level, z19 could amenity icon level.

kocio-pl:
Thanks for bringing more precise idea of what deprioritizing might mean.

I have a problem with amenity dot - this is kind of catch-all category. When I see violet dot, I know it's a shop, when I see blue dot, I know it's an office, but amenity can be a lot of different things for different purposes.

dieterdreist:
sent from a phone

> On 14. Sep 2018, at 23:08, kocio-pl <notifications@github.com> wrote:
> 
> I have a problem with amenity dot - this is kind of catch-all category. When I see violet dot, I know it's a shop, when I see blue dot, I know it's an office, but amenity can be a lot of different things for different purposes.


I agree. The dot does not make a lot of sense to me for amenity brown, I’d prefer to see a label rather than a dot. Unless we generally use a lot more dots and it becomes kind of a density map in certain zoom levels.

Adamant36:
Good points

kocio-pl:
> My "systematic" rule would be: the rendering level and priority level of POIs should be based on the POI's cartographic importance for a general-purpose map, determinded by consensus of a super-majority of style contributors, mappers and map users. So any change needs to be discussed and agreed upon.

Thanks for giving specific proposition, it helps me a lot when the criticism is constructive. 

I was thinking about it lately. It's not specific to general-purpose map, what you have proposed is essentially voting everything. Voting is used to show what people like when objective metric is hard to find or when measuring is not that important. One doesn't vote mathematics.

I think as the amount of objects in database is raising (and ~2 times in 3 years, like in this example, means "fast rising") finding objective rules is getting more and more important. I'm afraid of the following problems fueled by this trend:
- importance wars: there will be more politics used as a rule, because there will be more problems
- changing circumstances: if the decision looks OK for now, it might look worse 2 or 5 years later, so they will have to be repeated in the future

There are also inherent problems with voting paradigm that are independent from database size:
- lack of consistency: maybe X will be voted as more important than Y, and Y more important than Z, but Z might be voted as more important than X in a separate discussion 
- debugging problems: it would be hard to understand why it looks like this without checking each vote

Your proposition also requires new, much harder kind of voting, which can paralyze this project even faster than consensus rule almost did 2 years ago.

I believe universal map style deserves stable, long-term metric, which is also universal. There is enough more specific problems to solve, like colors, categorization, unclear tagging definitions, designing icons, writing code etc. This is where voting might be used as a support to make decisions. But if we have something more objective and independent of current struggles, the support it gives is much stronger.

jeisenbe:
Those are good points about the problems caused by too much voting,
@kocio-pl

I agree that not everything should be brought up to a vote. Most issues and
PRs should go through after discussion without an specific voting.
I was focused more on the example of where some maintainers oppose an idea,
but other maintainers support it. In this case, there should be some
attempt to find a consensus or at least a majority.

Maintainers should be the ones to make the final decision. As someone else
pointed out, anyone could make sock-puppet accounts or drum up outside
supporters to "stuff the ballot" if votes were counted. So expert opinion
and the fact that the maintainers are involved long-term are more important
than overall popularity.

But it doesn't make sense for 1 maintainer to merge a PR if almost all the
others oppose it, in particular.

kocio-pl:
We are discussing voting and governance problems again. They are related to this problem, of course, but what I'm most interested in is hearing what do you think about using universal rules when this is perfectly possible?

jeisenbe:
Universal rules are easy to make up and easy to apply. But do they make a
good map that fulfills the style guidelines?

I'm not sure that any set of hard rules will work for such a challenging
style as this. If we were just making a good general-purpose map, it might
work. If we were just making a map for use in JOSM by mappers, it could
work. But we are trying to do both of those things and more.

Specifically, I think a rule about the "size" of features is only useful
for comparing things that are otherwise the same. A big woodland is more
important cartographically than a small woodland, sure. A tall bulding is
more important than a 1 floor building, generally.

But a big warehouse is less important than a small hospital.
And a 1 floor supermarket may be more important than a 4 floor office
building that is objectively "bigger".

Adamant36:
@jeisenbe, I'm not sure how leaving ultimate say up to the majority of maintainers is any less prone to stacking a vote through any other way. Its not like a maintainer can't easily get three other maintainers on his side if need be, like with a quid pro qoue of "I'll scratch your back, if you scratch mine.' Or the person could just promote lower people to maintainer status. So person will then do their bidding in return for giving them the increased status. I wont say who or why, but I've had multiple people message me about similar things before on OSM. Even with my low social standing. So I know it happens. Especially with a small project of this size that actually has a lot of social/political capital under the surface that can be easily used as potential leverage to push an agenda. Plus, most people are stubborn and just want what they want. 

So, ultimately, there shouldn't be one overriding metric that decides things. Otherwise, people will figure out a way to exploit it. Whatever it is. Yes, the voice of the maintainers should be considered, but only as one factor of many. Instead, the more safe guards against special interests, factions, or just pointless arguing the better. What Kocio-pl is coming up with is a partial attempt to mitigate that stuff. Its not an ultimate answer, but it would help. The more standards that can be refereed to in cases where egos are getting in the way of progress the better. Guidelines also act as a buffer against potential corruption, help things to not be as personal, and can introduce a bit of accountability if done right, which isn't a bad thing. Also, I think an imperfect guideline is better then none at all. Especially since they can be modified and improved once they are implemented if need be. It would also help save a lot of time that is currently being used on extremely redundant conversations about details that are always being re litigated, even if they end up the same. Which would mean more time and energy for coding etc instead.

kocio-pl:
Thanks for your remarks. It helps me to fill the blank spaces I haven't thought of before.

> Universal rules are easy to make up and easy to apply. 

I agree that they might (and should) be easy to apply, but making up is much harder. I mean, if they have to be solid. It was easy for me to show the weakness of 3 proposed rules, because they were also subjective. And it took me many months to check if size rule is really worth using.

> But do they make a good map that fulfills the style guidelines? 

General purposes are not touched by size rule, the rest is just slightly related.

Goals:
- Legibility and clarity - more clarity
- Being understandable and supportive for mappers - easier to "debug" (probably the same as clarity)
- Diversity - not touched
- A rich map - it helps to make one if you have clean table, but not touched directly
- Maintainability - it's easier to maintain a project with less discussions (not about the code though)
- Adaptability and ease of use - not touched

> I'm not sure that any set of hard rules will work for such a challenging style as this. If we were just making a good general-purpose map, it might work. If we were just making a map for use in JOSM by mappers, it could work. But we are trying to do both of those things and more. 

It's not the only rule and I said it. When we have plenty of space, we can exaggerate things:
- we can show things earlier if they exist in the outdoor, because there will be never too many objects, even when they are fully mapped (but still relative size is important)
- we can follow standard practice of showing capitals with countries on low zoom levels

But this is when things are easy. Size rule become handy:
- to start with something sane
- if there are no other rules and people disagree what is important.

> Specifically, I think a rule about the "size" of features is only useful for comparing things that are otherwise the same. A big woodland is more important cartographically than a small woodland, sure. A tall bulding is more important than a 1 floor building, generally. 

In my opinion this is where size rule shows its strength: comparing otherwise uncomparable objects. It scales from trivial cases (but in trivial cases any rule might work) to the bunch of different types on z17 in the middle of a well mapped big city.

> But a big warehouse is less important than a small hospital. And a 1 floor supermarket may be more important than a 4 floor office building that is objectively "bigger".

The hidden question when dealing with importance is - "for whom" and "for what purpose?". It works relatively good for special styles, but:

_It's used in many map applications as a general purpose map._

Hospital is important for tourists probably, but for a transport map it's not. And for logistic app any warehouse might be more important than hospital. And we serve them all.

kocio-pl:
More ATM related things I discovered lately:

> In this situation, the ATM icon is more useful than the bank icon, even though the bank is larger.

@jeisenbe ATMs being part of banks are so common, that JOSM preset includes ATM presence (`atm=yes`) as default. So if you see bank icon, you can assume ATM is here, but when you see ATM, you can't assume the bank is here or not. Thus I think ATM icon is less useful thank bank icon in general.

> example where restaurants and cafes clutter the map at the same zoom level

@dieterdreist Clutter is always unwanted side effect, no matter what is cluttering map. But I see a difference here: gastronomy objects belong to this zoom level and it's just a dense area, while ATMs in Hanoi give impression that this area _at this scale_ is quite well mapped, while in fact it's not. They are just hiding the fact that there are not too many bigger objects tagged. I consider feedback loop claiming that it's well mapped city area to be a problem, not help.

[Hanoi, z17](https://www.openstreetmap.org/#map=17/21.01597/105.85068)

with ATMs
parcwpzv

without ATMs
se79tr61

kocio-pl:
Since this debate seems to be heated, I took extra effort to re-read historical discussion in #1884. 

I have found statements that every rule is subjective - this time I was able to show that some rules are more subjective than the others. 

I was also able to demonstrate this time how hard is it in fact to make such rules and in general what are the exact problems with using "importance" at all, which was not what I have analyzed before. 

The hidden questions "for whom" or "for what purpose" has been raised this time, which do not exist for a size rule.

In the historical discussion I have asked what an "importance rating as a technical term" means and that has been never even replied. I believe it can't be any more technical than size rule - as shown before every use of this word is subjective.

I was also able to show now that not only size rule is less subjective than many other rules, but also applying size rule is simple in the ATM case using relative size rule: there are ATMs inside bank or shop, but never the other way around.

In general I find historical discussion as less argument driven from both sides. It took years to test the size rule, but now it looks stronger than my initial intuition.

dieterdreist:
sent from a phone

> On 30. Sep 2018, at 01:34, kocio-pl <notifications@github.com> wrote:
> 
> I was also able to show now that not only size rule is less subjective than many other rules, but also applying size rule is simple in the ATM case using relative size rule: there are ATMs inside bank or shop, but never the other way around.
> 



you did not show though, that the size rule is relevant. There were people arguing that banks are less interesting to show than atms, for example, and it wasn’t actually rebutted

kocio-pl:
So which rule is most relevant for universal style? Which one is more universal in your opinion?

dieterdreist:
sent from a phone

> On 30. Sep 2018, at 19:10, kocio-pl <notifications@github.com> wrote:
> 
> So which rule is most relevant for universal style? Which one is more universal in your opinion?


IMHO there is no single rule that lets us decide everything, every issue has to be judged on an individual basis

matthijsmelissen:
> I believe size rule is good way to go

An alternative rule would be to prioritize objects based on how often they are used. This is of course hard to count exactly, but we could make an estimation. This would increase the prominence of ATMs, because ATMs are used quite often even though they are small. How would you feel about such rule?

dieterdreist:
Am Mo., 1. Okt. 2018 um 10:40 Uhr schrieb Matthijs Melissen <
notifications@github.com>:

> I believe size rule is good way to go
>
> An alternative rule would be to prioritize objects based on how often they
> are used. This is of course hard to count exactly, but we could make an
> estimation. This would increase the prominence of ATMs, because ATMs are
> used quite often even though they are small. How would you feel about such
> rule?
>


similarly, it could be an indicator to look at how big is the potential
userbase (not necessarily in absolute terms, which is hard to do, but also
"subjectively" relative, i.e. we guesstimate it). ATMs can be used 24/7
(typically) and by everybody (who has a card of one of the
circuits/networks that the atm supports), while a bank has typically very
limited opening hours and is useful only for people who hold an acocunt at
this specific branch.

We could also look at how useful we estimate showing the thing on a map.
Usually, maps are more important for tourists than for locals, because
locals know much more often without a map where the things are that they
need, especially if it is the branch they have their bank account at. Of
course, things on a map also serve for general orientation, i.e. it could
help you to see something on a map even if you don't want to go there, and
this would be a point in favor of prioritizing banks, because they will be
more visibly prominent than atms when you stand in the street.

The way the "size rule" is presented currently seems that it also useful
for comparing apples with pears. Maybe it is not intentional but zhe way it
is written, to me it suggests we should give priority to a slurry tank over
an atm, because the former is bigger. This is just nonsense.

jeisenbe:
Another consideration is that the current prominence rules are not size
based. For example, “major buildings” are shown mire prominently- like
train stations, churches, mosques, synagogues, airports etc

This is reasonable because these places are community landmarks and are
also public places that people are likely to look for. These
characteristics do not relate to size.

We are also currently considering deemphasizing minor buildings like
garages, based on their use and importance, not due to their size. This is
reasonable.

My folks live in the county, where the sheds and barns are often bigger
than the house, but certainly the house is more important. When there is a
forest fire, they count the outbuildings and houses burned separately even
though the outbuildings may be bigger

kocio-pl:
First of all thanks for all the comments. Sorry for not responding to all the problems this time - there are too many things and I will try to reply step by step:

> IMHO there is no single rule that lets us decide everything, 

Sure. There is not and I have said already that I don't propose to remove all the others, so it might be a straw man argument. I just think this is the best rule for universal style.

> every issue has to be judged on an individual basis

For such a rich and still growing database I believe deciding case by case (=effectively no rules) will not work good. All the downsides of "vote everything" idea (including rising lack of coherency between multiple elements) apply.

> Another consideration is that the current prominence rules are not size based.

Well, I don't propose anything new. Some of them are (we never show cities before countries, for example), some of them aren't. Whenever we have not too many elements, we can exaggerate (like showing capitals early or nearly all outdoor-specific objects) and we can modify colors like you said (or other attributes, like showing dots for shops initially). We will still have a lot of room for choosing and voting.

> My folks live in the county, where the sheds and barns are often bigger than the house, but certainly the house is more important. When there is a forest fire, they count the outbuildings and houses burned separately even though the outbuildings may be bigger

"More important" - OK, but but for whom and for what purpose? There is no universal meaning of "importance", free from the context. You just gave the reason for only one purpose. But it's very specific and sounds to me like HOT task, which has it's own very specific map style, and not for a general map we try to produce. We could modify shade or pattern for less important buildings - see #2532 (PR for which was rejected). But not showing big things before smaller ones is unpleasantly surprising idea for me - for the overview of this area hiding big, visible things does not work.

> The way the "size rule" is presented currently seems that it also useful for comparing apples with pears. Maybe it is not intentional but zhe way it is written, to me it suggests we should give priority to a slurry tank over an atm, because the former is bigger.

My English is too weak here, and so is my farmland knowledge, I don't know what is slurry tank - is it https://en.wikipedia.org/wiki/Slurry_pit ? 

If it's bigger, it should be rendered earlier. Still I can't imagine how does it look like... However I have very similar (also stinking...) analogy - we already show landfills from z10, which is much earlier than ATMs. Is there any problem for you? If yes, what do you want to do with it?

kocio-pl:
> They often serve a whole rural district, town or city and are well-known landmarks due to their function, not just due to their size.

They wouldn't be landmarks if they were just few meters big. It would be (probably) the case of slurry tank than.

AFAIK you can't even get into big landfill, at least in Poland, so for me personally it's less important than ATM. But for the local community it's the other way around. So importance is hard to get and compare and we're serving map style for all uses.

What if big landfill is operated by single entity, which for example imports trash from other countries and doesn't serve community at all? I know such places exist. Would you not show them early enough just because they are not useful?

kocio-pl:
I don't like to go into details of rendering private objects now (it's a problem for few separate issue tickets IMO), especially since I still haven't responded to some other questions. My point is that operator, access, relative usefulness or ugliness doesn't matter - if something is big enough, it will be always visible much earlier than ATMs.

dieterdreist:
Am Di., 2. Okt. 2018 um 02:01 Uhr schrieb kocio-pl <notifications@github.com
>:

> The way the "size rule" is presented currently seems that it also useful
> for comparing apples with pears. Maybe it is not intentional but zhe way it
> is written, to me it suggests we should give priority to a slurry tank over
> an atm, because the former is bigger.
>
> My English is too weak here, and so is my farmland knowledge, I don't know
> what is slurry tank - is it https://en.wikipedia.org/wiki/Slurry_pit ?
>



exactly, as the article states, "also known as slurry tank". These can come
in different shapes and structures, but the function is the same (storing
animal pee). Here's an example how they may look like:
https://de.wikipedia.org/wiki/G%C3%BCllegrube#/media/File:Don%27t_jump;_Haddon_House_Farm,_Bakewell_-_geograph.org.uk_-_1264500.jpg
or
https://www.picclickimg.com/d/l400/pict/202165400029_/Ehemaliger-Jauchetank-G%C3%BClletank-jetzt-Wassertank-2200Ltr-feuerverzinkt.jpg

these are objects, like atms, and unlike landfills or other landuse.

kocio-pl:
> An alternative rule would be to prioritize objects based on how often they are used. 

Thanks for another example of rule that could be useful, I'm glad that we're testing alternatives.

This one looks quite promising, since it gets away from the "importance wars" and deals with numbers, but is also not universal - it's not only harder to measure than size and can change over time, it's also very limited in scope. We might estimate the usage of ATMs, banks or shops at given moment, but there's no real equivalent for most memorials, towers or artworks for example. Niagara Falls might have some "userbase", but most other waterfalls just don't have any.

> How would you feel about such rule?

I think this is another nice helper rule, so it's good to have it in our toolset. It's more universal than just voting, but size still looks to me as the most universal rule for the map visualization.

> Usually, maps are more important for tourists than for locals

You just gave the example of local authorities needs, but there can be much more uses you don't even think of. It's not our job to dictate which usage is more important. I understand why it's so hard to not look from the specific points of view, but one ends up with switching them and lack of tools to compare them. 

As OSM is still growing in all the directions, we have more different consumers and I think it's time to make "universal" style really generic. We used to be "street map" - but it was just the beginning. We even had the UK roads style for many years (until 2015), because it was London-based project, but IMO as we get global it's getting more important to look for the most universal rules for default map style.

kocio-pl:
There were some arguments about the past - that we should probably trust in the original design reasons (even if we don't know them), that density is not that big problem, because this is only about tagging new places, that it's all the fault of adding new icons etc. 

Here is simple comparison of 2012 and 2018 data in the center of London with current style. Original ideas, which come from some time before 2012 (OSM Carto did not made a big departure from them), work reasonably good with smaller set of data, but look horrible with more of them (which was hard to predict, of course - it's also nobody's fault that OSM grows so fast). There is much bigger POI density even if London was probably mapped the best from the beginning, so this is not only about mapping new places. And there's only few new icons, much of the clutter here comes from adding more gastronomy and shops POIs. Small amenities are part of a problem now, even if we'll make orange dots on z17:

London center in 2012 (from the last CC-BY-SA licensed planet file)

oecwrnln

London center currently

nn903zm5

jeisenbe:
It looks like the current data includes many more offices, shops and
restaurants. The house/building address numbers also add to the crowding in
2018. The eating places are most noticeable since the shops and offices are
just dots.

But I only see 3 ATMs on each version of the map. They don’t seem to be the
problem?

It would be great if someone could make a code that would render the map
differently in crowded urban areas vs sparse rural areas. Maybe it could
also fix the difference in zoom levels vs map projection at 60 degrees
north vs the equator.

-Joseph

kocio-pl:
All small amenities are part of the problem. I see 9 ATMs there and 4 of them are eclipsing roads. There are also phones and post boxes also eclipsing roads. This is not obvious until gastronomy is dotted, but roads have already some related features (like names, arrows, traffic lights, some bus stops or bike rentals) and phones or ATMs do not belong there.

kocio-pl:
> It would be great if someone could make a code that would render the map differently in crowded urban areas vs sparse rural areas. 

Yes, this looks like a very important problem, but still we have no clue how to fix it: #1957.

Adamant36:
@jeisenbe, there is a discussion about it in issue #1957. I also brought up a similar idea in the gastronomy color change issue based on population density. Which I was reading an article about and could be totally doable. Both solutions could introduce their own issues though, like slowed down rendering speed due to the added code complexity or it still being interpretive. Plus, it would have to be thoroughly tested. Id imagine by the time the code got written and it was tested properly there would be vector tiles implemented on the main site that would allow for a better layers selection system, which would solve most of these types of issues. Like there is an interactivity function in the project file that would allow for informational popups on the map (at least I think that's what it does), but it slows things down to almost not be worth it from my understanding. Which is why the main CartoCSS maps don't use it, I think. Not to say something like that or your idea isn't worth trying to implement, but there would be technological hurdles etc and it might not be worth the time if something better is coming along soonish. Its kind of a similar issue with showing the surface of roads. There's one map that I know of that shows road surfaces and it renders way slowly compared to this one. Plus, it only shows it in a very small area. That's all just my opinion though.

kocio-pl:
Thanks all for the discussion and showing different points of view, it was good to inspect them to better understand the issue, which is more general than only ATMs.

It looks to me now that any systematic approach to cleaning rendering needs to regard size differences to achieve really universal and generic OSM map style. Other rules (like private/public differentiation) can be still useful, but they are secondary tools.

I also see the need to apply such cleaning, because database growth seriously changes the look of the map, even if the style hasn't changed for years in some aspects. I think ignoring such a big thing makes the resulting map gradually worse over the years.

matthijsmelissen:
> It looks to me now that any systematic approach to cleaning rendering needs to regard size differences to achieve really universal and generic OSM map style. 

For the record, I do not agree with this statement.

But from a community perspective, I think it's important that decisions are made by the people who actually do the work, so I'm happy you went through with this proposal.

kocio-pl:
> rejection of the evidence and argument based discourse

This is sad for me to see such statement. I find it so unfair that it's hard to believe you wrote it.

I took a lot of time and effort to look at multiple arguments, think about them and respond with my arguments and showing proofs. Everyone is free to have their own position and not like them. But they are here - you even took a small chance to respond and made your arguments against. And now you sound like there was just vague comments, probably. I took discussion very seriously. Not agreeing with arguments is one thing, rejecting their existence supports lack of discussion at all (why try, when I can simply wait and dismiss everything at the end?) and this is dangerous.

What I think is wrong, is discussion without the outcome, as we had before. But demoting active discussion (in any way) is also wrong. I believe in active discussing, deciding and then implementing it. Each step requires activity.

> I am also kind of glad about this outcome because it marks a clear end to any remainders of the cartographic consensus that formerly existed in the development of this style.

Nice to have it, but it's not something that just magically happens. What is your proposition to reach consensus then?

> For the record, I do not agree with this statement.

I see. What is your position and suggestion then?

PR #3373, opened 2018-08-31
Organize icons in key folders

Adamant36:
This PR adds folders for most icons that were not in ones already, and puts a few icons in folders that were already there. It also shortens the names of a few icons. It was tested in Kosmtik and seems to run fine. Hopefully this will help with keeping track of icons well doing PRs etc and also makes the symbols folder more organized overall.


Tomasz-W:
Icons from "towers" folder can be moved to "man_made" folder as they are all based on `man_made=tower` tag

Adamant36:
@Tomasz-W, done. 

@kocio-pl I'm glad its OK. The state of the Symbols folder has been bugging me for a minute now. Hopefully there wont be any issues on your end with the icons not being visible. If there is though, I'll fix it.

Adamant36:
@kocio-pl, would it be easier to test/make things go faster if I split things up into different PRs per folder or something instead of changing them all at once?

kocio-pl:
I don't think so, current code is OK for me.

kocio-pl:
I have checked that everything seems to be OK up to this point:

https://github.com/gravitystorm/openstreetmap-carto/pull/3373/files#diff-5d9237083037cf1d6d91c329a0ff61adL567

I will continue checking.

Adamant36:
Awesome. I'm glad to know it went OK. Thanks for merging it.

kocio-pl:
I think there are more things that could be pushed into folders, for example mud or waterfall belong to natural and one way arrow belongs to highway.

Tomasz-W:
My suggestions:
* as we have a mess with natural/ landuse tags for different landcover features, I propose to throw all patterns into "landcover" folder
* as "shields" and "unpaved" folders are for roads, maybe they should be moved to a subfolders of "highway"
* make "barrier" folder for barriers

jragusa:
In addition to the suggestions of @Tomasz-W, some others icons could be moved and some of them should be also renamed for clarity:
* `cliff`, `cliff2`, `spring` and `waterfall` to natural folder
* `motorcycle` to shop folder
* `bunker`, `advertising_column` and `embankment` to man_made ?
* `helipad16` should be be renamed to helipad and moved in an aerialway folder together with `aerodrome.12` (idem renamed to aerodrome).
* `square` should be renamed to station and moved to railway together with `level_crossing` and `level_crossing2`.
* please also note that `square` is also use for power_poles and need to have a dedicated icon which would be moved to a power folder together with `power_wind`.
* `entrance.10`, `rect`, `rectdiag`, `corners` should be renamed to entrance_subway, entrance, entrance_no, entrance_service, respectively, and moved to an entrance or building folder
* `oneway` and `oneway-reverse` to highway folder

PR #3374, opened 2018-09-01
Render all road links thinner

LaszloEr:
Currently trunk_link, primary_link, secondary_link and tertiary_link are all rendered identically to their main road types. These changes let them them render thinner.
Test renderings with links to the example places:

**primary_link before**
https://www.openstreetmap.org/#map=18/47.45893/18.98951
2o

**After**

2n

**trunk_link before:**
https://www.openstreetmap.org/#map=17/47.46131/18.87792
4o

**After**
4n

**secondary_link and tertiary_link before:**
https://www.openstreetmap.org/#map=18/47.46888/19.08032
3o

**After**
3n

### DIFF
--- a/roads.mss
+++ b/roads.mss
@@ -96,7 +96,6 @@
 @secondary-width-z12:             3.5;
 @secondary-link-width-z12:        1.5;
 @tertiary-width-z12:              2.5;
-@tertiary-link-width-z12:         1.5;
 @residential-width-z12:           0.5;
 @unclassified-width-z12:          0.8;
 
@@ -109,7 +108,6 @@
 @secondary-width-z13:             5;
 @secondary-link-width-z13:        4;
 @tertiary-width-z13:              4;
-@tertiary-link-width-z13:         3;
 @residential-width-z13:           2.5;
 @service-width-z13:               0.8;
 @living-street-width-z13:         2;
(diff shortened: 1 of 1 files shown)


kocio-pl:
Thanks for the proposition! I like it and it makes sense for me. 

It works good for color links and gives nice additional feedback if tagging is proper, but I would like to look closer at white links (tertiary) and compare them with some smaller types of roads.

kocio-pl:
@LaszloEr Could you test if white links are OK? Do they look good in comparison with residential or service roads?

LaszloEr:
No, as you said they look almost identical and I'm not sure how I could resolve this. Maybe it would be best to not change the tertiary (white) links at all for now?

kocio-pl:
For me it would be good enough to improve just the colored ones, since tertiary link is also less frequent than all the other links:

taghistory 40

LaszloEr:
Here's a version with no changes to tertiary links. Preview:
ggggg

kocio-pl:
Related to #1109 - probably closes it.

roads.mss
```diff
@@ -125,8 +131,11 @@
 @motorway-width-z15:             10;
 @motorway-link-width-z15:         7.8;
 @trunk-width-z15:                10;
+@trunk-link-width-z15:          7.8;
 @primary-width-z15:              10;
+@primary-link-width-z15:        7.8;
 @secondary-width-z15:             9;
+@secondary-link-width-z15:        7;
```
kocio-pl: Why 7 here instead of 7.8 as all the others?

LaszloEr: I just made it a bit smaller because secondary-with is also a bit smaller (9) than the other bigger road types above it (10).

kocio-pl:
Thanks for this tuning. There are some strange looking parts when links joining with roads, because width is different, but this is small effect, it helps make tagging better and also show roads design more accurately.

PR #3377, opened 2018-09-02
Move manors and castles to z16 and above

### LINKED ISSUE #3325: Move manors to higher zoom level
Curently both `historic=manor` and `historic=castle` are rendered from z15. It makes them too prominent and looking more important than they actually are. Examples:

manors:
https://www.openstreetmap.org/way/273470884#map=15/52.3669/16.9812
https://www.openstreetmap.org/way/273470883#map=15/52.3737/16.8508
https://www.openstreetmap.org/way/238038057#map=15/52.4118/16.9404
https://www.openstreetmap.org/node/5355328416#map=15/53.0894/16.1363
castles:
https://www.openstreetmap.org/relation/1483520#map=15/52.4077/16.9189
https://www.openstreetmap.org/way/274284937#map=15/52.9427/17.3060
https://www.openstreetmap.org/way/525949706#map=15/52.5731/17.0071

I propose to move them to z16 like eg. forts. Those buildings was the most important thing in cities hundreds of years ago. Now they are jus one of a tourist attractions, so they shouldn't be that prominent. 

I would consider moving an `amenity=townhall` to z15 instead.

Adamant36:
Fixes #3325 

This PR moves manors and castles to z16 and above. Although I did not take any pictures of it being test rendered, I did test it on my end and it worked as expected. If I need to provide rendering shots though I will, but it seems kind of pointless in this case.


kocio-pl:
Did you see #3352?

Adamant36:
Uuuummm, no. My bad. I have no clue how I missed it. Sorry.

PR #3387, opened 2018-09-11
Fix alignment in communications_tower

jragusa:
Remove artefact in icon of `man_made=communications_tower`

https://www.openstreetmap.org/node/1163258826

Before
communication_tower_before

After
communication_tower_after


kocio-pl:
Thanks for the fast reaction!

PR #3399, opened 2018-09-16
Unpaved rendering

### LINKED ISSUE #110: Render paved/unpaved

sommerluk:
Fixes #110

Closes #4137 (alternative version)

That’s essentially the same code as #3357 with the following differences:
- No special unpaved rendering for highway=platform and railway=platform. (There was a minor bug in #3357 rendering in wrong colour for railway=platform on bridges with layers ≠ 0 if there is no corresponding layer also in the “highway” part. Thought this case is rare, the fix in the SQL code could be a little complicate, so for the moment I’ve dropped it for “platform”.)
- Some small performance tuning in the SQL query for bridges. The bridges query (at difference to the roads-low-zoom or roads-fill query) can potentially be rendered various times: once for each layer=*. Together with the fact that it has yet a SELECT DISTINCT clause, it was my preferred starting point, as here with little change we might get better performance.

I’ve tested this locally, trying to eliminate influences of other programs running in the background. I got know the same rendering time as without unpaved rendering (while the old unpaved rendering took 10% more time). @rrzefox Could you take a look at the performance here also?


project.mml
```diff
@@ -1070,10 +1144,49 @@ Layer:
                   ELSE 'no'
                 END AS link,
                 COALESCE(layer,0) AS layernotnull,
-                z_order
+                z_order,
+                CASE
+                  WHEN highway IN ('service', 'tertiary_link') THEN 'tertiary_link'
+                  WHEN highway IN ('residential', 'unclassified', 'tertiary') THEN 'white'
+                  ELSE highway
+                END AS int_roadcolor
```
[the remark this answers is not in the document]
sommerluk: > Do you have indication that this actually improves performance in a meaningful way?

Well, the starting point was [the comment of talaj](https://github.com/gravitystorm/openstreetmap-carto/pull/3357#issuecomment-419658334) who explained that likely `dst-out` is no performance problem, but `dst-over` is. Furthermore, [he said that `dst-over` isn’t executed when there is no actual geometry that matches.](https://github.com/gravitystorm/openstreetmap-carto/pull/3357#issuecomment-419700435) Based on this, with this code I get the same performance on my local environment as without unpaved rendering. With the old code I get 10% slower performance. This is likely not representative. The last time I made I measured local performance, the results of @rrzefox where different from mine, and I suppose his results are more close to performance on production servers.

In short: There are some weak indications, but nothing of which I could say it’s representative.

> this makes adapting the roads code to different designs - like with tertiary roads in a distinct colour - significantly more complex.

Indeed. I share your concerns. If we get to the conclusion that we can live with the performance without this SQL change, than I prefer not doing the SQL change for exactly the reason you mentioned.

imagico:
Have you tried limiting the number of dst-over operations also in the roads-fill layer and not only in the bridges layer?  If this has a significant influence on performance it would be way more significant to do this on the road fills than just on the bridges.

Also checking if a performance difference is due to the bridges layer or the roads-fill layer is relatively easy since both layers exist in both versions of the code so you can simply disable the bridges layer in both and check if there still is a performance difference without it.

sommerluk:
screenshot_20180916_195842

In the underlying Python script I call the Mapnik rendering now various times instead of a single time, so with more overall time, perhaps the measurement becomes better.

It shows that [with full SQL tuning (also for roads-fill, as you proposed)](https://github.com/sommerluk/openstreetmap-carto/tree/unpaved25) it is indeed faster.

sommerluk:
Added the full SQL tuning to this branch.

project.mml
```diff
@@ -880,12 +893,51 @@ Layer:
                   ELSE 'no'
                 END AS link,
                 COALESCE(layer,0) AS layernotnull,
-                z_order
+                z_order,
+                CASE
+                  WHEN highway IN ('service', 'tertiary_link') THEN 'tertiary_link'
+                  WHEN highway IN ('residential', 'unclassified', 'tertiary') THEN 'white'
+                  ELSE highway
```
HolgerJeromin: Mixing "color" white and "class" tertiary_link feels odd.

sommerluk: I agree, it’s indeed ugly.

kocio-pl:
As v4.15.0 is near, I guess this code is not yet ready for merging?

We have to wait some more until https://github.com/mapnik/mapnik/pull/3980#issuecomment-422550474 will be usable for us, because we depend on multiple other code releases and packages, but it looks like proper solution is on the way.

rrzefox:
> @rrzefox Could you take a look at the performance here also?

I've deployed this on Friday, so excluding the weekend, I now have 3 days of stats:
```
    |         OLD  |          OLD |         OLD  |         NEW  |         NEW  |         NEW
  z |  2018-08-28  |  2018-08-29  |  2018-08-30  |  2018-09-24  |  2018-09-25  |  2018-09-26
 19 |        1.66  |        1.48  |        1.50  |        1.69  |        1.51  |        1.64
 18 |        1.44  |        1.56  |        1.60  |        1.85  |        1.69  |        1.78
 17 |        1.58  |        1.79  |        1.82  |        2.24  |        1.85  |        2.01
 16 |        2.34  |        2.37  |        2.53  |        3.06  |        2.74  |        2.63
 15 |        3.51  |        3.58  |        3.94  |        4.65  |        4.10  |        4.32
```
This does look fine for me, the difference is within the margin of error (remember, this is just average over all tiles rendered on a day, so fluctuations are to be expected because it's not the same tiles every day).

kocio-pl:
Just a quick status check: https://github.com/mapnik/mapnik/pull/4004 has been merged, so the next thing we wait for is a new Mapnik v3.0.x release (probably 3.0.22), then update of mapnik-reference package (https://github.com/mapnik/mapnik-reference/pull/153) and Kosmtik update, but also making Debian and Ubuntu Mapnik package and deployment of new Mapnik package on OSMF servers. New Kosmtik release might be also needed.

kocio-pl:
Oh, and CartoCSS probably will need a new release together with new npm package (8.10.0)  - https://github.com/mapnik/mapnik/pull/4004#issuecomment-434016216.

rrzefox:
Just a heads up: this is now no longer deployed on my server because it wouldn't apply on top of 4.17.0.

kocio-pl:
There are new Mapnik release (3.0.22), mapnik-reference releases (3.0.22 and npm 8.10.0) and also corresponding CartoCSS releases (1.2.0 and npm 1.2.0 - https://github.com/mapbox/carto/issues/498), what else is missing in our toolchain? Kosmtik update, release and npm release?

Adamant36:
Just because I find it interesting, what in particular from a technical standpoint is it about roads that make rendering surfaces on them slower then rendering them on anything else? also, what's the reason something like dashed casing, similar to how @StyXman says he does it in his style, isn't a possible alternative?

sommerluk:
@Adamant36 See #110, #2640, #3280, #3357

Adamant36:
@sommerluk, thanks. I'll look through them.

StyXman:
@Adamant36 I don't quite like how my style looks (the 'sausages'). I still can't find the time to implement mapnik/mapnik#3982.

sommerluk:
Okay. For some time, I was quite busy and therefor did not work much on this issue. While still being quite busy, I want to give it another try now.

In the meantime some things have changed: The guys from Mapnik have made a [native Mapnik implementation of polygon-like pattern rendering on line geometries](https://github.com/mapnik/mapnik/pull/3980) which has been [backported to v3.0.x](https://github.com/mapnik/mapnik/pull/4004)  and [released with Mapnik 3.0.22](https://mapnik.org/news/release-3.0.22). Also mapnik-reference has been [updated](https://github.com/mapnik/mapnik-reference/pull/153) to 3.0.22. Furthermore a [corresponding new npm package](https://www.npmjs.com/package/mapnik-reference/v/8.10.0) has been published with version number 8.10.0. Also CartoCSS had a [new release 1.2.0](https://github.com/mapbox/carto/releases/tag/v1.2.0) which updates its dependencies to Mapnik 3.0.22 respectively mapnik-reference 8.10.0. This CartoCSS release is also available as npm package [carto 1.2.0](https://www.npmjs.com/package/carto/v/1.2.0). I've opened an [issue](https://github.com/kosmtik/kosmtik/issues/295) requesting a new kosmtik release that uses all this. For openSUSE users, as the [Geo repository](http://download.opensuse.org/repositories/home:/sommerluk:/branches:/Application:/Geo/openSUSE_Leap_15.0/) still has Mapnik 3.0.20, for the moment I provide [RPM packages of Mapnik 3.0.22](https://build.opensuse.org/package/show/home:sommerluk:branches:Application:Geo/mapnik) for openSUSE in my home repository. For Ubuntu, there seem to be plans to integrate Mapnik 3.0.22 [into the next release 19.04](https://launchpad.net/ubuntu/+source/mapnik) (but this is not an LTS release).

The native support of Mapnik for polygon-like patterns on line geometries is really a great news. That fits perfectly our use case (except turning circles). Using it, the code can be much simpler and straightforward and without pitfalls. Given that it's only a matter of time until the new Mapnik version is available in Linux distributions, I'm inclined to close this PR and make a new one based on Mapnik 3.0.22 features. It's a pity because a lot of work had gone into the comp-op solution previously, but the new code would be much cleaner. What do you think about? How could this look like in practice? Waiting until Ubuntu 20.04 will be available on openstreetmap.org production servers?

kocio-pl:
Nice to hear you still focused on this task!

The good news is that there is a special PPA for packages not included in the standard Ubuntu repositories:

https://launchpad.net/~osmadmins/+archive/ubuntu/ppa

It's enough to backport Mapnik packages to Bionic.

sommerluk:
Cross-reference for updating Tilemill to support Carto 1.2.0/Mapnik 3.0.22: https://github.com/tilemill-project/tilemill/issues/2695

pnorman:
Note: If we bump the Mapnik version to 3.0.22, we should do that in a major release, not a minor.

sommerluk:
Okay.

Just for the record: I have updated the [documentation for the comp-op approach](https://github.com/sommerluk/roadpatternrendering) recently, considering the [earlier comment from @imagico](https://github.com/gravitystorm/openstreetmap-carto/pull/2640#issuecomment-399785129) and commenting also the performance issues we had with this approach and the possible solutions. And of course mentioning the now available native Mapnik solution.

I've switched to Kubuntu 19.04 in order to have up-to-date Mapnik version, using Kosmtik current master (seeing some [memory problems](https://github.com/kosmtik/kosmtik/issues/298) in Kosmtik at lower zoom levels, but that's not specific to unpaved rendering).

So: Time to write some actual code for the new Mapnik native support: https://github.com/sommerluk/openstreetmap-carto/tree/unpaved40 has the same pattern as here, but the changes to `roads.mss` and `project.mml` have been written from scratch.

No special code for turning circles because the Mapnik native support is only available on line geometries, and not on point geometries. (We could of course, if really desired, possibly fall back for the comp-op approach for turning circles, or maybe to some SVG marker files.).

sommerluk:
@tomhughes If we would like to switch to Mapnik 3.0.22 and CartoCSS 1.2.0 in the future, what would be a practicable roadmap for the openstreetmap.org servers?

tomhughes:
Well so long as there are mapnik packages we can use you pretty much just have to say when you want it to happen.

jeisenbe:
Is there any update on the status of this? In Indonesia some of the other local mappers have decided to start mapping highway=primary on unpaved roads, which makes a lot of sense based on the road system here (they are the only roads connecting large towns to other towns and cities), but it would be very helpful to distinguish between paved and unpaved roads.

@sommerluk, does the new code work in https://github.com/sommerluk/openstreetmap-carto/tree/unpaved40 with the upgrade to Carto 1.2.0/Mapnik 3.0.22? Any new test renderings available?

Should we be asking openstreetmap.org to upgrade to Carto 1.2.0/Mapnik 3.0.22 in the near future?

sommerluk:
@jeisenbe Sorry for the long delay. Currently I'm quite busy with other stuff. I hope that will get better next year… We have the same situation as in Indonesia also here in West Africa. I understand your point.

> does the new code work in https://github.com/sommerluk/openstreetmap-carto/tree/unpaved40 with the upgrade to Carto 1.2.0/Mapnik 3.0.22?

Yes.

The new code relies on a new feature in Mapnik 3.0.22 which the Mapnik maintainers kindly developed to help us with the rendering of unpaved roads. Not only Mapnik, but also software that depends on Mapnik (Carto, Kosmtik) has to be upgrades. Carto 1.2.0 is okay for the new feature. For Kosmtik, you have to clone the github repository and build it locally (as the last published npm package is too old).

A new version rebased to the current master of openstreetmap-carto is available at https://github.com/sommerluk/openstreetmap-carto/tree/unpaved46 now. Testing is welcome.

> Any new test renderings available?

No. But the rendering should be very much the same: It's the same pattern. The only difference is that it does not include support for special rendering for unpaved turning circles, and it adds support for unpaved platforms also for railways (not only for buses).

> Should we be asking openstreetmap.org to upgrade to Carto 1.2.0/Mapnik 3.0.22 in the near future?

Yes. @tomhughes said that basically there have to be Mapnik packages available for the distribution that is currently running on the openstreetmap servers. For now, this is Ubuntu 18.04. Apparently there are no Mapnik 3.0.22 packages available for Ubuntu 18.04. I've [requested a backport of Mapnik 3.0.22 to Ubuntu 18.04](https://bugs.launchpad.net/bionic-backports/+bug/1839056) now, but I don't know if this will become reality, and I don't know much about Ubuntu packaging process… The alternative is to wait until openstreetmap.org servers upgrade their Ubuntu version (probably 20.04 which means to wait about one year).

Maybe I should clode this PR and open a new one with the new code relying on the Mapnik 3.0.22?

tomhughes:
Your chances of getting an official backport are probably minimal but unless there is some significant technical hurdle there's nothing to stop us doing our own.

kocio-pl:
See also Kosmtik problems: #3717.

tomhughes:
I have a built a backport of 3.0.22 on bionic in our PPA but I haven't deployed it anywhere yet.

Should I be OK updating the servers? or are there changes needed to the style before I do so?

sommerluk:
@tomhughes 

That's great news! Thanks a lot!

I think Mapnik 3.0.22 should be backward compatible, so also the current release of this style should run out-of-the-box without problems. At least on my local machine I did not see any problems.

Apart from Mapnik 3.0.22 we would also need an upgrade to CartoCSS 1.2.0…

tomhughes:
The renderers have actually all been running it for about ten hours and no complaints so far...

tomhughes:
Two of the machines already had carto 1.2.0 and I've just updated the others.

sommerluk:
Great work!

sommerluk:
This PR contains now the code that relies on the native Mapnik support for line patterns. It has been rebased to current master. It contains also yet some documentation updates for the now required Mapnik and Carto versions and a comment about a 5.x release in README.md (because we require a newer rendering stack). For style development, we would have to switch to Kosmtik master. (A side effect might be [not to use index files](https://github.com/mapnik/node-mapnik/issues/917#issuecomment-518267096) in our development environment.)

Opinions? Reviews?

imagico:
I have not forgotten this but i need to update my test environment for it and this *always* tends to break things in one way or the other and then takes several hours of trial and error to get working again.  So i need to find time for it.  Damn node.js...

kocio-pl:
Seeing this feature being so close to being merged, I also curse my broken native testing environment... :disappointed: But maybe this could motivate me to take some time again to restore it.

@sommerluk is it possible that you could publish a modified Docker files containing your testing environment? It would help a lot and maybe we would update our Docker settings to keep up with it.

jeisenbe:
I'm using docker, so this would be important for me.

sommerluk:
I'm sorry I've never used Docker, so I'm not sure how to update these images.

Adamant36:
So I copied @sommerluk's repository, updated everything that I think I was suppose to, and tested it. Everything loaded and worked fine. Except there was no difference in rendering that I could find. It could just be me though. At least that means updating things and rendering it might not cause any major problems for whoever tests it. I plan to put some time into figuring out why I'm not seeing a difference to. Then maybe I can show some actual tests of it.

tomhughes:
This repository is just the web site - it doesn't include any map style or tile rendering support so the tiles it uses for the map are the same ones as the production site.

Adamant36:
Ah-ha. His repository has different code in it though. Hhhmmmm 😕 🤷‍♂

tomhughes:
Oh sorry I misread which ticket this was... There's another docker related ticket that I'm getting comments on currently and I mixed them u. Ignore me. This is the rendering repository.

sommerluk:
@Adamant36 Have you chosen the correct branch within my repository (unpaved24)? What Kosmtik version are you using?

kocio-pl:
@sommerluk IMHO it's dead easy - look at the https://github.com/gravitystorm/openstreetmap-carto/blob/master/Dockerfile, it's just a list of installation steps in a clean way, just a script with a simple, fancy syntax one defines for Docker instead of installing things manually. It starts with a pre-published (usually stripped a lot) image of Ubuntu Bionic and then there are some actions needed to install and run Kosmtik.

There is a good practice to make containers/apps atomic, so database and import script has their own definitions, and the whole is gathered together ("orchestrated") by Docker-Compose (configuration is in https://github.com/gravitystorm/openstreetmap-carto/blob/master/docker-compose.yml ). But you probably don't even need to touch them. 

If you have trouble updating Dockerfile, you can just publish what exact steps are needed to achieve your environment, so someone can translate it into the script.

Just ask if you need some guidance. Knowing Docker, even the basics, is quite useful these days.

Adamant36:
@sommerluk, yeah I did. I did the Kosmtik update per the instructions in the PR to. So I guess it's whatever version that is. Unless it didn't update it for some reason. Is there a command to check what version of it you have installed?

sommerluk:
@Adamant36 Hm, apparently there is no simple `-v` or `--version` parameter for kosmtik…

> IMHO it's dead easy - look at the https://github.com/gravitystorm/openstreetmap-carto/blob/master/Dockerfile

Well, these steps fail for me. I'm using Ubuntu disco. Is it necessary to use bionic? Anyway, I've followed exactly the described steps. When trying to load the database, it complaines about docker deamon not running. So I've done

* install docker from your distribution's package manager. In Ubuntu Bionic, do `sudo apt install docker-compose`, `sudo systemctl start docker`, `sudo systemctl enable docker`, `sudo groupadd docker`, `sudo usermod -aG docker $USER`. Than log out and log back in again to apply the groups

Now the docker deamon seams to run. But when doing `docker-compose up import` I get:

```
Starting openstreetmap-carto_db_1 ... error

ERROR: for openstreetmap-carto_db_1  Cannot start service db: b'driver failed programming external connectivity on endpoint openstreetmap-carto_db_1 (467caf30d3258b60e7e4dc2a1644878428d63015fa4720a4e07056c15c645a02): Error starting userland proxy: listen tcp 127.0.0.1:5432: bind: address already in use'

ERROR: for db  Cannot start service db: b'driver failed programming external connectivity on endpoint openstreetmap-carto_db_1 (467caf30d3258b60e7e4dc2a1644878428d63015fa4720a4e07056c15c645a02): Error starting userland proxy: listen tcp 127.0.0.1:5432: bind: address already in use'
ERROR: Encountered errors while bringing up the project.
```

Adamant36:
@sommerluk, hhmmm weird. As far as your problem, sounds like a port issue. Which is why it's listing a tcp address. I had that happen once. You have to restart Docker. That's what worked when it happened to me at least. The solution I used is mentioned [here](https://github.com/docker/compose/issues/4126#issuecomment-456724815).

kocio-pl:
Looks like you have native PostgreSQL running and we assume that it is not. We should probably change this port or at least give a proper warning.

kocio-pl:
> I'm using Ubuntu disco. Is it necessary to use bionic?

Bionic is stable, so we prefer having it, but if it would stop us from having full support for all the implemented functions, I would vote for updating it.

I also use Disco and I was not able to run Kosmtik natively after upgrading Ubuntu, so this ounds promising.

Could you just give the steps needed to have your Kosmtik environment on Disco? Maybe it could be backported, but Docker allows to not care that much for what is run inside the containers.

kocio-pl:
Oh, I'm not sure if you meant the system inside the container (I just answered it) or the host OS? Any host will be OK, they will share the kernel (this is why the speed on Linux is like native), but the rest is independent and you can run any distro inside.

pnorman:
We discussed this at SOTM this weekend. I'm of the opinion that the only way we can get the cartographic "space" to render unpaved surfaces is to drop something else, like access restriction rendering.


We had some discussion on if access restrictions are more important than surface type. In western Europe and coastal US they are, but we could use feedback from people where unpaved roads are common, and if that information is more important than access restrictions.

imagico:
Another possibility we considered in the discussion was dropping the color distinction of road importance class rendering (trunk to unclassified) in favor of surface.  This seems unlikely to get sufficient support because it would be a very radical change but i none the less want to mention it here - the idea would be to use a three color scheme for all roads - motorway, paved and unpaved - and to distinguish the non-physical road classes only by line width and starting zoom level.

jeisenbe:
> if access restrictions are more important than surface type. In western Europe and coastal US they are

Speaking for coastal USA: Access restrictions are more important on short residential roads and short service roads in rural areas, and on other minor roads in urban areas, but unpaved vs paved is more important for longer-distance roads of all times, especially in rural areas. If you are going to a destination on a service or residential road, it's ok if the surface is unpaved for that last kilometer, but you won't want to travel 20 kilometers on an unpaved road, unless that is what you are looking for in particular.

> We could use feedback from people where unpaved roads are common, and if that information is more important than access restrictions.

Here in eastern Indonesia the surface is much more important. Access restrictions are rarely posted or enforced, except at military and police facilities, but many `highway=primary` features have long unpaved sections, which can be impassable after rain, or for any vehicles with low clearance even when dry. Access restrictions with a practical impact on routing are also often visible by rendering of gates and other barriers at high zoom level.

While it would be ideal to render both types of information, I believe the access restrictions are rarely as significant, and generally are less useful for map rendering (though they are still important for routing).

sommerluk:
Rebased.

> is it possible that you could publish a modified Docker files containing your testing environment?

> I'm using docker, so this would be important for me.

The Docker environment has been updated. You have to execute `docker image ls` to get the image ID of the _kosmtik_ container, and you have to delete it with `docker image rm` to make sure the image is re-build with the new kosmtik version next time you start it.

So testing should be easy now.

> I'm of the opinion that the only way we can get the cartographic "space" to render unpaved surfaces is to drop something else, like access restriction rendering.

In general, I think we have designed the unpaved rendering rather subtle to avoid collisions with other features, so I think it could co-exist with access restriction rendering. But I'm also open to other solutions.

> if that information is more important than access restrictions.

In Benin and Ivory Coast definitely the information about surface is much more important than the information about access restrictions. And I'm quite sure the situation is the same in all West Africa.

As in Indonesia, here in West Africa the surface of the road gives the important information if the road can likely be use by a low clearance car or nor in dry season. In rain season, especially when it's actually currently raining, I would avoid short unpaved parts even when the alternative way is much longer.

Furthermore, here we have often more unpaved roads than paved roads, both in urban and in rural environments:

- Even in urban environments, often most roads are unpaved. And these that are paved are not always the important link roads you would expect to be paved.
  - Some important (secondary, tertiary) link roads have unpaved parts.
  - Some really large and long roads are maybe scheduled for paving, but actually unpaved since years.
  - And sometimes you have some really narrow, curvy paved (residential) roads that are strategically inconvenient locations but used nevertheless a lot because detours over paved roads are better than direct connections over unpaved roads.

- In rural environments, most roads are unpaved. Generally only mayor link roads are paved, but not always: It well possible that you have 50 km or 100 km unpaved on primary roads (without any alternative road).

In West Africa the surface has as much impact on routing as in Europe `access=private`: Avoid it whenever you can. The big difference is: While in Europe there are very few roads that have access restrictions, in West Africa about half of the road network is unpaved. So the surface information in West Africa has a much higher practical impact on utility for the map user than the access restrictions in Europa have.

jeisenbe:
I checked out this branch after using `docker image ls` and `docker image rm <kosmitik_image_id>` to remove the existing kosmtik container, and used `docker-compose build kosmtik` to rebuild the container. The rendering works, except I also got problems with shapefiles. Here's an example showing a private, unpaved residential highway:

z17 after
z17-potikelek-unpaved

z17 before
z17-potikelek-before

z16 after
z16-potikelek-unpaved

z16 before
z16-potikelek-before

z15 after
z15-potikelek-unpaved

z15 before
z15-potikelek-before

But surprisingly, I see Mapnik 4.0.0:
```
kosmtik_1  [Core] Generating Mapnik XML…
kosmtik_1  | [Core] Using mapnik version 4.0.0
```

There are also these errors, as mentioned in #3717:
``` kosmtik_1  | Warning: landcover.mss:574:29 Styles do not match layer selector #landcover.```

However it renders, just not with the shapefiles

I tried fixing the problems with shapefiles by deleting the .index files as suggested (https://github.com/gravitystorm/openstreetmap-carto/issues/3717#issuecomment-518592387), but then got this error: 

``` 
kosmtik_1  | Bus error
openstreetmap-carto_kosmtik_1 exited with code 135
```

**Comment by sommerluk** [comment]
> There are also these errors

These are not errors, but warnings. Since Carto version 1.0.0, there is a change in Carto which outputs now much more warnings than before, but still generates the very same XML result. Its consequences for openstreetmap-carto have been discussed in #3183 and it has been considered a non-issue. And actually, [on openstreetmap.org they are yet using Carto 1.2.0](https://github.com/gravitystorm/openstreetmap-carto/pull/3399#issuecomment-520025952) without a negative impact.

**Comment by sommerluk** [comment]
> I tried fixing the problems with shapefiles by deleting the .index files as suggested (#3717 (comment)), but then got this error:

Hm, I suspect the problem is that scripts/docker-startup.sh calls scripts/get-shapefiles.py which in turn recreates the shapefile indexes using the `shapeindex` binary provided by the Mapnik package provided by your (host) linux distriubution which might be different from the node-mapnik version used by kosmtik.

Could you try to delete the index files, then put your data/ folder in read-only mode to avoid recreating the indexes, and then test again?

**Comment by jeisenbe** [comment]
Sorry, I'm back home with a metered internet connection, and rebuilding the docker containers takes up nearly a gigabyte of data each time. I won't be able to try it again for a month. 

Perhaps @kocio-pl could try? Or @Adamant36?

If we can find the problem, perhaps it can be fixed for the next release of kosmtik.

**Comment by Adamant36** [comment]
@jeisenbe, I tried it in a couple of different Ubuntu versions, but I couldn't update either of them to Mapnik 3.0.22 from 3.0.00 because a Harfbuzz dependency issue that I was unable to figure out during make process. I'll spend some more time on it when I can though.

**Comment by jeisenbe** [comment]
Did you try using Docker? This should be much easier.

If you've never used Docker before, you can just download Docker,
fetch the latest commit from this branch, then use `docker-compose up
import` followed by `docker-compose up kosmtik` and hopefully it
should work, though perhaps the shapefiles will still show the same
issue.

**Comment by Adamant36** [comment]
So I tried it with Docker and as far as I know everything is updated. This is what I got. 

Screenshot from 2019-10-22 01-12-35

It won't load past the last line (unless I didn't wait long enough, but it was a while). Errors aside, it does seem to provide a lot more information then it use to when it's running through everything.

**Comment by sommerluk** [comment]
@Adamant36 Thanks for testing. That looks like a too old Carto version is used. Therefore, some rules (that are only available starting with Carto 1.2.0) are not recognized.

Did you remove old docker containers to make sure they are re-build with the up-to-date software versions?

See here:

> The Docker environment has been updated. You have to execute `docker image ls` to get the image ID of the _kosmtik_ container, and you have to delete it with `docker image rm` to make sure the image is re-build with the new kosmtik version next time you start it.

**Comment by jeisenbe** [comment]
@Adamant36 did you have time to try the steps suggested [above](https://github.com/gravitystorm/openstreetmap-carto/pull/3399#issuecomment-544859828)?

**Comment by Adamant36** [comment]
@jeisenbe, uummm no. I installed the new versions of Carto and Ksomtik on an otherwise clean install of Linux before setting up Docker etc. So one would think I wouldn't have to re-load them. Unless they overwritten to the default older version somewhere in the process. I had planned to look into that, but never got around to it. So, I'll just do what @sommerluk suggested instead when I get some time. 

How do you think it will work with the timeline for integrating the new versions of Carto and Kosmtik into the style's setup so this be an issue? I forgot if that's something that has wait for the reload or some other thing.

**Comment by jeisenbe** [comment]
Did you check out the branch `sommerluk/unpaved` prior to building the docker container for the first time (`docker-compose up import`  and `docker-compose up kosmtik`)? 

For me, if I've already built the containers while on the current `master` branch, and then I try to switch to `sommerluk/unpaved`, I will get errors since the version of Carto is wrong. 

I have to use `docker image rm <ID>` and then rebuild the whole container. Perhaps this is what happened?

**Comment by jeisenbe** [comment]
Does anyone have time to test this again?

What do we need to discuss about this change? There was a suggestion above that we might need to reconsider the access=* rendering along with this.

I am certainly in favor of showing unpaved roads differently, due to the importance of this information in lesser-developed countries and rural areas with unpaved roads.

**Comment by Adamant36** [comment]
I'll test it again when I get some time. If it doesn't go through in the next release maybe at least Carto and Kosmtik could be updated to their latest versions for it in meantime, if they aren't updated for it already.

File: INSTALL.md [hunk 3]
```diff
@@ -130,10 +130,10 @@ Some colours, SVGs and other files are generated with helper scripts. Not all us
 
 For deployment, CartoCSS and Mapnik are required.
 
-* [CartoCSS](https://github.com/mapbox/carto) >= 0.18.0 (we're using YAML)
-* [Mapnik](https://github.com/mapnik/mapnik/wiki/Mapnik-Installation) >= 3.0
+* [CartoCSS](https://github.com/mapbox/carto) >= 1.2.0 (we're using YAML)
+* [Mapnik](https://github.com/mapnik/mapnik/wiki/Mapnik-Installation) >= 3.0.22
```
pnorman: Just flagging that this isn't on the current Ubuntu LTS, and is likely to be a headache.

scripts/generate_unpaved_patterns.py
```diff
@@ -0,0 +1,121 @@
+#!/usr/bin/env python
```
pnorman: This would need porting to python3

INSTALL.md
```diff
@@ -107,7 +107,7 @@ For proper rendering of non-latin scripts, particularly those with complicated d
 ## Dependencies
 
 For development, a style design studio is needed.
-* [Kosmtik](https://github.com/kosmtik/kosmtik) - Kosmtik can be launched with `node index.js serve path/to/openstreetmap-carto/project.mml`
+* [Kosmtik](https://github.com/kosmtik/kosmtik) - Kosmtik can be launched with `node index.js serve path/to/openstreetmap-carto/project.mml` The 0.0.17 release of Kosmtik is not enough because we need up-to-date CartoCSS and Mapnik versions. To install the current master branch of Kosmtik, you can clone the Kosmtik repository and execute `npm install` within it.
```
pnorman: This is a hard blocker for me, but there's probably a release by now.

pnorman (dismissed):
I added a couple of technical comments, but I believe at a higher level we cannot continue to display access information like we do and add surface information to all the other properties we consider when rendering roads.

INSTALL.md
sommerluk: Yes. On the other hand: It will be part of Ubuntu 20.04 LTS, which will be available in only two months. There is also a backport to Ubuntu 18.04 LTS available, provided by @tomhughes [in a PPA](https://github.com/gravitystorm/openstreetmap-carto/pull/3399#issuecomment-519016170) and [yet deployed and running on openstreetmap.org servers](https://github.com/gravitystorm/openstreetmap-carto/pull/3399#issuecomment-520025952). So for openstreetmap.org this would make no difference. But of course for other downstream projects this might be different. I will comment this in INSTALL.MD.

Could we bump Mapnik und CartoCSS versions together with the schema changes to avoid two mayer releases in short time?

scripts/generate_unpaved_patterns.py
sommerluk: Yep. I'm porting this to Python 3. At the same time, I will suppress the C code (helper programm linking against littleCMS) and do everything within the Python script using colormath, which we are using yet in other scripts. So almost same result with much less code. For some colors (motorway-fill primary-fill primary-low-zoom secondary-fill trunk-fill trunk-low-zoom), colormath gives a mimimal different result (+1 or -1 within the RGB range of 0..255), but nothing that would be visible.

INSTALL.md
sommerluk: A kosmtik release would be great. We had asked for a new release about a year ago, but so for no reaction…

sommerluk:
Rebased. Addressed the Python 3 port. Some more comments about Mapnik 3.0.22 in Ubuntu 18.04 and 20.04.

jeisenbe:
So this is blocked by the lack of a new kosmtik release for now, but we also need to decide what to do about access rendering?

sommerluk:
I would suggest raising to Mapnik 3.0.22 and Carto 1.2.0 with the release of openstreetmap-carto 5. We are doing a mayor release anyway, so when raising Mapnik and Carto requirements, so we would be prepared for this PR and we do not need to do another mayor release soon.

(Apart from unpaved rendering, this can also be good for pattern rendering because Mapnik 3.0.22 [has true global pattern alignment](https://github.com/gravitystorm/openstreetmap-carto/issues/937#issuecomment-487484855) which removes the need for pattern sizes as powers of two, and gives us even the possibility of using arbitrary pattern sizes like 113 × 59)

This would not be a problem for openstreetmap.org because they are running yet now Mapnik 3.0.22 and Carto 1.2.0 anyway. Other deployments and forks would have to update, but given that backported packages are available for Ubuntu 18.04, and that Ubuntu 20.04 will be available soon and ship Mapnik 3.0.22 out-of-the-box, this might not be a big issue.

sommerluk:
Okay. Looks we depend on:

- Ubuntu LTS 20.04 being adopted
- A kosmtik release

The first is just a matter of time. Maybe summer 2020?

The second is more difficult to foresee…

Rebased. Scripts adapted to new MSS file directory.

Should I provide a version that does not longer render access restrictions?

pnorman:
> Okay. Looks we depend on:
> 
> * Ubuntu LTS 20.04 being adopted
> * A kosmtik release

The big thing we depend on is a rendering there is consensus around. My views are the same as earlier

> We discussed this at SOTM this weekend. I'm of the opinion that the only way we can get the cartographic "space" to render unpaved surfaces is to drop something else, like access restriction rendering.

sommerluk:
Well, that was why I asked if I should provide a version of this PR that removes access restriction rendering…

jeisenbe:
@sommerluk - could you update this PR to resolve the new merge conflicts? I would like to test it again with Docker.

Re: access restrictions: I'm okay with keeping them in for now, but I would also support removing them in favor of surface rendering, if this is considered a requirement by the other maintainers. They could also be removed in a separate PR after this, in my opinion.

jeisenbe:
I removed the docker images and rebuilt them, after checking out this branch, but now I get this error:
`npm ERR! code 1`
`ERROR: Service 'kosmtik' failed to build: The command '/bin/sh -c npm set prefix /usr && npm install -g "git+https://git@github.com/kosmtik/kosmtik.git#f176c4c"' returned a non-zero code: 1`

Full output of second attempt:
```
Building kosmtik
Step 1/9 : FROM ubuntu:bionic
 ---> c3c304cb4f22
Step 2/9 : RUN apt-get update && apt-get install --no-install-recommends -y     ca-certificates curl gnupg postgresql-client python fonts-hanazono     fonts-noto-cjk fonts-noto-hinted fonts-noto-unhinted mapnik-utils     nodejs npm ttf-unifont unzip git && rm -rf /var/lib/apt/lists/*
 ---> Using cache
 ---> e8d9135f0415
Step 3/9 : RUN npm set prefix /usr && npm install -g "git+https://git@github.com/kosmtik/kosmtik.git#f176c4c"
 ---> Running in 235c3d327cac
npm WARN deprecated nomnom@1.8.1: Package no longer supported. Contact support@npmjs.com for more info.
npm WARN deprecated request@2.88.2: request has been deprecated, see https://github.com/request/request/issues/3142
/usr/bin/kosmtik -> /usr/lib/node_modules/kosmtik/index.js

> mapnik@4.4.0 install /usr/lib/node_modules/kosmtik/node_modules/mapnik
> node-pre-gyp install --fallback-to-build

node-pre-gyp WARN Using request for node-pre-gyp https download 
node-pre-gyp WARN Tried to download(403): https://mapbox-node-binary.s3.amazonaws.com/mapnik/v4.4.0/node-v57-linux-x64-Release.tar.gz 
node-pre-gyp WARN Pre-built binaries not found for mapnik@4.4.0 and node@8.10.0 (node-v57 ABI, glibc) (falling back to source compile with node-gyp) 
gyp WARN EACCES current user ("nobody") does not have permission to access the dev dir "/root/.cache/node-gyp/8.10.0"
gyp WARN EACCES attempting to reinstall using temporary dev dir "/usr/lib/node_modules/kosmtik/node_modules/mapnik/.node-gyp"
/bin/sh: 1: mapnik-config: not found
gyp: Call to 'mapnik-config --cflags' returned exit status 127 while in binding.gyp. while trying to load binding.gyp
gyp ERR! configure error 
gyp ERR! stack Error: `gyp` failed with exit code: 1
gyp ERR! stack     at ChildProcess.onCpExit (/usr/lib/node_modules/kosmtik/node_modules/npm/node_modules/node-gyp/lib/configure.js:351:16)
gyp ERR! stack     at emitTwo (events.js:126:13)
gyp ERR! stack     at ChildProcess.emit (events.js:214:7)
gyp ERR! stack     at Process.ChildProcess._handle.onexit (internal/child_process.js:198:12)
gyp ERR! System Linux 4.19.76-linuxkit
gyp ERR! command "/usr/bin/node" "/usr/lib/node_modules/kosmtik/node_modules/npm/node_modules/node-gyp/bin/node-gyp.js" "configure" "--fallback-to-build" "--module=/usr/lib/node_modules/kosmtik/node_modules/mapnik/lib/binding/mapnik.node" "--module_name=mapnik" "--module_path=/usr/lib/node_modules/kosmtik/node_modules/mapnik/lib/binding" "--napi_version=1" "--node_abi_napi=napi" "--napi_build_version=0" "--node_napi_label=node-v57"
gyp ERR! cwd /usr/lib/node_modules/kosmtik/node_modules/mapnik
gyp ERR! node -v v8.10.0
gyp ERR! node-gyp -v v5.1.0
gyp ERR! not ok 
node-pre-gyp ERR! build error 
node-pre-gyp ERR! stack Error: Failed to execute '/usr/bin/node /usr/lib/node_modules/kosmtik/node_modules/npm/node_modules/node-gyp/bin/node-gyp.js configure --fallback-to-build --module=/usr/lib/node_modules/kosmtik/node_modules/mapnik/lib/binding/mapnik.node --module_name=mapnik --module_path=/usr/lib/node_modules/kosmtik/node_modules/mapnik/lib/binding --napi_version=1 --node_abi_napi=napi --napi_build_version=0 --node_napi_label=node-v57' (1)
node-pre-gyp ERR! stack     at ChildProcess.<anonymous> (/usr/lib/node_modules/kosmtik/node_modules/node-pre-gyp/lib/util/compile.js:83:29)
node-pre-gyp ERR! stack     at emitTwo (events.js:126:13)
node-pre-gyp ERR! stack     at ChildProcess.emit (events.js:214:7)
node-pre-gyp ERR! stack     at maybeClose (internal/child_process.js:925:16)
node-pre-gyp ERR! stack     at Process.ChildProcess._handle.onexit (internal/child_process.js:209:5)
node-pre-gyp ERR! System Linux 4.19.76-linuxkit
node-pre-gyp ERR! command "/usr/bin/node" "/usr/lib/node_modules/kosmtik/node_modules/.bin/node-pre-gyp" "install" "--fallback-to-build"
node-pre-gyp ERR! cwd /usr/lib/node_modules/kosmtik/node_modules/mapnik
node-pre-gyp ERR! node -v v8.10.0
node-pre-gyp ERR! node-pre-gyp -v v0.13.0
node-pre-gyp ERR! not ok 
Failed to execute '/usr/bin/node /usr/lib/node_modules/kosmtik/node_modules/npm/node_modules/node-gyp/bin/node-gyp.js configure --fallback-to-build --module=/usr/lib/node_modules/kosmtik/node_modules/mapnik/lib/binding/mapnik.node --module_name=mapnik --module_path=/usr/lib/node_modules/kosmtik/node_modules/mapnik/lib/binding --napi_version=1 --node_abi_napi=napi --napi_build_version=0 --node_napi_label=node-v57' (1)
/usr/lib
`-- (empty)

npm ERR! Linux 4.19.76-linuxkit
npm ERR! argv "/usr/bin/node" "/usr/bin/npm" "install" "-g" "git+https://git@github.com/kosmtik/kosmtik.git#f176c4c"
npm ERR! node v8.10.0
npm ERR! npm  v3.5.2
npm ERR! code ELIFECYCLE

npm ERR! mapnik@4.4.0 install: `node-pre-gyp install --fallback-to-build`
npm ERR! Exit status 1
npm ERR! 
npm ERR! Failed at the mapnik@4.4.0 install script 'node-pre-gyp install --fallback-to-build'.
npm ERR! Make sure you have the latest version of node.js and npm installed.
npm ERR! If you do, this is most likely a problem with the mapnik package,
npm ERR! not with npm itself.
npm ERR! Tell the author that this fails on your system:
npm ERR!     node-pre-gyp install --fallback-to-build
npm ERR! You can get information on how to open an issue for this project with:
npm ERR!     npm bugs mapnik
npm ERR! Or if that isn't available, you can get their info via:
npm ERR!     npm owner ls mapnik
npm ERR! There is likely additional logging output above.

npm ERR! Please include the following file with any support request:
npm ERR!     /npm-debug.log
npm ERR! code 1
ERROR: Service 'kosmtik' failed to build: The command '/bin/sh -c npm set prefix /usr && npm install -g "git+https://git@github.com/kosmtik/kosmtik.git#f176c4c"' returned a non-zero code: 1
```

sommerluk:
Rebased. All merge conflicts are resolved. I've also updated the Docker image to Ubuntu focal to make sure to get up-to-date Mapnik packages; however this change I could not test as my Internet connection is too slow to download new docker containers. It would be great if you could test that. Honestly I do not know why node-mapnik does not build, as your error message indicates - npm dependencies are the hell. When I tested it once, it worked fine.

I'll see what I can do with removing access rendering…

jeisenbe:
With the latest commit, using docker everything was working until this point:

```
Setting up libsasl2-modules-db:amd64 (2.1.27+dfsg-2) ...
Setting up tzdata (2019c-3ubuntu1) ...
debconf: unable to initialize frontend: Dialog
debconf: (TERM is not set, so the dialog frontend is not usable.)
debconf: falling back to frontend: Readline
Configuring tzdata
------------------

Please select the geographic area in which you live. Subsequent configuration
questions will narrow this down by presenting a list of cities, representing
the time zones in which they are located.

  1. Africa      4. Australia  7. Atlantic  10. Pacific  13. Etc
  2. America     5. Arctic     8. Europe    11. SystemV
  3. Antarctica  6. Asia       9. Indian    12. US
Geographic area: 
```
A search online suggests that a few lines need to be added to the dockerfile to deal with this:

https://serverfault.com/questions/949991/how-to-install-tzdata-on-a-ubuntu-docker-image

jeisenbe:
The suggestions in the link above didn't work for me. 

I got farther with the suggestion here: https://askubuntu.com/questions/909277/avoiding-user-interaction-with-tzdata-when-installing-certbot-in-a-docker-contai

I added this to Dockerfile:

```
RUN apt-get update && DEBIAN_FRONTEND=noninteractive apt-get install -y --no-install-recommends \
        tzdata \
    && rm -rf /var/lib/apt/lists/*
```

However, it still gets stuck here:

```
Setting up mapnik-utils (3.0.23+ds-1build1) ...
Processing triggers for libc-bin (2.31-0ubuntu9) ...
Processing triggers for ca-certificates (20190110ubuntu1) ...
Updating certificates in /etc/ssl/certs...
0 added, 0 removed; done.
Running hooks in /etc/ca-certificates/update.d...
done.
Removing intermediate container b8fdcf93036c
 ---> 5344c7db7bda
Step 4/10 : RUN npm set prefix /usr && npm install -g "git+https://git@github.com/kosmtik/kosmtik.git#f176c4c"
 ---> Running in 03f4c2cbc126
npm WARN deprecated nomnom@1.8.1: Package no longer supported. Contact support@npmjs.com for more info.
npm WARN deprecated request@2.88.2: request has been deprecated, see https://github.com/request/request/issues/3142
/usr/bin/kosmtik -> /usr/lib/node_modules/kosmtik/index.js

> mapnik@4.4.0 install /usr/lib/node_modules/kosmtik/node_modules/mapnik
> node-pre-gyp install --fallback-to-build

ERROR: Service 'kosmtik' failed to build: The command '/bin/sh -c npm set prefix /usr && npm install -g "git+https://git@github.com/kosmtik/kosmtik.git#f176c4c"' returned a non-zero code: 1
```

sommerluk:
Alternative version at #4137 without access rendering.

sommerluk:
> I added this to Dockerfile:

Could you post the entire Dockerfile with your modifications? So I can incorporate it in this PR.

> However, it still gets stuck here:

Does a new installation of kosmtik work for you outside Docker?

westis:
Does this mean there is now a way to render paved vs unpaved? When will it make it to the public map?

jeisenbe:
> When will it make it to the public map?

This PR has not yet been merged. There is also a [newer PR](https://github.com/gravitystorm/openstreetmap-carto/pull/4137), based off of this one, which also removes the `access=` rendering: #4137. That still needs to be reviewed. (Unfortunately, I'm busy at the moment and do not have time to do the review myself for the next couple weeks)

imagico:
Note i have in the meantime tuned the access restriction dashing to work both with paved and unpaved roads reasonably well i think in the ac-style:

ac-style

http://blog.imagico.de/rethinking-road-access-restrictions-rendering/

I am fine with the alternative to remove the access restrictions illustration instead - but combining the two while still being readable quite clearly seems to be possible.

sommerluk:
Okay.  Thanks for the hint!

I think, if we would change the access rendering, this would have to go into a separate pull request anyway.

So for now, maybe we could continue with _this_ PR preserving access rendering for the moment – and access rendering might be changed later…

INSTALL.md
sommerluk: Is available now since more than two years: has been part of Ubuntu since 20.04 LTS.

INSTALL.md
sommerluk: I don’t find the reference now, but the Kosmtik developer said he won’t do releases any more (actually, the last one is from 2017), because you can easily install it directly from master at Github. So I think this is a non-issue.

imagico (changes requested):
Got around testing this finally.

The fundamental things first:

Design wise I think this is a good solution for rendering unpaved surface roads in a subtle yet intuitive way that fits into the existing rendering system.  In particular because it works for all road types rendered with fill and casing (including pedestrian/service road polygons) and all aeroway types with a common styling paradigm.   And because it works well both in areas where paved roads are dominant as well as ones where unpaved roads are dominant - making it a good solution for a global project like this. No alternative has been suggested since #110 was opened 9 years ago that would even closely match these advantages.

The need to require a recent kosmtik version installed from source seems fine with me.  I have been taking that approach for kosmtik for some time anyway and it is not a problem from my perspective (beyond the awkwardness any node-js software comes with obviously).  We cannot in the long term continue this project if we essentially feature freeze our toolchain based on the last kosmtik release.

Now to testing - it looks good, i have not done extensive real world testing since design wise this is very similar to what i have been using in the ac-style for some time - you can find samples for that on https://imagico.de/map/ac_samples_en.php.

Here is an abstract test at z18 in comparison:

Master:
master

ac-style:
ac

This PR:
pr

Other zoom levels: [z12](https://imagico.de/files/road_unpaved_access_12_40.png), [z13](https://imagico.de/files/road_unpaved_access_13_40.png), [z14](https://imagico.de/files/road_unpaved_access_14_40.png), [z15](https://imagico.de/files/road_unpaved_access_15_40.png), [z16](https://imagico.de/files/road_unpaved_access_16_40.png), [z17](https://imagico.de/files/road_unpaved_access_17_40.png), [z19](https://imagico.de/files/road_unpaved_access_19_40.png)

I see the following issues:

* visibility of the access dashing is not good but it can be improved by toning the dashing color.  I am fine with doing that as a followup change.
* it could be considered to drop both the access dashing and the unpaved pattern rendering from roads with very narrow fill (like highway=road and highway=service at z15, minor service roads at z17, residential/unclassified at z13).  This also is a tuning idea that can be done later as well.
* pedestrian roads have the unpaved pattern on tunnels while the others don't.  For consistency i think this should be changed.

sommerluk:
Okay.

For the first and the second point, I agree to do this as possible follow-up changes.

For the third point (inconsistent tunnel rendering for pedestrian), I've fixed this now.

imagico (approved):
Looks good now.

davidbannon:
Great that this has been done. But how do we see it ?  I imagined we were talking about what is shown on osm.org but no sign of any way to tell the difference between sealed and unsealed roads there ??
Davo

imagico:
We have not yet made a new release since this was merged (and yes, we should probably do so).

davidbannon:
Ah, that makes perfect sense, thanks !  "Release" is hard work, I am about to do a new release of my app, tomboy-ng on a much smaller scale.  Doing the dev is lots more fun! Good luck.

Davo

imagico:
Since this issue has now been prominently linked to by [WeeklyOSM](https://weeklyosm.eu/archives/16243) i wanted to add that the majority of the development discussion can be found not here but in previous pull requests: #3357 and in particular #2640.  A lot of the early discussion happened also in #110.  This has been a lengthy process but reading this can provide some valuable insights into how cooperative map design can work and what obstacles this is practically faced with.

Overall a lot of people contributed over the year to discussing and trying out different ideas to make this work and quite a few were in some form involved in making the solution we have now chosen (which was first discussed back in 2015) work despite all the obstacles that lead to many delays.

pardthemonster:
I'm a little late to the discussion and I am sure this is discussed somewhere, but I just haven't found it yet.  My question is about all the other unpaved variations per the wiki (https://wiki.openstreetmap.org/wiki/Key:surface#Unpaved), i.e. `surface=compacted`, `surface=fine_gravel`, `surface=gravel`, etc.

imagico:
You have not actually formulated a question.  The way we interpret the surface tag can be found in

https://github.com/gravitystorm/openstreetmap-carto/blob/6591d7b97ce34395d8e214d25329c50742bce044/project.mml#L702-L706

This is the same logic we already used for differentiating path/footway so far.

pardthemonster:
Thank you on clarifying the others.

PR #3401, opened 2018-09-18
Add icon for tourism=gallery

### LINKED ISSUE #883: Add rendering for tourism=gallery
In 'Map Features', tourism=gallery has this explanation:

Art galleries are a popular visitor attraction 

but no icon
worse: even if the gallery has been given a name, nothing is shown

So, even though galleries are popular attractions, they are invisible on the map.

Adamant36:
This PR adds rendering for tourism=gallery. Closes #883 
### Definition and Rational for Rendering
Although there are some people who think this tag is to vague etc, none of the arguments against it use where very strong, it was only a small number of people, and the tags usage has increased quit a lot in the 5 years since objections to it started. Plus, as I have been reminded multiple times myself, this isn't a place for a tagging discussions anyway. That being said, although the word "gallery" might not explicitly mean an art gallery in a general sense, it is clearly the definition given on the wiki page. To me personally an art gallery is clearly distinct from shop=art or tourism=museum in the fact that unlike them it is a place that only displays art but does not mainly do it for profit. Whereas an art shop sells art and a museum features other things besides art. The wiki page gives the same definition, but as I stated, this isn't the place for tagging/tag definition discussions anyway. Its about if the tag should be rendered or not and nothing against the tags definition is compelling enough to make me think it shouldn't be rendered. Not to mention there is no alternative tag for the places this one covers anyway. So its either the things get rendered as tourism=gallery or they never get rendered. I rather they get rendered.

As far as the benefits to rendering it, it has 5,168 uses and the uses are increasing rather rapidly. So there is clearly a market for it being rendered. Also, the items tagged as tourism=gallery don't currently have their names rendered on the map. Which creates rather large gaps on the map in places where multiple galleries are on the same block, like in Oakland where did the test rendering. Also, whatever museums or art shops are wrongly tagged as tourism=gallery will probably never be fixed if the tag is not rendered. since mappers wont know the places are on the map in the first place to fix them.  
taghistory

### Test Renderings
https://www.openstreetmap.org/#map=19/37.81541/-122.26658 (way)
way
https://www.openstreetmap.org/#map=19/37.81454/-122.26766 (node)
node


kocio-pl:
Also this PR needs some rationale and summary, please add them.

dieterdreist:
can you explain which definition of tourism=gallery is represented by this
icon?

Adamant36:
@dieterdreist, I'm not really sure of any definition of a gallery outside of an art gallery and the wiki page is pretty clear on that being its definition. Plus, it seems more like a tagging question then a "why should this be rendered question." That being said, I updated my original post to answer your question anyway.

lakedistrictOSM:
If possible, it would be good to use a different icon. The one currently proposed makes me think of a place where you can do art activities, like a children's arts centre or an artist's studios.

I think that a framed painting design might work better - here's one I made a while ago:
artdraft01png

If you like it, feel free to test it (I'll upload it to gist later) or just feel free to take inspiration from it.

dieterdreist:
sent from a phone

> On 20. Sep 2018, at 11:50, Adamant36 <notifications@github.com> wrote:
> 
> I'm not really sure of any definition of a gallery outside of an art gallery and the wiki page is pretty clear on that being its definition.


there are 2 types of art galleries: those that only display and those that sell the works. 
The former are typically places which have their own permanent exhibition (collection), the latter are places where artists and collectors are brought together by the gallery owner.

Adamant36:
@dieterdriest, I semi agree with you. Although, Id say the places with permanent exibits are more like museums. Theres clearly a third category of places that have rotating displays but arent mainly in it to sell the art. The wiki makes the distinction and so does this guy (whoever he is) http://handmade-business.com/language-arts-theres-a-distinct-difference-between-a-gallery-and-a-shop-learn-what-it-is/

So the question do we follow the wiki and that guy by displaying galleries (how ever it might define it) on the map. Since theres clearly a bunch of "whatever you want to call them" things mapped..Or do we not display them because a slight ambiguity in the meaning of the word "gallery" might exist? If it was a new tag and hadnt already been dissussed id say dont render it until gets figured out at least. But the disscussions over it went know where in five years and the tag kept being used in ever increasing numbers. Its clear that the tag is one the community accepts and widely uses.  So I think thats good enough to say it should be displayed.

Adamant36:
@kocio-pl, I don't get what the conflict here is. Is it the empty line at 2349 or something?

dieterdreist:
2018-09-23 3:54 GMT+02:00 Adamant36 <notifications@github.com>:

> @dieterdriest, I semi agree with you. Although, Id say the places with
> permanent exibits are more like museums.
>


art galleries _are_ a kind of museum for paintings and today also photos. A
synonymous term is pinacotheca / pinacothèque / pinakothek ... (maybe would
be better, as it doesn't leave room for confusing it with commercial art
galleries)



> Theres clearly a third category of places that have rotating displays but
> arent mainly in it to sell the art.
>

can you give some examples, and with which terms they describe themselves?



> The wiki makes the distinction and so does this guy (whoever he is)
> http://handmade-business.com/language-arts-theres-a-
> distinct-difference-between-a-gallery-and-a-shop-learn-what-it-is/
>


there are also "art shops" that aren't galleries, I find this term
ambiguous as well.



> So the question do we follow the wiki and that guy by displaying galleries
> (how ever it might define it) on the map.
>


the wiki is relatively recent and the definition is not necessarily set in
stone (IMHO), rather the opposite I would say there are issues with the
current definition.



> Or do we not display them because a slight ambiguity in the meaning of the
> word "gallery" might exist? If it was a new tag and hadnt already been
> dissussed id say dont render it until gets figured out at least. But the
> disscussions over it went know where in five years and the tag kept being
> used in ever increasing numbers. Its clear that the tag is one the
> community accepts and widely uses. So I think thats good enough to say it
> should be displayed.
>

we're not talking about a "slight ambiguity", but about a significant one,
and increasing numbers for a tag with unclear meaning are not helpful.
An art gallery in the sense of a museum for paintings, usually merits
prominent display also at not the highest zoomlevel, for art galleries
where the gallery owner is an intermediary between the artist and the
customer / art collector the range of importance is huge, from similar to
the first meaning to small shop interesting to very few people.

Adamant36:
I would consider a museum a place that houses collections of things, like artifacts, that are of historical or cultural importance. A lot of art in art galleries dont fit those categories and often one off paintings, instead of collections. 

A lot of colleges have galleries that display art done by the students or other artists. They call themselves galleries. Theres also none profits that display art created by community members. Although they often call themselves "art centers," they still contain a gallery. There is also places like the Slate Contemporary in Oakland that has a gallery and a different building off site where the works are sold. Id imagine there are many places where the place that the is displayed is different then where its sold, or they just refer a buyer to an artist and get a cut later.  Those places would be galleries in my opinion. As the transaction is done off site. 

Whats ambiguous about "art shop" exactly? 

I think the fluidity of the wiki can apply in many cases, not just this one. That's not a reason to not render galleries though. IMHO, its set in stone at least in the fact that the discussion over the current definition ended five years ago and nothing was modified based on it. No one switched over to another tag or anything. So maybe the minor details need to be ironed out, at least in your opinion, but that doesnt put the whole tag in question. So theres no reason to wait until its %100 decided to render it. You could also put up a "this key is in dispute" banner or something and start the conversation over again. Although I doubt it will go anywhere. 

Its slight in the fact that theres only a few detractors, tag is used a bunch, and it seems like there is clearly a third category if you think so or not. Even if it might be hard to pin down what exactly it is. While I agree with you that usage numbers shouldnt be an excuse for rendering a bad tag under normal conditions, I think it comes down to the fact that these things clearly exist even if its fuzzy as to what qualifies. They arent  shops or museums though, they will probably never be tagged as those things because of the fuzzyness, and the fuzzyness isnt going to go away. So its better just to render them. Otherwise, your forcing something into a binary choice between museums or shops that it doesnt fit into. And you say shops are ambigious, which  means every place displaying art should be a museum then? That doesn't sound right.

dieterdreist:
2018-09-24 11:15 GMT+02:00 Adamant36 <notifications@github.com>:

> I would consider a museum a place that houses collections of things, like
> artifacts, that are of historical or cultural importance. A lot of art in
> art galleries dont fit those categories and often one off paintings,
> instead of collections.
>

"importance" is relative, there are surely a lot of small museums which
will not be considered universally "important".


A lot of colleges have galleries that display art done by the students or
> other artists. They call themselves galleries.
>


The term "gallery" has a lot of meanings, it is also an architectural term
for a building part, but these are not _tourism_=gallery.



> Theres also none profits that display art created by community members.
> Although they often call themselves "art centers," they still contain a
> gallery. There is also places like the Slate Contemporary in Oakland that
> has a gallery and a different building off site where the works are sold.
>


indeed, and this is the second meaning I was describing above.


> Whats ambiguous about "art shop" exactly?
>


places that call themselves art shops often do not sell original art but
reproductions, prints and books. The term for an art gallery is art
gallery. I would use shop=art_gallery as this is more clear. "art shop" is
a different category.

Adamant36:
@kocio-pl, what do you think about this one? I think dieterdreist makes some good points, but then I'd still like to see the places rendered on the map and galleries seem to be in enough of a gray area definition wise where I don't think they will ever get re-tagged as anything else. Even if a better tag for them comes along, which I don't see happening.

kocio-pl:
Sorry for not following this, I have less time lately and there were more PRs to check and discuss.

I have just looked at https://wiki.openstreetmap.org/wiki/Tag:tourism%3Dgallery and it looks quite solid. It is approved tag and the difference from both `shop=art` and `tourism=museum` is defined. I understand it means special kind of museum - "art museum". I have no problem with it, since some real life categories are special while being part of more generic category (like `shop=shoes` is in fact kind of `shop=clothes`) and some common objects being split between tags in OSM (like `historic=memorial` and `historic=monument`).

Adamant36:
No worries. Its been pretty busy around here lately. So, thanks for looking into it. I totally agree with you. Since your good on it, I'll fix the conflicts when I have some time and wait for it to be merged.

Adamant36:
@kocio-pl, I fixed the conflicts. I also consolidated the code for a few amenities that had the same attributes while I was there. It seems to have passed Travis fine.  So, its good to go.

kocio-pl:
Thanks! It works as expected in my test renderings.

Adamant36:
Awesome. Thanks for testing and merging it.

kocio-pl:
I have found also examples of open air gallery and gallery of photos operated by photographic society - both are not the shop (not focused on selling) nor museum (not in the common meaning of the word), so this tag is really useful.

kocio-pl:
@tomasz-w What do you think about tweaking art center icon to show it different than galleries? We have different colors to show if it's retail gallery or not, but art center is wider (for example it can contain some gallery space inside). Maybe it would be good to make some kind of objects mix, like we have for `shop=video` or `shop=music` (maybe adding similar note symbol would work).

Tomasz-W:
@kocio-pl I'll try with palette + note icon.

Tomasz-W:
@kocio-pl Palette + note:

amenity arts_centre

kocio-pl:
Could you switch them (note on the left)?

Tomasz-W:
@kocio-pl 
* amenity arts_centre (switched)
* amenity arts_centre2 (rotated palette)

lakedistrictOSM:
I prefer the original one (  )

Alternatively, we could instead use a different icon for gallery and art shop, maybe something like the one I suggested in https://github.com/gravitystorm/openstreetmap-carto/pull/3401#issuecomment-423739718 (  ).

BTW, you can't buy a painting palette at a shop=art, but you can at a [shop=craft](https://wiki.openstreetmap.org/wiki/Tag:shop%3Dcraft) (arts and craft supplies shop, 5.1k uses) which doesn't have an icon yet.

kocio-pl:
This could be also proper symbol for shop=frame:

taghistory 60

kocio-pl:
@Tomasz-W What about one theater mask with half of palette?

Tomasz-W:
@kocio-pl 
* amenity arts_centre1 (palette + note v1)
* amenity arts_centre2 (palette + note v2)
* amenity arts_centre3 (palette + mask)

kocio-pl:
Could someone make a test rendering for all these icons?

lakedistrictOSM:
@Tomasz-W The mask is too small for me. How about mask + paintbrush, with a bigger mask?

Tomasz-W:
@lakedistrictOSM mask + brush looked weird, so I tried with note + brush:

amenity arts_centre4

vertical brush looked more like a torch, so I rotated it.

PR #3420, opened 2018-09-29
Order amenity_lines by layer

### LINKED ISSUE #3413: Layer not respected in attraction=water_slide
The `layer`-tag is not respected in water-slides:
image

https://www.openstreetmap.org/way/603108779
https://www.openstreetmap.org/way/603108778

johsin18:
This way, we respect the layer tag for water slides, and it should not harm for fords.

Fixes #3413 

Order amenity_lines by layer

https://www.openstreetmap.org/#map=19/48.51073/9.03904

Before
screenshot_20180929_145854

After
screenshot_20180929_145912

### DIFF
--- a/project.mml
+++ b/project.mml
@@ -1564,6 +1564,7 @@ Layer:
         (SELECT
           way,
           name,
+          layer,
           COALESCE(
            'highway_' || CASE WHEN tags @> 'ford=>yes' OR tags @> 'ford=>stepping_stones' THEN 'ford' ELSE NULL END,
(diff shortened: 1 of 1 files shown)


kocio-pl:
Looks very good and simple, should not do any harm for both current features.

kocio-pl:
Interesting - my local rendering from the current version works as expected without this fix. Could anyone else check it?

I guess however that sorting should be safe anyway. At least it does not break my local rendering.

kocio-pl:
Thanks! With DESC it was inverted, but without it it worked as it should. I don't know why my setup works different than OSMF servers, but that should be OK.

Issue #3071, opened 2018-02-14
A main entrance does not stand out well if the building is a public building.


geozeisig:
A main entrance does not stand out well if the building is a church, train station or airport passenger building. Example: [Berliner Dom](https://www.openstreetmap.org/node/1967329593), [S-Bahrenfeld](https://www.openstreetmap.org/node/3527774968)
Maybe we should make the entrance **in this** case even darker.

kocio-pl:
@tpikonen What's your take on this? I'm also unsure about #3038.

dieterdreist:
2018-02-14 14:46 GMT+01:00 geozeisig <notifications@github.com>:

> A main entrance does not stand out well if the building is a church, train
> station or airport passenger building. Example: Berliner Dom
> <https://www.openstreetmap.org/node/1967329593>, S-Bahrenfeld
> <https://www.openstreetmap.org/node/3527774968>
> Maybe we should make the entrance *in this* case even darker.
>



the Berliner Dom case is not mapped well, because the entrance isn't part
of the building but only of a building part which is inside the building.

tpikonen:
I agree that these examples look pretty bad. Also, buildings-major and places of worship are also the places where map users probably benefit the most from seeing main entrances.

Technically this can be fixed in the SQL level by getting suitable building/way tags from a spatial intersection of the entrance with the building (similar to the solution proposed in PR #2932) and adjusting the color.

The more straightforward solution would be to make all main entrances dark enough to be visible also in these dark buildings.

dieterdreist:
2018-02-14 16:42 GMT+01:00 tpikonen <notifications@github.com>:

> The more straightforward solution would be to make all main entrances dark
> enough to be visible also in these dark buildings.



or the opposite: very bright

polarbearing:
Could we change the title of this issue, s/public/major/ ?

Tomasz-W:
I've missed a discussion about entrances rendering, and now I find these boxes not intuitive at all. Wouldn't be the classic door symbol from architectural plans suit better here?

[...]

Issue #2475, opened 2016-11-28
Render building=roof differently to expose underlying features


BushmanK:
There are a couple of issues, logically (but not technically) related to this situation: #2037 and #688 , but this is a bit different thing. 

Often, `building=roof` area covers something: a platform, a roadway, or even another building. Examples are: gas stations with offices or small stores, located under the roof, but are not constructed as a single building with it; public transport platforms; covered galleries or passage ways.

Logically,  `building=roof` should presume larger value of `layer=` than everything else under it. But using `layer=` on it explicitly seems redundant because of its nature, so it is actually rarely used. Another problem is, if (when) #688 will be resolved, it will be impossible to see objects hidden under the roofs.

Honestly, I don't know what kind of rendering to suggest, because transparency doesn't seem like a good idea (discussed in #688 ) while transparent patterns could look quite ugly and cause different issues. But I want to point at this situation anyway just in case if someone has a better idea.

mboeringa:
I have already shown this multiple times in the other threads, but just for the sake of keeping the discussion going, I post here two new examples of Amsterdam Centraal Station and Haarlem Station in the Netherlands, that show that a fine overlay pattern (non-transparent), can work for solving this issue.

> Logically, building=roof should presume larger value of layer= than everything else under it. But using layer= on it explicitly seems redundant because of its nature, so it is actually rarely used.

And yes, I render this on top of everything else (at least above buildings), and don't require and ignore any layer=x values.

As rendered by my ArcGIS Renderer. Notice the fine grating above the platforms of the railway stations, representing the platform / station roofs:
![arcgis_renderer_for_openstreetmap-amsterdam_centraal_station](https://cloud.githubusercontent.com/assets/7635726/20687260/18c2a818-b5bc-11e6-8c2c-6ae7493969ad.png)
![arcgis_renderer_for_openstreetmap-station_haarlem](https://cloud.githubusercontent.com/assets/7635726/20687261/18c589a2-b5bc-11e6-8001-a7175d024e69.png)

BushmanK:
I agree, that crosshatching looks well as a part of this particular style. However, this kind of fine pattern has one fundamental problem: it causes moire (regular interference pattern, actually) when printed. I honestly don't know, if printability is among the considerations of developers. 

[...]

PR #3426, opened 2018-10-05
 Buildings code rewrite

### LINKED ISSUE #2532: Rethinking major and minor buildings

meased:
(Most likely) a prerequisite for #2532

This PR simplifies the buildings SQL query and moves all styling decisions out of SQL and into CSS (as the software gods intended :wink:).

No visual changes.

## SQL

Before, normal buildings and major building were queried separately with each query specifically omitting the results of the other. This made it difficult to redefine the categories (and very difficult to add a third category).

The SQL query now queries all buildings together, along with any fields needed to make styling decisions:

 ``` sql
      (SELECT
            way,
            building,
            amenity,
            aeroway,
            aerialway,
            tags->'public_transport' as public_transport
          FROM planet_osm_polygon
          WHERE building IS NOT NULL
            AND building != 'no'
            AND way_area > 1*!pixel_width!::real*!pixel_height!::real
        ) AS buildings
```

## CSS

Now that the buildings class has access to all the fields needed, the CSS can be written in one block. As you can see, it is now very easy to reclassify buildings and add/remove the minor class:

``` css
  #buildings {
    [zoom >= 13] {
      polygon-fill: @building-low-zoom;
      polygon-clip: false;
      line-width: 0;
      [zoom >= 15] {
        polygon-fill: @building-fill;
        line-color: @building-line;
        line-width: .75;
        line-clip: false;
      }
      [amenity = 'place_of_worship'],
      [aeroway = 'terminal'],
      [aerialway = 'station'],
      [building = 'train_station'],
      [public_transport = 'station'] {
        polygon-fill: @building-major-fill;
        line-color: @building-major-line;
      }
    }
  }
```


jeisenbe:
Great work on the code!

I think it might be better with the major buildings a little more
prominent.

Could we try a color half-way between the old and new major building colors?

kocio-pl:
I like the idea of these complex changes. Did you however test the performance? We use SQL a lot to achieve reasonable speed.

kocio-pl:
Could you also test minor buildings darker, like half the way between the proposed color and standard buildings?

meased:
Minor buildings no change:
0perc

Minor buildings 4% lighter:
4perc

Minor buildings 7% lighter (same as screen shots above):
7perc

meased:
Major buildings no change:
maj2_20perc
maj_20perc

Major buildings 15% darker:
maj2_15perc
maj_15perc

Major buildings 10% darker (same as screenshots above):
maj2_10perc
maj_10perc

As you can see, I was more concerned with fixing the dreadful "gastronomy in the airport" problem than I was with spotting churches in neighbourhoods.

meased:
> Did you however test the performance?

No.

Anyone know how to run benchmarks in docker?

jeisenbe:
Would it make sense to lighten major buildings at higher zoom levels? At
mid-zoom it is very helpful that the airport terminal building is obvious,
and that’s when it is useful for places of worship etc.

But at zoom 18 there is no reason for the major building to stay so dark.

Would that be too complicated ?

meased:
That's not that complicated. Worth a test rendering at least.

kocio-pl:
I think both darker versions (4% for minor and 15% for major) work better. Since buildings are very popular map feature, we need much more testing in different areas (backgrounds). 

I like the idea of minor buildings instead of problems with ruins icon, especially for castles (see #331).

Unfortunately we have no performance framework (see #1287). However we can at least try to deploy it on some real server - @rrzefox, could you measure the effect of replacing some SQL code with Carto CSS here?

matthijsmelissen:
See also #68 and #490.

I'm not sure if I really see the advantage of rendering minor buildings lighter.

mboeringa:
> I'm not sure if I really see the advantage of rendering minor buildings lighter.

That surprises me. To me it makes utter sense to render relatively insignificant non-living quarters like small garden sheds and garages lighter.

It might actually stimulate good tagging practices in that people actually start to distinguish them in tagging, and add **building=shed** instead of just generic **building=yes**.

Tomasz-W:
@meased Can you test some lighter outline for minor buildings? It looks a little bit like some holes in ground currently.

mboeringa:
> @meased Can you test lighter outline for minor buildings? It looks a little bit like some holes in ground currently.

I think that is mainly a consequence of the relatively dark grey of the **landuse=residential**, which appears just a bit darker than the minor buildings. So either slightly lightening up of landuse=residential, or slightly darkening the minor buildings fill, might help.

meased:
> I think that is mainly a consequence of the relatively dark grey of the landuse=residential

Exactly.

I choose landuse=residential and airports for examples because they represent the worst cases for how light you can go on one end and how dark you can go on the other. There is not much space to play with in the middle. The minor buildings above would have looked better on bare background:
base_minor

kocio-pl:
> It might actually stimulate good tagging practices in that people actually start to distinguish them in tagging, and add building=shed instead of just generic building=yes.

I also believe it's more pleasant, less cluttering and more useful at the same time, so it's a triple win for me.

There are a lot of auxiliary buildings, but still most of them is tagged with `building=yes` (and only about 50 mln out of 300 mln of all the buildings has any other value!). I don't see how could we attract people to more precise tagging other than showing them a bit different as a "minor" super-class.

taghistory 55
taghistory 56

@meased I see that you have added carpark type, but omitted collapsed - was there any reason for this? I have tried to stay above 10k limit, but we don't have to stick to it, there are also some other types that make sense for me, what do you think about it?

taghistory 57

jragusa:
Please also consider underground #552 and roof #2475 buildings to get a consistent rendering.

meased:
I got the original minor buildings list from #2532. I was avoiding underground (because some people don't want it rendered at all) and roof (because #2475 is different than just rendering them lighter). But until those other issues are sorted out I don't see why they wouldn't qualify as minor buildings in the meantime.

@kocio-pl: Strangly, collapsed isn't on the [wiki page](https://wiki.openstreetmap.org/wiki/Key:building). I included it anyway because the usage is very high.

The list of minor buildings is now: 

``` css
    [building = 'roof'],
    [building = 'construction'],
    [building = 'damaged'],
    [building = 'collapsed'],
    [building = 'underground'],
    [building = 'service'],
    [building = 'slurry_tank'],
    [building = 'shed'],
    [building = 'garage'],
    [building = 'garages'],
    [building = 'farm_auxiliary'],
    [building = 'carport'],
    [building = 'barn'],
    [building = 'stable'],
    [building = 'cowshed'],
    [building = 'greenhouse'],
    [building = 'ger'],
    [building = 'ruins'] {
```

meased:
Playing with @jeisenbe's idea of variable major building darkness:

a14
a15
a16
a17


c14
c15
c16
c17

Tomasz-W:
About test renderings above: I find darkness of major buildings on high zoom levels definetely too low. When I looked at them with "fresh eye" I had to search for major buildings, but they should be distinguishable intuitively. Just to be clear: I very like the idea proposed in this PR, it just need some tuning IMO.

buildings.mss
```diff
@@ -12,29 +15,41 @@
...
+    [building = 'shed'],
+    [building = 'garage'],
+    [building = 'garages'],
+    [building = 'farm_auxiliary'],
+    [building = 'carport'],
+    [building = 'barn'],
+    [building = 'stable'],
+    [building = 'cowshed'],
+    [building = 'greenhouse'],
+    [building = 'ger'],
```
HolgerJeromin: What is that?

I miss =hut and perhaps =cabin in the list.

kocio-pl: You mean ger? See https://en.wikipedia.org/wiki/Yurt

kocio-pl:
What about construction or ruins?

dieterdreist:
sent from a phone

> On 9. Oct 2018, at 03:58, kocio-pl <notifications@github.com> wrote:
> 
> What about construction or ruins?


I’m undecided, they are not building types but they work like elsewhere, e.g. highway=construction, construction=primary

And they are at least buildings

kocio-pl:
@meased Variable darkness is smart trick and it can work for me. More tests would be useful to make sure, because buildings are very popular. 

The most important think for now is deciding what the list should contain. I like the idea of tagging `building=construction/ruins/collapsed/damaged` + `construction/ruins/collapsed/damaged=*`, but I'm not sure what should we render then - their uncomplete state as lighter or just their type.

***

Example of gers (yurts/tents) being less visible than regular buildings (in this area some of them are not tagged properly, so the difference in rendering is evident), using current code:

https://www.openstreetmap.org/#map=18/47.93434/106.91192

ft9dvqo1

rrzefox:
> However we can at least try to deploy it on some real server - @rrzefox, could you measure the effect of replacing some SQL code with Carto CSS here?

I will gladly deploy this once it has reached "beta" state, i.e. coloring has been decided.
I seriously doubt this has any measureable performance impact.

Adamant36:
@rrzefox, since we are on the subject of performance, do you know what impact there would be from creating more stylesheets with what is currently in the amenity file?

Tomasz-W:
@meased Can you try with 17% for major and 5% for minor due to some users questions about further tests? It would be good to see some new locations to have a 'fresh view' on it.

meased:
No problem, I'll do more testing this weekend.

kocio-pl:
I think deploying this idea will help people realize that there are many different building types and start adding them more often. In the looooong term this might lead to render `building=yes` as kind of minor/incomplete. So this is much more to this change than meets the eye!

I believe in such incremental, but more systematic changes. For example making more categories for POIs, as I did purely for making this style easier to customize, has led to making gastronomy icons orange, which may help to avoid clutter by starting rendering them as dots.

dieterdreist:
sent from a phone

> On 12. Oct 2018, at 19:29, kocio-pl <notifications@github.com> wrote:
> 
> I think deploying this idea will help people realize that there are many different building types and start adding them more often. In the looooong term this might lead to render building=yes as kind of minor/incomplete. So this is much more to this change than meets the eye!


+1, well said, the times that mapping projects called themselves proudly building=yes should be gone ;-)

meased:
I have removed `collapsed` and added `hut` and `cabin`.

I have left `ruins` off due to similarity to `collapsed`.

Still undecided on `construction`. I feel it almost warrants a new style altogether, which would be better implemented in a separate PR, so I've left it off for now.

There is a lot going on in this PR so I am thinking about making this PR only about adding minor buildings (and the query rewrite) and making a separate PR later for lightening major buildings. I think this would help get things done faster as we will be able to focus on one thing at a time. Unless anyone objects, I'll update the repository and change the title of this PR.

meased:
More testing (at 7% lighter):

https://www.openstreetmap.org/#map=17/45.58231/-122.73927
sj_b
sj_a

https://www.openstreetmap.org/#map=17/45.15308/-122.84857
wb_b
wb_a

https://www.openstreetmap.org/#map=18/45.46846/-122.73448
hv_b
hv_a

https://www.openstreetmap.org/#map=17/45.50819/-122.64908
la_b
la_a

aly_b
aly_a

meased:
@Tomasz-W wanted to see 5% lighter:

0%
la1_0p

5%
la1_5p

7%
la1_7p

0%
la2_0p

5%
la2_5p

7%
la2_7p

Tomasz-W:
Thanks! As 5% is too dark to distinguish them fastly, but 7% is minimally too light (too prominent) on residential areas, I would go with middle one of 6%.

meased:
6% version of the last image:
la2_6p

jeisenbe:
Looks good on residential landuse.

Can we try an example with minor buildings on Religious landuse?

Adamant36:
@meased, can you try a couple of tests with 6% and the new yellowish school grounds color?

kocio-pl:
I think it's important to test it on different backgrounds where one can find buildings (for example `landuse=garages` or `landuse=commercial`), not just the bare ground. They will look different there, so it's too early for precise shade tuning.

kocio-pl:
@meased My proposition is to start with just the code restructuring as a separate PR, so testing would be easier and less error prone, and next making another PR based on this which will be focused only on making minor category and visual testing the colors.

meased:
> My proposition is to start with just the code restructuring as a separate PR

I think I'll do this.

meased:
Done.

This PR is now just a rewrite of the buildings SQL and style sheet. There should be no visual differences. I have tested this in Kosmtik by flipping the overlay of the standard map on and off and I can't see any difference on any zoom level.

kocio-pl:
Great! Since there is a change from SQL code to CSS code, there might be some performance difference. @rrzefox Could you test this difference?

kocio-pl:
Well, this statement was not accurate - it's rather simpler both SQL and CSS code, so there should be no problems with performance. I will just wait with this until v4.16.0 is out (I plan to do this 19.10) and probably merge this PR soon after that. Then we can start testing minor buildings PR.

kocio-pl:
BTW: making major buildings lighter would also fix #3071.

kocio-pl:
@meased What about further tuning? Do you plan to get back to the proposed minor and major buildings changes?

meased:
The problem here is that buildings need to simultaneously serve as a background for our lightest foreground (I believe that is gastromony `#C77400`) and as a foreground to our darkest landusage (which I believe is religous `#d0d0d0`, but residential is also a problem), all while having three brightnesses of minor, normal, and major which are all distinguishable from each other.

I'm not convinced that all of these criteria can be met, hence the inactivity.

Regardless, I could start a PR so that these problems could be investigated.

matthijsmelissen:
That's the kind of thinking we need! Feel free to propose dropping/altering any of these criteria if you think it would make sense.

By the way, just curious, what's your username on openstreetmap.org?

meased:
> what's your username on openstreetmap.org?

[cowdog](https://www.openstreetmap.org/user/cowdog)

Tomasz-W:
Please remember about discussed half-transparency for some buildings (https://github.com/gravitystorm/openstreetmap-carto/issues/552#issuecomment-446540473).

Issue #3425, opened 2018-10-04
Simple 3d rendering of buildings


andrzej-r:
This is a request for discussion.

I've been thinking about implementing 3D rendering of buildings using Mapnik (rendering buildings like simple cube-like objects) to give some feedback on building heights and to make them look more like buildings.

Initial results were promising - see a screenshot below. However, I hit two road blocks:
- Wall hiding doesn't work correctly. Most examples sort buildings by coordinates before rendering. That works fine in simple cases (uniform size rectangular buildings) but fails in more complex scenarios. In the example below you can see some internal walls that shouldn't be rendered.
  - I've seen styles (e.g. osm-bexhill) that don't seem to have this issue, so perhaps a solution exists.
- If I render building below roads (as in the current openstreetmap-carto implementation) I get an unnatural effect of roads rendered on top of walls etc. If I render buildings above roads (like in the screenshot below), the buildings obscure roads. There is an option of making buildings semi-transparent, which would be a good solution to this problem, but it results in all internal walls being visible.

The [example](https://www.openstreetmap.org/#map=18/51.52293/-0.10397) below renders buildings at a fixed height and does not support major buildings. This can be fixed with extra code.

![z18-3d_buildings](https://user-images.githubusercontent.com/587630/46503900-3e09af00-c824-11e8-8b84-7213d91b9b4d.png)

andrzej-r:
Here is a screenshot with semi-transparent buildings (building-opacity=0.3). A nice feature is that undeground objects (here railway on the left) or landuse areas are still visible despite buildings being rendered above them. Unfortunately, all these sidewalls look too distracting, at least to me.

![z18-3d_buildings_transparent](https://user-images.githubusercontent.com/587630/46504740-b8d3c980-c826-11e8-98e1-6e0518cb0167.png)

andrzej-r:
It turns out that Bexhill OSM style uses a graphical trick to render drop shadows next to the buildings. That's a different technique from the one used in examples above (Mapnik building symbolizers).

Drop shadows look nice but because they require all buildings to be rendered at the same "height", so they don't improve mapping feedback. The feature can be implemented fairly easily, and has a minimum impact on the existing rendering. If there is an interest in having it, I can prepare a working PR.

![z18-3d_buildings_shadow](https://user-images.githubusercontent.com/587630/46506966-20414780-c82e-11e8-8dad-e9edff519548.png)

andrzej-r:
[...]

PR #3427, opened 2018-10-05
Added 3D buildings using Mapnik building symbolizers.

andrzej-r:
Implementation of #3425.

Buldings (3D and semi-transparent) are rendered _above_road features and land uses. This has a benefit of making land uses visible in dense urban areas. Also railway stations are now displayed correctly (not obsured by platforms and other internal objects, whilst making them still visible).

Building height are calculated using building:levels (defaults to 1) and building:height (in metres). Building:height has priority over building:levels.

There are some z-order issues due to Mapnik limitations. Hopefully will be fixed in future versions of Mapnik https://github.com/mapnik/mapnik/pull/3969.

Will probably not support complex building designs like [this one](https://www.openstreetmap.org/#map=19/31.24168/121.49560).

I've deferred rendering buildings to z15+, mostly for code simplicity, but then realized that looks better/more readable. Alternatively, we could use flat buildings at z13-14 or reduce opacity.

### DIFF
--- a/buildings.mss
+++ b/buildings.mss
@@ -9,33 +9,21 @@
 @entrance-normal: @building-line;
 
 #buildings {
-  [zoom >= 13] {
-    polygon-fill: @building-low-zoom;
-    polygon-clip: false;
-    [zoom >= 15] {
-      line-color: @building-line;
-      polygon-fill: @building-fill;
-      line-width: .75;
-      line-clip: false;
-    }
-  }
-}
-
-#buildings-major {
-  [zoom >= 13] {
+  [zoom >= 15] {
+    building-fill: @building-fill;
     [aeroway = 'terminal'],
     [amenity = 'place_of_worship'],
     [building = 'train_station'],
     [aerialway = 'station'],
     [public_transport = 'station'] {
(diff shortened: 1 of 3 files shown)


andrzej-r:
Examples from [London](https://www.openstreetmap.org/#map=16/51.5169/-0.0808).

z14 (no buildings in the current version):
3d_buildings_z14

z15:
3d_buildings_z15

z16:
3d_buildings_z16

z17:
3d_buildings_z17

z18:
3d_buildings_z18

z19:
3d_buildings_z19

Adamant36:
Although I am not against the idea per say, I think it should have a lot more discussion before being implemented and the Mapnik issue should be resolved also. From reading the comments there, it seemed like it stalled out. Also, I think @Tomasz-W's idea of it being implemented as an add on map instead of on the main is a better idea. Although, I guess it does fit the idea of the map being a showcase for OSM, but I think it might add to much clutter with the shadows etc to help the map act a feedback mechanism for mappers. So, while it might look cool in practice, I don't see it having useful application outside of aesthetics. Which is an issue.

kocio-pl:
Closing because original idea was rejected in this style.

PR #3430, opened 2018-10-06
Rendering icon for shop=food

kocio-pl:
- Rendering [shop=food](https://wiki.openstreetmap.org/wiki/Tag%3Ashop%3Dfood) with the restaurant icon, but mirrored (to make it easier to recognize)

[Node](https://www.openstreetmap.org/#map=19/52.22586/21.00867) (in comparison with restaurants)

x7qm0w6w

[Area](https://www.openstreetmap.org/#map=19/50.94857/19.85985)

9l6vfok

[Usage](http://taghistory.raifer.tech/)

taghistory 54


Tomasz-W:
I think that both icons should be mirrored once again. A correct table set is fork on the left side and soup spoon on the right side (see [here](https://www.google.pl/search?client=opera&hs=Btj&biw=1920&bih=977&tbm=isch&sa=1&ei=QWO4W-2II-2prgSvlaCwBA&q=cutlery+table+set&oq=cutlery+table+set&gs_l=img.3...6943.8297.0.8444.6.6.0.0.0.0.96.233.3.3.0....0...1c.1.64.img..3.3.231...0i7i30k1j0i7i5i30k1j0i8i7i30k1.0.q9vqM7xyY10)). It would be also logical to revert restaurant icon as dinner starts by soup and most of people are right-handed so spoon should be on the right side.

jeisenbe:
I’m confused by this tag. What is the difference between this tag and
shop=grocery or =greengrocer or =general etc?

Is it being used by non-English speakers who do not know the word grocery?
Is it being used for supermarkets or general shops? Or Delis?

Taginfo shows it is used in England and the USA, but it seems more popular
in Central Europe, Japan and the Phillipines. I couldn’t find info on
combinations used.

I checked overpass-turbo in London. The first few shop=food on the list are
“Nana’s Deli“, “The Chelsea Cake Shop”, “Green Apple”, “Holland And
Barrett”...

Those should be shop=deli, shop=bakery, shop=greengrocer(?). The fourth is
a vitamin and herbal supplement shop per google. So all 4 are mistagged.
Farther down the list are a few shops with description “Bulgarian Food” and
another “African and Zim food”, so could be restaurants or grocers, I
haven’t confirmed

In New York, overpass turbo shows:
1 “Busy Bee” note=Polish,
2 “Whole Foods Market”
3 “Caputo’s Fine Foods”

According to Google, #1 is a grocery store, #2 is a supermarket,  #3 is an
Italian Deli and grocer

I would recommend not rendering it at this time.

Joseph

Adamant36:
@jeisenbe, probably just a more generic, default type tag for "I don't know what exactly this place is, but I know it involves food somehow." Id say render it just for that. Its better then nothing. Plus, its not technically wrong. Since the places are food shops.

jeisenbe:
Can we try an icon that is more food-like?
Greengrocer has an apple icon, for example.
The utensils imply that you can eat there, but these shops are not
restaurants.

meased:
Perhaps reusing the `shop=convenience` icon? They are somewhat like a specialized convenience store (walk in, buy a few items, and leave without eating it).

kocio-pl:
Before I start digging into @jeisenbe questions, which are worth checking, but require some time and reading about English vocabulary, I can propose a shop bag with a fork for example (similar to charity shop). 

I also think that reworking restaurant icon is perfectly possible. It's just good to think also about food court - currently they look the same. Maybe they could be a bit different. I thought about showing it with this icon on z17, instead of a dot, to show that it's bigger space, but we can also do something more.

matthijsmelissen:
I think shop=food is almost always a mistagging of shop=supermarket, shop=convenience or shop=deli.

kocio-pl:
@jeisenbe Thanks a lot for writing about it on the Tagging list! :heart: I'd be happy to know the differences at last...

https://lists.openstreetmap.org/pipermail/tagging/2018-October/039710.html

jeisenbe:
No problem!

I also looked up the uses of shop=grocery on overpass turbo for London.

These look a little better. There are several specialty grocers that sell
Greek, African or Caribbean nonperishable food items
A few places on the list could also probably be shop=greengrocer or
shop=convenience.
And some are groceries but are closed now! We really need to be better
about keeping the map up to date.

I think shop=grocery may be the better of the two tags, but I'll let the
Brits weigh in on the mailing list about "proper" English usage.

dieterdreist:
sent from a phone

> On 6. Oct 2018, at 12:59, jeisenbe <notifications@github.com> wrote:
> 
> I would recommend not rendering it at this time.


I agree, this is a tag with very few usage (compared to other tags with similar meaning), unclear definition and generally disputed in the wiki: https://wiki.openstreetmap.org/wiki/Tag%3Ashop%3Dfood


please do not render it

dieterdreist:
sent from a phone

> On 6. Oct 2018, at 18:52, meased <notifications@github.com> wrote:
> 
> They are somewhat like a specialized convenience store (walk in, buy a few items, and leave without eating it).


who knows, they can also be something like a fast food place, it remains completely open from the tag name and definition in the wiki

jragusa:
I agree to reject this rendering but as @Tomasz-W suggested, we should flip the restaurant icon.

kocio-pl:
I don't think it's worth digging more now, there are other interesting PRs to take care of, but for a shop selling only home food (like dumplings+pancakes) ready to boil I don't think convenience or grocery (let alone supermarket!) makes any sense - and general is too general for this.

It's also good that we have such generic values - it's like building=yes: if I know only this, it's better to tag it according to my knowledge, maybe it will be tuned later.

Thanks for all the comments!

Tomasz-W:
@kocio-pl What about flipping amenity=restaurant icon due to reasons mentioned in https://github.com/gravitystorm/openstreetmap-carto/pull/3430#issuecomment-427553657?

kocio-pl:
I have nothing against it, I'm just focused on other issues. Typically this is knife and a fork, but spoon also works for me.

Tomasz-W:
Restaurant icon updating proposals:

* amenity restaurant2 (fork + spoon)
* amenity restaurant2 (fork + knife)

PR #3432, opened 2018-10-06
Render capacity next to bicycle_rental and parking markers.

### LINKED ISSUE #3428: Render capacity of amenity=bicycle_rental and amenity=bicycle_parking nodes
It would be nice to see the capacity tag of bike sharing stations and bicycle parking places rendered as text.

The capacity text should probably only be visible at at z>=18, so there will be enough free space around the node (bicycle_parking is only rendered at z=18 and above).

I will try to submit a PR with this change when I have time, but if someone beats me to it, I won't mind.

tpikonen:
Add 'capacity' tag to amenity-points-poly, amenity-points,
amenity-low-priority and amenity-low-priority-poly layers.

Render capacity text as a superscript to (up and to the right of) the
markers of amenity=bicycle_rental, parking, bicycle_parking and
motorcycle_parking at z>=18.

Fixes #3428.

Test rendering from a mall with several bicycle parkings, underground car park and a citybike station [here](https://www.openstreetmap.org/#map=18/60.15993/24.88132) (My test dataset is ~1 year old, so the positions have changed somewhat).

z=18:
lauttis_z18

z=19, showing also the name label of the citybike station:
lauttis_z19

The capacity text is a little bit too much to the right of the bicycle_parking marker IMO, but reducing the text-dx property apparently results in an overlap which hides the label. Markers for bicycle_rental and parking look fine to me. Hints on how to fix this would be appreciated.

### DIFF
--- a/amenity-points.mss
+++ b/amenity-points.mss
@@ -96,6 +96,19 @@
     marker-fill: @transportation-icon;
     marker-placement: interior;
     marker-clip: false;
+    [zoom >= 18] {
+      text-name: "[capacity]";
+      text-size: @standard-font-size;
+      text-wrap-width: @standard-wrap-width;
+      text-line-spacing: @standard-line-spacing-size;
+      text-fill: @transportation-text;
+      text-dy: -0.5;
+      text-dx: 8;
+      text-face-name: @standard-font;
+      text-halo-radius: @standard-halo-radius;
+      text-halo-fill: @standard-halo-fill;
+      text-placement: interior;
+    }
   }
 
   [feature = 'highway_bus_stop'] {
@@ -1468,8 +1481,25 @@
     marker-placement: interior;
     marker-clip: false;
     marker-fill: @transportation-icon;
+    [zoom >= 18] {
+      text-name: "[capacity]";
+      text-size: @standard-font-size;
+      text-wrap-width: @standard-wrap-width;
+      text-line-spacing: @standard-line-spacing-size;
+      text-fill: @transportation-text;
+      text-dy: -0.5;
+      text-dx: 8;
+      text-face-name: @standard-font;
+      text-halo-radius: @standard-halo-radius;
+      text-halo-fill: @standard-halo-fill;
+      text-placement: interior;
+    }
     [access != ''][access != 'permissive'][access != 'yes'] {
       marker-opacity: 0.33;
+      [zoom >= 18] {
+        text-opacity: 0.33;
+        text-halo-radius: 0;
+      }
     }
   }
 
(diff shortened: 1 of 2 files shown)


kocio-pl:
Thanks for a real rendering test, however I feel that showing capacity does not work as a substitution for showing available spaces. You need application which is counting free space in the real time, which is well beyond not only OSM Carto, but also any static map.

From my experience in Warsaw, most of small (<5) bicycle parkings have always some free slots, while a big bicycle parking near a shopping center is almost fully loaded most of the time. For a bicycle rental stations - you can check available bikes straight at the system website:

https://www.veturilo.waw.pl/en/

And if you want to show how big is that area, just draw it, like here:

https://www.openstreetmap.org/way/486558720

Adamant36:
@tpikonen, it might worth be worth starting a MapRoulette task or something to map more of the bike parking as ways. As it currently is only 33,000 out 233,000 are mapped that way. Judging how many bike racks there are by the size of the area is the next best thing to having a capacity number. Its not really possible in most cases though going by the numbers. Another good option is seeing if OsmAnd will support capicity if they dont already. As its more relevent to someone out and about using an app anyway. Also, it would be worth using Mappilary to take photos of the bike racks in your area if you can. So people can use that for a reference if nothing else. I do it sometimes for things where I live that arent supported by the main site that I still want a reference to. Maybe someday it will get integrated into the main site.

HolgerJeromin:
> Another good option is seeing if OsmAnd will support capicity if they dont already.

When you tap a poi it shows the capacity in the details view

Adamant36:
@HolgerJeromin, thanks for the heads up. All the more reason not to show it here. Especially since the cycle map shows it also apparently.

mboeringa:
> Test rendering from a mall with several bicycle parkings, underground car park and a citybike station [here](https://www.openstreetmap.org/#map=18/60.15993/24.88132) (My test dataset is ~1 year old, so the positions have changed somewhat).

Just for the record, in the Netherlands, we have official >1000 capacity bicycle parkings at almost every big city railway station. Utrecht, which is a central railway hub, has a brand new 12500(!) three-story underground bicycle parking build to the standards of an underground car parking, in which you can easily get lost (or at least totally forget where you left your bike if you don't memorize the parking lane's number ;-( ):

https://www.youtube.com/watch?v=L3M_GM_MDg8
https://www.youtube.com/watch?v=kGv7Nr6QQJI

matthijsmelissen:
I agree with @kocio-pl , I don't think this will help the map user a lot. Anecdotically, in The Hague we have a bike parking with capacity of 4100 which is almost always full. So I'll have to reject this proposal. Nice proof of concept though!

PR #3434, opened 2018-10-06
add icon for amenity=vehicle_inspection

### LINKED ISSUE #3378: Render amenity=vehicle_inspection
This is a growing tag closing to 2k limit and has a wiki page. I think that icon should be brown and contain car, probably with a seal of approval.

https://wiki.openstreetmap.org/wiki/Tag%3Aamenity%3Dvehicle_inspection

taghistory 38

Adamant36:
This PR adds an icon for amenity=vehicle_inspection. To quote @kocio-pl, it is a growing tag closing to 2k limit and has a wiki page. Closes #3378 
taghistory
https://www.openstreetmap.org/#map=19/40.35227/-122.28071 (node)
node
https://www.openstreetmap.org/#map=19/37.59953/-122.38768 (way)
way


kocio-pl:
Thanks, I like it. BTW: it already has 2149 uses now.

kocio-pl:
Could you test a node with a label to make sure that label rendering is fine there too?

matthijsmelissen:
How will places that are both shop=car_repair and amenity=vehicle_inspection be rendered? In some countries vehicle inspection is done by regular garages, in others it is done by the government. Is this allowed tagging?

dieterdreist:
sent from a phone

> On 7. Oct 2018, at 23:14, Matthijs Melissen <notifications@github.com> wrote:
> 
> In some countries vehicle inspection is done by regular garages, in others it is done by the government. Is this allowed tagging?


why not? As long as the definitions fit for the thing it is ok

Adamant36:
https://www.openstreetmap.org/#map=19/37.88545/-122.27402 (node with label)
node with name

kocio-pl:
Thanks, it's working good in my test.

boothym:
How will an object tagged as both shop=car_repair and amenity=vehicle_inspection be rendered?

Adamant36:
If I remember correctly, in my tests it rendered as vehicle inspection. Which seems like the proper rendering for me. Since a lot of places that do inspections might also do "repair" type things, but its usually secondary. Like places that check oil.

kocio-pl:
We have general problem with multiple renderable tags. Rendering depends on general categories in any given layer `project.mml`. For example `amenity-points-poly` layer checks for `amenity` first, and `shop` later, which means that it stops searching once it finds matching tag (hence amenities have priority over shops when tagged on the same object):

https://github.com/gravitystorm/openstreetmap-carto/blob/0ef4eb728b2a4b93448a50b39e8ffd0436ef6c2e/project.mml#L1473-L1501

I have no idea how to make it more specific without going insane. You can see using two sets of `tourism` in the beginning (for example `artwork`) and at the end of this list (`viewpoint` alone), most probably just to make some of such objects being selected even if they have `amenity` for example and some others (vieewpoint) only as a last resort, when no other renderable tags are found, but this solution is ugly and error prone.

polarbearing:
@Adamant36 - the situation is the opposite here in Europe. The inspection is a legal requirement every 2 years, and performed by institutions authorised for this job. They do not do any repair at all. There are repair shops which offer the formal inspection as an extra service, but then the authorised inspector visits the shop upon agreement. Thus in the European cases, shop=car_repair would have to prevail. 

@kocio-pl - it would be desirable to have an mechanism where we can formally define which of two conflicting tags wins the rendering. Otherwise these issues become forgotten after a while, and resurface as a side-effect once some layering etc is modified for other reasons.

HolgerJeromin:
> it would be desirable to have an mechanism where we can formally define which of two conflicting tags wins the rendering. Otherwise these issues become forgotten after a while, and resurface as a side-effect once some layering etc is modified for other reasons.

This would be a classic case for an automated regression test.  I opened #3446 for that

polarbearing:
Checking https://taginfo.openstreetmap.org/tags/amenity=vehicle_inspection#map shows the tag is mostly used in Europe where the legal requirement exists. I checked the wiki, and it clearly defines the legal safety checks and not the operational things like oil level. 
No combinations are shown in taginfo, so overpass reveals about 100 cases with inspection+repair. 
The 10 I probed in DE are all tagging errors, either trying to make the formal inspection sites visible in the renderer, or describing the additional service by a repair shop to have a visiting inspector.

dieterdreist:
sent from a phone

> On 11. Oct 2018, at 08:30, polarbearing <notifications@github.com> wrote:
> 
> They do not do any repair at all. There are repair shops which offer the formal inspection as an extra service, but then the authorised inspector visits the shop upon agreement. Thus in the European cases, shop=car_repair would have to prevail.


this is the German situation, it is not the same in all European countries AFAIK.
I agree though that checking oil is a completely different pair of shoes than European car inspection (the former takes a minute and is done in many places including gas stations, the latter requires a special license / authorization and takes much more time, as all safety and environment related features are checked (lights, brakes, rust on load bearing parts, honk, exhaust, suspension, ...)

matthijsmelissen:
> Since a lot of places that do inspections might also do "repair" type things, but its usually secondary.

In the Netherlands it's opposite. Almost all car repair shops are licensed to do inspections too, but it's not their primary purpose. I'm not sure if these should be tagged with amenity=vehicle_inspection_station though.

> The inspection is [...] performed by institutions authorised for this job. They do not do any repair at all.

Certainly not true in the Netherlands.

trigpoint:
> > Since a lot of places that do inspections might also do "repair" type things, but its usually secondary.
> 
> In the Netherlands it's opposite. Almost all car repair shops are licensed to do inspections too, but it's not their primary purpose. I'm not sure if these should be tagged with amenity=vehicle_inspection_station though.
> 
> > The inspection is [...] performed by institutions authorised for this job. They do not do any repair at all.
> 
> Certainly not true in the Netherlands.
Same in England/Wales/Scotland. Most garages (car repair) do vehicle testing but their primary business is service and repair. Vehicles 3 years old and over have to be tested every year.

jragusa:
It's a little bit off-topic (e.g. tagging discussion) but following the concept of [One feature, one OSM element](https://wiki.openstreetmap.org/wiki/One_feature,_one_OSM_element), `amenity=vehicle_inspection` should be displayed in a different node than `amenity=car_repair` in this case. It's the same problem with hotels encompassing a restaurant.

dieterdreist:
sent from a phone

> On 14. Oct 2018, at 14:11, Jérémy Ragusa <notifications@github.com> wrote:
> 
> It's a little bit off-topic (e.g. tagging discussion) but following the concept of One feature, one OSM element, amenity=vehicle_inspection should be displayed in a different node than amenity=car_repair in this case. It's the same problem with hotels encompassing a restaurant.


these are just 2 examples that „one feature one osm element“ isn’t a useful rule to decide anything. Things become a feature if you decide they are a feature. In the real world you can argue that a hotel and a restaurant are 2 things, or are one combined thing, and there will be arguments for both points of view.

PR #3246, opened 2018-05-26
Render remaining tourism_information features

jragusa:
Fixes #304
Fixes #324
Fixes #1978

Changes proposed in this pull request:
This PR completes #3198 by adding rendering of `audioguide`, `board`, `map`, `office`,` tactile_map`, `tactile_model` and `terminal`

Test rendering with links to the example places:

`information=audioguide`
https://www.openstreetmap.org/node/3066454413
tourism_audioguide

`information=board`
https://www.openstreetmap.org/node/2857924837
script-version
tourism_board
i-version
tourism_board2

`information=map`
https://www.openstreetmap.org/node/4939116583
tourism_map

`information=office`
https://www.openstreetmap.org/way/130442457
z=17
tourism_office_z17
z=19
tourism_office_z19
with other brown amenities
https://www.openstreetmap.org/node/5537445277
tourism_office

`information=tactile_map`
https://www.openstreetmap.org/node/3571444993
tourism_tactile_map

PR #3437, opened 2018-10-07
Fix icon for information=tactile_model

jragusa:
Replace board icon by map icon for `information=tactile_model ` following [comment](https://github.com/gravitystorm/openstreetmap-carto/pull/3246#issuecomment-427578416) of  @Tomasz-W from #3246

https://www.openstreetmap.org/node/4362875819 (the land of gift shops)

Before
tactile_model_before

After
tactile_model_after

### DIFF
--- a/amenity-points.mss
+++ b/amenity-points.mss
@@ -351,8 +351,7 @@
     [information = 'audioguide'] {
       marker-file: url('symbols/tourism/audioguide.svg');
     }    
-    [information = 'board'],
-    [information = 'tactile_model'] {
+    [information = 'board'] {
       marker-file: url('symbols/tourism/board.svg');
     }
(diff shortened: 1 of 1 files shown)


matthijsmelissen:
I think this is getting to detailed - and I don't think many people are going to use our icons to look for tacticle maps.

kocio-pl:
It's a small fix, but I think it's just closer to the map than board, so I would fix it.

kocio-pl:
Thanks for fixing this detail! Tested and it works.

>  I don't think many people are going to use our icons to look for tacticle maps

Even if this is small issue indeed, I think this might be important difference in how we see the map: I am not **looking for** things on a static map, rather I **look what is there**.

PR #3438, opened 2018-10-07
Rendering of strait from z10

### LINKED ISSUE #804: Add rendering for natural=strait

jragusa:
Fixes #804

Add label to `natural=strait` from z10

https://www.openstreetmap.org/relation/2667319

Before
strait_before

After
strait_after

### DIFF
--- a/project.mml
+++ b/project.mml
@@ -2019,7 +2019,7 @@ Layer:
             COALESCE(
               'landuse_' || CASE WHEN landuse IN ('forest', 'military', 'farmland') THEN landuse ELSE NULL END,
               'natural_' || CASE WHEN "natural" IN ('wood', 'glacier', 'sand', 'scree', 'shingle', 'bare_rock',
-                                                    'water', 'bay') THEN "natural" ELSE NULL END,
+                                                    'water', 'bay', 'strait') THEN "natural" ELSE NULL END,
               'place_' || CASE WHEN place IN ('island') THEN place ELSE NULL END,
               'boundary_' || CASE WHEN boundary IN ('national_park') THEN boundary ELSE NULL END,
               'leisure_' || CASE WHEN leisure IN ('nature_reserve') THEN leisure ELSE NULL END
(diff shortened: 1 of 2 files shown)


jragusa:
Node: https://www.openstreetmap.org/node/619215170

z=15
strait_point_z15

kocio-pl:
Great, thanks for picking it up!

Please make a test on some other node, because this one renders `place=locality` label (it should be probably removed as a tagging for rendering):

https://www.openstreetmap.org/node/619215170

jragusa:
Oups, I missed that. Corrected rendering from z17:
strait_point_z17_corrected

jeisenbe:
Is this only on nodes? Will it also render on linear ways?

The wiki suggests using a way for long, narrow straights. There are 1001
nodes, 600 ways.

kocio-pl:
I'm not aware of any linear rendering for such things, only nodes and areas (closed ways or relations) - this code is shared for bays for example.

jeisenbe:
The wiki page claims:

   - "when a strait is long, narrow and curved or has side channels or an
   otherwise complex and ambiguous coastline form it often makes sense to map
   it as a linear way. This should roughly follow the centerline of the water
   channel like in case of the Matochkin Strait
   <https://www.openstreetmap.org/way/163242449>" (link to OSM way
   163242449)
   - "As of March 2018 there are about 760 straits mapped with nodes, 540
   with linear ways and 290 with either multipolygons or closed ways"


So a linear way / line down the middle of the straight is almost as popular
as a node.

Rendering the lable at the center of the line or center of the area should
be fine, if that's possible?
In the future we could consider rendering the name alongside the linear
way, like was suggested for ridges and valleys, or like what is currently
done for waterways
But I'm happy with a lable at a central point. I just wanted to check if
the "areas" and linear ways would get rendered too.

-Joseph

kocio-pl:
I think adding linear rendering is a task for a separate PR (it will be good to check if some other similar objects might be rendered like that).

kocio-pl:
Thanks a lot for this code! Works as expected.

Now if anybody would like to research how to support linear shapes, it would be next big step...

geozeisig:
I found examples of areas ([example](https://www.openstreetmap.org/way/384457932)) but none for ways ([example](https://www.openstreetmap.org/way/486310481))? Maybe it will take time. Would it not be better to do the writing in the direction of the way? As it is already done at the ferry lines.

jragusa:
@geozeisig, look at the cap horn for ways
strait_overpass

geozeisig:
It looks like `natural=strait` is not rendered on ways.

HolgerJeromin:
This was [mentioned](https://github.com/gravitystorm/openstreetmap-carto/pull/3438#issuecomment-427824977) before:
> I think adding linear rendering is a task for a separate PR (it will be good to check if some other similar objects might be rendered like that).

geozeisig:
Rendering on linear ways could be done the same as with rivers. This could determine the direction of the font. It would be repeated at long ways. Only the font size could be larger in many places, which could be a problem to assess correctly.

PR #3441, opened 2018-10-07
Landcover order

### LINKED ISSUE #2782: Should landcover be ordered by osm_id in addition to way_area?

matthijsmelissen:
This resolves #2782


matthijsmelissen:
Is anybody willing to review this change?

sommerluk:
Haven’t tested it yet, but the code looks good to me.

PR #3443, opened 2018-10-09
Add rendering for leisure=track on lines

### LINKED ISSUE #2632: Rendering leisure=track on lines
We are rendering `leisure=track` on areas but not on ways - neither circular nor linear.  With the database reload we are also not going to interpret closed ways as areas by default any more.

Adding line rendering for tracks would be possible and would make sense.  It could be done in a way that is visually similar and compatible to area rendering with a fill and casing in the same colors as used for areas.

See also #574, #2238, #1504, #2533.

Adamant36:
This PR adds rendering for leisure=track on lines. As it is they are currently rendered on areas but not on ways. Closes #2632 
https://www.openstreetmap.org/#map=18/39.19833/-122.02187 (location 1)
track loction 1
https://www.openstreetmap.org/#map=18/39.14384/-121.67329 (location 2)
track location 2
There is also some other related issues in #2632 that it potentially closes, but I wasn't sure which ones. So I can add them later. Also, the line is a little thin and tracks are usually pretty wide. So maybe it could be made wider.

### DIFF
--- a/amenity-points.mss
+++ b/amenity-points.mss
@@ -2948,6 +2948,39 @@
     marker-fill: @transportation-icon;
   }
 
+  [feature = 'leisure_track'] {
+    [zoom >= 16] {
+      [zoom >= 17] {
+        bridgecasing/line-color: saturate(darken(@pitch, 30%), 20%);
+        bridgecasing/line-join: round;
+        bridgecasing/line-smooth: 1;
+        bridgecasing/line-width: 1.25;
+        [zoom >= 18] { bridgecasing/line-width: 2.5; }
+        [zoom >= 19] { bridgecasing/line-width: 5; }
+      }
+      line-color: @pitch;
+      line-join: round;
+      line-cap: round;
+      line-smooth: 1;
+      line-width: 1;
+      [zoom >= 18] { line-width: 2; }
+      [zoom >= 19] { line-width: 4; }
+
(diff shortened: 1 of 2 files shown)


Adamant36:
@tomasz-w, good call. I agree.

kocio-pl:
Thanks, this was a long missing feature.

Tomasz-W:
I've just found that there are no bridge rendering for leisure=track lines, Here is an exemple why it should be added for better readability. 

https://www.openstreetmap.org/way/327635778

@Adamant36 Could you take care of that?

Adamant36:
I'll look into it. Its also tagged as a summer toboggan. Which is something ive been meaning to work on. I think the track tag is wrong in this just and just so it will render.

Adamant36:
Well, I guess it says to add track to it on the wiki page. I think its a stretch though since summer toboggans aren't racing tracks per say. 

It also begs the question, if all of the summer toboggans are suppose to tag as tracks, then whats the point in added rendering for them? If there is none because they are covered by track rendering (which personally I dont think they should be) you might want to request the summer toboggan be closed. I feel like the track tagging is worth looking into more though just in case.

Adamant36:
@Tomasz-W, can you open a new for it please so it doesn't get lost and people can discuss it if they want?

PR #3444, opened 2018-10-09
Change color for apron in aerodromes

### LINKED ISSUE #3385: Apron color in aerodromes is too prominent
The airport apron is the area of an airport where park (for loading, unloading, refueling etc). Aprons are generally not open to the public, and are not features of interest on a general map style.  They are similar in use to aeroway=taxiway and amenity=parking, therefore it should be rendered in a similar way. 

But the tag aeroway=apron is currently rendered in a rather bright purple color, in contrast to the dark gray of taxiways and runways. The labels are more prominent than those of the terminal buildings.

I would recommend changing aprons to render in a light to mid-gray tone, similar to taxiways or private parking lots. 

Wikipedia article with definition: [https://en.wikipedia.org/wiki/Airport_apron](https://en.wikipedia.org/wiki/Airport_apron)
OSM Wiki article: [https://wiki.openstreetmap.org/wiki/Tag:aeroway%3Dapron](https://wiki.openstreetmap.org/wiki/Tag:aeroway%3Dapron)
Example of current rendering at Munich International Airport: [https://www.openstreetmap.org/#map=16/48.3526/11.7877](https://www.openstreetmap.org/#map=16/48.3526/11.7877)

(Edit: fixed link to Munich airport, mentioned label prominence issue, added image)

Adamant36:
This PR changes the color for aprons in aerodromes. As they are currently to prominent. Closes #3385 
apron before 1
apron after 1
apron before 2
apron after 2
after before 3
apron after 3

### DIFF
--- a/landcover.mss
+++ b/landcover.mss
@@ -28,7 +28,7 @@
 // --- Transport ----
 
 @transportation-area: #e9e7e2;
-@apron: #e9d1ff;
+@apron: #dadae0;
 @garages: #dfddce;
 @parking: #eeeeee;
 @parking-outline: saturate(darken(@parking, 40%), 20%);


Tomasz-W:
@jeisenbe @kocio-pl Comparing it to landuse=residential in the the last test rendering, I'm worried that this colour may be too light and taken as some cultural-like area rendering (like eg. landuse=residential or landuse=industrial etc.), but this is a landcover feature and we usually render these ones darker. 
@Adamant36 Can you try with darker `#ceced9`?

kocio-pl:
I have no such worries - this is one of the rare objects that have very clearly defined context,nobody will expect residential area on the airport, especially directly connected to the aeroways... However testing alternatives is always possible and I'm glad that you care.

kocio-pl:
Comparison:

original

923oeels

#ceced9

etydflji

Adamant36:
I don't know about #ceced9. I was thinking there might be a chance the other color is to close to residential, but I think #ceced9 is slightly to dark and a little on the purple side.

kocio-pl:
Thanks for another PR solving the issue! 

I also think lighter color is better and it didn't look like any discussion is still going on, so I merged it.

PR #3445, opened 2018-10-09
Move amenity=post_box to z19+

### LINKED ISSUE #3431: Move amenity=post_box to z19+
Now that ATMs are rendered at z19+ I think post boxes should also be. As they suffer from a lot of the same issues at higher zoom levels as ATMs did. I noticed @kocio-pl mentioned them in issue #1745 and the reasons are still valid. 

My own reasons for wanting it at z19+ is that they clutter things up to much otherwise. Plus, the icon blocks more important icons and names at higher zoom levels. Also, its a small feature that is not used that often, and is only important on a neighborhood or even block level.  Therefore, rendering it at z19+ just makes sense. Especially in more dense metro areas, where a lot of them are mapped. 

https://www.openstreetmap.org/#map=17/38.77231/-121.36215 (clutter)

https://www.openstreetmap.org/#map=17/37.78930/-122.40474 (blocking more important things)

Adamant36:
This PR moves post box's to z19+. There are multiple reasons why this is a good idea. A few are that they clutter things up to much and block more important icons and names at zoom levels above 19. Also, its a small feature that is not used that often, and is only important on a neighborhood or even block level most of the time. Closes #3431
post box z17 before
post box z17 after
san fransico post box
san franisco post box 2

### DIFF
--- a/amenity-points.mss
+++ b/amenity-points.mss
@@ -643,7 +643,7 @@
     marker-clip: false;
   }
 
-  [feature = 'amenity_post_box'][zoom >= 17] {
+  [feature = 'amenity_post_box'][zoom >= 19] {
     marker-file: url('symbols/amenity/post_box.svg');
     marker-fill: @amenity-brown;
     marker-placement: interior;
--- a/landcover.mss
+++ b/landcover.mss
@@ -28,7 +28,7 @@
 // --- Transport ----
 
 @transportation-area: #e9e7e2;
-@apron: #e9d1ff;
+@apron: #dadae0;
 @garages: #dfddce;
 @parking: #eeeeee;
 @parking-outline: saturate(darken(@parking, 40%), 20%);
--- a/landcover.mss
+++ b/landcover.mss
@@ -28,7 +28,7 @@
 // --- Transport ----
 
 @transportation-area: #e9e7e2;
-@apron: #dadae0;
+@apron: #e9d1ff;
 @garages: #dfddce;
 @parking: #eeeeee;
 @parking-outline: saturate(darken(@parking, 40%), 20%);


dieterdreist:
the area you have chosen seems strange, I guess the tag is not used
according to the definition. In a usual setting, there is no cluttering of
post boxes, never seen it.

kocio-pl:
I aree, some more typical area would be better to show the real problem.

Adamant36:
@dieterdreist, I addressed that in the original issue. The wiki says a post box "is a physical box used as a facility for the public to deposit outgoing postal items" As we do not live in that area, there is no way to determine if it fits the wiki definition. If not though, its a tagging issue anyway. Which I constantly get reminded this isn't the place for. 

that being said, I did check a few of them on Bing Street Side and they had outgoing slots. Plus, it looked like they were based on the per neighborhood level. So, I think they fit the wiki. If they don't, there was also the San Francisco example. I'm sure there are others, which I'm happy to provide. 

Lets say its just a few mail boxes somewhere though, they are still  not that important overall and block other things that are more important to show at z17, like restaurants or stores. Like in San Francisco. Plus, its just pointless clutter at that zoom level either way. Otherwise, provide some evidence of where a mail box should take precedence over a restaurant and where unnecessary clutter is a good thing. 

Or show me some style guide somewhere, that's not from imagico, about why exactly its so important to show post boxes or similar items at z17 instead of z19. I read through the whole ATM discussion multiple times and I never saw a compelling reason against why they shouldn't be at z19 and they very similar tags/items.

dieterdreist:
Am Di., 9. Okt. 2018 um 14:30 Uhr schrieb Adamant36 <
notifications@github.com>:

> @dieterdreist <https://github.com/dieterdreist>, I addressed that in the
> original issue. The wiki says a post box "is a physical box used as a
> facility for the public to deposit outgoing postal items" As we do not live
> in that area, there is no way to determine if it fits the wiki definition.
> If not though, its a tagging issue anyway. Which I constantly get reminded
> this isn't the place for.
>

Yes, I believe it is a tagging issue. This issue tracker is not for
discussing tagging issues, but it is clear you shouldn't use areas with
presumed tagging errors for test renderings. Just use any other area.

Adamant36:
@dieterdreist, as can be seen in the picture provided by @kocio-pl of an OverPass Turbo search, there are six post boxes that are not even displaying and a few more that are blocking other more important items. 
san franisco post box 2

Adamant36:
All either not displaying correctly or unnecessary clutter at z17
London
london 1
london 2
Paris
paris
Brussels
brussels

dieterdreist:
whether you call these "clutter" and think that other POIs that may not be
shown at some places are more relevant, depends how much importance you
assign to post boxes, what may also depend on how many there are in the
area. By some metrics they are clearly important (usable 24/24, interesting
for everybody).

Anyway, "omitted" post boxes are hardly an argument for moving them to
another zoom level, I'd rather argue these are situations that are already
"resolved" with the current rules, according to what you want to accomplish.

Adamant36:
First, a lot are in those areas and id bet most of them aren't being used by people that don't live close to them. So, although they might be important to people in those neighborhoods, they probably aren't that important to other people that would be looking at the map at z17. We don't really know though and I'm not saying "important" in that way anyway. I mean important from a cartographic and mapper perspective. Although, that is a factor. As well as the size thing. 

The map doesn't display cafes at the same level as mountains. A vast majority of people could care less about some random mountain, but they care a lot about where the local Starbucks is. So, why isn't Starbucks rendered at z11? In other words, popularity is a crappy metric and things are already being rendered based on metrics besides it anyway. So why dwell on it as a metric with this one particular issue? What specific problem do you have with showing these at only z19? Why is it even an issue when there are already other similar things only being rendered at that level?

Also, I think some of them not showing up at z17 is a perfectly valid reason for moving them to another zoom level. As its buggy and inconsistent otherwise. You might be fine with that because of whatever dogmatic render rule your stuck on, but I'm not. Further, how exactly are these situations already "resolved" by how it currently is? Rendering should be consistent and show what's there. If people expect post boxes to show at z17 because that's how they are coded, then that's what should happen. If not, we should fix it. I wouldn't called that "resolved" though. You shouldn't use a bug as an excuse to keep doing the same dogmatic thing. Otherwise, would you accept a bug like that If it had nothing to do with this issue? I wouldn't. Its odd you don't care about inconstant rendering, bugginess, or clutter in this situation.

boothym:
I don't agree with moving them to z19, maybe z18 would be better. If you don't mind, could you show us a before and after image to see what the areas you've shown look like at z17 and z18 without their post boxes?

Because looking at that screenshot of London a couple of the post boxes don't actually show at z17 as other POIs (a restaurant and a telephone box) actually hide the post box, e.g. https://www.openstreetmap.org/node/312159074#map=17/51.52189/-0.13850 Is there some way to give post boxes lower priority than other things?

Adamant36:
@boothym, z18 was brought up, but kocio-pl gave a reason why z19 is better/neccesary in the original issue. I think its the last post I agree with him. Priority rendering is its own issue, which from my understanding is hard to code. Its inconsistent anyway. Puting them at z19 is essentially giving them a lower priority anyway. Plus, if they arent being rendered half the time on z17 already, moving them to z19 is essentially a superficial change at best in a lot of cases. Since its already the only zoom where all of them can be seen consistently.

dieterdreist:
sent from a phone

> On 11. Oct 2018, at 01:55, Adamant36 <notifications@github.com> wrote:
> 
> Priority rendering is its own issue, which from my understanding is hard to code. Its inconsistent anyway. Puting them at z19 is essentially giving them a lower priority anyway.


tuning the same zoom level priorities should have precedence before more drastic measures like banning things completely from a zoom level. It is not utterly complicated, just a bit more than changing zoom levels.

kocio-pl:
I see it in a different way: some basic things were modeled using size rule first, and when there was a free space, some smaller things were exaggerated - like post boxes or highways. 

For highways adding more data on medium zoom levels didn't change much, because there are mostly natural and semi natural areas, which work as a background (when not too dark), and borders (for objects of a given size - not all of them!). So no problem, they can look much bigger (wider) than they really are, this exaggeration is still useful.

For post boxes there are a lot more data of the same type (POIs), so after few years this exaggeration stopped working properly and - together with some other small objects - became the visual clutter (see https://github.com/gravitystorm/openstreetmap-carto/pull/3372#issuecomment-426308795).

kocio-pl:
Thanks for this code, it's next step that helps to de-clutter map a bit. There was a big discussion about size principle in #3372 and I think this is the way to go.

For everybody which are worried about areas in the outdoor - the crucial problem is finding proper solution for #1957. This is important for me and I'd be happy to make almost every type of POI rendering much earlier (including post boxes) but currently we lack the idea how to deal with it.

Issue #846, opened 2014-08-05
Add rendering for barrier points


matthijsmelissen:
barrier=stile is not rendered. It should be where ever gate is being shown. Knowing where stiles are on country footpaths is useful when walking.

See also https://trac.openstreetmap.org/ticket/3546.

math1985:
I don't think the UTF symbols are very clear. Does anyone have a better icon for stiles and kissing gates? Alternatively, does anyone know of any rendering that renders these?

althio:
from http://wiki.openstreetmap.org/wiki/Tag:barrier%3Dkissing_gate ![image](https://cloud.githubusercontent.com/assets/9863567/5854849/6a5e0ec8-a232-11e4-9ced-908abfaf97c1.png)
from http://wiki.openstreetmap.org/wiki/Tag:barrier%3Dstile ![image](https://cloud.githubusercontent.com/assets/9863567/5854869/87a8478c-a232-11e4-9545-e541f2b92b76.png)

SomeoneElseOSM:
For stiles, https://github.com/SomeoneElseOSM/openstreetmap-carto-AJT/blob/master/symbols/stile.png works for me.  That's not SVG*; it's the same size as the graphic for gate that was in OSM-carto in May last year, and is designed to look "a bit like the gate icon".

There's also a cycle barrier and horse stile icon there too (currently experimenting with no transparency on the latter to make it a bit more prominent - the jury's still out on that).

I've not used a specific kissing gate icon because (compared to a normal gate) it doesn't restrict different traffic to that restricted by a normal gate (this'd be different if kissing gates occurred with any frequency on wheelchair routes of course, but I don't believe that they do).

It's less likely to be useful here, but for info I went trhough barrier usage in the UK to try and map some of the more esoteric values to more common ones - see https://github.com/SomeoneElseOSM/SomeoneElse-style/blob/master/style.lua for that.

* SVG is not an issue for me personally because I never print anything out; I'm aware that this is a significant use case for some people though.

math1985:
@kocio-pl @nebulon42 Do you have any suggestions for an icon for barrier=stile and barrier=kissing_gate?

[...]

PR #3450, opened 2018-10-15
Add icons for barrier=cattle_grid, barrier=stile

Adamant36:
This PR adds icons for barrier=cattle_grid and barrier=stile. The PR is a work in progress for issue #846 but doesn't close it. Its just easier for me to do a few at a time and these were already decided on. I also moved the other barrier icons into their own folder, along with these ones. 
Cattle grid z19
cattle grid z19
stile z19
stile z19

Other barriers that had icons moved still working
gate 
gate
lift gate
lift gate
toll booth
toll booth

@kocio-pl, if you can get this merged in the next few days before the new version roles out, I would really appreciate it. Then I can get to work on the other ones for the next update.


kocio-pl:
Sorry, it somehow got out of my sight before release, but this all looks good, so you can move on now with other barriers.

kocio-pl:
I had a problem with finding stile barriers. It would be easier for me if all the rendering images were like 2x smaller to both fit into the column without rescaling and show only local context, which should be enough.

PR #3462, opened 2018-10-20
Add icon for amenity=public_bookcase

### LINKED ISSUE #1288: Please render amenity=public_bookcase
http://wiki.openstreetmap.org/wiki/Tag:amenity%3Dpublic_bookcase

It would be nice to have a special icon for public bookcases, maybe a hand combined with a book to symbolize book sharing.

Adamant36:
Closes #1288 

This PR adds rendering for amenity=public_bookcase. Its a pretty well used tag already and the usage seems to be increasing. So why not? I left rendering to start at z19 because otherwise ones with names risk covering up addresses or other more important information. Also, its not suppose to be mapped on ways and there's only like 100 out of 2000ish mapped that way. So, I didn't do a test on them. 
taghistory
https://www.openstreetmap.org/#map=19/37.76514/-122.45442
public bookcase z19 1
https://www.openstreetmap.org/#map=19/37.75058/-122.41776
public bookcase z19 2


amenity-points.mss
```diff
@@ -90,7 +90,14 @@
     marker-placement: interior;
     marker-clip: false;
   }
-
+
```
kocio-pl: Introduced some unwanted spaces, please remove them.

sommerluk:
Looks rather like a library for me…

SomeoneElseOSM:
@sommerluk 

> Looks rather like a library for me

In a different style I couldn't think of a good differentiating icon so went with a small "L" (which wouldn't be appropriate here).  If you can think of a better idea I'd be all ears :)

Just to add another idea, maybe change "library" to incorporate the idea of "building" (like [here](https://map.atownsend.org.uk/maps/map/map.html#zoom=20&lat=-25.002872&lon=135.182616) for a different icon for a different thing, just as an example)

Adamant36:
@kocio-pl, my bad. I fixed it.

polarbearing:
I'd prefer @somerluk's idea, the shelf by @Tomasz-W is not recognisable to me.

Tomasz-W:
* amenity public_bookcase2 (v4)
* amenity public_bookcase3 (v5)

Gist link: https://gist.github.com/Tomasz-W/164f3f591a59e33d6d6663e2b0564033

polarbearing:
Sorry @Tomasz-W but my best interpretation is a book being raked. It does not work for me as the book is much larger than the self spaces.

kocio-pl:
For me even the same icon as library would be OK, because it's similar idea, but just much smaller size, so it's enough to see that it's rendered later.

meased:
Really quick idea combining the library and shelter icons:

bookcase

HolgerJeromin:
perhaps this is an candidate for a black color, too. But I do not remember the logic for black vs brown

polarbearing:
The open book with roof is good. We should not have too many different symbols based on a book .

Adamant36:
I prefer the open book with roof also. Can we go with it?
public_bookcase

kocio-pl:
I think this is good solution.

Adamant36:
@kocio-pl, its been updated.

project.mml
```diff
@@ -1585,7 +1585,8 @@ Layer:
                                                   'fuel', 'hospital', 'ice_cream', 'embassy', 'library', 'courthouse', 'townhall', 'parking',
                                                   'bicycle_parking', 'motorcycle_parking', 'pharmacy', 'doctors', 'dentist', 'place_of_worship',
                                                   'police', 'post_box', 'post_office', 'pub', 'biergarten', 'recycling', 'restaurant', 'food_court',
-                                                  'fast_food', 'telephone', 'taxi', 'theatre', 'toilets', 'drinking_water', 'internet_cafe', 
+                                                  'fast_food', 'telephone', 'taxi', 'theatre', 'toilets', 'drinking_water', 'public_bookcase',
```
kocio-pl: Looks like this line should be removed.

Adamant36: I removed it. Thanks for spotting it.

kocio-pl:
Thanks! It works as expected for both named nodes and named areas.

Issue #2629, opened 2017-05-10
waterway=ditch in a forest, colored over the forest


AllroadsNL:
In a forest, when drawn in the ditches.
On level 13 you already see the ditches in the forest.
This give a blue look, not seeing the forest anymore.

Example:
Here is a mountain bike route, through several ditches.
The reason to draw in the ditches.

Ditches are 50 cm deep. 1- 1,30 wide, most of the time dry and flows in to the 70-100 deep ditch.

For me showing at level 17 would be fine.
Two plots where drawn in yet, but the whole forest have ditches.

https://www.openstreetmap.org/#map=13/52.6587/5.9629

level 13
![ditch 13](https://cloud.githubusercontent.com/assets/8817345/25886365/9697cb62-355d-11e7-9a85-a1e94e2acb8f.PNG)

level 14
![ditch 14](https://cloud.githubusercontent.com/assets/8817345/25886376/9eec5fe4-355d-11e7-9add-e2a005adb443.PNG)

level 15
![ditch 15](https://cloud.githubusercontent.com/assets/8817345/25886395/ac257114-355d-11e7-82ff-762d4d1ba834.PNG)
 
level 16
![ditch 16](https://cloud.githubusercontent.com/assets/8817345/25886408/b5aca11c-355d-11e7-8c51-f2d59e1476e9.PNG)

level 17
![ditch 17](https://cloud.githubusercontent.com/assets/8817345/25886418/be22a4d6-355d-11e7-9775-10f13e2140d7.PNG)

imagico:
So the issue is about `waterway=ditch` being rendered too early at z13 in your opinion?

This is more or less the same as #1101 suggests for `waterway=drain`.  The minor waterways at the lowest zoom level (z13) could be slightly reduced in width and casing strength probably though there were increased in width for a reason (#967).  Note we do not want to start ditches later than streams due to #2346.

If these are dry most of the time you should tag them `intermittent=yes` by the way - no matter if that looks good on the map or not at the moment.

AllroadsNL:
Yes, rendered too earlier on z13.
I would say, z16 z17

Forget, to search for a earlier issue.!

Intermittent=yes, I know,
Now intermittent, the right side of the example, compare.

z13
![int ditch 13](https://cloud.githubusercontent.com/assets/8817345/25899579/a9be8ebe-3590-11e7-88f6-a59666d2cc2a.PNG)

 z14
[...]

PR #3467, opened 2018-10-21
Clean up z13

### LINKED ISSUE #3331: Further clean-up mid zoom levels
Some ideas to further clean up the mid zoom levels:

Current:

screen shot 2018-08-05 at 23 11 50

Dropping buildings from z13:
screen shot 2018-08-05 at 23 11 28

Dropping footways from z13:
screen shot 2018-08-05 at 23 18 02

Dropping buildings and footways from z13:
screen shot 2018-08-05 at 23 15 10

matthijsmelissen:
This pull request cleans up the rendering on z13 and z14, by removing a number of features from these zoom levels.

* Drop buildings from z13 ~~and z14~~
* Drop footways from z13
* Drop service roads from z13
* Drop pedestrian roads from z13
* Different rendering of living_street on z13
* Drop playground outlines from z13 and z14
* Lighten up the color of trams
* Drop streams, ditches and drains from z13 (and intermittent ones from z14)

Fixes #3331.

The primary purpose of this pull request is to remove visual clutter from z13 and z14. These zoom levels contain lots of tiny elements that are hard to interpret, and lead to noise in the images. As a consequence of removing these features, the remaining features (such as the street pattern) become easier to read.

The secondary purpose is to give more attention to landuse rendering on these zoom levels. Especially landuse=retail is currently hardly visible on the map on any zoom level, this pull request should improve this.

### DIFF
--- a/buildings.mss
+++ b/buildings.mss
@@ -9,12 +9,16 @@
 @entrance-normal: @building-line;
 
 #buildings {
-  [zoom >= 15] {
-    polygon-fill: @building-fill;
+  [zoom >= 14] {
+    polygon-fill: @building-low-zoom;
     polygon-clip: false;
-    line-color: @building-line;
-    line-width: .75;
-    line-clip: false;
+    line-width: 0;
+    [zoom >= 15] {
+      polygon-fill: @building-fill;
+      line-color: @building-line;
+      line-width: .75;
+      line-clip: false;
+    }
     [amenity = 'place_of_worship'],
     [aeroway = 'terminal'],
     [aerialway = 'station'],
--- a/buildings.mss
+++ b/buildings.mss
@@ -9,16 +9,12 @@
 @entrance-normal: @building-line;
 
 #buildings {
-  [zoom >= 13] {
-    polygon-fill: @building-low-zoom;
+  [zoom >= 15] {
+    polygon-fill: @building-fill;
     polygon-clip: false;
-    line-width: 0;
-    [zoom >= 15] {
-      polygon-fill: @building-fill;
-      line-color: @building-line;
-      line-width: .75;
-      line-clip: false;
-    }
+    line-color: @building-line;
+    line-width: .75;
+    line-clip: false;
     [amenity = 'place_of_worship'],
     [aeroway = 'terminal'],
     [aerialway = 'station'],
--- a/landcover.mss
+++ b/landcover.mss
@@ -108,8 +108,10 @@
   [feature = 'leisure_playground'][zoom >= 13],
   [feature = 'leisure_fitness_station'][zoom >= 13] {
     polygon-fill: @leisure;
-    line-color: darken(@leisure, 60%);
-    line-width: 0.3;
+    [zoom >= 15] {
+      line-color: darken(@leisure, 60%);
+      line-width: 0.3;
+    }
     [way_pixels >= 4]  { polygon-gamma: 0.75; }
     [way_pixels >= 64] { polygon-gamma: 0.3;  }
   }
--- a/project.mml
+++ b/project.mml
@@ -437,7 +437,7 @@ Layer:
           ORDER BY COALESCE(layer,0), way_area DESC
         ) AS buildings
     properties:
-      minzoom: 13
+      minzoom: 14
   - id: tunnels
     class: access
     geometry: linestring
@@ -2063,7 +2063,7 @@ Layer:
                                                   'school', 'college', 'kindergarten', 'hospital', 'ice_cream', 'pharmacy', 'doctors', 'dentist',
                                                   'atm', 'bicycle_rental', 'bureau_de_change', 'car_rental', 'car_wash', 'post_box', 'post_office',
                                                   'recycling', 'telephone', 'toilets', 'taxi', 'drinking_water', 'hunting_stand', 'internet_cafe', 'public_bookcase',
-                                                  'nightclub', 'veterinary', 'social_facility', 'vehicle_inspection', 'charging_station', 'arts_centre', 
+                                                  'nightclub', 'veterinary', 'social_facility', 'vehicle_inspection', 'charging_station', 'arts_centre',
                                                   'ferry_terminal', 'marketplace', 'shower', 'bbq', 'nursing_home', 'childcare', 'driving_school',
                                                   'casino', 'boat_rental', 'bicycle_repair_station') THEN amenity ELSE NULL END,
               'advertising_' || CASE WHEN tags->'advertising' in ('column') THEN tags->'advertising' else NULL END,
--- a/roads.mss
+++ b/roads.mss
@@ -109,11 +109,9 @@
 @secondary-link-width-z13:        4;
 @tertiary-width-z13:              4;
 @residential-width-z13:           2.5;
-@service-width-z13:               0.8;
 @living-street-width-z13:         2;
-@pedestrian-width-z13:            2;
 @bridleway-width-z13:             0.3;
-@footway-width-z13:               0.7;
+@footway-width-z14:               0.7;
 @cycleway-width-z13:              0.7;
 @track-width-z13:                 0.5;
 @track-grade1-width-z13:          0.5;
@@ -556,10 +554,9 @@
     }
 
     [feature = 'highway_pedestrian'] {
-      [zoom >= 13] {
+      [zoom >= 14] {
         line-color: @pedestrian-casing;
-        line-width: @pedestrian-width-z13;
-        [zoom >= 14] { line-width: @pedestrian-width-z14; }
+        line-width: @pedestrian-width-z14;
         [zoom >= 15] { line-width: @pedestrian-width-z15; }
         [zoom >= 16] { line-width: @pedestrian-width-z16; }
         [zoom >= 17] { line-width: @pedestrian-width-z17; }
@@ -573,19 +570,20 @@
           line-dasharray: 4,2;
         }
         #bridges {
-          [zoom >= 14] {
-            line-color: @bridge-casing;
-            line-join: round;
-          }
+          line-color: @bridge-casing;
+          line-join: round;
         }
       }
     }
 
     [feature = 'highway_living_street'] {
       [zoom >= 13] {
-        line-color: @living-street-casing;
+        line-color: @residential-casing;
         line-width: @living-street-width-z13;
-        [zoom >= 14] { line-width: @living-street-width-z14; }
+        [zoom >= 14] {
+          line-color: @living-street-casing;
+          line-width: @living-street-width-z14;
+        }
         [zoom >= 15] { line-width: @living-street-width-z15; }
         [zoom >= 16] { line-width: @living-street-width-z16; }
         [zoom >= 17] { line-width: @living-street-width-z17; }
@@ -655,7 +653,7 @@
       #bridges {
         [zoom >= 14][access != 'no'],
         [zoom >= 15] {
-          line-width: @footway-width-z13 + 2 * (@paths-background-width + @paths-bridge-casing-width);
+          line-width: @footway-width-z14 + 2 * (@paths-background-width + @paths-bridge-casing-width);
           [zoom >= 15] { line-width: @footway-width-z15 + 2 * (@paths-background-width + @paths-bridge-casing-width); }
           [zoom >= 16] { line-width: @footway-width-z16 + 2 * (@paths-background-width + @paths-bridge-casing-width); }
           [zoom >= 18] { line-width: @footway-width-z18 + 2 * (@paths-background-width + @paths-bridge-casing-width); }
@@ -667,7 +665,7 @@
       #tunnels {
         [zoom >= 14][access != 'no'],
         [zoom >= 15] {
-          line-width: @footway-width-z13 + 2 * (@paths-background-width + @paths-tunnel-casing-width);
+          line-width: @footway-width-z14 + 2 * (@paths-background-width + @paths-tunnel-casing-width);
           [zoom >= 15] { line-width: @footway-width-z15 + 2 * (@paths-background-width + @paths-tunnel-casing-width); }
           [zoom >= 16] { line-width: @footway-width-z16 + 2 * (@paths-background-width + @paths-tunnel-casing-width); }
           [zoom >= 18] { line-width: @footway-width-z18 + 2 * (@paths-background-width + @paths-tunnel-casing-width); }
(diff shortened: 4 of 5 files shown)


matthijsmelissen:
Preview z13:
Before:
screen shot 2018-10-21 at 12 51 03
screen shot 2018-10-21 at 12 51 55






Preview z14:

Before:
screen shot 2018-10-21 at 12 50 15

After:
screen shot 2018-10-21 at 12 49 28

jeisenbe:
How about moving intermittent and seasonal streams down one zoom level but
leaving the regular streams as-is?

At any rate I think intermittent streams should be rendered one zoom level
later than year-round streams.

dieterdreist:
sent from a phone

> On 21. Oct 2018, at 12:52, Matthijs Melissen <notifications@github.com> wrote:
> 
> Preview z13:
> Before:
> 
> 


missing buildings are considerably worse than having them at z13. Look for instance at the bigger buildings, how the situation becomes misrepresented by appearing void without them. 

Even without being able to discern every single small building, the sum of these create the morphology of the city (together with the roads), and this scale is interesting for seeing buildings and bigger urban structures and patterns.

In a bigger scheme, osm carto as a raster map is in the position of being able to show the buildings earlier than a vector map can (there is almost no cost in terms of tile size and no cost in terms of client rendering time, compared to vector tiles), and my stance is we should use this advantage, as an alternative for people interested in urban structures.

mboeringa:
Honestly, no one is going to understand this change... Especially dropping buildings from z13/14 will be totally alien to our openstreetmap community and heavily contested if implemented.

kocio-pl:
I will try to reply to the critics later (also look closer at all the proposed changes individually), for now just Warsaw example:

z13

Before
hkivvtea

After
tkskupza


z14

Before
mpgc8zxx

After
dvc6lzkr

kocio-pl:
It's good to look back to see how the database grows over the years creating more noise. Using current version of style (without this PR) for the last CC BY-SA licensed planet file (2012) and comparing with current data in Warsaw:

z13, 2012

tafxyayv

z13, 2018

hkivvtea

z14, 2012

b pm7f0j

z14, 2018

h2hvqj2d

mboeringa:
@kocio-pl ,

This adds nothing new to the existing images. What it actually mainly shows IMO is the less than ideal choice of a highly saturated **highway=footway** color, which is the main cause of the clutter, and the addition of many sidewalks. 

Unfortunately, although your Polish fellow mappers have been very active adding such sidewalk detail, they have been far less diligent in adding additional tags like footway=sidewalk/crossing, that might help filter out this clutter at medium zooms like Z13/14.

kocio-pl:
> This seems to go against the first item at https://github.com/gravitystorm/openstreetmap-carto/blob/master/CARTOGRAPHY.md "an important feedback mechanism for mappers". 

I think it doesn't. The key is that only some buildings are visible on midzoom, and they are related to some special area, which are eclipsed by them. z13 is the first zoom level where people can have proper feedback about landuses that are medium-level structures. It also supports diversity and helps to see where they are omitted.

This is related to #2896 - there's no one magical solution to show bigger entities/structures without eclipsing them with smaller features, since it would require considering very special relations (like "this park is part of the hospital, so render it one step later"), which is unlikely. However showing all the possible landuses without buildings on one zoom level at least gives the general impression what bigger areas are there before we go into details and loose this general view.

Having said that, I think that one special zoom level for landuses is enough and I would not remove buildings from z14.

matthijsmelissen:
> How about moving intermittent and seasonal streams down one zoom level but
leaving the regular streams as-is?
> At any rate I think intermittent streams should be rendered one zoom level
later than year-round streams.

Good idea. I have removed intermittent streams/ditches/drains from z14 too now.

> This seems to go against the first item at https://github.com/gravitystorm/openstreetmap-carto/blob/master/CARTOGRAPHY.md "an important feedback mechanism for mappers".

No, the opposite: the feedback mechanism for retail landuse is currently almost absent, this PR will improve this. Seeing the buildings one level later make not much difference in terms of feedback for building mappers.

> Also providing examples only of a densly-populated Central European city will do nothing to placate critics who've accused this style as being targeted at those areas only in the past ...

I have checked other areas as well, but I'm not planning on making screenshots of every arbitrary place on earth. Anyone who is interested in how this change behaves in other areas can run the code for themselves (and I'm happy to hear about potential issues in other areas).

> Having said that, I think that one special zoom level for landuses is enough and I would not remove buildings from z14.

I have adapted this pull request to only drop buildings from z13 for now.

matthijsmelissen:
@kocio-pl As far as I concerned this is ready to be merged now, what do you think?

kocio-pl:
Some more examples of cleaning the noise on z13 and giving better feedback on landuses (thanks to @komzpa I was able to import the whole Europe to check on his server):

London

screenshot_2018-10-28 openstreetmap carto kosmtik
screenshot_2018-10-28 openstreetmap carto kosmtik 1

Moscow

screenshot_2018-10-28 openstreetmap carto kosmtik 2
screenshot_2018-10-28 openstreetmap carto kosmtik 3

Stockholm

screenshot_2018-10-28 openstreetmap carto kosmtik 4
screenshot_2018-10-28 openstreetmap carto kosmtik 5

Madrid

screenshot_2018-10-28 openstreetmap carto kosmtik 6
screenshot_2018-10-28 openstreetmap carto kosmtik 7

Barcelona

screenshot_2018-10-28 openstreetmap carto kosmtik 8
screenshot_2018-10-28 openstreetmap carto kosmtik 9

Athens

screenshot_2018-10-28 openstreetmap carto kosmtik 10
screenshot_2018-10-28 openstreetmap carto kosmtik 11

jeisenbe:
Stockholm and Barcelona look odd in their medieval urban centers. Is this
because the streets are pedestrianized? The lack of landuse is also
apparent, but I understand that this should be added.

If we are going by a size rule, aren’t pedestrian streets equal in size to
residential streets?

kocio-pl:
Resolves part of #2629.

matthijsmelissen:
> If we are going by a size rule

Please note that this is not an official rule of the style, or a rule agreed on by the maintainers.

matthijsmelissen:
> Stockholm and Barcelona look odd in their medieval urban centers. Is this because the streets are pedestrianized? 

Yes, I think so. Think of it this way: if an car driver is driving from north to south Stockholm, he should know that he should navigate around the city centre.

vholten:
This PR changes the tram line color to #8E8E8E for zooms 13 and up, while it remains #6E6E6E at z12. Is it intentional that tram lines become lighter when zooming in from z12 to z13?

SomeoneElseOSM:
> Interesting take! I haven't thought about it this way, but it makes a lot of sense for me - medium scale is more about transportation than pedestrian-related objects.

That's definitely something that very much depends on the user.  Speaking personally, z13 is what I'd normally use for pedestrian route planning.  Obviously (due to changes over a number of years, actually) the style is a bit rubbish for that already (compare for example https://map.atownsend.org.uk/maps/map/map.html#zoom=13&lat=54.0967&lon=-1.0322 , which considers pedestrian horse/cycle routes important to show, and https://www.openstreetmap.org/#map=13/54.0971/-1.0530 in the standard style, which seems to give visual priority to streams and ditches, and the national park and admin boundaries.

mboeringa:
> @kocio-pl As far as I concerned this is ready to be merged now, what do you think?

I am almost certain this is going to lead to a barrage of complaints of our user base who are not involved in style development, and won't understand this change.

So unless you are "ready" to deal with that, and possibly have to revert this change ASAP if the masses arrrive here on the Github repository or in the forums to complain, I don't think you are ready yet to implement this.

matkoniecz:
> possibly have to revert this change ASAP if the masses arrrive here on the Github repository or in the forums to complain

Complaining alone is not sufficient for revert.

mboeringa:
> Complaining alone is not sufficient for revert.

Maybe not..., but loosing the goodwill of your user base, should be a big concern for any volunteer project like OpenStreetMap.

kocio-pl:
> I am almost certain this is going to lead to a barrage of complaints of our user base who are not involved in style development, and won't understand this change.

I don't really know - maybe you're right and it will. I am constantly taken by surprise when small details make people nervous or active in some way, wile at the same time some risky changes I'm afraid of, are met with total lack of response. Anyway it's just my opinion and I trust in Matthijs deciding what to do with this code.

matthijsmelissen:
> This PR changes the tram line color to #8E8E8E for zooms 13 and up, while it remains #6E6E6E at z12. Is it intentional that tram lines become lighter when zooming in from z12 to z13?

Good point. I have reduced the lightening now by making it #6E6E6E for all zoom levels now.

kocio-pl:
Thanks for the cleaning! 

Interesting evolution of clarity in London (starting from the old XML-based style):

screenshot_2018-11-05 your own openstreetmap rendering stack 1
screenshot_2018-11-05 openstreetmap carto kosmtik
screenshot_2018-11-05 openstreetmap carto kosmtik 1

Tomasz-W:
@matthijsmelissen As it was hard to imagine for me how this change would look like outside of presented test renderings, I wasn't active in discussion above, but looking at my hometown now with this change merged , I can tell that it's a very very good improvement making map a much nicer :)

Tomasz-W:
@matthijsmelissen What do you think about current state of z8-10? In my opinion they are quite messy. Most problematic things there are:
* too thin country borders on z8-9
* too prominent protected areas borders
* too prominent railways

These things (and beybe some another ones also) mixed together make an unreadable colorful 'spagetti' in most places.

matkoniecz:
@Tomasz-W 

That would be much better as new issues. New issues are cheap, do not be afraid to open them (it is certainly preferable over comments in slightly related issues).

If something turns out to be a poor idea, then in the worst case such issue will be simply closed - what is a nothing tragic.

dktue:
Oh noooooooooo!

Please bring back the buildings!

Look how strange the yellow and purple areas look without them (clinics, schools, other big buildings).

----
**Z13: Akward "empty" areas**
image

**Z14: Ahhhhhhh... now I understand the map**
image

kocio-pl:
This was done month ago. They look great to me, this zoom level is clear now, especially after moving tram and subway stations to z14+ in some bigger cities in the latest release. The visual noise is smaller.

dktue:
> This was done month ago. They look great to me, this zoom level is clear now, especially after moving tram and subway stations to z14+ in some bigger cities in the latest release. The visual noise is smaller.

I disagree and share @dieterdreist 's optinion:

> missing buildings are considerably worse than having them at z13. Look for instance at the bigger buildings, how the situation becomes misrepresented by appearing void without them.
> Even without being able to discern every single small building, the sum of these create the morphology of the city (together with the roads), and this scale is interesting for seeing buildings and bigger urban structures and patterns.

My opinion is the one of a confused mapper that though that there was a bug. Please treat this seriously.

Adamant36:
> Look how strange the yellow and purple areas look without them
Maps are always an abstraction. Considering the building were barely discernible at that level anyway, I don't know what the problem with being able to see the landuse areas only are. Especially since the buildings often got in the way of seeing it. 

There's nothing strange about someone going "oh look a retail area" and then zooming in one more level see the buildings if they want. Its already done at z12 to z13 to see specific landuse. Or is that strange also?

Hopefully this help people micro map landuse more now that it has its own zoom level at least. Instead of mapping whole towns as residential like they normally do.

kocio-pl:
> My opinion is the one of a confused mapper that though that there was a bug. Please treat this seriously.

I do really treat it very seriously, believe me. It's always bugging me when someone has a problem with this style, since I'd like it to be good for all the people if possible. But it's not possible and I just have different view on z13 buildings problem than you.

For me this is natural progression of the urban structure:
- On z12 we see the difference between green areas (natural and semi natural) and the rest of human landuses as uniform grey. And on this scale this is enough for me.
- Next, on z13 we see the difference between landuses - for example lack of buildings gives me general view how big industrial landuses are compared to residential areas. 
- Then on z14 buildings add more details to the picture, however we loose a bit of clarity (small details, especially dark ones, add high frequency noise), but this is next step of zoom and it makes sense to make it more detailed than z13.

Using intermediate step between what is shown on z12 and z14 is a nice idea for me. Smooth changes of details level is what I like.

Komzpa:
Hi @kocio-pl,

Looking at provided screenshot I tend to read what was clear "University" previously as "A large field or maybe parking". Ways of mitigating this that I see:

 - modify university color on the zoom by blending it towards building color. It works well on industrial areas that are clearly industrial even though buildings are not there.

 - show building that are as large as an average map icon, to represent themselves "as an icon". That approach was taken on the Belroad map style you can have a look at http://latlon.org/#13/53.8428/27.6855?layers=L

image

Komzpa:
Relevant code in Belroad:
https://github.com/jekhor/belroad/blob/master/project.mml#L273
(`where ... and way_area > 3*3*!pixel_height!*!pixel_width!`)

I can blindly adapt openstreetmap-carto to match it, I don't have any map rendering setup at hand now to test.

kocio-pl:
You could use this to test:

dktue:
@kocio-pl : I do agree that cleaning z13 is a good idea. But I do think that simply leaving out the buildings was not a good thing to do. @Komzpa 's idea looks like a good approach to clean z13 but still maintaining the relevant structures.

**My vote is to undo this change until an improved version is found for buildings.**

Adamant36:
Maybe it could be improved once the major/minor building thing is implimented or a similar solution is found like rendering based on building type. 

IMHO though a regression based on the negitive opinions of a few people that dont like how its rendered wouldnt be a good idea. As it not an actual bug/problem. So it should be treated like every other rendering improvement where a solution is found instead. Plenty of things that could look better in a few instances wait for solutions. Theres no reason this issue shouldnt be any different. Its not really that drastically bad. Plus, it did sit there for like a month without serious critism and recived the thumbs up from multiple people before it was merged. So...

dktue:
@Adamant36 : I completely agree with you that there are many other issues but I definitely think this should be treated differently.

Not having any opposing votes in this repository is not an indicator that people mostly think that this is a good change. I didn't read about this in here even if I do contribute here from time to time.

I came here because I was _confused_. And I do think that many other people are, too but don't know where to address their confusion.

It's easy to get in a trap when you're the developer and then live in a bubble where everything makes perfect sense. And you can, indeed, explain why this is a good idea. But almost nobody knows about those ideas behind it and is left behind alone with it's confusion.

We should definitively revert that change and re-release it once it's done right.

jragusa:
Problems such those underlined by @Komzpa happen with large areas without a great density of road or large buildings. Inversely, if you consider dense area with important road network such as some capitals ([Paris](https://www.openstreetmap.org/#map=13/48.8606/2.3433), [London](https://www.openstreetmap.org/#map=13/51.5084/-0.1013) or [Berlin](https://www.openstreetmap.org/#map=13/52.5019/13.3733), I have also tried with [Geneva](https://www.openstreetmap.org/#map=13/46.2021/6.1450)), the removal of buildings is beneficial and you can even better identify some specific areas (`amenity=school`, `landuse=retail`...).

matkoniecz:
In general it is better to open a new issue rather than start commenting in old.

This discussion may easily end forgotten unlike new issue that would need to be explicitly closed.

---

> My vote is to

Note that we are not voting. Single comment with a new argument should have weight greater than 100 +1 comments.

jeisenbe:
It sounds like one new issue might be that the new societal_amenities
rendering (light yellow for schools and hospitals) is too light. Is that
correct?

@Adamant36 recently mentioned that he would have preferred more time to
perfect this color. A slightly darker and more saturated shade of yellow
might help.

PR #3469, opened 2018-10-21
Move danger_area into landuse-overlay

### LINKED ISSUE #3454: Scrub overlaps danger area
### Expected behavior

The danger area[1] should be above the beach and scrub[2].

### Actual behavior

Scrub and beach are above the danger area.

map
[0]: https://www.openstreetmap.org/#map=18/54.15019/13.82067
[1]: https://www.openstreetmap.org/way/307904554/
[2]: https://www.openstreetmap.org/way/578224669

jragusa:
Fixes #3454

Render `military=danger_area` above other landuses by moving it into the `landuse-overlay` layer

https://www.openstreetmap.org/way/307904554
Before
map

After
danger_area

### DIFF
--- a/project.mml
+++ b/project.mml
@@ -87,17 +87,15 @@ Layer:
       table: |-
         (SELECT
             way, name, way_pixels, religion,
-            COALESCE(wetland, landuse, military, "natural") AS feature
+            COALESCE(wetland, landuse, "natural") AS feature
           FROM (SELECT
               way, COALESCE(name, '') AS name, religion,
-              ('landuse_' || (CASE WHEN landuse IN ('forest', 'military', 'farmland', 'residential', 'commercial', 'retail', 'industrial', 'meadow') THEN landuse ELSE NULL END)) AS landuse,
-              ('military_' || (CASE WHEN military IN ('danger_area') THEN military ELSE NULL END)) AS military,
+              ('landuse_' || (CASE WHEN landuse IN ('forest', 'farmland', 'residential', 'commercial', 'retail', 'industrial', 'meadow') THEN landuse ELSE NULL END)) AS landuse,
               ('natural_' || (CASE WHEN "natural" IN ('wood', 'sand', 'scree', 'shingle', 'bare_rock', 'heath', 'grassland') THEN "natural" ELSE NULL END)) AS "natural",
               ('wetland_' || (CASE WHEN "natural" IN ('wetland', 'mud') THEN (CASE WHEN "natural" IN ('mud') THEN "natural" ELSE tags->'wetland' END) ELSE NULL END)) AS wetland,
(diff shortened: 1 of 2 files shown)


kocio-pl:
@matkoniecz @matthijsmelissen Could you check this code? I'm not sure if it does not change anything else beside moving to overlay.

matkoniecz:
One thing that is certainly worth testing - is labelling of such areas still working correctly? I see nothing obviously problematic but it should be checked.

kocio-pl:
Could you make this additional check?

jragusa:
it's working from my side: 
https://www.openstreetmap.org/way/424081505 (without `landuse=military`)
danger_area_1

project.mml
```diff
@@ -87,15 +87,17 @@ Layer:
       table: |-
         (SELECT
             way, name, way_pixels, religion,
-            COALESCE(wetland, landuse, "natural") AS feature
+            COALESCE(wetland, landuse, military, "natural") AS feature
```
matthijsmelissen: I think we don't need military=danger_area in this layer. The layer `landuse-overlay` also works for lowzoom rendering.

We can even remove the landuse=military from this layer: the style for #landcover-low-zoom does not seem to use landuse=military.

project.mml
```diff
@@ -538,9 +538,11 @@ Layer:
         (SELECT
             way,
             landuse,
-            way_area/NULLIF(!pixel_width!::real*!pixel_height!::real,0) AS way_pixels
+            way_area/NULLIF(!pixel_width!::real*!pixel_height!::real,0) AS way_pixels,
+            military
```
matthijsmelissen: Maybe swap these lines for clarity.

project.mml
jragusa: fixed

project.mml
jragusa: fixed

matthijsmelissen:
Thanks!

matthijsmelissen:
Thanks!

project.mml
```diff
@@ -538,9 +536,11 @@ Layer:
         (SELECT
             way,
             landuse,
+            military,
             way_area/NULLIF(!pixel_width!::real*!pixel_height!::real,0) AS way_pixels
           FROM planet_osm_polygon
           WHERE landuse = 'military'
+            OR military = 'danger_area'
             AND building IS NULL
```
matthijsmelissen: In SQL, AND has precedence over OR, so this will be parsed as `landuse = 'military' OR (military = 'danger_area' AND building IS NULL)`. The intended meaning is of course `(landuse = 'military' OR military = 'danger_area') AND building IS NULL` so we are missing a pair of brackets here.

PR #3473, opened 2018-10-22
Render man_made=telescope telescope:type=radio

### LINKED ISSUE #3389: Render man_made=telescope
Render **man_made=telescope + tower:construction=dish** with
dish
It should be rendered as well as man_made=tower + tower:construction=dish.
Sometimes the dish is not on a tower and is called a telescope (telescope:type=radio). It should then also get an icon.

jeisenbe:
Related to #3389

_Changes proposed in this pull request:_
- Add rendering of radio telescopes:
- Render man_made=telescope when telescope:type=radio with a radio telescope dish icon in man_made-gray color. 

_Test rendering with links to the example places:_

**Mauna Kea, Hawaii at z17; nodes** [https://www.openstreetmap.org/#map=18/19.82366/-155.47612](https://www.openstreetmap.org/#map=17/19.82366/-155.47612)

_Before_ 
hawai-telescopes-before-z17

_After_ 

**LOFAR Radio telescopes in Germany, closed ways, z17**
https://www.openstreetmap.org/way/132557743
https://www.openstreetmap.org/#map=17/50.8976057/6.4233260
_Before_ 
lofar-before

_After_

**Radio telescopes in Cambridge, England; closed ways, z17:**
https://www.openstreetmap.org/way/32733809
https://www.openstreetmap.org/#map=17/52.1645763/0.0309299

_Before_ 
mrao-before

_After_

Questions: 
1) Should initial zoom level be constant, or should it depend on telescope:diameter and waypixels?
2) Should optical telescopes be included in this PR?


jeisenbe:
I don't understand why GitHub thinks that the amenity=internet_cafe commits are not already merged?

kocio-pl:
GitHub has some problems from yesterday:

https://status.github.com/messages

but it shows the merge properly:

https://github.com/gravitystorm/openstreetmap-carto/blob/master/symbols/amenity/internet_cafe.svg

kocio-pl:
1. The usage is smaller than typical, but these objects are not so popular, so it's OK for me. 
2. Could you check if all of them are located in the outdoor? If yes, we could render them much earlier.
3. If this is the only type of telescope that we will show, it will really close #3389, otherwise it's better to change the words into something neutral like "Related to" (some other than GitHub key words).

taghistory 16

jeisenbe:
Re 2.  "Could you check if all of them are located in the outdoor? If yes, we could render them much earlier."

By "the outdoor", do you mean in rural areas, rather than in towns or cities? Yes, all of the properly tagged examples that I found were in rural areas, even in very urban countries like England and Germany. I did find a few mistagged satellite communication dishes in towns, probably because the approved tag, tower:structure=dish, is not very intuitive.

Were you thinking z16 or z15 to start? I was also considering having the zoom level based on the telescope:diameter (like how towers render based on height) for nodes, and on waypixels for closed ways. Should I show some test examples?

Re: 3., I have changed it to "related to 3389" for now, but I actually would like to render optical telescopes, instead of rendering only one of the two major types. It will also make the code in project.mml file a little simpler if we add telescope:type=optical. 

I've asked @Tomasz-W if he can make a telescope dome icon for optical telescopes

kocio-pl:
> By "the outdoor", do you mean in rural areas, rather than in towns or cities? Yes, all of the properly tagged examples that I found were in rural areas, even in very urban countries like England and Germany. 

Yes, I mean that. Examples are always nice to see. It can be any zoom level (including z15 and z16) as long as it does not eclipse bigger entities like county or nearest city name label. Testing is the key and mistagged objects should be just repaired.

> I've asked @Tomasz-W if he can make a telescope dome icon for optical telescopes

I guess optical telescopes will need another PR, so this one won't resolve the issue anyway.

Tomasz-W:
@jeisenbe I'll try in the next couple of days.

kocio-pl:
This one is the biggest radio telescope, AFAIK:

https://en.wikipedia.org/wiki/Five_hundred_meter_Aperture_Spherical_Telescope
https://www.openstreetmap.org/way/384699313

jeisenbe:
The most common size for telescope:diameter are around 25 or 30 meters for radio telescopes, and those that are mapped as closed ways can be even bigger. Only a few had height marked, and the dish-shaped ones are often almost twice as tall as their diameter, eg 54 meters height with 30 meter diameter. So I would consider height to be about 2x diameter on average.

Towers are rendered at z13 for >100 meters, at z14 for >50 meters, and z15 for >30 meters. So I've tried z13 for telescope:diameter >= 60m, z14 for diameter >=30 meters, z15 for diameter >=15, and z16 otherwise.

I checked a few more countries: Canada, Russia, Australia, Italy, Germany and England. There were no radio telescopes within urban areas, except these in Bonn, which I believe are mistagged satellite uplink dishes (Deutsche Welle is a public broadcaster)
"description=Satellitenschüsseln Deutsche Welle (Uplink?), operator=Deutsche Welle"
https://www.openstreetmap.org/#map=18/50.7181635/7.1262950
bonn-uplinks

These are on the edge of Berlin, but it's surrounded by parks and forest. [EDIT: also they are in "Erdfunkstelle Telcom", so may be satellite uplink dishes, not radio telescopes] https://www.openstreetmap.org/#map=17/52.4083843/13.1252944
berlin-radio

Rendering at z15 (not visible, because diameter is null):
berlin-radio-z15

z16: 
berlin-radio-z16

This one is on the outskirts of Moscow, between allotments, industrial and forest. It is 64 meters in diameter and quite tall. https://www.openstreetmap.org/node/4461142791
https://www.openstreetmap.org/#map=16/55.8682423/37.9515804

Rendering at z13 (earliest shown)
moscow-z13

z15:
moscow-z15

z17: name label shown
moscow-z17


The other 2 dozen that I saw were in very rural areas, especially in the USA, Canada, Australia, Italy and Asian Russia.

**Complex in Hawaii on top of Mauna Kea, previously shown:** https://www.openstreetmap.org/#map=16/19.82366/-155.47612

Rendering at z15; only the largest radio telescope renders:
maunakea-z15

z16 (note that all the buildings visible contain optical telescope domes, not yet rendered)
maunakea-z16

z17: name labels shown
maunakea-z17


**The "Centrum Astronomii" near Piwnice, Poland**
https://www.openstreetmap.org/#map=15/53.09551/18.56104

z14 (first zoom level visible; diameter of telescope is 32m)
piwnice-z14

z15 (Smaller 15m telescope becomes visible)
piwnice-z15


**LOFAR telescopes in the fields near a big research complex; closed ways**
https://www.openstreetmap.org/#map=16/50.8971/6.4255

z15 (not rendered because diameter is null)
daubenrath-z15

z16
lofar-z16

**Cambridge, closed ways:**
https://www.openstreetmap.org/#map=16/52.1653/0.0378

z15: no icons rendered (telescope:diameter is null)
cambridge-z15

z16
cambridge-z16

z17: name label renders on closed way
merlin-z17


I see that almost all name labels start at z17. That's what I've gone with so far, but I could change it to z16 to show the names a little sooner?

I don't know if there is a way to use way_pixels to set the zoom level if telescope:diameter is null? That would be nice to have, but not essential, as seen above.

kocio-pl:
Thanks for thorough research. I guess z16 for labels are OK, since the main reason for not showing name along the icon is that it could clutter the space, so it appears gradually (for example I did it for castles to not compete with city name), but in this case there's a lot of free space around, so basically no risk.

I would not try every possible option for tuning, only diameter is documented on wiki page and for example the biggest radio telescope has no height (it's a hole in the ground), so this is enough. If this value is not added, it's a good indicator that tagging should be completed.

I think you should remove last commits in this branch to get rid of internet cafe problems. I hope this would help.

jeisenbe:
Ok I used
$ git reset 0f9646b
$ git push origin telescope -f
$ git add

And added the icon, amenity-points.mss and project.mml

$ git commit
$ git push origin telescope

But GitHub still showed lines of code as changes related to internet_cafe
etc

So I tried

$ git fetch upsteam
$ git merge upstream/master
$ git push origin telescope

That worked!

amenity-points.mss
```diff
@@ -3106,4 +3120,4 @@
       trunk/marker-height: 6;
     }
   }
-}
+}
```
kocio-pl: Just the small issue - there's a new line missing.

amenity-points.mss
```diff
@@ -1867,6 +1879,7 @@
   [feature = 'man_made_storage_tank'][zoom >= 18],
   [feature = 'man_made_silo'][zoom >= 18],
   [feature = 'man_made_communications_tower'][zoom >= 17],
+  [feature = 'man_made_telescope']["telescope:type" = 'radio'][zoom >= 17],
```
kocio-pl: I think you may make it safely z16+.

jeisenbe:
Ok, I fixed the white space again, and changed the name label initial zoom level to z16:

Hawaii Mauna Kea z16:
hawaii-lables-z16

Cambridge z16:
cambridgez16

kocio-pl:
It works as defined both on named nodes and closed ways, thanks a lot!

matthijsmelissen:
Is it not possible to find an icon that works both for radio telescopes and regular telescopes/observatories?

kocio-pl:
For now we have even no proposition for any other type of telescopes.

jeisenbe:
I don’t think so, they look very different.

I believe man_made=telescope is an optical telescope by default; that is
the common meaning of the word.

So I plan to do a PR using a icon that looks like an optical telescope
observatory dome for those, once there is a good icon available

geozeisig:
A ground station `man_made=satellite_dish` looks very similar to a telescope. Maybe we should render it with this symbol as well.

jeisenbe:
geozeisig  wrote:

> A ground station man_made=satellite_dish looks very similar to a
> telescope. Maybe we should render it with this symbol as well.
>

We could also use the dish icon used for communications “towers” with
structure or type=dish. It is quite similar, but has “waves”.

However, it is only used 175 times per the taginfo box on the wiki. There
is also man_made=communications_dish, used 505 times, and the previously
mentioned tower subtags (which are rendered)

So it needs a proposal or discussion on the tagging list first



>

PR #3475, opened 2018-10-22
Rendering orange dots for gastronomy on z17

### LINKED ISSUE #3410: Orange dot for gastronomy on z17
Follow up to #2945.

Now that we have special color for gastronomy, using orange dot on z17 should be clear enough. I also got used to the idea of urban priority on universal map (see https://www.openstreetmap.org/user/kocio/diary/44769 ) and dots are better known than it was a year ago (entrances have squares, we used blue dots for offices).

I would just leave food court icon on z17, since this is about bigger eating space, it would also distinguish it from restaurants.

[Warsaw, Old Town](https://www.openstreetmap.org/#map=17/52.24948/21.01222)

Before
3gineevx

After
r5vj1hit

[London, Soho](https://www.openstreetmap.org/#map=17/51.51348/-0.13332)

Before
48esq1ph

After
e7cizqu5

kocio-pl:
Fixes #3410
Follow up to #2945

- Rendering orange dot for gastronomy objects on z17, because they are very popular and create visual clutter (excluding food court, which is bigger and less popular, it would be still rendered as full orange icon)
- Removing night club from this category and showing it similar to other entertainment objects like cinema, since this is more about party than gastronomy, it's also much less popular and does not create clutter

[Warsaw, Old Town, z17](https://www.openstreetmap.org/#map=17/52.24951/21.01196)

Before

1agch66m

After

yo_jfsoy

### DIFF
--- a/amenity-points.mss
+++ b/amenity-points.mss
@@ -76,17 +76,49 @@
     marker-placement: interior;
     marker-clip: false;
   }
-  
+
+  [feature = 'amenity_bar'][zoom >= 17],
+  [feature = 'amenity_biergarten'][zoom >= 17],
+  [feature = 'amenity_cafe'][zoom >= 17],
+  [feature = 'amenity_fast_food'][zoom >= 17],
+  [feature = 'amenity_food_court'][zoom >= 17],
+  [feature = 'amenity_ice_cream'][zoom >= 17],
+  [feature = 'amenity_pub'][zoom >= 17],
+  [feature = 'amenity_restaurant'][zoom >= 17] {
+    marker-placement: interior;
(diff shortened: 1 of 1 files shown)


matthijsmelissen:
Great work!

I would keep nightclub in by the way, it is not that different from a bar.

kocio-pl:
It typically includes bar, but as far as I know its primary function is dancing and some other venues and shows. I see it as hotel where you can eat something (typically breakfast), but its primary function is sleeping place.

Tomasz-W:
+1 for moving `amenity=nightclub` to _amenity-brown_ for the reasons given above.

Issue #1704, opened 2015-07-29
dashed rail also on lower zoom levels


matkoniecz:
I am trying to modify railway=rail to make it harder to confuse with minor roads. One of ideas is to keep dashed style also on lower zoom levels.

Unfortunately to achieve truly good results it will be necessary to use overcomplicated SQL merging ways (as dash pattern has big problems with segmented ways, bridges are a typical problem).

Before doing that I want to check whatever the idea of dash style on lower zoom levels is considered as a good idea.

https://cloud.githubusercontent.com/assets/899988/8966556/4fa3601c-3632-11e5-8633-bf0ea717a6d4.png

kocio-pl:
Simply amazing! :astonished:

math1985:
Looks much better.

matkoniecz:
Thank - has anybody already managed to merge ways split into separate ways into one linestring? https://github.com/gravitystorm/openstreetmap-carto/issues/497#issuecomment-42429080 seems to indicate that it is possible.

Here I would want to do it at least for z11, z12 for railway=rail without tunnel=yes.

On higher zoom levels the same effects happen but it is quite hard to notice.

![selection_004](https://cloud.githubusercontent.com/assets/899988/8967050/0280d668-3635-11e5-8daa-29ac6207886e.png)


ping @pnorman

rmikke:
Wow. Simply wow.

imagico:
What would be needed here is probably something like a global alignment for the dash pattern - this would also help to avoid odd interferences between the dash patterns of neighboring lines like here:

http://www.openstreetmap.org/#map=15/52.2528/10.5508

but i guess it would be somewhere between tricky and impossible to implement something like that.

pnorman:
> Thank - has anybody already managed to merge ways split into separate ways into one linestring? #497 (comment) seems to indicate that it is possible.

Yes, and railways are a candidate for this. The technique is to use [ST_LineMerge](http://postgis.net/docs/manual-2.1/ST_LineMerge.html)

This comes with a few issues

* The SQL is more complicated
* It is slower
* It involves linestrings that differ depending on bounding box, resulting in artifacts at MT edges

Because we don't label railways, I don't think the MT edge problems matter here.

[...]

PR #3488, opened 2018-11-01
Rendering railway pattern on z12

kocio-pl:
Related to #3467.

- Rendering railway pattern on z12 instead of a simple black line. It allows to recognize railways from tram lines in urban areas and makes it easier to recognize them in the outdoor (we show power lines later, but one does not have to know that).

[Bratislava](https://www.openstreetmap.org/#map=12/48.1478/17.1361)

screenshot_2018-11-01 openstreetmap carto kosmtik
screenshot_2018-11-01 openstreetmap carto kosmtik 1

[Amsterdam](https://www.openstreetmap.org/#map=12/52.3756/4.8987)

screenshot_2018-11-01 openstreetmap carto kosmtik 6
screenshot_2018-11-01 openstreetmap carto kosmtik 7

[Zagreb](https://www.openstreetmap.org/#map=12/45.8123/15.9757)

screenshot_2018-11-01 openstreetmap carto kosmtik 2
screenshot_2018-11-01 openstreetmap carto kosmtik 3

[Bytom](https://www.openstreetmap.org/#map=12/50.3406/18.9142)

screenshot_2018-11-01 openstreetmap carto kosmtik 4
screenshot_2018-11-01 openstreetmap carto kosmtik 5

### DIFF
--- a/roads.mss
+++ b/roads.mss
@@ -1935,16 +1935,20 @@ tertiary is rendered from z10 and is not included in osm_planet_roads. */
           line-dasharray: 5,2;
         }
       }
-      [zoom >= 13] {
+      [zoom >= 12] {
         #roads-fill, #bridges {
           dark/line-join: round;
           light/line-color: white;
           light/line-join: round;
           [feature = 'railway_rail'] {
             dark/line-color: #707070;
-            dark/line-width: 3;
-            light/line-width: 1;
+            dark/line-width: 2;
+            light/line-width: 0.75;
             light/line-dasharray: 8,8;
+            [zoom >= 13] {
+              dark/line-width: 3;
+              light/line-width: 1;
+            }
             [zoom >= 15] {
               light/line-dasharray: 0,8,8,1;
             }


kocio-pl:
For me this problem is small (I would not notice it if you haven't shown it), but gains should be obvious. Split ways are bigger issue for road labels (names and shields are sometimes located on the edges, where the road splits into "connectors"), but even there it's not severe.

Cracow example looks great to me:

screenshot_2018-11-01 openstreetmap carto kosmtik 8

matkoniecz:
Artifacts (short/missing black parts) are also visible but problem is lesser than I remember.

Maybe I tested it on z11?

kocio-pl:
Cracow example on z11 looks also good:

screenshot_2018-11-01 openstreetmap carto kosmtik 9

matthijsmelissen:
This resolves part of #1704

kocio-pl:
Thanks for reminding, I have completely forgotten about it!... :smile: 

> Here I would want to do it at least for z11, z12 for railway=rail without tunnel=yes.

I've tried to fix the obvious problem here (trams vs. trains), which exists only on z12. However with this small pattern it might work on z11 too, probably with a bit faded color (z11 areas are also a bit muted and there is less other objects, so railways could be too visible then).

> The technique is to use ST_LineMerge
> This comes with a few issues
> - The SQL is more complicated

I'm afraid of that. We can't afford getting rid of SQL for performance reasons, but given that roads code is already hard to maintain and was meant to be cleaned, I don't think it's safe.

> - It is slower

It depends how much slower.

> - It involves linestrings that differ depending on bounding box, resulting in artifacts at MT edges
> Because we don't label railways, I don't think the MT edge problems matter here.

We already have railway labels, so this should be tested.

From what I see on test images, I don't feel glitches are big enough to try any workarounds. Maybe smaller pattern make them less visible than the pattern used on z12+? My goal was to make it small enough to not dominate everything around, and the result is that the pattern is there, but not anything above it, so probably glitches are also harder to be found.

matkoniecz:
> Cracow example on z11 looks also good:

For me glitches directly below "Kraków" label, on track below and toward right from city label (black parts are almost completely  gone) are significant at z11.

There are also other quite significant glitches.

kocio-pl:
BTW - after your comments I have also found glitches on current rendering, so this looks like general problem.

PR #3492, opened 2018-11-02
Render optical telescopes

### LINKED ISSUE #3389: Render man_made=telescope

jeisenbe:
Fixes #3389

- Add "radar" and "gamma" as types of non-optical telescopes, rendered the same as radio telescopes

- Render all other telescopes as optical telescopes (i.e. telescope:type is not radio, gamma or radar), with initial zoom level 17, increasing to 16, 15 and 14 for telescopes greater than 2, 4 or 8 meters in diameter (aperture). Name label rendered at zoom 17.

- Add telescope dome svg icon, courtesy of @Tomasz-W

This will complete the rendering of all man_made=telescopes.

**Mauna Kea, Hawaii**
Some of the world's biggest telescopes. Only about 100 telescopes are bigger than 2 meters, and only a dozen are bigger than 8 meters and would render at z14. Several are here:
https://www.openstreetmap.org/#map=17/19.82366/-155.47612
z14
maunakeaz14

z15
maunakeaz15

z16
maunakeaz16

z17
mauna-kea-17

**Paranal Observatory, Chile**
The other place with multiple huge telescopes >8 meters (the 4 shown are all part of the VLT; "Very Large Telescope"
https://www.openstreetmap.org/#map=15/-24.6294/-70.4018
z14
chile-z14

z15
chilez15

z16
chilez16

z17
chilez17

z18
chilez18

**Hampstead Observatory in urban London, England**
An example of a historic/educational telescope in a city. Renders at z17.
https://www.openstreetmap.org/#map=17/52.166756/0.034975
z17
hampsteadz17

z18
hampsteadz18

**Suburban Berlin Observatory**
Several optical telescopes at a research site on the outskirts of Berlin
https://www.openstreetmap.org/#map=17/52.4049489/13.1030466
z17
berlinz17

z18
berlinz18


kocio-pl:
> Add "radar" and "gamma" as types of non-optical telescopes, rendered the same as radio telescopes

Gamma is directly discouraged on the `man_made=telescope` wiki page ("use telescope:type=radio + telescope:spectrum=*") and radar falls into the same category. Their usage is very small (15 and 6 respectively), so we can retag them all in a short time and no special rendering support is needed.

jeisenbe:
The wiki is wrong about gamma ray observatories; gamma is shorter
wavelength and higher energy than X-rays, so it’s the opposite of radio
waves, which are very long wavelengths and very low energy per photon.

But I understand about the low usage.

Would it be ok to still exclude them from rendering with the optical
telescope dome icon? Or should I remove that line too?

kocio-pl:
It would be good to expand also this one:

https://wiki.openstreetmap.org/wiki/Key:telescope:type

jeisenbe:
Most optical telescopes are just tagged as man_made=telescope without
telescope:type. This is unsurprising, because the popular idea of a
telescope is an optical telescope, and these are also the most common.

So that’s why I wanted to render all man_made=telescope with the same icon,
unless they were specifically tagged as radio telescopes.

amenity-points.mss
```diff
@@ -553,6 +553,18 @@
     }
   }
 
+  [feature = 'man_made_telescope']["telescope:type" = 'optical'],["telescope:type" = null] {
```
kocio-pl: This should probably look like this:

  [feature = 'man_made_telescope'][telescope:type = 'optical'],
  [feature = 'man_made_telescope'][telescope:type = null] {

because AFAIK this would check if some object has "telescope:type" = null without checking if this is telescope at all, which would lead to strange problems. 

Also I don't think we need adding quotation marks for keys.

amenity-points.mss
```diff
@@ -1867,6 +1879,7 @@
   [feature = 'man_made_silo'][zoom >= 18],
   [feature = 'man_made_communications_tower'][zoom >= 17],
   [feature = 'man_made_telescope']["telescope:type" = 'radio'][zoom >= 16],
+  [feature = 'man_made_telescope'][zoom >= 17],
```
kocio-pl: If we use optical vs all the other for icons, I guess name rendering should also check it, not radio vs the rest.

amenity-points.mss
kocio-pl: The only key we need quotations is "natural", but this is just because this is a PostgreSQL keyword too by accident.

jeisenbe:
I thought the quotations were necessary when the key contains a semicolon?

This was also done for tower:type e.g. [ “tower:type” = ‘lighting’ ]

jeisenbe:
I just tried it, and mapnik fails to load, saying:
amenity-points.mss:544:33 Missing closing ] of filter

I did not add or subtract any brackets, I just changed all occurrences of
“telescope:type” to telescope:type in amenity-points.mss and project.mml

jeisenbe:
I changed the code as suggested, except I did not remove the
telescope:type=radio feature. Now telescope:type I null and optical render
with the done icon and have labels at z17, and only telescope:type = radio
has the dish.

I don’t want to assume that all other types are radio telescopes and render
at z16 with the dish icon. Because it is possible that infrared and x-ray
observatories will be tagged in the future, and gamma telescopes are
already tagged; I believe these should not be rendered the same as radio
telescopes

kocio-pl:
> I believe these should not be rendered the same as radio telescopes

It's easy to add special icons for them in the future, once we have same sane amount of them (not too much, since they are probably rather rare) and I think it's better to show them at all now.

jeisenbe:
I've updated the code as suggested. Here's the rendering of Mauna Kea in Hawaii again.
The NASA Infrared telescope now renders with an icon, but with a dish instead of a dome, unfortunately. However, all the others render correctly. 

(I read up on gamma ray telescopes; they look like dishes. X-ray telescopes are only usable in the high atmosphere or outer space. Radar telescopes are a rare, specialized type of radio telescope. Infrared telescopes are similar to optical telescopes, but only 4 are tagged so far)

Mauna Kea z16
maunakea-z16

kocio-pl:
This enhances rendering of telescopes nicely and sane, thanks! 

Some discussion will be needed on Tagging how to make the distinction between touristic view devices and astronomical ones, but there is already some tag with low usage, which might be documented:

https://taginfo.openstreetmap.org/tags/tourism=telescope

geozeisig:
I think `tourism=telescope` is the same as `amenity=binoculars`.

[https://taginfo.openstreetmap.org/search?q=amenity%3Dbinoculars](https://taginfo.openstreetmap.org/search?q=amenity%3Dbinoculars)

PR #3498, opened 2018-11-06
Add rendering for healthcare tag scheme

### LINKED ISSUE #2938: Unified rendering of healthcare objects

jragusa:
Fixes #2938

* change `health-color` to <del>`#CC0000`</del> `#BF0000`
* change icon of hospital
* add support of healthcare tag scheme with a dot from z17
* support duplicated healthcare tags with amenity tags (clinic/dentist/doctors/hospital)
* move icon of`amenity/healthcare=dentist` to z18 and  display a dot at z17
* move `amenity/healthcare=clinic`to z17
* preserve priority of `amenity=pharmacy` if tagged in combination with `healthcare=*`

https://www.openstreetmap.org/way/337002580 (new hospital icon and colour)
Before
healthcare_color_before
After (fix correct picture)
healthcare_z19_finalafter

https://www.openstreetmap.org/#map=17/46.19317/6.23683 (zoom level and dots)
`z17`
Before
healthcare_z17_before
After
healthcare_z17_finalafter

`z18`
Before
healthcare_z18_before
After
healthcare_z18_finalafter


Tomasz-W:
> support duplicated healthcare tags with amenity tags (clinic/dentist/doctors/hospital)

@jragusa Please add also `amenity=pharmacy` as priority one in case if there will be combination with `healthcare=*` to show an icon instead of dot then.

jragusa:
@Tomasz-W fixed

Tomasz-W:
@jragusa Just one more thing: I'm wondering if `#CC0000` isn't too prominent comparing it to other icons, can you test a little bit darker `#BF0000`?

jragusa:
`#BF0000`
healthcare_z18_finalafter_colour

healthcare_z19_finalafter_colour

jragusa:
@Tomasz-W yes I included the healthcare tags following this [comment](https://github.com/gravitystorm/openstreetmap-carto/issues/2938#issuecomment-366990786) and they are all displayed as a dot.

jeisenbe:
Have you tested it next to a named retail area and next to trunk roads and
motorways? Those retail name labels are fairly red.

jragusa:
found it: https://www.openstreetmap.org/#map=17/46.34787/6.43486
healthcare_z17_comparison

Tomasz-W:
@jragusa I've read some posts again, and there is a little misunderstanding, sorry for it. I mean to render every `heanlthcare=*` object with a dot at least, even e.g.` healthcare=gummy_bears_doctor ` (no matter of value). We have the same for `shop=*` and `office=*` keys and it's not causing any big problems. In this way we avoid series of issues named "Add rendering for healthcare=..." in the future.

matkoniecz:
>  render every heanlthcare=* object with a dot at least, even e.g.healthcare=gummy_bears_doctor (no matter of value)

I think that rendering freeform values is not preferable. With shops it may be acceptable as there are some weird rare shop, but in case of healthcare I would expect that all uses are covered by tagging scheme.

Tomasz-W:
Here is a TagInfo list of `healthcare=*` values in use: https://taginfo.openstreetmap.org/keys/healthcare#values . There is a lot of less known and undocumented 
(but still reasonable) values, so I don't see a problem with allowing all of them as red dots.

jragusa:
@Tomasz-W Sorry, it was a misunderstood from my side.

So what's the plan now ? Do I restrict following the TagInfo list or open to any value (`healthcare=*`) ? 
There is only 2 values missing in my PR (pharmacy and hospice).

Tomasz-W:
@jragusa open to any value (like shops and offices are)

matkoniecz:
I would restrict list to the documented values.

kocio-pl:
This case is not easy to decide. From one side - shops have much longer list of values and they do not have a global definition, while we have Healthcare 2.0. On the other hand having `healthcare=*` has very clear meaning and is limited by that, unlike gastronomy objects for example - they use generic `amenity=*` space, which makes it impossible to use such easy limits.

matkoniecz:
I think that it this case better mapper feedback would be to encourage use of documented values or documenting them if something is missing.

While in case of shops I see why rendering `shop=gummy_bears` may be considered desirable, I think that in this case not rendering, say `healthcare=gummy_bears_doctor` is desirable.

kocio-pl:
I think you proposition is perfectly valid for now. We can review this decision later if anyone will complain or just extend the rendering when documentation will be extended. Rendering healthcare is a big step ahead in itself, so we may stay on the safe side and see how it works.

jragusa:
So, I restrict following the TagInfo list for now ?

kocio-pl:
They are just used values. Documentation for tags is on the wiki:

https://wiki.openstreetmap.org/wiki/Key%3Ahealthcare#Values

ghost07tula:
What should we do to areas tagged like healthcare=yes? I.e. area around emergency=ambulance_station?

ghost07tula:
Makes sense. Will ask in a tagging list

dieterdreist:
sent from a phone

> On 9. Nov 2018, at 09:21, jeisenbe <notifications@github.com> wrote:
> 
> Is that really a healthcare facility? It’s more like an office plus garage
> or depot; not a place useful to the public.


it is different to a garage+office because you have the paramedics there, if you need help it would be „almost as good“ as a hospital (they would bring you there quickly)

jragusa:
@geowas-github yes it's supported in the PR

ghost07tula:
@jragusa and what about areas with healthcare=yes?

Adamant36:
@ghost07tula, its in the PR code. So I assume its supported.

openstreetmap-carto.lua
```diff
@@ -14,6 +14,7 @@ local polygon_keys = {
     'building',
     'building:part',
     'harbour',
+    'healthcare',
```
kocio-pl: I think this might be harmful - did you try first importing data with this code and then rendering? I suppose every healthcare object will be no longer in hstore column then, so the whole database code should be different.

We did not such change after v4.0.0, and it should be announced much earlier and coordinated with forks and OSMF admins.

kocio-pl:
Dentist code has a marker problem - it shows big red dot before z16, for example:

screenshot_2018-11-20 openstreetmap carto kosmtik

openstreetmap-carto.lua
jragusa: Fixed, I removed the variable from the Lua script.

kocio-pl:
I think that `amenity=doctors` is more like a `healthcare=centre` (so they should be rendered the same), while typical `healthcare=doctor` is more like dentist (different icon on z18+, of course, but probably just a dot on z17).

I would also drop `healthcare=pharmacy`. It's very popular (second next to `doctor`), but it's not even mentioned on the wiki and most of them are also tagged with `amenity=pharmacy` (only ~40 from 15k are not), so it's not changing too much at the moment. It can be discussed on a Tagging list and we can update this code later if some other main healthcare objects will be documented.

See also https://taginfo.openstreetmap.org/tags/healthcare=pharmacy#combinations.

jragusa:
@kocio-pl what could be the icon for `healthcare=doctor` on z18+ ? a dot or something else ?

Besides to be very popular, I added `healthcare=pharmacy` following this [comment](https://github.com/gravitystorm/openstreetmap-carto/issues/2938#issuecomment-433334766). There was some discussions about unifying all related healthcare POI into the `healthcare` key according to the [wiki page](https://wiki.openstreetmap.org/w/index.php?title=Proposed_features/Healthcare&oldid=1318635#healthcare_versus_amenity).

kocio-pl:
I would not change an icon for doctor on z18+, it can be still a medical bag. I just thought about z17 change for it (a dot instead of icon) to show that it's smaller. There are a lot of small doctor's offices, which are similar to dentists' offices or midwives' offices, because they're located in the ordinary residential houses, and only a plaque makes it visible.

I would be happy if more consistent, clear and extendable `healthcare` key would be used instead of already too vague `amenity` namespace (which covers pretty much everything and has no clear meaning as a category), but I think that at least basic documentation should be provided so all data consumers can rely on a single source. If something important is missing in the documentation, someone should update it first.

kocio-pl:
The comment you were referring to:

> The healthcare proposal already focused this migration and suggests just to render both tagging schemas. After a while, the legacy tags might be cleaned up and forgotten.

This is also what I think. However this shift can be a lengthy process (it can take rather years than months) and it's better to make a new namespace well documented, if we want something really better than a temporary fix. I think that the fact that it will start to be visible on default map will bring more attention to fixing and updating specification and documentation, which will be only a good thing.

There are some not clear things to decide - maybe pharmacy would be better tagged as `shop`? An what about veterinary - is `healthcare` only about humans? These details should be discussed, and once they are clear, we can easily implement them. I suspect that for such a broad topic the updates will be needed anyway.

jragusa:
OK, I corrected following your suggestions

kocio-pl:
Thanks! However there's still undocumented `chiropractor` and `hospice` items, but lack of documented `dialysis` entry to be fully synchronized with a wiki documentation.

kocio-pl:
There are still `chiropractor` entries in `project.mml` instead of `dialysis`, unfortunately. Please replace them and make sure it works. The rest looks OK.

jragusa:
I made tests for chiropractor and dialysis. It now work fine.

Sorry, for this trivial error, sometimes, I have my head in the clouds.

kocio-pl:
A lot of work, but I think it was worth all the hassle. Thanks for your dedication and patience! 

I think we will come back to this namespace sooner or later, I'm not only sure which one is more true in this case.

geozeisig:
Can someone tell me why the icon for the [Charité](https://www.openstreetmap.org/way/583306346#map=16/52.5258/13.3779) has no name on the map?

kocio-pl:
Most probably because of the road labels and road arrows.

geozeisig:
I give up. I've already changed a lot but the conflict persists rendering the name of [Charité](https://www.openstreetmap.org/way/583306346#map=17/52.52568/13.37636).

kocio-pl:
I would be surprised if anything would change without solving problem I mentioned lately. If you could try to solve it, the chances are high in my opinion. Anybody else interested in fixing this?

geowas-github:
> Personally I think its miss-tagged. Things there don't seem to fit the Wiki definition of a hospital. For instance the Institute of Chinese Life Sciences, which is north of the icon but still within the hospital bounds. You might try mapping the area as a university. Since that's what it actually is, even though some places in it provide patent services, and then applying the hospital tag to a more specific area like a building or something. Maybe university would render before hospital. So the name would display.

Which is tagging for the renderer.

Certainly Charité is both a hospital and (part of) a university, but the main purpose for its existence is that it is a hospital (incorporating education). And for the other point: Most hospitals also provides amenity=clinic type services.

Adamant36:
Its not tagging for the renderer if its miss-tagged in the first place and tagging it properly also happens to get the name to display. 

I wasnt suggesting just picking random tags until one worked. Which would qualify as tagging for renderer. The buildings should be properly tagged either way. Its missleading to just tag the whole area as a hospital and call it a day when it provides a bunch of other services that can also be tagged.

 Plus, I noticed the whole area is tagged as emergecy=yes and wheel_chair=yes. surely some of the buildings provide emergecy services and some of the compound is wheel chair friendly, but I doubt the whole does or is.

Anyway, its not a tagging disscussion. Except in the case where tagging might possibly be causing the problem. So it really doesnt matter outside of that.

geowas-github:
@Adamant36: even if we would tag Charité as university: why should it be rendered differently than a university and be hidden behind "road labels and road arrows" while the university is not. Science before road labels before charity?

BTW: I don't think there is a clear sceme for tagging hospitals linked to universities. If I look at the Universtitätsmedizin Mainz here around the corner: to most people it's clearly a hospital first of all. For students of medicine it will be the university. Either way: neither tagging the area as a hospital with university nodes/smaller areas or a university with hospital nodes/smaller areas inside will be correct. It's all mingled, so one has to decide for one (whatever). But this doesn't help solving the question why there is no text lable for Charité.

kocio-pl:
What about `healthcare=hospital` + `amenity=university`?

geozeisig:
The university is certainly not the reason. The [campus area](https://www.openstreetmap.org/way/235809307#map=17/52.52591/13.38265) is displayed correctly and does not overlap. Of course, the students go to the hospital.

Adamant36:
@geozeisig, there's a part [here](https://www.openstreetmap.org/#map=19/52.52589/13.38005) where the hospital grounds almost overlaps itself and there's a whole in it with a weird yellow line going south a bit. It might have something to do with that. There was a hospital a while back in the bay area somewhere where the icon displayed a few yards away from the hospital in a residential area and the name didn't rendering. It was because of a similar issue. I was never able to fix it unfortunately, but if there's not a clear center for the icon and name render or anything overlaps itself it tends to cause rendering problems. So you might check that (I would have fixed it myself but I don't want to screw with the area and the website keeps giving me a loading error for some reason).

Adamant36:
Instances of other places where the hospital name doesn't display correctly (not sure why. Could be rendering priority).

https://www.openstreetmap.org/#map=18/37.59184/-122.38181

https://www.openstreetmap.org/#map=18/37.67947/-122.47399 (only doesn't renders name at z18. Might be road oneway arrow blocking it)

https://www.openstreetmap.org/#map=18/37.37604/-122.06447 (doesn't render name at z17. Not sure why).

https://www.openstreetmap.org/#map=17/37.33501/-121.99849 (It blocks itself out in this situation. Probably miss-tagging).

https://www.openstreetmap.org/#map=19/37.33342/-121.88524 (another example of where the name is being blocked out by other stuff due to rendering priority. In this instance a post office, but zoom out).

geozeisig:
I've simplified something else. And now it works.
charite
Whether it really is related to my recent changes, I can not believe, because it was only a few minutes later. The server was down this morning, so we better wait.

Issue #3176, opened 2018-04-14
Tower baseline alignment


georgek:
There are a few icons representing tall objects now shown on the map. For example, wind turbines and communications towers. But the icons are aligned such that the centre of the icon is over the actual node position which leads to them looking misplaces, especially at lower zoom levels.

An example can be seen here with the wind turbine in the middle: https://www.openstreetmap.org/#map=16/52.5975/1.2626 It looks like it's in the water but it's not.

Can the alignment be changed so that the bottom of the icon is at the node position?

dieterdreist:
sent from a phone

> On 15. Apr 2018, at 08:23, geozeisig <notifications@github.com> wrote:
> 
> The icon is a symbol and not a small reflection of reality. This can lead to small inaccuracies in the placement.
> 


-1, the icons should be placed in a way that it is intuitive to understand to which spot they refer to. In case of e.g. a tower, pole, chimney or mast, this is the base.

matkoniecz:
The icons should be placed in a way that it is intuitive to understand to which spot they refer to. In case of e.g. a tower, pole, chimney or mast, this is the base.

@dieterdreist I reposted your comment as I agree with it and Github mail-to-cmment transform hides it (maybe it expects to post reply above quoted text? or to delete text of an original message?)

jeisenbe:
This looks fixable with
`      marker-transform: 'translate(0,6)';`
This code will move the marker icon up by 6 pixels, which should work for most 14 pixel icons.

Here's an initial test in Honolulu, Hawaii
z14 Communications towers Original
![z14-honolulu-0](https://user-images.githubusercontent.com/42757252/48302762-0806cb80-e544-11e8-8296-ef771ddcbf9a.png)
Up 6 pixels
![z14-honolulu-6](https://user-images.githubusercontent.com/42757252/48302774-44d2c280-e544-11e8-9ba4-bcd52c915a84.png)

z18 Lighting towers original
![z18-honolulu-0](https://user-images.githubusercontent.com/42757252/48302776-50be8480-e544-11e8-8ffc-b309fa8e8c88.png)
Up 6
![z18-honolulu-6up](https://user-images.githubusercontent.com/42757252/48302778-587e2900-e544-11e8-863b-8072e8537e75.png)

z15 Lighthouse, Original
![z15-diamondhead-0](https://user-images.githubusercontent.com/42757252/48302781-5fa53700-e544-11e8-99d5-3f69683d8fbc.png)
[...]

PR #3501, opened 2018-11-08
Add rendering for man-made cranes

### LINKED ISSUE #3478: Add rendering for man_made=crane
Why cranes do not appear rendered in OSM?

Wiki: https://wiki.openstreetmap.org/wiki/Tag:man_made%3Dcrane

Examples: https://www.openstreetmap.org/node/6000837250

jeisenbe:
Fixes #3478

- Add svg crane icon, rendered in man-made-color (gray) thanks to @Tomasz-W
- Add man_made=crane to project.mml file
- Render man_made=crane with initial zoom level based on height, following same zoom levels as man_made=tower
- Render name label at z17

EDITED: initial proposal was to render at z14 and z15 based on height >100 and >50, now changed to only z16 for height >50 and z17 otherwise.

**Port of Singapore**
https://www.openstreetmap.org/#map=17/1.2655/103.7928
z17
z17singapore
Future rendering at z14 (assuming height tags of 51 to 99 meters are added to these cranes; currently they would only render at z17 as above)
z14singapore-future

**Milan train station**
https://www.openstreetmap.org/#map=15/45.4851594/9.1832991
z14
z14milano
z15
z15milano
z16
z16milano

**Port of Venice**
https://www.openstreetmap.org/#map=15/45.4729014/12.2435279
z13
z13venice
z15
z15venice
z17
z17venice

**River port cranes with name**, Blauer Kran, height 24m
https://www.openstreetmap.org/#map=16/50.1129065/8.7519599
z16 Blauer Kran
z16blauer
z17 Blauer Kran
z17blauer
z17 Evo - height untagged; mapped as area (closed way)
z17evo

**Big Willi river boat crane** - height 12m
z17willi


Tomasz-W:
@jeisenbe Please test icons for all cranes types: https://github.com/gravitystorm/openstreetmap-carto/issues/3478#issuecomment-433604306

jeisenbe:
The additional icons are very nice.

But I don’t think the usage of different crane types is clear enough for us
to start rendering each type differently. I tried to ask about it on the
tagging list but did not get a response.

For now I would like to start rendering cranes with one icon, and later we
can consider adding others.

jeisenbe:
I added` marker-transform: 'translate(0,-6)';` so the base of the crane will be near the node or center of the area, as suggested in #3176 "Tower Baseline Alignment"

jeisenbe:
Renderings with marker-transform translate (0,-6) and reduced text-dy (can't forget that!):

Big Willi previous:

Big Willi new (up 6 pixels) z17:
willi-up6-z17

Singapore previous:

Singapore up 6 pixels z14:
z14-singapore-up6

I now think it might be better to limit the zoom level to z14 for height >100m, instead of 13m as with towers. Chimneys and masts do not render above z16 and z15 respectively; only towers render at z13.

 I find the tall towers a bit distracting on a city map at z13, especially in areas near the equator like Honolulu and Singapore, but at least there are not dozens of them in one place like cranes at a port. 

Towers at z13 are probably fine in Stockholm or St. Petersburg near 60 degrees N, where z13 is the same as z14 at the equator.

Lebanon *(tested with zoom z13 as initial level for all cranes as an example)
z13* test - rather too prominent
lebanon-z13
z14* test - probably ok (with marker-transform: 'translate(0,-6)';)
lebanon-z14
z14 without marker-transform
lebanon-0-z14

So I'm changing the code to ```

    [zoom >= 14][height >= 100],
    [zoom >= 15][height >= 50],
    [zoom >= 16][height >= 30],
    [zoom >= 17]
```
Agreed?

**Comment by kocio-pl** [comment]
Lebanon example shows me that it makes sense to reduce it.

**Comment by jeisenbe** [comment]
Ok, it should all be ready to merge now; I just tested it again.

**Comment by jeisenbe** [comment]
Final test rendering to show label with new text dy:
crane-new

**Comment by jeisenbe** [comment]
@Kocio-pl, sorry about that, it should be fixed now. I'll fix the telescope code when I adjust the zoom levels with the next PR.

I also removed `marker-transform translate(0,-5)` and the text-dy changes from this commit so that this can be merged now, even if we decide not to do the icon baseline adjustment for vertical features like cranes and towers in #3507; I will add the changes there after this is merged.

**Comment by Adamant36** [comment]
Just my two sense but I'm against this whole rendering things based on height thing for items like this. It just unnecessarily complicates things and doesn't help figuring out what needs to be mapped by looking at the map any easier. There should be some standard for what items render at what zoom levels and it shouldn't be dependent on arbitrary variables like height that most people don't know, care about, and can't even figure out by looking at the object in the real world. 

Especially for something like cranes where most of them are going to be supper tall anyway and the height data isn't even for them isn't even known because its not available on tag info. So the height ranges for rendering are completely arbitrary. If there are a few outlier small cranes, they shouldn't get special rendering rules anyway.

**Comment by kocio-pl** [comment]
I don't agree with that. Wen there are dependent objects (like bell tower, which is typically part of a church), their height is not that important, since I don't want parts to be visible more than the whole. But for standalone objects it's different. Eiffel tower is high and well visible, the same is tower on Alexanderplatz in Berlin. Their height is important.

When it comes to cranes it's possible to see them as part of some industrial area, so dropping height is possible for me.

**Comment by jeisenbe** [comment]
@Kocio-pl, would you like to see them all render at z17 only, or should we change the min zoom level to z15? Then the code similar to the code for chimneys, which are also found at industrial sites. Thus:
```
[zoom >= 15][height >= 50],
[zoom >= 16][height >= 30],
[zoom >= 17]
```
or 
```
[zoom >= 15][height >= 100],
[zoom >= 16][height >= 50],
[zoom >= 17]
```

**Comment by dieterdreist** [comment]
sent from a phone

> On 13. Nov 2018, at 12:31, kocio-pl <notifications@github.com> wrote:
> 
> But for standalone objects it's different. Eiffel tower is high and well visible, the same is tower on Alexanderplatz in Berlin. Their height is important.


both higher than 300m, comparing them to bell towers of churches is inappropriate, they are a different beast

**Comment by jeisenbe** [comment]
Here's some larger test images from Singapore, where part of the port is just next to Downtown and Chinatown. This city is one of the largest that is almost on the equator (about 1 degree north), so due to the Mercator projection it has some of the most crowded tiles at each zoom level. It looks like z14 is too soon, but z15 is probably fine:

z14 Singapore with cranes*
z14-singapore-cranes
z14 no cranes
z14-no-cranes

z15 with cranes*
z15-cranes-tanjong-pagar
z15 no cranes
singapore-z15-no-cranes

z16 with cranes*
z16-cranes
z16 no cranes
z16-no-cranes

*The pictured cranes do not have a height tag, so they would not actually render at z14 or z15 on any version of this PR, but I've seen them, and they are all tall enough that they will render if height tags are added in the future

I've pushed this code change:
```
[zoom >= 15][height >= 100],
[zoom >= 16][height >= 50],
[zoom >= 17]
```

jeisenbe:
A 12 meter (40ft) tall crane isn’t a landmark, except in the immediate
area, but a 120m (400ft) tall harbor crane (like those in Oakland or Long
Beach) is visible for kilometers.

But I’m happy to make more changes if any of the maintainers would like.

Adamant36:
That makes sense. 

As a side note, the opinions of the maintainers aren't the only ones that matter. Its a group effort beyond just what they think. Therefore everyone's feedback, irrespective of if they are a maintainer or not, should be considered.

kocio-pl:
I don't have a strong opinion on that matter, so both solutions are acceptable for me - both have their pros and cons. 

I treat public tickets as a way for everybody to give some input to improve the code (or at least show some problems and questions to be aware of).

Adamant36:
I think z15 and z16 is to soon to start rendering them. Especially z15 and considering how many are in a single area. Plus, they look out of place with other things rendered at level. So, id suggest starting at z17 for 100ft ones and Z18 for 50ft ones. Otherwise, they will stand out to much, compared to other things rendered at those levels. Which are more spread out and not as condensed.

jeisenbe:
Chimneys are rendered at z16 at height >50m and z17 at > 30m.

Cooling towers are also rendered at z15, and there are several in one
place, usually.

So I can’t imagine rendering tall cranes any later than z16, unless we also
change the zoom levels for chimneys, cooling towers and perhaps other
towers and masts

Adamant36:
Their different things though. In most cases your not going to find 10 chimneys lined up close next to each other. Whereas, you will with cranes. As your test renderings show. So the its irrelevant comparison. Also, anything can count as a "landmark." yes, rendering zoom levels are partly based on that, but its also increasingly based on the density the item being rendered. 

While we can sit here all day and argue about what constitutes a "landmark" or if something should be rendered at a certain zoom level because it can be seen from a certain number of miles away, everyone can easily tell if something is covering other things up at a certain zoom level or causing "over crowding." Being more mindful of things like that are the cost of dealing with a map that renders everything. There is no easy way to create a hierarchy of rendering importance with a map that shows as much as this one does, everywhere in the world. 

The trend has clearly been toward a more cleaner map at higher zoom levels and trying to move away from often contentious debates about "rendering importance" that turn people off of contributing to the project. See the scrum issue #3372 and the follow up in #3445 for example. Also the more recent comments in issue ##2436. Including around this comment https://github.com/gravitystorm/openstreetmap-carto/issues/2436#issuecomment-421625668 about project stagnation and its causes.

jeisenbe:
Many power plants have several chimneys or cooling towers in a row. For example, Aes Alamitos between Long Beach and Seal Beach.

Does anyone else think the cranes should only render at z16 or z17?
I believe z17 is too late, as seen by the views of harbors, which are quite empty with no icons at z16, even at areas near the equator.
But I could support rendering cranes at z16 to z18 based on size, same as chimneys.

jeisenbe:
It would prevent rendering small cranes too early. The 100m tall container
cranes are often 100m long by 20m wide, so they take up quite a bit of
space on the map. The fact that there are a whole line of cranes along a
harbor at z16 is an accurate representation of the port, just like a line
of cooling towers or chimneys at a power plant is appropriate.

Adamant36:
Its always a trade off between accuracy and clutter. I choose to prioritize clutter in this situation. Maps are an abstraction and no one, including the maintainers, expect the map to be a one to one recreation of reality. There's been plenty of issues about map clutter though. It doesn't help to spend a bunch of cleaning up certain zoom levels if they just get cluttered up again every time a new tag comes along. So there should be standards. I'm fine seeing what other people think though. I could be wrong in this situation, but I still think its something you should give more consideration to. The only reason the map isn't complete mess at any zoom level is because of constant effort to not let it get that way.

dieterdreist:
sent from a phone

> On 19. Nov 2018, at 21:25, Tomasz Wójcik <notifications@github.com> wrote:
> 
> I consider z17-18 as a good starting zoom level for cranes, even for a big ones, because at the one hand they might be a good orientation point, but on the other they are useless for almost every people exept factory/ construction workers.


a lot of things are useless, think about mountain peaks, totally useless, and we render them from z11

Cranes on construction sites are usually of temporary nature, but there are those in ports, factories and similar which can be significant landmarks.

jeisenbe:
All right, I have pushed a commit to render cranes starting at z16 for >50
meters, same as man-made=chimney.

Hopefully this is satisfactory. Any higher starting zoom level would
require changes to chimney rendering for consistency

Tomasz-W:
After checking some european cities, I still consider adding cranes from z16 as something which make them look much more important than they actually are (even taller ones). I think z17 would be better.

Adamant36:
The only reference I have for cranes is the bay bridge in SF. They aren't very noticeable there at all unless your looking directly at them. 
bay bridge
Although, that might be an outlier. Under normal circumstances though I don't think they would stand out in an industrial area anymore then anything else there though, like the barges or docks or whatever, since places where you find them are usually pretty "busy." Nore are they more important. For instance, they don't even stand out that much in a picture of the place provided in the test renderings. 
tan jong
If anything they are the least interesting part of the scene in the second and third images
tong jong port
tong jong

Whereas, compared to a chimney, the chimney is the main or even notable thing where they are. For instance at power plants. Which confirms to me that the comparison between the two is a bad one and how they are rendered shouldn't be a part of the decision. There's clearly no comparison between a crane at a dock, even a large one, and a nuclear power plant chimney.
nuclear-plant-cz
powerplant 1
powerplant 2
Or this fiber glass plant
knauf
three mile island

Therefore, on further introspective and in light of the evidence I agree with @Tomasz-W. Cranes should start at z17. If done by height then smaller ones should be rendered at z18, but none be rendered any early then z17.

dieterdreist:
cranes certainly are often landmarks, like here where they dominate the
skyline:
https://www.planet-wissen.de/hamburger-hafenimpressionen-110~_v-gseagaleriexl.jpg

I am not sure for chimneys, but certainly for cooling towers, as you
demonstrate on your pictures, cooling towers usually have a much greater
"spatial footprint" compared to chimneys.

I do not say we must show cranes below z17, but dismissing it because
cooling towers have a stronger visual impact seems not sufficient.

kocio-pl:
I still have no strong opinion on that. I guess we can merge any option now and tune in the future if we discover places with a need to change the zoom level.

Tomasz-W:
I think that zoom levels <17 should be left for objects with big size and big importance/ big cultural meaning. Cranes belong only for the first one group, so in my opinion it's way too little for rendering even bigger ones on <z17. Quoting the classic, "I'm strongly against" rendering cranes on z16 ;)

jeisenbe:
I’m interested in consistency for tall man-made objects. I’ve just opened a
PR that will render towers later (not before z14), but >160m towers would
still be rendered at z14 and >80m at z15.

I would find it odd to see a 50m talk, 1 meter wide lattice radio tower or
mast at z16 but not a 120m tall, 100m long, 20m wide crane.

Consider that we render power pylons / towers at z14 and z15 too, though as
small squares rather than an icon, and power poles at z16

Would you like to use a smaller crane icon for z15 and z16 which would take
up less space on the map, perhaps? The Humanitarian/ HDM style uses smaller
icons on lower zoom levels, and this looks good.

Adamant36:
Or, I suggest doing what is usually done in these types of circumstances, do the PR for adding the cranes at the zoom level me and @Tomasz-W suggest, wait for a while to give it time to be deployed, then open a new issue that illustrates places where rendering based on height would actually be a benefit. So it can be discussed properly. Because ultimately height is a separate issue to just rendering it in the first place and its better to have it rendered then not.

Adamant36:
@dieterdreist, obviously I meant zero relevance as far as map viewers are concerned. Since, I stated its highly doubtful 99% of the people that browse open street maps are crane operators or that rendering them at z16 versus z17 would even be useful to them if they where. Whereas, it does cause useless clutter to render them at z16 and its also wrong to say they should be rendered at that level because of being "important" landmarks (or just for the sake of "standardization").

dieterdreist:
sent from a phone

> On 30. Nov 2018, at 18:33, Adamant36 <notifications@github.com> wrote:
> 
> there's zero "crane tourists" out there that are dying to find a dock with cranes taller then a certain height. Is it good to know they are there?


there are even cranes used as tourist attractions, but it would probably merit more specific tagging, e.g.
http://dinnerinthesky.com/


Cheers, Martin

dieterdreist:
sent from a phone

> On 30. Nov 2018, at 18:33, Adamant36 <notifications@github.com> wrote:
> 
> Plus, they have no cultural significance.


in cities like Hamburg they do have cultural significance, they are part of the “panorama”

Adamant36:
There's always exceptions ;). So what rendering level are you for then?

dieterdreist:
sent from a phone

> On 30. Nov 2018, at 18:33, Adamant36 <notifications@github.com> wrote:
> 
> Even if they had the same height. Height doesn't determine importance in 99% of the cases.


citation needed. I would expect height to correlate with importance for many objects (not alone).

dieterdreist:
sent from a phone

> On 2. Dec 2018, at 10:58, Adamant36 <notifications@github.com> wrote:
> 
> There's always exceptions ;). So what rendering level are you for then?


it depends on the icon, if the icon is small enough not to be too intrusive I would probably render them a bit later than towers

Adamant36:
I wasn't taking issue with height in general. Yes it has importance for some objects, like crosses or memorials, only in this case. Where the difference between a 100 foot versus a 60 foot crane is minimal. In the case of how it should be rendered. Either cranes have importance in a landscape or they don't. The 40 foot difference probably doesn't matter.

dieterdreist:
sent from a phone

> On 2. Dec 2018, at 11:03, Adamant36 <notifications@github.com> wrote:
> 
> Where the difference between a 100 foot versus a 60 foot crane is minimal. In the case of how it should be rendered. Either cranes have importance in a landscape or they don't. The 40 foot difference probably doesn't matter.


exactly, the relevant difference I see is between cranes merely interesting for their function, like this 
https://upload.wikimedia.org/wikipedia/commons/b/bf/Bootskran_Typ_Maat.jpg

and those that have an impact on the landscape.

Adamant36:
Totally. There should be a different tag for them depending.

polarbearing:
> in cities like Hamburg they do have cultural significance, they are part of the “panorama”
Ho ho ho, Berlin is a tourist attraction for its plentifulness of construction sites!

jeisenbe:
The current code in this PR renders cranes at the approximately the same
zoom levels as man_made=chimney

That is, z16 for height > 50m and z17 otherwise.

These are both industrial features that are mainly rendered as orientation
points based on their height and visual prominence.

jeisenbe:
Not many cranes are tourist sites, but a few of the big ones have become icons:
https://en.wikipedia.org/wiki/Samson_and_Goliath_(cranes)

Belfast mentions these cranes on their tourist info website:
https://visitbelfast.com/see-do/partners/samson-goliath-cranes

And this crane, which I showed above, has been tagged as tourism=attraction

Adamant36:
So render all cranes at the same starting zoom level. Except add an exception that ones labeled as tourism=attraction get rendered a level higher. That seems like a more sane less arbitrary thing then height for me and it will also get people to tag the few significant ones as such.

jeisenbe:
Tourism=attraction is less arbitrary than height?
Certainly not less subjective.

dieterdreist:
sent from a phone

> On 2. Dec 2018, at 15:27, Adamant36 <notifications@github.com> wrote:
> 
> we can at least all agree that the tourism tag is how OSM intends "tourist attractions" to be tagged


let’s say, some mappers intend this tag for tourism attractions. It is not undisputed, personally I find it useless as a tag, too many attractions around, so it really depends on YOU what is attracting you. IMHO rendering the attraction tag (without additional tags that say what it is) could be more harmful than good for the map progress. Using it as a qualifier like you propose, could be ok (if the tag is applied in a consistent way around your area).

jeisenbe:
Here are more test renderings, from Jakarta. This huge city is quite close to the equator, so z16 at this level is the same as z17 at 60 degrees north, and is about half-way between z16 and z17 for cities at 40 to 45 degrees. The cranes that I found are all located in the port terminal area.

z17 current master
jakarta-port-z17-master

z17 cranes - after PR
jakarta-port-z17-cranes

z18 cranes - after
jakarta-z18-cranes

z16 (no change with this PR based on current data)
jakarta-port-z16

kocio-pl:
Since this discussion is inconclusive, I will merge it as it is and let's examine it on the real map for some time. I think rendering cranes is a good thing in general and such secondary problems should not eclipse that.

Thanks for both creating (icon + code) and discussing this problem!

Adamant36:
Kocio-pl, although I agree that we shouldn't let minor details get in the way of rendering something, doesn't merging a PR like this risk making people think they can get their way by just stalling on implementing, or just ignoring, a request for something to be changed? That's essentially what happened here. 

There's no reason Jeisenbe couldn't have just implement rendering at z17/z18 and then it if z16 was needed could have been decided later like multiple people requested instead of ignoring it and forcing it to now be the other way around. By merging it anyway, without him listening to feedback, your just incourging him to continue ignoring other peoples wishes. Which he's done on several issues already as it is. Not just this one. Personally, I don't think its a good way to go.

kocio-pl:
I see. My intention is not to help ignore anything. I stated before that both ways are acceptable for me and cranes are not that popular to create immediate problem, so I guess it's safe to change it if we see it does not work he thinks it does. 

From my experience it's more important to allow a bit of fuzziness in the project, including imperfect design, than allow discussing every bit to death to avoid any problem, because first one might be fixed in the future (or accepted eventually), while the second approach can simply paralyze team. I worry to not loose the nice balance between designing/discussing and coding we have currently.

Adamant36:
Ok. I can understand that. I know its not an easy thing to balance everything perfectly. Especially in a project like this. Thanks for the explanation.

jeisenbe:
I would have been happy to change the zoom level to 17/18 if any of the
maintainers requested the change, but I’m working on several other bugs and
issues, so I didn’t want to put more work into this PR than necessary.

If anyone finds any problems related to this PR, I will be happy to make a
new PR to fix it.

kocio-pl:
Thanks, nice to hear that. This is also one of my reasons to not get deeper into this problem - I have some big changes proposals pending at the table, which will affect the style much more than all cranes combined...

Adamant36:
@jeisenbe So if me or someone else starts a new issue about it (or even requests a change to another issue), you'll blow it off because they aren't "maintainers"? Seems like a good way to get people to not to work with you anymore and end up with a bad map. Maintainers might merge the PRs, but they only constitute a small part of the feedback on the and make even less design decisions then the contributors who arent maintainers do. We are relient on feedback from others. Including myself. Personally, I dont priortize feedback based on the "status" the person giving it. Otherwise, its a good way to alienate myself (probably not anymore then writting these types of messages though..But whatever).

PR #3505, opened 2018-11-09
Parking entrance

### LINKED ISSUE #270: Add rendering for amenity=parking_entrance

johsin18:
Resolves #270
Closes #2875

- Add rendering (marker and text for non-private parking entrances).
- project.xml grows from 53109 to 53163 lines due to this patch.
- access null, yes and permissive rendered as opaque, no and private not rendered, the rest rendered transparent
- only nodes are taken into account, as this is the only allowed element type

https://www.openstreetmap.org/#map=18/48.80225/9.01591
Before
image
After
image

https://www.openstreetmap.org/#map=19/48.80289/9.07147
Before
image
After
image


kocio-pl:
Thanks for rewriting and testing this! At the first sight it looks good for me, but could you explain why do you choose z18+ and not z19+?

I have changed the keywords in this PR (resolves/closes) and put the right tickets for them.

johsin18:
I took over z18+ from here:
https://github.com/gravitystorm/openstreetmap-carto/pull/2875/commits/8cd87daf3cf7ab2a7f57f00f784878f07379ad07
Wasn't the discussion about z19+ just for private ones (which my PR omits altogether)?
Most parkings are shown in z18+, so I would prefer that for the parking entrances as well (as they have the same relevancein the end).

kocio-pl:
Thanks, it works as expected. I'm glad that you took over my initial effort, seems like it's something people were really waiting for.

Now we can think if underground parking icon (and related elements) is still needed - see #3506.

Issue #468, opened 2014-04-09
national park boundary does not show up for closed ways


sommerluk:
The way 267307539 is the boundary of a national park in Ivory Coast. Basicly, this parc is formed by islands. The boundary passes water. The boundary currently doesn’t render at all. A bug?

matkoniecz:
It appears to be duplicate of reported later #1141. I am closing this rather than later as in #1141 it is clear from start what real problem is.

PR #3509, opened 2018-11-13
Adding rendering for boundary=protected_area

### LINKED ISSUE #603: Add rendering for boundary=protected_area

kocio-pl:
Follow up to #2830.
Resolves #603.

Since I have found no clear rules of how `boundary=protected_area` relates to national parks and nature reserves, it should be rendered the same. Only `protect_class=1-7,97-99` should be rendered, because they are more or less in the same category as them - see https://wiki.openstreetmap.org/wiki/Tag%3Aboundary%3Dprotected_area#Nature-protected-area (<s>to be implemented</s>). Smaller ones will appear later, so no clutter is expected. 

Example rendering ( https://www.openstreetmap.org/way/199324293 )

dmt4vdms

### DIFF
--- a/project.mml
+++ b/project.mml
@@ -1228,7 +1228,7 @@ Layer:
           FROM planet_osm_polygon
           WHERE (boundary = 'national_park'
                  OR leisure = 'nature_reserve'
-                 OR (boundary = 'protected_area' AND tags->'protect_class' IN ('1','2','3','4','5','6','7','97','98','99')))
+                 OR (boundary = 'protected_area' AND tags->'protect_class' IN ('1','1a','1b','2','3','4','5','6','7','97','98','99')))
             AND building IS NULL
             AND way_area > 1*!pixel_width!::real*!pixel_height!::real
         ) AS national_park_boundaries
@@ -2026,7 +2026,7 @@ Layer:
                                                     'water', 'bay', 'strait') THEN "natural" ELSE NULL END,
               'place_' || CASE WHEN place IN ('island') THEN place ELSE NULL END,
               'boundary_' || CASE WHEN boundary IN ('national_park') 
-                                       OR (boundary = 'protected_area' AND tags->'protect_class' IN ('1','2','3','4','5','6','7','97','98','99'))
+                                       OR (boundary = 'protected_area' AND tags->'protect_class' IN ('1','1a','1b','2','3','4','5','6','7','97','98','99'))
                                        THEN boundary ELSE NULL END,
               'leisure_' || CASE WHEN leisure IN ('nature_reserve') THEN leisure ELSE NULL END
             ) AS feature,
@@ -2038,7 +2038,7 @@ Layer:
               OR "natural" IN ('wood', 'glacier', 'sand', 'scree', 'shingle', 'bare_rock', 'water', 'bay', 'strait')
               OR "place" IN ('island')
               OR boundary IN ('national_park')
-              OR (boundary = 'protected_area' AND tags->'protect_class' IN ('1','2','3','4','5','6','7','97','98','99'))
+              OR (boundary = 'protected_area' AND tags->'protect_class' IN ('1','1a','1b','2','3','4','5','6','7','97','98','99'))
               OR leisure IN ('nature_reserve'))
             AND building IS NULL
             AND name IS NOT NULL
@@ -2105,7 +2105,7 @@ Layer:
                              THEN historic ELSE NULL END,
               'highway_' || CASE WHEN highway IN ('services', 'rest_area', 'bus_stop', 'elevator') THEN highway ELSE NULL END,
               'boundary_' || CASE WHEN boundary IN ('national_park')
-                                       OR (boundary = 'protected_area' AND tags->'protect_class' IN ('1','2','3','4','5','6','7','97','98','99'))
+                                       OR (boundary = 'protected_area' AND tags->'protect_class' IN ('1','1a','1b','2','3','4','5','6','7','97','98','99'))
                                        THEN boundary ELSE NULL END,
               'waterway_' || CASE WHEN waterway IN ('dam', 'dock') THEN waterway ELSE NULL END,
               'tourism_' || CASE WHEN tourism IN ('viewpoint', 'attraction') THEN tourism ELSE NULL END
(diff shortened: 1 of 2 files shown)


Adamant36:
Sounds good to me. This is a much needed addition.

kocio-pl:
I have added the limit to only render natural protected areas, so the first example is rendered, but this one is not:

https://www.openstreetmap.org/way/264800352#map=16/50.5569/17.5915

It might need some code checking, but looks like ready for testing.

rrzefox:
> But to be sure a preview on @rrzefox's server would be great.

deployed but not rerendered yet (this comment will be updated later).

kocio-pl:
Comparison is available here as always, please test and report how does it work:

http://bl.ocks.org/matthijsmelissen/raw/af7a602c222dbf1ff1a2c0d84ed755b7/#8.00/53.501/13.880

kocio-pl:
As far as I could test it, this is working as expected.

daganzdaanda:
Does this only work for relations atm? I can't make sense of the code, but this area ("Naturpark Altmühltal", protect_class=5) is mapped with a single way and does not show up in the demo:
https://www.openstreetmap.org/way/276483981
image

kocio-pl:
Thanks! Could you prepare a code to cover these cases?

jeisenbe:
This is related to issue #1141, which affects boundary=national_park tagged
on closed ways. The problem is that boundary ways are not assumed to be
polygons unless they are part of a relation, if I understand correctly.

A solution should address both types of boundaries if possible. Also see
#468 about the same issue.

daganzdaanda:
> Thanks! Could you prepare a code to cover these cases?

Sorry, I have no idea how to do that. I don't even understand the code of this PR... ;-) And atm I really don't have the time to try and learn to understand this :-(

kocio-pl:
OK, I see. :smile: Still it might be helpful if you try to play with our development environment one day, no big coding skills are required.

matkoniecz:
To answer old question from https://github.com/gravitystorm/openstreetmap-carto/issues/603#issuecomment-58159242

> Seems to overlap with existing established tags. What types of protected areas are there that are not covered by the tags already rendered?

(...)

I also agree with https://github.com/gravitystorm/openstreetmap-carto/issues/603#issuecomment-58262606

> Where there is an adequate existing tagging scheme, we don't particularly want to encourage a new incompatible one, as it makes it harder for all data consumers.

So to make clear I want to review whatever there are objects matched by

> 'protect_class' IN ('1','1a','1b','2','3','4','5','6','7','97','98','99')

that can not be correctly tagged with `boundary=national_park` or `leisure=nature_reserve`.

https://github.com/gravitystorm/openstreetmap-carto/issues/603#issuecomment-172944002 gives 

> I believe there is a consensus in the U.S. community that National Forests should be tagged boundary=protected_area, protect_class=6. It's a bit of a chicken-and-egg situation though because we'd like that to be rendered first.

So it this change in principle is desirable.

project.mml
```diff
@@ -1226,7 +1226,9 @@ Layer:
             boundary,
             way_area/NULLIF(!pixel_width!::real*!pixel_height!::real,0) AS way_pixels
           FROM planet_osm_polygon
-          WHERE (boundary = 'national_park' OR leisure = 'nature_reserve')
+          WHERE (boundary = 'national_park'
+                 OR leisure = 'nature_reserve'
+                 OR (boundary = 'protected_area' AND tags->'protect_class' IN ('1','1a','1b','2','3','4','5','6','7','97','98','99')))
```
matkoniecz: I think it is not desirable to render `boundary=protected_area` + `protect_class=2`. As I understand https://wiki.openstreetmap.org/wiki/Tag:boundary%3Dprotected_area#Nature-protected-area such objects should be always tagged rather as `boundary=national_park`.

Classes `1`, `1a`, `1b`, `3`, `4` look like cases that should be tagged as  `leisure = nature_reserve` with potential adding also `boundary = protected_area` so rendering border based solely on `boundary = protected_area` + `protect_class=*` would encourage undesirable rendering (but I am less sure here).

matkoniecz: I am not familiar with this tagging scheme so let me know if I am misunderstanding something.

kocio-pl: I think nobody is familiar with all 3 tagging schemes, which would be a key to make any subtle connections. I don't remember too much after a year or so, even if I did a lot of research to compare, but my last conclusion was that it's a mess that can't be easily resolved:

https://github.com/gravitystorm/openstreetmap-carto/issues/603#issuecomment-350587768

sylvain-m:
Congratulations for the work done that meets my expectations!
Just a small suggestion: I would find it relevant that the protected area labels be displayed at a higher zoom level.
As for France for example, the labels of the Regional Nature Parks ("Parcs naturels régionaux") cover the text territory, while other important themes are not displayed at these zoom levels (names of regions, large cities, ...).
In my opinion, it would be necessary for the perimeters to be visible from the zoom 8 (as is the case), but for the labels to appear only at the zoom 9.

project.mml
AdamHauner: @matkoniecz: I understand text in wiki, that `protected_area` should tag **all** protected areas including national parks. Look at first sentence of this page:

> boundary=protected_area is used to mark boundaries of protected areas **including national parks**, marine protection areas, heritage sites, wilderness, indigenous and other, similar areas.

kocio-pl:
Please open new ticket for it. This is so complicated and big problem, that it's better to make changes step by step. Rendering is established for years, so we talk about what we want to include at the moment, changing rendering will influence all such areas.

project.mml
kocio-pl: @matkoniecz What do you think about this problem now?

sig-pnrnm:
> 
> 
> Please open new ticket for it. This is so complicated and big problem, that it's better to make changes step by step. Rendering is established for years, so we talk about what we want to include at the moment, changing rendering will influence all such areas.

Done here : https://github.com/gravitystorm/openstreetmap-carto/issues/3569 :wink:

project.mml
matkoniecz: One object may not be tagged at once with `boundary=protected_area` and `boundary=national_park`. By rendering `boundary=protected_area` that should be tagged as `boundary=national_park` we basically allow `boundary=national_park` to be deprecated.

kocio-pl: Oh, I see now. I believe it's not deprecating, it's just description for one of valid (documented and used, I believe) tagging schemes. For me deprecation would be if there will be any reference which scheme should be used and which should not.

Still I feel this is mess that might be resolved at the tagging/documentation level, but now the best we could do now is to render all the "natural" classes. If anything will be more clear in the future, we can update the rendering.

matkoniecz: I will think a bit about this. For now I opened https://josm.openstreetmap.de/ticket/17111#ticket "convert boundary=protected_area + protect_class=2 into human readable equivalent boundary=national_park" (if someone thinks that boundary=national_park is not strictly preferable it would be a good idea to comment there).

kocio-pl:
@matkoniecz Since I have no idea what you like to fix or try, so I plan to merge it soon as it is and do the fine tuning later.

Issue #1017, opened 2014-10-06
wind turbines: do not show minor ones at zooms 15/16/17


matkoniecz:
PR #248 converted into a ticket

@danstowell suggests that we don't want to render all wind-turbines at zoom 15/16/17, only those marked `man_made=tower`.

Wind-turbines are sometimes significant landmarks (usually standlone towers) and sometimes not (small building-mounted turbines). The carto style currently errs on the side of eagerly showing wind turbines, which for some examples of tiny turbines in the city is undue prominence. Here's an example in Birmingham:

![birmingham-turbine](https://f.cloud.github.com/assets/202965/1416106/a21698da-3f37-11e3-9d67-2846a41701c3.png)
http://www.openstreetmap.org/#map=15/52.4836/-1.8881

Here's an example in London:
![highbury-turbine](https://f.cloud.github.com/assets/202965/1416107/a807465e-3f37-11e3-98d8-6aadb48664b8.png)
http://www.openstreetmap.org/#map=15/51.5439/-0.1037

I asked on the "tagging" mailing list if anyone had suggestions. There wasn't much controversy, but not much definite direction. One suggestion was that `landmark=yes` could disambiguate. Seemed OK to me, but searching actual use, I don't see this actually used for wind turbines.

However, in Europe there are 427 wind turbines also marked with `man_made=*`, of which 133 are `man_made=tower`. (Other values include things not encouraged by the wiki such as `man_made=wind_turbine`, and things not relevant to landmarkness such as `man_made=power`.)

@gravitystorm 

So to follow-up our in-person discussion:
- Yep, I think it's good to differentiate, if possible. The rendering of wind turbines on this style is as much (if not more) about their giant landmarkness than their electricity supply interestingness.
- I don't like "landmark=yes", since it's completely subjective
- I worry that since they are all on towers of one height or another, then that doesn't distinguish

Perhaps a height, or a tower height tag?

@mboeringa

> This pull-request suggests that we don't want to render all wind-turbines at zoom 15/16/17, only those marked man_made=tower.
> ...
> I worry that since they are all on towers of one height or another, then that doesn't distinguish

I think the better solution to this, might be to consider the power output in kW of a turbine. Small turbines of less than say 50-100 kW, could be left out of the rendering at certain zoom levels, or drawn with a smaller icon. Larger turbines than these, are less likely to be found in dense city quarters, and if they do, like in industrial zones, they should probably be rendered.

[...]

PR #3515, opened 2018-11-16
Render Wind Turbines names, Change icon to gray, Do not render before z19 if location is rooftop

### LINKED ISSUE #2734: Names for wind turbines
Named Windturbines currently have no rendered name.
Example: https://www.openstreetmap.org/node/5003725018

Perhaps this is a problem with [icon](https://github.com/gravitystorm/openstreetmap-carto/blob/5096a9e79023c4006ae5a5813b0cba9172d89050/amenity-points.mss#L945) collision or an area [problem](https://github.com/gravitystorm/openstreetmap-carto/blob/5096a9e79023c4006ae5a5813b0cba9172d89050/amenity-points.mss#L1916) with nodes.

jeisenbe:
Fixes #2734
Also related to #1017

- Start rendering names of wind turbines (power=generator, generator:source=wind)
- Change wind turbine icon color to man-made-gray
- Render wind turbines only at z19 if location=rooftop or =roof
- Remove rendering for deprecated tagging power=generator with power_source=wind

**Explanation:**
Wind turbine name labels do not currently render, because the names for power=generator are done in landcover style without any text-dy. This PR adds `text-dy: 10` for wind turbines, and the names will be rendered the same as other man_made features, starting at z17 (unless location=rooftop or roof)

The icon color is changed to match other man_made features, because it is currently black. This color is otherwise used only for religious feature icons.

Wind turbines on building rooftops are always much smaller than the large stand-alone towers common in rural areas, so they will now be rendered starting at z19. This will prevent them from blocking building names and the icons for amenities or shops in buildings at z17 and z18. The name label will also be rendered only at z19 in this case.

The tag power_source=wind is deprecated, and is now only used 200 times, vs 200,000 times for generator:source=wind, therefore it will be removed from rendering.

**Before**
Blackfriars road z18 before
https://www.openstreetmap.org/#map=18/51.504541/-0.10496
blackfriars-z18

Islington Council building 
https://www.openstreetmap.org/#map=18/51.54401/-0.10262 
z18 before
islington-z18-before
z19 before
islington-wind-19

Strata Building 
https://www.openstreetmap.org/#map=18/51.492857/-0.0995045
z18 before
strata-z18-before
z19 before
chase-wind-19

Prototype named wind turbine, before
https://www.openstreetmap.org/#map=17/53.5056716/8.5790005

Test rendering z19 before
z19-test-before

Test z18 before
z18-test-before

**After**
Blackfriars Road, London z18 (*Name added for test) - still renders because no location tag

AD 8-180 Prototype wind turbine, z17

Islington Council building, z18 - turbine hidden
islington-after-z18
z19 after
islington-wind-gray

Strata building, z18 - hidden
strata-after-z18
z19 after
chase-wind-grayicons

Prototype after
z17-prototyp-after

Test rendering z19
- Named rooftop wind turbines show with label
z19-test-after

Test z18
-rooftop wind turbines hidden, but others render with label.
-Power plants and other power generator labels still render properly as a node or polygon
z18-test-after

### DIFF
--- a/amenity-points.mss
+++ b/amenity-points.mss
@@ -1473,8 +1473,10 @@
   }
 
   [feature = 'power_generator']['generator:source' = 'wind'],
-  [feature = 'power_generator'][power_source = 'wind'] {
-    [zoom >= 15] {
+  [feature = 'power_generator']['generator:method' = 'wind_turbine'] {
+    [zoom >= 15][location != 'rooftop'][location != 'roof'],
+    [zoom >= 15][location = null],
+    [zoom >= 19] {
       marker-file: url('symbols/power_wind.svg');
       marker-placement: interior;
       marker-fill: black;
--- a/amenity-points.mss
+++ b/amenity-points.mss
@@ -1479,7 +1479,7 @@
     [zoom >= 19] {
       marker-file: url('symbols/power_wind.svg');
       marker-placement: interior;
-      marker-fill: black;
+      marker-fill: @man-made-icon;
       marker-clip: false;
     }
   }
--- a/amenity-points.mss
+++ b/amenity-points.mss
@@ -1472,8 +1472,7 @@
     marker-clip: false;
   }
 
-  [feature = 'power_generator']['generator:source' = 'wind'],
-  [feature = 'power_generator']['generator:method' = 'wind_turbine'] {
+  [feature = 'power_generator']['generator:source' = 'wind'] {
     [zoom >= 15][location != 'rooftop'][location != 'roof'],
     [zoom >= 15][location = null],
     [zoom >= 19] {
@@ -1908,6 +1907,9 @@
   }
 
   [feature = 'man_made_cross'][zoom >= 17],
+  [feature = 'power_generator'][location != 'rooftop'][location != 'roof'][zoom >= 17],
+  [feature = 'power_generator'][location = null][zoom >= 17],
+  [feature = 'power_generator'][zoom >= 19],
   [feature = 'historic_wayside_cross'][zoom >= 17],
   [feature = 'historic_wayside_shrine'][zoom >= 17],
   [feature = 'historic_city_gate'][zoom >= 17],
@@ -1932,6 +1934,7 @@
     [feature = 'historic_wayside_cross'] {
       text-dy: 6;
     }
+    [feature = 'power_generator'],
     [feature = 'historic_city_gate'],
     [feature = 'man_made_mast'],
     [feature = 'man_made_tower'],
@@ -2824,7 +2827,7 @@
 
   [feature = 'power_plant'][is_building = 'no'][zoom >= 10],
   [feature = 'power_station'][is_building = 'no'][zoom >= 10],
-  [feature = 'power_generator'][is_building = 'no'][zoom >= 10],
+  [feature = 'power_generator'][is_building = 'no']["generator:source" != 'wind'][zoom >= 10],
   [feature = 'power_sub_station'][is_building = 'no'][zoom >= 13],
   [feature = 'power_substation'][is_building = 'no'][zoom >= 13]{
     [way_pixels > 3000],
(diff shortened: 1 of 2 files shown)


Adamant36:
Does a "!=" clause not include a "null"? You'd think it would since "nothing" isn't equal to a roof, or if it was blank it just wouldnt use it as a rendering variable. I only ask because im still learning the code and think its interesting. Not because I think your wrong by doing that way or anything.

Adamant36:
Id say most of those aren't valid names. Id also imagine most wind turbines don't have legitimate names. Its not a problem specific to wind turbines though. For instance there's a bunch of parks named "park" out there. That being said, the best way to improve the data is to render it so people can see that its wrong and fix it. So I think its still worth rendering the names. Even if most of turn out to be junk.

jeisenbe:
**Re: name rendering:** power generator names should currently be rendered, but because wind generators have an icon the name rendering is blocked by the icon at the same location. This PR will make wind generator names render as intended, just like they currently do for solar panels, small hydroelectric, and any other type of power generator. 

I agree that most individual power generators do not have real names (usually the entire power plant is named, if there is a name), but hopefully the rendering will encourage mappers to remove incorrect names like "wind turbine" from the database.

**Icon change proposal:** both the old and new icon look fine to me, so I do not have a strong preference either way. I wish the 3 blades on the new icon were arranged evenly around the circle, but I think this was not possible due to the square shape of the canvas?

Test area with current icon z17:
windtest1-currenticon

Test with new icon proposal:
windtest1-newicon

London Blackfriars Road rooftop wind turbines current z18 (NOTE: I have added the "name" as "wind turbine"; this is not in the public database, fortunately)

Proposed icon:
wind-blackfriars-new-icon

German prototype current z17:
wind-protoptype-currenticon

Proposed icon:
wind-prototype-newicon

Dover wind turbine current z15:
wind-dover-z15-current

Proposed:
wind-dover-new-icon

jeisenbe:
Any opinions on the new or old wind turbine icon?

kocio-pl:
I prefer the new one, since it's more visible that it has 3 symmetric arms, old one is less clear with shape.

Adamant36:
I say new one also. I like that it has a thing in the middle of the blades. It looks clearly that way.

jeisenbe:
OK, I have pushed a commit with the new icon. I believe this is ready to merge

kocio-pl:
Thanks, it looks OK in general, however `symbols/power_wind.svg` file should be removed and new icon file has too much rights (755 instead of 644).

There are also few more icons with 755 mode - it'd be good to make some cleaning, but in a separate PR.

kocio-pl:
Please change file rights too for `symbols/generator_wind.svg`, since it belongs to this change, unlike other icons.

jeisenbe:
I can make a PR to fix permissions on all the other icons, if it is needed

kocio-pl:
I just wanted to ask you about it... :smile:

jengelh:
> Question - what are valid wind turbine names?

Whatever the owner/city administration decided it to be. Generally speaking, the name can be found on the owner's website (if such exists - not always a given), as a label on a transformer substation nearby (most reliable here in GER), or in news media or city council documents. A common name pattern for Germany is "WKA <locality> <number>", or "WKA <number>", implying the locality. If your UK wind turbines are called "Wind Turbine #1", I guess that's that. If it were just "Wind Turbine" though, I would be concerned. Names should have a disambiguating function, and "Wind Turbine" does not do that justice.

> Because I've looked at the ones with names in the UK, and seen the following:
>     * wind farm names
>     * manufacturer names/codes (which are also tagged under manufacturer/manufacturer:type)
>     * a reference number (e.g. A142 or 56)
>     * Wind | Turbine | #1

All are valid (and different things). That "reference number" btw is generally the manufacturer's serial number.

HolgerJeromin:
> A common name pattern for Germany is "WKA {locality} {number}", or "WKA {number}", implying the locality.

I would move this information to ref tag

jengelh:
> > A common name pattern for Germany is "WKA {locality} {number}", or "WKA {number}", implying the locality.
> 
> I would move this information to ref tag

That would be a strange thing to do, similar to moving "Umspannwerk {locality}" from name to ref.

PR #3521, opened 2018-11-22
Add aboriginal areas

### LINKED ISSUE #3520: Add rendering for aboriginal areas
Aboriginal areas (also called [Indigenous Protected Areas](https://en.wikipedia.org/wiki/Indigenous_Protected_Area), Native Reserves, First Nation Reserves, [Indian Reserves](https://en.wikipedia.org/wiki/Indian_reserve), [Indian Reservations](https://en.wikipedia.org/wiki/Indian_reservation), etc., depending on the country) are important features that belong on the map.

There had been some discussion in #603  about protected areas in general, but that discussion seems to have narrowed to discuss only natural protected areas, not cultural ones. So if #3509 proposes closing #603, we need to have a new issue to discuss aboriginal protected areas.

Setting aside the debate about whether administrative boundaries belong in OSM at all (the ship has sailed on that one), it seems obvious to me that aboriginal areas are at least as important as many other boundaries we currently show, such as municipal boundaries, or even state/province boundaries. However, I also agree that they should not be overly prominent either (some have proposed that they be mapped as `admin_level=2` which seems extreme). In many cases, you can cross in or out of an aboriginal area without knowing it or seeing other evidence on the ground. But in other cases, the boundaries are clearly marked on the ground, and sometimes even travel permits are required to enter these areas. Often there are different laws in effect within aboriginal areas, which seems like an important thing for a map user to know about.

Within OSM there does not appear to be a consensus about whether the tagging should be [`boundary=aboriginal_lands`](https://wiki.openstreetmap.org/wiki/Tag:boundary%3Daboriginal_lands) or whether we should use [`boundary=protected_area`](https://wiki.openstreetmap.org/wiki/Tag:boundary%3Dprotected_area) along with `protect_class=24`. Both are in relatively wide use, and I propose rendering both options until a tagging consensus has emerged. `boundary=aboriginal_lands` is quite popular in Canada and parts of the US, while `boundary=protected_area` + `protect_class=24` is popular in South America and Australia. It's also worth noting that mappers have been doing a lot of tagging for the renderer, for example using `leisure=nature_reserve` incorrectly to get these indigenous areas to show up on the map, especially in Brazil.

Overpass query for `boundary=aboriginal_lands`: http://overpass-turbo.eu/s/DV4
screen shot 2018-11-22 at 22 nov 11 18 51

Overpass query for `protect_class=24`: http://overpass-turbo.eu/s/DV5
screen shot 2018-11-22 at 22 nov 11 14 31

taghistory

These areas vary widely in size: some Indian Reservations in the United States are larger in area than many European countries, but in other cases, such as the [Nooksack Reservation](https://en.wikipedia.org/wiki/Nooksack_people) near where I live, they can be as small as one acre (0.4 hectares). For this reason, we need a cartographic style that varies appropriately according to zoom level, and filters out smaller reservations at low zoom levels.

I propose adapting the current `leisure=nature_reserve` styling, but using a brown color instead of green. Nature reserves have a nice subtle fill color at low zoom levels, but when you zoom in, the fill goes away and they only show a light outline. This level of detail seems appropriate for aboriginal areas as well. The aboriginal areas start showing up at z8, which is the same time we first see national parks and military areas. As with those other features, the aboriginal areas are filtered by area, so only large ones show up. At z10, there is no more polygon fill, just an outline.

Here's southern Arizona at z8, before the change. Note the large empty spaces, some of which are actually occupied by large native reservations.
screen shot 2018-11-22 at 22 nov 9 16 32

Here's the same view, with the change applied.
screen shot 2018-11-22 at 22 nov 10 20 28

Here it is with a different brown:
screen shot 2018-11-22 at 22 nov 10 25 33

In the first screenshot I started with the html "brown" color #835c3b (hsl(28, 38%, 37%)), but then in the second screenshot I adjusted the hue more in the direction of yellow away from red, to avoid too much similarity with the existing red military areas. Currently I'm using #82643a (hsl(35, 38%, 37%))

A few more samples:

With #835c3b (reddish brown):
screen shot 2018-10-11 at 11 oct 5 17 22

With #82643a (orangey brown):
screen shot 2018-11-22 at 22 nov 11 04 23

screen shot 2018-11-22 at 22 nov 11 08 00
screen shot 2018-11-22 at 22 nov 11 09 40
screen shot 2018-11-22 at 22 nov 10 35 28

Here are some cases in Canada where aboriginal areas were also tagged with a `type=boundary` relation (perhaps tagging for the renderer): https://www.openstreetmap.org/relation/2227709 and https://www.openstreetmap.org/relation/2092741. The interaction of the dashed purple boundary line with the solid brown aboriginal_area line is a bit weird, but not terrible IMO. Once we have correct aboriginal rendering, we can remove these unnecessary boundary tags.

screen shot 2018-11-22 at 22 nov 11 36 45
screen shot 2018-11-22 at 22 nov 11 44 04

I've tested this change for Arizona and Washington State in the US, and British Columbia in Canada. I would have tested against Brazil, but most of the areas tagged with `protect_class=24` there are incorrectly tagged as `leisure=nature_reserve`, so they already show up on the current rendering, just in green, not in brown.

Note: I based my code as closely as possible on #3509, which adds more natural protected areas to the map. If that PR gets merged first, I realize that this PR #3521 will probably need to be rebased to avoid some conflicts.

almccon:
Fixes #3520 

- Add rendering for `boundary=aboriginal_area` and for `boundary=protected_area` + `protect_class=24`.

Before
screen shot 2018-11-22 at 22 nov 9 16 32

After
screen shot 2018-11-22 at 22 nov 10 25 33

More samples:
screen shot 2018-11-22 at 22 nov 11 09 40
screen shot 2018-11-22 at 22 nov 11 08 00
screen shot 2018-11-22 at 22 nov 11 04 23
screen shot 2018-11-22 at 22 nov 10 35 28
screen shot 2018-11-22 at 22 nov 11 36 45


kocio-pl:
As I have written in the issue ticket, it's not easy to find good color - this rendering might be too close to the zoo. So maybe we need some dark orange border; of course the colors can be switched (brown-orange for aboriginal area and orange for zoo) if that would work better. That needs testing. Also wiki documentation is needed, following discussion on the Tagging list to make sure if we really need 2 different schemes (maybe we do).

Another, smaller issue would be to rename data layer into something like `protected-areas-boundaries` instead of creating another similar layer called `aboriginal-lands-boundaries`.

jeisenbe:
@almccon, if the orange doesn't work, another option would be to switch tourism boundaries (eg zoos, theme parks) to a shade of violet or purple, because we are planning to change administrative boundaries to gray, thus making purple available for tourism features.

sommerluk:
If violet becomes available, it should not be a minor feature like this one to occupy it again (blocking a hole coulor).

jeisenbe:
Lukas Sommer wrote:
> If violet becomes available, it should > not be a minor feature like this
one to > occupy it again

Zoos an amusement parks are fairly minor, but we could use the same border
for many other tourist areas and accommodations: e.g. resorts, campsites,
caravan sites, etc.

Violet is a rather strong color an uncommon in nature, so I think it should
be used for less common but interesting features; tourism and
transportation are good options.

But if orange works, that could also be a good option for tourism.

Adamant36:
Isn't violet/purple being used for shops currently? I think it works good there. The new gastronomy icon is sort of orangish already also. Maybe the dark blue color that's currently being used for railway stations would work. Since its the only thing that I know of with the color. 

There was a discussion somewhere around here about changing that and a few other colors around that never went anywhere. Maybe it would be worth revisiting and putting this stuff/that stuff under one umbrella. Since they are all sort of related. As far as zoos, amusement parks etc goes. I think it would be the wrong color for this issue though. 

Orange might actually work here despite its food association because I doubt people think aboriginal areas are related to restaurants. Plus, its kind of an earth tone. Which sorta fits.

matkoniecz:
@almccon Would it be OK to mark this PR as assigned to you for now?

almccon:
@matkoniecz Yes, please go ahead and assign it to me, thanks. I will try some alternative color treatments to avoid the zoo conflict. I just haven't had time to do it yet.

kocio-pl:
We have now `boundary=aboriginal_lands` documented:

https://wiki.openstreetmap.org/wiki/Tag:boundary%3Daboriginal_lands

and i agree that we should now render both schemes.

So, we just wait for solving rendering problems.

jeisenbe:
@almccon, here are some options for changing the tourism color (currently brown, used for tourism=zoo and tourism=theme_park). Tourism=attraction uses #660033, but I would like to change this to be the same.

Here's the current rendering of this test area that I created in JOSM (which shows province and national borders as well as several landuse colors that may be relevant):
**Current style**
z15
z15-theme-park-master
z16
z16-theme-park-master
z17
z17-theme-park-master

**#660033**
- this uses the current tourism=attraction color for zoos and theme parks as well
- It may work even without changing the administrative boundary colors
z15
z15-theme-park-660033
z16
z16-theme-park-660033
z17
z17-theme-park-660033

**Orange**
- This test is with #C77400,
z15
z15-theme-park-orange
z16
z16-theme-park-orange
z17
z17-theme-park-orange

**Dark violet** #864784
- Too similar to current administrative boundary color; so this might need to be changed first.
z15
z15-theme-park-darkviolet
z16
z16-theme-park-darkviolet
z17
z17-theme-park-darkviolet

kocio-pl:
`#660033` for attractions looks great for me. But I think this should be discussed in separate PR.

kocio-pl:
@almccon Quoting myself:

> Another, smaller issue would be to rename data layer into something like `protected-areas-boundaries` instead of creating another similar layer called `aboriginal-lands-boundaries`.

More layers create additional performance penalty and all these objects are equal in rendering (none of them should have rendering priority over the other), so it does not make sense to create additional data layer.

This is probably the last thing to be done here, the rest is just waiting for #3582 to be merged.

kocio-pl:
@almccon Ping?

Now the #3582 is merged, so merging this depends only on your action.

jeisenbe:
@almccon - do you have time to update this PR to fix the merge conflicts and per the suggestion in https://github.com/gravitystorm/openstreetmap-carto/pull/3521#issuecomment-449129496? 

If not, I could finish it for you (by submitting a new PR)

jeisenbe:
@almccon - do you have time to update this PR to fix the merge conflicts and per the suggestion in https://github.com/gravitystorm/openstreetmap-carto/pull/3521#issuecomment-449129496? 

If not, I could finish it for you (by submitting a new PR)

kocio-pl:
@jeisenbe I believe you can make your ow PR now to fix this.

almccon:
Hi all, sorry for the delay. I fixed the merge conflict, and I'll try to respond to the other comments tomorrow. But if you want to jump on any of it, be my guest @jeisenbe!

jeisenbe:
Go ahead, I think you’ve got it.

almccon:
Okay, @kocio-pl, I combined the two layers into a single layer as per the suggestion in https://github.com/gravitystorm/openstreetmap-carto/pull/3521#issuecomment-449129496

In this branch, the aboriginal areas are #82643a, and thanks to #3582 zoos and other tourism are #660033. The aboriginal brown is lighter than the original tourism brown #734a08, which is still used for amenity points.

screen shot 2019-01-27 at 27 jan 4 33 19
screen shot 2019-01-27 at 27 jan 4 31 44

This feels like a good outcome to me. I don't think there are any other outstanding issues. Did I miss anything?

kocio-pl:
Did you actually test the code (for both cases) after the changes? It is a common thing to do after each change to avoid errors.

I get SQL error when trying to render anything with this code:

```Postgis Plugin: ERROR:  column "boundary" does not exist```

If you need some help, please talk with @jeisenbe probably.

**Comment by almccon** [comment]
I did test the code, and I did see that error happen initially after I made the code change, but then when I restarted docker I thought I stopped seeing the error. Let me check again.

**Comment by almccon** [comment]
You're right, that error is showing up for me at z13+, due to the `protected-areas-text` layer not including the `boundary` and `class` column. My fault. Should be fixed now.

**Comment by kocio-pl** [comment]
Thanks, now it works, but the text label color is wrong here for example:

https://www.openstreetmap.org/relation/2740289#map=10/9.9377/-72.9568

screenshot_2019-01-29 openstreetmap carto kosmtik

**Comment by kocio-pl** [comment]
Looks like the problem exists only for `boundary=protected_area + protect_class=24` but not for `boundary=aboriginal_lands`:

https://www.openstreetmap.org/way/383485661#map=14/20.6472/-103.2405

screenshot_2019-01-29 openstreetmap carto kosmtik 1

File: project.mml [hunk 1]
```diff
@@ -2063,7 +2065,9 @@ Layer:
               'natural_' || CASE WHEN "natural" IN ('wood', 'glacier', 'sand', 'scree', 'shingle', 'bare_rock',
                                                     'water', 'bay', 'strait', 'cape') THEN "natural" ELSE NULL END,
               'place_' || CASE WHEN place IN ('island') THEN place ELSE NULL END,
-              'boundary_' || CASE WHEN boundary IN ('national_park')
+
```
kocio-pl: It looks like this line is unwanted.

amenity-points.mss
```diff
@@ -20,8 +20,10 @@
 @barrier-icon: #3f3f3f;
 @landform-color: #d08f55;
 @leisure-green: darken(@park, 60%);
+@aboriginal: #82643a;
 @religious-icon: #000000;
 
+
```
kocio-pl: This line should be removed probably.

amenity-points.mss
```diff
@@ -2311,6 +2314,10 @@
       [feature = 'landuse_military'] {
         text-fill: darken(@military, 40%);
       }
+      [feature = 'boundary_aboriginal_lands'],
+      [feature = 'boundary_protected_area'] {
```
kocio-pl: it's defined few lines later in a different (proper) color.

project.mml
```diff
@@ -1241,14 +1241,16 @@ Layer:
             way,
             name,
             boundary,
+            tags->'protect_class' AS class,
```
kocio-pl: I would leave protect_class as name because class sounds way too generic.

admin.mss
```diff
@@ -220,10 +220,14 @@ overlapping borders correctly.
   text-dy: -10;
 }
 
-#nature-reserve-text[zoom >= 13][way_pixels > 192000] {
+#protected-areas-text[zoom >= 13][way_pixels > 192000] {
   text-name: "[name]";
   text-face-name: @book-fonts;
   text-fill: green;
+  [boundary='aboriginal_lands'],
+  [class='24'] {
```
kocio-pl: I guess aboriginal_lands already includes 24 case? Anyway class=24 alone sounds hard to understand.

kocio-pl:
It works now, it just needs some code cleaning (I have not repeated it for every occurrence).

project.mml
almccon: Removed

amenity-points.mss
almccon: removed

amenity-points.mss
almccon: You're right, only `boundary_aboriginal_lands` should be used here. Fixed.

project.mml
almccon: Good recommendation. Changed.

admin.mss
almccon: `boundary=aboriginal_lands` doesn't need `24`, but you're right that it's not very clear to only use `[class=24]` to capture the other possible tagging approach. I made this line more verbose, so the second selector is now `[boundary='protected_area'][protect_class='24']`

almccon:
Thanks for the feedback, all of those code cleanup recommendations are good, and I think I've applied them all.

kocio-pl:
Now it looks OK for me. Thanks for all this work!

PR #3528, opened 2018-11-25
Add rendering for private POI

### LINKED ISSUE #1012: render more private POI icons lighter

jragusa:
Fixes #1012

- reorder some features
- add opacity for private POI

`amenity=bbq`
https://www.openstreetmap.org/node/1062462954
private_bbq

`amenity=bench`
https://www.openstreetmap.org/node/4570165702
private_bench

`amenity=bicycle_repair_station`
https://www.openstreetmap.org/way/266000780
private_bicycle_repair

`amenity=charging_station`
https://www.openstreetmap.org/node/5996039033
private_charging_station

`amenity=drinking_water`
https://www.openstreetmap.org/node/5275971271
private_drinking_water

`amenity=shelter`
https://www.openstreetmap.org/way/426143872
private_shelter

`amenity=shower` and `amenity=toilets`
https://www.openstreetmap.org/way/343862313
private_shower

`amenity=waste_basket`
https://www.openstreetmap.org/node/4078433504
private_waste_basket

`amenity=waste_disposal` (not included in the PR)
https://www.openstreetmap.org/way/400743572
private_waste_disposal2

`leisure=firepit`
https://www.openstreetmap.org/node/4358615257
private_firepit

`leisure=fitness_station`
https://www.openstreetmap.org/way/601801826
private_fitness_station

`leisure=picnic_table`
https://www.openstreetmap.org/node/4456465161
private_picnic_table

`tourism=picnic_site`
https://www.openstreetmap.org/way/305800094
private_picnic_site

`tourism=wilderness_hut`
https://www.openstreetmap.org/way/137043135
private_wilderness_hut

### DIFF
--- a/amenity-points.mss
+++ b/amenity-points.mss
@@ -1593,9 +1593,6 @@
     marker-file: url('symbols/amenity/waste_disposal.svg');
     marker-fill: @amenity-brown;
     marker-placement: interior;
-    [access != ''][access != 'permissive'][access != 'yes'] {
-      marker-opacity: 0.33;
-    }
   }
 }
 
@@ -1711,9 +1708,6 @@
     marker-file: url('symbols/amenity/waste_disposal.svg');
     marker-fill: @amenity-brown;
     marker-placement: interior;
-    [access != ''][access != 'permissive'][access != 'yes'] {
-      marker-opacity: 0.33;
-    }
   }
 }
 
--- a/amenity-points.mss
+++ b/amenity-points.mss
@@ -1856,15 +1856,38 @@
 
   [feature = 'amenity_vehicle_inspection'][zoom >= 17],
   [feature = 'amenity_car_wash'][zoom >= 17],
-  [feature = 'amenity_drinking_water'][zoom >= 17],
   [feature = 'amenity_internet_cafe'][zoom >= 17],
-  [feature = 'tourism_picnic_site'][zoom >= 17],
   [feature = 'leisure_bowling_alley'][zoom >= 17],
   [feature = 'leisure_beach_resort'][zoom >= 17],
   [feature = 'leisure_bird_hide'][zoom >= 17],
   [feature = 'leisure_amusement_arcade'][zoom >= 17],
   [feature = 'leisure_outdoor_seating'][zoom >= 19],
-  [feature = 'leisure_fishing'][zoom >= 17],
+  [feature = 'leisure_fishing'][zoom >= 17] {
+    text-name: "[name]";
+    text-size: @standard-font-size;
+    text-wrap-width: @standard-wrap-width;
+    text-line-spacing: @standard-line-spacing-size;
+    text-fill: @amenity-brown;
+    [feature = 'leisure_outdoor_seating'],
+    [feature = 'leisure_fishing'],
+    [feature = 'leisure_bowling_alley'],
+    [feature = 'leisure_bird_hide'],
+    [feature = 'leisure_amusement_arcade'],
+    [feature = 'leisure_beach_resort'] {
+      text-fill: @leisure-green;
+    }
+    text-dy: 10;
+    text-face-name: @standard-font;
+    text-halo-radius: @standard-halo-radius;
+    text-halo-fill: @standard-halo-fill;
+    text-placement: interior;
+  }
+
+  [feature = 'amenity_bbq'][zoom >= 17],
+  [feature = 'amenity_bicycle_repair_station'][zoom >= 19],
+  [feature = 'amenity_drinking_water'][zoom >= 17],
+  [feature = 'amenity_shower'][zoom >= 18],
+  [feature = 'tourism_picnic_site'][zoom >= 17],
   [feature = 'leisure_picnic_table'][zoom >= 17] {
     text-name: "[name]";
     text-size: @standard-font-size;
@@ -1872,12 +1895,6 @@
     text-line-spacing: @standard-line-spacing-size;
     text-fill: @amenity-brown;
     [feature = 'tourism_picnic_site'],
-    [feature = 'leisure_outdoor_seating'],
-    [feature = 'leisure_fishing'],
-    [feature = 'leisure_bowling_alley'],
-    [feature = 'leisure_bird_hide'],
-    [feature = 'leisure_amusement_arcade'],
-    [feature = 'leisure_beach_resort'] {
       text-fill: @leisure-green;
     }
     text-dy: 10;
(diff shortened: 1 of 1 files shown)


Adamant36:
@jragusa, instead of taking a screen shot of a test rendering you can click on "export" on the right side, click "show export extent on map" to on, click "current view," and then click export. That way you wont have the things on the left side in the picture. Plus, if you zoom out a little there's an adjustable bounding box so the area can be more precise. Also, exporting rendering can be a little more accurate then just doing a screen shot because Chrome can screw with colors sometimes. Whereas, exporting takes the screen capture directly from renderer instead of it being filtered through the browser. At least that's what I read.

jragusa:
@Adamant36 it's intentional to display the zoom level. I know this function but I don't know why, I use the embedded screen shot function of Firefox. I will test next time.

Adamant36:
Oh. I thought that might be it. Usually I'll include the zoom level in the name of the file and then reference it later in the image title when I post it.

kocio-pl:
> amenity=waste_disposal (not displayed previously)

@matkoniecz IIRC this was your pet issue (see https://github.com/gravitystorm/openstreetmap-carto/pull/2974#issuecomment-360704790)?

matkoniecz:
Yes, I still think that private amenity=waste_disposal should not be displayed before z20/z21, even z19 is too early.

sommerluk:
Agree with @matkoniecz that private waste disposals should be on much higher zoom levels.

sommerluk:
I’m not convinced that the transparency is an intuitive rendering for private features. And it’s not really good for legibility.

Adamant36:
Yeah, it might be hard for people with vision problems to see. It seems to look lighter on things besides parking for some reason also. Do you have an alternative?

sommerluk:
No, I haven’t an alternative. But I think it’s also not so important to render this difference, so I prefer a clear map.

Our map style does not target people with vision problems, so it doesn’t matter if they can’t recognize it. (If we would target people with vision problems — and there are quite a few different vision problems —, the remaining design and colour space would be to small for our rich map style.) But I think in this case it is even difficult to recognize for people with normal vision. Especially when rendered above boundaries between different colour surfaces.

kocio-pl:
Maybe we just need less opacity. It was designed for parkings, but maybe other objects should be sightly less muted.

sommerluk:
I do not think that would help. Either it’s still difficult to read or it is almost indistinguishable from the normal rendering.

kocio-pl:
I believe there is rather a sliding scale, not simple "1 or 0.33" opacity switch. But we might be wrong, it should be tested of course.

jragusa:
2 examples from above with `marker-opacity: 0.5;` (`leisure=playground` and `amenity=parking` are not concerned)
private_shower_05

private_waste_basket_05

kocio-pl:
@sommerluk What do you think of these samples?

amenity-points.mss
```diff
@@ -1557,12 +1590,11 @@
...
-    [access = 'permissive'],
-    [access = 'yes'] {
-      marker-file: url('symbols/amenity/waste_disposal.svg');
-      marker-fill: @amenity-brown;
-      marker-placement: interior;
+    marker-file: url('symbols/amenity/waste_disposal.svg');
+    marker-fill: @amenity-brown;
+    marker-placement: interior;
+    [access != ''][access != 'permissive'][access != 'yes'] {
+      marker-opacity: 0.33;
```
matkoniecz: rendering private waste disposal seems bizarre to me

jragusa: I removed the rendering of private `waste_disposal`

sommerluk:
> @sommerluk What do you think of these samples?

I’m still not convinced that the transparency is an intuitive rendering for private features.

Adamant36:
@sommerluk, can you suggest something else?

HolgerJeromin:
> I’m still not convinced that the transparency is an intuitive rendering for private features.

We have a lighter color for parking, recycling, playgrounds and normal highways.

matkoniecz:
While I am not going to claim that "less visible" has obvious meaning "private", I think it is fairly intuitive.

sommerluk:
Okay.

amenity-points.mss
```diff
@@ -1557,13 +1590,9 @@
 
   // waste_disposal tagging on ways - tagging on nodes is defined later
   [feature = 'amenity_waste_disposal'][zoom >= 19] {
-    [access = null],
-    [access = 'permissive'],
-    [access = 'yes'] {
```
kocio-pl: When you skip original value checking, waste disposals are rendered always. I think it should be just restored to the original state, so private waste disposals will not be rendered at all.

jragusa: oh yes I completely forgot the special rules for `waste_disposal`. Corrected for both nodes and ways.

kocio-pl:
Thanks for the fix, everything else was OK.

pnorman:
> Also, exporting rendering can be a little more accurate then just doing a screen shot because Chrome can screw with colors sometimes. Whereas, exporting takes the screen capture directly from renderer instead of it being filtered through the browser first. At least that's what I read.

No, do screenshots. It's very tricky to guarantee that export will get the scales right. https://github.com/gravitystorm/openstreetmap-carto/blob/master/CONTRIBUTING.md#reporting-issues talks about using screenshots for reporting issues, and the same is true for PRs.

amenity-points.mss
```diff
@@ -2499,6 +2560,10 @@
     [feature = 'amenity_shelter'] {
       text-fill: @amenity-brown;
     }
+    [access != ''][access != 'permissive'][access != 'yes'] {
+      text-opacity: 0.33;
+      text-halo-radius: 0;
+    }
```
jeisenbe, line 2566: This accidentally changes the text to be semi-transparent for all the above features if they have a value of `access=` other than `yes` or `permissive` - not just for amenity=shelter but also for tourism=hotel/motel/hostel/chalet/ etc

jeisenbe: This line was already fixed in #3616, sorry

PR #3463, opened 2018-10-20
Render amenity=internet_cafe

jeisenbe:
Fixes [#3451](https://github.com/gravitystorm/openstreetmap-carto/issues/3451)

Changes proposed in this pull request:

Adds rendering for amenity=internet_cafe on nodes and closed ways ("areas") with an icon based on the '@' symbol in "amenity brown" color. The icon and name label are rendered at z17 and up, similar to other amenities. Thanks @Tomasz-W for the icon.

Test rendering with links to the examples:

z18 Node with amenity=internet_cafe
https://www.openstreetmap.org/node/5130908200

z19 Node with amenity=internet_cafe

z18 Closed way (area) with amenity=internet_cafe
https://www.openstreetmap.org/way/436919789
internet-germany

PR #3531, opened 2018-11-27
Ignore .DS_Store files

jeisenbe:
-Update .gitignore file to include .DS_Store

**Explanation:**
.DS_Store files are directory files that are automatically created by MacOS. They should not be accidentally committed to Git

This PR will prevent [mistakes like this](https://github.com/gravitystorm/openstreetmap-carto/pull/3463/commits/1b321a2af6eb02208c8c83ff8f24c71be7cd6f0c) from occurring in the future.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,6 +9,7 @@ localconfig.json
 localconfig.js
 .kosmtik-config.yml
 .env
+.DS_Store
 
 # Generated at runtime by Python.
 *.pyc


polarbearing:
Thanks for suppressing these beasts at least here. I tried to remove them before uploads, but in general they are the most annoying invention in OS X. They spread like a pest, irritate other operating systems, and spoil the last-modified timestamp in my directory tree.

matkoniecz:
@jeisenbe You may also create a global gitignore that will apply to all your repos on your computer - see https://help.github.com/articles/ignoring-files/#create-a-global-gitignore https://stackoverflow.com/a/7335487/4130619

Issue #3534, opened 2018-11-28
Slow rendering on low zoom


rearden-steel:
### Actual behavior

Slow rendering on the latest version on full planet.
Slow query:
```sql
explain analyze SELECT ST_AsBinary("way") AS geom,"feature","way_pixels" FROM (SELECT
            way, name, way_pixels, religion,
            COALESCE(wetland, landuse, "natural") AS feature
          FROM (SELECT
              way, COALESCE(name, '') AS name, religion,
              ('landuse_' || (CASE WHEN landuse IN ('forest', 'farmland', 'residential', 'commercial', 'retail', 'industrial',
                                                    'meadow', 'vineyard', 'orchard') THEN landuse ELSE NULL END)) AS landuse,
              ('natural_' || (CASE WHEN "natural" IN ('wood', 'sand', 'scree', 'shingle', 'bare_rock', 'heath', 'grassland', 'scrub') THEN "natural" ELSE NULL END)) AS "natural",
              ('wetland_' || (CASE WHEN "natural" IN ('wetland', 'mud') THEN (CASE WHEN "natural" IN ('mud') THEN "natural" ELSE tags->'wetland' END) ELSE NULL END)) AS wetland,
              way_area/NULLIF(2445.98::real*2445.98::real,0) AS way_pixels,
              way_area
            FROM planet_osm_polygon
            WHERE (landuse IN ('forest', 'farmland', 'residential', 'commercial', 'retail', 'industrial', 'meadow', 'vineyard', 'orchard')
              OR "natural" IN ('wood', 'wetland', 'mud', 'sand', 'scree', 'shingle', 'bare_rock', 'heath', 'grassland', 'scrub'))
              AND way_area > 0.01*2445.98::real*2445.98::real
              AND building IS NULL
          ) AS features
          ORDER BY way_area DESC, feature
        ) AS landcover_low_zoom WHERE "way" && ST_SetSRID('BOX3D(-626172.1357124997 -626172.1357124997,10644926.3071125 10644926.3071125)'::box3d, 3857);


Subquery Scan on landcover_low_zoom  (cost=9467430.08..9479321.99 rows=78404 width=68) (actual time=39172.202..99529.120 rows=6726182 loops=1)
  ->  Gather Merge  (cost=9467430.08..9476577.85 rows=78404 width=320) (actual time=39031.807..49451.054 rows=6726182 loops=1)
        Workers Planned: 2
        Workers Launched: 2
[...]

PR #3539, opened 2018-11-30
Landcover low zoom cleaning

### LINKED ISSUE #1614: Odd coalescing of name
landcover-low-zoom has the SQL

``` sql
(SELECT
    way, name, way_pixels,
    COALESCE(landuse, "natural") AS feature
  FROM (SELECT
      ('landuse_' || (CASE WHEN landuse IN ('forest') THEN landuse ELSE NULL END)) AS landuse,
      ('natural_' || (CASE WHEN "natural" IN ('wood') THEN "natural" ELSE NULL END)) AS "natural",
      way_area/(!pixel_width!*!pixel_height!) AS way_pixels
    FROM planet_osm_polygon
    WHERE (landuse IN ('forest')
      OR "natural" IN ('wood'))
      AND way_area/(!pixel_width!*!pixel_height!) > 0.01
    ORDER BY z_order, way_area DESC
  ) AS landcover_low_zoom
) AS features
```

The `COALESCE(name, '') AS name` construction originated sometime between v2.15.0 and v2.14.0, but I can't track down exactly where. We were using a JSON MML at the time, not YAML, so `git blame` and other tools are not useful here.

kocio-pl:
Related to #3534.
Resolves #1614.

- removing needless buildings, religion and name related queries from `landcover-low-zoom` data layer

This should make the query marginally faster, but it's mostly code cleaning.

### DIFF
--- a/project.mml
+++ b/project.mml
@@ -100,7 +100,6 @@ Layer:
             WHERE (landuse IN ('forest', 'farmland', 'residential', 'commercial', 'retail', 'industrial', 'meadow', 'vineyard', 'orchard')
               OR "natural" IN ('wood', 'wetland', 'mud', 'sand', 'scree', 'shingle', 'bare_rock', 'heath', 'grassland', 'scrub'))
               AND way_area > 0.01*!pixel_width!::real*!pixel_height!::real
-              AND building IS NULL
           ) AS features
           ORDER BY way_area DESC, feature
         ) AS landcover_low_zoom
--- a/project.mml
+++ b/project.mml
@@ -86,10 +86,10 @@ Layer:
       <<: *osm2pgsql
       table: |-
         (SELECT
-            way, name, way_pixels,
+            way, way_pixels,
             COALESCE(wetland, landuse, "natural") AS feature
           FROM (SELECT
-              way, COALESCE(name, '') AS name,
+              way,
               ('landuse_' || (CASE WHEN landuse IN ('forest', 'farmland', 'residential', 'commercial', 'retail', 'industrial', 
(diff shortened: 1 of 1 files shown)


project.mml
```diff
@@ -100,7 +100,6 @@ Layer:
             WHERE (landuse IN ('forest', 'farmland', 'residential', 'commercial', 'retail', 'industrial', 'meadow', 'vineyard', 'orchard')
               OR "natural" IN ('wood', 'wetland', 'mud', 'sand', 'scree', 'shingle', 'bare_rock', 'heath', 'grassland', 'scrub'))
               AND way_area > 0.01*!pixel_width!::real*!pixel_height!::real
-              AND building IS NULL
```
matthijsmelissen: Did you check performance implications of this?

kocio-pl: No, I just checked that nothing is broken. That's why I claim only code cleaning, not a performance boost.

gravitystorm: This change will probably have a performance regression. There is an important index on the polygon table ([the "nobuilding" index](https://github.com/kocio-pl/openstreetmap-carto/blob/41c12c84e936aebece02351bae36c2b23b524522/indexes.yml#L20-L21)) and the `building is null` part of the query is there in order to use this index and speed up queries. 

It shouldn't be removed without careful testing.

pnorman: Yes, this change needs testing. If it hasn't been tested, we should revert this line.

kocio-pl: Could you make such test?

pnorman: Not in the near future. I think this line should be reverted as it's outside the scope of code cleanups to the landcover SQL.

kocio-pl: OK, PR is ready - #3622.

PR #3541, opened 2018-11-30
Move tram stops later

kocio-pl:
Follow up to #3467.

- After cleaning of z13 tram stops became too visible and make a noise, so they are pushed to z14+.

Warsaw, z13

Before

screenshot_2018-11-30 openstreetmap carto kosmtik

After

screenshot_2018-11-30 openstreetmap carto kosmtik 1

### DIFF
--- a/stations.mss
+++ b/stations.mss
@@ -101,7 +101,7 @@
   }
 
   [railway = 'tram_stop'] {
-    [zoom >= 13] {
+    [zoom >= 14] {
       marker-file: url('symbols/square.svg');
       marker-placement: interior;
       marker-fill: @station-color;


jragusa:
`railway=station` + `station=subway` could be also considered because they display isolated without any connection with a network transport since subways are hidden by the roads.

kocio-pl:
Yes, that's the next thing I thought about. In fact I was thinking about subway stations years ago, but it was not possible without hstore.

Maybe you'd like to make a PR for that?

kocio-pl:
Sure, but certainly not at this zoom level. I don't plan to eradicate them, just move them out to more specific zoom levels. Railway stations can be still visible, because they are located on rail ways, which makes them not too visible.

kocio-pl:
> Oh god, Paris looks like someone having measle.

:smile: 

> Currently, railway=station + station=subway is displayed zoom z12. Do you want just a shift to z14 or do you prefer a distinct behaviour for label ?

I would start with z14+ shift, but if you will find something else that could be tweaked in a sane way to make the slight difference between general station and a subway station, just try. Testing is the key here.

Tomasz-W:
What about rendering tram tracks a little bit less prominent on z13 (and maybe z 14 too)?

kocio-pl:
Well, maybe - somebody should just test it too...

kocio-pl:
@jragusa Any news about subway stations?

Issue #3511, opened 2018-11-14
rendering bugs with attraction=water_slide 


daganzdaanda:
Since #3335 we render attraction=water_slide. 
There are two problems at the moment:

-  The rendering uses smoothing (see #3499) which is unnecessary and distorts what is really in the database.

- We don't render water slides that are mapped as an area (with area=yes)

Examples from https://github.com/gravitystorm/openstreetmap-carto/issues/3499#issuecomment-437073314
A wide water_slide mapped as area, but without "area=yes" gets rendered as a closed way with smoothing: 
https://www.openstreetmap.org/way/110688957
![ws](https://user-images.githubusercontent.com/7693297/48213614-80239480-e37e-11e8-8711-ad457275bf2e.png)

A waterslide with "area=yes" is not rendered yet: https://www.openstreetmap.org/way/401026322
![ws-area](https://user-images.githubusercontent.com/7693297/48213656-9cbfcc80-e37e-11e8-938b-69b55421e16d.png)

Adamant36:
Related to #3499 dealing with smoothing on tracks. In my opinion more testing needs to be done with both. As I'm still not convinced smoothing is actually an issue on either one.

jeisenbe:
We cannot render features that are tagged with the key "attraction" as polygons until the next database reload, because closed ways tagged with only this key is not imported as polygons.

However, I also have reviewed the use of several of these "attraction" keys and the wiki page: https://wiki.openstreetmap.org/wiki/Tag:attraction%3Dwater_slide
It appears that it is not recommended to map these features as an area, but as a line or node. From the wiki:
"Map the water slide by drawing a way way in the direction of the stream and add attraction=water_slide"
This also appears to be the way the large majority of water slides are mapped. 

Therefore, I don't think we should support tagging as an area (with a closed way or multipolygon) with a special rendering. We could consider rendering just a name label on nodes.

The smoothing problem was fixed in #3547, therefore I recommend this issue be closed.

PR #3547, opened 2018-12-02
Remove smoothing for tracks and water slides..

### LINKED ISSUE #3499: Stop smoothing leisure=track as line as it results in very bad rendering in some cases
### Expected behavior
leisure=track should be rendered "as-is", how it is digitized, not smoothed

### Actual behavior
Lines of tracks are being smoothed, as the pull request introducing their rendering included smoothing, see line 2964 in this commit:
https://github.com/gravitystorm/openstreetmap-carto/pull/3443/commits/98d723c40873f7c3c46f5b28a72dfaadf285f01d

afbeelding

Note that in this case, the "tracks" are wrongly tagged, in fact, after inspection of the location, I have concluded these are pitches. I have re-tagged them as such. Primarily affected objects seem to be very ancient, many from before 2010, when leisure=pitch apparently wasn't yet documented, and people tagged them track instead.

Anyway, back to the issue: smoothing falsely suggests detail that isn't there, I would advocate to avoid smoothing on all objects, so as to truly represent them as they were digitized. Most other objects aren't smoothed either, so I think it an anomaly.

StyXman:
Fixes #3499 and partially #3511.

### DIFF
--- a/amenity-points.mss
+++ b/amenity-points.mss
@@ -3151,7 +3151,6 @@
       [zoom >= 17] {
         bridgecasing/line-color: saturate(darken(@pitch, 30%), 20%);
         bridgecasing/line-join: round;
-        bridgecasing/line-smooth: 1;
         bridgecasing/line-width: 1.25;
         [zoom >= 18] { bridgecasing/line-width: 2.5; }
         [zoom >= 19] { bridgecasing/line-width: 5; }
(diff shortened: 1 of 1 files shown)


kocio-pl:
Thanks, at last!

sommerluk:
Thank you!

Issue #2872, opened 2017-09-29
Unify railway rendering


pnorman:
We currently have subtly different styles for many similar types of rail

preserved: ![image](https://user-images.githubusercontent.com/1190866/31009598-e23301d0-a4bc-11e7-97bc-bab445f09af5.png)
rail: 
![image](https://user-images.githubusercontent.com/1190866/31009623-fbd8874a-a4bc-11e7-962b-52afb680c93d.png)
rail service: 
![image](https://user-images.githubusercontent.com/1190866/31009636-07b56718-a4bd-11e7-9bbf-bb1505c81f14.png)

light_rail and subway 
![image](https://user-images.githubusercontent.com/1190866/31009717-4b96aea6-a4bd-11e7-8467-5585476b2d03.png)
tram: 
![image](https://user-images.githubusercontent.com/1190866/31009818-91279174-a4bd-11e7-96df-b3167efb61a6.png)

monorail: 
![image](https://user-images.githubusercontent.com/1190866/31009785-7b2ba716-a4bd-11e7-85e4-f81e011daf4f.png)

miniature: 
![image](https://user-images.githubusercontent.com/1190866/31009867-bceb175e-a4bd-11e7-8452-7f933552e32b.png)

disused: ![image](https://user-images.githubusercontent.com/1190866/31009968-1a25220c-a4be-11e7-94d4-5c6fa9b9c4fd.png)

There are some others too.

This is too many similar renderings.

I think it makes sense to

- Unify light_rail, subway, tram, and monorail. light_rail, subway, tram are all very similarly rendered already, and the differences don't correspond to anything logical. monorail is rendered differently, but in purpose it's similar to the other types.

- Render preserved with the width and dash property of rail, but a lighter colour

I'm not sure what to do about miniature, it doesn't fit in with our style. Disused is also a problem.

pnorman:
This is inhibiting a bunch of code reuse with #2869

math1985:
I agree we have too many renderings now.

I'm against unifying tram and subway, there is an important and clear distinction between both in many European cities. Even the arbitrary difference that we currently have is useful.

Unifying light_rail and subway makes sense.

dieterdreist:
2017-09-29 11:33 GMT+02:00 Paul Norman <notifications@github.com>:

> Unify light_rail, subway, tram, and monorail. light_rail, subway, tram are
> all very similarly rendered already, and the differences don't correspond
[...]

PR #3553, opened 2018-12-08
[WIP] Borders and country name rework

### LINKED ISSUE #3489: Change the colors of admin borders
Current colors of admin borders are too bright, especially on low zoom levels. On low zoom they were gray once, but they were changed to violet for consistency.

Possible solutions of this problem include using gray this way or another, which result in darker colors:
1. German fork uses gray with a green tint (`#50705a`)
2. @imagico fork uses very similar mix of gray with a green tint (`#51785e`) on the land
3. I was testing plain gray (`#555555`) in #2695

Probably all of them are better than current color, but I think the best solution is to use gray with violet tint (probably 50% mix of `#555555` and current bright violet `#ac46ac`). It works quite nice - looks grayish on low zoom and more like violet on high zoom levels without any tuning. It allows to shift road colors like in Christoph fork, while not using green.

This is the final effect of violet tint for country borders, including road colors shift:

sbbo0u5x
6 nsthx_

This is the road colors shift alone (described in this blog post: http://blog.imagico.de/more-new-colors/ ) - the relevant part is only color shift from motorway to tertiary, the rest of the tuning is outside the scope of this issue:

roadcolor_z15

Of course we can also use green tint for admin borders, but green is better associated with (semi-)natural objects, in particular it's close to the natural reserve rendering with green labels along the boundary:

German green tint

screenshot_2018-11-01 openstreetmap deutschland karte

current bright violet

screenshot_2018-11-01 openstreetmap carto kosmtik 12

violet tint

screenshot_2018-11-01 openstreetmap carto kosmtik 13

Violet tint works good on all the zoom levels, even with roads color shift and expanded natural areas to z5+ (see #3458):

high

kyluz9v7

medium

hocaees6

low

ai0dig8q

### LINKED ISSUE #3526: Test more admin border colors
In #3489 @kocio-pl discussed changing the color of admin borders, and tested the look of dark violet borders. He mentioned two other possibilities: 
1. Gray with a green tint, used by the German fork and @imagico's fork
2. Plain gray, tested by @kocio-pl in #2695

I prefer to avoid anything involving violet for the borders, see also @Tomasz-W's comment https://github.com/gravitystorm/openstreetmap-carto/issues/3489#issuecomment-439676660
Here I test the results of the other two options, specifically:
1. The German-style gray with a green tint `#50705a`, with a slightly darker `#445f4d` for low zooms (z<8)
2. Plain gray `#646464` which is a desaturated version of the German-style color, with a slightly darker `#555555` for low zooms (z<8), which is the same color that @kocio-pl tested in #2695.

In addition, I feel that two other changes are beneficial:
* With gray or almost gray borders, the country labels would be shown in black and therefore in the same style as city labels. To distinguish country labels from the city labels I've made them bold.
* The second change is related to the line width of level 2 (international) borders. This line width jumps from 2 to 6 pixels between zoom 9 and 10, and I have made the change more gradual (see graph).

linewidth

z4 before
z4osm

z4 German-style color
z4ger

z4 gray
z4gray

z6 before
z6osm

z6 German-style color
z6ger

z6 gray
z6gray

z8 before
z8osm

z8 German-style color
z8ger

z8 gray
z8gray

z10 before
z10osm

z10 German-style color
z10ger

z10 gray
z10gray

z13 before
z13osm

z13 German-style color
z13ger

z13 gray
z13gray

kocio-pl:
Fixes #3489.
Closes #3526.
Related to #621 and #3102.
Related to #3563

- make borders much darker (`#ac46ac` +75% of grey =  `#6b516b`) 
- render country names in bold

This has been discussed and tested before, so now just a short summary: current borders are too bright, especially on low zoom levels and on the water. Other forks have already tried to make them darker using green tint, however it looks bad on the water, looks a bit blurry in general and is too close to natural reserve boundaries. Using pure grey makes some other (however less serious) problems, pointed out in https://github.com/gravitystorm/openstreetmap-carto/pull/2695#issuecomment-317164894.

Using dark violet looks good on low zoom, looks good on the water and is different than natural reserve boundaries. It's also similar to the current style, so people should be less surprised. Making country names bold helps to recognize them more easily than currently.

Some examples (I hope @rrzefox could apply it to test such important change on the whole planet):

cjyo 28y

https://www.openstreetmap.org/#map=6/51.849/20.116

1a8c4dr

https://www.openstreetmap.org/#map=16/49.5738/22.6873

birdt_1x

https://www.openstreetmap.org/#map=11/54.3842/19.7208

chtdclj_

https://www.openstreetmap.org/#map=12/52.2356/20.8412

tah3cucw

https://www.openstreetmap.org/#map=19/52.25929/20.99742

1iunm0hv

https://www.openstreetmap.org/#map=18/52.23921/21.03554

_bzqmee9


matkoniecz:
> https://www.openstreetmap.org/#map=19/52.25929/20.99742

It seem to be almost the same as `railway=disused`.

kocio-pl:
Currently they are already similar, because colors of slim lines are not too visible:

https://www.openstreetmap.org/way/461746771#map=19/52.25954/20.98289

current master

aaz9cdmy

z9+ rendered with current z7+8 pattern:

tn724nie

kocio-pl:
Rendering level 5 with 3 dots and longer dashes (`line-dasharray: 12,3,2,3,2,3,2,3;`):

https://www.openstreetmap.org/#map=18/50.17957/18.62354

Before 

dxnc2zit

After 

hy4ttvfd

Tomasz-W:
related to https://github.com/gravitystorm/openstreetmap-carto/issues/2872

vholten:
@kocio-pl Your test renders show bold country names without `text-character-spacing`, but this PR does not show that change in the code, the `text-character-spacing: 0.5;` is still there.

I think we agree that character spacing is no longer required (see also the [comment](https://github.com/gravitystorm/openstreetmap-carto/issues/3489#issuecomment-442863904) by sommerluk). Could you make that change in the code?

kocio-pl:
Thanks for checking the code, now it's removed.

rrzefox:
> I hope @rrzefox could apply it to test such important change on the whole planet

This has now been deployed together with #3563, and Z0-9 rerendered.

kocio-pl:
Example of different scripts in bold (Chinese font has general issues on this server, so I can't really test it) - click to see the full size:

screenshot_2018-12-14 rrze openstreetmap server - leaflet demo

screenshot_2018-12-14 rrze openstreetmap server - leaflet demo 1

screenshot_2018-12-14 rrze openstreetmap server - leaflet demo 2

sommerluk:
@rrzefox Thanks a lot!

@kocio-pl  Maybe `medium` font could be an option for regions (admin_level 4).

Initially, I thought `semi-bold` might work well for capitals, but after looking at the rendering live preview, I suppose this would be too invasive. Maybe it’s better to give slightly more emphasis to region names, but `semi-bold` might be too much. `Medium` is something between `semi-bold` and `regular`. The fallback would be `regular`.

kocio-pl:
For SemiBold this looks strange, like if using polish diacritics in one line triggers fallback to a font without this style -  I did not expect that:

SemiBold 

skbk0rkp

This is also how the country name would look in comparison (similar problems with Medium):

SemiBold 

xd6dvdyh

Medium

6qdtivwr

sommerluk:
@kocio-pl Well, the rendering is indeed strange. And I do not really understand why.

> I have added a simple Bold clone and turned it into medium in the code, I hope this is OK.

Theoretically, this should work. Pitfalls in the fallback chain might be
- "Noto Sans CJK JP Medium" (duplicates all LGC characters, but with different glyphs than Noto Sans)
- "Noto Naskh Arabic UI Medium" (duplicates _some_ LGC characters)
- "DejaVu Sans Medium" (doesn’t exist).

But nevertheless, the question is why the fallback chain is triggered. Let’s take [województwo śląskie](https://www.openstreetmap.org/relation/224462) as example. The label renders wrong – for both, medium (fallback to regular for the second word) and semi-bold (fallback for both words). If I understand it correctly, Mapnik chooses the font together with Harfbuzz, and they try to keep at least hole words always in the same font. If a font does not provide all characters for a given word, they fall back to the next font in the chain, and so on. The big question is: Why is the fallback chain triggered at all? I’ve tested the label “województwo śląskie” with Noto Sans in all relevant weight variants using InDesign (which does quite a good job in Unicode-conform rendering and shows missing characters): The result is that there are _no_ missing characters. So everything should be fine. I don’t understand why it doesn’t render correctly. Note that additionally to the fallback, there seems also to be a problem with different line spacing.

(By the way: Where did you get the Noto Sans Medium/Semi-bold font you are using? Github?)

Given that for both – Medium _and_ Semi-bold – there are rendering issues, and we do not understand the reason, I propose not to include Medium/Semi-bold in this PR. Maybe later there can be another PR that focuses exclusively on Medium/Semi-bold.

sommerluk:
Maybe also in a later PR, legacy DejaVu support could be dropped. Given that Noto is quite wide-spread today, dropping DejaVu could avoid some side effects in the fallback chain.

kocio-pl:
> (By the way: Where did you get the Noto Sans Medium/Semi-bold font you are using? Github?)

Most probably I was following installation instructions (and looking at files it was year ago):

https://github.com/gravitystorm/openstreetmap-carto/blob/master/INSTALL.md#installation-on-other-operation-systems

I will try to update them and see the results. Could you make independent tests on your system?

sommerluk:
> Most probably I was following installation instructions

Then you have downloaded from Github, which should be fine.

> Could you make independent tests on your system?

Not sure, I’m currently quite busy with other stuff, so I do not have much time…

kocio-pl:
OK, we have time to test things and find the problem before release. If that fails, I will just drop this part.

BTW: @rrzefox Could you update the code on your system so we could check if this is not just the issue with my setup?

vholten:
In the Medium examples, some text is shown in Noto Sans CJK JP Medium instead of Noto Sans Medium (the lowercase l has a tail in the CJK font):

lubelskie

noto sans

vholten:
I looked at all the Noto fonts on my system, and the CJK fonts are the only ones where the lowercase l has a tail, so I assume it's that one.

sommerluk:
> I looked at … so I assume it's that one.

That’s reasonable. Between Noto Sans and Noto Sans CJK, apart from “l”, there are some minor differences also in “s” and “e”, but that’s more difficult to see, especially on the low-resolution rendering we have here.

> Fallback is triggered because the CJK font doesn't have Polish diacritics.

Yes and no. Anyway, the LGC font _Noto Sans Medium_ is (correctly) the first choice in the list of @kocio-pl so I would assume that LGC font _Noto Sans Medium_ covers Polish diacritics and there is no further fallback. But apparently my assumption is wrong.

Google’s _Noto Sans CJK_ is _actually_ Adobe’s _Source Han Sans_ – just a copy that has been renamed: Google is commissioning most of it’s Noto fonts to Monotype. One exception are the CJK fonts: For these fonts, they have collaborated with Adobe because it’s really a huge project. So Adobe is producing its CJK font  _Source Han Sans_ with a design that match both, Adobe’s own LGC font _Source Sans Pro_ and also Google’s LGC font _Noto Sans_. After releasing _Source Han Sans_, they copy the font, rename it to _Noto_ and integrate it in the Noto family.

Now the problem is that Adobe adds the most common LGC characters from _Source Sans Pro_ directly to its _Source Han Sans_ font – as a sort of fallback. Google however follows a different policy with the other Noto fonts: Usually, there are no LGC fallback characters, so when you use for example _Noto Sans Hebrew_ than this font will not have any LGC characters. So _Noto Sans CJK_ is an exception, and additionally its LGC fallback glyphs are not what you would expect, but rather _Source Sans Pro_ glyphs.

*LGC = Latin, Greek & Cyrillic scripts

kocio-pl:
Could anybody else make such test to check what is the problem? @Adamant36? @jeisenbe?

jeisenbe:
I tried to fetch your branch, but I am getting errors that the project is
not loading properly. First there was an error about the fabric.svg
missing; I fixed this by rebasing on the master branch, but now I am told I
am missing many fonts, eg Noto Sans Medium, Noto Sans CJK JP Medium, etc.

Do I need to download a new set of fonts to render this branch?

I’m using the Docker container on MacOS to run kosmtik.

Adamant36:
I havent tried the branch yet, but I have been getting a massive amount of font errors when testing other stuff after I updated my master branch like a week ago. It happens multiple times when zooming in and slows things a lot. Although it usually still loads, there's been a few times it completely froze kosmtik (one time after I changed the landuse color docker wouldnt even load completely). Its one of the reasons I havent been doing much the last week. 

So I think this might be a wider issue caused by something other then just the border and country name rework.

vholten:
I fetched your branch and downloaded 
- `Noto Sans Medium` from https://github.com/googlei18n/noto-fonts/raw/master/hinted/NotoSans-Medium.ttf
- `Noto Sans CJK JP Medium` from https://github.com/googlei18n/noto-cjk/raw/master/NotoSansCJKjp-Medium.otf

I installed these two fonts by copying them to `~/.fonts/`
(For this test I didn't download any of the other Medium fonts)

Rendering was successful (landuse omitted to save time):
pl-kociobranch

A possible explanation for the behavior on @kocio-pl's system is that `Noto Sans Medium` cannot be found by Mapnik (maybe it's installed in a folder where it cannot be found) so the next font on the list, `Noto Sans CJK JP Medium` is used. And that font doesn't have all Polish diacritics, leading to further fallback.

kocio-pl:
In general yes, I'm happy with most of the changes, the only floating bit is rendering of `admin_level=4`. The main possibilities are:
- try Medium, proposed by @sommerluk (currently in the PR)
- try SemiBold (it looks more distinctive for me)
- leave it as it is (at least for this PR), which means Italic

What do you think?

sommerluk:
I would also rather tend to Semi-bold (but I could live also with the other options).

Anyway, _if_ this gets changed, we would have to make some smaller adaptions on the font list, and also change the install instructions. As far as I know, Noto Sans Medium is not packaged with Ubuntu – is this correct? If so, I could provide some alternative download instructions. And we need to warn the operations team when making a release, so that they are aware of the change. Not sure if they will actually install these fonts, but as we have a fallback, that’s not such a big problem.

kocio-pl:
Would you like to prepare separate code with Semi-bold? I can revert Medium changes and leave Italic for now.

vholten:
> I can revert Medium changes and leave Italic for now.

I would support leaving italic for now and looking at state names in a separate PR, as it seems that more tests are needed.

kocio-pl:
> Could you perhaps propose the change of the label font and the color change in different PRs? They are independent from each other, right?

Sure, state labels changes are different thing, so I will remove it soon from this PR.

> Perhaps we should abandon the idea of having the same border color on all zoom levels?

Ok, that sounds sane for me. 
- do you think that we should keep the old color for highest admin levels, new color for the lowest and make gradual color change for each admin level between them?
- what do you think about inserting new level 5 pattern and shifting the rest in such case?

matthijsmelissen:
> do you think that we should keep the old color for highest admin levels, new color for the lowest and make gradual color change for each admin level between them?

That sounds reasonable. Or perhaps the new color can be even more gray (or a different color) if it is only for low zoom. Perhaps an abrupt change in color works as well, we'd need to test that.

> what do you think about inserting new level 5 pattern and shifting the rest in such case?

I have no opinion on that at the moment.

kocio-pl:
> Perhaps an abrupt change in color works as well, we'd need to test that.

Well, we already had this once and it was changed... :smile: - see #1749 fixed by #1773.

matthijsmelissen:
I think I regret that change now.

vholten:
Test with the [7-color palette](https://meyerweb.com/eric/tools/color-blend/#6B516B:AC46AC:5:hex) suggested by @kocio-pl:
z12, https://www.openstreetmap.org/#map=12/50.4465/16.9049
This area contains many admin_level=8 borders, one admin_level=4 state border and an international border.

7colors

Tomasz-W:
@matthijsmelissen Can you provide some examples of places where gray borders might be a problem?

Btw. See https://github.com/gravitystorm/openstreetmap-carto/issues/2872#issuecomment-449655031. I think we should change some railway types rendering, because they cause conflicts with borders even when they are violet, so "palette solution" doesn't work for me anyway.

matthijsmelissen:
For example https://user-images.githubusercontent.com/5439713/49116116-ee33eb00-f29c-11e8-9a62-c415df1a1f21.png is for me hard to read.

vholten:
@kocio-pl https://github.com/vholten/openstreetmap-carto/tree/borders-blend

kocio-pl:
Thanks! Could you also adjust the label colors accordingly?

vholten:
The color of country labels was already adjusted, state labels may work with that color as well. 
I made another branch https://github.com/vholten/openstreetmap-carto/tree/borders-blend-statelabels where the color of state labels is the same as state borders.

In both branches I made country names bold.

kocio-pl:
Sorry for late reply, I am still recovering with my focus on this repo and was unable to get where the problem is. 

My idea is to make all the labels correspond to the line color, not only country and state labels. I think it will make it more clear that they are referring to the same admin level.

PR #3561, opened 2018-12-10
Unify code of power=tower and power=pole

### LINKED ISSUE #1179: Merge power-towers and power-poles
Currently we define two layers, `power-towers` and `power-poles`.

It would be a performance gain if these were merged into one layer, ordered with poles before towers in the SQL.

This would be a good task for someone wanting to learn how to edit the style.

jragusa:
Fixes #1179

* Merge layers `power-towers` and `power-poles` in `project.mml`
* Adjust modifications in `power.mss`

there is no rendering differences
https://www.openstreetmap.org/#map=17/46.17629/6.23976
`z=15`
power_towers_z15
`z=16`
power_towers_z16
`z=17`
power_towers_z17

### DIFF
--- a/project.mml
+++ b/project.mml
@@ -1722,15 +1722,14 @@ Layer:
         (SELECT
             way,
             power
-            CASE power
-              WHEN 'tower' THEN 1
-              WHEN 'pole' THEN 2
-              ELSE 3
-            END
-              AS prio
         FROM planet_osm_point
-        WHERE power = 'tower' OR power = 'pole'
-        ORDER BY prio
+        WHERE power IN ('tower', 'pole')
+        ORDER BY
+          CASE
+            WHEN power = 'tower' THEN 1
+            WHEN power = 'pole' THEN 2
+            ELSE NULL
+          END
         ) AS power_towers
     properties:
       minzoom: 14
--- a/project.mml
+++ b/project.mml
@@ -1722,8 +1722,15 @@ Layer:
         (SELECT
             way,
             power
-          FROM planet_osm_point
-          WHERE power = 'pole' OR power = 'tower'
+            CASE power
+              WHEN 'tower' THEN 1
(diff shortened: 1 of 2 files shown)


project.mml
```diff
@@ -1720,25 +1720,13 @@ Layer:
       <<: *osm2pgsql
       table: |-
         (SELECT
-            way
+            way,
+            power
           FROM planet_osm_point
-          WHERE power = 'tower'
+          WHERE power = 'tower' OR power = 'pole'
```
kocio-pl: As said by @pnorman in https://github.com/gravitystorm/openstreetmap-carto/issues/1179#issue-52973210:
> It would be a performance gain if these were merged into one layer, ordered with poles before towers in the SQL.

I'm not sure why is that, but it looks like the order is important then.

jragusa: ok, done

jragusa:
yes, I'm working on it.

project.mml
```diff
@@ -1720,25 +1720,13 @@ Layer:
       <<: *osm2pgsql
       table: |-
         (SELECT
-            way
+            way,
+            power
           FROM planet_osm_point
-          WHERE power = 'tower'
+          WHERE power = 'pole' OR power = 'tower'
```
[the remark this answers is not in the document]
jragusa: I modified the syntax to include ORDER BY condition. Is it correct ?

imagico:
@jragusa - do you still intend to complete this?  The current code does not work because of

https://github.com/gravitystorm/openstreetmap-carto/blob/00004485d8e35604c668a0905e4b623dae456b79/project.mml#L1724-L1725

Also note if you don't use the prio column anywhere else you don't need it, you can just inline the CASE in the ORDER BY (like done in many of the road queries for example).

jragusa:
@imagico I have modified according to your remark. What do you think ?

Issue #3538, opened 2018-11-29
Make railways lighter, minor highways less prominent at z8-10


Tomasz-W:
Zoom levels 8, 9 and 10 looks like a unreadable colorful spagetti in some places at the moment:

Examples:
https://www.openstreetmap.org//#map=8/49.246/15.430
https://www.openstreetmap.org//#map=9/51.0811/5.8173
https://www.openstreetmap.org//#map=10/50.2498/19.2906
https://www.openstreetmap.org//#map=10/53.5550/-2.0627

Problems to solve:

- [x]  too prominent protected areas borders -> done! https://github.com/gravitystorm/openstreetmap-carto/pull/3574
- [x]  too thin country borders on z8-9 (related to https://github.com/gravitystorm/openstreetmap-carto/issues/3489 and https://github.com/gravitystorm/openstreetmap-carto/issues/3526) -> done! https://github.com/gravitystorm/openstreetmap-carto/pull/3563
- [ ] too prominent railways
- [ ] roads system width tuning up

I believe it could be cleaned similar as z13-14 was in https://github.com/gravitystorm/openstreetmap-carto/pull/3467.

What do you think about it? Do you see a problem there, too?

kocio-pl:
I would change the title to be more clear that it's about lines.

- country borders on z8-9 could be bolder, I agree 
- protected areas borders might be thinner, although there are only some places when borders of adjacent areas add up which I consider to be a problem:
https://www.openstreetmap.org/#map=8/47.063/-75.457

![screenshot_2018-11-29 openstreetmap](https://user-images.githubusercontent.com/5439713/49248050-1ea09400-f419-11e8-881f-3ea5211d83e9.png)

- I don't feel railways could be thinner, because some roads are grey already

Adamant36:
I dont think protected areas borders should be thicker or brighter where the they join/overlap. It makes the map look really scrued up. For instance, in northern California it looks horrible and inconsistant.

kocio-pl:
The question is how to fix it...

vholten:
Protected area borders have shading on the inside. So borders between adjacent areas become twice as thick (shading on both sides). That could be fixed by shading that is centered on the border instead of inside.

kocio-pl:
That's also what I was suspecting. Could you prepare a PR fixing that?

Adamant36:
[...]

PR #3563, opened 2018-12-11
Thicker country borders at z8 and z9

vholten:
Related to #3538.

- Make country borders thicker at z8 and z9.

This will make these borders more prominent at these zoom levels, where they are currently quite thin. It will also make them easier to distinguish from railways when the border color from #3553 is implemented and when railways are made lighter in the future.

The "after" renders are shown with the border color from #3553. 

z8 before 
z8osm

z8 after
z8bordersonly

z9 before
z9osm

z9 after
z9bordersonly

### DIFF
--- a/admin.mss
+++ b/admin.mss
@@ -37,6 +37,14 @@ overlapping borders correctly.
       background/line-width: 2;
       line-width: 2;
     }
+    [zoom >= 8] {
+      background/line-width: 3;
+      line-width: 3;
+    }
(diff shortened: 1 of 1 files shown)


matthijsmelissen:
Does this depend on #3553?

kocio-pl:
No, it was just not tested with current borders style.

matthijsmelissen:
Thanks!

Issue #1280, opened 2015-02-01
Render support for highway=residential_link


aaronstar:
As above I suggest rendering support for highway=residential_link, highway=service_link, which is currently supported by various routing software and a natural extension of support for Link Roads.

mkoniecz:
http://taginfo.openstreetmap.org/tags/highway=residential_link - 57 instances worldwide
http://taginfo.openstreetmap.org//search?q=highway%3Dservice_link - 0 instances worldwide

Undocumented on wiki, very likely to be controversial.

It should be (a) documented (b) used to even consider rendering it.

mboeringa:
I also wonder what the "various routing software" are that support a hardly used tag... I guess you may be the developer of the one solution that now does, but it might be helpful to state it as such, and explain your reasons.

I also don't understand the concept of a **residential_link** or **service_link**. In most cases, if anything, I can only think of ***lanes***, but these aren't the typical "links" of a motorway junction...

aaronstar:
I am unsure if this can be reopened (quick decision) regardless,

Here is an example canidate of a road where highway=residential_link should be used.
https://www.openstreetmap.org/edit#map=19/-31.91795/115.86190

The intersecting road is two way with a turning segment bypass.

Link Roads may be mistaken for being motorway links. This has resulted in inconsistent usage. Link Roads are Slip Roads. Slip Roads include indeterminate and terminal ramps (I.e a T junction in cases) ingress and egress ramps and turning segments and intersection bypass segments (excluding those which turn unto themselves) I.e. U turns within the median.

Skobbler and Scout have native support for highway=residential_link and highway=service_link,
others including OsmAnd and OSRM support wildcards of highway values to accomodate link roads (Wherein they don't support highway=primary_link or the like at present as standalone values but rather they support link roads.).

pnorman:
I thought we did render residential_link roads, did this maybe change since the XML?

> Undocumented on wiki

This is not particularly relevant - none of the links except for motorway_link have their own page

pnorman:
Re-opening, residential_link is supported by MQ Open and Humanitarian so we shouldn't dismiss it out of hand.

mkoniecz:
> This is not particularly relevant - none of the links except for motorway_link have their own page

Neither http://wiki.openstreetmap.org/wiki/Key:highway nor http://wiki.openstreetmap.org/wiki/Highway_link is even mentioning highway=residential_link.

gravitystorm:
[...]

PR #3567, opened 2018-12-13
Select only motorway_link to tertiary_link as link in SQL

### LINKED ISSUE #3566: Undocumented link roads should not render
### Expected behavior
We previously rejected a request to render the undocumented tag residential_link. Similarly, unclassified_link, service_link and other undocumented "link" highways should not render. See #1280

### Actual behavior
All highway=xxx_link are currently being rendered the same as highway=xxx

Currently all highway=xxx_link are being rendered as highway=xxx, including residential_link, service_link and road_link. These features are also rendered above more important roads. 

This is a test area I created in JOSM, where I noticed the problem. Each of the small curving roads is a link (Left: road_link, bottom: residential_link, right: unclassified_link)

links-master

The current SQL query in project.mml for roads-casing and road-fill is:

```
            (CASE WHEN substr(feature, length(feature)-4, 5) = '_link' THEN substr(feature, 0, length(feature)-4) ELSE feature END) AS feature,
```

The result is that all highway_xxx_link are selected as highway_xxx, and currently render.

Test area showing all highway features and links being rendered. Each of the connecting, diagonal "highways" is in the form highway=xxx_link; eg highway=cycleway_link, highway=construction_link

links2-master

jeisenbe:
Fixes #3566
Related to #1280

- Modify SQL query in project.mml to only select as features motorway_link, trunk_link, primary_link, secondary_link and tertiary_link.

Before
links2-master

After
links2-after

### DIFF
--- a/project.mml
+++ b/project.mml
@@ -955,7 +955,8 @@ Layer:
         (SELECT
             way,
             COALESCE(
-              ('highway_' || (CASE WHEN substr(highway, length(highway)-4, 5) = '_link' THEN substr(highway, 0, length(highway)-4) ELSE highway end)),
+              ('highway_' || (CASE WHEN highway IN ('motorway_link', 'trunk_link', 'primary_link', 'secondary_link', 'tertiary_link')
+                                     THEN substr(highway, 0, length(highway)-4) ELSE highway end)),
               ('railway_' || (CASE WHEN (railway = 'rail' AND service IN ('spur', 'siding', 'yard')) THEN 'INT-spur-siding-yard'
                                    WHEN railway IN ('rail', 'tram', 'light_rail', 'funicular', 'narrow_gauge') THEN railway ELSE NULL END))
             ) AS feature,
--- a/project.mml
+++ b/project.mml
@@ -452,7 +452,7 @@ Layer:
       table: |-
         (SELECT
             way,
-            (CASE WHEN substr(feature, length(feature)-4, 5) = '_link' AND feature IN ('highway_motorway_link', 'highway_trunk_link', 'highway_primary_link', 'highway_secondary_link', 'highway_tertiary_link') THEN substr(feature, 0, length(feature)-4) ELSE feature END) AS feature,
+            (CASE WHEN feature IN ('highway_motorway_link', 'highway_trunk_link', 'highway_primary_link', 'highway_secondary_link', 'highway_tertiary_link') THEN substr(feature, 0, length(feature)-4) ELSE feature END) AS feature,
             horse,
             foot,
             bicycle,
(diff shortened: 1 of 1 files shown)


jeisenbe:
I have tested this code and it works properly.
GitHub isn't allowing an edit to the original comment right now, for some reason, nor will it allow me to upload the "after" image. We'll see if it works better in the morning.

jeisenbe:
Test area shows changes:

Before
links2-master

After
links2-after

matkoniecz:
Can you upload test file to http://gist.github.com/ or similar to avoid need for recreating it for a PR review testing?

jeisenbe:
You mean the osm.xml file? OK, here's a zip:
[road-test.osm.xml.zip](https://github.com/gravitystorm/openstreetmap-carto/files/2684803/road-test.osm.xml.zip)

kocio-pl:
@matkoniecz Do you think you could take care of this PR?

jeisenbe:
@matthijsmelissen do you want to review this? 
Or @matkoniecz ?

Here is an updated .osm.xml file with all sort of highways and construction types for reviewing:
[road-test.osm.xml.zip](https://github.com/gravitystorm/openstreetmap-carto/files/2722373/road-test.osm.xml.zip)

matthijsmelissen:
> Here is an updated .osm.xml file with all sort of highways and construction types for reviewing:


@jeisenbe  For your information, this is my test data: https://github.com/matthijsmelissen/openstreetmap-test-data

jeisenbe:
Thank you for catching that! I've tested the change and pushed a new commit.

jeisenbe:
> "replace the `substr()` later in the generation of the link column with an explicit list of the five values"

If I understand correctly, that line selects `motorway_link` as `motorway`, `trunk_link` as `trunk` and so on. This lets the roads.mss code have motorway and motorway_link as one feature. Later another line has `substr(highway, length(highway)-4, 5) = '_link' THEN 'yes'` as the column `link`, and this is used to distinguish link roads in the roads.mss code. 

How should we change that line, `THEN substr(feature, 0, length(feature)-4) ELSE feature END)`?
I don't know how to fix this, yet.

imagico:
As, sorry for being cryptic.

I meant

```sql
CASE
  WHEN highway IN ('motorway_link', 'trunk_link', 'primary_link', 'secondary_link', 'tertiary_link')` THEN 'yes'
  ELSE 'no'
END AS link,
```

instead of

```sql
CASE
  WHEN substr(highway, length(highway)-4, 5) = '_link' THEN 'yes'
  ELSE 'no'
END AS link,
```

This makes the conditions equivalent and makes sure the link column is only yes for the five link types supported.  Practically this makes no difference because the mss code only interprets link for those types anyway but it is more consistent overall.

jeisenbe:
Oh, I should have read your comment more carefully. Thank you for the
clarification.

I’ve tested the changes and pushed a new commit.

imagico:
In `roads-low-zoom` you have now changed the link column but not the feature column:

https://github.com/gravitystorm/openstreetmap-carto/blob/0b4faa8453909566e07d805d454f18b5499c6c54/project.mml#L958

Either variant is fine there  (`planet_osm_roads` only contains motorway to secondary) but you should use the same approach for both in this query.

jeisenbe:
OK, it should be fixed. I've tested at z9 and z10, and pushed the changes.

matthijsmelissen:
> Either variant is fine there (planet_osm_roads only contains motorway to secondary) but you should use the same approach for both in this query.

I preferred the way it was before this suggestion, because now we are duplicating (DRY) the list of valid link tags.

Anyway I'm fine either way, going to merge this now.

imagico:
As said i would have been fine with either variant.

Congrats to @jeisenbe for his first successful battle with the road layers SQL... :smile:

jeisenbe:
Yay! Thanks.

PR #3570, opened 2018-12-14
Add tertiary-link width, same as other links

### LINKED ISSUE #1109: Make highway _links same width?
Motorway_link seems to be the only "highway_link" that has it's own width.
Where a motorway_link meets a primary_link or secondary_link, the lower ranked road is wider.
http://www.openstreetmap.org/node/268585122
http://www.openstreetmap.org/node/412383929

Would a single width for all possible highway_links be a good idea?

jeisenbe:
**Fixes #1109** 
(This issue is currently closed, but tertiary_link was not added)

- Render highway=tertiary_link thinner than highway_tertiary (generally the same width as secondary_link)

**Explanation:**
Leaving different tertiary links causes oddities in connections to higher-order tertiary links, eg 
[](https://user-images.githubusercontent.com/19976610/45986666-4d7a4280-c043-11e8-97cb-28c287caaa12.PNG)

Tertiary_link is used over 150k times and is mentioned in the highway link documentation, and it is currently rendered, but at the same width as highway=tertiary. This makes the tertiary links significantly wider than motorway, trunk, primary and secondary links.

Previously this was not fixed, because tertiary has been rendered in white, the same as minor roads such as highway=unclassified and highway=residential, and the thinner tertiary links will be similar in appearance to these roads. However, there is currently a plan to change the highway colors, which will render tertiary in yellow. Even if this is not implemented, it is better to render the link roads consistently.

**Test rendering with links to the example places:**

Tertiary links in Wales, Great Britain
1.  Before
1-tertiary-master

After
1-tertiary-after

2. Before
2-tertiary-master

After
2-tertiary-after

3. Before
3-tertiary-master

After
3-tertiary-after

### DIFF
--- a/roads.mss
+++ b/roads.mss
@@ -96,6 +96,7 @@
 @secondary-width-z12:             3.5;
 @secondary-link-width-z12:        1.5;
 @tertiary-width-z12:              2.5;
+@tertiary-link-width-z12:         1.5;
 @residential-width-z12:           0.5;
 @unclassified-width-z12:          0.8;
 
@@ -108,6 +109,7 @@
 @secondary-width-z13:             5;
 @secondary-link-width-z13:        4;
 @tertiary-width-z13:              4;
+@tertiary-link-width-z13:         3;
 @residential-width-z13:           2.5;
 @living-street-width-z13:         2;
 @bridleway-width-z13:             0.3;
@@ -135,6 +137,7 @@
 @secondary-width-z15:             9;
 @secondary-link-width-z15:        7;
 @tertiary-width-z15:              9;
+@tertiary-link-width-z15:         7;
 @residential-width-z15:           5;
 @living-street-width-z15:         5;
 @pedestrian-width-z15:            5;
@@ -166,6 +169,7 @@
 @secondary-width-z17:            18;
 @secondary-link-width-z17:       12;
 @tertiary-width-z17:             18;
+@tertiary-link-width-z17:        12;
 @residential-width-z17:          12;
 @living-street-width-z17:        12;
 @pedestrian-width-z17:           12;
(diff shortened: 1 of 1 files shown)


kocio-pl:
This was skipped in #3374 to not mix with smaller roads, because they are all white.

Example 3 shows that probably tagging is inconsistent, but it can look pretty bad if all such connectors are tagged as links.

jeisenbe:
Re example 3.

Do you mean it looks bad because it is mistagged? Or that it would look bad
if properly tagged?

kocio-pl:
I am not sure which tagging is proper, but thinner rendering looks bad to me.

Adamant36:
For me it looks better in the first and third (although kind of wierd in the third) examples, but not in the second example.

jeisenbe:
Hmm, the wiki page / talk page suggests that there is some debate about
whether this sort of y-junctions and entrances to traffic circles should be
tagged as links, so perhaps this is a tagging error.

I think it looks better when the link is branching off from the main
tertiary road, but it looks odd when a highway=tertiary splits into two
highway=tertiary_link - and perhaps this is not the best way to tag.

Adamant36:
I could see why there would be a debate about it. Misstaging seems possible. As it seems weird. As you say though, it looks good branching off the main road. So I dont know.

jeisenbe:
I asked about this issue on the tagging mailing list, and a couple of people responded. They agreed that a road that splits to enter a roundabout, or at a Y-junction, should not be tagged as a highway=*_link. So examples 2 and 3 above are mistagged, by this standard. 

https://lists.openstreetmap.org/pipermail/tagging/2018-December/041500.html

This is good to know, because all of the non-motorway links (trunk thru secondary) can also be affected by this rendering issue.

kocio-pl:
Example 1 looks better with new rendering, so I would upvote it now.

kocio-pl:
I believe it's good to try this for catching tagging problems, thanks!

PR #3571, opened 2018-12-14
Use usual subway bridge style for subway construction bridges

tpetillon:
* Make subway bridges under construction thinner and only appear from zoom 14 on.

Currently, all railway construction bridges are rendered the same. However not all (non-construction) railway bridges are identical.

Namely, subway bridges are thinner and appear later (zoom 14 instead of 13) than standard railways.

This PR unifies the appearance of subway bridges, whether under construction or not.

Some other railway types have their own rules for bridges (light rail, funicular, narrow gauge, etc.). They are not handled in this PR, but could be if there is interest.

### Examples
(Area: https://www.openstreetmap.org/#map=14/48.1131/-1.6690)
#### Zoom 13
Before
image

After
image

#### Zoom 14
(This one is more subtle.)

Before
image
image

After
image
image

### DIFF
--- a/roads.mss
+++ b/roads.mss
@@ -772,7 +772,8 @@
       }
     }
 
-    [feature = 'railway_subway'] {
+    [feature = 'railway_subway'],
+    [feature = 'railway_construction']['construction' = 'subway'] {
       #bridges {
         [zoom >= 14] {
           line-width: 5.5;
@@ -817,7 +818,7 @@
     }
 
     [feature = 'railway_disused'][zoom >= 15],
-    [feature = 'railway_construction'],
+    [feature = 'railway_construction']['construction' != 'subway'],
     [feature = 'railway_miniature'][zoom >= 15],
     [feature = 'railway_INT-preserved-ssy'][zoom >= 14] {
       #bridges {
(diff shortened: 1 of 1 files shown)


matkoniecz:
From what I see in code it affects also trams. (Changing also tramway construction bridges is not a blocker for this PR, rather an idea for a new potential one).

matkoniecz:
Fun fact: there are exactly 126 ways representing subway bridges under construction in OSM:http://overpass-turbo.eu/s/EzD

tpetillon:
Actually, a bit more. I get 261 ways by querying for other values of the key `bridge`. (https://overpass-turbo.eu/s/EzL) I admit it's still a tiny number though. But because such bridges tend to appear in larger cities and stay in that state for a few years, I figured that it may be valuable.

Indeed currently tram bridges under construction use the style of standard railways. I can include them in the changes if you want.

matkoniecz:
Right. I forgot about other values of bridge key.

And to be clear: this is a very good change, I left this comment partially because I am in the middle of reviewing this PR.

Fixing team construction bridges may be done as part of this PR or as the next one, both are perfectly fine.

matkoniecz:
Thanks for fixing this inconsistency!

In case that you are would be interested in making other improvements: https://github.com/gravitystorm/openstreetmap-carto/issues/2872 is a nice railway related one - code is relatively simple but it requires a bit thinking over what should be unified

tpetillon:
Thanks for the merge, and thanks to everyone who made it easy to setup the project. With Docker it's a breeze. :)

I'm indeed interested in making further changes. I need to familiarise myself with the code first.

matkoniecz:
@tpetillon If you are stuck with something - open an unfinished PR and ask for help, hopefully it will be something that we were stuck before and we can help :)

Issue #3565, opened 2018-12-12
historic=memorial and tourism=artwork conflicts


Janjko:
There is a problem with tags _tourism=artwork_ and _historic=memorial_. They have subtags with the same meaning: artwork has _artwork_type=sculpture;statue;bust_ and memorial has _memorial=sculpture;statue;bust_. A bust icon  ![memorial bust](https://user-images.githubusercontent.com/25656654/36630295-83cc10f6-1964-11e8-8547-276221216235.png) is rendered if you put _historic=memorial+memorial=bust_. But if you add _tourism=artwork_ on the same node, it is rendered like a generic artwork  ![tourism artwork](https://user-images.githubusercontent.com/25656654/44617373-579cfb80-a861-11e8-80d5-e70764044744.png). Even if you put:

_historic=memorial+memorial=bust+tourism=artwork+artwork_type=bust_

it still doesn't render a bust.
In reality, all memorial=bust are also tourism=artwork, so it isn't wrong tagging.

In my opinion we should first decide which tag is stronger in case of different values, _memorial=*_ or _artwork_type=*_. After that, just render whatever is in those tags, or in the stronger tag if there are both.

kocio-pl:
It is documented on the wiki, so I think it's also worth rendering:

https://wiki.openstreetmap.org/wiki/Key:artwork_type

It might be good to also review other types of artwork by the way.

jragusa:
Bust and stone are concerned. I propose to render for now `artwork_type=bust` with the current `memorial=bust` icon. It's a rapid fix. Icon of `memorial=stone` suggests an inscription so I propose for this one to wait for the rendering of natural=stone as node (#2616).

dieterdreist:
sent from a phone

> On 12. Dec 2018, at 21:49, Janjko <notifications@github.com> wrote:
> 
> They have subtags with the same meaning:


can you explain what the problem is?

geozeisig:
Using the icon ![tourism artwork](https://user-images.githubusercontent.com/25656654/44617373-579cfb80-a861-11e8-80d5-e70764044744.png) for an entire building was for me confusing. This happens when `artwork_type=architecture` is applied. An example is the [Siemens-Villa](https://www.openstreetmap.org/way/30595355#map=18/52.43883/13.32918&layers=N). Maybe you should make an exception for `artwork_type=architecture`.

jragusa:
[...]

PR #3576, opened 2018-12-17
Change icon for artwork_type=bust

jragusa:
Related to #3565

Use `bust.svg` icon for `artwork_type=bust` instead of generic artwork icon

https://www.openstreetmap.org/node/3156678281
The upper right one is `artwork_type=sculpture`

Before
bust_before

After
bust_after

### DIFF
--- a/amenity-points.mss
+++ b/amenity-points.mss
@@ -202,6 +202,9 @@
     [artwork_type = 'statue'] {
       marker-file: url('symbols/historic/statue.svg');
     }
+    [artwork_type = 'bust'] {
+      marker-file: url('symbols/historic/bust.svg');
+    }
     marker-fill: @memorials;
     marker-placement: interior;
   }


Adamant36:
Although I'm generally in support of this, it only has 790 uses. Which is something to consider. Also, what's the diff between this and memorial=bust? The wiki isn't clear on it and I can't imagine a bust that's not a memorial. Otherwise, what would be the point in it? So the tag might be redundant. The example you provided is for the philosopher Jean-Jacques Rousseau. Am I correct to assume that its a memorial bust then? If so, it might be miss-tagging.

kocio-pl:
It is sometimes much more clear if you have a statue or bust than if that is the artwork or memorial, so this is nice change.

jragusa:
Yes there is only 790 occurrences of `artwork_type=bust`, but we have also more than 3000 uses of `memorial=bust`. Since icon is already use for the latter, It's not like adding a new icon and it contributes to grow the use of this icon to about 4000 features.

IMO, the difference between artwork and memorial is like between decorative and commemorative. And, yes, `memorial` tag is probably more suitable in the above example.

Issue #3579, opened 2018-12-19
Highway=construction rendering problems


jeisenbe:
This issue is a follow-up to #2595.

1) It has been suggested that highways under construction (construction=*) should have the same width like their finished counterparts (highway=*) on a given zoom level. However, this will lead to much wider highway=construction with construction=motorway/trunk/primary-tertiary on high zoom levels. This will make the following issue more prominent

2) Dashed white and colored lines are quite prominent. This is problem is reduced at high zoom levels currently, because highway=construction is rendered much narrower than major highways (motorway thru tertiary). It is very obvious right now with highway=service vs construction=service at z14 to z16, and with highway=residential vs construction=residential. If we make the motorways thru primary wider at lower zoom levels, they will appear very prominent.

- This could be fixed for motorway thru primary by using light gray instead of white for the dashes, and adjusting the length of the dashes vs the fill color.
- However, this is harder to fix for secondary highways, where the lightness is high, and for minor highways which are rendered in white

3) Highway=construction does not follow proper layering order. Link roads (eg construction=motorway_link) can render on top of construction=motorway, and minor highways sometimes render on to of major highways. The layering order will sometimes reverse between different zoom levels.

- This would need a new SQL query, or separating teach construction=* type into a separate feature, I think?

4) Highway=raceway does not have a specific rendering for construction=raceway

- This may not be worth fixing

5) The current highway=construction style is not very intuitive. It might be considered a different road surface; for example, some topographic maps use similar patterns to represent unpaved roads (eg USGS).

jeisenbe:
Thank you for merging #3580, @matthijsmelissen
Most of the issues above are still relevant. 

Does anyone know how to fix the third problem above:
"Highway=construction does not follow proper layering order. Link roads (eg construction=motorway_link) can render on top of construction=motorway, and minor highways sometimes render on to of major highways. The layering order will sometimes reverse between different zoom levels."

I believe there is a "z order" function for highways which is actually part of the database import process. Is that correct? Is there a way to make the different types of construction=* fit into the priority system so that they layer consistently?

@pnorman?

pnorman:
> "Highway=construction does not follow proper layering order. Link roads (eg construction=motorway_link) can render on top of construction=motorway, and minor highways sometimes render on to of major highways. The layering order will sometimes reverse between different zoom levels."

If it's inconsistent between zooms, we've got a bug different than it being the wrong order.

[...]

Issue #761, opened 2014-07-25
Highway=construction with tunnel=yes


TheFive:
See

http://www.openstreetmap.org/way/235731119

There seems to be no difference for construction highways with or without tunnel tag.

polarbearing:
Apparently the tunnel in the example was finished before we could render it correctly. At least the OSM object is deleted.

jeisenbe:
I'm not sure that it's necessary to render under-construction tunnels and bridges differently. 

This could be done, while increasing the `highway=construction` code complexity somewhat. 

But it would make it less obvious that the highway is under construction, rather than being some unique type of functional road which happens to be represented with a dashed pattern.

Issue #398, opened 2014-03-12
bridge casing not rendered on highway=construction


lest69:
See here for an example: http://www.openstreetmap.org/way/265561086
I understand this was supposed to have been taken care of by an earlier commit, but these bridges are not currently rendering. Looking through roads.mss, I can't see anywhere where bridges would be rendered for highway=construction (or highway=proposed for that matter, which would also be useful).

Lukas458:
It should also be in mind that osm carto **does** render the black casing of a bridge with **railway=construction**, but not with highay=construction. The rendering is inconsequent, because if you argue with math1985s point, it could also be that it would look like only the railway track is under construction and not the entire bridge.

kocio-pl:
Sorry, his was accidental assign.

PR #3580, opened 2018-12-19
Adjust width of highway=construction rendering

### LINKED ISSUE #2595: Construction highways width optimizations
Related to #2477.

Highways under construction (`construction=*`) should have the same width like their finished counterparts (`highway=*`) on a given zoom level.
For example residential highways under construction are much wider on lower zooms than residential roads. See also the image linked in #2477.

jeisenbe:
Fixes #2595

- Stop rendering undocumented construction=* types
- Specifically render raceway, pedestrian, living_street, road as types of other construction roads
- Render minor construction=* types later
- Reduce width for narrow highway=construction, to match finished road times (or narrower)
- Add paths, track and null construction name label rendering
- Change color of construction=null and construction=path/track to #aaa, medium gray (same as minor highway=construction)

**Explanation:**
There are several issues with the current highway=construction rendering. This PR addresses several:
1. Highways under construction (construction=*) should be no wider than their finished counterparts (highway=*) on a given zoom level.
2. Highway=construction with construction=null or =road and =track, =footway, etc currently render with a dark blue-gray color, which can look quite similar to an intermittent stream or river
3. Highway=construction with construction=road, =pedestrian, =living_street and =raceway currently render starting at z12. 
- In fact, all highway=construction without construction=* are rendering at this level, and all unspecified values of construction=* get rendered at z12".
4. Name labels are not rendered for construction=track/path/<null>

There are several issues that this PR does not address, which will be discussed in #3579

I previously attempted to match the width of all roads at all zoom levels to their construction=* counterpart. However, the wider major roads at low zoom levels do not look right. Extensive adjustments would be needed to the dash pattern length, casing, and layering. 

Dashed white and colored lines are quite prominent. This is problem is reduced currently, because highway=construction is rendered much narrower than major highways. Highway=construction does not follow proper layering order. Link roads (eg construction=motorway_link) can render on top of construction=motorway, and minor highways sometimes render on to of major highways. The layering order will sometimes reverse between different zoom levels. This is much more apparent if the roads are significantly wider.

These issues would need to be addressed before trying to match the width to wider roads at high zoom levels. But I believe this PR is a good start.

**Test rendering:**

**z12** Before
z12-construction-master

After
z12-construction-after

**z13**
Before
z13-construction-master
After
z13-construction-after

**z14** Before
z14-construction-master
After
z14-construction-after

**z15** Before
z15-construction-master
After
z15-construction-after

**z16** Before
z16-construction-master
After
z16-construction-after

**z17** Before
z17-construction-master
After
z17-construction-after

**Link roads z16** before
z16-construction-link-master
After
z16-construction-links-after
- construction=secondary_link is now slightly narrower at this zoom level, to match the finished road.

### DIFF
--- a/project.mml
+++ b/project.mml
@@ -1985,9 +1985,10 @@ Layer:
         (SELECT
             way,
             highway,
+            construction,
             name
           FROM planet_osm_line
-          WHERE highway IN ('bridleway', 'footway', 'cycleway', 'path', 'track', 'steps')
+          WHERE highway IN ('bridleway', 'footway', 'cycleway', 'path', 'track', 'steps', 'construction')
             AND name IS NOT NULL
         ) AS paths_text_name
     properties:
--- a/roads.mss
+++ b/roads.mss
@@ -1189,10 +1189,13 @@ tertiary is rendered from z10 and is not included in osm_planet_roads. */
           line-width: 8;
           b/line-width: 7;
           b/line-dasharray: 8,6;
-          [construction = 'secondary_link'],
-          [construction = 'tertiary_link'] {
+          [construction = 'secondary_link'] {
             line-width: @secondary-link-width-z15;
             b/line-width: @secondary-link-width-z15 - 2 * @casing-width-z15;
+          }
+          [construction = 'tertiary_link'] {
+            line-width: @tertiary-link-width-z15;
+            b/line-width: @tertiary-link-width-z15 - 2 * @casing-width-z15;
           }
         }
         [zoom >= 17] {
--- a/roads.mss
+++ b/roads.mss
@@ -1201,6 +1201,7 @@ tertiary is rendered from z10 and is not included in osm_planet_roads. */
         }
       }
 
+      [construction = null][zoom >= 14],
       [construction = 'residential'][zoom >= 14],
       [construction = 'unclassified'][zoom >= 14] {
         line-color: @minor-construction;
@@ -1302,7 +1303,6 @@ tertiary is rendered from z10 and is not included in osm_planet_roads. */
         }
       }
 
-      [construction = null][zoom >= 15],
       [construction = 'road'][zoom >= 15],
       [construction = 'raceway'][zoom >= 15] {
         line-color: @minor-construction;
--- a/roads.mss
+++ b/roads.mss
@@ -42,7 +42,7 @@
 
 @unimportant-road: @residential-casing;
 
-@residential-construction: #aaa;
+@minor-construction: #aaa;
 @service-construction: #aaa;
 
 @destination-marking: #c2e0ff;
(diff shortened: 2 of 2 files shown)


jeisenbe:
Here is a zip of the JOSM file (.osm.xml) that I made for testing. It includes all rendered types of construction=* and highway=* at different scales. There is also an area with construction=*_link roads, and areas of different difficult landcover colors for comparison to the highways:

[test2.osm.xml.zip](https://github.com/gravitystorm/openstreetmap-carto/files/2693581/test2.osm.xml.zip)

jeisenbe:
After checking some more areas of highway=construction, I realized that highway=construction without construction=* (i.e. construction is null) should render the same as construction = residential and unclassified, rather than as construction=road. The majority of roads are highway=residential or highway=unclassified (which render the same), so it makes sense to assume that highway=construction is one of these, if not otherwise specified. I've pushed a commit to fix this. 

**z16** master
z16-lorong-halus-master
z16 previous commit
z16-lorong-halus-current
z16 new commit
z16-lorong-halus-new

**z15** Master
z15-lorong-halus-construction-roads-master
z15 Previous commit
z15-lorong-halus-construction-current
z15 New commit
z15-lorong-halus-new

**z14** Master
z14-lorong-halus-master
z14 Previous commit
z13-lorong-halus-current
z14 New commit
z14-lorong-halus-new

jragusa:
Why the bridge is in motorway colour ?

BTW, could you also add rendering of bridge (#398) and tunnel (#761) for construction roads or is it better to do it in separate PRs ?

sommerluk:
Nice to see a PR for this issue. Better rendering for construction roads would be great!

I would keep the difference between “road” and “residential”. What is rendered for construction should follow the rendering for existing roads – yet for consistency.

jeisenbe:
The current PR does render construction=road and construction=residential
differently, as shown in the test renderings above.

The change was for highway=construction without any additional tags.

jeisenbe:
Perhaps we should open a new issue about that problem?

I’ll look into it after the holiday, once this PR is closed.

Tomasz-W:
@jeisenbe If you figure something out, you can post it in https://github.com/gravitystorm/openstreetmap-carto/issues/2872

jeisenbe:
I have updated this to include the a width adjustment for
construction=tertiary_link at the relevant zoom level, after #3570 was
merged, adding tertiary_link width to rendering.

sommerluk:
> The current PR does render construction=road and construction=residential differently, as shown in the test renderings above. The change was for highway=construction without any additional tags.

Okay, got it.

 `highway=construction` without any additional tags should render like `highway=construction` with `construction=road`.

The reason is that currently we avoid yet to make a “best guess” in the code for road rendering. Example: For normal roads (no construction) with `highway=road`, we render them differently from `residential` or `unclassified`. Indeed we could also just “guess” that it is likely that they are `residential` or `unclassified` and render them like that. But we don’t because we want to give the mapper the feedback that he should use a more specific value than `road`.

Now, semantically, `highway=construction` without any additional tags is the same thing as `highway=construction` with `construction=road`: In both cases, there is a lack of information about the road type.

So, for reasons of consistency, `highway=construction` without any additional tags should render like `highway=construction` with `construction=road`.

jeisenbe:
I thought that too, at first, but then I realized that many mappers use
just highway=construction without additional classification. This is
relatively much more common than highway=road.

The renderings did not work well with the width and light gray of
highway=road. The current commit is a less dramatic change from the current
rendering, and looks more harmonious with the other types of construction=*

sommerluk:
> many mappers use just highway=construction without additional classification. This is relatively much more common than highway=road.

Maybe, but that’s not a good argument to treat it differently. Especially as your PR provides the much nicer rendering for these cases as the current one.

dieterdreist:
sent from a phone

> On 27. Dec 2018, at 12:18, jeisenbe <notifications@github.com> wrote:
> 
> I thought that too, at first, but then I realized that many mappers use
> just highway=construction without additional classification. This is
> relatively much more common than highway=road.


it is something different, highway=road is kind of a fixme (no road class), highway=construction is for roads under construction

jeisenbe:
@matthijsmelissen could you take a look at this one? I believe it is ready to merge as-is, but I'd also be happy to change the rendering for generic highway=construction if needed.

matthijsmelissen:
Thanks, great work!

jeisenbe:
> > many mappers use just highway=construction without additional classification. This is relatively much more common than highway=road.
> 
> Maybe, but that’s not a good argument to treat it differently. Especially as your PR provides the much nicer rendering for these cases as the current one.

I don't quite understand this comment, I don't think the new rendering is "much nicer", though it is a minor improvement. 

Currently highway=construction is rendered the same as construction=residential, and it uses a blue-gray dash pattern. This PR will change both highway=construction and construction=residential to render a little thinner at z14, z15 and z16 to match highway=residential, but the width is unchanged at <= z13 and >= z17. The color of the dashes is changed from blue-gray to neutral gray.

sommerluk:
> I don't think the new rendering is "much nicer", though it is a minor improvement.

Well, you’ve changed the colour from blue-ish to grey-ish, and that’s good!

Issue #1490, opened 2015-04-17
Water parks areas should be rendered


daganzdaanda:
Issue #370 was closed by pull #725, where leisure=water_park got an icon. I think it should also get a fill. 
Possible colours:
- like landuse=meadow (since there are often meadows along the pools)
- something like sports_centre (maybe rethink that strong green at the same time...)

mboeringa:
I don't think this should have a fill, but some kind of outline only. It is a kind of "administrative" border, not a visible feature on the ground (In my personal ArcGIS Renderer, I render it as such: with just an outline). And yes, this is similar for sports_centres and many other OSM features...

matkoniecz:
Can you give an example of location where fill would be useful?

mboeringa:
@matkoniecz: Here is an example rendering of waterpark in Paris to show you why I think using an outline may be more appropriate. As you can see, there is other stuff being tagged within the outline of the waterpark. Setting a fill could only be done if it is one of the "lower" layers, drawn below most other stuff, but I think an outline is probably best, like with **leisure = marina** or **tourism = theme_park**. These features can be quite big as well.

![waterpark](https://cloud.githubusercontent.com/assets/7635726/7490108/cf9b9f38-f3dd-11e4-8635-dd960868c161.png)

daganzdaanda:
> Can you give an example of location where fill would be useful?

https://www.openstreetmap.org/way/70234702 used to be a leisure=water_park, and someone turned it into leisure=park, probably because it makes the area visible (as someone speculated in a note about it).
Here http://www.openstreetmap.org/way/127430479 there's also "surface=grass" tagged, which doesn't  help to make the area green (anymore?), and is actually wrong since the area also contains buildings and the pools...
Checking overpass for water_parks shows a lot of them that look "wrong" because there is nothing to show the extent of them. 

But I do agree with @mboeringa that an outline could do the same as a fill, or even more since then people might start to fill the outline with micromapped meadows and trees and so on.

KaiRo-at:
I've seen water parks getting a landuse=recreation_ground just so that it gets rendered as green on the map (given that most of the surface is usually grass and recreational stuff is mostly green in our scheme, that color sounds right anyhow).
Examples are http://www.openstreetmap.org/way/24443459 or http://www.openstreetmap.org/way/24444267 around my home town.

kocio-pl:
+1 for the outline.

[...]

PR #3582, opened 2018-12-20
Change tourism outline color

jeisenbe:
Related to #3045 "Outline Coordination Meta Issue"
and #3521 "Add Aboriginal Areas"

- Change tourism color from brown to `#660033` - this is currently the tourism=attraction text color

**Explanation:**
In #3521 it is planned to use brown for boundary=aboriginal_lands (eg American Indian reservations)
Therefore it is necessary to change the color of the outline used for zoos and theme parks.
This has also been desired, so that these tourist areas would be more visible on the map with a color more closely related to retail areas and other points of interest. The proposed change will also unify the color used for these two features with that used for tourism=attraction.

Purple and orange have also been discussed. Purple would work well, but would be more similar to the current administrative boundary rendering. Orange is another option.

Test renderings:

**Singapore zoos**
https://www.openstreetmap.org/#map=13/1.3950/103.8424
z17 Before
z17-singapore-zoo-master
z17 After
- Attractions within the zoom now are shown in the same color
- I plan a PR to add generic icons (dots) for tourism=attraction
z17-singapore-zoo-660033

z14 Before
- Can compare to military areas here
z14-singapore-zoo-master
z14 After
z14-singapore-zoo-660033

z13 Before
- Compare to the international administrative boundary at this level (top of image)
- Could we show zoos/theme parks with slightly smaller way_pixels at this zoom level?
z13-zoos-master
z13 After
z13-zoos-660033

**Universal Studios Singapore**
- The boundaries are hard to see in both the before and after, because the area has been (mis)-tagged with leisure=recreation_ground in addition to tourism=theme_park
https://www.openstreetmap.org/#map=16/1.2546/103.8293
z16 before
z16-universal-studios-master
z16 after
z16-universal-studios-660033

z17 before
z17-universal-studios-master
z17 after
z17-universal-studios-660033

**Jurong Bird Park, Singapore**
- Near industrial and commercial landuse
https://www.openstreetmap.org/#map=15/1.3205/103.7153
z15 before
z15-jurong-bird-master
z15 after
z15-jurong-660033

### DIFF
--- a/amenity-points.mss
+++ b/amenity-points.mss
@@ -2413,7 +2413,7 @@
         text-face-name: @bold-fonts; /*rendered bold to improve visibility since theme parks tend to have crowded backgrounds*/
       }
       [feature = 'tourism_attraction'] {
-        text-fill: #660033;
+        text-fill: @tourism;
         text-face-name: @standard-font;
       }
       [feature = 'amenity_kindergarten'],
--- a/landcover.mss
+++ b/landcover.mss
@@ -51,7 +51,7 @@
 @power-line: darken(@industrial-line, 5%);
 @sand: #f5e9c6;
 @societal_amenities: #ffffe5;   // Lch(99,13,109)
-@tourism: #734a08;
+@tourism: #660033;
 @quarry: #c5c3c3;
 @military: #f55;
 @beach: #fff1ba;


matkoniecz:
> Could we show zoos/theme parks with slightly smaller way_pixels at this zoom level?

If you think that it is desirable - please open a new issue (if it is not existing already), it will be missed otherwise.

Tomasz-W:
I prefer orange over violet (see https://github.com/gravitystorm/openstreetmap-carto/pull/3521#issuecomment-448830637)

kocio-pl:
Violet outline is not used for industrial areas (which BTW use lighter shade of violet anyway), so we reuse colors in a sane way. It also aligns nicely with attractions rendering, which makes it more consistent. Orange looks much lighter, so it could be used for something more "calm" than amusement.

jeisenbe:
Here are some test images with the other color options. I think the current commit – with #660033 – is the best option. The dark violet option would probably require changing the the administrative border color, and lighter violet or purple would certainly require it.

**Orange:** `#C77400`
Zoos z14
z14-singapore-zoos-orange
z17
z17-singapore-zoo-orange

Bird Park z15
z15-jurong-orange

Universal Studios z16
z16-universal-studios-orange
z17
z17-universal-studios-orange


**Dark Violet:** `#864784`
Zoos z14
z14-singapore-zoo-darkviolet
z17
z17-universal-studios-darkviolet

Bird Park z15
z15-jurong-bird-darkviolet

Universal Studios z16
z16-universal-studios-darkviolet
z17
z17-universal-studios-darkviolet

(Thank Telkomsel for the great internet connection today)

Adamant36:
Id go with violet. Orange just doesn't make sense and violet looks better. Although it could be because of the wrong park tagging color in the examples. Which go better with violet.

matkoniecz:
Violet for me looks too close to administrative boundaries.

Adamant36:
@matkoniecz, is that an endorsement of orange then?

kocio-pl:
@matkoniecz For me the difference is visible (this violet has more red shade and different line pattern), but I hope #3553 will make it even more clear soon. Unfortunately it makes the code dependency chain longer and more fragile (first #3553, then #3582 and after that #3521...).

jeisenbe:
@matkoniecz do you think #660033 will work with the current admin
boundaries?

I believe this color is sufficiently dissimilar.

turnsole80:
I'd definitely go with violet over orange.

jragusa:
Why about adding `leisure=water_park` (#1490) to the tourism outline ?

kocio-pl:
Thanks! I believe this is different enough from current border rendering and it's better to not hold the #3521 any longer.

polarbearing:
Would it make sense to move accommodation from the current transport-blue to the new tourism-violet?

kocio-pl:
I'm afraid it's too close to the shops when using icons.

polarbearing:
```#660033``` tourism
```#ac39ac``` shop-icon

Issue #3326, opened 2018-08-03
Change some amenities colour to grey


Tomasz-W:
I propose to change colour of some non-cultural, non-service, free standing objects from amenity-brown to man-made-grey. Amenity-brown is heavy loaded right now and it includes tags from many different categories. As we already render towers, windmills or crosses in man-made-gray, and will render tourist information objects too, I think we should use this colour also for:
* `amenity=bench`
* `leisure=picnic_table`
* `amenity=waste_basket`
* `amenity=waste_disposal`
* `amenity=recycling`

kocio-pl:
At the moment the difference is more or less whether the objects are usable (brown) or are mainly landmarks (gray), but that might make sense to "demote" some less important, but usable objects to gray to just offload brown.

Adamant36:
We have a lot of cement bench's and picnic tables where I live that know one really uses and serve more like landmarks or markers for interesting places. So I sort of like this idea.

kocio-pl:
It's possible to do it, but at the moment I'm not focused on this problem.

Adamant36:
@Tomasz-W, as it currently stands both the general amenity=recycling and recycling_type=centre is rendered the same way and the code is connected. Recycling_type=container is not rendered at all though. What I think we should do is that recycling_type=centre should remain as amenity brown and maybe rendering its area could happen also. Then add rendering for recycling_type=container in grey. Which might spur mappers to use the recycling type tags more. Plus, also leave the general amenity=recycling as amenity brown but make it a separate feature in the code from the other two. What do you think?

Adamant36:
Some test renders of everything except amenity=recycling and amenity=shelter
![benches and waste baskets in park](https://user-images.githubusercontent.com/30259065/45605403-9dc52500-b9f1-11e8-933f-622bc6d880dc.png)
![picnic tables park](https://user-images.githubusercontent.com/30259065/45605406-a289d900-b9f1-11e8-9c2c-10ae2ce35b2c.png)
![picnic tables retail area](https://user-images.githubusercontent.com/30259065/45605408-a4ec3300-b9f1-11e8-84e1-0c76fa5e5070.png)
![waste disposel grass](https://user-images.githubusercontent.com/30259065/45605410-a74e8d00-b9f1-11e8-888d-8dc022c0f3e5.png)
[...]

PR #3586, opened 2018-12-21
Change certain amenity icons to grey

Adamant36:
This PR changes certain amenity icons that are passive to grey in order to help off load things color with amenity brown a little and to reflect the fact that they aren't "services." Included in the PR is bench, picnic table, waste basket, shelter, and waste disposal. Not included in the PR is things related to recycling. Which will be changed in a separate PR. The PR is related to PR #3402 and issue #3326. 

bench, waste basket, picnic table
https://www.openstreetmap.org/#map=19/40.68154/-122.36317 (node)
bench waste basket picnic table
shelter
https://www.openstreetmap.org/#map=19/37.79619/-122.42138 (node)
shelter node
https://www.openstreetmap.org/#map=19/40.58366/-122.39271 (way)
waste disposal
https://www.openstreetmap.org/#map=19/37.48214/-122.22696 (node)
waste disposal node
https://www.openstreetmap.org/#map=19/37.49039/-122.22382 (way)
waste disposal way

### DIFF
--- a/amenity-points.mss
+++ b/amenity-points.mss
@@ -49,7 +49,7 @@
       marker-file: url('symbols/tourism/alpinehut.svg');
     }
     [feature = 'amenity_shelter'] {
-      marker-fill: @amenity-brown;
+      marker-fill: @man-made-icon;
     }
     marker-fill: @accommodation-icon;
     marker-placement: interior;
@@ -1427,7 +1427,7 @@
 
   [feature = 'leisure_picnic_table'][zoom >= 17] {
     marker-file: url('symbols/tourism/picnic.svg');
-    marker-fill: @amenity-brown;
+    marker-fill: @man-made-icon;
     marker-placement: interior;
     marker-clip: false;
     [access != ''][access != 'permissive'][access != 'yes'] {
@@ -1615,7 +1615,7 @@
     [access = 'permissive'],
     [access = 'yes'] {
       marker-file: url('symbols/amenity/waste_disposal.svg');
-      marker-fill: @amenity-brown;
+      marker-fill: @man-made-icon;
       marker-placement: interior;
     }
   }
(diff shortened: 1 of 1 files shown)


Tomasz-W:
@Adamant36 Thanks! I hope you will take care of grey recycling containers and brown recycling centres later :)

Adamant36:
Me neither. I feel like this type of thing has changed some since the original issue was opened. I'm not sure how though. 

@Tomasz-W, what do you think?

Tomasz-W:
> I have the feeling that "service" is everything you can interact with, not just observe. So for example you can sit on bench and throw some trash into the waste basket.

I totally agree. And that's the reason why in my opinion objects from this PR should be in man-made grey. If you kick a bench, it won't kick you back. If you throw a trach into a waste basket, it won't say you 'thank you" - there is no inteaction. As e.g. vending machines are mostly electronic I consider them as interactive so they should be in amenity-brown.

Adamant36:
Looking through the wiki hunting stand is the only similar one I could find, except maybe bbq, and its in grey. Bbq is brown, but its kind of passive isnt it? There's also fountains, but they are in blue.

matthijsmelissen:
Agree, was feeling the same for a long time already. Thanks!

Adamant36:
Yep. Thanks for merging it.

Tomasz-W:
@Adamant36 Do you need a ticket for different colours for recycling containers and centres or you will add it to yours some personal tasklist note?

Adamant36:
If you want to open a new ticket for it that would be great. I'll probably forget about it otherwise.

Tomasz-W:
Actually, https://github.com/gravitystorm/openstreetmap-carto/issues/3326 might be that ticket. I'll left it unchecked on Good first issues list.

PR #3598, opened 2018-12-26
Fix inconsistency of leisure=ice_rink

### LINKED ISSUE #3597: Inconsistency between polygon anf label  of leisure=ice_rink

jragusa:
Fixes #3597

remove variable `[way_pixels > 3000]` to allow correct display of `leisure=ice_rink` from z10

https://www.openstreetmap.org/way/87188308
Before
icerink_2_z17
After
icerink_2_z17_after

### DIFF
--- a/landcover.mss
+++ b/landcover.mss
@@ -269,7 +269,7 @@
   }
 
   [feature = 'leisure_ice_rink'][is_building = 'no'] {
-    [zoom >= 10][way_pixels > 3000] {
+    [zoom >= 10] {
       polygon-fill: @glacier;
       line-width: 0.5;
       line-color: saturate(darken(@pitch, 30%), 20%);


jeisenbe:
I’d recommend changing this PR to keep a way_pixels limit (perhaps reduced
to 900) for z10 to z13 and only show ice rinks of any size at >= z14

polarbearing:
I was wondering myself, and found that ```way_pixels > 3000``` is used on many text labels in lower zoom levels, apparently to prevent that a tiny area is cluttered with a label. 

We do not use ```way_pixels > 3000``` on any other polygon rendering in landcover.mss. 
We do have, for some 'important' objects (fire station, police), an 'early rendering' when they have ```way_pixels > 900```.

Thus I conclude that the 3000 was introduced accidentally by copy/paste from the label code in this case, and it can safely be removed as this PR proposes.

polarbearing:
@jeisenbe - what would be the reason for that? The only way-size exceptions I found currently in use are the fire/police for z8. 
```leisure=sports_centre``` is shown from z10 without restrictions, why not the ice_rink?

jeisenbe:
Ok, I see what you mean. I was thinking how the pattens only render from
z13 or z14.

It would be interesting to review this for all of the small urban features
like pitches, tracks, parking etc;

Perhaps they should instead be rendered as “built-up-lowzoom” at z10 to
z12, especially if smaller than 64 way_pixels, to avoid adding
high-frequency noise to the map.

PR #3599, opened 2018-12-26
Add support of parking=multi-storey to amenity=parking_entrance

### LINKED ISSUE #839: Add rendering for multistory parking icons
Multistories should be rendered with a P and roof.

See also https://trac.openstreetmap.org/ticket/3470.

### LINKED ISSUE #2586: Parking node without icon when housenumber is tagged
When an addr:housenumber is tagged, nodes with amenity=parking show the housenumber instead of the "P" icon. It would make more sense to prioritise the icon, as is done when the parking is mapped as a way. 

Examples:
http://www.openstreetmap.org/node/1540323229
http://www.openstreetmap.org/node/4035219789
http://www.openstreetmap.org/node/795888551
http://www.openstreetmap.org/node/3918195757

jragusa:
Fixes #839
Fixes #2586

* Render `amenity=parking_entrance` when `parking=multi-storey`
* This modification additionally resolves incoherent rendering of `addr:housenumber` tag.

https://www.openstreetmap.org/node/2493312559
Before
parking_multistorey_before
After
parking_multistorey_after


kocio-pl:
I think it makes sense, I just have to think a bit more if the same icon as for underground parking is OK.

kocio-pl:
@matthijsmelissen What's your opinion on that?

Tomasz-W:
@jragusa What about horizontal arrow for them? First meters of multistory parking entrances usually go staight, not down as underground ones.

svg amenity parking_entrance multi_storey

Gist link: https://gist.github.com/Tomasz-W/08b64a21cb537935f7cbe45238a35215

matthijsmelissen:
Why should we restrict it to underground and multi-storey parkings? Surface parking can also have a dedicated entrance.

https://www.travelaroundwithme.com/wp-content/uploads/gratis-parkeren-winter-scheveningen-zwarte-pad.jpg

Are there many examples of amenity=parking_entrance for surface parking that shouldn't be rendered? If not, I think I'd prefer to render all parking_entrances.

I'm marking this as [WIP] for now awaiting a response from @jragusa - please feel free to remove the [WIP]-tag when responding.

jragusa:
IMO, it's more easy to find the entrance of a surface parking since parking_aisle are rendered. Lets inverse the reasoning and imagine that we don't have any icon for parking_entrance. It remains easy to identify entrance of surface parking than those of underground/multi-storey parking.

Moreover, displaying two very similar icons for the same feature is useless and could clutter the map.

polarbearing:
> entrance of a surface parking since parking_aisle are rendered

service=parking_aisle is not supposed to be used for the entrance way to a car park.

polarbearing:
I just checked the wiki page for [parking_entrance](https://wiki.openstreetmap.org/wiki/Tag:amenity%3Dparking_entrance). 
It does not even mention any combination with parking=*
I see it problematic to encode a rendering rule for which the tagging is not even documented.

geozeisig:
@polarbearing I added it. `parking `was missing

polarbearing:
Fine thanks. It is used on entrances sufficiently to be documented:
parking=underground  15000
parking=multi-storey 1800
parking=surface surface

imagico:
The rendering of amenity=parking_entrance + parking=underground was added since amenity=parking + parking=underground is not being rendered.  Rendering multi-storey parkings the same way would kind of defeat the purpose of this.

I could imagine it making sense to render the entrances of multi-storey parkings in addition to the parking itself but this would IMO require a much less bold symbol, something like:

http://www.imagico.de/files/parking_small_horizontal.svg

and not rendering names (since they will most likely duplicate the name of the parking).

One problem with both underground and multi-storey parkings is that the tag does not seem to be exclusively meant or used for car entrances into the parking, i.e. also for exits and foot entrances.  This makes a differentiated and useful display more difficult.

jragusa:
1/ I'm not against to use a slightly different icon for multi-storey.
2/ About foot entrances, why not use the same icon than `railway=subway entrance` ? So we can distinguish both entrances.

jeisenbe:
@jragusa - were you considering updating this PR with the changes suggested: Smaller icon and no text label?

I agree with you that it's not really necessary to render a special icon for surface parking entrances, where the service roads can be mapped and rendered.

jeisenbe:
@jragusa - do you have any time this month to update the PR?

jragusa:
@jeisenbe what do you mean by smaller icon exactly ?

What about the icon suggested [here](https://github.com/gravitystorm/openstreetmap-carto/pull/3599/#issuecomment-451116029) by @Tomasz-W ?

jeisenbe:
> smaller icon

I was summarizing the comment [above](https://github.com/gravitystorm/openstreetmap-carto/pull/3599#issuecomment-480501239):

"I could imagine it making sense to render the entrances of multi-storey parkings in addition to the parking itself but this would IMO require a much less bold symbol, something like: http://www.imagico.de/files/parking_small_horizontal.svg
and not rendering names"

> What about the icon suggested [here](https://github.com/gravitystorm/openstreetmap-carto/pull/3599/#issuecomment-451116029)

That icon is full size, 14px tall

jragusa:
Yes I guess it's ready

project.mml
```diff
@@ -2195,6 +2195,7 @@ Layer:
               'barrier_' || CASE WHEN barrier IN ('bollard', 'gate', 'lift_gate', 'swing_gate', 'block', 'log', 'cattle_grid', 'stile', 'motorcycle_barrier', 'cycle_barrier', 'full-height_turnstile', 'turnstile', 'kissing_gate') THEN barrier ELSE NULL END
             )  AS feature,
             access,
+            tags->'parking' as parking,
```
jeisenbe: You also need to add this to the `amenity-points` layer, since `[parking='underground']` is used in line 2927 of `amenity-points.mss` for the text, but parking features are no longer in `amenity-low-priority`. Right now I get the error `column "parking" does not exist`.

jeisenbe:
@imagico - I agree with your [comment above](https://github.com/gravitystorm/openstreetmap-carto/pull/3599#issuecomment-480501239) that the icon could be lighter. However, the existing parking_entrance icon is equally heavy, and the main parking "P" icon is probably too large and bold as well. 

Should we expect @jragusa to change both in this PR, or would it be reasonable to use an icon similar to those shown in the most recent test images, to match the current parking_entrance icon (used for underground parkings), and then suggest another PR to update all the parking icons to be less bold?

Examples after:
parking_entrance_after

For comparison, here's a named `parking_entrance` + `parking=underground`:
z18-theatergarage-parking-entrance-underground

jeisenbe:
Also @jragusa the icon is not rendering for me, even after fixing the line above. Could you check the test rendering again on your system? Perhaps there is a problem with the svg file?

jeisenbe:
Ping @jragusa - there are a couple things that need to be fixed

jragusa:
My apologies, @jeisenbe. I was a bit busy last weeks. I'm working on it.

jragusa:
I corrected the svg file (initially saved in inkscape svg) and I update the branch since several modifications were not recorded in my project file. 

I think now that everything is ok

symbols/amenity/parking_entrance_multistorey.svg
```diff
@@ -0,0 +1,36 @@
...
+    <rdf:RDF>
+      <cc:Work
+         rdf:about="">
+        <dc:format>image/svg+xml</dc:format>
+        <dc:type
+           rdf:resource="http://purl.org/dc/dcmitype/StillImage" />
+        <dc:title></dc:title>
+      </cc:Work>
+    </rdf:RDF>
+  </metadata>
```
jeisenbe: This svg file (`parking_entrance_multistorey.svg`) could be optimized by removing the unused stroke-width and fill rules, and the metadata. If you are using Inkscape, open the svg and then save as "optimized svg" format. Select the options to remove metadata and unused rules. This will create a much shorter file, like this:

```
<?xml version="1.0" encoding="UTF-8"?>
<svg width="14" height="14" version="1.1" viewBox="0 0 14 14" xmlns="http://www.w3.org/2000/svg">
 <path d="m0 0v10h2v-4h2.3807c3.5045-0.023785 3.5045-6 0-6zm2 2.0115v1.9885l2-2.255e-4c1.5622 2.255e-4 1.5728-1.9998 0-2.0115z" fill-rule="evenodd"/>
 <path d="m5 11h5v3l4-4-4-4v3h-5v2z"/>
</svg>
```

jragusa: Done :)

I made a confusion between simple svg and optimized svg

jeisenbe:
Here is a test area comparing all the current and proposed parking icons:

parking-icon-examples

And this would be with the suggestion by @imagico with the smaller icon:

parking-entrance-small-test

jeisenbe:
https://www.openstreetmap.org/#map=18/53.07323/8.83855
This PR: 
z18-klinik-parking-entrance-icon-new

Compared with the smaller icon suggestion:
z18-parking-small-icon-klinik-18:53 07287:8 83851
Perhaps not legible enough?

kocio-pl:
@Tomasz-W What do you think about this proposition? Would you like to help tuning these icons?

jeisenbe:
@kocio-pl what are your thoughts about the proposed icon (and the current parking_entrance icon)? Do you think they are too bold or large, as mentioned by @imagico previously?

kocio-pl:
Sorry, Joseph, I'm reluctant to engage in yet another discussion - and I never know which small detail will lead to a big talk, especially after we spent a lot of time on icons. The problem here was that while they might be smaller, we wanted to retain readability and uniformity of symbols.

So just a short remark: I don't see these icons are any smaller. The one with horizontal arrow is smaller, but it's strange that it's smaller than the one with the arrow down, so some explanation is needed for this difference (or a further work).

jeisenbe:
I was asking because I wanted to know if you disapproved of this PR. I have already marked it as "approved" 3 days ago, with this icon:

z18-klinik-parking-entrance-icon-new

Since you asked @Tomasz-W about the icon, I want to make sure that you agree with merging this PR as-is. 

While I think that all of the parking icons might be improved by re-working them to be more consistent and a little less bold, I think that can be a separate PR, so I am willing to accept and merge this PR with the current commit, if you and @imagico (who has previously commented) agree.

kocio-pl:
Oh, now I see. I prefer waiting for @Tomasz-W answer for a few days. Feel free to nudge me if I forget to answer then.

jeisenbe:
Could you formally "request changes" to explain what needs to be changed with the proposed icon prior to this PR being merged?

symbols/amenity/parking_entrance_multistorey.svg
```diff
@@ -0,0 +1,14 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
kocio-pl: Let's wait for @Tomasz-W answer for a few days on the icon design.

jeisenbe:
It's been a month since this PR was approved. @kocio-pl was interested in seeing another icon option but that has not happened. Can it be merged now?

kocio-pl:
I guess it was enough of waiting for, let's go with the current icon.

PR #3600, opened 2018-12-26
Hide underground parking

### LINKED ISSUE #3506: Hide underground parkings
Continuation of https://github.com/gravitystorm/openstreetmap-carto/issues/2904 discussion since https://github.com/gravitystorm/openstreetmap-carto/issues/2904#issuecomment-346001797.

Proposition is to render underground parkings (`amenity=parking` + `parking=underground`) only as icon of "P" with down arrow.

Test renderings provided by @kocio-pl : 

Before:

After:

https://www.openstreetmap.org/way/235550028

Gist link: https://gist.github.com/Tomasz-W/b9655543be00a01d51c975b03009c654 -> v5

jragusa:
Fixes #3506

Exclude parking=underground from the rendering. Only `amenity=parking_entrance` should be displayed to indicate the entrance of these parkings.

https://www.openstreetmap.org/way/235550028
https://www.openstreetmap.org/way/231387829
Before
parking_underground_before

After
parking_underground_after


polarbearing:
How does this interfere with the current development in rendering underground buildings #552 ?

kocio-pl:
Well, it's not clear for me if these should be tagged as buildings or not. In the example cases they're not, so it does not interfere.

kocio-pl:
This code needs to allow also lack of `parking=*` tag. Currently simple `amenity=parking` does not show.

jragusa:
I added condition when parking is NULL. I tested and they are rendered now

literannn:
What about parking spaces (amenity=parking_space) on underground parkings? I suppose, we should hide them too. For example, we have many such items in Zaryadye park in Moscow: https://www.openstreetmap.org/#map=19/55.75122/37.63031

kocio-pl:
Yes, I was reporting this in https://github.com/gravitystorm/openstreetmap-carto/issues/3506#issuecomment-437602119. Too soon to consider problem solved, so I will reopen the issue.

jragusa:
What would be the rule : only consider `amenity=parking_space` with `parking=surface/rooftop` ?

kocio-pl:
I believe the problem is only with underground parkings, so I would not be so strict when only ~11% has `parking=*` subtag at all:

https://taginfo.openstreetmap.org/tags/amenity=parking_space#combinations

polarbearing:
I still don't understand the logic of hiding underground parking, while rendering underground buildings ( #552 ). Underground parking _is_ an underground building.

jragusa:
IMO, it's useless to display 2 icons for underground parking (`amenity=parking_entrance` and `amenity=parking`) and the most important one is the entrance not the parking itself.

polarbearing:
When people start mapping parking_entrance for surface parking as well, you'd have the same situation there.

jaffroy62:
Hi Dear, since this change #3600 the Rendering for the entrance to the underground parking (amenity=parking_entrance + parking=underground) vanished - was this also intended? 

I liked this feature very much :-)

Yours, Andy

kocio-pl:
Could you post a link where they are missing? This was intended to implement them first before doing anything with underground parkings and it looks that they are still there:

screenshot_2019-02-02 openstreetmap

jaffroy62:
Hi, thank You for checking this!

Analyzing myand your example: could be the "access=private" the reason? 
http://www.openstreetmap.org/node/4360517642
image


On the other hand this is an access to the underground parking via steps and not a vehicle access.

In your example: https://www.openstreetmap.org/node/4159128450#map=19/52.23873/21.01572
image

There is of course no "access=private"

Would it be an idea to render these with access=private in a lighter color/saturation?

Yours Andy

kocio-pl:
Yes, we were avoiding private entrances to nor clutter the map.

jaffroy62:
in my opinion this would not clutter but rather add useful information. however, thank you for clarification.

kocio-pl:
If this is an issue for you, please open new ticket so it could be discussed in a clear way.

Issue #1257, opened 2015-01-25
Add area render for tourism=attraction


matthijsmelissen:
@gravitystorm suggests [here](https://github.com/gravitystorm/openstreetmap-carto/pull/1063#issuecomment-67853592) about tourism=attraction:

> [..] we'll need to consider some sort of area rendering in the future otherwise we end up with floating labels with no indication what extent they refer to.

mkoniecz:
Where it would be useful? in my experience tourism=attraction is applied to very small object where rendering area is not useful or objects that are rendered anyway.

The only use-case where it maybe would be useful is Speakers' Corner in Hyde Park.

moliha:
Below are two examples of attractions that are areas, shown with:  
- The current grayish Mapnik-rendering  
- The current redish Cycle-rendering (and former Mapnik-rendering): 

"Den Fynske Landsby" in Odense, Denmark:  
- http://www.openstreetmap.org/#map=16/55.3647/10.3835  
- http://www.openstreetmap.org/#map=16/55.3647/10.3835&layers=C  

"Den Gamle By" in Aarhus, Denmark:  
- http://www.openstreetmap.org/#map=17/56.15878/10.19169  
- http://www.openstreetmap.org/#map=17/56.15878/10.19169&layers=C

The redish rendering makes it easy to see the extent of the attraction-area.

With the current grayish Mapnik-rendering it can be difficult to distinguish the attraction-area from some of the surrounding areas.

I think there are many such cases where a distinct coloring of atraction-areas is useful.

One additional point: the redish color used for attractions by Cycle and formerly by Mapnik is perhaps too close to the color used for retail and industrial areas. Some adjustnents of these colors could also be useful.

Note: In case the rendering is changed,images from the above links can be seen at  
http://mlha.dk/osm/attraction.htm

mboeringa:
> The only use-case where it maybe would be useful is Speakers' Corner in Hyde Park.

That is an extremely bold statement that unfortunately doesn't hold up. Look at this image of Berlin city centre made with my OSM Renderer for ArcGIS. Just in this small section of the German capital alone, there are at least 7 different **non-**building features tagged with **tourism=attraction**, including really prominent ones like:

- Museuminsel
- Nikolaiviertel
- M. Heckmann hofe
- Hackeschermarkt

There are *many* of such non-building tourist marked features world wide...

Please click the image to see it properly enlarged.

[...]

PR #3603, opened 2018-12-27
Reduce priority of tourism=attraction and render from z17

### LINKED ISSUE #3545: "tourism" refrains "man_made" node from rendering "name"
### Expected behavior

Related to: https://www.openstreetmap.org/node/1980961590

Combining `man_made=cross` and `tourism=attraction` should not refrain the feature from rendering its `name` beneath the cross icon.

### Actual behavior

See history: https://www.openstreetmap.org/node/1980961590/history

At the version 1, the node had only `tourism=attraction` and `name=Morro da Santa`. I remember its name rendered at the node.

Once I added `man_made=cross`, it started to render the cross icon but, on the other hand, its name no longer renders.

Now, compare with https://www.openstreetmap.org/node/5185486387 which has `man_made=cross` and `name`. Its name is rendered beneath the cross icon. My guess is that, the problem is on `tourism=attraction` side.

map

jeisenbe:
Fixes #3545
Related to https://github.com/gravitystorm/openstreetmap-carto/pull/1063#issuecomment-67853592

- Add generic "dot" icon for tourism=attraction
- Render tourism=attraction only at z17 and above
- Move tourism=attraction to low-priority label, after other features, so that it will only render if there is no other more specific tag (this fixes #3545 "tourism=attraction stops man_made=cross from rendering")
- Change text to standard font, render text and dot in tourism color ( `#660033` now)

Explanation:
- Features tagged tourism=attraction will currently render earlier than z17 if tagged on a large area (closed way or multi polygon), however there is no area fill or icon
- Area rendering was previously rejected in #1257 "Add area rendering for tourism=attraction"
- This PR will add a generic icon (a small circle) to show the location of the node or center of the area
- Large areas will no longer get a label before z17, because this invites misuse of the tag to get a text label to render on features that would otherwise not be rendered
- Generally we should avoid showing text name labels without an icon, outline or area fill, because this can lead to mistagging for rendering. This is particularly prone to vandalism when it can be used to render text at very early zoom levels (eg before z15)
- The feature is moved to the low-priority layer, and last in the SQL query, so that it will only be rendered if it is the only tag on the feature. This fixes a problem where wayside crosses and man_made=cross that are also tagged tourism=attraction do not render properly.

**Singapore zoos**
http://openstreetmap.org/#map=16/1.4034/103.7966
z16 - unchanged before/after
z16-zoos-after
z17 Before
z17-singapore-zoo-660033
z17 After
z17-zoo-after

**Universal Studios and Kidzania**
http://openstreetmap.org/#map=17/1.25438/103.82532
z17 before
z17-universal-studios-660033
z17 after
z17-universal-kidzania-after

### DIFF
--- a/amenity-points.mss
+++ b/amenity-points.mss
@@ -1728,13 +1728,6 @@
     }
   }
 
-  [feature = 'tourism_attraction'][zoom >= 17] {
-    marker-width: 4;
-    marker-line-width: 0;
-    marker-fill: @tourism;
-    marker-placement: interior;
-  }
-
   // waste_disposal tagging on nodes - tagging on ways is defined earlier
   [feature = 'amenity_waste_disposal'][zoom >= 19]::amenity {
     [access = null],
@@ -3234,7 +3227,6 @@
     text-size: @standard-font-size;
     text-wrap-width: @standard-wrap-width;
     text-line-spacing: @standard-line-spacing-size;
-    text-dy: 6;
     text-fill: @tourism;
     text-halo-radius: @standard-halo-radius;
     text-halo-fill: @standard-halo-fill;
--- a/amenity-points.mss
+++ b/amenity-points.mss
@@ -1728,6 +1728,13 @@
     }
   }
 
+  [feature = 'tourism_attraction'][zoom >= 17] {
+    marker-width: 4;
+    marker-line-width: 0;
+    marker-fill: @tourism;
+    marker-placement: interior;
+  }
+
   // waste_disposal tagging on nodes - tagging on ways is defined earlier
   [feature = 'amenity_waste_disposal'][zoom >= 19]::amenity {
     [access = null],
@@ -2362,7 +2369,6 @@
   [feature = 'landuse_construction'],
   [feature = 'tourism_theme_park'],
   [feature = 'tourism_zoo'],
-  [feature = 'tourism_attraction'],
   [feature = 'amenity_kindergarten'],
   [feature = 'amenity_school'],
   [feature = 'amenity_college'],
@@ -2474,10 +2480,6 @@
       [feature = 'tourism_zoo'] {
         text-fill: @tourism;
         text-face-name: @bold-fonts; /*rendered bold to improve visibility since theme parks tend to have crowded backgrounds*/
-      }
-      [feature = 'tourism_attraction'] {
-        text-fill: @tourism;
-        text-face-name: @standard-font;
       }
       [feature = 'amenity_kindergarten'],
       [feature = 'amenity_school'],
@@ -3225,6 +3227,19 @@
     text-halo-radius: @standard-halo-radius;
     text-halo-fill: @standard-halo-fill;
     text-placement: interior;
+  }
+
+  [feature = 'tourism_attraction'][zoom >= 17][is_building = 'no'] {
+    text-name: "[name]";
+    text-size: @standard-font-size;
+    text-wrap-width: @standard-wrap-width;
+    text-line-spacing: @standard-line-spacing-size;
+    text-dy: 6;
+    text-fill: @tourism;
+    text-halo-radius: @standard-halo-radius;
+    text-halo-fill: @standard-halo-fill;
+    text-placement: interior;
+    text-face-name: @standard-font;
   }
 }
 
--- a/project.mml
+++ b/project.mml
@@ -2317,9 +2317,9 @@ Layer:
                   'highway_' || CASE WHEN highway IN ('services', 'rest_area', 'bus_stop', 'elevator') THEN highway ELSE NULL END,
                   'boundary_' || CASE WHEN boundary IN ('national_park') THEN boundary ELSE NULL END,
                   'waterway_' || CASE WHEN waterway IN ('dam', 'weir', 'dock') THEN waterway ELSE NULL END,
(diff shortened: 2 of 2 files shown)


kocio-pl:
How would it compare to shop violet dot? I believe this will be too similar.

Also please make as small changes as possible and put them into separate PRs. We have a lot of discussions which push such complex propositions to the edge of being never finished and merged, which would be waste of your great work.

Adamant36:
Looks like one of your before photos screwed up in the upload process.

jeisenbe:
Thank you for catching my mistake. Should be fixed now.

jeisenbe:
I hope it's not too complicated. This seemed to be the smallest change that would properly fix #3545 at the time. I assumed the icon and text-dy were needed, though perhaps that is incorrect?

We could try without adding the generic dot, but it's not great to show a text label without an icon or outline or area, as @gravitystorm mentioned in 
https://github.com/gravitystorm/openstreetmap-carto/pull/1063#issuecomment-67853592

There are several purple shop dots in the test images above, by the Singapore Zoo entrance (upper left, near some orange fast food dots), and in the retail area "Resort World Sentosa" in the last image. 

Here's z18, where the shop icons show, to compare with the tourism dot and color:
z18-zoos-after

Adamant36:
Hhhmmm looks different enough in my opinion. Maybe the shop color could be slightly lightened to or something so they are that much easier to tell apart though.

kocio-pl:
For me the fact that you have used colon in the title and listed 4 changes in one PR is a simple indicator that this is complex solution.

When they are rendered near each other with a full icon I can see the difference, but when it comes to scattered dots they look the same, so I would not go with that.

jeisenbe:
Here's the location of the original issue:
https://www.openstreetmap.org/#map=19/-23.08121/-47.98788
Node with man_made=cross and tourism=attraction

Before z17
z17-cross-before
After z17
z17-cross-after

jeisenbe:
We could try a small square or circle instead of the dot:

1. Solid Square, 4x4 pixels:
z17-zoo-squares

2. Hollow Circle, 6 pixels:
z17-zoo-circles

kocio-pl:
Yes, that sounds sane for me.

jeisenbe:
3. Small circles, 4 pixel:
- not as recognizably circular as 2. (above), but still clearly different than the filled dots.
z17-zoom-small-circles

jeisenbe:
4. Square icon, 3 pixel width: 
3-pixel-squares-textdy4

The square icon is clearly different from the circular dots, yet is unobtrusive, and no new svg icon is needed.

I've pushed a new commit with this change. 

(I would also be happy to remove the icon from this PR, if this is too many changes to review in one pull request).

Adamant36:
Square icons are for specific things. They aren't going to be confused with transportation squares in this context. So, I think this would be a good usage of them.

kocio-pl:
I have just checked that 85% has name:

https://taginfo.openstreetmap.org/tags/tourism=attraction#combinations

Yet I'm not sure how the rest (15%) will look with just a dark violet square. Especially for areas. What do you think about it?

jeisenbe:
Is that 85% for tourism=attraction without any additional tags? I would be
surprised to see tourism=attraction without a name or any other, more
specific tag.

I had considered only rendering the dot with a name; it would be easy
enough to add [name != null], if in fact there are instances without a name
and no other tagging.

kocio-pl:
Hard to say, Taginfo provides only simple stats.

jeisenbe:
I've added a commit. Now the icon will only render if the name is not null.

matthijsmelissen:
I think the example images are mistagging - attraction=tourism is not intended for theme park attractions, but only for things that are tourist attractions themselves (the theme park would be, but the attractions are not, in my opinion).

For regular tourist attractions, I don't think a dot would improve the rendering situation.

kocio-pl:
After some more thinking I would also not like to add dots/squares. All the other propositions (including moving rendering of general "attractions" to z17 to not encourage abusing this tag) are OK for me.

My proposition is to remove dot rendering and then I can go on with code review.

jeisenbe:
Ok, I've removed the generic icons for now, as they are not required to solve the main issue. The code should be ready for review

My opinion is that tourism=attraction is most useful when it is used along with attraction=*; the most common values of attraction are animal (eg zoo animal), roller coaster, water slide, summer toboggan, etc - all theme park or zoo features. I almost all other cases there is a better primary tag for a feature that the generic tourism=attraction.

I also note that @gravitystorm previously recommended that features should have either an icon or an area rendering of some sort, in this comment:
https://github.com/gravitystorm/openstreetmap-carto/pull/1063#issuecomment-67853592

jeisenbe:
For the record, here are a couple of examples of why we should no longer the text labels render sooner than z17.

This is a real-world example of a very large "tourism=attraction", apparently the whole lake area, or perhaps the surrounding caldera too? It renders on z13, as I happened to notice while looking for areas with patterns:
z13-sete-cidades-mistagging

An arbitrarily large area tagged as "tourism=attraction" renders even on z10.
[Not real data]
z10-wamena-tourism-tagging-for-renderer

jeisenbe:
I've removed the two lines from amenity_low_priority and amenity_low_priority_poly in project.mml because we are not going to render an icon.

PR #3614, opened 2018-12-31
Add missing oneway arrows for tracks and paths

### LINKED ISSUE #1937: improve oneway arrow on cycle ways
The line of the arrow is not very good visible on (at least paved) cycle ways. 
Perhaps a white halo would help.

### LINKED ISSUE #2865: Oneway arrows for highway=cycleway/footway/path
Some time ago (I guess more than a year already) cycleways tagged with oneway=yes featured arrows. I think it would be very useful to reintroduce this feature as this motivates the correct mapping of unidirectional cycleways.

An example area with a mix of unidirectional and bidirectional cycleways:
http://www.openstreetmap.org/way/19649973#map=19/48.27127/16.42020

jeisenbe:
Fixes #2865 
Fixes #1937

- Add missing oneway arrows for tracks and paths
- Fixed rendering of arrows on footways: these will always use the footway color, rather than changing if there is (mis)-tagging with designated=bicycle or designated=bridleway.
- Fixed bridleway one-way arrow color. It is now based on brideway color instead of track color.

**EDITED: New changes**
- Change arrows to Unicode text with halo for better visibility.
- Adjust position of text based on `oneway = -1` to make sure text is always on the same side of one-way paths and tracks

### Test rendering:
(First option, without change to unicode text arrows and halos)

Before
oneway-master

After - first option
path-oneway-after

**EDITED: Current commit:**

### DIFF
--- a/project.mml
+++ b/project.mml
@@ -2001,7 +2001,7 @@ Layer:
             horse,
             bicycle
           FROM planet_osm_line
-          WHERE highway IN ('bridleway', 'footway', 'cycleway', 'path', 'track', 'steps')
+          WHERE highway IN ('bridleway', 'footway', 'cycleway', 'path', 'track', 'steps', 'construction')
             AND (name IS NOT NULL
               OR oneway IN ('yes', '-1')
               OR junction IN ('roundabout'))
--- a/project.mml
+++ b/project.mml
@@ -1982,6 +1982,7 @@ Layer:
     properties:
       minzoom: 13
   - id: paths-text-name
+    class: directions
     geometry: linestring
     <<: *extents
     Datasource:
@@ -1991,10 +1992,19 @@ Layer:
             way,
             highway,
             construction,
-            name
+            name,
+            CASE
+              WHEN oneway IN ('yes', '-1') THEN oneway
+              WHEN junction IN ('roundabout') AND (oneway IS NULL OR NOT oneway IN ('no', 'reversible')) THEN 'yes'
+              ELSE NULL
+            END AS oneway,
+            horse,
+            bicycle
           FROM planet_osm_line
-          WHERE highway IN ('bridleway', 'footway', 'cycleway', 'path', 'track', 'steps', 'construction')
-            AND name IS NOT NULL
+          WHERE highway IN ('bridleway', 'footway', 'cycleway', 'path', 'track', 'steps')
+            AND (name IS NOT NULL
+              OR oneway IN ('yes', '-1')
+              OR junction IN ('roundabout'))
         ) AS paths_text_name
     properties:
       minzoom: 15
--- a/roads.mss
+++ b/roads.mss
@@ -3262,24 +3262,24 @@ tertiary is rendered from z10 and is not included in osm_planet_roads. */
       text-placement: line;
       text-face-name: @book-fonts;
       text-vertical-alignment: middle;
-      text-dy: 5;
+      text-dy: -5;
       [oneway = '-1'] {
-        text-dy: -5;
+        text-dy: 5;
       }
       text-repeat-distance: @major-highway-text-repeat-distance;
     }
     [zoom >= 16] {
       text-size: 9;
-      text-dy: 7;
+      text-dy: -7;
       [oneway = '-1'] {
-        text-dy: -7;
+        text-dy: 7;
       }
     }
     [zoom >= 17] {
       text-size: 11;
-      text-dy: 9;
+      text-dy: -9;
       [oneway = '-1'] {
-        text-dy: -9;
+        text-dy: 9;
       }
     }
   }
@@ -3305,18 +3305,18 @@ tertiary is rendered from z10 and is not included in osm_planet_roads. */
       text-placement: line;
       text-face-name: @book-fonts;
       text-vertical-alignment: middle;
-      text-dy: 7;
+      text-dy: -7;
       [oneway = '-1'] {
-        text-dy: -7;
+        text-dy: 7;
       }
       text-repeat-distance: @major-highway-text-repeat-distance;
       [highway = 'steps'] { text-repeat-distance: @minor-highway-text-repeat-distance; }
     }
     [zoom >= 17] {
       text-size: 11;
-      text-dy: 9;
+      text-dy: -9;
       [oneway = '-1'] {
-        text-dy: -9;
+        text-dy: 9;
       }
     }
   }
@@ -3341,13 +3341,7 @@ tertiary is rendered from z10 and is not included in osm_planet_roads. */
     [highway = 'road'],
     [highway = 'service'],
     [highway = 'pedestrian'],
-    [highway = 'raceway'],
-    [highway = 'cycleway'],
-    [highway = 'footway'],
-    [highway = 'path'],
-    [highway = 'steps'],
-    [highway = 'track'],
-    [highway = 'bridleway'] {
+    [highway = 'raceway'] {
       [oneway = 'yes'],
       [oneway = '-1'] {
         marker-placement: line;
@@ -3393,45 +3387,56 @@ tertiary is rendered from z10 and is not included in osm_planet_roads. */
         [highway = 'raceway'] {
           marker-fill: @raceway-oneway-arrow-color;
         }
+      }
+    }
 
-        [highway = 'steps'] {
-          marker-fill: @steps-oneway-arrow-color;
-          marker-transform: 'translate(0,5)';
-          [oneway = '-1'] {
-            marker-transform: 'translate(0,-5)';
+    [highway = 'steps'],
+    [highway = 'cycleway'],
+    [highway = 'footway'],
+    [highway = 'path'],
+    [highway = 'track'],
+    [highway = 'bridleway'] {
+      [oneway = 'yes'],
+      [oneway = '-1'] {
+        text-name: "'⭢'";
+        text-size: 18;
+        text-clip: false;
+        text-spacing: 180;
+        text-placement: line;
+        text-halo-fill: @standard-halo-fill;
+        text-halo-radius: 1.5;
+        text-margin: 2;
+        text-dy: 4;
+        text-upright: right;
+        text-vertical-alignment: middle;
+        text-face-name: @book-fonts;
+        [oneway = '-1'] {
+          text-upright: left;
+          text-dy: -4;
+        }
+        [highway = 'footway'] {
+          text-fill: @footway-oneway-arrow-color;
+        }
+        [highway = 'path'] {
+          text-fill: @footway-oneway-arrow-color;
+          [horse = 'designated'] {
+            text-fill: @bridleway-oneway-arrow-color;
+          }
+          [bicycle = 'designated'] {
+            text-fill: @cycleway-oneway-arrow-color;
           }
         }
-
-        [highway = 'footway'],
-        [highway = 'path'],
-        [highway = 'cycleway'],
-        [highway = 'track'],
+        [highway = 'steps'] {
+          text-fill: @steps-oneway-arrow-color;
+        }
+        [highway = 'cycleway'] {
+          text-fill: @cycleway-oneway-arrow-color;
+        }
+        [highway = 'track'] {
+          text-fill: @track-oneway-arrow-color;
+        }
         [highway = 'bridleway'] {
-          marker-transform: 'translate(0,4)';
-          [oneway = '-1'] {
-            marker-transform: 'translate(0,-4)';
-          }
-          [highway = 'footway'] {
-            marker-fill: @footway-oneway-arrow-color;
-          }
-          [highway = 'path'] {
-            marker-fill: @footway-oneway-arrow-color;
-            [horse = 'designated'] {
-              marker-fill: @bridleway-oneway-arrow-color;
-            }
-            [bicycle = 'designated'] {
-              marker-fill: @cycleway-oneway-arrow-color;
-            }
-          }
-          [highway = 'cycleway'] {
-            marker-fill: @cycleway-oneway-arrow-color;
-          }
-          [highway = 'track'] {
-            marker-fill: @track-oneway-arrow-color;
-          }
-          [highway = 'bridleway'] {
-            marker-fill: @bridleway-oneway-arrow-color;
-          }
+          text-fill: @bridleway-oneway-arrow-color;
         }
       }
     }
--- a/roads.mss
+++ b/roads.mss
@@ -288,11 +288,11 @@
 @living-street-oneway-arrow-color: darken(@residential-casing, 30%);
 @pedestrian-oneway-arrow-color:   darken(@pedestrian-casing, 25%);
 @raceway-oneway-arrow-color:      darken(@raceway-fill, 50%);
-@footway-oneway-arrow-color:      darken(@footway-fill, 40%);
-@steps-oneway-arrow-color:        darken(@steps-fill, 45%);
-@cycleway-oneway-arrow-color:     darken(@cycleway-fill, 30%);
-@track-oneway-arrow-color:        darken(@track-fill, 20%);
-@bridleway-oneway-arrow-color:    darken(@bridleway-fill, 20%);
+@footway-oneway-arrow-color:      darken(@footway-fill, 35%);
+@steps-oneway-arrow-color:        darken(@steps-fill, 35%);
+@cycleway-oneway-arrow-color:     darken(@cycleway-fill, 25%);
+@track-oneway-arrow-color:        darken(@track-fill, 10%);
+@bridleway-oneway-arrow-color:    darken(@bridleway-fill, 15%);
 
 // Shield’s line wrap is based on OpenStreetMap data and not on line-wrap-width,
 // but lines are typically rather short, so we use narrow line spacing.
@@ -3263,15 +3263,24 @@ tertiary is rendered from z10 and is not included in osm_planet_roads. */
       text-face-name: @book-fonts;
       text-vertical-alignment: middle;
       text-dy: 5;
+      [oneway = '-1'] {
+        text-dy: -5;
+      }
       text-repeat-distance: @major-highway-text-repeat-distance;
     }
     [zoom >= 16] {
       text-size: 9;
       text-dy: 7;
+      [oneway = '-1'] {
+        text-dy: -7;
+      }
     }
     [zoom >= 17] {
       text-size: 11;
       text-dy: 9;
+      [oneway = '-1'] {
+        text-dy: -9;
+      }
     }
   }
 
@@ -3297,12 +3306,18 @@ tertiary is rendered from z10 and is not included in osm_planet_roads. */
       text-face-name: @book-fonts;
       text-vertical-alignment: middle;
       text-dy: 7;
+      [oneway = '-1'] {
+        text-dy: -7;
+      }
       text-repeat-distance: @major-highway-text-repeat-distance;
       [highway = 'steps'] { text-repeat-distance: @minor-highway-text-repeat-distance; }
     }
     [zoom >= 17] {
       text-size: 11;
       text-dy: 9;
+      [oneway = '-1'] {
+        text-dy: -9;
+      }
     }
   }
 }
@@ -3338,7 +3353,6 @@ tertiary is rendered from z10 and is not included in osm_planet_roads. */
         marker-placement: line;
         marker-spacing: 180;
         marker-max-error: 0.5;
-
         marker-file: url('symbols/oneway.svg');
         [oneway = '-1'] {
           marker-file: url('symbols/oneway-reverse.svg');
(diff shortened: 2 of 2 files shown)


jeisenbe:
1. Here's the current commit (steps darkened 35% and bridleway darkened 15%, based on the line fill color) at z17. The paths are the same width as at 16, but this image lets us see three arrows instead of 2. (The track is visible at top but horizontal)
z17-oneway-after-35-15

2. Here's a slightly darker version: step arrows darkened 45% and bridleway arrows darkened 20%:
z16
z16-paths-oneway-darker-45-20

z17 
z17-paths-oneway-darker-45-20

3. Even darker version: step arrows darkened 50% and bridleway arrows darkened 25%:
z16
z16-darker-50-25
z17
z17-darker-50-25

@pnorman would you like to see changes for footway, cycleway or highway=track as well? I think there is some utility in matching the arrows to the color of the highway=* in cases where there are parallel paths close together, but slightly darkened arrows would still be readable as related to the original color.

jeisenbe:
Here's another test area, showing all 5 different track grades with the track oneway arrows. This also allows us to see the other arrows in a different orientation.

1. Steps 35%, Bridleway 15% (current commit)
tracktype-arrows-35-15

2. Steps 45%, Bridleway 20%
tracktype-45-20

3. Steps 50%, Bridleway 25%
tracktype-50-25

4. Every arrow darker: footway 40%, steps 45%, cycleway 30%, track 20%, bridleway 20% 
(these are all 5% darker than the current code, except for steps which is 10% darker)
arrows-40-45-30-20-20

jragusa:
Arrows of bridleway and track (especially tracktype=1-4) are not well readable when located along the path

HolgerJeromin:
Would/should fix  #1937 too

Adamant36:
I like the idea of it being on the side of the path. Except where the symbol would overlap a road, like in the second to last test rendering. It would be interesting to see another exemple like that where the road is also one way. It would probably be either A. confusing or B. one arrow would out render the other. 

@jeisenbe, wasnt there a propsal a while back to make track roads or paths render more like regular roads since they aren't very visible in the first place (I think like the German style or something)? Maybe its worth revisiting. I kind of like the idea. It would give a better area for the arrows and showing paved or not later. As it is, on the path doesnt look great and off has its own issues.

HolgerJeromin:
> I had hoped this could be a simple fix, because it was meant to revert the arrows to the way they used to be rendered a year ago.

Did you found out with which commit we lost the arrows?

jeisenbe:
> Did you found out with which commit we lost the arrows?

No. I suspect it happened the last time that the project.mml file was
significantly changed in this area.

I wonder if there is a way on git or github to find the last change to a
line of code?

Adamant36:
You might be able to do an advanced search in the search box for any closed PR that has a specific word in the code. Or if not, at least by commit message. https://help.github.com/articles/searching-issues-and-pull-requests/

matthijsmelissen:
> I wonder if there is a way on git or github to find the last change to a
line of code?

Git: git blame [filename]
Github: navigate to the file, click the Blame button

jeisenbe:
Related to 
https://github.com/gravitystorm/openstreetmap-carto/issues/1937 - we could try using a "halo" around the arrows. I've checked the Mapnik/CartoCSS documentation, and there isn't a halo function for SVG icons or markers. So we would need to modify the oneway.svg file and oneway-reverse.svg to have a white, 0.5 pixel area with perhaps 50% opacity around the arrow. But I don't know if it is possible to do this and then still change the fill color for each type of arrow. We could make a separate svg file for each type of path, but this would make the style harder to adapt and maintain. 

Any ideas about how to do this?

Otherwise, darkening the arrow color and moving the arrows off of the line are the two best options, as shown above.

Adamant36:
(Or go the more complicated but good route and change paths to render more like roads first. Then this wouldnt be an issue. Plus, there would be other benefits)

jeisenbe:
I've pushed several commits that show the code changes for the different options in the previous test renderings. The most recent commit has the arrows off of the path.

jeisenbe:
> (Or go the more complicated but good route and change paths to render more like roads first. Then this wouldnt be an issue. Plus, there would be other benefits)

That's also an interested idea, I forgot about that. I've just downloaded the German and French styles to see how they work, but I thought it may have only been the footways that were changed? There still might be an issue with track roads, bridleways and cycleways to consider.

jeisenbe:
Here are some real-world renderings in the Netherlands, land of the cyclepath:

This road is only a few meters wide and has one-way cycle paths immediately next to it, so they overlap with the road rendering. This is a challenging rendering situation:
https://www.openstreetmap.org/#map=17/52.34378/5.20680
z19 arrows darkened, on cycle paths
z19-meerveldstaat-on
z19 arrows to side of cycle paths
z19-meerveldstraat-side

At z18 the version with the arrows to the side could be confusing, because the path is partially covered by the road. This is an unusual situation but can happen with a one-lane road like this. 
z18 arrows on path, darkened
z18-meerveldstraat-on

z18 arrows to side
z18-meerveldstraat-side

This bridge is a more reasonable width: 10 meters between the center of the two cycleways (33 feet)
https://www.openstreetmap.org/#map=18/52.33349/5.22320
z18 on path, darkened
z18-sluiskade-on
z18 arrows to side
z18-sluiskade

A one-way path (MTB trail?):
https://www.openstreetmap.org/#map=17/52.40464/5.33439
z17 on
z17-mtb-path-on
z17 side
z17-mtb-path-side

One-way footway(?):
https://www.openstreetmap.org/#map=18/52.53644/5.70264
z19 on
z19-oneway-footway-on
z19 side
z19-footway-oneway-side

One-way track road:
https://www.openstreetmap.org/#map=19/52.35675/5.22666
z19 on
z19-track-oneway-on
z19 side
z19-track-oneway-side

One-way steps:

Adamant36:
I think originally it was for footways. Track roads etc could be more view-able though and they would also benefit from paved/unpaved rendering later on. A while back I was screwing with an idea of having them render similar to normal roads, but with the brown colored dashes (or whatever they are) larger and inside of the road (I've always wondered why that wasn't explored as an option for showing paved/unpaved).

Kind of like this
path

jeisenbe:
One-way steps (an escalator in a mall in Harburg, Germany):
https://www.openstreetmap.org/#map=18/53.45756/9.98706
Arrows on steps
steps-oneway-on

Arrows to side
steps-oneway-side

jragusa:
When looking to the MTB trail example, I'm wondering if it would be better to also apply opacity to arrows

Adamant36:
@jragusa, see  https://github.com/gravitystorm/openstreetmap-carto/pull/3614#issuecomment-451315125

jragusa:
@Adamant36 I mean the arrow itself not the halo, like private tracks. Sorry for the misunderstanding

pnorman:
> @pnorman would you like to see changes for footway, cycleway or highway=track as well? I think there is some utility in matching the arrows to the color of the highway=* in cases where there are parallel paths close together, but slightly darkened arrows would still be readable as related to the original color.

It's hard to tell. I can't really see the colours in many of the sample images, they look black.

I suspect the answer lies in a halo, more below...

----

> I had hoped this could be a simple fix, because it was meant to revert the arrows to the way they used to be rendered a year ago.

We were probably rendering arrows with a catch-all, which we've been trying to get away from. I know in the more distant past, we built up arrows by using dash patterns.
-----

-----
For halos, it might work if we used arrow characters like `→` as text and then used text halo stuff.

----- 

> This road is only a few meters wide and has one-way cycle paths immediately next to it, so they overlap with the road rendering. This is a challenging rendering situation:
> https://www.openstreetmap.org/#map=17/52.34378/5.20680

This is probably the most common case of oneway arrows on non-motorized paths too, so it needs to look good here.

jeisenbe:
Here's U+1f816 "Rightwards arrow with equilateral arrowhead"; a little more similar to the current svg. This may be a better option:
Track roads:
tracks-arrow2
Vertical paths:
vertical-arrow2
Circles:
arrow2-circles

jeisenbe:
Test renderings with  U+1f816 "Rightwards arrow with equilateral arrowhead":

One-way bicycle paths along a road in Harburg: 
https://www.openstreetmap.org/#map=16/53.4715/9.9532
z16
z16-stader-cycleways

z17
z17-stader-cycleways

Harburg escalators in malls: 
https://www.openstreetmap.org/#map=18/53.45756/9.98706
z18
mall-arrows-steps

Mix of "paths" and bicycle paths in central Hamburg:
paths-and-cyclepaths-arrows-hamburg

Grade 1 tracks:
https://www.openstreetmap.org/#map=16/53.5435/10.0642
z17-arrows-tracktype1

Grade 2 and Grade 4 tracks:
https://www.openstreetmap.org/#map=16/53.4833/10.1197
z17-arrows-tracktype2-4
16/53.4833/10.1197

@pnorman - does this look like a good solution?

jeisenbe:
Small one-way bridleway in central Vienna
https://www.openstreetmap.org/#map=#17/48.20900/16.36208
z17-oneway-bridleway

I believe this is ready for review.

jeisenbe:
@matthijsmelissen - would be interested in reviewing this PR?

matthijsmelissen:
Will look at it once I have time.

matthijsmelissen:
I'm not convinced there is a need for footways and cycleways to have one-way arrows. I believe one-way footways are extremely rare. Oneway cycleways are common, but practically always oneway cycleways have an opposite cycleway right next to it (on the opposite side of the road), so this is not so useful for map users. On the other hand, displaying the direction of cycleways will clutter the map.

Is there any real-world evidence we need this?

matthijsmelissen:
I would be interested in the opinion of the other maintainers.

pnorman:
> Is there any real-world evidence we need this?

> I would be interested in the opinion of the other maintainers.

I'd defer to you on this, because you live in a place where one-way non-motorized is more common.

Locally, every oneway non-motorized I could find was either one I believed was tagged wrong, or one where the oneway arrows would add little value.

imagico:
Could someone explain to me what part of this PR is

* a simple bug fix for a regression
* a design change for oneway arrows that were (before the bug) already rendered in principle but that are rendered differently now
* a feature addition

For the last point it would be good to see some numbers on how many of the highway types for which arrows are to be added have a oneway tag.

pnorman:
> * a simple bug fix for a regression

I don't view it as a bug fix. Removing catch-alls and only rendering arrows on explicit objects caused arrows to disappear from non-motorized ways. I believe this to be intentional.

jeisenbe:
The initial PR was mainly a simple bug fix for a regression.

But I also changed the bridleway arrow color to @bridleway instead of
@track, and fixed a bug that rendered one-way footway arrows in the
cycleway or bridleway color.

Now the current PR also changes the cartography of all of the path arrows
to add a halo so that they are easier to see, especially on steps.

But there are no new features added. These arrows were apparently all
rendered a year ago, according to #2865

While most of the one-way footways that I’ve seen have bicycle=yes, I’m not
sure that this is mistagging?

Is there a clear division between footway and highway=path which requires
that footways be exclusively for the use of pedestrians?

jeisenbe:
I haven’t been able to find the PR where the arrows were removed. I tried
`git blame` but it wasn’t clear what happened.

Was this discussed at some point?

pnorman:
> I haven’t been able to find the PR where the arrows were removed. I tried `git blame` but it wasn’t clear what happened.

I'd go for using `git bisect` to try to narrow it down. It's a bit more of a pain, but can work better where bisect fails.

> Is there a clear division between footway and highway=path which requires that footways be exclusively for the use of pedestrians?

For the purpose of this style, highway=path is turned into highway=footway or highway=cycleway.

HolgerJeromin:
Perhaps oneway=no is more interesting to render. But i have not checked the usage

jeisenbe:
>
> For the purpose of this style, highway=path is turned into highway=footway
> or highway=cycleway.
>

Right. A highway=footway can have bicycle=yes, and a highway=path can have
bothe bicycle=yes and foot=yes. In either case it might be one way for
bicycles.

A number of paths on bridges in Portland, Oregon could be tagged this way;
the sidewalks are used for one way bike travel (actually some are two-way,
so it’s useful to show the difference)

>

imagico:
Ok, i will try to comment without making assumptions on if parts of this is a bug fix or not.

To me oneway arrows on steps definitely make sense.  In particular for escalators where up and down might be separated and it is important for navigation to know which is which.

But for escalators there is also conveying=* which is more common than oneway (not sure how many have both).  Also often a direction arrow is of little help with steps when incline=* is not rendered.  So improving rendering of steps to be more meaningful is kind of a separate matter.

Almost 10 percent of highway=cycleway (100k) have oneway=yes but only 6000 of nearly 10 million footways.  path: 20k of 7 million (many of them probably foot=designated + bicycle=designated), track 5600 of 15.7 million, bridleway: less than 1000.

I see the problem mainly in that cycleways are the primary use case here and there are practical cases where showing oneway arrows is useful - like a large street crossing with various direction segregated cycleway links where oneway arrows would be good to have and important for feedback or downhill mountainbike routes but this strongly interferes with local implicit rules.  In Germany for example all road parallel cycleway are legally oneway for bicycles unless designated otherwise.  Therefor tagging those oneway=no makes more sense and showing this like @HolgerJeromin would be more significant.

Oneway footways are a highly specialized thing and usually not really a legal "you may not go in the other direction" but more of a "this is the normal direction you are expected to go of you do not want people to think you are weird".

For track and bridleway i don't know.  Would be helpful if someone would extract those 5.6k from a planet file for reviewing.

In general when showing oneway arrows for non-motorized traffic it might also make sense to start later than z16.

I am not opposed to this change as is but my recommendation in principle would be to separate into different matters (which could of course all be tackled either with or without this PR merged):

* more meaningful rendering of steps possibly indicating conveying=*, incline=* and oneway=*
* rendering of direction restrictions and permissions on cycleways taking into account that there are implicit defaults that likely vary between countries
* analyzing the quality of oneway tagging on track, bridleway and footway if it makes sense to render oneway arrows on those.

jeisenbe:
Rendering the arrows a zoom level later is a reasonable idea, but I had
been hoping to keep the number of changes down, since several maintainers
have said that they prefer simpler PRs.

One of the main reasons to show the arrows, even on tracks and bridleway,
is to help mappers find mistakes which will affect routing applications.

It’s a pain when a router sends you the wrong way, or mistakenly avoids a
path because it is mistagged as oneway.

jeisenbe:
I removed a minor change to the text-dy for path name labels, which would have been necessary if using arrows next to the path rather than on top of it, but is no longer needed, and I fixed a mistake that would have stopped the rendering of the names of under-construction paths. I believe everything is now working properly.

matthijsmelissen:
I'm holding of merging this until we find out when and why the arrows were removed in the first place. Would be great if somebody could find this back.

jeisenbe:
Does anyone know a commit or version where the arrows were last shown, so I
can use `git bisect`? I don’t want to start back in 2015

matthijsmelissen:
@jeisenbe `git bisect` bisects the history, which means starting in 2015 goes very fast. If you start in 2015, after one step you already will know if it was before or after 2017, and after two steps you will know if it was before or after 2018, etc.

jeisenbe:
I believe the bug was introduced in Commit e455ee462e0ba1b16c6422b27ff734f2cb050f92 - "Move one way arrow rendering from several roads fill related layers to roads-text-name, so arrows get lower precedence than names. As the arrow symbol is quite small, most of the time it can still be squeezed in."
PR #2320

jeisenbe:
Also it looks like #2342 fixed the rendering of one-way arrows on roads, but probably it was missed that they were still not rendering on paths. 

There was never any discussion about removing the rendering of oneway arrows from paths; this was an unintentional change.

jeisenbe:
> Oneway cycleways are common, but practically always oneway cycleways have an opposite cycleway right next to it (on the opposite side of the road)

Perhaps this is true in the Netherlands, which has world-renowned bicycle infrastructure However, the UK, USA and other English-speaking countries are notorious for having one-way cycleways only on one side of a street (you get to ride with motor vehicle traffic on the other side), and having complicated ways to access one-way cycle paths on bridges, and at major intersections. 

We also have a number of one-way mountain bike trails.

Another situation that has been mentioned are one-way "via ferrata" in the mountains, foot/climbing paths.

jeisenbe:
I checked Singapore, which is relatively well-mapped. There are 111 ways tagged with oneway=yes and highway=path, =cycleway, =footway, =steps, or =track. Cycleway and footway have 39 each, the rest have 6 to 13 occurrences. These examples are the first I found. There are no examples of parallel one-way paths or cycleways on opposite sides of a street, but there are a number of one-way mountain bike paths, one-way pedestrian bridges and steps (at the international border, in a zoo and in a park), and one-way circles for bike and golf carts.

z19 Create Tower oneway bicycle path
https://www.openstreetmap.org/#map=19/1.30357/103.77436
- If you want to go west you can ride on the main roadway; bicycle=yes
z19-create-tower-cyclepath

z17 Chestnut Point Mountain Bike trails
https://www.openstreetmap.org/#map=17/1.3717944/103.7798446
z17-chestnut-point-mtb-oneway

z17 Singapore Island Country Club golf cart path
https://www.openstreetmap.org/#map=17/1.3637611/103.826251
z17-singapore-country-club-path

z16 HSBC-sponsored "Treetop Walk" (wooden path thru the trees and up hills)
https://www.openstreetmap.org/#map=17/1.36139/103.8094101
z16-hsbc-treetop-walk

z18 Woodlands checkpoint (international border)
https://www.openstreetmap.org/#map=18/1.4455896/103.7688039
- You probably don't want to try going the wrong way on these paths and steps!
z18-woodlands-checkpoint-paths-steps
z19
- You can see that the path and steps lead from the bus to the building entrance, where your passport is checked. On the other side you walk back to another bus stop. 
z19-woodlands-checkpoint-path-steps

z16 Pulau Ubin track road "Jalan Durian"
https://www.openstreetmap.org/#map=16/1.410072/103.982814
- An agricultural/forestry road in the city-state of Singapore?! Well, this is another small island that was used as a quarry and is now preserved for recreation and nature.
z16-jalan-durian-pulau-ubin

jeisenbe:
Bremen, Germany - there are a number of cycleways and paths, and a few tracks and footways with oneway=yes

https://www.openstreetmap.org/#map=17/53.09655/8.87581
- One-way highway=footway with footway=sidewalk, foot=designated and bicycle=yes
z17-berckstrathe-footway-oneway

https://www.openstreetmap.org/#map=18/53.08297/8.79848
- One-way paths along causeway/bridge over historic moat, connecting to one-way and two-way cycleways. Note that the cycleway along the north side of Eduard-Schopf-Allee is specifically tagged oneway=no, as is the path along the northeast side of Doventorscontrescarpe
z18-doventorscontrescarpe

https://www.openstreetmap.org/#map=16/53.0723428/8.8427867
- There are one-way cycleways only along some parts of the southern street (Hamburger)
- Still visible at z16 distinctly


https://www.openstreetmap.org/#map=16/53.0681/8.7594
- One-way track roads in allotments

jeisenbe:
**Canberra, Australia**

This capital city was purpose-built in mid-century style, with lots of motorways and roundabouts, including a motorway circling the parliament buildings. I guess they didn't think much about air pollution back then? 

I've shown the first 2 oneway cycleways and paths that I found by randomly searching. There was only one footway with oneway=yes, and no tracks or bridleways (which are quite rare in this area).

**Cycleways**
- Most of the oneway cycleways that I saw in Canberra are like these; short connectors that let bike riders avoid traffic circles:

https://www.openstreetmap.org/#map=18/-35.42391/149.07949
z18-drakeford-dr-circle-oneway-cycleway

https://www.openstreetmap.org/#map=18/-35.16602/149.13933
z18-mulligans-flat-cycleways

**Paths**
- most of the oneway paths are MTB trails (which may also be open to people on foot? This is not tagged either way):

https://www.openstreetmap.org/#map=17/-35.25670/149.10143
z17-daviesia-up-down-paths

https://www.openstreetmap.org/#map=16/-35.3146/149.0181
z16-mt-stromlo-mtb-paths

**Footway**
- This is the only oneway footway in the city. While it could be mistagged, it is right in front of the parliament building, by a "formal garden", so perhaps this one-way restriction is correct?

z17-parliament-house-footway

matthijsmelissen:
As this reverts an unintended change, I'd be happy to accept this PR (and certainly the part of the PR reverting this change).

I don't have time for a full review now though, is anyone else willing to help with that?

jeisenbe:
A few more examples of arrows on paths and cyclepaths in Hamburg:

https://www.openstreetmap.org/#map=16/53.4893/10.2108
z17-bergedorf-arrows-17 53 48944 10 20640

https://www.openstreetmap.org/#map=17/53.48944/10.20640
bergedorf-paths-oneway-16 53 4893 10 2108

https://www.openstreetmap.org/#map=18/53.51417/9.96960.png
hamburg-arrows-18 53 51417 9 96960

jragusa:
Since track/path/cycleway are thinner than highway=residential/tertiary/..., why not shift arrows for the formers to z18 ? They look prominent in the two first pictures.

jeisenbe:
That’s possible. But this PR was originally intended to fix a bug that had
stopped the arrows from rendering on paths, rather than significantly
change the old rendering.

matthijsmelissen:
Thanks for this PR, and sorry for the wait!

jeisenbe:
Thanks!
If anyone finds the problems with arrows on z16, or has other ideas for
improvements, I would be happy to do another PR to adjust the new rendering.

jengelh:
The choices made in osm-carto are generally well researched, and I am impartial to the arrow style. Just an inquiry though, did you consider JOSM style arrows (triangle with a colored outline), and if so, what was your opinion of those?
josmarrow

jeisenbe:
I did not consider that style. This change only affected the one-way arrows
on paths, so they needed to closely match the existing arrows on roads.

Issue #3517, opened 2018-11-18
Coordinate shades of non-physical green areas (meta-ticket)


Tomasz-W:
As we have a lot of different green shades on map, and some of them are discussed or considered as unnecessary, I'm opening meta-ticket to coordinate whole topic.

Ticket assumptions:
* ~~make park colour more greenish instead of blueish~~
* ~~make park colour more prominent~~
* ~~distinguish @_park_ and @_leisure_ colours more from each other~~
* eliminate some unnecessary green shades from map

Propositions of this ticket:
- [ ] test outlines for parks (https://github.com/gravitystorm/openstreetmap-carto/issues/3264)
- [ ] change `leisure=golf_course` and `leisure=miniature_golf` area fill to @_leisure_ (they are both just a kind of sports centres, so I don't see a reason for using special color here) (https://github.com/gravitystorm/openstreetmap-carto/issues/2069)
- [ ] change `leisure=common` area fill to @_leisure_ (rendering it the same as grass leads to bad tagging like here: https://www.openstreetmap.org/way/270339855)
- [ ]  drop `tourism=campsite` and `tourism=caravan_site` filling (current colour is too close to grass and might be confused with some landcover rendering; these sites are usually a group of tents and tourism-related nodes throwed on a bigger area of grass; we use fills rather for areas with some clear edges, in this case outline is usually random, so area filling is confusing) (https://github.com/gravitystorm/openstreetmap-carto/issues/1485)
- [ ] drop fill for `tourism=picnic_site` (these sites are usually a group of tourism-related nodes throwed on a bigger area of grass; we use fills rather for areas with some clear edges, in this case outline is usually random, so area filling is confusing)
- [ ] move `leisure=sport_centre` and `leisure=stadium` to @_leisure_ key (just code simplifying)
- [ ] test moving @_golf_course_ colour for pitches

ghost07tula:
Isn't new park color too aggressive?

Adamant36:
Its a tad over saturated. It might look different in rendering tests. We are kind of running out of options when it comes to green colors though.

jeisenbe:
I think it is too bright (saturated).
Perhaps we could try the current golf course or campground color, if we are
not using those anymore.

Adamant36:
![download](https://user-images.githubusercontent.com/30259065/48718633-9d533100-ebd0-11e8-984e-6da77a22549d.png)
To over saturated. I like whatever color it is zoomed out though. 
[...]

Issue #3305, opened 2018-07-15
Review some tags we already render


kocio-pl:
It's one of the things that came to mind when evaluating new tags for rendering: we do this tag by tag and without critical view on currently rendered tags.

1. In heated debate about `landcover=*` we talked only about relative numbers and problems with new proposition. Yet nobody touched the problems with `landuse=forest` and `natural=wood` - see  https://github.com/gravitystorm/openstreetmap-carto/issues/2548#issuecomment-399063510. I'm pretty sure that we would not let them in (especially both) if they would be proposed today.

2. Another issue and much smaller remark (https://github.com/gravitystorm/openstreetmap-carto/issues/3277#issuecomment-401345171) has shown that we also render some things which have no direct use for typical map user, like aeroways. So why we have them at all? And if it's what we want - why should we still show them?

3. And to touch also picking tag-by-tag, let's think about roads. They are poorly defined if we take the objectivity rule (just like `tourism=attraction`), but let's assume only some of them are bad. Yet lack of one type of the road would cause the map to loose its coherence. Similar new case are castles, where it's natural to use some system than pick tags one by one.

Do you see any other problems with existing tags we show?

lakedistrictOSM:
Interesting. A lot of the open issues are to do with features that are missing from the map (and I'm guilty of that) but not so much for things we shouldn't render.

#1003 (`waterway=derelict_canal`) was a good example of something that had rendering but unclear tagging.

`[leisure=common](https://wiki.openstreetmap.org/wiki/Tag:leisure=common?uselang=en-GB)` is another unclear tag that is currently rendered - in the UK it's an official designation so they should really be tagged under the `designation=` key. The green rendering for them is also misleading - I know of one common that is mainly paved! (also mentioned in https://github.com/gravitystorm/openstreetmap-carto/pull/2964)

I can't think of anything else that should lose rendering at the moment, but there are things that are too prominent or render too early for the scale of the feature.

matthijsmelissen:
Another one is leisure=common. It has a very UK-specific meaning, the tag is misused very often, and should normally always be tagged with a physical tag like natural=grass.

pnorman:
I think this is a good chance to *reduce* what we render as we've been adding so much.

Tomasz-W:
Default style renders highway=footway areas while due to OSM Wiki mapping them as areas is incorrect: https://wiki.openstreetmap.org/wiki/Tag:highway=footway

[...]

PR #3619, opened 2019-01-03
Drop leisure=common

matthijsmelissen:
This PR drops rendering of leisure=common.

The leisure=common tag has several problems. I therefore think we should not render it on the default style.

Some of these problems are:

* The tag defines a legal ownership concept, not a physical concept. According to the [wiki](https://wiki.openstreetmap.org/wiki/Tag:leisure%3Dcommon), a common is 'land which has an owner, but over which other people have certain rights'. I think we should render the physical appearance or purpose of features rather than their legal status.

* The tag is very UK centric. Commons are a UK concept that does not really translate to other countries.

* The tag is heavily misused. At least in Western Europe, almost all objects tagged with leisure=common are not actually commons. The tag is often used for random pieces of public grass, such as those alongside roads. The fact that the tag renders nicely green on the map probably has something to do with this.

* Other tags are available. Almost all areas that are currently tagged with leisure=common can also be tagged with other tags that are currently being rendered, such as leisure=park, landuse=grass and/or landuse=farmland.

### DIFF
--- a/amenity-points.mss
+++ b/amenity-points.mss
@@ -2385,7 +2385,6 @@
   [feature = 'leisure_recreation_ground'],
   [feature = 'landuse_recreation_ground'],
   [feature = 'landuse_village_green'],
-  [feature = 'leisure_common'],
   [feature = 'leisure_garden'],
   [feature = 'landuse_quarry'],
   [feature = 'landuse_vineyard'],
@@ -2464,7 +2463,6 @@
       [feature = 'leisure_recreation_ground'],
       [feature = 'landuse_recreation_ground'],
       [feature = 'landuse_village_green'],
-      [feature = 'leisure_common'],
       [feature = 'leisure_garden'] {
         text-fill: @leisure-green;
       }
--- a/landcover.mss
+++ b/landcover.mss
@@ -1,6 +1,6 @@
 // --- Parks, woods, other green things ---
 
-@grass: #cdebb0;        // Lch(90,32,128) also grassland, meadow, common, village_green, garden
+@grass: #cdebb0;        // Lch(90,32,128) also grassland, meadow, village_green, garden, allotments
 @scrub: #c8d7ab;        // Lch(84,24,122)
 @forest: #add19e;       // Lch(80,30,135)
 @forest-text: #46673b;  // Lch(40,30,135)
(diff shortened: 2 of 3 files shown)


Adamant36:
Id support this. Its miss used in America a lot to. I think people here just think it means any kind of "common" area. Like grass at a university. Plus I agree the tag isnt clear. leisure should be about the use case. Not the legal status. So I think removing rendering is a good idea.

kocio-pl:
Related to #3305.

Tomasz-W:
Realated to https://github.com/gravitystorm/openstreetmap-carto/issues/3517

I proposed color change, but dropping it at all is also good for me. Most of these areas in my neighbourhood are usually actually leisure=garden or landuse=recration_ground.

polarbearing:
Misused or not - dropping the rendering has the side-effect that it is also not rendered any more where it is applied correctly.
Do we have an evaluation how much of the tags are incorrect?

polarbearing:
Maybe changing the colour, like @Tomasz-W said, would be more reasonable. I could image the residential colour space. The current grass colour is indeed misleading.

jeisenbe:
Perhaps it could be rendered with an outline, like
leisure=nature_reserve (though with a different color), rather than
with a solid polygon fill?

But I confess I don't understand exactly what this tag means. It
doesn't seem to have an equivalent in any of the countries where I
have lived.

literannn:
Support dropping. In Russia mappers use this tag for everything including playgrounds, leisure spaces betwwen appartment buildings etc.

dieterdreist:
sent from a phone

> On 4. Jan 2019, at 14:59, jeisenbe <notifications@github.com> wrote:
> 
> Perhaps it could be rendered with an outline, like
> leisure=nature_reserve (though with a different color), rather than
> with a solid polygon fill?


I also wanted to mention nature reserve because it similarly is about a legal status and not about physical appearance:
https://wiki.openstreetmap.org/wiki/Tag:leisure%3Dnature_reserve

if we drop commons for this reason, we would have to drop other tags as well

Adamant36:
Id be for also dropping nature reserve. It has the same issues as leisure=commons and I see it miss-used about as much. boundary=protected_area is a much better tag for nature reserves.

dieterdreist:
> On 6. Jan 2019, at 11:28, polarbearing <notifications@github.com> wrote:
> 
> I see it miss-used about as much
> 
> Could you support this by some numbers, e.g. the percentage of misuse against correct tagging?
> 

and by an explanation what you consider misuse.

turnsole80:
Can I suggest dropping the landcover aspect of this tag, but not the naming?

jeisenbe:
We would need to render an outline or an icon to show the name, if the area
fill color is removed.

Otherwise the name label appears without context.

> Chris wrote:

> Can I suggest dropping the landcover aspect of this tag, but not the
naming?

turnsole80:
Okay, well that rather ruins my cunning plan. I figured it force mappers to make the necessary corrections, whilst still allowed for the correct usage to render where appropriate - at least until a new solution can be found.

matthijsmelissen:
@kocio-pl Could you review this?

kocio-pl:
I am not focused enough and this is quite general problem, sorry.

kocio-pl:
I also think of this problem along with `landuse=village_green`, which makes it even broader.

PR #3620, opened 2019-01-03
Render gate ref's black instead of purple

matthijsmelissen:
This renders gate ref's in black instead of in purple. Purple used to be the airport color, but apart from gates purple is no longer used for airports.

Before:
screen shot 2019-01-03 at 23 23 39

After:
screen shot 2019-01-03 at 23 20 02

### DIFF
--- a/amenity-points.mss
+++ b/amenity-points.mss
@@ -3012,7 +3012,7 @@
     text-size: @standard-font-size;
     text-wrap-width: @standard-wrap-width;
     text-line-spacing: @standard-line-spacing-size;
-    text-fill: #aa66cc;
+    text-fill: black;
     text-face-name: @standard-font;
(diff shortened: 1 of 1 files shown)


Adamant36:
matthijsmelissen, do a link to that place on the map by any chance?

matthijsmelissen:
It's [here](https://www.openstreetmap.org/#map=18/49.63387/6.21671) - but of course it's visible in any airport.

Adamant36:
Thanks. Your probably right. I guess I could have just checked other airports. My bad.

kocio-pl:
I agree that the color is not nice and after the change it's better and different from address nodes:

https://www.openstreetmap.org/#map=18/52.17299/20.97019

screenshot_2019-01-16 openstreetmap carto kosmtik 4

PR #3622, opened 2019-01-05
Reverting lowzoom nobuilding test change

kocio-pl:
Partial revert of #3539.

- bring back the "no building" checking line in `landcover-low-zoom` data layer, since this might have performance implications (see this comment: https://github.com/gravitystorm/openstreetmap-carto/pull/3539/files#r244279150)

Test rendering is positive (nothing is broken).


kocio-pl:
It would be good if someone made a test of the performance change with and without this line. Since low zoom tiles are refreshed quite rarely on OSMF servers, it's not primary concern, but it's still worth checking.

It would be also good to have performance testing framework to make checking such things more common practice.

PR #3626, opened 2019-01-08
Change zoom level and text size for place=hamlet

### LINKED ISSUE #1984: Zoom level for hamlet
I know this has been rejected before, but I feel this needs to be looked at again.

Currenty the zoom levels and name display is making Norway look rather uninhabited. Seems wrong to have to call small settlements of less than 250 people a village instead of a hamlet. And we got a lot of them in a country which recently passed 5 million people :)

At least move hamlets out one level please. Currently it is showing at 15 and that is the same as locality, farm and neighborhood. I would think a hamlet is more important to show than these? It is supposed to be the name of a larger area than a farm.

I see a lot of places tagged as village that should have been a hamlet because of this.

jeisenbe:
Fixes #1984

- Start rendering hamlets at z14 (instead of z15) through z17
- Use a slightly larger text label for hamlets, intermediate between place=quarter and place=neighborhood / locality / farm
- Move place=hamlet to same layer as place=quarter

### Explanation:
- In areas where hamlets are common, the map may not display any place names, or too few, at z14
- Mappers sometimes misuse place=village for very small settlements so that they will render sooner. This change will reduce this incentive. 
- Hamlets are larger and more important than farms, isolated_dwellings, and "localities", since they are found in rural areas but have more than one or two families, so they should be rendered sooner and with a subtly different text.

### Test renderings: 

**Wales**
_high latitude_
z14 Trefin - current
z14-trefin-master
After
z14-trefin-hamlet

z14 Dobshill - current
z14-dobshill-master
After
z14-dobshill-hamlet

**Sixaloa, Costa Rica**
_low latitude_
z14 current rendering:
z14-sixaola-master
After
z14-sixaola-hamlets

**Syokosimo, Papua Indonesia**
_low latitude_
z14 current
z14-syokosimo-master
z14 after
z14-syokosimo-hamlet2

_Soba, Papua Indonesia_
z18 current
z18-soba-master
z18 after
z18-soba-hamlet
- There is no need to show these labels on z18 and z19

Also see additional images in https://github.com/gravitystorm/openstreetmap-carto/issues/1984#issuecomment-452296750

### DIFF
--- a/placenames.mss
+++ b/placenames.mss
@@ -372,8 +372,9 @@
 }
 
 #placenames-small::quarter {
+  [place = 'hamlet'],
   [place = 'quarter'] {
-    [zoom >= 14][zoom < 17] {
+    [zoom >= 14][zoom < 18] {
       text-name: "[name]";
       text-fill: @placenames;
       text-face-name: @book-fonts;
@@ -403,8 +404,7 @@
   }
 }
 
-#placenames-small::hamlet {
-  [place = 'hamlet'],
+#placenames-small::neighborhood {
   [place = 'locality'],
   [place = 'neighbourhood'],
   [place = 'isolated_dwelling'],
--- a/placenames.mss
+++ b/placenames.mss
@@ -372,9 +372,8 @@
 }
 
 #placenames-small::quarter {
-  [place = 'hamlet'],
   [place = 'quarter'] {
-    [zoom >= 14][zoom < 18] {
+    [zoom >= 14][zoom < 17] {
       text-name: "[name]";
       text-fill: @placenames;
       text-face-name: @book-fonts;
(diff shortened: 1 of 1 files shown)


matthijsmelissen:
> Start rendering hamlets at z14 (instead of z15) through z17

This would revert #1247.

jeisenbe:
> This would revert #1247

That PR moved hamlet to z15 along with neighborhood, isolated_dwelling,
farm, and locality (among other changes).

kocio-pl:
In general I like this idea.

jeisenbe:
Does anyone want to see more test renderings? Any questions or concerns about rendering hamlets from z14, or about no longer rendering the name label at z18 and z19?

kocio-pl:
It looks OK for me - the difference in the font size is clear and I think it makes sense to not show the label of the area at every level.

https://www.openstreetmap.org/#map=14/52.0038/21.0371

Before 

screenshot_2019-01-16 openstreetmap carto kosmtik

After

screenshot_2019-01-16 openstreetmap carto kosmtik 1

PR #3627, opened 2019-01-09
Render ref of minor roads more than once

jeisenbe:
- Modify the code for display of the ref on unclassified and residential highways:
- Add `text-placement: line` for ref of minor roads so that the ref will display more than once on a way
- Remove deprecated `text-min-distance`, replace with `text-repeat-distance` instead.

### Explanation
- Currently the reference number of an unclassified or residential highway (aka "minor highway") is only displayed once at the center of the OSM way, because `text-placement` has not been set to `line`. This PR will allow ref numbers to display every 760 pixels approximately, as intended. 
- Also, `text-min-distance` was used, but this has been deprecated in the CartoCSS documentation. This has been changed to use the new recommended option, `text-repeat-distance`, which is already used for name label rendering in most other parts of roads.mss

### Test renderings:

**Before**
refs-master
ref-10-before

**After**
refs-res-unclass
ref-10-after

### DIFF
--- a/roads.mss
+++ b/roads.mss
@@ -3019,7 +3019,8 @@ tertiary is rendered from z10 and is not included in osm_planet_roads. */
 
       text-fill: #000;
       text-face-name: @book-fonts;
-      text-min-distance: 40;
+      text-placement: line;
+      text-repeat-distance: @major-highway-text-repeat-distance;
       text-halo-radius: 2;
       text-halo-fill: @standard-halo-fill;
       text-spacing: 760;


jeisenbe:
[test.osm.xml.zip](https://github.com/gravitystorm/openstreetmap-carto/files/2740757/test.osm.xml.zip)
Updated file for testing

kocio-pl:
Another thing worth updating should be probably moving rendering them from z13+ to z15+, since they can clutter space when residential details are still not really visible, especially when refs get longer for lower highway classes (as it is in Poland for example):

https://www.openstreetmap.org/#map=13/52.2739/21.0992

tkubap_g

jeisenbe:
Residential and Unclassified refs render at z15 and higher, but without a
shield.

I think the white roads with shields in that screenshot are tertiary.

I plan to make more extensive improvements, later, but I tried to keep this
PR simple.

kocio-pl:
Thanks, so it looks like another issue. However it would be good to make real tests of residential refs too, so we can see if it could be improved too.

jeisenbe:
It's tricky to find places where the OSM way is long enough that the ref repeats due to these changes, rather than a new way being started, but there are quite a number of unclassified and even residential roads with a ref in Northern Ireland:

z16 Samsonagh Road (highway=unclassified)
https://www.openstreetmap.org/#map=16/54.3505/-7.7417
Current master
- Ref renders once, and does not orient to way
samsonagh-master1

After changes
- Renders twice (in this 1000 x 500 pixel image), about half as often as the name
z16-samsonagh-refs-1

highway=unclassified B523, in Londonderry/Derry
- There are three different ways with different names
https://www.openstreetmap.org/#map=15/54.9924/-7.3110
z15 Current
- Only one of the 3 refs renders, due to collisions
londonderry-master

z15 After
z15-londonderry-refs

z19 Current
- at this zoom level we can see just one of the ways
z19-chapel-road-master

After
z19-chapel-road-refs

kocio-pl:
Does this code change rendering ref orientation from "landscape" to "along the road"? It's not clear for me from the description, but it looks like this in latest examples.

Adamant36:
I take it its from adding `text-placement: line;`

kocio-pl:
I guess it's better solution - refs are less visible that way and for such small roads it makes sense to not stand out too much.

kocio-pl:
Example I have found:

https://www.openstreetmap.org/#map=17/52.25744/21.03584

Before

o5y2rqix

After

_gnot3cc

jeisenbe:
I’ve noticed that refs are rendered before names (both now and with this
PR). It might be better to change this for highway=residential and
highway=classified, so that the name is rendered first and the ref is only
added if there is room.

For major highways it is probably correct to prefer the ref to the name if
there is not space for both (and this is less likely to happen for
long-distance roads), but for small residential highways, I would rather
see the name of there is only room for one or the other.

But that can be decided in another PR.

kocio-pl:
Thanks, I like that refs are less intrusive going along the line. However further tuning is also possible.

jeisenbe:
Thanks! Now I can test adding refs for highway=track (eg US Forest service roads, which are often unnamed)

PR #3636, opened 2019-01-12
[WIP] Switch to ST_PointOnSurface

### LINKED ISSUE #1644: Switch to ST_PointOnSurface?

matthijsmelissen:
Resolves #1644.

This is a necessary step towards vector rendering (see #1644).

This is a second trial of #3228, based on the word by @pnorman in #3233.

This is a first step, once we have this than we can probably merge the point and poly layers (for example amenity-points and amenity-points-poly).

### DIFF
--- a/addressing.mss
+++ b/addressing.mss
@@ -25,7 +25,6 @@
         }
       }
     }
-    text-placement: interior;
     text-face-name: @book-fonts;
     text-fill: @address-color;
     text-halo-radius: @standard-halo-radius;
@@ -56,7 +55,6 @@ no official postal addresses) */
   [zoom >= 14][way_pixels > 3000],
   [zoom >= 17] {
     text-name: "[name]";
-    text-placement: interior;
     text-face-name: @book-fonts;
     text-fill: #444;
     text-halo-radius: @standard-halo-radius;
--- a/amenity-points.mss
+++ b/amenity-points.mss
@@ -52,7 +52,6 @@
       marker-fill: @man-made-icon;
     }
     marker-fill: @accommodation-icon;
-    marker-placement: interior;
     marker-clip: false;
     [access != ''][access != 'permissive'][access != 'yes'] {
       marker-opacity: 0.33;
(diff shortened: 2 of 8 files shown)


matthijsmelissen:
@pnorman Would you be able to review this?

imagico:
The comments from #1644 regarding label placement quality still apply.

jeisenbe:
I’ve read through the comments on #1644, where @imagico concluded with:

“My main point here is that ultimately ST_PointOnSurface() is going to be a
dead end regarding quality label placement but Mapnik likely is as well so
it really does not matter.”

Is this still correct?

imagico:
Can we close this since most of it is superseded by #3712 and the rest would probably be easier to implement with new PRs?

Issue #3395, opened 2018-09-14
Icon colour categorization (Meta issue)


polarbearing:
There are a lot of new icons being added recently, and often the colour has to be decided. Doing that on a case-by-case basis apparently leads to inconsistencies, thus this issue tries to collect definitions and decisions, which I can refine in the top post according to the discussion evolving. I focus on the principal colours and leave out fine tuning of labels etc. 

### amenity-brown:  = memorials:, culture:, public-service:
734a08; ![#734a08](https://placehold.it/15/734a08/000000?text=+)
financial services (bank, ATM, bureau_de_change, casino)
unattended shelters and installations (amenity_shelter,  bbq)
unattended items (post_box, recycling, telephone, picnic_table, waste_disposal, bench, waste_basket)
artwork, memorials
culture (theatre, cinema, museum, community_centre, arts_centre
public_bath, shower, toilets, drinking_water, firepit
public service (fire station, embassy, courthouse, social_facility, town_hall, post_office, prison
tourism=office (don't see the rendering)
viewpoint
historic things (castle, fort, archaeological site etc)
_problematic:_ car_wash, car_service

### transportation-icon:  = accommodation-icon:
0092da; ![#0092da](https://placehold.it/15/0092da/000000?text=+)
Transport related (public transport, vehicle rent, taxi, filling/charging station, toll, elevator, parking)
Touristic accommodation (hotel, motel, hostel, tourism-apartment, guest house, camping/caravan)
serviced shelters (alpine_hut,
wilderness_hut appears more to be an unserviced shelter? 

### shop-icon: 
ac39ac; ![#ac39ac](https://placehold.it/15/ac39ac/000000?text=+)
_Shop (wiki) A place selling retail products or services._
shops, markets
driving_school
hygiene services (hair dresser, nail studio)

### health-color: 
da0092; ![#da0092](https://placehold.it/15/da0092/000000?text=+)
Hospital, Pharmacy, doctors, vet, clinic

### gastronomy: PR in prep for own colour
_Services where ready-to-eat food and drinks are served_ 
restaurant, fastfood, pub, cafe, bar, 
including those servings in a cultural context (nightclub)

### religious black 
000000 (needs a variable on top of file) ![#000000](https://placehold.it/15/000000/000000?text=+)
Place of worship, man-made=cross, wayside_cross

### power black
"black" (needs a variable on top of file) ![#000000](https://placehold.it/15/000000/000000?text=+)
generator

### water-color:
[...]

PR #3642, opened 2019-01-16
Add "religious-icon" as a color variable

jeisenbe:
Related to #3395

- Add `@religious-icon: #000000;` 
- Use this variable for place_of_worship, historic=wayside_cross and man_made=cross

In the future this could also be extended to other religious icons, such as wayside_shrine or other religious objects if they become defined in the database. 

Test rendering is not shown, because the cartography is unchanged. 
However, I have tested this commit and found that it works without problems.


mmelissen:
Thanks!

matthijsmelissen:
Thanks!

PR #3643, opened 2019-01-16
Add "barrier-icon" color variable with #3f3f3f for barriers

jeisenbe:
Related to #3395

- Add `@barrier-icon` color variable with value of `#3f3f3f `
- Change barriers marker-fill to use `@barrier-icon` instead of `#3f3f3f `

Note: 
-bollards, logs and blocks are unchanged, because these currently use a much lighter shade of gray.
- 'barrier_gate' does not use marker-fill, because the icon uses two colors (#3f3f3f plus a white halo)

In the future this variable could be changed to use @man-made-icon if we wish to reduce the number of shade of gray.

Test rendering is not shown, because the cartography is unaffected.
I have tested this commit and found that it works without changing the rendering.

### DIFF
--- a/amenity-points.mss
+++ b/amenity-points.mss
@@ -17,6 +17,7 @@
 @office: #4863A0;
 @man-made-icon: #666666;
 @advertising-grey: @man-made-icon;
+@barrier-icon: #3f3f3f;
 @landform-color: #d08f55;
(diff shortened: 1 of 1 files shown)


jeisenbe:
Edited; I meant to use #3f3f3f for barriers. Too many shades of gray...

Issue #3611, opened 2018-12-29
Different geometry decisions (point/line/polygon) on database re-import


sommerluk:
Currently there is no database re-import planned. And apparently also no urgent need to do so. Anyway it might be useful to have a list of geometry decisions (point/line/polygon) that we might maybe change if in the future we would make a database re-import.

- `barrier=toll_booth` as _line_ geometry (reference #3249 and #3244)

kocio-pl:
Does it cover import issues like #3606?

jeisenbe:
allotments=plot should imported as a polygon, so that we have the option of solving #432 "render ref numbers for allotments=plot"

HolgerJeromin:
Healthcare #3639

bhousel:
In case it helps, iD's areaKeys list is published here:  https://github.com/osmlab/id-area-keys

The main file in there [`areaKeys.json`](https://github.com/osmlab/id-area-keys/blob/master/areaKeys.json) contains a dictionary of keys that imply polygon, and subkeys of exceptions that imply linestring.

matthijsmelissen:
@kocio-pl I think this might be a bit too hasty. Wouldn't it be better to first discuss within our own repository if and when a database reload is desired, before asking externally?

kocio-pl:
Since this is about synchronizing efforts and we're dependent on them, it's good to know both sides IMO.

jeisenbe:
`boundary=national_park` and `boundary=protected_area` should be added to `local_polygon_values`

sommerluk:
> `boundary=national_park` and `boundary=protected_area` should be added to `local_polygon_values`

Cross-reference #1141

jeisenbe:
"Golf=*" as polygon:
'rough', 'fairway', 'driving_range', 'water_hazard', 'green', 'bunker'

Possibly 'tee'?

jeisenbe:
The key "emergency" should probably be imported as polygons.

jeisenbe:
For reference, here are the list of keys that the ID editor considers to be areas (polygons) when drawn as closed ways, with the exceptions.

"advertising" (except "billboard")
"aerialway": (Except "cable_car", "chair_lift", "drag_lift", "gondola", "goods", "magic_carpet", "mixed_lift":, "platter", "rope_tow", "t-bar")
"aeroway": (Except "runway", taxiway")
 "allotments"
"amenity": (except "bench")
"area:highway"
[...]

PR #3644, opened 2019-01-16
[WIP] Add healthcare to list of polygons in lua transformations

### LINKED ISSUE #3639: Healthcare tags do not render on closed ways without area=yes
### Expected behavior

### Actual behavior

### Solution

jeisenbe:
**Fixes #3639**

- Add 'healthcare' to list of polygon objects in openstreetmap-carto.lua

## Explanation:
- This will allow closed ways ("areas") that are tagged with "healthcare=*" to be imported as polygons, so that they can be rendered properly
- Currently imported objects will not be immediately affected, but newly added and edited objects should render properly now, and all should render properly after database reload.

## Test rendering:

Before
healthcare-areas-before

After
- Rendered after reloading database. Before database reload the rendering was unchanged.
healthcare-areas-after

### DIFF
--- a/openstreetmap-carto.lua
+++ b/openstreetmap-carto.lua
@@ -14,6 +14,7 @@ local polygon_keys = {
     'building',
     'building:part',
     'harbour',
+    'healthcare',
     'historic',
     'landuse',
     'leisure',


jeisenbe:
I see that `'amenity_hospital'` is rendered with a polygon fill in societal_amenities color, but `healthcare_hospital` was not added to landcover.mss

Was this intentional, or an oversight in the PR that added support for `healthcare=` ?

jeisenbe:
Hmm... adding healthcare_hospital and healthcare_clinic to landcover.mss does not work to render polygon fill and outlines for these features, even though there are already two lines in project.mml.

This is what I tried:
```
+++ b/landcover.mss
@@ -570,7 +570,9 @@
   }
 
   [feature = 'amenity_hospital'],
+  [feature = 'healthcare_hospital'],
   [feature = 'amenity_clinic'],
+  [feature = 'healthcare_clinic'],
   [feature = 'amenity_university'],
```
And in project.mml the landcover layer has (lines 124 and 150)
```
              ('healthcare_' || (CASE WHEN tags->'healthcare' IN ('clinic' ,'hospital') THEN tags->'healthcare' ELSE NULL END)) AS healthcare,
```
(FROM planet_osm_polygon)
WHERE ...
```
             OR tags->'healthcare' IN ('clinic', 'hospital')
```
I've also added 'healthcare' to openstreetmap-carto.lua already, as in the first commit in this PR.

Why are these features still not rendering with this code?

Is this due to "healthcare" being in the hstore column rather than in a standard column?

imagico:
I am not sure if you want to

* initiate a database reload on the OSMF servers
* roll out this change without a database reload (leading to inconsistencies in rendering between pre-existing and newly added or modified features)
* implement this change in OSM-Carto without rolling it out on OSMF servers (leading to rendering differences between osm.org and test environments)

I would be against the second and third options for the mentioned reasons.

matthijsmelissen:
I agree we cannot merge this now because there is no database coming up, and I wouldn't like our code to divert from the code on production. Let's add it to #3611 and take it into consideration during the next reload. For now I'm going to reject this PR.

jeisenbe:
The problem is, this feature is already rendered, but only on nodes or on
polygons tagged with area=yes, so the current situation is also not good
for tagging feedback.

Unless we make the more extreme choice of reverting the PR that added
healthcare rendering, this would be better than the other options.

For mapper feedback, new and edited features are most important. So it may
nit nene a large problem if old polygons only tagged with “healthcare” are
not yet rendered for a few months.

jeisenbe:
Would you like to stop rendering healthcare on all polygons, or entirely,
for the time being?

matthijsmelissen:
What do you think is the best way forward, @jeisenbe?

kocio-pl:
I would prefer to talk with OWG first and ask when they would be ready to reimport database. If we know it, we can plan our action, hopefully changing all the current hstore objects into plain columns.

jeisenbe:
Sorry, my first response was sent before last couple of hours of emails
loaded, so I didn’t realize the PR was rejected.

@imagico, do you think this feature should be removed until the next
database reload? Or would it be acceptable to only render “healthcare=“
tags on nodes for now?

imagico:
I have not reviewed #3498 so i cannot give you a full answer to the pros and cons to revert it (i see other problems with it as well) but yes, a change that would depend on this PR to be consistently rendered would IMO need to wait for a database reload.

mboeringa:
Do we actually need this?

It seems that e.g. iD already automatically double tags with additional "healthcare" tag for the most important amenities:
afbeelding

Also see this: 
https://github.com/openstreetmap/iD/commit/bfda2c543803961d569fd94d86fd2e240b6c1d26

In addition, healthcare tagging is a bit of a mess, with two competing schemes:
https://wiki.openstreetmap.org/wiki/Key:healthcare
https://wiki.openstreetmap.org/wiki/Proposed_features/Healthcare_2.0

While the first one seems used in iD, the second seems at least partly used by HOT:
https://github.com/hotosm/presets/blob/master/HDM.xml
afbeelding

**health_facility:type** is a key from the "officially" "abandoned" Healthcare 2.0 proposal...

This essentially means you would need to select:
**amenity IN ('hospital') OR healthcare IN ('hospital') OR health_facility:type IN ('hospital')**
to be reasonably sure to have most hospital facilities selected.

By introducing the secondary "healthcare" tag in rendering, there is a slight risk people will no longer double tag with **amenity=***, meaning all styles would need to start employ the above query if they want to catch all facilities of a certain type, although looking at the HOT preset, it also seems to promote or use double tagging with at least **amenity=***, so that this issue may not be to big in reality.

PR #3646, opened 2019-01-17
Fix layering order for "construction=*" highway categories

jeisenbe:
Related to #3579 

- Use osm_id as the last way to determine rendering order for roads-fill, so that different highway=construction roads will always layer in the same order, at different zoom levels. This change will take effect immediately.
- Add  "construction=" values to the z-order table in openstreetmap-carto.lua, so that the different categories of highway=construction will render in proper order (eg construction=motorway on top, construction=service on bottom). This will take effect on database reload, and for new or edited features. 

### Explanation

Highway=construction is currently set to z = 10 in the z-order table, so that it will render below almost all highways and paths. However, the different types of construction=*, such as construction=motorway, construction=trunk_link, construction=service, all have the same z value, so they are laid in apparently random fashion. The layering order often changes from one zoom level to the next. 

These two commits will eventually fix this problem.

The change to project.mml, adding osm_id as a final fall-back for ordering, will at least cause all highway=construction to rendering in a consistent way between zoom levels, effective immediately.

The improvement to rendering for different classifications of highway=construction is part of the lua transformations, and will take effect gradually for new and edited features, and then for old features when the database is next reloaded. 

EDITED: the lua transformation changes have been removed for now, because a database reload is not yet planned

EDITED: These were with the first commits

Before
A6 - New Trunk Interchange - Northern Ireland
https://www.openstreetmap.org/#map=16/54.7629/-6.4847
- construction=trunk_link above construction=trunk
z15 Before
z15-new-trunk-links-master
z16 Before
z16-new-trunk-links-master

After z15 
z15-new-trunk-links-after
After z16 
z16-new-trunk-links-after

Murry's Island - Northern Ireland
https://www.openstreetmap.org/#map=15/54.7734/-6.5189
- New trunk road crosses an under-construction minor road and an under-construction secondary road (right) and an under-construction tertiary road (left)
z15 Before
z15-murrays-island-construction-master
z15 After
z15-murrays-island-construction-after

Castledawson - new residential area
https://www.openstreetmap.org/#map=16/54.7692/-6.5512
z16 Before
z16-annaghmore-construction-master
z17 Before
z17-annaghmore-construction-master

After z16
z16-annaghmore-construction-after
After z17
z17-annaghmore-construction-after

EDITED: See new test renderings below


imagico:
Support for ordering by osm_id, regarding z_order (and therefore database related) changes see https://github.com/gravitystorm/openstreetmap-carto/pull/3644#issuecomment-454738080.

jeisenbe:
I could add a new SQL-based order just in the project.mml file for
roads-fill only.

But I thought that changing the z-order list in lua was better for
rendering speed.

Thoughts @imagico, @pnorman?

imagico:
Your choice of approach is perfectly fine IMO but it will have to wait until the next database reload.  I don't think adding complexity to the road layers' SQL to circumvent this would be a good idea.

Just have a bit of patience here - the last database reload was in August - https://github.com/openstreetmap/chef/issues/155#issuecomment-413818970

and for previous discussion about database reload patterns see #1286.  While this is a problem that would be good to fix it is not a high impact problem that frequently causes serious problems for map users or mappers.

jeisenbe:
Ok, I’ve removed the z-order changes. So the PR only adds ordering by
osm_id now

jeisenbe:
New tests with the current commit:
https://www.openstreetmap.org/#map=6/54.7629/-6.4847
- layering order is now consistent between different zoom levels

z17
z17-osm_id-layers

z16
z16-osm_id-layers

z15
z15-osm_id-layers

jeisenbe:
Thanks!

I plan to submit a new PR with the lua changes, at the time of the next
database reload.

Issue #2524, opened 2016-12-21
Label scaling incorrect with scale factor


pnorman:
This is from Germany with scale-factor 6, the resulting image scaled by 1/6 to get a result which should look the same except for antialiasing, quality of rendering, and PNG patterns

![image](https://cloud.githubusercontent.com/assets/1190866/21406427/7185471e-c780-11e6-84b4-aa32776e2994.png)

On osm.org the label doesn't even get placed at that scale
![image](https://cloud.githubusercontent.com/assets/1190866/21406451/965dc700-c780-11e6-8a25-bc45d266e850.png)

At a 4x higher scale, this is what it looks like

![image](https://cloud.githubusercontent.com/assets/1190866/21406473/b63f3860-c780-11e6-9115-6046c47a9f86.png)

cc @woodpeck

imagico:
> This is from Germany with scale-factor 6, the resulting image scaled by 1/6 to get a result which should look the same except for antialiasing, quality of rendering, and PNG patterns

Assuming that by _scale-factor 6_ you mean rendering at six times the normal resolution (scale factor to me usually means [something different](https://en.wikipedia.org/wiki/Mercator_projection#Scale_factor)) that seems not quite correct since we make a lot of use of `way_pixels` for labeling and this i would assume varies with rendering resolution.

pnorman:
> Assuming that by scale-factor 6 you mean rendering at six times the normal resolution

scale factor is a mapnik parameter: https://github.com/mapnik/mapnik/wiki/Scale-factor

> that seems not quite correct since we make a lot of use of way_pixels for labeling and this i would assume varies with rendering resolution.

I see a couple options

1. Use way_area directly. We used to do this, but it involves a lot of duplication with a selector needed for each zoom level. We might be able to simplify this if we can use the mapnik parameter zoom in a selector, but I don't think that's possible. @rory had to do something like this when adapting osm-carto to vector tiles

2. Alter the queries so their results do not depend on scale factor. This is my preferred approach, and what should have been done originally, but I don't think we adequately tested output for higher resolution devices.

imagico:
The underlying cause for the whole mess seems to be that mapnik does not properly separate between map scale and rendering resolution.  It makes sense for compact code in the style to normalize way_area with the map scale but it obviously does not make much sense to normalize with the rendering resolution if you want to be able to generate the same map scale in varying resolutions.

math1985:
Do we want to support scale_factor at all?

[...]

Issue #1094, opened 2014-10-26
way_area vs way_pixels


matkoniecz:
way_area is not an intuitive unit, it is quite easy to convert it in SQL to in my opinion much better way_pixels[1]. way_pixels also converts nicely from zoom level to zoom level allowing less complicated conditional structures. I propose to replace all usage of way_area in .mss files by way_pixels.

It would affect only amenity-points (10 occurrences - for leisure=playground and displaying names of water bodies).

Related to #710

[1]`way_area/(!pixel_width!*!pixel_height!) AS way_pixels`

Issue #703, opened 2014-07-04
The use of way_area


Rovastar:
I thought we could have a discussion in one place about the use of the way_area. It keep cropping up on many issues as a solution but until recently it is seldom used. 

Many issues on the carto style can be improved if we use way_area to define if they are displayed or not at each zoom level.

e.g. we have

``` mss
  [leisure = 'golf_course']::leisure,
  [leisure = 'nature_reserve']::leisure {
    [way_area >= 150000][zoom >= 14],
    [way_area >= 80000][zoom >= 15],
    [way_area >= 20000][zoom >= 16],
    [zoom >= 17] {
```

Now should we use this more. Personally I think we should it makes logical sense and will improve the map no end. It can be used in so many cases from displaying names for museums, lakes, recycling centres, etc to massive nature reverses being displayed at level 5.

Doing the catch all of x features should be rendered on y level as we do at the moment is not a best solution as bigger things are normally more important and should have more visibility.

The issues against this seem to be it will be computational heavy but I don't understand how much. Obviously there is a some.
And probably it would add more confusion/verboseness to the map code.

If we are to do this I propose having 
a) consistent values
I think in gneral x size will work on y zoom level
e.g. 

``` mss
    [way_area >= 150000][zoom >= 14],
    [way_area >= 80000][zoom >= 15],
    [way_area >= 20000][zoom >= 16]
```

are these the most accurate and optimized areas for these zoom.
b) have them as variable (well constants) defined
e.g.
way_area_for_z14 = 150000, etc
and use these through the style.

c) using these throughout the carto style for existing features and new features.

Any thoughts welcomed.

PR #3648, opened 2019-01-18
Render tourism outlines one zoom level sooner for small zoos and theme parks

jeisenbe:
Partially reverts #2835

- From z13 to z16, render tourism area outlines when way_pixels > 180 instead of the current limit of 750
- Removed redundant code that rendered thinner outlines when way_pixels < 60; this has not been used since #2835

### Explanation:

Prior to #2835, all zoos and theme parks were rendered from z10 if the way_pixels was > 20. This meant that very small features, for example 5 pixels tall and 5 pixels wide, would render. Those with less than way_pixels between 20 and 60 were rendered with a thin 1 pixel border, which was difficult to see and looked quite different than the wide, double border used for larger areas (as seen now).

#2835 improved this situation by only rendering tourism area outlines when the waypixels was > 750. This is 1/4 of 3000 way_pixels, the number used to show the name of the area, so the outline would show exactly 1 zoom level sooner than the name label. Very small areas were hidden.

However, this caused some mid-sized areas to disappear; for example a 30 by 20 pixel area is hidden, even though this is 3 times larger than an icon. This looks strange when there are two tourism areas adjacent, a situation sometimes found in areas with more than one theme park or zoo next to each other (as see in Singapore, below). If one of the areas is slightly smaller, it could disappear a zoom level sooner than it's neighbor, leaving a gap in the map. 

I initially considered rendering all zoos and theme parks from >=z13, because that is where all landcover colors and patterns are now shown. However, I found that it was not helpful to show very small areas less than 100 waypixels. The limit in this PR is now set to 180, which will cause small zoos and theme parks to render approximately 1 zoom levels sooner than before. This solves the problems without losing most of the benefits of #2835

### Test renderings:

**Sentosa Island, Singapore**
- Universal Studios (large) next to KidZania (small) theme park
https://www.openstreetmap.org/#map=15/1.2513/103.8229

z15 Current
z15-sentosa-master

z15 After
z15-sentosa-after

z16 (Same)
z16-universal-studios-660033

**Jurong Bird Park, Singapore**
- Zoo in an otherwise industrial area
https://www.openstreetmap.org/#map=14/1.3229/103.7269

z13 Current
z13-jurong-bird-park-master
z13 After
z13-jurong-bird-after

z14 (unchanged)
z14-jurong-bird-same

z15 (unchanged)
z15-jurong-660033

**Singapore Zoo and Night Safari**
https://www.openstreetmap.org/#map=14/1.4043/103.7957

z13 Current
z13-singapore-zoos-master
z13 After
z13-signapore-zoos-after

z14 (Same)
z14-singapore-zoos-same

**Bangor, Northern Ireland**
https://www.openstreetmap.org/#map=15/54.6637905/-5.6726521
- "Pickie Fun Park" is long and narrow, so it's rather challenging to render with an outline

z13 current
z13-bangor-master
z13 after
z13-bangor-after

z14 (same)
z14-bangor-master

z15 (same)
z15-pickie-fun-park-master

**Damhead Miniature Railway**
https://www.openstreetmap.org/#map=17/55.1115842/-6.5977664
- There are 4 or 5 of these railway theme parks in rural Northern Ireland, all are similar to this one.

z15 (Same)
z15-damhead-master

z14 Current
z14-damhead-master
z14 After
z14-damhead-after

z13 (same before and after)
z13-damhead-railway

_[Before #2835 this would have been the rendering at z13:]_
_z13 Damhead railway rendering with 20 way_pixels limit:_
z13-damhead-way_pixels-20-limit

### DIFF
--- a/landcover.mss
+++ b/landcover.mss
@@ -863,30 +863,29 @@
 }
 
 #tourism-boundary {
-  [tourism = 'zoo'][zoom >= 10][way_pixels >= 750],
-  [tourism = 'zoo'][zoom >= 17],
-  [tourism = 'theme_park'][zoom >= 10][way_pixels >= 750],
-  [tourism = 'theme_park'][zoom >= 17] {
-    a/line-width: 1;
-    a/line-offset: -0.5;
-    a/line-color: @tourism;
-    a/line-opacity: 0.5;
-    a/line-join: round;
-    a/line-cap: round;
-    [zoom >= 17],
-    [way_pixels >= 60] {
+  [tourism = 'zoo'],
+  [tourism = 'theme_park'] {
+    [zoom >= 10][way_pixels >= 750],
+    [zoom >= 13][way_pixels >= 180],
+    [zoom >= 17] {
+      a/line-width: 1;
+      a/line-offset: -0.5;
+      a/line-color: @tourism;
+      a/line-opacity: 0.5;
+      a/line-join: round;
+      a/line-cap: round;
       b/line-width: 4;
       b/line-offset: -2;
       b/line-color: @tourism;
       b/line-opacity: 0.3;
       b/line-join: round;
       b/line-cap: round;
-    }
-    [zoom >= 17] {
-      a/line-width: 2;
-      a/line-offset: -1;
-      b/line-width: 6;
-      b/line-offset: -3;
+      [zoom >= 17] {
+        a/line-width: 2;
+        a/line-offset: -1;
+        b/line-width: 6;
+        b/line-offset: -3;
+      }
     }
   }
 }


polarbearing:
Test renderings seem plausible. 

Could somebody educate me where the way_pixels values are generated?
I found them only being _used_ in this repo, but not _defined_. 
I found a formula in #1094 and #703, and I searched cartoCSS and mapnik docs without success.

polarbearing:
The ```pixel_height``` appears to come from PostGIS, but so far I did find neither ```way_area``` nor ```way_pixels``` being calculated. I would assume @pnorman could know it?

matthijsmelissen:
`way_pixels` comes from project.mml, for example `way_area/NULLIF(!pixel_width!::real*!pixel_height!::real,0) AS way_pixels`.

`pixel_width` and `pixel_height` in turn come from Mapnik:
https://github.com/mapnik/mapnik/wiki/PostGIS#other-tokens

jeisenbe:
So way_area is the actual area of the polygon in e.g. square meters, and
way_pixels is approximately the area of the polygon in pixels, correct?

matthijsmelissen:
> So way_area is the actual area of the polygon in e.g. square meters, and
way_pixels is approximately the area of the polygon in pixels, correct?

Correct! Forgot to mention, but way_pixels is calculated by [osm2pgsql](https://wiki.openstreetmap.org/wiki/Osm2pgsql#Special_.27tags.27).

imagico:
Note way_area is in Mercator square meters, not in actual square meters.

For making styling decisions regarding label size etc. normalization with `!pixel_width!`/`!pixel_height!` is not ideal because it is not invariant w.r.t. rendering resolution.  See #2524.  Normalization with `!scale_denominator!` would make more sense.

jeisenbe:
Should we redefine all occurrences of way_pixels as

```
way_area/!scale_denominator
```

?

matthijsmelissen:
Yes, would be great if that works.

jeisenbe:
It’s a shame that we don’t have a fast way to query the minimum dimensions
of a polygon. I think this PR looks good with zoos and theme parks that are
nearly square or circular (say width:height < 2) but it doesn’t work as
well with the example of the long, thin theme park.

Still, I find it odd to have land-color “holes” at z14 or z13 on a rendered
area that is actually 100% tagged with some sort of landcover or landuse,
so I prefer to see the area rendered, even when it looks more like a fill
color than an outline.

jeisenbe:
I've tested the way_pixels calculation change. 

The scale_denominator = map_width_in_metres/ (map_width_in_pixels * 0.00028m)

Because 0.28mm is the standardized pixel size. So at z17 scale_denominator=4265.4591677, which means that one standard pixel is approximately the same as one square meter at this zoom level (near the equator). See:  https://github.com/openstreetmap/mapnik-stylesheets/blob/master/zoom-to-scale.txt

So to convert way_area to pixels, divide by 0.28mm squared [=0.0000000784 meters] multiplied by the scale_denominator squared.
EG:

`way_area/NULLIF(.0000000784*POW(!scale_denominator!, 2),0) AS way_pixels`

I've pushed a commit to the border-labels branch, i.e. PR #3652 after testing.

imagico:
Thanks - to make the formula better understandable i would formulate it as

`way_area/NULLIF(POW(!scale_denominator!*0.001*25.4/90, 2),0)`

(that is millimeter-to-meter*millimeter-per-inch/assumed-ppi)

And i would not use it selectively for some features but roll it out universally in those cases where way_area is used in relation to the map display scale.

jeisenbe:
> millimeter-to-meter*millimeter-per-inch/assumed-ppi

That sounds sensible by an American perspective. I didn’t think
pixels-per-inch was the standard in other countries.

> roll it out universally in those cases where way_area is used in relation
to the map display scale.

I’ll do a PR for this, next

jeisenbe:
I tested with `way_area/NULLIF(POW(!scale_denominator!*0.001*25.4/90, 2),0)` but found that this is not precisely the same:

Current way_pixels calculation (based on pixel_width and pixel_height):
- Way_pixels = 478440. ...
boondina-waypixels-current

With `way_area/NULLIF(POW(!scale_denominator!*0.001*25.4/90, 2),0)`:
- Way_pixels = 470938. ...
boondina- pow scale_denominator 0 001 25 4 90 2 0

Compared to `way_area/NULLIF(POW(!scale_denominator!*0.001*0.28*, 2),0) AS way_pixels`
- way_pixels = 478443. ...
boondina- 0000000784 pow scale_denominator 2

In fact the assumed "ppi" for mapnik is based on a 0.28mm pixel, so it should be approximately 90.71 standardized "ppi" (pixels per inch):

`way_area/NULLIF(POW(!scale_denominator!*0.001*25.4/90.7, 2),0) AS way_pixels`:
- Way_pixels = 478292. ...
boondina- pow scale_denominator 0 001 25 4 90 7 2 0

I'm not convinced  that `way_area/NULLIF(POW(!scale_denominator!*0.001*25.4/90.71, 2),0)` is more intuitive than `way_area/NULLIF(POW(!scale_denominator!*0.001*0.28, 2),0)` or  `way_area/NULLIF(POW(!scale_denominator!*0.00028, 2),0)`

imagico:
I am sorry - my bad.  I just assumed mapnik - like almost all other programs converting between pixel and real world units - would assume a certain integer ppi (historically often 72, recently more often 90 or 92).

jeisenbe:
It looks like @matthijsmelissen is not in favor of the cartographic changes, and no one else has spoken out in support. I agree that the long and narrow areas do not look good with this change (though it works fine with shapes closer to square or circular) so I think this PR can be closed.

PR #3652, opened 2019-01-21
Improve rendering of text labels on admin and nature-reserve borders

### LINKED ISSUE #3152: Multipolygon relations: not all borders are labeled
If a multipolygon relation includes multiple discrete parcels of land, the name of the relation is ONLY drawn around the borders of the largest parcel.  This is confusing in situations where some of the land might be far out of view of the biggest chunk.

I supposed I can understand why only one "area label" is drawn per relation; it might look weird if there were fifty parcels visible, all with the same label... but I think the borders, which only appear when zoomed in, should be labeled on every parcel.

multipolygon rendering annotated

Attached file comes from here: https://www.openstreetmap.org/#map=15/45.7934/-121.6350

### LINKED ISSUE #3649: Administrative boundary text labels are displayed too often
### Expected behavior
- Administrative boundary text labels should repeat along the border a couple of times per screen, in a regular fashion, similar to the behavior of labels on waterways and roads 
- Text labels should not be displayed twice in row when a new OSM way starts or if the way curves back on itself

### Actual behavior
- Boundary text labels are often displayed several times in succession
- The spacing does not seem consistent

### Likely cause and solution
- the code in admin.mss does not contain `text-spacing:` or `text-repeat-distance:` as used for roads and waterways. Adding these missing properties should improve the rendering

**Near Wamena, Papua, Indonesia** (_Screenshots_)
https://www.openstreetmap.org/#map=17/-4.07715/138.97202
walelagama-boundary-label-x2

https://www.openstreetmap.org/#map=17/-4.08183/138.95416
wamena-x3-wesaput-border

**Damhead Railway, Northern Ireland** (_Screenshot from openstreetmap.org)
https://www.openstreetmap.org/#map=17/55.11269/-6.59828
z16-damhead-borders-2x

**Canberra, Australia** (_Exported from Kosmtik_):
z18-drakeford-dr-circle-oneway-cycleway

jeisenbe:
Fixes #3152 "Multipolygon relations: All borders should be labeled"
Fixes #3649 "Administrative boundary text labels are displayed inconsistently"

- 
- Add `text-largest-bbox-only: false` for nature-reserve-text and admin-text
- Add `text-spacing: 750`
- Add `text-repeat-distance: 250`
- Add `text-margin: 10`

### Explanation:

Currently the labels that follow admin borders at z16 and higher, and the labels along nature-reserve and national-park borders, are not shown on all borders of multi polygons. This is a problem when an country has an enclave within another country, or a national park consists of two parts.

Also, Administrative boundary text labels are displayed inconsistently. Boundary text labels are often displayed several times in succession, but sometimes the labels are more than 1600 pixels apart, even in areas with no other objects mapped. When there is more than one level of administrative border that shares the same OSM way, the labels are often rendered immediately adjacent, with no white space or gap between. 

The changes above will address these issues.

1) `text-largest-bbox-only: false` will cause the text labels to render along all borders of a multi-polygon, rather than on the largest only. 
2) `text-spacing: 750` will cause the labels along border to repeat approximately every 750 pixel (if they are not blocked by earlier-rendered objects, or tight bends in the border), which will usually render one or two labels per full screen.
3) `text-repeat-distance: 250` prevents the labels from being rendered close than about 250 pixels apart. This fixes a problem where multiple copies of the label will be rendered right next to each other, if the border is cut off by a metatile boundary.
- This also has the effect of preventing the border labels from rendering if they are within 250 pixels of identical text, such as a text label of a place that has exactly the same name as the administrative boundary.
4) `text-margin: 10` prevents two different text labels from rendering too close together. This also prevents the text labels from rendering in areas that are crowded with other icons and text labels. 
- Because admin-text and nature-reserve-text are the last 2 layers to render (except for the low-priority amenities layer), this will not prevent the rendering of other icons or text labels.

### Test renderings with links to the example places:
- These renderings were made with `metatile: 8` to match the live rendering on openstreetmap.org
- I have used screenshots with Grab on MacOS (rendering with Kosmtik and Safari browser), rather than the export function

**Wamena, Papua, Indonesia** (_Screenshots for current_)
https://www.openstreetmap.org/#map=17/-4.07715/138.97202
- Repeated text labels along border. This area is tagged correctly, there is no duplication in the data. 
Current rendering:
walelagama-boundary-label-x2
After:
z17-walelagama-after

https://www.openstreetmap.org/#map=17/-4.08183/138.95416
- Currently the border text labels are repeated, due to what appears to be a Mapnik bug related to polygons that cross metatile borders twice and therefore have two disconnected sections in the same metatile. This admin_level=6 area is tagged correctly.
Current:
wamena-x3-wesaput-border
After:
z17-wamena-after

**Andorra de la Vella - village boundary**
https://www.openstreetmap.org/#map=16/42.5096/1.5346
- The administrative boundary name currently renders twice, and right in a busy part of the map
- The changes in this PR prevent the labels from rendering twice in a row, and `text-margin: 10` causes them to be moved out into uncrowded parts of the rendering
z16 Current:
andorra-la-vella-2x-master-16 42 5096 1 5346
After:
z16-andora-la-vella-after

**Yahukimo/Jayawijaya border**
https://www.openstreetmap.org/#map=16/-4.2006/139.0261
- Two admin-level 5 areas and two admin-level 6 areas meet here (There is a 3rd admin_level=6 area that is not yet mapped)
- In the current rendering, the labels are too close together. There is almost no space between them.
- Adding `text-margin: 10` creates enough space. This can be done safely on these two layers, which are almost the last 2 in the project.mml (except for low-priority amenity points)
z16 Current
z16-aso-tipo-jayawijaya-yahukimo-master
z16 After
z16-asotipo-jayawijaya-yahukimo-after

**Central Luxembourg**
- Currently a couple of the labels are duplicated. This is fixed by these changes
https://www.openstreetmap.org/#map=16/49.6077/6.1190
Current z16
central-luxembourg-z16 49 6077 6 1190
After
central-luxembourg-z16 49 6077 6 1190

**National Park Multipolygon at Spain/Andorra Border**
https://www.openstreetmap.org/#map=16/42.5607/1.4081
- This data is from the Andorra extract but it shows the national park across the border correctly 
- The border labels currently do not display on the smaller polygon because it is part of a multi polygon with a larger area, and only the largest polygon is labeled by default
- `text-largest-bbox-only: false` causes the labels to render along the border of the smaller polygon as well (at high enough zoom level so that it is >196,000 waypixels)

z15 Current rendering
z15-border-park-master
z16 current
z16-master-border-park

z15 After
andorra-spain-park-z15-after
z16 After
z16-spain-andorra-border-park-after

**Boundary=administrative multi-polygon: Oecusse, East Timor**
https://www.openstreetmap.org/#map=16/-9.1772/124.4786
- This small enclave of East Timor is surrounded by Indonesia
- The border text labels do not currently render
Current z16
z16-timor-border-master
After z16
z16-oecusse-timor-border-after-16 -9 1772 124 4786

z10 unchanged (to show shape of enclave):
z10-oecusee-after-10 -9 2255 124 6056

### DIFF
--- a/admin.mss
+++ b/admin.mss
@@ -210,7 +210,11 @@ overlapping borders correctly.
   text-fill: @admin-boundaries;
   text-halo-radius: @standard-halo-radius;
   text-halo-fill: @standard-halo-fill;
+  text-largest-bbox-only: false;
   text-placement: line;
+  text-spacing: 750;
+  text-repeat-distance: 250;
+  text-margin: 10;
   text-clip: true;
   text-vertical-alignment: middle;
   text-dy: -10;
@@ -222,7 +226,11 @@ overlapping borders correctly.
   text-fill: green;
   text-halo-radius: @standard-halo-radius;
   text-halo-fill: @standard-halo-fill;
+  text-largest-bbox-only: false;
   text-placement: line;
+  text-spacing: 750;
+  text-repeat-distance: 250;
+  text-margin: 10;
   text-clip: true;
   text-vertical-alignment: middle;
   text-dy: -10;


jeisenbe:
This PR adds 4 additional properties to the admin-text and nature-reserve-text layers. I attempted to solve the 3 issues with fewer changes, but this was not possible.

**1) Text-margin:** 
Here's what it looks like in a few places without `text-margin`. `text-margin: 10` looks best and seems to work well. 
- Without text-margin:
madang-momase-no-margin
corner-no-text-margin

Current PR with text-margin:
with-text-margin-2
corner-with-margin

**2) Text-repeat:**
- Due to the mapnik bugs discussed above, the labels will sometimes render very close together, no matter what `text-spacing` is picked. `Text-margin` should not be very large, because it will cause collisions with other text and icons that prevent labels rendering. However, `text-repeat-distance` can be used to prevent the same text from being rendered twice in less than 250 pixels:

Without text-repeat-distance:
z17-wamena-metatile-corner-no-text-repeat
With text-repeat-distance
z17-wamena-after

**3) Text-spacing:**
- This causes the text labels to be repeated after a reasonable amount of time, if the metatile is large enough.
Without text-spacing: (zoomed out to 50% size in browser):
z16-malang-morobe-border-master

With  `text-spacing: 750` and `text-margin: 10`:
ramu-hwy-madang-morobe-after-lowres

jeisenbe:
As suggested by @talaj in the related issue (https://github.com/gravitystorm/openstreetmap-carto/issues/3649#issuecomment-456124838), changing `text-clip:` from `true` to `false` has solved the problem of the labels rendering very close together when a border crosses a metatile twice.

It is now it is possible to reduce `text-repeat-distance` to 100 pixels. This is enough to prevent two labels from being rendered right next to each other when a border bends back on itself (eg along a meandering river), or when there are two different admin_level with the same name which share a border (eg San Francisco (municipality, admin_level 8) and San Francisco (county, admin_level 6).

I've pushed a new commit with these two changes. 

**New test renderings:**

z16 Hamburg, Current:
https://www.openstreetmap.org/#map=1616/53.5640/10.0115
z16-hamburg-nord-borders-master
z16 New
hamburg-nord-less-textmargin-less-text-repeat

Uhlenhorst, Hamburg, z16 Current:
z16-uhlenhorst-3x-border-master
z16 New
uhlenhorst-less-textmargin-less-text-repeat

Hoheluft-west, Hamburg, z16 Current:
z16-hoheluft-west-master
z16 New
hoheluft-west-less-textmargin-less-text-repeat

matthijsmelissen:
> changing text-clip: from true to false

I think this is likely to cause performance issues. Is anybody willing to test this? Even a rough quick test would be helpful.

jeisenbe:
Or if anyone knows how to test with Kosmtik on Docker, I could do it.

Is there a command that will return the rendering time for a metatile or
tile?

matthijsmelissen:
You could use [`render_list`](http://www.volkerschatz.com/net/osm/render_list.html) which I believe comes with Mapnik. You'd have to do a manual install though instead of using Docker.

Alternatively, you could simply count in Kosmtik how long it takes before a given tiles appears (make sure Kosmtik has loaded/parsed the style sheet completely first). This should be enough to at least catch very big performance degradations.

I'd expect the problem to be worst with long borders, like national borders, at high zoom levels.

jeisenbe:
OK, I've done some rough tests with a stopwatch. I've tried to press the "start" button at the same time that I press enter/return to request that kosmtik render a new zoom level, and I try to stop the timer as soon as all of the tiles are rendered on my screen (usually about 2 dozen). I also tried to find places near metatile borders, for the slowest possible rendering.

The 4 test areas were:
 
1) Poland borders only
- An extract of just the admin_level=2 and admin_level=4 borders of Poland, which I had previously downloaded for some other reason. These borders are quite long and have many nodes, and probably render slower than the Canada or USA border due to the complexity (?). There were no other features, except for a couple of peaks and passes along the border. 

2) The border of southern Wales with England, along a curving river at a small town. This is admin_level=4 and another admin_level=6 border too

3) Central Wales, now with a national park border, as well as the admin_level=4 border, in a village

4) Minor borders in central Luxembourg - where several different small administrative borders met in the outskirts of the main town

Only the first was significantly slower, likely due to the complexity of the Poland national border:
These were the rendering times in seconds, with metatile: 8 (each metatile was 64 tiles, and in most cases 2 or 4 metatiles were on the screen). I tested each code three times at each zoom level and location, switching back and forth between `true` and `false`. Probably the precision is only valid to the nearest 1 second.

**Text-clip: true**
 | z20 | z19 | z18 | z17 | z16 | z15 | z14
Poland mean | 5.2 | 1.6 | 3.7 | 4.2 | 2.4 | 2.8 | 5.4
Wales 1 mean | 18.4 | 16.3 | 11.3 | 6.2 | 6.2 | 7.8 | 5.9
Wales 2 mean | 5.6 | 4.3 | 8.5 | 21.6 | 6.2 | 6.5 |  
Luxembourg mean | 4.1 | 5.3 | 5.4 | 11.7 | 13.7 | 15.2 |  

**Text-clip: false**
 | z20 | z19 | z18 | z17 | z16 | z15 | z14
Poland mean | 16.8 | 9.6 | 11.0 | 6.3 | 7.5 | 1.8 | 4.3
Wales 1 mean | 10.6 | 5.4 | 16.8 | 10.9 | 6.3 | 6.5 | 5.1
Wales 2 mean | 10.1 | 10.5 | 6.7 | 10.1 | 11.4 | 3.2 |  
Luxembourg mean | 4.4 | 5.1 | 6.3 | 22.0 | 17.7 | 14.4 |  

These were the average ratios between the tests with `text-clip: true` and `text-clip: false`
(Note that these border text labels are only shown from z16; the z15 results for as a control):

  | z20 | z19 | z18 | z17 | z16 | z15 | z14 |   | Average ratio z19-16 | Standard deviation of this ratio
Poland mean ratio | 3.3 | 6.1 | 3.0 | 1.5 | 3.2 | 0.6 | 0.8 |   | 3.42 | 1.91 | (Poland borders)
Wales 1 mean ratio | 0.6 | 0.3 | 1.5 | 1.8 | 1.0 | 0.8 | 0.9 |   | 1.16 | 0.63 | (Wales 1)
Wales 2 mean ratio | 1.8 | 2.5 | 0.8 | 0.5 | 1.8 | 0.5 |   |   | 1.39 | 0.93 | (Wales 2)
Luxembourg ratio | 1.1 | 1.0 | 1.2 | 1.9 | 1.3 | 0.9 |   |   | 1.33 | 0.39 | (Luxembourg)

So the Poland borders extract was 2 to 12 seconds slower (a ratio of around 3, neglecting outliers), but the standard deviation is high: the ratio is probably around 3 but could be 1.5 to 6.0
The Wales and Luxembourg tests with full data were not significantly different; most of the time `text-clip: false` was a couple of seconds slower, but in 5 out of 15 zoom levels it appeared to be faster.

I'll attach the spreadsheet with the full data:
[border-text-clip-tests.zip](https://github.com/gravitystorm/openstreetmap-carto/files/2796435/border-text-clip-tests.zip)

kocio-pl:
Even without touching `text-clip` I like the changes from this PR in general. https://github.com/gravitystorm/openstreetmap-carto/pull/3652#issuecomment-455916141 helps to understand what these all options really do. I didn't make my own independent testing however to double check the output.

jeisenbe:
I've pushed a new commit that changes the nature-reserve-text label to use scale_denominator instead of pixel_width and pixel_height to calculate way_pixels. This allows exporting or rendering at higher resolutions without changing the zoom level where labels appear.

talaj:
I just recollected that I have an optimisation of text rendering along huge geometries that go far beyond rendering area. That could mitigate the slowdown caused by `clip=true`. It's a pity I haven't proposed the optimisation to upstream Mapnik yet. Also it's not much motivational that the time needed to get a commit to Mapnik deployed to OSM servers is enormous.

kocio-pl:
Good to hear that, that would be nice improvement. I wish we have better communication between different OSM-related projects by default, so we at least know what is more possible and what is hard to expect.

kocio-pl:
@jeisenbe I think it's better to revert clipping removal. I would like to test this code after that.

jeisenbe:
OK, I've returned to `text-clip: true` and changed back to `test-repeat-distance: 250` to prevent too many duplicate labels from showing with the clipped geometries. 

I also removed the way_pixel changes, since this will now be a separate PR.

Unfortunately this means that some border segments will continue to be unlabeled at certain zoom levels, so #3649 will remain open for now. 

Perhaps this can be fixed in a few months, next time that Mapnik is upgraded, if @talaj is able to implement improved performance with large geometries as mentioned above.

Issue #2052, opened 2016-02-20
Render ref= for highway=track and highway=path


msebast2:
US forest service roads and trails often have both a name and a Forest Service road or trail number.
It would be nice to have these numbers rendered.

BalooUriza:
Or, more often, only a ref.  Not worth sweating whether it's a hex shield
or a rectangular slat, or whether that slat's horizontal or vertical until
proper shield rendering happens, though...

matkoniecz:
Can you give examples of ways where such rendering would be useful?

See also #514 (ticket with the same new feature request, closed as it seemed as it would mostly support incorrect rendering)

msebast2:
Hi Mateusz,

We need to render the ref= for paths and tracks for the exact same reason we render ref= for freeways and highways. Sometimes signs use the ref= designator and sometimes they use the name. So a useful map needs to show both.

For example, if someone was out hiking in the Los Padres National Forest, they might come across this sign:(See attached image)

But today the hiker looks at open street map, and they don't see any trails labeled as 22W08.On open street map this trail is shown as "Horn Canyon Trail".http://www.openstreetmap.org/way/10736646
And we SHOULD show the name "Horn Canyon Trail, because at the other end of the trail is a sign with that name.



Thanks for the link to #514. I'm confused about why it was closed?If I make the path part of a relation will that cause the trail number to get rendered?Should I create a relation that contains only one path and no other ways?
Thanks & Regards,Michael Sebastian

BalooUriza:
Pretty much anytime there's a trail, track or tertiary in the national
forests of the US.  For example and from the best of my recollection (so
don't take it as gospel if you're looking for things to map right now),
let's consider Larch Mountain Trail in the Columbia Gorge National Scenic
Area.  If you start at Historic US 30 at Multnomah Lodge, the first
wayfinder you'll see will tell you to head up the lodge's promenade for
#400 Columbia Gorge Trail, #441 Larch Mountain Trail to Multnomah Falls
Overlook, or go back through the parking lot to take #400 Columbia Gorge
Trail, #420 Wahkeenah Trail.  So, continue along Larch Mountain Trail, and
the next wayfinder at the first intersection after the concrete promenade
ends and you're on a hiking trail, you'll just get "400" pointing east and
"441" pointing north (though maybe a wayfinder that says "To Overlook").
Go up past the overlook (about two thirds of a mile north northeast and a
[...]

PR #3654, opened 2019-01-22
Add rendering of "ref" for track roads

jeisenbe:
Related to #2052 

- Add rendering of "ref" codes and numbers for track roads, using a similar style as used for residential and unclassified roads

### Explanation
Track roads, tagged with highway=track, are used to access agricultural and forestry areas. Many of these small roads are unnamed, but in some areas it is common to use a reference number or string of letter and numbers to refer to such roads. For example, Forest Service roads in North America often have a ref, without a name, or sometimes both. 

Adding rendering of the ref will help with routing in these areas. The style of rendering is rather subtle, using the same style as the name text label rendering.

One issue is that the ref layer is rendered first, so if there is not space for both the name and the ref, the ref alone will be rendered.  But in practice, most track roads with refs are long, so this problem is much more common on highway=residential, which already has a ref rendering. If the ref is more commonly used than the name, this could be seen as a benefit. 

**Allotments in the city of Hamburg, Germany**
https://www.openstreetmap.org/#map=15/53.6185/9.9781
z15 Before
z15-hamburg-ref-master
z17 Before
z17-hamburg-allotments-master

z15 After
hamburg-ref-track-15 53 6185 9 9782
z17 After
z17-hamburg-allotments

**Well-mapped rural area in Maine, USA**
https://www.openstreetmap.org/#map=15/45.7108/-68.8079
z15 Before
z15-pole-line-master
z17 Before
pole-line-trail-master-z17

z15 After
z15-pole-line-trail-after-15 45 7108 -68 8079
z17 After
z17-pole-line-trail-17 45 69977 -68 79609-after
https://www.openstreetmap.org/#map=17/45.69977/-68.79609

**Old State Road, Maine**
https://www.openstreetmap.org/#map=18/45.80773/-68.90040
z18 Before
old-state-road-master
After
z18-old-state-road-after-18 45 80773 -68 90040

**Logan Pond Trail, Maine**
https://www.openstreetmap.org/#map=16/45.8244/-68.8615
z16 Current
z16-logan-pond-master
After
z16-logan-pond-trail-after-16 45 8244 -68 8615

### DIFF
--- a/project.mml
+++ b/project.mml
@@ -1885,12 +1885,12 @@ Layer:
                     osm_id,
                     way,
                     COALESCE(
-                      CASE WHEN highway IN ('motorway', 'trunk', 'primary', 'secondary', 'tertiary', 'unclassified', 'residential') THEN highway ELSE NULL END,
+                      CASE WHEN highway IN ('motorway', 'trunk', 'primary', 'secondary', 'tertiary', 'unclassified', 'residential', 'track') THEN highway ELSE NULL END,
                       CASE WHEN aeroway IN ('runway', 'taxiway') THEN aeroway ELSE NULL END
                     ) AS highway,
                     string_to_array(ref, ';') AS refs
                   FROM planet_osm_line
-                  WHERE (highway IN ('motorway', 'trunk', 'primary', 'secondary', 'tertiary', 'unclassified', 'residential') OR aeroway IN ('runway', 'taxiway'))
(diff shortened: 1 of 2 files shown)


Penegal:
Is this in production? If so, it seems to not work everywhere, for instance here: https://www.openstreetmap.org/#map=17/48.34782/6.13247

kocio-pl:
No, the deployment has not yet happened for some reason: https://github.com/openstreetmap/chef/issues/218

Penegal:
Ok, thanks and sorry for the noise.

PR #3657, opened 2019-01-27
Change way_pixels calculation to use scale_denominator

jeisenbe:
- Way_pixels will now be calculated from way_area divided by the square of scale_denominator and the standard pixel size

### Explanation:
- Currently, the value of the column `way_pixels` is calculated from the way_area (which represents the approximate area of a polygon in square meters, but by Mercator projection) by dividing by pixel_width * pixel_height.
- However, this causes the calculation to change when the resolution is changed, for example, when exporting at 2x resolution in Kosmtik, or when rendering "retina" resolution tiles. 
- Using the scale_denominator instead of the actual pixel height/width results in a consistent rendering at different rendering resolutions.

_Southern Australia_
9/-32.7988/137.0764
**Screenshot at standard resolution** - unchanged before and after PR
z8
z8-master-x1-lake-gilles
z9
z9-master-lake-gilles-1x

**Export at 2x resolution**
z8 2x Before
z8-lake-gilles-master-2x

z8 2x After
z8-after-2x-gilles

### DIFF
--- a/project.mml
+++ b/project.mml
@@ -94,7 +94,7 @@ Layer:
                                                     'meadow', 'vineyard', 'orchard') THEN landuse ELSE NULL END)) AS landuse,
               ('natural_' || (CASE WHEN "natural" IN ('wood', 'sand', 'scree', 'shingle', 'bare_rock', 'heath', 'grassland', 'scrub') THEN "natural" ELSE NULL END)) AS "natural",
               ('wetland_' || (CASE WHEN "natural" IN ('wetland', 'mud') THEN (CASE WHEN "natural" IN ('mud') THEN "natural" ELSE tags->'wetland' END) ELSE NULL END)) AS wetland,
-              way_area/NULLIF(!pixel_width!::real*!pixel_height!::real,0) AS way_pixels,
+              way_area/NULLIF(POW(!scale_denominator!*0.001*0.28,2),0) AS way_pixels,
               way_area
             FROM planet_osm_polygon
             WHERE (landuse IN ('forest', 'farmland', 'residential', 'commercial', 'retail', 'industrial', 'meadow', 'vineyard', 'orchard')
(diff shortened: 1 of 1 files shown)


jeisenbe:
In #3648 we discussed the best way to show this code. @imagico  had suggested basing it on the ppi (pixels per inch) of the standard resolution, which is approximately 90.71, hence:

`way_area/NULLIF(POW(!scale_denominator!*0.001*25.4/90.71, 2),0)`

I think the current code, `way_area/NULLIF(POW(!scale_denominator!*0.001*0.28*, 2),0)` is somewhat simpler. This is based on the 0.28mm standard pixel size that Mapnik assumes.

However, I'm happy to change this.

imagico:
Looks good but i have not tested it so far.

@hholzgra and @woodpeck might be able to comment if this works well practically.

imagico:
Tested and works fine, thanks.

Would still be good to get some feedback if this is useful in practical for-print rendering

matthijsmelissen:
Good to have this finally fixed!

jeisenbe:
Thanks! I use Kosmtik’s export function for printing, so it will help me,
at least.

I hope it will help with generating PDFs and SVG files for printing too.

(BTW, does anyone know a good way to create PDFs for printing?)

Adamant36:
Have you looked into [MyOSMatic](https://maposmatic.osm-baustelle.de/)? It creates printable maps and I think it supports PDF. If so, maybe you could email the person that created for ideas.

sommerluk:
As it seems it's difficult to use way_pixels on vector tiles, I've played a bit with the MSS code. It is possible to define in MSS this variable `@way_pixels: "([way_area]*pow(4,@zoom)/24505740000)";` and then use `@way_pixels` (instead of the SQL based `way_pixels`) later on. But it does not work well: While it works in Kosmtik for normal rendering, but yet for export is fails (completly, even when exporting at 100% scale). It seems that `@zoom` is not reliably available. `scale_denominator` isn't available in MSS, is it? I've also [asked here](https://github.com/mapnik/mapnik/issues/2553#issuecomment-536346213).

pnorman:
Way pixels has two uses - MSS filtering and SQL filtering. The SQL filtering is important for performance.

PR #3658, opened 2019-01-27
Reduce saturation of pitch outline color 

jeisenbe:
Related to #3651 and #3517 

- Change the pitch outline color to be the pitch color, darkened 20% and desaturated 10%
- Also changes track outline color in the same way.  (The track and track outline colors are the same as the pitch and pitch outline colors currently)

### Explanation:
- In #3651 and #3517, it's been suggested that the pitch outline is not harmonious with the pitch background color.
- The current pitch outline is 30% darker and 20% more saturated than the pitch color (`#aae0cb` or  Lch(85,22,168). This is a much higher chroma than the pitch background color, so the outline does not appear to be based on the same color. 
- Examining the outline colors for other features, such as residential, commercial and industrial land, shows than most are 15 darker and just slightly higher chroma (in Lch). So a more appropriate color for pitch and track outlines would be  Lch 70,28,168 - 15 points darker than @pitch, and slightly higher chroma:  `#74b9a0`
- Some testing showed that desaturate(darken(@pitch, 20%), 10%) results in an outline color very similar to `#74b9a0` - yes, this is still higher chroma than @pitch, even with desaturate 10%.

https://www.openstreetmap.org/#map=17/40.55731/-122.34847
Before
pitches-master-17 40 55732 -122 34847

After (Darken20%, desaturate 10%)
pitches-darken20-desat10-17 40 55731 -122 34847

Compare to earlier test with hard-coded `#74b9a0` for pitch outline:
pitches-master-17 40 55732 -122 34847

Other examples (after PR):
https://www.openstreetmap.org/#map=16/40.5941/-122.3924
parkpitches-darken20-desat10-16 40 5941 -122 3924

https://www.openstreetmap.org/#map=16/40.5902/-122.4029.png
landcolor-darken20-desat10-16 40 5902 -122 4029

### DIFF
--- a/landcover.mss
+++ b/landcover.mss
@@ -63,7 +63,7 @@
 
 // --- Sports ---
 
-@pitch: #aae0cb; // also track
+@pitch: #aae0cb;           // Lch(85,22,168) also track
 @track: @pitch;
 @stadium: @leisure; // also sports_centre
 @golf_course: #b5e3b5;
@@ -684,7 +684,7 @@
     polygon-fill: @track;
     [zoom >= 15] {
       line-width: 0.5;
-      line-color: saturate(darken(@track, 30%), 20%);
+      line-color: desaturate(darken(@track, 20%), 10%);
     }
     [way_pixels >= 4]  { polygon-gamma: 0.75; }
     [way_pixels >= 64] { polygon-gamma: 0.3;  }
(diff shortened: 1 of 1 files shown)


imagico:
Thanks.

Note the convention is to list Lch color values only if the color was chosen in Lch color space.  If like here the color is chosen as an sRGB color this does not make sense.

jeisenbe:
Sorry, I originally picked the color using LCH, but then went back and
found a way to create the same color with `desaturate` and `lighten` but
forgot to remove that line

imagico:
That's not what i meant - i meant you added the Lch values in a comment in

https://github.com/gravitystorm/openstreetmap-carto/blob/bf13f67ed8a2f33d7841e3e22abda12a0cb71d68/landcover.mss#L66

despite the color being not changed and not originally chosen in Lch.  No big deal.

PR #3578, opened 2018-12-19
Add icons for sports pitches

Adamant36:
This PR adds icons for sports pitches. Its related to #844 but does not resolve it. Since it only involves pitches. Other sports icons besides ones attached to pitches involve other things that are more complicated. So they will be dealt with in another PR. Along with sports centers. 

One caveat about this PR is that names don't seem to render properly. As it is, they sometimes render at above the rendering level of the icons or sometimes not at all, but never with the icons.I don't know why there is the inconsistency. It might be something on my end with the code or it could be related to other issues with icons getting in the way of rendering names. As such, id appreciate it if someone could look through the code or test it to figure out what the problem might be. As I don't know enough about the subject myself. Although, I will look into it further and fix it if I can anyway. I figured it was still worth doing the PR in the meantime though. 

american_football (all test renderings are at z18)
american football z18
https://www.openstreetmap.org/#map=18/38.23502/-122.11689 (way)
archery
https://www.openstreetmap.org/#map=18/37.05016/-122.06383 (way)
archery z18 way
https://www.openstreetmap.org/#map=18/36.99722/-121.99118 (node)
archery z18 node
badminton
https://www.openstreetmap.org/#map=19/37.37623/-122.19847 (way)
badminton z18 way
https://www.openstreetmap.org/#map=18/36.99125/-122.06386 (node)
badminton z18 node
baseball
https://www.openstreetmap.org/#map=18/38.23317/-122.11794 (way)
baseball z18
basketball
https://www.openstreetmap.org/#map=19/37.80742/-122.28172 (way)
basketball z18
handball
https://www.openstreetmap.org/#map=18/37.42064/-122.11221 (way)
handball z18
running 
https://www.openstreetmap.org/#map=18/37.59517/-122.39120 (way)
running z18
Shooting 
https://www.openstreetmap.org/#map=18/37.58821/-122.32339 (way)
shooting z18
skateboarding
https://www.openstreetmap.org/#map=18/37.59583/-122.38964 (way)
skateboard z18
soccer 
https://www.openstreetmap.org/#map=18/38.23360/-122.12001 (way)
soccer z18
table tennis
https://www.openstreetmap.org/#map=18/37.71591/-122.45565 (way)
table tennis z18
tennis 
https://www.openstreetmap.org/#map=18/38.23590/-122.12113 (way)
tennis z18
volleyball
https://www.openstreetmap.org/#map=18/37.59496/-122.39029 (way)
volleyball z18

PR #3660, opened 2019-01-28
Add horse riding icon and Area

### LINKED ISSUE #2344: Add rendering for leisure=horse_riding
Could you add rendering for the [leisure=horse_riding](https://wiki.openstreetmap.org/wiki/Tag:leisure%3Dhorse_riding) tag?

Example: [323603669](https://www.openstreetmap.org/way/323603669).

Adamant36:
This PR adds rendering of an icon and the area for things tagged as leisure=horse_riding. It seems like a good addition to the map because horse riding areas are major parts of farms and other things. I left out rendering of an outline on areas because most of them have fences that can be mapped. So I don't think its necessary.
https://www.openstreetmap.org/#map=19/38.73239/-120.89836 (way)
horse riding way
 https://www.openstreetmap.org/#map=19/37.80047/-122.16968 (node)
horse riding node
Closes #2344


imagico:
I see no consistent use of leisure=horse_riding that would justify addition to the map.  The dominant use of leisure=horse_riding (probably about 2/3) seems to be for riding areas/tracks that should be (and are more frequently) tagged leisure=pitch/track + sport=equestrian.  The rest is partly combinations with building=* or landuse=farmyard.  Beyond that the tag competes with leisure=sports_centre + sport=equestrian with similar - if not higher - use numbers.

Adamant36:
>I see no consistent use of leisure=horse_riding that would justify addition to the map. The dominant use of leisure=horse_riding (probably about 2/3) seems to be for riding areas/tracks that should be (and are more frequently) tagged leisure=pitch/track + sport=equestrian. The rest is partly combinations with building=* or landuse=farmyard. 

Hmmm, its not that way where I looked in America. It seemed pretty consistent to me and none of them where tagged with tracks or should have been. How do you know they "should be (and are more frequently) tagged leisure=pitch/track + sport=equestrian" Anyway? I doubt you actually checked 2/3 of the 7,950 uses to see how they "should be" tagged. 

Even if some are miss-tagged as pitches though, its probably tagging for the renderer. Which is all the more reason to render this. So things will be tagged properly. 

> The rest is partly combinations with building=* or landuse=farmyard. 

According to the wiki "The tag leisure=horse_riding is used to tag an equestrian facility, often called equestrian centre, riding centre or riding stables, where people practise horse riding, typically in their spare time. It may include riding arenas, stables, riding halls, other buildings, paddocks and more (compare also Riding)." 

It sounds like that includes buildings and farmyards to me. Some things take place both inside and outside of buildings. Along with in farmyards. I'm not sure what's controversial about that. The farmyard tagging is probably the same reason as a above and again is fixed by this. 

>  Beyond that the tag competes with leisure=sports_centre + sport=equestrian with similar - if not higher - use numbers.

Actually, they don't "compete." They are different tags that serve different purposes. Not everything that involving a horse is a sport anymore then every road is a race track because both involve cars.

There was actually a conversation about the differences in either the original issue or the sports rendering issue that you clearly weren't involved in. its not like we wouldn't have considered and discussed it. 

That's where it helps to be involved in the original discusses, or if not, to at least review the issues the PRs are attached to before commenting on them. If you don't want to take the time to do that though, @Hufkratzer can probably tell you what the differences are between leisure riding and sports riding.

Hufkratzer:
> There was actually a conversation about the differences in either the original issue or the sports rendering issue [...]

It was in https://github.com/gravitystorm/openstreetmap-carto/issues/2344#issuecomment-246817877 ff

imagico:
I am aware of the discussion that happened in #2344.  But i see no indication that leisure=horse_riding is used for structures where a different kind of riding is practiced than in those tagged leisure=sports_centre + sport=equestrian.  Nothing like that is mentioned on the wiki either.  https://wiki.openstreetmap.org/wiki/Tag:leisure%3Dsports_centre specifically mentions leisure=horse_riding as alternative tagging without stating a difference in meaning.

Without intention to start a tagging discussion - the very idea to distinguish between non-sport leisure riding and sport leisure riding is nothing that is likely going to work in OSM on a global level since definitions and understanding of sports vary too much between different cultures.

Hufkratzer:
And why do you think that, if the majority of horse riders is less sport-oriented, we should tag all riding stables as sport_centres and not as leisure=horse_riding? Compare http://en.wikipedia.org/wiki/Pleasure_riding#Pleasure_riding mentioned in https://github.com/gravitystorm/openstreetmap-carto/issues/2344#issuecomment-248544339. Why can we have leisure=golf_course and leisure=bowling_alley but not leisure=horse_riding?

imagico:
I don't care what tag is used here, i just look at if it is used consistently and with a well defined and documented meaning.  The rest was just slightly off-topic advise on what tags are likely to work in OSM.  If you disagree - no problem.

matthijsmelissen:
I second @imagico's observations. The tag leisure=horse_riding seems to be mainly used in instances where leisure=pitch + sport=equestrian can also be used. And as the latter tag is far more common, I don't think we should support the leisure=horse_riding tagging scheme. I'll therefore reject this PR.

Adamant36:
Hhmm, that doesn't seem to be what me or 
Hufkratzer found and theirs clearly difference between them. With the leisure aspect of horse riding being being different then a sport. Can you provide some examples on the map where things are mistagged? (its not like miltitags for ambigious things aren't supported anyway sometimes though, but whatever). 

Also, if that is actually the case, can I do a PR for sport=equestarian instead or are you both going to do the same thing with it?Because at least one of them should be rendered.

jeisenbe:
I’d recommend adding sport=equestrian icons to pitches.

Hufkratzer:
@matthijsmelissen 
I have seen some of this kind of mistagging, but I do not agree that on average leisure=horse_riding is "**mainly** used in instances where leisure=pitch + sport=equestrian can also be used". I could provide lots of examples where it was used like intended. But I assume that this is not enough to reopen this PR.

I have some ideas how the observed mistaggings can be reduced in the future. I think the main reason what caused it was the unspecific labeling in the editors as "horse riding facility". Is it possible to reopen this PR when this kind of mistagging will have been reduced? leisure=horse_riding can be used for all types of horse riding, leisure oriented (**without** sport=equestrian), and sport-oriented (together **with** sport=equestrian), leisure=sports_centre can only be used together **with** sport=equestrian and is always sport-oriented!

@Adamant36 
What do you mean with "PR for sport=equestarian instead"?

Adamant36:
> I’d recommend adding sport=equestrian icons to pitches.

Yeah but then the same argument for not rendering it could be made in reverse. "seems like a lot of these are miss-tagged as sport=equestrian when they should be leisure=horse_riding so I'm closing this" etc etc. 

Which is why I think its a bad metric to use in the first place. Also rendering it, without also rendering leisure=horse_riding, would encourage  re-tagging things from leiusre=horse_riding to sport=equestrian for the renderer. Which we aren't suppose to do.

Hufkratzer:
@Adamant36 
I think it is unlikely that a lot of riding centres will be mistagged as pitches if you add the equestrian icon to pitches, If you use a pitch for the whole riding centre you can't see the pitches (riding arenas) inside of the centre anymore. You seem not to like riding arenas being tagged with leisure=pitch + sport=equestrian (noticable in https://github.com/gravitystorm/openstreetmap-carto/pull/3578#issuecomment-450452651). Why? It is used 23k times and documented in: 
* https://wiki.openstreetmap.org/wiki/Tag:leisure=pitch#Riding_arena and
* https://wiki.openstreetmap.org/wiki/Riding

I have removed the misleading recommendation on the wiki page for sports_centre to tag an outdoor riding arena with either sports_centre or horse_riding (see [diff](https://wiki.openstreetmap.org/w/index.php?title=Tag%3Aleisure%3Dsports_centre&type=revision&diff=1790054&oldid=1774697))

Adamant36:
> I think it is unlikely that a lot of riding centres will be mistagged as pitches if you add the equestrian icon to pitches. 

I didn't say they would. I said places tagged with leisure=horse_riding would be. Which is separate from riding centers per say.  As it is, there names don't even render. For example, see the places in my second rendering example in the first message of this PR. So why wouldn't they be miss-tagged if sport=equestrian is favored? It happens all the time with other tags. Those places aren't necessarily "sports" related either. 

(As a side note, even if icons or area fill rendering is a no go, I think there should be a basic principle that place names should be rendered even in cases like this. Its the basic purpose of a map to show people where and what things are. And its secondary to secondary to arguments about the differences between "leisure" and "sport" tagging or whatever).

> You seem not to like riding arenas being tagged with leisure=pitch + sport=equestrian

Personally, I'm neutral on it. I just think there's a difference between a place where you leisurely ride horses versus doing it for competition (or sport). The wiki makes the distinction also. The original issue and this PR wasn't about the sporting aspects anyway and making it about sport instead of the instances where its a leisure activity just convoluted things and was partly what lead to this being closed. There's zero reason they can't or shouldn't be considered separately. That doesn't mean I have something personal against the other tag though. 

Ultimately you know more about this then I do. I have owned a horse before and ridden them, but it was always for recreation and never for sport. Ultimately I think the sport tag is vague and probably the one that's miss-used. As in my opinion it should only apply to horse racing facilities or ones where there are competitions. Especially in cases where its combined with pitches. Everything else should be tagged as leisure=horse_riding. For instances, [checkout this example](https://www.openstreetmap.org/#map=18/40.62636/-122.24073). They might be considered pitches but the people aren't using them for sport purposes, and re-tagging them as such "just because" would be completely wrong. There's a bunch of places like them all over California. Id like to see them rendered without people having to wrongly re-tagged them for it to happen.

Hufkratzer:
@Adamant36 
If you want to tag riding arenas that are not mainly used for sport with leisure=horse_riding, with what do you want to tag the riding centres of which these riding areas are a part of? Probably not also with leisure=horse_riding. It would be possible to create a new tag leisure=riding_arena for all kinds of riding arenas, may they be used for sport or not (like your example). This would theoretically make some sense, but I took the established way of tagging riding arenas with with leisure=pitch + sport=equestrian for granted when I started to deal with OSM, like I also take for granted the established way of tagging tennis courts as pitches although they are not really pitches.

Adamant36:
> If you want to tag riding arenas that are not mainly used for sport with leisure=horse_riding, with what do you want to tag the riding centres of which these riding areas are a part of?

leisure=pitch + sport=equestrian obviously. I never said otherwise. This PR just had nothing to do with it. So I never brought it up. I'm not sure why there's any confusion about it. Like I said and you agree with, the places I provided examples for are tagged as leisure=horse_riding. So I'm not sure what the confusion is about or why leisure=pitch + sport=equestrian is even an issue in this case. As it doesn't apply to these places or the PR. 

> It would be possible to create a new tag leisure=riding_arena for all kinds of riding arenas

Its completely unnecessary to create a separate leisure tag just for none sports riding arena's since leisure=horse_riding covers them (I wouldn't consider them arena's anyway. More like horse pens). Plus, I'm not going to have a tagging discussion. Its completely pointless and has zero relevance. If your not clear about the definition of something or think a tag should be changed/created take it up in the wiki. Personally, I have more important things to do like figure out what will and won't get rendered. I don't feel like getting side tracked with a tagging discussion as part of that and its not the place for it anyway.

Hufkratzer:
@Adamant36 
> I'm not going to have a tagging discussion. Its completely pointless and has zero relevance.

It has relevance because it was the reason for closing this PR. If we don't know what caused the problem it's difficult to solve it. You seem to advocate a usage of leisure=horse_riding that is rejected by the maintainers. So it would be intersting, to find out what can be changed to please you and the maintainers at the same time. Ideas?

> Its completely unnecessary to create a separate leisure tag just for none sports riding arena's since leisure=horse_riding covers them (I wouldn't consider them arena's anyway. More like horse pens).

IMHO that is not what the wiki says, As you cited [above](https://github.com/gravitystorm/openstreetmap-carto/pull/3660#issuecomment-458274130): 
> "The tag leisure=horse_riding is used to tag an equestrian facility, often called equestrian centre, riding centre or riding stables, where people practise horse riding, typically in their spare time. It may include riding arenas, stables, riding halls, other buildings, paddocks and more (compare also Riding)."

As I understand it riding pens and riding arenas can be inside of leisure=horse_riding areas but not be leisure=horse_riding areas themselves. I think this was also the outcome of the initial discussion in https://github.com/gravitystorm/openstreetmap-carto/issues/2344#issuecomment-247044297 ff. We also have different tags for the whole golf_course and for all the suff that can be inside of it.
A single horse pen with some horses in it and nothing else around belonging to it, where people go, pick a horse and ride in the surroundings, could perhaps be tagged with leisure=horse_riding. But does this really exist? Your rendering examples [here](https://github.com/gravitystorm/openstreetmap-carto/issues/2344#issuecomment-457900212) don't look like this in the aereal view, they look more like riding arenas / riding pens; standard size 20 x 40 m for small ones (see [OSM wiki](https://wiki.openstreetmap.org/wiki/Tag:leisure=pitch#Riding_arena),  [wikipedia](https://en.wikipedia.org/wiki/Dressage#Arena)). Wikipedia doesn't make a difference between riding arenas and riding pens, see [here](https://en.wikipedia.org/wiki/Equestrian_facility#Arena).

>> If you want to tag riding arenas that are not mainly used for sport with leisure=horse_riding, with what do you want to tag the riding centres of which these riding areas are a part of?

> leisure=pitch + sport=equestrian obviously. I never said otherwise.

I am surprised. Do you really suggest to tag riding centres as pitches? Where is this documented? On the [wiki page for leisure=horse_riding](https://wiki.openstreetmap.org/wiki/Tag:leisure=horse_riding) is [this example](https://www.openstreetmap.org/way/485444067) which has the pitches inside of the area tagged with leisure=horse_ridng, so it's the opposite of what you propose.

>> I think it is unlikely that a lot of riding centres will be mistagged as pitches if you add the equestrian icon to pitches.

> I didn't say they would. I said places tagged with leisure=horse_riding would be. Which is separate from riding centers per say. As it is, there names don't even render. For example, see the places in my second rendering example in the first message of this PR. So why wouldn't they be miss-tagged if sport=equestrian is favored? [...]

Ok, these examples are nodes, I forgot that nodes can also be used for pitches. So there is a chance that some places currently have a leisure=horse_riding node will be retagged to leisure=pitch + sport=equestrian if only the last one will get an icon.

BTW: [Wikipedia](https://en.wikipedia.org/wiki/Equestrian_facility#Arena) says (contradicting what you said):
>A riding academy or **riding center** is a school for instruction in equestrianism, or for hiring of horses for **pleasure riding**.

jeisenbe:
> there is a chance that some places currently have a leisure=horse_riding
node will be retagged to leisure=pitch + sport=equestrian if only the last
one will get an icon.

Because of this risk, it is not a good idea for us to render sport=* icons
on nodes, but only on closed ways / polygons.

Luzandro:
I would like to revisit the topic, as the decision is not really understandable from todays perspective. @imagico  states that it is not used consistently and stands alternatively for leisure=pitch/track + sport=equestrian or leisure=sports_centre + sport=equestrian, among others. However, the wiki currently states quite clearly, both on the pages for the individual tags as well as in the [general article on riding](https://wiki.openstreetmap.org/wiki/Riding), where and how leisure=horse_riding should be used and when to use leisure=sport_centre. In fact, sport_centre is only mentioned in combination with sport=horse_racing but not sport=equestrian, which is also reflected in the usage (and the presets of the editors iD and JOSM). There is a discussion on the difference in the [talk section on leisure=horse_riding](https://wiki.openstreetmap.org/wiki/Talk:Tag:leisure%3Dhorse_riding#When_to_use_leisure.3Dhorse_riding_instead_of_leisure.3Dsports_centre.3F) and the usage comparison given there has shifted even further towards leisure=horse_riding with 13k by now, while sport_centre+equestrian stagnates at 2.6k even though only the last one is rendered on osm.org

PR #3574, opened 2018-12-15
Nature reserve boundaries revision

vholten:
Related to #3538, #2978

Changes proposed in this pull request:
- Make nature reserve boundaries less prominent at z8 and z9
- On all zoom levels: Reduce the "glowing" effect that occurs on boundaries that are shared between two adjacent nature reserves

Test renderings of [Northern California](https://www.openstreetmap.org/#map=8/41.263/-122.574)

z8 before:
z8osm

z8 after:
z8tuned

z9 before:
z9osm

z9 after:
z9tuned

z10 before:
z10osm

z10 after:
z10tuned

PR #3661, opened 2019-01-30
Reduce way_pixels limit to >750 for protected_area boundaries

jeisenbe:
- Renders national_park, nature_reserve, protected area, and aboriginal_lands boundaries at >750 waypixels instead of >3000.
- This causes the boundary outline to render 1 zoom level sooner than the name label for these areas. The text label renders at >3000 waypixels.
- This is also the same limit that is used for tourism areas, such as zoos and amusement parks, with outlines. (Military areas and police stations currently render with > 900 waypixels, similar to the number in this PR.)

### Explanation
- Currently, boundaries for protected areas, national parks, nature reserves and aboriginal_lands are only shown when the polygon is rather large: 3000 square pixels, approximately. For example, a 31 by 101 pixel rectangle will render, but a 30 by 90 pixel polygon would not show the borders at that zoom level.
- 3000 square pixels is also the same limit used to show the text label, because it should usually be large enough the fit the name within the area (though this depends on the shape of the area and the length of the name). This means that the area is not visible until it suddenly appears at a moderately large size, with a name. 
- Previously, tourism areas (zoos and theme parks) were changed to render at 750 way_pixels; this is 1/4 of 3000, so it is 1 zoom level sooner. For example, a 25 x 30 pixel polygon is 750 pixels square. This means the outline normally appears one zoom level before the name (if there is a name). See #2835
- It's logical to render this boundary lines with the same rules, so that the area will be shown one zoom level sooner than the name. 

**Lake Gilles, South Australia**
https://www.openstreetmap.org/#map=9/-33.0317/136.9267
Before: z8
z8-gilles-master
z9
z9-gilles-master-s33 0294-e136 9240

After: z8
z8-gilles-750px
z9
z9-lake-gilles-750px

**Singapore**
https://www.openstreetmap.org/#map=13/1.3804/103.8138
z12 Before
z12-singapore-protected-master
z12 After
z12-singapore-protected-after

Sungei Buloh
z13 Before
z13-sugei-buloh-master
After: z13
z13-sungei-buloh-after
z14 same
z14-sungei-buloh

Chestnut Nature Park (left)
z13 Before
z13-central-water-master
After z13
z13-central-water-master
z14 same
z14-chestnut-nature-park

### DIFF
--- a/admin.mss
+++ b/admin.mss
@@ -241,7 +241,7 @@ overlapping borders correctly.
 }
 
 #protected-areas {
-  [way_pixels > 3000] {
+  [way_pixels > 750] {
     [zoom >= 8][zoom < 10] {
       ::fill {
         opacity: 0.05;


kocio-pl:
This would partly revert #2119, but that should not be a big problem:
- I like the idea to make it similar to other areas 
- 750 instead of 3000 is still bigger than 100
- we had also some tuning of protected areas borders lately in #3574, especially on z8 and z9, where the problem is the most severe

kocio-pl:
Seattle example:

Old rendering from https://github.com/gravitystorm/openstreetmap-carto/pull/2119#issuecomment-219212784

2119_master

Current rendering

screenshot_2019-02-05 openstreetmap carto kosmtik

Proposed rendering - the missing pieces inside are now back, while borders are no longer screaming

screenshot_2019-02-05 openstreetmap carto kosmtik 1

kocio-pl:
Warsaw example:

Old from https://github.com/gravitystorm/openstreetmap-carto/pull/2119#issuecomment-350494433

aq2tiysm

Current

screenshot_2019-02-05 openstreetmap carto kosmtik 2

Proposed

screenshot_2019-02-05 openstreetmap carto kosmtik 3

kocio-pl:
> I hope this is a good compromise.

Both US examples are extreme and they still look sane now, so yes, I think so.

Thanks for making and explaining this unification.

jeisenbe:
> Both US examples are extreme...

It looks like the issue in those areas is that there are some moderately large protected_areas (wilderness or state forest?) which have many small outlying parcels, so it looks odd when these suddenly appear. If they were separate protected areas, they would not be render for another 2 or 3 zoom levels. 

Since the way_area is based on the whole multi-polygon, we can't easily render the outline for only the largest polygon without also rendering the tiny outlying polygons. 

For text labels, there's a property `text-min-path-length` which allows us to "Place labels only on polygons and lines with a bounding width longer than this value (in pixels)". Also there's `text-largest-bbox-only` which by default only labels the largest polygon in a multi polygon with a text label. 

We could ask the Mapnik developers if they could consider adding a similar feature for polygon fills and outlines?

kocio-pl:
Sure, please do.

PR #3663, opened 2019-01-31
Add rendering of "ref" on paths and create new layer for "ref" of minor roads and paths

jeisenbe:
Closes Issue #2052
Follow-up to PR #3654

1) Add rendering of "ref" numbers for paths (highway= bridleway, cycleway, footway, path, steps)

2)  Move aeroways, minor roads, tracks and paths "ref" to a later layer, after name labels are rendered

3) Minor changes:
- Change text-halo to standard size highway=unclassified/residential refs and aeroway refs
- Change residential / unclassified road "ref" size to 11 point at z17, to be the same size as the name text labels at this zoom level and higher. 

### Explanation

**1) Map users have requested rendering for the "ref" (reference numbers) for paths**
- Some paths are unnamed, but have a ref
- Other have both a name and a ref number, which can be useful for following longer routes
- This is most common with long-distance footways and cycleways, but can sometimes be found for other paths as well

**2) The minor roads, aeroways, and tracks with text-based ref labels currently render in the same layer as the shields for major roads.** 
- This means that the ref text is rendered before the name label, and can block the name from rendering when spaces is limited.
- Major road shields render much sooner than the ref text labels of minor roads and paths, which are not shown until z15 or z16.
- Major road shields are usually more important than the name; the number of a motorway is used more often, but for minor roads, such as residential streets, the name is more useful and should be shown with higher priority
- Changing residential, unclassified and paths to a later layer will fix this rendering. Aeroways "ref" text labels are also moved to this new layer because they are also rendered only at z15 and higher, though this does not change the cartography because they do not have a rendering for names

**3) The "ref" labels rendering for residential/unclassified roads needs minor adjustments**
- At z17, the "name" text label increases to 11pt font size, but the ref does not change size till z18 currently. This should be adjusted so that the "ref" label is the same size as the name label, as at z15 and z16.
- The halo needs to be changed to standard size. Right now it is 2 pixels wide, which would overlap with the casing at certain zoom levels. 

**highway=unclassified with name and ref** - Basilicata, Italy 
https://www.openstreetmap.org/#map=17/40.3776321/15.7281581
Before z17
via-chiusululle-unclassified-master
z18
z18-via-chiusululle-master
After z17
chiusululle-z17-after
z18
chiusululle-z18-after

**highway=residential with name and ref** - Basilicata, Italy 
https://www.openstreetmap.org/#map=17/40.1412792/16.0893142
Before z17
via-vittorio-res-z17-master
z18
via-vitorio-z18-master
After z17
z17-via-vittorio-after
z18
via-vittori-z18-after

**highway=path with name and ref** - Basilicata, Italy 
https://www.openstreetmap.org/#map=16/40.4133/15.6249
Before z16
z16-sentiero-master
After z16
z16-sentiero-refs-after-16-n40 4133-e15 624916

**highway=path with name and ref** - Montenegro
https://www.openstreetmap.org/#map=1616/42.6710/19.6521
Before z16
ref355-master
After z16
path-ref355-after-16 42 6710 19 6521

**highway=footway with ref** - Brussels, Belgium
https://www.openstreetmap.org/#map=19/50.85126/4.33989
Before z19
r20-master
After z19
r20-footway-ref-after-19 50 85126 4 33989

**highway=cycleway with ref** - Brussels, Belgium
https://www.openstreetmap.org/#map=17/50.87994/4.29554
Before z17
n9f-master
After z17
n9f-cycleway-ref-after-17 50 87994 4 29554

**highway=cycleway with ref** - Brussels, Belgium
https://www.openstreetmap.org/#map=18/50.86575/4.36460
Before z18
n277-master
After z18
n277-cycleway-ref-after-18 50 86575 4 36460

**highway=bridleway with ref** - Canada
https://www.openstreetmap.org/#map=16/52.8965/-118.1050
Before z16
bridleway-refs-master
After z16
bridleways-path-ref-after-16 52 8965 -118 1050

**highway=steps with ref** (and path with ref) - Germany
https://www.openstreetmap.org/#map=19/47.45867/11.12463
Before z19
kalbersteig-master
After z19
z19-kalbersteig-steps-path-19 47 45867 11 12463

### DIFF
--- a/project.mml
+++ b/project.mml
@@ -1888,13 +1888,10 @@ Layer:
                 SELECT
                     osm_id,
                     way,
-                    COALESCE(
-                      CASE WHEN highway IN ('motorway', 'trunk', 'primary', 'secondary', 'tertiary', 'unclassified', 'residential', 'track', 'bridleway,' 'cycleway,' 'path,' 'steps') THEN highway ELSE NULL END,
-                      CASE WHEN aeroway IN ('runway', 'taxiway') THEN aeroway ELSE NULL END
-                    ) AS highway,
+                    CASE WHEN highway IN ('motorway', 'trunk', 'primary', 'secondary', 'tertiary') THEN highway ELSE NULL END AS highway,
                     string_to_array(ref, ';') AS refs
                   FROM planet_osm_line
-                  WHERE (highway IN ('motorway', 'trunk', 'primary', 'secondary', 'tertiary', 'unclassified', 'residential', 'track', 'bridleway,' 'cycleway,' 'path,' 'steps') OR aeroway IN ('runway', 'taxiway'))
+                  WHERE highway IN ('motorway', 'trunk', 'primary', 'secondary', 'tertiary')
                     AND ref IS NOT NULL
               ) AS p) AS q
           WHERE height <= 4 AND width <= 11
@@ -1905,16 +1902,6 @@ Layer:
               WHEN highway = 'primary' THEN 36
               WHEN highway = 'secondary' THEN 35
               WHEN highway = 'tertiary' THEN 34
-              WHEN highway = 'unclassified' THEN 33
-              WHEN highway = 'residential' THEN 32
-              WHEN highway = 'track' THEN 30
-              WHEN highway = 'bridleway' THEN 10
-              WHEN highway = 'cycleway' THEN 10
-              WHEN highway = 'footway' THEN 10
-              WHEN highway = 'path' THEN 10
-              WHEN highway = 'steps' THEN 10
-              WHEN highway = 'runway' THEN 6
-              WHEN highway = 'taxiway' THEN 5
               ELSE NULL
             END DESC NULLS LAST,
             height DESC,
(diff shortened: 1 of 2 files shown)


jeisenbe:
The majority of paths with a ref also have a name, though there are some
unnamed paths that only have a ref, just as is the case with roads, based
on the search results that I found while looking for examples to render.

(Eg Hwy 96 is unnamed. I-5 sometimes has an official name but the ref is
almost always used, or in SoCal they call it “The Five”)

matthijsmelissen:
I'm still not very convinced about the need for this. In which countries would this be desirable?

I know many (hiking/cycling) routes have ref's, but are there also many instances when the paths/tracks themselves have refs?

I'm a bit concerned that this will lead to mistagging, where mappers will add the route ref to the individual roads rather than the route relations.

jeisenbe:
> I'm still not very convinced about the need for this. In which countries would this be desirable?

> I know many (hiking/cycling) routes have ref's, but are there also many instances when the paths/tracks themselves have refs?

Many forest trails in the USA have a ref, similar to those used for forest tracks and roads, especially on lands owned by the federal government (eg National Forests, National Parks, BLM lands). Not all of these are currently mapped. The trail has a reference number. Some may have started out as track roads, but now are trails because they are not maintained for 4-track motor vehicle access, or motor vehicle access is prohibited. 

For example, there are 934 ways (including >100,000 nodes!) in Oregon with highway=path and ref=*, usually a 4 digit numeric code that has been assigned by the US Forest Service. Most of these ways are 2 to 20 km long:

path-with-ref-oregon-usa
Search on Overpassturbo:
```
{{geocodeArea:Oregon}}->.searchArea;
  way["highway"="path"]["ref"](area.searchArea);
);
```

It appears some cycleways in the Netherlands have ref tags, eg "Fietssnelweg F35" around Hengelo. See https://www.openstreetmap.org/way/4797877 and connected ways for an example. 

> I'm a bit concerned that this will lead to mistagging, where mappers will add the route ref to the individual roads rather than the route relations.

I suppose this same issue is present with all types of roads? The ref is selected from `planet_osm_line` for all roads and paths, and if I understand how the data is imported correctly, this means it is from multipolygons and ways, but not from route relations

But I would think that the ref should be applied to the individual ways. The [wiki page for highway=](url
https://wiki.openstreetmap.org/wiki/Highways#Names_and_references) says:

"Names and references
"Use the name=* and/or the ref=* for each section of the road (way) which has a name or reference. If a stretch of highway has multiple reference numbers, they should be semicolon-delimited."

So all highways with a ref tag, including paths, footways, and cycleways, should have that ref=* tag on each of the ways, as is done with the name=* tag, according to this.

imagico:
The matter of route references has been discussed extensively on #596.

Note about half of the hiking and bicycle routes have a ref tag while only a small fraction of highway=footway/path/cycleway do.  So @matthijsmelissen's concern on this seems justified.

jeisenbe:
Thank you for the link to the other issue. I wasn't aware that the route relations were used so extensively in some other countries for walking and cycling routes. While walking and cycling route refs should not be rendered on most highways, it could be appropriate to render the ref found in a relation of type route=bicycle or =mtb on highway=cycleway, and the ref from a route=foot or =hiking would be appropriate for footways.

After reviewing issue #596 "Render Ref from Route Relation", it appears there is not yet a simple way to render the shields and ref numbers from the relations without affecting rendering performance. If I understand the state of the issue, the best solution would require improvements to osm2pgsql discussed at this link: https://github.com/openstreetmap/osm2pgsql/issues/230 - but no one is currently working on solving this. I don't have the technical ability to solve this issue.

So should we consider this to prevent the rendering of refs for footways and cycleways from the tags on the ways, since we cannot yet render refs from tags on route relations? I could remove these changes from the PR.

matkoniecz:
>     * This is most common with long-distance footways and cycleways

This should be tagged as route relations, not directly on `highway=*` elements, right?

jeisenbe:
According to [several](https://github.com/gravitystorm/openstreetmap-carto/pull/3663#issuecomment-465964091) [comments](https://github.com/gravitystorm/openstreetmap-carto/pull/3663#issuecomment-466939541) [above](https://github.com/gravitystorm/openstreetmap-carto/pull/3663#issuecomment-464249162), it is best if "ref=" tags are placed on route relations for footways and cycleways. This requires solving issue #596 before we can render the ref for paths, which could be done best by improving osm2pgsql. See openstreetmap/osm2pgsql#230

Therefore, I am closing this PR, and will open a new PR with the other changes.

Issue #3647, opened 2019-01-18
"The map appears washed out, more contrast wanted"


jeisenbe:
### Issue
**According to @kocio-pl, some map users have been complaining that the current rendering appears "washed out", and there "more contrast [is] wanted".** See is comment: https://github.com/gravitystorm/openstreetmap-carto/issues/3635#issuecomment-453684493
- "Washed out" suggests what happens to watercolor paints when too much water is added, or the color of a faded article of clothing after many launderings.
- This may refer to a reduction in contrast between different area fills or between some of the background colors, linear features and icons.
- Perhaps the complaint is that there are too many areas with high lightness values now?
- There may also be a complaint that more colors are now unsaturated (gray or grayish)
- **Note:** this issue is not my personal complaint. I did not notice this issue myself). Rather this is a place to discuss what has been said by others and decide if it is a problem or not.

### Possible causes
- A review of all the merged PRs in the past 3 years shows this list of pull requests which were related to the color of landcover, icons or linear features, and which may have reduced contrast and saturation or increased lightness.
- I have highlighted the PRs which are more likely to be related to this issue, in bold.

**#3625 Allotments color and pattern 2** - _changed from light green to less saturated, more blueish green_
**#3548 Allotments color and pattern 1** - _changed from orange-brown to light green_
**#3536 Landcover low zoom cleaning** - _removed buildings, some roads and paths_
**#3529 Change scrub color and pattern** - _slightly less blue and less saturated_
**#3493 Render religious landuse and place_of_worship lighter** - _lighter color, lower contrast_
#3475 Rendering orange dots for gastronomy on z17 - _smaller, less prominent icons_
#3470 Reduce noise of swimming pools at low zoom - _probably not related_
#3467 Clean up z13 - _removed buildings, some paths and minor roads from z13_
#3466 Render societal amenities like residential etc on mid zoom
**#3444 Change color of apron in aerodromes** - _changed apron color from light purple to light gray_
**#3412 Change societal amenities color**  - _changed color to very light yellow_
**#3327 Changing farmland and societal amenities colors** - _switched societal_amenities (lighter, less saturated) with farmland_
#3225 Brighten built-up areas on z12
#3096 Making gardens to use grass color with plant nursery pattern -  _probably not related?_
#3085 Change label colour of private parking -  _Minor change, but text is lighter_
**#3057 Made military area rendering less prominent**
[...]

PR #3458, opened 2018-10-17
Showing natural areas on z5+ [WIP]

kocio-pl:
Related to #2688
Implements parts of https://github.com/gravitystorm/openstreetmap-carto/blob/master/USECASES.md

Changes proposed in this pull request:
- Start rendering natural and semi-natural areas from z5 instead of z8+. Since we have usecases for some low zoom levels, it's time to show them there. z0-z4 might need separate attention, so it's good to make gradual progress and gather some experiences on z5-z7 first. This is also good opportunity for showing glacier and icesheets in a more balanced way (effectively reverting #2970).

Examples (click on images to see the full size):

**Poland**

z5

b61mojiq

z6

yj1jhpun

z7

yxdavqqx

**Egypt**

z5

vr4 7d1r

z6

t6dccn3i

z7

dn_8susl

PR #3670, opened 2019-02-04
Simplify low-zoom landcover rendering to 4 main colors (requires 3 water colors)

### LINKED ISSUE #3007: include spacial urban amenities in 'built-up-upper-lowzoom'
In mid and low zoom, currently the following landuses are rendered in gray (built-up-upper-lowzoom #c0c0c0, built-up-lower-lowzoom 11-12 #aaaaaa):
- landuse=residential
- landuse=retail, amenity=marketplace
- landuse=industrial
- landuse=commercial

Spacial amenities (i.e. campuses) and other landuses are often left without surrounding landuse tag, thus they leave holes in the midzoom areas that would be considered as built-up.
Examples (currently rendered from):
- landuse=garages (>=13) #2988
- landuse=religious (>=13)
- landuse=allotments (>=10)
- amenity=hospital/clinic/university/college/school/kindergarten/community_centre/social_facility (>=10 like residential, >=12 societal-yellow) #3004
- amenity=parking/bicycle/motorcycle parking (>=10)

To avoid the holes, these features should be assigned the build-up colour in the mid and zoom levels.

### LINKED ISSUE #3513: Missing Contrast in medium zoom levels
### Expected behavior
general map that gives indications about the place
### Actual behavior
road focused
Here some example comparisons between osm-carto and the geofabrik fork, look how the forests are important to understand the structure of the landscape and how they are missing in current osm-carto, until zoom 13 when they become visible but when the scale is so low that you cannot see the bigger picture any more:
Zoom 8
Zoom 9
Zoom 10
Zoom 11
Zoom 12
Zoom 13

I would say, color fading was an interesting experiment, but in the end it seems the results are worth with the fading than without.

### LINKED ISSUE #3936: Landcover and landuse features cannot be distinguished at mid and low-mid zoom levels
Related to #3935

### Expected behavior
- It should be possible to see when an area lacks landcover/landuse mapping, even at low-mid zoom levels
- Landcover distinctions that are rendered should be visually distinctive, even at mid zoom levels
- Landcover/landuse should be similar in visual prominence to water areas and waterways
- Landcover/landuse should be similar in prominence to roads and railways

### Actual behavior
- Landcover is strongly lightened by 0.4 (out of 1 in HSLA) at z12, and increasing to 0.5 at z11 and 0.6 at z10 and below
- Light landcover, like farmland, is not clearly visible at z10 and below because of the lightening
- Distinctions between similar landcover colors (like orchard/wood, grass/farmland, scrub/heath) are not clearly visible at z12, and are not visible at all at z10 and below
- Due to issues with how mapnik renders small polygons, the fading increases at z9 and below as smaller areas of landcover are no longer rendered
- Since the fading changes with each zoom level, it's not really possible to recognize each type of landcover without extensive research at z10 to z12 especially.
- Roads, railroads, and waterways are much more prominent than landcover at z12 and below, especially at z10 and below
- Water areas are not faded, and therefore increase in prominence relative to landcover 

See screenshots in #3935

jeisenbe:
Fixes #3513 "Missing contrast in mid-zoom levels"
Fixes #3007 "Include spacial urban amenities in 'built-up-upper-low-zoom'"
Fixes #3936 " Landcover and landuse cannot be distinguished at mid and low-mid zoom levels"
Related to #3935 "Roads are too prominent compared to landcover and rivers at mid and low-mid zoom levels"
Related to #3647 "The map appears washed out, more contrast wanted"

- Remove low-zoom fading
- Simplify low zoom (< z11) landcover rendering to use 4 main colors: darker green for tall vegetation, light green for low vegetation, light brown for bare ground, and gray for built-up (developed) land.
- Glacier, water, mud, and land-color will also be shown in the same colors as currently. Farmland and beach are still shown separately to z11, but combined with low-vegetation and bare_ground at z10, while most others are unified at z11. Built-up area is mainly unified at z12, as currently, except for airports.
- Also, this PR changes track/pitch/ice_rink lowzoom to be unified with builtup
- Police and Fire stations are also shown as built-up areas at low zoom, and the min zoom is changed to z10 which fits with the layer (landcover is limited to z10). Therefore the way_pixels-based rendering is removed.
- The airport transportation-area fill is shown from z11; apron is shown as the same color at this level. 

Wetland and sand patterns are still shown early, though this may need to be reconsidered.

These colors and zoom levels are very similar to those used in @imagico 's alternative-colors fork

One change that has NOT yet been made is to the 1 pixel limit for the size of landcover areas that are rendered. While reducing this limit would improve the rendering at low zoom levels, it also might affect performance, so this is one significant difference from the alternative-colors style. Further improving this should be considered in another PR. 

### List of landcover types in each of the 4 new low-zoom color categories:

**Vegetation-low:** `#ddecbd`
- picnic_site, caravan_site, camp_site,
- grave_yard, cemetery,
- park, recreation_ground, dog_park, golf_course, miniature_golf,
- allotments, farmland, greenhouse_horticulture
- meadow/grassland/grass, village_green, common, garden,
- heath, bog, string_bog, wet_meadow, fen, marsh

**Vegetation-tall:** `#b6dca6`
- vineyard, plant_nursery
- orchard
- wood, forest, swamp
- scrub

**Bare_ground:** `#efe7d9`
- quarry, landfill,
- bare_rock, scree, shingle, sand
- beach, shoal

(Sand and bare_rock still have patterns)

**Built-up:** `#d7d4d3`
- residential, retail, industrial, commercial,
- garages, parking
- farmyard
- railway, railway=station, 
- construction, brownfield
- power station/generator/substation
- hospital, clinic, university, college, school, kindergarten
- aeroway=apron, aerodrome,
- highway=services, rest_area
- fitness_center, sports_center, stadium, track, pitch, ice_rink

### Landcover and related area fill colors that are still rendered with separate colors
- Mud
- Glacier
- Water
- Land-color

## Explanation
In PR #2654 the landcover colors were faded from z12 to z10. At the time landcover was not shown at very low zoom levels. Later, landcover rendering was extended to as early as z5, however the fading was not changed ("Showing natural areas on z5+" PR #3458 )

Some other styles, such as the French fork, show landcover at full intensity even at lower zoom levels (though the French style uses a pre-processed z12 landcover rendering to speed up this process0

However, this leads to problems when a single pixel covers several different types of landcover, as often happens at low zoom levels. The final color may not be very intuitive.

@imagico suggested instead [combining many of the landcover colors into a simpler pallet](http://blog.imagico.de/on-basic-small-scale-landcover-rendering/). With only 4 main colors, it's clear that a darker green pixel is mainly tall vegetation, a light green pixel is low vegetation, mid-green is a pixel with a mix of vegetation types. Also, any shades toward gray are areas with developed land, and light orange/brown tones are bare land, such as rock and sand. 

### Further issues:
This PR is not a perfect solution to low-zoom rendering, because there is still a pixel size limit that prevents very small areas of landcover from rendering. However, this is a problem with the current rendering as well, especially at z8 and below. 

It could be possible to add a couple more types of landcover to the list of separate colors (currently water, glacier, and mud): for example farmland or sand and beach, which are currently rendered separately till z11 in this PR. I also tested another version with a mid-vegetation color for heath, scrub, and vineyards, about half-way between the tall and low vegetation colors. 

However, these complications would make the resulting blended colors slightly more difficult to interpret at low zoom levels, and the visible difference is quite small in most places. 

## Test renderings with links to the example places:

**Opolskie, Poland:**
https://www.openstreetmap.org/#map=10/50.506/18.545
z6 Before
z6-opolskie-master
z6 After
z6-opolskie-altcolors

z7 Before
z7-opolskie-master
After
z7-opolskie-altcolors

z8 Before
z8-opolskie-master
After
z8-opolskie-altcolors

z9 Before
z9-opolskie-master
After
z9-opole-alt-colors

z10 Before
z10-opole-master
After
z10-opole-altcolors

z11 Before
z11-opole-master
After
z11-opole-altcolors

z12 Before
z12-opole-master
After
z12-opole-altcolors

**Wales, Great Britain**
https://www.openstreetmap.org/#map=10/51.4901/-3.1603
z5 Before
z5-wales-master
After
z5-wales-altcolors

z6 Before
z6-wales-master
After
z6-wales-altcolors

z7 Before
z7-wales-master
After
z7-wales-altcolors

z8 Before
z8-wales-master
After
z8-wales-altcolors

z9 Before
z9-wales-master
After
z9-wales-altcolors

z10 Before
z10-wales-master
After
z10-wales-altcolors

z11 Before
z11-wales-master
After
z11-wales-altcolors

z12 Before
z12-wales-master
After
z12-wales-altcolors

### DIFF
--- a/style/water.mss
+++ b/style/water.mss
@@ -5,7 +5,7 @@
 @waterway-text-repeat-distance: 200;
 
 #water-areas {
-  [natural = 'glacier']::natural {
+  [natural = 'glacier'] {
     [zoom >= 5] {
       line-width: 1.0;
       line-color: @glacier-line;
@@ -17,10 +17,10 @@
     }
   }
 
-  [natural = 'water'][water != 'river'][water != 'canal'][waterway != 'riverbank']::natural,
+  [natural = 'water'][water != 'river'][water != 'canal'][waterway != 'riverbank'],
   [waterway = 'dock'][zoom >= 9],
   [landuse = 'basin'][zoom >= 7],
-  [landuse = 'reservoir']::landuse {
+  [landuse = 'reservoir'] {
     [int_intermittent = 'no'] {
       polygon-fill: @water-color;
     }
@@ -41,8 +41,8 @@
     }
   }
 
-  [natural = 'water'][water = 'river']::natural,
-  [natural = 'water'][water = 'canal']::natural,
+  [natural = 'water'][water = 'river'],
+  [natural = 'water'][water = 'canal'],
   [waterway = 'riverbank']::waterway {
     [int_intermittent = 'no'] {
       polygon-fill: @river-color;
@@ -54,7 +54,7 @@
       }
     }
     [int_intermittent = 'yes'] {
-      polygon-pattern-file: url('symbols/intermittent_water.png');
+      polygon-pattern-file: url('symbols/intermittent_river.png');
       [way_pixels >= 4] {
         polygon-pattern-gamma: 0.75;
       }


Adamant36:
Looks better. What's with the barely visible white checkered lines everywhere in the before and afters of Opolskie, Poland at z10? I take it those are track roads that have made to fade out to white starting at 12?that level? If so, they should probably be either not fading out anymore if we are getting rid of fading or just not rendered at all past z13.

jeisenbe:
**Luxembourg**
https://www.openstreetmap.org/#map=9/49.5895/6.0245
z8 Before
z8-luxembourg-master
After
z8-luxembourg-alt-colors

z9 Before
z9-luxembourg-master
After
z9-luxembourg-altcolors

z10 Before
z10-luxembourg-master
After
z10-luxembourg-altcolors

z11
_Current rendering but without fading_ (I have examples of this at all zoom levels as well)
z11-luxembourg-nofade
After
z11-luzembourg-alt-colors

z12 Before
z12-luxembourg-master
After
z12-luxembourg-alt-colors

jeisenbe:
**Shikoku island, Japan**
- Mainly forest
z5 Master
z5-shikoku-master
After
z5-shikoku-alt-colors

z6 Master
z6-shikoku-master
After
z6-shikoku-altcolors

z7 Master
z7-shikoku-master
After
z7-shikoku-altcolors

z8 Master
z8-shikoku-master
After
z8-shikoku-altcolors

z9 Master
z9-shikoku-master
After
z9-shikoku-altcolors

z10 Master
z10-shikoku-master
After
z10-shikoku-altcolors

z11 Master
z11-shikoku-master
After
z11-shikoku-altcolors

z12 Master
z12-shikoku-master
After
z12-shikoku-altcolors

jeisenbe:
> What's with the barely visible white checkered lines everywhere in the before and afters of Opolskie, Poland at z10

Good question! 

I checked the data, and it appears that the whole forest is divided up into small rectangular parcels about 300 x 1000m, or 400 x 400m in size, which share the same nodes at their borders, so it should render as one large area of forest. 

I'm not sure if these divisions represent land ownership parcels or if they are simply how the mappers in this area created the forest areas in small chunks:

data-inspector-opolskie

The borders between these parcels should not be visible, and I don't know why they are appearing at this zoom level only. There should be no change in the rendering with this PR; the polygon-gamma is the same before and after.

```
   [feature = 'landuse_forest'],
   [feature = 'natural_wood'] {
     [zoom >= 5] {
-      polygon-fill: @forest;
+      polygon-fill: @vegetation-tall-lowzoom;
+      [zoom >= 12] {
+        polygon-fill: @forest;
+      }
     }
   }
```

This issue also reminds us of the perils of filtering by way_pixels. Right now, some of the smaller forest parcels (300 x 200m) will no longer be rendered at z8, where they will be 0.8 pixels in size. 

If the limit was changed to 0.01 pixels, as on the alternative-colors branch, then an area of this size would continue to render until z5 at this latitude.

jeisenbe:
I've confirmed that the same artifacts are seen in the rendering from the master branch, and from the current branch with fading removed (which shows the lines more clearly):

Current rendering (Exported from Kosmtik) z10
opole-artifacts-nofade

Master branch, with fading removed z10
white-boundary-artifacts-opole

And here's a link to the same location on the live map for comparison: 
https://www.openstreetmap.org/#map=10/50.8554/17.9970

jeisenbe:
I've confirmed that the same artifacts are seen in the rendering from the master branch, and from the current branch with fading removed (which shows the lines more clearly):

Current rendering (Exported from Kosmtik) z10
opole-artifacts-nofade

Master branch, with fading removed z10
white-boundary-artifacts-opole

And here's a link to the same location on the live map for comparison: 
https://www.openstreetmap.org/#map=10/50.8554/17.9970

Screenshot from openstreetmap.org, zoomed in:
z10-opole-screenshot-artifacts

kocio-pl:
I have big general question (which I asked in the ticket issue probably): why clustering them, while the only problem for me seems to be dark forest/woods? There can be large areas of other things, so it's always good to recognize them. 

For me only making forest color lighter (maybe only on lower zoom levels) might be enough, but maybe also much less fading.

imagico:
> One change that has NOT yet been made is to the 1 pixel limit for the size of landcover areas that are rendered. While reducing this limit would improve the rendering at low zoom levels, it also might affect performance, so this is one significant difference from the alternative-colors style. Further improving this should be considered in another PR.

While this would certainly reduce the bias and the bad mapping incentive you have now (favoring big polygon mapping over small grained mapping) it would not create results similar to my low zoom demo.  This uses a completely different rendering approach for the landcover layers.

There are other design differences as well - in particular in roads rendering, various color differences and in not using any polygon pattern fills at low zoom levels (unification of base color is kind of pointless if you keep rendering a differentiated pattern above it).

jeisenbe:
>  it would not create results similar to my low zoom demo

Right, my understanding that the low-zoom demo isn't the same as the alternative-colors style, which uses Mapnik, correct?

(The pixel limit is only different in the landcover layer (z10 and up), not in the landcover-low-zoom layer, where it is already at 0.01 pixel in this style)

> not using any polygon pattern fills at low zoom levels

I think this is a good idea, but I was trying to limit this PR to changing the base colors, since it is already a very large PR to review. 

However, I could include these changes in this PR if the maintainers can reach a consensus to remove the sand and bare_rock patterns below z10 or z12. 

(I'm not sure if the wetland pattern is also considered a problem at low zoom levels; it looks like you have the pattern rendering at z9 for mud in the alternative-colors style)

> unification of base color is kind of pointless if you keep rendering a differentiated pattern above it

It's still useful for sub-pixel areas where the color mixing would otherwise be more confusing, though I agree that the pattern could be counterproductive.

imagico:
> Right, my understanding that the low-zoom demo isn't the same as the alternative-colors style, which uses Mapnik, correct?

Correct, the low zoom demo is an alternative low zoom continuation of the style.

> (The pixel limit is only different in the landcover layer (z10 and up), not in the landcover-low-zoom layer, where it is already at 0.01 pixel in this style)

Yes, this is just completely inconsistent here now - and the database indices are also not adjusted to the queries.  In terms of the results it does not really matter because the landcover rendering is purely decorative at these scales at the moment anyway.  But if you change that as you do in this PR it will start to matter again.

> (I'm not sure if the wetland pattern is also considered a problem at low zoom levels; it looks like you have the pattern rendering at z9 for mud in the alternative-colors style)

Starts at z10 for all wetlands:

https://github.com/imagico/osm-carto-alternative-colors/blob/master/landcover.mss#L799

imagico:
I should also mention that the low zoom color scheme used was designed for use with the three color scheme for water, in particular for the darker color for rivers.  It would therefore likely not work equally well with the single water color here.

jeisenbe:
> ... making forest color lighter (maybe only on lower zoom levels) might be enough, but maybe also much less fading.

Are there any places in the test renderings where useful information was lost?

We currently unify all of the developed landuses into one gray, built-up color at z12 and lower, because we've recognized that it would be difficult to see the different areas at lower zoom levels, and because small areas would lead to mixed colors when they overlap in one pixel. 

This same problem happens with the different vegetation colors and bare ground colors at low zoom. For example, if an area of `sand` overlaps with `heath`, it produces a very different colorthan `scree` with `grass`. By reducing the number of colors, the mixing is less confusing.

It's true that the dark forest color causes difficulty reading some linear features, but this isn't the only issue with using the same colors at high zoom. And - as I mentioned in https://github.com/gravitystorm/openstreetmap-carto/issues/3513#issuecomment-457117331 - we cannot use `#b6dca6` for wood/forest without merging these colors with orchard/vineyard, because these are almost the same color: `#aedfa3`

So if we use `#b6dca6` for forest, we will have already unified forest/wood, swamp, orchard and vineyard. (Scrub is the one other tall vegetation type that will be added by this PR.) And we have already unified the built-up landuse areas

If we've unified half of the types of landcover, it makes sense to simplify the others as well.

jeisenbe:
> the database indices are also not adjusted to the queries

I haven't yet studied how these work.

Would you have time to open an issue explaining the problem?

imagico:
Background info on the technical and cartographic aspects of way_area filtering can be found in:

#754, #1566, #1604, #2874.

When using a size threshold of 0.01 times the pixel size at z9 and 1 times the pixel size at z10 that is of course not something you can reasonably work with in terms of indices.

meased:
1. There is no contrast between rivers and forests (nearly invisible).
2. There is very little contrast between minor roads and forests.
3. There is very little contrast between minor roads and built up areas.

I don't think this helps with #3647. It *decreases* the contrast on some very important features.

jeisenbe:
> There is very little contrast between minor roads and forests.

While the contrast between forest and (gray) minor roads is less in this PR than in the current rendering at z11 and below, it is much better than with the current forest color without fading.

This PR intends to reduce the contrast between the background colors and the linear features such as roads, which are currently too prominent, especially compared to the rendering at z13 and above.

> There is very little contrast between minor roads and built up areas.

There is very little difference between the colors used for built-up areas in the current rendering and in this PR. However, the new color is actually a little lighter than the built-up area color in the current rendering at z10 and below, so the contrast with gray roads should be very slightly improved.

Here's a zoomed-in comparison from z10; current rendering on left and new rendering on right:

master-altcolors-z10-zoomed-in-built-up

> I don't think this helps with #3647 

This depends on what you understand by "washed out." I'm interpreting this complaint as talking about how the very light landcover looks next to the darker linear features and icons. While the contrast between near-white colors and darker lines is much higher,  I see this as a problem, especially since all of the colors were originally designed for the non-faded high zoom levels.

jeisenbe:
Oh, that's why we need to switch to ocean polygons first:
Sometimes the waterway=river continues out from the coastline, so the ocean needs to be above the water-lines layer, but right now the ocean is just the background color, underneath everything else:

z13-shikoku-river-color

Adamant36:
For what its worth, I'm 500% percent behind improving the water colors. I had actually tested a very similar thing to him a few months ago but scrapped it because of the problem with waterways showing over the ocean like in the last photo. If its not a problem though or there's a fix for it, I think changing the colors is a much needed change.

kocio-pl:
Unfortunately #2507 is stalled currently.

Well, it all depends what we'll define as a "bug" or as a "feature". If we want to give the best feedback, it's better to show both waterways and different water colors. But if we switch to ocean polygons, we can still give some feedback.

I opt for ocean polygons (which would be useful also for some other things), but I'm not sure about different water colors. Visually I don't think ocean needs lighter blue (I don't see it dominating the land, so it could be the same as for other standing waters), but rivers can be made stronger indeed. The joinings between rivers and lakes/oceans will look strange then (visually downside), but the upside would be to clearly show how arbitrary are all the water boundaries.

imagico:
Most past discussion on the water color in this style can be found in #1781.

Note the difference between inland water and ocean is already shown in this style currently - but only at z0 to z7 and in a way that is relatively confusing for mappers - both due to the styling and because it vanishes as you zoom in.

jeisenbe:
> the difference between inland water and ocean is already shown in this style currently - but only at z0 to z7

I believe this refers to the darker 0.5 pixel wide outline around the land polygons (i.e. on the coastline) at z0 to z7?

I'm marking this PR as WIP (work in progress) for now, because it looks like these 4 main colors will work if the river color can be changed, but this change depends on the switch to water polygons.

It might be possible to further lighten the forest/tall-vegetation color to work with the current river color, but the visibility of rivers is also an issue on forests at higher zoom levels, and the water polygons have other benefits, so I think we should work on that first.

jeisenbe:
Once we have ocean polygons, I will submit another pull request with the darker river color, and then this PR can move forward. If the new river color is accepted, these are examples of the results. I believe the rivers and water bodies are now visible over tall-vegetation at z8 to z12, even in areas with extensive forest:

z8 Wales, UK
z8-wales-4colors-fulldata

z9 Wales
z9-wales-4colors-generalized

z10 Shikoku, Japan
z10-shikoku-4colors-waterover

z11 Shikoku
z11-shikoku-4colors-waterover

z11 Opolskie, Poland - (please open to see at full size)
z11-stobrawski-4colors-waterover

z12 Opolskie
z12-stobrawski-4colors-waterover

z13 Opolskie after
z13-stobrawski-4colors-waterover
Compare  to current rendering:
z13-stobrawski-master

jeisenbe:
And here are some test renderings I made from central Libya, so we can see what an area of desert will look like. Unfortunately there are not very many arid areas with complete landcover mapping, but here you can see some circular areas of irrigated farmland and large areas of sand dunes at z8:

Current rendering at z9
z9-central-libya-master

z9 After
z9-central-libya-altcolors

z10 after
z10-central-libya-altcolors

z11 after
z11-central-libya-altcolors

z12 after
z12-central-libya-altcolors

jeisenbe:
I'm created a new branch that you can fetch for testing: `jeisenbe:landcover-water`
It contains the changes in this PR, as well as the water polygons and the different `water-color` and `river-color`, if anyone would like to test these changes early.
https://github.com/jeisenbe/openstreetmap-carto/tree/landcover-water

matthijsmelissen:
I like the idea of reducing the number of colours in the low zoom levels.

However, in the proposed images I think the colours compete too much with the roads. I think we should use less strong colours.

jeisenbe:
Continuing with this PR requires changing the river color, which will have additional benefits of making rivers more legible on darker backgrounds and providing better mapper feedback (e.g. when rivers are tagged as `natural=water` without `water=river` or `waterway=riverbank` this will show differently), however, it might be good to decide on #3854 first.

landcover.mss
```diff
@@ -68,19 +64,15 @@
 @stadium: @leisure; // also sports_centre
 @golf_course: #b5e3b5;
 
+// --- Low-zoom ---
+
+@built-up-lowzoom: #d7d4d3;
+@vegetation-tall-lowzoom: #b6dca6;
```
kocio-pl: I don't know where exactly to put this comment, but forest color is still a blocker for me, and it will be worse if used for more object types. The contrast between this and vegetation low zoom is too strong. This alone makes more problems than using more than 4 colors on low zoom.

This proves to be a problem for both small dark patches of green (like Wales and Luxembourg z8+ on provided examples) and solid dark green areas (Shikoku on z8+).

jeisenbe:
> Thanks for another example. Rivers look OK in every case, 4 landcover colors look bad for me in both cases (too strong high vegetation green).

> forest color is still a blocker for me, and it will be worse if used for more object types. The contrast between this and vegetation low zoom is too strong.

The `vegetation-high` color in this PR is slightly lighter than the forest color, BTW. Can you explain what problems are created by this color for woodland? Is it that the roads are less visible? Part of the goal of this PR is to balance the importance of roads vs other features at lower zoom levels. If you compare this PR to the openstreetmap.fr rendering at z11 or z12, the roads are still much more prominent in this style even after the PR. 

Or are there other features which you think are harder to see with the darker woodland color?

We could of course reduce the contrast between low and high vegetation, but this would continue to over-emphasize roads in this style in the lower zoom levels.

kocio-pl:
> The vegetation-high color in this PR is slightly lighter than the forest color, BTW

Thanks, it's good to know, but the problem is still huge.

> Can you explain what problems are created by this color for woodland? Is it that the roads are less visible? 

Nope, I didn't look at roads yet. 

My first impression is the main problem is a color saturation, the darkness might be second (or none).

jeisenbe:
> main problem is a color saturation

But what problem is caused by this?

Are some other features not legible? Is something not clear in the rendering, like it is unclear what features are in this color, or how they relate to other types of vegetation? Is it not intuitive that it's woodlands and other similar vegetation?

Subjectively, I think the color looks nice, but personal preferences about colors are strongly influenced by culture, and I can't claim that my cultural background and experience is representative of the global diversity of Openstreetmap.

kocio-pl:
> But what problem is caused by this?

In short - with patches, it makes the image to look like military camouflage, it's hard to focus on anything. With solid color, it "steals the show" for all the details located on this area.

jeisenbe:
Note that openstreetmap.fr uses quite dark colors for woodlands, even at z8:
z8-latvia-fr-style
z10-riga-fr-style

And the Humanitarian style uses dark woods without fading, though the saturation/chroma is lower (as with all colors in that style):
z8-latvia-hot-style
z10-riga-hot-style

imagico:
Like with the water colors i would like to point out that the choice of aggregated landcover colors is based on an overall different approach to color design at the low zoom levels, specifically not to use color fading.  Comparing the results with the faded low zoom landcover rendering right now is a bit non-productive IMO since it mixes two very different ideas in a single discussion:  The aggregation of landcover coloring into a smaller number of classes and the idea of fading colors using the technique currently used.

In other words:  In terms of discussing and finding consensus an agreement on moving away from fading landcover coloring at low zoom levels would probably need to come before it makes sense to do aggregation.

My experience in discussing low zoom landcover rendering with map users (in particular with feedback for my [low zoom rendering demo](http://maps.imagico.de/#map=6/49.500/5.800&lang=en&r=osmlz&o=aa&ui=8)) has essentially been that there are two groups of people:

* those who like and prefer the mostly decorative landcover rendering of many commercial maps at the lower zoom levels and who would like OSM maps to offer a similar look and feel.
* those who are excited about the possibilities offered by OSM data to go beyond the mainstream digital cartography and who value the mapper feedback at low zoom levels about landcover mapping.

What i also noticed is that the selective extension of specific landcover classes to lower zoom levels (sand, bare_rock and heath in particular) has lead to quite strong trends to specifically map these classes on larger areas with low accuracy:

https://www.openstreetmap.org/way/138017567
https://www.openstreetmap.org/relation/9350853
https://www.openstreetmap.org/way/597160562

Aggregate rendering of all landcover classes without specific preference from some of them would significantly help avoiding this kind of biased mapper feedback.

jeisenbe:
Also fixes #3935 " Landcover and landuse cannot be distinguished at mid and low-mid zoom levels"
Fixes #3007 "Include spacial urban amenities in 'built-up-upper-low-zoom'"
Related to #3935 "Roads are too prominent compared to landcover and rivers at mid and low-mid zoom levels"

jeisenbe:
Next step for this PR is to first merge PR #3952, reorder the water layers, and then discuss the river color again. After that we can discuss how dark woodlands should be at low zoom levels and the prominence of roads versus landcover.

project.mml
```diff
@@ -230,6 +230,7 @@ Layer:
             way,
             "natural",
             waterway,
+            water,
```
kocio-pl: (Since GitHub lacks general comments for reviews, I attach it to the first change, to make clear what problems that need to be fixed - sorry)

There are two problems I see here:
- dark wood spots (this is exactly the same problem as it was when testing color fading changes) - a blocker for me 
- I have a big question about what problem exactly resolves reducing colors at low zoom? The whole rationale is about fixing problems which will be probably fixed by #3952, it's not clear what this particular solution would change in comparison

jeisenbe:
Please explain the problem with the "high vegetation" color in the test images (used for scrub, woods and forest), considering that this color is actually lighter than the current `@forest` color used at z13 higher. 

Why is it a problem to show woodlands and scrub in this slightly lighter green at z11 or z10? 

Is it because roads, waterways, railways and borders are less prominent?

The reason for combining some of the landcover colors is the same as why we show all developed land as gray at z12 and lower: at these scales many types of vegetation and unvegetated areas are too small be be clearly visible, but if they are a large fraction of a pixel they will change the color of that pixel in somewhat unexpected ways. 

By simplifying to only 4 (or 5) types of landcover fill, it is easier to pick colors that will blend well together when a pixel is made up of 40% forest, 10% scrub, 10%heath, 20% farmland, 10% meadow, and 5% quarry and 5% sand, for example. 

I think it would also be reasonable to consider showing all the same landcover colors at low zoom in the same was as at high zoom, but since you have expressed concerns about the `@forest` color being still too dark even in this lighter version, it seems we need to use slightly different colors at low zoom. 

Currenly (and even with #3952) we have fading for every landcover fill color (but not water or linear features. 

This makes is harder to adjust the colors since we have to compare twice as many combinations, considering both the light versions at low zoom level and the darker versions at high zoom level. The current fading also makes it harder to use the map, impairing clarity and legibility, because map users have to learn twice as many landcover colors and need to realize that all the colors change between z10 and z13. 

While this will improve partially with #3952, if we are going to use a different `@forest` color at low zoom it makes sense to also use a smaller amount of different colors, focusing on the features which take up the most space

(BTW, to make a general comment as a review, you can got to "files changed" and click on "review changes" and just type a comment, approve or request changes there, without picking a particular line of code)

imagico:
For clarification:  This change depends on #3896 and both currently lack consensus but no one has so far suggested an alternative approach to addressing #3513, #3647,  #3935 and #3936.

jeisenbe:
To get to consensus it would be helpful to get an response to the questions above:

"Please explain the problem with the "high vegetation" color in the test images (used for scrub, woods and forest), considering that this color is actually lighter than the current @forest color used at z13 higher.

"Why is it a problem to show woodlands and scrub in this slightly lighter green at z11 or z10?"

"Is it because roads, waterways, railways and borders are less prominent?"

I would also appreciate thoughts and comments from other contributors about this PR or alternative ideas.

jeisenbe:
> This change depends on #3896 and both currently lack consensus but no one has so far suggested an alternative approach to addressing #3513, #3647, #3935 and #3936.

To be precise, the current PR is designed to work after the river color has been changed to be darker and more saturated, as suggested in #3896. In theory it would be possible to alter that suggestion:

1) Only use the darker river color at low zoom levels
or
2) Darken all water areas
or
3) Use the current water color, and accept that rivers will not be as visible in areas with many areas of woodland and scrubland. 

I don't think any of those ideas are better than the original plan.

To address #3513, #3647, #3935 and #3936, an alternative would be to completely remove the landcover fading at low zoom levels, instead of simplifying the number of rendered landcover colors as in this PR. However, this would probably not address @kocio-pl's dislike of the dark wood color at low zoom level.

pnorman:
> To be precise, the current PR is designed to work after the river color has been changed to be darker and more saturated, as suggested in #3896. In theory it would be possible to alter that suggestion:
> 
> 1. Only use the darker river color at low zoom levels
>    or
> 2. Darken all water areas
>    or
> 3. Use the current water color, and accept that rivers will not be as visible in areas with many areas of woodland and scrubland.
> 
> I don't think any of those ideas are better than the original plan.

I think decoupling this PR and #3896 would be good so we can get this unblocked.

jeisenbe:
> I think decoupling this PR and #3896 would be good so we can get this unblocked.

By which method? Just merging it without the water color changes, and accepting that waterways will be harder to see at low zoom level?

But kocio-pl was also unhappy with the color used for woodlands and other high vegetation areas (which is lighter than the current wood/forest/orchard colors, but darker than the faded landcover):

> "dark wood spots (this is exactly the same problem as it was when testing color fading changes) - a blocker for me"

I asked for clarification about this above (https://github.com/gravitystorm/openstreetmap-carto/pull/3670#issuecomment-555436174) but did not receive a response.

jeisenbe:
Rebased for v5.0.0, includes the 3 water colors.

Issue #661, opened 2014-06-24
Add rendering for golf=*


dieterdreist:
Please consider importing the "golf" key and render its values, they are documented here:
http://wiki.openstreetmap.org/wiki/Tag:leisure%3Dgolf_course
Particularly golf=
bunker (rendered as sand)
hole (the hole area)
pin (zoom 18/19?)
green
fairway
rough (rendered like natural=scrub)

geozeisig:
A long time has passed and nothing has happened. So everybody uses tags that are rendered. That may be landuse=grass, leisure=pitch, highway=path (with golf=hole) etc. That is **"Tagging for the renderer"** and it is provoked when golf=* has not been rendered.
It is right that there may be up to 4 tee boxes on a given hole. In most cases, however, only one (the furthest) is drawn. In any case, this should not prevent us from finding a solution.
How it made on osm.fr is quite a discussion basis.

nebulon42:
"Nothing has happened" is not quite correct. This depends on the rendering database schema change, which is covered in #2533. So there is already a PR for the prerequisite to this issue and it is close to being merged.

eFrysTon:
Given that all the prerequisites should be in place now, any chance something can be done about this in the near future?

I just tried mapping part of a golf course, and while this is how it looks in josm:

![image](https://user-images.githubusercontent.com/27037648/37750264-d07ed348-2dd7-11e8-8eab-2bd7d84d913b.png)

and this is how it's being rendered:

![image](https://user-images.githubusercontent.com/27037648/37750315-0bd27436-2dd8-11e8-9ec1-20d935037b96.png)

which is... somewhat disappointing.

As a minimum, I would be happy if areas (closed way or multipoly) with the following tags could just be rendered in different colors:

golf=bunker (as sand)
golf=rough (green, lighter than natural=scrub, darker than landuse=grass)
golf=fairway (green, either same or slighly lighter than landuse=grass)
golf=green (green, lighter than landuse=grass)
golf=tee (green, same as golf=green or between golf=fairway and golf=green)

Also, I would VERY much like leisure=golf_course to use a different color than natural=scrub.

Anything more complex than just coloring these areas can wait for later. But just these would greatly improve the visual feedback mappers get from mapping golf courses and would hopefully reduce the instances were golf features are tagged with nonsensical tags purely to make them show up in some way in the map.

kocio-pl:
Would you like to prepare the code for this?

[...]

PR #3675, opened 2019-02-05
[WIP] Render Golf Course Features

### LINKED ISSUE #3650: Change Color or Layering of Golf - Rough to Darker Green
### Expected behavior

Golf - Rough Area layer should be the lowest priority layer and the darkest in color.

Areas of rough, areas of fairway, and "greens" should be distinguishable and look good on the map.

### Actual behavior

When added to a course bounding areas that contain rough, the rough covers or is the same color as greens and fairways.  This causes the map to look bad and remove detail that users add.

Here is an example golf course.  When the rough was added to distinguish between the desert natural areas and the watered/lush areas, the greens and fairways are no longer visible:
https://www.openstreetmap.org/#map=18/36.17548/-115.28432

screenshot_2019-01-31 openstreetmap

I think they are just the same color, so changing Golf-Rough to be a darker shade of green should resolve the issue.  I would make a pull request, but there's a lot of discussion about colors, especially green, so I'd appreciate guidance on how critical these "very zoomed in detail" colors are.

Possibly related to meta ticket:
https://github.com/gravitystorm/openstreetmap-carto/issues/3517

chadrockey:
Fixes #3650 #661 

- Add golf.mss for specific coloring of golf course features (no new colors, just referencing existing)
- Support golf ways as closed polygons

https://www.openstreetmap.org/#map=18/36.17554/-115.28420
https://www.openstreetmap.org/#map=17/48.67095/9.41785
https://www.openstreetmap.org/#map=16/40.1428/-83.1404

Before
currentgolfcourse

After
proposedgolfcourse

@Adamant36 @jeisenbe @imagico Here's how far I got.

There's some issues with laying, I can't figure exactly why the line:
`ORDER BY COALESCE(layer,0), prio, way_area DESC`
isn't working as expected.

Two changes I see that need to be resolved before merging:

- [ ] The change of golf-course with an alpha is planned to be reverted, but if the alpha transparency isn't used, it covers all of the golf features.
- [ ] Line features are rendering behind the way areas.

Mostly I took @imagico 's fork and made it as vanilla/compatible as possible:
https://github.com/imagico/osm-carto-alternative-colors/commit/181360550f12bfdc3589d706e89ae60ae28edbfc#diff-2d6c6d06f63a6fab47286addedefc335

@imagico Are there any licensing needs for the assets (pin svg)/code snippets borrowed from your fork?

### DIFF
--- a/golf.mss
+++ b/golf.mss
@@ -3,6 +3,8 @@
 Coloring specific to golf and golf courses.
 derived from https://raw.githubusercontent.com/imagico/osm-carto-alternative-colors/master/golf.mss
 
+Inspired by the French rendering style and @cquest
+
 */
 
 /* Pinning most grass/plant features relative to grass (natural) and pitch (sporting) */
--- a/golf.mss
+++ b/golf.mss
@@ -1,0 +1,100 @@
+/*
+
+Coloring specific to golf and golf courses.
+derived from https://raw.githubusercontent.com/imagico/osm-carto-alternative-colors/master/golf.mss
+
+*/
+
+/* Pinning most grass/plant features relative to grass (natural) and pitch (sporting) */
+/* The desire is to "move with" changes of these other colors without looking out of place */
+/* Use @water-color for water hazards */
+
+#landcover[zoom >= 14] {
+  [feature = 'golf_rough'] {
+    polygon-fill: darken(@grass, 60%); // Darkest color, rendered lowest
+    [way_pixels >= 4]  { polygon-gamma: 0.75; }
+    [way_pixels >= 64] { polygon-gamma: 0.3;  }
+  }
+  [feature = 'golf_fairway'],
+  [feature = 'golf_driving_range'] {
+    polygon-fill: darken(@grass, 45%); // Typical green grass
+    [way_pixels >= 4]  { polygon-gamma: 0.75; }
+    [way_pixels >= 64] { polygon-gamma: 0.3;  }
+  }
+  [feature = 'golf_water_hazard'],
+  [feature = 'golf_lateral_water_hazard'] {
+    polygon-fill: @water-color;
+    [way_pixels >= 4]  { polygon-gamma: 0.75; }
+    [way_pixels >= 64] { polygon-gamma: 0.3;  }
+  }
+  [feature = 'golf_green'] {
+    polygon-fill: darken(@grass, 30%);  // Manicured sports turf
+    [way_pixels >= 4]  { polygon-gamma: 0.75; }
+    [way_pixels >= 64] { polygon-gamma: 0.3;  }
+  }
+  [feature = 'golf_bunker'] {
+    polygon-fill: @sand; // Typically filled with sand
+    [way_pixels >= 4]  { polygon-gamma: 0.75; }
+    [way_pixels >= 64] { polygon-gamma: 0.3;  }
+  }
+  [feature = 'golf_tee'] {
+    polygon-fill: darken(@grass, 45%); // Distinctive from rough, fairway, green
+    [way_pixels >= 4]  { polygon-gamma: 0.75; }
+    [way_pixels >= 64] { polygon-gamma: 0.3;  }
+  }
+}
+
+#golf-lines[zoom >= 16] {
+  [golf = 'path'][geo = 'line'] {
+    line-color: darken(@pitch, 15%); // Probably a grass trail
+    line-width: 2.0;
+    [name != ''] {
+      text-placement: line;
+      text-name: "[name]";
+      text-face-name: @book-fonts;
+      text-halo-radius: 1;
+      text-halo-fill: fadeout(white, 30%);
+    }
+  }
+  [golf = 'cartpath'][geo = 'line'] {
+    line-color: grey;
+    line-width: 2.0;
+    [name != ''] {
+      text-placement: line;
+      text-name: "[name]";
+      text-face-name: @book-fonts;
+      text-halo-radius: 1;
+      text-halo-fill: fadeout(white, 30%);
+    }
+  }
+  [golf = 'hole'][geo = 'line'] {
+    line-color: black;
+    line-width: 0.5;
+    [name != ''] {
+      text-placement: line;
+      text-name: "[name]";
+      text-face-name: @book-fonts;
+      text-halo-radius: 1;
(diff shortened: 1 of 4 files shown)


imagico:
> There's some issues with laying, I can't figure exactly why the line:
ORDER BY COALESCE(layer,0), prio, way_area DESC
isn't working as expected.

What do you mean by isn't working as expected?

> @imagico Are there any licensing needs for the assets (pin svg)/code snippets borrowed from your fork?

License is the same - but you should attribute the French style and @cquest, which is where the basic idea and coding are from.

Please note that as already indicated in https://github.com/gravitystorm/openstreetmap-carto/issues/3650#issuecomment-460223117 we likely do not want to use this workaround for the current database scheme with golf features in the lines table by default but wait until a database reload that would allow a simpler implementation.  So in this form this PR would mostly be for posterity probably (which does not mean it isn't a useful exercise - although technically that is pretty advanced stuff).

chadrockey:
@imagico I believe that line is meant to order the tags so that the vertical stacking or drawing tags on top of each other works correctly.  However, the main "golf course" tag covers all golf_ polygons, so I worked around it for proof-of-concept by using alpha transparency on the "golf course" landuse tag.

My understanding is that tags drawn last stay on top.  So all prior tags are set as 0, then we set the prio to 1, 10, 20, 30, etc which means they should draw on top, but that doesn't appear to be the case.  Within the golf_ set through, the tags are clearly drawn correctly, even the "bunker in the green" case I chose here to be an especially useful test case.

Also, it's not clear why the lines are rendering behind the polygons as well.

imagico:
The OSM-Carto landcover layer styling uses attachments which have priority over the SQL defined drawing order.  The ac-style does not.  So you would need to include the golf stuff in the same attachments as the other landcover stuff:

https://github.com/gravitystorm/openstreetmap-carto/blob/d4d7d55db07a9c0f39f134bba2f0d646c0bfe3f5/landcover.mss#L79-L83

Note the `prio` ordering is only needed because the line feature have no way_area attribute so they cannot be ordered by way_area.  This is another reason why a database update would be needed for this change - the drawing order would be fairly non-intuitive for mappers.

> Also, it's not clear why the lines are rendering behind the polygons as well.

Your golf-lines layer comes before the landcover layer.

kocio-pl:
From visual point of view, this PR introduces outstanding contrast between elements and especially between them and general area. I don't think it's necessary and it looks bad too, could you please test some softer version?

geozeisig:
The main thing that is missing is golf=hole and golf=tee/pin. That should be displayed with the appropriate number (ref).
golf=fairway, rough, green should be shown **without** landuse=grass  rendered in a color-coordinated manner.

matthijsmelissen:
@chadrockey Thanks for preparing a rendering demo. However, rendering golf-specific tags was rejected in #661. Before working on concrete rendering, I think we should first decide if we want to re-open #661 (I don't think we should).

matthijsmelissen:
I'm going to close this PR for now as it is a PR against a rejected issue. If we decide to re-open the issue, we can re-open the PR too.

Adamant36:
@matthijsmelissen, I was under the impression that #661 was closed because of a database issue, which was subsequently fixed, and not because anyone was against golf things being rendered. It just didn't get reopened for some reason, but I don't think it should disqulify it. 

Last time I checked PRs don't necessarly have to be attached to an issue do they? For instances I can open a PR for some random thing that's never been brought up and the maintainers aren't going to be like "Hhmm, I don't see an issue for this" and then close it right?

matthijsmelissen:
@kocio-pl gave the reason why he closed the issue here: https://github.com/gravitystorm/openstreetmap-carto/issues/661#issuecomment-405083848

PRs without issues are fine (but often discussing the issue first before working on an implementation is a good idea).

Adamant36:
Hhhhmmm, an argument for his comment could be made. I've done a lot of research on it though and my main issue with it is that since the golf tags aren't being rendered most of the places out there that should use it aren't being tagged correctly and people are going with other things that do render, like grass or sand, instead.

 You also can't just take his first comment and ignore the one after it where he said he'd reopen it if there was progress on the database issue https://github.com/gravitystorm/openstreetmap-carto/issues/661#issuecomment-441650499. He just never got around to it, but it would have been re-opened otherwise (approval of reopening is just as good as actually doing it in my mind). 

> PRs without issues are fine (but often discussing the issue first before working on an implementation is a good idea).

Alright. Good to know. I agree about the importance of discussing things first though. I prefer most things have issues and get discussed thoroughly before being implemented.

jeisenbe:
I’m certainly in favor of rendering the golf= tags, and I believe @imagico
also supports rendering, but we just have to wait till the next database
reload.

Adamant36:
> I’m certainly in favor of rendering the golf= tags, and I believe @imagico also supports rendering, but we just have to wait till the next database reload.

For some reason I thought that was dealt with since the last database reload or that it only applied to specific things. Like I don't think bunkers or the tee would be an issue. I could be wrong though. You have a running list of things that can be done with the database reload right? If so, would you mind adding this to the list?

jeisenbe:
Yes, it’s already on the list in issue #3611

Hstore lets us use pretty much any tag in the database, but only if the
feature already has a tag that is imported into the database with the
correct geometry.

Golf=* features are imported as linestrings, not polygons, so we can’t
render areas like greens, fairways and bunkers without reloading the
database in the osm.org rendering servers.

Adamant36:
> Yes, it’s already on the list in issue #3611

That's good. Thanks for keeping track of it. 

> Golf=* features are imported as linestrings, not polygons, so we can’t render areas like greens, fairways and bunkers without reloading the database in the osm.org rendering servers

How come @imagico's fork doesn't have the same problem and why was I able to render all those things when I originally tested it like a year ago? I assume that means osm.org trails behind Kosmtik by quit a lot.

jeisenbe:
The alt-colors fork and the French style both use a workaround. In effect
they query the linestrings and polygons separately, from planet_osm_polygon
and planet_osm_linestring, then merge them prior to rendering.

It’s a cool trick but a couple of the maintainers have said that we should
not use it in this style, rather we should wait to do it the normal way.

See
https://github.com/gravitystorm/openstreetmap-carto/issues/3650#issuecomment-460223117

Issue #1207, opened 2015-01-08
building=house and building=garage should render in slightly different colours


sladen:
It is useful to be able to tell the houses from the garages and traditionally slightly different shades have been used for building=house and building=garage/building=garages.  This is because it makes residential areas more readable.

Recently (early January 2015) a change appears to have been made to lighten buildings, and now garages and houses are the same colour.  This change is probably quite recent because some tiles have it and some don't.

I can't see anything explicit in the changelog, so this is probably more likely to be an accidental regression.  Likely from commit https://github.com/gravitystorm/openstreetmap-carto/commit/01acfb1b3a8bc1d6254bfdae91ab875ee695684e

pnorman:
> It is useful to be able to tell the houses from the garages and traditionally slightly different shades have been used for building=house and building=garage/building=garages. This is because it makes residential areas more readable.

It was more just something that was carried forwards for historical reasons.

> Recently (early January 2015) a change appears to have been made to lighten buildings, and now garages and houses are the same colour. This change is probably quite recent because some tiles have it and some don't.
> 
> I can't see anything explicit in the changelog, so this is probably more likely to be an accidental regression. Likely from commit 01acfb1

Not doing the old confusing logic for rendering some types of buildings lighter was intentional.

Rovastar:
I never understood building=house being lighter - i could just about understand garages.
Actually depending on the list of motivations for carto osm having building=yes could be considered rendering lighter (although it make little sense now they are all lighter) to encourage people to actually make a more descriptive building tag.

mkoniecz:
After #490 list of lighter buildings was sensible, though I am not sure whatever lighter sheds and tents are still necessary.

@sladen - can you give an example of location where it would result in really significant improvements?

sladen:
@mkoniecz, I noticed while working on an area of Wuerzburg, Germany where the buildings had been drawn, but the housenumbers not yet collected.  In particular this is an area of low-density detached housing, where there's roughly a 1:1 ratio of houses to double garages---which lots of "randomly" placed:

* http://overpass-turbo.eu/s/6Wz ... Wuerzburg, showing {building,use}=gauge{,s}
* http://osm.org/#map=16/49.7753/9.9488 Wuerzburg
* http://overpass-turbo.eu/s/6WC Nottingham Park Estate
* http://www.openstreetmap.org/#map=16/52.9510/-1.1621 Nottingham Park Estate
[...]

PR #1153, opened 2014-12-09
Rewrite buildings

matthijsmelissen:
Major rewrite of building code. This will render the buildings much lighter.
- Supersedes #565 
- Resolves #389
- Resolves #327
- Resolves #68
- Resolves #490
- Resolves #533
- Resolves #520
- Resolves #1061

See #565 for rationale.

Major thanks to @pnorman. Thanks to @mkoniecz for place of worship code.

Issue #68, opened 2013-06-13
building outlines are inconsistent


pnorman:
Buildings either have or don't have an outline, with no reason apparent to the user. It is not obvious what tags cause it without detailed inspection

PR #3679, opened 2019-02-12
Differentiate rendering of minor buildings with lighter fill and lighter outline

### LINKED ISSUE #2532: Rethinking major and minor buildings

jeisenbe:
Fixes #2532
Also see Issues #68 and #1207
See PRs #434, PR #490, #565, #568, #1153, #3426 

- Render minor buildings different than other buildings, by using a 10% lighter outline, and a 2% lighter fill color at z15 and higher. At z14 the outline is not used, but the fill color will be 4% lighter than other buildings at this zoom level. 
- The list of "minor buildings" initially includes: huts & yurts, garages, farm buildings, service buildings and sheds, roofs and under-construction buildings. (See #3426 for previous discussion)

### Initial list of minor buildings:
This list of values for the key "building" is based off of tags that have more than 9000 uses in the database, according to  [Taginfo](https://taginfo.openstreetmap.org/keys/building#values)

- barn 
- carport
- cowshed (9 ,159 uses)
- construction
- farm_auxiliary    [building = 'construction'],
- garage
- garages
- ger _[aka "yurt"]_
- greenhouse
- hut
- roof
- service
- shed
- slurry_tank
- stable

## Discussion
- Rendering sheds, garages, greenhouses and farm buildings differently than most buildings has previously been discussed and changed several times in this style over the past 4 years. Most recently it has been discussed in issue #2532 and in PR #3426
- "Major buildings", currently including train stations, airports, and places of worship, are currently rendered darker than other buildings.
- In the past, some buildings were rendered lighter, however the list was not very consistent. At that time most buildings, including building=yes, were rendered much darker than in the current style
- Recently there has been new discussion of improving the building rendering. By rendering some values of building=* with a subtle difference, this should reward mappers who provide more information than simply "building=yes". Because the change in fill and outline color is small, the new minor buildings style is still clearly a type of building, and sufficiently different than other sheds of brown, such as bare_earth and mud. 
- The new rendering is higher contrast with landuse=construction than the current rendering; this improves the contrast between building=construction and a landuse=construction background, especially at z14

It was considered to remove rendering of minor-buildings from z14, however, since farm buildings (like barns) are included in the list of minor buildings and these can be fairly large, they should still be shown. 

### Color comparison
- Mud: at z14, the minor-building-lowzoom color is the same as the current `@building-fill` `#d9d0c9`;  - Lch(84, 5, 68) which has a deltaE of Δ: 4.7 with mud over land - `#e6dcd1`. 
At z15, the `minor-buildings-fill` color is `#ddd5cf` - LCH(86, 4, 67) with Δ: 3.7 to mud `#e6dcd1`.
However at this level the outline color will be shown in `#cec3ba` - LCH(79, 6, 69) which has Δ: 8.9 to mud, so there should not be a risk of confusion.
- Quarry is `#c5c3c3` Lch(79,1,19) with Δ: 5.9 to `@minor-building-fill` `#d9d0c9`, 
and Δ: 7.8 to  `@minor-building-line` `#cec3ba`
- Construction (and brownfield) `#c7c7b4` - LCH(80, 10, 109) has Δ: 6.7 with `@minor-building-line` and Δ: 9.6 with `@minor-building-fill`. This is improved from the current Δ: 6.9 with building-fill-low-zoom
- Residential `#e0dfdf` has Δ: 11.3 with `@minor-building-line` and  Δ: 5.1 with `@minor-building-fill`, (versus  Δ: 5.3 with `@building-fill`)
- Place_of_worship / landuse=religious color is `#d0d0d0`, with Δ: 7.5 to `@minor-building-line` and  Δ: 4.8 to `@minor-building-fill`, (versus  Δ: 5.0 with current `@building-fill`)
- Landuse=garages `#dfddce` has Δ: 9.6 with `@minor-building-line` and  Δ: 5.6 with `@minor-building-fill`, compared to Δ: 4.8 with the current `@building-fill` and Δ: 8.6 with current `@building-low-zoom`

### Other values considered

- static_caravan (used 112, 365 times)
- cabin (132, 994 times)
- collapsed (79, 688)
- ruins (50,549)
- hangar (46,671)
- storage_tank (26, 478)
- silo (11, 220)

These are debatable. 
"Collapsed" and "ruins" should perhaps not be rendered as a building at all. 
"Static_caravans" (aka mobile homes) are similar to houses, but also similar to huts, as are cabins.
Hangars and storage_tanks are large, fairly solid structures, but are not intended for human use. 

There are other values in use with fewer than 10,000 occurrences which might be considered in the future, but for this PR the list has been limited to building values with nearly 10,000 or more uses.

It has also been considered to render building=roof and building=greenhouse with transparency, but this can be considered again in another PR.

## Test renderings with links to the example places:

Luxembourg, building=construction on landuse=construction at center
(Also at the lower zoom levels a building=construction on landuse=commercial is seen, and some building=garage, but quite small)
https://www.openstreetmap.org/#map=14/49.4984/6.1234
z14 Before
z14-construction-buildings-luxembourg-master
After
z14-construction-buildings-luxembourg

z15 Before
z15-construction-buildings-luxembourg-master
After
z15-construction-buildings-luxembourg

z17 Before
z17-construction-buildings-luxembourg-master
After
z17-construction-buildings-luxembourg

Lohbrugge - residential with garages
https://www.openstreetmap.org/#map=14/53.5093/10.2326.png
z14 before
z14-lohbrugge-master-14 53 5093 10 2326
after
z14-lohbrugge-minor-after

z15 before
z15-lohbrugge-master
after
z15-lohbrugge-minor-2-10

z17 before
z17-lohbrugge-master
after
z17-lohbrugge-minor-2-10

Carports 
https://www.openstreetmap.org/#map=18/53.48582/10.17360.png
z18 before
z18-carports-master-18 53 48582 10 17360
after
z18-carports-2-10

Farmyard
z14 Before
z14-farm-master
after
z14-farm-after

z15 Before
z15-farm-master
after
15-farm-2-10

z16 Before
z16-farm-master
after
z16-farm-2-10

### DIFF
--- a/buildings.mss
+++ b/buildings.mss
@@ -7,6 +7,10 @@
 @building-major-z15: darken(@building-major-fill, 5%);  // Lch(70, 9, 66)
 @building-major-z14: darken(@building-major-fill, 10%);  // Lch(66, 11, 65)
 
+@building-minor-fill: lighten(@building-fill, 2%);
+@building-minor-line: lighten(@building-line, 10%);
+@building-minor-low-zoom: @building-fill;
+
 @entrance-permissive: darken(@building-line, 15%);
 @entrance-normal: @building-line;
 
@@ -32,6 +36,27 @@
         [zoom >= 16] {
           polygon-fill: @building-major-fill;
         }
+      }
+    }
+    [building = 'barn'],
+    [building = 'carport'],
+    [building = 'cowshed'],
+    [building = 'construction'],
+    [building = 'farm_auxiliary'],
+    [building = 'garage'],
+    [building = 'garages'],
+    [building = 'ger'],
+    [building = 'greenhouse'],
+    [building = 'hut'],
+    [building = 'roof'],
+    [building = 'service'],
+    [building = 'shed'],
+    [building = 'slurry_tank'],
+    [building = 'stable'] {
+      polygon-fill: @building-minor-low-zoom;
+      [zoom >= 15] {
+        polygon-fill: @building-minor-fill;
+        line-color: @building-minor-line;
       }
     }
   }


kocio-pl:
Greenhouses from https://github.com/gravitystorm/openstreetmap-carto/issues/738#issuecomment-57156377 on current farmland color:

Before

screenshot_2019-02-12 openstreetmap carto kosmtik

After

screenshot_2019-02-12 openstreetmap carto kosmtik 1

jeisenbe:
Here's a building=construction on landuse=religious, next to a place_of_worship (and some church office buildings)

Wamena, GKI Effata
https://www.openstreetmap.org/#map=19/-4.09479/138.94243
z17 Before
z17-gki-master
After
z17-gki-effata-minor-buildings

z18 Before
z18-gki-effata-master
after
z18-gki-effata-minor-buildings

jeisenbe:
More areas with large greenhouses in Flevoland, the Netherlands

Buitenvaart
 https://www.openstreetmap.org/#map=14/52.4209/5.2762
z14 current
z15-buitenvaart-master
after
z14-buitenvaart-greenhouse-after

z15 current
z15-buitenvaart-master
after
z15-buitenvaart-greenhouse-after

Luttelgeest 
https://www.openstreetmap.org/#map=15/52.7407/5.8728
z14 current
z14-luttelgeest-master-15 52 7407 5 8728
after
z14-luttelgeest-greenhouses-after

z15 current
z15-luttlegeest-master
after
z15-luttelgeest-greenhouses-after

jeisenbe:
Tests on all landcovers; minor buildings on the right side of each test square.

z14
z14-minor-buildings-test

At z14, lighter minor buildings are slightly more visible on dark backgrounds such as quarry, forest, and construction/brownfield, and cemetery/graveyard, but less visible on religious landuse.

z16
z16-minor-buildings-text

At z16, the new minor-buildings fill is close to mud and garages, but the outline makes it clear.

polarbearing:
Much in favour. Thanks for the elaborate examples. Code reads plausible. 

Good encouragement for mappers to add the building type.

imagico:
My comments from https://github.com/gravitystorm/openstreetmap-carto/issues/2532#issuecomment-270081278 apply here as well.

jeisenbe:
That comment links to another comment!
I’ll copy and paste; less efficient but it will help me to follow the
discussion:

“I am open to the idea to rendering individual types of buildings in a
distinct way - although there are many different ways of doing that other
than using a different color.
“However all discussion on undefined abstract concepts like major and minor
is a waste of time in my eyes.
“Values for the building tag are a mixture of physical characterizations
(like building=church) and use characterizations (like
building=commercial). For any use of these values we need to see if the
individual tags are well defined and consistently used. Combinations of
these two in a common styling would likely be confusing for the map user.
“I am strictly against rendering in a way that encourages mappers to apply
tags to a building that do not actually apply to the building like for
shops which often only use parts of it. See also the comment by @pnorman
regarding the need for spatial queries.
“In any case the rule of differences in meaning and purpose as mentioned in
#2515 (comment) should IMO be a guiding principle in any decisions here.”

I agree with this; minor buildings should have a clearly different purpose
and use from most buildings.

@imagico, do you think the list of building values in this PR is
appropriate?

The current list of “minor” buildings is limited to non-habitable
structures used for storage of equipment, goods, plants or animals, rather
than for humans.

The exception may be building=hut, which may be used for small, simple
dwellings made from local materials in less developed countries, so we
might consider removing that from the list.

jeisenbe:
Oh, a ger or yurt might also be a dwelling, but I believe these are only
used seasonally, not year-round?

imagico:
> The current list of “minor” buildings is limited to non-habitable structures used for storage of equipment, goods, plants or animals, rather than for humans.

I don't know if that is the case.  Both in the sense that this criterion applies to the buildings on your list and in the sense that it does not apply to other buildings (like building=industrial).

Keep in mind also that many of the tags you listed could be - like building=church - applied to buildings that were originally constructed for a certain function but do not necessarily serve this function any more.  See https://wiki.openstreetmap.org/wiki/Key:building:use.

I would probably rather look for rendering individual tags in a distinct yet subtle form at the highest zoom levels rather than distinguishing aggregate classes with fairly non-intuitive meaning at mid zoom levels already.

jeisenbe:
> building=industrial

Industrial buildings are usually workplaces and most are designed to the same standards as other buildings intended for human occupation. There are probably some industrial buildings that serve as long-term warehouses or storage, but I don't think we should assume this for the entire building=industrial key.

There are certainly other values further down the list (with say 1000 to 9000 uses) that could be added to the list. I'm not sure if it's better to be comprehensive or to start with the most common tags only?

> I would probably rather look for rendering individual tags in a distinct yet subtle form

I'd be in favor of a distinctive rendering for the most common and most important building values, such as building=retail and building=commercial, but it's not possible to render more than a few of the building values in a distinctive fashion. There are hundreds of values for buildings, or rather thousands: tag info has 10542 values.

Garages and sheds both have over 1 million uses, and roof has almost a million. I don't see any need to render these in a prominent way; no one goes looking for sheds or garages. The idea is that there are a number of building values with similar function to garage/shed/roof, including greenhouses, barns, and other agricultural buildings.

Construction could benefit from a distinctive rendering, and perhaps roof could also be different. But I believe this PR would be a start to show that they are not finished, fully habitable buildings.

> the tags you listed could be - like building=church - applied to buildings that were originally constructed for a certain function but do not necessarily serve this function

That's true, but in developed countries it's usually illegal to use a garage, shed or barn as an apartment or office, without remodeling the structure to meet building and fire safety codes.

jeisenbe:
This was the comment in 
https://github.com/gravitystorm/openstreetmap-carto/pull/2515#issuecomment-269970266 that the other comment linked above:

> I believe the comment in I would also suggest to keep in mind the first guideline currently under discussion in #2462:

> > "The difference in rendering between different types of features should foremost be based on their difference in meaning and purpose for the target map users"

. which in case of color unification means we should make sure the differences in meaning and purpose of the features which are unified (...) are small compared to the differences towards other features (...) which are still being rendered in a different color.

jeisenbe:
@imagico, would you prefer that "minor" buildings be rendered the same as other buildings on z14? 

At z14 with the current commit, the difference is small, because there are no outlines, and the lack of building outline also increases the risk that large buildings (eg greenhouses) could be mistaken for a type of landcover or landuse.

kocio-pl:
My intuition is happy with "minor" buildings, but it's good to have some definition and what you proposed works for me - generally not dedicated for prolonged staying of people (garage, roof) or not solid (yurt, roof), additionally with a significant use (like ~10k limit) for practical reasons. I would also think in terms of "support" buildings (security booth is meant for people and can be solid, but the main building it's serving will be office for example, while kiosk selling newspapers is a service in itself).

I don't see how we can show more than 3 generalized types of buildings, but if anyone can come with the idea, it would be nice and it can be implemented then.

I see the reason is not only to have some basic type feedback, but also to make visual clutter smaller.

imagico:
> @imagico, would you prefer that "minor" buildings be rendered the same as other buildings on z14?

I currently see no meaningful function in showing either the "major" or "minor" categories - for the reasons explained.

And to me these are not classifications that naturally derive from tagging practice in OSM, to me these seem to be abstract classifications where developers here have said _i want this to be a meaningful classification_.  Mappers evidently do not see this the same way as illustrated by building=industrial etc. which defy this classification attempt.

Also keep in mind that in terms of mapper feedback the main point would be to encourage people to tag more specifically than building=yes.  If however the majority of building types including the most commonly used ones are rendered just like building=yes that goal is not really served.

Adamant36:
> Also keep in mind that in terms of mapper feedback the main point would be to encourage people to tag more specifically than building=yes. If however the majority of building types including the most commonly used ones are rendered just like building=yes that goal is not really served.

This is the rare instance where I agree with @imagico. IMHO, its more important if people start doing specific tagging on the main building types (retail, commercial, industrial, residential) then something like security booths. Which are fairly niche. As it is, if buildings are rendered based on major/minor building types, then building rendering would have to be further refined to separate the main four from each either. Which I don't see happening. 

At least not without more thought before this is Implemented first, but there's only so many brown building shades you can have before the distinctions becomes meaningless. So if major/minor buildings is implemented as currently discussed, it might be a missed opportunity to do something better.

jeisenbe:
I’m in favor of finding a better, more specific rendering for
building=construction (which is a lifecycle class and implies no access,
and may not yet have a roof or walls) and building=roof/gazebo/etc, which
lack walls.

But I see this PR as a first step, which should be refined further.

Re building=yes; since >90% of buildings are tagged with the value “yes”,
it’s not possible to distinguish these at the moment. But the hope is that
in the future we might render “building=yes” as a minor building type, if
tagging becomes more specific.

(In practice, the large majority of buildings tagged building=yes are
houses. These make up a large majority of buildings on the globe, and the
largest number of specific values in the database.)

There is certainly room in the color space between `@building-fill` and
`@major-building`-fill to add another class of buildings such as
apartments, retail, commercial etc, but I thought I remembered that
@pnorman and some other contributors disagreed with the idea of
differentiating these? Perhaps my recollection is mistaken?

Such a change would be much more significant cartographically than this PR,
so I would rather implement the slightly lighter fill and line color for
outbuildings before creating specific renderings for “public buildings” or
retail etc.

However, I’m happy to try that change first if there is consensus.

matthijsmelissen:
Thanks for the work on this, and thanks for providing an excellent pull request description, @jeisenbe .

That said, I do not think this change is a good idea for the following reasons.

* We are already struggling with a lack of available colours, so to introduce yet another colour we need to have a very good reason. In particular, introducing a lighter colour for buildings reduces the colour space that is available for landuse. I don't think the need for distinguishing major and minor buildings is high enough.

* Like @imagico, I don't think the current tagging (building values) allows us to adequately distinguish minor and major buildings. The tagging of the building tag is quite a mess, with functional and form factors being mixed. At best, we'd have to maintain a long list of minor building values - but given the long tail of building values we'll never be able to compile a complete list, and we'll have to keep adding and adding values.

* I don't find the use of a lighter colour to indicate 'under construction' intuitive. Also, I don't think buildings under construction are necessarily minor buildings (what about the [https://en.wikipedia.org/wiki/Sagrada_Fam%C3%ADlia](Sagrada Familia]?

Note that his reverts part of #1153.

imagico:
The statement from @pnorman on building types was in https://github.com/gravitystorm/openstreetmap-carto/pull/2515#issuecomment-268085855.

jeisenbe:
> I do not think this change is a good idea for the following reasons.

That’s ok, I thought there was a high chance of this PR being rejected when
I started working on it.

> I don't find the use of a lighter colour to indicate 'under construction'
intuitive

I agree that this wasn’t ideal. Perhaps we can try an outline with no fill,
or a dashed outline, for construction.

Does anyone have a good idea for rendering building=roof other than
transparency? Fill with no outline?

Adamant36:
Fill with lines like military area has. Only in building color and the lines closer together? That's all I can think of. I thought there was someone working on a similar thing at some point.

polarbearing:
> Fill with no outline?

Dashed outlines?

bdxd111:
Wouldn't dashed lines or a striped fill increase visual clutter on the map, and thus decrease overall readability?

jeisenbe:
> Wouldn't dashed lines or a striped fill increase visual clutter on the
map,

Yes. These could only be used at the highest zoom levels, i.e. z18/z19

jeisenbe:
Closing due to comments in opposition from 2 maintainers (@imagico (several comments above) and @matthijsmelissen https://github.com/gravitystorm/openstreetmap-carto/pull/3679#issuecomment-464078559) and previous comments suggesting that one building color is best from @pnorman in
https://github.com/gravitystorm/openstreetmap-carto/pull/2515#issuecomment-268085855

PR #3682, opened 2019-02-14
Update condition of index planet_osm_polygon_military

Nakaner:
This reflects changes to the SQL query of layer landuser-overlay introduced by commit 52085d3b6d60cfbe7f39fc44ae6c9345c89c1fd6 (pull request #3469). This partial index was introduced for this layer and its WHERE condition needs to be updated as well if the WHERE condition of the SQL query used by that layer is changed.

There are no visual changes but rendering performance of this layer should become better.

### DIFF
--- a/indexes.sql
+++ b/indexes.sql
@@ -36,7 +36,7 @@ CREATE INDEX planet_osm_polygon_way_area_z10
   WHERE way_area > 23300;
 CREATE INDEX planet_osm_polygon_military
   ON planet_osm_polygon USING GIST (way)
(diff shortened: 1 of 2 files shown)


matthijsmelissen:
Thanks!

matthijsmelissen:
@kocio-pl In the next release, we should recommend users of this style to re-create their indexes (or at least the `planet_osm_polygon_military` index) using the provided style.

kocio-pl:
I see, could you take care of this (or maybe the whole next release)?

PR #3688, opened 2019-02-20
Add rendering for lifeguard facilities

### LINKED ISSUE #3629: Render emergency=lifeguard facilities
Currently, none of the emergency=lifeguard options render.

Visibility of these facilities on the map could well be (without wanting to sound over-dramatic!) a matter of life or death, in that people can see where to go for help if a swimmer is in trouble.

If placing a node using iD, emergency=lifeguard does show an black & white life-ring icon, but it doesn't actually then render on the map eg https://www.openstreetmap.org/node/3491121545.

A better icon would be the one shown on the emergency=life_ring page https://wiki.openstreetmap.org/wiki/Tag%3Aemergency%3Dlife_ring: https://wiki.openstreetmap.org/w/images/0/0a/Life_ring.png
as red is the global distress / emergency colour & would stand out better on the map.

The same icon should be used for all the lifeguard options ie
https://wiki.openstreetmap.org/wiki/Tag%3Aemergency%3Dlifeguard
https://wiki.openstreetmap.org/wiki/Tag%3Aemergency%3Dlifeguard_base
https://wiki.openstreetmap.org/wiki/Tag%3Aemergency%3Dlifeguard_tower
https://wiki.openstreetmap.org/wiki/Tag%3Aemergency%3Dlifeguard_platform
https://wiki.openstreetmap.org/wiki/Tag:emergency%3Dwater_rescue_station
https://wiki.openstreetmap.org/wiki/Tag%3Aemergency%3Dlife_ring 

There are ~2000 of these facilities worldwide, so it is not an insignificant matter.

da1910:
### Rendering lifeguard facilities
Fixes #3629 
- Add rendering of emergency=lifeguard_tower and emergency=lifeguard_base at z=16+ to match other emergency healthcare prominence, in waterside areas lifeguard facilities, are arguably as important as hospitals/doctors surgeries
- Add rendering for life rings at z19, these are less important generally, but are emergency facilities that could be important to saving a life in waterside areas.

Of the lifeguard facilities that are tagged the two most popular are lifeguard_base (278 uses) and lifeguard_tower (489 uses), there has been discussion ongoing for a while about merging some of the lifeguard facility tags, but no consensus was reached. My rationale would be that lifeguard_base would represent permanently manned facilities with storage and maintenance facilities, like RNLI lifeboat stations in the UK. While lifeguard_tower would represent buildings that are manned seasonally and have limited to no facilities, but that immediate assistance would be available when open.

Before
screenshot from 2019-02-20 16-43-14
screenshot from 2019-02-20 16-43-56
screenshot from 2019-02-20 16-45-10

After
screenshot from 2019-02-20 16-38-00
screenshot from 2019-02-20 16-39-19
screenshot from 2019-02-20 16-41-44


Adamant36:
Just a heads up, but you have a space in the fixes. So it won't reference the original issue.

da1910:
I'm happy to split out the lifeguard facilities into a separate request if that simplifies things. The first three pictures show the before, where there are facilities present but currently are either absent completely, or just rendered as plain buildings. I have replaced the last image.

imagico:
Some observations:

* for all the tags suggested (in particular emergency=lifeguard_tower and emergency=lifeguard_base) it is somewhat questionable if they are established taggings.
* emergency=lifeguard_tower and emergency=lifeguard_base do not necessarily imply emergency=life_ring so using the same symbol could be misleading.
* emergency=lifeguard_base according to the sparse documentation does not necessarily imply any public services so showing it in a way implying such could also be misleading.
* symbol recognizability seems to highly depend on the specific colors used which would put severe constraints on adjusting the color scheme (and therefore the goal of adaptability) and clashes with the idea of use of colors in the style otherwise.
* the symbol looks heavy and therefore somewhat unbalanced for emergency=life_ring at z19 compared to the map context.  This would further advance the trend of the map at z19 becoming a pure POI data visualization instead of a map where features are put in context with each other to form a balanced whole.

Adamant36:
Emergency=life_ring is the only one that has the usage numbers for rendering my opinion. 

Emergency=lifeguard_tower base has 487 uses and * emergency=lifeguard_base has only 278. So the usage on both are probably much to low at this point for rendering them even without the consensus problems.

da1910:
Removing the rendering for the other lifeguard facilities addresses most of the issues above, perhaps at a later date when some kind of consensus is reached on the wiki for tagging of lifeguard facilities.

Regarding the icon, I'm not entirely convinced that the argument about recognisability is valid, it's an object that appears next to water in all cases and is coloured in the healthcare colour. It's also clearly a circular object, I struggle to think of another object that it could be confused with, even if one doesn't understand the icon. (It's much more obvious to me than the pill bottle that represents the pharmacy, particularly when it's close to other healthcare services)

I would also argue that the ring with alternating colours is somewhere between very common and universal for a device to save a drowning person. The use of orange and white seems to be the most common set of choices, but I would contend that regardless of colouration the icon is recognisable.

I can see the argument that the icon is a little heavy, when compared to other icons in close proximity it is certainly bold. Perhaps the contributors on the issue can suggest lighter looking forms of the same icon.

project.mml
```diff
@@ -1627,7 +1627,7 @@ Layer:
                                                'dentist', 'dialysis', 'doctor', 'hospital', 'laboratory', 'midwife', 'occupational_therapist', 'optometrist',
                                                'physiotherapist', 'podiatrist', 'psychotherapist', 'rehabilitation', 'speech_therapist', 'yes') THEN tags->'healthcare' ELSE NULL END,
               'advertising_' || CASE WHEN tags->'advertising' in ('column') THEN tags->'advertising' else NULL END,
-              'emergency_' || CASE WHEN tags->'emergency' IN ('phone') THEN tags->'emergency' ELSE NULL END,
+              'emergency_' || CASE WHEN tags->'emergency' IN ('phone','life_ring') THEN tags->'emergency' ELSE NULL END,
```
jeisenbe: I think there should be a space between `'phone'`, and `'life_ring'`, eg `'phone', 'life_ring'`

amenity-points.mss
```diff
@@ -2821,7 +2828,6 @@
     }
   }
 
-  [feature = 'amenity_nursing_home'],
```
jeisenbe: It looks like you've accidentally deleted this line containing`[feature = 'amenity_nursing_home'],`

da1910: Thanks for these, they're now fixed

jeisenbe:
@da1910 did you have interest in trying to redesign the life ring icon to address the problems mentioned [above](https://github.com/gravitystorm/openstreetmap-carto/pull/3688#issuecomment-466060761)

da1910:
I am by no means a designer, if someone else wants to propose a better icon then I can make the change and update this PR

Adamant36:
I can give some general feedback. I'd just modify the icon myself, but I'm not a fan of Inkscape. 
1. The the "ring" should be a full circle inside the 14x14 pixel box. As it is, the left side of it is cut off and the other sides are not full circles. As can be seen in the screen shot. Its fine if the icon doesn't 100% take up the dimensions of the 14x14 pixel box. But as its a circle, all sides should have equal proportions.
Screen shot

2. The black bars are off centered and they shouldn't be. If it where me, id do them in an X pattern. I've included an example of what I am talking about below. 

3. The black bars shouldn't be so thick. There should be more white space between the bars then "bar space." As can also been seen in the example. Or it should at least be equal. IMO the best would be like 50% more white space then bar space or something like that. For instance look at the proportions in a real life ring. The circle lines shouldn't be so thick either. 
high-quality-swimming-pool-life-ring-buoy
(Also notice the bars slant outward. Not inward like they do with the current icon)

You could almost go with something really simple like this
noun_life ring_1018239
noun_lifebuoy_919706
noun_lifeline_192779
noun_support_2441029

jeisenbe:
This will need to be rebased to work with v5.0.0 and above. @da1910 are you interested in trying some of the suggestions of @Adamant36 in https://github.com/gravitystorm/openstreetmap-carto/pull/3688#issuecomment-574532119 perhaps, or should we close this PR?

pnorman:
Closing as abandoned.

Fizzie41:
Trying to start this moving again, thanks. 

To clarify, this rendering (at least in the initial stages) should apply to emergency=life_ring only.

da1910's design from the first post, of a red & white ring icon, with transparent centre, certainly looks good, & would be appropriate.

What else needs to be done, please?

jeisenbe:
See the concerns about the icon design in this comment: https://github.com/gravitystorm/openstreetmap-carto/pull/3688#issuecomment-466060761 - there are some ideas for improvements in https://github.com/gravitystorm/openstreetmap-carto/pull/3688#issuecomment-596051204 though these will need to be tested by rendering on actual map data. I expect that to take several hours of work for someone who already has experience with modifying this map style. If you have not contributed before it can take several more hours to first set up the style and learn how to use it, especially if you do not have prior experience with github.

But on an encouraging note, anyone willing to spend sufficient time on this style can learn to be a contributor. I had never done any coding or development before getting involved with this style less than 2 years ago, and it took about 3 months to get fairly confident with submitting new PRs. 

However, this might not be the best issue for your first attempt, because it will require designing a new icon and getting consensus for rendering.

Instead see the tag “good first issue” for some options which are feasible for a first-time contributor https://github.com/gravitystorm/openstreetmap-carto/issues?q=is%3Aissue+is%3Aopen+label%3A%22good+first+issue%22

If you have a Linux machine see this tutorial by @Ircama:
https://ircama.github.io/osm-carto-tutorials/ 

If you have a Mac or PC you can use docker. I previously wrote a fairly detailed description of how to do this for your first PR, starting with downloading docker and setting up your own github branch: 
https://github.com/gravitystorm/openstreetmap-carto/issues/3782#issuecomment-495926741 - and also see this follow-up comment: https://github.com/gravitystorm/openstreetmap-carto/issues/3782#issuecomment-495957343

Fizzie41:
I've just added this comment to https://github.com/gravitystorm/openstreetmap-carto/issues/3629:

The proposal has now gone through to approve the tag https://wiki.openstreetmap.org/wiki/Tag:emergency%3Dlifeguard as the main way of indicating the presence of lifeguards at a beach etc.

Could I suggest / recommend that it should also be rendered with the life-ring icon as discussed above, with that same icon applying to both emergency=lifeguard & also emergency=life_ring?

Thanks

PR #3689, opened 2019-02-20
Changed rendering level of railway=disused

### LINKED ISSUE #2074: z14 - railway=level_crossing displayed on undisplayed railway=disused
https://www.openstreetmap.org/#map=14/51.4687/21.7988
selection_003

It is rather confusing and not particularly useful.

da1910:
Fixes #2074

- Move railway=disused rendering to z14+ to match level crossings

Currently level crossings over disused railways render without the railway, this looks odd and doesn't make sense.

Before
screenshot from 2019-02-20 17-09-46
screenshot from 2019-02-20 17-00-27
screenshot from 2019-02-20 16-59-29

After

screenshot from 2019-02-20 17-10-11
screenshot from 2019-02-20 17-01-09
screenshot from 2019-02-20 16-59-45

### DIFF
--- a/roads.mss
+++ b/roads.mss
@@ -2263,7 +2263,7 @@ tertiary is rendered from z10 and is not included in osm_planet_roads. */
     }
 
     [feature = 'railway_disused'] {
-      [zoom >= 14 {
+      [zoom >= 14] {
         line-color: #aaa;
         line-width: 2;
         line-dasharray: 2,4;
--- a/roads.mss
+++ b/roads.mss
@@ -831,7 +831,7 @@
       }
     }
 
-    [feature = 'railway_disused'][zoom >= 15],
+    [feature = 'railway_disused'][zoom >= 14],
     [feature = 'railway_construction']['construction' != 'subway'],
     [feature = 'railway_miniature'][zoom >= 15],
     [feature = 'railway_INT-preserved-ssy'][zoom >= 14] {
@@ -1024,7 +1024,7 @@
       }
     }
 
-    [feature = 'railway_disused'][zoom >= 15],
+    [feature = 'railway_disused'][zoom >= 14],
     [feature = 'railway_construction']['construction' != 'subway'],
     [feature = 'railway_miniature'][zoom >= 15],
     [feature = 'railway_INT-preserved-ssy'][zoom >= 14] {
@@ -2263,7 +2263,7 @@ tertiary is rendered from z10 and is not included in osm_planet_roads. */
     }
 
     [feature = 'railway_disused'] {
-      [zoom >= 15] {
+      [zoom >= 14 {
         line-color: #aaa;
         line-width: 2;
         line-dasharray: 2,4;


imagico:
Thanks for the PR.

However i don't think this is an advisable change, showing disused railways at an earlier zoom level would not benefit the map overall IMO.  I also don't think it would solve the problem you try to solve here (that railway crossings are displayed when the railway is not) because there are other railway types (like railway=tram + service=*) that are not shown at z14 either.

Whether it makes sense to have the starting zoom level for railway crossings depend on the type of intersecting railway is a different matter that could be discussed.  I am not sure if the gain justifies the complexity in code and in logic for the map user to understand.

imagico:
Also note your samples seem to be faulty - you have areas in a blue color in there (`#9dd3dd`) that does not otherwise occur in the style i think.  If you show samples it is important that they feature the change (and only the change) you are proposing.

da1910:
For me the presence of level crossings requires a railway of some sort, if the level crossings are rendered at a specific zoom level then the railway should be rendered too, it's like rendering Stiles without a barrier.

Rendering level crossings at different levels based on what kind of railway they cross is definitely a more general solution. The original argument on the issue was that bridges are definitely worth rendering at z14, so when they carry disused railways or cross disused railways it makes sense for the railway to be rendered.

Regarding improvement to the style I think that there's a few cases, when, as in the original ticket, the disused portion of the railway is well away from an existing railway then rendering it is likely to be valuable for wayfinding as it will be a valuable landmark.
When, as in the centre of Prague, there's a large amount of disused rail within an old station or similar then it makes sense to render it, we render the station in the normal way but then don't render any railway.
The last case is probably the one that bears discussion, which is where there are disused sections within live sections, where an argument could me made that it would clutter the map. There I would agree that it's not the best solution, but there I think a discussion could be had to reduce the number of lines shown in general, not just disused ones.

Regarding the blue colour, I have an issue with shapefiles not being shown in mapnik at the moment, even with a clean install of osmcarto,  I think it needs a reimport of all my osm data which I was disinclined to do immediately, it has no impact on the changes proposed.

Get Outlook for Android<https://aka.ms/ghei36>

________________________________

boothym:
>showing disused railways at an earlier zoom level would not benefit the map overall IMO

What are the downsides of showing disused railways at an earlier zoom level? Because IMO the benefits are:

- matches the zoom level of level crossings, so they are not floating about on their own at z14
- bridges/tunnels are shown earlier 
- important landmark/feature, especially in more rural areas. 

Most trams with service=* seem to be service=yard, where having level crossings rendered without tracks isn't as important as in public areas and they may not even be mapped anyway.

imagico:
> What are the downsides of showing disused railways at an earlier zoom level?

Lower zoom rendering of railways is for showing transportation infrastructure, not physical structures.  Disused railways do not serve as transportation.  A disused railway is more like a barrier and barriers start at z16 so rendering from z15 as fairly prominent barriers makes sense. And dashed/dotted lines do not generally work well at small scales, the line topology is not well visible and the dotting generates a lot of noise.  Specifically rendering disused railway bridges and tunnels earlier could be worth discussing.

The fact that this change is proposed as a means to solve a perceived problem with railway crossings and not as a desirable change on its own also tells me that this is not about a change that has merit on its own.

As a mapping related side note - i think the focus on the disused railway as a feature is somewhat ill advised.  The practical significance of a disused railway is often not the railway track itself but the structures around it.  You already mentioned bridges and tunnels but embankments and cuttings are much more widespread.  Mapping and rendering those would IMO be more important than just showing the disused railway as an abstract structure.

matkoniecz:
Thank you very much for the PR!

> Fixes #2074

I would rather say that it hides problem raised in #2074. I amended this report to make clear that problem is that crossings are rendered too early, not that railway=disused is rendered too late.

Also, "after" images are showing changes unrelated to changed railway rendering.

But anyway, after images make clear that disused railway symbols are not readable at this zoom level.

Also, as @imagico mentioned it is dubious that disused railway should be rendered so soon.

matkoniecz:
And again - thank you very much for the PR! Unfortunately in this case I think that this change will not make map better.

Hopefully next one will be merged!

PR #3690, opened 2019-02-22
Use ST_PointOnSurface for stations

pnorman:
Also a minor code cleanup to prio generation

### DIFF
--- a/project.mml
+++ b/project.mml
@@ -1422,13 +1422,7 @@ Layer:
             ref,
             railway,
             aerialway,
-            station,
-            CASE railway
-              WHEN 'station' THEN 1
-              WHEN 'subway_entrance' THEN 3
-              ELSE 2
-            END
-              AS prio
+            station
           FROM
           (SELECT
               ST_PointOnSurface(way) AS way,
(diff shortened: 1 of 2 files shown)


project.mml
```diff
@@ -1423,40 +1422,38 @@ Layer:
...
+            station
+          FROM
+          (SELECT
+              ST_PointOnSurface(way) AS way,
+              name,
+              ref,
+              railway,
+              aerialway,
+              tags->'station' AS station
+            FROM planet_osm_polygon
```
imagico: Don't you need  a `way && !bbox!` here as well?

imagico: Don't you need a `way && !bbox!` here as well?

matthijsmelissen:
> Note that this adds rendering of railway=subway_entrance on polygons

I'd prefer the ST_PointOnSurface changes not to have impact on rendering output. If there is something in the rendering we want to change, we can do so afterwards in a separate pull request.

matthijsmelissen:
> Note that this adds rendering of railway=subway_entrance on polygons

I'd prefer the ST_PointOnSurface changes not to have impact on rendering output. If there is something in the rendering we want to change, we can do so afterwards in a separate pull request.  That way we can keep the technical discussion separate from the discussion on how we want things to render.

pnorman:
I've fixed it to mostly match the existing rendering. Where it doesn't match is the ordering which was previously undefined. It now also orders by type of station before geometry type, because that's possible now.

Issue #2609, opened 2017-04-12
Water is not rendered for parks extending over coastline


artischocke:
In South Florida, I've come across cases where a park extending over coastline areas results in water not being rendered over the green color of the park. One such example is [Oleta River State Park](http://www.openstreetmap.org/way/119291279)
![1](https://cloud.githubusercontent.com/assets/19168045/24948133/9b150cd6-1f69-11e7-9a81-2a40cf333aec.PNG)

where, aside from the very odd location chosen by the rendered for the park's label, coastline features aren't visible over the green cover of the park. Here's the same view with the coastline features visible using OSMI's Coasline view
![2](https://cloud.githubusercontent.com/assets/19168045/24948232/f56084d6-1f69-11e7-871b-1d7a1f9dcbb5.PNG)

I had worked around this issue by creating relations which included the coastline way as the park's boundary, but in this case I believe it would be incorrect to exclude the surface of the water from the park's limits just to get the water to render (i.e. mapping for the renderer), so I thought I'd ask whether this was intended behavior or a bug. If the former, what would be a reasonable but still correct workaround?

imagico:
Here specifically could be considered a tagging problem, i.e. you want to use `leisure=park` for something other than what it is meant for (open, green area for recreation)

More generally this is a duplicate of #1473 which is meant to be be fixed by #1982.

imagico:
The label is indeed wrong by the way, does not appear to be a data glitch since it is the same in the German style - looks like a Mapnik bug to me - likely https://github.com/mapnik/mapnik/issues/3558

artischocke:
@imagico thanks for the answer, but I would say "open, green area for recreation" is a very narrow definition of what a park can be. I take 'open' to mean outdoors as opposed to covered (otherwise places with thick tree cover wouldn't fit the definition), and 'green' would necessitate that recreation activities take place exclusively over grass. Even if we ignore the fact that the name includes the word _Park_ , it would seem to me that the site is indeed a park as most of its [attributes and facilities](https://www.floridastateparks.org/park/Oleta-River) match the definition offered by [Wikipedia](https://en.wikipedia.org/wiki/Park).

All that aside, the official park boundaries are over coastal waters, so whether we think that's correct or not, it's simply something that exists in the wild.

![1](https://cloud.githubusercontent.com/assets/19168045/24969623/a09d126e-1fb1-11e7-9596-5250940b9fb3.PNG)

[...]

PR #3694, opened 2019-02-24
Change from land polygons to ocean polygons

### LINKED ISSUE #1982: Switch to ocean polygons

jeisenbe:
Fixes #1982 
Reverts #3065
See similar previous PR #2066 which was reverted due to #2101  
Also see #2507
Related to solving several issues including #621, #1547, #1781, #2013, #2025, #2609, #2688, #3695, and PR #3670

- Change from rendering  land polygons on an ocean-color background to rendering ocean polygons on a land-color background
- Remove outline from low-zoom coastlines

## Explanation

It is beneficial to render the oceans (that is, all water outside of the openstreetmap coastline) as polygons, rather than the present situation where the continents and islands are rendered as polygons. While this PR does not make any changes in the layering of land and water, with the introduction of water polygons it will become possible to move the oceans above water lines and landcover, which will be helpful in cases where these will show over the ocean. This will allow solving several issues related to mud flats, mangroves, wetlands and areas of parkland at the coast. It also makes it possible to use different colors for rivers and lakes, if desired. It also makes it easier to render maritime borders less prominently

Previously this same change was made in PR #2066 which had to be reverted due to an issue with the simplified water polygons that was seen on a few specific server setups. However, it appears that this issue may no longer be present with current software versions and the current simplified water polygons available from openstreetmapdata.com; see the discussion in issue #2101. Therefore we are attempting to implement ocean polygons using the current

Unfortunately it isn't possible to render a simplified coastline outline, as in PR #3065 and currently, from the simplified water polygons. This could be done by starting to use the generalized polygon files which are also available at openstreetmapdata.com, however this PR is attempting to make the smallest change possible. If the maintainers and contributors agree that a generalized outline is desirable at z0 to z7, I can easier add this feature now or in a separate PR by using the shapefiles at http://openstreetmapdata.com/data/generalized-coastlines

## Test rendering:

Before:
z3-europe-land-polygons-master

After
z3-europe-ocean-polygons-after

- Rendering is same as current rendering at z8 and higher. The only difference is reverting #3065 which simplified the coastline and added a blue outline at z0 to z7

### DIFF
--- a/scripts/get-shapefiles.py
+++ b/scripts/get-shapefiles.py
@@ -47,11 +47,11 @@ settings = {
     },
 
     2: {
-        'directory': 'simplified-land-polygons-complete-3857',
-        'url': 'http://data.openstreetmapdata.com/simplified-land-polygons-complete-3857.zip',  # noqa
+        'directory': 'simplified-water-polygons-complete-3857',
+        'url': 'http://data.openstreetmapdata.com/simplified-water-polygons-complete-3857.zip',  # noqa
         'type': 'zip',
-        'shp_basename': ['simplified_land_polygons'],
-        'long_opt': '--simplified-land'
+        'shp_basename': ['simplified_water_polygons'],
+        'long_opt': '--simplified-water'
     },
 
     3: {
@@ -63,11 +63,11 @@ settings = {
     },
 
     4: {
-        'directory': 'land-polygons-split-3857',
-        'url': 'http://data.openstreetmapdata.com/land-polygons-split-3857.zip',  # noqa
+        'directory': 'water-polygons-split-3857',
+        'url': 'http://data.openstreetmapdata.com/water-polygons-split-3857.zip',  # noqa
         'type': 'zip',
-        'shp_basename': ['land_polygons'],
-        'long_opt': '--land-polygons'
+        'shp_basename': ['water_polygons'],
+        'long_opt': '--water-polygons'
     },
 
     5: {
--- a/shapefiles.mss
+++ b/shapefiles.mss
@@ -12,9 +12,7 @@
 }
 
 .ocean {
-  [zoom >= 0] {
-    polygon-fill: @water-color;
-  }
+  polygon-fill: @water-color;
 }
 
 #icesheet-poly {
--- a/shapefiles.mss
+++ b/shapefiles.mss
@@ -11,22 +11,9 @@
   }
 }
 
-#world {
-  [zoom >= 0][zoom < 10] {
-    polygon-fill: @land-color;
-    polygon-simplify: 0.4;
-    [zoom < 8] {
-      line-color: darken(@water-color,20%);
-      line-simplify: 0.4;
-      line-width: 0.5;
-      line-offset: 0.5;
-    }
-  }
-}
-
-#coast-poly {
-  [zoom >= 10] {
-    polygon-fill: @land-color;
+.ocean {
+  [zoom >= 0] {
+    polygon-fill: @water-color;
   }
 }
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,8 +18,8 @@ env:
   - CARTO=0.18.0 MAPNIK='3.0.0 3.0.12'
 install:
   - npm install carto@$CARTO
-  - mkdir -p data/world_boundaries data/simplified-land-polygons-complete-3857 data/ne_110m_admin_0_boundary_lines_land data/land-polygons-split-3857
-  - touch data/simplified-land-polygons-complete-3857/simplified_land_polygons.shp data/ne_110m_admin_0_boundary_lines_land/ne_110m_admin_0_boundary_lines_land.shp data/land-polygons-split-3857/land_polygons.shp
+  - mkdir -p data/world_boundaries data/simplified-water-polygons-complete-3857 data/ne_110m_admin_0_boundary_lines_land data/water-polygons-split-3857
+  - touch data/simplified-water-polygons-complete-3857/simplified_water_polygons.shp data/ne_110m_admin_0_boundary_lines_land/ne_110m_admin_0_boundary_lines_land.shp data/water-polygons-split-3857/water_polygons.shp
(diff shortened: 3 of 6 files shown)


jeisenbe:
I've opened issue #3695 so that we can discuss whether the coastline outline and simplification or generalization is desireable at low zoom levels.

kocio-pl:
Do you plan to ask affected users if they still have a problem with rendering ocean polygons?

jeisenbe:
> Do you plan to ask affected users if they still have a problem with
rendering ocean polygons?

I’ve now mentioned this PR and tagged the affected users in #2101 -
hopefully they are still using the github.

jeisenbe:
I've now posted to the dev and talk mailing lists, and added a forum post. Are there any other places where we should mention this change to alert potentially affected users?

kocio-pl:
Sounds to me like enough to reach interested persons.

rrzefox:
I've deployed this on my server. Not all tiles have been completely rerendered yet, but I've not seen any completely broken tiles yet (as I did when this was last attempted).
Apart from the difference due to the outline, I've seen differences here:
http://bl.ocks.org/matthijsmelissen/raw/af7a602c222dbf1ff1a2c0d84ed755b7/#9.00/66.7063/13.0012
I'm not sure if that is to be expected or even related to this PR (might just be the tiles on osm.org being too old).

jeisenbe:
Thank you, @rrzefox for testing this again. It's very helpful to be able to see the whole planet for testing. 
I didn't see any differences at the link which you shared above, other than the newly visible protected areas (due to PR #3661). But perhaps I'm not looking at the right place. Is it visible on z9 here? http://bl.ocks.org/matthijsmelissen/raw/af7a602c222dbf1ff1a2c0d84ed755b7/#9.00/66.7063/13.0012

rrzefox:
The protected areas were what I meant - I simply forgot that osm.org hasn't deployed 4.20 yet.

pnorman:
I'm +1 to the change in principle, but haven't given this an in-depth review.

imagico:
Note it would be good to test this preemptively for the new version of the coastline process now available for evaluation on 

https://osmdata.openstreetmap.de/

This should be backwards compatible but would none the less be good to test.  Same applies if we decide to keep using land polygons. See

https://blog.jochentopf.com/2019-03-07-the-new-osmdata-service.html

jeisenbe:
I’ve pushed a new commit which changes the links for the ocean polygons to
osmdata.openstreetmap.de/download/

I did not change the Antarctic ice sheet files. This can be fixed in a
separate PR, to avoid too much complexity.

However, I haven’t yet had time to download the full shapefiles from the
new source; they are very large.

>

imagico:
Note i did not mean we should change to source the coastline data from osmdata.openstreetmap.de right now in this PR.  I meant we should check if it works this way.  You changing the PR makes this easier for people to do of course.  The eventual move to the new source should be done uniformly for all files and be coordinated with the OSMF sysadmins.

scripts/get-shapefiles.py
```diff
@@ -47,11 +47,11 @@
     },
 
     2: {
-        'directory': 'simplified-land-polygons-complete-3857',
-        'url': 'http://data.openstreetmapdata.com/simplified-land-polygons-complete-3857.zip',  # noqa
+        'directory': 'simplified-water-polygons-complete-3857',
+        'url': 'http://osmdata.openstreetmap.de/download/simplified-water-polygons-complete-3857.zip',  # noqa
```
imagico: This does not work with `http`, only with `https`.

scripts/get-shapefiles.py
```diff
@@ -63,11 +63,11 @@
     },
 
     4: {
-        'directory': 'land-polygons-split-3857',
-        'url': 'http://data.openstreetmapdata.com/land-polygons-split-3857.zip',  # noqa
+        'directory': 'water-polygons-split-3857',
+        'url': 'http://osmdata.openstreetmap.de/download/water-polygons-split-3857.zip',  # noqa
```
imagico: ditto.

shapefiles.mss
```diff
@@ -11,22 +11,9 @@
...
-      line-offset: 0.5;
-    }
-  }
-}
-
-#coast-poly {
-  [zoom >= 10] {
-    polygon-fill: @land-color;
+.ocean {
+  [zoom >= 0] {
```
imagico: Don't think that is really necessary.

jeisenbe:
I’ve pushed the requested changes and tested that the rendering is correct.
It should be ready to merge.

jeisenbe:
I downloaded the new shapefiles from [osmdata.openstreetmap.de](https://osmdata.openstreetmap.de/data/water-polygons.html) and checked that the rendering is still correct. I will submit a new PR to suggest using this source, once this PR is merged.

I believe this PR is ready to merge now. Are there any concerns?

imagico:
I have merged this now so we can move forward with the various matters that depend on ocean polygons.

It seems we have consensus that reverting #3065 at least an acceptable side effect of this change - if not desirable on its own because of the problems discussed in #3695.  If i misread anyone on this please say so.

As said we should move to osmdata.openstreetmap.de as shapefile data source with the same release as this change to avoid style users having to switch data source and deal with potential issues twice.

matkoniecz:
> It seems we have consensus that reverting #3065 at least an acceptable side effect of this change

I agree now.

kocio-pl:
Although I think it's better to have coastline simplification, this patch has higher priority for me, so I'm happy to see it merged - thanks.

PR #3699, opened 2019-02-28
Add toll gantries

Zverik:
We have supported `barrier=toll_booth` for quite a long time, but the current trend worldwide is to replace them with drive-through toll gantries that scan your license plate or transponder and deduct payment from your account. In OSM these are mapped with [highway=toll_gantry](https://wiki.openstreetmap.org/wiki/Tag:highway%3Dtoll_gantry). We have only 714 of these at the moment, but the number is steadily rising.

This PR adds an icon and styling for these nodes. Looks like this:

nj

nj2


matkoniecz:
`barrier=toll_booth` is useful to display as both barrier (one needs to stop there), as orientation point and to give feedback to mappers.

`highway=toll_gantry` are not barriers, are rather poor orientation points - so displaying them would be useful as feedback for mappers. But that justification may used to render very single thing with a separate icon.

Also, 714 usages seems to be too early to me.

imagico:
It is my impression that we do not really have an established tagging system here at the moment.  We have the concept of [enforcement relations](https://wiki.openstreetmap.org/wiki/Relation:enforcement) - which although not much used for toll systems (almost all uses are maxspeed) would make much more sense and would be more useful for applications that want to do more than just display the physical device.  I would not be comfortable with pushing the much simpler _device only_ mapping against the more complex but also more useful relation mapping here given the current prevalence in use.  Note that while both tagging ideas could be used in combination there is currently neither documentation nor significant practical use of this.  It all looks like there is not yet really a consensus on this matter.

Apart from that i also don't think the chosen symbol works here - it shows a perspective view of the gantry which can be confusing depending on the direction of the highway at the location and it does not provide usable feedback on accurate placement of the node on the highway - in fact both examples look like they have an offset to the road (which is due to the center of the symbol not being in the middle between the two pillars of the gantry).

Zverik:
Updated the icon.

Maintaining toll barriers and gantries on the East Coast of the US, I can't help noticing their increase. In Belarus where I live, I believe there are no toll booths left, only toll gantries.

Enforcement relation is used to mark regular cameras, which do not affect a driver's bill. Toll gantries mark points which would 100% cost you if you pass them. With that, I think they are useful to be displayed. Like, we display country borders and military areas already, despite these mostly not detectable on the ground.

nj2
nj

Gazer75:
I am in the process of moving all the AutoPass gantries over to this tag in Norway. Should be around 350-400 of them

Gazer75:
Make sure this is not just looking at highway=toll_gantry. There are many one directional ones in Norway which use the forward and backward subkey. Like highway:forward and highway:backward.

Zverik:
Toll gantries are [displayed](https://josm.openstreetmap.de/changeset/15201/josm) in the recently puslished tested version of JOSM. In the five months the number of toll_gantry points has risen 60%, to 1174.

pnorman:
I'm looking at this and wondering if it makes sense to have a unified icon with toll booths in order to reduce icon proliferation. What do other maps do for points where a toll is charged, do they use different icons for the different types of points, or one common icon?

Gazer75:
These are not toll booths though...

Zverik:
> What do other maps do for points where a toll is charged, do they use different icons for the different types of points, or one common icon?

Did a little research for that. The answer is yes. JOSM uses different icons (try to guess which is which):

Screenshot 2019-08-15 at 18 29 04

OsmAnd uses separate icons for [toll_booths](https://github.com/osmandapp/OsmAnd-resources/blob/master/icons/svg/barrier_colored/toll_booth.svg) and [toll_gantries](https://github.com/osmandapp/OsmAnd-resources/blob/master/icons/svg/transport_colored/toll_gantry.svg).

Gazer75:
OsmAnd use different icons...

jeisenbe:
Current usage has increased to 1207, including 1193 nodes: https://taginfo.openstreetmap.org/tags/highway=toll_gantry#overview

Similar tags: 

`enforcement=toll` is used 426 times on nodes: https://taginfo.openstreetmap.org/tags/enforcement=toll 

`highway=speed_camera` is used 40,000 times: https://wiki.openstreetmap.org/wiki/Tag:highway%3Dspeed_camera

`man_made=surveillance` is used >80,000 times: https://wiki.openstreetmap.org/wiki/Tag:man_made%3Dsurveillance

`man_made=gantry` is used 2300 times:
https://taginfo.openstreetmap.org/tags/man_made=gantry

jeisenbe:
@pnorman - what level of usage would be considered sufficient to reconsider rendering in this case?

Should we close this PR for now, or leave it open and wait for usage to increase to that level?

Gazer75:
The use has increased to about 1650 now. I just updated some one way gantries to use direction=* instead of using highway:forward and highway:backward. Those are not included in that number so it will be around 10-20 more.

Adamant36:
> Asking to have it more common than man_made=gantry is impossible.

I don't think he meant that it has to be more common then man_made=gantry to be rendered, just that the usage should be over the current amount for that tag of 2300 (around 2000 uses is often the threshold for rendering of most things).

jeisenbe:
This PR would need to be rebased for v5.0.0, so I am going to close it for now. But when usage of the tag increases, please submit a new, rebased PR.

Gazer75:
Usage is up to over 2000 now...

Gazer75:
@jeisenbe Please revisit this now. Taginfo shows 3000 of them so far.

cicku:
What is the purpose of tagging gantries?

pnorman:
> @jeisenbe Please revisit this now. Taginfo shows 3000 of them so far.

A new rebased PR needs to be submitted.

Gazer75:
> What is the purpose of tagging gantries?

They function like a toll booth, but you do not slow down. Will allow routers to mark toll roads.


> A new rebased PR needs to be submitted.

I wouldn't know what to do for that so i guess @Zverik would have to resubmit this.

Adamant36:
> They function like a toll booth, but you do not slow down. Will allow routers to mark toll roads.

Routers aren't really effected by what's rendered in this are they? What with them being separate pieces of software and whatnot. I know it would increase usage, but, personally, I'm of the opinion that if something along roadways is to be rendered at this point that there are probably things higher on the priority list that have way more usage then this currently has or will have in the future (I looked into it a while back, and there is. I just can't remember what right now).

Gazer75:
Problem is that people tag for visuals and thus a lot of these get toll booth instead.
I've had to fix people actually changing back my toll gantry tag to toll booth simply because it was no showing on map.

This renderer on the main osm page is very important for usage of tags. Things that actually render are far more likely to be mapped and tagged properly.

Adamant36:
> Problem is that people tag for visuals and thus alot of these get toll booth instead.

Rendering for the mapping isn't any better for mapping for the rendering. 

> I've had to fix people actually changing back my toll gantry tag to toll booth simply because it was no showing on map.

It's possible they were just confused. Toll gantries aren't super common and it's hard to tell without actually driving through one if it is that or a toll booth. Since they all sorta look the same on satellite photos. 

> This renderer on the main osm page is very important for usage of tags. Things that actually render are far more likely to be mapped and tagged properly.

I use to think that, but not so much anymore now that there are an abundance of other options. Especially on mobile. You specifically mentioned rendering this for the purposes of navigation, and I would argue that people aren't using the main site for  that. There isn't thing that the style renders specifically for that either from what I know. Tool booths likely are because they are geographical features and orientation points. Maybe barriers, but I'd argue they are rendered for similar reasons. It's not specifically to aid in navigation though. At least not that I'm aware of. 

Does OsmAnd rendering toll gantries? If not, I'd ask for them to. Since they are used for navigation and seem to much more laxed rendering standards. I often use OsmAnd as mapping feedback myself. Especially for mapping things related to navigation. So, I don't think it's a horrible or unrealistic option.

Gazer75:
OsmAnd started rendering them pretty early when this tag was approved.
Not using it myself as I never drive where I don't know the road/route.

Gazer75:
Usage has increased to over 4000 now, so please consider reopening this.

imagico:
It has already been mentioned in the reply to your request in https://github.com/gravitystorm/openstreetmap-carto/pull/3699#issuecomment-783266607 (https://github.com/gravitystorm/openstreetmap-carto/pull/3699#issuecomment-785558783) that this is a stale PR.  You are welcome to submit a new PR or to open an issue about adding rendering support for this.  The latter might be a good idea to do first because there are still competing tagging systems in place here (enforcement relations with `enforcement=toll` have about 3000 uses).

Gazer75:
enforcement=toll is not the same. That is just a check system in Germany for HGV. There is no actual toll collected passing through those.

cicku:
Can you stop bumping old threads here?

PR #3701, opened 2019-03-02
Render landuse with grass color from z5 instead of z10

### LINKED ISSUE #3696: Landuse=grass is not rendered at z9 and below
### Expected behavior
It should be apparent if a region has landcover tagging or if the landcover is still not mapped at z7 to z9

### Actual behavior
Areas with landuse=grass, such as the Netherlands and northern Belgium, look gray (unmapped) at z9 and below, but are clearly fully mapped when seen at mid and high zooms

Utrecht, z10
https://www.openstreetmap.org/#map=10/52.0331/4.7859
z10-utrecht-rotterdam-farmland

z9
https://www.openstreetmap.org/#map=9/52.0322/4.7873
z9-utretch-to-rotterdam

vholten:
Fixes #3696

- Add landuse=grass to the landcover-low-zoom layer and render it from z5 instead of z10

This allows large grassy areas, such as in the Netherlands, to be visible at low zoom. 

https://www.openstreetmap.org/#map=9/52.0322/4.7873
z9 before
z9-master

z9 after
z9-new

https://www.openstreetmap.org/#map=8/52.0322/4.7873
z8 before
z8-master

z8 after
z8-new

### DIFF
--- a/landcover.mss
+++ b/landcover.mss
@@ -359,7 +359,7 @@
   [feature = 'natural_grassland'][zoom >= 5],
   [feature = 'landuse_meadow'][zoom >= 5],
   [feature = 'landuse_grass'][zoom >= 5],
-  [feature = 'landuse_village_green'][zoom >= 10] {
+  [feature = 'landuse_village_green'][zoom >= 5] {
     polygon-fill: @grass;
     [way_pixels >= 4]  { polygon-gamma: 0.75; }
     [way_pixels >= 64] { polygon-gamma: 0.3;  }
--- a/landcover.mss
+++ b/landcover.mss
@@ -358,7 +358,7 @@
 
   [feature = 'natural_grassland'][zoom >= 5],
   [feature = 'landuse_meadow'][zoom >= 5],
-  [feature = 'landuse_grass'][zoom >= 10],
+  [feature = 'landuse_grass'][zoom >= 5],
   [feature = 'landuse_village_green'][zoom >= 10] {
     polygon-fill: @grass;
     [way_pixels >= 4]  { polygon-gamma: 0.75; }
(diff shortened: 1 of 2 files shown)


jeisenbe:
+1 👍

matthijsmelissen:
Looks good to me. Some colors, like grass/forest, are hard to distinguish, but that's not the fault of this pull request.

jeisenbe:
I’m actually surprised that these areas are mapped as landuse=grass instead
of landuse=meadow but perhaps this better translates the word used in Dutch?

Adamant36:
From what I remember with early discussions landuse=grass is one of those tags that's a little vague, over mapped in places where it probably shouldn't be due to that, and should probably be surface tag instead. More then likely in this case its over applied because managed nature of it. So landuse=meadow would probably be better, but there's still lots of other instances where rendering grass at z5 would be beneficial.

imagico:
I don't actually see a good reason for any starting zoom level differentiation here for features that are otherwise rendered identically at all zoom levels.  This is unexpected and confusing for mappers.  So the same should IMO be done for `landuse=village_green` as well.

aceman444:
I'm also surprised there are such large areas of landuse=grass, when this is intented for small managed patches of grass in residential/highway-side areas. Shouldn't the large country-side areas be natural=grassland or landuse=meadow or even landuse=farmland (if the grass is seeded to be farmed later).

imagico:
@vholten - i would prefer to remove all differentiation in starting zoom levels between different grass color tags, i.e. also for landuse=village_green.  If you can change that i would merge this.

vholten:
@imagico That sounds good, I'll probably do it this weekend.

vholten:
PR updated: landuse=village_green is now also rendered from z5 instead of z10. I couldn't find any village_green areas that are large enough to be visible below z10 however.

@imagico Do you also want leisure=garden rendered from z5? It has a grass color and is currently rendered from z10.

imagico:
Thanks.

leisure=garden is a somewhat different matter since that is differentiated in rendering at higher zoom levels.

PR #3709, opened 2019-03-08
Create new layer for "ref" of minor roads and paths 

jeisenbe:
Partially replaces PR #3663
Follow-up to PR #3654

1) Move unclassified, residential and track highways, and aeroways "ref" to a later layer, after name labels are rendered

2) Change text-halo to standard size for highway=unclassified/residential refs and aeroway refs

3) Change residential / unclassified road "ref" size to 11 point at z17, to be the same size as the name text labels at this zoom level and higher. 

### Explanation

**1) The minor roads, aeroways, and tracks with text-based ref labels currently render in the same layer as the shields for major roads.** 

- The ref text is rendered before the name label, and can block the name from rendering when spaces is limited.
- Major road shields render much sooner than the ref text labels of minor roads, which are not shown until z15 or z16.
- Major road shields are usually more important than the name; the number of a motorway is used more often, but for minor roads, such as residential streets, the name is more useful and should be shown with higher priority
- Changing residential, unclassified and tracks to a later layer will fix this rendering. Aeroways "ref" text labels are also moved to this new layer because they are also rendered only at z15 and higher, though this does not change the cartography because they do not have a rendering for names

**2) The "ref" labels rendering for residential/unclassified roads needs minor adjustments**
 At z17, the "name" text label increases to 11pt font size, but the ref does not change size till z18 currently. This should be adjusted so that the "ref" label is the same size as the name label, as at z15 and z16.

**3) The "ref" halo needs to be changed to standard size.**
 Right now it is 2 pixels wide, which would overlap with the casing at certain zoom levels. 

**highway=unclassified with name and ref** - Basilicata, Italy 
https://www.openstreetmap.org/#map=17/40.3776321/15.7281581
Before z17
via-chiusululle-unclassified-master
z18
z18-via-chiusululle-master
After z17
chiusululle-z17-after
z18
chiusululle-z18-after

**highway=residential with name and ref** - Basilicata, Italy 
https://www.openstreetmap.org/#map=17/40.1412792/16.0893142
Before z17
via-vittorio-res-z17-master
z18
via-vitorio-z18-master
After z17
z17-via-vittorio-after
z18
via-vittori-z18-after

### DIFF
--- a/project.mml
+++ b/project.mml
@@ -1883,10 +1883,13 @@ Layer:
                 SELECT
                     osm_id,
                     way,
-                    CASE WHEN highway IN ('motorway', 'trunk', 'primary', 'secondary', 'tertiary') THEN highway ELSE NULL END AS highway,
+                    COALESCE(
+                      CASE WHEN highway IN ('motorway', 'trunk', 'primary', 'secondary', 'tertiary') THEN highway ELSE NULL END,
+                      CASE WHEN aeroway IN ('runway', 'taxiway') THEN aeroway ELSE NULL END
+                    ) AS highway,
                     string_to_array(ref, ';') AS refs
                   FROM planet_osm_line
-                  WHERE highway IN ('motorway', 'trunk', 'primary', 'secondary', 'tertiary')
+                  WHERE (highway IN ('motorway', 'trunk', 'primary', 'secondary', 'tertiary') OR aeroway IN ('runway', 'taxiway'))
                     AND ref IS NOT NULL
               ) AS p) AS q
           WHERE height <= 4 AND width <= 11
@@ -1897,6 +1900,8 @@ Layer:
               WHEN highway = 'primary' THEN 36
               WHEN highway = 'secondary' THEN 35
               WHEN highway = 'tertiary' THEN 34
+              WHEN highway = 'runway' THEN 6
+              WHEN highway = 'taxiway' THEN 5
               ELSE NULL
             END DESC NULLS LAST,
             height DESC,
@@ -2069,13 +2074,10 @@ Layer:
                 SELECT
                     osm_id,
                     way,
-                    COALESCE(
-                      CASE WHEN highway IN ('unclassified', 'residential', 'track') THEN highway ELSE NULL END,
-                      CASE WHEN aeroway IN ('runway', 'taxiway') THEN aeroway ELSE NULL END
-                    ) AS highway,
+                    CASE WHEN highway IN ('unclassified', 'residential', 'track') THEN highway ELSE NULL END AS highway,
                     string_to_array(ref, ';') AS refs
                   FROM planet_osm_line
-                  WHERE (highway IN ('unclassified', 'residential', 'track') OR aeroway IN ('runway', 'taxiway'))
+                  WHERE highway IN ('unclassified', 'residential', 'track')
                     AND ref IS NOT NULL
               ) AS p) AS q
           WHERE height <= 4 AND width <= 11
@@ -2084,8 +2086,6 @@ Layer:
               WHEN highway = 'unclassified' THEN 33
               WHEN highway = 'residential' THEN 32
               WHEN highway = 'track' THEN 30
-              WHEN highway = 'runway' THEN 6
-              WHEN highway = 'taxiway' THEN 5
               ELSE NULL
             END DESC NULLS LAST,
             height DESC,
--- a/project.mml
+++ b/project.mml
@@ -2070,12 +2070,12 @@ Layer:
                     osm_id,
                     way,
                     COALESCE(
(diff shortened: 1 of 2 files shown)


jeisenbe:
Does anyone have time to look at this PR?

kocio-pl:
I don't have much time lately, but looking at the rendering this is a good idea.

imagico:
I am sorry for the long wait - I looked at this and am concerned about two things:

* by moving aeroway refs to the new layer you make them lower priority than road names.  I am not sure this would work so well - in particular for service roads and tunnels on airports
* rendering refs with labels indistinguishable from names does not seem a good idea.

jeisenbe:
> * by moving aeroway refs to the new layer you make them lower priority than road names.  I am not sure this would work so well - in particular for service roads and tunnels on airports

The "ref" for runways (and taxiways) is only rendered at z15 and greater, and it's rendered one time at the center of the way. So this is only an issue for airports where a named road crosses the center point a runway or taxiway. I can think of several examples of paths and roads that cross over or level with runways (eg Gibraltar, LAX, some airstrips in my area) but I don't know of any where the road is close enough to the runway center for this to cause interference with the label, so I doubt this is every an issue in practice. 

However, It would be easy enough to keep the aeroways in the main `roads-text-ref` layer. I only moved them because they render at z15 and higher only, so I thought it would slightly improve rendering performance for z13 and z14, which are included in `roads-text-ref` but not in the new layer.

> * rendering refs with labels indistinguishable from names does not seem a good idea.

This is the current situation with "ref" on residential, unclassified and track roads. The PR only makes 2 tiny adjustments for consistency:

```
-      [zoom >= 18] {
-        text-size: 10;
+      [zoom >= 17] {
+        text-size: 11;
```

```
-      text-halo-radius: 2;
```

If we want the 'ref' to be clearly distinct from the road name label on this minor roads, there would need to be a more significant distinction than the halo radius or a 1 point font size difference at z18 and z19. 

I could try rendering the ref with italic font, perhaps? Or should this be a separate PR?

jeisenbe:
Does anyone else have comments about this? Should I move the aeroways back to the earlier layer? Does the "ref" label need to be more clearly distinguished from the "name" label?

jeisenbe:
1) Do we need to keep  aeroways in the earlier layer? 
2) Should the "ref" rendering be changed in this PR, or can it wait till later?
See comments above https://github.com/gravitystorm/openstreetmap-carto/pull/3709#issuecomment-478300867

imagico:
Would be good to have another opinion on this, maybe from @matkoniecz, @kocio-pl or @sommerluk.  I mentioned two points of concern, but these are not essential for me.  Another look at this from someone else would be useful.

sommerluk:
Okay. I didn’t look to the code. But the idea to unify the `ref` label rendering with the `name` label rendering seems good to me. The current situation is that we have some differences in font size between `name` and `ref`, and these differences are not big enough within the same zoom level (and not systematic enough across different zoom levels) to allow a clear distinction anyway. Personally I also think different font sizes on the same zoom level for the same object in the same rendering (here: along the line) make the map harder to read (less intuitive).

Using the italic font instead of the regluar font for `ref` might also work. (Side note: We have currently no systematic approach about when using italic fonts and when not.)

jeisenbe:
@sommerluk - any thoughts on including aeroway refs in this new layer
or not? I'm happy to change it, but I hoped it could be slightly
better performance this way.

jeisenbe:
I've moved aeroway ref rendering back to the original layer (roads-text-ref)
I also changed the rendering of minor roads and aeroways to use an oblique (italic) font, as suggested above

jeisenbe:
Test renderings with oblique font for ref. It's a subtle change:

highway=unclassified with name and ref - Basilicata, Italy
https://www.openstreetmap.org/#map=17/40.3776321/15.7281581
z18-ref-oblique-via-chiusululle

highway=residential with name and ref - Basilicata, Italy
https://www.openstreetmap.org/#map=17/40.1412792/16.0893142
z18-ref-oblique-via-vittorio
z17-ref-oblique-via-vittorio

Runway with ref in new oblique font:
z15-wamena-runway-15-33

(Compare to current rendering:)
z15-current-runway-rendering

jeisenbe:
Any comments on the new commit?

jeisenbe:
@imagico, your original 2 points of concern were addressed by the most recent commits.

jeisenbe:
@imagico has approved this now, and it looks like @kocio-pl and @sommerluk were in favor, so I'm going to merge this PR tomorrow, once I figure out when the changelog should be updated.

pnorman:
> once I figure out when the changelog should be updated

It's not necessary, but a big help when releasing. Just edit the changelog directly on master.

PR #3715, opened 2019-03-12
Create groundskeeping.svg

woodmiser:
This is my first attempt at a pull request, so If something is not correct, please correct me.
I created a new tag and a wiki page for it. 
The tag is:  shop=groundskeeping
 Should be on taginfo with the next update. And I created this icon to go with it.  I would like to get it into the project.  Thank you.
groundskeeping

### DIFF
--- a/symbols/shop/groundskeeping.svg
+++ b/symbols/shop/groundskeeping.svg
@@ -1,0 +1,16 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!-- Created with Inkscape (http://www.inkscape.org/) -->
+<svg width="20" height="20" version="1.1" viewBox="0 0 20 20" xmlns="http://www.w3.org/2000/svg" xmlns:cc="http://creativecommons.org/ns#" xmlns:dc="http://purl.org/dc/elements/1.1/" xmlns:rdf="http://www.w3.org/1999/02/22-rdf-syntax-ns#">
+ <metadata>
+  <rdf:RDF>
+   <cc:Work rdf:about="">
+    <dc:format>image/svg+xml</dc:format>
+    <dc:type rdf:resource="http://purl.org/dc/dcmitype/StillImage"/>
+    <dc:title/>
+   </cc:Work>
+  </rdf:RDF>
+ </metadata>
+ <g transform="translate(0 -291.71)">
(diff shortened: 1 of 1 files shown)


jeisenbe:
Welcome, and thank you for contributing to Openstreetmap!

Before we add a new feature to the rendered map, it should be
discussed on the Tagging mailing list:
https://lists.openstreetmap.org/listinfo/tagging

There is also a proposal process for new features. While this is not
required, it would be a good idea to consider making a proposal for
your new tag: https://wiki.openstreetmap.org/wiki/Proposal_process

Before adding a new feature to the map, there should be proof that it
has been adopted by many mappers. Usually it's best to get the feature
added to a major editor, such as iD
(https://github.com/openstreetmap/iD) or JOSM
(https://github.com/openstreetmap/josm). Then, once more mappers are
using the tag, it is more likely to be added to the rendering in this
map style.

When you make a PR to add a new feature, please show test renderings
of the map with the feature added. You can do this by using Docker.
Please read and "docker.md" () to learn how to set this up on a Mac or
PC.

You should also provide a link to the wiki page documenting the tag. I
wasn't able to find this tag by searching the wiki today for
"groundskeeping."

There are more suggestions about how to contribute to this project at
"contributing.md"
(https://github.com/gravitystorm/openstreetmap-carto/blob/master/CONTRIBUTING.md)
- but this assumes some basic familiarity with Git, Github and coding.

This whole process may take at least a few months, but don't give up!
The first time is hard, but it will get easier.

Adamant36:
You have to specifically search for shop=groundskeeping for some reason (probably not enough key words in the article yet). 

https://wiki.openstreetmap.org/wiki/Tag:shop%3Dgroundskeeping

matkoniecz:
Used only 28 time worldwide - https://taginfo.openstreetmap.org//search?q=shop%3Dgroundskeeping

It is a bit early to render it with a special icon.

imagico:
As a general remark to @woodmiser and possibly other contributors:  Developing a feature addition that is actually beneficial for this style overall is much harder than it might look like at the first glance.  This style shows really a lot of features already so adding more of them in a matter that works well is tricky.  So for getting started i would rather recommend looking for simple bugs and consistency issues that can be solved with a simple change.  There are plenty of those although not all of them have issues documenting them and even fewer are indicated as good first issues:

https://github.com/gravitystorm/openstreetmap-carto/issues?q=is%3Aopen+is%3Aissue+label%3A%22good+first+issue%22

PR #3718, opened 2019-03-15
stop rendering shop=yes [ready for review]

### LINKED ISSUE #3697: stop rendering shop=yes
Originally only shop with their own icons were rendered.

Later popular shop values were also rendered - as dots without icons. This was deliberate decision to not render cases like `shop=supermarkte`

Since #2415 all shop values except `'no', 'vacant', 'closed', 'disused', 'empty'` are rendered.

In #2099 I proposed to stop rendering `shop=yes` to encourage more specific tagging but it was rejected.

I propose to reconsider this as since that time usage of `shop=yes` is increasing what is undesirable.

In 2016 "[Shop=yes is used only for 3.44% of all the shop types](http://taginfo.openstreetmap.org/keys/shop#values)" and "[10% of shop=yes](http://taginfo.openstreetmap.org/tags/shop=yes#combinations) are combined with amenity=fuel (from old? JOSM preset) which are not rendered anyway as shop." (quoted from https://github.com/gravitystorm/openstreetmap-carto/issues/2099#issuecomment-254543095 and https://github.com/gravitystorm/openstreetmap-carto/issues/2099#issuecomment-254580699 ). It was used 81 483 times ( https://github.com/gravitystorm/openstreetmap-carto/issues/2099#issuecomment-254836689 )

Currently shop=yes usage is growing and reached 149 194 uses (4.05% of all shops). It includes 9 476 `shop=yes` on `amenity=fuel` (6.35% of `shop=yes`), for 139 718 unspecified shops.

EDIT: 149 222 at 2019-02-25

matkoniecz:
fixes #3697

Originally only shop with their own icons were rendered.

Later popular shop values were also rendered - as dots without icons. This was deliberate decision to not render cases like shop=supermarkte

Since #2415 all shop values except 'no', 'vacant', 'closed', 'disused', 'empty' are rendered.

Given that any, even never used shop value is rendered and `shop=yes` is unreasonably popular it is desirable to encourage more specific shop tagging.

Dropping `shop=yes` may encourage using some new shop values and in some cases people will mistakenly create new tags duplicating existing ones,
but standard shop types are still encouraged by using icons.

shop tag currently has long tail of unusual shop types (with most tagged correctly - for example `shop=maps`, some shop like this really exist),
so data consumers need anyway some strategy for handling rare shop types.

`shop=yes` is currenly reported as a tagging mistake by decent validators (for example JOSM and Osmose).

Note that this is not deprecation of a tag but stopping to render a deprecated tag.
This tag was never considered as a sufficient for tagging shop type

https://www.openstreetmap.org/way/237537359 and other now fixed objects, found using http://overpass-turbo.eu/s/H0Q

Before (this place has both node and area with `shop=yes`)
Selection_002

After
Selection_003

### DIFF
--- a/project.mml
+++ b/project.mml
@@ -1513,7 +1513,7 @@ Layer:
                 'amenity_' || CASE WHEN amenity IN ('vending_machine') AND tags->'vending' IN ('excrement_bags', 'parking_tickets', 'public_transport_tickets') THEN amenity ELSE NULL END,
                 'advertising_' || CASE WHEN tags->'advertising' in ('column') THEN tags->'advertising' else NULL END,
                 'emergency_' || CASE WHEN tags->'emergency' IN ('phone') AND way_area IS NULL THEN tags->'emergency' ELSE NULL END,
-                'shop' || CASE WHEN shop IN ('no', 'vacant', 'closed', 'disused', 'empty') OR shop IS NULL THEN NULL ELSE '' END,
+                'shop' || CASE WHEN shop IN ('yes', 'no', 'vacant', 'closed', 'disused', 'empty') OR shop IS NULL THEN NULL ELSE '' END,
                 'leisure_' || CASE WHEN leisure IN ('amusement_arcade', 'beach_resort', 'bird_hide', 'bowling_alley', 'dog_park', 'firepit', 'fishing',
                                                     'fitness_centre', 'fitness_station', 'garden', 'golf_course', 'ice_rink', 'marina', 'miniature_golf',
                                                     'nature_reserve', 'outdoor_seating', 'park', 'picnic_table', 'pitch', 'playground', 'recreation_ground',


matkoniecz:
Note that all and each shop value except blacklisted `'no', 'vacant', 'closed', 'disused', 'empty', 'yes'` will continue to be rendered in this style.

Dropping rendering of `shop=yes` will stop rendering of about 140-150k objects (note that some `shop=yes` are not rendered due to their presence on `amenity=fuel`).

imagico:
I think with #2415 taken as given this change makes sense but i would be in favor of revisiting the decision of #2415 to render a catchall for shops.

We currently have 105k different values for shop=* but only ~850 with more than ten uses - which means that there are nearly as many special choice values for shop=* out there as there are cases of shop=yes and many of them are just typos or pointless synonyms for existing established values.

I am kind of inclined to suggest rendering neither shop=yes nor any special choice values - but that would require creating and maintaining a fairly long whitelist of values.  While this could be scripted based on taginfo data this is still a maintenance problem.

matkoniecz:
Note that script already existed - see https://github.com/gravitystorm/openstreetmap-carto/blob/c17cc64cbf37b96710439b1ca6c64ca73e514883/scripts/shop_values.rb (at that time it whitelisted everything above 100 usages except some explicitly blacklisted entries).

I would also support revisiting this decision (AKA reverting #2415) in addition to stopping rendering `shop=yes` combined with lowering minimal count to 50 - but to do in addition to dropping `shop=yes` rendering.

imagico:
My idea was more of a combined rule - including values that are not on blacklist and either

* sufficient use numbers to indicate it is not just a typo or unqualified choice
* a tag page on the wiki
* support for the specific value in any taginfo registered project

But the question is not really how the rules should look like exactly but if we want to have a dynamic list of values that needs to be updated on a regular basis.

I see this also in relation to #3635 - if a generic shop dot ceases to indicate any shop value and starts indicating a meaningful shop value that would probably somewhat reduce the incentive to add special individual symbols for all kind of shops to separate them from _all the meaningless dots_.

matkoniecz:
> a tag page on the wiki

Not a good filter. I created many tag pages on wiki to document specific tag as a horrible mistake.

matkoniecz:
Though note that potential catch-all removal is a separate project and it is unnecessary to discuss it here.

kocio-pl:
I don't believe such list makes sense if there is no person dedicated to "gardening" this particular problem (not a big task, but needed to be done in a regular fashion for a long time). We had a short list of shops once, when there was much more activity in this project, and even then it was abandoned and I'm afraid we would get back to this fiction.

At this moment I support removing `shop=yes` only, so we take no additional obligation and people still have clearly defined and quite large task (~150.000 uses) to take care of.

matkoniecz:
Yes, certainly dropping `shop=yes` is not implying dropping also catch-all that would be a separate discussion.

BTW, I looked at old script that resulted in some updates (of script, OSM Wiki and mailing list). If someone is interested in shop values it is a good idea to look at them.

imagico:
Note since #2415 the same logic as used for shop=* was also added for office=*.  While it is certainly not necessary to look at all of this (catch-alls and yes-values, shop and office) at once it could help deliberating on these things and trying to find consensus together.

matkoniecz:
I opened https://github.com/gravitystorm/openstreetmap-carto/pull/3718 (certainly not ready, but posted as PR as I made some updates to an old script).

kocio-pl:
Another similar tag to look at is `building=*` (I mean especially `building=yes`). Once in a while we can check if something can be changed in rendering them without causing too much stir in a bowl.

PR #3730, opened 2019-03-25
Stop shop catchall [WIP, do not merge]

matkoniecz:
- [ ] switch script generating valid shop names to Python
- [ ] replace the literal list in SQL code in project.mml with an SQL function call to a function that is in a separate SQL script and is generated by script or edit .mml file by scipt
- [ ] checks support of shop tagging in JOSM, iD, Vespucci - request/PR support for valid shop values and propose validator warnings/fixes for unwanted (note to self: /home/mateusz/Documents/install_moje/OSM software/test_tag_support that resulted in https://josm.openstreetmap.de/ticket/18310 https://josm.openstreetmap.de/ticket/18335 https://josm.openstreetmap.de/ticket/18336 https://josm.openstreetmap.de/ticket/18337 https://josm.openstreetmap.de/ticket/18338 )
- [ ] document unwanted and unneded shop values as unwanted and unneded (in progress, see tagging mailing list thread about ice cream)

----

This is work in progress based on idea floated in https://github.com/gravitystorm/openstreetmap-carto/pull/3718#issuecomment-476166845 by @imagico to drop catch-all for shop types.

> i would be in favor of revisiting the decision of #2415 to render a catchall for shops.

> We currently have 105k different values for shop=* but only ~850 with more than ten uses - which means that there are nearly as many special choice values for shop=* out there as there are cases of shop=yes and many of them are just typos or pointless synonyms for existing established values.

> I am kind of inclined to suggest rendering neither shop=yes nor any special choice values - but that would require creating and maintaining a fairly long whitelist of values. While this could be scripted based on taginfo data this is still a maintenance problem.

For this moment it is solely an updated shop script that

- takes all shop types mapped 100 times or more
- add all shop types that have their own page on the wiki
- discards items from blaklist

The idea is to revert #2415 and go back to rendering only specific shop values.

To avoid not rendering well mapped shops it is necessary to have easy to update list of valid values. This script is supposed to be used for this purpose.

Main change from old one is "add all shop types that have their own page on the wiki" what allows to handle rare shop values that are valid.

It may be a good idea to further reduce minimal count from 100 to something even lower.

-----

posted as PR rather than issue as I have some related code and with PR it is easier to group it with discussion


imagico:
A number of things i would like to see if we decide to go this way:

* not introduce dependency on another scripting language (ruby)
* have a fully automated update process for updating the list (which currently would require creating an SQL function or script editing project.mml)
* not basing it on a hand curated whitelist (which requires continuous work and commitment to maintain and would inevitably come with cultural bias)
* having the flexibility of reasonably easy adaption to other tags (like office=*).

Another thing i wonder is of there are any performance issues with using a list of possibly several hundred different values for filtering.

matkoniecz:
> not introduce dependency on another scripting language (ruby)

Switching to other language would be simple.

> not basing it on a hand curated whitelist (which requires continuous work and commitment to maintain and would inevitably come with cultural bias)

Note that whitelist is used solely to list new entries appearing in the list making easier to catch blacklist candidates.

> have a fully automated update process for updating the list (which currently would require creating an SQL function or script editing project.mml)

SQL in turn would require full planet to run it, right? Querying Taginfo API in SQL may be in theory possible but sounds like a poor idea.

script editing project.mml would be much more reasonable (and planned as the next step)

> having the flexibility of reasonably easy adaption to other tags (like office=*).

This is simple, but separate blacklist (what requires most of the time) would be necessary.

imagico:
> SQL in turn would require full planet to run it, right? Querying Taginfo API in SQL may be in theory possible but sounds like a poor idea.

No, the list of valid shop values would always be generated as a literal list by a script - based on taginfo API calls or other sources.

But right now this list is printed out as a plain list to be copy and pasted into project.mml manually.  This would either need to be scripted  (which is somewhat ugly) or you you could replace the literal list in SQL code in project.mml with an SQL function call to a function that is in a separate SQL script that is re-generated and updated whenever the list is updated - like i do in

https://github.com/imagico/osm-carto-alternative-colors/blob/master/line-widths-generated.sql

matkoniecz:
Yes, that would make a perfect sense. Maybe it would be also possible to avoid repeating the same long list in multiple places.

matkoniecz:
Once #3718 is merged I plan on porting script to Python (existing scripts are in Python) to avoid introducing Ruby as dependency.

In meantime I will try to document some existing shop values (posts on tagging will appear soon).

Adamant36:
Slightly off topic, but @matkoniecz, you think you could open an issue on StreetComplete for something like "what type of building is this" if there isn't one already to deal with the over use of building=yes tag? Id do it, but you to know how StreetComplete works. Whereas, I totally don't.

matkoniecz:
> "what type of building is this"

Already implemented in StreetComplete.

Adamant36:
OK cool. Its odd it never comes up for me when I use it. At least now I know it's there though.

matkoniecz:
> Its odd it never comes up for me when I use it.

Maybe you have no nearby `building=yes`? You can also look in setting for quest list and drag building type higher to give it a higher priority.

Adamant36:
> Maybe you have no nearby building=yes? You can also look in setting for quest list and drag building type higher to give it a higher priority.

Could be. I'm pretty much the only consistent mapper within like a 150 miles of me and I'm pretty good at making sure buildings are tagged correctly. So it just be that there aren't any instances of building=yes where I am. I'll try the sorting thing though. I wasn't aware it was an option. So thanks for the tip.

pnorman:
@matkoniecz this has been listed as WIP for several months. Is it still under work?

matkoniecz:
Whops, I thought that it is waiting for review. I will try to reimplement (or retest) this.

jeisenbe:
@matkoniecz are you going to update this PR?

matkoniecz:
Yes - I know that it is not ready but if someone objects to idea proposed by it, please comment now.

pnorman:
> replace the literal list in SQL code in project.mml with an SQL function call to a function that is in a separate SQL script and is generated by script or edit .mml file by scipt

This would be a significant change from our current policy of not using any functions, and changes the developer experience. If we do go for the route of storing style logic in the DB, I'd rather make another table and join to it.

jeisenbe:
Re: "If we do go for the route of storing style logic in the DB, I'd rather make another table and join to it.” - @matkoniecz is that something you would be interested in trying, or are we uncertain about this idea?

I support switching back to an approved list of tags rather than the current catch-all for any value of shop=*, even though it will increase complexity in the code.

Issue #3706, opened 2019-03-05
healthcare=counselling no icon no text


thetornado76:
At this node https://www.openstreetmap.org/node/1252888814 there is no icon or text.

Please add icon and text for healthcare=counselling similar healthcare=laboratory
![screenshot_20190305-175802_chrome](https://user-images.githubusercontent.com/20454020/53822860-da211900-3f70-11e9-913b-5562d91f53a3.jpg)

matkoniecz:
https://taginfo.openstreetmap.org/tags/healthcare=counselling - used just 37 times so lack of rendering it is not too surprising

From https://github.com/gravitystorm/openstreetmap-carto/blob/4df96c4e4927c3e029b31e34c0cf1be2dda6f637/amenity-points.mss#L716 it seems to not be list of rendered tags

Adamant36:
Maybe it could get a dot at least so the building doesn't look un-tagged or empty (do healthcare objects even have dots? I can't remember).

wilmaed:
> 
> 
> do healthcare objects even have dots? 

yep:
https://www.openstreetmap.org/node/5445761071

Adamant36:
Thanks, I thought they did.

woodmiser:
I had the same problem a couple days ago.  As a workaround, I tagged the place with
amenity=social_facility
social_facility=outreach
social_facility=mental_health
and it gets rendered now.  You might consider whether the place you are trying to tag fits that description, and tag it with that temporarily.
Anyway, I just checked taginfo and it is up to 72 now.
https://taginfo.openstreetmap.org/tags/healthcare=counselling
Also ran an overpass global query, which found 110.  So, hopefully this tag will get the attention it need soon.
Cheers

polarbearing:
> Also ran an overpass global query, which found 110

This is incorrect, you were including the nodes in polygons probably. 

The counselling value has no wiki page and no proposal. I would expect a healthcare value to describe some _therapeutic_ approach, and not counselling. Among the values I found a WeightWatchers office, which certainly does not fit. 

It seems that in indeed 'social facility' would be a better fit in particular in the context of mental health. 

I propose to close the issue.

Adamant36:
[...]

Issue #3609, opened 2018-12-28
healthcare=nurse not rendered at all


deuzeffe:
Hello,

Thx for the work done on rendering for healthcare tag scheme. Very appreciate.

One question: healtcare=nurse is not rendered. Is it intended, forgotten or just a bug ?
For example, here: https://www.openstreetmap.org/#map=19/45.79709/1.13787 (on n°34). On an other hand, maybe I didn't tagged correctly the POI.. Feel free to show me the right way :)

Thx for your answer, whatever it is.

PS : sorry for my bad and straightforward english. As you guess, I'm not a english language native...

kocio-pl:
Hi, thanks for the report!

This tag is not documented currently and has very low usage, so it's unlikely at the moment:

https://wiki.openstreetmap.org/wiki/Key:healthcare

![taghistory 37](https://user-images.githubusercontent.com/5439713/50531019-42ccbf00-0b04-11e9-8a5b-234a14fd5f76.png)

Somebody should complete the wiki and add more such objects to the OSM database.

deuzeffe:
Oh, I see, you're right, of course! 

Mainly used in France (according to https://overpass-turbo.eu/s/EQ1), probably because a tuto about healthcare tagging (https://wiki.openstreetmap.org/wiki/FR:Comment_cartographier_un_(sant%C3%A9) ) to tag nurse's offices. An other wiki page to write ^^

PS: thx for `features` label and adding in the road map ;)

Have a nice 2018's end & 2019's beginning!

kocio-pl:
I will close this issue now for the reasons explained above, but feel free to come back when the usage will be much higher and there will be wiki page for this. Technically it should be easy fix then.

jeisenbe:
This tag is now used 1429 times, after an import or organized activity last year, and contiued usage in the past couple of years. https://taginfo.openstreetmap.org/tags/healthcare=nurse#chronology

<img width="953" alt="Screen Shot 2021-09-18 at 22 28 00" src="https://user-images.githubusercontent.com/42757252/133916538-8de02017-a3fc-4a0c-97c8-5f8617719725.png">

However geographically the usage is quite limited, mostly in France, and a few regions in other countries:
<img width="700" alt="Screen Shot 2021-09-18 at 22 27 36" src="https://user-images.githubusercontent.com/42757252/133916531-3f3fa338-7424-4819-af5b-549e254de3a6.png">
:

[...]

Issue #3606, opened 2018-12-27
healthcare=rehabilitation on areas not rendered


geowas-github:
Areas (closed ways) tagged healthcare=rehabilitation are not rendered.

https://www.openstreetmap.org/way/551730150
https://www.openstreetmap.org/way/644404595

However, there is a health care dot if the tag is on a building. 

https://www.openstreetmap.org/way/127701228

Adamant36:
Where would healthcare=rehabilitation be tag on something as a way thats not a building? Can you provide a location of an example?

geowas-github:
I added examples above. You were to fast ;)

In Germany there are rehabilitation centres which can be as big as hospitals (several buildings, parks, helicopter landing sites, etc. included).

kocio-pl:
We have more problems with some areas rendered only with buildings. Looks like some areas are not imported into database when not using any of the popular tags.

jeisenbe:
Do'h, I just opened a duplicate issue! The problem is that 'healthcare' is not in openstreetmap-carto.lua as a polygon

kocio-pl:
I will close this one now as your ticket is more general.

PR #3731, opened 2019-03-26
Revert rendering of healthcare key

### LINKED ISSUE #3639: Healthcare tags do not render on closed ways without area=yes

jeisenbe:
Reverts #3498
Fixes #3639, #3606 
Related to #3706, #3594, #3609 

Can be reverted after a new version of #3644 is merged and database reloaded (probably some months in the future)

- Remove "healthcare" key from project.mml and amenity-points.mss
- Removes rendering for features tagged with "healthcare" only (most healthcare features, which are also tagged with amenity=hospital, =clinic, =doctors, =pharmacy etc, will continue to render)

### Explanation
Features mapped with the key "healthcare" are not currently loaded into the database as a polygon when mapped as a closed way. Fixing this requires database reload; see #3644 which was closed. 

In the meantime, features mapped as closed ways are not be rendered when tagged with "healthcare" alone. This provides inconsistent mapper feedback, suggesting that features need to be double-tagged with "amenity" or mapped as a node to render.

Therefore, healthcare should not be rendered at this time. Ideally, the next database reload on the OSMF servers can take place sometime this year, and the healthcare tagging can be rendered again at that time.

### DIFF
--- a/amenity-points.mss
+++ b/amenity-points.mss
@@ -655,30 +655,14 @@
   }
 
   [feature = 'amenity_clinic'][zoom >= 17],
-  [feature = 'healthcare_clinic'][zoom >= 17],
-  [feature = 'amenity_doctors'][zoom >= 17],
-  [feature = 'healthcare_centre'][zoom >= 17] {
+  [feature = 'amenity_doctors'][zoom >= 17] {
     marker-file: url('symbols/amenity/doctors.svg');
     marker-fill: @health-color;
     marker-placement: interior;
     marker-clip: false;
   }
 
-  [feature = 'healthcare_doctor'][zoom >= 17] {
-    [zoom >= 17][zoom < 18] {
-      marker-width: 4;
-      marker-line-width: 0;
-    }
-    [zoom >= 18] {
-      marker-file: url('symbols/amenity/doctors.svg');
-    }
-    marker-fill: @health-color;
-    marker-placement: interior;
-    marker-clip: false;
-  }
-
-  [feature = 'amenity_dentist'][zoom >= 17],
-  [feature = 'healthcare_dentist'][zoom >= 17] {
+  [feature = 'amenity_dentist'][zoom >= 17] {
     [zoom >= 17][zoom < 18] {
       marker-width: 4;
       marker-line-width: 0;
@@ -691,8 +675,7 @@
     marker-clip: false;
   }
 
-  [feature = 'amenity_hospital'][zoom >= 15],
-  [feature = 'healthcare_hospital'][zoom >= 15] {
+  [feature = 'amenity_hospital'][zoom >= 15] {
     marker-file: url('symbols/amenity/hospital.svg');
     marker-fill: @health-color;
     marker-placement: interior;
@@ -711,34 +694,6 @@
     marker-fill: @health-color;
     marker-placement: interior;
     marker-clip: false;
-  }
-
-  [feature = 'healthcare_alternative'],
-  [feature = 'healthcare_audiologist'],
-  [feature = 'healthcare_birthing_center'],
-  [feature = 'healthcare_blood_bank'],
-  [feature = 'healthcare_blood_donation'],
-  [feature = 'healthcare_dialysis'],
-  [feature = 'healthcare_laboratory'],
-  [feature = 'healthcare_midwife'],
-  [feature = 'healthcare_occupational_therapist'],
-  [feature = 'healthcare_optometrist'],
-  [feature = 'healthcare_physiotherapist'],
-  [feature = 'healthcare_podiatrist'],
-  [feature = 'healthcare_psychotherapist'],
-  [feature = 'healthcare_rehabilitation'],
-  [feature = 'healthcare_speech_therapist'],
-  [feature = 'healthcare_yes'] {
-    [zoom >= 17] {
-      marker-width: 4;
-      [zoom >= 18] {
-        marker-width: 6;
-      }
-      marker-line-width: 0;
-      marker-placement: interior;
-      marker-clip: false;
-      marker-fill: @health-color;
-    }
   }
 
   [feature = 'amenity_place_of_worship'][zoom >= 16] {
--- a/project.mml
+++ b/project.mml
@@ -126,7 +126,6 @@ Layer:
                                                     'marketplace', 'community_centre', 'social_facility', 'arts_centre', 'parking_space', 'bus_station',
                                                     'fire_station', 'police')
                               OR amenity IN ('parking') AND (tags->'parking' NOT IN ('underground') OR (tags->'parking') IS NULL) THEN amenity ELSE NULL END)) AS amenity,
-              ('healthcare_' || (CASE WHEN tags->'healthcare' IN ('clinic' ,'hospital') THEN tags->'healthcare' ELSE NULL END)) AS healthcare,
               ('landuse_' || (CASE WHEN landuse IN ('quarry', 'vineyard', 'orchard', 'cemetery', 'residential', 'garages', 'meadow', 'grass',
                                                     'allotments', 'forest', 'farmyard', 'farmland', 'greenhouse_horticulture',
                                                     'recreation_ground', 'village_green', 'retail', 'industrial', 'railway', 'commercial',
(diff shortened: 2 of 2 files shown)


Adamant36:
> In the meantime, features mapped as closed ways are not be rendered when tagged with "healthcare" alone. This provides inconsistent mapper feedback, suggesting that features need to be double-tagged with "amenity"

What's wrong with double tagging exactly? Last time I checked both tagging schemes are correct aren't they? More then removing the rendering of one, which kind of pushes people in the direction of over tagging the other, it should figured out which is the better tagging scheme and the none favored one should be phased out. I seem to remember there being a discussion about doing just that somewhere. It seems like a better option then encouraging people to not map healthcare as a closed way just because its not supported here yet.

jeisenbe:
There is no problem with double-tagging clinics, hospitals, and pharmacies.
But a number of the healthcare values have no equivalent. If they are
mapped as a polygon they will not render, right now. We can’t fix this
problem until the database is reloaded

imagico:
Frankly i am not sure what to make of this because i don't understand the reasoning behind the change this reverts.  It to me looks like this was adding rendering of synonyms for existing hospital/dentist etc. tags - which is something we as a principle like to avoid here to support tagging consistency and avoid meaningless proliferation of tags.

In other words:  This change makes sense to me because it reverts a change that does not seem to make sense.  But i suspect i might be missing something.

jeisenbe:
The original PR also added rendering (with red dots and a text label) for a number of healthcare-related features that have no alternate tagging:

```
          [feature = 'healthcare_alternative'],
	  [feature = 'healthcare_audiologist'],
	  [feature = 'healthcare_birthing_center'],
	  [feature = 'healthcare_blood_bank'],
	  [feature = 'healthcare_blood_donation'],
	  [feature = 'healthcare_dialysis'],
	  [feature = 'healthcare_laboratory'],
	  [feature = 'healthcare_midwife'],
	  [feature = 'healthcare_occupational_therapist'],
	  [feature = 'healthcare_optometrist'],
	  [feature = 'healthcare_physiotherapist'],
	  [feature = 'healthcare_podiatrist'],
	  [feature = 'healthcare_psychotherapist'],
	  [feature = 'healthcare_rehabilitation'],
	  [feature = 'healthcare_speech_therapist'],
	  [feature = 'healthcare_yes'] {
```

I think adding `healthcare=yes` was questionable, but I would support rendering the other features in the above list with a generic dot and name label, when it become possible.

However, because these features can be tagged on polygon geometires (closed ways and multipolygon relations), and at this time we cannot render these without a database reload, I have submitted this PR. 

I believe this is reason enough to merge the change for the time being; we don't have to discuss the double-tagging issue at this time.

jajajaneeneenee:
I personally think this is not a very good idea. Why should mappers be frustrated who already used the new healthcare tagging scheme for POIs, which are not represented by the old scheme (which was **one** important reason for the newer, wider scheme, which is more powerful)? I mean those who tagged nodes, with healthcare=* (without amenity) which are rendered already now (except healthcare=counselling, which is in the wiki since March 2019, and I personally would like to render, too, because I think, it's an important part of a healthcare system). And this will disapear after this change for some month, for a year? Or perhaps longer? Only to avoid discussions and questions when it was used to map closed ways?

This could be made clearer in the Wiki ... and that it will take some month to fix this. If you remove the rendering for nodes (will have an effect only to those which are tagged with "healthcare" only, I know), you also will have to explain this in the wiki, I think ... and this may cause also some discussions and will take some time until every mapper did read it  ... so what is the real advantage? Or is there any **advantage** for the mappers at all? Or is the only reason to remove an temporary inconsistency in the rendering – seen from a developer viewpoint? I don't get it completely ...

And I am not familiar enough with the problems of the database reload: but why don't put more energy in a quicker database reload (a PR for that?) to fix this problem completely much faster than some month or a year? But as I mentioned, this may be out of the scope of my knowledge (how difficult this may be). Only a question I had for myself ...

And I think, the new healthcare scheme makes a lot of sense an is quite well-thought-out ... it's a good advancement of the old "amenity scheme". And now we are in the transition phase with double tagging (no problem I think)  ...but taggers should be encouraged to use (also) the new scheme. And the PR would perhaps cause the opposite. That's my opinion, or better: my feeling ...

(Quick view in the [taginfo of "healthcare"](https://taginfo.openstreetmap.org/keys/?key=healthcare#overview) – 2019-04-02: nodes=72 791, ways=36 004, so the ratio is 2:1 – very often combinded with amenity, of course – 77.56% – but there should be > 20.000 without amenity and perhaps around 14000 nodes WITHOUT amenity based on the 2:1 ratio – I didn't check it completely = potential of 14000 frustrations? I know ... that only an abstract and rough calculation ...)

Only wanted to add some thoughts ... and I'm coming more from the mapper side (as you may already have noticed), but I also have some developer backgrounds since the 80's/90's ... so I decided to write this comment.

Adamant36:
> why don't put more energy in a quicker database reload (a PR for that?) to fix this problem completely much faster than some month or a year?

Unfortunately, it's not that easy. From my understanding it's not really done alone, and gets dealt with when new LTS versions are Ubuntu are deployed. Along with other things, like newer stable versions the back end site stuff. Which usually happens once a year, usually in October if I remember correctly. There was talk about doing it every six months, but there just enough stuff to implement on that cycle most of the time to make it worth doing that frequently. 

My personal opinion on this is that it might have been a good thing to do a while back, but now that it's almost the middle of the year and it will only be like 6 months until the reload, I don't it's worth the temporary gain to implement. Given the other problems that will come along with it. Like confusing people as to why rendering a perfectly valid tagging scheme was removed, just to be implemented again in a few months. As @jajajaneeneenee says also, tagging both along side each other is helping it to get used more widely. Not rendering it might screw that up and cause the older tagging scheme to gain ground it shouldn't. Plus, the whole thing is a really minor issue in the first place in my opinion. It's definitely not on the top of the list of rendering problems. I've really only seen a few complaints about it. There will probably be more issues opened if its removed then there currently are about whatever problems it causes in the short term.

imagico:
> I believe this is reason enough to merge the change for the time being; we don't have to discuss the double-tagging issue at this time.

Fine by me - i just wanted to make sure i did not miss any non-obvious reason for introducing rendering of synonyms for established tags here.

Regarding database reload - this does not depend on any external software version conditions - though a system upgrade would of course be a logical choice for doing a database reload.  Anyone who is interested in making this happen sooner can help doing that by working on implementing this change - to be maintained in a separate branch until it can be integrated as discussed in #3611.

And keep in mind tagging discussion is doubly off-topic here (in this PR and in this repository in general).

kocio-pl:
This problem has no good solution for me, unfortunately. I believe if there is now someone to take care of the full upgrade, we could merge this for now, otherwise it's better to leave the current state to not create additional confusion and code mangling.

matthijsmelissen:
>  the described problem is definitely something that should have prevented merging of #3498

+1

imagico:
@kocio-pl - i don't quite understand your statement, could you re-formulate to make it clearer?

kocio-pl:
If somebody **plans** to take care of the code to be properly rendered in all te cases (in few months probably), we could revert now. If nobody is ready **now** to take care of that process, I think we should not revert it. 

I hope this is more clear now.

imagico:
That's clearer now, thanks.

I think preparing a change reverting this PR later would not pose significant technical difficulty.  But i also think asking anyone here for a long term assurance that they will prepare a change at an undefined time in the future is not something we can or should ask in a volunteer project.

kocio-pl:
Nobody is forced to do anything. I just vote with my trust on somebody willing to provide a solid solution, because it would overweight other problems and risks I'm still afraid of (expressed by @jajajaneeneenee and @Adamant36).

pnorman:
> It to me looks like this was adding rendering of synonyms for existing hospital/dentist etc. tags - which is something we as a principle like to avoid here to support tagging consistency and avoid meaningless proliferation of tags.

+1

Leaving aside the synonyms for existing tags, the new ones don't render properly, so +1 on this PR.

I resolved the conflicts and merged it.

HolgerJeromin:
Was this part overlooked in the revert? 

https://github.com/gravitystorm/openstreetmap-carto/blob/635dae6e78e44a059a45a2d23485b992f763605b/amenity-points.mss#L2521-L2540

jeisenbe:
Yes, that was my mistake. These lines are not needed at the moment.

Issue #3634, opened 2019-01-10
Limit natural labels rendering to z4+


kocio-pl:
There are several natural tags that can be rendered pretty early, like bays, capes or islands. No island shows its label earlier than z4, so it makes sense to limit also other natural objects to this. Currently Baffin Bay stands out (it's the only label on z2+), which makes it a practical problem to solve:

![screenshot_2019-01-10 openstreetmap](https://user-images.githubusercontent.com/5439713/50999864-a6ef7a00-152b-11e9-94b1-8ae6b85e6e47.png)

Please be aware that it's highly controversial problem discussed (currently there is a try to define peninsulas: https://github.com/gravitystorm/openstreetmap-carto/issues/788#issuecomment-453248593) and I don't think we can resolve it in the single ticket, because the root causes are much deeper. I just like to make an upper limit as a fix to create some balance for labels.

If you are interested in more details why it's so heated debate, here is a diary entry from @imagico who presents this problem as the opponent of rendering such areas at all:

https://www.openstreetmap.org/user/imagico/diary/47432

imagico:
No amount of zoom level threshold tuning will solve this.

I stated my position on this in https://github.com/gravitystorm/openstreetmap-carto/pull/2700#issuecomment-317236118

This would mean removing way_area logic from all features that are rendered label only.  That AFAIK is currently:

* natural=bay
* natural=strait
* shop=mall
* natural=cape

tourism=attraction was removed in #3603.  shop=mall is a legacy case that is not really that significant practically (see #2702).  The other three were all added during the last year.

Note it is not really a heated debate, it is just a point where this style has during the past year moved most strongly from mapper support towards active mapper steering - maybe unintentionally originally but the effect is still the same.  Adjusting zoom level thresholds would just tune the mapper steering into a different direction, it would not remove it.

kocio-pl:
> No amount of zoom level threshold tuning will solve this.

Please follow the rationale with the scope clearly defined (what I mean by "this" which is to be solved here and what is "that" which is not to be solved here).

imagico:
Then i rephrase:  This issue is not something i consider worth solving because it would only (poorly) hide the real problem, which is using way_area for making prominent rendering decisions on features without giving the mapper feedback on the geometry in question and thereby creating counterproductive mapping incentives (that is using the polygon geometry essentially to manually specify way_area).

kocio-pl:
> Note it is not really a heated debate

[...]

PR #3732, opened 2019-03-27
Restrict rendering of natural=cape to nodes, change font to standard

jeisenbe:
Partial fix for #3634

- Remove rendering of natural=cape polygons (Nodes will continue to be rendered as at present)
- Change cape name label text to use standard font, rather than landcover font

### Explanation:
- [According to the wiki](https://wiki.openstreetmap.org/wiki/Tag:natural%3Dcape), a cape is the extreme point of land that projects out into a body of water, such as a sea or lake. 
- [Taginfo confirms](https://taginfo.openstreetmap.org/tags/?key=natural&value=cape) that 12 ,747 capes are mapped as nodes, and only 148 as ways, 355 as relations.
- The recently approved tag "natural=peninsula" is available to tag areas of land that are mainly surrounded by water. From the [wiki](https://wiki.openstreetmap.org/wiki/Tag:natural%3Dpeninsula):

> "The tag natural=cape refers to a coastal extreme point, that is the tip of a land area that projects into water. In contrast, natural=peninsula refers to a land area that projects into water and that is nearly surrounded by water."

- Currently the natural=cape name label is rendered with an italic landcover-style font, which is the same as is used for islands. The font is changed to the standard font to be more appropriate for a point feature. 

**Kolkas rags, Latvia** - mapped as node
https://www.openstreetmap.org/#map=14/57.7488/22.6323
z14 Current
z14-kolkas-rags-cape-before

z14 After
z14-kolkas-rags-after

**Castle Neck** - mapped as a closed way
https://www.openstreetmap.org/#map=15/42.6744/-70.7476
z15 Current
z15-castle-neck-cape-before

z15 After
z15-castle-neck-cape-after

### DIFF
--- a/amenity-points.mss
+++ b/amenity-points.mss
@@ -1790,8 +1790,6 @@
 /* Note that .text is also used in water.mss */
 .text-low-zoom[zoom < 10],
 .text[zoom >= 10] {
-  [feature = 'natural_cape'][zoom >= 4][way_pixels > 3000][way_pixels < 800000],
-  [feature = 'natural_cape'][zoom >= 16][way_pixels < 800000],
   [feature = 'place_island'][zoom >= 4][way_pixels > 3000][way_pixels < 800000],
   [feature = 'place_island'][zoom >= 16][way_pixels < 800000],
   [feature = 'place_islet'][zoom >= 11][way_pixels > 3000][way_pixels < 800000],
@@ -2065,13 +2063,10 @@
   [feature = 'natural_cape'][zoom >= 14] {
     text-name: "[name]";
     text-fill: #000;
-    text-size: @landcover-font-size;
-    text-wrap-width: @landcover-wrap-width-size;
-    text-line-spacing: @landcover-line-spacing-size;
-    text-size: @landcover-font-size-big;
-    text-wrap-width: @landcover-wrap-width-size-big;
-    text-line-spacing: @landcover-line-spacing-size-big;
-    text-face-name: @oblique-fonts;
+    text-size: @standard-font-size;
+    text-wrap-width: @standard-wrap-width;
+    text-line-spacing: @standard-line-spacing-size;
+    text-face-name: @standard-font;
     text-halo-radius: @standard-halo-radius;
     text-halo-fill: @standard-halo-fill;
     text-placement: interior;
--- a/project.mml
+++ b/project.mml
@@ -2071,7 +2071,7 @@ Layer:
               'landuse_' || CASE WHEN landuse IN ('forest', 'military', 'farmland') THEN landuse ELSE NULL END,
               'military_' || CASE WHEN military IN ('danger_area') THEN military ELSE NULL END,
               'natural_' || CASE WHEN "natural" IN ('wood', 'glacier', 'sand', 'scree', 'shingle', 'bare_rock',
-                                                    'water', 'bay', 'strait', 'cape') THEN "natural" ELSE NULL END,
+                                                    'water', 'bay', 'strait') THEN "natural" ELSE NULL END,
               'place_' || CASE WHEN place IN ('island') THEN place ELSE NULL END,
               'boundary_' || CASE WHEN (boundary = 'protected_area' AND tags->'protect_class' = '24') THEN 'aboriginal_lands'
                                   WHEN boundary IN ('aboriginal_lands', 'national_park')
@@ -2084,7 +2084,7 @@ Layer:
           FROM planet_osm_polygon
           WHERE (landuse IN ('forest', 'military', 'farmland')
               OR military IN ('danger_area')
-              OR "natural" IN ('wood', 'glacier', 'sand', 'scree', 'shingle', 'bare_rock', 'water', 'bay', 'strait', 'cape')
+              OR "natural" IN ('wood', 'glacier', 'sand', 'scree', 'shingle', 'bare_rock', 'water', 'bay', 'strait')
               OR "place" IN ('island')
               OR boundary IN ('aboriginal_lands', 'national_park')
               OR (boundary = 'protected_area' AND tags->'protect_class' IN ('1','1a','1b','2','3','4','5','6','7','24','97','98','99'))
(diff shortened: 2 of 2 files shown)


Adamant36:
As the person who originally did the PR for rendering capes, my opinion is that it might have been worth waiting until the discussion in #3634 was in a semi-complete state and it was agreed that this is an actual solution. As it is, the agreements put forth in #3634 that support this PR are rather weak, originally only put forward by a few people (or really one person), and I don't see much support of them. It might a solution after the discussion is concluded though. 

Also, there was a reason why capes were rendered like islands in the original issue/PR. I don't remember what the reason was, but maybe you should read it over and state why rendering them that way is problematic in the PR message.

jeisenbe:
> As the person who originally did the PR for rendering capes

I appreciate the work you did in the original PR to get these rendered in #3452 - especially all the tests you did to confirm that z14 was the right zoom level to show capes. That was a lot of work, thank!

Unfortunately, I missed the commit that changed the PR to include capes mapped as areas. I was under the misimpression that the plan was to only render nodes, so I didn't say anything at the time. 

> There was a reason why capes were rendered like islands in the original issue/PR. ...

I read through all of the discussion in #3452 prior to making this PR, and it was not clear why the maintainer requested that they be rendered like islands. The wiki description and taginfo at the time were clear that nodes were used for over 98% of capes (as is the case now). I can only speculate that perhaps there was some confusion between cape and peninsula? 

However, now that natural=peninsula is an approved feature, this is the correct way to map an area of land that is mostly surrounded by water. See: https://wiki.openstreetmap.org/wiki/Tag:natural%3Dcape and https://wiki.openstreetmap.org/wiki/Tag:natural%3Dpeninsula: 

_"The tag natural=cape refers to a coastal extreme point, that is the tip of a land area that projects into water. In contrast, natural=peninsula refers to a land area that projects into water and that is nearly surrounded by water."_

This was clear during the proposal process for natural=peninsula and when it was discussed on the tagging list: https://wiki.openstreetmap.org/w/index.php?oldid=1813152 includes the text above.

So natural=cape on a natural=peninsula is like a natural=peak on a mountain range or ridge; it's a point on a larger feature.

Adamant36:
> Unfortunately, I missed the commit that changed the PR to include capes mapped as areas. I was under the misimpression that the plan was to only render nodes, so I didn't say anything at the time.

That explains it. I seem to remember you being involved in another natural feature being rendered at that time. So I was wondering why you didn't speak up about it back then. 

I don't have a problem with removing area rendering per say, but I don't see how point rendering would necessarily be anymore accurate. Since it's impossible to really tell where the center of a cape is. As apposed to something like peaks that have clear end points at the top. So, it doesn't really solve the problem @imagico has with improper mapping. Which is why I why I think he wants it reverted completely. 

Ultimately,  I think there's really only two options here. 1. accept that some things will be arbitrarily sometimes due to their "fuzzyness," but its still worth it for name rendering 2. Remove any none ground verifiable things from rendering. Including retail/residential/etc landuse and leisure=park. Otherwise, if rendering is removed just for these tags, it will be majorly slanted and not actually deal with the underlining problems of encouraging mapping things that aren't verifiable. 

Personally, id love to see residential/retail/commercial landuse not be rendered. All of them are pretty much worthless due to people mapping them wrong to make the map look pretty. It usually comes at the cost of mapping other more important things also. If there's going to be a stand about this though, it should apply outside of just the natural=* domain. That's just my opinion. 

Personally

imagico:
Note while i would have added rendering of natural=cape for nodes only - based on mostly the same reasons you provided - i have not specifically objected to the idea of rendering polygon features tagged natural=cape.  I just objected to taking the polygon way_area into account for rendering decisions.

I think it is unnecessary to render polygons tagged natural=cape because of the low volume of use and it would support mappers in consistent mapping not doing so.  But i think it is important not to distract from the important part of not making rendering decisions based on way_area with the minor matter of nodes only vs. nodes or polygons.

kocio-pl:
> I just objected to taking the polygon way_area into account for rendering decisions.

Hm, I don't really understand why. For me it's quite natural and efficient tool to give the impression of the size f the object (and it works also in practice in this style). Could you explain it a bit?

(This is just a general question, which is important to me while designing this style, not something related to this change in particular.)

imagico:
I explained this on many occasions -  way_area is derived from the polygon geometry.  If you use that for making rendering decisions without giving visual feedback in the map on the geometry you create an incentive to draw non-verifiable abstract geometries as a method to manually specify the way_area producing the subjectively desired rendering result.

In this specific case i don't even think a cape has a specifiable, let alone verifiable size.  I couldn't specify such for any of the following major capes even if i wanted to:

https://www.openstreetmap.org/node/32532727
https://www.openstreetmap.org/node/2510985983
https://www.openstreetmap.org/node/2098928265
https://www.openstreetmap.org/node/2696775247

kocio-pl:
I see, so it's not rwally way_area problem. I would however ask how can one verify position of these 4 nodes for example? To make a problem simpler for a start - why Kap Morris Jesup is on the coastline and three others are not?

jeisenbe:
> why Kap Morris Jesup is on the coastline and three others are not?

I believe capes should be mapped at the extreme point of land, which would be the node on the coastline where it extends farthest into the body of water. This is the same idea with mountain peaks; you map the highest point. However, it doesn't make a large difference if the node is not exactly on the coastline. 

Similarly, many peaks are not exactly on the highest point of the mountain, but database users can make minor adjustments if needed when processing peaks.

jeisenbe:
Now that natural=peninsula is an approved feature, this is the better way to map an area of land that is mostly surrounded by water. Many of the features that are currently mapped as a closed way or multipolygon and tagged with natural=cape can be retagged to natural=peninsula
https://wiki.openstreetmap.org/wiki/Tag:natural%3Dcape and https://wiki.openstreetmap.org/wiki/Tag:natural%3Dpeninsula:

"The tag natural=cape refers to a coastal extreme point, that is the tip of a land area that projects into water. In contrast, natural=peninsula refers to a land area that projects into water and that is nearly surrounded by water."

imagico:
> I would however ask how can one verify position of these 4 nodes for example?

_Cabo da Roca_ - westmost point of continental Europe
_Cape Agulhas_ - southmost point of Africa
_Cape Comorin_ - southmost point of mainland India
_Cape Morris Jesup_ - northmost point of mainland Greenland

Node location relative to the coastline is not really that meaningful since the cape and the coastline are not necessarily mapped together and the coastline in neither of these cases is particularly accurate (South Africa coastline in particular is largely imported and systematically off)  Also people will often locate the cape not at the high water mark on the beach or scree at the shoreline but at the out-most part of the cliff or other prominent feature - like in case of Cabo da Roca.

Apart from peaks and saddles capes are probably among the most easily and most precisely localizable physical geography features we have.

kocio-pl:
If I understand you right, the location of capes is fuzzy (correct me if you didn't mind it). But still I see no answer for my question - **how** can one verify cape node location? The answer was **where** they are, but this exactly is a claim, and I look for verification of that claim. What source did you use for this claim?

I guess only in the case of Cape Agulhas there is an artwork that - if it contains inscription with this name - can be taken as the location of a node. But what about others?

imagico:
Trying to address the question in a compact form to not too much venture outside the scope here - this is largely a discussion on verifiability in general.

You need to distinguish between the verifiability of names and the verifiability of coordinates.  Using one of the examples given - verifying the location of the westmost point of continental Europe is one thing, verifying that it is called the _Cabo da Roca_ and verifying that it is this point that is called this way and not a different one is another one.  Both of these work essentially like most other physical geography features - like for example a natural=peak.  Determining coordinates of anything is always subject to variance of different independent determinations - sometimes more, sometimes less, depending on the situation and the type of feature.  Determining the name can work either through physical manifestations like signs (which don't necessarily have to be directly at location obviously but can have the form of something like a guidepost some distance away) or through evaluation of local knowledge of people (which is tricky and more work intensive, especially if you are not a local yourself).

Issue #3707, opened 2019-03-05
Render beaches and rocks in the tidal range differently


SelfishSeahorse:
It would be useful if `natural=beach` and `natural=bare_rock` in the tidal range, i.e., below `natural=coastline` would be rendered differently, e.g. by overlaying them with the `natural=wetland` pattern:

![rendering-area-beach2](https://user-images.githubusercontent.com/14875493/53822620-65e67580-3f70-11e9-98e3-01496dd7e177.png)

imagico:
Depends on #1982.  See also #1473, #1547.

Note rendering of beaches with a surface tag is already designed to look differently over water and over land - see #1990.

jeisenbe:
This issue was partially fixed by PR #3738 - now beaches and shoals with a surface tag should have a pattern that renders over the ocean water, but the fill color is covered by the water.

The remaining issue is with `natural=bare_rock` which now is covered completely by the marine water, outside of the coastline. While natural=bare_rock has a pattern, this pattern is not shown above the water fill, but rendered in the main `landcover` layer. 

This can be easily changed if this is considered worthwhile.

jeisenbe:
@SelfishSeahorse - I have code to test which moves the bare_rock pattern above water areas, but I need some examples of where this would make a difference. Can you provide a link to a place outside of the coastline and a place in a lake or river where there is bare_rock?

SelfishSeahorse:
@jeisenbe wrote:

> This issue was partially fixed by PR #3738 - now beaches and shoals with a surface tag should have a pattern that renders over the ocean water, but the fill color is covered by the water.

Very nice! Thanks a lot for your work!

@jeisenbe wrote:

> @SelfishSeahorse - I have code to test which moves the bare_rock pattern above water areas, but I need some examples of where this would make a difference. Can you provide a link to a place outside of the coastline and a place in a lake or river where there is bare_rock?

Please excuse my late reply. Here's a large area of bare rock that is only visible at low tide (i've just added the missing `tidal=yes`, but i guess that it's not needed):

https://www.openstreetmap.org/way/660708388

Unfortunately i don't have an example of bare rock in a lake or river. I don't even know how to map and tag elements in a lake or river only visible at baseflow. (How is baseflow defined on OSM?)

georgek:
[...]

PR #3738, opened 2019-04-06
Reorder layers to render ocean and water areas above water lines and landcover

jeisenbe:
**Fixes** #426, #1473, #1547 and #2609
Related to issues #1781, #2013, #2025, #2688, #3707 and PR #3670 as a necessary step in the solution.
Continuation of PR #3694
Partially reverts #3189

- Move ocean and water above landcover
- Icesheet polygons are moved before water lines, which are now before water-areas, and ocean polygons are after these. Landcover is now first.
- Also, springs are moved above water-barriers, marinas are moved below water-barriers, and `necountries` is moved to immediately before admin-low-zoom
- Natural=mud is rendered in the `landcover-area-symbols` layer, above all the water bodies

### Explanation
- This PR is the first application of the switch to water (ocean/marine) polygons instead of land polygons. The layering changes follow the order used in the `alternative-colors` stylesheet by @imagico

- By moving the ocean and water areas above the water lines, it is now possible to use different colors for linear water features (such as rivers and streams), seas, lakes and rivers.

- The landcover fill color is now rendered below all types of water. This means that natural areas and areas with `landuse` and `amenity` which extend over water (such as golf courses, parks, schools etc) will be rendered properly. This solves #1547, #2013 and #3707 by not rendering the fill color for wetlands, rock, and beaches over the water, but the pattern will still show.

- Marinas are moved before water barriers because the barriers should be shown over the area outline, as is done for tourism areas and protected areas on land. 

- Springs are moved up above water features so that they will render properly when under a stream or river. The icon is the same as that introduced in PR #3189 but the layering changes are reverted.

- `necountries`, the z1 to z3 country borders, sourced from Natural Earth, are moved immediately before the `admin-low-zoom` layer. This makes sure that the borders are rendered over the water areas at all zoom levels.

- The semi-transparent mud color fill is moved from the main landcover layer to  landcover-area-symbols, which remains above all of the water areas, so that it will render properly over all water areas (If it were not changed, then all areas of mud over seas and water would render as they currently do over natural=water (only the wetland pattern would show); this is also a reasonable option)

In the future it will be possible to solve #2013 and #2025 by adding a fill color for wetland=reedbed and wetland=mangroves. These do not have a fill color currently, because they often grow over water areas. But after this PR the fill color can be rendered below the water, so it will only show on the land. 

### Before
z10 Zeeland - current 
https://openstreetmap.org/#map=10/51.5079/3.6996
(mud is over the sea in the south, but under the natural=water in the north)
z10-zeeland-current
z12 - current
z12-zeeland-before

z9 North Wales - current 
https://openstreetmap.org/#map=9/53.2258/-3.6090
(the sand banks, covered at low tide, show up the same as beaches)
z9-north-wales-current
z13
z13-wales-salisbury-current

z14 Newborough forest - current 
https://openstreetmap.org/#map=14/53.1682/-4.3945
 - the spring is under the stream, the marsh is over the sea
z14-newborough-current

### After
z10 Zeeland - after 
(mud areas are the same over the sea and over natural=water)
z10-zeeland-after-mud-over
z12 - after
z12-zeeland-after-mud-over

z9 North Wales after 
(sand banks are covered, but the pattern still shows the location)
z9-north-wales-after
z13 after
z13-north-wales-after-salisburybanks

z14 Newborough after - the spring is now over the stream, and the part of the marsh outside of the coastline is distinct
z14-newborough-after-spring-over-marsh-under

### DIFF
--- a/landcover.mss
+++ b/landcover.mss
@@ -82,17 +82,6 @@
   ::upper-mid-zoom[zoom >= 12][zoom < 13],
   ::high-zoom[zoom >= 13] {
 
-  ::first {
-    [feature = 'wetland_mud'],
-    [feature = 'wetland_tidalflat'] {
-      [zoom >= 5] {
-        polygon-fill: @mud;
-        [way_pixels >= 4]  { polygon-gamma: 0.75; }
-        [way_pixels >= 64] { polygon-gamma: 0.3;  }
-      }
-    }
-  }
-
   [feature = 'leisure_swimming_pool'][zoom >= 14] {
     polygon-fill: @water-color;
     [zoom >= 17] {
@@ -718,6 +707,18 @@
 }
 
 #landcover-area-symbols {
+
+  ::first {
+    [natural = 'mud'],
+    [int_wetland = 'tidalflat'] {
+      [zoom >= 9] {
+        polygon-fill: @mud;
+        [way_pixels >= 4]  { polygon-gamma: 0.75; }
+        [way_pixels >= 64] { polygon-gamma: 0.3;  }
+      }
+    }
+  }
+
   [natural = 'sand'][zoom >= 5] {
     polygon-pattern-file: url('symbols/beach.png');
     polygon-pattern-alignment: global;
--- a/project.mml
+++ b/project.mml
@@ -281,7 +281,8 @@ Layer:
                     END
                 END
               END AS int_wetland,
-            tags->'leaf_type' AS leaf_type
+            tags->'leaf_type' AS leaf_type,
+            way_area/NULLIF(POW(!scale_denominator!*0.001*0.28,2),0) AS way_pixels
           FROM planet_osm_polygon
           WHERE ("natural" IN ('marsh', 'mud', 'wetland', 'wood', 'beach', 'shoal', 'reef', 'scrub', 'sand') OR landuse = 'forest')
             AND building IS NULL
--- a/project.mml
+++ b/project.mml
@@ -54,33 +54,6 @@ Stylesheet:
   - admin.mss
   - addressing.mss
 Layer:
-  - id: ocean-lz
-    geometry: polygon
-    class: ocean
-    <<: *extents
-    Datasource:
-      file: data/simplified-water-polygons-split-3857/simplified_water_polygons.shp
-      type: shape
-    properties:
-      maxzoom: 9
-  - id: ocean
-    geometry: polygon
-    class: ocean
-    <<: *extents
-    Datasource:
-      file: data/water-polygons-split-3857/water_polygons.shp
-      type: shape
-    properties:
-      minzoom: 10
-  - id: necountries
-    geometry: linestring
-    <<: *extents84
-    Datasource:
-      file: data/ne_110m_admin_0_boundary_lines_land/ne_110m_admin_0_boundary_lines_land.shp
-      type: shape
-    properties:
-      minzoom: 1
-      maxzoom: 3
   - id: landcover-low-zoom
     geometry: polygon
     <<: *extents
@@ -176,6 +149,14 @@ Layer:
         ) AS landcover_line
     properties:
       minzoom: 14
+  - id: icesheet-poly
+    geometry: polygon
+    <<: *extents
+    Datasource:
+      file: data/antarctica-icesheet-polygons-3857/icesheet_polygons.shp
+      type: shape
+    properties:
+      minzoom: 5
   - id: water-lines-casing
     geometry: linestring
     <<: *extents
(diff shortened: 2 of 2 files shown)


imagico:
A few remarks here:

* mud/tidalflat was already not rendered as intended originally before all other landcovers since the landcover color fading was introduced.  This change does not significantly worsen the situation (and fixes #1547) but it also does not solve the fundamental problem of the transparency leading to undesirable color mixing.  This could be solved by rendering mud/tidalflat twice - once completely opaque before the water layers but above the landcover layers (or as last element of the landcover layer via attachment) and once after the water layers with transparency.
* i don't like that `natural=beach` and `natural=sand` are shown identically when overlapping water.  Removing the pattern from sand would solve that.  natural=sand in general does not make much sense overlapping water - either `natural=beach/shoal/reef` or `wetland=tidalflat` with `surface=sand` would always be a more accurate mapping.  This is not something this PR should solve but it is an issue this would open.
* #3707 is not fully fixed by this since `natural=bare_rock` is not rendered with an overlay pattern.  But i am not so sure if it is actually desirable to change that for the same reason as with `natural=sand`, i.e. the possibility this could lead to abuse of `natural=bare_rock` for subsea mapping (i.e. `natural=shoal/reef`).  This would need to be discussed separately.
* there are some features from the landcover layer that would afterwards need to be considered how to suitably render over water, in particular things like `amenity=grave_yard` and `leisure=pitch`.

jeisenbe:
I should also mention that one side-effect of this layering change is that landcover symbols are now above streams. It's a subtle difference, but can be noticed when streams are in wetlands, Eg:
z16 bogs before
z16-bog-before
after
z16-streams-bog-wales-after

z15 woods before
z15-pen-lan-wood-before
after
z15-Pen-lan-wood-after

And here's an example of how the marine relayering would work, in theory. (I haven't found any real examples of a marina outline that is crossed by a weir, dam or lock gate, so this is unlikely to occur)

Marina test area, before z17 
(marina outline is below the pier and building, but above the dam, weir and lock gate)
z17-marina-test-before
after
z17-marina-test-after

jeisenbe:
> mud/tidalflat was already not rendered as intended originally before all other landcovers since the landcover color fading was introduced. ... This could be solved by rendering mud/tidalflat twice...

I'm in favor of removing the low-zoom color fading in #3670, which should be possible to implement soon after this PR is merged, so I don't think we need to do the double rendering yet. Were #3670 to be rejected then I would submit another PR to fix the mud rendering, so I think I doesn't need to be addressed in this PR.

> ... natural=beach and natural=sand are shown identically when overlapping water. 

I believe this is true of natural=beach (and natural=shoal) with surface=sand. If the surface is not specified, then there is no pattern overlay. The coarser surfaces (eg pebbles, stones, shells) have a slightly different pattern. But as mentioned, there should be an issue opened about this problem. (Another possible solution would be to render a sand pattern in the `landcover` layer, so that it is not shown over water, similar to the rendering for natural=bare_rock:)

> #3707 is not fully fixed by this since natural=bare_rock is not rendered with an overlay pattern.
I hadn't realized this. I see now that the bare_rock, shingle and scree patterns are rendered in `landcover`, so they will not be shown over water. Perhaps there needs to be some tagging discussion first. I will change the PR description.

> there are some features from the landcover layer that would afterwards need to be considered how to suitably render over water, in particular things like amenity=grave_yard and leisure=pitch.

I can imagine that leisure=pitch might be used for water sports. But I'm surprised to learn that there are partially-submerged graveyards?! Perhaps the pattern could be moved to the `landcover-area-symbol`. Should this be considered in this PR, or better left till later?

imagico:
I am sorry for not being clear - the mud rendering has nothing in substance to do with the fading, it is just that the introduction of fading broke the original rendering of mud before all other landcovers.

> I believe this is true of natural=beach (and natural=shoal) with surface=sand.

Correct.

> I can imagine that leisure=pitch might be used for water sports. But I'm surprised to learn that there are partially-submerged graveyards?! Perhaps the pattern could be moved to the landcover-area-symbol. Should this be considered in this PR, or better left till later?

That's for later.  Just wanted to point out this being a thing to follow up on.

What i had in mind is for example dedicated sea burial sites like:

https://www.openstreetmap.org/way/48980133

kocio-pl:
It's only the visual impression (I didn't look at the code), but I'm very happy to see all these changes. The sand pattern on the water with orange name looks a bit strange, but it makes sense if this is a sand covered with water.

jeisenbe:
> removing the pattern from natural=sand to not render it identical to natural=beach + surface=sand when overlapping water.

It would also be possible to render the sand pattern in the
`landcover` overlay to get the same result. But I understand that you
think removing the sand pattern is a good thing, even on land?

imagico:
I don't see any gain from the sand pattern in its current form for the map (there is no information transported in it while it generates a lot of noise) but i would be open to any rendering of sand as long as it makes sure it is well distinguishable from beaches both with and without a surface tag.

Issue #3250, opened 2018-05-28
Why is area-barriers rendering so different from line-barriers rendering?


sommerluk:
Our current `project.mml` has very different SQL queries for the layers with id “area-barriers” and “line-barriers”:

| area-barriers | line-barriers|
| ------------- |-------------|
| Has not `historic=citywalls`      | Has `historic=citywalls`  |
| Has everything with a `barrier` key      | Has only elements with a `barrier` key  who’s value is whitelisted | 

Therefore, closed ways with barrier=block (which is not in our whitelist) without further tags are not rendered like https://www.openstreetmap.org/way/350624000#map=19/43.63777/10.42239 shows. But if combined with another tag that is recognized as are, like landuse=grass, the barrier is rendered like https://www.openstreetmap.org/way/337388805#map=19/51.22629/6.65681 shows. I think that’s not consistent…

dieterdreist:
sent from a phone

> On 28. May 2018, at 20:50, Lukas Sommer <notifications@github.com> wrote:
> 
> Our current project.mml has very different SQL queries for the layers with id “area-barriers” and “line-barriers”:
> 
> area-barriers	line-barriers
> Has not historic=citywalls	Has historic=citywalls
> Has everything with a barrier key	Has only elements with a barrier key who’s value is whitelisted


just in case it is not clear to everyone, there is also another tag for citywalls, barrier=city_wall

jeisenbe:
Closed by PR #3745

PR #3745, opened 2019-04-12
Render only specific values of barrier and historic=citywalls on polygons

### LINKED ISSUE #3742: Barriers on areas are rendered for all values including barrier=no
### Expected behavior
- Barrier tags on linear ways and closed ways should render the same

### Actual behavior
- `line-barrier` only selects certain values of barrier:
        `'chain', 'city_wall', 'embankment', 'ditch', 'fence', 'guard_rail', 'handrail', 'hedge', 'kerb', 'retaining_wall', 'wall'`
- `area-barrier` selects the entire key barrier, so all values of "barrier" are rendered the same,  
Even barrier=no, misspellings and undocumented tags are rendered with a 0.4 pixel width line in `#444` (dark gray) from z16 when mapped on a closed way tagged with area=yes or tagged with a key that is considered to be an areas feature. 

Amenity=community_centre + barrier=no (probably meant to describe wheelchair accessibility?)
https:/www.openstreetmap.org/way/392903881/#map=18/53.28442/-3.80404
memorial-hall-barrier-no

Rendering test without barrier=no rendering:
Memorial-hall-after

jeisenbe:
Fixes #3742

-  Select only specific values of barrier for rendering from the polygon geometries, matching the same list used for line-barriers 
- Start rendering historic=citywalls from polygon geometries for consistency with rendering of barrier=city_wall
- Stop rendering barriers on polygons where building is not NULL

## Explanation:
- Currently all values of the key barrier are selected when mapped as polygons. 
- - This includes all relations of type=multipolgyon and all closed ways tagged with area=yes or with any of the keys that are considered a polygon, such as "building", "amenity", "tourism", etc.
- - All values are picked, including barrier=no and barrier=nonsene
- historic=citywalls is rendered on line geometries, but not on polygon geometry. This is a problem if a feature is mapped as a closed way and double-tagged with historic=citywalls and tourism=attraction; the feature will be imported as a polygon, even though editors and mappers think of this as a linear feature.
- Buildings do not need rendering of walls; our standard building rendering already suggests that there are walls around the outline of a building. In particular, "building=no" should not result in area fill rendering when tagged along with "barrier=hedge"

## Test renderings:

Amenity=community_centre + barrier=no (probably meant to describe wheelchair accessibility?)
https:/www.openstreetmap.org/way/392903881/#map=18/53.28442/-3.80404
Before

After

Test area
Before
barriers-test-before
After
barrier-test-after

### DIFF
--- a/project.mml
+++ b/project.mml
@@ -621,11 +621,16 @@ Layer:
       <<: *osm2pgsql
       table: |-
         (SELECT
-            way, barrier AS feature
+            way, COALESCE(historic, barrier) AS feature
           FROM (SELECT way,
-            ('barrier_' || barrier) AS barrier
+            ('barrier_' || (CASE WHEN barrier IN ('chain', 'city_wall', 'embankment', 'ditch', 'fence', 'guard_rail',
+                  'handrail', 'hedge', 'kerb', 'retaining_wall', 'wall') THEN barrier ELSE NULL END)) AS barrier,
+            ('historic_' || (CASE WHEN historic = 'citywalls' THEN historic ELSE NULL END)) AS historic
             FROM planet_osm_polygon
-            WHERE barrier IS NOT NULL
+            WHERE (barrier IN ('chain', 'city_wall', 'embankment', 'ditch', 'fence', 'guard_rail',
+                  'handrail', 'hedge', 'kerb', 'retaining_wall', 'wall')
+              OR historic = 'citywalls')
+              AND building IS NULL
           ) AS features
         ) AS area_barriers
     properties:


jeisenbe:
One question: should barriers be rendered on closed ways with with `building=no`? 

This could happen in the case of an open-air market with a wall, tagged `barrier=wall` + `amenity=marketplace` + `building=no` on a single closed way,
or `barrier=wall` + `area=yes` + `building=no` on a closed way, for example.  

I believe the current code `WHERE (...) AND building IS NULL` provides the best performance, because the building index can be used to discard all building features, right? 

This is also the line used for many of the polygon layers, eg `landcover`, `water-areas`, `landcover-area-symbols`, `landcover-overlay`, `protected-areas`, and the related text layers.

But alternately, I could change it to `AND building IS NULL OR  building = 'no'`, which allows rendering in these cases, eg:
test-area-building-no

imagico:
Much in favor of stopping to render barrier=no and moving to an explicit list of values instead of a catchall.

The question of `building IS NULL` vs.  `building IS NULL OR building = 'no'` is a more generic question for anything where we use that condition.  We are currently not consistent about this by the way, we use this in landcover-low-zoom, water-areas and landcover-area-symbols but not in the landcover layer.  This is both inconsistent in rendering results and inefficient in use of indices.

I would in general be careful with excluding features from rendering based on complex conditions with secondary tags - this can be rather confusing.

jeisenbe:
> We are currently not consistent about this by the way, we use this in landcover-low-zoom, water-areas and landcover-area-symbols but not in the landcover layer. 

Right, `AND building IS NULL` is used for those layers including `landcover-low-zoom` (and also `landuse-overlay` and `protected-areas`).

But the filter is not used in `landcover`. 

I speculate this is because buildings are relatively small features at the z9 and below, usually smaller than 0.01 pixel, so filtering out anything double-tagged with `building` and a landcover tag is unlikely to make a visual difference at the zoom levels of `landcover-low-zoom` (z9 and below). But `landcover` is used to render z10 to z14, where landuse and natural areas are shown but not buildings. At z12 to z14 large buildings are larger than a pixel, so excluding a feature double-tagged with `building=warehouse` and `landuse=industrial` could make a significant difference.

> I would in general be careful with excluding features from rendering based on complex conditions with secondary tags 

I decided to check how common it is to add `building=no` to a barrier. Out of the top 4 line barrier values (wall, fence, hedge and retaining_wall), only "wall" had enough features combined with `building` to be shown in taginfo. There are 2.5 million barrier=wall features, and only 2,700 have "building", and over 2000 of those are "building=yes". So there cannot be more than 700 uses of building=no with barrier=wall, or not more than 3 in 10,000. 

So I don't think it's necessary, especially if it will slow down rendering performance to check for building=no

imagico:
> I speculate this is because buildings are relatively small features at the z9 and below, usually smaller than 0.01 pixel, so filtering out anything double-tagged with building and a landcover tag is unlikely to make a visual difference at the zoom levels of landcover-low-zoom (z9 and below). But landcover is used to render z10 to z14, where landuse and natural areas are shown but not buildings. At z12 to z14 large buildings are larger than a pixel, so excluding a feature double-tagged with building=warehouse and landuse=industrial could make a significant difference.

At the low zoom levels the landcover layer should use the way_area based indices.  These automatically also eliminate almost all of the buildings which are smaller than the thresholds.  At high zoom level however excluding several hundred million buildings would be a significant advantage.  I am not sure why it was never used in the landcover layer.  Your argument is that it ensures that buildings also tagged with a landcover tag are shown in color.  But since that is poor tagging it does not really make much sense and landcover-low-zoom does it differently which is also inconsistent, in particular since landcover-low-zoom should use the way_area based indices so the building IS NULL condition will not make a performance difference there it seems.

Tomasz-W:
@jeisenbe Can you test thinner citywalls lines? I think they are way too bold in comparsion with simple walls now (in some places they could be even confused with some kind of power lines). Example places which shows the problem:
https://www.openstreetmap.org//#map=19/52.40885/16.93095
https://www.openstreetmap.org//#map=19/52.40556/16.93329
https://www.openstreetmap.org//#map=18/52.24915/21.01100

jeisenbe:
Good idea. Could you open a new issue?

Tomasz-W:
@jeisenbe #3747

jeisenbe:
This PR is ready to merge, unless there are any other comments or concerns?

jeisenbe:
1) building IS NULL vs. building IS NULL OR building = 'no'

- I'll make a new issue about this, and about updating the indexes to match

2) I agree that the polygon fill can be confusing. I've got another PR ready to discuss this, after this PR is merged

jeisenbe:
Also fixes #3250

PR #3749, opened 2019-04-17
Reduce rendering priority of place=locality and raise initial zoom level to z16

jeisenbe:
- Change place=locality initial zoom to z16
- Move place=locality to `amenity-points` layer instead of `placenames-small`

## Explanation
- The tag [place=locality](https://wiki.openstreetmap.org/wiki/Tag:place=locality) was developed in 2007 for tagging an "unpopulated location for which there is no extant feature to which the tag could be associated." 
- Most of the uses of this tag are from imports in the United States, from the GNIS database and similar public data sources. These are usually "former human settlements for which all taggable features are gone, but which still see use (and/or have value) as named locations. This is most frequent in the case of former hamlets, or for historical railroad stations or railroad junction locations."
- The wiki page for this tag advises that it should not be used when another tag is appropriate, such as another `place=*` tag, administrative boundaries, or natural features such as `natural=cape`.
- However, a review of tag use in Hawaii, Delaware, Andorra and Liechtenstein  shows that that the majority of uses of this tag are for features that could be described more precisely. See the discussion the [Tagging mailing list](https://lists.openstreetmap.org/pipermail/tagging/2019-April/044648.html).
- Many of these features are small and should not be shown at z15, or with priority over other features such as `amenity=hospital` or `place=isolated_dwelling`.
- Therefore, place=locality should be rendered in the `amenity-points` layer, similar to `place=island` and `place=islet`
- The initial zoom level should be changed from z15 to z16 to better represent the kind of features that are tagged, including crossroads within a hamlet, small squares, guideposts, forest clearings, etc.
- This will also encourage mappers to consider adding other tags which will render sooner or with a more specific rendering

Ka'awaloa, Hawaii - tagged as locality with a name only
https://www.openstreetmap.org/#map=16/19.4793/-155.9342
Before z15
z15-ka-awaloa-before
After
z15-ka-awaloa-after

z16 Before
z16-ka-awaloa-before
After (not rendered because the Park label has higher priority)
z16-ka-awaloa-after

z17 Before
z17-ka-awaloa-before
After
z17-ka-awaloa-after

Kawainui, Hawaii - tagged as locality and name only
https://www.openstreetmap.org/#map=17/19.82840/-155.09774
z15 Before
z15-kawainui-before
After

z16 Before
z16-kawainui-before
After
z16--Kawainui-after

z17 Before = After
z17-Kawainui-after

### DIFF
--- a/amenity-points.mss
+++ b/amenity-points.mss
@@ -1769,6 +1769,26 @@
     text-placement: interior;
   }
 
+  [feature = 'place_locality'][zoom >= 16] {
+    text-name: "[name]";
+    text-size: 10;
+    text-fill: @placenames;
+    text-face-name: @book-fonts;
+    text-halo-fill: @standard-halo-fill;
+    text-halo-radius: @standard-halo-radius * 1.5;
+    text-wrap-width: 45; // 4.5 em
+    text-line-spacing: -0.8; // -0.08 em
+    text-margin: 7.0; // 0.7 em
+    [zoom >= 17] {
+      text-size: 12;
+      text-wrap-width: 60; // 5.0 em
+      text-line-spacing: -0.60; // -0.05 em
+      text-margin: 8.4; // 0.7 em
+      text-fill: @placenames-light;
+      text-halo-fill: white;
+    }
+  }
+
   [feature = 'amenity_pub'][zoom >= 18],
   [feature = 'amenity_restaurant'][zoom >= 18],
   [feature = 'amenity_food_court'][zoom >= 17],
--- a/placenames.mss
+++ b/placenames.mss
@@ -436,7 +436,6 @@
 
 #placenames-small::neighborhood {
   [place = 'neighbourhood'][zoom >= 15][zoom < 20],
-  [place = 'locality'][zoom >= 15],
   [place = 'isolated_dwelling'][zoom >= 15],
   [place = 'farm'][zoom >= 15] {
     text-name: "[name]";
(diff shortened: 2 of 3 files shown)


jeisenbe:
Does anyone have any comments on this PR? I think it's reasonable to render localities with a lower priority and later zoom level than inhabited places. See the more extensive research and discussion on the tagging mailing list: https://lists.openstreetmap.org/pipermail/tagging/2019-April/044648.html

imagico:
Forgot to mention:  place=locality should also be last in the COALESCE() to make sure any other tag has priority over it.

jeisenbe:
Thank you, I've made the recommended changes, and I checked that the rendering is now correct for localities mapped as points (render at >= z16) and areas (do not render).

I didn't quite understand how the new combined `amenity-points` layer works. 

So it looks like we need to add "AND way_area IS NULL" to any new feature that is only to be rendered from points, or it would be rendered when mapped as an area (if the key is imported as a polygon). Example:

```
                'barrier_' || CASE WHEN barrier IN ('toll_booth') AND way_area IS NULL THEN barrier ELSE NULL END,
```

And if a feature icon or name label is only supposed to be rendered when it is mapped as an area, we need to check for "AND way_area IS NOT NULL", eg:

```
CASE WHEN amenity IN ('waste_disposal') AND way_area IS NOT NULL THEN amenity ELSE NULL END,
```

Issue #3705, opened 2019-03-05
Do not render water areas names at z2, z3


matkoniecz:
### Expected behavior

No water area names are rendered at z2 (only oceans are a potential fit and ocean naming is a subjective cultural issue, there are even multiple competing ideas how many oceans exist, making it a really poor idea)

### Actual behavior

![selection_001](https://user-images.githubusercontent.com/899988/53807772-b8b03500-3f50-11e9-88ee-23b3e2f3aa15.png)

I strongly suspect tagging for renderer here, but it is a separate problem.

### Links and screenshots illustrating the problem

See above

imagico:
Duplicate of #3634.

The solution would IMO be to stop rendering labels based on way_area for features for which there is no visual feedback on the geometry in the map.  Moving the threshold up to z4 would just attempt hiding the problem but not solve it.  The incentive for free style label polygon drawing with no basis in verifiable reality is not limited to z3 and below.

matkoniecz:
Thanks for spotting duplicate!

Issue #3284, opened 2018-07-02
Labeling based on size (meta issue)


Tomasz-W:
There is an issue with labeling based on size for some POI areas, but the discussion about this problem and how to resolve it broken into few pieces and get messed. I propose to close some dead tickets (which actually are the same multiplicated ticket) and move whole discussion into one meta issue.

The most substantive parts was in: 
* https://github.com/gravitystorm/openstreetmap-carto/pull/2361#issuecomment-249083347
* https://github.com/gravitystorm/openstreetmap-carto/issues/2704#issuecomment-328608781

Involved tags:
- [ ] amenity=school (https://github.com/gravitystorm/openstreetmap-carto/issues/120)
- [ ] amenity=hospital (https://github.com/gravitystorm/openstreetmap-carto/issues/1019)
- [ ] amenity=prison (https://github.com/gravitystorm/openstreetmap-carto/issues/1022)
- [ ] tourism=camp_site (https://github.com/gravitystorm/openstreetmap-carto/issues/1020)
- [ ] amenity=place_of_worship (https://github.com/gravitystorm/openstreetmap-carto/issues/1052)
- [ ] historic=archaeological_site (https://github.com/gravitystorm/openstreetmap-carto/issues/2073)
- [ ] amenity=university, college (https://github.com/gravitystorm/openstreetmap-carto/pull/2361)
- [ ] amenity=childcare, kindergarden (https://github.com/gravitystorm/openstreetmap-carto/issues/2657)
- [ ] tourism=museum (https://github.com/gravitystorm/openstreetmap-carto/issues/2704)

Related tickets:
* Render labels of large POI larger https://github.com/gravitystorm/openstreetmap-carto/issues/1055
* Render icon and icon+text for large POIs on lower zoom levels https://github.com/gravitystorm/openstreetmap-carto/issues/1057
* Outline coordination https://github.com/gravitystorm/openstreetmap-carto/issues/3045

polarbearing:
IIRC what blocks scalable labelling for some objects above is the **presence of icons**.
We have not yet agreed how to handle this, i.e. the icon would need to scale with the text label.
One idea was to have a set of label sizes, e.g. three or so, with according text font sizes.

@Tomasz-W - Thanks for your summary in this ticket!

PR #3144, opened 2018-03-24
Adding rendering for big natural=bay

kocio-pl:
Rendering big bay labels is currently missing, notable example is [Port Phillip Bay](https://www.openstreetmap.org/relation/1221199#map=9/-37.9962/145.0923) in Australia:

Using Mapnik 3.0.15 (currently in Kosmtik):

z8
rik1uo8i

z9 (probably eclipsed by a town name)
p3rcx_d4

Mapnik 3.0.19 (latest release):

z8
cdc4iu6v

z9 
hxf tbsq

Related to #2068.

PR #3750, opened 2019-04-17
Render name labels of bays and straits from z14 only, lakes from z5 only

### LINKED ISSUE #3634: Limit natural labels rendering to z4+

jeisenbe:
Fixes #3634 and #3705
Reverts #3438 and #3144
Related to discussion of `way_area` based name label rendering in #3284 and #2700, also see #2702

- Stop using `way_area` to set zoom level and text size for natural=bay and natural=strait text labels
- Render name labels for straits and bays at z14 and above
- Do not render text labels for lakes, reservoirs or docks before z5

## Explanation

- Since #3438 and #3144, natural=strait and natural=bay have been rendered identically to lakes (natural=water and landuse=reservoir) when mapped as an area (closed way or multipolygon relation). Depending on the `way_area`, they will render as soon as z0 in theory, and in practice there is currently a bay rendered on z3. 
- This led to issue #3634, which recommends that natural area labels should not be shown at z4 and below.

- The mapping of bays and straits as areas is controversial, because the non-coastline borders of these features are not verifiable by mappers, and the coastline is already contained in the database. Also, mapping large bays or straits with multipolygons relations can lead to difficult to manage relations with many hundreds of members. 
- Because of this, some mappers have taken to mapping a bay or strait with a rough outline that does not follow the coastline. This is encouraged by the current rendering, because these features are only shown as a text label with the name, and there is no mapper feedback about the shape or borders of the area in the map rendering.
- See also: https://www.openstreetmap.org/user/imagico/diary/43957 [Warning, dark humor]
- Recently a text label rendering was added for straits mapped as linear ways in #3733

- We recently stopped rendering tourism=attraction based on `way_area` because of this same problem, in PR #3603
- Previous to PR #3732  natural=capes mapped as an area were rendered based on `way_area`, but this has now been removed.
- Bays, straits, and malls are the only 3 features that currently have text labels that are rendered based on `way_area` without an outline or area fill rendering.

- This PR will fix #3634 and #3705 by limiting natural=water, landuse=reservoir, landuse=basin and waterway=dock rendering to no sooner than z5. Even the largest lakes in the world do not need text labels before this zoom level, and natural=water is being abused to map marine water areas for label rendering.
- The initial zoom level will be set to z14 for all straits mapped as nodes, ways or areas and all bays mapped as nodes and areas. This level has previously been tested and deemed an acceptable initial zoom level in PR #3733 .

(The PR author plans to submit another PR to add rendering for bays (specifically fjords) mapped as linear ways similar to #3732, and an additional PR to limit the rendering of malls to no sooner than z13 (where the retail landuse fill color is shown) and add a fill or outline in retail color, which should finish resolving the`way_area` based text labels issue.)

While many of us would like to see bays and straits rendered at low zoom levels in this style (the author included), we have a responsibility to support good mapping by providing appropriate feedback. Therefore in the short term this PR is the best solution.

@imagico has suggested a script could be written that would calculate the correct initial zoom level based on a node and the position of the coastline. This would allow us to restore the rendering of bays and straits at low zoom levels, and not only for areas but also for nodes. It would also make it possible to render seas in a consistent fashion, which is not currently possible because these are mapped as nodes. However, there is not yet a volunteer to implement this. [EDITED]

## Test renderings:

**Yam Bay, Hawaii** - mapped as an oval-shaped closed way rather than straight across from the ends of the bay; this causes it to render 1 zoom level sooner than otherwise:
z13 before
z13-yam-bay-before
After
z13-yam-bay-after

z14 before
z14-yam-bay-before
after
z14-yam-bay-after

z15 before
z15-yam-bay-before
after
z15-yam-bay-after

**Waioka Bay, Hawaii**
- I was surpised to find that "Waioka Pond", a very small "bay", currently renders at z14. This is because the initial zoom was set to z14 in amenity-points.mss, which overrides the `way_area` based filtering which should have prevented rendering until later. So the current `way_area` adjustment only shows labels sooner, never later.
z13 unchanged
z13-waioka-bay-after
z14 Before
z14-waioka-bay-before
After
z14-waioka-bay-after

z16 Before
z16-waioka-bay-before
z16 After
z16-waioka-bay-after

**Shelikof Strait, Alaska**
z6 Before
z6-shelikof-strait-before
after
z6-strait-after

### DIFF
--- a/amenity-points.mss
+++ b/amenity-points.mss
@@ -2521,7 +2521,6 @@
     }
   }
 
-  [feature = 'natural_bay'][zoom >= 14],
   [feature = 'natural_spring'][zoom >= 16] {
     text-name: "[name]";
     text-size: 10;
@@ -2532,9 +2531,7 @@
     text-halo-radius: @standard-halo-radius;
     text-halo-fill: @standard-halo-fill;
     text-placement: interior;
-    [feature = 'natural_spring'] {
-      text-dy: 6;
-    }
+    text-dy: 6;
   }
 
   [feature = 'amenity_atm'][zoom >= 19],
--- a/project.mml
+++ b/project.mml
@@ -2035,7 +2035,7 @@ Layer:
           ORDER BY way_area DESC
         ) AS text_poly_low_zoom
     properties:
-      minzoom: 0
+      minzoom: 4
       maxzoom: 9
   - id: text-line
     geometry: linestring
--- a/project.mml
+++ b/project.mml
@@ -2012,8 +2012,7 @@ Layer:
             COALESCE(
               'landuse_' || CASE WHEN landuse IN ('forest', 'military', 'farmland') THEN landuse ELSE NULL END,
               'military_' || CASE WHEN military IN ('danger_area') THEN military ELSE NULL END,
-              'natural_' || CASE WHEN "natural" IN ('wood', 'glacier', 'sand', 'scree', 'shingle', 'bare_rock',
-                                                    'water', 'bay', 'strait') THEN "natural" ELSE NULL END,
+              'natural_' || CASE WHEN "natural" IN ('wood', 'glacier', 'sand', 'scree', 'shingle', 'bare_rock', 'water') THEN "natural" ELSE NULL END,
               'place_' || CASE WHEN place IN ('island') THEN place ELSE NULL END,
               'boundary_' || CASE WHEN (boundary = 'protected_area' AND tags->'protect_class' = '24') THEN 'aboriginal_lands'
                                   WHEN boundary IN ('aboriginal_lands', 'national_park')
@@ -2026,7 +2025,7 @@ Layer:
           FROM planet_osm_polygon
           WHERE (landuse IN ('forest', 'military', 'farmland')
               OR military IN ('danger_area')
-              OR "natural" IN ('wood', 'glacier', 'sand', 'scree', 'shingle', 'bare_rock', 'water', 'bay', 'strait')
+              OR "natural" IN ('wood', 'glacier', 'sand', 'scree', 'shingle', 'bare_rock', 'water')
               OR "place" IN ('island')
               OR boundary IN ('aboriginal_lands', 'national_park')
               OR (boundary = 'protected_area' AND tags->'protect_class' IN ('1','1a','1b','2','3','4','5','6','7','24','97','98','99'))
--- a/project.mml
+++ b/project.mml
@@ -243,7 +243,7 @@ Layer:
           ORDER BY COALESCE(layer,0), way_area DESC
         ) AS water_areas
     properties:
-      minzoom: 0
+      minzoom: 5
   - id: ocean-lz
     geometry: polygon
     class: ocean
--- a/water.mss
+++ b/water.mss
@@ -334,12 +334,10 @@
 .text-low-zoom[zoom < 10],
 #text-point[zoom >= 10] {
   [feature = 'natural_water'],
-  [feature = 'natural_bay'],
-  [feature = 'natural_strait'],
   [feature = 'landuse_reservoir'],
   [feature = 'landuse_basin'],
   [feature = 'waterway_dock'] {
-    [zoom >= 0][way_pixels > 3000],
+    [zoom >= 5][way_pixels > 3000],
     [zoom >= 17] {
       text-name: "[name]";
       text-size: 10;
@@ -370,6 +368,7 @@
 }
 
 #text-point[zoom >= 14] {
+  [feature = 'natural_bay'],
   [feature = 'natural_strait'] {
     text-name: "[name]";
     text-size: 10;


imagico:
I support this but it is clear that we need to establish consensus on the matter before this can be merged.

As a point for discussion:  A theoretical alternative to removing the way_area based labeling decisions would be to actually render the geometry by drawing things in a different color than the ocean.  This would provide feedback on the geometry and thereby address some of the arguments against using way_area.  I would be against this since (a) it is bad cartography even with accurate and verifiable mapping and (b) it would still render bays and straits fundamentally different depending on how they are mapped which is not desirable IMO.

I like to clarify one thing regarding:

> @imagico has suggested that he will eventually be able to write a script that will calculate the correct initial zoom level based on a node and the position of the coastline. This will eventually allow us to restore the rendering of bays and straits at low zoom levels, and not only for areas but also for nodes. It will also make it possible to render seas in a consistent fashion, which is not currently possible because these are mapped as nodes.

I have not made any promises that i will actually write code for that.  What i have said is that this is possible to do, most experienced developers here would be able to do this (put the coastline data and the bay/strait features into a temporary PostGIS or spatialite database and do some processing).  Anyone who wants to work on something in that direction is welcome to ask for advise on specific practical questions.

But independent of the question of implementation we would need to decide on if we want to actually have something like that in this style.  This we would need to discuss based on a specific proposal that can be tested.  I think this would probably be a good test case for trying the fundamental idea and evaluating the practical effects of it but I would not be in favor of connecting a decision on this PR with a definite forward promise to make this decision in a certain way in the future.

jeisenbe:
Sorry, I'm not sure why I thought there was a plan to develop the script. Wishful thinking perhaps? I've edited the original comment.

Should I open a specific issue to discuss preprocessing the bay, sea and strait nodes in this way?

jeisenbe:
I did not mention in the first comment that z5 is chosen as the earliest level to render natural=water (lake) labels, because this is the earliest that any of the lakes render. There are 5 that render at this level, but none are large enough for the name label to render at z4:

z5 Canada screenshot
z5-great-bear-great-slave-lake-current

z5 Russia screenshot
z5-lake-baukal

z5 Great Lakes - current screenshot
z4-great-lakes

z4 Canada current
z4-canada

z4 Russia current
z4-russia

kocio-pl:
> Do not render text labels for lakes, reservoirs or docks before z5

:+1: - does not change current rendering, but helps to prevent vandalism in a sane way, so I'm OK with that
 
> Render name labels for straits and bays at z14 and above

:+1::-1: - I'm OK with rendering them later, as they typically have less land borders, but since they can be quite big, I vote for z7+ to avoid looking for labels on the water (this way even biggest ones will be still detectable with the land still in sight)

> Stop using way_area to set zoom level and text size for natural=bay and natural=strait text labels

:-1: - I'm against both dropping area rendering and way_area (at least for now)

In short: the main reason is there are more problems with big (especially global) objects and with water objects in OSM. The discussion on tagging [is currently open](https://lists.openstreetmap.org/pipermail/tagging/2019-April/thread.html#44659), but is far from being resolved, so I believe the most responsible action is not to decide now how to change rendering and wait for the outcome.

A bit longer: these problems are really general and verifiability of nodes or lines is questionable as well. For small objects it is not that painful, but for global objects local knowledge does not really apply - the area shape can be fuzzy, the line can have unreliable ends and shape and position of node can differ by many kilometers. Since their size may vary, I would not remove rendering for dots and oppose rendering ways (if they are small enough we can reasonably round the area down to the line or node), but for larger ones using nodes is abusing the simplification - this way or another you start with some shape to find the middle of the object (at least to know you really put it inside), so it's better to show it instead of hiding the problem (this is exactly making it less verifiable IMO).

There could be different solutions to this - maybe global objects would be banned from OSM (but then geometry doesn't matter anyway), maybe there will be some special rules to cover such special cases - who knows? Maybe we would define bays and straits in a different way, so there would be more clear how to tag them. Yet it's too early to know and act, it needs some time to settle and let people decide with both discussion and practice.

imagico:
@kocio-pl - for understanding your position:  Would you be fine with keeping the way_area logic but introducing in addition a cap of way_area at 270000 (i.e. any polygon larger than this is interpreted to have only this way_area)?  That would remove the incentive to mappers to draw labeling polygons for straits and bays to make them show up at the zoom level they desire since bays/straits mapped with nodes or linear ways are not shown later than ones mapped with large polygons then.

kocio-pl:
I'm not sure if I understand you properly, so I'll try explain this way: I'd like to keep the way_area logic to retain a sense of size (since this gives better feedback than removing it), but the initial zoom level is moved so they start showing later when zooming in.

imagico:
Which is exactly what a cap on the raw way_area value would do.  A way_area of 270000 is approximately 3000 way_pixels at z14.

jeisenbe:
The thing is, the way_area logic currently only works at zoom levels below z14. Small bays currently all render at z14 no matter how small they are. That's why I just removed the `way_area` code. I don't know if this was always the case, or if something happened when the amenity-points layer was changed recently to use `ST_PointOnSurface`. 

Can anyone confirm if the `way_area` filters for bays previously prevented small bays from rendering before z14?

imagico:
Current way_area logic for bays and straits can be found in:

https://github.com/gravitystorm/openstreetmap-carto/blob/b117a298b5d153c861c118f745bf316bc4de6427/water.mss#L342-L368

That means the zoom level threshold is disabled at z17+ but the zoom level logic for the label size is the same at all zoom levels.

jeisenbe:
> Current way_area logic for bays and straits can be found in:

That's what I thought, but when I did test renderings of Waikoka pond (the smallest "bay" or probably an estuary in Hawaii), it still rendered on z14. (I think you are correct about straits).

Here's a screenshot from openstreetmap.org:
z14
z14-waioka-pond-screenshot

z18 with the closed way selected:
z18-waioka-pond-screenshot

I believe these lines in the `text-point` layer of `amenity-points.mss` are the problem:

https://github.com/gravitystorm/openstreetmap-carto/blob/b117a298b5d153c861c118f745bf316bc4de6427/amenity-points.mss#L2524-L2538

jeisenbe:
I added another commit which changes the `minzoom` from 0 to 4 for `text-low-zoom`, since water text labels will not longer render at z0 to z3 (place=island labels start at z4)

imagico:
Ah, yes, there we have some chaos in the code that needs to be cleaned up - there is also

https://github.com/gravitystorm/openstreetmap-carto/blob/b117a298b5d153c861c118f745bf316bc4de6427/water.mss#L372-L389

So all together three different variants in two different files with rules that overlap in application.  All the more reason to remove the way_area logic - since it is not working consistently at the moment anyway. :wink:

jeisenbe:
> Ah, yes, there we have some chaos in the code that needs to be cleaned up - there is also
> [openstreetmap-carto/water.mss](https://github.com/gravitystorm/openstreetmap-carto/blob/b117a298b5d153c861c118f745bf316bc4de6427/water.mss#L372-L389)
> Lines 372 to 389 in [b117a29](/gravitystorm/openstreetmap-carto/commit/b117a298b5d153c861c118f745bf316bc4de6427)
>  #text-point[zoom >= 14] { 
>    [feature = 'natural_strait'] { 
>      text-name: "[name]"; 
>      text-size: 10; 
>      text-wrap-width: 25; // 2.5 em 
>      text-line-spacing: -1.5; // -0.15 em 
>      text-fill: @water-text; 
>      text-face-name: @oblique-fonts; 
>      text-halo-radius: @standard-halo-radius; 
>      text-halo-fill: @standard-halo-fill; 
>      text-placement: point; 
>      [zoom >= 15] { 
>        text-size: 12; 
>        text-wrap-width: 37; // 3.1 em 
>        text-line-spacing: -1.6; // -0.13 em 
>      } 
>    } 
>  } 

Ah, who added that! How careless. Let me check git blame ... Oh, I did it. Oops. 

So the current initial zoom level for both straits and bays mapped as polygons is:

``
[zoom >= 0][way_pixels > 3000], 
[zoom >= 14]
```
That is, currently they are rendered as early as z0 if they are mapped as polygons that take up more than approximately 3000 square pixels on the screen. But at z14 and higher all bays and straits are rendered no matter what size. Nodes are also rendered at z14 and above. 

(Straits mapped as linear ways are rendered at z14 and above, but bays mapped as linear ways are not yet rendered; I'll fix this soon)

**Comment by jeisenbe** [comment]
@kocio-pl said "I vote for z7+ to avoid looking for labels on the water ... I'm against both dropping area rendering and way_area".

The PR does not propose stopping rendering straits or bays mapped as areas (that is, those mapped as closed ways or multipolygon relations and imported as polygons). As far as rendering, there is not any area fill or outline rendering for bays and straits currently.

Do you mean you think it should still be something like this for natural=strait and natural=bay?
``
[zoom >= 7][way_pixels > 3000], 
[zoom >= 14]
```
The problem with this code is that it has been pushing mappers to change the mapping of large bays, straits and seas. 

As of March 2018, 1 year ago, the majority of all seas, bays and straits were mapped as nodes. But in the past year the number of natural=bay relations has increased from 2,000 to almost 5,000, while the number of natural=bay nodes only increased from 45k to 46k.

Relations tagged natural=strait increased from 290 in March 2018 to 945 now in April 2019, while nodes increased only slightly.

In early 2018 there were over 120 place=sea mapped as nodes and less than 20 mapped as relations. Now there are 68 place=sea nodes and 63 place=sea relations: the total number of mapped seas has not changed, but nodes are being replaced with polygons and being double-tagged with "natural=bay". 

Notice how stable the tagging was for many years until the past 6 months:
![seas-nodes-relations](https://user-images.githubusercontent.com/42757252/56426645-fe07a680-62f3-11e9-8d6a-16367db74649.png)

We shouldn't be discouraging mappers from using nodes or ways for large bays and straits by only rendering the labels for those mapped as areas at low zoom levels.

kocio-pl:
Sorry for editing your posts - I've just added missing "`" in both because formatting was broken and I had hard time reading it.

The problem with nodes is that they lack information about size, which can vary a lot, so their rendering is trimmed to the lowest common size and I have found no way to improve it currently. That is the cost of avoiding area geometry for describing areas - some additional information is needed anyway (for example for cities we use population data to rank them). 

For small bays it works, but for big ones it doesn't. If there was tagging type which can differentiate small bays from big ones, it could be easier even on nodes, but I'm not aware of this.

jeisenbe:
>  I've just added missing "`" in both

Thanks! Sorry about that

> nodes lack information about size

The coastline is already there, and we can compute the distance from the node to the coast to find out approximately how to render the strait or bay (or sea). Above:

"this is possible to do, most experienced developers here would be able to do this (put the coastline data and the bay/strait features into a temporary PostGIS or spatialite database and do some processing). " https://github.com/gravitystorm/openstreetmap-carto/pull/3750#issuecomment-484093119

The alternative is that we could add a darker outline for bays and straits so that mappers would know when they were incorrectly drawn as a rough sketch instead of along the coastline. But I don't think this would work for seas, and it will look not so great. 

 I'm quite concerned that seas are being double-tagged as bays and redrawn as polygons, eg: the Bothnian Sea has been mapped as a node for over 7 years as place=sea - but this doesn't render yet: https://www.openstreetmap.org/node/1195287427

But now there is a "Gulf of Bothnia" relation made of just 2 ways roughly drawn 5 months ago: https://www.openstreetmap.org/relation/9048256
gulf of bothnia

imagico:
Note the idea of _avoiding area geometry for describing areas_ is somewhat misleading here.  A polygon (closed way or multipolygon relation in OSM) is just a way of defining a two dimensional entity (an area) through an explicit representation of its outline.  It is not that mapping a bay or strait with a node or linear way typically lacks significant information about the feature in question, that information is already available in form of the coastline geometry.  It is just that in our rendering toolchain that information is not readily available for making rendering decisions.

As @jeisenbe explained with #3438 and #3144 we have created an enormous incentive for creating polygons to represent bays and straits but provide no feedback on the geometry of these polygons.  This has led mappers to create thousands of polygons, many of which they would not have created without this incentive.  This is something we should not do - no matter how convenient it is for us within the technological framework we have chosen.

Note it is not that for small bays there is any less need for decent importance rating and label placement than for large bays.  In a zoomable map this is a completely scale independent problem.  It is just that at higher zoom levels the problem is more hidden and can be ignored more easily than at low zoom levels.

If we can't achieve consensus on not using way_area any more for labeling decisions for bays and straits at z<14 i would - as already indicated - propose to explicitly display the geometry of bay and strait polygons for providing proper mapper feedback on these geometries.  I don't really like that but it would be much better than the status quo.  It might be a solution _to the equal dissatisfaction of everyone_.

kocio-pl:
Sorry for not being too verbose - I still have not too much time and energy. You both pointed out a lot of interesting problems I'd like to talk about or ask about details, but it looks like this PR is a "fork bomb". :wink: So I would like to propose:
- making small PR with the non-controversial changes, which could be merged soon
- opening a ticket for discussing how to improve rendering of nodes
- keep discussing tagging problems on Tagging list (there are more people involved, which might help)

Anything that could make this problem less complex would also help, do you have some ideas?

Some background (I doubt it will close the discussion, but I believe it can at least clarify a bit my views):
- I'm all for improving rendering nodes, but if it hasn't been done for years, it's not the fault of different geometry, which allows for better rendering.
- I share the concern about tagging seas for rendering as bays. But:
    - I also think that many "node bays" was just a tagging for rendering practice that lasted for years. 
    - I also think that the problem is deeper - how do you know what is the difference between bay and sea? If we know the definitions, it can be retagged, if we don't know, there is simply a grey area to be better defined, both things are related to tagging and real QA (there are not too many seas to watch if their tagging is changed).

imagico:
@kocio-pl - i try to understand the long term strategy behind your suggestions.  You seem to insist on maintaining making the decision at which zoom level to start labeling a bay or strait mapped with a polygon from way_area - which is the central idea of this PR to abolish.  If that is not the case please say so - consensus hinges on this here so it is important not to have misunderstandings on the matter.

But apart from that you indicate to support improving the labeling of bays/straits mapped with nodes - which based on the discussions about this we had so far would mean pre-calculating an importance rating for them using the coastline data to select the starting zoom level based on that.

What i am not sure about is this:  If such a method is developed and if we achieve consensus to use it in this style if you'd want to use it for nodes only, that would mean still keeping the way_area logic for polygons while rendering nodes with a completely different system in parallel or if you then would be fine with abolishing the way_area logic and labeling both nodes and polygons using the same pre-calculated rating for selecting the starting zoom level.

jeisenbe:
I've updated this branch to fix a merge conflict.

It would be good to have input from the other maintainers. 

I see 3 possible long-term solutions:

1) Merge this PR: no longer render name labels based on the polygon size or shape, and set a policy that we do not render name labels for areas that do not have an outline or polygon fill rendering

2) Add an outline rendering for bays and straits (eg a subtle lighter or darker blue line) to provide mapper feedback, and agree to render place=sea in the same way in the future (so that mappers do not continue to map seas as natural=bay)

3) Reject this PR and agree that we can render Seas and Oceans mapped as polygons.  [For consistency this would also suggest that we would eventually add support for the rendering of name labels mountain ranges and valleys that are mapped as polygons (and probably plateaus, plains, drainage basins and other natural regions), although we could also say that this was limited to water features only.]

kocio-pl:
@imagico My long term strategy is to balance the gains/problems with each geometry choice, since I see no perfect solution for this problem. I find using way_area for water objects pretty natural. If we find a way to prioritize water objects, I would use it to make sure that parts are never rendered before the whole objects, so it's rather about initial zoom level, not taking away the sense of object's scale.

@jeisenbe 

1. :-1: - it suggests that there should be special policy for areas while there is no such for nodes and lines.

2. :-1: :-1:  :+1: - I'm not sure with that, maybe at some zoom level a delicate border could be rendered, but if so, I think we should render a dot for nodes, to show real geometry choice in all the cases. I guess using nodes works currently exactly because we hide the real geometry used and pretend they are areas, otherwise the problem with nodes would be already clear.

3. :+1: :-1: - Favoring nodes (and lines) is what I would like to stop in general, because they have their own problems. However I think that problem is much wider and trying to solve all the cases in one ticket is too far stretched.

jeisenbe:
> "it suggests that there should be special policy for areas while there is no such for nodes and lines.

I don't understand this comment. With this PR, bays and straits would be rendered the same whether they were mapped as nodes or areas: they would get a name label at z14, on the center of the polygon or node. 

Currently there is a special rendering for bays and straits that are mapped as polygons which allows them to be rendered at lower zoom levels and with larger text.

> "I think we should render a dot for nodes, to show real geometry choice in all the cases"

This isn't necessary to show the geometry, because the name label is centered exactly over the node, so if you zoom in to z19 it shows the location of the node quite precisely. This is the same situation with place=village/hamlet: we just render the name.

> "Favoring nodes (and lines) is what I would like to stop in general, because they have their own problems"

This PR would not favor bays or straits that are mapped as nodes. They would get an identical rendering to those mapped as areas. Currently it is those that are mapped as areas that are favored by the special rendering.

I'm mentioning seas because they are quite similar to bays, and some seas (Banff Bay, the Bay of Bengal) are even named "X Bay" in English, or "X Gulf" (eg. Gulf of Mexico). If we render bays as soon as z5 when mapped as an area, then we also need to render seas in this way. It's rather inconsistent to show the Bay of Bengal but not the equally large Arabian Sea on the other side of India.

imagico:
@kocio-pl - i still don't understand your position.  You write _I'm all for improving rendering nodes_ and i am trying to understand what you mean by that.  Does this mean you are only for improving rendering of nodes but insist on keeping the way_area logic (and thereby direct mapper control over the labeling) for polygons?

This is a pretty fundamental question regarding this matter.

@jeisenbe

:+1: on 1.
:+1: on providing visual feedback on the polygon geometry when way_area is used.  This is cartographically not a good solution but it would be a step forward from the status quo. :-1: on adding any new features with way_area logic for rendering decisions.
:-1: on 3. for the same reason.

kocio-pl:
> Does this mean you are only for improving rendering of nodes but insist on keeping the way_area logic (and thereby direct mapper control over the labeling) for polygons?

Yes, that's what I mean.

kocio-pl:
> I'm mentioning seas because they are quite similar to bays, and some seas (Baffin Bay, the Bay of Bengal) are even named "X Bay" in English, or "X Gulf" (eg. Gulf of Mexico).

Seas and oceans have two huge problems with regard to rendering, which make them different enough:
- identifying them - some say there is only one big ocean for example
- rendering names - there can be for example 10 countries using different languages around; rendering all of them seems too much and there is no easy way to say which ones to choose

That's why I would keep them in separate tickets.

imagico:
Thanks, this is much clearer now.

Chances that someone will develop an automated rating system for bays and straits for this style are very small of course if that cannot be used for a uniform rendering of all straits and bays no matter how they are mapped.  I would probably not support such a change without having the advantage of a unified rendering justifying the complexity.

I would still suggest to provide feedback on the geometry of bay and strait polygons when way_area is used for the rendering decisions.

kocio-pl:
I look for some more information about your ideas, because it seems to me that I don't get some important details. Here is what I think currently:

I thought that improving nodes rendering would just look like rendering areas. Adding some information like "size=300 sqkm" would effectively work as declaring the size which is computed for areas, just without declaring the boundaries (but declaring where is a center of them instead).

There can be more complex algorithm for ranking water objects, of course. We use some for urban places and it uses population data as a base. For water objects area size seems to me as a natural base for such algorithm, no matter if this is declared or computed. I have no idea what could be the base instead in this case.

Rendering every water object from z14 is something I consider something highly annoying and under-designed. This rendering looks to me as a secondary to "tagging for searching", because it works only for small bays. For everything medium- and large-sized it demands user to guess there can be some label at all, then guess where the center of it is tagged, and then zoom a lot seeing no land or any other hint around. I failed many times doing this even when suspecting the bay is labelled, and it was not possible without combining my guessing with 3rd party informations to see the name, which I personally consider horrible and I call it anti-experience form the user perspective.

Since I think that node is reasonable only for small bays anyway, I can stand it (and consider area tagging as a proper solution), but I don't like it still. Ideally there should be no case when the label is not visible where the whole object fits the screen, no matter what geometry user chooses.

How do you see this problem and what solution would you like to apply (if you do consider it as a problem at all)?

imagico:
Automated importance rating of bays and straits would be based on the coastline geometries.  It would therefore not depend on the mapper manually choosing how the label should be rendered by either drawing a labeling geometry accordingly or using other means.  But it would of course also deny the mapper the ability to subjectively decide which bays are important and should be shown at low zoom level and which bays are not and should be shown later.

jeisenbe:
It would really help to have a working sample where we could show how
it would work.

Several people on the tagging list were skeptical about using the
coastline and nodes alone.

While I understand how I could be done, I don't know how to write that
kind of script myself.

I hope that if we have a way to show name labels at a good zoom level and size,
even without mapping as polygons, most of the objections will be resolved?

imagico:
> I hope that if we have a way to show name labels at a good zoom level and size, even without mapping as polygons, most of the objections will be resolved?

That was my hope as well but @kocio-pl indicates otherwise.

Obviously we can't make a definitive decision without a concrete proposal for a change.  But my hope was that we could remove the bad and ultimately pointless incentive for drawing labeling polygon geometries we have right now to avoid both mappers investing a lot of time into this and at the same time to increase the incentive to develop an alternative.

My fear is that with the status quo in rendering if in maybe a year or two someone develops an automated importance rating system there will be pressure to reject this to maintain the justification for the investment of hundreds or thousands of mapper hours into drawing and maintaining labeling polygons made in the meantime, i.e. we run into an escalation of commitment problem.

kocio-pl:
To state obvious thing at the beginning: all the geometries for bays/straits etc have problems, including nodes. What makes me very cautious is lack of confirmation of this simple fact.

I can perfectly understand why someone feels better with nodes instead of areas. But I also understand when someone feels better with areas than nodes (and I share this attitude). The problem is that there are always some not clear borders that can't be observed on the water surface. Geometry will suffer because of this, no matter which one will we choose. That's why I think we should allow people to decide and render all of them.

If I had to choose just one, I would remove nodes rendering. It is easy to verify that it's always the area, so this important rule is working at the lowest level against them, while at the same time it strongly supports area tagging. Node can work as a simplification, but when you want to verify if this is inside you have to declare some boundaries, so they are always less verifiable, because there are many ways to calculate centroid of natural shapes - the center is easy to assign only for simple, symmetrical shapes, for all the rest this is subjective and nobody gives the source borders nor the algorithm chosen for this centroid.

This way or another, one always has to rely on 3rd party sources, because the named area relies on naming, which is not a natural, visible feature. People can name any arbitrary part of the water as bay and it's up to mapper who does she trust and how much. The only physical verification I can imagine is a plaque along the border showing the name.

OSM Carto should not work as a QA tool when there are some controversies and there are, so I do not propose to remove rendering nodes. I believe they are less accurate than areas (or ways), but for small objects this can be used in a sane way - probably for z14+ objects, as currently. For bigger objects it is not reasonable assumption and should not be supported. Unfortunately we have no additional properties to drop node rendering only for bigger objects. 

That's why I support current state of the things. Moving initial zoom level for areas is meant to make things more balanced, just like we don't start rendering USA, China or Russia earlier than other countries, but this is secondary concern and I support it as long as there is no searching the labels.

imagico:
Probably needless to say that i disagree with almost everything you wrote.  This disagreement is not primarily on the specific level of cartographic design but on the more fundamental level of the function and mandate of this map style (mapper support vs. mapper steering) and of OpenStreetMap as a whole (documenting verifiable information about the geographic reality vs. providing a collection of useful data for data users).

Anyway - these subjects we have discussed several times already without this leading to a productive result so i don't think it is productive to do so again here.

The problems i see here - and i have partly indicated that already in https://github.com/gravitystorm/openstreetmap-carto/pull/3750#issuecomment-490961195 - are:

* that the status quo (which i like to point out is not the result of a consensus decision) is strongly pressing mappers into investing a lot of time into creating and maintaining these labeling geometries - time that could much more productively be invested into other things and that mappers would invest in other things if they were not strongly incited to do otherwise by this style.  This is something i find irreconcilable with our mandate here.

* connected to that on a more abstract level this project in that regard is in a classic pre-enlightenment situation where we are - to quote Kant - in a state of self-incurred immaturity.  I can see how people believe that because polygon mapping of bays and straits seems universally helpful for more differentiated map rendering this style of mapping has to be a positive thing and it has to be morally right to encourage it.  But i can also clearly see how exactly you are mistaken in that belief.  Unfortunately i have no way to make you see this beyond what i have tried already.

Eventually automated rule based cartography will - likely universally - move towards importance rating and label placement for bays and straits based on automated analysis of the physical geography.  The question we have to ask ourselves is if we want to contribute to shaping this future or not.

kocio-pl:
General disagreement is hard to surpass, but I see some pretty concrete problems here:

- general remark about mapper support vs. mapper steering - I don't see a problem here, steering is OK for me as long as we don't push people and don't drop proper cases; rendering is a "poor man's QA tool", since in typical QA you get warnings, explanations and errors, in rendering you get only visible/invisible, so we should be extra careful with it

- verifiable vs useful - my point is exactly that node rendering is _both_ less useful and not verifiable (or "verifiable as false"), so no contradiction in this case; my question is why do you insist on nodes when _nobody_ claims that bays are nodes?

- maintaining any complex shape is harder, but this is true for all the cases, with coastlines and admin borders as the most prominent examples, yet nobody proposes to replace them with nodes for example; it's also not our responsibility or mandate to tell mappers what they should spend their time on doing

- I heard no moral argument there and I don't see it as any kind of moral fight, it's just that all the geometries are hard on water - and bays, straights etc. as nodes are clearly against geometry verification, so this is technical problem for me

- I have thought about importance rating for bays/straights and I see not anything we could use instead of size, so I don't believe in better solution; if there is, please show them so we could discuss in separate ticket

@jeisenbe Do you see any part of this ticket to be universally agreed or how to modify it? I guess every part of it met resistance, so it should be probably closed.

imagico:
@kocio-pl - all your points have been discussed in the past - either here on the issue tracker or (in particular on the broader subject of verifiability) in other venues.  You are making the same arguments based on the same questionable interpretations of mechanisms, conventions and statements made as before.  I don't think it is productive to repeat this discussion once more with the same predictable outcome.

As far as the moral dimension of the whole subject is concerned.  A question of responsible decisions is always a moral question.  And as said the idea that a decision is necessarily morally right if you think it is at the moment you make it goes too short.

Apart from all disagreement on the matter what worries me most is that in the whole discussion you have not with a single word acknowledged that

* the status quo in bay and strait rendering is not the result of consensus
* that - for better or worse - you are largely responsible for mappers investing a lot of work into drawing labeling polygons for bays and straits and will - if the status quo is maintained - likely invest even more time in further adding and maintaining those.

Such acknowledgement would in no way prejudice the consensus building process but it would be immensely important to demonstrate that you are willing to accept responsibility for decisions you made in the past, are aware of the extent of the effects these decisions have and are at least in principle willing to critically evaluate these decisions based on arguments and evidence presented without using the fact that it is the status quo as an argument per se.

kocio-pl:
> all your points have been discussed in the past

I'm not sure about most of the other problems, but I don't remember my main question here to be related by anyone, so I'm still waiting for it.

Could anyone give me a link to any response for the question about bay nodes being verified as false geometry choice ("is this bay a node?") or - even better - reply now?

> the status quo in bay and strait rendering is not the result of consensus

I was not aware what answers do you expect, so sorry for lack of addressing it, but **I acknowledge** it now: you can check that there was a consensus for areas of bays and for straits (#3144 and #3438), for lines it was the same (#3733 and #3754). For nodes I don't even know when they were introduced and if more than one person agreed on rendering them, so it would be good to check it.

Moral arguments are what I ask myself ("does it bring more good than bad?"), not anyone else, because I find it offensive to ask others about (in my view "why do you make immoral choice?" would be even against CoC). I feel responsible for my actions and I'm general happy with results, and when I'm not, I propose the change (or support someone else's action), so **I acknowledge also this**. Mappers are not forced by my choice with bay tagging, they can still tag the way they feel is right - lack of size for nodes is clearly reflected in rendering and this is the consequence they take. Please don't take away responsibility from them.

I guess you also want the best and you are also responsible for your actions (and lack of actions), so I do not question moral value of your choices even if I disagree with some of them.

imagico:
> Could anyone give me a link to any response for the question about bay nodes being verified as false geometry choice ("is this bay a node?") or - even better - reply now?

You are arguing a statement that no one has ever made here and you are using a claimed lack of verifiability of some data (which you never supported with any evidence) as justification for adding other non-verifiable data.  And you are not using this as an argument to remove rendering of bays and straits in this style (which would be the logical thing to do) but to favor and encourage one method of mapping compared to another. This is not a convincing line of argument.

The question of verifiability of node locations for representing bays and straits in the database has been discussed many times - the first time i am aware of was in October 2014 in the thread _natural=bay as nodes are evil_ on the tagging mailing list but plenty of other cases later as well.  I have explained the verifiability both practically and theoretically in different ways on many occasions.  But it does not really matter in the end because as explained above you use it as a straw man argument for something different anyway.

Your acknowledgements are the opposite of what i indicated i miss in your statements here.  You are specifically rejecting responsibility in those points i consider important.  This to me once more underlines that our difference is on a very fundamental level.  Your statements on the moral aspects are still valuable though because they at least seem to give a bit of a hint how you come to the views and opinions you have.

And i do not question the moral value of your choices, i encourage you to critically reflect on it yourself.

But in the end i don't think this discussion should overshadow the admirable attempt of @jeisenbe to come to a consensus on this subject to allow us moving forward in a productive way.  If - as it seems right now - that is not successful here that is definitely not his fault.

kocio-pl:
I see the big missing piece here. Node location (which you seam to treat as the core of the problem) is _secondary_ to deciding if node is right geometry at all. And there is a simple question which shows that it's not. So let's start with basics - what are the reasons to skip this check before doing anything else? How do you deal with that _primary_ node verifiability problem (and I don't mean just you, Christoph, but anyone supporting this idea)? Geometry is much more than just location - I guess we can agree here?

This is not a straw man, I really believe that once we got into the basics, let's explore them. It was not my idea to push hard toward verifiability rule to resolve this case, but this is the consequence of sticking to this rule as necessary. Or maybe is it not necessary, just quite important? I don't get your position on that and I try to figure it out (and still I don't mean just you, Christoph - English is misleading here).

I believe I explained that the sources of such verification are 3rd party mostly (just like translations) for any geometry choice and you have not shown any source for nodes that is better or not questionable when I asked about it. There are no better sources for nodes than for any other geometry choice, but there are better and worse questions and answers. If you don't start with "is this the area? If yes, what are the borders?" you might end up with node (for example a local tapping a finger on the screen or paper print). But this is the same oversimplified question ("where is this?") that would make a country a node also and I believe no one would agree that was a good enough attempt for verification, even if there are questionable or undefined admin borders.

I don't remember any practical nor theoretical proof for choosing nodes for bays, only instruction how to set locations once we know the area, which is not the same. As I understand you personally reject 3rd party sources, but have not shown other type of source for nodes. Please show a source for nodes that is both:
- not 3rd party
- can not be used for verification of an area

Is there such a thing at all?

My responsibility is for my decisions. OSM is heavily distributed, decisions are not centralized and everybody gets her part of it. If I took responsibility for people doing hard work with mapping, I would take away their credit also. But if you like to think of it as only my work, I would be even more happy with this choice - it's your choice what you prefer to think about it.

Adamant36:
> that - for better or worse - you are largely responsible for mappers investing a lot of work into drawing labeling polygons for bays and straits and will

It seems like from the tagging history that the rate of mapping bays has been largely unchanged since 2018 when @kocio-pl made his decisions that you take issue with. If anything, it looks like rendering of nodes for bays has gone down since then and most bays were already mapped before that point. Unless I'm missing the point you are making, the rendering decisions by @kocio-pl haven't had any effect on things, or even a negative effect. At least as far as raw mapping numbers goes. 

nodes 
bay node
way
bay way
both
bay any type

jeisenbe:
> nodes for bays has gone down since then

Right, they've been changed to multipolygons. These are mapped as relations, which are missing from the chart above. 

While the number change is not huge compared to the number of bay nodes, the visually effect is larger than this chart shows, because many of the new multipolygon relations are very large features: large bays, gulfs and even seas like the Bay of Bengal.

Chart of natural=bay from taghistory with nodes, ways and relations tagged natural=bay
(Most of the ways are closed ways, i.e. areas, but some are linear ways. Most of the relations are multipolygon areas)
bays-ways-nodes-relations-taghistory


Chart of natural=strait showing taghistory for nodes, ways (includes linear ways and closed ways) and relations (mainly multipolygons):
straits-ways-nodes-relations-taghistory

* Since 1-1-2018, the number of natural=bay relations increased from about 2000 to over 5000: a 250% increase

* The number of natural=strait relations increased from about 170 to over 950 in the same time: a 550% increase

jeisenbe:
Please show a source for nodes that is both:
* not 3rd party
* can not be used for verification of an area

Local knowledge of the mappers.

I know that the water north of Tacoma and west of Seattle is "Puget Sound" - but I couldn't guess how far north that name extends. 

Locally, I know that the bay off the coast of the beach west of Sentani is called [Teluk Tanamerah](https://www.openstreetmap.org/node/6444203534) (or Teluk Depapre sometimes), so I can place a node off the beach with this name, but I couldn't say exactly how far west along the coast this name extends.

I can confirm that the local name of the strait between Sumatra and Java is "Selat Sunda" - and I know from history books that it's called the Sunda Strait in English - but that doesn't tell me where the strait ends along the coastline of each island, so I'm pretty skeptical of this multipolygon: https://www.openstreetmap.org/relation/9444965/  -

Especially this corner: https://www.openstreetmap.org/relation/9444965#map=14/-5.9336/104.5774
And this corner: https://www.openstreetmap.org/relation/9444965#map=13/-6.8194/105.2258

The situation is analogous to place names.

It's easy enough to say that a node near the center of Paris is within Paris, but if you try to draw an outline of the whole place, where do you put it? Do you only include the old city within the walls? Do you include La Defense and the other new edge cities? What about the major suburbs? Do you include all of Ile de France?

While the precise location of the center node for Paris is always somewhat debatable, mappers can agree that the node is actually in Paris (even if they think it should be 100 m this way or that way), and that's good enough. And Paris as a relatively easy example for areas, since it has a well-defined minimum (the old city wall) and maximum (Île-de-France). Take a look at Atlanta instead, and despair of defining an area for the place.

We don't rely on 3rd party authorities for OSM data, because OSM data needs to be something that individual mappers can add and correct. But if we did, that wouldn't help the situation. Here's wikipedia's chart of possible areas for the Gulf of Guinea: 

https://en.wikipedia.org/wiki/Gulf_of_Guinea#/media/File:Limites_du_golfe_de_Guinée-fr.svg

Limites_du_golfe_de_Guinée-fr svg

At best this helps us to say that a node near "A" (the center of the smallest option) would probably be correct - but this would have to be confirmed by local mappers in the Nigeria and Cameroon.

Adamant36:
@jeisenbe, thanks for the extra details. Your correct that I left out multipolygons from @imagico's quote above "The question of verifiability of node locations for representing bays and straits in the database has been discussed many times" and a few other times nodes had came up, it seems like they still a large part of the problem. The vast majority of bays are still mapped as nodes also, 46,960 versus around 7,000 or so for other types. So I think the point that rendering them had zero effect on "bad mapping" still stands. In other words, there's been zero effect on node mapping. Which is still the way the vast majority of them are mapped a year later. 

Also, while the increase in multipolygon mapping might seem drastic, the conversion of nodes to multipolygons is still only a tiny amount of of all nodes that have been mapped. It's not enough to say there's a trend or a preference for it. Even if it's 250% than what it was. Since it was small to start with. 

Either way, none of the increases or changes in mapping support @imagico's claim about @kocio-pl being solely or otherwise responsible for "bad mapping." As the vast majority of bays were already mapped before he added rendering and there's been no increase overall. A switch from one way of mapping them to another doesn't matter. As they were already inaccurate before rendering was added. However they were mapped.

jeisenbe:
>  there's been zero effect on node mapping.

I disagree. If you draw a strait line from 1-1-2018 to 4-1-2018 and extrapolate that, there appear to be about 2000 fewer nodes than expected:

trendline-bay-nodes

This fits with the 3000 feature increase for relations, suggesting that about 2k features were changed from a node to a (multipolygon) relation:

>  Even if it's 250% than what it was... it was small to start with.

Straits mapped as relations have increased by 950% and now are nearly as common as nodes and more common than ways. The change has been much more dramatic for straits than for bays.

And as I speculated above, most of the bay multipolygons that have been added seem to be larger features, so the impact is much greater. Many of the remaining nodes are small features. 

>As they were already inaccurate before rendering was added, however they were mapped

Almost all of the bay and strait nodes I've reviewed have been close enough to the center of the bay or strait that they could be used for an accurate rendering based on the nodes and coastline alone, if someone with more programming skills than myself were to make a script. 

But perhaps you didn't mean that they were "inaccurate" (wrong) but "imprecise". This I can understand: it's the same situation as an `amenity=school` that is mapped as a node in the middle of the school grounds, rather than as an area. Sometimes this is necessary when it isn't possible to tell how far the school ground extend out, for example in a rural area where the school is surrounded by grass and there are no fences - unusually but sometimes seen in American high schools. 

It's true that mapping bays or straits as a point is going to be imprecise. While every node in OSM is tagged with a latitude and longitude which include a large number of digits, suggesting centimeter precision, many features that are mapped as nodes are not points but the center of a place with undefined outer boundaries, for example an unincorporated American place=town.

For a large feature like the Gulf of Guinea, precision to the nearest degree of latitude and longitude would be more than enough. But the current data model does not provide a way to show that a feature is less precisely mapped than any other. Whether a node is a member of a way or relation, or unconnected, it will be recorded with the same number of digits. 

While this is a small problem, it isn't nearly so serious as the lack of distinction between areas and linear ways: a way is merely a list of node IDs, and database users have to make assumptions about whether a way is an area or a linear way, depending on if it self-intersects, or which tags are added, such as area=yes and other keys that are assumed to be polygons. Multipolygons are clearly intended as areas, but these also can be easily broken by deleting a single or way or a terminal node, or by changing a "role=*" tag from "outer" to "inner", and there is no way to show that one side of a multipolygon is imprecise (eg the ocean side of a bay or strait). 

Ideally we would improve the database model to include a native area feature, to avoid problems with interpreting closed ways as lines vs polygons, but I don't know of any database model for marine areas with uncertain borders that is superior to a node plus mapping the coastline.

It could also be beneficial to have a type of nodes that is defined with less precision, for mapping features such as place=city, place=sea, natural=bay and natural=strait - eg. a node defined to the nearest tenth or hundreth of a degree. 

But it's unlikely that the database model will change anytime soon, so we will have work with it as is. This includes using nodes to define the center of features with non-verifiable edges.

Adamant36:
@jeisenbe, did you do research on other tags? There seems to be a general trend away from node and towards relations. For instance check out parks and schools. Which both started changing around the same time bays did. I think they would be analogous to bays in the way you stated above as to not having a precise center or clear border in some cases. For colleges nodes haven't gone down, but there's still a trend toward mapping more relations that was affected at the same time in 2018. For ridges, mapping as nodes seemed to taper off in the start of 2018 and then grow again at the same rate as relations. There was also a marked decline of a couple of hundred woods at the start of 2018. I'm wondering if there was just something happening with re tagging in general around then. It might not be related though. Either way, I don't think the changes in Bay tagging are unique or can be equated to adding rendering.  

parks
school
college relations
ridge
wood nodes

jragusa:
My 2 cents :

It's much more easier to put a node for a bay/strait/etc than create a complex polygon/relation with a great number of nodes. This would also explain the great amount of node. It would be interesting to see the correlation between the type (node/polygon/relation) and the size of these objects.

I suspect than small features are mainly polygon while very large ones are mainly nodes. Did you see such behaviour @jeisenbe ?

An other feature to consider is about the size of the label. How put large font size from low zoom for large features and small one from high zoom for small ones ?

imagico:
@jeisenbe is right on point here - and he has way more patience in explaining things than i have.

The level of whataboutism in this discussion is becoming fairly annoying so please keep in mind that this PR is about changing the rendering of polygon features tagged natural=bay and natural=strait.  If you argue exclusively with observations of how other features are mapped in OSM and reject any discussion opened up by others regarding the mapping of those features that are subject of this PR that is not helpful.

I would also like to ask that if you make claims what other people have said you carefully read what they wrote.  Arguing with something that you perceived others to have said but which they have not said does not bring us forward.  If you have the impression someone is clearly mistaken on something the best first reaction is usually to re-read the statement and check if you might have misread.

To try bringing the discussion back on track - the suggestions made so far have been

* maintaining the status quo (that is rejecting this PR)
* removing the way_area logic (that is what this PR is primarily about)
* adding mapper feedback on the polygon geometries when way_area is used (a compromise suggestion i made)

Based on the recent commentary another possible compromise (which is obviously fairly radical) would be

* to completely remove rendering of bays and straits no matter how they are mapped.

Reasoning behind this is that the status quo provides counterproductive mapper feedback (as explained in detail before - see links by @jeisenbe in the initial comments) and we can't find a consensus how to solve that otherwise not providing any mapper feedback on bays and straits would be an improvement.  And since @kocio-pl indicated he considers all geometries for bays and straits to be non-verifiable that might be a solution we can agree on.

imagico:
Regarding

> It would be interesting to see the correlation between the type (node/polygon/relation) and the size of these objects.

I made no thorough analysis of this but based on overall observations it seems that before #3144/#3438 most bay/strait polygons in existence were relatively small.  Polygon mapping afterwards OTOH also and in particular created large polygons.  The free form labeling polygons not following the coastline for most of their outline are mostly a new invention.  The total polygon area of all bay/strait polygons mapped before late 2018 is completely negligible compared to what has been created afterwards.

As @jeisenbe already mentioned most of the larger bay polygons recently added are replacing pre-existing nodes.

Adamant36:
@imagico, as @jeisenbe says "Almost all of the bay and strait nodes I've reviewed have been close enough to the center of the bay or strait that they could be used for an accurate rendering based on the nodes and coastline alone." So why isn't removing rendering multipolygons but keeping it for nodes an option since nodes appear to be mostly mapped accurately?

imagico:
@Adamant36 - for completeness of course just plainly reverting #3144 and #3438 is an option.  But i am pretty sure that for @kocio-pl this is less desirable than accepting this PR (which would keep the rendering of polygons but without use of way_area).

@jeisenbe's remark you quoted however was not about that but about the option to render with automated importance rating based on an analysis of the spatial relationship between the node and the coastline.  This would not solve the problem we are discussing here though (the use of way_area logic without feedback on the polygon geometry) because @kocio-pl indicated he would not want to use such an approach on polygons in https://github.com/gravitystorm/openstreetmap-carto/pull/3750#issuecomment-489446248.

Adamant36:
> for completeness of course just plainly reverting #3144 and #3438 is an option. But i am pretty sure that for @kocio-pl this is less desirable than accepting this PR (which would keep the rendering of polygons but without use of way_area).

Personally, I think only rendering nodes is the best compromise. As it still retains name rendering but gets rid of the clearly contentious less accurate polygons. Ultimately, it doesn't matter if polygons aren't rendered as long as the name still shows up on the map and it's clear from reading the wiki that nodes are the preferred way to map bays anyway. I'm not sure adding mapper feedback on the polygon geometries when way_area is used would work that well considering polygon mapping is explicitly discouraged and the other options, like removing rendering altogether, don't seem viable in my opinion either. So node rendering is the only thing I see that would work.

jeisenbe:
> "why isn't removing rendering multipolygons but keeping it for nodes an option since nodes appear to be mostly mapped accurately?"

The wiki pages for both bay and strait suggest that mapping as an area is an option. Since this style has the difficult responsibility of supporting mappers by providing feedback about correct mapping, we should render both nodes and areas (and linear ways - which were recently added for bays and straits.

I personally would map most straits as nodes or linear ways, and most bays as nodes (except for fjords and similar long, narrow bays, which can be linear ways), but there have always been a minority of these features mapped as areas and this should be supported in this style. 

But the goal would be to show the same rendering whether a feature was mapped as a node or an area, by using the coastline (or ocean shapefiles) to pick the right initial zoom level and text label size, rather than using the way_area of the polygon to show areas sooner but showing nodes only at z14.

When I first made this PR, it was with the mistaken assumption that we could soon show some examples of this type of rendering - but I was hoping that @imagico already had a script written for this, or could do it in an hour's time. 

And I hoped that if we could show labels for large bays at an appropriate zoom level and size, without relaying on mapping polygons and checking their way_area, then @kocio-pl would be happy with accepting this PR, since it would lead to a way forward for rendering other large features that are defined by the coastline, like place=sea and peninsulas.

With the status quo we will have rendering for large bays and straits, but only if mappers change the remaining nodes and linear ways to areas (closed ways or multipolygons). But we will probably never get a good rendering for seas or peninsulas. 

This is particularly a problem for large marine water bodies like the Gulf of Mexico and the Bay of Bengal, which should be place=sea, but will be changed to natural=bay for rendering purposes. I've also seen a number of seas tagged as natural=water for rendering, since we currently are allowing this.

Another problem with the current code is seen in [these](https://github.com/gravitystorm/openstreetmap-carto/pull/3750#issuecomment-484537520) [comments](https://github.com/gravitystorm/openstreetmap-carto/pull/3750#issuecomment-484521868) - because I messed up the last PR which added straits and linear ways:

So the current initial zoom level for both straits and bays mapped as polygons is:
`[zoom >= 0][way_pixels > 3000], 
[zoom >= 14]`
That is, currently they are rendered as early as z0 if they are mapped as polygons that take up more than approximately 3000 square pixels on the screen. But at z14 and higher all bays and straits are rendered no matter what size. Nodes are also rendered at z14 and above."

 The status quo is not a good long-term solution.

kocio-pl:
> >Please show a source for nodes that is both:
> >
> >   not 3rd party
> >   can not be used for verification of an area

> Local knowledge of the mappers.

Thanks for your explanation, but this is not matching any of these conditions:
- this is 3rd party, because you need to rely on somebody else, you can't check it yourself without any other sources of human knowledge, spoken or written
- local knowledge can be used for verification of areas too, if somebody knows more precisely

> We don't rely on 3rd party authorities for OSM data, because OSM data needs to be something that individual mappers can add and correct.

Yes, we do, and the most common cases are admin or reservation boundaries and foreign names. These can be not visible on the ground in many cases (like Berlin name in many languages or the continuous line for French border), but we have them in OSM database because we rely on other sources.

> While the precise location of the center node for Paris is always somewhat debatable, mappers can agree that the node is actually in Paris (even if they think it should be 100 m this way or that way), and that's good enough.

There is analogical reasoning for areas - the borders can be debatable (in fact, many things in OSM are debatable and debated), but it's sure thing that this is an area, not a point, and that's good enough. Plus: you know that a node is inside the area only if somebody told you so, and bay being an area is fundamental feature of this object, you don't have to ask anyone.

But this is theory and I know practical examples where it's clear that area tagging is proper, like Port Phillip Bay (this was a direct reason for me to propose it in #3144).

> This is particularly a problem for large marine water bodies like the Gulf of Mexico and the Bay of Bengal, which should be place=sea, but will be changed to natural=bay for rendering purposes.

How do you know what is a sea or bay? The definition is imprecise, but we have a lot other imprecise definitions, including park/garden, path/footway, restaurant/fast_food etc. It's up to the mapper to decide.

> And since @kocio-pl indicated he considers all geometries for bays and straits to be non-verifiable that might be a solution we can agree on.

This is misunderstanding of my position on this. That's probably because I was pushing the reasoning of nodes as the best solution for bays to see how far does it go and found that this position implies big verifiability problem. I will try to speak directly now to make it clear what I actually think.

I have only asked how do you deal with node for area violating verifiability rule to make a stress test of this assumption. I was interested if there is something unusual to be said, which I might overlook, but nobody ever responded to this. I guess it led to more emotions than I wanted (sorry for that), but it makes me sure that nobody deals with it other than by "who cares?".

We checked verifiability rule down to the core and found that either (a) people are wrong with using nodes for many areas or (b) in OSM rules (including verifiability) are not that strict and people decide in a more complex way.

Which leads me to conclusion that this way or another we will be not 100% sure, but this is not a reason to expect this kind of confidence for every kind of object. Buildings are more reliable than bays or admin borders, but that does not mean we should remove them from rendering (and possibly from database).

I have not responded to all the things being discussed there and maybe I may do it later, but this is not certain thing to happen, sorry.

imagico:
@kocio-pl - as i see it you have a fundamental misunderstanding or disagreement with the [basic principles of OSM](https://wiki.openstreetmap.org/wiki/How_We_Map), in particular:

* the nature and the role of the verifiability principle, in particular how convergence of independent observations and measurement errors relate to verifiability.  Sometimes this seems to go down to the point that you reject the concept of an inter-subjectively observable reality.
* the fundamental role local knowledge and _original research_ (in Wikipedia-speech) play in OSM.

In any case this is not the place to discuss the fundamental values of OSM obviously and i would appreciate if you would refrain from making claims about agreement on those things ("We checked verifiability rule down to...") where there clearly is none as well as from making assumptions about the understanding of others of things ("but it makes me sure that nobody deals with it other than by who cares?") based on your own views of things.

I completely understand that this is not purely a one sided problem and i (and others) have not managed to explain our reasoning in a way you can understand.  But given i have really extensively written about these fundamental questions of OSM addressing many of the misunderstandings you continue to present as arguments in discussion here there is not really much more i can do.

Regarding the concrete matter here - i would summarize your position as follows:

* You are mostly satisfied with the status quo, in particular with the fact that it strongly encourages representing bays and straits with polygons independent of their specific nature and gives mappers direct control over position and size of the label without providing feedback on the polygon geometry.
* You are strictly against any change that deprives mappers of direct control over label position and size of bays/straits represented with polygons by removing way_area logic or basing labeling decisions on the spatial relationship with the coastline.

Which leaves (other than just maintaining the status quo which as @jeisenbe said is not a good long term solution - even when seen independent of the mapper feedback aspect) two options:

* adding mapper feedback on the polygon geometries when way_area is used.
* to completely remove rendering of bays and straits no matter how they are mapped.

Both would ultimately be solutions to the _mutual dissatisfaction of everyone_ - but as i see it that is all the options that are left.

kocio-pl:
Thanks for the concise summary, it helps me to discuss this way.

> I completely understand that this is not purely a one sided problem and i (and others) have not managed to explain our reasoning in a way you can understand. But given i have really extensively written about these fundamental questions of OSM addressing many of the misunderstandings you continue to present as arguments in discussion here there is not really much more i can do.

There is one thing you did not do until now, even if I asked few times: you can answer if you see the problem with mapping area objects as nodes as problem with verifiability rule (and if yes, whayt it might be exactly). For me it's a fundamental question.

> Which leaves (other than just maintaining the status quo which as @jeisenbe said is not a good long term solution

It's fine for me to get back to this problem in the mid term (like dozen months or years) and revisit it.

> You are mostly satisfied with the status quo, in particular with the fact that it strongly encourages representing bays and straits with polygons independent of their specific nature and gives mappers direct control over position and size of the label without providing feedback on the polygon geometry.

I am satisfied only with bays and straits as polygons exactly because their nature is area. With nodes and lines I feel disappointed and still think about them.

I think direct control is a good thing as long as one makes proper mapping - we do it for every other object type and this is called feedback and this is why OSM is about people tagging things, not automatic interpretation of data. Nodes and lines give even more control over label positioning and shape, which would be OK in itself for me, yet this is not accurate representation, but this has nothing to do with control.

> the nature and the role of the verifiability principle, in particular how convergence of independent observations and measurement errors relate to verifiability. Sometimes this seems to go down to the point that you reject the concept of an inter-subjectively observable reality.

Inter-subjective can be a problem if people disagree with indirect observations. I do not dispute buildings for example, because it's directly observable.

I believe this is perfectly verifiable area mapping for - say - Gulf of Guinea, as shown earlier. Mapper comes to the local people and ask them:
- what exact name has this bay
- what is the most northern point where it ends
- what is most southern point where it ends

There are some other simple helper questions - if there are some islands inside or the outside of it, is this a bay or a sea and so on, but nothing complicated or unusual. It's not guaranteed that Jane Doe will know it, so one can ask some other local person.

Do you see some problem with verifiability of this?

imagico:
As said many times now - this is not the place to discuss the fundamental values and principles of OSM.  I have on several occasions stated my position both specifically on the mapping of bays and straits and on the representation of geometries in OSM in general.  Statements i have made are for example

* polygons are a way to geometrically define a two dimensional entity through (and only through) an explicit delineation of its one dimensional boundaries.
* for the decision what kind of feature to use to represent a certain real world feature in the OSM database mappers should put mapping and data maintenance efficiency above perceived desires of data users.  The main criterion should be how to most efficiently represent verifiable information on the feature in question without storing either redundant or non-verifiable data.
* there is no principal connection between the nature of a real world object and how it can or should be represented in OSM above the mapping efficiency criterion previously mentioned.
* straits between concave coasts are one dimensional entities, they have a width but they have no length.
* the verifiability of a node location representing a feature exclusively depends on if multiple independent placements of the node converge to a single location.  This is a completely scale independent problem meaning variance of different placements can be anywhere from less than a meter to hundreds of kilometers.  This has no bearing on the principal verifiability.

I consider most of these statements pretty self evident but i invite anyone who disagrees to challenge them.  But not here please - all of this has no bearing on the what we discuss here.  Here we should not look at how mappers should map based on the subjective preferences of anyone, we should look at how mappers do map (preferably minus the influence we exercise to steer mappers into changing this already).

And i like to point out once more:  We do not currently despite taking into account the geometry of polygons representing bays and straits (via way_area) to decide on starting zoom level and size of labels provide any feedback to mappers on the geometries of those.

kocio-pl:
I have tried to get the fundamental geometry answers on this on the list and on the comment section on your blog, with no reaction. This set of propositions is the first time I remember which deals with these problems and I will try to comment them, thanks.

This PR really does not seem to be the best place to discuss general issues. It proposes very detailed solutions which are not agreed, so I will close it now. The rest should be discussed on the lists and in the issue tickets before any action is proposed. Thanks @jeisenbe for trying to solve some problems, but we clearly have some groundwork to do first.

jeisenbe:
OK, I guess I'll open a new PR to add the outlines and a separate PR
to render lakes from z5 only?

kocio-pl:
Yes, please. In a matter of fact I consider such complex PRs to be a problem and prefer smaller ones, which are easier to test and apply.

jeisenbe:
@kocio-pl - in issue #2068 it was clear that @imagico opposed PR #3144 (which rendered bay areas sooner) but the PR was merged without discussion only 1 day after it was opened.

Perhaps we can reconsider whether a lack of consensus for the original changes in #3144 should mean that the change should be reverted, as in #3955

pnorman (commented):
I agree with these changes, but would like to see a rebase before giving an approving review.

pnorman:
@jeisenbe would you have a chance to rebase this PR?

pnorman (approved):
I figured out how to resolve the conflicts, it wasn't too bad

imagico (approved):
Checked that this works as intended.

As already indicated above in https://github.com/gravitystorm/openstreetmap-carto/pull/3750#issuecomment-484093119 i support this.

This also resolves #4546.

pnorman:
The entire water names section of code could use cleanup.

PR #3754, opened 2019-04-20
Add text label rendering for natural=bay mapped as linear way

jeisenbe:
- Render name labels for natural=bay mapped as linear ways, using same style as natural=strait

## Explanation
- Currently we render name labels for natural=bay mapped as an area or as a node. However, there are almost 400 bays mapped as linear ways, especially in areas with fjords such as Norway, Greenland and Chile. Many of these bays are long, narrow and curving, so it is reasonable that some mappers like to map these as linear ways.
- Recently we added rendering for straits mapped as linear ways in PR #3733 - this new PR will complete the rendering for bays as well, using the same style as that used for straits in the former PR
- 99% of the bays mapped as ways could be rendered as early as z13, but the initial zoom level is set at z14 to be consistent with the current rendering of bay nodes. 

Chile: https://www.openstreetmap.org/#map=14/-42.2631/-72.4470
Before z14
z14-before-chile

After z14
z14-fiordo-cahuelmo
After z15
z15-fiordo-cahuelmo

Norway: https://www.openstreetmap.org/#map=14/59.0483/5.6642
Before z14 - straits render (mapped as linear ways), but not the bay
z14-norway-before
After
z14-svinavika-bay-after

### DIFF
--- a/project.mml
+++ b/project.mml
@@ -2150,7 +2150,7 @@ Layer:
             CASE WHEN tunnel IN ('yes', 'culvert') THEN 'yes' ELSE 'no' END AS int_tunnel
           FROM planet_osm_line
           WHERE (waterway IN ('river', 'canal', 'stream', 'drain', 'ditch', 'wadi')
-                 OR "natural" = 'strait')
+                 OR "natural" IN ('bay', 'strait'))
             AND (tunnel IS NULL or tunnel != 'culvert')
             AND name IS NOT NULL
           ORDER BY COALESCE(layer,0)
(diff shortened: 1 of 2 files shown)


matkoniecz:
Sorry for a long time without review. For now I created https://josm.openstreetmap.de/ticket/17727#ticket (JOSM validator complaint is not a blocker, but maybe there will be some useful comment there)

matkoniecz:
Thanks, and again sorry for a long wait. 

BTW, iD and JOSM now accept linear `natural=bay` though it may require updating to version more recent than released to a general public.

PR #3763, opened 2019-04-24
Road width at z20 adjusted

sommerluk:
- Making the roads code z20-ready
- Larger service roads at z20
- Larger motorways at z20
- The other roads types without changes. It seemed to me that only motorway and service roads are in current master at z20 significantly smaller than on the ground, so only changing these types.

Before:
a1

After:
a2

Before:
b1

After:
b2

Before:
c1

After:
c2

Before:
d1

After:
d2

Before:
e1

After:
e2

### DIFF
--- a/roads.mss
+++ b/roads.mss
@@ -193,6 +193,8 @@
 @road-width-z18:                  8.5;
 @service-width-z18:               8.5;
 @minor-service-width-z18:         4.75;
+@footway-width-z18:               1.3;
+@cycleway-width-z18:              1;
 
 @motorway-width-z19:             27;
 @motorway-link-width-z19:        16;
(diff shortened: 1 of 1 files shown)


imagico:
All roads are drawn in larger-than-real width at low zoom levels and all are drawn less wide than their real width at the highest zoom levels.  Both is intended and necessary if you render roads in a nominal width indicating the road type.

The transit from wider to less wide happens at different zoom levels depending on road type and latitude.  In most cases between z16 and z19, for the smaller large roads, in particular _tertiary_ it can also be at z20 at low latitudes.  Actual ground width rendering without optimizing for a specific latitude would only be possible with variable width rendering as discussed in #1853.

z20 means 0.15m per pixel at the equator which based on current values means:

class | width (pixel) | ground width z20 @ 0° in m
-------|-----------------|--------------------
motorway to tertiary | 27 | 4.05
motorway to tertiary link | 16 | 2.4
residential | 17 | 2.55
living-street | 17 | 2.55
pedestrian | 17 | 2.55
road | 11 | 1.65
service | 11 | 1.65
minor-service | 5.5 | 0.825

The biggest potential for improvements in the road width design at the moment without adding technical complexity would be smoothing out the non-uniform width progression of the various road types.  For most road types for example the ground width stays the same or almost the same as you zoom in from z16 to z17 because the width in pixel doubles or almost doubles but then from z17 to z18 the ground width halves because the width in pixel stays nearly the same for most road types between those zoom levels.

matkoniecz:
I agree that further improvements are possible (@imagico, this comment may be worth opening an issue), but this PR improves situation a bit for z20 so I will merge it.

sommerluk:
Here the pixel width and the ground width change on various zoom levels to compare:

Screenshot_20190810_122648

PR #3771, opened 2019-04-30
Reuse sql statements for turning-circle and roads layers

daveol:
This makes sure the casing and fill queries are the same for both
turning-circle-{fill,casing} and roads-{fill,casing} layers

### DIFF
--- a/project.mml
+++ b/project.mml
@@ -641,7 +641,7 @@ Layer:
     <<: *extents
     Datasource:
       <<: *osm2pgsql
-      table: |-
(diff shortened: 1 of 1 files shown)


project.mml
```diff
@@ -820,122 +825,15 @@ Layer:
       # It is necessary to
```
imagico: This comment should be moved to the actual SQL code.

daveol:
I moved the comment, and amended my previous commit.

PR #3781, opened 2019-05-11
Use ST_PointOnSurface for building labeling

### LINKED ISSUE #3780: Double rendering name  for some features on buildings
Probably regression after #3712, which was replacing centroid algorithm for amenities from the one used by current Mapnik to PostGIS `ST_PointOnSurface`.

### Expected behavior
When building contains some feature we render, the name should be rendered only for this feature.

### Actual behavior
A name can be rendered both for the feature and for the building in different places if the shape is non-trivial.

[Example](https://www.openstreetmap.org/relation/2965344) with `amenity=library`+`building=*`+`name=*`:

Screenshot_2019-05-10 OpenStreetMap Carto — Kosmtik

### Proposed solutions

Possible solutions might be:
1. Remove selecting buildings for name rendering when they share the geometry with amenity tag.
2. Pass the building name rendering also to PostGIS.

pnorman:
Fixes #3780
Tested locally where I found some differing points.

### DIFF
--- a/project.mml
+++ b/project.mml
@@ -2094,10 +2094,11 @@ Layer:
       table: |-
         (SELECT
             name,
-            way,
+            ST_PointOnSurface(way) AS way,
             way_area/NULLIF(POW(!scale_denominator!*0.001*0.28,2),0) AS way_pixels
           FROM planet_osm_polygon
-          WHERE building IS NOT NULL
+          WHERE way && !bbox!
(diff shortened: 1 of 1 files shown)


imagico:
@pnorman - do you want to do a bugfix release with this and #3774?

pnorman:
Yep, I want to get 4.21.1 out

Issue #1981, opened 2015-11-20
Add rendering for healthcare=* tags


morray:
Although the healthcare= tags is still only at about 10% usage of amenity=doctors, it is wothwhile to include it. One advantage is that via the healthcare:speciality=* and healthcare:speciality:*=yes tags can add further information missing in the amenity tagging.

matkoniecz:
Please, be more precise - which tags you propose to display?

See also https://github.com/gravitystorm/openstreetmap-carto/blob/master/CARTOGRAPHY.md#the-mapper-feedback-loop

morray:
Hi matkoniecz,

as a start I think 
tags | occurrences 
healthcare=doctor |  4158
healthcare=dentist | 849
healthcare=centre | 654
healthcare=clinic | 459
should be rendered using the same icons as the corresponding amenity tags

additionally 
healthcare=physiotherapist | 1057
healthcare=alternative | 664
might be interesting because of their frequency for dedicated symbols. The rest healthcare=* should get a generic icon

Concerning the specialities: 
healthcare:speciality=general
healthcare:speciality:general =yes
----
healthcare:speciality=ophthalmology		
healthcare:speciality:ophthalmology = yes
----
healthcare:speciality=gynaecology
healthcare:speciality:gynaecology = yes
----	
healthcare:speciality=internal
healthcare:speciality:internal = yes
----
healthcare:speciality = orthopaedics
healthcare:speciality:orthopaedics =yes
----
healthcare:speciality = otolaryngology	
healthcare:speciality:otolaryngology = yes
---
healthcare:speciality = paediatrics
healthcare:speciality:paediatrics = yes

should add a detail to the doctor symbol.

matkoniecz:
And which ones would not be covered by already rendered amenity=doctors, amenity=dentist etc?

matkoniecz:
> should add a detail to the doctor symbol.

I consider it highly dubious that it is possible to find icon that would have a clear meaning, would not result in confusion and be readable (14x14 px, single color).

morray:
Of course these are covered by amenity=doctor and dentist now. But especially for the doctor healthcare is much cleaner, since it puts the type of doctor in the speciality tag. And the detail could be as big as the icon as now done for the dentist.

matkoniecz:
> Of course these are covered by amenity=doctor and dentist now.

[...]

PR #3784, opened 2019-05-19
Add healthcare tag to lua transformations as a polygon key

### LINKED ISSUE #3639: Healthcare tags do not render on closed ways without area=yes

jeisenbe:
Fixes #3639
Related to #1981 and #2938

- Add "healthcare" to list of keys imported as polygons (when mapped as a closed way)
- This will make it possible to render healthcare objects properly, after the next database reload

- N/A: does not change rendering currently

EDIT: this should be merged to the `schema_changes` branch, not to `master`. Not sure if I can change this?

### DIFF
--- a/openstreetmap-carto.lua
+++ b/openstreetmap-carto.lua
@@ -14,6 +14,7 @@ local polygon_keys = {
     'building',
     'building:part',
     'harbour',
+    'healthcare',
     'historic',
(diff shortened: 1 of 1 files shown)


pnorman:
Merged in 09eefd90

PR #3785, opened 2019-05-19
Add boundary= aboriginal_lands, national_park, and protected_area to polygon keys

### LINKED ISSUE #1141: The boundary=national_park tag is not rendered when tagged on a closed way (not a relation)

jeisenbe:
Fixes #1141

- Add boundary=aboriginal_lands, boundary=national_park, and boundary=protected_area to polygon_values in lua file
- Adding these features to the local polygon values list in openstreetmap-carto.lua will allow these features to be imported as polygons when they are mapped as closed ways, so that they will render properly. Currently only those mapped as boundary relations or multipolygons are imported and rendered, or closed ways tagged "area=yes"

This should be merged to the schema_changes branch so that it can be deployed when the database is reloaded


matkoniecz:
It seems that rebase is missing or Github is making some mistake - see list of commits

pnorman:
This came up with a few of them - for work on the schema change stuff you want to start with a checkout of `schema_change`, not of `master`.

To fix it I took the relevant commits, rebased to remove the commits between them and v4.21.1, then rebased those commits onto schema_change.

jeisenbe:
Sorry! I’ll check for that next time

pnorman:
> * Adding these features to the local polygon values list in openstreetmap-carto.lua will allow these features to be imported as polygons when they are mapped as closed ways, so that they will render properly. Currently only those mapped as boundary relations or multipolygons are imported and rendered, or closed ways tagged "area=yes"

I'm not sure on this. If we wanted to change this, wouldn't it also apply to other `type=boundary` relation-only objects?

jeisenbe:
> wouldn't it also apply to other type=boundary relation-only objects?

The wiki pages and taginfo data for boundary=aboriginal_lands, boundary=national_park, and boundary=protected_area show that these features can be tagged on closed ways, not only on relations. 

I thought that boundary=administrative is supposed to be tagged on relations, not on closed ways. But according to the current wiki page this is possible: https://wiki.openstreetmap.org/wiki/Tag:boundary%3Dadministrative

"Because ways have a maximum number of nodes, you are not required to loop around a whole administration with a single way to form a closed area (although this is a good idea for very small administrative areas)"

If this statement is correct, I could add another PR or a new commit to this PR that would address boundary=administrative, or just import the whole boundary key as polygons when mapped as closed ways.

The other commons of boundary=* are 
- "boundary=political", 
- "boundary=postal_code", 
- "boundary=landuse", 
- "boundary=marker", 
- "boundary=lot",
- "boundary=census"
- 'boundary=forest_compartment'
etc.

It looks like we won't be rendering any of these. But only boundary=marker (which should only be mapped as a node) would be a problem if imported as a polygon

pnorman:
I'm iffy on this change, but need to sit down and think about why.

jeisenbe:
I've updated the branch. @pnorman have you had any time to think about this? 

I think we should support closed ways for protected areas and national parks, or else the long-standing wiki documentation that suggests this mapping method needs to be changed.

jeisenbe:
@pnorman - had you had a chance to think about this one?

jeisenbe:
@pnorman are you available to review this? Does anyone else have thoughts about importing this `boundary=` features as polygons, and whether to include `boundary=administrative` closed ways as well?

pnorman:
Including boundary=administrative on ways would go against long-standing practice, so I'm against that. To me, it's always been type=boundary, boundary=* relations.

jeisenbe:
OK. 
Are you in favor of the current changes in this pr?
boundary=aboriginal_lands, boundary=national_park, and boundary=protected_area

jeisenbe:
@pnorman and @matkoniecz - what are your thoughts about this PR, which does not address administrative relations?

I just edited a few protected_area relations near my home town, and many are mapped as closed ways. 

These are rendered if they are tagged leisure=nature_reserve because leisure=* closed ways are imported as polygons, but if they are boundary=national_park or =protected_area they are not rendered. 

I believe we should fix this inconsistency.

pnorman:
> OK.
> Are you in favor of the current changes in this pr?
> boundary=aboriginal_lands, boundary=national_park, and boundary=protected_area

No. My view is that these should be on relations, where type=boundary will cause them to be processed as boundaries, not on ways.

jeisenbe:
That view is not consistent in the wiki documentation and current use in Taginfo:

https://wiki.openstreetmap.org/wiki/Tag:boundary%3Daboriginal_lands:
Nodes 4
Ways	427 (closed ways 328 http://overpass-turbo.eu/s/PIt)
Relations	520
"Use boundary=aboriginal_lands to mark the jurisdiction area.
Usually this tag is added to a relation of type=boundary, but sometimes it is applied to an area (closed way) or a relation with type=multipolygon."

https://wiki.openstreetmap.org/wiki/Tag:boundary%3Dprotected_area
Nodes	1 833	
Ways	64 211 (closed ways 52864 http://overpass-turbo.eu/s/PIu)
Relations	22 311
"For small protected areas, draw a closed way to define the area.
Larger or more complex protected areas are usually mapped as relations with type=boundary. Some mappers use type=multipolygon instead."

https://wiki.openstreetmap.org/wiki/Tag:boundary%3Dnational_park
Nodes	15	
Ways	10 721  (closed ways 4969 http://overpass-turbo.eu/s/PIw)
Relations	3 095
"The minimum requirements are exemplary for the Snowdonia National Park
name=Snowdonia National Park
boundary=national_park
protect_class=2
area=yes[1]
[... other useful tags for protected areas in general, see boundary=protected_area ...]
↑ As of 2018-05, area=yes is necessary to display a simple closed polyline that is not a type = boundary relation on the OSM standard map."

So all 3 mention that closed ways are an option (the last is implied by suggesting adding area=yes), and in fact ways are more common for 2 out of the 3 features. There are twice as many boundary=protected_area closed ways than relations.

Why should mappers need to convert a simple way into a relation with 1 member or add `area=yes` to get these to render?

jeisenbe:
We are currently only rendering `boundary=` on closed ways when `area=yes` is added. As quoted above this has even been suggested on the Tag:boundary=national_park page, which is inconsistent with our goals. Another option to "tag for the render" is to add `leisure=nature_reserve` to a `boundary=protected_area` closed way.

There are 
* 278 boundary=national_park closed ways with `area=yes` (http://overpass-turbo.eu/s/PIz) 
* 1456 boundary=protected_area closed ways with `area=yes` (http://overpass-turbo.eu/s/PIA)
* 115 boundary=aboriginal_lands closed ways with `area=yes` (http://overpass-turbo.eu/s/PIz)

And with leisure=nature_reserve
* 27125 boundary=protected_area closed ways with `leisure=nature_reserve` http://overpass-turbo.eu/s/PIB
* 1513 boundary=national_park closed ways with `leisure=nature_reserve`http://overpass-turbo.eu/s/PIC

In the past it was stated that we should not render a feature if adding `area=yes` was required, instead we needed to wait for database reload (e.g. heathcare=*, allotments=*). And certainly it is against our goals to encourage mappers to add `leisure=nature_reserve` to `boundary=national_park` to get it to render.

Based on our policy, we should remove rendering of boundary=national_park, and boundary=aboriginal_lands for consistency, if this PR is rejected. 

I am willing to consider this as an alternative - and I believe @imagico was already in favor of removing protected areas rendering in favor of more clearly showing landuse/natural areas.

Adamant36:
> Based on our policy, we should remove rendering of boundary=national_park

National park is kind of a contentious tag that is used in cases it should't be to make currently none rendered protected area classes and other things render. So, that would be a good thing IMO.

kennykb:
> > OK.
> > Are you in favor of the current changes in this pr?
> > boundary=aboriginal_lands, boundary=national_park, and boundary=protected_area
> 
> No. My view is that these should be on relations, where type=boundary will cause them to be processed as boundaries, not on ways.

@pnorman, I'm sorry that this is a long message. We're dealing with a complex issue here, and your repeated one-line statements such as the one above (or an earlier one where you simply shared words to the effect that you had a bad feeling about it), don't really provide a framework to address your concerns. I'd appreciate if you could read through this message and correct my reasoning.

I've done enough map rendering by now to understand why administrative boundaries need a particular treatment. I intend to discuss below the reasons that they do, and the information that is garnered from the ways, and from the relations.  The other `boundary` types, for the most part, are a different animal, as I try to explain. It's possible that `national_park`, `protected_area` and possibly `aboriginal_lands` are therefore abuses of `boundary=*`. Nevertheless, they are really well established tagging at this point, and I'm extremely reluctant to ask mappers to retag everything yet again!

The classic `boundary` relation is `boundary=administrative`. Its purpose is to render governmental cadastre. The ordinary expectation is that a land area will divide into a set of regions, sharing boundaries. A _way_ that is marked `boundary=administrative admin_level=4` will typically either be shoreline, or will be shared between two relations representing the administrative regions, at least one of which is at level <=4, that it divides. Rendering will often be based on the ways, rather than the relations, in order to avoid drawing the border twice. 

In particular, drawing the border of the (multi)polygons will ordinarily draw the shared border in opposite directions, because the renderer walks polygons according to the right-hand rule. If different administrative levels are represented by different dash patterns, as is common, then the dash patterns will be obscured or even spoilt by the overdrawing.

The boundary _relations_ are then used to give the topology, so that a data consumer can look at them and distinguish the inside and outside of an administrative region. This information is necessary if regions are to be shaded, for label placement, for statistical analysis, and so on.

Ordinarily, you'll have an isolated administrative region only in particular cases. It happens often in New York for Villages and Hamlets, which are often isolated and not surrounded by other districts at `admin_level=8`. Towns and Cities (and Indian Reservations), all `admin_level=7`, by contrast, tile the entire state; every point in the state is in exactly one town, city or Indian Reservation. Likewise, counties, `admin_level=6`, tile the entire state, although there is a broken hierarchy; a couple of Cities cross county lines.

_Side note:_ As it happens, I haven't checked whether there are villages and hamlets around that are represented by single closed ways, rather than having boundary relations. Promoting those to boundary relations would certainly be a task for a mechanically assisted edit! (I _never_ do full-up automated edits, everything is checked by eyeball and the JOSM validator before upload.) I had not recognized until fairly recently that having a closed way in place of a boundary relation is a data integrity issue for these; it seemed to me that the way could still be rendered linearly (and would not be shared with a neighbour, so there would be no dash-pattern confusion), and the fact that the way is closed would give the topology, so thanks for the heads-up!

By contrast, `boundary=national_park` and `boundary=protected_area` are not ordinarily objects that are expected to tile an area closely. It's common for national parks, nature reserves, and other areas to be isolated; in fact, it would be rather unusual to see two national parks sharing a common border. Lesser protected areas often will share borders - the Adirondack Mountains are a crazy quilt - but that can be handled readily by making multipolygons that incorporate the shared ways. While initial imports didn't do so, when I've had occasion to adjust these boundaries, I've been conflating common boundaries into shared ways as I go. (At present, I don't see a compelling reason to go through the thousands of protected areas that I curate to identify and conflate all the shared boundaries; rendering and topology mostly work fine as it is.)

In other words `boundary=national_park` and `boundary=protected_area` ordinarily are something that a mapper would expect to work like the common run of (multi)polygons. Rendering issues involving double-drawing of shared borders would be expected to be uncommon.

`boundary=aboriginal_lands` lies somewhere in between.  In some jurisdictions, it's simply an identification of land ownership, restrictions on land use, and perhaps a secondary government.  In others, it bounds a domestic dependent nation, and is expected to work like an administrative boundary.  For instance, as I mentioned before, in New York, all land is part of a City, a Town, or an Indian Reservation. In addition to having the `aboriginal_lands` relation, I'd also include the Indian Reservations as level-8 administrative boundaries, for the purpose of clean border rendering.  

_Another side note:_ Akwesasne is a unique case where the claim of the Six Nations of the Haudenosaunee crosses the international border between the US and Canada. While neither of the larger powers recognizes Haudenosaunee autonomy, the Six Nations have from time to time issued passports, fielded Olympic teams, and performed similar minor functions of a sovereign state. There is at present no passport control for Akwesasne inhabitants within its borders. Attempts to institute such controls have been made multiple times by both the larger powers. The result has generally been violent protest. I'd tag this particular area as a _de facto_ `admin_level=3`, given that it is in practice a domestic dependent nation of two larger nations simultaneously.

So, to my mind what's important is:

 - _shared_ borders between administrative regions _must_ be discrete single ways; no border between two administrative regions should be represented twice or it will be drawn twice, messing up the rendering.  
 - administrative regions also must have topology - so `boundary` (or, equivalently, `multipolygon`, I suppose) relations must be provided at least for ones that do not comprise a single closed way.
 - administrative enclaves cannot be represented by single closed ways, but require boundary relations, because the border way is shared with the enclosing region.
 - an administrative region that is isolated (e.g., many Villages in New York, many islands) _could_ logically be represented by a simple closed way. By convention, we do not do so, because the case of enclaves is otherwise difficult to distinguish from the case of islands.

Aboriginal lands should be treated the same way as administrative regions if they fit into the jurisdiction's administrative structure and are expected to share boundaries. There's no convincing reason why New York's Indian Reservations could not, for instance, be surrounded by `boundary=administrative admin_level=8` ways and then participate in a `boundary=aboriginal_lands` relation.

National parks and protected areas are not ordinarily expected to share boundaries, so the problem of double-rendering is less of an issue. For clean rendering, shared borders _should_ be conflated where possible, which then requires the creation of multipolygons (or boundary relations, but mappers have consistently used multipolygons for these). I'm tentatively against elevating the _should_ in the last sentence to a _must_, simply because existing renderings do not generally adopt a visual style that is confused by adjacency.

@pnorman, have I adequately summarized your concerns? Given this summary, is there a way to go forward without retagging thousands of protected areas? 

Does it suffice to have the following rules? _(a)_ Two regions of the same type (administrative, national park, aboriginal lands, protected area), if they have a border in common, _should_ have that border represented by a shared way. _(b)_ If a shared way is used, the two regions _must_ be represented by `boundary` or `multipolygon` relations. _(c)_ In the data model, rule _(b)_ is impossible to break except in the case where one such region is an enclave within another; in this last case, the relation is required anyway, to distinguish the properties of the inner region from properties that belong to the boundary.

I could see these rules extending to any type of area where the rendering emphasizes the border, rather than simply filling the interior or providing an interior highlight on the border. There may be other objects (some land uses or land covers, perhaps) that fit the pattern; I've not done any detailed investigation into how others render these. In any case, it's a separate discussion.

pnorman:
> That view is not consistent in the wiki documentation and current use in Taginfo:

...

> So all 3 mention that closed ways are an option (the last is implied by suggesting adding area=yes), and in fact ways are more common for 2 out of the 3 features. There are twice as many boundary=protected_area closed ways than relations.

Given these numbers I guess we should add it. I'm not particularly overjoyed about the current representation in the data.

jeisenbe:
>  I'm not particularly overjoyed about the current representation in the data.

Agreed. So while we should add these features to the polygon list in lua, we might also want to reconsider whether they should all be rendered.

Adamant36:
> we might also want to reconsider whether they should all be rendered.

I think the same arguments could be made for removing rendering of national parks and protected areas as there is for the removal bays and similar things. No clear boundaries, rendering encourages inaccurate tagging and mapping etc etc.

imagico:
Regarding the change - i think that is fine although i also don't really like the ambiguities in tag use this would support.

Regarding stopping rendering of protected areas and other non-general-administration boundaries - yes i think from a cartographic viewpoint it would be good to drop those, for reasons explained in https://github.com/gravitystorm/openstreetmap-carto/pull/3887#issuecomment-531549827 and https://github.com/gravitystorm/openstreetmap-carto/pull/3887#issuecomment-531557484.  But this is very different from the considerations for bays and straits.

For bays and straits all arguments and evidence speak against the current rendering, for protected areas it is a matter of cartographic preferences and of making best use of the limited map real estate to show map users the most meaningful data in a way that is well readable.  Even if objective arguments can be made (like i linked to above) it is ultimately a matter of setting priorities.  

@kennykb - i think most of what you write about is outside the scope of this issue tracker.  But what we definitely do not want to do here is pushing mappers to map things in a way that is convenient for us.  If mappers map things consistently in a certain way that is difficult for us to properly render it is our task to deal with that, not to try nudging the mappers to change their habits to make things easier for us.

jeisenbe:
This is the one schema change that we still need to discuss before the next
release

My thought is that we should import closed ways as polygons with these tags
because mappers are often using them in this way.

If there is consensus that this mapping method is a serious problem, it
would be better to remove rendering of protected areas entirely, rather
than requiring mapper to add ‘area=yes’ or create single-object
multipolygon relations to force rendering.

In that case we still ought to import these closed ways properly in case
other adaptations of this style wish to render such features.

(I have been guilty of using both those options myself as a mapper, and one
of my changesets was criticized for creating multipolygon relations out of
a single closed way.)

pnorman:
> Agreed. So while we should add these features to the polygon list in lua, we might also want to reconsider whether they should all be rendered.

I believe they should continue to be rendered, and don't see any connection between that and this issue.

PR #3786, opened 2019-05-19
Import "allotments" as a polygon key

jeisenbe:
Related to #432

- Import "allotments" as a polygon feature by adding it to the list of polygon_keys in the lua transformations file
- This will import closed ways tagged as "allotments=plot" properly as a polygon rather than a linestring, which makes it possible to render the ref numbers for plots (see #432) or plot outlines
- Current rendering is unchanged

Should be merged to the `schema_changes` branch, which will be merged to master at the time of the next database reload

### DIFF
--- a/openstreetmap-carto.lua
+++ b/openstreetmap-carto.lua
@@ -9,6 +9,7 @@ local polygon_keys = {
     'abandoned:landuse',
     'abandoned:power',
     'aeroway',
+    'allotments',
     'amenity',
     'area:highway',
     'building',


matkoniecz:
It seems that rebase is missing or Github is making some mistake - see list of commits

jeisenbe:
This is because the schema_changes branch is still behind master, and I
created my branches from tree up-to-date master branch

jeisenbe:
Any objections to this? I have not been able to confirm the total number of landuse=allotments which have plots specified, but there are a total of 13 ,266 individual plots tagged, and 8, 166 (61.56%) have a ref=*

So I suspect there are several hundred landuse=allotments areas which have been subdivided in this way and could have the ref rendered, after the database reload.

PR #3790, opened 2019-05-28
Add retail color fill for shop=mall areas

### LINKED ISSUE #2702: Render non-building shop=mall like landuse=retail
Based on discussion in #2700 we could consider rendering shop=mall like landuse=retail in case it is used on features without a building tag and has no landuse tag.  This would essentially affect malls that contain outdoor areas (which is not uncommon in countries with appropriate climate).

It would improve mapper feedback because you would see the mall as mapped independent of labels meaning there would be feedback on accuracy of the geometry.

This would probably only affect about 10 percent of the malls currently mapped - others have a building or a landuse tag.

jeisenbe:
- Add retail color fill for shop=mall areas
- But do not rendering shop=mall fill when underground=yes

## Explanation
Currently most malls are mapped with the same geometry as a building, or are tagged with landuse=retail. However, some malls area mapped as areas but lack either of these tags. These malls are labeled based on the size of the area, but there is no mapper feedback from this style about the shape of the area.

It would improve mapper feedback to render these areas, because you would see the shape of the mall as mapped, meaning there would be feedback on accuracy of the geometry.

However, there are some malls that are located underground, beneath streets, for example near subway stations. These should not be shown with the landuse fill color, because the mall does not represent the landuse of the area in this case.

**Before**
Citylink Singapore (location=underground) - unchanged
https://www.openstreetmap.org/#map=17/1.2923/103.8546
z17-citylink-label-only

z19 Esplanade Xchange - before (located underground but not tagged)
https://www.openstreetmap.org/#map=19/1.2931/103.8555
z19-Esplanade-Xchange-before

Diamond Run, z17 - mall mapped as an area; above ground
https://www.openstreetmap.org/#map=17/43.58043/-72.96340
z17-diamond-run-before

Sicilia Outlet Village - outdoor outlet mall, with detailed mapping
https://www.openstreetmap.org/#map=17/37.573296/14.480978
z17-sicilia-before

**After**
Citylink Singapore (location=underground) - unchanged
z17-citylink-label-only

z19 Esplanade Xchange after - fill rendered
z19-esplanade-xchnange-mall-after-19:1 29310:103 85549

Diamond Run z17
z17-diamond-run-mall-fill

Sicilia Outlets z17
z17-sicilia-fill

### DIFF
--- a/landcover.mss
+++ b/landcover.mss
@@ -355,6 +355,7 @@
   }
 
   [feature = 'landuse_retail'],
+  [feature = 'shop_mall'],
   [feature = 'amenity_marketplace'] {
     [zoom >= 8] {
       polygon-fill: @built-up-lowzoom;
--- a/project.mml
+++ b/project.mml
@@ -90,7 +90,7 @@ Layer:
       table: |-
         (SELECT
             way, name, religion, way_pixels, is_building,
-            COALESCE(aeroway, amenity, wetland, power, landuse, leisure, man_made, "natural", tourism, highway, railway) AS feature
+            COALESCE(aeroway, amenity, wetland, power, landuse, leisure, man_made, "natural", shop, tourism, highway, railway) AS feature
           FROM (SELECT
               way, COALESCE(name, '') AS name,
               ('aeroway_' || (CASE WHEN aeroway IN ('apron', 'aerodrome') THEN aeroway ELSE NULL END)) AS aeroway,
@@ -103,6 +103,7 @@ Layer:
                                                     'allotments', 'forest', 'farmyard', 'farmland', 'greenhouse_horticulture',
                                                     'recreation_ground', 'village_green', 'retail', 'industrial', 'railway', 'commercial',
                                                     'brownfield', 'landfill', 'construction', 'plant_nursery', 'religious') THEN landuse ELSE NULL END)) AS landuse,
+              ('shop_' || (CASE WHEN shop IN ('mall') AND (tags->'location' NOT IN ('underground') OR (tags->'location') IS NULL) THEN shop ELSE NULL END)) AS shop,
               ('leisure_' || (CASE WHEN leisure IN ('swimming_pool', 'playground', 'park', 'recreation_ground', 'garden',
                                                     'golf_course', 'miniature_golf', 'sports_centre', 'stadium', 'pitch', 'ice_rink',
                                                     'track', 'dog_park', 'fitness_station') THEN leisure ELSE NULL END)) AS leisure,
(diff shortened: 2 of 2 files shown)


jeisenbe:
The fill color will only render from z10, as with amenity=marketplace, and it will be the gray built-up-area fill color until z13, as now with landcover=retail and amenity=marketplace

I could add [zoom >= 10 ] to the CSS code in landcover.mss for clarity, if that's preferred.

matkoniecz:
I still feel that explicit `landuse=retail` is preferable, but I am not going to strongly protest.

jeisenbe:
I agree, and I will always add landuse=retail to malls that I map.

But we need to show something, since the size and zoom level of the text
label depends on the way_area

I tried just an outline, but the results are poor.

Though I would consider an outline for underground malls, if needed.

imagico:
I would be fine with other rendering styles as long as there is feedback on the geometry.  And it has to work design wise and i agree with @jeisenbe that outline rendering does not.

Regarding not rendering the fill but rendering the label for underground malls - this does not really work for me.  The only other features where we specifically do not render them based on `location=underground` we don't render at all, not even with a label.  Your 'Esplanade Xchange' example IMO shows that this is quite confusing - the map user has no idea what this label refers to.  The approach to use here in my eyes would be similar to underground parkings to develop a way to highlight the entrances.

jeisenbe:
> The approach to use here in my eyes would be similar to underground parkings to develop a way to highlight the entrances.

Do you think we should keep rendering the name label as well?

imagico:
> Do you think we should keep rendering the name label as well?

Not in the middle of the polygon in a size based on way_area.

The map user tends to try connecting the label to the context around it and if that does not work because the context is on ground level, for example a road as in the case you showed and the mall is underground this is very confusing IMO, in particular since the label is very generic and does in no way imply the type of feature.

jeisenbe:
I've removed the rendering of text labels for underground malls.

z17 before - Citylink Singapore (location=underground)
https://www.openstreetmap.org/#map=17/1.2923/103.8546
z17-citylink-label-only

z17 after
z17-citylink-after-no-label

z18 after
z18-citylink-after-no-label

jeisenbe:
@imagico, the changes that you requested were made. Can this be merged?

imagico:
I merged this but notice comment by @matkoniecz about having a distinction between landuse=retail and shop=mall in rendering being a good idea.  It might be worth thinking about a minor distinction, maybe only at very high zoom levels, highlighting shop=mall as different from landuse=retail other than just the label.  Maybe that could be even something that could be used for hinting underground malls as well.

That would be a worthy task for someone who is into delicate design tuning and experimentation.

PR #3816, opened 2019-06-29
python3 support

### LINKED ISSUE #3815: scripts/**.py did not work with python3
python2 has been no longer supported, so these scripts need to work with python3.

scripts/generate_road_colours.py
scripts/generate_shields.py
scripts/indexes.py

samaiyou:
Fixes #3815

- 
python3 support.
changed from iteritems to items on dict(), lxml.etree.tostring returns to bytes, so convert to utf-8.


scripts/generate_road_colours.py
```diff
@@ -32,7 +32,10 @@ def rgb_error(self):
 
 def load_settings():
     """Read the settings from YAML."""
-    return yaml.load(open('road-colors.yaml', 'r'))
+    with open('road-colors.yaml', 'r') as fd:
+        y=yaml.load(fd, Loader=yaml.SafeLoader)
```
StyXman: Nitpick: could you add spaces around `=` like in the rest of the file?

matkoniecz: `autopep8` should catch such formatting, and fix most of them

matkoniecz:
Sorry for a long delay!

It seems to me that `scripts/generate_road_colours.py` used to output elements in constant order and now it is doing in random order.

Though maybe earlier version worked by accident thanks to an undocumented behaviour?

meased:
This behaviour is coming from the YAML loader, which has changed since Python 2:
https://github.com/yaml/pyyaml/issues/110

This Python 3 version is listing out items in the order they are listed in the YAML file. The Python 2 version was essentially randomized (I believe it would hash the keys and then sort the hashes, but don't quote me on that). I would argue that this version is correct and the previous Python 2 version is "wrong".

Sjord:
Beside the ordering, the colors are also different for me. I have no explanation for this.

* python 2: @trunk-low-zoom-casing: #cf6649;
* python 3: @trunk-low-zoom-casing: #d1684a;

Sjord:
I have figured out why the colors are different:

    delta_l = (line_colour_info.end_l - line_colour_info.start_l) / colour_divisions

In Python 2, this is an integer division: 20/3 = 6.
In Python 3, this is a float division: 20/3 = 6.667

pnorman:
After changing the scripts the generated files need to be regenerated.

pnorman:
Could you also change indexes.py to call python3

scripts/generate_road_colours.py
prmtl: If you want to enforce pep8 (any codestyle), check should be added to CI (.travis.yml)

prmtl: But of course this should not be part of this PR.

mboeringa:
> I have figured out why the colors are different:
> 
> ```
> delta_l = (line_colour_info.end_l - line_colour_info.start_l) / colour_divisions
> ```
> 
> In Python 2, this is an integer division: 20/3 = 6.
> In Python 3, this is a float division: 20/3 = 6.667

@Sjord 

Have you considered using the **decimal** module of Python?:
https://docs.python.org/2/library/decimal.html

pnorman:
> Have you considered using the **decimal** module of Python?:
> https://docs.python.org/2/library/decimal.html

I don't see that decimal would give us any practical advantages over floats here.

mboeringa:
> I don't see that decimal would give us any practical advantages over floats here.

The calculations would be consistent between Python 2 and 3. But of course, if you are dropping Python 2 support entirely, which is logical given the end of life status, then yes, there isn't much to gain.

prmtl:
@mboeringa It is not needed to use `decimal` module to have consistent division. Just use `__future__` module and replace `/` with `//`:
```python
# python 2 & 3
> from __future__ import division
> 20 // 3
6
```

pnorman:
I've addressed the PR comments, but still need to get Travis working with python3

pnorman:
> This Python 3 version is listing out items in the order they are listed in the YAML file. The Python 2 version was essentially randomized (I believe it would hash the keys and then sort the hashes, but don't quote me on that). I would argue that this version is correct and the previous Python 2 version is "wrong".

Either would be correct, but the problem is that Python 3 prior to Python 3.7 doesn't have a fixed ordering. I think the options here are to require Python 3.7 or load into an OrderedDict.

jeisenbe:
@pnorman is this nearly ready to release?

pnorman:
> @pnorman is this nearly ready to release?

It needs reviewing.

matkoniecz:
This PR changes 665 files due to removal of ".0" in generated SVGs. I am considering adding special additional code to preserve it for now, to avoid changing so many files and potentially drowning other changes.

pnorman:
> It needs reviewing.

To be clear, it needs reviewing by someone other than me since I took the PR over

scripts/generate_road_colours.py
```diff
@@ -32,7 +32,10 @@ def rgb_error(self):
 
 def load_settings():
     """Read the settings from YAML."""
-    return yaml.load(open('road-colors.yaml', 'r'))
+    with open('road-colors.yaml', 'r') as fd:
+        y = yaml.load(fd, Loader = yaml.SafeLoader)
```
Sjord: Couldn't you return immediately here, without assigning to `y` in between?

scripts/generate_shields.py
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
 
 # Generate highway shields as SVG files in symbols/shields.
```
Sjord: Line 5 says:

    from __future__ import print_function

but this can be removed now. In other files as well.

.travis.yml
```diff
@@ -11,14 +11,16 @@ addons:
...
+    - postgresql-9.4-postgis-2.4
     - osm2pgsql
+    - python3-pip
+    - python3-setuptools
 env:
   - CARTO=0.18.0 MAPNIK='3.0.0 3.0.12'
 install:
   - npm install carto@$CARTO
-  - pip install --user colormath
+  - pip3 install --user colormath
```
Sjord: Shouldn't this also install PyYAML and lxml?

.travis.yml
```diff
@@ -11,14 +11,16 @@ addons:
     packages:
     - lua5.1
     - libxml2-utils
-    - python-yaml
-    - postgresql-9.4-postgis-2.3
+    - python3-yaml
+    - postgresql-9.4-postgis-2.4
```
Sjord: Does this package exist? In bionic there seems to be `postgresql-10-postgis-2.4`.

pnorman:
I'm going to close this. The original author never addressed the comments, and while I attempted to do some of them, there are still some left. I'd rather split the python 3 upgrade to multiple PRs, and have been doing so.

PR #3825, opened 2019-07-23
Drop 'highway=steps' from z=13

### LINKED ISSUE #3808: Drop highway=steps from z13
I think this was simply forgotten about when highway=footway was removed from z13 in #3467.

As is, highway=steps barely appear as more than a smudge of red pixels on that zoom level. See e. g. https://www.openstreetmap.org/#map=13/49.6363/7.8418&layers=N (which on that note still has a lot of staircases not mapped, it's been on my todo list for a good while...)

I can't think of a good usecase where you'd want to look at steps at z13, especially because they tend to be very short ways (the staircase in Rockenhausen is one of the longest in Germany period).

btwhite92:
Fixes #3808

### Changes proposed in this pull request
- Drop steps from z13

### Description

Footway rendering was dropped with #3467, steps should likewise be dropped. For future reference, if more cleanup/adjustments happens to z13:
>When taking the current status of z13 otherwise as a given removing steps would be an improvement. But i like to point out that #3467 (and to some extent already #747) left z13 and the whole zoom level progression in a fairly inconsistent state with several contradicting design ideas being thrown together. I have my doubts that in the long term a z13 design and a zoom level progression with no footways (and accordingly no steps) at z13 makes a lot of sense. #3467 did not really present a vision how this is supposed to work outside European urban settings and even there it has its problems.

>So yes, removing the steps from z13 would be consequential in the current situation but it should not prejudice the long term strategy for z13 and around.

_Originally posted by @imagico in https://github.com/gravitystorm/openstreetmap-carto/issues/3808#issuecomment-503034693_

#### Before

steps-ex-old

([Vitória, Espírito Santo, Brasil](https://www.openstreetmap.org/#map=13/-20.3090/-40.3167)) - thanks @LucFreitas for the excellent example location

#### After

steps-ex-new

### DIFF
--- a/roads.mss
+++ b/roads.mss
@@ -646,7 +646,7 @@
         }
       }
       #tunnels {
-        [zoom >= 13][access != 'no'],
+        [zoom >= 14][access != 'no'],
         [zoom >= 15] {
           line-width: @steps-width-z14 + 2 * (@paths-background-width + @paths-tunnel-casing-width);
           [zoom >= 15] { line-width: @steps-width-z15 + 2 * (@paths-background-width + @paths-tunnel-casing-width); }
--- a/roads.mss
+++ b/roads.mss
@@ -118,7 +118,6 @@
 @track-width-z13:                 0.5;
 @track-grade1-width-z13:          0.5;
 @track-grade2-width-z13:          0.5;
-@steps-width-z13:                 0.7;
 
 @secondary-width-z14:             5;
 @tertiary-width-z14:              5;
@@ -127,6 +126,7 @@
 @pedestrian-width-z14:            3;
 @road-width-z14:                  2;
 @service-width-z14:               2;
+@steps-width-z14:                 0.7;
 
 @motorway-width-z15:             10;
 @motorway-link-width-z15:         7.8;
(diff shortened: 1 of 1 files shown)


kocio-pl:
Thanks for taking care of this. It looks to me like pretty obvious change at this moment. Unfortunately my testing environment is broken currently and I have not too much time to fix it, so I hope that maybe somebody else would be able to test and merge it.

jeisenbe:
I've reviewed this PR. The code works as expected. Here are some test images from Singapore. It's on the equator, so z13 looks like z12 in northern Europe. There are a number of steps in the two hilly parks here:

z14: Before/After (unchanged):
z14-singapore-no-steps-after

z13: Before
z13-singapore-with-steps-before

z13: After
z13-singapore-no-steps-after

Certainly this change works with the scheme after #3467 - although I would also like to reconsider some of the changes from that PR.

It does not fix the problems with the red dotted footways and paths on z14 or 15 - these might be improved by changing the rendering style for paths and footways, perhaps using something similar to the German style.

However, overall I would approve the PR as a reasonable step to take at the moment.

imagico:
Thanks - and sorry for the delay.

As already said - this is consequential to #3467 - but z13 is in any case after #3467 quite out of balance and not really usable outside urban environments so requires overall work - which might ultimately also mean getting back footway rendering (and consequently steps).

Issue #3830, opened 2019-08-01
Check for drift in generated MSS and SVGs


pnorman:
We check that indexes.sql is up to date with
https://github.com/gravitystorm/openstreetmap-carto/blob/080b8eaaef81fc98b77d9c58505dec62a0e195be/.travis.yml#L37

We should check other generated files.

ref #3816

jeisenbe:
What are the others besides #3831?

`.kosmtik-config.yml`?

`/symbols/shields`?

pnorman:
The shields are the one I'm not sure how to do. 

> .kosmtik-config.yml

This file doesn't exist in the repo

PR #3831, opened 2019-08-02
Check the generated road colours for drift

pnorman:
This catches when the generated road colours file has drifted out of sync with the source.

Part of #3830, but doesn't address all the generated code.


sommerluk:
Looks good. Thanks.

PR #3832, opened 2019-08-02
Render rock

### LINKED ISSUE #2616: Tag natural=rock (node and area) is not rendered
I've changed incorrectly mapped [natural=cliff ](https://wiki.openstreetmap.org/wiki/Tag:natural%3Dcliff ) area to [natural=rock](https://wiki.openstreetmap.org/wiki/Tag:natural%3Drock) and it is not render at all.

https://www.openstreetmap.org/way/263176747

I thought it is only connected to area, but later on I've find that even node is not rendered.

http://www.openstreetmap.org/node/3841474372

JOSM uses following icon: https://josm.openstreetmap.de/export/11963/josm/trunk/images/presets/misc/rock.svg

Sjord:
Fixes #2616

- Rename rock_overlay to bare_rock_overlay, so that rock_overlay can be used to:
- render natural=rock. The used overlay is rock 8 as mentioned in #2616. Rock is rendered differently than bare_rock, especially for cases when a notable rock feature is surrounded by bare rock.

[This rock](https://www.openstreetmap.org/way/375390567) at zoom level 16:
Before
rock-before

After
rock-after


imagico:
Thanks for the PR.

At a quick look i see two problems with it:

* I don't see the reasoning behind the design of the pattern.  It is more prominent and less homogeneous than the bare_rock pattern but i am not sure what this is supposed to communicate to the map user.  And given the difficulties we have with the bare_rock pattern in terms of readability of other features drawn on top (like paths) a significantly stronger pattern is likely to cause problems.
* At the moment i see no consistent use of `natural=rock` on polygons for something different than `natural=bare_rock` in the database or in the example you showed.  Historically `natural=rock` has been used as a synonym for `natural=bare_rock` - including several large imports.  And there is plenty of data that follows this use, though `natural=bare_rock` is clearly the only generally accepted tagging for this now.  There might be other uses of `natural=rock` that are meant to indicate something distinctly different - locally or by individual mappers.  But i don't really see an overall agreement on this being reflected in the data.

matkoniecz:
Also, OSM Wiki claims that

> natural=rock describes a notable rock feature or small group of rocks, attached to the underlying bedrock, mainly as a single node element.

> On the contrary, single boulders not attached to the bedrock are better tagged as natural=stone.

> For larger bedrock areas you may use natural=bare_rock. 

https://wiki.openstreetmap.org/wiki/Tag:natural%3Drock (in other  words for areas `natural=bare_rock` should be used)

Adamant36:
> https://wiki.openstreetmap.org/wiki/Tag:natural%3Drock (in other words for areas natural=bare_rock should be used)

That was the main reason I stopped working on it before taking it to the PR phase. Maybe one of the maintainers could close the original issue as won't fix after this is rejected so it's clear there isn't a chance of rendering being added in the future (which I think is for the better).

Sjord:
From the comments I gather that the change this PR proposes is not necessarily seen as an improvement. I agree that there are some problems with it, and I'll close this PR.

The distinction between the tags `rock` and `bare_rock` areas is small and unclear. Even though the proposed pattern makes clear that `rock` is different from `bare_rock`, it does not show in what way it differs, which is hard to do because the tags themselves are ambigious. It is not immediately clear what the proposed pattern should signify.

I think there are two possible ways forward:
* Not render `natural=rock`, and communicate to mappers that should tag areas as `bare_rock` instead of `rock`.
* Render `rock` the same as `bare_rock`. This admits that the distinction between the two is negligable, but at least renders the areas currently mapped as `rock`.

Please let me know what you think, either here or in #2616.

matkoniecz:
I opened https://josm.openstreetmap.de/ticket/18031#ticket "Complain about natural=rock used for large areas and suggest natural=bare_rock instead"

PR #3833, opened 2019-08-03
Add tourism=yes as a linestring exception

### LINKED ISSUE #3812: Tag:tourism refrains turntable from rendering
### Expected behavior
Tag `tourism` should not affect the rendering of railway turntables.

### Actual behavior
I'll use this turntable (https://www.openstreetmap.org/way/444942851/history) as example.

Changeset [71486896](https://www.openstreetmap.org/changeset/71486896): before adding `tourism=yes`.

turntable_1

Changeset [71487174](https://www.openstreetmap.org/changeset): after adding `tourism=yes`.

turntable_2

I also tried removing `name` and `wikimedia_commons`: no effect.

jeisenbe:
Fixes #3812 

- Add `tourism=yes` as a linestring exception in `openstreetmap-carto.lua`

### Explanation:
- Currently all features in the key `tourism=*` are imported as polygons when mapped as closed ways. This means that features tagged with the property `tourism=yes` are always imported as polygons when mapped as a closed way with a key such as `railway=` or `highway=*` which is normally a linestring. This tag is intended to say that a feature is "interesting to tourists", so it should not define an object as an area. By adding this tag to the list of linestring exceptions, it will not be imported as a polygon, unless the main feature tag on the object is a polygon feature.

This PR is submitted to the branch schema_changes, since it will require database reload to take effect.

I've tested the rendering by downloading the turntable mentioned in the issue and adding back tourism=yes, then reloading the database with this branch. Also note that the building in the northwest (upper left) is building=* with tourism=yes and this still renders properly

https://www.openstreetmap.org/map=18/-23.776/-46.301

**Before:**
turntable-tourism-yes-before

**After:** (plus database reloaded)
turntable-tourism-yes-after

### DIFF
--- a/openstreetmap-carto.lua
+++ b/openstreetmap-carto.lua
@@ -39,6 +39,7 @@ local linestring_values = {
     man_made = {embankment = true, breakwater = true, groyne = true},
     natural = {cliff = true, tree_row = true, ridge = true, arete = true},
     power = {line = true, minor_line = true},
+    tourism = {yes = true},
     waterway = {canal = true, derelict_canal = true, ditch = true, drain = true, river = true, stream = true, wadi = true, weir = true}
 }
 


pnorman:
:+1: in principle, I haven't tested it yet.

mboeringa:
To me, this is trying to solve the wrong issue. 

Rendering turntables as a line feature is the real problem: If the turntable is digitized as closed way, it should additionally get **area=yes** and then ideally openstreetmap-carto should render it is a polygon, not line.

A turntable is not a railway track after all, the line actually defines or represents the entire _**surface area**_ covered by the turntable, but openstreetmap-carto doesn't render it like that. It is the outline of the object, not the object itself.

I would suggest to switch rendering of turntable from line to polygon, and than possibly augment Wiki and iD presets to add **area=yes** by default to closed ways with **railway=turntable**.

jeisenbe:
The tag tourism=yes is used on a large number of different features,
many of which could be rendered incorrectly if we continue to consider
it a polygon feature tag, rather than recognizing it as a property tag
which does not imply a particular geometry.

> Rendering turntables as a line feature is the real problem

I considered changing the rendering of turntables. This would also
require a schema change, to import these features as polygons rather
than linestrings. I believe this should be a separate PR.

jeisenbe:
Re: > Wiki and iD presets to add area=yes by default to closed ways
with railway=turntable

This is a bit off-topic for this PR, but I believe this suggestion
would be incorrect. The wiki suggests the tag railway=turntable should
only be used on nodes and areas (e.g. closed ways), not on linear
ways: https://wiki.openstreetmap.org/wiki/Tag:railway%3Dturntable

If this is the case, area=yes is not required. This tag is necessary
for features that can be mapped both as linear ways and as areas - in
that case, a closed way is potentially ambiguous in meaning.

However, I have not carefully checked current usage to confirm that
mappers are representing these features only as areas rather than as
linear ways.

mboeringa:
> The tag tourism=yes is used on a large number of different features, many of which could be rendered incorrectly if we continue to consider it a polygon feature tag, **rather than recognizing it as a property tag which does not imply a particular geometry**.

OK, I now get where you want to go with this, and I think I agree with this, as long as the implementation follows what you wrote in the quote below, and specifically the bold part, so features having a valid polygon tag next to the **tourism=yes** tag are still imported as polygon after this change.

> By adding this tag to the list of linestring exceptions, it will not be imported as a polygon, **unless the main feature tag on the object is a polygon feature.**

> Rendering turntables as a line feature is the real problem

> I considered changing the rendering of turntables. This would also require a schema change, to import these features as polygons rather than linestrings. I believe this should be a separate PR.

Fair.

mboeringa:
By the way, just a side note, but of all the **tourism=x** tags, I think **tourism=yes** is the least useful and informative. It doesn't explicitly imply being a visitable attraction like **tourism=attraction**, or **tourism=museum**.

The tag may well be used on almost anything related to tourists, even hotels or appartments could well use it. As such, I personally ignore this tag for rendering.

jeisenbe:
I agree that this tag has a vague meaning, and we shouldn’t consider it for
rendering. Even tourism=attraction is has a quite broad meaning, being used
as both a property (“this feature is interesting to tourists”) like
tourism=yes, and as a feature (“this is a tourism attraction, and I don’t
know if a more specific tag for it”).

Ideally we should also stop rendering tourism=attraction after some more
specific tags are developed.

mboeringa:
> Ideally we should also stop rendering tourism=attraction after some more specific tags are developed.

Looking at my rendering results, I don't entirely agree with this conclusion. Although there are a few mishaps, overall, I think **tourism=attraction** is used quite well. I have rendered objects tagged with this tag for large parts of the globe, and most of the time, I do think they are valid tags that would classify as a "touristic attraction" in the sense that most people would see it like that.

The most mis-use I have seen is by e.g. climbers or other extreme sports. E.g. I have seen some climbers tag every scalable peak (**natural=peak**) in a rock climbing area with an additional **tourism=attraction** tag. _That is just plain nonsense IMO_. Luckily, such mis-use is rather rare.

To me, a **tourism=attraction** is something that anybody, _irrespective of age, ability, gender, race_ or whatever, should be able to visit.

If your grandma of 75 can't visit it, it is not a **tourism=attraction** ;-)

matkoniecz:
> To me, a tourism=attraction is something that anybody, irrespective of age, ability, gender, race or whatever, should be able to visit.

I just want to note that I really disagree with this. But discussing it should happen elsewhere (OSM Wiki/tagging mailing list/etc).

mboeringa:
> I just want to note that I really disagree with this. But discussing it should happen elsewhere (OSM Wiki/tagging mailing list/etc).

We don't need to discuss it, it was just a personal opinion voiced by me. I fully understand not everybody will agree, and I understand there are notable exceptions even to my own 'rule', e.g. while I don't see the average pensioner of 80 bungee jump from a bridge, that still is likely a tourism=attraction.

PR #3838, opened 2019-08-10
Use “cache-feature: true” for better performance with attachments

### LINKED ISSUE #3756: Set “cache-features: true” in project.mml?
Set `cache-features: true` in project.mml?

The option `cache-features: true` makes Mapnik cache SQL results between attachments within the same layer.

Currently, we do not use it. Would it be for performance reasons a good idea to enable it on all layers that have attachments?

Mapnik does not use  `cache-features: true` because of [possible memory issues](https://github.com/mapnik/mapnik/issues/657), but memory issues [does not seem to be likely](https://github.com/mapnik/mapnik/issues/3977#issuecomment-417923243).

sommerluk:
Fixes #3756

- Add `cache-feature: true` for all layers that use attachments to get better performance. We have quite a lot of them…

Test rendering with links to the example places: Should trigger no rendering change.

This PR would need performance testing, but I am not able to do that.

Another question: What about adding `cache-feature: true` to _all_ layers? If this would be performance-neutral, it might be worth to do so, to prevent adding new attachments while forgetting about adding `cache-feature: true`?

### DIFF
--- a/project.mml
+++ b/project.mml
@@ -80,6 +80,7 @@ Layer:
           ORDER BY way_area DESC, feature
         ) AS landcover_low_zoom
     properties:
+      cache-features: true
       minzoom: 5
       maxzoom: 9
   - id: landcover
@@ -135,6 +136,7 @@ Layer:
           ORDER BY way_area DESC, feature
         ) AS features
     properties:
+      cache-features: true
       minzoom: 10
   - id: landcover-line
     geometry: linestring
@@ -243,6 +245,7 @@ Layer:
           ORDER BY COALESCE(layer,0), way_area DESC
         ) AS water_areas
     properties:
+      cache-features: true
       minzoom: 0
   - id: ocean-lz
     geometry: polygon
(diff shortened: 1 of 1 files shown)


pnorman:
> Another question: What about adding `cache-feature: true` to _all_ layers? If this would be performance-neutral, it might be worth to do so, to prevent adding new attachments while forgetting about adding `cache-feature: true`?

I would expect it to boost RAM usage if done to all layers, but I have no idea how much.

I'm loading up a DB to run low zoom tests on, but if they're inconclusive, then I'll need to do high-zoom testing too. My current test setup use mapproxy, so because it creates the map object each render that flaw might overwhelm any performance differences, and I might need to go to renderd

sommerluk:
For comparison: https://github.com/sommerluk/openstreetmap-carto/tree/cache02 provides code that uses `cache-features` for _all_ layers.

sommerluk:
Thanks for testing!

Issue #3347, opened 2018-08-16
Multipolygon relations 'Island within a hole' not rendering as expected


DaveF63:
Hi

### Expected behaviour
https://www.openstreetmap.org/#map=18/51.02480/-3.93326&layers=C
### Actual behaviour
https://www.openstreetmap.org/#map=18/51.02480/-3.93326

https://wiki.openstreetmap.org/wiki/Relation:multipolygon#Island_within_a_hole

Why isn't OSM-Carto rendering the inner island? I could have sworn it used to.

matthijsmelissen:
This is a data issue: the water area doesn't have a hole in the data (it's not a multipolygon).

SomeoneElseOSM:
@DaveF63 I'm not sure exactly how you'd like @matthijsmelissen to phrase it, but https://www.openstreetmap.org/way/616701768 is an inner to the wood but is not itself a relation.  https://www.openstreetmap.org/way/615985619 is not an inner to the water relation and hence is both "wood" and "water".  This change was made here a while back to accommodate places where those things happen together (think mangroves).  See e.g. https://github.com/gravitystorm/openstreetmap-carto/issues/1754 for some previous discussion.

dieterdreist:
sent from a phone

> On 16. Aug 2018, at 17:56, SomeoneElseOSM <notifications@github.com> wrote:
> 
> This change was made here a while back to accommodate places where those things happen together (think mangroves).


+1, we have to look at the actual data, not whether other renderers show or don’t show the problems. And yes, this style should try to show the problems rather than hiding them in favor of a more pleasant map, because people will fix the problems much more likely if they suffer from them.

kocio-pl:
This is a follow up to this question:

https://help.openstreetmap.org/questions/65303/could-someone-check-this-multipolygon-outerinnerouter-relation

The only problem I see is that when water was properly tagged as a multipolygon with a hole inside, the island has the tree symbols, but not the green background. Which is interesting, but I would probably just tag the forest on the island instead of complicated two part multipolygon with forest-inside-a-forest shape.

matthijsmelissen:
@DaveF63 Happy to expand on my answer.

First: the current tagging indicates that the middle island is both trees and water, i.e. that the trees are growing in the water. Does this describe the situation correctly? If not, the tagging needs changing.

> How comes the island renders the graphic? How come the other renders manage it?

[...]

Issue #1954, opened 2015-11-04
Forest Mapnik Landuse Face Partially Overlying Water Face


PaulColinSimpson:
When looking at area http://www.openstreetmap.org/#map=17/51.72011/-0.01848 you will notice that the water face only partially overlaps the forest face. For some reason the forest landuse colour disappears only revealing the forest symbols which looks strange.

math1985:
This is a tagging error (assuming there are no trees growing in the lake). The forest should be tagged as a multipolygons with the lakes as 'inner' components.

matkoniecz:
See #1754 for more detailed explanation.

PaulColinSimpson:
Many thanks matkoniecz! I much clearer on this now

PR #3851, opened 2019-08-26
Render bare_rock pattern in landcover-area-symbols

### LINKED ISSUE #1473: natural=* covered by freshwater or oceans treated inconsistently?
works nicely:
https://www.openstreetmap.org/way/336816028#map=17/46.18740/8.75470
there is bare rock on both sides of the rivers and also underwater.  The rendering is fine.

The same does not appear to work for natural=bare_rock which is across natural=coastline
and partly underwater - the whole area is rendered as rock and water is not shown
in that place.

What is right? Should we use something like natural=underwater:bare_rock ? Could
the rendering be made consistent whether the water is natural=water or natural=coastline?

### LINKED ISSUE #3707: Render beaches and rocks in the tidal range differently

jeisenbe:
Fixes #3707  - I believe this is the only missing feature that is currently not rendered when in the intertidal range.
Also fixes #1473 which mentioned bare_rock

- <s>Remove pattern rendering for natural=sand</s>
- Render bare_rock pattern in landcover-area-symbols, above water areas

## Explanation
- <s>The tag `natural=sand` is currently rendered with a unique color fill, different than beaches or bare_earth, so the pattern is not needed to distinguish sand from similar areas. Removing the pattern will make the map less busy. This will also make it easier to add a pattern for `natural=dune` areas, if desired.</s>
- <s>Removing the natural=sand pattern will also hide areas tagged natural=sand outside of the coastline. These should be tagged `natural=beach` or `natural=shoal`</s>
- Rendering the `natural=bare_rock` pattern in the `landcover-area-symbols` layer, above water areas, will show areas of rock at the coastline, between the high and low tide lines. This will match with the rendering for `natural=beach` areas which also show the pattern above water. 

(I considered changing the initial zoom level for the bare_rock pattern from z5 to z8 or z10, since at low zoom levels the pattern is usually not visible and mainly adds visual noise without conveying significant information, but I believe the `natural=scree` and `natural=shingle` patterns should also be changed at the same time, so this will be another PR.)

### Horton's Nose - `sand` next to `beach` and a patch of `bare_rock` (on land, and another outside of coastline, upper right)
http://www.openstreetmap.org/#map=17/53.31607/-3.50920
z17 before
z17-hortons-nose-sand-before-17:53 31607:-3 50920
z17 after
z17-hortons-nose-sand-rock-after

### Ty Anquer - `natural=bare_rock` outside and inside of coastline, next to `natural=beach`
z16 before
z16-Ty-Anquer-bare_rock-before
z16 after
z16-ty-anquer-bare_rock-after

z17 before
z17-ty-anquer-bare_rock-before
z17 after
z17-ty-anquer-bare_rock-after

### Pointe de Trefeuntec
z16 before
z16-pointe-de-trefeuntec-bare_rock-before
z16 after
z16-pointe-de-trefeuntec-bare_rock-after

z17 before
z17-pointe-d-trefeuntec-bare_rock-before
z17 after
z17-pointe-de-trefeuntec-bare_rock-after

### Beg an ty-Guard - bare_rock
z16 before
z16-beg-an-ty-guard-bare_rock-before
z16 after
z16-beg-an-ty-guard-bare_rock-after


imagico:
The pattern over water looks different in different examples.  And neither of them works for me as a pattern - the foreground-background contrast is too small for that.

This is technically tricky to tune because the the two versions with different background colors cannot be adjusted independent of each other.  You probably need to use two pattern images - one in the base layer that is covered by water and one in landcover-area-symbols.  This way you could get the same visual results over land while getting a darker pattern over water.

Care also needs to be taken not to incentivize tagging rocky reefs with natural=bare_rock.  Differentiating reef rendering by `reef=*` could prevent that.

The sand change looks fine to me - that is essentially reverting #2746.

jeisenbe:
> You probably need to use two pattern images - one in the base layer that is covered by water and one in landcover-area-symbols. This way you could get the same visual results over land while getting a darker pattern over water.

I don't quite understand how that would work. I can see how it would be easier to get a lighter pattern this way, by adding the two patterns. Would it require transparency?

imagico:
Yes, you could also try other ideas using comp-op but that would probably be less intuitive.  I have not tried this and i don't know if it would actually work as intended.

jragusa:
I notice that the current rendering of `natural=beach` is different to your pictures. Is it normal ?

I agree about the too subtle rendering of `natural=bare_rock`.

Please also fix your addresses, they link to 404 error due to wrong place of "#" (map# → #map). And it's Point of Ayr ;)

imagico:
> I notice that the current rendering of natural=beach is different to your pictures. Is it normal ?

The map on osm.org currently lags significantly behind this style.

jeisenbe:
> Differentiating reef rendering by reef=* could prevent that.

https://taginfo.openstreetmap.org/keys/reef#values - used a little under 4000 times, vs a little under 19,000 reefs mapped as areas.

Main values are coral (3146 times), rock (493 times) and sand (46 times). So `reef=rock` could be rendered in a similar way to `bare_rock`. Could we rendering reef=sand with the sand pattern (as used for beach and shoal) instead, perhaps? Coral reefs could keep the current rendering.

imagico:
I tried and you can get a reasonable color balance while keeping the results over land with a single pattern image by darkening the color and increasing transparency:

rock pattern sample

http://imagico.de/files/rock_overlay.png
http://imagico.de/files/rock_overlay.svg

The problem is that rendering the rock pattern in landcover-area-symbols would mean the pattern is not only rendered over ocean but also over all other water features including water lines.  This creates a significant amount of noise.  Overall there are currently a lot of areas where lakes, glaciers etc. are not cut out of a larger base mapping of bare_rock.  Such rendering would provide feedback for cleaner mapping but it would certainly at first look fairly ugly in a lot of places and the reason for the different rendering results would not always be so clear to mappers.  We have had a similar situation when we introduced the wood/forest pattern in the overlay.

I also produced a pattern that could be used for coral reefs:

http://imagico.de/files/reef_coral.svg

jeisenbe:
> rendering the rock pattern in landcover-area-symbols would mean the pattern is not only rendered over ocean but also over all other water features including water lines.

True, this also happens with the forest/woodland pattern and scrub pattern now. 

Changing the river/stream color to a darker shade of blue, as planned, will help some.

Another option would be to use a less dense pattern for bare_rock. Some other patterns were already tried out in https://github.com/gravitystorm/openstreetmap-carto/issues/2616#issuecomment-445691221 and https://github.com/gravitystorm/openstreetmap-carto/issues/2616#issuecomment-446070453 - but these are probably not an improvement yet.

jeisenbe:
4 new options which are less dense, and mostly smaller pattern size, compared to current pattern. All use the same color and opacity suggested above (opacity 0.3, color is `#8a99a5`)

1) Smaller, evenly spaced
bare_rock-new-1

2) Smaller but closer
bare_rock-new-2

3) Mid-sized, but more space
bare_rock-new-3

4) Even smaller, some space between
bare_rock-new-4

For comparison, here's the modified pattern that @imagico created; the shapes are the same as the current bare_rock pattern, (but opacity is 0.3 and color is `#8a99a5`, as in the others):
rock_overlay

imagico:
> True, this also happens with the forest/woodland pattern and scrub pattern now.

Yes, and i have mentioned that when these were moved to landcover-area-symbols there was confusion because of unexpected rendering results, see #1754, #1796, #1821, #1954, #3347.

But also note these are pictorial symbol patterns with individually identifiable and readable symbols while bare_rock uses a structure pattern that is read and understood by map users based on its overall structure in combination with the base color.  Due to this it is much more confusing seeing such pattern on a different background than with wood/scrub.

kocio-pl:
Hi, @jeisenbe! Please remove the sand part, because it makes this PR more complicated in terms of discussion and these two are only remotely related. 

Currently I don't agree with removing sand pattern (but this is not my final say - we can discuss the dune problem elsewere), while I don't have any opinion on rock rendering (but I don't see anything wrong with it, so I leave it to the interested people to decide).

imagico:
I outlined a technical approach that could solve the drawing order and independent design for water and land problems in #3854.

jeisenbe:
#3854 would change the water layers to use `dst-out` to cut out the shape of water and ocean areas from all the landcover, land patterns, water-lines and land, then use `dst-over` so that water and water patterns are rendered _behind_. This would work very nicely, and would not require changing the current bare_rock pattern on land, as suggested in this PR.

However, we could still consider changing the `bare_rock` pattern, if it's considered to be too busy compared to the `scree`/ `shingle` and `sand` patterns. While the changes in #3854 would allow using different colors and patterns over land and over water, it might still be sensible to use the same svg pattern for `bare_rock` over land and over water, for consistency, and the current pattern may be too solid for use over water?

Should I show some examples with the new options above?

jeisenbe:
Here are tests with the new rock_overlay - same pattern, different color and transparency, which @imagico created above: https://github.com/gravitystorm/openstreetmap-carto/pull/3851#issuecomment-525342666 - (opacity 0.3, color is `#8a99a5`). 

It's certainly improved, but I will also show some other options:

z16 Pasir Putih 
http://www.openstreetmap.org/#map=16/-4.0400/138.947516
pasir-putih-bare_rock5-16:-4 0400:138 947516

z14 Puncak Jaya
http://www.openstreetmap.org/#map=14/-4.0719/137.1774.png
puncak-jaya-bare_rock5-14:-4 0719:137 1774

z17 ty Anquer
http://www.openstreetmap.org/#map=17/48.15064/-4.27294
z17-ty-anquer-bare_rock5-17:48 15064:-4 27294

z17 Pointe de Trefeuntec
A couple kilometers south of ty Anquer
z17-pointe-de-trefeuntec-bare_rock5

(These are with `bare_rock5`, from the most recent commit 5fd8ffd4ce12e86b053a7c71a35c76e03eff3cba)

jeisenbe:
**1) `bare_rock1.svg`** (with opacity 0.3, color `#8a99a5`). 
This pattern is smaller and has some space between each symbol.

z16 Pasir Putih 
http://www.openstreetmap.org/#map=16/-4.0400/138.947516
pasir-putih-bare_rock1-z16

z14 Puncak Jaya
http://www.openstreetmap.org/#map=14/-4.0719/137.1774.png
puncak-jaya-bare_rock1-z14

z17 ty Anquer
http://www.openstreetmap.org/#map=17/48.15064/-4.27294
z17-ty-anquer-barerock1

z17 Pointe de Trefeuntec
z17-point-de-trefeuntec-barerock1

jeisenbe:
**2) bare_rock2.svg** (with opacity 0.3, color `#8a99a5`). 
This pattern is denser than 1) but a little less dense and the elements are a little smaller than the current pattern

z16 Pasir Putih 
pasir-putih-bare_rock2-z16

z14 Puncak Jaya
puncak-jaya-bare_rock2-z14

z17 ty Anquer
z17-ty-anquer-bare_rock2

z17 Pointe de Trefeuntec
z17-point-de-trefeuntec-bare_rock2

jeisenbe:
**3) bare_rock3.svg** (with opacity 0.3, color `#8a99a5`). 
This pattern is larger than the others. I think this makes it too strong; it's hard to see paths or streams.

z16 Pasir Putih 
pasir-putih-bare_rock3-z16

z14 Puncak Jaya
puncak-jaya-bare_rock3-z14

z17 ty Anquer
z17-ty-anquer-bare-rock3

z17 Pointe de Trefeuntec
z17-point-de-trefeuntec-bare_rock3

From commit 7baec1055af10fddd69ada5628a642268027e308

jeisenbe:
**4) bare_rock4.svg** (with opacity 0.3, color `#8a99a5`). 
This pattern is the smallest and slightly more closely spaced than 1), though still not overlapping like the current pattern. 

This is the best-looking in some situations, though it might need to be a little darker. 
I believe 2) is the next best, and then the current pattern is ok too.

z16 Pasir Putih 
pasir-putih-bare_rock4-z16

z14 Puncak Jaya
puncak-jaya-bare_rock4-z14

z17 ty Anquer
z17-ty-anquer-bare_rock4

z17 Pointe de Trefeuntec
z17-pointe-de-trefeuntec-bare_rock4

Tested in commit 3691c7032cf851260a92b710fe81b65b47bed9df

imagico:
I think you should concentrate on one thing in this PR and not mix it with other things.  Changing the bare_rock pattern is separate from how to render bare_rock over water.

When you compare different structure patterns it is often useful to separate differences in pattern weight from differences in pattern structure.  The most reliable way of doing that is to generate for every pattern variant a sequence of different rendering intensities and then move these relative to each other so they match in overall weight.  Technically this is much more efficient to do by compositioning the pattern using imagemagick than by rendering again and again with mapnik.

The reason why the bare_rock pattern is currently heavier than scree/shingle is because bare rock areas are usually 'heavier' (in other words: more serious obstacles, more weighty appearance) than scree/shingle.  Most maps rendering rock and scree use a similar or even stronger difference in weight of rendering styles.

jeisenbe:
The current commit uses the existing pattern, with the new color and transparency, as shown in https://github.com/gravitystorm/openstreetmap-carto/pull/3851#issuecomment-526052372 - but I thought I should try some other options to see if it was easy to improve the rendering.

Since none of the 4 new patterns was clearly an improvement, I agree with keeping the existing pattern.

jeisenbe:
@imagico - are you in favor of this PR in the short-term? I imagine that implementing #3854 will take some time.

imagico:
I would not be in favor of it but i would not block it if others see it positively.

#3854 is not a matter of difficult implementation i think but i am not sure if there is consensus in support of it.

jeisenbe:
>  I am not sure if there is consensus in support of it.

Best way to find out would be to submit a PR. I believe this could be done before changing the river color (or ocean color), and in fact it might be good to discuss first, since it could influence color choices for water areas. Would you be willing to finish the code? I'm willing to do test images and review.

jeisenbe:
Closing for now in favor of the solution proposed in issue #3854. If that solution is rejected we can reconsider this PR.

PR #3856, opened 2019-08-31
Adjust `aeroway=aerodrome` zoom levels

### LINKED ISSUE #2664: Airport names gone above zoom level 13
This airport only has names at
http://www.openstreetmap.org/#map=12/23.5180/119.5891
http://www.openstreetmap.org/#map=13/23.5180/119.5891
When you reach
http://www.openstreetmap.org/#map=14/23.5180/119.5891
and above, the name is gone. Imagine an airport without a name.

### LINKED ISSUE #3809: Make sure features are still queryable when user is told to zoom in first
It doesn't matter what the reason is, just make sure you don't "play
bait and switch" with the user... especially as he has to mouseover the
"Zoom in to query features" button at each zoom level to see if it has
finally changed its name to "Query features".

On https://www.openstreetmap.org/#map=11/25.5688/-80.4893 user sees
plenty of airports.

User clicks the "Zoom in to query features" button... OK, he needs to
zoom in first before he can query "Redland Field".

Finally https://www.openstreetmap.org/#map=14/25.5448/-80.4574 the "Zoom
in to query features" button now is called "Query features", but alas,
it is just at this zoom level that "Redland Field" has disappeared from
the map!

Even if #2664 / #232 are resolved, it is just going to happen again for some
other feature.

All the user knows it he is told to zoom in to query, but then the thing
disappears, "like one big practical joke". He looks over his shoulder to
see if this isn't one more trick his office mate set up on his computer
like the last time he went for coffee break without locking the keyboard.

jeisenbe:
Adjust aerodrome and airport zoom levels

Fixes #2664 and fixes #3809 
Related to issues #1028 and #1143
Follow-up of PR #2674

#### 1) For public aerodromes (`aeroway=aerodromes` with `iata=*` and `icao=*`, and not `access=private`)
- Move initial zoom level for text for airports from z10 to z11. (Icon continues to render from z10).
- Continue rendering icon and text until z16, instead of stopping at z13 as currently.

#### 2) For other aerodromes (`aeroway=aerodromes` lacking `iata=*`, `icao=*` or both, or with `access=private`)
- Increase icon initial zoom level from z11 to z12 
- Move initial text label zoom level from z11 to z13.
- Continue rendering till z17

#### 3) _Optional_: 
- Stop rendering icon and labels earlier (usually no sooner than z15 or z16) if `way_pixels>768,000` (this only works for large `aeroway=aerodrome` mapped as closed ways or multipolygons, however). 

## Explanation
The current initial and final zoom levels for aeroways are z10 and z13. This appears to have been picked to work with large, international airports in the high lattitudes, though even there it's surprising that airports are not rendered at z14. 

In #2664 and #3809  it is noted that aerodrome icons and labels disappear after z13, much sooner than almost any other feature. This is unexpected, and does not work well for average-sized airports and small aerodromes. 

It also does not work well at low latitudes near the equator, and it fails badly with small airstrips and airfields. Such features are rare in Europe, but common in parts of North America, Australia, New Zealand and the whole tropical region.

Also, in #1143 it was noted that too many small aerodromes were previously shown at mid-zoom levels, (z10 to z12). This was partially address in PR #2674 by moving aerodromes with `access=private` or lacking `iata`= (or `icao=`) to z11, however, this did not address any issues at z11 or z12, and it appears not to have considered the issues closer to the equator.

Another issue is that the text label or icon of an airport is often blocked by the name of the town at low zoom levels (z10 to z12), leading to just the text label or just the icon showing. This is particularly a problem for small airfields and airstrips, which show a zoom level before `place=village` currently, though they area usually less significant.

Recently, PR # solved #1028 by stopping area labels based on waypixels for most area features at `way_pixels > 768,000`

Many larger airports are now mapped as areas, but of the 100 aerodromes with an IATA and ICAO code in Germany there are still 11 mapped as nodes, and in Canada there are 78 nodes out of 260, so it's not possible to use `way_pixels` to adjust rendering in general.  *(However, it might be used just to stop rendering large airports a little sooner than others)

However, checking the `way_pixels` of various aerodromes in Canada, Northern Ireland and Indonesia, it's apparent that most should continue rendering till z16 or z17, rather than stopping at z13, if we were to use a `way_pixels < 768,000` cut-off.

See https://github.com/gravitystorm/openstreetmap-carto/issues/1028#issuecomment-526809401: 

I have tested various initial and final zoom levels for rendering airport icons and text labels, so far in Papua, Indonesia (4 degrees south), Prince Edward Island, Canada (46 degrees north) and Northern Ireland (55 degrees north). 

By changing airports (those with IATA codes) to show only the icon at z10, the feature is not too dominating on the map at this zoom level. At z11 larger airports are clearly visible and the text label is not too distracting (though it still may be a problem near the equator).

For smaller aerodromes without IATA codes or private aerodromes, the initial zoom level is changed from z11 to z12, and only the icon is shown at this zoom level, with the text label first shown at z13. This allows `place=village` to be shown 1 zoom level before these minor features have labels. 2 other renders mentioned using z12 or z13 as the initial zoom level in #1143, see https://github.com/gravitystorm/openstreetmap-carto/issues/1143#issuecomment-209842594 and https://github.com/gravitystorm/openstreetmap-carto/issues/1143#issuecomment-252240948

The final zoom level is changed to z16 for aerodromes with an IATA code and z17 without, based on the usual size of these two categories. Some large aerodromes at high lattitudes will not benefit from the icon and text label rendering at z16; this could be solved by checking for `way_pixels`, or we can accept that such airports are rare, and the majority of `aeroway=aerodrome` features will benefit from the rendering at z16 (or z17 for smaller ones).

### Belfast International Airport
https://www.openstreetmap.org/#map=11/54.6500/-6.2272
- Major airport in a rural area far from the city which takes up a large area. Since it's at 55 degrees north, it's almost twice as wide (and 4x the area) on-screen as similar-sized airports at the equator

#### z10 before 
(also visible: Belfast city airport, at right, but text labels is blocked by place label)
(At this zoom level, the text lable is several times larger than the airport, even for this large airport at high latitude)
z10-belfast-airports-before
**z10 after** - symbols still shown, but no text at this zoom level
z10-belfast-international-and-city-after

#### z11 before
z11-belfast-international-before
**z11 after** - small aerodrome to southwest is not shown now, because at this low zoom level 
z11-belfast-international-after

#### z13 before (same after)
(last zoom level where icon and text label are rendered currently)
z13-belfast-international-before

#### z14 before (no label or icon)
z14-belfast-international-before-no-label
**z14 after**
z14-belfast-international-after

#### z15 before - inspector view
(not rendered, even though < 525k `way_pixels`)
z15-belfast-international-before-waypixels-524418
**z15 after**
z15-belfast-international-after

#### z16 before 
(first zoom level where terminal building name is shown)
z16-belfast-international-before-waypixels-2million
**z16 after** - not shown, if `way_pixels` are checked. 
z16-belfast-international-before-waypixels-2million

### Belfast city airport
https://www.openstreetmap.org/#map=11/54.6198/-5.8705
Also an airport with a full-length runway at high lattitude, but it only hits 36,000 way_pixels at z14, so it should render till z16 (576k waypixels). 
* **z10 - see above**
* ** z11-z13** unchanged

#### z14 Before
Only 36k `way_pixels`, but no icon or text label rendering
z14-belfast-city-airport-waypixels-35864
**z14 after**
z14-belfast-city-airport-after

#### z15 After 
(No rendering before)
z15-belfast-city-airport-after

#### z16 After - inspector view
(Only 574k `way_pixels`, so it should be rendered - and this is true for almost all airports at z16)
z16-belfast-city-airport-after-waypixels-574000

### Dunnamanagh Airstrip
https://www.openstreetmap.org/#map=12/54.8889/-7.2669
- This aerodrome is a good example of the many small runways without an IATA or ICAO code, yet it's been mapped as an area with a building, apron and taxiway. Shouldn't render before z12, but can render until z17 without problems (and near the equator z17 would be important)

#### z11 before
- Renders before any villages, label as big as a good-sized town.
z11-dunnamanagh-airstrip-before
(After: not rendered)

#### z12 before
- Label is too prominent compared to villages
z12-dunnamanagh-airstrip-before
** z12 after** - just icon
z12-dunnamanagh-airstrip-after

#### z15 after
- No icon or text label rendering before
z15-dunnamanagh-airstrip-after

#### z17 after
z17-dunnamanagh-airstrip-after-last-zoom-level

### DIFF
--- a/amenity-points.mss
+++ b/amenity-points.mss
@@ -1505,7 +1505,7 @@
       marker-clip: false;
       marker-fill: @airtransport;
       [way_pixels > 48000] { marker-file: url('symbols/aerodrome.15.svg'); }
-      [way_pixels > 19200] { marker-file: url('symbols/aerodrome.18.svg'); }
+      [way_pixels > 192000] { marker-file: url('symbols/aerodrome.18.svg'); }
     }
   }
 
@@ -1526,7 +1526,7 @@
       marker-fill: @airtransport;
       [zoom >= 11] { marker-file: url('symbols/aerodrome.12.svg'); }
       [way_pixels > 48000] { marker-file: url('symbols/aerodrome.15.svg'); }
-      [way_pixels > 19200] { marker-file: url('symbols/aerodrome.18.svg'); }
+      [way_pixels > 192000] { marker-file: url('symbols/aerodrome.18.svg'); }
     }
   }
 
@@ -3236,7 +3236,7 @@
         text-line-spacing: @landcover-line-spacing-size-big;
         text-dy: 11;
       }
-      [way_pixels > 19200] {
+      [way_pixels > 192000] {
         text-size: @landcover-font-size-bigger;
         text-wrap-width: @landcover-wrap-width-size-bigger;
         text-line-spacing: @landcover-line-spacing-size-bigger;
--- a/amenity-points.mss
+++ b/amenity-points.mss
@@ -1488,39 +1488,34 @@
     marker-fill: @airtransport;
   }
 
-  [feature = 'aeroway_aerodrome'][is_airport = 'yes'][zoom >= 10][zoom < 16],
-  [feature = 'aeroway_aerodrome'][is_airport = 'no'][zoom >= 12][zoom < 17] {
+  [feature = 'aeroway_aerodrome']['access' = 'private'][zoom >= 12][zoom < 17],
+  [feature = 'aeroway_aerodrome']['icao' = null][zoom >= 12][zoom < 17],
+  [feature = 'aeroway_aerodrome']['iata' = null][zoom >= 12][zoom < 17] {
+    [way_pixels <= 192000],
+    [way_pixels = null] {
+      marker-file: url('symbols/aerodrome.12.svg');
+      marker-placement: interior;
+      marker-clip: false;
+      marker-fill: @airtransport;
+      [zoom >= 15] { marker-file: url('symbols/aerodrome.15.svg'); }
+      [zoom >= 16] { marker-file: url('symbols/aerodrome.18.svg'); }
+      [way_pixels > 12000] { marker-file: url('symbols/aerodrome.15.svg'); }
+      [way_pixels > 48000] { marker-file: url('symbols/aerodrome.18.svg'); }
+    }
+  }
+
+  [feature = 'aeroway_aerodrome']['access' != 'private']['icao' != null]['iata' != null][zoom >= 10][zoom < 16] {
     [way_pixels <= 192000],
     [way_pixels = null] {
       marker-file: url('symbols/aerodrome.8.svg');
       marker-placement: interior;
       marker-clip: false;
       marker-fill: @airtransport;
-      [zoom >= 11] {
-        marker-file: url('symbols/aerodrome.12.svg')
-      }
-      [is_airport = 'yes'] {
-        [zoom >= 14] {
-          marker-file: url('symbols/aerodrome.15.svg')
-        }
-        [zoom >= 15] {
-          marker-file: url('symbols/aerodrome.18.svg')
-        }
-      }
-      [is_airport = 'no'] {
-        [zoom >= 15] {
-          marker-file: url('symbols/aerodrome.15.svg')
-        }
-        [zoom >= 16] {
-          marker-file: url('symbols/aerodrome.18.svg')
-        }
-      }
-      [way_pixels > 12000] {
-        marker-file: url('symbols/aerodrome.15.svg')
-      }
-      [way_pixels > 48000] {
-        marker-file: url('symbols/aerodrome.18.svg')
-      }
+      [zoom >= 11] { marker-file: url('symbols/aerodrome.12.svg'); }
+      [zoom >= 14] { marker-file: url('symbols/aerodrome.15.svg'); }
+      [zoom >= 15] { marker-file: url('symbols/aerodrome.18.svg'); }
+      [way_pixels > 12000] { marker-file: url('symbols/aerodrome.15.svg'); }
+      [way_pixels > 48000] { marker-file: url('symbols/aerodrome.18.svg'); }
     }
   }
 
@@ -3139,8 +3134,9 @@
     text-placement: interior;
   }
 
-  [feature = 'aeroway_aerodrome'][is_airport = 'yes'][zoom >= 11][zoom < 16],
-  [feature = 'aeroway_aerodrome'][is_airport = 'no'][zoom >= 13][zoom < 17] {
+  [feature = 'aeroway_aerodrome']['access' = 'private'][zoom >= 13][zoom < 17],
+  [feature = 'aeroway_aerodrome']['icao' = null][zoom >= 13][zoom < 17],
+  [feature = 'aeroway_aerodrome']['iata' = null][zoom >= 13][zoom < 17] {
     [way_pixels <= 192000],
     [way_pixels = null] {
       text-name: "[name]";
@@ -3161,28 +3157,53 @@
           text-dy: 13;
         }
       }
-      [is_airport = 'yes'] {
-        [zoom >= 14] {
-          text-size: @landcover-font-size-big;
-          text-wrap-width: @landcover-wrap-width-size-big;
-          text-line-spacing: @landcover-line-spacing-size-big;
-          text-dy: 11;
-        }
-        [zoom >= 15][is_airport = 'yes'] {
+      [zoom >= 15] {
+        text-size: @landcover-font-size-big;
+        text-wrap-width: @landcover-wrap-width-size-big;
+        text-line-spacing: @landcover-line-spacing-size-big;
+        text-dy: 11;
+        [zoom >= 16] {
           text-size: @landcover-font-size-bigger;
           text-wrap-width: @landcover-wrap-width-size-bigger;
           text-line-spacing: @landcover-line-spacing-size-bigger;
           text-dy: 13;
         }
       }
-      [is_airport = 'no'] {
-        [zoom >= 15] {
-          text-size: @landcover-font-size-big;
-          text-wrap-width: @landcover-wrap-width-size-big;
-          text-line-spacing: @landcover-line-spacing-size-big;
-          text-dy: 11;
+      text-face-name: @oblique-fonts;
+      text-halo-radius: @standard-halo-radius;
+      text-halo-fill: @standard-halo-fill;
+      text-placement: interior;
+    }
+  }
+
+
+  [feature = 'aeroway_aerodrome']['access' != 'private']['icao' != null]['iata' != null][zoom >= 11][zoom < 16] {
+    [way_pixels <= 192000],
+    [way_pixels = null] {
+      text-name: "[name]";
+      text-size: @landcover-font-size;
+      text-wrap-width: @landcover-wrap-width-size;
+      text-line-spacing: @landcover-line-spacing-size;
+      text-fill: darken(@airtransport, 15%);
+      text-dy: 9;
+      [way_pixels > 12000] {
+        text-size: @landcover-font-size-big;
+        text-wrap-width: @landcover-wrap-width-size-big;
+        text-line-spacing: @landcover-line-spacing-size-big;
+        text-dy: 11;
+        [way_pixels > 48000] {
+          text-size: @landcover-font-size-bigger;
+          text-wrap-width: @landcover-wrap-width-size-bigger;
+          text-line-spacing: @landcover-line-spacing-size-bigger;
+          text-dy: 13;
         }
-        [zoom >= 16] {
+      }
+      [zoom >= 14] {
+        text-size: @landcover-font-size-big;
+        text-wrap-width: @landcover-wrap-width-size-big; 
+        text-line-spacing: @landcover-line-spacing-size-big;
+        text-dy: 11;
+        [zoom >= 15][is_airport = 'yes'] {
           text-size: @landcover-font-size-bigger;
           text-wrap-width: @landcover-wrap-width-size-bigger;
           text-line-spacing: @landcover-line-spacing-size-bigger;
--- a/amenity-points.mss
+++ b/amenity-points.mss
@@ -1488,11 +1488,11 @@
     marker-fill: @airtransport;
   }
 
-  [feature = 'aeroway_aerodrome']['access' != 'private']['icao' != null]['iata' != null][zoom >= 10][zoom < 17],
-  [feature = 'aeroway_aerodrome']['access' = 'private'][zoom >= 12][zoom < 18],
-  [feature = 'aeroway_aerodrome']['icao' = null][zoom >= 12][zoom < 18],
-  [feature = 'aeroway_aerodrome']['iata' = null][zoom >= 12][zoom < 18] {
-    [way_pixels <= 768000],
+  [feature = 'aeroway_aerodrome']['access' != 'private']['icao' != null]['iata' != null][zoom >= 10][zoom < 16],
+  [feature = 'aeroway_aerodrome']['access' = 'private'][zoom >= 12][zoom < 17],
+  [feature = 'aeroway_aerodrome']['icao' = null][zoom >= 12][zoom < 17],
+  [feature = 'aeroway_aerodrome']['iata' = null][zoom >= 12][zoom < 17] {
+    [way_pixels <= 192000],
     [way_pixels = null] {
       marker-file: url('symbols/aerodrome.12.svg');
       marker-placement: interior;
@@ -3116,11 +3116,11 @@
     text-placement: interior;
   }
 
-  [feature = 'aeroway_aerodrome']['access' != 'private']['icao' != null]['iata' != null][zoom >= 11][zoom < 17],
-  [feature = 'aeroway_aerodrome'][zoom >= 13][zoom < 18]['access' = 'private'],
-  [feature = 'aeroway_aerodrome'][zoom >= 13][zoom < 18]['icao' = null],
-  [feature = 'aeroway_aerodrome'][zoom >= 13][zoom < 18]['iata' = null] {
-    [way_pixels <= 768000],
+  [feature = 'aeroway_aerodrome']['access' != 'private']['icao' != null]['iata' != null][zoom >= 11][zoom < 16],
+  [feature = 'aeroway_aerodrome'][zoom >= 13][zoom < 17]['access' = 'private'],
+  [feature = 'aeroway_aerodrome'][zoom >= 13][zoom < 17]['icao' = null],
+  [feature = 'aeroway_aerodrome'][zoom >= 13][zoom < 17]['iata' = null] {
+    [way_pixels <= 192000],
     [way_pixels = null] {
       text-name: "[name]";
       text-size: @standard-font-size;
--- a/amenity-points.mss
+++ b/amenity-points.mss
@@ -1488,10 +1488,10 @@
     marker-fill: @airtransport;
   }
 
-  [feature = 'aeroway_aerodrome']['access' = 'private'][zoom >= 12][zoom < 17],
-  [feature = 'aeroway_aerodrome']['icao' = null][zoom >= 12][zoom < 17],
-  [feature = 'aeroway_aerodrome']['iata' = null][zoom >= 12][zoom < 17] {
-    [way_pixels = null] {
+  [feature = 'aeroway_aerodrome']['access' = 'private'],
+  [feature = 'aeroway_aerodrome']['icao' = null],
+  [feature = 'aeroway_aerodrome']['iata' = null] {
+    [zoom >= 12][zoom < 17][way_pixels = null] {
       marker-file: url('symbols/aerodrome.12.svg');
       marker-placement: interior;
       marker-clip: false;
@@ -1499,25 +1499,18 @@
       [zoom >= 15] { marker-file: url('symbols/aerodrome.15.svg'); }
       [zoom >= 16] { marker-file: url('symbols/aerodrome.18.svg'); }
     }
-  }
-
-  [feature = 'aeroway_aerodrome']['access' = 'private'][zoom >= 12][zoom < 17],
-  [feature = 'aeroway_aerodrome']['icao' = null][zoom >= 12][zoom < 17],
-  [feature = 'aeroway_aerodrome']['iata' = null][zoom >= 12][zoom < 17] {
-    [way_pixels <= 768000] {
+    [zoom >= 12][zoom < 18][way_pixels <= 768000] {
       marker-file: url('symbols/aerodrome.12.svg');
       marker-placement: interior;
       marker-clip: false;
       marker-fill: @airtransport;
-      [zoom >= 15] { marker-file: url('symbols/aerodrome.15.svg'); }
-      [zoom >= 16] { marker-file: url('symbols/aerodrome.18.svg'); }
-      [way_pixels > 12000] { marker-file: url('symbols/aerodrome.15.svg'); }
-      [way_pixels > 48000] { marker-file: url('symbols/aerodrome.18.svg'); }
+      [way_pixels > 48000] { marker-file: url('symbols/aerodrome.15.svg'); }
+      [way_pixels > 19200] { marker-file: url('symbols/aerodrome.18.svg'); }
     }
   }
 
-  [feature = 'aeroway_aerodrome']['access' != 'private']['icao' != null]['iata' != null][zoom >= 10][zoom < 16] {
-    [way_pixels = null] {
+  [feature = 'aeroway_aerodrome']['access' != 'private']['icao' != null]['iata' != null] {
+    [zoom >= 10][zoom < 16][way_pixels = null] {
       marker-file: url('symbols/aerodrome.8.svg');
       marker-placement: interior;
       marker-clip: false;
@@ -1526,19 +1519,14 @@
       [zoom >= 14] { marker-file: url('symbols/aerodrome.15.svg'); }
       [zoom >= 15] { marker-file: url('symbols/aerodrome.18.svg'); }
     }
-  }
-
-  [feature = 'aeroway_aerodrome']['access' != 'private']['icao' != null]['iata' != null][zoom >= 10][zoom < 16] {
-    [way_pixels <= 768000] {
+    [zoom >= 10][zoom < 18][way_pixels <= 768000] {
       marker-file: url('symbols/aerodrome.8.svg');
       marker-placement: interior;
       marker-clip: false;
       marker-fill: @airtransport;
       [zoom >= 11] { marker-file: url('symbols/aerodrome.12.svg'); }
-      [zoom >= 14] { marker-file: url('symbols/aerodrome.15.svg'); }
-      [zoom >= 15] { marker-file: url('symbols/aerodrome.18.svg'); }
-      [way_pixels > 12000] { marker-file: url('symbols/aerodrome.15.svg'); }
-      [way_pixels > 48000] { marker-file: url('symbols/aerodrome.18.svg'); }
+      [way_pixels > 48000] { marker-file: url('symbols/aerodrome.15.svg'); }
+      [way_pixels > 19200] { marker-file: url('symbols/aerodrome.18.svg'); }
     }
   }
 
@@ -3157,10 +3145,10 @@
     text-placement: interior;
   }
 
-  [feature = 'aeroway_aerodrome']['access' = 'private'][zoom >= 13][zoom < 17],
-  [feature = 'aeroway_aerodrome']['icao' = null][zoom >= 13][zoom < 17],
-  [feature = 'aeroway_aerodrome']['iata' = null][zoom >= 13][zoom < 17] {
-    [way_pixels = null] {
+  [feature = 'aeroway_aerodrome']['access' = 'private'],
+  [feature = 'aeroway_aerodrome']['icao' = null],
+  [feature = 'aeroway_aerodrome']['iata' = null] {
+    [zoom >= 13][zoom < 17][way_pixels = null] {
       text-name: "[name]";
       text-size: @landcover-font-size;
       text-wrap-width: @landcover-wrap-width-size;
@@ -3172,12 +3160,36 @@
         text-wrap-width: @landcover-wrap-width-size-big;
         text-line-spacing: @landcover-line-spacing-size-big;
         text-dy: 11;
-        [zoom >= 16] {
-          text-size: @landcover-font-size-bigger;
-          text-wrap-width: @landcover-wrap-width-size-bigger;
-          text-line-spacing: @landcover-line-spacing-size-bigger;
-          text-dy: 13;
         }
+      [zoom >= 16] {
+        text-size: @landcover-font-size-bigger;
+        text-wrap-width: @landcover-wrap-width-size-bigger;
+        text-line-spacing: @landcover-line-spacing-size-bigger;
+        text-dy: 13;
+      }
+      text-face-name: @oblique-fonts;
+      text-halo-radius: @standard-halo-radius;
+      text-halo-fill: @standard-halo-fill;
+      text-placement: interior;
+    }
+    [zoom >= 13][zoom < 18][way_pixels <= 768000] {
+      text-name: "[name]";
+      text-size: @landcover-font-size;
+      text-wrap-width: @landcover-wrap-width-size;
+      text-line-spacing: @landcover-line-spacing-size;
+      text-fill: darken(@airtransport, 15%);
+      text-dy: 9;
+      [way_pixels > 48000] {
+        text-size: @landcover-font-size-big;
+        text-wrap-width: @landcover-wrap-width-size-big;
+        text-line-spacing: @landcover-line-spacing-size-big;
+        text-dy: 11;
+      }
+      [way_pixels > 192000] {
+        text-size: @landcover-font-size-bigger;
+        text-wrap-width: @landcover-wrap-width-size-bigger;
+        text-line-spacing: @landcover-line-spacing-size-bigger;
+        text-dy: 13;
       }
       text-face-name: @oblique-fonts;
       text-halo-radius: @standard-halo-radius;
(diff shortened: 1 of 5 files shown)


jeisenbe:
Here are the other 3 aerodromes with IATA and ICAO codes in Northern Ireland:

### Derry airport
#### z10 before
- Derry airport text label shows, but RAF airport to east has text label blocked, unexpectedly.
z10-derry-airport-before
**After** - just the icons are rendered for both; less confusing
z10-londonderry-airport-after

#### z11 same before/after
z11-derry-airport-before

#### z14 before
Only 48k `way_pixels` but no rendering; should render to about z16 (~768 way_pixels)
z14-derry-airport-before-waypixels-48133

### Enniskillen airport
#### z10 before
- Label renders; rather unhelpful when it's right above a town label with the same name.
z10-enniskillen-airport-before
**After** - icon only
z10-enniskillen-after

#### z14 after
- No icon or label in "before"
z14-enniskillen-airport-after

#### z16 after
- 194k `way_pixels` at z16 so it should be rendered.
z16-enniskillen-airport-after-waypixels-194000-last-z-level

### Castlerock
- Small aerodrome without IATA or ICAO code, has same name as village, currently renders sooner
#### z11 before
z11-castle-rock-before
**after** - not rendered
z11-castlerock-after

#### z12 before
z12-castlerock-before
**z12 after** - does not dominate over village
z12-castlerock-after

#### z14 before
- did not render
z14-castelrock-before

#### z17 after
- Last zoom level. Still just a small portion of screen;  about an 800m long runway.
z17-castle-rock-after

jeisenbe:
## Charlottetown, Canada
- 46 degrees north; largest airport for the main city on Prince Edward Island
https://www.openstreetmap.org/way/377214479#map=14/46.2890/-63.1173

### z10 before
(Only 315 way_pixels at this zoom level: about 11x30, only a little larger than the 14x14 pixel icon)
z10-charlottetown-airport-before
**after**
z10-charlottetown-airport-after

### z11 - unchanged
z11-charlottetown-airport-after

### z14 before
z14-charlottetown-airport-before
**After**
z14-charlottetown-airport-after

## Cable Head Airpark
- On the northeast side of the island, 46 degrees
- https://en.wikipedia.org/wiki/Cable_Head_Airpark 
- unpaved grass runways, no IATA or ICAO code, private
### z11 before
z11-cable-head-airpark-before
** After ** - not rendered

### z12 before
z12-cable-head-airpark-before
**After** - icon only
z12-cable-head-airpark-after2

### z14 before
z14-cable-head-airpark-before

### z15 AFTER
z15-cable-head-airpark-after

imagico:
I have a few concerns:

* that a two zoom level gap between the two classes of airport defined by the presence of codes and access tagging might put too much weight on criteria that are not really that meaningful in the end.
* that showing symbol and label up to z16 will often look weird even for smaller airports with a <3000m runway at medium latitudes - like https://www.openstreetmap.org/#map=16/38.1794/13.1005 and especially of course for any large size airport with 5000m+ runways that will almost always exceed screen size at this scale.
* that many regions where air transport is the primary mode of longer distance transport and even small airfields are important infrastructure elements are at high latitudes where the current starting zoom levels are already quite late.  Like https://www.openstreetmap.org/#map=10/66.7301/-160.6290
* that runways start getting rendered at z11 meaning that if you don't start showing the symbol before z12 the runway will potentially be visible before the airport symbol.
* that frankly from the examples you have shown - while i see a few cases of minor improvement i see no example where this change solves a serious problem.

One thing that could be an option:  The plane symbol could in principle also be recognizable at somewhat smaller size.  You could think about starting with a smaller version of the symbol if that is of benefit.

jeisenbe:
Sorry, I should have shown the good examples first, but then I decided to show the "worst-case" scenario of N Ireland. It's true that the changes are not very impressive at high latitudes in Europe, where you tend to put your airports far from towns and don't have very many small ones. In contrast, in the tropics it's quite common for every important `village` and small `town` to be built right around the landing strip, and even small cities grew up around airports. This makes label collisions very common, and it's also a problem to show the airstrip name before the village name.

# Papua, Indonesia
https://www.openstreetmap.org/#map=10/-4.000/139.000
- About 4 degrees south

## Wamena & Baliem valley
- Wamena airport is busy with 200k passengers annually and many cargo flights via Boeing 737s
- The other airports in the z10 image have turboprop or prop planes, but have scheduled passenger service; every town has an airport, and some large villages with ~2000 or 3000 people (maybe I should change those to towns? But the national standard was for >10k for towns)

### z10 before
z10-bandar-udara-baliem-valley
**after** - there are still some collisions between icons and town text labels; ideally only major international airports should be shown at z10 at the equator.
z10-baliem-valley-airports-after

### z11 before / after (same)
- Currently this is the highest zoom level where the airport name is shown! 
Text label isn't shown after this due to collisions. Icon is blocked by collision with city name
z11-bandar-udara-wamena-only-z-level-with-label

#z13 same
z13-bandar-udara-wamena-symbol-last

### z13 Before / After (same)
- Last zoom level with icon rendered; text label still blocked by collision with secondary highway
z13-bandar-udara-wamena-symbol-last

z14 Before
- No icon or label, but airport is < 7, `way_pixels`
z14-bandar-udara-wamena-no-symbol
**After** - symbol shown, quite helpful (airport is still a small part of screen)
z14-wamena-airport-after

### z15 AFTER
- We finally get the text label again; `way_pixels` are 27k
z15-wamena-airport-after

### z16 after
- 108k `way_pixels` - I believe it's still quite helpful to have the symbol and label here. Note that this is only the third z level where the text label is shown.

## Yahukimo
https://www.openstreetmap.org/#map=11/-4.500/139.500
- These airstrips have just and unpaved runway, a wind sock, and a small shed for a radio and storage; they are built by the local communities in every sizable village since it's the only way to get to town, besides for walking for 2 to 5 days, but there is not scheduled commercial service, only irregular and chartered flights. Large areas of New Guinea and Borneo look like this

### z11 Before - south of mountains
- Far too many small aerodromes shown; no village names shown
z11-yahukimo-too-many-aerodromes
** After **  - too bad villages don't render at z11. I'll check if that's feasible eventually.
z11-yahukimo-airstrips-after

### z12 After
z12-yahukimo-soba-after


### z11 before - north of mountains
z11-n-yahukimo-many-aerodromes
** After **
z11-n-yahukimo-airstrips-after

### z12 North Before
**After**
z12-n-yahukimo-airstrips-after

### z13 Before / After - unchanged
z13-anggruk-airstrips-after-same

### z14 AFTER 
- Not rendered before
z14-anggruk-walma-airstrips-after

## Pengunungan Bintang 
- Star Mountains, near border with Papua New Guinea

### z11 before
z11-pengunungan-bintang-too-many-aerodromes
** After **
z11-pengunungan-bintang-aerodromes-after

### z12 After - all the labeled villages have an aerodrome, but most icons are still blocked by collisions at this level
z12-kiwirok-airstrips-after

## Sentani International Airport
- 2 million annual passengers, busiest airport for the Indonesian half of New Guinea, though it rarely has international flights scheduled. The town of Sentani grew up right next to it. 

**z10** - icon and text labels blocked by town / lake)
z10-sentani-airport-after

### z11 Before
z11-sentani-airport

### z12 Before
z12-sentani-airport-blocked-by-village

### z13 Before
z13-sentani-international-airport-no-label

### z14 AFTER
z14-sentani-airport-after

### z15 before
- No label, yet airport is not very large on screen
z15-sentani-airport-current

### z17 Before/After
- Terminal name starts rendering, so no need for airport icon/label now
z17-bandar-udara-sentani-terminal-before

jeisenbe:
> * that a two zoom level gap between the two classes of airport defined by the presence of codes and access tagging might put too much weight on criteria that are not really that meaningful in the end.

I would prefer to have "airports" (aerodromes with `iata`) show up starting at z9, since at z10 even the huge Chiangi airport in Singapore looks odd, and many airport icons are blocked by town and city labels in the tropics. I believe it is a principal that the majority of icons should not be blocked by collisions at the initial zoom level. But I was reluctant to do this, since it would be surprising for major airports at high latitudes (e.g. Belfast International).

Re: IATA codes; all airports with commericial air transport (regular scheduled airline flights) have an IATA code, even in Indonesia and most developing countries. While some aerodromes with IATA codes are old military airports or old airports which no longer have scheduled service, it's nearly certain that all "airports" (in the popular sense) have one of these codes, and mappers have been very diligent about adding them. I agree that checking for `access=private` isn't very helpful; it only matters for 32 features world-wide currently (vs >6800 aeroway=aerodrome with both `iata=` and `icao=`)

I would like to develop the `aerodrome=*` subtag further with something like `=commercial` vs `=general_aviation` to distinguish airports that do not have scheduled commercial flights, but that's not an option yet. In the future, we might use `aerodrome=international` to render big airports sooner, and `aerodrome=airstrip` to render the small ones later; say international at z10 and airstrip not till z13.

> * showing symbol and label up to z16 will often look weird even for smaller airports with a <3000m runway at medium latitudes - like https://www.openstreetmap.org/#map=16/38.1794/13.1005 and ... 5000m+ runways ... exceed screen size.

Yes, that's why I included `[way_pixels <= 768000 ]` - but perhaps that restriction is too loose? It was designed for `landcover` labels, which get larger, unlike the constant-size `aerodrome` label and icon. It would look better to adjust the `aerodrome` text label font size with zoom level or `way_pixels`, but this would also probably require adjusting the size of the svg icon. 

Otherwise, we could change this limit to `[way_pixels <= 192000 ]`, or use <16 for airports and <17 for others, or obth.

> * that many regions where air transport is the primary mode of longer distance transport and even small airfields are important infrastructure elements are at high latitudes where the current starting zoom levels are already quite late.  Like https://www.openstreetmap.org/#map=10/66.7301/-160.6290

66.7 degrees is quite high. I believe <1% of the global population lives north of 60 degrees (none live south of 60), and I suspect <0.01% live north of 66 degrees. Since cos of 66.7 is 0.40, a pixel at the equator is 6.2 times more area than a pixel at 66.73 degrees north, and a line is 2.5 times longer at the equator when drawn the same length, which makes it very difficult to get reasonable rendering for features in the tropics and at 66.7 degrees when using mercator.  I'll open an issue to discuss this further. 

> * runways start getting rendered at z11 meaning that if you don't start showing the symbol before z12 the runway will potentially be visible before the airport symbol.

Stadiums and Universities render from z10 and `leisure=track` renders at z10, but we don't show the text labels for these features until `way_pixels > 3000`, which usually doesn't happen until z14 or later.

Belfast International has `way_pixels 512` at z10 and `2048` at z11, and Charlottetown airport is `way_pixels 315` at z10 and `1260` at z11, so it would be more reasonable to render the text label starting at z12, rather than z11 as in this PR, even at these higher latitudes. Belfast City, like the other mid-sized airports in N. Ireland, only reaches >3000 at z13, and the small airstrips don't achieve this size till z14 or z15.

> One thing that could be an option: The plane symbol could in principle also be recognizable at somewhat smaller size. You could think about starting with a smaller version of the symbol if that is of benefit.

That would be worth trying, especially if are to keep the symbol at z10. I notice that the `natural=peak` symbol is significantly smaller and less obtrusive (and that the name does not render till z13).

What do you think about using an even larger icon at higher zoom levels and adjusting the text size large as well, so that it works more like the `landcover` text labels? This was discussed in #1055, but we haven't tried it yet.

jeisenbe:
# Singapore
### z10 before
- Whole country (>6 million people) in this image. Only 1 airport is of general interest (Changi), text labels are too big especially for minor airports
z10-singapore-airports-before
**After**  - better
z10-singapore-airports-after

### z11 before
- Whole country/city still fits in one screen, too many small military airports shown, one blocks the text label from Changi (the big international airport)
z11-singapore-airports-before
** After ** - small airports unlabeled, just icon
z11-singapore-airports-after

## Changi
- The real airport, one of the busiest in southeast Asia. 
### z12 before
- Military airfield is shown with same prominence
z12-singapore-airports-east-before
**After** only label for Changi (commercial airport) shows
z12-east-singapore-airports-after


### z13 before / after
z13-changi-singapore-airports

### z14 before
- no label
z14-changi-singapore-airports-before-no-icons
**after**
z14-changi-singapore-airports-after

### z15 AFTER
Way_pixels is 500k, so this would be filtered by change to 192k as discussed
z15-changi-airport-singapore-after-waypixels-499521

## Selatar
- General aviation; no commercial flights
### z13 Before / After
z13-selatar-singapore-airports-after

### z14 Before
- not rendered
z14-seletar-singapore-airports-before
** After **
z14-seletar-singapore-airports-after

### z15 AFTER
z15-seletar-singapore-airport-after

### z16 after
- way_pixels = 307,000, so this could be filtered out if limit is changed to 192k
z16-seletar-singapore-airport-after-waypixels-307k

## Murai
### z12 Before / After
- Unchanged; aerodrome text label blocked by Air Base text lable
z12-singapore-airports-west-before

### z13 Before/After
z13-murai-singapore-airports-after-same

### z14
z14-murai-singapore-airports-after
 AFTER
- Not rendered before

### z15 after
z15-murai-singapore-airports-after

### z16 after
z16-murai-singapore-airports

meased:
I like these changes as I feel aerodromes are somewhat messy right now.

An old idea of mine was to render small aerodromes with a prop plane icon:

aerodrome_small 12 svg

But really, the [aerodrome proposal](https://wiki.openstreetmap.org/wiki/Proposed_features/Aerodrome) needs to get approved first so we can use `aerodrome=international`, as using the `iata=*` / `icao=*` tags seems to be a bit of a hack.

jeisenbe:
Based on the additional test images in Singapore and suggestions above, I've pushed a new commit which changes the last zoom level to z15 for airports with IATA codes and z16 for other aerodromes, and `way_pixels < 192k` instead of 768k (for those mapped as areas).

The removes the icon and text label rendering from z16 for those with an IATA code, and from z17 for other aerodromes, and also removes the rendering if way_pixels is between 192k and 768k (one zoom level)

jeisenbe:
@meased - I've thought about using a different icon, but I agree that we would need a more precise tag like `aerodrome=general_aviation`, `aerodrome=airstrip` for aerodromes without commercial air transport. 

Re: IATA codes; [every airport with scheduled airline service](https://aviation.stackexchange.com/questions/8644/are-there-any-airports-that-do-not-have-an-iata-code-but-have-a-huge-number-of-t), including very small "airlines" flying prop planes, has an IATA code. And these codes are quite useful for identifying airports, since they are often printed on tickets, boarding passes (e.g. for transfers and connecting flights) and baggage tags, and shown on ticket buying websites, so I think it's quite important for mappers to add these. See: 

From what I've seen, there are no airports with an [IATA code](https://www.iata.org/services/Pages/codes.aspx) that lack an ICAO code, but there are some ICAO-coded aerodromes that have no IATA code: https://aviation.stackexchange.com/questions/8767/when-do-we-use-iata-codes-and-when-do-we-use-icao-codes

I considered if we might just check for `iata`, to save a few lines of code.

In the OSM database, there are currently:
8775 `aeroway=aerodrome` + `iata` (2950 nodes, 5825 ways and relations) 

In contrast, for ICAO there are 12,158 (4098 nodes, 8060 ways and relations)

There are [5339 aerodromes](https://overpass-turbo.eu/s/LZW) with ICAO codes but without IATA codes in the database; some of these are unpaved grass airstrips with no weather or control (https://aviation.stackexchange.com/questions/14585/why-do-some-airports-not-have-icao-codes?rq=1)
Examples: 
1) [Lincoln Regional Airport](https://www.openstreetmap.org/node/369169970)  - general aviation ([wikipedia](https://en.wikipedia.org/wiki/Lincoln_Regional_Airport_(California)))
2) [Twentynine Palms Strategic Expeditionary Landing Field](https://www.openstreetmap.org/node/973472283)  - special [military airfield](https://en.wikipedia.org/wiki/Twentynine_Palms_Strategic_Expeditionary_Landing_Field)
3) [Pine Valley Airport](https://www.openstreetmap.org/node/1042053236) - private airport, doesn't actually have an ICAO code; FAA id was mistakenly entered (common problem). https://en.wikipedia.org/wiki/Pine_Valley_Airport
4) [Bandar Udara Kiwirok](https://www.openstreetmap.org/node/511570215)  -see test images above - 600m long runway, irregularly scheduled flights with small prop planes.

There are also 1956 aerodromes with an IATA code but no ICAO code currently. 

I checked on these: there aren't any in the UK, Japan, Germany, or France, and there are only 19 in all of Russia, so places with active mapper communities have added ICAO codes to all airports with IATA codes.

Examples of those missing the ICAO code in Russia
https://en.wikipedia.org/wiki/Teply_Klyuch_Airport - which should be icao=UEMH according to wikipedia, and https://en.wikipedia.org/wiki/Kirovsk-Apatity_Airport - ICAO= ULMK 

But in the USA there are [510 aerodromes](https://overpass-turbo.eu/s/LZZ) with an IATA but no ICAO, especially up in Alaska, and many actually don't have an ICAO code assigned.

Examples:
1) First one I checked, [Alitak Seaplane Base](https://www.openstreetmap.org/node/369144747) https://en.wikipedia.org/wiki/Alitak_Seaplane_Base in Kodiak Island, Alaska, has an IATA and no ICAO per wikipedia, but has "air taxi" service; on-demand airline flights with small sea planes.  
2) Next, [Lime Village Airport](https://www.openstreetmap.org/node/369144748) - no other info, no ICAO.
3) First node on list [Barksdale](https://www.openstreetmap.org/node/29559516) https://en.wikipedia.org/wiki/Barksdale_Air_Force_Base - actually has an ICAO, but not tagged. Also missing military=airfield or landuse=military tag
4) Most recent node in lower 48 states: [Grove City Airport](https://www.openstreetmap.org/node/5471871595) (https://en.wikipedia.org/wiki/Grove_City_Airport) - listed as `iata=29D` but this is actually the FAA id; there is no IATA code or ICAO code per wikipedia. 
5) [Loring "international" airport](https://www.openstreetmap.org/node/1042047849)- https://en.wikipedia.org/wiki/Loring_International_Airport - old army airfield, now repurposed, has IATA but no ICAO code per wikipedia, appears not to have commercial flights yet. Per http://www.airnav.com/airport/ME16 it is "private use".  See https://loring.org

Indonesia also has 23 aerodromes with `iata` but no `icao` - and most of these seem to be mistakes, imported from the crowdsourced webites.

And there are 50 aerodromes with `icao` only - most are quite small, and many seem to be mistagged: [ICAO codes for Indonesia](https://www.notams.faa.gov/common/icao/R24C1.html) start with [WA, WI, WQ or WR]((https://en.wikipedia.org/wiki/List_of_airports_by_ICAO_code:_W#WA_WI_WQ_WR_-_Indonesia)), but many in OSM start with WZ and WX: I think these are local codes, not real ICAO codes 

Overall there are 
6819 aerodrome with both codes (1724 nodes, 4728 ways, and 367 relations) (16.5%)
14,024 aerodromes (5232 nodes, 8275 ways, and 517 relations) with iata or icao (34%)
27,228 aerodromes without either code (66%)
41252 aerodromes total

It looks like all the important airports have both an ICAO code and IATA code, so by checking for these we are getting all the big ones in the earlier rendering, and avoiding some bad data among the airports that are tagged with just an iata code.

jeisenbe:
I tried adjusting the text and icon size by zoom level and way_pixels, but I'm having trouble getting the SQL right. 

Rather than checking for `['iata' != null]['icao' != null]['access' != 'private'] I tried to make this SQL query:

```
              CASE 
                WHEN aeroway = 'aerodrome' THEN
                  CASE
                    WHEN tags->'icao' = NULL OR tags->'iata' = NULL OR access = 'private' THEN 'no' 
                    ELSE 'yes'
                  END
                ELSE NULL 
              END AS is_airport,
```

And then this:

```
[feature = 'aeroway_aerodrome'][is_airport = 'yes'][zoom >= 10][zoom < 16],
  [feature = 'aeroway_aerodrome'][is_airport = 'no'][zoom >= 12][zoom < 17] {
    [way_pixels <= 192000],
    [way_pixels = null] {
      marker-file: url('symbols/aerodrome.8.svg');
      marker-placement: interior;
      marker-clip: false;
      marker-fill: @airtransport;
      [zoom >= 11] {
        marker-file: url('symbols/aerodrome.12.svg')
      }
      [is_airport = 'yes'] {
        [zoom >= 14] {
          marker-file: url('symbols/aerodrome.15.svg')
        }
        [zoom >= 15] {
          marker-file: url('symbols/aerodrome.18.svg')
        }
      }
      [is_airport = 'no'] {
        [zoom >= 15] {
          marker-file: url('symbols/aerodrome.15.svg')
        }
        [zoom >= 16] {
          marker-file: url('symbols/aerodrome.18.svg')
        }
      }
      [way_pixels > 12000] {
        marker-file: url('symbols/aerodrome.15.svg')
      }
      [way_pixels > 48000] {
        marker-file: url('symbols/aerodrome.18.svg')
      }
    }
  }
```

But now all the airports without IATA and ICAO codes are still rendering as `is_airport=yes`, so I did something wrong clearly.

jeisenbe:
Here are test renderings from the airports in Singapore with the adjusted icon and text size:

### z10 Singapore - 8 pixel icon (current 12 pixel icon scaled down)
z10-singapore-airports-after-8px-icon

z11 and z12 are same as previous tests, no change in icon or text size.

### z12 Changi and Seletar airports
z12-east-singapore-airports-after

### z13 Changi airport - 15 pixel icon, larger text
z13-changi-singapore-airports-after-large-icon-text

### z14 Changi airport - larger icon and text
z14-changi-singapore-airports-after-large-icon-text

### z15 Changi air base - larger text
- The big international airport no longer renders at this zoom level, due to way_pixels > 192k, but this military airport would render with the large icon and text
z15-changi-air-base-east-after-large-icon-text

## Seletar Airport
- Same from z12 to z13, like above:
z13-selatar-singapore-airports-after

z14 Larger text and icon
z14-seletar-singapore-airport-after-large-icon-text

z15 Large icon - text blocked by runway number
z15-seletar-singapore-airport-after-large

jeisenbe:
I've pushed a new commit with my partially successful attempt at adjusting the icon and text size; perhaps someone can help me find my mistakes.

jeisenbe:
I fixed the problems with the previous commit, though the code is not as compact as I would like. I would appreciate some advice about how to condense it.

Here are the current results, with icon size and text size adjusted based on zoom level (and way_pixels). I returned the way_pixels max limit to 768,000 because the text size now matches other landcover areas which use this same limit. 

### Northeast Wales
- 3 small airstrips and one airport which is mainly used for cargo, but had scheduled service years ago, and also serves an Airbus wing factory:
https://en.wikipedia.org/wiki/Hawarden_Airport
z10 Hawarden latest commit - small icon
z10-hawarden-airport-after-small

z11 Hawarden latest commit - unchanged
z11-hawarden-airport-after-same

z11 airstrips before
- labels of the 2 in the north are blocked by collisions
z11-st-asaph-airfields-before

z11 Airstrips - latest commit - no longer shown
- These two have no IATA or ICAO codes, they are quite small, but are mapped as areas.
z11-st-asaph-airstrips-after

z12 airstrips - latest commit - just icon
z12-st-asaph-airstrips-after

z13 airstrips - latest commit - inspector view (just 680 way_pixels here, but text label shown)
z13-st-asaph-airstrips-after-inspector-waypixels-680

z14 airstrips - latest commit
z16-rhedyn-airstrip-after-43500-waypixels

z14-st-asaph-airstrips-after

z16 rhedyn airstrip - latest commit - still could be shown at z17, but stops here
- Larger text and icon
z16-rhedyn-airstrip-after-43500-waypixels

z16 bryngwyn farm airfield - latest commit - last zoom level rendered (but z17 would be ok)
z16-bryngwyn-airfiled-after


### Northwest Wales
- There are 2 current and 1 former RAF (airforce) military airfield in this view, so these are the worst-case scenarios as far as larger aerodromes far from the equator without an IATA code:
 
z10 current commit
z10-caernarfon-airports-after

z11 current commit
z11-caernarfon-airports-after

### RAF Mona
- Military airfield in Wales, so larger than most aerodromes mapped as an area but without an IATA code. I wonder if the icon color should match the military area rendering, or be dropped with military=airfield and the military text color used? 
z12 latest commit
z12-raf-mona-after

z13 latest commit - same as current rendering
z13-raf-mona-after-same

z14 latest commit
z14-raf-mona-after

z15  latest commit
z15-raf-mona-after

z16 latest commit
z16-raf-mona-after

- Not rendered on z17

### Caernarfon airport
- Former airforce base, now used for general aviation, no IATA code; mapped as a node. (This is one of the larger airfields without an IATA code, due to it's former military use and high latitude)
- https://en.wikipedia.org/wiki/Caernarfon_Airport

z12 latest commit
z12-caernarfon-airports-after

z13 latest commit - unchanged from current rendering
z13-caernarfon-airport-same

z14 latest commit
z14-caernarfon-airport-after

z15 latest commit
z15-caernarfon-airport-after

z16 latest commit
z16-caernarfon-airport-after
- Not rendered on z17 or higher

jeisenbe:
New tests in Papua, Indonesia; compare with images in https://github.com/gravitystorm/openstreetmap-carto/pull/3856#issuecomment-526876176 above for "before".

## Wamena airport

(z10 airport icon and label blocked by city name, before and after)

z11 latest commit
z11-wamena-airport-after

z12 latest commit - unchanged (blocked before and after)
z12-wamena-airport-not-shown-collisions

z13 latest commit - unchanged
z13-wamena-airport-after-no-text
 
z14 latest commit - larger icon
z14-wamena-airport-after-big-icon

z15 latest commit - larger icon, text still blocked by colllision
z15-wamena-airport-after-big

z16 Latest commit - large text now shown
z16-bandar-udara-wamena-after-large

z17 latest commit - icon and text shown (last level)
z17-wamena-after-large-icon-text

## Tulem airstrip
- Small rarely-used airstrip, built before the current airport, about 10 km to the north. One of the larger airstrips in the area; there about a dozen others, mostly smaller, in the district. 

z13 latest commit - same as current
z13-tulem-after-same

z14 latest commit
z14-tulem-after

z15 latest commit
- not larger because area is still quite small
z15-tulem-after

z16 latest commit
z16-tulem-after

z17 latest commit
- Last zoom level shown, text is still small since the area is small.
z17-tulem-after

jeisenbe:
I fixed a couple more bugs. More test images:

## Darwin International Airport
- Main airport for the largest city in Northern Territory, Australia. Located in the tropics at 14 degrees: https://www.openstreetmap.org/#map=14/-12.4146/130.8881
- For some reason it is tagged as landuse=military; this is probably incorrect, but it makes it much easier to see the area. But again, perhaps aerodromes with landuse=military / military=airfield should have the label in the military color rather than purple.

### z10 before
z10-darwin-airport-before

z10 after
z10-darwin-airport-after

### z11 before / after - unchanged
z11-darwin-airport-after

### z12 before / after - unchanged
z12-darwin-airport-after

### z13 before
z13-darwin-airport-before

z13 after
z13-darwin-airport-after-new

### z14 before - not shown
z14-darwin-airport-before

z14 after
z14-darwin-airport-after

### z15 after - last zoom level - 636k way_pixels, just makes it under the cut-off
z15-darwin-airport-after

### z16 after 
- not rendered; too large (2.5 million waypixels)
z16-darwin-airport-not-rendered

jeisenbe:
Northern Australia has lots of tiny airports / airstrips:

### z10 before
z10-lake-evelia-airports-before

z10 after
z10-lake-evelia-airports-after

### z11 before
z11-lake-evelia-airport-before

z11 after
z11-lake-evelia-airport-after

### z12 before
- Village names still not shown, because they are `place=hamlet`
z12-lake-evelia-airstrips-before

z12 after
z12-lake-evelia-airstrip-after

### z13 before / after - unchanged
- Village names still not shown, because they are `place=hamlet`
z13-raymangirr-before

### z14 before - no label or icon
z14-raymangirr-before
z14 After
z14-raymangirr-airport-after

### z15 AFTER
z15-raymangirr-airport-after

### z16 AFTER
z16-raymangirr-after-new

### z17 AFTER
z17-raymangirr-airport-after-new

Not shown at z18

imagico:
Not sure when you consider this ready - when you work on a PR for longer it is usually good to indicate this with a WIP note in the title.

As a general note on what you are trying now - from a perspective of a well readable and intuitively understandable map design i am against scaling an icon + label annotation of features.  The scaled labels for polygons we have been using are already a pretty delicate thing that only works when used sparsely and we are already overusing that quite significantly.  Scaling also symbols and offset labels next to them (which by the way are shown and hidden independently - see #234) would significantly aggravate this problem.  My suggestion was to - if that helps - reduce the symbol size at the starting zoom level when it is shown without a label.  This could help maintaining a relatively low starting zoom level without too much filling the map at low latitude.  Apart from that i would keep the logic relatively simple.

The problem about showing symbol/label at relatively high zoom levels when the airports are large compared to the size of the labels is the lack of sophistication in label placement.  A well working label would need to be located near the functional center of the airport - near the main tower, the terminal building or between those and the runway.  Scaling symbol and label is a superficial attempt to hide this but it does not really work, it just disguises the problem.

jeisenbe:
Sorry, it's ready now. I thought it was ready before, but kept finding new mistakes when I made test renderings in different places. 

> Scaling symbol and label is a superficial attempt to hide this but it does not really work, it just disguises the problem.

It certainly would be better if the label placement could be adjusted. But I believe the larger text size is helpful because it makes it clear that this is a large feature. Since we do label `landcover` features with the larger labels when they start to take up a significant portion of the screen, adding this feature for airports helps to show that the label is for a large area, not just for the point where the icon is located. I find the "after" images at z14 to z16 much improved because of this. 

Objectively, the larger labels are only about 50% bigger than the standard text size, so this doesn't actually do anything to hide the sub-optimal label placement (especially since the ST_PointOnSurface changes), but the larger text does make it clear that it isn't a specific point. 

I considered removing the airport icon at higher zoom levels, but at that point it's almost never blocked by collisions with other features. This tends to happen at z10 to z12, which is why I removed the text labels for non-commercial aerodromes (lacking IATA codes) at z11 to z12 and from z10 for airports. 

(I'd be interested in trying to solve #234 too.)

jeisenbe:
It would also be better if the aerodrome polygon fill color was a little more purple. Right now it looks too similar to parking lots and untagged lands. Matching the text, icon and fill color helps make the map more intuitive.

For this reason, I'm also considering a separate PR to render military airfields differently, so that the purple text does not mis-match with the military fill/pattern.

kocio-pl:
I made some general notes about PR review here: https://github.com/gravitystorm/openstreetmap-carto/issues/2291#issuecomment-529113602, so now I will only try to spot on this specific ticket. I gave there my explanation why trying to make a full review is sadly just hopeless.

I believe text scaling is a great tool, but I'm less of a fan for icons. I think making them bigger to stick with the labels is justified (uniformity of elements describing the same object), but scaling down does not work for me at all. I hope just removing labels at initial levels should be enough.

The problem of stopping to render icons for me is related to the airport size. The last time I want to see the icon is when the whole airport i still visible on the screen. Since they differ in size it is not a simple choice, but we can estimate average size for public/private airports for example (or use some other tags also).

imagico:
> But I believe the larger text size is helpful because it makes it clear that this is a large feature.

I know this is the reasoning but it works less so than you might think.  Basing the label size on way_area rather than on semantics (like it is done for populated places for example or administrative units) removes the label size as an element that allows identifying the type of label or that transports useful information not shown otherwise and limits intuitive recognition of the overall appearance of certain types of features.

Airports - unlike for example lakes - are not a featureless two-dimensional entity defined exclusively (or even foremost) by the geometry of their outline.

Also keep in mind that if you extend the label rendering above z15 it becomes very likely that the label is blocked by a runway/taxiway label or some other label and the likeliness of this happening increases with increasing label size.  You might not get a very consistent rendering of labels at the higher zoom levels because of that.

kocio-pl:
To keep things simple: the shortest general explanation why label scaling works is this image - the semantic part is that smaller objects belong to larger ones and it is clearly visible at just one sight:

Screenshot_2019-09-07 OpenStreetMap

imagico:
@kocio-pl - i don't doubt that there are examples where label scaling based on way_area works well - that is not what i have put into question here.  I also explained that there is an enormeous difference between labeling something like a lake and labeling something like an airport.

Apart from that your sample mainly illustrates how dysfunctional our rendering of nature reserve boundaries currently is in a lot of cases.  Many map readers in your example by the way will not necessarily read one large nature reserve and three smaller ones but one of a more important class of nature reserve and three ones with a less significant classification.  The main reason why your example works is because the size is congruent with the semantic classification in this case so this interpretation is not in conflict with reality.  This works less so for example here:

https://www.openstreetmap.org/#map=9/37.9160/-111.8367

where you have prominently visible the _Dixie National Forest_ but much smaller and less prominently the much more significant (because more known and more visited)  _Capitol Reef National Park_ - not to mention _Bryce Canyon National Park_, which is not even labeled at the linked to zoom level.  Someone who is not familiar with the details of classification of protected areas in the US (National Park/Monument/Forest/Recreation Area, Wilderness Area, Conservation Area and countless other terms) is very likely to be completely confused because the label size does not match what they'd expect it to indicate, namely a meaningful measure of importance of the features labeled.

kocio-pl:
For the reasons I just gave in https://github.com/gravitystorm/openstreetmap-carto/issues/2291#issuecomment-529128300 I won't go further with discussing size-based labels here - we can talk this to death with no conclusion for this ticket.

I don't see anything blocking this code from merging. Maybe there are things that could be improved, but nothing scary appears on the surface and proper review would too time consuming, so I give you a benefit of doubt. The only thing I'm opposed would be downscaling icons, but I'm not sure how hypothetical this proposition was.

kennykb:
@imagico - The specific example that @kocio-pl gave would have worked in either approach. The Adirondack Park is tagged `boundary=national_park` (I've discussed the reasons at length elsewhere) while the smaller protected areas that it contains are all `boundary=protected_area protect_class=* leisure=nature_reserve`. The rendering of those smaller regions will always, unavoidably, be a mess because in many cases they are very badly fragmented. (Sorry, they're fragmented in the field as well!) 

Unfortunately, I don't think that there's any mapping that can associate `protect_class` or `protection_title` with relative importance of the feature.  As you correctly observe, there is a whole zoo of different sorts of protected areas over here, and while you give examples where a visitor would have an expectation of what is 'more important', the idea breaks down when trying to decide questions like, "is a small 'State Historic Site' more important that the huge 'Wild Forest' next to it?"  I'm forced to conclude that "relative importance" is, in general, subjective and not verifiable, while "area" is a measurable, verifiable quantity that in a great many cases is a useful surrogate measure.

jeisenbe:
> close the old one and open a new PR with your new approach

(From https://github.com/gravitystorm/openstreetmap-carto/issues/2291#issuecomment-529125294) 

Would you recommend that in this case, @imagico, @kocio-pl?

jeisenbe:
> scaling down does not work for me at all. I hope just removing labels at initial levels should be enough

Is the new 8 pixel size aerodrome icon too small to be clearly legible at z10? Are you saying that you would prefer that we keep the current 12 pixel icon at z10 or remove the airport icon at z10 for now (though later we might add it back for `aerodrome=international` if this is used widely enough)?

jeisenbe:
@imagico - it sounds like you are opposed to the use of larger icons and text at higher zoom level / higher way_pixels. 

This option was not originally part of the PR, so I don't mind removing it, and also removing the rendering at z16/z17 where otherwise the small icon/text is confusing.

amenity-points.mss
```diff
@@ -1488,12 +1488,46 @@
...
+  [feature = 'aeroway_aerodrome']['iata' = null] {
+    [zoom >= 12][zoom < 17][way_pixels = null] {
+      marker-file: url('symbols/aerodrome.12.svg');
+      marker-placement: interior;
+      marker-clip: false;
+      marker-fill: @airtransport;
+      [zoom >= 15] { marker-file: url('symbols/aerodrome.15.svg'); }
+      [zoom >= 16] { marker-file: url('symbols/aerodrome.18.svg'); }
+    }
+    [zoom >= 12][zoom < 18][way_pixels <= 768000] {
```
jeisenbe: @imagico - are you ok with using way_pixels to stop rendering the icon/text sooner for large aerodromes which are mapped as areas, as in this line and in lines 1522, 3175 and 3226?

jeisenbe:
I've removed a couple of my comments which included outdated test images, which are not necessary to understand the discussion.

imagico:
> @imagico - it sounds like you are opposed to the use of larger icons and text at higher zoom level / higher way_pixels.

I consider this an advise rather than a firm opposition.

I don't think this is a good idea but at this point my main concern is not so much that that the results would be so bad that it is a serious problem to try it but that we (the maintainers here) currently lack an agreement on how to evaluate if such a change is positive in the results or not.

Judging from various recent discussions about the benefit of past changes (in particular #3750/ #3634 about the benefit of #3438 and #3144, #3513/#3647 about the benefit of #2654, #3855 about the benefit of #2746 and of course #3635) i have the strong impression that we currently lack the ability to critically evaluate past design decisions in cases where there is not a clear technical bug and reach a common assessment and consensus on such.  This in my eyes consequently means we need to be extra careful when making new design choices.

> @imagico - are you ok with using way_pixels to stop rendering the icon/text sooner for large aerodromes which are mapped as areas, as in this line and in lines 1522, 3175 and 3226?

In principle yes, however you need to be careful what way_pixels value you assume for nodes.  If this is a value other than zero that is likely confusing for mappers.

@kennykb - What i tried to point out here is that label scaling based on way_area in this style works well in some cases but sufficiently often is also quite counterproductive as shown so a critical evaluation of this and careful consideration if to extend it to other features are a good idea.

kocio-pl:
> Would you recommend that in this case, @imagico, @kocio-pl?

The short list of what you plan to do currently would be enough for me.

> This in my eyes consequently means we need to be extra careful when making new design choices.

I think another general remark and long comments do not help to make the decision in this PR. It does not answer the key questions of this change, like the one above. 

Please, Christoph, open general tickets for general problems. I seriously think it deserves separate thread and not mixing it with - we agree that already overloaded - PRs.

> Is the new 8 pixel size aerodrome icon too small to be clearly legible at z10? Are you saying that you would prefer that we keep the current 12 pixel icon at z10 or remove the airport icon at z10 for now (though later we might add it back for aerodrome=international if this is used widely enough)?

Removing labels at initial levels is OK for me if there is a problem with cluttering. I said nothing about initial level the icons would start showing, but I prefer to keep them on z10 (just without the labels, if needed).

jeisenbe:
It looks like the current small icon at z10 is not acceptable, and the more complex label and icon size changes at high zoom levels will need to be deferred to another PR, where we can discuss them more generally, so I will close this PR and create a fresh one.

Issue #1693, opened 2015-07-28
Add icon for leisure=hackerspace


miniwark:
[leisure=hackerspace](http://wiki.openstreetmap.org/wiki/Tag:leisure%3Dhackerspace) is used to map hackerspaces, makerspaces (and sometime fablabs). There is actually no icon to render them in Mapnik. Actually there are [365 tagged points](https://taginfo.openstreetmap.org/tags/leisure=hackerspace).

As, almost all this leisures are for DIY stuff, a simple render solution is to use the [shop_diy](https://github.com/gravitystorm/openstreetmap-carto/blob/master/symbols/shop_diy.16.svg) logo.

The other easy alternate solution is to use the [computer]([https://github.com/gravitystorm/openstreetmap-carto/blob/master/symbols/computer-14.svg) logo.

If not accepted i would think of a gear like or lab logo (because this symbol is popular in hacker spaces logos around the world), but i think it is not necessary to create a new special logo, when is possible to use an already existing one.

HolgerJeromin:
duplicate of #451

dieterdreist:
> Am 28.07.2015 um 13:29 schrieb Bertrand Lecervoisier <notifications@github.com>:
> 
> i think it is not necessary to create a new special logo, when is possible to use an already existing one


using the same icon for different stuff reduces the usefulness of the icon for all these things. It does only make sense if the tags are synonyms or very close and should not be differentiated in rendering

kocio-pl:
Is it close to the community centre or not?

HolgerJeromin:
I think the target audience of a community_centre and a hackerspace have no or a really small intersection. :)

kocio-pl:
Thanks. :smile: I was wondering what is community centre really, cause I'm not familiar with such things.

math1985:
> I was wondering what is community centre really, cause I'm not familiar with such things.

I think the Polish equivalent is dom kultury / dom ludowy. After Google'ing, I found http://www.sdk.pl, which I would call a community centre, not an arts centre: they offer various art courses and performances, but also things like computer courses for the elderly.

kocio-pl:
Yet some of the "dom kultury" ("house of culture") POIs should be rather tagged as arts centre (6k+ uses), so it looks like we have a few of similar, but not identical type of objects, and the name is not the ultimate hint:
- amenity=community_centre
- amenity=arts_centre
- club=art (and the likes)

[...]

Issue #1628, opened 2015-07-02
Add rendering for amenity=coworking_space


vdeparday:
Coworking space and innovation hubs are/can be a big part of growing the OSM community, however they do not get rendered on the map, so it would be nice to add these at these two tags to be rendered:
http://wiki.openstreetmap.org/wiki/Tag:amenity=coworking%20space?uselang=en-US
http://wiki.openstreetmap.org/wiki/Tag:amenity%3Dcommunity_centre

There may be some other ones used for similar spaces.

kocio-pl:
I guess we will have a bigger coloring change, including your examples, because [we tend to agree now](https://github.com/gravitystorm/openstreetmap-carto/pull/1450#issuecomment-117227190), that filling color is not universal enough and we don't need it everywhere (too broad) or have many filling colors (that would make a map look like a rainbow), so instead we may end up with color schemes for boundaries.

pnorman:
edited to reflect just `amenity=coworking_space` as the two tags are quite unrelated.

pnorman:
225 uses is insufficient to add as a POI

vdeparday:
Thanks Paul for the quick answer. Is that for technical reasons and maintenance? Because in terms of importance or meaning, I am not sure why if a tag is not used that much, it would be mean that it is not relevant to be shown? Or would that mean there is a flaw in the tag itself?

kocio-pl:
AFAIK there are currently no policy on the amount of tag uses yet. It's just a matter of individual taste and feeling. In general the more occurrences, the more contributors agree it's worth showing on the map, of course. See the comments [in copy shop pull request](https://github.com/gravitystorm/openstreetmap-carto/pull/1615): @pnorman is more demanding in this regard, @math1985 has lower expectations, and mine are even less strict than them both.

The bottom line is you can not agree and try to prove your reasons are justified, but until we have no such policy, the people able to merge the pull requests have final say in practice.

dieterdreist:
2015-07-03 10:48 GMT+02:00 Vivien Deparday <notifications@github.com>:

> Thanks Paul for the quick answer. Is that for technical reasons and
> maintenance? Because in terms of importance or meaning, I am not sure why
> if a tag is not used that much, it would be mean that it is not relevant to
> be shown? Or would that mean there is a flaw in the tag itself?



Completely agree, as has been pointed out previously, pure usage numbers
should not dictate which features get rendered (and haven't been in the
[...]

Issue #451, opened 2014-03-31
Add rendering for leisure=hackerspace


devjb:
I just noticed that the standard renderer does not label leisure=hackerspace in the map. The bike map does render the label. Maybe it might nice to have the names of hackerspaces also in the standard map.

matkoniecz:
It is neither widely used nor passed through wiki proposal process. I propose to close this ticket.

xamanu:
I don't think hackerspaces are a niche thing.

> making icon of hackerspace that would be clearly recognisable would be really hard.

There is quite a good symbol by [Eric Steven Raymond](https://en.wikipedia.org/wiki/Eric_S._Raymond): the glider [hacker emblem](http://www.catb.org/hacker-emblem/)

![hackerspace svg](https://user-images.githubusercontent.com/97706/64209517-40d7ae80-ce90-11e9-84ba-c51f237446bd.png)

Anyway, just made a PR: #3867

dreua:
Rendering it with a blue dot like association/offices would be better than hiding it imho. Can we reopen this to reassess the situation? Or would you rather have a new issue?

The wiki lists it as "de facto", used over 1500 times: https://wiki.openstreetmap.org/wiki/Tag:leisure=hackerspace?uselang=en

I'm considering to tag it as "office" or something else just to have it on the map, but nobody wants that, right?

immanuel-h:
Good point actually, there has always been a close connection between the OSM and hacking communities so this decision to not represent that connection seems rather weird.

imagico:
Reasons for not rendering this other than the low use volume have been provided in #3867.  I particular important:

* the tag seems to be deliberately non-generic and culture specific.  The wiki text is full of _definition by example_ and weasel words like 'normally', 'typically' and 'often'.
* delineation of the tag is unclear, the tag seems widely used for commercially run places as well as community run places.  At the same time it is unclear if something like a community workshop providing classical engineering tools qualifies (and if not: if that is because those are not primarily for recreational use or because they are non-digital).

And keep in mind that our target audience is the potential global map user, we are not a special interest map for groups with strong voices on digital channels.

As i wrote in https://github.com/gravitystorm/openstreetmap-carto/pull/3867#issuecomment-527998308:

[...]

PR #3867, opened 2019-09-03
Render leisure=hackerspace

xamanu:
Adds rendering for leisure=hackerspace on nodes and closed ways ("areas") with an icon based on the [hacker glider](http://www.catb.org/hacker-emblem/) shop color. The icon and name label are rendered at z17 and up, similar to other amenities.

This PR includes the first of the following symbols. Please let me know if you had a different preference of icon or colour and I would be happy to adjust.

foo

### DIFF
--- a/amenity-points.mss
+++ b/amenity-points.mss
@@ -1424,6 +1424,13 @@
      marker-clip: false;
    }
 
+  [feature = 'leisure_hackerspace'][zoom >= 17]  {
+     marker-file: url('symbols/leisure/hackerspace.svg');
+     marker-fill: @shop-icon;
+     marker-placement: interior;
+     marker-clip: false;
+   }
+
   [feature = 'leisure_beach_resort'][zoom >= 16] {
      marker-file: url('symbols/leisure/beach_resort.svg');
      marker-fill: @leisure-green;
@@ -1857,6 +1864,19 @@
     text-wrap-width: @standard-wrap-width;
     text-line-spacing: @standard-line-spacing-size;
     text-fill: @leisure-green;
+    text-dy: 11;
+    text-face-name: @standard-font;
+    text-halo-radius: @standard-halo-radius;
+    text-halo-fill: @standard-halo-fill;
+    text-placement: interior;
+  }
+
+  [feature = 'leisure_hackerspace'][zoom >= 17] {
+    text-name: "[name]";
+    text-size: @standard-font-size;
+    text-wrap-width: @standard-wrap-width;
+    text-line-spacing: @standard-line-spacing-size;
+    text-fill: @shop-text;
     text-dy: 11;
     text-face-name: @standard-font;
     text-halo-radius: @standard-halo-radius;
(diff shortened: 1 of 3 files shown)


imagico:
Thanks for the suggestion.  I would be against this in this form for the following reasons:

* the definition of leisure=hackerspace is fairly vague and largely overlaps with amenity=social_centre and amenity=community_centre.  There does not appear to be a clear delineation towards these.
* the symbol is not going to be intuitively understandable for the majority of map users, it would essentially be an insider code for those who know it but useless and confusing for most.
* we need to keep the number of distinct specialty symbols down for this style to stay usable and maintainable - see #3635, so rendering something as narrow in meaning as this this with a unique symbol is not a good idea.  The only suitable way i see is integrating it into a broader class of features with a common symbol - which however would still require a clear definition and use according to it.

xamanu:
I can follow you arguments, @imagico. What about rendering just as a dot, and render the name text? This would allow hackerspace at least to appear on the main layer.

imagico:
I am not sure this would provide constructive feedback to mappers in the current situtation.  The whole tagging landscape in this domain seems very fluid.  Looking at the use of leisure=hackerspace there are standalone uses of the tag as well as combinations with the mentioned amenity=social_centre and amenity=community_centre but i have also seen combinations with amenity=coworking_space (#1628, which itself overlaps meaning wise with office=coworking) - possibly indicating that those are more commercial workshops with digital manufacturing infrastructure for rent.

My impression is that the whole domain needs some work in properly defining the meaning and the limits of the different tags independent of the culture specific interpretation of certain English language terms.  Like is a community workshop for repairing and maintaining farming equipment a leisure=hackerspace?  Is it not because it is not digital or is it not because it is not used as a leisure activity?  In general i think the whole domain (community infrastructure for common use and cooperation) is a worthy field to render in this style - maybe even to the level that we can throw out some less important things for it.  But doing so would require there to be well defined tags that are used consistently according to a broadly accepted meaning.

jeisenbe:
I read up on the meaning of "hackerspace" on Wikipedia, and it's all rather confusing: 

"A hackerspace is a community-operated, often "not for profit", workspace where people with common interests, such as computers, machining, technology, science, digital art, or electronic art, can meet, socialize and collaborate. Hackerspaces are comparable to other community-operated spaces with similar aims and mechanisms such as Fab Lab, men's sheds, and commercial "for-profit" companies."

This could include all sorts of clubs and workspaces, in theory, and Wikipedia goes on to list spaces in public libraries, schools and universities as options, commercial hackerspaces (for-profit), "Men's sheds" in Australia, and other variations. Some people think this is different than a "makerspace", which involves creating physical things via machining or 3D printin and such, rather than just software, but this is debated. Some hackerspaces seem to be co-working places, others are more like social clubs. Some are democratic, some are operated by a "benevolent dictator for life", others are commercial for-profit places.

It's hard to see how all that can fit under `leisure=hackerspace`.

matthijsmelissen:
We rejected rendering hacker spaces in #451 and #1693.

xamanu:
I can understand the doubts about the icon! But not even considering to render hackerspaces with a dot and a name is not nice. It will lead to bad tagging (people will put anything so it gets rendered) and this way the carto layer hides community spaces over any commercial place such as shops and offices.

jeisenbe:
I believe we would be happy to render more types of community spaces, but it's important that the tags are well designed and clearly defined. As I mentioned, the definition of "hackerspace" seems to include community centers and clubs, as well as commercial co-working spaces, and it's also not clear what sort of activities are required to define a hackerspace versus some other sort of social or community space.

It was also mentioned that the features are frequently also tagged with more common tags like `amenity=social_centre`, `amenity=community_centre` and `office=coworking`. 

We currently render `amenity=community_centre` with a special icon, and all values of `office` (including `coworking`) with a dot and text label. 

I would be open to considering `amenity=social_centre` for rendering, if we can clarify when it is needed instead of the more common tag `amenity=community_centre`.

PR #3874, opened 2019-09-07
Fix low priority layers

### LINKED ISSUE #3860: amenity=parking rendered with two parking icons
This also seems to be related to our switch to ST_PointOnSurface.

Example: https://www.openstreetmap.org/#map=18/51.94370/13.89742

jeisenbe:
Fix low priority layers

Fixes #3860 

- Remove duplicate parking, motorcycle_parking, bicycle_parking, wayside_shrine, wayside_cross and man_made_cross from amenity-points layer.
- Create text-low-priority layer to render text labels of parkings and wayside shrines/crosses
- Move `parking_entrance` to `low-priority-point` layer, to match other parking features
- Fix rendering text labels for `parking` on nodes
- <s>Adjust initial zoom level for `amenity=parking` nodes from z17 to z18 to match `bicycle_parking` and `motorcycle_parking` - (the code would be more complex otherwise). </s> - removed
- Use ST_PointOnSurface to combine `amenity-low-priority` and `amenity-low-priority-poly` into one layer, and re-use the same SQL table for low-priority text labels layer.

## Test rendering:
https://www.openstreetmap.org/#map=18/-12.35622/130.88415

Before
z18-pasteur-road-before

<s>After</s>
z18-pasteur-road-after

After (ST_PointOnSurface):
z18-pasteur-after-st-point-on-surface

### DIFF
--- a/amenity-points.mss
+++ b/amenity-points.mss
@@ -1575,7 +1575,8 @@
   [feature = 'amenity_motorcycle_parking'],
   [feature = 'amenity_parking_entrance'] {
     [zoom >= 14][way_pixels > 900],
-    [zoom >= 18][way_pixels = null] {
+    [zoom >= 17][feature = 'amenity_parking'],
+    [zoom >= 18] {
       [feature = 'amenity_parking'] { marker-file: url('symbols/amenity/parking.svg'); }
       [feature = 'amenity_bicycle_parking'] { marker-file: url('symbols/amenity/bicycle_parking.svg'); }
       [feature = 'amenity_motorcycle_parking'] { marker-file: url('symbols/amenity/motorcycle_parking.svg'); }
@@ -3122,7 +3123,7 @@
   [feature = 'amenity_motorcycle_parking'],
   [feature = 'amenity_parking_entrance'] {
     [zoom >= 10][way_pixels > 900],
-    [zoom >= 18][way_pixels = null] {
+    [zoom >= 18] {
       text-name: "[name]";
       text-size: @standard-font-size;
       text-wrap-width: @standard-wrap-width;
--- a/amenity-points.mss
+++ b/amenity-points.mss
@@ -1554,7 +1554,7 @@
   }
 }
 
-.amenity-low-priority {
+#amenity-low-priority {
   [feature = 'man_made_cross'][zoom >= 16],
   [feature = 'historic_wayside_cross'][zoom >= 16] {
     marker-file: url('symbols/man_made/cross.svg');
@@ -3115,8 +3115,7 @@
   }
 }
 
-.text-low-priority {
-
+#text-low-priority {
   [feature = 'amenity_parking'],
   [feature = 'amenity_bicycle_parking'],
   [feature = 'amenity_motorcycle_parking'],
--- a/amenity-points.mss
+++ b/amenity-points.mss
@@ -713,13 +713,6 @@
     marker-clip: false;
   }
 
-  [feature = 'historic_wayside_shrine'][zoom >= 17] {
-    marker-file: url('symbols/historic/shrine.svg');
-    marker-fill: @man-made-icon;
-    marker-placement: interior;
-    marker-clip: false;
-  }
-
   [feature = 'amenity_police'][zoom >= 16] {
     marker-file: url('symbols/amenity/police.svg');
     marker-fill: @public-service;
@@ -1570,6 +1563,13 @@
     marker-clip: false;
   }
 
+  [feature = 'historic_wayside_shrine'][zoom >= 17] {
+    marker-file: url('symbols/historic/shrine.svg');
+    marker-fill: @man-made-icon;
+    marker-placement: interior;
+    marker-clip: false;
+  }
+
   [feature = 'amenity_parking'],
   [feature = 'amenity_bicycle_parking'],
   [feature = 'amenity_motorcycle_parking'],
@@ -2090,8 +2090,7 @@
     text-placement: interior;
   }
 
-  [feature = 'military_bunker'][zoom >= 17],
-  [feature = 'historic_wayside_shrine'][zoom >= 17] {
+  [feature = 'military_bunker'][zoom >= 17] {
     text-name: "[name]";
     text-size: @standard-font-size;
     text-wrap-width: @standard-wrap-width;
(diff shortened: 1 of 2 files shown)


jeisenbe:
I've also created another branch which has the same changes, but with both `amenity-low-priority` point and poly layers combined into one with `ST_Point_On_Surface`, and the text layer based off of this. 

See: https://github.com/jeisenbe/openstreetmap-carto/tree/parking-combined and commit f43805971f7a90e7f1b54e6e6332138cb33157f7

I can add that commit to this PR if desired, or we can discuss it later, since it's not required to fix these bugs.

jeisenbe:
Rendering with ST_PointOnSurface; position is slightly less central:

z18-pasteur-after-st-point-on-surface

imagico:
I think if you change this you should also use the same approach as in amenity-points.  If ST_PointOnSurface() leads to acceptable label positions is a different issue - if it is good enough for amenity-points it is also good enough for low priority. :wink:

In any case please make sure the rendering rules are the same as before and any change in actual rendering behavior is made separately.  From a quick look it seems you are not rendering parkings with way_pixels<=900 any more, not even at z18+.

kocio-pl:
As a general note: I was wondering what is the difference between the weighted-centroid algorithm carefully tailored by @talaj for Mapnik and stock `ST_PointOnSurface()` function. Sadly the effect is disappointing for me in many places. However for me it is still acceptable (good enough, as you said), so I think we should stick to this at least to avoid double position rendering.

talaj:
According to [this answer](https://gis.stackexchange.com/questions/76498/how-is-st-pointonsurface-calculated) `ST_PointOnSurface()` works the same as the former interior algorithm in Mapnik was working. What is the problem with the new interior algorithm?

kocio-pl:
It was both substantially higher speed and much less code complexity - see https://github.com/gravitystorm/openstreetmap-carto/pull/3712#issuecomment-471359838.

Hm, it would be great to deploy new algorithm directly in PostGIS too, if possible. What do you think about it?

jeisenbe:
> I think if you change this you should also use the same approach as in amenity-points

Done. I've pushed the commit that was shown.

>  please make sure the rendering rules are the same as before and any change in actual rendering behavior is made separately. From a quick look it seems you are not rendering parkings with way_pixels<=900 any more, not even at z18+.

D@mn it Jim, I'm a doctor, not a computer engineer! 

Oops. I'll fix that. 

(If we reconsider this, should teeny tiny parking smaller than 30 x 30 pixels have an icon or text label even at z18?  I suspect z18 was picked back before we had z19 rendered on openstreetmap.org - Perhaps this should be changed to z19 in a separate PR.)

jeisenbe:
>  it seems you are not rendering parkings with way_pixels<=900 any more, not even at z18+

I've pushed a new commit which fixes this, and tested the rendering. Should be ready for review/merge now.

jeisenbe:
Ok, now it's ready: latest commit removes duplicate rendering of wayside_shrine and adds `text-dy: 10` needed for rendering of wayside_shrine name labels. 

Now all the features in the `amenity-low-priority` layer should only render in this layer, I believe, except for waste_disposal areas which were intentionally treated differently than nodes.

talaj:
> it would be great to deploy new algorithm directly in PostGIS too

@kocio-pl It is possible. GEOS library should accept it. Once it's there, it should be doable to convince PostGIS to use it.

amenity-points.mss
```diff
@@ -3181,6 +3115,50 @@
   }
 }
 
+.text-low-priority {
```
imagico: I think you should use an id based selector here since this styling is only used in one layer.

project.mml
```diff
@@ -2196,52 +2188,80 @@ Layer:
     <<: *extents
     Datasource:
       <<: *osm2pgsql
-      table: |-
+      table: &amenity_low_priority_sql |-
         (SELECT
             way,
+            name,
             COALESCE(
-              'highway_' || CASE WHEN highway IN ('mini_roundabout') THEN highway ELSE NULL END,
```
imagico: You are newly allowing mapping various things with polygons which were previously only rendered on nodes.  This should not be part of the bugfix PR.

kocio-pl:
> GEOS library should accept it. Once it's there, it should be doable to convince PostGIS to use it.

Considering weak choices the old algorithm is making, I think that would win/win for us and general audience of GEOS users (which includes some important projects: https://en.wikipedia.org/wiki/JTS_Topology_Suite#Applications_Using_GEOS ). 

The only question is who could do it. @talaj Would you like to try that or you prefer not to?

talaj:
> Would you like to try that or you prefer not to?

@kocio-pl I would like to do that but my spare time and energy is quite limited, so I cannot promise anything.

kocio-pl:
I perfectly understand you, there's no need to make any promises. If you would like to use some help to make it happen, don't hesitate to ask.

amenity-points.mss
jeisenbe: Thanks! I forgot to remove that when the point and poly layers were unified.

project.mml
jeisenbe: Which ones? I thought I caught them all with `AND way_area IS NULL` - only barriers and parking (not parking entrance) should be allowed if way_area is not null?

jeisenbe:
I've pushed a new commit which uses an id based selector for #text-low-priority and #amenity-low-priority, since there are no longer separate point and polygon layers.

project.mml
jeisenbe: I already made test images which show that wayside_shrine and crosses do not render when mapped as areas, so I believe `AND way_area IS NULL` is working, as in `amenity-points`

imagico: Oh, sorry - my bad.

jeisenbe:
@imagico the changes you requested have been made. Ready to approve?

jeisenbe:
Thanks! Yes, I noticed that too. I think we can limit them both to >=z14
but I will make a separate PR. Probably we should open an issue to discuss
the low-priority layers in general.

jeisenbe:
I've pushed a new commit which uses `COALESCE(way_area/NULLIF(POW(!scale_denominator!*0.001*0.28,2),0), 0) AS way_pixels` as in PR#3878 so that text labels render on parking nodes again at z18 and higher. I've tested it.  @imagico, would you have the time to review this new commit?

imagico:
There is no way_pixels condition on z18 and above for parkings so i don't think this change makes sense here.

jeisenbe:
OK, I realize now that the prior commits had already fixed the rendering on nodes, which was broken when done from the `amenity-points` layer. I was thinking that it would be good to use the same code `COALESCE(way_area/NULLIF(POW(!scale_denominator!*0.001*0.28,2),0), 0) AS way_pixels` in the `amenity-low-priority` layer so that in the future, contributors will not be confused why this layer works differently than `amenity-points`.

But that's not really necessary for this PR, so I will revert to the previously reviewed commit.

PR #3878, opened 2019-09-08
Restore rendering of text labels on points

### LINKED ISSUE #3868: Nodes with leisure=sports_centre have no icon or name rendered
Any node tagged as leisure=sports_centre is currently rendered as just an address. There is no icon or name.

examples:
https://www.openstreetmap.org/node/2174921622
https://www.openstreetmap.org/node/4029273583

### LINKED ISSUE #3876: Schools and kindergartens mapped as a node not shown anymore.
Since a few days schools and kindergartens that have been mapped on a node are not shown anymore. This seems to be a bug [issue 3868](https://github.com/gravitystorm/openstreetmap-carto/issues/3868). 

Schools are often mapped as an area that contains the entire school area, but in a case of multiple schools sharing a building or a school and a library in the same building building mapping them on a node makes more sense.

### Expected behavior
The name of the school or kindergarten is shown in a dark-yellow italic font, like until a few days ago.

### Actual behavior
The housenumber is shown instead of the name.

- [A kindergarten shown as a house number](https://www.openstreetmap.org/node/4575567969)
- [A school shown as a house number](https://www.openstreetmap.org/node/5693493497)

For both the name used to be rendered by OSM-carto. This problem is also being discussed on the [Dutch forum](https://forum.openstreetmap.org/viewtopic.php?id=67300), in a discussion that started with [issue 3860](https://github.com/gravitystorm/openstreetmap-carto/issues/3860).

jeisenbe:
**Fixes #3876**
**Fixes #3868**

- Restore rendering of text labels on points, including landcover and islands

## Explanation: 
Prior to #3764 most features in `landcover` had text labels rendered at high zoom level when tagged on nodes. For islands this was the case before  #1444. It appears that text labels of the name for points was unintentionally lost in these two PRs.
This PR restores this rendering by adding a check for way_pixels=null.
Suprisingly, `way_pixels != null` has to be checked for each instance of `[way_pixels > 12000]` or else the points render with the large lancover text size

## Test rendering:

Old rendering, before #3764 but after #1444
z17-landcover-poi-before-3764

Current test rendering:
z17-landcover-poi-text-current-rendering

After this PR:
z17-landcover-poi-test-after

### DIFF
--- a/amenity-points.mss
+++ b/amenity-points.mss
@@ -1745,20 +1745,22 @@
 .text-low-zoom[zoom < 10],
 #text-point[zoom >= 10] {
   [feature = 'place_island'][zoom >= 4][way_pixels > 3000][way_pixels <= 768000],
+  [feature = 'place_island'][zoom >= 16][way_pixels = null],
   [feature = 'place_island'][zoom >= 16][way_pixels <= 768000],
   [feature = 'place_islet'][zoom >= 11][way_pixels > 3000][way_pixels <= 768000],
+  [feature = 'place_islet'][zoom >= 17][way_pixels = null],
   [feature = 'place_islet'][zoom >= 17][way_pixels <= 768000] {
     text-name: "[name]";
     text-fill: #000;
     text-size: @landcover-font-size;
     text-wrap-width: @landcover-wrap-width-size;
     text-line-spacing: @landcover-line-spacing-size;
-    [way_pixels > 12000] {
+    [way_pixels > 12000][way_pixels != null] {
       text-size: @landcover-font-size-big;
       text-wrap-width: @landcover-wrap-width-size-big;
       text-line-spacing: @landcover-line-spacing-size-big;
     }
-    [way_pixels > 48000] {
+    [way_pixels > 48000][way_pixels != null] {
       text-size: @landcover-font-size-bigger;
       text-wrap-width: @landcover-wrap-width-size-bigger;
       text-line-spacing: @landcover-line-spacing-size-bigger;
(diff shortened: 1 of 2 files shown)


kocio-pl:
I remember that you have made similar rendering tests for different features. Do you have the idea how could we make a standard (possibly automated) test for indicating visual changes in every feature we render, to avoid most basic problems, like this one?

jeisenbe:
I don't think there is a universal solution. Often bugs are found in
the things that we forget to test.

kocio-pl:
I mean a list of objects and dumb automated file diff, not high level tests. It is trivial to catch the difference between tile with object and without it. The idea is to do it for every commit (possibly by Travis), so we do not have to remember about running it.

jeisenbe:
Oh, that's interesting. I have no idea how to implement something like that.

It would be difficult to get all the possible combinations in one test
file: many changes have to do with one feature layered on top of
another.

kocio-pl:
@pnorman You have some experience with Travis checks, do you have some idea for implementing such test?

Maybe every feature should be in separate file, to make diffs atomic. If anybody else has some suggestion how can we deal with it, it would be great.

pnorman:
If someone comes up with checks I can help get Travis running them.

jeisenbe:
Thanks. I'll merge this in a couple of days if there are no other comments or objections.

jeisenbe:
> ... simply set way_pixels to 0 instead of NULL in SQL since we essentially treat (and will likely always treat) point features the same way as very small polygon features in cases where both are rendered identically in principle.

@pnorman - any opinion on this idea? Would it slow things down to have more non-NULL fields in the rendering database?

pnorman:
> @pnorman - any opinion on this idea? Would it slow things down to have more non-NULL fields in the rendering database?

We would not be altering the database. We could go for 0.

jeisenbe:
Should we merge this PR now, and then try out the solution with 0 in
the next release, or would it be better to switch now?

jeisenbe:
OK, I've pushed a new commit which changes from `NULL as way_area` to `O as way_area` for the points, and removed the other changes. The test renderings above are unchanged, and all of the amenity-points.mss features in my test rendering appear to still render correctly, including subway entrances and stations (that layer is also affected by the change).

imagico:
I don't mind this approach but what i actually had in mind is changing

https://github.com/gravitystorm/openstreetmap-carto/blob/eae09090b64c361be3ff7bfd1975be7638a6b36c/project.mml#L1537

using `COALESCE(..., 0)`.

The differentiation between NULL and zero is not practically that relevant because polygons with zero area are not valid and AFAIK osm2pgsql does not generate those.  But for clarity selecting the node only feature types with `way_area IS NULL` is still clearer.

jeisenbe:
I thought about that but didn't know how to do it. :-)

I've pushed a new commit using `COALESCE(way_area/NULLIF(POW(!scale_denominator!*0.001*0.28,2),0), 0) AS way_pixels` and switched back to `way_area IS NULL`

jeisenbe:
@imagico or @sommerluk do you have time to review this before Friday?

PR #3898, opened 2019-09-22
Switch addresses to use ST_PointOnSurface

jeisenbe:
### Related to #1644
- This appears to be the last relevant layer that needs to be switched now, unless we wish to switch all polygon layers to use ST_PointOnSurface

- Switch `addresses` to use ST_PointOnSurface
- This makes addresses consistent with `amenity-points` and `building` names.
- Address numbers will no longer be displayed separately than the POI or building name for oddly-shaped polygons

## Test rendering:

https://www.openstreetmap.org/#map=19/50.63295/3.06320

### Passage 57
Before
passage-57-before
After
z19-passage-57-after

### Monop
Before
z19-monop-before
After
z19-monop-after

### DIFF
--- a/addressing.mss
+++ b/addressing.mss
@@ -25,7 +25,6 @@
         }
       }
     }
-    text-placement: interior;
     text-face-name: @book-fonts;
     text-fill: @address-color;
     text-halo-radius: @standard-halo-radius;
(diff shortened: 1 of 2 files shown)


matthijsmelissen:
I thinjk we can/should also get rid of text-placement: interior; in addressing.mss.

jeisenbe:
Thanks, good idea. I have pushed a new commit that fixes this for `addresses` in this PR, but it looks like we also need to remove  `text-placement: interior` from `buildings-text` and from `amenity-points.mss` in a separate PR

jeisenbe:
Thanks, I've made the changes requested

PR #3900, opened 2019-09-24
Use 0 for point areas rather than coalesing

pnorman:
Rather than setting point areas to NULL and turning nulls to 0, we can directly set them to 0.


pnorman:
Ah, good point.

PR #3902, opened 2019-09-24
Use ST_PointOnSurface for bridge names

pnorman:
I didn't identify any labels likely to appear on the same objects, so we shouldn't run into issues like the other cases where moving one label (e.g. shop) but not the other (e.g. addr) results in duplicates.


jeisenbe:
Should the geometry still be `polygon` even though we are producing a point for labeling with `ST_PointOnSurface`, because the original geometry was a polygon?

pnorman:
> Should the geometry still be `polygon` even though we are producing a point for labeling with `ST_PointOnSurface`, because the original geometry was a polygon?

I'll change that, although it doesn't matter that much.

jeisenbe:
Should we also change the geometry to point for all the other layers which now use ST_PointOnSurface?

jeisenbe:
@pnorman is this PR ready to merge?

jeisenbe:
@pnorman there is now a merge conflict that needs resolving

jeisenbe:
@pnorman do you have time to fix the merge conflict?

pnorman:
Not in the next few days

jeisenbe:
@pnorman will you have to work on this in the next month? If you are busy, I'm willing to fix the merge conflict for you, if that is acceptable.

pnorman:
> @pnorman will you have to work on this in the next month? If you are busy, I'm willing to fix the merge conflict for you, if that is acceptable.

If you have time you're welcome to fix it.

PR #3909, opened 2019-09-27
Update Index for Ferry Layers

Nakaner:
Pull request #3727 introduced the new condition osm_id > 0 but did not update the condition for the indexes inteded to be used by the SQL queries of these layers.

### DIFF
--- a/indexes.sql
+++ b/indexes.sql
@@ -4,7 +4,7 @@
 
 CREATE INDEX planet_osm_line_ferry
   ON planet_osm_line USING GIST (way)
-  WHERE route = 'ferry';
+  WHERE route = 'ferry' AND osm_id > 0;
 CREATE INDEX planet_osm_line_name
   ON planet_osm_line USING GIST (way)
   WHERE name IS NOT NULL;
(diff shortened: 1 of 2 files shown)


pnorman:
The old index will still work. I wouldn't expect many ways to match the old index but not the one proposed here, so I'd expect the performance difference to be small.

jeisenbe:
@pnorman - your comments suggest that this change would make a small but positive improvement in the indexes. Do you approve of this change, or are there problems?

pnorman:
> Do you approve of this change, or are there problems?

Operational costs of changing indexes

jeisenbe:
Should it be merged now, for this release, or do you want this delayed until we have another PR that requires updating indexes? For example, we could merge it to the schema-changes branch.

jeisenbe:
@pnorman - is it ok to merge this PR for the next release, or do we need to wait?

pnorman:
I'm not sure why we need it, but if we're going to merge it, now is okay

pnorman:
Let's ship it with the schema changes

jeisenbe:
Plan is to merge this PR at the same time the `schema_changes` branch is merged to `master` - I hope it will be this month.

pnorman:
I merged this in to schema_changes, so it'll get brought in to master when we merge that.

PR #3912, opened 2019-09-28
Don't use classes anymore

sommerluk:
- Do not use classes any more.
- Check with Travis that classes are not used any more.

This is a step towards vector tiles.

It seems surprisingly simple to eliminate the classes. Not sure if I'm missing something?


matthijsmelissen:
No, I think that's it. We only use classes as a shortcut so we don't need to spend multiple id's in some cases. I'm happy with this change, also forward looking to vector tiles.

jeisenbe:
Looks good

PR #3915, opened 2019-09-29
Precedence of junctions over POIs

sommerluk:
- Give junction names and bridge names precedence over amenity-points.

Before:
before

After:
after

Rationale:
Often I see junction names that appear on z15, but later disappear on z16 or z17 because of some shops near to the junction, just to re-appear again on higher zoom levels when there is enough space for both, the junction name and the shop icon. I think the junction name is more important as the shop icon. Furthermore, for more consistency features that have yet appeared on z15 should in general not be hidden later by features that only appear on z16 or z17.

The “bridge” layer was just next to the junction layer. I've moved it also, because I think the same arguments apply here. There should be fewer conflicts anyway, because bridges often pass over water - but not always!

### DIFF
--- a/project.mml
+++ b/project.mml
@@ -1383,6 +1383,52 @@ Layer:
     properties:
       cache-features: true
       minzoom: 12
+  - id: junctions
+    geometry: point
+    <<: *extents
+    Datasource:
+      <<: *osm2pgsql
+      table: |-
+        (SELECT
+            way,
+            highway,
+            junction,
+            ref,
+            name,
+            NULL AS way_pixels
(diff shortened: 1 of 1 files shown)


jeisenbe:
Occasionally this makes a difference with motorway junctions. In this case the text label is currently blocked by power tower icons:

https://www.openstreetmap.org/#map=15/49.6330/6.1800
z15-grunewald-compare

jeisenbe:
Does anyone else want to review this PR? It would be nice to merge it soon so that I can resolve the conflicts with #3933

Issue #3880, opened 2019-09-10
Symbol and label prioritization is not in sync with the starting zoom levels


imagico:
Currently the drawing order of point symbols and labels (layers amenity-points and text-point) but also of other text layers is not in sync with the starting zoom levels as defined by the MSS code in amenity-points.mss.  This leads to in some cases symbols and labels disappearing as you zoom because other symbols and labels appear at a later zoom level but with higher priority.

This is related to #234.

The superficial solution would be to manually work through these layers and adjust the drawing order to match the order of starting zoom levels.  This is however very maintenance intensive. The alternative would be to define the starting zoom levels in SQL and use that for starting zoom level selection rather than MSS logic.

This would also have the advantage it could easily be adapted to dynamic starting zoom levels based on continuous importance rating and local map scale/latitude.

Neither of these solutions would of course affect the same problem when to comes due to labels and symbols from different layers competing with each other.

Examples:

https://www.openstreetmap.org/node/59646817#map=16/47.9973/7.8364
https://www.openstreetmap.org/node/59646817#map=17/47.99652/7.83686

https://www.openstreetmap.org/node/3913744216#map=18/47.99380/7.85089
https://www.openstreetmap.org/node/3913744216#map=19/47.99384/7.85109

sommerluk:
The big question is how the MSS code should look like.

Example:

```cartocss
#amenity-points[zoom >= 10][int_startzoom <= 10],
#amenity-points[zoom >= 11][int_startzoom <= 11],
#amenity-points[zoom >= 12][int_startzoom <= 12],
#amenity-points[zoom >= 13][int_startzoom <= 13],
#amenity-points[zoom >= 14][int_startzoom <= 14],
#amenity-points[zoom >= 15][int_startzoom <= 15],
#amenity-points[zoom >= 16][int_startzoom <= 16],
#amenity-points[zoom >= 17][int_startzoom <= 17],
#amenity-points[zoom >= 18][int_startzoom <= 18],
#amenity-points[zoom >= 19][int_startzoom <= 19],
#amenity-points[zoom >= 20] {
  // Here comes all the existing icon/label code
}
```

[...]

PR #3612, opened 2018-12-30
Render parking at z14 and greater

jeisenbe:
### Changes proposed in this pull request:
- Render parking at z14 and above only
(No longer render parking lots at z10 to z13)

### Explanation:
- Parking lots currently render as soon as z10, and the icon is rendered as soon as the way_pixels (height x width) is greater than 900 (eg 30 pixels wind and high), without consideration of zoom level
- However, very few parking lots are large enough to have an icon at z13 (I have yet to find one), and generally only the fill color is visible
- The fill color for parking is nearly the same as the background land-color, or the color of untagged land, so parking lots are only visible at z13 and lower when they are surrounded by other land uses or natural areas, and they then appear as a "hole" in the landuse or other area.
- This is a particular problem in the USA, where many town centers have more than 50% of the land covered in parking lots, thus obscuring most of the retail or commercial area.
- By stopping rendering parking lots at z10 to z13, it will be easier to see the correct outline of retail, commercial and industrial areas at z13, and developed land in general at z12 and below. 

### Test renderings with links to the example places:

**Kahului, Hawaii**
https://www.openstreetmap.org/#map=13/20.9042/-156.4423
z12 Before
z12-kahului
After
z12-kahului-no-parking

z13 Before
z13-kahului
After
z13-kahului-no-parking

Double Resolution to show details:
Before z12 - double
z12-kahului-master-2x
After - double
z12-kahului-no-parking-2x

Before z13 - double
z13-kahului-master-2x
After - double
z13-kahului-no-parking-2x

**Washington, DC**
z12 Before
z12-washington-master
z12 After
washington-z12-no-parking
z13 Before
z13-washington-master
z13 After
z13-washington-no-parking

**Newcastle and Newark, Delaware**
z12 Before
z12-newark-newcastle-master
After
z12-newark-newcastle-no-parking

z13 Before
z13-newcastle-airport-master
After
z13-newcastle-airport-no-parking

z12 - double resolution
Before
z12-newark-master-2x
After
z12-newar-no-parking-2x

z13 - 4x resolution
Before
z13-newark-master-4x
After
z13-newark-no-parking-4x

**Derry, N. Ireland:**
https://www.openstreetmap.org/#map=13/54.9936/-7.3074
z12 before
z12-derry-master
z12 after
z12-derry-after

z13 before
z13-derry-master
z13 after
z13-derry-after

z12 double resolution - before
z12-derry-master-2x
after
z12-derry-2x-after

**Belfast, N. Ireland:**
[...]

PR #3923, opened 2019-10-05
Move parking back to amenity-points layers, change way_pixels limit and fix initial zoom level

### LINKED ISSUE #3908: amenity=parking icon no longer rendered on buildings
Example illustrating the problem: https://www.openstreetmap.org/way/221032567

This is a regression - amenity=parking icon used to render on buildings.

jeisenbe:
Fixes #3908
Partial revert of #3874 
Follow-up to fix bug in #3612

### 1) Move parking back to amenity-points layers
- Moves `amenity=parking`, `amenity=motorcycle_parking`, `amenity=bicycle_parking` and `amenity=parking_entrance` to `amenity-points` and `text-point` layers instead of `amenity-low-priority` and `text-low-priority`
- This fixes bugs where the building text or address number is shown instead of the icon and parking text, introduced by #3874 
- Parking had previously been rendered in both amenity-points and amenity-low-priority.
- Recently PR #3874 had consolidated parking rendering in low-priority only, but this leads to problems when there is a building or an address on the same feature
- This commit moves parking back to the amenity-points layer, but near the end. 
- The minzoom for text-low-priority is now changed back to z17 since there are no longer any features rendered sooner that this in the layer.

### 2) Change initial zoom level to be z14 for both text label and icon
- This bug was introduced by my PR #3612 where the initial zoom was changed to z14 for the icon, but not for the text. 
- In PR #3612 it was planned to render parking from z14 only, however this was not changed for the text labels, only for the icon.
- This commit fixes the parking rendering at z10 to z13, though in practice almost no parking lots are large enough to have been rendered at these levels.

### 3) Change way_pixels limit for parking icon and text
- Start showing icon at 750 pixels instead of 900, but don't show text label till 3000 way_pixels, as with most other features
- Only `amenity=marketplace` and parking have an icon that appears based on `way_pixels`. The other uses 3000 way_pixels, which is the limit also used for all text labels.
- Since the icon is only 14 x 14 pixels, it's reasonable to show it one zoom level sooner than the text, at 750 way_pixels (1/4 of 3000 - so 1 zoom level sooner)

### Hanau, Germany
https://www.openstreetmap.org/way/221032567#map=17/50.13289/8.92118
z16 compare current (left), after (right)
z16-parkhaus-compare

z17
z17-parkhaus-after

z18
z18-parkhaus-compare

### Montpellier, France
https://www.openstreetmap.org/#map=17/43.61338/3.88557

z16 comparison (left before, right after)
z16-parking-joffre-compare

z17
z17-parking-joffre-compare

### DIFF
--- a/amenity-points.mss
+++ b/amenity-points.mss
@@ -2929,7 +2929,7 @@
   [feature = 'amenity_bicycle_parking'],
   [feature = 'amenity_motorcycle_parking'],
   [feature = 'amenity_parking_entrance'] {
-    [zoom >= 10][way_pixels > 900],
+    [zoom >= 14][way_pixels > 900],
     [zoom >= 18] {
       text-name: "[name]";
       text-size: @standard-font-size;
--- a/amenity-points.mss
+++ b/amenity-points.mss
@@ -1434,21 +1434,6 @@
       marker-fill: @man-made-icon;
     }
   }
-}
-
-#amenity-low-priority {
-  [feature = 'man_made_cross'][zoom >= 16],
-  [feature = 'historic_wayside_cross'][zoom >= 16] {
-    marker-file: url('symbols/man_made/cross.svg');
-    marker-fill: @religious-icon;
-    marker-clip: false;
-  }
-
-  [feature = 'historic_wayside_shrine'][zoom >= 17] {
-    marker-file: url('symbols/historic/shrine.svg');
-    marker-fill: @man-made-icon;
-    marker-clip: false;
-  }
 
   [feature = 'amenity_parking'],
   [feature = 'amenity_bicycle_parking'],
@@ -1465,6 +1450,21 @@
       marker-fill: @transportation-icon;
       [access != ''][access != 'permissive'][access != 'yes'] { marker-opacity: 0.33; }
     }
+  }
+}
+
+#amenity-low-priority {
+  [feature = 'man_made_cross'][zoom >= 16],
+  [feature = 'historic_wayside_cross'][zoom >= 16] {
+    marker-file: url('symbols/man_made/cross.svg');
+    marker-fill: @religious-icon;
+    marker-clip: false;
+  }
+
+  [feature = 'historic_wayside_shrine'][zoom >= 17] {
+    marker-file: url('symbols/historic/shrine.svg');
+    marker-fill: @man-made-icon;
+    marker-clip: false;
   }
 
   [feature = 'railway_level_crossing'][zoom >= 14]::railway,
@@ -2924,9 +2924,7 @@
     text-halo-fill: @standard-halo-fill;
     text-face-name: @standard-font;
   }
-}
 
-#text-low-priority {
   [feature = 'amenity_parking'],
   [feature = 'amenity_bicycle_parking'],
   [feature = 'amenity_motorcycle_parking'],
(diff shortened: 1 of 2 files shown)


jeisenbe:
I tried the previous rendering at 900 way_pixels for the icon and text, but I think this often renders the text outside of the parking area. This can be confusing, and over-emphasizes small parking lots.

Here's Hanau (same as above) at z17 but with 900 way_pixels limit:
https://www.openstreetmap.org/way/221032567#map=17/50.13289/8.92118 
z17-parking-compare-900-way_pixels

This doesn't work well for long, narrow parking areas.

jeisenbe:
Does anyone have time to review this PR? I would love to get it in the next release if possible.

matkoniecz:
> amenity-points and text-point layers instead of amenity-low-priority and text-low-priority

It was intended as to prevent parking icon from blocking for example shop icon or museum icon. Is it now achieved using a different method or considered as unimportant or it was not working anyway?

jeisenbe:
The problem was that parking garages, tagged `building=*` + `amenity=parking` were rendered with the building text and no icon, since `amenity=parking` was only in the low-priority layer after my last PR. Also, any other amenity-points icon would block parking features, even though they would otherwise render rather early if they are large enough (z10 minimum in theory, though in practice at z14 for very large parking lots). This worsens issue #3880.

To properly get the priority of icons right in all cases, we need to fix #3880 in a general way, or consider how we want to set the priority of all the amenity-points icons (and text) in a consistent way.

matkoniecz:
> any other amenity-points icon would block parking features

 That is why parkings were in both amenity and amenity-low-priority (maybe it was overengineered and supporting only just parkings and therefore not a proper way to do this) but it seemed to work.

jeisenbe:
After PR #3612 `amenity=parking` icons and text were rendering in the regular `amenity-points` layers as soon as z14  if imported as a large enough polygon (>900 square pixels at each zoom level). From z17 points and small areas were also rendered in the `low-priority` layers. Potential duplicate rendering was blocked by the later icons and text colliding with the previously rendered icon or text. But this then led to a visible duplicate rendering when we switched the `amenity-points` layers to use ST_PointOnSurface, before switching the `low-priority` layer.

I initially tried fixing this by switching all parking into the `low-priority` layer only, but #3908 is a problem - icons and the parking text style never showed for parking garages, which are usually also tagged `building=`

In theory we could restore the previous duplicate rendering, with only larger polygons rendered in the main `amenity-points` layers, and nodes/small areas only rendered in `low-priority`, but this would risk #3908 reappearing in the future if changes were made to how the label/icon position is calculated for one of the layers.

But I would recommend merging this PR now, which will fix the rendering in most cases, and then discuss next week whether the nodes / small parking areas should instead be rendered in the `low-priority` layer. Alternately, we might think about how to solve the problems with prioritization in a general way by fixing #3880

jeisenbe:
@matkoniecz - would you like to request changes formally, or do we have consensus to merge this PR today, before the release of v4.24.0?

matkoniecz:
I just wanted to check whatever it was a deliberate decision to drop this part. I am not going to block merge.

jeisenbe:
OK, I will merge this now and we can see if we can solve the prioritization problem in general, but if not then we can consider using the low-priority layer again.

Issue #2909, opened 2017-10-26
`man_made=silo` is not being rendered


ThomasA:
I recently changed a lot of silos and storage tanks along Aalborg Harbour from `building=yes` to `man_made=silo` and `man_made=storage_tank`, because I find that a more accurate reflection of what is actually there. However, now they are no longer rendered on openstreetmap.org. Why does it not render these - physically quite prominent - features?

kocio-pl:
#588 was closed, but can be reopened (see also #372). Please change name of this ticket to drop `man_made=silo`, so the tickets don't overlap.

For every such object we need an idea how should they be rendered. Do you have one?

ThomasA:
Sorry, didn't see that issue...

dieterdreist:
2017-10-26 10:53 GMT+02:00 Thomas Arildsen <notifications@github.com>:

> I recently changed a lot of silos and storage tanks along Aalborg Harbour
> from building=yes to man_made=silo and man_made=storage_tank, because I
> find that a more accurate reflection of what is actually there. However,
> now they are no longer rendered on openstreetmap.org. Why does it not
> render these - physically quite prominent - features?
>


IMHO you shouldn't replace tagged properties ("is a building in OSM") with
different properties, if both apply. You can keep the building tag (or
replace it with a more accurate value like "silo") and still add the
man_made tags. Obviously only in case, the tagged objects are considered
buildings in OSM (to see this, I'd ask on the tagging list, look up the
wiki _and taginfo_).

matkoniecz:
I suspect that silos should also have a building tag.

matkoniecz:
I propose to close this issue and reopen it if there are man_made=silo objects that should not be tagged as building.

matkoniecz:
If there are man_made=silo where building tag is wrong it would be a good idea to give example (with photo) at https://wiki.openstreetmap.org/wiki/Tag%3Aman_made%3Dsilo

In my very limited experience building tag is applicable for such structures.

kocio-pl:
I reopen this as I made a mistake closing this - bunker silo needs a different visual representation probably and I think silo is closer to storage tank (closed space), see the proposition for rendering them: https://github.com/gravitystorm/openstreetmap-carto/issues/588#issuecomment-373903971.

matkoniecz:
> I reopen this as I made a mistake closing this

Do you have any example of object correctly tagged with man_made=silo where adding also building=silo would be wrong?

[...]

Issue #588, opened 2014-05-31
Suggestion for man_made=storage_tank


JoshuaGore8:
The below is my suggestion for rendering the feature man_made=storage_tank in main OSM layer:
![storage_tank](https://cloud.githubusercontent.com/assets/7251556/3138970/8cf7fad6-e8bd-11e3-9186-3bd72d96b0d2.png)

simonpoole:
Sorry for adding to this and essentially reopening. Fhe current numbers for storage_tank use are

145 900 man_made storage_tank
vs
12 762 building storage_tank

(this is likely due to presets). IMHO rendering the former would make sense (but not necessarily anything differnt than for building=storage_tank).

jojo4u:
This is man_made instead of building. Man_made is also used for silo, water_tower, gasometer and reservoir_covered. The [feature page](https://wiki.openstreetmap.org/wiki/Tag:man_made%3Dstorage_tank) is now in good shape and usage is now 153 467 : 13 560.

ThomasA:
Simply *not* rendering `man_made=storage_tank` (and `man_made=silo` as well, for example) seems like almost the worst solution of all. At least the tanks and silos I am thinking of are often quite prominent features on the ground. They often stand out much clearer from the landscape than surrounding buildings and not rendering them at all when they are actually tagged seems like a mistake.
If no consensus can be reached on the actual appearance, how about simply rendering them with the same signature as `building=industrial`? At least they show up on the map then.

ThomasA:
To elaborate a bit: These are examples of the silos and storage tanks that I think it is a problem not to render:
![Silo](https://imageservice.nordjyske.dk/images/nordjyske.story/2012_05_15/1b016c1e-d89f-4343-b57a-a3fde3cd4d4e.jpg)
![Storage tanks](http://scm.dk/sites/default/files/oversigtsbillede_oliehavnen.jpg)

math1985:
Shouldn't they be tagged as building=yes in addition? At least the first one seems to be a building.

We can fix it on our side, but I don't think we should expect all other renderings to follow suit.

kocio-pl:
The question is not only how to render such polygons (143 941), but also how to render nodes (56 091)?

ThomasA:
Combining it with `building=yes` would seem an easy fix, but [the documentation](http://wiki.openstreetmap.org/wiki/Tag:man_made%3Dstorage_tank) says:

[...]

PR #3928, opened 2019-10-09
Add 14px storage tank icon used by iD

### LINKED ISSUE #3927: Reuse iD editor storage tank icon
> If you think that something should be changed - please open a new issue.

_Originally posted by @matkoniecz in https://github.com/gravitystorm/openstreetmap-carto/issues/588#issuecomment-539653830_

The current tall storage tank icon looks like a gasoline pump.

It would just be safer to use the storage tank icon from the iD editor. Thanks.

bhousel:
closes #3927 
I named it storage_tank2.svg, so it wouldn't collide with the existing icon, but you can move it if you want.

Anyway, there are a bunch more icons over in [bhousel/temaki](https://github.com/bhousel/temaki).  They are 15px to match what maki does, but redrawing them to pixel align to 14px is usually not a big deal. Enjoy!

### DIFF
--- a/project.mml
+++ b/project.mml
@@ -214,7 +214,7 @@ Layer:
             CASE WHEN tunnel IN ('yes', 'culvert') THEN 'yes' ELSE 'no' END AS int_tunnel,
             'no' AS bridge
           FROM planet_osm_line
-          WHERE waterway IN ('river', 'canal', 'stream', 'drain', 'ditch', 'wadi')
+          WHERE waterway IN ('river', 'canal', 'stream', 'drain', 'ditch')
             AND (bridge IS NULL OR bridge NOT IN ('yes', 'aqueduct'))
           ORDER BY COALESCE(layer,0)
         ) AS water_lines
@@ -594,7 +594,7 @@ Layer:
             WHERE barrier IN ('chain', 'city_wall', 'embankment', 'ditch', 'fence', 'guard_rail',
                   'handrail', 'hedge', 'kerb', 'retaining_wall', 'wall')
               OR historic = 'citywalls'
-              AND (waterway IS NULL OR waterway NOT IN ('river', 'canal', 'stream', 'drain', 'ditch', 'wadi'))
+              AND (waterway IS NULL OR waterway NOT IN ('river', 'canal', 'stream', 'drain', 'ditch'))
           ) AS features
         ) AS line_barriers
     properties:
@@ -916,7 +916,7 @@ Layer:
             CASE WHEN tunnel IN ('yes', 'culvert') THEN 'yes' ELSE 'no' END AS int_tunnel,
             'yes' AS bridge
           FROM planet_osm_line
-          WHERE waterway IN ('river', 'canal', 'stream', 'drain', 'ditch', 'wadi')
+          WHERE waterway IN ('river', 'canal', 'stream', 'drain', 'ditch')
             AND bridge IN ('yes', 'aqueduct')
           ORDER BY COALESCE(layer,0)
         ) AS waterway_bridges
@@ -2120,7 +2120,7 @@ Layer:
               THEN 'yes' ELSE 'no' END AS int_intermittent,
             CASE WHEN tunnel IN ('yes', 'culvert') THEN 'yes' ELSE 'no' END AS int_tunnel
           FROM planet_osm_line
-          WHERE (waterway IN ('river', 'canal', 'stream', 'drain', 'ditch', 'wadi')
+          WHERE (waterway IN ('river', 'canal', 'stream', 'drain', 'ditch')
                  OR "natural" IN ('bay', 'strait'))
             AND (tunnel IS NULL or tunnel != 'culvert')
             AND name IS NOT NULL
--- a/water.mss
+++ b/water.mss
@@ -122,8 +122,7 @@
 
 .water-lines {
   [waterway = 'canal'][zoom >= 12],
-  [waterway = 'river'][zoom >= 12],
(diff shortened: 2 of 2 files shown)


kocio-pl:
It's OK for me. 

Since you mean to replace it, it's better to rename it, we don't collect all the possible icons if not used here currently. However one has to check if this is not being used for different tagging and then decide how should it look like.

jeisenbe:
Looking back at the previous discussion in the issues #2909 and #588, we did not really reach consensus about rendering `man_made=storage_tank` with an icon instead of similar to buildings when tagged on areas. See https://github.com/gravitystorm/openstreetmap-carto/issues/2909#issuecomment-340767093,  https://github.com/gravitystorm/openstreetmap-carto/issues/2909#issuecomment-390165339, and https://github.com/gravitystorm/openstreetmap-carto/issues/588#issuecomment-45058778 plus following comments and https://github.com/gravitystorm/openstreetmap-carto/issues/588#issuecomment-339616474 and https://github.com/gravitystorm/openstreetmap-carto/issues/588#issuecomment-339624211 about rendering like buildings. 

Also see discussion about tagging on https://wiki.openstreetmap.org/wiki/Talk:Tag:man_made%3Dstorage_tank#building.3Dyes_.3F and elsewhere on that same Talk page.

It seems that a number of mappers feel that `man_made=storage_tank` and `man_made=silo` are sufficient, without a `building` tag, though many also add a `building=` tag.

Should we start rendering `man_made=storage_tank` and `=silo` like buildings? If so, it's probably not necessary to show an icon when they are mapped as areas.

Should silos and storage tanks have the same icon? This storage tank icon looks more like a tank, but less like a silo, than the current icon.

bhousel:
I have no opinion about the rendering.

There is a silo icon too if you want it.. (they do look different from storage tanks)
preview page here:  https://bhousel.github.io/temaki/docs

jeisenbe:
@bhousel - I'm guessing that you are quite busy with iD and probably do not have time to develop this PR further. Is that correct? If so, is it ok if we edit it for you, or is it better for us to make a different PR if further changes are required?

jeisenbe:
Rendering comparison at z18. These storage tanks, like many (56%), are also tagged with `building=`

z18-storage-tanks-buildings-compare

jeisenbe:
Comparison of silos:

z19-silos-compare
z18-silos-compare

bhousel:
> I'm guessing that you are quite busy with iD and probably do not have time to develop this PR further. Is that correct? If so, is it ok if we edit it for you, or is it better for us to make a different PR if further changes are required?

1. I don't really work on iD anymore.
2. We discussed storage tanks here:  https://github.com/openstreetmap/iD/issues/4265 .  They should probably have a building tag when mapped as an area.
3. You can make whatever changes to the PR you want - "allow edits from maintainers" is checked, and I wasn't planning on doing any follow on work.

imagico:
A few comments:

* i think neither the current symbol nor the newly proposed one is ideal but i also don't think the newly proposed one is worse so i would not mind a one-to-one replacement if otherwise seen positively
* in general when @nebulon42 re-designed many point symbols of this style only having profile views and no perspective views was part of his design paradigm.  This was not fully maintained later in all cases but that is the reason why we have many symbols like the current storage tank symbols.
* i think any more elaborate change in how things are rendered in this field should be preceded by an analysis of actual tag use (like combination with building tags, relationship to other semantically similar tags etc.).

polarbearing:
I would not require a building tag to render the silo/storage. I can imagine various construction types, for some I'd be inclined as a mapper to add 'building' to some of them, and explicitly not on some others. 
For comparison, we discourage 'building' on man_made=bridge with a few notable exceptions.

HolgerJeromin:
Both icons are pretty heavy with no real value. Who uses this map for searching for storage tanks?
Showing the outline would be ok imo, and why not with the bu

jeisenbe:
So far @kocio-pl has said this new icon is "ok", which I believe means it's better than the current icon, @imagico does not mind the replacement though it's not entirely clear if he thinks it is an overall improvement, and @HolgerJeromin does not like the use of an icon, but has not commented if the new icon is an improvement on the old one. 

I think the new icon is more legible, but this is because it uses a perspective view, partially from above, rather than a straight-on, side view. As mentioned, this would be a departure from our usual icon design style, but it might be helpful to allow such variation.

Does anyone else have an opinion about the new icon?

Should we allow icons which are designed with a perspective view?

Adamant36:
I say go for it. It's better then no icon. I thought there was a discussion somewhere a while back about if storage tanks count as buildings that was inconclusive. I can't seem to find it though, but I think whatever the case is, it's best to go with what is already established. Which in this case is an icon and the current one clearly needs changing. Maybe allowing for a different perspective will lead to improvements in other icons later on also.

kocio-pl:
> So far @kocio-pl has said this new icon is "ok", which I believe means it's better than the current icon, 

I meant that it is also acceptable for me. To be precise, it has some downsides compared to a current icon - it uses perspective, which we try to avoid, so I would not point out which one is really better.

jeisenbe:
That was silly of me. I tried to update this PR, but pushed my master branch instead of the changes, and this apparently closed the PR. And now that the PR is closed, I can no longer push changes to this branch. Is there any way to fix this remotely?

imagico:
I am not sure but i think once there is a change on the branch in question you can reopen.

jeisenbe:
@bhousel - mea culpa, could you please run `git push -f`?

bhousel:
> mea culpa, could you please run `git push -f`?

it didn't seem to work, so I just opened another PR.

Issue #3883, opened 2019-09-12
Render waterways over natural=wood pattern


LucFreitas:
Waterways are rendered below the wood pattern. It seems more appropriate to me that the waterway overlaps the natural=wood.

![wood](https://user-images.githubusercontent.com/19976610/64798896-2ec5d200-d55a-11e9-8099-f57c01d85df1.png)

(https://www.openstreetmap.org/#map=17/-20.33539/-40.54868)

Pikse:
Waterway lines were displayed over wood/forest until very recently, as opposed to water areas that weren't. It probably isn't useful to map narrow waterways as area and these narrow waterways, even if in forest, are not necessarily covered by canopy. No overlap also looked kind of cleaner to me. So it would make sense to me if layer order was restored the way it was.

imagico:
What has been changed recently is making the layer ordering more consistent with the way things are rendered.  Rendering water linework above the landcover overlay while rendering it in the same color as water areas has always been a strange quirk and a source of confusion for the map user.

The reason why the landcover overlay is rendered above the water layers is

* because for wetlands in particular this is necessary for correct rendering since the wetland edge does not necessarily coincide with the edge of the water and the wetland mapping should be accurately shown in cases of overlap.
* because it provides better feedback on accurate mapping of landcover.

As @jeisenbe explained it would in principle be possible to make an exception for water lines while keeping the logical drawing order using some comp-op magic.  But i am not sure if that would actually be desirable for the pictorial pattern overlays (wood/scrub/wetland).

jeisenbe:
I think we should stop referring to composition operations as "magic."

It's just cutting things out instead of drawing on top.

Then we add the new layer behind instead of on top.

Nothing magic to it.

LucFreitas:
The Ribeirão Santo Agostinho is large enough to be a river, but the rainforest is too dense to show it. You cannot determine its width with aerial images at the points it passes under the forest. Not only this, thousands of other rivers in the Amazônia or Mata Atlântica, or any other rainforest in the world will exhibit the same configuration.

Also, one of the basics of cartography is generalization. You demand too much when you ask to map the area of a river. A mapper will take twice or triple the time and effort to map the margins, while it can only map the riverbed. Unfortunately we do not have the same labor density as there is in the US, Germany or the UK.

I don't know of any other maps that make the same way. Overlaying a forest on a highway follows the same logic.

- - - - 
pt/br

[...]

Issue #3926, opened 2019-10-09
Restore ocean outline


jeisenbe:
### Prior behavior
Between PR #3065 and PR  #3694 (which reverted the prior PR while switching to ocean polygons), the oceans were outlined at z0 to z7. This was not done at higher zoom levels because it would lead to a line across river mouths and other places where inland water bodies meet the coastline. The coastline was also simplified to make this look better, which led to some problems: see #3695. But in PR #3694 we switched from land polygons to ocean polygons, and the outline was removed.

### Proposed solution

It would be possible to restore the outline based on the water polygons, if we use `comp-op: dst-out` and `comp-op: dst-over` as discussed in #3854, and use a different color for oceans and rivers as discussed in #3895 and #3896 - this would allow drawing an outline or gradient around the oceans, without having this line overlap with the land at places where the coastline is deeply indented, in addition to other benefits. The line would not be more visible across river mouths, but will actually make the transition between rivers and ocean more subtle and subjectively pleasing (IMO).

The coastline outline could be in the same color as lakes or rivers, or a gradient made of 2 or 3 different colors that transitions between the ocean and river color. 

This outline would somewhat emphasize small islands, so it would need to be small at lower zoom levels, but could be somewhat wider at high zoom levels. 

### Screenshots illustrating the idea:

z15 Age river mouth, Latvia - current
![z15-age-before](https://user-images.githubusercontent.com/42757252/66454115-b8a58400-eaa1-11e9-9b9f-8ff6f09dcc62.png)

z15 Age - Separate colors for ocean and rivers, 1 pixel wide outline for ocean
![z15-age-lowcontrast-1pxoutline](https://user-images.githubusercontent.com/42757252/66454150-de328d80-eaa1-11e9-9d8d-6212682bf37a.png)

z15 Age - 3 pixel gradient outline
![z15-age-lowcontrast-3px](https://user-images.githubusercontent.com/42757252/66454162-f0143080-eaa1-11e9-872a-c21b76fcfc7a.png)

z17 before
![z17-age-before](https://user-images.githubusercontent.com/42757252/66454195-10dc8600-eaa2-11e9-8d2c-e5e7fa254200.png)

z17 1 pixel outline
![z17-age-lowcontrast-1pxoutline](https://user-images.githubusercontent.com/42757252/66454207-1df97500-eaa2-11e9-90c2-115f51fda58b.png)

z17 3 pixel gradient
[...]

Issue #3895, opened 2019-09-21
Render ocean and seas differently than fresh water


jeisenbe:
### Expected behavior
- The location of the `natural=coastline` should be clearly visible to provide good mapper feedback
- There should be a visible difference between sea water and inland fresh water
- The color of the seas should be less intense than inland waters, since they take up a very large part of the rendering at lower zoom levels
- Mappers should not be able to use `natural=water` polygons over the oceans to render labels for seas

### Actual behavior
- The ocean polygons are rendered in the same color as all inland water areas
- Mappers do not have feedback on the location of the coastline transit in debatable areas near estuaries and rivers. It's not obvious if another user changes coastline to natural=water or waterway=riverbank
- Since the ocean and `natural=water` are the same, some mappers tag areas of seas with `natural=water` to get a label rendered.

### See Also #3893 - rendering salt water lakes differently

kocio-pl:
> Mappers should not be able to use natural=water polygons over the oceans to render labels for seas

How do you expect to achieve this?

jeisenbe:
> How do you expect to achieve this?

Rendering oceans in a different color and moving water areas back above oceans would work, but this conflicts with the first point: "The location of the natural=coastline should be clearly visible to provide good mapper feedback" - that is more important, so I believe the oceans polygons should continued to be rendered above the other water areas.

jeisenbe:
It's quite common for printed maps to render marine water bodies in a lighter shade of blue than lakes or rivers. Here are examples from atlases and wall maps in my kids' school (which also show rivers in a much darker blue, as is common):

![Southern Netherlands-Atlas](https://user-images.githubusercontent.com/42757252/65383123-57fd1400-dd4b-11e9-9b6b-e6006708155c.jpg)

![Shanghai-Wall Map](https://user-images.githubusercontent.com/42757252/65383125-5b909b00-dd4b-11e9-87a5-dca4bee8e58e.jpg)

![Bangladesh-Atlas](https://user-images.githubusercontent.com/42757252/65383127-60554f00-dd4b-11e9-9eb4-9956e8262945.jpg)

StyXman:
What about salty lakes? https://en.wikipedia.org/wiki/Salt_lake#List

jeisenbe:
See issue #3893 about rendering lakes differently when tagged with
`salt=yes` - I agree this is also a good idea to consider at the same
time.

kocio-pl:
[...]

PR #3930, opened 2019-10-10
Create darker river-color for river & canal areas and ditch, drain, stream, canal & river lines

### LINKED ISSUE #3896: Rivers and canal lines and areas should be darker than other water areas
### Expected behavior
- River and stream lines should be a darker and more visible blue color than lakes and seas, so that they can be seen at low zoom levels when over dark landcover like forests.
- For the same reason, river areas should be darker and the same color as river/stream lines
- Areas of `waterway=riverbank` and `water=river` should be distinguished from lakes, reservoirs, ponds and other non-flowing water bodies
- Mappers should get feedback about river areas that are missing `water=*` tags

### Actual behavior
- All areas of water and water lines are rendered in the same color
- No feedback is given if a river is mapped incorrectly with just `natural=water`
- River lines are not very easy to follow on areas of forest, especially at lower zoom levels

This issue blocks merging PR #3670 to fix #3513 and possibly #3647

jeisenbe:
Fixes #3896
Related to #3895 and #1781
Required for #3926 

- The color of river, canal, ditch and drain lines is changed to `#8fcadd` (Lch78,21,227) to be more visible over forest and other dark polygon fills.
- The color of river and canal areas is changed to match. This also improves mapper feedback about tagging of water areas

## Explanation:
-Narrow river areas and lines are difficult to see on woodland areas and some other dark landcovers currently. This is particularly visible at mid and lower zoom levels. 
- Most mid-scale paper maps use a much darker and more saturated blue for river lines to make it easier to visually follow these linear features. The openstreetmap.de rendering currently does this as well. But currently all rivers and streams are rendered the same as water areas in this style. 
- Since rivers lines and areas sometimes overlap (when the area is nearly the same width as the line), the linear and area rendering needs to be the same color to prevent visual artifacts.
- Using a different color for `waterway=riverbank` and `water=river` areas also provides feedback when rivers are not properly tagged, or when the transition from river to lake or sea is at the wrong location.
- In this PR, the color and pattern of intermittent river areas has not yet been adjusted. This will be done in the next PR, since it may require rethinking the pattern rather than just changing the color. 

### Skjolden, Norway
https://www.openstreetmap.org/#map=13/61.5033/7.6847
Before z13
z13-skjolden-before-13-61 5033-7 6847

After
z13-skjolden-river-color

### Sogndal, Norway
https://www.openstreetmap.org/#map=15/61.2285/7.1051
Before z15
z15-sogndal-before

After
z15-sogndal-lowcontrast


jeisenbe:
Årdalstangen, Norway
https://www.openstreetmap.org/#map=14/61.2393/7.7130
- Ocean lower left, lake upper right

z14 before
z14-ardalstangen-before-small

z14 after
z14-ardalstangen-river-color-small

Førde, Norway
https://www.openstreetmap.org/#map=13/61.4441/5.8803
z13 before
z13-forde-before-se

z13 after
z13-forde-se-river-color

z15 before
z15-forde-nw-before

z15 after
z15-forde-river-color

jeisenbe:
Bremerhaven 
https://www.openstreetmap.org/#map=14/53.5468/8.5955
- Shows ditch, river, dock and coastline. 

z14 before
z14-bremerhaven-before

z14 after
z14-bremerhaven-river-color

imagico:
Note this significantly interferes with implementing #3854 so we should make a choice of either first implementing water color differentiations or first implementing #3854.  Mixing the two things would complicate making changes.

I am open to either variant.

I have not tested it yet but do i read the code right that you differentiate rivers and non-rivers but only if they are non-intermittent?  That might be a bit confusing for the map user.

jeisenbe:
> do i read the code right that you differentiate rivers and non-rivers but only if they are non-intermittent?

Only the areas are not yet distinguished, the linear renderings all use the new `river-color`. 

I was too lazy to make and test a new file for intermittent river areas so I just kept `symbols/intermittent_water.png` for now. I think the current intermittent water pattern does not work for rivers, so we need another PR to fix that, rather than just updating the file with the new river-color.

jeisenbe:
> Note this significantly interferes with implementing #3854 so we should make a choice of either first implementing water color differentiations or first implementing #3854

I'd be happy to do either order, but #3854 is more work, while this PR seems to have a good chance of being accepted soon.

Also I thought it would be easier to settle on the new water colors before trying to adjust the rendering of intertidal features.

imagico:
In reference to https://github.com/gravitystorm/openstreetmap-carto/issues/3926#issuecomment-540463678 - in case you are not aware, this is how your above samples look like to me in the browser:

screenshot

jeisenbe:
Crazy, that's quite different that what I'm seeing. The ocean color is #A8D3DE in both the original images according to the color picker in Terminal. Perhaps this happed because the "before" pictures are screenshots from a 1.0 opacity overlay of the current rendering, and the "after" are exports from Kosmtik? I'll avoid doing this in the future (I'll lose this nice fiberoptic connection in a few days anyway, so it's back to rendering everything instead of downloading tiles)

jeisenbe:
We could still do part of #3854: the re-layering and use of `comp-op: dst-out`, which would fix #3883 and related issues while making it possible to fix other issues like #3102 and #3926.

Then there could be a second PR to adjust the rendering of all the features over water and remove the mud transparence once the new colors are settled.

kocio-pl:
The fundamental problem I have with this solution is still the same - the inner water borders are sharp.

The rendering color shift is what i get when using Chromium/Chrome, it is known issue with this browser. That's why I also tried to export from Kosmtik.

jeisenbe:
> the inner water borders are sharp.

I'm happy to fix this for the ocean/river border with #3926, but that's only possible after we start using the darker river color. Recall that in the past we only had the outline for land polygons from z0 to z7 because at mid and high zoom levels there was a strong line across the river/coastline transit. To avoid that, we have to change the river color first.

kocio-pl:
I have to recall this when I have a bigger bit of conciousness, at the moment I don't remember this problem.

But this means that I'm looking for gradient border on inland waters first. Since riverbeds are usually constructed with multiple polygons it does not seem reasonable to use border around them, but for the rest of the inland water it sounds OK for me. Of course any other idea might work too.

jeisenbe:
> gradient border on inland waters first

As you mention, this isn't feasible for rivers since they are always made of multiple polygons. It would also show rendering artifacts with lakes which are mapped as more than one polygon. That practice is still somewhat common, especially in areas with many interconnected lakes, so I do not believe it is a good idea. 

The plan is that there will be only half as much contrast between rivers and lakes as between oceans and rivers, so I don't see a problem here:

z14 after this PR
z14-ardalstangen-river-color-small
(Lake is upper right) 

z15 with 3 water colors (first tests)
z15-ardalstangen-altcolors
- I think the lake-river transition looks fine.

z15 lower contrast 3 colors
z15-ardalstangen-lowcontrast
- Lake-river transition is almost too subtle here

kocio-pl:
Try z19, please, and without bridge near the junction, this would be much more reliable test of this effect.

kocio-pl:
Good testing places (even without reaching z19):

- https://www.openstreetmap.de/karte.html?zoom=16&lat=12.52105&lon=104.38533&layers=B000TF
- https://www.openstreetmap.de/karte.html?zoom=17&lat=-31.63332&lon=-60.67576&layers=B000TF
- https://www.openstreetmap.de/karte.html?zoom=18&lat=52.61599&lon=13.20916&layers=B000TF

jeisenbe:
I've already done a couple tests on z17 (just in the higher-contrast 3 color style):

### z17 Afon rhythallt river-lake transition before
z17-afon-rhythallt-river-lake-before

z17 high-contrast 3-color
z17-afon-rhythallt-river-lake-after

### z17 Sure river-lake border before
z17-sure-border-before

z17 after (high-contrast 3-color)
z17-sure-border-rivercolor-after

jeisenbe:
https://www.openstreetmap.org/#map=19/53.04425/8.87248
Lower contrast 3 colors
z19-river-lake-low-contrast

Higher contrast 3 colors
z19-river-lake-higher-contrast

jeisenbe:
It's difficult to find places without a bridge, dam, weir or lock:

https://www.openstreetmap.org/#map=19/53.03678/8.87313
Lower contrast
z19-canal-lake-lower-contrast

Higher contrast
z19-canal-lake-higher-contrast

jeisenbe:
> https://www.openstreetmap.de/karte.html?zoom=16&lat=12.52105&lon=104.38533&layers=B000TF

Mapped badly: the river-lake transition is a few hundred meter out in the lake. We should not use poor data for testing

> https://www.openstreetmap.de/karte.html?zoom=17&lat=-31.63332&lon=-60.67576&layers=B000TF

Incomplete: the wide part of lake is tagged like a river, but the narrow part is tagged like natural=water only, so it looks backwards currently (screenshot from openstreetmap.de):
laguna-setubal

> https://www.openstreetmap.de/karte.html?zoom=18&lat=52.61599&lon=13.20916&layers=B000TF
This one is mapped fine. I don't have this data downloaded, but here's the screenshot from the link:
havel-german

kocio-pl:
I have general question regarding water data in general, since you were making these comments in different cases lately - how do you know it's really badly mapped?

And the second one - if we use different colors for water, aren't you afraid people will start moving inner water borders so they look nicer (real mapping for rendering), especially to make the border small and unintrusive?

This does not mean I would be against lower contrast with gradient, I just like to be sure we know what we're choosing and what are the downsides of the choice.

jeisenbe:
> how do you know it's really badly mapped?

Re: lake-river transition: a river has flowing water, a lake doesn't. The boundary between this two should not be out in the middle of the lake. 

This is how the riverbank in first example is currently mapped: 

https://www.openstreetmap.org/relation/1348576#map=14/12.5875/104.4208
delta-riverbank

The line goes around a peninsula and along the edge of the marshes. I suspect this was mapped coarsely originally, and has now been improved somewhat but still needs more work. This way still goes around the entire 100km long river: https://www.openstreetmap.org/way/378290382#map=12/12.5671/104.5050

> aren't you afraid people will start moving inner water borders so they look nicer?

In some cases they will need to be moved because the current data is not accurate, and often that will improve the rendering. But most of the current data is quite good. It's the less accurate examples that stand out.

kocio-pl:
You're right, I don't withdraw my agreement in any way. I just wanted to ask about your take about some examples illustrating real life problems, which can be interesting to extrapolate on many others similar cases, like this:

> >    https://www.openstreetmap.de/karte.html?zoom=16&lat=12.52105&lon=104.38533&layers=B000TF

> Mapped badly: the river-lake transition is a few hundred meter out in the lake. We should not use poor data for testing

When talking about moving/standing water border it seems that this is more correct in most of the cases, since I suspect water still moves there:

- Do you agree this particular example can be a tagging problem?
- What speed limit would you take as this kind of a border? 
- How would you expect the people should measure it?
- Are you not afraid that people will start "aligning" the borders so they make the smallest possible visual verge, without even taking water speed into account?
- Any other possible problems that come to your mind?

imagico:
Your questions are mostly related to tagging and mapping, not to rendering so i am reluctant to start a discussion about these here.  The idea of differentiated water rendering is based on what is currently tagged and mapped - which as i mentioned before i think is done quite consistently.  Same as for natural=wood and natural=scrub or landuse=industrial and landuse=commercial.

> Are you not afraid that people will start "aligning" the borders so they make the smallest possible visual verge, without even taking water speed into account?

Apart from as indicated this not being about water speed but about the ecological and physical characteristics of the waterbody - No.  Mappers as explained do differentiate the different waterbody classes and globally seen they do it quite well (in case of the ocean/coastline i would even say extremely well).  There is IMO no reason to assume they'd stop doing that because OSM-Carto starts acknowledging their work in rendering.  It is not that the colors suggested are so different that mappers are likely to have strong preferences and for example think: _This color is not suited for a lake so i am going to re-tag it as waterway=riverbank_.  If a mapper tags a lake `natural=water`+`water=lake` and a river flowing into it `natural=water`+`water=river` they would not be surprised if that difference in tagging is visible in rendering in the form of differently colored polygon fills and would likely not be inclined to change the tagging or the mapped geometries.

For mappers who don't care and just tag any waterbody with some tag to make it blue the best we can hope for is that the differentiated rendering encourages them to inform themselves on accurate tagging and encourages them to use it.

kocio-pl:
OK, thanks for responding anyway, I'm not too much into discussing it, rather asking about your position on (inevitable, as always in the real world) problems. My impression is that you don't worry, whatever mappers will do with it, you trust them when it comes to position of water borders. It is essentially on a par with my general laissez-faire attitude, so I don't complain, I'm just surprised that you don't see any potential data problems as usual. Nothing to discuss, though. I just confirmed that you're comfortable with all the consequences and the given example does not bother you.

I'd like to ask something else (if you're interested in explaining a bit more) - I'm just curious, they are "consistent" with what and "excellent" measured how exactly? I mean, how do you personally evaluate the quality of water borders? What benchmarks do you use, not being able to see the water yourself in most of the cases? Are satellite images of special use here, for example, when the definition you gave is so complex?

@jeisenbe, I'd like to hear your position on the data errors and possibility that people might start relocating borders just to make them small and what is your reference to say if data are bad (sorry for repeating the question, I'm just still interested in hearing that)?

imagico:
Consistent means consistent use of tags by mappers as it has been discussed in a lot of cases in case of feature additions in this style.  If different mappers use the same tag in the same way according to the same understanding what it means and the same view of the geographic reality that is consistent use.

The ocean/coastline mapping being quite excellent in its consistency is owed to the fact that there are a lot of people performing routine quality checks on coastline mapping and because the coastline data is one of the few feature types in OSM that is routinely processed on a global level.  You can see that by comparing it to other data sources.  That there are a few cases world wide with really blatant inconsistencies does not diminish that significantly.

My view of the likely effects of differentiated rendering of different waterbody classes is not a laissez-faire attitude, it is based on what can be expected from the experiences we made with various other cases of providing visual feedback to mappers.  And the idea of providing positive feedback on mapping consistently performed by mappers already to a large extent is part of the primary goals of this style.  Should this expectation be disappointed and the data quality degrade as a result of our rendering decision we could and should of course revisit that decision.

jeisenbe:
> https://www.openstreetmap.de/karte.html?zoom=16&lat=12.52105&lon=104.38533&layers=B000TF
> What speed limit would you take as this kind of a border?
> How would you expect the people should measure it?

In the example above, the water is actually flowing out of the lake into the river "delta", unusually. I suspect the water flow is not apparent until you are in the narrower, linear river, though I have not visited the location in person. As with the transition between river and ocean, one would expect the linear flow of water in the river area to be dominant, not reversing with winds or tides, unlike currents found in a lake.  The actual velocity or speed of the water will vary depending on the size of the river and how quickly it drops in elevation. 

> Are you not afraid that people will start "aligning" the borders so they make the smallest possible visual verge, without even taking water speed into account?

I expect that good visual results will be consistent with a reasonable border between lake and river in all common cases. If I were going to use OpenStreetMap data to find the total surface area of all reservoirs in California, for example, I would expect to find the maximum area, represented by the high water line. So the river should end as soon as it touches the high water line of the reservoir, not extending into the lake: this also produces good visual results, as shown in the second example in this comment https://github.com/gravitystorm/openstreetmap-carto/pull/3930#issuecomment-540585639. At the dam end of a reservoir the river should start at the dam. For a natural lake I would expect the same method of mapping, based on the season high water line of the lake. Generally this is what is currently done with the great majority of lakes that I have checked in various places.

jeisenbe:
> I'd like to hear your position on the data errors and possibility that people might start relocating borders just to make them small and what is your reference to say if data are bad

Mainly "bad" data is just "coarse" or "roughly mapped": in developing countries like here in Indonesia, many mappers think "it's good enough if it looks ok at z10" at first, but over time the mapping is slowly improved. Rendering features in this style helps to encourage this process of improvement: a mapper sees that the forest is mapped as a jagged line that overlaps with areas of farmland, so they adjust it a bit. 

By showing this subtle difference between rivers and lakes and seas, mappers will start noticing situations like in your comment here: https://github.com/gravitystorm/openstreetmap-carto/issues/3895#issuecomment-537488876 - I suspect that the local mappers near Kalingrad and Archangelsk will discuss where the coastline should be in these cases, which will be healthy and may lead to improvements in overall tagging of some other estuaries and river mouths. 

Right now with no visual feedback there is nothing to show when something like this happens: https://www.openstreetmap.org/changeset/73057949 - (This made me quite sad, since it erased hours of work fixing the coastline around these tidal channels and estuaries. The coastline shouldn't be below the mangroves: they grow in sea water. It took me 2 months to notice, since the rendering was unchanged. )

jeisenbe:
BTW, here's what that river/lake(reservoir) transition might look like at z19 in Luxembourg after all 3 colors are used:

z19 lower contrast
z19-sure-lake-lowcontrast9 

z19 higher contrast
z19-sure-lake-higher-contrast

Notice that even in a well-mapped country like Luxembourg, the landcover areas are quite jagged at z19. This is what the aerial imagery looks like: https://www.openstreetmap.org/edit#map=19/49.86787/5.81849

z19-mapbox-sure

I suspect that most mappers and map users very rarely zoom in to z19 in areas like this.

jeisenbe:
Both @kocio-pl and @imagico have commented on this PR in generally positive ways, but it's not yet approved. Any chance of a formal review this week?

jeisenbe:
I've made the suggested changes: I created a new png with the same pattern as the current intermittent_water but using the new river-color.

Test in Montpellier, France
https://www.openstreetmap.org/#map=17/43.61361/3.89126
z17-le-verdanson-river-color

kocio-pl:
@jeisenbe I'm puzzled, why have you merged this code? I thought I made it crystal clear that it's not ready quite early with https://github.com/gravitystorm/openstreetmap-carto/pull/3930#issuecomment-540557790:

> The fundamental problem I have with this solution is still the same - the inner water borders are sharp.

and they still are.

jeisenbe:
It would be helpful in the future to formally request specific changes if there are objections to a PR.

The linked comment was made 15 days ago. I responded to it https://github.com/gravitystorm/openstreetmap-carto/pull/3930#issuecomment-540560016 - 

"I'm happy to fix this for the ocean/river border with #3926, but that's only possible after we start using the darker river color. Recall that in the past we only had the outline for land polygons from z0 to z7 because at mid and high zoom levels there was a strong line across the river/coastline transit. _To avoid that, we have to change the river color first._" (emphasis added)

It's not possible to make the coastline linear gradient unless we first change the river-color. Since it is recommended to make small PRs, I could not include that change in this PR, since major layering changes are required.

4 days ago, I commented https://github.com/gravitystorm/openstreetmap-carto/pull/3930#issuecomment-544772443 "Both @kocio-pl and @imagico have commented on this PR in generally positive ways, but it's not yet approved. Any chance of a formal review this week?" - which described my understanding of the situation. There was no response requesting any changes except for imagico's request to change the intermittent water pattern, so I thought that meant there was consensus.

kocio-pl:
Sorry, I try to make formal requests most of the time, but I don't want to make more tension when discussing then necessary. And I thought in this case you know what I accept and what I don't, which appears was not true.

I don't understand where did you find me commenting this in a positive way? In my opinion this was a critical review all the time. The only positive thing I remember was in response to "since you agreed that in principle the differentiation of different waterbody classes" - and yes, I still support this **principle** (you're right, we got the consensus here), but not this **implementation**, you have probably mixed these things. My approval is conditional and relies on the soft border rendering. This condition is roughly OK for ocean and seas, so I approved that, but for inland waters the border is still sharp, so I'm still against.

OK, back to the current point. My concerns about people moving lines to just look smaller is further amplified by your responses - Christoph gave no clear rules and only rejected one (speed of the water), and you mentioned only high water level, which is about water/land, not how deep river goes into the lake for example and what shape does it make, so you see no easy tools for people willing to verify borders, which is bad. So it becomes more important that people don't see artificial strong lines if they decide the line should be long. They should be not pushed by visual solution which is suitable only for small borders (and even then it's inferior).

I propose to revert this code and make a new release (4.24.1 or 4.25.0), which could be deployed any time, so we could get back to the work with inland waters implementation.

The possible solutions I see are:
- making preprocessed data for joinings between the inland waters
- making all the inland waters the same darker color, so we could deploy #3895 soon

The first one is technically more complex, but it allows making soft borders and is vector friendly (it does not rely on any special, innovative rendering features) and is fallback friendly too (if it breaks, we simply get the hard borders).

The second one needs some more testing around the world and does not have to stay forever, once we have the proper solution.

jeisenbe:
Thank you for suggesting specific changes, this is helpful

>making preprocessed data for joinings between the inland waters

In the past the use of preprocessed data has been rejected by @pnorman and others, because it increases the complexity of maintaining, using and adapting this style. (However, this is only really necessary to show an outline around river areas, not for the ocean or lakes.)

> making all the inland waters the same darker color, so we could deploy #3895 soon

This is certainly possible as an intermediate step, but I don't understand which problem it will solve?

>  I still support this principle (you're right, we got the consensus here), but not this implementation

The implementation is simple: water=river, water=canal and waterway=riverbank areas as well as waterway=* lines are changed to `#8fcadd`. No objection has been presented to this change, either to the color chosen or the features chosen. 

The objection appears to be to releasing v4.24.0 before other PRs were made and accepted?

I has been said that it is better if a PR changes one thing at a time, rather than changing many features all at once - I have made this mistake before, and am trying to improve.

It would be hard to continue development if a large number of changes have to be accepted all in the span of one release. Many of the maintainers do not have time to review multiple PRs. 

In the past we have usually made incremental changes over several releases, even when the individual changes do not make an improvement.

For example, the ST_PointOnSurface PRs have resulted in worse label placement in some cases, but they will eventually allow vector tiles so we have been willing to accept this.

kocio-pl:
> This is certainly possible as an intermediate step, but I don't understand which problem it will solve?

It will allow us to show ocean/inland border the way we have agreed, exactly.

> The implementation is simple: (...) No objection has been presented to this change, either to the color chosen or the features chosen.

Of course, the objection has been raised exactly about the fact that it's _too_ simple, not to any part of this change. It's just lacking an important part. Please, let's focus on it.

> The objection appears to be to releasing v4.24.0 before other PRs were made and accepted?

My objection is that this change is not agreed upon, it has nothing to do with other PRs. 

> I has been said that it is better if a PR changes one thing at a time, rather than changing many features all at once - I have made this mistake before, and am trying to improve.

I appreciate that and I'm happy with how do you handle such things now - thank you, Joseph. Oceans PR just depends on this code and that's fine, it's better than one big chunk.

> For example, the ST_PointOnSurface PRs have resulted in worse label placement in some cases, but they will eventually allow vector tiles so we have been willing to accept this.

But we have fully agreed on that and such decisions are made on a case by case basis. not as a general rule. My reason was that vector tiles is a big, long and complicated goal, which has the potential to resolve many of our disputes (what color should it be? how many icons we can show? from what zoom level?) and fix some serious problems (rendering names in different languages). It was worth to make such compromise, even if I don't like the current rendering of labels. But if they would go outside the boundaries, like it was before, I would be against instant changes and would concentrate on fixing rendering first.

With this change it's much simpler and I don't fear it will take years to resolve, like the vector case, so I think it is not worth pushing such raw changes on a fast path. Why do you think we need to hurry up with this?

jeisenbe:
> > This is certainly possible as an intermediate step, but I don't understand which problem it will solve?
> 
> It will allow us to show ocean/inland border the way we have agreed, exactly.

I don't understand. If you mean that we should add the gradient line around the coastline, it's not necessary to change the color of lakes to do that.

(It might actually be possible to do a gradient line for lakes right now, though I'm concerned it may cause problems at some rare places where 2 lakes are connected without a river in between)

> > The implementation is simple: (...) No objection has been presented to this change, either to the color chosen or the features chosen.
>
> Of course, the objection has been raised exactly about the fact that it's _too_ simple, not to any part of this change. It's just lacking an important part. Please, let's focus on it.
> Oceans PR just depends on this code and that's fine, it's better than one big chunk.

I don't understand this part. Do you mean the PR to render oceans in a slightly lighter color, which I've previously suggested? Or the relayering in #3854? Or the coastline gradient?

I would like to know what you are requesting as a next step.

kocio-pl:
> (It might actually be possible to do a gradient line for lakes right now, though I'm concerned it may cause problems at some rare places where 2 lakes are connected without a river in between)

I'm also very concerned about this. That's the only reason why I don't want to merge ocean PR right now.

> I don't understand this part. Do you mean the PR to render oceans in a slightly lighter color, which I've previously suggested? Or the relayering in #3854? Or the coastline gradient?

I mean missing a gradient on the water borders. In a case of this PR, only internal waters. 

> I would like to know what you are requesting as a next step.

I have proposed how to resolve the problem of sharp water borders - if you don't like to involve preprocessing, then making all the internal waters dark is what can be done, so I would go with that.

jeisenbe:
So you are saying that you are more concerned about where lakes meet rivers, rather than at the coastline? I did not get that from the previous discussion.

So far there has only been 2 examples shown where the transition between a lake and a river would look unusually: that was the large lake in Cambodia where the river area had been mapped coarsely with a single, large outline originally, and then later was made into a multipolygon, but the transition between the lake and river wasn't fixed when the rest of the area of updated (probably with better aerial imagery). There was also an areas where the lake and river were not tagged properly yet - see https://github.com/gravitystorm/openstreetmap-carto/pull/3930#issuecomment-540602131 - 

There are going to be some examples of missing tags which are shown by this change, as with any time that we render a new feature. 

I do not believe there is a valid issue with showing a difference between the colors of lakes and rivers, and more than in showing a difference between heath and scrub areas. 

This style should represent accurately the geometries in the database.

Adding a gradient around the coastline is something that could help distinguish the location of the coastline, but it's not necessary around lakes, and even for the coastline it's a "nice to have" feature.

jeisenbe:
Do you think this PR should be reverted?

imagico:
My view here is that I think @jeisenbe did nothing wrong in his principal approach and that both me and @jeisenbe have to some extent misunderstood the comments by @kocio-pl - which was probably at least partly because they were somewhat unclear and contradicting (at least from my point of view)

@kocio-pl - the way i read your comments now it seems you kind of say you support differentiating different classes of waterbodies but you don't want this difference to actually be explicitly visible.  This seems a contradiction to me.  What i think would make the situation much clearer is if you either

* state clearly that you currently see no way to differentiate different classes of waterbodies in a way you find suitable for this style or
* make a specific suggestion how to differentiate different classes of waterbodies.

Either of these would give a positive perspective on how to move forward.

As far as reverting the change is concerned - i am probably not the right person to decide this.  But i am kind of with @jeisenbe here that this change is the logical first step for any design strategy towards differentiating different classes of waterbodies - you cannot differentiate without differentiating after all.  In other words:  Reverting this change would not actually solve the problem of us needing to come to an agreement on the actual design decision.

kocio-pl:
> Do you think this PR should be reverted?

Yes, nothing new from https://github.com/gravitystorm/openstreetmap-carto/pull/3930#issuecomment-546550403. I see no reason for the rest of the code to not be deployed soon, while we keep discussing inland waters.

> This style should represent accurately the geometries in the database.

I think you're oversimplifying the problem. Real geometry of the database is sometimes hidden - deliberately, I suppose. 

For example we hide waterways under water areas (the intermittent case is an exception here). Also when I suggested outline for the lakes, we would see the actual geometry of all the parts used there, but this was exactly why you rejected this idea.

> you support differentiating different classes of waterbodies but you don't want this difference to actually be explicitly visible

I think you're getting me right, but where do you see the contradiction in these claims? It's similar to "I like this color, but not so strong (explicit)".

>    state clearly that you currently see no way to differentiate different classes of waterbodies in a way you find suitable for this style or

I don't understand what do you miss from my propositions. In my opinion the code from this PR does the good job with showing different classes of waterbodies, but it also does the bad job of suggesting the solid difference, which is just an artifact. What I want is to keep the good things while not introducing bad things.

> make a specific suggestion how to differentiate different classes of waterbodies.

Making all the inland water darker fulfills this requirement and this is what I proposed lately as an alternative take, which also fulfills my requirements.

> Adding a gradient around the coastline is something that could help distinguish the location of the coastline, but it's not necessary around lakes, and even for the coastline it's a "nice to have" feature.

For me adding a gradient for the coastline is a minimal solution that made me accept this code, as long as there is no hard water border, which is my primary concern. The fact that it tunes the coastline rendering back to the previous state is a nice to have feature for me.

***

I hoped you all understand that and I wont have to repeat what is the fundamental problem in every comment, but it looks like I have to repeat it. 

So please make a very simple test - just check if any given solution shows the soft water borders or not, and you instantly know whether this moves the proposition forward (or not).

jeisenbe:
I've opened #3955 which would revert this PR.

jeisenbe:
> > This style should represent accurately the geometries in the database.
> 
> I think you're oversimplifying the problem. Real geometry of the database is sometimes hidden - deliberately, I suppose.
> 
> For example we hide waterways under water areas (the intermittent case is an exception here).

We can not show everything in the database. When there are 2 features in the same place, we can only shown one of them at a time. With waterway line in water areas, the mapper feedback would be best if we put the line on top, but this would violate map users expectations, and would encourage mappers to incorrectly delete the water lines after mapping water areas to get a "proper" rendering. 

>  Also when I suggested outline for the lakes, we would see the actual geometry of all the parts used there, but this was exactly why you rejected this idea.

I think you are misunderstanding the history. I mentioned the outline to try to compromise and meet your request to not have a "sharp" meeting between river-color and lake areas, but I mentioned that sometimes it might show a line between two adjacent lakes, and you rejected this:

 "@jeisebe: It might actually be possible to do a gradient line for lakes right now, though I'm concerned it may cause problems at some rare places where 2 lakes are connected without a river in between."
@kocio-pl "I'm also very concerned about this. That's the only reason why I don't want to merge ocean PR right now."

It would be ok to show a line between lakes in theory: since we render a text label for each lake mappers already have an incentive to map lakes as one polygon or closed way. I don't think they would remap them due to our adding an outline. But I also did not mention it before because I do not think it is necessary to show an outline for lakes.
 
We should not show a line between different river polygons, because rivers are properly mapped as a large number of separate areas. A single huge polygon, extending for 1000's of kilometers in the case of the Danube or Mekong, would be hard to maintain and hard to use, so mappers rightly have chosen to map rivers areas as many smaller segments (just as the waterway is broken up in 10 to 20 kilometer pieces). The proper expectation of mappers is that this will be rendered as one continuous feature.

> > you support differentiating different classes of waterbodies but you don't want this difference to actually be explicitly visible 
> I think you're getting me right, but where do you see the contradiction in these claims? It's similar to "I like this color, but not so strong (explicit)".

I find the frequent request to use visually weaker colors is a frequent problem in reviews here. See the issue #3647 for examples of many times when contrast has been reduced, sometimes so that two different colors cannot be clearly distinguished. 

Features should be shown clearly. It should not be required to have "fuzzy" borders between two areas. 

If we accept this for water bodies, doesn't that mean that we should also make fuzzy boundaries between all natural vegetation areas, like grass, heath, scrub, woodland? This actually look "fuzzy" on aerial imagery. We don't do this, because we are representing information in a vector database, which stores everything as lines.

Similarly, it would be more realistic to represent rivers and streams as curving lines, like on opentopomap and other styles. But we do not do this, because mappers should be able to see the actual geometry as clearly as possible. 

> > make a specific suggestion how to differentiate different classes of waterbodies.
> 
> Making all the inland water darker fulfills this requirement and this is what I proposed lately as an alternative take, which also fulfills my requirements.

That would be rejecting @river-color, and only changing the ocean-color. 

While that is better than nothing, it would not help improve mapping of rivers, and it would make the lakes darker than necessary. 

This is based on insisting that we cannot have a release which shows a pixel of water-color next to a pixel of ocean-color or river-color. 

I don't believe this is a valid argument. From what I can see, there has not been an argument presented to explain why it would cause problems for map users or mappers (the openstreetmap.de style is already using a similar rendering without problems). I think this objection is probably based on a subjective, cultural-determined or perhaps individual view of how things should look.

Please take a look at #3951 - consensus-based decision making will not work if our decisions are based on tastes or preferences. It will also not work if maintainers veto changes which are not exactly what they prefer.

imagico:
>>    you support differentiating different classes of waterbodies but you don't want this difference to actually be explicitly visible

> I think you're getting me right, but where do you see the contradiction in these claims? It's similar to "I like this color, but not so strong (explicit)".

As i said elsewhere i think attempts to create a compromise between showing something and not showing something will always result in bad map design.

Since you - even on specific inquiry - do not make any suggestion how you think different classes of waterbodies can be differentiated without the difference being visible (think of the example i gave for comparison:  You want to differentiate natural=wood and natural=scrub but don't want the difference to be visible) i can at this point interpret your answer only that you don't see a way to differentiate in a way that you would find acceptable.  Postulating the existence of some hypothetical possibility to differentiate without actually showing the difference does not change that - what you postulate seems physically impossible, at least doing it in a way that is compatible to our documented goals of creating a well readable map and providing constructive mapper feedback.  If you disagree please show us how you think this can be done.

And as @jeisenbe said - so far your rejection of a visible differentiation seems based exclusively on a subjective dislike for it - you have so far not presented any arguments and reasoning why you think this would be problematic w.r.t. any of the goals of this style.

kocio-pl:
Well, I think we could simplify a problem then for a start: please provide an objective proof for 3-color water solution - not 2 and not any other number of classes. It seems to be missing from and it looks to me like a very subjective choice.

imagico:
You seem to be taking a very destructive approach to this now, we have discussed the reasoning for this change at length and you (at least seemed to) have agreed that differentiating different classes of waterbodies is a good idea.  @jeisenbe has - largely based on your request - split this into small step-by-step changes and now you are demanding proof for something (not sure for what but apparently not for anything specifically about this change).

If you want to question and argue with the reasons for differentiating different types of waterbodies or for this change here in isolation that is fine with me.  But i would ask you to read about the arguments that have been made for this in the past discussion and start from there instead of re-iterating the discussion yet again from the beginning.  @jeisenbe has showed and explained in detail the advantages of using different colors, we also have patiently answered a lot of questions raised by you so far.  I think it is a reasonable request for you now to present a constructive approach making specific suggestions how to proceed from here and how to address the various issues this change tries to address if you disagree with the approach taken.

kocio-pl:
I have proposed 3 different solutions and 2 of them preserves original 3-color proposition, I think this is quite a lot of constructive feedback. Rejecting them does not seem like a constructive step. Consensus means all the sides trying to reach it - so now it's your turn, I guess.

I agree with the general differntiation and this can be easily done with 2 colors (ocean and inland waters). I think any number of colors and any other choice by definition is subjective and I don't know why you treat objectivity as a new rule, but if you can prove me wrong and show me that 3 is not the subjective number, that would make finding a technical solution easier.

jeisenbe:
>3 different solutions and 2 of them preserves original 3-color proposition

Would it be possible to explain the steps that would be part of each of those solutions? I'm a little confused about the next step to take, and how many steps need to be included in 1 PR.

(Please, let's discuss at #3895 or #3896 - not in this closed PR where the comments will get hidden)

kocio-pl:
Sure, I will do it soon there.
EOF

PR #3931, opened 2019-10-14
Remove rendering of waterway=wadi

### LINKED ISSUE #1365: drop rendering for waterway=wadi
As discussed on tagging mailing list[a] and described on wiki[b] this tag is ambiguous. It may mean ephemeral river, valley, typical course of flash floods. What worse word wadi is used in multiple languages with multiple meanings. There are also additional problems.

It is recommended to use waterway=river/waterway=stream + intermittent=yes for waterways, natural=valley for valleys.

I am proposing to drop rendering of this tag.

[a] https://lists.openstreetmap.org/pipermail/tagging/2015-January/thread.html - thread "waterway=wadi problem"
[b] http://wiki.openstreetmap.org/wiki/Tag:waterway%3Dwadi

jeisenbe:
Fixes #1365

- Removes rendering of deprecated feature waterway=wadi, which should instead be mapped as `intermittent=yes` + `waterway=stream` or `waterway=river` if it is a waterway, or as `natural=valley` if it is a dry valley. Removing the rendering will support mappers in consistent use of these tags
- Does not remove wadi from the list of linestring exceptions in `openstreetmap-carto.lua` because these features should be imported as linear ways, in case this style is adapted for other purposes. 

### Trend of `waterway=wadi` usage over time:

waterway-wadi-taghistory

In comparison, intermittent=yes (which is used with waterway=* over 92% of the time) has increased to over 2 million uses, with over 1.9 million waterway=stream + intermittent=yes 

Also, as mentioned in the related issue (https://github.com/gravitystorm/openstreetmap-carto/issues/1365#issuecomment-464341386), "three groups from individual mappers in Tunisia, Pakistan and southern Turkey from about 4 years ago account for nearly half of the data].

### DIFF
--- a/project.mml
+++ b/project.mml
@@ -214,7 +214,7 @@ Layer:
             CASE WHEN tunnel IN ('yes', 'culvert') THEN 'yes' ELSE 'no' END AS int_tunnel,
             'no' AS bridge
           FROM planet_osm_line
-          WHERE waterway IN ('river', 'canal', 'stream', 'drain', 'ditch', 'wadi')
+          WHERE waterway IN ('river', 'canal', 'stream', 'drain', 'ditch')
             AND (bridge IS NULL OR bridge NOT IN ('yes', 'aqueduct'))
           ORDER BY COALESCE(layer,0)
         ) AS water_lines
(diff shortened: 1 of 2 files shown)


jeisenbe:
There are only 6 waterway=wadi features in Lebanon (http://overpass-turbo.eu/s/N6L), and most are shown in this image:

https://www.openstreetmap.org/#map=14/33.9475/35.7315

z14 Deir Chamra, Lebanon - before

After
z14-deir-chamra-wadi-after

In contrast, there are 112 rivers and streams with `intermittent=yes` in Lebanon: http://overpass-turbo.eu/s/N6M

sommerluk:
Note: In the image you have posted, the villages use Latin script while on openstreetmap.org they use Arabic script…

jeisenbe:
I used a year-old lebanon.osm.pbf extract

Issue #3201, opened 2018-04-27
Migration to vector tiles


kocio-pl:
Follow up to https://github.com/gravitystorm/openstreetmap-carto/issues/1975#issuecomment-384858383

As discussed privately among team members, it's good to migrate osm-carto to be rendered with [vector tiles](https://wiki.openstreetmap.org/wiki/Vector_tiles). To allow as smooth and painless migration path as possible, it should be ready for server side raster tiles rendering. It should be non-disruptive change for clients while allowing to start developing client side rendering and testing vector tiles in general.

This is not a fresh and untested approach, @gravitystorm has been using it for his own styles for a few years already - you can watch his presentation about it from SotM EU 2014:

https://www.youtube.com/watch?v=NoLJHgqDCzc

There's also a vector fork of osm-carto by @rory (also his presentation from SotM 2016):

https://github.com/geofabrik/openstreetmap-carto-vector-tiles
https://2016.stateofthemap.org/2016/converting-a-regular-carto-project-to-vector-tiles-osmcarto-case-study/

However this fork is quite old and happened before our v4.x database migration. It's not clear now how useful it could be, other than being proof of concept.

This ticket is about coordination of the effort, more specific tickets might be needed for subtasks and technical background (like choosing the proper software stack) is yet to be discussed and tested.

jbelien:
Would be really awesome to support Vector Tiles ! 🎉 
But imho I do not think we should migrate ... We should add Vector Tiles support while still providing Raster Tiles :)

StyXman:
I would agree, but maintaining two different styles is too complicated, specially because they don't seem to be compatible, or at least nobody has written a converter from one to the other. Think about all the contributions, having to be implemented twice...

kocio-pl:
This problem is much wider than style, it's also related to deployment strategy (especially on OSMF servers, but most probably also for the forks following us, like German or Belgian) and tools to be used.

@jbelien The whole point with server side raster tiles rendering is about planning deployment so that vector osm-carto tiles are then transformed into raster tiles. In other words instead of:

osm-carto -> raster tiles

it would be:

osm-carto -> vector tiles -> raster tiles

so the final output should be very similar. 

Please note however that from the osm-carto perspective it's just a migration. Raster tiles rendering will be done with some other tools, so it's a deployment problem. Yet my intention is to plan the whole process with our partners (including you :smile:), not just abrupt change of the styling output.

[...]

PR #3932, opened 2019-10-15
Use ST_PointOnSurface for roads-area-text-name

jeisenbe:
Related to #3201

- Use ST_PointOnSurface for roads-area-text-name in project.mll
- Remove no longer needed `text-placement: interior` from roads.mss

This is necessary to be able to use this style with server-side vector tiles.

## Test rendering

https://www.openstreetmap.org/#map=18/47.13734/9.52334
before:
z18-peter-kaiser-platz-before

after:
z18-peter-kaiser-platz-st-pointonsurface

### DIFF
--- a/project.mml
+++ b/project.mml
@@ -1796,13 +1796,13 @@ Layer:
             name
           FROM planet_osm_polygon
           WHERE way && !bbox!
-            AND highway IN ('residential', 'unclassified', 'pedestrian', 'service', 'footway', 'cycleway', 'living_street', 'track', 'path', 'platform')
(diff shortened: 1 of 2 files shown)


jeisenbe:
I've pushed a new commit which adds parenthesis around all of the OR statements, so that the ```WHERE way && !bbox!``` statement will apply properly to all of the features.

jeisenbe:
@pnorman are you interested in reviewing this?

jeisenbe:
Anyone available to review?

PR #3938, opened 2019-10-19
Do not render stream, ditch and drain tunnels at z14

### LINKED ISSUE #3676: Inconsistent tunnel=yes rendering for streams
There is inconsistency for rendering tag+key values tunnel=yes on waterway=stream.

On zoom level 13, [the stream is not rendered](https://www.openstreetmap.org/#map=13/49.3835/16.7418).
On zoom level 14, [the stream is rendered as regular waterway](https://www.openstreetmap.org/#map=14/49.3840/16.7393).
On zoom level 15, [the stream is rendered with dashed line symbolizing tunnel](https://www.openstreetmap.org/#map=15/49.3838/16.7378).

My concern here is about zoom level 14: it should either not be rendered (the same as zoom 0-13) or rendered by a dashed line (the same as zoom level 15-19). Now it appears as "normal" waterway just for one zoom level, which is strange.

(just to note: the whole area I am linking is a karst zone with underground creeks/rivers, so it is tagged correctly).

jeisenbe:
Fixes #3676

- Do not render stream, ditch and drain tunnels at z14

### Explanation:
- Currently stream, ditch and drain waterway tunnels are rendered the same as above-ground waterways at z14
- At z15 and above they have a special tunnel rendering. 
- This commit removes the rendering at z14, since at this level intermittent streams are no longer rendered. It would not make sense to render stream culverts at z14 but not intermittent streams.

### Test rendering:

https:/www.openstreetmap.org/#map=15/36.9649/-121.9659
Before
z15-rodeo-creek-same

z14-rodeo-creek-before

After
z14-rodeo-creek-after-no-tunnel

**https:/www.openstreetmap.org/#map=15/37.1896/-122.2085**
Before
z14-sempervirens-stream-tunnel-before

z15-sempervirens-stream-tunnel-before

After
z14-sempervirens-stream-tunnel-after

### DIFF
--- a/water.mss
+++ b/water.mss
@@ -194,7 +194,7 @@
   [waterway = 'stream'],
   [waterway = 'ditch'],
   [waterway = 'drain'] {
-    [int_intermittent != 'yes'][zoom >= 14],
+    [int_intermittent != 'yes'][int_tunnel != 'yes'][zoom >= 14],
     [zoom >= 15] {
       // the additional line of land color is used to provide a background for dashed casings
       [int_tunnel = 'yes'] {


kocio-pl:
Before and after images in the wild looks the same to me, probably the line is too short here for the changes to be seen. 

Otherwise it makes sense to me.

jeisenbe:
I've fixed the images for the second example, they were the same previously. The difference is subtle for both examples, because even a 100 meter long tunnel is quite short at this zoom level.

imagico:
I have my doubts about this.  Removing selectively rendering of waterway tunnels removes display continuity of waterways along their route - they'd disappear and re-appear potentially at unconnected points.

jeisenbe:
They are not visible on the ground, so it properly represents the
situation. In most cases the gaps are very short and will not be
visible at z14: most drain, ditch and stream tunnels are short
culverts under roads, which will be covered up by the road fill and
casing at z14. I had to hunt around for examples that would be visibly
different in the before and after.

I would also be happy to change this to show the tunnel style at z14
instead, except that intermittent streams are not being rendering at
z14. It would be strange to not show intermittent streams at this
level, since they can actually be seen, but to show tunnels which are
not visible.

But if we reconsider removing intermittent streams from z14 we could
also then show these tunnels with the tunnel rendering style.

imagico:
> I would also be happy to change this to show the tunnel style at z14
instead, except that intermittent streams are not being rendering at
z14. It would be strange to not show intermittent streams at this
level, since they can actually be seen, but to show tunnels which are
not visible.

I don't really see this connection.  The intermittent nature of a stream is completely independent of it running in a tunnel or not.

The main purpose of stream display at the lower zoom levels is showing the drainage network.  For this it makes sense to start by rendering streams uniformly and continuously and not interrupt their display if there is a tunnel.  I don't really mind much showing the tunnels distinctly but IMO rendering would be much cleaner if they are shown like normal waterways at the lower zoom levels.

The example shown in #3676 is not very good since it is  showing incorrect use of waterway=stream + tunnel=yes.

jeisenbe:
> The example shown in #3676 is not very good since it is showing incorrect use of waterway=stream + tunnel=yes.

There are waterway=drain and waterway=ditch features in long tunnels in some places, and they are not part of the natural drainage network. Should these be shown at z14?

Adamant36:
This waterway [here](https://www.openstreetmap.org/#map=18/40.57208/-122.35091) is probably miss-tagged due to my ignorance at the time, but imagine it wasn't. In my opinion there's zero reason to show it going through the retail area at z14 as it just adds clutter and get's in the way of the parking lot being rendered. Plus, it makes it seem like it's over ground when it's not. Is Anyone going to be looking for a long tunnel full of water (or not) like that one at z14 on this map? My guess is probably not. Nothing would be taken away from it not being displayed I don't think either. If it wasn't displayed it would be more cohesive visually because it would be clearer where it goes under ground and comes out again. 

z17
z17
z14
z14

jeisenbe:
@imagico, there are some waterway=drain and waterway=ditch features in long tunnels in some places, and they are not part of the natural drainage network. Should these be shown at z14?

jeisenbe:
@imagico please see question above: https://github.com/gravitystorm/openstreetmap-carto/pull/3938#issuecomment-544771087

imagico:
I would need to see specific cases of long ditch/drain tunnels that cannot be reasonably rendered as normal waterways at lower zoom levels without causing major confusion.  Keep in mind we have no way of distinguishing between long and short tunnels in rendering because waterways can be split arbitrarily.

PR #3942, opened 2019-10-21
Apply bbox to part of "addresses" query

### LINKED ISSUE #3937: Part of query is not restricted by bounding box
WHERE conditions on line:
https://github.com/gravitystorm/openstreetmap-carto/blob/05dc392c9b828814b3ce741e8b5ad397b934b4f5/project.mml#L2100

are different than on line:
https://github.com/gravitystorm/openstreetmap-carto/blob/05dc392c9b828814b3ce741e8b5ad397b934b4f5/project.mml#L2090

which causes sequential table scan (long execution) and collecting unwanted data

nexces:
Adding parenthesis around additional conditions allows to use index and prevents from selecting all rows from DB containing "addr:housename" or "addr:unit".

Fixes #3937 

- Adding parenthesis around additional conditions

No changes in rendering

### DIFF
--- a/project.mml
+++ b/project.mml
@@ -2097,7 +2097,7 @@ Layer:
             tags->'addr:unit' AS addr_unit,
             NULL AS way_pixels
           FROM planet_osm_point
-          WHERE way && !bbox! AND ("addr:housenumber" IS NOT NULL) OR ("addr:housename" IS NOT NULL) OR ((tags->'addr:unit') IS NOT NULL)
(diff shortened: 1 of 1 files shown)


nexces:
You can perform further checks with pure query grabbed from PostgreSQL slow query log. Run "EXPLAIN" on it and you will find how costs are distributed. Also original query returns points that are outside of bounding box which is irrelevant for rendering but has massive impact on data gathering. For Europe+Russia extract data collection went from ~45 seconds to 0.4 seconds

jeisenbe:
Thank you for that advice (I need to learn more about PostgreSQL), and for submitting for this PR.

It would be very helpful if we had an automated or semi-automated way to test performance, to prevent bugs like this in the future. There's an open issue #1287 - perhaps you have an idea about solving this?

PR #3945, opened 2019-10-23
Add 14px storage tank icon used by iD

### LINKED ISSUE #3927: Reuse iD editor storage tank icon

bhousel:
closes #3927
replaces #3928

### DIFF
--- a/amenity-points.mss
+++ b/amenity-points.mss
@@ -646,7 +646,7 @@
 
   [feature = 'man_made_storage_tank'][zoom >= 18],
   [feature = 'man_made_silo'][zoom >= 18] {
-    marker-file: url('symbols/man_made/storage_tank.svg');
+    marker-file: url('symbols/man_made/storage_tank2.svg');
     marker-fill: @man-made-icon;
     marker-clip: false;
   }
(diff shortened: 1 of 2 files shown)


jeisenbe:
Rendering examples:

z18-storage-tanks-buildings-compare

z19-silos-compare

kocio-pl:
Maybe making the gap a bit bigger would make the shape more obvious?

meased:
1. I disagree that the current icon looks like a gas pump.
2. The new icon is not pixel aligned.
3. The new icon is just a cylinder and could be mistaken for a lot more objects than the current icon (it reminds me of a hard drive).
4. Is an icon even needed for these features?
    - Whether it is "proper" or not, there is clearly a user expectation that these features will render like buildings. Why is that not enough?

jeisenbe:
Ok, now there's @jidanni and @adamant36 in favor but some specific critical comments by @kocio-pl and @meased about lack of pixel-alignment and the size of the gap between the top and the side. 

@bhousel - do you have time and interest in trying to address those concerns?

There have been a couple of comments about rendering this feature with building fill - this could be done in another PR, though it would only work for those mapped as areas (as with buildings).

bhousel:
> @bhousel - do you have time and interest in trying to address those concerns?

Not really, sorry.  I just sent the thing originally because I had the program open and could throw something together in a few minutes.

As a maintainer you should just pick something. It probably doesn't really matter which icon you use.  They'll all look pretty janky until this project is generating retina tiles, and you can't please everyone who shows up to your GitHub to complain.  Arguing over which storage tank icon is best is not a great use of anyone's time.

jeisenbe:
This PR would need to rebased now for v5.0.0, so I am going to close it as stale.

PR #3947, opened 2019-10-24
Remove name from SQL select when unused

jeisenbe:
- The column `name` was selected but not used for `water-areas`, `water-barriers-poly`, `water-barriers-line`, `bridge`,  `waterway-bridges`, `tourism-boundary`, and `protected-areas`. 
- The name labels for these features are rendered in different layers. 
- Therefore, `name` can be removed from the sql query for these layers

Test rendering is unchanged:

Before z16
z16-name-test-before

After
z16-names-test-after

### DIFF
--- a/project.mml
+++ b/project.mml
@@ -230,7 +230,6 @@ Layer:
             "natural",
             waterway,
             landuse,
-            name,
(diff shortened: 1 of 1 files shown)


jeisenbe:
Anyone available to review this minor code clean-up?

PR #3950, opened 2019-10-25
Remove way_pixels selection from bridge layer

jeisenbe:
- Remove way_pixels selection from bridge layer SQL query

The `bridge` layer for rendering man_made=bridge polygons does not use way_pixels, so this is not needed. Removing the line will slightly improve performance.

Test rendering is unchanged:

Before
way_pixels-bridge-before

After
way_pixels-bridge-after

### DIFF
--- a/project.mml
+++ b/project.mml
@@ -418,7 +418,6 @@ Layer:
       table: |-
         (SELECT
             way,
-            way_area/NULLIF(POW(!scale_denominator!*0.001*0.28,2),0) AS way_pixels,
             man_made,
             name
           FROM planet_osm_polygon


matkoniecz:
I am pretty sure that name of bridges is checking way_pixels. Is it impacted? Probably not, it is in https://github.com/gravitystorm/openstreetmap-carto/blob/4b24de74857ab93e278fcc624d5654cb538b4510/roads.mss#L2710 with its own query in https://github.com/gravitystorm/openstreetmap-carto/blob/e290f365ebc242282f1176394a85c57c64286963/project.mml#L1405

jeisenbe:
Right, `bridge-text` is it's own layer, separate from the `bridge`
layer that is involved in this PR

jeisenbe:
Anyone interested in reviewing this minor code cleanup?

Issue #3935, opened 2019-10-17
Roads are too prominent compared to landcover and rivers at mid and low zoom levels


jeisenbe:
## Expected behavior
- Highways should be visible but not overly prominent at mid and low zoom levels
- Landcover features should be balance the linear features
- Rivers and railways should be visually balanced with primary and trunk highways
- These visual features should encourage mappers to add waterways, water features, landcover and landuse features, not only roads

## Actual behavior
- Because the landcover layer is lightened by 0.6, roads are much more prominent than landcover at z10 and below. 
- At z11 and z12 the 0.4 and 0.5 lightening of landcover still makes roads more prominent, even the tertiary roads and unclassified/residential roads rendered as think gray lines
- Railways and rivers are too prominent compared to landcover at z12 and below, but are less prominent than motorways and trunk roads
- This suggests that this style is focused on roads, especially motorroads, at mid and low-mid zoom levels.

### Schwerin
https://www.openstreetmap.org/#map=10/53.70615/11.31001
![z10-lubeck-schwerin-current](https://user-images.githubusercontent.com/42757252/66988179-0d2dac80-f0fd-11e9-9afb-c57278a57fc3.png)

z11 Schwerin
![z11-schwerin-current](https://user-images.githubusercontent.com/42757252/66988600-ca200900-f0fd-11e9-928f-d84a6a236dd6.png)

### z10 Luxembourg
https://www.openstreetmap.org/#map=12/49.6490/6.1601
![z10-luxembourg-before](https://user-images.githubusercontent.com/42757252/66988213-1dde2280-f0fd-11e9-8e59-7216419d54ac.png)

z11 Luxembourg
![z11-luxembourg-before](https://user-images.githubusercontent.com/42757252/66988231-220a4000-f0fd-11e9-9f25-398c5869985d.png)

z12 Luxembourg
![z12-n-luxembourg-before](https://user-images.githubusercontent.com/42757252/66988241-2898b780-f0fd-11e9-9496-1da7517b3200.png)

### z8 Cardiff, Wales
![z8-cardiff-faded](https://user-images.githubusercontent.com/42757252/66988281-38b09700-f0fd-11e9-9299-900b1324e47d.png)

z9 Cardiff
![z9-cardiff-faded](https://user-images.githubusercontent.com/42757252/66988296-3fd7a500-f0fd-11e9-9513-b6c58661b623.png)

z10
[...]

Issue #3951, opened 2019-10-26
Consensus based decision making


jeisenbe:
## Define the method of making decisions based on consensus

Proposal: we should agree to a process for consensus-based decision making and document this in `CONTRIBUTING.md`, `CARTOGRAPHY.md` or another appropriate location. 

As @gravitystorm has stated, we should:

"Build consensus and agreement through discussion and compromise, using a combination of the cartography guidelines and common sense."

"Consensus doesn't mean unanimity, but there shouldn't be anything merged here against firm opposition from a maintainer. On the other hand, a maintainer shouldn't block a change without offering a very good reason, or another way forward. If there's opposite opinions on something that's not covered by the guidelines, then we can add more guidelines or update the existing ones."

We should discuss how this should work and how to resolve cases when two maintainers disagree. 

We should also discuss whether a review by any contributor should be given equal weight, which would mean that all objections to a proposed change would need to be treated equally. This would also increase the need to define when consensus has or has not been achieved.

jeisenbe:
## Background

If I understand the history correctly, back when this style started, @gravitystorm created the original design based on the old XML stylesheets used on openstreetmap.org previously, with the help of a few others. Several other people were made maintainers of the style, and there was an rule (perhaps unwritten?) that all changes should be approved by one of the maintainers prior to being merged. If there were objections, they needed to be addressed.

In [April 2017](https://github.com/gravitystorm/openstreetmap-carto/issues/2436#issuecomment-291008629) there a change which allowed maintainers to merge their own PRs even if there was a negative review requesting changes: 

> "Maintainers should take into account feedback they receive. In particular, negative reviews should be a reason to seriously reconsider the proposal. However, in the end each maintainer is allowed to merge pull requests independent on the amount of positive or negative reviews, giving maintainers ownership of their own changes. If there is a review that requests changes and the pull request is merged there should be some reasoning why this was not taken into account."

However, a review was still required before merging.

Then in  [July 2017](https://github.com/gravitystorm/openstreetmap-carto/issues/2436#issuecomment-318955397) there was a suggestion that maintainers should be allowed to merge their own changes even without a review (Perhaps after a 1 month wait https://github.com/gravitystorm/openstreetmap-carto/issues/2436#issuecomment-323137393)

[...]

PR #3952, opened 2019-10-26
Reduce landcover fading at mid-low zoom levels

jeisenbe:
### Related to #3935 and #3936
- Intermediate step before #3670

- Only fade from z11 and only 1 level - reduces the 3 level fading to only one level, using `lighten 0.2` instead of 0.4.
- Remove built-up-z11 color - this is no longer needed, since z11 is now the same as z10 and below
- Change `@built-up-z12` to `#dddddd` - this maintains the current rendered color at z12
- Change `@built-up-low-zoom` to `#d0d0d0` - same as current `@built-up-z12` color

### Explanation
As discussed in issue #3936 and the WIP PR#3670, the current fading of the landcover layers at low and mid-low zoom levels makes it difficult to distinguish different types of landcover/landuse at z12 and below.

Roads, railroads, and waterways are much more prominent than landcover at z12 and below, especially at z10 and below. Water areas are not faded, and therefore increase in prominence relative to landcover

This PR reduces the fading. While this transition still makes it more difficult to distinguish between light landcovers (like farmland), this situation is partial improved. This change will also make the transition smoother between the current fading and a future goal of no fading at low zoom levels, but few landcover classes, as shown in #3670 

## Test rendering :

Before
z10 Luxembourg
z10-luxembourg-before

z11
z11-luxembourg-before

z12
z12-n-luxembourg-before

After
z10
z10-luxembourg-reduce-fade

z11
z11-luxembourg-reduce-fade

z12
z12-luxembourg-reduce-fade

Also see more examples in https://github.com/gravitystorm/openstreetmap-carto/issues/3936#issuecomment-544090680

### DIFF
--- a/landcover.mss
+++ b/landcover.mss
@@ -11,9 +11,8 @@
 
 // --- "Base" landuses ---
 
-@built-up-lowzoom: #aaaaaa;
-@built-up-z11: #c0c0c0;
-@built-up-z12: #d0d0d0;
+@built-up-lowzoom: #d0d0d0;
+@built-up-z12: #d7d4d3;
 @residential: #e0dfdf;      // Lch(89,0,0)
 @residential-line: #b9b9b9; // Lch(75,0,0)
 @retail: #ffd6d1;           // Lch(89,16,30)
@@ -71,17 +70,11 @@
 
 #landcover-low-zoom[zoom < 10],
 #landcover[zoom >= 10] {
-  ::low-zoom[zoom < 10]                   { image-filters: scale-hsla(0,1,0,1,0.6,0.95,0,1); }
-  ::lower-mid-zoom[zoom >= 10][zoom < 11] { image-filters: scale-hsla(0,1,0,1,0.6,0.95,0,1); }
-  ::mid-zoom[zoom >= 11][zoom < 12]       { image-filters: scale-hsla(0,1,0,1,0.5,0.96,0,1); }
-  ::upper-mid-zoom[zoom >= 12][zoom < 13] { image-filters: scale-hsla(0,1,0,1,0.4,0.97,0,1); }
-  ::high-zoom[zoom >= 13]                 { image-filters: scale-hsla(0,1,0,1,0,  1,   0,1); }
+  ::low-zoom[zoom < 12] { image-filters: scale-hsla(0,1,0,1,0.2,1,0,1); }
+  ::high-zoom[zoom >= 12]                 { image-filters: scale-hsla(0,1,0,1,0,  1,   0,1); }
 
-  ::low-zoom[zoom < 10],
-  ::lower-mid-zoom[zoom >= 10][zoom < 11],
-  ::mid-zoom[zoom >= 11][zoom < 12],
-  ::upper-mid-zoom[zoom >= 12][zoom < 13],
-  ::high-zoom[zoom >= 13] {
+  ::low-zoom[zoom < 12],
+  ::high-zoom[zoom >= 12] {
 
   [feature = 'leisure_swimming_pool'][zoom >= 14] {
     polygon-fill: @water-color;
@@ -231,7 +224,6 @@
 
   [feature = 'landuse_residential'][zoom >= 8] {
     polygon-fill: @built-up-lowzoom;
-    [zoom >= 11] { polygon-fill: @built-up-z11; }
     [zoom >= 12] { polygon-fill: @built-up-z12; }
     [zoom >= 13] { polygon-fill: @residential; }
     [zoom >= 16] {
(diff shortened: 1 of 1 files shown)


imagico:
I am not really fond of this idea of a compromise between fading and not fading colors.  Most of the reasons why fading is a problem, in particular the problem of consistent design across zoom levels, do not vanish if you reduce the amount of fading.  But if this change helps weaning us off the fading idea i would not mind.

As i explained in https://github.com/gravitystorm/openstreetmap-carto/pull/3670#issuecomment-538914719 i think a consensus to fully get rid of fading would be a prerequisite for #3670 because both ideas represent opposing strategies towards low zoom map design.

jeisenbe:
This is meant as an intermediate step, since entirely removing the fading would be a much stronger change. I still intend to advocate for merging #3670 later: since we already combine urban land uses into one rendering at z12 and below, it makes the most sense to simplify the vegetation and rural landuse colors as well. 

Also note that the test images show the darker river-color. That's why I had been waiting to submit this PR until #3930 was merged.

kocio-pl:
I still have not working testing environment, but the examples look very good for me - they are both readable and nicely balanced.

I see problems with #3670, but I think they should be discussed there, not here.

jeisenbe:
>  it helps us working towards a consensus position

Yes, that is my understanding.

My goal is still to merge #3670

But by accepting this PR, it will make it easier to get to unfaded colors.

Right now there is a large difference between the "before" and "after" images in #3670 due to the current strong fading, and this distracts from the more subtle difference proposed there.

Unfortunately, @kocio-pl is not able to review this technically and @matthijsmelissen has not been available recently, so I am not sure who is going to be able to do a full review of this PR.

Adamant36:
@jeisenbe, can I do a review of it or is that something that has to be done by a Collaborator? Id like to see it go through. As the current fading irritates me.

jeisenbe:
Yes, anyone is welcome to do a review, and I would encourage contributors who have some experience with this style to make a review when they are interested in an issue, especially if they see a specific problem, but also if they have reviewed the PR carefully and think it should be approved.

Ideally a reviewer would [fetch and check out](https://github.com/gravitystorm/openstreetmap-carto/pull/3901#issuecomment-554737522) the branch ([jeisenbe:reduce-fade](https://github.com/jeisenbe/openstreetmap-carto/tree/reduce-fade)), and check how the cartographical changes look in a couple of representative areas. 

It's also possible to do a partial review of just the code changes, or just the cartography changes based on test images. Either way, it's best to use the button to "review changes" under the tab "[File changed](https://github.com/gravitystorm/openstreetmap-carto/pull/3952/files)", or if there is a comment about a particular line of code, it is helpful to select that line and comment there first.

jeisenbe:
Anyone willing to review this PR?

jeisenbe:
@kocio-pl are you willing to approve this PR?

kocio-pl:
Nothing changed from https://github.com/gravitystorm/openstreetmap-carto/pull/3952#issuecomment-546708377 - I still have no tools to test this technically, but I accept what is seen on visual examples you provided.

jeisenbe:
Anyone else willing to review this? 

While I agree that it is not a long-term solution to low zoom landcover rendering, I think incremental changes like this might make it easier to get back to a consensus-based solution, since any change would not be as large or sudden.

jeisenbe:
OK, I'm taking the comment by @kocio-pl as a positive cartographic review, @adamant36 is "thumbs up", and @imagico is not in favor of this as a long-term solution, but not opposed either if it helps get us back towards consensus.

Are there any other thoughts or reviews about this? I am considering merging this for the next release.

imagico:
> I'm taking the comment by @kocio-pl as a positive cartographic review

While as indicated i have no serious objections to this change although i don't see it solving any problem i would like to point out that our principle of every PR being reviewed by a maintainer in addition to the person making the PR requires actual independent testing to be meaningful.  The whole idea is to make sure someone independently looked at the change and its effect - testing it under circumstances other than those tested by the PR developer.  This is not the case here.

On a more general note - as we discussed in #3951 everyone constructively working towards a solution is a fundamental prerequisite for consensus based decision making to work.  If @kocio-pl objects to #3670 but is not able to actively work towards a different route to consensus either using this PR as a step or using a different approach that is not going to work.

jeisenbe:
Good point, so this PR still needs a formal review. Perhaps @Adamant36 would be able to test it?

Adamant36:
> Perhaps @Adamant36 would be able to test it?

Sure. I was actually going to ask if I could test it even though I'm not a maintainer.

Adamant36:
From the minor testing so far I'd say its a major improvement. Not having the fading transition between z12 and z13 makes zooming seem much more fluid. Also, the docker colors make farmland much easier to see zoomed out and parks look slightly less prominent because they are darker and therefore slightly more inline with other green colors. 

z12
z11

pnorman:
> Sure. I was actually going to ask if I could test it even though I'm not a maintainer.

Anyone can test and add a review. It's best to add a review rather than issue comments because then it's clear that it's been reviewed through the Github UI.

Adamant36:
> Anyone can test and add a review. It's best to add a review rather than issue comments because then it's clear that it's been reviewed through the Github UI.

Good call. I didn't know that was an option.

Issue #3607, opened 2018-12-28
Adjust societal amenity colors


jeisenbe:
### Background

- The societal amenties color was previously `#eef0d5` - Lch(94,14,112), a moderately light, slightly-greenish yellow
- This was recently switched (on 3-Sep-2018) in PR #3327, so that this color became the new farmland color, and societal amenities used the old farmland color `#fbecd7` - Lch(94,12,80), as suggested by @imagico 
- However, this new color was considered to have too high of contrast with pitch, grass and park areas, often found within university and school campuses, so PR #3412 was merged on 9-Oct-2018 to change societal_amenities color to `#ffffe5` - Lch(99,13,109)
- This new color was considered to look more harmonious with the park, pitch and garden green colors
- However, several people involved in that PR have mentioned that they would have liked more time to perfect the color, eg @Adamant36 

### Expected behavior

- The societal amenities color should has a similar lightness to other urban areas, so that roads, buildings and icons have the proper contrast with the background color. 
- The societal amenities line should be visible between two adjacent campuses
- The text label for an area should be similar to the area color.
- The color should be sufficiently different from that of other landcovers so that there is low risk of confusion

### Actual behavior

- The new societal amenities color is a very light yellow, much lighter than other landcovers and land uses. 
- White roads have too little contrast on this background
- Dark buildings, major roads, and POI icons have too much contrast
- The line color is too light, and it is difficult to see the division between adjacent schools or hospitals.
- The text label is much darker than the area color, so it is not clear that it is related
- The color is a little too close to that of farmland (delta 5.2, though this depends on the display used)

### Possible solutions

- Increase the chroma/saturation and darken the color slightly to better harmonize with other landcovers and land uses. 
- Perhaps change the hue to be slightly more orange and less green?
- Increase the contrast of the outline

**Options I've considered:**

1. slightly darker yellow, such as   `#f8f7c1`   LCH(96,27,107)
 - This color is a little close to beach (delta 5.2)
 - Too close to secondary road fill

2. Less saturated, a touch greener   `#f6f7c7`   Lch(96,24,109)
- Better difference from beach and sand than the first option
- But still close to secondary road fill

3. Less saturated, a touch lighter:  `#f5f3c6`  LCH(95,23,106)
- less similar to beach than the first option, still a little close to secondary road fill (delta 7.0), closer to sand (delta 6.6)

4. Similar to old farmland color but less saturated:  `#f8ecdd` - LCH(94,9,80)
[...]

PR #3957, opened 2019-10-28
Change sand to @bare_ground color and beach to sand, move sand pattern out of overlay and render till z13

jeisenbe:
Alternative to PR #3855

- Change `@sand` color to be same as @bare_ground `#eee5dc`  (also used for bare_rock, shingle, scree and mud on land)
- Rename pattern from beach.png to sand.png
- Move sand pattern to landcover layer so it will not render over water, from z13
- Change `@beach` to `#f5e9c6` - former @sand color

Sand (used for dunes and other area of wind-blow sand, not beaches) is distinguished from similar non-vegetated areas, scree and shingle, by both a pattern and a color currently.

I originally proposed removing the pattern in PR #3855, which would have made it possible to use a different pattern for dunes, but that PR is stalled.

By instead unifying sand with the `@bare_ground` color `#eee5dc` there is one less landcover color used. However, this makes beaches rather dissimilar to the adjacent sand dunes, so `@beach` is switched from the current color `#fff1ba` to the current sand color `#f5e9c6` 

(This may also make it possible to fix the societal-amenities color by using the current beach color or a similar color)

The pattern is moved to the `landcover` layer instead of `landcover-overlay` since it should not be rendered over water: sand which is in the intertidal zone should usually be tagged natural=beach (or natural=shoal). It now renders from z13, same as shingle and scree.

### Point of Ayr
https://www.openstreetmap.org/#map=15/53.3549/-3.3122
Before
z15-ayr-sand-beach-master

After
z15-ayr-beach-f5e9c6-sand-bare_ground

### Harlech
https://www.openstreetmap.org/#map=15/52.8613/-4.1034
before
z15-harlech-master-societal

After
z15-harlech-beach-f5e9c6-sand-bare_groun

### Rhyl
https://www.openstreetmap.org/#map=14/53.3151/-3.4828
before
z14-rhyl-master-societal

After
z14-rhyl-beach-f5e9c6-sand-bare_ground


imagico:
Regarding the sand color change - i don't think this would be beneficial compared to #3855 as already explained there - the choice in #1217 to use a distinct color for natural=sand had very good reasons and these reasons have not changed.  For practical human use the difference between natural=sand and natural=scree/bare_rock is huge and this change would reduce this difference to a minor footnote visible only at the higher zoom levels at a closer look.  See https://github.com/gravitystorm/openstreetmap-carto/pull/3855#issuecomment-529950933.

Regarding the beach color change - i don't think that is beneficial either.  The only reason given seems to be to make beaches appear less contrasting with natural=sand in the proposed different color.  But at the same time they would also be much closer to various other colors - or in other words:  They would be much less distinctive although beaches in reality are a very distinctive feature and are by definition of limited extent so well suited for a strong color.

Both changes together i think have a high potential to be confusing for mappers because of the confusing similarities and dissimilarities and are likely to blur the meaning of tags by providing less clear intuitive feedback to mappers.  So far i see no real reason for this change being proposed other than the 'political' reason that this change might achieve consensus among maintainers while #3855 - while being much more strait away, much better supported with arguments and with less side effects - cannot.

Also in terms of the use of colors in this style in general - i think we should not reduce the number of different colors indicating physical geography features but rather work on reducing the number of colors in use for human geography features.

kocio-pl:
For the comparison - this is how the shingle looks like (in Firefox):

Screenshot_2019-10-28 Linia 486985920 OpenStreetMap

This is too close IMO. However since shingle basically consists of bigger grains, maybe just making the grains pattern bigger would work.

jeisenbe:
Re #1217 - 44 comments from 5 people (including 2 maintainers at the time) in 4 days, and almost every one was helpful and specific. I wish we had that sort of development environment now. 

I see that in #1217 @imagico originally proposed unifying sand with the @bare_ground color used for scree and bare_rock, by using a somewhat coarser pattern for scree, but this was rejected because as mentioned above, "For practical human use the difference between natural=sand and natural=scree/bare_rock is huge", see e.g. https://github.com/gravitystorm/openstreetmap-carto/pull/1217#issuecomment-69462189.

> Regarding the beach color change ... they would be much less distinctive although beaches in reality are a very distinctive feature and are by definition of limited extent so well suited for a strong color.

Well, that's a good point. The current beach color does work better

> we should not reduce the number of different colors indicating physical geography features but rather work on reducing the number of colors in use for human geography features.

Since I was trying to steal the current @beach color to fix @societal-amenities in #3607 this is an important criticism.

> i see no real reason for this change being proposed other than the 'political' reason

I agree with you that sand shouldn't have both a separate pattern and a separate color, so if #3855 is rejected then this would be the only other option. I think it's helpful to see both choices and consider which is better. As mentioned, reducing the number of different landcover colors is also a worth goal

But I concede your point that it is mainly a problem for developed, human land-use rather than for natural areas, therefore it would be better to work on combining different human land-use colors like quarry, landfill and construction/brownfield, and golf_course, rather than removing the distinctive color for sand or beach.

jeisenbe:
@kocio-pl after seeing this option, do we now agree that it makes more sense to keep the current `@beach` and `@sand` colors, and merge #3855 instead?

kocio-pl:
Hold on, please... This was just a simple, partial observation, I still need to look at the different solution to make a synthesis. 

In general, you're trying to make a lot of important changes at once and I try to keep up and focus, which is hard. Testing different way helps, but more analysis in a short time is not the only tool to help decide things and sometimes too much data is a problem in itself.

kocio-pl:
Is it possible to make a list of all the expected effects we'd like to have (or avoid)? I mean any effects, even if they are contradicting, to avoid "premature optimization" problem and get back to the raw requirements. I think that could help review, since the subject seems even more complex after your remarks.

For example I want the similar things to have a similar look and different things to have a different look.

jeisenbe:
Would this be a list of all the important cartography principles and design
choices that we should follow in general?

Or specifically about how to choose landcover fill colors and patterns?

Or just the things that are important about natural=sand in particular?

kocio-pl:
I would like to focus on specific `natural=sand` rendering, since I'd like to resolve the issue eventually, not to go offtopic.

beach:
Not sure why I’m on this thread. Keep getting emails. Can’t unsubscribe. 

-- 
David Beach (He/Him)
Product Leadership and Strategy
davidbeach@gmail.com
408-896-0466
https://www.linkedin.com/in/beach/

jeisenbe:
So sorry, we’ve been discussing a color variable named ‘@beach’ and this
must have notified you again after you unsubscribed the first time.

I will avoid using the ‘@‘ symbol with this variable in future comments

beach:
Hah. Yeah I get this on Twitter too. I’m @beach there. Good luck with your project. :)

-- 
David Beach (He/Him)
Product Leadership and Strategy
davidbeach@gmail.com
408-896-0466
https://www.linkedin.com/in/beach/

jeisenbe:
More detailed explanation for proposed changes:

Sand is currently distinguished from similar areas, scree and shingle, by both a pattern and a color. The pattern is rendered in the landcover-overlay layers, so it shows over water and over other landcover/landuse areas.

The pattern is moved to the landcover layer instead of landcover-overlay since it should not be rendered over water: sand which is in the intertidal zone should usually be tagged natural=beach or natural=shoal + surface=*.

This PR would use the same color for sand and bare_ground, so one less landcover color would be used.

This may also make it possible to fix the societal-amenities color by using the current sand or beach color or a similar color, or allow another type of natural feature to be rendered.

Since this style shows many different types of area features, we need to minimize the number of different colors and patterns used. It is not necessary to use both a unique pattern and a unique color for `natural=sand`, so either the color or the pattern should be removed.

Another general principle is that similar colors should represent features that are functionally similar for a general map user. As mentioned [above](https://github.com/gravitystorm/openstreetmap-carto/pull/3957#issuecomment-547373147), we need "similar things to have a similar look and different things to have a different look"

For example, currently all areas of low vegetation use various hues in the light green range (grass, garden, meadow, heath, allotments, and parks), while high vegetation uses darker greens (orchard, forest, scrub).

Also, retail is similar to commercial, which is similar to industrial, and all have the same lightness as residential, as does farmyard - these are the basic types of developed landuse. 

Natural areas without vegetation all use similar colors: mud (over land), scree/shingle, bare_rock are the same, sand is similar, and beach is similar to sand, but not so close to mud. This makes sense because a beach is quite different to a mud flat as far as a general map user is concerned: you go to a beach for fun or a nice walk, while trudging across mud is difficult.

jeisenbe:
Also, patterns cannot be used at low zoom levels, because they are only legible when the area is moderately large: in the case of the sand pattern, the area needs to be at least a hundred square pixels for the pattern to be recognized. In contrast an area fill color can be clearly visible at just a few pixels, so it can be shown a few zoom levels sooner than a pattern.

10x10 pixel sand with pattern

kocio-pl:
I thought about requirements, like "sand should/shouldn't be similar to scree and shingle" - rather "what", not so much about "why". We can have very different explanations and visions (and we do), but still be able to agree on what to do.

jeisenbe:
> sand should/shouldn't be similar to scree and shingle

Yes, precisely. :-)

Sand should be somewhat similar to scree/shingle, but not too similar.

The same goes for the visual difference between the rendering of
`natural=sand` and `natural=beach`

kocio-pl:
You see? ;-) Defining desired properties first helps to make the discussion simpler and for me this is exactly the case where it's better to simplify a bit first. We still have our reasons and we can still discuss the results, but every "because" has a potential to go offtopic and discuss big things like visions instead of a sand rendering. This is interesting, but I have wasted much too much time this way and even more hard feelings, because we're attached to our visions, while not so much to the specific solutions.

So I guess we have following basic propositions:
- sand should be similar to scree/shingle and beach
- sand should be distinctive from scree/shingle and beach

What else?

jragusa:
Shingle and scree are mostly made of coarse materials (pebbles, cobbles and even boulders), while beach could be either sandy or entirely made polished pebble to cobble. Since sands and more especially clays ad silts are the most abundant sediments on Earth's surface, I would propose to only render a pattern for other kind of bare surface than sand which is already the case for bare_rock for example.

Hence, IMO, `natural=beach` should have a pattern when not associated to `surface=sand`suggesting the occurrence of coarse material as well as for shingle and scree.

The current rendering looks very strange in the first capture because beach without sand (implicitly with cobbles and pebbles) looks very flat while the rendering of sandy beach is very granular with this pattern. The inverse rendering would be correct.

jeisenbe:
> only render a pattern for other kind of bare surface than sand

OK, so you are opposed to this PR (3957) but in favor of #3855 instead, right?

> natural=beach should have a pattern when not associated to surface=sand suggesting the occurrence of coarse material as well as for shingle and scree.

That might be a reasonable cartographic decision for another map style, but since this style is used by mappers for feedback on the consistent use of tags, we need to show something for `surface=sand` if we are also going to render `surface=gravel/pebbles/stones/etc` differently, otherwise mappers will not be able to distinguish between a beach which is missing the `surface=` tag and one which is tagged with `surface=sand`

(Also, surface=sand is the by far the most common type of surface for `natural=beach` - 29k `natural=beach` features have a `surface=` tag and of those over 22k are `surface=sand` - https://taginfo.openstreetmap.org/tags/?key=natural&value=beach#combinations - so if we were not worried about mapper feedback, it would be best to assume that most beaches are sandy unless tagged otherwise)

jragusa:
> otherwise mappers will not be able to distinguish between a beach which is missing the surface= tag and one which is tagged with surface=sand

You raise a good point that I forgot: the lack of surface tag. So a pattern for sand is needed for mapper feedback.

jragusa:
Ok, I read the discussion of #3855 and I expect to not misunderstood it.

I think that the main problem is that `natural` tag encompasses two different kind of features in the framework of this discussion: (1) there is geomorphological features (beach, scree, shingle, dune) and features related to the sediment/lithology (sand, bare_rock). 

However, the first group of features is characterised by the second group. For example, as I said previously, you can have a sandy beach or a pebbly beach and that's the same for scree while dune is only made of sand and shingles are mainly composed of coarse grained materials.

So, I would propose to (1) standardise all the features (except beach) with the current colour of bare_ground and (2) add a pattern regarding to the sediment/lithology.

Hence, we should have 4 configurations of pattern regardless of the colour background: without pattern when there is no information about the surface (mapper feedback), a sandy pattern (beach, dune, scree, shoal), a coarser pattern for pebbles-cobbles (beach, scree, shingle) and the bare_rock pattern (bare_rock, reef, rock, cliff as a polygon ?).

Consequently, `natural=sand` should be rendered with both bare_ground colour and sandy pattern although I consider this feature as a catch-all. And we have not yet discussed about `natural=mud`...

kocio-pl:
@jragusa Thanks for bringing your voice, because it looks like the discussion stalled. I'm not sure right now about specific rendering, but the main idea of separating features from their surface is exactly what I think will make the things really clear. However that might be broader issue than just changing the rendering of `natural=sand` in this way or another.

jeisenbe:
>   sandy ... or pebbly... that's the same for scree

Are there areas of mainly sand, not including gravel or stones, tagged as `natural=scree`? I thought that `natural=scree` was used for slopes of loose gravel and stones: 

"Scree is an accumulation of unconsolidated angular rocks." [Tag:natural=scree](https://wiki.openstreetmap.org/wiki/Tag:natural%3Dscree)

"a collection of broken rock fragments at the base of crags, mountain cliffs, volcanoes or valley shoulders that has accumulated through periodic rockfall from adjacent cliff face" [wikipedia](https://en.wikipedia.org/wiki/Scree)

"large, loose broken stones on the side of a mountain, or an area covered with stones like this" - [Cambridge dictionaries](https://dictionary.cambridge.org/dictionary/english/scree)

I've tagged areas of sand in the mountains as `natural=sand`, not `natural=scree` - that I use for landslides and areas below cliffs with a mix of small and large stones (sometimes there is some sand mixed in as well).

> (1) standardise all the features (except beach) with the current colour of bare_ground

- Do you mean that you approve of #3855 which would change sand to @bare_ground. A

- Are you suggesting including other features in `@bare_ground` color, besides `mud`, `sand`, `scree`/`shingle`, and `bare_rock`?

> without pattern when there is no information about the surface

Only `natural=beach` lacks sufficient information: both `scree` and `shingle` imply loose stones, pebble to cobble sized, `mud` is fine particles (clay and silt), `sand` is sand, and `bare_rock` implies very large, non-moblie bolders, or solid bedrock. So I think approving #3855 is sufficient to match your suggestion?

jeisenbe:
Sorry, I got the PR numbers reversed. It appears @jragusa is approving of the general idea of this PR, #3957, which means closing #3855, if I understand the comment above correctly?

jeisenbe:
> > otherwise mappers will not be able to distinguish between a beach which is missing the surface= tag and one which is tagged with surface=sand

> You raise a good point that I forgot: the lack of surface tag. So a pattern for sand is needed for mapper feedback.

Only for beaches, not for `natural=sand`, because this implies `surface=sand`

jragusa:
> [...] Are there areas of mainly sand, not including gravel or stones, tagged as `natural=scree`? I thought that `natural=scree` was used for slopes of loose gravel and stones [...]

That's happen especially for weathered granitic rocks and unconsolidated or poorly cemented sedimentary rocks (mainly of quaternary age). In french, we use the term _arène_ for this kind of deposit (see also the english equivalent [wiki page](https://en.wikipedia.org/wiki/Decomposed_granite)). But yes, that's not a frequent case and they frequently of small size, so it can be neglected.

> [...]
>     * Do you mean that you approve of this PR, which would change sand to @bare_ground?
>     * Are you suggesting including other features in `@bare_ground` color, besides `mud`, `sand`, `scree`/`shingle`, and `bare_rock`?

Yes, I approve using @bare_ground colour for `natural=sand` to get an uniform background for these features excepted for `natural=beach`.

Concerning mud-clay, they are especially accumulated in quiet water areas such as tidal flat, marsh and essentially in deep marine settings. For some of them we have already a specific pattern.

> [...] Only `natural=beach` lacks sufficient information: both `scree` and `shingle` imply loose stones, pebble to cobble sized, `mud` is fine particles (clay and silt), `sand` is sand, and `bare_rock` implies very large, non-moblie bolders, or solid bedrock. So I think approving this PR is sufficient to match your suggestion?

I noticed that we have already a coarse pattern for beach `beach_coarse`. This pattern should also be used for scree and shingle. The current pattern for the latter is very slight in comparison and there is no reason IMO since both have made of equivalent coarse deposit. Furthermore, large block are preferentially found in rivers and not along beaches. For example:

https://www.openstreetmap.org/#map=18/46.40333/6.51925
beach-shingle

If you rename `beach.png` to `sand.png`, I recommend to do same for `beach_coarse.png` into `coarse.png` and eventually create a distinct issue to discuss about replacing `scree_overlay.png` by `coarse.png`

jeisenbe:
Closed, #3855 was merged instead.

PR #3964, opened 2019-11-04
Only use text-dy for wind generators, not for other power=generator features

### LINKED ISSUE #3963: Name rendered twice on power/building
Observed:
The name of building (way 737365087) is rendered twice.
https://osm.org/#map=19/51.54569/9.76438

Expected: just one label should be enough.

jeisenbe:
Fixes #3963

- Limit text rendering using `text-dy` to `generator:source=wind` only

Other power generators which are not buildings have their text labels rendered in the landcover font style, adjusted by way_area. Buildings do not need a separate rendering. Only wind generators have an icon, so only they need `text-dy: 10` for the name to be rendered properly

Test rendering:

Before
z19-fermenter-before

After
z19-fermenter-after

### DIFF
--- a/amenity-points.mss
+++ b/amenity-points.mss
@@ -1872,9 +1872,9 @@
     }
   }
 
-  [feature = 'power_generator'][location != 'rooftop'][location != 'roof'][zoom >= 17],
-  [feature = 'power_generator'][location = null][zoom >= 17],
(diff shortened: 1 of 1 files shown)


jeisenbe:
Anyone available to review this bugfix?

jeisenbe:
Review needed

PR #3969, opened 2019-11-11
Remove rendering of barrier=kerb

### LINKED ISSUE #3714: Different rendering for barrier=kerb
**barrier=kerb** has the same rendering as **barrier=wall** so there is no difference between a kerb and a wall on the map.

It could be rendered as a dashed line. Here is a example of the french cadastre were the dashed lines are kerbs.

kerb_cadastre

### LINKED ISSUE #3744: Stop rendering barrier=embankment
### Barrier=embankment vs man_made=embankment

I researched the values of barrier in taginfo to see if there were any that we could start or stop rendering. I found that barrier=embankment was not documented on the wiki.  

It turns out that most features with this tag were imported into the database in 2011, before man_made=embankment was an established tag. A few more were imported in 2013. As of April 2019, barrier=embankment is used 4750 times, less than the usage in 2013.

A better documented and supported tags this feature is man_made=embankment. This tag is well documented on the wiki and also rendered with a more specific style, and man_made=embankment has been used over 80,000 times.

The comparison graph of usage speaks loudest:
embankment-man_made-vs-barrier

jeisenbe:
Closes #3714 

- <s>Remove the rendering of barrier=embankment</s>
- Remove the rendering of barrier=kerb 

### Explanation:
- Kerbs (aka curbs) are rendered the same as significant obstacles like walls and fences. In issue #3714 I attempted to find a subtle rendering for kerbs which would not look like a fence or wall, but failed to find a good option for z19. Without rendering sidewalk and street areas these features are not intuitive.
- <s>The tag `barrier=embankment` was once somewhat common but never was documented in the wiki. Instead `man_made=embankment` or `man_made=dyke` are the much more common ways to make one-sided embankments and levees. The definition of `barrier=embankment` was not clear, so it's best to no longer support this tag.</s> (See #3744)

### Test rendering:

(Test shows fence, kerb, wall, embankment from top to bottom)
Before
kerb-embankment-test-before

After
kerb-embankment-barrier-after

### DIFF
--- a/landcover.mss
+++ b/landcover.mss
@@ -861,6 +861,10 @@
     line-width: 0.4;
     line-color: #444;
   }
+  [feature = 'barrier_embankment'][zoom >= 14] {
+    line-width: 0.4;
+    line-color: #444;
+  }
   [feature = 'barrier_hedge'][zoom >= 16] {
     line-width: 1.5;
     line-color: @hedge;
--- a/landcover.mss
+++ b/landcover.mss
@@ -861,10 +861,6 @@
     line-width: 0.4;
     line-color: #444;
   }
-  [feature = 'barrier_embankment'][zoom >= 14] {
-    line-width: 0.4;
-    line-color: #444;
-  }
   [feature = 'barrier_hedge'][zoom >= 16] {
     line-width: 1.5;
     line-color: @hedge;
--- a/project.mml
+++ b/project.mml
@@ -583,11 +583,11 @@ Layer:
             way, COALESCE(historic, barrier) AS feature
           FROM (SELECT way,
             ('barrier_' || (CASE WHEN barrier IN ('chain', 'city_wall', 'ditch', 'fence', 'guard_rail',
-                  'handrail', 'hedge', 'kerb', 'retaining_wall', 'wall') THEN barrier ELSE NULL END)) AS barrier,
+                  'handrail', 'hedge', 'retaining_wall', 'wall') THEN barrier ELSE NULL END)) AS barrier,
             ('historic_' || (CASE WHEN historic = 'citywalls' THEN historic ELSE NULL END)) AS historic
             FROM planet_osm_line
             WHERE barrier IN ('chain', 'city_wall', 'ditch', 'fence', 'guard_rail',
-                  'handrail', 'hedge', 'kerb', 'retaining_wall', 'wall')
+                  'handrail', 'hedge', 'retaining_wall', 'wall')
               OR historic = 'citywalls'
               AND (waterway IS NULL OR waterway NOT IN ('river', 'canal', 'stream', 'drain', 'ditch'))
           ) AS features
@@ -618,11 +618,11 @@ Layer:
             way, COALESCE(historic, barrier) AS feature
           FROM (SELECT way,
             ('barrier_' || (CASE WHEN barrier IN ('chain', 'city_wall', 'ditch', 'fence', 'guard_rail',
-                  'handrail', 'hedge', 'kerb', 'retaining_wall', 'wall') THEN barrier ELSE NULL END)) AS barrier,
+                  'handrail', 'hedge', 'retaining_wall', 'wall') THEN barrier ELSE NULL END)) AS barrier,
             ('historic_' || (CASE WHEN historic = 'citywalls' THEN historic ELSE NULL END)) AS historic
             FROM planet_osm_polygon
             WHERE (barrier IN ('chain', 'city_wall', 'ditch', 'fence', 'guard_rail',
-                  'handrail', 'hedge', 'kerb', 'retaining_wall', 'wall')
+                  'handrail', 'hedge', 'retaining_wall', 'wall')
               OR historic = 'citywalls')
               AND building IS NULL
           ) AS features
--- a/project.mml
+++ b/project.mml
@@ -582,11 +582,11 @@ Layer:
         (SELECT
             way, COALESCE(historic, barrier) AS feature
           FROM (SELECT way,
-            ('barrier_' || (CASE WHEN barrier IN ('chain', 'city_wall', 'embankment', 'ditch', 'fence', 'guard_rail',
+            ('barrier_' || (CASE WHEN barrier IN ('chain', 'city_wall', 'ditch', 'fence', 'guard_rail',
                   'handrail', 'hedge', 'kerb', 'retaining_wall', 'wall') THEN barrier ELSE NULL END)) AS barrier,
             ('historic_' || (CASE WHEN historic = 'citywalls' THEN historic ELSE NULL END)) AS historic
             FROM planet_osm_line
-            WHERE barrier IN ('chain', 'city_wall', 'embankment', 'ditch', 'fence', 'guard_rail',
+            WHERE barrier IN ('chain', 'city_wall', 'ditch', 'fence', 'guard_rail',
                   'handrail', 'hedge', 'kerb', 'retaining_wall', 'wall')
               OR historic = 'citywalls'
               AND (waterway IS NULL OR waterway NOT IN ('river', 'canal', 'stream', 'drain', 'ditch'))
@@ -617,11 +617,11 @@ Layer:
         (SELECT
             way, COALESCE(historic, barrier) AS feature
           FROM (SELECT way,
-            ('barrier_' || (CASE WHEN barrier IN ('chain', 'city_wall', 'embankment', 'ditch', 'fence', 'guard_rail',
+            ('barrier_' || (CASE WHEN barrier IN ('chain', 'city_wall', 'ditch', 'fence', 'guard_rail',
                   'handrail', 'hedge', 'kerb', 'retaining_wall', 'wall') THEN barrier ELSE NULL END)) AS barrier,
             ('historic_' || (CASE WHEN historic = 'citywalls' THEN historic ELSE NULL END)) AS historic
             FROM planet_osm_polygon
-            WHERE (barrier IN ('chain', 'city_wall', 'embankment', 'ditch', 'fence', 'guard_rail',
+            WHERE (barrier IN ('chain', 'city_wall', 'ditch', 'fence', 'guard_rail',
                   'handrail', 'hedge', 'kerb', 'retaining_wall', 'wall')
               OR historic = 'citywalls')
               AND building IS NULL
--- a/project.mml
+++ b/project.mml
@@ -582,11 +582,11 @@ Layer:
         (SELECT
             way, COALESCE(historic, barrier) AS feature
           FROM (SELECT way,
-            ('barrier_' || (CASE WHEN barrier IN ('chain', 'city_wall', 'ditch', 'fence', 'guard_rail',
+            ('barrier_' || (CASE WHEN barrier IN ('chain', 'city_wall', 'embankment', 'ditch', 'fence', 'guard_rail',
                   'handrail', 'hedge', 'retaining_wall', 'wall') THEN barrier ELSE NULL END)) AS barrier,
             ('historic_' || (CASE WHEN historic = 'citywalls' THEN historic ELSE NULL END)) AS historic
             FROM planet_osm_line
-            WHERE barrier IN ('chain', 'city_wall', 'ditch', 'fence', 'guard_rail',
+            WHERE barrier IN ('chain', 'city_wall', 'ditch', 'embankment', 'fence', 'guard_rail',
                   'handrail', 'hedge', 'retaining_wall', 'wall')
               OR historic = 'citywalls'
               AND (waterway IS NULL OR waterway NOT IN ('river', 'canal', 'stream', 'drain', 'ditch'))
@@ -617,11 +617,11 @@ Layer:
         (SELECT
             way, COALESCE(historic, barrier) AS feature
           FROM (SELECT way,
-            ('barrier_' || (CASE WHEN barrier IN ('chain', 'city_wall', 'ditch', 'fence', 'guard_rail',
+            ('barrier_' || (CASE WHEN barrier IN ('chain', 'city_wall', 'ditch', 'embankment', 'fence', 'guard_rail',
                   'handrail', 'hedge', 'retaining_wall', 'wall') THEN barrier ELSE NULL END)) AS barrier,
             ('historic_' || (CASE WHEN historic = 'citywalls' THEN historic ELSE NULL END)) AS historic
             FROM planet_osm_polygon
-            WHERE (barrier IN ('chain', 'city_wall', 'ditch', 'fence', 'guard_rail',
+            WHERE (barrier IN ('chain', 'city_wall', 'ditch', 'embankment', 'fence', 'guard_rail',
                   'handrail', 'hedge', 'retaining_wall', 'wall')
               OR historic = 'citywalls')
               AND building IS NULL


Adamant36:
How about some real world examples?

jeisenbe:
https://www.openstreetmap.org/#map=16/31.9764/10.6907
barrier=embankment in Libya - there is a `barrier=ditch` parallel to it which still renders

before
barrier-embankment-before

after
barrier-embankment-after

https://www.openstreetmap.org/#map=19/43.60627/3.87779
barrier=kerb in Montpellier, France

before
barrier-kerb-before

after
barrier-kerb-after

Adamant36:
Yeah, looks better without them. No big surprise there though, but thinks for the test renderings anyway.

polarbearing:
I disagree with removing them in the same PR. The reasons are completely different.
* barrier=embankment removal is fine for me, since is is a mistagging and deprecated
* kerbs are not deprecated, and are a physical barrier, see my reasoning in #3714, they should not be removed.

Prince-Kassad:
a curb is no more a physical barrier than a stairwell is, except possibly to wheelchair users but we don't even distinguish between high curb and low curb so that's pretty useless. OpenStreetMap is not really suited to routing for wheelchair or similarly impared people (elderly people for example) and it's not a trivial thing to do.

If you want an extreme example look at Neu-Isenburg: https://www.openstreetmap.org/#map=19/50.05338/8.68732&layers=N It makes the city look like some kind of walled city.

polarbearing:
That Neu-Isenburg example has other weird mapping, e.g. mapping small green patches on the sidewalk as 'garden'. 
However it shows that if we remove kerb but keep rendering parking_space, we would render lines painted on the road but ignoring barriers set in stone.

jeisenbe:
I tried a number of different renderings for `barrier=kerb` above ([1](https://github.com/gravitystorm/openstreetmap-carto/issues/3714#issuecomment-483084362), [2](https://github.com/gravitystorm/openstreetmap-carto/issues/3714#issuecomment-483082525), [3](https://github.com/gravitystorm/openstreetmap-carto/issues/3714#issuecomment-482978790) [4](https://github.com/gravitystorm/openstreetmap-carto/issues/3714#issuecomment-482907766)), some of which required extensive changes to the rendering of other barriers, but none worked well.

Currently kerbs are rendered the same as retaining walls, walls and fences, which are mostly tall barriers that prevent entry by pedestrians. Since barrier=wall and barrier=fence are used millions of times and barrier=kerb is rare, the cases when barrier=kerb is rendered are going to be interpreted by most map users as a fence or wall. This is misleading and should be fixed.

Prince-Kassad:
> However it shows that if we remove kerb but keep rendering parking_space, we would render lines painted on the road but ignoring barriers set in stone.

I don't think I need to say I'm not a fan of rendering parking spaces like this. It kind of reminds me of back when barrier= was rendered regardless of value, and people used barrier=line to draw the white lines of soccer fields so it shows as a soccer field on the map instead of just green square. We got rid of that for a reason.

polarbearing:
> barrier=kerb is rare

225000 uses is not 'rare'.
> misleading and should be fixed

rendering it later would fix it.

jeisenbe:
> rendering it later would fix it.

Rendering it only at z20, but still the same as `barrier=wall`, would still be highly misleading at z20. I attempted a subtle distinction between `barrier=kerb` and the other tags in https://github.com/gravitystorm/openstreetmap-carto/issues/3714#issuecomment-482907766 but this did not work well.

At z19 the current rendering is poor (I did not hand-pick this example, it was the first place I found in Montpellier with kerbs):

barrier-kerb-before

>225000 uses is not 'rare'.

OK, you caught me (I didn't check the numbers). It would be better to say they are "uncommon" compared to fence+wall+other barriers rendered in the same way. Note that when I first looked for `barrier=kerb` features, I did not find very many: 

"There are none mapped in Prince Edwards Isle (Canada) and only 3 in Liechtenstein. Delaware, USA has only 1. Washington, DC has a couple dozen, but all of them are also part of another feature. Luxembourg was the first place I found that had some kerbs mapped as separate features"

This means that many map users are not going to be used to seeing them. Rendering them only at high zoom levels would increase this problem.

kocio-pl:
> Rendering it only at z20, but still the same as barrier=wall, would still be highly misleading at z20.

I don't think there is a chance of such confudion. Rendering black line has multiple uses and using zoom level context helps here (for example train line and tram line has this distinction). There are also common life observation of wall not being built exactly on the edge of the road.

jeisenbe:
> wall not being built exactly on the edge of the road.

Many of the currently-rendered kerbs are not right on the enge of the road (and at low to mid lattitudes most kerbs right on the side of the road are covered by the highway rendering). Example where kerb is about the same distance from the road as a fence:

z19-rue-carlo-current

Kerb on northeast (upper left), fence on southeast (lower right) side.

Here's the test I tried with slightly wider walls/fences and slightly narrower kerb:

z19 after - 0.3 pixel wide kerbs (center, north, west), 0.6 pixel wide fence (southeast)
z19-rue-carlo-after2

It is not clear that there is kerb in the upper left and a fence in the lower right; one would be more inclined to guess fence vs wall perhaps.

I also tried this more complicated change, with a new fence/wall rendering as well:

z19-kerb-halo-fence-dots

But it doesn't work well because kerbs are not consistently mapped in a certain direction

kocio-pl:
In the given example there are trees outside the fence, you would not expect them on the road.

> But it doesn't work well because kerbs are not consistently mapped in a certain direction

I don't get why the certain direction is needed here?

Just making kerbs lighter like on the last eample (no need to change fences closer to the walls) and rendering them on z19+ would make it different enough. You skipped the core of the comment about zoom level context.

kocio-pl:
There are enough context tips from [this location](https://www.openstreetmap.org/#map=18/49.63435/6.17173) you don't show by skipping z18 rendering - the fence is attached to a building, there's a gate and private road leading inside; certainly not a kerb properties:

Screenshot_2019-11-12 OpenStreetMap

I think that kerb on z18 is still too detailed feature, but moving it to appear on z19 together with lighter rendering is good enough combination.

jeisenbe:
> I don't get why the certain direction is needed here?

Note the inconsistency between the curb in the middle, and the ones to the north around the grassy area: the white lines on opposite sides (it should be on the outside, or lower side, of the curb in both cases, theoretically). 

> Just making kerbs lighter like on the last example 

Do you mean the second to last example?

> (no need to change fences closer to the walls) 

 Currently `barrier=fence` and `barrier=wall` (and barrier=retaining_wall) are all rendered identically, and in the tests above that is still the case. 

But note that none of my ideas above achieved consensus, see critical comments: https://github.com/gravitystorm/openstreetmap-carto/issues/3714#issuecomment-482910247, https://github.com/gravitystorm/openstreetmap-carto/issues/3714#issuecomment-482987207 (matkoniecz), https://github.com/gravitystorm/openstreetmap-carto/issues/3714#issuecomment-482994980 (imagico), and my own comments after the last tests.

> core of the comment about zoom level context.

2 other contritubors agreed that this feature (`barrier=kerb`) will only work for rendering on z20. Currently we are not rendering any features at z20 only, since this is not used on openstreetmap.org yet, and I believe we need to wait until there is a possibility of z20 being used before reconsidering rendering `barrier=kerb` at that level.

jeisenbe:
> there's a gate and a private road leading inside

Right now there is no established way to tag a "curb cut", or a place where wheelchair and bicycle/scooter users can roll across a curb. This is one of the problems of rendering kerbs currently. See https://github.com/gravitystorm/openstreetmap-carto/pull/3969#issuecomment-552425725 and https://github.com/gravitystorm/openstreetmap-carto/issues/3714#issuecomment-482925441 which mentioned this issue with kerbs.

kocio-pl:
> 2 other contritubors agreed that this feature (barrier=kerb) will only work for rendering on z20.

There is also a https://github.com/gravitystorm/openstreetmap-carto/pull/3969#issuecomment-552421266 about just different rendering (though I'm not sure what initial zoom level @polarbearing meant).

> Right now there is no established way to tag a "curb cut", or a place where wheelchair and bicycle/scooter users can roll across a curb.

We talk about rendering, not routing, which has it's own problems (the existence of a gate does not automatically inform you if it's passable nor if it's not passable). There are plenty of lacking gates on the map and that is just incomplete, not wrong.

> Do you mean the second to last example?

I mean "z19 after - 0.3 pixel wide kerbs (center, north, west)" example.

> Currently barrier=fence and barrier=wall (and barrier=retaining_wall) are all rendered identically, and in the tests above that is still the case.

I meant just no need only in the context of kerb. There might be different reasons to change them and that would be fine. "No need" does not mean "they can't be changed".

> Note the inconsistency between the curb in the middle, and the ones to the north around the grassy area: the white lines on opposite sides (it should be on the outside, or lower side, of the curb in both cases, theoretically).

What white lines are you talking about? Did you use them instead of just a lighter shade of grey? I would not try composing lines here at all.

Adamant36:
> a curb is no more a physical barrier than a stairwell is

@Prince-Kassad, they are in parking lots. Like for instance in this one that I probably over mapped, but it gives some visual context of where not to drive, or something. 

Curb

> Without rendering sidewalk and street areas these features are not intuitive.

@jeisenbe, it seems pretty intuitive with just sidewalk and street ways. Know one will confuse these examples for walls for fences. Its pretty obvious from the context that they are curbs. 

Curbs and sidewalks

Garden

I like the way the curbs in this case kind of create a visual border for the space of the theater from the other stuff around it. In some cases, like this one and the parking lot example, they add something to the map. 

Theater

imagico:
I support:

* stopping rendering barrier=embankment due to it being widely considered an undesirable synonym for man_made=embankment.
* stopping rendering of barrier=kerb in the same styling as other barriers due to this being highly misleading and against our documented goals.
* stopping rendering barrier=kerb for the moment waiting for a convincing suggestion how to render it and arguments why we should render it.

polarbearing:
I propose:
* splitting the PR, thus stopping rendering barrier=embankment which is undisputed
* keeping barrier=kerb while discussing how to change rendering. I find dropping it for a while and then introducing differently counter-productive.

kocio-pl:
Thank you both for concise propositions.

@jeisenbe In general it is always good to separate features. I understand that it's easier to write one ticket for a short and similar code, and in some cases that is OK, but you never know what will be the problem to be discussed.

jeisenbe:
I don't think that removing `barrier=embankment` from this discussion will make a significant difference to our decision, but to avoid the distraction I have pushed a new commit which removes the changes to `barrier=embankment`, so we can focus the discussion on `barrier=kerb`.

I would be happy to see other rendering ideas for `barrier=kerb`. So far I spent at least 8 hours trying to adjust the various barrier features to find a reasonable rendering for kerbs at z19 which could not be confused with other barriers, but was still clearly visible, but was not successful.

If there are no new idea, then we should remove the rendering for now, until a successful rendering option is found. As I've mentioned, I think this is much more likely at z20, in concert with showing `area:highway=` features, which would make curbs much more intelligible in context.

imagico:
> keeping barrier=kerb

Obviously this is what will happen if there is no consensus but as mentioned before this would actually be in clear conflict with the goals of this style because rendering kerbs the same way as physical barriers is misleading mapper feedback.

In other words: Our inability to achieve consensus on questions like this results in an inability to pursue the goals and the mission of this project.

jeisenbe:
> rendering kerbs the same way as physical barriers is misleading mapper feedback.

It also interferes with map **Legibility and clarity** when very different features are rendered identically. 

_"The map should be intuitively readable by users with some general experience using maps without a map key, preferrably with relatively little effort"_ - Line 34, CARTOGRAPH.md

jeisenbe:
> Like for instance in this one that I probably over mapped, but it gives some visual context of where not to drive, or something. 

Curb

This is a good example of how the current rendering is misleading. There is no way to tell if the lines around the spots with trees in the parking lots are walls, fences, or just curbs. A person walking can cross the area without breaking stride if there is only a curb, but if there is a fence or wall they will go around.

kocio-pl:
> This is a good example of how the current rendering is misleading. 

We cluster some features into the same rendering, so it depends on how much difference is needed. Barrier is a common part (class) of kerb/fence/wall and on some level it is good enough - you know that it's not a road or a building. So I would rather say this is not detailed enough, not misleading.

So the proposition has been given that will be clear that this is not the same as other barriers, but still looks like an instance of this class.

jeisenbe:
> So the proposition has been given that will be clear that this is not the same as other barriers, but still looks like an instance of this class

Which proposition? I don't understand the sentence above, I'm afraid.

> We cluster some features into the same rendering, so it depends on how much difference is needed

Consider how differently we render `railway=rail` and `service=` which
look nearly the same on-the-ground and both could have trains. We also
have a quite distinctive rendering for gondolas versus chairlifts -
that is, enclosed versus open "aerialways", and we distinguish
cycleways from footpaths, and even highway=pedestrian from
highway=footway

We distinguish hedges from fences/walls and from citywalls.

I believe it's very reasonable to expect that a kerb/curb would be
rendered differently than a wall.

polarbearing:
I propose to render kerb identical to parking_space, which is a light line at zoom>=18.

The rationale is that both are "soft" demarcations that however can be easily walked or driven over.

kocio-pl:
> Which proposition? I don't understand the sentence above, I'm afraid.

I already replied when you were asking which proposition I mean (_I mean "z19 after - 0.3 pixel wide kerbs (center, north, west)" example._).

I am not attached to any specific rendering however. The general guiding rules are to make the line lighter and start rendering later than fence and wall.

> I believe it's very reasonable to expect that a kerb/curb would be rendered differently than a wall.

I agree that it's reasonable. I'm not sure why are you trying to explain that. Did you think that I don't support that or just analysing aloud?

What I mean is that it is not an error to render barriers the same way, but it's better if we can tune the rendering to show them in a different way. Proposition has been made which meets this wish.

I don't remember what is the parking lots color, but I would use darker grey, because kerbs are much more passable then fences, but still less then lines on the ground.

jeisenbe:
Thank you for the clarification. So you prefer this rendering option

**0.3 pixel wide kerbs and 0.6 pixel wide fences/walls**

Note that this requires changing the rendering of fences/walls from 0.4 to 0.6 pixels at z19, in addition to changing the rendering of kerbs from 0.4 to 0.3

The problem with that rendering suggestion is that some of the other maintainers and contributors already gave arguments against this rendering of kerbs as simply thinner barriers:

https://github.com/gravitystorm/openstreetmap-carto/issues/3714#issuecomment-482910247:
"even with a lighter, thinner line, the kerb isn't significantly different than another barrier. If you were to see only a kerb on a map, you wouldn't have anything to compare to, so it would still be tough to differentiate"

"I think that simply not rendering barrier=kerb is worth considering. I think that it is significantly less important barrier than other ones.
"Note that we render barrier=gate and similar passages through walls and fences. We are not doing it with kerbs." - @matkoniecz 
https://github.com/gravitystorm/openstreetmap-carto/issues/3714#issuecomment-482925441

"fortunately this is still not clear. I worry that wall is so far away from kerb that it may be not doable to make both visible and properly different."- @matkoniecz 
https://github.com/gravitystorm/openstreetmap-carto/issues/3714#issuecomment-482987207

The last comment was in response to the more complicated change, which tried to make walls/fences even more different, but still showed kerbs as a thin dark line (0.3pixels, with a thin white line beside it).

Since our decision-making here is based on getting consensus amongst all of the maintainers and hopefully among all contributors, it appears that the option of "z19 only kerbs 0.3 pixels wide, walls/fences 0.6 pixels wide on z19" is not going to get consensus. I should mention that I agree with @matkoniecz that the rendering was not sufficiently different.

jeisenbe:
> > I propose to render kerb identical to parking_space, which is a light line at zoom>=18. The rationale is that both are "soft" demarcations that however can be easily walked or driven over.

> I don't remember what is the parking lots color, but I would use darker grey, because kerbs are much more passable then fences, but still less then lines just painted on the ground.

There were already 2 critical comments about rendering parking spaces the same as kerbs above: https://github.com/gravitystorm/openstreetmap-carto/pull/3969#issuecomment-552427916 and https://github.com/gravitystorm/openstreetmap-carto/pull/3969#issuecomment-552430974

The current parking_space rendering has a number of issues. The color is a 50% mix of the parking outline color with the parking fill color, so it can't be reused for kerbs directly. 

@kocio-pl suggests using gray for kerbs, presumably lighter than the current barrier color but darker than the parking outline. This would be slightly different than the previous tests where the kerb line as still in `#444` (dark gray), same as fences and walls, but just thinner.

The one other proposition previously was rendering kerbs with a dashed line. Due to the poor rendering results of dashed lines on small or strongly curved features, this will not work at z18 and might not work at z19, though it would probably be acceptable at mid to high lattitudes at z20.

I will test these ideas, though I admit I am not looking forward to spending more time on this issue. I feel like we are spending all this time discussing a minor, uncommon and not very important map feature which will only be rendered at one zoom level, when there are 400 open issues, many of which are more significant, and almost 30 active PRs which could also be discussed.

kocio-pl:
I am still not attached to any specific kerb rendering. This is why I wrote my guiding rules. I think that https://github.com/gravitystorm/openstreetmap-carto/issues/3714#issuecomment-482910247 gives a new context which has not been discussed yet and it's good as a starting point.

> Note that we render barrier=gate and similar passages through walls and fences. 

But even without them nobody claims that it's misleading if the road goes through them, so I don't see this as a reason to not render barriers.

> I will test these ideas, though I admit I am not looking forward to spending more time on this issue.

Since this is a different discussion than simple code deployment, it might be better to open an issue ticket for kerb improvements.

jeisenbe:
> nobody claims that it's misleading if the road goes through them [fences and walls], so I don't see this as a reason to not render barriers.

I would claim that it is misleading. That's part of the reason why we render `line-barriers` underneath roads, so that the road does not appear to be blocked by the fence or wall. But it makes sense when there is a `barrier=gate` or `barrier=stile` or similar point barrier, to show that there is something there. 

> issue ticket for kerb improvements

That would be #3714 - unfortunately I will not have time to work on this for the next couple of weeks due to traveling.

> I think that #3714 (comment) gives a new context which has not been discussed yet

What part of that comment are you referring to? Here it is in context:

>> Right now, with kerbs rendered the same as fences and walls, the rendering is unhelpful. A kerb certainly needs to be rendered differently than a wall. It's not possible to make kerbs much thinner or lighter than the current barrier rendering without confusing them with the outline of an area. ....
>> I don't think a dashed or dotted line would work, because it would not be clear that this was a barrier type feature.

> I agree with everything you've posted here. My only thought is that even with a lighter, thinner line, the kerb isn't significantly different than another barrier. If you were to see only a kerb on a map, you wouldn't have anything to compare to, so it would still be tough to differentiate. I don't have a better solution, though.

> I've seen versions of fences with a dash-dot-dash style, which made them very obvious, but that doesn't match the current barrier style of a solid line...

Were you referring to the suggestion to render `barrier=fence` with a dash-dot-dash style, like on some small scale Swiss maps and old OS (British) small scale maps?

pnorman:
👍 in principle to removing barrier=kerb rendering, but I haven't done a technical review

polarbearing:
> we render line-barriers underneath roads, so that the road does not appear to be blocked by the fence or wall

Do we? So if I erect a wall across a road, and map accordingly, the road appears to go through? Why?

Adamant36:
@jeisenbe, I take it this no longer fixes #3744 like your original comment suggests since it's not dealing with embankments anymore? Thought I'd mention it so the issue doesn't get auto closed when this gets merged if it ever does. 

Also, I put some thought into this and I'm changing my vote to remove. The reasons are 

1. I realized I, and probably a lot others, mostly map curbs for the rendering since they serve no real world purpose as concerns the style (for instance acting as a visual aid to wheel chair bound people). 

2. Like I @matkoniecz mentioned, places where the curb goes down are not rendered. Which makes things look like one long, continuous curb when they might not be. Which is just miss-leading. 

3. There's also the problem of them looking like fences that seems to have no current, acceptable way of dealing with. Even I didn't find any of @jeisenbe's alternate tests compelling and usually pretty liberal about that kind of thing. 

4. The case could be made that by rendering kerbs it is encouraging micro-mapping and in an extremely inaccurate way. Since in a good portion of the world the satellite photos aren't good enough to map such small features accurately yet. I actually tried to map some kerbs down the street from me once using a combination of the satellite photos, which aren't to bad around where I live, and GPS tracks. Even then it has hard to do and not as precise as it should have been. It's more a novelty then anything. 

5.  I was trying to find the original issue here on Github for when kerbs were added to see what the rational or thinking behind adding them was and I couldn't find it. Normally that wouldn't be a problem, but I feel like in cases like this, where there are clearly problems with the current rendering, maybe it would be better to remove the rendering and go through the rationalization process all over again if someone wants them to be added back. While taking into account the "modern" shortcomings of the style. Which, honestly, I doubt will happen. More then likely know-one will even notice if they are removed. Let alone ask for them to be rendered again. I don't think it will effect how much they are mapped or not either.

imagico:
Regarding the origin of kerb rendering - i mentioned this in #3714:

> Side note: By rendering barriers with a catch-all until #1986 we are largely responsible for the invention of barrier=kerb as a textbook _tagging for the renderer_. We failed to correct this in #1986 unfortunately.

This is a great example why we need to avoid catch-all renderings.

jeisenbe:
> I take it this no longer fixes #3744

Right, thank you for noticing that. I've changed the comment.

> We failed to correct this in #1986 unfortunately.

I think it is important to remember this: there was never a discussion about rendering barrier=kerb in particular

matkoniecz:
I fully support removal of `barrier=curb` rendering, especially in case where it is rendered like fences, wall etc.

polarbearing:
How would the paradox be solved that we'd still render painted lines for parking_space?

Adamant36:
What do the two have to do with each other? My understanding is that parking space rendering is light and different enough to not be confused with anything else or to give the impression that its a true barrier.

jeisenbe:
> How would the paradox be solved that we'd still render painted lines for parking_space?

See #3974 and my suggestion in https://github.com/gravitystorm/openstreetmap-carto/issues/3974#issuecomment-554692212

matkoniecz:
> How would the paradox be solved that we'd still render painted lines for parking_space?

Decision to change rendering of this feature (or maybe doing it) can be done in a separate PR.

polarbearing:
Sure it could, but it will take a while. In the meantime, we are replacing a minor discrepancy (rendering small and high barriers equal) with a larger one (rendering a painted line but suppressing a stone feature).

jeisenbe:
4 out of 5 maintainers of this style who have commented have approved of this PR, and of the other 3 contributors who commented 2 are in favor.

I've tried a number of different renderings, in good faith (I wanted to be able to render kerbs) without success, and I don't have any other good ideas for how this feature can be clearly rendered in a way that cannot be confused with significant linear barriers like fences and walls.

I believe this is rough consensus to remove the rendering of barrier=kerb. Of course anyone is welcome to submit a PR with a new type of kerb rendering, especially if it is related to a significant redesign for z20 or z21 and above where it would be more feasible.

kocio-pl:
The only thing that IIRC is currently shared by everyone is that it does not need to be rendered before z19, removing is not even proposed in some last comments, so this PR does not work for any of them, so it should be closed anyway.

jeisenbe:
That's an interesting interpretation of: 

1) "[I fully support removal of barrier=curb rendering](https://github.com/gravitystorm/openstreetmap-carto/pull/3969#issuecomment-562964256)" @matkoniecz 

2) "[👍 in principle to removing barrier=kerb rendering, but I haven't done a technical review](https://github.com/gravitystorm/openstreetmap-carto/pull/3969#issuecomment-556688728)" @pnorman 

3) "[I support: stopping rendering of barrier=kerb in the same styling as other barriers due to this being highly misleading and against our documented goals.](https://github.com/gravitystorm/openstreetmap-carto/pull/3969#issuecomment-552876168)" @imagico 

4) Myself "Kerbs (aka curbs) are rendered the same as significant obstacles like walls and fences. In issue #3714 I attempted to find a subtle rendering for kerbs which would not look like a fence or wall, but failed to find a good option for z19. Without rendering sidewalk and street areas these features are not intuitive."

5) "[a curb is no more a physical barrier than a stairwell is](https://github.com/gravitystorm/openstreetmap-carto/pull/3969#issuecomment-552425725)" @Prince-Kassad 

6) "[ I put some thought into this and I'm changing my vote to remove. The reasons are ...](https://github.com/gravitystorm/openstreetmap-carto/pull/3969#issuecomment-557777232)" @Adamant36 

Consensus-based development does not mean that you get a 1-vote veto on any change which you dislike.

imagico:
> removing is not even proposed in some last comments

No @kocio-pl - this is not how we have discussions here - reducing the view to a selection of latest comments and thereby re-iterating every discussion ad infinitum is not how we work.  We have had an extensive discussion here and in #3714.  There was plenty of opportunity for everyone to make different PRs to resolve #3714.  And everyone still would have the change for such with this PR merged of course.  

I agree with @jeisenbe that consensus based development here can only work if we all put the goals of the project above our personal preferences and accept decisions against those if we are not able to provide convincing arguments for our point of view.

kocio-pl:
> reducing the view to a selection of latest comments

IIRC reducing rendering was exactly what you have supported last time I checked. 

> There was plenty of opportunity for everyone to make different PRs to resolve #3714. 

And this was one of the failed attempt. We have a lot of unresolved problems, so I see nothing special about this one.

Prince-Kassad:
Abusing the "request changes" feature just to prevent a pull request one doesn't like from getting merged is something that I would report to @gravitystorm or perhaps even GitHub admins.

kocio-pl:
You are referring to discussions that did not yet even ended, let alone ended with finding common ground, so it lacks some backing and I get it as your personal view. I plan to show what fundamental problems are with removing features, which make it exactly counterproductive when looking at the goals of this style. (There are cases where it makes perfect sense, but this is not the one.)

I also don't agree with "abusing". Being clear is what is needed to decide.

pnorman:
Reviewing this issue, there is a consensus about merging it. @kocio-pl disagrees with removing it, but a consensus does not require everyone agree. My reason for not merging it at this time is primarily that I have not tested the code myself.

Also, for the record, a requested changes review is entirely appropriate here - it makes the status clear.

tordans:
Hey guys, I now understand why I don't see my changes on the map anymore :-). The rendering for barrier=kerb was very helpful to us in Berlin where we use the tag for small structures on roads but also to show where a block of houses is. It helps to show where the sidewalk (…) ends and the car-road starts. See https://overpass-turbo.eu/s/QzA for more cases.

I hope there will be a new rendering in the future. In the meantime, we will have to review our tagging :-).

nickvet419:
I am hope for a new rendering also. I had also been using them in downtown Chicago to mark the many curbed areas in the city, including the road medians that separate the direction of travel and create a protected island in crosswalks.  Without the renderings it creates a lack of detail.

Adamant36:
@nickvet419 There's a couple of open issuses about rendering traffic islands. Maybe you comment on those. Mapping curbs shouldn't be a substitute for using the proper tags for those things just because it rendered anyway. Its better we get traffic islands rendered instead.

jaffroy62:
I can not understand why you removed the rendering of barrier=kerb, also the discussion on removing barrier=embankment from the style does not make sense for me. 

This is not an improvement but rather a step back in my eyes. Why don't come up with a better suggestion but rather remove it?

Adamant36:
@jaffroy62, did you look through the discussion? Unfortunately there wasn't a usable way to make them look different then regularly fences. The amount of them mapped doesn't really matter if them being rendered potentially leads to confusion. Perhaps someone can come up with a usable solution at some point though so they can be rendered again.

nickvet419:
A solution would be to change the way fences look and maybe lighten the
color of the kerb line.

Adamant36:
How do you think fences should look instead? The problem with changing how they look is that since they look the same as curbs did  we are probably confined to the same solutions that already didnt work for them. There's really only so many ways you can render somthing like until its not intuative anymore and most ways lines can be rendered are already done with other thigs. Not that a lot of them would work for showing fence or curbs anyway.

kocio-pl:
It seems to be enough in #4063 to show the lack of a fence by making the line lighter. Just that, no changes in fence rendering are proposed there.

nickvet419:
I would go with something like this to represent fencing to solve many issues relating to fences.  
New Bitmap Image   There are a few issues this would fix to help differ them between quarry outlines, kerbs, walls etc...  I would also make walls a little thicker for the same reason.

jaffroy62:
> @jaffroy62, did you look through the discussion? Unfortunately there wasn't a usable way to make them look different then regularly fences....

@Adamant36 Thank You for the reply. Yes i did look through the discussion(s) of course but not in close detail. But i have already recocnized the problem of the wish of different apearances. From my perspective the logical task would have been to resolve the ambiguities by changing the appearances rather then remove them.

As You probably know, the apearance of some other barriers is ambiguous too (see: https://wiki.openstreetmap.org/wiki/Key:barrier) 

For me it yould be for example also a good suggstion to use dotted or dashed lines for some of them. I think this alsready has been suggested elsewhere in this discussion. Different colors could also be an option.

At least i am very happy about the movement that has come now again into this issue.

Greetings!

Adamant36:
@nickvet419, looks like a path or railroad track to me. Walls could maybe be a little thicker. I think there was an issue about that. Quarry outlines are being dealt with through lightening them I think. Different colors might work, but most colors are already taken and there's a general hesitation with the style to reuse colors for different elements. 

jaffroy62, yep. I agree some other barriers are also ambagious, but you have to start somewhere and this had the advantage of not really being a barrier in the first place. Maybe when vector tiles come around, which should allow for different map layers, we can have a barrier layer.

swedneck:
I agree with @nickvet419 that making it lighter would help, maybe it could also have triangles (or some other marker) like in JOSM/iD?
image

kocio-pl:
Clever observation! Since this is a new idea and this code is already merged, so it makes no sense to continue discussion here, could you please open a fresh ticket for this?

Issue #640, opened 2014-06-16
Pipeline should be visible


Cetusek:
Why are pipelines not rendered?
They are good landmarks.

kocio-pl:
> Maybe dark gray line, wider than barriers? It is a special kind of barrier, so it should be OK to have it similar. With bridge casings for elevated ones (with bridge=*).

Sounds good for me. 

I'm also thinking about segmented line, resembling pipe joinings (rough ASCII sketch):

`|——||——||——||——||——|`

javbw:
> Maybe With bridge casings for elevated ones (with bridge=*).
> 
> I'm also thinking about segmented line, resembling pipe joinings (rough ASCII sketch):
> 
> |――||――||――||――||――|
> 
Where I am there are a lot of "pipe bridges" - a buried 1m pipe pops out of the ground, crosses a river, and goes back underground. 

Any render that would help people not confuse them with a pedestrian crossable bridge would be wonderful - as they are often near or go over other ways. I know that the routing engines would ignore them, but visually would look like a bridge a person could cross (but are gated off and very dangerous). 

Javbw

kocio-pl:
It looks like Mapnik currently offers only simple line endings - and of course CartoCSS too:

https://github.com/mapnik/mapnik/wiki/LineSymbolizer#linesymbolizer
https://www.w3.org/TR/SVG/painting.html#StrokeLinecapProperty
https://carto.com/docs/carto-engine/cartocss/properties#line-cap-keyword

kocio-pl:
I have finally found a way to do it:

![screenshot-2018-5-2 openstreetmap carto kosmtik](https://user-images.githubusercontent.com/5439713/39498616-60569230-4da9-11e8-8027-4438121e9a8b.png)

The trick was to use image (`LinePatternSymbolizer` in Mapnik), the same as in the cliff rendering - above rendering is done simply with replacing the [SVG file](https://gist.github.com/kocio-pl/5db38f80ca0db92b60d22d829486546b#file-pipeline-svg). The only problem was that segments had no space between them, despite the space on both ends. However adding invisible blocks (alpha=0) helped.

Now we can get back to design. What color should be used? What segment length (currently 13 px total, with endings x 1 px), endings size (5x1 px) and the space between segments (1+1=2 px)?

Tomasz-W:
Surely it's too strong at the moment. It also is too sililar to aerialway=* rendering, eg.
[...]

PR #3973, opened 2019-11-15
Import man_made=pipeline, man_made=cutline and power=cable as linestrings

jeisenbe:
This PR adds each of the 3 features as local linestring_values in openstreetmap-carto.lua:
- import `man_made=pipeline` closed ways as linestrings
- import `man_made=cutline` closed ways as linestrings
- import `power=cable` closed ways as linestrings

### Explanation:
- All closed way features (ways which start and end on the same node and do not self-intersect) which are tagged with `man_made` or `power` are normally imported as polygons when added to the rendering database, as are linear `man_made` features like embankments and breakwaters.
- But there are exceptions for features that are only mapped as lines: currently `power=line` and `power=minor_line`, but `power=cable` is missing from this list of linestring exceptions.
- The tag `man_made=cutline` is also rendered and is only mapped as a line, so it should not be imported as a polygon
- We do not yet render `man_made=pipeline`, but it is only mapped as a line, and there is a plan to add rendering soon, see #640 
- Importing these features as polygons takes more server resources, so performance during import is slightly improved by this change.

This PR will be applied to the schema_changes branch, not to `master`, because it requires a rendering database reload at the time the release is applied to the rendering servers.

### DIFF
--- a/openstreetmap-carto.lua
+++ b/openstreetmap-carto.lua
@@ -37,7 +37,7 @@ local polygon_keys = {
 local linestring_values = {
     historic = {citywalls = true},
     leisure = {track = true, slipway = true},
-    man_made = {breakwater = true, cutline = true, embankment = true, groyne = true},
+    man_made = {breakwater = true, cutline = true, embankment = true, groyne = true, pipeline = true},
     natural = {cliff = true, tree_row = true, ridge = true, arete = true},
     power = {cable = true, line = true, minor_line = true},
     tourism = {yes = true},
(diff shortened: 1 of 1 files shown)


jeisenbe:
Anyone available to review this?

Adamant36:
For what it's worth it looks fine to me. As you mention it would be useful for #640 and the minor improvement on the server side is good. Plus, it would be cool if everything that needed to be for the reload got done at least sometime this year. So, if it where me I'd merge it. 

Its been here quite a while without objectors to. Which is a positive for it IMO. If there was anything detrimental about it you'd think someone would have commented by now. I seem to remember someone saying at one point that not commenting on things was a tacit form of approval.

jeisenbe:
I've tested this in a few medium extracts: Oregon (USA), Drenth (Netherlands), and Shikoku island (Japan). The import times were approximately the same before and after, though there is 20% variation between different trials so only a larger change would be noticable. The import and renderings were successful and appeared unchanged.

Issue #3594, opened 2018-12-24
healthcare=laboratory has no icon just a dot


532910:
Bad:
![bad](https://user-images.githubusercontent.com/14617699/50407653-ee1c0200-07ec-11e9-8f36-d0fdb38f79b8.png)

Good:
![good](https://user-images.githubusercontent.com/14617699/50407654-eeb49880-07ec-11e9-86ed-44c7850be594.png)

polarbearing:
If you consider this style bad, why do you use it?

Tomasz-W:
As we have a lot of things to show in OSM-Carto, we have to show some of them in simplest and less prominent possible way (e.g. by dots for healtcare and office objects or squares for public transport stops). If we would use dedicated icons for all features, map propably would look like this:

![](https://mhscattalk.com/wp-content/uploads/2017/05/791px-Plain-MMs-Pile.jpg)

matkoniecz:
I changed the title to make this issue fixable. If some other tags should be rendered with icon - open an issue for them.

I considered closing this as `declined` and opening new one for just laboratory, but this version should be less confusing.

thep:
> healthcare=laboratory has over a thousand uses and it could be something like a beaker (the container. Not the Sesame Street character). So that's a possibility at least.

I asked medical lab people and they said a microscope should represent their work best. A beaker or a flask could still be read as a chemical lab instead.

Adamant36:
A side view of a microscope probably wouldn't be to hard of an icon to make. It would probably be easier to tell its a lab 
icon compared to a beaker also.

jeisenbe:
Note that we cannot render anything tagged under the `healthcare=*` key until the next database reload. This could happen sooner rather than later, but someone will need to submit a PR which adds `healthcare=*` as a polygon key in the `openstreetmap-carto.lua` file. See #3611

Adamant36:
> Note that we cannot render anything tagged under the healthcare=* key until the next database reload. This could happen sooner rather than later, but someone will need to submit a PR which adds healthcare=* as a polygon key in the openstreetmap-carto.lua file.

Hopefully done correctly in PR #4005

imagico:
This issue in the current form is invalidated by #3731.  We should either close it or the issue opener should re-formulate it to adding rendering of healthcare=laboratory, see also #4402.

jeisenbe:
Closing since the original concern is no longer valid

PR #4005, opened 2020-01-17
Import healthcare= as a polygon feature

Adamant36:
Related to #3594 and #3611

Import the tag healthcare as a polygon. Which allows rendering the healthcare tag when mapped as areas. As allowed by the wiki documentation. The PR should be merged to the schema_changes branch and then merged with it into the master branch when the next database reload is planned. I think I set it up probably for that.


jeisenbe:
Next time, make sure to fetch and check out the `schema_changes` branch, rather than the `master` branch, prior to making the new branch for the PR.

polarbearing:
This is just about importing, not about rendering any of them specifically?

jeisenbe:
I am in favor of this. While I would not necessarily support rendering the values which duplicate common tags (e.g. healthcare=hospital vs amenity=hospital), there are a number of significant features which are not currently found except under healthcare values, like healthcare=laboratory, healthcare=midwife, healthcare=birthing_centre, etc. 

It will also help users who adapt this style for their own purposes if this key is already available.

polarbearing:
No problem with ```healthcare``` in general, my concern is just that we should only render a whitelist of values and not a catch-all.

jeisenbe:
>
> No problem with healthcare in general, my concern is just that we should
> only render a whitelist of values and not a catch-all.
>

Right, agreed.

This is just about importing, not about rendering any of them specifically?


Yes, rendering would have to wait until after the Openstreetmap.org
rendering servers can be reloaded. This PR only adds “healthcare=“ to the
list of keys which result in a closed way being imported as a polygon into
our rendering database.

imagico:
In favor of the schema change (@jeisenbe already mentioned that there are some technical issues with using the wrong base branch) - but just to be clear: This facilitates the possibility to render healthcare features, it does not mean an automatism to actually implement this.  Discussion on if and how this can be done while maintaining constructive mapper feedback and not supporting tagging proliferation needs to happen if and when this should be implemented.

Adamant36:
> Next time, make sure to fetch and check out the schema_changes branch, rather than the master branch, prior to making the new branch for the PR.

So what's the process for that? When you fork and clone a repository it only does it under the master branch. In this case in the terminal when doing git branch or git checkout whatever it didn't show schema_changes as a branch. When I initiated the PR on the websites side though, I selected the schema_changes branch as the thing I was applying the PR to in the comparing changes drop down menu thing. So I figured that would do it.

jeisenbe:
> I selected the schema_changes branch as the thing I was applying the PR to

Yes, that was important. But you also should start the branch by checking out schema_changes from "upstream" (or whatever you have named the remote https://github.com/gravitystorm/openstreetmap-carto.git on your system).

To do this, first enter in Terminal / Command Line:
```
git fetch upstream schema_changes
```
This will fetch the branch "schema_changes" from your remote named "upstream" which should be pointing to https://github.com/gravitystorm/openstreetmap-carto.git

(If you are not sure about this, use ```git remote -v``` to list the name and url of each remote, first)

Next:
```
git checkout -b schema_changes
```

Now you will have a new local branch named `schema_changes` which is the same as the `schema_changes` branch in this repository. 

Finally:

```
git checkout -b <new_PR_branch>
```
This will create a new branch, named "new_PR_branch", with the same commits as `schema_changes`. 

(I made this same mistake the first couple of times, no worries)

jeisenbe:
Silly me, I forgot that we already added `healthcare` to `schema_changes` in PR #3784 

(Previously when I looked at the list, I thought it had not been accepted because it has the red "closed without merging" icon, but that is because it is not yet merged to master)

Closing because this duplicates PR #3784

Adamant36:
No worries. At least now there's an explanation of how to the branch thing correctly for next time if nothing else. As I'll probably do PRs for some of the other ones at some point.

Adamant36:
@jeisenbe, if I follow the last two steps `git checkout -b schema_changes` and then do `git checkout -b <new_PR_branch>` wont that just switch to the new PR branch making the creation of the schema_changes branch useless because I won't be working from it anymore?

PR #4010, opened 2020-01-26
Remove rendering of barrier=embankment

### LINKED ISSUE #3744: Stop rendering barrier=embankment

jeisenbe:
Closes #3744

- Remove the rendering of barrier=embankment
- Change minzoom for barrier layer to z15 since this is now the earliest that any `barriers` feature is shown (city walls).

### Explanation:
- The tag `barrier=embankment` was once somewhat used, but never was documented in the wiki. Instead `man_made=embankment` or `man_made=dyke` are the much more common ways to make one-sided embankments and levees. The definition of `barrier=embankment` was not clear and it is not similar to the other barriers rendered in the same way (fences and walls), so it's best to no longer support this tag. (See #3744)

### Test rendering:
(Note: these images also show changes proposed in #3969)
(Test shows fence, kerb, wall, embankment from top to bottom)
Before
kerb-embankment-test-before

After
kerb-embankment-barrier-after

### DIFF
--- a/landcover.mss
+++ b/landcover.mss
@@ -850,10 +850,6 @@
     line-width: 0.4;
     line-color: #444;
   }
-  [feature = 'barrier_embankment'][zoom >= 14] {
-    line-width: 0.4;
-    line-color: #444;
-  }
   [feature = 'barrier_hedge'][zoom >= 16] {
     line-width: 1.5;
     line-color: @hedge;
--- a/project.mml
+++ b/project.mml
@@ -583,7 +583,7 @@ Layer:
             way, COALESCE(historic, barrier) AS feature
           FROM
             (SELECT way,
-              ('barrier_' || (CASE WHEN barrier IN ('chain', 'city_wall', 'embankment', 'ditch', 'fence', 'guard_rail',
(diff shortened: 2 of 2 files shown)


jeisenbe:
I believe this has already been approved as part of #3969 previously, but I will wait for another maintainer to formally approve this PR prior to merging. 

There is one change: I've also adjusted the initial zoom level for the barriers layer from z14 to z15, since only `barrier=embankment` was shown at z14 previously, and now z15 is the earliest (for city walls).

Issue #2288, opened 2016-08-18
render natural=gully or other tag similar using current natural=cliff style


d1g:
natural=cliff was defined at wiki with close relation to natural rocks, but  natural=cliff was heavily tagged for rendering: 
http://www.openstreetmap.org/way/191570269
http://www.openstreetmap.org/way/191570270

By many users, everywhere. Personally I relate this because of the symbols in "карты генштаба" maps.

We got new tag(s) natural=gully/valley to indicate a non-rocky change in the elevation.

mboeringa:
A cliff does not necessarily have to be solid rock. There are many cases of soft sediment erosion along coasts, with pretty giant and high "cliffs" being formed that would definitely not classify as gully or valley, e.g. along the coast of the UK. I think the usage you are showing is within the limits of acceptable tagging for this. Alternatively, if the "gully" is actually a man_made=cutting, using the man_made=embankment tag would be an option.

dieterdreist:
sent from a phone

> Il giorno 18 ago 2016, alle ore 11:14, mboeringa <notifications@github.com> ha scritto:
> 
> Alternatively, if the "gully" is actually a man_made=cutting, using the man_made=embankment tag would be an option.


??? Why the embankment and not the cutting tag if it's a cutting?

d1g:
> A cliff does not necessarily have to be solid rock. 

Agree. natural=cliff is about "almost vertical drop" (see wiki). Almost always you get them with solid formations (simply put, "rocks"). But with coastal erosion you can get natural=cliff (a vertical drop) out of everything, yes.

> Alternatively, if the "gully" is actually a man_made=cutting, using the man_made=embankment tag would be an option.

Gullies and valleys are different from "vertical drop" because... they are not always as steep as "cliffs", right?
Gullies and valleys are different from man_made=embankment because they are not man made, right?

http://taginfo.openstreetmap.org/tags/man_made=cutting has only 60 uses ATM and wasn't described at wiki.

I don't see "almost vertical drops" based on Bing/other imagery here:
http://www.openstreetmap.org/way/173655182
http://www.openstreetmap.org/way/368611288
http://www.openstreetmap.org/way/368623276

Yes, they are steep, but not "near vertical" (natural=cliff). They are natural (not man_made=embankment) and most likely natural=gully or natural=valley (see wiki for details)

d1g:
[...]

PR #4012, opened 2020-01-28
Import natural=earth_bank as a linestring only

jeisenbe:
Related to  #3611 and #2288

- Add natural=earth_bank to the list of linestring exceptions in openstreetmap-carto.lua so that closed ways with this tag will not be imported as polygons, but as a linestring only.
- This tag is only used on lines, similarly to natural=cliff or man_made=embankment
- While it is not currently rendered, there are over 4,000 features with this tag, including 202 closed ways (where the bank encircles an area) which are not intended to represent areas. http://overpass-turbo.eu/s/Qa3
- Adding this to the lua transformations file will make it easier to render this feature correctly in the future or in other map styles which are based off of this style.

### DIFF
--- a/openstreetmap-carto.lua
+++ b/openstreetmap-carto.lua
@@ -36,7 +36,7 @@ local linestring_values = {
     historic = {citywalls = true},
     leisure = {track = true, slipway = true},
     man_made = {embankment = true, breakwater = true, groyne = true},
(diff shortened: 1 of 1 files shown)


jeisenbe:
Oops, should be merged to schema_changes not to master

PR #4017, opened 2020-01-30
Import the key "golf=" as polygon features, with exceptions for golf=hole, path and cartpath

jeisenbe:
Related to #3611 and #661

- Add `golf` key to list of polygon keys in openstreetmap-carto.lua so that closed ways with this key will be imported as polygons, except for `golf=hole`, `golf=path` and `golf=cartpath`, which are always mapped as lines

While there is some debate in #661 whether rendering golf=* features in this style is appropriate, some other styles based on this one are already rendering golf=* features. Adding this key to properly import the relevant features will make it easier to adapt this stylesheet in the future.

See documentation of `golf=*` values at https://wiki.openstreetmap.org/wiki/Key:golf


jeisenbe:
Fixed the merge conflict

PR #4018, opened 2020-01-31
Switch Travis to Xenial

pnorman:
This partially replaces #3816

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,5 @@
 language: node_js
-dist: trusty
+dist: xenial
 sudo: false
 node_js:
   - "6"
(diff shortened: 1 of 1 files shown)


pnorman:
Still needs a review.

jeisenbe:
The PR just changes "Trusty" to "Xenial" and "postgresql-9.4-postgis-2.3" to "postgresql-9.4-postgis-2.4". According to the Travis documentation it is correct for us to update to Xenial, and I see that postgresql-9.4-postgis-2.4 has been available since mid 2018 so these changes seem appropriate. 

But I'm not certain how to test these changes.

pnorman:
The test is that Travis runs successfully ;) For a review like this, all that's needed is verifying the code and that Travis passed.

PR #4025, opened 2020-02-06
Don't form areas from tags like building=no

pnorman:
ref. https://github.com/gravitystorm/openstreetmap-carto/issues/3611#issuecomment-582672393

Checked by loading an extract and verifying that building=no with other polygon tags are still areas, and then loading just a building=no way


jeisenbe:
Tests with highway=pedestrian closed ways before:
export-9

And after:
export-11

PR #4032, opened 2020-02-13
Change area decisions at import time

### LINKED ISSUE #3611: Different geometry decisions (point/line/polygon) on database re-import

pnorman:
Fixes #3611


pnorman:
Originally I was wondering if this needed a benchmarking run, but looking at the diff I don't see anything with performance implications, either at import time or at run time.

pnorman:
I wasn't expecting that much of a performance difference, so I'm running a benchmark myself

pnorman:
I don't find a noticeable performance difference on Geofabrik's Canada extract with `--slim --drop --flat-nodes`.

jeisenbe:
That's great. Is there documentation about how to use these options (`--slim --drop --flat-nodes `) when importing with docker?

pnorman:
> Is there documentation about how to use these options (`--slim --drop --flat-nodes `) when importing with docker?

I don't use docker, so don't know.

PR #4033, opened 2020-02-13
Upgrade get-shapefiles to python3

pnorman:


StyXman:
I tested it with py3.5 and py3.8. I think it shoudld be good enough.

Issue #3982, opened 2019-11-27
Label of large protected areas never shown


jeisenbe:
### Expected behavior
- We should show the label of (mercator)-large protected areas at z8

### Actual behavior
- If the protected area is too large the label is never shown, due to the max 768,000 way_pixels limit, which applies even at z8, the first zoom levels where protected areas are shown
- This mainly affects Alaska, Northern Canada and Siberia

### Links and screenshots illustrating the problem

https://www.openstreetmap.org/relation/8537673#map=8/68.876/-144.456
![z6-anwr-relation](https://user-images.githubusercontent.com/42757252/69758788-5538e980-1115-11ea-931b-7dcfb3c6bba2.png)

z8 Artic National Wildlife Refugee - no central label shown
![z8-alaska-wildlife-refugee](https://user-images.githubusercontent.com/42757252/69758791-566a1680-1115-11ea-9f53-64af31a14c72.png)

jeisenbe:
Possible solutions:

1) Show protected areas and labels even sooner. The current initial level of 8 is already quite early, but we now show landcover from z5 so this could be changed - but only if issues with low-zoom landcover are also addressed.

2) Do not apply the way_pixels limit until z9 or z10 - this would be an easy solution.

jeisenbe:
Any comments on solution 2: "Do not apply the way_pixels limit until z10" which would mean that the central label of protected areas, lakes, large forests, and military areas would still be shown at z8 and z9 even if the area takes up the whole screen.

kennykb:
I don't know enough about the internals to make sense of what 'way_pixels' actually means, to evaluate the proposed solution 2.

My favorite large protected area is https://www.openstreetmap.org/relation/1695394

1. It's not shown at z7, but it could be - it's bigger than some of the neighboring states and wouldn't really collide with anything. Even at z6, it could probably be labeled without confusion. (I don't consider these very low zoom levels to be terribly important, so this is probably not worth doing anything about - but if the proposed solution 1  addresses this, that would be nice.)
2.  At z8 and z9, it looks reasonable.
3. At z10, the interior label disappears, so at z10-z12 all that is shown is an external border without labels. I think this is the behaviour that relates to the way_pixels limit?
4. At z13 and beyond, the name is used to label the external border. Probably can't do much better than this.

jeisenbe:
[...]

PR #4034, opened 2020-02-16
Adding upper way_pixels limit in SQL for amenity-points and text-poly-low-zoom layers

### LINKED ISSUE #4030: Add conditions and functional indexes on ST_PointOnSurface(way)
A typical layer with points from polygons looks like
```sql
(SELECT
    way,
    "natural"
  FROM
  (SELECT
      "natural"
    FROM planet_osm_polygon
    WHERE way && !bbox!
  UNION ALL
  SELECT
      way,
      "natural"
    FROM planet_osm_point
    WHERE way && !bbox!
    ) _
  WHERE "natural" IN ('spring')
) AS springs
```

Adding a functional index with `CREATE INDEX ON planet_osm_polygon USING GIST (ST_PointOnSurface(way))` and adding `ST_PointOnSurface(way) && !bbox!` to the polygon branch of the union all allows a query plan which avoids calculating ST_PointOnSurface for polygons where the point isn't within the metatile. This doesn't matter much for some layers like buildings, addresses, or springs, but it can be a savings for layers with lots of polygons that bounding box intersect the current MT but don't have the label point within the bbox.

On the other hand, these polygons are by definition large in relation to the current screen size, and we tend to exclude those with area filters (#1028). Perhaps we could get the same or greater gains by moving the area filters from MSS to SQL.

imagico:
Related to #4030.

Requires testing.

Implemented this at the [hack weekend](https://wiki.openstreetmap.org/wiki/Karlsruhe_Hack_Weekend_February_2020).

Changes contained:

* moving the upper way_pixels limit (< 768000) in amenity-points and text-poly-low-zoom to SQL and extending it to all polygon features.  This violates the principle not to use way_area filtering on features without visual feedback on the geometry but for a *high upper limit* this is OK i think.
* move remaining way_area rules in amenity-points.mss to way_pixels based rules.  This is a change with visual impact so requires visual evaluation.
* <strike>adding functional index on ST_PointOnSurface(way) on the polygons layer and adding rule to use it in amenity-points.  This calls for being tested on a full planet database.</strike>

### DIFF
--- a/project.mml
+++ b/project.mml
@@ -1590,6 +1590,7 @@ Layer:
                   way_area
                 FROM planet_osm_polygon
                 WHERE way && !bbox!
+                  AND ST_PointOnSurface(way) && !bbox!
               UNION ALL
               SELECT
                   way,
--- a/project.mml
+++ b/project.mml
@@ -1590,6 +1590,7 @@ Layer:
                   way_area
                 FROM planet_osm_polygon
                 WHERE way && !bbox!
+                  AND way_area < 768000*POW(!scale_denominator!*0.001*0.28,2)
               UNION ALL
               SELECT
                   way,
@@ -1621,7 +1622,6 @@ Layer:
               ) _
             ) AS features
           WHERE feature IS NOT NULL
-            AND (way_pixels < 768000 OR way_pixels IS NULL)
           ORDER BY score DESC NULLS LAST,
             way_pixels DESC NULLS LAST
           ) AS amenity_points
@@ -1990,7 +1990,7 @@ Layer:
               OR leisure IN ('nature_reserve'))
             AND building IS NULL
             AND name IS NOT NULL
-            AND way_area/NULLIF(POW(!scale_denominator!*0.001*0.28,2),0) < 768000
+            AND way_area < 768000*POW(!scale_denominator!*0.001*0.28,2)
           ORDER BY way_area DESC
         ) AS text_poly_low_zoom
     properties:
--- a/project.mml
+++ b/project.mml
@@ -1991,6 +1991,7 @@ Layer:
               OR leisure IN ('nature_reserve'))
             AND building IS NULL
             AND name IS NOT NULL
+            AND way_area/NULLIF(POW(!scale_denominator!*0.001*0.28,2),0) < 768000
           ORDER BY way_area DESC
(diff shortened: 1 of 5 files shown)


imagico:
I just realized that this:

```
                  AND ST_PointOnSurface(way) && !bbox!
```
probably will cause a significant number of metatile boundary inconsistencies because if a label location is just outside the metatile but it reaches in it won't be rendered.  Since large polygon labels tend to be large in size this would be more of a problem than with current point labels (where this in principle already happens).

The previous `way && !bbox!` had the advantage that unless the label size exceeded the polygon bounding box (which can of course occur in rare cases) this did not happen.

jeisenbe:
Also consider issue #3982 - if an area is large enought that it is over 768,000 pixels in area at the first zoom level when the text label would be rendered, then it will never be shown. I had been testing a solution to this issue which would remove the way_pixels based filtering for the first zoom level (z8) for the features where this can be a problem (mainly large national parks, nature reserves and protected areas in the artic)

pnorman:
I think we should split this up, because the SAT_PointOnSurface changes require a different type of review than the way_pixels changes

imagico:
Removed the functional index addition - so this is just the way_pixels upper limit now.

@jeisenbe - i think this is a problem that requires a different solution.  Starting to show a label when a feature is already filling the whole screen is never going to work well.

project.mml
```diff
@@ -1621,6 +1621,7 @@ Layer:
               ) _
             ) AS features
           WHERE feature IS NOT NULL
+            AND (way_pixels < 768000 OR way_pixels IS NULL)
```
pnorman: Can you move this to inside the subselect, applying only to the polygon table and dropping the point case.

I'm not sure the planner will evaluate what amounts to `COALESCE(way_area/NULLIF(POW(!scale_denominator!*0.001*0.28,2),0), 0) < 768000` before st_pointonsurface.

Where we have a way_area condition, we have way_area > 1*!pixel_width!::real*!pixel_height!::real. iirc, this doesn't work work properly with different screen resolutions and scale factors, so I think this would work: `way_area < 768000 * POW(!scale_denominator!*0.001*0.28,2)`

imagico: This is changed now - and also unified in amenity-points and text-poly-low-zoom.

Issue #3273, opened 2018-06-20
Don't filter riverbanks by size


eigenwillig:
Has somebody already proposed some code to avoid unmanageable big riverbanks multipolygons (see [imagico's post](Say hello to the giant Multipolygons
))?

If not what about:
```
  [natural = 'water']::natural,
  [landuse = 'reservoir']::landuse {
    [zoom >= 0][zoom < 1][way_pixels >= 4],
    [zoom >= 1][zoom < 2][way_pixels >= 16],
    [zoom >= 2][zoom < 8][way_pixels >= 32],
    [zoom >= 8] {
      [int_intermittent = 'no'] {
        polygon-fill: @water-color;
        [way_pixels >= 4] {
          polygon-gamma: 0.75;
        }
        [way_pixels >= 64] {
          polygon-gamma: 0.6;
        }
      }
      [int_intermittent = 'yes'] {
        polygon-pattern-file: url('symbols/intermittent_water.png');
        [way_pixels >= 4] {
          polygon-pattern-gamma: 0.75;
        }
        [way_pixels >= 64] {
          polygon-pattern-gamma: 0.6;
        }
      }
    }
  }
}

  [waterway = 'riverbank']::waterway {
    [zoom >= 0] {
      [int_intermittent = 'no'] {
        polygon-fill: @water-color;
        [way_pixels >= 4] {
          polygon-gamma: 0.75;
        }
        [way_pixels >= 64] {
          polygon-gamma: 0.6;
        }
      }
      [int_intermittent = 'yes'] {
        polygon-pattern-file: url('symbols/intermittent_water.png');
        [way_pixels >= 4] {
          polygon-pattern-gamma: 0.75;
        }
        [way_pixels >= 64] {
          polygon-pattern-gamma: 0.6;
        }
      }
    }
  }
}
```

kocio-pl:
Please tell more about this change in simple words, the code alone doesn't speak for itself.

eigenwillig:
@kocio-pl This is the code [before](https://github.com/gravitystorm/openstreetmap-carto/blob/master/water.mss):

```
  [natural = 'water']::natural,
  [landuse = 'reservoir']::landuse,
  [waterway = 'riverbank']::waterway {
    [zoom >= 0][zoom < 1][way_pixels >= 4],
    [zoom >= 1][zoom < 2][way_pixels >= 16],
    [zoom >= 2][zoom < 8][way_pixels >= 32],
```

[...]

PR #4042, opened 2020-02-26
way_area conditions for all ST_PointOnSurface layers

### LINKED ISSUE #4030: Add conditions and functional indexes on ST_PointOnSurface(way)

pnorman:
Fixes #4030 

I used 768000 for POIs, which is about 12 tiles, and 4000000, which is about 64 tiles for cases like countries where we want to show larger objects.

### DIFF
--- a/project.mml
+++ b/project.mml
@@ -357,6 +357,7 @@ Layer:
               "natural"
             FROM planet_osm_polygon
             WHERE way && !bbox!
+            AND way_area < 768000*POW(!scale_denominator!*0.001*0.28,2)
           UNION ALL
           SELECT
               way,
@@ -1185,6 +1186,7 @@ Layer:
             AND admin_level = '2'
             AND name IS NOT NULL
             AND way_area > 100*POW(!scale_denominator!*0.001*0.28,2)
+            AND way_area < 4000000*POW(!scale_denominator!*0.001*0.28,2)
             AND osm_id < 0
           ORDER BY way_area DESC
         ) AS country_names
@@ -1228,7 +1230,8 @@ Layer:
             AND boundary = 'administrative'
             AND admin_level = '4'
             AND name IS NOT NULL
-            AND way_area > 100*POW(!scale_denominator!*0.001*0.28,2)
+            AND way_area > 3000*POW(!scale_denominator!*0.001*0.28,2)
+            AND way_area < 4000000*POW(!scale_denominator!*0.001*0.28,2)
             AND osm_id < 0
           ORDER BY way_area DESC
         ) AS state_names
@@ -1336,6 +1339,7 @@ Layer:
               way_area
             FROM planet_osm_polygon
             WHERE way && !bbox!
+              AND way_area < 768000*POW(!scale_denominator!*0.001*0.28,2)
           UNION ALL
           SELECT
               way,
@@ -1389,6 +1393,7 @@ Layer:
           FROM planet_osm_polygon
           WHERE way && !bbox!
             AND junction = 'yes'
+            AND way_area < 768000*POW(!scale_denominator!*0.001*0.28,2)
           ORDER BY way_pixels DESC NULLS LAST
         ) AS junctions
     properties:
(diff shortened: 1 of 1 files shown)


imagico:
I am strictly against introducing a lower (starting) way_area limit for features where there is no visual feedback on the geometry - in this case place=square.

Regarding changing `way_area > 1*!pixel_width!::real*!pixel_height!::real` - i am not sure this is a good idea.  This limit had been introduced as arbitrary limit based on the cargo culted assumption that the threshold of one pixel in rendering has somehow a magic significance.  Promoting this to a scale independent limit would make it even more baseless.  What should be done here is to get rid of this arbitrary and distorting limit in the first place - a matter that i have discussed at length in the past (see #3273 for the most recent collection of references).

In the single case of the `0.01*!pixel_width!::real*!pixel_height!::real` - that should definitely stay tied to the rendering resolution because the reasoning behind that limit is connected to the actual pixel size of the rendering, see also #1566.

pnorman:
> I am strictly against introducing a lower (starting) way_area limit for features where there is no visual feedback on the geometry - in this case place=square.

I am not introducing it - it already exists: https://github.com/gravitystorm/openstreetmap-carto/blob/e7de1cd7c48d4efbb801f0b33fd7cc0dea143c36/roads.mss#L3307-L3309

For the other limits, I believe it is important that the feature selection and queries are the same independent of scale factor. If you want to get rid of the limit, that belongs in a different PR.

imagico:
> I am not introducing it - it already exists: 

Ah, sorry.  See #4043.

> For the other limits, I believe it is important that the feature selection and queries are the same independent of scale factor.

Well - as explained the reasoning behind these limits is not independent of the scale factor.

project.mml
```diff
@@ -133,7 +133,7 @@ Layer:
               OR tourism IN ('camp_site', 'caravan_site', 'picnic_site')
               OR highway IN ('services', 'rest_area')
               OR railway = 'station')
-              AND way_area > 1*!pixel_width!::real*!pixel_height!::real
+              AND way_area > 1*POW(!scale_denominator!*0.001*0.28,2)
```
jeisenbe: Let's leave this line the same. If rendering 3x retina tiles, there will be 9 real pixels per standard pixel, and print quality could be even higher. At these higher resolutions the smaller polygons will be more significant and should not be excluded.

project.mml
```diff
@@ -241,7 +241,7 @@ Layer:
               OR landuse IN ('reservoir', 'basin')
               OR "natural" IN ('water', 'glacier'))
             AND building IS NULL
-            AND way_area > 1*!pixel_width!::real*!pixel_height!::real
+            AND way_area > 1*POW(!scale_denominator!*0.001*0.28,2)
```
jeisenbe: Water areas should also use real pixel_width

project.mml
```diff
@@ -1184,7 +1185,8 @@ Layer:
             AND boundary = 'administrative'
             AND admin_level = '2'
             AND name IS NOT NULL
-            AND way_area > 100*!pixel_width!::real*!pixel_height!::real
+            AND way_area > 100*POW(!scale_denominator!*0.001*0.28,2)
```
jeisenbe: These changes and those following are fine for using scale_denominator, so that the display of these features will be consistent at different resolutions.

jeisenbe: These changes and those following are fine. Using scale_denominator for country_names, state_names, text etc makes sense, so that the display of these features will be consistent at different resolutions.

Issue #4002, opened 2020-01-11
Drop amenity=telephone from z17


Prince-Kassad:
Split from #1745

The amount of public telephones is rapidly decreasing both in developed and developing countries (not so much in OSM, unfortunately, because we can't keep up with the deconstruction of public telephones). Where public telephones still exist, they're usually not used for their telephone feature. For example, in Germany, public telephones in larger cities also function as a wifi hotspot, and most people will use the telephone for the wifi hotspot rather than its phone feature. Public telephones in small villages, where they still exist, are simple models that cannot be used for anything other than emergency calls.

See an example where current amenity=telephone clutters the map:

https://www.openstreetmap.org/#map=17/50.13321/8.91777&layers=N

It's debatable at which zoom level it is appropriate to render public telephones. I personally think z19 is too late (and think emergency=phone should be pushed back up to z18 for the same reason), but z17 is definitely too early, as you can see from the provided link.

imagico:
I don't think the idea of hiding the rendering of phones at lower zoom levels because there are many of these mapped which do not exist any more is a good approach.

In general my opinion on this is already stated in https://github.com/gravitystorm/openstreetmap-carto/pull/1884#issuecomment-144378212 - if the map is too full at z17 my approach would be to move other things to higher zoom levels like shops, restaurants etc.  Public phones as a possibility to call for help in a case of emergency and as one of the few anonymous communication possibilities that exists in many countries today are IMO an important infrastructure, *especially if they become more rare these days*.

kocio-pl:
Yet this is not an emergency map, it's a general map (if it was, I believe #3629 should be very visible). Also relying on importance as a primary feature for such general thing as zoom levels is very subjective and makes it very impossible to create any order of objects.

Adamant36:
Like 67 percent of the worlds population has cell phones now. So, I think an argument could be made that are mostly sorta historical artifacts at this point. Even in places where they are available I doubt they get used that much. Except for the anonymous reasons you state, but id guess even that is largely done on a cell phone now. 

Personally, I would put them in the same category as things like changing tables (except less so). Something important to some people, that would be good on a specialized map, but probably not great for a general map at this point. This issue isn't even about getting rid of them completely either. 

[...]

PR #4044, opened 2020-02-27
Move small amenities to z18

jeisenbe:
Related to #4002, #1745 and PR #1884

- This PR, similar to closed PR #1884, moves amenity=atm and amenity=post_box to z18, though in this case they are currently rendered only at z19, while previously they were rendered at z17. 
- Also emergency phones are moved back to z18 from z19. Previously they had been rendered at z17

This is an attempt to compromise. While these features were moved from z17 to z19 without consensus, rather than attempting to move them back to z17 we can use z18.

In rural areas these features should be shown at z17, while in some very large cities (e.g. Warsaw) there may be many icons in one screenshot at that zoom level due to the high density of features in capital cities.

### DIFF
--- a/style/amenity-points.mss
+++ b/style/amenity-points.mss
@@ -61,7 +61,7 @@
     }
   }
 
-  [feature = 'amenity_atm'][zoom >= 19] {
+  [feature = 'amenity_atm'][zoom >= 18] {
     marker-file: url('symbols/amenity/atm.svg');
     marker-fill: @amenity-brown;
     marker-clip: false;
@@ -664,7 +664,7 @@
     marker-clip: false;
   }
 
-  [feature = 'amenity_post_box'][zoom >= 19] {
+  [feature = 'amenity_post_box'][zoom >= 18] {
     marker-file: url('symbols/amenity/post_box.svg');
     marker-fill: @amenity-brown;
     marker-clip: false;
@@ -692,7 +692,7 @@
     marker-clip: false;
   }
 
-  [feature = 'emergency_phone'][zoom >= 19] {
+  [feature = 'emergency_phone'][zoom >= 18] {
     marker-file: url('symbols/amenity/emergency_phone.svg');
     marker-fill: @amenity-brown;
     marker-clip: false;


kocio-pl:
What rule would we have after that for choosing zoom levels? I have proposed size rule as rule of the thumb and all the other proposed rules were more subjective or not applicable for some types of objects.

jeisenbe:
In PR #1884 you proposed z18 for these features.

Is there a particular reason that you are no longer in favor of this?

> What rule would we have after that for choosing zoom levels? I have proposed size rule

The "size rule" was never accepted by any of the other maintainers of
this style, from what I have read.

Most initial zoom levels for features in this style have been
determined by the significance of a particular feature for general map
users at that scale. Large size might make a feature significant at
lower scales, but it is not the most important consideration.

Places of worship are shown at z16 not because they are larger than
other features, but because they are important orientation points at
that scale.

kocio-pl:
I have thought more about the size and would be more following it currently than then, because it's simple, universal and used already in this style (at different scales).

Usually places of worship are indeed larger (higher, which also applies to at least some peaks).

> Large size might make a feature significant at lower scales, but it is not the most important consideration.

Which leads me to a fundamental question about importance, which is - for whom it is important and specifically why do you think it applies only to low zoom?

imagico:
> This is an attempt to compromise. While these features were moved from z17 to z19 without consensus, rather than attempting to move them back to z17 we can use z18.

More specifically this change would be a partial revert of #2993, #3372 and #3445.

While i don't mind this change in principle if it helps moving us back towards consensus i am unsure if it actually does.

With these changes (as in particular became clear in the discussion on #3372) the massive schism in the fundamental approach to cartographic design that we struggle with became very visible to everyone i think.

I don't think ultimately trying to make compromises between two very different and fundamentally opposite strategies will lead us very far.  Yet it is good to try addressing this matter as one of the key manifestations for the loss of consensus on cartographic strategy we struggle with now.

Regarding the change itself and independent of the whole backstory - Essentially most has already been said in #1884 back in 2015.

jeisenbe:
Rebased for v5.0.0

Adamant36:
I had suggested a while back that z18 is kind of an under used zoom level that could be used sort of in this way, except more planned out. Just rendering random things there just to compromise though would cause it to be an unplanned out icon dump in a similar way to how z19 currently is. That said, I think z18 has some potential as a place for related icons that don't fit in well at z17 or z19, or something.

jeisenbe:
I will wait till after v5.0.0 to merge this.

PR #4047, opened 2020-03-03
Move MSS files to style directory

### LINKED ISSUE #4041: Move MSS files to their own directory
We now have about 50 items in the root directory of the project, which is becoming unwieldy. 17 of these are MSS files.

I propose moving all the MSS files to their own directory. I'm not sure if symbols should move to that directory too. I would keep project.mml where it is.

The only con to this is that most PRs would need rebasing, so we should try to get as many merged as are ready.

If we do this, I'd like to do it before we tag 5.0.0.

pnorman:
Fixes #4041

All style PRs will need rebasing with this move.
No rendering changes.


jeisenbe:
Do we need to change `generate_road_colours.py` as well?

Line3: `# Generates a set of highway colors to be stored in road-colors-generated.mss.`

Late:
`print "/*   ./scripts/generate_road_colours.py > road-colors-generated.mss  */"`

Also, should the mention of `fonts.mss` in INSTALL.md be changed to `style/fonts.mss` ?

pnorman:
> Line3: `# Generates a set of highway colors to be stored in road-colors-generated.mss.`

I'm keeping this line as-is intentionally, I feel it reads better using just the filename.

PR #4051, opened 2020-03-05
fix broken links in get-shapefiles.py

capan:
- Fixed broken links in get-shapefiles.py


pnorman:
You still have a revert of the removal of water_boundaries and the switch to water polygons.

pnorman:
We don't use the shapefile you've added.

PR #4052, opened 2020-03-06
Make link roads the same z_order as their base class

### LINKED ISSUE #3719: Rescale z_order and treat links the same as base type
Currently all orders are divisible by 10. The z_order would be easier to work with if all current orders were divided by 10.

At the same time, we should make highway=X and highway=X_link have the same z_order, then add link status to the order by in the queries.

pnorman:
By adding link status to the ORDER BY this then allows the ordering
logic to be changed at low zoom or in forks.

Tested locally at several complex interchanges and everything looks as it does on tile.osm.org.

Fixes #3719


jeisenbe:
This reminds me that we had intended to add construction=* to the z_order list to fix inconsistent ordering of highway=construction types, as mentioned in #3579 and the initial version of PR #3646. 

Would you be willing to add the `z_order` for different values of `construction=` to this PR, for example something like this commit: https://github.com/gravitystorm/openstreetmap-carto/pull/3646/commits/810448b808ef2dbec7d583bc2e31721bac16b12b ? Or else I can open a separate PR.

pnorman:
Let's keep them in different PRs

pnorman:
Thinking about this, it's not going to work with our peculiar ordering of links between high-class roads and low-class roads.

PR #4054, opened 2020-03-06
Change scree_overlay of scree and shingle to beach_coarse pattern

jragusa:
Related to #3957 (see [comment](https://github.com/gravitystorm/openstreetmap-carto/pull/3957#issuecomment-555961353))

- Remove `scree_overlay.png`
- Standardise pattern of coarse beach, scree and shingle

The current pattern of both scree and shingle is crude and there is no reason to better emphasise pebble beach than scree and shingle: they are all made of the same materials.

`natural=scree`
https://www.openstreetmap.org/#map=15/46.2032/7.1130
coarse_scree_z14

coarse_scree_z17

`natural=shingle`
https://www.openstreetmap.org/#map=17/46.07486/6.72905
coarse_shingle_z15

coarse_shingle_z17

comparison between `natural=beach` and `natural=shingle`
https://www.openstreetmap.org/#map=16/46.4013/6.5202
coarse_beach_z15

coarse_beach_z17

coarse_beach_z18


imagico:
> The current pattern of both scree and shingle is crude

How so?

> and there is no reason to better emphasise pebble beach than scree and shingle: they are all made of the same materials.

First of all it is not the same material.

Second if your argument is that _there is no reason to better emphasise pebble beach_ the logical step would be to change the pattern for coarse beaches to make it weaker.

Apart from that i think your examples show that this does not work well for scree - the pattern is very noisy on larger areas, it does not harmonize well with natural=bare_rock and paths and other features are poorly visible on it.

One argument for having a strong pattern for beaches is that it is supposed to be well readable with water as a background color as well.

jragusa:
> How so?

I mean it's much more a noisy pattern than a clear pattern such as those used for `orchard` or `allotments` for example

> First of all it is not the same material.

Could you please further explain ?

> Second if your argument is that _there is no reason to better emphasise pebble beach_ the logical step would be to change the pattern for coarse beaches to make it weaker.

That's another alternative

> Apart from that i think your examples show that this does not work well for scree - the pattern is very noisy on larger areas, it does not harmonize well with natural=bare_rock and paths and other features are poorly visible on it.

We can tune it to provide a more adequate rendering and there are already some issues opened about bad rendering of path along `natural=bare_rock`. See #1765 and #1748.
 
> One argument for having a strong pattern for beaches is that it is supposed to be well readable with water as a background color as well.

In other hand, it's mean that pebbly beach are better emphasised than shingle and scree although the latter (can) comprise much more larger block than pebbly beach.

imagico:
> I mean it's much more a noisy pattern than a clear pattern such as those used for orchard or allotments for example

The orchard and allotments patterns work very differently.  The scree/shingle pattern is a structure pattern that functions not by the mapper being able to see the individual symbols and their arrangement but by communicating an overall statistical structure.

In any case i don't quite understand the point here because your suggestion effectively would make natural=scree much more noisy.

> >   First of all it is not the same material.

> Could you please further explain ?

The coarse beach pattern is used for any beaches consisting of coarser material than sand - which is a very wide range of materials.  natural=shingle in terms of physical properties has a much more narrow meaning.  natural=scree is in terms of physical properties something again completely different.

I would not mind differentiating natural=scree and natural=shingle via pattern but it would need to be intuitively understandable to some extent at least.  And i don't think this should mean making natural=scree significantly heavier in rendering.

jragusa:
The basic assumption is that the pattern is "proportional" to the grain-size (IMO) and the latter is decreasing with the distance from the source. Hence, we should have scree > shingle > beach although large blocks also occur along rivers in relation to flooding.

jeisenbe:
@jragusa - would you be interested in designing a new pattern for just `natural=scree` - leaving the exiting pattern for `natural=shingle`?

PR #4055, opened 2020-03-07
Fix layering order for highway=construction by adding z-order for construction=* values

jeisenbe:
Fixes part of #3579

Follow-up of #3646

- Add "construction=" values to the z-order table in openstreetmap-carto.lua, so that the different categories of highway=construction will render in proper order (eg construction=motorway on top, construction=service on bottom). 
- This will take effect on database reload with v5.0.0
- Same as Commit 810448b808ef2dbec7d583bc2e31721bac16b12b from PR #3646 except that I've added construction=platform with z=9, same as construction=steps.

Highway=construction is currently set to z = 10 in the z-order table, so that it will render below almost all highways and paths. However, the different types of construction=*, such as construction=motorway, construction=trunk_link, construction=service, all have the same z value, so they are not layered according to classifaction.

Before
A6 - New Trunk Interchange - Northern Ireland
https://www.openstreetmap.org/#map=16/54.7629/-6.4847
- construction=trunk_link above construction=trunk
z16 Before
z16-new-trunk-links-master

After z16 
- Now the link is under.
z16-new-trunk-links-after

Murry's Island - Northern Ireland
https://www.openstreetmap.org/#map=15/54.7734/-6.5189
- New trunk road crosses an under-construction minor road and an under-construction secondary road (right) and an under-construction tertiary road (left)
z15 Before
z15-murrays-island-construction-master
z15 After
z15-murrays-island-construction-after


jeisenbe:
I notice that line 219-220 has:

```
construction    = {z = 10,  roads = false}
    },
```

But line aeroway and railway end like this:
```
        platform        = {z = 90,  roads = false},
    },
    aeroway = {
        runway          = {z = 60,  roads = false},
        taxiway         = {z = 50,  roads = false},
    },
```
Should the final comma be added to line 220 or removed from the other two lines? (though it doesn't seem to matter...)

jeisenbe:
>  test z_order({highway="construction", construction="service"}) 

This would be in `scripts/lua/test.lua`, probably after line 28?:
```
assert(z_order({highway="motorway", railway="rail"}) == 440 , "test failed: motorway + rail")
```

So I should add... 
```
assert(z_order({highway="construction", construction="service"}) == 15 , "test failed: highway=construction + construction=service")
```
Would that work right? I've pushed a commit with this.

pnorman:
> This would be in `scripts/lua/test.lua`, probably after line 28?:

Yes, although if you can review and merge #4056 there will be more examples of z_order tests.



> So I should add...
> 
> ```
> assert(z_order({highway="construction", construction="service"}) == 15 , "test failed: highway=construction + construction=service")
> ```
> 
> Would that work right? I've pushed a commit with this.

The test is correct, but the code is failing it. You can run the test locally with `lua scripts/lua/test.lua`

jeisenbe:
Well, the Travis build failed, so either the new test is wrong or something else is wrong. Perhaps `highway=construction` is overriding `construction=service`?

jeisenbe:
Oh, now I see your previous comment, sorry.

This is the current lua code, unfortunately I don't know much about how lua works.
```
function z_order(tags)
    local z = 0
    for k, v in pairs(tags) do
        if roads_info[k] and roads_info[k][v] then
            z = math.max(z, roads_info[k][v].z)
        end
    end
    return z ~= 0 and z or nil
end
```
The problem seems to be that all ways with `construction=*` also are going to have `highway=construction`, which gets z_order=33, and that seems to be happening first. 

I don't want to prioritize `construction=*` over `highway=`, since if there is a `highway=motorway` + `construction=motorway` this will still render as a motorway and should get `z_order = 380` not 38.

The pseudocode needs to be something like:

```
if roads_info[k=highway][v=construction] and roads_info[k=construction][v] then
  z = math.max(z, roads_info[k=construction][v].z)
if roads_info[k=highway] is not [v=construction] then
  z = math.max(z, roads_info[k][v].z)
```
Sorry, I'm not much help here.

pnorman:
> Well, the Travis build failed, so either the new test is wrong or something else is wrong. Perhaps `highway=construction` is overriding `construction=service`?

Your code is wrong, the test is right.

z_order({highway="construction", construction="service"}) returns 33. If you take the function and unroll the loop you get

```lua
local z = 0
if roads_info["highway"] and roads_info["highway"]["construction"] then
    z = math.max(z, roads_info["highway"]["construction"].z)
end
if roads_info["construction"] and roads_info["construction"]["service"] then
    z = math.max(z, roads_info["construction"]["service"].z)
end
return z ~= 0 and z or nil
```

Which boils down to
```lua
local z = 0
z = math.max(z, 33)
z = math.max(z, 15)
return z ~= 0 and z or nil
```

The problem is you've defined construction the same way as railway and areoway, and the logic is to take the max z_order of any of them.

It's not the most elegant, but one way to approach it would be to pull construction out of roads_info["highway"] completely and then add to the end of `z_order`
```lua
if tags["highway"] == "construction" then
    z = math.max(z, roads_info["highway"][tags["construction"]]/10 or 33)
end
```

I'm not happy with that approach, but I can't think of a better one off hand.

jeisenbe:
Sorry, I need the "for dummies" version. :-)

I tried this:
```
function z_order(tags)
    local z = 0
    for k, v in pairs(tags) do
        if roads_info[k] and roads_info[k][v] then
            z = math.max(z, roads_info[k][v].z)
        end
        if tags["highway"] == "construction" then
           z = math.max(z, roads_info["highway"][tags["construction"]]/10 or 33)
        end
    end
    return z ~= 0 and z or nil
end
```

And this:
```
function z_order(tags)
    local z = 0
    for k, v in pairs(tags) do
        if roads_info[k] and roads_info[k][v] then
            z = math.max(z, roads_info[k][v].z)
        end
    end
    return z ~= 0 and z or nil
    if tags["highway"] == "construction" then
        z = math.max(z, roads_info["highway"][tags["construction"]]/10 or 33)
    end
end
```
But both do not work.

pnorman:
> ```
> function z_order(tags)
>     local z = 0
>     for k, v in pairs(tags) do
>         if roads_info[k] and roads_info[k][v] then
>             z = math.max(z, roads_info[k][v].z)
>         end
>     end
>     return z ~= 0 and z or nil
>     if tags["highway"] == "construction" then
>         z = math.max(z, roads_info["highway"][tags["construction"]]/10 or 33)
>     end
> end
> ```
> 
> 
> But both do not work.

Move the return statement to the end, and check that you've deleted roads_info["construction"]

jeisenbe:
> check that you've deleted roads_info["construction"]

Yes, I had already deleted the section that starts with 
```
    construction = {
        motorway        = {z = 38, roads = false},
```

> Move the return statement to the end

Things that didn't work:

1)
```
function z_order(tags)
    local z = 0
    for k, v in pairs(tags) do
        if roads_info[k] and roads_info[k][v] then
            z = math.max(z, roads_info[k][v].z)
        end
        if tags["highway"] == "construction" then
           z = math.max(z, roads_info["highway"][tags["construction"]]/10 or 33)
        end
    end
    return z ~= 0 and z or nil
end
```

2) (Latest commit)
```
function z_order(tags)
    local z = 0
    for k, v in pairs(tags) do
        if roads_info[k] and roads_info[k][v] then
            z = math.max(z, roads_info[k][v].z)
        end
    end
    if tags["highway"] == "construction" then
       z = math.max(z, roads_info["highway"][tags["construction"]]/10 or 33)
    end
    return z ~= 0 and z or nil
end
```

Errors for both of these are: ```Failed to execute lua function for basic tag processing: openstreetmap-carto.lua:264: attempt to perform arithmetic on field '?' (a table value)
Failed to execute lua function for basic tag processing: openstreetmap-carto.lua:264: attempt to perform arithmetic on field '?' (a nil value)```

3) A separate function doesn't work either:
```
function z_order(tags)
    local z = 0
    for k, v in pairs(tags) do
        if roads_info[k] and roads_info[k][v] then
            z = math.max(z, roads_info[k][v].z)
        end
    end
    return z ~= 0 and z or nil
end

function z_order(tags)
    local z = 0
    for k, v in pairs(tags) do
        if tags["highway"] == "construction" then
            z = math.max(z, roads_info["highway"][tags["construction"]]/10 or 33)
        end
    end
    return z ~= 0 and z or nil
end
```

jeisenbe:
How about this option? 

Remove `construction    = {z = 10,  roads = false}` and just add
```
    construction = {
        motorway        = {z = 38, roads = true},
        trunk           = {z = 37, roads = true},
...
        null            = {z = 33, roads = false},
```
So this would make z_order=33 if there is no highway=* value (except highway=construction), right?

jeisenbe:
That last option didn't work ( `null = {z = 33...` did not work, nor did `nil = {z = 33...`), so how about something like this?

```
function z_order(tags)
    for k, v in pairs(tags) do
        if roads_info[k] and roads_info[k][v] then
            z = math.max(z, roads_info[k][v].z)
        end
        if not tags["construction"] then 
            z = 33
        end
    end
    return z ~= 0 and z or nil
end
```
"Something like" since that doesn't actually work, but you see what I'm going for...

jeisenbe:
@pnorman, if you are ready to merge v5.0.0, please don't let this PR delay things. It would be nice to have, but it's a fairly minor change, so I don't consider it worth holding things up.

jeisenbe:
Replaced by #4075

PR #4066, opened 2020-03-09
Change bridge-text way_pixels minimum and maximum

### LINKED ISSUE #4019: Text labels of man_made=bridge are shown too soon for short or narrow bridges
Related to #3979

### Expected behavior
- The text labels of man_made=bridge areas should generally be shorter than the longest dimension of the bridge, in most cases
- Common bridge names in languages which use an alphabetic script may be expected to include words about 10 characters long, which equals to about 50 pixels
- Many shorter bridges and viaducts are only 10 times longer than they are wide, therefore the bridge label can be shown at about 50*5= 250 way_pixels
- Most other area features, which are usually not so thin, are only labeled with their names at 3000 way_pixels: assuming an average length to width ratio of 2:1 for random landcover and landuse polygons, this is about 80 by 40 pixels.

### Actual behavior
- The way_pixels limit for man_made=bridge text lables is only 62.5 pixels, one zoom level sooner than suggested by the calculation above.
- While this is fine for long bridges (more than 30 times length vs width) or bridges with short name (no words longer than 6 characters), it shows the text label too soon for shorter bridges, or mid-length bridges with long names

### Suggested change
- Adjust way_pixels limit to 250 (1 zoom level later than current) or 187.5 (2 zoom levels sooner than other features)
- This will still show the bridge name almost 2 zoom level sooner than most area features, but not quite so soon as now.

### Examples:
Short (1 block long: 60 to 120 meters) road and pedestrian bridges in Chicago are already shown at z15, but the text label is usually significantly longer than the bridge itself:
chicago-bridges

At z16 the bridge outline is barely visible. The text label is still as long as the bridge, but it looks reasonable:

bridges-chicago-z16

z17 shows the extent of the bridges clearly, for comparison:
bridge-chicago-z17

### LINKED ISSUE #4050: man_made=bridge label upper way_area threshold should match that of landcovers
Otherwise we get irritating label style switches like from z18 to z19 on

https://www.openstreetmap.org/way/124882971

https://github.com/gravitystorm/openstreetmap-carto/blob/e7de1cd7c48d4efbb801f0b33fd7cc0dea143c36/roads.mss#L2712

https://github.com/gravitystorm/openstreetmap-carto/blob/eaa94ad01f6f530f92a8ddb144721335af626ea2/project.mml#L1592

See also #4019.

jeisenbe:
Fixes #4050
Fixes #4019 

- Change bridge-text way_pixels minimum limit to 125 (previously 62.5) and maximum to 768000
- This doubles the minimum area in pixels from the current limit (equivalent to 1/2 zoom level), and increases the max pixel area limit to match other features, to prevent double-tagged features from showing a different text label style at low zoom.
- A higher limit of 187.5 pixels (2 zoom levels sooner than other area features) was also tested, but this excludes some mid-sized bridges. 

Chicago before z15:
chicago-bridges

After 125 way_pixel limit:
chicago-after-125-z15

Bridge with landuse=railway before - https://www.openstreetmap.org/way/124882971
z18 (same)

z18-after-railway-bridge

z19 before
z19-before-railway-bridge

z19 after
z19-after-railway-bridge

### DIFF
--- a/project.mml
+++ b/project.mml
@@ -1411,8 +1411,8 @@ Layer:
           FROM planet_osm_polygon
           WHERE way && !bbox!
             AND man_made = 'bridge'
-            AND way_area > 62.5*POW(!scale_denominator!*0.001*0.28,2)
-            AND way_area < 64000*POW(!scale_denominator!*0.001*0.28,2)
+            AND way_area > 125*POW(!scale_denominator!*0.001*0.28,2)
+            AND way_area < 768000*POW(!scale_denominator!*0.001*0.28,2)
           ORDER BY way_area DESC
         ) AS bridge_text
     properties:
--- a/style/roads.mss
+++ b/style/roads.mss
@@ -2709,7 +2709,7 @@ tertiary is rendered from z10 and is not included in osm_planet_roads. */
 
 #bridge-text  {
   [man_made = 'bridge'] {
-    [zoom >= 12][way_pixels > 62.5][way_pixels <= 64000] {
+    [zoom >= 12][way_pixels > 125][way_pixels <= 768000] {
       text-name: "[name]";
       text-size: 10;
       text-wrap-width: 30; // 3 em


jeisenbe:
@matkoniecz re: your comment: https://github.com/gravitystorm/openstreetmap-carto/issues/4019#issuecomment-597264845 - are you ok with this PR?

PR #4068, opened 2020-03-10
Remove duplicate selection of cave_entrance, peak, volcano and saddle

jeisenbe:
- Remove duplicate selection of natural = cave_entrance, peak, volcano and saddle
- The features `natural=cave_entrance`, `natural=peak`, `natural=volcano` and `natural=saddle` are always mapped as nodes, so they should not be selected when mapped as areas
- They are selected in the previous line when way_area is null. It appears that this was accidentally duplicated when text-poly, text-point, amenity-points etc were merged into one layer.

We could consider still rendering `natural=cave_entrance` on areas. JOSM allows it, but this is only done 1% of the time. However, I'm keeping it simple and just removing this bug now. If someone is in favor of rendering `natural=cave_entrance` areas, that can be done in a separate PR.

https://www.openstreetmap.org/#map=16/56.3770/22.6305
https://www.openstreetmap.org/way/187727062
https://www.openstreetmap.org/way/187727047
- Two peaks mapped as closed ways

Before - render currently
natural-peak-areas-before

After
natural-peak-areas-after

### DIFF
--- a/project.mml
+++ b/project.mml
@@ -1477,8 +1477,8 @@ Layer:
                                                     'greenhouse_horticulture', 'retail', 'industrial', 'railway', 'commercial', 'brownfield', 'landfill',
                                                     'construction', 'military', 'plant_nursery') THEN landuse ELSE NULL END,
                 'natural_' || CASE WHEN "natural" IN ('peak', 'volcano', 'saddle', 'cave_entrance') AND way_area IS NULL THEN "natural" ELSE NULL END,
-                'natural_' || CASE WHEN "natural" IN ('wood', 'peak', 'volcano', 'saddle', 'cave_entrance', 'water', 'mud', 'wetland', 'bay', 'spring',
-                                                      'scree', 'shingle', 'bare_rock', 'sand', 'heath', 'grassland', 'scrub', 'beach', 'glacier', 'tree', 'strait', 'cape')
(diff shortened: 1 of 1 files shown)


kocio-pl:
> If someone is in favor of rendering natural=cave_entrance areas, that can be done in a separate PR.

This is another patch trying to change too much things in one PR, which is a constant source of problems even for a simple review. Please, try to avoid this, especially since you're making a tons of different changes, which makes it much harder to follow. This is not only this case.

If you want to delete duplication, this should be clearly defined PR exactly for this task. And breaking rendering should never be included as a subtask, anyway.

jeisenbe:
This is a very simple PR: it just removes `'peak', 'volcano', 'saddle', 'cave_entrance'` from line 1480 in `amenity-points.mss`

I am not changing the rendering: the new rendering is the same as the old rendering from April 2019

The rendering was unintentionally changed by @pnorman in the huge PR which moved everything to one amenity-points layer: #3712 (April 2019)

Previously to that PR, the tags `natural=peak`, `volcano`, `saddle` and `cave_entrance` were only rendered when tagged on nodes.

imagico:
> I am not changing the rendering: the new rendering is the same as the old rendering from April 2019

Exactly.  This fixes a bug introduced in #3712 that i failed to notice during review.

Not tested this yet but in principle :+1:.

kocio-pl:
I have misunderstood it then, sorry. Cleaning code is OK for me.

Issue #3461, opened 2018-10-19
Remake natural=spring as dot+icon


kocio-pl:
Follow up to #3189.
Proof of concept code is here (based on v4.10.0): https://github.com/gravitystorm/openstreetmap-carto/compare/master...kocio-pl:spring-dot

I'm not satisfied with our new spring rendering. It was based on a clever idea, but in practice it's not too good. My proposition is to get back to the original "s"-based design and follow the way we're rendering fountains. Here's why:
- it's best not to use letters and stay with geometric shapes ("s" may be misleading in other languages than English)
- on medium zoom levels we need just something to spot, and since we already have blue dot as a symbol of a fountain, it should be easier to get the meaning (man made or natural water source)
- on higher levels there's enough space to show real icon depicting object, so it's clear that this is not a fountain
- it fixes the problem of rendering on water areas
- dot is smaller than ring

I also tried to make both dots and labels more visible, so I used the same darker blue color as fountain and made bigger halo.

Examples:

[Named spring is better visible than currently](https://www.openstreetmap.org/node/1100451137)

![hgkypdhd](https://user-images.githubusercontent.com/5439713/47247583-f2026100-d404-11e8-82c9-e7594c9f0b04.png)
![ll8xgnpg](https://user-images.githubusercontent.com/5439713/47247588-f464bb00-d404-11e8-96ad-a6762954e0bf.png)

[Spring in the middle of a water area](https://www.openstreetmap.org/node/5442671644)

![_mimirc](https://user-images.githubusercontent.com/5439713/47247539-9fc14000-d404-11e8-922f-a0ec1bf461d0.png)
![epsvrlfa](https://user-images.githubusercontent.com/5439713/47247538-9e901300-d404-11e8-915e-b166bd510bd2.png)

[Lot of springs on the grass](https://www.openstreetmap.org/#map=14/48.2824/24.1234) (even better than "s")

![yei2krqw](https://user-images.githubusercontent.com/5439713/47247470-3e996c80-d404-11e8-8596-232329b2aa07.png)

![_uohlkab](https://user-images.githubusercontent.com/5439713/47247363-78b63e80-d403-11e8-93f3-cb9c075c5f42.png)

[Lot of springs in the forest](https://www.openstreetmap.org/#map=14/48.5753/9.0193)

![wmkncwpn](https://user-images.githubusercontent.com/5439713/47247762-ecf1e180-d405-11e8-9cda-38512449cfc8.png)

[...]

PR #4069, opened 2020-03-10
Move natural=spring back to amenity-points layer

### LINKED ISSUE #4011: natural=spring blocks place node labels
Example illustrating the problem: https://www.openstreetmap.org/#map=15/49.6064/8.1085&layers=N

natural=spring has priority over place node labels. As a result, the label for this village gets blocked by the spring at z14 and thus does not render. It gets rendered again at z13 when we stop rendering natural=spring. 

We definitely should not render natural=spring in such a way that it can block place node labels, even if it is uncommon for place nodes and springs to be close together like this.

Related to #3461

jeisenbe:
Fixes #4011

- Move `natural=spring` icon back to amenity-points layer, delete current `springs` layer
- Previously springs were rendered below water-lines. This is not longer the case, so springs can be rendered with other symbols again
- Springs now render above paths and highways, like other point symbols.
- This change should slightly improve performance, and springs will no longer render with priority to placenames such as hamlets and villages.

z14 Before
z14-japan-springs-before png

z14 After - place label now seen (upper left)
z14-japan-springs-after

z15 before
z15-japan-springs-before png

z15 after
z15-japan-springs-after

z16 before
z16-japan-springs-before-

z16 after
z16-japan-springs-after
- Shinto place of worship, place name, and funicular station now all show with priority over the springs.

### DIFF
--- a/project.mml
+++ b/project.mml
@@ -342,33 +342,6 @@ Layer:
         ) AS water_barriers_poly
     properties:
       minzoom: 13
-  - id: springs
-    geometry: point
-    <<: *extents
-    Datasource:
-      <<: *osm2pgsql
-      table: |-
-        (SELECT
-            way,
-            "natural"
-          FROM
-          (SELECT
-              ST_PointOnSurface(way) AS way,
-              "natural"
-            FROM planet_osm_polygon
-            WHERE way && !bbox!
-            AND way_area < 768000*POW(!scale_denominator!*0.001*0.28,2)
-          UNION ALL
-          SELECT
-              way,
-              "natural"
-            FROM planet_osm_point
-            WHERE way && !bbox!
-            ) _
-          WHERE "natural" IN ('spring')
-        ) AS springs
-    properties:
-      minzoom: 14
   - id: piers-poly
     geometry: polygon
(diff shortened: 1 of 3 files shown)


imagico:
Note this symbol for springs was never intended to be used as a POI style icon.

I would probably rather prefer to see a full revert of #3189 for the reasons explained in https://github.com/gravitystorm/openstreetmap-carto/issues/325#issuecomment-365714748 but there is likely no consensus for that.

jeisenbe:
I am in favor of changing the spring icon. I think @kocio-pl also thinks it is too large at low zoom level. We discussed this previously but did not reach consensus in #3461 

I would support using a much smaller, abstract icon at low zoom levels and a slightly smaller icon at high zoom levels. 

But this PR will be an improvement, since springs are currently blocking placenames that appear sooner.

jeisenbe:
I also would like to see geysers and hot springs rendered: https://github.com/gravitystorm/openstreetmap-carto/issues/3461#issuecomment-431665748 - also see test images in https://github.com/gravitystorm/openstreetmap-carto/issues/3461#issuecomment-437582441

PR #4081, opened 2020-03-16
Remove unneeded `ELSE NULL` from case statements

### LINKED ISSUE #2735: Explicit nulls on feature statements
Right now our feature statements are like
```sql
COALESCE(
  'aeroway_' || CASE WHEN aeroway IN ('helipad', 'aerodrome') THEN aeroway ELSE NULL END,
  'power_' || CASE WHEN power IN ('generator') THEN power ELSE NULL END,
  ...
) AS feature
```

The explict `ELSE NULL` does nothing, as a case statement where none of the conditions are true and there is no else clause is already null.

Does this help people understand the statements? If I was writing for myself, I wouldn't add it, but I write a lot of SQL.

Either way, we should document it in the code standards.

jeisenbe:
Fixes #2735

- Remove unneeded explicit `ELSE NULL` from case statements in project.mml

These statements are not needed, because "NULL" is the default when none of the cases are true.

In #2735 there was agreement that these explicit NULL statements are not helpful

Rendering appears to be unchanged.

### DIFF
--- a/project.mml
+++ b/project.mml
@@ -66,9 +66,9 @@ Layer:
           FROM (SELECT
               way,
               ('landuse_' || (CASE WHEN landuse IN ('forest', 'farmland', 'residential', 'commercial', 'retail', 'industrial',
-                                                    'meadow', 'grass', 'village_green', 'vineyard', 'orchard') THEN landuse ELSE NULL END)) AS landuse,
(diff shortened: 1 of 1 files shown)


jeisenbe:
Rebased. It would be nice to get this reviewed soon, since it will frequently lead to merge conflicts.

Issue #4084, opened 2020-03-19
Re-evaluation of the rendering of labels for place=square


imagico:
This issue is being separated out of #4043 based on https://github.com/gravitystorm/openstreetmap-carto/issues/4043#issuecomment-593045858 and following.

Please keep in mind that the criterion for rendering something/not rendering something based on our [goals](https://github.com/gravitystorm/openstreetmap-carto/blob/master/CARTOGRAPHY.md) needs to be if a tag is consistently used with a well defined and verifiable meaning and if the suggested rendering encourages such use.

@jeisenbe has in https://github.com/gravitystorm/openstreetmap-carto/issues/4043#issuecomment-593046473 and https://github.com/gravitystorm/openstreetmap-carto/issues/4043#issuecomment-593046673 illustrated a large variation in what the tag is used for.  There in particular seems a strong overlap between `place=square` and `highway=pedestrian` (and to some extent other highway tags), `leisure=park` as well as to `junction=yes`.

[The documentation on the wiki](https://github.com/gravitystorm/openstreetmap-carto/issues/4043#issuecomment-593046673) lacks in a precise, verifiable and language independent definition of the meaning probably largely contributing to the lack of consistency in use.  Until today the documented definition was largely language based (what is called a square or similar words in other languages) and will therefore widely vary depending on cultural context.  This part has been removed (except for the English language reference to the term 'square') but there are new non-verifiable criteria added now (suitability for certain uses, architectural significance).  None of these actually seems to reflect the practical use of the tag world wide.

Those who are in favor of keeping rendering place=square labels should therefore concentrate on demonstrating a consistent and broadly accepted meaning of the tag and how the current rendering supports use of the tag specifically according to this meaning.

polarbearing:
The request for a more precise definition is often heard here when someone requests to render a feature. However, how can we achieve "precise, verifiable and language independent definition" in OSM? 

This is OSM, not public administration where a definition has to be court-proof. I just came across this definition of amenity=brothel : ["Prostitutionsstätten im Sinne des Prostituiertenschutzgesetzes vom 21. Oktober 2016 (BGBl. I S. 2372), das durch Artikel 57 des Gesetzes vom 20. November 2019 (BGBl. I S. 1626) geändert worden ist,..."](https://www.berlin.de/rbmskzl/aktuelles/rathaus-aktuell/2020/meldung.906890.php) (...are to be closed because of COVID-19). Precise and verifiable, probably not language independent yet.

[...]

PR #4085, opened 2020-03-20
Render place=square in amenity-points layer with placenames style from z17

### LINKED ISSUE #4043: place=square label display should not depend on way_area without visual feedback on the geometry
As said in the subject - polygons tagged `place=square` are rendered depending on way_area - see

https://github.com/gravitystorm/openstreetmap-carto/blob/e7de1cd7c48d4efbb801f0b33fd7cc0dea143c36/roads.mss#L3307-L3309

without there being any visual feedback on the geometry.

Introduced in #2673, see also https://github.com/gravitystorm/openstreetmap-carto/pull/2673#issuecomment-313035750, https://github.com/gravitystorm/openstreetmap-carto/pull/2700#issuecomment-317236118.

Side note:  The whole rendering of place=square with nodes in `placenames-small` and polygons in `roads-area-text-name` is confusing and difficult to maintain.

jeisenbe:
Fixes  #4043

- Removes rendering of `place=square` areas from roads_area_text_name and moves rendering from placenames to amenity-points like place=locality, since the text label is only rendered at z17 and above. 
- This change improves the rendering priority system, since now place=square features will not block icons which are rendered at z15 and z16. 
- The text style is slightly adjusted to match `place=locality` and `place=neighborhood` features

This makes place=square rendering consistent with `place=locality`, and more similar to `place=neighborhood` and `place=quarter`, with the rendering starting at a fixed zoom level.

Currently, very large place=square areas will render sooner than z17, and very small features will ony render from z18. Now all will initially render at z17, if they are not blocked by a higher-priority feature (such as a `highway=pedestrian` area label or a junction=yes node, or by an amenity=marketplace or leisure=park tag on the same feature)

However, unlike those other "place" features, this will still render `place=square` when mapped as an area or as a node. (Most place=square features are mapped as areas)

### DIFF
--- a/project.mml
+++ b/project.mml
@@ -1279,25 +1279,21 @@ Layer:
         (SELECT
             way,
             place,
-            leisure,
             name
           FROM planet_osm_point
           WHERE place IN ('village', 'hamlet')
              AND name IS NOT NULL
              AND NOT tags @> 'capital=>yes'
              OR (place IN ('suburb', 'quarter', 'neighbourhood', 'isolated_dwelling', 'farm')
-                 OR (place IN ('square')
-                     AND (leisure is NULL OR NOT leisure IN ('park', 'recreation_ground', 'garden')))
              ) AND name IS NOT NULL
           ORDER BY CASE
-              WHEN place = 'suburb' THEN 8
-              WHEN place = 'village' THEN 7
-              WHEN place = 'hamlet' THEN 6
-              WHEN place = 'quarter' THEN 5
-              WHEN place = 'neighbourhood' THEN 4
-              WHEN place = 'isolated_dwelling' THEN 3
-              WHEN place = 'farm' THEN 2
-              WHEN place = 'square' THEN 1
+              WHEN place = 'suburb' THEN 7
+              WHEN place = 'village' THEN 6
+              WHEN place = 'hamlet' THEN 5
+              WHEN place = 'quarter' THEN 4
+              WHEN place = 'neighbourhood' THEN 3
+              WHEN place = 'isolated_dwelling' THEN 2
+              WHEN place = 'farm' THEN 1
             END DESC,
             length(name) DESC,
             name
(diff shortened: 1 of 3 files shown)


jeisenbe:
Test images:

https://www.openstreetmap.org/#map=18/1.4042908/103.7898478
Before - same text style as highway=pedestrian
square2-before-18:1 4042908:103 7898478

After - placenames-small text style
square2-after-18:1 4042908:103 7898478


https://www.openstreetmap.org/#map=18/1.3585403/103.7511304
Before - not rendered (too small)
square1-before-18:1 3585403:103 7511304

After - rendered (z18)
square1-after-18:1 3585403:103 7511304


https://www.openstreetmap.org/#map=18/1.28187/103.84498
Before - rendered
square4-before-18:1 28187:103 84498

After - blocked by tourism information icon (shows at z19 still)
square4-after-18:1 28187:103 84498


https://www.openstreetmap.org/#map=18/50.84506/4.35649
* Brussels,  double-tagged as landuse=residential + place=square
Before - rendered like highway=pedestrian (due to place=square tag)
square-brussels-before-18:50 84506:4 35649

After - rendered with landuse=residential style
square-brussels-after-18:50 84506:4 35649


https://www.openstreetmap.org/#map=15/54.6953/20.5027.png
Large square in Kaliningrad (not just the central green square area, but all 3)
Before - label renders already at z15
square-kaliningrad-big-before-15:54 6953:20 5027

After - not rendered till z17
square-kaliningrad-big-after-15:54 6953:20 5027

kocio-pl:
> The text style is slightly adjusted to match place=locality and place=neighborhood features 

If I understand you right, this will make a departure from rendering of a square consisting of just a pedestrian area, which would make different types of square inconsistent?

jeisenbe:
The tag`place=square` is sometimes added to a `highway=pedestrian` area, but it is also added to `leisure=park` features, road junctions (which are also mapped as `junction=yes`) and for neighborhoods that are named after a square. All 4 of those features are rendered differently. Since the key is `place`, I believe it is most sensible to render it similarly to `place=neighborhood` or `place=locality`. 

Using the `highway=pedestrian` area style does not work, because that style is designed to render on top of the gray pedestrian area color. There is no fill color for `place=square`.

We could instead use the text style from junction=yes (shared by highway=traffic_signal), which has a little darker text and slightly smaller halo, but I think the placenames style is best for mapper feedback, since it matches the "place" key.

polarbearing:
> Since the key is place, I believe it is most sensible to render it similarly to place=neighborhood or place=locality. [...] There is no fill color for place=square

Indeed. As we do not have fill colours for place=village or place=country. Consequently there should be no polygon fill for place=square. Mapping place=square as polygon is useful for other uses of OSM, e.g. in Nominatim.

jeisenbe:
See https://github.com/gravitystorm/openstreetmap-carto/issues/4084#issuecomment-608200518 - the definition for this feature has been improved

Though the area that should be mapped is still not well defined and we could not render this as an area, I believe this PR will be an acceptable way to render the name of a square which does not require adding the name tag directly to the highway=pedestrian area

project.mml
```diff
@@ -1468,11 +1464,11 @@ Layer:
                                                     'greenhouse_horticulture', 'retail', 'industrial', 'railway', 'commercial', 'brownfield', 'landfill',
                                                     'construction', 'military', 'plant_nursery') THEN landuse END,
                 'natural_' || CASE WHEN "natural" IN ('peak', 'volcano', 'saddle', 'cave_entrance') AND way_area IS NULL THEN "natural" END,
-                'natural_' || CASE WHEN "natural" IN ('wood', 'water', 'mud', 'wetland', 'bay', 'spring', 'scree', 'shingle', 'bare_rock', 'sand', 'heath',
-                                                      'grassland', 'scrub', 'beach', 'glacier', 'tree', 'strait', 'cape')
+                'natural_' || CASE WHEN "natural" IN ('wood', 'peak', 'volcano', 'saddle', 'cave_entrance', 'water', 'mud', 'wetland', 'bay', 'spring',
```
imagico, line 1467: This change does not belong here.

jeisenbe: Sorry, it is now fixed

jeisenbe:
> My concern here is ... that features with point labels but not otherwise rendered in the style will have priority over features in roads-area-text-name

Right, prior to this PR `place=square` is rendered from `placenames-small`, which has higher priority than many features, including roads-area-text-name.

After this PR, `place=square` feature will be rendered from the `text-point` layer (though the code is in `amenity-points` as usual), which is after all of the important icon and text layers. It will be rendered in the same text style like place=locality, which is also now rendered in `text-point`.

But `highway=pedestrian` + `area=yes` features with a name= will still have this rendered in `roads-area-text-name`, much earlier. 

Polygon features that are double-tagged as `highway=pedestrian` + `place=square` will still be rendered in the `highway=pedestrian` area style, from `roads-area-text-name`

imagico:
> After this PR, place=square feature will be rendered from the text-point layer (though the code is in amenity-points as usual), which is after all of the important icon and text layers.

Right - sorry, i got confused since i had forgotten that the query for amenity-points gets reused later for the labels.

jeisenbe:
@imagico could you review this again when you have the time?

Issue #3400, opened 2018-09-17
Render tunnel=flooded with tunnel styling


jragusa:
### Expected behavior
Support `tunnel=flooded` for waterways

### Actual behavior
Only `tunnel=yes` and `tunnel=culvert` are rendered.

There is only 573 flooded tunnels but the lack of support creates confusion with superficial waterways.

### Links and screenshots illustrating the problem
https://www.openstreetmap.org/way/586127462

before
![flooded_before](https://user-images.githubusercontent.com/9897203/45653186-62048b00-bad7-11e8-9045-8c644b8c20ca.png)

after
![flooded_after](https://user-images.githubusercontent.com/9897203/45653194-692b9900-bad7-11e8-8529-4e4ef50f19cb.png)

HolgerJeromin:
Thanks for your suggestion.
Is there a documentation for this tag available?

Adamant36:
@HolgerJeromin, https://wiki.openstreetmap.org/wiki/Tag:tunnel%3Dflooded

It comes up fifth in the search for the term tunnel=flooded for some reason.

matkoniecz:
I changed the title a bit to make it more clear what is requested.

Gazer75:
This is due to the approval of this:
https://wiki.openstreetmap.org/wiki/Proposed_features/Hydropower_water_supplies

matkoniecz:
Given that it is just new value of tunnel tag I think that it would be OK. I would recommend checking random `tunnel=flooded`, check history and look whatever the edit was problematic (problematic as in "there is an ongoing discussion in changeset or proposal to revert it for some good reason")

imagico:
Small correction:  About 1/3 of uses of tunnel=flooded + waterway=canal are correctly used for qanats (together with canal=qanat) - which is probably also the most widespread real life situation where this tag could be used.  This is something that would be good to render (though better with a dedicated styling rather than a generic waterway tunnel signature)

Gazer75:
https://www.openstreetmap.org/#map=17/58.65856/6.73393
That hydro plant I mapped is a good one to test at. You have headrace, penstock and tailrace.
Zoom out a bit and you'll see all the tunnels feeding it from different lakes and river intakes.

imagico:
Yes, but as said it is highly unfortunate if mappers choose to re-purpose established waterway tags for those so data users have no reliable way to distinguish between these and something like

https://www.openstreetmap.org/way/242158791

[...]

Issue #3354, opened 2018-08-20
Render waterway=canal thinner


SelfishSeahorse:
Please render `waterway=canal` thinner. The current rendering looks very strange for small irrigation canals and seems to lead people to wrongly tag them as `waterway=ditch`, `waterway=drain` (which are both used for carrying superfluous water) or `waterway=stream`.

Example of a 1 m wide canal ([location](https://www.openstreetmap.org/#map=18/46.95915/7.38393)) which is rendered ca 5 m wide at zoom 18:

![untitled](https://user-images.githubusercontent.com/14875493/44354900-5ce9f700-a4ab-11e8-9291-ce9e7dddf96c.png)

See: [key:waterway](https://wiki.openstreetmap.org/wiki/Key:waterway) on the wiki the [this email](https://lists.openstreetmap.org/pipermail/tagging/2018-August/038460.html) from the tagging mailing list ([full discussion](https://lists.openstreetmap.org/pipermail/tagging/2018-August/thread.html#38452)).

kocio-pl:
Maybe there's a way to show how big (wide) are they without measuring them? I would probably render these with `boat/motorboat/ship=yes` and `name` as currently, but by default show them the same as ditch, drain or stream:

https://taginfo.openstreetmap.org/tags/waterway=canal#combinations

Upsides:
- this would probably not change rendering of big, named canals (unless the name is in relation probably)
- this would encourage people to add more data when possible
- it's quite simple

Maybe we could also decide using `width=*` tag.

What do you think about it?

eehpcm:
> Maybe we could also decide using width=* tag.

That's one way.  The other is to map an area with natural=water + water=canal in addition to the linear canal feature, as detailed [here](https://wiki.openstreetmap.org/wiki/Tag:waterway%3Dcanal). I've tried both, neither currently have any effect.

The effect is particularly weird at lower zooms.  I have a mill pond with mill race tagged as a canal using water=canal as an area.  At z=19 it looks OK.  Zoom out a little and the mill race balloons up until it's almost as big as the mill pond.  It did the same thing before I tried water=canal, when I used width=0.5 and width=1 (the latter just in case it was rejecting non-integer values).

It would be nice if the carto handled both ways of dealing with a narrow canal, but if you're only going to implement one then please update the wiki to reflect that fact (update the wiki even before you actually implement it to reduce the number of canals added before implementation that later have to be re-tagged).

polarbearing:
[...]

PR #4087, opened 2020-03-20
Remove the rendering of waterways with tunnel=flooded

### LINKED ISSUE #3179: Render waterway with tunnel=flooded like tunnel=yes
### Expected behavior
Ways with waterway=* + tunnel=flooded may be rendered like ways with waterway=* + tunnel=yes
I think canal in tunnels should be thinner on the map from zoom 16 to 19 than now. Colors may be inverted, with way in blue and casing in white or grey

### Actual behavior
Tag tunnel=flooded has no effect, tunnel=yes shows a really wide way with solid white and blue casing

https://www.openstreetmap.org/way/216220208
https://www.openstreetmap.org/way/301303197

According to https://wiki.openstreetmap.org/wiki/Tag:tunnel%3Dflooded description and review, tunnel=yes shouldn't be compatible with waterway=canal or waterway=river but since tunnel=flooded has just been introduced and it needs lot of time to carefully move tunnel=yes, render can remain the same.

Thanks in advance and best regards

jragusa:
Fixes #3179

- exclude `tunnel=flooded` for waterways

https://www.openstreetmap.org/way/369442787
Before
flooded_before_z15

After
flooded_after_z15

### DIFF
--- a/project.mml
+++ b/project.mml
@@ -176,6 +176,7 @@ Layer:
             CASE WHEN tunnel IN ('yes', 'culvert') THEN 'yes' ELSE 'no' END AS int_tunnel
           FROM planet_osm_line
           WHERE waterway IN ('stream', 'drain', 'ditch')
+            AND tunnel NOT IN ('flooded')
         ) AS water_lines_casing
     properties:
       minzoom: 13
--- a/project.mml
+++ b/project.mml
@@ -2122,7 +2122,7 @@ Layer:
           FROM planet_osm_line
           WHERE (waterway IN ('river', 'canal', 'stream', 'drain', 'ditch')
                  OR "natural" IN ('bay', 'strait'))
-            AND (tunnel IS NULL or tunnel NOT IN ('culvert', 'flooded'))
+            AND (tunnel IS NULL OR tunnel NOT IN ('culvert', 'flooded'))
             AND name IS NOT NULL
           ORDER BY COALESCE(layer,0)
         ) AS water_lines_text
--- a/project.mml
+++ b/project.mml
@@ -215,6 +215,7 @@ Layer:
           FROM planet_osm_line
           WHERE waterway IN ('river', 'canal', 'stream', 'drain', 'ditch')
             AND (bridge IS NULL OR bridge NOT IN ('yes', 'aqueduct'))
+            AND (tunnel IS NULL OR tunnel NOT IN ('flooded'))
           ORDER BY COALESCE(layer,0)
         ) AS water_lines
     properties:
@@ -2121,7 +2122,7 @@ Layer:
           FROM planet_osm_line
           WHERE (waterway IN ('river', 'canal', 'stream', 'drain', 'ditch')
                  OR "natural" IN ('bay', 'strait'))
-            AND (tunnel IS NULL or tunnel != 'culvert')
+            AND (tunnel IS NULL or tunnel NOT IN ('culvert', 'flooded'))
             AND name IS NOT NULL
           ORDER BY COALESCE(layer,0)
         ) AS water_lines_text
--- a/project.mml
+++ b/project.mml
@@ -173,10 +173,11 @@ Layer:
             CASE WHEN tags->'intermittent' IN ('yes')
               OR tags->'seasonal' IN ('yes', 'spring', 'summer', 'autumn', 'winter', 'wet_season', 'dry_season')
               THEN 'yes' ELSE 'no' END AS int_intermittent,
-            CASE WHEN tunnel IN ('yes', 'culvert') THEN 'yes' ELSE 'no' END AS int_tunnel
+            CASE WHEN tunnel IN ('yes', 'culvert') 
+              OR waterway='canal' AND tunnel = 'flooded'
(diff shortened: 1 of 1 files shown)


jragusa:
Fixed

imagico:
I would be very reluctant to completely hide major features with widespread use based on the presence of a secondary tag.  This would make data interpretation much more difficult for other data users because they would not be able to rely on all data with for example `waterway=stream` being evaluated by visual mapper feedback through this style but they would have to specifically exclude features with the tag `tunnel=flooded` from their interpretation because these lack such visual feedback.

Independent of that i would like to see evaluation that `tunnel=flooded` is actually used consistently on the waterway types we render in a way that is distinct from `tunnel=yes`.  There is a high risk that with this change `tunnel=flooded` would change into a synonym for essentially `tunnel=yes` + `render=no` - in other words:  That it would be used by mappers specifically for features they subjectively would want to hide from the map.

jeisenbe:
>  i would like to see evaluation that tunnel=flooded is actually used consistently on the waterway types we render in a way that is distinct from tunnel=yes

In theory this tag is supposed to mean that the tunnel is not passable by humans, while tunnel=yes might allow humans to enter, but I suspect that this may not be happening. "A long (>100m) tunnel where flowing water or other fluid prevent humans from safely walking inside".

But according to the author of the proposal, "tunnel=yes shouldn't be compatible with waterway=canal or waterway=river" (https://github.com/gravitystorm/openstreetmap-carto/issues/3179#issue-315550655) - so I suppose he thinks all river and canal tunnels should be flooded=yes? Not sure why, clearly their are rivers which are shallow enough to walk along.

From the comment https://github.com/gravitystorm/openstreetmap-carto/issues/3179#issuecomment-529864020 I had though you were in favor of removing the rendering of waterways with this tag, but perhaps that was for waterway=canal exclusively? 

There are 54 uses with waterway=drain (https://overpass-turbo.eu/s/RNE) and 69 with waterway=river (https://overpass-turbo.eu/s/RNF), in addition to 1515 with waterway=canal (https://overpass-turbo.eu/s/RNG) and some combinations with pipeline=* or waterway=pressurised

jragusa:
Since the definition of `tunnel=flooded` is not correctly constrained, the best would be to entirely remove the rendering. In addition, I suppose these waterways are not permanently flooded and hence, tunnel are accessible during dry period or of low output.

imagico:
> In theory this tag is supposed to mean that the tunnel is not passable by humans, while tunnel=yes might allow humans to enter, but I suspect that this may not be happening.

Well, even if it is - the question arises if it makes sense to differentiate between the two by rendering one in a dedicated form and not rendering the other at all.  The difference in meaning for the map user is small while the difference in rendering is extreme (even to the point that it completely overrides the primary tagging).

> From the comment #3179 (comment) I had though you were in favor of removing the rendering of waterways with this tag, but perhaps that was for waterway=canal exclusively?

That was for waterway=canal only - and it was based on the assumption that use of tunnel=flooded differs clearly from tunnel=yes.  Overall the best would be an overall concept for canal rendering (taking into account in particular also #3354).

> Since the definition of tunnel=flooded is not correctly constrained, the best would be to entirely remove the rendering

But the thing is we don't render `tunnel=flooded`, we render `waterway=*`.

jeisenbe:
@imagico - would you be in favor of continuing to ignore `tunnel=flooded` for now (thus rendering the canal, ditch or drain like an above-ground waterway), or could we consider rendering it the same as `tunnel=yes`?

imagico:
I would only interpret tunnel=flooded if there is agreement among mappers on how to use it and if it is used accordingly.

If tunnel=flooded is used on waterway=canal in a well defined meaning i would be fine with rendering it like tunnel=yes or not rendering it if that well represents this meaning.  But only on the waterway types where it is actually used for.  So far it seems to me that use would more likely warrant rendering like tunnel=yes (with #3354 applying in that case as well of course).  Not rendering waterway=canal + tunnel=flooded should not be a substitute for solving #3354.

imagico:
Note the reasoning behind my original consideration not to render waterway=canal in combination with tunnel=flooded was because the choice of waterway tag (waterway=canal vs. waterway=pressurized) seemed often arbitrary.  It might be a best not to look at interpreting tunnel=flooded on existing waterway tags without considering that as well as interpreting usage=*/canal=*.

See also https://github.com/gravitystorm/openstreetmap-carto/issues/3400#issuecomment-481286679 and https://github.com/gravitystorm/openstreetmap-carto/issues/3400#issuecomment-481306064

pnorman:
@imagico if you have changes to request or objections, can you do a review indicating so so it appears  in the status?

jeisenbe:
@jragusa - could you change this to just consider tunnel=flooded with waterway=canal, for now?

jragusa:
With the tunnel styling or as currently rendered ?

jeisenbe:
With the tunnel styling

jragusa:
Rendering of `waterway=canal` + `tunnel=flooded`
flooded_after2_z15

jeisenbe:
@imagico - does the latest commit address your request for changes?

jeisenbe:
@imagico would you review this?

PR #4096, opened 2020-03-29
Remove rendering of residential, unclassified, cycleway, footway, path highway areas

### LINKED ISSUE #1967: Problems with rendering "highway=track; area=yes" (and to a lesser extent other highway types)
Essentially, the problem is this:

"highway=blah; area=yes" is rendered by OSM carto, and "area:highway=blah" is not.  The former is intended for "true" highway areas (pedestrian plazas etc.); the latter for adding the "width" of linear highways (a bit like "waterway=riverbank" is on "waterway=river").  The problem is, as discussed on https://github.com/gravitystorm/openstreetmap-carto/issues/180 and the links from there such as https://github.com/gravitystorm/openstreetmap-carto/issues/1504 , area:highway can't be supported yet, so people are using "highway=blah; area=yes" on linear features.

In some cases (pedestrian plazas) it's perfectly valid; in others (track, cycleway) it surely isn't.  Here's an example I found today:

http://www.openstreetmap.org/#map=17/53.23152/-0.48302

(for completeness I tried to explain the problem on https://www.openstreetmap.org/changeset/35066497 ).

It's a similar logic to used at https://github.com/gravitystorm/openstreetmap-carto/issues/1280#issuecomment-72629181 - if OSM Carto renders something, it is in some sense a "seal of approval"; we all see that mappers shouldn't "tag for the renderer" but we all know that many tag specifically for OSM Carto, and a specific goal of CARTOGRAPHY.md is to provide a "feedback mechanism for mappers to validate their edits".

So, which of the "current highway area values" (which I suspect is the list at https://github.com/gravitystorm/openstreetmap-carto/blob/982bc27a01143aea334c1e1a13edd8c25630b9c5/project.yaml#L763 ) are really valid?

I'd suggest 'residential', 'unclassified', 'pedestrian', 'service', 'platform' only; can anyone point to a real-world feature that's validly mapped as an area=yes highway (i.e. is not primarily a linear feature) that is 'footway', 'cycleway', 'track', or 'path'?

### LINKED ISSUE #3995: Drop rendering for highway=unclassified and highway=residential areas
### Expected behavior

highway=unclassified + area=yes is not rendered

### Actual behavior

highway=unclassified + area=yes is rendered as an area

Based on #3985 research it appears that there are nearly no (or maybe not even a single one) highway=unclassified + area=yes that is mapped correctly.

pnorman:
Fixes #3995
Fixes #1967

Removes the rendering of highway=residential area=yes and highway=unclassified area=yes polygons. In #3995 these were established to commonly be a tagging mistake.

image

### DIFF
--- a/project.mml
+++ b/project.mml
@@ -642,7 +642,7 @@ Layer:
         (SELECT
             way,
             COALESCE((
-              'highway_' || (CASE WHEN highway IN ('pedestrian', 'service', 'footway', 'cycleway', 'track', 'path', 'platform') THEN highway END)),
+              'highway_' || (CASE WHEN highway IN ('pedestrian', 'service', 'platform') THEN highway END)),
               ('railway_' || (CASE WHEN (railway IN ('platform')
                               AND (tags->'location' NOT IN ('underground') OR (tags->'location') IS NULL)
                               AND (tunnel NOT IN ('yes', 'building_passage') OR tunnel IS NULL)
@@ -650,7 +650,7 @@ Layer:
                               THEN railway END))
             ) AS feature
           FROM planet_osm_polygon
-          WHERE highway IN ('pedestrian', 'service', 'footway', 'cycleway', 'track', 'path', 'platform')
+          WHERE highway IN ('pedestrian', 'service', 'platform')
             OR (railway IN ('platform')
                 AND (tags->'location' NOT IN ('underground') OR (tags->'location') IS NULL)
                 AND (tunnel NOT IN ('yes', 'building_passage') OR tunnel IS NULL)
@@ -764,8 +764,8 @@ Layer:
         (SELECT
             way,
             COALESCE(
-              ('highway_' || (CASE WHEN highway IN ('pedestrian', 'service', 'footway', 'cycleway', 'living_street',
-                                                    'track', 'path', 'platform', 'services') THEN highway END)),
+              ('highway_' || (CASE WHEN highway IN ('pedestrian', 'service', 'living_street',
+                                                    'platform', 'services') THEN highway END)),
               ('railway_' || (CASE WHEN (railway IN ('platform')
                               AND (tags->'location' NOT IN ('underground') OR (tags->'location') IS NULL)
                               AND (tunnel NOT IN ('yes', 'building_passage') OR tunnel IS NULL)
@@ -774,7 +774,7 @@ Layer:
               (('aeroway_' || CASE WHEN aeroway IN ('runway', 'taxiway', 'helipad') THEN aeroway END))
             ) AS feature
           FROM planet_osm_polygon
-          WHERE highway IN ('pedestrian', 'service', 'footway', 'cycleway', 'living_street', 'track', 'path', 'platform', 'services')
+          WHERE highway IN ('pedestrian', 'service', 'living_street', 'platform', 'services')
             OR (railway IN ('platform')
                 AND (tags->'location' NOT IN ('underground') OR (tags->'location') IS NULL)
                 AND (tunnel NOT IN ('yes', 'building_passage') OR tunnel IS NULL)
--- a/project.mml
+++ b/project.mml
@@ -642,7 +642,7 @@ Layer:
         (SELECT
             way,
             COALESCE((
-              'highway_' || (CASE WHEN highway IN ('residential', 'unclassified', 'pedestrian', 'service', 'footway', 'cycleway', 'track', 'path', 'platform') THEN highway END)),
+              'highway_' || (CASE WHEN highway IN ('pedestrian', 'service', 'footway', 'cycleway', 'track', 'path', 'platform') THEN highway END)),
               ('railway_' || (CASE WHEN (railway IN ('platform')
                               AND (tags->'location' NOT IN ('underground') OR (tags->'location') IS NULL)
                               AND (tunnel NOT IN ('yes', 'building_passage') OR tunnel IS NULL)
@@ -650,7 +650,7 @@ Layer:
                               THEN railway END))
             ) AS feature
           FROM planet_osm_polygon
-          WHERE highway IN ('residential', 'unclassified', 'pedestrian', 'service', 'footway', 'track', 'path', 'platform')
+          WHERE highway IN ('pedestrian', 'service', 'footway', 'cycleway', 'track', 'path', 'platform')
             OR (railway IN ('platform')
                 AND (tags->'location' NOT IN ('underground') OR (tags->'location') IS NULL)
                 AND (tunnel NOT IN ('yes', 'building_passage') OR tunnel IS NULL)
@@ -764,7 +764,7 @@ Layer:
         (SELECT
             way,
             COALESCE(
-              ('highway_' || (CASE WHEN highway IN ('residential', 'unclassified', 'pedestrian', 'service', 'footway', 'cycleway', 'living_street',
+              ('highway_' || (CASE WHEN highway IN ('pedestrian', 'service', 'footway', 'cycleway', 'living_street',
                                                     'track', 'path', 'platform', 'services') THEN highway END)),
               ('railway_' || (CASE WHEN (railway IN ('platform')
                               AND (tags->'location' NOT IN ('underground') OR (tags->'location') IS NULL)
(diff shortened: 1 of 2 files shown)


project.mml
```diff
@@ -642,15 +642,15 @@ Layer:
...
+              'highway_' || (CASE WHEN highway IN ('pedestrian', 'service', 'footway', 'cycleway', 'track', 'path', 'platform') THEN highway END)),
               ('railway_' || (CASE WHEN (railway IN ('platform')
                               AND (tags->'location' NOT IN ('underground') OR (tags->'location') IS NULL)
                               AND (tunnel NOT IN ('yes', 'building_passage') OR tunnel IS NULL)
                               AND (covered NOT IN ('yes') OR covered IS NULL))
                               THEN railway END))
             ) AS feature
           FROM planet_osm_polygon
-          WHERE highway IN ('residential', 'unclassified', 'pedestrian', 'service', 'footway', 'track', 'path', 'platform')
+          WHERE highway IN ('pedestrian', 'service', 'footway', 'cycleway', 'track', 'path', 'platform')
```
jeisenbe: I see this adds back highway=cycleway, which is currently missing from this selection. 

Do we actually want to render highway=cycleway areas? 

Are there really areas where bicycles may travel freely in any direction?

pnorman: I haven't given it much thought either way. The WHERE clauses were the only places missing cycleway, we were attempting to render it in the MSS and the COALESCE

pnorman:
#1967 requests removing footway, cycleway, track, and path so I'm going to add those

polarbearing:
What is wrong with area=yes on a footway (unless of course it is abused for the width of a linear feature)? It is very popular for minor areas of omnidirectional foot traffic, such as the round area in the middle of a park where all footways meet and you have some benches around. This is too insignificant for me to be a "highway=pedestrian". What is the logic to render area=yes on _pedestrian_ but not on _footway_?

The example above with the hidden house is a tagging error, it should have been cut out as a MP. 

Do we start dropping all rendering where a tagging mistake has been found?

HolgerJeromin:
> What is wrong with area=yes on a footway (unless of course it is abused for the width of a linear feature)? 

I added another example in:
https://github.com/gravitystorm/openstreetmap-carto/issues/1967#issuecomment-155718094

jeisenbe:
> What is wrong with area=yes on a footway (unless of course it is abused for the width of a linear feature)

While many examples are misused in such a way, it appears that majority of features with this tag are small pedestrian areas. 

The problem with the current rendering is that it works well with highway=pedestrian, since the fill and casing colors are the same, but it doesn't look anything like footways or cycleways.

polarbearing:
> The problem with the current rendering is that it works well with highway=pedestrian, since the fill and casing colors are the same, but it doesn't look anything like footways

A solution would be to unify the appearance of linear highway=footway and =pedestrian, in the current pedestrian style (grey) just with different width (footway a bit narrower). That would also reduce the noise from the red footways. 

IIRC there were attempt to do something like that here in the past.

polarbearing:
As for highway=track areas, they are certainly rendered too prominently, a good approach would to adapt them more to the track appearance, e.g. making them darker and/or use some hatching or pattern.

polarbearing:
I **disagree strongly**. 

First, this is yet another example of the current practice, that after discussing a certain change in rendering for one feature, another feature is included in this change on the last minute, without being sufficiently discussed. 

In this case, the proposed change is not even present in the title of this PR, it is kind of a stealth removal. 

Second, the PR now removes a feature used 55 k + 7k = 62k times, documented in the wiki and _not declared as deprecated_. 

The removal _will cause pointless mechanical changing_ such areas to highway=pedestrian, which they often aren't.

While @imagico  sees them as synonyms, I see a difference. There are plenty other examples that are tagged differently but rendered identical. OSM is a database that is not just driving this style. It has other applications where this distinction is significant.

imagico:
Clearing up a few misconceptions:

* reviews of this PR are based on all the proposed changes - both me and @jeisenbe specifically discussed the changes that were not originally included.
* the main basis of rendering decisions here is actual use of tags as it can be observed in the database.  Neither the wiki nor what individual mappers view different tags to mean or supposed to mean have a particular significance.
* any analysis and demonstration that highway=footway is used on polygons systematically with a different meaning than highway=pedestrian would be most welcome.

jeisenbe:
To clarify: I am happy with merging this PR as-is, because the `highway=footway` + `area=yes` features are usually small areas, they are not large pedestrian plazas or public squares which require a special rendering so that the center of a town or city will be represented properly.

However, if @pnorman decides to keep the current rendering of `highway=footway` + `area=yes` as identical to `highway=pedestrian` + `area=yes`, I would also support that. Either way is fine by me.

pnorman:
> First, this is yet another example of the current practice, that after discussing a certain change in rendering for one feature, another feature is included in this change on the last minute, without being sufficiently discussed.

Additional changes were added to the PR in response to comments from a review.

> In this case, the proposed change is not even present in the title of this PR, it is kind of a stealth removal.

I've now edited the title to reflect the changes.

> Second, the PR now removes a feature used 55 k + 7k = 62k times, documented in the wiki and _not declared as deprecated_.

I don't consider the wiki status as important as the actual usage numbers.

-----

> Based on that for unclassified, residential, track and path i'd say removing rendering due to a lack of broad use and acceptance by mappers is in order.

I'm not sure, based on the table I see use. Not as much as pedestrian, obviously, but more than the other roads in the table.

jeisenbe:
> > Based on that for unclassified, residential, track and path i'd say removing rendering due to a lack of broad use and acceptance by mappers is in order.

> I'm not sure, based on the table I see use. Not as much as pedestrian, obviously, but more than the other roads in the table.

@pnorman - do you mean you are not sure about highway=footway areas? I

t's true that area tagging of those features is more common that the others on the list: about 1 out of 200 footway features are currently mapped as areas, while for the others on the list there are less than 1 per 1000.

Like I've said, I support removing the rendering of residential, unclassified, track and cycleway area renderings, but I can go either way about highway=footway - it's odd to render it the same as highway=pedestrian, but we could work on fixing that later, if necessary.

jeisenbe:
I've checked the local data in Oregon, and I noticed that the great majority of highway=footway + area=yes closed ways also have a man_made=pier or public_transport=platform feature:

http://overpass-turbo.eu/s/Sob - 44 total features
http://overpass-turbo.eu/s/Soc - 39 with man_made=pier or public_transport=platform

Overall, a smaller percentage of `highway=footway` areas have this combination: http://overpass-turbo.eu/s/Sod - 3416 total. I believe the iD editor was adding these tags to piers and platforms automatically, so this mapping is probably not intentional. 

In the case of platform, the correct tag is highway=platform for the bus platforms which were mapped in Oregon (or sometimes railway=platform in other contexts):

https://www.openstreetmap.org/#map=19/44.04760/-123.07723
highway-pedestrian-oregon-bus-platform-19:44 04760:-123 07723

For the piers, the tagging is not exactly wrong, but it's not commonly added by mappers who are not using iD:



Here are the other 5 features in the State:

https://www.openstreetmap.org/#map=19/45.50853/-122.67222
- Two adjacent square areas in a park:
highway-pedestrian-oregon-18:44 56114:-123 27742

highway-pedestrian-oregon-19:45 50853:-122 67222

highway-pedestrian-oregon-18:44 76010:-124 06489

https://www.openstreetmap.org/#map=17/44.04177/-123.06829
- Error: addition of highway=footway + area=yes to leisure=track
highway-ped-leisure-track-oregon-17:44 04177:-123 06829

Rendering with this PR:
highway-ped-leisure-track-oregon-after

jeisenbe:
Man made piers + highway=pedestrian:

https://www.openstreetmap.org/#map=18/44.80906/-124.05909

Current
highway-ped-man_made-pier-before

After
highway-ped-man_made-pier-before-18:44 80906:-124 05909

https://www.openstreetmap.org/#map=19/44.98982/-123.99457
before
man-made-pier-highway-ped-before-19:44 98982:-123 99457

after
pier-highway-ped-after-2

jeisenbe:
@pnorman - I'm waiting for you to merge this PR as-is or make changes if you want. 

Do you plan to merge it soon, including removing the rendering of highway=footway areas, or change it to keep the rendering of highway=footway for now?

pnorman:
> Do you plan to merge it soon, including removing the rendering of highway=footway areas, or change it to keep the rendering of highway=footway for now?

I'll change it

project.mml
```diff
@@ -1758,7 +1758,7 @@ Layer:
             name
           FROM planet_osm_polygon
           WHERE way && !bbox!
-            AND (highway IN ('residential', 'unclassified', 'pedestrian', 'service', 'footway', 'cycleway', 'living_street', 'track', 'path', 'platform')
```
jeisenbe, line 1761: 'footway' is missing here

polarbearing:
If I understand correctly, area/footway is being kept for now. Could this please been reflected in the title? It seems that the message to osm-talk still included the removal, probably by copy/pasting?

jeisenbe:
Sorry, that should not have been in the list of changes for v5.1.0. I've updated the title of this PR.

polarbearing:
footway still in the title?

jeisenbe:
Title updated, thank you for catching that.

PR #1788, opened 2015-08-26
path/footway/cycleway - render missing surface separately, more prominent rendering for unpaved ways

matkoniecz:
It changes rendering for highway=path/footway/cycleway.

Now there are three [buckets](https://github.com/gravitystorm/openstreetmap-carto/issues/1698#issuecomment-134938213)*:

- highway=path/cycleway/footway tagged as paved
- valid surface tag is missing
- highway=path/cycleway/footway tagged as unpaved

There are contradictory goals that I tried to satisfy:

- make unpaved footways more prominent
- fit third stage - between paved and unpaved for missing surface
- avoid making paved footways more prominent
- keep each footway type recognisable
- keep footway style less prominent than highway=tracks
- make footways visible across different landcovers
- and many additional that in this case were not highly problematic

I think that it is works fairly well (or at least better than currently and previously used style).

Footways are still too dominant in fully mapped cities, important walking routes are not visible enough, paved/unknown/unpaved are not different enough etc. But it is near limits of useing the same style used across the world (yes, there is currently no available method to display [highway=path, surface=unpaved] differently depending whatever it is an unimportant path in crowded park or the main route across the mountains).

*Yes, this link leads to comment that is worth reading.

Examples for cities:
https://cloud.githubusercontent.com/assets/899988/9492585/18a7c242-4bf9-11e5-948c-2b9460eb48fe.png
https://cloud.githubusercontent.com/assets/899988/9492587/18aae9ea-4bf9-11e5-9455-bd627c5529da.png
https://cloud.githubusercontent.com/assets/899988/9492588/18ab11fe-4bf9-11e5-99ac-59786f54a2ab.png
https://cloud.githubusercontent.com/assets/899988/9492586/18aab8c6-4bf9-11e5-850a-227b50635331.png

Outside cities:
https://cloud.githubusercontent.com/assets/899988/9492602/404eaa72-4bf9-11e5-9189-bab5509794c3.png
https://cloud.githubusercontent.com/assets/899988/9492603/40520c4e-4bf9-11e5-8628-30ab33d54d93.png
https://cloud.githubusercontent.com/assets/899988/9492604/4079d972-4bf9-11e5-96ab-398d551f8619.png
https://cloud.githubusercontent.com/assets/899988/9492605/40fb4bd8-4bf9-11e5-8280-324af165b9c1.png

fixes #1765 and #1766

[...]

PR #4097, opened 2020-03-29
Render foot/cyclepaths without regard to surface

### LINKED ISSUE #1765: unpaved footways on rocky areas are poorly visible

### LINKED ISSUE #1793: footway line styles difficult to distinguish and in general much weaker than cycleway

pnorman:
Rendering with a different dash pattern to indicate paved, unpaved, or unknown was a good idea, but in practice the difference between them is too subtle to be seen and understood.

The existing rendering also constrained us cartographically, making it near-impossible to balance the patterns against a variety of backgrounds.

I expect this idea to be somewhat controversial, and I like the theory of showing the surface with different dash patterns. It just doesn't work in practice with our wide assortment of backgrounds.

Fixes #1765
Fixes #1793

### DIFF
--- a/project.mml
+++ b/project.mml
@@ -430,7 +430,6 @@ Layer:
             foot,
             bicycle,
             tracktype,
-            int_surface,
             access,
             construction,
             service,
@@ -444,11 +443,6 @@ Layer:
                 foot,
                 bicycle,
                 tracktype,
-                CASE WHEN surface IN ('unpaved', 'compacted', 'dirt', 'earth', 'fine_gravel', 'grass', 'grass_paver', 'gravel', 'ground',
-                                      'mud', 'pebblestone', 'salt', 'sand', 'woodchips', 'clay', 'ice', 'snow') THEN 'unpaved'
-                  WHEN surface IN ('paved', 'asphalt', 'cobblestone', 'cobblestone:flattened', 'sett', 'concrete', 'concrete:lanes',
-                                      'concrete:plates', 'paving_stones', 'metal', 'wood', 'unhewn_cobblestone') THEN 'paved'
-                END AS int_surface,
                 CASE WHEN access IN ('destination') THEN 'destination'::text
                   WHEN access IN ('no', 'private') THEN 'no'::text
                 END AS access,
@@ -477,7 +471,6 @@ Layer:
                 foot,
                 bicycle,
                 tracktype,
-                'null',
                 CASE
                   WHEN access IN ('destination') THEN 'destination'::text
                   WHEN access IN ('no', 'private') THEN 'no'::text
@@ -496,8 +489,7 @@ Layer:
             z_order,
             CASE WHEN substring(feature for 8) = 'railway_' THEN 2 ELSE 1 END,
             CASE WHEN feature IN ('railway_INT-preserved-ssy', 'railway_INT-spur-siding-yard', 'railway_tram-service') THEN 0 ELSE 1 END,
-            CASE WHEN access IN ('no', 'private') THEN 0 WHEN access IN ('destination') THEN 1 ELSE 2 END,
-            CASE WHEN int_surface IN ('unpaved') THEN 0 ELSE 2 END
+            CASE WHEN access IN ('no', 'private') THEN 0 WHEN access IN ('destination') THEN 1 ELSE 2 END
         ) AS tunnels
     properties:
       cache-features: true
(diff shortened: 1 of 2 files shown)


jeisenbe:
I agree, the current rendering is not clear. I've always forgotten that there is a difference, because the rendering seems too subtle to notice.

It would be great to have a clear way of distinguishing paved and unpaved paths and cycleways. I would rather show this information instead of access=. 

But I would support this change for the moment, until we can develop an effective rendering.

jeisenbe:
I'm not sure that #1793 will be fixed by this? Also see previous discussion in PR #1788

imagico:
I think this would strategically point us in the opposite direction of where we need to be going.  And i think this neither fixes #1765 nor #1793 in a convincing and long term viable fashion and it does not fix or promises a path to fix #1748.

This style has moved with past changes (in particular #747) very much in direction towards emphasis of automobile transport and de-emphasis of other modes of transports.  This trend should be reversed because one of the key unique selling points of OSM compared to other map data sources is the extensive and differentiated mapping of non-road infrastructure.

There have been plenty of demonstrations already been made how well readable and balanced differentiation of narrow line signatures can be designed - assuming you accepts these being more prominently visible than they are right now - @SomeoneElseOSM has shown that  (https://map.atownsend.org.uk/maps/map/map.html#zoom=15&lat=-24.9959&lon=135.05924) - i have also demonstrated it specifically for the problem discussed here (http://blog.imagico.de/drawing-the-lines/).

Frankly if the desire is to reduce differentiation of non-automobile path rendering i would rather remove the difference between footway and cycleway (which is in parts based on an irritating ruleset anyway) and maintain the paved/unpaved difference.

jeisenbe:
I find your arguments pretty convincing. After looking back at the attempts to solve this previously, I'm not sure if I could quickly replace this with a better design.

@imagico, would you be willing to submit a PR to implement the ideas in (http://blog.imagico.de/drawing-the-lines/) or something similar?

Solid purple, red, brown (and green for bridleways?) lines would not be mistaken for the admin boundaries, but we can work on changing the colors of those if needed.

imagico:
> @imagico, would you be willing to submit a PR to implement the ideas in (http://blog.imagico.de/drawing-the-lines/) or something similar?

I don't think this would work well with the purple boundaries, in particular for access restricted paths.  And it would mean a partial revert of #3467.

jeisenbe:
>  access restricted paths

I'm willing to lose the distinction of access=no if necessary to
render surface= in a sensible way.

But the purple boundaries will be changing shortly, I think. Even the
slightly desaturated purple (e.g. Lch(47,30,327) is sufficiently
different that there will not be a problem implementing the purple
cycleways and magenta motorways, based on test images and delta E
calculations.

I'm also happy to see #3467 reverted partially or entirely.

PR #4102, opened 2020-04-02
Render man_made=goods_conveyor

### LINKED ISSUE #1999: Render conveyor belts (man_made=goods_conveyor)
man_made=goods_conveyor is used [nearly 1000 times](http://taginfo.openstreetmap.org/tags/?key=man_made&value=goods_conveyor). Conveyor belts are relevant for bulk materials (coal, iron ore, copper ore, lime, wheat, corn) at open-pit mines, port terminals and industrial plants.

As alternative sometimes railway=monorail is used (example: [http://www.openstreetmap.org/relation/62366#map=14/51.3573/10.5061](http://www.openstreetmap.org/relation/62366#map=14/51.3573/10.5061)) which then could be abandoned.

jragusa:
Fixes #1999

- Add the rendering of `man_made=goods_conveyor` using a broadly similar pattern as `aerialway`
- `line-dasharray: 12,6` + `#999999`

### TGV Lyon-Turin - Descenderie de Saint-Martin la Porte
https://www.openstreetmap.org/way/627540810
z=18
conveyor12-6_z18_colour3
z=17
conveyor2_z17
z=16
conveyor2_z16
z=15
conveyor2_z15

### Picampoix
https://www.openstreetmap.org/way/163719137
https://www.openstreetmap.org/way/163719136
https://www.openstreetmap.org/way/163719139
https://www.openstreetmap.org/way/163719140
z=18
conveyor3_z18
z=17
conveyor3_z17
z=16
conveyor3_z16
z=15
conveyor3_z15


jeisenbe:
Thank you, @jragusa, for submitting this PR. 

Since it is currently listed as a draft and WIP (work in progress), are you planning to make some more changes or do further tests prior to requesting a review?

project.mml
```diff
@@ -1647,13 +1647,15 @@ Layer:
           COALESCE(
            'highway_' || CASE WHEN tags @> 'ford=>yes' OR tags @> 'ford=>stepping_stones' THEN 'ford' ELSE NULL END,
            'leisure_' || CASE WHEN leisure IN ('slipway', 'track') THEN leisure ELSE NULL END,
-           'attraction_' || CASE WHEN tags @> 'attraction=>water_slide' THEN 'water_slide' ELSE NULL END
+           'attraction_' || CASE WHEN tags @> 'attraction=>water_slide' THEN 'water_slide' ELSE NULL END,
+           'man_made_' || CASE WHEN man_made IN ('goods_conveyor') THEN man_made ELSE NULL END
```
jeisenbe, line 1651: Rather than rendering this in the amenity-line layer, which will be on top of power-lines and many text features, please move this to the aerialways layer, like man_made=pipeline - this will render below bridges, power-lines, and text.

style/amenity-points.mss
```diff
@@ -3028,6 +3028,38 @@
       }
     }
   }
+
+  [feature = 'man_made_goods_conveyor'] {
+    [zoom >= 12] {
```
jeisenbe, line 3033: Have you tested this at z12 and z13? Are there long enough goods_conveyor features to be shown clearly at z12? With this pattern the goods_conveyor will need to be at least 40 pixels long for the pattern to be understood.

style/amenity-points.mss
```diff
@@ -3028,6 +3028,38 @@
...
+      line/line-width: 1;
+      line/line-join: round;
+      line/line-cap: round;
+      line/line-color: #999999;
+
+      dash/line-width: 3.5;
+      dash/line-join: round;
+      dash/line-color: #999999;
+      dash/line-dasharray: 12,6;
```
jeisenbe, line 3042: Rather than starting with the dash, try using half of the gap at the beginning and the end, like this: `dash/line-dasharray: 0,3,12,3;` 

Or you could split the long dash in half instead: `dash/line-dasharray: 6,6,6,0;`

style/amenity-points.mss
```diff
@@ -3028,6 +3028,38 @@
...
+      dash/line-clip: false;
+      [zoom >= 17] {
+        text-name: "[name]";
```
jeisenbe, line 3045: It would be better to render the name in the `text-line` layers. See PR #4107 for an example.

project.mml
jragusa: fixed

style/amenity-points.mss
jragusa: Indeed, z14 is suitable in most cases but there are rare large scale features requiring z12 to z13. What do you suggest ?

https://www.openstreetmap.org/relation/2606647
z12
conveyor4_z12

z13
conveyor4_z13

z14
conveyor4_z14

style/amenity-points.mss
jragusa: I put the second suggestion in the PR
`dash/line-dasharray: 0,3,12,3;`
conveyor2_z18-03123

`dash/line-dasharray: 6,6,6,0;`
conveyor2_z18-6660

style/amenity-points.mss
jragusa: fixed

style/amenity-points.mss
jeisenbe: In these images, at z12 the current rendering is too strong: it is much wider than most types of highways and other features, so it would need to be a thinner version.

But even then, in this example many of the goods_conveyor ways are very close to overlapping at z12. By z14 this is not a problem. Most of these ways are several kilometers long

I downloaded all man_made=goods_conveyor ways, and checked their length in JOSM. Only 370 out of 4324 are longer than 2 kilometers, which is where they are long enough to see clearly at z12. 3130 are less than 500 meters long, and the median length is only 200 meters.

So I would recommend starting the rendering at z14.

project.mml
```diff
@@ -816,8 +816,9 @@ Layer:
             name
           FROM planet_osm_line
           WHERE aerialway IS NOT NULL
-            OR (man_made IN ('goods_conveyor', 'pipeline')
+            OR (man_made = 'pipeline'
                 AND tags-> 'location' IN ('overground', 'overhead', 'surface', 'outdoor')
+                OR man_made = 'goods_conveyor'
```
jeisenbe, line 821: This line needs to move down below "OR bridge IN...", outside of the close parenthesis .

Also, I see that there are 217 man_made=goods_conveyor ways with `tunnel=yes` or `location=underground`. It would be good to exclude these from rendering.

project.mml
```diff
@@ -1615,9 +1616,9 @@ Layer:
           name,
           layer,
           COALESCE(
-           'highway_' || CASE WHEN tags @> 'ford=>yes' OR tags @> 'ford=>stepping_stones' THEN 'ford' END,
-           'leisure_' || CASE WHEN leisure IN ('slipway', 'track') THEN leisure END,
-           'attraction_' || CASE WHEN tags @> 'attraction=>water_slide' THEN 'water_slide' END
+           'highway_' || CASE WHEN tags @> 'ford=>yes' OR tags @> 'ford=>stepping_stones' THEN 'ford' ELSE NULL END,
```
jeisenbe, line 1619: "ELSE NULL" has accidentally been added on these three lines, please remove

project.mml
```diff
@@ -1965,7 +1966,7 @@ Layer:
             NULL AS way_area,
             CASE WHEN building = 'no' OR building IS NULL THEN 'no' ELSE 'yes' END AS is_building
           FROM planet_osm_line
-          WHERE (man_made IN ('pier', 'breakwater', 'groyne', 'embankment')
+          WHERE (man_made IN ('pier', 'breakwater', 'groyne', 'embankment', 'goods_conveyor')
```
jeisenbe, line 1969: 'goods_conveyor' is added here, but there is no mss code to render the name. Is this intentional? 

I see that there are only 78 different names for goods conveyors, and at least half of those are "band", "belt", "conveyor" or a similar description rather than a name, so perhaps we don't need to render the names of these features?

project.mml
jragusa: fixed

project.mml
jragusa: fixed

project.mml
jragusa: yes, that's better to remove it

project.mml
```diff
@@ -1616,7 +1618,7 @@ Layer:
           name,
           layer,
           COALESCE(
-           'highway_' || CASE WHEN tags @> 'ford=>yes' OR tags @> 'ford=>stepping_stones' THEN 'ford' ELSE NULL END,
+           'highway_' || CASE WHEN tags @> 'ford=>yes' OR tags @> 'ford=>stepping_stones' THEN 'ford' END,
            'leisure_' || CASE WHEN leisure IN ('slipway', 'track') THEN leisure ELSE NULL END,
```
jeisenbe, line 1622: These two lines still have "ELSE NULL"

project.mml
jeisenbe: Only the first line was fixed

project.mml
jragusa: My apologies, I should have better check, it's now fixed

jragusa:
With the corrected pattern
z14
conveyor2_newpattern_z14

z15
conveyor2_newpattern_z15

z16
conveyor2_newpattern_z16

z17
conveyor2_newpattern_z17

z18
conveyor2_newpattern_z18

imagico:
I think the geometric design looks quite good now - but i am a bit worried about the color.  Introducing more and more gray features to the style for a large variety of things is tricky in a style that otherwise uses color for differentiating a lot.  In particular the fact that the color for industrial features like this is identical to that for natural features like cliffs is a bit weird.

I have no specific suggestion to solve this but i would encourage you to give this some thought.

jeisenbe:
I didn't remember that cliff.svg is in `#999999` - it would be nice if this was encoded in the .mss file rather than in the svg. 

@kocio-pl previously suggested using violet or purple for industrial features like pipelines.

Since the pipeline rendering uses color to specify the substance (water, oil or gas), I did not use purple for that feature, also because the gray is less visually prominent and these features are not so important. 

But it could be considered, with low chroma, especially if we plan to change the admin-boundaries away from violet.

e.g 

**1) `#aa91a6` Lch(63,15,330)** - same lightness as `#999999`, same hue as `@industrial`:
- Might be too saturated still

conveyor-after-4a

conveyor-after-4b

conveyor-after-4d

conveyor-after-4f


**2) `#a494a1` Lch(63,10,330)**
Less saturated, but the violet hue is still nice. 

conveyor-after-5a

conveyor-after-5b

conveyor-after-5d

conveyor-after-5f


**3) `#a1959f` Lch(63,7,330)**
Even less saturated, perhaps too gray

conveyor-after-6a

conveyor-after-6b

conveyor-after-6d

conveyor-after-6f

jeisenbe:
The other color that might make sense would be something similar to the building outline color (`#d9d0c9`), so hue 66

4) `#a69588` - Lch(63,10,66)
conveyor-after-7a

conveyor-after-7b

conveyor-after-7d

conveyor-after-7f


5) `#a2968d ` - Lch(63,7,66)
conveyor-after-8a

conveyor-after-8b

conveyor-after-8d

conveyor-after-8f

jragusa:
Comparison over quarry
`#aa91a6`
conveyor4_aa91a6_quarry_z17

`#a494a1`
conveyor4_a494a1_quarry_z17

`#a1959f`
conveyor4_a1959f_quarry_z17

`#a69588`
conveyor4_a69588_quarry_z17

`#a2968d`
conveyor4_a2968d_quarry_z17

jeisenbe:
Thank you, that's helpful. What is your preference?

jragusa:
colour of 5 and especially 4 are close to track ways, 1 is a bit to saturated, so I suggest 2 or 3.

imagico:
For colors in general:  Small features call for stronger colors to achieve good readability.  

To be honest given the current use of colors in this style there is most likely not any good option to just pick (and it is probably also confusing to have different colors for pipelines and this of course).  The whole thing would probably require a bit more of a boarder look in terms of color systematics.

Going with gray as a generic color and differentiating different gray lines only by structure is not necessarily a bad idea.  Note aerialway=goods and man_made=pipeline are both conceptually similar to this proposal so it might make sense to unify them in color.  Tough this might require adjustment in line widths to avoid making things more or less heavy without a good reason.

jeisenbe:
> Small features call for stronger colors to achieve good readability.

Do you think the line needs to be a bit darker then? The current pipeline color is `#909090` which is L60 instead of the L63 used here so far. We could try switching to that instead, e.g:

6) `#9d8c99` - Lch(60,10,331)
conveyor-9d8c99-10g

conveyor-9d8c99-10h

conveyor-9d8c99-10i

conveyor-9d8c99-10j


7) `#9f8b9b` - Lch(60,12,331)
conveyor-9f8b9b-12h

conveyor-9f8b9b-12i

conveyor-12j-9f8b9b


> aerialway=goods and man_made=pipeline are both conceptually similar to this proposal so it might make sense to unify them in color.

Yes, I would consider changing man_made=pipeline to use the same color. One can also argue that aerialway=goods should continue to use the same darker gray color as other aerial cableway features which it resembles in physical form, though the function is similar to a conveyor or pipeline, so either color might work in that case.

> this might require adjustment in line widths

All the aerialway patterns at z12-13 and z14-15 should probably be adjusted to be less heavy and thinner than at z16-z19, but that's another issue.

But do you think the latest line widths for goods_conveyor look about right compared to man_made=pipeline?

The width of the line and dash are the same as man-made=pipeline at z14 and thinner at z15 (though the dashes are much longer here than with pipeline), and the total line width of piplines is the same as the dash width at z18, wider at z19, to accomodate a wider centerline with the substance color, which isn't really needed here.

jeisenbe:
Some more tests:

**7) `#9f8b9b` - Lch(60,12,331)**
conveyor-9f8b9b-12a

conveyor-9f8b9b-12b

conveyor-9f8b9b-12d

conveyor-9f8b9b-12f

jeisenbe:
Comparison of 2) vs 7):

**2) `#a494a1` Lch(63,10,330)**
conveyor-a494a1-5k

conveyor-a494a1-5l

conveyor-a494a1-5m


**7) `#9f8b9b` - Lch(60,12,331)**
conveyor-9f8b9b-12k

conveyor-9f8b9b-12l

conveyor-9f8b9b-12m

jeisenbe:
Check of man_made=pipeline:

**0) Current rendering: `#909090` Lch(60,0,0)** - pipelines
pipeline-current-A

pipeline-current-B

pipeline-current-C

pipeline-current-D

pipeline-current-E


**2) `#a494a1` Lch(63,10,330)** - pipelines
conveyor-a494a1-5-A

conveyor-a494a1-5-B

conveyor-a494a1-5-C

conveyor-a494a1-5-D-

conveyor-a494a1-5-E


**7) `#9f8b9b` - Lch(60,12,331)** - pipelines
conveyor-9f8b9b-12-A2png

conveyor-9f8b9b-12-B

conveyor-9f8b9b-12-C

conveyor-9f8b9b-12-D

conveyor-9f8b9b-12-E



@jragusa - could you show test area from above, where there is a conveyor next to a pipeline?

jeisenbe:
@jragusa - could you show test area from above, where there is a conveyor next to a pipeline, using `#a494a1` or `#9f8b9b`?

jragusa:
https://www.openstreetmap.org/#map=17/51.45170/3.72808
`#9f8b9b`
conveyor5_9f8b9b_z17

`#a494a1`
conveyor5_a494a1_z17

`#999999`
conveyor5_999999_z17

jragusa:
https://www.openstreetmap.org/#map=18/51.27466/3.84638
#9f8b9b
conveyor5-2_9f8b9b_z18

#a494a1
conveyor5-2_a494a1_z18

#999999
conveyor5-2_999999_z18

jeisenbe:
I'm okay with either option. Please pick your preference, @jragusa

jragusa:
I would suggest  `#999999` to keep an overall grey tint for all of these features. This also avoid confusion with purple tint of administrative borders.

jeisenbe:
OK, pnorman is also in favor of just using gray for now, so I will merge this PR as-is. Later we can consider changing both `man_made=pipeline` and `man_made=goods_conveyor` to a different color, at the same time, especially if the admin-boundaries hue is changed again.

PR #4113, opened 2020-04-10
Remove rendering of protect_class = 7, 24, 97, 98, 99 boundary=protected_area features

### LINKED ISSUE #3656: Different rendering of protected areas by protect_class

jeisenbe:
Needed prior to #3656

- Stop rendering boundary=protected_area with protect_class = 7,  protect_class = 24, and  protect_class = 97, 98, or 99
- Move leisure=nature_reserve selection after boundary=aboriginal_lands in `amenity-points`, to match `text-point-low-zoom` layer. 

### Explanation
- The tag `boundary=protected_area` is mostly tagged with protect_class = 1a, 1b, 2, 3, 4, 5, 6 which match the IUCN categories for "Strict Nature Reserve (IUCN Category 1a)", "Wilderness" (IUCN 1b), "National Park" (2), "Natural Monument or Natural Feature" (3), " " (4), "Protected Landscape or Seascape" (5), "Area with sustainable use of natural resources (6). 
- Sometimes these IUCN categories, from the Internation Union of Conservation Naturalists, are used with boundary=national_park or leisure=nature_reserve as well
- The proposal for boundary=protected_area also created new values of protect_class which do not match any IUCN category. The tag `protect_class=7` was for local areas which do not have an IUCN level, and 97 to 99 were for international areas. However, these tags are not well defined, and the IUCN categories actually include international parks, so these values seem to break the system.
- There is also protect_class=24 for "political protection" which has been used mainly for Indigenous Lands in Brazil and some other countries. However, the approved tag boundary=aboriginal_lands is more common and more clearly defined.
- I also noticed that boundary=aboriginal_lands is selected prior to leisure=nature_reserve in text-poly-low-zoom but afterwards in amenity-points (used for text-point), so the color of labels for areas double-tagged as aboriginal_lands + nature_reserve will change between z9 and z11. This should be fixed to keep the aboriginal_lands color at all zoom levels, since the leisure=nature_reserve tag was often added for rendering in the past.

Also see discussion in https://github.com/gravitystorm/openstreetmap-carto/issues/3656#issuecomment-609041048 and https://github.com/gravitystorm/openstreetmap-carto/issues/3656#issuecomment-609373726

See https://wiki.openstreetmap.org/wiki/Proposal:_Named_protection_class_for_protected_areas

Additional changes to the protected areas rendering should be made, but this PR is a pre-requisites to designing a sensible system to fix #3656, since protect_class 7, 97, 98, and 99 don't fit into a logical structure. 

It was also considered to remove rendering of protect_class=5 or 6, since these areas have limited protection, but in some countries National Parks are these classes and would still be rendered if tagged boundary=national_park, so they are kept for now.

There was also discussion about removing protect_class=2 since this is the usually category for boundary=national_park, but again, some protect_class=2 features are called "National Monuments" or "State Parks" or "Provincial Parks" etc, and some "National Park" (or related term) features do not have the same level of protection.

### DIFF
--- a/project.mml
+++ b/project.mml
@@ -1118,7 +1118,7 @@ Layer:
           FROM planet_osm_polygon
           WHERE (boundary IN ('aboriginal_lands', 'national_park')
                  OR leisure = 'nature_reserve'
-                 OR (boundary = 'protected_area' AND tags->'protect_class' IN ('1','1a','1b','2','3','4','5','6','7','24','97','98','99')))
+                 OR (boundary = 'protected_area' AND tags->'protect_class' IN ('1','1a','1b','2','3','4','5','6')))
             AND building IS NULL
             AND way_area > 1*!pixel_width!::real*!pixel_height!::real
         ) AS protected_areas
@@ -1465,7 +1465,7 @@ Layer:
                 'highway_' || CASE WHEN highway IN ('services', 'rest_area', 'bus_stop', 'elevator', 'traffic_signals') THEN highway END,
                 'highway_'|| CASE WHEN tags @> 'ford=>yes' OR tags @> 'ford=>stepping_stones' AND way_area IS NULL THEN 'ford' END,
                 'boundary_' || CASE WHEN boundary IN ('aboriginal_lands', 'national_park')
-                                          OR (boundary = 'protected_area' AND tags->'protect_class' IN ('1','1a','1b','2','3','4','5','6','7','24','97','98','99'))
+                                          OR (boundary = 'protected_area' AND tags->'protect_class' IN ('1','1a','1b','2','3','4','5','6'))
                                           THEN boundary END,
                 'tourism_' || CASE WHEN tourism IN ('information') THEN tourism END,
                 'office' || CASE WHEN tags->'office' IN ('no', 'vacant', 'closed', 'disused', 'empty') OR (tags->'office') IS NULL THEN NULL ELSE '' END,
@@ -1922,9 +1922,8 @@ Layer:
               'natural_' || CASE WHEN "natural" IN ('wood', 'glacier', 'sand', 'scree', 'shingle', 'bare_rock',
                                                     'water', 'bay', 'strait') THEN "natural" END,
               'place_' || CASE WHEN place IN ('island') THEN place END,
-              'boundary_' || CASE WHEN (boundary = 'protected_area' AND tags->'protect_class' = '24') THEN 'aboriginal_lands'
-                                  WHEN boundary IN ('aboriginal_lands', 'national_park')
-                                       OR (boundary = 'protected_area' AND tags->'protect_class' IN ('1','1a','1b','2','3','4','5','6','7','97','98','99'))
+              'boundary_' || CASE WHEN boundary IN ('aboriginal_lands', 'national_park')
+                                       OR (boundary = 'protected_area' AND tags->'protect_class' IN ('1','1a','1b','2','3','4','5','6'))
                                        THEN boundary END,
               'leisure_' || CASE WHEN leisure IN ('nature_reserve') THEN leisure END
             ) AS feature,
@@ -1937,7 +1936,7 @@ Layer:
               OR "natural" IN ('wood', 'glacier', 'sand', 'scree', 'shingle', 'bare_rock', 'water', 'bay', 'strait')
               OR "place" IN ('island')
               OR boundary IN ('aboriginal_lands', 'national_park')
-              OR (boundary = 'protected_area' AND tags->'protect_class' IN ('1','1a','1b','2','3','4','5','6','7','24','97','98','99'))
+              OR (boundary = 'protected_area' AND tags->'protect_class' IN ('1','1a','1b','2','3','4','5','6'))
               OR leisure IN ('nature_reserve'))
             AND building IS NULL
             AND name IS NOT NULL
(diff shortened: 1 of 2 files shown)


jeisenbe:
Any comments about removing the rendering of protect_class 7, 24, 97, 98, and 99?

Anyone have time to review this?

Prince-Kassad:
> Any comments about removing the rendering of protect_class 7, 24, 97, 98, and 99?

Looks good to me. A step in the right direction until we can figure out what to do with protected areas in general.

Stalfur:
Tread carefully with protected areas, I only just saw this. I've added comments on the wiki page.

Adamant36:
> Tread carefully with protected areas, I only just saw this. I've added comments on the wiki page.

Why? What's so special about them? They are kind of convoluted crap at the moment IMO. So, personally, I'm glad rendering was removed for the odder ones.

Stalfur:
@Adamant36 Because they have very definite legal protection. Doing something outside the protected area might be fine but do the same thing within it and you'll be liable to fines and prosecution. Just to think of one thing that might have real life repercussions.

1ec5:
[Changeset 87,776,466](https://www.openstreetmap.org/changeset/87776466) updates tags on the reservations in Oklahoma so they render again, just in time for today’s U.S. Supreme Court ruling that puts these reservations in the headlines.

Mahabarata:
Hi,

I'm not agree with the choice not to render 97 and 98 :
* a natura 2000 area is clearly a code 97 because many have no other protection in particular are not listed on the IUCN website. A natura 2000 is neither a national park nor a nature_reserve because "nature_reserve" is a value for the key "leisure" and a natura 2000 is not an area for "leisure", it is clearly a protected area and nothing else. There is a website with, for each, a map of the area, for example https://natura2000.eea.europa.eu/Natura2000/SDF.aspx?site=PTCON0002, and not to be able to see this area on osm is a problem for me : https://www.openstreetmap.org/edit?relation=9855739#map=10/41.8683/-6.8466 (the link is with zoom 10 but the area is invisible at any zoom).
* same thing for a ramsar area, with the code 98 and only this code, with this example https://rsis.ramsar.org/ris/1394 not visible on osm : https://www.openstreetmap.org/edit?way=367821180#map=13/44.9296/35.2301

I can't understand that we have a rendering for leisure=nature_reserve when, a lot of time, these areas are small, local, with nothing special scientifically speaking, and no rendering for international protected area where you can find a lot of plants and/or animals of scientific interest.

Stalfur:
I agree @Mahabarata that the people suggesting these changes have no idea of the protection and legal values of the places they removed rendering from, as I have stated here and on the wiki.

However I don't think this is a either or issue, we should not remove rendering for nature reserves when they are in fact destinations for people to enjoy nature, even if they are not harboring rare and protected species (although some do).

imagico:
I would advise again following what i said in https://github.com/gravitystorm/openstreetmap-carto/issues/4517#issuecomment-1053586789:

> If you want to suggest changes in rendering i suggest opening a new issue where you make a specific suggestion of what you would like to have changed and why. The reasoning should take the actual use of the tagging in question in OpenStreetMap into account. So if you want to suggest rendering additional protected_class=* values it is highly advisable to take a look at how these values are used practically first.

Mahabarata:
@imagico :
* this issue #4113 is still open so I imagine I can talk about it, why to open a new issue to talk about the same problem ? Jeisenbe opened it 2 years ago, nobody except Prince-Kassad answered, I answer today : is it a problem to you ?
* you say I should explain why I want to change rendering about protected_class, i'm very sad to see that you don't understand my last above comment because it is exactely what I wanted to do, sorry if my english is not so good to you.

@Stalfur : I'm not asking to delete rendering for nature_reserve, I'm just asking to add rendering for something with a bigger protection than national parks.

We have a rendering for protected_class=2 (national parks with an IUCN codification) but some national parks have no IUCN codification so there is a rendering for national_park : like that you covers every case and it is perfect.

We have a rendering for protected_class=1 and 1a (IUCN codification) but many natural 2000 areas and ramsar areas have no IUCN codification (if they had, they would have 1 or 1a), so please put a rendering for 97 and 98 because it is exactely the same that 1 and 1a protected areas when there is no official IUCN codification.

imagico:
This is a merged pull request (merged two weeks after it was opened), as such it is closed (meaning it will not be visible to people unless they explicitly look for it).  You are free to comment here but if you want to suggest a change in rendering to be made it is advisable to open a new issue.  That issue will be visible to everyone as a topic of active discussion and it can be closed if it is resolved or if it is declined.  I know this procedure might appear a bit mechanical to those who are not used to it but it helps us to manage this issue tracker.

I understood what you wrote but it is neither clear what you specifically want to change (exactly how you think `protect_class=97/98` should be rendered) nor do i see convincing reasoning under the goals of this style and based on the use of these tags in the database for that change.  That is not meant to say i don't think your request has merit but so far i lack pertinent information to assess if it has.

PR #4120, opened 2020-04-15
Render populated place without name

### LINKED ISSUE #4115: Render populated place without a name
Corresponding PR #4120.

### Expected behavior

In many areas of the world small villages have no name or an unknown name. These villages are tagged with `place` but without a `name`. These places should be rendered and be visible on the map, preferably with a small icon (a dot, a house?).

In addition to roads, showing villages makes the map more structured. I would expect that the villaged would be rendered from zoom 12 onwards (maybe depending on the.

### Actual behavior

Nothing is rendered.

At the moment, the only way view a meaningful map in those areas is to make sure every village has a mapped `landuse=residential`, and view the map at high zoom levels.

Overpass turbo link with some of these issues in Nigeria: https://overpass-turbo.eu/s/SFG

Render of the map (normally)

image

Overpass interface, showing the missing villages

image

Distribution of items tagged with `place`: https://taginfo.openstreetmap.org/keys/place#values

hiddewie:
Fixes #4115 

### Changes proposed
- New marker for places with no name
- Show marker at places with no name with the following `place` tags:
   - `city` from zoom 11 (circle with white background and a dot)
   - `village` from zoom 12 (circle with white background)
   - `town` from zoom 12 (dark dot, enlarged slightly from zoom 13 & 14)
   - `hamlet` from zoom 14 (small dark dot)
   - `isolated_dwelling` from zoom 15 (small dark dot)
   - `farm` from zoom 15 (small dark dot)
- `quarter`, `suburb` & `neighbourhood` are not included

It would be possible to include all places without names in the amenity query, but it feels off and the rendering order would be different than places *with* names.

### Test rendering:

 **Note that in this data set I added a fake city, village, town, hamlet isolated dwelling and farm without name**

(Left: current OSM, right: this PR)

Zoom 12 http://localhost:6789/openstreetmap-carto/#12/11.5344/5.6264

image

Zoom 13 (slightly bigger village dot) http://localhost:6789/openstreetmap-carto/#13/11.5178/5.6089

image

Zoom 14 (includes hamlet) http://localhost:6789/openstreetmap-carto/#14/11.5067/5.5432

image

Zoom 15 (now includes isolated dwelling and farm) http://localhost:6789/openstreetmap-carto/#15/11.5237/5.5615

image

---

By the way: thanks for the great Docker guide! This makes testing a change much easier than installing the whole stack on my host machine.

### DIFF
--- a/project.mml
+++ b/project.mml
@@ -1304,7 +1304,7 @@ Layer:
             leisure,
             name
           FROM planet_osm_point
-          WHERE place IN ('city', 'town', 'village', 'hamlet', 'isolated_dwelling')
+          WHERE place IN ('city', 'town', 'village', 'hamlet', 'isolated_dwelling', 'farm')
              AND name IS NULL
         ) AS placenames_no_name
     properties:
--- a/project.mml
+++ b/project.mml
@@ -1301,7 +1301,6 @@ Layer:
         (SELECT
             way,
             place,
-            leisure,
             name
           FROM planet_osm_point
           WHERE place IN ('city', 'town', 'village', 'hamlet', 'isolated_dwelling', 'farm')
--- a/project.mml
+++ b/project.mml
@@ -1292,6 +1292,24 @@ Layer:
     properties:
       cache-features: true
       minzoom: 12
+  - id: placenames-no-name
+    geometry: point
+    <<: *extents
+    Datasource:
+      <<: *osm2pgsql
+      table: |-
+        (SELECT
+            way,
+            place,
+            leisure,
+            name
+          FROM planet_osm_point
+          WHERE place IN ('city', 'town', 'village', 'hamlet', 'isolated_dwelling')
+             AND name IS NULL
+        ) AS placenames_no_name
+    properties:
+      cache-features: true
+      minzoom: 12
   - id: stations
     geometry: point
     <<: *extents
--- a/style/placenames.mss
+++ b/style/placenames.mss
@@ -491,10 +491,22 @@
 }
 
 #placenames-no-name {
-  [zoom >= 12] {
+  [place = 'city'][zoom >= 11] {
+    marker-file: url('symbols/place/place-no-name-city.svg');
+    marker-width: 6;
+  }
+  [place = 'town'][zoom >= 12] {
+    marker-file: url('symbols/place/place-no-name-town.svg');
+    marker-width: 6;
+  }
+  [place = 'village'][zoom >= 12],
+  [place = 'hamlet'][zoom >= 14],
+  [place = 'isolated_dwelling'][zoom >= 15],
+  [place = 'farm'][zoom >= 15] {
     marker-file: url('symbols/place/place-no-name.svg');
-    marker-line-color: @placenames-light;
-    marker-width: 5;
+    marker-width: 4;
+    [place = 'village'][zoom >= 13][zoom < 14] { marker-width: 5; }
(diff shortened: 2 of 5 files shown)


jeisenbe:
Thank you for this contribution, @hiddewie 

- The benefit to doing this would be to show a distinction between each category, so mappers could tell if there was a place=town or just a place=isolated_dwelling. So it would be better to have a subtle difference for each class of settlement that was going to get a symbol. Using the same symbol for towns, villages, hamlets and isolated dwellings is not as helpful. 

- Usually place=farm is treated identically to place=isolated_dwelling.

- I believe it will be confusing to show an SVG icon only when there is no name label. For all other features, the icon (or dot) is shown whether or not there is a name. 

- The circle is a bit heavy for hamlets and isolated dwellings. A thinner line might be better. Perhaps try varying this, or adding a central dot or square, to distinguish different values of place?

hiddewie:
@jeisenbe Great. Enough to improve. I also just found the rendering rules for the capitals in the very low zoom levels in the code.

To anwer your questions:

- I will definitely experiment with some different renderings for different kinds of places.

- I updated the query to include farms. Quarters, suburbs and others are still excluded from the query, because it gives no meaning to render those points as they should be part of another annotated place in a rural area (https://wiki.openstreetmap.org/wiki/Key:place#Populated_settlements,_urban)

- This is a good discussion, and should be discussed further. I noticed a few things as things are rendered now:
  - On low zoom levels the larger cities and capitals have an icon and a name. Then from zoom level 8 onwards, they only have a name. 
  - 'Normal', non-important cities have a name only when whey are rendered.
  - I think it would be even more confusing to render a dot with every city, village and other `place=*` that is rendered. This would clutter the map badly in densely populated places. Also, rendering a name without an icon makes the name represent the area instead of a single point on the map. This is a good thing.
  - I do agree that rendering an icon for a place without a name makes it look like it is a different 'thing' than whether or not the place has a name. Somehow the icon should indicate that it is a replacement for a name.
  - The situation that a place does not have a name will only occur in certain areas, and should be treated as an exception.

- I will vary the design of the icons!

As for the 3rd bullet point: I would be interested if there is a good reason to render dots for places *with* names.

imagico:
As mentioned in https://github.com/gravitystorm/openstreetmap-carto/issues/4115#issuecomment-612110925 i currently see no agreement on the meaning of populated place tags without a name or a consistent use of those tag for such.  Hence i don't think we should support such use through a distinct style rendering.

hiddewie:
@imagico Comment https://github.com/gravitystorm/openstreetmap-carto/issues/4115#issuecomment-612847759 indicates that there are indeed places without a name. But I agree that the use case must be agreed upon and specified in the wiki before rendering it. 

I will post an email to the OSM tagging mailing list to get a discussion going.

hiddewie:
Until there is concensus about the use case for places without names, this PR will be closed, just like the issue.

SomeoneElseOSM:
As an aside, it's worth bearing in mind that in some areas there's quite a lot of misuse of "place" with no name (I ran https://overpass-turbo.eu/s/TkM locally).  For example, https://www.openstreetmap.org/way/474771372 is unnamed and completely surrounds https://www.openstreetmap.org/node/323161416 .

It could be argued that rendering these would be a _good_ idea if it'll allow mismapping to be more easily identified.  

To be clear I'm not describing someone adding what is obviously a place=town or place=village from aerial imagery when the name is unknown as "mismapping", but some (like my example above) clearly is.

Issue #333, opened 2014-02-09
landuse=retail areas have a border, unlike other landuses


matkoniecz:
It suggests fence or some other barrier.

example: http://www.openstreetmap.org/?mlat=50.07836&mlon=19.88075#map=19/50.07836/19.88075

PR #4121, opened 2020-04-16
Add rendering for mountain_pass

### LINKED ISSUE #244: Render mountain_passes
Is it possible to render on the openstreetmap moutain_passes with symbols like >=< ?

Thanks a lot!

hiddewie:
Fixes #244

- Add rendering for nodes annotated with `mountain_pass=yes` (taginfo https://taginfo.openstreetmap.org/keys/mountain_pass)
- Added symbol for mountain passes, using grey color `#333` (text same color)
- Name is shown if available (just like saddle)
- Elevation is shown if available (just like saddle)
- If a node is both a saddle and a mountain pass, a saddle is shown
- ~~If a node is both a saddle and a mountain pass, it is shown from zoom 14 onwards, instead of zoom 15 onwards.~~
- No fancy visualization concerning the direction of the pass (`direction` tag or in line with the corresponding way).

Icon:

image

##### After

Mountain pass z15 (no saddle) (http://localhost:6789/openstreetmap-carto/#15/47.4588/7.5113)

image

Mountain pass z16 (no saddle) (http://localhost:6789/openstreetmap-carto/#16/47.4588/7.5113)

image

Mountain pass z17 (no saddle) (http://localhost:6789/openstreetmap-carto/#17/47.4588/7.5113)

image

Three mountain_passes that are also saddles (http://localhost:6789/openstreetmap-carto/#14/47.2314/7.1906)

image

Nothing changed for existing saddles (also being a mountain_pass) (http://localhost:6789/openstreetmap-carto/#17/47.24972/7.18446)

image

Some other mountain passes (makes me wonder if all mountain passes should be visualized for zoom >= 14)

http://localhost:6789/openstreetmap-carto/#15/46.1789/8.6183

image

http://localhost:6789/openstreetmap-carto/#15/46.6085/8.9415

image

http://localhost:6789/openstreetmap-carto/#15/42.8415/-0.2871

image


project.mml
```diff
@@ -1438,15 +1438,16 @@ Layer:
...
+                        OR tags->'mountain_pass' = 'yes')
                     THEN CONCAT(REPLACE(ROUND((tags->'ele')::NUMERIC)::TEXT, '-', U&'\2212'), U&'\00A0', 'm') END,
                   CASE
                     WHEN (tags ? 'height') AND tags->'height' ~ '^\d{1,3}(\.\d+)?$'
                       AND waterway = 'waterfall'
                     THEN CONCAT(ROUND((tags->'height')::NUMERIC)::TEXT, U&'\00A0', 'm') END
                   )
                 ) AS name,
-                tags->'parking' as "parking",
+              tags->'parking' as "parking",
```
jeisenbe, line 1450: This line shouldn't be changed in this PR, though you can open a different Pull Request to fix it and any other formatting problems which you notice.

project.mml
```diff
@@ -1483,6 +1484,8 @@ Layer:
                 'natural_' || CASE WHEN "natural" IN ('wood', 'water', 'mud', 'wetland', 'bay', 'spring', 'scree', 'shingle', 'bare_rock', 'sand', 'heath',
                                                       'grassland', 'scrub', 'beach', 'glacier', 'tree', 'strait', 'cape')
                                                       THEN "natural" END,
+                -- mountain_pass come after the natural=saddle, to give priority to nodes being both a mountain_pass and a saddle
```
jeisenbe, line 1487: "comes", and remove "the" before "natural=saddle". Or maybe this could be shorted into a comment on the next line like `  // after natural=saddle to give priority to that tag on the same node`

style/amenity-points.mss
```diff
@@ -1371,11 +1371,19 @@
...
+  [feature = 'natural_saddle'][mountain_pass = 'yes'][zoom >= 14] {
     marker-file: url('symbols/natural/saddle.svg');
     marker-fill: @landform-color;
     marker-clip: false;
   }
+  
+  [feature = 'mountain_pass'][zoom >= 15] {
+    marker-file: url('symbols/mountain_pass.svg');
+    marker-fill: none;
+    marker-line-color: @transportation-icon;
```
jeisenbe, line 1384: I would prefer this to be in landform-color or a neutral tone. It's not a transportation feature like a ferry, bus stop, bike share station, taxi stand, or parking lot, but rather a landcape feature: the highest elevation point where a highway crosses a ridgeline, so `@landform-color` might make sense semantically. Gray or black could also work if we consider this more like a man-made feature since the presence of a road or path is required.

jeisenbe (dismissed):
Thank you for working on this, @hiddewie. 

The CI syntax check has failed for the latest commit. It looks like there is a problem with the SVG file.

Try opening it in Inkscape again, use "Save As..." and choose the format Optimized SVG (preferable) or Plain SVG.

We had previously discussed orienting the icon to match the local direction of the road at the mountain_pass node: so if the road is traveling east-west the symbol would be rotated horizontally. This is not required, but might look better. You could give it a try if the SQL query is not too difficult.

hiddewie:
@jeisenbe Thanks, I will definitely improve the PR in response to your comment.

A question about the rotation. In comment https://github.com/gravitystorm/openstreetmap-carto/issues/244#issuecomment-456261072 it was mentioned that there is some previous work we can build upon, but could not find it in the source code. I also searched in https://github.com/imagico/osm-carto-alternative-colors. Could you give a pointer where to find that existing work?

hiddewie:
Updated the color to something dark. Looks better. Screenshots are updated

imagico:
The ac-style uses the same technique for mountain_pass as for fords:

https://github.com/imagico/osm-carto-alternative-colors/blob/56a74359fbd70cde5603627e4e76ed07333b2e60/project.mml#L1773-L1833

This is technically rather complex.

I would be against rendering mountain passes with a static symbol that would cover an area most important for reading the map.  We have this problem already with fords.  I would also be against using a symbol indicating a specific direction without that direction matching that of the road.  A rotated symbol could work but would also lead to problems with curved roads.

jeisenbe:
> A rotated symbol could work 

@hiddewie are you able to try this?

jeisenbe:
Re: problems with curved roads. It might help to use a symbol which has more space in the middle, between the two curves. This will make it less likely to overlap with a curved road or path, at most zoom levels

hiddewie:
I will definitely try adding rotation to the rendered mountain pass!

The most interesting problem will probably occur when two or more roads intersect at the mountain pass. For example a road crossing the pass, and a walking trail crossing the road exactly at the pass. In that case the direction is not clearly defined (we can take the first way, or fall back to a `direction` tag).

hiddewie:
I actually meant the previous comment as a theoretical problem, but the first rendering I see has exactly that problem (47.45925/7.51031)

image

I am pretty sure that the pass is actually north-south here (notice the two mountains on the east and west of the pass). There is no tagging data to back that up. Rotating the rendered icon in this way is wrong, just as not rotating it might be wrong.

imagico:
Yes, mountain_pass=yes on nodes where more than one path goes through or more than two start/end are inherently ambiguous.  You could try to address this partially by adding an `ORDER BY` to the subquery prioritizing paths based on their class.  But that will fail if the crest route for example is a highway=track but what goes over the pass is just a highway=path (like https://www.openstreetmap.org/node/269964518). The only reliable option you would have when using a rotated symbol would be to not rendering in such a case - which however is not necessarily intuitive for the mapper (adding a path would result in the rendering vanishing).

hiddewie:
I would actually propose to render the mountain pass icon
- with a certain direction if the direction tag is set (either a number or a bearing in degrees)
- with the direction of the road crossing the node at the path if there is exactly one way
- otherwise vertical, 0 rotation.

I think there will be few cases which would render no rotation, and having an icon is in my opinion more important than the correct direction. (I have to check the numbers on that)

mboeringa:
>     * with a certain direction if the direction tag is set (either a number or a bearing in degrees)

I would actually stick with this option, instead of fancy rotation based on path direction. Using the direction tags puts a nice positive incentive to actually add this information to the database, based on actual surveying or close inspection of aerial imagery (although the last can be a bit tricky at times). 

If the majority of the rotated mountain passes visible in the map is automatically rotated instead of based on actual data, no one will ever add it this cartographically important information to the mountain pass nodes.

And it is not that we need to adjust this continuously, once added, the direction tag stays valid essentially indefinitely (unless a catastrophic natural event).

>     * either a number or a bearing in degrees

The direction key is defined as either a number representing 0-359 degrees, or cardinal directions, see:

https://wiki.openstreetmap.org/wiki/Key:direction

jeisenbe:
The key `direction=` is rarely used with `mountain_pass=yes`: less than 1% of passes ([133 total](http://overpass-turbo.eu/s/TeX)) have this tag, and it isn't mentioned on the wiki page, so it will not help us with rendering most passes.

hiddewie:
The direction tag is mentioned in the wiki for saddle points. Maybe it could be added for mountain passes as well, as the two are geometrically closely related. But adding it in the wiki is a future solution and not a present solution.

But the question remains, is 1% of mountain passes with rotated icons considered an incentive for mapping, or is it considered a rendering problem?

mboeringa:
> But the question remains, is 1% of mountain passes with rotated icons considered an incentive for mapping, or is it considered a rendering problem?

If there is anywhere a dedicated community willing to add such data, it is hikers in the mountain regions. I do not consider 1% a rendering problem. If users become aware that their added tagging of direction leads to visually superior cartography, they will add it.

If automated routines, that are not guaranteed to get it right, try to simulate it, no incentive is left to add such useful information.

jeisenbe:
> If automated routines, that are not guaranteed to get it right, try to simulate it, no incentive is left to add such useful information.

There is actually a better way to do this automatically, demonstrated by opentopomap: use a DEM to determine the topography. Compare https://www.openstreetmap.org/#map=16/47.2501/7.1836 and https://www.opentopomap.org/#map=17/47.25011/7.18359, or https://www.opentopomap.org/#map=16/47.43778/7.40220https://www.openstreetmap.org/#map=16/47.43778/7.40220 - in these examples, the natural=saddle icon is oriented to match the contours of the topography. 

While we would not want to add a whole DEM file to this style to improve the rendering of one feature, it is quite feasible for any style which is already showing contour lines or hillshading based on a DEM. See open source code to do this at https://github.com/der-stefan/OpenTopoMap/blob/master/mapnik/tools/saddledirection.c

So there isn't a need to make mappers add a `direction=` tag, in most cases.

mboeringa:
> So there isn't a need to make mappers add a `direction=` tag, in most cases.

I understand, but it still isn't "mapper feedback". In fact, it is deceptive, because it suggests there is something that isn't there.

jeisenbe:
>  it suggests there is something that isn't there.

Using the DEM might be misleading, but using the direction of the `highway=` way is appropriate: that is information in the OSM database.

mboeringa:
>  is appropriate: that is information in the OSM database.

I am not challenging that is appropriate or feasible, both potentially are as you justly indicate.

But it still isn't mapper feedback about the status of having a direction key on the mountain pass node or not.

jeisenbe:
> I would actually propose to render the mountain pass icon
> 1 with a certain direction if the direction tag is set (either a number or a bearing in degrees)
> 2 with the direction of the road crossing the node at the path if there is exactly one way
> 3 otherwise vertical, 0 rotation.

Since only 0.5% of these nodes have a direction tag and it is not documented, I don't think that step 1 is necessary.

Number 2 is good, if possible.

Number 3 might work, but we will need to test how it looks. It might be better to orient the pass to the higher class of highway, when relevant, as mentioned in  https://github.com/gravitystorm/openstreetmap-carto/pull/4121#issuecomment-618355123

imagico:
It would be good to cut the tagging discussion here because as @jeisenbe said current use of that combination is negligible and undocumented.  Also semantics are not at all trivial and in case of the popular combination of mountain_pass=yes + natural=saddle ambiguous.  We are not going to try steering mappers to add certain data because it seems to simplify things for us - especially if the solution this promises would ultimately be quite mediocre.

Regarding the rendering of mountain passes - i see mainly the following options:

* using a very compact abstract and non-oriented symbol to indicate a mountain pass without obscuring a lot of the map in the area and without potentially blocking a lot of other symbols.  Because of the widespread use of the combination of mountain_pass=yes + natural=saddle this would also call for handling the double tagging in an intuitive way.
* finding consensus on using a bridge like rendering.  This isn't rocket science - though as said not trivial.  It would have the advantage of being equally usable for fords - which are a long standing problem in this style.  And it would have tons of advantages otherwise - like the option to intuitively handle the complex cases of junctions on the pass and that it would not need to block other symbols or labels around.

I don't see that the idea of a rotated static symbol is ultimately going to work well - no matter how you source the rotation.  Surely you would for any variant of it find cases where it looks good but at the same time you will also find tons of cases where it does not.  Taking care of that would quickly result in a lot of special casing and different symbols for different road types and zoom levels and ultimately probably a higher complexity in code than the bridge like solution.

mboeringa:
Well, for anyone who would like to join me, here is now a MapRoullete task to add the direction tag to mountain passes. iD's option to show a direction on nodes, is really helpful as well:

https://maproulette.org/browse/challenges/13205

mboeringa:
By the way, this is why you need to be careful making assumptions based on path direction. Notice how the path follows a ridge, and the actually tagged mountain pass is almost perpendicular to it as a consequence:

afbeelding

jeisenbe:
Then that’s a natural=saddle, not a mountain_pass=yes

> [image: afbeelding]
> <https://user-images.githubusercontent.com/7635726/80222594-6267ef00-8647-11ea-89e6-698d84e0c5c7.png>
>

StyXman:
>  Also semantics are not at all trivial and in case of the popular combination of mountain_pass=yes + natural=saddle ambiguous

> Then that’s a natural=saddle, not a mountain_pass=yes

I have the intuitive feeling that most mountain passes are built in saddles (but definitely not most saddles have mountain passes as @jeisenbe says). I wouldn't think much about it.

hiddewie:
### OpenTopoMap examples

> Col de Fuet https://github.com/gravitystorm/openstreetmap-carto/pull/4121#issuecomment-618887861

https://www.opentopomap.org/#map=16/47.25084/7.18120 barely shows the icon because the icon alighs with the road and there is no text.

> Blauenpass https://github.com/gravitystorm/openstreetmap-carto/pull/4121#issuecomment-617302064

https://www.opentopomap.org/#map=16/47.45826/7.50961 nothing is rendered

image

### Bridge like rendering & taking ways into account for direction

The more I think about the property of saddle points and mountain passes, and taking all the comments above into account, the more I feel that paths, roads and other ways have little to do with the orientation of the pass or the rendering of it.

That also means that bridge like rendering (https://github.com/gravitystorm/openstreetmap-carto/pull/4121#issuecomment-618921694) for roads leading up to a mountain pass seems unusual. For fords, this would be more logical, because the way actually crosses something wet.

So I would agree with the first option of https://github.com/gravitystorm/openstreetmap-carto/pull/4121#issuecomment-618921694, and try a non-oriented icon which does not need rotation.

> https://en.wikipedia.org/wiki/Mountain_pass: A mountain pass is a navigable route through a mountain range or over a ridge.

> https://wiki.openstreetmap.org/wiki/Key%3Amountain_pass: mountain_pass=yes marks the highest point of a mountain road, railway or path as it passes over a crest. [...] A mountain_pass=yes tag should be the highest point of the way, so it's a single node on that way. [...] Be aware that highways don't always go exactly through the saddle point, so that the highest point on the highway and the saddle point can have different positions.

There is no need for a mountain pass to be built in a topographic saddle. It only has to cross a ridge as the highest point of a way.

hiddewie:
image

Non rotated mountain pass icon proposal (own creation)

hiddewie:
The PR and description + screenshots have been updated with the proposed icon.

jeisenbe:
At 14 pixels square the proposed icon is not very clear. Also, it looks larger and heaviery than the `natural=saddle` or `natural=peak` icons.

hiddewie:
Yes, it looks a bit heavier. But I actually like the visualization at 14px. Compare it to the castle icon or the alpine_hut. Maybe the natural peak and saddle should be enlarged.

I might lighten the color a bit, but this would reduce the readability. Or the icon should have some (light colored) bleed, although this is against current practices in this cartography style.

jeisenbe:
> Compare it to the castle icon or the alpine_hut.

The difference is that a `mountain_pass=yes` is a feature found on a `highway=` such as a path or road, making it more similar to a bridge or tunnel (which are rendered by changing the highway rendering), gate or toll booth (which have quite small symbols). Unlike a gate or toll booth, a pass is not a point barrier, but rather a routing way-point and sometimes a named feaure. 

We don't want a big, heavy symbol which obscures the geometry of the road as it goes through the pass, since many passes are on tight curves, and sometimes they have junctions between a path or road along the ridgeline and the path/road which crosses the ridge. 

Previously, another maintainer of this stylesheet, @imagico, suggested these 2 options which he would be willing to approve:

> 1) using a very compact abstract and non-oriented symbol to indicate a mountain pass without obscuring a lot of the map in the area and without potentially blocking a lot of other symbols. 

This might just be a round dot or circle, or a smaller version of the `) (` type symbol above.

> 2) using a bridge like rendering

Adamant36:
Option 2 of using a bridge like rendering is always how I imagined it. It just seems intuitive for some reason. Whereas, a symbol doesn't.

hiddewie:
> Unlike a gate or toll booth, a pass is not a point barrier, but rather a routing way-point and sometimes a named feaure.

I do not agree. The mountain pass itself is a single point along the way, exactly at the point where the road reaches its highest point. In particular, it is usually marked by a large stone, a post, a traffic sign or an information sign. 

As for obscuring the map: the text and elevation information of the mountain pass will obscure more than the icon.

I also want to stress that a mountain pass may sometimes be as important as a city or town on a map, because this is used for selecting the way to go from A to B. And not only for roads; in particular when hiking in the mountains, you usually go from hut to pass to hut (and to a mountain top). These features are more important than saddles.

Along roads, there is usually a restaurant or some other facilities because it is the highest point with a nice view. Those facilities are less important than the mountain pass itself, but should not be obscured in high zoom levels.

@jeisenbe Do you have some examples of complex moutain pass road stuctures that are obscured by the icon? I want to render these areas with the current code, and share them here.

---

Some examples of single points indicating a mountain pass, not a section of a way:

[Blauenpass](https://eu-browse.startpage.com/av/anon-image?piurl=https%3A%2F%2Fwalterdiener.ch%2FGallerien%2FGallerien%25202014%2F03.01.2014%2520Flueh%2520-%2520Choepfli%2520-%2520Blauenpass%2520-%2520Blattenpass%2520-%2520Aesch%2Falbum%2Fslides%2FDSCN1374.jpg&sp=1588419792T91dbdc40dc72dcd46e6ab7cea1c78a95c3faa7ba9fc2aa7b57ff23dee71d30d5)

[Col de Jau](https://eu-browse.startpage.com/av/anon-image?piurl=https%3A%2F%2Fcyclingthecatharcountry.files.wordpress.com%2F2016%2F05%2Fimg_4181.jpg&sp=1588419897Tb47ecc5b59dd518a2d74418e159113ffd21504744ab6f2c9ca84e77488339c98)

[Passo Greina](https://upload.wikimedia.org/wikipedia/it/e/ed/Passo_della_Greina.jpg)

imagico:
> As for obscuring the map: the text and elevation information of the mountain pass will obscure more than the icon.

No, because the label would be off center and the actual location of the mountain pass would not be obscured.

Use of large pictorial point symbols for identifying mountain passes is very rare in maps in general - i know of no classically produced maps that do this and even in automatically rendered digital maps it seems - where it is done - a technical convenience choice rather than a conscious cartographic decision.

hiddewie:
OK, with the `) (` style icon it looks like:

image

(http://localhost:6789/openstreetmap-carto/#15/46.6086/8.9406)

hiddewie:
And with some pointy ends (I like this)

image

Icon (own creation)

image

jeisenbe:
@imagico would you be able to review this cartographically?

Adamant36:
> adjusting the starting zoom level of natural=saddle depending on the presence of mountain_pass=yes without showing the double tagging in rendering does not seem a good idea.

I totally agree with that.

hiddewie:
- I removed the differentiating zoom level logic for saddles that are also a mountain pass. Mountain passes and saddles are both rendered from zoom 15 now.
- Color of the icon + text: https://github.com/gravitystorm/openstreetmap-carto/pull/4121#discussion_r409986372 indicated that a darker non-transportation color would be better. I agree, but I am willing to change it to a different color if there is enough support.

mboeringa:
> The difference is that a `mountain_pass=yes` **is a feature found on a `highway=` such as a path or road**, making it more similar to a bridge or tunnel (which are rendered by changing the highway rendering), gate or toll booth (which have quite small symbols). Unlike a gate or toll booth, a pass is not a point barrier, but rather a routing way-point and sometimes a named feaure.

One more general remark about features tagged with **mountain_pass=yes**: from what I have seen up to now, there are quite a few **mountain_pass=yes** features, that don't have a **highway=x** running through them, e.g. in Russia.

While my first instinct was to assume that these nodes were falsely tagged mountain_pass=yes, and should be re-tagged to natural=saddle only, I have come to the conclusion that this is to simple an assumption.

Looking at aerial imagery, I have the distinct feeling that many of these features, sometimes based on official government registries of the corresponding country, are what I would like to call **"_potential mountain passes_"**, that is, the saddle visible in the imagery and tagged with **mountain_pass=yes**, appears to be _**passable**_ by a human without specialized climbing gear, e.g. ropes etc. 

So, while it is obvious that these "mountain passes" are not in active use today, they could be used during e.g. a mountain rescue operation (or war time...).

hiddewie:
@jeisenbe No problem. The PR is still open, and I am willing to update it to any wishes :)

I think that rotating the icon conditionally would be possible, but would give a bad result because many passes contain more than one road passing through. Defaulting to an 'upright' position would give a false impression that the pass is in the north-south direction because other mountain pass icons are be rotated. Furthermore, as mentioned in https://github.com/gravitystorm/openstreetmap-carto/pull/4121#issuecomment-619039860, there are examples that only contain one highway passing through the pass, but the direction is different from the highway direction. The only reason I would show a rotated icon is combined with a `direction` tag, but this is not a solution as commented in https://github.com/gravitystorm/openstreetmap-carto/pull/4121#issuecomment-618810055.

As for the icon, I will try to make a proposal with a small abstract icon that does not suggest a direction. I just have to come up with a suitable image (suggestions are welcome).

I will experiment with the existing colors you mentioned. :+1:

hiddewie:
Some examples with landform color and new icon

Icon (abstract enough?)

image

image

image

image

image

hiddewie:
`transportation-icon` color:

image

hiddewie:
In https://github.com/cyclosm/cyclosm-cartocss-style/pull/462#issuecomment-749631404 it was suggested that the mountain pass could also be rendered with the saddle icon with a transport or dark color. The CyclOSM style currently uses that solution. 

They use no rotation of the mountain pass icon, which I agree with.

imagico:
I would definitely think reusing the same symbol as used for saddles in transportation color is a better option than what has been suggested otherwise so far.  I still think an integrated bridge-like rendering would offer better readability and be more intuitive but as already said i am not tied to this - if others think a technically simple and non-contextualized static point symbol is the way to go i am not opposed to that.

hiddewie:
:+1: I will update the PR to use the saddle icon in a suitable color.

hiddewie:
The screenshots in the PR description have been updated, and old comments have been collapsed with a `<details>` block.

style/amenity-points.mss
```diff
@@ -1847,7 +1847,7 @@
     text-line-spacing: @standard-line-spacing-size;
     text-fill: darken(@landform-color, 30%);
     [feature = 'natural_volcano'] { text-fill: #d40000; }
-    [feature = 'mountain_pass'] { text-fill: @mountain-pass; }
+    [feature = 'mountain_pass'] { text-fill: darken(@mountain-pass, 30%); }
```
jeisenbe, line 1850: Please change this to use `@transportation-text` instead, so it will match the text for other `transportation-icon` colored symbols

style/amenity-points.mss
```diff
@@ -1383,9 +1383,9 @@
     marker-fill: @landform-color;
     marker-clip: false;
   }
-  
+
   [feature = 'mountain_pass'][zoom >= 15] {
-    marker-file: url('symbols/mountain_pass.svg');
+    marker-file: url('symbols/natural/saddle.svg');
     marker-fill: @mountain-pass;
```
jeisenbe, line 1389: I think we can just simplify this to `@transportation-text` and skip the intermediate color name.

jeisenbe (changes requested):
I’ve checked the rendering. The text color needs to be changed to match other `transportation-icon` features, but otherwise it is ok.

(This change reminds me that it would be better to change our transportation-icon and transportation-text colors to be something different than this shade of blue, since they are rather close to water features, but that should not need to be addressed in this PR)

Examples:
Screen Shot 2021-01-24 at 15 37 34

Screen Shot 2021-01-24 at 15 38 46

Note that nodes with natural=saddle + mountain_pass=yes are not changed by this PR, but are still rendered in the landform color like mountain peaks.

hiddewie:
@jeisenbe Thanks. I updated the code as per your comments.

The saddles are not modified to let the existing rendered features on the map remain the same. Also ref https://github.com/gravitystorm/openstreetmap-carto/issues/244#issuecomment-456261072 that saddle is more 'important' to render than a moutain pass.

Extra screenshot for verification:

image

style/amenity-points.mss
```diff
@@ -1383,6 +1383,12 @@
     marker-clip: false;
   }
 
+  [feature = 'mountain_pass'][zoom >= 15] {
+    marker-file: url('symbols/natural/saddle.svg');
+    marker-fill: @transportation-text;
```
imagico, line 1388: `transportation-icon` is the color for symbols, `transportation-text` is the color for the corresponding labels:

https://github.com/gravitystorm/openstreetmap-carto/blob/faead203a2fb4400050d59ac2d5ded7ff5bcd003/style/amenity-points.mss#L5-L6

hiddewie:
Yes ofcourse, I missed the correct name.

image

jeisenbe:
@imagico does this work for you cartographically? 

It would be better if `@transportation-icon` were not blue, though at least passes are always at ridgelines and therefore are less likely to be confused with water features such as fords or springs, when the context is clear.

imagico:
> @imagico does this work for you cartographically?

As said i would prefer a contextualized, bridge-like rendering and i also think we need to move towards such methods to maintain a rich style in the future without cluttering the map - see #3635.

But as i also said within the spectrum of static point symbols this is the best solution IMO.  So if it otherwise finds support i will not oppose it.  But it should be clear that it would further add to #3635 and not be a step towards solving it.

ppete2:
I really like the result how this PR has been developed. It's nice to stick to the saddle-like icon with the blue transport color. It's also good, not to keep with the idea to use a rotated, bridge-like icon. I think it would have been to confusing for non-map experts reading the map - especally on passes where multiple ways are crossing.

I've though about the situation if a node shares both `natural=saddle` + `mountain_pass=yes` . And I accord with the current solution to render as saddle, to illustrate the saddle-property of a pass. Since usually not each pass is also a saddle, but each saddle along a highway is a pass too.

So for me this PR would be ready to publish.

jeisenbe (approved):
While this PR is not ideal (I would still prefer a rendering that emphasized the orientation of the road or path through the mass), it is acceptable. I’m inclined to merge this, though will wait to see if there are any additional comments or reviews

hiddewie:
Perfect, thanks!

PR #4141, opened 2020-05-08
Icon proposal for amenity=carpet_washing

xkr47:
Fixes -

Add icon for amenity=carpet_washing

Before https://www.openstreetmap.org/node/829492460

After -


jeisenbe:
Thank you for contributing, @xkr47 

The tag `amenity=carpet_washing` has been used a couple dozen times, but it does not have a definition or description: https://taginfo.openstreetmap.org/tags/amenity=carpet_washing

Usually, a tag should be well-defined, accepted by the community and widely used, before it is rendered in this style. 

Consider discussing this tag on the [Tagging mailing list](https://lists.openstreetmap.org/listinfo/tagging) and then documenting it on the OpenStreetMap wiki with a proposal page or Tag page.

xkr47:
Thanks for your feedback! I did create the page https://wiki.openstreetmap.org/wiki/Tag:amenity=carpet_washing but it's possible I actually posted this PR before that. It seems taginfo has picked up at least the headline now.

As you suggest maybe I should direct further discussion to the mailing list, but I'll post a brief update here for the record;

I looked around a lot at the current tags in use for the purpose of publicly available carpet washing facilities, and there is a plethora of creative tagging like (:heavy_check_mark: = documented tag):
* amenity=carpet_wash
* amenity=carpet_washing
* amenity=drinking_water :heavy_check_mark: 
* amenity=lavoir :heavy_check_mark: 
* amenity=matonpesupaikka (finnish translation of carpet_washing_place)
* amenity=outdoor_carpet_wash
* amenity=washing
* landuse=yes :heavy_check_mark:
* leisure=park :heavy_check_mark:
* shop=laundry (even if there is no actual shop or building)

These are often combined with name=* describing the actual purpose of the feature.

Of the three (undocumented) tags with clear association with carpet washing, the `carpet_washing` was the most used one by far, so after talking briefly with a seasoned mapper on #osm-fi irc channel and reading up on adding new tags, I documented it.

PR #4153, opened 2020-06-02
get-external-data.py: store downloaded archives for further use

jacekkow:
This allows to recreate the database objects without the need
to redownload all the (not so small) files.

Additionally it outputs stderr of `ogr2ogr` on error (which was collected but not displayed).

Motivation:
I got hit by https://github.com/OSGeo/gdal/issues/1692 and every time I re-ran the get-external-data.py script it tried to download ~600 MB of data (water-polygons-split-3857.zip).

### DIFF
--- a/scripts/get-external-data.py
+++ b/scripts/get-external-data.py
@@ -290,6 +290,7 @@ def main():
                         logging.critical("Command line was {}".format(
                             subprocess.list2cmdline(e.cmd)))
                         logging.critical("Output was\n{}".format(e.output))
+                        logging.critical("Error was\n{}".format(e.stderr))
                         raise RuntimeError(
                             "ogr2ogr error when loading table {}".format(name))
 
--- a/scripts/get-external-data.py
+++ b/scripts/get-external-data.py
@@ -22,7 +22,6 @@ import shutil
 # modules for getting data
 import zipfile
 import requests
-from requests_file import FileAdapter
 import io
 
 # modules for converting and postgres loading
@@ -138,7 +137,6 @@ class Table:
 class Downloader:
     def __init__(self):
         self.session = requests.Session()
-        self.session.mount('file://', FileAdapter())
         self.session.headers.update({'User-Agent': 'get-external-data.py/osm-carto'})
 
     def __enter__(self):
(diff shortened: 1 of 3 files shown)


jeisenbe:
@pnorman were you planning to review this?

pnorman:
I would prefer to not have it save files by default.

jeisenbe:
@pnorman if I understand your comment correctly, you would prefer to reject this PR... or are you requesting changes?

pnorman:
> @pnorman if I understand your comment correctly, you would prefer to reject this PR... or are you requesting changes?

I'd prefer to reject it, but if others feel the need for a save download file option, it should be a non-default option.

imagico:
I would support having a local file storage and reuse option as non-default functionality.  If you have a setup with multiple test databases and switch between them via localconfig.json reusing downloads is convenient.  As well as an option to create the tables without any data.

pnorman (changes requested):
We can add this functionality, but it should be off by default.

pnorman:
@jacekkow, are you interested in making the changes above? note that #4468 may impact the need for this.

jacekkow:
Hi, I completely forgot about this Pull Request. I'll update it in a moment.

jacekkow:
As a first step I have refactored whole download logic into a separate class (Downloader). I'll introduce caching there (maybe removing the requests-file dependency in the process).

Please check if such changes are fine by you.

pnorman:
I think it still needs to work with a file URL

jacekkow:
@pnorman: We didn't understand each other - by "removing the requests-file dependency" I meant only the dependency, not the support for file:// URLs itself.

This is what I meant:
https://github.com/jacekkow/openstreetmap-carto/commit/bef7e38c6411515c47ebb083f871947d964b7dfd

Caching is now behind a feature flag:
https://github.com/jacekkow/openstreetmap-carto/commit/09453c4c29d6266c1d212a08caae7f9cfea5fa8d

Please verify whether this is what you wanted.

scripts/get-external-data.py
```diff
@@ -135,13 +134,95 @@ def replace(self, new_last_modified):
...
+        return self
+
+    def __exit__(self, *args, **kwargs):
+        self.session.close()
+
+    def _download(self, url, headers=None):
+        if url.startswith('file://'):
+            with open(url[7:], 'rb') as fp:
+                return DownloadResult(status_code = 200, content = fp.read(),
+                                      last_modified = str(os.fstat(fp.fileno()).st_mtime))
```
pnorman, line 152: Why this over requests_file?

jacekkow: To eventually support last_modified with local files, too, as requests-file does not return the modification date at all.

Note it is not there yet, but I may implement in an extra commit this weekend if you'd like to see it.

pnorman: Ah, that makes sense.  I'll give the rest of this a fuller review.

pnorman (approved):

Issue #3737, opened 2019-04-05
Adding specific rendering for the new office=diplomatic (new tag for embassies and consulate)


lgiard:
Hello, 

Since the **new method of tagging embassies and consulate** ([with office=diplomatic](https://wiki.openstreetmap.org/wiki/Tag:office=diplomatic)) has been approved end of last year ([look at the accepted proposal](https://wiki.openstreetmap.org/wiki/Proposed_features/office%3Ddiplomatic)), it should be useful to support the rendering for the new tagging scheme. The old tag "amenity=embassy" is now in the process of [being deprecated](https://wiki.openstreetmap.org/wiki/Tag:amenity%3Dembassy). 

I would say that using the **same icon than "amenity=embassy"** should be enough :
-> ![embassy_logo](https://wiki.openstreetmap.org/w/images/3/36/Embassy-16.svg)

Because I don't think it is absolutely needed to visually differentiate between embassy, consulate and liaison (the three sub-category of this new tagging scheme via the sub-tag _diplomatic=*_). It is generally described in the name anyways.  

Concerning the old tag "amenity=embassy", I don't know if we generally keep the rendering of the "deprecated tag" (like during a transition phase) or remove it directly from rendering ?

imagico:
Thanks for pointing this out.

In general proposal votes and deprecating of tags in them has no direct effect on our decisions here.  The numbers for the tags in question are quite clear, there are 12k features with `amenity=embassy`, 600 with `office=diplomatic` and 5.2k with `diplomatic=*`.  Should this change significantly we could reconsider this.

Also on a general note we do not typically render several different tags for indicating the same type of feature because this would perpetuate the use of both tags (a.k.a. tagging proliferation) rather than supporting a specific consistent tagging.  If it is clear that mappers are adopting a new tagging in favor of an existing one we'd move by removing the old one and at the same time adding the new one.

matkoniecz:
I think that for now we should wait whatever the new scheme will be adopted by mappers. Not that `office=diplomatic` is rendered.

I would suggest to ask for support in editors first (if it is not supported and not rejected). Note also there are validators - standalone and built-in within JOSM and iD.

If there is consensus among mappers to promote `office=diplomatic` then maybe validators can suggest mappers to use this tag in addition/instead of `amenity=embassy` (or reverse, suggest to add/replace with ` amenity=embassy` if there is no consensus that new tag makes sense).

[...]

PR #4168, opened 2020-06-22
moving from amenity=embassy to office=diplomatic and differentiating embassy and consulate

### LINKED ISSUE #3886: amenity=embassy is deprecated, show office=diplomatic or diplomatic=consulate with the flag symbol instead
### Expected behavior
Diplomatic offices such as embassies, consulates etc (see [office=diplomatic](https://wiki.openstreetmap.org/wiki/Tag:office%3Ddiplomatic) for full list) without the amenity=embassy tag should be rendered with the flag symbol
### Actual behavior
Only amenity=embassy is rendered, however, this tag is [deprecated](https://wiki.openstreetmap.org/wiki/Tag:amenity%3Dembassy)
https://www.openstreetmap.org/node/6665879459 is rendered, but tagged with the deprecated amenity=embassy
https://www.openstreetmap.org/node/2953707091 is not rendered, but tagged correctly

(edit: corrected first link)

imagico:
Fixes #3886

Removes rendering of `amenity=embassy` and adds rendering of `office=diplomatic` + `diplomatic=embassy` and `office=diplomatic` + `diplomatic=consulate` using distinct symbols.  Any other `office=diplomatic` will remain being rendered like other offices with a dot.

Symbol and label color is the same as for the office dots.

Reasoning for the switch of removing support for `amenity=embassy` and adding support for the new tagging:  The new tagging has proven to have replaced the old tagging in practical mapping and is broadly being adopted by mappers.  Of the 6200 remaining amenity=embassy features nearly 1500 have also an `office=diplomatic` tag.  Currently >6000 features would be rendered in the distinct forms introduced by this change, another ~1500 `office=diplomatic` would remain being rendered with the plain office dots.

Having a distinct symbol for `diplomatic=consulate` helps avoiding abuse of `diplomatic=embassy` as tagging for the renderer.  Other values of `diplomatic=*` are either not widely used or are not consensus among mappers.

Examples:

embassy
embassy


grischard:
It’s always cool to see something you’ve mapped becoming an example in a PR :)

I’ve certainly used the other values for diplomatic, and am a bit concerned that only rendering some will encourage mistagging, say honorary consulates as consulates or liaison offices as embassies. What about rendering diplomatic=* with flag icon?

grischard:
Actually, the first example should probably be permanent_mission since it represents Luxembourg not to a country but to an organisation. I’ll fix my tagging.

grischard:
it’s always cool to see something you’ve mapped used as an example in a PR :)

kocio-pl:
I like this proposition in general, however I see two things to be discussed:

1. The consulate icon proposition includes a person - what is the rationale for this? Wouldn't a smaller flag be enough here?

2. Even if the old embassy tag is deprecated and losing ground, it's still both over our traditional limits for rendering features and like a half of its top amount, so still a lot:

taghistory(4)

Removing it right now sounds like an invitation for a mass retagging, which would be bad for the database quality, of course. I think that it'd be much safer to wait until it goes below ~2k uses.

imagico:
The reason for displaying a person with `diplomatic=consulate` is to indicate this provides a public service of some sort.  There are other classes of diplomatic offices that do not and might be in the future differentiated by a symbol without a person.

As indicated in #3886 keeping the previous rendering while introducing a new rendering for the new tagging scheme is out of the question - it would be confusing for mappers and would be bad cartography - short:  It would be against several of our documented goals here.

kocio-pl:
I have never seen any documented goal to do changes in one step. It's better to ask people what they will find better than claim for them. I personally see no more confusion there than in the whole idea of migrating things and avoiding automated mass changes, which is crucial issue when designing tags - nobody likes that complication, but it's there for a reason.

I'd like to hear some more details about symbol design. So embassies are just general representation of the country, while consulates are different than them and deals more with people? I'm not sure that is the main difference. Smaller flag for consulates makes sense for me, since they clearly subordinated.

imagico:
I am not going to revisit the question of rendering tags as synonyms here - we have done that plenty of time.  If you want to re-visit that topic please open a separate issue and state any new arguments.

In any case what is important here is that there is no consensus on adding distinct rendering for office=diplomatic without at the same time (or before) removing amenity=embassy.  If it is too early for that change is something that can be discussed obviously.

The difference between an embassy and a consulate is that an embassy is an official representation of a country abroad.  It serves diplomatic purposes mostly.  A consulate mostly provides services to either the host or the home country citizens.  Hence it is more service oriented and therefore the symbol with a person.  Most embassies also provide consular services but tags to indicate that are so far not very well established.

imagico:
Ok, there do not seem to be any objections to the design aspect of this change.  I would like to make a decision to either merge this or to close this because we are not ready to move to the new tagging scheme.  I think these are the only two directions for which consensus is achievable.  Any comments on which of these is preferred is welcome.

jeisenbe:
I'm in favor of this idea. While I have personally felt ambivalent about the switch to `office=diplomatic`, it appears the mappers have been willing to change the tags and there is a clear trend toward use of the new scheme. Since the main benefit of this change is to distinguish consulates from embassies, this PR is the right idea. However, I have not reviewed the PR technically, yet. Perhaps I will have time in a few days.

kocio-pl:
> I am not going to revisit the question of rendering tags as synonyms here - we have done that plenty of time. If you want to re-visit that topic please open a separate issue and state any new arguments.

I don't remember any other argument than "it's been documented", so it would be good to prove this, since you brought it here yourself.

On the other hand I don't remember you referring to the argument about mass edits, so it would also be good to hear that.

> If it is too early for that change is something that can be discussed obviously.

OK, I asked the question here:

https://lists.openstreetmap.org/pipermail/talk/2020-July/085115.html

> Hence it is more service oriented and therefore the symbol with a person. Most embassies also provide consular services but tags to indicate that are so far not very well established.

I understand this, but the problem is that's not a clear difference in reality, and I think just a smaller flag would be enough, since it also changes symbol proportions. Maybe making a gap a little bigger could be used to make it more distinct.

imagico:
Note i have no strong desire to see this change being merged - if the general feeling is that it needs more time for mappers to adopt the new tagging until we can do this that is fine with me.  I just saw strong desire being expressed to render the new tagging and wanted to show and offer a suitable implementation for this that provides useful feedback to mappers to demonstrate that it is not the lack of a viable design solution that prevents addressing #3886.

Note right now we have - as mentioned on talk - 4611 features with amenity=embassy only - that is down from 6692 at the beginning of the year according to https://github.com/gravitystorm/openstreetmap-carto/issues/3886#issuecomment-569810310.

During the same time span the use of diplomatic=embassy/consulate (which this change would add being rendered in a distinct and feedback providing form) rose from 2769+927 to 4901+1905 - i.e. it almost doubled.  Given that embassies and consulates are probably a class of objects where the vast majority of real world features that exist are already mapped in OSM the usefulness of feedback through the map is highest while mappers adopt the new tagging and assess the nature of existing features for the new tagging scheme.  This is different from previous changes like landuse=farm/landuse=farmland or highway=ford/ford=yes where the usefulness of mapper feedback always was and still is predominantly in the future.

I don't think differentiating embassies and consulates with just a large/small flag symbol would be desirable.  This distinction as explained is not one of size or importance, it is one of qualitatively different function.

jeisenbe:
@imagico - please merge this if you agree

kocio-pl:
What I miss in this design is lack of clear extensibility - they are carefully crafted for exactly 2 cases. This is not a blocker for me, but do you think it could be extended for all the other diplomatic offices?

Since nobody on the list commented for a long time, I think it's safe enough to test this change and see how it impacts tagging changes.

imagico:
> What I miss in this design is lack of clear extensibility - they are carefully crafted for exactly 2 cases.

I am not sure what you mean by this.  Could you point to POI symbol additions from the past 2-3 years where you see a clearer extensibility than with this?

kocio-pl:
No, I was not thinking about anything else, I just focused on this proposition.

imagico:
As i said i am not sure what you mean with that remark.  An example would have helped.

grischard:
The world of diplomatic structures tends to be relatively static. If another type of `office=diplomatic` comes along one day and becomes significant enough, we can amend the rendering to add it. I'm glad this is now merged.

I'm also amused that @jeisenbe has also used something I've mapped as an example. Are you guys following me? :)

imagico:
Luxembourg is a popular test area for testing in central Europe since the country extract is small and does not take a lot of space in a test database.  I used Strasbourg as sample because i have no national capital in my test database and therefore Strasbourg was about the only location with a feature tagged diplomatic=embassy.

jeisenbe:
Also fixed #3737

PR #4174, opened 2020-07-03
Solve overlapping of amenity=prison over natural features

### LINKED ISSUE #3841: Render amenity=prison over natural features
### Expected behavior
Render amenity=prison over natural features, like a landuse=military.

### Actual behavior
Anotação 2019-08-11 130503
Natural features overlap the amenity=prison.
https://www.openstreetmap.org/way/205946580#map=15/-20.3745/-40.5007

LorenzoStucchi:
Fixes #3841

- correct rendering of amenity=prison over natural elements

Before
immagine

After
solved_issue

The original data that cause the problem have been delated, the test could be done with these data. 
[prison.zip](https://github.com/gravitystorm/openstreetmap-carto/files/4869761/prison.zip)

### DIFF
--- a/project.mml
+++ b/project.mml
@@ -92,7 +92,7 @@ Layer:
               way, COALESCE(name, '') AS name,
               ('aeroway_' || (CASE WHEN aeroway IN ('apron', 'aerodrome') THEN aeroway END)) AS aeroway,
               ('amenity_' || (CASE WHEN amenity IN ('bicycle_parking', 'motorcycle_parking', 'university', 'college', 'school', 'taxi',
-                                                    'hospital', 'kindergarten', 'grave_yard', 'prison', 'place_of_worship', 'clinic', 'ferry_terminal',
+                                                    'hospital', 'kindergarten', 'grave_yard', 'place_of_worship', 'clinic', 'ferry_terminal',
                                                     'marketplace', 'community_centre', 'social_facility', 'arts_centre', 'parking_space', 'bus_station',
                                                     'fire_station', 'police')
                               OR amenity IN ('parking') AND (tags->'parking' NOT IN ('underground') OR (tags->'parking') IS NULL) THEN amenity END)) AS amenity,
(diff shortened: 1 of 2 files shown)


imagico:
Thanks for the pull request.

Personally i don't like the idea of transparent overlays with line hatching obscuring the underlying map through color mixing so i would rather think about getting rid of this than extending it.  Especially since this implies to the map user that these tags are something else than normal landuse tags.  But if others think extending this from of rendering to prisons is a good idea i won't object.

Technically there are a number of issues with this PR though:

* you don't remove the existing rendering of `amenity=prison` so it is rendered twice - that is probably not a good idea.
* you interpret `natural=scrub` - apparently to specifically handle the example case you demonstrate.  That is not a good idea.  We don't want to special case specific combinations of semantically unrelated tags.

It would be good in general to test in an area of correct tagging based on current data.

LorenzoStucchi:
> Personally i don't like the idea of transparent overlays with line hatching obscuring the underlying map through color mixing so i would rather think about getting rid of this than extending it. Especially since this implies to the map user that these tags are something else than normal landuse tags. But if others think extending this from of rendering to prisons is a good idea i won't object.

I saw as an open issue dedicated to basic contributors so I tried to do it. I think that it is better to have an overlayer than before, but I understand that it is not soo cool

> Technically there are a number of issues with this PR though:
>     * you don't remove the existing rendering of `amenity=prison` so it is rendered twice - that is probably not a good idea.

I will fix this, sorry I don't know this issue.

>     * you interpret `natural=scrub` - apparently to specifically handle the example case you demonstrate.  That is not a good idea.  We don't want to special case specific combinations of semantically unrelated tags.

Should it be generalized with all the natural elements?

> It would be good in general to test in an area of correct tagging based on current data.

I tried to find existing elements with this issue but I was not able to find them.

imagico:
> Should it be generalized with all the natural elements?

The rendering of amenity=prison should not depend on the presence of any other primary tags at all.

> I tried to find existing elements with this issue but I was not able to find them.

Huh?

https://www.openstreetmap.org/way/160091489
https://www.openstreetmap.org/way/149658767
https://www.openstreetmap.org/way/37392144
https://www.openstreetmap.org/way/34605696

style/landcover.mss
```diff
@@ -821,6 +813,19 @@
       line-offset: -1.0;
     }
   }
+  [natural = 'scrub'][zoom >= 8][way_pixels > 900],
+  [natural = 'scrub'][zoom >= 13],
```
jeisenbe, line 817: These lines should be removed

project.mml
```diff
@@ -531,10 +531,14 @@ Layer:
             way,
             landuse,
             military,
+            amenity,
+            "natural",
```
jeisenbe, line 535: This line is not needed

project.mml
```diff
@@ -531,10 +531,14 @@ Layer:
...
             way_area/NULLIF(POW(!scale_denominator!*0.001*0.28,2),0) AS way_pixels
           FROM planet_osm_polygon
           WHERE (landuse = 'military'
             OR military = 'danger_area')
+            OR ("natural" = 'scrub'
```
jeisenbe, line 540: This line is also not needed - you just want to render prisons as an overlay, not all natural=scrub.

pnorman:
Closing as abandoned.

PR #4176, opened 2020-07-05
Travis: Build with NodeJS 10 & 14

hiddewie:
At the moment there is only version 10 which has been released back in 2018. This project should also be able to build with a modern version of NodeJS.

- Build against NodeJS 14 as well as NodeJS 10

Older versions can be removed in the future.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,7 @@ dist: bionic
 sudo: false
 node_js:
   - "10"
+  - "14"
 services:
   - postgresql
 addons:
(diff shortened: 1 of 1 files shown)


pnorman:
We shouldn't need to build with multiple node versions, we have no nodejs code.

hiddewie:
OK, although the Carto library may or may not work with node 10 or 14. I changed the Travis configuration to only use version 14.

jeisenbe:
@pnorman - is this PR beneficial?

HolgerJeromin:
>is this PR beneficial? 

IMO yes. Old versions should be left behind. And it only affects CI build. As it build, there is no harm to drop 10 in travis CI.

PR #4177, opened 2020-07-05
fix rendering of highway=secondary_link at z12

crimsondusk:
- fixes the rendering of highway=secondary_link so that the casing becomes properly visible

https://www.openstreetmap.org/way/30071699

Before
z12-prior

After
z12-after


jeisenbe:
This PR fixed the width of the secondary_link fill at z12, which was previously too wide and thus covering the casing.

PR #4182, opened 2020-07-23
Switch to un-blessed (but still valid) osmdata

JesseWeinstein:
Due to a complicated and unfortunate mess, the blessed/marked good versions of the osmdata.openstreetmap.de output files have not been updated since Jan 9, 2020; and even though the original cause is (more or less) resolved, there have been enough changes in the past six months that the automatic script that blesses new versions still rejects them.

Happily, the un-blessed (but still valid, non-flooding-the-world) versions are still made available, just under slightly different URLs.

This is the minimal change needed to use those un-blessed files.

It is theoretically possible that joto, the maintainer of the osmdata box, may do a manual override (and I've requested that here: fossgis/osmdata#12 ) -- but as an alternative, and in the meantime, this enables the coastline data to be updated.

I previously proposed this by mistake in https://github.com/openstreetmap/chef/pull/326 ; TomH was kind enough to point me over here as the relevant place to ask the actual decision makers.

### DIFF
--- a/external-data.yml
+++ b/external-data.yml
@@ -9,7 +9,7 @@ sources:
     # The type of file this source is
     type: shp
     # Where to get it
-    url: https://osmdata.openstreetmap.de/download/simplified-water-polygons-split-3857.zip
+    url: https://osmdata.openstreetmap.de/new/simplified-water-polygons-split-3857.zip
     # The location within the archive
     file: simplified-water-polygons-split-3857/simplified_water_polygons.shp
     archive:
@@ -23,7 +23,7 @@ sources:
         - simplified-water-polygons-split-3857/simplified_water_polygons.shx
   water_polygons:
     type: shp
-    url: https://osmdata.openstreetmap.de/download/water-polygons-split-3857.zip
+    url: https://osmdata.openstreetmap.de/new/water-polygons-split-3857.zip
     file: water-polygons-split-3857/water_polygons.shp
     archive:
       format: zip


imagico:
@tomhughes did not point you here from what i can see.

Anyway - until someone demonstrates reliably that this is not the case we are going to stick with the well established knowledge that the global coastline assembly process needs to be subjected to a sanity check to avoid errors in the data to lead to large scale flaws in the map.  Many years of experience with that (see https://blog.jochentopf.com/2013-10-16-preventing-the-flooding-of-the-planet.html for example) have shown this to be the case.

If someone develops an alternative - either with a tiled coastline assembly and local error checks or just with a different error detection strategy and demonstrates this to work reliably we should look at that.  That is however not the case here, this change is suggesting to use data not subject to any error checks.

Suggestion to close this as not compatible with the goals of this style.

JesseWeinstein:
Tom pointed me here on IRC, after saying that he wasn't the decision maker, and that this group was. I would very much prefer to continue using the blessed files -- but they aren't being updated. 

I mainly created this to formally document this option -- I agree that it is non-optimal, but so is never again updating the coastlines.

tomhughes:
What I said on IRC was that (having considered the matter more after the initial response on the chef ticket) the real issue is that, regardless of my opinions, the chef recipe is just installing the data sources required by the style and it's therefore a decision for the style what data sources to use.

I none the less agree personally that this should be fixed upstream in some way, by resurrecting the coastline update process rather than by routing round it.

imagico:
All right, @tomhughes is right, this would be the primary place to change the coastline data source though OSMF operations would of course be able to override this in their deployment.

Regarding the issue - we are all aware that the coastline data on the source we use has not been updated for more than half a year (which is unfortunate but historically seen is not in any way unprecedented).  Hence as said we would look carefully at any viable alternatives (and independent of the specific issue with the stalled updates it would be great to have more than one maintained coastline data source for OSM in the same way we have several providers of OSM planet files and extracts).

What this issue suggests is as i have explained not a viable option for us so i am going to close it.  Any suggestion to adopt different sources of coastline data being maintained with a different process or with a different method of error detection and management would be looked at favorably under the criterion if it can be expected to reliably offer an error free and preferably  frequently updated assembly of the coastline data.

I would like to emphasize that this style is fairly neutral w.r.t. the Rio de la Plata matter and the positioning of the coastline by mappers.  This is relevant only in so far that a lack of consistency in coastline mapping could in principle lead us to no render the coastline at all.  However a lack of consistency in mapping in a single location does not raise to a level of affecting the global mapping quality in a statistically significant way.  Overall coastline mapping has a very good level of consistency and accuracy.

Thanks @JesseWeinstein for the PR - you are obviously welcome to contribute in this style otherwise on map design related subjects.

JesseWeinstein:
Osmdata has now made a manual override, so this should be resolved for now, happily. The underlying issue of people controversially marking things as bays remains, though. But now we can deal with it individually.

PR #4189, opened 2020-08-13
fix python script not running

Harmonious4:
The Python script that downloads the shapefiles needs to be run with python3

Please let me know if I did this correctly, this is my first pull request. I'm interested in contributing here, this line in docker-startup.sh was preventing the import from completing successfully.

### DIFF
--- a/scripts/docker-startup.sh
+++ b/scripts/docker-startup.sh
@@ -52,7 +52,7 @@ EOF
   $OSM2PGSQL_DATAFILE
 
   # Downloading needed shapefiles
-  scripts/get-external-data.py
+  python3 scripts/get-external-data.py
   ;;
 
 kosmtik)


pnorman:
The script starts with `#!/usr/bin/env python3` so it shouldn't need python3 before the command. What error were you getting?

Harmonious4:
`import_1   | /usr/bin/env: 'python3\r': No such file or directory`
`openstreetmap-carto_import_1 exited with code 127`
I noticed that in [this commit](https://github.com/gravitystorm/openstreetmap-carto/commit/67824003e02b27574de9b6f49313043248d5bcff) you replaced `python3 scripts/get-shapefiles.py -n` with `scripts/get-external-data.py` ... the old script get-shapefiles.py was run with python3 even though it also started with `#!/usr/bin/env python3`

pnorman:
It looks like it might be a newline issue

KevKo1990:
By applying the proposed change (https://github.com/gravitystorm/openstreetmap-carto/pull/4189/files/61e3d9083d36adf21decb7d78b505955fd2adbb3), the script run in my case successfully.

ZeLonewolf:
The instructions at https://github.com/gravitystorm/openstreetmap-carto/blob/master/DOCKER.md currently fails for me when running `docker-compose up import` on Windows 10.  This script change fixes the problem for me.

matkoniecz:
> The script should be runnable as an executable. If it is not, we need to fix that.

Is approach used in this PR to fixing the issue is wrong/problematic for some reason? If yes - why?

pnorman:
> Is approach used in this PR to fixing the issue is wrong/problematic for some reason? If yes - why?

Yes. The script is supposed to be an executable which can be run directly. 

Something is putting a \r in there on either git checkout or docker, where it is **not** present in the source.

ZeLonewolf:
I submitted a hopefully more acceptable solution to this in #4329

pnorman:
Fixed by #4330

PR #4132, opened 2020-04-29
Update natural earth links to use https

jeisenbe:
Changes proposed in this pull request:
- Update natural earth links to use https

The current links no longer work to download shapefiles correctly.

PR #4203, opened 2020-09-10
Update external-data.yml

arnaudblancher:
Correct external url
http://www.naturalearthdata.com/http//www.naturalearthdata.com/download/110m/cultural/ne_110m_admin_0_boundary_lines_land.zip
->  https://naciscdn.org/naturalearth/110m/cultural/ne_110m_admin_0_boundary_lines_land.zip

- 

Before

After

### DIFF
--- a/external-data.yml
+++ b/external-data.yml
@@ -62,7 +62,7 @@ sources:
 
   ne_110m_admin_0_boundary_lines_land:
     type: shp
-    url: http://www.naturalearthdata.com/http//www.naturalearthdata.com/download/110m/cultural/ne_110m_admin_0_boundary_lines_land.zip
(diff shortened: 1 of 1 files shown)


imagico (changes requested):
The canonical url seems to be

https://www.naturalearthdata.com/http//www.naturalearthdata.com/download/110m/cultural/ne_110m_admin_0_boundary_lines_land.zip

See https://www.naturalearthdata.com/downloads/110m-cultural-vectors/

And that seems to work.  Even if that is redirecting to naciscdn.org we should use what is advertised i think because the redirect can change any time.

If you want to re-purpose this PR to change the link to https (superseding #4132) that would be welcome.

pnorman:
Closing as abandoned.

PR #4218, opened 2020-10-11
Render mini-roudabout just like a turning circle

### LINKED ISSUE #4167: Road shield blocks highway=mini_roundabout
Example illustrating the problem:

https://www.openstreetmap.org/#map=19/49.15015/7.77923&layers=N

highway=mini_roundabout should have priority over everything else since it's a road feature, yet the road shield can block it.

mini_roundabout

hiddewie:
Fixes #4167

- Make `mini_roundabout` be rendered exactly like a turning circle
- Includes `secondary` and `primary` roads in the SQL query for turning circles.

The icon is still rendered. It can be removed, or changed to a single dot in the *casing* color of the road which contains the roundabout.

Test rendering with links to the example places: https://www.openstreetmap.org/#map=19/49.15015/7.77923&layers=N

Before/after renderings (zoom is shown in image):

image

image

image

image

image

image


imagico:
In principle a good idea.  But you are not actually fixing #4167 this way because you keep rendering the mini roundabout symbol as a blocking symbol:

https://github.com/gravitystorm/openstreetmap-carto/blob/b10aef3866bacf387581b8fea4eec265010b0d14/style/amenity-points.mss#L1486

I would suggest to follow the suggestion by @jeisenbe - continue rendering the center symbol, possibly in a less heavy variant (and possibly in casing color of the road) but without it blocking other symbols or labels.

hiddewie:
@imagico  Done. It is now a mini-dot in the middle of the roundabout. Not a blocking icon anymore.

That styling only applies to the `mini_roundabout`, not the turning-circles.

The color is matching the casing of the road that contains the roundabout.

polarbearing:
Looks nice in the examples above, thanks. 
Do we have any test rendering for lower roads, such as tertiary and residential, where mini roundabouts are more frequent?

PR #4250, opened 2020-11-14
Use naciscdn URL for NE data

### LINKED ISSUE #4249: scripts/get-external-data.py fails because ne_110m_admin_0_boundary_lines_land.zip has moved
(this is on Debian 11 testing)

```
scripts/get-external-data.pyINFO:root:Checking table simplified_water_polygons
INFO:root:Checking table water_polygons
INFO:root:Checking table icesheet_polygons
INFO:root:Checking table icesheet_outlines
INFO:root:Checking table ne_110m_admin_0_boundary_lines_land
  File "scripts/get-external-data.py", line 258, in <module>
    main()
  File "scripts/get-external-data.py", line 206, in main
    download.raise_for_status()
  File "/usr/lib/python3/dist-packages/requests/models.py", line 941, in raise_for_status
    raise HTTPError(http_error_msg, response=self)
requests.exceptions.HTTPError: 500 Server Error: Internal Server Error for url: http://www.naturalearthdata.com/http//www.naturalearthdata.com/download/110m/cultural/ne_110m_admin_0_boundary_lines_land.zip
```

It's odd that that URL returns 500 there.  A "wget" on the same URL gets a 500 too as opposed to a 302; but a browser download follows a redirect to https://naciscdn.org/naturalearth/110m/cultural/ne_110m_admin_0_boundary_lines_land.zip .

Obviously one "fix" would be to change the URL in "get_external_data.yml", but is it actually possible to get "get_external_data.py" to follow the redirect in the same way that a browser does?

pnorman:
The naturalearthdata.com links are broken, so we can use the naciscdn ones.

Fixes #4249

### DIFF
--- a/external-data.yml
+++ b/external-data.yml
@@ -62,7 +62,7 @@ sources:
 
   ne_110m_admin_0_boundary_lines_land:
     type: shp
-    url: http://www.naturalearthdata.com/http//www.naturalearthdata.com/download/110m/cultural/ne_110m_admin_0_boundary_lines_land.zip
+    url: http://naciscdn.org/naturalearth/110m/cultural/ne_110m_admin_0_boundary_lines_land.zip
     file: ne_110m_admin_0_boundary_lines_land.shp
     ogropts: &ne_opts
       - "--config"


pnorman:
There's some odd redirects going on still causing problems for browser users, but they don't impact us so the scripts work again.

ref https://github.com/nvkelso/natural-earth-vector/issues/434#issuecomment-731490754

PR #4251, opened 2020-11-20
Render landuse=flowerbed

liotier:
landuse=flowerbed is a gap in the rendering of urban vegetation. 13k occurrences according to Taginfo. I propose using the same rendering as landuse=orchard, as orchard and flowerbed are unlikely to be adjacent. I propose starting at zoom 11 because leisure=garden starts at zoom 10 and flowerbeds are smaller detail.

### DIFF
--- a/style/landcover.mss
+++ b/style/landcover.mss
@@ -167,6 +167,20 @@
     }
   }
 
+  [feature = 'landuse_flowerbed'] {
+    [zoom >= 11] {
+      polygon-fill: @orchard;
+      [way_pixels >= 4]  { polygon-gamma: 0.75; }
+      [way_pixels >= 64] { polygon-gamma: 0.3;  }
+    }
+    [zoom >= 13] {
+      polygon-pattern-file: url('symbols/orchard.png');
+      polygon-pattern-alignment: global;
+      [way_pixels >= 4]  { polygon-pattern-gamma: 0.75; }
+      [way_pixels >= 64] { polygon-pattern-gamma: 0.3;  }
+    }
+  }
+
   [feature = 'landuse_plant_nursery'] {
     [zoom >= 10] {
       polygon-fill: @orchard;


jeisenbe:
This tag was documented with a [draft proposal in 2014](https://wiki.openstreetmap.org/w/index.php?title=Proposed_features/flowerbed&oldid=1071989) and with a wiki page in [2015](https://wiki.openstreetmap.org/w/index.php?title=Tag:landuse%3Dflowerbed&oldid=1160264), with more details added in 2019. Usage increase substantially in the last year, perhaps because it is now supported by presets in JOSM and iD: https://taginfo.openstreetmap.org/tags/landuse=flowerbed#projects

https://taginfo.openstreetmap.org/tags/landuse=flowerbed#chronology - now used >10,000 times. 
Screen Shot 2020-11-19 at 20 39 05

The only similar tags, man_made=flowerbed, man_made=flower_bed, and landcover=flowerbed, are undocumented and only used a couple hundred times: https://taginfo.openstreetmap.org/search?q=flowerbed#values https://taginfo.openstreetmap.org/search?q=flower_bed#values

Perhaps this is now common enough to be supported with a similar rendering to leisure=garden?

literannn:
I'd prefer rendering like leisure=garden.

liotier:
> I'd prefer rendering like leisure=garden.

That was my first idea. But then flowerbeds in gardens would be invisible - hence why I suggested another way.

literannn:
maybe better to unify rendering of gardens and leisure=park, and then pattern of garden use for flowerbeds

liotier:
Considering the semantic overlap between leisure=park and leisure=garden - and the constant confusion between the two because the distinction is too subtle, unifying their rendering feels logical to me and I'm comfortable with the idea of using the garden pattern for flowerbeds.

imagico:
Thanks for the PR.  Use of the tag still seems to be somewhat patchy but it also looks like being highly consistent compared to broader tags like `landuse=garden` so in principle something i would support.

But not in a styling like `landuse=orchard`.  Those are both physically and purpose-wise very different.  A good design probably requires looking at the urban green rendering in more depth - which is unfortunately quite a mess right now.  If you are thinking of using the plant_nursery/garden pattern - keep in mind that this is broken at the moment - see #2771/#2773.  An alternative design can be found in the ac-style: third one in

http://blog.imagico.de/wp-content/uploads/2018/11/ac_patterns_vis.png

jeisenbe (changes requested):
This is now marked as “de facto” on the wiki and that appears to be correct since there are no competing tags for this kind of feature and it is fairly well defined. 

I would be happy to approve this PR if there are significant changes to the current rendering suggestion. Like @imagico I also think that rendering with the orchard/vineyard color and pattern does not make sense: orchards and vineyards have regular plantations of perineal woody plants which are the height of shrubs or trees, used to produce food, while flowerbeds usually contain flowering herbs, intendent for decoration or aesthetic value.

jeisenbe:
Re: other rendering options.

I also agree that rendering leisure=garden like leisure=park and landuse=recreation_area might work, perhaps while keeping the pattern as a distinction? 

@liotier are you able to take the time to look at all the kinds of urban / recreational vegetation currently rendered and make a rendering which will make sense? You might have to change the rendering of several features, which can take some time to get right.

liotier:
My understanding of the requirement for a landuse=flowerbed style:
- It is about vegetation, so green must be the dominant color
- It must express density higher than grass but lower than wood or scrub. A color intensity somewhere between leisure=garden and natural=wood might fit - but maybe not pastel like natural=scrub. That color intensity should be understood as the total of background and pattern
- Flowerbeds are not entirely flat, especially compared to grass, so some texture would be fitting - expressed through some pattern. The wood and scrub patterns are out of question, hence why out of the existing patterns I thought that some dotted pattern might do the job

jeisenbe:
> It is about vegetation, so green must be the dominant color

Agreed.

> It must express density higher than grass...

Not necessarily. Recall that landuse=meadow, natural=grassland and landuse=grass use the same color of green. The meadows and natural grasslands often includes wildflowers and non-grass herbs, and in areas with sufficient rainful can easily grow to 1 meter height or more. 

> Flowerbeds are not entirely flat, especially compared to grass, so some texture would be fitting

The patterns are mainly used to distinguish 2 different features which use the same base color. The use of a regularly-spaced pattern can represent human designed features (which are often planted in repeating patterns) - see landuse=orchard, landuse=vineyard, the current leisure=garden pattern and landuse=plant_nursery for examples. 

In contrast the semi-natural area patterns are somewhat random looking, e.g. natural=scrub, natural=wetland etc.

In this case since we already have a number of different green colors, it would be reasonable to use the same color as grass / meadow / grassland, which also represent herbaceous vegetation, and use a pattern to distinguish this from areas of grass. 

However, you can consider other options, as long as they fit in with the existing colors.

jeisenbe:
@liotier this is now marked as a draft, meaning it is not ready to merge yet, but if you have time to update it, we will be happy to review it again.

matkoniecz:
rendering it like leisure=garden / leisure=park / landuse=recreation_area seems a good idea. Main negative is that flowerbed in garden/park would not be visible

pnorman:
> @liotier this is now marked as a draft, meaning it is not ready to merge yet, but if you have time to update it, we will be happy to review it again.

Closing as abandoned. Please feel free to make the adjustments requested and open a new PR, or ask for this to be re-opened.

Issue #4279, opened 2021-01-09
Optimize queries using info from OSMF servers


pnorman:
In https://github.com/openstreetmap/operations/issues/495 I got a sysadmin to sample the queries, allowing for an analysis of what's slow.

Grouped by the first lines of the query, this identifies the slow queries

sum        | prefix                                                                                                      
--|--
 00:03:16.497222  | ``SELECT ST_AsBinary("way") AS geom,"feature","is_building","name","religion","way_pixels" FROM (SELECT\n    way, name, religion, way_pixels, is_building,\n    COALESCE(aeroway, amenity, wetland, power, landuse, leisure, man_made, "natural", shop, tourism, highway, railway) AS feature``
 00:02:39.464784  | ``SELECT ST_AsBinary("way") AS geom,"bridge","int_intermittent","int_tunnel","waterway" FROM (SELECT\n    way,\n    waterway,``
 00:01:01.774667  | ``SELECT ST_AsBinary("way") AS geom,"feature","link" FROM (SELECT\n    way,\n    (CASE WHEN feature IN ('highway_motorway_link', 'highway_trunk_link', 'highway_primary_link', 'highway_secondary_link', 'highway_tertiary_link') THEN substr(feature, 0, length(feature)-4) ELSE feature END) AS feature,``
 00:00:44.244365  | ``SELECT ST_AsBinary("way") AS geom,"access","bicycle","construction","feature","horse","int_surface","link","tracktype" FROM (SELECT\n    way,\n    (CASE WHEN feature IN ('highway_motorway_link', 'highway_trunk_link', 'highway_primary_link', 'highway_secondary_link', 'highway_tertiary_link') THEN substr(feature, 0, length(feature)-4) ELSE feature END) AS feature,``
 00:00:43.933502  | ``SELECT ST_AsBinary("way") AS geom FROM (SELECT\n    way\n  FROM planet_osm_line``
 00:00:36.795382  | ``SELECT ST_AsBinary("way") AS geom,"access","feature","generator:source","height","iata","icao","is_building","location","name","shop","sport","way_pixels" FROM (SELECT\n    *\n  FROM``
 00:00:30.663109  | ``SELECT ST_AsBinary("way") AS geom,"access","feature","height","iata","icao","name","way_pixels" FROM (SELECT\n    *\n  FROM``

The math to prove it is complicated, but the higher the sum of the durations is, the more time the render server spent on those queries.

[...]

PR #4280, opened 2021-01-09
Create a partial index for waterways

pnorman:
Part of #4279

I was seeing a 50x speedup on these queries on my machine, and they have been identified as some of the slower queries.


imagico (approved):
Works fine and makes sense from my perspective, did not check if there are any substantial performance improvements.

Issue #3613, opened 2018-12-31
Current healthcare icon color too strong


matthijsmelissen:
I think the current healthcare icon is too strong compared to other colors on the map. Would it be possible to turn down the saturation a bit?

<img width="425" alt="screen shot 2018-12-31 at 10 52 19" src="https://user-images.githubusercontent.com/5251909/50557946-2ee2b200-0cea-11e9-928d-2ebed21ba5c6.png">

CC @jragusa

jeisenbe:
> turn down the saturation a bit

I agree, after comparing the new health color to  the other colors used for icons. It's the color with the highest chroma (saturation, "c") currently

_Ton line Left to right: gastronomy, **health**, shop, air-transport_:
![almost-all-icon-colors-current](https://user-images.githubusercontent.com/42757252/50966507-794d0580-1518-11e9-98ae-8f7473d21b02.png)
- Second row left to right: office, transportation, marina-text (fountains), water-text (waterfalls)
- Third row: amenity-brown, landform-color, man-made-icon, leisure (park darkened 60%)
- These are almost all of the current icon colors, except for black, used for religious features, and the blue used for springs, which is similar to water-color

I'd suggest reducing the chroma slightly to 72, which would be the same chroma as `@shop` - this would be `#b52613` or something similar.

It also looks like we should reduce the number of shades of blue.

seav:
Aside from the color, I find it weird that the hospital and clinic icon starts rendering at zoom level 15, earlier than any other POI icon.

kocio-pl:
Only hospital is drawn from z15 and it does not surprise me, since it is usually a big entity.

https://github.com/gravitystorm/openstreetmap-carto/blob/762b292b7ab9d353ed1fd99adabe6c44070977d4/amenity-points.mss#L654-L655

seav:
![openstreetmap_-_2019-01-12_02 04 44](https://user-images.githubusercontent.com/873653/51051411-a8de3980-160e-11e9-8d8d-e229174a61d2.png)

Yes, they are usually big but not always. (In my hometown, most hospitals are actually small office-like buildings a few storeys high.) As such, they look really out of place on the map because they are the only large POIs that have icons at z15 and they stand out even more because of the saturated color. Note that other large POIs like colleges and universities don't even have icons.

kocio-pl:
I would look closer on tagging, because from my experience many hospitals are de facto clinics, doctor's offices, hospital departments etc.

[...]

PR #4300, opened 2021-01-27
Render vaccination_centres.

### LINKED ISSUE #4295: Render Vaccination centers/centres.
This tag has been approved recently: https://wiki.openstreetmap.org/wiki/Proposed_features/Tag:healthcare%3Dvaccination_centre

Given it usefulness in this time and age, I propose we render it. I will try to some up with the code myself, otherwise I'll let you know that I gave up (unluckily I have not much free time in my hands).

StyXman:
Fixes #4295.

- add `healtcare=vaccination_centre`

https://www.openstreetmap.org/#map=15/51.5566/-0.2794

Before
Screenshot_20210127_153709

After:
Screenshot_20210127_153623

The `vaccination_centre` is just north of the Wembley Stadium.

### DIFF
--- a/style/amenity-points.mss
+++ b/style/amenity-points.mss
@@ -593,6 +593,13 @@
     marker-clip: false;
   }
 
+  [feature = 'healthcare_vaccination_centre'][zoom >= 15] {
+    marker-file: url('symbols/amenity/hospital.svg');
+    marker-width: 10;
+    marker-fill: @health-color;
+    marker-clip: false;
+  }
+
   [feature = 'amenity_pharmacy'][zoom >= 17] {
     marker-file: url('symbols/amenity/pharmacy.svg');
     marker-fill: @health-color;
--- a/project.mml
+++ b/project.mml
@@ -1535,6 +1535,7 @@ Layer:
                                         AND way_area IS NULL
                                    THEN amenity END,
                 'tourism_' || CASE WHEN tourism IN ('viewpoint', 'attraction') THEN tourism END,
+                'healthcare_' || CASE WHEN tags->'healthcare' in ('vaccination_centre') then tags->'healthcare' END,
                 'place_' || CASE WHEN place IN ('locality') AND way_area IS NULL THEN place END
               ) AS feature,
               access,


StyXman:
The icon is a 10px hospital one, as I couldn't find any that looked like a syringe :)

polarbearing:
We cannot re-use the hospital icon while telling mappers not to tag the vacc-centres as hospital. 
Works neither cartographically nor as mapper feedback. 
Icons usually have to be created, not found.

imagico:
Thanks for the submission - code looks fine, although i have not tested it.

Right now i see the following problems with this:

* Use of the tag is not very widespread so far - about a hundred compared to >500 for healthcare:speciality=vaccination.  We should not push for a newly created tag and instead wait until it is clear if said tag finds broad support from the community (which is likely here but not guaranteed).
* Many of the places tagged with this seem not open to the general public but only accessible with an appointment you get subject to varying conditions.  So far however taggings indicating such things are not widely used.  Indiscriminately showing  vaccination centers without indicating if these are open to the general public would limit usefulness of rendering to the map user.
* A distinct and intuitively readable symbol is an absolute necessity for adding rendering.  If a smaller version of an existing symbol is used (and i don't think that is a good idea here) it would need to be adjusted to good quality rendering at the target size.
* As with other POI rendering suggestions an overall concept of symbology in the domain in question (here: medical/healthcare facilities) that the symbol fits in here should be a prerequisite.  At hoc, compartmentalized additions of individual symbols are not a good idea - see #3635.  In this case specifically the question that would need to be considered is should we decide to render vaccination centers if we should also render vaccination opportunities at places other than dedicated vaccination centers (i.e. healthcare:speciality=vaccination or vaccination=* as addon tags for other medical facilities).  Not doing so could very well create counterproductive mapping incentives (i.e. people adding healtcare=vaccination_centre to indicate a vaccination opportunity that is not a dedicated vaccination center).

StyXman:
This tag has been approved recently: https://wiki.openstreetmap.org/wiki/Proposed_features/Tag:healthcare%3Dvaccination_centre

> Use of the tag is not very widespread so far - about a hundred compared to >500 for healthcare:speciality=vaccination

Let's use that one too, then.

> Many of the places tagged with this seem not open to the general public but only accessible with an appointment you get subject to varying conditions. So far however taggings indicating such things are not widely used. Indiscriminately showing vaccination centers without indicating if these are open to the general public would limit usefulness of rendering to the map user.

Let's narrow down the filters then. Any suggestions?

> A distinct and intuitively readable symbol is an absolute necessity

Yeah, I put the hospital icon just to prove the SQL query works. Does anyone know a syringe icon somewhere? I _will_ try to find one myself, but my time is quite limited.

polarbearing:
It's unlikely to _find_ an icon matching the requirements here, which is a 14px-optimised monochrome svg.
I'd imagine a syrringe can be drawn as a T for the handle, a rectangle and a thin line, diagonally. Will see if I find time to try.

jeisenbe:
While I was personally involved in creating this tag, I don’t think it is well established enough to render right now, considering that we don’t render more common tags like healthcare=laboratory. 

However, it would be worth considering adding rendering along with a few of the most common healthcare=* tags once it is more common - see https://wiki.openstreetmap.org/wiki/Template:Map_Features:healthcare

Perhaps we coudl use a generic, abstract icon such as a filled circle, rather than creating new icons for each feature? 

See also #3635 for discussion about the abundance of icons in this style.

style/amenity-points.mss
```diff
@@ -593,6 +593,13 @@
     marker-clip: false;
   }
 
+  [feature = 'healthcare_vaccination_centre'][zoom >= 15] {
+    marker-file: url('symbols/amenity/hospital.svg');
```
jeisenbe, line 597: The symbol needs to be changed, as discussed, either to a generic icon which can be used for a number of healthcare features which are currently not yet rendered (e.g. laboratories, test centres, birth centres, midwives, nurses, chiropractors, acupuncture, etc) or a unique symbol needs to be created which is recognizable, e.g. a syringe.

jeisenbe (changes requested):
This PR currently would not render the names of healthcare=vaccination_centre features. That would require adding this to line 2552: https://github.com/gravitystorm/openstreetmap-carto/blob/d13ed55f5ea4a1ae3e263bce0622baad1a80b5cf/style/amenity-points.mss#L2552 - and as mentioned the icon needs to be changed

enzet:
Here are some sketches of 14 × 14 px icon:

syringe_1

The same × 4:

syringe_4

Any thoughts? What is better: more detailed, more pixel-perfect?

imagico:
Those look very good.  The other issues mentioned in https://github.com/gravitystorm/openstreetmap-carto/pull/4300#issuecomment-768402991 still apply of course.  Considering as mentioned vaccination opportunities are often tagged with addon tags (healthcare:speciality=vaccination, vaccination=*) to various medical facilities it could be worth trying to indicate this with symbol variations (i.e. adding a syringe symbol to existing symbols for doctors/clinics etc.)  I am not saying that this would necessarily work well in this style but it could be worth trying out (also potentially for other medical specialities of course).

enzet:
> it could be worth trying to indicate this with symbol variations (i.e. adding a syringe symbol to existing symbols for doctors/clinics etc.)

Sorry, not sure I got you right. Do you mean something like that?

syringe_cross

pnorman:
Scaled to 8x
image
Numbering from the top-left and going across, the first three should all have a one pixel gap like the ones below. 
Four and eight are quite blurry because they're angled.

enzet:
> the first three should all have a one pixel gap like the ones below

I've intentionally added half-pixel gaps to some icons and one-pixel gaps to others to compare and select the best elements.

imagico:
> Sorry, not sure I got you right. Do you mean something like that?

In principle yes.  But that is delicate work to design with non-misleading semantics.  It would need to be clear that the added syringe indicates a subtype of the symbol without it.  We have something like that for different kinds of parking ('P' symbol plus additional symbol) but this would be more difficult for healthcare symbols, in particular because of #3613 and because current healthcare symbols here use a Swiss coat of arms/negative space symbol rather than an actual red cross.

enzet:
Well, I realized that the syringe shape wasn't right. I also changed the needle shape. There are updated and rearranged sketches:

syringe_2

× 4:

syringe_2_x4

For references:
  * without gaps: 2.1 — 2.4,
  * with half-pixel gaps: 2.5 — 2.8,
  * with one-pixel gaps: 2.9 — 2.12.

So far I prefer 2.2 cause it's the most recognizable one.

jeisenbe:
@StyXman could you update this PR with the new symbol and address the previous comment about line 2552?

Now used 700 times, probably not going away anytime soon.

pnorman:
Closing as stale. If you're still interested in this, feel free to open a new PR addressing the above comments.

Issue #4197, opened 2020-08-27
Don’t render amenity=parking access=private with low capacity


pwithnall:
### Expected behavior

I think small, private car parks should not be rendered at all in osm-carto. Typically, this kind of car park is a couple of parking spaces assigned to specific houses, but grouped together rather than being on the private curtilage of any specific house. They end up being mapped, as they occupy some amount of space. But the information that they are there is not useful to anyone except the owners of those houses and people visiting them. It does not need to be on a public map, clutters the map, and potentially confuses others into parking where they’re not supposed to.

So I think anything with amenity=parking access=private capacity<10 should be rendered with the grey parking background, but not have a ‘P’ rendered.

### Actual behavior

Lots of small parking areas are rendered, and the fact they’re lighter/transparent is not obvious because there’s no public car park to contrast them with. People get confused and park where they’re not supposed to. The map is cluttered.

### Links and screenshots illustrating the problem

https://www.openstreetmap.org/#map=19/54.32177/-2.75600

![image](https://user-images.githubusercontent.com/53172/91452019-0243bc80-e876-11ea-8ed9-a7e27aafaced.png)

jeisenbe:
If the issue is cluttering of the map, it's possible that this could be resolved by filtering out parking areas with small areas. However, since we would still want to render nodes tagged with `amenity=parking`, this would lead to the odd result that mapping the parking as an area might cause the symbol to disappear. 

Using the `capacity=` tag is unlikely to work in most cases, because only 6.6% of `amenity=parking` features have a `capacity=` tag: 
https://taginfo.openstreetmap.org/tags/amenity=parking#combinations

cquest:
On the french rendering, I'm using more and more the pixel size of objects to prioritize rendering or decide not to render and not only by zoom level.
It is computed by PG:  way_area / (!pixel_width! * !pixel_height!) as pixels

For parkings, I prioritize multistorey ones.

capacity=* could be used as it is more often set on large ones (and as usual, rendering them using this tag will be an incentive to add it).

matkoniecz:
> The map is cluttered.

Given example seems to not be couriered to me.

And if that works be considered as clutter that should be fixed then we have many, many, many things to remove. I would be against that.

> potentially confuses others into parking where they’re not supposed to.

That seems ridiculously unlikely. And if someone believes his/her misinterpretation of an online map over signage - then I hope that they will lose traffic license before killing someone in traffic accident caused by their stupidity.

[...]

PR #4301, opened 2021-01-28
De-emphasize parking=street_side

### LINKED ISSUE #4262: Streetside parking should be rendered differently than normal parkings
### Expected behavior
Streetside parking is only rendered at zoom 19 (or maybe 18 too) in a less intrusive way than a blue P

### Actual behavior
Each parking tagged with parking=street_side shows a blue P

[A new way](https://wiki.openstreetmap.org/wiki/Tag:parking%3Dstreet_side) of tagging streetside parking was approved a few days ago, with the goal to clarify how to map this type of feature.
Given that this method only adds an additional tag to the already in use `amenity=parking` tag, the parking spaces are represented each with a blue P, the same as actual parkings.
As a representation, this is probably too heavy, and something similar to [parking spaces](https://wiki.openstreetmap.org/wiki/Tag:amenity%3Dparking_space) should be used
#### Zoom 18
C1
#### Zoom 19
C2

[Link](https://www.openstreetmap.org/#map=19/45.47721/9.16270) to the area of the screenshots

jdhoek:
For `amenity=parking` of the type `parking=street_side` (street-side parking in bays and other demarcated parking areas) and `parking=lane` (parking on a lane alongside a street), the amenity icon is replaced with a smaller one that is shown from `z18` onwards.

With the introduction and formal acceptance of `parking=street_side`, mappers now have a way of indicating that a mapped parking amenity is a parking bay/area on the side of a street. This tag gives renderers a chance to solve the problem of these parts of the street layout from cluttering the map with large blue **P**'s that are better suited for larger parking lots.

This commit introduces a smaller **P** icon that is pixel aligned to match the crispness of the existing generic parking **P**.

Fixes #4262.

-----

### Current rendering at `z17`

orig-z17

### Proposed at `z17`

No **P** for `parking=street_side` yet; the other `amenity=parking` are unchanged.

p-z17

### Current rendering at `z18`

orig-p18

### Proposed at `z18`

Subtler small **P**.

p-z18

### Proposed at `z19`

p-z19

### Proposed in comparison with `parking=surface`

comparison-with-generic-p

-----

This PR is limited to the introduction of the smaller **P** for `parking=street_side` and `parking=lane`. During testing of this feature I've noticed that the grey background of `amenity=parking` makes larger parking lots and areas lose their association with the icon and the meaning of the area.

I have found that a slight blue tint pulls the icon and the background together, creating a more coherent rendering:

blue-2

blue-1

Here I've used `@parking: #ededef;`.

This is not for this PR, but worth considering as a follow up development, and it may solve what @jeisenbe mentioned in #4262:

> Another issue is that the current parking background color is not very distinctive, so if we don’t show the icon it might not be clear what we are looking at. 

Here the icon remains at `z >= 18` (but de-emphasized), but the same concern is present here too.

### DIFF
--- a/style/amenity-points.mss
+++ b/style/amenity-points.mss
@@ -1446,10 +1446,14 @@
   [feature = 'amenity_bicycle_parking'],
   [feature = 'amenity_motorcycle_parking'],
   [feature = 'amenity_parking_entrance'] {
-    [zoom >= 14][way_pixels > 750],
-    [zoom >= 17][feature = 'amenity_parking'],
+    [zoom >= 14][way_pixels > 750]["parking" != 'street_side']["parking" != 'lane'],
+    [zoom >= 17][feature = 'amenity_parking']["parking" != 'street_side']["parking" != 'lane'],
     [zoom >= 18] {
       [feature = 'amenity_parking'] { marker-file: url('symbols/amenity/parking.svg'); }
+      [feature = 'amenity_parking']["parking" = 'street_side'],
+      [feature = 'amenity_parking']["parking" = 'lane'] { 
+        marker-file: url('symbols/amenity/parking_subtle.svg'); 
+      }
       [feature = 'amenity_bicycle_parking'] { marker-file: url('symbols/amenity/bicycle_parking.svg'); }
       [feature = 'amenity_motorcycle_parking'] { marker-file: url('symbols/amenity/motorcycle_parking.svg'); }
       [feature = 'amenity_parking_entrance']["parking"='underground'] { marker-file: url('symbols/amenity/parking_entrance_underground.svg'); }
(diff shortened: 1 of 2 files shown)


jdhoek:
A comment on the OSM wiki reminded me that `parking=lane` has essentially the same rendering requirements as `parking=street_side`. I have added that `parking` value to this PR for good measure.

jeisenbe:
@jdhoek - this looks pretty good. I was hoping to address some of the other issues with the parking rendering as mentioned in the comments, but it is reasonable to start with just this change. 

I am a little concerned about supporting such a new tag, but the usage trends are strong:

Screen Shot 2021-02-24 at 22 16 15

Screen Shot 2021-02-24 at 22 15 37

jdhoek:
It's a Catch-22. I think `parking=street_side` solves a lot of the issues ([like this one](https://github.com/gravitystorm/openstreetmap-carto/issues/4197#issuecomment-694821613); adding the `parking=street_side` tag is required of course) with rendering parking amenities (many stem from the use of `parking=surface` for street-side parking), but for broader adoption rendering support is needed at some point. The usage of the tag seems to grow organically and at a steady pace.

kocio-pl:
This helps un-clutter cities and is clearly supported by community, so thanks for your submission and sorry for waiting so long. I'm especially happy with solving one, easy to understand problem at the time,

jdhoek:
Thanks! Looking forward to seeing this on the live map.

PR #4307, opened 2021-01-30
Add an option to grant permissions to a user for rendering from the shapefile tables

### LINKED ISSUE #4305: Need some way to get a list of external data tables
In order that the production deployment can set permissions correctly it needs to be able to get a list of the external data tables so that it can grant select permission on them.

pnorman:
This is necessary on production systems where a different user is used for rendering.

Tested locally and select was granted on the new table.

cc @tomhughes

Fixes #4305

### DIFF
--- a/scripts/get-external-data.py
+++ b/scripts/get-external-data.py
@@ -65,6 +65,12 @@ class Table:
             if results is not None:
                 return results[0]
 
+    def grant_access(self, user):
+        with self._conn.cursor() as cur:
+            cur.execute('''GRANT SELECT ON "{temp_schema}"."{name}" TO "{user}";'''
+                        .format(name=self._name, temp_schema=self._temp_schema, user=user))
+            self._conn.commit()
+
     def index(self):
         with self._conn.cursor() as cur:
             # Disable autovacuum while manipulating the table, since it'll get clustered towards the end.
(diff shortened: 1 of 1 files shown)


scripts/get-external-data.py
```diff
@@ -270,6 +282,8 @@ def main():
                             "ogr2ogr error when loading table {}".format(name))
 
                     this_table.index()
+                    if renderuser is not None:
+                        this_table.grant_access(renderuser)
                     this_table.replace(new_last_modified)
```
mmd-osm, line 287: As discussed earlier on, the idea of this table switch operation is to ensure an atomic table switch once a table has been successfully loaded.

My concern was that the table switch should only be done after the loading process for *all* tables has finished. This way, none of the production tables would be replaced by their temporary table counterpart in case of some failure.

@pnorman remarked, that all tables are being treated independently by design and in practice they're either updated separately or used on different zoom levels. Hence, there's no need to have the table switch for all tables at the very end the process.

mmd-osm:
One thing I've noticed during testing: if you run the script once without specifying the render user, and then later run the script again, it might happen that the data hasn't changed in the meantime. In that case the script would print a message `"INFO:root:Table simplified_water_polygons did not require updating"` without updating the render user.

PR #4311, opened 2021-01-31
Fixed #4270

### LINKED ISSUE #4270: Public transport shelter should not be labeled with elevation
### Expected behavior

A bus shelter tagged `amenity=shelter` `building=yes` `shelter_type=public_transport` should be rendered as an ordinary building with or without an `ele` tag.

### Actual behavior

The `ele` tag causes the bus shelter to be rendered like an alpine hut, labeled with a 🌧️ and an elevation. For example, [this way](https://www.openstreetmap.org/way/851972354) was added as part of a building import that included elevations and heights for every building. The bus shelter is useful for avoiding the rain, but the elevation is not particularly interesting for a 2D renderer. (It might be useful for a 3D renderer.)

[vta](https://www.openstreetmap.org/way/851972354)

/ref #3382

GoutamVerma:
Fixes #4270

Changes proposed in this pull request :  **Public transport shelter should not be labeled with elevation
Test rendering with links to the example places:**

Add  :` AND tags->shelter_type NOT IN (public_transport, picnic_shelter) 
`

### DIFF
--- a/project.mml
+++ b/project.mml
@@ -1466,7 +1466,7 @@ Layer:
                     WHEN (tags ? 'ele') AND tags->'ele' ~ '^-?\d{1,4}(\.\d+)?$'
                       AND ("natural" IN ('peak', 'volcano', 'saddle')
                         OR tourism = 'alpine_hut' OR (tourism = 'information' AND tags->'information' = 'guidepost')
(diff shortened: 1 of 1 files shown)


imagico:
Why have you closed this?  You need to change the expression as explained in https://github.com/gravitystorm/openstreetmap-carto/issues/4270#issuecomment-770379869 but otherwise this looks fine.

PR #4312, opened 2021-01-31
Fixed #4207

### LINKED ISSUE #4207: Is there a way to update for a specific layer without having to reload the map tiles each time?
Hi
I would like to display the location of specific agents in real time (or pseudo-real time as it relates to a simulation) on a map. I am happy to begin with to explore movement of the agents within a specific set of bounds so the zoom and area of the map can stay constant. Is there a way of being able to update the page with the new locations of the agents without having to render the tiles for the whole map each time. I have been able to add a layer relating to the agents which appends to the project.mml file (although there may be a more appropriate way of doing this - I am pretty new to openstreetmap-carto). This allows me to plot the agents on the map. I have a second container which calculates the new location for each agent which updates the project.mml file and this is eventually updated on the rendered map. However the rendering is not fast enough to show the agents moving across the map. I think if I could update the map with just the layer relating to the agents location it should speed things up significantly.

GoutamVerma:


imagico:
Make sure you understand the condition you are modifying here - `(tags ? 'ele') AND tags->'ele' ~ '^-?\d{1,4}(\.\d+)?$'` is `true` if an ele tag is present and contains a valid number.  That is followed by `AND (/conditions in tags under which we want to show the elevation/)`.  And those conditions so far contain an `amenity = 'shelter'` - which you need to modify into the mentioned compound condition based on amenity=shelter and `tags->shelter_type`.  And never chain `AND` and `OR` without braces for explicit operator precedence.

Also make sure to test the change if it actually has the desired effect.

PR #4317, opened 2021-02-02
Change icon for shelter_type=public_transport

jdhoek:
The default shelter icon looks a bit strange for shelters tagged with `shelter_type=public_transport`; the minimal shelter type you would find at a bus or tram stop.

This commit introduces a distinct icon for this shelter type, and changes the colour of it to `@transportation-icon` to match the public transport stop itself.

This type of icon was suggested by @polarbearing in #4270.

## Rendering

Screenshot from 2021-02-02 19-23-08

Seems to work okay with shelters mapped as buildings too:

Screenshot from 2021-02-02 19-29-16

At the very least no worse than the generic shelter icon.

## Current icon and colour

Screenshot from 2021-02-02 19-30-14

Screenshot from 2021-02-02 19-31-19

### DIFF
--- a/project.mml
+++ b/project.mml
@@ -1592,6 +1592,7 @@ Layer:
                                 ELSE 'other' END AS shop,
               CASE WHEN building = 'no' OR building IS NULL THEN 'no' ELSE 'yes' END AS is_building,
               tags->'operator' AS operator,
+              tags->'shelter_type' AS shelter_type,
               ref,
               way_area,
               COALESCE(way_area/NULLIF(POW(!scale_denominator!*0.001*0.28,2),0), 0) AS way_pixels
--- a/style/amenity-points.mss
+++ b/style/amenity-points.mss
@@ -54,6 +54,10 @@
     [feature = 'amenity_shelter'] {
       marker-fill: @man-made-icon;
     }
+    [shelter_type = 'public_transport'] {
+      marker-file: url('symbols/amenity/shelter_public_transport.svg');
+      marker-fill: @transportation-icon;
+    }
     marker-fill: @accommodation-icon;
     marker-clip: false;
     [access != ''][access != 'permissive'][access != 'yes'] {
--- a/symbols/amenity/shelter_public_transport.svg
+++ b/symbols/amenity/shelter_public_transport.svg
@@ -1,0 +1,5 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<svg xmlns="http://www.w3.org/2000/svg" width="14" height="14" viewBox="0 0 14 14">
+  <path d="M11.227.004L2.693 2.922v11.103H4.18V9.021h.578l.166.616.734-.198-.476-1.78-.737.196.159.592H4.18V4.535l7.047-3.514V.004z"/>
+</svg>
+


imagico:
This suggestion clashes with a number of important principles of map design we tend to follow in OSM-Carto:

* *semantic homogeneity* - map elements that are similar in meaning and purpose for the target map user should be similar in appearance while elements that fulfill very different functions should look different.  The visual difference between different subtypes of shelters being larger than that between a shelter and an alpine hut or wilderness hut would therefore be fairly misleading.
* *function before physical appearance* - we like to focus on function in map rendering and not on physical appearance.  There is nothing in your symbol design that hints at the function of the feature in an explicit form without the mental detour of recognizing the similarity in physical shape to a very specific variant of the feature.
* *mind the target map user* - our target map user is the potential global map user and not the de facto dominating urban European/North American population.  That public transport shelters world wide predominantly have a form resembling the minimalist roof and wall urban tram stop type shelter is unlikely.
* *intuitive readability* - we like the map to be intuitively readable without a map key as much as possible.  That is the case for the shelter symbol, it is quite definitely often not the case for your suggestion.
* *avoiding confusion with mapped geometries* - we try to avoid symbols to be mistaken for mapped geometries.  The very abstract shape of your symbol of two lines at an angle is very prone to that.

It would be perfectly fine to use a variation of the shelter symbol to indicate a public transport shelter is tagged specifically as such with the variation not being that intuitive as long as the base symbol is still recognizable as the general shelter symbol and the variation is not directly misleading.

jeisenbe:
In Indonesia public transport shelters are more likely to look like this:

Pangkalan_ojek_Sinakma

That’s an ojek (motorcycle taxi?) stand, but local bus shelters are similar.

jeisenbe:
@jdhoek would you be interested in either modifying the main `amenity=shelter` icon or modifying this new icon so that they are more similar, and so the function is more clear, as mentioned in https://github.com/gravitystorm/openstreetmap-carto/pull/4317#issuecomment-774139211 ?

jdhoek:
Public transport shelters all share the characteristic that there is no guarantee that they are suitable for more than a modicum of shelter while waiting for a bus or tram. Most are open to at least one side and are size constrained, and the icon should indicate that the goal of this type of shelter is quite different from a generic `amenity=shelter`, and may not provide the map user with the shelter sought.

In many, if not most countries such shelters are variations on a roof on stilts without walls or a roof with some minimal walls on at most three sides. Representing it as a little house that will definitely keep you dry in stormy weather is misleading.

I don't think anything I can suggest will address the list of negative points raised by @imagico. In fact, if I apply those conditions to almost any of the existing icons, they all fail. It effectively discourages the contribution of any new icons. I mean, the current shelter icon fails on points 2, 3, and 4 too.

polarbearing:
> That public transport shelters world wide predominantly have a form resembling the minimalist roof and wall urban tram stop type shelter is unlikely.

Well is is meant to protect against the occasional rain shower and a bit from the wind. What else than a minimalist roof and some wall would be suitable for that? Closed buildings maybe, but they would probably be mapped as a bus station. 

I respect @imagico's perfectionism, but lacking a perfect solution a pragmatic one will help as well. The icon is in transport-blue, and typically found in proximity to the bus stop. Thus some clues for the map user to relate it correctly.

imagico (changes requested):
This needs revisiting the symbol design - it would need to be something that is universally recognizable as a subtype of shelter.  More guidance can be found in the comments above.

Blijbol:
I added the rain from the existing shelter symbol to make it more similar and easier to recognise as a shelter:

shelter_public_transport

If one further wants to minimise the difference, it could also retain the same shelter colour instead of adopting the public transport colour.

pnorman:
With the requested changes in https://github.com/gravitystorm/openstreetmap-carto/pull/4317#issuecomment-785636796 and https://github.com/gravitystorm/openstreetmap-carto/pull/4317#pullrequestreview-708936139 unaddressed, I'm going to go ahead and close this as abandoned.

PR #4330, opened 2021-02-10
Windows line ending fix

### LINKED ISSUE #4329: get-external-data.py fails to run on Windows
When running `docker-compose up import` on Windows, the script `get-external-data.py` fails.  The cause is that git on windows automatically converts LF to CRLF which causes an unexpected character to be added to the python shebang line.

I am submitting a PR which fixes this by adding python files to the list of files that should be excluded from automatic CRLF conversions.

ZeLonewolf:
Fixes #4329

Adds .py files to the list of files that should be excluded from git CRLF conversions.
This has been tested on Windows, but needs to be tested on Linux to ensure no unexpected behavior.


matkoniecz (approved):

matkoniecz:
Thanks for figuring it out and submitting PR!

matkoniecz:
It appears that it also closes  #4189

jeisenbe:
@ZeLonewolf thank you for this PR, it was quite helpful.
If you are interested in contributing futher, we have a list of [good first issues](https://github.com/gravitystorm/openstreetmap-carto/labels/good%20first%20issue) and also some more difficult issues where [help is needed](https://github.com/gravitystorm/openstreetmap-carto/issues?q=is%3Aopen+label%3A%22good+first+issue%22+label%3A%22help+wanted%22).

PR #4331, opened 2021-02-13
Render volcanoes with a distinct icon from peaks

### LINKED ISSUE #1790: render natural=volcano different from natural=peak
I mean in general design, the color is already different.

Reason: natural=volcano is meant (see [wiki](http://wiki.openstreetmap.org/wiki/Tag:natural%3Dvolcano)) and is practically used for all kind of volcanic features and not just for volcanic mountain peaks:

http://www.openstreetmap.org/node/2281252654
http://www.openstreetmap.org/node/1764264178
http://www.openstreetmap.org/node/1366173854
http://www.openstreetmap.org/node/270951619

Implying a peak in rendering is also confusing for the mapper since it contradicts the idea of placing the node at the volcanic opening or the middle of the main caldera.

It probably makes sense to render natural=volcano as label only.

ZeLonewolf:
Fixes #1790

Adds a new icon for volcanoes (`natural=volcano`), to make them distinct from the `natural=peak` icon.

Test rendering with links to the example place:
Location: https://www.openstreetmap.org/#map=13/19.3878/-155.2169

Before:
image

After:
image


jeisenbe:
Thank you for working on this, @ZeLonewolf. I agree that this is an important change. 

The initial rendering looks pretty good.

Since the tag natural=volcano is for a volcanic vent, which is not always at the top of a mountain but often in a low crater or fissure, could we try an icon that is slightly more crater-like and less peak-like? 

I also think the “smoke” / ash cloud could be adjusted, though I can’t quite say how. Perhaps thinner, more like the whisp of smoke on the man_made=chimney icon?

ZeLonewolf:
Were you thinking something more like this?
image

I was actually thinking we wanted something closer in shape to a spatter cone or stratovolcano, vice a more geologically accurate pit crater, since it's a more stereotypically recognizable volcano shape.  That's what I tried to make before I gave up and dropped in @imagico's icon, but I could be convinced either way.

ZeLonewolf:
Here's a version that simply combines the peak icon with the chimney icon's smoke trail:

image

Simple, but does the job.  Though it doesn't achieve the "pit" you're looking for.

ZeLonewolf:
And lastly, here's a version with a more flat-ish top and the smoke wisp from `man_made=smokestack`.
image

imagico:
Good idea to move forward with this.

When i implemented a similar change in the ac-style (https://github.com/gravitystorm/openstreetmap-carto/commit/99e573814e7307aad20adbc123c3130f1cc85966) i chose to

* eliminate the singular use color and chose to use the same color as for peaks and saddles (in particular important here where the red color clashes with the one used for healthcare).
* start the volcano symbol two zoom levels later and use a smaller version for the first zoom level and this way avoid filling the map with lots of heavy icons at low zoom levels in some areas, in particular since we have no meaningful prioritization, areas to test for example:

https://www.openstreetmap.org/#map=12/28.9590/-13.6977
https://www.openstreetmap.org/#map=12/14.9304/-24.3809
https://www.openstreetmap.org/#map=11/-2.8371/35.9469
https://www.openstreetmap.org/#map=13/37.7480/15.0072

ZeLonewolf:
Here are some sample renderings at z10-13 using small icons at z10 and regular icons below that.  I've also switched the volcanoes to use the same color as the regular peaks.  I'm not really a fan of losing the red color as it's visually a lot more difficult to see the difference now.

Z10:
v_z10
Z11:
v_z11
Z13:
v_z13

What do people think of this approach?

jeisenbe:
1) I agree with changing the color to match peaks and saddles. It does not make semantic sense that these would use the same color as healthcare features. 

2) The smaller icon is a clever idea, but it would need to be implemented with smaller icons for peaks at those zoom levels as well, as in https://github.com/gravitystorm/openstreetmap-carto/commit/99e573814e7307aad20adbc123c3130f1cc85966 - else the peaks will look too heavy in contrast, I think. 

3) I like what you tried in https://github.com/gravitystorm/openstreetmap-carto/pull/4331#issuecomment-778569432 - but could we perhaps try that crater shape plus the original ash cloud?

ZeLonewolf:
I've updated the PR to use the flat-ish top with the original wisp.  I think it looks pretty good like this.  Also, peaks and volcanoes now use the same style and zoom sections, as I brought them into a common stylesheet block.

z10:
image
z11:
image
z12:
image
z13:
image

imagico:
Like in case of https://github.com/gravitystorm/openstreetmap-carto/pull/4317#issuecomment-774139211 also here a few cartographic design principles we tend to follow here this change has issues with:

* We try to avoid combining several unrelated changes into one - in this case you extend rendering of peaks to z10 which has nothing to do with the change of the symbol for volcanoes (and would be a questionable change on its own as well - see areas like [here](https://www.openstreetmap.org/#map=11/42.5533/0.4185) or [here](https://www.openstreetmap.org/#map=11/44.0528/40.4451)).
* When changing symbol sizes, line widths etc. from one zoom level to the other it is advisable to enlage them by less than a factor of two.  This way you avoid the symbol geometries aligning to ground features across several zoom levels and therefore map users potentially mistaking symbols for real world geometries or getting the impression that they are zooming into a static image rather than a dynamic map.
* Symbols should always be readable for what they are meant to depict.  If a symbol is too small for that you either have to produce a simplified version that meets this requirement or you cannot show the symbol in that form.  As a general rule we typically design all symbols specifically for the target scale and don't reuse the same symbol in different sizes.
* When there is no meaningful prioritization (and in case of volcanoes the `ele` tag is not a meaningful criterion) the starting zoom level should be chosen so that the majority of features are shown at the starting zoom level already and no significant arbitrary selection takes place.

To me your tests seem to confirm that the approach taken (a pictorial point symbol to visualize volcanoes) does only work well in this style when you start the symbol significantly later than the abstract symbol for peaks - probably no earlier than z13.  But i would be happy to be proven wrong about this.

Also keep in mind that the whole point of the exercise is to solve #1790 and not render volcanoes any more in a way that implies they are a subtype of peaks.  I am not sure if your current proposal really accomplishes that.  It removes the additional color used for volcanoes (which is a good thing) but otherwise by rendering a symbol which clearly looks like a derivative of the peak symbol in exactly the same fashion to me still communicates this being a subtype of the same feature class.

ZeLonewolf:
What I think I'm hearing is that we want to:
* Start the rendering of peaks/volcanoes at zoom 10, with the triangle peak symbol
* At zoom 13, begin showing a difference in icon between peak and volcano.  This matches the same zoom level that you start to see crater rims, so that's visually consistent.

Would that be satisfactory?

imagico:
I don't think that would be a good idea.  As written in #1790 what is non-satisfying right now is that we render natural=volcano in a way that implies it is a type of peak.  It would be good to change that in a way that would communicate to the map user that it is not and at the same time it would be good to avoid using a color for the symbol that is very similar to the healthcare color.  If you'd render peaks and volcanoes identically at lower zoom levels that would communicate the opposite and reinforce the current conflation so i don't think that would be desirable.

My idea - which i showed in https://github.com/gravitystorm/openstreetmap-carto/commit/99e573814e7307aad20adbc123c3130f1cc85966 - was to show only peaks with the abstract triangle symbol starting at z11 and show volcanoes with a normal pictorial symbols starting later (at z13).  I encourage you to explore other options though and hope that the design principles i sketched might help a bit with that.

ZeLonewolf:
Can you clarify what you mean by the "abstract triangle symbol"?  Since it's in vector form, why would you need a separate image for the small version vice just shrinking it in CSS?

imagico:
The peak symbol is a simple triangle and is recognizable as such quite easily.  The volcano symbol OTOH is a pictorial symbol, it requires recognition of the specific and more complex shape and its meaning to be readable.

You don't absolutely need a separate symbol design for different scales in all cases but usually, especially if as recommended the scale difference is something other than a power of two, the best readability calls for varying the symbol depending on the scale it is to be rendered at.  When the symbol is rendered larger, details help recognition of the shape for what it is meant to depict while if it is shown smaller that additional detail will lead to blurriness and more difficult readability.

ZeLonewolf:
Okay, (try again) so what I think I'm hearing is:

1. Render peaks as they are currently (zoom 11+)
2. Render volcanoes from zoom 13+
3. Volcano icons don't look like peaks

Perhaps the volcano icon could follow a concept of "smoke rising from the ground"?

ZeLonewolf:
Here is a concept I've been playing with, to go for more of a "pit" style icon.  Thoughts?
z12
z13

HolgerJeromin:
> Here is a concept I've been playing with, to go for [...] icon. Thoughts?

I like the many birds in this rendering :-)

ZeLonewolf:
Well.  I can't unsee that now.  Back to the drawing board.

jeisenbe:
I think this looked good at z13 and higher zoom levels:


It was just the z11 / z12 that was a problem.

ZeLonewolf:
Unless I'm misreading @imagico's comments above, it sounded like there was an objection to iconography that indicated that volcanoes were a subtype of peak.

jeisenbe:
I believe that comment was in response to your plan to "Start the rendering of peaks/volcanoes at zoom 10, with the triangle peak symbol”, not in response to the new volcano icon.

imagico:
The peak symbol is eight pixels in size so working out something suitable in the standard 14 pixel symbol size should typically allow an impression sufficiently different from the peak symbol to make clear it is a fundamentally different type of feature.

I don't think the pit style design you showed is intuitively readable but that is in principle the kind of experiment i wanted to encourage.  The question is ultimately what does most clearly say 'volcano' that can work in a 14 pixel symbol.  The *cratered mountain with cloud* design you showed earlier in principle seems to work though it could probably be a bit more distinct from peaks in the base shape and it also appears to be a bit blurry as is.  See the ac-style version for comparison here:

http://www.imagico.de/map/pict/ac_samples/tenerife_z16.png

ZeLonewolf:
Ah, I was working off the volcano icon file to start from so it sounds like I'm using the wrong proportions.  14 pixels is considerably more breathing room.  Is there a style guide somewhere that I should be looking at?

ZeLonewolf:
Latest experiment -- "smoke rising from crater", now in 14 pixel version.

image

ZeLonewolf:
Below rendering is with a 14px version of the volcano with smoke icon.  This seems to be the preference.  I have the volcanoes starting to render at zoom=13.

z12
download (1)

z13
download (2)

z14
download (3)

Also, if anyone has any tips for the best way to generate these screen shots, I'd appreciated it.  Are you using the export function in kosmtik or just taking screen shots?

imagico:
I think i am probably not the best person to make a neutral assessment here having worked on the very same problem before and in doing that having already formed an opinion on the matter.

So i would encourage others to chime in here with comments on the specifics of the symbol design.

Regarding sample rendering productions - the most common method is screenshots, the export function makes it too easy to produce wrong scale rendering in my experience.

Nic787:
Hi,

jumping in here, because of the recent eruption in Iceland. I prefer having the red color personally, because it would help users differentiate normal peak and dangerous area (volcano). I love your flat peak with a bit of smoke.

imagico:
As said above in https://github.com/gravitystorm/openstreetmap-carto/pull/4331#issuecomment-789649795 someone else should review this (both maintainer and non-maintainer reviews welcome) since i am a bit biased here having addressed this problem in a different fashion elsewhere.

symbols/natural/volcano.svg
```diff
@@ -0,0 +1,77 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
```
pnorman, line 1: The [icon guildelines](https://github.com/gravitystorm/openstreetmap-carto/blob/master/CONTRIBUTING.md#map-icon-guidelines) say
> All new icons must be SVG format only. The SVG must be saved as standards compliant SVG without any proprietary tags. In Inkscape software, you will need to "Save As..." and choose the format Optimized SVG (preferable) or Plain SVG.

It doesn't look like that was done hyere by the amount of sodipodi stuff in the SVG.

style/amenity-points.mss
```diff
@@ -1365,16 +1365,19 @@
...
+    [feature = 'natural_peak'] {
+      marker-file: url('symbols/natural/peak.svg');    
+    }
+    [feature = 'natural_volcano'] {
+      marker-file: url('symbols/natural/volcano.svg');    
+    }
+    [zoom = 10] {
+      marker-width: 4;
+    }
   }
```
pnorman, line 1381: I would rather split this up into two blocks than put both peaks and volcanos ihn the same one. 

e.g.
```
[feature = 'natural_peak'][zoom >= 10] {
...
}
[feature = 'natural_volcano'][zoom >= 10] {
...
}
```

pnorman (changes requested):

symbols/natural/volcano.svg
pitdicker: @ZeLonewolf If you want I have a simplified and pixel-aligned version in https://github.com/pitdicker/openstreetmap-carto/commit/3b89a05d7f3b9ad74739d158b0c4905d72fc04fb.

ZeLonewolf: Thanks @pitdicker!

ZeLonewolf:
Updated with changes requested.  However, due to the issues noted at https://github.com/gravitystorm/openstreetmap-carto/pull/4381#issuecomment-917513271, I am not able to run this PR.  I temporarily removed the golf code in order to generate these render samples.

**Test renders, zoom 10-13**:
image
image
image
image

style/amenity-points.mss
ZeLonewolf: done!

ZeLonewolf:
Closing so other contributors can work on this.

PR #4341, opened 2021-03-04
Update conditions of two indexes

Nakaner:
This pull request updates the where conditions of indexes defined in `indexes.yml` and `indexes.sql`. The queries of the layers using the indexes have been changed in previous commits but the changes did not get reflected in the index creation scripts.

The performance gain is not expected to be large because the indexes are still useful for the layers they have been created for. However, it makes sense to me to make the index conditions match the conditions of their layers exactly if they have been created for them.

People who set up the style after this patch can benefit from slightly improved performance while people already running the style can decide whether they want to change the indexes or use the time for other purposes.

This pull request has no visual effects on rendering. It follows the spirit of my previous pull requests of this kind: #3909, #3682.

Note: The water areas index uses the condition `building IS NULL`. This condition was added long time ago in order to make use of the nobuilding index. In the meanwhile, someone added a special water index. Instead of adding `building IS NULL` to the water index's condition, I could remove `building IS NULL` from the condition of that layer as well but this can lead to mistagged water areas appear on the map if they are not rendered (and covered) by the buildings layer as well.

### DIFF
--- a/indexes.sql
+++ b/indexes.sql
@@ -44,7 +44,7 @@ CREATE INDEX planet_osm_roads_admin
   WHERE boundary = 'administrative';
 CREATE INDEX planet_osm_roads_admin_low
   ON planet_osm_roads USING GIST (way)
-  WHERE boundary = 'administrative' AND admin_level IN ('0', '1', '2', '3', '4');
+  WHERE boundary = 'administrative' AND admin_level IN ('0', '1', '2', '3', '4') AND osm_id < 0;
 CREATE INDEX planet_osm_roads_roads_ref
   ON planet_osm_roads USING GIST (way)
   WHERE highway IS NOT NULL AND ref IS NOT NULL;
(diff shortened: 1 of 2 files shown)


pnorman (changes requested):
CI is failing because indexes.sql does not match the index definitions. You can fix this with `scripts/indexes.py > indexes.sql`.

> This pull request updates the where conditions of indexes defined in `indexes.yml` and `indexes.sql`. The queries of the layers using the indexes have been changed in previous commits but the changes did not get reflected in the index creation scripts.

No, the indexes were written [in 2011 in the old XML stylesheets](https://github.com/openstreetmap/mapnik-stylesheets/commit/17a371b5cf53be4701e7497476c067c0b66a4e24) and the mismatch between the SQL and newer indexes is intentional.

I would prefer to not add add the `building IS NULL` constraint to the index. For the osm_id constraint, I don't see a real advantage or disadvantage to adding it.

pnorman:
Closing as abandoned.

PR #4353, opened 2021-03-26
Rendering for manhole covers

ZeLonewolf:
This PR adds rendering for manhole covers at zoom 19.  

There is [growing usage](https://taginfo.openstreetmap.org/tags/man_made=manhole#chronology) of this tag, with over 76,000 usages globally.  Rendering manhole covers are useful for clearing flooded streets (when manhole drains can be covered by flood waters), and are useful as navigational landmarks at the highest zoom levels.

Sample locations:
[Berlin, Germany](https://www.openstreetmap.org/#map=19/52.47929/13.43034)
[Rhode Island, USA](https://www.openstreetmap.org/#map=19/41.65523/-71.45390)

Before
Berlin, Germany
Rhode Island, USA

After
Berlin, Germany
Rhode Island, USA


symbols/man_made/manhole.svg
```diff
@@ -0,0 +1,4 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<svg height="4" width="4">
+  <circle cx="2" cy="2" r="2" fill="black" />
```
HolgerJeromin, line 3: Current dot renderings (shop=* and others) does not need an svg file but uses native mapnik circles.

polarbearing:
I would be sceptical about the significance of this feature for a generalised map.

imagico:
Cartographically i don't think this is going to work.  Rendering of manholes would only be workable IMO in combination with a much more sophisticated and much more expensive high zoom level cartography than we currently have (and also than we can afford ressources-wise with a real time rendered map).

At the same time the practical value in a general purpose map is rather limited.

More technical feedback - this style of display would lead to strong confusion with other features displayed with a dot - in particular bollards.  Also use of opacity for the dot will lead to confusing color mixing on different backgrounds.

mvexel:
I'm with @polarbearing and @imagico on this one. Man hole covers don't have a place on a general purpose map at this zoom level. From a cartographic point of view, the chosen symbol is too generic.

pnorman:
Thanks  for the PR, but I'm with the others on manhole covers not having a place on a general purpose maps, as well as the problems with this particular symbol causing confusion.

Issue #3977, opened 2019-11-21
Feature removal policy


imagico:
When adding new features in this style our obvious and natural policy is that someone being willing to develop a concrete PR for adding the feature is prerequisite for it actually being added.  When removing features however the actual code change to remove them is usually fairly simple and the hurdle that in most cases prevents the removal is opposition to the change.

This opposition sometimes is fairly non-productive because no concrete suggestions are made how to resolve the issues that suggest the removal of said feature but instead just call for keeping the status quo without resolving the issues or call non-specifically for resolving the problems without removing the feature.  #3714/#3969 is a good example for that.

My idea to address this problem and to allow us better to manage the complexity in this style and to keep both code and design manageable is to move to a policy that opposition to feature removal PRs should be accompanied by commitment to develop the feature in question in a way that achieves consensus within the project.

This in my eyes could help us developing and maintaining consensus on design questions and it would also encourage developers to adopt certain features and take up the responsibility to maintain them, fix problems and to keep their rendering compatible with the rest of the style and the goals of the project.

Note my idea is not for this rule to supersede the consensus principle but for us (the maintainers) to agree that we will not oppose a feature removal PRs against a majority of the other maintainers if neither we nor anyone else is willing to address the problems because of which the feature removal is suggested by other means.

Related to #3951.

kocio-pl:
Removing features is already far easier than adding them or updating. I see no reason to further increase this (unfortunately, quite natural) huge lack of balance. IMO each change deserves the same amount of scrutiny, so the opposite direction is needed (to make the amount of work more similar), not the shortcuts for one special category of changes. However thanks for the clear description of what you see as the problem in one place.

Consensus rules as discussed in #3951 have their own load of problems and we are rather on the beginning of the road, as more fundamental things are not even addressed, so at this moment this proposition goes too far in this respect.

imagico:
> Removing features is already far easier than adding them or updating.

The opposite is the case - and it has been this way throughout the history of this style.  See for example https://github.com/gravitystorm/openstreetmap-carto/issues/1630#issuecomment-384056965.

Feature removals are extremely rare - usually only something like three or four per year and that are often just removal of deprecated synonyms.  This year we had AFAIK so far only #3718, #3931 and #3829.  Removals of features because of cartographic prioritization or for quality issues with the chosen rendering are almost non-existent.

kocio-pl:
[...]

PR #4357, opened 2021-04-02
Update README.md on v5

kocio-pl:
Added the line about major changes in v5.


imagico (approved):
This looks ok to me - however note this section of the readme is titled *roadmap* - over time however this has more degraded into a changlog summary.  It might make sense to title it as such.

This points to the issue that for quite some time we have had no consensus among the maintainers about the strategic direction of this style.  See #3951, #3977, #2270, #1975, #3635, #1630.  Us not being able to present interested volunteers with consensus on a clear strategic vision for this style severely hampers recruitment of qualified volunteers.

PR #4361, opened 2021-04-06
Update project.mml for railway:preserved=yes parity with railway=preserved

### LINKED ISSUE #2615: railway:preserved=yes + railway=rail not rendering as preserved
There will be more and more use of this tag so that one can also add what kind of preserved railway it is.
Basically railway:preserved=yes + railway=*

Suggest add this in so that it will work like the more common railway=preserved.

Better option would be to just move over to railway:preserved=yes only to encourage better tagging.

kmpoppe:
Fixes #2615

# Changes proposed in this pull request #
`railway:preserved=yes` should become the preferred option over `railway=preserved`. A [proposal is underway](https://wiki.openstreetmap.org/wiki/Proposed_features/Discourage_railway%3Dpreserved) to discourage the use of the latter.

Very active railway mappers were concerned, that `railway:preserved` isn't rendered as of yet, this is the first try to fix these concerns. 

# Test rendering with links to the example places #
https://www.openstreetmap.org/#map=18/52.66410/8.64290&layers=N

## Before ##
grafik

## After ##
grafik

# Unresolved issue #

I wasn't able to display the `name=*` next to the line, just like it does next to `railway=preserved` lines. I would need help in fixing this.

Thanks everyone for your support!

^Kai

### DIFF
--- a/project.mml
+++ b/project.mml
@@ -487,7 +487,7 @@ Layer:
             UNION ALL
             SELECT
                 way,
-                'railway_' || (CASE WHEN railway = 'preserved' AND service IN ('spur', 'siding', 'yard') THEN 'INT-preserved-ssy'::text
+                'railway_' || (CASE WHEN (tags->'railway:preserved' IN ('yes') OR (railway = 'preserved' AND service IN ('spur', 'siding', 'yard'))) THEN 'INT-preserved-ssy'::text
                                  WHEN (railway = 'rail' AND service IN ('spur', 'siding', 'yard')) THEN 'INT-spur-siding-yard'
                                  WHEN (railway = 'tram' AND service IN ('spur', 'siding', 'yard')) THEN 'tram-service'
                                  ELSE railway END) AS feature,
@@ -743,7 +743,7 @@ Layer:
             UNION ALL
             SELECT
                 way,
-                ('railway_' || (CASE WHEN railway = 'preserved' AND service IN ('spur', 'siding', 'yard') THEN 'INT-preserved-ssy'::text
+                ('railway_' || (CASE WHEN (tags->'railway:preserved' IN ('yes') OR (railway = 'preserved' AND service IN ('spur', 'siding', 'yard'))) THEN 'INT-preserved-ssy'::text
                                      WHEN (railway = 'rail' AND service IN ('spur', 'siding', 'yard')) THEN 'INT-spur-siding-yard'
                                      WHEN (railway = 'tram' AND service IN ('spur', 'siding', 'yard')) THEN 'tram-service'
                                      ELSE railway END)) AS feature,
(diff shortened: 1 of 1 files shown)


jdhoek:
I can't judge the edits you've done, but when I scanned through `project.mml` yesterday I noticed that `railway=preserved` is also filtered out for low-zoom rendering on line 881. You might want to change that query as well to also exclude `tags->'railway:preserved' IN ('yes')` and verify that the same behaviour applies.

kmpoppe:
Hey @jdhoek, thanks for reminding me. I changed that line but I didn't see any difference - I might have CTRL+Z'd once too often then ;-)
I hope that there's a "help wanted" label here, so that we can manage to get this solved sooner than later :)

kmpoppe:
Hey @gravitystorm, would you be able to direct me to some assistance with the one or two issues that have arisen from this change (Displaying name next to the line and finding a way in testing the `roads-low-zoom´-layer)?

imagico:
Thanks for the pull request, this is definitely something we need to change considering the shift in tag use by mappers observable.

I don't understand the issue of testing roads-low-zoom - you need to elaborate on that problem.

Railway labels are handled in layer `railways-text-name`.

Regarding the change itself:

* we don't want to encourage proliferation of taggings so we would likely prefer removing rendering of `railway=preserved` if and when we add support for `railway:preserved=yes`.
* we want the style to be intuitively understandable for mappers - hence it seems problematic to me if we change and unify the basic styling of features tagged `railway=*` based on the presence of a secondary tag.  Secondary tags should normally be indicated by a styling variation while maintaining the distinction based on the primary tag.
* we definitely don't want a catchall rendering of everything with `railway:preserved=yes` no matter of the value of `railway=*`, in particular things like `railway=razed` and `railway=abandoned` - see also https://github.com/gravitystorm/openstreetmap-carto/issues/2615#issuecomment-599210182.

kmpoppe:
Thank you for your insightful responses @imagico!

* `roads-low-zoom`: I couldn't correctly interpret what exactly the rendering difference was supposed to be. Could you show me two screenshots what the difference should be so that I could check that?
* `railway-text-name`: Will look into it, thanks!
* If I read your points correctly, they lead to this conclusion: create a `railway:preserved=yes` special style for all different `railway=*`-styles, which is ideally so unique and recognizeable at the same time, that `railway=preserved` can safely be removed from rendering, right?
^Kai

imagico:
roads-low-zoom is not for a different styling but for rendering efficiency.  Current starting zoom level for rendering railway=preserved is z12 so it is not relevant in roads-low-zoom anyway.

> If I read your points correctly, they lead to this conclusion: create a railway:preserved=yes special style for all different railway=*-styles, which is ideally so unique and recognizeable at the same time, that railway=preserved can safely be removed from rendering, right?

Ideally it should be visible that the styling railway=rail + railway:preserved=yes is related to the styling railway=rail and that railway=rail + railway:preserved=yes is related to railway=narrow_gauge + railway:preserved=yes.  This is very hard considering the variety of line signatures we already have in this style.  If that turns out to be impossible the solution should be not rendering railway:preserved=yes rather than dropping the distinction between railway=* in favor of rendering the secondary tag railway:preserved=yes in a unified fashion aggregating physically very different kinds of railways just based on how they are used.

pnorman:
@kmpoppe Are there additional questions you have in order to get this PR out of a draft state?

Although not essential, you might want to rebase your PR on the latest version.

kmpoppe:
@pnorman the problem was reaching consensus on the tagging-ML which I thought was vital - I also tried encouraging openrailway-ML to come forward and voice their ideas on how the "correct" tagging should be henceforth but there wasn't any concrete answer.

So, after all, this is WIP because my idea didn't pan out as I thought it would and I didn't have the time to discuss this through and through.

Issue #3143, opened 2018-03-24
Change meadow and grassland rendering


Tomasz-W:
Currently, this (landuse=grass):
![](https://www.lawnstarter.com/blog/wp-content/uploads/2015/03/grass-small.png)

is rendered in the same way as this (landuse=meadow):
![](https://i2.wp.com/www.christchurchdailyphoto.com/wp-content/uploads/2018/01/DSC02482.jpg)

or this (natural=grassland):
![](https://img-aws.ehowcdn.com/877x500p/photos.demandstudios.com/getty/article/241/7/470864887.jpg)

As grass can reach max. 15-20 cm, meadows and grasslands can reach more than 50 cm. I think they are different levels of landcover, so they should be distinguished on map. 
For example - you can walk or ride a bike through grass without problem, but is more problematic in a meadow/ grassland, so rendering this features in the same way may be confusing.

My opening proposition is to use wetland pattern - change it's colour to black, turn 90 degree, and set opacity to 15/25%.

Photoshop mock-up:
![v1](https://user-images.githubusercontent.com/25656654/37863262-c12fc462-2f5b-11e8-95bd-091c91620f28.png)

dieterdreist:
sent from a phone

> On 24. Mar 2018, at 12:09, Tomasz Wójcik <notifications@github.com> wrote:
> 
> Currently, this (landuse=grass):
> 


actually this isn’t the correct application of landuse=grass according to the wiki:
https://wiki.openstreetmap.org/wiki/Tag:landuse%3Dgrass

kocio-pl:
I like this idea. Anybody willing to make the code and test it a bit?

Tomasz-W:
Another options (click to view full size):
* 50% transparent `scree_overlay` pattern (lower left square)
* 100% `beach_coarse` pattern (lower right square)

![meadow](https://user-images.githubusercontent.com/25656654/44992253-0692b300-af97-11e8-8166-9eb3f9d5f7eb.jpg)

imagico:
For reference - the question of differentiating natural=grassland and landuse=meadow was previously discussed in #2433

Tomasz-W:
@Adamant36 

Here is a file with PNG pattern which worked the best set at 10% overlaying `landuse=grass` colour in my Photoshop tests, but it needs a real test renderings.

![meadow](https://user-images.githubusercontent.com/25656654/45013065-17383d00-b01a-11e8-8864-de3c071c2cbc.png)

[...]

PR #4381, opened 2021-04-21
Update leisure=golf_course color and render golf=* features (gravitystorm#661)

### LINKED ISSUE #661: Add rendering for golf=*

jgruca:
_(Comment updated June 5 to current state of PR, test renderings also updated)_

Fixes #661

- Change leisure=golf_course/miniature_golf_course to `@campsite`, eliminating `@golf_course` color
- Update `landcover` layer to include golf area features
- Update `amenity-points` layer to include golf point features
- Add `golf-line` layer
- Add `style/golf.mss` and specify rendering for:
  - `golf=tee`/`fairway`/`driving_range` (`@grass`, z>=12)
  - `golf=green` (`@pitch`, z>=13)
  - `golf=bunker` (`@sand`, z>=13)
  - `golf=rough` (`@grass`, z>=12; with texture, z>=15)
  - `golf=hole` as way (solid `@address-color` line, labeled with ref or name, z>=16)
  - `golf=hole` as point and `golf=pin` (`symbols/golf_pin.svg`; labeled with `ref` in `@leisure-green`, z>=16)

See #4212 for previous discussion of how these colors and textures were chosen. Corrections, feedback, critiques welcome.

Test rendering with links to the example places (before/after):

https://www.openstreetmap.org/#map=15/41.7849/-93.7247

z15
loc1-z15
z16
loc1-z16
z17
loc1-z17

https://www.openstreetmap.org/#map=14/41.8705/-93.2144

z14
loc2-z14
z15
loc2-z15
z16
loc2-z16
z17
loc2-z17

https://www.openstreetmap.org/#map=15/41.2767/-95.7367

z15
loc3-z15
z16
loc3-z16
z17
loc3-z17


imagico (changes requested):
I think the polygon fill colors are overall a good choice - though the rough color is IMO too close to scrub color and too different from grass color.  A rough is AFAIK usually still just longer grass so a much darker color indicating something different is a bit misleading.  See the following comparison with how the ac-style renders it:

this
ac-style

I also don't think it is a good idea to add further labels/linework/symbols in green color that are not strictly vegetation related.  We would most certainly maneuver us seeing eyes into a situation similar to that with blue cycleways conflicting with the use of blue for water related features.

The hole symbol not being centered at the base of the flag seems to be a bit confusing to me.

jgruca:
> I think the polygon fill colors are overall a good choice - though the rough color is IMO too close to scrub color and too different from grass color.

I've updated the rough color; see screenshots below.

> I also don't think it is a good idea to add further labels/linework/symbols in green color that are not strictly vegetation related.

I tried changing these to `@admin-boundaries` and like the appearance; it also makes a little sense to me to use a boundary color for the play path since it is not observable on the ground, similar to a boundary. However, I'm open to suggestions on the best color to use here.

> The hole symbol not being centered at the base of the flag seems to be a bit confusing to me.

This appears to be a [bug in mapnik](https://github.com/mapnik/mapnik/issues/1122) with a workaround. I've adjusted the SVG file so the bounding box of the marker is calculated correctly.

Updated images:

https://www.openstreetmap.org/#map=15/41.7849/-93.7247

z15
loc1-z15
z16
loc1-z16
z17
loc1-z17

https://www.openstreetmap.org/#map=14/41.8705/-93.2144

z14
loc2-z14
z15
loc2-z15
z16
loc2-z16
z17
loc2-z17

https://www.openstreetmap.org/#map=15/41.2767/-95.7367

z15
loc3-z15
z16
loc3-z16
z17
loc3-z17

project.mml
```diff
@@ -1044,6 +1047,19 @@ Layer:
...
+  - id: golf-line
+    geometry: linestring
+    <<: *extents
+    Datasource:
+      <<: *osm2pgsql
+      table: |-
+        (SELECT
+            way, tags->'golf' AS golf, ref, name
+          FROM planet_osm_line
+          WHERE tags->'golf' = 'hole' AND way && !bbox!
```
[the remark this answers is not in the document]
jgruca: Fixed.

style/landcover.mss
```diff
@@ -66,7 +66,6 @@
 @pitch: #aae0cb;           // Lch(85,22,168) also track
 @track: @pitch;
 @stadium: @leisure; // also sports_centre
-@golf_course: #b5e3b5;
```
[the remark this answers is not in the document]
jgruca: Fixed.

style/golf.mss
```diff
@@ -0,0 +1,62 @@
+#landcover[zoom >= 12] {
+  ::high-zoom[zoom >= 12] {
```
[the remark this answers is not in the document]
jgruca: Could you clarify what this means? Would you prefer all of these rules be moved to the `landcover.mss` file?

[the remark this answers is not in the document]
imagico: No, just that if you use an attachment (the `::high-zoom`) that will be rendered outside the normal drawing order of the layer - which we don't want to happen here.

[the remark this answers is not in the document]
jgruca: I see. I need some help, then. Removing the attachment here makes these changes disappear. I specified `::high-zoom` because that was specified on the [existing `#landcover` layer](https://github.com/gravitystorm/openstreetmap-carto/blob/master/style/landcover.mss#L76-L77), and unless I directly place my new layers in `landcover.mss`, I'm not sure how to make sure these new styles are painted on top of the other landcover styles.

I see that in your [ac style](https://github.com/imagico/osm-carto-alternative-colors/blob/master/golf.mss) you do not have to specify an attachment, so maybe I just need help understanding how mapnik is ordering these layers internally.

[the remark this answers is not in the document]
imagico: Oh, sorry, forgot about the fading, yes, you have to use the same attachment as the other landcover features then (and we should really get rid of the fading of course).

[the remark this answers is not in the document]
jgruca: Ah, thanks for the clarification.

neuhausr:
I like this! My only criticism is that the color of the greens seems a little too close to the color of the water. Maybe modify it to be less blue. #B5E3B5, which I think is the current carto golf course fill, might work?

Also a question: if a hole has a name _and_ a ref, how will that be handled? It would probably be ideal to include both ref and name, at least at higher zoom levels.  (For example see the [Jubilee Course](http://tile.openstreetmap.fr/?zoom=16&lat=48.72648&lon=2.60806) at St Andrew's)

jgruca:
> My only criticism is that the color of the greens seems a little too close to the color of the water.

I agree. This came up [in the previous PR too](https://github.com/gravitystorm/openstreetmap-carto/pull/4212#issuecomment-708108239), with [@imagico agreeing to some extent](https://github.com/gravitystorm/openstreetmap-carto/pull/4212#issuecomment-708184878). I'm trying to build consensus, though, and so far consensus seems to mean `@pitch` for the green. I'd be happy to hear from @imagico and @jeisenbe on acceptable alternatives; `#B5E3B5` does look decent, as does `@pitch` sent through `polygon-comp-op: lighten`. But both of those options introduce  single-use shades of green back into the style.

> Also a question: if a hole has a name and a ref, how will that be handled? It would probably be ideal to include both ref and name, at least at higher zoom levels.

It looks like the tile.openstreetmap.fr renderer will display hole `name` on lines and fall back to `ref`, so I have copied that behavior on the latest commit. Thanks for pointing out Jubilee Course as an example, thought, because that made me realize my PR did not correctly render `ref` on `golf=pin` (a less common combo, but perhaps what you were talking about for a hole with a `name` and `ref`). I have fixed that as well.

image

imagico:
If the pitch color has issue with being too similar to other colors with a very different meaning the right way is to change the pitch color, not to introduce a new color to address the issue only for new color uses.  Current pitch color was selected in #2363.

I don't think introducing yet another point symbol color and overloading the dark purple for administrative boundaries for other stuff are a good idea.  Generally speaking dark purple is always a very problematic color where mixing with other (background) colors happens because it leads to all kind of weird colors in combination with the green tones in many areas dominating the base map (in particular also on golf courses).  For the lines i would probably go either with a neutral color or try a very feint green drawn directly above the landcover and water layers.  Not sure if that would work though, just an idea.

jgruca:
Thanks for the input. Using `@address-color` (`#666666`, dark gray) for the lines/labels/markers seems like a good neutral yet visible color, and semantically the hole ref seems similar to its address.

image

image

> [...] the right way is to change the pitch color [...]

I don't disagree. Is that something that should be proposed in a separate PR?

jgruca:
I updated the first comment in this PR to match the current state of the PR. Comments welcome.

imagico:
Like with #4331 this should be reviewed by someone else because i am a bit prejudiced by having picked a different design approach to this elsewhere.

michaelabon:
I'd love to have something like this merged, as someone who loves mapping golf courses. Most existing golf courses I see are mapping for the renderer, despite [the wiki saying](https://wiki.openstreetmap.org/wiki/Tag:leisure%3Dgolf_course):

> Do not use the tag `landuse=grass` in order to make the rendering more appealing. Use `surface=grass`.

I don't make sweeping changes to existing golf courses when I see people having used `landuse=grass`, but it makes me a little sad every time I see it. This PR would go a long way towards helping address the mis-tagging issue.

pnorman:
Reviewing isn't just for maintainers. Everyone can review.

vincentvd1:
I like the new color. Maybe one thing to consider. In issue #3143 it was also proposed to change the greens in order to be able to render landuse=meadow and natural=grassland differently. There it was proposed to give campsite the same color as the golfcourse (the old color) and use the campsite color for landuse=meadow ( you now used the old campsite color or golf courses). 

I don't have a problem with the new golf course color, it is just because the limited amount of available greens, it is good to also take this into account. Maybe (in a new PR for #3143) the new golf course color can be used for both golf courses and meadows. The campsite need to be changed to the green of leisure (see image at the end of #3143).

project.mml
```diff
@@ -2024,7 +2041,8 @@ Layer:
               OR leisure IN ('slipway', 'track')
               OR waterway IN ('dam', 'weir')
               OR "natural" IN ('arete', 'cliff', 'ridge'))
-            AND name IS NOT NULL
+            AND name IS NOT NULL)
+            OR tags @> 'golf=>hole'
```
pnorman, line 2045: This is going to have negative performance impacts since it will be unable to use the [planet_osm_line_name index](https://github.com/gravitystorm/openstreetmap-carto/blob/a2077c0c8d40fb7f6d308b4ca4e8941d4c8b699a/indexes.sql#L7-L9)

style/golf.mss
```diff
@@ -0,0 +1,76 @@
...
+    text-placement: line;
+    text-size: 11;
+    text-fill: @address-color;
+    text-face-name: @book-fonts;
+    text-halo-radius: @standard-halo-radius;
+    text-halo-fill: @standard-halo-fill;
+    text-name: "";
+
+    [ref != ''] { text-name: "[ref]"; }
+    [name != ''] { text-name: "[name]"; }
```
pnorman, line 45: Set one of these as default, then check if that one is null, and if so, set the other.

symbols/golf_pin.svg
```diff
@@ -0,0 +1,9 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<svg version="1.1"
+    id="golf_pin"
+    xmlns:svg="http://www.w3.org/2000/svg"
+    xmlns="http://www.w3.org/2000/svg"
+    width="9"
+    height="13">
+   <path d="M 4 1 L 4 10.773438 A 2.107161 0.75 0 0 0 2.3925781 11.5 A 2.107161 0.75 0 0 0 4.5 12.25 A 2.107161 0.75 0 0 0 6.6074219 11.5 A 2.107161 0.75 0 0 0 5 10.773438 L 5 4 L 9 2.5566406 L 5 1 L 4 1 z"/>
```
pnorman, line 8: needs review for pixel alignment

project.mml
```diff
@@ -1044,6 +1047,19 @@ Layer:
...
+  - id: golf-line
+    geometry: linestring
+    <<: *extents
+    Datasource:
+      <<: *osm2pgsql
+      table: |-
+        (SELECT
+            way, tags->'golf' AS golf, ref, name
+          FROM planet_osm_line
+          WHERE tags->'golf' = 'hole'
```
pnorman, line 1059: Use `tags @> 'golf=>hole'` like elsewhere

style/golf.mss
```diff
@@ -0,0 +1,76 @@
...
+#amenity-points[zoom >= 16] {
+  [feature = 'golf_hole'],
+  [feature = 'golf_pin'] {
+    marker-file: url('symbols/golf_pin.svg');
+    marker-fill: @address-color;
+    marker-transform: translate(2, -5);
+  }
+  [zoom >= 17][feature = 'golf_hole'],
+  [zoom >= 17][feature = 'golf_pin'] {
+    marker-transform: "scale(1.5) translate(2, -5)";
```
pnorman, line 60: `scale(1.5)` will cause problems with pixel alignment

style/golf.mss
```diff
@@ -0,0 +1,76 @@
...
+    [zoom >= 17] { text-size: 13; }
+  }
+}
+
+#amenity-points[zoom >= 16] {
+  [feature = 'golf_hole'],
+  [feature = 'golf_pin'] {
+    marker-file: url('symbols/golf_pin.svg');
+    marker-fill: @address-color;
+    marker-transform: translate(2, -5);
```
pnorman, line 56: marker-transform ignores map scale factor. Does this still work at high DPI?

pnorman (changes requested):
I still need to review the cartography, but found some technical issues.

project.mml
jgruca: What's the best way to address this?

style/golf.mss
jgruca: Updated.

symbols/golf_pin.svg
jgruca: Any guidelines on pixel alignment? Here is the SVG with a 1px grid overlaid:

image

project.mml
jgruca: Updated.

style/golf.mss
jgruca: I've removed this scaling altogether.

style/golf.mss
jgruca: I've switched this to `marker-geometry-transform`, which does not have that caveat in the docs. Happy to test at a higher DPI, just need a pointer on how to do that.

project.mml
pnorman: I'm thinking `WHERE ([existing conditions] AND name IS NOT NULL) OR ([new conditions] AND ref IS NOT NULL)` then rename the planet_osm_line_name index to planet_osm_line_label and make it have the condition `name IS NOT NULL OR ref IS NOT NULL`, but it could use more thought and testing.

imagico: I concur - we also have other queries from planet_osm_line which have a `ref IS NOT NULL` condition, that is `roads-text-ref` and `roads-text-ref-minor`.

jgruca: I've updated the index defined in `indexes.sql`/`indexes.yml` and reformatted the query's condition as described.

jgruca: I ran `EXPLAIN` on a sample query, and it does appear the index is being used and has a positive effect on query cost.

With only `planet_osm_line_name`:
```
 Bitmap Heap Scan on planet_osm_line  (cost=55.42..3376.13 rows=7 width=85)
   Recheck Cond: (way && '<snip>'::geometry)
   Filter: ((((man_made = ANY ('{pier,breakwater,groyne,embankment}'::text[])) OR ((man_made = 'pipeline'::text) AND ((tags -> 'location'::text) =
ANY ('{overground,overhead,surface,outdoor}'::text[]))) OR (bridge = ANY ('{yes,aqueduct,cantilever,covered,trestle,viaduct}'::text[])) OR (tags @>
 '"attraction"=>"water_slide"'::hstore) OR (aerialway = ANY ('{cable_car,gondola,mixed_lift,goods,chair_lift,drag_lift,t-bar,j-bar,platter,rope_tow
,zip_line}'::text[])) OR (leisure = ANY ('{slipway,track}'::text[])) OR (waterway = ANY ('{dam,weir}'::text[])) OR ("natural" = ANY ('{arete,cliff,
ridge}'::text[]))) AND (name IS NOT NULL)) OR ((tags @> '"golf"=>"hole"'::hstore) AND (ref IS NOT NULL)))
   ->  Bitmap Index Scan on planet_osm_line_way_idx  (cost=0.00..55.42 rows=951 width=0)
         Index Cond: (way && '<snip>'::geometry)
```
After creating `planet_osm_line_label`:
```
 Bitmap Heap Scan on planet_osm_line  (cost=18.47..1116.75 rows=7 width=85)
   Recheck Cond: ((way && '<snip>'::geometry) AND ((name IS NOT NULL) OR (ref IS NOT NULL)))
   Filter: ((((man_made = ANY ('{pier,breakwater,groyne,embankment}'::text[])) OR ((man_made = 'pipeline'::text) AND ((tags -> 'location'::text) =
ANY ('{overground,overhead,surface,outdoor}'::text[]))) OR (bridge = ANY ('{yes,aqueduct,cantilever,covered,trestle,viaduct}'::text[])) OR (tags @>
 '"attraction"=>"water_slide"'::hstore) OR (aerialway = ANY ('{cable_car,gondola,mixed_lift,goods,chair_lift,drag_lift,t-bar,j-bar,platter,rope_tow
,zip_line}'::text[])) OR (leisure = ANY ('{slipway,track}'::text[])) OR (waterway = ANY ('{dam,weir}'::text[])) OR ("natural" = ANY ('{arete,cliff,
ridge}'::text[]))) AND (name IS NOT NULL)) OR ((tags @> '"golf"=>"hole"'::hstore) AND (ref IS NOT NULL)))
   ->  Bitmap Index Scan on planet_osm_line_label  (cost=0.00..18.47 rows=292 width=0)
         Index Cond: (way && '<snip>'::geometry)
```

style/golf.mss
pitdicker: @jgruca I made a pr to your branch that handles the aligning in the SVG file, so you don't need to do transformations in the style. https://github.com/jgruca/openstreetmap-carto/pull/1

project.mml
```diff
@@ -1044,6 +1047,19 @@ Layer:
...
       cache-features: true
       minzoom: 11
+  - id: golf-line
+    geometry: linestring
+    <<: *extents
+    Datasource:
+      <<: *osm2pgsql
+      table: |-
+        (SELECT
+            way, tags->'golf' AS golf, ref, name
```
pnorman, line 1057: `ref` and `name` are not used in this layer and should be removed.
```suggestion
            way, tags->'golf' AS golf
```

project.mml
```diff
@@ -2007,15 +2024,15 @@ Layer:
         (SELECT
           way,
             NULL as way_pixels,
-            COALESCE('aerialway_' || aerialway, 'attraction_' || CASE WHEN tags @> 'attraction=>water_slide' THEN 'water_slide' END, 'leisure_' || leisure, 'man_made_' || man_made, 'waterway_' || waterway, 'natural_' || "natural") AS feature,
+            COALESCE('aerialway_' || aerialway, 'attraction_' || CASE WHEN tags @> 'attraction=>water_slide' THEN 'water_slide' END, 'leisure_' || leisure, 'man_made_' || man_made, 'waterway_' || waterway, 'natural_' || "natural", CASE WHEN tags @> 'golf=>hole' THEN 'golf_hole' END) AS feature,
```
pnorman, line 2027:
```suggestion
            COALESCE('aerialway_' || aerialway, 'attraction_' || CASE WHEN tags @> 'attraction=>water_slide' THEN 'water_slide' END, 'leisure_' || leisure, 'man_made_' || man_made, 'waterway_' || waterway, 'natural_' || "natural", 'golf_' || tags->'golf') AS feature,
```
You want to keep the coalesce formatted the same as the others

pnorman (changes requested):
Just a couple minor technical changes, then it looks ready to go

pnorman (approved):

jgruca:
Thanks, everyone!

ZeLonewolf:
I am seeing the following error, and I'm wondering if it might be related to this PR?

```
kosmtik_1  | Postgis Plugin: ERROR:  Unexpected end of string
kosmtik_1  | LINE 4: ...'waterway_' || waterway, 'natural_' || "natural", 'golf_' ||...
kosmtik_1  |                                                              ^
kosmtik_1  | in executeQuery Full sql was: 'SELECT * FROM (SELECT
kosmtik_1  |   way,
kosmtik_1  |     NULL as way_pixels,
kosmtik_1  |     COALESCE('aerialway_' || aerialway, 'attraction_' || CASE WHEN tags @> 'attraction=>water_slide' THEN 'water_slide' END, 'leisure_' || leisure, 'man_made_' || man_made, 'waterway_' || waterway, 'natural_' || "natural", 'golf_' || tags->'golf') AS feature,
kosmtik_1  |     access,
kosmtik_1  |     name,
kosmtik_1  |     tags->'operator' as operator,
kosmtik_1  |     ref,
kosmtik_1  |     NULL AS way_area,
kosmtik_1  |     CASE WHEN building = 'no' OR building IS NULL THEN 'no' ELSE 'yes' END AS is_building
kosmtik_1  |   FROM planet_osm_line
kosmtik_1  |   WHERE ((man_made IN ('pier', 'breakwater', 'groyne', 'embankment')
kosmtik_1  |       OR (man_made = 'pipeline'
kosmtik_1  |          AND tags-> 'location' IN ('overground', 'overhead', 'surface', 'outdoor')
kosmtik_1  |          OR bridge IN ('yes', 'aqueduct', 'cantilever', 'covered', 'trestle', 'viaduct'))
kosmtik_1  |       OR tags @> 'attraction=>water_slide'
kosmtik_1  |       OR aerialway IN ('cable_car', 'gondola', 'mixed_lift', 'goods', 'chair_lift', 'drag_lift', 't-bar', 'j-bar', 'platter', 'rope_tow', 'zip_line')
kosmtik_1  |       OR leisure IN ('slipway', 'track')
kosmtik_1  |       OR waterway IN ('dam', 'weir')
kosmtik_1  |       OR "natural" IN ('arete', 'cliff', 'ridge'))
kosmtik_1  |     AND name IS NOT NULL)
kosmtik_1  |     OR (tags @> 'golf=>hole' AND ref IS NOT NULL)
kosmtik_1  | ) AS text_line LIMIT 0'
kosmtik_1  |   encountered during parsing of layer 'text-line' in Layer
kosmtik_1  | [httpserver] /openstreetmap-carto/tile/4/5/6.png?t=1631407022029 500
```

ZeLonewolf (commented):
This change appears to be causing SQL errors that I don't fully understand.

pnorman:
Fixed in 871804d13f561c416c4423fc10ca2e7ea4aa413a

DaveF63:
Even though I don't like golf, this is a good addition to the OSM 'standard' map. It's these details which makes it stand out from its competitors.

I concur with the feeling that golf=green (@pitch) is a bit too blue, especially given it's shape & size making it look similar to a pond.  There /are/ differences between it & water features, but they have to be studied intently to notice them. 

Aside: I've never been keen on the current @pitch colour being used on sports pitches.  Should be a bit 'greener' IMO.

boothym:
What about reusing the old golf course colour - rgb(181 227 181) - for golf greens, as that is less blue than the current pitch colour?

PR #4382, opened 2021-04-22
Add rendering for `highway=busway`

kolgza:
- #4226

The rendering for `higway=bus_guideway` would be applied to instances of `highway=busway`.

- https://www.openstreetmap.org/relation/4051402#map=15/40.4294/-80.0636

#### Before
image

### After
image

### DIFF
--- a/style/amenity-points.mss
+++ b/style/amenity-points.mss
@@ -165,10 +165,24 @@
     }
   }
 
-  [feature = 'amenity_bus_station'][zoom >= 16] {
-    marker-file: url('symbols/amenity/bus_station.svg');
-    // use colors from SVG to allow for white background
-    marker-clip: false;
+  [feature = 'amenity_bus_station'] {
+    [zoom >= 16] {
+      marker-file: url('symbols/amenity/bus_station.svg');
+      // use colors from SVG to allow for white background
+      marker-clip: false;
+    }
+    [zoom < 16][zoom >= 12] {
+      marker-file: url('symbols/square.svg');
+      marker-fill: @transportation-icon;
+      marker-clip: false;
+      marker-width: 4;
+      [zoom >= 13] {
+        marker-width: 6; 
+      }
+      [zoom >= 15] {
+        marker-width: 6;
+      }
+    }
   }
 
   [feature = 'amenity_taxi'][zoom >= 17] {


pnorman (changes requested):
I agree with a unified rendering for both `bus_guideway` and `busway`. I'm not certain this is the right rendering for them when combined. On the other hand, I can't think of something that I'm certain would work. I was thinking of blue variations of railway=subway, but don't think that's any better.

My request changes is just about the syntax.

kolgza:
> I was thinking of blue variations of railway=subway, but don't think that's any better.

I think this would have the same problem where it *just sort doesn't look right* where a busway meets a roadway. It should really have more road-like qualities.

imagico (changes requested):
There has not yet been a meaningful discussion in #4226 about the suitability of this tag for rendering in this style.  At this time this is highly questionable.  The tag is used 1500 times (compared to 2900 of service=busway), more than 1200 of these uses are last edited by a single user (colgza) - possibly indicating a mechanical edit.  It is going to take quite some time to see if local mappers mapping based on local knowledge are going to broadly adopt this tagging and how the tag is actually going to be used.

Regarding the design - i would strongly suggest not using blue for non-water related line features in this style - this is just confusing.  I know this is the case already for bus_guideway since the beginning (just like for cycleway) - but this is something best to be fixed rather than extended to other features IMO.

kolgza:
> There has not yet been a meaningful discussion in #4226 about the suitability of this tag for rendering in this style. At this time this is highly questionable. The tag is used 1500 times (compared to 2900 of service=busway), more than 1200 of these uses are last edited by a single user (colgza) - possibly indicating a mechanical edit. It is going to take quite some time to see if local mappers mapping based on local knowledge are going to broadly adopt this tagging and how the tag is actually going to be used.


Most busways are applied with `highway=service` which renders them poorly. However, applying `highway=busway` would cause them *to instead not render at all*. For this reason, very few mappers would apply this tag (and some may even remove it, thereby intensifying the catch-22), even if it is accepted as opposed to `highway=service` + `service=busway`.

If you want Carto to not affect how quickly this tag is adopted, then what you should do is render it the same way as `highway=service` because Carto would no longer be favor one tagging scheme over the other. However, because the proposal for `highway=busway` was accepted, this experiment would be moot.

imagico:
We have had this situation plenty of times - like highway=ford -> ford=yes, amenity=embassy -> office=diplomatic.  If a tag is popular among mappers it will get used independent of if we render it here or not.

Our aim is to support mappers in consistent use of tags (which they have developed on their own), not to proactively steer them to use certain tags.

Recoil016:
Ah yes, the circle of death: New tag => people don't use it because carto doesn't render it => carto doesn't render it because people don't use it => people don't use it because carto doesn't render it...

kolgza:
And, I accidentally closed the pull request because I don't know how to use git :p

zekefarwell:
> Ah yes, the circle of death: New tag => people don't use it because carto doesn't render it => carto doesn't render it because people don't use it => people don't use it because carto doesn't render it...

Yup.  Most people don't care what the tags are.  They care that the thing they mapped shows up on a map.  Carto influences mappers just as much by _not_ rendering a feature as it does by rendering a feature.  Denying this influence is disingenuous.

zekefarwell:
Moral responsibility and physical attack are far beyond anything I intended to imply.  All I'm saying is that in general mappers will choose a tag that renders over a tag that does not.  This dynamic upholds the status quo and hinders adoption of improved tagging schemes.

kolgza:
Please don't have any arguments on my account. If you do, please put it in the mailing lists and not here.

PR #4383, opened 2021-04-23
Make `amenity=bus_station` render more like `railway=halt`

kolgza:
**_This comment is outdated!_**

See https://github.com/gravitystorm/openstreetmap-carto/pull/4383#issuecomment-826095714

---

- Change `style/amenity-points.mss` to render `amenity=bus_station` at z>=12
  - Use `symbols/square.svg` at z<16
    - Rendering is unaffected above z>=16
  - The marker width at each zoom level is the same as for `railway=station`
 

- [Pittsburgh, PA](https://www.openstreetmap.org/#map=12/40.4232/-80.0157)
- [Adelaide, NSW](https://www.openstreetmap.org/?mlat=-34.8707&mlon=138.6603#map=12/-34.8707/138.6603)

### Before
Pittsburgh at z=12:
image
Adelaide at z=12:
image

#### After

Adelaide at z=12:
z12 Adelaide
Adelaide at z=14:
z14 adelaide
Pittsburgh at z=12
z12 Pittsburgh without bus_guideway
Pittsburgh at z=12, except the rendering for `highway=bus_guideway` is applied to `highway=busway`:
z12 Pittsburgh with bus_guideway

### DIFF
--- a/style/amenity-points.mss
+++ b/style/amenity-points.mss
@@ -171,14 +171,11 @@
       // use colors from SVG to allow for white background
       marker-clip: false;
     }
-    [zoom < 16][zoom >= 12] {
+    [zoom < 16][zoom >=13] {
       marker-file: url('symbols/square.svg');
       marker-fill: @transportation-icon;
       marker-clip: false;
       marker-width: 4;
-      [zoom >= 13] {
-        marker-width: 6; 
-      }
       [zoom >= 15] {
         marker-width: 6;
       }
--- a/style/amenity-points.mss
+++ b/style/amenity-points.mss
@@ -165,10 +165,24 @@
     }
   }
 
-  [feature = 'amenity_bus_station'][zoom >= 16] {
-    marker-file: url('symbols/amenity/bus_station.svg');
-    // use colors from SVG to allow for white background
-    marker-clip: false;
+  [feature = 'amenity_bus_station'] {
+    [zoom >= 16] {
+      marker-file: url('symbols/amenity/bus_station.svg');
+      // use colors from SVG to allow for white background
+      marker-clip: false;
+    }
+    [zoom < 16][zoom >= 12] {
+      marker-file: url('symbols/square.svg');
+      marker-fill: @transportation-icon;
+      marker-clip: false;
+      marker-width: 4;
+      [zoom >= 13] {
+        marker-width: 6; 
+      }
+      [zoom >= 15] {
+        marker-width: 6;
+      }
+    }
   }
 
   [feature = 'amenity_taxi'][zoom >= 17] {


imagico:
This does not seem to make much sense to me - an `amenity=bus_station` is - both according to actual use and documentation on the wiki - something semantically very different from `railway=station`.  And your example seems to show what would normally be tagged a bus stop rather than a bus station.

Independent of that bus stations are currently rendered at z16+ - i don't think moving this down four (!!) zoom levels is a good idea.  I also don't see what purpose this would serve in a typical case at this zoom level.

kolgza:
Thank you for the reply. I should have included my rationale and the history in the comment.

I started working on the proposal for `highway=busway` because of #4226. I wanted BRT networks to render more-similarly to light-rail networks.

Initially, I intended to additionally propose `station=busway` — which would be to `highway=busway` what `station=light_rail` is to `railway=light_rail`. After all, the difference between a light-rail station and a BRT station is the types of vehicles used, and [most transit maps that I've seen show "BRT stations" and "light-rail stations" as the same](https://github.com/gravitystorm/openstreetmap-carto/issues/4226#issue-724131173). However, I had difficulty trying to create a definition that differentiates `station=busway` from `highway=bus_stop` and `amenity=bus_station`. 

---

Anyway, `amenity=bus_station` is generally applied to bus stations along a BRT corridor, even if they don't meet the specific [definition in the wiki page](https://wiki.openstreetmap.org/wiki/Tag:amenity%3Dbus_station).

An example of this occurring would be in Bogotá, Columbia:


https://overpass-turbo.eu/s/16wt
image
image

kolgza:
TLDR: `amenity=bus_station` is applied to bus stations along BRT corridors like `station=light_rail` is applied to light-rail stations along light-rail corridors.

kolgza:
> I also don't see what purpose this would serve in a typical case at this zoom level.

I suppose it would serve the same purpose that`railway=station` serves at this zoom level.

imagico:
Well - at the moment amenity=bus_station is predominantly used for what are bus stations according to what is documented on the wiki and we render it accordingly.  What the appropriate tag would be to tag BRT bus stops/stations and if it makes sense to overload amenity=bus_station for that purpose is a topic for a different venue.

> I suppose it would serve the same purpose that railway=station serves at this zoom level.

Rendering of railway=station at z12 serves to display further structure and details on the rail network (which we start displaying at z8 for the major rails).  They serve this purpose well without adding too much noise because typical distances between train stations are from a few km to more than 50km and therefore sparse enough to not overfill the map and since the choice of color is with only subtle contrast with the color of the railway line.  Note we have railway=halt for simple train stops en route without more structure typical for a railway station.  Use of that varies between countries and in your two samples the majority of train stations shown would elsewhere probably be tagged as railway=halt.

kolgza:
Well, this is why I like `public_transport=station` + `bus=yes`. 

- It serves the general purpose as what `station=busway` would have served.
- It is already accepted.
- [It has a own definition](https://wiki.openstreetmap.org/wiki/Tag:public_transport%3Dstation#Definition) that does not automatically imply something meets the definition of `amenity=bus_station`.

It would probably be more appropriate if this tagging scheme were applied to the examples I provided, and `amenity=bus_station` were removed.

The only problem is that Carto wouldn't render it. There are [less than 24k uses of this combination of tags](https://taginfo.openstreetmap.org/tags/public_transport=station#combinations), while there are [over 60k uses of `amenity=bus_station`](https://taginfo.openstreetmap.org/tags/amenity=bus_station). Therefor, if Carto were to use this tagging scheme for rendering, it would be unduly influencing mappers to use a use it.

But, I digress. Such a discussion probably belongs in the tagging mailing list.

kolgza:
> Note we have railway=halt for simple train stops en route without more structure typical for a railway station. Use of that varies between countries and in your two samples the majority of train stations shown would elsewhere probably be tagged as railway=halt.

I suppose I should change this PR to instead render `amenity=bus_station` more according to the rendering of `railway=halt`.

kolgza:
Screenshot_2021-04-23 OpenStreetMap Carto — Kosmtik(1)
Screenshot_2021-04-23 OpenStreetMap Carto — Kosmtik

kolgza:
BRT stations are usually supposed to be between 300 meters and 800 meters apart. I'm not completely sure if the distance between railway halts is about the same.

kolgza:
Pittsburgh, PA at z=13
Screenshot_2021-04-23 OpenStreetMap Carto — Kosmtik(2)

Los Angeles, CA at z=13
image

kolgza:
- Change `style/amenity-points.mss` to render `amenity=bus_station` at z>=13
  - Use `symbols/square.svg` at z<16
    - Rendering is unaffected where z>=16
  - The marker width at each zoom level is the same as for `railway=station`

- [Pittsburgh, PA](https://www.openstreetmap.org/#map=12/40.4232/-80.0157)
- [Los Angeles, CA](https://www.openstreetmap.org/?mlat=34.1908&mlon=-118.5954#map=13/34.1908/-118.5954)

#### Before

image
image

#### After

Pittsburgh, PA at z=13
Screenshot_2021-04-23 OpenStreetMap Carto — Kosmtik(2)

Los Angeles, CA at z=13
image

kolgza:
Some might be interested in how this PR would affect cities in Europe where `amenity=bus_station` is used much more sparingly. Here is a demonstration of this rendering in Paris, France.

---
[Link to area in question](https://www.openstreetmap.org/#map=13/48.8978/2.2814)


#### Before:
image


#### Overpass Turbo query for `amenity=bus_station`:
image

#### At z=13
Paris 2

#### At z=15
Paris 3

#### At z=16 (unchanged)
Paris 4

kolgza:
Closing this PR so I can reopen it under a different branch.

Issue #4260, opened 2020-12-02
Render amenity=animal_shelter


Ivans-Git:
The symbol could be the same as the dog park, in brown

imagico:
3800 uses

https://wiki.openstreetmap.org/wiki/Tag:amenity%3Danimal_shelter

jeisenbe:
Usage has been increasing steadily since hitting 1000 in 2016:
https://taginfo.openstreetmap.org/tags/?key=amenity&value=animal_shelter#chronology

<img width="941" alt="Screen Shot 2020-12-04 at 16 10 52" src="https://user-images.githubusercontent.com/42757252/101227018-4a2f6800-364b-11eb-8c9b-7b752b9e1318.png">

Usage is widespread:
<img width="683" alt="Screen Shot 2020-12-04 at 16 14 06" src="https://user-images.githubusercontent.com/42757252/101227144-be6a0b80-364b-11eb-98ab-e90eb1811f4a.png">

This was approved back in 2012, along with animal_boarding (see request to add that feature in #3966) and animal_breeding.

Ivans-Git:
> 3800 uses
> 
> https://wiki.openstreetmap.org/wiki/Tag:amenity%3Danimal_shelter

And? In many cases, like this one, the uses do not reveal anything, there is simply not this type of amenity in all cities, but it deserves rendering as it is a unique installation...

imagico:
Sorry for the brevity - this was just providing data for future reference.

jeisenbe:
@Ivans-Git while there is no minimum number required for rendering, we are more likely to add a tag which has been used thousands of times, instead of a tag used hundreds of times, since the more frequently-used tags are likely to be more significant to mappers and map users. 

I also posted the charts showing geographical distribution of use and how the usage has increased over time, because we want to add features which are used widely around the globe and which are increasing steadily in usage (rather than having been added all at once in one or two organized activities such as an import) - both of these are signs that a tag has obtained consensus as the common way to tag a particular feature.

polarbearing:
As for the rendering style, yes the dog paw would be suitable, but there should be an element for the shelter/boarding. I could imagine a roof symbol like ^ above the paw. Breeding would be different however.

Khomille:
Something like this?

![Animal_shelter_2](https://user-images.githubusercontent.com/23643419/106957860-d556d800-6738-11eb-8d17-4eebdb6a6f3b.png)

[...]

Issue #3966, opened 2019-11-05
Add icon for amenity=animal_boarding


Adamant36:
It would be cool if there was an icon for animal boarding facilities. As it is their name's are already rendered in a generic text label color. So it wouldn't be that much of a leap to also add an icon. My suggestion is going with paw icon of veterinary, but without the cross in the middle. Since there is some overlap there and it makes sense.

HolgerJeromin:
> As it is their name's are already rendered in a generic text label color.

There is no such thing since a many years.
Can you add a screenshot and a link to a location?

Adamant36:
@HolgerJeromin, yes, it is when the amenity is combined with a building, but ultimately that really shouldn't matter. If there is an amenity attached to the building the name of the amenity shouldn't render the same as a building name. Since they are separate objects IMHO. 

> We have 720 nodes and 1195 ways with this "pet hotel" feature.

That seems close enough to the threshold to me. There's actually about a thousand more uses of amenity=animal_shelter. Which I find interesting.

jeisenbe:
The tag [amenity=animal_boarding](https://wiki.openstreetmap.org/wiki/Tag:amenity%3Danimal_boarding) is slightly less common than the two other animal-facility tags which were approved around the same time: [amenity=animal_shelter](https://wiki.openstreetmap.org/wiki/Tag:amenity%3Danimal_shelter) and [amenity=animal_breeding](https://wiki.openstreetmap.org/wiki/Tag:amenity%3Danimal_breeding)

### Numbers:
animal_boarding - [1933 uses (taginfo)](https://taginfo.openstreetmap.org/tags/?key=amenity&value=animal_boarding#overview)
animal_breeding - [2 841 uses ](https://taginfo.openstreetmap.org/tags/?key=amenity&value=animal_breeding#overview)
animal_shelter - [3002 uses[(https://taginfo.openstreetmap.org/tags/?key=amenity&value=animal_shelter#overview)

### Maps: 
Animal shelter is more common in North America and Australia, animal breeding is more limited in extent. Here in Indonesia animal_boarding is the only one of the 3 that is common. 

animal_boarding:
![animal_boarding-map](https://user-images.githubusercontent.com/42757252/68363439-3080d000-016e-11ea-82ec-b7fbda24c7fa.png)

animal_breeding:
![animal_breeding-map](https://user-images.githubusercontent.com/42757252/68363462-3c6c9200-016e-11ea-9902-9053ae457192.png)

[...]

PR #4384, opened 2021-04-23
Update dog park rendering

### LINKED ISSUE #3041: Dog park is too prominent

ZeLonewolf:
This PR remove the `leisure=dog_park` pattern fill and unifies the rendering style with playgrounds at high zoom levels.

Fixes #3041 (id of the issue to be closed)

Berlin: https://www.openstreetmap.org/#map=16/52.4769/13.2607
New York City: https://www.openstreetmap.org/#map=18/40.74897/-74.00707

Berlin, zoom 13
dog_park_de_13_before
dog_park_de_13_after

Berlin, zoom 16
dog_park_de_16_before
dog_park_de_16_after

New York City, zoom 17
dog_park_ny_17_before
dog_park_ny_17_after

New York City, zoom 18
dog_park_ny_18_before
dog_park_ny_18_after


style/amenity-points.mss
```diff
@@ -1226,14 +1226,14 @@
...
-
+  [feature = 'leisure_dog_park'][zoom >= 17],
   [feature = 'leisure_playground'][zoom >= 17] {
-    marker-file: url('symbols/leisure/playground.svg');
+    [feature = 'leisure_playground'] {
+      marker-file: url('symbols/leisure/playground.svg');
+    }
+    [feature = 'leisure_dog_park'] {
+      marker-file: url('symbols/shop/pet.svg');
+    }
```
ZeLonewolf, line 1236: This change unifies the `dog_park` and `playground` styling.

style/amenity-points.mss
```diff
@@ -2307,11 +2308,6 @@
           text-halo-radius: 0;
         }
       }
-      [feature = 'leisure_dog_park'] {
-        text-fill: @leisure-green;
-        text-halo-radius: @standard-halo-radius * 1.5; /* Extra halo needed to stand out from paw pattern. */
-        text-halo-fill: @standard-halo-fill;
-      }
```
ZeLonewolf, line 2314: This removes pattern fill styling and replaces it with the simple fill used by park/village green/recreation_ground

style/landcover.mss
```diff
@@ -88,6 +88,7 @@
 
   [feature = 'leisure_recreation_ground'][zoom >= 10],
   [feature = 'landuse_recreation_ground'][zoom >= 10],
+  [feature = 'leisure_dog_park'][zoom >= 13],
```
ZeLonewolf, line 91: Adds casing to dog_parks to match playgrounds.

style/landcover.mss
```diff
@@ -261,20 +262,6 @@
...
-      [way_pixels >= 64] { polygon-gamma: 0.3;  }
-    }
-    [zoom >= 16] {
-      polygon-pattern-file: url('symbols/dog_park.png');
-      polygon-pattern-alignment: global;
-      [way_pixels >= 4]  { polygon-pattern-gamma: 0.75; }
-      [way_pixels >= 64] { polygon-pattern-gamma: 0.3;  }
-    }
-  }
-
```
ZeLonewolf, line 277: Removes pattern fill from dog parks.   Polygon fill is already handled elsewhere.

ZeLonewolf (commented):
Adding comments only to assist reviewers in understanding the changes.

ZeLonewolf:
Additional before/after example from Bristol, Connecticut, USA (location: https://www.openstreetmap.org/#map=14/41.6877/-72.9577)

Screen shots demonstrating no change in label rendering behavior at lower zoom (zoom 14 in this example):

bristol_14_before
bristol_14_after

polarbearing:
Thanks, that is a significant improvement.

imagico:
You are probably not consciously aware of this but this is kind of the last step in a sequence of highly questionable design choices - most of which were made without consensus among the maintainers and which according to what this PR suggested would lead to rendering dog parks identical to playgrounds - an idea that seems ridiculous if you subscribe to the paradigm that differences in color should reflect differences in function and purpose for the target map user and which by the way massively supports the stereotype that OSM is a project of male tech nerds without kids and who view caring for a kid and caring for a dog as similar in nature as well as the view that this project is dominated by the narrow perspective of an urban European/North American temperate climate clientele with their specific views of animal keeping (leisure=dog_park is still the only feature specific to animal keeping we have in this style so far and we render it in a green color implying vegetation as if globally dogs are always kept/walked in vegetated areas).

The history starts essentially with adding dog_park rendering in #2216 (sample rendering in #341), criticized in #2250.  Playground rendering was changed from the traditional well recognizable cyan tone to a meaningless non-distinctive green in #2249 to which various other features were piggy-bagged later (#2891 - note that leisure=fitness_station and leisure=fitness_centre are unified with playgrounds, not with pitch or track - or stadium, sports_centre, which were later unified in #2954) and recreation_ground (#2964 which was unified with playground and not with park to which is is practically a synonym) including dog_park with its base color in #2268 (while keeping a distinction via pattern).  I don't want to repeat all the considerations why these changes were in a disadvantageous direction from these issues - see also #2270.

The dog_park rendering is also still a lingering problem in the ac-style by the way, i have not yet gotten around addressing that.  My thought however goes into the direction of looking towards the farmyard/farmland colors to establish a distinct color for animal keeping features (see also #4260).

This is all of course not your fault - in the current situation i understand this change might make some sense.  But the better approach would IMO be to take a step back and look at the larger picture of landcover fill colors in this style and how to get back to a systematic strategy in that domain.

1ec5:
> an idea that seems ridiculous if you subscribe to the paradigm that differences in color should reflect differences in function and purpose for the target map user and which by the way massively supports the stereotype that OSM is a project of male tech nerds without kids and who view caring for a kid and caring for a dog as similar in nature as well as the view that this project is dominated by the narrow perspective of an urban European/North American temperate climate clientele with their specific views of animal keeping

In fact, in the United States, dog parks are very much designed to function as playgrounds for pets, and their popularity as such is by no means limited to males, tech nerds, or dense urban environments. But I suppose this is the americentrism you refer to, given your observation in https://github.com/gravitystorm/openstreetmap-carto/issues/2270#issuecomment-260938327 that the two features couldn’t be more dissimilar. What are dog parks like in other places that would be difficult to reconcile with this rendering?

> leisure=dog_park is still the only feature specific to animal keeping we have in this style so far

#3966 and #4260 propose more animal-related features, but to me it seems that dog parks are much more closely associated with playgrounds than animal boarding, animal shelters, or veterinarians (which are rendered). Do you have other not-yet-rendered features in mind that could be part of a whole coherent category of animal-related features? Or is your core point that dog parks simply should not be rendered in the first place?

> we render it in a green color implying vegetation as if globally dogs are always kept/walked in vegetated areas

I wouldn’t think the fill or label color could be taken literally anyways. After all, playgrounds in the U.S. aren’t usually associated with greenspace: mulched or rubberized surfaces are the norm, and trees and landscaping are a luxury not often found in simple neighborhood or school playgrounds. Yet green is a sensible color for playgrounds, since they are so commonly park amenities.

polarbearing:
The green is used for various leisure purposes. I see no problem with having the same colour at low zoom, and we have clearly distinctive icons when looking close. 
Anyway, another idea I had was an outline instead of a fill.

imagico:
I am not really keen on discussing the culture and role of animal keeping in different parts of the world and its similarity to child-rearing here - my comments in that direction were meant to encourage developers to reflect on their selective perception of the field.

As said what is missing here (seriously and desperately missing IMO since 2016) is an overall color design strategy with consensus support.  That is what #2270 is for.  And the way to get there is not by making local convenience choices for selecting individual colors by finding a free spot in color space or overloading a suitably undefined existing color with an additional meaning and then ex post facto rationalizing that to be reconcilable with reality.

> Do you have in mind other not-yet-rendered features that could be part of a whole coherent category of animal-related features? Or is your core point that dog parks simply should not be rendered in the first place?

I have no issue with rendering leisure=dog_park provided that a suitable rendering can be found (and so far i do not consider this to be the case for any of the variants tried).  We can only render what mappers actually record in a consistent fashion.  That dog parks are the only animal-keeping related feature we render so far is not exclusively caused by the perception and experience bias of the developers here, it is at least as much due to an equally narrow perspective among most mappers.  The lack of  clear, well defined, documented and consistently used tagging schemes for practically the whole domain of livestock keeping is testimony to that.

imagico:
For clarification - due to https://github.com/gravitystorm/openstreetmap-carto/issues/2270#issuecomment-826347185 - the above comments are not a negative review, they are meant to provide useful historic context to the change this PR proposes.

I will however not positively review or merge this change unless someone can convince me of its merits.

imagico:
This change requires review by someone other than me.

ZeLonewolf:
Closing so other contributors can work on this.

Issue #3581, opened 2018-12-20
highway=bus_guideway layering, tunnels, covered, bridges


sladen:
Per request @kocio-pl in gravitystorm/openstreetmap-carto/issues/2965#issuecomment-448795111:

for `highway=bus_guideway`, the following are not rendered:

* `layer=0` (ie. under `bridge=yes`; `layer=1`) https://www.openstreetmap.org/way/54965900#map=17/-34.89753/138.61934
* `layer=-1` (ie. under `bridge=yes`) https://www.openstreetmap.org/way/55053642#map=18/-34.89048/138.63145
* `tunnel=yes*` https://www.openstreetmap.org/way/545944857#map=16/-34.92069/138.61394
* `bridge=yes` https://www.openstreetmap.org/way/420096851#map=17/51.888/-0.46729
* `bridge=viaduct` https://www.openstreetmap.org/way/92924790#map=18/52.31489/-0.05584
* `covered=yes` https://www.openstreetmap.org/way/656244701#map=19/33.92338/-118.34427

Ideally these would be rendered as per a railway.  There were some historical locations in Essen and Mannheim where `highway=bus_guideway` and `railway=tram` would have shared the same physical space, but these appear to disappeared about when OSM came into existance.

sladen:
Essen example:

* `highway=bus_guideway`; `railway=tram` https://www.openstreetmap.org/way/81149695#map=19/51.45231/7.02779

As an aside, but for completeness… This was the former Mannheim example, it would appear to have been paved over as a normal tram/bus-only service road:

* https://www.openstreetmap.org/way/204347211#map=19/49.48900/8.51716

Ideally the `tram` rendering should be on top of the `bus_guideway`.

imagico:
Going to keep this open as the aggregate issue for all the `highway=bus_guideway` layering issue and closing #2489 and #2897 instead.

The solution for this is IMO to include `highway=bus_guideway` in the roads layers.  It seems that would depend on a database reload because it would require defining a z_order for `highway=bus_guideway` - see 

https://github.com/gravitystorm/openstreetmap-carto/blob/a2077c0c8d40fb7f6d308b4ca4e8941d4c8b699a/openstreetmap-carto.lua#L188-L214

Issue #3396, opened 2018-09-15
Make bus guideways less prominent and less water-like


polarbearing:
Bus guideways were touched a while ago in #2965 and #2966. They still stand out unnaturally on the map and look like "water railways". 

I propose to both
- make the colour darker, e.g. a greyish blue 
- change the dasharray to have wider white and smaller colour elements. 

An alternative proposal, I hear in the hackweekend room, would be to render like a minor serviceway marked private. 

Below is the current rendering ([Essen, Germany](https://www.openstreetmap.org/#map=18/51.45545/7.04322)) across some railways. Cambridge examples were posted in #2965.
![bus_guideway](https://user-images.githubusercontent.com/8266355/45588359-f3c39b00-b913-11e8-97da-6f32d6e09839.png)

PR #4385, opened 2021-04-23
Add rendering for `highway=busway`

kolgza:
This is basically the same pull request as #4382, except from a different branch. In other words, I don't know how to use git. @imagico: please request the same changes as before.

---

- #4226

The rendering for `higway=bus_guideway` would be applied to instances of `highway=busway`.

- https://www.openstreetmap.org/relation/4051402#map=15/40.4294/-80.0636

#### Before
image

### After
image

### DIFF
--- a/project.mml
+++ b/project.mml
@@ -1005,7 +1005,7 @@ Layer:
         (SELECT
             way
           FROM planet_osm_line
-          WHERE highway = 'bus_guideway'
+          WHERE highway IN ('bus_guideway', 'busway')
         ) AS guideways
     properties:
       minzoom: 11
--- a/project.mml
+++ b/project.mml
@@ -1005,7 +1005,7 @@ Layer:
         (SELECT
             way
           FROM planet_osm_line
-          WHERE highway IN ('bus_guideway', 'busway')
+          WHERE highway = 'bus_guideway'
         ) AS guideways
     properties:
       minzoom: 11


imagico (changes requested):
See comments on #4382.

I am not opposed to the idea of rendering bus only roads built and used for specific bus routes in a distinct fashion.  But i don't think the tag highway=busway has shown clear enough adoption and consistency in use by mappers to support rendering it here and i don't think the suggested rendering (in particular the blue color) is a good choice for that - in particular also where these roads connect to the general road network.  That is different for bus_guideway - which is more rail like and does not transit to regular roads.

kolgza:
> i don't think the suggested rendering (in particular the blue color) is a good choice for that

This relates to  #3396

kolgza:
> That is different for bus_guideway - which […] does not transition to regular roads.

Here is a counterexample:
image
image

This issue is inherent to the fact that `highway=bus_guideway` is rendered like a railway, but must somehow merge into roads.
@pnorman discussed both redoing the rendering for bus guideways, and unifying it with `highway=busway`.

kolgza:
@imagico Would you merge this PR if it instead applied the rendering of `highway=service` to `highway=busway`?

imagico:
Also keep in mind #3581 - the rendering of `highway=bus_guideway` currently suffers from various issues related to layering relative to other roads features.  It would not be advisable to extend this highly suboptimal rendering to additional features without addressing those issues first.

kolgza:
Closing, because I lack the expertise to implement this.

Issue #2346, opened 2016-09-14
rendering of waterway=ditch/drain incentivizes bad mapping


imagico:
Currently this style renders waterway=ditch/drain thinner than waterway=stream at z15+ (2 vs. 3 pixels) but otherwise identical leading to widespread abuse for tagging small streams - there are literally thousands of cases of this in mountainous parts of southern Germany and Switzerland.

I have no solution for this at the moment, rendering them the same would reduce the incentive but ideally rendering should clearly indicate the difference between natural and artificial waterways.

Related to #1101

ray66:
The width and zoom level visibility should reflect the appearance of natural and artificial waterways in the real world. According the [waterway wiki page](http://wiki.openstreetmap.org/wiki/Key:waterway)  the major difference between streams and ditches is that the latter are artificial. The pictures illustrating both values are very similar. According to the same wiki page, the width of a ditch is similar to that of a small river. Therefore, my proposal is

- render ditches with the same width as streams 
- render ditches and streams less thick than currently and only from zoom level 15 or 16 on
- render drains thicker than drains, slightly less or equal to rivers

kocio-pl:
> render drains thicker than drains

What did you mean here?

imagico:
> The width and zoom level visibility should reflect the appearance of natural and artificial waterways in the real world.

No, a map is always a more or less abstract depiction.  Nearly all maps that display rivers and other waterways display the important ones earlier and more prominently than they would appear in reality.  The ability of this style to do the same is primarily limited by the difficulty to determine a reasonable and consistent measure of importance in real time from local properties.

Practically `waterway=ditch` and `waterway=drain` are used almost interchangeably, in particular since ditch is defined through the purpose of drainage.  There is not really a basis in the data for rendering these two in different ways.

But it would make sense to make a clear difference between artificial and natural waterways in rendering, i am just not sure what would be a good way to do this.

ray66:
@kocio-pl : Sorry, it should read "drains thicker than **ditches**". I corrected my original post.

jeisenbe:
FYI, @imagico wrote a blog post showing a possible option for fixing this issue:

http://blog.imagico.de/water-under-the-bridge/

Complete code available here: https://github.com/imagico/osm-carto-alternative-colors/blob/master/water.mss

```
[waterway = 'ditch'],
      [waterway = 'drain'] {
        [zoom >= 15][intermittent != 'yes'][seasonal != 'yes'] {
[...]

PR #4388, opened 2021-04-25
water: render waterway=fish_pass

### LINKED ISSUE #2895: waterway=fish_pass
Fish ladders are currently not rendered. I recommend to draw them identically as waterway=stream.

TheJJ:
Fixes #2895

- Render `waterway=fish_pass`

I don't have tested this at all and I still need to figure out how to produce a test rendering :)

### DIFF
--- a/style/water.mss
+++ b/style/water.mss
@@ -52,6 +52,7 @@
 
 #water-lines::casing {
   [waterway = 'stream'],
+  [waterway = 'fish_pass'],
   [waterway = 'ditch'],
   [waterway = 'drain'] {
     [int_tunnel = 'no'] {
@@ -144,6 +145,7 @@
   }
 
   [waterway = 'stream'],
+  [waterway = 'fish_pass'],
   [waterway = 'ditch'],
   [waterway = 'drain'] {
     [int_intermittent != 'yes'][zoom >= 14],
@@ -174,23 +176,27 @@
         water/line-clip: false;
       }
 
-      [waterway = 'stream'][zoom >= 15] {
+      [waterway = 'stream'][zoom >= 15],
+      [waterway = 'fish_pass'][zoom >= 15] {
         water/line-width: 3;
 
         [int_tunnel = 'yes'] {
           background/line-width: 3;
         }
       }
+
       [int_tunnel = 'yes'][zoom >= 15] {
         background/line-width: 3.5;
         water/line-width: 3.5;
-        [waterway = 'stream'] {
+        [waterway = 'stream'],
+        [waterway = 'fish_pass'] {
           background/line-width: 4.5;
           water/line-width: 4.5;
         }
         water/line-dasharray: 4,2;
         tunnelfill/line-width: 1;
         [waterway = 'stream'] { tunnelfill/line-width: 2; }
+        [waterway = 'fish_pass'] { tunnelfill/line-width: 2; }
         tunnelfill/line-color: #f3f7f7;
       }
     }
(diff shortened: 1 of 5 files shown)


HolgerJeromin:
At least 
https://github.com/gravitystorm/openstreetmap-carto/blob/42a6baf41a6e8f7020ff005a18e2f63e75a97d1b/project.mml#L180
and line 220 (and others) seems to be missing.

imagico:
Thanks for the PR.  We generally require PRs to be tested and examples of before and after rendering to be shown to allow a quick assessment of what the change tries to achieve.  This does not necessarily have to be done by the person submitting the PR of course.  When you implement a change with visual impact on the rendering it is generally highly advisable to visually check the results.

TheJJ:
Definitely :smile: I just wanted to get the ball rolling.
What would be the simplest possible way to test the changes? Apart from of course somebody with the setup in place creating the renders.

style/water.mss
```diff
@@ -181,16 +183,30 @@
           background/line-width: 3;
         }
       }
+
+      [waterway = 'fish_pass'][zoom >= 15] {
```
pitdicker, line 187: Can you merge this section with the one above for `[waterway = 'stream'][zoom >= 15]`?

pitdicker, line 187: Can you merge this section with the one above for `[waterway = 'stream'][zoom >= 15]`?

style/water.mss
```diff
@@ -181,16 +183,30 @@
...
+      }
+
       [int_tunnel = 'yes'][zoom >= 15] {
         background/line-width: 3.5;
         water/line-width: 3.5;
         [waterway = 'stream'] {
           background/line-width: 4.5;
           water/line-width: 4.5;
         }
+        [waterway = 'fish_pass'] {
```
pitdicker, line 202: Can you merge this section with the one above for `[waterway = 'stream']`

pitdicker (commented):
Looks good to me.

Before (https://www.openstreetmap.org/#map=18/52.49547/6.58054):
before

After:
after

TheJJ:
bump bamp, what is remaining to bring this upstream?

jeisenbe:
It looks like @pitdicker approved this in a prior review. Does anyone else have time to test this?

imagico (changes requested):
This seems to work as intended.

As long as we render waterway=stream and waterway=ditch/drain differently (see #2346) i think it would be more logical to have waterway=fish_pass match the rendering of waterway=ditch/drain and not waterway=stream.

Apart from that i think this is something we can add, i am just somewhat wondering if it would be better to render waterway=fish_pass like waterway bridges rather than like a plain waterway.  This would make fish passes in the middle of a river across a weir for example visible.  Like:

https://www.openstreetmap.org/way/714071249
https://www.openstreetmap.org/way/466865795
https://www.openstreetmap.org/way/993880729
https://www.openstreetmap.org/way/285989535

I am not sure about this given the different styles in mapping that exist (see https://github.com/gravitystorm/openstreetmap-carto/issues/2895#issuecomment-1147438074) - so just a thought, not a request for changes.

TheJJ:
rebased. somebody with a working setup please test :)

imagico:
Setting this as draft.  It requires

* changing rendering of `waterway=fish_pass` to match that of `waterway=ditch`/`waterway=drain` rather than `waterway=stream` as it does at the moment (or alternatively: fixing #2346).
* practical testing and samples.
* optionally: evaluation if this would work better when rendered like a wayterway bridge rather than a normal waterway so fish passes within a riverbank polygon (for example across a weir) are visible.

PR #4391, opened 2021-04-27
Render bus stations more prominantly

### LINKED ISSUE #4390: Render bus stations at lower zoom levels.
The tag `amenity=bus_station` is used in two primary ways:

1. for busway stations along a BRT corridors
2. for transportation hubs in which  multiple bus routes converge [(the original intended use of this tag)](https://wiki.openstreetmap.org/wiki/Tag:amenity%3Dbus_station)

Both uses of this tag suggests that `amenity=bus_station` describes important pieces of transportation infrastructure wherever it is used.
### Actual behavior

`amenity=bus_station` is rendered at z=15, the same zoom level in which `highway=bus_stop is rendered`.

### Expected behavior

`amenity=bus_station` should be rendered identically to how `railway=halt` is rendered.

kolgza:
Fixes #4390

This PR attempts to increase rendering for `amenity=bus_station` without making the map too busy. Between zoom levels 13 and 16, `amenity=bus_station` will render as small blue squares, the width of each icon at each zoom level being identical to `railway=halt`.

This PR does not affect how bus stations render at zoom levels z>=16. Personally, however, I think it is worth noting how seamless the transition between the generic blue square and the bus station pictograph is when zooming in from z=15 to z=16.

- Change `style/amenity-points.mss` to render `amenity=bus_station` at z>=13
  - Use `symbols/square.svg` at z<16
    - Rendering is unaffected where z>=16
  - The marker width at each zoom level is the same as for `railway=halt`

- [Los Angeles, CA](https://www.openstreetmap.org/?mlat=34.1908&mlon=-118.5954#map=13/34.1908/-118.5954)
- [Pittsburgh, PA](https://www.openstreetmap.org/#map=12/40.4232/-80.0157)
- [Paris, France](https://www.openstreetmap.org/?mlat=48.8977&mlon=2.2813#map=13/48.8977/2.2813)

###### Before
image
###### After
image

###### Before
image
###### After
Screenshot_2021-04-23 OpenStreetMap Carto — Kosmtik(2)

###### Before:
image

###### Overpass Turbo query for `amenity=bus_station`:
image

###### At z=13
Paris 2

###### At z=15
Paris 3

###### At z=16 (unchanged)
Paris 4

### DIFF
--- a/style/amenity-points.mss
+++ b/style/amenity-points.mss
@@ -171,14 +171,11 @@
       // use colors from SVG to allow for white background
       marker-clip: false;
     }
-    [zoom < 16][zoom >= 12] {
+    [zoom < 16][zoom >=13] {
       marker-file: url('symbols/square.svg');
       marker-fill: @transportation-icon;
       marker-clip: false;
       marker-width: 4;
-      [zoom >= 13] {
-        marker-width: 6; 
-      }
       [zoom >= 15] {
         marker-width: 6;
       }
--- a/style/amenity-points.mss
+++ b/style/amenity-points.mss
@@ -165,10 +165,24 @@
     }
   }
 
-  [feature = 'amenity_bus_station'][zoom >= 16] {
-    marker-file: url('symbols/amenity/bus_station.svg');
-    // use colors from SVG to allow for white background
-    marker-clip: false;
+  [feature = 'amenity_bus_station'] {
+    [zoom >= 16] {
+      marker-file: url('symbols/amenity/bus_station.svg');
+      // use colors from SVG to allow for white background
+      marker-clip: false;
+    }
+    [zoom < 16][zoom >= 12] {
+      marker-file: url('symbols/square.svg');
+      marker-fill: @transportation-icon;
+      marker-clip: false;
+      marker-width: 4;
+      [zoom >= 13] {
+        marker-width: 6; 
+      }
+      [zoom >= 15] {
+        marker-width: 6;
+      }
+    }
   }
 
   [feature = 'amenity_taxi'][zoom >= 17] {


imagico:
I am fairly critical of this for a number of reasons:

* it explicitly seems to be designed to cater a non-consensus overloading of the tag amenity=bus_station to map bus stops considered 'high importance' (for some essentially subjective or local context dependent definition of importance).
* reusing the bus stop symbol for bus stations at z13-z15 further supports mappers in the wrong assumption that bus stations are just high importance bus stops (which is not how the tag is predominantly used).
* it moves down the starting zoom level from z16 to z13 - three levels, which is a massive change in map design without discussing the effects this has on the map in general.  The balance of what we show and what we don't show is fairly delicate especially at the mid zoom levels so just starting to add some POI type three zoom levels down is going to significantly mess with that balance.
* we typically start rendering labels for POI symbols one or two zoom levels after we start rendering the symbol.  Your change would mean the symbol starts at z13 while the label starts at z17 which would make it highly unique in this style.

What would go a long way towards convincing me of the merits of this change is if you can show me how for normal use of the tag (as documented on the wiki) this change would have a positive effect on the map - in particular at relatively low latitudes where the map is dense.

jeisenbe:
`amenity=bus_station` is more similar to a light rail station, subway station or tram stop than a railway station.  In contrast, `station=subway` and `tram=stop` are not rendered till z14. I would suggest testing `amenity=bus_station` with the smaller icon at z14, and the text label at z15.

@Imagico the bus stop and train station icons are currently both blue squares (buses are light blue, the transport color, while railway stations are darker), and in your alt-colors branch the stations are still the same color though buses have been changed to purple along with other transportation icons, matching with airports. Do you have other ideas about an icon change or color change that would help here?

imagico:
Well - the train station / train halt rendering is cartographically very different from the bus stop rendering because the symbol is located on the railway track which is always shown as context (OpenTopoMap here for example goes a step further and orients the symbol in direction of the railway track) while for the bus stop the symbol is more or less hanging in the air without a clear context - and it is not uncommon that at z16 it is not quite clear what road a bus stop is on (like [here](https://www.openstreetmap.org/#map=16/48.0110/7.7234)).

The current strong color combined with the fairly heavy weight of the bus stop squares, the lack of context and the fact that bus stops tend to be mapped separately for both sides of the road so they tend to show up in pairs - this all leads to a less than optimal rendering result.  And tagging of public transport has evolved quite a bit (though there is no full agreement among mappers on this of course).  So i think it would be valuable if someone would look into if some better interpretation of mapping is possible within the constraints we have here, in particular for the early zoom levels (~z15-z17).  This requires finding out which mapping and tagging techniques have broad acceptance and consistent use, what kind of contextualized styling could work at lower zoom levels and how that could be generated without too much computational costs and code complexity.  I know that is a tough call - but IMO the direction in which we have the most room to improve here.

style/amenity-points.mss
```diff
@@ -165,10 +165,21 @@
...
-    marker-file: url('symbols/amenity/bus_station.svg');
-    // use colors from SVG to allow for white background
-    marker-clip: false;
+  [feature = 'amenity_bus_station'] {
+    [zoom >= 16] {
+      marker-file: url('symbols/amenity/bus_station.svg');
+      // use colors from SVG to allow for white background
+      marker-clip: false;
+    }
+    [zoom < 16][zoom >=13] {
```
jeisenbe, line 174: Usually we would have this part above, using [feature = 'amenity_bus_station'][zoom >= 13], then [zoom >= 16] below

jeisenbe (changes requested):
I would suggest testing amenity=bus_station with the smaller icon at z14, and the text label at z15.

Issue #4235, opened 2020-10-23
Show name for some healthcare POIs


Mannivu:
`healthcare=physiotherapist` and `healthcare=laboratory` are shown on carto using the `addr:housenumber` value (see for example https://www.openstreetmap.org/node/5511376932 and https://www.openstreetmap.org/node/6644718876), but showing the `name` value would make it easier for user to find them if they need to edit them. Moreover, if a POI doesn't have an `addr:housenumber` value, the POI isn't shown on map (e.g. https://www.openstreetmap.org/node/6852375985)

imagico:
Duplicate of #3594 (which needs updating due to #3731) as far as `healthcare=laboratory` is concerned.  Duplicate of #1697 as far as `craft=photographer` is concerned.

`healthcare=physiotherapist` was part of #1981 - which however has been closed w.r.t. the rendering of synonyms for amenity=hospital etc.  So the request for rendering healthcare=physiotherapist would be valid, if you want to edit this issue to limit it to that please do so.  In general having separate issues for individual feature requests makes managing them easier.

healthcare=physiotherapist has about 9000 uses, mostly in Western Europe and North America.  It seems unlikely comparable services don't exist or are not mapped elsewhere so possibly use of the tag is tied to cultural aspects and comparable services elsewhere are more frequently tagged differently (possibly shop=massage).

Mannivu:
Ok, I've edited my comment deleting references to `healthcare=laboratory` and `craft=photographer`.

Issue #4222, opened 2020-10-16
Add rendering of healthcare=blood_donation


TheAdventurer64:
Quick question, why was the rendering for [this approved proposal](https://wiki.openstreetmap.org/w/index.php?oldid=1206637)  never added?

polarbearing:
Apparently the issue opener refers to https://wiki.openstreetmap.org/wiki/Tag:healthcare%3Dblood_donation

According to the code, the feature is rendered as a healthcare-coloured dot, see
``` [feature = 'healthcare_blood_donation']``` in amenity-points.mss

Could somebody make the issue title more meaningful, and close for being invalid?

imagico:
Because:

* proposal approval itself has no bearing on this style
* the tag is not very widely used
* it is questionable if this is suitable for rendering in a general purpose map style like this
* we have currently no consensus on the strategy for POI feature additions (#3635)
* a previous addition of rendering of this without consensus in bulk with various other healthcare tags (#3498) was reverted in #3731, see also #4183.

crse:
As for December 2025, the tag is approved and have emerging usage, with 2300+ usage.
https://taginfo.openstreetmap.org/tags/healthcare%3Dblood_donation

PR #4402, opened 2021-05-08
Add healthcare=laboratory

corentinfuret:
- Laboratories became facilities of major importance during the last year. These modifications will render healthcare=laboratory. At zoom 17 it will be only a red dot to be less visible than doctors or pharmacy. At zoom 18 and higher the full name and the icon will be displayed.

The icon came from The Noun Project (https://thenounproject.com/term/microscope/648/), has been adjusted to fit openstreetmap-carto guidelines by cquest and is under CC0.

https://www.openstreetmap.org/#map=19/47.46202/-0.55274

Before
example_laboratory_void_final

After
example_laboratory_final


imagico:
healthcare=laboratory has 7400 uses.

If this is merged it should be preceded or combined with a fix for #4403 - otherwise the code in amenity-points.mss will get even more confusing.

I am generally wary of adding new POIs before fixing #3880.  I also have doubts with the use of the healthcare color for features that by mapping conventions do not necessarily indicate facilities which provide services open to the public.

jeisenbe:
I agree that #4403 needs to be fixed. We should also think about this systematically. 

There are 2 other healthcare=* feature which are more common and which are not rendered yet:
- physiotherapist - https://taginfo.openstreetmap.org/tags/healthcare=physiotherapist#overview - see #4235
- healthcare=alternative is used 9000 times, mostly for chiropractors -  https://taginfo.openstreetmap.org/tags/healthcare=alternative

And there are several other healthcare=* tags which are fairly common but not rendered. It may make sense to address several of these at once. 
healthcare=psychotherapist - 4023 uses - https://taginfo.openstreetmap.org/tags/healthcare=psychotherapist
healthcare=rehabilitation - 2370 uses - https://taginfo.openstreetmap.org/tags/healthcare=rehabilitation
healthcare=podiatrist - 2140 uses - https://taginfo.openstreetmap.org/tags/healthcare=podiatrist
healthcare=optometrist - 2710 uses - https://taginfo.openstreetmap.org/tags/healthcare=optometrist

Also see less common tags: #4222 healthcare=blood_donation, #4295 healthcare=vaccination_centre

We previously closed #3706 and #3609   about healthcare=counselling, and healthcare=nurse which were rare at the time.

style/amenity-points.mss
```diff
@@ -605,6 +605,18 @@
     marker-clip: false;
   }
 
+  [feature = 'healthcare_laboratory'][zoom >= 17] {
+    [zoom >= 17][zoom < 18] {
+      marker-width: 4;
```
jeisenbe, line 610: We do not currently use any dots in this color. I do not think it will be intuitive

jeisenbe (changes requested):

PR #4419, opened 2021-06-05
updated the code for tourism_information

### LINKED ISSUE #4247: Remove rendering for tourism=information without supported value of information=*
### Expected behavior
* The tag `tourism=information` can be used for a wide variety of barely-related features. 
* 93% of features have the key `information=*` which more clearly specifies the feature. 
* The most common values are already supported with specific icons in this style: information=guidepost is most common, followed by information=board,   information=map, information=office. A tourism office is quite different from a guidepost or a public map and should not be confused. 
* Therefore, we should not be showing a symbol for features missing information=* or features with undocumented values of information=*

### Actual behavior
* We are rendering a non-specific “i” icon for all `tourism=information` objects
* This includes misleading uses such as `information=route_marker` and `information=trail_blaze` - the later appears to be a simple cut or paint blotch on a tree which helps mark the location of a rural pathway or trail, though it’s undocumented. See https://taginfo.openstreetmap.org/keys/information#values
* We are also rendering this symbol for the 7% of `tourism=information` objects which have no sub-tag, and could be anything from a guidepost to a visitor centre to a tourism office or a public map. 
 
### Examples with location

https://www.openstreetmap.org/node/7656793394
Screen Shot 2020-11-07 at 12 05 41

Link to image of this feature: https://images.mapillary.com/6kr3SGuwakA3RAxMDut2Ax/thumb-2048.jpg

https://www.openstreetmap.org/node/929360773
Screen Shot 2020-11-07 at 12 07 58

map-per:
Fixes #4247

Made tourism=information without or with undocumented information=* Tag disapperar

Test with tourism=information + information=yes: https://www.openstreetmap.org/node/4945458569
Before
OSM (3)

After
OSM (2)

Tested information=office, board, guidepost as well and as expekted no changes occured.


map-per:
For Points only tagged with tourism=information the icon doesn't render any more but the name ("Schloss Mühlental") is still displayed.
https://www.openstreetmap.org/node/4946241495
OSM (4)

map-per:
Fixed the names as well. They don't show up now:
https://www.openstreetmap.org/node/4946241495
OSM (5)

imagico (changes requested):
Thanks for working on this issue.

But it is not quite as simple.  In

https://github.com/gravitystorm/openstreetmap-carto/blob/b86de31da9e44e1a308eb7c8f57c3f4a9bce8956/project.mml#L1527

you still have all features with `tourism=information`.  So adding that tag to something with lower priority in the COALESCE() (like an office or a `place=locality`) will make that feature vanish from the map - which would be highly confusing.  So you need to limit the query to those features which are actually going to be rendered.

You also should remove the unused SVG.

map-per:
Thanks for the feedback. I limited the query to features that are actually rendered and deleted the general information symbol. But I think that these changes in project.mml make the former changes in amenity-points.mss superfluous. Thats why I proposed a change with less changes to amenity-points.mss as well ( #4420)

map-per:
Solved with #4420

PR #4420, opened 2021-06-06
Simplified version fixing tourism=information issuses #4247 as well

### LINKED ISSUE #4247: Remove rendering for tourism=information without supported value of information=*

map-per:
Fixes #4247
Simplified version that fixes tourism=information issuses #4247 as well

Made tourism=information without or with undocumented information=* Tag disapperar

Test with tourism=information + information=yes: https://www.openstreetmap.org/node/4945458569
Before
OSM (3)

After
Screenshot (653)

Tested information=office, board, guidepost as well and as expekted no changes occured. 
The Names of unwanted points and tourism=information points without information=* don't show up either.


map-per:
Is this change Ok or do I have to improve something?

imagico (dismissed):
Works fine.

project.mml
```diff
@@ -1524,7 +1524,7 @@ Layer:
                                           OR (boundary = 'protected_area' AND tags->'protect_class' IN ('1','1a','1b','2','3','4','5','6'))
                                           THEN boundary END,
                 'leisure_' || CASE WHEN leisure IN ('nature_reserve') THEN leisure END,
-                'tourism_' || CASE WHEN tourism IN ('information') THEN tourism END,
+                'tourism_' || CASE WHEN tourism IN ('information') AND tags->'information' IN ('audioguide', 'board', 'guidepost', 'office', 'map', 'terminal') THEN tourism END,
```
imagico, line 1527: tactile_map is missing in the list of values here.

imagico (changes requested):
Works as expected, just missing a value in SQL.  If you can add that this would be great.  Sorry for the delay in reviewing.

project.mml
map-per: Thanks, corrected the mistake and tested

imagico (approved):
Looks good now, thanks.

map-per:
Sorry, requesting review again as accidental.

PR #4424, opened 2021-06-19
Fix advertising column shape

enzet:
Side-to-side shape connections can lead to artifacts on some zoom levels.

I didn't create an issue since the fix is pretty small.

- Fix `advertising_column.svg` shape.

Before (you can see pale white line between `tip` and `top` shapes)

Screen Shot 2021-06-19 at 23 29 46
Screen Shot 2021-06-19 at 23 30 21

After

Screen Shot 2021-06-19 at 23 31 15
Screen Shot 2021-06-19 at 23 31 29

### DIFF
--- a/symbols/advertising_column.svg
+++ b/symbols/advertising_column.svg
@@ -6,6 +6,5 @@
 	 viewBox="0 0 14 14" enable-background="new 0 0 14 14" xml:space="preserve">
 <rect id="posters" x="4.5" y="4.108" style="fill-opacity:0.3;stroke-width:0.01526906;opacity:0.9;fill:#000000" width="5" height="8"/>
 <rect id="bottom" x="4" y="13" width="6" height="1"/>
-<rect id="top" x="4" y="2" width="6" height="1"/>
-<polygon id="tip" points="4,2 7,0 10,2 "/>
+<path id="top" d="M 4,3 L 4,2 7,0 10,2 10,3 Z"/>
 </svg>


polarbearing (commented):
Can you show examples how it affects rendering?
You apparently provide an overlap between the top bar and the tip. However how would I see such a white line as there is no pixel in between? At which magnification? Could it be an SVG-editor artefact rather an rendering effect?

enzet:
Hello @polarbearing! Thank you for the reply!

> Can you show examples how it affects rendering?

I think it doesn't affect the rendering of the map itself since we use pretty small pins but it may affect icon usage in other places. I spotted the artifact on the wiki page (screenshot from Chrome browser):

Screen Shot 2021-06-20 at 21 55 57

You can check it in your browser: https://wiki.openstreetmap.org/wiki/File:Column-14.svg

> You apparently provide an overlap between the top bar and the tip. 

I understand that this technique seems strange, but it's quite standard for such cases.  Some examples and explanations can be found [here](https://graphicdesign.stackexchange.com/questions/73601/is-there-a-reason-hillary-clintons-logo-has-hidden-notches) and [here](https://graphicdesign.stackexchange.com/questions/65058/image-looks-embossed-when-converted-to-svg).

> However how would I see such a white line as there is no pixel in between? At which magnification? Could it be an SVG-editor artefact rather an rendering effect?

All SVG-renderers that I saw had creates such artifacts on some fractional magnifications.

Another possible solution is to simply join `tip` and `top` shapes.  I didn't do that in this pull request because I'm not sure if we have to have them as separate shapes (e.g. for recoloring).

polarbearing:
Ok so I'm not sure as how urgent the maintainers here see that. 
I'd probably prefer the solution to join the top into a single polygon. In this style, the icons are coloured from black SVG  into the semantic colour as a whole by the renderer, thus the elements of the construction getting different colour is very unlikely.

enzet:
Well, in this case, merging is the better solution.

imagico (approved):
Looks good, thanks for spotting that.

As a note - most SVG symbols we have use a `path` feature for this kind of geometry - but a `polygon` seems perfectly fine as well.

Intend to merge this in case no one has objections.

enzet:
Yes, I think the `path` element is better. I've fixed it.

pnorman:
Why make overlapping shapes instead of one five-sided path?

enzet:
Now it is one five-sided path.

First, I made two overlapping shapes based on the assumption that they can be used in different ways, for example, to be recolored.

PR #4426, opened 2021-06-21
Add phantom labels for all bridges

### LINKED ISSUE #478: River label renders on top of bridge
Example:
screen shot 2014-04-16 at 1 23 20 pm

Bridge: http://www.openstreetmap.org/way/8147286
Waterway: http://www.openstreetmap.org/way/27600092

alfmarcua:
**Partially fixes #478** 

- Add a phantom label to all bridges in order to displace water labels for not to be on top of bridges

**Test rendering with links to the example places:**

Before:
Selección_006

After:
Selección_007

### DIFF
--- a/project.mml
+++ b/project.mml
@@ -1418,8 +1418,6 @@ Layer:
           WHERE ST_PointOnSurface(way) && !bbox!
             AND name IS NOT NULL
             AND man_made = 'bridge'
-            AND way_area > 125*POW(!scale_denominator!*0.001*0.28,2)
-            AND way_area < 768000*POW(!scale_denominator!*0.001*0.28,2)
           ORDER BY way_area DESC
         ) AS bridge_text
     properties:
--- a/project.mml
+++ b/project.mml
@@ -1822,6 +1822,32 @@ Layer:
         ) AS roads_area_text_name
     properties:
       minzoom: 15
+  - id: roads-text-name-phantom
+    geometry: linestring
+    <<: *extents
+    Datasource:
+      <<: *osm2pgsql
+      table: |-
+        (SELECT
+            way,
+            CASE WHEN substr(highway, length(highway)-4, 5) = '_link' THEN substr(highway, 0, length(highway)-4) ELSE highway END,
+            construction,
+            length(name)*!scale_denominator!*0.001*0.28/st_length(way) as char_per_pixel
+          FROM planet_osm_line l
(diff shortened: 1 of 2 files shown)


alfmarcua:
It is just a partial humble solution. There are cases where the solution still does not apply, the bridge label itself shall be displayed but it is large enough. Hence, it is not displayed. Here it is an example:

Before:
Selección_009

After:
Selección_008

pnorman:
I think the bigger concern is roads that are bridges, not the man_made=bridge structure.

imagico:
Yes, this is nice to see as a proof of concept for using pro forma labels to influence label placement in mapnik (and the limitations of this) but this will not solve the core of #478 and furthermore the pro forma labels for unnamed bridges will also block labels of bridge=yes roads which would be highly confusing for mappers.

alfmarcua:
I might have some ideas for roads as well. I will try to update with an approach. Although, I think that the main problem of this solution is that there are labels which are not displayed due to surrounding elements.

alfmarcua:
In this new commit I have tried to overcome the issue regarding roads that are bridges. 

In particular, a new layer is created, roads-text-name-phantom, just before the roads labels layer (roads-text-name). This layer contains only roads that are bridges and that have their names not null. In brings a new field which I called char_per_pixel which represents the number of characters that are per pixel.

The idea is that if the number of characters per pixel is below a threshold, then the label will fit the bridge and it could be represented. Hence, in the style it is declared that a phantom label will be placed whenever the number of characters per pixel of a road's label is above that threshold.

Notice that this threshold is text size dependant, so that property should be taken into account. In fact, the property phantom-line-text-threshold has been estimated from testing the limit of text size for a certain label of a bridge to be displayed or not. The tests have been with several different bridges.

Here is a snashot of the difference in Kosmtik comparison view:
- Left: OpenStreetMap
- Right: latest implementations

phantom-roads

alfmarcua:
@pnorman @imagico What do you think about this new approach? Did you have time to review it? 

It covers all bridges and it takes into account the size of the labels regarding the space they have in order to be represented.

pnorman:
I'd rather clean up the roads layers before adding any complexity around them,

jeisenbe (changes requested):
@alfmarcua - thank you for trying to solve this problem. 

Unfortunately, as mentioned in the issue #478 there is no simple way to fix these sort of problems with automated label placement in Mapnik. Creating an additional layer, as done in this PR, will add complexity to the code, increase computational cost, and slow down rendering, for a small improvement in the rendering result.

alfmarcua:
Thank you for your comments. As you comment, it is a complex problem as the optimal solution requires a lot of effort. Hence, I close this pull request.

PR #4430, opened 2021-06-26
only render addr:flats where value is 9 characters or less

### LINKED ISSUE #4160: Ugly rendering of addr:flats
### Expected behavior

The `addr:flats` tag is used to encode apartment numbers into OSM. Big apartments will usually have a big number of flats, and thus a long list under the `addr:flats` tag.

In contrast to `addr:unit` (which is meant to address a single unit), `addr:flats` is usually a list of apartments. These apartments are usually also on top of each other.

It's geographical info that can be used for address completion/verification, but IMO is very unsuited for rendering on a 2D map.

### Actual behavior

Ugly long lists of numbers which makes it almost impossible to find the housenumbers.

afbeelding

afbeelding

### Solution

Revert #3988 and display only `addr:unit` for separate addressed building units.

andrewharvey:
The vast majority of addr:flats values are 7 characters or less and this captures values like "101-909", however since significantly longer values are valid tags and do occur throughout OSM data, they are not rendered to prevent the map being overwhelmed by these labels, and to reduce temptation for people to delete these valid tags to "improve the map rendering".

Fixes #4160

### DIFF
--- a/project.mml
+++ b/project.mml
@@ -2082,7 +2082,7 @@ Layer:
             "addr:housenumber" AS addr_housenumber,
             "addr:housename" AS addr_housename,
             tags->'addr:unit' AS addr_unit,
-            CASE char_length(tags->'addr:flats') <= 7 WHEN TRUE THEN tags->'addr:flats' ELSE NULL END AS addr_flats,
+            CASE char_length(tags->'addr:flats') <= 9 WHEN TRUE THEN tags->'addr:flats' ELSE NULL END AS addr_flats,
             way_area/NULLIF(POW(!scale_denominator!*0.001*0.28,2),0) AS way_pixels
           FROM planet_osm_polygon
           WHERE way && !bbox! -- Not ST_PointOnSurface(way) because name might be NULL
@@ -2095,7 +2095,7 @@ Layer:
             "addr:housenumber" AS addr_housenumber,
             "addr:housename" AS addr_housename,
             tags->'addr:unit' AS addr_unit,
-            CASE char_length(tags->'addr:flats') <= 7 WHEN TRUE THEN tags->'addr:flats' ELSE NULL END AS addr_flats,
+            CASE char_length(tags->'addr:flats') <= 9 WHEN TRUE THEN tags->'addr:flats' ELSE NULL END AS addr_flats,
             NULL AS way_pixels
           FROM planet_osm_point
           WHERE way && !bbox!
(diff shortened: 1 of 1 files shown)


imagico:
As said in https://github.com/gravitystorm/openstreetmap-carto/issues/4160#issuecomment-868987197 i don't think dropping labels exceeding a certain length is a good idea.  Confusion about the logic behind some labels showing up while others don't (due to collision and prioritization) is already one of the most frequent complaints of map users - despite the current mechanism being reasonably easy to understand.  Adding additional complexity to that by hiding some labels based on length despite there being enough space to show them would significantly add to that problem.

pnorman (changes requested):
I'm okay with the idea of either not showing or cutting the labels short, but this PR won't do that reliably. If an object has addr:housenumber it will be included even if the addr:flats is > 7 characters.

andrewharvey:
I'll probably need to work out how to get this running locally to test it properly.

Personally I'm still leaning more towards dropping anything >7 chars than doing ..., but that's just my opinion.

jeisenbe (changes requested):
A limit of 7 characters will cause a simple and valid tag such as `addr:flats=1001-1099` not to render. This would be confusing.

andrewharvey:
> A limit of 7 characters will cause a simple and valid tag such as addr:flats=1001-1099 not to render. This would be confusing.

I increased it to 9.

project.mml
```diff
@@ -2082,7 +2082,7 @@ Layer:
             "addr:housenumber" AS addr_housenumber,
             "addr:housename" AS addr_housename,
             tags->'addr:unit' AS addr_unit,
-            tags->'addr:flats' AS addr_flats,
+            CASE char_length(tags->'addr:flats') <= 9 WHEN TRUE THEN tags->'addr:flats' ELSE NULL END AS addr_flats,
```
pnorman, line 2085: Don't use the switch-statement style case statement with an expression that evaluates to a boolean, instead do `CASE WHEN char_length(tags->'addr:flats') <= 9 THEN tags->'addr:flats' END AS addr_flats,`

project.mml
```diff
@@ -2095,7 +2095,7 @@ Layer:
...
```
pnorman, line 2098: Same as above for the switch-style case statement

pnorman (changes requested):
Some SQL changes.

I'm still not that convinced on the idea of restricting the length, but will defer to other maintainers there.

pnorman:
There are unresolved technical issues with this PR. I also don't think any of the maintainers think dropping long addr:flats tags is a good idea.

PR #4435, opened 2021-07-06
Added option for using local files when importing external data

nmaludy:
N/A

- I work in several disconnected (no internet) environments where the `get-external-data.py` script fails because it can't download the polygon files. However, we still need these files imported into the Postgres database in order to use openstreemap-carto with renderd. The changes here allow a user to pass in `-l/--local` and the download step will be skipped when importing the files. Instead it just reads from the data dir cache and re-runs the import.

### DIFF
--- a/scripts/get-external-data.py
+++ b/scripts/get-external-data.py
@@ -145,6 +145,8 @@ def main():
                         help="Name of configuration file (default external-data.yml)")
     parser.add_argument("-D", "--data", action="store",
                         help="Override data download directory")
+    parser.add_argument("-l", "--local", action="store_true",
+                        help="Expect files to exist locally in the data download directory. With this option, external downloads will not attempt to be retrieved from the internet. Instead they will simply be imported into the database.")
 
     parser.add_argument("-d", "--database", action="store",
                         help="Override database name to connect to")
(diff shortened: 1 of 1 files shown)


scripts/get-external-data.py
```diff
@@ -145,6 +145,8 @@ def main():
                         help="Name of configuration file (default external-data.yml)")
     parser.add_argument("-D", "--data", action="store",
                         help="Override data download directory")
+    parser.add_argument("-l", "--local", action="store_true",
+                        help="Expect files to exist locally in the data downloa directory. With this option, external downloads will not attempt to be retrieved from the internet. Instead they will simply be imported into the database.")
```
pnorman, line 149: typo, "downloa"

scripts/get-external-data.py
```diff
@@ -227,15 +230,16 @@ def main():
...
-                download.raise_for_status()
+                if not opts.local:
+                    download = s.get(source["url"], headers=headers)
+                    download.raise_for_status()
 
-                if (download.status_code == 200):
-                    if "Last-Modified" in download.headers:
+                if opts.local or (download.status_code == 200):
+                    if not opts.local and "Last-Modified" in download.headers:
                         new_last_modified = download.headers["Last-Modified"]
```
pnorman, line 239: If `ops.local` is set, `download` will be uninitialized. What will the lines below do?

pnorman (changes requested):

scripts/get-external-data.py
nmaludy: `opts.local` guards against utilizing `download`, the if conditional is read left to right and bails out early, never comparing `download.status`.

Conversely `not opts.local` is a check to allow using `download`, guarded above.

pnorman:
Closing in favour of #4468

PR #4205, opened 2020-09-12
Clean up some svg images

pitdicker:
When rendering my local area with Mapnik I get a bunch of errors (apparently not fatal). The first bunch:
```
Mapnik LOG> 2020-09-11 21:09:25: SVG PARSING ERROR:"SVG support error: <font-style> attribute is not supported"
Mapnik LOG> 2020-09-11 21:09:25: SVG PARSING ERROR:"SVG support error: <font-variant> attribute is not supported"
Mapnik LOG> 2020-09-11 21:09:25: SVG PARSING ERROR:"SVG support error: <font-weight> attribute is not supported"
Mapnik LOG> 2020-09-11 21:09:25: SVG PARSING ERROR:"SVG support error: <font-stretch> attribute is not supported"
Mapnik LOG> 2020-09-11 21:09:25: SVG PARSING ERROR:"SVG support error: <font-size> attribute is not supported"
Mapnik LOG> 2020-09-11 21:09:25: SVG PARSING ERROR:"SVG support error: <font-family> attribute is not supported"
Mapnik LOG> 2020-09-11 21:09:25: SVG PARSING ERROR:"SVG support error: <text-decoration> attribute is not supported"
Mapnik LOG> 2020-09-11 21:09:25: SVG PARSING ERROR:"SVG support error: <letter-spacing> attribute is not supported"
Mapnik LOG> 2020-09-11 21:09:25: SVG PARSING ERROR:"SVG support error: <word-spacing> attribute is not supported"
Mapnik LOG> 2020-09-11 21:09:25: SVG PARSING ERROR:"SVG support error: <writing-mode> attribute is not supported"
Mapnik LOG> 2020-09-11 21:09:25: SVG PARSING ERROR:"SVG support error: <direction> attribute is not supported"
Mapnik LOG> 2020-09-11 21:09:25: SVG PARSING ERROR:"SVG support error: <dominant-baseline> attribute is not supported"
Mapnik LOG> 2020-09-11 21:09:25: SVG PARSING ERROR:"SVG support error: <baseline-shift> attribute is not supported"
Mapnik LOG> 2020-09-11 21:09:25: SVG PARSING ERROR:"SVG support error: <text-anchor> attribute is not supported"
Mapnik LOG> 2020-09-11 21:09:25: SVG PARSING ERROR:"SVG support error: <clip-rule> attribute is not supported"
Mapnik LOG> 2020-09-11 21:09:25: SVG PARSING ERROR:"SVG support error: <overflow> attribute is not supported"
[...]

PR #4453, opened 2021-08-26
Shield dimensions

pitdicker:
In https://github.com/gravitystorm/openstreetmap-carto/pull/4205 I offered to clean up all svg files, and work on the problems that keep some patterns from being usable as vector images.
I have been at it for a couple of weeks, and am mostly there (also working on three related Mapnik issues).
This is the first of three PRs. The rest wil come as time allows ;-)

This PR is a tiny one that only touches the shields.
Previously they were generated with a width and height of 100%. Percentages are allowed by the SVG specification, but don't make much sense. Inkscape, viewers, browsers and Mapnik differ in their interpretation. Is the percentage relative to a screen, the whole map, a tile, or user supplied marker dimensions? Or to implicit image dimensions?

In the case of mapnik it falls back to the width and height of the viewBox, and scales them by the percentage. If the SVG contains no viewBox, it acts as if no width and height are given.

It all doesn't matter much in the end, as the rendering in Mapnik was already as desired. But it was little work to update the shield generation script to Python 3 (no expert by far, but it seems to work), and change it to simply write out the true width and height.


pitdicker:
@imagico Thank you for your quick reply. And I have definitely studied a number of your PRs around patterns, jsdotpattern, your blog and the repo! I'd like to hear your thoughts once the PR is up.

@pnorman This week I can make the other two PRs. At least the first one is hopefully not hard to review, that does some cleanup on the SVG of all symbols. (Only thing left for the first is write a good description.) Maybe worth waiting for with https://github.com/gravitystorm/openstreetmap-carto/issues/4454?

Issue #3653, opened 2019-01-21
Alternative for comp-op="darken"


Klaus-Tockloth:
The current style (4.19) uses the feature 'comp-op="darken"' three times:

```
<Style name="admin-low-zoom" filter-mode="first" comp-op="darken" opacity="0.4">
<Style name="admin-mid-zoom" filter-mode="first" comp-op="darken" opacity="0.5">
<Style name="admin-high-zoom" filter-mode="first" comp-op="darken" opacity="0.5">
```

Question: Is there an alternative to achieve a similar result?

Background: "comp-op" prevents mapnik/carto to create a svg file with "proper" vector data. 
Instead a svg file with an (huge) embedded png map (base64 encoded) will be created. Same problem for pdf.

kocio-pl:
I have heard complaints about SVG export at least 2 times on forum and on help.osm.org, I would be happy if this problem could be fixed. Anybody interested in fixing this?

imagico:
To my knowledge the only other way to get consistent rendering of admin boundaries with semi-transparency is by preprocessing the data.  This is a workable approach but not necessarily a desirable one here.

sommerluk:
Does https://github.com/mapnik/mapnik/pull/4012 help here? Is part of the newest Mapnik release 3.0.22…

talaj:
It seems like comp-op is only available in Cairo SVG surface for SVG version >= 1.2, see https://github.com/behdad/cairo/commit/5336cf5a984cdeb4d4887cd3c643a3e051aca338. Since Mapnik outputs SVG with `version="1.1"`, I think this is the culprit.

On the other hand PDF with `comp-op="darken"` should be rendered correctly. I tried that with Mapnik 3.0.x and Cairo 1.15.

talaj:
Yes, SVG 1.1 is the default standard for Cairo SVG Surface. How I could overlook that  :angry:!

https://github.com/behdad/cairo/blob/7a923b1ff5bf3e2b79ce4f2fe14d352faef5bccb/src/cairo-svg-surface.c#L276

talaj:
Is this ancient script still being used for expert to PDF and SVG?
https://trac.openstreetmap.org/browser/sites/tile.openstreetmap.org/cgi-bin/export

If yes, to allow `comp-op` in SVG, this has to be put after line 120 of the script:
```python
surface.restrict_to_version(cairo.SVG_VERSION_1_2)
```

Or is it somewhere in git?

[...]

PR #4467, opened 2021-09-13
Small fixes to golf features

pitdicker:
Two minor fixes after the merge of https://github.com/gravitystorm/openstreetmap-carto/pull/4381. They are mostly to keep my own setup working, but I hope it makes sense for others.

- Define a `golf-color` for the golf holes, lines and text instead of reusing `address-color`.
- Don't use `comp-op: soft-light` with the `golf_rough`.
- Not fixed: the pattern is a cropped 40x40px SVG, that doesn't repeat properly. I would like to change that in https://github.com/gravitystorm/openstreetmap-carto/pull/4463.

**About `comp-op`**
This is a feature supported by Mapnik, but not really by its SVG export. https://github.com/gravitystorm/openstreetmap-carto/issues/3653 hit that issue. What happens is that the exported SVG either contains a rasterized image, or uses non-standard extensions (when exporting with the non-existing SVG 1.2). At some point proposals for SVG 1.2 including `comp-op` floated around and were implemented in Inkscape, librsvg and Mapnik, but those proposals were later dropped.

The result in most image viewers (those that are not based on librsvg) and in all browsers:
golf_rough_blend
5x:
golf_rough_blend_5x

The only current use of `comp-op` before https://github.com/gravitystorm/openstreetmap-carto/pull/4381 was for admin boundaries, and it is a lot less noticeable when those don't look perfectly as intended.

I believe using `comp-op` in OSM-carto is desirable, and it would be nice to get workarounds at some point. SVG supports masks and filters, so a lot should be possible. But can the use of `comp-op` and the need for workarounds (or dropping support for exporting vector images) be delayed until there is a feature that is really worth it, such as https://github.com/gravitystorm/openstreetmap-carto/issues/3854?

For now I would just like to use the simple solution: adjust the color of the SVG pattern for `golf_rough`, to use the color `comp-op: soft-light` would have resulted in:
golf_rough

@jgruca Are these changes acceptable?


jgruca:
No objections from me. Would it be preferable to set the color for the SVG in the MSS, though, per the [current map icon guidelines](https://github.com/gravitystorm/openstreetmap-carto/blob/master/CONTRIBUTING.md#map-icon-guidelines)?

pitdicker:
:+1: If only we could set the color... Mapnik doesn't support such a setting for patterns ([polygon-pattern reference](http://mapnik.org/mapnik-reference/#3.0.22/polygon-pattern)).

imagico:
comp-op should not be used in patterns - if i had seen that in #4381 i would have requested to change that.  Just design the pattern to directly generate the desired results on the background it is to be used on (or with the background included in the pattern).  That is simple enough to do, jsdotpattern allows to preview the pattern in a chosen color.

But in general we will likely continue to use comp-op elsewhere - currently in particular for admin boundaries, potentially also for #3854 and in #4128.

jeisenbe:
@pitdicker would you be able to address this? 

"design the pattern to directly generate the desired results on the background it is to be used on (or with the background included in the pattern)"

jgruca:
I believe this was addressed by setting the pattern color directly in the SVG file.

pitdicker:
@jeisenbe As @jgruca says. To say a bit more:

Before making this PR I played a few hours with jsdotpattern, trying to get a similar looking pattern that would fit in a small tile such as 64x64. Just an experiment, and I got nowhere :-).

To keep the changes minimal, I made this PR with just the color adjusted in the already existing SVG, based on jsdotpattern. The end result is no different from setting the color in jsdotpattern and regenerating the SVG, except that I now didn't need to figure out a similar looking command sequence.

In https://github.com/gravitystorm/openstreetmap-carto/pull/4463#issuecomment-918538518 I added a 128x128 version of the `golf_rough` pattern, based on the jsdotpattern sequence of the `beach` pattern. But the intention of this PR was to just be a minimal quick fix.

PR #4468, opened 2021-09-14
Allow file:// urls in the config file

stephan2012:
Signed-off-by: Stephan Austermühle <au@hcsd.de>

This small technical PR allows using `file://` urls in the config file (`.sources.*.url`) which is helpful for loading external data in air-gap environments, i.e., systems without Internet connection.

This workflow illustrates the need for the change:

1. Download external files (requires [yq](https://github.com/mikefarah/yq)):

    ```bash
    yq eval '.sources.*.url' external-data.yml |xargs -P4 -i curl -x "${https_proxy:-}" --silent --fail --output-dir external-data -LO {}
    ```

2. Replace external by file urls:

    ```bash
    yq -i eval '.sources.*.url |= sub("^https://.*/(.*)$", "file:///var/lib/openstreetmap-carto/external-data/${1}")' external-data.yml
    ```

3. `get-external-data.py` can now load the files locally


pnorman (changes requested):
When everything is installed as per the documentation, you get this error
```
  File "scripts/get-external-data.py", line 25, in <module>
    from requests_file import FileAdapter
ModuleNotFoundError: No module named 'requests_file'
```

The documentation needs updating. Additionally, I would prefer not to require an additional package in the common case where local files are not being used.

pnorman:
Both this and #4435 would accomplish the goals, correct? I'm not sure which is the better approach.

stephan2012:
> Both this and #4435 would accomplish the goals, correct? I'm not sure which is the better approach.

@nmaludy and I implemented the same feature at the same time, so obviously there is a need for it. 😉

Agree, the documentation requires an update. Will add this unless someone decides for the other approach. Besides that, IMHO adding a package (`python3-requests-file`) is really nothing to worry about.

INSTALL.md
```diff
@@ -102,7 +102,7 @@ To display any map a database containing OpenStreetMap data and some utilities a
 * [PostgreSQL](https://www.postgresql.org/)
 * [PostGIS](https://postgis.net/)
 * [osm2pgsql](https://github.com/openstreetmap/osm2pgsql#installing) to [import your data](https://switch2osm.org/loading-osm-data/) into a PostGIS database
-* Python 3 with the psycopg2, yaml, and requests libraries (`python3-psycopg2` `python3-yaml` `python3-requests` packages on Debian-derived systems)
+* Python 3 with the psycopg2, yaml, and requests libraries (`python3-psycopg2`, `python3-yaml`, `python3-requests`, `python3-requests-file` packages on Debian-derived systems)
```
pnorman, line 105: If we use commas, we need to specify "and". Personally, I slightly prefer without commas so you can copy and paste the package list

pnorman (changes requested):
I haven't tested the technical changes yet - I need to get home to do that - but there's a minor documentation issue.

Otherwise, good with the approach.

PR #4431, opened 2021-06-26
Use flex backend

pnorman:
Fixes #4112
This PR implements the flex backend and is ready for a first review of the changes. So far only new tables are added, but I'd like to get feedback on the lua changes.

Two new tables are added. `planet_osm_route` contains route relation information but no geometries. This allows shields from route relations. `planet_osm_admin` contains linestrings for ways that are part of admin relations. This enables deduplicating the linestrings on import, rather than relying on tricky overlaps and transparency.

I have not yet made any SQL or MML changes

Issue #3504, opened 2018-11-09
Virtual borders on antimeridian


kocio-pl:
There are virtual borders on antimeridian, which of course should not be there:

https://www.openstreetmap.org/way/239992336#map=6/67.306/-177.726

![screenshot_2018-11-09 openstreetmap 1](https://user-images.githubusercontent.com/5439713/48283794-dc111a80-e45d-11e8-815b-19bd285224b4.png)

https://www.openstreetmap.org/way/42394837#map=9/51.9933/179.8860

![screenshot_2018-11-09 openstreetmap](https://user-images.githubusercontent.com/5439713/48283746-b421b700-e45d-11e8-895e-b4fee26c97e9.png)

kocio-pl:
This problem is visible also on HOT layer, but is not present on French fork:

http://tile.openstreetmap.fr/?zoom=9&lat=51.87829&lon=-180.54689&layers=B000000FFFFFF
https://github.com/cquest/osmfr-cartocss

![screenshot_2018-11-09 tile openstreetmap fr](https://user-images.githubusercontent.com/5439713/48284073-b1739180-e45e-11e8-9f61-15d77dbde46d.png)

jeisenbe:
This bug is still visible at https://www.openstreetmap.org/#map=9/51.9933/179.8860

@imagico - I recall you mentioned a way to solve a similar problem for archipelagos. How can we fix this problem for admin boundary relations?

pitdicker:
Isn't it the case that all data is split up on the antimeridian? I see roads and rivers that don't even visually attach when crossing the boundary.

mboeringa:
> Isn't it the case that all data is split up on the antimeridian?

Yes, in most cases. However, sometimes multipolygon relations are defined that "connect" the two parts in an "administrative" way. However, these MP relations cause issues with many GIS systems and PostGIS functions designed for planar coordinate systems, as there is no inherent knowledge of the antimeridian in a planar coordinate system like Web Mercator.

E.g. if you try to calculate a bounding box for such a multipolygon spanning the antimeridian, it will cover the globe:
https://github.com/openmaptiles/openmaptiles/issues/1122#issuecomment-913147529

pitdicker:
Then I wonder if there is any solution possible here except pre-processing admin boundaries (and maybe other geometry).

kocio-pl:
Does anyone know how the French fork deals with it?

pitdicker:
[...]

Issue #712, opened 2014-07-10
Don't render high(local) administrative borders on coastlines


matthijsmelissen:
Don't render adminlevel 8 boundaries if the coincede with coastlines.

See also https://trac.openstreetmap.org/ticket/2670

kocio-pl:
If I remember, the main problem were proposed changes to the tagging borders, not the rendering. If this one would not mean such changes, it wouldn't be hard to deploy it.

matthijsmelissen:
There are two types of tagging: tagging the relation that includes all boundary lines of an admin area, of tagging the boundary lines themselves. We render based on the first approach (relation), and people have a problem with rendering based on the second approach (boundary lines).

However, whether a boundary is maritime or not is stored in the individual boundary lines, so if we render based on the relation, we have no access to that.

jeisenbe:
If #4092 is merged we would have ways for the coastlines in the database, which might make it easier to solve this issue for admin boundaries with precisely follow the coastline

PR #4471, opened 2021-09-16
Use masking instead of darken blending of admin boundaries and ferry routes

pitdicker:
### Current
The admin boundaries are currently drawn with some complicated tricks to avoid strange renderings where boundaries overlap. It is build around `comp-op: darken`, applied at the attachment level. Within an attachment layer, for every boundary there is first drawn a white line to cover any previous boundaries. Then the new boundary is drawn on top. The entire attachment layer is blended on top of the map so far, where only the pixels that are darker than the map are blended. Because the white lines that cover up overlapping boundaries are lighter, they fall away when blending. This has the tricky assumption that boundary lines are always darker than what is already drawn on the map.

The admin boundaries have different styles for every admin level. The most importand ones, `admin_level <= 4`, are a wide line with a thin center line. The others admin levels are drawn with thin lines. This is currently drawn with three attachment layers:
- `:firstline`. All lines of `admin_level > 4`. To keep blending with `comp-op: darken` working as intended, lines from `admin_level <= 4` are drawn as white background lines.
- `:wideline`. Wide lines of `admin_level <= 4`.
- `:narrowline`. The center lines of `admin_level <= 4`.

The only reason not to combine `:firstline` and `:wideline` is so that their opacity can be seperately controlled. Until now they both have been the same, 0.5. The cost is ca. 100 lines of mss, and having to draw the background lines of `admin_level <= 4` one more time than necessary.

White background lines are always drawn, even when not needed in the cases where a boundary line is solid.

### New
- Use `comp-op: dst-out` at the line level instead of `comp-op: darken` at the layer level. The result is the same, but works it independent of lightness.
- Merge `:firstline` and `:wideline`.
- Don't draw background lines behind solid lines.

### Aliasing artifacts
`comp-op: darken` doesn't take the opacity of a pixel into account, only the lightness. I don't know the lightness formula for aliased pixels, but it worked out. With the current foreground and background colors the covered aliasing pixels along a border line were lighter than the background color. With the darken blending filter they became invisible.

Masking doesn't have such a luxury. Any aliased pixel with not be fully masked out. A solution is to adjust the gamma AGG uses for anti-aliasing. Setting it to 0 for the mask lines makes them fully mask out the aliased pixels. But it does lead to small artifacts at three-way intersections. In my opinion this is not a problem, and maybe even looks good. At these points there would always be small aliasing artifacts anyway, and because a three-way point rarely look perfect because the dashes would have to align just right. Now it looks like one of the dashes ends just before the intersection.

Aliasing issue:
aliasing
With `gamma: 0`:
aliasing_gamma0

### Ferry routes
Ferry routes work around the same problem as admin boundaries: multiple dashed lines that can overlap, but then the dashes may not overlap and cause the line to look solid. The current solution is to draw a background line with the water color. This doesn't look optimal where a ferry route crosses a tunnel. Also it doesn't work with https://github.com/gravitystorm/openstreetmap-carto/pull/4128. I changed them to use the same solution as admin boundaries.

### Advantages
- I am quite happy with how these changes shaved off a quarter of the mss lines related to admin borders.
- Instead of two background lines and a border line for boundaries at `admin_level <= 4`, now only the border line is drawn.
- Basically all hidden complexity around drawing admin boundaries is now gone, such as why `:firstline` was used to draw invisible lines for `admin_level <= 4` and only starting at `zoom >= 8`.
- Masking is a simpler operation than `comp-op: darken`. This helps with my usecase: SVG files generated with Mapnik, which get some further processing before turning them into pdf. Any files with `comp-op` generated by Mapnik are non-standard (='broken'), and have to be fixed by a script. But an SVG with a `comp-op` that does masking is much easier to fix than one that does blending, because masking is a well-supported feature by viewers while blending is not even supported well enough by browsers. This is the first time since https://github.com/gravitystorm/openstreetmap-carto/pull/1107 that I have fully working SVG exports again 🎉.

### Previews
No visible changes to admin boundaries. After:

Ferry routes crossing a tunnel before:
ferry_before
Ferry routes crossing a tunnel after:
ferry_after

### DIFF
--- a/style/admin.mss
+++ b/style/admin.mss
@@ -8,10 +8,6 @@
 // The SQL has `ORDER BY admin_level DESC`, so the boundary with the lowest
 // admin_level is rendered on top.
 //
-// Three attachments are used, with minor borders before major ones, and the
-// thin centerline last, to handle overlapping borders correctly and allow each
-// type to have a different level of opacity.
-//
 // Where borders overlap, the overlapping border maskes the one below it. This
 // is done by drawing a background line behind each boundary line with
 // comp-op: dst-out.
@@ -19,24 +15,10 @@
 #admin-low-zoom[zoom < 8],
 #admin-mid-zoom[zoom >= 8][zoom < 13],
 #admin-high-zoom[zoom >= 13] {
-  ::firstline { opacity: 0.5; }
-  ::wideline { opacity: 0.5; }
-  ::narrowline { opacity: 0.6; }
+  ::fill { opacity: 0.5; }
+  ::centerline { opacity: 0.6; }
 
-  [admin_level = '2']::firstline {
-    [zoom >= 8] {
-      background/line-join: bevel;
-      background/line-comp-op: dst-out;
-      background/line-width: 3;
-    }
-    [zoom >= 9] { background/line-width: 3.5; }
-    [zoom >= 10] { background/line-width: 4.5; }
-    [zoom >= 11] { background/line-width: 5.5; }
-    [zoom >= 12] { background/line-width: 6; }
-    [zoom >= 13] { background/line-width: 7; }
-    [zoom >= 14] { background/line-width: 8; }
-  }
-  [admin_level = '2']::wideline {
+  [admin_level = '2']::fill {
     [zoom >= 4] {
       background/line-join: bevel;
       background/line-comp-op: dst-out;
@@ -87,7 +69,7 @@
       line-width: 8;
     }
   }
-  [admin_level = '2']::narrowline {
+  [admin_level = '2']::centerline {
     [zoom >= 8] {
       background/line-join: bevel;
       background/line-comp-op: dst-out;
@@ -125,20 +107,7 @@
     }
   }
 
-  [admin_level = '3']::firstline {
-    [zoom >= 8] {
-      background/line-join: bevel;
-      background/line-comp-op: dst-out;
-      background/line-width: 1.8;
-    }
-    [zoom >= 9] { background/line-width: 2.5; }
-    [zoom >= 10] { background/line-width: 3.2; }
-    [zoom >= 11] { background/line-width: 4; }
-    [zoom >= 12] { background/line-width: 4.5; }
-    [zoom >= 13] { background/line-width: 5; }
-    [zoom >= 14] { background/line-width: 5.5; }
-  }
-  [admin_level = '3']::wideline {
+  [admin_level = '3']::fill {
     [zoom >= 4] {
       background/line-join: bevel;
       background/line-color: white;
@@ -189,7 +158,7 @@
       line-width: 5.5;
     }
   }
-  [admin_level = '3']::narrowline {
+  [admin_level = '3']::centerline {
     [zoom >= 10] {
       background/line-join: bevel;
       background/line-comp-op: dst-out;
@@ -219,20 +188,7 @@
     }
   }
 
-  [admin_level = '4']::firstline {
-    [zoom >= 8] {
-      background/line-join: bevel;
-      background/line-comp-op: dst-out;
-      background/line-width: 1;
-    }
-    [zoom >= 9] { background/line-width: 1.5; }
-    [zoom >= 10] { background/line-width: 2; }
-    [zoom >= 11] { background/line-width: 2.5; }
-    [zoom >= 12] { background/line-width: 3; }
-    [zoom >= 13] { background/line-width: 3.5; }
-    [zoom >= 14] { background/line-width: 4; }
-  }
-  [admin_level = '4']::wideline {
+  [admin_level = '4']::fill {
     [zoom >= 4] {
       background/line-join: bevel;
       background/line-comp-op: dst-out;
@@ -284,7 +240,7 @@
       line-width: 4;
     }
   }
-  [admin_level = '4']::narrowline {
+  [admin_level = '4']::centerline {
     [zoom >= 10] {
       background/line-join: bevel;
       background/line-comp-op: dst-out;
@@ -313,7 +269,7 @@
     }
   }
 
-  [admin_level = '5'][zoom >= 8]::firstline {
+  [admin_level = '5'][zoom >= 8]::fill {
     background/line-join: bevel;
     background/line-comp-op: dst-out;
     background/line-width: 0.6;
@@ -347,7 +303,7 @@
       line-dasharray: 20,2,8,2;
     }
   }
-  [admin_level = '6'][zoom >= 10]::firstline {
+  [admin_level = '6'][zoom >= 10]::fill {
     background/line-join: bevel;
     background/line-comp-op: dst-out;
     background/line-width: 1;
@@ -371,7 +327,7 @@
       line-dasharray: 16,2,3,2;
     }
   }
-  [admin_level = '7']::firstline {
+  [admin_level = '7']::fill {
     [zoom >= 11] {
       background/line-join: bevel;
       background/line-comp-op: dst-out;
@@ -393,7 +349,7 @@
       line-dasharray: 12,2,3,2,3,2;
     }
   }
-  [admin_level = '8']::firstline {
+  [admin_level = '8']::fill {
     [zoom >= 12] {
       background/line-join: bevel;
       background/line-comp-op: dst-out;
@@ -411,7 +367,7 @@
     }
   }
 
-  [admin_level = '9'][zoom >= 13]::firstline {
+  [admin_level = '9'][zoom >= 13]::fill {
     background/line-join: bevel;
     background/line-comp-op: dst-out;
     background/line-width: 1.2;
@@ -426,7 +382,7 @@
       line-dasharray: 0,4,2,2,3,2,2,4;
     }
   }
-  [admin_level = '10'][zoom >= 14]::firstline {
+  [admin_level = '10'][zoom >= 14]::fill {
     background/line-join: bevel;
     background/line-comp-op: dst-out;
     background/line-width: 1.2;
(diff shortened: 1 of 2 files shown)


pitdicker:
Note that this PR pioneers a new technique for OSM-carto: masking. I see no real problems caused by it, but had to learn two tricks:
- setting a gamma value for the masking lines to avoid aliasing issues.
- limit the masking operation to only the current layer, by applying a layer-level opacity.

It can be a good tool for other currently hard to solve situations, such as properly rendering the `highway-area` layer seemingly on top of road casings, but underneath tunnels, buildings and barriers.

pitdicker:
One more comparison: with `comp-op: darken` the pixels along a boundary where not aliased consistently, while they are with masking. 

With `comp-op: darken`:
aliasing_darken
With masking:
aliasing_gamma0

jeisenbe:
I haven’t reviewed the code but the preview images look good.

kocio-pl:
> This is the first time since Reduce number of overlapping admin borders #1107 that I have fully working SVG exports again tada.

I consider it to be very useful change in itself.

mboeringa:
I think this whole issue is superseded by @pnorman's work on the [new flex style](https://github.com/gravitystorm/openstreetmap-carto/pull/4431), that includes a new de-duplicated 'planet_osm_admin' table, which means there will no longer be overlapping admin boundary lines.

pitdicker:
> I think this whole issue is superseded by @pnorman's work on the [new flex style](https://github.com/gravitystorm/openstreetmap-carto/pull/4431)

True, should have mentioned that. This PR would be useful in the meantime, and after that only for the ferry routes.

It is silently the reason I am looking forward to https://github.com/gravitystorm/openstreetmap-carto/pull/4431 (don't understand the rest of the advantages yet). But I have tried to work on similar boundary de-duplication in an other project before, and gave up because I could imagine too many edge cases. So I am cautiously hopeful, but thought an alternative solution would be good to share.

To give a few concerns, mostly because I haven't studied the solution well, and don't know how clean the data is: Can https://github.com/gravitystorm/openstreetmap-carto/pull/4431 handle it when two borders have their points in opposite order? When one border has an extra point falling on a straight line, not shared by the other? Or T-intersection of boundaries, where the border at the bar of the T doesn't have a point? Would the slicing up of a border affect text placement, or the dash pattern near intersections? Nothing against https://github.com/gravitystorm/openstreetmap-carto/pull/4431, it would be the cleaner solution, it just feels more tricky to work an the data than figure things out at rasterizing time.

imagico:
I have looked at this a bit and have to say i don't fully understand how it works technically but the combination of using `opacity: 0.99999;` and `mask/line-gamma: 0;` makes me wary that you are making use of some fragile particularity in the way Mapnik implements comp-op.

Here a comparison at normal scale:

**This PR:**
this

**Master:**
this

mboeringa:
> To give a few concerns, mostly because I haven't studied the solution well, and don't know how clean the data is: Can #4431 handle it when two borders have their points in opposite order? When one border has an extra point falling on a straight line, not shared by the other? Or T-intersection of boundaries, where the border at the bar of the T doesn't have a point? Would the slicing up of a border affect text placement, or the dash pattern near intersections? Nothing against #4431, it would be the cleaner solution, it just feels more tricky to work an the data than figure things out at rasterizing time.

I think Paul is by far the best person to answer these questions, but I have the feeling that you assume that de-duplicating takes place in PostgreSQL after creating duplicate lines. AFAIU the code of the flex style now, that is not the case. OpenStreetMap boundary relations, by themselves as stored in the API database, are not storing duplicate geometries, only referencing the shared boundary line from multiple relations. It is the processing in osm2pgsql and other frameworks to generate e.g. Simple Feature "Polygon" objects from them, that potentially creates duplication. 

By adjusting the processing at the osm2pgsql import stage through the new flex options, it is possible to avoid this duplication.

pitdicker:
> I have looked at this a bit and have to say i don't fully understand how it works technically but the combination of using `opacity: 0.99999;` and `mask/line-gamma: 0;` makes me wary that you are making use of some fragile particularity in the way Mapnik implements comp-op.

About the opacity part: it is a hack, but practical. All the masking should take place within the `admin-*` layer, and within the `ferry-routes` layer. But `comp-op` has no way to specify the destination of its blending/masking. It applies to the entire canvas, with everything that is already drawn on it.

Mapnik documents one solution: instead of directly drawing on the main canvas, we can get a blank canvas for a layer by using `comp-op` or `opacity` [at the style level](https://github.com/mapnik/mapnik/wiki/Compositing#style-level-compositing). Then all `comp-op` operations are restricted within that layer.

How it used to work on master was that `comp-op: darken` was applied at the layer attachment level (style level). So all admin lines were first drawn on an empty canvas, and then blended onto the main canvas. We still want to have that seperate canvas somehow.

The wiki does say:
> The default (if no `comp-op` is set on a Style) is to skip the creation of a temporary canvas. So, while setting the `comp-op` on a style to `src-over` will invoke the default blending method, but it will also be triggering the rendering all the entire style to a separate canvas, which can lead to different output - perhaps desirable, perhaps not - just be aware of this.

This documentation is an entirely reasonable way for `comp-op` to behave, but turns out not to work anymore. I consider this a bug in Mapnik, that is probably not too hard to fix. `opacity: 1.0;` is documented to have the same effect, but is also broken. Maybe I'll even give fixing it a try. For now we need another solution.

`opacity: 0.99999` works because it rounds to `1.0` but is nog recognized as such. An alternative is `comp-op: src-atop` which is identical to `comp-op: src-over` when the main canvas is a solid background, as it is for us. I have no strong preference. Both don't really feel fragile to me, changing how `opacity` or `comp-op` work at the layer level would be a breaking change for Mapnik.

------------

`mask/line-gamma: 0;` is documented in the first post under `Aliasing artifacts`. Suppose we have line A below, and a masking line B covering it. Pixels drawn along the line are partially transparent because of aliasing. The same is true for the masking line. A partially transparent pixel from the masking line will not fully blot out the underlying pixel.

The current hack in `ferry-routes` was to draw an arbitrary wider background line (where the width did not get updated when the dashed line becomes wider at a higher zoom level). How wide should the line exactly be? Instead of guessing, I remembered that the AGG feature to deal with aliasing issues is to control gamma. With it we can just draw a line with the same width, and `line-gamma: 0`.

I have no problem with guessing a wider masking line, but using `line-gamma` feels much cleaner.

You made a good preview image again to show the differences! With GIMP I can measure tiny color changes in the image from master that show the background line does not fully cover the aliased pixels, but invisible to the eye.

Master looks better in this pretty artificial test with close almost parallel lines. I suppose we don't need to set gamma to all the way to `0.0`, but can approach master with something like `0.15`.

------------

> OpenStreetMap boundary relations, by themselves as stored in the API database, are not storing duplicate geometries, only referencing the shared boundary line from multiple relations.

Great. Than I continue to look forward to https://github.com/gravitystorm/openstreetmap-carto/pull/4431! Would this PR make sense in the meantime, or maybe just for the ferry routes?

jeisenbe:
@pnormal #4431 might also fix this, but would this PR make sense in the meantime?

@pitdicker how do the ferry routes look with gamma 0.15?

mboeringa:
> @pnorman #4431 might also fix this, but would this PR make sense in the meantime?

I have been using my own variant of Paul's work in #4431 for quite some time now, and think it quite solid by now, and have commented on a couple of issues with the original implementation that have been fixed (one outstanding: https://github.com/gravitystorm/openstreetmap-carto/pull/4549).

I haven't tested this in the full context of the openstreetmap-carto rendering stack though, just for one-time imports using osm2pgsql (v1.6.0 or higher highly recommended). I think a couple of others have been using the new flex style options of osm2pgsql in continuously updated databases and stacks though for quite some time, so maybe someone else can comment on the solidity of that with latest osm2pgsql and such complete rendering stack.

I do think switching to flex opens up a whole lot of new possibilities, and wouldn't necessarily be disruptive for the current rendering stack with the work Paul did. It should largely be a "drop-in" replacement with #4431.

pnorman:
> @pnorman #4431 might also fix this, but would this PR make sense in the meantime?

This PR would make sense, but I wouldn't bother with work on it in favour of doing other work, as I think flex is close enough, and enables much simpler methods.

dch0ph:
Just to update now that mapvik v4 is being distributed in the new LTS Ubuntu (resolute).

PDF and SVG are both now correct using the ferry routes code in this PR:
image

But the ferry routes are being rasterised in both cases. In terms of PDF, this is probably inevitable since PDF doesn't seem to support the Porter-Duff compositing operations. SVG now does support them (all, most?) and the mapnik output seems make use of them, but the ferry routes layer is still being rasterised, perhaps as a result of the `opacity: 0.99999` trick (still needed, setting the layer opacity to 1 doesn't work).

Obvious this is not an issue for raster tiles, but it is a little frustrating for print maps.

pitdicker:
It seems best to close my PRs that deal with improving vector support. I maintain my local working setup, and can no longer invest in trying to upstream it.

imagico:
Thanks for your work on this none the less.

The technique demonstrated is intriguing, but i still don't fully understand how it works and i did not have time so far to delve into the Mapnik source code for that.

My suggestions going forward:

* separate the ferry lines issue from the admin boundaries.
* for the admin boundaries a look at our strategic direction is advisable.  I would, in particular, want to look at:

#2172
#3504
#712
#5027

Personally i would be more inclined to support testing a somewhat unproven technique for the ferry lines than for the admin borders.

dch0ph:
> Personally i would be more inclined to support testing a somewhat unproven technique for the ferry lines than for the admin borders.

Agreed, but I think it also depends on how much significance we give to supporting vector output from OSM Carto, #5117. At least in my hands, testing this approach for ferry routes leads to extensive rasterisation and much larger PDF output file sizes. I suspect that this could be avoided using current mapnik functionality for [nesting layers](https://github.com/mapnik/mapnik/pull/3474), but it is hard to see layer nesting becoming accessible via `carto`.

PR #4476, opened 2021-09-20
Make water-lines-casing an attachment of water-lines

pitdicker:
The SQL query of `water-lines-casing` was a subset of the one for `water-lines`, so it is easy to turn it into an attachment to that layer. This removes one small query.

I have checked the generated Mapnik XML, the styles are in the right order now part of `water-lines`. No visual changes.

### DIFF
--- a/project.mml
+++ b/project.mml
@@ -164,26 +164,6 @@ Layer:
         ) AS icesheet_polygons
     properties:
       minzoom: 5
-  - id: water-lines-casing
-    geometry: linestring
-    <<: *extents
-    Datasource:
-      <<: *osm2pgsql
-      table: |-
-        (SELECT
-            way,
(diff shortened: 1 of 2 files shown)


jeisenbe:
Thank you, @pitdicker 
This seems like a good idea, though I have not tested it. Did you find any performance improvement with this change?

imagico (approved):
This seems to work fine and is a good idea in the current situation as we draw the casing and lines directly after each other.

As far as i understand it to have any performance benefits we would need to add `cache-features: true` to the layer properties though.

And if we at some point want to implement #3854 we might need to revert this change to get the rendering desired.  But that is a bridge to cross if and when we come to it of course.

pnorman (approved):

PR #3548, opened 2018-12-03
Change allotments color and pattern

jeisenbe:
**Fixes #3411** 

**Changes proposed in this pull request:**
- Change allotments color to garden/grass green, i.e. #cdebb0, Lch(90,32,128)
- Change pattern to near-white, light green 2x2 pixel squares on a 4x4 pixel grid
       *Pattern is #f2faeb (Lch97,8,130) approximately 75% white with 25% grass color 
- Consolidate allotments on the same lines as garden and grass for text label rendering.
- Add allotments.svg and allotments.md to describe generating this pattern

**Explanation**:
Allotment gardens are currently a terra-cotta colour, similar to farmyards, with a pattern of white dots every 8 pixels. This looks nice and the relation to farmyards is sensible, but it does not immediately suggest that allotments are a type of garden. 

This PR will change allotments to use the same shade of green as landuse=garden (and landuse=grass), while changing the pattern to be distinct from gardens and grass areas.

**Test rendering with link:**
http://openstreetmap.org/#map=16/52.1940/21.0404
Warsaw z15 master
warsaw-z15-master
z16
warsaw-z16-master

After z15
warsaw-z15-new
z16 after
warsaw-allotments-after-z16

Also see previous test renderings in issue #3411, eg:
https://github.com/gravitystorm/openstreetmap-carto/issues/3411#issuecomment-443203388 
https://github.com/gravitystorm/openstreetmap-carto/issues/3411#issuecomment-443192445
https://github.com/gravitystorm/openstreetmap-carto/issues/3411#issuecomment-442807940

PR #4480, opened 2021-09-30
Update pitch color to be less blue

### LINKED ISSUE #4479: [suggestion] golf course colors - recent change - color for putting greens should be more green
### Expected behavior
Putting greens should be brighter green as it usually has the best quality grass on a golf course.

### Actual behavior
The recent change in default colors for the default layer has a more blue-ish hue for putting greens, making them somewhat look like water hazards which is something that golfers actively try to avoid.

https://www.openstreetmap.org/#map=19/40.45731/-104.97406

jgruca:
Fixes #4479 

- Update `@pitch` from `#aae0cb` to `#88e0be`

Golf course with water hazards
https://www.openstreetmap.org/#map=15/41.5966/-93.6898

z15
1-z15
z16
1-z16
z17
1-z17

Park with pitches and pool
https://www.openstreetmap.org/#map=15/41.5731/-93.7213

z15
2-z15
z16
2-z16
z17
2-z17

School with pitches and sports center, nearby wooded area
https://www.openstreetmap.org/#map=15/41.5982/-93.7211

z15
3-z15
z16
3-z16
z17
3-z17

Comparison with the rest of the colors in this style
comparison

### DIFF
--- a/style/landcover.mss
+++ b/style/landcover.mss
@@ -63,7 +63,7 @@
 
 // --- Sports ---
 
-@pitch: #bfdfc3;           // Lch(86,18,146) also track
+@pitch: #88e0be;           // Lch(83,35,166) also track
 @track: @pitch;
 @stadium: @leisure; // also sports_centre
 @golf_course: @campsite;
--- a/style/landcover.mss
+++ b/style/landcover.mss
@@ -63,7 +63,7 @@
 
 // --- Sports ---
 
-@pitch: #aae0cb;           // Lch(85,22,168) also track
+@pitch: #bfdfc3;           // Lch(86,18,146) also track
 @track: @pitch;
 @stadium: @leisure; // also sports_centre
 @golf_course: @campsite;


imagico:
I will not oppose this change if it finds support by other maintainers but i would strongly advise against it because it would go against almost every concept of systematic color design we have followed or considered in the past.

The logical alternative would be to roll back #2363 (and it might be helpful to read up the discussion that led to that choice in #2071 - which contains some of the considerations regarding systematic color choices i consider important).  You could also have a look at the ac-style which uses `#88e0be` - although that is designed to work together with other changes, in particular a distinct color for playgrounds.

jeisenbe:
This proposed color, as noted, is almost the same as the allotments color.
Similar colors should have similar meanings. Allotments are not very
similar to sports tracks or pitches, so this proposed change is not a good
idea.

#88e0be looks interesting. I would also suggest looking at #b8dabd

Adamant36:
Any change in pitch color is going to have to account the color of the landuse below it and IMO this color makes the current color for parks stand out way to much. Especially at z17.

jeisenbe (dismissed):
We should not have 2 colors that are only barely visually different. In the unusually case with a pitch and allotments overlapping or adjacent it will be hard to see the difference. The differnence between the new pitch and current allotments is only Δ: 4.3 in Lch, which is not enough to be clearly understood.

However, we cannot simply unifiy the color with that of allotments. Similar colors should have similar meanings. Allotments are not very similar to sports tracks or pitches.

Alternatives that could be accepted would be to pick a different new color for pitches which was not too similar to another color. As mentioned above, `#88e0be` might work. I would also suggest looking at `#b8dabd`.

Another option would be to change the allotments colors. These are similar to farmland and gardens, but also sometimes related to residential areas. I could be challenging to find a better color, we have already changed it 3 times (see #2267, #3625 and #3548) - so I would recommend trying a different pitch color instead.

jgruca:
I've updated the color to `#88e0be` per your suggestion. The images in the initial comment have also been updated.

imagico (approved):
I had not noticed this had been changed to use the color from the AC-style.  This seems fine to me now, maybe a tiny bit too strong because OSM-Carto uses somewhat weaker colors overall than the AC-style.  But that is something that should be reconsidered anyway (#3647) so this is completely fine as a subtle move in that direction.

I am sorry for the long time it took to review this - it somehow slipped my attention.

PR #4506, opened 2021-12-21
Shapefiles

traitor6789:

- 

Before

After


traitor6789:
dffg

pnorman (changes requested):
We now load the shapefiles into the DB. See [INSTALL.md](https://github.com/gravitystorm/openstreetmap-carto/blob/master/INSTALL.md#scripted-download).

PR #4524, opened 2022-03-15
Render landuse=education

### LINKED ISSUE #774: Add rendering for landuse=education
The following issue has been moved over from [trac](https://trac.openstreetmap.org/ticket/3608):

amenity=school should be used only for the actual facilities. but fellow mappers use it as a substitute for landuse, because it renders nicely in a yellowish color. the problem is, that search results then show two schools (one for the building and one for the area), because there there is not enough data to differentiate between the landuse and the actual school. it gets even worse, when different schools share one estate.

therefore i suggest that landuse=education should be rendered in the same color like amenity=school on areas to give mappers an alternative for defining/coloring the landuse and at the same time solving the problem with too many search results.

jdhoek:
Render `landuse=education` the same way as `amenity=school` etc.; this is in line with the stated goals of this tag, and actively stimulates mappers to stop abusing `amenity-school` to force rendering of shared grounds.

Fixes #774

# Before

## 19

Screenshot from 2022-03-15 19-27-45

## 17

Screenshot from 2022-03-15 19-27-39

# After

## 19

Screenshot from 2022-03-15 19-21-30

## 17

Screenshot from 2022-03-15 19-21-39

### DIFF
--- a/project.mml
+++ b/project.mml
@@ -101,7 +101,7 @@ Layer:
               ('landuse_' || (CASE WHEN landuse IN ('quarry', 'vineyard', 'orchard', 'cemetery', 'residential', 'garages', 'meadow', 'grass',
                                                     'allotments', 'forest', 'farmyard', 'farmland', 'greenhouse_horticulture',
                                                     'recreation_ground', 'village_green', 'retail', 'industrial', 'railway', 'commercial',
-                                                    'brownfield', 'landfill', 'salt_pond', 'construction', 'plant_nursery', 'religious') THEN landuse END)) AS landuse,
+                                                    'brownfield', 'landfill', 'salt_pond', 'construction', 'plant_nursery', 'religious', 'education') THEN landuse END)) AS landuse,
               ('shop_' || (CASE WHEN shop IN ('mall') AND (tags->'location' NOT IN ('underground') OR (tags->'location') IS NULL) THEN shop END)) AS shop,
               ('leisure_' || (CASE WHEN leisure IN ('swimming_pool', 'playground', 'park', 'recreation_ground', 'garden',
                                                     'golf_course', 'miniature_golf', 'sports_centre', 'stadium', 'pitch', 'ice_rink',
@@ -1524,7 +1524,7 @@ Layer:
                 'landuse_' || CASE WHEN landuse IN ('reservoir', 'basin', 'recreation_ground', 'village_green', 'quarry', 'vineyard', 'orchard', 'cemetery',
                                                     'residential', 'garages', 'meadow', 'grass', 'allotments', 'forest', 'farmyard', 'farmland',
                                                     'greenhouse_horticulture', 'retail', 'industrial', 'railway', 'commercial', 'brownfield', 'landfill',
-                                                    'construction', 'salt_pond', 'military', 'plant_nursery') THEN landuse END,
+                                                    'construction', 'salt_pond', 'military', 'plant_nursery', 'education') THEN landuse END,
                 'natural_' || CASE WHEN "natural" IN ('peak', 'volcano', 'saddle', 'cave_entrance') AND way_area IS NULL THEN "natural" END,
                 'natural_' || CASE WHEN "natural" IN ('wood', 'water', 'mud', 'wetland', 'bay', 'spring', 'scree', 'shingle', 'bare_rock', 'sand', 'heath',
                                                       'grassland', 'scrub', 'beach', 'glacier', 'tree', 'strait', 'cape')
--- a/style/amenity-points.mss
+++ b/style/amenity-points.mss
@@ -2175,6 +2175,7 @@
   [feature = 'amenity_school'],
   [feature = 'amenity_college'],
   [feature = 'amenity_university'],
+  [feature = 'landuse_education'],
   [feature = 'landuse_religious'],
   [feature = 'natural_heath'],
   [feature = 'natural_scrub'],
@@ -2286,7 +2287,8 @@
       [feature = 'amenity_kindergarten'],
       [feature = 'amenity_school'],
       [feature = 'amenity_college'],
-      [feature = 'amenity_university'] {
+      [feature = 'amenity_university'],
+      [feature = 'landuse_education'] {
         text-fill: darken(@societal_amenities, 80%);
       }
       [feature = 'landuse_religious'] {
--- a/style/landcover.mss
+++ b/style/landcover.mss
@@ -582,7 +582,8 @@
   [feature = 'amenity_kindergarten'],
   [feature = 'amenity_community_centre'],
   [feature = 'amenity_social_facility'],
-  [feature = 'amenity_arts_centre'] {
+  [feature = 'amenity_arts_centre'],
+  [feature = 'landuse_education'] {
     [zoom >= 10] {
       polygon-fill: @built-up-lowzoom;
       [way_pixels >= 4]  { polygon-gamma: 0.75; }


imagico (changes requested):
I have looked at the use of `landuse=education`.  The tag is used predominantly as an umbrella tag to tag what is otherwise tagged as `amenity=school`, `amenity=kindergarten`, `amenity=college` etc. but with less specific meaning.  This umbrella tag use consists mainly of:

* double tagging the grounds of a single education facility with `landuse=education` and  `amenity=*`
* tagging the grounds of a single education facility with `landuse=education` with a separate intersecting polygon tagged `amenity=*`
* tagging the grounds of a single education facility with `landuse=education` with a node or a building inside tagged `amenity=*`
* tagging the grounds of a single education facility with `landuse=education` with nothing specific inside - this is often coming from some sort of import.

Conservatively estimated these uses account for roughly 2/3 of the uses of the tag, possibly significantly more.

The use for tagging the union of overlapping facilities for which different `amenity=*` values apply - like an `amenity=kindergarten` and an `amenity=school` is clearly a minority use of the tag.  And consistency of that minority use is furthermore poor - there are a lot of weird constructions with that.  Like:

https://www.openstreetmap.org/relation/1580741
https://www.openstreetmap.org/way/959800915

The dominant mapping of such cases is quite clearly use of `amenity=*` in the way all landcover tags in OSM tend to be used - by tagging the predominant meaning.  There is clearly no consensus among mappers that overlapping features tagged with educational `amenity=*` is disallowed.

In addition there is also some use for what is essentially land ownership mapping, i.e. using the tag to map land that is the property of a school - without necessarily being used for education purposes.

Long story short::  I don't think rendering this tag as a synonym for the much more specific and much more broadly and consistently used educational `amenity=*` tags would be compatible with the goals of this style.

If we'd seriously want to consider rendering this tag (in case its use improves in consistency and in semantic delineation towards the competing and much more widely used `amenity=*` tags) that would IMO have to:

* be in a styling distinct from the educational `amenity=*` features so it is clear that it is not a synonym for them.
* be in terms of prominence of rendering and in particular labeling clearly semantically subordinate to educational `amenity=*`.  Our current `way_area` based polygon labeling is exactly the opposite of that.

boothym:
>Long story short:: I don't think rendering this tag as a synonym for the much more specific and much more broadly and consistently used educational amenity=* tags would be compatible with the goals of this style.

For a shared campus with two schools already individually mapped as nodes/buildings, how should its campus area be tagged if not with landuse=education?

jeisenbe:
Rendering `landuse=education` risks losing the established `amenity=school` tagging on areas. 

I am concerned that some mappers would assume that only `landuse=education` was needed and that `amenity=school` could be removed, were we to render the two tags identically.

There has always been a vocal minority of mappers who have a dislike for the broad key `amenity=` and would seek to replace it with alternatives - see for example promotion of `healthcare=hospital` instead of `amenity=hospital`.

jeisenbe:
The first place I randomly checked confirms my concerns above:

https://www.openstreetmap.org/way/969746864/history
https://www.openstreetmap.org/way/969746874/history
https://www.openstreetmap.org/way/862525322/history

In one changeset 9 months ago, a user in the Netherlands added `landuse=education` to each of these areas, and removed the `name=`, instead the name is now on a separate node for each of the 3 schools. They are all individual schools, not shared campuses. 

I am concerned that were we to render `landuse=education`, the `amenity=school` tag would be removed entirely from these areas and consigned to the node instead, which would be a loss of information in comparison to mapping the school grounds as recommended at `amenity=school`, while also being more complicated since there are now 2 database objects instead of 1. This is not just one user, just a few kilometers away the next village has https://www.openstreetmap.org/way/1011247217 - which is also now a closed way with a single `amenity=school` note inside. This one has the tag `fixme = Remove amenity=school when landuse=education is rendered.`

We do not want to force all database users to update their schema to use landuse=education for rendering without a clear benefit, nor should we encourage mappers to spend time changing `amenity=school/kindergarten/etc` areas to `landuse=education` when not necessary.

jeisenbe:
Here in Oregon the tag `landuse=education` has only been used a few times.

5 months ago 3 existing `amenity=school` areas had `landuse=education` tags added, in the same changest, eg. https://www.openstreetmap.org/way/398642092 - This was harmless, but adds no new information.

But next door at https://www.openstreetmap.org/way/1020441106 - a `landuse=education` area was added around a single exiting `amenity=school` node, and also `amenity=school` was added to the building, duplicating the feature. At this time the mistake is clear due to the lack of area rendering.

https://www.openstreetmap.org/way/1017749189 is "Springfield Schools Maintenance” - it appears to be a school bus parking lot and maintenance facility, which appears to not be the intended use of the tag, and should not be rendered like a school.

Locally https://www.openstreetmap.org/way/131079290 and https://www.openstreetmap.org/way/1030598824/ are the only 2 examples which closely fit the proposal - with a couple different schools within the same grounds.

jdhoek:
> In one changeset 9 months ago, a user in the Netherlands added landuse=education to each of these areas, and removed the name=, instead the name is now on a separate node for each of the 3 schools. They are all individual schools, not shared campuses.

`landuse=education` isn't rendered, so that isn't the cause of that tagging style. Mappers make mistakes. Documentation and guidance address that.

That mapper would have done the same with or without rendering: the cause here is that some Dutch mappers still follow an old informal guideline that is valid for shops and such, but not for large amenities with clear area boundaries. That is, the notion that all POI's should only be mapped on the imported and existing address nodes. That is bullshit when it comes to things like hospitals, and it is pointless if the area is known. This is not something that is broadly considered correct either. I will address those errors with that mapper.

That example has nothing to with _rendering_ `landuse=education`, and you should address it at a more appropriate venue, like the Tagging ML, the wiki, or the relevant community channel (the Dutch forum in this case).

boothym:
The user that you highlight also moved details of a sports centre from the building outline to a node in that changeset. Perhaps they wanted the school name rendered differently, or the sports centre name rendered in green? (as it isn't rendered green on a building).

https://overpass-api.de/achavi/?changeset=110602651

> while also being more complicated since there are now 2 database objects instead of 1.

As mentioned in previous comments, an extra database object is already being used in many cases to map the entire shared campus containing two schools - but is something that rendering this tag would help avoid.

jdhoek:
> The user that you highlight also moved details of a sports centre from the building outline to a node in that changeset. Perhaps they wanted the school name rendered differently, or the sports centre name rendered in green?

Yeah, that's the same misguided notion that all POI's should only be tagged on address nodes. It has nothing to do with this tag at all. I'm addressing that with this mapper.

jeisenbe (changes requested):
I recommend rejecting the rendering of landuse=education at this time. This can be reconsidered in the future if use changes. But at the moment it would not be beneficial to show this in most cases, and in many cases it could be harmful by providing incorrect feedback to mappers, for example when an `amenity=school` area is missing.

HolgerJeromin:
> in many cases it could be harmful by providing incorrect feedback to mappers, for example when an amenity=school area is missing.

We could drop `name` rendering with missing `amenity` tag.

pnorman (changes requested):
I agree with https://github.com/gravitystorm/openstreetmap-carto/pull/4524#pullrequestreview-996173739 above

cicku:
I used this tag for vocational school like Job Corps, not sure what tag can substitute it.

vincentvd1:
Given the explosive growth of landuse=education, can rendering be re-considered?

https://taginfo.openstreetmap.org/tags/landuse=education#chronology

raspbeguy:
Also using `landuse=education` areas + `amenity=school` single nodes is the official way of tagging multiple school on one signe ground, stated in the wiki. This is sad that the rendering conflicts with official instructions in my opinion.

teemka:
As of writing this there are 115 objects with "wiki official" fixme tag relating to this PR:
https://taginfo.openstreetmap.org/tags/?key=fixme&value=Remove+amenity%3Dschool+when+landuse%3Deducation+is+rendered.

imagico:
Discussion on if and how to render `landuse=education` should go to #774, not here on a closed PR.

PR #4535, opened 2022-04-21
Fix shop=car_repair label/icon colour mismatch (purple)

### LINKED ISSUE #2658: shop=car_repair color inconsistency dot vs icon
Hi, just noticed this: For shop=car_repair, as long as it's displayed as just a dot, the color is violet, but when the icon shows up, it is brown. 
It seems this is the only shop=* value that is rendered with a brown icon. 
It would look better if either the dot for car repairs was brown, or even if there was no dot for this (because we do not use any brown dots atm, IIRC)

Related PRs #2589 and #1778, also discussion #1783 and closed PR #1868

jdhoek:
The colour of the icon (purple) and the label (brown) don't match.

This commit makes the label purple to match most other `shop=*` tags.

Fixes #2658

-----

_This is one of two ways to solve this. The other colour choice is shown in #4536. These two PR's are mutually exclusive._

_Either one is probably better than:_ nope

-----

# Before

17-now

18-now

# After

17-purple

18-purple

### DIFF
--- a/style/amenity-points.mss
+++ b/style/amenity-points.mss
@@ -1024,7 +1024,6 @@
 
     [shop = 'car_repair'][zoom >= 18] {
       marker-file: url('symbols/shop/car_repair.svg');
-      marker-fill: @amenity-brown;
     }
 
     [shop = 'dairy'][zoom >= 18] {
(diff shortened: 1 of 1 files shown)


imagico (approved):
This looks fine and renders as expected.  Nice to see this being fixed.

I don't have any strong preference regarding the color, both choices seem in principle fine with me.  So far we don't have a very consistent scheme of colors with symbols.  I would suggest to:

* think about a practically verifiable criterion to determine if some feature is to be rendered in shop color or a different color.
* Then apply it to the choice here and
* document it so others can follow the same principle with other changes

And the criterion should of course not be the key used by the tag - that does not follow any consistent rules.

jeisenbe:
I approve this change (I have not tested the rendering but the code change looks fine)

PR #4536, opened 2022-04-21
Fix shop=car_repair label/icon colour mismatch (brown)

### LINKED ISSUE #2658: shop=car_repair color inconsistency dot vs icon

jdhoek:
The colour of the icon (purple) and the label (brown) don't match.

This commit makes the icon brown in line with other service providing
amenities.

Fixes #2658

-----

_This is one of two ways to solve this. The other colour choice is shown in #4535. These two PR's are mutually exclusive._

_Either one is probably better than:_ nope

-----

# Before

17-now

18-now

# After

17-brown

18-brown

### DIFF
--- a/style/amenity-points.mss
+++ b/style/amenity-points.mss
@@ -847,7 +847,7 @@
   }
 
   [feature = 'shop'] {
-    [shop != 'mall'][shop != 'massage'][zoom >= 17],
+    [shop != 'mall'][shop != 'massage'][shop != 'car_repair'][zoom >= 17],
     [shop = 'supermarket'][zoom >= 16],
     [shop = 'department_store'][zoom >= 16] {
       marker-clip: false;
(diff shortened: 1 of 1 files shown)


imagico (changes requested):
This in principle works fine as well, however it introduces brown dot rendering at z17 - which would be a novelty for this style.  Unlike shops where almost all variants start at the same zoom level brown symbols start at very different zoom levels so introducing a dot only rendering at the starting zoom level could seem a bit weird.  Still this could be discussed as a new design idea, however if you want to do so i would suggest to separate that from the bug fix.

As written in #4535 i am not partial to either of the color choices but would like to see a consistent criterion being developed, applied and documented for that.

strongwillow:
> however it introduces brown dot rendering at z17 - which would be a novelty for this style.

It does not introduce a dot at z17. A purple dot is rendered in the current rendering. See https://www.openstreetmap.org/#map=18/51.42782/5.44144

Car repair shops belong to amenity shop types like `car_wash`. I think it's correct to show them at z17.

jeisenbe (changes requested):
I believe shop icons should be consistent in one color, so I recommend closing this PR without merging

pnorman:
I also agree on the purple colour being the better option. Closing in favour of that PR.

PR #4542, opened 2022-05-06
Render boundary=protected_area+protect_class=22 (cultural protection)

Blijbol:
This adds rendering for boundary=protected_area+protect_class=22, which is used for culturally-protected areas.

It reuses the existing rendering of boundary=aboriginal_lands. Both can be thematically categorised as socially-protected area, so it makes sense to render them identically. They are rendered the same as all variants of naturally-protected areas, but coloured in yellow instead of green.

In the past, this same rendering has also been used for protect_class=24 (politically-protected areas), before it was removed in #4113 to encourage use of clearer and more specific tags like boundary=aboriginal_lands. However, this issue does not apply to protect_class=22, as there are no better alternative tags available to indicate culturally-protected areas. Although the wiki suggests to add historic=* and/or heritage=* to the area, neither actually defines that the reason for its protection is cultural.

Code disclaimer: I have zero experience with this codebase and have no test setup to test the code. If you agree with the proposed rendering change, please check and test the code carefully.

### DIFF
--- a/project.mml
+++ b/project.mml
@@ -1166,7 +1166,7 @@ Layer:
           FROM planet_osm_polygon
           WHERE (boundary IN ('aboriginal_lands', 'national_park')
                  OR leisure = 'nature_reserve'
-                 OR (boundary = 'protected_area' AND tags->'protect_class' IN ('1','1a','1b','2','3','4','5','6')))
+                 OR (boundary = 'protected_area' AND tags->'protect_class' IN ('1','1a','1b','2','3','4','5','6','22')))
             AND building IS NULL
             AND way_area > 1*!pixel_width!::real*!pixel_height!::real
         ) AS protected_areas
@@ -1537,8 +1537,9 @@ Layer:
                 'highway_' || CASE WHEN highway IN ('services', 'rest_area', 'bus_stop', 'elevator', 'traffic_signals') THEN highway END,
                 'highway_'|| CASE WHEN tags @> 'ford=>yes' OR tags @> 'ford=>stepping_stones' AND way_area IS NULL THEN 'ford' END,
                 'boundary_' || CASE WHEN boundary IN ('aboriginal_lands', 'national_park')
-                                          OR (boundary = 'protected_area' AND tags->'protect_class' IN ('1','1a','1b','2','3','4','5','6'))
                                           THEN boundary END,
+                'boundary_protected_area_' || CASE WHEN boundary = 'protected_area' AND tags->'protect_class' IN ('1','1a','1b','2','3','4','5','6','22')
+                                          THEN tags->'protect_class' END,
                 'leisure_' || CASE WHEN leisure IN ('nature_reserve') THEN leisure END,
                 'tourism_' || CASE WHEN tourism IN ('information') AND tags->'information' IN ('audioguide', 'board', 'guidepost', 'office', 'map', 'tactile_map', 'terminal') THEN tourism END,
                 'office' || CASE WHEN tags->'office' IN ('no', 'vacant', 'closed', 'disused', 'empty') OR (tags->'office') IS NULL THEN NULL ELSE '' END,
@@ -1991,7 +1992,7 @@ Layer:
                                                     'water', 'bay', 'strait') THEN "natural" END,
               'place_' || CASE WHEN place IN ('island') THEN place END,
               'boundary_' || CASE WHEN boundary IN ('aboriginal_lands', 'national_park')
-                                       OR (boundary = 'protected_area' AND tags->'protect_class' IN ('1','1a','1b','2','3','4','5','6'))
+                                       OR (boundary = 'protected_area' AND tags->'protect_class' IN ('1','1a','1b','2','3','4','5','6','22'))
                                        THEN boundary END,
               'leisure_' || CASE WHEN leisure IN ('nature_reserve') THEN leisure END
             ) AS feature,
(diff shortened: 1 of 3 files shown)


ZeLonewolf:
In the United States, we've been working to remove the invented (i.e. not related to IUCN categories) values of `protect_class`, including value 22.  Most of the cases that remain are from legacy imports that we haven't come up with better replacement values for.  Adding rendering for `protect_class=22` would harm this community effort to replace cryptic numeric values with better semantics.

Adamant36:
> In the United States, we've been working to remove the invented (i.e. not related to IUCN categories) values of protect_class, including value 22.

If you don't mind me asking, what are the values of protect_class being replaced with? It's not like the alternatives, one of which I assume is leisure=park, are any less "invented" or cryptic.

kennykb:
On Tue, May 10, 2022 at 7:19 PM Adamant36 ***@***.***> wrote:

> In the United States, we've been working to remove the invented (i.e. not
> related to IUCN categories) values of protect_class, including value 22.
>
> If you don't mind me asking, what are the values of protect_class being
> replaced with? It's not like the alternatives, one of which I assume is
> leisure=park, are any less "invented" or cryptic.
>

As far as I know, we've not got around to 21 or 22.  There is no good
replacement yet. Which doesn't keep Brian from bugging me to replace them.
:)

-- 
73 de ke9tv/2, Kevin

ZeLonewolf:
It depends on the property in question, but typically something under `historic` or `heritage` is a more sane tagging.

ZeLonewolf:
For 21, `boundary=water_protection_area` and `landuse=reservoir_watershed` are emerging candidates, again depending on the specific nature of the property, because there's a wide variety, and often they're really just `leisure=nature_reserve` or one of the existing rendering values of `protect_class=*`.  But building consensus takes time.  In any case, `protect_class=22` does not a level of community support that would justify adding it to rendering in my opinion.

kennykb:
On Wed, May 11, 2022 at 11:23 PM Brian Sperlongano ***@***.***>
wrote:

> For 21, boundary=water_protection_area and landuse=reservoir_watershed
> are emerging candidates, again depending on the specific nature of the
> property, because there's a wide variety, and often they're really just
> leisure=nature_reserve or one of the existing rendering values of
> protect_class=*.
>

That's 12 (resource protection), not 21 (community values).  (Yes, I agree
with you that the non-mnemonic nature of the numeric codes does us a
disservice).

> But building consensus takes time. In any case, protect_class=22 does not
> a level of community support that would justify adding it to rendering in
> my opinion.
>
I'm afraid that I'm largely the perpetrator of all three of these tags in
the US. I adopted them because nobody had a better alternative to describe
the objects to which they attach.  I've been convinced to retag
`protect_class=12` according to Brian's suggestion. None of the available
alternatives for 21 (which I used for State Park, State Multiple-Use Area,
State Unique Area, County Park, Town Park, and similar objects) and 22
(which I used on State Historic Sites for which `museum` didn't seem quite
correct).  I can locate readily the objects that I've tagged thus, and if a
consensus emerges, I'll be happy to retag them (or simply step back and let
Brian do it).


-- 
73 de ke9tv/2, Kevin

Adamant36:
> For 21, boundary=water_protection_area and landuse=reservoir_watershed are emerging candidates

That's kind of a weird take when the only usage landuse=reservoir_watershed has is from a couple of mass-edits in 2010 and it's been on a steady decline ever since. Same goes for boundary=water_protection_area and it's 213 uses. 

> I'll be happy to retag them (or simply step back and let
Brian do it).

What's with all this deferring to Brian stuff about?

kennykb:
On Thu, May 12, 2022 at 12:19 AM Adamant36 ***@***.***> wrote:

> I'll be happy to retag them (or simply step back and let Brian do it).
>
> What's with all this deferring to Brian stuff about?
>
Just that I'm tired of this interminable argument.  What I've done in NY
might be the wrong thing, but it's a _consistent_ wrong thing.  Which means
that it's relatively easy to change mechanically into something else.

-- 
73 de ke9tv/2, Kevin

Adamant36:
> Just that I'm tired of this interminable argument.  

Makes sense. Can't say I blame you considering how long and needlessly contentious this whole thing has been.

> Which means that it's relatively easy to change mechanically into something else.

Since we're on the subject do either of you happen to know what the mass edit that gutted 30,000 uses of the protect_class tag in December was about?

ZeLonewolf:
> Since we're on the subject do either of you happen to know what the mass edit that gutted 30,000 uses of the protect_class tag in December was about?

There was [no such edit](https://taginfo.openstreetmap.org/keys/protect_class#chronology) and this has now gone off-topic for this PR so I'm going to unsubscribe :)

imagico:
Please keep the discussion on topic, especially on a PR.

@Blijbol - thanks for the PR.  I don't think this suggestion is going to find consensus support at this time though.  At a quick look there does not seem to be much consistency in use of `protect_class=22` and i don't even see how the tag is meant to be used exactly based on verifiable criteria.  Anyone presenting evidence that this observation is wrong is welcome to make their case but from how it looks to me right now i would suggest to reject this PR.

jeisenbe:
@Blijbol thank you for your contribution. Unfortunately I do not support merging the PR right now.

I do not think `boundary=protected_area` + `protect_class=22` should be rendered the same as `boundary=aboriginal_lands`

In the past some similar areas of land which were areas for indigenous peoples were tagged `protect_class=24` but the tag `boundary=aboriginal_lands` is now more widely used, and neither is related to `protect_class=22`

 `protect_class=22` is said to be for "protection of sites with special architecture or historic interest, designed and created intentionally by people: urban layout, buildings, landscape, gardens, art-goods” - this is on https://wiki.openstreetmap.org/wiki/Key:protect_class but there is no specific page for protect_class=22 in particular. Taginfo shows that usage is limited to a few regions in a few countries: https://taginfo.openstreetmap.org/tags/protect_class=22#map

Most historic sites are tagged with other methods such as `historic=` or `heritage=` which has specific values. In particular `heritage=` is much more common, and `historic=` has more specific values. I do not think there is a clear consensus to use  `protect_class=22` in a clear and consistent way, so we should not render it at this time.

pnorman:
I also agree with the above reasons for declining this PR.

Issue #3961, opened 2019-10-31
Render highway=pedestrian same as highway=living_street


jeisenbe:
As mentioned in https://github.com/gravitystorm/openstreetmap-carto/issues/3849#issue-484899355 ("Render highway=pedestrian and highway=living_street same as highway=residential on z13") there is not a strong difference between the meaning of `highway=pedestrian` and `highway=living_street` - both are pedestrian-priority areas with limits on motor vehicle access and speed, but both almost always allow motor vehicle access at least at certain times of the day.

The current `highway=pedestrian` rendering is quite dark, with the same lightness as `landuse=residential`, `=retail`, `=commercial` and `=industrial`, so there is low contrast with the surrounding polygon fill in most places.

On the other hand, there are problems with the current `highway=living_street` fill, for example it is identical to the current `parking` area fill, which is a problem for areas. At z13 the color is hard to see against residential areas, according to https://github.com/gravitystorm/openstreetmap-carto/issues/3849#issue-484899355 

By combining the two features into one rendering, we would have one less different road color, which will make future changes and maintenance of the style easier. And we could adjust the color to compensate for the 3 problems mentioned above.

Rendering `highway=pedestrian` like `highway=living_street` would also mean showing these features at z13. Right now it is strange that cycleways are shown but not large pedestrian streets at z13.

Current rendering of Luxembourg city, z13 (no pedestrian roads, but living streets shown in southeast part of city centre):
![z13-luxembourg-current](https://user-images.githubusercontent.com/42757252/63674631-e8fecf00-c821-11e9-9a71-bff4c29eb9c2.png)

z14 Luxembourg current
![z14-luxembourg-city-current](https://user-images.githubusercontent.com/42757252/63674632-e9976580-c821-11e9-96f7-00bac7fc8d46.png)

polarbearing:
> both almost always allow motor vehicle access at least at certain times of the day

I disagree. For hw=pedestrian that might be true if there is delivery access in a shopping zone, but there are other examples where they are indeed pedestrian only.

ClarkstonCorrect:
I also thought that living street rendering doesn't stand out enough from landuse=residential.

jeisenbe:
https://github.com/gravitystorm/openstreetmap-carto/pull/4555#issue-1245520424 - former maintainer @matkoniecz recommended to close this issue "as not worth doing" or perhaps "not improving things" - please explain?

jeisenbe:
[...]

Issue #3958, opened 2019-10-29
Render archeological site polygons as tourism boundary


StyXman:
Archeological sites are usually tourist attractions. We already render them as amenity points, but some of them are frankly quite big, and mapped as polygons. Just render the same polygon effect for zoos and theme parks for those polygons.

matkoniecz:
repeat of #732 but not a duplicate, as provides a specific rendering idea that makes sense and has decent chance to be an overall improvement

imagico:
I suggest to close this - if an archeological site has a physically manifested verifiable boundary then that should be mapped and rendered as such (like `barrier=*`) same applies if ruins are visible and can as such be mapped.  But rendering legal bounds of sites as such (which practically is often simply guesswork or abstract wrapper polygons) is something we should reduce rather than extend IMO.

Indicating visually in some form which mapped ruins belong to a certain (named) archeological site would be something interesting to consider i think - but it would require (a) an established tagging for the ruins of an archeological site, (b) an established way to map which ruins belong to a site and (c) probably quite a bit of work to develop an efficient and visually suitable way to depict that.

Issue #3849, opened 2019-08-25
Render highway=pedestrian and highway=living_street same as highway=residential on z13


Prince-Kassad:
Partially reverts #3467

Currently, highway=pedestrian is not rendered at all on z13, while highway=living_street is rendered just like in higher zoom levels.

Both are problems on their own: Cities that make heavy use of pedestrian zones look really funny on z13 because only the residential roads render and the pedestrian ones don't. They appear to have huge gaps of nothingness in their historic city centers.

See Barcelona: https://www.openstreetmap.org/#map=13/41.4047/2.1827
Zurich: https://www.openstreetmap.org/#map=13/47.3734/8.5621
Basel: https://www.openstreetmap.org/#map=13/47.5667/7.6446

As for highway=living_street, it's more or less impossible to make out against the landuse=residential gray on z13, which is also a problem in cities that make heavy use of living streets.

See Madrid: https://www.openstreetmap.org/#map=13/40.4317/-3.6813
The Hague (district De Bras): https://www.openstreetmap.org/#map=13/52.0478/4.4201
Zinnowitz (island of Usedom): https://www.openstreetmap.org/#map=13/54.0879/13.9499

If today were 2006, I would not have supported a highway=living_street and instead have suggested tagging the living_street property as a separate tag i. e. living_street=yes, which would have solved the problem with service roads connecting to living streets looking super ugly as well. But 13 years later it's almost impossible to change a long established tagging practice, so we have to make do with this.

Anyway, on this zoom level, it's not important to distinguish whether the road is accessible to cars or not, as all you'd ever want to do at z13 is get a general idea of the city's road network, which doesn't require differentiation between residential and living_street or pedestrian. We already generalize for landuse on lower zoom levels, so I don't see an issue with generalizing for roads too, at least for z13.

jeisenbe:
Residential areas are lightness 89 (in Lch), living_street fill is L 94, and residential is 100; they are all close to neutral gray, so lightness is the main difference. Living streets on commercial or retail landuse are visible enough.

Current rendering of Luxembourg city, z13 (no pedestrian roads, but living streets shown in southeast part of city centre):
![z13-luxembourg-current](https://user-images.githubusercontent.com/42757252/63674631-e8fecf00-c821-11e9-9a71-bff4c29eb9c2.png)

z14 current
[...]

Issue #3782, opened 2019-05-18
Propose new railway station logo


wodndb:
Bus stop logo looks like bus and taxi logo looks like taxi.

However railway station logo is not.

So I propose new railway station logo like below. (I draw these using Power Point)

![train sample](https://user-images.githubusercontent.com/8624058/57970603-34514800-79be-11e9-8be6-85583ac99c71.png)

imagico:
Thanks for your suggestion.

At the moment railway stations and halts are not rendered with a pictorial symbol but with a square marker the meaning of which is meant to derive from context.  This is rendered starting at z12 where a pictorial symbol in quite definitely not an option.  If it makes sense to use a pictorial symbol at higher zoom level would need to be tested.  Currently the same marker is used for railway and tram stops so this would also need to be considered.

In general symbols need to work at native resolution (typically 14 pixel in size) and therefore should be tested in that form.  You might want to try out using Inkscape which is the typical tool used for symbol design here.

wodndb:
@imagico I try to use inkscape for symbol design. Thanks.

wodndb:
How can I test symbol what I designed?

jeisenbe:
@wobndb - there are several ways you can test this new symbol properly.

A) The easy way: make a few different options for symbols to test, and then create a Zip archive. This can be directly uploaded here. You could ask one of the other contributors, like @Adamant36 or me, to make some test renderings for you. 

However, this won't let you try different options directly, and it may take a while to get done. 

B) The best way, although it requires the most time and work, is to create test renderings with Kosmtik, based on current Openstreetmap data from several different representantive places. This will also allow you to make a [Pull Request](https://help.github.com/articles/about-pull-requests/) here in this Github repository to actually change the code.

To do this, you would need to download data onto your computer and set up a test rendering server. 

If you use Linux you can do this directly by following the step-by-step instructions in [INSTALL.MD](https://github.com/gravitystorm/openstreetmap-carto/blob/master/INSTALL.md)

For Mac and PC there is a [Docker container option]((https://github.com/gravitystorm/openstreetmap-carto/blob/master/DOCKER.md)). This is what I use, on my 2012 Macbook.

For this option, I'm pretending that you know nothing about git or the command line, because this was the state of my knowledge a few months ago.

1) download [Docker](https://www.docker.com/get-docker) and install it. 

2) Next, [create a "fork"](https://help.github.com/articles/fork-a-repo/) of this repository. 

[...]

PR #4555, opened 2022-05-23
document growing issue count

matkoniecz:
BTW, I think I would wontfix #2958 #2986 #2704 #3692 #3849 #3491 #3958 #3264 #3961 #3782 as not worth doing, not improving things or outright impossible


pnorman (approved):
Accurate, but I'd like to work out a path to getting it back under 300.

PR #4557, opened 2022-05-29
Render disc golf holes and leisure=disc_golf_course

### LINKED ISSUE #3766: Render some disc_golf features (similar to analagous golf=* features)
Currently, no disc_golf features are rendered, but I think they could be added to the map similarly to golf features, probably using the same code but adding a query for `disc_golf=*`.

At this time, OSM Carto shows `golf=tee`, but not `golf=hole` or `golf=pin`. `disc_golf=tee` could be rendered the same as `golf=tee` (light green rectangle). Ideally, then, there could be some rendering of `disc_golf=basket`, as this is a fixed point, unlike the `golf=pin` which moves around the green. The basket could be a new point icon, or perhaps a leisure-green dot with basket number?

Finally, we have icons for `leisure=golf_course` and `leisure=miniature_golf`. Could there be a `leisure=disc_golf_course` icon for polygons or relations of features?

There is not a huge number of any single `disc_golf=*` feature, but it looks like there are about 1450 holes (in 380 courses) of disc golf out there: https://taginfo.openstreetmap.org/keys/disc_golf#values

jdmonin:
Fixes #3766, now that iD has presets and the tags are more widely used and standardized.

- Render disc golf course holes, and `leisure=disc_golf_course` areas, in the same style as standard golf
- Made an icon to use for course POIs and baskets (end of each hole)

Disc golf courses are often part of a park, sometimes in the woods, not always with the best visibility when you're out on the course.

OSM now has [ten thousand](https://taginfo.openstreetmap.org/keys/disc_golf) `disc_golf` features and [over a thousand](https://taginfo.openstreetmap.org/tags/leisure=disc_golf_course) `leisure=disc_golf_course`, and this rendering can help people find and enjoy them. id [recently added](https://github.com/openstreetmap/id-tagging-schema/blob/main/CHANGELOG.md#330) presets for disc golf as well.

POI for a course within a park: https://www.openstreetmap.org/#map=17/42.87260/-76.97040

Geneva, NY

Area for a course: https://www.openstreetmap.org/#map=18/43.04604/-77.61741

Henrietta, NY

A course with mapped holes, from tees to baskets: https://www.openstreetmap.org/#map=16/43.1520/-76.1828 

z16:

Clay, NY z16

z17:
Clay, NY z17

z18:
Clay, NY z18


imagico:
Thank you for the pull request.

#3766 has been closed with the decision not to render that feature.  If you think the situation has changed in substance since then i suggest to make a case there for reopening.  Also be aware of #844 for the long standing idea of differentiated rendering of various kinds of sport infrastructure by type of sport in a consistent design (sport=disc_golf seems to have much more uses than leisure=disk_golf_course).

Independent of that i am probably not the right person to review this change because i have never heard of "disc golf" and am not familiar with either the sport or its social context.  To me the symbol you use does not suggest anything, especially nothing sports related.  With some imagination it could be read to look a bit like a trident.

jdmonin:
Thank you for your quick response!

It's been 2 years since #3766 was closed, and it does look like lot more usage nowadays; I think it's a useful addition. I'll comment on that issue. Will that comment be noticed enough to reopen discussion?

The symbol is a simplified disc golf basket, as seen at https://wiki.openstreetmap.org/wiki/Tag:leisure%3Ddisc_golf_course , and would be recognizable by a player.

jdmonin:
And I'd be fine with also mapping `sport=disc_golf` POIs/areas.
I see iD's newly added disc golf course preset adds both `sport=disc_golf` and `leisure=disc_golf_course`

pnorman:
I can't do a proper review right now but a few thoughts

- I support the idea of rendering disk golf features, and similar to golf features is the natural way. Disk golf courses tend to be in areas not used by other uses, so there are likely to be few symbol collisions
- I'm not on my normal monitor so I can't tell for sure, but it looks like parts of the image that should be pixel aligned aren't
- I'm not convinced that the symbol will be obvious to those not familiar with disk golf as a sport. I could tell what it is instantly, but to someone with no awareness of disk golf, is it obvious that it's some kind of sport icon? The same is true for the golf icon, but the number of people who will associate the golf symbols with golf is much greater because golf is a much better known sort.

imagico:
> Disk golf courses tend to be in areas not used by other uses

Please note that in https://github.com/gravitystorm/openstreetmap-carto/issues/3766#issuecomment-1146617348 i indicated that almost all the examples i could find at a first quick look indicate otherwise.

jdmonin:
@pnorman has a good point about overlap: The "holes" of a park's disc golf course are parts of the park that aren't otherwise used, like the woods or remote grassy areas. The area covered by the _entire course_ would overlap with a lot of park features. That's why my PR renders the details of each hole, which won't overlap that much, as we do already for golf.

imagico (changes requested):
I tested this now - which was difficult because i had almost no matching features in my test database and those which there were were not mapped with the scheme this change seems to support.  I also looked a bit more in depth at the mapping of this in OSM.

My observations:

* The symbol seems to be mostly pixel aligned (except for the bottom bar).  It is however non-recognizable to the non-insider, >90 percent of our target map users will not be able to understand it intuitively, not even in the way that it indicates some kind of sport.  It also does not match the symbol paradigm we use for golf courses and which i have suggested for illustrating sports in general - to depict a person practicing the sport in question (which is also in general by far the most common paradigm in sports symbology).
* Since the vast majority of leisure=disc_golf_course features in the OSM database seem to be in public places intersecting general purpose public infrastructure the fact that the rendering of details visually interferes with other infrastructure (in particular public paths) and can be confused with other features rendered in similar design (like barriers, power lines) is a problem.  See examples in #3766.
* This change renders disc_golf=tee with a dark gray polygon fill while this tag is ten times as often used on nodes as on polygons.  That would mean nudging mappers to change their established mapping practice to hand drawing map symbology.

There seem to be various competing tagging schemes for disc golf infrastructure at the moment:

- mapping the whole disc golf site with a wrapping polygon or a node
- mapping the whole disc golf site with a site relation containing the individual holes as members (probably chosen frequently because a disc golf course in a public setting does not usually have a verifiable perimeter)
- mapping the individual holes with leisure=disc_golf_course
- mapping either the whole site or the individual holes with leisure=pitch + sport=disc_golf

This PR supports the first scheme while there seems clear indication that there is no mapper consensus that this is the only valid scheme.  As such i have doubts that it would be advisable under the goals we have to specifically support this and discourage the others.

I would suggest that we

* not render this at this time because of the lack of a well established mapping scheme and a good and documentation of that
* if we decide to render leisure=disc_golf_course in principle we choose a symbol that matches the symbolization paradigm of golf courses and that can be used for other sports (see #844).
* if we decide to render leisure=disc_golf_course in principle we refrain from rendering the details of disc golf courses because they are likely and in many cases of existing mapping in OSM interfering with not sport specific public infrastructure display that is of higher importance for the target map user.

imagico:
Pixel aligned sample renderings:

jdmonin:
Thank you both for your feedback.

- I can try to draw a leisure=disc_golf_course symbol that's a person throwing a frisbee disc
- Although the current tagging seems to be messy, what's in the wiki is much more straightforward, and the iD presets support that, so it's likely that people will be mapping more disc golf features with that straightforward scheme:
  - course: leisure=disc_golf_course  sport=disc_golf
  - hole: disc_golf=hole  with optional name, par, ref (hole number)
  - tee: disc_golf=tee  with optional surface
  - basket: disc_golf=basket 
- I'd be glad to add more details to the wiki about that tagging scheme
- I can remove the disc_golf=tee polygon rendering, it's probably too much detail.
- What if we increase the minimum zoom level to render hole/basket details? My inspiration for this PR was playing on a course which had no map and wondering which hole is next; I think course details would be useful to people to see on OSM. At a higher zoom level, they wouldn't be mistaken for very short power lines or other infrastructure.

jeisenbe:
I agree with the preceeding concerns by @imagico and @pnorman about the symbol. 

Also the line used for the “hole” does not as work as well here as it does for standard golf, where there should be an area rendering for the pin, fairway, and green which provides visual context. A thin line through a park or trees does not clearly look like a disk golf feature. 

While in principle I would be fine with showing this if we can find an easy to understand rendering, at this time my comments in [3766](https://github.com/gravitystorm/openstreetmap-carto/issues/3766#issuecomment-486961360) are still valid: The tags used for rendering in this PR are not widely accepted and established, though they may be in the future, since they were only recently added to the iD editor presets.

Also, there are many popular sports which are not currently distinguished - see #844 - and any new symbol for this sport ought to be designed to fit in with potential renderings for other more common types of sport. 

See also concerns at https://github.com/gravitystorm/openstreetmap-carto/issues/3766#issuecomment-1146617348

@jdmonin would you be interested in the bigger problem of trying an initial PR for #844 in the meantime?

pnorman:
Closing because the above issues (in particular symbol issues) are unresolved. If you want to fix them, feel free to open a new PR.

PR #4561, opened 2022-06-01
Mateusz Konieczny (matkoniecz) - no longer active

matkoniecz:
I am no longer really active.

Thanks for working together and improving OSM Carto together!

I am no longer active for several reasons:

- I implemented most of what I wanted and could improve
- lack of time
- people kept trying to use OSM Carto to force their pet tagging schemes on others, bypassing discussions
- people were unhappy, no matter which decision I have taken (see also the first point)
- people were unhappy also if no decision was taken
- repeated insults from some of people who expected me to work on OSM Carto

as long as there were things that I could improve and wanted improve other issues were not important, but hobby projects are worth it when you are either doing something that you really want or people are really appreciating/using/liking what you are doing. In case of neither not really applying it is better to change the hobby project.

I am removing myself as I am neither active anymore, stopped planning to restart my activity and removed my permissions on this repository.

It is possible that I will contribute again to this or another map style.

EDIT: more background at https://www.openstreetmap.org/user/Cristoffs/diary/399287#comment52602 https://lists.openstreetmap.org/pipermail/talk/2022-June/087533.html (to be clear, that is not a main reason - just a final straw)

### DIFF
--- a/CODE_OF_CONDUCT.md
+++ b/CODE_OF_CONDUCT.md
@@ -62,7 +62,7 @@ When using the OpenStreetMap Carto spaces you should act in the spirit of the va
 
 The OpenStreetMap Carto maintainers are responsible for handling conduct-related issues. Their goal is to de-escalate conflicts and try to resolve issues to the satisfaction of all parties.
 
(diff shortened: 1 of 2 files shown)


imagico (approved):
Thanks for all your contributions.  Your input will be missed.

If you have the time - it would be good to have the list of retired maintainers in README.md in consistent order - either by seniority (like the list of active maintainers) or by time of retirement.

matkoniecz:
> by time of retirement.

done

PR #4579, opened 2022-06-27
Do not render railway crossings if only tramways are involved (#4559).

### LINKED ISSUE #4559: Do not render railway=level_crossing for tramways
### Expected behavior
railway=level_crossing should not be rendered if it is part of a way tagged railway=tram

### Actual behavior
railway=level_crossing is always rendered as a large diagonal cross.  For tramways, which naturally have many level crossing, this pollutes the map, and in certain areas dominates the look of the map, without adding much useful information.  This is particularly bad for printed maps, like the ones generated by MapOSMatic.
grafik

https://www.openstreetmap.org/#map=16/47.5462/7.5634&layers=N
grafik

It was debated whether to map level crossings for tramways, with a tendency to not do so:
https://lists.openstreetmap.org/pipermail/tagging/2017-April/031941.html
I'm not sure this is implementable with the software infrastructure for carto.  However, e.g. in Basel, there are hundred of them mapped already, so it would be awkward to remove them manually.

johsin18:
Fixes #4559 

Do not render railway crossings if only tramways are involved. 

https://www.openstreetmap.org/#map=16/47.5403/7.5733

Before
grafik

After
grafik

### DIFF
--- a/project.mml
+++ b/project.mml
@@ -2220,7 +2220,6 @@ Layer:
             way,
             name,
             COALESCE(
-              'railway_' || CASE WHEN railway IN ('level_crossing', 'crossing') AND way_area IS NULL THEN railway END,
               'amenity_' || CASE WHEN amenity IN ('bench', 'waste_basket', 'waste_disposal') AND way_area IS NULL THEN amenity END,
               'historic_' || CASE WHEN historic IN ('wayside_cross', 'wayside_shrine') AND way_area IS NULL THEN historic END,
               'man_made_' || CASE WHEN man_made IN ('cross') AND way_area IS NULL THEN man_made END,
@@ -2275,6 +2274,30 @@ Layer:
               END DESC,
               way_pixels DESC NULLS LAST
           ) AS amenity_low_priority
+    properties:
+      cache-features: true
+      minzoom: 14
+# Render railway crossings except if they concern tramways only
+  - id: railway-crossings
+    geometry: point
+    <<: *extents
+    Datasource:
+      <<: *osm2pgsql
+      table: &railway-crossing_sql |-
+        (SELECT DISTINCT ON (crossing.way)
+            'railway_' || crossing.railway AS feature,
+            crossing.way AS way,
+            crossing.access AS access,
+            crossing.railway AS type,
+            crossing.railway AS railway,
+            track.railway AS int_lc_type
+          FROM planet_osm_point crossing
+            JOIN planet_osm_line track
+              ON ST_DWithin(crossing.way, track.way, 0.1) -- Assumes Mercator
+          WHERE crossing.railway IN ('crossing', 'level_crossing')
+            AND track.railway IS NOT NULL AND track.railway <> 'tram'
+          ORDER BY crossing.way
+        ) AS railway_crossing_sql
     properties:
       cache-features: true
       minzoom: 14
--- a/style/amenity-points.mss
+++ b/style/amenity-points.mss
@@ -1471,7 +1471,8 @@
   }
 }
 
-#amenity-low-priority {
+#amenity-low-priority,
+#railway-crossings {
   [feature = 'man_made_cross'][zoom >= 16],
   [feature = 'historic_wayside_cross'][zoom >= 16] {
     marker-file: url('symbols/man_made/cross.svg');


project.mml
```diff
@@ -2278,6 +2277,30 @@ Layer:
...
+    Datasource:
+      <<: *osm2pgsql
+      table: &railway-crossing_sql |-
+        (SELECT DISTINCT ON (crossing.way)
+            'railway_' || crossing.railway AS feature,
+            crossing.way AS way,
+            crossing.access AS access,
+            crossing.railway AS type,
+            crossing.railway AS railway,
+            track.railway AS int_lc_type
```
imagico, line 2293: These appear to be unused so should be removed.

project.mml
```diff
@@ -2278,6 +2277,30 @@ Layer:
...
+          FROM planet_osm_point crossing
+            JOIN planet_osm_line track
+              ON ST_DWithin(crossing.way, track.way, 0.1) -- Assumes Mercator
+          WHERE crossing.railway IN ('crossing', 'level_crossing')
+            AND track.railway IS NOT NULL AND track.railway <> 'tram'
```
imagico, line 2298: That means all railway crossings that are wrongly mapped not on any kind of railway (tram or otherwise) are hidden and not visible to the map user.  That is incompatible with our goal of providing constructive mapper feedback i think.  We - as a general rule - do not hide features from rendering based on spatial analytics telling us that they are mapped incorrectly.  If a mapper maps a railway crossing in the middle of nowhere we show that because the mapper either has a good reason for that or it is an error and should be corrected and showing it helps mappers in seeing such errors.

project.mml
```diff
@@ -2278,6 +2277,30 @@ Layer:
...
+          ORDER BY crossing.way
```
imagico, line 2299: Although this is not a newly introduced problem i would like to mention that we generally want to use explicit ordering in queries where the order of features returned matters (which is the case here because the symbols are blocking each other).  It would therefore be advisable to order also by (a) priority based on railway type and (b) osm_id to always have a defined drawing order.

project.mml
```diff
@@ -2278,6 +2277,30 @@ Layer:
     properties:
       cache-features: true
       minzoom: 14
+# Render railway crossings except if they concern tramways only
+  - id: railway-crossings
+    geometry: point
+    <<: *extents
+    Datasource:
+      <<: *osm2pgsql
+      table: &railway-crossing_sql |-
```
imagico, line 2286: The `&railway-crossing_sql` is unused and therefore should be removed.

style/amenity-points.mss
```diff
@@ -1471,7 +1471,8 @@
   }
 }
 
-#amenity-low-priority {
+#amenity-low-priority,
+#railway-crossings {
```
imagico, line 1475: Since there is no overlap in style rules between the `amenity-low-priority` and `railway-crossings` layers the style rules for those should be separate.

imagico (changes requested):
Thanks for the pull request.  This is in principle the correct approach to implement rendering of railway crossings differentiated by type or railway.

My assessment of the idea from https://github.com/gravitystorm/openstreetmap-carto/issues/4559#issuecomment-1142659597 is unchanged:

> Considering how widely these are mapped and how significant they are often for navigation i would be skeptical about removing rendering. But adjusting the starting zoom level or the symbol specifically for tram crossings might be a good idea.

I would encourage you to try modifying your PR along these lines, i.e. instead of dropping rendering of railway crossings when they are located on a tram track to render those in a distinct styling with a later starting zoom level and a less massive symbol.  Starting at z16/z17 and using the smaller symbol currently used for z14/z15 would be a starting point but you can also try different symbol variants.

dch0ph:
@joto Is there any functionality in the flex input of `osm2pgsql` that would simplify the process of extracting tagging information from the railway way to which a level crossing node is attached? Ideally we would extract the railway type at import, rather than perform multiple local spatial searches at render time.  

My understanding is that the only "connections" that can be made between OSM objects are between relations and their member ways?

But there are a few contexts where it would be useful to "mark" nodes and subsequently extract tag values from their "parent" ways e.g. fords on waterways. Other cases may be too awkward, e.g. mini roundabouts generally link multiple ways.

dch0ph:
Many thanks for confirming. It would better to avoid these spatial searches if alternatives were available. But, in the end, cases where we need to "cross-refer" between nodes and intersecting ways are pretty rare; generally all the information needed to render a node is tagged on the node itself.

imagico:
Lets not intermix discussion of data preprocessing strategy with a concrete pull request.  We have *developing a data preprocessing strategy* in #5027 - for more detailed discussion a new issue can be opened.

To avoid confusing @johsin18 - the requests for changes in the review remain, if you need advice/help with any of them please ask.

PR #4589, opened 2022-07-08
Fail CI if YAML files aren't valid

### LINKED ISSUE #4583: YAML file validation in CI does not work
[Summary](https://github.com/gravitystorm/openstreetmap-carto/actions/runs/2607866115)

```
Run find . \( -type f -name '*.yaml' -o -name '*.yml' -o -name '*.mml' \) \
  find . \( -type f -name '*.yaml' -o -name '*.yml' -o -name '*.mml' \) \
  -exec python -c "from yaml import safe_load; safe_load(file('{}'))" \;
  shell: /bin/bash -e {0}
  env:
    pythonLocation: /opt/hostedtoolcache/Python/3.7.[1](https://github.com/gravitystorm/openstreetmap-carto/runs/7174601674?check_suite_focus=true#step:8:1)[3](https://github.com/gravitystorm/openstreetmap-carto/runs/7174601674?check_suite_focus=true#step:8:3)/x6[4](https://github.com/gravitystorm/openstreetmap-carto/runs/7174601674?check_suite_focus=true#step:8:4)
  File "<string>", line 1, in <module>
NameError: name 'file' is not defined
  File "<string>", line 1, in <module>
NameError: name 'file' is not defined
  File "<string>", line 1, in <module>
NameError: name 'file' is not defined
  File "<string>", line 1, in <module>
NameError: name 'file' is not defined
  File "<string>", line 1, in <module>
NameError: name 'file' is not defined
  File "<string>", line 1, in <module>
NameError: name 'file' is not defined
  File "<string>", line 1, in <module>
NameError: name 'file' is not defined
```

There's two things wrong here:
1. It's failing to check the files
2. Even though it errored, it passed the check

pnorman:
Fixes #4583

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -21,8 +21,8 @@ jobs:
       run: set -o pipefail
     - name: Validate YAML files
       run: |
-        find . \( -type f -name '*.yaml' -o -name '*.yml' -o -name '*.mml' \) \
-        -exec python -c "from yaml import safe_load; safe_load(file('{}'))" \;
+        find . \( -type f -name '*.yaml' -o -name '*.yml' -o -name '*.mml' \) -print0 \
+        | xargs -0 -I '{}' python3 -c 'from yaml import safe_load; safe_load(open("{}"))'
     - name: Validate SVGs for valid XML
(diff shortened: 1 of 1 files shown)


StyXman (approved):
ooops, good catch.

PR #4592, opened 2022-07-10
Recommend disabling JIT

### LINKED ISSUE #4095: Evaluate use of PostgreSQL JIT
In https://github.com/openstreetmap/operations/issues/386 I encountered a problem with one of Mapnik's `LIMIT 0` queries having a high cost and use of the JIT being planned, causing the query to take a second to execute.

My development setup is not normal, so this needs testing to see if it reproduces in smaller extracts with Kosmtik, and I haven't done testing with renderd to see if it impacts a production setup.

pnorman:
PostgreSQL JIT cripples Kosmtik, and slows down rendering of some zooms.

Fixes #4095


imagico (approved):
Frankly i have never heard of a practical map rendering use case where JIT is of benefit and there are plenty of map rendering related situations (and not just the more esoteric problems of `LIMIT 0` queries) where it is making things prohibitively expensive.

PR #4596, opened 2022-07-10
Increase readability of code comments related to input/output parameters

mboeringa:

- 

Before

After

### DIFF
--- a/openstreetmap-carto.lua
+++ b/openstreetmap-carto.lua
@@ -485,8 +485,8 @@ local excluded_railway_service = {
     yard = true
 }
 --- Gets the z_order for a set of tags
--- @param tags OSM tags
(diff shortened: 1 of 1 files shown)


pnorman (changes requested):
The @ syntax is a standard documentation syntax as part of LuaDoc and can be processed by tools like [LDoc](https://github.com/lunarmodules/LDoc).

It's also basically the same as doxygen, except C uses /* */ for comments.

mboeringa:
OK, although I do think the choice of not using some sort of separator between the parameters name and the explanation does make the readability of the code comment and the parameter definition suffer, because it is unclear at first glance where the name of the parameter ends, and the explanation starts. 

But that then is a kind of upstream problem / mildly problematic historic choice.

PR #4600, opened 2022-07-13
Use barrier icon for gate fill

### LINKED ISSUE #4599: Barrier=gate is too prominent after recent changes
### Expected behavior
Objects tagged with barrier=gate are not so prominent on the map. 

### Actual behavior
Currently, its black color (#000000) dominates everything else around and gathers unnecessary focus on this (honestly - not so important) object. It may have something to do with cleaning SVG commit: https://github.com/gravitystorm/openstreetmap-carto/pull/4457 .

Zoom 17 - screenshot:
obraz
https://www.openstreetmap.org/#map=17/52.14917/20.82287

Old vs new rendering:
obraz

Previously:
obraz

Now:
obraz

pnorman:
Fixes #4599

#4457 removed the fill from symbols/barrier/gate.svg, which was correct, but there wasn't a marker-fill for this icon in the MSS. This PR adds that.

With PR:
image

master:
image

cc @pitdicker

### DIFF
--- a/symbols/barrier/gate.svg
+++ b/symbols/barrier/gate.svg
@@ -1,3 +1,11 @@
-<svg xmlns="http://www.w3.org/2000/svg" width="6" height="5" viewBox="0 0 6 5">
-  <path d="m0,1 h6 v1 h-6 z m0,2 h6 v1 h-6 z m1,2 l3,-5 h1 l-3,5 z"/>
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+
+<svg xmlns="http://www.w3.org/2000/svg"
+   version="1.1"
+   width="6"
+   height="5">
+	<path d="M 0,0 6,0 6,5 0,5 z" fill="#ffffff" />
+	<path d="M 0,1 6,1 6,2 0,2 z" fill="#3f3f3f" />
+	<path d="M 0,3 6,3 6,4 0,4 z" fill="#3f3f3f" />
+	<path d="M 4,0 1,5 2,5 5,0 z" fill="#3f3f3f" />
 </svg>


pnorman:
After merging I propose releasing v5.5.1 with this bugfix

pitdicker:
Ai, I am sorry. Thanks for fixing this!

pnorman:
> Ai, I am sorry. Thanks for fixing this!

Could you add a review?

pitdicker (approved):
Reviewed en tested locally :+1:

pitdicker:
Was it intended to have a white square behind the gate? But you are right @imagico, that is a difference. `level_crossing` and `level_crossing2` also have/had a casing, `kissing_gate`, `cattle_grid` and `cycle_barrier` don't have it.

imagico:
> I'm okay with that because our standard is monochromatic icons.

Please no major design change as a footnote to a technical change that was merged under the assumption of having no visible effect.

We can discuss removal of white casing of the small gray barrier symbols in this style, but lets have a proper discussion of advantages and disadvantages and proper testing in different contexts for that.

imagico (changes requested):
Tested this and can confirm that the gate symbol is the only one that differs in content (and not just in alignment like many other) from 5.4.0.

I would suggest to revert gate.svg to the old version with white casing and we open a new issue to discuss the future design of small barrier symbols using `@barrier-icon`.  This is not only inconsistent right now (with gate and level crossings having a casing and the others not) but also has issues with the other symbols lacking a casing, in particular lift gate and cycle barrier - like here:

https://www.openstreetmap.org/#map=17/48.01547/7.83017

We should see if we can develop a uniform design here that works well in most situations, even with a context involving dark linework like bridges and cycleways.

pitdicker:
Adding a white line will allow the symbol to still be colored by the mss, like so (the same method is used for other symbols with casing):
```svg
<svg xmlns="http://www.w3.org/2000/svg" width="6" height="5" viewBox="0 0 6 5">
  <path d="m0,2.5 h6" stroke="#ffffff" stroke-width="5"/>
  <path d="m0,1 h6 v1 h-6 z m0,2 h6 v1 h-6 z m1,2 l3,-5 h1 l-3,5 z"/>
</svg>
```

imagico (approved):
Renders as before now.

PR #4601, opened 2022-07-13
Use indexes for turning circle queries

pnorman:
While testing I noticed that on zoom 15, the turning circle query is one of the slowest. This change reduces the duration from 20ms to to 1.2ms. The query times were several seconds with cold caches, but that is difficult to benchmark and reproduce.

No rendering changes.

### DIFF
--- a/project.mml
+++ b/project.mml
@@ -635,8 +635,9 @@ Layer:
           WHERE p.highway IN (
             'turning_circle',
             'turning_loop',
-            'mini_roundabout'
-          )
+            'mini_roundabout')
+            AND l.way && !bbox!
+            AND p.way && !bbox! -- Both conditions are necessary for good index usage, even with the DWithin above
           ORDER BY p.way, v.prio
         ) AS turning_circle_sql
     properties:


imagico:
Do you know why `l.way && !bbox!` is necessary here?  You would expect it to use `l.way && ST_Expand(p.way, 0.1)`.

pnorman:
old query explain analyze: https://explain.depesz.com/s/zlAz
new query explain analyze: https://explain.depesz.com/s/z1YA

The difference is one of query planning. PostgreSQL believes old query is much more expensive, so plans a parallel query, but because it's fast, this is a bad plan. If you look at the raw explains, you can see the cost estimates. With both bbox conditions, it estimates 1 row, with only one it estimates 26k rows.

I believe this is because although we know that `p.way && !bbox! and st_dwithin(p.way, l.way, 0.1)` implies `l.way && st_expand(!bbox!, 0.1)`, PostgreSQL doesn't, so it can't plan the query as well.

Unfortunately, I can't reproduce the queries I was getting before I ran a new VACUUM ANALYZE

imagico (approved):
Works and i can confirm that in a bit of non-systematic testing it moves from being the slowest query to being somewhere in the middle among the slower queries (it is still not fast obviously with two tables being involved).

pnorman:
> That is a bit weird. I wonder if it changes if you reformulate the query with a subquery for the way rather than with a join.

It will probably turn the subquery into a join internally, being equivalent.

In general, if you're doing geometry operations of some kind and can add more constraints that you can infer from what you're doing, it will help give more index usage and query estimation options.

Issue #3759, opened 2019-04-22
Text labels of shops are no longer displayed


jeisenbe:
### Expected behavior: 
- shop=mall and other shop name labels should be displayed at appropriate zoom levels

### Actual behavior: 
- since PR #3712 the shop=mall name label and other shop name labels are not rendered.
- I noticed this problem while trying to add an area fill for shop=mall, similar to that used for amenity=marketplace and landuse=retail, both in test data and when looking at malls (without building=*) in Singapore

### Links and screenshots illustrating the problem
Marina Bay, Singapore
z19 from immediately before PR #3712
![marina-bay-before-z19](https://user-images.githubusercontent.com/42757252/56496189-4cef4f00-6533-11e9-9b61-e47b873a2605.png)

z19 current master (and same as immediately after PR #3712)
![z19-marina-bay-after](https://user-images.githubusercontent.com/42757252/56496184-45c84100-6533-11e9-8333-2c99160b6676.png)

z22 with current master - still no rendering
![z22-marina-bay-mall-master](https://user-images.githubusercontent.com/42757252/56496194-55478a00-6533-11e9-94cc-638670f0fb4e.png)

Inspector view shows that the feature is in the rendering database at this location:
![z22-marina-bay-link-mall-inspector](https://user-images.githubusercontent.com/42757252/56496221-67c1c380-6533-11e9-84d6-da31e8fe6095.png)

[Note that this is an underground mall]

jeisenbe:
I've updated the title; it's not just shop=mall
All shop text labels are no longer being displayed after #3712

This is another real example: Diamond Run Mall in Vermont. https://www.openstreetmap.org/#map=17/43.58196/-72.96550

It's mapped both as an area (which surrounds the service road loop) tagged shop=mall and the building is also tagged shop=mall.

Hence before it would render like this with 2 labels. You can also see that the department store names are shown (rendering with commit 712afb87f8c7d506437cedb5273cea9fa367f524 ):
z17 before: 
![z17-diamond-run-mall-before](https://user-images.githubusercontent.com/42757252/56498850-8bd6d200-653e-11e9-9126-b92e8390ab0f.png)

Now it renders like this after commit fee9ce2e67ef7b96935bfe61104041d515119759 - no text labels for the mall or the shops:
z17 current master:
![z17-diamond-run-mall-after](https://user-images.githubusercontent.com/42757252/56498848-89747800-653e-11e9-878d-45aa35f55bc0.png)

[...]

PR #4603, opened 2022-07-14
Require feature check when checking for mall

pnorman:
#3759 + previous PRs changed the text-points feature=shop_mall into feature=shop shop=mall. This was done incorrectly in one place where the [feature='shop'] condition was not added, leading to having to check for shop=mall on everything else in the layer, even if it didn't have feature=shop.

This change brings compilation time from 9s to 6.5s and the text-point layer from 2.3s to 1.1s. The total XML decreases from 46670 to 45404 lines.

ref #1941

### DIFF
--- a/style/amenity-points.mss
+++ b/style/amenity-points.mss
@@ -2202,8 +2202,8 @@
   [feature = 'leisure_pitch'] {
     [zoom >= 10][way_pixels > 3000][is_building = 'no'],
     [zoom >= 17][is_building = 'no'],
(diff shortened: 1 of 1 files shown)


pnorman:
There might be changes to objects tagged with tags like `leisure=fitness_station shop=mall`, but just to restore the rendering to what they were prior to the previously mentioned PRs. I couldn't find any examples.

pnorman:
Pinging for PR review.

imagico (approved):
Looks good and i also did not observe any rendering differences in various tagging combination tested.

PR #4520, opened 2022-03-02
update INSTALL.md with fonts-unifont package

andrewharvey:
ttf-unifont is now fonts-unifont on recent Debian releases

Old: https://packages.debian.org/buster/all/ttf-unifont/filelist
Current: https://packages.debian.org/sid/all/fonts-unifont/filelist

PR #4606, opened 2022-07-16
Use local copy of fonts

### LINKED ISSUE #4490: Diacritics in a particular vilage label are rendered at wrong places
The name of the Czech village "Břežany II" is rendered with the carons over 'e' and 'a' instead of 'r' and 'z'. This happens both on the label for the village and for border of its area. Also, similar issues are found on several other layers featured at [osm.org](https://osm.org). The name seems to be written in correct UTF-8 (the roman 'II' is written using U+2161, rather than two i's).

The village is here: https://www.openstreetmap.org/relation/441243#map=13/50.0865/14.7830

The wrong renderings:
image
image

For some reason, the bus stop in the village has correct rendering:
image

I don't think I've seen such misrendering for any other place.

(Also, I am not sure whether this belongs here or to the operations bug tracker. The [wiki](https://wiki.openstreetmap.org/wiki/Standard_tile_layer) says that it belongs rather to the operations' repo, but that issue tracker doesn't seem to me as being map/rendering related.)

### LINKED ISSUE #4602: Should we vendor fonts?
Out of #4490 I investigated the versions of the fonts we use, and what systems have
Most Debian-derived systems have Noto fonts from 2020. Since then, there have been [updates to noto](https://github.com/notofonts/noto-fonts#recent-updates), with bug fixes, including bold support diacritics in NotoNaskhArabic.

Even if we got Debian to update the fonts-noto package today, it wouldn't be in a Ubuntu LTS release until 2024, and wouldn't be in backports until late 2022

Vendoring the fonts would involve writing a script that copies them from the noto-fonts repo into a ./fonts directory, and adding `font-directory: url('fonts');` to fonts.mss

This would add an estimated 18MB for all NotoSans Regular fonts[^1], 5.5MB for NotoSans bold[^2], 52MB for hanazono, and 14MB for unifont.

Advantages

1. Style users get up to date fonts without needing to manually install unpackaged stuff
2. We can get rid of fallback fonts, since we will control the fonts included
3. We can get rid of missing font warnings, which will also allow strict mode loading of the XML
4. We would remove significant deployment to deployment variations in text rendering that we currently have (e.g. people testing #4490)

Disadvantages
1. Repo size goes up by about 90MB
2. We're vendoring third-party content rather then relying on build systems and packaging
3. We have to keep the vendored fonts up to date with a script

[^1]: `find noto-fonts/unhinted/ -name 'NotoSans*-Regular.ttf' -type f | xargs du -hc --apparent-size`
[^2]: `find noto-fonts/unhinted/ -name 'NotoSans*-Bold.ttf' -type f | xargs du -hc --apparent-size`

### LINKED ISSUE #4610: Syriac language not showing in catro
### Expected behavior

Syriac language showing in carto (it was working before for years, it stop showing in the last days)

### Actual behavior

Syriac language showing as "▯▯▯▯"...

screenshot before and after: https://i.ibb.co/SQq0xCz/image.png      

one of the examples:     
https://www.openstreetmap.org/node/1972159484

pnorman:
Fixes #4602
Fixes #4490
Fixes #4520
Fixes #4610

This moves to downloading a copy of the fonts locally, allowing more recent versions to be used, guarantee coverage, and eliminate fallback fonts for old systems. In addition the CJK font is re-ordered with the Arabic font to stop overwriting the design of commonly used signs in other fonts.

By having a more recent copy, it allows the number of scripts with bold support to be significantly improved.

The total download size is 98MB

image

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -55,48 +55,15 @@ The script downloads shapefiles, loads them into the database and sets up the ta
 ## Fonts
 The stylesheet uses Noto, an openly licensed font family from Google with support for multiple scripts. The stylesheet uses Noto's "Sans" style where available. If not available, this stylesheet uses another appropriate style of the Noto family. The "UI" version is used where available, with its vertical metrics which fit better with Latin text.
 
-DejaVu Sans is used as an optional fallback font for systems without Noto Sans. If all the Noto fonts are installed, it should never be used.
-
 Hanazono is used a fallback for seldom used CJK characters that are not covered by Noto.
-
-Unifont is used as a last resort fallback, with it's excellent coverage, common presence on machines, and ugly look. For compatibility reasons, we support two Linux-distributions-specific versions of Unifont, therefore it's expected that you *always* get a warning about a missing Unifont version.
-
-If you do not install all the fonts, the rendering itself will not break, but missing glyphs will be ugly.
 
 For more details, see the documentation at [fonts.mss](style/fonts.mss).
 
-### Installation on Ubuntu/Debian
-
-On Ubuntu 16.04 or Debian Testing you can download and install most of the required fonts
+To download the fonts, run the following script
 
 ```
-sudo apt-get install fonts-noto-cjk fonts-noto-hinted fonts-noto-unhinted fonts-hanazono ttf-unifont
+scripts/get-fonts.sh
 ```
-
-Noto Emoji Regular (*not* Noto Color Emoji) can be downloaded [from the Noto Emoji repository](https://github.com/googlefonts/noto-emoji).
-
-It might be useful to have a more recent version of the fonts for [rare non-latin scripts](#non-latin-scripts). The current upstream font release has also some more scripts and style variants than in the Ubuntu package. It can be installed [from source](https://github.com/googlefonts/noto-fonts/blob/master/FAQ.md#where-are-the-fonts).
-
-DejaVu is packaged as `fonts-dejavu-core`.
-
-### Installation on other operating systems
-
-The fonts can be downloaded here:
-
-* [Noto homepage](https://www.google.com/get/noto/) and [Noto github repositories](https://github.com/googlefonts?utf8=%E2%9C%93&q=noto)
-* [DejaVu homepage](https://dejavu-fonts.org/)
-* [Hanazono homepage](http://fonts.jp/hanazono/)
-* [Unifont homepage](http://unifoundry.com/)
-
-After the download, you have to install the font files in the usual way of your operating system.
-
-### Non-latin scripts
-
-For proper rendering of non-latin scripts, particularly those with complicated diacritics and tone marks the requirements are
-
-* FreeType 2.6.2 or later for CJK characters
-
-* A recent enough version of Noto with coverage for the scripts needed.
 
 ## Dependencies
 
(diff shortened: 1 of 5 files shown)


pnorman:
I don't have a docker setup on my system, so the docker change could use testing.

INSTALL.md
```diff
@@ -55,49 +55,16 @@ The script downloads shapefiles, loads them into the database and sets up the ta
...
 For more details, see the documentation at [fonts.mss](style/fonts.mss).
 
-### Installation on Ubuntu/Debian
-
-On Ubuntu 16.04 or Debian Testing you can download and install most of the required fonts
+To download the fonts, run the following script
 
 ```
-sudo apt-get install fonts-noto-cjk fonts-noto-hinted fonts-noto-unhinted fonts-hanazono ttf-unifont
+scripts/get-fonts.sh
```
imagico, line 65: That is not specific to Ubuntu/Debian so it should be in a different section i think.

imagico (changes requested):
Not yet tested but wondering:

* should we introduce a dependency on running shell scripts and curl when we use python for all other scripting so far and the necessary means to do this in python are already used in `get-external-data.py` so this would not introduce any new dependencies?
* do we need to maintain two separate font lists (in mss and for the download) or could we combine that into one?

pnorman:
> should we introduce a dependency on running shell scripts and curl when we use python for all other scripting so far and the necessary means to do this in python are already used in get-external-data.py so this would not introduce any new dependencies?

I looked at this but it was a lot of work to reimplement curl -z properly.

> do we need to maintain two separate font lists (in mss and for the download) or could we combine that into one?

I considered this, but ruled it out. We have a specific ordering for some fonts, exceptions, and mapping between file name and font name isn't easy.

INSTALL.md
pnorman: There's no mention of Ubuntu or Debian here.

imagico: Ah, sorry, misread the diff.

mmd-osm:
https://github.com/notofonts/noto-fonts mentions:

_This repository is in the process of being migrated._

_New versions of Noto fonts will soon be available at https://notofonts.github.io/_

Has this been considered already? My impressionwas that the script is still using the old location, but I may be misreading things here.

pnorman:
> https://github.com/notofonts/noto-fonts mentions:
> 
> _This repository is in the process of being migrated._
> 
> _New versions of Noto fonts will soon be available at https://notofonts.github.io/_

The move is still in progress. If you look at their commit log they're adding more fonts to the new site each day. It shouldn't be much work to change the URL when all the fonts are at the new location.

sommerluk (approved):
Looks good.

While this still does not guarantee that all users have exactly the same font version (because of different moments when doing the font download), it is a big step forward.

Also, the better support for bold fonts is great, and of course that we do not have any longer the warnings for missing fonts.

Moreover, removing the fallback fonts is good. Especially, DejaVu appeared in the regular font list and the bold font list, which made things complicate.

I've made a PR correcting a bit the documentation about Arabic.

pnorman:
> I've made a PR correcting a bit the documentation about Arabic.

Note: I don't see this PR

sommerluk:
> Note: I don't see this PR

See here: https://github.com/pnorman/openstreetmap-carto/pull/8

I've rebased it and made a PR directly to this repository now: #4634

Issue #3925, opened 2019-10-09
Utility poles rendered within tertiary highways at z16 to z18


jidanni:
Despite one's best efforts in the editor
https://github.com/openstreetmap/iD/issues/6925#issuecomment-539834155
it is still a fact that most utility poles will end up getting rendered **within** a road.

Everybody knows that utility poles are on the side of a road, not within it. Don't get me wrong: rendering utility poles is very important. Just please, when rendering a road, render the pole to the nearest edge. Thanks.

Autopilot cars will also thank you, no more fear of slamming in to them.

Don't worry. If they are really in the center of a road, they will be on a [traffic island](https://wiki.openstreetmap.org/wiki/Proposed_features/Street_area#traffic_island).

jeisenbe:
Area discussed in the linked comment: https://www.openstreetmap.org/#map=17/24.19249/120.87513

Node example: https://www.openstreetmap.org/node/4778301333#map=19/24.19948/120.87570
(Is this a very narrow road, with the poles quite close to the edge? This node is only 1 meter from the centerline of the `highway=tertiary` way. Perhaps the source of the data for the road and the source for the poles are not aligned properly? It will also be easier to understand once the `power=minor_line` is mapped)

However, many of the properly-aligned poles and roads will show overlap at z16 and z17 in places like Taiwan which are near the equator, since the tertiary road rendering at z16 and z17 is over 20 meters wide (10 pixels is 24 meters at the equator at z16, and 18 pixels is 21.6 meter at z17). Since many tertiary roads are only 6 to 8 meters wide, the rendering is quite a bit wider than the real feature at these zoom levels. 

We could improve this a little by narrowing the rendering of tertiary highways at z16 to z19. This would probably require changing the color to yellow, which means that motorways would need a new rendering as well. 

> when rendering a road, render the pole to the nearest edge.

This isn't very easy to do with this map style and the rendering toolchain we use. It might also confuse mappers if poles close to roads were moved farther out, but those farther away were at the precise location. Such pre-processing of data is more feasible for specialized map styles and custom maps.

jidanni:
[Streetview](https://goo.gl/maps/wBC9GLLJyTnGxY4MA) gives you a general idea. Anyway trying to better position these pole will in fact probably just make things worse, as we are not supposed to "edit for the renderer" anyway.

Different sized roads at different latitudes? There should be a formula to keep them the same size no matter what latitude!

[...]

Issue #3891, opened 2019-09-18
Parking fill color is the same as living street fill color 


jeisenbe:
### Expected behavior
- Parking lots should be clearly distinguished from living street ways and areas

### Actual behavior
- The newer `parking-fill` color, `#eeeeee`  is Lch (94,0,180)
- This is indistinguishable from `living-street-fill` color, `#ededed` - also Lch (94,0,180)
- Only the casing is different

jeisenbe:
Also, `parking-outline` still uses `saturate(darken` but the `saturate` doesn't do anything, since saturation is 0.

imagico:
Not as close but there are various other cases where unrelated features are rendered with very similar fill colors.  Specifically apron and pedestrian colors [are very close now](http://davidjohnstone.net/pages/lch-lab-colour-gradient-picker#dadae0,dddde8) - as well as #2353, #2905.

I am generally not sure if there is agreement among maintainers that differences in fill colors should represent differences in meaning and colors should therefore not be overloaded for unrelated purposes.  Similarity to other unrelated color played no role in discussion of most of these changes (#2292, #2934, #3444) and i am not sure if that is because it was being overlooked or if it was considered not to matter when the change was made.  We also have cases with a deliberate choice to aggregate things fairly dissimilar in meaning into a common color (https://github.com/gravitystorm/openstreetmap-carto/pull/3843#issuecomment-524541234).  From my perspective it comes all back to #2270.

jeisenbe:
The apron vs pedestrian color similarity appears to have been overlooked in #3444 and the related issue #3385, from what I can see. There was discussion of using the `parking` color, since aprons are parking areas for aircraft, but then the current color was picked to be a little more visibly related to taxiways.

It doesn't look like there was any discussion of the similarity between the new parking color and living-street-fill in #2934 or #2904 - these things are easy to miss if you don't look for them carefully.

kocio-pl:
Could you show real problems with that? I don't remember a single time I had mistaken them.

jeisenbe:
For some reason highway=living-street areas do not have a casing, unlike highway=pedestrian and highway=residential areas, otherwise they look identical to parking lots:

z15 https://www.openstreetmap.org/#map=14/52.5183/13.1686
![z15-parking-vs-living-street-area](https://user-images.githubusercontent.com/42757252/65369999-2e7eb280-dc8f-11e9-81ea-607278868cb3.png)

[...]

Issue #3540, opened 2018-11-30
Use violet-yellow range for roads


kocio-pl:
This ticket is about implementing roads color palette from alternative-colors style @imagico fork, described in this blog post: http://blog.imagico.de/more-new-colors/ . 

![roadcolor_z15](https://user-images.githubusercontent.com/5439713/47878586-f1af9000-de1e-11e8-9b21-1e0287b276ce.png)

This is just an improvement of the current roads color scheme. Big road style redesign in 2015 (#1736) was largely a successful change, except some corner cases. like paths or tertiary roads. @matkoniecz used a clever hack to make red-yellow palette distinguishable enough, but it meant that tertiary roads need to be white - just bolder than all the lower classes. Yet this was still just a hack, design wise tertiary roads were left behind.

Unfortunately attempt to add lighter yellow was a failure (https://github.com/gravitystorm/openstreetmap-carto/pull/2228#issuecomment-236075503), because the palette was just too short. Extending it with a dark violet does the trick. It allows to make tertiary to not be oversized, which looks bad on two-carriage ways:

![screenshot_2018-11-30 openstreetmap](https://user-images.githubusercontent.com/5439713/49268382-76fa8480-f45f-11e8-8ff3-2753227c7052.png)

Current works on admin borders color change lead to some grey+violet shade, which does not create problems - even 50% gray+violet is different enough (#3489):

![sbbo0u5x](https://user-images.githubusercontent.com/5439713/47878423-8cf43580-de1e-11e8-9270-fce69551563c.png)
![6 nsthx_](https://user-images.githubusercontent.com/5439713/47879083-0a6c7580-de20-11e8-96c3-7c606bfb81f8.png)

Now I feel 75% grey proposition is even better, which means further color distance from the borders.

A look at the roads comparison shows also that this change makes secondary roads thinner, which might help with similar two carriage ways problem with secondary:

![screenshot_2018-11-30 openstreetmap 1](https://user-images.githubusercontent.com/5439713/49268679-25eb9000-f461-11e8-9687-1f82bd525032.png)

jeisenbe:
Are you considering the changes to cycle paths that were shown in that
comparison, or just the roads?

kocio-pl:
This chunk is only about the roads to keep the problem manageable. However footway and path rendering change looks can be discussed in a separate ticket.

kocio-pl:
[...]

PR #4607, opened 2022-07-16
Runway and taxiway rendering rework

### LINKED ISSUE #2169: separate aeroways from rail/road query
Currently there is a massive query that extracts data about rail, roads and aeroways.

Rail and roads need to be together to handle layering, but it should be possible to extract aeroways to separate query.

extracted from https://github.com/gravitystorm/openstreetmap-carto/pull/2168/files#r67127757

### LINKED ISSUE #4340: Runway, taxiway areas should not be rendered

pnorman:
- Refactor aeroways to their own layers. Fixes #2169
- Add a dashed centerline to runways
- Scale runway and taxiways past zoom 16
- Stop rendering runway and taxiway areas. Fixes #4340

[zoom 15 YVR](https://www.openstreetmap.org/#map=#15/49.1940/-123.1950)
image

[zoom 17 YNJ](https://www.openstreetmap.org/#map=17/49.10051/-122.62892)
image

[zoom 19 WA07](https://www.openstreetmap.org/#map=19/48.37482/-122.33878)
image

[zoom 16 FRA](https://www.openstreetmap.org/#map=16/50.0444/8.5342)
image

[zoom 15 SIN](https://www.openstreetmap.org/#map=15/1.3602/103.9856)
image

cc @jdhoek

### DIFF
--- a/project.mml
+++ b/project.mml
@@ -1032,6 +1032,7 @@ Layer:
         (SELECT
             way,
             aeroway,
+            tags->'runway' AS runway,
             bridge IN ('yes', 'boardwalk', 'cantilever', 'covered', 'low_water_crossing', 'movable', 'trestle', 'viaduct') AS bridge
           FROM planet_osm_line
           WHERE aeroway IN ('runway', 'taxiway')
--- a/project.mml
+++ b/project.mml
@@ -1728,8 +1728,11 @@ Layer:
             aeroway,
             ref
           FROM planet_osm_line
-          WHERE aeroway IN ('runway', 'taxiway')
-            AND ref IS NOT NULL
+          WHERE ref IS NOT NULL
+            AND (aeroway = 'taxiway'
+              OR (aeroway = 'runway'
+                AND NOT tags @> 'runway=>displaced_threshold'
+                AND NOT tags @> 'runway=>blast_pad'))
           ORDER BY CASE aeroway
             WHEN 'runway' THEN 2
             WHEN 'taxiway' THEN 1
--- a/project.mml
+++ b/project.mml
@@ -40,6 +40,7 @@ Stylesheet:
   - style/water-features.mss
   - style/road-colors-generated.mss
   - style/roads.mss
+  - style/aeroways.mss
   - style/power.mss
   - style/placenames.mss
   - style/buildings.mss
@@ -779,8 +780,7 @@ Layer:
                               AND (tags->'location' NOT IN ('underground') OR (tags->'location') IS NULL)
                               AND (tunnel NOT IN ('yes', 'building_passage') OR tunnel IS NULL)
                               AND (covered NOT IN ('yes') OR covered IS NULL))
-                              THEN railway END)),
-              (('aeroway_' || CASE WHEN aeroway IN ('runway', 'taxiway', 'helipad') THEN aeroway END))
+                              THEN railway END))
             ) AS feature
           FROM planet_osm_polygon
           WHERE highway IN ('pedestrian', 'footway', 'service', 'living_street', 'platform', 'services')
@@ -788,7 +788,6 @@ Layer:
                 AND (tags->'location' NOT IN ('underground') OR (tags->'location') IS NULL)
                 AND (tunnel NOT IN ('yes', 'building_passage') OR tunnel IS NULL)
                 AND (covered NOT IN ('yes') OR covered IS NULL))
-            OR aeroway IN ('runway', 'taxiway', 'helipad')
           ORDER BY COALESCE(layer,0), way_area desc
         ) AS highway_area_fill
     properties:
@@ -1010,6 +1009,20 @@ Layer:
         AS entrances
     properties:
       minzoom: 18
+  - id: aeroway-area-fill
+    geometry: polygon
+    <<: *extents
+    Datasource:
+      <<: *osm2pgsql
+      table: |-
+        (SELECT
+            way,
+            aeroway
+          FROM planet_osm_polygon
+          WHERE aeroway IN ('runway', 'taxiway', 'helipad')
+        ) AS aeroway_area_fill
+    properties:
+      minzoom: 11
   - id: aeroways
     geometry: linestring
     <<: *extents
@@ -1703,6 +1716,28 @@ Layer:
         ) AS power_towers
     properties:
       minzoom: 14
+  - id: aeroway-ref
+    geometry: linestring
+    <<: *extents
+    Datasource:
+      <<: *osm2pgsql
+      table: |-
+        (SELECT
+            way,
+            aeroway,
+            ref
+          FROM planet_osm_line
+          WHERE aeroway IN ('runway', 'taxiway')
+            AND ref IS NOT NULL
+          ORDER BY CASE aeroway
+            WHEN 'runway' THEN 2
+            WHEN 'taxiway' THEN 1
+            END DESC,
+            ref,
+            osm_id
+        ) AS aeroway_ref
+    properties:
+      minzoom: 15
   - id: roads-text-ref-low-zoom
     geometry: linestring
     <<: *extents
@@ -1773,13 +1808,10 @@ Layer:
                 SELECT
                     osm_id,
                     way,
-                    COALESCE(
-                      CASE WHEN highway IN ('motorway', 'trunk', 'primary', 'secondary', 'tertiary') THEN highway END,
-                      CASE WHEN aeroway IN ('runway', 'taxiway') THEN aeroway END
-                    ) AS highway,
+                    highway,
                     string_to_array(ref, ';') AS refs
                   FROM planet_osm_line
-                  WHERE (highway IN ('motorway', 'trunk', 'primary', 'secondary', 'tertiary') OR aeroway IN ('runway', 'taxiway'))
+                  WHERE highway IN ('motorway', 'trunk', 'primary', 'secondary', 'tertiary')
                     AND ref IS NOT NULL
               ) AS p) AS q
           WHERE height <= 4 AND width <= 11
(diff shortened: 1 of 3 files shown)


imagico:
Not looked at this in more detail but a quick remark on the line width progression.  What you have right now is this:

For runways:

zoom  | line width | max. ground width (in m at 0°) | ratio to previous
------|------------|--------|------
12|4|153|-
13|6|115|0.75
14|12|115|1
15|18|86|0.75
16|24|57|0.67
17|40|48|0.83
18|56|33|0.7
19|72|21|0.64

For taxiways:

zoom  | line width | max. ground width (in m at 0°)|ratio to previous
------|------------|--------|------
13|2|38|-
14|4|38|1
15|6|29|0.75
16|8|19|0.67
17|11|13|0.69
18|16|10|0.73
19|72|7|0.69

As you can see in both cases you have a non-uniform progression, for runways more so than for taxiways.

Also note that the runway width at low latitude will at low latitudes not drop to less than 21 meters - which is wider than most small informal airstrips you can find in many countries in rural areas.  Like for example here:

https://www.openstreetmap.org/#map=12/3.9992/-73.2627

Such runways would with this change even at the highest zoom level be drawn wider than the actual width and potentially covering other features in the map.

The highly variable width runways can have from less than ten meters to 50 meters or more is the main reason why i had suggested in #1853 this would be a good candidate for introducing ground unit rendering based on tagged or heuristically estimated width.  We don't have to do this but it would make it much easier to achieve a decent display of very small and very large runways likewise and at different latitudes.

mboeringa:
> Such runways would with this change even at the highest zoom level be drawn wider than the actual width **and potentially covering other features in the map.**

I don't think that is a particular problem with runways. Even if the actual width of the prepared rural runway is narrow, obstacle clearance will usually be much bigger, and no significant overlap is expected in most cases.

pnorman:
I've added some ideas from #4346

> Also note that the runway width will at low latitudes not drop to less than 21 meters - which is wider than most small informal airstrips you can find in many countries in rural areas. Like for example here:

I didn't find any examples in your link where anything was mapped beside the runway, so nothing looks out of place there. I did look at another small local airport where there was overlap

[zoom 18 27WA](https://www.openstreetmap.org/#map=18/48.32484/-122.31419)
image

I'm okay with there being overlap. The same happens with roads and other features, and it's likely to be less of an issue with runways because there are limited obstacles beside them.

> The highly variable width runways can have from less than ten meters to 50 meters or more is the main reason why i had suggested in https://github.com/gravitystorm/openstreetmap-carto/pull/1853 this would be a good candidate for introducing ground unit rendering based on tagged or heuristically estimated width. We don't have to do this but it would make it much easier to achieve a decent display of very small and very large runways likewise and at different latitudes.

I considered using `width` but decided against it, as I didn't want to add too much in one PR, and feel these changes stand alone.

imagico:
> I'm okay with there being overlap. The same happens with roads and other features

No, road width progression typically transits from larger than ground width to less than ground width at z17 or z18, in extreme cases at z19. And we depend on this being the case to reliably provide accurate geometry feedback on all the features we render at least on the highest zoom level.  In rare cases, in particular for very narrow tertiary roads (see #3925) we do not accomplish that but this is something we should fix (see #3540) and not extend to other features.

I am not aware of any line features we routinely and deliberately draw wider in their ground width at z19.

> I considered using width but decided against it, as I didn't want to add too much in one PR, and feel these changes stand alone.

In that case i suggest to limit the drawing width at z19 to less than the physical width in practically common cases.  As i see it that would match our approach with other line features, in particular other road types.

The further design changes seem unclear in motivation to me.  For example the chosen line color has less contrast towards the neutral gray fill colors we use for platforms/bridges and it harmonizes less well with the air transport symbol color.  I am not aware of any problems with the current color.  If you mean to free the current aeroway color for other purposes i could understand but if that makes sense would depend what purpose that might be.  If there are other reasons i would like to understand them.

Like in #4346 i would suggest not to mix technical refactoring, design changes and feature additions all in one PR.

pnorman:
> The further design changes seem unclear in motivation to me. For example the chosen line color has less contrast towards the neutral gray fill colors we use for platforms/bridges and it harmonizes less well with the air transport symbol color. I am not aware of any problems with the current color.

There was a transcription error, I got the hue wrong. Fixed. Neither the current colour, incorrect change, nor corrected change bear any particular similarity to `@airtransport: #8461C4` with different lightness, chroma, and hues.

imagico:
Ah, i see.  The corrected color is only marginally different from the current one so that is fine with me.

For understanding the concept of harmonizing colors:  Originally this style had a distinct violet color scheme for airports composed of the airport terminal building color, the apron color and the runway color.  When choosing the air transport symbol and label color in #1033 fitting into that scheme was a major factor:

https://davidjohnstone.net/lch-lab-colour-gradient-picker#cc99ff,e9d1ff,bbbbcc,8461c4

Although we have meanwhile almost completely eliminated that color scheme by changing both the terminal building and apron colors (which introduced problems - see https://github.com/gravitystorm/openstreetmap-carto/issues/3891#issuecomment-533173493 and also note the various issues we have with the 'major' building color) the choice and the reasoning behind it still echoes in the combination of runways and the air transport symbols, which practically in particular is visible for helipads.  The hues of the air transport symbol color and the runway color are relatively close - close enough to form a harmonic unit.  There were and are obviously other constraints.

https://davidjohnstone.net/lch-lab-colour-gradient-picker#bbbbbb,bbbbcc,babacb,8461c4,b1b9c7

The label design as shown would be a major departure from the current labeling paradigm in this style.  I am not opposed to that in principle if it makes sense overall and we have agreement that this is a workable direction but i need to study this in more detail and i would also like to understand the reasoning behind the various aspects of this design choice.

One thing in particular to keep in mind here is that the display of runway/taxiway labels and of airport labels and symbols overlaps in zoom levels and having larger and stronger labels for runways than for the airport overall could be highly irritating.

jdhoek:
> I considered using width but decided against it, as I didn't want to add too much in one PR, and feel these changes stand alone.

Understandable. Is it something you would be willing to consider as a follow up? For runways (not for taxiways) it is a marked improvement in rendering due to the large area runways can have in real life.

Thank for picking the label design from my PR. It looks great. I would recommend taking one other feature from my PR though. This is a screenshot using this PR:

Screenshot from 2022-07-17 18-13-56

Compared with #4346 (different zoom-level):

110848275-34ff5080-82ae-11eb-9ce0-885758d505eb

This runway is mapped in a manner fairly common by now: it splits the runway proper from the displaced threshold and the stopway or blast pad (only runway `05/23`, runway `09/27` doesn't have a displaced threshold). Stopways can be ignored here if you are limiting your PR to only the conceptual runway, but for displaced thresholds you might want to consider dropping the label like I did. Runways often are implemented like this in terms of scale and division:

```
>>>>>>>-------------------------------------------------------<<<<<<<
```

(`>>>` being the displaced thresholds)

Not labelling the displaced thresholds fixes runways having multiple unneeded labels, and limits the label to the main runway part. For runways this seems to work quite well.

Displaced runways are tagged like this:

```
aeroway=runway
runway=displaced_threshold
ref=05/23
```

pnorman:
> Not labelling the displaced thresholds fixes runways having multiple unneeded labels, and limits the label to the main runway part. For runways this seems to work quite well.

I agree with not labeling displaced thresholds. I'm not sure about using the on-runway marking for them, and would keep that for follow-up work.

The dashed line helps make it clearer what's going on in complex airports with lots of intersecting runways and taxiways. Without it, I find they tend to look like a mess of uninterrupted gray. [JFK](https://www.openstreetmap.org/#map=15/40.6394/-73.7738) is one example of it.

jdhoek:
> I'm not sure about using the on-runway marking for them, and would keep that for follow-up work.

Those markings work a lot better when `width` is used to render the runway line as wide as it is in reality. Leaving out the markings is a good first step.

crjwalton:
CYPT has a paved runway that is 75 feet wide and a gravel runway that is 15 feet (4.6metres) wide.  Painting both runways with the same line thickness is a bit deceiving.
[https://www.openstreetmap.org/#map=16/41.7782/-82.6771](https://www.openstreetmap.org/#map=16/41.7782/-82.6771)

Nic787:
Any update on this project?

I think it should consider width used when tagging, but if it's not tagged, it should have a default width.

PR #4609, opened 2022-07-16
User friendly improvements to get-external-data.py

### LINKED ISSUE #4590: get-external-data.py leaves file in root directory of project
When run with `--cache`, `get-external-data.py` creates files like `water-polygons-split-3857.zip` and `water-polygons-split-3857.zip.lastmod` in the root directory of the project.

These files should be in the data directory and ignored by git.

Indieberrie:
Most important improvement is `--no-update` option for the script. When set, script will not download newer data to save time if you are constantly importing stuff around and you don't need latest and greates. Useful when database needs to be regenerated, you have cache files and don't want to wait for newer version to download. 
Added ability to pass options to script in docker via env variables.
Update documentation accordingly to help users better understand how to do it.

Apologies that this took a while.

Also
Fixes #4590 (cache option saving data in place where it was called)


pnorman:
I'm probably going to use this to load empty shapefiles into the DB as part of CI

Indieberrie:
I'm glad there is immediate use case!
I forgot to do this, but changes should probably be mentioned in changelog or somewhere else so users know of new functionality.

scripts/get-external-data.py
```diff
@@ -310,73 +333,74 @@ def main():
...
-                    this_table.index()
-                    if renderuser is not None:
-                        this_table.grant_access(renderuser)
-                    this_table.replace(download.last_modified)
+                    continue
+
+
+                shutil.rmtree(workingdir, ignore_errors=True)
+                os.makedirs(workingdir, exist_ok=True)
+                if "archive" in source and source["archive"]["format"] == "zip":
```
Indieberrie, line 347: From here on, changes are mostly identation, I don't know how to make github not display those, but `git diff` has `--ignore-all-space`

Indieberrie: Aha, I found how to do it!

pnorman (changes requested):
Some exception handling issues, as well as a couple of python style issues, but otherwise looks good.

pnorman:
A re-review was requested, but https://github.com/gravitystorm/openstreetmap-carto/pull/4609/files?w=1#r922931521 is still unaddressed

Indieberrie:
Do you want me to throw exception there? I hope I explained my reasoning well.
In prior versions it didn't throw exception, and it makes sense to let it download next file, since other files might still be available.

pnorman (approved):

PR #4614, opened 2022-07-21
Test MML queries are valid in CI

pnorman:
Sometimes PRs come in with basic syntax errors in the queries in the MML file.

This could be because someone committed the wrong changes, or, less charitably, because they never tested. In either case, this is something that can be caught by automation.

Because the tables created depend on the shapefile layout, shapefiles with the same layout are required. The ones here were created by ogr2ogr, filtering all the data out so that only the columns remain.


scripts/empty_files/antarctica-icesheet-outlines-3857.zip.lastmod
```diff
@@ -0,0 +1 @@
+Thu 01 Jan 1970 12:00:00 AM UTC
```
pnorman, line 1: This (and other) files has a different format than external_data.last_modified, causing an issue. TODO: Change these to `Thu, 01 Jan 1970 00:00:00 GMT` with no newline.

sommerluk (approved):

sommerluk (approved):

PR #4616, opened 2022-07-22
Improve country label placement by splitting

### LINKED ISSUE #4241: Bad label placement at high latitudes due to coordinate calculations in mercator projection
### Expected behavior
The label for norway (Norge) renders in the middle of mainland norway, next to sweden. 
### Actual behavior
The label for norway is rendered in the middle of Svalbard, but not mainland norway.

pnorman:
By splitting MPs into polygons, taking the largest, and finding a point in 4326 we get better points for placing labels. 4326 does distort the placement, but it does so in a way that results in better results for most countries.

When monitoring for slow queries, this query does not take significant time compared to other layers. I have some upcoming work that will improve the other queries, as some of them are quite slow on low zoom.

I am making this change for countries only because
- splitting and calculating true area is only different from what st_pointonsurface does if there is a significant north/south difference in the different polygons that make up a MP. In practice, this means only countries are impacted.
- The difference for ST_PointOnSurface between 4326 and 3857 is only significant for large objects. This could in theory impact large bays, but people tend to not have strong opinions on bays, label placement isn't as crucial when the entire area is water, and we're considering removing them from the very low zooms.

Fixes #4241

For the images, left is the tile.osm.org deployment, right is this PR
image
image
image
image
image

From my tests, I have a geojson with all the country labels computed this way, because this PR is easiest to test with a full planet import.

[point_4326.zip](https://github.com/gravitystorm/openstreetmap-carto/files/9163902/point_4326.zip)

### DIFF
--- a/project.mml
+++ b/project.mml
@@ -1178,13 +1178,20 @@ Layer:
     <<: *extents
     Datasource:
       <<: *osm2pgsql
+      # By splitting MPs into polygons, finding the largest in real area, and
+      # finding a label point in 4326 we get better labeling points. 4326 does
+      # distort the placement, but it does so in a way that results in better
+      # results for most countries.
       table: |-
         (SELECT
-            ST_PointOnSurface(way) AS way,
+            (SELECT ST_Transform(ST_PointOnSurface(geom),3857)
+              FROM ST_Dump(ST_Transform(way,4326))
+              ORDER BY ST_Area(geom::geography) DESC
+              LIMIT 1) AS way,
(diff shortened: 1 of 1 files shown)


pnorman:
cc @swedneck, if you want to review this

swedneck:
Looks great!

pnorman:
Note that adding the index in #4617 makes this substantially faster to view with large databases.

sommerluk (approved):
Also, some changes in the Pacific Ocean (small island states like Kiribati), looks good.

PR #4617, opened 2022-07-22
Add an index for names at low zoom

pnorman:
One of the slowest queries at zoom 5 is for the text-poly-low-zoom
layer. This index speeds it up, and similar queries up substantially.

No rendering changes.

This index brings a text_poly_low_zoom query down from 120s-31s warm and 6.5s hot to 825ms hot, and improves the experience when developing low zooms.

Index build time on my system is 6:35, and index size is 31MB.


sommerluk (approved):
Honestly, I do not know much about databases. But I tested this, and the speed improves a lot at z5. And that's great!

PR #4622, opened 2022-07-26
Emit one index statement per line, add -0 option

### LINKED ISSUE #4621: Emit one index statement per line
By emitting one index statement per line and stripping any newlines out of the WHERE conditions, this allows use of our indexes with xargs, e.g.

```
scripts/indexes.py --fillfactor 100 | xargs -d'\n' -P20 -I_CREATE_ psql -Xqat -d flex -c "_CREATE_"
```

With newline changes, the above command will create up to 20 indexes in parallel.

Unknown: should this be standard or a command-line option? If a command-line option, should it use `\0` instead, which allows use with `xargs -0` instead of relying on GNU-specific behavior around quotes and `-d`?

pnorman:
These options make it easier to create indexes in parallel.

Fixes #4621


INSTALL.md
```diff
@@ -35,18 +35,24 @@ Disabling JIT is **essential** for use with Kosmtik and other style development
...
 
-```
+```sh
 psql -d gis -f indexes.sql
 ```
 
+The indexes can be created in parallel with
+
+```sh
+scripts/indexes.py -0 | xargs -0 -P0 -I{} psql -d gis -c "{}"
```
mboeringa, line 47: Maybe add some comments here what the different arguments to 'indexes.py' / xargs mean? E.g., it is not immediately obvious 'P0' refers to the number of parallel threads, and must be changed to the desired number, if I understood it well from some other remark I saw you make elsewhere.

scripts/indexes.py
```diff
@@ -8,63 +8,61 @@
 
 import argparse, sys, os, yaml
 
+parser = argparse.ArgumentParser(description='Generates custom index statements')
+parser.add_argument('--concurrent', dest='concurrent', help='Generate indexes CONCURRENTLY', action='store_true', default=False)
+parser.add_argument('--fillfactor', help='Custom fillfactor to use')
+parser.add_argument('--notexist', help='Use IF NOT EXISTS (requires 9.5)', action='store_true', default=False)
+parser.add_argument('--osm2pgsql', help='Include indexes normally built by osm2pgsql', action='store_true', default=False)
+parser.add_argument('--reindex', help='Rebuild existing indexes', action='store_true', default=False)
+parser.add_argument('--zero', '-0', help='Emit zero-separated index statements for use with xargs', action='store_true', default=False)
```
mboeringa, line 17: I am not to familiar with xargs, so maybe it is an xarg thing, but the line 'zero-separated index statements' is rather cryptic to me. What does it mean? Maybe spell it out a bit more?

mboeringa (changes requested):

INSTALL.md
pnorman: The text above mentions `scripts/indexes.py --help`. I don't want to go into explaining xargs here - the user can just run the above command, and if they want to modify it, they'll need to understand more about what xargs does than can be reasonably explained here.

scripts/indexes.py
pnorman: I've renamed this `--null` like xargs, and reused language from its man page.

pnorman:
Rebased because the index and CI files changed meantime

sommerluk (changes requested):
Looks good. I would just propose these two minor changes.

sommerluk (approved):

PR #4646, opened 2022-08-15
Remove name rendering for information=board

### LINKED ISSUE #4619: Drop rendering of name labels on information boards
Currently, features of type `information=board` will render with an icon and label, for example:

image
Location: https://www.openstreetmap.org/#map=19/41.44931/-71.39906

At this location, the feature:
https://www.openstreetmap.org/node/9779113378
is tagged with a name that corresponds to the title of the information board, which seems like an interesting/questionable but not unreasonable practice.  The title of the information board is supported by a link to a photograph of the object in the object's tags.

I propose that `name` no longer be rendered for information boards.  The rendering of `name` on these features adds visual clutter and encourages the adding of possibly questionable values into the `name` tag for the purpose of rendering.

wyskoj:
Fixes #4619

- Remove  name rendering of `information = board`

https://www.openstreetmap.org/node/8423226477

**Before**
download
**After**
download (1)

This is my first pull request here, so please forgive me if my method of removing the name is not the preferred way :)

### DIFF
--- a/style/amenity-points.mss
+++ b/style/amenity-points.mss
@@ -1882,15 +1882,18 @@
     text-halo-fill: @standard-halo-fill;
   }
 
-  [feature = 'tourism_information'][zoom >= 19],
-  [feature = 'tourism_information']["information"='office'][zoom >= 17] {
+  [feature = 'tourism_information'][information = 'audioguide'][zoom >= 19],
+  [feature = 'tourism_information'][information = 'guidepost'][zoom >= 19],
+  [feature = 'tourism_information'][information = 'map'][zoom >= 19],
+  [feature = 'tourism_information'][information = 'office'][zoom >= 17]
+  [feature = 'tourism_information'][information = 'tactile_map'][zoom >= 19],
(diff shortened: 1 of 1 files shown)


pnorman:
My inclination is that it would be better to restrict the text rendering to exclude information=board, rather than keep it for information=board and set the string to ''.

wyskoj:
Okay, I've refactored it such that the selector lists every other type of `tourism_information` except `information = board`. I've also ensured that other `tourism_information` nodes keep their name rendering:

Here is a `tourism = information` `information = terminal` node.
https://www.openstreetmap.org/node/8735193256
download (2)

pnorman:
> Okay, I've refactored it such that the selector lists every other type of `tourism_information` except `information = board`. I've also ensured that other `tourism_information` nodes keep their name rendering:

Is an information != board condition an option? It would keep the number of lines is the same

wyskoj:
Yes, this also works!

imagico:
Looks good, thanks.

Considering if there are other types of `information=*` that would be candidates to drop label rendering for might be a good idea.  But it makes sense to look at this carefully on a case-by-case basis.

Issue #4568, opened 2022-06-06
Stop catch-all interpretation of access tags for POI symbols


imagico:
Related to #4078.

We have meanwhile a lot of POI symbols which in `amenity-points.mss` have a rendering logic like:

```
    [access != ''][access != 'permissive'][access != 'yes'] {
      marker-opacity: 0.33;
    }
```

See #4078 for a history of how that came to be.

That means we interpret every access tag that is not `yes` or `permissive` as _restricted_.

I don't think that is a good idea or is compatible to our goal to give mappers constructive feedback.  It includes on many POI types values like:

* access=FIXME
* access=unknown
* access=official
* access=customers (for things like amenity=charging_station or amenity=shelter + shelter_type=public_transport)

Possible improvements are:

* extend the list of access tags indicating the feature to be _open_.
* interpret unclear access values the same way as the lack of an access tag (which currently always is _open_ - so this would mean moving to a list of _restricted_ values - like we do for roads).
* move towards three rendering variants in some form with _open_, _restricted_ and _unknown_ as third class.
* more restrictively choose which POI types are rendered with a separate access restricted symbolization

jeisenbe:
Re: "move towards three rendering variants in some form with open, restricted and unknown as third class.”

The current rendering of private POIs already has problems. I doubt we could create 3 classes which are clearly visible on different backgrounds and clearly distinct from one another. 

We might want to re-consider how many features have this unique private rendering.

imagico:
> I doubt we could create 3 classes which are clearly visible on different backgrounds and clearly distinct from one another.

Yes, most certainly not in the form of color variations.  But it could be considered to have symbol variations as well - the parking symbol is in principle a good candidate for that.

> We might want to re-consider how many features have this unique private rendering.

I added that to the list of possible improvements.

Issue #4469, opened 2021-09-15
Please render amenity=lounger


demsjf8:
...ideally with its own symbol. Or, as stop-gap, or permanently (to counter one aspect of "[Too many icons](https://github.com/gravitystorm/openstreetmap-carto/issues/3635)"), e.g. with the same symbol as "amenity=bench", given they serve a similar purpose, very roughly at least. And showing loungers as benches may be better than showing nothing at all.

Currently, default OSM style does not render "amenity=lounger" [at all](https://www.openstreetmap.org/node/8897597571).

OsmAnd recently made changes to [display loungers with their own symbols](https://github.com/osmandapp/OsmAnd-resources/commit/61b90120ab9002a977d05fcd4be9cd79ca446f51) going forward.

iD also shows loungers with their [own symbol](https://www.openstreetmap.org/edit?node=8897597571).

At this time, there aren't too many loungers yet (OSM wiki indicates about 1500 nodes carrying the tag). But they are being installed more and more in recent years, and mappers start [mapping them with that tag right from the start](https://www.openstreetmap.org/node/8284376683) (rather than using amenity=bench), or even re-tag loungers previously tagged as [bench to lounger](https://www.openstreetmap.org/node/8897597571/history).

So there is a clearly [accelerating trend](https://taginfo.openstreetmap.org/tags/amenity=lounger#chronology) that the number of loungers mapped as amenity=lounger is increasing since it was apparently first introduced in 2018, in my view warranting for it to be rendered in the OSM default map in one way or another (rather than being "invisible" there).

As additional input, please see the discussion regarding [OsmAnd adding support for rendering loungers](https://github.com/osmandapp/OsmAnd/issues/12251).

polarbearing:
https://wiki.openstreetmap.org/wiki/Tag:amenity%3Dlounger

These items seem to violate the OSM principle:
https://wiki.openstreetmap.org/wiki/Good_practice#Don.27t_map_temporary_events_and_temporary_features

demsjf8:
I understand those items are "approved", so I'd assume that there was some kind of consensus as to their "legality".

Are you saying that process was flawed, and it should be reopened?

Ltrlg:
> These items seem to violate the OSM principle:
> https://wiki.openstreetmap.org/wiki/Good_practice#Don.27t_map_temporary_events_and_temporary_features

Both usages documented on the wiki page are permanent enough for being accepted by this rule. The first type is just as permanent as most benches.

[...]

Issue #3187, opened 2018-04-22
Different icon for benches with backrest


Tomasz-W:
Due to Taginfo, out of 872k benches, 294k of them are tagged with backrest=yes.
https://taginfo.openstreetmap.org/tags/amenity=bench#overview

As sitting on non-backrest benches is not comfortable, I think there should be 2 different icons for 2 different bench types.

* ![pobrany plik](https://user-images.githubusercontent.com/25656654/39094321-f59df91a-462d-11e8-82a8-c14284494991.png) (current amenity=bench icon)
* ![amenity bench backrest yes](https://user-images.githubusercontent.com/25656654/39094324-096fe516-462e-11e8-85c9-0f87f7099477.png) (amenity=bench + backrest=yes icon proposal)

matkoniecz:
I am strongly against this idea. It is not important info, it is not relevant for orientation, it is too detailed for a general purpose map.

eehpcm:
Sounds good to me.

IVAN-RSM:
Nice

matkoniecz:
See https://github.com/gravitystorm/openstreetmap-carto/blob/master/CARTOGRAPHY.md#main-goals

> The aim is not however to show all or even most of the OSM data.

Unfortunately showing everything is not feasible.

eehpcm:
> Unfortunately showing everything is not feasible.

Indeed.  Last week I mapped a public space which had two picnic tables.  Sadly, they are too close together and so only one is shown by the renderer.  I can live with that.  I understand.  There's not enough room to fit in two symbols that close together.  I can wish for higher zoom levels so they can be shown, but I don't expect that to happen (my wish for lots of money didn't come true, either, but 
that doesn't stop me wishing).

This is different.  Icons are standard sizes and I think the proposed bench-with-back icon fit into that standard size.  If I'm wrong and it was a little larger, it could be redrawn to fit.  So if it's feasible to fit a bench-without-back icon it's feasible to fit a bench-with-back icon.  The hard part (once a symbol that fits the standard icon size has been created) is modifying the carto code to choose one symbol or the other depending on whether or not it has backrest=yes.

I can understand "it's too much work when we have so many other issues to deal with" but not "showing everything is not feasible" when in this case showing it is feasible.  Nor can I understand "it's too much detail" when, for some people, it may be an important consideration.

Just my opinion.  You have the final say.

kocio-pl:
[...]

PR #4647, opened 2022-08-15
Render amenity=lounger

starsep:
Render amenity=lounger, right now 3269 uses.
https://taginfo.openstreetmap.org/tags/amenity=lounger#overview
https://wiki.openstreetmap.org/wiki/Tag:amenity%3Dlounger

https://osm.org/#map=19/52.22657/20.93735
Before
image
After
image

Current icon:
image

It is my first icon and I'd appreciate feedback. I modified bench.svg

### DIFF
--- a/project.mml
+++ b/project.mml
@@ -2214,7 +2214,7 @@ Layer:
             name,
             COALESCE(
               'railway_' || CASE WHEN railway IN ('level_crossing', 'crossing') AND way_area IS NULL THEN railway END,
-              'amenity_' || CASE WHEN amenity IN ('bench', 'waste_basket', 'waste_disposal') AND way_area IS NULL THEN amenity END,
+              'amenity_' || CASE WHEN amenity IN ('bench', 'waste_basket', 'waste_disposal', 'lounger') AND way_area IS NULL THEN amenity END,
               'historic_' || CASE WHEN historic IN ('wayside_cross', 'wayside_shrine') AND way_area IS NULL THEN historic END,
               'man_made_' || CASE WHEN man_made IN ('cross') AND way_area IS NULL THEN man_made END,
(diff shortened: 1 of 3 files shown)


pnorman:
I think the cartography needs consideration, but is the icon pixel aligned?

starsep:
> I think the cartography needs consideration, but is the icon pixel aligned?

I am not sure yet what it means exactly, does this image answer this question?
image

imagico (changes requested):
Related issues: #4469, #3187, #4568.

I have two things that IMO would need to be changed:

* The symbol is not intuitive in my eyes - it is essentially the bench symbol with a diagonal line on top which is not how these typically look like and not intuitive.  This requires some more design work.
* The special rendering of access restricted loungers is not a good idea.  Because of #4568, because access restriction tagging is not common on those and because practically access restricted loungers are very rare.  Either the lounger is open to use by everyone or it is located in an access restricted area (but is open to use within that area none the less).  In the latter case the tagging and display of access restrictions should be on that area and not on the bench (and it is common practice to map this way i might add - of the 1.8M `amenity=bench` only <2000 have an `access!=yes` tag)

I think this is in principle a feature suitable for rendering here although i am a bit concerned about the idea of rendering this with only 3300 uses while we have decided not to distinctly render backrests on benches in #3187 - although these are used on 860k features.  Yes, this is a question of rendering at all vs. rendering in a distinct fashion and for benches we have the additional difficulty of needing three symbol variants.  Based on that a point of consideration might be to render `amenity=lounger` like `amenity=bench` (as suggested in #4469).  I would make this dependent on if a good distinct symbol can be developed.

pnorman (changes requested):
In #4469 we rejected rendering loungers with a dedicated icon, and in #3187 we rejected differentiating benches based on if they had a back.

I do not see that anything has changed since we rejected #4469. I could see maybe rendering them like benches.

pnorman:
Closing for reasons above. Rendering them like a bench is an option that could be discussed.

PR #4651, opened 2022-08-17
Add SRID to postgis datasources

StyXman:
This way `mapnik` does not have to make expensive queries at XML loading time just to find the layer's projection. This helps a lot shortening feedback loops while modifying the style and viewing the result in f.i. `kosmtik`.

### DIFF
--- a/project.mml
+++ b/project.mml
@@ -29,6 +29,7 @@ _parts:
     dbname: "gis"
     key_field: ""
     geometry_field: "way"
+    srid: "3857"
     extent: "-20037508,-20037508,20037508,20037508"
 
 Stylesheet:
--- a/project.mml
+++ b/project.mml
@@ -22,7 +22,7 @@ _parts:
   # Extents are used for tilemill, and don't actually make it to the generated XML
   extents: &extents
     extent: *world
-    srs-name: "900913"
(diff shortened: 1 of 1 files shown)


pnorman (approved):

StyXman:
Thanks @pnorman but I can't merge myself. Can someone else do it for me? TIA.

imagico (changes requested):
Fine by me in principle - but if we want to move from `900913` to `3857` we should also change it here:

https://github.com/gravitystorm/openstreetmap-carto/blob/5c0bbb2e011df909f9df3f3efa378d3dc64c70ce/project.mml#L25

StyXman:
I wonder is the srs's are still needed...

imagico (approved):
Thanks, looks good now

StyXman:
Thanks @imagico can you merge it? I don't have the perms :)

imagico:
> I wonder is the srs's are still needed...

I don't really know.  I remember in the past we had problems with Mapnik performing a round trip conversion Mercator -> geographic -> Mercator because of some mismatch in SRS definitions.

sommerluk:
That’s really a great speed-up. Thanks a lot!

PR #4658, opened 2022-08-21
Rework font fallbacks

### LINKED ISSUE #4644: Recommendation to include additional Arabic Naskh fallback font
Carto currently uses the Noto set of font families, which typically provide quite comprehensive coverage of Unicode characters used by a variety of languages. However, there are a handful of recently added Unicode characters which are not covered by the Noto Naskh Arabic fonts. An additional fallback font would be able to fill in these characters wherever they may occur.

For example, ࣇ is "lam with small tah above," a character added in 2020 which is not covered by Noto Naskh. It is described here https://en.wikipedia.org/wiki/Lam_with_tah_above

I have identified a few open source fonts which have more comprehensive character coverage which could be used to fill in the handful of characters which Carto currently does not render. These are each designed to fit the model of a "Naskh" font style, and examples can be seen on their respective web pages. They also seem to be relatively popular and used within a number of existing applications, so their use in Carto would likely be uncontroversial. In order of preference (based on coverage and style):

1. PakType Naskh Basic, GPL 2.0 license, available at http://paktype.sourceforge.net/index.html
This one is very well documented and is clearly informed by in depth knowledge of various Arabic-based scripts used in South Asia which may include characters which are not always supported by other fonts. I find it to be quite readable and the proportions Note that the developers have a similarly named font called PakType Naqsh, but I think that one would look more out of place alongside the Noto letters or when considered against the larger Carto style. Looking at the samples here, the "Wide" version of Naskh Basic may also be viable, particularly if readability at smaller font sizes is a concern: https://svn.code.sf.net/p/paktype/code/Fonts/Deployment/Sura-Fatiha.html

2. Scheherazade New, Open Font License, available at https://software.sil.org/scheherazade/ 
The Unicode character coverage here is impressive and Google apparently includes this one internally in their font set if that says anything about its comparability to the Noto fonts. Some of the character proportions are decidedly awkward though, and I am admittedly a bit more skeptical of this one knowing that its designers seem to be largely unfamiliar with any language which uses an Arabic-based script.

Honorable mention: Amiri Regular, Open Font License, available at https://www.amirifont.org/
I will be opinionated here and say I found this to be one of the most attractive options I came across, but unfortunately it has at least a few outstanding bugs or missing characters that disqualify it at the moment. It's a bit more ornate than the Noto fonts, but clearly very carefully considered and the sample renderings on the website look quite clear and readable. It was also funded in part by the Google Fonts project, so it has that in common with Noto. The developer(s) seem responsive enough on the GitHub page that it seems like the current problems could be fixed in the near future.

It is possible that the use of a fallback from a different family of fonts may end up making strings of text involving combinations of fonts look awkward. If that is the case, it may be worth considering just switching to one of these fonts for anything currently rendered in the Arabic Naskh style.

Please note that this is different from the issue I made a few months ago at #4547 or the older issue #2208 which concern the application of region or country specific fonts. That is an issue which is much more technically challenging than I had realized when I made it (and one I realized later is further complicated by the fact that some may prefer that Sindhi and languages using related variants of the Arabic script). With this new issue, I am just proposing what to do about representing characters which may occur anywhere on the map, within the Naskh style which Carto currently uses and which represents a reasonable compromise between the various script styles used in different parts of the world.

### LINKED ISSUE #4653: Adlam font duplicate
Currently, our font list has _Noto Sans Adlam_ (with letters that are joining up like Arabic) and then _Noto Sans Adlam Unjoined_ (as the name says, with letters that do _not_ join up).

As far as I understand, both fonts cover the same range of Unicode code points. Therefore, as we are vendoring fonts now and all fonts in our list will actually be available, the second one (_Noto Sans Adlam Unjoined_) will never be used, so obviously such duplicates should be removed.

Here is how these two fonts should render (screenshot of a PDF created with Scribus):
grafik
Above the joined version, below the unjoined version.

Wikipedia says:

> The letters are found either joined akin to Arabic or separate - the joined form is commonly used in a cursive manner; however, separate or block forms are also used as primarily for educational content.

The link specified as source by Wikipedia however does not say that.

The [Noto documentation](https://notofonts.github.io/noto-docs/specimen/NotoSansAdlam/) says:

> The unjoined variant is suitable for headlines and for educational content. The cursive variant, in which letters join the same way as in Arabic and N’Ko, is suitable for most texts. Needs software support for complex text layout (shaping). 

Apparently, [support in the shaping engine HarfBuzz was added in 2017](https://github.com/notofonts/adlam/issues/5#issuecomment-1188251406).

Obviously, this is working today in up-to-date HarfBuzz versions, as my Scribus example shows. It also works in KDE’s text editor and in LibreOffice. However, it does _not_ work in Mapnik:

_Noto Sans Adlam_ is supposed to join, but does not join in Mapnik:
joined

As comparision, here is _Noto Sans Adlam Unjoined_ (which is similar, but not identical):
unjoined

I don’t know why it doesn’t work in Mapnik.

In any case, it doesn’t make sense to keep both fonts. What should we do?
- Keep the joined font, knowing it will not actually join in current Mapnik, but maybe will join in later Mapnik versions.
- Keep the unjoined font, making sure that the rendering will not change, even if later Mapnik versions will support joining.

Note that the text in the example is 𞤪𞤢𞤤𞤵𞤆 𞤥𞤢𞤤𞤣𞤀 (diplays probably in wrong order Github): I did not find a real-world example in OSM, and therefore used this artificial example.

sommerluk:
Fixes #4653

Adds fallback fonts and simplifies the font groups.

- Treat Noto Sans Arabic like all other Sans fonts. Until now, it had a special treatment because it could redefine some Latin characters; therefore, it was placed behind. However, this is apparently something that happened in Noto Nashk Arabic (which we used in the past). Since a few years, we are using yet Noto Sans Arabic, which does not have this problem. Also, Arabic is not the only script for which Noto provides particular styles: Hebrew for instance has the Noto Rashi style, so Arabic is not special in this sense either.
- Simpifies the font subgroups to only three subgroups.
- Solves #4653 by choosing Noto Sans Adlam Unjoined. This choise is because Noto recommends the Unjoined variant for headlines, which seems to me appropiate for carthography, but this choise is subjective.

### DIFF
--- a/scripts/get-fonts.sh
+++ b/scripts/get-fonts.sh
@@ -17,8 +17,7 @@ download() {
 REGULAR_BOLD_ITALIC="NotoSans"
 
 # Fonts available in regular and bold
-REGULAR_BOLD="NotoSansAdlam
-NotoSansAdlamUnjoined
+REGULAR_BOLD="NotoSansAdlamUnjoined
 NotoSansArabicUI
 NotoSansArmenian
 NotoSansBalinese
--- a/style/fonts.mss
+++ b/style/fonts.mss
@@ -6,7 +6,14 @@ design across various scripts. We use Noto for most text and some fallback fonts
 
 By order:
 
-1. Noto Sans is available for most scripts and it is used as a first choice.
+1. Noto Sans:
+
+Noto Sans is available for most scripts and it is used as a first choice. Noto
+provides also a Serif style for quite a few scripts, Rashi for Hebrew, Nask
+for Arabic. This CartoCSS style uses the Sans style because it’s the most
+consistent and the most legible choice. Also, the other styles might redefine
+parts of the Latin script, which is not desired.
+
 Where available the UI version of the fonts – which provides tighter vertical
 metrics – is used. We intent to have all scripts of Noto in
 our list except dead (historic) scripts of whom we assume that they are not
@@ -15,9 +22,8 @@ but there are some exceptions.
 
   - Noto Sans is before all other fonts
 
-  - Noto Sans Arabic and Noto Sanc CJK JP are placed behind sans fonts because
-    commonly used signs like parenthesis or quotation marks, and their design
-    should not overwrite the standard design.
+  - Noto Sanc CJK JP is placed behind the other fonts because it redefines the
+    Latin script, and their design should not overwrite the standard design.
 
 The used CJK font covers all CJK languages, but defaults to the Japanese
 glyph style if various glyph styles are available. We have to default to one
@@ -27,21 +33,19 @@ better to default to either Chinese or Japanese. As Chinese exists in the two
 variants SC/TC, it won’t be a uniform rendering anyway. So we default to
 Japanese. However, this choice stays somewhat arbitrary and subjective.
 
-Noto provides multiple designs for the Arabic script. This style uses
-Noto Sans Arabic UI because it’s consistent with the other Sans
-fonts and legible.
+2. Noto fonts that are not available in a Sans style:
 
-2. Noto Serif is used for scripts that are not supported by Noto Sans. Currently
-this is only Tibetan: The old Noto Sans Tibetan has been renamed to Noto Serif
+The old Noto Sans Tibetan has been renamed to Noto Serif
 Tibetan in 2015, since then only Noto Serif Tibetan gets updated.
 
-3. Noto provides two variants of Emoji: Noto Color Emoji and Noto Emoji. The
+Noto provides two variants of Emoji: Noto Color Emoji and Noto Emoji. The
 colour variant is a SVG flavoured OpenType font that contains coloured emojis.
 This is not useful in cartography, so we use the “normal” monochromatic
 Noto Emoji.
 
-4. Fallback fonts. Hanazono covers almost all CJK characters, even in Unicode
-Plane 2.
+3. Fallback fonts:
+
+Hanazono covers almost all CJK characters, even in Unicode Plane 2.
 */
 
 /* Use vendored fonts. This allows for more recent versions and better coverage */
@@ -53,8 +57,8 @@ Map {
 A regular style.
 */
 @book-fonts:    "Noto Sans Regular",
-                "Noto Sans Adlam Regular",
                 "Noto Sans Adlam Unjoined Regular",
+                "Noto Sans Arabic UI Regular",
                 "Noto Sans Armenian Regular",
                 "Noto Sans Balinese Regular",
                 "Noto Sans Bamum Regular",
@@ -113,10 +117,8 @@ A regular style.
                 "Noto Sans Vai Regular",
                 "Noto Sans Yi Regular",
                 "Noto Sans CJK JP Regular",
-                "Noto Sans Arabic UI Regular",
                 
                 "Noto Serif Tibetan Regular",
-
                 "Noto Emoji Regular",
 
                 "HanaMinA Regular", "HanaMinB Regular";
@@ -126,8 +128,8 @@ A bold style is available for almost all scripts. Bold text is heavier than
 regular text and can be used for emphasis. Fallback is a regular style.
 */
 @bold-fonts:    "Noto Sans Bold",
-                "Noto Sans Adlam Bold",
                 "Noto Sans Adlam Unjoined Bold",
+                "Noto Sans Arabic UI Bold",
                 "Noto Sans Armenian Bold",
                 "Noto Sans Balinese Bold",
                 "Noto Sans Bamum Bold",
@@ -161,10 +163,8 @@ regular text and can be used for emphasis. Fallback is a regular style.
                 "Noto Sans Thaana Bold",
                 "Noto Sans Thai UI Bold",
                 "Noto Sans CJK JP Bold",
-                "Noto Sans Arabic UI Bold",
 
                 "Noto Serif Tibetan Bold",
-
                 "Noto Emoji Bold",
 
                 @book-fonts;


scripts/get-fonts.sh
```diff
@@ -131,3 +133,10 @@ mv "${FONTDIR}/static/NotoEmoji-Bold.ttf" "${FONTDIR}"
 curl --fail -A "get-fonts.sh/osm-carto" -o "${TMPDIR}/hanazono.zip" -L 'https://osdn.net/frs/redir.php?f=hanazono-font%2F68253%2Fhanazono-20170904.zip'
 
 unzip -oqq "${TMPDIR}/hanazono.zip" HanaMinA.ttf HanaMinB.ttf -d "${FONTDIR}"
+
+curl --fail -A "get-fonts.sh/osm-carto" -o "${TMPDIR}/unifont.tar.gz" -L 'https://unifoundry.com/pub/unifont/unifont-14.0.04/unifont-14.0.04.tar.gz'
+
+tar -xvf "${TMPDIR}/unifont.tar.gz" -C "${FONTDIR}/static/" unifont-14.0.04/font/precompiled/unifont-14.0.04.otf unifont-14.0.04/font/precompiled/unifont_upper-14.0.04.otf
```
pnorman, line 139: Use `--strip-components` to remove the directory from the extracted filename. This will avoid creating another temporary directory.

pnorman (changes requested):
> If we want to support such seldom characters, it makes IMO sense to re-introduce also Unifont again, which has a big coverage

Do you have some examples of glyphs in unifont that we don't have coverage of?

Do you have any suggestions of good locations to review this with?

sommerluk:
> > If we want to support such seldom characters, it makes IMO sense to re-introduce also Unifont again, which has a big coverage
> 
> Do you have some examples of glyphs in unifont that we don't have coverage of?

There is a good graphical coverage table for Noto: https://notofonts.github.io/overview/ (The site takes some seconds to load…) Everything that is red (and also everything that is green, but where we don't include the font in openstreetmap-carto because it's a dead script) is not covered.

Another question is, however, if these code points are actually used in the OSM database.

> Do you have any suggestions of good locations to review this with?

Honestly, no. I didn’t manage to make an overpass query to find some occurrences. For #4644 even the bug reporter could not come up with a real usage in OSM. And also for #4653 I did not find data. (I searched for “name:ff” in the hope to find some node that also has its name tag in this language, and maybe in Adlam script. No success.) It seems that Unicode support of Overpass practically does not exist, at least for regular expressions. Not only everything beyond the BMP does not work, but also Arabic letters within the BMP let the regular expression fail.

The tests that I've done locally have been done with mock-up data.

scripts/get-fonts.sh
sommerluk: Fixed.

sommerluk:
I've added a commit that removes the font-directory declaration, fixing #4667.

pnorman:
> I've added a commit that removes the font-directory declaration, fixing https://github.com/gravitystorm/openstreetmap-carto/issues/4667.

I'd prefer testing *why* we have #4667 to see if we have an error, and keeping it separate.

sommerluk:
Okay. I've removed this commit.

pnorman:
I'm running a query on my DB to look for names with characters outside Noto's coverage range, not counting CJK which we expect to be covered by Hanazono.

I have no concerns except the unicode, and want to validate if this is an actual issue.

bgo-eiu:
I have noticed that certain characters do not come up in taginfo either. I can add a name to the map which has a currently unsupported character if that would help

I appreciate the inclusion of my issue together with the Adlam one, but I want to reiterate that using a Nastaliq font as a fallback in Carto is not a good idea. The problem is that letter position presentations and connections are different between the two script styles (Naskh and Nastaliq, I am considering Noto Sans Arabic a Naskh font here to be clear).

Here is how lam + alef connect in a typical Naskh font like Noto Sans Arabic:
image

Here is how they connect in a Nastaliq font:
image

Now if you have a variant of lam that expects the second connection but gets the first, the result is this:
image

Here everything is Noto Sans Arabic except the lam with tah above, which has fallen back to Noto Nastaliq Urdu. These are both forms of lam and need to connect to alef the same way. This is why I use `alt_name` for letters like this until there is consistent font support, because the letter needs to connect the same way as other letters, not just be rendered itself.

Unifont has impressive character coverage, but apparently at least some of its Arabic characters have incomplete presentation forms and do not even connect cursively. (Their version of this character is also incorrect, the "tah" is supposed to be on top of the staff.)

Anyway, I just tried this with a few different fonts, and none of them render this correctly. Noto Sans Arabic is one of the few fonts which renders ݨ correctly, so it is good overall. I think PakType Naskh looks the best out of the fallback options:
image

However, I think you can leave the Arabic fallback out for now as this is still not great. Whenever I have time, I want to just try a hand at adding the missing characters to Sans Arabic and making a pull request to the Noto font repository

sommerluk:
Okay. If I understand you correctly, you think…
- There should be a fallback font, so that `U+08C7 ARABIC LETTER LAM WITH SMALL ARABIC LETTER TAH ABOVE` and potentially some other non-supported rare Arabic characters get rendered.
- The fallback font should not be Noto Nastaliq Urdu, as it has a completely different design than our standard fonts and therefore does not match them well.

I chose Noto Nastaliq Urdu because it is at least in the same font family as our default fonts, so installing the font with script was trivial, and I was hoping for at least some similarities in design, and ultimately it is just a fallback font and not a default font.

However, I get your point that the fallback font has a very different style than the default font, which is visually intrusive.

I will wait for the results of @pnorman's query before moving on.

pnorman:
> I'm running a query on my DB to look for names with characters outside Noto's coverage range, not counting CJK which we expect to be covered by Hanazono.

I did not find any characters.

sommerluk:
Okay. It might have been a bad idea of mine to put the Arabic script issue and the other changes in the same PR. That's why I removed the part about the Arabic script, which we can further discuss separately in #4644. Here thus remain the undisputed changes.

pnorman (approved):

PR #4666, opened 2022-08-28
feat: Render roller_coaster=track ✨🎢 

### LINKED ISSUE #3596: Add rendering for roller_coaster
The track should be mapped with [roller_coaster=track](https://wiki.openstreetmap.org/wiki/Tag:roller_coaster%3Dtrack). An example is [this](https://www.openstreetmap.org/way/209771601) and there are even more next door. A narrow black line should be drawn on the map. At the moment, unfortunately, many use railway=* to get the track  rendered. On the [Key:attraction](https://wiki.openstreetmap.org/wiki/Key:attraction)-page this is expressly disapproved (_Do not tag them as any railway!_). Since the track is meandering the name should not be placed here.
For the [attraction](https://wiki.openstreetmap.org/wiki/Tag:attraction%3Droller_coaster), another node or area should be drawn on which the name stands. The name should be added in black. You could also add tourism=attraction, but that would be too much importance to the roller coaster.

tjur0:
Fixes #3596

- Render roller_coaster=track

I based the design after typical rollercoaster track. I made sure that the track is not too wide, so small roller coasters and tracks next to each other render properly. While also making sure that it cannot be confused with railways.

Before
Maverick:
before

After
Maverick zoom 18:
Maverick1_Z18

Cedar Point Ohio USA zoom 16:
CedarPoint1_Z18

Gemini zoom 18:
Gemini1_Z18

Millennium Force zoom 19:
MillenniumForce2_Z19

Millennium Force zoom 19:
MillenniumForce_Z19

Millennium Force zoom 18:
MillenniumForce3_Z18


Eiim:
The tunnel rendering seems a bit too subtle. Also, in your last screenshot, some overlapping tracks conflict strangely - not sure if that's fixable or not. Definitely like the general design though, and roller coaster rendering is overdue.
image

tjur0:
Yea, I’m not quite sure how that happens, but I think this is caused if the layers don’t change and it’s the same way, see: [way/106610418](https://www.openstreetmap.org/way/106610418)
I agree that tunnels could use some work, I will fix that asap.

openstreetmap-carto.lua
```diff
@@ -227,6 +227,9 @@ local roads_info = {
         construction    = {z = 400, roads = false},
         platform        = {z = 90,  roads = false},
     },
+    roller_coaster = {
+        track           = {z = 300, roads = true},
```
pnorman, line 231: Don't add rollar_coaster to the roads table, and it doesn't need a z_order because it doesn't participate in the roads layering

openstreetmap-carto.style
```diff
@@ -41,6 +41,7 @@ node,way   railway      text         linear
 node,way   ref          text         linear
 node,way   religion     text         linear
 way        route        text         linear
+way        roller_coaster      text  linear
```
pnorman, line 44: It doesn't need its own column

project.mml
```diff
@@ -1005,6 +1005,23 @@ Layer:
...
+    geometry: linestring
+    <<: *extents
+    Datasource:
+      <<: *osm2pgsql
+      table: |-
+        (SELECT
+            way,
+            roller_coaster,
+            CASE WHEN (tunnel = 'yes' OR tunnel = 'building_passage' OR covered = 'yes') THEN 'yes' ELSE 'no' END AS tunnel,
+            name
```
pnorman, line 1018: name isn't used for this layer, is it?

project.mml
```diff
@@ -1005,6 +1005,23 @@ Layer:
...
+          FROM planet_osm_line l
```
pnorman, line 1019: Don't alias planet_osm_line as l, as you don't use the alias anywhere

project.mml
```diff
@@ -1950,6 +1967,22 @@ Layer:
...
+    Datasource:
+      <<: *osm2pgsql
+      table: |-
+        (SELECT
+            way,
+            roller_coaster,
+            name
+          FROM planet_osm_line l
+          ORDER BY
+            COALESCE(layer, 0) -- put top layered track first
```
pnorman, line 1982: The layering of a line layer and a text layer needs to be opposite.

pnorman (changes requested):
I haven't reviewed the cartography, but there's a few technical issues first.

I *think* you want to use attachments to get the layering correct for overlapping ways, but am not positive without checking.

imagico (changes requested):
I have not tested this yet, but i would suggest to reconsider rendering of the name tag.

In contrast to roads/paths a roller coaster track does not usually have a local and locally verifiable name.  The name that gets tagged will almost always be inferred from the name of the overall structure.  And for tagging the name of the overall structure `attraction=roller_coaster` is clearly the more established method.

Also note cases of roller coaster tracks tagged with railway tags at the moment show that line labels of those often work poorly because they often end up overlapping other parts of the track, badly affecting readability in general:

https://www.openstreetmap.org/#map=18/49.03685/9.05351

I would suggest also to check and make a conscious choice about the results when a way is double tagged with `roller_coaster=track` and `railway=*` - which is a common occurrence at the moment.

openstreetmap-carto.style
tjur0: could you explain how i would do this instead?

tjur0:
Thank you all for the feedback, I have removed the name rendering as I agree it looks cluttered. I chose not to render roller_coaster=track if there is a railway=*, to avoid double rendering. And I adjusted the tunnels and widths at z 20 and 19. I’m not sure how I can solve the extra table issue, so suggestions are welcome.

Example of double tagged and single tagged roller coasters:

[Toverland, The Netherlands](https://www.openstreetmap.org/#map=18/51.39683/5.98540)

Toverland1_Z18

Improved tunnels:

MillenniumForce5_Z18

Eiim:
> I chose not to render roller_coaster=track if there is a railway=*, to avoid double rendering.

As most (if not all - can anyone come up with another real reason for double tagging? Perhaps there's a semantic argument to be made that a rollercoaster track is a type of rail track?) `roller_coaster=track`+`railway=*` taggings seem to be tagging for the renderer, it would probably be more *correct* to render roller_coaster rather than railway in those cases. However, from a mapper feedback perspective, rendering the "wrong" tag might get people to switch to the "correct" tagging quicker.

tjur0:
> can anyone come up with another real reason for double tagging?

I’ve thought about this the last past days, but I can’t think of a good reason why you can double tag a rollercoaster and a railway. You are right that roller coaster track is a type of rail track, however it is not a railway. This Is because the functions are different. A roller coaster only gets build for entertainment purposes. While a railway is a road that transports people or goods, except for some exceptions like educational railways.
These different functions lead to different design requirements for the roller coaster track, for example the roller coaster trains need to wrap around the track etc.
The reason that I chose not to render roller_coaster=track is there is railway=* are:

- Less spread-out git diffs
- Changes the least amount of stuff on the map
- Gives an upgrade path for the roller coasters
- If I do it the other way around, I’m afraid the double tagging remains

The few day I have been experimenting with the correct background layer, I have added an line-cap:round background that makes shure there are no gaps between two track pieces. A problem with this is that where there is al looping, and the track makes an almost 180 degree turn the round fill-in is undesirable and looks weird. I need some more time and experimenting to get that right.

imagico:
I think prioritizing the more established tag is a reasonable choice.

For completeness (i am not necessarily suggesting to do this) - if you'd render roller coaster tracks in the road layers like railways you could render both and the railway above the roller coaster - making the double tagging explicitly visible (in a similar fashion as we render double tagging with highway and railway).  That would have the strong disadvantage to add complexity to the road layers and would have the advantage that in those few cases where there are bridges above roller coaster tracks it would layer correctly.

> The last few days I have been experimenting with the correct background layer, I have added an line-cap:round background that makes sure there are no gaps between two track pieces. A problem with this is that where there is a looping, and the track makes an almost 180 degree turn the round fill-in is undesirable and looks weird. I need some more time and experimenting to get that right.

I don't think there is a good way to do this because there is no established tagging for upside down sections of roller coaster tracks.  You would have to use fairly fragile heuristics for this - which we probably don't want to.

I think the current results or round line caps for these cases are acceptable - they might even incentivize mappers to develop more elaborate tagging in those cases that can help to produce better rendering.

tjur0:
> That would have the strong disadvantage to add complexity to the road layers

I would like to keep the roller coasters a separate from the roads. I’m satisfied with the current look of track to bridge interactions. And think it adds

Is there anything else that could be improved? Should I move the code to a separate file instead of the roads file? What do you all think of the width progression?

ppete2:
@tjur0 Thanks for you work!
How about adding support for `attraction=alpine_coaster` and `attraction=summer_toboggan` as well with the same design?
Both are very similar to `roller_coaster=track`, they are tracks for entertainment only. No sports tracks and no railways.

HolgerJeromin:
I really love to have summer_toboggan rendered. Being quite similar to roller_coaster in function the proposed visual design would not so nice match the summer attraction.
alpine_coaster is very seldom used in the database...

pnorman:
I agree with keeping these out of the roads layer. It's complex enough already.

I don't recommend adding summer coasters or toboggan runs to this PR.

pnorman (changes requested):
The casing+fill isn't working correctly - see the T junctions at 19/41.48294/-82.68539

image

I think overall the coasters are a bit big, tending to overwhelm other features. 18 and 19 feel okay.

The other problem is that a dashed fill + casing rendering doesn't work when the fill is too small to be seen. See this example from 15/8857/12226, scaled up 8x to better see the pixels.

image

Service and minor service roads appear as 2px fill + 0.55px or 0.60px casing, and are pushing it for being able to distinguish what's going on. When you add dashes to that mix, it's too much.

I see two options for fixing this. The first is the increase the minimum zoom. I recommend doing this to at least z15 anyways. If you went as far as z16, you might be able to get it working with a dashed rendering on all zooms.

The other option is to switch rendering for low zooms, as is done with roads, having a solid line for low zooms.

tjur0:
Thanks for the feedback. I have:
•	Moved the code to a separate “tourism” file.
•	Adjusted the widths from all the lower zoom levels.
•	Removed dashes from level 14 and 15.
•	Make the color a mix on level 14 and 15.

I did a lot of experimenting with the T junctions. However, I cannot seem to get a result I’m satisfied with. I want to add that the geometry of the given example is incorrectly mapped. Real transfer tracks are either completely disconnected or go with a gentle curve. Because roller coaster trains cannot turn 90 degrees that fast. As a possible solution I have sorted the tracks based on length (smaller first).

I thought about starting the rendering of roller coasters later than level 14, but if I start at 15 or 16 the roller coasters suddenly pop in. I think it’s better to have transition layers between rendering and not rendering.

Current situation:
T junction 1

Here I have removed the middle track outline. Which looks acceptable but not perfect. However this change results into the same behavior even when the layers change (see third picture)
T junctions bad

When joining the 
intersection

Do you have any ideas on how to fix this?

pnorman:
> As a possible solution I have sorted the tracks based on length (smaller first).

You probably want a Mapnik group-by, like the road layers.

tjur0:
I have implemented the Mapnik group by, and taken more inspiration from the roads layer to also add a separate casing layer, this is to prevent the gaps between the track.

rc1
rc2
rc3

tjur0:
Is there anything that could be improved, Would it be a good idea to add name rendering for attraction=rollercoaster in this PR?

style/tourism.mss
```diff
@@ -0,0 +1,87 @@
...
+    }
+    [zoom >= 17] { line-width: 4; }
+    [zoom >= 18] { line-width: 6; }
+    [zoom >= 19] { line-width: 8; }
+    [zoom >= 20] { line-width: 12; }
+  }
+}
+
+#roller-coaster[roller_coaster = 'track'][zoom >= 14] {
+  ::casing {
```
pnorman, line 42: This section shouldn't be needed, as it's now its own layer.

style/tourism.mss
```diff
@@ -0,0 +1,87 @@
...
+      line-color: @track-casing;
+      line-width: 2.5;
+    }
+    [zoom >= 17] { line-width: 4; }
+    [zoom >= 18] { line-width: 6; }
+    [zoom >= 19] { line-width: 8; }
+    [zoom >= 20] { line-width: 12; }
+  }
+
+  ::fill[zoom >= 16] {
```
pnorman, line 61: Doesn't need to be an attachment with the removal of ::casing above

project.mml
```diff
@@ -1005,6 +1006,45 @@ Layer:
...
+      table: |-
+        (SELECT
+            way,
+            COALESCE(layer,0) AS layernotnull,
+            tags->'roller_coaster' AS roller_coaster,
+            railway,
+            CASE WHEN (tunnel = 'yes' OR tunnel = 'building_passage' OR covered = 'yes') THEN 'yes' ELSE 'no' END AS tunnel,
+            CASE WHEN (bridge = 'yes' OR bridge = 'covered' OR bridge = 'viaduct') THEN 'yes' ELSE 'no' END AS bridge
+          FROM planet_osm_line
+          WHERE railway IS NULL
```
pnorman, line 1023: The WHERE clause must restrict this layer to only roller coasters.

I would also prefer not to see the railway IS NULL check, as tagging a rollercoaster as a railway is wrong generally.

project.mml
```diff
@@ -1005,6 +1006,45 @@ Layer:
...
+  - id: roller-coaster-casing
+    geometry: linestring
+    <<: *extents
+    Datasource:
+      <<: *osm2pgsql
+      table: |-
+        (SELECT
+            way,
+            COALESCE(layer,0) AS layernotnull,
+            tags->'roller_coaster' AS roller_coaster,
```
pnorman, line 1018: Not necessary, as this layer should only contain roller coasters

project.mml
```diff
@@ -1005,6 +1006,45 @@ Layer:
...
+            railway,
+            CASE WHEN (tunnel = 'yes' OR tunnel = 'building_passage' OR covered = 'yes') THEN 'yes' ELSE 'no' END AS tunnel
+          FROM planet_osm_line
+          WHERE railway IS NULL
```
pnorman, line 1041: Same as above

project.mml
```diff
@@ -1005,6 +1006,45 @@ Layer:
...
+  - id: roller-coaster
+    geometry: linestring
+    <<: *extents
+    Datasource:
+      <<: *osm2pgsql
+      table: |-
+        (SELECT
+            way,
+            COALESCE(layer,0) AS layernotnull,
+            tags->'roller_coaster' AS roller_coaster,
```
pnorman, line 1037: Same as above

project.mml
```diff
@@ -1005,6 +1006,45 @@ Layer:
...
+            railway,
```
pnorman, line 1019: This column isn't used, even right now, so can be removed.

project.mml
```diff
@@ -1005,6 +1006,45 @@ Layer:
...
```
pnorman, line 1038: Same as above

pnorman (changes requested):
There's still issues with the casing and overlaps.

Just north of Super Himalaya we have

image

For the roads layers, the solution is to have no group-by on `roads-casing`, no group-by on `roads-fill`, and group-by on `bridges`.

I'm not sure there's a good solution here. We're treating the cartography of all roller coasters are bridge-like, which I think is a reasonable choice. In particular, it is not common practice to tag elevated sections of roller coaster with bridge=yes, even when they pass over other sections of roller coaster. I looked at Cedar Point, Vancouver PNE, and all of Florida and this was true.

I'm not sure there's a good solution here. If we did things the same way as roads, it would look worse with the current data because all the crossings would go away. It also wouldn't help long-term, because this could drive bridge=yes on all the elevated ways, so we'd still be facing the same problem as the image above.

Some of the issues here are technically present with roads, but in practice complex connections of roads where all elements have bridge=yes don't exist very often, so we don't see the problem.

Given this, I think the current handling of layers and overlap is acceptable.

I still would like to see rendering start on zoom 15, not 14.

tjur0:
The image you provided is working as expected as the extension: https://www.openstreetmap.org/way/844472689 has a higher layer than the main track: https://www.openstreetmap.org/way/106372945

I want to clarify why I added the second layer.
The first layer is purely for the gaps between two sections, it’s not strictly necessary but without it the gaps look weird. Is you think that its not necessary than I’m willing to remove this layer.
If I add line-cap: round; to the second layer the round fillet overlaps the previous track. Which is unacceptable. If I remove the second casing there will never be any layering, everything will merge like it has the same layer=*.

I have tried to get the result I want with less layers. I was not able to achieve this.

The railway where check is a result of: “I would suggest also to check and make a conscious choice about the results when a way is double tagged with roller_coaster=track and railway=* - which is a common occurrence at the moment.”

Where I decided it would be the best if we do not render roller_coaster=track if there is a railway=* tag.

I still think that we should start rendering at z14 as this creates two transition layers between not rendering(z13) and fully rendering(z16). But if you insist, I’m willing to start rendering at z15.

pnorman:
> The image you provided is working as expected as the extension: https://www.openstreetmap.org/way/844472689 has a higher layer than the main track: https://www.openstreetmap.org/way/106372945

No, it's not working. If you change the tags to a highway tag, you'll see that it renders normally. However, as indicated above, it's somewhat unavoidable.

I'll review the multiple layers and end-caps once the CI failures are fixed.

I'd prefer not to have the check, but am okay with having it in.

project.mml
```diff
@@ -1005,6 +1006,39 @@ Layer:
...
+    <<: *extents
+    Datasource:
+      <<: *osm2pgsql
+      table: |-
+        (SELECT
+            way,
+            CASE WHEN (tunnel = 'yes' OR tunnel = 'building_passage' OR covered = 'yes' OR indoor = 'yes') THEN 'yes' ELSE 'no' END AS tunnel,
+            CASE WHEN (bridge = 'yes' OR bridge = 'covered' OR bridge = 'viaduct') THEN 'yes' ELSE 'no' END AS bridge
+          FROM planet_osm_line
+          WHERE tags->'roller_coaster' = 'track' AND railway IS NULL
```
pnorman, line 1020:
```suggestion
          WHERE tags @> 'roller_coaster=>track' AND railway IS NULL
```

As per [SQL style guidelines](https://github.com/gravitystorm/openstreetmap-carto/blob/master/CONTRIBUTING.md#sql-style-guidelines)

> To check if a tag is in the tags hstore, use tags @> 'foo=>bar', relying on automatic conversion from text to hstore.

pnorman (changes requested):
Because stylistically all roller coaster tracks are like bridges, I think I see what's going on.

`#roller-coaster-casing::casing` is like roads-casing, `#roller-coaster-casing::fill` is like roads-fill, `#roller-coaster::casing` is like `#bridges::casing`, and `#roller-coaster::fill` is like `#bridges::fill`.

This is the approach of drawing everything twice - once without group-by and with round end-caps, and once with group-by and without round end-caps. The first pass of drawing exists solely to get the ends right.

This looks good to me, except I'd like different layer names so there's no confusion caused by `#roller-coaster-casing::fill` being named both casing and fill.

I'd also like a review by someone else familiar with writing roads layers, as they're complicated. Maybe @matkoniecz?

imagico (changes requested):
Design wise i think this is fine - the whole dashing concept is rather tricky and not possible to do really well with the builtin means of mapnik without excessive geometry processing.  But i am fine with having this as is to see how it works out practically.  The only real issue i see is that the bridge casing is hard to read.  I suggest using a black bridge casing (which is standard in this style in general for bridges) and maybe brighten the dark part of the line signature for bridges to limit the overall weight.

What does not work is the layering.  If you want proper layering of bridges and tunnels you will need to use the same layer stack as with the roads (i.e. separate bridges and tunnels layers) or duplicate geometries in SQL to get the right drawing order.

What you get right now is this:

roller_coaster_layering

left is with layer tags, right is without.  Bottom is the same with highway=service for comparison.  With the layer tags it is nearly right - except for the bridge casing, which is not correctly layered.

Ultimately i think there is not going to be maintainability gain of separating the roller coaster rendering out of the road layers if you want to have correct layering because you will end up duplicating the road layer stack in full.  And that adds more complexity than extending the existing road layers with roller coaster lines.  But that is my opinion - if others differ in that assessment i am fine with keeping separate layers.

matkoniecz:
Is left one, without `layer` tags even a correct mapping?

imagico:
I don't know - but we have for ages rendered bridges/tunnels above/below normal roads independent of the layer tag - it would be inconsistent and confusing if we do this differently for roller coaster tracks.

In any case - my key point was that layering does not work correctly even with the layer tags.

tjur0:
Thank you for the feedback, I have:
•	Changed the bridge colour to black.
•	Refactored the tourism file to reduce duplicated code.
•	Change name of layer `roller-coaster-casing` to `roller-coaster-gap-fill`.

imagico (changes requested):
This is much clearer now - both in code and in rendering.

And it is actually a fairly interesting approach to the road rendering problem.  You could even extend it to work in the case i showed by modifying the `layernotnull` with something like:

```
        (SELECT
            way,
            layernotnull + CASE WHEN tunnel = 'yes' THEN -1000 WHEN bridge = 'yes' THEN 1000 ELSE 0 END AS layernotnull,
            tunnel,
            bridge
          FROM 
            (SELECT
                way,
                COALESCE(layer,0) AS layernotnull,
                CASE WHEN (tunnel = 'yes' OR tunnel = 'building_passage' OR covered = 'yes' OR tags->'indoor' = 'yes') THEN 'yes' ELSE 'no' END AS tunnel,
                CASE WHEN (bridge = 'yes' OR bridge = 'covered' OR bridge = 'viaduct') THEN 'yes' ELSE 'no' END AS bridge
              FROM planet_osm_line
              WHERE tags @> 'roller_coaster=>track' AND railway IS NULL) AS _
          ORDER BY
            layernotnull -- put bottom layered track first
        ) AS roller_coaster
```

However, it ultimately does not really correctly interpret the layer tag correctly according to the meaning it has by convention in the OSM community.

See here:

roller_coaster_layering2

The curved segments have a layer tag (the bridge `layer=1`, the non-bridge one `layer=-1`), the other ones have not.  This is typically considered perfectly correct tagging.  For comparison here how it renders for roads:

roller_coaster_layering2a

I am not saying this needs to be changed - if others find this to be ok i am not going to object.  But i wanted to point out that the chosen method - even if it is fairly neat - does not represent a shortcut to get the same layering logic we have for roads.

tjur0:
Yes, this is not a replacement for the way we render roads. Because roller coaster track are not roads. nevertheless, the rendering of higher layer=* tags over lower layer=* is very necessary to get the look of the coaster right. Look at the loops in this coaster:

condor
https://www.openstreetmap.org/#map=19/52.44040/5.76195

While the tunnel is useful for indoor coaster and covered sections (like in most stations). I added the bridge rendering more for mapper feedback than anything else. And I would like to see how the usage of the bridge tag evolves overtime. With this render method all the variation renders like bridges.

I also made sure that the bridge rendering and the tunnel rendering can work together. As I can imagine an indoor coaster having a bridge.

matkoniecz:
I am just looking at images: but this rendering seems extremely prominent. Maybe a bit subtler would be preferable?

imagico:
@tjur0 - i see what you mean - but your example is incorrectly tagged - the loop is evidently a bridge and should be tagged as such - and then you will get the desired rendering, even with road-like layering semantics.

While i concur that roller coaster tracks are not roads so it is acceptable to render with differences in rendering logic, the meaning of secondary tags like `bridge=*` `tunnel=*` and `layer=*` is the same as for roads and we should not indicate to mappers otherwise.  If a roller coaster track crosses over another without a bridge or tunnel tag on either that is a level crossing - independent of the layer tags applied.

The reason why i put so much focus on this is because on the matter of layer tag interpretation this style has - probably more than on any other aspect of tagging - an obligation to support consistent tagging.  Meaning of the layer tag is a highly abstract topic and without clear and consistent feedback from this style tagging would likely quickly deteriorate.

I think you can get the road-like layering semantics w.r.t. the aspects discussed so far with something like:

```
        (SELECT
            way,
            layernotnull,
            CASE WHEN tunnel = 'yes' THEN layernotnull - 1000 WHEN bridge = 'yes' THEN layernotnull + 1000 ELSE 0 END AS grouplayer,
            tunnel,
            bridge
          FROM 
            (SELECT
                way,
                COALESCE(layer,0) AS layernotnull,
                CASE WHEN (tunnel = 'yes' OR tunnel = 'building_passage' OR covered = 'yes') THEN 'yes' ELSE 'no' END AS tunnel,
                CASE WHEN (bridge = 'yes' OR bridge = 'covered' OR bridge = 'viaduct') THEN 'yes' ELSE 'no' END AS bridge
              FROM planet_osm_line
              WHERE tags @> 'roller_coaster=>track' AND railway IS NULL) AS _
          ORDER BY
            grouplayer,
            layernotnull
        ) AS roller_coaster
```

imagico:
@matkoniecz - i think it is fine in general - especially compared to the monorail rendering, which is occasionally used right now as tagging for the renderer.

The bridge rendering with the changed black casing is however - as already hinted - somewhat too heavy.  I'd suggest to narrow the bridge casing width (half of the current seems to work quite well, maybe a bit more at the highest zoom levels) and limit the bridge casing display to z16+.

tjur0:
I have adjusted the bridge casing widths.

On roads a bridge is used when there is extra structure to hold up the road. With roller coasters this structure is always there and therefore you could say that roller_coaster=track has an implied bridge=yes.

I’m currently content with the rendering. But to comply with your statements above an option could be to remove the bridge=yes completely or always show it. What do the other maintainers think of this?

Like pnorman pointed out earlier there is not a real solution for this problem it is going to be a compromise anyhow. I feel like I have balanced the meaning of the secondary tags and rendered the data as is as good as possible.

In my opinion it is wrong to render covered sections always below normal sections. Let’s say there is a covered lift hill high up in the air, the track below the covered lift hill is not covered because of the height difference. And therefore, should be below the tunnel section.

But I cloud do something like this for the render order:
•	Tunnel: Layer * 3
•	Normal: Layer * 3 + 1
•	Bridge: Layer * 3 + 2
Then it would respect the layer and tunnel/bridge render order.

01110229:
Any updates ongoing?

imagico (commented):
My view here is essentially unchanged since i last reviewed this.  The bridge casing seems fine now - because the signature of the roller coaster track is fairly heavy it is a bit difficult to recognize at times, but this is something that can be adjusted later as well.

I still think it would be best if the layering logic matches that of roads.  After having looked at various roller coaster tracks in the database it is clear that there is no consensus among mappers about using layer/bridge/tunnel tagging on these with different semantics than for roads - hence providing feedback on this that is consistent with that for roads/railways seems prudent.  If this is the goal, putting roller coaster tracks into the road layers would probably be more convenient than separate layers - even if interleaving between roads and roller coaster tracks is rarely necessary.

But this is not a big deal for me - if others are fine with the current approach i won't object.

I also still think the two things i remarked on the code (interpreting `bridge` and `tunnel` the same way as for roads, using `@bridge-casing` instead of a color literal) would be good to change.  Yes,cases where `bridge=*` and `covered=yes` are used in combination are an open question - but this is the case for roads as well, nothing specific to roller coasters IMO.

But likewise - if others are fine with the change as is i would be ok with that.

tjur0:
I'm happy to consider implementing changes to the rendering if there is a consensus that something needs to be changed.

In my opinion, adding road logic to the mapping of roller coaster tracks could be problematic since it may require the use of bridge=yes, which could potentially create confusion when it comes to verifying the actual bridge structure. As in contrast with roads there is not a clear point where an overlapping track “bridge” starts and ends. That in turn may incentivize adding bridge=yes to all parts of the track, which could make bridge=yes on roller coaster track meaningless.

I understand the importance of finding a solution that works well for everyone. I believe that the current rendering system is flexible enough to accommodate the forming of a consensus. Not having the extra tag bridge=* and covered=yes and adding the layering heuristics could inhibit the formation of a new consensus.

I would like to hear some more opinions on this matter.

arminkollascheck:
Thanks you for working on this topic.

I have the same opinion as you @tjur0 . On roller coasters you should not have to mark everthing as bridge=yes (what makes it indead meaningless). To add bridge=yes only to parts where it passing over a other roller coaster track is even more problematic. As you said, there would not be a clear end or start of this bridge, a bad idea. Using just layers is better imo.

@imagico 
In the rules of Mapping it reads "Bridge=* tags should not be used just because tracks overlap. There should be an actual bridge structure underneath the track." (https://wiki.openstreetmap.org/wiki/Tag:roller_coaster%3Dtrack)

But as you also said, since many mappers map for the openstreetmap carto map (even it should not be goal to map for a map layout), there is much chaos in mapping roller coasters like having bridge=yes everywhere or railway=..... It even gets changed many times on the same roller coaster over the time.

imagico:
My (certainly limited) studying of mapping of roller coaster tracks indicates that the semantics of `bridge=yes` and `layer=*` of roller coaster tracks matches that of roads where it is applied, the main difference seems to be that it is applied less (i.e. there are quite a lot of tracks which completely lack this kind of information).  Yes, there are a few cases where mappers either tag a whole track with `bridge=yes` because they are assuming that tag to mean _feature is above ground level_ rather than an actual bridge (in the sense of how the tag is used on roads) or cases where no bridge tagging is used at all, not even for the actual places where the track bridges over itself or a road or something else.  But these are clearly in the minority or represent cases of coarse, preliminary, non-detailed mapping.

The key here is to consider our goal to support mappers in consistent use of tags.  For that the bridge vs. non-bridge difference is not an issue, both this change and the traditional road rendering provide intuitive feedback.  The handling of `layer=*` on non-bridge features is the problem here.  On that this change would provide fundamentally different feedback indicating different meaning of the `layer=*` tag on roller coasters.  Mapper consensus on `layer=*` on non-bridge features is that it describes the relative vertical relationship between *non-connecting* features.  Where features connect on a shared node different layer values are not considered to disconnect the features.  For roads this is pretty fundamental to allow data users to correctly interpret the tagging and most map styles rely on this.

tjur0:
I can’t wrap my head around that you keep comparing roller coaster tracks to roads. Maybe that is my mistake from implementing many road features.

From the start I took the water slide rendering as inspiration because IMO the closes thing to roller coaster tracks are water slides. And those also layer correctly without additional tags.
https://www.openstreetmap.org/way/936392687
I do not see the issue of having this different implementation of the layer tag.

imagico:
> I can’t wrap my head around that you keep comparing roller coaster tracks to roads.

I can see that and therefore i tried to explain in detail how - as far as i can see - the semantics of layer/bridge/tunnel tags on roller coaster tracks and on roads are the same for mappers and that there does not seem to be a consensus on a different meaning of those on roller coaster tracks.  I am open to the possibility that this is a wrong impression but i would need evidence for that.  Someone would need to explain to me the distinct meaning of those tags for roller coaster tracks.

> From the start I took the water slide rendering as inspiration because IMO the closes thing to roller coaster tracks are water slides.

I see.  Two things to keep in mind here:

* Water slides were added in #3346 without consensus (and initially even without any explicit layering - see #3413).  There was no conscious decision back then regarding how this supports or goes against our goal of providing constructive mapper feedback.  If such a discussion had happened back then the arguments brought up here could well have already been brought up back then.
* The situation with water slides is a bit different because those contain junctions/switches much less frequently than roller coaster tracks (in other words: these tend to be topologically simple).

Again - i am not categorically against the approach chosen, my aim is for us to make a conscious and well informed choice here taking the different arguments into account.

arminkollascheck:
I looked at some parks in the US and Europe and I find no evidence that the bridge tag is often used for roller coasters/crossing roller coaster tracks. Most of the time the layers get used. Obviouly this is not a complete overlook.

https://www.openstreetmap.org/edit#map=18/38.51458/-90.67468 [no bridge tag, different layers]
https://www.openstreetmap.org/edit#map=17/42.03964/-72.61536 [no bridge tag, different layers]
https://www.openstreetmap.org/edit#map=17/40.13800/-74.43518 [mapped as monorail]
https://www.openstreetmap.org/edit#map=18/34.42291/-118.59733 [no bridge tag, but also no layers]
https://www.openstreetmap.org/edit#map=16/48.2656/7.7222 [some mapped as railway, some not but with layers]
https://www.openstreetmap.org/edit#map=16/53.0238/9.8836 [bridge tag gets used]
https://www.openstreetmap.org/edit#map=16/53.0238/9.8836 [as railway, only tunnel tag gets used]
https://www.openstreetmap.org/edit#map=17/52.98740/-1.89116 [as railway, only layers, sometimes bridge tag]

pnorman (approved):

jespervd2:
Is this mergeable? It looks so cool!😁

arminkollascheck:
Also wonder what is needed now for a merge? Would be cool to have it included.

PR #4672, opened 2022-09-01
Make font-directory statement effectife

### LINKED ISSUE #4667: font-directory declaration ineffective?
We have in `fonts.mss` the following line of code:
```cartocss
font-directory: url('fonts');
```
This line was added when we started vendoring fonts.

I wonder if this line works at all. When calling CartoCSS, the resulting XML does not contain any data about the font directory, nevertheless everything works fine and Kosmtik does not complain about any missing fonts. It doesn't complain either when I remove the `font-directory` statement completely. But when the font directory name is changed to `fonts2` (both, the `font-directory` declaration in `fonts.mss` and the name of the actual font directory), it doesn’t work any more and Kosmtik complains now about missing fonts.

Is `font-directory` completely ineffective, but Mapnik has hard-coded to search for additional fonts always in a subdirectory named `fonts`?

If so, should we drop `font-directory` and replace it by a comment explaining the situation?

sommerluk:
Fixes #4667

### DIFF
--- a/style/fonts.mss
+++ b/style/fonts.mss
@@ -45,7 +45,9 @@ Plane 2.
 */
 
 /* Use vendored fonts. This allows for more recent versions and better coverage */
-font-directory: url('fonts');
+Map {
+  font-directory: url('fonts');
+}
 
 /*
 A regular style.


pnorman (approved):
Tested with kosmtik and verified with cartocss that `font-directory="fonts"` appears on the map element.

PR #4675, opened 2022-09-04
Avoid repetition in project.mml

sommerluk:
- “Do not repeat yourself.” Use references instead of repetition in project.mml. If generating a JSON file out of project.mml the generated JSON file should be identical before and after this PR.

### DIFF
--- a/project.mml
+++ b/project.mml
@@ -1,28 +1,16 @@
-scale: 1
-metatile: 2
-name: OpenStreetMap Carto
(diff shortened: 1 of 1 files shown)


pnorman (changes requested):
Can you move _parts higher up, then define the srs there, and use aliases elsewhere?

sommerluk:
New commit added, moving _parts higher and defining everything that is reused there-in.

PR #4685, opened 2022-09-13
YAML highlighting on GitHub webpage for CartoCSS MML files

sommerluk:
- YAML highlighting on GitHub webpage for CartoCSS MML files

It is possible to [force a specific syntax highlighting](https://github.com/github/linguist/blob/master/docs/overrides.md) for specific file types on the GitHub web page. So we can make `*.mml` files, which are currently not highlighted in our repository, get YAML highlighting: `project.mml`

Note that this PR will not have an immediate effect. [Only after the _next_ commit that touches the MML file, the syntax highlighting will be available.](https://webapps.stackexchange.com/questions/31654/force-github-syntax-highlighting-language-on-source-files#comment139871_108676) (I've verified this in my fork before making this PR.)

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -6,4 +6,4 @@
 *.py text eol=lf
 
 # YAML highlighting on GitHub webpage for CartoCSS MML files.
-*.mml linguist-language=YAML
+*.mml linguist-language=YAML linguist-detectable
--- a/.gitattributes
+++ b/.gitattributes
@@ -4,3 +4,6 @@
 # Declare files that will always have LF line endings on checkout.
 *.sh text eol=lf
 *.py text eol=lf
+
+# YAML highlighting on GitHub webpage for CartoCSS MML files.
+*.mml linguist-language=YAML


.gitattributes
```diff
@@ -4,3 +4,6 @@
 # Declare files that will always have LF line endings on checkout.
 *.sh text eol=lf
 *.py text eol=lf
+
+# YAML highlighting on GitHub webpage for CartoCSS MML files.
+*.mml linguist-language=YAML
```
HolgerJeromin, line 9:
```suggestion
*.mml linguist-language=YAML linguist-detectable
```
https://github.com/github/linguist/blob/master/docs/overrides.md#detectable

sommerluk: This will try to make MML files count for the language statistics on GitHub. I've tested it, but it doesn’t work. The [documentation](https://github.com/github/linguist/blob/master/docs/overrides.md#detectable) says:

> Languages that are not yet mentioned in languages.yml will not be included in the language statistics, even if you specify something like *.mycola linguist-language=MyCoolLang linguist-detectable in the .gitattributes file.

I suppose it is unlikely that MML will get support in the future. Therefore, I would leave this PR as-is for now.

sommerluk: This will try to make MML files count for the language statistics on GitHub. I've tested it, but it doesn’t work. The [documentation](https://github.com/github/linguist/blob/master/docs/overrides.md#detectable) says:

> Languages that are not yet mentioned in languages.yml will not be included in the language statistics, even if you specify something like *.mycola linguist-language=MyCoolLang linguist-detectable in the .gitattributes file.

I suppose it is unlikely that MML will get support in the future. Therefore, I would leave this PR as-is for now.

pnorman: We should still set `linguist-detectable`. We're not setting the language to MML, we're setting it to YAML.

pnorman (changes requested):

.gitattributes
sommerluk: I've added a corresponding commit.

pnorman (approved):

HolgerJeromin:
Nice:
image
https://github.com/gravitystorm/openstreetmap-carto

sommerluk:
That's great.

Issue #4322, opened 2021-02-07
Stop rendering tracktype for highway=track


imagico:
This is meant to supersede  #2513.

`tracktype` on tracks (https://wiki.openstreetmap.org/wiki/Key:tracktype) has always been a poorly defined tagging scheme with a very climate specific documentation and is widely misused as a generic importance rating for tracks.

At the same time rendering it through different dashing patters adds a lot of visual noise to the map.

I suggest we stop rendering tracktype on tracks and maybe instead add rendering of paved/unpaved surface (which would mean going for a 6 class system to a 3 class system).  This would be related to #110/#3399/#4137.  And it could make rendering of tracks consistent with footway/cycleway/path thereby significantly improving intuitive readability of the map.

Current use numbers are 6.8M cases of tracktype=* on highway=track and 4.6M cases of surface=*.

TiliaJ:
I don't see poor definition, neither climate-specific characteristics.

Yes, it can be misused as an importance rating.

Tracktype is used in combination with key:surface and key:trail_visibility. I encounter the combination of tracktype=2 & surface=sand quite frequently and this mis-tagging is already in the JOSM validation rules set.

A proper definition of tracktype <> importance PLUS a proper definition of allowed combinations (and subsequent validation) with surface and trail_visibility should be the preferrable solution

kocio-pl:
I think it makes sense - we could also use more patterns for admin level boundaries. Paved/unpaved classification rendering would be very useful here.

pnorman:
I don't see it misused as importance either. I agree with removing the many dash patterns as they're confusing, and I'm the one who introduced them originally.

imagico:
Regarding misuse as a generic importance rating - the only verifiable criterion for tagging tracktype that is either known or documented is 'softness' of the ground - which is a criterion that does not apply at all for much of the earth surface, in particular in dry climate of all kinds.  In those areas (and partly also in regions where the softness criterion would be applicable) mappers therefore often apply the tracktype rating as a generic importance rating to indicate the cartographic importance of the track - which will often be based on a local and subjective rating of how much the track seems to be used or how the 'quality' of the track is rated (f.s.v.o. quality that is subjective to the mapper).  The source of confusion might have been that i did not clarify that with 'importance' i here meant cartographic importance and not some real world physical/functional importance.

But ultimately it of course does not matter that much - there seems to be broad agreement that rendering the paved/unpaved distinction instead of tracktype would be a desirable change.

[...]

PR #4692, opened 2022-09-19
Introduce sac_scale and trail_visibility on highway=path

### LINKED ISSUE #1500: Stop rendering extreme mountain paths without giving indication of difficulty
Some mappers use highway=path instead of via_ferrata ( https://github.com/gravitystorm/openstreetmap-carto/issues/156 ) to make the paths they like visible.

As long as via_ferrata is not rendered and ferrata/path difficulty not displayed I believe it would be good not to render 
* highway=path with via_ferrata=yes and no further difficulty indicator
* paths with via_ferrata_scale>=2
* paths with sac_scale>=demanding_mountain_hiking
* http://www.klettersteig.de/klettersteig/via_attrezzata_rino_pisetta/760

yvecai:
This is a first step to fix #1500 , attempted to be the least disturbing as possible from the current style. From this, future change can be made and (heavily) discussed, either more elaborate (fine-grain tuning for hikers), or less (don't render at all). 

Introduce tags sac_scale and trail_visibility, and draws an arbitrary limit on what an average walker would expect.
```
                CASE WHEN (tags->'sac_scale' IN ('demanding_mountain_hiking', 'alpine_hiking', 'demanding_alpine_hiking', 'difficult_alpine_hiking')
                  OR
                  tags->'trail_visibility' IN ('intermediate','bad','horrible','no')) THEN 'demanding'
                  ELSE 'normal'
                END AS int_difficulty,
```
The style change is 'naive' and aim at reducing visibility of most 'demanding' path by introducing wide dash and reduced background visibility:
```
          [int_difficulty = 'demanding'] {background/line-opacity: 0.1;}
          ...
            [int_difficulty = 'demanding'] {line/line-dasharray: 1,12;}
```
This style is extended to bridleways and cycleways in order to keep in line with the current stylesheet.

**Test rendering with comparison with surface tags (null, paved, unpaved), bridleway, cycleway, and access=no:**
demanding_path

### DIFF
--- a/style/roads.mss
+++ b/style/roads.mss
@@ -2285,10 +2285,12 @@ tertiary is rendered from z10 and is not included in osm_planet_roads. */
           background/line-join: round;
           background/line-width: @bridleway-width-z15 + 2 * @paths-background-width;
           background/line-opacity: 0.4;
+          [int_difficulty = 'demanding'] {background/line-opacity: 0.1;}
         }
         line/line-color: @bridleway-fill;
         [access = 'no'] { line/line-color: @bridleway-fill-noaccess; }
         line/line-dasharray: 4,2;
+        [int_difficulty = 'demanding'] {line/line-dasharray: 1,12;}
         line/line-width: @bridleway-width-z13;
         [zoom >= 15] { line/line-width: @bridleway-width-z15; }
         #tunnels {
@@ -2308,6 +2310,7 @@ tertiary is rendered from z10 and is not included in osm_planet_roads. */
           background/line-join: round;
           background/line-width: @footway-width-z15 + 2 * @paths-background-width;
           background/line-opacity: 0.4;
+          [int_difficulty = 'demanding'] {background/line-opacity: 0.1;}
           [zoom >= 16] {
             background/line-width: @footway-width-z16 + 2 * @paths-background-width;
           }
@@ -2321,18 +2324,22 @@ tertiary is rendered from z10 and is not included in osm_planet_roads. */
         line/line-color: @footway-fill;
         [access = 'no'] { line/line-color: @footway-fill-noaccess; }
         line/line-dasharray: 1,3;
+        [int_difficulty = 'demanding'] {line/line-dasharray: 1,12;}
         line/line-join: round;
         line/line-cap: round;
         line/line-width: @footway-width-z14;
         [zoom >= 15][int_surface = 'paved'] {
           line/line-dasharray: 2,3.5;
+          [int_difficulty = 'demanding'] {line/line-dasharray: 1,12;}
           line/line-width: @footway-width-z15;
           [zoom >= 16] {
             line/line-dasharray: 3,3.5;
+            [int_difficulty = 'demanding'] {line/line-dasharray: 1,12;}
             line/line-width: @footway-width-z16;
           }
           [zoom >= 17] {
             line/line-dasharray: 3,3;
+            [int_difficulty = 'demanding'] {line/line-dasharray: 1,12;}
           }
           [zoom >= 18] {
             line/line-width: @footway-width-z18;
@@ -2345,11 +2352,13 @@ tertiary is rendered from z10 and is not included in osm_planet_roads. */
           line/line-color: @footway-fill;
           [access = 'no'] { line/line-color: @footway-fill-noaccess; }
           line/line-dasharray: 1,3,2,4;
+          [int_difficulty = 'demanding'] {line/line-dasharray: 1,12;}
           line/line-join: round;
           line/line-cap: round;
           line/line-width: @footway-width-z15;
           [zoom >= 16] {
             line/line-dasharray: 1,4,2,3;
+            [int_difficulty = 'demanding'] {line/line-dasharray: 1,12;}
             line/line-width: @footway-width-z16;
           }
           [zoom >= 18] {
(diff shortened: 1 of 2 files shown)


style/roads.mss
```diff
@@ -2321,18 +2324,22 @@ tertiary is rendered from z10 and is not included in osm_planet_roads. */
         line/line-color: @footway-fill;
         [access = 'no'] { line/line-color: @footway-fill-noaccess; }
         line/line-dasharray: 1,3;
+        [int_difficulty = 'demanding'] {line/line-dasharray: 1,12;}
         line/line-join: round;
         line/line-cap: round;
         line/line-width: @footway-width-z14;
         [zoom >= 15][int_surface = 'paved'] {
           line/line-dasharray: 2,3.5;
+          [int_difficulty = 'demanding'] {line/line-dasharray: 1,12;}
```
imagico, line 2333: Isn't this (and subsequent rules added) unnecessary?

imagico (changes requested):
Thanks for working on improving rendering of paths.

First of all a technical remark - please always post sample images in native resolution reflecting what the map user will see if the change is merged and released.

The change introduces a new, fourth line signature for paths in addition to the existing surface=paved/unpaved/unknown stylings.  This is based on having any of eight tags present on the way:

* `sac_scale=demanding_mountain_hiking`
* `sac_scale=alpine_hiking`
* `sac_scale=demanding_alpine_hiking`
* `sac_scale=difficult_alpine_hiking`
* `trail_visibility=intermediate`
* `trail_visibility=bad`
* `trail_visibility=horrible`
* `trail_visibility=no`

It does so in a way that overrides the surface classification, so a path with `surface=concrete`, `sac_scale=demanding_mountain_hiking` (yes, such exist in the database) will be rendered according to the sac_scale and not (like right now) according to the surface.

The rendering of this starts at z14 (surface distinction starts at z15).

I will comment separately on the idea this change implements and how it implements that.


## The implementation

The newly introduced line signature is almost always unreadable at z14 and usually very poorly readable at higher zoom levels.  It aggrevates the existing problems we have with #1748, #1793, #1765.

Having any of eight poorly defined and not very consistently used tags override existing distinction of paved and unpaved paths (that is taggged by mappers with a very high level of consistency) does not seem a good idea.


## The idea of the change

What the change here essentially tries to do is to interpret eight tags which are all rather poorly defined and documented and mostly not used very consistently as code for _this path is to be considered as special in some form_.  That this special nature of the path is meant to represent some form of difficulty for the user is not intuitively visible.  The chosen design is more communicating some sort of _path light_ or something between there being a path and there not being a path.  This is indeed a frequent use case of the tags in question, on the extreme end for example a pathless walking route on rocks or on a beach, but there is clearly no consensus among mappers that this is a valid use case for highway=path.

I already commented on the consistency and definition of the sac_scale and trail_visibility tags in #1500 and elsewhere.  The idea that collectively these eight tags can be aggregated as being meant to indicate some sort of difficulty or as being discouraged to be used by the mapper is sound (though of course subjective and non-verifiable) but there is clearly also no consensus among mappers that this is the main meaning of these tags.

I am kind of torn between supporting or rejecting this change.  On the one hand it is clearly not in support of the goal of this style to support mappers in consistent use of tags.  And it is also not very useful for the target map user because while, as explained, the interpreted tags indeed typically indicate some sort of difficulty or being problematic to use in some form these tags are in practical use not even remotely the generally accepted form to indicate any sort of difficulty of a path.

On the other hand it seems that all arguments and reasoning about the flaws of these tagging ideas have not substantially diminished the desire of mappers for sac_scale/trail_visibility to have meaning (independent of how much wishful thinking that is).  So maybe the most constructive way forward is to actually support degrading the eight tags this change wants to interpret to being synonyms for `render_in_the_map_as_path_light=yes` to help creating space for mappers to develop tagging concepts to document specific, well defined and locally verifiable risks/difficulties/etc. of paths.  This is a long game of course, it could easily take a decade for this to happen.  But we kind of have precedents for that - see #4322.

As said - i am unsure about this, in particular because the second alternative is based on an extremely long term scenario and it would mean us encouraging mappers to invest enormeous amounts of time into work we expect to be pointless in the long term.  In other words:  I have serious moral scruple about supporting such a course of action.

In any case - since there is an active discussion going on in the community about new tagging ideas in that context (see https://wiki.openstreetmap.org/wiki/Proposed_features/highway%3Dscramble and links on that page) i would be strongly against prejudicing consensus building in that discussion with a change like this at this time.

style/roads.mss
yvecai: Wherever a line dash is specified in a filter, I'm afraid I have to 'overwrite' it with this new [int_difficulty = 'demanding'] . I could probably refactor the code in some way, but been as it is, I doubt we gain in clarity.

yvecai:
"tags which are all rather poorly defined and documented", really ?
https://wiki.openstreetmap.org/w/index.php?title=Key:sac_scale&oldid=124302 (2008->2022)

"prejudicing consensus building in that discussion with a change like this at this time" I also agree this is a major blocker. Anyway I just wanted to show that things could also evolve here. Also because the status quo of issue #1500 (2015 !!) is probably not completely sinless in the recent tagging proposals around more demanding hiking trails.

style/roads.mss
imagico: You obviously would have to make the other dasharrays conditional to the  attributes they are used for.

But i agree your version makes the logic of int_difficulty overriding int_surface most explicit.

PR #4695, opened 2022-09-24
fix: move tourism-boundary above the road layers

### LINKED ISSUE #4682: Theme park outline not rendering correctly over highway areas
### Expected behavior
Theme park (`tourism=theme_park`) and zoo (`tourism=zoo`) outlines should render on top of everything else, since they define a sort of boundary. This would be similar to how protected area outlines currently render.

### Actual behavior
Theme park outlines appear to render below highways, but above other objects like buildings and piers, leaving to an undesirable broken outline. This is especially noticeable under pedestrian areas (`highway=footway`+`area=yes`), but appears to be true for all `highway=*`, whether they are mapped as ways or areas.

### Screenshots with links illustrating the problem
Screen Shot 2022-09-11 at 11 11 50 AM

https://www.openstreetmap.org/way/173532595, Pacific Park, Santa Monica, CA, USA

Screen Shot 2022-09-11 at 11 13 15 AM

https://www.openstreetmap.org/#map=19/33.80960/-117.91858, Disneyland, Anaheim, CA, USA

Screen Shot 2022-09-11 at 11 17 33 AM

https://www.openstreetmap.org/#map=19/33.80960/-117.91858, Universal Studios, Los Angeles, CA, USA

Compare:
Screen Shot 2022-09-11 at 11 21 45 AM

https://www.openstreetmap.org/#map=19/33.80960/-117.91858, Circle X Ranch, Thousand Oaks, CA, USA

Notice how the protected area boundary goes on top of the roadways. I find this desirable behavior, though it's been proposed to change it in #4127, due to rendering issues with closely parallel boundaries (#3685)

tjur0:
Fixes #4682

- move the "tourism-boundary" between "power-line" and "protected-areas" as the function and rendering of protected-areas and tourism-boundary are very similar.

[Cedar Point](https://www.openstreetmap.org/relation/11491811)

Before
z16
boundary-before-z16
z19
boundary-before

After
z16
boundary-after-z16
z19
boundary-after

### DIFF
--- a/project.mml
+++ b/project.mml
@@ -523,22 +523,6 @@ Layer:
         ) AS landuse_overlay
     properties:
       minzoom: 8
-  - id: tourism-boundary
-    geometry: polygon
-    <<: *extents
-    Datasource:
-      <<: *osm2pgsql
-      table: |-
-        (SELECT
-            way,
-            way_area/NULLIF(POW(!scale_denominator!*0.001*0.28,2),0) AS way_pixels,
-            tourism
-          FROM planet_osm_polygon
-          WHERE tourism = 'theme_park'
-            OR tourism = 'zoo'
-        ) AS tourism_boundary
-    properties:
-      minzoom: 10
   - id: barriers
     geometry: linestring
     <<: *extents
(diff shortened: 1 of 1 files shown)


imagico (approved):
Looks good and works as intended.

It could be worth considering to move all the semitransparent boundary line signatures (this, protected areas and admin boundaries) together.  The current setup with the power lines in between is kind of odd.

tjur0:
> The current setup with the power lines in between is kind of odd.

Agreed, I’ll make a ticket for this. I do think we need to try and keep bug fixes and refactor prs sperate.

PR #4702, opened 2022-10-01
Use a variable for opacity of private amenity points

sommerluk:
- Use a variable instead of individual values for the opacity of private amenity points. 

- No rendering changes.


sommerluk:
Rebased and resolved merge conflicts.

